PR #1093, opened 2023-08-02
Improve handling of unselecting languanges in composer language menu

### LINKED ISSUE #1075: Unselecting languanges in Post Languange selection is confusing.
**Describe the bug**
There are and will be a lot of bilingual and multilingual users coming to Bluesky so there will be users who enabled multiple languages in their post by default. However, there might moments where said user wants to only post in English, or other languages they are proficient in. 

Unselecting languages in Post Selection leads to a weird behaviour where when the user clicks or taps on the checkmark next to the language name, instead of unselecting it, it will select said language as the post language

**To Reproduce**

1. Pick multiple languages in post-composer.
2. Try to unselect any languages by clicking the checkbox of said language.

**Expected behavior**
When the user unselects a language by clicking or tapping, it should unselect the language and not selecting the one they just clicked. 

**Screenshots**
the error

**Details**

 - Platform: Desktop Web (Microsoft Edge 115.0.1901.183)
 - Platform version: 1.41.0
 - App version: 1.41.0 (2) production

estrattonbailey:
Fixes #1075 

Revised approach after discussion in Slack, so deleted my previous description.

New approach is to store the post language as a comma separated string of 2-character language codes on the `postLanguage` prop. This string is split and still accessible by the previous `postLanguages` prop via a getter.

Also added here is a rolling history of the language combinations the user has selected.

https://github.com/bluesky-social/social-app/assets/4732330/602db981-b606-4af5-bf70-ea285718c5cd

Also updated to use a Radio icon at Ansh's recommendation:
Screen Shot 2023-08-22 at 3 47 50 PM

### DIFF
--- a/src/state/models/ui/preferences.ts
+++ b/src/state/models/ui/preferences.ts
@@ -372,7 +372,7 @@ export class PreferencesModel {
   }
 
   getReadablePostLanguages() {
-    const all = this.postLanguages.filter(Boolean).map(code2 => {
+    const all = this.postLanguages.map(code2 => {
       const lang = LANGUAGES.find(l => l.code2 === code2)
       return lang ? lang.name : code2
     })
--- a/src/state/models/ui/preferences.ts
+++ b/src/state/models/ui/preferences.ts
@@ -55,7 +55,7 @@ export class PreferencesModel {
   adultContentEnabled = !isIOS
   contentLanguages: string[] = deviceLocales || []
   postLanguage: string = deviceLocales ? deviceLocales.join(',') : 'en'
-  postLanguageHistory: string[] = []
+  postLanguageHistory: string[] = DEFAULT_LANG_CODES
   contentLabels = new LabelPreferencesModel()
   savedFeeds: string[] = []
   pinnedFeeds: string[] = []
--- a/src/state/models/ui/preferences.ts
+++ b/src/state/models/ui/preferences.ts
@@ -54,7 +54,7 @@ export class LabelPreferencesModel {
 export class PreferencesModel {
   adultContentEnabled = !isIOS
   contentLanguages: string[] = deviceLocales || []
-  postLanguage: string = deviceLocales ? deviceLocales.join(',') : 'en'
+  postLanguage: string = deviceLocales[0] || 'en'
   postLanguageHistory: string[] = DEFAULT_LANG_CODES
   contentLabels = new LabelPreferencesModel()
   savedFeeds: string[] = []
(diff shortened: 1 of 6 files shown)


pfrazee:
This actually doesn't fix #1075 because the issue is which languages are shown as options. If you select from one of the non-default options using the deeper menu, it will show as an option. However once you deselect that non-default option, it is "forgotten." This is why we need to retain a rolling history of past selections to displace the defaults.

I'm also -1 to this change because it's less common that people post in multiple languages, and so making the quick-UI go for multi-language isn't ideal. (IMO It's likely to cause more people to incorrectly tag posts as multilingual.)

pfrazee:
I misread the original complaint. Reopening and discussing in slack

src/state/models/ui/preferences.ts
```diff
@@ -45,7 +48,8 @@ export class LabelPreferencesModel {
 export class PreferencesModel {
   adultContentEnabled = !isIOS
   contentLanguages: string[] = deviceLocales || []
-  postLanguages: string[] = deviceLocales || []
+  postLanguage: string = deviceLocales ? deviceLocales.join(',') : 'en'
```
pfrazee, line 51: we should take the first lang instead of defaulting to a multilang when deviceLocales are available

src/state/models/ui/preferences.ts
```diff
@@ -95,16 +100,35 @@ export class PreferencesModel {
...
         this.contentLanguages = deviceLocales
       }
-      // check if post languages in preferences exist, otherwise default to device languages
+      if (hasProp(v, 'postLanguage') && typeof v.postLanguage === 'string') {
+        this.postLanguage = v.postLanguage
+      } else {
+        /*
+         * Check if the OLD `postLanguages` in preferences exist, set new field
+         * to old value if exists, otherwise use deviceLocales
+         */
```
pfrazee, line 109: now that I think about this, it may not be worth looking for the old postLanguages. This is a small enough UI affordance that we could "forget" what was used last and it's not that big of a deal

src/state/models/ui/preferences.ts
```diff
@@ -299,20 +324,47 @@ export class PreferencesModel {
...
-      this.postLanguages = this.postLanguages.filter(lang => lang !== code2)
+      this.postLanguage = this.postLanguages
+        .filter(lang => lang !== code2)
+        .join(',')
     } else {
-      this.postLanguages = this.postLanguages.concat([code2])
+      // sort alphabetically for deterministic comparison in context menu
+      this.postLanguage = this.postLanguages
+        .concat([code2])
+        .sort((a, b) => (a > b ? 1 : -1))
```
pfrazee, line 351: Very minor deal that probably wont ever matter for these comparisons but `(a, b) => a.localeCompare(b)` is more reliable

src/state/models/ui/preferences.ts
```diff
@@ -299,20 +324,47 @@ export class PreferencesModel {
...
+  }
+
+  /**
+   * Saves whatever language codes are currently selected into a history array,
+   * which is then used to populate the language selector menu.
+   */
+  savePostLanguageToHistory() {
+    this.postLanguageHistory = [this.postLanguage]
+      .concat(this.postLanguageHistory)
+      .slice(0, 6)
```
pfrazee, line 367: Need to remove duplicates

src/state/models/ui/preferences.ts
estrattonbailey: Sgtm!

src/state/models/ui/preferences.ts
estrattonbailey: Removed dupe and prepended value to start of array 👍

pfrazee:
Okay we're looking good!

WDYT about updating the language selector modal to have the multilingual option?

ansh:
Small comment: I would suggest using a radio icon rather than a checkbox because radio is implicitly: "only one" whereas checkbox can be "one or more"

pfrazee (approved):
Good stuff, thanks Eric!

PR #1095, opened 2023-08-03
[APP-786] Native notifications

ansh:
Related to https://github.com/bluesky-social/atproto/pull/1428
Pull the above branch, build using `yarn` and then use the `dist` from the `api` package instead of the current one

To test on Android, you also require `google-services.json` from Firebase

(lint errors exist because of the above)

`addNotificationResponseReceivedListener` will not run in dev mode, so please start the frontend server in production with `npx expo start --no-dev --minify`

There are no regressions in this PR, however there are a few improvements that can be made in the future, but aren't implemented because they need more discussion:
- [ ] Deeplink to a post when clicking on a notification (right now it just goes to the notifications screen)
- [ ] Display images in notifications when the post has an image
- [ ] Potentially display notifications when app is in foreground


pfrazee:
PR seems fine so far. Let's get backend merged and the API changes published, then we can test this in testflight. Until the backend stuff is in prod, testflight doesn't do us much good.

pfrazee (approved):
🚀 🔔

PR #1097, opened 2023-08-03
Use a post and handle-resolution cache to enable quick postthread loading

pfrazee:
- Adds a handle-resolution cache
- Adds a posts cache
- Uses the posts cache to optimistically render post threads

https://github.com/bluesky-social/social-app/assets/1270099/f57a0997-2d18-4ebe-bc7a-b79f94a1eef8

### DIFF
--- a/src/state/models/cache/handle-resolutions.ts
+++ b/src/state/models/cache/handle-resolutions.ts
@@ -1,0 +1,5 @@
+import {LRUMap} from 'lru_map'
+
+export class HandleResolutionsCache {
+  cache: LRUMap<string, string> = new LRUMap(500)
+}
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -29,10 +29,24 @@ export async function resolveName(store: RootStoreModel, didOrHandle: string) {
   if (didOrHandle.startsWith('did:')) {
     return didOrHandle
   }
-  const res = await store.agent.resolveHandle({
-    handle: didOrHandle,
-  })
-  return res.data.did
+
+  // we run the resolution always to ensure freshness
+  const promise = store.agent
+    .resolveHandle({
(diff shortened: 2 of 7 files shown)


pfrazee:
One big TODO: when viewing a reply, we need to make sure we maintain the current scroll position when parents get loaded

pfrazee:
Another improvement would be to cache the quoted posts

src/state/models/cache/handle-resolutions.ts
```diff
@@ -0,0 +1,5 @@
+import {LRUMap} from 'lru_map'
+
+export class HandleResolutionsCache {
```
ansh, line 3: `lru_map` is a good solution for this but I don't think we would want this to be clogging up the JS thread. Not sure about the performance implications but it is something to be wary of (re: memory issues). Could use https://reactnative.dev/docs/asyncstorage for async storage or https://github.com/mrousavy/react-native-mmkv for sync storage

ansh: I will caveat this and say its _probably_ fine

ansh (approved):
Small comment

src/state/models/cache/handle-resolutions.ts
pfrazee: Wouldnt the context switching of messaging the main process be worse than the processing overhead of the LRU manager?

ansh: I don't think I know the internals of react native to give an answer I am confident about with this matter.

But one more thing to consider is that the `lru_map` will be wiped out on every app restart

pfrazee: Yeah for now I'm good with that, we dont want to persist that long

PR #1106, opened 2023-08-04
fix feed UI alignment issues

estrattonbailey:
Noticed some mis-alignment of feed item components the other day and decided to fix it quickly. Screenshots below, iOS looks good too, have not tested on Android (need to fix my env apparently).

To fix these, I basically rearranged the components to create new layout containers. That way, instead of `absolute` positioning, we can let things flex and move based on the height of surrounding content.

For review, have a quick scroll through some of your feeds and confirm there aren't any edge cases I missed.

Before -> After
Screen Shot 2023-08-04 at 11 00 10 AM
Screen Shot 2023-08-04 at 11 00 38 AM
Screen Shot 2023-08-04 at 11 05 53 AM
Screen Shot 2023-08-04 at 11 07 23 AM

### DIFF
--- a/src/view/com/posts/FeedItem.tsx
+++ b/src/view/com/posts/FeedItem.tsx
@@ -352,6 +352,7 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     marginTop: 2,
     marginBottom: 4,
+    marginLeft: -20,
   },
   includeReasonIcon: {
     marginRight: 4,
--- a/src/view/com/posts/FeedItem.tsx
+++ b/src/view/com/posts/FeedItem.tsx
@@ -32,10 +32,12 @@ import {makeProfileLink} from 'lib/routes/links'
 export const FeedItem = observer(function ({
   item,
   isThreadChild,
+  isThreadLastChild,
   isThreadParent,
 }: {
   item: PostsFeedItemModel
   isThreadChild?: boolean
+  isThreadLastChild?: boolean
   isThreadParent?: boolean
   showReplyLine?: boolean
 }) {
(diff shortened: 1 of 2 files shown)


src/view/com/posts/FeedSlice.tsx
```diff
@@ -57,6 +57,9 @@ export const FeedSlice = observer(
             item={item}
             isThreadParent={slice.isThreadParentAt(i)}
             isThreadChild={slice.isThreadChildAt(i)}
+            isThreadLastChild={
+              slice.isThreadChildAt(i) && slice.items.length === i + 1
+            }
```
estrattonbailey, line 62: Is this true 100% of the time?

src/view/com/posts/FeedItem.tsx
```diff
@@ -208,6 +225,18 @@ export const FeedItem = observer(function ({
             avatar={item.post.author.avatar}
             moderation={item.moderation.avatar}
           />
+          {isThreadParent && (
+            <View
```
estrattonbailey, line 229: Within the avatar container, this flexes to fill the remaining height

src/view/com/posts/FeedItem.tsx
```diff
@@ -154,51 +160,62 @@ export const FeedItem = observer(function ({
...
-                  sanitizeHandle(item.reasonRepost.by.handle),
-              )}
-              href={makeProfileLink(item.reasonRepost.by)}
-            />
-          </Text>
-        </Link>
-      )}
       <PostSandboxWarning />
+
+      <View style={{flexDirection: 'row', gap: 10, paddingLeft: 8}}>
```
estrattonbailey, line 165: Basically sits on top of the post and flexes height, which at minimum is 12px due to the padding on the `reasonRepost` wrapper, which itself can expand if needed.

src/view/com/posts/FeedSlice.tsx
pfrazee: I think so yeah

pfrazee (changes requested):
I'm +1 on everything except the reason line -- I think it flows visually much better to align the text but have the icon offset to the left. Can we use relative positioning to maintain that?

estrattonbailey:
Oh, thought it looked weird initially but I agree now that I see it. Blew right through chesterton's fence on that one hehe

Fixed 👍 

Screen Shot 2023-08-04 at 2 02 01 PM

pfrazee (approved):
Great! Nice cleanup

PR #1108, opened 2023-08-04
Fix repeated firing of scroll into view on large threads

pfrazee:
Fixes an issue in large threads where the windowing behavior would fire `onContentSizeChange` when scrolling, cause it to "fight" the user

https://github.com/bluesky-social/social-app/assets/1270099/5883d7f1-0b71-4bff-a378-31729120657e


estrattonbailey (approved):
Nice!

PR #1109, opened 2023-08-04
fix post thread feedback regression

estrattonbailey:
This [was previously](https://github.com/bluesky-social/social-app/commit/b154d3ea2#diff-0980d61e6f9f147f46edbf4f397b45ec438ae46571a0f1e3cf39748ca9061f0fL162) `PostHider`, which has `noFeedback` and `accessbility={false}`, so I added that back.


pfrazee (approved):
Ah excellent, thanks for catching that

PR #1110, opened 2023-08-04
Add more shades of blue for random app passwords.

alimony:
I've noticed collisions every now and then when creating app passwords using the default random name. This adds more shades of blue (sourced from web colours and Wikipedia) to the list of potential names to reduce the risk of collision from ~3% to ~1%.


pfrazee:
I appreciate the suggestion but some of these are a bit weird. Maybe we should just check if the name is in use, and if it is roll again (or append a number)?

alimony:
Yeah, that's definitely the better solution, this was just the 10-minute quick and dirty fix. 🙃

PR #1125, opened 2023-08-07
Fix: dont use profile preview modal on mobile web

pfrazee:
The profile preview modal now depends on a swipe gesture which web doesnt support. This PR just disables it on mobile web.


ansh (approved):
👍🏽

estrattonbailey (approved):

PR #1132, opened 2023-08-08
fix long thread styling

estrattonbailey:
Noticed that my fixes from #1106 caused a mis-alignment of this "View full thread" section, so I applied the same styling to fix.

Before and after

Screen Shot 2023-08-08 at 9 08 43 AM
Screen Shot 2023-08-08 at 9 49 31 AM


pfrazee (approved):
🚀

PR #1133, opened 2023-08-08
fix thread spacing and reply lines

estrattonbailey:
Noticed this last week too, decided to just knock it out using the same styling changes I made in #1106. In addition to reply line alignment, some replies were not styled as replies (no reply line), and some posts were missing borders between them. 

Part of the issue here was a bug in how we set `_showParentReplyLine`. I've edited that here and I believe it's correct now, but @ansh should have a look to confirm.

Note for reviewers: turn off whitespace to more easily see the actual changes.

Before and afters:
Screen Shot 2023-08-08 at 11 33 59 AM
Screen Shot 2023-08-08 at 11 31 33 AM
Screen Shot 2023-08-08 at 11 32 26 AM

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -159,171 +159,189 @@ export const PostThreadItem = observer(function PostThreadItem({
 
   if (item._isHighlightedPost) {
     return (
-      <Link
-        testID={`postThreadItem-by-${item.post.author.handle}`}
-        style={[styles.outer, styles.outerHighlighted, pal.border, pal.view]}
-        noFeedback
-        accessible={false}>
-        <PostSandboxWarning />
-        <View style={styles.layout}>
-          <View style={styles.layoutAvi}>
-            <PreviewableUserAvatar
-              size={52}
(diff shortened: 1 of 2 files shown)


src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -159,171 +157,189 @@ export const PostThreadItem = observer(function PostThreadItem({
...
+                    backgroundColor: pal.colors.replyLine,
+                  },
+                ]}
+              />
+            </View>
           </View>
-          <View style={styles.layoutContent}>
-            <View style={[styles.meta, styles.metaExpandedLine1]}>
-              <View style={[s.flexRow]}>
+        )}
```
estrattonbailey, line 175: Added a little shim with a reply line on top of the highlighted post, hence the fragment and the huge whitespace edit in this PR.

estrattonbailey: You can turn off whitespace using the gear menu in the files view, much easier to read

pfrazee (approved):
🚀 looks right, good fix on that parent reply line

PR #1136, opened 2023-08-08
Fix to react-key uniqueness in feeds

pfrazee:
Turns out the react key I chose for https://github.com/bluesky-social/social-app/pull/1113 wasn't unique enough (it shows up in profile pages when people repost themselves). This fixes that, unless somebody manages to repost themselves with the exact same timestamp.


ansh (approved):
Good change!

PR #1137, opened 2023-08-08
Adds profile media tab

estrattonbailey:
Depends on https://github.com/bluesky-social/atproto/pull/1439

Adds a "Media" tab to the profile screen, and replaces the client-side filtering of replies with a server-side filter.

Little shaky rn, we need a smoother transition:
https://github.com/bluesky-social/social-app/assets/4732330/b2558677-5197-4050-b8b1-d2147ba75d65


pfrazee:
Okay this is looking great, two things we need address:

1. Switching tabs on mobile is a bit screwy. The loading state causes the spinner to show up briefly, jogging the UI a bit. We do need to indicate the loading state. There's a couple ways to handle this; it _might_ be worth having 3 different feed models rather than swapping out the one.
2. Mobile now has the ability to overflow the tab headers (smaller devices and/or larger accessibility text). We can fix that either by using a scrollview or by trying to simplify the tab names (eg "Post & replies" -> "w/Replies")

pfrazee (approved):
🚀 nice work!

PR #1138, opened 2023-08-08
Add alerts to embeds

pfrazee:
Embeds currently don't have alerts when they should:

CleanShot 2023-08-08 at 12 37 48@2x
CleanShot 2023-08-08 at 12 34 11@2x

This PR fixes that:

CleanShot 2023-08-08 at 12 33 37@2x

This also applies to a "downgrading" behavior† which applies to self quote posts:

CleanShot 2023-08-08 at 12 45 59@2x

† Moderation UI "downgrades" from filters & blurs to just alerts for your own content. This is to avoid confusingly absent content, but it's currently a little inconsistent -- it correctly downgrades on quote posts but not on images. I'm going to tweak this behavior a bit, but for now it's useful to get this fix done.

### DIFF
--- a/__e2e__/mock-server.ts
+++ b/__e2e__/mock-server.ts
@@ -144,7 +144,7 @@ async function main() {
           await server.mocker.labelProfile('porn', 'porn-profile')
           await server.mocker.labelPost(
             'porn',
-            await server.mocker.createPost('porn-posts', 'porn post'),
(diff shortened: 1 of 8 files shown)


estrattonbailey (approved):
LGTM, was able to run e2e mock server and seeds locally to confirm

Issue #1081, opened 2023-07-31
Add to mute lists dialog clears selection after I check a checkbox


mackuba:
**Describe the bug**

The "add user to mute lists" dialog seems to be loading the current state of that user separately from the UI of the dialog itself and reloading the dialog a moment after it appears. The result is that when I open the dialog and check a checkbox, the checkbox is unchecked and I need to check it again.

**To Reproduce**

This happens very consistently and I've seen it many times.

1. Go to someone's profile.
2. Choose "Add to lists" from the menu.
3. A dialog appears with a list of your mutelists and empty checkboxes.
4. Check one of the checkboxes within a few seconds.
5. A few seconds after the dialog appeared, the checkbox you've checked is unchecked again.

**Expected behavior**

If it takes a few seconds to load current selection state, then the dialog should not display its contents until that data has loaded and the dialog can be shown with the checkboxes already selected correctly (or not selected, if the user hasn't been added to any yet). But it should not reset the dialog state after I have already selected some checkboxes.

**Details**

 - Platform: macOS 12.6.8, Safari 16.6

pfrazee:
Ew thanks for the report, assigning to @estrattonbailey

estrattonbailey:
Hey @mackuba, thanks for this. I haven't been able to reproduce yet, even trying flaky/slow connections in various browsers. Are you still seeing this issue on your end?

mackuba:
Ah, I see what's going on. Yeah, I'm still seeing this.

I've checked the web inspector, and it looks like the dialog is loading my complete list of `listitem` records, which seems kind of… suboptimal? So it probably isn't noticeable if someone has just one list with a few entries, but if you have a few hundred entries in total on your list(s), then it takes a bit of time to load all of this, just to check on which lists this specific user appears.

estrattonbailey:
Thanks for the quick response! Copy that, I'll do some more investigating 🫡

estrattonbailey:
@mackuba we've merged an improved loading state that should patch the wonky behavior you've noticed. That'll be included in the next release. In the longer term, there are some backend improvements in the backlog that will make this more performant, but TBD on when we get to that.

Thanks again for this report! Will close this for now, but feel free to open again if you continue to see issues.

PR #1140, opened 2023-08-09
wait for list memberships to load before becoming interactive

estrattonbailey:
A small improvement to help address #1081 until we have time to improve the performance of this data fetching.

Adds an improved loading spinner and grays out interactive elements until everything has loaded.

This _removes_ the spinner from the `ListsList` component, which means `ModerationMuteListsScreen` loses it too (the component is reused there). However, the `FlatList` skeleton UI state already obscures the loading spinner and the spinner's disappearance was a little jumpy sometimes anyway. I actually think it's better off without it, but I may be missing an edge case.

https://github.com/bluesky-social/social-app/assets/4732330/28bac020-97a9-4c41-89a5-5eff8fb8689d

For reference, here's what the other load state now looks like:

https://github.com/bluesky-social/social-app/assets/4732330/68974a23-77f2-440d-a6fd-3399fd4c6552

### DIFF
--- a/src/view/com/lists/ListsList.tsx
+++ b/src/view/com/lists/ListsList.tsx
@@ -1,6 +1,5 @@
 import React, {MutableRefObject} from 'react'
 import {
-  ActivityIndicator,
   RefreshControl,
   StyleProp,
   StyleSheet,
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
🚀 looks good!

PR #1141, opened 2023-08-09
Add self-labeling controls

pfrazee:
Adds the ability to set labels on your own post. Currently just some NSFW labels, will add more over time.

https://github.com/bluesky-social/social-app/assets/1270099/d8a67b20-c7e9-4100-bb4e-8ba265ee936a

Before merging:

- [ ] Wait for backend deploy on prod
- [x] E2E tests

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -4,6 +4,7 @@ import {
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
   AppBskyRichtextFacet,
+  ComAtprotoLabelDefs,
   ComAtprotoRepoUploadBlob,
   RichText,
 } from '@atproto/api'
@@ -77,6 +78,7 @@ interface PostOpts {
   }
   extLink?: ExternalEmbedDraft
   images?: ImageModel[]
+  labels?: string[]
   knownHandles?: Set<string>
   onStateChange?: (state: string) => void
   langs?: string[]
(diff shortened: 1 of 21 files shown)


src/lib/api/index.ts
```diff
@@ -234,6 +236,15 @@ export async function post(store: RootStoreModel, opts: PostOpts) {
...
   }
 
+  // set labels
+  let labels: ComAtprotoLabelDefs.SelfLabels | undefined
+  if (opts.labels?.length) {
+    labels = {
+      $type: 'com.atproto.label.defs#selfLabels',
+      values: opts.labels.map(val => ({val})),
+    }
+  }
```
estrattonbailey, line 246: Just general question: should we have a SDK to ease the creation of lexicon records like this?

estrattonbailey (approved):
LGTM, tested in staging

src/lib/api/index.ts
pfrazee: Possibly, but I think that can wait for later

skunecuscre1:
This is great!

Small nitpick but could you reconsider the 🥵 emoji? It kinda has a bad vibe, I'd suggest replacing it with 🔥 instead.

pfrazee:
Hmm! The 🔥 isnt quite right either because it's a little too _positive_ and the point is not to be positive or negative, just descriptive.

I'm not really sure what's a better fit than 🥵 (which is commonly used to mean horny). Maybe "🍆🍑"?

Maybe a better move is to just show 🔞 on the ones that have that effect (adult only cant be disabled within the iOS app)?

DavidBuchanan314:
The textual descriptions are pretty clear to me, but the emojis are vague, and vary a lot depending on platform/font, and cultural interpretation - I would question whether an emoji is needed at all

pfrazee:
Needed or helpful, no. They were just meant to add some fun

ShreyanJain9:
The existing emojis mostly look fine to me, since they're not a hugely prominent part of the UI. That said, some people may find them annoying, so maybe they should be togglable off. But it also doesn't make sense to have too many toggles in the app? Idk just make of that what you will

pfrazee:
Eh I'm going to just drop them. They're fun but I don't want to imply anything weird with them

ShreyanJain9:
Sure, the emojis aren't a huge deal anyway, some third party app can add those later if they want

The actual functionality looks pretty good to me 🙂

pfrazee:
See https://github.com/bluesky-social/social-app/pull/1147

PR #1144, opened 2023-08-10
Fix issue where small web browsers were being miscategorized as mobile web

Kingles:
Mostly done to fix an issue where currently links don't show as / behave strictly as (no underline, no cursor change) links when the agent's viewport is smaller than 1230px, rather than if it is a mobile device by our best guess. 

Definitely will have some other changes but getting a hopefully more accurate flag in early seems worth it. A less controversial change may be to adjust `Link` to switch based on `isNative` rather than the current `isDesktopWeb` 🤷🏼.

I've been testing locally and things seem to be great after the change, but I've of course got limited context here. Thanks for the opportunity to contribute!

### DIFF
--- a/src/platform/detection.ts
+++ b/src/platform/detection.ts
@@ -10,7 +10,7 @@ function hasTouchscreen() {
   if ('maxTouchPoints' in navigator) {
     return navigator.maxTouchPoints > 0
   } else if ('msMaxTouchPoints' in navigator) {
-    // @ts-ignore navigator should exist
+    // @ts-ignore msMaxTouchPoints is nonstandard/deprecated but may exist
     return navigator.msMaxTouchPoints > 0
   } else {
     const mQ = matchMedia?.('(pointer:coarse)')
@@ -35,11 +35,10 @@ export const isAndroid = Platform.OS === 'android'
 export const isNative = isIOS || isAndroid
 export const isWeb = !isNative
 export const isMobileWebMediaQuery = 'only screen and (max-width: 1230px)'
-export const isMobileWeb =
-  isWeb &&
-  // @ts-ignore we know window exists -prf
-  global.window.matchMedia(isMobileWebMediaQuery)?.matches &&
-  hasTouchscreen()
+export const shouldUseMobileLayout = window.matchMedia(
+  isMobileWebMediaQuery,
+)?.matches
+export const isMobileWeb = isWeb && hasTouchscreen() && shouldUseMobileLayout
 export const isDesktopWeb = isWeb && !isMobileWeb
 
 export const deviceLocales = dedupArray(
--- a/src/platform/detection.ts
+++ b/src/platform/detection.ts
@@ -2,6 +2,34 @@ import {Platform} from 'react-native'
 import {getLocales} from 'expo-localization'
 import {dedupArray} from 'lib/functions'
 
+/**
+ * Attempt to ascertain if the agent is a touchscreen device.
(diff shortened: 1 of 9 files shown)


Kingles:
Drafting this while I chase down some CSS issues; mostly with padding on `Feed` instances

src/platform/detection.ts
```diff
@@ -2,6 +2,34 @@ import {Platform} from 'react-native'
...
 
+/**
+ * Attempt to ascertain if the agent is a touchscreen device.
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent
+ */
+function hasTouchscreen() {
+  if ('maxTouchPoints' in navigator) {
+    return navigator.maxTouchPoints > 0
+  } else if ('msMaxTouchPoints' in navigator) {
+    // @ts-ignore navigator should exist
```
ansh, line 13: An `if (navigator)` or `if (isWeb)` check should make this confirmed

src/platform/detection.ts
```diff
@@ -2,6 +2,34 @@ import {Platform} from 'react-native'
...
+    return navigator.msMaxTouchPoints > 0
+  } else {
+    const mQ = matchMedia?.('(pointer:coarse)')
+    if (mQ?.media === '(pointer:coarse)') {
+      return !!mQ.matches
+    } else if ('orientation' in window) {
+      return true // deprecated, but good fallback
+    } else {
+      // Only as a last resort, fall back to user agent sniffing
```
ansh, line 22: Why is `user-agent` sniffing the last resort?

Kingles: It's mostly unreliable and not considered best practice.

Refs:
https://javascript.plainenglish.io/3-facts-you-should-know-about-user-agent-a476251ee3b5
https://css-tricks.com/browser-detection-is-bad/
https://stackoverflow.com/questions/661213/why-is-browser-sniffing-not-a-recommended-practice
https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent


That being said I've done it in most cross-platform projects I've worked on over the years so...... ya know. Just thought it was worth keeping the comment in there.

Kingles:
Alright, ready for review now I think! I ended up adding a `shouldUseMobileLayout` variable to account for places where we still want the mobile look/feel but aren't actually mobile. 

Def not sold on the name, but the concept seems valuable/relevant if we're going to keep the mobile-esque view for smaller screens in general.

Feels really good locally, and I haven't seen any additional issues!

src/platform/detection.ts
ansh: Appreciate it!

src/view/com/auth/SplashScreen.web.tsx
```diff
@@ -6,7 +6,7 @@ import {ErrorBoundary} from 'view/com/util/ErrorBoundary'
 import {s, colors} from 'lib/styles'
 import {usePalette} from 'lib/hooks/usePalette'
 import {CenteredView} from '../util/Views'
-import {isMobileWeb} from 'platform/detection'
+import {shouldUseMobileLayout} from 'platform/detection'
```
ansh, line 9: `shouldUseMobileWebLayout` might be better since its more specific. cc @pfrazee

pfrazee:
@Kingles So I just redid all of our responsive breakpoint code; I want to work this in but I need to understand the issue a little better. I'm guessing https://github.com/bluesky-social/social-app/issues/1410 might be manifesting the issue this is meant to solve?

The goal is to be deciding what to render based on screen real-estate only -- at least for now. Is our query botching that somehow?

PR #1150, opened 2023-08-10
More useful block-state information in embeds

pfrazee:
This PR updates how blocks are handled with embeds to give more information about what's going on. It also pulls in an updated SDK which causes QPs with block-states to be filtered from feeds.

There are some active discussions about how aggressive blocking should be -- it is now _extremely_ aggressive, applying to everyone (the third state below) and essentially prunes replies by blocked users from the world. These behaviors are not resolved and will get another design pass in the near future. 

# Blocked by author of quoted post

CleanShot 2023-08-10 at 13 43 16@2x
CleanShot 2023-08-10 at 13 43 23@2x

# Blocking author of quoted post

CleanShot 2023-08-10 at 13 43 38@2x
CleanShot 2023-08-10 at 13 43 45@2x

# Author of quoted post or of quote post block each other

CleanShot 2023-08-10 at 13 44 04@2x
CleanShot 2023-08-10 at 13 44 33@2x

### DIFF
--- a/src/lib/moderation.ts
+++ b/src/lib/moderation.ts
@@ -18,13 +18,13 @@ export function describeModerationCause(
   }
   if (cause.type === 'blocking') {
     return {
-      name: 'Blocked Author',
+      name: 'User Blocked',
       description: 'You have blocked this user. You cannot view their content.',
     }
   }
   if (cause.type === 'blocked-by') {
     return {
-      name: 'Blocked',
+      name: 'User Blocking You',
       description: 'This user has blocked you. You cannot view their content.',
     }
   }
--- a/src/lib/moderation.ts
+++ b/src/lib/moderation.ts
@@ -18,14 +18,21 @@ export function describeModerationCause(
   }
   if (cause.type === 'blocking') {
     return {
-      name: 'Blocked User',
+      name: 'Blocked Author',
       description: 'You have blocked this user. You cannot view their content.',
     }
   }
   if (cause.type === 'blocked-by') {
     return {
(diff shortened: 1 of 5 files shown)


src/lib/moderation.ts
```diff
@@ -18,16 +18,23 @@ export function describeModerationCause(
...
     return {
-      name: 'Blocked User',
+      name: 'Blocked Author',
       description: 'You have blocked this user. You cannot view their content.',
     }
   }
   if (cause.type === 'blocked-by') {
     return {
-      name: 'Blocking You',
+      name: 'Blocked',
```
ansh, line 27: This might be confusing. Since 'User Blocks You' is used in `ModerationDetails.tsx` I suggest we keep this as 'Blocked you'

src/view/com/modals/ModerationDetails.tsx
```diff
@@ -29,11 +29,15 @@ export function Component({
     description =
       'Moderator has chosen to set a general warning on the content.'
   } else if (moderation.cause.type === 'blocking') {
-    name = 'Account Blocked'
+    name = 'User Blocked'
     description = 'You have blocked this user. You cannot view their content.'
   } else if (moderation.cause.type === 'blocked-by') {
-    name = 'Account Blocks You'
+    name = 'User Blocks You'
     description = 'This user has blocked you. You cannot view their content.'
```
ansh, line 36: Might be useful to link to the moderation screen here

ansh (approved):
Works as expected on all platforms 👍🏽

PR #1157, opened 2023-08-14
🗺️ fix an issue where web devices that see the Drawer couldn't access Profile from said Drawer

Kingles:
Previously, when using the web and the viewport was smaller than 1230px, nothing would happen when the user clicked on the "Profile" option in the Drawer component. This fixes that!

* Pulled over the screen props into `FlatNavigator` from the `MyProfileTabNavigator` directly above.
* Added the appropriate type to FlatNavigatorParams
* Kept the `ts-ignore` from before, as there still seems to be a type issue there.


ansh:
Thank you for creating this @Kingles !

There were a few changes that I needed to make to make everything work well. So, I had to create a new PR to wrap it up. Closing this in favor of that #1437

PR #1158, opened 2023-08-14
[APP-798] add key to `ProfileCardPills` render method

ansh:
Was getting this error when `ProfileCardPills` was used

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ProfileCardPills`. 
```

This just fixes that.

### DIFF
--- a/src/lib/moderation.ts
+++ b/src/lib/moderation.ts
@@ -92,3 +92,16 @@ export function isCauseALabelOnUri(
   }
   return cause.label.uri === uri
 }
+
+export function getModerationCauseKey(cause: ModerationCause): string {
+  const source =
+    cause.source.type === 'labeler'
+      ? cause.source.labeler.did
+      : cause.source.type === 'list'
+      ? cause.source.list.uri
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):

pfrazee:
I'm not 100% sure that will product a unique key, actually. A labeler could assign multiple labels to something. The question is just whether all of them would show.

The safer move would be to distill down the key with a utility function in `lib/moderation.ts`. This is off the top of my head, you'll need to tune it up (types will guide ya)

```typescript
export function getModerationCauseKey(cause: ModerationCause): string {
  const source = cause.source.type === 'labeler' ? cause.source.labeler.did : cause.source.type === 'list' ? cause.source.list.uri : 'user'
  if (cause.type === 'label') {
    return `${source}:label:${cause.label.val}`
  }
  return `${source}:${cause.type}
}
```

ansh:
Yes, let me create a function to do that

pfrazee:
Went ahead and added it

PR #1159, opened 2023-08-14
Fixes #1124: Trim alt text before saving

### LINKED ISSUE #1124: Seemingly empty alt text due to special whitespace chars
**Describe the bug**
I recently saw a post where the author had added the blank space character (char code: 32) as the only alt text, resulting in the alt banner appearing over the image, but then there's no actual alt text.

so it's currently possible to include only whitespace as the alt text, with no other characters present.

**To Reproduce**

1. make a new post containing an image
2. add the alt text " " <- copy paste this or use String.fromCharCode(32), or any other special whitespace char you wanna test.

**Expected behavior**

if the alt text contains ONLY special whitespace characters, strip them out.

**Screenshots**

image

image

**Details**

 - Platforms: all
 - Platform version: n/a; latest
 - App version: latest

**Additional context**

seems like there's a few ways this could be solved
but the right way is probably just adding a guard to the model or sanitizing for special whitespace chars when the alt text is set: [setAltText in the image model](https://github.com/bluesky-social/social-app/blob/89fc975a150d8c1f79d2a728f9f252fc3438a613/src/state/models/media/image.ts#L123-L125)

ansh:
Trim alt text before saving

### DIFF
--- a/src/lib/strings/helpers.ts
+++ b/src/lib/strings/helpers.ts
@@ -15,3 +15,7 @@ export function enforceLen(str: string, len: number, ellipsis = false): string {
   }
   return str
 }
+
+export function trim(str: string): string {
+  return str.trim()
+}
--- a/src/lib/strings/helpers.ts
+++ b/src/lib/strings/helpers.ts
@@ -15,7 +15,3 @@ export function enforceLen(str: string, len: number, ellipsis = false): string {
   }
   return str
 }
-
-export function trim(str: string): string {
-  return str.trim()
-}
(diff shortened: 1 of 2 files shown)


src/lib/strings/helpers.ts
```diff
@@ -15,3 +15,7 @@ export function enforceLen(str: string, len: number, ellipsis = false): string {
   }
   return str
 }
+
+export function trim(str: string): string {
```
pfrazee, line 19: can we drop this?

pfrazee (approved):
Just need to drop the `trim()` utility function and we're good

src/lib/strings/helpers.ts
ansh: Yes

PR #1163, opened 2023-08-14
fix prefetching lightbox images

estrattonbailey:
We switched to `expo-image` a few months ago in #452, and Expo provides its own [prefetch](https://docs.expo.dev/versions/latest/sdk/image/#prefetchurls) handling. Calling `react-native`'s built-in prefetch may be causing unnecessary calls to our CDN.

### DIFF
--- a/src/view/com/util/post-embeds/index.tsx
+++ b/src/view/com/util/post-embeds/index.tsx
@@ -89,15 +89,8 @@ export function PostEmbeds({
       const openLightbox = (index: number) => {
         store.shell.openLightbox(new ImagesLightbox(items, index))
       }
-      const onPressIn = (index: number) => {
-        const firstImageToShow = items[index].uri
-        Image.prefetch(firstImageToShow)
-        items.forEach(item => {
-          if (firstImageToShow !== item.uri) {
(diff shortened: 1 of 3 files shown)


src/view/com/util/post-embeds/index.tsx
```diff
@@ -95,14 +96,9 @@ export function PostEmbeds({
...
-        const firstImageToShow = items[index].uri
-        RNImage.prefetch(firstImageToShow)
-        items.forEach(item => {
-          if (firstImageToShow !== item.uri) {
-            // First image already prefetched above
-            RNImage.prefetch(item.uri)
-          }
+      const onPressIn = (_: number) => {
+        InteractionManager.runAfterInteractions(() => {
+          Image.prefetch(items.map(i => i.uri))
```
estrattonbailey, line 101: Tried a few things with no luck. Only significant change to the post-embed functionality was here.

Evidently `prefetch` was blocking the UI thread, so `TouchableOpacity` was triggered and then never released/blurred. This seems to fix it.

pfrazee:
👍 Confirmed fixed

PR #1165, opened 2023-08-15
use user.handle response from getProfile, not getSession

estrattonbailey:
Fixes APP-802

`getSession` returns un-validated `user.handle`. `getProfile` on the other hand _does_ validate, and will return `handle.invalid` for un-resolveable handles.


pfrazee (approved):
👍 Looks good!

PR #1179, opened 2023-08-15
Clear caches on session change (close #1177)

### LINKED ISSUE #1177: [APP-806] Clear caches on session change
Some information that's specific to the authed user gets cached now. When the session changes, we need to clear the caches.

<sub>[APP-806](https://linear.app/blueskyweb/issue/APP-806/clear-caches-on-session-change)</sub>

pfrazee:
Closes #1177

Ensures profile and post caches clear on session change to avoid carrying over session-specific state


ansh (approved):
Looks good to me!

Issue #752, opened 2023-05-19
`TouchableNoFeedback` on link embeds looks/feels odd on all platforms


llllvvuu:
**Describe the bug**
It's particularly strange on desktop as it also removes the `<a>` element (which on web provides ctrl+click, URL preview in bottom left corner, etc) and thus the `cursor: pointer`. But it feels weird on mobile as well. I actually thought it was an image and didn't realize I could click it.

**To Reproduce**

1. Sample post: https://bsky.app/profile/bnewbold.bsky.team/post/3jw2p3xhdks2j (in this particular example you can also see with the white image that a divider is missing)

**Expected behavior**

Something to happen when I press (Twitter mobile uses an effect like `TouchableHighlight` - our `Link`s use `TouchableOpacity` which looks fine as well)

**Screenshots**

Desktop:
https://github.com/bluesky-social/social-app/assets/5601392/75680741-4500-4cfb-9036-4603e9d005a6

Android:
https://github.com/bluesky-social/social-app/assets/5601392/bd190b6c-2b30-438b-8dd3-cf4040496fb6

**Details**

 - Platform: all platforms
 - App version: `main` 15c1b6e

**Additional context**

This draft PR demonstrates the difference with the proposed behavior: https://github.com/bluesky-social/social-app/pull/750

It's ready to go if desired.

llllvvuu:
> visual feedback triggers due to the scroll touch

Seems like different apps behave differently here, Discord removes the feedback entirely, Twitter/IG/FB keep it on. It's cleaner with `delayPressIn={130}` which is what [React Native uses by default in most of their components](https://github.com/facebook/react-native/blob/a7f7f8aca9eba32d490275024f7cee359cb3cfcd/packages/react-native/Libraries/Pressability/Pressability.js#L282) (not sure why not in `TouchableOpacity`). This will ensure that if you're scrolling at a reasonable speed, you won't get the feedback flicker. Demo updated: #750 

If we do end up keeping no feedback, we can add `cursor: pointer` to make it look better on web.

If we want a darkening effect instead of a lightening effect, that's slightly more work.

estrattonbailey:
Thanks again for opening this! I think we've solved core intent of this issue and the PR, and we have more link/pressable improvements on the todo list, so gonna close for now. If you feel differently please let us know! 🙂

PR #1181, opened 2023-08-15
Use `Pressable` for most links/embeds

estrattonbailey:
Relates to #752, basically removes pressed styles for links/embeds.

This PR moves to using `Pressable` for interactive elements. Essentially the following is true:
- Links have a `pointer: cursor` and do NOT have a pressed style (yet)
- Links with `href` (sometimes via `asAnchor`) render as `<a>` on the web
- `noFeedback` behaves as it was

`Pressable` is RNs successor to the `Touchable*` components. We _should_ add pressed styles to some elements, but this PR does not include that, since we'll want to consider case-by-case with some of these.

A11y properties and keyboard nav still works the same afaict.

### DIFF
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -115,6 +115,7 @@ export const Link = observer(function Link({
     <TouchableOpacity
       testID={testID}
       style={style}
+      delayPressIn={130}
       onPress={onPress}
       accessible={accessible}
       accessibilityRole="link"
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -9,8 +9,9 @@ import {
   TextProps,
   View,
   ViewStyle,
-  TouchableOpacity,
+  Pressable,
   TouchableWithoutFeedback,
+  TouchableOpacity,
 } from 'react-native'
 import {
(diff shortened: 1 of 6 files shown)


src/view/com/util/Link.tsx
```diff
@@ -9,8 +9,9 @@ import {
   TextProps,
   View,
   ViewStyle,
-  TouchableOpacity,
+  Pressable,
   TouchableWithoutFeedback,
+  TouchableOpacity,
```
estrattonbailey, line 14: Still using this for types

src/view/com/util/post-embeds/YoutubeEmbed.tsx
```diff
@@ -23,9 +23,9 @@ export const YoutubeEmbed = ({
 
   return (
     <Link
+      asAnchor
```
estrattonbailey, line 26: Also noticed that YouTube embeds were rendered as anchors on web, so fixed that

ansh: Nice!

pfrazee (approved):
Aaaawesome. I think this finally fixes that tap animation misfiring.

PR #1182, opened 2023-08-15
use greater of indexedAt or machine clock

### LINKED ISSUE #1064: listNotifications can miss or "duplicate" notifications due to local clock drift
**Describe the bug**

My system clock is a few seconds behind real-time. When I check my notifications, often one or more notifications I've already seen are re-included and highlighted as unread.

Even if my system clock were 100% accurate, there's still a race condition window here.

**To Reproduce**

Set your system clock to something a few minutes into the past, receive a new notification, and see what happens (click onto and away from the notifications tab, multiple times)

**Expected behavior**

On repeated visits to the notification tab, a previously-seen notification should never be highlighted as unseen.

**Additional context**

The last-seen timestamp sent to the server via `app.bsky.notification.updateSeen` is ultimately based on the local clock of the client

https://github.com/bluesky-social/social-app/blob/44b3929240736ead1740efa791e9f4aaf792f2f4/src/state/models/feeds/notifications.ts#L349

whereas the server uses its own clock to determine notification unseen-ness when returning notifications via `app.bsky.notification.listNotifications`.

I'm not sure what the best fix for this would be - perhaps the `listNotifications` query ought to return the server's timestamp (or maybe an opaque cursor-like string) in its response (in which case, perhaps this issue ought to be moved to the atproto repo)

Edit: not 100% sure about the above description, I'm testing now and getting some weird results (I set my clock backwards, and now my notifications seem to have stopped updating entirely...), I'll update once I figure out what's actually going on

Edit: Ok, the weirdness was just because I confused my browser's cache with the time changes, if I clear my cache after the clock change, everything behaves roughly as described above. (tested in both Firefox and Chromium now)

estrattonbailey:
Should fix #1064

### DIFF
--- a/src/state/models/feeds/notifications.ts
+++ b/src/state/models/feeds/notifications.ts
@@ -529,16 +529,12 @@ export class NotificationsFeedModel {
     const latest = res.data.notifications[0]
 
     if (latest) {
-      // not in action
-      this.mostRecentNotificationUri = latest.uri
-
       const now = new Date()
       const lastIndexed = new Date(latest.indexedAt)
       const nowOrLastIndexed = now > lastIndexed ? now : lastIndexed
 
-      runInAction(() => {
-        this.lastSync = nowOrLastIndexed
-      })
+      this.mostRecentNotificationUri = latest.uri
+      this.lastSync = nowOrLastIndexed
     }
 
     return this._appendAll(res, true)
--- a/src/state/models/feeds/notifications.ts
+++ b/src/state/models/feeds/notifications.ts
@@ -241,6 +241,12 @@ export class NotificationsFeedModel {
   loadMoreError = ''
   hasMore = true
   loadMoreCursor?: string
+
+  /**
+   * The last time notifications were seen. Refers to either the
+   * user's machine clock or the value of the `indexedAt` property on their
+   * latest notification, whichever was greater at the time of viewing.
+   */
   lastSync?: Date
 
   // used to linearize async modifications to state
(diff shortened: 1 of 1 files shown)


src/state/models/feeds/notifications.ts
```diff
@@ -523,9 +526,21 @@ export class NotificationsFeedModel {
   // =
 
   async _replaceAll(res: ListNotifications.Response) {
-    if (res.data.notifications[0]) {
-      this.mostRecentNotificationUri = res.data.notifications[0].uri
+    const latest = res.data.notifications[0]
+
+    if (latest) {
+      // not in action
```
pfrazee, line 532: What does "not in action" refer to?

src/state/models/feeds/notifications.ts
```diff
@@ -523,9 +526,21 @@ export class NotificationsFeedModel {
...
+      this.mostRecentNotificationUri = latest.uri
+
+      const now = new Date()
+      const lastIndexed = new Date(latest.indexedAt)
+      const nowOrLastIndexed = now > lastIndexed ? now : lastIndexed
+
+      runInAction(() => {
```
pfrazee, line 539: This `runInAction()` closure shouldnt be necessary (there hasn't been an await prior, at least that I can see in the diff)

pfrazee (changes requested):
Seems right, some small questions/changes

src/state/models/feeds/notifications.ts
estrattonbailey: Ok I follow, fixing rn

pfrazee (approved):
LGTM

PR #1205, opened 2023-08-17
Fix to the web composer's state management (close #1193)

### LINKED ISSUE #1193: cmd+enter post submission on desktop app removes link cards
When you submit a post using `cmd+enter` (this is on mac, not sure what the shortcut is for non-mac) on https://bsky.app the link card gets removed

pfrazee:
The tiptap `useEditor()` hook creates an awkward challenge for passing event handlers into its plugins and native events. By introducing a memoized editor, we should be able to shuttle events out of tiptap without retriggering the useEditor hook. The emitter can then change its registered handlers with each state update.

This closes #1193 which was caused by stale state getting captured in the `onPressPublish` closure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "base64-js": "^1.5.1",
     "bcp-47-match": "^2.0.3",
     "email-validator": "^2.0.4",
+    "eventemitter3": "^5.0.1",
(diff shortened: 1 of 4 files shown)


estrattonbailey (approved):
Nice

PR #1207, opened 2023-08-17
default translate to device language

### LINKED ISSUE #1126: [Translation] Content Language setting overrides target translation language
**Describe the bug**

The language that was enabled _first_ under Content Languages becomes the pre-selected target translation language for the Translate feature of Bluesky regardless of the OS default language. 

**To Reproduce**

1. Go to Settings -> Content Languages.
2. Make sure no languages are enabled.
3. Enable any language other than the OS default language first, followed by any others.
4. Save the changes.

As a result, the Translate feature will use the language that was enabled first as the pre-selected target language and _not_ the OS default language.

**Expected behavior**

The pre-selected target translation language must be the default OS language regardless of which language was enabled first under Content Languages.
Alternatively, you can turn this into a feature by adding a note stating that the first language selected here will be used as a target language for the translation. :)

**Details**

 - Platform: Android
 - Platform version: 10
 - App version: 1.43.0

estrattonbailey:
Fixes #1126

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -79,8 +79,6 @@ export function isPostInLanguage(
   return bcp47Match.basicFilter(lang, targetLangs).length > 0
 }
 
-export function getTranslatorLink(lang: string, text: string): string {
-  return encodeURI(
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Great, thanks!

PR #1213, opened 2023-08-18
Add x.com to `isTwitterUrl` heuristic

mozzius:
It's only a matter of time before it switches fully


pfrazee:
includes `x.com` is a bit too general (though includes `twitter.` isnt great either). Let's just deal with this when it happens

PR #1227, opened 2023-08-21
Add likes tab

estrattonbailey:
Depends on https://github.com/bluesky-social/atproto/pull/1440

**Note: you can only view your own likes. The likes tab will not appear on other users' profiles.**


pfrazee (approved):
👍 Looks good. I'm inclined not to add any messaging for now. We might add some in the future but this seems fine

PR #1228, opened 2023-08-21
add title attr to text text links

### LINKED ISSUE #1226: Add a tooltip on the post date label link which shows an exact timestamp

I often scroll down some timelines and see posts with relative dates like "25d", "1mo", but I would like to see what day exactly it was posted on, since it's not convenient to have to calculate in your mind what day was 25 days ago. You can see the timestamp on the post detail page, but then you need to keep opening posts and going back to the list.

It's a common UI pattern on this kind of sites to have a tooltip that appears when hovering the date label for a moment which shows the exact date and time. This could be either a standard HTML tooltip (`title`) or something custom.

**Examples**

Mastodon (title attribute):

<p></p>

Twitter (custom tooltip):

<p></p>

estrattonbailey:
Fixes #1226 

Sets the `title` HTML attribute so that the date shows on hover.

Screen Shot 2023-08-21 at 10 41 05 AM

### DIFF
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -135,6 +135,7 @@ export const TextLink = observer(function TextLink({
   numberOfLines,
   lineHeight,
   dataSet,
+  title,
 }: {
   testID?: string
   type?: TypographyVariant
@@ -144,10 +145,12 @@ export const TextLink = observer(function TextLink({
   numberOfLines?: number
   lineHeight?: number
   dataSet?: any
+  title?: string
 } & TextProps) {
   const {...props} = useLinkProps({to: sanitizeUrl(href)})
   const store = useStores()
   const navigation = useNavigation<NavigationProp>()
+  const anchorRef = React.useRef(null)
 
   props.onPress = React.useCallback(
     (e?: Event) => {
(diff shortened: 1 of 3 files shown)


src/view/com/util/Link.tsx
```diff
@@ -166,8 +169,14 @@ export const TextLink = observer(function TextLink({
     return {}
   }, [href])
 
+  React.useEffect(() => {
+    // @ts-expect-error web only, shows date on hover -esb
+    anchorRef.current?.setNativeProps?.({title})
```
pfrazee, line 174: Oof, is this really the only way to do this?

estrattonbailey: Yeah the `title` attr is ignored afaict :((((((

pfrazee: I almost wonder if we'd be better off putting it in a `data-` attribute and adding a css-only tooltip library

estrattonbailey: Also just remembered I'd heard that the new arch will deprecate `setNativeProps`. This is still a first party API, so could work for now, but yeah I can do a simple CSS tooltip too nbd.

pfrazee:
Sweeeet. Don't forget to add the css to bskyweb

pfrazee:
Yeah this is looking great. Three things

1 - Get the css in bskyweb

2 - We have some floating elements that occlude and I dont know if we'll be able to z-index our way out of it. Should we consider hacking out of it by having them go below instead of above?

CleanShot 2023-08-21 at 16 51 46@2x

3 - Should probably add to the expanded state of posts too

pfrazee (approved):
👍 Looks good!

PR #1231, opened 2023-08-21
allow handles to start with a number

### LINKED ISSUE #1229: User handles cannot start with a number
**Describe the bug**
Handle input doesn't allow handles to start with a number, but leading-digit domains are still valid.

**Expected behavior**
We should allow handles to start with a number.

**Additional context**
Some more info here: https://serverfault.com/questions/638260/is-it-valid-for-a-hostname-to-start-with-a-digit

estrattonbailey:
Fixes #1229

### DIFF
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -3,7 +3,7 @@ export function makeValidHandle(str: string): string {
     str = str.slice(0, 20)
   }
   str = str.toLowerCase()
-  return str.replace(/^[^a-z]+/g, '').replace(/[^a-z0-9-]/g, '')
+  return str.replace(/[^a-z0-9-]/g, '')
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/strings/handles.ts
```diff
@@ -3,7 +3,7 @@ export function makeValidHandle(str: string): string {
     str = str.slice(0, 20)
   }
   str = str.toLowerCase()
-  return str.replace(/^[^a-z]+/g, '').replace(/[^a-z0-9-]/g, '')
+  return str.replace(/[^a-z0-9-]/g, '')
```
devinivy, line 6: One tiny thing that might slip through here is handles beginning with a hyphen.

estrattonbailey: Ah geez good call

pfrazee (approved):
LGTM

PR #1237, opened 2023-08-21
Add profiling sampling to sentry

ansh:


pfrazee (approved):
Let's give it a shot, shouldn't affect perf but we should keep an eye on it

PR #1240, opened 2023-08-22
Fix: include link-meta abort signal for timeout (close #1214)

### LINKED ISSUE #1214: `AbortController` for the fetch to the link metadata service is not wired up to the `fetch` call
https://github.com/bluesky-social/social-app/blob/257c15ec1f1b327e21f6e1f5707e6c0724d3afc6/src/lib/link-meta/link-meta.ts#L57

It should probably be

```ts
const controller = new AbortController()
const to = setTimeout(() => controller.abort(), timeout || 5e3)

const response = await fetch(
  `${LINK_META_PROXY(
    store.session.currentSession?.service || '',
  )}${encodeURIComponent(url)}`,
  { signal: controller.signal },
)
```

pfrazee:
Fix to a small bug introduced a bit ago that drops the timeout for link meta fetches (thanks @mozzius!)

Closes #1214

### DIFF
--- a/src/lib/link-meta/link-meta.ts
+++ b/src/lib/link-meta/link-meta.ts
@@ -58,6 +58,7 @@ export async function getLinkMeta(
       `${LINK_META_PROXY(
         store.session.currentSession?.service || '',
       )}${encodeURIComponent(url)}`,
+      {signal: controller.signal},
     )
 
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Nice!

PR #1242, opened 2023-08-22
[APP-835] trim end of email during login

### LINKED ISSUE #1202: Trailing whitespace in email field will cause sign in to fail
**Describe the bug**

**To Reproduce**

1. Go to https://bsky.app/
2. Sign in using email and password
3. See you've signed in successfully.
4. Sign out
5. Got to sign in again, but this time put a single space after your email address.

**Expected behavior**

You sign in successfully, since your details are correct and most platforms use a "trim()" to remove trailing whitespace from any username/email login details

**Actual behavior**

The login fails, unless you remove the space

**Additional context**

On Android phone I used autocomplete for the email address, but often adds a space after it by default. I couldn't figure out why I was signing in no problem on PC, but had problems on mobile, even though I used the same email and password.

ansh:
Fixes #1202

### DIFF
--- a/src/view/com/auth/login/Login.tsx
+++ b/src/view/com/auth/login/Login.tsx
@@ -403,7 +403,7 @@ const LoginForm = ({
             keyboardAppearance={theme.colorScheme}
             value={identifier}
             onChangeText={str =>
-              setIdentifier((str || '').toLowerCase().trimEnd())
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
:+1: Out an abundance of caution I switched it from trimEnd to trim, mainly because I'm concerned trimEnd is a recent enough standard library addition to not be present in some engine

PR #1246, opened 2023-08-22
Don't submit search if query is empty

darnfish:
Right now if you submit the search field on the Discover page with no content it submits an empty result, which shouldn't happen and rightly returns zero results. This PR fixes that.

| Posts | Users |
| --- | --- |
| Screenshot 2023-08-22 at 4 28 36 AM jpeg | Screenshot 2023-08-22 at 4 28 38 AM jpeg |


pfrazee:
Thanks!

PR #1253, opened 2023-08-22
do not thread author media filtered feed

### LINKED ISSUE #1212: Improve how the media tab handles threads
Currently the new media tab does not show all media when scrolling through it. Media within threads requires the someone browsing the tab click on "view full thread" to see that media, which could make sense in the main profile feed, but not in the media tab, where users are typically looking for a more exhaustive list of all media from the account.

Displaying media in thread form like this also prevents the media tab from being a fully chronological view of the media posted from an account, as you can scroll down seeing posts one by one in reverse chronological order except for posts in threads which are ordered by the chronology of the first media post in the thread, rather than when the reply iself was posted. If someone was to add an image as a reply to a month old thread that image would be a month back in their media tab despite having just been posted.

estrattonbailey:
Fixes #1212 

Adds an options object to the `PostsFeedModel` so that we can specify feeds that should display as a flat array of posts and not be threaded in situ into replies.

This will be useful for the upcoming Likes tab as well.

### DIFF
--- a/src/state/models/feeds/posts.ts
+++ b/src/state/models/feeds/posts.ts
@@ -11,8 +11,17 @@ import {cleanError} from 'lib/strings/errors'
 import {FeedTuner} from 'lib/api/feed-manip'
 import {PostsFeedSliceModel} from './posts-slice'
 import {track} from 'lib/analytics/analytics'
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
🚀

PR #1256, opened 2023-08-23
update golang to v1.21; set GOEXPERIMENT=loopvar

bnewbold:
We're hitting this in all the other golang repos at the same time.


pfrazee:
@Jacob2161 good to merge?

Jacob2161 (approved):
:bow:

PR #1270, opened 2023-08-24
Add analytics event for when the app is opened from a notification

pfrazee:


estrattonbailey (approved):

src/lib/analytics/types.ts
```diff
@@ -8,6 +8,8 @@ export type ScreenEvent = (
   properties?: ScreenPropertiesMap[keyof ScreenPropertiesMap],
 ) => Promise<void>
 interface TrackPropertiesMap {
+  // OS events
+  'OpenApp:FromNotification': {}
```
ansh, line 12: I would probably call this

`Notifications:Click` or `Notificatons:OpenApp`

My reasoning is that in the future we will probably deep-link notifications to the exact reply, quote, etc. and then we would want to track that differently as well.

ansh (approved):
👍🏽 Left a small comment

src/lib/analytics/types.ts
pfrazee: Sure sg

Issue #1233, opened 2023-08-21
[APP-829] Too many pinned feeds creates performance issues


pfrazee:
See why this might be -- is it too many views in the pager, are we loading things optimistically when we shouldn't? Etc. Then see if there are some reasonable fixes available to us without changing the existing behavior. This might require unloading views from the pager if too many are there. (Realistically might not be possible to have more than 3.) 

<sub>[APP-829](https://linear.app/blueskyweb/issue/APP-829/too-many-pinned-feeds-creates-performance-issues)</sub>

estrattonbailey:
Performance related to feeds should be improved after #1271, gonna close this

PR #1271, opened 2023-08-24
defer loading of feeds until visible

estrattonbailey:
Should help #1233

Each of the tab screens renders when you open the app, so this relies on `isPageFocused` to trigger a first load of the feed itself. And since loading happens lazily, we need a loading state ready for the user on each swipe, hence the change to loading logic to show a default loading state only if `isInitialLoading = true`.

`isInitialLoading` is named after the [same feature](https://tanstack.com/query/v4/docs/react/guides/disabling-queries#isinitialloading) in React Query.

### DIFF
--- a/src/state/models/feeds/posts.ts
+++ b/src/state/models/feeds/posts.ts
@@ -272,7 +272,7 @@ export class PostsFeedModel {
    * Check if new posts are available
    */
   async checkForLatest() {
-    if (this.hasNewLatest || this.isLoading) {
+    if (!this.hasLoaded || this.hasNewLatest || this.isLoading) {
       return
     }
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
🚀 Excellent!

estrattonbailey:
@onmyouji no problem. The issue here was that we were loading posts for every pinned feed on startup. That part has been fixed. We do still need to load a feed's info (like display name) for every pinned feed, but we only load the posts of the feed that you're currently viewing.

We definitely want to help if we can. What kind of device are you using? Is feed loading from your home screen slow too, or is it only the settings page where you reorder your feeds? Thanks!

PR #1279, opened 2023-08-25
don't mention own user from composer

### LINKED ISSUE #1277: Entering the composer from your own profile prefills the editor with your handle
**Describe the bug**
Nav to your profile, then click on the compose post button. Your handle is already present in the editor.

**Expected behavior**
This is nice for other accounts, but on your own account it doesn't make sense.

**Details**

 - Platform: iOS
 - Platform version: 17
 - App version: TestFlight

estrattonbailey:
Closes #1277 

Saw we had this on web, figured we wanted on mobile too.


pfrazee (approved):
👍

PR #1284, opened 2023-08-25
[APP-812] add custom feed discovery to onboarding

ansh:
- Makes onboarding into a screen not modal
- Allow users to add recommended custom feeds during onboarding
- Saves position in onboarding locally
RocketSim_Recording_iPhone_14_Pro_2023-08-28_18 43 42


src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+  RecommendedFeeds: 'RecommendedFeeds',
+  Home: 'Home',
+} as const
+
+type OnboardingStep =
+  (typeof OnboardingScreenSteps)[keyof typeof OnboardingScreenSteps]
+const OnboardingStepsArray = Object.values(OnboardingScreenSteps)
+export class OnboardingModel {
+  // state
+  step: OnboardingStep = 'Welcome'
```
pfrazee, line 17: The default should be the off state so that existing users don't get put through this flow. Then we can add a `start()` function or similar during account creation that sets the step to 'Welcome' and goes through this flow

src/Navigation.tsx
```diff
@@ -219,6 +221,26 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
...
+        name="Welcome"
+        component={Welcome}
+        options={{
+          title: title('Welcome'),
+          presentation: 'card',
+          headerShown: true,
+          headerTransparent: true,
+          headerTitle: '',
+          headerBackVisible: false,
+          headerRight: props => <WelcomeHeaderRight {...props} />,
```
pfrazee, line 234: These titles & headings don't have any effect on web, so you're likely going to need to put them in the UIs themselves

src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+  constructor(public rootStore: RootStoreModel) {
+    makeAutoObservable(this, {
+      rootStore: false,
+      hydrate: false,
+      serialize: false,
+    })
+  }
+
+  serialize(): unknown {
+    console.log('serializing onboarding', this.step)
```
pfrazee, line 28: Log

src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  hydrate(v: unknown) {
+    if (typeof v === 'object' && v !== null) {
+      if (
+        hasProp(v, 'step') &&
+        typeof v.step === 'string' &&
+        OnboardingStepsArray.includes(v.step as OnboardingStep)
+      ) {
+        console.log('hydrating onboarding', v.step)
```
pfrazee, line 41: Log

src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.step = v.step as OnboardingStep
+      }
+    } else {
+      // if there is no valid state, we'll just reset
+      this.reset()
+    }
+  }
+
+  nextScreenName(currentScreenName?: OnboardingStep) {
```
pfrazee, line 50: The semantic here is a little unclear -- `nextScreenName()` seems like a getter but it's actually effectful.

I'm +1 to having this function return the next screen name but I'd probably just call it `next()` or similar

src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+    if (currentScreenName === 'Welcome' || this.step === 'Welcome') {
```
pfrazee, line 51: You can simplify these by just doing

```typescript
currentScreenName = currentScreenName || this.step
```

at the start of the function

src/state/models/discovery/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+  skip() {
+    track('Onboarding:Skipped')
+    this.step = 'Home'
+  }
+
+  get isComplete() {
+    return this.step === 'Home'
+  }
+
+  get isRemaining() {
```
pfrazee, line 81: `isRemaining` is a little confusing, maybe `isActive` instead

src/Navigation.tsx
```diff
@@ -67,6 +67,8 @@ import {getRoutingInstrumentation} from 'lib/sentry'
 import {bskyTitle} from 'lib/strings/headings'
 import {JSX} from 'react/jsx-runtime'
 import {timeout} from 'lib/async/timeout'
+import {Welcome, WelcomeHeaderRight} from 'view/com/auth/onboarding/Welcome'
+import {RecommendedFeeds} from 'view/com/auth/onboarding/RecommendedFeeds'
```
pfrazee, line 71: These should probably be in the screens folder and treated like normal screens if we're going to use the navigation model.

If we don't want to use navigation, the other option is to use the `withAuthRequired` higher-order component

pfrazee: In fact, if we're using the navigation approach we need to address what happens if the user navigates away via the sidenav (currently it looks like you do so successfully but the next refresh will resume onboarding where you left off)

pfrazee (changes requested):
Off to a good start, got some comments.

Web needs to use the `util/Views` helpers to properly center the elements.

src/state/models/discovery/onboarding.ts
ansh: Agreed

src/state/models/discovery/onboarding.ts
ansh: Good catch, thanks!

src/state/models/discovery/onboarding.ts
ansh: Done

src/Navigation.tsx
ansh: Fixed.

estrattonbailey (approved):
LGTM, only thing I can find is that the bottom bar covers the continue button on mobile

Screen Shot 2023-08-30 at 6 03 30 PM
Screen Shot 2023-08-30 at 6 07 57 PM

pfrazee:
Yep I just caught that, I introduced that regression. Going to fix that and merge

PR #1299, opened 2023-08-28
Only show clear button while focused

darnfish:
As of today, the clear button on the search field shows as long as there is a query inside the search bar. This is not typical behaviour, as the clear button is usually only shown while the user is editing the search field (focused).

This PR moves the shown/hidden state of the clear button into expected functionality.


pfrazee:
Is that the expected behavior? Isn't it useful to have the clear button even if it's not focused?

darnfish:
Yeah I think you're right, let me refactor this PR a little to show the input field when the clear button is tapped which is more in line with the standards

Issue #1302, opened 2023-08-28
Infinite fetch loop on notifications page


mozzius:
**Describe the bug**

https://github.com/bluesky-social/social-app/assets/10959775/c31ab7be-1383-4fe6-a449-b58b2e5f5c39

**To Reproduce**

Not sure, sorry :(
I think I when from notifications > home > notifications to get new notifications (or perhaps just clicked notifications from the notifications screen)

**Expected behavior**

Don't do that

**Details**

 - Platform: Web
 - Platform version: Firefox
 - App version:

pfrazee:
Yeah I ran into this over the weekend myself, managed to get myself logged out due to rate limiting. Going to diagnose today and deploy a hotfix

PR #1307, opened 2023-08-28
revert potentially problematic logic change

estrattonbailey:
I think this is probably causing #1302, though I've only reproduced once and I'm not sure how to reliably repro.


estrattonbailey:
Can reopen this if we need to hotfix, but would be good to fix-foward if possible

pfrazee:
Yeah thanks for opening, agree this is the likely problem spot. I think we can solve the issue; my guess is it's when the request fails

PR #1318, opened 2023-08-29
Improvements to feed assembly to avoid possible state issues

pfrazee:
The first commit is not strictly necessary but wise to do -- the feed tuner is stateful and the `checkForLatest()` call was modifying that state. That state resets whenever "load new posts" is pressed so it didn't manifest as a bug, but it's still smart to avoid issues.

The second commit is a somewhat gross sanity check to ensure that we never let a duplicate react-key through. Because a duplicate react key is _disastrous_ to rendering, I think this is worth doing.


estrattonbailey (approved):
Makes sense to me

PR #1319, opened 2023-08-29
Replace web editor link behavior

### LINKED ISSUE #1292: Clicking in the New Post text field such that the cursor is placed on a link opens the link
**Describe the bug**
When composing a post, left-clicking anywhere such that the cursor ends up on top of a link will open the link.
After clicking, the cursor is placed behind the non-whitespace last character of the post each time.

This does not occur with RMB (opens context menu) or MMB (normal behavior). Click-dragging to select text works as expected too, including dragging and returning to make a zero-width selection (effectively placing the cursor with LMB) - this only happens when strictly clicked with LMB.

Mobile versions don't allow clicking in text fields to open the link, so this does not apply there.

**To Reproduce**

1. Begin composing a New Post and write a link anywhere.
2. Click outside of the text such that the cursor ends up on top of the link (including after the last link character, but not before the first link character - anywhere where the cursor turns blue instead of white/black).
3. The link will be opened, and the text cursor will be placed behind the last non-whitespace character in the text field (that is, excluding trailing spaces and newlines).

Examples include placing a link at the end and clicking well-outside the text, e.g. in the bottom-right corner, to move the cursor to the end; or making a one-line post with a link in the middle, and then clicking somewhere perpendicular to the text to place the cursor onto the link.

**Details**

 - Platform: Desktop, Chrome 115.0.5790.171
 - App version: 1.47.0

### LINKED ISSUE #1293: Link underlining breaks if the link is edited within the middle of the text (or in some other cases)
**Describe the bug**
This seems like a really far-reaching bug with a lot of extra cases, but I'll try my best. Each time I returned to test further, I found more weird behavior.

In the post compositor, returning the cursor to a link not at the end of the text, and either editing it or appending text, causes all the extra text to get underlined, including whitespace and URL-incompatible characters. Newline at the end of the link is an exception, but not when inserted into the middle. Pasting from Bluesky, other browser tabs, or external sources all produces different results (elaborated below)

I couldn't repro this on mobile - the text field seems to re-parse pasted links correctly and extra text added onto a link doesn't get underlined.

This appears to be an entirely client-side visual discrepancy, and only before a post is sent.

**To Reproduce**

1. Write a new post with a link in the middle of the text
2. Return your cursor to the link, into the middle or to the end of it, and start typing further
3. All newly-typed characters - except for a newline if at the end - will be underlined too
4. Pasting non-link text from the compositor keeps it as a non-link. Pasting another link from the text field, even a subsection, keeps that subsection linked to the url that it came from. Pasting text copied from some other sources (e.g. notepad, but not the browser) behaves one way or another, some formatting correctly, some extending the link incorrectly.
5. Deleting the link never invalidates it in this case either, not until it's deleted completely. This is also evident by the interface suggesting a card for the link until all the blue underlined text from it is erased.

Additionally, this also occurs when, after writing a link at the end of a post:

- Hitting enter, then backspace, then typing more
- Deleting at least one character from the link (at any point, even after erasing previous text ahead of the link), then typing more
  - The card suggestion only gets invalidated after
- Pasting text copied from certain sources
  - Copying from notepad extends the link (even if it starts with a newline)
  - Copying from elsewhere in the browser, even another tab, keeps links or plain text, even if the links were hrefs that then don't get parsed after posting (that is, blue text that doesn't match its target url).

**Expected behavior**

- When pasting any text, link/underline behavior should be consistent with how the server parses it after posting. Currently, links are only valid when they write out the whole link, but the client/app shows the underline even when the server won't put it there afterwards.
- When editing a link in the middle of text (or, as mentioned, at the end after making a newline and, it should be re-parsed correctly on the frontend. Non-URL characters should invalidate links if inserted, or end links if appended. Likewise with deleting parts of links currently, as they aren't invalidated until removed entirely.
  - Since rich text is probably gonna eventually support completely custom linking, this behavior should still be consistent with user expectations. If I type example.com into my post, and then erase the last part to make it say example.org, my link should lead to the latter, but right now the client suggests that it leads to the former. The server re-parses it as expected currently, but this expected behavior should be ensured for directly-written links if/once full rich-text linking is supported. Modifying text with a link shouldn't change the link, but modifying an auto-generated link should auto-regenerate the link too. (if that makes sense)
- Overall, the text compositor should be in sync with how the server parses the post after it's sent

**Screenshots**

chrome_2023-08-27_15-54-15
chrome_2023-08-27_15-54-21
chrome_2023-08-27_15-54-41
chrome_2023-08-27_15-59-56

**Details**

 - Platform: Desktop, Chrome 115.0.5790.171 (Official Build) (64-bit) (cohort: Stable)
 - App version: 1.47.0

**Additional context**

I discovered this while reporting #1292, and testing it yielded a lot of weird edge cases.
Feel free to refer to a thread where I tried out various things: https://bsky.app/profile/metaflame.bsky.social/post/3k5wu6gboov2o
The server side of things seems to handle everything properly, though I raised concerns there about a potential typo-squatting exploit with links and non-matching cards; that's for another Issue though.

pfrazee:
Closes #1293
Closes #1292

This PR replaces tiptap's link plugin with a simple "decorator" plugin. The multiline comment at the top explains the situation pretty well:

> TipTap is a stateful rich-text editor, which is extremely useful when you _want_ it to be stateful formatting such as bold and italics. However we also use "stateless" behaviors, specifically for URLs where the text itself drives the formatting.

> This plugin uses a regex to detect URIs and then applies link decorations (a <span> with the "autolink") class. That avoids
adding any stateful formatting to TipTap's document model. We then run the URI detection again when constructing the RichText object from TipTap's output and merge their features into the facet-set.

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -113,9 +113,9 @@
     .ProseMirror .mention {
       color: #0085ff;
     }
-    .ProseMirror a {
+    .ProseMirror a,
+    .ProseMirror .autolink {
       color: #0085ff;
-      cursor: pointer;
     }
     /* OLLIE: TODO -- this is not accessible */
     /* Remove focus state on inputs */
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "@tiptap/extension-document": "^2.0.0-beta.220",
     "@tiptap/extension-hard-break": "^2.0.3",
     "@tiptap/extension-history": "^2.0.3",
-    "@tiptap/extension-link": "^2.0.0-beta.220",
     "@tiptap/extension-mention": "^2.0.0-beta.220",
     "@tiptap/extension-paragraph": "^2.0.0-beta.220",
     "@tiptap/extension-placeholder": "^2.0.0-beta.220",
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -1,12 +1,11 @@
 import React from 'react'
 import {StyleSheet, View} from 'react-native'
-import {RichText} from '@atproto/api'
+import {RichText, AppBskyRichtextFacet} from '@atproto/api'
 import EventEmitter from 'eventemitter3'
 import {useEditor, EditorContent, JSONContent} from '@tiptap/react'
 import {Document} from '@tiptap/extension-document'
 import History from '@tiptap/extension-history'
 import Hardbreak from '@tiptap/extension-hard-break'
-import {Link} from '@tiptap/extension-link'
 import {Mention} from '@tiptap/extension-mention'
 import {Paragraph} from '@tiptap/extension-paragraph'
 import {Placeholder} from '@tiptap/extension-placeholder'
@@ -17,6 +16,7 @@ import {createSuggestion} from './web/Autocomplete'
 import {useColorSchemeStyle} from 'lib/hooks/useColorSchemeStyle'
 import {isUriImage, blobToDataUri} from 'lib/media/util'
 import {Emoji} from './web/EmojiPicker.web'
+import {LinkDecorator} from './web/LinkDecorator'
 
 export interface TextInputRef {
   focus: () => void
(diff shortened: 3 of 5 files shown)


estrattonbailey (approved):
LGTM

PR #1320, opened 2023-08-29
Make load latest centered button primary

P4tr0ll:
"Load latest" centered button - the one displaying at the top in web whenever there are new posts/notifications available is barely visible.

Buttons that do primary actions like showing/producing new content that wasn't there before, encouraging engagement or in general making the user do something that's desired to be as often as possible should be primary - very visible.

(This is the case on _similar platforms_)

Before:
image
After:
image


pfrazee:
I agree that it's way too hard to see in dark mode but I'm not sure the primary action color is right. For an element of that size, it conflicts strongly in the IA with the post button, which _is_ the primary action. Instead we should be giving it a lighter background from the palette or similar.

pfrazee (changes requested):
-

pfrazee:
We've got a new approach to this coming in a separate PR

PR #1322, opened 2023-08-29
fix: dont display reply input before thread fetches

P4tr0ll:
"Write your reply" was showing up when thread was still loading or non existent
image
image


pfrazee (changes requested):
Appreciate the PR! Rather than adding a state item, I think we should just check the view for if it's loaded successfully before rendering

PR #1330, opened 2023-08-29
fix lightbox on search page

estrattonbailey:
On views that use the `Feed` component and therefore `FlatList`, we have `desktopFixedHeight` that applies `height: 100vh; overflow-y: auto`. On the search page, we only load 30 posts, and so don't use the `FlatList`. The reason it's 100vh and not 100% is because it's nested deep enough in the tree that our wrapping containers that ARE 100% aren't direct parents.

I think the preferable way to fix this is to use do it as high up the tree as possible so that we can use a relative height of 100%. This PR also only applies these styles to web, which is a little nicer than having to conditionally apply them when `isWeb` within the `SearchResults` components.

### DIFF
--- a/src/view/screens/Search.web.tsx
+++ b/src/view/screens/Search.web.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import {View} from 'react-native'
 import {SearchUIModel} from 'state/models/ui/search'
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
👍 Looks good!

PR #1198, opened 2023-08-16
Fix #1146 Android fatal crash

ansh:
Sentry Reports have suggested this is a bug in in `react-native-reanimated`

This PR upgrades `reanimated` to the latest version to hope the issue is fixed.

Hopefully fixes #1146

PR #1331, opened 2023-08-29
Upgrade to Expo 49

estrattonbailey:
Upgrades to Expo 49 using the migration guide in [their blog post](https://blog.expo.dev/expo-sdk-49-c6d398cdf740).

Resulted in a handful of dep updates, including React Native. For the latter I needed to re-patch the package, but afaict nothing had changed there so the patch is essentially the same.

One of the things we wanted from this was to upgrade `expo-camera` to fix console errors on web. It looks like they published over their [latest version accidentally](https://www.npmjs.com/package/expo-camera?activeTab=versions), so I pinned it to their actual latest for right now.

This PR builds on web/iOS/Android, but we should test more extensively before merging.

Another note: I feel better about this PR after running `npx expo install --fix` off `main` right now. Looks like we've upgraded a few things like `expo-image` (2 minor versions) and `reanimated` (major version) beyond what's "compatible" with our version of Expo. This PR contains only `reanimated`, and that's only a minor version now.


package.json
```diff
@@ -127,20 +127,20 @@
     "react-native-linear-gradient": "^2.6.2",
     "react-native-pager-view": "6.1.4",
     "react-native-progress": "bluesky-social/react-native-progress",
-    "react-native-reanimated": "^3.4.2",
+    "react-native-reanimated": "~3.3.0",
```
estrattonbailey, line 130: Slight downgrade, need to test this

package.json
```diff
@@ -168,7 +168,6 @@
     "@types/lodash.set": "^4.3.7",
     "@types/lodash.shuffle": "^4.2.7",
     "@types/react-avatar-editor": "^13.0.0",
-    "@types/react-native": "^0.67.3",
```
estrattonbailey, line 171: Apparently don't need this anymore

src/view/com/auth/SplashScreen.web.tsx
```diff
@@ -90,6 +90,7 @@ const styles = StyleSheet.create({
   containerInner: {
     height: '100%',
     justifyContent: 'center',
+    // @ts-ignore web only
```
estrattonbailey, line 93: Almost all of these are due to new types trying to protect us from ourselves: https://github.com/facebook/react-native/issues/39179

package.json
```diff
@@ -29,19 +29,19 @@
...
     "@gorhom/bottom-sheet": "^4.4.7",
     "@mattermost/react-native-paste-input": "^0.6.4",
     "@miblanchard/react-native-slider": "^2.3.1",
-    "@react-native-async-storage/async-storage": "^1.17.6",
+    "@react-native-async-storage/async-storage": "1.18.2",
     "@react-native-camera-roll/camera-roll": "^5.2.2",
     "@react-native-clipboard/clipboard": "^1.10.0",
     "@react-native-community/blur": "^4.3.0",
-    "@react-native-community/datetimepicker": "6.7.3",
+    "@react-native-community/datetimepicker": "7.2.0",
```
estrattonbailey, line 44: Appears to work just fine

package.json
estrattonbailey: Seems fine on all platforms

ansh: Why was this downgraded?

ansh: If its because of expo v49 compatibility, that only matters in the managed workflow afaik. We can use versions that aren't directly compatible with our version of expo because we use `prebuild` and build the app. 

Specifically this reanimated update was to fix a bug #1198 which was a fatal crash issue

estrattonbailey: Oh good call I forgot about that. Yeah we'll just ignore this one then 👍

estrattonbailey: Done

pfrazee (approved):
Great stuff, thanks Eric

PR #1336, opened 2023-08-30
Moderation settings fixes

### LINKED ISSUE #633: Content filtering sync between devices
**Describe the bug**

#### While the feature is enabled on the web platform, I initially assumed it would also be enabled within the application.
**Screenshots**

app-bug

**Details**

 - Platform: Mobile 
 - Platform version: iOS 16.3
 - App version: 1.28

### LINKED ISSUE #1139: race condition - get set preferences
**Describe the bug**
I can see that to save preferences a "read and save" happens, but it seems a race condition can be done opening just two tabs same page

**To Reproduce**

for example go to https://bsky.app/profile/bsky.app in two tabs
be already in the first feed but not in the second one
in first tab click on remove (trash icon) and very fast go to second tab
on second tab add second feed 
now fast go to first tab and click on confirm

you are not in any of both feeds now (I saw different results also)

there is an automatic refresh of the preferences, if that refresh happens it works, I assume that when the confirm dialog shows the auto refresh is not happening that fast. but even if it is happening you can be in 5 tabs also

**Expected behavior**

IMO read/write will be always a problem. one solution at least to mitigate it, broadcast changes on preferences to tabs. on one browser a host.domain can broadcast to same host.domain msgs. broadcast that a change happened. that way without realtime updates you know a change happened in another tab. anyway if you are in mobile and at the same time on web, read/write will be a problem...

**Additional context**

wasn't sure to write it here or in the protocol repo. not sure why an updatePreferences isn't there? it is because of how preferences are stored?

pfrazee:
- Moves preference mutations to the SDK
- FINALLY fixes the moderation sync issue with the adult content flag (close #633).
- Fixes a race condition with saved feed updates (close #1139). This was caused by failing to pull the latest saved feed state before doing add/remove mutations. The SDK now does that.
- Removes the "Adult content can only be toggled" warning on iOS if adult content has been turned on, to reduce confusion.
- Removes a bunch of dead code.


src/view/com/modals/ContentFilteringSettings.tsx
```diff
@@ -188,7 +192,7 @@ function SelectGroup({current, onChange, group}: SelectGroupProps) {
       />
       <SelectableBtn
         current={current}
-        value="show"
+        value="ignore"
         label="Show"
```
ansh, line 196: label is still `show` but value is `ignore`

src/view/com/modals/ContentFilteringSettings.tsx
```diff
@@ -48,15 +48,19 @@ export const Component = observer(({}: {}) => {
...
-              Adult content can only be enabled via the Web at{' '}
-              <TextLink
-                style={pal.link}
-                href="https://bsky.app"
-                text="bsky.app"
-              />
-              .
-            </Text>
+            store.preferences.adultContentEnabled ? (
+              <></>
```
ansh, line 52: nit: this can be `null` to stick with convention of the rest of the codebase

src/view/com/modals/ContentFilteringSettings.tsx
pfrazee: This is intentional. The internal logic is to ignore the label, but I figure the clearest move is to have the UI say "Show". Can revisit, but not just a missed change.

src/view/com/modals/ContentFilteringSettings.tsx
pfrazee: ah yep

src/view/com/modals/ContentFilteringSettings.tsx
ansh: Gothca!

estrattonbailey (approved):
Magic

ansh (approved):
❤️ seeing PRs that remove code

PR #1339, opened 2023-08-30
Fix 1304 refresh notifications when a notif is clicked or received

### LINKED ISSUE #1304: Notification via app or push notification should trigger refresh of notification tab
**Describe the bug**
When I get notified about new events, regardless of via app or via push notification, the event is not displayed, when I navigate to notifications. I always have to refresh the tab to get to read the event. However, in push notifications the content of the event is already displayed.

**To Reproduce**

1. Get notified.
2. Navigate to notifications.
3. Nothing new to see 
4. Refresh.
5. Notification event appears.

**Expected behavior**
When I get notified I expect to see the notified event, when I move there.

**Screenshots**

**Details**

 - Platform: Android
 - Platform version: 13
 - App version: 1.47.0

**Additional context**

ansh:
Fixes #1304 
- Refreshes `NotificationModel` when notif is clicked on
Demo video:
https://github.com/bluesky-social/social-app/assets/8207733/422159c7-11d7-4056-a524-185a5fae8654

~~Also does some extra stuff~~ (EDIT: Going to have to get rid of this because of how strict our rate limits are)
- ~~Removes code for 30s `NotificationModel.syncQueue()`~~
- ~~Runs `NotificationModel.syncQueue()` when notification is received while user is in foreground. This makes the `syncQueue` a lot smarter (only called when necessary) and also gives instant feedback for users when a notification is received while the app is on foreground~~


estrattonbailey (approved):

pfrazee:
Oh interesting. So you're essentially dropping the polling. This means the app cant have different notification-surfacing policies than push, right? (We don't push every notification)

PR #1341, opened 2023-08-30
Improvements to UI in web logged-out views

pfrazee:
Small UI update to logged out views so that we don't look quite so amateur

CleanShot 2023-08-30 at 16 56 45@2x
CleanShot 2023-08-30 at 16 56 54@2x

### DIFF
--- a/src/state/models/ui/create-account.ts
+++ b/src/state/models/ui/create-account.ts
@@ -120,7 +120,7 @@ export class CreateAccountModel {
       })
       track('Create Account')
     } catch (e: any) {
-      this.rootStore.onboarding.skip() // whoops jk
+      this.rootStore.onboarding.skip() // undo starting the onboard
       let errMsg = e.toString()
       if (e instanceof ComAtprotoServerCreateAccount.InvalidInviteCodeError) {
         errMsg =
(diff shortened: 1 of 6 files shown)


ansh:
No option to skip onboarding on desktop web

pfrazee:
yeah I dunno, right now it's 1 step so I'm kind of inclined to just get that in later

ansh:
It'd be v useful to get the metrics to see how many people still skip it @pfrazee

ansh (approved):
Works well on all platforms and light/dark mode too

src/state/models/ui/create-account.ts
```diff
@@ -125,7 +120,7 @@ export class CreateAccountModel {
       })
       track('Create Account')
     } catch (e: any) {
-      sessionReadySub.remove()
+      this.rootStore.onboarding.skip() // whoops jk
```
ansh, line 123: is this supposed to be here?

pfrazee: hah yes, I was being a bit goofy with my comment. I'll fix that

pfrazee:
> It'd be v useful to get the metrics to see how many people still skip it @pfrazee

okay let's plan to get that in with the additional screens you've got planned

Issue #1300, opened 2023-08-28
Web post composer can produce garbled text after inserting an emoji


intrnl:
**Describe the bug**

Inserting an emoji and them messing around with the post composer can cause the composer's text content to become garbled, emojis turning into `�` for example

**To Reproduce**

1. Open post composer
2. Insert an emoji from the emoji picker
3. Text input briefly focuses before blurring, so you try to focus again, but just before the emoji you just inserted
4. Press backspace after focusing.

**Expected behavior**

Nothing should be deleted as the cursor is on the first index.

**Screenshots**

[Kooha-2023-08-28-14-01-24.webm](https://github.com/bluesky-social/social-app/assets/20620901/96fc077a-4f28-49be-aa88-46b9bc823706)

**Details**

 - Platform: Desktop Firefox Linux 
 - Platform version: Firefox 118.0a1 (2023-08-27), Arch Linux 6.4.11-1-cachyos
 - App version: 1.47

**Additional context**

Not entirely sure what's currently happening, I'll dig around for causes

intrnl:
seems like this exclusively happens when messing around with the emoji specifically, you can see here how messing around with other parts of the text is unaffected, only when you happen to focus on *just before* the emoji and press backspace

[Kooha-2023-08-28-14-09-35.webm](https://github.com/bluesky-social/social-app/assets/20620901/8f57fc3f-6bdd-41bc-821d-229270e0cb6d)

simonetthomas:
It seems the emoji is made of two "parts", and when you add it and focus the composer again, the cursor appears in the middle of them. So when you press backspace, it deletes the left part and scrambles the emoji. When you add text immediately after focusing, the letters appear in the "middle" of the emoji (I don't know how emojis work technically)

https://github.com/bluesky-social/social-app/assets/36693311/11ffbcab-7f9d-405a-8339-d3ede489655f

pfrazee:
Now _that_ is an interesting behavior. Emojis are indeed made of multiple "parts", and it seems like tiptap (our web rich text editor) is in fact mispositioning the cursor into the middle of the emoji.

@gaearon I'm going to let you tackle this one; it looks like this might be a Firefox issue but we should check other browsers

mschwendener:
> looks like this might be a Firefox issue but we should check other browsers

Can reproduce in Firefox on Mac, not in Chrome or Safari. 

Position of the cursor is different after starting a post with an emoji via the emoji picker: 

Safari: 
[...]

PR #1254, opened 2023-08-22
[APP-836] Emoji picker for web

ansh:
Closes #1199 
- [x] Add emoji mart package for emoji picker functionality
- [x] Make it accessible with a screen reader
- [x] Allow emoji to be inserted
- [x] Close picker when emoji is selected
- [x] Cache emoji assets for 1 yr (h/t @Jacob2161)
- [x] Load emoji assets separate from main bundle

PR #1241, opened 2023-08-22
[APP-834] Allow @ing someone in post directly from profile

ansh:
Closes #1206 

- [x] web
- [x] ios
- [x] android

Web
https://github.com/bluesky-social/social-app/assets/8207733/e0cab755-d7f1-4e55-9c6c-00d9a3db5786

iOS
https://github.com/bluesky-social/social-app/assets/8207733/a98c42c8-1cc3-45da-8d73-66f0252abbe4

PR #1374, opened 2023-09-04
Fix caret jumps for web composer

gaearon:
- Fixes https://github.com/bluesky-social/social-app/issues/1300
- Fixes caret jumping to the end when inserting an emoji in the middle of the post using our picker 
- Fixes caret jumping to the end when blurring out of the composer and focusing it back again (maybe controversial)

---

See https://github.com/bluesky-social/social-app/issues/1300#issuecomment-1705753534 for an explanation. TLDR:
- We shouldn't be jumping the caret when the user picks an emoji. They might be in the middle of the post anyway.
- We should not be reading mutable `.length` in order to "jump" the caret to the end — it appears that our text editing library does not guarantee whether the new value is flushed yet by the time `onFocus` fires.
  - Although we *could* have fixed that by using `focus('end')` instead of `focus().setTextSelection(...)`, we probably shouldn't be jumping the caret to the end _at all_ on focus anyway — since it's not what you actually want e.g. when inserting an emoji in the middle of the text. So I removed that line altogether unless we have some motivation to keep it. <s>I haven't found a case where it's needed yet.</s> (Edit: it was needed for first focus, so I fixed it [like this](https://github.com/bluesky-social/social-app/pull/1374#issuecomment-1705788704) instead.)

This might need a review from someone familiar with #1241 and #1254.

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -111,7 +111,7 @@ export const TextInput = React.forwardRef(
           },
         },
         content: textToEditorJson(richtext.text.toString()),
-        autofocus: true,
+        autofocus: 'end',
         editable: true,
         injectCSS: true,
         onUpdate({editor: editorProp}) {
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -111,9 +111,6 @@ export const TextInput = React.forwardRef(
           },
         },
         content: textToEditorJson(richtext.text.toString()),
-        onFocus: ({editor: e}) => {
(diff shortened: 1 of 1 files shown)


gaearon:
This needs more work. @ansh pointed out that we need to jump the cursor to the end when the user opens the composer on somebody else's profile. This is because we want to place it after the mention.

gaearon:
This needs more work. @ansh pointed out that we need to jump the cursor to the end when the user opens the composer on somebody else's profile. This is because we want to place it after the mention.

gaearon:
Pushed another commit. This ensures that if we press Compose on someone's profile, the cursor will appear after the handle.

ansh (approved):
Looks great. Thanks for catching this 👍🏽

PR #1383, opened 2023-09-05
Enable Fast Refresh for web

gaearon:
React has a feature (Fast Refresh) that applies code edits in-place without reloading the entire app.

This feature is turned on for Android and iOS by default, but it wasn't enabled by Expo for web. In this PR, I followed [this guide](https://forums.expo.dev/t/enabling-fast-refresh-with-expo-web-in-2022/63102) to turn it on.

Here's a demo. The key thing is that editing code no longer reloads the entire page on the web:

https://github.com/bluesky-social/social-app/assets/810438/e323f4c2-06d1-47c6-b1f8-b71ea0bf23f3

In the future Expo will have this by default, but Evan says it would require moving to Expo Router V2 (and Metro). For now I think this will do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -153,6 +153,7 @@
     "@babel/preset-env": "^7.20.0",
     "@babel/runtime": "^7.20.0",
     "@did-plc/server": "^0.0.1",
+    "@pmmmwh/react-refresh-webpack-plugin": "^0.5.11",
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Nice

PR #1385, opened 2023-09-05
Fix thread positioning on load

pfrazee:
This affects the positioning of posts when viewing a thread

- Fixes the minimum index to match some changes to rendering
- Updates android to use the RN tool now that support has landed

In general this will greatly improve the thread scroll-positioning so that things dont jump around after load


estrattonbailey (approved):
It's lit

PR #1390, opened 2023-09-06
Update time.ts to handle very small or negative time differences

Bossett:
Right now, posts can appear to be from the future with a negative time difference (i.e. -3s appears). This change defines 'NOW' as less than 5 seconds old, and returns 'now' in that case.

It's not clear how localisation is handled - this may need translation.


pfrazee (approved):
Nice thanks!

pfrazee (changes requested):
Change looks good, do you think you could update the failing test?

pfrazee (approved):
Awesome, thank you so much! 🙏

PR #1393, opened 2023-09-06
Remove unnecessary deps from package.json

gaearon:
Running `yarn` on main produces a change in lockfile now, presumably after the Expo upgrade.
Rather than updating the lockfile I just removed unnecessary overrides.

Verified error box still works.


pfrazee (approved):
👍

PR #1394, opened 2023-09-06
Use frozen lockfile for CI workflows

gaearon:
Currently, running `yarn` on main produces a change in `yarn.lock`. This means that our CI doesn't check whether the lockfile is matching what's in `package.json`.

I think this flag is supposed to do that. I'd like to add it, <s>verify it'll fail the CI</s> (it doesn't, maybe due to https://github.com/yarnpkg/yarn/issues/5840, but it's still right to have it on), and then merge https://github.com/bluesky-social/social-app/pull/1393 with the fix.


pfrazee (approved):
sgtm

PR #1396, opened 2023-09-06
add development codesigning config plugin

EvanBacon:
We'll upstream this functionality in the future, but for reference, here's how you can configure your project to have development codesigning added during prebuild. Just swap out the `XX11XX1XXX` with the Apple team ID. The team ID can be located in the `ios/Bluesky.xcodeproj/project.pbxproj` file, under `DevelopmentTeam = "XX11XX1XXX";` where the `XX11XX1XXX` is your signing ID. Add this value to the `app.json` and it should always be used for development profiles.

The logic comes from [what we use](https://github.com/expo/expo/blob/main/packages/%40expo/cli/src/run/ios/codeSigning/xcodeCodeSigning.ts) in Expo CLI during `npx expo run:ios` to configure signing.

### DIFF
--- a/app.json
+++ b/app.json
@@ -69,6 +69,7 @@
       "url": "https://u.expo.dev/55bd077a-d905-4184-9c7f-94789ba0f302"
     },
     "plugins": [
+      ["./plugins/withIosCodeSigning", "XX11XX1XXX"],
       "expo-localization",
       "sentry-expo",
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks! Ended up resolving this locally based on your suggestions (not sure what helped tho).

PR #1398, opened 2023-09-06
Merge feed prototype [WIP]

pfrazee:
This PR introduces the "Merge feed" which is a home feed that samples from all of the user's saved feeds. The goal is to produce a feed which interleaves posts from the custom feeds into the following feed.

At time of writing, this aims for an interleave of:

```
********************************_____**********__**********_____*******__...
0 1 2 3 4 5 6 7 8 9 10 11 12 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28...
```

Where the `*` is follow feed and the `_` is custom feeds. Aka:

```
sampleFromCustomFeed = i >= 15 && (i % 4 === 0 || i % 5 === 0)
```

This happens entirely in the client so the interesting challenge is producing this outcome while papering over the multiple requests that are occurring. Here's the bag of tricks that are used to accomplish this:

- The following feed is treated as an anchor for the feed to ensure content is available (it should be more reliable than custom feeds)
- Every custom feed is treated like a queue that is topped up in increments of 10 and then sampled from. (The UI is pulling pages of 25 posts.) 
  - The custom feed requests are capped at 500ms; if they don't respond in time, they won't be used in that page but their response will be saved for future sampling.
  - The top-ups occur for only 3 feeds with each page-event to ensure we don't run a bunch of parallel requests.
  - Because the page size is 25, the pattern in use only requires 3 custom feeds in the initial page to provide the desired output pattern; if one of the custom feeds doesn't respond in the 500ms window, it's not a huge deal.

One other note: I removed the `update()` call which the feed would run to do an in-place update of like/repost stats because it was just a bit of a pain for our usecase. If we want to bring that back, we can probably do a `getPosts()` batch on the posts which are visible in the UI.

https://github.com/bluesky-social/social-app/assets/1270099/1bcecf76-c1eb-42ac-900d-d1eb6bdf19f4


pfrazee:
Still todo: make sure error handling works correctly. (It currently papers over network failures in some cases but the following feed should likely bubble up errors)

PR #1401, opened 2023-09-07
Fix scrollbar gutters on web

pfrazee:
This got broken with a "fix" I snuck into the responsive layout, but this finally fixes it for real. (See the comment for an explanation.)

Before:

CleanShot 2023-09-06 at 17 19 46@2x

After:

CleanShot 2023-09-06 at 17 19 31@2x


estrattonbailey (approved):
:shock_and_awe:

ansh (approved):
👍🏽

PR #1402, opened 2023-09-07
Fix lint-staged not allowing edits to Markdown files

gaearon:
I tried editing `docs/build.md` but committing it didn't work. Our lint-staged setup calls ESLint on a Markdown file, which breaks it. I've tweaked it so that we only run ESLint on the files that actually can be linted.

As a drive-by fix, noticed a seemingly unnecessary `overrides` option. I've folded it into the main config.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,6 +7,9 @@ module.exports = {
   ],
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint', 'detox'],
+  rules: {
+    'react-native/no-inline-styles': 0,
+  },
   ignorePatterns: [
     '**/__mocks__/*.ts',
     'src/platform/polyfills.ts',
@@ -20,14 +23,5 @@ module.exports = {
     'bskyweb',
     '*.html',
     'bskyweb',
-  ],
-  overrides: [
-    {
-      files: ['*.js', '*.mjs', '*.ts', '*.tsx'],
-      rules: {
-        semi: [2, 'never'],
-        'react-native/no-inline-styles': 0,
-      },
-    },
   ],
 }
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -21,13 +24,4 @@ module.exports = {
     '*.html',
     'bskyweb',
   ],
-  overrides: [
-    {
-      files: ['*.js', '*.mjs', '*.ts', '*.tsx'],
```
gaearon, line 26: Got rid of this so that it runs for whichever files are being linted

.eslintrc.js
```diff
@@ -21,13 +24,4 @@ module.exports = {
...
-      rules: {
-        semi: [2, 'never'],
-        'react-native/no-inline-styles': 0,
```
gaearon, line 29: Moved to main config's rules

.eslintrc.js
```diff
@@ -21,13 +24,4 @@ module.exports = {
     '*.html',
     'bskyweb',
   ],
-  overrides: [
-    {
-      files: ['*.js', '*.mjs', '*.ts', '*.tsx'],
-      rules: {
-        semi: [2, 'never'],
```
gaearon, line 28: This isn't necessary, Prettier is already taking care of this

package.json
```diff
@@ -252,6 +252,6 @@
     ]
   },
   "lint-staged": {
-    "*": "yarn eslint --ext .js,.jsx,.ts,.tsx --fix"
```
gaearon, line 255: No need for `ext` here, it would be ignored because `lint-staged` already passes a specific file. Instead, specify the glob to match staged files against.

estrattonbailey (approved):
Nice one

PR #1412, opened 2023-09-07
Add ESLint React plugin

gaearon:
This adds `eslint-plugin-react` with the recommended rules:

- https://github.com/bluesky-social/social-app/commit/a27c22428e4e3b715ba8441b533f4c3a14364b55 adds the plugin and disables the rules that fail.
- https://github.com/bluesky-social/social-app/commit/a278839c3f47c1d253fb3a93cd59a494261b58ee ([review without indentation changes](https://github.com/bluesky-social/social-app/pull/1412/commits/a278839c3f47c1d253fb3a93cd59a494261b58ee?w=1)) enables [`react/display-name`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/display-name.md) and fixes the violations. I chose `*Impl` suffix for inner functions so we don't accidentally change the behavior. E.g. if there's a recursive component, you pretty much always want the outer function.
  - The purpose here is to have readable names in React DevTools (and component stacks in general). Currently there's a whole lot of "Anonymous" components which makes getting acquainted with the codebase by poking the app a bit difficult. There's still a way bigger problem of "observerComponent" everywhere due to [anonymous MobX closures](https://github.com/mobxjs/mobx/blob/main/packages/eslint-plugin-mobx/README.md#mobxno-anonymous-observer-deprecated). I'll fix this in a follow-up PR cause it's a bit invasive.
- I didn't bother fixing up [`react/no-unescaped-entities`](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md) and left it disabled. We can reconsider if we ever run into the bugs it's supposed to catch.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,12 +2,15 @@ module.exports = {
   root: true,
   extends: [
     '@react-native-community',
+    'plugin:react/recommended',
     'plugin:react-native-a11y/ios',
     'prettier',
   ],
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'detox'],
+  plugins: ['@typescript-eslint', 'detox', 'react'],
   rules: {
+    'react/display-name': 0,
+    'react/no-unescaped-entities': 0,
     'react-native/no-inline-styles': 0,
   },
   ignorePatterns: [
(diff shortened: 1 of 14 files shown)


pfrazee (approved):
Yeah lgtm

PR #1416, opened 2023-09-08
Make Sentry optional

gaearon:
I see no need for requiring Sentry for local development. It's an impediment to the open source contributors (they have to set it up or rip it out), and uploading symbols is an extra step every time you run `yarn ios` / `yarn android`.

Until there's movement on https://github.com/getsentry/sentry-react-native/issues/1042, we can do this:

- https://github.com/bluesky-social/social-app/commit/748ff61b1b87f0d1b1bf942f3c257dccd625c233: Make the `app.json` config programmatic (`app.config.js`). No other changes there, I just ran it through our Prettier config.
- https://github.com/bluesky-social/social-app/commit/fde776d493a79a307b3168bbdfa5630994aba55b: Change the behavior so that if `SENTRY_AUTH_TOKEN` is missing or empty, the `expo-sentry` plugin is turned off. Update the contributing docs to match that.


pfrazee (approved):
Nice

PR #1417, opened 2023-09-08
fix theme flashing

estrattonbailey:
There are two open issues concerning this in the RN repo: https://github.com/facebook/react-native/issues/36713 and https://github.com/facebook/react-native/issues/35972

This PR applies the same approach [taken here](https://github.com/marklawlor/nativewind/issues/478), seems to work well. Fixes iOS/Android.

Noticed a related issue on web where `useColorScheme` doesn't report the correct system scheme, it always comes back dark.


src/lib/ThemeContext.tsx
```diff
@@ -91,13 +96,31 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({
...
+  React.useEffect(() => {
+    if (nativeColorSchemeChangeTimeout.current)
+      clearTimeout(nativeColorSchemeChangeTimeout.current)
+    nativeColorSchemeChangeTimeout.current = setTimeout(
+      () => setNativeColorScheme(colorSchemeFromRN),
+      1000,
+    )
+  }, [colorSchemeFromRN])
 
   const value = useMemo(
```
gaearon, line 114: Why is this `useMemo` at all? It doesn't do any expensive calculation. Let's remove it and make this a plain statement.

gaearon:
Is this explanation wrong? https://github.com/facebook/react-native/issues/35972#issuecomment-1705304553

gaearon:
How does it manifest in the UI? Can you record the video of it flashing?

src/lib/ThemeContext.tsx
estrattonbailey: Oh no reason, was just there before. I'll remove 👍

estrattonbailey:
>Is this explanation wrong?

Seems plausible to me

gaearon:
Do we know that this explanation is wrong (i.e. iOS isn't supposed to do this), or could it be that RN is not able to respond to iOS synchronously (due to async rendering) and so this manifests as a flash?

estrattonbailey:
I did some more searching and testing. It's not a new issue for RN (I tried back to 0.70), but I'm struggling to find other users with this issue. So I'm not sure which layer is wrong here, but agree it'd be great to learn one day.

Made us an internal ticket to track this, [APP-848](https://linear.app/blueskyweb/issue/APP-848/system-color-scheme-not-reported-correctly-on-web).

mozzius:
I've also had this issue if it helps. You may be interested in this person's solution that uses AppState to make sure the color scheme only changes if the app is in the foreground: https://github.com/marklawlor/nativewind/issues/478#issuecomment-1675212661

(Note: I have not tested this so I can't speak to it's effectiveness, I solved the issue differently)

ansh:
Thanks @mozzius 

Yeah the Nativewind approach makes a lot of sense. @estrattonbailey definitely check that out

estrattonbailey:
Hey @mozzius thanks for your thoughts! Saw this earlier and finally checked it out. It piqued my interest bc solution A (timeout) and solution B (AppState) both smell a little bit like a race condition.

For general posterity and other's having this issue, I'll post my findings:

It appears that `useColorScheme` reports changes in the value prior to the app being fully foregrounded. Even directly imported from RN, I get the inverse value of whatever scheme is currently active, followed immediately by the currently active scheme.

By creating and destroying this event listener based on changes to the value of `useColorScheme`, it's kinda acting as a debounce, and whatever value was last passed will be within the listener's closure and used once the app is fully foregrounded and it emits the `change` event.

I used this code to test this:

```ts
  const colorSchemeFromRN = useColorScheme()

  React.useEffect(() => {
    console.warn('Outside', {
      appStateIsActive: AppState.currentState === 'active',
      scheme: colorSchemeFromRN
    })

    const subscription = AppState.addEventListener('change', (state) => {
      const isActive = state === 'active';

      console.warn('Inside', {
        appStateIsActive: isActive, scheme: colorSchemeFromRN
      })

      if (!isActive) return;

      // set color
    });
    return () => subscription.remove();
  }, [colorSchemeFromRN]);
```

So let's say we're in `light` mode, and we change our system scheme to `dark`. Upon returning to the app, we'll see:
1. `Outside { appStateIsActive: false, scheme: 'light' }`
2. `Outside { appStateIsActive: false, scheme: 'dark' }`
3. `Inside { appStateIsActive: true, scheme: 'dark' }`

Handling this via event listener is much better than a timeout, since we can apply the correct theme as soon as we know the app is ready. Thanks for the helpful link!

pfrazee (changes requested):
👍 Yeah looks good

pfrazee:
Lol misclicked, meant to approve

pfrazee (approved):

gaearon:
I verified that applying https://github.com/facebook/react-native/pull/38214 + https://github.com/facebook/react-native/pull/39439 to RN locally fixes this (without the workaround in this PR).

gaearon:
I can't reproduce the web part of the issue btw. What's the exact sequencing to repro it?

I kind of have the opposite problem:

- On `main`, the theme gets initialized to the system setting correctly. Changing system setting midway does *not* change the theme until I refresh the page.
- If I revert this PR, then it works correctly.

PR #1427, opened 2023-09-11
Fix 1423 Prevent emoji picker clicks from bubbling up

### LINKED ISSUE #1423: [Web] Emoji picker closes instantly
**Describe the bug**
If I open the emoji picker and click in the search field, for example, it closes immediately

**To Reproduce**

1. open the emoji picker on web and click in the search field

**Expected behavior**

I expect to be able to enter a search term or access the emoji categories like I was able to do in 1.48 before the scroll bug was fixed

**Details**

 - Platform: Windows 11
 - Platform version: Firefox 117.0
 - App version: 1.49

ansh:
Similar to the composer or the modal, we want to prevent Emoji Picker clicks from bubbling up!

Fixes #1423

### DIFF
--- a/src/view/com/composer/text-input/web/EmojiPicker.web.tsx
+++ b/src/view/com/composer/text-input/web/EmojiPicker.web.tsx
@@ -58,21 +58,27 @@ export function EmojiPicker({close}: {close: () => void}) {
     // eslint-disable-next-line react-native-a11y/has-valid-accessibility-descriptors
     <TouchableWithoutFeedback onPress={close} accessibilityViewIsModal>
       <View style={styles.mask}>
-        <View
-          style={[
-            styles.picker,
-            {
-              paddingTop: noPadding ? 0 : reducedPadding ? 150 : 325,
-              display: noPicker ? 'none' : 'flex',
-            },
-          ]}>
-          <Picker
-            // @ts-ignore we set emojiMartData in `emoji-mart-data.js` file
-            data={window.emojiMartData}
-            onEmojiSelect={onInsert}
-            autoFocus={false}
-          />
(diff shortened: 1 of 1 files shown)


gaearon:
Is there some logic in some ancestor that closes it on click?

If yes, instead of preventing propagation here, have we considered doing a `node.contains` check on `e.target` there, and not closing if the target is within the picker?

src/view/com/composer/text-input/web/EmojiPicker.web.tsx
```diff
@@ -58,21 +58,27 @@ export function EmojiPicker({close}: {close: () => void}) {
     // eslint-disable-next-line react-native-a11y/has-valid-accessibility-descriptors
     <TouchableWithoutFeedback onPress={close} accessibilityViewIsModal>
```
ansh, line 59: @gaearon The ancestor is simply just this one.

I was following the pattern in the `Composer.web.tsx` file and the `Modal.web.tsx` file, but your idea sounds good. 

Let me take a look

pfrazee: Either way should work

ansh: Doesn't seem like the React Native `ref` on a `View` has a `contains` method

pfrazee: probably not worth sussing out the other approach, but I _believe_ there's a way to get the dom element out of the ref in the RNW build

pfrazee (approved):
👍 Seems to work!

PR #1429, opened 2023-09-11
Fix 651 Make followers and following button into a `<Link>`

### LINKED ISSUE #651: "followers" and "following" should be <a> elements

<a> have additional functionality such as link preview (bottom left of Chrome) on hover, open in new tab, probably some a11y benefits, etc

Change or wrap <div> in <a>

Chrome "duplicate tab"

**Additional context**

I can implement this if people think it's a good idea. Actually I reported a similar issue prior to open-source, and it seems to have been fixed. So we can probably `git blame` and follow the same pattern.

ansh:
Fixes #651

### DIFF
--- a/src/lib/analytics/types.ts
+++ b/src/lib/analytics/types.ts
@@ -56,8 +56,12 @@ interface TrackPropertiesMap {
   }
   // PROFILE HEADER events
   'ProfileHeader:EditProfileButtonClicked': {}
-  'ProfileHeader:FollowersButtonClicked': {}
-  'ProfileHeader:FollowsButtonClicked': {}
+  'ProfileHeader:FollowersButtonClicked': {
+    handle: string
+  }
+  'ProfileHeader:FollowsButtonClicked': {
+    handle: string
+  }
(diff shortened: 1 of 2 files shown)


pfrazee:
Does this actually turn those buttons into `<a>` tags? I dont think it will unless we specify an href attribute on them

ansh:
@pfrazee Even `asAnchor` and adding `href` doesn't seem to make them into `<a>` tags. I can look more into that but might be how some accessibilityType components get rendered.

Current solution makes the buttons accessible and gives current pointer type which is the main issue.

pfrazee:
Ah yeah that's possible. If we include an `href` that should enable cmd/ctrl+click though yeah? Also maybe we want to drop the button accessibilityType prop if we're able to make it an `<a>`?

ansh:
Makes sense @pfrazee I can do that

ansh:
@pfrazee Done

Also made some improvements are de-duping some code, and making tracking events better

estrattonbailey (approved):
V nice

pfrazee (approved):
🚀

PR #1435, opened 2023-09-12
Add birth date gating to moderation settings

pfrazee:
- Adds new birth date preference
- Adds modal to set birth date
- Updates settings screen to include age setting and moderation, rearranged a bit
- Updates moderation settings for adult content to require 18+

NOTE: does not disable adult content if not 18+ because I don't want to mess with existing users. New users will have to set age to turn on adult content, which defaults to off.

CleanShot 2023-09-11 at 17 08 27@2x

CleanShot 2023-09-11 at 17 08 47@2x

CleanShot 2023-09-11 at 17 09 46@2x

### DIFF
--- a/src/state/models/ui/create-account.ts
+++ b/src/state/models/ui/create-account.ts
@@ -118,6 +118,7 @@ export class CreateAccountModel {
         password: this.password,
         inviteCode: this.inviteCode,
       })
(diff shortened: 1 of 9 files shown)


pfrazee:
That last screenshot is an artifact of having adult content enabled but no age set. It shouldn't happen in practice.

estrattonbailey (approved):

PR #1437, opened 2023-09-12
Fix Profile link from drawer on Mobile Web

ansh:
Fixes following bugs on mobile web:
- Active state was not registering with the bottom tab when on Profile
- Profile link in drawer was not working


estrattonbailey (approved):

pfrazee:
I feel like this _shouldnt_ work just glancing at the code. The web uses a single flat navigation system, and this approach puts two routes on with the same path. My guess is that either `Profile` and `MyProfile` are going to get hit by the routes

ansh:
@pfrazee It did work but it was such a hack. I created a new solution that is also a hack but a little better hack in my opinion since it isolates the hack into 3 lines of code

src/view/shell/Drawer.tsx
```diff
@@ -64,6 +64,10 @@ export const DrawerContent = observer(function DrawerContentImpl() {
       const state = navigation.getState()
       store.shell.closeDrawer()
       if (isWeb) {
+        // hack because we have flat navigator for web and MyProfile does not exist on the web navigator -ansh
+        if (tab === 'MyProfile') {
+          navigation.navigate('Profile', {name: store.me.handle})
+        }
```
pfrazee, line 70: Need an else here, yeah?

pfrazee (changes requested):
Okay yeah this looks better to me. One small question

src/view/shell/Drawer.tsx
ansh: oops yeah thanks for the catch

ansh: Fixed! @pfrazee Pls approve so I can merge

ansh: @pfrazee Done, can you approve so I can merge?

pfrazee (approved):
🚀

PR #1446, opened 2023-09-13
Fix 1405 handle android back button in composer

### LINKED ISSUE #1405: Pressing back in the composer deletes the post without warning
**Describe the bug**
In the Android App, pressing the back button when you are in the composer deletes the post without any warning. So you lose the content of the post, including the pictures and ALT values you typed if there were any.

**To Reproduce**

1. Enter the composer
2. Type in some text, add pictures etc.
3. Press back to close the keyboard, and back again to exit the composer.

**Expected behavior**
A confirmation message should display, to ask the user if they want to lose what they typed. (Even better would be the possibility to save it as a draft)

**Details**
 - Platform: Android
 - App version: 1.48.0

ansh:
Fixes #1405


ansh:
@pfrazee I agree. We should. It will be used elsewhere too. I can update this PR to do that.

pfrazee:
sg

ansh:
I thought about this a lot but I don't see any reason to introduce another event listener since the only component that would fire this "event" would be the `Composer` itself. I think the code is good as is. @pfrazee

estrattonbailey (approved):
Works well, interesting idea with listeners 🤔

pfrazee:
Does this run any chance of a double handling of the back button?

pfrazee:
@ansh I still had open questions about this, it wasn't ready for a merge

Issue #1371, opened 2023-09-04
reposted replies don't appear on main timeline


ameliamnesia:
**Describe the bug**
when reposting your own replies they aren't showing up on the main profile page, but they do appear on the "replies" tab and apparently in feeds as normal (i noticed a couple of people like it after reposting but can't see it on my main profile page without going to replies)

**To Reproduce**

1. repost one of your own replies in a thread
2. view main profile page

**Expected behavior**

the reposted reply (in the past) has shown up on the normal "posts" tab, visible to other users without moving to the "posts & replies" tab

**Screenshots**

![Screenshot_20230904-150953.png](https://github.com/bluesky-social/social-app/assets/133761448/0687e693-d709-4948-b2e3-7451890f36cf)

![Screenshot_20230904-151146.png](https://github.com/bluesky-social/social-app/assets/133761448/cc483a57-86a2-4730-9651-cc91c056c896)

**Details**

 - Platform: android and mobile chrome
 - Platform version: 14
 - App version: 1.48.0

**Additional context**
this seems to happen whether you're logged in to the account in question or not, i thought maybe it only applied to seeing your own reposted replies but am seeing the same behavior when logged into a different account

direct link to the reply i noticed this on:
https://bsky.app/profile/ameliamnesia.xyz/post/3k6iiv7dopa2g

pfrazee:
Heh got another one for you @estrattonbailey

estrattonbailey:
Thanks! This is related to #1166, I think we can do both at the same time 👍

PR #1450, opened 2023-09-14
don't thread a self-repost of a self-reply

estrattonbailey:
Fixes https://github.com/bluesky-social/social-app/issues/1371

Originally thought this was a backend thing, but actually we were just mistakenly threading the reposted `reply` if the feed response didn't also contain the plain-ol' `reply` itself. See below for example:

Screen Shot 2023-09-13 at 3 45 52 PM

Whereas in the posts-and-replies view we get both the plain-ol' reply _and_ the reposted reply, so it doesn't get threaded.
Screen Shot 2023-09-13 at 3 45 58 PM

So the fix looks like this (see last two posts in feed):
Screen Shot 2023-09-14 at 12 28 06 PM

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -281,7 +281,10 @@ export class FeedTuner {
 
 function getSelfReplyUri(item: FeedViewPost): string | undefined {
   if (item.reply) {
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Makes sense!

PR #1451, opened 2023-09-14
fix theme switching on web without refresh

estrattonbailey:
It's allll coming together now.

Dan brought this up, great call: https://github.com/bluesky-social/social-app/pull/1417#issuecomment-1719879738

### DIFF
--- a/src/lib/ThemeContext.tsx
+++ b/src/lib/ThemeContext.tsx
@@ -1,3 +1,4 @@
+import {isWeb} from 'platform/detection'
 import React, {ReactNode, createContext, useContext} from 'react'
 import {
   AppState,
@@ -101,6 +102,10 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({
     React.useState<ColorSchemeName>(colorSchemeFromRN)
 
   React.useEffect(() => {
+    if (isWeb) {
+      setNativeColorScheme(colorSchemeFromRN)
+    }
+
     const subscription = AppState.addEventListener('change', state => {
       const isActive = state === 'active'
 
--- a/src/lib/ThemeContext.tsx
+++ b/src/lib/ThemeContext.tsx
@@ -3,7 +3,7 @@ import React, {ReactNode, createContext, useContext} from 'react'
 import {
   AppState,
   TextStyle,
-  useColorScheme,
+  useColorScheme as useColorScheme_BUGGY,
   ViewStyle,
   ColorSchemeName,
 } from 'react-native'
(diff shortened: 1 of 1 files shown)


src/lib/ThemeContext.tsx
```diff
@@ -101,6 +102,10 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({
     React.useState<ColorSchemeName>(colorSchemeFromRN)
 
   React.useEffect(() => {
+    if (isWeb) {
```
gaearon, line 105: Instead of doing this in an effect, can we just treat `colorSchemeFromRN` as the source of truth when `isWeb` is true? And ideally rewrite the ternary below to be easier to follow.

src/lib/ThemeContext.tsx
```diff
@@ -92,33 +93,32 @@ export const ThemeContext = createContext<Theme>(defaultTheme)
...
 }) => {
   const colorSchemeFromRN = useColorScheme()
-  const [nativeColorScheme, setNativeColorScheme] =
+  const [statefulColorScheme, setStatefulColorScheme] =
     React.useState<ColorSchemeName>(colorSchemeFromRN)
+  const colorScheme = isWeb ? colorSchemeFromRN : statefulColorScheme
+  const themeValue = getTheme(theme === 'system' ? colorScheme : theme)
 
   React.useEffect(() => {
+    if (!isWeb) return
```
gaearon, line 110: I find the use of negation a bit tricky to follow here. Why do we exit early when we're _not_ on the web? Isn't this block of code relevant *when* we're not on the web (i.e. specifically for iOS)? 

If I'm not wrong, we should flip this to `if (isWeb) return`.

As a nitpick these conditions are tricky to think about so I'd prefer less use of early returns here. (Curious if you disagree!) Assuming my earlier suggestion was correct:

```js
  React.useEffect(() => {
    // With RN iOS, we can only "trust" the color scheme reported while the app is active.
    // This bug will probably be fixed by https://github.com/facebook/react-native/pull/39439
    if (!isWeb) {
      const subscription = AppState.addEventListener('change', state => {
        if (state === 'active') {
          setStatefulColorScheme(colorSchemeFromRN)
        }
      })
      return () => subscription.remove()
    }
  }, [colorSchemeFromRN])
```

gaearon: This is also nitpicking, but I'll throw in a few more naming habits I picked up from FB as a proposal.

It might be nice to import the original as `useColorScheme_BUGGY` and then extract this part into a separate `useColorScheme_FIXED` Hook which encapsulates this fix. Then it'll be easy to get rid of the forked behavior without making mistakes when https://github.com/facebook/react-native/pull/39439 lands into a release we're using.

```js
import {
  // ...
  useColorScheme as useColorScheme_BUGGY,
  // ...
} from 'react-native'

// ...

const useColorScheme_FIXED = () => {
  const colorSchemeFromRN = useColorScheme_BUGGY()
  const [savedColorScheme, setSavedColorScheme] = useState<ColorSchemeName>(colorSchemeFromRN)
  const colorScheme = isWeb ? currentScheme : savedColorScheme
  // ...
  return colorScheme
}

export const ThemeProvider: React.FC<ThemeProviderProps> = ({
  // ...
}) => {
  const colorScheme = useColorScheme_FIXED()
  // ...
```

Then when the bug is fixed, we'll remove alias from the import, delete `useColorScheme_FIXED`, and call the imported `useColorScheme` directly.

estrattonbailey: >If I'm not wrong

You're not wrong 🤦 good catch

estrattonbailey: I like the hook idea too, let's adopt that throughout 👍

src/lib/ThemeContext.tsx
```diff
@@ -92,33 +93,37 @@ export const ThemeContext = createContext<Theme>(defaultTheme)
...
+  return theme === 'dark' ? darkTheme : defaultTheme
+}
+
+function useColorScheme_FIXED() {
+  const colorScheme = useColorScheme_BUGGY()
+  const [currentColorScheme, setCurrentColorScheme] =
+    React.useState<ColorSchemeName>(colorScheme)
 
   React.useEffect(() => {
+    // we don't need to be updating state on web
```
gaearon, line 106: I would prefer if we link to the supposed fix so that we can know later when to rip this out without going through the blame and PR history.

gaearon (approved):
lgtm

PR #1452, opened 2023-09-14
bump picker deps

### LINKED ISSUE #1278: option to choose between photo picker or file browser when uploading images on android
todays change to using only the photo picker is not a good one. 

an option to choose between photo picker or file browser would be a better way, photo picker cannot access hidden folders or cloud storage

estrattonbailey:
Maybe fix #1278


pfrazee (approved):
No reason no to merge, but it doesnt look like this fixes our problemo

estrattonbailey:
Yeah agreed, need to do more looking

PR #1454, opened 2023-09-14
add build:apk and apk config

estrattonbailey:
Sets us up to build APK bundles for install on real Android devices.

Log into Expo, then run:

```
yarn build:apk
```

You can download the build artifacts from the Expo dashboard.


pfrazee (approved):
nice

PR #1455, opened 2023-09-14
Drop all the in-post translate links except when expanded

pfrazee:
This drops it down to the expanded view:

CleanShot 2023-09-14 at 14 42 10@2x

While for others you just use the menu:

CleanShot 2023-09-14 at 14 42 21@2x
CleanShot 2023-09-14 at 14 42 36@2x


estrattonbailey (approved):
🚢

PR #1457, opened 2023-09-15
Onboarding recommended follows

ansh:
https://github.com/bluesky-social/social-app/assets/8207733/79778941-3929-4cba-bb4f-afdd0264e438


estrattonbailey (approved):
Nice one! I think we should return here and smooth the experience a bit, esp on web, but I also think the utility is excellent and it will benefit users, so I stamp 🚢

pfrazee (approved):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #1462, opened 2023-09-15
Sanitize BIDI characters from display names

pfrazee:
Before

CleanShot 2023-09-15 at 13 53 37@2x

---

after

CleanShot 2023-09-15 at 13 53 17@2x


estrattonbailey (approved):
Looks like it works great

agentjabsco:
to be clear, not only do I fully support this, [I think it should go further](https://bsky.app/profile/jabsco.cia.fyi/post/3jyf4sucadx25)

Issue #650, opened 2023-05-16
Middle click/ctrl click to open post in new tab


foxydevloper:
**Is your feature request related to a problem? Please describe.**

Middle clicking/ctrl clicking to open a post in a new tab is a very very useful shortcut, available on most websites (like twitter) and in html, however this app doesn't appear to support it. It makes opening multiple posts a pain with no clear easy way to get around it.

**Describe the solution you'd like**

Middle click & ctrl click to open post in new tab, which is standard web behavior

**Describe alternatives you've considered**

**Additional context**

Supported by most websites, even electron applications will literally open up a new tab in your browser if you middle click on something

Nazeeh21:
Would like to work on this issue. Can you please assign me this?

benharri:
as a workaround for now you can middle click on the timestamp

Shohreh:
What's middle click? My mouse has two buttons and a scroll wheel.

The lack of the usual CTRL+left click on a link to open a thread in a new tab is really annoying.

benharri:
> What's middle click? My mouse has two buttons and a scroll wheel.

click with the scroll wheel

Shohreh:
Doesn't work on the mouse I have.

I'll just wait until BS adds this common feature.

icecube45:
+1ing this, it's a shortcut I'm sorely missing. 
The referenced PR does not seem to have addressed this.

icecube45:
Both chrome (120.0.6099.130) and firefox (117.0) don't work for me (windows 10)

benharri:
seems to be working with firefox 121 on win 11

icecube45:
Very happy to have this working on the main feeds now! I did notice it doesn't work on search results, though

haileyok:
@icecube45 feel free to open up an issue, i'll get that working whenever i have some time!

PR #1469, opened 2023-09-18
:bug: Handle middle mouse click on feed list items

foysalit:
Addresses https://github.com/bluesky-social/social-app/issues/650

This is a PoC that attempts to handle middle mouse click on post feed items and opens the post in a new tab. This is default behavior when using `a` tags but on non-a tags, links only open in new tab on safari (which is an anomaly and will probably be patched down the road on safari too).

I'd imagine the handler will have to be added on other wrappers where posts are rendered.
More details on the default web-supported handler for this event https://github.com/facebook/react/pull/11571

https://github.com/bluesky-social/social-app/assets/1919066/33a3ae5f-dcb6-4a2b-965b-b3fa5cd5eaa0

### DIFF
--- a/src/view/shell/index.web.tsx
+++ b/src/view/shell/index.web.tsx
@@ -16,11 +16,13 @@ import {useWebMediaQueries} from '../../lib/hooks/useWebMediaQueries'
 import {BottomBarWeb} from './bottom-bar/BottomBarWeb'
 import {useNavigation} from '@react-navigation/native'
 import {NavigationProp} from 'lib/routes/types'
+import {useAuxClick} from 'lib/hooks/useAuxClick.web'
 
 const ShellInner = observer(function ShellInnerImpl() {
   const store = useStores()
   const {isDesktop, isMobile} = useWebMediaQueries()
   const navigator = useNavigation<NavigationProp>()
+  useAuxClick()
 
   useEffect(() => {
     navigator.addListener('state', () => {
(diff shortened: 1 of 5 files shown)


pfrazee (commented):
Is it not possible to handle this by attaching to document.body and firing a synthetic `click` even with the metaKey set?

foysalit:
> Is it not possible to handle this by attaching to document.body and firing a synthetic `click` even with the metaKey set?

hmm... we could but then we would have to ignore safari. let me draw that up.

foysalit:
alright @pfrazee check out the new approach.

src/view/shell/index.web.tsx
```diff
@@ -16,11 +16,13 @@ import {useWebMediaQueries} from '../../lib/hooks/useWebMediaQueries'
 import {BottomBarWeb} from './bottom-bar/BottomBarWeb'
 import {useNavigation} from '@react-navigation/native'
 import {NavigationProp} from 'lib/routes/types'
+import {useAuxClick} from 'lib/hooks/useAuxClick.web'
```
pfrazee, line 19:
```suggestion
import {useAuxClick} from 'lib/hooks/useAuxClick'
```

The `.web` extension gets automatically added by the bundler

pfrazee (approved):
Okay let's ship it!

pfrazee:
...after I fix the thing I broke

PR #1472, opened 2023-09-19
Nicer 'post processing status' in the composer

pfrazee:
Small visual improvement that I've been wanting to get done forever. The information line is now anchored on the bottom (no more reflowing layout) and looks a smidge nicer

CleanShot 2023-09-18 at 17 16 14@2x


gaearon (approved):
lg

PR #1475, opened 2023-09-19
Add thread sort settings

pfrazee:
As requested!

CleanShot 2023-09-18 at 18 01 03@2x

### DIFF
--- a/src/state/models/content/post-thread.ts
+++ b/src/state/models/content/post-thread.ts
@@ -241,7 +241,7 @@ export class PostThreadModel {
       res.data.thread as AppBskyFeedDefs.ThreadViewPost,
       thread.uri,
     )
-    sortThread(thread)
+    sortThread(thread, this.rootStore.preferences)
     this.thread = thread
   }
 }
@@ -263,11 +263,16 @@ function pruneReplies(post: MaybePost) {
   }
 }
 
+interface SortSettings {
+  threadDefaultSort: string
+  threadFollowedUsersFirst: boolean
+}
+
 type MaybeThreadItem =
   | PostThreadItemModel
   | AppBskyFeedDefs.NotFoundPost
   | AppBskyFeedDefs.BlockedPost
-function sortThread(item: MaybeThreadItem) {
+function sortThread(item: MaybeThreadItem, opts: SortSettings) {
   if ('notFound' in item) {
     return
   }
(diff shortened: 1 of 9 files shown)


emilyliu7321:
`Sort Replies By ... ` -> `Sort Replies`
`Replies always show by ... ` -> `Sort replies to the same post by: `
`Followed Users First` -> `Prioritize Your Follows`
`Set this setting to...` -> `Show replies by people you follow before all other replies.` (and keep the `Yes` on the toggle itself)

Sininini:
Since you're poking at reply sorting may I suggest always keeping, or at least having an option to keep your own replies at the very top. It's annoying to have to scroll through lots of replies to find your own.

src/state/models/content/post-thread.ts
```diff
@@ -296,13 +301,31 @@ function sortThread(item: MaybeThreadItem) {
       if (modScore(a.moderation) !== modScore(b.moderation)) {
         return modScore(a.moderation) - modScore(b.moderation)
       }
-      if (a.post.likeCount === b.post.likeCount) {
-        return b.post.indexedAt.localeCompare(a.post.indexedAt) // newest
-      } else {
-        return (b.post.likeCount || 0) - (a.post.likeCount || 0) // most likes
+      if (opts.threadFollowedUsersFirst) {
+        const af = a.post.author.viewer?.following
+        const bf = b.post.author.viewer?.following
```
estrattonbailey, line 306: Sick

src/state/models/content/post-thread.ts
```diff
@@ -296,13 +301,31 @@ function sortThread(item: MaybeThreadItem) {
...
+      } else if (opts.threadDefaultSort === 'most-likes') {
+        if (a.post.likeCount === b.post.likeCount) {
+          return b.post.indexedAt.localeCompare(a.post.indexedAt) // newest
+        } else {
+          return (b.post.likeCount || 0) - (a.post.likeCount || 0) // most likes
+        }
+      } else if (opts.threadDefaultSort === 'random') {
+        return 0.5 - Math.random() // this is vaguely criminal but we can get away with it
       }
+      return b.post.indexedAt.localeCompare(a.post.indexedAt)
```
estrattonbailey, line 326: This default is technically `newest` and not `oldest` like it is in preferences, right?

src/state/models/ui/preferences.ts
```diff
@@ -189,6 +194,21 @@ export class PreferencesModel {
...
       }
+      // check if thread sort order is set in preferences, then hydrate
+      if (
+        hasProp(v, 'threadDefaultSort') &&
+        typeof v.threadDefaultSort === 'string' &&
+        THREAD_SORT_VALUES.includes(v.threadDefaultSort)
+      ) {
+        this.threadDefaultSort = v.threadDefaultSort
+      }
+      // check if tread followed-users-first is enabled in preferences, then hydrate
```
estrattonbailey, line 205: Nit, don't need it
```suggestion
      // check if thread followed-users-first is enabled in preferences, then hydrate
```

src/view/screens/PreferencesThreads.tsx
```diff
@@ -0,0 +1,155 @@
...
+              Sort replies to the same post by:
+            </Text>
+            <View style={[pal.view, {borderRadius: 8, paddingVertical: 6}]}>
+              <RadioGroup
+                type="default-light"
+                items={[
+                  {key: 'oldest', label: 'Oldest replies first'},
+                  {key: 'newest', label: 'Newest replies first'},
+                  {key: 'most-likes', label: 'Most-liked replies first'},
+                  {key: 'random', label: 'Random (aka "Poster\'s Roulette")'},
```
estrattonbailey, line 59: Would be ideal to generate this from the `THREAD_SORT_VALUES` array

estrattonbailey: Also lawl Poster's Roulette

estrattonbailey (approved):
Couple nits, v nice

src/state/models/content/post-thread.ts
pfrazee: yes but that should never hit unless something goes real wrong

PR #1481, opened 2023-09-19
add suggested follow section to profile header

estrattonbailey:
Adds a section to `ProfileHeader` that displays the results of `getSuggestedFollowsByActor`.
- fetches fresh results each time you open this section
- section is opened and closed automatically upon follow/unfollow
- section can be toggled separately using new button
- not rendered on preview
- shows skeleton UI when loading

https://github.com/bluesky-social/social-app/assets/4732330/e3f874a3-e7b0-4b74-a800-fbe1b1e419d0

Screen Shot 2023-09-19 at 3 38 07 PM
Screen Shot 2023-09-19 at 3 38 16 PM
Screen Shot 2023-09-19 at 3 40 42 PM

### DIFF
--- a/src/lib/hooks/useFollowDid.ts
+++ b/src/lib/hooks/useFollowDid.ts
@@ -1,0 +1,35 @@
+import React from 'react'
+
+import {useStores} from 'state/index'
+import {FollowState} from 'state/models/cache/my-follows'
+
+export function useFollowDid({did}: {did: string}) {
+  const store = useStores()
+  const state = store.me.follows.getFollowState(did)
+
+  return {
+    state,
+    following: state === FollowState.Following,
+    toggle: React.useCallback(async () => {
+      const freshState = await store.me.follows.fetchFollowState(did)
+
+      if (freshState === FollowState.Following) {
+        try {
+          await store.agent.deleteFollow(store.me.follows.getFollowUri(did))
+          store.me.follows.removeFollow(did)
+        } catch (e: any) {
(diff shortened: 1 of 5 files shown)


src/lib/hooks/useFollowDid.ts
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+
+import {useStores} from 'state/index'
+import {FollowState} from 'state/models/cache/my-follows'
+
+export function useFollowDid({did}: {did: string}) {
```
estrattonbailey, line 6: Logic copied from `FollowButton`

ansh (approved):
Works well 👍🏽 

I would probably add a "close" button to the top right of suggested follows scrollview, as well as a header on the top left with just says "Suggested follows" as people might be confused about this.

gaearon:
I think a close button is definitely needed. I also find the "suggested follows" button *very* unintuitive (both the icon that's used and its toggling behavior). It really looks like a "Follow" action based on the icon.

I would suggest:

- Removing the toggle button altogether (so only Follow triggers it)
- Adding the dismiss button to the panel

estrattonbailey:
Updated!

Screen Shot 2023-09-20 at 1 51 15 PM
Screen Shot 2023-09-20 at 1 51 27 PM
Screen Shot 2023-09-20 at 1 48 54 PM
Screen Shot 2023-09-20 at 1 49 19 PM

pfrazee (approved):
Nice!

PR #1486, opened 2023-09-20
Feeds tab fixes

pfrazee:
- Bold the saved feeds on mobile (looked wrong to me)
- Better loading state of saved feeds (placeholders)
- Add soft reset to the feeds page (tapping on the # when already open now reloads)
- Show feed descriptions in profile listings
- Add an "About this feed" button to the dropdown to see the details

I'm not in love with the "about this feed" UI but it's enough for now.

CleanShot 2023-09-19 at 20 39 42@2x
CleanShot 2023-09-19 at 20 40 10@2x

### DIFF
--- a/src/state/models/ui/my-feeds.ts
+++ b/src/state/models/ui/my-feeds.ts
@@ -10,6 +10,11 @@ export type MyFeedsItem =
     }
   | {
       _reactKey: string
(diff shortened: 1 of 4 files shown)


pfrazee:
going to merge w/o review so I can put up a testflight tonight

PR #1489, opened 2023-09-20
Allow touch at the top of the lightbox

gaearon:
## What

The lightbox top bar (which contains the close button) was eating touch even outside the close button tap area.

This was confusing because the top bar is invisible.

Let's fix it so that it doesn't eat touch.

Verified close button still works.

## Before

https://github.com/bluesky-social/social-app/assets/810438/f8039da1-341b-44e0-9d92-0e84b5b91432

## After

https://github.com/bluesky-social/social-app/assets/810438/97fde8b3-5d30-4112-a4c9-f0ad0d717a7b


ansh (approved):
👍🏽 Good to merge

redian:
Looks like this Action is not planning to stop anytime soon:
https://github.com/bluesky-social/social-app/actions/runs/6248036305/job/16961810293

PR #1498, opened 2023-09-21
Prefilter the mergefeed to ensure a better mix of following and custom feeds

pfrazee:
During QA I found that the merge feed was giving the wrong mix of following & custom feeds. This turned out to be because the sampling between these sources occurred prior to the client-side filters. With this PR, a prepass of the filtering is added, causing the sample rate to be much more accurate. I also bumped the following fetch count to offset the loss of posts due to that filter.


estrattonbailey (approved):
LGTM

PR #1499, opened 2023-09-21
Test suite improvements & tests for the mergefeed

pfrazee:
- Adds a `<TestCtrls>` component which only renders in the e2e sim build and provides some fast actions, dramatically improving the execution time of tests.
- Disables the invite codes test temporarily due to the drawer crash issue.
- Adds mergefeed tests.


estrattonbailey (approved):
Great idea

PR #1505, opened 2023-09-21
Simplify Array.from + map

Connormiha:
`Array.from` has optional second argument. Calling `map` is not required.


pfrazee:
You're right and I appreciate the contribution, but I don't see the upside of moving to a more niche behavior.

PR #1506, opened 2023-09-21
Tune the treeview mode to use larger immediate-children

pfrazee:
Another iteration after a night of QA. We need to balance horizontal screen usage with overall visual clarity and appeal. With this update, I modified the immediate children to show as with normal linear posts, but then further children go into the thread mode. I think this feels more natural.

**Desktop**

|Before|After|
|-|-|
|CleanShot 2023-09-21 at 10 56 05@2x|CleanShot 2023-09-21 at 10 55 22@2x|

**Mobile**

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/src/view/com/composer/Prompt.tsx
+++ b/src/view/com/composer/Prompt.tsx
@@ -33,7 +33,7 @@ export function ComposePrompt({onPressCompose}: {onPressCompose: () => void}) {
 
 const styles = StyleSheet.create({
   prompt: {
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
I like the top part a lot, nice to not have the left-most line. But I actually think I really like how the avi is stack on top of the text, even on immediate-children. Feels like a more efficient use of space. Just my 2¢

PR #1507, opened 2023-09-21
Move home feed and thread preferences to server

pfrazee:
Updates home feed and thread preferences to store server-side.

### DIFF
--- a/src/state/models/ui/preferences.ts
+++ b/src/state/models/ui/preferences.ts
@@ -1,7 +1,8 @@
 import {makeAutoObservable, runInAction} from 'mobx'
 import {
   LabelPreference as APILabelPreference,
-  AppBskyActorDefs,
+  BskyFeedViewPreference,
+  BskyThreadViewPreference,
 } from '@atproto/api'
 import AwaitLock from 'await-lock'
 import isEqual from 'lodash.isequal'
@@ -16,11 +17,12 @@ import {LANGUAGES} from '../../../locale/languages'
 
 // TEMP we need to permanently convert 'show' to 'ignore', for now we manually convert -prf
 export type LabelPreference = APILabelPreference | 'show'
-export type FeedViewPreference = Omit<AppBskyActorDefs.FeedViewPref, '$type'>
-export type ThreadViewPreference = Omit<
-  AppBskyActorDefs.ThreadViewPref,
-  '$type'
->
+export type FeedViewPreference = BskyFeedViewPreference & {
+  lab_mergeFeedEnabled?: boolean | undefined
+}
+export type ThreadViewPreference = BskyThreadViewPreference & {
+  lab_treeViewEnabled?: boolean | undefined
+}
 const LABEL_GROUPS = [
   'nsfw',
   'nudity',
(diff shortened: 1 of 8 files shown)


src/state/models/ui/preferences.ts
```diff
@@ -576,25 +558,25 @@ export class PreferencesModel {
...
-      if (areRepliesEnabled) {
+      if (this.homeFeed.hideReplies) {
+        feedTuners.push(FeedTuner.removeReplies)
+      } else {
         feedTuners.push(
           FeedTuner.thresholdRepliesOnly({
             userDid: this.rootStore.session.data?.did || '',
-            minLikes: repliesThreshold,
-            followedOnly: areRepliesByFollowedOnlyEnabled,
+            minLikes: Number(this.homeFeed.hideRepliesByLikeCount),
```
estrattonbailey, line 573: This can be undefined right? Would result in NaN? I see there's an `|| 0` elsewhere.

src/state/models/ui/preferences.ts
```diff
@@ -13,6 +16,11 @@ import {LANGUAGES} from '../../../locale/languages'
 
 // TEMP we need to permanently convert 'show' to 'ignore', for now we manually convert -prf
 export type LabelPreference = APILabelPreference | 'show'
+export type FeedViewPreference = Omit<AppBskyActorDefs.FeedViewPref, '$type'>
+export type ThreadViewPreference = Omit<
+  AppBskyActorDefs.ThreadViewPref,
+  '$type'
+>
```
estrattonbailey, line 23: These types are resulting in `unknown`s, unsure why it's flattening to it's most basic type value 🤔

src/state/models/ui/preferences.ts
```diff
@@ -298,6 +254,14 @@ export class PreferencesModel {
     await this.rootStore.me.savedFeeds.updateCache(clearCache)
   }
 
+  async syncLegacyPreferences() {
+    if (this.legacyPreferences) {
```
estrattonbailey, line 258: Do we need to also delete the legacy prefs from local storage so that `getLegacyPreferences` returns undefined, or is it enough to have removed them from `serialize`?

estrattonbailey (commented):
Types are funky, but that might require a `api` package change so not a huge deal rn.

Everything works great except `hideRepliesByLikeCount` isn't being reflected between devices? Maybe I'm doing it wrong.

src/state/models/ui/preferences.ts
pfrazee: Yeah it's bugging me too. You know what, gonna manually spec it locally.

pfrazee: Yeah it seems to be the `Omit<>` modifier that ruins it

pfrazee: https://github.com/bluesky-social/atproto/pull/1653

src/state/models/ui/preferences.ts
pfrazee: It's enough to remove them from serialize

src/state/models/ui/preferences.ts
pfrazee: It can't actually ever be undefined due to JS serialization behaviors; the updated types PR will more properly reflect this.

pfrazee:
Merging for nightly

PR #1510, opened 2023-09-21
Switch the date input web version to the date-input web control

pfrazee:
Some users were struggling with the date localization of the freeform web input, so this switches us to using `<input type="date">`. Tested on FF, Chrome, and Safari on desktop. Caniuse gives good support for the element and for `valueAsDate` on all browsers (including mobile).

CleanShot 2023-09-21 at 16 10 57@2x

Dark mode doesnt show the date picker icon. Frustratingly that would require CSS changes and I don't really want to bother with that right now.

CleanShot 2023-09-21 at 16 12 05@2x


estrattonbailey (approved):
Nice one.

Calendar icon works in Firefox 😏

Issue #1268, opened 2023-08-24
cmd+click on user name should open the profile in a new tab, as does cmd+click on avatar


mschwendener:
**Describe the bug**, **To Reproduce**

https://bsky.app/profile/benjaminpercy.bsky.social/post/3k5pob23uhr2e 

Clicking on the poster's name or photo both bring me to https://bsky.app/profile/benjaminpercy.bsky.social 

On the **replies**, also cmd+click on the name or the photo of a person behave the same. 

But: for the **first** post, cmd+click only works on the avatar, not on the name. 

(let me know if it's not clear what I mean, then I'll try to describe it better)

**Expected behavior**

Also for the first post, cmd+click on the name should behave the same as cmd+click on the avatar. 

**Details**

Chrome on MacBook Pro with trackpad.

**Additional context**

Related? https://github.com/bluesky-social/social-app/issues/650

mschwendener:
PS Same in Chrome on Windows. 

And to add: when I cmd+click / right click on the same name, looking at the same post, but as part of a feed, it also works as expected.

mschwendener:
This might be related? https://github.com/bluesky-social/social-app/issues/1286

intrnl:
doesn't seem to be, the issue you linked is about mentions not being linked properly in profiles so you can't click them at all

mschwendener:
Thank you, @intrnl 

> the issue here is about being able to ctrl+click ... the mentions to open them in a new tab, right?

Yes, exactly.

mschwendener:
While **cmd**+click now behaves as expected, on Firefox and Chrome on MacBook the following is still happening: 

Example: https://bsky.app/profile/mpgphd.bsky.social/post/3kg5si5p6vq2w 

I can ctrl+click on the avatar and open the link in a new tab, but this doesn't work with the user name.

PR #1512, opened 2023-09-22
Improve Device Detection For Better Responsiveness

liby:
## Motivation

The creation of this Pull Request was motivated by the need to address a particular issue, as referenced in the issue https://github.com/bluesky-social/social-app/issues/1268. 

After further investigation, I found that when users adjust the size of their desktop browser, [a specific variable `isDesktopWeb` will erroneously return `false`](https://github.com/bluesky-social/social-app/blob/3c4899b3c4ae55285559215023a25e365c9f4e30/src/view/com/util/Link.tsx#L225). At the same time, I also noticed that `isDesktopWeb` is used in many places in the project, which may cause some style issues because `isDesktopWeb` does not perform responsive updates.

Then I found [`useWebMediaQueries` hook](https://github.com/bluesky-social/social-app/blob/764c7cd5694a41c98d8543b68d7791fa90db4291/src/lib/hooks/useWebMediaQueries.tsx#L4) in the project. This Hook integrates responsive updates in device detection and represents a marked improvement over the existing method. Hence, in this pull request, I propose using this Hook as a replacement for the current logic.

### https://github.com/bluesky-social/social-app/commit/d83738a0e709bbb407a53fba37b014a26251925e
  - Replace static `isDesktopWeb` with `useWebMediaQueries` hook to enable dynamic device type detection.
  - Create `useDeviceLimits` hook to dynamically determine `DY_LIMIT_UP` and `DY_LIMIT_DOWN` based on device type.
  - Update dependency arrays for the `useCallback` hooks to include new dynamic variables.
  

### https://github.com/bluesky-social/social-app/commit/63521ce896e675e311b9a24ec2c3b9b71455cd12
  - Create `useStyles` hook that generates styles object based on device type detected from `useWebMediaQueries`.
  - Replace static styles object with dynamic styles object generated from `useStyles` hook in components.
  - This allows `paddingLeft` values for 'ul' and 'ol' styles to adapt to device type dynamically.
  - This allows `maxWidth` values for 'metaItem'' styles to adapt to device type dynamically.
  
### https://github.com/bluesky-social/social-app/commit/61f68711d2842c50630e43202590680382102960
  - Removed the unused `isDesktopWeb` from _src/platform/detection.ts_.
  - Replaced `isDesktopWeb` with `useWebMediaQueries().isDesktop` in the remaining React components. 

### https://github.com/bluesky-social/social-app/commit/20251c6843d142244196d0884744068b2f864ae8
 - Replaced the usage of `isMobileWeb` exported from _src/platform/detection.ts_ in `SplashScreen` component with `isWeb && useWebMediaQueries().isTabletOrMobile`.

## Others

1. This does not solve issue https://github.com/bluesky-social/social-app/issues/1268 as using Media Query inside `DesktopWebTextLink` to determine if it is desktop web is fundamentally inappropriate. User Agent would be more reasonable for such a purpose. Considering the complexity, should we consider introducing [react-device-detect](https://www.npmjs.com/package/react-device-detect)?

2. There is still [one location where `isMobileWeb`](https://github.com/bluesky-social/social-app/blob/49356700c31a1cb34c252e3aecf18561114916b9/src/lib/styles.ts#L176) is being used that hasn't been replaced. As it affects a wide range of areas, careful planning and consideration are required for this task.

### DIFF
--- a/src/view/com/composer/photos/OpenCameraBtn.tsx
+++ b/src/view/com/composer/photos/OpenCameraBtn.tsx
@@ -7,11 +7,11 @@ import {
 import {usePalette} from 'lib/hooks/usePalette'
 import {useAnalytics} from 'lib/analytics/analytics'
 import {useStores} from 'state/index'
-import {isDesktopWeb} from 'platform/detection'
 import {openCamera} from 'lib/media/picker'
 import {useCameraPermission} from 'lib/hooks/usePermissions'
 import {HITSLOP_10, POST_IMG_MAX} from 'lib/constants'
 import {GalleryModel} from 'state/models/media/gallery'
+import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 
 type Props = {
   gallery: GalleryModel
@@ -22,6 +22,7 @@ export function OpenCameraBtn({gallery}: Props) {
   const {track} = useAnalytics()
   const store = useStores()
   const {requestCameraAccessIfNeeded} = useCameraPermission()
+  const {isDesktop} = useWebMediaQueries()
 
   const onPressTakePicture = useCallback(async () => {
     track('Composer:CameraOpened')
@@ -43,7 +44,7 @@ export function OpenCameraBtn({gallery}: Props) {
     }
   }, [gallery, track, store, requestCameraAccessIfNeeded])
 
-  if (isDesktopWeb) {
+  if (isDesktop) {
     return null
   }
 
--- a/src/view/com/composer/photos/OpenCameraBtn.tsx
+++ b/src/view/com/composer/photos/OpenCameraBtn.tsx
@@ -11,7 +11,7 @@ import {openCamera} from 'lib/media/picker'
 import {useCameraPermission} from 'lib/hooks/usePermissions'
 import {HITSLOP_10, POST_IMG_MAX} from 'lib/constants'
 import {GalleryModel} from 'state/models/media/gallery'
-import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
+import {isMobileWeb, isNative} from 'platform/detection'
 
 type Props = {
   gallery: GalleryModel
@@ -22,7 +22,6 @@ export function OpenCameraBtn({gallery}: Props) {
   const {track} = useAnalytics()
   const store = useStores()
   const {requestCameraAccessIfNeeded} = useCameraPermission()
-  const {isDesktop} = useWebMediaQueries()
 
   const onPressTakePicture = useCallback(async () => {
     track('Composer:CameraOpened')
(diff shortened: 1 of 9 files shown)


src/view/com/composer/photos/SelectPhotoBtn.tsx
```diff
@@ -19,16 +19,17 @@ export function SelectPhotoBtn({gallery}: Props) {
   const pal = usePalette('default')
   const {track} = useAnalytics()
   const {requestPhotoAccessIfNeeded} = usePhotoLibraryPermission()
+  const {isDesktop} = useWebMediaQueries()
 
   const onPressSelectPhotos = useCallback(async () => {
     track('Composer:GalleryOpened')
 
-    if (!isDesktopWeb && !(await requestPhotoAccessIfNeeded())) {
```
pfrazee, line 26: This is actually an issue of the platform not the device size. We should change it to `isNative`

src/view/com/composer/photos/OpenCameraBtn.tsx
```diff
@@ -43,7 +44,7 @@ export function OpenCameraBtn({gallery}: Props) {
     }
   }, [gallery, track, store, requestCameraAccessIfNeeded])
 
-  if (isDesktopWeb) {
+  if (isDesktop) {
```
pfrazee, line 47: This was originally intended as a platform test not a device-size test. The interesting question is whether the camera module will work with web.

pfrazee (changes requested):
Really useful, thank you. I left a couple of small comments

src/view/com/composer/photos/OpenCameraBtn.tsx
liby: I've revised to use `isNative` and `isMobileWeb` checks for showing the camera button, which should cover native apps and mobile web

src/view/com/composer/photos/SelectPhotoBtn.tsx
```diff
@@ -23,12 +23,12 @@ export function SelectPhotoBtn({gallery}: Props) {
...
+    if (!isNative && !(await requestPhotoAccessIfNeeded())) {
```
pfrazee, line 26:
```suggestion
    if (isNative && !(await requestPhotoAccessIfNeeded())) {
```

pfrazee (approved):
Looks great! Thanks very much!

src/view/com/composer/photos/SelectPhotoBtn.tsx
liby: Sorry, I didn't notice the problem here. I recently contracted COVID-19, which has left me a bit out of sorts.

pfrazee: No problem -- please get better!

PR #1535, opened 2023-09-26
fix 'Reposted by' text overflow

estrattonbailey:
(Extra "Reposted by" text just for testing)

Looks like this now:
Screen Shot 2023-09-26 at 10 17 59 AM

Prev:
image


pfrazee (approved):
Nice, works great

PR #1537, opened 2023-09-26
:lipstick: Remove report option on my own post

foysalit:
Doesn't make a lot of sense for the author of a post to report their own post and with the dropdown growing ever so slightly, getting rid of an unnecessary item feels nice.


pfrazee (approved):
🚀

PR #1542, opened 2023-09-26
Small PFP on desktop

Tanza3D:
Aligns with back button when visible, and generally looks less wonky in my opinion.

image

Also just about lines up with the icons for each page.


pfrazee:
Thanks! I went _slightly_ larger but I do think it's better
CleanShot 2023-09-28 at 12 11 36@2x

PR #1543, opened 2023-09-26
Move sidebar by 10px for more even padding

Tanza3D:
Looks nicer as it adds even padding on the left relative to the main content as there is on the top. Not reflected on the left sidebar as that already has padding on each button / element.

image


pfrazee (approved):
Yeah +1. This is how it was, we temporarily pulled it in while our responsive styles were bad. Glad to pull it back out

PR #1544, opened 2023-09-26
fix link highlighting with mention present

estrattonbailey:
While implementing hashtags I noticed that highlighting gets off IF there's a mention in the body text preceding the link (or tag). In fact, for each mention present, the subsequent highlighting is off by n+1.

While digging around I also learned of `generateJSON` which takes into account our plugins and does the work for us. I haven't been able to find an issue with this.

After (full link highlighted):
Screenshot 2023-09-26 at 6 26 18 PM

Before (last char not highlighted):
Screenshot 2023-09-26 at 6 26 06 PM


pfrazee (approved):
🔥

PR #1545, opened 2023-09-27
Add example google-services file

gabrieldonadel:
As an external contributor trying to build the app for the first time on Android throws a few errors due to a missing google-services.json.

image

This PR adds an example google-services.json file that users can use as a base to build the app on Android. We can't simply use an empty JSON since Google enforces validation on specific fields within it. 

Note: all values used in this file are mocked and are not related to any existing Firebase projects

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -8,6 +8,7 @@
   - brew tap wix/brew
   - brew install applesimutils
 - After initial setup:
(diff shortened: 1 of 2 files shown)


mozzius:
This would be great, maybe also add a note to the build instructions too?

gabrieldonadel:
> This would be great, maybe also add a note to the build instructions too?

Sure, updated

pfrazee (approved):
Thanks!

PR #1552, opened 2023-09-27
Composer prefilled mention - handle edge cases with DID URLs and users with invalid handles

mozzius:
I found that the behaviour of the composer where it prefills a mention when viewing a user's profile was producing incorrect behaviour when viewing a user with an invalid handle. This manifests in two bugs:

mobile: composer is prefilled with `@invalid.handle`

web: composer is prefilled with `@did:plc...`

Screenshot 2023-09-27 at 10 25 07

This is because the compose button simply looks at the URL - this would happen when viewing a profile with a valid handle too if the DID URL is used

For the first bug, I simply added a check for `handle === 'handle.invalid'`. For the second one, I added a check for `handle.startsWith('did:')` - this is not a great solution because this does not handle accounts with valid handles viewed with their DID in the URL, but it's better than the current, broken behaviour.

Profile used for testing:
https://bsky.app/profile/did:plc:3nqrhu5mthmias3zc4a2ovzj

### DIFF
--- a/src/view/screens/Profile.tsx
+++ b/src/view/screens/Profile.tsx
@@ -91,7 +91,10 @@ export const ProfileScreen = withAuthRequired(
     const onPressCompose = React.useCallback(() => {
       track('ProfileScreen:PressCompose')
       const mention =
-        uiState.profile.handle === store.me.handle ? '' : uiState.profile.handle
+        uiState.profile.handle === store.me.handle ||
+        uiState.profile.handle === 'handle.invalid'
+          ? undefined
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Love it 👍 

Great call on the problem of not having access to the handle in the nav. I was curious and and played around with solving this with our profile cache and seems to work, so gonna just merge into my branch and push up what I was working on since I've already got it locally ✌️

pfrazee:
Appreciate it @mozzius !

PR #1554, opened 2023-09-27
Larger user edit modal

Tanza3D:
This ends up making all modals larger, but I feel that ends up looking okay.
image
image


pfrazee:
I went with a slightly more modest change but did add some space

PR #1556, opened 2023-09-27
Improvements to auto-mentioning users from their profiles

estrattonbailey:
@mozzius pointed out two issues with our auto-mentioning of users when creating a new post:
- from the mobile new post button, if the user's handle was invalid we'd end up using `@handle.invalid`
- from the desktop nav, we didn't have profile UI state to check, so we ended up using the user's did

~The first is a real bug, the second technically works, but isn't a great experience.~ Edit: ope, for some reason I thought we did support linking to dids 🤔 

I pushed a commit to resolve the profile data from the cache within the desktop nav. This is just a hit to `AsyncStorage` so should typically be fast enough that users won't notice the await.

Invalid handle test profile: http://localhost:19006/profile/did:plc:3nqrhu5mthmias3zc4a2ovzj
Profile with did in url (Paul): http://localhost:19006/profile/did:plc:ragtjsm2j2vknwkz3zp4oxrd

### DIFF
--- a/src/view/screens/Profile.tsx
+++ b/src/view/screens/Profile.tsx
@@ -91,7 +91,10 @@ export const ProfileScreen = withAuthRequired(
     const onPressCompose = React.useCallback(() => {
       track('ProfileScreen:PressCompose')
       const mention =
-        uiState.profile.handle === store.me.handle ? '' : uiState.profile.handle
+        uiState.profile.handle === store.me.handle ||
+        uiState.profile.handle === 'handle.invalid'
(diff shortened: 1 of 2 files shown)


mozzius:
I tried posting a DID mention and while it came out blue in the composer, it was just plain text after posting. Is it supposed to work?

estrattonbailey:
@mozzius my b, you're correct, updated description 👍

pfrazee (approved):
Yeah lgtm

PR #1561, opened 2023-09-28
Don't render a pressable if there's only one notification author

### LINKED ISSUE #1559: Links from https://bsky.app/notifications to posts should work for likes as they do for replies
Looking at my https://bsky.app/notifications, I can click on "1d" in the notification for a reply, but not on "2d" in the notification for a like. 

If this is related to performance, I'm happy with it as it is. 

If performance is not the issue, I'd like the date stamp "2d" for the like to link back to my post, as the "1d" does for the reply. 

screenshot_4569

mozzius:
Fixes #1559

Allows the click to fall through to the underlying pressable, rather than a noop if there's only one person in the group of notifications

Also I've added the tooltip hover for the timestamp too


pfrazee (approved):
Awesome! Thanks mozz!

PR #1563, opened 2023-09-28
[WIP] Rewrite Android lightbox

### LINKED ISSUE #1485: Pinch gesture gets stuck on Android
Pinching an image in the lightbox gets stuck. There's likely two issues:

- Scroll view's pan recognizer "steals" the pinch gesture. We should fix this.
- The pan/pinch recognizer doesn't deal with termination well. We should fix this too.

gaearon:
Fixes #1485.

This is a comprehensive rewrite of the Android lightbox aligning it closer with behavior you see in stock Android apps. We're using Reanimated + RNGH since these provide more control and will let us compose with scroll views better.

**Not ready for a full review yet.**

- [x] Pinch and pan gestures
  - [x] Basic interaction
  - [x] Panning and pinching doesn't go offscreen (new!)
  - [x] Pinching is centered around focal point (new!)
  - [x] Panning is disabled while black bars are visible (new!)
  - [x] Enforce max zoom level
- [x] Feature parity
  - [x] Swipe away gesture
    - [x] Basic implementation
    - [x] Make it more tolerant X-wise
  - [x] Double tap gesture
  - [x] Show/hide controls based on zoom level
  - [x] Loading indicator
    - [ ] Check why sometimes it appears when you don't expect (multiple images?)
- [x] The Long Tail
  - [x] Regression: Pan during first pinch doesn't work
  - [x] Existing bug: It's possible to spam gestures during scroll swipe
    - [x] Fix overflowing images (e.g. quick double tap and swipe)
    - [x] Fix the scroll view getting stuck (e.g. pinch gesture during swipe)
  - [ ] Existing bug: swiping may sometimes get stuck
- [ ] Integration
  - [x] Re-enable the "gallery" scroll view above (temporarily disabled)
    - [x] Fix conflict with pinching
  - [ ] Verify we haven't regressed iOS
  - [ ] Delete dead code from old implementation
  - [ ] Inline weird helpers from old implementation
  - [ ] Fix types / lint
  - [ ] Move utility functions into own file


src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -42,7 +38,112 @@ type Props = {
...
+  number,
+  number
+];
+
+function createTransform(): Matrix3 {
+  'worklet'
+  return [1, 0, 0, 0, 1, 0, 0, 0, 1];
+}
+
+function prependTranslate(t: Matrix3, x: number, y: number) {
```
gaearon, line 58: Wrote these as in-place mutations to avoid allocating a bunch of stuff during gestures

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -42,7 +38,112 @@ type Props = {
...
+
+function prependScale(t: Matrix3, value: number) {
+  'worklet'
+  t[0] *= value;
+  t[1] *= value;
+  t[3] *= value;
+  t[4] *= value;
+}
+
+function prependTransform(ta: Matrix3, tb: Matrix3) {
```
wood1986, line 72: I do not know if this is matrix multiplication. But if it is, I would suggest you should use https://github.com/wcandillon/react-native-redash/blob/master/src/Matrix3.ts#L262. Then you do not need to have this code. Not only this, it has scaleMatrix and translation matrix.

image

gaearon: I’ve looked at it for inspiration and intentionally chose against it. I don’t want the extra bundle size on the web (and I don’t trust tree shaking to be reliable). I’d also like to avoid any unnecessarily allocations in the gesture handling code. Even if it’s fast, seems unnecessary to add GC churn for things that are enough to mutate in place. The version you linked to always allocates.

gaearon:
Still unhappy with how pan and pinch gestures compose but maybe it's not blocking. Left a comment in https://github.com/software-mansion/react-native-gesture-handler/issues/2616#issuecomment-1745657371 for the RNGH maintainers.

gaearon:
As for other notes, I'll do them in a follow-up. I want this PR to just get us to the baseline with the custom implementation. I'll polish up the code and make some videos to show the difference with what we had.

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -52,106 +158,238 @@ const ImageItem = ({
...
+    width: SCREEN.width,
+    height: SCREEN.height,
+    overflow: 'hidden'
   },
-  imageScrollContainer: {
-    height: SCREEN_HEIGHT * 2,
+  image: {
+    flex: 1,
   },
+  loading: {
```
chrisnojima, line 386: This is provided by RN if that's useful https://reactnative.dev/docs/stylesheet#absolutefill

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -42,7 +42,113 @@ type Props = {
...
+  ta[2] = a00 * tb[2] + a01 * tb[5] + a02 * tb[8];
+  ta[3] = a10 * tb[0] + a11 * tb[3] + a12 * tb[6];
+  ta[4] = a10 * tb[1] + a11 * tb[4] + a12 * tb[7];
+  ta[5] = a10 * tb[2] + a11 * tb[5] + a12 * tb[8];
+  ta[6] = a20 * tb[0] + a21 * tb[3] + a22 * tb[6];
+  ta[7] = a20 * tb[1] + a21 * tb[4] + a22 * tb[7];
+  ta[8] = a20 * tb[2] + a21 * tb[5] + a22 * tb[8];
+}
+
+function readTransform(t): [number, number, number] {
```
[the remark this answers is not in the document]
gaearon: yea i haven’t started fixing up types yet

gaearon:
I'll keep this for historical reasons but will resubmit cause a rebase was too hard and I needed to restructure things anyway.

gaearon:
Resubmit: https://github.com/bluesky-social/social-app/pull/1624

Issue #1514, opened 2023-09-22
Long press profile tab to quick switch accounts


mozzius:
**Is your feature request related to a problem? Please describe.**

Switching accounts is a common action for me, so I find going to the drawer menu then settings every time to be quite arduous

**Describe the solution you'd like**

It'd be great if long pressing the  profile tab would bring up a bottom sheet to quick switch accounts, like twitter does

**Describe alternatives you've considered**

Perhaps a quick switcher in the drawer menu?

**Additional context**

Here's a screenshot of twitter's menu - I'm imagining something pretty close to this:

![image](https://github.com/bluesky-social/social-app/assets/10959775/9b208a8d-c0fe-4525-89a3-9370def01831)

- [x] I would be willing to submit a PR for this

ansh:
Please submit a PR if you want/can!

PR #1555, opened 2023-09-27
Account quick switch modal

mozzius:
Closes #1514

PR #1567, opened 2023-09-28
Account quick switch modal

pfrazee:
Supercedes #1555

Closes https://github.com/bluesky-social/social-app/issues/1514

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -246,6 +246,7 @@ function TabsNavigator() {
     ),
     [],
   )
+
   return (
     <Tab.Navigator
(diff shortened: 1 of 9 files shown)


estrattonbailey (approved):
LGTM

Idea: kinda want a lil haptic beep-beep when the sheet opens. Instagram does this well in the same use case.

mozzius:
Yeah agreed - I left it out because Twitter doesn't do it, but it would definitely make it much more satisfying

pfrazee:
I added a small haptic feedback but not as cool as instagram (which does an additional one to indicate the long press has triggered, then another on release)

PR #1571, opened 2023-09-28
Fix alt text selectability in lightbox on web (close #1548)

### LINKED ISSUE #1548: cant select alt text - desktop firefox or android
cannot select alt text on android or firefox (to copy/paste )

tried multiple times

ideal situation, to be able to select alt text on photos

current behavior of mouse when hovering over alt text

PXL_20230927_041741015 MP

pfrazee:
A recent upgrade to react-native-web changed how `accessibilityRole` gets interpreted and turns it into a `<button>` when the value is `"button"` which then broke the selectability of the alt text in the lightbox. This fixes that.


ansh (approved):
Ah, looks good!

PR #1573, opened 2023-09-29
Warn the user on links that dont match their text

pfrazee:
Adds an interstitial for links in posts that dont match their text. If the text is not a URL, it just gives the user a normal "Leaving this site". If the text is a URL that doesnt match the target, it gives a "Possibly misleading" warning. If the link is the URL, it opens as usual.

CleanShot 2023-09-28 at 19 32 13@2x
CleanShot 2023-09-28 at 19 33 31@2x
CleanShot 2023-09-28 at 19 33 38@2x


ansh:
CleanShot 2023-09-29 at 07 18 47@2x
Dark mode broken

ansh (approved):

src/lib/strings/url-helpers.ts
```diff
@@ -166,3 +166,25 @@ export function getYoutubeVideoId(link: string): string | undefined {
   }
   return videoId
 }
+
+export function linkRequiresWarning(uri: string, label: string) {
+  try {
+    const urip = new URL(uri)
+    label = label.replace(/^http(s?):\/\//i, '')
+    return !label.startsWith(urip.hostname)
```
intrnl, line 174: URL#host and URL#hostname are normalized, so `new URL('https://Example.com')` with `Example.com` wouldn't match at all

intrnl:
because of the startsWith use, wouldn't it be worth to check what comes *after* the host? if it doesn't end with `/`, `?`, `#` or undefined, then it should be rejected

can imagine a scenario where someone would link to `website.co` but in actually it links to `website.com` which is a different site entirely.

DavidBuchanan314:
Something you might want to consider is an additional emphasis on the apex domain. Web users often do not understand the counterintuitive order of trust anchoring in domains. For example:

https://**paypal.com.billing-department-enquiries.<ins>example.com</ins>**/blah-blah-blah

intrnl:
@DavidBuchanan314

> Something you might want to consider is an additional emphasis on the apex domain, in the interstitial

Might as well ask it here, what's the right way you consider to check whether something is a domain or an IP address? is checking the end of the hostname to see if they're alphabetical or numerical the right choice here?

pfrazee:
Good suggestion @DavidBuchanan314 

Listed TLD:

CleanShot 2023-10-02 at 14 13 11@2x

Non-listed TLD:

CleanShot 2023-10-02 at 14 13 31@2x

pfrazee:
@intrnl dumb thing for me to be lazy about. Fixed now.

src/lib/strings/url-helpers.ts
pfrazee: Updated the code, it normalizes now and ignores case

pfrazee:
@ansh Can't believe I missed dark mode, thanks for hitting that

Should be good to go

PR #1587, opened 2023-10-03
Fix: to post positioning after thread load

pfrazee:
I was noticing some cases where quote posts would position incorrectly after tapping them and loading their thread screen. I believe this is because the index passed to `maintainVisibleContentPosition` is correct when rendering replies but not when rendering toplevel posts (you have to do some annoying component counting to establish the index).

When viewing a toplevel post you don't need to do any scroll repositioning, so this PR checks that and disables it for non-replies.

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -367,7 +367,7 @@ export const PostThread = observer(function PostThread({
       data={posts}
       initialNumToRender={posts.length}
       maintainVisibleContentPosition={
-        isNative && view.isFromCache
+        isNative && view.isFromCache && view.isCachedPostAReply
           ? MAINTAIN_VISIBLE_CONTENT_POSITION
           : undefined
       }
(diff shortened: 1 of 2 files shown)


src/view/com/post-thread/PostThread.tsx
```diff
@@ -367,7 +367,7 @@ export const PostThread = observer(function PostThread({
       data={posts}
       initialNumToRender={posts.length}
       maintainVisibleContentPosition={
-        isNative && view.isFromCache
+        isNative && view.isFromCache && view.isCachedPostAReply
```
gaearon, line 370: Wanna leave a comment mentioning the other case is broken? So that it's clear why the condition

gaearon (approved):

src/view/com/post-thread/PostThread.tsx
pfrazee: The other case isn't broken in the sense that this is a hackfix. This change is just another way to do it correctly -- if the post we're loading isnt a reply, we don't want to reposition.

Issue #1594, opened 2023-10-03
[APP-880] NullPointerException on android thread view


pfrazee:
Got reports that threads werent loading for an android user. Tracked it down to a null pointer exception (sentry REACT-NATIVE-4R8) which is caused by the `maintainScrollPosition` directive. Turns out this was a bug in RN that's fixed in 0.72.5 (https://github.com/facebook/react-native/commit/1a1a79871b2d040764537433b431bc3b416904e3).

If expo functions fine with a premature RN upgrade then we should do that; otherwise need to disable `maintainScrollPosition` in android until this can be fixed.

<sub>[APP-880](https://linear.app/blueskyweb/issue/APP-880/nullpointerexception-on-android-thread-view)</sub>

PR #1599, opened 2023-10-03
Update to react-native@0.72.5

pfrazee:
Closes https://github.com/bluesky-social/social-app/issues/1594 by pulling in the bugfix highlighted below:

CleanShot 2023-10-03 at 16 43 26@2x


patches/react-native+0.72.5.patch
```diff
@@ -0,0 +1,1787 @@
...
+   }
+-  _lastStringStateWasUpdatedWith = _backedTextInputView.attributedText;
+   [self textViewProbablyDidChangeSelection];
+ }
+ 
+diff --git a/node_modules/react-native/ReactCommon/react/renderer/components/rncore/ComponentDescriptors.h b/node_modules/react-native/ReactCommon/react/renderer/components/rncore/ComponentDescriptors.h
+new file mode 100644
+index 0000000..37b7d8d
+--- /dev/null
++++ b/node_modules/react-native/ReactCommon/react/renderer/components/rncore/ComponentDescriptors.h
```
gaearon, line 22: What's up with this patch? Is this intentional? The old patch was way smaller

pfrazee: Oh weird, no. `patch-package` must have bugged out with its process for updating the patch. Glad you caught this, I'll dig into it.

gaearon (approved):
assuming you'll fix the patch

patches/react-native+0.72.5.patch
pfrazee: Fixed

PR #1602, opened 2023-10-04
Onboarding & feed fixes

### LINKED ISSUE #1601: [APP-884] Issues with feed behavior post-onboarding
1. The feed is not automatically refreshed after onboarding, leaving the empty state until the user refreshes. We should trigger that automatically after onboard.
2. It's possible to not follow enough people to fill the feed (esp if they havent been active) leaving the feed under-filled. If the end is reached, we should give a CTA.
3. The following feed is weirdly never hitting the "no more posts" state, instead giving an endless flickering spinner that suggests it's still retrying.

<sub>[APP-884](https://linear.app/blueskyweb/issue/APP-884/issues-with-feed-behavior-post-onboarding)</sub>

pfrazee:
Closes #1601. I started fixing issues and couldn't stop finding new ones, turning this into a bit of a bloated PR.

- Feed is now automatically refreshed after onboarding to ensure the content is loaded
- Following feed now has a CTA if the user reaches the end, which is more possible than you'd think because `getTimeline` only goes back a certain amount.
- Improved detection of the "end of feed" condition (no posts returned)
- Stop autofocusing the search input when going to that tab. This has been a long-standing request but it's important for the empty feed state, thus I'm rolling it into this.
- Fixes to preference management to optimize the behavior a bit and avoid inflight changes clobbering each other
- Small additional fixes
  - Fixed the button to go to feeds on mobile from empty state
  - Fixed layout of the feed empty state on web
  - Fixed the error screen rendering
  - Reduced the default feeds saved for the user to "Discover" now that we have users selecting feeds
  - Fixed some issues with the loading spinner at the footer of the feed
  - Fixed an edgecase where the verify-email reminder could fire during onboarding if the device had been used to create an account before

Here's what the footer CTA looks like:
Simulator Screenshot - iPhone 15 Pro - 2023-10-03 at 18 11 51

### DIFF
--- a/src/view/com/posts/FollowingEndOfFeed.tsx
+++ b/src/view/com/posts/FollowingEndOfFeed.tsx
@@ -1,0 +1,100 @@
+import React from 'react'
+import {StyleSheet, View} from 'react-native'
+import {useNavigation} from '@react-navigation/native'
+import {
+  FontAwesomeIcon,
+  FontAwesomeIconStyle,
+} from '@fortawesome/react-native-fontawesome'
+import {Text} from '../util/text/Text'
+import {Button} from '../util/forms/Button'
+import {NavigationProp} from 'lib/routes/types'
+import {usePalette} from 'lib/hooks/usePalette'
+import {s} from 'lib/styles'
+import {isWeb} from 'platform/detection'
+
+export function FollowingEndOfFeed() {
+  const pal = usePalette('default')
+  const palInverted = usePalette('inverted')
+  const navigation = useNavigation<NavigationProp>()
+
+  const onPressFindAccounts = React.useCallback(() => {
+    if (isWeb) {
+      navigation.navigate('Search', {})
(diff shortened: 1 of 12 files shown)


src/view/com/posts/FollowingEndOfFeed.tsx
```diff
@@ -0,0 +1,100 @@
...
+      navigation.navigate('FeedsTab')
+      navigation.popToTop()
+    }
+  }, [navigation])
+
+  return (
+    <View style={[styles.container, pal.border]}>
+      <View style={styles.inner}>
+        <Text type="xl-medium" style={[s.textCenter, pal.text]}>
+          You've reached the end of your feed! Find some accounts to follow to
```
gaearon, line 42: "to fix this" sounds overly technical to me, maybe just end with "to follow."

gaearon (approved):

PR #1603, opened 2023-10-04
Improve time to content in the search page

pfrazee:
The search page's user discovery algorithm is currently a mix of two things:

1. Suggested follows from the server
2. A client-side crawl of the FOAFs

2 is temporary until we can get some better facilities in the server and it performs very poorly. We've been historically waiting for both to finish to render something. Breaking them up and rendering the server's suggestions first makes things much better.

### Before

https://github.com/bluesky-social/social-app/assets/1270099/e06cbd50-29a8-487d-a586-a4c7ccf1c331

### After

https://github.com/bluesky-social/social-app/assets/1270099/2e23bcbf-0f7e-4975-9d45-352940637291


gaearon (approved):
nice

PR #1606, opened 2023-10-04
Pull upstream bugfixes to bottom-sheet

pfrazee:
Noticed the swipe gestures failing on the bottom-sheet. Pulling latest upstream fixed it.


gaearon:
How to repro the issue?

pfrazee:
Before updating, I found all bottom sheets were failing to swipe away (down motion) and sheets with multiple snap points such as the account switcher would not swipe upward.

gaearon (approved):
i trust you but couldn't repro for some reason

pfrazee:
That's worrisome. I wonder how that can happen -- I was able to repro on device and in a simulator. No harm merging but, very odd.

PR #1612, opened 2023-10-04
Improve typeahead search with inclusion of followed users (temporary solution)

pfrazee:
Uses a client-side cache of the followed users to prioritize the user's follows in autocomplete search (seach page, mentions).

Should replace this with serverside as the fetch of follows has to happen in pages (of 100) and stops at 1k, all of which adds more load than we want. Still, these results are much better

### Before

https://github.com/bluesky-social/social-app/assets/1270099/8abb6798-507d-4c56-a898-766684fa46e6

Notice how much I'm struggling to find ppl I know

### After

https://github.com/bluesky-social/social-app/assets/1270099/3d0a2a2a-15ae-4af8-bfe7-431f6f4274f7

Eat viola, I'm immediately finding my friendos

### DIFF
--- a/src/state/models/content/profile.ts
+++ b/src/state/models/content/profile.ts
@@ -137,7 +137,7 @@ export class ProfileModel {
       runInAction(() => {
         this.followersCount++
         this.viewer.following = res.uri
-        this.rootStore.me.follows.addFollow(this.did, res.uri)
+        this.rootStore.me.follows.hydrate(this.did, this)
       })
       track('Profile:Follow', {
         username: this.handle,
--- a/src/state/models/content/profile.ts
+++ b/src/state/models/content/profile.ts
@@ -290,8 +290,8 @@ export class ProfileModel {
     this.labels = res.data.labels
     if (res.data.viewer) {
       Object.assign(this.viewer, res.data.viewer)
-      this.rootStore.me.follows.hydrate(this.did, res.data.viewer.following)
     }
+    this.rootStore.me.follows.hydrate(this.did, res.data)
   }
 
   async _createRichText() {
--- a/src/state/models/discovery/user-autocomplete.ts
+++ b/src/state/models/discovery/user-autocomplete.ts
@@ -4,11 +4,7 @@ import AwaitLock from 'await-lock'
 import {RootStoreModel} from '../root-store'
 import {isInvalidHandle} from 'lib/strings/handles'
 
-interface Suggestion {
-  handle: string
-  displayName: string | undefined
-  avatar: string | undefined
-}
+type ProfileViewBasic = AppBskyActorDefs.ProfileViewBasic
 
 export class UserAutocompleteModel {
   // state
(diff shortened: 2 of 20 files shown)


src/state/models/discovery/user-autocomplete.ts
```diff
@@ -77,28 +102,40 @@ export class UserAutocompleteModel {
...
-      term: this.prefix,
-      limit: 8,
-    })
-    runInAction(() => {
-      this.searchRes = res.data.actors
-      for (const u of this.searchRes) {
-        this.knownHandles.add(u.handle)
-      }
-    })
+function prefixMatch(prefix: string, info: ProfileViewBasic): boolean {
```
gaearon, line 133: Currently `dho` prefix matches `joshdholtz` but I think it shouldn't. I suggest maybe splitting the `info.handle` and `info.displayName` into "words" by separators (spaces and punctuation), then returning `true` if any of the "words" starts with the given prefix (rather than includes).

gaearon (approved):
looks ok but not familiar with the code overall, lmk if you want closer testing.

have one concern about matching algorithm, commented

src/state/models/discovery/user-autocomplete.ts
pfrazee: I debated this, figured it's better to match too many than too few. You're selecting from the menu anyway, so the odds that our lack of specificity could be a problem seem low -- eventually you'll type `dholms` and that's unlikely to double match. Meanwhile you're more likely to type `holm` for holmgren and if dan has a nonreal display name, then holm wouldnt match on a startswith

Would probably have to test against a big sample set to give a definitive answer but that's my thinking

gaearon: I guess my concern is more that some matches may look "unexplainable" and push down the matches you actually want. Even if you keep typing, that seems unfortunate.

I see your point for handles but maybe not for display names.

src/state/models/content/profile.ts
```diff
@@ -137,7 +137,7 @@ export class ProfileModel {
       runInAction(() => {
         this.followersCount++
         this.viewer.following = res.uri
-        this.rootStore.me.follows.addFollow(this.did, res.uri)
+        this.rootStore.me.follows.hydrate(this.did, this)
```
ansh, line 140: Definitely a better name

src/state/models/discovery/user-autocomplete.ts
pfrazee: Think about how often people do something like this:

`🧛Paul Frazee`

gaearon: Oh interesting, I guess punctuation alone wouldn't be enough. Okay I'd need to see how it feels in practice but point taken

pfrazee: We'll keep an eye on it

PR #1616, opened 2023-10-05
Remove unused lightbox options

gaearon:
This is a purely mechanical refactor of the lightbox code. Since we vendored the code, I simplified the code structure a bit and removed the code that we don't use. This narrows down to the features we actually rely on.

- https://github.com/bluesky-social/social-app/commit/a5218607de62a4b0feaf1b163d659e77749e875c: There's no reason for these to be reusable Hooks. First, let's move them to the only file where they're used.
- https://github.com/bluesky-social/social-app/commit/df839e9229e876df1d8309bece6e86a4e8757630: This file is unused entirely. Delete.
- https://github.com/bluesky-social/social-app/commit/bcfc92800ec763efc2991519ef2d34afede99d57: We don't use the long press gesture. Delete the implementation. (We can readd later if needed.)
- https://github.com/bluesky-social/social-app/commit/bc354bfedcb7a6a8f07ec95a50749667da567674: Double tap is always on. Hardcode the option as `true`.
- https://github.com/bluesky-social/social-app/commit/193629cbda0a597e17cb6546211c06440255a8e5: Swipe to close is always on. Hardcode the option as `true`.
- https://github.com/bluesky-social/social-app/commit/40fcf9a8d7751773c9eec5bca4159e0d44157dea: We don't use this callback. Delete the option and the Effect.
- https://github.com/bluesky-social/social-app/commit/1234f1737561d6f401bfc08c491647dc5ed742f1: Let's inline these custom Hooks directly into the component. There's no reason for them to be abstracted.
- https://github.com/bluesky-social/social-app/pull/1616/commits/0a80312e2dacfcce1c48c1e4575a9d54ec5e4bc6: Move the nested `LightboxFooter` out of `Lightbox` since that's not valid in React. Same logic as before.
- https://github.com/bluesky-social/social-app/pull/1616/commits/5d05e6dfe3187d9a4e540eb615c7172ae610f36f: Inline iOS-specific `useDoubleTapToZoom` into the only component using it.
- https://github.com/bluesky-social/social-app/pull/1616/commits/fec6fdae2bf211d10af9284a98ec300f016c108a: Remove unused utilities, move some to the only place they're used.

Tested on Android and iOS, exact same behavior like before.


pfrazee (approved):
Tested on-device both platforms, no regressions found. Looks good!

PR #1622, opened 2023-10-06
Dont trim before posting (close #1621)

### LINKED ISSUE #1621: [APP-885] Trimming whitespace at the start of posts is throwing off ascii art
See https://bsky.app/profile/lyndseyluvs.bsky.social/post/3kazx2ofnld2w. Need to look at whether any trimming is needed at all. I believe the main goal was to avoid runs of empty space, but that's more of a newline problem than a starting-space problem.

<sub>[APP-885](https://linear.app/blueskyweb/issue/APP-885/trimming-whitespace-at-the-start-of-posts-is-throwing-off-ascii-art)</sub>

pfrazee:
The richtext library runs a sanitization pass that reduces excess newlines. Trimming doesnt really add anything and sometimes breaks people's posts, so this PR stops doing that.

EDIT: tweaked to still trim end, the web composer sticks empty space on the end if we dont

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -94,7 +94,7 @@ export async function post(store: RootStoreModel, opts: PostOpts) {
     | undefined
   let reply
   let rt = new RichText(
-    {text: opts.rawText},
+    {text: opts.rawText.trimEnd()},
     {
       cleanNewlines: true,
     },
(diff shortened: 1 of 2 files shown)


gaearon:
what happens if you prefix with a bunch of newlines? i guess it's no worse than doing a bunch of newlines in the middle?

gaearon:
how do you feel about trimming all initial _newlines_ but allowing spaces on the first line?

gaearon (approved):
with suggestion

pfrazee:
The richtext sanitization pass will reduce any run of newlines (start or end) down to two newlines -- so, presently, that's what'll happen

```



foo
```

will become

```

foo
```

I agree that stripping those empty newlines would be better, I'm just not sure I want to invest the dev time in it.

PR #1624, opened 2023-10-06
Rewrite the Android lightbox (this time for real)

gaearon:
Re-submit of https://github.com/bluesky-social/social-app/pull/1563 which has the original history.

This is a complete rewrite of the Android part of the lightbox with several purposes:

- It fixes https://github.com/bluesky-social/social-app/issues/1485
- It fixes a few existing bugs that I haven't documented yet (will do)
- Makes it match native Android stock app gesture behavior better (e.g. Photos), e.g. limited panning
- Moves it to a more powerful gesture/animation framework so that further improving the lightbox is easier

TODO

- [x] Code
- [ ] Explanatory comments
- [ ] Proper PR description

PR #1627, opened 2023-10-06
Fix stuck lightbox header after double tap

gaearon:
Fixes a regression introduced in https://github.com/bluesky-social/social-app/pull/1624.

It seems like Reanimated has some kind of compiler behavior that _infers_ `useAnimatedReaction` "dependencies" but since it doesn't analyze functions, it doesn't know that my `getCommittedScale` helper reads `committedTransform.value`.

If I understood it correctly, this seems like a pretty big footgun — would be nice to lint against render scope function calls there then? Anyway, I inlined the helper to its usage points and clarified the intention of the code.

### Scenario 1

1. Double tap
2. [Without this fix] the controls don't disappear
3. [With this fix] the controls disappear, as they should

### Scenario 2

1. Pinch in (controls disappear)
2. Double tap out
3. [Without this fix] the controls don't appear
4. [With this fix] the controls appear, as they should


pfrazee (approved):
Interesting, 👍

PR #1640, opened 2023-10-08
fix accessibility in AutoSizedImage

elijah-wright:
setting the alt tag on the parent component (the `Pressable`) won't apply to divs on web. the image should receive the alt tag.

additionally, [VoiceOver Recognition](https://support.apple.com/en-us/HT211899) uses on-device machine learning to detect what an image is depicting, so it's probably not necessary to state that it's an image (which is what the purpose of the role is). I don't think TalkBack has this functionality but I might be wrong there.


pfrazee:
Oh my goodness, correct you were and I can't believe we didn't notice web was dropping this. 🙏

PR #1644, opened 2023-10-09
fix: warn on unicode domain confusion

intrnl:
Remembered about Unicode domains, and realized that JS' URL API will always normalize them to punycode form, which means that it's currently possible to do some [Unicode domain trickery](https://www.xudongz.com/blog/2017/idn-phishing/) and have it pass through the code that checks if a link is suspicious

image

This pull request makes it warn on *all* Unicode domains, and adds the relevant tests for it.

This *does* mean that it would catch Japanese sites using [`.みんな` TLD](https://www.registry.google/tlds/minna/) as an example, where I'd assume that TLD would encourage having your domains be entirely in Unicode, but at the same time, the Unicode usage is concerning, perhaps we could limit the scope down to only Greek/Cyrillic letters where it's known to be actually problematic? Let me know what y'all think.

### DIFF
--- a/__tests__/lib/strings/url-helpers.test.ts
+++ b/__tests__/lib/strings/url-helpers.test.ts
@@ -29,6 +29,7 @@ describe('linkRequiresWarning', () => {
     ['http://site.pages', 'site.pages.dev', true],
     ['http://xn--pple-43d.com', '\u0430pple.com', true],
     ['http://subdomain.xn--pple-43d.com', 'subdomain.\u0430pple.com', true],
+    ['http://notphishing.foo-xn--pple.com', 'notphishing.foo-xn--pple.com', false],
     ['http://xn--s7y.co', '短.co', true],
 
     // bad uri inputs, default to true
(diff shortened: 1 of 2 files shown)


intrnl:
This seems to be how Chrome is approaching the issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1332714#c17

intrnl:
It's a very deep rabbit hole, I'm not sure whether to continue pursuing this just yet, at the very least I'll mark this pull request as draft

intrnl:
fwiw I've looked at Chrome's spoof checks and they're not very trivial to implement, at worst we might just have to give up on this and have the browsers show them properly, or even warn if Safe browsing has flagged them

https://source.chromium.org/chromium/chromium/src/+/main:components/url_formatter/spoof_checks/idn_spoof_checker.cc

intrnl:
Alternative solution:

1. We extract domains out of raw labels manually, check if it contains Unicode, and that it gets converted into punycode `xn--`, if so, show a warning
2. Show a different warning UI, one that shows Unicode and the actual punycode domains
3. Allow users to exclude the domain they're visiting, so they don't get warned on this the next time

It'd still be an inconvenience for non-English users, but this seems like the path to go as it tries to balance the risk of Unicode spoofing, with trying not to inconvenience non-English speakers much

intrnl:
closed, discussion over at #1647

PR #1645, opened 2023-10-09
Refactor iOS lightbox to Reanimated

gaearon:
For consistency with the Android version, let's use Reanimated across the lightbox.
I'll also have a follow-up for the root view though not sure about that one yet.


ansh:
Looks good to me! Works fine

ansh (approved):
👍🏽 Works well

PR #1659, opened 2023-10-10
Drive-by lightbox refactors

gaearon:
No behavior change, just cleaning up before more significant changes.

- https://github.com/bluesky-social/social-app/commit/b274de6757aa86ed4e4102371c45b589fc6a6af1: Removes support for passing a number as image source. This used to be a lie in the types, and we don't use this feature anyway. (In the original library, this was used to support showing local images from disk.) Also removes the `keyExtractor` prop which we're not using anywhere.
- https://github.com/bluesky-social/social-app/commit/2464c8fca8ed84afd820913c82da0b69a30aa56a: Renames the prop to clarify it's only used to seed the initial state.
- https://github.com/bluesky-social/social-app/commit/bd7ef0cec5ed9c0c5b6be8ff69a9aa2ac0b5a513: Renames the state variable.


pfrazee (approved):
LGTM

PR #1662, opened 2023-10-10
Only warn on links to bsky.app if it represents itself as another url

### LINKED ISSUE #1652: Fix/remove the interstitial for internal links
**Describe the bug**
The interstitial added by #1573 warns that we leave Bluesky even if we actually don't, when making a link to bsky.app with a regular text (not a 'false URL')

**To Reproduce**

1. Create a link to an element of bsky.app (post, feed...) but with a regular text (like : "check my thread here")
2. Click on the link

**Expected behavior**

Either change the text of the card link, or remove the card link? I don't know what would be best.

**Screenshots**

image
image

**Details**

 - Platform: Firefox 
 - Platform version: 118.0.1 Windows
 - App version: 1.5.2 I think?

**Additional context**

This feature could also be used to create abuse (like writing insults that lead to someone's account). That might be taken into consideration.

pfrazee:
Closes #1652

Updates the link-warning interstitial to only trigger for `bsky.app` links if the richtext represents itself as another URL. 

CleanShot 2023-10-10 at 12 26 10@2x
CleanShot 2023-10-10 at 12 26 18@2x

### DIFF
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -194,11 +194,7 @@ export function linkRequiresWarning(uri: string, label: string) {
     if (!labelDomain) {
       return true
     }
-    try {
-      return labelDomain !== urip.hostname
-    } catch {
-      return true
-    }
+    return labelDomain !== urip.hostname
   }
 }
 
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -170,14 +170,35 @@ export function getYoutubeVideoId(link: string): string | undefined {
 
 export function linkRequiresWarning(uri: string, label: string) {
   const labelDomain = labelToDomain(label)
-  if (!labelDomain) {
-    return true
-  }
+  let urip
   try {
-    const urip = new URL(uri)
-    return labelDomain !== urip.hostname
+    urip = new URL(uri)
   } catch {
     return true
+  }
+
+  if (urip.hostname === 'bsky.app') {
+    // if this is a link to internal content,
+    // warn if it represents itself as a URL to another app
+    if (
+      labelDomain &&
+      labelDomain !== 'bsky.app' &&
+      isPossiblyAUrl(labelDomain)
+    ) {
+      return true
+    }
+    return false
+  } else {
+    // if this is a link to external content,
+    // warn if the label doesnt match the target
+    if (!labelDomain) {
+      return true
+    }
+    try {
+      return labelDomain !== urip.hostname
+    } catch {
+      return true
+    }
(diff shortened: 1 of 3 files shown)


src/state/models/root-store.ts
```diff
@@ -223,6 +223,7 @@ export class RootStoreModel {
   }
   emitSessionReady() {
     DeviceEventEmitter.emit('session-ready')
+    window.agent = this.agent
```
gaearon, line 226: ?

src/lib/strings/url-helpers.ts
```diff
@@ -170,15 +170,36 @@ export function getYoutubeVideoId(link: string): string | undefined {
...
-    const urip = new URL(uri)
-    return labelDomain !== urip.hostname
+    urip = new URL(uri)
   } catch {
     return true
   }
+
+  if (urip.hostname === 'bsky.app') {
+    // if this is a link to internal content,
+    // warn if it represents itself as a URL to another app
```
gaearon, line 182: what if it's a link to bsky but a different page than claimed?

src/lib/strings/url-helpers.ts
```diff
@@ -170,15 +170,36 @@ export function getYoutubeVideoId(link: string): string | undefined {
...
+    }
+    return false
+  } else {
+    // if this is a link to external content,
+    // warn if the label doesnt match the target
+    if (!labelDomain) {
+      return true
+    }
+    try {
+      return labelDomain !== urip.hostname
```
gaearon, line 198: do you expect `.hostname` accessor to throw? why try/catch?

src/state/models/root-store.ts
pfrazee: of course I forgot to remove that

src/lib/strings/url-helpers.ts
pfrazee: eugh moving too quickly. leftover code, will fix that

src/lib/strings/url-helpers.ts
pfrazee: this is the meat of it here. if you look at the issue that triggered this, people do thinks like "see [this thread]". the question is what exactly do we want to mitigate, and how

a likely issue would be faking mentions, like doing "@bob.xyz" and it actually goes to "@bob.site". right now that's not getting covered. after that it's just misleading labels like "this post by paul frazee" and it's a post by "@pfrazee.io." I'd rather not do an interstitial on *every* link to internal content that's not a raw URL, so I'm not sure what kinds of mitigations to aim for there

gaearon (approved):

Issue #1654, opened 2023-10-10
Typing << or >> In Compose Post Freezes the App


jcsalterego:
**Describe the bug**

Attempting to type `<<` or `>>` in a compose post/reply modal on the web hangs the view.

**To Reproduce**

1. Compose Post or Reply to a new post
2. Attempt to type `<<` or `>>`

**Expected behavior**

One can type `<<` or `>>`

**Screenshots**

N/A

**Details**

 - Platform: Safari, Chrome
 - Platform version:
 - App version: 1.52.0

**Additional context**

Feels like an infinite loop is looping.

jcsalterego:
sry esb

haileyok:
>.>

PR #1663, opened 2023-10-10
resolve fork of zeed-dom

estrattonbailey:
Helps with #1654 

I made a [PR to zeed-dom](https://github.com/holtwick/zeed-dom/pull/9) to fix an infinite loop issue. For now, I've pushed a branch (to my personal github, we may want to change that) that we can install from to use the new version until everything can get updated downstream.

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,8 @@
     "webpack-dev-server": "^4.11.1"
   },
   "resolutions": {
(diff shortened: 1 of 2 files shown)


pfrazee:
👍 merging, we'll go back to their official distribution once they merge our PR

PR #1666, opened 2023-10-10
Change lightbox to use Pager

gaearon:
This removes a bunch of hacks we've done before to make individual image gestures work with the wrapping scroll view.

The problem was that the wrapping scroll view was overly sensitive (e.g. picking up pinches) and there was no good way to control that or to coordinate gestures. Here I'm replacing `ScrollView` (which we used via `VirtualizedList`) with `react-native-pager-view` which we already use in the project.

In my experience testing the lightbox, `react-native-pager-view` feels a lot less flaky on Android and is much better resistant to "spamming" the gallery with swipes and other gestures. Previously the lightbox would sometimes still get "stuck" but I can't get it to do that anymore. It also recognizes pinches more reliably now.

One downside here is that this is also going to use `react-native-pager-view` for iOS, and I think on iOS the behavior with the vanilla scroll view was slightly better. For example, overscroll used to work in the left direction (before the first image) but now it only works at the end. This just seems like a bug in `react-native-pager-view`. I think it's OK to accept that and maybe file a bug for them later cause it's just a small detail. Or we could fork by platform later if we want to.

I've tested on device on Android, but for iOS only on the Simulator.

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -1,9 +1,8 @@
-import React, {MutableRefObject, useState} from 'react'
+import React, {useState} from 'react'
 
 import {ActivityIndicator, Dimensions, StyleSheet} from 'react-native'
 import {Image} from 'expo-image'
 import Animated, {
-  measure,
   runOnJS,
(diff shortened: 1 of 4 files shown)


pfrazee:
Tested on my iPhone, feels solid there. Saw the lack of overscroll on the left edge, not super bothered by it

pfrazee (approved):

PR #1667, opened 2023-10-10
Fix feed switching crash

ansh:
When some custom feeds were offline, it caused a crash on device

CleanShot 2023-10-10 at 13 21 36@2x

CleanShot 2023-10-10 at 13 28 24@2x


pfrazee (approved):
Surprised this was an issue, good to merge

gaearon:
how did this happen? does this mean one of `renderEmptyState` callbacks has a Hook call inside? the linter shouldn't allow that

pfrazee:
@gaearon let's plan to have a call on this, I'm somewhat confused about how we're supposed to handle render-function parameters in the RN components

gaearon:
tbh I really don't understand what this fix is doing

PR #1668, opened 2023-10-10
Fix MobX crash for Android lightbox

gaearon:
Same as https://github.com/bluesky-social/social-app/pull/1666/commits/ed8523c117934a96ef2c2fe7d7ab7534d20bd961 but for Android. Added comments.

Don't want to do this at a higher level to avoid breaking memoization.


pfrazee (approved):
:+1:

PR #1669, opened 2023-10-10
Port remaining lightbox code to Reanimated

gaearon:
This removes the last usage of Animated in the lightbox, replaced by Reanimated.
Should be no observable changes except maybe a slightly different spring formula.

I'll take a pass at making the UX itself nicer after, this just aims to preserve the behavior.

### DIFF
--- a/src/view/com/lightbox/ImageViewing/index.tsx
+++ b/src/view/com/lightbox/ImageViewing/index.tsx
@@ -8,7 +8,7 @@
 // Original code copied and simplified from the link below as the codebase is currently not maintained:
 // https://github.com/jobtoday/react-native-image-viewing
 
-import React, {ComponentType, useMemo, useState} from 'react'
+import React, {ComponentType, useCallback, useMemo, useState} from 'react'
 import {StyleSheet, View, Platform} from 'react-native'
 
 import ImageItem from './components/ImageItem/ImageItem'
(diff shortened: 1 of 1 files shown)


src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -23,19 +24,12 @@ type Props = {
   initialImageIndex: number
   visible: boolean
   onRequestClose: () => void
-  presentationStyle?: ModalProps['presentationStyle']
-  animationType?: ModalProps['animationType']
```
gaearon, line 27: Drive by removal of unused props

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -49,36 +43,24 @@ function ImageViewing({
...
+    transform: [
+      {
+        translateY: withClampedSpring(isScaled ? 300 : 0),
+      },
+    ],
+  }))
 
   const onZoom = (nextIsScaled: boolean) => {
-    toggleBarsVisible(!nextIsScaled)
-    setIsScaled(false)
```
gaearon, line 81: This didn't really make sense and just seems like a mistake in my old code.

pfrazee (approved):
Tested on iPhone, lgtm

PR #1673, opened 2023-10-11
Fix layout shift for multi-image posts

gaearon:
This has been bugging me for a while.

When we encounter a post with a gallery in feed (or profile — anywhere we display posts), we try to render it without images first, then measure the container's width, and then relayout images based on that width. As a result, there is a noticeable layout shift after the first render — even though the actual layout is entirely dependent on its container and doesn't *even* depend on the image sizes.

This shift is very bad both because it's visually noticeable (the screen always paints first with empty images) and because it screws the perf — everything below has to relayout again.

The fix is to use flexbox instead of manual calculations. I don't actually _know_ flexbox so I had some help from ChatGPT with this. I *might* have uncovered a potential RN bug with how the `gap` property works, so there's a bit of a hack in the code. I've checked iOS, Android, and web (Chrome, FF, Safari), and the hack seems to work.

## Before

Notice the layout jump *after* the glimmer. 

https://github.com/bluesky-social/social-app/assets/810438/c227db79-ffd4-45df-84ab-ff476bfc2b34

## After

No layout jump after the glimmer. 

https://github.com/bluesky-social/social-app/assets/810438/58e49a38-a988-4e35-8e6d-dc5e6a0bd2a4

### DIFF
--- a/src/view/com/util/images/Gallery.tsx
+++ b/src/view/com/util/images/Gallery.tsx
@@ -1,5 +1,5 @@
 import {AppBskyEmbedImages} from '@atproto/api'
-import React, {FC} from 'react'
+import React, {ComponentProps, FC} from 'react'
 import {StyleSheet, Text, Pressable, View} from 'react-native'
 import {Image} from 'expo-image'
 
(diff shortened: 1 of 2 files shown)


src/view/com/util/images/Gallery.tsx
```diff
@@ -54,6 +52,13 @@ export const GalleryItem: FC<GalleryItemProps> = ({
 }
 
 const styles = StyleSheet.create({
+  fullWidth: {
+    flex: 1,
+  },
+  image: {
+    flex: 1,
+    borderRadius: 4,
```
gaearon, line 60: do I need `resizeMode: 'cover'` here? realized i dropped it

gaearon: nope, it's default https://reactnative.dev/docs/image-style-props#resizemode

pfrazee (approved):
Great stuff

PR #1676, opened 2023-10-12
Performance optimization 

ansh:
This PR does the following:
- Creates scripts for automated performance testing using Flashlight and Maestro
   - Flashlight only works for Android
   - Maestro allows us to write tests with YAML files instead and avoids the issues that come with using Detox and React Native
- Fixes warnings in `eas.json`
- Removes stray `console.logs` with a plugin in production code to help JS thread (as recommended by RN dos)
- Set up Sentry performance monitoring to get performance insights in production applications
- Moves shell animations to Reanimated and wraps them in a mobx reaction to help avoid re-renders
- Moves shared shell animations to centralized hook
- Moves `LoadLatestBtn` to an animation instead of dynamically changing styles and re-rendering the component
- Moves `FeedPage` out of `Home` and separates concerns of components
- Re-write some logic of Empty screens and renderItem in `Feed` and `FeedPage` so we don't unnecessarily pass components down when not required


docs/testing.md
```diff
@@ -0,0 +1,10 @@
+# Testing instructions
+
+### Using Maestro E2E tests
+1. Install Maestro by following [these instuctions](https://maestro.mobile.dev/). This will help us run the E2E tests.
+
+
+### Using Flashlight for Performance Testing
+1. Make sure Maestro is installed (optional: only for auomated testing)
+2. Install Flashlight by following [these instructions](https://flashlight.dev)
+3.
```
gaearon, line 10: How do these tools relate to Detox?

Do we only use them for perf?

Can you add instructions on actually running these in our repo? E.g. which yarn scripts, how to interpret results

ansh: Yes. I am finishing writing instructions right now.

ansh: Done

ansh: It can be used for other things than perf too. Added all instructions. Please open a new comment if its still confusing.

gaearon (approved):
the Reanimated parts lgtm, but haven't tested

pfrazee:
Load more btn lost it's alignment on desktop web

CleanShot 2023-10-13 at 14 04 02@2x

pfrazee:
I fixed the load more and FAB behaviors on desktop

Last one, on web tablet the feeds pager tabs are off

CleanShot 2023-10-13 at 14 35 22@2x

pfrazee:
Those are the only issues I saw, scroll perf seems definitely stronger on web

PR #1683, opened 2023-10-12
Make shell hide/show animation smoother

gaearon:
This animation was too jumpy. I believe the reason is that we've set the distance pretty high, but the duration pretty low. So even if the device is powerful, each frame it has to travel "too much" resulting in jerkiness. Additionally, things "going offscreen" always feel jarring — usually this is solved by fading things out.

I've changed the animation to travel a smaller distance over a slightly longer time, plus added opacity.

## Before

https://github.com/bluesky-social/social-app/assets/810438/4a74aaa8-fecc-4d7f-a513-4d0c459a218b

## After

https://github.com/bluesky-social/social-app/assets/810438/537f90dc-8bee-4669-995c-bec67afff5c2

### DIFF
--- a/src/lib/hooks/useMinimalShellMode.tsx
+++ b/src/lib/hooks/useMinimalShellMode.tsx
@@ -7,21 +7,22 @@ export function useMinimalShellMode() {
   const store = useStores()
   const minimalShellInterp = useAnimatedValue(0)
   const footerMinimalShellTransform = {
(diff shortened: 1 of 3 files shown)


ansh (approved):
Looks good! #1676 has similar changes but can merge this in the interim.

pfrazee (approved):
👍

EverStarck:
nice 👍

Issue #1265, opened 2023-08-24
Add the option to hide alt text on images


enmaku:
**Is your feature request related to a problem? Please describe.**

Alt text is useful, but frequently covers the entire lower portion of tall images, making their lower portions unreadable.

**Describe the solution you'd like**

A button or swipe gesture to close the alt text panel, and perhaps a small corner button or reversed gesture to bring it back. The option to display or not display alt text by default in settings would also be nice. 

**Describe alternatives you've considered**

Users can kind of get around the issue now by zooming and swiping, but this is not ideal and does not work 100% of the time. 

**Additional context**

None.

pfrazee:
+1 to this or some similar solution

PR #1687, opened 2023-10-12
Toggle lightbox controls on tap

gaearon:
Closes https://github.com/bluesky-social/social-app/issues/1265.

This makes the lightbox controls (including alt text) toggle on tap, similar to the Photos app behavior:

- Initially, controls are visible.
- A single tap toggles visibility (hide -> show, show -> hide).
- Zooming in (via pinch or double tap) immediately hides controls.
- Once controls are hidden, only tap shows them again (zooming out doesn't).

Note that, of course, two quick taps (double tap to zoom) should not toggle controls there and back.

I've also tweaked the animation to feel softer, similar to https://github.com/bluesky-social/social-app/pull/1683.

## Note on reviewing

It seems like sometimes Reanimated scroll view events just stop working in the iOS simulator. I can't quite pinpoint the pattern. This doesn't seem to happen when I build for device in the Release configuration, so I presume it's DEV-only. There also seems to be a weird way to get them to unstuck: I put `console.log()` into `registerForEvents` in `node_modules/react-native-reanimated/src/reanimated2/WorkletEventHandler.ts`. This might be placebo though, I'm not sure if this is a reliable fix. So far this _seems_ to be some kind of DEV-only bug in Reanimated. I'll try to report it to them.

## iOS

https://github.com/bluesky-social/social-app/assets/810438/dded1c68-ce54-40c4-84cf-cce03345a316

## Android

https://github.com/bluesky-social/social-app/assets/810438/1a1c4eea-3bef-4a88-862d-2971f1f1973b

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -34,11 +34,13 @@ const initialTransform = createTransform()
 type Props = {
   imageSrc: ImageSource
   onRequestClose: () => void
+  onTap: () => void
   onZoom: (isZoomed: boolean) => void
   isScrollViewBeingDragged: boolean
 }
 const ImageItem = ({
   imageSrc,
+  onTap,
   onZoom,
   onRequestClose,
   isScrollViewBeingDragged,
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Tested on iOS and android simulators, works great

PR #1691, opened 2023-10-13
Don't re-render bars when showing/hiding them

gaearon:
Currently, when we show/hide bars on scroll, there is a React re-render for both of them. However, that's unnecessary because we only use that React re-render for an Effect which triggers a chance in the Animated value.

Let's just use a MobX reaction here. (In the future, we'll replace this with a Reanimated style so this won't happen anyway.)

### DIFF
--- a/src/lib/hooks/useMinimalShellMode.tsx
+++ b/src/lib/hooks/useMinimalShellMode.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import {reaction} from 'mobx'
 import {useStores} from 'state/index'
 import {Animated} from 'react-native'
 import {useAnimatedValue} from 'lib/hooks/useAnimatedValue'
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useMinimalShellMode.tsx
```diff
@@ -12,22 +13,27 @@ export function useMinimalShellMode() {
...
-      }).start()
-    } else {
-      Animated.timing(minimalShellInterp, {
-        toValue: 0,
-        duration: 150,
-        useNativeDriver: true,
-        isInteraction: false,
-      }).start()
-    }
-  }, [minimalShellInterp, store.shell.minimalShellMode])
```
gaearon, line 30: The fix is that `store.shell.minimalShellMode` isn't read on render anymore, so MobX won't subscribe this component to its changes. We don't need to subscribe the component to its changes because we only care about re-running the imperative code, which `reaction` will do for us.

gaearon:
This feels way faster on lower-end Android, though I only checked in DEV mode — might not be as important in prod.

There is a regression though for initial load though:

Screenshot 2023-10-13 at 04 03 34

gaearon:
Pushed another commit that uses `autorun` instead. This fixes the first render.

estrattonbailey (approved):
Nice!

PR #1693, opened 2023-10-13
Fix profile preview jump

gaearon:
Fixes a bunch of issues that caused the profile preview -> profile transition to be jumpy.

- We now prevent the sheet from being dragged higher than the safe area so that it aligns with the profile screen.
- Fixed the profile preview glimmer — it used to be very misaligned.
- Added an artificial delay during the transition to avoid flicker. (Couldn't find a more reliable way.)

The 200ms delay is consistently enough to avoid flashing the previous screen (fixing an existing bug). It also gives us a chance to load the images — but we should figure out why the images aren't being instantly available in the cache (this is clear if you remove artificial the delay). I wonder if this is also related to them being too large. Left a TODO for this.

## Before

https://github.com/bluesky-social/social-app/assets/810438/117b2403-bc84-4d21-8844-97da4ed5dc7e

## After

https://github.com/bluesky-social/social-app/assets/810438/74b8c710-97a0-4b7b-a990-d91f15a936c0

### DIFF
--- a/src/view/com/modals/Modal.tsx
+++ b/src/view/com/modals/Modal.tsx
@@ -6,6 +6,7 @@ import BottomSheet from '@gorhom/bottom-sheet'
 import {useStores} from 'state/index'
 import {createCustomBackdrop} from '../util/BottomSheetCustomBackdrop'
 import {usePalette} from 'lib/hooks/usePalette'
+import {timeout} from 'lib/async/timeout'
 import {navigate} from '../../../Navigation'
 import once from 'lodash.once'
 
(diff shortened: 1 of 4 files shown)


src/view/com/modals/Modal.tsx
```diff
@@ -36,11 +37,13 @@ import * as SwitchAccountModal from './SwitchAccount'
 import * as LinkWarningModal from './LinkWarning'
 
 const DEFAULT_SNAPPOINTS = ['90%']
+const HANDLE_HEIGHT = 24
```
gaearon, line 40: Same as the default but explicit here so we can use it in the calculation.

gaearon:
Changed from 500ms to 200ms artificial delay. 500ms felt perceptible because you notice the modal isn't being dismissed.

With a 200ms delay, the images don't always load in time (so there can be a flash) but this is a separate problem we need to address. They're supposed to load instantly from cache so we'll need to figure out why they don't.

pfrazee (approved):
Tested on both simulators, works great

PR #1694, opened 2023-10-13
Scope drawer drag workaround to feed pagers

gaearon:
We have a patch that makes the drawer gesture coexist with feed pagers. However, this patch breaks the overdrag in iOS lightbox. We want to enable this patch only for feed pagers. We could've introduced a prop for this, but conceptually, we really just want to disable the hack when overdrag is on (because it breaks overdrag, and feed pagers don't use overdrag).

By default it's `false`. We only turn it on for the lightbox right now. So this fixes the lightbox.

This does that.

## Test plan

- Verified dragging the drawer still works for the feed pager
- Verified overdrag is fixed for the lightbox.

You'd need to `yarn prebuild` to get the new behavior.

### DIFF
--- a/patches/react-native-pager-view+6.1.4.patch
+++ b/patches/react-native-pager-view+6.1.4.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/react-native-pager-view/ios/ReactNativePageView.m b/node_modules/react-native-pager-view/ios/ReactNativePageView.m
-index ab0fc7f..fbbf19f 100644
+index ab0fc7f..1ace752 100644
 --- a/node_modules/react-native-pager-view/ios/ReactNativePageView.m
 +++ b/node_modules/react-native-pager-view/ios/ReactNativePageView.m
 @@ -1,6 +1,6 @@
@@ -19,7 +19,7 @@ index ab0fc7f..fbbf19f 100644
  
  @property(nonatomic, strong) UIPageViewController *reactPageViewController;
  @property(nonatomic, strong) RCTEventDispatcher *eventDispatcher;
-@@ -80,6 +80,10 @@
+@@ -80,6 +80,10 @@ - (void)didMoveToWindow {
          [self setupInitialController];
      }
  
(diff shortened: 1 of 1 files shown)


patches/react-native-pager-view+6.1.4.patch
```diff
@@ -30,13 +30,13 @@ index ab0fc7f..fbbf19f 100644
      if (self.reactViewController.navigationController != nil && self.reactViewController.navigationController.interactivePopGestureRecognizer != nil) {
          [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
      }
-@@ -463,4 +467,21 @@
+@@ -463,4 +467,21 @@ - (NSString *)determineScrollDirection:(UIScrollView *)scrollView {
```
gaearon, line 33: idk why my git decided to add these but ok

patches/react-native-pager-view+6.1.4.patch
```diff
@@ -30,13 +30,13 @@ index ab0fc7f..fbbf19f 100644
...
  - (BOOL)isLtrLayout {
      return [_layoutDirection isEqualToString:@"ltr"];
  }
 +
 +- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
-+    if (otherGestureRecognizer == self.scrollView.panGestureRecognizer) {
++    if (!_overdrag && otherGestureRecognizer == self.scrollView.panGestureRecognizer) {
```
gaearon, line 39: `if (!_overdrag &&` is the actual change here

pfrazee (approved):
Okay seems to work!

troZee:
@gaearon would you mind adding this fix into https://github.com/callstack/react-native-pager-view?

gaearon:
This isn't really a fix tbh, more of a hack for our case. I'm not sure what a proper fix would be since it depends on how your drawer works.

PR #1696, opened 2023-10-13
Synchronize the feed switcher with the feed drag gesture

gaearon:
WIP.

- [x] Scroll the feed switcher on drag
- [x] Interpolate the indicator to current item
- [x] Interpolate color
- [x] Fix flicker
- [x] Fix jump on drag + tap
- [x] Fix no selection on initial render
- [ ] Revert to a simpler model, this is too complicated and doesn't work
- [ ] Fix web
- [ ] What's up with that right margin
- [ ] What should happen if number of feeds changes (relayout, adjust scroll, etc)
- [ ] Types, naming, etc


gaearon:
implemented in https://github.com/bluesky-social/social-app/pull/6868

PR #1707, opened 2023-10-17
Replaced local style with global style

scozdev:


scozdev:
@pfrazee could you please review this?

pfrazee:
Thanks @scozdev but we don't really need to make minor tweaks like this

PR #1713, opened 2023-10-18
Visually align magnifying glass icon in search bar

Tanza3D:
Similar thing with the new post button, just this really improves things.
| Before  | After |
| ------------- | ------------- |
| image | image |

### DIFF
--- a/src/view/shell/desktop/Search.tsx
+++ b/src/view/shell/desktop/Search.tsx
@@ -57,10 +57,12 @@ export const DesktopSearch = observer(function DesktopSearch() {
       <View
         style={[{backgroundColor: pal.colors.backgroundLight}, styles.search]}>
         <View style={[styles.inputContainer]}>
-          <MagnifyingGlassIcon2
-            size={18}
-            style={[pal.textLight, styles.iconWrapper]}
-          />
+          <View style={styles.searchIconWrapper}>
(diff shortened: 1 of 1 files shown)


pfrazee:
Which browser are you using? It doesn't look like that "before" shot for me in chrome or firefox, and it looks too low after the patch

CleanShot 2023-10-30 at 19 21 39@2x

Tanza3D:
I'd guess it might be something to do with fonts? It seems there's alot of different fonts in use depending on which device is being used. Disabling "Segoe UI" and instead opting for Roboto or another font seems to fix the issue in itself - I had a few friends send screenshots from various different web devices and browsers

## Windows
### Chrome(ium)
image
image
### Firefox
image
## macOS
### Chrome(ium)
image
### Firefox
image
### Safari
image
## Linux
### Chrome(ium)
image
### Firefox
image
[pop!_os]
image
(this one has the opposite issue??)

[zorin os]
image


The only thing I can think of to fix this really is having just one font used loaded through google fonts or such, though I feel that may come at the dismay of many users - the fix applied in this commit will make it worse for alot it seems, especially some Linux users (like pop!_os??)

Tanza3D:
any thoughts on this?

mozzius:
Component no longer exists

PR #1717, opened 2023-10-19
Allow back navigation to home on profile header (on empty navigation history)

pdelfan:
Group 1454

### Problem
The back button on profile header doesn't navigate back to home if there's no navigation history. 

### To reproduce
Navigate directly to a profile on a new tab and click the back button on profile header.

### Proposed fix
We can check if we can go back (there's something to go back to), and if not, go home instead. 

A similar behaviour is currently [implemented here.](https://github.com/bluesky-social/social-app/blob/c6e28f88e5e96453005ff6f11194f3ccf091f4b6/src/view/com/post-thread/PostThread.tsx#L171)


pfrazee (approved):
Oh good catch, thank you!

pdelfan:
You're welcome—happy to contribute!

PR #1728, opened 2023-10-20
Use more beautiful title separator

leo:
Bluesky currently uses a [hyphen](https://en.wikipedia.org/wiki/Hyphen) for separating the individual parts of its tab title in the web app:

2023-10-20 at 19 48 03@2x

As mentioned in the intro of the Wikipedia article above, hyphens are typically used for **connecting multiple words**, and therefore aren't an ideal choice for **separating multiple words** in a tab title.

I replaced the hyphen with a [dash](https://en.wikipedia.org/wiki/Dash), and more specifically an [em dash](https://en.wikipedia.org/wiki/Dash#Em_dash), which, in my opinion, is not only more typographically correct, but also looks much better:

2023-10-20 at 19 48 28@2x

Also happy to add an [en dash](https://en.wikipedia.org/wiki/Dash#En_dash) instead, which is still longer than a hyphen, but shorter than a [em dash](https://en.wikipedia.org/wiki/Dash#Em_dash).

### DIFF
--- a/src/lib/strings/headings.ts
+++ b/src/lib/strings/headings.ts
@@ -1,4 +1,4 @@
 export function bskyTitle(page: string, unreadCountLabel?: string) {
   const unreadPrefix = unreadCountLabel ? `(${unreadCountLabel}) ` : ''
-  return `${unreadPrefix}${page} - Bluesky`
+  return `${unreadPrefix}${page} — Bluesky`
 }


ansh:
This looks good! But, does it work for all pages?

subashcs:
Looks Good !

pfrazee (approved):
Sure, looks good

PR #1743, opened 2023-10-25
Update Analytics 

### LINKED ISSUE #1364: Event tracking is inverted in some cases
**Describe the bug**

There are some incorrect tracking events being sent, they're inverted. Not sure if this is a matter of the logic being wrong, a data race, incorrect viewer state, or what.

Liking an individual post:
* `Post:Unlike`

Unliking an individual post:
* `Post:Like`

Liking a post on a feed:
* `FeedItem:PostLike`
* `Post:Unlike`

Unliking a post on a feed:
* `FeedItem:PostUnlike`
* `Post:Like`

There seems to be similar problems with a number of events, but I haven't tested them all:

Following from a profile:
* `ProfileHeader:UnfollowButtonClicked`
* `Profile:Follow` 

Unfollowing from a profile:
* `ProfileHeader:FollowButtonClicked`
* `Profile:Unfollow` 

When clicking Reply (but before replying, maybe intentional?):
* `FeedItem:PostReply`

**To Reproduce**

* Open DevTools and filter HTTP requests for `api.evt.bsky.app`
* Like and Repost things
* Check the event payload for `https://api.evt.bsky.app/t` (track) endpoints events

**Expected behavior**

Tracking events to be emitted that match the action.

**Screenshots**

The result of liking two posts:

image

Aggregate results showing the issue:
image
image

**Details**

 - Platform: Desktop/Chrome/Linux and iOS
 - Platform version:
 - App version: `1.48.0`

ansh:
1. Uses hashed did for analytics instead of hashed email
2. Fixes #1364

### DIFF
--- a/src/lib/analytics/analytics.tsx
+++ b/src/lib/analytics/analytics.tsx
@@ -52,9 +52,7 @@ export function init(store: RootStoreModel) {
     const sess = store.session.currentSession
     if (sess && sess.did) {
       if (sess.email) {
-        const email_hashed = sha256(sess.email)
         const did_hashed = sha256(sess.did)
-        segmentClient.identify(email_hashed, {email_hashed})
         segmentClient.identify(did_hashed, {did_hashed})
         store.log.debug('Ping w/hash')
       } else {
--- a/src/lib/analytics/analytics.tsx
+++ b/src/lib/analytics/analytics.tsx
@@ -50,8 +50,8 @@ export function useAnalytics() {
 export function init(store: RootStoreModel) {
   store.onSessionLoaded(() => {
     const sess = store.session.currentSession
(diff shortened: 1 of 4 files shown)


src/lib/analytics/analytics.tsx
```diff
@@ -52,9 +52,7 @@ export function init(store: RootStoreModel) {
     const sess = store.session.currentSession
     if (sess && sess.did) {
       if (sess.email) {
```
pfrazee, line 54: Still checking for email here and in the other snippet, is that needed?

ansh: Fixed.

pfrazee (approved):
LGTM

PR #1744, opened 2023-10-25
Simplify the avatar component

gaearon:
We render a lot of avatar components. Especially as you navigate around.

All avatar components currently contain these Hooks:

```js
  const store = useStores()
  const {requestCameraAccessIfNeeded} = useCameraPermission()
  const {requestPhotoAccessIfNeeded} = usePhotoLibraryPermission()
  const dropdownItems = useMemo(
    () =>
    // ...
  )
```

This seems like a bunch of unnecessary overhead since these are only used for editable avatars. And as the comment says, avatars are only editable when `onSelectNewAvatar` is passed. This only happens at two callsites.

I separate the avatar component into two: `UserAvatar` and `EditableUserAvatar`. See individual commits.

I don't expect this to have a noticeable perf effect so this is just a drive-by refactor for simplicity. That said we should generally try to avoid adding a ton of logic to super commonly used leaf components.

## Before

Screenshot 2023-10-25 at 04 22 22

## After

Screenshot 2023-10-25 at 04 23 41

## Commits

- https://github.com/bluesky-social/social-app/commit/b538115e7b421f759e36efae042d979bdd8acc13: This creates `EditableUserAvatar` by copypasting `UserAvatar` 1:1.
- 236b9330d8a69335f7b8e112c808885f8789d4ba: We only pass `onSelectNewAvatar` from two places. Make those use our new `EditableUserAvatar`.
- f87bb4e88a376e464b2cdc07a3cd1c939cf645ac: Separate TypeScript definitions for `UserAvatar` and `EditableUserAvatar` props. 
- 8e55fd25d12c39d6be66f36f96f2e9c5c637a3e5: `EditableUserAvatar` always accepts `onSelectNewAvatar`. Make it required. Remove dead code from other branches. This exposes that `moderation` prop is never used in the editable version. Remove its code and move its prop type to the `UserAvatar` type definition to make this clearer.
- https://github.com/bluesky-social/social-app/commit/5689b24ca330ae08c23577c4eea595c2acd145f3: `UserAvatar` now never accepts `onSelectNewAvatar`. Remove it from the type. Prune dead branches. 

## Test Plan

Verified avatars show up. Verified avatar editing UI still shows up and works.

The commits themselves are fairly mechanical.

### DIFF
--- a/src/view/com/modals/CreateOrEditMuteList.tsx
+++ b/src/view/com/modals/CreateOrEditMuteList.tsx
@@ -18,7 +18,7 @@ import {ListModel} from 'state/models/content/list'
 import {s, colors, gradients} from 'lib/styles'
 import {enforceLen} from 'lib/strings/helpers'
 import {compressIfNeeded} from 'lib/media/manip'
-import {UserAvatar} from '../util/UserAvatar'
+import {EditableUserAvatar} from '../util/UserAvatar'
 import {usePalette} from 'lib/hooks/usePalette'
 import {useTheme} from 'lib/ThemeContext'
 import {useAnalytics} from 'lib/analytics/analytics'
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Yeah looks great

PR #1745, opened 2023-10-25
Remove MobX from Link

gaearon:
We should try to keep very commonly used components lightweight. E.g. avoid unnecessary Effects and subscriptions. Especially when they don't really "change" since this influences the init cost.

I don't see a reason for Link components to be MobX observers. They do reference the `store` but only to call `store.shell.openModal` and `store.shell.closeModal` so I don't think they require listening to `store` changes.

## Test plan

Clicked around. Links seem to work.

## Methodology

I set up a small logpoint tracking which observers exist on the page

Screenshot 2023-10-25 at 04 52 57

## Before

Feed has ~160 observers:

Screenshot 2023-10-25 at 05 04 06

## After

Feed has ~100 observers:

Screenshot 2023-10-25 at 05 03 48

### DIFF
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -1,4 +1,4 @@
-import React, {ComponentProps, useMemo} from 'react'
+import React, {ComponentProps, memo, useMemo} from 'react'
 import {
   Linking,
   GestureResponderEvent,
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
100%, no idea why that happened

Issue #1754, opened 2023-10-26
Timestamp Tooltip in Notifications Page obscured


liby:
**Describe the bug**

In the Notifications Page on the Bluesky web interface, when the mouse hovers over a notification's timestamp, a tooltip appears that contains the timestamp. However, this tooltip is obstructed or overlapped by other UI elements, making it difficult to read.

**To Reproduce**

1. Go to the Notifications Page on the Bluesky web interface.
2. Hover the mouse cursor over the timestamp of a notification.
3. Notice that the tooltip containing the timestamp is obscured.

**Expected behavior**

The tooltip containing the timestamp should appear clearly without any obstructions. Either its position should be adjusted or other overlapping UI elements should be moved so that the tooltip can be easily read.

**Screenshots**

<img width="872" alt="image" src="https://github.com/bluesky-social/social-app/assets/38807139/9cd38ee6-758f-4e06-bdc7-50587d982e0b">

**Details**

 - Platform: desktop / Chrome 118.0.5993.117 / macOS 13.5.2 
 - Platform version: Chrome 118.0.5993.117 / macOS 13.5.2 
 - App version: n/a

**Additional context**

thebabydino:
This still seems to be a problem, I happened to notice it today, so I looked into it. There seem to be two parts to it: a wrapper that has `overflow: hidden` in a style attribute and the entire feed item needing a `z-index` on `:hover`. In the video below, I'm disabling that `overflow: hidden` and adding a `z-index` on `:hover`.

https://github.com/user-attachments/assets/89ea2ba8-9d3c-4fd7-afae-389a23d149e6

Alternatively, the tooltip could just be moved above.

Issue #376, opened 2023-04-03
Accessibility: Ensure the screen reader works on all platforms


pfrazee:
We need a framework for accessibility that will apply reliably on all platforms

PR #1755, opened 2023-10-26
Fix Obscured Timestamp Tooltip in Notifications Page

liby:
This PR addresses an issue observed in the Notifications Page of our Bluesky Web Interface, where the timestamp tooltip was obscured when trying to view it. This behavior diminished the overall user experience, as users had trouble reading the timestamp which is crucial to understand the chronological context of the notification.

Fix: https://github.com/bluesky-social/social-app/issues/1754

## Context

The root of this problem is traced back to an inherent issue with React Native `FlatList` component. Upon further investigation, similar concerns have been discussed widely in the React Native community. For reference, the following are related issues:

- [Issue #376 on hossein-zare/react-native-dropdown-picker](https://github.com/hossein-zare/react-native-dropdown-picker/issues/376)
- [Issue #18616 on facebook/react-native](https://github.com/facebook/react-native/issues/18616)
- [Issue #1903 on necolas/react-native-web](https://github.com/necolas/react-native-web/issues/1903)

## Solution

To resolve this, the PR [includes a workaround](https://github.com/hossein-zare/react-native-dropdown-picker/issues/376#issuecomment-1717443831) of applying a dynamic z-index to the `CellRendererComponent` by using `index` value from `props`.

```
CellRendererComponent={({children, index, style, ...props}) => {
    return (
        <View style={[style, {zIndex: -1 * index}]} index={index} {...props}>
            {children}
        </View>
    );
}}
```

image

This change effectively brings the tooltip to the forefront, improving its visibility and thereby enhancing the overall user experience.

### DIFF
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -1,7 +1,14 @@
-import React, {MutableRefObject} from 'react'
+import React, {MutableRefObject, PropsWithChildren} from 'react'
 import {observer} from 'mobx-react-lite'
 import {CenteredView, FlatList} from '../util/Views'
-import {ActivityIndicator, RefreshControl, StyleSheet, View} from 'react-native'
+import {
+  ActivityIndicator,
+  RefreshControl,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+} from 'react-native'
 import {NotificationsFeedModel} from 'state/models/feeds/notifications'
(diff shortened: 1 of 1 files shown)


pfrazee:
This is because RNW applies `z-index: 0` to all views (https://github.com/necolas/react-native-web/issues/689) which I need to wrap my head around a little bit. I'd prefer not to go with this approach unless we have to.

PR #1757, opened 2023-10-26
Filter non-self threads from profile view

### LINKED ISSUE #1166: Profile posts tab shouldn't filter out self-threads

In a social media where short-form text is the way to communicate, it often happens that you would create multiple chained posts, I usually refer this as "self-threading" as it is primarily a thread containing of only your posts.

Bluesky's profile posts tab currently filters out any subsequent posts within a self-thread, and that isn't nice for users who prefer this way of communicating, especially when they're coming from Twitter

It would be nice if self-threads were not filtered out from the posts tab

None yet.

**Additional context**

None.

estrattonbailey:
Fixes #1166

Frontend component to https://github.com/bluesky-social/atproto/pull/1776. This logic walks the reply chain to ensure that all replies (available on this page of results, anyway) are by the `actor` and in reply to a post by the `actor`.

Example below. See the second image "Posts and replies" for the full picture. "Posts" shows ONLY if those replies that are (1) by the author you're viewing and (2) their reply root is a post by the author you're viewing.

Screenshot 2023-10-26 at 3 11 10 PM
Screenshot 2023-10-26 at 3 11 24 PM

### DIFF
--- a/src/lib/api/feed/author.ts
+++ b/src/lib/api/feed/author.ts
@@ -35,11 +35,57 @@ export class AuthorFeedAPI implements FeedAPI {
       this.cursor = res.data.cursor
       return {
         cursor: res.data.cursor,
-        feed: res.data.feed,
+        feed: this._filter(res.data.feed),
       }
     }
     return {
       feed: [],
     }
   }
+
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Works for me!

PR #1758, opened 2023-10-27
[NOT FOR MERGE] Add instrumentation for module init

gaearon:
Would be nice to have some non-intrusive way to check this in under a toggleable flag but for now I'll just hardcode it. This adds instrumentation that logs when a module initializes.

Screenshot 2023-10-27 at 02 06 43

It's interesting that after https://github.com/bluesky-social/social-app/pull/1756, we can see lazy init in action:

https://github.com/bluesky-social/social-app/assets/810438/0d3d8d9d-1af1-43f1-933f-7fccba3e218c


gaearon:
https://github.com/bluesky-social/social-app/pull/1787

PR #1759, opened 2023-10-27
Fix white flash on startup

gaearon:
Two separate problems here.

1. We need to provide `initialMetrics` for the `SafeAreaProvider`. Otherwise the first frame is empty. See [here](https://github.com/th3rdwave/react-native-safe-area-context#optimization).
2. Doing that isn't sufficient though because `NavigationContainer`, [for whatever reason](https://reactnavigation.org/docs/navigation-container/#onready), can't render in a single pass. So wait for `onReady` which, as the docs say, is a good place for "hiding your native splash screen". That's exactly what we're doing.

## Before

https://github.com/bluesky-social/social-app/assets/810438/794e7b57-0f08-4ba8-83fe-0e8b5c630aef

## After

https://github.com/bluesky-social/social-app/assets/810438/302cd600-fb0f-4cdf-988f-f84c18fae31f

## Test Plan

Verified there's no flash on iOS and Android. No behavior changes on web.

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -34,7 +34,6 @@ const App = observer(function AppImpl() {
       setRootStore(store)
       analytics.init(store)
       notifications.init(store)
-      SplashScreen.hideAsync()
       Linking.getInitialURL().then((url: string | null) => {
         if (url) {
           handleLink(url)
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Good stuff

gaearon:
#movefast

PR #1763, opened 2023-10-27
Initialize screen modules lazily

gaearon:
This builds on https://github.com/bluesky-social/social-app/pull/1756.

Normally, inline requires already ensure that components' modules only initialize when they actually get used in JSX. However, this navigation file references all the screens in its JSX even though only few of them are actually rendered at a time.

<s>To solve this, we create one level of indirection. Instead of referencing component functions during render, we wrap them into `React.lazy` that would immediately resolve (so there is no async gap). This gives us lazy module initialization.</s>

I'm a clown, React Navigation [already supports this](https://github.com/bluesky-social/social-app/pull/1763#issuecomment-1783260213) so I just changed to use their `getComponent` API.

As a result, we only load the modules that are used by the current screen.

## Before

Modules for all screens block startup on their initialization:

Screenshot 2023-10-27 at 18 10 44

Note this isn't too bad because normally they don't initilaize transitive dependencies until rendering, but it's still not ideal. In particular, if you *do* use some API at the module init time in a screen, you'd have to pay for it upfront in the startup cost.

## After

Only the initially visible screen gets initialized at startup:

Screenshot 2023-10-27 at 17 59 00

Other screen modules get initialized as you go around the app:

https://github.com/bluesky-social/social-app/assets/810438/afaf4573-1f22-4ca4-b0f1-1d4f40a0a05b

## Test Plan

Walked around the app on iOS, Android, Web. Seems fine.

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -92,42 +92,42 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
     <>
       <Stack.Screen
         name="NotFound"
-        component={NotFoundScreen}
+        component={lazyInit(() => NotFoundScreen)}
         options={{title: title('Not Found')}}
       />
       <Stack.Screen
         name="Moderation"
-        component={ModerationScreen}
+        component={lazyInit(() => ModerationScreen)}
         options={{title: title('Moderation')}}
       />
       <Stack.Screen
         name="ModerationMuteLists"
-        component={ModerationMuteListsScreen}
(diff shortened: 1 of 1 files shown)


ecreeth:
Why don't we use [`getcomponent`](https://reactnavigation.org/docs/screen/#getcomponent)? Is used for lazily get a react component.

gaearon:
Ohhh I've never seen this! Lemme try it

gaearon:
It works! Thanks @ecreeth for the tip :)

pfrazee (approved):
sick

PR #1773, opened 2023-10-31
Don't compile Hermes-supported features with Babel

gaearon:
This simplifies the generated code to be closer to the code we actually write.

Concretely:

- Hermes already supports a bunch of features that we're currently compiling with Babel. Pass an option to target the subset needed by Hermes specifically, and let Hermes handle the rest.
- Although Hermes supports object spread, the Expo preset adds it back. Explicitly patch the Expo preset to disable it.

We will be able to remove both of these overrides when we upgrade Expo:

- The transform profile is being [auto-detected](https://github.com/expo/expo/blob/8156a6ff57353fc50f02537d0b6aefd1033c9f29/packages/babel-preset-expo/src/index.ts#L55) now.
- The object spread plugin is [conditional](https://github.com/expo/expo/blob/8156a6ff57353fc50f02537d0b6aefd1033c9f29/packages/babel-preset-expo/src/index.ts#L65) now.

However, I don't know when we'll get to updating, so why not fix this now.

## Test plan

Walked around the app. App seems to work. Equivalent changes being in newer Expo releases give some confidence.

## Bundle output changes

Glance over the bundle diff.

Function names are now implied (since Hermes can do this):

Screenshot 2023-10-30 at 23 49 46

Native support for `??` and `?.` and friends:

Screenshot 2023-10-30 at 23 50 00

No runtime code needed for rest/spread:

Screenshot 2023-10-30 at 23 49 51

Also, we finally get rid of this `_objectSpread` helper:

Screenshot 2023-10-30 at 23 52 35

Fun fact: this `_objectSpread` helper was duplicated in our bundle almost three thousand times.

Total minified PROD bundle went down from 8.9 MB to 8.6 MB.

Spreads in particular are very common so having them go through fewer indirections is nice.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -8,6 +8,9 @@ module.exports = function (api) {
         {
           lazyImports: true,
           native: {
+            // We should be able to remove this after upgrading Expo
+            // to a version that includes https://github.com/expo/expo/pull/24672.
+            unstable_transformProfile: 'hermes-stable',
             // Disable ESM -> CJS compilation because Metro takes care of it.
             // However, we need it in Jest tests since those run without Metro.
             disableImportExportTransform: !isTestEnv,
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
Ahhh very nice

PR #1774, opened 2023-10-31
Don't inline-require Babel runtime helpers

gaearon:
There's no benefit to treating Babel runtime helpers as lazy. With this change, we go back to eagerly importing them:

Screenshot 2023-10-31 at 00 50 14

I sent a corresponding change upstream in https://github.com/facebook/metro/pull/1127.

If that PR doesn't survive, we can get rid of this patch. It's not super consequential in either direction but I want to remove these indirections from the render path since there's currently a lot of places where Babel generates calls to them.

## Test plan

Walked around the app. It works.


pfrazee (approved):
👍

PR #1776, opened 2023-10-31
Use Sentry only for errors

gaearon:
Sentry has a bunch of fancy tracing features that we're not using.
To keep things simple, let's just disable them.


pfrazee:
@ansh this look okay to you?

ansh:
@pfrazee Yeah this is fine. 

We can use those Sentry features when `sentry-expo` implements them.

ansh (approved):

PR #1787, opened 2023-11-01
Instrument module init in DEV

gaearon:
I end up hacking ad-hoc versions of this into the build all the time so let's just add it as a DEV-only patch. That Metro file barely ever changes so it shouldn't be a burden.

## Test plan

Change `shouldLog` to `true` locally. Observe the output:

Screenshot 2023-11-01 at 13 39 26


ecreeth:
I don't know if this may be helpful for you, but there's a way for ["Investigating the Loaded Modules"](https://reactnative.dev/docs/ram-bundles-inline-requires#investigating-the-loaded-modules)

gaearon:
Yeah that's nice but not helpful enough because I want to see the dependency tree and the actual timings

estrattonbailey (approved):
Sweet!

Issue #718, opened 2023-05-17
Remove Intl polyfill


dholms:
We're doing this with a dependency in `@atproto/api` now (same dep that was used in the polyfill), so yall can stop polyfilling now:

https://github.com/bluesky-social/atproto/pull/1044

PR #1789, opened 2023-11-01
Remove unused Intl.Segmenter polyfill

gaearon:
Closes https://github.com/bluesky-social/social-app/issues/718.

The `graphemer` dep and `Intl` polyfill were originally added in https://github.com/bluesky-social/social-app/pull/362/commits/1b939863e6aef77c2d4e903dffed9c0529d5f94e for composer (https://github.com/bluesky-social/social-app/pull/362/commits/45498937127dcbbb13061f1cdf91134bd47a1331).

However, `@atproto/api` no longer uses `Intl` and instead uses `graphemer` directly: https://github.com/bluesky-social/atproto/pull/1044. So there is no need for these.

## Test Plan

Verified composer works on iOS, Android, and web (Chrome and FF).


src/platform/polyfills.web.ts
```diff
@@ -9,12 +9,8 @@ window.setImmediate = (cb: () => void) => setTimeout(cb, 0)
 
 // @ts-ignore not on the TS signature due to bad support -prf
 if (!globalThis.Intl?.Segmenter) {
-  // NOTE loading as a separate script to reduce main bundle size, as this is only needed in FF -prf
-  const script = document.createElement('script')
-  script.setAttribute('src', '/static/js/intl-segmenter-polyfill.min.js')
-  document.head.appendChild(script)
-
   // loading emoji mart data
+  // TODO: This condition doesn't make sense; emojimart has nothing to do with Intl.
```
gaearon, line 13: Accidentally noticed this. I presume it was added without looking at the condition above. What's happening here is bizarre:

- Emojimart data is undefined in local development (because the condition never triggers) but the emoji picker somehow works. (Is data not required??)
- In production, emojimart data is defined, but it's only because our `scripts.html` partial template adds a `<script>` tag for the thing that triggers emojimart data loading anyway. So this code isn't necessary.

I'd remove it now but I want to understand why it works without data locally. So for now, not touching this.

PR #1794, opened 2023-11-02
Make choose account form scrollable

flicknow:
If you have more than ~10 logins, the Sign in as form fills up the screen and you cannot access "Other account" or "Back". This should be made scrollable


pfrazee:
Thanks!

PR #1798, opened 2023-11-02
Improved list and feed errors

pfrazee:
- Improve the error state of feedgens to appear less frightening
- Fix the error state of profile lists
- Unsave lists when deleted

## Feedgen errors

|||
|-|-|
|CleanShot 2023-11-02 at 15 09 01@2x|CleanShot 2023-11-02 at 15 09 07@2x|

## List error state

CleanShot 2023-11-02 at 13 47 55@2x

### DIFF
--- a/src/state/models/feeds/posts.ts
+++ b/src/state/models/feeds/posts.ts
@@ -25,6 +25,17 @@ import {MergeFeedAPI} from 'lib/api/feed/merge'
 
 const PAGE_SIZE = 30
 
+type FeedType = 'home' | 'following' | 'author' | 'custom' | 'likes' | 'list'
+
+export enum KnownError {
+  FeedgenDoesNotExist,
+  FeedgenMisconfigured,
+  FeedgenBadResponse,
+  FeedgenOffline,
+  FeedgenUnknown,
+  Unknown,
+}
+
 type Options = {
   /**
    * Formats the feed in a flat array with no threading of replies, just
@@ -49,6 +60,7 @@ export class PostsFeedModel {
   isBlocking = false
   isBlockedBy = false
   error = ''
+  knownError: KnownError | undefined
   loadMoreError = ''
   params: QueryParams
   hasMore = true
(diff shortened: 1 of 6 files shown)


src/state/models/feeds/posts.ts
```diff
@@ -25,6 +25,17 @@ import {MergeFeedAPI} from 'lib/api/feed/merge'
...
+type FeedType = 'home' | 'following' | 'author' | 'custom' | 'likes' | 'list'
+
+export enum KnownError {
+  FeedgenDoesNotExist,
+  FeedgenMisconfigured,
+  FeedgenBadResponse,
+  FeedgenOffline,
+  FeedgenUnknown,
+  Unknown,
+}
```
estrattonbailey, line 37: Not a hill I'd die on, but I think we should only use string enums. These get evaluated to numbers, which makes rearranging difficult, etc.

```
export enum KnownError {
  FeedgenDoesNotExist = 'FeedgenDoesNotExist'
  ...
}
```

Works 100% of the time, very clear.

src/view/com/posts/FeedErrorMessage.tsx
```diff
@@ -0,0 +1,119 @@
...
+}: {
+  feed: PostsFeedModel
+  onPressTryAgain: () => void
+}) {
+  if (
+    typeof feed.knownError === 'undefined' ||
+    feed.knownError === KnownError.Unknown
+  ) {
+    return (
+      <ErrorMessage message={feed.error} onPressTryAgain={onPressTryAgain} />
```
estrattonbailey, line 39: Hehe so we fall back to the red error?

estrattonbailey (approved):
LGTM, no blockers

src/view/com/posts/FeedErrorMessage.tsx
pfrazee: Hah ya but only for non-feedgens

PR #1801, opened 2023-11-03
Work around a startup stall caused by expo-image on low-end Android

gaearon:
There seems to be an issue in `expo-image` that blocks the UI thread on startup on low-end Android. As a result, we get stuck on the splash screen for 5 extra seconds if even one of those is rendered. As a temporary fix, let's render the startup-blocking images (own avatars in the drawer and the tabbar) as plain RN `Image`.

This lets us hide the splash ~4.5 seconds earlier. The final time to content seems to only improve by ~2.5s, but that's harder to measure reliably because it's network-bound. (We should optimize the load time too though.)

The proper fix is to get this reported to Expo and see if it can be fixed upstream. Maybe it's something like https://github.com/expo/expo/issues/21921.

(The measurements are taken with https://github.com/bluesky-social/atproto/pull/1790 patched in since it was another cause of the slowdown.)

## Before

Time to end of splash: 8s
Time to content: 13s

https://github.com/bluesky-social/social-app/assets/810438/90e8e111-4269-4130-883b-99c55d2b7ae1

## After

Time to end of splash: ~3.5s
Time to content: ~10.5s

https://github.com/bluesky-social/social-app/assets/810438/798d1fc4-d015-46cd-98da-6d74910a6dda

## Test Plan

Verified both the drawer and the tab bar render my avatar fine on iOS, Android, and web.

### DIFF
--- a/src/view/com/util/UserAvatar.tsx
+++ b/src/view/com/util/UserAvatar.tsx
@@ -1,5 +1,5 @@
 import React, {useMemo} from 'react'
-import {StyleSheet, View} from 'react-native'
+import {Image, StyleSheet, View} from 'react-native'
 import Svg, {Circle, Rect, Path} from 'react-native-svg'
 import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {HighPriorityImage} from 'view/com/util/images/Image'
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Amazing, really great

gaearon:
Filed Expo bug: https://github.com/expo/expo/issues/25201

PR #1809, opened 2023-11-03
ListAddUser modal UX improvements

pfrazee:
- Add loading state
- Improve hitslops
- Improve keyboard persistence behavior (wont annoyingly dismiss now when tapping things)
- Fix rendering issues that would cause weird relayouts (some kind of RN flexbox bug due to padding)


gaearon (approved):
cool

PR #1811, opened 2023-11-04
Fix iOS jump in the new pager

gaearon:
I think maybe this works.

Haven't tested on other screens where we use `Feed`.


gaearon:
let's do https://github.com/bluesky-social/social-app/pull/1825 for now which works around it

PR #1813, opened 2023-11-04
Add new logger

estrattonbailey:
This is a port of the logger I wrote at a previous job for a RN application. Couple of key benefits:
- supports stuff you'd expect like log levels and debug contexts
- helps us report more specific/actionable errors to Sentry
- handles formatting and pretty printing
- will help clear up noise in the console once we port over usages

Peep first commit for actual code. Subsequent commits were fixing TS errors and replacing instances of passing an `Error` class as the second argument. This would really be fine, but we should be consistent.


package.json
```diff
@@ -13,21 +13,21 @@
...
     "start:prod": "expo start --dev-client --no-dev --minify",
     "clean-cache": "rm -rf node_modules/.cache/babel-loader/*",
-    "test": "jest --forceExit --testTimeout=20000 --bail",
-    "test-watch": "jest --watchAll",
-    "test-ci": "jest --ci --forceExit --reporters=default --reporters=jest-junit",
-    "test-coverage": "jest --coverage",
+    "test": "NODE_ENV=test jest --forceExit --testTimeout=20000 --bail",
+    "test-watch": "NODE_ENV=test jest --watchAll",
+    "test-ci": "NODE_ENV=test jest --ci --forceExit --reporters=default --reporters=jest-junit",
+    "test-coverage": "NODE_ENV=test jest --coverage",
```
estrattonbailey, line 19: Making sure logging isn't enabled in test modes

.env.example
```diff
@@ -1 +1,3 @@
 SENTRY_AUTH_TOKEN=
+EXPO_PUBLIC_LOG_LEVEL=debug
```
estrattonbailey, line 2: This env var will need to be added to prod builds, probably set to `info` but up for discussion

package.json
```diff
@@ -80,6 +80,7 @@
     "babel-plugin-transform-remove-console": "^6.9.4",
     "base64-js": "^1.5.1",
     "bcp-47-match": "^2.0.3",
+    "date-fns": "^2.30.0",
```
estrattonbailey, line 83: Do we have something else in here already...

src/env.ts
```diff
@@ -0,0 +1,9 @@
+export const IS_TEST = process.env.NODE_ENV === 'test'
+export const IS_DEV = __DEV__
+export const IS_PROD = !IS_DEV
+export const LOG_DEBUG = process.env.EXPO_PUBLIC_LOG_DEBUG || ''
+export const LOG_LEVEL = (process.env.EXPO_PUBLIC_LOG_LEVEL || 'info') as
+  | 'debug'
+  | 'info'
+  | 'warn'
+  | 'error'
```
estrattonbailey, line 9: I've found a module file to be super helpful with testing since you can mock the variables more easily than globals. Expo automatically replaces `process.env.<var>` calls with the strings.

pfrazee (approved):
Nice! I just reviewed the code, havent tested, but lgtm

src/logger/debugContext.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * *Do not import this directly.* Instead, use the shortcut reference `logger.DebugContext`.
+ *
+ * Add debug contexts here. Although convention typically calls for enums ito
+ * be capitalized, for parity with the `LOG_DEBUG` env var, please use all
+ * lowercase.
+ */
+export const DebugContext = {
+  // e.g. composer: 'composer'
+} as const
```
estrattonbailey, line 10: First ones I'd add are contexts for session model, feeds related (especially polling), analytics, notifications

src/state/models/root-store.ts
```diff
@@ -41,7 +41,7 @@ export type AppInfo = z.infer<typeof appInfo>
 export class RootStoreModel {
   agent: BskyAgent
   appInfo?: AppInfo
-  log = new LogModel()
+  log = logger
```
estrattonbailey, line 44: Just did this for now to show how it's a drop-in replacement

PR #1814, opened 2023-11-04
Remove old logger

estrattonbailey:
Just swaps in the new logger everywhere, should be no functional changes.


src/logger/debugContext.ts
```diff
@@ -7,4 +7,6 @@
  */
 export const DebugContext = {
   // e.g. composer: 'composer'
+  session: 'session',
+  notifications: 'notifications',
```
estrattonbailey, line 11: Added these two. Thinking we actually may want to be able to silence certain debugs also (inverse of what is possible rn). Can layer that on later.

pfrazee (approved):
haha what a PITA for you. lgtm!

PR #1819, opened 2023-11-05
Add async storage interface

estrattonbailey:
Storage interface for selectively updating our existing local storage object using keypaths. Data is persisted using AsyncStorage, but all get/set calls are synchronous for ease of use.

```typescript
import * as storage from '#/storage'

await storage.init() // MUST be called first

// calls are synchronous
storage.get('shell.colorMode') // => 'system'
storage.set('shell.colorMode', 'light')
```


src/storage/__tests__/storage.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  storage.set('shell.colorMode', 'light')
+
+  jest
+    .mocked(AsyncStorage.setItem)
+    .mockImplementationOnce(() => Promise.reject(new Error('test error')))
+
+  await storage.set('shell.colorMode', 'system')
+
+  expect(storage.get('shell.colorMode')).toBe('light')
+})
```
estrattonbailey, line 39: I don't really expect this to happen, and not sure we can recover very well if it does, but added it anyway

src/storage/schema.ts
```diff
@@ -0,0 +1,70 @@
+// import {LabelPreference} from '@atproto/api'
+
+/**
+ * The shape of the object we store in local storage
+ */
+export type Schema = {
+  shell: {
+    colorMode: 'system' | 'light' | 'dark'
+  }
+  // session: {
```
estrattonbailey, line 10: The rest of this file is based on `localStorage.getItem('root')` dump in the browser. Actually pretty slim.

estrattonbailey: Left it commented out bc we should double check this before we start using.

src/storage/index.ts
```diff
@@ -0,0 +1,48 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+import getObj from 'lodash.get'
+import setObj from 'lodash.set'
```
gaearon, line 3: How do we feel about not doing these and always doing a flat namespace? I think we want to avoid situations where we query for nested things from different places without shared parsing logic, but the keypath APIs encourage exactly that.

pfrazee:
LGTM, gotta update the PR description

pfrazee:
I believe this was replaced with https://github.com/bluesky-social/social-app/pull/1830

PR #1821, opened 2023-11-05
Restore autofocus on emoji drop down

eliemichel:
Okey, I know somebody must have explicitly turned this feature off by stating `autoFocus={false}` in the `EmojiPicker`, but as an end user it feels really weird to me not to have this auto-focus (I realized several time I was typing my search query directly in the post's content).

Maybe there is a good reason it was deactivated in one specific scenario? At least on my standard desktop browser I believe autofocus is a nicer UX.


ansh:
This works correctly on Firefox, Chrome, and Safari. @pfrazee I will leave merging up to you as this change auto-focuses the search bar when the emoji picker is opened, which is a UX question

pfrazee (approved):
Eyyy finally got to this.

Yeah actually this works great. Focus returns to the input after selecting. Nothing but upside.

PR #1822, opened 2023-11-06
Internationalization & localization

ansh:
Allow for the app to be consumed in multiple languages and locales

This PR installs Lingui and converts most text to support localization. Check `internationalization.md` for detailed info


docs/internationalization.md
```diff
@@ -0,0 +1,111 @@
...
+6. Enjoy translated app!
+
+### Common pitfalls
+```jsx
+import { msg } from "@lingui/macro";
+import { i18n } from "@lingui/core";
+
+const welcomeMessage = msg`Welcome!`;
+
+// ❌ Bad! This code won't work
```
gaearon, line 70: This (and the one below) are quite confusing. Where is this documented and explained? Is anything going to error if we do this, or does it just silently do the wrong thing?

docs/internationalization.md
```diff
@@ -0,0 +1,111 @@
...
+  const buggyWelcome = useMemo(() => {
+    return i18n._(welcomeMessage);
+  }, [i18n]);
+
+  return <div>{buggyWelcome}</div>;
+}
+
+// ✅ Good! `useMemo` has i18n context in the dependency
+export function Welcome() {
+  const linguiCtx = useLingui();
```
gaearon, line 92: Do I understand correctly that these pitfalls only matter if we allow changing the locale without reloading the app? I guess that's unavoidable on native?

src/App.web.tsx
```diff
@@ -26,6 +29,7 @@ const App = observer(function AppImpl() {
       setRootStore(store)
       analytics.init(store)
     })
+    dynamicActivate(defaultLocale) // async import of locale data
```
gaearon, line 32: This looks concerning to me. What will it be doing while the locale data is loading? Is this blocking? If it's not blocking, how does the UI work in the meantime?

docs/internationalization.md
ansh: Yes, that is correct.

If we use the hook and properly memoize any functions that use data from the hook, then the UI will react to changes in the locale. 

When it's implemented in the whole app, I doubt it'll be be instantaneous as the new locale will need to be loaded, then all components will need to be re-rendered. It'll be like switching from light mode -> dark mode, I think.

docs/internationalization.md
ansh: These pitfalls are just memoization pitfalls. The components won't re-render if the theme changes in the "Bad" examples.

src/App.web.tsx
ansh: You can check the `i18n.ts` code -- it's just a dynamic import

gaearon: I understand it's a dynamic import — what I don't understand is what strings will we be showing until the locale loads?

ansh: We can either show nothing or the default locale (English)

ansh: Let's see if any screen flashing becomes an issue...

gaearon: Things to verify on mobile:

- we shouldn't flash the wrong locale
- we shouldn't introduce any new async delays; all files are already on the disk

Things to verify on web:

- we shouldn't flash the wrong locale
- we shouldn't introduce a waterfall (fetch code -> try to render -> discover we don't have a locale -> download locale). if locale is known at the request time, it should download in parallel with the code.


At FB I _think_ we'd just create a complete JS bundle per locale and force a reload on locale change. I wonder if an approach like this would work for us? 

Also, its docs are not great, but I strongly recommend investigating how https://facebook.github.io/fbt/ handles this.

ansh: Lingui works very similar to FBT from what I can tell. I'll take a deeper look and then we can adjust the method that we use. However, I do think that *not* reloading the app on locale change would be a better user experience.

docs/internationalization.md
```diff
@@ -0,0 +1,113 @@
+# Internationalization
+
+We want the official Bluesky app to be supported in as many languages as possible. If you want to help us translate the app, please reach out to us at info@blueskyweb.xyz
```
pfrazee, line 3: Is that email right? Either way, we need to replace this with the actual instructions once we have them

src/view/com/auth/create/Step3.tsx
```diff
@@ -31,11 +34,11 @@ export const Step3 = observer(function Step3Impl({
           editable
           onChange={model.setHandle}
           // TODO: Add explicit text label
-          accessibilityLabel="User handle"
+          accessibilityLabel={_(msg`User handle`)}
           accessibilityHint="Input your user handle"
         />
         <Text type="lg" style={[pal.text, s.pl5, s.pt10]}>
-          Your full handle will be{' '}
+          <Trans>Your full handle will be </Trans>
```
pfrazee, line 41: This kind of thing is going to cause problems -- we don't want trailing spaces to be in the translations because they won't be reliably preserved, and then we'll get spacing issues

src/view/com/composer/Composer.tsx
```diff
@@ -409,10 +416,10 @@ export const ComposePost = observer(function ComposePost({
                   style={[pal.borderDark, styles.addExtLinkBtn]}
                   onPress={() => onPressAddLinkCard(url)}
                   accessibilityRole="button"
-                  accessibilityLabel="Add link card"
+                  accessibilityLabel={_(msg`Add link card`)}
                   accessibilityHint={`Creates a card with a thumbnail. The card links to ${url}`}>
                   <Text style={pal.text}>
-                    Add link card:{' '}
+                    <Trans>Add link card: </Trans>
```
pfrazee, line 422: Another padding space

src/view/com/modals/AddAppPasswords.tsx
```diff
@@ -119,15 +122,19 @@ export function Component({}: {}) {
...
         ) : (
           <Text type="lg" style={[pal.text]}>
-            <Text type="lg-bold" style={[pal.text]}>
-              Here is your app password.
-            </Text>{' '}
-            Use this to sign into the other app along with your handle.
+            <Trans>
+              <Text type="lg-bold" style={[pal.text]}>
+                Here is your app password.
+              </Text>{' '}
```
pfrazee, line 135: Does this work?

src/view/com/modals/SelfLabel.tsx
```diff
@@ -116,23 +119,26 @@ export const Component = observer(function Component({
...
           ) : (
             <View>
               <Text style={[pal.textLight]}>
                 <Text type="md-bold" style={[pal.textLight]}>
-                  Not Applicable
+                  <Trans>Not Applicable</Trans>
                 </Text>
-                . This warning is only available for posts with media attached.
+                <Trans>
+                  . This warning is only available for posts with media
```
pfrazee, line 139: Let's co-locate the `.` with `Not Applicable`

src/view/com/modals/lang-settings/ConfirmLanguagesButton.tsx
```diff
@@ -28,14 +31,16 @@ export const ConfirmLanguagesButton = ({
...
+        accessibilityLabel={_(msg`Confirm content language settings`)}
         accessibilityHint="">
         <LinearGradient
           colors={[gradients.blueLight.start, gradients.blueLight.end]}
           start={{x: 0, y: 0}}
           end={{x: 1, y: 1}}
           style={[styles.btn]}>
-          <Text style={[s.white, s.bold, s.f18]}>Done{extraText}</Text>
+          <Text style={[s.white, s.bold, s.f18]}>
+            <Trans> Done{extraText}</Trans>
```
pfrazee, line 42: Extra space snuck in

src/view/com/modals/report/InputIssueDetails.tsx
```diff
@@ -35,14 +38,16 @@ export function InputIssueDetails({
         style={[s.mb10, styles.backBtn]}
         onPress={goBack}
         accessibilityRole="button"
-        accessibilityLabel="Add details"
+        accessibilityLabel={_(msg`Add details`)}
         accessibilityHint="Add more details to your report">
         <FontAwesomeIcon size={18} icon="angle-left" style={[pal.link]} />
-        <Text style={[pal.text, s.f18, pal.link]}> Back</Text>
+        <Text style={[pal.text, s.f18, pal.link]}>
+          <Trans> Back</Trans>
```
pfrazee, line 45: Move padding-space out

src/view/com/notifications/FeedItem.tsx
```diff
@@ -232,7 +234,9 @@ export const FeedItem = observer(function FeedItemImpl({
             />
             {authors.length > 1 ? (
               <>
-                <Text style={[pal.text]}> and </Text>
+                <Text style={[pal.text]}>
+                  <Trans> and </Trans>
```
pfrazee, line 238: Move padding space out

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -645,9 +648,11 @@ function ExpandedPostDetails({
       <Text style={pal.textLight}>{niceDate(post.indexedAt)}</Text>
       {needsTranslation && (
         <>
-          <Text style={pal.textLight}> • </Text>
+          <Text style={pal.textLight}>•</Text>
```
pfrazee, line 651: Lost padding space -- need to either preserve it or replace with styles

pfrazee (changes requested):
Whew! Nice work @ansh. Added some fixes. If you can hit those, then we can merge but finish the conversation about how locales are loaded and do a followup PR.

src/view/com/modals/AddAppPasswords.tsx
ansh: Yes! Lingui has support for nested components since `<Trans>` is a macro

docs/internationalization.md
ansh: That is fair. I will just ask people to create a github PR or issue for now instead of the email.

src/view/com/auth/create/Step3.tsx
ansh: Good point. Let me find all these occurrences and fix them.

src/view/com/composer/Composer.tsx
ansh: Done

src/view/com/modals/SelfLabel.tsx
ansh: Done

src/view/com/modals/lang-settings/ConfirmLanguagesButton.tsx
ansh: Fixed

src/view/com/modals/report/InputIssueDetails.tsx
ansh: Fixed

src/view/com/notifications/FeedItem.tsx
ansh: Done

src/view/com/post-thread/PostThreadItem.tsx
ansh: Done

src/App.web.tsx
ansh: On mobile, everything will be bundled automatically and then switch when we want it to. I expect our locale `messages.js` files to be around `20kb`. Right now its `14kb` for English.

On web, it will be fetched when required. We can cache it and that will definitely help.  Here is how it works on web: https://lingui.dev/guides/dynamic-loading-catalogs#conclusion

On RN, we would have to implement our own solution. There has been some prior art on this: https://github.com/lingui/js-lingui/issues/503

pfrazee (approved):
Great work!

src/view/com/modals/AddAppPasswords.tsx
pfrazee: Nice nice

PR #1823, opened 2023-11-06
Fix removal of old lists from saved feeds

estrattonbailey:
Noticed that the remove button did not work for user lists. This updates the `unsave` handling to include `list` type sources. It also ensures that the unpin functionality in this view works the same as the unpin handling on `ProfileList` screen. Added TODO comments in both places.

Screenshot 2023-11-06 at 8 32 14 AM

### DIFF
--- a/src/state/models/ui/saved-feeds.ts
+++ b/src/state/models/ui/saved-feeds.ts
@@ -38,12 +38,18 @@ export class SavedFeedsModel {
     return this.hasLoaded && !this.hasContent
   }
 
-  get pinned() {
-    return this.all.filter(feed => feed.isPinned)
+  get pinned(): FeedSourceModel[] {
+    return this.rootStore.preferences.savedFeeds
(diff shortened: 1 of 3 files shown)


src/state/models/ui/saved-feeds.ts
```diff
@@ -38,12 +38,18 @@ export class SavedFeedsModel {
     return this.hasLoaded && !this.hasContent
   }
 
-  get pinned() {
-    return this.all.filter(feed => feed.isPinned)
+  get pinned(): FeedSourceModel[] {
+    return this.rootStore.preferences.savedFeeds
+      .filter(feed => this.rootStore.preferences.isPinnedFeed(feed))
+      .map(uri => this.all.find(f => f.uri === uri))
+      .filter(Boolean) as FeedSourceModel[]
```
estrattonbailey, line 45: Use `preferences` model as source of truth, but hydrate using already loaded data on this model.

pfrazee (approved):

PR #1824, opened 2023-11-06
Extract shell state into separate context

estrattonbailey:
Pulls out drawer handling and minimal shell mode into its own React context provider and methods.

Couple notes:
- closing currently active elements will be easier once they're on on React context — for now I'll manually co-locate these calls to replicate existing functionality


src/state/shell/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [isDrawerSwipeDisabled, _setIsDrawerSwipeDisabled] = React.useState(
+    defaultContextValue.isDrawerSwipeDisabled,
+  )
+  const value = {
+    minimalShellMode,
+    setMinimalShellMode: _setMinimalShellMode,
+    isDrawerOpen,
+    setIsDrawerOpen: _setIsDrawerOpen,
+    isDrawerSwipeDisabled,
+    setIsDrawerSwipeDisabled: _setIsDrawerSwipeDisabled,
```
estrattonbailey, line 39: Could create callbacks for each of these too

src/state/shell/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  return <context.Provider value={value}>{children}</context.Provider>
+}
+
+export function useShellState() {
+  return React.useContext(context)
+}
```
estrattonbailey, line 47: Also considered putting methods here instead of on context. Might be a good move? Then leave context as just the props themselves with no way to update them 🤔

src/view/com/feeds/FeedPage.tsx
```diff
@@ -38,7 +38,7 @@ export const FeedPage = observer(function FeedPageImpl({
   const store = useStores()
   const pal = usePalette('default')
   const {isDesktop} = useWebMediaQueries()
-  const [onMainScroll, isScrolledDown, resetMainScroll] = useOnMainScroll(store)
+  const [onMainScroll, isScrolledDown, resetMainScroll] = useOnMainScroll()
```
estrattonbailey, line 41: No longer need to pass in the root store

src/view/shell/index.tsx
```diff
@@ -38,35 +41,39 @@ const ShellInner = observer(function ShellInnerImpl() {
...
+    if (isAndroid) {
+      listener = BackHandler.addEventListener('hardwareBackPress', () => {
+        setIsDrawerOpen(false)
+        return store.shell.closeAnyActiveElement()
+      })
+    }
     return () => {
-      listener()
+      listener.remove()
     }
```
estrattonbailey, line 62: Needed to use `useShellState`, so refactored this logic up a level into the component 👍

src/view/shell/index.web.tsx
```diff
@@ -17,18 +17,21 @@ import {BottomBarWeb} from './bottom-bar/BottomBarWeb'
...
   const store = useStores()
+  const {isDrawerOpen, setIsDrawerOpen} = useShellState()
   const {isDesktop, isMobile} = useWebMediaQueries()
   const navigator = useNavigation<NavigationProp>()
   useAuxClick()
 
   useEffect(() => {
     navigator.addListener('state', () => {
+      setIsDrawerOpen(false)
       store.shell.closeAnyActiveElement()
```
estrattonbailey, line 32: Need to do this here instead of in a centralized `closeAnyActiveElement`. We will migrate this later on.

pfrazee (approved):
Yeah looks great

src/state/shell/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [minimalShellMode, _setMinimalShellMode] = React.useState(
+    defaultContextValue.minimalShellMode,
+  )
+  const [isDrawerOpen, _setIsDrawerOpen] = React.useState(
+    defaultContextValue.isDrawerOpen,
+  )
+  const [isDrawerSwipeDisabled, _setIsDrawerSwipeDisabled] = React.useState(
+    defaultContextValue.isDrawerSwipeDisabled,
+  )
+  const value = {
```
gaearon, line 33: This should have `useMemo` in case we add more providers above.

src/lib/hooks/useMinimalShellMode.tsx
```diff
@@ -38,7 +39,7 @@ export function useMinimalShellMode() {
 
   React.useEffect(() => {
     return autorun(() => {
```
gaearon, line 41: The `autorun` here doesn't make sense anymore.

src/lib/hooks/useMinimalShellMode.tsx
```diff
@@ -50,7 +51,7 @@ export function useMinimalShellMode() {
         })
       }
     })
-  }, [minimalShellInterp, store.shell.minimalShellMode])
+  }, [minimalShellInterp, minimalShellMode])
```
gaearon, line 54: This seems like a regression — we're triggering re-render when the shell mode changes. We really shouldn't. However, it's not a new regression. It looks like we regressed on this in https://github.com/bluesky-social/social-app/commit/8e9cf182c2e247203b6b5ea9ae701c039945d6a0#r131878136. The original fix was https://github.com/bluesky-social/social-app/pull/1691. We'd need a different fix for this, but since it's already broken, I guess it doesn't block this PR.

src/view/com/auth/LoggedOut.tsx
```diff
@@ -19,15 +20,16 @@ enum ScreenState {
 export const LoggedOut = observer(function LoggedOutImpl() {
   const pal = usePalette('default')
   const store = useStores()
+  const {setMinimalShellMode} = useShellState()
```
gaearon, line 23: OK, so this pattern is a problem. We only need a setter but we're listening to the context which changes (and will trigger re-renders). So this and other components that only want to send messages like `setMinimalShellMode` are now subscribing to the actual shell mode changes. 

I propose that instead we separate two contexts and two values:

- `useShellState` contains UI stuff that changes (like `isDrawerOpen` and `minimalShellMode`)
- `useShellDispatch` exposes a stable `dispatch` method that accepts commands like `'openDrawer'`, `'closeDrawer'`, `setShellMode'`

I think you can implement this as a single `useReducer` instead of multiple state variables, and do

```
const [state, dispatch] = useReducer(shellReducer)

return (
  <ShellStateContext.Provider value={state}>
    <ShellDispatchContext.Provider value={dispatch}>
      ...
```

gaearon (changes requested):
See the last comment — we need to separate setters from the state. Two contexts.

src/lib/hooks/useMinimalShellMode.tsx
pfrazee: Let's try to fix it while we're here. How can we react to the state change without passing in `minimalShellMode` as the dep?

gaearon: I think we’d need to do an explicit subscription — ie hold an array of listeners, and invoke them instead of having it be a state variable. I’d prefer to decouple that though since it’s a bit more involved.

gaearon: Tbh the actual thing i wanted to do is just to make it a “shared value” from Reanimated. So no setState — maybe a set function that toggles it. We’d do this for all gesture/animation-driving things.

gaearon: We should also change it from a boolean to a float and have it be driven by scroll so that it scrolls away and back “with” the content. Don’t have to do now ofc

pfrazee: Okay sg

src/state/shell/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  useIsDrawerSwipeDisabled,
+  useSetDrawerSwipeDisabled,
+} from './drawer-swipe-disabled'
+export {useMinimalShellMode, useSetMinimalShellMode} from './minimal-mode'
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  return (
+    <DrawerOpenProvider>
+      <DrawerSwipableProvider>
+        <MinamalModeProvider>{children}</MinamalModeProvider>
```
gaearon, line 17: typo: Minimal

gaearon (approved):
lets go

PR #1826, opened 2023-11-06
Fix accidental unwrapped strings

gaearon:
Noticed this on the list page on the web:

Screenshot 2023-11-06 at 22 34 21

The issue is a pattern like `foo.error && <Bla />` where `foo.error` can be `0` or an empty string.

If it's an empty string, React no longer throws (it did in the past), so we actually "get away" with it. But `0` could still be an issue in theory, plus we don't want to see warning messages, so let's just write this explicitly.

### DIFF
--- a/src/view/com/composer/ExternalEmbed.tsx
+++ b/src/view/com/composer/ExternalEmbed.tsx
@@ -51,14 +51,14 @@ export const ExternalEmbed = ({
             {link.meta.description}
           </Text>
         )}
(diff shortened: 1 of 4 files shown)


pfrazee (approved):

KingAmo:
excuse me, but what's wrong with `!!foo?.error && <Bla />` ?

gaearon:
it’s fine, i just did a search for all `error &&` and converted it for consistency in the style with other places

PR #1827, opened 2023-11-07
Pull animated scroll handler down from pager

gaearon:
This works around https://github.com/software-mansion/react-native-reanimated/issues/5345.

We don't currently have a codepath surfacing this bug, but if I try to port Profile to use the new pager, the last tab loses its scroll events. The fix is to avoid reusing the same "animated scroll handler" between the pages.

Initially I tried to pull it all the way down to the flat lists. But this change ended up affecting everything in between and turned into a mess. So this is a more scoped fix that "wraps" `onScroll` into a handler as soon as we have a notion of a page.

Types should help us distinguish the two events. (I considered naming it `onScrollWorklet` but that didn't seem worth it.)

### DIFF
--- a/src/view/screens/ProfileFeed.tsx
+++ b/src/view/screens/ProfileFeed.tsx
@@ -1,7 +1,14 @@
 import React, {useMemo, useCallback} from 'react'
-import {FlatList, StyleSheet, View, ActivityIndicator} from 'react-native'
+import {
+  FlatList,
+  NativeScrollEvent,
+  StyleSheet,
+  View,
+  ActivityIndicator,
+} from 'react-native'
 import {NativeStackScreenProps} from '@react-navigation/native-stack'
 import {useNavigation} from '@react-navigation/native'
+import {useAnimatedScrollHandler} from 'react-native-reanimated'
 import {usePalette} from 'lib/hooks/usePalette'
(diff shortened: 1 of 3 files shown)


src/view/screens/ProfileFeed.tsx
```diff
@@ -345,17 +351,14 @@ export const ProfileFeedScreenInner = observer(
             />
           )}
           {({onScroll, headerHeight}) => (
-            <ScrollView
```
gaearon, line 348: I moved the scroll view inside.

src/view/screens/ProfileFeed.tsx
```diff
@@ -430,82 +434,93 @@ const AboutSection = observer(function AboutPageImpl({
...
-            text={`Liked by ${feedInfo.likeCount} ${pluralize(
-              feedInfo.likeCount,
-              'user',
-            )}`}
-            style={[pal.textLight, s.semiBold]}
+    <ScrollView
+      scrollEventThrottle={1}
+      contentContainerStyle={{paddingTop: headerHeight}}
+      onScroll={scrollHandler}>
+      <View
```
gaearon, line 460: The rest is indentation changes.

pfrazee (approved):
Good deal. Small chore but probably the best solution

Issue #1817, opened 2023-11-05
Grammatical Error in Log Message: 'me' should be 'my'


8ortiz4:
**Describe the Bug**

I've identified a grammatical inconsistency in one of the log messages within the codebase. Specifically **src/state/models/me.ts**, in the line of code:

`logger.debug('Updating me profile information')`

Alternatively, this grammatical inconsistency can also be observed through the following URL: https://bsky.app/static/js/main.1b8ff191.js. By specifying the following line of code:

`this.rootStore.log.debug("Updating me profile information")`

**Steps to Reproduce**

1. Access the following URL: https://bsky.app/sys/log
2. Observe the log entry that includes the phrase 'Updating me profile information'
3. Alternatively, through the following URL: https://bsky.app/static/js/main.1b8ff191.js

**Expected Behavior**

The log message **should display** 'Updating my profile information' to maintain grammatical accuracy.

**Actual Behavior**

The current log message exhibits 'Updating me profile information' which **deviates from proper grammar**.

**Screenshot**

<img width="600" alt="Captura de pantalla 2023-11-05 a las 12 17 15" src="https://github.com/bluesky-social/social-app/assets/136113382/178a7817-0650-42a0-a3cd-8c7d70fff2b7">

gaearon:
https://github.com/bluesky-social/social-app/pull/1828#issuecomment-1798117199

PR #1828, opened 2023-11-07
Fix grammatical error

kaichii:
Issue https://github.com/bluesky-social/social-app/issues/1817


gaearon:
it’s intentional, the model is called “me”. the logs are internal and not meant to be read by users.

kaichii:
> it’s intentional, the model is called “me”. the logs are internal and not meant to be read by users.

Make sense.

PR #1830, opened 2023-11-07
Add persistent state provider

estrattonbailey:
Creates a new state provider called "persisted" (name is not final) that manages ALL state that is persisted to to disk in React context, just like any other of our upcoming state providers. The key difference is that after a state update, we write the state object to disk.

Additionally, this implementation uses `BroadcastChannel` to alert other tabs to state updates, and handles refreshing the in memory persisted state representation when a refresh event is received.

I migrated `colorMode` from shell as an example.


src/state/persisted/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // should move to server?
+  invitedUsers: {
+    seenDids: string[]
+    copiedInvites: string[]
+  }
+
+  onboarding: {
+    step: string
+  }
```
estrattonbailey, line 36: The other properties felt OK to flatten, but these two I decided to leave as objects in case we need to expand their child props.

src/state/persisted/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+  colorMode: 'system',
+  accounts: [],
+  currentAccount: undefined,
+  lastEmailConfirmReminder: undefined,
+  primaryLanguage: deviceLocales[0] || 'en',
+  contentLanguages: deviceLocales || [],
+  postLanguage: deviceLocales[0] || 'en',
+  postLanguageHistory: (deviceLocales || [])
+    .concat(['en', 'ja', 'pt', 'de'])
+    .slice(0, 6),
```
estrattonbailey, line 49: This handling is copied from the preferences model

src/state/persisted/store.ts
```diff
@@ -0,0 +1,14 @@
...
+const BSKY_STORAGE = 'BSKY_STORAGE'
+
+export async function write(data: Schema) {
+  await AsyncStorage.setItem(BSKY_STORAGE, JSON.stringify(data))
+}
+
+export async function read(): Promise<Schema | undefined> {
+  const rawData = await AsyncStorage.getItem(BSKY_STORAGE)
+  return rawData ? JSON.parse(rawData) : undefined
+}
```
estrattonbailey, line 14: I wanted to re-implement these so that they DO throw if `AsyncStorage` is unavailable, and I wanted to abstract writing to this key.

pfrazee (approved):
👍 LGTM, let's also get @gaearon's sign-off

Dan- the reason we moved to a hook was mainly so that we can react to state changes, and in particular because on Web the state-changes can come from other tabs (thus the broadcastchannel).

estrattonbailey:
I'm keen on the reducer pattern @gaearon suggested too, looking at that now

gaearon (approved):
let's try this

estrattonbailey (commented):
👍

pfrazee (approved):

PR #1833, opened 2023-11-07
Add modal state provider, replace usage except methods

estrattonbailey:


src/state/modals/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+      setActiveModals(activeModals => [...activeModals, modal])
+      setIsModalActive(true)
+    },
+    [setIsModalActive, setActiveModals],
+  )
+
+  const closeModal = React.useCallback(() => {
+    let totalActiveModals = 0
+    setActiveModals(activeModals => {
+      activeModals.pop()
```
gaearon, line 237: This is mutative. We shouldn't mutate inside updaters, which need to be pure functions.

We should enable Strict Mode to catch these better.

src/state/modals/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+const ModalControlContext = React.createContext<{
+  openModal: (modal: Modal) => void
+  closeModal: () => void
+}>({
+  openModal: () => {},
+  closeModal: () => {},
+})
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const [isModalActive, setIsModalActive] = React.useState(false)
```
gaearon, line 222: I think maybe we can replace this with `const isModalActive = activeModals.length > 0`?

src/state/modals/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+      activeModals.pop()
+      totalActiveModals = activeModals.length
+      return activeModals
+    })
+    setIsModalActive(totalActiveModals > 0)
+  }, [setIsModalActive, setActiveModals])
+
+  return (
+    <ModalContext.Provider
+      value={{
```
gaearon, line 246: Objects we pass to providers should be memoized, let's useMemo them.

src/state/modals/index.tsx
```diff
@@ -0,0 +1,267 @@
...
+        isModalActive,
+        activeModals,
+      }}>
+      <ModalControlContext.Provider
+        value={{
```
gaearon, line 251: Same.

src/state/models/ui/shell.ts
```diff
@@ -319,10 +118,11 @@ export class ShellUiModel {
       this.closeLightbox()
       return true
     }
-    if (this.isModalActive) {
-      this.closeModal()
-      return true
-    }
+    // if (this.isModalActive) {
+    //   // TODO
```
gaearon, line 122: In this PR or later?

gaearon (approved):
nice. accept with small changes

src/state/modals/index.tsx
estrattonbailey: Good call!

src/state/models/ui/shell.ts
estrattonbailey: In this PR, just did it last

src/state/modals/index.tsx
estrattonbailey: Yeah good shout, this should be improved at some point. We should also do the portaling thing we talked about.

This is just a direct port of the existing functionality, so I'd like to just leave as is until we can come back and do it right.

src/state/modals/index.tsx
estrattonbailey: Sounds like a good idea. Oddly, without it, setting `activeModals` to an empty array doesn't trigger a re-render? Might be a shallow compare issue between arrays, idk. I'm sure we could figure this out but it works right now and the code is clear

src/lib/media/picker.web.tsx
```diff
@@ -14,12 +15,12 @@ export async function openCamera(
...
 
 export async function openCropper(
-  store: RootStoreModel,
+  _store: RootStoreModel,
   opts: CropperOptions,
 ): Promise<RNImage> {
   // TODO handle more opts
   return new Promise((resolve, reject) => {
-    store.shell.openModal({
+    unstable__openModal({
```
estrattonbailey, line 23: Without this hacky global I'd need to pass through 2 mobx models and into this file 🤔

src/state/modals/index.tsx
pfrazee: We can make this a non-mutation now though right? Not _terribly_ hard

pfrazee (approved):
LGTM, I think any further changes can be in followups

PR #1835, opened 2023-11-07
Move onboarding state to new persistence + reducer context

pfrazee:
Refactors onboarding state away from the mobx model and into reducer contexts which use the new persistence layer


src/state/models/ui/create-account.ts
```diff
@@ -9,6 +9,7 @@ import {cleanError} from 'lib/strings/errors'
 import {getAge} from 'lib/strings/time'
 import {track} from 'lib/analytics/analytics'
 import {logger} from '#/logger'
+import {DispatchContext as OnboardingDispatchContext} from '#/state/shell/onboarding'
```
estrattonbailey, line 12: Appreciate the rename 🙌

src/state/persisted/schema.ts
```diff
@@ -7,9 +7,9 @@ const accountSchema = z.object({
   did: z.string(),
   refreshJwt: z.string().optional(),
   accessJwt: z.string().optional(),
-  handle: z.string(),
-  displayName: z.string(),
-  aviUrl: z.string(),
+  handle: z.string().optional(),
+  displayName: z.string().optional(),
+  aviUrl: z.string().optional(),
```
estrattonbailey, line 12: If our plan is to fetch this data to render screens that need it, should we just remove these entirely and only store the bare minimum in storage?

src/state/shell/onboarding.tsx
estrattonbailey, line 1: When we update onboarding, we should be able to localize this state a little more since the rest of the app doesn't really need to be aware of onboarding state, aside from the kinda boolean "is onboarding complete/active or not" we may need elsewhere

estrattonbailey (approved):
Dig this pattern

src/state/persisted/schema.ts
pfrazee: that might be better yeah

PR #1839, opened 2023-11-08
Move require alt-text to new persistence + context

pfrazee:


src/state/shell/alt-text-required.tsx
estrattonbailey, line 1: Should this be shell state or preferences?

estrattonbailey (approved):
LGTM but question on file location — can happen later too nbd

src/state/shell/alt-text-required.tsx
pfrazee: Totally preferences - I just put it here for convenience since the other waiting PR adds preferences

PR #1840, opened 2023-11-08
Move invite-state to new persistence + context and replace the invite-use notifications

pfrazee:
Moves local invite-state to a context/hook and the new persistence layer. Also replaces the janky fake notification of an invite usage with just showing who used the invites in the modal:

CleanShot 2023-11-07 at 17 30 24@2x

I swapped that out to simplify this PR and because that fake notification caused way more trouble than it was worth


estrattonbailey (approved):
Genius, visionary

PR #1845, opened 2023-11-08
State layer cleanup: move alt-text-required into preferences and fix a bug in reminders

pfrazee:
- Alt-text-required was temporarily in shell state to avoid conflicts. This PR moves it into preferences.
- The reminders and onboarding refactors happened in parallel PRs. Reminders depends on onboarding and this PR updates it to use that new system.


estrattonbailey (approved):
Beaut

PR #1849, opened 2023-11-09
Hide/show header and footer without re-renders, take two

gaearon:
Re-doing the fix from https://github.com/bluesky-social/social-app/pull/1691 which regressed in https://github.com/bluesky-social/social-app/pull/1676.

This time it's easier to fix at the source. Now that we've moved the shell state into React, I've changed the source of truth to be an Animated value. We can change it directly and make all animations derived from it.

In the future, we should also change this to update with scroll but for now I left it discrete.

## Verification

```diff
// ReactNativeRenderer-dev.js
+let RENDERS = []
    
+setInterval(() => {
+  console.log('new renders:', RENDERS.join(', '))
+  RENDERS = []
+}, 1000)
    
function renderWithHooks(
  current,
  workInProgress,
  Component,
  props,
  secondArg,
  nextRenderLanes
) {
+  RENDERS.push(Component.name)
```

### Before

https://github.com/bluesky-social/social-app/assets/810438/08fd683e-f3cd-4769-814e-9dc6911924a8

### After

https://github.com/bluesky-social/social-app/assets/810438/d18fb4bb-b37d-4e22-b3c7-fa85591cb88d

## Impact

Low/mid-range Android.

### Before

https://github.com/bluesky-social/social-app/assets/810438/be57d152-d1e8-4bc3-9a2d-b5161c2ac63d

### After

https://github.com/bluesky-social/social-app/assets/810438/1a74a3bf-b871-40d6-9eb0-e30a818268b0

## Test Plan

Verified these elements hide/show as expected:

- Top bar
- Bottom bar
- FAB
- Load latest
- Sticky "reply to thread"

### DIFF
--- a/src/state/shell/minimal-mode.tsx
+++ b/src/state/shell/minimal-mode.tsx
@@ -1,16 +1,28 @@
 import React from 'react'
+import {useSharedValue, SharedValue} from 'react-native-reanimated'
 
-type StateContext = boolean
+type StateContext = SharedValue<boolean>
 type SetContext = (v: boolean) => void
 
-const stateContext = React.createContext<StateContext>(false)
+const stateContext = React.createContext<StateContext>({
+  value: false,
+  addListener() {},
+  removeListener() {},
+  modify() {},
+})
 const setContext = React.createContext<SetContext>((_: boolean) => {})
 
 export function Provider({children}: React.PropsWithChildren<{}>) {
-  const [state, setState] = React.useState(false)
+  const mode = useSharedValue(false)
+  const setMode = React.useCallback(
+    (v: boolean) => {
+      mode.value = v
+    },
+    [mode],
+  )
   return (
-    <stateContext.Provider value={state}>
-      <setContext.Provider value={setState}>{children}</setContext.Provider>
+    <stateContext.Provider value={mode}>
+      <setContext.Provider value={setMode}>{children}</setContext.Provider>
     </stateContext.Provider>
   )
 }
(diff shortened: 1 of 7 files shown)


src/state/shell/minimal-mode.tsx
```diff
@@ -1,16 +1,28 @@
...
+  addListener() {},
+  removeListener() {},
+  modify() {},
+})
 const setContext = React.createContext<SetContext>((_: boolean) => {})
 
 export function Provider({children}: React.PropsWithChildren<{}>) {
-  const [state, setState] = React.useState(false)
+  const mode = useSharedValue(false)
+  const setMode = React.useCallback(
```
gaearon, line 17: I kept the setter so that writes are centralized and easy to trace. Technically code can screw around with `.value` but we can catch that in review.

src/view/screens/PostThread.tsx
```diff
@@ -83,17 +83,17 @@ export const PostThreadScreen = withAuthRequired(
             treeView={!!store.preferences.thread.lab_treeViewEnabled}
           />
         </View>
-        {isMobile && !minimalShellMode && (
-          <View
+        {isMobile && (
+          <Animated.View
             style={[
               styles.prompt,
+              fabMinimalShellTransform,
```
gaearon, line 90: We could rename this since this isn't technically FAB, I think it's fine though.

pfrazee (approved):
Nice

src/view/com/pager/FeedsTabBarMobile.tsx
```diff
@@ -38,7 +38,6 @@ export const FeedsTabBar = observer(function FeedsTabBarImpl(
         pal.border,
         styles.tabBar,
         headerMinimalShellTransform,
-        minimalShellMode && styles.disabled,
```
gaearon, line 41: I moved these into the animated styles so that we don't expose the raw value.

PR #1850, opened 2023-11-09
First pass at a session handler

estrattonbailey:


src/state/persisted/legacy.ts
```diff
@@ -35,7 +35,7 @@ type LegacySchema = {
     description: string
     avatar: string
   }
-  onboarding: {
+  onboarding?: {
```
estrattonbailey, line 38: Migration was failing bc we've removed these mobx models now. We should be careful going forward to make sure we can resume from a partial migration. So far so good, and in the next release with these changes, it will be a full migration, so probably not a big deal...

src/state/persisted/schema.ts
```diff
@@ -2,15 +2,17 @@ import {z} from 'zod'
...
-  refreshJwt: z.string().optional(),
-  accessJwt: z.string().optional(),
-  handle: z.string().optional(),
-  displayName: z.string().optional(),
-  aviUrl: z.string().optional(),
+  handle: z.string(),
+  refreshJwt: z.string().optional(), // optional because it can expire
+  accessJwt: z.string().optional(), // optional because it can expire
+  // displayName: z.string().optional(),
+  // aviUrl: z.string().optional(),
```
estrattonbailey, line 13: Thinking that anything NOT returned from session response should be loaded with RQ

src/state/persisted/schema.ts
```diff
@@ -2,15 +2,17 @@ import {z} from 'zod'
...
+// TODO agent.resumeSession requires the following fields
 const accountSchema = z.object({
   service: z.string(),
   did: z.string(),
-  refreshJwt: z.string().optional(),
-  accessJwt: z.string().optional(),
-  handle: z.string().optional(),
-  displayName: z.string().optional(),
-  aviUrl: z.string().optional(),
+  handle: z.string(),
```
estrattonbailey, line 9: Handle is returned from session, and the resumeSession method actually requires it in types but still works without it

src/state/shell/onboarding.tsx
```diff
@@ -82,12 +82,16 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
 
   React.useEffect(() => {
     return persisted.onUpdate(() => {
-      dispatch({
-        type: 'set',
-        step: persisted.get('onboarding').step as OnboardingStep,
-      })
+      const next = persisted.get('onboarding').step
+      // TODO we've introduced a footgun
+      if (state.step !== next) {
```
estrattonbailey, line 87: Hehe this was cyclically updating in the background. Wonder if we can protect against this better at a higher level?

src/view/com/auth/login/Login.tsx
```diff
@@ -344,6 +346,11 @@ const LoginForm = ({
...
       }
 
+      await login({
+        service: serviceUrl,
+        identifier: fullIdent,
+        password,
+      })
       await store.session.login({
         service: serviceUrl,
         identifier: fullIdent,
```
estrattonbailey, line 356: Just dual-booting rn

src/state/session/index.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+      logger.debug(`session: logged in`, {
+        service,
+        identifier,
+      })
+    },
+    [upsertAccount],
+  )
+
+  const logout = React.useCallback<ApiContext['logout']>(async () => {
```
pfrazee, line 214: We should send the delete session request, right?

src/state/shell/onboarding.tsx
pfrazee: mm that was on me, I think the only way to solve it is to avoid fires in response to this event -- the `type: set` needs a `silent: true` param

src/state/persisted/schema.ts
pfrazee: I guess we can fetch every time

pfrazee (commented):
This is looking like the right approach to me

src/App.web.tsx
```diff
@@ -34,10 +42,16 @@ const InnerApp = observer(function AppImpl() {
       setRootStore(store)
       analytics.init(store)
     })
-  }, [])
+  }, [resumeSession])
```
gaearon, line 45: not used here

src/state/session/index.tsx
```diff
@@ -0,0 +1,383 @@
...
+    agent: PUBLIC_BSKY_AGENT,
+  })
+
+  const setStateWrapped = React.useCallback(
+    (fn: (prev: StateContext) => StateContext) => {
+      let next: StateContext | undefined
+      setState(s => {
+        next = fn(s)
+        return next
+      })
```
gaearon, line 103: This looks dubious to me. In React, the updater functions run during the *next* render. I think this happens to be synchronous today just because it's called outside a batching context, but this will change in React in the future, so we shouldn't rely on this.

I think persisting in an Effect is the easiest solution here. If that doesn't work, I can suggest some other options.

gaearon (commented):
seems good overall

src/state/session/index.tsx
estrattonbailey: Good call! Yeah no longer need this. Updated to use an effect 👍

src/state/session/index.tsx
estrattonbailey: Is there a method like this? I don't see one. Atm no, we just clear the tokens from memory and storage

src/state/session/index.tsx
```diff
@@ -317,9 +297,16 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         }
       })
     },
-    [setStateWrapped],
+    [setState],
   )
 
+  React.useEffect(() => {
+    persisted.write('session', {
```
gaearon, line 304: Note you probably don't want to do this after first render. So maybe:

- Keep a wrapper that sets a dirty flag in a ref
- In the Effect, if the dirty flag is true, persist and unset the flag.

src/state/session/index.tsx
pfrazee: https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/server/deleteSession.json

can do this later though

pfrazee (approved):
Going to go ahead and merge so others can build off this. More will come in followup PRs. Good work @estrattonbailey

PR #1851, opened 2023-11-09
Refactor post threads to use react query

pfrazee:
This PR refactors the Post Thread code to use React Query (RQ).

In addition to refactoring the Post Thread, this PR introduces a "Post Shadow" cache which is used for optimistic updates that we want to keep synced across views (like, repost, and deletion state).

Some notes:

- ~This PR is using a temporary `useSession` hook that lightly wraps the existing session manager.~
- There are some commented out TODOs in the component which have to do with the previous post cache. We don't want to get rid of that yet because it's a case we're still going to need to handle -- using posts from a feed to create a pre-fetch skeleton of threads.
- There are a couple of components which I've temporarily replaced by adding a `2` to the filename, since there's still existing code depending on them.


src/state/queries/post-thread.ts
```diff
@@ -0,0 +1,200 @@
+import {AppBskyFeedDefs, AppBskyFeedGetPostThread} from '@atproto/api'
+import {useQuery, QueryClient, useQueryClient} from '@tanstack/react-query'
+import {useSession} from '../session'
+import {RQKEY as POST_RQKEY, getCachedPost} from './post'
+import {ThreadViewPreference} from '../models/ui/preferences'
+
+export const RQKEY = (uri: string) => ['post-thread', uri]
```
estrattonbailey, line 7: Not positive we'd really see an issue with this, but we might want to include the user did or something in this keyset so that switching users without a reload to clear memory never results in stale data being shown from cache.

src/state/queries/post-thread.ts
```diff
@@ -0,0 +1,200 @@
...
+        hydrateCache(queryClient, res.data.thread)
+        return threadViewToSkeleton(res.data.thread)
+      }
+      return {type: 'unknown', uri: uri!}
+    },
+    {enabled: !!uri},
+  )
+}
+
+export function sortThreadSkeleton(
```
estrattonbailey, line 70: These functions are so clear 🥺 I love them

src/state/queries/post-thread.ts
estrattonbailey, line 1: Nit: conceptually I think I like @gaearon's distinction between state and "cache" i.e. a local cache of remote data that only changes on a server. In respect to that, I'm kinda inclined to say we should put queries in a `src/data/*` dir or something outside the state dir, but that's a very soft opinion. Remote cache/data is "state" in a way too. Idk!

src/state/queries/post-thread.ts
pfrazee: I assumed we'd drop all RQ cache on a session change

src/state/queries/post-thread.ts
pfrazee: oh yeah why dont you marry them

src/state/queries/post.ts
```diff
@@ -0,0 +1,236 @@
+import {AppBskyFeedDefs} from '@atproto/api'
+import {
+  useQuery,
+  useQueryClient,
+  useMutation,
+  QueryClient,
+} from '@tanstack/react-query'
+import {useSession} from '../session'
+
+export const RQKEY = (postUri: string) => ['post', postUri]
```
estrattonbailey, line 10: Yeah same here. If we update cache with a like, then switch users, I imagine the post could in some cases still show as "liked" if we aren't careful. So might want did here? I could be wrong

src/state/queries/post-thread.ts
ansh: Would be nice to have our React Query Keys strongly typed in one place so we can refer to them for future updates. A very simple factory model should work: https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories

src/state/queries/post-thread.ts
ansh: Would love a comment overview of how this function sorts the posts for future contributors:
```
// 1. If node is not a post return it unmodified, otherwise continue.
// 2. If the node has replies, start the sorting process.
// 3. Retrieve the cached post for the current node using its URI.
// 4. Sort the replies array in place using a custom comparator.
// 5. Inside the comparator, give priority to posts over non-posts.
// 6. Retrieve cached posts for both comparison nodes using their URIs.
// 7. If the cached post for 'a' is not found, it is moved down in the sort order.
// 8. If the cached post for 'b' is not found, it is moved up in the sort order.
// 9. Check if the author of post 'a' is the original poster (OP) and the same for post 'b'.
// 10. If both are by the OP, sort by the timestamp 'indexedAt' to get the oldest post.
// 11. If only 'a' is by the OP, it gets higher priority.
// 12. If only 'b' is by the OP, 'a' gets higher priority.
// 13. If user preference is to prioritize followed users, compare the following status of each post's author.
// 14. Sort by followed status, giving priority to followed user's posts.
// 15. If sorting by 'oldest', sort by 'indexedAt' to get the oldest post.
// 16. If sorting by 'newest', sort by 'indexedAt' in reverse to get the newest post.
// 17. If sorting by 'most-likes', compare like counts; in case of a tie, use 'indexedAt' to prioritize newer posts.
// 18. If sorting by 'random', return a random order by using Math.random.
// 19. Default to sorting by 'indexedAt' in reverse if no other conditions are met.
// 20. Recursively apply the sorting function to each reply.
// 21. Finally, return the sorted node.
```

src/state/queries/post.ts
ansh, line 10: Would be nice to have this in a typed key generating factory as previously mentioned

src/state/queries/post.ts
```diff
@@ -0,0 +1,236 @@
...
+  const {agent} = useSession()
+  return useQuery<AppBskyFeedDefs.PostView>(
+    RQKEY(uri || ''),
+    async () => {
+      const res = await agent.getPosts({uris: [uri!]})
+      if (res.success && res.data.posts[0]) {
+        return res.data.posts[0]
+      }
+
+      throw new Error('No data')
```
ansh, line 29: Is this error propagated somewhere?

src/state/queries/post.ts
```diff
@@ -0,0 +1,236 @@
...
+        likeCount = post.likeCount || 0
+        return {
+          ...post,
+          likeCount: (post.likeCount || 0) + 1,
+          viewer: {...(post.viewer || {}), like: 'pending'},
+        }
+      })
+      return {likeCount}
+    },
+    onSuccess(data, variables) {
```
ansh, line 59: React Query v5 is out which simplifies the optimistic update workflow. We no longer have to manually update the cache every time: 

https://medium.com/@stojanovic.nemanja71/optimistic-updates-in-tanstack-query-v5-dfbcbb124113#:~:text=So%20instead%20of%20waiting%20for,won't%20notice%20any%20delay.

ansh: This applies to other optimistic updates as well. Not adding the comment everywhere fyi.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -23,109 +19,169 @@ import {ViewHeader} from '../util/ViewHeader'
...
+  const {data: resolvedUri} = useResolveUriQuery(uri)
+  const {
+    isLoading,
+    isError,
+    error,
+    refetch,
+    isRefetching,
+    data: threadSkeleton,
+  } = usePostThreadQuery(resolvedUri)
+  const {data: rootPost} = usePostQuery(threadSkeleton?.uri)
```
ansh, line 78: Love how the code is simplified!

src/view/com/post-thread/PostThread.tsx
```diff
@@ -278,68 +331,107 @@ export const PostThread = observer(function PostThread({
...
+  }, [navigation])
+
+  return (
+    <CenteredView>
+      <View style={[pal.view, pal.border, styles.notFoundContainer]}>
+        <Text type="title-lg" style={[pal.text, s.mb5]}>
+          Post hidden
+        </Text>
+        <Text type="md" style={[pal.text, s.mb10]}>
+          You have blocked the author or you have been blocked by the author.
```
ansh, line 385: Please merge with main and add `<Trans>` here

src/view/com/post-thread/PostThread.tsx
```diff
@@ -278,68 +331,107 @@ export const PostThread = observer(function PostThread({
...
-    )
-  }
-  if (view.isBlocked) {
+  }, [navigation])
+
+  return (
+    <CenteredView>
+      <View style={[pal.view, pal.border, styles.notFoundContainer]}>
+        <Text type="title-lg" style={[pal.text, s.mb5]}>
+          Post hidden
```
ansh, line 382: `<Trans>` here too

src/view/screens/PostThread.tsx
```diff
@@ -23,62 +23,49 @@ type Props = NativeStackScreenProps<CommonNavigatorParams, 'PostThread'>
...
-            handle: view.thread.post.author.handle,
-            displayName: view.thread.post.author.displayName,
-            avatar: view.thread.post.author.avatar,
+            handle: post.author.handle,
+            displayName: post.author.displayName,
+            avatar: post.author.avatar,
           },
         },
-        onPost: () => view.refresh(),
+        // onPost: () => view.refresh(), TODO need to trigger this elsewhere
```
ansh, line 59: We can probably create a helper function that takes in a RQ key and refreshes it, right?

src/state/queries/post.ts
pfrazee: it becomes RQ error state, right?

src/state/queries/post-thread.ts
estrattonbailey: I'm +1 to this, nice to have a quick reference to what query keys are out there

estrattonbailey: Actually @ansh does it change your mind any if we have separate query hooks for each query like Paul has here? I.e. `usePostQuery` instead of `useQuery(RQKEY('uri'), () => {})`. So finding all usages is pretty easy either way.

src/state/queries/post-thread.ts
pfrazee: I'm not a huge fan of describing code a second time in the comments -- only summarizing or explaining

src/view/screens/PostThread.tsx
pfrazee: Yep, done

src/state/queries/post.ts
pfrazee: The link above is how we should do things in _most_ cases. In this case, the post shadow is synced across components and it's better to put in these handlers

PR #1855, opened 2023-11-09
Sync top/bottom bar disappearance to the scroll

gaearon:
OK, I've wanted to do this since forever.

The most natural way for the "minimal shell" showing and hiding to work is to stay exactly in sync with the scroll gesture. It shouldn't "fly away" independently from what your finger is doing. This avoids showing weird blank spaces.

This might need a bit of testing. Happy to take feedback on the code/naming, not sure if it's a bit messy. Note the screen captures are missing the "is scrolled down" logic for the left floating button but I fixed it after recording.

## iOS

https://github.com/bluesky-social/social-app/assets/810438/f713857e-bacd-407e-a9b5-18bc1dc14f75

## Android

https://github.com/bluesky-social/social-app/assets/810438/b046a14b-7f2c-42b1-85fe-9a2c5bab5882

## Web

https://github.com/bluesky-social/social-app/assets/810438/d19cd42b-4083-4c17-b2a4-adb84d2331e9

PR #1856, opened 2023-11-09
Show tabs when swiping feeds

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/1855.

@ansh noticed that switching tabs in minimal shell mode shows blank space at the top initially (and you don't see where you're swiping). This fixes it by disabling the minimal shell as the swipe starts.

https://github.com/bluesky-social/social-app/assets/810438/0e08a9d8-9d8c-42d2-9a67-9e8bba8cf820

## Test Plan

This one is pretty straightforward. Verified it works on iOS and Android. Doesn't affect web.


pfrazee (approved):
Nice yearh

PR #1857, opened 2023-11-09
Fix tab alignment on the web

gaearon:
Fixes an issue @ansh pointed out in another PR.

This line seems to only affect web; I don't see any difference on iOS and Android.

It's pretty odd that none of these styles have changed in a while:

Screenshot 2023-11-09 at 21 52 10

So I don't know how it could have regressed recently, but whatever.

## Test Plan

Before:

https://github.com/bluesky-social/social-app/assets/810438/d245406b-e3f0-42da-ac9e-02a5b8afe7eb

After:

https://github.com/bluesky-social/social-app/assets/810438/4e9daf14-2464-4d2b-a296-8e18fd4b8c0b


pfrazee (approved):
ez pz

PR #1860, opened 2023-11-10
Web login/signup and shell

estrattonbailey:


src/state/persisted/schema.ts
```diff
@@ -7,6 +7,8 @@ const accountSchema = z.object({
   service: z.string(),
   did: z.string(),
   handle: z.string(),
+  email: z.string(),
+  emailConfirmed: z.boolean(),
```
estrattonbailey, line 11: These come back on session responses too

src/view/com/auth/login/ChooseAccountForm.tsx
```diff
@@ -1,52 +1,93 @@
...
+}) {
+  const pal = usePalette('default')
+  const {_} = useLingui()
+  const {isError, data} = useGetProfile({did: account.did})
+
+  const onPress = React.useCallback(() => {
+    onSelect(account)
+  }, [account, onSelect])
+
+  if (isError) return null
```
estrattonbailey, line 31: So yeah, we can fetch profile info, but we kinda need to handle failures here too 🤔

src/view/com/auth/withAuthRequired.tsx
```diff
@@ -6,25 +6,25 @@ import {
...
   observer(function AuthRequired(props: P) {
-    const store = useStores()
+    const {isInitialLoad, hasSession} = useSession()
     const onboardingState = useOnboardingState()
-    if (store.session.isResumingSession) {
+    if (isInitialLoad) {
       return <Loading />
     }
-    if (!store.session.hasSession) {
+    if (!hasSession) {
```
estrattonbailey, line 27: Should Just Work™

src/view/shell/desktop/LeftNav.tsx
```diff
@@ -41,18 +41,31 @@ import {router} from '../../../routes'
...
+  const {
+    isLoading,
+    isError,
+    data: profile,
+  } = useGetProfile({did: currentAccount!.did})
   const {isDesktop} = useWebMediaQueries()
   const size = 48
-  return store.me.handle ? (
+
+  if (isError || !profile || !currentAccount) return null
```
estrattonbailey, line 57: Some checks are just to appease TS. `LeftNav` isn't rendered unless we're authed (see shell), so we _could_ just ignore these with a `!`. We could maybe do some type trickery too...

src/data/useGetProfile.ts
pfrazee, line 1: This is different from the conventions I've been using. I've been using the `queries` folder, using the name of the object (in this case it would be `profile`), and naming the hooks `use*Query` (eg `useProfileQuery`).

So basically you're using a method-driven name and I'm using a resource-driven name. We can bikeshed which to use but we gotta get on the same page.

src/view/com/auth/login/ChooseAccountForm.tsx
pfrazee: Hmm, I think in this case we just render without the data and ignore the error, right? The data being fetched just enhances what we see

src/view/com/modals/SwitchAccount.tsx
```diff
@@ -19,89 +18,115 @@ import {BottomSheetScrollView} from '@gorhom/bottom-sheet'
...
+  const isCurrentAccount = account.did === currentAccount?.did
+  const {onPressSwitchAccount} = useAccountSwitcher()
+
+  const onPressSignout = React.useCallback(() => {
+    track('Settings:SignOutButtonClicked')
+    logout()
+  }, [track, logout])
+
+  // TODO
+  if (isError || !currentAccount) return null
```
pfrazee, line 42: Same story here, I think we can just render best-effort and ignore the error

src/view/screens/Settings.tsx
```diff
@@ -67,6 +68,73 @@ import {STATUS_PAGE_URL} from 'lib/constants'
...
+function SettingsAccountCard({account}: {account: SessionAccount}) {
+  const pal = usePalette('default')
+  const {isSwitchingAccounts, currentAccount} = useSession()
+  const {logout} = useSessionApi()
+  const {isError, data} = useGetProfile({did: account.did})
+  const isCurrentAccount = account.did === currentAccount?.did
+  const {onPressSwitchAccount} = useAccountSwitcher()
+
+  // TODO
+  if (isError || !currentAccount) return null
```
pfrazee, line 80: Same

pfrazee (changes requested):
A few small questions to address but otherwise solid

src/data/useGetProfile.ts
estrattonbailey: Switched to your conventions

src/view/com/auth/login/ChooseAccountForm.tsx
estrattonbailey: Sg I'll update

pfrazee (approved):
🔥

PR #1862, opened 2023-11-10
Refactor feeds to use react-query

pfrazee:
Refactors the feed to use react-query.

The feed query-key is the `FeedDescriptor` type, which is a string that uses template-literal-types to ensure correct usage. I decided to go with a string because it gets passed around a fair amount and it'll be easier to avoid needless re-renders with that. Also note: it uses a pipe (`|`) separator for the params to avoid colliding with potential values (initially I used `:` but that collides with URIs).


src/state/preferences/feed-tuners.tsx
```diff
@@ -0,0 +1,45 @@
+import {FeedTuner} from '#/lib/api/feed-manip'
+import {FeedDescriptor} from '../queries/post-feed'
+import {useLanguagePrefs} from './languages'
+
+export function useFeedTuners(feedDesc: FeedDescriptor) {
```
gaearon, line 5: Do you expect there'll be more user-specific stuff here pulled from context etc? If not I think I'd prefer this to just be `getFeedTuners(...)` and then the rest is filled in at the callsite. So that we don't get into a situation where we want this but it only works in a component.

src/state/queries/post-feed.ts
```diff
@@ -0,0 +1,168 @@
...
+    if (post) {
+      const slices = tuner.tune([post], feedTuners, {
+        dryRun: true,
+        maintainOrder: true,
+      })
+      if (slices[0]) {
+        if (
+          !moderatePost(
+            slices[0].items[0].post,
+            store.preferences.moderationOpts,
```
gaearon, line 106: Since we're not guaranteed to be in an `observer`, this won't necessarily "subscribe" to store.preferences.moderationOpts so need to be careful if we want this to update on existing screens. I think you can solve this by wrapping in autorun while we still depend on MobX state.

src/state/queries/post-feed.ts
```diff
@@ -0,0 +1,168 @@
...
+          ).content.filter
+        ) {
+          return true
+        }
+      }
+    }
+    return false
+  }, [api, tuner, feedTuners, store.preferences.moderationOpts, enabled])
```
gaearon, line 114: Note this memoization is not doing anything because `feedTuners` is always a fresh array. Could fix it there, or remove it here and not bother with it.

gaearon (commented):
lgtm overall

src/state/queries/post-feed.ts
pfrazee: We're going to move all preferences to hooks in the nearterm so I'm just leaving this stuff unfixed temporarily

src/state/queries/post-feed.ts
pfrazee: Oh no, yep I need to fix that in the feed tuners hook, good catch

src/state/preferences/feed-tuners.tsx
pfrazee: We will need to make the final call depending based on how the UX feels. If we want to automatically trigger re-fetches due to changes in the user preferences, then we need them in a hook. If we dont, we can just use a getter and then the user will refresh their feed manually.

estrattonbailey (commented):
Looking good to me. One smaller note: stale data flashes and is replaced when you toggle back and forth between feeds. We might want to play with staleTime, maybe even only manually marking as stale to trigger a refresh.

src/state/queries/post-feed.ts
```diff
@@ -0,0 +1,171 @@
...
+      // shouldnt happen
+      return new FollowingFeedAPI(agent)
+    }
+  }, [feedDesc, agent])
+  const tuner = useMemo(
+    () => (opts?.disableTuner ? new NoopFeedTuner() : new FeedTuner()),
+    [opts?.disableTuner],
+  )
+
+  const pollLatest = useCallback(async () => {
```
estrattonbailey, line 94: I wonder if we shouldn't just do polling, `hasNew` state, etc in this hook and expose as state 🤔  Would remove some logic from `Feed` and make that simpler.

In general, I think we should aim for something as simple as:
- fetch data and handle refresh and state up top (like in this hook)
  - all logic lives here
- construct item skeleton
- pass to FlatList to render

Maybe can't get to that point, or not right now anyway, but could be a good guiding light moving forward.

src/state/preferences/feed-tuners.tsx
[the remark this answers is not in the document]
pfrazee: not 100% sure yet, but I'd suggest we do a code reorg once the dust settles

src/view/com/posts/FeedItem.tsx
```diff
@@ -1,151 +1,147 @@
...
+        facets: record.facets,
+      }),
+    [record],
+  )
+  if (postShadowed === POST_TOMBSTONE) {
+    return null
+  }
+  if (richText && moderation) {
+    return (
+      <FeedItemLoaded
```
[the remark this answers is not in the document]
pfrazee: oh yep, I was copying patterns and not paying attention

src/state/queries/post-feed.ts
pfrazee: That's probably possible, though we do need the `onHasNew` emitter so that we can emit that state back out to the wrapper components

estrattonbailey: If outer components need state, could we move this hook up a level and flatten children out more? Then we don't need to emit anything, we just check for `hasNew` and render stuff.

But yeah not saying we should do all this here 👍

src/state/preferences/feed-tuners.tsx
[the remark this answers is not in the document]
estrattonbailey: Yeah nvm me, let's wait

src/state/queries/post-feed.ts
pfrazee: I'd like to keep the hook inside the `<Feed>` component but maybe the answer is to move that state and UI behaviors into the Feed component

src/lib/api/feed/merge.ts
```diff
@@ -190,11 +200,11 @@ class MergeFeedSource_Following extends MergeFeedSource {
     cursor: string | undefined,
     limit: number,
   ): Promise<AppBskyFeedGetTimeline.Response> {
-    const res = await this.rootStore.agent.getTimeline({cursor, limit})
+    const res = await this.agent.getTimeline({cursor, limit})
     // run the tuner pre-emptively to ensure better mixing
     const slices = this.tuner.tune(
       res.data.feed,
-      this.rootStore.preferences.getFeedTuners('home'),
+      this.agent.preferences.getFeedTuners('home'),
```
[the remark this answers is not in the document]
pfrazee: Nah I'm working on the merge feed now

PR #1863, opened 2023-11-10
Make the pager non-jumpy

gaearon:
Currently the pager "jumps" the header translation if needed when you switch between tabs.

https://github.com/bluesky-social/social-app/assets/810438/97c8c2c5-0413-4bb7-b689-21d63becfa6e

Although that looks kinda cool on this screen, it becomes disorienting when you have multiple feed tabs, like on Profile.

This PR changes the pager to keep the tabs in sync by adjusting their *content* scroll positions instead:

https://github.com/bluesky-social/social-app/assets/810438/06d72dc3-51c9-491b-9ffa-0b914765ed24

This looks more boring which, with this kind of interface, is probably a good thing.

One concession I've had to make is that the scrollables now have to have a large enough height to make sure they can always be scrolled to the state where the header is collapsed. If there's a better way to do it, let me know.

## Test Plan

Went through different lists on iOS, Android, and Web.

### DIFF
--- a/src/view/com/lists/ListItems.tsx
+++ b/src/view/com/lists/ListItems.tsx
@@ -1,6 +1,7 @@
 import React, {MutableRefObject} from 'react'
 import {
   ActivityIndicator,
+  Dimensions,
   RefreshControl,
   StyleProp,
   View,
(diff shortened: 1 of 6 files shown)


gaearon:
Not super happy with how complicated it is. I'll think more.

pfrazee (approved):
Yeah looks good

PR #1866, opened 2023-11-10
Push useAnimatedScrollHandler down everywhere to work around bugs

gaearon:
It's not safe to use it anywhere except leaf components due to https://github.com/software-mansion/react-native-reanimated/issues/5364.

So we can't call it upwards and pass the result downwards. It "works" in some cases but we risk having stale props and state inside every handler, which is happening in my other refactors.

Since the functionality itself is extremely useful, I'm converting our code to a pattern where the thing we pass down is the config — i.e. the handlers. Then we wrap it at the lowest level.

There are effectively two sources of these handlers right now:

- `useOnMainScroll` handlers
- `PagerWithHeader` handlers

I've converted both sources and tightened up the types.

## Test Plan

Walk around the affected screens:

- Home
- Notifications
- Search (when typing)
- Profile Feeds
- Profile Lists
- Mod Lists

Do it on every platform.

Observe no regressions.

### DIFF
--- a/src/lib/hooks/useAnimatedScrollHandler_FIXED.ts
+++ b/src/lib/hooks/useAnimatedScrollHandler_FIXED.ts
@@ -1,1 +1,15 @@
+// Be warned. This Hook is very buggy unless used in a very constrained way.
+// To use it safely:
+//
+// - DO NOT pass its return value as a prop to any user-defined component.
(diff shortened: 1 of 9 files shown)


pfrazee (approved):
Yeah makes sense

PR #1868, opened 2023-11-10
Fix initial pager gap after fast scroll

gaearon:
This fixes a bug in the new pager sync scrolling logic I've added in https://github.com/bluesky-social/social-app/pull/1863.

The previous code was saying: "sync scroll only if we didn't scroll past the header". The problem is that if we scroll fast, we're gonna have a large gap between the last event (which didn't reach the limit, but got synced) and the next event (which went over the limit, and didn't get synced).

Instead, let's track the last value that got synced. During every scroll, we'll calculate the new target, applying the limit. If we need to sync up the scroll, we'll do so and store the current value as the last one that got synced.

As a result, if we get an event that "overshoots" the header hiding threshold, we'll still adjust other pages to be exactly at that limit.

## Test Plan

Flicking one of the tabs down really fast no longer causes a blank space on other tabs.

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -138,17 +138,20 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
       })
     }
 
+    const lastForcedScrollY = useSharedValue(0)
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
nice

PR #1869, opened 2023-11-10
Use min height for pager lists and increase it

gaearon:
For pager pages, we need to pad out enough space so that even if the pager is scrolled down, switching tabs does not cause a jump. The previous calculation I used didn't really make sense — it was too small in some cases.

I don't know the science behind which number to pick because it doesn't seem to actually correspond to the scrolled distance. Something's up with layout there. So I picked 1.5 screen heights which seems to be enough in my testing.

Also, it should be min height rather than bottom padding so that when the content is loaded, it doesn't stay at the tail.

¯\\_(ツ)_/¯

### DIFF
--- a/src/view/com/lists/ListItems.tsx
+++ b/src/view/com/lists/ListItems.tsx
@@ -227,7 +227,7 @@ export const ListItems = observer(function ListItemsImpl({
           />
         }
         contentContainerStyle={{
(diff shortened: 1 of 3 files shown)


pfrazee (approved):

leoleitedev:
LGTM!

PR #1870, opened 2023-11-10
Refactor profile screen to use new pager and react-query

gaearon:
- Updates profile screen to use the new `<PagerWithHeader>`
- Updates profile screen to use react-query
- Introduces a profile shadow cache, similar to the posts shadow cache

Still todo in followup PRs:

- [ ] Profile preview
- [ ] Edit profile modal
- [ ] Profile avatar lightbox
- [ ] Dynamic tabs
- [ ] Feed tab behaviors
- [ ] Lists tab
- [ ] Feeds tab

### DIFF
--- a/src/state/cache/profile-shadow.ts
+++ b/src/state/cache/profile-shadow.ts
@@ -1,0 +1,88 @@
+import {useEffect, useState, useCallback, useRef} from 'react'
+import EventEmitter from 'eventemitter3'
+import {AppBskyActorDefs} from '@atproto/api'
+
+const emitter = new EventEmitter()
+
+export interface ProfileShadow {
+  followingUri: string | undefined
+  muted: boolean | undefined
+  blockingUri: string | undefined
+}
+
+interface CacheEntry {
+  ts: number
+  value: ProfileShadow
+}
+
+type ProfileView =
+  | AppBskyActorDefs.ProfileView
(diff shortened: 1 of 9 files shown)


src/state/cache/profile-shadow.ts
```diff
@@ -0,0 +1,88 @@
...
+    return () => {
+      emitter.removeListener(profile.did, onUpdate)
+    }
+  }, [profile.did, onUpdate])
+
+  // react to profile updates
+  useEffect(() => {
+    // dont fire on first run to avoid needless re-renders
+    if (!firstRun.current) {
+      setState({ts: Date.now(), value: fromProfile(profile)})
```
gaearon, line 52: Can you describe how you want this to work? Generally we shouldn't be setting state from effects. If there's a need to adjust state, the canonical way is to do [this](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes). But I think I'm not quite following the logic in general so I want to understand the intended behavior.

src/state/cache/profile-shadow.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    firstRun.current = false
+  }, [profile])
+
+  return state.ts > ifAfterTS ? mergeShadow(profile, state.value) : profile
```
gaearon, line 57: We should probably useMemo this.

PR #1875, opened 2023-11-11
Refactor lists to use new queries

pfrazee:
Adds queries:

- `my-follows` Cached list of the users follows, needed by `actor-autocomplete`. Just a first pass; may need more tuning around cache-invalidation.
- `actor-autocomplete` Autocomplete view, needed by one of the modals. Same as `my-follows`, just a first pass.
- `my-lists` Get a list of the current user's lists, possibly filtered by mod or curate
- `profile-lists` Get a list of the given user's created lists.
- `list` Get a list's information, includes mutators for acting on a list (create, update metadata, delete, mute, block)
- `list-members` Get a list's members
- `list-memberships` A temporary solution to managing users' membership in a list. For a variety of reasons, had to be a separate query than `list-members`. See the comments for more info. Also includes mutators for adding/remove users memberships.

Includes some minor changes to the UIs:

- `ListAddUserModal` is now `ListAddRemoveUsersModal`, supports the remove action now
- `UserAddRemoveListsModal` switched from checkboxes to just add/remove buttons


src/state/queries/list.ts
```diff
@@ -0,0 +1,285 @@
...
+      await whenAppViewReady(agent, uri, (v: AppBskyGraphGetList.Response) => {
+        return !v?.success
+      })
+    },
+    onSuccess() {
+      invalidateMyLists(queryClient)
+      queryClient.invalidateQueries({
+        queryKey: PROFILE_LISTS_RQKEY(currentAccount!.did),
+      })
+      // TODO!! /* dont await */ this.rootStore.preferences.removeSavedFeed(this.uri)
```
pfrazee, line 227: Dont forget this with the saved/pinned feeds updates @estrattonbailey

src/state/queries/my-follows.ts
```diff
@@ -0,0 +1,43 @@
...
+          actor: currentAccount.did,
+          cursor,
+          limit: PAGE_SIZE,
+        })
+        // TODO
+        // res.data.follows = res.data.follows.filter(
+        //   profile =>
+        //     !moderateProfile(profile, this.rootStore.preferences.moderationOpts)
+        //       .account.filter,
+        // )
```
pfrazee, line 33: We should decide if we want to do this filter here or not, and if so bring in the moderation opts. (Might not be the best place for it)

src/state/queries/list.ts
```diff
@@ -0,0 +1,285 @@
...
+}
+export function useListCreateMutation() {
+  const {agent, currentAccount} = useSession()
+  const queryClient = useQueryClient()
+  return useMutation<{uri: string; cid: string}, Error, ListCreateMutateParams>(
+    {
+      async mutationFn({purpose, name, description, avatar}) {
+        if (!currentAccount) {
+          throw new Error('Not logged in')
+        }
```
estrattonbailey, line 51: I wonder if we should just add a `useSafeSession` or something that will throw if used outside a authed context. May not work with the public view in all cases? Maybe easier to just do it like this in a bunch of places. Just thinking out loud.

src/state/queries/list.ts
```diff
@@ -0,0 +1,285 @@
...
+      })
+    },
+  })
+}
+
+async function whenAppViewReady(
+  agent: BskyAgent,
+  uri: string,
+  fn: (res: AppBskyGraphGetList.Response) => boolean,
+) {
```
estrattonbailey, line 274: This is a handy little fella

src/state/queries/my-follows.ts
estrattonbailey: We could def do that here if we need it. Might make sense to do it all inside this query instead of later on with the results of this query? Although a filter here and a mute/unmute or something later could get caches out of sync. So maybe best to leave this response raw for easy caching and filter later with fresh moderation data each time.

src/view/com/lists/ListMembers.tsx
```diff
@@ -94,53 +105,44 @@ export const ListItems = observer(function ListItemsImpl({
...
-        onAdd(listUri: string) {
-          if (listUri === list.uri) {
-            list.cacheAddMember(profile)
-          }
-        },
-        onRemove(listUri: string) {
-          if (listUri === list.uri) {
-            list.cacheRemoveMember(profile)
-          }
-        },
```
estrattonbailey, line 132: Nice

src/view/screens/ProfileList.tsx
```diff
@@ -60,233 +65,204 @@ interface SectionRef {
...
   const palInverted = usePalette('inverted')
   const {_} = useLingui()
   const navigation = useNavigation<NavigationProp>()
+  const {currentAccount} = useSession()
   const {openModal, closeModal} = useModalControls()
+  const listMuteMutation = useListMuteMutation()
+  const listBlockMutation = useListBlockMutation()
+  const listDeleteMutation = useListDeleteMutation()
+  const isCurateList = list.purpose === 'app.bsky.graph.defs#curatelist'
+  const isModList = list.purpose === 'app.bsky.graph.defs#modlist'
```
estrattonbailey, line 253: Any ideas for where we could keep these types of values and utils?

src/view/screens/ProfileList.tsx
```diff
@@ -60,233 +65,204 @@ interface SectionRef {
...
-    list.togglePin().catch(e => {
-      Toast.show('There was an issue contacting the server')
-      logger.error('Failed to toggle pinned list', {error: e})
-    })
-  }, [list])
+    // TODO
+    // list.togglePin().catch(e => {
+    //   Toast.show('There was an issue contacting the server')
+    //   logger.error('Failed to toggle pinned list', {error: e})
+    // })
```
estrattonbailey, line 265: Noted

src/view/screens/ProfileList.tsx
```diff
@@ -309,18 +285,18 @@ const Header = observer(function HeaderImpl({
         closeModal()
       },
     })
-  }, [openModal, closeModal, list])
+  }, [openModal, closeModal, list, listMuteMutation])
 
   const onUnsubscribeMute = useCallback(async () => {
     try {
-      await list.unmute()
+      await listMuteMutation.mutateAsync({uri: list.uri, mute: false})
```
estrattonbailey, line 293: Mutes don't invalidate query, so profile list items don't show the label. Do we need that here? If you nav away and come back the muted label shows.

estrattonbailey (approved):

src/state/queries/list-memberships.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ * This API is dangerous! Read the note above!
+ */
+export function useDangerousListMembershipsQuery() {
+  const {agent, currentAccount} = useSession()
+  return useQuery<ListMembersip[]>({
+    queryKey: RQKEY(),
+    async queryFn() {
```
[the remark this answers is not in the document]
pfrazee: Eh only problem is that 30s is too long, and anything that's the right interval will cause a _lot_ of extraneous refetches

src/state/queries/list.ts
pfrazee: Not a bad thought, though `Safe` usually means won't throw, so maybe `useAssertedSession`

src/state/queries/list.ts
pfrazee: ya

src/view/screens/ProfileList.tsx
pfrazee: I dunno if we need to, the intellisense picks up on the purpose values pretty well

src/view/screens/ProfileList.tsx
pfrazee: Eh, probably not the biggest thing in the world

PR #1881, opened 2023-11-13
Update clear account behavior, ensure agent is set

estrattonbailey:


src/state/session/index.tsx
```diff
@@ -178,6 +178,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         }),
       )
 
+      setState(s => ({...s, agent}))
```
estrattonbailey, line 181: Set agent after login, was missing this

src/state/session/index.tsx
```diff
@@ -373,9 +374,14 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
   )
 
+  /**
+   * Clears the `currentAccount` from session. Typically used to drop the user
+   * back to the sign-in page.
+   */
   const clearCurrentAccount = React.useCallback(() => {
     setStateAndPersist(s => ({
       ...s,
+      agent: PUBLIC_BSKY_AGENT,
```
estrattonbailey, line 384: This is a soft log out. The user was authed and wanted to sign in with a different account.

src/state/session/index.tsx
```diff
@@ -425,10 +431,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
           logger.DebugContext.session,
         )
 
-        logout()
+        clearCurrentAccount()
       }
     })
-  }, [state, logout, initSession])
+  }, [state, clearCurrentAccount, initSession])
```
estrattonbailey, line 437: This happens in response to a broadcast channel update. A user could simply be adding a new account, in which case we don't want to nuke their tokens.

If a user _does_ logout, this still still be called, and the other tab will have already nuked their tokens, so same difference.

pfrazee (approved):

PR #1887, opened 2023-11-13
Factor our feed source model

estrattonbailey:


src/state/queries/like.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function useUnlikeMutation() {
+  const {agent} = useSession()
+
+  return useMutation({
+    mutationFn: async ({uri}: {uri: string}) => {
+      await agent.deleteLike(uri)
+    },
+  })
+}
```
estrattonbailey, line 24: Figured some generic helpers might be good?

src/state/queries/suggested-feeds.ts
```diff
@@ -0,0 +1,29 @@
...
+    InfiniteData<AppBskyFeedGetSuggestedFeeds.OutputSchema>,
+    QueryKey,
+    string | undefined
+  >({
+    queryKey: suggestedFeedsQueryKey,
+    queryFn: async ({pageParam}) => {
+      const res = await agent.app.bsky.feed.getSuggestedFeeds({
+        limit: 10,
+        cursor: pageParam,
+      })
```
estrattonbailey, line 23: We only return 10 right now regardless, but in the future we may actually paginate

src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -23,7 +22,7 @@ import {
 } from '#/state/queries/preferences'
 import {useFeedSourceInfoQuery} from '#/state/queries/feed'
 
-export const NewFeedSourceCard = observer(function FeedSourceCardImpl({
+export const FeedSourceCard = observer(function FeedSourceCardImpl({
```
estrattonbailey, line 25: Had built a new version alongside the old. Replaced last old version with new and renamed.

src/view/com/util/post-embeds/index.tsx
```diff
@@ -72,7 +72,13 @@ export function PostEmbeds({
     // custom feed embed (i.e. generator view)
     // =
     if (AppBskyFeedDefs.isGeneratorView(embed.record)) {
-      return <CustomFeedEmbed record={embed.record} />
+      return (
+        <FeedSourceCard
+          feedUri={embed.record.uri}
+          style={[pal.view, pal.border, styles.customFeedOuter]}
+          showLikes
+        />
```
estrattonbailey, line 80: Portable, only needs a URI

pfrazee (approved):
Lookin great

PR #1893, opened 2023-11-14
Port Profile Followers/Follows to RQ

gaearon:
- [x] Port `ProfileFollowers`
- [x] Port `ProfileFollows`
- [x] Port `FollowButton`
  - [x] Fix the button to update on completion
  - [x] Check if shadow cache is working
  - [x] Check if we broke the usage in Onboarding
- [x] Delete or replace now-broken `useFollowProfile`
  - [x] Used from `SuggestedFollow`


src/view/com/profile/FollowButton.tsx
```diff
@@ -20,28 +21,59 @@ export const FollowButton = observer(function FollowButtonImpl({
...
+      Toast.show(`An issue occurred, please try again.`)
     }
-  }, [toggle, onToggleFollow])
+  }
 
-  if (state === FollowState.Unknown) {
+  const onPressUnfollow = async () => {
+    if (!profile.viewer?.following) {
+      return
+    }
```
estrattonbailey, line 42: There's a split second where `profile.viewer.following` can be set to `pending`, and this fn should return early in that case too. Rn we're reusing this prop that's typically either `undefined | URI`, so that we know a follow request is in-flight.

estrattonbailey (commented):
Looking good to me!

src/view/com/profile/ProfileCard.tsx
```diff
@@ -188,34 +189,33 @@ const FollowersList = observer(function FollowersListImpl({
...
-    )
-  },
-)
+  return (
+    <ProfileCard
+      profile={profile}
+      noBg={noBg}
+      noBorder={noBorder}
+      followers={followers}
+      renderButton={isMe ? undefined : () => <FollowButton profile={profile} />}
```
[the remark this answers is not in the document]
gaearon: yup, thanks!

src/view/com/profile/FollowButton.tsx
gaearon: So, this is based on the pattern in `ProfileHeader` so I'd rather align with that. If it's wrong we should fix it in both places — and probably many more places because `ProfileHeader` uses a similar pattern for a bunch of other handlers. So I'd keep it out of scope of this PR.

pfrazee: Added to tracker

pfrazee (approved):
looks good!

PR #1896, opened 2023-11-14
fix: show emoji picker on desktop at smaller window widths

### LINKED ISSUE #1880: Emoji picker disappears when window size is shrunk
**Describe the bug**
Odd UI bug where the emoji picker entirely disappears when the window is shrunk (ie, to half frame). This occurs after the window has shrunk to where the sidebars disappear, around ~1308px on a 1920px wide display.

There's also a tangential issue with the sidebar getting screwy for an edge case at the crossover period when the window is shrunk. See the second screenshot. Should be noted as well that there is still space for the sidebars when the sidebars choose to cut off. Might be better to have them gradually go with one sidebar going before the other.

**To Reproduce**

1. Open post editor on Web (browser agnostic)
2. Adjust window size
3. Observe

**Expected behavior**

The emoji picker should still be there and the sidebars should not be screwy.

**Screenshots**

The emoji picker is gone:
image
Also, as a side note, there exists an odd in-between state for a few pixels that appears to be an edge case where the sidebar gets very screwy:
image

**Details**

 - Platform: desktop Windows 10 web (browser agnostic)
 - Platform version:
 - App version:

voidstar0:
Currently, `isDesktop` media query is used to conditionally display the emoji picker. 

However, this will return false on desktop at smaller window widths. Instead, we should use the web platform detection.

Fixes #1880 

https://github.com/bluesky-social/social-app/assets/25884226/a48df591-fead-40d2-948d-efda63a64c28

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -40,7 +40,13 @@ import {OpenCameraBtn} from './photos/OpenCameraBtn'
 import {usePalette} from 'lib/hooks/usePalette'
 import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {useExternalLinkFetch} from './useExternalLinkFetch'
-import {isWeb, isNative, isAndroid, isIOS} from 'platform/detection'
+import {
+  isWeb,
+  isNative,
+  isAndroid,
+  isIOS,
+  isMobileWeb,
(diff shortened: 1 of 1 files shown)


pfrazee:
Ended up doing this in https://github.com/bluesky-social/social-app/pull/2257, apologies -- didnt see this sooner

PR #1898, opened 2023-11-14
Handle end of feed

estrattonbailey:
Forgot to handle this


src/view/screens/Feeds.tsx
```diff
@@ -154,6 +155,15 @@ export const FeedsScreen = withAuthRequired(function FeedsScreenImpl(
     await refetchPopularFeeds()
     setIsPTR(false)
   }, [setIsPTR, refetchPopularFeeds])
+  const onEndReached = React.useCallback(() => {
+    if (isUserSearching || !hasNextPopularFeedsPage || popularFeedsError) return
```
pfrazee, line 159: You also need to `!isFetching` per the docs https://tanstack.com/query/latest/docs/react/guides/infinite-queries

PR #1899, opened 2023-11-14
Composer update (react-query refactor)

pfrazee:
- Moves composer state to a context
- Reworks the composer to use RQ

Notes

- I kept two mobx models which are used for the image galleries. They don't touch the rootstore god object. If we don't get to them during this refactor we can survive; just trying to keep us moving.
- I had to create a non-RQ autocomplete class to interact with tiptap. Cest la vie.


src/state/models/media/gallery.ts
```diff
@@ -8,10 +7,8 @@ import {getImageDim} from 'lib/media/manip'
 export class GalleryModel {
   images: ImageModel[] = []
 
-  constructor(public rootStore: RootStoreModel) {
-    makeAutoObservable(this, {
-      rootStore: false,
-    })
+  constructor() {
+    makeAutoObservable(this)
```
estrattonbailey, line 11: Beautiful

estrattonbailey (approved):
Works great

PR #1900, opened 2023-11-14
Add Shadow type

pfrazee:
We use a "shadow cache" for profiles and posts to ensure synchronized state across views and to handle optimistic updates. As we've begun passing around the shadowed values, we noticed that it's easy to miss when a component depends on the shadowing. This is a problem because passing unshadowed posts/profiles will cause the component to break.

This PR adds a `Shadow<T>` type to help avoid that footgun.


estrattonbailey (approved):
Yep, perfect

PR #1902, opened 2023-11-14
Fix root sibling context issue

estrattonbailey:
Noticed that the language selector in the composer broke bc of no access to i18n context. I moved any non-component-related providers higher up.


pfrazee (approved):
good good

PR #1903, opened 2023-11-14
Fix self mention, resolve handle

estrattonbailey:
Replaces logic with new patterns, adds a couple new utils for getting a `handle` by `handleOrDid`. Prev handling was synchronous, new is async, so I added a disabled state to this button.

### DIFF
--- a/src/state/queries/handle.ts
+++ b/src/state/queries/handle.ts
@@ -1,0 +1,34 @@
+import React from 'react'
+import {useMutation} from '@tanstack/react-query'
+
+import {useSession} from '#/state/session'
+
+export function useGetHandle() {
+  const {agent} = useSession()
+
+  return React.useCallback(
+    async (handleOrDid: string) => {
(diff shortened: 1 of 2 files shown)


src/state/queries/handle.ts
```diff
@@ -0,0 +1,27 @@
...
+export function useFetchHandle() {
+  const {agent} = useSession()
+  const queryClient = useQueryClient()
+
+  return React.useCallback(
+    async (handleOrDid: string) => {
+      if (handleOrDid.startsWith('did:')) {
+        const res = await queryClient.fetchQuery({
+          // cache in memory forever, page reload clears
+          staleTime: Infinity,
```
estrattonbailey, line 17: @pfrazee

pfrazee: we should invalidate the cache when the user changes their handle, and might want to manually update the cache any time we use the get profile hook

pfrazee (approved):
Small note

PR #1904, opened 2023-11-14
Fix image order in 4-image post embeds (close #1895)

### LINKED ISSUE #1895: The order in which the four images are displayed in the post is different from the previous version.
**Describe the bug**
The order in which thumbnails are displayed for posts with four images attached is different between bsky.app V1.54 and V1.55 or later.

**To Reproduce**

Sample post URL
https://bsky.app/profile/shoimada.bsky.social/post/3kbroxvwhyo2o

When displayed in the v1.54 version of the web version:
bskyImageOrder 154

If displayed in v1.55 or later:
bskyImageOrder 156

**Expected behavior**

Even for posts like this post, where images are attached with the order of display in thumbnails also in mind, or when considering the enlarged display of one image at a time, it seems natural for the second image to be displayed to the right of the first image.

I would like to see a return to the order in which images were displayed up to the v1.54 version, as the thumbnail display of previously posted images is different.

pfrazee:
# Before

image

# After

CleanShot 2023-11-14 at 12 14 26@2x

### DIFF
--- a/src/view/com/util/images/ImageLayoutGrid.tsx
+++ b/src/view/com/util/images/ImageLayoutGrid.tsx
@@ -69,12 +69,12 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
               <GalleryItem {...props} index={0} imageStyle={styles.image} />
             </View>
             <View style={styles.smallItem}>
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):

gaearon:
oops, i should've tested with different images 🤦

PR #1906, opened 2023-11-14
Desktop user autocomplete search

estrattonbailey:
Just the desktop autocomplete portion

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "expo-system-ui": "~2.4.0",
     "expo-updates": "~0.18.12",
     "fast-text-encoding": "^1.0.6",
+    "fuse.js": "^7.0.0",
     "history": "^5.3.0",
     "js-sha256": "^0.9.0",
     "lande": "^1.0.10",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -106,6 +106,7 @@
     "expo-system-ui": "~2.4.0",
     "expo-updates": "~0.18.12",
     "fast-text-encoding": "^1.0.6",
+    "fuse.js": "^7.0.0",
```
estrattonbailey, line 109: Gives us fuzzy matching for follows in 5kb gzipped

src/view/shell/desktop/Search.tsx
```diff
@@ -1,59 +1,137 @@
...
 import {Text} from 'view/com/util/text/Text'
-import {Trans, msg} from '@lingui/macro'
-import {useLingui} from '@lingui/react'
+import {UserAvatar} from '#/view/com/util/UserAvatar'
+import {useActorSearch} from '#/state/queries/actor-autocomplete'
+import {useModerationOpts} from '#/state/queries/preferences'
 
-export const DesktopSearch = observer(function DesktopSearch() {
-  const store = useStores()
+export function SearchResultCard({
```
estrattonbailey, line 28: `ProfileCard` was doing a lot, I think it's probably better to not use a shared component like that that bloats over time with more config like `noBorder` etc.

pfrazee (approved):

PR #1911, opened 2023-11-15
Autocomplete updates (react-query refactor)

pfrazee:
- Updates the composer's web autocomplete to use the new hook
- Unifies the autocomplete hooks
- Drops the `fuse.js` usage. I did a side-by-side with the results and the fuzzy matching just added mostly-wrong guesses. We should let the backend do that behavior.
- Updates the mobile composer autocomplete to persist results while new ones are loading

### DIFF
--- a/src/view/com/composer/text-input/mobile/Autocomplete.tsx
+++ b/src/view/com/composer/text-input/mobile/Autocomplete.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect} from 'react'
+import React, {useEffect, useRef} from 'react'
 import {Animated, TouchableOpacity, StyleSheet, View} from 'react-native'
 import {observer} from 'mobx-react-lite'
 import {useAnimatedValue} from 'lib/hooks/useAnimatedValue'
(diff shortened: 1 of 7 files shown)


src/view/com/composer/text-input/mobile/Autocomplete.tsx
```diff
@@ -19,7 +21,13 @@ export const Autocomplete = observer(function AutocompleteImpl({
   const positionInterp = useAnimatedValue(0)
   const {getGraphemeString} = useGrapheme()
   const isActive = !!prefix
-  const {data: suggestions} = useActorAutocompleteQuery(prefix)
+  const {data: suggestions, isFetching} = useActorAutocompleteQuery(prefix)
+  const suggestionsRef = useRef<
```
gaearon, line 25: Why is this needed?

gaearon: To rephase, why is the code below not using `suggestions` directly?

estrattonbailey: Looks like it's so it renders the old results while new results are being fetched. I'll add a comment

estrattonbailey: Looks like it's working as intended without though 🤔

estrattonbailey (approved):
LGTM, one question on the ref

src/view/com/composer/text-input/mobile/Autocomplete.tsx
pfrazee: Yep, to show old results while new ones are being fetched

gaearon: I see. The canonical pattern for this looks a bit different (writing to refs during render isn't supported by React) but by the time React adds a warning/error for this we'll probably rewrite this five times anyway so lg

PR #1915, opened 2023-11-16
Shell behaviors update (react-query refactor)

pfrazee:
- Move tick-every-minute into a hook/context
- Create a new global events bus for soft-resets and session events
- Update UIs, push-notifications, and analytics to use new systems

### DIFF
--- a/src/state/shell/tick-every-minute.tsx
+++ b/src/state/shell/tick-every-minute.tsx
@@ -11,7 +11,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       setTick(Date.now())
     }, 60_000)
     return () => clearInterval(i)
-  })
+  }, [])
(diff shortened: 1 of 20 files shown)


src/state/shell/tick-every-minute.tsx
```diff
@@ -0,0 +1,20 @@
...
+const stateContext = React.createContext<StateContext>(0)
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const [tick, setTick] = React.useState(Date.now())
+  React.useEffect(() => {
+    const i = setInterval(() => {
+      setTick(Date.now())
+    }, 60_000)
+    return () => clearInterval(i)
+  })
```
gaearon, line 14: add `[]` here so it doesn't get reset if state update flows through it

gaearon (approved):
dope

PR #1918, opened 2023-11-16
No need to await follows anymore, fix cold load search

estrattonbailey:
- was previously checking for follows before searching so that the prev functionality would work as best it could, but Paul's changes make this unnecessary
- on cold load we should set an initial value for query so that you can see results right away


pfrazee (approved):
Solid

PR #1921, opened 2023-11-16
Improve types

estrattonbailey:
We were [doing this before too,](https://github.com/bluesky-social/social-app/pull/1920/files#diff-e25b9c34f92225fd470d531409295da0a5a027efe4ff901922a25ed76974404bL14) but my new types should be optional since they're not part of the spec


pfrazee (approved):
Cool

PR #1924, opened 2023-11-16
Fix memo params causing tuner to lose state

pfrazee:
The tuner handles de-duplication. If it doesnt maintain its state properly, it can result in duplicate react keys showing in the feed.

I noticed that and tracked down that the useMemo was being too loose with the dependency.


estrattonbailey (approved):
Nice one

PR #1926, opened 2023-11-16
Close active elems (react-query refactor)

pfrazee:


src/state/shell/composer.tsx
```diff
@@ -51,11 +53,17 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         setState(opts)
       },
       closeComposer() {
-        setState(undefined)
+        let wasOpen = false
+        setState(v => {
+          wasOpen = !!v
+          return undefined
+        })
+        return wasOpen
```
gaearon, line 61: This will break in future versions of React, can we just return `state`?

gaearon: I think this can actually even break today too in some cases. React provides no guarantee at all that the updater function would run synchronously during the `setState` call

src/state/modals/index.tsx
```diff
@@ -238,18 +245,22 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
     let totalActiveModals = 0
+    let wasActive = false
     setActiveModals(activeModals => {
+      wasActive = activeModals.length > 0
       activeModals = activeModals.slice(0, -1)
       totalActiveModals = activeModals.length
       return activeModals
     })
     setIsModalActive(totalActiveModals > 0)
+    return wasActive
```
gaearon, line 258: Same, this kind of pattern is not to be trusted — there's no guarantee the function above would have executed by now

pfrazee: Fixed

src/state/shell/composer.tsx
pfrazee: Fixed

gaearon (approved):
cool

PR #1928, opened 2023-11-16
Some me model cleanup

estrattonbailey:


src/view/com/composer/Prompt.tsx
```diff
@@ -3,13 +3,15 @@ import {StyleSheet, TouchableOpacity} from 'react-native'
...
 import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {Trans, msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
+import {useSession} from '#/state/session'
+import {useProfileQuery} from '#/state/queries/profile'
 
 export function ComposePrompt({onPressCompose}: {onPressCompose: () => void}) {
-  const store = useStores()
+  const {currentAccount} = useSession()
+  const {data: profile} = useProfileQuery({did: currentAccount!.did})
```
pfrazee, line 14: This is actually a place where we want to handle the possibility of no session

src/view/shell/bottom-bar/BottomBar.tsx
```diff
@@ -47,6 +49,7 @@ export const BottomBar = observer(function BottomBarImpl({
     useNavigationTabState()
   const numUnreadNotifications = useUnreadNotifications()
   const {footerMinimalShellTransform} = useMinimalShellMode()
+  const {data: profile} = useProfileQuery({did: currentAccount!.did})
```
pfrazee, line 52: Another where we need to handle it

pfrazee (changes requested):
Great - I think we just need to not use `currentAccount!` because that'll reference error. If the profile query doesn't support `enabled: false` on no `did` yet, you can add that to avoid the conditional hook

src/view/com/composer/Prompt.tsx
estrattonbailey: Yep yep, the `useProfileQuery` has the `enabled` check, and `UserAvatar` falls back to `DefaultAvatar` in the case of undefined. That work or were you thinking of something else?

src/view/shell/bottom-bar/BottomBar.tsx
estrattonbailey: Updated this to match the above, will fall back to default avi

src/view/com/composer/Prompt.tsx
estrattonbailey: Tbh I'd rather not rely on those enable checks much if we can help it, but we'll need to for the time being

pfrazee: Wont `currentAccount!.did` reference error if `currentAccount` is undefined?

pfrazee (approved):

PR #1929, opened 2023-11-16
Refactor ChangeHandle modal

estrattonbailey:


src/view/com/modals/ChangeHandle.tsx
```diff
@@ -25,77 +24,71 @@ import {logger} from '#/logger'
...
+import {useSession, useSessionApi, SessionAccount} from '#/state/session'
 
 export const snapPoints = ['100%']
 
-export function Component({onChanged}: {onChanged: () => void}) {
-  const store = useStores()
-  const [error, setError] = useState<string>('')
+export type Props = {onChanged: () => void}
+
+export function Component(props: Props) {
```
estrattonbailey, line 35: Just pulled the loading state up to simplify child components

src/view/com/modals/ChangeHandle.tsx
```diff
@@ -25,77 +24,71 @@ import {logger} from '#/logger'
...
-    return () => {
-      aborted = true
-    }
-  }, [store.agent.service, store.session, retryDescribeTrigger])
+  // TODO refactor verify logic
+  const userDomain = serviceInfo.availableUserDomains?.[0]
+    ? serviceInfo.availableUserDomains[0].includes('bsky.network')
+      ? 'bsky.social'
+      : serviceInfo.availableUserDomains[0]
+    : 'bsky.social'
```
estrattonbailey, line 85: @pfrazee

src/view/com/modals/ChangeHandle.tsx
```diff
@@ -106,13 +99,22 @@ export function Component({onChanged}: {onChanged: () => void}) {
...
   }, [setCustom, isCustom, track])
   const onPressSave = React.useCallback(async () => {
-    setError('')
-    setProcessing(true)
+    if (!userDomain) {
+      logger.error(`ChangeHandle: userDomain is undefined`, {
+        service: serviceInfo,
+      })
+      setError(`The service you've selected has no domains configured.`)
+      return
```
estrattonbailey, line 107: Not sure this will ever happen, but if a PDS returns no domains at least we'll get a log in Sentry and alert the user

src/view/com/modals/ChangeHandle.tsx
pfrazee: Yeah don't fall back. If this happens the server is not configured for assigning user handles (probably misconfigured)

src/view/com/modals/ChangeHandle.tsx
pfrazee: Yep report an error

pfrazee (approved):
Small changes, otherwise looks good

src/state/session/index.tsx
```diff
@@ -14,8 +14,8 @@ export type SessionState = {
   agent: BskyAgent
   isInitialLoad: boolean
   isSwitchingAccounts: boolean
-  accounts: persisted.PersistedAccount[]
-  currentAccount: persisted.PersistedAccount | undefined
+  accounts: SessionAccount[]
+  currentAccount: SessionAccount | undefined
```
estrattonbailey, line 18: Just using semantically correct types, same type though

src/state/session/index.tsx
```diff
@@ -164,8 +164,8 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         throw new Error(`session: createAccount failed to establish a session`)
       }
 
-      const account: persisted.PersistedAccount = {
-        service,
+      const account: SessionAccount = {
+        service: agent.service.toString(),
```
estrattonbailey, line 168: Just using the agent/session as the source of truth for this stuff so that the persisted account never gets out of date

src/view/com/modals/ChangeHandle.tsx
```diff
@@ -25,77 +24,66 @@ import {logger} from '#/logger'
...
-        setError(
-          'Unable to contact your service. Please check your Internet connection.',
-        )
-      },
-    )
-    return () => {
-      aborted = true
-    }
-  }, [store.agent.service, store.session, retryDescribeTrigger])
+  const userDomain = serviceInfo.availableUserDomains?.[0]
```
estrattonbailey, line 80: So at least one needs to be configured, and we don't support multiple right now

pfrazee (approved):

PR #1933, opened 2023-11-16
Add a mutation queue to fix race conditions in toggles

gaearon:
This introduces a helper that lets us run "toggle" mutations (e.g. Follow/Unfollow) without races.

Currently, our approach is to treat each mutation as independent. However, their optimistic updates can clobber each other — e.g. if you press Follow immediately after Unfollow, you can't predict in which order they'll get to the server, or in which order they will come back. Our existing code did not deal with that.

Additionally, our API treats Unfollow/Unblock/Unmute as deletions of a record and requires a record URI. However, if you press Follow immediately followed by Unfollow (haha), it's "too early" to send the Unfollow request because we don't have any record URI for it yet. Our existing code tried to deal with that by guarding event handlers so that you _can't_ Unfollow until Follow comes through — but that wasn't being applied consistently. And even if it was, you want to be able to "undo" an accidental Follow immediately.

In this PR, we take the following approach. The user is expected to be able to press the toggle as much as they like. The changes are applied immediately through our existing optimistic mutation mechanism. We still fire the underlying mutations via RQ. However, we don't "commit" or "rollback" the response immediately.

Instead, if you fire off multiple mutations immediately one after another, they run via a state machine that carries the returned server state through all the queued up actions one by one. This lets us feed the resource URI returned by the Follow action right into the pending Unfollow action. I chose to do it this way instead of reading the current shadow because I don't want to rely on the timing of React rendering (and updating the state) when orchestrating mutations. The state of the queue is reset when the queue is drained. So it only lives for as long as you keep smashing the button. This should reduce the surface area for bugs.

The queue automatically deduplicates pending actions. If you spam the button (e.g. Follow -> Unfollow -> Follow -> Unfollow -> Follow), as soon as it's done with the current request, it will apply at most one request if necessary. E.g. if we're waiting for Follow and the final requested state is also Follow, then there's nothing left to do. But if we're waiting for Follow and you requested Unfollow -> Follow -> Unfollow, the final Unfollow is still necessary — so we will apply just the one last Unfollow, and drop the intermediate actions.

API-wise, I opted to wrap our existing RQ mutations so that, if this abstraction is bad, it should be easy to rip out. At the callsites, I've removed the guards that protect against missing URIs (since the queue handles this now), but kept the overall structure the same. The individual Promises from queued calls resolve as the queue progresses, so we can still show toasts as individual requests come through, even if the entire queue is not drained yet. This provides feedback on slow network. I've ripped out spinners in the UI because now you can press Follow/Unfollow anytime and expect to get to the final state.

For testing, it might be helpful to turn on 3G emulation or to add manual Promise delays.

I only tested the web so far.

**[Review without whitespace](https://github.com/bluesky-social/social-app/pull/1933/files?w=1)**

---

https://github.com/bluesky-social/social-app/assets/810438/4bb4c629-6c73-4457-b8de-6523864b14af

### DIFF
--- a/src/lib/hooks/useToggleMutationQueue.ts
+++ b/src/lib/hooks/useToggleMutationQueue.ts
@@ -11,6 +11,12 @@ type TaskQueue<TServerState> = {
   queuedTask: Task<TServerState> | null
 }
 
+function AbortError() {
+  const e = new Error()
+  e.name = 'AbortError'
+  return e
+}
+
 export function useToggleMutationQueue<TServerState>({
   initialState,
   runMutation,
@@ -47,6 +53,7 @@ export function useToggleMutationQueue<TServerState>({
         queue.queuedTask = null
         if (prevTask?.isOn === nextTask.isOn) {
           // Skip multiple requests to update to the same value in a row.
+          prevTask.reject(new (AbortError as any)())
           continue
         }
         try {
@@ -68,6 +75,9 @@ export function useToggleMutationQueue<TServerState>({
   function queueToggle(isOn: boolean): Promise<TServerState> {
     return new Promise((resolve, reject) => {
       // This is a toggle, so the next queued value can safely replace the queued one.
+      if (queue.queuedTask) {
+        queue.queuedTask.reject(new (AbortError as any)())
+      }
       queue.queuedTask = {isOn, resolve, reject}
       processQueue()
     })
--- a/src/lib/hooks/useToggleMutationQueue.ts
+++ b/src/lib/hooks/useToggleMutationQueue.ts
@@ -1,0 +1,77 @@
+import {useState} from 'react'
+
+type Task<TServerState> = {
+  isOn: boolean
+  resolve: (serverState: TServerState) => void
+  reject: (e: unknown) => void
+}
+
+type TaskQueue<TServerState> = {
+  activeTask: Task<TServerState> | null
+  queuedTask: Task<TServerState> | null
+}
+
+export function useToggleMutationQueue<TServerState>({
+  initialState,
+  runMutation,
+  onSuccess,
+}: {
+  initialState: TServerState
+  runMutation: (
+    prevState: TServerState,
+    nextIsOn: boolean,
+  ) => Promise<TServerState>
+  onSuccess: (finalState: TServerState) => void
+}) {
+  // We use the queue as a mutable object.
+  // This is safe becuase it is not used for rendering.
+  const [queue] = useState<TaskQueue<TServerState>>({
+    activeTask: null,
+    queuedTask: null,
+  })
+
+  async function processQueue() {
+    if (queue.activeTask) {
+      // There is another active processQueue call iterating over tasks.
+      // It will handle any newly added tasks, so we should exit early.
+      return
+    }
+    // To avoid relying on the rendered state, capture it once at the start.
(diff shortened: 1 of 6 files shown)


src/state/queries/profile.ts
```diff
@@ -99,104 +102,294 @@ export function useProfileUpdateMutation() {
...
+  >({
     mutationFn: async ({did}) => {
       return await agent.follow(did)
     },
     onMutate(variables) {
-      // optimstically update
-      updateProfileShadow(variables.did, {
-        followingUri: 'pending',
-      })
+      if (!variables.skipOptimistic) {
```
gaearon, line 171: If we don't want to use `useProfileFollowMutation` standalone then we can just rip these conditions out. However, I think it's plausible they might be useful somewhere else (or maybe the abstraction doesn't work out) so I kept them for now.

src/view/com/auth/onboarding/RecommendedFollowsItem.tsx
```diff
@@ -77,22 +74,15 @@ export function ProfileCard({
...
-      if (
-        profile.viewer?.following &&
-        profile.viewer?.following !== 'pending'
-      ) {
-        await unfollow({did: profile.did, followUri: profile.viewer.following})
-      } else if (
-        !profile.viewer?.following &&
-        profile.viewer?.following !== 'pending'
-      ) {
+      if (profile.viewer?.following) {
```
gaearon, line 81: Haven't tested this codepath. I did test the others.

src/view/com/profile/ProfileHeader.tsx
```diff
@@ -267,22 +249,16 @@ function ProfileHeaderLoaded({
...
       onPressConfirm: async () => {
-        if (!profile.viewer?.blocking) {
-          return
-        }
         try {
-          await unblockMutation.mutateAsync({
-            did: profile.did,
-            blockUri: profile.viewer.blocking,
-          })
+          await queueUnblock()
```
gaearon, line 253: Strictly speaking, didn't have to convert this one since it stays modal.

src/lib/hooks/useToggleMutationQueue.ts
```diff
@@ -0,0 +1,88 @@
...
+        queue.queuedTask = null
+        if (prevTask?.isOn === nextTask.isOn) {
+          // Skip multiple requests to update to the same value in a row.
+          continue
+        }
+        try {
+          // The state received from the server feeds into the next task.
+          // This lets us queue deletions of not-yet-created resources.
+          confirmedState = await runMutation(confirmedState, nextTask.isOn)
+          nextTask.resolve(confirmedState)
```
gaearon, line 56: Normally we resolve these to allow the callers to await on individual calls. However, I don't resolve the actions we've skipped (e.g. when you spam the button, and a bunch are unnecessary). Currently `await`ing them would wait forever. This isn't an issue for our code in practice but I think it *might* cause a memory leak — or at least break expectations. I think a proper way would be to reject with `AbortError` or such, and to change the callsites to _not_ show the test for that. I can fix this but wasn't sure if it would come up.

gaearon: Eh. Fixed in https://github.com/bluesky-social/social-app/pull/1933/commits/ec6e53315428284791cee49cfcc85d4435c6e343.

pfrazee (approved):
Yeah, looks great. +1

PR #1935, opened 2023-11-16
Fix blank feed

gaearon:
The issue was the feed tuner marking all feed items as stale because we've "seen" them. We've "seen" the initial feed fetch was immediately superseded by a second one. That happened because the pinned feeds (which are part of the parameters) were not immediately available, so loading them caused the query to refetch. The fix is to block on the prefs query.

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import {View, ActivityIndicator, StyleSheet} from 'react-native'
 import {useFocusEffect} from '@react-navigation/native'
 import {NativeStackScreenProps, HomeTabNavigatorParams} from 'lib/routes/types'
 import {FeedDescriptor, FeedParams} from '#/state/queries/post-feed'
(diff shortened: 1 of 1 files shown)


src/view/screens/Home.tsx
```diff
@@ -7,44 +8,52 @@ import {FollowingEmptyState} from 'view/com/posts/FollowingEmptyState'
...
   const setDrawerSwipeDisabled = useSetDrawerSwipeDisabled()
-  const pagerRef = React.useRef<PagerRef>(null)
   const [selectedPage, setSelectedPage] = React.useState(0)
-  const [customFeeds, setCustomFeeds] = React.useState<FeedDescriptor[]>([])
-  const {data: preferences} = usePreferencesQuery()
-
-  React.useEffect(() => {
-    if (!preferences?.feeds?.pinned) return
 
+  const customFeeds = React.useMemo(() => {
```
gaearon, line 43: This is a driveby change, it seems okay? Didn't see a reason for this to be an effect

pfrazee (approved):
Great

PR #1938, opened 2023-11-16
Caching heuristics

estrattonbailey:


src/state/queries/app-passwords.ts
```diff
@@ -1,12 +1,15 @@
...
+
+import {useSession} from '#/state/session'
+import {STALE} from '#/state/queries'
 
 export const RQKEY = () => ['app-passwords']
 
 export function useAppPasswordsQuery() {
   const {agent} = useSession()
   return useQuery({
+    staleTime: STALE.INFINITY,
```
pfrazee, line 12: we want infinity for this one?

pfrazee (approved):
Interesting. We'll definitely need to test each of these but let's roll the dice

PR #1941, opened 2023-11-16
Notifs refresh improvement

pfrazee:
1. Lack of visual feedback on notifs refetch was feeling off. This adds a spinner which could use visual improvements but does the job for now.
2. Turns out invalidation wasn't always triggering, you need `refetchType: 'all'` to ensure it happens


gaearon:
Let's not show on initial?

Screenshot 2023-11-16 at 23 34 02

gaearon (approved):
cool but needs fix for first load

pfrazee:
Done

PR #1945, opened 2023-11-17
Fix jumpy modal regression

gaearon:
## Why/What

See comments. We're gonna be using this new Hook in a few more places.

## Before

https://github.com/bluesky-social/social-app/assets/810438/9418e57f-e36e-4871-b5e6-0a41ab1ed462

## After

https://github.com/bluesky-social/social-app/assets/810438/1760048c-1206-42c3-acbf-2df8b78ff160


pfrazee (approved):
Nice

PR #1947, opened 2023-11-17
Add some memoization back

gaearon:
We've lost some memoization when dropping MobX because its `observer` included `memo` inside.
Re-add some `memo`s in a few strategic places.

This should fix (most of) the profile scroll jank. The issue was that we were triggering some re-render up high which also re-rendered _all content in other tabs that were previously open_.

I'll do a closer pass to see what can benefit from memo'ing the most, but this is an easy fix.

## Code style

I opted to wrap *after* definition because the extra indent level has been annoying.
This is a bit less obvious but reads cleaner IMO.

## Before

Screenshot 2023-11-17 at 02 44 15

## After

Screenshot 2023-11-17 at 02 47 22

### DIFF
--- a/src/view/com/posts/Feed.tsx
+++ b/src/view/com/posts/Feed.tsx
@@ -1,4 +1,4 @@
-import React, {MutableRefObject} from 'react'
+import React, {memo, MutableRefObject} from 'react'
 import {
   ActivityIndicator,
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
🔥

PR #1950, opened 2023-11-17
Store reviews

ansh:
Set up `expo-store-reviews` which asks users to rate the app on the App Store or Play Store.

Our top priority is not intrusive to users and disrupt them from their flow of using the app. So, we only remind the users to rate the app when we think they have had a great experience and aren't in the middle of something.

They can easily opt-out of rating the app as well, and we won't remind them again.


surfdude29:
I think this change would help to give a more accurate representation of how actual Bluesky users feel about the app and avoid dissuading new users from downloading. Currently on the UK App Store, for instance, the ratings are not particularly encouraging for someone thinking of trying out Bluesky:
IMG_6288

PR #1951, opened 2023-11-17
Fix some key errors

estrattonbailey:


src/view/com/lists/ProfileLists.tsx
```diff
@@ -75,7 +75,12 @@ export function ProfileLists({
       items = items.concat([EMPTY])
     } else if (data?.pages) {
       for (const page of data?.pages) {
-        items = items.concat(page.lists)
+        items = items.concat(
```
gaearon, line 78: Do we have other places needing a similar fix?

src/view/screens/Search/Search.tsx
```diff
@@ -137,11 +137,11 @@ function SearchScreenSuggestedFollows() {
             ),
         )
       ).flat()
-
-      setSuggestions(
-        // dedupe
-        friendsOfFriends.filter(f => !friends.find(f2 => f.did === f2.did)),
+      const deduped = friendsOfFriends.filter(
+        (f, i) => friendsOfFriends.findIndex(f2 => f.did === f2.did) === i,
```
gaearon, line 141: Can we do `Array.from(new Set(friendsOfFriends))` to avoid loops inside a loop?

src/view/com/lists/ProfileLists.tsx
estrattonbailey: Haven't noticed any yet, just grouped these fixes in a pr for now

src/view/screens/Search/Search.tsx
estrattonbailey: Not with objects, but we could probably do this faster by creating a `Set(did, obj)` and reading values.

gaearon: yeah

gaearon (approved):
cool but let's fix nested loops too

estrattonbailey:
Just gonna merge this and continue in a sep PR

PR #1953, opened 2023-11-17
Do less work during rendering (part 1)

gaearon:
Let's fix a bunch of regressions that crept in with the state refactor.

- We should [avoid `setState` in `useEffect`](https://react.dev/learn/you-might-not-need-an-effect#caching-expensive-calculations) like plague. To avoid recalculations, `useMemo`. 
- If we're calculating something just for the initial state, put `() =>` before it to [avoid running it again](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state).
- Don't recreate a bunch of objects that have the same content every time. Especially when they're being passed down later. Then add some memoization to take advantage of that.

## Before

https://github.com/bluesky-social/social-app/assets/810438/1df1513a-169c-410b-a388-2000067cbc4c

With 6x CPU throttling, scrolling a post thread stalls for two seconds:

Screenshot 2023-11-17 at 17 23 00

## After

https://github.com/bluesky-social/social-app/assets/810438/57613d05-deb2-4f9b-b9de-89cf428b9104

With 6x CPU throttling, scrolling a post thread stalls for 60 milliseconds (~30x faster):

Screenshot 2023-11-17 at 17 21 30

### DIFF
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -1,4 +1,4 @@
-import {useEffect, useState, useCallback, useRef} from 'react'
+import {useEffect, useState, useMemo, useCallback, useRef} from 'react'
 import EventEmitter from 'eventemitter3'
 import {AppBskyFeedDefs} from '@atproto/api'
 import {Shadow} from './types'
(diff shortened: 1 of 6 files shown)


pfrazee (approved):
Nice, good stuff

PR #1954, opened 2023-11-17
Add "Who can reply" controls [WIP]

pfrazee:
Adds this button to the composer:

CleanShot 2023-11-17 at 09 23 40@2x

Which opens this modal:

CleanShot 2023-11-17 at 09 24 02@2x

Which leads to this behavior:

|1|2|3|
|-|-|-|
|CleanShot 2023-12-01 at 16 29 17@2x|Simulator Screenshot - iPhone 15 Pro - 2023-12-01 at 16 32 10|Simulator Screenshot - iPhone 15 Pro - 2023-12-01 at 16 32 25|

Still todo:

- [x] Disable the reply button if not able to reply. This is blocked by backend because `canReply` is only being provided in the threadview, and we need it in all postview viewer states.
- [ ] ~Simulation tests~ - going to add this later as we need to get sim tests working again


pfrazee:
Also worth noting: the `threadgate` record is external to the post record and it should be mutable, so we can add a control for a thread author to change the threadgate after posting.

src/lib/api/index.ts
```diff
@@ -247,6 +251,52 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
...
     }
   }
+
+  try {
+    // TODO: this needs to be batch-created with the post!
+    if (opts.threadgate?.length) {
+      await createThreadgate(agent, res.uri, opts.threadgate)
+    }
+  } catch (e: any) {
+    console.error(`Failed to create threadgate: ${e.toString()}`)
```
estrattonbailey, line 261: Nit: use logger

src/lib/api/index.ts
```diff
@@ -247,6 +251,52 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
       throw e
     }
   }
+
+  try {
+    // TODO: this needs to be batch-created with the post!
```
estrattonbailey, line 256: Todo?

src/lib/api/index.ts
```diff
@@ -247,6 +251,52 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
...
+    if (opts.threadgate?.length) {
+      await createThreadgate(agent, res.uri, opts.threadgate)
+    }
+  } catch (e: any) {
+    console.error(`Failed to create threadgate: ${e.toString()}`)
+    throw new Error(
+      'Post reply-controls failed to be set. Your post was created but anyone can reply to it.',
+    )
```
estrattonbailey, line 264: Where does this show up?

src/view/com/composer/threadgate/ThreadgateBtn.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <TouchableOpacity
+      testID="openReplyGateButton"
+      onPress={onPress}
+      style={styles.button}
+      hitSlop={HITSLOP_10}
+      accessibilityRole="button"
+      accessibilityLabel={_(msg`Who can reply`)}
+      accessibilityHint="">
```
estrattonbailey, line 44: Kinda meh copy

src/view/com/util/post-ctrls/PostCtrls.tsx
```diff
@@ -108,9 +108,16 @@ export function PostCtrls({
     <View style={[styles.ctrls, style]}>
       <TouchableOpacity
         testID="replyBtn"
-        style={[styles.ctrl, !big && styles.ctrlPad, {paddingLeft: 0}]}
+        style={[
+          styles.ctrl,
+          !big && styles.ctrlPad,
+          {paddingLeft: 0},
+          post.viewer?.replyDisabled ? {opacity: 0.5} : undefined,
```
estrattonbailey, line 115: Does this mean that when disabled, there's no visual change on touch/click?

src/lib/api/index.ts
pfrazee: ah hell, yeah this can wait but it needs to be on our near-term todos. the wonky error message you commented on won't be necessary afterwards

src/lib/api/index.ts
pfrazee: in the composer

estrattonbailey (approved):

pfrazee:
Move the button to the top right
Simulator Screenshot - iPhone 15 Pro - 2023-12-10 at 11 48 29

PR #1957, opened 2023-11-17
Optimize notifications rendering

gaearon:
## Before

Scrolling notifications with 6x CPU throttling: 700ms stall.

Screenshot 2023-11-17 at 17 44 07

## After

Scrolling notifications with 6x CPU throttling: 30ms stall.

Screenshot 2023-11-17 at 17 47 15

### DIFF
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -1,4 +1,4 @@
-import React, {useMemo, useState, useEffect} from 'react'
+import React, {memo, useMemo, useState, useEffect} from 'react'
 import {
   Animated,
   TouchableOpacity,
@@ -56,7 +56,7 @@ interface Author {
   moderation: ProfileModeration
 }
 
-export function FeedItem({
+let FeedItem = ({
   item,
   dataUpdatedAt,
   moderationOpts,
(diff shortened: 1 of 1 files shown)


pfrazee (approved):

src/view/com/notifications/FeedItem.tsx
```diff
@@ -56,15 +56,15 @@ interface Author {
   moderation: ProfileModeration
 }
 
-export function FeedItem({
+let FeedItem = ({
```
Hartaithan, line 59: hi, i'm just stranger
but can you explain this approach, please? interesting
why you need to store component in let variable and apply memo later, instead of just wrapping component in memo?

Hartaithan: wow, i doesn't know that every stranger can comments on pr's like that
if i'm not allowed to, please sorry

recursive-beast: In the react docs, [memo](https://react.dev/reference/react/memo) accepts a component to memoize, which is in this case a funcitonal component. You can either store the component's function in a variable then pass it to memo (what is done here), or just inline it in the `memo()` call (what is done in [the docs](https://react.dev/reference/react/memo#memo).

gaearon: It doesn't really matter, I just want to avoid an extra indentation level in the code.

gaearon:
To people getting excited by this PR — this just returns us to the baseline perf we had before a bunch of refactor. So this won't by itself make anything faster, it just avoids us making things slow :)

PR #1959, opened 2023-11-18
Shadow refactoring and improvements

gaearon:
## Part 1

Small refactor to how shadows work. This isn't supposed to *significantly* change anything but might improve perf a bit.

- fa334350fb04d3ede175829cb9a2723155826bc6: This removes the `isShadowed` field from the runtime. This lets us avoid doing an extra object copy for each post (even before it gets shadowed). Theoretically this also might help with monomorphism (avoiding all the methods below from dealing with differently-shaped objects) though in practice probably not. To enforce that you can't can't make a shadow out of thin air, I used [this trick](https://evertpot.com/opaque-ts-types/).
- 0128feefd4012f16984f231235fb74a1d66a95d9: Just preventing creating a few unnecessary objects on every render.
- 67cf7a5824ae8fcdd7274a8949fe836da4d0a9ed: This applies [this technique](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) to adjust state during render instead of in an effect. That allows us to do this without waiting for a whole tree to complete first, i.e. avoids a cascading change. Should be faster.
- f10ed66d5f3983d73b6e43e6003d521f5c152e1e: Process updates to multiple components (if there are multiple) in a batch. The API says `unstable_` but it's actually very stable and will even be the default behavior in the future anyway.

## Part 2

I pushed to the same PR but these are more serious changes.

- cbf813f20810fc6b337af4c57d597b878d3f53e7: Introduces a WeakMap where we track the first time we see a post/profile. This assumes the object travels down unchanged from RQ. The "first seen" timestamp for a given object will never change. The idea is simple: if a mutation was done _after_ we've first seen _that particular_ object, we apply the shadow. Otherwise we don't. This lets us...
- 90af3df2afa461920c4b2af760eed2a0fe790545: Rip out the passing of `dataUpdatedAt` all the way down. This was causing unnecessary re-renders for all the previous pages when the next page is fetched. It also wasn't correct anyway because fetching a new page doesn't make the *previous* pages any less stale.

### Correctness

Liked my own post in one tab, checked it's liked in another tab. Unliked it in the browser (a different device) and verified that PTR in the app also reset it to unliked despite what the shadow said. 

#### Before

Feed item re-render batches are getting larger on scroll:

https://github.com/bluesky-social/social-app/assets/810438/47d75509-3654-4df8-a270-93bb189f98cc

Same for notifications:

https://github.com/bluesky-social/social-app/assets/810438/3996b2d5-146b-43f3-9123-da9ce6c4b15e

#### After

Feed item render batches are bounded to the new pages:

https://github.com/bluesky-social/social-app/assets/810438/5534e0ff-0aee-4c15-b627-94c6ea0850de

Same for notifications:

https://github.com/bluesky-social/social-app/assets/810438/597f8308-8339-43c5-8946-d98b8a7346e0

### DIFF
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -1,6 +1,7 @@
 import {useEffect, useState, useMemo, useCallback} from 'react'
 import EventEmitter from 'eventemitter3'
 import {AppBskyFeedDefs} from '@atproto/api'
+import {batchedUpdates} from '#/lib/batchedUpdates'
 import {Shadow, castAsShadow} from './types'
 export type {Shadow} from './types'
 
@@ -64,8 +65,11 @@ export function usePostShadow(
 }
 
 export function updatePostShadow(uri: string, value: Partial<PostShadow>) {
-  emitter.emit(uri, value)
+  batchedUpdates(() => {
+    emitter.emit(uri, value)
+  })
 }
+
 function fromPost(post: AppBskyFeedDefs.PostView): PostShadow {
   return {
     likeUri: post.viewer?.like,
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -25,10 +25,10 @@ export function usePostShadow(
   post: AppBskyFeedDefs.PostView,
   ifAfterTS: number,
 ): Shadow<AppBskyFeedDefs.PostView> | typeof POST_TOMBSTONE {
-  const [state, setState] = useState<CacheEntry>({
+  const [state, setState] = useState<CacheEntry>(() => ({
     ts: Date.now(),
     value: fromPost(post),
-  })
+  }))
   const firstRun = useRef(true)
 
   const onUpdate = useCallback(
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -1,7 +1,7 @@
 import {useEffect, useState, useMemo, useCallback, useRef} from 'react'
 import EventEmitter from 'eventemitter3'
 import {AppBskyFeedDefs} from '@atproto/api'
-import {Shadow} from './types'
+import {Shadow, castAsShadow} from './types'
 export type {Shadow} from './types'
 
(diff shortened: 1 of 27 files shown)


gaearon:
Actually wait, seeing some (maybe?..) issues. Debugging.

gaearon:
Let's wait with this until I figure out what's going on.

gaearon:
I couldn't reproduce what I thought was a bug, so I'd say this is ready for review after all.

The bug I was seeing was some feed items failing to re-render on Pull To Refresh despite new data coming in. I don't see that happening anymore and I don't know how to reproduce this again. It also doesn't seem related to my change per se so I'd say let's go ahead with this, and if the issue comes up again, we'll try to narrow it down.

src/state/cache/post-shadow.ts
```diff
@@ -21,15 +22,36 @@ interface CacheEntry {
...
-  const firstRun = useRef(true)
+  }))
+
+  const [prevPost, setPrevPost] = useState(post)
+  if (post !== prevPost) {
+    // if we got a new prop, assume it's fresher
+    // than whatever shadow state we accumulated
+    setPrevPost(post)
+    setState({
+      ts: postSeenTS,
```
estrattonbailey, line 51: Profile shadow uses the getter for this value. Need that here too?

src/state/cache/types.ts
```diff
@@ -1 +1,7 @@
-export type Shadow<T> = T & {isShadowed: true}
+// This isn't a real property, but it prevents T being compatible with Shadow<T>.
+declare const shadowTag: unique symbol
+export type Shadow<T> = T & {[shadowTag]: true}
+
+export function castAsShadow<T>(value: T): Shadow<T> {
+  return value as any as Shadow<T>
+}
```
estrattonbailey, line 7: Rad

estrattonbailey (approved):
Fire. Just our convo in Slack, but this LGTM overall

PR #1964, opened 2023-11-20
PWI Base

estrattonbailey:
View with [no whitespace](https://github.com/bluesky-social/social-app/pull/1964/files?diff=split&w=1)


src/state/queries/preferences/index.ts
```diff
@@ -23,63 +24,71 @@ export * from '#/state/queries/preferences/types'
...
-          ...(res.feedViewPrefs.home || {}),
-        },
-        threadViewPrefs: {
-          ...DEFAULT_THREAD_VIEW_PREFS,
-          ...(res.threadViewPrefs ?? {}),
-        },
-        userAge: res.birthDate ? getAge(res.birthDate) : undefined,
+      if (currentAccount?.did === undefined) {
+        return DEFAULT_PREFERENCES
+      } else {
```
estrattonbailey, line 42: This is the key design choice here. If no user, it returns our default settings, which should be restrictive.

src/state/queries/preferences/moderation.ts
```diff
@@ -26,9 +26,9 @@ export const DEFAULT_LABEL_PREFERENCES: Record<
...
   nsfw: 'hide',
-  nudity: 'warn',
-  suggestive: 'warn',
-  gore: 'warn',
+  nudity: 'hide',
+  suggestive: 'hide',
+  gore: 'hide',
   hate: 'hide',
   spam: 'hide',
   impersonation: 'hide',
```
estrattonbailey, line 34: Now defaulting to hiding everything

src/state/queries/preferences/index.ts
```diff
@@ -23,63 +24,67 @@ export * from '#/state/queries/preferences/types'
...
-          ...DEFAULT_THREAD_VIEW_PREFS,
-          ...(res.threadViewPrefs ?? {}),
-        },
-        userAge: res.birthDate ? getAge(res.birthDate) : undefined,
+      const agent = getAgent()
+
+      if (agent.session?.did === undefined) {
+        return DEFAULT_PREFERENCES
+      } else {
+        const res = await agent.getPreferences()
```
estrattonbailey, line 39: This is the key design choice here. If no user, it returns our default settings, which should be restrictive.

src/state/queries/preferences/types.ts
```diff
@@ -43,7 +43,10 @@ export type UsePreferencesQueryResponse = Omit<
   }
 }
 
-export type ThreadViewPreferences = Omit<BskyThreadViewPreference, 'sort'> & {
+export type ThreadViewPreferences = Pick<
+  BskyThreadViewPreference,
+  'prioritizeFollowedUsers'
+> & {
```
estrattonbailey, line 49: Fixing a type error

src/state/queries/preferences/index.ts
```diff
@@ -139,7 +144,7 @@ export function usePreferencesSetAdultContentMutation() {
       await getAgent().setAdultContentEnabled(enabled)
       // triggers a refetch
       await queryClient.invalidateQueries({
-        queryKey: usePreferencesQueryKey,
+        queryKey: preferencesQueryKey,
```
estrattonbailey, line 147: The `use*` prefix was confusing eslint

src/view/com/auth/SplashScreen.tsx
```diff
@@ -17,9 +28,32 @@ export const SplashScreen = ({
...
+          onPress={() => setShowLoggedOut(false)}>
+          <FontAwesomeIcon
+            icon="x"
+            size={24}
+            style={{
+              color: 'white',
+            }}
+          />
+        </Pressable>
+      )}
```
estrattonbailey, line 55: Adds a way to return to the previous screen if you arrived at this screen by attempting an "authed" action. Same on mobile.

Screenshot 2023-11-20 at 3 07 32 PM

src/view/com/auth/SplashScreen.tsx
```diff
@@ -17,9 +28,32 @@ export const SplashScreen = ({
 }) => {
   const pal = usePalette('default')
   const {_} = useLingui()
+  const {showLoggedOut} = useLoggedOutView()
+  const {setShowLoggedOut} = useLoggedOutViewControls()
```
gaearon, line 32: Seems wrong that `SplashScreen` "knows" about these if the logic to show the splash screen _for this particular reason_ is inside `withAuthRequired`. I suggest that the `SplashScreen` takes an `onDismiss` prop. If `onDismiss != null`, show the close button which calls `onDismiss`. Then we can hoist the "show logged out" logic somewhere above — maybe to `withAuthRequired` (which can pass it down to `LoggedOut` -> `SplashScreen`).

src/view/com/util/post-ctrls/PostCtrls.tsx
```diff
@@ -107,7 +109,7 @@ export function PostCtrls({
       <TouchableOpacity
         testID="replyBtn"
         style={[styles.ctrl, !big && styles.ctrlPad, {paddingLeft: 0}]}
-        onPress={onPressReply}
+        onPress={authedMethod(onPressReply)}
```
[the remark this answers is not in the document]
gaearon: Another option to consider:

```js
if (ensureAuth()) {
  // ...
}
```

Allows calling it standalone like `ensureAuth()` without the callback, similar to what you have for the report button.

gaearon (approved):
lgtm for a start

src/state/queries/preferences/const.ts
```diff
@@ -25,3 +26,26 @@ export const DEFAULT_PROD_FEEDS = {
   pinned: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
   saved: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
 }
+
+export const DEFAULT_PREFERENCES: UsePreferencesQueryResponse = {
+  birthDate: new Date('2022-11-17'), // first post on network
```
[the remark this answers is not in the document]
estrattonbailey: This should actually be a todo, waiting to hear back on how we want to handle this. I'll add a TODO to return to this.

src/view/com/util/post-ctrls/PostCtrls.tsx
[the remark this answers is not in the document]
estrattonbailey: Sure can do 👍

PR #1965, opened 2023-11-20
Add dev-only button to clear storage

estrattonbailey:
Helpful during dev and testing the new refactor

### DIFF
--- a/src/state/persisted/store.ts
+++ b/src/state/persisted/store.ts
@@ -16,3 +16,9 @@ export async function read(): Promise<Schema | undefined> {
     return objData
   }
 }
+
+export async function clear() {
+  await AsyncStorage.removeItem(BSKY_STORAGE, e => {
+    if (e) console.error(e)
+  })
+}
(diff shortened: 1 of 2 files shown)


src/state/persisted/legacy.ts
```diff
@@ -128,10 +128,18 @@ export async function migrate() {
...
     }
   } catch (e) {
     logger.error('persisted state: error migrating legacy storage', {
       error: String(e),
     })
   }
 }
+
+export async function clearLegacyStorage() {
+  await AsyncStorage.removeItem(DEPRECATED_ROOT_STATE_STORAGE_KEY, e => {
```
gaearon, line 142: Why both await and a callback? I think they're supposed to be used as alternatives. Let's just await + try/catch?

src/state/persisted/store.ts
```diff
@@ -16,3 +16,9 @@ export async function read(): Promise<Schema | undefined> {
     return objData
   }
 }
+
+export async function clear() {
+  await AsyncStorage.removeItem(BSKY_STORAGE, e => {
```
gaearon, line 21: same

gaearon (approved):
lg, nits

src/state/persisted/legacy.ts
estrattonbailey: Ooof yeah, just a remnant. My brain was not on yesterday 😅

PR #1971, opened 2023-11-22
Allow dynamic locale imports for web

ansh:
Doesn't work on native.

### DIFF
--- a/src/locale/i18nProvider.web.tsx
+++ b/src/locale/i18nProvider.web.tsx
@@ -1,0 +1,9 @@
+import React from 'react'
+import {I18nProvider as DefaultI18nProvider} from '@lingui/react'
+import {i18n} from '@lingui/core'
+import {useLocaleLanguage} from './i18n.web'
+
+export default function I18nProvider({children}: {children: React.ReactNode}) {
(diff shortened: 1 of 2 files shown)


src/locale/i18nProvider.web.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import {I18nProvider as DefaultI18nProvider} from '@lingui/react'
+import {i18n} from '@lingui/core'
+import {useLocaleLanguage} from './i18n.web'
+
```
gaearon, line 5: This file doesn't need to be forked right? The bundler should already pick up the right version of `./i18n` from the original one.

gaearon (approved):
let's remove the unnecessary file but lg

src/locale/i18nProvider.web.tsx
ansh: It was throwing errors for me in native because when the bundler was compiling the `async import` it would throw an error

ansh: Actually sorry -- you're right. I got confused.

PR #1988, opened 2023-11-24
[PWI] Fall back for non-public feeds

estrattonbailey:
Screenshot 2023-11-24 at 2 45 32 PM

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -160,6 +160,35 @@ export function useFeedSourceInfoQuery({uri}: {uri: string}) {
   })
 }
 
+export const isFeedPublicQueryKey = ({uri}: {uri: string}) => [
+  'isFeedPublic',
+  uri,
+]
+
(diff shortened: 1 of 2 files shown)


src/state/queries/feed.ts
```diff
@@ -160,6 +160,35 @@ export function useFeedSourceInfoQuery({uri}: {uri: string}) {
...
+          feed: uri,
+          limit: 1,
+        })
+        return Boolean(res.data.feed)
+      } catch (e: any) {
+        const msg = e.toString() as string
+
+        if (msg.includes('missing jwt')) {
+          return false
+        }
```
estrattonbailey, line 184: This is what we're getting now, we'll need to review this [when we get backend merged](https://github.com/bluesky-social/atproto/pull/1884/files#diff-fb75fb08331538c6b23a0496087b1fb92ee1a1c3cd801a938d99ffcf401b7f74R13).

gaearon (approved):

PR #1976, opened 2023-11-22
[PWI] Fix splash screen

estrattonbailey:

PR #1989, opened 2023-11-24
PWI: Refactor Shell

gaearon:
Alternative to https://github.com/bluesky-social/social-app/pull/1976.

Some of the routes need to be completely public (like Profile). Other routes (like Settings) need to show a "log in" shim when you're logged out. The "log in" shim has a different layout from a normal screen (e.g. it should have no rails or bottom bar). However, the code that determines whether to render those shell elements currently lives outside of the route tree, so it doesn't *know* whether we're showing a protected route--and so it can't determine which layout to use.

One way to solve this could be to move the rails into the route screens themselves. However, that would result in resetting the rails state whenever you navigate, which is bad. So this doesn't work for us.

We could try having the route itself somehow affect the layout outside of it, but that breaks top-down data flow. Alternatively, we could try to hoist the route configuration upwards, which is what https://github.com/bluesky-social/social-app/pull/1976 was trying to do. But maintaining another config is annoying. There are too many places (shell, HOC, route config) that have to coordinate to stay in sync.

In this PR, we implement a different solution. We are using the same approach that React Navigation itself uses to maintain state in the tab bar — i.e. we create *our own custom navigator* (it's [surprisingly easy](https://reactnavigation.org/docs/custom-navigators/#extending-navigators)). This gives us access to the route info, let us have precise control over what's being rendered, and lets us show something else instead.

Note I'm basing this navigator on the Stack navigator, which is what we already use on the web. I'm not fully sure why we don't use the Tab navigator for tabs on the web, but I decided not to mess with that, so I follow the existing structure.

Concretely, this fixes the issue we have on main where the rails "aren't aware" that we're on a protected route.

Here's how this works.

- https://github.com/bluesky-social/social-app/commit/f06aa7e67518f3432e72dc176f824a427f147b03: First, we vendor `createNativeStackNavigator` from the upstream package. It's very short.

- https://github.com/bluesky-social/social-app/commit/ba3e33c50e5629c8cd81d8a0404a483c8dcc781b: Remove all logic in `withAuthRequired`. We'll be pulling it upwards into our custom navigator.

- https://github.com/bluesky-social/social-app/commit/eb829cf5827c60d54ce128f39f20b2b3b3b2c3ff: Add a new `requireAuth: boolean` option to our route `options`, defaulting to `false`. Make our custom navigator render `<LoggedOut />` if we try to render such a route with no active session.

- https://github.com/bluesky-social/social-app/commit/e2cab7b926df715f9e010c2249bcbf1d23de8b48: Move the rail and the bottom bar into the navigator which now "knows" whether the active route is protected.

- https://github.com/bluesky-social/social-app/commit/2eb3769aedb45c12ee444a203011a117a9ba321b: Found a simpler way to do this.

- https://github.com/bluesky-social/social-app/commit/451f218557bbc1a22abd298b15cbd45c7353d617: Adds the remaining bit — the dimissable version of the shim.

- https://github.com/bluesky-social/social-app/commit/4364afa711b618927618e0a7498b7a49da7bda7f: We can now delete `withAuthRequired` and all its usages. **This commit looks large but it's just indentation.**

- https://github.com/bluesky-social/social-app/commit/deea4005c8b3f276c50b214ad084d957bd240eee: Reset the app state on session change. This will probably fix a few other things too but I need this so that the "is showing logged out shim" state gets auto-reset on login.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/1989/files?w=1)


estrattonbailey (approved):
Fire

PR #1993, opened 2023-11-26
Use node 18 for dev builds

estrattonbailey:

PR #2000, opened 2023-11-26
Create build v1.57

ansh:
In favor of #1993


estrattonbailey (approved):
We clear to upgrade the image size? If it speeds up I'm in favor

ansh:
@estrattonbailey its the same image size, expo just changed the name so this fixes that

PR #2013, opened 2023-11-28
Guard against following tab stub being used un-authed

estrattonbailey:


src/view/com/pager/FeedsTabBar.web.tsx
```diff
@@ -81,9 +81,10 @@ function FeedsTabBarTablet(
 ) {
   const feeds = usePinnedFeedsInfos()
   const pal = usePalette('default')
+  const {hasSession} = useSession()
   const {headerMinimalShellTransform} = useMinimalShellMode()
   const {headerHeight} = useShellLayout()
-  const items = feeds.map(f => f.displayName)
+  const items = hasSession ? feeds.map(f => f.displayName) : []
```
estrattonbailey, line 87: This file isn't rendered without a session, but added a guard here anyway in case we ever change that.

PR #2017, opened 2023-11-28
Refactor the notifications to cache and reuse results from the unread-notifs checks

pfrazee:
This PR:

- Improves the performance of loading notifs (by using a background cache)
- Fixes an issue on mobile where the feed would jump around while scrolling (by removing the loading spinner I added, which is no longer needed anyway)

The way this works is, it ties together the unread-notif syncing API with the feed query. The unread-notif checker has to periodically fetch a page to count unreads, so this just holds on to that first page to use in the notifs-feed query.

This video shows the performance. It could use a little more visual feedback about the loading state, as well as some better handling of some waterfall-effect behaviors, but it's still much snappier than before:

https://github.com/bluesky-social/social-app/assets/1270099/60b3d4e2-28d2-4814-8d08-ffc717cd12c5

### DIFF
--- a/src/state/queries/notifications/feed.ts
+++ b/src/state/queries/notifications/feed.ts
@@ -1,11 +1,22 @@
-import {
-  AppBskyFeedDefs,
-  AppBskyFeedPost,
-  AppBskyFeedRepost,
-  AppBskyFeedLike,
-  AppBskyNotificationListNotifications,
-} from '@atproto/api'
-import chunk from 'lodash.chunk'
+/**
+ * NOTE
+ * The ./unread.ts API:
+ *
+ * - Provides a `checkUnread()` function to sync with the server,
+ * - Periodically calls `checkUnread()`, and
+ * - Caches the first page of notifications.
+ *
+ * IMPORTANT: This query uses ./unread.ts's cache as its first page,
+ * IMPORTANT: which means the cache-freshness of this query is driven by the unread API.
+ *
+ * Follow these rules:
+ *
+ * 1. Call `checkUnread()` if you want to fetch latest in the background.
+ * 2. Call `checkUnread({invalidate: true})` if you want latest to sync into this query's results immediately.
+ * 3. Don't call this query's `refetch()` if you're trying to sync latest; call `checkUnread()` instead.
+ */
+
+import {AppBskyFeedDefs} from '@atproto/api'
 import {
   useInfiniteQuery,
   InfiniteData,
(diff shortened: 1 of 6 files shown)


src/state/queries/notifications/feed.ts
```diff
@@ -1,63 +1,50 @@
...
+import {useUnreadNotificationsApi} from './unread'
+import {fetchPage} from './util'
+import {FeedPage} from './types'
 import {useMutedThreads} from '#/state/muted-threads'
-import {precacheProfile as precacheResolvedUri} from '../resolve-uri'
 
-const GROUPABLE_REASONS = ['like', 'repost', 'follow']
+export type {NotificationType, FeedNotification, FeedPage} from './types'
+
 const PAGE_SIZE = 30
```
estrattonbailey, line 35: This page size and the unread page size should match, right?

src/state/queries/notifications/unread.tsx
```diff
@@ -46,61 +63,88 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   // listen for broadcasts
   React.useEffect(() => {
     const listener = ({data}: MessageEvent) => {
-      lastSyncRef.current = new Date()
+      cacheRef.current = {
+        sessDid: currentAccount?.did || '',
+        syncedAt: new Date(),
+        data: undefined,
+      }
```
estrattonbailey, line 70: What do you think about invalidating the query here too if the last sync is behind in this tab?

src/state/queries/notifications/unread.tsx
```diff
@@ -46,61 +63,88 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+            sessDid: currentAccount?.did || '',
+            data: page,
+            syncedAt: !lastIndexed || now > lastIndexed ? now : lastIndexed,
+          }
+
+          // update & broadcast
+          setNumUnread(unreadCountStr)
+          if (invalidate) {
+            queryClient.invalidateQueries({queryKey: RQKEY_NOTIFS()})
+          }
```
estrattonbailey, line 132: If we kept tabs in sync, could we only invalidate IF we're behind the latest unread? It would mean no feedback to the user, but saves a useless load state 🤔

src/state/queries/notifications/unread.tsx
```diff
@@ -46,61 +63,88 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
-
-        // update & broadcast
-        setNumUnread(num)
-        broadcast.postMessage({event: num})
+      getCachedUnreadPage() {
+        // return cached page if was for the current user
+        // (protects against session changes serving data from the past session)
+        if (cacheRef.current.sessDid === currentAccount?.did) {
+          return cacheRef.current.data
+        }
```
estrattonbailey, line 144: When does this happen? If it doesn't match, should we trash the cache and start fresh?

estrattonbailey (approved):
Took me a minute to grok this but I think it's Great™

src/state/queries/notifications/feed.ts
pfrazee: Yeah I ask for 40 in the unread because I'm trying to detect 30+, and I give some headroom for filtering

src/state/queries/notifications/unread.tsx
pfrazee: Might trigger more work than we need, but worth considering. Let's see how things feel

src/state/queries/notifications/unread.tsx
pfrazee: Yeah I'm just not sold on keeping the tabs' render view in sync yet

src/state/queries/notifications/unread.tsx
pfrazee: We currently trash all RQ caches but this cache is outside of RQ. I went with this approach instead of adding an imperative clear() call. Both approaches seem fine to me?

PR #2023, opened 2023-11-29
[WIP] Shadow Queue

gaearon:
Builds on top of https://github.com/bluesky-social/social-app/pull/2022.

This is a draft of an approach that attempts to fix existing bugs with the shadow logic.

## The Problem

Our current approach to optimistic mutations (with local "shadowing" of the query cache) is not very reliable because it only works for syncing views that are currently on the screen (or at least in the rendered part of the navigation stack). On the web, this breaks down on most push/pop navigations because our web router doesn't keep the entries below always rendered. On *both* mobile and web, this breaks on scenarios where you *push* a new view that contains a post that's supposed to be shadowed — our shadowing logic only works for *existing* views but not *newly mounted* ones.

For example, here's how this breaks on the web:

1. Push post
2. Like post
3. Pop to feed (expected: liked, actual: not liked)
4. Push post again (expected: liked, actual: not liked)

And here's how it breaks on mobile:

1. Push post
2. Like post
3. Pop to feed
4. Push post again (expected: liked, actual: not liked)

The exact outcome is different but the core issue is the same. We don't keep track of active shadow for newly mounted views, so they're unable to take the already committed mutations into account without an explicit refetch.

## The Proposal

This is a draft so it's a bit messy. But it's also a slightly messy solution. I'd like to give it a try though.

- There is a new concept of a "queue". A queue is a list of partial updates. Each partial update has a timestamp.
- To compute a shadow, we go over the queue and apply only the mutations *after* the post's "seen" timestamp.
- There is a top-level Map from a URI to a "record".
  - A record represents a post that's currently rendered on the screen *somewhere*, possibly more than in one place.
  - A record is refcounted, like the old event emitter approach. It gets deleted when all components for this post unmount.
  - A record contains references to two things. The listener array and the queue for this post. This enables coordination.
- By itself, this still wouldn't be enough to let components communicate with a gap in their lifetime (like push on the web).
  - This is why we also add a WeakMap from a post to a queue.
    - This adds an extra layer: we associate every post with the queue for it forever, so we can "resume" with the same queue.

So effectively the lifetime of the queue is as long as there's either components rendering these posts *or* something holds onto the old post objects (so that the queue can be "brought back from the dead" via the WeakMap when the posts mount).

In theory there is a concern that a queue for some post can grow long. This seems not super likely to me (why would you interact with some post so many times?) but in principle it can happen. Then it's not great that we keep the entire mutation queue forever. One way to address it would be to "prune" the beginning of the queue when we associate it with some post for the first time. I've made it a linked list to make this easy but I have not actually implemented this part.

This might be a bit over-engineered so I'm asking for feedback first, but I do like that it seems to cover known cases well.

## Demo

This appears to fix all the scenarios I've observed so far.

https://github.com/bluesky-social/social-app/assets/810438/9b267b24-b72b-4e1f-af6d-8e31f318cfee

https://github.com/bluesky-social/social-app/assets/810438/d0629597-8c92-47b7-b27e-8d0e25e570e2

## TODO

- [ ] Add types
- [ ] Add back memoization
- [ ] Fix lint
- [ ] (Tentative) Prune the queue
- [ ] Do this for profile shadow cache as well
- [ ] Maybe extract a shared helper so I can stop copy pasting this

### DIFF
--- a/src/lib/react-query.ts
+++ b/src/lib/react-query.ts
@@ -8,6 +8,10 @@ export const queryClient = new QueryClient({
       // so we NEVER want to enable this
       // -prf
       refetchOnWindowFocus: false,
+      // Structural sharing between responses makes it impossible to rely on
+      // "first seen" timestamps on objects to determine if they're fresh.
+      // Disable this optimization so that we can rely on "first seen" timestamps.
+      structuralSharing: false,
     },
   },
 })
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -36,76 +36,39 @@ function getFirstSeenTS(post: AppBskyFeedDefs.PostView): number {
 export function usePostShadow(
   post: AppBskyFeedDefs.PostView,
 ): Shadow<AppBskyFeedDefs.PostView> | typeof POST_TOMBSTONE {
-  const postSeenTS = getFirstSeenTS(post)
-  const [state, setState] = useState<CacheEntry>(() => ({
-    ts: postSeenTS,
-    value: fromPost(post),
-  }))
-
-  const [prevPost, setPrevPost] = useState(post)
-  if (post !== prevPost) {
-    // if we got a new prop, assume it's fresher
-    // than whatever shadow state we accumulated
-    setPrevPost(post)
-    setState({
-      ts: postSeenTS,
-      value: fromPost(post),
-    })
-  }
-
-  const onUpdate = useCallback(
-    (value: Partial<PostShadow>) => {
-      setState(s => ({ts: Date.now(), value: {...s.value, ...value}}))
-    },
-    [setState],
-  )
-
-  // react to shadow updates
   useEffect(() => {
-    emitter.addListener(post.uri, onUpdate)
-    return () => {
-      emitter.removeListener(post.uri, onUpdate)
+    function onChange() {
+      // TODOO
     }
-  }, [post.uri, onUpdate])
 
-  return useMemo(() => {
-    return state.ts > postSeenTS
-      ? mergeShadow(post, state.value)
-      : castAsShadow(post)
-  }, [post, state, postSeenTS])
+    const unsub = addListener(post.uri, onChange)
+    return unsub
+  }, [post.uri])
+
+  return post
 }
+
+function addListener(uri, onChange) {
+  let record
+  if (map.has(uri)) {
+    record = map.get(uri)
+  } else {
+    record = {listeners: []}
+    map.set(uri, record)
+  }
+  record.listeners.push(onChange)
+  return () => {
+    record.listeners = record.listeners.filter(l => l !== onChange)
+    if (record.listeners.length === 0) {
+      map.delete(uri)
+    }
+  }
+}
+
+const map = new Map()
+
+console.log(map)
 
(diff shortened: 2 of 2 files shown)


gaearon:
Screenshot 2023-11-29 at 04 07 33

gaearon:
Why two Maps?

- I need `Map<URI, { queue, listeners }>` because we need a source of truth for the queue (and the listeners) between all currently mounted versions of the same post. Even if they use different post objects (like different mounted tabs). So using a post as a key wouldn't be enough. Since this Map has to be keyed by URIs, we also can't have it grow forever. So this is why it's refcounted. This would mean losing the queue when this post's components unmount, which is why...
- I need `WeakMap<Post, queue>` so that *even* if all components for some post unmount, if we later mount the same post object (like a tab that was removed but later gets readded with cached data), we can recover the associated queue (since the data is stale!) It has to be a `WeakMap` because the only notion of the lifetime here is "as long as those post objects are alive somewhere". It can't be keyed by URI because then we'd never be able to remove it.

pfrazee:
Yeah I think we can simplify this. I did some debugging and this is what I think is going on:

1. The shadow cache isn't getting reliably synced into all of the shadow "subscribers".
2. We're not pulling latest from the server.

I saw signs of (1) with some logging. It's hard to isolate exactly why it's happening because the logic is complex, but I suspect that your queue is effective because it's maintaining persistent state outside of the hook lifecycles. I replaced the shadow hook with this and it fixed all visible issues in the web client when using only one tab:

```typescript
const _cache = new Map()

export function usePostShadow(
  post: AppBskyFeedDefs.PostView,
): Shadow<AppBskyFeedDefs.PostView> | typeof POST_TOMBSTONE {
  const [_, trigger] = useState({})

  // react to shadow updates
  useEffect(() => {
    const onUpdate = () => trigger({})
    emitter.addListener(post.uri, onUpdate)
    return () => {
      emitter.removeListener(post.uri, onUpdate)
    }
  }, [post.uri, trigger])

  const cache = _cache.get(post.uri)
  return cache ? mergeShadow(post, cache) : castAsShadow(post)
}

export function updatePostShadow(uri: string, value: Partial<PostShadow>) {
  // update cache
  let curr = _cache.get(uri)
  if (curr) {
    value = {...curr, ...value}
  }
  _cache.set(uri, value)

  // trigger re-renders
  emitter.emit(uri, value)
}
```

So that solves (1). This leaves (2) the issue of actions in other tabs not being introduced -- aka, becoming stale from the server. To prove this bit to myself, I just tweaked the post-feed and post-thread queries as follows:

|Post-feed|Post-thread|
|-|-|
|CleanShot 2023-11-28 at 22 24 26@2x|CleanShot 2023-11-28 at 22 24 50@2x|

|Post-feed|Post-thread|
|-|-|
|CleanShot 2023-11-28 at 22 25 24@2x|CleanShot 2023-11-28 at 22 25 57@2x|

And that solves (2).

So from here I think the questions are:

1. How do we maintain a well-synced (aka "global") shadow cache that GCs in the way we need it to? an LRU?
2. Can we get the weakmap of timestamps to compare correctly against the shadow cache time, or do we need to push into the shadow cache on fetch like I did in this example? Or a secret third option?

gaearon:
https://github.com/bluesky-social/social-app/pull/2045

PR #2026, opened 2023-11-29
Get more rigorous about getAgent() consistency

pfrazee:
The `getAgent()` is a global method that gives the current session's agent. We want to make sure to never hold onto its return value because that would enable stale agent usage. We've seen a few surprise "missing jwt" errors that suggest that stale agents are being used, so this PR gets a bit more rigorous about it.

### DIFF
--- a/src/state/queries/notifications/unread.tsx
+++ b/src/state/queries/notifications/unread.tsx
@@ -70,12 +70,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       },
 
       async checkUnread() {
-        const agent = getAgent()
-
-        if (!agent.session) return
+        if (!getAgent().session) return
 
         // count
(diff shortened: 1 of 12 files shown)


src/state/queries/notifications/unread.tsx
```diff
@@ -70,12 +70,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       },
 
       async checkUnread() {
-        const agent = getAgent()
-
-        if (!agent.session) return
+        if (!getAgent().session) return
```
estrattonbailey, line 73: Seems like a code smell now, right? Ideally we render this unnecessary 🤔 maybe it's already not needed?

estrattonbailey (approved):
Yep I think this is a good move, makes use of this much easier to reason about

PR #2032, opened 2023-11-29
Improve feed reordering with optimistic updates

estrattonbailey:
Opted not to include pinning or saving in the optimistic updates since saving is within `FeedSourceCard` and overall it results in a lot of state to manage.

### DIFF
--- a/src/view/com/feeds/FeedSourceCard.tsx
+++ b/src/view/com/feeds/FeedSourceCard.tsx
@@ -1,5 +1,12 @@
 import React from 'react'
-import {Pressable, StyleProp, StyleSheet, View, ViewStyle} from 'react-native'
+import {
+  Pressable,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+  ActivityIndicator,
+} from 'react-native'
 import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {Text} from '../util/text/Text'
 import {RichText} from '../util/text/RichText'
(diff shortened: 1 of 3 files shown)


src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -23,24 +23,33 @@ import {
...
   const {data: feed} = useFeedSourceInfoQuery({uri: feedUri})
 
-  if (!feed || !preferences) return null
+  if (!feed || !preferences) {
+    return LoadingComponent ? (
+      LoadingComponent
+    ) : (
+      <FeedLoadingPlaceholder style={{flex: 1}} />
+    )
+  }
```
estrattonbailey, line 52: Looks better on Feeds page and embedded in posts too

src/view/screens/SavedFeeds.tsx
```diff
@@ -151,22 +175,30 @@ export function SavedFeeds({}: Props) {
...
-  const {isPending: isMovePending, mutateAsync: setSavedFeeds} =
-    useSetSaveFeedsMutation()
+  const isPending = isPinPending || isUnpinPending
 
   const onTogglePinned = React.useCallback(async () => {
     Haptics.default()
 
     try {
+      resetSaveFeedsMutationState()
+
```
estrattonbailey, line 201: Clears optimistic state in order to ensure we read the new pinned values. Can conflict if a user re-orders and pins/unpins on a slow connection, but responses do catch up and result in correct state.

pfrazee (approved):
Yep looks good

PR #2035, opened 2023-11-29
Handle nsfw feeds in public view

estrattonbailey:


pfrazee (changes requested):
What if we just add a check in the feed-query for nsfw labels if you're signed out and throw an error if you are? Then the KnownError handler could pick it up and render what you're trying to do.

PR #2043, opened 2023-11-30
Change "Share" to "Copy link to {thing}" on web

### LINKED ISSUE #1107: Suggestion: rename "Share" button on the web to "Copy link"

Since in the web version the "Share" button simply copies the link to a post/profile to the clipboard (unlike in the mobile app, where it displays the native share sheet), I think it would be a bit clearer if it said something like "Copy link" or "Copy link to post/profile" instead.

(It could also use a more link-like icon like Twitter does)

(On mobile it should stay "Share" since it does something else)

Here's how it looks on Twitter:

<p></p>
<p></p>

Bluesky version:

<p></p>
<p></p>

mozzius:
Fixes #1107

Are we supposed to run the lingui cli for every change like this by the way?


pfrazee:
> Are we supposed to run the lingui cli for every change like this by the way?

Not sure, this is all new territory

pfrazee:
Another great PR, thanks!

PR #2046, opened 2023-11-30
Fix lag when switching tabs

gaearon:
The `select` function is re-executing on every render, and since it is always producing new slices, bailouts below in the tree no longer work. The fix is to [memoize the `select` function](https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option).

## Before

https://github.com/bluesky-social/social-app/assets/810438/d90d4b2e-6c67-46d5-bfaf-40b39e725d5b

## After

https://github.com/bluesky-social/social-app/assets/810438/5c2c4e44-7a3c-4925-ba35-5c47a147df6f


pfrazee:
This makes total sense and really validates our intuitions about how the select function is working (which is great news)

PR #2048, opened 2023-11-30
Hide normal mute if muted by list, and invalidate profile query upon list mute

estrattonbailey:
It's not perfect — we don't currently have a way to wait for list records to be commited to the repo, so we use a timeout — but it's a step in the right direction.

Also noticed a flicker when invalidating profile query that I fixed by only showing load state on a cold hit. Once cached (stale time is 5 min now), it just replaces data in background. See before, after:

https://github.com/bluesky-social/social-app/assets/4732330/2a5f49b4-c649-4c5f-a4c2-1ae4e062d273

https://github.com/bluesky-social/social-app/assets/4732330/50618d2b-03fc-4ffb-ab53-f1ded8fba43b

### DIFF
--- a/src/view/com/profile/ProfileHeader.tsx
+++ b/src/view/com/profile/ProfileHeader.tsx
@@ -7,6 +7,7 @@ import {
 } from 'react-native'
 import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {useNavigation} from '@react-navigation/native'
+import {useQueryClient} from '@tanstack/react-query'
 import {
   AppBskyActorDefs,
   ProfileModeration,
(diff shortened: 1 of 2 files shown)


src/view/com/profile/ProfileHeader.tsx
```diff
@@ -325,22 +336,24 @@ let ProfileHeaderLoaded = ({
...
-            label: profile.viewer?.muted
-              ? _(msg`Unmute Account`)
-              : _(msg`Mute Account`),
-            onPress: profile.viewer?.muted
-              ? onPressUnmuteAccount
-              : onPressMuteAccount,
-            icon: {
-              ios: {
-                name: 'speaker.slash',
+          if (!profile.viewer?.mutedByList) {
```
estrattonbailey, line 339: This matches our block/blockList handling

pfrazee (approved):
Yep looks good

PR #2049, opened 2023-11-30
Eric/profile feeds list

estrattonbailey:
Lists looked OK, but added a load state for it. Feeds looked like this:

Screenshot 2023-11-30 at 3 34 32 PM

Now they look like this:

Screenshot 2023-11-30 at 3 31 22 PM
Screenshot 2023-11-30 at 3 31 31 PM


pfrazee (approved):
Good stuff

PR #2052, opened 2023-11-30
Disable OTA for now

estrattonbailey:
Adds a little bit of Sentry noise. If we do use this, just close this PR.


pfrazee:
Yeah I'm good with this. We really can't use OTA until we test it in prod and have full confidence in it.

PR #2054, opened 2023-12-01
Maintain some feed data to avoid needless glimmers

pfrazee:
Fully resetting the query caches was causing needless "glimmers" (aka showing load-in-progress feed). This PR replaces the resets with a truncate to the first page and then invalidate, keeping some of the cache in place while more data loads.

### DIFF
--- a/src/state/queries/util.ts
+++ b/src/state/queries/util.ts
@@ -1,0 +1,17 @@
+import {QueryClient, QueryKey, InfiniteData} from '@tanstack/react-query'
+
+export function truncateAndInvalidate<T = any>(
+  queryClient: QueryClient,
+  querykey: QueryKey,
+) {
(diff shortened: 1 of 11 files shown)


gaearon (approved):

src/state/queries/util.ts
```diff
@@ -0,0 +1,17 @@
+import {QueryClient, QueryKey, InfiniteData} from '@tanstack/react-query'
+
+export function truncateAndInvalidate<T = any>(
```
estrattonbailey, line 3: Any?

estrattonbailey: Maybe doesn't matter since it's all infinite huh

estrattonbailey (approved):

src/state/queries/util.ts
pfrazee: yeah we dont touch that part of the object

PR #2055, opened 2023-12-01
Optimize pager rendering

gaearon:
Some misc perf improvements.

- Pull out the tabbar into its own component so we can memo it
- This also solves the animated style being recreated too often, causing header re-renders
- Adjust the other tabs' scroll views on a throttle instead of on every frame
- Pull out the usage of `isScrolling` out of `onScroll` to avoid invalidating `onScroll` when you scroll down
- Make `isScrolling` update non-blocking

We're still managing the header positioning with JS and it's laggy.

I'm not sure if we're gonna have to change the strategy there but at least it's an incremental improvement.

### Before (6x throttled)

https://github.com/bluesky-social/social-app/assets/810438/5ea2fc27-86df-40a3-a452-ba203863170e

### After (6x throttled)

https://github.com/bluesky-social/social-app/assets/810438/299925c1-ba48-4017-8ab0-cac023a60b11

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -138,7 +138,9 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
       }
     }
 
-    const throttleTimeout = React.useRef(null)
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Definitely improves it a bit.

From here I think the only way we get Web 100% is by looking at an alternative Web implementation that is entirely CSS.

PR #2058, opened 2023-12-01
Add Lingui check to CI

estrattonbailey:


.eslintrc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     'prettier',
   ],
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'detox', 'react'],
+  plugins: ['@typescript-eslint', 'detox', 'react', 'lingui'],
```
estrattonbailey, line 10: Unclear how helpful this will be, but it exists and we might as well try it

.gitignore
```diff
@@ -102,4 +102,8 @@ ios/
 google-services.json
 
 # Performance results (Flashlight)
-.perf/
\ No newline at end of file
+.perf/
+
+# i18n
+src/locale/locales/_build/
+src/locale/locales/**/*.js
```
estrattonbailey, line 109: Docs recommend ignoring compiled files

Issue #2061, opened 2023-12-01
Possible visual shift on feed tab bar when pinned feeds load


CooperEdmunds:
**Describe the bug**
The Feeds ✨button (added by #2033) could be rendered before the pinned feeds are loaded, which would cause a visual shift in the case that the user does in fact have custom feeds. I'm not sure if this would show up in prod, where the pinned feeds tend to load before the tab bar is rendered (at least for me), but I've been seeing it on the simulator which is much slower.

**To Reproduce**

This bug isn't in prod. It was just added and is on the main branch.

1. Open the app on an account that has a pinned feed after quitting the app
2. Look at the feeds tab bar and notice that the Feeds ✨button sometimes flashes before the pinned feeds are loaded

**Expected behavior**

Given the prevalence of that flow (it will happen for every user who has custom feeds every time they open the app after quitting), I think it's better to wait until the preferences have loaded to display the Feeds ✨ button so there isn't a visual shift in the case that the user does have pinned feeds.

**Screenshots**

I will add a screen recording in a minute

**Details**

 - Platform: iOS simulator
 - Platform version: iPhone 15 running iOS 17
 - App version: main branch

CooperEdmunds:
@pfrazee putting up a PR for this. I don't know if it will be a problem in prod, but might as well fix it. I also noticed that in #2033 I removed what's-hot from DEFAULT_FEEDS for prod in src/lib/constants.ts, but as of 18 days ago that was no longer being used - I should've removed it from DEFAULT_PROD_FEEDS in src/state/queries/preferences. Updated the issue and will include that in the PR. Sorry to not have done this right the first time. Thanks.

PR #2033, opened 2023-11-29
Feeds link in feeds tab bar

CooperEdmunds:
Implements the feed link in tab bar for users who do not have any pinned feeds or lists, per #2011.

[ed391c3](https://github.com/bluesky-social/social-app/pull/2033/commits/ed391c346d6e6858d0d24c08def2974df8ccbde7): Add a hasPinnedCustomFeedOrList boolean property to the usePinnedFeedsInfos hook (used in the next commit) and refactor the 3 places that were using that hook to get feeds. I included the property in this hook because usePinnedFeedsInfos will, at worst, update only as often as the list of feeds, so this won't cause any extra renders and has the benefit of colocation.

[8ceabe2](https://github.com/bluesky-social/social-app/pull/2033/commits/8ceabe2a11742973447a6e3b4489c8a5660f48c3): Use the hasPinnedCustomFeedOrList property added in the last commit to see if the user has any "real" (non-Following) items pinned. If not, display the "Feeds ✨" link. The button is rendered as just another item in the TabBar, but a new onSelect checks if the selected item is the "Feeds ✨" link. If so, open the feeds page; if not, call props.onSelect as before.

[3475979](https://github.com/bluesky-social/social-app/pull/2033/commits/34759798ebb2aaa4c292f000df8f19c7b9f75cb6): Removes the whats-hot feed from the default saved and pinned list for new users, so they will instead see the "Feeds ✨" link in the tab bar next to Following.

[3fbac46](https://github.com/bluesky-social/social-app/pull/2033/commits/3fbac466ac8554fcb690fccbe0053a5be07801d7): Updated existing home-screen tests that swipe and tap between feeds and added tests against the "Feeds ✨" link appearing, navigating to the Feeds tab when tapped, and disappearing when a feed is pinned.

Test results (all home-screen tests pass now):
2011-test-results-after-changes

Screenshots & demo videos:

When I'm signed out (unchanged):
2011-feeds-tab-bar-signed-out-iPhone

When I have no feeds or lists pinned:
2011-feeds-tab-bar-no-pinned-iPhone

When I have a feed pinned (unchanged):
2011-feeds-tab-bar-feed-pinned_iPhone

When I have a list pinned (unchanged):
2011-feeds-tab-bar-list-pinned_iPhone

iPhone demo:
https://github.com/bluesky-social/social-app/assets/6691295/4260e7ba-c720-4db4-abea-26318401e30f

Android demo:
[...]

PR #2063, opened 2023-12-01
Fix feeds tab bar layout shift

CooperEdmunds:
Related to issue #2061 - fixing some problems with #2033.

[c2bd86e](https://github.com/bluesky-social/social-app/pull/2063/commits/c2bd86eb61512aa0c98a3f66327a533c6b20d5ea): Make hasPinndCustom undefined until pinned feeds have been loaded and strictly check against it being false, so the Feeds ✨ button won't show until it is confirmed that the user has no pinned feeds or lists.

[041235e](https://github.com/bluesky-social/social-app/pull/2063/commits/041235e5d93e6f39f1e3dc6280da98a60f8a36f0): Removed what's-hot from being autopinned so this new button appears for new users. #2033 attempted to do this but deleted the defaults from the wrong file.

Video of the simulator before this change (look at the tab bar when it is loaded - you'll see a slight flash of "Feeds ✨" before the pinned feeds preferences loaded:
[todo]

Video after the change:
[todo]

### DIFF
--- a/src/state/queries/preferences/const.ts
+++ b/src/state/queries/preferences/const.ts
@@ -20,11 +20,9 @@ export const DEFAULT_THREAD_VIEW_PREFS: ThreadViewPreferences = {
   lab_treeViewEnabled: false,
 }
 
-const DEFAULT_PROD_FEED_PREFIX = (rkey: string) =>
-  `at://did:plc:z72i7hdynmk6r22z27h6tvur/app.bsky.feed.generator/${rkey}`
 export const DEFAULT_PROD_FEEDS = {
-  pinned: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
-  saved: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
+  pinned: [],
+  saved: [],
 }
 
 export const DEFAULT_LOGGED_OUT_PREFERENCES: UsePreferencesQueryResponse = {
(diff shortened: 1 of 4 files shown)


estrattonbailey (commented):
Makes sense to me, I'll let Paul stamp though

src/state/queries/preferences/const.ts
```diff
@@ -20,11 +20,9 @@ export const DEFAULT_THREAD_VIEW_PREFS: ThreadViewPreferences = {
   lab_treeViewEnabled: false,
 }
 
-const DEFAULT_PROD_FEED_PREFIX = (rkey: string) =>
-  `at://did:plc:z72i7hdynmk6r22z27h6tvur/app.bsky.feed.generator/${rkey}`
 export const DEFAULT_PROD_FEEDS = {
-  pinned: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
-  saved: [DEFAULT_PROD_FEED_PREFIX('whats-hot')],
```
pfrazee, line 27: I'm not quite ready to pull the trigger on this. We want to be sure the user has a feed; one of my current tasks is to ensure if a new user sees content, and the way we'll do that is by sampling from their saved feeds -- so we want to guarantee that one is set.

CooperEdmunds: @pfrazee just want to make sure you see this

CooperEdmunds:
@pfrazee @estrattonbailey Hey guys - bumping this because the "Feeds ✨" button is in prod now and has a visual shift when the user's preferences load which this PR solves (specially [c2bd86e](https://github.com/bluesky-social/social-app/pull/2063/commits/c2bd86eb61512aa0c98a3f66327a533c6b20d5ea))

PR #2064, opened 2023-12-01
Only poll feed when feed page is focused

estrattonbailey:
Noticed this happening and thought this was an easy enough fix. Easiest way to test is to throw some logs in, and drop the `pollInterval` in `Feed` to like 3s.


pfrazee (approved):
Good stuff

PR #2065, opened 2023-12-01
Use lastActiveTab from AsyncStorage for non-desktop

CooperEdmunds:


pfrazee:
Yeah that basically means we need to check if the index is too high

CooperEdmunds:
Now instead of tracking the last index they were on, I'm tracking the string so it persists correctly if they re-order. If the string that was last active is no longer pinned, it defaults to Following.

pfrazee (approved):
Yeah this works for me. Thanks for making the changes!

PR #2072, opened 2023-12-03
Tweak the loading placeholders to use the full space and position better

pfrazee:
|Before|After|
|-|-|
|Simulator Screenshot - iPhone 15 Pro - 2023-12-02 at 17 11 50|Simulator Screenshot - iPhone 15 Pro - 2023-12-02 at 17 11 11|

|Before|After|
|-|-|
|CleanShot 2023-12-02 at 17 12 35@2x|CleanShot 2023-12-02 at 17 00 13@2x|


pfrazee:
BTW I tried an off-the-shelf library that uses animated SVGs. Scaling the svgs for different platforms was a nightmare. Would not recommend again.

gaearon (approved):

PR #2081, opened 2023-12-04
Fix bottom space on feeds

estrattonbailey:
This is what the end of a profile feed looked like (test with `test.esb.lol`):
Screenshot 2023-12-04 at 9 45 38 AM

So I added a spacer:
Screenshot 2023-12-04 at 9 45 44 AM

And then figured we might as well tell the user what's up:
Screenshot 2023-12-04 at 9 44 38 AM

While scrolling, I noticed the loader was cut off, and you couldn't even see if before. Now looks like this (had to cut my internet to get this shot, hence no images):
Screenshot 2023-12-04 at 9 52 28 AM

And also noticed that the empty state at the bottom of following feed looked like this:
Screenshot 2023-12-04 at 9 40 03 AM

So I also added some min-spacing to that:
Screenshot 2023-12-04 at 9 48 21 AM

### DIFF
--- a/src/locale/locales/cs/messages.po
+++ b/src/locale/locales/cs/messages.po
@@ -83,7 +83,7 @@ msgstr ""
 
 #: src/view/com/modals/ListAddRemoveUsers.tsx:264
 #: src/view/com/modals/UserAddRemoveLists.tsx:187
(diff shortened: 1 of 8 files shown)


pfrazee (approved):
LGTM

PR #2085, opened 2023-12-04
Update feeds on post created

pfrazee:
This one is a little tricky

- After post, have to wait for appview to update
- In feed, only invalidate the cache _if_ just one page of data is loaded. This is really to act as a proxy to a UI issue that react-native can have with flatlists; if the user is scrolled far down, prepending data can cause the scroll position to begin vacillating wildly. Of course, RN doesn't give us a way to just get the scroll position, so this will have to do.


estrattonbailey (approved):
Makes sense to me!

PR #2088, opened 2023-12-05
Fix: correctly partial-match when calling truncateAndInvalidate()

pfrazee:
It turns out that `setQueryData` relies on an exact match for the react-key but `setQueriesData` can do partial matches (which sort of makes sense). This meant that the post feeds were still fetching all subsequent pages -- not truncating.


estrattonbailey (approved):
Ahhhhhhhhhhhh

PR #2090, opened 2023-12-05
Fix jump when toggling suggestions

gaearon:
## What?

The animation was buggy as hell on iOS and somewhat laggy on Android. The reason in both cases is that it triggers the pager `onLayout` logic which tries to readjust the FlatList padding which wreaks havoc *especially* on iOS.

For simplicity I ripped out the animation everywhere.

## Before

https://github.com/bluesky-social/social-app/assets/810438/c2496b01-3771-4ccd-b1f1-f6881678e2d9

https://github.com/bluesky-social/social-app/assets/810438/777c2936-2c93-4448-a79e-31e37dd91683

## After

https://github.com/bluesky-social/social-app/assets/810438/9166a4e8-d771-4d41-836d-baac3b98be42

https://github.com/bluesky-social/social-app/assets/810438/98dd7fdd-ef3a-4b2c-8869-d4f9ca0ba303

### DIFF
--- a/src/view/com/profile/ProfileHeaderSuggestedFollows.tsx
+++ b/src/view/com/profile/ProfileHeaderSuggestedFollows.tsx
@@ -1,11 +1,5 @@
 import React from 'react'
 import {View, StyleSheet, Pressable, ScrollView} from 'react-native'
-import Animated, {
-  useSharedValue,
-  withTiming,
-  useAnimatedStyle,
-  Easing,
-} from 'react-native-reanimated'
 import {AppBskyActorDefs, moderateProfile} from '@atproto/api'
 import {
   FontAwesomeIcon,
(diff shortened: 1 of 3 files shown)


src/view/com/profile/ProfileHeaderSuggestedFollows.tsx
```diff
@@ -34,112 +28,84 @@ const TOTAL_HEIGHT = 250
...
-
   return (
-    <Animated.View
-      pointerEvents="box-none"
-      style={[{overflow: 'hidden', opacity: 0}, animatedStyles]}>
-      <View style={{paddingVertical: OUTER_PADDING}} pointerEvents="box-none">
+    <View
+      style={{paddingVertical: OUTER_PADDING, height: TOTAL_HEIGHT}}
+      pointerEvents="box-none">
+      <View
```
gaearon, line 44: The rest is whitespace changes.

gaearon:
This has a nice side effect of *not* requesting suggestions until you toggle. We should preserve this behavior even if we reintroduce the animation later.

pfrazee (approved):

haileyok:
We're rounding this to 0.5 now in https://github.com/bluesky-social/social-app/pull/4294/commits/4f8331fafbd2916290e5032f36b50dba2f971d17

PR #2093, opened 2023-12-05
More treeview UI tweaks

pfrazee:
A little more treeview UI tweaking, as a treat

- More consistent spacing
- Always include one level of "reply bar" to avoid the awkward left-edge visual flow
- Slightly shrink the expanded post's avi (Fons is going to reduce avi sizes in general with the updated designs)

|Before|After|
|-|-|
|Simulator Screenshot - iPhone 15 Pro - 2023-12-05 at 12 31 54|Simulator Screenshot - iPhone 15 Pro - 2023-12-05 at 12 31 16|

|Before|After|
|-|-|
|Simulator Screenshot - iPhone 15 Pro - 2023-12-05 at 12 32 03|Simulator Screenshot - iPhone 15 Pro - 2023-12-05 at 12 31 41|

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -280,6 +280,9 @@ function PostThreadLoaded({
         const prev = isThreadPost(posts[index - 1])
           ? (posts[index - 1] as ThreadPost)
           : undefined
+        const next = isThreadPost(posts[index - 1])
+          ? (posts[index - 1] as ThreadPost)
+          : undefined
         return (
           <View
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Subjective, but idk if I dig the left-most line. It's effectively a border now, and makes me think there is an outer-most reply that I just haven't scrolled down to yet. Esp on mobile it seems odd to me 🤔 

Not a blocker though! I don't even typically use the thread view :P

pfrazee:
Maybe, but the leftmost indentation was confusing for the opposite reason (and I think just visually ugly)

gaearon:
I kinda like how Reddit does it

PR #2097, opened 2023-12-05
Fix email confirmation dialog on startup

gaearon:
This was previously crashing because `unstable__openModal` was being called too early. Also nice to move it out of the critical path. I'll probably move a few other things here too.

Screenshot 2023-12-05 at 23 44 47
Screenshot 2023-12-05 at 23 44 55
Screenshot 2023-12-05 at 23 45 04

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -35,6 +35,12 @@ import {bskyTitle} from 'lib/strings/headings'
 import {JSX} from 'react/jsx-runtime'
 import {timeout} from 'lib/async/timeout'
 import {useUnreadNotifications} from './state/queries/notifications/unread'
+import {useSession} from './state/session'
+import {useModalControls} from './state/modals'
+import {
+  shouldRequestEmailConfirmation,
+  setEmailConfirmationRequested,
+} from './state/shell/reminders'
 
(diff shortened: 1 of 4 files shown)


src/Navigation.tsx
```diff
@@ -464,19 +470,25 @@ const LINKING = {
 
 function RoutesContainer({children}: React.PropsWithChildren<{}>) {
```
gaearon, line 471: I did it here because `Shell` component is forked and `App` is also forked but the logic is the same. Also, this gives us the precise point where extra logic doesn't block the first render so it's a good consolidation point.

src/Navigation.tsx
```diff
@@ -464,19 +470,25 @@ const LINKING = {
...
   const theme = useColorSchemeStyle(DefaultTheme, DarkTheme)
+  const {currentAccount} = useSession()
+  const {openModal} = useModalControls()
+
+  function onReady() {
```
gaearon, line 476: Verified this re-fires when we change the account — due to the key thing at the top.

I.e. logging out and in (barring the suppression condition for the storage) shows the modal.

estrattonbailey (approved):
Fire

PR #2098, opened 2023-12-05
Improvements to persisted state migration

estrattonbailey:
Noticed a [Sentry log](https://blueskyweb.sentry.io/issues/4699579547/?query=is%3Aunresolved+dist%3Aios.1.57.0.7&referrer=issue-stream&statsPeriod=24h&stream_index=9) about a failed migration. The logs pointed to a validation failure around `email` and `emailValidated`. Sure enough, these were optional fields in the old architecture.

This PR marks those fields as optional and any usages to be more safe. I also added a preemptive validation check so that we can more gracefully handle any unforeseen issues. 

I've tested this a few times by resetting back to 46c2564e6531712538e44ee6880fb4bfce612ab9, clearing storage, etc, and then jumping back to the head of this branch. While doing that I clarified a few logs.

### DIFF
--- a/src/state/persisted/__tests__/legacy.test.ts
+++ b/src/state/persisted/__tests__/legacy.test.ts
@@ -1,0 +1,9 @@
+import {expect, test} from '@jest/globals'
+
+import {transform} from '#/state/persisted/legacy'
+import {schema} from '#/state/persisted/schema'
+
+test('transform', () => {
+  const data = transform({})
+  expect(() => schema.parse(data)).not.toThrow()
+})
--- a/src/state/persisted/__tests__/legacy.test.ts
+++ b/src/state/persisted/__tests__/legacy.test.ts
@@ -1,7 +1,11 @@
 import {expect, test} from '@jest/globals'
 
 import {transform} from '#/state/persisted/legacy'
-import {schema} from '#/state/persisted/schema'
+import {defaults, schema} from '#/state/persisted/schema'
+
+test('defaults', () => {
+  expect(() => schema.parse(defaults)).not.toThrow()
+})
 
 test('transform', () => {
   const data = transform({})
--- a/src/state/persisted/legacy.ts
+++ b/src/state/persisted/legacy.ts
@@ -1,7 +1,7 @@
 import AsyncStorage from '@react-native-async-storage/async-storage'
 
 import {logger} from '#/logger'
-import {defaults, Schema} from '#/state/persisted/schema'
+import {defaults, Schema, schema} from '#/state/persisted/schema'
 import {write, read} from '#/state/persisted/store'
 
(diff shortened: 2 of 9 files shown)


src/state/persisted/__tests__/legacy.test.ts
estrattonbailey, line 1: Sanity checks

src/state/persisted/legacy.ts
```diff
@@ -138,34 +137,40 @@ export async function migrate() {
...
               legacy?.session?.data?.did,
           ),
         })
-      } else {
-        logger.info(`persisted state: no legacy to debug, fresh install`)
       }
-    } catch (e) {
-      logger.error(`persisted state: legacy debugging failed`, {error: e})
+    } catch (e: any) {
+      logger.error(e, {message: `persisted state: legacy debugging failed`})
```
estrattonbailey, line 150: Passing the exception directly here for more detail

src/state/persisted/legacy.ts
```diff
@@ -138,34 +137,40 @@ export async function migrate() {
...
-      logger.log('persisted state: migrated legacy storage')
+      const validate = schema.safeParse(newData)
+
+      if (validate.success) {
+        await write(newData)
+        logger.log('persisted state: migrated legacy storage')
+      } else {
+        logger.error('persisted state: legacy data failed validation', {
+          error: validate.error,
+        })
```
estrattonbailey, line 166: If it fails, this will give us a nice pretty-printed reason as to why

src/state/persisted/schema.ts
```diff
@@ -2,17 +2,14 @@ import {z} from 'zod'
...
 // only data needed for rendering account page
-// TODO agent.resumeSession requires the following fields
 const accountSchema = z.object({
   service: z.string(),
   did: z.string(),
   handle: z.string(),
-  email: z.string(),
-  emailConfirmed: z.boolean(),
+  email: z.string().optional(),
+  emailConfirmed: z.boolean().optional(),
```
estrattonbailey, line 10: These were the culprits

src/view/com/modals/ChangeEmail.tsx
```diff
@@ -118,8 +118,8 @@ export function Component() {
           ) : stage === Stages.ConfirmCode ? (
             <Trans>
               An email has been sent to your previous address,{' '}
-              {currentAccount?.email || ''}. It includes a confirmation code
-              which you can enter below.
+              {currentAccount?.email || '(no email)'}. It includes a
```
estrattonbailey, line 121: I've chosen to fall back to `(no email)` instead of empty strings throughout

gaearon (approved):

PR #2099, opened 2023-12-06
Prevent invalid locale from being loaded

estrattonbailey:
Reported by [Jake in Slack](https://bluesky-builders.slack.com/archives/C04BH54QDLH/p1701818542377789). The selector library we're using doesn't support `disabled` on its placeholder option (or any option, ugh).

This PR just prevents any locale we're not ready for from being loaded + disables the placeholder option entirely.

### DIFF
--- a/src/view/screens/LanguageSettings.tsx
+++ b/src/view/screens/LanguageSettings.tsx
@@ -48,6 +48,7 @@ export function LanguageSettingsScreen(_props: Props) {
 
   const onChangePrimaryLanguage = React.useCallback(
     (value: Parameters<PickerSelectProps['onValueChange']>[0]) => {
+      if (!value) return
       if (langPrefs.primaryLanguage !== value) {
         setLangPrefs.setPrimaryLanguage(value)
       }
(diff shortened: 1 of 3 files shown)


src/view/screens/LanguageSettings.tsx
```diff
@@ -48,6 +48,7 @@ export function LanguageSettingsScreen(_props: Props) {
 
   const onChangePrimaryLanguage = React.useCallback(
     (value: Parameters<PickerSelectProps['onValueChange']>[0]) => {
+      if (!value) return
```
estrattonbailey, line 51: In case the placeholder ever shows back up, it has a value of `null`.

src/view/screens/LanguageSettings.tsx
```diff
@@ -100,6 +102,7 @@ export function LanguageSettingsScreen(_props: Props) {
 
           <View style={{position: 'relative'}}>
             <RNPickerSelect
+              placeholder={{}}
```
estrattonbailey, line 105: This is how you disable it 🤷

src/view/screens/LanguageSettings.tsx
ansh: Cool

ansh (approved):

pfrazee (approved):

PR #2102, opened 2023-12-06
Split notification init, add cleanup

gaearon:
Similar in spirit to https://github.com/bluesky-social/social-app/pull/2097.

Currently, the logic for initializing push notifications is inside a single `init` function which contains a mix of event handlers, some async code, and a `listenSessionLoaded` callback. There are a few problems with this:

- There's no consistent cleanup cycle.
- The `Notifications.addPushTokenListener` leaks when the user changes.
- It's not easy to fix that because the subscription happens after an async function.
- In general, `listenSessionLoaded` is a footgun because moving this code anywhere is risking a race.

This PR breaks the logic apart into three pieces:

- The RQ-related subscribers that never need to unmount and remain in the top-level init path.
- The per-user async "ask for permissions and register a token" flow which does not need cleanup.
- The per-user "subscribe in case we get a new token" flow which *does* need cleanup.

There's also a little tidying up of forked logic to be more similar and removal of superficial guards (per types).

I don't know how to test this so I didn't.


src/view/shell/index.tsx
```diff
@@ -67,6 +68,19 @@ function ShellInner() {
...
 
+  React.useEffect(() => {
+    if (currentAccount) {
+      notifications.requestPermissionsAndRegisterToken(currentAccount)
+    }
+  }, [currentAccount])
+
+  React.useEffect(() => {
+    if (currentAccount) {
+      const unsub = notifications.registerTokenChangeHandler(currentAccount)
```
gaearon, line 79: This is fine to run even if we don't have perms cause it's just an event emitter.

estrattonbailey (commented):
Conceptually makes sense to me, also a little unsure how to test all cases

PR #2104, opened 2023-12-06
Perf: Include quote posts in the post-thread placeholder

pfrazee:
Updates the search by the post-thread query for placeholder data to include quoted posts, reducing even more cases where the user might get a spinner during the PostThread load time.


gaearon (approved):
beautiful

PR #2105, opened 2023-12-06
Update to actions/setup-node@v4

pfrazee:
Trying to fix the flakey install step in our GH actions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install node 18
-        uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -17,7 +17,11 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v3
       - name: Yarn install
-        run: yarn --frozen-lockfile
+        uses: Wandalen/wretry.action@master
+        with:
+          command: yarn --frozen-lockfile
+          attempt_limit: 3
```
gaearon, line 23: 5?

pfrazee:
Ended up adding retries, which solves it. Thus revealing a type error that got missed because of this problem in the first place

PR #2106, opened 2023-12-06
Check Analytics

ansh:
I went through all the analytics events on web and mobile and added, removed, and modified any events that were lost/changed during our state transition.

### DIFF
--- a/src/view/com/posts/Feed.tsx
+++ b/src/view/com/posts/Feed.tsx
@@ -294,7 +294,7 @@ let Feed = ({
         scrollEventThrottle={scrollEventThrottle}
         indicatorStyle={theme.colorScheme === 'dark' ? 'white' : 'black'}
         onEndReached={onEndReached}
-        onEndReachedThreshold={3} // number of posts left to trigger load more
+        onEndReachedThreshold={2} // number of posts left to trigger load more
         removeClippedSubviews={true}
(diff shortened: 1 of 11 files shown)


src/view/com/posts/Feed.tsx
```diff
@@ -294,7 +294,7 @@ let Feed = ({
         scrollEventThrottle={scrollEventThrottle}
         indicatorStyle={theme.colorScheme === 'dark' ? 'white' : 'black'}
         onEndReached={onEndReached}
-        onEndReachedThreshold={2}
+        onEndReachedThreshold={3} // number of posts left to trigger load more
```
estrattonbailey, line 297: Do you want this in this PR?

src/view/com/posts/FeedItem.tsx
```diff
@@ -121,7 +119,6 @@ let FeedItemInner = ({
   }, [record?.reply])
 
   const onPressReply = React.useCallback(() => {
-    track('FeedItem:PostReply')
```
estrattonbailey, line 124: Intentional?

estrattonbailey (approved):

src/view/com/posts/Feed.tsx
ansh: Probably not. We can test it more later.

PR #2107, opened 2023-12-06
3rd party embedded player (can't finish right now, tossing up here)

pfrazee:
This was a weekend project we don't have time to finish, but if anybody in the community is inclined, feel free to take this to the finish line.

Here are my TODO notes from when I was working on it

- [ ] twitch parent param
- [ ] needed origin?
- [ ] parse tests
- [ ] dark mode
- [ ] no image render
- [ ] fix android in viewport
- [ ] fix android dimensions
- [ ] remove hack


mackuba:
That would definitely be cool to have! 👍

chilblane:
possibility of getting soundcloud and bandcamp support in this pr as well? (whoever does it)

PR #2110, opened 2023-12-06
Throttle non-critical Sentry messages

gaearon:
Let's get Sentry logs out of the critical path.

Two changes:

- Non-exception messages are put in a queue throttled with a leading edge of 7 seconds. I picked a number that's likely enough to unclog the critical path on a slow network but still reasonable that the app will likely stay foregrounded etc.
- Disable [session tracking](https://docs.sentry.io/platforms/javascript/configuration/options/#auto-session-tracking) cause I don't think we care about this? This was causing an extra request at startup.

## Before (slow 3G)

Screenshot 2023-12-06 at 16 11 38

## After (slow 3G)

Screenshot 2023-12-06 at 16 10 57

### DIFF
--- a/src/lib/sentry.ts
+++ b/src/lib/sentry.ts
@@ -37,6 +37,7 @@ const dist = `${Platform.OS}.${release}${
 }`
 
 init({
+  autoSessionTracking: false,
   dsn: 'https://05bc3789bf994b81bd7ce20c86ccd3ae@o4505071687041024.ingest.sentry.io/4505071690514432',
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):

estrattonbailey (commented):
Wait, will we lose release health if we disable sessions? https://docs.sentry.io/platforms/javascript/guides/electron/configuration/integrations/mainprocesssession/

gaearon:
Oh didn't see the comment.

PR #2114, opened 2023-12-06
Refetch some data on an interval

estrattonbailey:
`refetchInterval` re-fetches and updates data regardless of `staleTime`, so at face value, I think it makes sense to match these time periods.

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -32,6 +32,7 @@ export function useProfileQuery({did}: {did: string | undefined}) {
     // if you remove it, the UI infinite-loops
     // -prf
     staleTime: STALE.MINUTES.FIVE,
+    refetchInterval: STALE.MINUTES.FIVE,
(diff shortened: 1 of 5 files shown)


src/state/queries/profile.ts
```diff
@@ -32,6 +32,7 @@ export function useProfileQuery({did}: {did: string | undefined}) {
     // if you remove it, the UI infinite-loops
     // -prf
     staleTime: STALE.MINUTES.FIVE,
+    refetchInterval: STALE.MINUTES.FIVE,
```
estrattonbailey, line 35: Five is good for viewing other user's profiles, but for your own, we may want to conditionally lower this to like 30s? @pfrazee thoughts?

pfrazee: Mmm yeah that might be the move

pfrazee (changes requested):
Let's add that conditional on profile, otherwise lgtm

PR #2118, opened 2023-12-06
Add tests for migration and persisted state

estrattonbailey:
Started to write tests for all persisted state APIs but got bogged down by handling the broadcast channel mocks. I can come back to that in a separate PR.


src/state/persisted/legacy.ts
```diff
@@ -7,15 +7,15 @@ import {write, read} from '#/state/persisted/store'
...
+export type LegacySchema = {
   shell: {
     colorMode: 'system' | 'light' | 'dark'
   }
   session: {
     data: {
       service: string
       did: `did:plc:${string}`
-    }
+    } | null
```
estrattonbailey, line 18: Small type update. Access below was using optional chaining, but this is more correct.

src/state/persisted/legacy.ts
```diff
@@ -61,7 +61,7 @@ type LegacySchema = {
     copiedInvites: string[]
   }
   mutedThreads: {uris: string[]}
-  reminders: {lastEmailConfirm: string}
+  reminders: {lastEmailConfirm?: string}
```
estrattonbailey, line 64: Small type update. Access below was using optional chaining, but this is more correct.

pfrazee (approved):
good stuff

pfrazee:
@estrattonbailey Getting some weird sentry errors in the CI

PR #2119, opened 2023-12-06
Fixes issue with

ansh:
Allow going directly to password input screen when switching accounts and password is required

Solves: https://github.com/orgs/bluesky-social/projects/9/views/1?pane=issue&itemId=46244229

https://github.com/bluesky-social/social-app/assets/4732330/24956111-8606-452e-a016-d354ee694d64

### DIFF
--- a/src/state/shell/logged-out.tsx
+++ b/src/state/shell/logged-out.tsx
@@ -24,6 +24,11 @@ type Controls = {
      */
     requestedAccount?: string
   }) => void
+  /**
+   * Clears the requested account so that next time the logged out view is
+   * show, no account is pre-populated.
+   */
+  clearRequestedAccount: () => void
 }
 
 const StateContext = React.createContext<State>({
@@ -34,6 +39,7 @@ const StateContext = React.createContext<State>({
 const ControlsContext = React.createContext<Controls>({
   setShowLoggedOut: () => {},
   requestSwitchToAccount: () => {},
+  clearRequestedAccount: () => {},
 })
 
 export function Provider({children}: React.PropsWithChildren<{}>) {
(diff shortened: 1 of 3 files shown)


src/state/shell/logged-out.tsx
```diff
@@ -1,23 +1,43 @@
 import React from 'react'
 
-type StateContext = {
+const StateContext = React.createContext<{
   showLoggedOut: boolean
-}
-
-const StateContext = React.createContext<StateContext>({
+  requestedAccountSwitchTo?: string // did of account to switch to
```
estrattonbailey, line 5: Nit: if we're gonna go with this approach, can you make these named arguments via an object? Just want it to be very clear 🙏

src/view/com/auth/LoggedOut.tsx
```diff
@@ -36,6 +38,12 @@ export function LoggedOut({onDismiss}: {onDismiss?: () => void}) {
     setMinimalShellMode(true)
   }, [screen, setMinimalShellMode])
 
+  React.useEffect(() => {
+    if (requestedAccountSwitchTo) {
+      setScreenState(ScreenState.S_Login)
+    }
+  }, [requestedAccountSwitchTo])
```
estrattonbailey, line 45: Can't just set this in the useState? Or does that break back button functionality 🤔

pfrazee (changes requested):
Yeah this looks like a good approach. Will let you answer eric's questions, but otherwise +1

src/view/com/auth/LoggedOut.tsx
gaearon: As much as possible, we shouldn't have effects that update state. It means something else in the logic is not quite right.

src/state/shell/logged-out.tsx
```diff
@@ -1,23 +1,43 @@
...
+    [showLoggedOut, requestedAccountSwitchTo],
+  )
+
+  const controls = React.useMemo(
+    () => ({
+      /**
+       * Sets the visibility of the logged-out state and optionally specifies the account to switch to.
+       * @param show - Whether to show the logged-out state.
+       * @param did - The account to switch to (optional).
+       */
```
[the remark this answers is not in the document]
ansh: Thanks for the force push my dude!

PR #2122, opened 2023-12-07
Add PWI opt-out toggle

pfrazee:
This just adds the opt-out toggle. The behaviors for handling the opt-out can come in a follow-up PR (not a blocker until we enable the PWI).

- Added to the moderation screen
- Open to discussing the copy, but we need to settle it by Friday

## Current copy

CleanShot 2023-12-06 at 18 45 17@2x

## Previous iterations (no longer current)

CleanShot 2023-12-06 at 16 12 39@2x

### DIFF
--- a/src/view/screens/Moderation.tsx
+++ b/src/view/screens/Moderation.tsx
@@ -182,14 +182,15 @@ function PwiOptOut() {
 
   return (
     <View style={[pal.view, styles.toggleCard]}>
-      <View style={{flexDirection: 'row', alignItems: 'center'}}>
+      <View
+        style={{flexDirection: 'row', alignItems: 'center', paddingRight: 14}}>
         <ToggleButton
           type="default-light"
           label={_(msg`Request to limit the visibility of my account`)}
           labelType="lg"
           isSelected={isOptedOut}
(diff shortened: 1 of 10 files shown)


src/view/screens/Moderation.tsx
```diff
@@ -109,10 +121,112 @@ export function ModerationScreen({}: Props) {
...
+    profile?.labels?.some(l => l.val === '!no-unauthenticated') || false
+  const canToggle = profile && !updateProfile.isPending
+
+  const onToggleOptOut = React.useCallback(() => {
+    if (!profile) {
+      return
+    }
+    updateProfile.mutate({
+      profile,
+      updates: existing => {
```
gaearon, line 158: I realize this is not new, but are we mutating something that's also used for rendering? Generally in React we shouldn't mutate stuff we render — can this be a new object?

gaearon (approved):
lg but mutation seems shady

src/view/screens/Moderation.tsx
pfrazee: It's not being used for rendering. I agree with the general approach as a safe default approach, but this is a narrow non-react api that handles this okay. Not against changing it, but not an issue

PR #2126, opened 2023-12-07
New Web Layout

gaearon:
## What?

Reworking of web layout that makes the document itself scrollable instead of the list node. Despite the dicey PR title, how the layout *looks* is meant to remain the same. What I'm aiming to change is how it achieves that look. 

Removes virtualization on the web although custom one could plausibly be re-added.

## Why?

Fixes scroll jump on clicking into post (before/after):

https://github.com/bluesky-social/social-app/assets/810438/6eb2c7d6-c61d-4bad-9276-8372547d6ff4

Fixes scroll lag on profile:

https://github.com/bluesky-social/social-app/assets/810438/dfb6b6c5-f281-4a8b-bf73-1152a00a10df

Will also unblock fixing unselectable text in profile header.

## Plan

- [x] Essentials
  - [x] Rip out `FlatList` which assumes the scrollable node is owned by it
  - [x] Bind `onScroll` handlers to the `window`
  - [x] Implement `scrollToOffset`
  - [x] Make tab headers `position: sticky`
  - [x] Fix bugs
    - [x] Modal and lightbox can be scrolled away
    - [x] Headers and footers
    - [x] More screen headers need to be changed to fixed
    - [x] What is causing Back to animate?
    - [x] Layout fixes
      - [x] Profile header width at different breakpoints
    - [x] Fix screen backgrounds
      - [x] When cmd+down'ing too hard
      - [x] In light mode during load
    - [x] Safari doesn't trigger "load more"
    - [x] Scroll
      - [x] Move towards a single scroll handler
        - [x] https://github.com/bluesky-social/social-app/pull/2200
        - [x] https://github.com/bluesky-social/social-app/pull/2216
        - [x] Reduce dependency on `onScroll`
          - [x] Implement `onEndReached` without `onScroll`
          - [x] Implement `onScrolledDownChange` without `onScroll`
        - [x] Track logical list visibility for a single scroll handler
        - [x] Delegate the scroll context to the single scroll handler
      - [x] Preserve and reset scroll as needed
        - [x] Figure out why Back resets scroll
        - [x] In pagers, reset and restore position per tab
        - [x] For routes, reset and restore on navigation
          - [x] Proof of concept
          - [x] Polish it up
            - [x] Integrate with pager restore
            - [x] Make sure Back/Forward is sensible
        - [x] Fix post thread jump
- [x] Perf
  - [x] Add memoization
  - [x] Add containment CSS properties
- [x] Tidying up
  - [x] Fix native if regressed
  - [x] Clean up the code
  - [x] Fix type errors
  - [x] Check other browsers
  - [x] Check mobile Safari etc
- [x] Polish
  - [x] Search should jump to top on query change
  - [x] Add scroll lock to modal and lightbox
  - [x] Figure out why iOS Safari lags when scrolling feeds header away
- [x] Bugs and regressions
  - [x] Search field and pager should be sticky
- [ ] Nitpicks
  - [x] Missing border on Home tablet feeds bar
  - [x] 1px vertical gap between borders on Home tablet feeds bar
  - [ ] <s>Empty screen during popstate gesture on iOS Safari</s> (won't fix)
  - [x] Scroll lock mode causes the scrollbar to show/hide which moves content (dropdowns trigger this)
  - [x] When scrollbars are "always on", there's a layout jump on some nav (eg Home -> Notification) on desktop
- [ ] Future follow-ups
  - [ ] Fix `any` types, enforce that `.web` and `.native` types match
  - [ ] Remove `desktopFixedHeight` prop if we can (is it needed anymore?)
  - [ ] Remove `minHeight: Dimensions.get('window').height * 1.5` styles if we can (do we need them anymore?)
  - [ ] Get rid of the `pointer-events` hack to fix unselectable text in profile header
  - [ ] Consider deleting `FlatList` from `Views.web` and possibly forcing exclusion of its code from the bundle


src/view/com/util/List.web.tsx
```diff
@@ -0,0 +1,281 @@
...
+      <View
+        style={[
+          styles.contentContainer,
+          contentContainerStyle,
+          desktopFixedHeight ? styles.minHeightViewport : null,
+          pal.border,
+        ]}>
+        <Visibility
+          onVisibleChange={handleAboveTheFoldVisibleChange}
+          style={[styles.aboveTheFoldDetector, {height: headerOffset}]}
```
gaearon, line 142: This is where I'm using https://github.com/bluesky-social/social-app/pull/2216. I think it's helpful to know at this level that this is a specific concept (`headerOffset`) and not just some padding.

pfrazee:
Starting with the positives:
- Profile header on web is so much smoother
- Scrolling on web generally feels smoother? maybe placebo effect
- Text is selectable in profile headers again on web
- Scroll position dejanked on web when opening replies
- Scroll wheel on web works no matter where your cursor is!!! ugh god that's nice to have fixed

Small issues that I'd be 100% okay with solving in followups:
- ProfileList and ProfileScreen on web hit a point (seems like 100% screen height) where the sticky header scrolls away
- If you go back, forward, then back (via browser tools) you lose your scroll position on the last back. So it's back (scroll position is correct), forward, back (scroll position lost).

Somewhat less minor issues but still okay with putting in followups:
- Scroll position on web isn't correctly focusing reply posts when I open them from notifications
- Screens that still need to be updated:
  - Home feed preferences
  - Moderation preferences
  - Saved feeds
  - etc -- just need to do a pass to ensure they're all using the new layout

pfrazee (approved):
This was an enormous PITA slog and I appreciate you sticking with it dan. I'm excited to have this done.

PR #2129, opened 2023-12-07
More session improvements

estrattonbailey:


src/state/session/index.tsx
```diff
@@ -104,31 +102,43 @@ function createPersistSessionHandler(
   }) => void,
 ): AtpPersistSessionHandler {
   return function persistSession(event, session) {
-    const expired = !(event === 'create' || event === 'update')
+    const expired = event === 'expired' || event === 'create-failed'
```
estrattonbailey, line 105: Better to be more explicit

src/state/session/index.tsx
```diff
@@ -104,31 +102,43 @@ function createPersistSessionHandler(
...
+     * If the session expired, or it was successfully created/updated, we want
+     * to update/persist the data.
+     *
+     * If the session creation failed, it could be a network error, or it could
+     * be more serious like an invalid token(s). We can't differentiate, so in
+     * order to allow the user to get a fresh token (if they need it), we need
+     * to persist this data and wipe their tokens, effectively logging them
+     * out.
+     */
     persistSessionCallback({
```
estrattonbailey, line 142: I think ideally, we'd like to know if it's a potentially recoverable issue or not so that we could keep the user's tokens around to try again later. But since we can't, we need clear them for two reasons:
1. To bump the user out to sign in
2. To prevent a loop of attempting to sign in with bad tokens and them never getting wiped

This hasn't changed, I'm just highlighting it for future reference.

src/state/session/index.tsx
```diff
@@ -355,23 +364,54 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+            /*
+             * Note: `agent.persistSession` is also called when this fails, and
+             * we handle that failure via `createPersistSessionHandler`
+             */
+            logger.info(`session: resumeSessionWithFreshAccount failed`, {
+              error: e,
+            })
+
+            __globalAgent = PUBLIC_BSKY_AGENT
+          })
```
estrattonbailey, line 383: Want to avoid the unhandled rejection, since we handle this elsewhere.

The global agent also needs to be reassigned so that auth failures don't try to use an agent with a session, which will result in 401s.

src/state/session/index.tsx
```diff
@@ -355,23 +364,54 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+            error: e,
+          })
+
+          __globalAgent = PUBLIC_BSKY_AGENT
+        }
       }
 
       async function resumeSessionWithFreshAccount(): Promise<SessionAccount> {
+        // TODO on flaky networks, this takes ages to time out
         await networkRetry(3, () => agent.resumeSession(prevSession))
```
estrattonbailey, line 405: Just noting that if we wait for 4 total requests to time out, the user sees a spinner for multiple minutes depending on browser behavior

pfrazee (approved):

PR #2130, opened 2023-12-07
Fix: remove replies to deleted and blocked posts from feeds

pfrazee:
Finally found a test case.


estrattonbailey (approved):
Not sure how best to test, but logic makes sense to me

PR #2135, opened 2023-12-07
Improve account switching on mobile

estrattonbailey:


src/lib/hooks/useAccountSwitcher.ts
```diff
@@ -23,8 +23,13 @@ export function useAccountSwitcher() {
       try {
         if (account.accessJwt) {
           closeAllActiveElements()
-          navigation.navigate(isWeb ? 'Home' : 'HomeTab')
```
estrattonbailey, line 26: This is probably what was causing a flash of the home screen for the wrong account, since we were navigating and THEN changing accounts 🤦

pfrazee (approved):
:fire:

PR #2137, opened 2023-12-07
Disable animation on scrollToTop for web

pfrazee:
In some cases on Chrome and always on FF, the scroll to top would fail because the animation wouldnt finish before content changed in the feed, causing the scroll to totally fail.

Perhaps a better fix would be to wait until the scroll finishes to queue the work, but I went with the simple solution first.


estrattonbailey (approved):
💆‍♂️

PR #2138, opened 2023-12-07
Fix Home navigation after account switch on the web

gaearon:
Fixes https://github.com/bluesky-social/social-app/pull/2135/files#r1419618341.

Any navigations adjacent to app remounts might need similar fixes.

A more comprehensive alternative could be to adjust our code to always `pushState` early. But that his might break the router so let's wait with that.

https://github.com/bluesky-social/social-app/assets/810438/2e89c7c7-e8b2-45b0-ab96-69d29b32ded7

### DIFF
--- a/src/lib/hooks/useAccountSwitcher.ts
+++ b/src/lib/hooks/useAccountSwitcher.ts
@@ -1,8 +1,6 @@
 import {useCallback} from 'react'
-import {useNavigation} from '@react-navigation/native'
 
 import {isWeb} from '#/platform/detection'
-import {NavigationProp} from '#/lib/routes/types'
 import {useAnalytics} from '#/lib/analytics/analytics'
 import {useSessionApi, SessionAccount} from '#/state/session'
 import * as Toast from '#/view/com/util/Toast'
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useAccountSwitcher.ts
```diff
@@ -23,16 +20,15 @@ export function useAccountSwitcher() {
...
-           * specifically.
-           */
           if (isWeb) {
-            navigation.navigate('Home')
+            // We're switching accounts, which remounts the entire app.
+            // On mobile, this gets us Home, but on the web we also need reset the URL.
+            // We can't change the URL via a navigate() call because the navigator
+            // itself is about to unmount, and it calls pushState() too late.
+            // So we change the URL ourselves. The navigator will pick it up on remount.
+            history.pushState(null, '', '/')
```
gaearon, line 29: I put this before `selectAccount` to emphasize that we're changing something for the action caused by `selectAccount` to "pick up" later. By itself this call doesn't re-render.

pfrazee (approved):
LGTM

PR #2146, opened 2023-12-08
Add manual per-page memoization to post select

gaearon:
## What?

We're always regenerating new data in `select()` so every real re-render with new data would lead to recalculation of all slices and blow away our memoization. Let's add a manual memoization step that reuses pages from the previous selection (which in many cases will work out — notably for pagination). I'm assuming pages are immutable (I think so).

Note that we need to advance the tuner while going over the reused pages.

Also note `structuralSharing` wouldn't help us here because the issue is with the pages that were *not* refetched.

[Review without whitespace.](https://github.com/bluesky-social/social-app/pull/2146/files?w=1)

## Before

Every time you fetch a new page, all previous items re-render.

First commit (extraneous and slow -- all pages):

Screenshot 2023-12-08 at 04 40 07

Second commit (expected:

Screenshot 2023-12-08 at 04 40 16

## After

When you fetch a new page, only items for that page render (for the first time). 

There's just one commit (it was second in "Before") which only touches the new items.

Screenshot 2023-12-08 at 04 39 12

### DIFF
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -1,4 +1,4 @@
-import {useCallback, useEffect} from 'react'
+import React, {useCallback, useEffect} from 'react'
 import {
   AppBskyFeedDefs,
   AppBskyFeedPost,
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Basically our own manual structural sharing. LGTM

PR #2148, opened 2023-12-08
Close account switcher before logging out

estrattonbailey:


src/view/com/modals/SwitchAccount.tsx
```diff
@@ -32,11 +33,15 @@ function SwitchAccountCard({account}: {account: SessionAccount}) {
...
   const isCurrentAccount = account.did === currentAccount?.did
   const {onPressSwitchAccount} = useAccountSwitcher()
+  const closeAllActiveElements = useCloseAllActiveElements()
 
   const onPressSignout = React.useCallback(() => {
     track('Settings:SignOutButtonClicked')
-    logout()
-  }, [track, logout])
+    closeAllActiveElements()
+    setTimeout(() => {
```
ansh, line 41: Curious why this `logout` needs to be inside a `setTimeout`?

gaearon, line 41: What happens without it?

gaearon (approved):

src/view/com/modals/SwitchAccount.tsx
estrattonbailey: Ah good call, added comment. Without that, the modal never finishes closing and remains open. CC @gaearon

PR #2150, opened 2023-12-08
Use new filter for Posts tab on profile

estrattonbailey:
Updates atproto deps and uses the new approach in https://github.com/bluesky-social/atproto/pull/1776

Will wait for app view deployment to merge this 👍


pfrazee (approved):
Awesome, works great

PR #2154, opened 2023-12-09
Multiple notifications fixes

pfrazee:
- c1a2819fc79546a443ffcdc48e492486789bdf7b Removes a reset of the notifications feed cache that I'm fairly sure is the cause of the random loads of the feed while you're scrolling
- d78d5c00ceb97376b6aa9fdfa6d6242978e8a501 Small tweak of grouping behavior (reverting a change I tried earlier)
- b541f94f0ec6c61d7b06fa6ac2a3da25997dd725 Takes a new approach to resetting the notifs feed query on focus which avoids losing the user's scroll state 
- 05c39c92ee30d9c57870826d47847f6ed79199f8 Is a minor optimization, bringing back the use of the cached page from unread but only if the page is known to be fresh, avoiding the issue driving its removal in the first place (losing likes by the user)

### DIFF
--- a/src/lib/notifications/notifications.ts
+++ b/src/lib/notifications/notifications.ts
@@ -81,7 +81,7 @@ export function registerTokenChangeHandler(
 
 export function init(queryClient: QueryClient) {
   // handle notifications that are received, both in the foreground or background
-  // TODO: is this needed?
+  // NOTE: currently just here for debug logging
   Notifications.addNotificationReceivedListener(event => {
     logger.debug(
       'Notifications: received',
--- a/src/lib/notifications/notifications.ts
+++ b/src/lib/notifications/notifications.ts
@@ -81,6 +81,7 @@ export function registerTokenChangeHandler(
 
 export function init(queryClient: QueryClient) {
   // handle notifications that are received, both in the foreground or background
+  // TODO: is this needed?
   Notifications.addNotificationReceivedListener(event => {
     logger.debug(
       'Notifications: received',
@@ -88,8 +89,6 @@ export function init(queryClient: QueryClient) {
       logger.DebugContext.notifications,
     )
     if (event.request.trigger.type === 'push') {
-      // refresh notifications in the background
-      truncateAndInvalidate(queryClient, RQKEY_NOTIFS())
       // handle payload-based deeplinks
       let payload
       if (isIOS) {
--- a/src/state/queries/notifications/types.ts
+++ b/src/state/queries/notifications/types.ts
@@ -28,7 +28,7 @@ export interface FeedPage {
 }
 
 export interface CachedFeedPage {
-  sessDid: string // used to invalidate on session changes
+  usableInFeed: boolean // if true, the cached page is recent enough to use as the response
   syncedAt: Date
   data: FeedPage | undefined
 }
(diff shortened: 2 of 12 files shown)


src/state/queries/notifications/unread.tsx
```diff
@@ -138,12 +138,12 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       getCachedUnreadPage() {
         // return cached page if was for the current user
         // (protects against session changes serving data from the past session)
```
estrattonbailey, line 140: Outdated comment

src/lib/notifications/notifications.ts
```diff
@@ -81,15 +81,14 @@ export function registerTokenChangeHandler(
 
 export function init(queryClient: QueryClient) {
   // handle notifications that are received, both in the foreground or background
+  // TODO: is this needed?
```
estrattonbailey, line 84: Doesn't seem like it. If someone taps on a push, we handle the invalidation and such then. No need to do anything ahead of time right now. Maybe keep this around just for debug logging?

src/state/queries/notifications/util.ts
```diff
@@ -119,8 +119,7 @@ function groupNotifications(
           Math.abs(ts2 - ts) < MS_2DAY &&
           notif.reason === groupedNotif.notification.reason &&
           notif.reasonSubject === groupedNotif.notification.reasonSubject &&
-          notif.author.did !== groupedNotif.notification.author.did &&
-          notif.isRead === groupedNotif.notification.isRead
+          notif.author.did !== groupedNotif.notification.author.did
```
estrattonbailey, line 122: I THOUGHT I saw something like this, nice

src/view/screens/Notifications.tsx
```diff
@@ -63,16 +64,26 @@ export function NotificationsScreen({}: Props) {
     }
   }, [scrollToTop, queryClient, unreadApi, hasNew])
 
+  const onFocusCheckLatest = React.useCallback(() => {
+    // on focus, check for latest, but only invalidate if the user
+    // isnt scrolled down to avoid moving content underneath them
+    unreadApi.checkUnread({invalidate: !isScrolledDown})
```
estrattonbailey, line 70: Yep that's nice

src/state/queries/notifications/types.ts
```diff
@@ -28,7 +28,7 @@ export interface FeedPage {
 }
 
 export interface CachedFeedPage {
-  sessDid: string // used to invalidate on session changes
+  usableInFeed: boolean // if true, the cached page is recent enough to use as the response
```
estrattonbailey, line 31: Nit: could chuck this in a doc block and get it on hover where we use it

estrattonbailey (approved):
Works well from my testing

src/lib/notifications/notifications.ts
pfrazee: yeah that's what I'm thinking

src/state/queries/notifications/types.ts
pfrazee: Done

src/state/queries/notifications/unread.tsx
pfrazee: Fixed

PR #2159, opened 2023-12-11
Added czech translation

ondrejholan:


mmatlak:
How does it look with the Czech translation today?

surfdude29:
The Czech translation was never added, however I'm sure the devs would be open to accepting a PR for one :)

surfdude29:
> How does it look with the Czech translation today?

@mmatlak It's a long time ago that you asked this, so apologies if this is irrelevant to you now, but a Czech localization for Bluesky was just added recently, and it's now available in version 1.126 :)

- #10994

PR #2168, opened 2023-12-11
Fix scroll on profile lists/feeds

gaearon:
It didn't work due to https://github.com/software-mansion/react-native-reanimated/issues/5488.

We already had a workaround checked in, but it wasn't prominent enough so old imports slipped through.

I renamed the export name itself consistently and added it in the places that missed it so that it's harder to copy paste the wrong thing.

[Without whitespace](https://github.com/bluesky-social/social-app/pull/2168/files?w=1)

### DIFF
--- a/src/view/com/feeds/ProfileFeedgens.tsx
+++ b/src/view/com/feeds/ProfileFeedgens.tsx
@@ -19,7 +19,7 @@ import {OnScrollHandler} from '#/lib/hooks/useOnMainScroll'
 import {logger} from '#/logger'
 import {Trans} from '@lingui/macro'
 import {cleanError} from '#/lib/strings/errors'
-import {useAnimatedScrollHandler} from 'react-native-reanimated'
+import {useAnimatedScrollHandler_FIXED} from '#/lib/hooks/useAnimatedScrollHandler_FIXED'
 import {useTheme} from '#/lib/ThemeContext'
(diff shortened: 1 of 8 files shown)


src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
```diff
@@ -14,10 +14,10 @@ import Animated, {
   interpolate,
   runOnJS,
   useAnimatedRef,
-  useAnimatedScrollHandler,
   useAnimatedStyle,
   useSharedValue,
 } from 'react-native-reanimated'
```
gaearon, line 19: The fix

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -19,7 +19,7 @@ import {OnScrollHandler} from '#/lib/hooks/useOnMainScroll'
 import {logger} from '#/logger'
 import {Trans} from '@lingui/macro'
 import {cleanError} from '#/lib/strings/errors'
-import {useAnimatedScrollHandler} from 'react-native-reanimated'
+import {useAnimatedScrollHandler_FIXED} from '#/lib/hooks/useAnimatedScrollHandler_FIXED'
```
gaearon, line 22: The fix

pfrazee (approved):

PR #2170, opened 2023-12-11
Fix scroll on native

gaearon:
Fixes crashes on native that I introduced in https://github.com/bluesky-social/social-app/pull/2168.

I forgot the Reanimated compiler looks at the function name. That's why I didn't rename it back then.

So reverting to the old function name but preserving the fix.


pfrazee (approved):
Hahaha jfc

PR #2171, opened 2023-12-11
Header blurred overscroll behaviour in profile

mozzius:
tl;dr: pulling down the header in the profile blurs the banner image and reveals a spinner. iOS only.

Video:

https://github.com/bluesky-social/social-app/assets/10959775/59de008a-2f07-44e7-a27a-af3d3f04b137

### What I've done

- Added `allowHeaderOverscroll` prop to `PagerWithHeader` which removes the clamp on the header Y value.
- Passed the `scrollY` value to the header and grows the banner container accordingly
- Overlayed a `BlurView` on the banner and bound the intensity prop to the `scrollY`

### Notes

- I had to pass the `scrollY` `sharedValue` all the way up through the render functions - not a very nice way of doing it. `react-native-collapsible-tab-view` has a `useHeaderMeasurements` hook - might be worth looking into recreating that.
- I had to install `expo-blur` because `@react-native-community/blur` wasn't playing nice with Reanimated
- Seems to be some glitchiness - you sometimes see a white flash when the banner isn't sticking in place properly. Needs further investigation
- Spinner is not actually bound to the actual refresh state - simply appears when overscrolled by greater than 50px. A real pain to pass this data all the way back up - I've yet to solve this in Graysky.
- Now the header has the `scrollY` value, you could implement a header minimum size to let the header stay fixed (see Graysky's behaviour)


pfrazee:
Okay this is nice! I appreciate you sharing this, @mozzius.

@gaearon is currently doing a heavy rework of our layout system, mainly RE how FlatLists and ScrollViews operate on the Web but it will also affect mobile. Once that's complete, he's going to be modifying how the header handles scroll actions to try to fix our dead-zone issue. After _that_ I think is the right time to give this a look.

I'm going to assign to @gaearon, but I figure it'll be a couple weeks before we can give it a serious look.

mozzius:
Superceded by #5474 🎉

PR #2175, opened 2023-12-11
Stub broadcast channel for unspported contexts

estrattonbailey:
Tested using separate tabs and the `Stub` implementation.


pfrazee (approved):
solid

PR #2176, opened 2023-12-11
Hide keyboard when pressing threadgate button

mozzius:
Just copied the same code from the other button


mozzius:
whoops, already existed

PR #2179, opened 2023-12-11
Patch bad migration, sanitize bad value

estrattonbailey:
This is NOT done. We need to handle a comma-sep locales value in the language settings too.


pfrazee (approved):
Okay looks good, gratzi

PR #2181, opened 2023-12-12
Fix View Full Thread

gaearon:
This was broken because we pass `''` in one case (from `FeedSlice`), and `isInvalidHandle('')` is `true`. I figured I'd fix it here since `isInvalidHandle` seems to be about a very specific concept that's distinct.


pfrazee (approved):
Solid

PR #2184, opened 2023-12-12
i18n settings improvements

estrattonbailey:
Just wanted to add some type hints and precautions. We can't enforce the value at the storage level unless we do another migration, but we can help ensure we continue using valid values going forward.

### DIFF
--- a/src/view/screens/LanguageSettings.tsx
+++ b/src/view/screens/LanguageSettings.tsx
@@ -21,6 +21,7 @@ import {useModalControls} from '#/state/modals'
 import {useLanguagePrefs, useLanguagePrefsApi} from '#/state/preferences'
 import {Trans, msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
+import {sanitizeAppLanguageSetting} from '#/locale/helpers'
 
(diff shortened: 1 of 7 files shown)


src/view/screens/LanguageSettings.tsx
```diff
@@ -103,7 +104,7 @@ export function LanguageSettingsScreen(_props: Props) {
           <View style={{position: 'relative'}}>
             <RNPickerSelect
               placeholder={{}}
-              value={langPrefs.appLanguage}
+              value={sanitizeAppLanguageSetting(langPrefs.appLanguage)}
```
estrattonbailey, line 107: Sanitizes existing data

src/view/screens/LanguageSettings.tsx
```diff
@@ -60,7 +61,7 @@ export function LanguageSettingsScreen(_props: Props) {
     (value: Parameters<PickerSelectProps['onValueChange']>[0]) => {
       if (!value) return
       if (langPrefs.appLanguage !== value) {
-        setLangPrefs.setAppLanguage(value)
+        setLangPrefs.setAppLanguage(sanitizeAppLanguageSetting(value))
```
estrattonbailey, line 64: Just a precaution for the future

ansh (approved):

PR #2189, opened 2023-12-12
Don't create embeds of profiles/posts from users who ask not to be sh…

ericvolp12:
…own in public views


pfrazee (approved):
probably good, I'm not a Go-ologist

PR #2191, opened 2023-12-12
Add largeHeap to AndroidManifest.xml

estrattonbailey:
Can test with `yarn prebuild -p android` and inspect the built file.

## Background

Some android users hit this error when they try to add images to posts:

CleanShot 2023-12-12 at 13 21 18@2x

It looks like this is caused by large images running out of memory (https://github.com/expo/expo/issues/12167). Increasing the heap size [may come with downsides](https://stackoverflow.com/questions/27396892/what-are-advantages-of-setting-largeheap-to-true) but without a clear repro it's hard to establish if this is in our control. If we roll this out, it should be phased so we can monitor error rates.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -121,6 +121,7 @@ module.exports = function () {
             username: 'blueskysocial',
           },
         ],
+        './plugins/withAndroidManifestPlugin.js',
       ].filter(Boolean),
       extra: {
         eas: {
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
Nice job, confirmed the manifest update works. Let's hope this helps

Issue #2185, opened 2023-12-12
Crash when going back from the profile screen


nucleartux:
**To Reproduce**

1. Open app
2. Open someone profile
3. Scroll sheet to open full profile
4. Scroll feed a little bit
5. Click back

**Screenshots**

https://github.com/bluesky-social/social-app/assets/199706/99539e17-3697-475d-bc8e-a3a3a48b627e

**Details**

 - Platform: iOS
 - Platform version:
 - App version: 1.58

PR #2194, opened 2023-12-12
fix: patch reanimated to fix crash when using `useAnimatedScrollHandler()`

haileyok:
Small patch that makes sure `animatedRef` is a function before calling in regards to https://github.com/bluesky-social/social-app/issues/2185

This seems like a check that should be there in any case, but there could also be something else with the implementation. This is mainly just a reference as to what's causing the issue 👍


gaearon:
Thanks for looking into it! I think this should do: https://github.com/bluesky-social/social-app/pull/2195

pfrazee:
Really appreciate the assists @haileyok

PR #2195, opened 2023-12-12
Clean up pager refs to fix crash

gaearon:
Fixes https://github.com/bluesky-social/social-app/issues/2185.

Previously, child page refs were registered on render and never cleaned up. The idea was that they'd get GC'd anyway with the parent eventually. However, since the onScroll handler is throttled, sometimes we'd try to access the refs when they're not good anymore. The leads to a weird crash in https://github.com/bluesky-social/social-app/issues/2185.

This PR changes refs to be synced with the React lifecycle so that they would actually get cleaned up on unmount. The throttled listener would still fire, but it would be a no-op because I added a null check before trying to adjust the scroll. To avoid over-firing, I've made `registerRef` a stable function that takes the `index`. The `index` is now passed down by the child itself.

The easiest way to test this is to follow instructions in https://github.com/bluesky-social/social-app/issues/2185 but bump the throttle delay from `queueThrottledOnScroll` to something like `1000`. Then it's easier to trigger. After the fix, that no longer shows the redbox.

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -121,13 +121,16 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
     )
 
     const scrollRefs = useSharedValue<AnimatedRef<any>[]>([])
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Good stuff

PR #2197, opened 2023-12-13
Add an intermediate List component

gaearon:
This adds a `<List>` component which currently doesn't do anything and proxies all props (and its ref) to the `FlatList` from `Views`. I think it's a good idea for a few reasons:

- It lets us constrain the `FlatList` usage across the app down to things we actually rely on
- It lets us add custom props and logic across all usages (something I might want to do soon)
- It lets us narrow down the types a bit more (or fix clunky types, as in this PR)

I kept the forks in `Views` (which pretend to be API-compatible with RN `FlatList`) but later might move some of our custom logic into the `List` instead. E.g. non-standard props like `desktopFixedHeight` would better fit in the `List`.

## Test Plan

Walked around the app.

### DIFF
--- a/src/view/com/util/ViewSelector.tsx
+++ b/src/view/com/util/ViewSelector.tsx
@@ -6,7 +6,7 @@ import {
   View,
   ScrollView,
 } from 'react-native'
-import {FlatList} from './Views'
+import {FlatList_INTERNAL} from './Views'
 import {OnScrollCb} from 'lib/hooks/useOnMainScroll'
 import {useColorSchemeStyle} from 'lib/hooks/useColorSchemeStyle'
 import {Text} from './text/Text'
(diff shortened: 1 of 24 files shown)


src/view/com/util/ViewSelector.tsx
```diff
@@ -6,7 +6,7 @@ import {
   View,
   ScrollView,
 } from 'react-native'
-import {FlatList} from './Views'
+import {FlatList_INTERNAL} from './Views'
```
estrattonbailey, line 9: Just curious, why not use `List` here?

estrattonbailey: Oh nvm https://github.com/bluesky-social/social-app/pull/2200/files#r1424875513

gaearon:
merged via https://github.com/bluesky-social/social-app/pull/2200#pullrequestreview-1780856732

PR #2200, opened 2023-12-13
Make scroll handling contextual

gaearon:
Stacks on top of https://github.com/bluesky-social/social-app/pull/2197.

This is a pretty chunky refactor but it's meant to mostly keep the data flow as is (with less plumbing).

The main change is we no longer pass down two things:

- `isScrolledDown`. This is now maintained by whoever renders `<List onScrolledDownChanged={...}>`.
- `onScroll` et al (`onBeginDrag` / `onEndDrag` / `scrollEventThrottle`). These are now in context (pager or top level).

More detailed:

- The new `List` abstraction has `onScrolledDownChanged` (similar in spirit to `onEndReached`). You can use this to track `isScrolledDown` yourself next to the list. This is nice because it is *used* very close to the list in most cases. So it's just local state where it's needed. (In the future, we could make this use intersection observer instead of scroll events.)
- The new `List` abstraction does *not* take `onScroll` at all. (We could allow that, but so far it's not used.)
- Instead, there is a new concept of a "scroll provider". It's a thing with `onScroll`, `onBeginDrag`, and `onEndDrag` properties. **`List` now delegates to the handlers from the closest scroll provider—this is the key change.**
- `useOnMainScroll` is gone, replaced by `MainScrollProvider`. That's used on Home and Notifications.
- Pager doesn't pass `onScroll` down anymore, that's replaced by a `ScrollProvider`.
- In principle scroll providers could compose but we don't have examples of that yet.
- The `List` is the one doing `useAnimatedScrollHandler` now. Which avoids spreading that across the product code. I've removed throttle customization because we just always run this on the main thread already.

Note we could still support one-off `onScroll` handlers if needed. But all current uses are contextual.

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import {
   LayoutChangeEvent,
-  FlatList,
   ScrollView,
   StyleSheet,
   View,
@@ -22,6 +21,7 @@ import {TabBar} from './TabBar'
 import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {OnScrollHandler} from 'lib/hooks/useOnMainScroll'
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
+import {ListMethods} from '../util/List'
 
 const SCROLLED_DOWN_LIMIT = 200
 
@@ -30,7 +30,7 @@ export interface PagerWithHeaderChildParams {
   isFocused: boolean
   onScroll: OnScrollHandler
   isScrolledDown: boolean
-  scrollElRef: React.MutableRefObject<FlatList<any> | ScrollView | null>
+  scrollElRef: React.MutableRefObject<ListMethods | ScrollView | null>
 }
 
 export interface PagerWithHeaderProps {
@@ -331,7 +331,7 @@ function PagerItem({
     isScrolledDown,
     onScroll: scrollHandler,
     scrollElRef: scrollElRef as React.MutableRefObject<
-      FlatList<any> | ScrollView | null
+      ListMethods | ScrollView | null
     >,
   })
 }
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -21,6 +21,7 @@ import {TabBar} from './TabBar'
 import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {ListMethods} from '../util/List'
+import {ScrollProvider} from '#/lib/ScrollContext'
 
 export interface PagerWithHeaderChildParams {
   headerHeight: number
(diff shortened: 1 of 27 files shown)


src/view/com/util/MainScrollProvider.tsx
```diff
@@ -58,13 +46,6 @@ export function useOnMainScroll(): [OnScrollHandler, boolean, ResetCb] {
   const onScroll = useCallback(
     (e: NativeScrollEvent) => {
       'worklet'
-      // Keep track of whether we want to show "scroll to top".
-      if (!isScrolledDown && e.contentOffset.y > s.window.height) {
-        runOnJS(setIsScrolledDown)(true)
-      } else if (isScrolledDown && e.contentOffset.y < s.window.height) {
```
gaearon, line 64: This logic is in `List` now and it's comparing to the threshold that was hardcoded in the pager (`200`). We could make that customizable or choose a number in the middle. Regardless, `s.window.height` didn't quite make sense because it's frozen at startup and doesn't take window resize into account.

src/view/com/util/ViewSelector.tsx
```diff
@@ -110,7 +111,7 @@ export const ViewSelector = React.forwardRef<
     [items],
   )
   return (
-    <FlatList
+    <FlatList_INTERNAL
```
gaearon, line 114: I left this as is cause it's only used in some debug view.

src/view/com/util/MainScrollProvider.tsx
```diff
@@ -102,24 +83,15 @@ export function useOnMainScroll(): [OnScrollHandler, boolean, ResetCb] {
         startMode.value = mode.value
       }
     },
-    [headerHeight, mode, setMode, isScrolledDown, startDragOffset, startMode],
```
gaearon, line 105: Thanks to pulling this out, combined with https://github.com/bluesky-social/social-app/pull/2201, we'll finally stop re-rendering `Feed` when scrolling down/up:

Screenshot 2023-12-13 at 06 01 47

src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -155,15 +150,7 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
       if (!throttleTimeout.current) {
         throttleTimeout.current = setTimeout(() => {
           throttleTimeout.current = null
-
           runOnUI(adjustScrollForOtherPages)()
-
-          const nextIsScrolledDown = scrollY.value > SCROLLED_DOWN_LIMIT
-          if (isScrolledDown !== nextIsScrolledDown) {
-            React.startTransition(() => {
-              setIsScrolledDown(nextIsScrolledDown)
```
gaearon, line 164: Pulling this down to the `List` improves perf here too.

Scrolling profile down before:

Screenshot 2023-12-13 at 06 23 31

Scrolling profile down after:

Screenshot 2023-12-13 at 06 24 43

src/view/com/util/MainScrollProvider.tsx
```diff
@@ -102,24 +83,15 @@ export function useOnMainScroll(): [OnScrollHandler, boolean, ResetCb] {
...
+      {children}
+    </ScrollProvider>
   )
-
-  return [
-    scrollHandler,
-    isScrolledDown,
-    useCallback(() => {
-      setIsScrolledDown(false)
-      setMode(false)
```
gaearon, line 122: I replaced calls to these with just `setMinimalMode(false)` because `isScrolledDown` is now managed by the list.

src/view/screens/ProfileFeed.tsx
```diff
@@ -558,21 +543,20 @@ function AboutSection({
...
   feedInfo: FeedSourceFeedInfo
   headerHeight: number
-  onScroll: OnScrollHandler
   scrollElRef: React.MutableRefObject<ScrollView | null>
   isOwner: boolean
 }) {
   const pal = usePalette('default')
   const {_} = useLingui()
-  const scrollHandler = useAnimatedScrollHandler(onScroll)
+  const scrollHandlers = useScrollHandlers()
```
gaearon, line 558: Note: this is interesting because it shows how to get the contextual handlers when you're not using a `List`.

pfrazee (commented):
Okay perfect, much clearer to me now where you're headed with this. I'm +1.

Let's hold off merging until after today's hotfix release is created so we have more time for QA

src/view/screens/ProfileFeed.tsx
estrattonbailey: Do you think we'll ever want to replace this `ScrollView` with an impl similar to `List`?

estrattonbailey (commented):
This looks great, haven't pulled and tested anything yet though

src/view/screens/ProfileFeed.tsx
gaearon: So far it looks like almost all long scrolls we have are lists but could be done

pfrazee (approved):
Yep works great, great stuff

PR #2214, opened 2023-12-15
Bump react-native to 0.73.1, expo to 50

estrattonbailey:
Couldn't get 0.73.0 to work with Expo 49, which was expected. Some changes to Podfile config and such seems to be handled by internal Expo plugins, and we'd need to replicate that to make it work.

So I upgraded to Expo 50 beta, and upgraded all expo packages to `@next` tags for compat. Thankfully Expo has good npm hygiene! 

Steps to get this set up on your machine:
1. Upgrade to Java 17 using brew or whatever you've used before
2. I nuked Android Studio, made sure [this box was checked](https://stackoverflow.com/questions/17727645/how-to-update-gradle-in-android-studio), and pointed Gradle at Java 17
3. Run `yarn nuke` (new script)
4. Then `yarn && yarn prebuild`

For reference:
- [RN upgrade guide](https://reactnative.dev/blog/2023/12/06/0.73-debugging-improvements-stable-symlinks)
- [RN upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.72.7&to=0.73.0)
- [Expo 50 beta announcement](https://expo.dev/changelog/2023/12-12-sdk-50-beta)


app.config.js
```diff
@@ -110,6 +110,9 @@ module.exports = function () {
         [
           'expo-build-properties',
           {
+            ios: {
+              deploymentTarget: '13.4',
```
estrattonbailey, line 114: @pfrazee special call out for this

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "bsky.app",
   "version": "1.59.0",
   "private": true,
+  "engines": {
+    "node": ">=18"
+  },
```
estrattonbailey, line 7: Also important to note, though I think we're probably all on 18+ already

package.json
```diff
@@ -135,11 +140,11 @@
     "react-avatar-editor": "^13.0.0",
     "react-circular-progressbar": "^2.1.0",
     "react-dom": "^18.2.0",
-    "react-native": "0.72.5",
+    "react-native": "^0.73.0",
     "react-native-appstate-hook": "^1.0.6",
     "react-native-drawer-layout": "^3.2.0",
     "react-native-fs": "^2.20.0",
-    "react-native-gesture-handler": "^2.12.1",
+    "react-native-gesture-handler": "~2.14.0",
```
estrattonbailey, line 147: Both this and screens needed to be updated to unblock build errors

patches/babel-preset-expo+9.5.2.patch
estrattonbailey, line 1: @gaearon cool to axe this? Patch failed otherwise. I can re-add this and fix the patch if we need it.

src/Splash.tsx
```diff
@@ -36,8 +36,6 @@ type Props = {
   isReady: boolean
 }
 
-SplashScreen.preventAutoHideAsync().catch(() => {})
```
estrattonbailey, line 39: After updating, I was getting a `cannot read catch of undefined` error. This is already called in `App.native.tsx`. Removing this fixed 🤷

package.json
ansh: Probably good to run `npx expo install --fix` just so we are at least at the package numbers that v50 wants

patches/react-native-pager-view+6.1.4.patch
pfrazee, line 1: This patch does not appear to be needed anymore. `pager-view`'s latest seems to have updated everything we need; the gesture works smoothly with the left-hand drawer (the original need of the patch) and overscroll works correct in the image viewer (an update to the patch).

src/platform/crypto.ts
pfrazee, line 1: irritating hack

src/lib/hooks/usePermissions.web.ts
pfrazee, line 1: unclear why `expo-camera` doesn't work at all on web now but this saves us from that issue

pfrazee:
Whelp, here goes nothing

PR #2215, opened 2023-12-15
Traffic reduction and tuned caching strats

pfrazee:
- b50e5d82f6f67450196271cf7f175c5edac205be Updates the post Feed component to 1) reduce an extraneous poll request on load, and 2) do a full reset after 60 minutes.
- 551b026d7a88f48a3bfa1531a8cb232c0a637611 Removes the `isFeedPublic` query. It's not needed now, but in the future we need to find another solution because this was adding a lot of low-value traffic.
- f45ac60798f546df3eb25f201ea0c2c6106f66a1 Reduces how frequently we poll for updates by changing the timers.
- 5ecf724cf399c3fb653ef8f6eff129cb3d815d87 Guards against a double-fire of `fetchNextPage` which unfortunately can happen because `isFetchingNextPage` doesn't update immediately. It does this by just tracking the page count at the time of last trigger.
- be87c370abc14c3442788ccab27883cf57dd007a Disables the feeds (and thus stops polling) when the ProfileFeed and ProfileList screens are blurred, reducing needless polling.


pfrazee:
yolo merge

PR #2216, opened 2023-12-15
Consolidate List props 

gaearon:
### `refreshControl.* -> onRefresh + refreshing`

We always render the same `RefreshControl`. Moreover, this prop has some weird behaviors and from what I recall you can’t actually put some different kind of content there. So I think it makes sense to tighten the control for it and to get back to the `refreshing` and `onRefresh` shortcut props. This is the main FlatList API for this anyway — I think we only use the `refreshControl` because we wanted custom colors and offset, which doing it in `List` solves.

(If we later decide to support a customizable `refreshControl` again, we should make it a render prop like `refreshControl({ headerOffset })` because it needs to know the layout information.)

---

### `style.paddingTop + contentOffset.y -> headerOffset`

These are always either missing or used together for the same goal, and in those scenarios forgetting one would cause a bug. Let's lock it down a little bit under a single concept. We could in principle add `contentOffset` as a separate concept back, but so far we haven’t needed it, so I do not expose `contentOffset` from in the `List` wrapper.

## Why

Originally, my motivation was `headerOffset`. It is [useful](https://github.com/bluesky-social/social-app/pull/2126/files#r1427591304) to preserve as a concept until the end instead of losing it in the middle. That makes it easier to understand what to do in special cases or on different platforms. I pulled the `refreshControl` change in along the way because that control also has to "know" the `headerOffset` too and is also always the same. Seems easy enough to support again if it needs to diverge.

As a bonus, these changes would let us later optimize `List` re-renders by making each prop memoizable again. One notable exception there is `keyExtractor` which is always inline right now but we could make that one a default extractor in a future PR. For now I want to keep it manageable for review.

## Test Plan

Walked around the lists on web and Android.

My simulators seem busted after upgrading macOS. I'll try to figure that out tomorrow.

### DIFF
--- a/src/view/com/util/List.tsx
+++ b/src/view/com/util/List.tsx
@@ -1,27 +1,42 @@
 import React, {memo, startTransition} from 'react'
-import {FlatListProps} from 'react-native'
+import {FlatListProps, RefreshControl} from 'react-native'
 import {FlatList_INTERNAL} from './Views'
+import {addStyle} from 'lib/styles'
 import {useScrollHandlers} from '#/lib/ScrollContext'
 import {runOnJS, useSharedValue} from 'react-native-reanimated'
 import {useAnimatedScrollHandler} from '#/lib/hooks/useAnimatedScrollHandler_FIXED'
+import {usePalette} from '#/lib/hooks/usePalette'
 
 export type ListMethods = FlatList_INTERNAL
 export type ListProps<ItemT> = Omit<
   FlatListProps<ItemT>,
-  'onScroll' // Use ScrollContext instead.
+  | 'onScroll' // Use ScrollContext instead.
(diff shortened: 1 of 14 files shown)


src/view/com/util/List.tsx
```diff
@@ -49,11 +64,35 @@ function ListImpl<ItemT>(
...
+        titleColor={pal.colors.text}
+        progressViewOffset={headerOffset}
+      />
+    )
+  }
+
+  let contentOffset
+  if (headerOffset != null) {
+    style = addStyle(style, {
+      paddingTop: headerOffset,
```
gaearon, line 83: Note: this overrides it. This is a bit blunt but maybe also more obvious when it's wrong than if I tried to sum them up or something. I decided throwing on it might be a bit much. Maybe could console warn.

pfrazee (approved):

PR #2217, opened 2023-12-15
3rd party embed player

haileyok:
Continuing from this: https://github.com/bluesky-social/social-app/pull/2107

On all devices:
[https://streamable.com/p48euh?src=player-page-share](https://streamable.com/p48euh?src=player-page-share)

<details>

<summary>Screenshots</summary>

Screenshot 2023-12-19 at 7 02 36 PM
Screenshot 2023-12-19 at 7 02 26 PM
Screenshot 2023-12-19 at 7 02 16 PM
Screenshot 2023-12-19 at 7 02 04 PM
Screenshot 2023-12-19 at 7 01 44 PM
Screenshot 2023-12-19 at 7 01 37 PM
Screenshot 2023-12-19 at 7 01 26 PM

</details>

The whole "player" area is pressable to begin playback. The area below the player is pressable to redirect to the external site.

### Twitch parent param, needed origin
From what I can tell, `localhost` works for the parent param and there is no need for an origin.

### Dark mode
No needed changes since this uses the same styling as other external embeds

### No Image Render
See below

### Android
Setting `collapsable` to false on the player allows us to measure it.

### Remove hack
I'm assuming this was only for being able to use the local dev server. Is that the case?

### Testing
Added test cases for parsing links from all embed sources.

### Move render logic to `ExternalLinkEmbed`
- Memoize the player params
- Conditionally render the regular external thumbnail or the player view based on whether we have player params or not
- Allows us to easily keep all the other styling that exists for external embeds

### Blur event listener
The webview is not unloaded after we navigate away from whatever screen we are on. Listening for the blur event and running `onLeaveViewport()` resolves this.

### Add SoundCloud
Use the SoundCloud embed URL for tracks and sets (playlists or albums)

### Add YouTube Shorts
Uses the same embed as regular YouTube videos. Just added an additional check in the parse function.

### Change Spotify album size
Was a tad bit tall on smaller screens and made scrolling away from it difficult. We can go smaller, but then Spotify's styling shrinks the player to the same size as a track. Wasn't a great look to me personally.

### Use absolute "layers" for the placeholder and the player
- Image is rendered whenever a thumbnail is available and the player is either not active or is loading. No re-rendering this image when changing between states
- Placeholder is placed on top of the image at a zIndex of 2. Changes states from either a play button or an activity indicator.
- YoutubePlayer or WebView is rendered ontop of the image at a zIndex of 3. This allows us to avoid any flickering or resizing issues. By setting the WebView's background color to transparent, we continue to see the placeholder until the content has loaded.
- Also avoids any odd differences in rendering between platforms.

### No Image Render
This one is a bit annoying, maybe someone has some ideas. Ideally I would like to be able to display something like this, except with the size of the player much shorter, maybe around 80-100px then change the size once the player has been started. (This would be nice for some of the taller players like Spotify/SoundCloud playlists too)

<details>
<summary>Example Screenshot</summary>
	
Screenshot 2023-12-19 at 6 10 18 PM

</details>

The issue with that is when we scroll away, we then set the player to inactive again, which changes the size back and causes a jump in scroll position. I can add some additional logic to check if we have already played the video once and maintain whatever the content's size is.

Alternatively - since I assume that most cases of not having an image are from the poster not completing the thumbnail upload for whatever reason - we could try and fetch it whenever the element comes into the feed, but I'm not sure if that's a good idea or if it's something you'd want to do. So for now, it's just an empty space as shown.

### Use StyleSheet API
Just cleaned up a bunch of the inline styles


pfrazee:
This wonderful work again, @haileyok. I ran some QA on all platforms and everything seemed to work just right.

> I'm assuming this was only for being able to use the local dev server. Is that the case?

Yep, good to remove.

> The issue with that is when we scroll away, we then set the player to inactive again, which changes the size back and causes a jump in scroll position. I can add some additional logic to check if we have already played the video once and maintain whatever the content's size is.

I think this may be the right move for now.

Did I miss any other open questions?

pfrazee:
Great!

pfrazee:
People are going to really appreciate this. I hope we can get it out soon

pfrazee:
@haileyok oh right. I'm thinking the move is to make sure the player's updated height gets kept even after it's closed to ensure we don't move the feed around on people

pfrazee:
Okay I'll see how it feels with some testing

pfrazee (approved):
LFG

higgins:
it's circle back season...so excited to see this merged! thank you!!! 🙇‍♂️

PR #2227, opened 2023-12-18
bskyweb: update golang indigo dep (from May!), and some small devex tweaks

bnewbold:
Amazingly hadn't updated the indigo package since May 2023. Guess we've been pretty stable!

There were only a couple *required* lines of changes, but I also had the refactor to "flatten" the code structure of some methods in my working tree, and it was hard to untangle those commits. No change in behavior expected here. Did some minimal local testing (logged-out!) against prod and seems to be working fine.


Jacob2161 (approved):
:+1:

PR #2230, opened 2023-12-18
Fix `Logotype` fill for dark mode

estrattonbailey:


pfrazee (approved):
[shakes first at dark mode]

PR #2236, opened 2023-12-18
Move label appeal to post dropdown

estrattonbailey:
Applies for the highlighted post in a thread only. Other post views do not allow appealing their labels.


pfrazee (approved):
Good stuff

PR #2238, opened 2023-12-18
Set numberOfLines for Android text input

estrattonbailey:
Followed docs to set a fixed # of lines. I used a red background to view changes in the inputs height, and on small phones and a link card, the input does get squished. This ensures there's always at least 4 lines of space.

As per docs, this also required adding the `textAlignVertical` for Android.

Screenshot 2023-12-18 at 2 21 00 PM


pfrazee (approved):
Cool cool. (Worth noting: numberOfLines is android only)

PR #2239, opened 2023-12-18
Fix 2231 Modal overlap on double modal open

ansh:
This is a known bug. Basically the issue was that the modal was being `expanded` to the largest position. Now we just snap to the first position.


pfrazee (approved):
Awesome yeah

PR #2249, opened 2023-12-19
Sketching

estrattonbailey:


src/view/com/posts/FeedItem.tsx
```diff
@@ -130,18 +130,19 @@ let FeedItemInner = ({
...
+    paddingLeft: 'm',
+    pr: 'l',
+    cursor: 'pointer',
+    overflow: 'hidden',
+    borderColor: 'l2',
+    paddingBottom:
+      isThreadLastChild || (!isThreadChild && !isThreadParent)
+        ? 's'
+        : undefined,
+  }), []))
```
estrattonbailey, line 145: Didn't replace `Link`, so stuck with the style object. Easier to read like this and know what styles are being applied.

src/view/com/posts/FeedItem.tsx
```diff
@@ -152,8 +153,8 @@ let FeedItemInner = ({
       accessible={false}>
       <PostSandboxWarning />
 
-      <View style={{flexDirection: 'row', gap: 10, paddingLeft: 8}}>
-        <View style={{width: 52}}>
+      <Box row gap='s' pl='s'>
```
estrattonbailey, line 156: Terse, nice imo

src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -72,54 +70,54 @@ export function ProfileSubpageHeader({
...
+        <Box
+          row aic
+          pt={web('s')}
+          pb='s'
+          px='m'
+          borderColor='l3'
+          borderBottomWidth={1}
+          gtMobile={{
+            px: 14
+          }}>
```
estrattonbailey, line 84: Some of the styles in here could probably be simplified when we take a more holistic approach to components

src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -72,54 +70,54 @@ export function ProfileSubpageHeader({
...
+            px={canGoBack ? 0 : 6}
             accessibilityRole="button"
             accessibilityLabel={canGoBack ? 'Back' : 'Menu'}
             accessibilityHint="">
             {canGoBack ? (
               <FontAwesomeIcon
                 size={18}
                 icon="angle-left"
-                style={[styles.backIcon, pal.text]}
+                style={{ marginTop: tokens.space.s, color: tokens.color.l5 }}
```
estrattonbailey, line 99: A wrapper around icons would be a early todo so we can just style them directly using attrs

src/view/com/util/LoadingPlaceholder.tsx
```diff
@@ -42,58 +43,57 @@ export function PostLoadingPlaceholder({
...
-          <View style={s.flex1}>
+      <Box flex={1}>
+        <Box gap='s'>
+          <Box bg='l1' width={100} height={6} radius='s' />
+          <Box bg='l1' width="95%" height={6} radius='s' />
+          <Box bg='l1' width="95%" height={6} radius='s' />
+          <Box bg='l1' width="80%" height={6} radius='s' mb='s' />
+        </Box>
+        <Box row>
+          <Box column>
```
estrattonbailey, line 66: I replaced all this with Box just to show how easy it is to make basic components

src/view/com/util/fab/FABInner.tsx
```diff
@@ -18,6 +20,26 @@ export function FABInner({testID, icon, ...props}: FABProps) {
...
+    },
+    outer: {
+      position: 'absolute',
+      zIndex: 1,
+    },
+    inner: {
+      justifyContent: 'center',
+      alignItems: 'center',
+    },
+  }), []))
```
estrattonbailey, line 42: Could definitely do this better with some better button/pressable primitives, but I did this just to show that we _can_ just copy pasta in the defs from `StyleSheet` if there are cases where we must.

For entirely static styles in small specific components like this — especially in perf-intensive situations — maybe StyleSheet is still the way to go.

src/view/com/util/load-latest/LoadLatestBtn.tsx
```diff
@@ -20,61 +21,62 @@ export function LoadLatestBtn({
   label: string
   showIndicator: boolean
 }) {
-  const pal = usePalette('default')
   const {isDesktop, isTablet, isMobile} = useWebMediaQueries()
   const {fabMinimalShellTransform} = useMinimalShellMode()
 
+  const tokens = useTokens()
+  const styles = useStyles({
```
estrattonbailey, line 28: Again didn't really need to do this one here, but few of the components it applies to accept style props (which they could in our ideal world) and it was easy to copy over + part of this draft PR is testing perf so might as well

src/view/screens/ProfileFeed.tsx
```diff
@@ -335,44 +332,47 @@ export function ProfileFeedScreenInner({
...
+              </Button>
+            </>
+          )}
+
+          <NativeDropdown
+            testID="headerDropdownBtn"
+            items={dropdownItems}
+            accessibilityLabel={_(msg`More options`)}
+            accessibilityHint="">
+            <Box bg='l1' py='s' px='m' radius='round'>
```
estrattonbailey, line 362: We have a lot of ad-hoc re-used UI like this that we should better abstract. But even so, it's easy to create shapes and layouts on the fly

PR #2250, opened 2023-12-20
Add `yarn intl:compile` to postinstall hook

mozzius:
This PR adds `yarn intl:compile` to the postinstall hook.

```diff
-    "postinstall": "patch-package",
+    "postinstall": "patch-package && yarn intl:compile",
```

This trips me up every time - hoping that this reduces the fricton of the i18n stuff for new contributors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "prepare": "is-ci || husky install",
-    "postinstall": "patch-package",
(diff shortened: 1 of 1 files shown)


ansh (approved):
Agreed with this but I'll let @pfrazee give the final stamp

pfrazee (approved):
Sg!

PR #2251, opened 2023-12-20
Invalid handle autocomplete

mozzius:
Fixes autosuggestion so it doesn't match on `'invalid.handle'` if the user has an invalid handle.

Before:
Screenshot 2023-12-20 at 10 04 58

After:
Screenshot 2023-12-20 at 10 05 04

I also found a couple of places where a comparison was done to `'invalid.handle'` and replaced it with the `isInvalidHandle` utility.


pfrazee (approved):
Nice thanks!

PR #2261, opened 2023-12-21
Bump react-native-drawer-layout@4.0.0-alpha.3

pfrazee:
While testing android in the sim I was able to trigger one of our crash conditions. It happened when using the slide gesture to open the drawer and releasing the gesture. The error stack tracked with https://github.com/react-navigation/react-navigation/commit/3cfb3e63949f0aa6f4b14db02161dd88fd10cb12 which merged in [4.0.0-alpha.1](https://github.com/react-navigation/react-navigation/blob/main/packages/react-native-drawer-layout/CHANGELOG.md#400-alpha1-2023-09-25). Tested this, seemed to work fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "react-dom": "^18.2.0",
     "react-native": "0.72.5",
     "react-native-appstate-hook": "^1.0.6",
-    "react-native-drawer-layout": "^3.2.0",
+    "react-native-drawer-layout": "^4.0.0-alpha.3",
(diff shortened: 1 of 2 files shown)


mozzius:
Be aware that this version breaks drawers that come in from the right (although I don't think this app has any of those)

gaearon (approved):

pfrazee:
@mozzius appreciate the heads up

PR #2264, opened 2023-12-21
minor search screen ux improvements

haileyok:
Hopefully I am not PR spamming here...apologies if so. This is just some very minor changes but with very noticeable UX improvements:

- Add a hitslop to the cancel and clear buttons on the search screen (and the clear button on the feeds screen search). These are currently rather difficult to press
- Add `keyboardShouldPersistTaps` to`SearchScreenSuggestedFollows`, `SearchScreenMobile` (autocomplete results), and `Feeds`.  Currently you have to dismiss the keyboard before any interaction with the screen can be made.
- Dismiss the keyboard on scroll on `SearchScreenSuggestedFollows`, `SearchScreenMobile`, and `Feeds`. This resolves issues where some content is hidden below the keyboard.

On the last issue, this is certainly a good fix (both in terms of ease and feel). The alternative would be to adjust the content insets when the keyboard is visible. Unfortunately I can't seem to get that to work on `SearchScreenMobile` after the `ScrollView` re-renders. `automaticallyAdjustKeyboardInsets` is known to be a bit buggy, and there are some way better alternatives available if that's something you want to explore: See https://twitter.com/hirbod_dev/status/1668528361457451009 (sorry for bird link)

UX wise, I think that dismissing the keyboard on scroll is the most common behavior, so I don't think there's a big need in this particular case. (I also think this is default behavior on Android even without the prop, but I'm not a big Android user so don't quote me on that)

RocketSim_Recording_iPhone_15_Pro_2023-12-21_01 11 54

RocketSim_Recording_iPhone_15_Pro_2023-12-21_01 12 40

### DIFF
--- a/src/view/com/util/forms/SearchInput.tsx
+++ b/src/view/com/util/forms/SearchInput.tsx
@@ -16,6 +16,7 @@ import {useTheme} from 'lib/ThemeContext'
 import {usePalette} from 'lib/hooks/usePalette'
 import {useLingui} from '@lingui/react'
 import {msg} from '@lingui/macro'
+import {HITSLOP_10} from 'lib/constants'
 
 interface Props {
   query: string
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
literally had this on my todo list, another awesome contrib. thank you!

estrattonbailey (approved):
Yesssss

pfrazee:
I still need to test this before we merge but I'll try to today

pfrazee (approved):
Yeah great stuff as expected, thanks so much hailey

Issue #1740, opened 2023-10-24
French Translation


gorlanova:
**Is your feature request related to a problem? Please describe.**

No previous issue yet
I would like to help translating the web app / mobile app to French

**Describe the solution you'd like**

If I can access a file containing all english strings / labels, I will translate them to French (I'm French) and commit them here

**Describe alternatives you've considered**

None (pardon me if there are any, I'm not an i18n expert)

**Additional context**

No additional context to provide, but here's a baguette emoji 🥖

iamwinner422:
Wanna help for French Translation too

pfrazee:
I appreciate it! We have "internationalization" tooling on our roadmap for the next couple months. I'm going to close this issue for now but we'll keep everyone updated.

PR #2266, opened 2023-12-21
Add German localization

ansh:
This PR adds German localization which has been verified internally. However, we are open to any community feedback as we work on translating the app into German!

This PR should be merged after #1740


estrattonbailey (approved):

ansh:
Closed in favor of #2267

PR #2270, opened 2023-12-22
CI: run all actions on every push

bnewbold:
It sounds like from chat that maybe we *don't* actually want to do it this way for the `social-app` repo specifically, but this PR at least makes the issue clearer.

The github docs on what other "triggers" can be used to control when CI runs are here:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows


estrattonbailey (commented):
Yeah for lint/test, we might as well be doing that on every `pull_request` at least. I'm not sure how much it helps us to have it run upon `push` into main, for instance, but I could be wrong.

PR #2284, opened 2023-12-23
Mark more text for localization

noritada:
This PR adds marks for localization to more texts.
Only these messages are left out, while the texts before and after are subject to localization.

Many thanks!


pfrazee (approved):
Appreciate it!

PR #2289, opened 2023-12-23
yet more RSS tweaks

bnewbold:
- use the fully qualified URL in the profile view HTML head `<link>`
- use the actual request Host as hostname in RSS (not hardcoded to `bsky.app`). note that it is still hard-coded to `https://`, so not perfect for local dev, but should fix staging etc
- handle `/profile/{handle}/rss` as an HTTP 302 (non-permanent) redirect to `/profile/{did}/rss`, as a UX affordance

I think this the is last round of enhancements I can afford to spend time on right now. Eg, not planning to bite off Atom feeds in the near future (though it probably isn't a ton of work, can just copy the current RSS code).

cc: @scripting

### DIFF
--- a/bskyweb/cmd/bskyweb/rss.go
+++ b/bskyweb/cmd/bskyweb/rss.go
@@ -39,6 +39,7 @@ type rss struct {
 
 func (srv *Server) WebProfileRSS(c echo.Context) error {
 	ctx := c.Request().Context()
+	req := c.Request()
 
 	didParam := c.Param("did")
 	did, err := syntax.ParseDID(didParam)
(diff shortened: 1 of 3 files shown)


pfrazee (approved):

scripting:
@bnewbold -- 

thanks for these fixes, i had already [updated feedland](http://scripting.com/2023/12/23/141939.html) so it understood relative urls in `<links>` but there's a lot of software out there that probably wasn't going to work with a relative url there. 

when you're ready to do some more work in this area, i'd be happy to review the ideas. 

dave

bnewbold:
@scripting thanks for all the review and feedback!

Issue #2277, opened 2023-12-22
[APP-923] Muted & blocked users not showing in the muted & blocked users lists


pfrazee:
The profilecard moderation filter is incorrectly applying in the moderation screen's listings interfaces

<sub>[APP-923](https://linear.app/blueskyweb/issue/APP-923/muted-and-blocked-users-not-showing-in-the-muted-and-blocked-users)</sub>

heikadog:
i wanna note that it's also not showing in any mod lists !

pfrazee:
This should be fixed

PR #2290, opened 2023-12-23
Fix some account filtering behaviors

pfrazee:
I hastily added a moderation filter on profile cards while adding the PWI behaviors. This missed that we don't want this behavior in 3 settings:

1. Muted users list
2. Blocked users list
3. Curation/moderation user lists

This PR fixes that. I also reviewed the other cases ProfileCard is used and I believe we do want this filter in all of them.

Additionally, the `moderation.account.filter` is true when users are muted, and in some cases that's too extensive (autocomplete, any user listing) so this relaxes that as well.


pfrazee:
Closes #2277

gaearon (approved):

PR #2297, opened 2023-12-24
Patches and compiled i18n

estrattonbailey:
Bumps a couple patches by a minor version, extracts missing translations, and removes 3 compiled translations that snuck back in.


pfrazee (approved):
Nice

PR #2299, opened 2023-12-24
Hide posts tool

estrattonbailey:


pfrazee:
Well I tried to make an e2e test for this but somehow our detox config got FUBARed by the recent react native upgrade so I need to put some time into that

pfrazee:
Gotta yolo merge this due to time pressure

PR #2302, opened 2023-12-24
Apply post-hide to quote posts

pfrazee:
Followup to https://github.com/bluesky-social/social-app/pull/2299, apply the "hide post" behavior to quotes


pfrazee:
Another time pressure yolo merge

PR #2311, opened 2023-12-26
use the window hostname on web for twitch embeds

haileyok:
Twitch requires parent param, use the window hostname if on web, otherwise use `localhost` on native.


pfrazee (approved):
🙏 yep looks good, gratzi

PR #2328, opened 2023-12-27
Stop the user from posting before the link card finishes loading

pfrazee:
There have been a lot of attempts to share youtube videos where no metadata is extracted into the link card. The only way I was able to reproduce that was by not waiting for the link card to finish fetching.

This isn't an ideal way to solve that, but it does solve it: it stops the user from posting while it's loading.

CleanShot 2023-12-26 at 19 25 28@2x

Closes #2328


pfrazee:
Confirmed by discussing with users this is the cause.

estrattonbailey (approved):

PR #2210, opened 2023-12-14
Mark more text for localization

quiple:

PR #2334, opened 2023-12-27
Add Korean localization

lens0021:
This PR adds Korean localization. As a Korean speaker and as a Bluesky use, I have translated few messages, which are frequently exposed I think.


lens0021:
I didn't check there was an existing PR.
Should be rebased onto #2210, or maybe closed.

ansh:
It would be great if you can check the work done in #2210 @lens0021

PR #2337, opened 2023-12-27
Update IS_PROD check

### LINKED ISSUE #2313: [APP-931] A user is reporting they see the "SANDBOX" watermark on their posts
I asked if they do anything unusual with their signin. They said they had to run forgot password recently and this is the result. I wonder if we somehow created a session for them using their actual server URI.

estrattonbailey:
Fix #2313 

What this PR does is **assumes we're in production unless we match a localhost PDS URL, OR our staging PDS URL.** This means that the SANDBOX warning labels **only show for local dev and on staging.**

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -45,7 +45,15 @@ export function IS_STAGING(url: string) {
 }
 
 export function IS_PROD(url: string) {
-  return !IS_LOCAL_DEV(url) && !IS_STAGING(url)
+  // NOTE
+  // until open federation, "production" is defined as the main server
+  // this definition will not work once federation is enabled!
+  // -prf
+  return (
+    url.startsWith('https://bsky.social') ||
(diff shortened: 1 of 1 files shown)


pfrazee (changes requested):
Nah this isn't right. The `url` being tested is the PDS.

pfrazee:
Yo @dholms we're hitting situations where users are seeing the SANDBOX watermark now that internal federation has deployed. It's rare but it can happen. We don't have a thorough listing of the federation sandbox hostnames and so we don't really have a way to establish whether they are in the federation sandbox or not. This PR will essentially ensure prod users never get the watermark but it will also break it for the federation sandbox

estrattonbailey:
Update: this still won't work in a full-federated world — so no change from present — but for now it solves our problem of not telling users it's "sandbox" if they log in with their PDS URL directly.

PR #2339, opened 2023-12-27
Detect offline state and alert user

ansh:
This PR adds a check for seeing if there is no internet connection before attempting to resume a session

CleanShot 2023-12-27 at 22 38 16@2x


estrattonbailey:
Going to fix in #2399

PR #2340, opened 2023-12-27
Remove patched color scheme code

estrattonbailey:
The [PR in question](https://github.com/facebook/react-native/pull/39439) was merged in late September.

I can't find what release it was actually included in, but according to [their release docs](https://reactnative.dev/contributing/release-faq#when-will-my-fix-make-it-into-a-release), it should be in there by now. Plus, I built locally and couldn't replicate what we saw before: flashing of the theme dark-light-dark upon foregrounding the app.


pfrazee (approved):
Cool

Issue #2326, opened 2023-12-27
[APP-935] Embed video playback stops if you go landscape


pfrazee:
With the youtube player

>  If you start a video, then maximise it and then rotate the screen to landscape- video stops/disappears and you are back in BlueSky app. So maximise is possible but only if you stay in portrait orientation.

<sub>[APP-935](https://linear.app/blueskyweb/issue/APP-935/embed-video-playback-stops-if-you-go-landscape)</sub>

PR #2349, opened 2023-12-27
Fix various issues with screen dimension detection in external player

haileyok:
This resolves https://github.com/bluesky-social/social-app/issues/2326 as well as normalizes the active/inactive logic on all platforms.

1. The position of the item in the feed never actually changes on native, since landscape is disabled in the app. Therefore, when the screen rotates in the native full screen player, we need to take that into account. Here we will use either the window height when in portrait or the window width when in landscape to get the "real height" of the feed. Desktop behavior stays the same. This is working on both iOS and Android now.
2. Instead of checking against static values like 0 and screenHeight when determining whether or not to dismiss the player, we instead use the safe area insets.
 - If the bottom of the player reaches the top of the safe area, then we dismiss. Likewise, if the top reaches the bottom of the safe area we do the same.
 - This normalizes dismiss behavior on all platforms. Currently on Android, the video may dismiss even if it is just *barely* off screen. On web, the player would sometimes not dismiss until significant scrolling had occurred.
 - This logic works on all platforms, so the only platform-specific check that has to be done is the check from # 1
3. Stop running the interval if the player isn't actually active.

If we were only concerned with native, we could have just used the static `Dimensions.get('window')` outside of the component and it would remain stable, removing the need to check. This won't work on web though since the user can resize their window as they desire.

### DIFF
--- a/src/view/com/util/post-embeds/ExternalPlayerEmbed.tsx
+++ b/src/view/com/util/post-embeds/ExternalPlayerEmbed.tsx
@@ -1,10 +1,10 @@
 import React from 'react'
 import {
   ActivityIndicator,
-  Dimensions,
   GestureResponderEvent,
   Pressable,
   StyleSheet,
+  useWindowDimensions,
   View,
 } from 'react-native'
 import {Image} from 'expo-image'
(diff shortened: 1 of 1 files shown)


pfrazee:
This seems solid, we good to merge now or do we want to wait for those additions?

estrattonbailey:
This is great stuff 🚀 thanks for all the details!

pfrazee (approved):
Incredible work hailey. Tested on all 3 platforms and works great. Very clean solution (I wasnt aware of that renaimated api at all) and wonderful writeup. Just superb.

Issue #2344, opened 2023-12-27
[APP-939] Reporting profiles broken on android


pfrazee:
https://bsky.app/profile/furrydev.xyz/post/3khjbtaxbzi27

>  The textbox doesn't accept input until you put away your keyboard, plus pressing Send Report does nothing.

A video is available in the link

pfrazee:
Fixed by https://github.com/bluesky-social/social-app/pull/2351

PR #2351, opened 2023-12-28
Disable BlurView on android

pfrazee:
Something is horribly wrong here.

We've had various reports from users that a `<TextInput>` will drop user input. Then we got #2344 reported, which gave us two lucky breaks:

1. [A video](https://www.youtube.com/watch?v=e66IKwV7vi4&t=17s)
2. I could repro it in the simulator, thank god

With the repro I massively struggled to figure out the _cause_ of this. Here are all of the bizarre behaviors:

- The input does actually occur, it's just that the text is invisible. You can actually select the invisible text.
- The `onChange` fires and the value is updated and the correct string is passed into the `<TextInput>`.
- Despite the value updating, the `<CharCounter>` wasn't updating either.

I tested a variety of possible causes. I pulled out everything to do with BottomSheet. I removed all properties on the `<TextInput>`. I turned off autocomplete, autocorrect, etc. I also googled this aggressively and tried a variety of known hackfixes. None of this had any effect.

The `<CharCounter>` failing to update despite the value changing was the final clue. That led me to do 569ce385f8870ed5f6c47fff1390bb61825a7b34 which simply sets a `key` on _any_ component in the render tree, and bizarrely that fixed both of the problems.

What this suggests to me is that somehow the render is failing to make its way through the react/react-native pipeline. I have absolutely no idea how that could be happening.

---

UPDATE

My god. I kept digging after this and discovered that running a `setInterval` in the header of a profile would show similar characteristics -- the UI would just stop updating -- but in the header of the home screen this didn't happen. Then I noticed that scrolling down a little bit in the profile fix this problem entirely.

Turns out, the `@react-native-community/blur` `<BlurView>` being used for the background of the back button was to blame. As soon as that was out of view, everything worked fine.

I need a drink.

### DIFF
--- a/src/view/com/modals/report/InputIssueDetails.tsx
+++ b/src/view/com/modals/report/InputIssueDetails.tsx
@@ -42,7 +42,8 @@ export function InputIssueDetails({
         accessibilityHint="Add more details to your report">
         <FontAwesomeIcon size={18} icon="angle-left" style={[pal.link]} />
         <Text style={[pal.text, s.f18, pal.link]}>
-          <Trans> Back</Trans>
+          {' '}
+          <Trans>Back</Trans>
         </Text>
       </TouchableOpacity>
       <View style={[pal.btn, styles.detailsInputContainer]}>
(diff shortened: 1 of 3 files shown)


pfrazee:
This issue is also affecting the profile editor on android

AndreBordignon:
This bug is also affecting the register page, but it happened on my iPhone 11

estrattonbailey (approved):

PR #2362, opened 2023-12-28
Disable spanish translation until it's more thoroughly reviewed

pfrazee:
After what happened with German, we're slowing these down until we can get more thorough reviews

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -119,8 +119,9 @@ export function sanitizeAppLanguageSetting(appLanguage: string): AppLanguage {
       // DISABLED until this translation is fixed -prf
       // case 'de':
       //   return AppLanguage.de
-      case 'es':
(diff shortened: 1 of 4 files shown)


maisondasilva:
@pfrazee I saw that both translations had been checked internally, what happened? I ask this because I am sending the translation in Portuguese!

pfrazee:
@maisondasilva We have to get a better process for verifying translations. Unless we can be sure it's good, we have to wait to ship a translation.

maisondasilva:
@pfrazee I understand, why don't you use Crowdin for translations? Since the files there can be updated on Github through integration, I already translate several open projects here on GitHub with translation hosted in Crowdin, I believe this could be the best way to have control over the translations, as they can be approved and reviewed in a more user-friendly interface.

What do you think of the idea @ansh 

Example:

image

image

PR #2363, opened 2023-12-28
Handle birth dates as UTC, handle locale formatting

estrattonbailey:
Tested using a prod account on web/ios/android and editing the dates between them.


pfrazee (approved):
✅

PR #2367, opened 2023-12-29
Change paddingTop to marginTop for web emoji picker (EmojiPicker.web.tsx)

Juicysteak117:
I was going to open an issue, but I figured this would be a quick CSS fix I could do. Currently when opening the emoji picker on web, you can't close it in a significant amount of screen space due to the `padding` lacking pointer events. See attached screenshot. There's a `padding-right: 50px` as well that seems unavoidable regarding `marginHorizontal: auto` without something hacky to my hobby-level CSS knowledge in order to get the emoji picker to line up, but changing the `padding-top` to `margin-top` mostly resolves the issue. There are no secondary effects that I can tell due to the vertically fixed position of the emoji picker and post composer.

As a side note, it'd be nice if SHIFT+LMB did not close the emoji picker in order to allow for multiple emojis to be selected similar to how it works in Discord, but I don't know enough about the libraries to tweak `onEmojiSelect` or `onInsert` confidently. 

image

### DIFF
--- a/src/view/com/composer/text-input/web/EmojiPicker.web.tsx
+++ b/src/view/com/composer/text-input/web/EmojiPicker.web.tsx
@@ -50,8 +50,8 @@ export function EmojiPicker({close}: {close: () => void}) {
     textInputWebEmitter.emit('emoji-inserted', emoji)
     close()
   }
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Rad, works great, thanks you thank you!

PR #2373, opened 2023-12-29
Localize options in "Thread Preferences" screen

noritada:
This PR adds marks for localization to options in "Thread Preferences" screen.

Many thanks!


ansh (approved):
LGTM! Thanks

PR #2378, opened 2023-12-29
Remove splashscreen blink

zoontek:
Hi 👋

`react-native-bootsplash` author here. I noticed, when using your app, that the splash screen implem doesn't wait for the first render to hide the native splash. This result in a white flash, which is not great given the fact that it's our first experience with the app.

This PR fixes this (without switching lib, I only removed `react-native-splash-screen` as it's unused and autolinked / built for nothing)

**Before (notice the blink) 👇**

https://github.com/bluesky-social/social-app/assets/1902323/a2279214-fce7-46cf-8f8d-9816af8440b8

**After 👇**

https://github.com/bluesky-social/social-app/assets/1902323/5d38486d-5af9-40a2-b85d-3b87e982ccb6

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,6 @@
     "react-native-root-siblings": "^4.1.1",
     "react-native-safe-area-context": "4.7.4",
     "react-native-screens": "~3.27.0",
-    "react-native-splash-screen": "^3.3.0",
     "react-native-svg": "14.0.0",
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
Rad, thank you for this!

pfrazee:
Oh wonderful contribution, thank you!

PR #2379, opened 2023-12-30
fix web thumbnail resizing

haileyok:
Right now when resizing thumbnails on web, the image gets drawn to a canvas that starts the drawing at a y or x value based on the scale of the image. In some cases where images are landscape, this is problematic. See: https://bsky.app/profile/retr0.id/post/3khpgllcczr2j

We should always set the x and y axes to start drawing at (0,0). The height and width of the canvas should be set to the scaled height and width of the image.

Before: 
Screenshot 2023-12-29 at 3 49 17 PM

After: 
Screenshot 2023-12-29 at 3 58 22 PM

Note still though that the image is a bit too zoomed in. This was an error in the styles. Right now, there's a style applied to the image that sets the height to 200 regardless of platform, while the container is set to 200 on native and 120 on web. I fixed that by just applying `flex: 1` to the image so it conforms to the correct bounds on all platforms.

Now they resize correctly and appear correctly on both platforms.

Screenshot 2023-12-29 at 4 44 32 PM

### DIFF
--- a/src/lib/media/manip.web.ts
+++ b/src/lib/media/manip.web.ts
@@ -117,9 +117,6 @@ function createResizedImage(
         return reject(new Error('Failed to resize image'))
       }
 
-      canvas.width = width
-      canvas.height = height
-
       let scale = 1
       if (mode === 'cover') {
         scale = img.width < img.height ? width / img.width : height / img.height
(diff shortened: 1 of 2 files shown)


pfrazee (commented):
Holy moley, I had completely forgotten frontend was responsible for grabbing these images and was leaving this alone until backend had time to fix it. Another massive contribution.

A huge motivation for desktop giving the different link card than mobile was to mask this issue. I'm going to look at using the same card on desktop & mobile as part of this PR. If that works, I might tweak your changes a bit more and then we can shipit.jpeg

pfrazee:
Sweet. Folded into https://github.com/bluesky-social/social-app/pull/2398

Issue #1617, opened 2023-10-05
Gallery access to all images in local files is dependent on Android version


The-Peezoo:
**Describe the bug**
When you add an image to your post on a mobile device, you can add it from a folder in your local files. On Android devices with Android version below 12, you can access all images in all folders, including folders you create yourself, but on devices with Android 12 and above, you can access only the default Android folders Camera, Screenshots, and Downloads. 

**To Reproduce**

(Expected behavior with Android <12 and unexpected with Android >12).
1. Create a post by clicking the icon at the bottom right.
2.  Then click the left icon to add an image. 
3. Then click on 'Album'. 

**Expected behavior**

You should be able to access all images in all folders, including those you have created yourself.

**Details**

 - Platform: Android
 - Platform version: 9 (Works as expected),11 (Works as expected),12 (bug),13 (bug)
 - BlueSky App version: 1.51.0

h-fernand:
This appears to partially be an issue with Android itself. As someone experiencing this bug, the same issue occurs when trying to upload images to mobile sites like Facebook or Craigslist. However, on mobile websites I am given the option to tap three dots to expand the "browse" menu which lets me find the rest of my folders. On Bluesky this option does not exist. A custom media gallery for the app itself would be great, but in the interim giving us access to this "browse" window would absolutely be an acceptable solution.

aendra-rininsland:
I suspect it's the partial photo permissions update they've rolled out in Android, and given that this is using Expo, that hasn't updated and is running in [compatibility mode](https://developer.android.com/about/versions/14/changes/partial-photo-video-access#compatibility-mode) because it's [using the `READ_MEDIA_IMAGES` permission](https://github.com/expo/expo/blob/9643aea6268fbcf714c8d3fb99cea354aabd41cb/android/expoview/src/main/java/host/exp/exponent/utils/ScopedPermissionsRequester.kt#L140) instead of the new `READ_MEDIA_VISUAL_USER_SELECTED` permission.

I have no idea how to fix this without doing away with Expo's image picker altogether, but truly this bug makes posting images SUPER annoying.

pfrazee:
I haven't been able to find any information about this so far, @aendra-rininsland it sounds like we need to open an issue with expo

aendra-rininsland:
@pfrazee Might be this issue: https://github.com/expo/expo/issues/25361 or this PR https://github.com/expo/expo/pull/20907

Are you using the latest version of the image picker? It seems they fixed that back in January.

PR #2384, opened 2023-12-30
patch expo-image-picker to support "browse" on android

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/1617

Right now, expo-image-picker is using `PickVisualMedia`. This does not support the "Browse" option to select files elsewhere on the device except the gallery. Instead, we want to use `GetContent`. See https://android-developers.googleblog.com/2023/04/photo-picker-everywhere.html

This patches expo-image-picker to do so. Tested on SDK 34, 28, and ~~23~~  25. Tested accessing images from files as well as Google Drive with good success.

This is meant to be a temporary fix, and I didn't implement the logic for selecting videos. It's only a few extra lines, however we have to be careful that we properly create the intent (the mime can't be anything other than `image/*` or `video/*` otherwise it won't use the image picker).

We also have to use `Intent.EXTRA_ALLOW_MULTIPLE` which only accepts a boolean, not a limit. To that end, I've made a minor change in `picker.shared.ts` to slice the result and make sure we only return four images. Not the most ideal option, but I think the benefits of fixing this seemingly annoying issue outweigh that. To try and make it clear to the user, we present a toast if they select more than four images.

Of course whenever multiple is disabled we still only allow the selection of one file.

I think Google's intent is to make `PickVisualMedia` work in the future in the same way, so hopefully this isn't something that will need to be in the codebase for too long. They said by the end of the year though and...here we are 🙃

### DIFF
--- a/patches/expo-image-picker+14.7.1.patch
+++ b/patches/expo-image-picker+14.7.1.patch
@@ -1,0 +1,64 @@
+diff --git a/node_modules/expo-image-picker/android/src/main/java/expo/modules/imagepicker/contracts/ImageLibraryContract.kt b/node_modules/expo-image-picker/android/src/main/java/expo/modules/imagepicker/contracts/ImageLibraryContract.kt
+index ff15c91..bc868d1 100644
+--- a/node_modules/expo-image-picker/android/src/main/java/expo/modules/imagepicker/contracts/ImageLibraryContract.kt
++++ b/node_modules/expo-image-picker/android/src/main/java/expo/modules/imagepicker/contracts/ImageLibraryContract.kt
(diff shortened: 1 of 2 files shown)


pfrazee:
@haileyok going to review this asap but dropping in real quick to say if you're interested in a job, I definitely want to chat (paul@blueskyweb.xyz)

pfrazee (approved):
I can't tell you how much I appreciate this one. It's been on the todo list for so long.

I'm cleaning up the patchfile with 95%ish confidence that I'm just removing build intermediates. If I get that wrong, we've got this PR to look at again in the future.

Issue #2390, opened 2023-12-30
Keep web emoji picker open with a shift click


Juicysteak117:
**Is your feature request related to a problem? Please describe.**

I mentioned this in #2367 as a side note, but I figured it would be better to break this out into a feature request. For the emoji picker, it would be nice if `SHIFT+LMB` did not close the emoji picker in order to allow for multiple emojis to be selected similar to how it works in Discord, but I don't know enough about the libraries to tweak `onEmojiSelect` or `onInsert` confidently myself.

**Describe the solution you'd like**

I would like for multiple emojis to be selected without having to repeatedly open the emoji picker poput.

**Describe alternatives you've considered**

The alternative is tedious.

**Additional context**

mschwendener:
Planned: https://github.com/bluesky-social/social-app/issues/1425

Issue #1425, opened 2023-09-10
[Web] Emoji picker should not close automatically after picking an emoji


n0vedad:
I'm not really sure if this is a bug or work as designed, but I really would like to see an emoji picker which doesn't close automatically after picking an emoji and I have to reopen it if I want to add another emoji.

**Describe the solution you'd like**

See above.

**Additional context**

This afflicts only the web client. I'm usig Windows 11 and Firefox 117.0

mschwendener:
Related https://github.com/bluesky-social/social-app/issues/1423

ansh:
@mschwendener Is describing a bug with the search bar in the emoji picker. This issue is about a feature request of allowing multiple emojis to be inserted.

zarakk:
Hi @ansh, is this feature still in discussions? I've added this feature locally. If this is the expected behavior I'll submit a PR which would resolve this.  

https://github.com/bluesky-social/social-app/assets/52151884/707b7898-73fd-4a55-8e24-db2bb56eede8

PR #2391, opened 2023-12-30
keep emoji picker open when shift is pressed while selecting emoji

haileyok:
closes https://github.com/bluesky-social/social-app/issues/2390
closes https://github.com/bluesky-social/social-app/issues/1425

simple stuff. when the emoji picker is open, listen for shift down/up. store the value in a ref, and don't close the picker on select if the ref is true. remove listeners on close.


ansh:
looking forward to your next PR

Issue #2389, opened 2023-12-30
Web emoji picker misplaced for extended post composer fields


Juicysteak117:
**Describe the bug**
I noticed this while writing #2388 as a related issue to #2372, but the emoji picker having a fixed position is a problem in the case of an extremely long post composer field as shown in the first screenshot.

**To Reproduce**

1. Open the post composer
2. Insert a large number of new lines
3. Open the emoji picker
4. Observe

**Expected behavior**

The picker to stay lined up with the button, I assume? Or perhaps it would be better to do it similar to the native Windows emoji picker which opens the popout beneath the cursor (see attached for example).

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/20595808/413d0b71-b087-459c-8f3f-c0246a4a1b96)

![image](https://github.com/bluesky-social/social-app/assets/20595808/e8ec7a63-80eb-4082-9700-fae94dc81bb3)

**Details**

 - Platform: Desktop chromium
 - Platform version:
 - App version:

**Additional context**

PR #2392, opened 2023-12-31
emoji picker improvements

haileyok:
Fixes: https://github.com/bluesky-social/social-app/issues/2390
Fixes: https://github.com/bluesky-social/social-app/issues/2389
Fixes: https://github.com/bluesky-social/social-app/issues/1425

Demo: https://www.youtube.com/watch?v=eGJY9y7i4BY&feature=youtu.be

## Problem

Right now the picker is being presented as a button dropdown. This limits the "mobility" of the picker, as we cannot control where it's going to be displayed at on the screen.

Screenshot 2023-12-30 at 6 41 21 PM

This creates some issues like what we see here: https://github.com/bluesky-social/social-app/issues/2389

Instead, we want the picker to be movable about the entirety of the screen so that we can adjust to the position of the cursor.

## How

First, we need to move the picker outside of the composer modal so that we can set the position to anywhere on the screen. We do so by placing it inside of `Composer.web.tsx`.

To control the picker, we have a small state object with `isOpen` and `top`. The top value is the y position of the text input cursor plus 50 pixels, so that it will always be visible no matter where the picker is. This also gives us room to make further adjustments such as moving it to the left of the text on smaller screens where the entire picker can't show up. I.e., right now it gets a bit messy:

Screenshot 2023-12-30 at 6 45 38 PM

To get the current cursor, we can use tiptap's editor state to do this. We add `getCursorPosition()` to `TextInput`'s methods:

```tsx
    getCursorPosition: () => {
      const pos = editor?.state.selection.$anchor.pos
      return pos ? editor?.view.coordsAtPos(pos) : undefined
    }
```

Now when we press the emoji button, we just call `openPicker(textInput.current?.getCursorPosition())`.

A little bit of logic can tell us where we need to place the picker:

```tsx
  const position = React.useMemo(() => {
    const fitsBelow = state.pos.top + PICKER_HEIGHT < height
    const fitsAbove = PICKER_HEIGHT < state.pos.top
    const placeOnLeft = PICKER_WIDTH < state.pos.left
    const screenYMiddle = height / 2 - PICKER_HEIGHT / 2

    if (fitsBelow) {
      return {
        top: state.pos.top + HEIGHT_OFFSET,
      }
    } else if (fitsAbove) {
      return {
        bottom: height - state.pos.bottom + HEIGHT_OFFSET,
      }
    } else {
      return {
        top: screenYMiddle,
        left: placeOnLeft ? state.pos.left - PICKER_WIDTH : undefined,
        right: !placeOnLeft
          ? width - state.pos.right - PICKER_WIDTH
          : undefined,
      }
    }
  }, [state.pos, height, width])
```

Examples: 

Space below:
Screenshot 2023-12-30 at 7 37 10 PM

Space above:
Screenshot 2023-12-30 at 7 37 29 PM

Space to the left (always prefer the left if there is space):
Screenshot 2023-12-30 at 7 37 43 PM

No space on the left so we move to the right:
Screenshot 2023-12-30 at 7 38 10 PM

## Don't close while pressing shift

Simple event listener for shift up/down. Don't close the picker when pressing shift so we can select multiple emojis.

### DIFF
--- a/src/state/shell/composer.tsx
+++ b/src/state/shell/composer.tsx
@@ -30,6 +30,7 @@ export interface ComposerOpts {
   onPost?: () => void
   quote?: ComposerOptsQuote
   mention?: string // handle of user to mention
+  openPicker?: (pos: DOMRect | undefined) => void
 }
 
 type StateContext = ComposerOpts | undefined
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -6,6 +6,7 @@ import {
   Keyboard,
   KeyboardAvoidingView,
   Platform,
+  Pressable,
   ScrollView,
   StyleSheet,
   TouchableOpacity,
@@ -46,7 +47,6 @@ import {Gallery} from './photos/Gallery'
 import {MAX_GRAPHEME_LENGTH} from 'lib/constants'
 import {LabelsBtn} from './labels/LabelsBtn'
 import {SelectLangBtn} from './select-language/SelectLangBtn'
-import {EmojiPickerButton} from './text-input/web/EmojiPicker.web'
 import {insertMentionAt} from 'lib/strings/mention-manip'
 import {Trans, msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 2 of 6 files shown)


pfrazee (approved):
Incredible work @haileyok. Repeating to make sure it reaches you - if you're at all interested in a job, please email me at paul@blueskyweb.xyz.

Juicysteak117:
Bless you @haileyok for fixing all of the minor grievances I have had with with the emoji picker lately 🙏

PR #2393, opened 2023-12-31
Improve French localization

Signez:
This pull request tries to improve the French localization effort that was already done earlier this month in #2265 (quite well, actually! Congratulations, coming from a native speaker who is nitpicky about translations in software products! 💛).

- Add missing translations (I just run `yarn intl:extract` and translated the empty ones)
- Always use French guillemets («») when applicable — it was already done so congrats for that, but some English ones still slipped through
- Always add an article (le/la/l’) before nouns — those should never be omitted in my opinion, that is never done in other software products unless there are very hard space contraints
- Use … character instead of three dots
- Change _"Mettre en sourdine"_ into _"Masquer"_, used in other similar products to translate _"Mute"_
- Change other usages of _"Masquer"_ into its synonym _"Cacher"_ to prevent confusion between concepts
- ~~Shorten _"Nouvelle publication"_ into _"Nouv. publication"_ to prevent it overflowing in the main blue button~~
- With the greenlight of @ansh, renamed all occurrences of _Publication_ into _Post_, that is less mouthful and takes less space in critical UI elements (i.e. the blue button on the left)

It's my first contribution to Bluesky, feel free to tell me if I did something that could have been better for my future contributions :) (I may submit improvements to i18n too, there are some strings that are hard to fix as-is for now because of some string concatenations, so I may suggest some improvements on that later if you are open for those.)

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -1433,7 +1433,7 @@ msgstr "Seul {0} peut répondre."
 #: src/view/com/composer/Composer.tsx:468
 #: src/view/com/composer/Composer.tsx:469
 msgid "Open emoji picker"
-msgstr ""
+msgstr "Ouvrir le sélecteur d’emoji"
 
 #: src/view/com/pager/FeedsTabBarMobile.tsx:76
 msgid "Open navigation"
@@ -1764,7 +1764,7 @@ msgstr "Demande de modification"
 
 #: src/view/screens/Settings.tsx:422
 msgid "Require alt text before posting"
-msgstr ""
+msgstr "Nécessiter un texte alt avant de publier"
 
 #: src/view/com/auth/create/Step2.tsx:68
 msgid "Required for this provider"
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -68,7 +68,7 @@ msgstr "Une nouvelle version de l’application est disponible. Veuillez faire l
 msgid "Accessibility"
 msgstr "Accessibilité"
 
-#: src/view/com/auth/login/LoginForm.tsx:159
+#: src/view/com/auth/login/LoginForm.tsx:163
 #: src/view/screens/Settings.tsx:286
 msgid "Account"
 msgstr "Compte"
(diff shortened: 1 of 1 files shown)


ansh:
Really appreciate your feedback @Signez ! Till we figure out a formal process for reviewing translations, right now we are just waiting until there is a 2+ native speaker consensus on changes before merging. So, I'll wait till anyone else has anything to chime in here.

ansh:
Would also love your feedback here #2404 @Signez 

We got some professional review on our translation file (I would've sent your changes to be reviewed too but this was last week). Would appreciate your feedback on that PR if you have time.

src/locale/locales/fr/messages.po
```diff
@@ -1334,7 +1338,11 @@ msgstr "Nouvelle publication"
 
 #: src/view/shell/desktop/LeftNav.tsx:258
 msgid "New Post"
-msgstr "Nouvelle publication"
+msgstr "Nouv. publication"
```
ansh, line 1341: Do you think this could be translated as "Nouvelle Poste?"

For example, other sites use the french word "Poster" for "Post"

Signez: Indeed, this could be translated to `Nouveau post`! (Yeah, `post` is _masculine_ in French, whereas `publication` is feminine, so it would be `Nouveau` and not `Nouvelle` in that case.)

I didn't want to change `publication` to `post` in my pull request because I didn't want, as a first contribution, change all mentions of `publications` into `post` — but I think it would be a good call, as users are used to `post` to refer to it, both in writing and in speaking.

If you don't mind this change, I can make the move from `publication` to `post` in that pull request. Waiting for your green light to do so :)

Signez:
> Would also love your feedback here #2404 @Signez
> 
> We got some professional review on our translation file (I would've sent your changes to be reviewed too but this was last week). Would appreciate your feedback on that PR if you have time.

I've just done it! Note that even though I'm not a professional translator, I used to work on a product that spanned several languages ( on a pan-European website selling train tickets), and I helped localize that product, so I'm quite familiar with this kind of endeavor, and the challenges you face. 🤗

src/locale/locales/fr/messages.po
ansh: Yeah, that would be great! 

Again, want to thank you a lot for your help here. Please add yourself to the `"Last-Translator: \n"
"Language-Team: \n"` header in the `messages.po` file if you haven't done so already!

Signez:
This pull request is now rebased, updated with the two new strings extracted from the codebase, and with `publication` switched with `post` everywhere (30 substitutions + _accords_ redone, wrt moving from a _feminine_ noun to a _masculine_ one).

ansh (approved):
Merging so it does not go stale

PR #2399, opened 2024-01-02
Do not nuke session on unknown backend errors

estrattonbailey:
Relies on https://github.com/bluesky-social/atproto/pull/2010

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -103,9 +103,9 @@ function createPersistSessionHandler(
     refreshedAccount: SessionAccount
   }) => void,
   {
-    unstable_networkErrorCallback,
+    networkErrorCallback,
   }: {
-    unstable_networkErrorCallback?: () => void
+    networkErrorCallback?: () => void
   } = {},
 ): AtpPersistSessionHandler {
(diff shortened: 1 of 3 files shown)


src/state/session/index.tsx
```diff
@@ -102,10 +102,21 @@ function createPersistSessionHandler(
...
+  } = {},
 ): AtpPersistSessionHandler {
   return function persistSession(event, session) {
     const expired = event === 'expired' || event === 'create-failed'
 
+    if (event === 'network-error') {
+      logger.warn(`session: persistSessionHandler received network-error event`)
+      unstable_networkErrorCallback?.()
+      return
+    }
```
estrattonbailey, line 118: Only real addition is this block, just kicks the user out to signup and sends a warning to sentry

pfrazee: Why in a sub-object with the unstable_ prefix?

estrattonbailey: Oh, just removed, was just what I called it initially

pfrazee (approved):
Was able to successfully test on web by blocking `getSession`. Good step in the right direction, since this makes it so you don't need to input your password.

PR #2400, opened 2024-01-03
Remove masking from splash screen

estrattonbailey:
We've been having some trouble with the masked view. Though we may be able to continue to fix forward, a more simple fade-out looks and feels almost exactly the same and hopefully avoids the pitfalls we've hit previously.


pfrazee:
Replaced by other PRs

PR #2402, opened 2024-01-03
Add some more logging

estrattonbailey:


src/state/session/index.tsx
```diff
@@ -181,14 +181,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
-        {
-          service,
-          handle,
-        },
-        logger.DebugContext.session,
-      )
+      logger.info(`session: creating account`, {
+        service,
+        handle,
+      })
```
estrattonbailey, line 187: Just makes sure this gets sent to Sentry prior to a thrown exception (handled in create account state reducer)

src/view/com/auth/create/state.ts
```diff
@@ -136,7 +136,13 @@ export async function submit({
         msg`Invite code not accepted. Check that you input it correctly and try again.`,
       )
     }
-    logger.error('Failed to create account', {error: e})
+
+    if ([400, 429].includes(e.status)) {
```
estrattonbailey, line 140: Kinda basic, but should give us a place to go from here

src/view/com/auth/login/LoginForm.tsx
```diff
@@ -107,17 +107,21 @@ export const LoginForm = ({
...
         setError(_(msg`Invalid username or password`))
       } else if (isNetworkError(e)) {
+        logger.warn('Failed to login due to network error', {error: errMsg})
         setError(
           _(
             msg`Unable to contact your service. Please check your Internet connection.`,
           ),
         )
       } else {
+        logger.warn('Failed to login', {error: errMsg})
```
estrattonbailey, line 124: This is the existing event, so unknown errors will continue to accrue here. Numbers should go down.

src/view/com/auth/create/state.ts
```diff
@@ -136,7 +136,13 @@ export async function submit({
...
+      logger.warn('Failed to create account', {error: e})
```
estrattonbailey, line 141: Most of these are just validation errors. We should look out for spikes though.

pfrazee (approved):
Yeah good stuff

PR #2404, opened 2024-01-03
Improve French Localization

ansh:
Improve French localization based on feedback

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -49,7 +49,7 @@ msgstr "<0/> membres"
 
 #: src/view/com/auth/onboarding/RecommendedFeeds.tsx:30
 msgid "<0>Choose your</0><1>Recommended</1><2>Feeds</2>"
-msgstr "<0>Choisissez vos</0><1>fils d’actualité</1><2>recommandés</2>"
+msgstr "<0>Choisissez vos</0><1>recommandés</1><2>fils</2>"
 
 #: src/view/com/auth/onboarding/RecommendedFollows.tsx:37
 msgid "<0>Follow some</0><1>Recommended</1><2>Users</2>"
@@ -57,7 +57,7 @@ msgstr "<0>Suivre certains</0><1>utilisateurs</1><2>recommandés</2>"
 
 #: src/view/com/util/moderation/LabelInfo.tsx:45
 msgid "A content warning has been applied to this {0}."
-msgstr ""
+msgstr "Un avertissement sur le contenu a été appliqué à ce {0}."
 
 #: src/lib/hooks/useOTAUpdate.ts:16
 msgid "A new version of the app is available. Please update to continue using the app."
@@ -68,7 +68,7 @@ msgstr "Une nouvelle version de l’application est disponible. Veuillez faire l
 msgid "Accessibility"
 msgstr "Accessibilité"
 
-#: src/view/com/auth/login/LoginForm.tsx:159
+#: src/view/com/auth/login/LoginForm.tsx:163
 #: src/view/screens/Settings.tsx:286
 msgid "Account"
 msgstr "Compte"
@@ -98,6 +98,7 @@ msgstr "Ajouter un compte"
 
 #: src/view/com/composer/photos/Gallery.tsx:119
 #: src/view/com/composer/photos/Gallery.tsx:180
+#: src/view/com/modals/AltImage.tsx:93
 msgid "Add alt text"
 msgstr "Ajouter un texte alt"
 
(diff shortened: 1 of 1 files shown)


src/locale/locales/fr/messages.po
```diff
@@ -49,15 +49,15 @@ msgstr "<0/> membres"
 
 #: src/view/com/auth/onboarding/RecommendedFeeds.tsx:30
 msgid "<0>Choose your</0><1>Recommended</1><2>Feeds</2>"
-msgstr "<0>Choisissez vos</0><1>fils d’actualité</1><2>recommandés</2>"
+msgstr "<0>Choisissez vos</0><1>recommandés</1><2>fils</2>"
```
Signez, line 52: This sounds weird. We never use _recommandés fils_, always _fils recommandés_ in French. The `recommandé(s)` adjective is never written _before_ the noun in French, always _after_.

src/locale/locales/fr/messages.po
```diff
@@ -186,13 +187,12 @@ msgstr "Mots de passe de l’appli"
 
 #: src/view/com/util/forms/PostDropdownBtn.tsx:236
 msgid "Appeal content warning"
-msgstr ""
+msgstr "Avertissement sur le contenu de l'appel"
 
 #: src/view/com/modals/AppealLabel.tsx:65
 msgid "Appeal Content Warning"
-msgstr ""
+msgstr "Avertissement sur le Contenu de L'appel"
```
Signez, line 194: This translation is OK, but let's note that we don't do Title Casing in UI in software localization in French. I would recommand keeping it like `Avertissement sur le contenu de l'appel`, without special casing applied (beside the first letter, of course).

src/locale/locales/fr/messages.po
```diff
@@ -1532,18 +1547,16 @@ msgstr "Veuillez également entrer votre mot de passe :"
...
-#: src/view/com/modals/AppealLabel.tsx:72
-#: src/view/com/modals/AppealLabel.tsx:75
 #~ msgid "Please tell us why you think this decision was incorrect."
 #~ msgstr "Dites-nous pourquoi vous pensez que cette décision était incorrecte."
 
-#: src/view/com/composer/Composer.tsx:214
+#: src/view/com/composer/Composer.tsx:215
 msgid "Please wait for your link card to finish loading"
-msgstr ""
+msgstr "Attendez la fin du chargement de votre carte de liaison"
```
Signez, line 1557: `carte de liaison` doesn't mean what you think it means (it sounds like an administrative physical card you handle to someone?!). Either use `carte de lien`, or a more liberal translation, like `aperçu de lien` (literally `link preview`).

src/locale/locales/fr/messages.po
```diff
@@ -2171,14 +2186,14 @@ msgstr "Cet avertissement n’est disponible que pour les messages contenant des
 
 #: src/view/com/util/forms/PostDropdownBtn.tsx:178
 msgid "This will hide this post from your feeds."
-msgstr ""
+msgstr "Cette opération permet de cacher cet article dans vos fils d'actualité."
```
Signez, line 2189:
```suggestion
msgstr "Ceci cachera cet article dans vos fils d'actualité."
```

(Literally moving from `This operation enable hiding…` to `This will hide…`. I don't understand why this translation diverge that much from the original English text, I don't subscribe to that vision.)

src/locale/locales/fr/messages.po
```diff
@@ -2350,13 +2365,9 @@ msgstr "Bienvenue sur <0>Bluesky</0>"
...
 msgstr "Quel est le problème avec cette {collectionName} ?"
 
-#: src/view/com/auth/SplashScreen.tsx:34
-#~ msgid "What's next?"
-#~ msgstr "Et après ?"
-
 #: src/view/com/auth/SplashScreen.tsx:34
 msgid "What's up?"
-msgstr ""
+msgstr "Qu'est-ce qui se passe ?"
```
Signez, line 2370: This translates to `What is happening?`. If you want something that is really translating to `What's up?`, I would suggest translating it to `Quoi de neuf ?` instead (literally `What's new?`, but used in the same place as English's `What's up?`.).

Signez (commented):
Here are my review; if I had the rights (I don't of course 😅), I would have marked it as "Request changes", as my first comment about `recommandés` makes it really unnatural in my humble, native speaker opinion.

src/locale/locales/fr/messages.po
ansh: That is great. Thanks, made the change.

src/locale/locales/fr/messages.po
ansh: Done

Signez:
Somewhat superseeded by #2404, but it's still an improvement so LGTM 👍

estrattonbailey (approved):

PR #2405, opened 2024-01-03
Enable Spanish Localization

ansh:
We put our Spanish localization through another 3rd party review and now we are confident to enable it


pfrazee (approved):
Excellent, thanks!

PR #2415, opened 2024-01-03
Adding catalan translation

ivanbea:
Here is the catalan translation of messages.po and all the required files.


ivanbea:
closing it as I will make a new PR with just the catalan messages.po

PR #2419, opened 2024-01-03
add expandable context to composer when replying to post

haileyok:
Small addition for something that really bothers me. When replying to a larger post, we lose the ability to see the full context of the post. Sometimes I find myself needing to copy my reply and discard it real quick to see the full post...

This just makes the reply to area a pressable and adds a little `LayoutAnimation` to the expand/collapse so it looks pretty. Also displays images in a square similar to how Twitter does when applicable, as well as shows the text of the quoted post.

https://github.com/bluesky-social/social-app/assets/153161762/961c163d-d8f5-48cc-b3f8-a26c0b3ebccc

https://github.com/bluesky-social/social-app/assets/153161762/3ca6559e-b3bf-4e3e-9d8c-072920a7538b

### DIFF
--- a/src/state/shell/composer.tsx
+++ b/src/state/shell/composer.tsx
@@ -11,6 +11,7 @@ export interface ComposerOptsPostRef {
     displayName?: string
     avatar?: string
   }
+  embed?: AppBskyEmbedRecord.ViewRecord['embed']
 }
 export interface ComposerOptsQuote {
   uri: string
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -5,7 +5,9 @@ import {
   BackHandler,
   Keyboard,
   KeyboardAvoidingView,
+  LayoutAnimation,
   Platform,
+  Pressable,
   ScrollView,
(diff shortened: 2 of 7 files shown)


mozzius:
Might be neat to also include embeds in the expanded view (I do this for graysky)

haileyok:
@mozzius I thought about that. Do you think it's irritating at all (often would end up taking up most of the screen and hiding the reply you're writing)? Not hard to do though if that's preferred.

mozzius:
Definitely can be irritating for tall images; I like how Twitter uses the tiny square image next to the text. But I think surfacing quotes is very helpful

image

image

haileyok:
Ah yea quote are definitely nice since there could certainly be some context there. Images are great too and I like how Twitter does that, but what about if there are multiple images? Just the first one? (I guess I could download Twitter again 🙄)

haileyok:
@mozzius Like this?

mozzius:
Perfect! Looks great!

pfrazee:
Fire as ever

pfrazee (approved):

PR #2420, opened 2024-01-04
Reduce web requests

pfrazee:
We're still trying to track down the excessive web traffic. This PR makes two changes:

- 4e7c886fb2c21497b48e00325ae5bc0a97b0796d Stop auto-refetching app passwords and invites on an interval
- bd62ce7668b1ea57654a5dede4990ef738dabcae Stop polling posts and notifications when the tab is backgrounded

The second commit is likely to reduce traffic dramatically on web, where people often have multiple tabs open


ansh (approved):
Makes sense to me.

PR #2423, opened 2024-01-04
Update Japanese localization (addition + language code correction)

noritada:
This PR adds translations for newly introduced messages, and also corrects language code (ISO 639-1).

Dear @Hima-Zinn @tkusano @dolciss @oboenikui
I would appreciate your review and feedback on these text updates.

Many thanks!

### DIFF
--- a/src/locale/locales/ja/messages.po
+++ b/src/locale/locales/ja/messages.po
@@ -5,7 +5,7 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: @lingui/cli\n"
-"Language: jp\n"
+"Language: ja\n"
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
 "PO-Revision-Date: \n"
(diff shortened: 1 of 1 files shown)


src/locale/locales/ja/messages.po
```diff
@@ -1445,7 +1445,7 @@ msgstr "{0}のみ返信可能"
 #: src/view/com/composer/Composer.tsx:468
 #: src/view/com/composer/Composer.tsx:469
 msgid "Open emoji picker"
-msgstr ""
+msgstr "絵文字ピッカーを開く"
```
oboenikui, line 1448: 「絵文字ピッカー」はITに詳しくないと伝わらない用語という気がするので、「絵文字を入力」などの方が一般向けとして相応しいように感じました。

I feel that "絵文字ピッカー" is a term that only those who are familiar with IT would understand, so I think that something like "絵文字を入力" would be more appropriate for the general public.

oboenikui (commented):
I commented on one point of concern. We would appreciate it if you would consider changing it.

src/locale/locales/ja/messages.po
ansh: Has this change been made?

noritada:
@ansh I made a change suggested by @oboenikui.
Thanks a lot!

ansh (approved):

PR #2427, opened 2024-01-04
Include expanded urls in social cards

zachlipton:
#1188 had the side effect that post social cards have truncated links, which means that bsky posts displayed in Slack, among other places, display non-functional URLs. This is increasingly a problem with broader sharing of publicly viewable posts from the web interface. For example:

This PR undoes the effect of shortenLinks() when formatting social cards, expanding these truncated urls back to their original content from the post's facets. Additionally, posts with embedded links get the url of the embedded link appended to the end of the social card if it's not already included in the post so those posts don't appear without their links.

Full disclosure: I don't know much Go and this is my first bluesky-social PR. I've tested this pretty extensively locally, including with multibyte emoji, posts with multiple links, and a wide variety of posts on the live site, but someone who knows the language and the codebase should please let me know if I've missed anything non-obvious or made a mess of it.

**Before**:

**After**:

### DIFF
--- a/bskyweb/cmd/bskyweb/server.go
+++ b/bskyweb/cmd/bskyweb/server.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"slices"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


zachlipton:
> how does it handle link facets with mismatching text/uri? https://bsky.app/profile/techmeme.com/post/3kbzrhz3v3p2p

I, uh, did not realize that was a thing, oops. Converting this PR to a draft until that's sorted. 

I could either put the uri after the facet text (perhaps in parens) or just not expand those links at all (they currently aren't expanding)?

zachlipton:
Ok I modified it to leave facets with mismatching text/uri alone (i.e. it will extend links that were shortened by the composer and include missing uris from embeds, but will leave the Techmeme example alone)

Testcases (I'd write tests but the go server doesn't appear to have a test suite yet):

- https://bsky.app/profile/techmeme.com/post/3kbzrhz3v3p2p: 

> `<meta property="og:description" content="YouTube adds a news page on its mobile apps to suggest content from &quot;authoritative sources&quot;, and plans to spend $1.6M to promote the creation of news on Shorts (Jon Porter/The Verge)Main Link | Techmeme Permalink">` (unchanged from production)

- https://bsky.app/profile/zachlipton.com/post/3ki7evh767y25 (multibyte unicode with both link embed and link facet): 

> ```
> <meta property="og:description" content="test 😀œ´®†¥¨ˆøπ¬˚∆˙©ƒ∂ßåΩ≈ç√∫˜µ≤ユーザーコードa
> 
> https://defector.com/the-messenger-is-circling-the-drain
> 
> test
> https://www.politico.eu/article/hydroxychloroquine-could-have-caused-17000-deaths-during-covid-study-finds/">
> ```

- https://bsky.app/profile/morgfair.bsky.social/post/3ki7etjvqmi2e (only expand embeds once): 
> `<meta property="og:description" content="https://apple.news/A_Qzd_O-fT5KFmpaSQfaaiA">`

pfrazee:
@bnewbold Sorry to do this but I gotta bounce this one to you 🙏

bskyweb/templates/post.html
```diff
@@ -22,8 +22,8 @@
   <meta property="og:title" content="@{{ postView.Author.Handle }}">
   {% endif -%}
   {%- if postView.Record.Val.Text %}
```
bnewbold, line 24: this check should become `if postText`

bskyweb/cmd/bskyweb/server.go
```diff
@@ -342,6 +343,42 @@ func (srv *Server) WebPost(c echo.Context) error {
 		}
 		data["imgThumbUrls"] = thumbUrls
 	}
+
+	// expand links in rich text back to full urls, replacing shortened urls in
+	// social card meta tags and the noscript output. this essentially reverses the effect
+	// of shortenLinks() in src/lib/strings/rich-text-manip.ts
+	if postView.Record != nil {
```
bnewbold, line 350: this section should get refactored out in to a separate function; that will make testing easier. it should probably return a string, and maybe an error as well.

bskyweb/cmd/bskyweb/server.go
```diff
@@ -342,6 +343,42 @@ func (srv *Server) WebPost(c echo.Context) error {
...
+		postText := rec.Text
+		var charsAdded int64 = 0
+		// iterate over facets, check if they're link facets, and if found, grab the uri
+		for _, facet := range rec.Facets {
+			linkUri := ""
+			if slices.ContainsFunc(facet.Features, func(feat *appbsky.RichtextFacet_Features_Elem) bool {
+				if feat.RichtextFacet_Link != nil && feat.RichtextFacet_Link.LexiconTypeID == "app.bsky.richtext.facet#link" {
+					linkUri = feat.RichtextFacet_Link.Uri
+					// only expand uris that have been shortened (as opposed to those with mismatched text/uris)
+					if strings.HasSuffix(postText[facet.Index.ByteStart+charsAdded:facet.Index.ByteEnd+charsAdded], "...") &&
```
bnewbold, line 361: String indexing like this should work in terms of correct offsets: the facet byte ranges are UTF-8 byte offsets, and that is how string indexing works in golang.

However, there needs to be some bounds checking here, to prevent out-of-bounds index panics.

bnewbold (commented):
I left some review comments. If they are outside your golang comfort zone, I can take a pass myself when I get a moment.

It is kind of a bummer that opengraph doesn't include a separate tag for "external link" or similar. I don't love just appending the link at the end, but I don't have an alternative to propose for now.

For completeness, we should probably also handle quote-posts ("embed record") and the "embed record and media" cases soon. Don't need to handle all those in this one PR though.

pfrazee:
Thanks zach! I'll let @bnewbold review again

bskyweb/cmd/bskyweb/server.go
bnewbold: as a style nit, the code in this block would be less-indented if the check was:

```go
if postView.Record == nil {
    return ""
}
```

bnewbold (approved):
this looks good, thanks for responding to the review feedback!

I'm going to make 1-2 other small changes in a follow-up, but good to merge

bnewbold:
Lint failure is unrelated. I think the right move for attribution is to squash-merge this, then i'll rebase and squash-merge my other fixes on top of this.

Issue #2430, opened 2024-01-05
User Mentions Follow-Prioritization Lowercase Only


Juicysteak117:
**Describe the bug**
The popout list when attempting to `@` a user breaks the prioritization of followed users when using any caps. When attempting to use caps (and even matching case), the list appears to prune ANY followed users.

I am not 100% that `actor-autocomplete.ts` is the right file for this, so no PR, but `if (info.displayName?.toLocaleLowerCase().includes(prefix))` under the `prefixMatch` function seems to be the relevant offending piece of code assuming I am reading it correctly.

**To Reproduce**

1. Attempt to tag a followed user
2. Use caps
3. Observe

**Expected behavior**

The list to still prioritize followed users.

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/20595808/15075624-32ad-4c8a-a8e1-60ba3f96a916)
![image](https://github.com/bluesky-social/social-app/assets/20595808/1f33e951-02e4-4ef3-be46-337dd898927b)

In fact, ANY capitalization breaks it.
![image](https://github.com/bluesky-social/social-app/assets/20595808/6769b82b-1def-4947-90d2-5048e08264bf)
![image](https://github.com/bluesky-social/social-app/assets/20595808/d52fd026-0bc6-425f-a04f-05e7ed9f25e1)

**Details**

 - Platform: web chromium
 - Platform version:
 - App version:

**Additional context**

Juicysteak117:
@haileyok No imposter, that's just me keeping the placeholder account to make sure nobody ELSE is an impostor haha. Thank you!

Juicysteak117:
@pfrazee Can I get a double check on if this change is actually reflected in production since I'm still running into this issue regularly where caps in a search causes the results to break.

haileyok:
@Juicysteak117 Thanks, curious about this one. Wonder if it's a v2 change since I think this worked before?

Juicysteak117:
@haileyok I'm unsure as I don't know too much about the v2 stuff but at no point did I see this working in production. I only didn't bump it sooner since I figured it might've been delayed by an update or two for whatever reason.

PR #2431, opened 2024-01-05
convert prefix to lowercase in actor autocomplete query

haileyok:
Case shouldn't matter when making autocomplete suggestions, and followed users should be prioritized but without converting the prefix to lowercase, they are not.

Here, convert the prefix to lowercase in the respective query. 

Before:

After: 

fixes https://github.com/bluesky-social/social-app/issues/2430


pfrazee (approved):
Yep yep!

PR #2436, opened 2024-01-05
Mark more texts for localization

noritada:
This PR adds marks for localization to unlocalized texts left on several screens.

Many thanks!


dolciss:
It seems to me that this change is probably all included in #2210

noritada:
Thank you for letting me know that great PR.  I am sorry I did not notice it.
I checked the PR and it looks like a very large piece. I also understood that the review and discussion is ongoing.

I have tweaked this PR and removed where there is controversy so that this PR is a subset of the previous PR.
If it would take too long to merge that monolithic PR at once, I would appreciate it if you would merge this PR.
If you think you can merge that PRs, please feel free to ignore this PR as it will be time-consuming.

Many thanks.

PR #2437, opened 2024-01-05
Use android mode, improve fallback

estrattonbailey:
Makes sure mask animates on Android, and improves fallback for <v25


ansh (approved):
Tested and looks great!

haileyok:
@estrattonbailey hmm, using android mode breaks images 😓 

Screenshot 2024-01-06 at 9 22 45 PM

estrattonbailey:
Oh geez, will look into that!

PR #2439, opened 2024-01-06
Update Ukrainian translations

imax9000:


Andygol:
@imax9000 could you consider sync translation of bsky.app with 🇺🇦 translation of [decl.blue](https://deck.blue) (https://github.com/deckblue/l10n/tree/main)?

imax9000:
> @imax9000 could you consider sync translation of bsky.app with 🇺🇦 translation of [deck.blue](https://deck.blue) (https://github.com/deckblue/l10n/tree/main)?

I don't think I'll have enough bandwidth to do it myself, but I agree that having some consistency would be good

ansh (approved):

PR #2440, opened 2024-01-06
Internationalize more strings (i.e. add _, msg and Trans everywhere)

Signez:
So... you know that internationalization effort you made two months ago? It was great. But... it was incomplete. 😅

But don't worry! I've just completed(-ish) the internationalization work for you. It took me four hours and I lost a few brain cells in the process, but you know, it's a way of thanking all of you, the Bluesky team, for taking the first step.

Let us translate the rest of Bluesky 💪 

---

So, what was missing?

- All toasts. They should be translated: they are visible to users, and are important feedback for their actions.
- Nearly all the `accessibilityHint`s were untagged. It doesn't make sense: those hints are read by humans, so they should be translated.
- Nearly all the `placeholder`s. Those are really visible by users, so it was clearly missing from the first work. Let's translate those words!
- Page titles. They appear in tabs, so they are quite visible too!
- …and some markup, too, that were missed in the previous work. With all the new `<Trans>` and `_(msg` I add in this pull request, we will be able to translate some of the UI that were not available for translation before!

Sparkled in this pull request are some `context` added to some strings, where it is necessary for translators to differentiate between, for example, `Reply` as an action (on a button), and `Reply` as a description (next to it).

Let's note that even though I tried to cover all the codebase, I am sure that there are some parts that will be missing, so it's not the last internationalization pull request you'll have to review.

ping @ansh who did the first, huge i18n work — hope you can review that before it diverge too much from main 😅


Signez:
Oh, I just noticed, after all that work, that @quiple also did some of that work on their pull request #2210. Of course, feel free to move forward with either or both! I think mine is a bit more in-depth on some stuff (including labeling all `accessibilityLabel`, for the reasons I explained above), and @quiple's work tackled some I overlooked (moderation messages, for example) but I'm sure both our works are useful in aiming for even better internationalization of Bluesky!

ansh (approved):
I looked through all the files individually and this looks great.

Well done!

I'll let @pfrazee merge it though, just because it adds about 300 translations per language.

pfrazee:
Insane PR. I gave it a thorough scan and it looks clean. Nicely done

PR #2448, opened 2024-01-08
Splash: reduce motion + dark mode

estrattonbailey:
Adds a separate animation for reduced motion that simply fades in the logo and app interface, no logo zoom.

Adds a dark-mode splash image and handling.


pfrazee (approved):
👍 Solid

PR #2457, opened 2024-01-09
Add accept-language header

CooperEdmunds:
Added an Accept-Language header (where the value is the user's Content Language selections joined by commas) to the three getFeed calls. There will be a PR on the backend to forward this header to the feed generator.


CooperEdmunds:
@pfrazee I discussed this with Ansh but will need someone else to review. Thanks!

pfrazee (approved):
Looks good! Thanks!

PR #2462, opened 2024-01-10
Mark more texts for localization

quiple:
This is about some localizations that were not included in #2440.

* Localize time period strings in `time.ts`. The `now` should be localized, and depending on the language, the `h`, `m`, `s`, `d`, `mo` may also need to be localized.
* Localize date strings in `time.ts` and number format strings in `format.ts`. It works, but I'm not sure if it's good as is.
* Localize any other strings.

Thank you!


src/view/screens/Debug.tsx
```diff
@@ -177,14 +177,17 @@ function ErrorView() {
 }
 
 function NotifsView() {
+  const {_} = useLingui()
   const triggerPush = () => {
     // TODO: implement local notification for testing
   }
   const triggerToast = () => {
-    Toast.show('The task has been completed')
+    Toast.show(_(msg`The task has been completed`))
```
quiple, line 185: @ansh I'm confused because there are already localized strings in the `Debug.tsx` file, should we don't localize them all in `Debug.tsx`?

src/view/screens/Settings.tsx
```diff
@@ -699,7 +699,7 @@ export function SettingsScreen({}: Props) {
         {__DEV__ ? (
           <ToggleButton
             type="default-light"
-            label="Experiment: Use AppView Proxy"
+            label={_(msg`Experiment: Use AppView Proxy`)}
```
quiple, line 702: @ansh If this shouldn't be localized either, I'll remove it.

src/view/screens/Debug.tsx
Signez: Let's note that I didn't want to translate all the "fake" text in most of this page, for obvious reasons (it's a debug page, and translating it won't add much value), but translating those two Toasts is not that much work and would make sure we could maybe check later through CI (or types?) that we are always translating toasts.

src/view/com/auth/create/Policies.tsx
```diff
@@ -72,19 +77,21 @@ export const Policies = ({
       0,
       <Text key="and" style={pal.textLight}>
         {' '}
-        and{' '}
+        <Trans>and</Trans>{' '}
```
quiple, line 80: Maybe `<Trans>{' '}and{' '}</Trans>` would be better?

quiple:
```tsx
        <Trans>
          <Text type="sm" style={[{flexWrap: 'wrap'}, textStyles]}> and </Text>
        </Trans>
```
This one shows both whitespaces correctly,

```tsx
        <Trans>
          <Text type="sm" style={[{flexWrap: 'wrap'}, textStyles]}>
            {' '}
            and{' '}
          </Text>
        </Trans>
```
But this removes the left whitespace.

Why is this and what can I do about it?

PR #2464, opened 2024-01-10
fix rendering of external in quote embeds

haileyok:
External embeds are not currently displayed in a quote post, which results in context loss.

I also think it makes more sense here to remove the `.find()` since - unless this is done in anticipation of future changes - there will never be more than a single embed for a post: https://github.com/bluesky-social/atproto/blob/50f70453a9e49621956e4e820226c0e220fe138e/packages/bsky/src/services/feed/views.ts#L434 (thanks @mary-ext). Instead we just need to get the first possible embed from the array.

We do still need to check for types, since we don't want to show a quote's quote.

RocketSim_Recording_iPhone_15_Pro_6 1_2024-01-09_18 36 15

Of course we could add
```tsx
  const externalEmbed = React.useMemo(
    () =>
      quote.embeds?.find(
        embed =>
          AppBskyEmbedExternal.isView(embed)
      ),
    [quote.embeds],
  )
```
but this feels like some additional logic that isn't needed. Correct me if wrong.

### DIFF
--- a/src/view/com/util/post-embeds/QuoteEmbed.tsx
+++ b/src/view/com/util/post-embeds/QuoteEmbed.tsx
@@ -6,6 +6,7 @@ import {
   AppBskyEmbedImages,
   AppBskyEmbedRecordWithMedia,
   ModerationUI,
+  AppBskyEmbedExternal,
 } from '@atproto/api'
 import {AtUri} from '@atproto/api'
 import {PostMeta} from '../PostMeta'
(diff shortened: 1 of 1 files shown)


pfrazee:
Strong +1 to this. It was the wrong call not to do link cards.

I actually don't recall why `app.bsky.embed.record#viewRecord` uses an array of `embeds` but I assume it was for future defensibility.

pfrazee (approved):

PR #2465, opened 2024-01-10
Reduce polling

### LINKED ISSUE #2451: Reduce polling based on known state
We can reduce overall traffic by ensuring that

- Feed polls don't fire if they already known there is unread
- Notifications poll fires less frequently if there are is unread (1/2 as often)

### LINKED ISSUE #2454: Change preferences polling into on-foreground fetches
Currently we poll for preferences regularly. I believe we can switch to a fetch on foreground events (web and mobile) and then also fetch when viewing preferences directly and that should cover it.

pfrazee:
- a4c51da869f321a9f39f2da253e76566bb6f4450 Stops fetching preferences and profiles on an interval and instead does it on focus (close #2454)
- 7f9827674b1f8900457ba5f9b925b2a0926fbc56 Makes it possible to refresh the saved feeds in the feeds screen
- 2963d5789c1a7f02f7227636c1264465e6d982f2 and b7686080c9868777feb79daece148d8777412b32 Reduces feed and notifications polling if there's known notifications (close #2451)
- 4d87706c3513edc58c0200d3849eb0b5d723f2ca Disables the hard refresh after 1 hour to test if that's the cause of the random reload on start bug

### DIFF
--- a/src/lib/react-query.ts
+++ b/src/lib/react-query.ts
@@ -1,11 +1,39 @@
-import {QueryClient} from '@tanstack/react-query'
+import {AppState, AppStateStatus} from 'react-native'
+import {QueryClient, focusManager} from '@tanstack/react-query'
+import {isNative} from '#/platform/detection'
+
+focusManager.setEventListener(onFocus => {
+  if (isNative) {
+    const subscription = AppState.addEventListener(
+      'change',
+      (status: AppStateStatus) => {
+        focusManager.setFocused(status === 'active')
+      },
+    )
+
+    return () => subscription.remove()
+  } else if (typeof window !== 'undefined' && window.addEventListener) {
+    // these handlers are a bit redundant but focus catches when the browser window
+    // is blurred/focused while visibilitychange seems to only handle when the
+    // window minimizes (both of them catch tab changes)
+    // there's no harm to redundant fires because refetchOnWindowFocus is only
+    // used with queries that employ stale data times
+    const handler = () => onFocus()
+    window.addEventListener('focus', handler, false)
+    window.addEventListener('visibilitychange', handler, false)
+    return () => {
+      window.removeEventListener('visibilitychange', handler)
+      window.removeEventListener('focus', handler)
+    }
+  }
+})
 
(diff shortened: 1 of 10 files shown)


src/view/com/posts/Feed.tsx
```diff
@@ -146,11 +149,12 @@ let Feed = ({
   React.useEffect(() => {
     if (enabled) {
       const timeSinceFirstLoad = Date.now() - lastFetchRef.current
-      if (timeSinceFirstLoad > REFRESH_AFTER) {
+      // DISABLED need to check if this is causing random feed refreshes -prf
+      /*if (timeSinceFirstLoad > REFRESH_AFTER) {
```
estrattonbailey, line 153: Ooooooooh yeah this looks promising

src/view/screens/Feeds.tsx
```diff
@@ -151,9 +152,12 @@ export function FeedsScreen(_props: Props) {
   }, [query, debouncedSearch])
   const onPullToRefresh = React.useCallback(async () => {
     setIsPTR(true)
-    await refetchPopularFeeds()
+    await Promise.all([
+      refetchPreferences().catch(_e => undefined),
+      refetchPopularFeeds().catch(_e => undefined),
```
estrattonbailey, line 157: Do we need to wait for prefs before fetching other feeds or nah

src/lib/react-query.ts
```diff
@@ -1,11 +1,39 @@
-import {QueryClient} from '@tanstack/react-query'
+import {AppState, AppStateStatus} from 'react-native'
+import {QueryClient, focusManager} from '@tanstack/react-query'
+import {isNative} from '#/platform/detection'
+
+focusManager.setEventListener(onFocus => {
```
estrattonbailey, line 5: TIL

estrattonbailey (approved):

src/view/screens/Feeds.tsx
pfrazee: nah

PR #2470, opened 2024-01-10
(optional) In app browser

mozzius:
Adds `expo-web-browser`. Keeps current behaviour on web, and does not get used unless the user specifically opts in.

When opening a link, if there's no saved preference for the in app browser it will open a model asking if you want to open in the in app browser or in the default browser. A toggle has been added to the settings to change this setting later.

Simulator Screenshot - iPhone 15 - 2024-01-10 at 14 00 50


mackuba:
Omg yes thank you! 🙏

haileyok:
goated feature

pfrazee:
@mozzius Beginning review but FYI -- we decided not to run the intl extraction for PRs because it adds risk of conflicts. I just run it before every release. It'll lag behind but that's going to happen no matter what.

mozzius:
> @mozzius Beginning review but FYI -- we decided not to run the intl extraction for PRs because it adds risk of conflicts. I just run it before every release. It'll lag behind but that's going to happen no matter what.

Yeah that makes way more sense. You might want to write that in the README somewhere

beverloo:
Hi both - this is a cool interface! Would you be able to share more about the motivation to present this decision to users in such an upfront manner? Do you know (or measure) what people generally prefer?

PR #2472, opened 2024-01-10
Add Latin American Spanish localization

Weeesly:
Latin American Spanish language


pfrazee:
I really hate to ask this but, how significantly different is this from the existing spanish translation? The problem is that we have very little resources to be maintaining multiple translations, and are already struggling to keep up with the currently active ones.

Weeesly:
My goal in undertaking this translation was to provide a version more in line with Latin American linguistic nuances, thus improving the experience for Spanish-speaking users. Although not explicitly stated, the existing Spanish translation reflects that of Spain.

The current Spanish translation presents some contextual issues, which prompted me to address them and align the language more closely with a "Latin American" style. Here are some examples with terms widely used in the file:

Feed ---> muro (I'm not from Spain but apparently "feed" is "noticias", "news").

Un-mute list ---> "quitar silencio" (In the existing translation, it is "Desactivar la opción de silenciar la lista" which is incorrect).

Cancel quote post ---> "Cancelar cita" (In the exiting translation, "Cancelar la publicación de un presupuesto", literally, "Cancel the publication of a budget".

The differences between Spanish in Spain and Latin America are mainly grammatical. For example, some verb conjugations, vocabulary choices and regional expressions may vary. The goal is to make the translation more linguistically consistent with the nuances of Latin American Spanish.

Issue #1742, opened 2023-10-24
Broken links from "Copy Post Text"


YuriGal:
**Describe the bug**

If a post text contains a long URL, using "Copy Post Text" copies text with broken URL.

**To Reproduce**

1. Create post with a long URL in post's text, e.g https://chrome.google.com/webstore/detail/sky-follower-bridge/behhbpbpmailcnfbjagknjngnfdojpko
2. Use "..." -> Copy Post Text
3. Paste the text somewhere

Text will contain a broken link, e.g. in the above example `chrome.google.com/webstore/det...`

**Expected behavior**

Text should contain full unbroken URL

intrnl:
I suppose one question before trying to make this work would be what to do for links that aren't valid, e.g. [this post from Techmeme](https://bsky.app/profile/did:plc:pv7fudnt4dspurzdnyq73pfe/post/3kbzrhz3v3p2p) where it uses a completely different label.

should links from those be copied as well, and how would that look like?

YuriGal:
That is a good question. I can imagine 2 possible solutions:
1. On copying replace the labels with the original URLs. But without labels it might get confusing, and post risks to go over character limit.
2. Add a general markdown support for links. So every link can be represented in post text as `[label](url)`

estrattonbailey:
This will also need to be fixed for, say, when Slack unfurls a URL 👍

PR #2481, opened 2024-01-11
Use the RichText facets when copying post text

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/1742

- Adds a `richTextToString` function that looks at the rich text facets and adjusts the URL being copied as necessary
- Wire it up into the "Copy Post Text" functionality

If an invalid/misleading link is being copied, we use the `[label](url)` format, but otherwise it just unfurls the url, based on https://github.com/bluesky-social/social-app/issues/1742#issuecomment-1779051689


pfrazee (approved):
Yep this is great. Very nice

Issue #2443, opened 2024-01-08
Website Search Bar behaves like Password Text Field on Chrome


xtianpu:
**Describe the bug**
as Title says

**To Reproduce**

1. Just click search bar with Chrome browser

**Expected behavior**

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/8172235/8792423d-9f9b-4937-b2a6-5d465fc4d17c)

**Details**

 - Platform: Chrome
 - Platform version: 120.0.6099.131
 - App version: 1.63.0

**Additional context**

Issue #1648, opened 2023-10-10
Modal dialogs make Chrome treat search fields as password inputs


MarioLiebisch:
**Describe the bug**
It seems that at least some modal dialogs shown inside the app will make Chrome treat search input fields as password inputs, unintentionally offering auto-completion/password generation.

While this might appear as a minor nuisance, it's actually rather disruptive, since this also causes Chrome to offer or directly save the input text as a password, if you try to confirm a search with the return button, potentially overwriting the actual password used for accounts or at least spamming its built-in password manager with non-accounts.

**To Reproduce**

* Refresh the page to start from a clean slate.
* Open the "Settings" page.
* Click a few times into the search input box on the top right.
* The box gains input focus but works as expected.
* Click the link to change your account's email address.
* Cancel editing the email address.
* Click twice into the search input box on the top right.
* Observe the browser suddenly offering autofill options for passwords.

**Expected behavior**

Search input fields should never be interpreted as password inputs by the browser.

**Screenshots**

![Chrome offering to generate a password since it assumes the search input field is actually a password field.](https://github.com/bluesky-social/social-app/assets/1854245/4406688a-556c-48c2-9706-ce39f377fdb5)

![GIF demonstrating the issue/reproduction steps.](https://github.com/bluesky-social/social-app/assets/1854245/b85ab4c4-d81b-42e5-9978-f1e7e1121f8d)

**Details**

 - Platform: Chrome (Windows 11, x64)
 - Platform version: 117.0.5938.150
 - App version: 1.52.0

**Additional context**

jonolds:
still having this issue when I click on some search results.

PR #2482, opened 2024-01-11
Don't set autocomplete on search input

mary-ext:
Potentially fixes https://github.com/bluesky-social/social-app/issues/2443
Potentially fixes https://github.com/bluesky-social/social-app/issues/1648

I can't reproduce this, but `autocomplete=on` was being put on text inputs, that means the browser gets to make its own judgement on what it should be autocompleting, so let's turn it off and see if this magically solves the issue.

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -572,6 +572,7 @@ export function SearchScreen(
             accessibilityLabel={_(msg`Search`)}
             accessibilityHint=""
             autoCorrect={false}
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
yeah this seems like a good bet, appreciate it

xtianpu:
Thanks! It's fixed now.

jonolds:
I'm still getting the popup to update my bluesky password when I click on search results. It happens often but not always. When it does happen, the suggested password to save is the search term.

Issue #2449, opened 2024-01-08
Detect language during authoring and suggest a correction if needed


pfrazee:
Right now the selected language carries over from the last post. It's very easy to miss that control! So people often post with the wrong language tag.

It would not be a heavy lift to use `lande` to detect which language they're posting in and, if lande is confident enough, suggest they change the tag to that language.

pfrazee:
Yeah I think that's the move

PR #2484, opened 2024-01-11
Suggest post language correction

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/2449

- Adds `code3ToCode2Strict` mapping function because it seems that `lande` returns language codes that we don't have in our mapping, we should just filter those out. (Mixed English+Japanese seems to be returning `cmn` and I can't figure out what language is)
- It skips over short posts (currently set to length < 10) because `lande` is going to return incorrect matches for them anyway
- `lande` is being run in a `requestIdleCallback`, not sure if this is the right call, perhaps a timeout would be a better idea instead?
- I'm hoping that a threshold of 0.85 should be "correct" enough, but we could probably bump this up to 0.90

image

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -22,6 +22,14 @@ export function code3ToCode2(lang: string): string {
   return lang
 }
 
+export function code3ToCode2Strict(lang: string): string | undefined {
+  if (lang.length === 3) {
+    return LANGUAGES_MAP_CODE3[lang]?.code2
+  }
+
+  return undefined
+}
+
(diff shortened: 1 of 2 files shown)


pfrazee:
PR is great. I factored it out into a separate component and now I'm testing a bit.

Going to need to tune the confidence level from lande and see if it can get to a better spot. False positives are kinda ick and lande immediately thought I was writing in french.

pfrazee:
Superseded by https://github.com/bluesky-social/social-app/pull/2486

PR #2485, opened 2024-01-11
Resolve facets on list descriptions

mary-ext:
List descriptions can contain facets and in the app is rendered with a `<RichText>`, all that's left is to add the resolution step

- Description state is initialized by serializing it to string first and then creating a RichText from it, this way we're starting with a pretty much blank state.
- `richTextToString` function has been modified to add a `loose` property, when it's `true`, invalid links are to be copied with the `[label](url)` format, but if not then it just uses the label. The idea is that copying post text usually means it being shared elsewhere, while for this form we just want to reuse it for the user to edit.
- Description now has a visible counter now rather than mysteriously appending with an ellipsis.

image

image


pfrazee (approved):
Wonderful as usual Mary. Thank you!

PR #2486, opened 2024-01-11
Suggest post language correction

pfrazee:
Building on #2484

- Factors out into a separate component
- Tunes the lande confidence to reduce false positives

Copying over original description:

Fixes https://github.com/bluesky-social/social-app/issues/2449

- Adds `code3ToCode2Strict` mapping function because it seems that `lande` returns language codes that we don't have in our mapping, we should just filter those out. (Mixed English+Japanese seems to be returning `cmn` and I can't figure out what language is)
- It skips over short posts (currently set to length < 10) because `lande` is going to return incorrect matches for them anyway
- `lande` is being run in a `requestIdleCallback`, not sure if this is the right call, perhaps a timeout would be a better idea instead?
- I'm hoping that a threshold of 0.85 should be "correct" enough, but we could probably bump this up to 0.90

image

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -22,6 +22,14 @@ export function code3ToCode2(lang: string): string {
   return lang
 }
 
+export function code3ToCode2Strict(lang: string): string | undefined {
+  if (lang.length === 3) {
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
This is super nice! Works great

PR #2493, opened 2024-01-11
Truncate long email addresses

mary-ext:
My email gets pretty long :sweat_smile: 

image


haileyok:
test yourself but i think you should be able to just set `numberOfLines={1}` and `{flex: 1}` to the text style without need for `flexShrink` on up the tree

pfrazee (approved):
Sweet

Issue #1102, opened 2023-08-04
Make it possible to copy ALT text from an image on mobile


mackuba:
**Is your feature request related to a problem? Please describe.**

If someone is cross-posting content manually to multiple platforms, I think it's a pretty common action to make a post with an image with alt text and then copy the post text and the alt text to the other app like Twitter or Ivory. The problem here is that it doesn't seem to be possible to copy alt text from an image in the Bluesky app. The alt text field can't be selected and does not react to taps.

The workflow:

1. Open Bluesky app
2. Open create new post
3. Enter post text
4. Add an image
5. Add image alt text
6. Submit the post
7. Go to Twitter app, open create new post dialog
8. Go back to Bluesky, copy post text, go to Twitter, paste it there
9. Go back to Bluesky, click the image, copy alt text  - ??? - no profit

<p><img src="https://github.com/bluesky-social/social-app/assets/28465/d9dcb8d1-398b-4f50-9a4a-214ee13568e7" width="500"></p>

**Describe the solution you'd like**

1) Option 1: make the alt text textfield selectable, so the native text actions popover appears and you can tap "copy"
2) Option 2: copy the alt text on tap or long-tap

deceptionagent:
regarding option 2:

maybe double-tap instead of single-tap to reduce accidentally blowing away your pasteboard with errant touches.

deceptionagent:
> regarding option 2:
> 
> maybe double-tap instead of single-tap to reduce accidentally blowing away your pasteboard with errant touches.

oh right - since this bug was written, the alt text UI has changed so single-tap expands or collapses the alt text overlay. so yeah you'd probably need it to be double-tap, press-and-hold (P&H), a discrete button, or a selectable text label on P&H.

deceptionagent:
when you go to do this, make sure whatever interaction method you decide on can actually be performed using VoiceOver (and the Android equivalent). i don't know what RN's accessibility story will be here with regard to interactions on the text label itself, so if you need to use a discrete button then that's what you gotta do.

remember, people who use screenreaders need to reproduce alt text as well.

deceptionagent:
it occurs to me that a discrete button would be more easily discoverable to screenreaders. could just be a small, simple icon. (assuming RN can properly set NSAccessibilityDescriptionAttribute.)

pfrazee:
Appreciate it chelsea

mozzius:
We fixed this

Issue #954, opened 2023-07-03
Make it possible to select text in posts etc


0x53A:
I would like all/most text to be selectable with the default system tools.

For example, on web, I can select parts of a post (by long-tapping), and then do anything with it, including (but not limited to) using the native translation feature.

![image](https://github.com/bluesky-social/social-app/assets/4236651/c0cfcf33-96ac-423e-8ce2-9afcaf26339e)

![image](https://github.com/bluesky-social/social-app/assets/4236651/2bf75286-1840-40df-a7b8-907874bf7919)

On the iOS app, I would have to use the triple-dot menu to copy&paste the whole post into the notes app.

In addition to post text, I would also like alt-text of images to be selectable in app (it is on web). Maybe also profile descriptions?

Lastly, on iOS, there is automatic  (on-device!) OCR for images. I believe apps can opt into this. This would be great for images with text.

This is how it looks on web:

![image](https://github.com/bluesky-social/social-app/assets/4236651/7447d728-888b-43e6-a789-f25c82fcce58)

I’m not an iOS developer but found this article which seems to describe how to enable this: https://developer.apple.com/documentation/visionkit/enabling_live_text_interactions_with_images

ansh:
I would love for this to be possible as well, unfortunately it is quite hard to implement in React Native. I got to a decent place in this branch #453 but the issue happens when you have links and rich text which makes it very difficult to do without having unintended side-effects.

intrnl:
would a toggle suffice as a compromise? I'd imagine when selecting text you wouldn't want to accidentally tap a link and ruin your selection

fabriziobertoglio1987:
I'm available to work on this issue. Thanks

haileyok:
This got merged in 👍

PR #2495, opened 2024-01-11
RNUITextView, selectable post contents for main thread post

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/954
Fixes https://github.com/bluesky-social/social-app/issues/1102 although it might be worth exploring easier alt-copy options in the future. At a minimum though, this allows for that selection.

https://www.youtube.com/watch?v=0aufAdmfKkE

## RNUITextView Module
React Native's `Text` component allows for "infinite" nesting of further `Text` components. To make a true "drop-in",
we want to do the same thing.

To achieve this, we first need to handle determining if we are dealing with an ancestor or root `UITextView` component.
We can implement similar logic to the `Text` component [see Text.js](https://github.com/facebook/react-native/blob/7f2529de7bc9ab1617eaf571e950d0717c3102a6/packages/react-native/Libraries/Text/Text.js).

We create a context that contains a boolean to tell us if we have already rendered the root `UITextView`. We also store
the root styles so that we can apply those styles if the ancestor `UITextView`s have not overwritten those styles.

All of our children are placed into `RNUITextView`, which is the main native view that will display the iOS `UITextView`.

We next map each child into the view. We have to be careful here to check if the child's `children` prop is a string. If
it is, that means we have encountered what was once an RN `Text` component. RN doesn't let us pass plain text as
children outside of `Text`, so we instead just pass the text into the `text` prop on `RNUITextViewChild`. We continue 
down the tree, until we run out of children.

On the native side, we make use of the shadow view to calculate text container dimensions before the views are mounted.
We cannot simply set the `UITextView` text first, since React will not have properly measured the layout before this
occurs.

As for `Text` props, the following props are implemented:

- All accessibility props
- `allowFontScaling`
- `adjustsFontSizeToFit`
- `ellipsizeMode`
- `numberOfLines`
- `onLayout`
- `onPress`
- `onTextLayout`
- `selectable`

All `ViewStyle` props will apply to the root `UITextView`. Individual children will respect these `TextStyle` styles:

- `color`
- `fontSize`
- `fontStyle`
- `fontWeight`
- `fontVariant`
- `letterSpacing`
- `lineHeight`
- `textDecorationLine`

## Parity on Android

Few small changes:
- Remove the `Link` wrapper on the main post thread item. There is no need for this to be a touchable opacity, and it looks bad when selecting text
- Unsure about this one...I set `PostThread`'s `removeClippedSubviews` to `false`. There's a bug in RN (seemingly an Android ListView related thing) where `selectable` won't work inside of a `FlatList`. Oddly, this prop wasn't set to begin with, and the default is supposed to be `false`. Regardless, I only considered this viable since we probably can rely on `PostThread` to be relatively small (instead of say the feed that will infinitely continue). I don't notice perf issues on the few devices I have, but...

See: https://github.com/facebook/react-native/issues/26264
https://stackoverflow.com/questions/11064115/make-items-within-listview-selectable
https://stackoverflow.com/questions/31726470/how-i-can-create-text-selection-in-listview-item

This is probably fixable though, if it is a perf issue.


pfrazee:
🚨 s-tier PR alert 🚨

is this ready for review, @haileyok ?

haileyok:
@pfrazee s-tier is the only way to live

ready now, made alt text selectable as well now. there's some changes/improvements that can continually get made to the module, but right now all the needed Text features work.

haileyok:
Oh, this was also tested with voiceover and accessibility inspector. Things are fine there (aside from the just overall screen reader issues that already exist)

haileyok:
cleaned up a bunch of stuff while writing a little post about doing this. otherwise been testing it all day and its working great on my machine™

pfrazee:
fire fire fire

haileyok:
just to make sure, ran a quick profile on this just to see if there were any perf implications (that would be negligible anyway but). looks like like its is on near equal footing with Text (i mean this is <1ms renders here so its really irrelevant, but point is there's nothing goofy 👍 )

Screenshot 2024-01-16 at 3 26 42 PM

Screenshot 2024-01-16 at 3 28 47 PM

also sneaking in a little adjustment to the alt text expansion so it is less jarring.

haileyok:
okay that should do it, fixed a few of the minor things i noticed over the past couple days.

haileyok:
@pfrazee yea thats great. im actually dropping expo-modules since you unfortunately cant use the shadowview there. for perfect parity, we want to override flexbox which requires calculating before mount in the shadowview. would rather test that for a bit anyway.

haileyok:
merged main in, this supersedes https://github.com/bluesky-social/social-app/pull/2563 since we need to move the `onPress` event to `<Text>`. Otherwise, we will get interference with the native selection gestures (press and hold) and the view itself. Otherwise, same thing.

pfrazee (approved):
Works super well. Tested in simulators and on-device iOS. Now waiting for android test device to juice up and just triple check there.

I'm okay with removing the `<Link>` wrapper on posts in the post thread. Didn't notice any real difference in the behaviors on any platform.

Android's alt-text selection is still wonky, even testing here. What I see now is that it enters selection state but then fails to retain afterwards. I'll try to address this in a separate PR.

I did have to make one change for Web. It was giving me a function-not-found:

```diff
diff --git a/modules/react-native-ui-text-view/src/index.tsx b/modules/react-native-ui-text-view/src/index.tsx
index ada5c4d9..d5bde136 100644
--- a/modules/react-native-ui-text-view/src/index.tsx
+++ b/modules/react-native-ui-text-view/src/index.tsx
@@ -24,6 +24,7 @@ export interface RNUITextViewChildProps extends TextProps {
 }

 export const RNUITextView =
+  UIManager.getViewManagerConfig &&
   UIManager.getViewManagerConfig('RNUITextView') != null
     ? requireNativeComponent<RNUITextViewProps>('RNUITextView')
     : () => {
@@ -31,6 +32,7 @@ export const RNUITextView =
       }

 export const RNUITextViewChild =
+  UIManager.getViewManagerConfig &&
   UIManager.getViewManagerConfig('RNUITextViewChild') != null
     ? requireNativeComponent<RNUITextViewChildProps>('RNUITextViewChild')
     : () => {
```

haileyok:
Unfortunate about the Android issue. I'll play with it on a real device and see what it feels like there.

pfrazee:
Well, weirdly on my test device the alt text selection worked fine. Let's give it a deploy and see how it goes.

pfrazee:
Had to do a local squash & merge in my CLI because of my changes, so GitHub isn't detecting as merged but indeed this is!!

Issue #2303, opened 2023-12-25
The `lang` attribute of the `html` should change based on the app language


quiple:
**Is your feature request related to a problem? Please describe.**

Changing the app language doesn't change the `lang` property of the `html`.

**Describe the solution you'd like**

The lang attribute of the HTML should change based on the app language. This will typically render the correct font for each language.

**Describe alternatives you've considered**

**Additional context**

<img width="706" alt="스크린샷 2023-12-25 21 27 18" src="https://github.com/bluesky-social/social-app/assets/11014257/1ab076b4-e950-492c-85dc-95facce451de">

----

I'd also like you to consider giving each post a `lang` attribute to match the language of the post. (Add lang attribute on the post content #1404)

PR #2496, opened 2024-01-11
Set document lang attribute to the app language

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/2303

Wanted to do it in `src/state/preferences/languages.tsx` to match `src/state/shell/color-mode.tsx` but that would've meant setting an unsanitized value which might be incorrect as translations changes/gets removed/etc


pfrazee (approved):
Oh nice, yea 👍

Issue #1846, opened 2023-11-08
Embedding lists that have a handle in the URL fails


Bossett:
**Describe the bug**

When embedding a list as a link card, the embed fails when using a URL with a handle in it: https://bsky.app/profile/bossett.bsky.social/lists/3jx3w32axax2f 

The same list works if you instead construct the URL with did: https://bsky.app/profile/did:plc:jfhpnnst6flqway4eaeqzj2a/lists/3jx3w32axax2f

**To Reproduce**

1. Visit any user's list page through the app
2. Copy the list URL
3. Paste into new post, click the embed card
4. The embed card begins to generate, then fails

**Expected behavior**

The list should embed correctly

**Details**

 - Platform: chrome
 - Platform version: 118.0.5993.120
 - App version: 1.56.0

PR #2497, opened 2024-01-11
Resolve handle in feed/list URLs

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/1846

It looks like the new feed/list pages are able to resolve the handle without issues unlike before, should we make the "Share feed" and "Share list" menu items use the handle instead of DID for the URL?


pfrazee (approved):
Wonderful, works great!

PR #2498, opened 2024-01-11
Ensure sessions and tokens are synced between tabs

estrattonbailey:
This should address the multiple reports of unexpected signouts.

The issue was with users using multiple tabs on web. Both tabs would retain the same `refreshJwt` in memory, and refresh tokens can be used only once. So once a refresh token was rotated in tab A, the tab B _was_ notified but did not update that token on the `agent`, and so when tab B tried to get a new access token, `refreshSession` would fail.

This was a little tricky to pinpoint because if both tabs were periodically reloaded within the access token expiry window, our session handling would pick up the new refresh token from storage and continue as normal for a time. So this issue only existed in long-running multi-tab sessions.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -193,11 +193,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const clearCurrentAccount = React.useCallback(() => {
-    logger.debug(
-      `session: clear current account`,
-      {},
-      logger.DebugContext.session,
-    )
+    logger.warn(`session: clear current account`)
     __globalAgent = PUBLIC_BSKY_AGENT
     queryClient.clear()
     setStateAndPersist(s => ({
@@ -322,8 +318,8 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const logout = React.useCallback<ApiContext['logout']>(async () => {
+    logger.info(`session: logout`)
     clearCurrentAccount()
-    logger.debug(`session: logout`, {}, logger.DebugContext.session)
     setStateAndPersist(s => {
       return {
         ...s,
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -551,30 +551,36 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     return persisted.onUpdate(() => {
       const session = persisted.get('session')
 
-      logger.debug(`session: onUpdate`, {}, logger.DebugContext.session)
+      logger.info(`session: persisted onUpdate`, {})
 
-      if (session.currentAccount) {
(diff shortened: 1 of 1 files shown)


src/state/session/index.tsx
```diff
@@ -551,41 +551,59 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     return persisted.onUpdate(() => {
       const session = persisted.get('session')
 
-      logger.debug(`session: onUpdate`, {}, logger.DebugContext.session)
+      logger.info(`session: persisted onUpdate`, {})
```
estrattonbailey, line 550: Sending these to Sentry as breadcrumbs so we can track this issue

src/state/session/index.tsx
```diff
@@ -193,11 +193,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const clearCurrentAccount = React.useCallback(() => {
-    logger.debug(
-      `session: clear current account`,
-      {},
-      logger.DebugContext.session,
-    )
+    logger.warn(`session: clear current account`)
```
estrattonbailey, line 196: Sending these to Sentry so we can monitor for spikes

src/state/session/index.tsx
```diff
@@ -322,8 +318,8 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const logout = React.useCallback<ApiContext['logout']>(async () => {
+    logger.info(`session: logout`)
```
estrattonbailey, line 321: Will be included in any `session: clear current account` warnings

src/state/session/index.tsx
```diff
@@ -551,41 +547,59 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     return persisted.onUpdate(() => {
       const session = persisted.get('session')
 
-      logger.debug(`session: onUpdate`, {}, logger.DebugContext.session)
+      logger.info(`session: persisted onUpdate`, {})
```
estrattonbailey, line 550: Sending these to Sentry as breadcrumbs so we know if a broadcast resulted in a logout

src/state/session/index.tsx
```diff
@@ -551,41 +547,59 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+         * or by a `logout` call  which nukes all accounts tokens)
+         */
         clearCurrentAccount()
       }
+
+      setState(s => ({
+        ...s,
+        accounts: session.accounts,
+        currentAccount: session.currentAccount,
+      }))
```
estrattonbailey, line 600: We don't directly access tokens anywhere atm, but I thought it made sense to keep React state in sync too so we don't surprise ourselves later.

pfrazee (approved):
Looks good. Will test by leaving open in my dev env all day

PR #2499, opened 2024-01-11
Toggle minimal shell on any scroll for web

gaearon:
On native, the minimal shell is shown/hidden in sync with the gesture. However, on the web, this doesn't work well because we don't actually know when the dragging starts and ends. We only get scroll events.

As a result, it gets "stuck" on the web in a half-visible state. This is especially confusing on mobile web where it's easy to mistap.

Let's make the web implementation show/hide simply based on whether we're scrolling up or down. I'm also forcing the header to always show for a little bit (first 200 pixels) so that it disappearing out of sync with the top post doesn't feel weird.

[Review without whitespace.](https://github.com/bluesky-social/social-app/pull/2499/files?w=1)

Should have no effect on native.

## Before

https://github.com/bluesky-social/social-app/assets/810438/94949aec-cf1b-40d5-a139-e8dfd73ee3d6

## After

https://github.com/bluesky-social/social-app/assets/810438/4b0d55b7-4fc4-40da-a95d-e5d721850801

### DIFF
--- a/src/view/com/util/MainScrollProvider.tsx
+++ b/src/view/com/util/MainScrollProvider.tsx
@@ -3,8 +3,10 @@ import {ScrollProvider} from '#/lib/ScrollContext'
 import {NativeScrollEvent} from 'react-native'
 import {useSetMinimalShellMode, useMinimalShellMode} from '#/state/shell'
 import {useShellLayout} from '#/state/shell/shell-layout'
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Tested locally, works great

pfrazee (approved):
sgtm!

PR #2502, opened 2024-01-12
Fix typos, improve localization and unify separators

quiple:
* Correct string `Reposted by {0})` to `Reposted by {0}` in `FeedItem.tsx`.
* For ease of localization, separate `liked your custom feed{0}` into two versions.
* Replace the bullet in `PostThreadItem.tsx` with a middle dot, the same as the separator used in other parts.

Thank you!


pfrazee (approved):
Thanks!

Issue #849, opened 2023-06-03
Add support for ”from:username” search operator


alimony:
It would be nice if I could search only for posts from a specific user, most often myself. A simple interface for this would be the search operator ”from:” which is familiar from several other platforms.

HeyItsLollie:
I often do this on other platforms when I'm trying to find one of my own posts to re-share or reply to, it's extremely useful.

If "from:" is implemented as a search operator, I'd also like to request "self:" as additional shorthand for "from:`your.username`"

potatoqualitee:
Looking forward to this one! Thank you

OmarShehata:
Tried this today and was surprised that it isn't supported yet! Would like to add that it is critical for me when finding both my own posts to link back to as well as others' posts from a few days ago I want to respond to

tkasci:
I agree, it would be very useful.

mschwendener:
In the meantime, there's https://oyin.bo/receipts 

(via https://bsky.app/profile/rvbdrm.com/post/3kg7iz6ztct23 )

brendanjones:
> If "from:" is implemented as a search operator, I'd also like to request "self:" as additional shorthand for "from:your.username"

`from:me` is even simpler, and arguably more intuitive. Would definitely want this, I use it all the time on Mastodon.

mschwendener:
> you should be able to do from:bsky.app etc 

Interesting. 

I can search within the account bsky.app, for example 
from:bsky.app rolling 

(This does not change the URL)

But when I want to search in my own posts or some random other example, I get no hits, although I know the posts are there. 

from:manuschwendener.bluesky.social paris
should find 
https://bsky.app/profile/manuschwendener.bsky.social/post/3ki6b7lpriw2a 

from:geeka.bsky.social
should find 
https://bsky.app/profile/geeka.bsky.social/post/3ke3tn4l6tt2p

mschwendener:
Oh, wait: it works if I use the search field in the right column, but not in the left column? 

<img width="1064" alt="screenshot_4716" src="https://github.com/bluesky-social/social-app/assets/4497705/f1c4faa2-ecbf-4fab-a22f-ac07ad8ce099">

mschwendener:
Yeah, it's strange. I now get the hits for both examples, from both search places. 

But it really didn't work just a few seconds ago. I can still reproduce it when I search via the search field in the middle column, which gets visible _after_ a search. 

[...]

PR #2504, opened 2024-01-12
Add from:me search functionality

mary-ext:
Addresses https://github.com/bluesky-social/social-app/issues/849#issuecomment-1876982941
nit: we should probably close that issue already since it's been fixed for a while now

Adds the ability to search for your own posts using the `from:me` search syntax, this is done by replacing instances of `from:me` with account's DID, special care is taken to avoid string replacements inside quotes, because quotes are exact search matches

image

### DIFF
--- a/src/lib/strings/helpers.ts
+++ b/src/lib/strings/helpers.ts
@@ -37,3 +37,27 @@ export function countLines(str: string | undefined): number {
   if (!str) return 0
   return str.match(/\n/g)?.length ?? 0
 }
+
+// Augments search query with additional syntax like `from:me`
+export function augmentSearchQuery(query: string, {did}: {did?: string}) {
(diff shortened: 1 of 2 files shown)


DavidBuchanan314:
By the way, the query parameters are parsed for real as part of the indigo codebase, maybe it would be cleaner to implement the feature there?

https://github.com/bluesky-social/indigo/blob/630059eb1ce90f897527b02612788c05ef95b5e6/search/parse_query.go#L36

I'm not familiar enough with it to say how easy that would be though, obviously the requester's DID needs to be known.

pfrazee (approved):
Really nice. The quoted text handling was a nice touch as well

pfrazee:
I do want to move this into structured query params at some point btw

PR #2525, opened 2024-01-13
Add lang attribute to content of posts

### LINKED ISSUE #1404: Add `lang` attribute on the post content

I'm always frustrated when viewing posts from Japan and China. Because their Chinese characters don't display correctly for me in my Korean environment. Some Simplified / Traditional Chinese characters (hanzi) and Chinese characters in Japan (kanji) are rendered as Chinese characters in Korea (hanja) due to font issues.

This is due to the different shapes of the kanji used by CJK, and the fact that Unicode assigned the same code to characters with no significant difference in shape.

The `lang` attribute must be defined on the post that matches the language of the post content. This will typically render the correct font for each language.

**Additional context**

Screenshot 2023-09-07 at 19-43-06 Quiple (@quiple dev)
Screenshot 2023-09-07 at 19-43-55 Quiple (@quiple dev)

Above is how the same post looks in Korean environment, and below is how it looks with `lang="ja"` applied.

quiple:
Add the lang attribute to the post content. Don't add it if the app language and post language are the same. Resolves #1404.

Thank you!

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -19,7 +19,11 @@ import {sanitizeDisplayName} from 'lib/strings/display-names'
 import {sanitizeHandle} from 'lib/strings/handles'
 import {countLines, pluralize} from 'lib/strings/helpers'
 import {isEmbedByEmbedder} from 'lib/embeds'
-import {getTranslatorLink, isPostInLanguage} from '../../../locale/helpers'
+import {
+  getTranslatorLink,
+  isPostInLanguage,
+  getPostLanguage,
+} from '../../../locale/helpers'
 import {PostMeta} from '../util/PostMeta'
 import {PostEmbeds} from '../util/post-embeds'
 import {PostCtrls} from '../util/post-ctrls/PostCtrls'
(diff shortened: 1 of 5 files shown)


src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -203,6 +209,13 @@ let PostThreadItemLoaded = ({
       ),
     [post, langPrefs.primaryLanguage],
   )
+  const postLang = useMemo(
+    () =>
+      sanitizedLanguage && !isPostInLanguage(post, [sanitizedLanguage])
+        ? getPostLanguage(post)
+        : undefined,
+    [post, sanitizedLanguage],
+  )
```
pfrazee, line 218: Can you explain why you're doing this instead of just putting the post's declared language in the field? If there are multiple languages declared and only one is supported, we can either abort or just take the first

quiple: @pfrazee Sorry, I don't know much about Typescript, and I wrote this by referencing code from elsewhere, so it's going to be hard to explain what I wrote clearly...

If you prefer to implement this feature in a different way, please feel free to close this PR. Thank you!

PR #2534, opened 2024-01-15
Add a new home feed-api wrapper and give a header indicating the fallback behavior

pfrazee:
Currently we're falling back to the "Discover" feed when the following feed comes back empty. We're doing this to patch over weaknesses in our onboarding, which 50% of users skip and therefore land on an empty feed. This is a temporary solution while we work to improve the onboarding; therefore this is a somewhat hacking temporary fix.

|Desktop|Mobile light|Mobile dark|
|-|-|-|
|image|image|image|

### DIFF
--- a/src/lib/api/feed/home.ts
+++ b/src/lib/api/feed/home.ts
@@ -1,0 +1,88 @@
+import {AppBskyFeedDefs} from '@atproto/api'
+import {FeedAPI, FeedAPIResponse} from './types'
+import {FollowingFeedAPI} from './following'
+import {CustomFeedAPI} from './custom'
+import {PROD_DEFAULT_FEED} from '#/lib/constants'
+
+// HACK
+// the feed API does not include any facilities for passing down
+// non-post elements. adding that is a bit of a heavy lift, and we
+// have just one temporary usecase for it: flagging when the home feed
+// falls back to discover.
+// we use this fallback marker post to drive this instead. see Feed.tsx
+// for the usage.
+// -prf
+export const FALLBACK_MARKER_POST: AppBskyFeedDefs.FeedViewPost = {
+  post: {
+    uri: 'fallback-marker-post',
+    cid: 'fake',
+    record: {},
+    author: {
+      did: 'did:fake',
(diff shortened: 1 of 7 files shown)


src/view/com/posts/Feed.tsx
```diff
@@ -265,6 +266,12 @@ let Feed = ({
         )
       } else if (item === LOADING_ITEM) {
         return <PostFeedLoadingPlaceholder />
+      } else if (item.rootUri === 'fallback-marker-post') {
```
estrattonbailey, line 269: Nit: could create a string constant for this to tie the two code locations together

src/view/screens/Home.tsx
```diff
@@ -112,7 +111,7 @@ function HomeScreenReady({
       mergeFeedEnabled: Boolean(preferences.feedViewPrefs.lab_mergeFeedEnabled),
       mergeFeedSources: preferences.feedViewPrefs.lab_mergeFeedEnabled
         ? preferences.feeds.saved
-        : [PROD_DEFAULT_FEED('whats-hot')],
+        : [],
```
estrattonbailey, line 114: Nice

src/lib/api/feed/home.ts
```diff
@@ -0,0 +1,88 @@
...
+
+// HACK
+// the feed API does not include any facilities for passing down
+// non-post elements. adding that is a bit of a heavy lift, and we
+// have just one temporary usecase for it: flagging when the home feed
+// falls back to discover.
+// we use this fallback marker post to drive this instead. see Feed.tsx
+// for the usage.
+// -prf
+export const FALLBACK_MARKER_POST: AppBskyFeedDefs.FeedViewPost = {
```
estrattonbailey, line 15: Oh or just reference like `FALLBACK_MARKER_POST.post.uri`

estrattonbailey (approved):
Works beautifully

src/lib/api/feed/home.ts
pfrazee: good call ya

PR #2536, opened 2024-01-16
Improved search autocomplete view

mary-ext:
Users seems to be getting confused about whether they're currently in the "autocomplete view" or the actual "search view" when trying to search for posts (https://github.com/bluesky-social/social-app/issues/849#issuecomment-1888569647). This pull request improves on the UI.

Not sure about the removal of borders on profiles, thought I'd do it to make the distinction between "Search for ..." and profile suggestions even clearer.

`Go to @handle` is powered by a regex that matches domain-like substrings, when it can find one, then it's shown to the user.

---

| Before | After |
| --- | --- |
| image | image |
| image | image |
|  image | image |

---

| Before | After |
| --- | --- |
| image | image |
| image | image
| image | image |


pfrazee:
Nice! Thanks again Mary!

Issue #684, opened 2023-05-17
Data Saver / Explicit Consent Mode


bnb:
**Is your feature request related to a problem? Please describe.**

I'm currently on an airplane and was scrolling the web app on my Mac. Refreshed the home page and at the top of the page was some pretty, uh, Classic Hellthread Content. I feel exceptionally bad for accidentally exposing the people around me to that and would love to see a way to ensure I (and others!) have complete control of what's shown when scrolling the Good Site.

**Describe the solution you'd like**

I'd specifically like to see a Settings toggle that forces all images behind the blurred image UI that I believe comes up on some content when you've got Content Filtering set to Warn. This would put all images behind the blured UI.

**Describe alternatives you've considered**

- Not Using BlueSky in situations where seeing Classic Hellthread Content isn't appropriate
  - This is not a reasonable solution, as I have ADHD and will forget.
- Toggle to disable images altogether 
  - More aggressive for no benefit 
- Using filtering
  - Tried this, still saw Classic Hellthread Content. No OCR tools will ever be good enough to filter everything. 

**Additional context**

<3

cooljeanius:
> **Describe alternatives you've considered**
>    * Toggle to disable images altogether
>       * More aggressive for no benefit

I'd disagree about the "for no benefit" part here; I think it'd still be worthwhile

PR #2539, opened 2024-01-16
Data saving mode

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/684

Adds data saving mode

I originally wanted to show a button on individual images that would load that specific image thumbnail, ala Twitter, but it seemed like it can get pretty messy with how image embeds are written at the moment, so I took the alternate option which is to show a warning instead.

- Adds a warning when a post contains an image
  image
- Disables thumbnails on external embeds, incl. players
- Adds a toggle on appearance section (not quite sure where else to put)
  image


surfdude29:
Would it be possible to automatically enable data saving mode if the user has turned on Low Data Mode on iOS or Data Saver on Android?

iOS:
https://support.apple.com/102433

Android:
https://source.android.com/docs/core/data/data-saver

surfdude29:
> * It might be cool if we make it actively clear that data saving mode is being used, without the need for there being a post containing an image, perhaps like a fixed bar or something...

As I understand it, enabling data saving mode only affects the loading of images (including thumbnail images for embeds), so I think it's reasonable to only remind the user that it's enabled when it actually changes the experience.

What about changing the message to something like: "Data saving mode is enabled. This post contains 1 image." ?

surfdude29:
Yeah I guess a fixed bar does make more sense.

pfrazee:
Okay cool, I'm going to save this for next release

src/state/persisted/schema.ts
```diff
@@ -54,6 +54,7 @@ export const schema = z.object({
   }),
   hiddenPosts: z.array(z.string()).optional(), // should move to server
   useInAppBrowser: z.boolean().optional(),
+  dataSaverEnabled: z.boolean(),
```
pfrazee, line 57: Needs to be optional or else pre-existing data will fail validation and be discarded

src/view/com/util/post-embeds/DataSaverHider.tsx
```diff
@@ -0,0 +1,153 @@
...
+  // The images are shown
+  SHOWN,
+}
+
+async function isImageCached(uri: string) {
+  if (isWeb) {
+    const img = document.createElement('img')
+    img.src = uri
+
+    const complete = img.complete
```
pfrazee, line 28: TIL! I never knew about this attribute

src/view/com/util/post-embeds/DataSaverHider.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      promises.then(results => {
+        if (aborted) {
+          return
+        }
+
+        const valid = results.every(cached => cached)
+        setState(valid ? State.SHOWN : State.HIDDEN)
+      })
+    }
```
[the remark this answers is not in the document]
pfrazee: Hmm! The post shadow cache is designed to do this kind of thing -- it's a local cache that ties its lifetime to post records. I wonder if we just toss this state onto there.

PR #2544, opened 2024-01-16
Remove excess l10n related markups

tkusano:
There are excess l10n markups which produce meaningless `msgid`/`msgstr` pairs in message catalogue.


quiple:
https://github.com/bluesky-social/social-app/blob/9cbd3c0937d22e8dccbd9c086d3a3a24dbd27b3a/src/Navigation.tsx#L147
It looks like this can be removed as well.

pfrazee (approved):
Awesome, appreciate it!

PR #2548, opened 2024-01-16
Patch `@lingui/core` to fix `unraw` import resolution error

### LINKED ISSUE #2535: Typing A\x41 on search field crashes Bluesky
**Describe the bug**

A CJS-ESM module confusion in how LinguiJS attempts to import the `unraw` module causes Bluesky to crash if a Unicode escape like `\x41` appears in a localized string 

**To Reproduce**

1. Open Bluesky
2. Type `A\x41` in the search field

**Expected behavior**

There shouldn't be any autocompletion results, so it would show the localized string for "No results found..." 

**Screenshots**

`r` is the function from `unraw` module that LinguiJS tried to import, it's the `default` export specifically.

**Details**

 - Platform: Web, any browser and device
 - Platform version: Chrome 120
 - App version: 1.63

**Additional context**

https://bsky.app/profile/mary.my.id/post/3kix2upox6k23

Web-only bug, does not happen in mobile app.

estrattonbailey:
Fixes #2535 

Temp patch, see filed issue in Lingui repo https://github.com/lingui/js-lingui/issues/1836

CleanShot 2024-01-16 at 15 17 39@2x
CleanShot 2024-01-16 at 15 17 45@2x
CleanShot 2024-01-16 at 15 17 49@2x

### DIFF
--- a/patches/@lingui+core+4.5.0.patch
+++ b/patches/@lingui+core+4.5.0.patch
@@ -1,0 +1,10 @@
+diff --git a/node_modules/@lingui/core/dist/index.mjs b/node_modules/@lingui/core/dist/index.mjs
+index 9759736..881f67b 100644
+--- a/node_modules/@lingui/core/dist/index.mjs
++++ b/node_modules/@lingui/core/dist/index.mjs
+@@ -1,4 +1,4 @@
+-import unraw from 'unraw';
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Love that for us

PR #2549, opened 2024-01-16
Reduce some Sentry logs

estrattonbailey:
This PR stops sending 3 logs to Sentry as top-level issues (reduces to `info` level, sends as breadcrumbs which should be batched), and reduces one log from `error` to `warn` for monitoring.


pfrazee (approved):
sg

pfrazee:
Oh unit tests are relying on some of those logs

PR #2557, opened 2024-01-17
new Web rich text composer

mary-ext:
this is an experimental pull request that introduces a new RichText input for web, with the goal of aligning with what's currently on native. This has the benefit of side-stepping any headaches that currently comes with TipTap and other WYSIWYG editors like Lexical, plus, makes a decent dent on reducing bundle size.

this RichText input works pretty similarly to how it is on native, it does this by making the input's text color transparent (and fixing up any adjacent styling like caret, placeholder and selection text), and overlaying our own rendered text below it (such that your cursor will always land on the input)

- [x] Basic text input functionality
- [x] Text highlighting
- [x] Autocomplete suggestions
  - This is entirely new code, should be able to align it with native, but native doesn't handle selection changes
- [x] Emoji picker (broke range positioning, should be able to make it work again)
- [x] Paste to add images
- [x] Keyboard shortcuts (Ctrl/Meta+Enter)

I know that Paul has said that there might be plans to move away from being entirely driven by syntax, but this was just a fun challenge I came up on a whim, feel free to close this PR if necessary :)

TODO

- [x] Don't pull in ReactDOM for portals, I didn't realize we already have a portal solution going
- [ ] Figure out the performance issue, see below comments
- [ ] Check if ligatures could pose a problem, is turning them off an option?
- [ ] Test non-Latin usage


pfrazee:
Intriguing. I'll definitely want to play with this at some point.

ocodista:
Could you add some screenshots or videos of what it would look like?

mary-ext:
It's intended to look exactly the same as before, so there is no screenshot or video.

PR #2564, opened 2024-01-18
Phone number verification in account creation

pfrazee:
Adds phone-number verification to account creation driven by server config. Is *not* 2FA. Still needs the api package to get bumped so won't run yet.

### DIFF
--- a/src/view/com/auth/create/CreateAccount.tsx
+++ b/src/view/com/auth/create/CreateAccount.tsx
@@ -117,7 +117,7 @@ export function CreateAccount({onPressBack}: {onPressBack: () => void}) {
 
   return (
     <LoggedOutLayout
-      leadin={`Step ${uiState.step}`}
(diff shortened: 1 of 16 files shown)


src/view/com/auth/create/CreateAccount.tsx
```diff
@@ -117,7 +118,7 @@ export function CreateAccount({onPressBack}: {onPressBack: () => void}) {
 
   return (
     <LoggedOutLayout
-      leadin={`Step ${uiState.step}`}
+      leadin=""
```
estrattonbailey, line 121: 👀

estrattonbailey (approved):
LGTM

Nit: verification code input could strip bad characters 🤔

Zero3K:
Why was this added? There are people that hate it from what I have seen on Reddit, etc.

PR #2566, opened 2024-01-19
Use new palette values for main palette

estrattonbailey:
Only swaps colors that will be very obviously different alongside new components.


pfrazee (approved):
Yep that's nice

haileyok:
im guessing this is not intended?

Screenshot 2024-01-19 at 1 28 25 AM

estrattonbailey:
@haileyok hmmmm (oh geez I hope I'm not missing something obvious), does something look off there?

haileyok:
@estrattonbailey i didn't look closely but I'm betting the scroll view and the scroll view content container styles don't match up. Likely that the view itself's background is still solid black.

PR #2569, opened 2024-01-19
flex and truncate content hider text

haileyok:
IMG_4754

Screenshot 2024-01-19 at 1 40 01 AM

caps line count to two. could go to one, but better to err on side of showing more context


pfrazee (approved):
Oh nice, gratzi

PR #2570, opened 2024-01-19
properly shorten links in quote embeds

haileyok:
if the user doesn't shorten links before posting, quotes look janky

Screenshot 2024-01-19 at 1 59 35 AM

instead, we can pass the facets into the quote, and add a prop to `RichText` that can optionally disable links, but still deal with shortening them.

Screenshot 2024-01-19 at 2 04 34 AM


pfrazee (approved):
Cool nice work

PR #2576, opened 2024-01-19
fix: add onContentSizeChange scrollToEnd in Composer component

### LINKED ISSUE #2413: Input field does not scroll with content iOS
**Describe the bug**
When replying to a post, the input field where I enter text does not scroll as the text exceeds the length of the viewport, and must be scrolled manually over and over. With a long enough post, it often cannot be scrolled to the end at all, or appears to get “stuck” and has to be swiped over and over to cause it to actually scroll. 

**To Reproduce**

1. Reply to a post
2. Type in text longer than the viewport

**Expected behavior**

For the viewport to scroll to where the insert cursor is

**Screenshots**

**Details**

 - Platform: iOS
 - Platform version: 16.7.2
 - App version: 1.62.0, has been happening since I installed the app four months ago 

Uploading RPReplay_Final1704307218.mov…

trantus:
fixes #2413 

| iOS before | iOS after |
| ------- | --- |
| iOS before | iOS after |

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -548,7 +548,7 @@ const styles = StyleSheet.create({
     marginRight: 5,
   },
   scrollView: {
-    // flex: 1,
+    flex: 1,
     paddingHorizontal: 15,
   },
   textInputLayout: {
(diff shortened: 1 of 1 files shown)


pfrazee:
I appreciate this PR! Unfortunately this doesn't work because if you're editing the middle of a long post, it will cause the scroll to happen inappropriately

However! While testing I noticed something which I think is the root of the issue, and opened a separate PR here: https://github.com/bluesky-social/social-app/pull/2603

Thank you again!

Issue #2556, opened 2024-01-17
Cap external embed description at 2 lines


mozzius:
**Is your feature request related to a problem? Please describe.**

Link cards are very tall. The primary issue is #2555, but it's confounded by the description being capped at 4 lines. This text is more often than not, not very helpful - usually you can get the gist from the first 2 lines, and at current it's mostly clutter. I'd much prefer to see it capped to only 1 or 2 lines (exception being if there's no thumbnail, in which case I suppose you can keep the description as is since you've got more space)

**Describe the solution you'd like**

Cap the description to 2 lines if there's a thumbnail, 4 lines if not.

Issue #2555, opened 2024-01-17
Use a 1.91:1 aspect ratio for external embed thumbnails


mozzius:
**Is your feature request related to a problem? Please describe.**

The app uses an unusally tall aspect ratio for displaying external embed thumbnails. It seems like there's a loose consensus around the 1.91:1 ratio that Facebook uses (1200x630) - this means that a lot of opengraph images get cropped, especially ones with text on them. See this link to github:

<img width="520" alt="Screenshot 2024-01-17 at 10 26 48" src="https://github.com/bluesky-social/social-app/assets/10959775/45dcc8c2-7766-4e41-9ed0-b4ae4f72f942">

Reducing the vertical height of the embeds would also be good in general, as the link embeds take up a lot of space.

**Describe the solution you'd like**

Set the aspect ratio of the link embed thumbnail to 1.91:1

**Additional context**

I did this for Graysky and I see a lot fewer clipping issues

PR #2577, opened 2024-01-19
improve external link design

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2556
fixes https://github.com/bluesky-social/social-app/issues/2555

Improves the sizing of external embeds so they don't take up so much screen space.

- Reduce thumb size on mobile from 200 to 180
- Reduce thumb size on web from 300 to 250
- Reduce number of description lines from 4 to 2 unless there is no thumbnail
- Cleanup unnecessary styles and move everything to the stylesheet

https://github.com/bluesky-social/social-app/assets/153161762/dec24810-ae9e-4715-bd8f-10186c822e1b

### DIFF
--- a/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
+++ b/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
@@ -32,7 +32,7 @@ export const ExternalLinkEmbed = ({
     <View style={styles.container}>
       {link.thumb && !embedPlayerParams ? (
         <Image
-          style={{height: isMobile ? 150 : 250, width: '100%'}}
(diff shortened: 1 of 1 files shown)


pfrazee:
@haileyok wdyt about moving to an aspect ratio for the thumbs? we could try to match what the existing platforms have established

haileyok:
@pfrazee yea just asked @mozzius about that, sounds like 1.91 is the way to go. looks better on mobile and web that way, good call

Screenshot 2024-01-19 at 3 13 11 PM
Screenshot 2024-01-19 at 3 15 17 PM

pfrazee:
wunderbar! thanks @haileyok and @mozzius

pfrazee (approved):

PR #2578, opened 2024-01-19
Fix orphan replies in linear mode

gaearon:
The way we show replies in a linear mode isn't quite right IMO.

Consider this thread: http://bsky.app/profile/danabra.mov/post/3kjemen7ies27

Opening the root post shows replies to further posts on the same level:

Screenshot 2024-01-19 at 23 51 40

This breaks the flow of conversation. You see replies out of the context to which they were replying — which is not just confusing but can also cause miscommunication (like "I agree" replying to a different post in the thread).

Instead, I think this should work like on Twitter. You only see the reply chains that are *for* the post you currently highlighted.

So on the root post I should see just this:

Screenshot 2024-01-19 at 23 54 21

For the "post 1" post, I should see this:

Screenshot 2024-01-19 at 23 54 51

And so on.

This only affects linear mode. In threaded mode, all replies are still shown.

Screenshot 2024-01-19 at 23 55 45

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -157,7 +157,11 @@ function PostThreadLoaded({
   const posts = React.useMemo(() => {
     let arr = [TOP_COMPONENT].concat(
       Array.from(
-        flattenThreadSkeleton(sortThread(thread, threadViewPrefs), hasSession),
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Yep, this is a huge improvement to linear mode. Muchos gracias

PR #2580, opened 2024-01-20
use theme's color scheme in in-app browser

haileyok:
instead of using the system's color scheme, use the user's selected color scheme for the in-app browser

RocketSim_Recording_iPhone_15_Pro_6 1_2024-01-19_16 17 05


pfrazee (approved):
awesome

PR #2581, opened 2024-01-20
Debounce language suggestion

mary-ext:
using requestIdleCallback might've been a bad idea for this, let's properly debounce it so it only runs when the user finishes typing

later on I'll try tuning the neural network directly and hopefully we can improve on the suggestions then

### DIFF
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -18,10 +18,6 @@ import {
   FontAwesomeIconStyle,
 } from '@fortawesome/react-native-fontawesome'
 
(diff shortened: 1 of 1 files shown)


pfrazee:
> having it always clear might be annoying though, let me do some more testing...
> edit: yeah it's bad

okay so I figure what you were aiming for was to trigger the detection after a small delay from typing, but now what this means is that it also _clears_ after a small delay, and that is kind of crummy too? If so then we'd need to debounce _only_ if a recommendation isn't made yet. Think that's worth bothering?

pfrazee:
up to you. this doesn't really bother me that hard

mary-ext:
Closing for now

PR #2582, opened 2024-01-20
Small language detection fixes

mary-ext:
- Fix broken Chinese language detection
- Make sure that post language detection is skipping over uncertain languages
- ~~Only suggest languages that are also set in the user's content languages preference, it's a bit unfortunate, but it seems like the only best option given how much false positives it's giving off~~

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -48,44 +48,40 @@ export function codeToLanguageName(lang: string): string {
 export function getPostLanguage(
   post: AppBskyFeedDefs.PostView,
 ): string | undefined {
+  const record = post.record
+
   let candidates: string[] = []
-  let postText: string = ''
-  if (hasProp(post.record, 'text') && typeof post.record.text === 'string') {
-    postText = post.record.text
+  let text: string = ''
+
+  if (AppBskyFeedPost.isRecord(record)) {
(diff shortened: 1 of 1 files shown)


pfrazee:
> Only suggest languages that are also set in the user's content languages preference, it's a bit unfortunate, but it seems like the only best option given how much false positives it's giving off

not a bad solution, but I feel like it might kneecap this pretty hard

some ideas:

- we could try subtracting from the score by, like, 0.1 if it's not one of their configured languages
- the model reacts differently to strings at various lengths. one thing we could try is to average the scores from the string at full length, len-1, len-2
- I noticed the scores tend to weight differently between languages. for instance, it seemed to cap at (iirc) 0.97 for korean, but go to 0.99 for english. we could test against some different languages and set different thresholds for each one if I have this right

pfrazee:
@mary-ext btw on PRs you contribute, if I ever have a round of feedback and you want to just leave it for us to do, always feel free to say so

pfrazee:
> haven't got a lot of time to look on this just yet, though I'd be okay with passing this and https://github.com/bluesky-social/social-app/pull/2581 off to y'all ^^

okay I'll consider these in our court! if you feel like picking them up again feel free as well; my main concern is us clobbering your work, not the other way around

mary-ext:
Closing for now

Issue #1257, opened 2023-08-23
Support for drag & drop of images in the compose dialog


mackuba:
**Is your feature request related to a problem? Please describe.**

When composing a post in the web app in a desktop browser, it would be nice to be able to drag & drop an image from Finder into the window to attach it. This works on Twitter and Mastodon, and is generally a pattern used everywhere where images can be uploaded.

Bluesky right now does not handle this in JS, so dragging an image to the app causes the browser to just open the image on the whole page, replacing the Bluesky webapp, possibly losing the entered post text.

**Describe the solution you'd like**

Mastodon:

<p><img width="600" src="https://github.com/bluesky-social/social-app/assets/28465/67b38b25-4f47-49ba-8e29-74230cad73cd"></p>

Twitter:

<p><img width="600" src="https://github.com/bluesky-social/social-app/assets/28465/97bdc93f-3785-4909-bbed-024746e4e641"></p>

PR #2583, opened 2024-01-20
Image drag and drop on web composer

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/1257

Adds the ability to insert images by drag-and-drop.
The drop trigger covers the whole screen so you don't have to go very far to add the image.

The overlay modal has a box-shadow to compensate for reduced opacity on the backdrop, because having the backdrop opacity be the same as the composer modal seems to make it *too* dark.

image
image

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -111,24 +111,32 @@ export const TextInput = React.forwardRef(function TextInputImpl(
       event.preventDefault()
       setIsDropping(false)
     }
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Fantastic! Works perfectly

pfrazee:
>Is there a reason behind the text/plain handling? I'm a bit confused because I don't remember having to deal with that sort of thing

I forget exactly why that got done -- I seem to remember there's some flow that gets recognized as a text-link getting pasted. But it does mean that pasting an image URI gets added automatically as an image if it can.

Issue #2574, opened 2024-01-19
Logged-out visibility toggle not applying for <title> HTML tag


gildaswise:
**Describe the bug**
If you have the Logged-out visibility toggle on, the bsky.app web version is still showing the contents of your post in the <title> HTML tag even if the UI is hiding it;

**To Reproduce**

1. Toggle Logged-out visibility on
2. Check your own post in an incognito tab

**Expected behavior**

All post content should be hidden

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/1339236/2b76d320-04fb-4dce-a1a4-7be0215cc65b)

![image](https://github.com/bluesky-social/social-app/assets/1339236/33456966-95f2-4edb-b127-5f9eff2c357c)

**Details**

 - Platform: Web
 - Platform version: 1.65

**Additional context**

Reported by [@no381.bsky.social ](https://bsky.app/profile/no381.bsky.social/post/3kje5225j222b)

pfrazee:
Oh good catch, thank you. We'll get this fixed.

PR #2584, opened 2024-01-20
Don't expose post content on no-unauthenticated

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/2574


pfrazee (approved):
Excellent, thank you

PR #2585, opened 2024-01-20
use ISO 3166 names for UK and Virgin Islands (U.S.)

surfdude29:
(this is my first-ever PR so I apologise if I got something wrong)

#2579 added a list of alpha-2 codes and country names according to ISO 3166 (as well as their international dialling codes). All were correct except GB and VI.

GB should be [United Kingdom of Great Britain and Northern Ireland](https://www.iso.org/obp/ui/#iso:code:3166:GB)

VI should be [Virgin Islands (U.S.)](https://www.iso.org/obp/ui/#iso:code:3166:VI)

This small PR amends these names.

If the full name for the UK is too long for the UI, it can simply be shortened to 'United Kingdom'.


pfrazee (approved):
Awesome, thank you!

surfdude29:
Oops, I think my PR might be causing the linter to throw an error because the line is too long? But I saw that @quiple has a patch as part of #2502 that looks like it would fix it

PR #2586, opened 2024-01-20
Ensure OG:Image is filled when post contains a record

Gargaj:
Posts that contain images AND a record (e.g. [quoted posts](https://bsky.app/profile/amagire.bsky.social/post/3kjgfxufwfn2z)) currently don't generate any `og:image` tags; this - in theory - fixes that.

### DIFF
--- a/bskyweb/cmd/bskyweb/server.go
+++ b/bskyweb/cmd/bskyweb/server.go
@@ -336,12 +336,20 @@ func (srv *Server) WebPost(c echo.Context) error {
 	postView := tpv.Thread.FeedDefs_ThreadViewPost.Post
 	data["postView"] = postView
 	data["requestURI"] = fmt.Sprintf("https://%s%s", req.Host, req.URL.Path)
-	if postView.Embed != nil && postView.Embed.EmbedImages_View != nil {
-		var thumbUrls []string
-		for i := range postView.Embed.EmbedImages_View.Images {
-			thumbUrls = append(thumbUrls, postView.Embed.EmbedImages_View.Images[i].Thumb)
+	if postView.Embed != nil {
+		if postView.Embed.EmbedImages_View != nil {
+			var thumbUrls []string
+			for i := range postView.Embed.EmbedImages_View.Images {
+				thumbUrls = append(thumbUrls, postView.Embed.EmbedImages_View.Images[i].Thumb)
+			}
+			data["imgThumbUrls"] = thumbUrls
+		} else if postView.Embed.EmbedRecordWithMedia_View != nil && postView.Embed.EmbedRecordWithMedia_View.Media != nil && postView.Embed.EmbedRecordWithMedia_View.Media.EmbedImages_View != nil {
+			var thumbUrls []string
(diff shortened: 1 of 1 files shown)


bskyweb/cmd/bskyweb/server.go
```diff
@@ -336,12 +336,20 @@ func (srv *Server) WebPost(c echo.Context) error {
 	postView := tpv.Thread.FeedDefs_ThreadViewPost.Post
 	data["postView"] = postView
 	data["requestURI"] = fmt.Sprintf("https://%s%s", req.Host, req.URL.Path)
-	if postView.Embed != nil && postView.Embed.EmbedImages_View != nil {
-		var thumbUrls []string
-		for i := range postView.Embed.EmbedImages_View.Images {
-			thumbUrls = append(thumbUrls, postView.Embed.EmbedImages_View.Images[i].Thumb)
+	if postView.Embed != nil {
+		if postView.Embed.EmbedImages_View != nil {
+			var thumbUrls []string
```
Gargaj, line 341: This could be lifted out to be only declared once outside the nested `if`, but I wasn't sure how the template engine would react.

pfrazee:
@bnewbold have to pass to you for review

bskyweb/cmd/bskyweb/server.go
bnewbold: the variable could be defined outside, it is only the assignment in to `data` that will impact the template engine.

bskyweb/cmd/bskyweb/server.go
```diff
@@ -336,12 +336,20 @@ func (srv *Server) WebPost(c echo.Context) error {
...
+			for i := range postView.Embed.EmbedImages_View.Images {
```
bnewbold, line 342: you kept it the way it was before, so this is fine, but the code would be more compact if this iterator was `for _, img := range postView.Embed.EmbedImages_View.Images`, so the long variable names don't need to be indexed below

bnewbold (approved):
looks good! this is Ok to merge and deploy as-is. if I get time I might take a small second pass to make the code more compact, but it is fine and safe as-is.

thanks for the contribution! good to be hunting down these "embed record with media".

bnewbold:
lint is failing, but that should get fixed with squash merge

bskyweb/cmd/bskyweb/server.go
Gargaj: Yes but that would require me to actually know Go which I decidedly don't, I just looked at the code and sorta inferred how the language works 😁

bnewbold: all the more impressed!

Gargaj:
@bnewbold Just out of curiosity, when do these go live? Is that done with certain regularity or is it a versioned manual deploy?

bnewbold:
they tend to go out weekly, coordinated with mobile app deploys. you can follow https://bsky.app/profile/bsky.app for announcement when a versioned release happens; the last one was 3 days ago, just before this merged.

Issue #1540, opened 2023-09-26
[APP-870] Compose intent URL


pfrazee:
Add a URL for triggering the composer with some pre-filled fields. This is useful for websites that want to do share links/buttons.

See https://github.com/bluesky-social/atproto/discussions/1671

<sub>[APP-870](https://linear.app/blueskyweb/issue/APP-870/compose-intent-url)</sub>

ShreyanJain9:
When you do this, I would suggest figuring out some way to make it client-agnostic, if possible. Might not be easy, but I think you can do it!

pfrazee:
Copying from dup:

**Is your feature request related to a problem? Please describe.**

Currently users wanting to share off-site links directly from the page need to do a variety of URL copy/paste actions to get a link into the Bluesky editing interface. Further, offsite content creators who want to "pre-fill" the sharing post's text are unable to do so. This makes it impossible to create off-site sharing buttons, such as you'd have for every other social media.

**Describe the solution you'd like**

Two APIs:
* `/share?url=<url>&prefill=<prefill text>&origin=<handle of account linked to shared content>`
  * User journey: _When a user clicks a share button outside of Bluesky, she is taken to a new post dialog. If the optional `prefill` query parameter has a value, that text precedes the string given via the required `url` query parameter. If the optional `origin` query parameter has a value, a translated version of a string in the format ", via @{handle}." is appended to the text._
* `/repost?did=<did>&prefill=<prefill text>[&allow_quote]`
  * User journey: _A user clicks a link in this format on an off-site page, which opens the related DID in the Bluesky interface, possibly without any attached replies. If the originating website included the optional `allow_quote` query parameter, the "Repost"/"Quote post" menu is already unfurled, allowing the user to select either option. If "Quote post" is selected by the user, the quote post editor is opened; if the `prefill` query parameter is included, the editor is already populated with the text from that. If the "enable_quote" query parameter is omitted, the user is given the option to repost the linked DID without a quote. Of course if the user really wants to quote post a link, she's still able to copy/paste the URL into a new post._

**Describe alternatives you've considered**

Prior art:
* X: `/intent/tweet?url=<url>&text=<prefill>&via=<handle>`
* Facebook: `/sharer.php?u=<url>&t=<prefill>`
* LinkedIn: `/sharing/share-offsite/?url=<url>`

Of these, X's implementation fulfils the needs of media orgs the best by allowing both prefill text as well as the ability for a shared post be linked to the org's canonical user.

[...]

Issue #66, opened 2023-01-20
[APP-482] Add Bluesky as a "share to" target


pfrazee:
null

<sub>[APP-482](https://linear.app/blueskyweb/issue/APP-482/add-bluesky-as-a-share-to-target)</sub>

revanmj:
How is it completed when Android app does not have such option. Tried with sharing text or images from other app, in both cases Bluesky did not show up as a target.

pfrazee:
This isn't completed, it just got moved to the internal tracker when we were adding that

Splunger:
> This isn't completed, it just got moved to the internal tracker when we were adding that

Any updates? This is a pretty critical feature.

pfrazee:
Internal tracker didn't work out, reopening

jhwheuer:
So a year and no resolution. Facepalm.

Splunger:
Open to the public but still no fix to quick share in Android. I'm a little confused by these tactics.

jhwheuer:
Nothing confusing. Somebody's bonus depended on going GA, and did not on being feature-complete.

MarkBennett:
Does anyone know of a way to add this? I'm a dev with lots of react experience but none with React Native.

Would be happy to make a PR if the team is open and someone could point me in the general direction of how best to start implementing this.

MarkBennett:
Just did a quick search and React Native has a Linking API:

https://reactnative.dev/docs/linking

I assume implementing this would work? Thoughts @pfrazee?

haileyok:
@MarkBennett Thank you! We actually have this in a branch right now but just need some time to review it. Hoping we get it done soon. https://github.com/bluesky-social/social-app/pull/2587

jhwheuer:
Not to sound alarmist.... I stopped using Blue sky (no not BS, bad acronym) because sharing is simply painful

MarkBennett:
> @MarkBennett Thank you! We actually have this in a branch right now but just need some time to review it and hash out a few things. Hoping we get it done soon. #2587

Thanks @pfrazee! Will definitely make sharing easier once that's enabled. 😄

PR #2587, opened 2024-01-21
Share Extension/Intents

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/66
fixes https://github.com/bluesky-social/social-app/issues/1540

## What I've tested
- [x] iOS
	- [x] Share from Photos
	- [x] Share screenshot
	- [x] Share URL from Safari
	- [x] Share text from selection
	- [x] Share note (uses the note's text)    
- [x] Android
	- [x] Share from Photos
	- [x] Share URL from Chrome
	- [x] Share text from selection  
	- [x] Platforms 23, 25, 28, and 33 
- [x] Web
	- [x] Share text with `?compose=true&text=`

Still to do
- [x] Probably should remove the copies of the images after everything is done (we have to wait until the post is actually made or discarded unfortunately otherwise we won't have access to the original image) - Later note: I'll handle this in a separate PR since we're migrating away from `react-native-fs` to `expo-file-system` and the logic will need to be different anyway.

https://github.com/bluesky-social/social-app/assets/153161762/e12527b4-bc26-4891-bbf5-ceaef3ecab15

# Intent

To open the composer, user navigates to `https://bsky.app/intent/compose?text=Hello world!`. The existing router logic handles passing those params to the home screen.

# Native scheme

The same logic is used on native, allowing for `bluesky://intent/compose?text=Hello world!`.

In addition, image URIs from the extension are also allowed. These are passed in as `imageUris?=uri1,uri2,uri3,uri4`. Currently, only local URIs are supported.

# Native share extension/module

On iOS we add an extension to the project's build phases. On Android, we add an expo module that handles the incoming intents for us and reroutes them to the correct scheme.

There's also a little script, `updateExtensions.sh` that updates the tracked native files in `/modules/Share-with-Bluesky` with the ones in `/ios/Share-with-Bluesky`, for development purposes, since unlike a module, the extension code needs to live inside of the `ios` directory.

# Config plugin

In an effort to support the current prebuild configuration (ffs...🙃), an expo plugin handles moving the native code to the correct directories and adding the necessary build phases, entitlements, etc. to the projects. Config-defined schemes are supported, for those who may be doing their own testing in an effort to not overlap with the official build's scheme.

The Android build plugin just updates the `AppManifest.xml` with the necessary allowed intents.


extensions/Share-with-Bluesky/ShareViewController.swift
```diff
@@ -0,0 +1,146 @@
...
+      self.completeRequest()
+    }
+  }
+
+  private func handleImages(items: [NSItemProvider]) async -> Void {
+    // Verify that they are all images and that there are not more than four, just in case
+    guard items.count <= 4 else {
+      self.completeRequest()
+      return
+    }
```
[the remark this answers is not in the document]
haileyok: That guard was actually unnecessary, the extension doesn't show up based off the plist setting. But yea, I think just accepting however many and taking the first four is better. Android doesn't let us select a max anyway so that's the logic we need there.

pfrazee:
good gravy those build plugins

haileyok:
tfw writing the feature takes an hour and writing getting it to build takes a day

haileyok:
Feeling like this is probably ready to test out. I've tested the scenarios I can think of and they have all worked so far...

haileyok:
Changing this to draft because I want to check over one thing I noticed (sharing an image that isn't in the photo gallery but rather from Safari or something) and I want to see what we're doing for intent URIs.

haileyok:
Okay back to work.

scripts/updateExtensions.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+IOS_SHARE_EXTENSION_DIRECTORY="./ios/Share-with-Bluesky"
+MODULES_DIRECTORY="./modules"
+
+if [ ! -d $IOS_SHARE_EXTENSION_DIRECTORY ]; then
+  echo "$IOS_SHARE_EXTENSION_DIRECTORY not found inside of your iOS project."
+  exit 1
+else
+  cp -R $IOS_SHARE_EXTENSION_DIRECTORY $MODULES_DIRECTORY
+fi
```
haileyok, line 10: This is just a helper script for updating the extension files inside of the `./modules` directory with changes made in Xcode while dev/testing the extension.

src/lib/hooks/useIntentHandler.ts
```diff
@@ -45,18 +47,39 @@ function useComposeIntent() {
...
 
+      const imageUris = imageUrisStr
+        ?.split(',')
+        .filter(part => {
+          // For some security, we're going to filter out any image uri that is external. We don't want someone to
+          // be able to provide some link like "bluesky://intent/compose?imageUris=https://IHaveYourIpNow.com/image.jpeg
+          // and we load that image
+          if (part.includes('https://') || part.includes('http://')) {
+            return false
+          }
```
haileyok, line 66: I think this covers the main security concern with this. Any others I'm not considering?

src/lib/hooks/useIntentHandler.ts
```diff
@@ -45,18 +47,39 @@ function useComposeIntent() {
...
+        .map(part => {
+          const [uri, width, height] = part.split('|')
+          return {uri, width: Number(width), height: Number(height)}
+        })
+
       setTimeout(() => {
-        openComposer({}) // will pass in values to the composer here in the share extension
+        openComposer({
+          text: text ?? undefined,
+          imageUris: isNative ? imageUris : undefined,
```
haileyok, line 81: No support for `imageUris` on web for right now. At some point if we do allow external URLs through cardyb or something, we could do this.

pfrazee (approved):
This is fantastic work Hailey. Narrowing it all down into the intent URL is very smart. Tested on simulators w/all platforms and felt great. So excited for this.

PR #2596, opened 2024-01-23
New Onboarding

estrattonbailey:
- [x] Handle initial request failure
- [x] Potentially separate global onboarding state
- [x] Metrics
- [x] Handle links in feed descriptions
- [x] Preserve suggestions between steps
- [x] Prevent layout shift toggling adult content


src/components/Button.tsx
```diff
@@ -40,13 +42,18 @@ export type VariantProps = {
    * The size of the button
    */
   size?: ButtonSize
+  /**
+   * The shape of the button
+   */
+  shape?: ButtonShape
```
estrattonbailey, line 48: We have round and square buttons now

src/components/Button.tsx
```diff
@@ -490,14 +511,24 @@ export function ButtonText({children, style, ...rest}: ButtonTextProps) {
 
 export function ButtonIcon({
   icon: Comp,
+  position,
 }: {
   icon: React.ComponentType<SVGIconProps>
+  position?: 'left' | 'right'
```
estrattonbailey, line 517: Helps with positioning, since artboards for icons contain some whitespace.

src/components/Link.tsx
```diff
@@ -24,66 +23,62 @@ import {
...
- * A interactive element that renders as a `<a>` tag on the web. On mobile it
- * will translate the `href` to navigator screens and params and dispatch a
- * navigation action.
- *
- * Intended to behave as a web anchor tag. For more complex routing, use a
- * `Button`.
- */
-export function Link({
-  children,
+export function useLink({
```
estrattonbailey, line 51: Did some work in this file to differentiate between inline links and other custom links. This hook powers both, and can be used elsewhere for the same effect.

src/components/Portal.tsx
```diff
@@ -12,45 +12,54 @@ type ComponentMap = {
...
+    return null
+  }
+
+  return {Provider, Outlet, Portal}
 }
+
+const DefaultPortal = createPortalGroup()
+export const Provider = DefaultPortal.Provider
+export const Outlet = DefaultPortal.Outlet
+export const Portal = DefaultPortal.Portal
```
estrattonbailey, line 65: Default portal here, but can be used for layouts to create "slots"

src/components/RichText.tsx
```diff
@@ -0,0 +1,126 @@
...
+import {getAgent} from '#/state/session'
+
+const WORD_WRAP = {wordWrap: 1}
+
+export function RichText({
+  testID,
+  value,
+  style,
+  numberOfLines,
+  disableLinks,
```
estrattonbailey, line 17: I felt that `noLinks` wasn't as clear as `disableLinks` 🤔

src/components/RichText.tsx
```diff
@@ -0,0 +1,126 @@
...
+    if (
+      mention &&
+      AppBskyRichtextFacet.validateMention(mention).success &&
+      !disableLinks
+    ) {
+      els.push(
+        <InlineLink
+          key={key}
+          to={`/profile/${mention.did}`}
+          style={[a.leading_normal, style, {pointerEvents: 'auto'}]}
```
estrattonbailey, line 88: Main features of this re-implementation are support line height styles more generically

src/components/Typography.tsx
```diff
@@ -110,15 +179,16 @@ export function P({style, ...rest}: TextProps) {
...
-    (_style?.lineHeight || atoms.text_md.lineHeight) *
-    atoms.leading_normal.lineHeight
   return (
     <RNText
       {...attr}
       {...rest}
-      style={[atoms.text_md, t.atoms.text, _style, {lineHeight}]}
+      style={normalizeTextStyles([
+        atoms.text_md,
+        atoms.leading_normal,
```
estrattonbailey, line 188: Only the `P` comp has leading applied, all others default to a relative value of 1.0

src/components/forms/TextField.tsx
```diff
@@ -208,7 +208,7 @@ export function createInput(Component: typeof TextInput) {
               paddingBottom: 2,
             }),
             {
-              lineHeight: a.text_md.lineHeight * 1.1875,
+              lineHeight: a.text_md.fontSize * 1.1875,
```
estrattonbailey, line 211: Removed `lineHeight` from atoms, so needed to use `fontSize` (same value) for this hack in inputs

src/lib/build-flags.ts
```diff
@@ -1,2 +1,4 @@
 export const LOGIN_INCLUDE_DEV_SERVERS = true
 export const PWI_ENABLED = true
+// TODO
+export const NEW_ONBOARDING_ENABLED = true
```
estrattonbailey, line 4: TODO decide how we want to ship this

src/view/screens/Storybook/Typography.tsx
```diff
@@ -25,6 +26,14 @@ export function Typography() {
       <Text style={[a.text_sm]}>atoms.text_sm</Text>
       <Text style={[a.text_xs]}>atoms.text_xs</Text>
       <Text style={[a.text_2xs]}>atoms.text_2xs</Text>
+
+      <RichText
+        value={`This is rich text. It can have mentions like @bsky.app or links like https://blueskyweb.xyz`}
+      />
+      <RichText
+        value={`This is rich text. It can have mentions like @bsky.app or links like https://blueskyweb.xyz`}
+        style={[a.text_xl]}
```
estrattonbailey, line 35: Text size or leading applies as normal here now

src/components/RichText.tsx
```diff
@@ -0,0 +1,126 @@
...
+  disableLinks,
+}: TextStyleProp & {
+  value: RichTextAPI | string
+  testID?: string
+  numberOfLines?: number
+  disableLinks?: boolean
+}) {
+  const detected = React.useRef(false)
+  const [richText, setRichText] = React.useState<RichTextAPI>(
+    value instanceof RichTextAPI ? value : new RichTextAPI({text: value}),
```
gaearon, line 26: Let's do `useState(() => ...)` to avoid re-running the code inside (like the constructor) needlessly

src/components/RichText.tsx
```diff
@@ -0,0 +1,126 @@
...
+    if (!detected.current) {
+      detected.current = true
+      detectFacets()
+    }
+  }, [richText, setRichText])
+
+  const {text, facets} = richText
+
+  if (!facets?.length) {
+    if (/^\p{Extended_Pictographic}+$/u.test(text) && text.length <= 5) {
```
gaearon, line 45: Let's do `text.length <= 5` check first before the regex check?

gaearon (commented):
small nits

src/components/RichText.tsx
```diff
@@ -0,0 +1,126 @@
...
+      const rt = new RichTextAPI({text: richText.text})
+      await rt.detectFacets(getAgent())
+      setRichText(rt)
+    }
+
+    if (!detected.current) {
+      detected.current = true
+      detectFacets()
+    }
+  }, [richText, setRichText])
```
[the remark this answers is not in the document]
pfrazee: Yeah this needs to be disabled by default and only enabled by a prop. Runtime resolution breaks people changing their handles after initial publish.

src/lib/build-flags.ts
pfrazee: _probably_ won't need this but let's make sure our server routes get shipped before we remove it

src/components/RichText.tsx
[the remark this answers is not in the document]
estrattonbailey: Word, fixed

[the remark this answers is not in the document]
estrattonbailey: Will need to do selectable too, but that's a larger lift that needs to touch the new type components

src/components/RichText.tsx
estrattonbailey: Ah good call!

pfrazee (approved):
Fantastic work all

PR #2589, opened 2024-01-22
basic implementation of search history

skiniks:
This pull request introduces a search history feature to the application, limiting the history to the 5 most recent searches. This enhancement tracks and displays the user's latest searches, automatically removing the oldest entry when a new search is conducted. The history is displayed under the search bar, allowing for easy repetition of previous searches. Duplicate searches are intelligently handled by bringing the existing query to the top, avoiding redundant entries. This feature, integrated with AsyncStorage for persistence, is designed to blend seamlessly into the existing UI, offering an intuitive and improved user experience.

PR #2597, opened 2024-01-23
Add local search history (builds on #2589)

pfrazee:
Supersedes #2589

From original PR by @skiniks

> This pull request introduces a search history feature to the application, limiting the history to the 5 most recent searches. This enhancement tracks and displays the user's latest searches, automatically removing the oldest entry when a new search is conducted. The history is displayed under the search bar, allowing for easy repetition of previous searches. Duplicate searches are intelligently handled by bringing the existing query to the top, avoiding redundant entries. This feature, integrated with AsyncStorage for persistence, is designed to blend seamlessly into the existing UI, offering an intuitive and improved user experience.

Changes added:

- Put most recent searches on top
- Only show search history when input is focused and no query is active
- Only add to search history when a full query is run (not the autocomplete queries)


pfrazee:
Thanks again @skiniks !

PR #2599, opened 2024-01-23
more social card tweaks, and include in RSS as well

bnewbold:
This PR is on top of https://github.com/bluesky-social/social-app/pull/2427

Additions:

- merges main to fix an unrelated CI failure
- adds a single simple test for URL expansion
- tweaks code a bit to be more readable
- include a note when there is a quote post, though doesn't (yet) actually expand that quoted post text
- also uses this "expanded" text in RSS feeds


pfrazee (approved):
Awesome thanks all

PR #2602, opened 2024-01-23
Scroll feed search bar to top of screen on focus

mozzius:
This PR scrolls the feed search bar to the top of the screen on focus, to avoid the situation where it's too low on the screen and you can't see the results as you type.

https://github.com/bluesky-social/social-app/assets/10959775/dcfd30e5-6e59-41f4-bcf3-735743b106bc

Implementation notes:

`scrollToIndex` is used on native and works flawlessly. However, that's not available on web, so we use `scrollToOffset` there instead and have a guess at the offset. I measured the height of all the elements so it should be roughly accurate, but in my testing it wasn't perfectly lining up. Luckily there's a fair bit of padding above the search bar so there's plenty margin for error.


pfrazee (approved):
aahh very nice

PR #2603, opened 2024-01-24
Disable scroll on the composer TextInput (close #2413)

### LINKED ISSUE #2413: Input field does not scroll with content iOS

pfrazee:
In iOS, a multiline textinput has a scrolling behavior of its own. This created a scrollview-within-scrollview problem on iOS, because we embed the entire composer in a scrollview.

Disabling the scrolling behavior on the textinput solves this issue.

Closes #2413

### DIFF
--- a/src/view/com/composer/text-input/TextInput.tsx
+++ b/src/view/com/composer/text-input/TextInput.tsx
@@ -217,6 +217,7 @@ export const TextInput = forwardRef(function TextInputImpl(
         autoFocus={true}
         allowFontScaling
         multiline
+        scrollEnabled={false}
         numberOfLines={4}
         style={[
           pal.text,


markaudacity:
Nice. Thanks for the hard work. <3

markaudacity:
This is either not fully fixed or has been unfixed; it happened to me today on iOS 17/Bsky 1.67.0 (6)

Issue #875, opened 2023-06-13
Replies to your own posts are not visible on following feed


pepijndevos:
**Is your feature request related to a problem? Please describe.**

It is basically impossible to do long running threads because later additions to the thread don't show up in any timelines that I know of so nobody sees the later posts.

On Twitter adding a post to a thread bumps it to the top of your timeline. On Bluesky it only appears to show up on your profile.

I _do_ see replies of other people on my following feed, so it seems like this is either some super weird bug or an intentional feature?

IMO this encourages people to do less thoughtful threads and QT themselves more.

**Describe the solution you'd like**

I think bumping the thread when you reply to it makes sense, at least on the following feed.

For algorithmic feeds I guess it's up to the feed author if they want to include replies, and if/how they want to promote additions to threads.

**Describe alternatives you've considered**

An alternative would be to show replies to your own posts like it shows replies from other people, where you see the tail end of the thread and have to open it up to see the full context.

**Additional context**

https://bsky.app/profile/wilderless.quest/post/3jxzrpxg5242r

Drilski:
Current Behaviour:

It seems replies to your own posts (threads) will fall back in the follow feed - And be added to the original post way back in the timeline. They also seem to be hidden (if they don't have 2 likes?).

Desired behaviour:

When you reply to your own post (thread) the reply post should be prioritised and brought to the top of the timeline. This shouldn't need two likes to show on the follow feed.

pfrazee:
Yeah it makes sense to me to exclude your own posts from some of the rules. We can do this.

mschwendener:
I keep stumbling over the following, not sure if this is the right place to add it: 

I'm using the Only posts feed https://bsky.app/profile/did:plc:tenurhgjptubkk5zf5qhi3og/feed/only-posts 

Example of what I see: 
<img width="608" alt="screenshot_4385" src="https://github.com/bluesky-social/social-app/assets/4497705/96bf96c3-948a-4241-baf1-b1c00d52821c">

But the "2 replies" is misleading: it's a thread with 12 posts by the OP.

PR #2604, opened 2024-01-24
Bump replies

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/875

This partly addresses the complaints of low engagements on replies by bumping the entire slice up to the top of timeline. Given that we're requesting 60 posts at a time in the following feed, this means that threads can be much further apart and still be eligible for bumping.

It's somewhat open for abuse, but I don't think it should be an issue.

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -177,8 +177,11 @@ export class FeedTuner {
 
             parent.insert(item)
 
-            slices.splice(index, 1)
-            slices.unshift(parent)
+            // If our slice isn't currently on the top, reinsert it to the top.
+            if (index !== 0) {
+              slices.splice(index, 1)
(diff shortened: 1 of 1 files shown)


DavidBuchanan314:
I think this will be less useful for people who follow lots of users, or for threads composed over a longer time window (because a given thread will be less likely to be within the window of bump-ability), but if I understand the logic correctly then it'll never make things "worse". i.e. it'll either do nothing or be an improvement, so that's a :+1: from me (just based on your description, I haven't looked at or tested the impl)

pfrazee (approved):
Yep yep. Doesn't trigger often but when it does, it does the job exactly right

pfrazee:
Thanks Mary!

PR #2605, opened 2024-01-24
cleanup moderation styles

haileyok:
@estrattonbailey 

This should cleanup the issues. Simplified the AdultContentPref so all the containers use the same background style/rounded corners, etc. w/ less code. Also, made a few adjustments to ToggleButton so it adjust correctly. Tested on an iPhone SE and buttons stay within the container bounds now too.

Hopefully doesn't conflict, or if you've already done this just close it 👍


estrattonbailey (approved):
Sweet, thank you!

PR #2607, opened 2024-01-24
Fix server icon in sign up flow not responding to dark mode

mozzius:
Fixes a bug where an icon didn't respond to dark mode and thus was near impossible to see

Before:

Screenshot 2024-01-24 at 15 31 10

After:

Screenshot 2024-01-24 at 15 30 53


pfrazee:
Never fails. Gratzi mozzy

Issue #1017, opened 2023-07-15
add mute user button to posts


ghost:
ADd the ability to mute a user from clicking the 3 dots below a post.

Currently, we have to view the users profile we want to mute - which is an additional step and can also lead to people seeing things they do not want to see while viewing the home timeline, a comment section or a feed post.

this can make it faster to mute, easier and is a quality of life improvement 
![image](https://github.com/bluesky-social/social-app/assets/66647238/4c288e48-1ad5-4da7-8ffd-2c7991d9de51) 
![image](https://github.com/bluesky-social/social-app/assets/66647238/e4dfdca9-83a7-48f0-a107-19fbab61ad80)

pfrazee:
I'll give this a look as I work on these tools over the next couple weeks.

monicaellerose:
thinking about this-  it should probably also include a quck block option,

skiniks:
All set! PR #2610 ready for review.

monicaellerose:
adding this here https://github.com/bluesky-social/social-app/pull/3708

monicaellerose:
i hope this one can get boosted and reviewed pretty fast- seeing a huge increase in bad actors.

surfdude29:
Update: #6091 has just been merged which adds the ability to block someone from the post dropdown menu (but it doesn't include an option to mute them).

PR #2610, opened 2024-01-24
adding mute account and block account buttons to post dropdown

skiniks:
This PR adds Mute Account and Block Account buttons to the Post dropdown menu to close issue #1017.


pfrazee:
At a quick glance, one thing this needs is to check if the target user is already muted and give an unmute option if so. (This shouldn't be necessary for blocks, because it shouldn't be possible to see posts by blocked users.)

skiniks:
Thanks for the notes, @pfrazee! A toggle has been implemented with `useMyMutedAccountsQuery()`.

PR #2611, opened 2024-01-24
Use consistent spinners for loading screens

gaearon:
Post likes, reposts, profile follows and profile followers were showing a tiny spinner with no paddings.

Now they're using a large spinner with a padding, like post thread does.

This looked especially bad on the web. On the mobile it was fine but I'm going for consistency.

This is now one component; if we want to change how it looks, let's change it across all screens with this layout.

## Before

https://github.com/bluesky-social/social-app/assets/810438/7ac5db19-8fa0-4f36-8b97-d7d609027cfd

## After

https://github.com/bluesky-social/social-app/assets/810438/82d2a373-4f49-460a-9cfc-4fae5712f0ed

### DIFF
--- a/src/view/com/util/LoadingScreen.tsx
+++ b/src/view/com/util/LoadingScreen.tsx
@@ -1,0 +1,14 @@
+import React from 'react'
+import {ActivityIndicator, View} from 'react-native'
+import {s} from 'lib/styles'
+import {CenteredView} from './Views'
+
+export function LoadingScreen() {
+  return (
+    <CenteredView>
+      <View style={s.p20}>
+        <ActivityIndicator size="large" />
+      </View>
+    </CenteredView>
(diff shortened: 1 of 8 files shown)


src/view/com/util/LoadingScreen.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import {ActivityIndicator, View} from 'react-native'
+import {s} from 'lib/styles'
+import {CenteredView} from './Views'
+
+export function LoadingScreen() {
```
gaearon, line 6: Calling this "loading screen" to emphasize it's meant for cases where the entire screen is empty. Not meant as a generic "spinner" component.

pfrazee (approved):

PR #2613, opened 2024-01-25
Add account-activation queueing to signup

pfrazee:
Depends on https://github.com/bluesky-social/atproto/pull/2076

With that PR, it becomes possible for the server to create an account but place it in "deactivated" state. Deactivated accounts will go through a queue where they are enabled as the server has capacity.

This information is revealed to us through an `accessJwt` with the scope of `"com.atproto.deactived"`. Under that scope we're only able to call `erver.com.atproto.temp.checkSignupQueue` which gives us information about our status in the queue.

When that tells us we're activated, we refresh the token for a good one and continue onboarding.


src/components/Button.tsx
```diff
@@ -337,6 +337,7 @@ export function Button({
         a.flex_row,
         a.align_center,
         a.overflow_hidden,
+        a.justify_center,
```
pfrazee, line 340: @estrattonbailey we should talk about this

estrattonbailey (approved):

pfrazee:
Lint error fixed in another PR

PR #2615, opened 2024-01-25
Remove dead code for optimistic updates

gaearon:
In https://github.com/bluesky-social/social-app/pull/1933 I've moved the optimistic update handling from raw mutations to the "queue" methods like `useProfileFollowMutationQueue`. The individual optimistic update logic remained in the underlying mutation logic but was disabled by passing `{skipOptimistic: true}`. I left this as an option in case we want to use the raw mutations somewhere in the interface after all.

A few months later, we're still not using raw mutation methods for those toggles anywhere. And really, we shouldn't—if you're building some new UI, you should just use the queue anyway. So let's remove this option and assume it's always `true`. This lets us simplify the mutations and make it clearer what's actually going on.

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -149,7 +149,6 @@ export function useProfileFollowMutationQueue(
       if (shouldFollow) {
         const {uri} = await followMutation.mutateAsync({
           did,
-          skipOptimistic: true,
         })
         return uri
       } else {
@@ -157,7 +156,6 @@ export function useProfileFollowMutationQueue(
           await unfollowMutation.mutateAsync({
             did,
             followUri: prevFollowingUri,
-            skipOptimistic: true,
           })
         }
         return undefined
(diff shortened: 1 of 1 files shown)


src/state/queries/profile.ts
```diff
@@ -191,68 +189,22 @@ export function useProfileFollowMutationQueue(
...
-        })
-      }
-    },
     onSuccess(data, variables) {
-      if (!variables.skipOptimistic) {
-        // finalize
-        updateProfileShadow(variables.did, {
-          followingUri: data.uri,
-        })
-        track('Profile:Follow', {username: variables.did})
```
gaearon, line 216: Technically we accidentally were missing this event entirely because it was in a falsy branch. So moving it out is kind of a bugfix. I don't see it emitted elsewhere.

pfrazee (approved):
LGTM

jonstuebe:
Do you all have any blog posts that describe your architecture of optimistic updates and these queues? Sounds fascinating

gaearon:
Not really, but I've described this one in https://github.com/bluesky-social/social-app/pull/1933.

PR #2616, opened 2024-01-25
Derive shadow like/repost count

gaearon:
This is a simplification meant to avoid potential bugs. Not sure if it uncovers any bugs right now.

In principle we should always be able to derive the shadow like count instead of explicitly tracking it. Shadow like count should be `post.likeCount + 1` when `!wasLiked && isLiked`, `post.likeCount - 1` when `wasLiked && !isLiked`, or `post.likeCount` otherwise. A larger divergence than +/– 1 should not be possible because we only use the shadow for optimistic updates. So if it ever happened in practice, it would be due to a bug. Let's make this bug impossible.

### DIFF
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -12,9 +12,7 @@ export type {Shadow} from './types'
 
 export interface PostShadow {
   likeUri: string | undefined
-  likeCount: number | undefined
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Okay yep, had to turn this over in my head a couple times but I believe this will work.

PR #2621, opened 2024-01-25
Long press external embed to share

mozzius:
Opens the sharing menu when you long press the external link card. Only applies to native.


pfrazee (approved):
Works great!

PR #2622, opened 2024-01-25
Show composer reply prompt on tablet

gaearon:
On mobile, it's always at the bottom. On desktop, it's always inline. But on tablet sizes, we didn't show it at all. This makes the interface feel read-only. This change makes the tablet behavior the same as desktop — it's shown inline.

## Before

Screenshot 2024-01-25 at 13 26 48

## After

Screenshot 2024-01-25 at 13 26 56

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -152,7 +152,7 @@ function PostThreadLoaded({
   const {hasSession} = useSession()
   const {_} = useLingui()
   const pal = usePalette('default')
-  const {isTablet, isDesktop, isTabletOrMobile} = useWebMediaQueries()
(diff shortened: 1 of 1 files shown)


src/view/com/post-thread/PostThread.tsx
```diff
@@ -271,7 +271,7 @@ function PostThreadLoaded({
       } else if (item === REPLY_PROMPT && hasSession) {
         return (
           <View>
-            {isDesktop && <ComposePrompt onPressCompose={onPressReply} />}
+            {!isMobile && <ComposePrompt onPressCompose={onPressReply} />}
```
gaearon, line 274: Opted for `!isMobile` because the case where we show it at a fixed position uses `isMobile &&` as the condition. Ideally we'd unify those (so that it's always rendered from here) but whatever.

pfrazee (approved):
Yep yep great

PR #2625, opened 2024-01-25
Dejank navigation between thread posts

gaearon:
Currently, when opening a thread post, we'd show two spinners: the one below and the one above. This feels a bit janky/busy and distracts from actually reading the post you clicked on. Twitter solves this by scrolling you down just enough that you *don't see* the parent post (and so there is no need to show a spinner). I'm doing the same behavior here.

This also lets us remove some fragile logic that's bound to eventually get out of sync.

## Before

Web:

https://github.com/bluesky-social/social-app/assets/810438/dec6c9a3-0864-4515-a19f-cd668b556200

iOS:

https://github.com/bluesky-social/social-app/assets/810438/87bc5bac-0e1f-484e-96d4-8d7a97133f21

## After

Web:

https://github.com/bluesky-social/social-app/assets/810438/37b51efb-04c0-4375-a35c-f578689d9ec6

iOS:

https://github.com/bluesky-social/social-app/assets/810438/158463d5-1f29-4c26-9cf9-8ea835d665a4

(Sorry, my Android emulator is still hosed.)

PR #2628, opened 2024-01-25
Present highlighted post as part of the thread

gaearon:
I meant to include it in https://github.com/bluesky-social/social-app/pull/2625 but lost the change somehow.

The highlighted post is "the end of the thread so far" so it should stay avatar-connected to the parent ones.

## Before

Screenshot 2024-01-25 at 21 40 37

## After

Screenshot 2024-01-25 at 21 40 25


pfrazee (approved):
LGTM

Issue #2592, opened 2024-01-22
Android users unable to reliably select alt text


pfrazee:
Word is that some Android users can't get text selection to work. At times it seems that there's a secret spot in the UI that they have to aim for. Not clear why this is -- it worked for me in testing.

haileyok:
guessing this had to do with the fact the selection and on press for expand kinda interfere. afaik should be resolved in that selection pr, although i admit to not testing much on android there. trick though was to move the on press to the text itself rather than wrap it as a pressable.

Juicysteak117:
For whatever it's worth, I have never once been able to select alt text on Android to the point where I thought it simply wasn't implemented. I'd expect a long press to do it but it never has. Glad to know it's being looked into at least.

PR #2563, opened 2024-01-18
fix: make alt text selectable in Lightbox

trantus:
fixes #2293 
| Android | iOS |
| ------- | --- |
| Android | iOS |

PR #2629, opened 2024-01-25
android alt text selection fix for expandable alt

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2592

tested on device, alt text is selectable now regardless of whether it is expandable or not. right now, since RN handles the selection long press as a press regardless of length, it expands/collapses the text upon selection which is obviously janky.

similar to what was done in https://github.com/bluesky-social/social-app/pull/2563 except we needed to apply it to the `Text` instead of `Pressable`. Also verified that things still work fine on both other platforms.


pfrazee (approved):
Nice, gratzi

Issue #2631, opened 2024-01-25
Can't auxiliary click on posts (web)


mozzius:
**Describe the bug**

Clicking on a post body takes you to the post thread, but since the links are are fake (not anchor tags), there's no auxiliary click action on web (aka middle click to open in new tab)

**To Reproduce**

1. middle click on a post
2. observe nothing happening

**Expected behavior**

It opens the post thread in a new tab

**Details**

 - Platform: web
 - Platform version: firefox
 - App version: any

**Additional context**

~~Just need to add an `auxclick` handler (weirdly I can find next to no info on this for RNWeb)~~

Turns out Firefox is the issue here! Doesn't support `auxclick` properly

haileyok:
I actually have a local branch where I fixed this afaik. I'll look and clean it up.

PR #2633, opened 2024-01-26
fix web aux click on all browsers

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2631
fixes https://github.com/bluesky-social/social-app/issues/650

Fixes the "aux click" handler to work on all browsers. Currently, `auxclick` support seems to be a bit sketchy: https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event#browser_compatibility

Instead, wrap the `Link` component children inside of `WebAuxClickWrapper`, which adds a `div` with `onMouseDown` and `onMouseUp` handlers.

`onMouseDown` prevents the middle click scroll functionality from starting. `onMouseUp` checks to make sure the target is not an anchor tag, and if not simulates a meta click on the targeted element. This works on all browsers.

https://github.com/bluesky-social/social-app/assets/153161762/01611111-cebc-46ad-b2a9-ac6646f0c24f

ts error seems to be unrelated


pfrazee (approved):
Looks good!

PR #2634, opened 2024-01-26
Prefetch profile links on web

gaearon:
This adds prefetching on hover for profile user avatars and profile links in post meta.

We don't need to go wild with this but it's worth doing in some places.

The goal is to instantly jump to the next page's shell.

## Before

https://github.com/bluesky-social/social-app/assets/810438/0f23b910-6d53-4a7d-bd7e-a83b9459763b

## After

https://github.com/bluesky-social/social-app/assets/810438/62ec2668-3caa-4545-a095-6bb361279b19

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -57,6 +57,23 @@ export function useProfilesQuery({handles}: {handles: string[]}) {
   })
 }
 
+export function usePrefetchProfileQuery() {
+  const queryClient = useQueryClient()
+  const prefetchProfileQuery = useCallback(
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Nice. Going to merge, but one other thing we could look at is prefilling the profile placeholder cache using a search like we do elsewhere. The profile objects inside posts don't have all the data of the full `getProfile()` call but it has enough to optimize the loads significantly.

That approach would reduce traffic and work for mobile too.

Ferrozo (commented):
Well done :)

Issue #2595, opened 2024-01-23
Offer full-black dark mode setting


Juicysteak117:
**Is your feature request related to a problem? Please describe.**

As of the latest update (On Web at least, I'm unsure if my phone simply has yet to receive the update), the dark mode has changed from being black to being a hazy blueish gray. It's similar to GitHub's dark mode, but distinctly less readable. The contrast ratio for the main post text has only gone from 21 to 19.82, but the perceived contrast difference in context is far worse. It looks like there is a layer of fog on my screen with how much every color blends together because the ratio between the background and the borders is 1.93. Likewise, links on the sidebar to various feeds have dropped to 9.16. I find this much more difficult to read.

Now, aside from this being a personal preference (because with all due respect I find these color choices to be deeply revolting), the lack of a black mode is upsetting for OLED users given that pixels on an OLED screen are off when they are black. This difference in power usage is less appreciated on desktop, but for mobile users with OLED screens, it is absolutely noticeable. 

**Describe the solution you'd like**

Either reverting to the previous dark mode or offering a black mode (the previous dark theme) along side the current light and dark mode options.

**Describe alternatives you've considered**

I have three options:

1. Live with it.
2. Pump in my own CSS via an extension like Stylus.
3. Ask politely on GitHub for a change.

I do not wish to do Option 1, nor do I wish to deal with the tedium of Option 2, so respectfully I am here for Option 3.

**Additional context**

quiple:
I think it has to do with accessibility for users with astigmatism. #647

Personally, I find the gray background a little easier to look at. But I agree with providing an option for a full-black background!

Hima-Zinn:
It appears that such a feature was implemented in build version 1.66.0.

Personally, I liked the traditional black background with white text, but I think it would be better to add a new accessibility setting in the "Basics" section, where background color, text color, and other handicap-accessible settings can be combined and enabled for those who want to enable them, or disabled for those who want to disable them.

surfdude29:
FYI @pfrazee said this on Bluesky:

["yeah we're iterating toward 3 modes, one of which will have the pure black"](https://bsky.app/profile/pfrazee.com/post/3kjm2ksoxcl2g)

Juicysteak117:
@surfdude29 Ah, I must've missed that. That's great to hear, although I would've appreciated had they rolled out both changes simultaneously instead of leaving us stuck with the current colors, but that's how it is sometimes.

pfrazee:
[...]

PR #2637, opened 2024-01-26
Options for selecting preferred dark theme

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2595

Adds preference in settings to choose between the default dark theme or the OLED dark theme. To support this, we needed to make some changes (which incidentally will make adding additional themes down the road much easier):

- Create a `oledTheme` theme which uses all of the base tokens from `darkTheme` but sets `backgroundColor` to `trueBlack`
- Both the ALF and the legacy `ThemeProvider` take in the name of a theme now rather than a particular color mode. Options are `light`, `dark`, or `oled`. Assuming that `dim` is another theme option that is anticipated, it's now as easy as just updating the theme name type to add more.
- `useColorMode` and `useSetColorMode` is changed to `useThemePrefs` and `useSetThemePrefs`. Respectively, both contexts have `colorMode, darkTheme` and `setColorMode, setDarkTheme` as their values
- `useColorModeTheme` is changed to return the memoized theme name, and we handle the logic here to determine which theme name to use regardless of the color mode being system/light/dark.

Additionally on web since there are some styles inside of the root HTML document, it seems better to change this logic a bit. Instead of relying on `@media` here, we can use use the existing `updateDocument()` from the color mode context to set the particular theme that is being used (light, dark, or oled). Instead of adding/changing the class name inside of the context's `useEffect` however, we can handle that when the `useColorModeTheme` memo updates.

(If there is some reason that `@media` was being used over RN's `useColorScheme()` we can add that logic back. I tested that on Gnome and KDE, macOS, and Windows 11 and it seems to work fine)

https://github.com/bluesky-social/social-app/assets/153161762/d3fd59d5-fdeb-46cc-ab3a-8ba3e7b9d1ec


haileyok:
Inclined to maybe have a separate screen for appearance preferences, but I'll leave as is for now. Assuming that eventually things here will migrate to using ALF anyway.

OLED probably isn't the right name you want for the theme itself either. Probably could give the two different themes better names than just "Default" and "OLED".

surfdude29:
> OLED probably isn't the right name you want for the theme itself either. Probably could give the two different themes better names than just "Default" and "OLED".

Many users may not be familiar with the term OLED so I think you're right. A few options that come to mind are "Dim / Dark / Lights down" for dark but not pure black and "Dark / Night / Lights out" for pure black/OLED.

Hima-Zinn:
I think "Pale Blue"or"Sky Blue"... and "Black"or"True Dark"... are good option names.

"Default" → "Pale Blue"or"Sky Blue"...
"OLED" → "Black"or"True Dark"...

Also, it would be more useful if the "Dark Theme" could be displayed only when "Dark" is selected.

haileyok:
@Hima-Zinn Yea, it only displays if dark mode is selected or if "System" is selected (since you'll still end up in dark mode at night, and you'll want to pick an option for that time period)

Hima-Zinn:
@haileyok Thank you. I appreciate your concern.

Juicysteak117:
Fantastic, thank you Hailey!

haileyok:
moving this to https://github.com/bluesky-social/social-app/pull/2722

PR #2641, opened 2024-01-26
basic export repository link in settings

bnewbold:
EDIT: sneak attack by paul

Absolutely no prior React experience, and limited TypeScript, so probably doing all kinds of things wrong!

~I tried to make it a download button instead of link but that didn't work.~ I got that attribute on there, but my web browser isn't respecting it. We might need to make a specific endpoint that doesn't produce an idiotic filename.

~There is probably a safer way to construct the URL string.~ fixed

I think having the download open in the browser is reasonable, as opposed to an in-app save flow in mobile. But i'm not sure.

Did not commit the updated translation files (.po)

---

Final UI pics:

|Settings|Dialog|
|-|-|
|CleanShot 2024-02-12 at 14 58 55@2x|CleanShot 2024-02-12 at 15 03 44@2x|


haileyok (approved):
LGTM, love the dialogs on web 🎉

bnewbold:
:heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

PR #2646, opened 2024-01-26
Always show post dropdown button at the bottom of the post, add share button to highlighted post

gaearon:
It's currently inconsistent between highlighted and non-highlighted posts. I don't see a good reason for the inconsistency. We could always tuck it away to upper right (like twitter does) but maybe that's more controversial so I'm doing the opposite.

We definitely have space for that button.

## Before

Screenshot 2024-01-26 at 17 18 25

## After

Screenshot 2024-01-26 at 17 18 10


src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -187,9 +184,6 @@ let PostThreadItemLoaded = ({
     return makeProfileLink(post.author, 'post', urip.rkey, 'reposted-by')
   }, [post.uri, post.author])
   const repostsTitle = _(msg`Reposts of this post`)
-  const isModeratedPost =
-    moderation.decisions.post.cause?.type === 'label' &&
-    moderation.decisions.post.cause.label.src !== currentAccount?.did
 
   const translatorUrl = getTranslatorLink(
     record?.text || '',
```
arkhurst, line 189: can we use "??" instead of "||"

 const translatorUrl = getTranslatorLink(
    record?.text ?? ""

src/components/icons/TEMPLATE.tsx
```diff
@@ -46,3 +46,25 @@ export function createSinglePathSVG({path}: {path: string}) {
...
+        ref={ref}
+        viewBox="0 0 18 18"
+        width={size}
+        height={size}
+        style={[style]}>
+        <Path fill={fill} fillRule="evenodd" clipRule="evenodd" d={path} />
+      </Svg>
+    )
+  })
+}
```
haileyok, line 70: @estrattonbailey Can you give me some thoughts on why I might be needing to do this? Did I export the icon wrong? (Very very possible lol)

haileyok:
image

Added share button to the row. On web copies the link to clipboard.

src/components/icons/TEMPLATE.tsx
haileyok: nevermind, i figured it out. figma noob.

src/view/com/post-thread/PostThreadItem.tsx
znareak: I Guess that in this case, It's same behavior with ??.

PR #2651, opened 2024-01-26
Remove dead code for pager

gaearon:
Bottom position is never being used and already doesn't work properly. Let's rm.


luan03:
@gaearon In short, are you removing this only because the code is not used and adds extra complexity?

pfrazee (approved):
Yep good w/me

PR #2652, opened 2024-01-26
Update pt-BR translation to 1.67

gildaswise:
Adds new strings from 1.67


src/locale/locales/pt-BR/messages.po
```diff
@@ -4538,7 +4538,7 @@ msgstr "Sua escolha será salva, mas você pode trocá-la nas configurações de
 
 #: src/screens/Onboarding/StepFollowingFeed.tsx:61
 msgid "Your default feed is \"Following\""
-msgstr ""
+msgstr "Seu feed inicial é o \"Seguindo\"\"
```
surfdude29, line 4541: I think there might be an extra \ at the end here? 🤔

gildaswise: Nice catch!

src/locale/locales/pt-BR/messages.po
```diff
@@ -630,7 +630,7 @@ msgstr "Altere o Seu Email"
 #: src/screens/Deactivated.tsx:73
 #: src/screens/Deactivated.tsx:77
 msgid "Check my status"
-msgstr ""
+msgstr "Verificar minha situação"
```
luan03, line 633:
```suggestion
msgstr "Verificar meu status"
```

**Status** is already used in the current translation, so I'd suggest using that instead of situação.

gildaswise:
```
#: src/view/screens/Settings.tsx:804
msgid "Status page"
msgstr "Página de status"
```

The other usage of Status is related to Bluesky's status page, in this case it's used in a queue context. The button will check if your account is already activated. "Verificar minha situação na fila" would be a better translation but "fila" is already used in the screen, so I didn't want to repeat it there

pfrazee (approved):
Awesome, thanks!

PR #2654, opened 2024-01-27
Only re-reg push noty token when account changes

estrattonbailey:
`currentAccount` is mutated a couple times as we init e.g. updated the refresh token, which caused these effects to re-run 3x and register a push token 5x (since the listener was destroyed and re-instantiated also).

This PR consolidates these two effects and does a check for a did to make sure it only runs when the account did changes. So `registerPush` should only be called 2x now instead of 5x.

I considered doing this in our `session` state, but as mentioned, some props on `currentAccount` DO need to mutate. So for this use case I thought this was more direct.


pfrazee (approved):
lgtm

Issue #2645, opened 2024-01-26
Tapping Translate from post dropdown doesn't respect in-app browser pref


surfdude29:
**Describe the bug**

When you tap Translate from the post dropdown menu (not the blue Translate link that is shown for posts tagged in a different language) the app opens the Google Translate webpage in the browser, even if the 'Open links with in-app browser' preference is toggled on.

**To Reproduce**

1. Go to Settings and toggle on the 'Open links with in-app browser' preference.
2. Go to any post and tap Translate from the post dropdown menu.
3. Observe that Google Translate is opened in the browser app.

**Screen recording**

https://github.com/bluesky-social/social-app/assets/149612116/ed12a141-0c07-4e41-a5f5-b9748097f729

**Expected behavior**

Google Translate should open in the in-app browser when the preference for that is toggled on.

**Details**

 - Platform: iOS
 - Platform version: 15.8.1
 - App version: 1.66

surfdude29:
Cheers @haileyok and @pfrazee ! 🙏

PR #2663, opened 2024-01-29
use in app browser for translations when enabled

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2645

just use the `useOpenLink()` hook for opening translations instead of `Linking.openURL()` directly


pfrazee (approved):
Nice

Issue #2656, opened 2024-01-27
Composer reply avatar doesn't respect profile moderation blur


mozzius:
**Describe the bug**

When replying to someone with a blurred avatar (due to profile moderation), the avatar in the composer shows the unblurred image.

**To Reproduce**

1. Find someone with a blurred avatar
2. Reply to one of their posts

**Expected behavior**

Avatar is blurred

**Details**

 - Platform: macOS firefox
 - App version: 1.67

PR #2665, opened 2024-01-29
moderate avatars and embeds in composer reply to

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2656

pass `PostModeration` into composer `replyTo` and:

- blur avatars if needed
- hide image embeds if needed

Screenshot 2024-01-28 at 5 03 03 PM


pfrazee (approved):
Works great

Issue #1876, opened 2023-11-12
unable to scroll the following and followers list to the end


sowtiny:
**Describe the bug**
some users can not scroll the following and followers list until the end, it gets stuck in the middle. and the numbers shown on the profile don't match the actual number of following and followers on the list.

**To Reproduce**

1. open bluesky profile.
2. tap the following/followers list.
3. scroll the list until the end.

**Expected behavior**

if we scroll all the way to the end, we will find our first following/followers. and the number of following/followers on the list will be the same as the number shown on the profile.

**Screenshots**

**Details**

 - Platform:  android
 - Platform version: 
 - App version: 1.56.0 (46)

**Additional context**

pfrazee:
Is this still happening?

sowtiny:
yes, this issue is still happening.

haileyok:
@sowtiny does this seem like an issue where you just can't scroll down far enough to see the followers or they are just not showing up at all when loading them?

Timepixels:
It's happening with me too, I created multiple accounts and same things happening with other account.

aalku:
I think there is a misunderstanding.
I think the reporters want the following. I talked to other people complaining about the same.

Let's say you use the app until 5pm on day one and read everything up to that point. Then you close it and that's it that day.
Then on day two you open the app again at 4pm and refresh to see new content but you want to scroll back to read absolutely everything new, so you expect to find every post from day 1 at 5pm to day 2 at 4pm to load and they don't, some are missing.

That's what they told me, I never tried.

barryr:
Why was this bug closed? It is still happening, both on the website (Firefox browser) and the app (iPhone). I have over 350 followers, but can only see the 20 - 30 most recent. I can look at the profiles of people I know are still following me and see the "Follows You" flag, but they don't show up in my Followers list.

PR #2675, opened 2024-01-29
add padding to user's lists/feeds scrollviews

haileyok:
Original issue in https://github.com/bluesky-social/social-app/issues/1876 isn't reproducible as a padding/margin issue on any platform. Seems that the `ListFooterComponent` always adds the appropriate amount of padding to the view...

This does fix issues where a large list of lists/feeds isn't scrollable on user profiles though, since there was no accounting for the header offset on those tabs. Will keep playing with different follower screens to see if it is ever reproducible.


pfrazee (approved):
LGTM

PR #2676, opened 2024-01-29
Handle URL params

estrattonbailey:
Adds handling to store UTM params locally to track referrals from our marketing site. WIP, will require additional code to emit events from create account flow. No personal data will be transmit.


haileyok:
Stealing this PR to do some linking work for share extension and other intents.

Issue #2674, opened 2024-01-29
Scrolling in profiles' list inclusion modal is now broken


venteto:
**Describe the bug**
When I visit someone's profile so I can add or remove them from curated user lists, I cannot scroll inside the modal that is opened, and I have too many lists to fit within the default modal's dimensions. Thus many of my lists are not accessible for list member management actions.

**To Reproduce**

1. Go to someone's profile
2. Click the 3 horizontal dots beside the Follow button (on the desktop/web client)
3. Click Add To Lists
4. If you have made more than a few lists, only a few will be visible in the viewport, and you will not be able to scroll down to the others
5. This prohibits you from adding the account to one of your older lists, or from removing them from one of your older lists (and potentially also putting them into a different list instead)

**Expected behavior**

I was previously able to scroll up and down the entire list of my user lists, and add/remove accounts from lists at will

**Screenshots**

(I have "placeholders" because I cannot arbitrarily sort / order my lists, and would have to know in advance every list I will ever want to have, and then create each in reverse alphabetical order)
![Screenshot from 2024-01-29 15-58-22](https://github.com/bluesky-social/social-app/assets/7062385/3ccbed65-2e91-49a3-a0a0-d1c12efcb8da)

**Details**

 - Platform: desktop Firefox on Linux

PR #2678, opened 2024-01-29
adjust height of "add to list" modal based on screen size on web

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/2674

Adding `flex: 1` to the list in the modal doesn't work on web, since a smaller screen will cause it to overflow. Should be handled better with ALF dialog since the content container will adjust height accordingly, but for now a quick fix to just use the screen's height and adjust accordingly.


pfrazee (approved):
Yep looks good

PR #2679, opened 2024-01-30
clear cache when blocking/unblocking and whenever we get blocked, better invalidation logic for `useProfileQuery`

haileyok:
Right now we don't clear the post cache at all whenever we block/unblock someone or they block us. We might visit a profile and see that we are blocked since the avatar is blurred and we get the blocked label in the bio, but the posts are still visible. Instead, if we are blocked we should reset the posts query to ensure we can't see any content there.

Same for when we block someone. We technically shouldn't be still seeing the content from the user, rather it should be removed as soon as we block. Here we need to wait a moment for the AppView to receive the block.

We also don't need to invalidate profile queries if they are only for display name resolution. By setting those queries to not invalidate, we can set all the others to a lower 15 seconds ensuring we always have the most recent block state.

No extra queries are getting made in the feed:

https://github.com/bluesky-social/social-app/assets/153161762/e6b2cc3d-088e-4ebc-a7bd-c099e8142688

Profile and posts invalidate after blocks:

https://github.com/bluesky-social/social-app/assets/153161762/e0f63f82-0130-47c3-a6ab-ff1c385ceb74

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -17,12 +17,12 @@ import {updateProfileShadow} from '../cache/profile-shadow'
 import {uploadBlob} from '#/lib/api'
 import {until} from '#/lib/async/until'
 import {Shadow} from '#/state/cache/types'
+import {resetProfilePostsQueries} from '#/state/queries/post-feed'
 import {useToggleMutationQueue} from '#/lib/hooks/useToggleMutationQueue'
 import {RQKEY as RQKEY_MY_MUTED} from './my-muted-accounts'
 import {RQKEY as RQKEY_MY_BLOCKED} from './my-blocked-accounts'
 import {STALE} from '#/state/queries'
 import {track} from '#/lib/analytics/analytics'
-import {queryClient} from '#/lib/react-query'
 
 export const RQKEY = (did: string) => ['profile', did]
 export const profilesQueryKey = (handles: string[]) => ['profiles', handles]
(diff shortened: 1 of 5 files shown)


src/view/com/modals/ProfilePreview.tsx
```diff
@@ -27,12 +27,12 @@ export function Component({did}: {did: string}) {
...
     refetch: refetchProfile,
-    isFetching: isFetchingProfile,
+    isLoading: isLoadingProfile,
   } = useProfileQuery({
     did: did,
   })
 
-  if (isFetchingProfile || !moderationOpts) {
+  if (isLoadingProfile || !moderationOpts) {
     return (
```
haileyok, line 36: we'd rather use `isLoading` instead of `isFetching` since we want to show the avatar/banner/bio as long as it's been loaded once. the 15 second stale time will handle updating it if it has changed.

src/state/queries/profile.ts
```diff
@@ -426,3 +434,15 @@ export function* findAllProfilesInQueryData(
     }
   }
 }
+
+export function resetProfilePostsQueries(did: string, timeout = 0) {
```
pfrazee, line 438: Can we move this into `post-feed`? We try to colocate all functions that affect a query in the file that controls the query

src/view/com/util/UserInfoText.tsx
```diff
@@ -29,7 +30,10 @@ export function UserInfoText({
   attr = attr || 'handle'
   failed = failed || 'user'
 
-  const {data: profile, isError} = useProfileQuery({did})
+  const {data: profile, isError} = useProfileQuery({
+    did,
+    staleTime: STALE.INFINITY,
```
pfrazee, line 35: Let's set this to fifteen minutes so we don't accumulate a huge cache (the profile objects are large enough to care)

pfrazee (approved):
Okay great, seems to work well.

I had to look up why this works. `staleTime` is per-hook. I believe that means you can actually use different staleTimes for the same query, which is counter-intuitive for useQuery because you often have to think in terms of each query's cache and its shared lifetime between the hooks that use it.

src/view/com/util/UserInfoText.tsx
pfrazee: Yeah for future documentation -- `staleTime` does not drive useQuery cache lifetimes

PR #2680, opened 2024-01-30
implementing client-side muted words feature

skiniks:
This PR adds a Muted Words feature to the moderation settings. Muted terms are hidden from feeds, notifications, and from profiles that contain a muted word in the handle, displayName or bio.

Screenshot 2024-01-29 at 9 31 52 PM

NOTE: In Moderation.tsx, it's currently using the same ban icon as blocked users.


pfrazee:
This is an awesome contribution. Let me sync with the team on how we intended to implement this and see how big of a delta there is between this PR and our plans. Then we can try to take this to the finish line. I'll try to keep this top of mind.

pfrazee:
Thanks again for this contrib -- as you probably know, we got this in

Issue #2562, opened 2024-01-18
Notifications from muted thread


YuriGal:
**Describe the bug**

If a thread is muted - posts from that thread no longer appear in Notifications tab, but I still receive notifications (number on the bell icon)

1. Create a thread of a few posts and replies
2. Mute the thread
3. Add a few posts to the thread

**Expected behavior**

No notification should appear on the bell icon.

**Details**

- Platform: Chrome Desktop on Mac
- Platform version: 120.0.6099.234
- App version: 1.65.0

**Screenshots**

1st screen - 5 new notifications count

![Screenshot 2024-01-20 at 1 58 40 PM](https://github.com/bluesky-social/social-app/assets/50464505/07327e4d-8b7f-4326-8bfb-766215e4cbdf)

2nd screen - when clicked/refreshed - no new notifications

![Screenshot 2024-01-20 at 1 58 52 PM](https://github.com/bluesky-social/social-app/assets/50464505/ce5434bb-03b0-41e7-a7ad-19e406f714d8)

mozzius:
This is because they're just filtered out on the clientside, and the API call to fetch the unread count can't tell whether or not the count includes posts that should be ignored. This also means that they can come through as push notifications - I hope one day we can get muted threads moved to the protocol level

YuriGal:
When the client makes a request to get the unread count - can it pass information about muted threads along with the request?

pfrazee:
Yeah there are three issues, two of which we can address now:

1. Push notifications aren't applying the filter (can't address yet)
2. It looks like the unread counter isn't filtering properly (can address)
3. Also gotten reports that quote post filtering isnt working (can hopefully address)

PR #2686, opened 2024-01-30
check if a thread is muted before incrementing notif badge, filter out quotes

haileyok:
Adding some logic to `isThreadMuted` for when we don't have a subject URI (when counting)

- We can always check if the thread itself is muted from the post
- We can check root level quotes to see if they are muted
- Same for reposts, we can check if they are muted when the repost is the root

Hopefully will cover the majority of cases for now.

https://github.com/bluesky-social/social-app/issues/2562

### DIFF
--- a/src/state/queries/notifications/unread.tsx
+++ b/src/state/queries/notifications/unread.tsx
@@ -5,7 +5,11 @@
 import React from 'react'
 import * as Notifications from 'expo-notifications'
 import {useQueryClient} from '@tanstack/react-query'
-import {AppBskyFeedPost} from '@atproto/api'
+import {
+  AppBskyEmbedRecord,
+  AppBskyFeedPost,
+  AppBskyFeedRepost,
+} from '@atproto/api'
 import BroadcastChannel from '#/lib/broadcast'
 import {useSession, getAgent} from '#/state/session'
 import {useModerationOpts} from '../preferences'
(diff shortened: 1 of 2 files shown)


src/state/queries/notifications/unread.tsx
```diff
@@ -136,7 +136,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
 
             // only fetch subjects when the page is going to be used
             // in the notifications query, otherwise skip it
-            fetchAdditionalData: !!invalidate,
+            fetchAdditionalData: true,
```
haileyok, line 139: If we fetch the additional data every time, we can filter out root quotes and all levels of reposts

src/state/queries/notifications/util.ts
```diff
@@ -205,8 +206,15 @@ function getSubjectUri(
...
-  if (type === 'reply' || type === 'quote' || type === 'mention') {
+  if (type === 'reply' || type === 'mention') {
     return notif.uri
+  } else if (type === 'quote') {
+    if (
+      AppBskyFeedPost.isRecord(notif.record) &&
+      AppBskyEmbedRecord.isMain(notif.record.embed)
+    ) {
+      return notif.record.embed.record.uri
+    }
```
haileyok, line 217: Shouldn't the subject be the quoted post's URI here? A quote post will never be muted since it's new, but we might have the quoted muted.

pfrazee: Ah no -- the subject should be the quote itself. The subject is the content that drove the notification; it drives what will show in the notifs. If we return the embed target (the quoted post) then the notification would render as your own post

pfrazee (changes requested):
Cool this is close, a couple things

1. Pretty sure we don't want to change the subject per my comment, which means that the subject codepath in `isThreadMuted` needs to be modified.
2. Added codepath in `isThreadMuted` looks good otherwise
3. There's a leftover `// TODO: thread muting is not being applied` that I think can get removed now, it's done in a separate area

src/state/queries/notifications/util.ts
haileyok: Right, good catch thanks. Changed logic to just check the embed inside the `isThreadMuted` codepath the same way we would without the subject. Only applies to the root, of course

pfrazee (approved):
Okay solid improvement until we can do a more thorough rebuild, thanks!

PR #2691, opened 2024-01-30
Don't interpolate rkey for custom feed notification

estrattonbailey:
Proposal: not sure how useful the `rkey` is here, since feed card shows below anyway. I can only repro the interpolation failure on mobile.

This PR changes what you see below to simply read "liked your custom feed".

CleanShot 2024-01-30 at 16 14 34@2x


pfrazee (approved):
Yeah that works for me

PR #2692, opened 2024-01-30
i18n interests, allow for fallbacks

estrattonbailey:
- moves interest mapping within a component so we can translate it
- adds fallback handling so if we add an interest group, it will simply display the untranslated tag value, capitalized


pfrazee (approved):
Cool lgtm

Issue #2522, opened 2024-01-13
Post did not display full length when focused


Juicysteak117:
**Describe the bug**

On mobile (Android), I experienced an odd bug where a post on my timeline was fine, but the moment I tapped on the post, the bottom of the post cut off. Nothing looks particularly unusual about the post so I am unsure what happened. It is fine when viewed in the feed and it is fine when viewed from a reply, but being focused on the post itself is the problem.

**To Reproduce**

1. I honestly have no idea this is the first time I've ever seen an issue like this.
2. Repeatedly clicking in or out of the post, including after a restart and on latest app version, has the issue persist.
3. Notice how it cuts off after "just"

Link to the post in question: https://bsky.app/profile/mary.my.id/post/3kis5dgivpt27

**Expected behavior**

**Screenshots**

![Screenshot_20240112-095213](https://github.com/bluesky-social/social-app/assets/20595808/c473f0fc-56e3-45bb-9a73-5db8978237ea)
![Screenshot_20240112-095228](https://github.com/bluesky-social/social-app/assets/20595808/8eb6afaf-0df2-4584-bc8d-2d7d424e4842)
![Screenshot_20240112-095240](https://github.com/bluesky-social/social-app/assets/20595808/bce3d95d-e855-4c20-9310-9fd0842565b7)

**Details**

 - Platform: Android
 - Platform version: 12
 - App version: 1.64

**Additional context**

pfrazee:
Pretty odd. Appreciate the lengthy report, we'll look into it

haileyok:
@Juicysteak117 Do you have any settings that might change fonts or anything? (I don't use android so...not sure what's available but first thing that comes to mind would be font size or custom fonts?) Using 1.64 on platform 12 doesn't repro this for me unfortunately...

Juicysteak117:
@haileyok No custom fonts or anything that I can think od. Android 13 on a Pixel 4a and the issue still persists on that specific post, but I've never seen it elsewhere.

haileyok:
Heh nice. Let me see if it repros on that specific device (simulator...)

haileyok:
@Juicysteak117 Okay you have a very specific screen size and found a really strange bug as a result lol, thanks for giving me the device model!

PR #2693, opened 2024-01-30
remove unnecessary right padding in main thread item (fixes weird android cutoff)

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2522 (?)

Very specific case where Android on a Pixel 4a and platform 31 (yea, that specific...probably related to that precise screen size) doesn't want to render text when we add the additional padding to the main post. I'm not actually sure that this is a real "fix" as far as it never happening again (I'm inclined to think its an Android bug in text measurement), but it does fix this particular case to remove the padding.

Tested a ton of posts by removing the padding, if anything the visual affect is a bit better. Right now text kind of line breaks a bit too early, removing that bit of padding helps a good deal from all the posts I opened up.

We could say forget it too and just mark it up as an OS bug, but I don't see the harm in removing this anyway.

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -775,6 +775,7 @@ const useStyles = () => {
     },
     postTextLargeContainer: {
       paddingHorizontal: 0,
+      paddingRight: 0,
       paddingBottom: 10,
     },
     translateLink: {


pfrazee (approved):
https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExbmJuNWprMHNib3l4dHFld3UyY3dwN3lmMHd2YjhwejNtYXdubG52dyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/ukGm72ZLZvYfS/giphy.gif

Issue #2601, opened 2024-01-23
Left swipe is buggy under very specific circumstances


gpp-0:
**Describe the bug**
While on the Following feed or on the Posts tab in someone's profile, if you swipe left on a post with two or more images, instead of switching to the next feed/tab it tries to open the side drawer in a weird way

**To Reproduce**

1. Go to the Following feed, or open up someone's profile and go to the Posts tab
2. Find a post with two or more images in it. Posts with one image (or no images) work as they should
3. Swipe left, starting the gesture at one of the pictures
4. The app gets confused and tries to extend the drawer

**Expected behavior**

No drawer shenanigans, and the timeline should switch to the next feed

**Screenshots**

Showing first how it's buggy under the above conditions, and then how it works fine if you swipe anywhere else

https://github.com/bluesky-social/social-app/assets/52042597/ca8e38a1-70ff-4493-af28-e2973822984e

**Details**

 - Platform: Android
 - Platform version: 10
 - App version: 1.66

**Additional context**

This faulty gesture capturing also randomly happens under different circumstances, but this was the most reliable way to reproduce it, it basically happens 100% of the time. Posts with 2 images are common so I encounter this bug very often when trying to switch feeds

pfrazee:
Seems to be related to https://github.com/bluesky-social/social-app/issues/2677

PR #2695, opened 2024-01-31
fix left swipe on android

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2601

Optimistic that we can use `gap` on native. Removing the negative margins from the container is the key to fixing the swipe issue, and looks like we were using that because of the yoga issue.

Noted that on web, using `gap` with `aspectRatio` definitely didn't work right for 3 images, but it seems to be okay on native. We don't need `aspectRatio` there because expo-image handles the sizing (based on the containers height from the other two images with `aspectRatio: 1`, whereas it looks like on web it doesn't.

Did have to slightly adjust the alt text label as well, since removing that margin moved it in a bit.

### DIFF
--- a/src/view/com/util/images/ImageLayoutGrid.tsx
+++ b/src/view/com/util/images/ImageLayoutGrid.tsx
@@ -51,7 +51,7 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
           <View style={{flex: 2, aspectRatio: isWeb ? 1 : undefined}}>
             <GalleryItem {...props} index={0} imageStyle={styles.image} />
           </View>
-          <View style={{flex: 1, gap: !isWeb ? 5 : undefined}}>
+          <View style={{flex: 1, gap: !isWeb ? IMAGE_GAP : undefined}}>
             <View style={styles.smallItem}>
               <GalleryItem {...props} index={1} imageStyle={styles.image} />
             </View>
--- a/src/view/com/util/images/ImageLayoutGrid.tsx
+++ b/src/view/com/util/images/ImageLayoutGrid.tsx
@@ -48,10 +48,10 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
     case 3:
       return (
         <View style={styles.flexRow}>
-          <View style={{flex: 2, aspectRatio: isWeb ? 1 : undefined}}>
+          <View style={styles.threeSingle}>
             <GalleryItem {...props} index={0} imageStyle={styles.image} />
           </View>
-          <View style={{flex: 1, gap: !isWeb ? IMAGE_GAP : undefined}}>
+          <View style={styles.threeDouble}>
             <View style={styles.smallItem}>
               <GalleryItem {...props} index={1} imageStyle={styles.image} />
             </View>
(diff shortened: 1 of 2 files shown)


src/view/com/util/images/ImageLayoutGrid.tsx
```diff
@@ -93,13 +94,19 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
...
-  flexRow: {flexDirection: 'row'},
+  container: isWeb
+    ? {
+        marginHorizontal: -IMAGE_GAP / 2,
+        marginVertical: -IMAGE_GAP / 2,
+      }
+    : {
+        gap: IMAGE_GAP,
+      },
+  flexRow: {flexDirection: 'row', gap: !isWeb ? IMAGE_GAP : undefined},
```
gaearon, line 105: Nit: let's order so that `isWeb` is either always checked positively like `isWeb ? foo : bar`. Since the hacks are all related, it's a bit hard to read what the end result is when half declarations in this file are negated and the other half aren't.

src/view/com/util/images/ImageLayoutGrid.tsx
```diff
@@ -47,10 +48,10 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
     case 3:
       return (
         <View style={styles.flexRow}>
-          <View style={{flex: 2, aspectRatio: 1}}>
+          <View style={{flex: 2, aspectRatio: isWeb ? 1 : undefined}}>
```
gaearon, line 51: Nit: let's move these to `styles` below so that all related hacks are visually close in the code

src/view/com/util/images/ImageLayoutGrid.tsx
```diff
@@ -93,13 +94,28 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
 const IMAGE_GAP = 5
 
 const styles = StyleSheet.create({
-  container: {
-    marginHorizontal: -IMAGE_GAP / 2,
-    marginVertical: -IMAGE_GAP / 2,
-  },
-  flexRow: {flexDirection: 'row'},
+  flexRow: {flexDirection: 'row', gap: !isWeb ? IMAGE_GAP : undefined},
```
gaearon, line 97: Let's make this one positive too for consistency

src/view/com/util/images/ImageLayoutGrid.tsx
```diff
@@ -93,13 +94,28 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
 const IMAGE_GAP = 5
```
gaearon, line 94: Is the comment above still correct? Might be nice to update if we changed the strategy in any way. Ideally it should be clear what the hacky part is (and whether we can ever remove the hack).

gaearon (approved):

src/view/com/util/images/ImageLayoutGrid.tsx
haileyok: Yea, the issue is actually still valid. But adding something to clarify why we're avoiding the margin solution on native.

haileyok:
Tested using various dimensions for the photos (landscape/portrait), layout remains correct on all platforms.

PR #2696, opened 2024-01-31
Fix notification mark-read behaviors

pfrazee:
Noticed these when testing the thread muting

Essentially the notification filtering was interrupting the correct `markAllRead` behavior. We need to be using the `indexedAt` of the most recent notification in `updateSeen`, and so we need to capture that timestamp prior to filtering. I also removed an optimization that assumed that you'd never filter out a whole page of notifs, which you indeed can with a thread mute


haileyok (approved):
LGTM!

PR #2697, opened 2024-01-31
Fix notification invalidation condition on web

gaearon:
Fixes an issue where `isScrolledDown` is not correct on focus, which caused us to not invalidate when we should have.

This is kind of a hack.

The `isScrolledDown` is not always reliably synchronous after the web layout PR. I've tried to fix that in a few different ways but it's a bit difficult due to the interplay of React Navigation, React's own rendering cycle, and the scroll restoration behaviors.

It's easier to just synchronously query it directly on web.

## Test Plan

Scroll down notifications. Go to another tab. Return to notifications. We're still scrolled down, so no invalidation.

Scroll notifications to top. Go to another tab. Return to notifications. We're scrolled up, so we invalidate.

No changes to behavior on other platforms.


pfrazee (approved):
Yep works great, gratzi

PR #2699, opened 2024-01-31
Workaround Safari content-visibility bug

gaearon:
This property should in theory be useful for improving paint perf for lists (I haven't measured but I know it can bring good results from past practice). However, it seems currently borked in Safari Technology Preview and Orion. It might be a temporary regression (they're actively landing PRs related to this property in WebKit repo) but for now let's turn it off for Safari.


pfrazee (approved):
SGTM

PR #2700, opened 2024-01-31
Reduce the odds of false positives in the language suggester

### LINKED ISSUE #2598: Language detection sometimes gets outright offensive
[https://bsky.app/profile/psyhfuck.bsky.social/post/3kjoea4227q2r](https://bsky.app/profile/psyhfuck.bsky.social/post/3kjoea4227q2r)

bafkreifzokbcbdbupxr35o5n4vdmiwwq4tb67vnaftw7cdspu6r7xp6pt4

Suggesting "russian" on a text clearly written in Ukrainian is very, very bad (especially if you capitalize "russian"). Please do better.

pfrazee:
Hopefully closes #2598 

I did more testing sessions on lande's output for given test strings. I noticed that false positives happen seem to correlate to having multiple matches with a score higher than `0.000x` (4 decimals of certainty) and so this adds logic to discard suggestions that match those conditions.

### DIFF
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -23,7 +23,9 @@ const onIdle = globalThis.requestIdleCallback || (cb => setTimeout(cb, 1))
 const cancelIdle = globalThis.cancelIdleCallback || clearTimeout
 
 export function SuggestedLanguage({text}: {text: string}) {
-  const [suggestedLanguage, setSuggestedLanguage] = useState<string>()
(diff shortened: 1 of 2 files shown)


pfrazee:
I downloaded a corpus of language samples from https://tatoeba.org/en/downloads to test this against. Could probably get more rigorous.

With this PR, I got `16715 right, 133 wrong, 99318 unsure`. Without this PR, I got `30353 right, 1370 wrong, 84443 skipped`. A 10x reduction of false positives for a 2x reduction of true positives is a good trade IMO.

haileyok (approved):
Yea logic seems fine, and the reduction is good. Definitely better to err on the side of not guessing than guessing incorrectly.

PR #2701, opened 2024-01-31
Prevent post meta misclicks in QT

gaearon:
I often accidentally click on the author when attempting to click on the embedded post. This is hard to get muscle memory for because the author name is dynamic and the click area is often very small.

Screenshot 2024-01-31 at 03 46 20

I'd argue that this direct click-through is unnecessary, and it's ok if clicking the meta always lands you on the quoted post. From there you can click into the author. (That's how it works on Twitter.)

That's what this PR does.


pfrazee (approved):
Works great

Issue #2659, opened 2024-01-28
Right-side list of pinned feeds isn't updating when deleting a user list


PratesM:
**Describe the bug**
When deleting a User List that has been pinned, the right-side bar isn't updating to reflect that said list is no longer available, only updates after you refresh the browser window.

**To Reproduce**

1. Create a User List
2. Click "Pin to home"
3. Delete said User List with the "Delete List" button

**Expected behavior**

Given that the right-side bar list updates when a list is pinned and unpinned, it would be expected to also update when a list is deleted.

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/32964925/d3a32ff0-b5f5-4b96-bdee-0cf86b52ae19)

**Details**

 - Platform: Web, Microsoft Edge, Desktop, Windows 11
 - Platform version: Version 121.0.2277.83 (Official build) (64-bit)
 - App version: 1.67.0

PR #2705, opened 2024-01-31
Remove user list from pinned once deleted

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/2659

Checks if `isSaved` or `isPinned` is true, if so, fires a non-blocking mutation to remove the user list from saved/pinned array


pfrazee (approved):
Works great! Thanks Mary!

PR #2706, opened 2024-01-31
Remove superseded strings from en/messages.po

surfdude29:
A cleanup PR that removes strings from en/messages.po that have been superseded and are no longer used.


pfrazee (approved):
Thanks!

PR #2711, opened 2024-01-31
Remove superseded strings from hi/messages.po

surfdude29:
A cleanup PR that removes strings from hi/messages.po that have been superseded and are no longer used.

I also matched some of the superseded strings that had already been translated to their new equivalents.


pfrazee (approved):
Thanks!

Issue #2669, opened 2024-01-29
Language selection in the post editor sometimes fails to open (Web)


pfrazee:
On Web, there are some cases where pressing the language selector fails to open the selector. This is not consistent but I was able to repro it.

PR #2719, opened 2024-01-31
Display the language selection dialog correctly on web

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2669

Not sure when this broke, but the language selection dialog currently doesn't take into account the scroll position of the container, so if we are not at the top of the page it renders somewhere in the void. This uses click/press position instead of the page position for better accuracy.

https://github.com/bluesky-social/social-app/assets/153161762/705ce7bc-4c52-4676-9c77-71b246251e64

### DIFF
--- a/src/view/com/util/forms/Button.tsx
+++ b/src/view/com/util/forms/Button.tsx
@@ -9,14 +9,12 @@ import {
   PressableStateCallbackType,
   ActivityIndicator,
   View,
+  NativeSyntheticEvent,
+  NativeTouchEvent,
 } from 'react-native'
 import {Text} from '../text/Text'
 import {useTheme} from 'lib/ThemeContext'
 import {choose} from 'lib/functions'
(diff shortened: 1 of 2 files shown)


src/view/com/util/forms/Button.tsx
```diff
@@ -9,15 +9,13 @@ import {
...
 } from 'react-native'
 import {Text} from '../text/Text'
 import {useTheme} from 'lib/ThemeContext'
 import {choose} from 'lib/functions'
 
-type Event =
-  | React.MouseEvent<HTMLAnchorElement, MouseEvent>
-  | GestureResponderEvent
-
 export type ButtonType =
```
haileyok, line 19: RN's gesture event type has all of this

pfrazee (approved):
Tested on all platforms, works great, well done

PR #2720, opened 2024-01-31
Fix string in onboarding screens in fr/messages.po

surfdude29:
This is a very small follow-up PR to #2709. I messed up one of the strings in the French translation of the string for number of steps on the onboarding screens, sorry.


pfrazee (approved):
No worries! Thanks!

PR #2721, opened 2024-01-31
Submit App Password on pressing Enter

mary-ext:
Tested on Android and Web, can't test on iOS.

- Fixes the annoying issue with Enter key not submitting the form on web
- Makes `selectTextOnFocus` actually work (it didn't do anything before, my guess is that this was used to select the default name so you can easily change it, so I did just that)

### DIFF
--- a/src/view/com/modals/AddAppPasswords.tsx
+++ b/src/view/com/modals/AddAppPasswords.tsx
@@ -62,7 +62,8 @@ export function Component({}: {}) {
   const {_} = useLingui()
   const {closeModal} = useModalControls()
   const {data: passwords} = useAppPasswordsQuery()
-  const createMutation = useAppPasswordCreateMutation()
+  const {mutateAsync: mutateAppPassword, isPending} =
+    useAppPasswordCreateMutation()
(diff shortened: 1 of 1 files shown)


src/view/com/modals/AddAppPasswords.tsx
```diff
@@ -136,6 +137,17 @@ export function Component({}: {}) {
...
+  const textInputRef = React.useCallback((node: TextInput | null) => {
+    if (!node) {
+      return
+    }
+
+    // `selectTextOnFocus` isn't working with `autoFocus={true}` or without a timeout going.
+    setTimeout(() => {
+      node.focus()
+    }, 0)
+  }, [])
```
pfrazee, line 149: Appreciate what you were going for here, but it's a bit cludgey and didn't work in my testing. I went back to autofocus because that does work on iOS. I think web users will have to just suffer.

pfrazee (approved):
If you dont mind, I'm going to just drop the autofocus behavior. It's just a bit too cludgey and I dont feel like it's

pfrazee:
Thanks Mary!

PR #2722, opened 2024-02-01
Options for selecting dark theme, fix some white flashes when in dark mode

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2595

Adds preference in settings to choose between the default dark theme or the OLED dark theme. To support this, we needed to make some changes (which incidentally will make adding additional themes down the road much easier):

- Create a `dimTheme` theme which uses all of the base tokens from `darkTheme` but sets `backgroundColor` a dim color instead of black
- Both the ALF and the legacy `ThemeProvider` take in the name of a theme now rather than a particular color mode. Options are `light`, `dark`, or `dim`. `dim` is the default dark theme on a new install.
- `useColorMode` and `useSetColorMode` is changed to `useThemePrefs` and `useSetThemePrefs`. Respectively, both contexts have `colorMode, darkTheme` and `setColorMode, setDarkTheme` as their values
- `useColorModeTheme` is changed to return the memoized theme name, and we handle the logic here to determine which theme name to use regardless of the color mode being system/light/dark.
- Small modification to default to a black background when loading the web app and your system color mode is set to dark

Additionally on web since there are some styles inside of the root HTML document, it seems better to change this logic a bit. Instead of relying on `@media` here, we can use use the existing `updateDocument()` from the color mode context to set the particular theme that is being used (light, dark, or dim). Instead of adding/changing the class name inside of the context's `useEffect` however, we can handle that when the `useColorModeTheme` memo updates.

Makes sense to also make sure we are updating the system UI background color in this PR. You'll notice that on some screens when a modal is open that the keyboard in dark mode appears quite bright as if there's something white behind it. However, on Android this is very noticeable because the system animations reveal this white background every now and then.

By using `expo-system-ui` to update the background color, we fix this issue and we no longer get white flashes when opening modals, displaying the keyboard, etc.

### Before:
https://github.com/bluesky-social/social-app/assets/153161762/636bfefe-dcee-4b9f-9ab3-756aae2516f1

### After:
https://github.com/bluesky-social/social-app/assets/153161762/d75ba123-da07-4575-bf6b-bd6e9c2dc359

https://github.com/bluesky-social/social-app/assets/153161762/d264b21a-2845-4f70-8cc7-a78b3272650e


Hima-Zinn:
"Dim" and "Dark" are good name!

src/alf/util/useColorModeTheme.ts
```diff
@@ -1,10 +1,54 @@
...
+      SystemUI.setBackgroundColorAsync(light.atoms.bg.backgroundColor)
+      break
+    case 'dark':
+      SystemUI.setBackgroundColorAsync(dark.atoms.bg.backgroundColor)
+      break
+    case 'dim':
+      SystemUI.setBackgroundColorAsync(dim.atoms.bg.backgroundColor)
+      break
+  }
 }
```
haileyok, line 54: I'm conveniently throwing this fix in as well. We need to update the UI background color as well when the theme is set. You'll notice that on some screens when a modal is open that the keyboard in dark mode appears quite bright as if there's something white behind it. However, on Android this is *very* noticeable because the system animations reveal this white background every now and then.

This fixes that, and we no longer get white flashes when opening modals, displaying the keyboard, etc.

pfrazee (approved):
LGTM

Issue #2487, opened 2024-01-11
YouTube embedded player gives "Video unavailable" on android


pfrazee:
![CleanShot 2024-01-10 at 21 35 55@2x](https://github.com/bluesky-social/social-app/assets/1270099/0a01bb24-c582-4286-b940-ffadad199ccf)

This doesn't seem to happen on Web or iOS so we need to figure out why Android is getting this.

haileyok:
odd, i see the same behavior on ios on this one. if you're not, curious what the distinction is.

regardless, i suspect it's the environment. came across something that mentions needing a referrer header. might be worth trying to set the headers in the request in the webview.

pfrazee:
yeah we should take a look at that youtube embed module we dropped and see if there was something they did

haileyok:
Yea, its the lack of the referrer header. Notice that if you navigate to https://www.youtube.com/embed/h81Ojd3d2rY you get the same result. The previously used library doesn't work if you use the local HTML either, only if you use the GitHub hosted HTML page.

They never want to make it easy. I suppose the solution is indeed to have a hosted file. Twitter does this for example: https://twitter.com/i/cards-frame/postId

pfrazee:
Are we not able to manually set a referrer with the webview api?

haileyok:
You can set headers in the `source` but it seems they don't actually get applied to each request, only the initial one (minimal test to be fair, but setting the exact same headers from the web request didn't seem to help).

PR #2726, opened 2024-02-01
fix some youtube videos not properly loading

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2487

Content now loads that previously would not due to YouTube restrictions. If there's no referrer for the request, it doesn't work. By hosting the iframe on `bsky.app` we can get around this.

Screenshot 2024-02-01 at 1 57 13 PM


pfrazee (approved):
Okay looks good!

PR #2733, opened 2024-02-02
fix scrolling in delete account modal

haileyok:
just adding scroll to the delete account modal and an android adjustment so it presents nicely.

https://github.com/bluesky-social/social-app/assets/153161762/2c0bd7ba-5faf-4ec4-936f-37f6960babf5


pfrazee (approved):
Thanks, that's been bugging me for a while

Issue #2734, opened 2024-02-02
Cannot add a link card for certain links, link card isn't loading 


hildwin:
**Describe the bug**
I try to include a link to the website of the German parliament in a new post, and generate a link card, but it fails. 
I see the message "Please wait for your link card to finish loading", posting is not possible.
Even after waiting a while, I can only cancel. A link card cannot be generated. 

**To Reproduce**

1. Paste the following link into the compose box
https://www.bundestag.de/mediathek/live
2. click add link card

**Expected behavior**
Expected to have a link card generated. 

**Screenshots**
![Screenshot 2024-02-02 103032_waitloading](https://github.com/bluesky-social/social-app/assets/5394190/70089d24-15c6-4664-b58a-b2ed5984d028)

**Details**

 - Platform: Desktop Chrome Version 121.0.6167.139 Windows 10
 - App version: 1.67.0 

**Additional context**

surfdude29:
I wonder, is the behaviour different on the app to the web? I just tried to reproduce it on iOS (1.66) and no image loaded, but I was able to make the post fine with the link card:

https://bsky.app/profile/surfdude29.ispost.ing/post/3kkh5o67bhs2z

hildwin:
> I wonder, is the behaviour different on the app to the web? I just tried to reproduce it on iOS (1.66) and no image loaded, but I was able to make the post fine with the link card:
> 
> https://bsky.app/profile/surfdude29.ispost.ing/post/3kkh5o67bhs2z

I tried to generate a link card on Android (1.66): No image loaded, but I was able to post with link card, too.

hildwin:
Checking the link with one of thousand OG-preview-site 
shows no error. What is the difference with bsky? E. g. https://www.opengraph.xyz/url/https%3A%2F%2Fwww.bundestag.de%2Fmediathek%2Flive

kainsavage:
I am also experiencing this issue with the web client. I tried OP's post and it seems to be working now, but I can request a card for https://blog.teamclerks.net/post/renaissance-man/ and it will never get the image.

I checked documentation and source, but I couldn't find anything suggesting limits to size, aspect ratio, or anything else. Also, I've used cards with webp before that do work, so I know it's not a mime type issue.

I can even see the request in the network tab is trying to get the correct image: 
![image](https://github.com/bluesky-social/social-app/assets/3082056/80f931a2-dba1-4122-8f77-33652244636f)
[...]

PR #2736, opened 2024-02-02
Listen for error event on creating resized image

mary-ext:
Partial fix for https://github.com/bluesky-social/social-app/issues/2734

It returned a broken unparseable image, and because there was no event listener to catch it, the promise is just stuck unfulfilled.

This only addresses the fact that cardyb is returning a broken image, this doesn't fix *why* cardyb is returning a broken image to begin with.


pfrazee (approved):
🙏

PR #2738, opened 2024-02-03
Fix right pane Feeds selector getting out of sync with tabbar

gaearon:
## The Problem

If you open the app with mobile layout, you'll see the current feed in the tab bar:

Screenshot 2024-02-03 at 02 10 58

If you keep expanding the browser window, we'll flip to desktop mode. Notice that it **shows the wrong feed as selected:**

Screenshot 2024-02-03 at 02 11 09

This is because they don't have the same source of truth.

Additionally, if you use the right pane selector, it'll switch you out of the Home feed and will navigate you to the standalone views of those individual feeds. As opposed to showing them in Home like the mobile tab bar does.

## The Solution

I'm making URL the source of truth:

- `/` shows Following (with a clean URL)
- `/?feed=<some feed URL>` shows a specific feed in Home

There is no difference in behavior between mobile and desktop views now. The tab bar always updates the URL. The right pane also always updates the URL. They stay in sync even if you render both at the same time (which we don't currently do but could in theory).

I've kept persistence for native apps, but I've restructured the data flow so that the persistence value flows into the navigator. From that point on, we're reusing the same code on all platforms.

I decided not to persist last selected tab on the web. Web already has a concept of URLs so you could bookmark a specific feed if you wanted to. Persisting the last selection doesn't play very well with that concept — it's weird if you open multiple tabs and they clobber each other's values. On the web, *navigating to* a link usually doesn't imply you want to persist that selection for future tabs. We could revisit this decision though if the difference between web and native is too jarring.

## Test Plan

- On web, click around the tabbar in mobile mode and the right pane in desktop mode.
  - Verify that they stay synchronized within a browser session.
  - Verify they respond to Back/Forward browser buttons in both layouts.
- Verify right pane can be opened in a new tab.
- Verify right pane works outside the Home tab.
- Verify refreshing preserves the selected tab.
- Verify Following always has a clean `/` URL.
- Verify unpinning and/or deleting a feed does not break either selector, and falls back to Following.
- Verify that on mobile, the last selection is still persisted between reloads.

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -431,6 +431,9 @@ const FlatNavigator = () => {
         name="Home"
         getComponent={() => HomeScreen}
         options={{title: title(msg`Home`), requireAuth: true}}
+        initialParams={{
+          selectedPage: 'Following',
+        }}
       />
       <Flat.Screen
         name="Search"
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -314,9 +314,6 @@ function HomeTabNavigator() {
         name="Home"
         getComponent={() => HomeScreen}
         options={{requireAuth: true}}
-        initialParams={{
-          selectedPage: 'Following',
-        }}
       />
       {commonScreens(HomeTab)}
     </HomeTab.Navigator>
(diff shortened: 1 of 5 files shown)


gaearon:
One possible alternative could be to give the Following feed a separate URL, and to redirect `/` to the last opened one.

gaearon:
Gonna approach differently. Can of worms.

Issue #2740, opened 2024-02-03
Appeal content warning option missing if Moderation option set to 'Show'


Bossett:
**Describe the bug**

When moderation settings are configured to 'Show' (instead of 'Warn') on content, it appears that the 'Appeal content warning' option on the post is missing.

**To Reproduce**

1. Set Content Filtering to 'Show' for a label
2. Create a post containing content using that label
3. Note that the 'Appeal content warning' option is not present on the post in the UI
4. Set Content Filtering to 'Warn' for that label
5. Note that the 'Appeal content warning' option is present on the post in the UI

**Expected behavior**

'Appeal content warning' should appear on all labelled posts, regardless of the end user's Content Filtering settings.

**Screenshots**

Hidden content, appeal button appears:
![image](https://github.com/bluesky-social/social-app/assets/114112113/e77f4138-2393-44e1-8b76-46ca539e58b0)

Shown content, appeal button not appearing:
![image](https://github.com/bluesky-social/social-app/assets/114112113/e22ff143-086e-4133-b75d-d123c965dcd1)

**Details**

 - Platform: all
 - Platform version: 
 - App version: 1.67.0

**Additional context**

Investigation in this thread: https://bsky.app/profile/bossett.bsky.social/post/3kkkbz6fcok26

Issue #2286, opened 2023-12-23
"Appeal content warning" is always displayed in post drop down menu on my posts


tkusano:
**Describe the bug**

**To Reproduce**

1. Display one of my posts (posts posted by myself)
2. Select ellipsis and display "drop down menu"
3. "Appeal content warning" item appears even if the post is not labeled.

**Details**

 - Platform: web, iOS (may be on Android)
 - App version: 1.60

**Additional context**

In `src/view/com/post-thread/PostThreadItem.tsx`, `!isSelfLabeledPost` is always `true` when `moderation.decisions.post.cause?.type !== 'label'`, I think.

So,  when a post is not labeled, 

```typescript
<PostDropdownBtn

    showAppealLabelItem={
                post.author.did === currentAccount?.did && !isSelfLabeledPost
   }
```
becomes
```typescript
<PostDropdownBtn

    showAppealLabelItem={
                post.author.did === currentAccount?.did && true
   }
```
.
.

pfrazee:
Yep yep thanks for the issue, good sleuthing too

PR #2550, opened 2024-01-17
3p moderation services [WIP]

pfrazee:
**General**
- [ ] Translations
- [ ] Review filtering behaviors in search
- [ ] Default setting for label concern
- [ ] QPs of blocked showing in feed
- [ ] Takedowns of labels
- [ ] Set the default avatar for labelers in profilecards
- [ ] "Show badge" on impersonation having no effects (see newyorktimes.bsky.social)

**Labelers**
- [ ] Handle labelers that go away
- [ ] Update handling of "tombstones" for takedowns: on content, and on your own account

**Label handling**
- [ ] Apply labels to labelers
- [ ] Apply labels to feedgens
- [ ] Apply labels lists

* * *

**Done:**
- [x] Update SDK docs
- [x] Double check that old native clients and new web client won't bork each other
- [x] Actually send the reports (needs coord with backend)
- [x] Report labelers
- [x] Report feedgens
- [x] Report lists
- [x] Finalize how we refer to labelers
- [x] **Do a design pass on all the dialogs (labels on me, moderation details)**
- [x] **Update the alert/inform pill designs**
- [x] **Adjust quote post spacing between post ctrls**
- [x] **Remove bad focus state from overflow menu**
- [x] **Make sure hidden posts are working**
- [x] **Make sure the adult-disabled behavior for custom labels is hide**
- [x] **Add liked by page to go server**
- [x] **Make sure embeds are driving top level filtering!**
- [x] **Choose a default avatar**
- [x] **Bluesky Safety special-casing in the UI**
- [x] **Fix optimistic likes**
- [x] **Change likes to a screen**
- [x] Make sure mutewords are working
- [x] After setting age, update moderation screen
- [x] Drop the tertiary button color and make the secondary solid color what the tertiary color is
- [x] Date input font is weird on web
- [x] Switch gore to graphic-media
- [x] Create dev-env mod authority
- [x] Update profile=target behaviors
- [x] Move nudity out of adult content choice
- [x] Fix type errors everywhere
- [x] Finalize the global labels
- [x] Rebase
- [x] Add moderation service view screen
- [x] Add callout for moderation services on the profile
- [x] Implement un/subscribe
- [x] Implement un/like
- [x] Update moderation preferences screen to list the user's currently saved moderators
- [x] Update the SDK to support the labeler header and update the agent according to the user's preferences
- [x] Sync label definitions to local storage
- [x] Refine the content-hiding and content-warning UIs
[...]

PR #2741, opened 2024-02-04
always show appeal content label, show notice when post is moderated

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2740
fixes https://github.com/bluesky-social/social-app/issues/2286

Right now, we are not showing any info if a post has been auto-moderated. The only way that someone can see if their post was auto-modded is by checking the dropdown and seeing if the `Appeal Content Warning` option is shown.

We also are not showing the option at all if the user has that particular label set to `Show`, so there's no way for them to know if their post got labelled nor to appeal it without changing that option to `Warn`.

This fixes both problems by doing a little logic on the `PostView` itself.

1. If there are no labels on the `PostView` at all, then it isn't moderated
2. If there are labels on the post but none in the record, then we know it is moderated
3. Finally, if there are labels in both we just check if any of the app view's labels are not on the post record.

Also, since this logic will return false if the post is self labeled, we can add back the `LabelInfo` component to post thread items.

Screenshot 2024-02-03 at 5 08 34 PM
Screenshot 2024-02-03 at 5 09 05 PM
Screenshot 2024-02-03 at 5 09 34 PM

### DIFF
--- a/src/lib/moderatePost_wrapped.ts
+++ b/src/lib/moderatePost_wrapped.ts
@@ -1,8 +1,13 @@
 import {
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
+  AppBskyFeedDefs,
+  AppBskyFeedPost,
   moderatePost,
 } from '@atproto/api'
+interface SelfLabel {
+  val: string
+}
 
 type ModeratePost = typeof moderatePost
 type Options = Parameters<ModeratePost>[1] & {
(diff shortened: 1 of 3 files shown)


haileyok:
What is the plan for 3p-labelers? Are those labels going to be appealable and how do we differentiate between our labels and 3p labels?

haileyok:
What is the plan for 3p-labelers? How do we differentiate between our labels and 3p labels?

estrattonbailey:
@haileyok oh just saw this! Nice one. This will be addressed in #2550 👍 but let's keep it open for a hot sec to confirm

haileyok:
Haha I forgot about this one too. Sounds like the overall fix is better in the other branch anyway 🔥

haileyok:
Yep, closing in favor of 3p mod branch.

PR #2742, opened 2024-02-04
Validate phone number based on country

RonyVidaur:
I signed up today using my Honduran phone number (8 digits) and the app showed me an error saying `here's something wrong with this number. Please choose your country and enter your full phone number!`

Given that international phone numbers are often < 9 characters in length and there's already a `parsePhoneNumber` included from `libphonenumber-js` I think it's better to use the built-in validator.


pfrazee (approved):
Good call, sorry for the error!

PR #2743, opened 2024-02-04
Support previews for links using DIDs and not just handles

ericvolp12:
The current template rendering logic for OpenGraph data only supports handles in the URL, this should update it to support DIDs as well for previewing profiles and posts.


Jacob2161 (approved):
:+1:

PR #2744, opened 2024-02-04
update docs to include info on installing `cocoapods`

haileyok:
I've seen this come up three times in the past week, and it isn't documented in the React Native docs. On macOS we need to make sure we are running ruby >= 2.7.6 (macos ships with 2.6) and a downgraded version of cocoapods which gets installed through the project's gemfile.


pfrazee (approved):
Cool sg

PR #2748, opened 2024-02-05
Show original query in error message when using "from:me"

mozzius:
Currently, the error message shows the query *after* it's been transformed, resulting in the DID being visible in the error message:

Screenshot 2024-02-05 at 09 58 55

It seems like there's a general aim to avoid DIDs being surfaced in the UI where possible, so I don't think this behaviour is ideal.

This PR shows the original query in the error message, only using the augmented query for the actual seach request.

Screenshot 2024-02-05 at 09 58 32


pfrazee (approved):
Nice yeah many thanks

Issue #2745, opened 2024-02-04
Under certain conditions, link in Moderation screen cannot be tapped


surfdude29:
**Describe the bug**
Under certain combinations of device screen size, text size setting and app language setting, the "Learn more about what is public on Bluesky." link at the bottom of the Moderation screen cannot be tapped.

**To Reproduce**

1. Use a device with a screen size at the smaller end of the spectrum, such as iPhone 7, iPhone 8 or iPhone SE (1st, 2nd or 3rd generation).
2. Set Text Size to the default setting or larger.
3. Open the Moderation screen, scroll down and try to tap the "Learn more about what is public on Bluesky." link at the bottom.
4. You cannot tap the link.

**Expected behavior**

The user should be able to tap the link.

**Screenshots**

https://github.com/bluesky-social/social-app/assets/149612116/4d34e10f-84b9-42a3-8fec-d3fe1ef24b0c

**Details**

 - Platform: iOS
 - Platform version: 15.8.1
 - App version: 1.66

**Additional context**

When some languages other than English are set as the app language, this occurs on my iPhone 7 even when Text Size is set at the smallest setting.

surfdude29:
This was fixed by #2749, thanks! :)

PR #2749, opened 2024-02-05
link in Moderation screen cannot be tapped #2745

Pranith-Kanchan:
link in Moderation screen cannot be tapped #2745


pfrazee (approved):
🙏 🙏 🙏 🙏 🙏

PR #2752, opened 2024-02-05
use ios live text interaction in alt text modal and image viewer

haileyok:
It is pretty common on iOS now for images to use iOS's live text interaction feature. We can enable that in the alt text modal and in the image viewer (when the image is not scaled) with expo-image.

https://github.com/bluesky-social/social-app/assets/153161762/8733507d-de3d-447a-9485-82a0e30963ef


pfrazee (approved):
Sick

PR #2753, opened 2024-02-06
Add missing word in ModerationDetails.tsx

surfdude29:
This very small PR adds the missing word 'in' to the message:

"This user is included the [listname] list which you have muted."

to follow the equivalent message on L43:

"This user is included **in** the [listname] list which you have blocked."


pfrazee (approved):
Oof thanks

Issue #2728, opened 2024-02-01
Thread screen jumping around on android


pfrazee:
A user reported that large threads have a habit of jumping around on them (see https://bsky.app/profile/trisalvagni.bsky.social/post/3kkehebyx7e2g)

It's possible that this is a bug in how the scroll position is maintained on load.

mozzius:
I had this at the bottom of a very long thread on iOS too, I don't think it's Android-specific. I'll try and find the thread and replicate it

mozzius:
Not sure if this is the same bug as what the other person was sharing, but this is the thread: https://bsky.app/profile/mozzius.dev/post/3kjqgddalf22e

https://github.com/bluesky-social/social-app/assets/10959775/4193e68c-840e-431a-8a0a-d2e50732e66d

gaearon:
We've fixed / worked around _some_ of these cases.

Here's the latest on known jumps by type of post.

```
[OK] Timeline -> Text
[OK] Timeline -> QT of Text
[OK] Timeline -> Link
[OK] Timeline -> Image
[OK] Timeline -> QT of QT of Text
[JUMP] Timeline -> QT of Link
[JUMP] Timeline -> QT of Image
[OK] Timeline -> QT of QT of QT of Text
[OK] Timeline -> QT of QT of Link
[JUMP] Timeline -> QT of QT of Image
[JUMP] Timeline -> YouTube
[JUMP] Timeline -> QT of YouTube
[OK] Timeline -> QT of QT of YouTube
```

Other known issues:

- Posts with no likes/RTs jump
  - When navigated from timeline (but we have the metrics, so should be fixable on front-end)
  - When navigated via QT (we don't have the metrics here, might need backend help)

Some unexplained problems:

- Why does [this post](https://bsky.app/profile/haileytesting10.bsky.social/post/3klsh75p4h52q) jump
- Why is `QT of QT of Link` OK but `QT of QT of Image` jumps (aren't both embeds)
- Why doesn't [this post](https://bsky.app/profile/dansshadow.bsky.social/post/3klsh3rezsl2i) jump on native despite YouTube above?

gaearon:
I'm closing this issue as we've resolved most of the related issues (on main — pending the release). There's still more to do for quote posts, but I've opened https://github.com/bluesky-social/social-app/issues/2946 to track that.

Thanks everyone!

PR #2755, opened 2024-02-06
Improve rendering behaviors in PostThread

haileyok:
fixes https://github.com/bluesky-social/social-app/issues/2728#issue-2113566343

## Problem

We are experiencing a few different issues right now when rendering post threads, especially ones that have a large number of posts:

- After content has loaded, there are jumps in the scroll position after the parents or replies have loaded
- If our data is cached, we completely lose the scroll position whenever there is a large number of items
- In threads with an excessive parent count, we are experiencing performance issues and jumpiness when scrolling up
- In those same threads, when scrolling down, `FlatList` oddly tries to render the items at the top, causing extreme jankiness and bouncing for a few seconds

These problems stem from two different issues:

1. We are currently using `onContentSizeChanged` to set our scroll position after rendering items if the data is cached. This is necessary with our current implementation because `maintainVisibleContentPosition` requires some index to "latch onto". However, if the data is cached, all of the posts including the parents are rendered *immediately*, so we never get that initial position to latch onto. We can make some adjustments to our `onContentSizeChanged` logic, but since this runs *after* data is rendered, we often see a "jump" to the correct position, which is not ideal.
2. Sometimes, like in the case of opening a post from a quote embed, we don't have all the needed data to render the post. Regardless we display the info we do have while the rest of the post data loads. This causes an awkward resize once we get the metrics and embeds (this is especially bad with embeds!)
3. `maintainVisibleContentPosition` seems to have a few issues, discussed more in detail elsewhere, but here is the repro for the issue: https://snack.expo.dev/@haileyok/flatlist-repro

https://github.com/bluesky-social/social-app/assets/153161762/218c4f12-6199-491e-9312-de32545d4d34

This particular thread illustrates the problem: https://bsky.app/profile/mozzius.dev/post/3kjqgddalf22e

https://github.com/bluesky-social/social-app/assets/153161762/bd11cf84-8e4c-450f-b88f-d3e67151e989

## Solution

First of all, we do not change any logic on web. We continue to use `onContentSizeChanged` and 

Both of these issues are solved first by changing how we render the post thread.

1. Get our post data (whether cached or not)
2. Create `ThreadSkeletonParts` which consists of `parents`, `highlightedPost`, and `replies` (for this, we split `flattenThreadSkeleton` into two parts, the parents and the replies)
4. On native, initially render only the `highlightedPost`. This should render *immediately* whenever the highlighted post is cached (i.e. coming from the feed).
5. After it has rendered, we next add the parents and the replies to the `FlatList`'s data. Because we initially rendered *only* the `highlightedPost`, we now have an item to latch to and `maintainVisibleContentPosition` works in all cases.

Now, because of the `maintainVisibleContentPosition` issue, we have these options:

- Increase the `maxToRenderPerBatch` value. This isn't ideal, as if we increase this render times will increase. Additionally, this is *not* a guarantee to work, unless we set this to some extremely high value (obviously not good).
- Prepend batches of 10 posts to the start of our data array whenever `onStartReached` is called by the `FlatList`.

The latter option is the option used in some other apps including Twitter, and is what we use here. Ideally, I'd like to not have to use this, but until we can successfuly dejank `maintainVisibleContentPosition`, this is the most elegant way to fix both the jumpy behavior when scrolling up and the extreme jumps seen when scrolling down in a thread with excessive parents.

Result:

https://github.com/bluesky-social/social-app/assets/153161762/ba49de44-dc45-4741-b1d3-80f0ab8944b2

Before/After:

https://github.com/bluesky-social/social-app/assets/153161762/462c669e-9cab-4e59-a127-34a475c9939e


src/state/queries/post-thread.ts
```diff
@@ -81,6 +82,12 @@ export function usePostThreadQuery(uri: string | undefined) {
       if (!uri) {
         return undefined
       }
+      {
+        const item = findPostInPostQueryData(queryClient, uri)
+        if (item) {
+          return postViewToPlaceholderThread(item)
+        }
+      }
```
haileyok, line 90: This will only happen if we have precached the quote's post.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -365,20 +414,29 @@ function PostThreadLoaded({
...
+      maintainVisibleContentPosition={
+        isWeb ? undefined : MAINTAIN_VISIBLE_CONTENT_POSITION
+      }
+      ListHeaderComponent={
+        isTabletOrMobile ? (
+          <ViewHeader
+            title={_(msg({message: `Post`, context: 'description'}))}
+          />
+        ) : undefined
+      }
```
haileyok, line 435: I'm moving the header render to the `ListHeaderComponent` so that we can simplify the logic in slicing the parents array. It also prevents some layout issues when we have to load more items when we scroll up. The behavior seems to be identical to what we currently have.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -163,42 +167,69 @@ function PostThreadLoaded({
...
+
+    // HACK
+    // This lets us delay rendering of the additional items in the flatlist for a little while so we can latch onto the
+    // correct post
+    if (items.highlightedPost && !readyToRender) {
+      setTimeout(() => {
+        setReadyToRender(true)
+      }, 300)
+    }
+
```
haileyok, line 210: This was the best way I could think of to first render the highlight and then render the rest of the posts, but it feels strange. Other ideas?

src/view/com/post-thread/PostThread.tsx
haileyok: Nevermind, it makes more sense to do this how we did before.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -163,42 +169,81 @@ function PostThreadLoaded({
...
+    if (postsToPrepend.length > 0) {
+      // If there are any parent posts, we need to add the header here
+      postsToPrepend = [TOP_COMPONENT, ...postsToPrepend]
+    } else if (
+      postsToPrepend.length === 0 &&
+      !highlightedPost?.ctx.isParentLoading
+    ) {
+      // If there are no parents and there isn't a parent loading, then we just add the top component
+      postsToPrepend = [TOP_COMPONENT]
+    }
```
haileyok, line 199: We need to handle the logic of adding the view header here now, since we are continually loading in posts at the top.

haileyok:
Another consideration to make before we merge this is the increased number of requests to `getProfile()` that comes with it. From a UX side this is definitely something we need to do so:

1. We should get the necessary info we need in the response (uri resolutions)
2. At a minimum, this should be fixed https://github.com/bluesky-social/atproto/issues/2152 so that we can batch request's in groups of 25 (as is the case with `getPosts()`)

haileyok:
Went over this tonight and scrolled through a few feeds. I'm seeing that with our existing logic, we should get a spinner whenever we don't have all the data we need. I think I'm okay with that behavior, this is uncommon and should only happen on some quote posts that have embeds. In any case, there are no more jumps in those feeds that I can find. We might be able to get backend to add some more info for us in the quote info so we don't need to make any calls, but I think we might be okay just letting these spinners happen.

src/view/com/post-thread/PostThread.tsx
haileyok: @gaearon I think this is the only remaining quirk to think about. This works perfectly (visually) but curious if there's a better option.

src/state/queries/post-thread.ts
haileyok: Dropped this.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -163,42 +169,75 @@ function PostThreadLoaded({
     [threadViewPrefs, thread],
   )
 
-  // construct content
+  // On web we always are ready to render, we only need this on mobile.
+  const [readyToShowAll, setReadyToShowAll] = React.useState(isWeb)
+  const [topPageCount, setTopPageCount] = React.useState(1)
+  const isPrepending = React.useRef(false)
+
```
haileyok, line 176: `readyToShowAll` gets set to true after we have rendered the first item. We need to wait a moment before setting this to true, hence the 300ms `setTimeout`.

`topPageCount` is just us keeping track of how many pages we have displayed at the top (15 per page)

`isPrepending` ensures we do not run the `onStartReached` callback too many times.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -163,42 +169,75 @@ function PostThreadLoaded({
...
+      : undefined
+
+    // Create the posts to prepend
+    const postsToPrepend = [
+      ...(!highlightedPost?.ctx.isParentLoading ? [TOP_COMPONENT] : []),
+      ...(!isWeb
+        ? [...items.parents.slice(-(topPageCount * 15))]
+        : [...items.parents]),
+    ]
+
```
haileyok, line 196: Posts to prepend has these possibilities:

1. We have the highlighted post, and we see the highlighted post has no parents. Therefore, we add the TOP_COMPONENT here.
2. We have the highlighted post, and we see that there is more posts that will be added above, so we don't add the TOP_COMPONENT and let the generator add it instead.
3. If we are on web, we just render every child. There is no reason to simulate the load on web.
4. On native, we add 15 posts per page. Each time the `onStartReached` callback is fired, `topPageCount` will be updated if necessary.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -215,24 +254,52 @@ function PostThreadLoaded({
...
+    // - There are no more parents to show
+    // - The top post is the highlighted post
+    if (
+      isPrepending.current ||
+      !isThreadPost(first) ||
+      !first.parent ||
+      first.ctx.isHighlightedPost
+    ) {
+      return
+    }
```
haileyok, line 290: 1. We don't want to update `topPageCount` multiple times (`onStartReached` doesn't get debounced, so it's possible it will fire multiple times once the user reaches the top). So, we use a ref to keep track of whether we're awaiting new items to render. Once they do, we will set the ref back to false.
2. `isThreadPost` just gives us the type we need.
3. `!first.parent` checks if there is anything else for us to add. If there isn't then we don't need to do anything and can just return.
4. Similar to `!first.parent`, I think we can remove this now that I think about it.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -215,24 +254,52 @@ function PostThreadLoaded({
...
+    // Start prepending
+    isPrepending.current = true
+    // We wait a moment both to appear like a "load" event and to let the scroll "settle". 850ms is the sweet spot
+    // that we should be confident the scroll has settled at. Lower is slightly janky, higher is too long.
+    setTimeout(() => {
+      // Increment the top page count and set prepending to false
+      setTopPageCount(prev => prev + 1)
+      isPrepending.current = false
+    }, 850)
+  }, [posts])
```
haileyok, line 301: Because the animation needs to settle, the 850 ms does that. I'm going to see if we can instead use `onMomentumScrollEnd` to run the `setState()` so we can ensure reliability.

haileyok: Oh, I stand corrected. There's an `onScrollAnimationEnd` that I think is even better. Checking now.

haileyok: Oh, weird. There's a typehint for `onScrollAnimationEnd` but it isn't documented (and doesn't do anything). Is this something we added?

Anyway, `onMomentumScrollEnd` does what we want.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -215,24 +256,53 @@ function PostThreadLoaded({
...
+    isPrepending.current = true
+  }, [posts])
+
+  // We want the scroll animation to settle before we increment the page count. We will check every time the scroll
+  // animation ends and prepend posts once settled if necessary.
+  const onMomentumScrollEnd = React.useCallback(() => {
+    if (isPrepending.current) {
+      setTopPageCount(prev => prev + 1)
+    }
+  }, [])
```
haileyok, line 304: This works well to make sure the scroll animation is completed before prepending. We can't run all of the logic here though, because we don't have any information about the scroll offset. So instead we rely on the ref that gets updated inside of `onStartReached`

gaearon:
Merged via https://github.com/bluesky-social/social-app/pull/2862, https://github.com/bluesky-social/social-app/pull/2865, https://github.com/bluesky-social/social-app/pull/2878, and https://github.com/bluesky-social/social-app/pull/2929.

PR #2758, opened 2024-02-06
European (Danish) phone numbers often just 8 digits.

melpomene:
When not counting the country code the number is most often 8 digits in Denmark.


surfdude29:
It seems to affect some users outside of Europe too, as in PR #2742 .

pfrazee:
Yeah I think this has been solved best with #2742. Thank you!

PR #2765, opened 2024-02-06
Use new persistence API for selected feed

gaearon:
This isn't strictly necessary but it's a bit simpler.

Verified persistence still works. (Ofc it would reset once due to a different key.)


gaearon:
Gonna roll this into a bigger related change.

PR #2766, opened 2024-02-06
Variable translation broke the number of "following" in pt-BR

maisondasilva:
Variable translation broke the number of "following" in pt-BR
I believe need to be more careful when translating to avoid variables being translated, breaking the website experience!

Imagem


maisondasilva:
@pfrazee @ansh

src/locale/locales/pt-BR/messages.po
```diff
@@ -1,6 +1,6 @@
 msgid ""
 msgstr ""
-"POT-Creation-Date: 2023-12-22 01:56+0530\n"
+"POT-Creation-Date: 2024-02-07 00:25+0530\n"
```
surfdude29, line 3: Sorry to jump in, I think you meant to update `PO-Revision-Date` instead of the creation date?

maisondasilva: Thanks for the warning, for some reason the editor I use changed the wrong line, I already corrected it.

surfdude29: No worries :)

pfrazee (approved):
🙏

PR #2768, opened 2024-02-07
Refactor Home feed pager rendering logic

gaearon:
I found the existing logic difficult to trace due to the source of truth being slightly ambiguous. We have several types of information (selected index, a list of pinned feeds, fetched/resolved pinned feeds, feed descriptors) and the relationship between them wasn't very clear. In particular, we weren't very careful about which indexes correspond to which arrays. This is why the code was written defensively (e.g. clamping) and a bit unnaturally.

In this PR, I restructure it so that there is always an `allFeeds` array of resolved `FeedDescriptor` values (including the first `'home'` one), the `selectedIndex` always points to an item in that array (and is guaranteed to be in-bounds, i.e. no `-1`), and so that `selectedFeed` is always `allFeeds[selectedIndex]`. If the persisted value is not in the array, it is effectively discarded — we fall back to `'home'` (with index `0`).

## Test Plan

- Reload to verify persistence works
- Pin or unpin unselected feed, verify pager updates and we stay on the same page
- Unpin selected feed, verify we fall back to Following
- Verify gestures work on mobile

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -16,9 +16,9 @@ import {usePinnedFeedsInfos, FeedSourceInfo} from '#/state/queries/feed'
 import {UsePreferencesQueryResponse} from '#/state/queries/preferences/types'
 import {emitSoftReset} from '#/state/events'
 import {useSession} from '#/state/session'
-import {loadString, saveString} from '#/lib/storage'
 import {useWebMediaQueries} from '#/lib/hooks/useWebMediaQueries'
 import {clamp} from '#/lib/numbers'
+import * as persisted from '#/state/persisted'
 
 type Props = NativeStackScreenProps<HomeTabNavigatorParams, 'Home'>
 export function HomeScreen(props: Props) {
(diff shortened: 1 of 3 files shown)


src/view/screens/Home.tsx
```diff
@@ -16,45 +16,25 @@ import {usePinnedFeedsInfos, FeedSourceInfo} from '#/state/queries/feed'
...
-  ) {
+  if (preferences && pinnedFeedInfos && !isPinnedFeedsLoading) {
     return (
       <HomeScreenReady
         {...props}
         preferences={preferences}
-        pinnedFeeds={pinnedFeeds}
-        initialPage={isDesktop ? 'Following' : initialPage}
+        pinnedFeedInfos={pinnedFeedInfos}
+        rawInitialFeed={isDesktop ? 'home' : rawInitialFeed}
```
gaearon, line 37: I'm sticking with `FeedGenerator` format as the main key/identifier/"data structure" here so I switched to the `FeedGenerator` value we use for the first `<FeedPage>`.

src/view/screens/Home.tsx
```diff
@@ -68,77 +48,54 @@ export function HomeScreen(props: Props) {
...
-  const [selectedPage, setSelectedPage] = React.useState<string>(initialPage)
-
-  /**
-   * Used to ensure that we re-compute `customFeeds` AND force a re-render of
-   * the pager with the new order of feeds.
-   */
-  const pinnedFeedOrderKey = JSON.stringify(preferences.feeds.pinned)
-
-  const selectedPageIndex = React.useMemo(() => {
-    const index = ['Following', ...preferences.feeds.pinned].indexOf(
```
gaearon, line 90: It's possible this code might have been the cause of some cases where the wrong feed was being displayed. This logic searches for the selected index based on the `preferences.feeds.pinned` array. However, rendering pages was done using the `customFeeds` array (derived from `pinnedFeeds`), and their items aren't guaranteed to be in sync. So before the refactor, I think it was possible for the index to correspond to a wrong item. Which is also why clamping below was necessary.

pfrazee (approved):
LGTM. Tested on all platforms

PR #2770, opened 2024-02-07
convert query to lowercase before performing query

haileyok:
We needed to convert the query to lowercase here before searching. Wondering if this was a v2 change?

We should do this regardless so that searching `KATIE` then a subsequently searching for `Katie` will be cached.

https://github.com/bluesky-social/social-app/issues/2430#issuecomment-1931046814


gaearon:
Seems like it would be good if backend handled this instead (rather than every client). Any reason for it not to?

gaearon (approved):

pfrazee:
> Seems like it would be good if backend handled this instead (rather than every client). Any reason for it not to?

nope just need to ping em

PR #2772, opened 2024-02-07
Update some styles

quiple:
Sets `paddingBottom` on `End of feed` and remove `s.mr5`, `s.ml5` on `and` in `FeedItem`.

Replace `margin` with `space` wherever it feels more natural than `margin`.

Before:

After:

---

Before:

After:

---

Before:

After:

---

Before:
Screenshot 2024-02-08 at 10-08-33 Bluesky
Only `TextLink` has a bigger font size.

After:
Screenshot 2024-02-08 at 10-08-39 quiple 테스트 — Bluesky
Same font size.

---

Before (an issue with Firefox only on narrow screens):
Screenshot 2024-02-14 at 16-33-21 Search — Bluesky

After:
Screenshot 2024-02-14 at 16-33-32 Search — Bluesky

---

Before:
Screenshot 2024-02-15 at 13-36-23 quiple (@quiple dev) — Bluesky

After:
Screenshot 2024-02-15 at 13-36-47 quiple (@quiple dev) — Bluesky
Increases the size of reply icon slightly because I think it's too small.

---

Before:
Screenshot 2024-02-16 at 22-24-52 홈 — Bluesky

After:
Screenshot 2024-02-16 at 22-24-17 홈 — Bluesky

---

Before:
Screenshot 2024-02-17 at 03-38-03 Bluesky

After:
Screenshot 2024-02-17 at 03-37-36 피드 — Bluesky

---

Removes this top border.
스크린샷 2024-02-12 14 12 25

---

Set the `paddingHorizontal` of the Feeds header to 18, just like the headers on other pages, such as Settings, Notifications, etc.
Screenshot 2024-02-17 at 03-40-56 피드 — Bluesky

Thank you!


src/view/com/notifications/FeedItem.tsx
```diff
@@ -231,10 +231,9 @@ let FeedItem = ({
             />
             {authors.length > 1 ? (
               <>
-                <Text style={[pal.text, s.mr5, s.ml5]}>
-                  {' '}
-                  <Trans>and</Trans>{' '}
-                </Text>
+                <Trans>
+                  <Text style={[pal.text]}> and </Text>
+                </Trans>
```
pfrazee, line 236: is this one necessary? it will throw off existing translations and force the translations to include the surrounding spaces, which we try to avoid

src/view/com/util/moderation/ScreenHider.tsx
```diff
@@ -85,10 +84,12 @@ export function ScreenHider({
...
-            <Trans>This {screenDescription} has been flagged:</Trans>
-            <Text type="2xl-medium" style={[pal.text, s.ml5]}>
-              {desc.name}.
-            </Text>
+            <Trans>
+              This {screenDescription} has been flagged:{' '}
+              <Text type="2xl-medium" style={[pal.text]}>
+                {desc.name}.{' '}
+              </Text>
+            </Trans>
```
pfrazee, line 92: Same question here

pfrazee (commented):
Thanks so much @quiple ! A couple of questions

quiple:
@pfrazee
> In Japanese, there is no need for spaces before and after "and", so I think it would be a good idea to correct it that way.

As @tkusano mentioned, I included spaces in the localization string because some non-spaced languages, such as Japanese, require spaces to be removed.

Thanks for review!

pfrazee:
> @pfrazee
> 
> > In Japanese, there is no need for spaces before and after "and", so I think it would be a good idea to correct it that way.
> 
> As @tkusano [mentioned](https://github.com/bluesky-social/social-app/pull/2462#discussion_r1457443034), I included spaces in the localization string because some non-spaced languages, such as Japanese, require spaces to be removed.
> 
> Thanks for review!

Oh dear. Okay I'm going to have to give this a closer look then.

PR #2775, opened 2024-02-07
Keep pager feeds in sync with the right pane

gaearon:
Another approach to https://github.com/bluesky-social/social-app/pull/2738.

This one is a bit weird but I think it works.

The idea is to have a hierarchy:

- If opened with `/?feed=...`, prefer that. This is for "open in new tab" in the right pane.
- Otherwise, see if we have something in `sessionStorage`. This is tab-local so it ensures refreshing a tab doesn't reset the selected feed. Each browser tab tracks its own selected feed across refreshes.
- If we don't have that either, fall back to `persisted` (i.e. `localStorage`). This is for regular newly opened tabs.
- Finally, fall back to Following.

On native, there shouldn't be any changes.

I haven't tested extensively, but cursory testing seems to show this works.

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -16,25 +16,19 @@ import {usePinnedFeedsInfos, FeedSourceInfo} from '#/state/queries/feed'
 import {UsePreferencesQueryResponse} from '#/state/queries/preferences/types'
 import {emitSoftReset} from '#/state/events'
 import {useSession} from '#/state/session'
-import {useWebMediaQueries} from '#/lib/hooks/useWebMediaQueries'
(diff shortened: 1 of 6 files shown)


src/view/screens/Home.tsx
```diff
@@ -93,7 +97,7 @@ function HomeScreenReady({
       setDrawerSwipeDisabled(index > 0)
       const feed = allFeeds[index]
       setSelectedFeed(feed)
-      persisted.write('lastSelectedHomeFeed', feed)
```
gaearon, line 96: Note: this still happens but is now taken care by `setSelectedFeed` from the provider.

PR #2776, opened 2024-02-07
Quickly fix Japanese mistranslation of "Suggestive"

Jessidhia:
The word used in the previous translation ([提案](https://dictionary.goo.ne.jp/word/en/提案/)) means "an offer", and has not at all the meaning it needs to have. Switch to [色気](https://dictionary.goo.ne.jp/word/en/色気/)[^1] which is specific to "sex appeal".

[^1]: don't be misled by the first couple example sentences as they show the word being used in the *negative*.

### DIFF
--- a/src/locale/locales/ja/messages.po
+++ b/src/locale/locales/ja/messages.po
@@ -3837,7 +3837,7 @@ msgstr "あなたへのおすすめ"
 
 #: src/view/com/modals/SelfLabel.tsx:95
 msgid "Suggestive"
-msgstr "提案"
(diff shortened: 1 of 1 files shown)


Jessidhia:
(soapbox) This might be related to why the Bluesky mods are permanently banning so many Japanese illustrators, even when what they are uploading is not sexual content at all.

Jessidhia:
https://github.com/bluesky-social/social-app/pull/2777#issuecomment-1931733207 also addresses this, with likely better wording

PR #2780, opened 2024-02-07
Update zh-CN localization to 1.68

auroursa:
Translated the Chinese (zh-CN) to version 1.68, and made some adjustments to certain translated terms.


mitian233:
Referring to Twitter's translation, translate 'mute' into “隐藏” is better here.

auroursa:
I agree with you.

pfrazee (approved):
Thanks all!

PR #2784, opened 2024-02-07
Fix google translate appear

liberaldev:
screenshot 2024-02-07 234059

On non-English Windows Chrome, Google Translate appears as shown in the image above. This PR fixes this. Also, we've removed the X-UA-Compatible code as IE support has now ended.


pfrazee (approved):
This works for me -- thanks!

Issue #2756, opened 2024-02-06
Temporary suspension takedown message is not showing properly in app


surfdude29:
**Describe the bug**

The message saying that an account has been temporarily suspended is not always displaying properly in the app.

**To Reproduce**

1. Open the URL of the profile of a user who has been temporarily suspended, such as: https://bsky.app/profile/did:plc:queiogp4cavuzrnq4k3hxh6t
2. See a mangled error page with no visible text.

**Expected behavior**

See the correctly rendered error page as it is shown on the web (including mobile web).

**Screenshots**

https://github.com/bluesky-social/social-app/assets/149612116/794e6a64-3b5e-42ba-ae3e-50596524a08e

**Details**

 - Platform: iOS
 - Platform version: 15.8.1
 - App version: 1.67

**Additional context**

At least two other users have also run into this issue, reported here:

https://bsky.app/profile/forcolumb.us/post/3kkpp4lola22t

PR #2789, opened 2024-02-07
Fix error screen on native, use `Not Found` for profile errors instead of `Oops!`

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/2756

Profile errors were not rendering correctly since they were accidentally wrapped inside of an additional `<CenterView>`. Also adding a view header on the error screen when rendered in `Profile` (still no header in `ProfilePreview`).

### Before

Screenshot 2024-02-07 at 2 06 02 PM

### After

Screenshot 2024-02-07 at 2 06 12 PM

Also using `Not Found` for profile errors found in https://github.com/bluesky-social/atproto/blob/77118a66a7a5703e495464186246fb29fa9fa9df/packages/bsky/src/api/app/bsky/actor/getProfile.ts

### DIFF
--- a/src/view/com/modals/ProfilePreview.tsx
+++ b/src/view/com/modals/ProfilePreview.tsx
@@ -46,7 +46,7 @@ export function Component({did}: {did: string}) {
   if (profileError) {
     return (
       <ErrorScreen
-        title={_(msg`Oops!`)}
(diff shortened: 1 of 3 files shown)


pfrazee (approved):
Muchos muchos gracias

PR #2790, opened 2024-02-07
Remove profile preview modal

gaearon:
This was disabled on Android and isn't used on web so it was iOS-only. It seems more trouble than it's worth. This lets us delete some hacks with the bottom sheet and, importantly, gives us back a consistent push/pop animation on iOS and a consistent Back navigation behavior.

One thing we should do in a follow-up is to precache the display name and avatar so that they're available synchronously during the push. I'm not doing this in this PR though.


pfrazee (approved):
👋 So long, preview modal!

PR #2795, opened 2024-02-08
Precache basic profile from posts for instant future navigations

haileyok:
## Problem

Currently, we fetch profile data for some - but not all - posts in the timeline (currently it is only the "main" post, not the parent it's in response to). We need to be caching this info for all posts. 

The next issue is that we are unnecessarily making calls to `resolveHandle` to resolve DIDs when pushing to posts. This behavior isn't great, because now we have to first resolve the DID and *then* load the thread. This is the main cause of the "full spinners"  we get when opening posts. We also lack enough data to present anything on the profile screen in many cases.

### 1. Cache the `ProfileViewBasic` (`PostView`'s author)

We start by taking the author from each `PostView` in the feed (including parents and authors in quotes) and adding them to the cache with a query key of `['profileBasic', handle]`. These won't ever get prefetched, since there isn't even a query function for them.

Also adding quote post caching for FlatList changes.

### 2. Add initial data to `useResolveDidQuery()`

We maintain the same behavior of resolving the handle when necessary (i.e. if we just open a new tab, we wont' have anything so we do need to resolve in that case).

What we are changing is setting the `initialData` on the query by checking the query cache for an existing `ProfileViewBasic` item. If we find one, we set the initial data to the DID. `initialData` unlike `placeholderData` updates the cache and blocks any query from being done, so there's no `resolveHandle()` call.

### 3. Add placeholder data to `useProfileQuery()`

Similar to `useResolveDidQuery()`, here we query the cache using `getQueriesData()` to find any existing `ProfileViewBasic` with the DID we are looking up.

This gives us a bit of data for the shell whenever we open up a profile.

https://github.com/bluesky-social/social-app/assets/153161762/b98c6131-d0d3-40fe-ad04-10a2521ccf74


haileyok (commented):
We can revert this if we want, but it's a good test case for showing what we'll end up with in the profile shell.

src/state/queries/profile.ts
```diff
@@ -25,28 +25,30 @@ import {STALE} from '#/state/queries'
...
-    // this staleTime is load-bearing
-    // if you remove it, the UI infinite-loops
-    // -prf
-    staleTime,
+export function useProfileQuery({did}: {did: string | undefined}) {
+  const queryClient = useQueryClient()
+  // TODO Figure out a good staleTime for this. We should refetch on every profile push, because we need to check for
+  // blocks
+  return useQuery<AppBskyActorDefs.ProfileView>({
+    staleTime: STALE.SECONDS.FIFTEEN,
```
haileyok, line 36: We should figure out what we really want to do with this. Is the removed comment still applicable after these changes?

src/state/queries/resolve-uri.ts
```diff
@@ -22,40 +35,114 @@ export function useResolveUriQuery(uri: string | undefined): UriUseQueryResult {
...
+    if (AppBskyEmbedRecord.isViewRecord(embed?.record)) {
+      precacheProfile(queryClient, embed.record.author)
+    } else if (
+      AppBskyEmbedRecordWithMedia.isView(embed) &&
+      AppBskyEmbedRecord.isViewRecord(embed.record.record)
+    ) {
+      precacheProfile(queryClient, embed.record.record.author)
+      quoteEmbedUris.push(embed.record.record.uri as string)
+    }
+  }
```
haileyok, line 127: This isn't applicable yet, but will be with FlatList changes

src/state/queries/profile.ts
```diff
@@ -25,28 +25,30 @@ import {STALE} from '#/state/queries'
 import {track} from '#/lib/analytics/analytics'
 
 export const RQKEY = (did: string) => ['profile', did]
+export const profileBasicKey = (handle: string) => ['profileBasic', handle]
```
haileyok, line 28: We can technically use the DID for the query key here, but since the majority of our look ups are going to be using the handle, I think this makes more sense. When we use this as placeholder data for `useProfileQuery()` we can use `getQueriesData()` with a `.find()` to get the profile using the DID.

src/state/queries/profile.ts
```diff
@@ -25,28 +25,30 @@ import {STALE} from '#/state/queries'
...
-    // this staleTime is load-bearing
-    // if you remove it, the UI infinite-loops
-    // -prf
-    staleTime,
+export function useProfileQuery({did}: {did: string | undefined}) {
+  const queryClient = useQueryClient()
+  // TODO Figure out a good staleTime for this. We should refetch on every profile push, because we need to check for
+  // blocks
+  return useQuery<AppBskyActorDefs.ProfileView>({
+    staleTime: STALE.SECONDS.FIFTEEN,
```
haileyok, line 36: We should figure out what we want to do with this `staleTime`. Is the removed comment still applicable after these changes?

src/state/queries/profile.ts
haileyok: Caching twice instead.

src/view/screens/Profile.tsx
```diff
@@ -84,12 +85,18 @@ export function ProfileScreen({route}: Props) {
...
+        ? moderateProfile(profile, moderationOpts)
+        : null
+
     return (
       <CenteredView>
         <ProfileHeader
-          profile={null}
-          moderation={null}
+          profile={profile}
+          moderation={moderation}
```
haileyok, line 99: Instead of waiting for `moderationOpts` to render, just pass null if we don't get it and let `ProfileHeader` determine whether it's ready or not.

src/view/com/profile/ProfileHeader.tsx
```diff
@@ -76,7 +83,7 @@ export function ProfileHeader({
   if (!profile || !moderation) {
     return (
       <View style={pal.view}>
-        <LoadingPlaceholder width="100%" height={153} />
+        <LoadingPlaceholder width="100%" height={150} />
```
haileyok, line 86: the height is supposed to be 150, not 153. fixes a slight shift we see.

src/view/com/profile/ProfileHeader.tsx
```diff
@@ -42,22 +42,29 @@ import {usePalette} from 'lib/hooks/usePalette'
...
 import {s, colors} from 'lib/styles'
 import {logger} from '#/logger'
 import {useSession, getAgent} from '#/state/session'
 import {Shadow} from '#/state/cache/types'
 import {useRequireAuth} from '#/state/session'
 import {LabelInfo} from '../util/moderation/LabelInfo'
+import {useProfileShadow} from 'state/cache/profile-shadow'
 
 interface Props {
-  profile: Shadow<AppBskyActorDefs.ProfileViewDetailed> | null
```
gaearon, line 60: We should keep this forced being a shadow. Otherwise it's easy to pass an unshadowed profile to the follow button — but it requires a shadow to work properly. (We should probably express that in types.)

gaearon: actually wait, I might be wrong here, you do turn it to a shadow below

gaearon (approved):

pfrazee (approved):
LGTM!

PR #2800, opened 2024-02-08
Improve discovery of the quick account switcher

mary-ext:
This pull request makes the following changes:
- Show a chevron beside the profile icon to show that you can indeed, quickly switch accounts.
- ~~Only allow quick account switcher open if there's multiple accounts~~

| Light | Dim |
| --- | --- |
|  |  |

The chevron only appears if there's multiple accounts signed in, if the switcher had the ability to add new accounts I'd probably have it always shown, thoughts? I'll put it in a separate pull request if that's the direction we want.

Additionally, should this be announced in accessibility label as well?

### DIFF
--- a/src/lib/icons.tsx
+++ b/src/lib/icons.tsx
@@ -974,3 +974,32 @@ export function ListIcon({
     </Svg>
   )
 }
+
+// Copyright (c) 2020 Refactoring UI Inc.
+// https://github.com/tailwindlabs/heroicons/blob/master/LICENSE
+export function ChevronUpDownIcon({
+  style,
+  size,
+  strokeWidth = 1.5,
+}: {
+  style?: StyleProp<ViewStyle>
+  size?: string | number
+  strokeWidth?: number
+}) {
+  return (
+    <Svg
+      fill="none"
+      viewBox="0 0 24 24"
+      strokeWidth={strokeWidth}
+      stroke="currentColor"
+      width={size}
+      height={size}
+      style={style}>
+      <Path
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        d="M8.25 15 12 18.75 15.75 15m-7.5-6L12 5.25 15.75 9"
+      />
+    </Svg>
+  )
+}
--- a/src/view/shell/bottom-bar/BottomBar.tsx
+++ b/src/view/shell/bottom-bar/BottomBar.tsx
@@ -234,7 +234,7 @@ export function BottomBar({navigation}: BottomTabBarProps) {
                       <ChevronUpDownIcon
                         size={16}
                         strokeWidth={1.9}
-                        style={[pal.text, styles.profileIconChevron]}
+                        style={[pal.textLight, styles.profileIconChevron]}
                       />
                     )}
                   </View>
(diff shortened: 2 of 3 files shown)


pfrazee:
Hey mary! Not against this idea but not sure that icon is the move. I'd probably just do a chevron up. Even then I'm not sure -- is that enough for people to understand that it's a long press? (Could be just not sure)

surfdude29:
Some form of visual indication of the quick account switcher would be great.

I don't know how feasible this is in the UI, but what about something like how GitHub shows co-authored contributions, with the pfp of another of your accounts behind the currently active one?

IMG_8764

My thinking is that this would be a good visual indication for the user as to how they can access the quick account switcher and encourage them to try long-pressing the icon.

I'm not sure how it would handle more than two accounts though, would it add another pfp behind the second one (up to three max?) or would it only show two at most?

mary-ext:
The alternative would be to have a onboarding/tour when you login into your second account

mary-ext:
> I don't know how feasible this is in the UI, but what about something like how GitHub shows co-authored contributions, with the pfp of another of your accounts behind the currently active one?

Might attempt doing this tomorrow, GitHub Mobile app does this as well when you have multiple accounts signed in.

mary-ext:
Here's the alternative design that was requested, this is basing off GitHub Mobile app's design where the secondary account has a slightly smaller avatar.

| Theme | Unfocused | Focused |
| --- | --- | --- |
| Dark |  |  |
| Light |  |  |

pdelfan:
I think the chevron would make more sense if there was an option to add another account on the account switcher modal.

mary-ext:
If there was the option to add another account I'd probably make the chevron always visible too.

mary-ext:
It does seem like the avatar stack makes for a busy-looking bottom bar though, a slightly smaller avatar might work, but I do prefer the chevron for the most part.

surfdude29:
The alternative design based off the GitHub Mobile app's design looks really good, very nicely done @mary-ext 👍

And I think it would make it more likely that a user who has added a second account would figure out how to access the quick account switcher on their own.

However, I think Mary is right that the chevron design looks better overall.

Perhaps going with the chevron design (both chevrons as Mary proposed or with just a chevron up as Paul suggested) and finding a good way to make users aware of the feature would be the best way forward?

And for doing that, what about showing a brief hint, something like: "You can quickly switch between accounts by long-pressing your profile picture" immediately after the user adds a second account?

Basically, what Mary suggested previously, and a first step in implementing #2802 (which imo is a very good idea):

> The alternative would be to have a onboarding/tour when you login into your second account, might be worth tackling alongside #2802

pdelfan:
> Yeah, I might just add that in, I'd probably make the chevron always visible too.

Yeah that would make sense. It would be good to look at other UI components in the app and see how they handle affordances, especially those designed for power-users which tend to be subtle or hidden away. In general, though, recognition is better than recall.

Invertex:
Little worried this idea got forgotten about! It's definitely a very helpful UI improvement, as currently the account switching feels like a little secret you either have to stumble on or have someone tell you about.

I do like the faded icon approach, but the chevron is probably a more clear indicator of switchability...

Turtlepaw:
You could also add something like Twitter does. You click open the side panel and click the more button and it opens the account sheet.

Manu people migrate from X to Bluesky and expect to have the same functionality and can't find it (like me) in that spot.

Issue #2793, opened 2024-02-08
UI overflow caused by german translation of "Notifications" in sidebar


lenke182:
**Describe the bug**
When using German as the app language on https://bsky.app, the German translation for "Notifications" in the sidebar causes an overflow that makes the sidebar scrollable.

**To Reproduce**

1. Use "German" as the app language and visit https://bsky.app

**Expected behavior**

The sidebar remains unscrollable and the German term doesn't cause an overflow in the UI.

**Screenshots**

[recording.webm](https://github.com/bluesky-social/social-app/assets/13052748/8edc40e2-6680-4565-a3b2-475538ff33d0)

**Details**

 - Platform: Desktop, Firefox 122.0 (https://bsky.app)
 - App version: 1.68.0

**Suggested action**

- Replace the term "Benachrichtigungen" with the term "Mitteilungen" in the German translation, which is shorter, has a similar meaning and doesn't cause an overflow. (This also what Twitter did in their sidebar.)
- OR change the code for the sidebar to trim the term with "..." to fix the overflow.

surfdude29:
I think this was fixed by #2805

PR #2805, opened 2024-02-08
Solved Issue #2793

Armaan48985:
Changed "Benachrichtigungen" to "Mitteilungen"


pfrazee (approved):
Cool I'm good with this (as good as I can be for a totally uninformed opinion)

pfrazee:
Well -- I'm informed by a lot of feedback and this one got a general +1, but personally no idea

Martin gave a good breakdown (as did others):

> iOS and X use “Mitteilungen” while Facebook and Instagram use “Benachrichtigungen”, so I guess either of those should be ok. “Mitteilungen” risks being confused with DMs (although I’d translate DMs as “Nachrichten”). I wouldn’t use “Meldungen” or the English word, as some suggested.

PR #2806, opened 2024-02-08
Add composer button to Notifications

gaearon:
I often have a thought while reading notifications — let's add the button there.

(I considered adding to Search too but it wouldn't work super well in our layout due to overlapping Follow and Compose buttons. Maybe we can do that if we redesign a bit.)

Notifications is logged in only so I didn't add a session check.

Screenshot 2024-02-08 at 17 30 16
Screenshot 2024-02-08 at 17 30 19
Screenshot 2024-02-08 at 17 30 49
Screenshot 2024-02-08 at 17 34 14


pfrazee (approved):
Great!

PR #2808, opened 2024-02-08
Add pressed state to profile counts

gaearon:
For iOS:

https://github.com/bluesky-social/social-app/assets/810438/b5b52d54-4f65-4e29-b852-5a2f67ac5a33

No changes on web (where this isn't needed). 

Doesn't do anything for Android, but we generally don't handle it there for other text links either.


gaearon:
Ah, this lost the hover underline on the web. I'll need to re-add.

PR #2822, opened 2024-02-09
Design system tweaks

estrattonbailey:
After using this the last couple weeks I had some improvement todos piling up.

Tweaked color palette a bit:
- darker grays pair better with a little more saturation
- dim/dark have consistently spaced values now

Tweaked theme atom names:
- text contrast options are more specific, increase consistency
- border contrast options match text contrast naming convention, much better options that previous
- expanded bg color options to include all bc seems like I had to manually do this a lot so far

Tweaked `Typography` to reduce confusion:
- just use `Text`
- use semantic HTML where appropriate (TBD where)

Then I swapped our few usages of these things with new values, mostly direct swaps.


pfrazee (approved):
🧑‍🎨

src/components/Typography.tsx
```diff
@@ -41,138 +41,42 @@ function normalizeTextStyles(styles: TextStyle[]) {
...
+        {...rest}
+        style={normalizeTextStyles([t.atoms.text, flatten(style)])}
+      />
+    )
+  }
 }
 
+/*
+ * Use semantic components when it's beneficial to the user or to a web scraper
+ */
```
gaearon, line 73: In my experience this is nice to automate using context. See this example: https://react.dev/learn/passing-data-deeply-with-context#context-passes-through-intermediate-components

PR #2828, opened 2024-02-10
Add follow button to highlighted post

haileyok:
Doing some modifications on the highlighted post.

1. Adding a follow button
  - Button is initially visible if the user is not being followed.
  - The button stays visible even after the user is followed until the user leaves the screen
  - The button will be removed/added both on blur and on focus, so it will maintain visibility with the `viewer` in the profile's shadow
2. Remove the `TimeElappsed` from the highlighted post. We are displaying the timestamp below, and having this there is hindering us having the button there. This is the common behavior in other apps as well.
3. Removing static width on the display name and handle. We flex these properly now so that they take up the full line instead of being truncated. This prevents  some issues that we were seeing with longer display names (especially whenever using larger font sizes)

https://github.com/bluesky-social/social-app/assets/153161762/55be7ef1-4ae7-4abf-ac3c-988b8a8a6ed7

### DIFF
--- a/src/view/com/post-thread/PostThreadFollowBtn.tsx
+++ b/src/view/com/post-thread/PostThreadFollowBtn.tsx
@@ -110,8 +110,6 @@ function PostThreadFollowBtnLoaded({
     }
   }, [isFollowing, requireAuth, queueFollow, _, queueUnfollow])
 
-  if (!showFollowBtn) return null
-
   return (
     <View style={styles.container}>
       {showFollowBtn && (
@@ -149,7 +147,7 @@ function PostThreadFollowBtnLoaded({
 
 const styles = StyleSheet.create({
   container: {
-    width: 95,
+    width: 100,
   },
   btnOuter: {
     marginLeft: 'auto',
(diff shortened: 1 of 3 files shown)


surfdude29:
Sorry for jumping in here unsolicited, but what about simply using a plus sign as the icon on mobile? It would match the **+ Follow** button, just without the text (given the lack of space). I think the function of the button would be more intuitive to users and would prompt fewer "Oh **** I didn't realise tapping that meant Follow" situations.

Here is a crude iOS mockup (based on 1.67 so the layout of other elements is out-of-date):

IMG_8344

haileyok:
Hey it's open for a reason, always feel free to jump in!

I had the same thought about the icon, but wasn't sure that the plus by itself was clear enough. I also feel the same way about the current icon though, so... :) 

I'm not opposed at all though to having a modal that verifies your decision to follow, I'll see how nicely that implements, but I suspect it would be pretty nice actually. Would only use it on mobile I think?

gaearon:
Re: three seconds, I'd just keep it there (so it's shows if you weren't following at the time you pressed into the post).

haileyok:
> Re: three seconds, I'd just keep it there (so it stays if you weren't following at the time you pressed into the post).

Yea agree. I made some more changes here as noted above. I think this is actually a better look. We update the `wasFollowing` state on focus as well so that we can stay in sync with the profile shadow.

haileyok:
@emilyliu7321 Just tagging you for your input here on whats above

surfdude29:
@haileyok The solution you've arrived at is great, nice job! :)

pfrazee:
If we're looking to reduce down to an icon, maybe this:

CleanShot 2024-02-12 at 11 30 40@2x

PR #2829, opened 2024-02-10
Increase stroke width for active hashtag icon in left nav

mozzius:
HashtagIcon's active state behaviour seems to generally be to increase the stroke width from 2.25 to 4 in most places, but in the left nav it was left as 2.25, so there was effectively no active state. This fixes that.


gaearon (approved):
sg

PR #2830, opened 2024-02-10
Update `blueskyweb.xyz` links to `bsky.social`

surfdude29:
This PR updates links to `blueskyweb.xyz` found in various places to the equivalent URLs on `bsky.social`.

It makes two other minor changes:

- The superfluous localisation parameters at the end of the Play Store link in README.md are removed.
- The link in `CopyrightPolicy.tsx` which currently points to the Community Guidelines has been changed to point directly to the Copyright Policy (which I'm assuming is the intended location?)

This PR does not, of course, make any changes to any references to `blueskyweb.xyz` relating to how the app is identified for the purposes of the app stores etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Get the app itself:
 
 - **Web: [bsky.app](https://bsky.app)**
 - **iOS: [App Store](https://apps.apple.com/us/app/bluesky-social/id6444370199)**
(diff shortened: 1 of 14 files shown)


haileyok (approved):
All of these are correct, don't see any other instances we need to change. Thank you!

PR #2831, opened 2024-02-11
change `follow all` to `follow` if any recommendation is deselected in onboarding

haileyok:
Small one. Just changing `Follow All` to `Follow` if we deselect anyone in the follow recommendations during onboarding.


surfdude29:
This is a welcome change (and it has saved me opening an issue about it, thanks!) because I saw this review in the App Store where a user had this problem during onboarding:

IMG_8410

pfrazee (approved):

pfrazee:
The unticks aren't lost -- it presents new suggestions if you go back

pfrazee:
But the ticks/unticks you did previously are retained. Not super clear though

PR #2841, opened 2024-02-12
Use EAS managed build number, run build/submit on GH Actions

haileyok:
This creates a better workflow for building and deploying. Also does automatic nightly TestFlights for iOS

- Use GitHub runners for deployments instead of EAS. iOS runs on `macos-14` and Android runs on `ubuntu-latest`.
- Use EAS's remove build number instead of saving locally to `app.config.js`. This eliminates the need for committing to the repo each time we deploy to TestFlight.
- The iOS job runs either manually or automatically each night at 9PM PST. The Android job only runs manually.
- Both jobs automatically deploy to the store after building. (Build times for both platforms average 15 minutes)
- On both platforms we obtain the Sentry auth tokens from a GitHub secret. On Android `google-services.json` is also obtained from another GitHub secret.


app.config.js
```diff
@@ -169,7 +149,7 @@ module.exports = function () {
               organization: 'blueskyweb',
               project: 'react-native',
               release: VERSION,
-              dist: `${PLATFORM}.${VERSION}.${DIST_BUILD_NUMBER}`,
+              dist: `${PLATFORM}.${VERSION}.${new Date().toISOString()}`,
```
estrattonbailey, line 152: Sorry none of this has comments on it, but we need to keep these values in sync with our Sentry config in order for source maps to work. Let's chat in Slack 👍

estrattonbailey (changes requested):
Small thing, should be easy to resolve

app.config.js
haileyok: I think this will cover it. Going to download the IPA and install to make sure it sets the version correctly.

haileyok: Bundle looks good, just check Sentry and we should be good to go.

estrattonbailey (approved):
LGTM let's give it a go 🎉

PR #2844, opened 2024-02-12
remove line height from text input for composer on ios

haileyok:
Again addressing RN weirdness with line height in the text composer. This only applies on iOS, there are no visible issues on android using `lineHeight` styles.

The previous patch for this fixed the line height from causing the text to continue off the screen, however, it seems to have introduced a small movement on the first line of text. I think until RN merges a full fix for this the best option is to just not apply a line height on iOS.

Before and after:

https://github.com/bluesky-social/social-app/assets/153161762/48b67175-3838-43ae-aac9-fb3b75da1cc8


pfrazee (approved):
LGTM

PR #2846, opened 2024-02-12
Improve a11y of inputs

estrattonbailey:


src/components/forms/TextField.tsx
```diff
@@ -72,35 +69,17 @@ export function Root({children, isInvalid = false}: RootProps) {
...
-        onPress={() => inputRef.current?.focus()}
-        onHoverIn={onHoverIn}
-        onHoverOut={onHoverOut}>
+      <View
+        style={[a.flex_row, a.align_center, a.relative, a.w_full, a.px_md]}
+        {...web({
+          onClick: () => inputRef.current?.focus(),
+          onMouseOver: onHoverIn,
+          onMouseOut: onHoverOut,
+        })}>
```
estrattonbailey, line 80: This was resulting in a separately tabbable element wrapping the input (duh) which caused inputs to require two taps to enter when using a screen reader.

This was here so that clicks/taps to the icon/suffix provided caused the input to focus. This retains this functionality on web. On mobile, you must tap the center of the input.

But with a screen reader, this is now fixed and allows normal navigation.

src/components/forms/TextField.tsx
```diff
@@ -187,7 +166,6 @@ export function createInput(Component: typeof TextInput) {
         <Component
           accessibilityHint={undefined}
           {...rest}
-          aria-label={label}
           accessibilityLabel={label}
```
estrattonbailey, line 169: Evidently react-native-web applies `aria-label` based on `accessibilityLabel`.

pfrazee (approved):
LGTM

PR #2848, opened 2024-02-12
Add search button to header on feeds screen

haileyok:
Adds a button to the header of the feeds list that focus the search input. I think we should focus on reworking this screen a good bit later, think we'll be discussing that coming up.

https://github.com/bluesky-social/social-app/assets/153161762/51243315-e727-4959-a9c5-1ddb26cb8b15


pfrazee (approved):
Nice, lgtm

PR #2850, opened 2024-02-12
Unmark DEV strings for translation in Settings

surfdude29:
As I understand it, there are strings in Settings which are only visible in DEV builds, which I think only Bluesky devs and maybe some others who are building the app themselves use?

Given this, my understanding is that these strings are not visible to any end-users, and so I don't think it's worth marking them for translation, as this adds an (admittedly small) extra burden to translators in return for virtually no user benefit, as their translations will only be seen by a very small number of people (if anyone).

Therefore, this PR proposes unmarking these strings for translation, which would decrease the number of unique strings for translators by 13. It also makes two minor tweaks to `accessibilityHint`s, changing `Clear` to `Clears`.

I apologise if I have misunderstood the situation, and in that case please don't hesitate to close this PR 🙏


kisaragi-hiu:
My opinion as a translator is that it's just a few strings, and we might as well make the development build also pleasant to use for developers who can use English but feel more comfortable with their native language.

surfdude29:
> My opinion as a translator is that it's just a few strings, and we might as well make the development build also pleasant to use for developers who can use English but feel more comfortable with their native language.

That's a good point, and I appreciate you sharing your perspective as a translator. I'll leave this open for a bit longer in case anyone else (especially translators and/or those who use dev builds) has feedback, but based on what you said I'm inclined to close it.

Signez:
> I'll leave this open for a bit longer

While I searched for another PR, I stumbled upon this one, and as a co-translator I'd suggest to close this one. It's not too big of a deal to translate those few strings anyway!

surfdude29:
> > I'll leave this open for a bit longer
> 
> While I searched for another PR, I stumbled upon this one, and as a co-translator I'd suggest to close this one. It's not too big of a deal to translate those few strings anyway!

True, 13 extra strings is not that big of a burden to translate.

PR #2851, opened 2024-02-12
Remove invites code UI

estrattonbailey:


pfrazee (approved):
Looks good

-- note for others, this UI will come back based on the server config in the future, that's why it's just commented out in settings

PR #2852, opened 2024-02-13
Add copy to feeds page

haileyok:


src/screens/Onboarding/StepAlgoFeeds/index.tsx
```diff
@@ -20,7 +20,7 @@ import {
   OnboardingControls,
 } from '#/screens/Onboarding/Layout'
 import {FeedCard} from '#/screens/Onboarding/StepAlgoFeeds/FeedCard'
-import {IconCircle} from '#/screens/Onboarding/IconCircle'
+import {IconCircle} from '#/components/IconCircle'
```
haileyok, line 23: I think we might find more use for this icon circle throughout the app as we implement ALF so moving it to the root components directory.

src/components/IconCircle.tsx
```diff
@@ -30,8 +30,8 @@ export function IconCircle({
         a.align_center,
         a.rounded_full,
         {
-          width: 64,
-          height: 64,
+          width: size === 'lg' ? 52 : 64,
+          height: size === 'lg' ? 52 : 64,
```
haileyok, line 34: This will only affect this component, we should come back and create different size categories for this @estrattonbailey

pfrazee (approved):

PR #2854, opened 2024-02-13
Fix duplicate keys in PostThread

gaearon:
In theory these could be duplicate. I couldn't actually repro this case because deletion cuts off the parents but let's fix just in case (and for when we'll fix it to not cut off the parents).


pfrazee (approved):
LGTM, seems simpler anyway

PR #2862, opened 2024-02-13
Move visual display logic out of PostThread generators

gaearon:
## What?

This is based on the work @haileyok started in https://github.com/bluesky-social/social-app/pull/2755.

Currently, the generator functions produce the FlatList items for the UI, including elements like:

- Top page header
- Individual parent items
- The highlighted post
- The reply prompt
- "Replies are loading" spinner
- Individual reply items
- The footer

They do this while iterating over the post structure (upwards for parents, downwards for children).

There are some issues with this approach.

One issue is that the generators were written to be more expressive than what the backend sends *and* what the component can actually handle. E.g. we could never end up with two child spinners, but the data structure makes it seem like we can. On the other hand, if we *did* end up with two child spinners, we'd actually get a key error because their keys are currently written to be unique. To sum up, that code was far more general than both our needs and what the rest of the code can handle.

The other, more pressing, issue is that we'd like to make more rendering behavior adjustments. E.g. in https://github.com/bluesky-social/social-app/pull/2755, we want to reveal posts in "two steps" (first, the root post, then the rest) even if both parents and replies are synchronously available. We also want to provide our own paging on top of the paging done by the virtualized list itself. This logic is difficult to write when it's weaved into the model traversal logic — these are two separate domains.

In this PR, I'm splitting this logic to be a bit less fancy. The generators only take care of enumerating the parents and the replies. The rendering code pieces them together and inserts the appropriate UI elements where needed. This will make it easier to adjust the visual behavior independently of the model traversal. It's also less mindbendey (each traversal only goes in one direction). It will also let us retain memoization of the generator results independently from the state that https://github.com/bluesky-social/social-app/pull/2755 adds.

## Test Plan

Navigate around threads. Should have exactly the same behavior as before.

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -160,13 +160,18 @@ function PostThreadLoaded({
 
   // construct content
   const posts = React.useMemo(() => {
-    let arr = Array.from(
-      flattenThreadSkeleton(
-        sortThread(thread, threadViewPrefs),
-        hasSession,
-        treeView,
-      ),
-    )
+    const sortedThread = sortThread(thread, threadViewPrefs)
+    let arr = []
+    for (let parent of flattenThreadParents(sortedThread, hasSession)) {
+      arr.push(parent)
+    }
+    for (let reply of flattenThreadReplies(
+      sortedThread,
+      hasSession,
+      treeView,
+    )) {
+      arr.push(reply)
+    }
     if (arr.length > maxVisible) {
       arr = arr.slice(0, maxVisible).concat([LOAD_MORE])
     }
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Yeah seems fine to me

src/view/com/post-thread/PostThread.tsx
```diff
@@ -160,19 +168,31 @@ function PostThreadLoaded({
...
   const posts = React.useMemo(() => {
-    let arr = Array.from(
-      flattenThreadSkeleton(
-        sortThread(thread, threadViewPrefs),
-        hasSession,
-        treeView,
-      ),
-    )
+    const root = sortThread(thread, threadViewPrefs)
+    let arr: RowItem[] = []
```
gaearon, line 172: I guess this by itself could've been a generator but meh.

PR #2865, opened 2024-02-14
Adjust post thread scroll for cached posts

gaearon:
I'm splitting https://github.com/bluesky-social/social-app/pull/2755 into independent parts so that it's easier to understand how the fix is layered, and to ensure that it's correct (especially if we revisit this in the future). The overall problem and the solutions are described in https://github.com/bluesky-social/social-app/pull/2755 in detail.

In this PR, we're including these parts of the fix:

- The native side will no longer try to adjust scroll in `onContentSizeChange`. That's not a good place to adjust scroll because on the native it is asynchronous—so if we failed to maintain the scroll position by this point, the user has already seen a jump. We'd like to always use `maintainVisibleContentPosition` on native, including for cached posts.
- In order to make `maintainVisibleContentPosition` consistently work well, we'll need a few adjustments.
  - For `maintainVisibleContentPosition` to work at all, we need to have some stable index to "latch onto". In other words, we need to *first* get the root post on the screen, and *then* prepend the parents before it. This is why this PR introduces `deferParents` which artificially defer parent rendering (even if the data for the parents is cached). Now `maintainVisibleContentPosition` works for cached posts so we can make `onContentSizeChange` web-only.
  - **This PR alone won't be sufficient to fix jumps caused by other reasons.** Those will be handled in separate PRs:
    - There's a bug in `maintainVisibleContentPosition` that breaks it when there are too many prepended rows. @haileyok found a workaround for it in https://github.com/bluesky-social/social-app/pull/2755 — we can limit how many we render. I'll send that fix separately since it's conceptually unrelated to this PR.
    - In some cases we don't have enough data (like metrics or embeds), causing jumps when they load. This is unrelated to this fix, so I will also separate that out in another PR.

## Test Plan

Open a thread with >1 posts. Tap into post 1. Tap into post 2. Tap into post 1 (uses cache). Tap into post 2 (uses cache).

Before this PR: the cached case either resets scroll or is imprecise.

After this PR: the cached case preserves scroll and is precise.

Does not affect web.

## Before

https://github.com/bluesky-social/social-app/assets/810438/ab17849b-feda-4612-84c3-eb451fc23c73

https://github.com/bluesky-social/social-app/assets/810438/249ee71d-ae2e-405d-8bde-3f19d42e3afc

## After

https://github.com/bluesky-social/social-app/assets/810438/43dafc17-b745-488c-9805-b84f65e65f70

https://github.com/bluesky-social/social-app/assets/810438/adb0dbea-b37f-4d8f-aedd-a068c958ef14

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -47,7 +47,11 @@ import {isAndroid, isNative} from '#/platform/detection'
 import {logger} from '#/logger'
 import {moderatePost_wrapped as moderatePost} from '#/lib/moderatePost_wrapped'
 
-const MAINTAIN_VISIBLE_CONTENT_POSITION = {minIndexForVisible: 1}
+const MAINTAIN_VISIBLE_CONTENT_POSITION = {
+  // We don't insert any elements before the root row while loading.
+  // So the row we want to use as the scroll anchor is the first row.
+  minIndexForVisible: 0,
+}
 
 const TOP_COMPONENT = {_reactKey: '__top_component__'}
 const REPLY_PROMPT = {_reactKey: '__reply__'}
@@ -64,6 +68,12 @@ type RowItem =
   | typeof CHILD_SPINNER
   | typeof LOAD_MORE
   | typeof BOTTOM_COMPONENT
+
+type ThreadSkeletonParts = {
+  parents: YieldedItem[]
+  highlightedPost: ThreadNode
+  replies: YieldedItem[]
+}
 
 export function PostThread({
   uri,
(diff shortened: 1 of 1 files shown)


src/view/com/post-thread/PostThread.tsx
```diff
@@ -337,7 +347,8 @@ function PostThreadLoaded({
           : undefined
         return (
           <View
-            ref={item.ctx.isHighlightedPost ? highlightedPostRef : undefined}>
+            ref={item.ctx.isHighlightedPost ? highlightedPostRef : undefined}
+            onLayout={deferParents ? () => setDeferParents(false) : undefined}>
```
haileyok, line 351: nice

haileyok (approved):

Issue #2861, opened 2024-02-13
Hover preview for accounts


BubiBalboa:
**Is your feature request related to a problem? Please describe.**

On the web UI, when I'm reading a conversation I'd like to have a quick way to see whether I'm following an account or not and to have a preview of their bio.

**Describe the solution you'd like**

Hover preview cards i.e. hovering with the mouse cursor over a username brings up a card with the user's bio and the (un)follow button. 

**Describe alternatives you've considered**

Some other follow indicator that's always visible would also work, I guess. It would clutter the UI but might be a good option for the app and mobile view where hover preview isn't possible. 

**Additional context**

I think having as few barriers as possible to (un)following others is good for the users and the social network. Keeps it from getting stale. The bio preview helps to quickly gauge a users background while reading comments.

surfdude29:
This has just been implemented in #3508, thanks @mary-ext and @estrattonbailey! 🙏

BubiBalboa:
Fantastic! Thank you!

PR #2867, opened 2024-02-14
Profile card hover preview

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/2861

Adds profile card preview hovering, this is a very rough implementation meant to get the ball rolling, please take it over :)

There's no delay between hovering-in and hovering-out at the moment, it seems super finicky because if I add a debounce to `hovering`, the preview gets stuck mounted if you head to another page (by clicking the link), it only goes away once you move the cursor

It could be compounded by the use of our custom portal component, and/or Reanimated, but I'm not sure how much time I should be spending on it

I've decided to use `@floating-ui/react-dom` for this because it's a little easier to do floats, if this pull request makes it in, I'll make a subsequent pull request to refactor the web composer to make use of it as well and remove `tippy.js` from the dependency (we don't need multiple popper implementations)

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "@expo/webpack-config": "^19.0.0",
     "@floating-ui/dom": "^1.6.3",
     "@floating-ui/react-dom": "^2.0.8",
-    "@floating-ui/react-native": "^0.10.4",
     "@fortawesome/fontawesome-svg-core": "^6.1.1",
     "@fortawesome/free-regular-svg-icons": "^6.1.1",
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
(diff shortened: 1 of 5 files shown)


mary-ext:
here's what I tried debouncing the hovering value with

```ts
import React from 'react'

export function useDebouncedValue<T>(value: T, delay: number): T {
  const [state, setState] = React.useState(value)
  const ref = React.useRef(value)

  React.useEffect(() => {
    if (ref.current !== value) {
      const timeout = setTimeout(() => setState(value), delay)
      return () => clearTimeout(timeout)
    }
  }, [value, delay, ref, setState])

  return state
}
```

gaearon:
Refs don’t update values on their own (the value you pass to useRef is only used during first render). Try writing to it from the effect?

mary-ext:
Ah right, I *think* I've set both the ref value and the state in the timeout, and unknowingly removed it while pasting it here

I'll correct that, it doesn't solve the issue with the preview not getting unmounted during page navigation though

gaearon:
No prob, I'll have a look. Thanks!

PR #2878, opened 2024-02-15
Fix jumps when navigating into long threads

gaearon:
The other half of the fix from https://github.com/bluesky-social/social-app/pull/2755. (The first one was https://github.com/bluesky-social/social-app/pull/2865.)

## What?

Navigating to a post in the thread needs to keep this post visible while the parents load. On the native side, we do this using `maintainVisibleContentPosition`. However, that prop is buggy if you prepend too many items at once — it causes wild scroll jumps. This can likely be fixed on the RN side ([repro case](https://snack.expo.dev/@haileyok/flatlist-repro), cc @janicduplessis) but for now we'll try to work around it.

The workaround is to prepend parent rows in chunks, and only do this while you reach the top, and only do this after the scroll is resting. This is somewhat annoying but hopefully not by too much. The jumps are still possible if you scroll more aggressively — I don't think there's a way to completely get rid of them but I'm open to ideas.

I've removed pull-to-refresh from the post thread. This is an usual interaction for a thread, and I don't think people do it super commonly. It was conflicting with the scrolling at the edge, and in some cases I was seeing the pull-to-refresh indicator getting stuck with no way to get rid of it. So it seems like it's better just not to have it here.

There's also a drive-by change that adds a little glimmer to the engagement bar ("X likes, Y reposts") for highlighted posts. If we don't know the metrics, we're probably coming from a QT — and QT'd posts tend to have other engagement. So it would be better to assume the engagement exists (and to hide the bar if we're wrong, causing a layout shift) than to assume there's no engagement, and then to cause a layout shift if there is. In longer term, we should solve this by including the metrics for QT'd posts into the parent timeline response. Then we wouldn't need to be guessing at all.

This PR does not affect the web, except for the engagement glimmer (which is added consistently on all platforms).

## Before

On iOS, navigating into a long thread is very jumpy and doesn't point you to the correct post.

https://github.com/bluesky-social/social-app/assets/810438/ef972fb4-3a87-43c7-934d-74e2cc1d6724

On Android, it was better, although during scroll up it would also jump around sometimes.

https://github.com/bluesky-social/social-app/assets/810438/6c3f946c-3fb9-47d5-8aba-1349361a75f3

## After

We now correctly get to the right place in the thread.

The tradeoff is that parents load "on demand", i.e. if you scroll up, you'll have to bounce into the edge (and rest) until it loads more. We haven't been able to solve the jumps without that limitation. There's hope that we can remove this once `maintainVisibleContentPosition` is less buggy.

https://github.com/bluesky-social/social-app/assets/810438/683a8b1c-4c03-4545-8f51-52168e83d46d

https://github.com/bluesky-social/social-app/assets/810438/cbb3727e-fe5e-402e-bd8b-c2c87fb4c826


haileyok (approved):
Thanks for much for working on this, LGTM.

pfrazee (approved):
Strong +1 to this, and my only question is whether we should put an activityindicator in the topmost position so that it's clear when you reach the top that there is more?

PR #2882, opened 2024-02-16
Implement captcha

haileyok:
Removing phone verification from the flow and adding captchas.

- Move handle selection from step 3 to step 2.
- Make step 3 the captcha.
- Move `submit` to a hook so we can easily access it from the two required places
- Remove all the phone number code, including the phone number validation lib.
- Move errors to the top of forms for when we get taken back in the process

https://github.com/bluesky-social/social-app/assets/153161762/9d0ae9bf-0ce0-4d5a-95bd-2256ee1952b3


haileyok:
Just waiting for us to determine the prod URL for this and we'll be set 👍

src/view/com/auth/create/state.ts
```diff
@@ -328,23 +292,16 @@ function compute(state: CreateAccountState): CreateAccountState {
...
+    canNext = false
   }
   return {
     ...state,
     canBack: state.step > 1,
     canNext,
     isInviteCodeRequired: !!state.serviceDescription?.inviteCodeRequired,
-    isPhoneVerificationRequired:
-      !!state.serviceDescription?.phoneVerificationRequired,
+    isCaptchaRequired: !!state.serviceDescription?.phoneVerificationRequired,
```
haileyok, line 305: We should check that this is still going to be `phoneVerificationRequired` for now or if we are changing it.

haileyok: Okay we are leaving this here for now while we transition over. Will come back in the future.

src/view/com/auth/create/state.ts
```diff
@@ -77,144 +73,127 @@ export function useCreateAccount() {
...
-    uiDispatch({type: 'set-error', value: cleanError(errMsg)})
-    throw e
-  }
+        const error = cleanError(errMsg)
+        const isHandleError = error.toLowerCase().includes('handle')
+
+        uiDispatch({type: 'set-processing', value: false})
+        uiDispatch({type: 'set-error', value: cleanError(errMsg)})
+        uiDispatch({type: 'set-step', value: isHandleError ? 2 : 1})
+        throw e
```
haileyok, line 177: @estrattonbailey Is there a reason that we throw here for create account error? Is that for Sentry?

estrattonbailey: Looks like previously it exited early at its callsite (we should retain that I suppose), but otherwise I don't think it's captured or handled anywhere.

haileyok: Yea it isn't handled anywhere, so just throws a red screen on dev. Wasn't sure if this was some trick to get sentry to pick it up or not. Remove is fine?

haileyok:
@pfrazee Ready to glance this over, let's not merge yet though. Logic should be sound.

Can't test locally on web because of CORS, but simulators should (right now) display `ROUTE NOT FOUND` where the captcha will be (waiting for this to go live).

src/view/com/auth/create/CreateAccount.tsx
```diff
@@ -84,33 +75,47 @@ export function CreateAccount({onPressBack}: {onPressBack: () => void}) {
...
+    uiState.isCaptchaRequired,
+    uiState.handle,
     uiDispatch,
-    onboardingDispatch,
-    createAccount,
-    setBirthDate,
-    setSavedFeeds,
     _,
+    submit,
   ])
```
haileyok, line 119: This should validate handles for us before continuing.

https://github.com/bluesky-social/social-app/assets/153161762/1ed6ad74-589f-49c8-9a1b-125bb37714fb

@pfrazee

src/view/com/auth/create/CaptchaWebView.tsx
```diff
@@ -0,0 +1,86 @@
...
+      if (urlp.host !== redirectHost) return
+
+      const code = urlp.searchParams.get('code')
+      if (urlp.searchParams.get('state') !== stateParam || !code) {
+        onError()
+        return
+      }
+
+      wasSuccessful.current = true
+      onSuccess(code)
```
haileyok, line 64: `onNavigationStateChange` can get called multiple times, so we don't want to submit multiple times as well.

src/view/com/auth/create/state.ts
estrattonbailey: Remove seems fine to me

devinivy (approved):
💎

PR #2887, opened 2024-02-16
Italian localization

Titianbeetle:
Italian localization revised with latest translations for Bluesky-Social V. 1.68


pfrazee (approved):
These look good! Thank you!

PR #2888, opened 2024-02-16
Bump contrast on dim mode for old ds

estrattonbailey:
Just overrides more colors using dimPalette, which is 1 stop lighter on the scale.


haileyok (approved):
my eyes thank you

pfrazee (approved):
Yeah lgtm

PR #2889, opened 2024-02-16
Some button updates

estrattonbailey:


src/components/Button.tsx
```diff
@@ -48,25 +48,32 @@ export type VariantProps = {
...
-    focused: boolean
-    pressed: boolean
-    disabled: boolean
+    testID?: string
+    label: string
+    style?: StyleProp<ViewStyle>
+    children:
+      | React.ReactNode
+      | string
+      | ((context: ButtonContext) => React.ReactNode | string)
```
estrattonbailey, line 72: Can now use a render function, which is passed the memoized context values.

I did this so that we can more easily define pressables and links (`Link` inherits `Button`) and apply interactive styles. Previously you had to use context, which mean defining the "inner" button components separately. Just not tenable long term.

src/components/Button.tsx
```diff
@@ -379,21 +395,31 @@ export function Button({
...
-          style={[a.absolute, a.inset_0]}
-        />
+        <View
+          style={[
+            a.absolute,
+            a.inset_0,
+            a.overflow_hidden,
+            {borderRadius: flattenedBaseStyles.borderRadius},
+          ]}>
+          <LinearGradient
```
estrattonbailey, line 405: This was added so that the button itself doesn't require `overflow: hidden`, since that was preventing using a shadow on a `Link` that wrapped a modservice card. The overflow is only here for gradients, so only added within a gradient button.

src/components/Button.tsx
```diff
@@ -514,9 +542,11 @@ export function ButtonText({children, style, ...rest}: ButtonTextProps) {
 export function ButtonIcon({
   icon: Comp,
   position,
+  size: iconSize,
 }: {
   icon: React.ComponentType<SVGIconProps>
   position?: 'left' | 'right'
+  size?: SVGIconProps['size']
```
estrattonbailey, line 549: Allows for override if needed

pfrazee (approved):
LGTM

PR #2891, opened 2024-02-16
Fix default loader color

estrattonbailey:


pfrazee (approved):
LGTM, didnt even notice we had this

PR #2896, opened 2024-02-16
Commit to theme mode of system

estrattonbailey:
Issue: OS-level theme mode and app theme mode can be different (light or dark), which was causing the splash screen to begin in dark mode, then abruptly switch to light mode (or vice versa).

This PR makes it so that the splash screen matches whatever mode your system is set to.


haileyok (approved):
Yup that looks right!

PR #2898, opened 2024-02-16
Drop the Following feed settings onboarding step

gaearon:
I think this is enough to remove it.


surfdude29:
This can probably be closed now as it was removed in #4224.

PR #2899, opened 2024-02-16
Add `selectable` to new text components

estrattonbailey:


src/components/Link.tsx
```diff
@@ -253,43 +258,41 @@ export function InlineLink({
   const flattenedStyle = flatten(style)
 
   return (
-    <TouchableWithoutFeedback
-      accessibilityRole="button"
+    <Text
```
estrattonbailey, line 261: Tbh idk why I had this wrapped in a touchable, but UITextView didn't like it since it's not Text. No functional changes here, so nice win win.

haileyok (approved):
Yup, LGTM! If you notice any of these have weird styles lmk, there might be some missing styles in the UITextView component, but it should be fine.

PR #2914, opened 2024-02-18
Connect replies from other users

mary-ext:
I dug up the original commit that introduced the threading, https://github.com/bluesky-social/social-app/commit/ba837ad9afd3f4faa5a47b7ea620cd633e99d9cf and there doesn't seem to be much of a reason as to why it was being done this way.

The problem is that it makes replies like this look awkward and confusing, they're part of the same reply chain, but they're being detached.

If we connect it, then it's clear that it's part of the same chain.

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -84,8 +84,8 @@ export class FeedViewPostsSlice {
   }
 
   insert(item: FeedViewPost) {
-    const selfReplyUri = getSelfReplyUri(item)
-    const i = this.items.findIndex(item2 => item2.post.uri === selfReplyUri)
+    const replyUri = getReplyUri(item)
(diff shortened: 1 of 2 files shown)


mary-ext:
I've seen new users and existing users getting confused with this weird chaining behavior as well, so while it's not perfect by any means, it seems worth getting this in

Especially considering the timeline logic tries to unwrap the slices as well

gaearon:
This has been implemented in https://github.com/bluesky-social/social-app/pull/4871 in a different way. I initially was planning to stack that work on top of this PR, but the implementation ended up completely different so it didn't really make sense. Thinking back on it, it might've been nice to include this commit there just for the initial exploration, so I'm sorry I forgot to do that before merge. Regardless, huge thanks for looking into it and raising this as a problem.

PR #2918, opened 2024-02-18
Add theme-color meta tag and apply selection of theme

edo-bari-ikutsu:
By applying this commit, theme color users selected on settings is applied to the theme-color tag. When users install the web page, color of status bar is synchronized to the theme color.

theme-color

### DIFF
--- a/src/alf/util/useColorModeTheme.ts
+++ b/src/alf/util/useColorModeTheme.ts
@@ -49,7 +49,7 @@ function updateDocument(theme: ThemeName) {
     html.className = html.className.replace(/(theme)--\w+/g, '')
     html.classList.add(`theme--${theme}`)
     // set color to 'theme-color' meta tag
-    meta.setAttribute('content', getBackgroundColor(theme))
+    meta?.setAttribute('content', getBackgroundColor(theme))
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/alf/util/useColorModeTheme.ts
```diff
@@ -46,6 +46,13 @@ function updateDocument(theme: ThemeName) {
     html.className = html.className.replace(/(theme)--\w+/g, '')
 
     html.classList.add(`theme--${theme}`)
+
+    // set color to 'theme-color' meta tag
+    const themeColor = window
+      .getComputedStyle(html)
+      .getPropertyValue('--background')
```
haileyok, line 53: Would it make more sense if we got the background color from the theme directly here, instead of getting it from the document's styles?

haileyok:
Thank you! I wasn't aware that this was a thing! Where I left the comment, I'm wondering if it might be easier to just use the background color directly from the theme instead of getting it from the document's styles? We could use the same logic from `useSystemBackground()` I think.

- Add `getBackgroundColor(theme: ThemeName)` to get the background color
- Replace the `updateSystemBackground(theme)` on line 16 with `SystemUI.setBackgroundColorAsync(getBackgroundColor(themeName))`
- Use `getBackgroundColor(themeName)` to get the background color inside of `updateDocument()` rather than getting it from the computed style.

Sorry, just looking at how we could simplify the rest of the code here as well, not just your changes! I could be off too. Really appreciate this!

edo-bari-ikutsu:
@haileyok Thank you for your suggestions! I've updated the implementation (e7e3fd0). Please review it.

haileyok (approved):
Nice, works great. Thank you! (And thanks for the other PR, will look at that later)

Issue #1194, opened 2023-08-16
Offer a PWA


scottdotjs:
**Is your feature request related to a problem? Please describe.**

N/a

**Describe the solution you'd like**

I would like there to be a [progressive web app (PWA)](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps] of the Bluesky web interface that I can install on my phone. That would be comparable to Twitter, Mastodon, and T2, all of which offer this. 

**Describe alternatives you've considered**

I just use Bluesky in my browser. I'm not interested in installing the app at present as it doesn't seem to offer a significant advantage. 

**Additional context**

TheHCJ:
Very easy to make it an installable PWA all you need to add is a manifest.json

scottdotjs:
@TheHCJ Technically yes, but with a service worker, caching, the usual stuff that makes it more than just "a site that opens in its own app window", you know, is a bit more effort than that. I'll happily wait till all that's in place.

scottdotjs:
Just updating here to say that the app got significantly better not long after I originally placed this request and I've been using it ever since! But still think this would be nice to offer.

FinallyCanuck:
This would still be a good option for desktop users. (Yes, we still exist.)

waynesbrain:
I would love a standalone PWA. I don't think that you actually need a service worker for a standalone installation to work. What I've found is that you need a service worker to *offer* installation. However, on an iPhone where there is no offer allowed, the Add to Home Screen action would work.

My goal is to escape the tyranny of the Apple app store wherever possible without buying a phone from an ad-company, so I love using PWAs and the PWA for X works great.

waynesbrain:
One trick that I've found to force standalone mode on any site is to use a userscript like [this one](https://greasyfork.org/en/scripts/490784-pwas-anywhere) along with the iOS [Userscripts extension](https://apps.apple.com/us/app/userscripts/id1463298887) and when I try it on BSky soon I'll update this comment to let anyone who's interested know how well it works...

Meanwhile, `Install page as app...` works fine on desktop Chrome to launch a standalone BlueSky.

thegreatGreenstar:
This suggestion is similar to https://github.com/bluesky-social/social-app/issues/1584 so I linked it to mention back to here.
[...]

Issue #1584, opened 2023-10-02
PWA: Convert website into a real app.


peerem:
Why don't you turn your website into a real app (Progressive Web Apps/PWA)?

Simple instructions for converting the website can be found here.

https://medium.com/@OPTASY.com/3-essential-steps-to-convert-your-website-to-a-progressive-web-app-c63d7c1c6322

The icons for all platforms can be created here. The icon Json for the manifest is also generated there at the same time.

https://www.pwabuilder.com/imageGenerator

An example of an (almost) complete manifest can be found here:

https://desktop.peerem.com/manifest.json

By the way, Twitter is also an example of a good PWA!

https://twitter.com/manifest.json

intrnl:
btw as a side note, might want to look into inverting the document title from `XXX - Bluesky` to `Bluesky - XXX` or even remove the app name at all if PWA is detected

this is because Chrome tries to omit the app name off the page title, but it only does so if the page title *starts* with the app name.

Doing `XXX - Bluesky` would just make the actual window title be `Bluesky - XXX - Bluesky`

thegreatGreenstar:
This suggestion is similar to https://github.com/bluesky-social/social-app/issues/1584 so I linked it to mention back to here.
BlueSky should be published to Microsoft Store as a [PWA](https://web.dev/learn/pwa/progressive-web-apps), so a [PWA version](https://developer.chrome.com/blog/getting-started-pwa/#progressive_web_apps_are) of BlueSky would be great.
When done, this will also allow the site to be [published in multiple other app stores as a PWA](https://web.dev/articles/pwas-in-app-stores), which should make updating easier.
Also, this would allow BlueSky to be [installed directly in browser](https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/#requirements) on pcs which again, improves user experience.

thegreatGreenstar:
This suggestion is similar to https://github.com/bluesky-social/social-app/issues/1584 so I linked it to mention back to here.
BlueSky should be published to Microsoft Store as a [PWA](https://web.dev/learn/pwa/progressive-web-apps), so a [PWA version](https://developer.chrome.com/blog/getting-started-pwa/#progressive_web_apps_are) of BlueSky would be great.
When done, this will also allow the site to be [published in multiple other app stores as a PWA](https://web.dev/articles/pwas-in-app-stores), which should make updating easier.
[...]

PR #2919, opened 2024-02-18
Add manifest.json and related files

edo-bari-ikutsu:
By adding manifest.json, web page can be installed as application. I also have added icons optimized for PWA.

pwa001

Note: Currently offline support with service worker is not implemented, thus I set `display` setting of manifest.json to `minimal-ui` to let users use reload button.

pwa002


surfdude29:
It looks like this would fix #1194 and #1584

hichemfantar:
What's the difference between this and the existing manifest in the build folder?
image

PR #2922, opened 2024-02-18
Finnish localization

jaoler:
Finnish (fi) translation for Bluesky. The translation has been done and reviewed by three individuals who speak Finnish as their native language.


msarjakoski:
Having reviewed the Finnish translations I feel they are very well crafted and that they fit very well as the first public version. I might have some minor suggestions, but I think it would be better to see the translations live in the real context.

Yangbblood (commented):
Translation appears ok to me!

valoisin:
Very good translation. Hyvää työtä.

jaoler:
Added new untranslated strings in to Finnish messages.po. I accidentally edited the English file first, then replaced it with the original. Everything in okay now.

pfrazee (approved):
Wonderful! Thank you!

PR #2923, opened 2024-02-19
Fix typos and clarify two `accessibilityHint`s

surfdude29:
While I was working on updating the German localisation I noticed a couple of typos and one `accessibilityHint` that didn't seem to quite make sense. This small PR corrects those.

Edited to add: I found one more `accessibilityHint` that I think could be made a bit clearer, so I have added that to this PR.


pfrazee (approved):
Thanks!

PR #2930, opened 2024-02-19
Update IS_PROD check

haileyok:
`IS_PROD` should only be false if it's a local dev service of the bsky staging service.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -5,7 +5,6 @@ export const LOCAL_DEV_SERVICE =
 export const STAGING_SERVICE = 'https://staging.bsky.dev'
 export const PROD_SERVICE = 'https://bsky.social'
 export const DEFAULT_SERVICE = PROD_SERVICE
-
 const HELP_DESK_LANG = 'en-us'
 export const HELP_DESK_URL = `https://blueskyweb.zendesk.com/hc/${HELP_DESK_LANG}`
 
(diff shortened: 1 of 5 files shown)


src/lib/constants.ts
```diff
@@ -36,92 +35,12 @@ export const MAX_GRAPHEME_LENGTH = 300
...
-  // this definition will not work once federation is enabled!
-  // -prf
-  return (
-    url.startsWith('https://bsky.social') ||
-    url.startsWith('https://api.bsky.app') ||
-    /bsky\.network\/?$/.test(url)
-  )
+export function IS_PROD_SERVICE(url?: string) {
+  return url && url !== STAGING_SERVICE && url !== LOCAL_DEV_SERVICE
 }
```
haileyok, line 40: The only cases where we shouldn't be in prod are staging and local right?

src/lib/constants.ts
```diff
@@ -135,13 +54,11 @@ export const STAGING_LINK_META_PROXY =
...
-    return STAGING_LINK_META_PROXY
-  } else if (IS_STAGING(serviceUrl)) {
-    return STAGING_LINK_META_PROXY
-  } else {
+  if (IS_PROD_SERVICE(serviceUrl)) {
     return PROD_LINK_META_PROXY
   }
+
+  return STAGING_LINK_META_PROXY
 }
```
haileyok, line 63: Again, we should always use cardyb as long as we are not on staging regardless of service iiuc.

estrattonbailey (approved):
LGTM! Thanks for cleaning all that up. Just wanna get a triple check from @dholms on prod vs everything else logic.

dholms (approved):
yup looks right 👍

PR #2931, opened 2024-02-19
Make bio area scrollable on iOS

haileyok:
## Why

Our implementation of a header with a tab bar has caused us quite a bit of pain for a while now on iOS. Unfortunately, because of how iOS handles touches, we were unable to properly scroll in the area of a user bio that is pressable - i.e. in the bio text, the followers/following links, or the Edit Profile button.

This PR addresses those issues by adding some gesture detectors to the profile header. Those detectors can then be used to update the underlying `UIScrollView`'s content offset and to trigger a refresh of the scroll view's content.

In an effort to make this act as closely as possible to a normal `UIScrollView` (i.e. just a `<ScrollView>` with a `<RefreshControl>`, we needed to make some changes to React Native's `RCTRefreshControl `. Most of the methods and variables on the `RCTRefreshControl` are private, so we needed to expose a new method, `forwarderBeginRefreshing()` to handle triggering a refresh.

I've used the Expo Modules API for this as that will allow us to use this same code with Fabric, rather than have to rewrite or update it whenever we migrate to Fabric.

### `findNodeHandle()`

The first and smallest part of the code happens on the JS side. We need to tell our `ExpoScrollForwarderView` which `RCTScrollView` it should be modifying. To do that, we need to get the native tag of the view via `findNodeHandle()`.

An initial concern we had with this was having to use `svRef.current.getNativeScrollRef()._nativeTag`. As it turns out, this was a valid concern. A few libraries - including `@gorhom/bottom-sheet` - have used this method of obtaining the native view. While this method _does_ work right now with Paper, when migrating to Fabric this method seems to have broken. See https://github.com/discord/react-native-bottom-sheet/pull/11/files. `findNodeHandle()` however does work with Fabric.

`findNodeHandle()` is also used in React Native's own `ScrollView` component, for `svRef.current.getScrollableNode()`. https://github.com/facebook/react-native/blob/6014dce04c86fde5c7d4050859fca20007ff1200/packages/react-native/Libraries/Components/ScrollView/ScrollView.js#L845. 

> In fact, we could actually use `getScrollableNode()` directly here, however, I am opting to use `findNodeHandle()` instead, because it is properly typed as returning `number | null` whereas `getScrollableNode()` is typed as returning `any | null`.

We use a `useEffect()` to update the native tag on tab changes. From my testing, this has not caused any problems. It's a bit unfortunate that we need to prop drill to get a `setState()` down to where we need to call it at, but oh well.

### `RCTRefreshControl` Patch

One additional piece of code that we needed to implement for this to act nicely was a little patch to `RCTRefreshControl`. Most of the methods in `RCTRefreshControl` are not exposed, and as a result we couldn't just do this right inside of the new `ExpoScrollForwarderView` module.

The patch adds a new method - `forwarderBeginRefreshing()` to `RCTRefreshControl` that mimics the normal behavior of the component. Whenever our scroll view reaches a content offset of -130 (only when we are scrolling from within the header, otherwise we just let the `RefreshControl` act normally) we call this new method, mostly mimicing the behavior that we expect.

```objc
- (void)forwarderBeginRefreshing
{
  _refreshingProgrammatically = NO;
  
  [self sizeToFit];
  
  if (!self.scrollView) {
    return;
  }
  
  UIScrollView *scrollView = (UIScrollView *)self.scrollView;
  
  [UIView animateWithDuration:0.3
    delay:0
    options:UIViewAnimationOptionBeginFromCurrentState
    animations:^(void) {
      // Whenever we call this method, the scrollview will always be at a position of
      // -130 or less. Scrolling back to -65 simulates the default behavior of RCTRefreshControl
      [scrollView setContentOffset:CGPointMake(0, -65)];
    }
    completion:^(__unused BOOL finished) {
      [super beginRefreshing];
      [self setCurrentRefreshingState:super.refreshing];
    
      if (self->_onRefresh) {
        self->_onRefresh(nil);
      }
    }
  ];
}
```

### `ExpoScrollForwarder`

Most of the new logic is inside of this new module. There are a few different pieces here to take note of.

#### Finding the underlying `RCTScrollView`

Whenever we update the `scrollViewTag` prop, we call `tryFindScrollView()`. This method will attempt to find an `RCTScrollView` with a tag that matches the `scrollViewTag` prop. If we do find one, we also add two gesture detectors to the view (see below).

#### `cancelGestureRecognizers`

Because there is some new code that is responsible for modifying the scroll position of the view - which runs separate of the real `UIScrollView` logic for scroll animations, velocity, etc - we need to ensure that we _stop_ that animation whenever we touch the actual `UIScrollView`. For example:

- User swipes down the screen from within the header
- The expected decay scroll begins
- The user swipes again - now within the `UIScrollView`
- Because the user landed a finger inside of the `UIScrollView`, we no longer need to run the custom scroll animation to completion

This isn't a super easy task - we need to make sure that our new gesture recognizers do not interfere with the `UIScrollView`'s `UIPanGestureRecognizer`. We also need to make sure that a tap within the `UIScrollView` does not accidentally trigger a tap on a post or other button. To achieve this, we add a `UITapGestureRecognizer` as well as a `UILongPressGestureRecognizer` to the `UIScrollView`. 

The former is used only to intercept taps - so that we do not accidentally trigger any navigation events, post likes, etc. whenever we bring the scroll to a halt.

The latter is used to actually cancel the scroll animation. It is a `UILongPressGestureRecognizer` with a `minimumPressDuration` of 0.01, which will result in a nearly immediate halt of the previous animation and let the `UIScrollView` take over animations on its own. (We can't use a `UIPanGestureRecognizer` here because it would interfere with the `UIScrollView`'s `UIPanGestureRecognizer`)

Every time we switch tabs - i.e. switch from "Posts" to "Lists" - we remove the cancel recognizers from the current `UIScrollView` and add them to the new one that we are switching to. This happens inside of `tryFindScrollView()`.

#### The main `UIPanGestureRecognizer`

The main `UIPanGestureRecognizer` is added to the `UIView` - the view that contains our header. This operates in the same way as (or rather is the underlying mechanism behind of) Reanimated's `Gesture.Pan()`.

We use this recognizer to "simulate" and "forward" the scroll events to the `UIScrollView`. We run a bit of logic to "dampen" the scroll whenever we end up at an offset of < 0, handle refresh logic whenever we reach an offset of -130, and apply a decay "animation" all by using this gesture recognizer's events.

## Caveats

This operates _very_ closely to the real thing. However, there are a few quirks that I would definitely like to solve, although I believe they would require a more extensive (i.e. a completely custom `UIScrollView` module) to achieve.

### The decay animation

We would like to be able to use `UIView.animate` (https://developer.apple.com/documentation/uikit/uiview/1622418-animate) to achieve this. Unfortunately, from my testing this interferes with the `FlatList`'s virtualization. As a result, we instead use a timer - which runs on every frame - to update the scroll position instead. This isn't optimal but it appears to be the best way to achieve what we are looking for. I'll play with this some more in the future to see if a better alternative could come up

### Quirky animation on the `UIRefreshControl`

There's a bit of a strange animation that occurs with the refresh control here. This - from what I can tell - is a result of "hacking" the entire thing to do this programatically. There may be a better alternative here (I suspect that looking more closely at the `RCTScrollView` view might turn up something), but for right now this is a _very_ close replica that I feel comfortable shipping (I was not comfortable shipping the prior iteration I did of this, where the refresh control was extremely janky).

## Video

https://github.com/bluesky-social/social-app/assets/153161762/d980541c-3bb2-4740-9c10-7339537dd23a

## Test Plan

We should use this internally for a bit to find any quirks. I am sure there will be some that come up, although I feel confident at this point that a few small iterations should get us extremely close to what we want.

I've tested switching between tabs without issue, including starting each tab at a different scroll position. I've also tested opening multiple windows with profiles.

Additionally, I've done some "attempts" at breaking it. I.e., switch tabs and ***immediately*** start scrolling to try and find a delay between the tab switch and the calling of `tryFindScrollView()`. So far I have not been able to turn up anything.

And of course, I have done some pretty significant testing of the refresh control, and aside from the above noted quirk, have not seen anything that has caused problems.

Lastly, I've tested the gate by setting the percentage to 100% of internal users and verified the scroll behavior works, then setting it to 0% of internal users and verified that it is disabled.

## General Usage

You should have two views: a `ScrollView` (or a `FlatList`, `VirtualizedList`, etc. that uses `RCTScrollView`) and another
view that you want to detect pan on to apply to the `RCTScrollView`.

Next, you need to get the `_nativeTag` of your `ScrollView`, like below:

```tsx
const [scrollViewTag, setScrollViewTag] = React.useState<number | null>()

return (
  <ScrollView ref={(ref) => {
	setScrollViewTag(findNodeHandle(ref))
  }}>
    <View>
      <Text>Some children</Text>
    </View>
  </ScrollView>
)
```

Next, wrap the view you're trying to detect gesture on inside of `ScrollForwarder`.

```tsx
import { ScrollForwarder } from 'react-native-scroll-forwarder/src/ScrollForwarder'

return (
  <ScrollForwarder scrollViewTag={scrollViewTag}>
    <View>
      <Text>Probably a header for a list</Text>
    </View>
  </ScrollForwarder>
)
```


modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
```diff
@@ -0,0 +1,169 @@
...
+      }
+    }
+  }
+  
+  func dampenOffset(_ offset: CGFloat) -> CGFloat {
+    // We can probably uncomment this later, but for now let's just not allow refreshing
+    //    if offset < 0 {
+    //      return offset - (offset * 0.55)
+    //    }
+
```
haileyok, line 112: Adding `refreshControl` support to this is pretty janky. I think the move might be to apply a few changes to the actual `RCTRefreshControl` instead. There are a few private methods there that we could use to our advantage if we make them public. This is something that I'd rather do in a separate PR though, so I say let's test the scroll behavior in general before applying patches to RN.

haileyok:
I've gotten a bit closer here to reducing the changes needed on the JS side, so I want to spend a bit of time trying to clean that up more before we decide to merge this in.

modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
gaearon: What's the consequence in UI if we don't implement this?

haileyok: not great. i'm going to spend some time to see if i can find a good way in this PR.

not having this will prevent pull to refresh behavior which is not great. the problem is that the underlying RCTRefreshControl does some of its own logic (inside of a private method that we can't access here), so there's an overlap in us calling beginRefreshing here and the underlying method calling it as well.

will give this some thought, because i'd like to ship with refreshControl support since it's a pretty crucial thing to have.

haileyok:
@gaearon I feel a lot more comfortable about `findNodeHandle()` (see above for reasoning on that) and have not been able to produce any issues here. Will continue testing though to see.

Do you feel that the `useEffect()` is the correct way to update the `scrollViewTag` value/prop?

haileyok:
Ah, missing animation for scrolling back up if the offset is < 0 and > -130 😭 will fix tomorrow.

haileyok:
Alright, fixed the missing animation as well as fixed interference with the native swipe back gesture.

PR #2934, opened 2024-02-19
Swap `RichText`

estrattonbailey:
Changes between old and new `RichText`:
- text size now defaults to 14, was 15
- line height now defaults to 1.3, was 1.2

Posts in feeds and threads have no (or very very minimal) visual changes. After/before (reverse order sorry) shots:

Group 1
Group 2
Group 3
Group 4
Group 5

### DIFF
--- a/src/alf/atoms.ts
+++ b/src/alf/atoms.ts
@@ -176,33 +176,43 @@ export const atoms = {
   },
   text_2xs: {
     fontSize: tokens.fontSize._2xs,
+    letterSpacing: 0.25,
   },
   text_xs: {
     fontSize: tokens.fontSize.xs,
+    letterSpacing: 0.25,
   },
   text_sm: {
     fontSize: tokens.fontSize.sm,
+    letterSpacing: 0.25,
   },
   text_md: {
     fontSize: tokens.fontSize.md,
+    letterSpacing: 0.25,
   },
   text_lg: {
     fontSize: tokens.fontSize.lg,
+    letterSpacing: 0.25,
   },
   text_xl: {
     fontSize: tokens.fontSize.xl,
+    letterSpacing: 0.25,
   },
   text_2xl: {
     fontSize: tokens.fontSize._2xl,
+    letterSpacing: 0.25,
   },
   text_3xl: {
     fontSize: tokens.fontSize._3xl,
+    letterSpacing: 0.25,
   },
   text_4xl: {
     fontSize: tokens.fontSize._4xl,
+    letterSpacing: 0.25,
   },
   text_5xl: {
     fontSize: tokens.fontSize._5xl,
+    letterSpacing: 0.25,
   },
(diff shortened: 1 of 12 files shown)


src/alf/atoms.ts
```diff
@@ -176,43 +176,59 @@ export const atoms = {
   },
   text_2xs: {
     fontSize: tokens.fontSize._2xs,
+    letterSpacing: 0.25,
```
estrattonbailey, line 179: All our existing type styles have built in letter spacing, so for consistency I've added here too

src/alf/atoms.ts
```diff
@@ -176,43 +176,59 @@ export const atoms = {
...
   },
   leading_normal: {
     lineHeight: 1.5,
   },
+  tracking_normal: {
+    letterSpacing: 0,
+  },
+  tracking_wide: {
+    letterSpacing: 0.25,
+  },
```
estrattonbailey, line 231: Couple more Tailwind atoms

src/alf/atoms.ts
```diff
@@ -176,43 +176,59 @@ export const atoms = {
...
   text_5xl: {
     fontSize: tokens.fontSize._5xl,
+    letterSpacing: 0.25,
   },
   leading_tight: {
     lineHeight: 1.15,
   },
   leading_snug: {
-    lineHeight: 1.25,
+    lineHeight: 1.3,
```
estrattonbailey, line 221: Post text defaults to 1.3, so for consistency through, I bumped this up to match. 0.5 difference is negligible, and we hardly use this atom so far elsewhere.

src/components/Link.tsx
```diff
@@ -82,7 +83,9 @@ export function useLink({
 
   const onPress = React.useCallback(
     (e: GestureResponderEvent) => {
-      outerOnPress?.(e)
+      const exitEarlyIfFalse = outerOnPress?.(e)
+
+      if (exitEarlyIfFalse === false) return
```
estrattonbailey, line 88: Something I've been meaning to do, was gonna use for tags but going to do something else now. Might as well include.

src/components/RichText.tsx
```diff
@@ -29,7 +29,7 @@ export function RichText({
   const [richText, setRichText] = React.useState<RichTextAPI>(() =>
     value instanceof RichTextAPI ? value : new RichTextAPI({text: value}),
   )
-  const styles = [a.leading_normal, style]
+  const styles = [a.leading_snug, flatten(style)]
```
estrattonbailey, line 32: Default's to what we use for post text. `leading_normal` (1.5) is too wide for anything but longform, and `tight` (1.15) is better for titles. `snug` (1.3) is a good default.

Previous `RichText` component defaulted to 1.2, which is a little tight imo.

haileyok (approved):
LGTM!

PR #2936, opened 2024-02-20
Remove async resizing from external embed player

haileyok:
Instead of using `measure()`, we can just use `aspectRatio` to determine the size of the player. This fixes some jumps when opening threads with player embeds.

We were also able to simplify the styles considerably. The player can use a single wrapper for the `WebView` vs the three from before.

### DIFF
--- a/src/lib/strings/embed-player.ts
+++ b/src/lib/strings/embed-player.ts
@@ -343,45 +343,45 @@ export function parseEmbedPlayerFromUrl(
   }
 }
 
-export function getPlayerHeight({
+export function getPlayerAspect({
   type,
-  width,
   hasThumb,
+  width,
 }: {
   type: EmbedPlayerParams['type']
-  width: number
   hasThumb: boolean
-}) {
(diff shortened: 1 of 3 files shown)


src/lib/strings/embed-player.ts
```diff
@@ -343,45 +343,45 @@ export function parseEmbedPlayerFromUrl(
...
+  width: number
+}): {aspectRatio?: number; height?: number} {
+  if (!hasThumb) return {aspectRatio: 1.78}
 
   switch (type) {
     case 'youtube_video':
     case 'twitch_video':
     case 'vimeo_video':
-      return (width / 16) * 9
+      return {aspectRatio: 1.78} // This is 16:9
```
gaearon, line 361: Could also just write `16 / 9` etc?

haileyok: Heh yea duh, more clear that way.

gaearon (approved):
lgtm. thank you!

PR #2956, opened 2024-02-21
Fix `numberOfLines` not updating on iOS 15

haileyok:
Beats me as to why, but the main view does not have access to shadow properties on iOS 15. We instead need to have them as both a shadow property and a view property if we want to access the property from both.


pfrazee (approved):
great!

PR #2959, opened 2024-02-22
Add handle validation to create account UI

haileyok:
This adds some validation to handles in the UI. Right now we are not doing any validation aside from checking if the handle is already registered, so the backend receives a lot more invalid handles than it should.

Using regex from https://github.com/bluesky-social/indigo/blob/main/atproto/syntax/handle.go#L10 and other logic from https://github.com/bluesky-social/atproto/blob/main/packages/pds/src/handle/index.ts#L72

While we technically are only allowing handles with a length of <= 30, this sounds like it might change so I'm leaving that validation out and letting the service handle that.

https://github.com/bluesky-social/social-app/assets/153161762/30cd6217-3c14-43ec-b43e-83eb990bc1e9

### DIFF
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -1,3 +1,8 @@
+// Regex from the go implementation
+// https://github.com/bluesky-social/indigo/blob/main/atproto/syntax/handle.go#L10
+const VALIDATE_REGEX =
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Looks great

PR #2966, opened 2024-02-22
Make dim theme dim

gaearon:
Screenshot 2024-02-22 at 16 47 22


src/alf/themes.ts
```diff
@@ -132,21 +132,21 @@ export const darkPalette: Palette = {
...
-  contrast_500: tokens.color.gray_500,
-  contrast_600: tokens.color.gray_400,
-  contrast_700: tokens.color.gray_300,
-  contrast_800: tokens.color.gray_200,
-  contrast_900: tokens.color.gray_100,
-  contrast_950: tokens.color.gray_50,
-  contrast_975: tokens.color.gray_25,
+  contrast_25: `hsl(211, 28%, 14%)`,
+  contrast_50: `hsl(211, 28%, 17%)`,
+  contrast_100: `hsl(211, 28%, 23%)`,
```
haileyok, line 139: We use this for the border in `lib/themes.ts`. This makes sense I think whenever we are dealing with solid white/black, but I feel like in this case we might want something a bit less "blue"? Maybe it's just me, but on some monitors this is just slightly too blended in with the background for the borders.

haileyok, line 139: We use this for the border in lib/themes.ts. This makes sense I think whenever we are dealing with solid white/black, but I feel like in this case we might want something a bit less "blue"? Maybe it's just me, but on some monitors this is just slightly too blended in with the background for the borders.

haileyok: (Specifically `contrast_100`)

haileyok (approved):
Much better now imo, thanks eric!

haileyok:
Updated image:

image

pfrazee (approved):
This isn't final but it gets our contrasts to a usable point

PR #2967, opened 2024-02-22
Expo library migration, dependency culling

haileyok:
This started as a migration of some libraries to Expo equivalents, but there was a lot of unused or unnecessary libraries lingering around here as well. This is an attempt to do a cleanup of those.

There are also a few fixes for existing bugs as a result of these changes as noted when applicable.

## Migrate Libraries

- [x] Bump all Expo packages to latest
- [x] Remove `@react-native-community/blur`. This was used in a single place, seemingly without any particular need. Pretty sure our intention was to not use any blur views anywhere in the app anyway.
- [x] Remove `rn-fetch-blob` in favor of `expo-file-system`. See below for more
- [x] Migrate `react-native-version-number` to `expo-application`
- [x] Migrate `react-native-clipboard` to `expo-clipboard`
- [x] Migrate `react-native-fs` to `expo-file-system`
- [x] Migrate `react-native-haptics` to `expo-haptics`
- [x] Migrate `react-native-image-resizer` to `expo-image-manipulator`
    - This fixes an issue on Android where the resizer would save the image with the same name sometimes (honestly not sure if it was library related or some of the other refactors around the code, but the issue is no longer shown).
- [x] I'm leaving `react-native-image-crop-picker` alone for now. `expo-image-picker` nor `expo-image-manipulator` (that I can tell) offer a way to do crops the way we want to natively. If necessary, we can build another module.
- [x] Migrate `react-native-linear-gradient` to `expo-linear-gradient` 

## Image manipulation and downloading

- Removing `rn-fetch-blob` and using `expo-file-system` for downloads instead. All the same features are supported (i.e. timeouts)
- Improving the download logic and disposal afterwards
- Removing unnecessary `isIos` share logic (previously looks like `expo-sharing` didn't work on iOS, but this seems to be fixed now)
- Prefer cached images whenever available. We will almost always have the image available when we want to share it or download it from the lightbox, so no need to download it again.

## Other dropped libraries (unused)

- [x] `@react-native-camera-roll/camera-roll`
- [x] `react-circular-progress-bar`
- [x] `react-native-appstate-hook`
- [x] `react-native-web-linear-gradient`
- [x] `use-deep-compare`
- [x] `normalize-url`
- [x] `lru_map` 

 ## Test replaced libraries

- [x] `expo-application`
	- [x] Android
	- [x] iOS
	- [x] Web
- [x] `expo-clipboard`
	- [x] Android
	- [x] iOS
	- [x] Web
- [x] `expo-file-system`
	- [x] Android
	- [x] iOS
	- [x] Web
- [x] `expo-haptics`
	- [x] Android
	- [x] iOS
	- [x] Web (N/A)
- [x] `expo-image-manipulator`
	- [x] Android
	- [x] iOS
	- [x] Web
- [x] `expo-linear-gradient`
	- [x] Android
	- [x] iOS
	- [x] Web
- [x] Image manipulation changes for saving
    - [x] Android
    - [x] iOS
    - [x] Web (N/A)


haileyok:
Shouldn't merge this until after 1.69 release if that's going to happen this week. Should do some testing to make sure everything lined up correctly.

src/lib/media/manip.ts
```diff
@@ -147,24 +153,35 @@ interface DoResizeOpts {
...
 
 async function doResize(localUri: string, opts: DoResizeOpts): Promise<Image> {
+  // We should run the first iteration outside of the loop. This is the only one that should resize the image and the
+  // loop can deal with lowering the quality if needed. Iterating on the resize causes problems on iOS
+  let resizeRes = await manipulateAsync(
+    localUri,
+    [{resize: {height: opts.height, width: opts.width}}],
+    {format: SaveFormat.JPEG, compress: 1},
+  )
+
```
haileyok, line 163: Seems like the reason we may have been using `react-native-image-resizer` was because of a weird issue on iOS with the image getting resized to the wrong dimensions. What seems to happen is that if we iterate more than once on resizing it, the dimensions get wonky.

Moving this outside of the loop and only iterating on the compression resolves that problem.

PR #2968, opened 2024-02-22
Add tags and mute words

estrattonbailey:


src/lib/moderatePost_wrapped.ts
```diff
@@ -2,18 +2,72 @@ import {
...
+    )
+    .map(t => t.toLowerCase())
+
+  const words = text.toLowerCase().split(/\s+/)
+
+  for (const mute of mutedWords) {
+    const mutedWord = mute.value.toLowerCase()
+
+    // `content` applies to tags as well
+    if (tags.includes(mutedWord)) return true
```
haileyok, line 42: I think I agree with this. If we mute a word in the text, it should be muted in both the regular text *and* the tags. If we're going to do that though, I think we should force `Mute in tags` to be true in the dialog. I partly expected the tags to still show up but not in the text. Minor, but could be confusing.

src/components/dialogs/MutedWords.tsx
```diff
@@ -0,0 +1,317 @@
...
+            label={_(msg`Mute this tag`)}
+            name="tag"
+            value={tagTargetEnabled}
+            onChange={setTagTargetEnabled}
+            style={[a.flex_1]}>
+            <TargetToggle>
+              <View style={[a.flex_row, a.align_center, a.gap_sm]}>
+                <Toggle.Checkbox />
+                <Toggle.Label>
+                  <Trans>Mute in tags</Trans>
```
haileyok, line 134: Other comment is regarding this checkbox. Do you think we should just have the value be `contentTargetEnabled || tagTargetEnabled` and disable this whenever `contentTargetEnabled` is true?

src/view/screens/Moderation.tsx
```diff
@@ -81,6 +83,23 @@ export function ModerationScreen({}: Props) {
...
+          style={[styles.linkCard, pal.view]}
+          onPress={() => mutedWordsDialogControl.open()}
+          accessibilityRole="tab"
+          accessibilityHint=""
+          accessibilityLabel={_(msg`Open muted words settings`)}>
+          <View style={[styles.iconContainer, pal.btn]}>
+            <FontAwesomeIcon
+              icon="eye"
+              style={pal.text as FontAwesomeIconStyle}
+            />
```
haileyok, line 97: Could we use this?
Screenshot 2024-02-22 at 4 09 00 PM

src/view/com/util/forms/PostDropdownBtn.tsx
```diff
@@ -210,6 +212,20 @@ let PostDropdownBtn = ({
...
     },
+    hasSession && {
+      label: _(msg`Mute words & tags`),
+      onPress() {
+        mutedWordsDialogControl.open()
+      },
+      testID: 'postDropdownMuteWordsBtn',
+      icon: {
+        ios: {
+          name: 'speaker.slash',
```
haileyok, line 223: Mmm I don't think there's a funnel SFSymbol...Not sure what we'd use for this rather than the speaker slash. There's some options, but they don't like up with what we have for icons in the pack :(

src/components/dialogs/MutedWords.tsx
```diff
@@ -0,0 +1,317 @@
...
+            size="small"
+            color="primary"
+            variant="solid"
+            style={[!gtMobile && [a.w_full, a.flex_0]]}
+            onPress={submit}>
+            <ButtonText>
+              <Trans>Add</Trans>
+            </ButtonText>
+            <ButtonIcon icon={Plus} />
+          </Button>
```
haileyok, line 156: Should probably disable this and display a spinner or something in the button while we save the new tag.

haileyok:
Really awesome work, this works great and looks even better!

Just a few visual nits that I noticed. Also, what do we think about just pushing to a search screen on native rather than taking the user to the search tab? My initial muscle memory was to "swipe back" after going to the search screen. (I know this is temporary of course, so it's probably not the biggest deal, but I think the UX is better regardless)

src/components/RichText.tsx
```diff
@@ -136,3 +161,79 @@ export function RichText({
...
+  }, [control])
+
+  /*
+   * N.B. On web, this is wrapped in another pressable comopnent with a11y
+   * labels, etc. That's why only some of these props are applied here.
+   */
+
+  return (
+    <React.Fragment>
+      <TagMenu control={control} tag={tag} authorHandle={authorHandle}>
```
pfrazee, line 201: Oh this is really clever how you handled both web and mobile, well done man

src/lib/moderation.ts
```diff
@@ -67,6 +67,13 @@ export function describeModerationCause(
       description: 'You have hidden this post',
     }
   }
+  // @ts-ignore Temporary extension to the moderation system -prf
+  if (cause.type === 'muted-word') {
+    return {
+      name: 'Post hidden by muted word',
+      description: `You've chosen to hide a word or tag within this post.`,
```
pfrazee, line 74: Oh bleh. We've got this thing rewritten in the moderation PR to be easy to localize but it's not easy here. Not a trivial change either IIRC.

src/lib/moderatePost_wrapped.ts
```diff
@@ -29,15 +83,65 @@ export function moderatePost_wrapped(
...
+              embedHidden || hasMutedWord(mutedWords, image.alt, [], [])
+          }
+        }
+      }
     }
     if (
       AppBskyEmbedRecordWithMedia.isView(subject.embed) &&
       AppBskyEmbedRecord.isViewRecord(subject.embed.record.record)
     ) {
+      // TODO what
```
pfrazee, line 144: If we decide to leave this edge-case, it's trivial in the new moderation rework

src/lib/moderatePost_wrapped.ts
```diff
@@ -2,18 +2,72 @@ import {
...
+    // check individual character groups
+    for (const word of words) {
+      if (word === mutedWord) return true
+
+      const wordNoPunc = word.replace(/\p{P}+$/gu, '')
+      const mutedWordNoPunc = mutedWord.replace(/\p{P}+$/gu, '')
+
+      if (mutedWordNoPunc === wordNoPunc) return true
+    }
+  }
```
pfrazee, line 61: If we want to enable support for phrases, I think what we can do is this:

- Instead of splitting the text into words, run your punctuation strip on it
- Convert the muteword into a regex like this (havent tested this regex): `new RegExp('[\s\t\n\r\f\,\.\;\!\?\_\-\"\(\)]' + stripPunctuation(muteWord) + '[\s\t\n\r\f\,\.\;\!\?\_\-\"\(\)]')`
- Just run the muteword regexes against the punctuation-stripped post text

pfrazee (commented):
Awesome work Eric. Really just into the tweaking stage.

src/lib/moderatePost_wrapped.ts
mary-ext: Here's how Mastodon is more or less doing its matching:

https://github.com/mary-ext/langit/blob/70cf8ed820d27337370039f517f4968c6553cd11/app/desktop/components/settings/settings-views/keyword-filters/KeywordFilterFormView.tsx#L258

src/components/TagMenu/index.web.tsx
```diff
@@ -0,0 +1,125 @@
...
+          web: 'magnifying-glass',
+        },
+      },
+      {
+        label: _(msg`See ${tag} posts by this user`),
+        onPress() {
+          navigation.navigate({
+            name: 'Search',
+            params: {
+              q: tag + (authorHandle ? ` from:${authorHandle}` : ''),
```
mary-ext, line 67: No checks for invalid handle? Could probably use their DID as a fallback so their posts are at least searchable

src/lib/moderatePost_wrapped.ts
estrattonbailey: Thanks for the link! Will have a look and compare 🙏

src/components/TagMenu/index.web.tsx
estrattonbailey: Ah good call! Yeah should fall back to something, will def get that in, thank you

pfrazee:
Hot diggity dog

dolciss:
Glad to see this PR merged🎉, but forgive me for commenting when it is already closed.

Mute words appear to split the body text at the following locations,
https://github.com/bluesky-social/social-app/blob/e2ae534fde4a351ac4c8ebf18a492f0fc6ab7abc/src/lib/moderatePost_wrapped.ts#L62
However, this often fails to mute words in languages that are not separated by spaces, such as Japanese.
https://github.com/bluesky-social/social-app/blob/e2ae534fde4a351ac4c8ebf18a492f0fc6ab7abc/src/lib/__tests__/moderatePost_wrapped.test.ts#L74-L91
There are testing mute of a single kanji character here, but we often mute Japanese characters on a word-by-word basis as well. For example, "ハッシュタグ"(means Hashtags) in "次のバージョンでハッシュタグが実装される！"
It would be great if you could consider a non word-by-word word mute option!

estrattonbailey:
@dolciss no problem at all! Appreciate you bringing this to our attention. We've considered adding options for allowing users to select which type of matching they'd like to do, and we may continue that investigation. 

For now, we pushed a [new change](https://github.com/bluesky-social/social-app/pull/3018/commits/b3adcf6708fbef6f8d2cb9ea2c48d417a0dbae9f) just now that checks for the post language, and if it's a language that doesn't use spaces — or uses spaces in ways other than separating words — then we check for the string directly instead of word-by-word.

In English, this direct matching would return false-positives e.g. `AI` would mute the word `brain`.

If you don't mind, may I ask: with [this new change](https://github.com/bluesky-social/social-app/pull/3018/commits/b3adcf6708fbef6f8d2cb9ea2c48d417a0dbae9f), do you happen to know if this type of partial-match issue will be a concern in Japanese as well?

dolciss:
@estrattonbailey Thank you very much for your improve at #3018 ! 

> If you don't mind, may I ask: with [this new change](https://github.com/bluesky-social/social-app/pull/3018/commits/b3adcf6708fbef6f8d2cb9ea2c48d417a0dbae9f), do you happen to know if this type of partial-match issue will be a concern in Japanese as well?

Yes, partial-match problems are a concern in Japanese as well. (e.g. `ネット` would mute the word `インターネット` `ネットワーク`)
However, solving this problem is distressing because it requires complex processing, such as morphological analysis.
(I am sorry, but I am not that familiar with morphological analysis.)

src/lib/moderatePost_wrapped.ts
```diff
@@ -2,18 +2,122 @@ import {
...
+  for (const mute of mutedWords) {
+    const mutedWord = mute.value.toLowerCase()
+    const postText = text.toLowerCase()
+
+    // `content` applies to tags as well
+    if (tags.includes(mutedWord)) return true
+    // rest of the checks are for `content` only
+    if (!mute.targets.includes('content')) continue
+    // single character, has to use includes
+    if (mutedWord.length === 1 && postText.includes(mutedWord)) return true
```
DavidBuchanan314, line 52: This checks the utf16 code-unit length, which may give an unexpected result for certain international scripts and/or emojis (maybe it's fine? I'm not sure I totally understand the purpose of this check)

estrattonbailey: Good shout. Yeah the original intention was to handle international scripts with this. In response to the comments above I added [a check](https://github.com/bluesky-social/social-app/pull/3018/files#diff-844b6937574f4df0a4d948456421f7fad4fd358a62357bcaa7f00ba15bbab8e0R74) a couple weeks ago for a few languages that commonly don't use spaces as well.

But what you've called out here is still an issue: trying to match a single character (potentially with `length > 1`) in a language _not_ part of our exceptions may fail unless we can split the text at the word boundary. Interesting...

DavidBuchanan314: As a practical test-case if someone mutes "💩", then it it'll match when it's on its own, but not as part of a "word"
image

Muting an emoji is probably quite a common thing to want to do, and I think counting graphemes would give the expected behavior in most cases (which is normally an annoying thing to do, but I know the app already has grapheme counting logic in it elsewhere)

PR #2969, opened 2024-02-22
Norwegian localization

janlindso:
Translation for Norwegian Bokmål (nb). All strings are translated manually by native speaker/writer, and double checked with functionality when in doubt.


eklem:
This would be really nice to have!

pfrazee:
Thank you @janlindso! Do you think you could get a second norwegian speaker to verify the translation? We have a rule that there needs to be approval from a native speaker

eklem:
I can do than @pfrazee

pfrazee:
🙏 thanks so much!

eklem:
I'll try to finish the review over the weekend.

PR #2975, opened 2024-02-23
use `useOpenLink` hook for links in ALF

haileyok:
Replaces `Linking.openURL` with the `useOpenLink` hook in ALF links.

Also fixes a style error, `outline: 0` was an invalid style, supposedly?

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {GestureResponderEvent, Linking} from 'react-native'
+import {GestureResponderEvent} from 'react-native'
 import {
   useLinkProps,
   useNavigation,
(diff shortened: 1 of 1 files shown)


src/components/Link.tsx
```diff
@@ -260,7 +263,6 @@ export function InlineLink({
       style={[
         {color: t.palette.primary_500},
         (hovered || focused || pressed) && {
-          outline: 0,
```
estrattonbailey, line 263: Yeah leave this here, since it removes the ugly focus outlines on web. We have an underline to serve as a focus state. We should replace with this to avoid errors `...web({ outline: 0 })`.

estrattonbailey (approved):
Oh whoops, TIL `useOpenLink`

PR #2977, opened 2024-02-23
Remove top border from highlighted post when there's a parent, connect line to avatar (with better centering)

haileyok:
Right now it's not always obvious that there are parents when we open a post with parents

image

This makes it more obvious that there's a parent by making the line connect to the avatar
Screenshot 2024-02-23 at 1 36 00 PM

Still looks the same when we scroll up as well
Screenshot 2024-02-23 at 1 36 17 PM

And we don't make any changes if there are no parents
Screenshot 2024-02-23 at 1 36 29 PM


haileyok:
Going to put this on the shelf for now until we figure out the sticky header here. Still looking into fixing this though.

PR #2932, opened 2024-02-19
Adds `.nvmrc` file to `bskyweb`

EvandroLG:
This tiny PR introduces a `.nvmrc` file to the `bskyweb` project.

PR #2979, opened 2024-02-23
feat(nvm): add support for nvm & minor actions housekeeping

vinayakkulkarni:
- [x] add `.nvmrc` from #2932
- [x] update github action `setup-node` actions
- [x] update github action workflows to use `node-version-file` instead of hardcoded v18.x
- [x] bump nvm version installed in Dockerfile


haileyok:
Thanks for this! It seems like the actions would rather have a specific node version though?

vinayakkulkarni:
> Thanks for this! It seems like the actions would rather have a specific node version though?

hey @haileyok, the workflow runs are now successful :D

haileyok (approved):
Thank you!

PR #2981, opened 2024-02-23
Respect chosen theme for splash

gaearon:
We were reading the system theme color for the splash screen. Even if the app's theme is set up to be different. This resulted in a flash if the system theme is light but the app theme is dark. This fixes that.

### DIFF
--- a/src/Splash.tsx
+++ b/src/Splash.tsx
@@ -4,7 +4,6 @@ import {
   StyleSheet,
   Image as RNImage,
   AccessibilityInfo,
-  useColorScheme,
 } from 'react-native'
 import * as SplashScreen from 'expo-splash-screen'
 import {Image} from 'expo-image'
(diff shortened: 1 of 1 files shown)


src/Splash.tsx
```diff
@@ -74,8 +74,8 @@ export function Splash(props: React.PropsWithChildren<Props>) {
     isLayoutReady &&
     reduceMotion !== undefined
 
-  const colorScheme = useColorScheme()
-  const isDarkMode = colorScheme === 'dark'
+  const colorTheme = useColorModeTheme()
```
gaearon, line 77: This is the main change.

src/Splash.tsx
```diff
@@ -74,8 +74,8 @@ export function Splash(props: React.PropsWithChildren<Props>) {
...
+  const isLightMode = colorTheme === 'light'
```
gaearon, line 78: Flip the condition because we want to match both `'dark'` and `'dim'`. So do it by excluding `'light'` and flipping the cases below.

gaearon:
As discussed, this regresses on the initial splash image -> splash screen transition so let's shelve until we have some solution to that

PR #2992, opened 2024-02-27
Intent handler

haileyok:
Separating this logic out from https://github.com/bluesky-social/social-app/pull/2587. Making this a precursor just so we can figure out a good way to handle intent links while also being able to use hooks for things like opening modals.


src/Navigation.tsx
```diff
@@ -460,7 +460,8 @@ const FlatNavigator = () => {
  */
 
 const LINKING = {
-  prefixes: ['bsky://', 'https://bsky.app'],
+  // TODO figure out what we are going to use
+  prefixes: ['bsky://', 'bluesky://', 'https://bsky.app'],
```
haileyok, line 464: We still need to figure out what scheme we want to use before we actually release intents.

src/lib/intents/useIntentHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+      if (!isIntent) return
+
+      switch (intentType as IntentType) {
+        case 'compose': {
+          onComposeIntent({
+            text: params.get('text'),
+            imageUris: params.get('imageUris'),
+            openComposer,
+            hasSession,
+          })
```
estrattonbailey, line 37: Nit: feels kinda odd to me to pass in `openComposer` and `hasSession` here. Is our plan to re-use `onComposeIntent` elsewhere?

haileyok: This would only happen if you were explicitly linked to `https://bsky.app/intent/compose?text=Hello`. Passing the `hasSession` just lets us know to not actually open the modal if we are not yet logged in.

We could put the `hasSession` check right above us calling `onComposeIntent()` though if we wanted to, but no, we are not reusing `onComposeIntent` anywhere but here.

estrattonbailey (approved):
Nice!

PR #2998, opened 2024-02-27
Show tab bar on desktop web

gaearon:
We need to improve the tabbar itself so that it's more usable (e.g. I don't think the way it manages scroll is great right now), but as an incremental improvement, let's replace the "Bluesky" desktop-only header with it. This brings it to parity with tablet and mobile layouts and clarifies what you're looking at (and aligns the information architecture between platforms).

Before:

Screenshot 2024-02-27 at 04 36 41

After:

Screenshot 2024-02-27 at 04 36 58

One awkward point is that the current feed is repeated twice on Home. I think we can live with that.

### DIFF
--- a/src/view/com/feeds/FeedPage.tsx
+++ b/src/view/com/feeds/FeedPage.tsx
@@ -1,30 +1,24 @@
 import React from 'react'
-import {
-  FontAwesomeIcon,
-  FontAwesomeIconStyle,
-} from '@fortawesome/react-native-fontawesome'
 import {useNavigation} from '@react-navigation/native'
 import {useAnalytics} from 'lib/analytics/analytics'
 import {useQueryClient} from '@tanstack/react-query'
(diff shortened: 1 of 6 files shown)


gaearon:
Updated:

Screenshot 2024-02-27 at 17 07 45

gaearon:
New scroll behaviors.

On native, we adjust tabs on each tap (similar to before) but with a bit of the previous tab visible.

https://github.com/bluesky-social/social-app/assets/810438/79639a24-07fd-4cf0-9034-f9993dfdf20d

On web, we don't adjust things on tap unless the selected tab is out of viewport. In which case we get it into viewport, and additionally scroll by almost an entire additional page. So it feels closer to traditional pagination.

We can add "left" / "right" arrows similar to Twitter and TikTok in a follow-up.

https://github.com/bluesky-social/social-app/assets/810438/091cf77d-a96a-4f05-940b-9619ee671204

pfrazee (approved):
Yeah this feels great. Well done

PR #3004, opened 2024-02-27
Theme tweaks

estrattonbailey:


haileyok (approved):
Awesome, some of those darker blues that were on dim look a lot better now. Added one additional tweak we were missing to the user lists (wrong color was used for the "Users" label)

PR #3015, opened 2024-02-28
Fix wrong feed being shown

gaearon:
This might solve the problem where the app would sometimes show a wrong feed header for the selected feed.

## The Problem

The problem (or at least a part of it) is that `usePinnedFeedsInfos` was not deterministic across components. Instead of always using the same query cache, it used state and an effect that fired off requests and saved the successful ones. The set of successful requests could be different depending in the `Home` component and in the `HomeHeader` components, but we assumed in the code that the indexes "match up" between components.  So we could show the wrong feed.

This was easy to reproduce by adding a non-deterministic `throw` when fetching individual items.

## The Solution

First, we're replacing ad-hoc state + effect combo with `useQuery`. I set the stale cache as infinite because (I think) we don't want to revalidate the list of feeds while they're being displayed. They do seem to be invalidated correctly when I actually change the pinned/unpinned status.

Next, I've taken the opportunity to batch the requests to feeds. (Lists still have to be resolved one by one.) This cuts down on requests at startup.

Before:

Screenshot 2024-02-28 at 15 09 27

After:

Screenshot 2024-02-28 at 15 09 18

I removed the special logic that showed "Following" while the rest was loading. I don't think it's actually beneficial to the UX since we fill in the list of feeds right after. In fact, in the UI we _already_ wait for them to load (except in the right pane) so we might as well embrace that by delaying the entire query until the list is fetched.

Finally, I've changed `Home` to pass props to `HomeHeader` because we already have the query result there anyway. So even if they share the cache, the data flow is clearer this way. This also lets us avoid `isLoading` checks in `HomeHeader` where we don't have anything meaningful to show as a fallback.

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -1,11 +1,9 @@
-import React from 'react'
 import {
   useQuery,
   useInfiniteQuery,
   InfiniteData,
   QueryKey,
   useMutation,
-  useQueryClient,
 } from '@tanstack/react-query'
 import {
   AtUri,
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Tested, seems solid

PR #3018, opened 2024-02-28
Improve muted word matching

estrattonbailey:
Updates API for the match function, and adds exceptions for languages either without spaces, or with space usage that's different from latin and other languages that use spaces between each word.


pfrazee (approved):
Yeah LGTM

PR #3019, opened 2024-02-28
Protect against non functions being passed to close callback

estrattonbailey:


src/components/Dialog/index.tsx
```diff
@@ -56,7 +57,7 @@ export function Outer({
   )
 
   const close = React.useCallback<DialogControlProps['close']>(cb => {
-    if (cb) {
+    if (cb && typeof cb === 'function') {
```
estrattonbailey, line 60: Only allow functions to be passed (not an event object as was the case here)

src/components/Dialog/index.tsx
```diff
@@ -74,8 +75,16 @@ export function Outer({
...
-        closeCallback.current = undefined
+        try {
+          closeCallback.current?.()
+        } catch (e: any) {
+          logger.error(`Dialog closeCallback failed`, {
+            message: e.message,
+          })
+        } finally {
+          closeCallback.current = undefined
+        }
```
estrattonbailey, line 86: Make sure even if something in the callback throws, that we're able to reset

src/components/Dialog/types.ts
```diff
@@ -6,8 +6,13 @@ import {ViewStyleProp} from '#/alf'
 
 type A11yProps = Required<AccessibilityProps>
 
+export type DialogControlProps = {
+  open: (options?: DialogControlOpenOptions) => void
+  close: (callback?: () => void) => void
+}
+
 export type DialogContextProps = {
-  close: () => void
```
estrattonbailey, line 10: Missed this type update last week, didn't get type errors from the usages below 🤦

pfrazee (approved):

PR #3024, opened 2024-02-28
always use utf8 offsets

haileyok:
Because of varying emoji lengths, relying on Swift's `String.Index` instead of utf-16 offsets is a safer option when determining a text range. Right now, some emojis may cause links to not be pressable, since we are scanning the incorrect offset.


estrattonbailey (approved):
Tested using the problem posts we've found in the last couple days, all works great 👍

PR #3025, opened 2024-02-28
Only trigger the middle aux click "new tab" action if we start and stop within a post.

haileyok:
## Why

Right now, in some browsers that support it such as Firefox, middle clicking outside of a post allows you to "smooth scroll" (or whatever this is called) the screen by just moving the mouse up and down. This action gets triggered if we are outside of a post.

If we are within a post, we cancel the `onMouseDown` event so that the "smooth scroll" do not enable. We then simulate an aux/meta click `onMouseUp`.

The problem with this logic is that we only care about receiving the `onMouseUp` event to determine if we should open the new tab. We need to change the logic a bit.

## How

The order of events for default behavior is as follows:

1. `onMouseDown` is fired. The browser will enable/disable the "smooth scroll" behavior unless the pointer is over an anchor tag.
2. `onMouseUp` is fired. There is no special behavior that deals with "smooth scroll" here. The `onMouseUp` event is fired regardless of the previous state of the "smooth scroll".

What our logic currently does is as follows:

1. `onMouseDown` is fired. If we are hovering a post, we cancel the default behavior so that "smooth scroll" does not become enabled. If "smooth scroll" is already enabled, we don't intercept the `onMouseDown` event at all, since the browser handles it instead.
2. `onMouseUp` is fired. Regardless of the previous state of `onMouseDown`, we simulate the meta click.

What's missing in step two is a way to tell if we were smooth scrolling prior to receiving this `onMouseUp` event.

This PR adds a flag to tell if we originally received an `onMouseDown` event within the post, and _only if so_ do we simulate the click when we receive an `onMouseUp` event.

https://github.com/bluesky-social/social-app/assets/153161762/17975ee6-8eeb-450e-9ba6-66594451b0b5

## Testing

1. See the current production build, where if you press down middle click outside of a post then move your mouse into a post and release, the post will open in a new tab.
2. Also, if you enable "smooth scroll" outside of a post, then middle click again inside of a post, the new tab will also be opened.

3. In this PR, see that pressing down middle click outside of a post then moving into the post and releasing does not cause a new tab to be opened.
4. Similarily, see that if you enable "smooth scroll" from outside the post, then middle click inside the post while it is still enabled, the new tab will not be opened.


haileyok:
We'll come back to this, there are probably some better ways to remove the hack completely.

PR #3027, opened 2024-02-28
Remove `Platform.select()`

haileyok:
This causes problems during compile on web platform, likely due to minification. We can revisit to see if there's another way, but for now let's stick to what we know works.


gaearon (approved):

estrattonbailey (approved):
Verified locally

PR #3037, opened 2024-02-29
Don't mute your own posts

haileyok:
If you make a post with a mute word in it, it should still show up in your timeline and on your profile.


estrattonbailey (approved):
Nice

PR #3041, opened 2024-02-29
Enable tags inside of quotes

haileyok:
Highlighting tags, links, etc. in quotes but keeping them unpressable. We should discuss if we actually want to do this though.

Screenshot 2024-02-28 at 7 01 37 PM

Screenshot 2024-02-28 at 7 05 29 PM

### DIFF
--- a/src/view/com/util/post-embeds/QuoteEmbed.tsx
+++ b/src/view/com/util/post-embeds/QuoteEmbed.tsx
@@ -93,7 +93,9 @@ export function QuoteEmbed({
       quote.text.trim()
         ? new RichTextAPI({
             text: quote.text,
-            facets: quote.facets?.filter(f => !!f.features[0].tag),
+            facets: quote.facets?.filter(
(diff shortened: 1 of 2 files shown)


src/view/com/util/post-embeds/QuoteEmbed.tsx
```diff
@@ -127,14 +130,15 @@ export function QuoteEmbed({
...
-        />
+        <View pointerEvents="none">
+          <RichText
+            enableTags
+            value={richText}
+            style={[a.text_md]}
+            numberOfLines={20}
+            authorHandle={quote.author.handle}
+          />
+        </View>
```
estrattonbailey, line 141: Looks like this doesn't work on web, may be because Radix is using event delegation 🤔

haileyok: Changed the fix to work on both platforms 👍 Think you were onto something.

pfrazee (approved):
I'm +1 to this

PR #3045, opened 2024-02-29
Simplify intent path logic, ensure all variations work for bluesky://, bluesky:///, and https://bsky.app/

haileyok:
## Why

When checking the intent path, we use `new URL(url)` to parse the url. Generally, it is awkward for someone to use `bluesky:///intent/compose` since we commonly only use two slashes in a URL scheme. However, if we use `bluesky://intent/compose`, this makes the first part of the intent the hostname rather than part of the pathname.

Right now, we work around that with an `isNative` check. However, it is also possible for Safari to deep link into our app through `https://bsky.app/`. In that case, our `isNative` check will fail since we are checking the hostname for `=== intent` rather than the first part of the path name.

## How

The best option is to ensure that `bluesky://` is converted to `bluesky:///` so that we can just always check the first part of the pathname for `=== intent` rather than having to apply `isNative` logic and possibly missing some variations of the deeplink.

## Test

To test, use both `bluesky://intent/compose?text=Test One` and `bluesky:///intent/compose?text=Test Two` from a simulator.

### DIFF
--- a/src/lib/hooks/useIntentHandler.ts
+++ b/src/lib/hooks/useIntentHandler.ts
@@ -19,7 +19,7 @@ export function useIntentHandler() {
       // slashes, like bluesky:///intent/follow. However, supporting just two slashes causes us to have to take care
       // of two cases when parsing the url. If we ensure there is a third slash, we can always ensure the first
       // path parameter is in pathname rather than in hostname.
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
lgtm

PR #3047, opened 2024-02-29
Dedicated screen for hashtags, POC ALF list

haileyok:
Got slightly carried away with ALF here, as you know, ALF is very persuasive.

We can hold off on the ALF stuff if we want, it's pretty small though and we should probably start normalizing our lists, list footers, etc anyay. This is an example of how we can do that.

## Why

The current behavior of switching to the Search tab is a bit awkward. You cannot easily use native gestures on iOS to swipe back, and there are no animations that show the intended action of "pushing". We should prefer to move forward in the stack rather than switch tabs.

Adding a few components that can be reused in lists elsewhere that all use ALF will make the continual migration fairly easy. These are mainly just spinner components, one for the footer and one for the empty, loading state. This will also let us maintain styles between each screen. Right now, we have varying views for different screens. These should all be the same across every screen.

Another small fix removes some ghost padding from the `ViewHeader` that caused the title to not be aligned correctly in the center of the screen.

## What

- Create a new route specifically for hashtags instead of switching to the search tab
- Update the `RichTextTag` dropdown items to push to this screen
- Update `server.go` with the new route
- Create an ALF list footer for a spinner, errors, etc.
- Create an ALF "empty state" spinner
- Create an ALF desktop header
- Use `https://bsky.app/hashtag/{HASHTAG}` for route. Handle can optionally be applied with `?author={HANDLE}`.
- Adjust `ViewHeader` to ensure all items are centered correctly in all cases

## Testing

- Verify that `http://localhost:19006/hashtag/hashtags` displays hashtags for #hashtags
- Verify that `http://localhost:19006/hashtag/hashtags?author=bsky.app` displays hashtags only for @bsky.app
- Verify that `http://localhost:19006/hashtag/#somehashtag` returns not found. The router strips the # symbol out of the URL, so we can't use that.
- Verify that `http://localhost:19006/hashtag` returns not found.
- Verify that opening a hashtag from the feed, profile, etc. on different platforms works correctly by pushing to the proper screen, and that it applies the author filter when selected.

## List with error

Screenshot 2024-02-29 at 9 01 52 AM

## Web

https://github.com/bluesky-social/social-app/assets/153161762/ec3df535-fd72-4e9c-b5ae-e376ed137c72

## Empty/Error State Screens

Screenshot 2024-02-29 at 2 22 12 PM
Screenshot 2024-02-29 at 2 22 33 PM
Screenshot 2024-02-29 at 2 23 12 PM
Screenshot 2024-02-29 at 2 23 16 PM

### DIFF
--- a/bskyweb/cmd/bskyweb/server.go
+++ b/bskyweb/cmd/bskyweb/server.go
@@ -180,6 +180,7 @@ func serve(cctx *cli.Context) error {
 	e.GET("/", server.WebHome)
 
 	// generic routes
+	e.GET("/hashtag", server.WebGeneric)
 	e.GET("/search", server.WebGeneric)
 	e.GET("/feeds", server.WebGeneric)
 	e.GET("/notifications", server.WebGeneric)
--- a/bskyweb/cmd/bskyweb/server.go
+++ b/bskyweb/cmd/bskyweb/server.go
@@ -180,7 +180,7 @@ func serve(cctx *cli.Context) error {
 	e.GET("/", server.WebHome)
 
 	// generic routes
-	e.GET("/hashtag", server.WebGeneric)
+	e.GET("/hashtag/:tag", server.WebGeneric)
 	e.GET("/search", server.WebGeneric)
 	e.GET("/feeds", server.WebGeneric)
 	e.GET("/notifications", server.WebGeneric)
--- a/src/components/Lists.tsx
+++ b/src/components/Lists.tsx
@@ -1,0 +1,126 @@
+import React from 'react'
+import {atoms as a, useTheme} from '#/alf'
+import {Text, View} from 'react-native'
+import {Loader} from '#/components/Loader'
+import {Trans} from '@lingui/macro'
+import {cleanError} from 'lib/strings/errors'
+import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
+import {Button} from '#/components/Button'
+
+export function ListFooter({
+  isFetching,
+  isError,
+  error,
+  onRetry,
+}: {
+  isFetching: boolean
+  isError: boolean
+  error?: string
+  onRetry?: () => Promise<unknown>
+}) {
+  const t = useTheme()
+
+  return (
+    <View
+      style={[
+        a.w_full,
+        a.align_center,
+        a.justify_center,
+        a.border_t,
+        t.atoms.border_contrast_low,
+        {height: 100},
+      ]}>
+      {isFetching ? (
+        <Loader size="xl" />
+      ) : (
+        <ListFooterMaybeError
+          isError={isError}
+          error={error}
+          onRetry={onRetry}
+        />
+      )}
+    </View>
+  )
+}
+
+function ListFooterMaybeError({
+  isError,
+  error,
+  onRetry,
+}: {
+  isError: boolean
+  error?: string
+  onRetry?: () => Promise<unknown>
+}) {
+  const t = useTheme()
+
+  if (!isError) return null
+
+  return (
+    <View style={[a.w_full, a.px_lg]}>
+      <View
+        style={[
+          a.flex_row,
+          a.gap_md,
+          a.p_md,
+          a.rounded_sm,
+          a.align_center,
+          t.atoms.bg_contrast_25,
+        ]}>
+        <Text
+          style={[a.flex_1, a.text_sm, t.atoms.text_contrast_medium]}
+          numberOfLines={2}>
+          {error ? (
+            cleanError(error)
+          ) : (
+            <Trans>Oops, something went wrong!</Trans>
+          )}
+        </Text>
+        <Button
+          variant="gradient"
+          label="Press to retry"
+          style={[
+            a.align_center,
+            a.justify_center,
+            a.rounded_sm,
+            a.overflow_hidden,
+            a.px_md,
+            a.py_sm,
+          ]}
+          onPress={onRetry}>
+          Retry
+        </Button>
+      </View>
+    </View>
+  )
+}
+export function ListMaybeLoading({isLoading}: {isLoading: boolean}) {
+  if (!isLoading) return
+  return (
+    <View style={[a.w_full, a.align_center, {top: 100}]}>
+      <Loader size="xl" />
+    </View>
+  )
+}
+
+export function ListHeaderDesktop({
+  title,
+  subtitle,
+}: {
+  title: string
+  subtitle?: string
+}) {
+  const {isDesktop} = useWebMediaQueries()
+  const t = useTheme()
+
+  if (!isDesktop) return null
+
+  return (
+    <View style={[a.w_full, a.py_lg, a.px_xl, a.gap_xs]}>
+      <Text style={[a.text_3xl, a.font_bold, t.atoms.text]}>{title}</Text>
+      {subtitle ? (
+        <Text style={[a.text_md, t.atoms.text]}>{subtitle}</Text>
+      ) : undefined}
+    </View>
+  )
+}
--- a/src/components/Lists.tsx
+++ b/src/components/Lists.tsx
@@ -1,10 +1,9 @@
 import React from 'react'
-import {atoms as a, useTheme} from '#/alf'
+import {atoms as a, useBreakpoints, useTheme} from '#/alf'
 import {View} from 'react-native'
 import {Loader} from '#/components/Loader'
 import {Trans} from '@lingui/macro'
 import {cleanError} from 'lib/strings/errors'
-import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {Button} from '#/components/Button'
 import {Text} from '#/components/Typography'
 import {StackActions} from '@react-navigation/native'
@@ -106,10 +105,10 @@ export function ListHeaderDesktop({
   title: string
   subtitle?: string
 }) {
-  const {isDesktop} = useWebMediaQueries()
+  const {gtTablet} = useBreakpoints()
   const t = useTheme()
 
-  if (!isDesktop) return null
+  if (!gtTablet) return null
 
   return (
     <View style={[a.w_full, a.py_lg, a.px_xl, a.gap_xs]}>
--- a/src/components/Lists.tsx
+++ b/src/components/Lists.tsx
@@ -123,15 +123,6 @@ export function ListHeaderDesktop({
   )
 }
 
-function ListMaybeLoading({isLoading}: {isLoading: boolean}) {
-  if (!isLoading) return
-  return (
-    <View style={[a.w_full, a.align_center, {top: 100}]}>
-      <Loader size="xl" />
-    </View>
-  )
-}
-
 export function ListMaybePlaceholder({
   isLoading,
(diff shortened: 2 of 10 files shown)


mozzius:
Screenshot 2024-02-29 at 16 57 34

Should the back chevron be centered on the large header? I'd be inclined to fix it to the top, so that its position stay the same across header sizes

haileyok:
@mozzius thanks. also fixed that pesky issue with the title not being perfectly centered too...

bskyweb/cmd/bskyweb/server.go
```diff
@@ -180,6 +180,7 @@ func serve(cctx *cli.Context) error {
 	e.GET("/", server.WebHome)
 
 	// generic routes
+	e.GET("/hashtag/:tag", server.WebGeneric)
```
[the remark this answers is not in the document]
estrattonbailey: @Jacob2161 how's this look?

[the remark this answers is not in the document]
haileyok: For the record, the react-navigation router will treat `https://bsky.app/hashtag` as not found, which is why I explicitly add `:tag` to the path. I think that works, I can build the server locally and test, but also want to make sure it's the right logic in general.

src/screens/Hashtag.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <View style={a.flex_1}>
+      <ViewHeader
+        title={headerTitle}
+        subtitle={author ? `By ${author}` : undefined}
+        canGoBack={true}
+      />
+      <ListMaybeLoading isLoading={isLoading} />
+      {!isLoading && (
+        <List<PostView>
```
estrattonbailey, line 90: Looks like we could use an empty/no results state

src/components/Lists.tsx
```diff
@@ -0,0 +1,126 @@
...
+  title,
+  subtitle,
+}: {
+  title: string
+  subtitle?: string
+}) {
+  const {isDesktop} = useWebMediaQueries()
+  const t = useTheme()
+
+  if (!isDesktop) return null
```
haileyok, line 116: By the way @estrattonbailey not sure if we still use this with ALF stuff, do we have a replacement for `useWebMediaQueries()`, maybe in the breakpoints?

src/view/com/util/ViewHeader.tsx
```diff
@@ -71,42 +73,53 @@ export function ViewHeader({
...
+            <View style={styles.titleContainer} pointerEvents="none">
+              <Text type="title" style={[pal.text, styles.title]}>
+                {title}
+              </Text>
+            </View>
+            {renderButton ? (
+              renderButton()
+            ) : showBackButton ? (
+              <View style={canGoBack ? styles.backBtn : styles.backBtnWide} />
+            ) : null}
```
estrattonbailey, line 113: Not quite centered vertically with and without subtitle, could use a lil finesse. Unsure how this affects other screens, will look deeper.

Edit: on mobile web only, looks good on mobile

src/screens/Hashtag.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }, [refetch])
+
+  const onEndReached = React.useCallback(() => {
+    if (isFetching || !hasNextPage || error) return
+    fetchNextPage()
+  }, [isFetching, hasNextPage, error, fetchNextPage])
+
+  return (
+    <View style={a.flex_1}>
+      <ViewHeader
```
estrattonbailey, line 83: Idea: add a share icon to the top right for easier sharing on mobile

estrattonbailey (commented):
Looking sweeeet

src/screens/Hashtag.tsx
```diff
@@ -79,7 +82,7 @@ export default function HashtagScreen({
     <View style={a.flex_1}>
       <ViewHeader
         title={headerTitle}
-        subtitle={author ? `By ${author}` : undefined}
+        subtitle={author ? `${_(msg`By`)} ${author}` : undefined}
```
estrattonbailey, line 85: Wonder if the interpolated flavor would be easier for translators to grok instead of a lone word 🤔

haileyok: Wait what's the syntax on that?

src/components/TagMenu/index.web.tsx
```diff
@@ -66,11 +66,9 @@ export function TagMenu({
...
           label: _(msg`See ${truncatedTag} posts by user`),
           onPress() {
-            navigation.navigate({
-              name: 'Search',
-              params: {
-                q: tag + (authorHandle ? ` from:${authorHandle}` : ''),
-              },
+            navigation.push('Hashtag', {
+              tag: tag.replace('#', ''),
+              author: authorHandle,
```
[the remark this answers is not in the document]
haileyok: If we add the '#' into the URL, it gets removed by the router, so the route just ends up being `/hashtag` on web...I wanted to do this too but alas :/

src/screens/Hashtag.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const query = React.useMemo(() => {
+    const queryTag = !tag.startsWith('#') ? `#${tag}` : tag
+
+    if (!author) return queryTag
+    return `${queryTag} from:${author}`
+  }, [tag, author])
+
+  const headerTitle = React.useMemo(() => {
+    return `#${enforceLen(tag.toLowerCase(), 24)}`
```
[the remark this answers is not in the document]
haileyok: Hmm maybe we should just use the same logic we use on web dropdowns then...

bskyweb/cmd/bskyweb/server.go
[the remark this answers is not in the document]
Jacob2161: @estrattonbailey :+1:

src/components/Lists.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <View style={[a.w_full, a.py_lg, a.px_xl, a.gap_xs]}>
+      <Text style={[a.text_3xl, a.font_bold, t.atoms.text]}>{title}</Text>
+      {subtitle ? (
+        <Text style={[a.text_md, t.atoms.text]}>{subtitle}</Text>
+      ) : undefined}
```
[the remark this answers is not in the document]
haileyok: I swore I tried that and it was white...probably was the wrong Text.

Good call on contrast.

src/components/TagMenu/index.web.tsx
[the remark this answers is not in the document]
mozzius: Should you be using encodeURIComponent here? What if it's the ##hashtag case?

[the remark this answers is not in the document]
haileyok: @estrattonbailey What's your take on that now that we've changed the way they get sanitized?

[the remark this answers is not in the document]
estrattonbailey: Yep `##hashtag` is still valid, so gotta encode. You're right though, the URLs would look kinda jank for "normal" tags with just one hash 🤔 

I'm actually wrong, what gets passed to `TagMenu` is the _text_ portion of the tag. We should really pass in the `tag` itself, since that's already sanitized. Then we know there's no leading # (unless it's a double hash like `##hashtag`), and yeah we need to encode for those cases. That make sense?

src/components/Lists.tsx
estrattonbailey: Yep! `useBreakpoints from '#/alf'`

src/components/TagMenu/index.web.tsx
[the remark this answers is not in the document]
haileyok: Okay this logic should work now, leaving a comment below.

src/components/TagMenu/index.web.tsx
```diff
@@ -49,8 +49,8 @@ export function TagMenu({
       {
         label: _(msg`See ${truncatedTag} posts`),
         onPress() {
-          navigation.navigate('Search', {
-            q: tag,
+          navigation.push('Hashtag', {
+            tag: tag.slice(1).replaceAll('#', '%23'),
```
haileyok, line 53: Taking the tag and just removing the first character. When we press this button, we know it's going to have at least one # which we'll remove.

The remaining '#' should all be encoded.

src/screens/Hashtag.tsx
```diff
@@ -0,0 +1,131 @@
...
+  route,
+}: NativeStackScreenProps<CommonNavigatorParams, 'Hashtag'>) {
+  const {tag, author} = route.params
+  const setMinimalShellMode = useSetMinimalShellMode()
+  const {_} = useLingui()
+  const [isPTR, setIsPTR] = React.useState(false)
+
+  const fullTag = React.useMemo(() => {
+    return `#${tag.replaceAll('%23', '#')}`
+  }, [tag])
```
haileyok, line 42: Here we prepend a # and replace any '%23' with a '#'. URLs should usually come in with no leading # so this will allow for `/hashtag/somehashtag`. If there are nested ones, they'll need `/hashtag/%23sometag` for `##sometag`

src/components/Lists.tsx
```diff
@@ -0,0 +1,228 @@
...
+  title,
+  subtitle,
+}: {
+  title: string
+  subtitle?: string
+}) {
+  const {gtTablet} = useBreakpoints()
+  const t = useTheme()
+
+  if (!gtTablet) return null
```
haileyok, line 111: This looks like the equivalent of `isDesktop`, is that right?

src/screens/Hashtag.tsx
mozzius: lgtm 👍

src/screens/Hashtag.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const onEndReached = React.useCallback(() => {
+    if (isFetching || !hasNextPage || error) return
+    fetchNextPage()
+  }, [isFetching, hasNextPage, error, fetchNextPage])
+
+  return (
+    <CenteredView style={a.flex_1}>
+      <ViewHeader
+        title={headerTitle}
+        subtitle={author ? _(msg`From @${sanitizeHandle(author)}`) : undefined}
```
haileyok, line 90: @mozzius This here

src/components/Lists.tsx
estrattonbailey: Yeah the breakpoint should line up, is it not?

haileyok: Yea yea it is. Was just making sure. 👍

src/screens/Hashtag.tsx
mozzius: Yeah perfect, although be sure to test it carefully - have ran into issues with inline function calls in the lingui macros before, see here:

https://github.com/mozzius/graysky/issues/470

haileyok: What a pain. Thank you!

haileyok:
Added share button

Screenshot 2024-02-29 at 4 37 42 PM

estrattonbailey (approved):
Incredible stuff, can't wait to use this 🚀

PR #3049, opened 2024-02-29
Remove trailing comma from md files

lens0021:


pfrazee:
Appreciate it but this isnt really needed

PR #3055, opened 2024-02-29
Bump API SDK, add validation to MutedWords

estrattonbailey:
Bumps SDK to fix ## parsing and removal bugs, as well as sanitization like zero width space and newlines. Adds a little validation to the frontend.
image

### DIFF
--- a/src/components/TagMenu/index.tsx
+++ b/src/components/TagMenu/index.tsx
@@ -215,14 +215,12 @@ export function TagMenu({
                           if (isMuted) {
                             resetUpsert()
                             removeMutedWord({
-                              value: sanitizedTag,
+                              value: tag,
                               targets: ['tag'],
                             })
                           } else {
                             resetRemove()
-                            upsertMutedWord([
-                              {value: sanitizedTag, targets: ['tag']},
-                            ])
+                            upsertMutedWord([{value: tag, targets: ['tag']}])
                           }
                         })
                       }}>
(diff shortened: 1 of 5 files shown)


src/components/TagMenu/index.tsx
```diff
@@ -215,14 +215,12 @@ export function TagMenu({
...
-                              value: sanitizedTag,
+                              value: tag,
                               targets: ['tag'],
                             })
                           } else {
                             resetRemove()
-                            upsertMutedWord([
-                              {value: sanitizedTag, targets: ['tag']},
-                            ])
+                            upsertMutedWord([{value: tag, targets: ['tag']}])
```
estrattonbailey, line 223: These should use the sanitized version from the facet value that we're adding in #3047

Same in web file

src/components/dialogs/MutedWords.tsx
```diff
@@ -48,24 +55,29 @@ function MutedWordsInner({}: {control: Dialog.DialogOuterProps['control']}) {
...
+    if (!sanitizedValue || !targets.length) {
+      setField('')
+      setError(_(msg`Please enter a valid word, tag, or phrase to mute`))
+      return
+    }
 
     try {
-      await addMutedWord([{value, targets}])
+      // send raw value and rely on SDK as sanitization source of truth
+      await addMutedWord([{value: field, targets}])
```
estrattonbailey, line 74: It's sanitized above in order to do validation, but since it's done in the SDK too, we don't need to do it twice here.

src/components/dialogs/MutedWords.tsx
```diff
@@ -87,7 +99,12 @@ function MutedWordsInner({}: {control: Dialog.DialogOuterProps['control']}) {
           label={_(msg`Enter a word or tag`)}
           placeholder={_(msg`Enter a word or tag`)}
           value={field}
-          onChangeText={setField}
+          onChangeText={value => {
+            if (error) {
+              setError('')
+            }
```
estrattonbailey, line 105: Clears error upon typing a new value

PR #3063, opened 2024-03-01
Integrate new tag detection and regexes

estrattonbailey:
Integrates new detection regex


src/view/com/composer/text-input/web/LinkDecorator.ts
```diff
@@ -78,8 +80,7 @@ function linkDecorator() {
 
 function iterateUris(str: string, cb: (from: number, to: number) => void) {
   let match
-  const re =
```
estrattonbailey, line 81: Confirmed before doing this that this is the exact same regex, just wasn't shared before.

haileyok (approved):
👍

PR #3068, opened 2024-03-01
Fix icons on blurred posts

estrattonbailey:
Noticed this could be improved


haileyok (approved):
Yup. Just the consideration of if we want to show the muted word in the highlighted post for a thread. I'm impartial, don't see the need since it's explained why it's muted anyway.

PR #3069, opened 2024-03-01
Filter muted words from all feeds

estrattonbailey:
Since mute words are so new, the case where users desire muted words to filter out from _all_ feeds is not currently supported by our moderation system. We're in the process of rewriting the mod system for greater flexibility, so we need to discuss this change and determine if we want to hack it like this, or wait until a proper handling is in place.


estrattonbailey:
Just want to note that since the logic in #3071 sets the UI behaviors on `content` that this logic will apply to those as well

PR #3071, opened 2024-03-01
Fix muting of embedded data

estrattonbailey:
Uses a pretty heavy hand to mute embeds — basically hides the whole post — but fixes cases where mute words were not applied.


pfrazee (approved):
Yeah sucks but this is the right move for now

haileyok (approved):
👍

PR #3077, opened 2024-03-02
Sticky desktop header

gaearon:
In https://github.com/bluesky-social/social-app/pull/2998, I've added a tabbar to dekstop web. I believe this was necessary because having multiple feeds is a core feature of the app and you need to be aware of which feed you're on. Third-party feeds should feel just as "first-class" as our own feeds, so we shouldn't give preferential treatment to Following.

However, the tabbar was getting pretty noisy due to being tall and due to custom triggers. We could tune those but I think it's actually better to always keep the current header visible. Whereas the very top part can hide on scroll. In other words, we can apply the same UI as we do in Profile — and this also lets us just "use the platform" and do `position: sticky`.

Which this PR does.

## Before

You know, whatever's in prod now.

## After

https://github.com/bluesky-social/social-app/assets/810438/1ce805cc-b224-4f54-a206-84c53b4d438d

### DIFF
--- a/src/view/com/home/HomeHeader.tsx
+++ b/src/view/com/home/HomeHeader.tsx
@@ -52,7 +52,7 @@ export function HomeHeader(
   )
 
   return (
-    <HomeHeaderLayout>
+    <HomeHeaderLayout tabBarAnchor={props.tabBarAnchor}>
       <TabBar
         key={items.join(',')}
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Yeah seems promising, let's give it a shot

PR #3090, opened 2024-03-03
Upgrade `UITextView` to latest

haileyok:
Took some time over the weekend to build this into a solid library rather than just an app-specific wrapper. There were a variety of styles that were missing/not supported that have been added, as well as support for a few additional props.

An additional change is that you can now use `<UITextView>` on any platform without needing to do the `Platform.OS` check (that's handled by the library now instead of inside our `RichText` code). Usage is rather simple, and there's no changes to our workflow since we use `RichText` everywhere and just add the `selectable` prop when needed.

An additional bug that's currently present is fixed as well, which caused links to be "missing" a character in pressability. The string `Press Me` would only be pressable on the letters `r` and after, not the `P`. Spent about an hour reading over the Swift string API and figured out that there was a much more elegant solution to the problem than what I was doing before.

```swift
  func getTouchChild(_ location: CGPoint) -> RNUITextViewChild? {
    let fullText = self.textView.attributedText.string
    
    // Get utf16 index of the pressed character
    let charIndex = self.textView.layoutManager.characterIndex(
      for: location,
      in: textView.textContainer,
      fractionOfDistanceBetweenInsertionPoints: nil
    )
    
    var currIndex = -1
    for child in self.reactSubviews() {
      guard let child = child as? RNUITextViewChild,
            let childText = child.text
      else {
        continue
      }
      
      // We need to account for grapheme length here, so use utf16.count
      currIndex += childText.utf16.count
      
      if charIndex <= currIndex {
        print(charIndex)
        print(currIndex)
        return child
      }
    }
    
    return nil
  }
```

I've gone through now and tested a variety of situations in the app, and it looks like the only missing piece should be a "highlight" when we press in on a link. This is a bit annoying, because the native `UITextView` does not handle `touchesBegan` events. I'll have to experiment with some possible solutions on that front.


pfrazee (approved):
Tested on simulators for iOS and Android and on Web. Looks great!

PR #3092, opened 2024-03-04
expo-updates implementation

haileyok:
# Expo Updates

This PR implements the `expo-updates` client into the app. It replies on the Bluesky updates server which uses the expo-updates protocol.

I've opted to stray from the default options here and never check automatically at app launch. I'm not certain of the start-time implications across devices and networks if we take that route. I believe that setting the `fallbackToCacheTimeout` to zero would prevent that, but I would rather remain safe and manually check for the update once the app has launched.

Once a user opens the app, we set a timeout for 15 seconds that will check for an update. If there is an update available, it will be fetched in the background but will not be used for that launch. We also check for an update each time the user opens the app from the background, assuming that 30 minutes has passed between the last time they opened the app and now.

New updates will be used whenever one of two things happens:

1. The next time the user launches the app (i.e. after a force quit or after watchdog terminates the app)
2. After the app has been in the background for 30 minutes or more

In the first situation, there isn't much thinking that we need to do. In the latter, we use a 30 minute time period to ensure that this wasn't the user backgrounding the app to go watch a short video or something, then come back to the app expecting to be in the same state. A 30 minute period is generally enough time for to feel confident in reloading the app without affecting the user's flow. This only happens _if_ an update is available.

## Deploy Process

The deploy process is pretty straight forward.

- We merge in changes to main
- We run the action, inputting the version that the update is for. By inputting a version, we allow ourselves two options:
  - The version that is currently live on the app store
  - The version that is being used internally for TestFlight

By having this distinction, we can deploy changes rapidly for testing internally while maintaining the version that production users are using.

## Test Plan

Created a TestFlight release and an APK that has an `Alert` at launch. This update was deployed and installed on devices. After restarting the app, the message no longer appears, indicating that the new bundle has been properly installed.

We should use this internally on a TestFlight release for a bit before using it in production.


haileyok:
Closing in favor of https://github.com/bluesky-social/social-app/pull/3291

PR #3094, opened 2024-03-04
Improve dialogs a11y

estrattonbailey:
I recommend [reviewing without whitespace.](https://github.com/bluesky-social/social-app/pull/3094/files?diff=split&w=1)

With this PR, with VoiceOver/TalkBack enabled:
- dialogs should be "focused" when opened
- focus should be trapped within the dialog

Easiest to test on a real device: `npx expo run:ios --device` with your phone plugged in.

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -147,9 +147,11 @@ export function Outer({
     isOpen && (
       <Portal>
         <View
+          // iOS
           accessibilityViewIsModal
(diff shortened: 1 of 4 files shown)


haileyok:
Okay, I went through and tested on both platforms on-device, and looks like the dialogs work well now. Easy to close out of them as well where it wasn't before.

Next step is to fix that context button for android, can't get to it at all.

haileyok (approved):

PR #3096, opened 2024-03-04
Fabric migration prereqs - Library Migrations Base PR | Fix issue with uploading images on Android

haileyok:
While investing text accessibility issues (including selection of hashtags), I stumbled into some PRs in the React-Native repo that explain how this issue has been fixed in the Fabric renderer. The gist of the issue is that we need to be able to add the `focusable` prop to nested text elements, however, right now Paper does not support this. It's unlikely that we'll see that change get made any time soon either, as the focus has been moved to Fabric.

A Fabric migration will also be useful for general performance improvements. The Expo team has been willing to provide some help in this endeavor, so I think this might be worthwhile to start working towards for multiple reasons.

I'm incrementally breaking down https://github.com/bluesky-social/social-app/pull/2967 so that we can easily tell where issues come from during the migration.

This is the base PR for all of the changes required. This PR only consists of the below (mainly complete removals and upgrades). Individual upgrades are in these PRs:

## Other PRs

- [x] https://github.com/bluesky-social/social-app/pull/3099
- [x] https://github.com/bluesky-social/social-app/pull/3100
- [x] https://github.com/bluesky-social/social-app/pull/3101
- [x] https://github.com/bluesky-social/social-app/pull/3102
- [x] https://github.com/bluesky-social/social-app/pull/3104

### Remove unused libraries

These libraries are not used anywhere in the app and are being removed. Checked both our code as well as other libraries inside of `node_modules`.

- [x] `@react-native-camera-roll/camera-roll`
- [x] `react-circular-progress-bar`
- [x] `react-native-appstate-hook`
- [x] `react-native-web-linear-gradient`
- [x] `use-deep-compare`
- [x] `lru_map`

### Remove `react-native-community/blur`

Just removing this library without a replacement. There was a single instance of us using it, and it was actually not necessary at all.

### Replace `react-native-linear-gradient` with `expo-linear-gradient`

This is the only "drop-in" replacement we can make, so I'm making this change here. The other expo swaps will require some minor code changes.

### Upgrade Expo deps to latest versions supported by SDK 50

This will help whenever we start adding some of the other Expo replacements. From my past experience, there were some issues with compatibility between older version of expo packages and the ones we were wanting to add.

# Testing

This PR contains *all* of the new packages required to test each individual PR above.

- Checkout this branch
- `yarn install` and `pod install`
- `yarn ios` or `yarn android`
- Checkout each individual branch separately to test that specific change


src/components/GradientFill.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import LinearGradient from 'react-native-linear-gradient'
+import {LinearGradient} from 'expo-linear-gradient'
```
haileyok, line 2: I don't think we are actually using this `GradientFill` component at all, can we RM or is this going to be used somewhere else?

estrattonbailey: Almost used in mod services, no need to keep it around for now if ya don't want to

mozzius:
Oh, and the segment-related packages need upgrading, but I couldn't find any info about breaking changes so I am loathe to touch them without knowing how to test them properly

haileyok:
Let's make sure we add all of those to Trello, thanks for compiling a list for us!

haileyok:
Closing this in favor of the individual PRs.

PR #3097, opened 2024-03-04
Add `Menu` component

estrattonbailey:
This PR adds a new abstraction to replace Zeego and the native context menus. It provides essentially the same interface as our `NativeDropdown` wrapper, but replaces the Radix dropdown on web, and uses a bottom sheet on native.

Usage looks like this:
```tsx
        <Menu.Root control={menuControl}>
          <Menu.Trigger label="Open menu" style={[a.flex_1]}>
            {({state, props}) => {
              return (
                <Text {...props}>Open menu</Text>
              )
            }}
          </Menu.Trigger>

          <Menu.Outer>
            <Menu.Group>
              <Menu.Item label="Click me" onPress={() => {}}>
                <Menu.ItemIcon icon={Search} />
                <Menu.ItemText>Click me</Menu.ItemText>
              </Menu.Item>

              <Menu.Divider />

              <Menu.Item
                label="Another item"
                onPress={() => menuControl.close()}>
                <Menu.ItemText>Another item</Menu.ItemText>
              </Menu.Item>
            </Menu.Group>
          </Menu.Outer>
        </Menu.Root>

```

CleanShot 2024-03-05 at 13 44 53@2x

CleanShot 2024-03-05 at 14 17 55@2x

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -213,6 +213,7 @@
     }
 
     /* NativeDropdown component */
+    .radix-dropdown-item:focus,
     .nativeDropdown-item:focus {
       outline: none;
     }
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "@lingui/react": "^4.5.0",
     "@mattermost/react-native-paste-input": "^0.6.4",
     "@miblanchard/react-native-slider": "^2.3.1",
+    "@radix-ui/react-dropdown-menu": "^2.0.6",
     "@react-native-async-storage/async-storage": "1.21.0",
     "@react-native-camera-roll/camera-roll": "^5.2.2",
     "@react-native-clipboard/clipboard": "^1.10.0",
--- a/package.json
+++ b/package.json
@@ -149,6 +149,7 @@
     "react-avatar-editor": "^13.0.0",
     "react-circular-progressbar": "^2.1.0",
     "react-dom": "^18.2.0",
+    "react-keyed-flatten-children": "^3.0.0",
     "react-native": "0.73.2",
     "react-native-appstate-hook": "^1.0.6",
     "react-native-drawer-layout": "^4.0.0-alpha.3",
--- a/src/components/Dialog/context.ts
+++ b/src/components/Dialog/context.ts
@@ -21,7 +21,8 @@ export function useDialogControl(): DialogOuterProps['control'] {
     open: () => {},
     close: () => {},
   })
-  const {activeDialogs} = useDialogStateContext()
+  const {activeDialogs, openDialogs} = useDialogStateContext()
+  const isOpen = openDialogs.includes(id)
 
   React.useEffect(() => {
     activeDialogs.current.set(id, control)
@@ -31,14 +32,18 @@ export function useDialogControl(): DialogOuterProps['control'] {
     }
   }, [id, activeDialogs])
 
-  return {
-    id,
-    ref: control,
-    open: () => {
-      control.current.open()
-    },
-    close: cb => {
-      control.current.close(cb)
-    },
-  }
+  return React.useMemo<DialogOuterProps['control']>(
+    () => ({
+      id,
+      ref: control,
+      isOpen,
+      open: () => {
+        control.current.open()
+      },
+      close: cb => {
+        control.current.close(cb)
+      },
+    }),
+    [id, control, isOpen],
+  )
 }
--- a/src/components/Dialog/types.ts
+++ b/src/components/Dialog/types.ts
@@ -22,6 +22,7 @@ export type DialogControlRefProps = {
 export type DialogControlProps = DialogControlRefProps & {
   id: string
   ref: React.RefObject<DialogControlRefProps>
+  isOpen: boolean
 }
 
 export type DialogContextProps = {
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -166,9 +166,7 @@ export function Group({children, style}: GroupProps) {
         style,
       ]}>
       {flattenReactChildren(children).map((child, i) => {
-        // ignore null children, like Dividers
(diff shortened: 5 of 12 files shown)


src/components/Menu/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+  })
+
+  if (!React.isValidElement(child)) {
+    logger.error(
+      'Menu.Trigger children must be a function that returns a valid element',
+    )
+    return null
+  }
+
+  return React.cloneElement(child, {
```
gaearon, line 76: This seems unnecessary to me — I think you can instead pass it to `children({})`?

```js
  const child = children({
    isNative: true,
    control,
    state: {
      hovered: false,
      focused,
      pressed,
    },
    props: {
      onPress: control.open,
      onFocus,
      onBlur,
      onPressIn,
      onPressOut,
      accessibilityLabel: label,
    },
  })
```

I'm renaming `handlers` to `props` so that sticking it there makes sense while keeping usage the same (spreading them).

estrattonbailey: Love it, much clearer!

src/components/Menu/index.web.tsx
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    <DropdownMenu.Item asChild onSelect={onPress}>
+      <Pressable
+        className="radix-dropdown-item"
+        accessibilityHint=""
+        accessibilityLabel={label}
+        onPress={onPress}
+        onFocus={onFocus}
+        onBlur={onBlur}
+        style={flatten([
```
haileyok, line 161: Small nit, but we don't have to flatten here. That happens internally anyway. Same for the couple other places where `flatten` is used.

estrattonbailey: Discussed in slack, and added comment: Radix is merging some props here so we do need to flatten ahead of time

src/components/Dialog/context.ts
```diff
@@ -21,7 +21,8 @@ export function useDialogControl(): DialogOuterProps['control'] {
     open: () => {},
     close: () => {},
   })
-  const {activeDialogs} = useDialogStateContext()
+  const {activeDialogs, openDialogs} = useDialogStateContext()
+  const isOpen = openDialogs.includes(id)
```
estrattonbailey, line 25: Since closing is ultimately handled by the `BottomSheet`, and we need to have a top-level reference to open dialogs for a11y prop setting at the `shell` level, I'm reusing the global state here for this localized reference.

src/components/Dialog/types.ts
```diff
@@ -22,6 +22,7 @@ export type DialogControlRefProps = {
 export type DialogControlProps = DialogControlRefProps & {
   id: string
   ref: React.RefObject<DialogControlRefProps>
+  isOpen: boolean
```
estrattonbailey, line 25: Added so that we can control the Radix dropdowns using the same `control` interface

src/components/Menu/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+      {...rest}
+      accessibilityHint=""
+      accessibilityLabel={label}
+      onPress={e => {
+        onPress(e)
+
+        if (!e.defaultPrevented) {
+          control?.close()
+        }
+      }}
```
estrattonbailey, line 110: Same functionality as Radix on web and native: `preventDefault()` and the sheet/dropdown won't close on press of an item.

src/components/Menu/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+        a.rounded_md,
+        a.overflow_hidden,
+        a.border,
+        t.atoms.border_contrast_low,
+        style,
+      ]}>
+      {flattenReactChildren(children).map((child, i) => {
+        // ignore null children, like Dividers
+        // @ts-ignore
+        return React.isValidElement(child) && child.props?.children ? (
```
estrattonbailey, line 171: Allows for conditional logic using fragments. Group still expects `Item` components as immediate children, and filters `Divider`s since they have no `children` of their own. May be a better way...

src/view/screens/Storybook/Menus.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+            <Menu.Group>
+              <Menu.Item label="Click me" onPress={() => {}}>
+                <Menu.ItemIcon icon={Search} />
+                <Menu.ItemText>Click me</Menu.ItemText>
+              </Menu.Item>
+
+              <Menu.Item
+                label="Another item"
+                onPress={() => menuControl.close()}>
```
estrattonbailey, line 63: This control works for both web and native

package.json
```diff
@@ -58,6 +58,7 @@
     "@lingui/react": "^4.5.0",
     "@mattermost/react-native-paste-input": "^0.6.4",
     "@miblanchard/react-native-slider": "^2.3.1",
+    "@radix-ui/react-dropdown-menu": "^2.0.6",
```
estrattonbailey, line 61: Zeego relies on `2.0.1`, which was the last breaking release. Patches since then were all trivial, and I've observed no issues with existing dropdowns on web.

bskyweb/templates/base.html
```diff
@@ -213,6 +213,7 @@
     }
 
     /* NativeDropdown component */
+    .radix-dropdown-item:focus,
```
estrattonbailey, line 216: Just being more specific here, since now we're using Radix directly. I'd like to remove the other class once `NativeDropdown` is phased out.

src/components/Menu/index.tsx
estrattonbailey: Ope, there is, silly me. Updated.

haileyok (approved):
Tested this on Fabric and it works. There are some things that need to get fixed up around dialogs in general, but those are not related to this PR. Works great on paper in the meantime!

PR #3102, opened 2024-03-05
Fabric migration - `rn-fetch-blob`, `react-native-fs` -> `expo-file-system`

haileyok:
### Testing requires building a client after checking out https://github.com/bluesky-social/social-app/pull/3096

This removes migrates our use of `react-native-fs` and `rn-fetch-blob` to `expo-file-system`. It also makes use of `expo-image-manipulator` instead of using `react-native-image-resizer`.

It also fixes a bug where the image manipulator would try to save to the same file path again. We are now cleaning up after ourselves, so this can no longer happen. Of course, in cleaning up we are also preventing the unnecessary growth of the local file store as well.

Tests are also updated to reflect the use of new libraries.

## Testing

- Save image (Tests saving and cache)
	- [x] iOS
	- [x] Android
	- [x] Not applicable on web
- Share image (Tests saving and cache)
	- [x] iOS
	- [x] Android
	- [x] Not applicable on web
- Add a link card (Tests both saving and resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload an image (Tests resizing/compression) 
 	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload an avatar (Tests resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload a banner (Tests resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works


src/lib/media/manip.ts
```diff
@@ -81,47 +72,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
+  // Usually whenever we share an image it will already be available in the cache. If it isn't, then we
+  // will download it.
+  let imageUri = await ExpoImage.getCachePathAsync(uri)
+  if (!imageUri) {
+    // NOTE
+    // assuming PNG
+    // we're currently relying on the fact our CDN only serves pngs
+    // -prf
+    imageUri = await downloadImage(uri, createPath('png'), 5e3)
   }
```
haileyok, line 85: We can take advantage of the expo-image cache here. We will almost always have the image downloaded already by the time we want to share it, so there's no need for us to create duplicates or wait for downloads (especially annoying on slower connections)

src/lib/media/manip.ts
```diff
@@ -81,47 +72,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
-  imagePath = normalizePath(await moveToPermanentPath(imagePath, '.png'), true)
+  let imageUri = await ExpoImage.getCachePathAsync(uri)
+  if (!imageUri) {
+    // download the file to cache
+    // NOTE
+    // assuming PNG
+    // we're currently relying on the fact our CDN only serves pngs
+    // -prf
+    imageUri = await downloadImage(uri, createPath('png'), 5e3)
+  }
```
haileyok, line 106: Same cache use here

src/lib/media/manip.ts
```diff
@@ -203,11 +209,36 @@ function joinPath(a: string, b: string) {
...
-  if (isAndroid || allPlatforms) {
-    if (!str.startsWith('file://')) {
-      return `file://${str}`
-    }
+function normalizePath(str: string): string {
+  if (!str.startsWith('file://')) {
+    return `file://${str}`
   }
   return str
 }
```
haileyok, line 217: There's no need to make a distinction between adding `file://` and not on iOS. Both are supported (Android only supports adding paths with `file://`). We can simplify this logic by just always adding it.

src/lib/media/manip.ts
```diff
@@ -203,11 +209,36 @@ function joinPath(a: string, b: string) {
...
+
+function createPath(ext: string) {
+  // cacheDirectory will never be null on native, so the null check here is not necessary except for typescript.
+  // we use a web-only function for downloadAndResize on web
+  return `${cacheDirectory ?? ''}/${uuid.v4()}.${ext}`
+}
```
haileyok, line 223: This function (and all `expo-file-system` usage) is only used on native, so there are not any situations where `cacheDirectory` would be null.

src/lib/media/manip.ts
```diff
@@ -147,27 +139,48 @@ interface DoResizeOpts {
...
-        path: normalizePath(resizeRes.path),
+        path: normalizePath(resizeRes.uri),
         mime: 'image/jpeg',
-        size: resizeRes.size,
+        size: info.size,
         width: resizeRes.width,
         height: resizeRes.height,
       }
+    } else {
+      await deleteAsync(resizeRes.uri)
```
haileyok, line 183: By cleaning up properly between each resize, we prevent growing local storage and a bug in the other `react-native-image-resizer` package that would prevent resizes from completing properly.

src/lib/api/api-polyfill.ts
```diff
@@ -33,9 +33,26 @@ async function fetchHandler(
...
+          from: reqBody,
+          to: newPath,
+        })
+        reqBody = newPath
+      } catch (e) {
+        await copyAsync({
+          from: reqBody,
+          to: newPath,
+        })
+      }
```
haileyok, line 55: This is a fix for some devices preventing writing to the directory we are using. There are some users who report _never_ having been able to upload a photo. I've been able to finally reproduce this back to this function. The `moveAsync` or previously `RNFS.moveFile` would error out before the upload would succeed.

Here, we attempt to move first (that's the optimal way to do it, since it doesn't duplicate the file and fill up storage) but if we can't, we'll just copy it.

haileyok:
Quick test of link card sizes: The first is made on native (with these changes) and the second is made on web (so no changes). Both line up proportions-wise.

Screenshot 2024-03-06 at 10 04 08 AM

estrattonbailey (approved):
Aside from the warning I slacked you about, this works great on all my devices

Issue #2292, opened 2023-12-24
Dark mode on application startup


maisondasilva:
It would be nice to have an image with a darker background for those who use dark mode activated, because at night when opening the application it displays a bright image!

![Imagem do WhatsApp de 2023-12-24 à(s) 11 21 11_c3856e46](https://github.com/bluesky-social/social-app/assets/11423362/a59fdfd2-d825-4b87-9d4d-949b9325852e)

pfrazee:
we'll have to see if that's possible. the splash is registered with the operating system and we may not be able to control it

maisondasilva:
I use an app from Ubiquiti where opening the app changes according to light mode and dark mode, I just don't know how they did it!

ansh:
We can do this via the `ios.splash.dark` flag in iOS and its equivalent in Android

https://docs.expo.dev/versions/latest/config/app

@pfrazee Im happy to take this on. Will be fun!

pfrazee:
@ansh go for it

pfrazee:
Closed by https://github.com/bluesky-social/social-app/pull/2448

benharri:
There's still a full white screen before the splash image loads. I'm not seeing splash.dark changed in that PR

scottdotjs:
I think this issue should be reopened. [This post by `@mukuuji.bsky.social`](https://bsky.app/profile/mukuuji.bsky.social/post/3klji4fvsdk2c) demonstrates why:

![bafkreidmvtj6xguczgxc6l3esds7aekk6gai2epxr7v2mz2hirebue45ru](https://github.com/bluesky-social/social-app/assets/32358661/47b9b4c1-8270-4036-ba36-8f0c7302a073)

To be slightly less tongue in cheek, for me as an Android user in dark mode, what I see after tapping the app icon is:
1. Dark blue screen with blue/white app icon
1. brief fade to completely white screen
1. night sky with stars
1. white butterfly appears
1. butterfly zooms to full white screen
1. fade to black dark mode of app.

So that's dark, full brightness, dark, full brightness, dark. It's actually quite unpleasant and I now tend to look away until the app is fully loaded. :(

estrattonbailey:
Update here, we've tweaked this again, and the butterfly is now a deep, almost-black blue. Should prevent blinding users. Will be included in next release, pending any further changes 👍 

I don't think we've seen #2 before, I'll note that and investigate. Thanks for the details on that!

scottdotjs:
[...]

PR #3105, opened 2024-03-05
Improve splash display on android

edo-bari-ikutsu:
The current Android app displays a white screen before the splash screen. A white status bar is also displayed. I think this white display degrades the user experience of the app (especially on dark mode).

current_light current_dark

By applying this fix, white display is gone and startup and splash become more continuous and natural.

pr_light pr_dark

Note: this fix is only for Android apps, as I don't have iOS devices. To avoid side effects to the iOS app, I have applied some conditional branches and code separation between Android and iOS. 
* Example 1: Apart from the existing `SPLASH_CONFIG` and `DARK_SPLASH_CONFIG` in `app.config.js`,
`SPLASH_CONFIG_ANDROID` and `DARK_SPLASH_CONFIG_ANDROID` added.
* Example 2: `<StatusBar />` in App.native.tsx is only added on Android.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -144,6 +144,7 @@ module.exports = function (config) {
           },
         ],
         './plugins/withAndroidManifestPlugin.js',
+        './plugins/withAndroidStylesWindowBackgroundPlugin.js',
         './plugins/shareExtension/withShareExtensions.js',
       ].filter(Boolean),
       extra: {
--- a/app.config.js
+++ b/app.config.js
@@ -11,6 +11,17 @@ const DARK_SPLASH_CONFIG = {
   resizeMode: 'cover',
 }
 
+const SPLASH_CONFIG_ANDROID = {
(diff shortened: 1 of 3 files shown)


surfdude29:
It looks like this would fix #2292.

Also fwiw I think the white screen on startup is only an issue on Android. In my experience, iOS displays the splash image (starry sky or blue sky) immediately after opening the app. I also couldn't find any issues reporting the white screen pre-splash on iOS.

haileyok:
Another awesome PR!

Yup, this isn't an issue on iOS. Wasn't aware of this expo config plugin at all, thank you so much for sharing!

haileyok (approved):
Thank you!

haileyok:
@edo-bari-ikutsu By the way, this is definitely something that might be worth bringing up in the @expo repo if you feel like it. Seems to be something that a ton of people run into!

PR #3112, opened 2024-03-05
Refactor `PostDropdownBtn` to use new `Menu`

estrattonbailey:


haileyok:
Okay I really don't understand this one 🥴

- On Firefox, this works perfect!
- On Chrome, I still get the weird issue where it's slightly offset
- On Safari...


https://github.com/bluesky-social/social-app/assets/153161762/a25865c0-1f3e-4011-b8ab-ae639bc7f5bb

Almost certain it's because of the scroll bar appearing/disappearing. Is there any way we can stop that from happening? Could it be the `<List>` component we are using on web causing that?

haileyok (approved):
Okay, I fixed the scrollbar disappearing act. Works great now on Chrome, Safari, and Firefox!

haileyok (changes requested):
Oh Chrome still has an offset issue.

haileyok (approved):
Yea, that's fine. Let's get accessibility fixed and look at why the gutter is messed up later.

PR #3114, opened 2024-03-05
Fabric - Running on iOS and Android 🥳

haileyok:


patches/metro-runtime+0.80.4.patch
```diff
@@ -1,50 +0,0 @@
...
-+}
-+
- // Don't use a Symbol here, it would pull in an extra polyfill with all sorts of
- // additional stuff (e.g. Array.from).
- const EMPTY = {};
-@@ -303,7 +310,30 @@ function loadModuleImplementation(moduleId, module) {
-     throw module.error;
-   }
-   if (__DEV__) {
--    var Systrace = requireSystrace();
```
gaearon, line 23: This would be nice to keep / re-add. I use it sometimes.

gaearon: Ah nvm it's not main.

PR #3122, opened 2024-03-06
Initial feature gating and A/B testing integration

gaearon:
This introduces a proof of concept setup we plan to use for gated feature rollout and AB testing with Statsig.

## Scope

This proof of concept is very limited:

- Currently it's web-only. I'll hook up native later on after we get the initial sense of event volume.
- We're not yet sending any product analytic events except the built-in ones statsig does (like crash reporting and some page load metrics). We'll be adding those later.
- We're not yet applying any sampling. We'll do that later after we get a sense of the volume.

The goal here is just to get something running and make sure the setup works end-to-end.

## Details

I've chosen a slightly quirky approach. I want to preserve two constraints:

1. We should never block startup on fetching the A/B gate config.
2. We also shouldn't change A/B values midsession for the user. (At least on the web.)

So the approach I'm going with is that we always use stale values from `localStorage` from the last run. This means that feature rollout settings (e.g. bumping the %) kicks in on the _next_ page load rather than on the page load after it was done. So rollouts lag behind a bit, but the upside is the UI is always consistent and startup is fast.

Statsig SDK already deals with `localStorage` for us, so no need to do something special.

See code comments for more details.

https://github.com/bluesky-social/social-app/assets/810438/970c0db9-6968-4528-acae-ff7b73692200

Let's experiment with this approach and see if there are any downsides.

## How to use this

You call `useGate('test_gate')` and it gives you `false` (default) or `true`.

Gates can be created in the Statsig console.

Here's the toy code I used in the video above.

```js
import {useGate} from '#/lib/statsig/statsig'

function TestGate() {
  const value = useGate('test_gate')
  let color = value ? 'green' : 'red'
  let text = value ? 'PASS' : 'FAIL'
  return (
    <div
      style={{
        position: 'fixed',
        left: '50%',
        transform: 'translateX(-50%)',
        backgroundColor: color,
        width: 200,
        height: 200,
        color: 'white',
        fontSize: 30,
        zIndex: 100,
        fontFamily: 'monospace',
      }}>
      {text}
    </div>
  )
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,6 +177,7 @@
     "react-responsive": "^9.0.2",
     "rn-fetch-blob": "^0.12.0",
     "sentry-expo": "~7.0.1",
+    "statsig-react": "^1.36.0",
     "statsig-react-native-expo": "^4.6.1",
     "tippy.js": "^6.3.7",
     "tlds": "^1.234.0",
--- a/package.json
+++ b/package.json
@@ -177,6 +177,7 @@
     "react-responsive": "^9.0.2",
     "rn-fetch-blob": "^0.12.0",
     "sentry-expo": "~7.0.1",
+    "statsig-react-native-expo": "^4.6.1",
     "tippy.js": "^6.3.7",
     "tlds": "^1.234.0",
     "use-deep-compare": "^1.1.0",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -177,6 +177,8 @@
     "react-responsive": "^9.0.2",
     "rn-fetch-blob": "^0.12.0",
     "sentry-expo": "~7.0.1",
+    "statsig-react": "^1.36.0",
+    "statsig-react-native-expo": "^4.6.1",
```
gaearon, line 181: This one technically isn't used yet.

haileyok (approved):

PR #3134, opened 2024-03-06
Adjust FlatList performance in main feeds

haileyok:
## Why

This is the first step of working to improve FlatList performance. There's a couple of different parts that need to be fixed up, but this one addresses window size and the initial render.

### Current Behavior

Screenshot 2024-03-06 at 2 10 04 PM

The behavior right now renders 10 posts initially (10 posts here means 10 `FeedItem`s, not 10 individual posts, but that is a separate issue). After those ten posts, we fill ten more viewports worth of posts.

Window Size: https://reactnative.dev/docs/optimizing-flatlist-configuration#windowsize
Initial Number: https://reactnative.dev/docs/optimizing-flatlist-configuration#initialnumtorender

In the average feed, we will see between 4-6 posts. On some of the largest screens, the largest number of posts that we will be able to see on the initial render is eight. (There's also a somewhat relevant correlation between screen size and device performance, although that's not a solid test and shouldn't be relied on)

#### Maximum viewable posts

We also see that there is a _very_ large number of posts rendered after those initial posts. In this screenshot, the number is 30 posts that are rendered. (This number varies as FlatList determines the number of posts that are necessary to fill the viewport. Shorter posts will result in a higher number of posts being rendered, and vice versa. The main point is that it prepares - by default - ten viewports for scrolling).

### initialNumToRender

In this first part of the solution (that would need to be applied in the different list implementations throughout the app, for now only implemented in the main `Feed` component and `Hashtag` for more testing), we create a `useInitialNumToRender()` hook that determines the appropriate number of posts to render initially for that specific device.

#### What we know

- We only want to render what we need
- We want to make sure that if a post is cut off at the bottom but not fully visible, we want that post to be rendered
- We want to add a bit more wiggle room, so we should add one more post on to the initial render.
- The screen size determines the number of posts that will be required to fill the viewport
- **The minimum height of a post is 100 pixels**

With these constraints in mind, we can reliably calculate the number of posts that a user needs for their viewport by simply taking taking the ceiling of `SCREEN_HEIGHT` * `MINIMUM_ITEM_HEIGHT` and adding one to that value.

### windowSize

This is the same adjustment that we made to post threads. During testing, we have not seen any issues in decreasing the window size from 21 to 11 (so, 5 viewports above, 5 below, and one in the middle). This applies the same adjustment to the feeds.

## Results

On the feeds screen, we end up doing significantly less work to display the user's feed.
Screenshot 2024-03-06 at 2 11 27 PM

Another way to look at it is the significantly lower amount of upfront CPU cycles on the render (top is before, bottom is after)

We can use the hashtags screen to see the results of scroll performance, before and after

https://github.com/bluesky-social/social-app/assets/153161762/9bece367-f0a4-4b6a-ac5c-f165efb47988

## Next steps

Note in the Feed renders, some items have two items being rendered inside of them. This is because of how we are creating a <FeedItemGroup> (or similar name, cannot remember off the top of my head). This isn't ideal, as we're essentially rendering (as of right now) up to 20 posts for the initial render, whenever a _maximum_ of 4 might be required (again, dependent on screen size). Ideally we would have each of these be its own item given to the FlatList.

### DIFF
--- a/src/lib/hooks/useInitialNumToRender.ts
+++ b/src/lib/hooks/useInitialNumToRender.ts
@@ -1,0 +1,13 @@
+import React from 'react'
+import {useWindowDimensions} from 'react-native'
+
+const MIN_POST_HEIGHT = 100
+
+export function useInitialNumToRender(minItemHeight: number = MIN_POST_HEIGHT) {
+  const {height: screenHeight} = useWindowDimensions()
+
+  return React.useMemo(
+    () => Math.ceil(screenHeight / minItemHeight) + 1,
+    [screenHeight, minItemHeight],
+  )
+}
--- a/src/lib/hooks/useInitialNumToRender.ts
+++ b/src/lib/hooks/useInitialNumToRender.ts
@@ -1,13 +1,11 @@
 import React from 'react'
-import {useWindowDimensions} from 'react-native'
+import {Dimensions} from 'react-native'
 
 const MIN_POST_HEIGHT = 100
 
 export function useInitialNumToRender(minItemHeight: number = MIN_POST_HEIGHT) {
-  const {height: screenHeight} = useWindowDimensions()
-
-  return React.useMemo(
-    () => Math.ceil(screenHeight / minItemHeight) + 1,
-    [screenHeight, minItemHeight],
-  )
+  return React.useMemo(() => {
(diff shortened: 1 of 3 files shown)


gaearon (approved):
good investigation, thanks for diving in

PR #3143, opened 2024-03-07
ALF confirmation dialogs (Dialogs Pt. 3)

haileyok:
### This PR builds off of https://github.com/bluesky-social/social-app/pull/3144

We need to fix a good number of our dialogs for cases where accessible font sizes are used. Some dialogs that need to be scrollable are not, which results in some users not being able to do things like delete posts. In an effort to fix a number of accessibility problems with the old Modals, this PR creates a "Confirmation Prompt" that can be re-used.

Screenshot 2024-03-06 at 8 02 52 PM


haileyok:
Recorded a little video of all of this stuff combined 👀 

https://github.com/bluesky-social/social-app/assets/153161762/d231d5b6-9a8f-4243-949c-301f74511a71

src/lib/hooks/useOTAUpdate.ts
```diff
@@ -2,25 +2,9 @@ import * as Updates from 'expo-updates'
...
-      name: 'confirm',
-      title: t`Update Available`,
-      message: t`A new version of the app is available. Please update to continue using the app.`,
-      onPressConfirm: async () => {
-        Updates.reloadAsync().catch(err => {
-          throw err
-        })
-      },
-    })
-  }, [openModal])
```
haileyok, line 23: Whenever we do OTA updates we don't want to do this anyway. Could be breaking iOS guidelines.

src/components/Prompt.tsx
```diff
@@ -119,7 +120,7 @@ export function Action({
   return (
     <Button
       variant="solid"
-      color="primary"
+      color={color}
       size={gtMobile ? 'small' : 'medium'}
```
haileyok, line 124: I think it makes sense to do this here, because `Action` is just built off of a `Button`, and we'd have access to the color on `Button` anyway.

src/view/screens/AppPasswords.tsx
```diff
@@ -260,6 +258,23 @@ function AppPassword({
         </Text>
       </View>
       <FontAwesomeIcon icon={['far', 'trash-can']} style={styles.trashIcon} />
+      <Prompt.Outer control={control}>
+        <Prompt.Title>
+          <Trans>Delete app password?</Trans>
+        </Prompt.Title>
+        <Prompt.Description>
+          <Trans>Are you sure you want to delete the app password</Trans> "
+          {name}"?
```
surfdude29, line 267: Sorry to jump in here, could I suggest that all of this string is marked for localisation? As I was reminded recently while updating the French translation, the sentence structure can be different in other languages, there can be differences such as _guillemets_ being used instead of quotation marks and spaces can be needed before certain punctuation marks like `?`

```suggestion
          <Trans>Are you sure you want to delete the app password "
          {name}"?</Trans>
```

haileyok: Heh yea, I forget lingui handles that. Thanks!

estrattonbailey (approved):

PR #3146, opened 2024-03-07
Fix reactivity of dialogs (Dialogs Pt. 1)

haileyok:
## Why

We recently made a change for the dialogs that allowed for updating the `importantForAccessibility` prop of the app's root view. We have to set the value to `no-hide-descendants` when a dialog is open, otherwise TalkBack will be able to highlight the text/elements beneath the bottom sheet.

When we did this, we started storing the list of active modals in state, and passed around a few values that were reactive to the open/closed status of a sheet. After this, every dialog began to re-render when one was opened up.

Screenshot 2024-03-07 at 12 41 49 PM

## Solution

Instead of leveraging state to know when to set the `no-hide-descendants` value, we can instead change the root view to an `Animated.View` and use a `SharedValue` to update the `importantForAccessibility` value. We still make use of the new `setDialogIsOpen` function, but instead of updating state, we just add/remove the dialog's ID to the old `Set` and also update the `importantForAccessibility` shared value.

Screenshot 2024-03-07 at 3 52 40 PM

https://github.com/bluesky-social/social-app/assets/153161762/cb6bc79b-77c1-44cc-a9bf-280ecbd18f3c

(The few "missing swipes" are the Android video recorder modal picking up the gestures)

This does not include any changes for iOS or web.


src/components/Dialog/context.ts
```diff
@@ -36,14 +35,13 @@ export function useDialogControl(): DialogOuterProps['control'] {
     () => ({
       id,
       ref: control,
-      isOpen,
```
estrattonbailey, line 39: This prop is used by the new `Menu` 🤔

src/state/dialogs/index.tsx
```diff
@@ -41,26 +47,42 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+    return openDialogs.current.size > 0
+  }, [])
 
   const setDialogIsOpen = React.useCallback(
     (id: string, isOpen: boolean) => {
-      setOpenDialogs(prev => {
-        const filtered = prev.filter(dialogId => dialogId !== id) as string[]
-        return isOpen ? [...filtered, id] : filtered
-      })
+      if (isOpen) {
```
estrattonbailey, line 62: The other thing I considered here was an emitter to pass state around outside the react tree

src/view/shell/index.tsx
```diff
@@ -80,19 +81,9 @@ function ShellInner() {
...
-   *
-   * @see https://reactnative.dev/docs/accessibility#importantforaccessibility-android
-   */
-  const importantForAccessibility =
-    openDialogs.length > 0 ? 'no-hide-descendants' : undefined
-
   return (
     <>
-      <View
+      <Animated.View
```
estrattonbailey, line 86: Super clever I love it

src/components/Dialog/context.ts
haileyok: From what I could tell, the menu handles it on its own right? So I changed the `isOpen` value to be optional. I merged in the dropdown menu and it still worked fine.

```tsx
export function useMenuControl(): Dialog.DialogControlProps {
  const id = React.useId()
  const [isOpen, setIsOpen] = React.useState(false)

  return React.useMemo(
    () => ({
      id,
      ref: {current: null},
      isOpen,
      open() {
        setIsOpen(true)
      },
      close() {
        setIsOpen(false)
      },
    }),
    [id, isOpen, setIsOpen],
  )
}
```

src/components/Dialog/types.ts
```diff
@@ -22,7 +22,7 @@ export type DialogControlRefProps = {
 export type DialogControlProps = DialogControlRefProps & {
   id: string
   ref: React.RefObject<DialogControlRefProps>
-  isOpen: boolean
+  isOpen?: boolean
```
haileyok, line 25: This right here makes it optional, so we can use it in the menu but not need it for the dialog. The way the menu works on web doesn't cause re-renders either like we saw with the dialogs, as it doesn't need to change a prop on the root `<View>`

src/state/dialogs/index.tsx
haileyok: That's an option too! This felt like the lower lift to keep the old behavior, but if we want to do that I'm totally open to it.

src/view/shell/index.tsx
haileyok: Yea, it's bizarre to think about at first as to why it works, but the gist is that the `SharedValue` creates a reference on the native side that can be updated synchronously, and therefore the underlying `UIView` (or whatever it is on android) can just update on the native side rather than causing a React re-render. Not something we want to do very often obviously, but in situations like this it's pretty useful.

estrattonbailey (approved):
Nice one!

PR #3148, opened 2024-03-08
rm waitlist modal, button during sign up

haileyok:
There is no longer a waitlist, and the only time this shows up now is if you try to make a new account on a third-party service. Should be good to 🗑️


gaearon:
Are third-parties allowed to have invites? I think maybe they were.

haileyok:
Yea they have invites, but the waitlist API is dead now. It pointed here to https://bsky.app/api/waitlist so it doesn't apply to any third parties.

gaearon (approved):
got it

PR #3150, opened 2024-03-08
Make ALF prompt scrollable for accessibility

haileyok:
One of the driving motivators of moving these modals to ALF is accessibility. If users are using a very large font size, they are currently unable to scroll down to things like "Delete Post" inside of a modal. This is also an issue with the Prompt, since we just use a `Dialog.Inner`. We should instead use a `Dialog.ScrollableInner`, so that if the font size is too large, the user can scroll to the needed button.

Screenshot 2024-03-07 at 8 44 09 PM
Screenshot 2024-03-07 at 8 44 13 PM

### DIFF
--- a/src/components/Prompt.tsx
+++ b/src/components/Prompt.tsx
@@ -3,7 +3,7 @@ import {View, PressableProps} from 'react-native'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
-import {useTheme, atoms as a} from '#/alf'
+import {useTheme, atoms as a, useBreakpoints} from '#/alf'
 import {Text} from '#/components/Typography'
 import {Button} from '#/components/Button'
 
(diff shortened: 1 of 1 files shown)


src/components/Prompt.tsx
```diff
@@ -71,8 +72,17 @@ export function Description({children}: React.PropsWithChildren<{}>) {
...
+
   return (
-    <View style={[a.w_full, a.flex_row, a.gap_sm, a.justify_end]}>
+    <View
+      style={[
+        a.w_full,
+        a.flex_row,
+        a.gap_sm,
+        a.justify_end,
+        gtMobile && [a.pb_4xl],
```
haileyok, line 84: Adding padding on mobile keeps the buttons from reaching the edge of the safe area.

estrattonbailey (approved):

PR #3151, opened 2024-03-08
Use ALF for signup flow, improve a11y of signup

haileyok:
Going through the sign up flow right now is a bit strange. We jump from one UI look to another once we get to the onboarding. The forms also do not scale well with scaled font sizes.

This PR keeps the same layout for signup as before, but instead uses ALF for the forms. It also improves on validation prior to submission (i.e. a valid email is required before proceeding, unlike currently where you can complete the captcha and submit before realizing it is invalid). State logic is the same as the old logic, except laid out like that of the Onboarding.

Wouldn't bother merging until the login flow is also at the same place.

https://github.com/bluesky-social/social-app/assets/153161762/60f211ac-58cd-499a-91bb-685df9e63899

Screenshot 2024-03-08 at 3 27 58 AM
Screenshot 2024-03-08 at 2 57 52 AM
Screenshot 2024-03-08 at 2 57 55 AM
Screenshot 2024-03-08 at 3 20 44 AM
Screenshot 2024-03-08 at 3 20 53 AM

Forgot to fix the service selector button before recording. Added a "pencil edit" to the right hand side for clarity.

Screenshot 2024-03-08 at 3 11 21 AM


src/screens/Signup/StepHandle.tsx
```diff
@@ -0,0 +1,123 @@
...
+        {state.error ? (
+          <View style={[a.w_full, a.flex_row, a.align_center, a.gap_sm]}>
+            <IsValidIcon valid={false} />
+            <Text style={[a.text_md, a.flex]}>{state.error}</Text>
+          </View>
+        ) : undefined}
+        <View style={[a.w_full, a.flex_row, a.align_center, a.gap_sm]}>
+          <IsValidIcon valid={validCheck.handleChars} />
+          <Text style={[a.text_md, a.flex]}>
+            <Trans>May only contain letters and numbers</Trans>
```
surfdude29, line 95: Could this string be updated please to clarify that hyphens are also allowed in handles? (albeit not at the start or end, but I don't think that needs to be mentioned)

```suggestion
            <Trans>May only contain letters, numbers and hyphens</Trans>
```

haileyok: Agree it doesn't need to be shown usually, but still want to clarify if a user does put a `-` at the start or end. Added some logic to account for that and only display the message if thats the case. Thank you!

surfdude29: That is a much better solution, actually 💫

Also, I think the new strings are clearer 👍  (I forgot the Oxford comma in en-US in my suggested version, oops 😅)

haileyok: Yea the other ones felt TOO polite haha, who says _MAY_? 😆

surfdude29:
I noticed something that I think is worth looking into: some of the `accessibilityLabel`s and `accessibilityHint`s seem to have gone AWOL in the ALF-ification?

In `src/view/com/auth/create/Step1.tsx` ➡️ `src/screens/Signup/StepInfo.tsx`:

When you compare the `Hosting provider` field

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step1.tsx#L73-L109

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepInfo.tsx#L53-L85

The `Invite code` field

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step1.tsx#L138-L158

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepInfo.tsx#L92-L114

The `Email address` field

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step1.tsx#L162-L184

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepInfo.tsx#L115-L135

The `Password` field

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step1.tsx#L186-L208

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepInfo.tsx#L136-L155

And when you compare the DOB field

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step1.tsx#L210-L231

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepInfo.tsx#L156-L178

you see that the `accessibilityLabel` and `accessibilityHint` are there, but I think they should be marked for translation?



In `src/view/com/auth/create/Step2.tsx` ➡️ `src/screens/Signup/StepHandle.tsx`:

When you compare the handle field:

https://github.com/bluesky-social/social-app/blob/4f8381678da505737a96b7420c0f1ea8329f3672/src/view/com/auth/create/Step2.tsx#L68-L81

with

https://github.com/bluesky-social/social-app/blob/ae8070d339f75bdf28b400dabb712b3066df6142/src/screens/Signup/StepHandle.tsx#L58-L69

haileyok:
Good find on the `Date`, should be localized.

Re: the other ones, I need to actually verify (or have you?) but this should be handled underneath in `TextField`. We wanted it to be more straight forward for us so we can't accidentally forget to add them. If it _doesn't_ work then it needs to get fixed underneath. Let me see what it sounds like w/ VoiceOver.

surfdude29:
Oops, I think you're completely right, sorry 😬

I hadn't realised that accessibility props were handled like that in ALF, that's what comes from when I comment on coding stuff and I don't know what I'm talking about... 🤦🏻‍♂️

haileyok:
All good! You're 100% right on the `DateInput` props so those need fixing. Going to see if we can move that into ALF too so I can't forget those 😅 

Still want to test this anyway to make sure that what I'm saying _is_ correct. That's how it was explained to me but not sure we've given it a full spin yet.

src/screens/Signup/StepCaptcha.tsx
```diff
@@ -62,33 +44,32 @@ export function Step3({
...
-  }, [_, uiDispatch])
+  }, [_, dispatch])
 
   return (
-    <View>
-      <StepHeader uiState={uiState} title={_(msg`Complete the challenge`)} />
+    <Animated.View
+      style={[a.gap_lg]}
+      entering={FadeInRight}
+      exiting={FadeOutLeft}>
```
mozzius, line 57: In the login flow, I disabled this on web. Do we want to do the same here? (or we could reenable for the login)

mozzius (commented):
Looks a little jank on web due to transitions, looks good on iOS though!

src/screens/Signup/StepCaptcha.tsx
mozzius: I am leaning towards disable on web

mozzius:
Screenshot 2024-03-19 at 18 41 00

Text overflow :(

src/screens/Signup/StepCaptcha.tsx
haileyok: Just used `ScreenTransition`. Just had to change where I used it. Reanimated layout transitions confuse me sometimes :/

PR #3156, opened 2024-03-08
Fix RSS URLs treated as internal

estrattonbailey:
Emily noticed that RSS URLs are treated as internal atm. This fixes that.

Test post: https://bsky.app/profile/emily.bsky.social/post/3kn52xdg3cc2a

### DIFF
--- a/src/lib/routes/router.ts
+++ b/src/lib/routes/router.ts
@@ -2,9 +2,15 @@ import {RouteParams, Route} from './types'
 
 export class Router {
   routes: [string, Route][] = []
-  constructor(description: Record<string, string>) {
+  constructor(description: Record<string, string | string[]>) {
     for (const [screen, pattern] of Object.entries(description)) {
-      this.routes.push([screen, createRoute(pattern)])
+      if (typeof pattern === 'string') {
+        this.routes.push([screen, createRoute(pattern)])
+      } else {
+        pattern.forEach(subPattern => {
+          this.routes.push([screen, createRoute(subPattern)])
+        })
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


haileyok:
Work on web but on native it's giving `useOpenLink` the value `/profile/haileyok.com/rss`. We might need a router update for this too (I guess could also just prepend `https://bsky.app` if the url starts with `/` but that doesn't feel good)

estrattonbailey:
@haileyok I think it's OK to prepend the host, since we DO want RSS links to open in a browser. Right?

But we do need some sort of router update to handle Safari trying to link the user back to the app, which results in a 404.

CleanShot 2024-03-08 at 17 37 11@2x

haileyok:
Okay yea, that should be fine. As for the Safari issue, yea the router is probably trying to push to the same relative path I think...maybe the router should just push to the profile page in that case?

estrattonbailey:
Yeah I think just route to profile is fine. Not sure what else we could do since iOS will try to universal link to the app for those URLs 🤔

haileyok:
Okay, this might be a solution. We can modify the router to allow for an array of possible paths rather than a single one.

_That being said_, now need to make sure that the Go server is handling `/rss` for us before we even hit the router. Going to build and see.

haileyok:
Okay yea, this works. The go server handles the RSS on its own:

```go
	// profile RSS feed (DID not handle)
	e.GET("/profile/:ident/rss", server.WebProfileRSS)
```

Screenshot 2024-03-08 at 5 49 14 PM

The router will still catch `/rss` here though on native since the deeplink goes to `bluesky://` instead of the Go server. Might be useful to have multiple routes for another reason in the future anyway.

haileyok (approved):
As long as we are okay with the router change, the rest of this works on both platforms.

PR #3159, opened 2024-03-09
Fix `aria-label` on the Share button

aliceisjustplaying:
The `aria-label` on the Share button was the same as the one on the Like button.

Before:
image

After:
image


haileyok:
Ahh this has been on the list for a bit and just never got to it. Thank you! 🙏

haileyok (approved):

PR #3165, opened 2024-03-10
Fix button widths in the Settings on web

### LINKED ISSUE #3060: "Dim" button must have more characters to fit the larger translation in just one line!
image

aliceisjustplaying:
Make the buttons full width in settings so it supports longer strings in other languages.

This is the least bad solution with the current constraints per my research. This would be solved neatly if RN would support `box-sizing: content-box` but it does not.

It's possible there is a more elegant solution, I'm not a CSS expert but I did spend quite a bit googling about this.

Before:
image

After:
image

Fixes #3060.

### DIFF
--- a/src/view/com/util/forms/SelectableBtn.tsx
+++ b/src/view/com/util/forms/SelectableBtn.tsx
@@ -57,6 +57,7 @@ const styles = StyleSheet.create({
   btn: {
     flexDirection: 'row',
     justifyContent: 'center',
+    flexGrow: 1,
     borderWidth: 1,
     borderLeftWidth: 0,
     paddingHorizontal: 10,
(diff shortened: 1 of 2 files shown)


aliceisjustplaying:
Ah this is so much worse than I thought.

aliceisjustplaying:
Okay, whatever, let's just set fixed width, least bad solution.

haileyok:
I think adding `flexGrow: 1` to the `btn` style in `SelectableBtn.tsx` would be what you're looking for. Looks like all the places we use `SelectableBtn` would be fine with this.

aliceisjustplaying:
Plenty more flexbox to learn :) Thanks! Looks good here as well.

haileyok:
8iqque

Flexbox regularly confuses the heck out of me all the time, especially in RN. Thanks for fixing this!

haileyok (approved):

Issue #2028, opened 2023-11-29
There is no way to unfollow a user if they blocked you


Ai-rin:
**Describe the bug**

Users who block me still counts as followed, but i can't unfollow them

**To Reproduce**

Have somebody block you then try to unfollow them

**Expected behavior**

Users who blocked me should still be present in the follow list so I can unfollow them.

**Details**

Build version: 1.56.0; Platform: web

meicpalmer:
Blocking someone should make them unfollow you, so people can perform softblocks.

Ai-rin:
> Blocking someone should make them unfollow you, so people can perform softblocks.

Either way, if i still follow somebody, they shouldn't disappear from my follow list.

scottdotweb:
Could we get confirmation from the team that this will be addressed, please?

surfdude29:
This absolutely should be fixed so that there's a way to do it through the app, but in the meantime you can do it using this great tool:

https://cleanfollow-bsky.pages.dev/

mschwendener:
Not sure if this is related: 
I have unfollowed https://bsky.app/profile/gladlib.bsky.social but still see their posts in the OnlyPosts feed. 
This has been going on for a while; following & unfollowing again does not fix it.

X1Vi:
Guys I have added a fix and added a video I hope it is to your expectations.

eddieoz:
Is there a chance this could resolve the inability to unfollow people, potentially due to asynchronous issues like this: https://github.com/bluesky-social/atproto/issues/3256 ?

welaughindoors:
> Is there a chance this could resolve the inability to unfollow people, potentially due to asynchronous issues like this: [bluesky-social/atproto#3256](https://github.com/bluesky-social/atproto/issues/3256) ?

Echoing this, experiencing the same issues and its infuriating to have hundreds of people (likely bots) that I cannot unfollow and are cluttering my feed.

PR #3168, opened 2024-03-11
Use new menu for Profile

haileyok:
This converts the Profile dropdown menu to the new dropdown. Also adding a `minWidth` of 170 to the dropdown on web, since some menus appear very small if there is only a few items/those items have a short description.


src/view/com/profile/ProfileMenu.tsx
```diff
@@ -0,0 +1,274 @@
...
+                        }
+                        onPress={onPressBlockAccount}>
+                        <Menu.ItemText>
+                          {profile.viewer?.blocking ? (
+                            <Trans>Unblock Account</Trans>
+                          ) : (
+                            <Trans>Block Account</Trans>
+                          )}
+                        </Menu.ItemText>
+                        <Menu.ItemIcon icon={Block} />
```
haileyok, line 250: Suggestions for icons here would be nice. Can't find something that works well for both block _and_ unblock. Maybe something in the mod PR will be nice?

src/view/com/profile/ProfileMenu.tsx
```diff
@@ -0,0 +1,274 @@
...
+                />
+              </TouchableOpacity>
+            )
+          }}
+        </Menu.Trigger>
+
+        <Menu.Outer>
+          <Menu.Group>
+            <Menu.Item
+              testID="profileShareBtn"
```
estrattonbailey, line 188: Was `profileHeaderDropdownShareBtn`, should check if this affects e2e

src/components/Menu/index.web.tsx
```diff
@@ -125,6 +125,7 @@ export function Outer({children}: React.PropsWithChildren<{}>) {
             a.p_xs,
             t.name === 'light' ? t.atoms.bg : t.atoms.bg_contrast_25,
             t.atoms.shadow_md,
+            {minWidth: 170},
```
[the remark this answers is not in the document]
estrattonbailey: Could also allow a `style` prop here on web for cases where we do want this.

estrattonbailey (approved):
LGTM, will have a look at icons quick

src/view/com/profile/ProfileMenu.tsx
haileyok: oops, nice find.

mozzius:
Added a fix for #2028

src/components/Menu/index.tsx
```diff
@@ -68,7 +68,9 @@ export function Trigger({children, label}: TriggerProps) {
   })
 }
 
-export function Outer({children}: React.PropsWithChildren<{}>) {
+export function Outer({
+  children,
+}: React.PropsWithChildren<{style?: ViewStyle | ViewStyle[]}>) {
```
estrattonbailey, line 73: You can use `StyleProp<ViewStyle>`

haileyok: 🤯

haileyok:
@mozzius Did you want to keep that follow/unfollow button there? I think it's fine but just making sure.

PR #3174, opened 2024-03-11
Combine actions, convert to new menu

estrattonbailey:
CleanShot 2024-03-11 at 14 04 13@2x
CleanShot 2024-03-11 at 14 04 07@2x
CleanShot 2024-03-11 at 14 04 04@2x

### DIFF
--- a/src/view/screens/ProfileFeed.tsx
+++ b/src/view/screens/ProfileFeed.tsx
@@ -1,11 +1,10 @@
 import React, {useMemo, useCallback} from 'react'
-import {Dimensions, StyleSheet, View} from 'react-native'
+import {Dimensions, StyleSheet, View, Pressable} from 'react-native'
 import {NativeStackScreenProps} from '@react-navigation/native-stack'
 import {useIsFocused, useNavigation} from '@react-navigation/native'
 import {useQueryClient} from '@tanstack/react-query'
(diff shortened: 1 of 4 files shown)


haileyok (approved):
LGTM, we'll see if Dan has any suggestions here. We do need to make this same change for the UserList screen, but that should be a fairly easy copy-paste over.

src/view/screens/ProfileFeed.tsx
```diff
@@ -263,130 +270,132 @@ export function ProfileFeedScreenInner({
...
+              <NewButton
+                testID={isPinned ? 'unpinBtn' : 'pinBtn'}
+                disabled={isPinPending || isUnpinPending}
+                size="small"
+                variant="solid"
+                color={isPinned ? 'secondary' : 'primary'}
+                label={isPinned ? _(msg`Unpin from home`) : _(msg`Pin to home`)}
+                onPress={onTogglePinned}>
+                <ButtonText>
+                  {isPinned ? _(msg`Unpin`) : _(msg`Pin to home`)}
```
gaearon, line 299: How about `Pin to Home`? To capitalize the screen name.

estrattonbailey, line 299:
```suggestion
                  {isPinned ? _(msg`Unpin`) : _(msg`Pin to Home`)}
```

PR #3177, opened 2024-03-11
Convert profile edit avatar/banner dropdown menus to new menu

haileyok:
Also adding a prop, `showCancel`, to the `Menu.Outer` component. The cancel button is copied from `TagMenu`. Since it only applies on native, it seems easier to keep the logic for showing this inside of the `Menu` component instead of having to add it with an `isNative` check each time we make a menu. Having the Cancel button on some menus gives a better appearance imo. We can add that to some of the others later if we feel like they need it.

## Native
Screenshot 2024-03-11 at 2 09 32 PM

## Web
Screenshot 2024-03-11 at 2 07 34 PM


src/view/com/util/UserAvatar.tsx
```diff
@@ -216,118 +221,121 @@ let EditableUserAvatar = ({
...
+                  style={aviStyle}
+                  source={{uri: avatar}}
+                  accessibilityRole="image"
+                />
+              ) : (
+                <DefaultAvatar type={type} size={size} />
+              )}
+              <View style={[styles.editButtonContainer, pal.btn]}>
+                <FontAwesomeIcon
+                  icon="camera"
```
estrattonbailey, line 284: Could use the new icon here and on the banner if ya want

src/view/com/util/UserAvatar.tsx
```diff
@@ -216,118 +221,121 @@ let EditableUserAvatar = ({
...
-          size={12}
-          color={pal.text.color as string}
-        />
-      </View>
-    </NativeDropdown>
+    <EventStopper onKeyDown={false}>
+      <Menu.Root>
+        <Menu.Trigger label={_(msg`Edit avatar`)}>
+          {({props}) => (
+            <TouchableOpacity {...props} activeOpacity={0.8}>
```
estrattonbailey, line 271: Note to us: should test this again after #3182

estrattonbailey (approved):
LGTM, tested and works. Left a few nits!

PR #3179, opened 2024-03-12
Dedupe navigation events (push, navigate, pop, etc)

haileyok:
## Why

It's currently possible to encounter situations where you accidentally navigate to a screen twice, i.e. if you quickly press on a post or press twice when switching to a different tab. While this is certainly possible in various `onPress` events throughout the app, most are better handled with logic such as adding `disabled` states to buttons. However, in the case of our `Link` component or `BottomBar` component, it's more intuitive to create a small wrapper for `useNavigation()` that offers deduped functions such as `.push()` and `.navigate()`.

I am only applying this in the places where it is most needed. Generally, `<Link>` components are where we see this the most, as there is a small delay between press and rendering of the next screen. There's also some noticable delay it seems on Android when switching tabs, so the dedupe logic is added to the `BottomBar` navigation logic.

We don't need to do this on web, so the logic is not used there. It also is not generally necessary in some other places where we dismiss a modal or close the drawer upon press, since there is not time to press those links twice. Regardless, `Link` covers most cases where we encounter these events.

## Before

https://github.com/bluesky-social/social-app/assets/153161762/f0add5d4-6130-4965-8341-315bec68de53

## After

https://github.com/bluesky-social/social-app/assets/153161762/5524f589-c07d-44f1-baa9-a4e9e63696cc

### DIFF
--- a/src/view/shell/bottom-bar/BottomBar.tsx
+++ b/src/view/shell/bottom-bar/BottomBar.tsx
@@ -36,6 +36,7 @@ import {Button} from '#/view/com/util/forms/Button'
 import {s} from 'lib/styles'
 import {Logo} from '#/view/icons/Logo'
 import {Logotype} from '#/view/icons/Logotype'
+import {useDedupe} from 'lib/hooks/useDedupe'
 
 type TabOptions = 'Home' | 'Search' | 'Notifications' | 'MyProfile' | 'Feeds'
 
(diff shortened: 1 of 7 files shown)


src/view/shell/bottom-bar/BottomBar.tsx
```diff
@@ -74,12 +76,12 @@ export function BottomBar({navigation}: BottomTabBarProps) {
...
-        navigation.dispatch(StackActions.popToTop())
+        dedupe(() => navigation.dispatch(StackActions.popToTop()))
       } else {
-        navigation.navigate(`${tab}Tab`)
+        dedupe(() => navigation.navigate(`${tab}Tab`))
       }
     },
-    [track, navigation],
+    [track, navigation, dedupe],
   )
```
haileyok, line 85: `BottomBar` can't use the `useNavigation()` hook, so we'll just dedupe these with the `useDedupe()` hook instead.

estrattonbailey (approved):
Nice solve!

PR #3180, opened 2024-03-12
Save image to user media library when taken from camera during composing

haileyok:
Simple one-liner. Saves pics taken with the camera to the media library as well when composing a post.

### DIFF
--- a/src/view/com/composer/photos/OpenCameraBtn.tsx
+++ b/src/view/com/composer/photos/OpenCameraBtn.tsx
@@ -14,6 +14,7 @@ import {isMobileWeb, isNative} from 'platform/detection'
 import {logger} from '#/logger'
 import {useLingui} from '@lingui/react'
 import {msg} from '@lingui/macro'
+import {saveImageToMediaLibrary} from 'lib/media/manip'
 
 type Props = {
   gallery: GalleryModel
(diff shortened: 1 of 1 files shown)


surfdude29:
Apologies for jumping in (yet again), I'm wondering: does `{saveImageToMediaLibrary}` automatically take care of camera roll/photo gallery permissions if the user hasn't granted that? Sorry if it's a stupid question

haileyok:
It actually should (for me it worked fine on Android) but there's no harm in checking again (might not work on iOS).

surfdude29:
Sounds sensible to me (but you obviously know much more about these things than I do :)  The thought occurred as I know Camera and Photos are separate permissions on iOS (I don't have any experience with Android) and when I just tested on a fresh install of the current app, taking a photo prompted for Camera permission but not Photos.

And sorry for the annoying questions btw, thank you for being so friendly and patient in answering and explaining to someone who's very much a newbie 🙏

haileyok:
Feel free! Very likely that the permissions are different, and even when testing it could just be lingering perms from an old install.

We should actually probably move the logic for checking perms right into these helper functions rather than having to do do the perms check logic each time we call openCamera() or openImagePicker() and prevent forgetting to do it. Something for another time though 😅

surfdude29:
> We should actually probably move the logic for checking perms right into these helper functions rather than having to do do the perms check logic each time we call openCamera() or openImagePicker() and prevent forgetting to do it. Something for another time though 😅

Yeah I know you all have a long enough TODO list as is 😅

estrattonbailey (approved):

PR #3182, opened 2024-03-12
Improve `Menu.Trigger` on web

estrattonbailey:
On the first go-around on Menu I hadn't considered just passing the props from `asChild` through, but after using it a couple times I realized this would be the better move.

This PR makes it so that most `Menu`s should just render a `Pressable` within `Menu.Trigger`. Should be pretty easy since we just spread `props`.

This PR is web only.


mozzius (approved):
Yeah way better

PR #3183, opened 2024-03-12
`PostThread` cleanup

haileyok:
## Why

1. A continued effort to normalize error screens, spinners, etc. across the app. In `PostThread` we had two duplicated error screens (some of which didn't work at all, as in the case of a blocked post. Right now you just get an empty page). We also had some unnecessary logic to add padding to the bottom of the screen, which in some cases such as on web would allow us to scroll far enough to make all of the content disappear.
2. Given the perf changes we've made to `FlatList`, we don't need to be concerned with the "Load More" feature. Instead, we can let virtualization handle this.
3. Overall cleanup of this screen allows us to remove the extra wrapper for the loading states. We can re-use our loading screen logic from `Lists.tsx` - which will handle loads, errors, and empty states - instead, all from within the single component.

### Part One
The error screen from `Lists.tsx` is extracted to it's own component, so that we can use it anywhere. The props `title` and `message` allow configuration of the screen. Also, a `retry` prop is added for cases where we want to offer that ability. Not passing anything to `retry` will result in the button not being shown.

We then use that error screen to display our various errors in `PostThread`. We create a `error` object that has the different error messages, and just pass that into `ListMaybePlaceholder`. Any time we encounter an error, that screen will be shown instead of the thread. `ListMaybePlaceholder` also handles the "full screen" spinner whenever we don't have any content to display on push (i.e. no placeholder data).

Finally, we add a prop to `ListFooter` that lets us set a height. On most screens, this could be a rather small value such as 180 pixels (the default). However, in the thread view, we want this value to be approximately the height of the screen minus the minimum post size. This gives us the scrollability we want when maintaining position on a post, even when that is the last post in the chain.

### Part Two

- Add `initialNumToRender` to `PostThread` using the `useInitialNumToRender()` hook.
- Remove the `LoadMore` button.
- Add logic to update `maxReplies` when we reach the end of the screen.

Initially, I tried removing `maxReplies` entirely. On native, this worked alright. Virtualization handled things elegantly, and there was not a noticeable performance drop (at least on iOS). However, on web we would quickly reach 700MB of memory use and there would be a brief freeze in the page load.

Instead, I opted to use `onEndReached` to update `maxReplies` in increments of 50 posts. The user won't see any buttons anymore however, and the posts will be automatically loaded in.

### Part Three

Because we are adding a single error screen instead of various screens for different error states, it makes sense to consolidate the logic into a single component, rather than a wrapper that renders `PostThreadLoaded` whenever the query has finished. There is little logic changes needed here. The only things that have changed are that `skeleton` may possibly be null. In that case, we simply return an empty array for `posts` until the skeleton has been created (this happens once thread query finishes loading or has placeholder data to display).

### Normal post load

https://github.com/bluesky-social/social-app/assets/153161762/e7ca3f3d-9b92-400a-8e20-b0293d1df66b

[screen-capture.webm](https://github.com/bluesky-social/social-app/assets/153161762/3e589ad5-93f3-4918-a899-5b1474aeb3aa)

### Initial Load

https://github.com/bluesky-social/social-app/assets/153161762/f45f624d-719b-401d-97c6-03b72e865083

[screen-capture (1).webm](https://github.com/bluesky-social/social-app/assets/153161762/5d5267ba-37c6-42e7-8235-0175299d5646)

### Large Thread

https://github.com/bluesky-social/social-app/assets/153161762/81bbe9bc-6754-49f6-8062-6dea170e6a4f

[screen-capture (2).webm](https://github.com/bluesky-social/social-app/assets/153161762/c8b59c6f-5018-40bb-ae8b-b43ed29bc86e)

### Not Found

https://github.com/bluesky-social/social-app/assets/153161762/d6702909-3a06-4810-82c4-f994fc6b5db1

[screen-capture (3).webm](https://github.com/bluesky-social/social-app/assets/153161762/d57723e3-7642-422a-809d-5c535fd78003)

### Blocked User

Screenshot 2024-03-12 at 12 54 01 PM

Screenshot 2024-03-12 at 12 55 22 PM


src/view/com/post-thread/PostThread.tsx
```diff
@@ -325,46 +307,6 @@ function PostThreadLoaded({
...
-            // @ts-ignore web-only
-            style={{
-              // Leave enough space below that the scroll doesn't jump
-              height: isNative ? 600 : '100vh',
-              borderTopWidth: 1,
-              borderColor: pal.colors.border,
-            }}
-          />
-        )
-      } else if (item === CHILD_SPINNER) {
```
haileyok, line 362: We don't need this border hack anymore, because the `ListFooter` applies a top border on its own.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -230,17 +213,11 @@ function PostThreadLoaded({
       if (!highlightedPost.post.viewer?.replyDisabled) {
         arr.push(REPLY_PROMPT)
       }
-      if (highlightedPost.ctx.isChildLoading) {
-        arr.push(CHILD_SPINNER)
-      } else {
```
haileyok, line 235: Spinners are also added through the `ListFooter` component now.

pfrazee (approved):
This is great!

PR #3188, opened 2024-03-12
Add warning about sharing if post author has the !no-unauthenticated label

### LINKED ISSUE #2538: When I choose "Copy link to post" - please let me know if viewing the post requires sign-in

When the user who created the post has Moderation/Logged out visibility = false, I might unknowingly share a useless link with the public on some other site.

After I click "Copy link to post," maybe show me a modal stating "Only signed-in users will be able to view this post." If that is indeed the case.

I could manually test the link in a private browser window.

I could take screenshots of the app, and share the post that way.

mozzius:
Fixes #2538 

Screenshot 2024-03-13 at 18 21 22

Screenshot 2024-03-12 at 20 56 36
^outdated wording

Same warning is also on the share option in the profile

Screenshot 2024-03-13 at 18 20 39

Let me know if the wording can be improved


mozzius:
Really messed up git here, I don't even know what happened

PR #3189, opened 2024-03-12
Sort imports using eslint-plugin-simple-import-sort

mozzius:
Did some spot checking, seems ok

There's an option to use regex to manually sort the imports, if we're not happy with the default order


.eslintrc.js
```diff
@@ -17,7 +17,29 @@ module.exports = {
...
+          ['^\\u0000'],
+          // Node.js builtins prefixed with `node:`.
+          ['^node:'],
+          // Packages.
+          // Things that start with a letter (or digit or underscore), or `@` followed by a letter.
+          // React/React Native priortized, followed by expo
+          [
+            '^(react/(.*)$)|^(react$)|^(react-native(.*)$)',
+            '^(expo(.*)$)|^(expo$)',
+            '^@?\\w',
```
mozzius, line 34: Unfortunately relative unprefixed imports that we have get grouped into here, because we can't really tell them apart

mozzius:
Maybe should it be set to `warn` instead of `error` so that the checks pass? I think the auto-fix will still catch it

estrattonbailey (approved):
Hells yeah, this is awesome 🔥

PR #3191, opened 2024-03-13
Fix dropdown close via a portaled backdrop

estrattonbailey:


haileyok (approved):

haileyok:
Tested on mobile Safari and mobile Chrome  (Android), works great.

PR #3193, opened 2024-03-13
Tweak prompt styles on mobile web

estrattonbailey:
CleanShot 2024-03-12 at 20 02 38@2x
CleanShot 2024-03-12 at 20 02 26@2x

### DIFF
--- a/src/view/com/profile/ProfileMenu.tsx
+++ b/src/view/com/profile/ProfileMenu.tsx
@@ -305,7 +305,7 @@ let ProfileMenu = ({
         confirmButtonCta={
           profile.viewer?.blocking ? _(msg`Unblock`) : _(msg`Block`)
         }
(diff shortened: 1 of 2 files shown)


src/view/com/profile/ProfileMenu.tsx
```diff
@@ -305,7 +305,7 @@ let ProfileMenu = ({
         confirmButtonCta={
           profile.viewer?.blocking ? _(msg`Unblock`) : _(msg`Block`)
         }
-        confirmButtonColor="negative"
+        confirmButtonColor={profile.viewer?.blocking ? undefined : 'negative'}
```
estrattonbailey, line 308: Just makes the button blue when unblocking, red when blocking

haileyok (approved):

PR #3195, opened 2024-03-13
[Statsig] Track likes, reposts, follows

gaearon:
## What

Adds `post:like` and `post:unlike` events, similar to our Segment setup. Using lowercase for the new ones to distinguish from existing Segment events, as I don't aim to port them 1:1.

I'm introducing the notion of "log context", in the sense of which-part-of-the-UI-the-action-originated-from. It will be a required parameter for most mutations. Open to other names but I want to keep it relatively short, have "log" in the name to discourage any other attempts to use this information, and something to imply it's contextual to the UI.

I'm using component names here. I didn't reify this in an explicit type which is maybe a bad idea but I want to see how far we can take it with just copy paste / find-and-replace. We don't change the fundamental product UI structure that much.

I decided against inferring this "log context" automatically (e.g. via React context) to prevent refactor hazards. E.g. you don't want a refactor that moves `useMutation` hook up the tree to start losing the log context. So it's threaded through at every level and you can see what it is.

I used component names for the values of log context but they could've been something more specific too. E.g. we could do something like `PostThreadItem[highlighted]`. Or some kind of breadcrumbs. I think this is an ok start. I don't know what will be useful.

I made log context required so that if you're adding a new call to the mutation, you're forced to figure out what it should be. I kept the existing decision to fire events in mutations because this ensures we wouldn't miss them.

I'm keeping the Segment events as is for now. The new stuff is additive.

## Test Plan

Like and unlike a post. Observe the events.

Screenshot 2024-03-13 at 03 50 48

### DIFF
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -1,4 +1,4 @@
-export type Events = {
+export type LogEvents = {
   init: {
     initMs: number
   }
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -2,4 +2,10 @@ export type Events = {
   init: {
     initMs: number
   }
+  'post:like': {
+    logContext: 'FeedItem' | 'PostThreadItem' | 'Post'
+  }
+  'post:unlike': {
+    logContext: 'FeedItem' | 'PostThreadItem' | 'Post'
+  }
 }
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -14,4 +14,22 @@ export type Events = {
   'post:unrepost': {
     logContext: 'FeedItem' | 'PostThreadItem' | 'Post'
   }
+  'profile:follow': {
+    logContext:
+      | 'RecommendedFollowsItem'
+      | 'PostThreadItem'
+      | 'ProfileCard'
+      | 'ProfileHeader'
+      | 'ProfileHeaderSuggestedFollows'
+      | 'ProfileMenu'
+  }
+  'profile:unfollow': {
+    logContext:
+      | 'RecommendedFollowsItem'
+      | 'ProfileCard'
+      | 'PostThreadItem'
+      | 'ProfileHeader'
+      | 'ProfileHeaderSuggestedFollows'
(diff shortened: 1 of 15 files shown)


src/lib/statsig/events.ts
```diff
@@ -2,4 +2,10 @@ export type Events = {
   init: {
     initMs: number
   }
+  'post:like': {
+    logContext: 'FeedItem' | 'PostThreadItem' | 'Post'
+  }
+  'post:unlike': {
+    logContext: 'FeedItem' | 'PostThreadItem' | 'Post'
```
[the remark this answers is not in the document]
gaearon: i guess my concern is that i do want to get more specific sometimes (like profile header) so we’ll want canonical names for these concepts — and so we might as well just use component names as first approximation.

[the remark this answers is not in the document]
gaearon: and then if the component structure changes sufficiently to have them renamed, that might be a good signal that we want to rename the contextual names too — since they correspond to different kind of UI elements. assuming we don’t go around renaming things for its own sake. also note that there’s no requirement the call is done in the same exact component — as long as it’s somewhere above i think the approximation is close enough.

estrattonbailey (approved):
Works for me!

src/state/queries/post.ts
```diff
@@ -111,22 +113,32 @@ export function usePostLikeMutationQueue(
...
+    },
     onSuccess() {
       track('Post:Like')
     },
   })
 }
 
-function usePostUnlikeMutation() {
+function usePostUnlikeMutation(
+  logContext: 'FeedItem' | 'PostThreadItem' | 'Post',
```
estrattonbailey, line 135: Oh ya know, should we re-use the `Events['post:unlike']['logContext']` type for these to keep in sync?

gaearon: yeah i was leaning towards that as well.. ok lemme do that real quick

gaearon: ok yeah i did this for mutations, but kept explicit enumerations for components since those are more local and i think you might only want a subset for props.

PR #3201, opened 2024-03-13
Filter out non-image files from image picker

### LINKED ISSUE #2369: Uploading non-image files results in incorrect behavior.
**Describe the bug**
When uploading non-images (mp4, mkv), etc., an error is displayed only in the console and a preview of the image that is just black.
Uploading a long video will cause the browser to run very heavy and display a black preview after a few seconds.

**To Reproduce**

1. Open Post window from New Post
2. Press the Image Button
3. Change the file extension to All Files in Windows Explorer and select files with extensions other than the one specified.

**Expected behavior**

It should indicate in a user-visible manner an error if an unsupported file extension is loaded.
Also, videos should not be previewed.

**Screenshots**

スクリーンショット 2023-12-29 204409

**Details**

 - Platform: desktop chrome windows
 - Platform version:  120.0.6099.130
 - App version: 1.61.0

**Additional context**

mozzius:
Fixes #2369 

Checks if the images returned from the image picker are indeed images, since the browser file picker lets you override the requested mime type. Shows a toast if anything gets filtered out.

### DIFF
--- a/src/lib/media/picker.shared.ts
+++ b/src/lib/media/picker.shared.ts
@@ -18,11 +18,18 @@ export async function openPicker(opts?: ImagePickerOptions) {
     Toast.show('You may only select up to 4 images')
   }
 
-  return (response.assets ?? []).slice(0, 4).map(image => ({
-    mime: 'image/jpeg',
-    height: image.height,
-    width: image.width,
-    path: image.uri,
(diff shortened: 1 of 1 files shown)


haileyok (approved):
lgtm!

PR #3212, opened 2024-03-14
Add package diff PR labeler

haileyok:
Adding a GitHub action based on https://github.com/expo/expo/blob/main/.github/workflows/pr-labeler.yml to find changes to packages and native modules.


mozzius (approved):
go for it

Issue #3059, opened 2024-02-29
Tapping posts doesn't reliably work on Android


gaearon:
See thread: https://bsky.app/profile/whiskerfatigue.bsky.social/post/3kl5ga3ducl2g

>I experience the same thing in the Android app. My feeling is that the tap detection is overly sensitive to the scrolling motion. Even the slightest amount of y-movement prevents the tap from registering.

Video repro:

https://github.com/bluesky-social/social-app/assets/810438/5943978e-58b7-44b6-b234-6445c54202d0

Issue #2729, opened 2024-02-01
Notifications not responding to taps on android


pfrazee:
Users are reporting that notifications become unresponsive to taps after some usage in android (see https://bsky.app/profile/yourhappyhost.com/post/3kkekzkqqsp2o)

PR #3214, opened 2024-03-14
Remove `FixedTouchableHighlight` , fix Android press issues

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/3059
Fixes https://github.com/bluesky-social/social-app/issues/2729

## Why

We are currently using `TouchableHighlight` for some of our `Link` components. It appears that `TouchableHighlight` has some issues when used with `PagerView`. To fix that, we are using a `FixedTouchableHighlight` "patch" that was found in an older GitHub issue.

This fix checks for a translation in the press on both the X and Y axes before "allowing" the press. If the translation is over 2 pixels in any direction, the press is canceled. This creates weird behaviors - especially on lower end devices. We should remove this "fix" and use `Pressable` instead, with `android_ripple` prop to add an effect that is expected on Android.

We only use this `FixedTouchableHighlight` for things like `FeedItem`, `PostThreadItem`, etc. This also only affects Android. iOS and web hovers, presses, etc. will remain the same.

## Before

Notice that even the most minor of translations causes the press to be cancelled.

https://github.com/bluesky-social/social-app/assets/153161762/6da4e7f8-4492-4ecc-89b3-7d00077b388d

## After

https://github.com/bluesky-social/social-app/assets/153161762/fb39c6e5-80d3-4fa2-bc1c-c47318addb0a

## Test Plan

Run through the app in various locations such as notifications, profile, etc. (preferably on a mid or low range Android device). When tapping, try to move your finger every so slightly while tapping and notice how easy it is to cancel a press.

Test this build (should be a production build, but still on a mid or low end device) and notice that pressing is significantly easier.


gaearon (approved):

haileyok:
Adjusted slightly so that we can move the ripple to the background instead of the foreground. Unnecessary `pal.view` styles were removed (the parent background colors are the same, so the duplicate style is unnecessary). Colors look much better now and are in-line with Twitter.

PR #3217, opened 2024-03-15
Use ALF for login & signup flow

mozzius:
The signup portion was in #3151, which has since been merged into this one.

I've broadly kept the state management the same, while moving the components over to ALF. The biggest difference in functionality is adding a ScrollView to all the forms, which should hopefully improve a11y for users with larger font sizes.
 
I also added transition animations between the various screens - let me know if we should keep it or not, it's very simple to remove (just wrapped the content in an animated view that has enter/exit animations with a key of the screen name). We could make it native only perhaps, since the web one looks a little jankier and there's no expectation of screen transitions there?
_Edit: I've since made it native-only and it was the right call, web feels way better without it. The video still show it with the transitions so you can see what it looked like_

https://github.com/bluesky-social/social-app/assets/10959775/ef0a8b7c-d756-41f7-94f1-22f793771a7c

https://github.com/bluesky-social/social-app/assets/10959775/c4dee49c-9d4d-4719-9d69-8aa1ded4fdc5

https://github.com/bluesky-social/social-app/assets/10959775/52e28410-b456-4d9c-acb5-4346a5493277


haileyok:
yes

src/screens/Login/SetNewPasswordForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+        password,
+      })
+      onPasswordSet()
+    } catch (e: any) {
+      const errMsg = e.toString()
+      logger.warn('Failed to set new password', {error: e})
+      setIsProcessing(false)
+      if (isNetworkError(e)) {
+        setError(
+          'Unable to contact your service. Please check your Internet connection.',
```
surfdude29, line 76: This looks so good, very nicely done, especially the animations on native 👌

Could this string be kept localisable please? 🙏

mozzius: Sure

src/screens/Login/FormError.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import {StyleSheet, View} from 'react-native'
+
+import {Warning_Stroke2_Corner0_Rounded as Warning} from '#/components/icons/Warning'
+import {Text} from '#/components/Typography'
+import {atoms as a, useTheme} from '#/alf'
+import {colors} from '#/lib/styles'
+
+export function FormError({error}: {error?: string}) {
```
haileyok, line 9: We should just move this to `/src/components/forms` to reuse ❤️

src/screens/Login/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+      )
+      logger.warn(`Failed to fetch service description for ${serviceUrl}`, {
+        error: String(serviceError),
+      })
+    } else {
+      setError('')
+    }
+  }, [serviceError, serviceUrl, _])
+
+  const onPressRetryConnect = () => refetchService()
```
haileyok, line 88: Small nit, I think this just got copied over when you moved it from the other directory, but let's just put `refetchService` right in the `onPress` prop instead of this.

src/screens/Login/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+      title = _(msg`Forgot Password`)
+      description = _(msg`Let's get your password reset!`)
+      content = (
+        <ForgotPasswordForm
+          error={error}
+          serviceUrl={serviceUrl}
+          serviceDescription={serviceDescription}
+          setError={setError}
+          setServiceUrl={setServiceUrl}
+          onPressBack={gotoForm(Forms.Login)}
```
haileyok, line 136: Could just make these arrow functions too, at a first glance it looks weird having the function get called right here until you scroll up 😅

mozzius: Sure lol

mozzius:
Screenshot 2024-03-19 at 18 06 16

This screen is pretty terrible - do we have a reference for a prettier way to show some text and a confirm button?

haileyok (approved):
Did a full pass here on all platforms

- Signup flow works correctly
- Service selection works correctly
- Password reset works correctly
- Login works correctly
- Selecting an already logged in acc works correctly
- Logic to handle invite code req in signup works
- Logic for requiring a captcha also works

Unless I'm missing a flow here, this looks great.

src/components/icons/Pencil.tsx
```diff
@@ -0,0 +1,5 @@
+import {createSinglePathSVG} from './TEMPLATE'
+
+export const Pencil_Stroke2_Corner0_Rounded = createSinglePathSVG({
```
estrattonbailey, line 3: Nit: name of the file is fine, since this can contain all pencil variations, but looks like the name of this icon is "pencil-line" as opposed to "pencil-wave" or just "pencil". Nit, but leaves room for others in the future.

assets/icons/pencil_stroke2_corner0_rounded.svg
estrattonbailey, line 1: Looks like this export doesn't match the icons file. There should be a lil whitespace round the icon.

src/components/forms/FormError.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <View style={styles.error}>
+      <Warning fill={t.palette.white} size="sm" />
+      <View style={(a.flex_1, a.ml_sm)}>
+        <Text style={[{color: t.palette.white}, a.font_bold]}>{error}</Text>
+      </View>
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
```
estrattonbailey, line 24: Old styles and colors here

src/components/forms/HostingProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+          a.flex_row,
+          a.align_center,
+          a.rounded_sm,
+          a.px_md,
+          a.gap_xs,
+          {paddingVertical: isAndroid ? 14 : 9},
+          t.atoms.bg_contrast_25,
+        ]}
+        onPress={onPressSelectService}>
+        <TextField.Icon icon={Globe} />
```
estrattonbailey, line 52: Just FYI, idk if this was like this before: this Icon component is design for use in a TextField, we can just use the Globe SVG component directly here

src/components/forms/HostingProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <View
+          style={[
+            a.rounded_sm,
+            t.atoms.bg_contrast_100,
+            {marginLeft: 'auto', left: 6, padding: 6},
+          ]}>
+          <Pencil
+            style={{color: t.palette.contrast_500}}
+            height={18}
+            width={18}
```
estrattonbailey, line 63: Let's stick to using consistent sizes via `size` if we can, but otherwise all g ✌️

src/components/forms/HostingProvider.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }, [onOpenDialog, serverInputControl])
+
+  return (
+    <>
+      <ServerInputDialog
+        control={serverInputControl}
+        onSelect={onSelectServiceUrl}
+      />
+      <TouchableOpacity
```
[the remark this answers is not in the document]
estrattonbailey: Also forces us to add a `label` for a11y

src/components/forms/HostingProvider.tsx
haileyok: I think this one was on me from that whitespace oops. Shouldn't need it when we fix that 👍

src/screens/Login/ChooseAccountForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+    },
+    [currentAccount, track, initSession, onSelectAccount, setShowLoggedOut, _],
+  )
+
+  return (
+    <FormContainer
+      testID="chooseAccountForm"
+      title={<Trans>Select account</Trans>}>
+      <View>
+        <TextField.Label>
```
estrattonbailey, line 140: Nit: let's not use TextField components outside a TextField, this can just be a normal Text

src/screens/Login/ChooseAccountForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+            {i > 0 ? (
+              <View style={[a.border_b, t.atoms.border_contrast_low]} />
+            ) : null}
+            {React.cloneElement(child, {
+              // @ts-ignore
+              style: {
+                borderRadius: 0,
+                borderWidth: 0,
+              },
+            })}
```
estrattonbailey, line 44: Not really a big deal, but I think `Group` is maybe a little unnecessary here e.g. I don't think these styles are applied. Just a cleanliness thing.

assets/icons/pencil_stroke2_corner0_rounded.svg
mozzius: Tried reexporting, is the new one good?

src/screens/Login/ChooseAccountForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  return (
+    <TouchableOpacity
+      testID={`chooseAccountBtn-${account.handle}`}
+      key={account.did}
+      style={[a.flex_1]}
+      onPress={onPress}
+      accessibilityRole="button"
+      accessibilityLabel={_(msg`Sign in as ${account.handle}`)}
+      accessibilityHint={_(msg`Double tap to sign in`)}>
```
estrattonbailey, line 77: Double tap 🤔 

Could also be a `Button`, could add hover states on web

src/screens/Login/ChooseAccountForm.tsx
estrattonbailey: You know what, never mind me, the consistency is nice here

src/screens/Login/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,183 @@
...
+      </View>
+      <View>
+        <Text style={[t.atoms.text_contrast_high, a.mb_md]}>
+          <Trans>
+            Enter the email you used to create your account. We'll send you a
+            "reset code" so you can set a new password.
+          </Trans>
+        </Text>
+      </View>
+      <FormError error={error} />
```
estrattonbailey, line 125: General thought, don't need in this PR: `TextField` has an `isInvalid` boolean to give it an error state, and I'd like to start doing field errors adjacent to the field itself.

That said, a larger error state below the form is still good for network errors and the like though. All good here for now though!

mozzius: Noted - yeah, that'll be much better looking if we do that

estrattonbailey (approved):
Absolutely fire work on this

PR #3219, opened 2024-03-15
ProfileFollows and ProfileFollowers cleanup

haileyok:
## Why

Continuing on the normalization of lists, list footers, and error screens. Also cleaning up logic (although these two components were already generally okay).

The "no followers" error state was broken as well, which is now fixed as shown below.

## How

Not many noticeable changes are made on native unless you visit a list with no results. However, on web there are some more noticeable ones (or if you click links to take you into the native app).

#### Before

Screenshot 2024-03-15 at 8 48 31 AM

#### After

Screenshot 2024-03-15 at 8 49 50 AM

#### Before

Screenshot 2024-03-15 at 8 54 12 AM

#### After

Screenshot 2024-03-15 at 8 53 45 AM

## Test Plan

Navigate the follows and followers of a few users. Find some accounts that do not have any followers/following and check those pages. Also check the follows/following for a handle that does not exist.


mozzius:
Looks great, but maybe add some feedback to the retry button? Disabled it, or add a spinner or something

haileyok:
Can try, not sure how useful it will be. It should actually be showing a spinner if it does a real retry, but if the result is cached it's just "instant".

I removed the retry button from the error though if it's just a "no results" case because it felt useless.

haileyok:
Ah @mozzius you know what it is? React Query doesn't clear the error when you call `refetch`, huh. Fixed this in the base PR [`e7064bf` (#3183)](https://github.com/bluesky-social/social-app/pull/3183/commits/e7064bf54448145fb2f9ec51eb56a2bb14052407)

mozzius (approved):
Awesome, all good then 👍

pfrazee (approved):
LGTM

PR #3227, opened 2024-03-15
add `Partial<GestureResponderEvent>` to `.open()` on dialog control

haileyok:
In order to facilitate easy passing of `.open` to a `Pressable` or `TouchableX` `onPress` prop, adding the `GestureResponderEvent` type to the `options` parameter is necessary.

For example right now we have to do one of the following.

```tsx
const control = useDialogControl()

const onPressTwo = React.useMemo(() => {
	control.open()
}, [])

<Pressable onPress={() => control.open()}>
	<Text>Hello</Text>
</Pressable>

<Pressable onPress={onPressTwo}>
	<Text>Hello Again</Text>
</Pressable>
```

If we add `Partial<GestureResponderEvent>` to the `options` type, we can simply use a pressable like so

```tsx
const control = useDialogControl()

<Pressable onPress={control.open}>
	<Text>Yay</Text>
</Pressable>
```

Making this `Partial` allows us to still use `.open({index: 10})` for example if we wanted to without any of the `GestureResponderEvent`s required types.

### DIFF
--- a/src/components/Dialog/types.ts
+++ b/src/components/Dialog/types.ts
@@ -1,5 +1,5 @@
 import React from 'react'
-import type {AccessibilityProps} from 'react-native'
+import type {AccessibilityProps, GestureResponderEvent} from 'react-native'
 import {BottomSheetProps} from '@gorhom/bottom-sheet'
 
 import {ViewStyleProp} from '#/alf'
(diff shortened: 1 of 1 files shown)


mozzius (approved):
LGTM 👍

PR #3233, opened 2024-03-17
Add 'Liberation Sans' font for FireFox; lightbox buttons adjustments

alexkuz:
As seen on Firefox v123.0, Linux Mint
1. Consistent button styles in lightbox

| Before | After |
|--------|--------|
| image | image |
| image | image | 

2. Add `Liberation Sans` to prevent broken `Helvetica` replacement in Firefox on Linux (**TBD**: check other platforms/browsers)

| Before | After |
|--------|--------|
| image | image |

3. Hide scrollbar gutter if body scroll is locked (it breaks the layout in Chrome)

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -44,7 +44,7 @@
       scrollbar-gutter: stable both-edges;
     }
     html, body {
-      font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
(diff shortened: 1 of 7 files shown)


pfrazee (approved):
This LGTM. I'm not on linux so I can't test the font fallback, but I don't see any changes on FF or Chrome on macos.

In the future, if you could break up PRs into individual tasks I'd appreciate it!

pfrazee:
Also the improvements to the lightbox are all great!

alexkuz:
>In the future, if you could break up PRs into individual tasks I'd appreciate it!

absolutely!

Issue #3213, opened 2024-03-14
YouTube Shorts issues on Vivaldi


yevvie:
**Describe the bug**
**YouTube Shorts** embed in web app has odd scaling & doesn't properly launch in full screen in Vivaldi, then refreshes the feed?

**To Reproduce**
1. Click on YouTube Shorts embed in a feed.
2. Enable External Media if disabled.
3. Click Full Screen button.

**Expected behavior**
I was expecting see the content of the video...

**Preview**
Recording of the bug: https://www.youtube.com/watch?v=fpVjM6knowQ

**Details**
 - Platform: Vivaldi Browser on Windows
 - Platform version: 6.6.3271.50 (Stable channel) (64-bit) 
 - App version: Build version: 1.72.0

**Additional context**
- In Firefox the fullscreen launches properly but the initial scaling is still wrong. 
- Tested on two separate Vivaldi browsers with different addons, with and without workspaces, disabled uBlock and other stuff too, with no fix.

Issue #2985, opened 2024-02-25
YouTube shorts embeds load horizontal resolution


slrgt:
**Describe the bug**
The embed of a youtube short resizes to a horizontal resolution after it loads. ![image](https://github.com/bluesky-social/social-app/assets/114287507/d5137ca4-4d08-4a8e-a96f-cc1df82383b0)

**To Reproduce**

1. Upload youtube short
2. Press play on embed
3. It will go from the correct portrait resolution to a horizontal resolution once it loads

**Expected behavior**

It should load in the correct portrait

 - Platform: ios
 - Platform version: 1.6.8 (7)
 - App version:

PR #3238, opened 2024-03-18
Properly fill container for YT Shorts videos

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/3213
Fixes https://github.com/bluesky-social/social-app/issues/2985

## Why

When we started using the Bluesky web server to host the YouTube iframes, the CSS from the used library did not account properly for YT Shorts, instead always using a 16:9 ratio. This change instead fills the viewport with the video player, who's dimensions we set ourselves to ensure proper aspect.

## Test Plan

- Run `yarn build-web`
- Start the go server locally
- Open http://localhost:8100/profile/asfpoef.bsky.social
- In a simulator, view the same profile while the go server is still running.

Screenshot 2024-03-17 at 9 00 58 PM

Screenshot 2024-03-17 at 9 00 39 PM

Screenshot 2024-03-17 at 9 01 06 PM

Screenshot 2024-03-17 at 9 01 10 PM


mozzius (approved):
Nice, tested and it works well

Issue #3236, opened 2024-03-18
Android image picker can be double opened which crashes the app


Juicysteak117:
**Describe the bug**
On Android, double tapping the image picker results in two sets of the imager picker being opened. Attempting to close them causes the app to crash.

**To Reproduce**

1. Double click the image picker
2. Close both image pickers
3. Observe

**Expected behavior**

Only one should open and it shouldn't crash

**Screenshots**

Cut the video short since I realized it doxxed me on the home screen accidentally, but it fully crashes to home with a "Bluesky stopped responding" message after the second time.

https://github.com/bluesky-social/social-app/assets/20595808/536c950d-6b38-465c-a4ac-1e64b0b9ec3c

**Details**

 - Platform: Android
 - Platform version: 13
 - App version: Latest

**Additional context**

PR #3240, opened 2024-03-18
prevent duplicate pickers from opening on android

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/3236

## Why

On Android, the image picker can accidentally be opened twice by the user. This doesn't happen often, but when it does it causes a native crash. We should keep track of whether the picker is open or not before launching the picker again.

I opted to keep track of this on the native side, since if we kept track of it on the JS side we could still potentially run into async issues.

## Test Plan

Use a current build and open the picker twice by quickly pressing the image library button in the post composer. Observe that you have to close it twice, and on the second close it crashes.

Use this build and attempt to open the picker twice again, but observe it only opens once.


mozzius:
I can still seem to be able to trigger the crash, at least on the emulator (my real device uses a different picker which doesn't have the issue in the first place)

haileyok:
@mozzius Heh nice. Uploaded some videos to show it. Might be hard to trigger in a production build, but with the dev client it's a bit easier since its slower.

pfrazee (approved):
LGTM

Juicysteak117:
Thanks Hailey!

PR #3247, opened 2024-03-18
Fix instances of static translations (`t` macro)

mozzius:
Changed all the places I could find where the `t` macro is used, since this doesn't change when changing the app language, and replaced it with `msg`

### DIFF
--- a/src/view/com/composer/select-language/SelectLangBtn.tsx
+++ b/src/view/com/composer/select-language/SelectLangBtn.tsx
@@ -88,7 +88,7 @@ export function SelectLangBtn() {
       ...arr.slice(0, 6),
       {sep: true},
       {
-        label: _(`Other...`),
+        label: _(msg`Other...`),
         onPress: onPressMore,
       },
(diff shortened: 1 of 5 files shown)


src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -84,15 +84,15 @@ export function SelectLangBtn() {
     }
 
     return [
-      {heading: true, label: t`Post language`},
+      {heading: true, label: _(msg`Post language`)},
       ...arr.slice(0, 6),
       {sep: true},
       {
-        label: t`Other...`,
+        label: _(`Other...`),
```
gildaswise, line 91:
```suggestion
        label: _(msg`Other...`),
```

I guess?

mozzius: Good catch, thanks!

haileyok (approved):
👍

PR #3252, opened 2024-03-18
Bday modal tweaks

estrattonbailey:
Mostly set out to integrate the ALF DateField, but decided we [should do a little more work](https://github.com/orgs/bluesky-social/projects/9/views/1?pane=issue&itemId=56843822) on that before we start using it.

Also wanted to fix:
- flash of load state when submitting
- remove need to pass in prefs (prefs should be cached, might as well use query hook)

So just small tweaks:
- hoisted some UI up for initial load state (probably won't ever happen, prefs should be in cache)
- moved `isPending` state to button icon

CleanShot 2024-03-18 at 13 51 59@2x
CleanShot 2024-03-18 at 13 58 22

### DIFF
--- a/src/components/dialogs/BirthDateSettings.tsx
+++ b/src/components/dialogs/BirthDateSettings.tsx
@@ -43,7 +43,25 @@ export function BirthDateSettingsDialog({
         </View>
 
         {isLoading ? (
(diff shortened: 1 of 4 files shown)


haileyok (approved):
Nice

PR #3253, opened 2024-03-18
Fix nested border radius issue on square avatars

mozzius:
Cherrypicked commits from #3230 due to 3p moderation branch being merged


haileyok (approved):
BRIT

PR #3257, opened 2024-03-18
Use consistent avatar shape/defaults for labelers

estrattonbailey:
Uses `labeler` variant of avatar in all cases


pfrazee (approved):
LGTM

PR #3258, opened 2024-03-18
Update the reporting flow to first select a recipient if the user has multiple labelers

pfrazee:
- If the user has >1 labeler subscribed, adds an initial screen where you choose the moderation service
- Doesn't change any other logic, including the ability to check multiple recipients in the final step
- If the user has only 1 labeler, doesnt do that initial step

CleanShot 2024-03-18 at 15 58 48@2x
CleanShot 2024-03-18 at 15 58 58@2x
CleanShot 2024-03-18 at 15 59 07@2x


estrattonbailey (approved):
🧙‍♂️

Issue #3246, opened 2024-03-18
On desktop, a horizontal scrollbar appears on the feed list div, when it does not seem necessary. (Windows, not MacOS)


adamatic-me:
**Describe the bug**
When the vertical scrollbar appears on the feed list, an unnecessary horizontal scrollbar also appears.

**To Reproduce**

1. On Windows, sub to enough feeds so that the feed list div requires a vertical scrollbar.
3. Notice the horizontal scrollbar which now appears on the feed list div. 

**Expected behavior**

There is no reason for a horizontal scrollbar to appear.

A CSS hack to make it go away: 
```
.r-1ipicw7 {
  width: unset; 
}
```

**Screenshots**
The image below shows Bksy as it is currently. The CSS hack above makes the horizontal scroll not appear.

![image](https://github.com/bluesky-social/social-app/assets/146130101/8d373eba-472d-4c1e-af10-adb6f3a00b96)

**Details**

 - Platform: desktop chrome firefox windows (does not happen on macos safari chrome, as scrollbars are generally handled much better on macos)
 - Platform version: all
 - App version: Build version: 1.73.0; Platform: web

adamatic-me:
Fixed in Build version: 1.96.2; Bundle info: 3aa1a5f (prod); Bundle date: 24122022; Platform: web; Platform version: undefined

Issue #3248, opened 2024-03-18
Add CSS { scrollbar-width: thin; } to div and root so that scrollbars look better on Windows, as they do on MacOS


adamatic-me:
**Is your feature request related to a problem? Please describe.**

On Windows, (and maybe Linux?) scrollbars are ugly by default. On MacOS, scrollbars are handled much prettier and thinner. Let's pass that on to Windows users.

**Describe the solution you'd like**
I am a muggle, but this CSS hack led to the results that I would like to see:

:root, div { 
  scrollbar-width: thin; 
}

**Additional context**

https://caniuse.com/?search=scrollbar-width

Before screenshot:
![image](https://github.com/bluesky-social/social-app/assets/146130101/a18eff47-f654-401d-9977-4b32640c8b49)

After screenshot:
![image](https://github.com/bluesky-social/social-app/assets/146130101/14ed19ce-d931-4dc2-ae12-33e80c93c9f5)

adamatic-me:
An even prettier, more MacOS-like, and IMO the ideal experience hides the scroll bar on divs until hover:

  :root {
    scrollbar-width: thin; 
  }
  div {
    scrollbar-width: none;
  }
  div:hover {
    scrollbar-width: thin;
  }


https://github.com/bluesky-social/social-app/assets/146130101/a09359d0-e911-4367-a6ce-b716e8e4a375

PR #3276, opened 2024-03-19
don't show vertical scroll bar in desktop sidebars

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/3246

## Why

As seen in this image, scrollbars on Windows in our left and right navs cause a pretty ugly layout. Generally this isn't as bad on macOS since the OS handles scrollbars a bit better, but they can also look bad there too.

img

## How

If we replace the current `View` that wraps these with a `ScrollView` and set the `showVerticalScrollIndicator` prop to `false`, these are not displayed, creating a better look.

## Before - Right Nav

Screenshot 2024-03-19 at 1 35 50 PM
Screenshot 2024-03-19 at 1 36 01 PM

## After - Right Nav

Screenshot 2024-03-19 at 1 35 41 PM
Screenshot 2024-03-19 at 1 36 33 PM
Screenshot 2024-03-19 at 1 36 37 PM

## Before - Left Nav

Screenshot 2024-03-19 at 1 39 00 PM
Screenshot 2024-03-19 at 1 39 06 PM

## After - Left Nav

Screenshot 2024-03-19 at 1 39 27 PM
Screenshot 2024-03-19 at 1 40 38 PM

### DIFF
--- a/src/view/shell/desktop/RightNav.tsx
+++ b/src/view/shell/desktop/RightNav.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {StyleSheet, View} from 'react-native'
+import {ScrollView, StyleSheet, View} from 'react-native'
 import {usePalette} from 'lib/hooks/usePalette'
 import {DesktopSearch} from './Search'
 import {DesktopFeeds} from './Feeds'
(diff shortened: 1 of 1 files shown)


haileyok:
Mm, after looking at Twitter's layout, they do this a bit differently. Closing in favor of another option.

adamatic-me:
oh, @haileyok  - somehow I missed in my search before making an issue related to this: https://github.com/bluesky-social/social-app/issues/3248#issuecomment-2004329588

TL;DR: I proposed only showing the scrollbar on hover, and also setting scrollbar-width: thin;

PR #3286, opened 2024-03-20
[Statsig] Track login/logout

gaearon:
Adds tracking for login/logout operations.

This only tracks operations *initiated by the user*. We don't count automatic logins here. (The way those work is pretty convoluted so I think for now I'd just like to keep track of explicit actions to err on the side of understanding the events.)

I'm also only counting success for now. If we see significant gaps compared to `init` events we can dive deeper.

I've placed the events themselves closer to the API calls (so we don't miss adding new calls to them) but I require passing in `logContext` from outside which forces us to add attribution for which part of the UI triggered the event.

It's a bit ugly not not too bad to thread it through.

## Test Plan

- Login form
- Sign out button in Settings
- Switch between accounts in Settings
- Switch between accounts in mobile bottom bar

Observe `logEvent` calls match expectations by logging locally.

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -565,7 +565,11 @@ function RoutesContainer({children}: React.PropsWithChildren<{}>) {
 }
 
 function getCurrentRouteName() {
-  return navigationRef.getCurrentRoute()?.name
+  if (navigationRef.isReady()) {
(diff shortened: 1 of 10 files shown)


pfrazee (approved):
Yeah lgtm

tuzkundev (commented):
LGTM

PR #3288, opened 2024-03-20
Set keyboardDismisMode to interactive on the report dialog

pfrazee:
Because the default behavior is to dismiss on-drag, we run into this:

https://github.com/bluesky-social/social-app/assets/153161762/a3e215d7-162d-49cc-b6e8-adb8165ebd18

This adds a param to control the behavior and switches it to interactive on the report dialog.

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -206,12 +206,16 @@ export function Inner({children, style}: DialogInnerProps) {
   )
 }
 
-export function ScrollableInner({children, style}: DialogInnerProps) {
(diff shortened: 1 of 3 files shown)


haileyok (approved):

haileyok:
https://github.com/bluesky-social/social-app/assets/153161762/a3e215d7-162d-49cc-b6e8-adb8165ebd18

haileyok:
@estrattonbailey Any idea on this one from the video above? Would be nice if we could get this to work correctly.

PR #3291, opened 2024-03-20
Add OTA updates support for `testflight` channel

haileyok:
## Why

This PR adds OTA update functionality to the app. Updates will use the `updates.bsky.app` service, which is based on the Expo Updates protocol. https://github.com/expo/custom-expo-updates-server

There are two channels for updates: `production` and `testflight`. In this PR, we are only enabling the `testflight` channel. Users that are using the production app will not receive OTA updates, and the production client will not make any requests to the server at this time.

After some time of testing the internal channel, we can enable the `production` channel to start getting a sense of how many requests are going to the server and to see if there are any changes we need to make there. The good news is that even if the server goes down or starts returning errors, the client will simply ignore the requests/errors and continue to operate normally.

We should also update the workflow at a later point to support automatic deployments to the staging server, as well as manual workflow for releasing to web.

## Workflow

Untitled

## How Channels Work

There is an environemtn variable inside of `.env` called `EXPO_PUBLIC_ENV` that can be either set to `production` or `testflight` depending on how we run the release. Using `eas build --profile testflight` for example will automatically set `EXPO_PUBLIC_ENV` to `testflight`. This will set the app to download updates from the `testflight` channel. Likewise, using `eas build --profile production` will set the app to download updates from the `production` channel.

There should not be any changes to our current workflow when submitting new releases to the app store. The GitHub action is configured to either use the `testflight` or `production` channel now. @pfrazee notably it might be worth taking a moment to properly add the credentials for Google Play to make Android releases easier, but the workflow that you are using for submitting Android builds right now will also continue to be supported.

## How Updates Work

New builds (i.e. a build that we are submitting to the App Store) will not immediately have any OTA updates available for it. In that case, the JS bundle included in the .ipa/.aab will be used by the client. Whenever an OTA update is released however, the .ipa/.aab's JS bundle will not be used in favor of the JS bundle downloaded from our updates server.

In the event that the downloaded JS bundle causes any issues, the client will fall back to using the .ipa/.aab's JS bundle.

We use `expo-fingerprint` in the GitHub action as well to prevent any accidental deployments that are not compatible with the current runtime version. I.e., if we updated a package that uses native code such as `react-navigation`, there's a chance that the JS code used by `react-navigation` would not be compatible with the native code the library uses. Therefore, we will cancel and fail these deployments.


.github/workflows/build-submit-ios.yml
```diff
@@ -69,7 +68,7 @@ jobs:
           echo "${{ secrets.GOOGLE_SERVICES_TOKEN }}" > google-services.json
 
       - name: 🏗️ EAS Build
-        run: yarn use-build-number eas build -p ios --profile production --local --output build.ipa --non-interactive
+        run: yarn use-build-number eas build -p ios --profile ${{ inputs.profile || 'testflight' }} --local --output build.ipa --non-interactive
```
haileyok, line 71: the `--profile` flag here determines what `EXPO_PUBLIC_ENV` gets set to.

eas.json
```diff
@@ -16,14 +16,20 @@
...
       "distribution": "internal",
-      "channel": "preview",
+      "channel": "production",
       "ios": {
         "resourceClass": "large"
+      },
+      "env": {
+        "EXPO_PUBLIC_ENV": "production"
       }
     },
```
estrattonbailey, line 34: I think this `preview` profile was only added for APK builds (happens bc of `distribution: 'internal'`). Could maybe remove if we can build APKs with the `testflight` profile.

estrattonbailey (approved):
LGTM! So pumped for this 🔥 

As you mentioned in Slack, let's do another quick look at Android too

eas.json
haileyok: Ah interesting. I think what I'll do for now then is set this one to use `testflight` channel, since we'll still need to do Android builds locally until we figure out the Google Play alpha builds stuff.

PR #3296, opened 2024-03-20
Mods UI fixes

estrattonbailey:
This PR:
- tweaks ReportDialog styles so that button states aren't cut off by Android — just made these all consistently a "card", and retained interaction styles
- consolidated our label prefs components into a single "macro" component, along with exports for the specific usages we have on the Moderation and Labeler Profile screens
  - Should be no visible or functional changes

CleanShot 2024-03-20 at 15 04 37@2x
CleanShot 2024-03-20 at 15 07 25@2x
CleanShot 2024-03-20 at 15 07 13@2x
CleanShot 2024-03-20 at 15 07 17@2x

CleanShot 2024-03-20 at 15 36 24@2x
CleanShot 2024-03-20 at 15 37 34@2x
CleanShot 2024-03-20 at 15 36 32@2x


pfrazee (approved):
Yep looks good

PR #3297, opened 2024-03-20
Fix problems with `BottomSheet` and the report dialog

haileyok:
## Why

Currently, when selecting the text input box in a report dialog, the input gets hidden behind the keyboard. This isn't great, as it requires the user to scroll down after selecting the box. It also messes with our "swipe to dismiss" gesture, since we have to stop using the `on-drag` keyboard dismiss mode.

## Solution

- Migrate to using Discord's fork of `@gorhom/bottom-sheet` found at https://github.com/discord/react-native-bottom-sheet. I've forked this again over to https://github.com/bluesky-social/react-native-bottom-sheet for us to use in our `package.json`.
- Migrating to this fork fixes our keyboard dismiss woes, so we revert some changes that ran a `Keyboard.dismiss()` for `onTouchDown` callbacks.
- Create a subscription to the `keyboardDidShow` event to call `scrollToEnd()` on the `ScrollView` whenever the keyboard appears.

## Future

Using React Native's `keyboardDidShow` event isn't the optimal solution for iOS. It does not fire until the entire keyboard shows. There are some libraries that handle this, however as @pfrazee predicted, they do not play very well with the bottom sheet. This might be something worth investigating in the future if we feel like giving it some time.

https://github.com/kirillzyusko/react-native-keyboard-controller is one option that uses Reanimated and provides us with an `onMove` callback. This could work as a solution, however it comes with new difficulties on Android. This will require a deeper dive to figure out what our best option is on the two different platforms.

https://github.com/bluesky-social/social-app/assets/153161762/27152bba-f2f2-4d25-a0c5-5d7d00ff0d0a

https://github.com/bluesky-social/social-app/assets/153161762/db136835-3906-419d-8802-54df7962f4ad


pfrazee (approved):
Okay looks good to me! tested on ios and android

PR #3303, opened 2024-03-21
Support NicoVideo inline play

haoyayoi:
Support nicovideo.jp in embed target. Autoplay not support.

Target Host:
- www.nicovideo.jp
- nico.ms

pr_preview_1
pr_preview_2
pr_preview_3


haileyok:
@haoyayoi Thanks for cleaning those up, I'll review it here later today probably.

haileyok:
Screenshot 2024-03-21 at 10 48 38 AM

Hi! I just merged in main and added a few small changes. This is great! One question about the image above. Have you tested this on web to see if it works for you? I can't find any good documentation on their embed API so not sure if that's something that would be fixed on a non-localhost server or if we'd need to host an embed frame for this.

Thanks!

haoyayoi:
I tried this on html like youtube(bskyweb/static/iframe/youtube.html), reference https://qiita.com/fanta_/items/13e0819e0c1db111186c. But blocked web and ios 
 on localhost. So, change to react-native-webview. 

Nicovideo chief developer has account in Bluesky, it might be a good idea to check.

thank you

haoyayoi:
The cause of the error is probably CORS.

haileyok:
Hey! Do you know the handle of the Nicovideo dev? I'll send them a mention to see if we can clear this up.

Issue #2308, opened 2023-12-26
[APP-929] Thread screen dimming on android


pfrazee:
Got multiple reports of this. Description from a user:

> ok I did some tinkering after it happened again just now. i can reliably reproduce it by going to someone's profile page, [@juicysteak117.gay](https://bsky.app/profile/did:plc:ywbm3iywnhzep3ckt6efhoh7)'s for testing as is tradition, and pressing a post that has media and is quote posted. pressing the quote posted media dims the screen

|Normal|Dimmed|
|-|-|
|![image](https://github.com/bluesky-social/social-app/assets/1270099/04902fec-3a9f-4f69-a3da-4cb61d14bd17)|![image](https://github.com/bluesky-social/social-app/assets/1270099/9e6a3aa5-51df-410a-b8ff-6d82a17b4fba)|

This looks like the lightbox dimming mask applying for some reason?

<sub>[APP-929](https://linear.app/blueskyweb/issue/APP-929/thread-screen-dimming-on-android)</sub>

MariusVatasoiu:
I also have a recording with this bug, besides other bugs.

https://github.com/bluesky-social/social-app/assets/18071013/cec58bbc-55f3-40ae-9a20-635929c2e265

PR #3305, opened 2024-03-21
Upgrade `react-navigation`, remove patch, use transition animations on Android

haileyok:
We removed Android transition animations in https://github.com/bluesky-social/social-app/pull/2352 because of https://github.com/bluesky-social/social-app/issues/2308.

This reverts that change and upgrades `react-navigation` to the latest stable release. It also removes the patch, since that fix was merged in and was released in 6.1.11 https://github.com/react-navigation/react-navigation/releases/tag/%40react-navigation%2Fnative%406.1.11.

## Testing

First, I checked out https://github.com/bluesky-social/social-app/commit/705f9b61efebe8ca0d044f1a53586b6fe4614195 which is two commits prior to the fix in #2352. I am reliably able to reproduce the issue with this change using the reproduction steps in #2308.

Next, I checked this branch out. Following the same reproduction steps does not reproduce the issue. Below are a few theories:

- There are a few animation fixes in `react-navigation` throughout the versions between our current and latest 6.X release. None specifically call out this issue, but they could be related.
- Our old `TouchableHighlight` animation was _very_ janky in that old commit. The color (dimness) of the text when the post was pressed in was the _same color_ as the next screen's text. It's possible that some combination of issues between react-navigation and the `TouchableHighlight` was causing an issue.


mozzius:
There is also the `ios` animation on android, we could use that?

gaearon:
merged via https://github.com/bluesky-social/social-app/pull/4216 + https://github.com/bluesky-social/social-app/pull/4207

PR #3313, opened 2024-03-21
Specify package manager in package.json

mozzius:
I'm not sure if it's yarn or node, but something keep trying to add this field automatically. I think its probably good to have anyway, so why not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "engines": {
     "node": ">=18"
   },
+  "packageManager": "yarn@1.22.19",
   "scripts": {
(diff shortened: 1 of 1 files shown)


haileyok (approved):

haileyok:
@mozzius Does this warn you if your yarn version is out of date?

mozzius:
PNPM does, I'm unsure if Yarn will

mozzius:
Ah yeah pretty sure it will

hichemfantar:
corepack adds this field automatically. corepack is like a `package manager` manager

PR #3314, opened 2024-03-21
Fix dialogs state references, fixes Android back button

estrattonbailey:


src/state/dialogs/index.tsx
```diff
@@ -74,15 +78,11 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
   const context = React.useMemo<IDialogContext>(
     () => ({
-      activeDialogs: {
-        current: new Map(),
-      },
-      openDialogs: {
-        current: new Set(),
-      },
+      activeDialogs,
+      openDialogs,
```
estrattonbailey, line 82: We were recreating these references each time 🤦

src/state/dialogs/index.tsx
```diff
@@ -53,7 +54,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   >('auto')
 
   const closeAllDialogs = React.useCallback(() => {
-    activeDialogs.current.forEach(dialog => dialog.current.close())
+    openDialogs.current.forEach(id => {
+      const dialog = activeDialogs.current.get(id)
+      if (dialog) dialog.current.close()
+    })
```
estrattonbailey, line 60: When this was originally written we didn't have a reference to ONLY the open dialogs. Now we do, so can close only those. `id` is the same. Test using the Storybook components.

PR #3316, opened 2024-03-21
Fix android splash jump

estrattonbailey:
Merged to a branch to fix conflict and format, which evidently wasn't really needed. Already tested on Android, no change on iOS. This one is ready to go.


haileyok (approved):
Yea LGTM. Thank you @edo-bari-ikutsu

PR #3327, opened 2024-03-21
Use ALF for the account quick switch dialog

mozzius:
This takes the new "choose account" thing from the login flow, moves it to it's own component, and then refreshes the account quick switch modal to be this new `AccountList` in an ALF dialog.

This means that you now have the ability to add a new account from this dialog, however the ability to sign out here has been removed - I doubt that was a common thing to do here, the settings menu is a much more sensible place for that stuff.

Screenshot 2024-03-21 at 22 27 14


pfrazee (approved):
Tested on all platforms by signing out and also by using the switcher. Works well.

PR #3330, opened 2024-03-21
fix android back after closing composer

haileyok:
## Why

See https://bsky.app/profile/mary.my.id/post/3koagxpdim72g

Because we're using two modal systems at once, we're running into problems here when we use a new dialog inside of the composer for discarding.

- Type some text into the composer
- Press "Cancel"
- Dialog presents asking if you want to discard
- Can no longer use the back button on Android

The prompt's confirm `onPress` calls both the `onClose()` for the composer and the `close()` for the dialog at the same time. The `close()` call though does not complete however and we end up with some ghost navigation state.

Instead, in this case we should use the callback parameter of `control.close(cb)` to close the composer _after_ closing the dialog.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1,5 +1,4 @@
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react'
-import {observer} from 'mobx-react-lite'
 import {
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Nice one

PR #3321, opened 2024-03-21
Replace queryClient

estrattonbailey:

PR #3333, opened 2024-03-22
Scope query client per DID

gaearon:
## What

Inspired by #3321, this is an attempt to have a guarantee that we never ever reuse the same cache instance (and the associated tree) between different DIDs. If we have that guarantee then we also should not have a need to `clear` the cache or `resetQueries` or any of the related things since we should be able to expect a cache instance to eventually get GC'd.

So I removed those `clear` calls.

## Test Plan

Switched between users. I think it works (not seeing cached feed etc). We should check if it fixes the issue https://github.com/bluesky-social/social-app/pull/3254 was set out to solve. I haven't done extensive testing here.

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -77,7 +77,7 @@ function InnerApp() {
           <React.Fragment
             // Resets the entire tree below when it changes:
             key={currentAccount?.did}>
-            <QueryProvider>
+            <QueryProvider currentDid={currentAccount?.did}>
               <PushNotificationsListener>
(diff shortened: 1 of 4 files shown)


estrattonbailey (approved):

pfrazee (approved):
Tested this on iOS and Android sims and on Web. Switched between accounts multiple times, and tried logged out.

Seems to work great. Didn't run into any bugs, never saw the incorrect feeds in home screen, and seemed to perform a little better.

PR #3334, opened 2024-03-22
Simplify list logic further to prevent misuse

haileyok:
@estrattonbailey had some good points on ensuring that we don't miss anything when creating lists. To that end, I have simplified this logic down further so that it can be easily reused anywhere while keeping it obvious as to what information we need to pass to either `ListMaybePlaceholder` or `ListFooter`.

### Notes

When using React Query, `isLoading` means that there is no cached data/placeholder data. `isLoading` will only be true whenever we have no data at all. See https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#isloading-previously-isinitialloading

`isFetching` is true whenever we *do or do not have data* and we are fetching new data. This is an important distinction to consider. In most of our cases, we do *not* want to show a full screen spinner if we have *some* data to display at first. For that reason, we use `isLoading` to determine whether or not to show a full screen spinner in most cases.

`isFetchingNextPage` is true whenever we *have data already and are getting the second or greater page*. For that reason, we use this to determine if we want to show the footer spinner when scrolling.

## `<ListMaybePlaceholder>`

This is the screen that should appear in the following states:

- Query is performing the initial load
- An error has occurred that we want to display to the user *before any data has loaded*
- There are not results from the query

This component should only be rendered while `data.length < 1`. In all other cases, we will display errors to the user in the `ListFooter` component.

Usage is like this:

```tsx
if (data.length < 1) {
	return (
		<ListMaybePlaceholder
			isLoading={isLoading} // Comes from our query
			isError={isError} // Comes from our query
			errorTitle="" // Optional. Title to display when `isError` is true
			errorMessage="" // Optional. Message to display when `isError` is true
			emptyTitle="" // Optional. Title to display when there are no results
		    emptyMessage="" // Optional. Message to display when there are no results.
		    onRetry={refetch} // Optional. Usually will be `refetch` from React Query. Leave `undefined` if you do not want to show a retry button.
		/>
	)
}
```

## `<List>`

Nothing changes for our usage of `List`. However, do note that we should use the `useInitialNumToRender()` hook when possible to provide the `FlatList` with the correct number to fill the user's screen with. This is a big performance booster, since by default we paint 10 items, regardless of their height. If we are rendering posts, we can easily render 2x the number of required posts, taking a good bit longer.

```tsx
const initialNumToRender = useInitialNumToRender() // Minimum height of an item defaults to 100. This is the height of posts and user profile cards (i.e. the liked by screen)
// OR
const initialNumToRender = useInitialNumToRender(300) // Pass your own value if there's a use case. Use the element inspector to figure out the smallest possible size for your component

return (
	<List
		// Other props
		initialNumToRender={initialNumToRender}
	/>
)
```

## `<ListFooter>`

The `ListFooter` is rendered at all times in a list, however it is not always visible to the user.

- The footer is a `View` with a height of `180px` by default. You can pass a custom value however if you want to adjust this. `180px` should be enough for any screen to be visible without the tab bar overlapping on native
- If we are fetching the next page, we display a spinner in the view.
- If there is an error *after we have data*, we display an error message in the view. Optionally display a retry button

Usage is like this

```tsx
<List
	// Other props
	listFooter={
		isFetchingNextPage={isFetchingNextPage} // Optional. comes from RQ
		error={cleanError(error)} // Optional. `error` comes from RQ
		onRetry={onRetry} // Optional. Shows a retry button if there is an error.
	}
/>
```

## Complete Example

This is a complete example of how we can use these components together. Note that there is no `isEmpty` prop on the `ListMaybePlaceholder` component. Because we are already conditionally rendering this component if the length of the data array is less than one and because we render the spinner or error over the empty error, we don't need to pass anything here.

In the occasional case where you *don't* want to show an error message at all if there is the request is successful and there is no error _but_ there is no data, you can use the `noEmpty` prop (or change the conditions for rendering).

```tsx
  if (follows.length < 1) {
    return (
      <ListMaybePlaceholder
        isLoading={isDidLoading || isFollowsLoading}
        isError={isError}
        emptyType="results"
        emptyMessage={
          isMe
            ? _(msg`You are not following anyone.`)
            : _(msg`This user isn't following anyone.`)
        }
        errorMessage={cleanError(resolveError || error)}
        onRetry={isError ? refetch : undefined}
      />
    )
  }

  return (
    <List
      data={follows}
      renderItem={renderItem}
      keyExtractor={keyExtractor}
      refreshing={isPTRing}
      onRefresh={onRefresh}
      onEndReached={onEndReached}
      ListHeaderComponent={<ListHeaderDesktop title={_(msg`Following`)} />}
      ListFooterComponent={
        <ListFooter
          isFetchingNextPage={isFetchingNextPage}
          error={cleanError(error)}
          onRetry={fetchNextPage}
        />
      }
      initialNumToRender={initialNumToRender}
      windowSize={11}
    />
  )
```


pfrazee (approved):
Clicked around like lists and post-threads on all platforms. Looks good.

PR #3336, opened 2024-03-22
Use ALF for the embed consent modal

mozzius:
Self-explanatory. Nothing fancy going on here, although if you don't like the web layout please say 😅 I can't decide if it's good or not

Screenshot 2024-03-22 at 12 17 02

Screenshot 2024-03-22 at 12 17 13


haileyok:
If we used `Prompt` instead of `Dialog` for this would it look any better? (And then put the buttons ontop of each other instead of side by side) No idea, just a thought.

mozzius:
@haileyok how's this?

Screenshot 2024-04-05 at 10 32 26

haileyok:
Screenshot 2024-04-05 at 4 17 54 PM

I slightly modified the spacing here. As long as you're good with this, everything else LGTM.

haileyok (approved):
Ran this through a couple of times on web and native. This looks good to me.

PR #3337, opened 2024-03-22
Move scrollview to `LoggedOutLayout` to fix scrolling on web

mozzius:
Currently, the signup/login scrollviews for the desktop layout don't work. This normally doesn't show up, but on a smaller screen you can see it on the captcha step (see below). I'm not 100% sure what the exact cause is (something to do with height: 100% vs height: 100vh I think?) but even if it did work, it would be broken anyway because `LoggedOutView` adds horizontal padding to the content area, leaving the scrollbars inset which looks terrible. This PR adds an optional scrollview to `LoggedOutView`, and removes the two different scrollviews from both the sign up and log in flows. Also, the way that it was centred using flexbox didn't work with the scrollview, so I changed it to use vertical auto margins so that once the scrolling kicks in, it stops being centred.

Screenshot 2024-03-22 at 12 24 36

### DIFF
--- a/src/screens/Login/FormContainer.tsx
+++ b/src/screens/Login/FormContainer.tsx
@@ -1,13 +1,6 @@
 import React from 'react'
-import {
-  ScrollView,
-  type StyleProp,
-  StyleSheet,
-  View,
-  type ViewStyle,
-} from 'react-native'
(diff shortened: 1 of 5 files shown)


haileyok (approved):
Nice find

PR #3346, opened 2024-03-23
Allow text selection of the post being replied to in composer

haileyok:
## Why

Users frequently want to be able to pull a piece of text out of the post they are replying to. By adding the `selectable` prop to the `ComposerReplyTo`'s `Text`, we can easily allow users to do this.

We also need to add a dummy callback to the `onLongPress` here so that the `Pressable` does not interfere with the native press and hold to select.

### DIFF
--- a/src/view/com/composer/ComposerReplyTo.tsx
+++ b/src/view/com/composer/ComposerReplyTo.tsx
@@ -1,21 +1,22 @@
 import React from 'react'
 import {LayoutAnimation, Pressable, StyleSheet, View} from 'react-native'
 import {Image} from 'expo-image'
-import {useLingui} from '@lingui/react'
-import {msg} from '@lingui/macro'
 import {
   AppBskyEmbedImages,
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
   AppBskyFeedPost,
 } from '@atproto/api'
-import {ComposerOptsPostRef} from 'state/shell/composer'
+import {msg} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+
 import {usePalette} from 'lib/hooks/usePalette'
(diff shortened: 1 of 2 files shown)


src/view/com/composer/ComposerReplyTo.tsx
```diff
@@ -100,7 +110,8 @@ export function ComposerReplyTo({replyTo}: {replyTo: ComposerOptsPostRef}) {
             <Text
               type="post-text"
               style={pal.text}
-              numberOfLines={!showFull ? 6 : undefined}>
+              numberOfLines={!showFull ? 6 : undefined}
+              selectable>
```
mozzius, line 114: Should it be a `selectable uitextview`? Would that cause issues?

haileyok: This `<Text>` does that underneath in https://github.com/bluesky-social/social-app/blob/67fca733faafdc54f85846ca0f0bbf0f4f174d13/src/view/com/util/text/Text.tsx#L35

(Reminds me though that we can also merge this in https://github.com/bluesky-social/social-app/pull/3090, has a few fixes that came up from other people using it)

mozzius: Ah my bad, should've checked that :)

pfrazee:
Tested on iOS and Android, works well.

But I think the letter spacing changed on iOS? or maybe the font weight or the aliasing or something?

|Reply in composer|In feed|
|-|-|
|CleanShot 2024-04-03 at 19 37 19@2x|CleanShot 2024-04-03 at 19 38 00@2x|

haileyok:
Oh interesting. Looks like it's that way in prod as well. Let me see if I can just hit that real quick here.

PR #3348, opened 2024-03-24
Show profile icon on bottom bar of web

edo-bari-ikutsu:
On current version of web, when width of browser is narrow, bottom bar is shown as follows: 

<Light, user profile not selected>
current_light_1

<Light, user profile selected>
current_light_2

<Dark, user profile not selected>
current_dark_1

<Dark, user profile selected>
current_dark_2

User profile icon is not shown on the bottom bar on the web version although shown on the iOS / Android version. Thus user cannot recognize current account by the profile icon.

I have implemented that user profile is shown on the web version same as iOS / Android version as follows:

<Light, user profile not selected>
pr_light1

<Light, user profile selected>
pr_light2

<Dark, user profile not selected>
pr_dark1

<Dark, user profile selected>
pr_dark2

Along with this fix, I have changed the border width when a user profile is selected from 1px to 2px. This is because border do not display correctly in lower resolution environments (for example, low-end smartphones and 96dpi PCs). This change has been applied both web (`BottomBarWeb.tsx`) and app (`BottomBar.tsx`) version of bottom bar.

[Note] Please review and confirm this modification based on these points:

+ I have confirmed this modification on Web (PC), Android (real device) and iOS (simulator). I cannot confirm on real iOS device because I do not own it. 
+ I have confirmed by using my personal Bluesky account, but I cannot confirm by using labeler account.

### DIFF
--- a/src/view/shell/bottom-bar/BottomBar.tsx
+++ b/src/view/shell/bottom-bar/BottomBar.tsx
@@ -1,42 +1,43 @@
 import React, {ComponentProps} from 'react'
 import {GestureResponderEvent, TouchableOpacity, View} from 'react-native'
 import Animated from 'react-native-reanimated'
-import {StackActions} from '@react-navigation/native'
-import {BottomTabBarProps} from '@react-navigation/bottom-tabs'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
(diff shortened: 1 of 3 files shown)


edo-bari-ikutsu:
I have recognized that fine tune is needed. After applying it, I will post new pull request.

PR #2483, opened 2024-01-11
Reorder the locales alphabetically

quiple:
I reordered the locales alphabetically.

Other fixes include:
* https://github.com/bluesky-social/social-app/blob/c75771047cb49a2cd529f72a2756a456c843a730/lingui.config.js#L3 There were two `es`, so I removed one.
* https://github.com/bluesky-social/social-app/blob/c75771047cb49a2cd529f72a2756a456c843a730/src/locale/languages.ts#L27-L37 Corrected `Indonesian` to `Bahasa Indonesia`. (since they are all written in their own language)

Thank you!

PR #3350, opened 2024-03-24
Alphabetise languages by language code

surfdude29:
This PR alphabetises the list of languages according to the two-letter ISO code (keeping English first, or last, as appropriate).


pfrazee:
Any reason you feel like we should do this?

surfdude29:
> Any reason you feel like we should do this?

The only reordering that is visible to end-users afaik is in `languages.ts`, where Catalan, Italian and Turkish are put in more logical places than at present. Reordering by language code seemed to make more sense to me than the current order, I guess :)

https://github.com/bluesky-social/social-app/blob/a1cbc9a8139567efed87ecd4e5eac45a314c7e05/src/locale/languages.ts#L32-L49

And as was done in #2483 I thought it made sense to reorder the locales in places where it isn't visible to end-users too.

pfrazee (approved):
Okay! Lgtm, thanks!

Issue #3349, opened 2024-03-24
Clicking in search box selects entire text


groupboard:
**Describe the bug**

When you click in the search box, it selects the entire text rather than simply placing the cursor where you click. It seems that it deliberately overrides the default input behaviour for no apparent reason, and it's really annoying.

**To Reproduce**

1. Go to https://bsky.app/search?q=bob+smith
2. Click somewhere inside the "bob smith" text

**Expected behavior**

Cursor should move to where you click, and nothing should be selected

**Screenshots**

**Details**

 - Platform:  macOS web, on Safari, Firefox and Chrome
 - Platform version:
 - App version:

**Additional context**

PR #3353, opened 2024-03-25
Clicking in search box selects entire text #3349

shanuv000:
clicking inside the search box should no longer select the entire text, and it should behave as expected, placing the cursor at the clicked position.


src/view/com/util/forms/SearchInput.tsx
```diff
@@ -66,7 +66,7 @@ export const SearchInput = React.forwardRef<SearchInputRef, Props>(
           ref={textInput}
           placeholder={_(msg`Search`)}
           placeholderTextColor={pal.colors.textLight}
-          selectTextOnFocus
+          // selectTextOnFocus
```
shanuv0002, line 69: Looks good

PR #3363, opened 2024-03-29
disable inline predictions to prevent ios composer jank

haileyok:
## Why

Inline predictions are causing some pretty bad jank with the post composer. This patch disables those suggestions. This is fairly common on other apps, so disabling it should not cause any issues for users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bsky.app",
-  "version": "1.75.0",
+  "version": "1.76.0",
   "private": true,
   "engines": {
     "node": ">=18"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bsky.app",
-  "version": "1.76.0",
+  "version": "1.75.0",
   "private": true,
   "engines": {
     "node": ">=18"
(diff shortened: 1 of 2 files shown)


mozzius (approved):
works in the TF, LGTM 👍

gaearon:
This doesn't seem like it gets at the root of the problem though — there is an ongoing issue in the composer where if you type fast enough, the typing happens at the wrong place. Do you think you could have a look at fixing that? And then whatever the fix for that is, it may likely fix the problem this PR works around too. I'm guessing the `setTimeout` calls are a good place to start with.

gaearon:
Hm, looks like I can't repro the fast typing problem, so maybe that's a bit hard to chase

haileyok:
I think the better option here might be to remove the current paste component and create another one using an expo module. The current one is not compatible with Fabric so we will need to do this anyway. I've looked at both the android and ios code that the current one uses and it doesn't seem to be a massive undertaking. 

But yes, do agree we need a more complete solution. I have something in a local branch right now that tries to address both the cursor problem and the flickering text color when adding mentions or links without affecting the default iOS behavior which is what i'd like to do optimally.

gaearon:
Yeah this sounds ideal.

haileyok:
by the way, this should fix the problem with the fast typing, which is why I did this. I haven't been able to reproduce since merging. Are you still experiencing it in the TestFlight?

gaearon:
Yeah I haven't been able to repro either! Didn't realize this PR fixes that too. That makes more sense now, thanks. We should maybe report upstream.

haileyok:
Yea, that might be a good idea. At the very least I was considering opening a small PR for a prop to turn off this iOS `UITextView` property (it should probably be configurable, there are legitimate reasons to turn it off), but from other users' experiences in other RN apps, seems like there's a chance this might be a bug in general.

PR #3364, opened 2024-03-30
Apply background color to entire scrollview in delete account modal

mozzius:
Before:

Screenshot 2024-03-30 at 16 27 29

After:

Screenshot 2024-03-30 at 16 27 09


haileyok (approved):
rad

PR #3368, opened 2024-03-31
[ipad support] Fix header overlap issue

mozzius:
simulator_screenshot_87EC86B3-6FC8-4227-909E-5A753D1B42CB

PR #3369, opened 2024-03-31
[iPad support] Fix header overlap issue

mozzius:
Supercedes #3368

Fixes the header overlap issue, and also fixes the way we do scrollviews in the logged out layout

simulator_screenshot_BFEB8741-4D62-4E8F-94CF-D8E4E694CC0F
Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2024-03-31 at 17 00 34


src/view/com/util/layouts/LoggedOutLayout.tsx
```diff
@@ -73,9 +72,7 @@ export const LoggedOutLayout = ({
             contentContainerStyle={styles.scrollViewContentContainer}
             keyboardShouldPersistTaps="handled"
             keyboardDismissMode="on-drag">
-            <View style={[styles.contentWrapper, isWeb && a.my_auto]}>
-              {children}
-            </View>
+            <View style={[styles.contentWrapper]}>{children}</View>
```
mozzius, line 75: Should consider making this a KeyboardAvoidingView or something

haileyok: Agree on this. We actually should probably use one of the native libraries for this like the one you used in Graysky or the other popular one. `KeyboardAvoidingView` can be a bit janky in general especially on Android.

src/view/com/util/load-latest/LoadLatestBtn.tsx
```diff
@@ -43,7 +45,7 @@ export function LoadLatestBtn({
           (isTallViewport
             ? styles.loadLatestOutOfLine
             : styles.loadLatestInline),
-        isTablet && styles.loadLatestInline,
+        isTablet && !isNativeTablet && styles.loadLatestInline,
```
haileyok, line 48: Just curious how often we think this will happen? Could add a `isWebTablet` for this if it will happen a lot. Nbd if this is a rare one-off thing though.

haileyok (approved):
Everything here looks good! Clicked through the app, not seeing any obvious UI concerns other than just eventually wanting to keep posts a maximum width.

Oh, we also need to fix the image viewer to support screen rotation, but that can be a separate PR since it isn't iPad specific.

mozzius:
I'm able to just rotate the app in the simulator, I think iPad might be immune to rotation lock?

mozzius:
Oh dang ok. Better fix that due to the warning here: https://docs.expo.dev/versions/latest/sdk/screen-orientation/#

gaearon:
This PR seems to touch every file (?), can we undo that? I think it probably ran `eslint --fix` on the entire codebase instead of just the changed files?

haileyok:
Oh lord, is _that_ how it happened? Probably on me. Bet that webstorm did something goofy, will fix here in a few.

haileyok:
Ah, the mass-rewrite happened in this commit https://github.com/bluesky-social/social-app/pull/3369/commits/24c9539af162a4f8667e51da096b690819ef4160

haileyok:
I was able to easily roll back all of the changes made to the tests. These changes in `src/*` are a lot gnarlier though. Happy to help work through that next week, but don't want to mess anything up here right now re: the PR's changes themselves.

haileyok:
Okay strike that, I think I was able to work through all of them. I'm not going to squash the commits right now incase a specific batch messed something up, but this should be good now.

src/view/com/util/load-latest/LoadLatestBtn.tsx
mozzius: That's probably a good idea, yeah

mozzius:
I'll keep a closer eye out in future. Thanks for fixing this 👍

haileyok:
No problem. FWIW Webstorm did it to me in another branch I was working on. Looks like what happens (at least for me) is if I merge in main and if there are any type errors (i.e. a new package that I have not installed yet), then the git hooks _prevent_ the merge and somehow the linter gets ran on everything (I think it's the on-save prettier that's causing it for me) :(

PR #3340, opened 2024-03-22
Wrap Button children in an error boundary

estrattonbailey:

PR #3372, opened 2024-04-02
Remove Button error boundary

gaearon:
We've added an error boundary _inside_ each button in https://github.com/bluesky-social/social-app/pull/3340 in case there are some existing places in the app where we accidentally render raw text inside (not wrapped in `<ButtonText>`) which would lead to a crash. In that case, the error boundary would display "ERROR" inside the button instead. 

Let's remove this.

## Why it's ok to remove now

The release including this fix has largely propagated:

Screenshot 2024-04-01 at 19 02 57

We are not observing any related logs in Sentry:

Screenshot 2024-04-01 at 19 03 34

## Why we shouldn't re-add this

It's confusing if a button just says `ERROR`. You might even do a destructive action (e.g. deletion) because you're not aware what the button actually does. We shouldn't be showing buttons without real labels:

Screenshot 2024-04-01 at 19 35 53

Screenshot 2024-04-01 at 19 37 06

In general, defensive coding can be helpful for working around issues in short term, but it makes it harder to understand where the points of failure are, and gradually leads to more bugs being introduced unknowingly.

And of course this doesn't catch issues with any other components so it's a pretty narrow fix.

Instead, there are a few things we could do:

- Add sensible error boundaries for major UI elements. E.g. "something went wrong" around big screen areas. This requires some design consideration because they actually need to look decent in the UI.
- (Maybe:) Stop using `<Trans>` in favor of `_(msg``)`. The benefit of this is that strings will be typed as strings, instead of being sometimes strings and sometimes React elements (as it happens now). This would let us start enforcing that some components don't take strings as children. It's still not protecting against extracting `<Foo>` which then returns raw text, but it's a decent start which should catch most cases.
- We could also use linting more heavily, for example we could adapt [this rule](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/lib/rules/no-raw-text.js) or [this rule](https://www.npmjs.com/package/eslint-plugin-lingui#no-unlocalized-strings) to our needs. There may be some simple middle ground where we don't try to catch all cases but instead make our own rule that's specific to some convention that we use in our design components.

I think none of this should be blocking the removal of the error boundary though.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -14,7 +14,6 @@ import {
 import LinearGradient from 'react-native-linear-gradient'
 import {Trans} from '@lingui/macro'
 
-import {logger} from '#/logger'
 import {android, atoms as a, flatten, tokens, useTheme} from '#/alf'
 import {Props as SVGIconProps} from '#/components/icons/common'
 import {normalizeTextStyles} from '#/components/Typography'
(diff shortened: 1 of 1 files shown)


haileyok:
Yea, this should be removable now. We should technically be getting red screen errors for these as well, so we do need to pay closer attention in general to testing new changes. Thanks for the dig and info on alternative solutions!

Regarding possible solutions, I think I most closely agree with the second option of using `_(msg``)`. This seems like the safeset option imo _in addition_ to using the lint rules. The lint rule by itself here I don't think would have caught the error (iirc, it's been a bit since we last looked at this).

Also, regarding an error boundary, do agree we should be using them - and I believe we are. In this case I think that because the modals are getting portaled out (to the root? Not sure myself) they end up being outside of our normal boundary. We might benefit from placing a boundary around the `Dialog.Outer`?

Leaving this here for @estrattonbailey's thoughts as well.

estrattonbailey (approved):
Yep lgtm!

PR #3373, opened 2024-04-02
Separate ESLint and Prettier

gaearon:
Our current setup is running Prettier *as a part of* ESLint run because of https://github.com/facebook/react-native/pull/43756. This wasn't super intentional and is just a consequence of us using the default configs which have this arcane setting turned on.

The recommended modern approach is to treat Prettier and ESLint as separate unrelated tools.
This has a few benefits:

- Editors that support Prettier will no longer run Prettier twice on every save (currently they do because we lint *in addition* to formatting, but linting also *includes* formatting).
- Formatting nits will no longer show up in IDEs as errors or warnings (currently they do because, well, for the same reason).
- This may also prevent some cases where lint has to run more than once (due to cycles).

See Prettier's [Integrating with Linters](https://prettier.io/docs/en/integrating-with-linters) and [Josh Goldberg's blog post](https://www.joshuakgoldberg.com/blog/you-probably-dont-need-eslint-config-prettier-or-eslint-plugin-prettier/) for more background.

## Okay, so what are we doing?

- First, I get rid of `eslint-plugin-prettier` by explicitly disabling the `prettier/prettier` rule. We'll be able to remove this special case when https://github.com/facebook/react-native/pull/43756 ships.
- Next, I add explicit `yarn prettier --check .` to CI.
- I tweak the `.eslintignore` to select the stuff we want (code in `src`) and exclude the rest. Previously it was kind of selectively ignoring a bunch of stuff but it was _also_ being called outside just for `src` so there was no source of truth.
- I extend the lint staged commands to run Prettier after ESLint. Note I removed the `yarn` indirection there to cut down time on node init. Otherwise we have a chain of Node processes which is slower than we want.
- Finally, added `--cache` to both commands as [suggested by Christoph Nakazawa](https://cpojer.net/posts/fastest-frontend-tooling-in-2022), it seems to be working decent in my testing.

If I didn't mess it up, this should give us roughly same behavior as before but faster and less annoying.


haileyok:
Amazing! Pulled this down and I'm getting the expected behavior in Webstorm now. Prettier only runs on save if I explicitly tell Webstorm to do so, rather than it running all the time w/ eslint. Seriously appreciate you looking into this, it was killing performance for me!

Will be curious to see over time if the `--cache` flag improves performance as well. Never knew about this one. Code in `./modules/*/src` is also linted correctly so `.eslintignore` looks solid.

haileyok (approved):

gaearon:
OK let's give this a go.

gaearon:
#codingonaplane

estrattonbailey:
Nice one!

PR #3375, opened 2024-04-02
Use Prettier at the root

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3373. I think this makes more sense — we want to format all JS-ey stuff across the project. Including build scripts etc. I only added explicit ignores for folders with generated code.


haileyok (approved):
Ah yea, useful for the various `*.config.js` files at the root. Don't usually touch them, but the `app.config.js` does get changed occasionally.

PR #3383, opened 2024-04-03
Add webpack analyzer, PR comments for webpack bundle size

haileyok:
This PR adds a GitHub action to reply to PRs with a webpack bundle size diff comment (as seen below, yikes!). It also add a `yarn open-analyzer` command for local viewing of the bundle in a browser.

The PR action will compare the current PR commit against the latest commit in main. Resulting `stats-base.json` will be cached the first time it is created to improve the speed at which the action runs.


.github/workflows/pull-request-commit.yml
```diff
@@ -14,9 +14,109 @@ concurrency:
...
+        run: yarn install
+
+      - name: 🔦 Generate stats file for PR
+        run: |
+          yarn generate-webpack-stats-file
+          mv stats.json ../stats-new.json
+
+      - name: Get the base commit
+        id: base-commit
+        run: echo base-commit=$(git show-ref -s $(git describe --abbrev=0 --tags)) >> "$GITHUB_OUTPUT"
```
haileyok, line 44: Thinking about this again. This uses the latest _tag_ as the base commit to compare to. I think this makes the most sense, but maybe it would be better to just use the latest commit to main. If we merge in 4 changes that reduce the bundle size by 300kb, then create another commit that increases the size by 10KB, we'd still see a reduction of 290kb even though we actually increased the size in this PR.

haileyok: Changing it to the latest main commit for now, but if we do end up wanting to use a tag, this is what I used before: `        run: echo base-commit=$(git show-ref -s $(git describe --abbrev=0 --tags)) >> "$GITHUB_OUTPUT"`.

Or if we have a better thought on how to do it, totally can!

haileyok: Okay, I changed this to use the latest commit from the PR's base branch. Because we want to check _only_ the changes made in the PR, we will merge in the base commit to the PR's branch before generating the stats file. See comment below.

estrattonbailey (approved):
LGTM

pfrazee (approved):
SGTM!

PR #1569, opened 2023-09-28
iPad support

pfrazee:
Start of iPad support. Unfortunately this is slightly less trivial than expected, though still doable. The web versions of the various components support the tablet form (_decently_) but the native versions all currently expect the mobile form-factor. Updating things to work correctly with the native tablet may be a task best left for a future refactor, such as when we introduce the design system.

How this looks currently:

CleanShot 2023-09-28 at 14 09 40@2x

This also fixes the authentication flow:

Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2024-03-31 at 17 00 34

# Test plan

This PR is the absolute basics of iPad support - basically just enabling it and making sure it's not entirely broken.

1. does it boot up into iPad
2. does the auth flow work
3. is the auth flow still working on other platforms

Note that the header overlaps with the feed - this is fixed in #3465

PR #3385, opened 2024-04-03
[iPad support] Custom tab navigator for native tablet, with side nav

mozzius:
Stacked on:
#1569 
#3369 

Goal: get the web tablet layout working on native tablet.

This PR:

- Renames `DesktopSideNav` to `SideNav`
- Branches `TabNavigator` into `TabNavigatorMobile` or `TabNavigatorTablet`, depending on platform
- Turns the list, moderation, and settings screens into their own tab stacks on native tablet
- Created `TabNavigatorTablet` which adds the extra tab stacks and uses `SideNav` rather than `TabBar`

Not a huge fan of some of the ways it splits the navigator logic depending on platform. Suggestions on how to branch this better are welcome.

image

### DIFF
--- a/src/view/com/profile/ProfileSubpageHeader.tsx
+++ b/src/view/com/profile/ProfileSubpageHeader.tsx
@@ -13,7 +13,7 @@ import {sanitizeHandle} from 'lib/strings/handles'
 import {makeProfileLink} from 'lib/routes/links'
 import {NavigationProp} from 'lib/routes/types'
 import {BACK_HITSLOP} from 'lib/constants'
-import {isNative} from 'platform/detection'
+import {isNative, isNativeTablet} from 'platform/detection'
 import {useLightboxControls, ImagesLightbox} from '#/state/lightbox'
 import {useLingui} from '@lingui/react'
 import {Trans, msg} from '@lingui/macro'
@@ -71,9 +71,11 @@ export function ProfileSubpageHeader({
     }
   }, [openLightbox, avatar])
 
+  const useMobileLayout = isMobile || isNativeTablet
+
   return (
     <CenteredView style={pal.view}>
-      {isMobile && (
+      {useMobileLayout && (
         <View
           style={[
             {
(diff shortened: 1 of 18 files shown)


haileyok:
Screenshot 2024-04-05 at 3 46 21 PM

Not sure what's causing the white line there between sidebar and content (might not matter if we decide to rm anyway?)

src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -72,9 +72,11 @@ export function ProfileSubpageHeader({
     }
   }, [openLightbox, avatar])
 
+  const useMobileLayout = isMobile || isNativeTablet
```
haileyok, line 75: Small nit, let's use `isMobileLayout` for this just so it's more clear that it's actually a bool and not a hook. Mostly a skill issue probably, but when I saw `useMobileLayout` below I went 🤔 looking for a hook, heh.

haileyok (approved):
As long as we can clean up that margin between the sidebar and content, this LGTM!

Like I said, we might be rming that anyway so might not be a _huge_ deal

src/view/com/profile/ProfileSubpageHeader.tsx
mozzius: Damn yeah good call, was literally thinking that while typing but I couldn't come up with a better name

2 hardest problems in computer science etc etc

mozzius:
Oh no, turns out non of the options in the hamburger menu work 😅

Could we rm the hamburger menu on tablet in this case? Since we have the full sidebar?

haileyok:
Oh actually you're doing that in the other PR anyway right? rm seems fine to me.

mozzius:
The sidebar is in this PR but yeah

haileyok:
As long as you are removing it in the other PR it's fine with me i think. I preferred the look of that anyway.

mozzius:
Good news - rm-ing the drawer lets me centre the butterfly!

Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-04-07 at 10 25 13

mozzius:
Seems like the drawer is not present on the web tablet either. So this is an uncontroversial change

PR #3386, opened 2024-04-03
[iPad support] add right nav

mozzius:
Stacked on:
#1569 
#3369 
#3385 

Xnapper-2024-04-03-15 24 23


mozzius:
Going to close this - I can't quite get the right nav to behave as it should, and our current right nav is not great anyway. Worth looking into in the future though

PR #3388, opened 2024-04-03
Italian Localization

Titianbeetle:
New strings translated for v.1.75

### DIFF
--- a/src/locale/locales/it/messages.po
+++ b/src/locale/locales/it/messages.po
@@ -36,7 +36,7 @@ msgstr "<0>{0}</0> following"
 
 #: src/screens/Profile/Header/Metrics.tsx:46
 msgid "<0>{following} </0><1>following</1>"
-msgstr "<0>{following} </0><1>folowing</1>"
+msgstr "<0>{following} </0><1>following</1>"
 
 #: src/view/com/auth/onboarding/RecommendedFeeds.tsx:30
 msgid "<0>Choose your</0><1>Recommended</1><2>Feeds</2>"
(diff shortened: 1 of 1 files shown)


pfrazee:
@Titianbeetle can you get a second italian speaker to review this? We always want a confirmation

marcomaroni:
I'm available to collaborate with @Titianbeetle to review the italian translation.

Titianbeetle:
@pfrazee I am in a bit of a pickle here! @marcomaroni came in as a second participant and did a few corrections (see [https://github.com/Titianbeetle/social-app/commit/71bdb7b806b806e8aab3013b097f2a28e9de69c4](url) ,which I have implemented. However, for reason beyond my jurisdiction, his account has been cancelled by GitHub (go figure!) so I am bit stuck as I cannot find any other Italian speaker available on this platform. Also, there is now an error in the check which wasn't there early and I have no idea what caused it. I will put out a call to see if someone else is willing to help and if not I may put an end to my stint as volunteer translator :(

surfdude29:
> Also, there is now an error in the check which wasn't there early and I have no idea what caused it.

I don't think you need to worry about the PR labeler /webpack-analyzer error, I got that too on my PR. I might be wrong on this but I think that is only applicable for Bluesky team members?

haileyok:
Heh yea, the labeler is a bit wonky when working with other people's PRs. I disabled it for external contributors now, so if you merge in main it should stop yelling at you.

pfrazee (approved):

PR #3393, opened 2024-04-03
[Session] Root

estrattonbailey:
The core issue being solved here is a race condition between the update of `currentAccount` and the `__globalAgent`. #3333 set us up so that changing accounts gives us a fresh query cache, which was the impetus behind `__globalAgent` in the first place (stale closures).

This PR builds on that by deriving `currentAccount` from the currently active `BskyAgent`, removing the need to manage updates to two objects (`currentAccount` and `__globalAgent`) that have the same lifetime.

In the process, I also split out state handling into separate `useState` calls to help avoid loops and isolate updates.

- derive `currentAccount` from `BskyAgent.session` 4482cbf1436113e1a34c4903bb26635417d1a8cd
- replace mutation of this global state with a call to refresh the session and re-derive ff36564c9b9fc5776de5936e991f33487c4ceafe
- replace usages of previous mutation of global state 80b562cfacf703d2807e9c81f67e187ad43c6017
- refactored `persistSession` handler for more clarity 634995db38a2f7a3d6b4871b66cafb8480d86877

With this PR, we should now reference the active `BskyAgent` via `{ agent } = useSession()` and drop the `__globalAgent`.

Review `state/session.ts` [without whitespace](https://github.com/bluesky-social/social-app/pull/3393/files?diff=split&w=1#diff-ba67025ff0027278031f772f88a9ed2f7fa4b12421d0a423ad74b4937aa2ca53).

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -1,22 +1,33 @@
 import React from 'react'
-import {BSKY_LABELER_DID, BskyAgent} from '@atproto/api'
-import {jwtDecode} from 'jwt-decode'
+import {BskyAgent} from '@atproto/api'
 
 import {track} from '#/lib/analytics/analytics'
 import {networkRetry} from '#/lib/async/retry'
-import {IS_TEST_USER} from '#/lib/constants'
 import {PUBLIC_BSKY_SERVICE} from '#/lib/constants'
-import {logEvent, LogEvents} from '#/lib/statsig/statsig'
-import {hasProp} from '#/lib/type-guards'
+import {logEvent} from '#/lib/statsig/statsig'
 import {logger} from '#/logger'
 import {isWeb} from '#/platform/detection'
 import * as persisted from '#/state/persisted'
+import {
+  SessionAccount,
+  SessionApiContext,
+  SessionStateContext,
+} from '#/state/session/types'
+import {
+  agentToSessionAccount,
+  configureModeration,
+  createAgentAndCreateAccount,
+  createAgentAndLogin,
+  isSessionExpired,
+  sessionAccountToAgentSession,
+} from '#/state/session/util'
 import {useLoggedOutViewControls} from '#/state/shell/logged-out'
 import {useCloseAllActiveElements} from '#/state/util'
 import * as Toast from '#/view/com/util/Toast'
 import {IS_DEV} from '#/env'
 import {emitSessionDropped} from '../events'
-import {readLabelers} from './agent-config'
+
+export type {CurrentAccount, SessionAccount} from '#/state/session/types'
 
 /**
  * Only used for the initial agent values in state and context. Replaced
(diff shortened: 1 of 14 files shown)


src/state/session/index.tsx
```diff
@@ -209,6 +207,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   const clearCurrentAccount = React.useCallback(() => {
     logger.warn(`session: clear current account`)
     __globalAgent = PUBLIC_BSKY_AGENT
+    BskyAgent.configure({appLabelers: [BSKY_LABELER_DID]})
```
estrattonbailey, line 210: This slipped in with modservices. I think this should be reset when the account is logged out.

src/state/session/index.tsx
```diff
@@ -276,20 +275,24 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         createPersistSessionHandler(
           account,
           ({expired, refreshedAccount}) => {
-            upsertAccount(refreshedAccount, expired)
+            if (expired) {
+              clearCurrentAccount()
+            } else {
+              setCurrentAccount(refreshedAccount)
+            }
```
estrattonbailey, line 282: All persist handlers should handle token expiration AND update e.g. refresh token is used to get a new access token

src/state/session/index.tsx
```diff
@@ -392,97 +408,34 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
-        refreshJwt: account.refreshJwt || '',
-        did: account.did,
-        handle: account.handle,
-        deactivated:
-          isSessionDeactivated(account.accessJwt) || account.deactivated,
-      }
+      // optimistic, we'll update this if we can't reuse or resume the session
+      await configureModeration(agent, account)
 
       if (canReusePrevSession) {
```
estrattonbailey, line 414: This block is just simplified.

If we know the session hasn't expired:
- set it on `agent.session`
- update `__globalAgent`
- update state, trigger app rerender

If the session is expired:
- check if it can be refreshed
  - if it succeeds, the persist callback will handle the rest
- if it fails, clear the current account

src/state/session/index.tsx
```diff
@@ -368,14 +373,25 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         persistSession: createPersistSessionHandler(
           account,
           ({expired, refreshedAccount}) => {
-            upsertAccount(refreshedAccount, expired)
+            if (expired) {
+              clearCurrentAccount()
+            } else {
+              __globalAgent = agent
+              setCurrentAccount(refreshedAccount)
```
estrattonbailey, line 380: This persist callback is special because it swaps the `__globalClient` along with `currentAccount` at the same time.

src/state/session/index.tsx
```diff
@@ -594,13 +547,13 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
             },
           })
 
-          initSession(session.currentAccount)
+          await initSession(session.currentAccount)
```
estrattonbailey, line 550: All other callsites wait for init to finish, so added this just for consistency

src/state/session/index.tsx
```diff
@@ -658,11 +563,14 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
       selectAccount,
-      updateCurrentAccount,
+      refreshSession,
       clearCurrentAccount,
     ],
   )
 
+  // as we migrate, continue to keep this updated
+  __globalAgent = agent
+
```
estrattonbailey, line 573: @gaearon special note of this

src/state/session/index.tsx
```diff
@@ -108,112 +114,129 @@ const ApiContext = React.createContext<ApiContext>({
...
-    },
-    [setState],
+  const [agent, setAgent] = React.useState<BskyAgent>(PUBLIC_BSKY_AGENT)
+  const [accounts, setAccounts] = React.useState<SessionAccount[]>(
+    persisted.get('session').accounts,
+  )
+  const [isInitialLoad, setIsInitialLoad] = React.useState(true)
+  const [isSwitchingAccounts, setIsSwitchingAccounts] = React.useState(false)
+  const currentAccount = React.useMemo(
+    () => agentToSessionAccount(agent),
```
gaearon, line 158: This assumes that once `agent` gets into state, its `.session` and nested fields (e.g. `.session.refreshJwt`) never changes. Can we make that assumption?

src/state/session/index.tsx
```diff
@@ -108,112 +114,129 @@ const ApiContext = React.createContext<ApiContext>({
...
-    currentAccount: undefined, // assume logged out to start
-  })
-
-  const setStateAndPersist = React.useCallback(
-    (fn: (prev: SessionState) => SessionState) => {
-      isDirty.current = true
-      setState(fn)
-    },
-    [setState],
+  const [agent, setAgent] = React.useState<BskyAgent>(PUBLIC_BSKY_AGENT)
```
gaearon, line 151: There's already `agent` in nested scopes. Let's rename this one to `currentAgent`.

src/state/session/index.tsx
gaearon: Alternatively, can rename those to `nextAgent`.

src/state/session/index.tsx
gaearon: Should we do a pass to use a hook first? And then replace the implementation.

src/state/session/index.tsx
estrattonbailey: Yeah good shout. Those fields are managed internally by `BskyAgent`. Example: when `accessJwt` expires, the agent automatically calls `refreshSession` which updates the `session` internally. Calls to the `currentAgent` in client code then automatically use the updated session.

This made me realize how important it is to ensure that fresh tokens get synced across tabs though, gonna triple check that...

gaearon: If `agent` is something that can be mutated, it isn't right to use it as a `useMemo` input. Since nothing will trigger re-evaluation when these fields change. Let's sync up tomorrow to chat more about this?

estrattonbailey: `setAgent` is called when switching accounts, so each account always has its own `agent`, and `currentAccount` can be derived from it. What I'm explaining above isn't something we need to handle directly in this case.

gaearon (commented):
chatted out of band about this, seems like a promising direction with a few things to fix

estrattonbailey:
Superseded by #3541 and stack

PR #3394, opened 2024-04-03
Show redbox in DEV on string <View> child for web

gaearon:
Suppose we add a string somewhere like this:

```js
<View>stuff</View>
```

On RN, this would be a crash, but it works on the web. So if you mostly develop on web, it's easy to miss.

RNW includes some basic protection against this with [this DEV-only `console.error` call](
https://github.com/necolas/react-native-web/blob/a0cd8ffba4ade1b5860164cb4ef89a652a68580f/packages/react-native-web/src/exports/View/index.js#L71-L79). Note it isn't as exhaustive as the real check RN needs because it will only catch direct `<View>` children. It's still better than nothing though.

The problem is, it's very easy to miss because we have a bunch of `console.error` console logspam on web. We should fix that logspam, but still, for this particular error we should just show the redbox in DEV on web. So we can't miss it.

This is a hack to do exactly that.

## Before

Screenshot 2024-04-04 at 00 11 10

## After

Screenshot 2024-04-04 at 00 18 16

Note you can see the component stack in the console.

## Test Plan

See above.

Doesn't affect native because in `.web.ts` file.

Doesn't affect prod because in `process.env.NODE_ENV !== 'production'` block.

## Alternatives

This could be an RNW patch instead. But this seems easier. We don't need this to be a strong guarantee, let's just update it if needed. In longer term we might want to override console anyway.

### DIFF
--- a/src/platform/polyfills.web.ts
+++ b/src/platform/polyfills.web.ts
@@ -6,3 +6,24 @@ findLast.shim()
 
 // @ts-ignore whatever typescript wants to complain about here, I dont care about -prf
 window.setImmediate = (cb: () => void) => setTimeout(cb, 0)
+
+if (process.env.NODE_ENV !== 'production') {
+  // In development, react-native-web's <View> tries to validate that
+  // text is wrapped into <Text>. It doesn't catch all cases but is useful.
+  // Unfortunately, it only does that via console.error so it's easy to miss.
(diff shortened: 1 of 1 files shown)


haileyok (approved):
LGTM! Never realized it didn't show a redbox for this on web.

PR #3398, opened 2024-04-04
Lint against strings without wrapping <Text>

gaearon:
This adds a lint rule enforcing that JSX text has an immediate `<Text>` parent. There are a few nuances:

- It skips over `<Trans>` when looking upwards for `<Text>`. So `<Text><Trans>text</Trans></Text>` is OK.
- Aside from `<Trans>`, the parent *has to* be immediate. I.e. `<Foo><Text>text</Text></Foo>` is OK but `<Text><Foo>text</Foo></Text>` is not.
- It treats any component ending with -`Text` as a valid parent. E.g. `<FooText>text</FooText>` is OK. This is useful for the design system.
  - Our components don't currently follow this scheme so I also added a list of exceptions. I do want to narrow them down though. The advantage of the suffix is that we'll be able to lint for it too — i.e. we'd be able to enforce that a component named `FooText` *must* return `Text`. This would close the remaining loophole.
- It treats any prop called `text` (or ending with `Text`) as an implied guarantee it's wrapped. So `<Foo bla={<Trans>hello</Trans>} />` is not OK but `<Foo blaText={<Trans>hello</Trans>} />` is OK.
  - Our components don't currently follow this scheme so I also added a list of exceptions.
- Note that we don't validate literals. So `<Foo>{'text'}</Foo>` will pass despite being not OK. Only JSX text is validated. This leaves much less space for false positives, and is probably fine because we have to wrap stuff in `<Trans>` anyway.

This gets us pretty far.

### Raw text in `<View>`

Screenshot 2024-04-04 at 05 04 48

### Raw text in a custom component

Screenshot 2024-04-04 at 05 04 56

### Raw text in a prop

Screenshot 2024-04-04 at 14 50 45


.eslintrc.js
```diff
@@ -13,12 +15,50 @@ module.exports = {
...
   ],
   rules: {
     // Temporary until https://github.com/facebook/react-native/pull/43756 gets into a release.
     'prettier/prettier': 0,
     'react/no-unescaped-entities': 0,
     'react-native/no-inline-styles': 0,
+    'bsky-internal/avoid-unwrapped-text': [
+      'error',
+      {
+        impliedTextComponents: [
```
gaearon, line 28: The goal is to get this list close to zero, maybe except H1..H6. For components that always return Text (e.g. links), I propose adding Text suffix and making the rule enforce that they actually return Text. Will need follow-up work.

package.json
```diff
@@ -234,6 +234,7 @@
     "babel-preset-expo": "^10.0.0",
     "detox": "^20.14.8",
     "eslint": "^8.19.0",
+    "eslint-plugin-bsky-internal": "link:./eslint",
```
gaearon, line 237: Don't forget to run yarn.

haileyok (approved):
Went through a bunch of test cases on this, works great. Really like the prop name idea as well.

PR #3400, opened 2024-04-04
Use ALF for post controls

mozzius:
Changes:

- New icons!
- New repost menu
  - Web: ALF `<Menu>`
  - Native: ALF `<Dialog>`
 - Hover states
 - Cleaned up sizes/padding/etc 

Screenshot 2024-04-04 at 20 28 35
Screenshot 2024-04-04 at 20 28 49
Screenshot 2024-04-04 at 20 29 01
Screenshot 2024-04-04 at 20 29 14


src/components/icons/Quote.tsx
estrattonbailey, line 1: Beautiful

src/view/com/util/post-ctrls/RepostButton.tsx
```diff
@@ -1,108 +1,129 @@
...
+                  onQuote()
+                }}
+                size="large"
+                variant="outline">
+                <ButtonIcon
+                  icon={Quote}
+                  size="lg"
+                  color={t.palette.primary_500}
+                />
+                <ButtonText style={a.text_xl}>{_(msg`Quote post`)}</ButtonText>
```
estrattonbailey, line 110: These two components are really just for consistency when creating "normal" buttons. For custom stuff like this I don't think there's any benefit from using them (and extending their APIs to support stuff like `color`).

src/components/Button.tsx
```diff
@@ -562,7 +564,7 @@ export function ButtonIcon({
         size={
           iconSize ?? (size === 'large' ? 'md' : size === 'tiny' ? 'xs' : 'sm')
         }
-        style={[{color: textStyles.color, pointerEvents: 'none'}]}
+        style={[{color: color ?? textStyles.color, pointerEvents: 'none'}]}
```
estrattonbailey, line 567: Yeah could pry drop this, would rather not encourage its use. Happy to discuss though!

estrattonbailey (commented):
LGTM, just the one nit, gonna sketch out my thoughts in a doc to discuss

src/components/Button.tsx
mozzius: Yeah np, I don't think I ended up needing it

haileyok:
hell yea, this has been bothering me every time i go to repost, seeing that ugly round button 😭

mozzius:
also, are we happy with the hover state?

Screenshot 2024-04-04 at 21 56 06
Screenshot 2024-04-04 at 21 56 30

pfrazee:
|ios feed before|after|
|-|-|
|CleanShot 2024-04-04 at 15 35 29@2x|CleanShot 2024-04-04 at 15 26 01@2x|

|ios thread before|after|
|-|-|
|CleanShot 2024-04-04 at 15 35 51@2x|CleanShot 2024-04-04 at 15 27 12@2x|

|web feed before|after|
|-|-|
|CleanShot 2024-04-04 at 15 32 03@2x|CleanShot 2024-04-04 at 15 29 02@2x|

|web thread before|after|
|-|-|
|CleanShot 2024-04-04 at 15 31 43@2x|CleanShot 2024-04-04 at 15 29 30@2x|

pfrazee:
Excited for this change, but you know Imma nitpick.

First two thoughts:

CleanShot 2024-04-04 at 15 38 28@2x

Also the icons seem slightly larger in general. I may need to see it on an actual device but my initial reaction was they seemed big.

mozzius:
Hmm, interesting that the spacing has all changed. I think this is from the padding needed for the hover state

src/view/com/util/post-ctrls/PostCtrls.tsx
```diff
@@ -141,105 +145,108 @@ let PostCtrls = ({
...
 
+  const btnStyle = React.useCallback(
+    ({pressed, hovered}: PressableStateCallbackType) => [
+      a.gap_xs,
+      a.rounded_full,
+      a.flex_row,
+      a.align_center,
+      a.justify_center,
+      {padding: 5},
+      (pressed || hovered) && t.atoms.bg_contrast_50,
```
estrattonbailey, line 156: 50 feels kinda dark to me, 25 is kinda our default bg state elsewhere

src/view/com/util/post-ctrls/RepostButton.tsx
```diff
@@ -1,108 +1,127 @@
...
+                size="large"
+                variant="outline">
+                <Repost size="lg" fill={t.palette.primary_500} />
+                <Text style={[a.font_bold, a.text_xl]}>
+                  {isReposted
+                    ? _(msg`Remove repost`)
+                    : _(msg({message: `Repost`, context: 'action'}))}
+                </Text>
+              </Button>
+              <Button
```
estrattonbailey, line 96: Nit: these should really have `pressed` states on native so the user gets feedback when they tap (or drag their finger away). Could do with the child function of Button, or via Pressable as you did for the PostCtrls.

src/view/com/util/post-ctrls/PostCtrls.tsx
mozzius: I added a `hoverStyle` prop to `Button` to achieve this because `secondary`+`ghost` uses 100, is this ok? Alternatively, I could change `secondary`+`ghost` to be 25, or I could use a raw `<Pressable>` (not ideal)

gaearon (approved):

PR #3402, opened 2024-04-04
rm 'pornography' label wording

haileyok:
This is a follow up to https://github.com/bluesky-social/social-app/pull/3282 where we reverted the change to the "Adult Content". The appeals screen still shows "Pornography", so this PR replaces that with "Adult Content" again.

### DIFF
--- a/src/lib/moderation/useReportOptions.ts
+++ b/src/lib/moderation/useReportOptions.ts
@@ -1,7 +1,7 @@
-import {msg} from '@lingui/macro'
-import {useLingui} from '@lingui/react'
 import {useMemo} from 'react'
 import {ComAtprotoModerationDefs} from '@atproto/api'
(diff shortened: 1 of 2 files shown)


src/lib/moderation/useReportOptions.ts
```diff
@@ -68,7 +68,7 @@ export function useReportOptions(): ReportOptions {
         {
           reason: ComAtprotoModerationDefs.REASONSEXUAL,
           title: _(msg`Unwanted Sexual Content`),
-          description: _(msg`Nudity or pornography not labeled as such`),
+          description: _(msg`Nudity not labeled as such`),
```
pfrazee, line 71: Maybe "Nudity or adult content not labeled as such"

pfrazee (approved):
Small suggestion, otherwise +1

PR #3404, opened 2024-04-04
Extract query key roots to constants

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3401 intended to avoid this problem by always using constants. We'd need to be disciplined about this. If we slip again, I'll write a lint rule, but I think this will do for now.

## Test Plan

Careful read. Plus walked around the app.

### DIFF
--- a/src/lib/react-query.tsx
+++ b/src/lib/react-query.tsx
@@ -11,7 +11,8 @@ import {
 import {isNative} from '#/platform/detection'
 
 // any query keys in this array will be persisted to AsyncStorage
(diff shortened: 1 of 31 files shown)


pfrazee (approved):
Visually checked all of the changes, looks correct

gaearon:
Pushed https://github.com/bluesky-social/social-app/pull/3404/commits/37b92f73c983885c5bff3fc3fa2b5e1155124449 to dedupe a potential future misalignment.

PR #3405, opened 2024-04-04
replace deprecated `term` with `q` in search actor

haileyok:
Removes deprecated query param, see https://github.com/bluesky-social/atproto/pull/2376/files#diff-94910126c7c4193579395efd9490525ceeef6a8ce01ab7a9d7687e099af9ac62L13


pfrazee (approved):
Didnt test

haileyok:
Screenshot 2024-04-04 at 11 12 33 AM

Tested both of these endpoints using `q`, works the same as before 👍

PR #3407, opened 2024-04-04
Enforce Text suffix for Text-rendering components

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3398.

This renames a bunch of components that always "promise" to render Text:

- `Description` -> `DescriptionText`
- `InlineLink` -> `InlineLinkText`
- `Label` -> `LabelText`
- `Suffix` -> `SuffixText`

It's not super "pretty" but it very clearly communicates the expectations that they render Text.

I've left H1..H7 and P "as is" because they're obviously text.

The last commit adds a lint rule check that enforces that such components *always* return `Text`.

Screenshot 2024-04-04 at 21 09 57

That's to prevent regressions.

This doesn't address buttons which I'll need to take a pass on separately.

## Test Plan

Did this via IDE renames (TS language server). Walked around the app too.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,21 +25,15 @@ module.exports = {
       {
         impliedTextComponents: [
           'Button', // TODO: Not always safe.
-          'DateField.Label',
           'H1',
           'H2',
           'H3',
           'H4',
           'H5',
           'H6',
-          'InlineLink',
-          'Label',
           'P',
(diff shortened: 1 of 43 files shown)


haileyok:
@estrattonbailey Not needed in this PR, but are we planning on using H1-7 at all? Can we rm if not? I think we decided not to but idr for sure.

PR #3408, opened 2024-04-04
Search - extra tabs

mozzius:
This PR adds the additional Top/Latest tabs, behind a feature gate. The new tabs just have the same content, because we don't have the new APIs yet. It also allows the tab names to be translated

Screenshot 2024-04-04 at 21 15 04


mozzius:
huh, commit b2727d9a isn't appearing. odd.

gaearon:
Seeing some duplicate views when in the gate. Both empty states and loading states are duplicated.

Screenshot 2024-04-05 at 14 15 25

Also People tab comes up empty for me. We should populate it with the same content as current Users tab.

Screenshot 2024-04-05 at 14 17 08

gaearon (changes requested):
see above

src/view/screens/Search/Search.tsx
```diff
@@ -316,8 +324,33 @@ function SearchScreenUserResults({query}: {query: string}) {
   )
 }
 
-const SECTIONS_LOGGEDOUT = ['Users']
-const SECTIONS_LOGGEDIN = ['Posts', 'Users']
+const SECTIONS_LOGGEDOUT = ['PEOPLE'] as const
```
gaearon, line 327: I think we’ll want to open up posts search logged out too. Let’s keep track of this somewhere.

src/view/screens/Search/Search.tsx
```diff
@@ -316,8 +324,33 @@ function SearchScreenUserResults({query}: {query: string}) {
...
+  i18n: I18n,
+  section:
+    | (typeof SECTIONS_LOGGEDIN)[number]
+    | (typeof SECTIONS_LOGGEDOUT)[number]
+    | (typeof SECTIONS_LOGGEDIN_LEGACY)[number]
+    | (typeof SECTIONS_LOGGEDOUT_LEGACY)[number],
+) {
+  switch (section) {
+    case 'TOP':
+      return i18n._(msg`Top`)
```
gaearon, line 342: I think (?) we shouldn’t use the global `_` because it won’t update on lang change. https://lingui.dev/tutorials/react-patterns

Let’s just move this inline into the component. Maybe right into the useMemo that wants it.

src/view/screens/Search/Search.tsx
```diff
@@ -330,6 +363,10 @@ export function SearchScreenInner({
   const setDrawerSwipeDisabled = useSetDrawerSwipeDisabled()
   const {hasSession} = useSession()
   const {isDesktop} = useWebMediaQueries()
+  const {i18n} = useLingui()
+
+  // New Search feature gate
```
gaearon, line 368: redundant comment

gaearon (approved):
Looks good if nits are fixed.

I wonder if it might feel more natural to move all branching to one place. eg you can have `sections` be an array of `{ title: _(…), content: <Stuff /> }` and then iterate over those. Then there’s no risk that the titles will get out of sync with the content.

src/view/screens/Search/Search.tsx
```diff
@@ -330,6 +363,10 @@ export function SearchScreenInner({
...
+  const isNewSearch = useGate('new_search')
```
estrattonbailey, line 369: Nit: could add these gates as `const` values in a file near StatSig stuff. May be nice for features that span lots of components.

mozzius: Good idea! done

haileyok:
I don't mind this as far as UI goes, and this isn't something that needs to land in this PR, but I want to call out that we probably are going to want to add logic to only perform the search for the selected tab. Right now AFAIK the Posts tab and the Users tab both complete the search and render data in both tabs at once. This is going to be a performance issue on some devices if we start adding more tabs here (especially if it's a tab of posts).

Not sure what the logic we use in home is for this, but maybe similar?

mozzius:
That's a good shout, I'll look at that next week

src/view/screens/Search/Search.tsx
```diff
@@ -339,6 +347,63 @@ export function SearchScreenInner({
     [setDrawerSwipeDisabled, setMinimalShellMode],
   )
 
+  const sections = React.useMemo(() => {
+    if (!query) return []
```
gaearon, line 351: This seems suspicious — previously we never had an empty array of pager items. I'm not sure the pager supports zero items.

gaearon: Hm I see, this shouldn't happen anyway because then the conditionals below would render a different codepath.

src/view/screens/Search/Search.tsx
```diff
@@ -389,13 +449,11 @@ export function SearchScreenInner({
...
           style={[pal.border, pal.view, styles.tabBarContainer]}>
-          <TabBar items={SECTIONS_LOGGEDOUT} {...props} />
+          <TabBar items={sections.map(section => section.title)} {...props} />
         </CenteredView>
       )}
       initialPage={0}>
-      <View>
-        <SearchScreenUserResults query={query} />
-      </View>
+      {sections.map(section => section.title)}
```
gaearon, line 456: You probably meant `component` here

gaearon: Edited, this fixed the crash

src/view/screens/Search/Search.tsx
```diff
@@ -347,16 +412,11 @@ export function SearchScreenInner({
...
           </CenteredView>
         )}
         initialPage={0}>
-        <View>
-          <SearchScreenPostResults query={query} />
-        </View>
-        <View>
-          <SearchScreenUserResults query={query} />
-        </View>
+        {sections.map(section => section.component)}
```
gaearon, line 419: According to the docs, its direct children have to be `View`s: https://github.com/callstack/react-native-pager-view?tab=readme-ov-file#usage

gaearon (approved):
I think this works, let's merge after the next release

PR #3410, opened 2024-04-04
Fix false positive DEV check

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3394.

React can already handle stray _empty_ strings just fine, so the `console.error` from RNW is overzelaous.

Let's silence it to match the actual behavior.

## Test Plan

Moderation -> Muted words and tags

No longer crashes in DEV on web.


haileyok (approved):
👍

PR #3411, opened 2024-04-04
Use CSS keyframe animation for spinner on web

mozzius:
We're using reanimated to make something spin in a circle. This is fine on native due to the UI thread, but on web it's prone to stuttering, especially on lower end devices.

End result is exactly the same:

https://github.com/bluesky-social/social-app/assets/10959775/41c0be98-bde2-48a2-b33a-b19ac56fdde5

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -220,6 +220,21 @@
     .nativeDropdown-item:focus {
       outline: none;
     }
+
+    /* Spinner component */
+    @keyframes rotate {
+      0% {
+        transform: rotate(0deg);
+      }
+      100% {
+        transform: rotate(360deg);
+      }
+    }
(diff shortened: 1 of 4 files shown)


bskyweb/templates/base.html
```diff
@@ -220,6 +220,21 @@
     .nativeDropdown-item:focus {
       outline: none;
     }
+
```
gaearon, line 223: Is it enough to change `base.html` or should changes be mirrored in `index.html`? @pfrazee

gaearon (approved):
lgtm

bskyweb/templates/base.html
mozzius: I suspect `index.html` is for the local dev version, so I've opted to mirror it so that it works there too

PR #3412, opened 2024-04-04
Remove special cases for some buttons with text

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3407.

I'm continuing removing special cases so that we can confidently render stuff and know whether it supports text or not.

- `Prompt.Cancel` and `Prompt.Action`: Those are barely used, and all usages are text-only. I'm just enforcing that they _don't_ take `children`. The `cta` prop is then sufficient. (Side note: We should really stop using `PropsWithChildren`.)
- `Toggle.Button` -> `Toggle.ButtonWithText`: This case is more interesting. I settled on a convention:
  - `Button` is a raw vanilla button with no opinion on children. Can't nest strings there (lint rule would flag it).
  - `ButtonText` is a button-specific text component. Similar to `ButtonText` in normal buttons.
  - <s>`ButtonWithText` **(new!)** combines them as a shortcut. The suffix means the lint rule would let you nest text there.</s>
    - This didn't fit very well, see below. We can make it work but doesn't seem worth it. If we hate the long names, can rename to `ToggleButton.Text` and do the same for others but I didn't do it here.

The strategy for `Toggle.Button` seems viable for the main `Button` as well. But I won't do it here.

## Test Plan

Storybook.

Would appreciate someone checking the UI.

Screenshot 2024-04-04 at 22 53 38

Screenshot 2024-04-04 at 22 35 16

### DIFF
--- a/src/components/Prompt.tsx
+++ b/src/components/Prompt.tsx
@@ -91,15 +91,13 @@ export function Actions({children}: React.PropsWithChildren<{}>) {
 }
 
 export function Cancel({
-  children,
   cta,
-}: React.PropsWithChildren<{
+}: {
   /**
-   * Optional i18n string, used in lieu of `children` for simple buttons. If
-   * undefined (and `children` is undefined), it will default to "Cancel".
+   * Optional i18n string. If undefined, it will default to "Cancel".
    */
   cta?: string
-}>) {
+}) {
   const {_} = useLingui()
   const {gtMobile} = useBreakpoints()
   const {close} = Dialog.useDialogContext()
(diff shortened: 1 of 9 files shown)


src/components/forms/ToggleButton.tsx
```diff
@@ -47,49 +63,42 @@ function ButtonInner({children}: React.PropsWithChildren<{}>) {
   const t = useTheme()
   const state = Toggle.useItemContext()
 
-  const {baseStyles, hoverStyles, activeStyles, textStyles} =
-    React.useMemo(() => {
-      const base: ViewStyle[] = []
-      const hover: ViewStyle[] = []
-      const active: ViewStyle[] = []
-      const text: TextStyle[] = []
+  const {baseStyles, hoverStyles, activeStyles} = React.useMemo(() => {
```
gaearon, line 66: This is same code but with no `textStyles` calculation.

src/components/forms/ToggleButton.tsx
```diff
@@ -110,19 +119,37 @@ function ButtonInner({children}: React.PropsWithChildren<{}>) {
...
+      {children}
     </View>
   )
 }
+
+export function ButtonText({children}: {children: React.ReactNode}) {
+  const t = useTheme()
+  const state = Toggle.useItemContext()
+
+  const textStyles = React.useMemo(() => {
```
gaearon, line 131: Lifted from code above, just the `textStyles` calculation.

gaearon:
Actually I think this doesn't quite work if we remove `Button` itself from the "special" list. I think we have to bite the bullet here and just explicitly use `ButtonText` when needed. It's also simpler to convert back and forth if needed.

haileyok (approved):
I don't mind using `ButtonText` myself, as long as it tells me if I forget to! This seems good to me.

I clicked through the app where we use the new buttons at, everything looks good. Storybook seems fine as well.

estrattonbailey (approved):
I ❤️ conventions

src/components/Prompt.tsx
```diff
@@ -151,7 +147,7 @@ export function Action({
       label={cta || _(msg`Confirm`)}
       onPress={handleOnPress}
       testID={testID}>
-      {children ? children : <ButtonText>{cta || _(msg`Confirm`)}</ButtonText>}
+      <ButtonText>{cta || _(msg`Confirm`)}</ButtonText>
```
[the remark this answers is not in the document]
gaearon: Yea I figured we can revisit when types are tighter

PR #3414, opened 2024-04-05
Fabric migration - `react-native-version-number` -> `expo-application`

haileyok:
This PR removes `react-native-version-number` and uses `expo-application` to retrieve the same values. Tested on all platforms and the outputs line up with each other.

## Test Plan

Open the settings screen, scroll to bottom, see that the "Build Version" is still correct (it should be 1.71.0 (1) on native, 1.71.0 on web)

- [x] iOS
- [x] Android
- [x] Web


haileyok:
Didn't work, will do this again manually.

PR #3421, opened 2024-04-05
Enforce that text is wrapped in <Text>, remaining cases

gaearon:
# Stacked on https://github.com/bluesky-social/social-app/pull/3412

This is the final boss in the https://github.com/bluesky-social/social-app/pull/3398 stack that gets rid of the last significant exception to the rule.

Concretely, `<Button>` no longer tries to treat its string children specially. Instead, you're supposed to put `<ButtonText>` in there if you want to place a string or `<Trans>`. In return for some extra verbosity, you get higher safety — the lint rule and types would catch most cases where you might accidentally leave raw text (or raw `<Trans>`) in a wrong place.

Unlike the previous approach (which auto-wrapped strings and `<Trans>` components at runtime), this doesn't change the behavior when things get wrapped an extra level (e.g. into a shared `<View>`) which is a pitfall of the original approach. It also works the same whether you use the children function form of `<Button>` or the plain children form.

## Commits to Review

- https://github.com/bluesky-social/social-app/pull/3421/commits/f83c1a78af7fe8b183cf2c7bc80249e4199657d7?w=1 is the big switchover, adding `<ButtonText>` where it was previously implied
- https://github.com/bluesky-social/social-app/pull/3421/commits/ff6684964a35b04fd332af6ca6c168778cc427bc tightens up the rule (see https://github.com/bluesky-social/social-app/pull/3421#issuecomment-2038818549)
- https://github.com/bluesky-social/social-app/pull/3421/commits/f60705460b3335834c78664f632bcc457c67698d further tightens up the rule (see https://github.com/bluesky-social/social-app/pull/3421#issuecomment-2038936194)

I kept the implementation low-abstraction, i.e. there's some copy paste. Intentional so it's easy to change.

## Methodology

I did a project search for `<Button`, went through all cases manually, and wrapped in `<ButtonText>` when a plain string or `<Trans>` was being passed as a direct child. That's the only case in which auto-wrapping was being performed.

I also went through "find all references" of `Button` to see if my search missed something. Then I went through "find all references" of all components passing their own `children` to a `Button`. That didn't lead me to any further violations.

## So what?

This catches JSX text inside `<Button>` (whether with `<Trans>` or not) and ensures it's wrapped.

Screenshot 2024-04-05 at 04 09 50

Screenshot 2024-04-05 at 04 10 30

Works for this version too.

Screenshot 2024-04-05 at 04 12 09

Note the lint rule only catches string literals, template literals, and JSX tags. It won't catch embedded variables like `<Button>{someString}</Button>`. However, now that we locked it down a bit, these cases can be caught with types:

Screenshot 2024-04-05 at 12 18 12

If we merge this, we need to rebase and re-review PRs against `main` that add more buttons. I expect most issues to be caught at the build time though so I think it shouldn't be much trouble.

## Test Plan

Careful inspection.

I also walked around the app and checked most screens.

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -405,7 +405,6 @@ export function Button({
         </View>
       )}
       <Context.Provider value={context}>
-        {/* @ts-ignore */}
         {typeof children === 'function' ? children(context) : children}
       </Context.Provider>
     </Pressable>
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -12,7 +12,6 @@ import {
   ViewStyle,
 } from 'react-native'
 import {LinearGradient} from 'expo-linear-gradient'
-import {Trans} from '@lingui/macro'
 
 import {android, atoms as a, flatten, tokens, useTheme} from '#/alf'
 import {Props as SVGIconProps} from '#/components/icons/common'
@@ -59,6 +58,10 @@ export type ButtonState = {
 
 export type ButtonContext = VariantProps & ButtonState
 
+type NonTextElements =
+  | React.ReactElement
+  | Iterable<React.ReactElement | null | undefined | boolean>
+
 export type ButtonProps = Pick<
   PressableProps,
   'disabled' | 'onPress' | 'testID'
@@ -68,11 +71,9 @@ export type ButtonProps = Pick<
     testID?: string
     label: string
     style?: StyleProp<ViewStyle>
-    children:
-      | React.ReactNode
-      | string
-      | ((context: ButtonContext) => React.ReactNode | string)
+    children: NonTextElements | ((context: ButtonContext) => NonTextElements)
   }
+
 export type ButtonTextProps = TextProps & VariantProps & {disabled?: boolean}
 
 const Context = React.createContext<VariantProps & ButtonState>({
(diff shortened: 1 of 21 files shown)


gaearon:
>We could in principle check them in the future by catching usage of <Trans> itself and treating it similar to JSX text.

I guess I'll explore this for a bit so we have a stronger guarantee.

gaearon:
Pushed https://github.com/bluesky-social/social-app/pull/3421/commits/5902f0d657c8470f4f449533375b701049617358.

Now we flag any `<Trans>` that isn't directly nested into some `<Text>`.

Screenshot 2024-04-05 at 04 45 29
Screenshot 2024-04-05 at 04 45 41

This should solve the remaining cases.

gaearon:
Note that this is still a pitfall:

Screenshot 2024-04-05 at 04 50 00

However, the old approach wouldn't catch it either. We might be able to improve on this with some effort but since it's not a regression, that's definitely out of scope.

gaearon:
Pushed https://github.com/bluesky-social/social-app/pull/3421/commits/f221048ff56fb496221540eefbcb0f3b8dc5368a.

This covers these cases:

Screenshot 2024-04-05 at 05 53 02

Screenshot 2024-04-05 at 05 55 11

src/components/Button.tsx
```diff
@@ -59,6 +58,10 @@ export type ButtonState = {
 
 export type ButtonContext = VariantProps & ButtonState
 
+type NonTextElements =
+  | React.ReactElement
+  | Iterable<React.ReactElement | null | undefined | boolean>
```
estrattonbailey, line 63: Handy

estrattonbailey (approved):
🔥

PR #3428, opened 2024-04-05
fix dialog back button for android

haileyok:
This is a follow-up to https://github.com/bluesky-social/social-app/pull/3330 and https://github.com/bluesky-social/social-app/pull/3343

I am implementing the change in #3330 here for _all_ dialogs now. I am also adding a callback to `close()` on web, so that the functionality remains the same across all platforms. This prevents issues where the dialog gets unmounted before the removal of the dialog from the dialog state is finished.

### On Android

- Hide post through the post menu
- Open another post
- Try to use the back button

### On Web

- Open the post composer on web
- Insert some text
- Press cancel, dismiss the composer

### DIFF
--- a/src/components/Dialog/index.web.tsx
+++ b/src/components/Dialog/index.web.tsx
@@ -50,7 +50,7 @@ export function Outer({
   }, [control.id, onClose, setDialogIsOpen])
 
   const close = React.useCallback<DialogControlProps['close']>(
-    async cb => {
+    cb => {
       if (cb && typeof cb === 'function') {
         cb()
       }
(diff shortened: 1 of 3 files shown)


src/components/Dialog/index.web.tsx
```diff
@@ -37,22 +40,32 @@ export function Outer({
...
-  }, [onClose, setIsOpen, setDialogIsOpen, control.id])
+  }, [control.id, onClose, setDialogIsOpen])
+
+  const close = React.useCallback<DialogControlProps['close']>(
+    cb => {
+      if (cb && typeof cb === 'function') {
+        cb()
+      }
+      onCloseInner()
+    },
```
estrattonbailey, line 58: Fine to make this a separate method imo, but let's also wrap in a try/catch to match native impl?

estrattonbailey (approved):

src/components/Dialog/index.web.tsx
estrattonbailey: Oh I see, types, makes sense to have seprate 👍

PR #3429, opened 2024-04-05
Remove vertical scrollbars from views on native

haileyok:
## Rationale

Because a lot of our lists are virtualized, there is a bit of "jank" whenever the view resizes. For example in the feed, each time a new batch of items gets rendered below our current scroll position, the scrollbar changes position. Additionally, whenever we load the next batch of items from the server, the scrollbar jumps up the page to account for the new items.

Other apps, such as Meta products, do not show the vertical scrollbar in any scroll view - including non-virtual lists such as settings, etc.

My opinion is that we definitely should now show scrollbars on our virtualized lists for two reasons:

1. They are janky and just look weird
2. They take away from the sense of "infinite" items (where applicable). I expect my feed to be continuous no matter how far I scroll. Having a scrollbar makes takes away from that thought a bit.

And since we are removing the scrollbars from the main parts of the app, I also believe we should just remove them from the other few `ScrollView` areas. I have two reasons for that as well:

1. If we don't show them everywhere, it's a bit strange to show them only in some places. Users don't care or know about _why_ virtualization causes jank and why we removed them from those screens, they just know that sometimes they see a scrollbar and sometimes they do not.
2. Even without the virtualization jank, our scrollbars still get hidden under parts of our UI shell (since we are not using the native navigation shell from `react-navigation`, primarily.

## How

This was pretty simple. There are two places where we get these: `List` and `ScrollView`. Thankfully, we have nice little wrappers for these already, so I only needed to modify this prop in two places. I went through the app and can't find anywhere that I missed this.

There were a few places that we are using `FlatList` directly as well, which can get fixed in a separate PR (there's some other things there that are no good like using anonymous functions as props). For now I added `showsVerticalScrollIndicator={false}` to those `FlatList`s manually.

## Test Plan

This is mainly a visual change, so really its a matter of going around the app and seeing if not having the scrollbars is something that seriously bothers you or not. We might decide some areas _need_ them, but from my experience so far I have not ran into any of those cases.


src/view/com/auth/onboarding/RecommendedFeeds.tsx
```diff
@@ -130,6 +131,7 @@ export function RecommendedFeeds({next}: Props) {
               renderItem={({item}) => <RecommendedFeedsItem item={item} />}
               keyExtractor={item => item.uri}
               style={{flex: 1}}
+              showsVerticalScrollIndicator={false}
```
haileyok, line 134: These screens actually look better in general without indicators (and already do not show them on web) so let's just set the prop to false.

src/view/com/util/List.tsx
```diff
@@ -37,7 +40,8 @@ function ListImpl<ItemT>(
   const isScrolledDown = useSharedValue(false)
   const contextScrollHandlers = useScrollHandlers()
   const pal = usePalette('default')
-
+  const showsVerticalScrollIndicator =
+    useGate('shows_vertical_scroll_indicator') && isWeb
```
haileyok, line 44: I am using `shows_vertical_scroll_indicator` for the gate name, even though we're actually _removing_ the indicators. I think it makes more sense to use the wording `shows_vertical_scroll_indicator` just so we are not inverting the boolean. We can change the name though if we want to keep the gate name identical to what we are actually testing.

src/view/com/util/Views.jsx
```diff
@@ -2,8 +2,22 @@ import React from 'react'
...
 export const FlatList_INTERNAL = Animated.FlatList
-export const ScrollView = Animated.ScrollView
 export function CenteredView(props) {
   return <View {...props} />
 }
+
+export function ScrollView(props) {
+  const showsVerticalScrollIndicator = useGate(
+    'shows_vertical_scroll_indicator',
+  )
```
haileyok, line 15: This component is only used on native, so we don't need an `isWeb` check here.

src/view/com/util/List.tsx
haileyok: Ah, I just realized this isn't ideal because the `useGate` will return `false` IIRC.

src/view/com/util/List.tsx
```diff
@@ -37,7 +40,8 @@ function ListImpl<ItemT>(
   const isScrolledDown = useSharedValue(false)
   const contextScrollHandlers = useScrollHandlers()
   const pal = usePalette('default')
-
+  const showsVerticalScrollIndicator =
+    !useGate('hide_vertical_scroll_indicators') && isWeb
```
pfrazee, line 44: should this be && or should it be || ?

pfrazee: nm that's correct

pfrazee: wait. is it?

haileyok: God I had my explanation typed out and then realized you're right. What a mind fuck. Sec.

haileyok: okay fixed...

Case 1:
- Gate false -> inverted -> true
- Is not on web -> false
- Shows indicator

Case 2:
- Gate false -> inverted -> true
- Is on web -> true
- Shows indicator

Case 3:
- Gate true -> inverted -> false
- Is not on web -> false
- Doesn't show

Case 4:
- Gate true -> inverted -> false
- Is on web -> true
- Shows indicator

pfrazee (approved):
Yeah I'm +1 to this, especially with the new profile header behavior it becomes super noticable

PR #3430, opened 2024-04-06
[Experiment] Disable min shell on foregrounding

gaearon:
Just a simple feature gate to start playing with it.

When `disable_min_shell_on_foregrounding` is on, the app will disable minimal shell on foregrounding. The thing I'm curious about is whether this will lead to more Notifications tab navigations (since we won't be hiding the jewel).

I do this on the home tab because it's unsafe to do globally. E.g. don't want to screw up onboarding or logged out state.

## Test Plan

Overrode it in the console to `true` for self. Minimal shell gets disabled on foregrounding.

https://github.com/bluesky-social/social-app/assets/810438/5cdb88c0-dcfe-497e-b5fc-9e3c4c96d40d

Also verified on (mobile) web.

Verified logged out state isn't affected by this.

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -1,23 +1,25 @@
 import React from 'react'
-import {View, ActivityIndicator, StyleSheet} from 'react-native'
+import {ActivityIndicator, AppState, StyleSheet, View} from 'react-native'
 import {useFocusEffect} from '@react-navigation/native'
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Cool. I think I actually prefer this myself. In fact, part of me wonders if it might be something we want to keep visible at all times. I can't think of another app that hides it _except_ for Twitter. I'm not sure!

PR #3431, opened 2024-04-06
[Experiment] Turn off interval polling for Discover

gaearon:
The first commit adds the feed URL to the metadata so we can check metrics against individual feeds.

In the second commit, I add a feature gate that disables polling for Discover. The hypothesis is that the blue dot is getting people to refresh it more often instead of scrolling to the end, but the blue dot doesn't quite make sense for algo feeds where the content is _always_ different. This removes the polling-based blue dot for Discover so we can check if this changes the user behavior. The blue dot will still appear when foregrounding the app.

We can run more variations on this experiment so this is just something to play with while learning how to do it.

## Test Plan

Decreased polling interval to 5 seconds. Verified blue dot stopped getting blue on timer for the Discover feed. Still gets blue on timer for other feeds. Foregrounding still makes it blue for Discover.

### DIFF
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -45,10 +45,12 @@ export type LogEvents = {
   'onboarding:moderation:nextPressed': {}
   'onboarding:finished:nextPressed': {}
   'feed:endReached': {
+    feedUrl: string
     feedType: string
     itemCount: number
   }
   'feed:refresh': {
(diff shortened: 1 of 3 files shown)


haileyok (approved):
Dropped this in Slack as well, but think there's a lot of experimentation we could do around this button. Might be worth considering just rming it completely at some point based on some user discussions.

PR #3432, opened 2024-04-06
[Experiment] Auto-expand follow suggestions on web

gaearon:
If you're in the `autoexpand_suggestions_on_profile_follow` gate, you'll get this behavior:

https://github.com/bluesky-social/social-app/assets/810438/08656e5c-ea18-4e5a-8745-a6ed24ea2903

It's web-only because of scrolling issues on iOS/Android.

I've gated it as web-only in the code but I also had to add web-only condition to the gate setup itself. This is because we don't want to count exposures to the flag being on unless we're actually providing the changed behavior.

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -78,6 +78,10 @@ function toStatsigUser(did: string | undefined) {
   return {
     userID,
     platform: Platform.OS,
+    custom: {
+      // Need to specify here too for gating.
+      platform: Platform.OS,
+    },
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/statsig/statsig.tsx
```diff
@@ -78,6 +78,10 @@ function toStatsigUser(did: string | undefined) {
   return {
     userID,
     platform: Platform.OS,
+    custom: {
+      // Need to specify here too for gating.
+      platform: Platform.OS,
```
gaearon, line 83: Gating by platform didn't work unless I put it under `custom` namespace. Maybe we don't have to send both but let's keep both for now.

haileyok (approved):
Strong +1 to this. Verified that we will be able to display this on iOS as well whenever we land the scroll fixes, so looking forward to that.

PR #3437, opened 2024-04-06
Migrate to `expo-file-system`

haileyok:
This implements the changes made in https://github.com/bluesky-social/social-app/pull/3102

This also will help us to resolve an additional missing API declaration for `NSPrivacyAccessedAPICategoryDiskSpace`. Those changes have not yet been merged into Expo, but if we get too close to the May deadline without them merging, we can patch with some changes in this PR: https://github.com/expo/expo/pull/27770 (or available in this canary, though I have not looked to see what other changes this includes https://www.npmjs.com/package/expo-file-system/v/0.0.1-canary-20240328-24ecc5e)

### Copied from previous PR:

This removes migrates our use of `react-native-fs` and `rn-fetch-blob` to `expo-file-system`. It also makes use of `expo-image-manipulator` instead of using `react-native-image-resizer`.

It also fixes a bug where the image manipulator would try to save to the same file path again. We are now cleaning up after ourselves, so this can no longer happen. Of course, in cleaning up we are also preventing the unnecessary growth of the local file store as well.

Tests are also updated to reflect the use of new libraries.

## Testing

- Save image (Tests saving and cache)
	- [x] iOS
	- [x] Android
	- [x] Not applicable on web
- Share image (Tests saving and cache)
	- [x] iOS
	- [x] Android
	- [x] Not applicable on web
- Add a link card (Tests both saving and resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload an image (Tests resizing/compression) 
 	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload an avatar (Tests resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works
- Upload a banner (Tests resizing/compression)
	- [x] iOS
	- [x] Android
	- [x] Web - No changes, but verify it still works


src/lib/api/index.ts
```diff
@@ -39,10 +41,14 @@ export async function uploadBlob(
...
     // `blob` should be a path to a file in the local FS
-    return agent.uploadBlob(
+    const res = await agent.uploadBlob(
       blob, // this will be special-cased by the fetch monkeypatch in /src/state/lib/api.ts
       {encoding},
     )
+    try {
+      deleteAsync(blob)
+    } catch (e) {} // Don't need to handle
+    return res
```
haileyok, line 51: No logical change, only cleaning up.

src/lib/media/manip.ts
```diff
@@ -81,47 +76,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
-  let imagePath = downloadResponse.path()
-  imagePath = normalizePath(await moveToPermanentPath(imagePath, '.png'), true)
-
-  // NOTE
-  // for some reason expo-sharing refuses to work on iOS
-  // ...and visa versa
-  // -prf
-  if (isIOS) {
-    await RNShare.share({url: imagePath})
-  } else {
```
haileyok, line 102: This is no longer the case, and so we can simplify this logic. No real changes here.

src/lib/media/manip.ts
```diff
@@ -81,47 +76,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
-  RNFS.unlink(imagePath)
+
+  const imagePath = await moveToPermanentPath(imageUri, '.png')
+
+  await Sharing.shareAsync(imagePath, {
+    mimeType: 'image/png',
+    UTI: 'image/png',
+  })
+
+  deleteAsync(imagePath)
```
haileyok, line 98: Same logic, just a different library.

src/lib/media/manip.ts
```diff
@@ -81,47 +76,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
+  if (!imageUri) {
+    // download the file to cache
+    // NOTE
+    // assuming PNG
+    // we're currently relying on the fact our CDN only serves pngs
+    // -prf
+    imageUri = await downloadImage(uri, createPath('png'), 5e3)
+  }
+
+  const imagePath = await moveToPermanentPath(imageUri, '.png')
```
haileyok, line 112: This is a very small logic change. The only difference from before is that we attempt to use the already downloaded image for sharing. This is extremely helpful on poorer connections.

src/lib/media/manip.ts
```diff
@@ -81,47 +76,45 @@ export async function shareImageModal({uri}: {uri: string}) {
...
 
   // save
   await MediaLibrary.createAssetAsync(imagePath)
+
+  deleteAsync(imagePath)
```
haileyok, line 117: No logic change, just cleanup.

src/lib/media/manip.ts
```diff
@@ -203,11 +239,53 @@ function joinPath(a: string, b: string) {
...
+  const ratio = Math.min(
+    POST_IMG_MAX.width / originalDims.width,
+    POST_IMG_MAX.height / originalDims.height,
+  )
+
+  return {
+    width: Math.round(originalDims.width * ratio),
+    height: Math.round(originalDims.height * ratio),
+  }
+}
```
haileyok, line 291: This portion is necessary to remove the old image manipulator package. There's some tests to ensure that this works properly.

src/lib/media/manip.ts
```diff
@@ -203,11 +239,53 @@ function joinPath(a: string, b: string) {
...
+  const to1 = setTimeout(() => downloadResumable.cancelAsync(), timeout)
+  const downloadRes = await downloadResumable.downloadAsync()
+  clearTimeout(to1)
+
+  if (!downloadRes?.uri) {
+    throw new Error()
+  }
+
+  return normalizePath(downloadRes.uri)
+}
```
haileyok, line 269: This is from the removal of `rn-fetch-blob`. We can revert this portion if we want to for now.

src/lib/media/manip.ts
```diff
@@ -203,11 +239,53 @@ function joinPath(a: string, b: string) {
...
 }
 
-function normalizePath(str: string, allPlatforms = false): string {
-  if (isAndroid || allPlatforms) {
-    if (!str.startsWith('file://')) {
-      return `file://${str}`
-    }
+function normalizePath(str: string): string {
+  if (!str.startsWith('file://')) {
+    return `file://${str}`
```
haileyok, line 244: Same logic as before, but there's no need to differentiate platforms in _any_ case.

src/lib/media/manip.ts
```diff
@@ -182,12 +202,28 @@ async function moveToPermanentPath(path: string, ext = ''): Promise<string> {
...
+  // image_manager_disk_cache is used by Glide
+  if (
+    !path.includes('com.hackemist.SDImageCache') &&
+    !path.includes('image_manager_disk_cache')
+  ) {
+    try {
+      deleteAsync(path)
+    } catch (e) {
+      // No need to handle
+    }
```
haileyok, line 224: This is new logic, but it has no affect on the client. It only removes old files.

src/lib/media/manip.ts
```diff
@@ -147,27 +140,54 @@ interface DoResizeOpts {
...
+        path: normalizePath(resizeRes.uri),
         mime: 'image/jpeg',
-        size: resizeRes.size,
+        size: info.size,
         width: resizeRes.width,
         height: resizeRes.height,
       }
+    } else {
+      await deleteAsync(resizeRes.uri)
     }
```
haileyok, line 191: Another piece of new logic that only cleans up, no changes.

src/lib/api/api-polyfill.ts
```diff
@@ -33,9 +33,27 @@ async function fetchHandler(
       // we get around that by renaming the file ext to .bin
       // see https://github.com/facebook/react-native/issues/27099
       // -prf
-      const newPath = reqBody.replace(/\.jpe?g$/, '.bin')
-      await RNFS.moveFile(reqBody, newPath)
-      reqBody = newPath
```
haileyok, line 38: Even if we revert, this isn't a good move. From testing, some of these files could not be moved and would result in an error.

src/lib/media/manip.ts
```diff
@@ -53,26 +61,13 @@ export async function downloadAndResize(opts: DownloadAndResizeOpts) {
...
-    downloadRes = await downloadResPromise
-    clearTimeout(to1)
-
-    let localUri = downloadRes.path()
-    if (!localUri.startsWith('file://')) {
-      localUri = `file://${localUri}`
-    }
+  const path = createPath(appendExt)
 
-    return await doResize(localUri, opts)
```
haileyok, line 71: Only deduplication. No logical change.

src/lib/media/manip.ts
haileyok: (In regards to "expo-sharing refuses to work on iOS"). I'm fine with reverting though.

PR #3438, opened 2024-04-06
[Remove FontAwesome] Remove from `BottomBar`, `LeftNav`; remove some excessive styling from `BottomBar` and `BottomBarWeb`

haileyok:
## Why

We would like to remove the FontAwesome icon library from the app, as it is currently the biggest library that we are using in the app. This would shave about 1 MB off of our bundle size (parsed), which would be a significant improvement. This PR replaces the FontAwesome icons that are used in the `LeftNav` and the `BottomBar` with icons from the new icon library.

Additionally, while working on the `BottomBar` replacements, I realized that there was some excessive styling that could be achieved with flexbox rather than margins, negative margins, etc. I reworked these aspects of the styles here.

Note that these icons are _not_ a part of FontAwesome, but the natural first step in working through these feels like replacing the main navigation icons, which is why I started here.

## Test Plan

The icons that have changed here are on the left nav and the bottom bar. I've gone through all of these icons to ensure that they properly switch between filled/not filled when the respective tab is selected. I have also verified that the updated styles work on 

The more important thing that we will want to test here is if we _like_ the icons. Some of the icons do not have 1:1 replacements as far as design goes, so there are some subtle differences between them - primarily the home icon and the alert/bell icon. If we feel strongly about a particular icon not looking great, then we might want to replace.

### Android

Screenshot 2024-04-06 at 5 28 38 PM

### iOS

Screenshot 2024-04-06 at 5 28 50 PM

### Web Desktop

Screenshot 2024-04-06 at 5 29 07 PM

### Web Mobile

Screenshot 2024-04-06 at 5 29 47 PM


haileyok:
Note that the bottom bar on web being slightly offset to the right is due to the ghost scroll gutter issue that has not yet been fixed.

src/view/shell/desktop/LeftNav.tsx
```diff
@@ -299,141 +319,57 @@ export function DesktopLeftNav() {
...
           <NavItem
             href="/lists"
-            icon={
-              <ListIcon
-                style={pal.text}
-                size={isDesktop ? 26 : 30}
-                strokeWidth={2}
-              />
-            }
+            icon={<Bars style={t.atoms.text} size="lg2" />}
```
mozzius, line 351: I don't love this icon, but I don't see anything better in figma

mozzius (approved):
The icons look a little bit heavier than the current ones, however I don't think that's an issue. Tested iOS sim + web, LGTM

haileyok:
Yea I wanted to see what @estrattonbailey thought on that. Honestly, we don't need to get rid of these icons - they are not using FontAwesome anyway - if we don't like them. Really was just playing around w/ stuff on the weekend. With that said I'd just move the old icons into using the same system as we are using now if we wanted to keep them.

src/components/icons/TEMPLATE.tsx
```diff
@@ -28,7 +28,14 @@ export const IconTemplate_Stroke2_Corner0_Rounded = React.forwardRef(
...
 
-export function createSinglePathSVG({path}: {path: string}) {
+export function createSinglePathSVG({
+  path,
+  strokeWidth = 0,
+}: {
+  path: string
+  strokeWidth?: number
+}) {
+  console.log(strokeWidth)
```
estrattonbailey, line 38: Log :P

src/components/icons/Sample.tsx
estrattonbailey, line 1: Nice nice

src/view/shell/bottom-bar/BottomBarWeb.tsx
```diff
@@ -53,68 +63,47 @@ export function BottomBarWeb() {
...
               <NavItem routeName="Feeds" href="/feeds">
                 {({isActive}) => {
-                  return (
-                    <HashtagIcon
-                      size={22}
-                      style={[styles.ctrlIcon, pal.text, styles.feedsIcon]}
-                      strokeWidth={isActive ? 4 : 2.5}
-                    />
-                  )
+                  const Icon = isActive ? Hashtag : HashtagFilled
```
estrattonbailey, line 99: Should this be switched around?

estrattonbailey (approved):
Looks great!

haileyok:
We did it!

PR #3440, opened 2024-04-07
[Remove FontAwesome] Replace icons in the side drawer

haileyok:
A continuation of https://github.com/bluesky-social/social-app/pull/3438. Here we are just replacing the same icons from the previous PR in the side drawer as well on native.

## Test Plan

Open the drawer, select the various icons. In the instances where you can still open the drawer from the page you navigate to, verify that the icons become filled.


mozzius (approved):
Tested iOS sim + web, LGTM

haileyok:
We did it!

PR #3441, opened 2024-04-07
[Remove Font Awesome] Cleanup unused icons, replace a few more icons with the new icon library

haileyok:
Another continuation of https://github.com/bluesky-social/social-app/pull/3438 stacked on https://github.com/bluesky-social/social-app/pull/3440. This just cleans up some of the now unused icons, as well as a few other pieces of cleanup.

## Test Plan

Icons were replaced in the settings screen. Verified that the hashtag icon and the raised hand icon still look correct in that screen.

Also verified that the replaced `CogIcon` in the feeds screen looks correct.


haileyok:
A few of the remaining icons in `icons.tsx` will be replaced in https://github.com/bluesky-social/social-app/pull/3400. At that point we can probably get close to just RMing the entire `icons.tsx` file.

src/view/screens/Feeds.tsx
```diff
@@ -433,7 +435,7 @@ export function FeedsScreen(_props: Props) {
                     href="/settings/saved-feeds"
                     accessibilityLabel={_(msg`Edit My Feeds`)}
                     accessibilityHint="">
-                    <CogIcon strokeWidth={1.5} style={pal.icon} size={28} />
+                    <Gear style={pal.textLight} size="lg" />
```
mozzius, line 438: Should be `pal.icon`, otherwise there's a mismatch with the icon next to it

Screenshot 2024-04-08 at 15 58 21

mozzius (commented):
Just one small issue, otherwise looks great!

PR #3449, opened 2024-04-08
Small logic cleanups

estrattonbailey:
Tl;dr we don't want to pass nested callbacks into the Dialog `close(cb)` callback. I added a error log for this, but it's something that we need to be aware of going forward.

### DIFF
--- a/src/components/Prompt.tsx
+++ b/src/components/Prompt.tsx
@@ -123,6 +123,13 @@ export function Action({
   cta,
   testID,
 }: {
+  /**
+   * Callback to run when the action is pressed. The method is called _after_
+   * the dialog closes.
+   *
+   * Note: The dialog will close automatically when the action is pressed, you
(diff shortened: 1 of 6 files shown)


haileyok:
Merged change in from https://github.com/bluesky-social/social-app/pull/3451. Review discussion was in that PR.

gaearon (approved):

src/components/Prompt.tsx
```diff
@@ -123,6 +123,13 @@ export function Action({
   cta,
   testID,
 }: {
+  /**
+   * Callback to run when the action is pressed. The method is called _after_
+   * the dialog closes.
+   *
+   * Note: The dialog will close automatically when the action is pressed, you
+   * should NOT close the dialog as a side effect of this method.
```
estrattonbailey, line 131: Still seems relevant, will leave

pfrazee (approved):

PR #3451, opened 2024-04-09
Small logic cleanups

haileyok:
## Test Plan

- Test edge cases inside of the Storybook. Press `Open Tester` to see the various edge cases and watch the console logs.
- Composer
	- Type some junk
	- Attempt to close
	- On Android, attempt to use the back button now
- Hide Post
	 - On Android, attempt to use the back button now
- Report Post
    - Again, try to use the back button on Android (I usually just do a trash report to Beans Labeler for testing, lol)
- Just swipe to close a dialog. Everything should still work.
- Use the mute words dialog
    - Delete a mute word, and make sure the back button is working

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -12,6 +12,7 @@ import BottomSheet, {
   WINDOW_HEIGHT,
 } from '@discord/bottom-sheet/src'
 
+import {logger} from '#/logger'
 import {useDialogStateControlContext} from '#/state/dialogs'
 import {isNative} from 'platform/detection'
 import {atoms as a, flatten, useTheme} from '#/alf'
@@ -129,7 +130,9 @@ export function Outer({
       for (const cb of closeCallbacks.current) {
         try {
           cb()
-        } catch {}
+        } catch (e: any) {
+          logger.error('Error running close callback', e)
+        }
       }
     }
     onClose?.()
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -128,7 +128,6 @@ export function Outer({
       // No need to handle this here
     } finally {
       onClose?.()
-      setDialogIsOpen(control.id, false)
     }
   }, [control.id, onClose, setDialogIsOpen])
 
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -83,7 +83,6 @@ export function Outer({
   const sheetOptions = nativeOptions?.sheet || {}
   const hasSnapPoints = !!sheetOptions.snapPoints
   const insets = useSafeAreaInsets()
-  const isClosing = React.useRef<boolean>(false)
   const closeCallbacks = React.useRef<(() => void)[]>([])
   const {setDialogIsOpen} = useDialogStateControlContext()
 
@@ -116,7 +115,6 @@ export function Outer({
       setOpenIndex(index || 0)
 
       callQueuedCallbacks()
-      isClosing.current = false
     },
     [setDialogIsOpen, control.id, callQueuedCallbacks],
   )
@@ -126,11 +124,7 @@ export function Outer({
     if (typeof cb === 'function') {
       closeCallbacks.current.push(cb)
     }
-
-    if (!isClosing.current) {
-      isClosing.current = true
-      sheet.current?.close()
-    }
+    sheet.current?.close()
   }, [])
 
   // This is the actual thing we are doing once we "confirm" the dialog. We want the dialog's close animation to
@@ -143,7 +137,6 @@ export function Outer({
 
     callQueuedCallbacks()
     onClose?.()
-    isClosing.current = false
   }, [callQueuedCallbacks, control.id, onClose, setDialogIsOpen])
 
   useImperativeHandle(
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -113,8 +113,8 @@ export function Outer({
     }
 
     if (!isClosing.current) {
-      sheet.current?.close()
       isClosing.current = true
+      sheet.current?.close()
     }
   }, [])
 
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -110,11 +110,13 @@ export function Outer({
 
   const open = React.useCallback<DialogControlProps['open']>(
     ({index} = {}) => {
+      // Run any leftover callbacks that might have been queued up before calling `.open()`
+      callQueuedCallbacks()
+
       setDialogIsOpen(control.id, true)
       // can be set to any index of `snapPoints`, but `0` is the first i.e. "open"
       setOpenIndex(index || 0)
-
-      callQueuedCallbacks()
+      sheet.current?.snapToIndex(index || 0)
     },
     [setDialogIsOpen, control.id, callQueuedCallbacks],
   )
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -124,8 +124,6 @@ export function Outer({
       if (closeCallback.current) {
         closeCallback.current()
       }
-    } catch (e) {
-      // No need to handle this here
     } finally {
       onClose?.()
     }
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -12,7 +12,6 @@ import BottomSheet, {
   WINDOW_HEIGHT,
 } from '@discord/bottom-sheet/src'
 
-import {logger} from '#/logger'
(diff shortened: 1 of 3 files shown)


src/components/Dialog/index.web.tsx
```diff
@@ -113,17 +110,15 @@ export function Outer({
                   gtMobile ? a.p_lg : a.p_md,
                   {overflowY: 'auto'},
                 ]}>
-                {isVisible && (
-                  <Animated.View
```
haileyok, line 117: Do we need this `isVisible`? See above

src/components/Dialog/index.web.tsx
```diff
@@ -33,20 +33,17 @@ export function Outer({
...
     setDialogIsOpen(control.id, true)
+    setIsOpen(true)
   }, [setIsOpen, setDialogIsOpen, control.id])
 
   const close = React.useCallback<DialogControlProps['close']>(
     cb => {
       setDialogIsOpen(control.id, false)
-      setIsVisible(false)
       setIsOpen(false)
-      setIsVisible(true)
```
haileyok, line 49: `isVisible` has a default value of `true` and here we never actually would have changed this to `false`.

haileyok: I also can't see a visual difference after removing this in the web version, so unsure if we need it. We're conditionally rendering in the dialog once `isOpen` changes to true, so we still get the `Animated.View` `entering` animations.

src/components/Dialog/index.tsx
```diff
@@ -105,19 +104,34 @@ export function Outer({
...
+    if (typeof cb === 'function') {
       closeCallback.current = cb
     }
-    // initiates a close animation, the actual "close" happens in `onCloseInner`
     sheet.current?.close()
   }, [])
 
+  // This is the actual thing we are doing once we "confirm" the dialog. We want the dialog's close animation to
+  // happen before we run this. It is passed to the `BottomSheet` component.
+  const onCloseAnimationComplete = React.useCallback(() => {
```
haileyok, line 117: Calling this `onCloseAnimationComplete` I think gives us a better sense of the purpose of this function. Too many `onClose`s gets a bit confusing I think.

src/components/Dialog/index.tsx
```diff
@@ -105,19 +104,34 @@ export function Outer({
...
-        logger.error(
-          `Dialog close was passed multiple callbacks, you shouldn't do that`,
-        )
-      }
+    if (typeof cb === 'function') {
       closeCallback.current = cb
     }
-    // initiates a close animation, the actual "close" happens in `onCloseInner`
     sheet.current?.close()
   }, [])
```
haileyok, line 113: Do we actually care if the callback changes? Or can it? From what I can tell if we call `close` then the dialog's lifetime is going to end and there won't be another chance for this to get called. Unless we are worried about `control.close()` getting called multiple times for some reason?

src/components/Dialog/index.web.tsx
```diff
@@ -135,7 +130,7 @@ export function Outer({
                       minHeight: web('calc(90vh - 36px)') || undefined,
                     },
                   ]}>
-                  {isVisible ? children : null}
```
haileyok, line 138: Same here

src/components/Dialog/index.tsx
gaearon: In principle we should take multiple calls into account since there's no mechanism to prevent those, and if they happen in different parts of the codebase (which at first might be independent and then due to another change might start firing together), each of them may have an expectation that the callback will fire.

src/components/Dialog/index.web.tsx
gaearon: State calls during events should be batched so I'm not sure there'd be a difference either way.

src/components/Dialog/index.tsx
haileyok: Okay. Perhaps I am misunderstanding what is going on outside of here, but I would imagine that we wouldn't actually hit this case. Let me look some more. I'd rather that just not be a possibility at all.

src/components/Dialog/index.tsx
```diff
@@ -105,19 +105,36 @@ export function Outer({
...
-      }
-      closeCallback.current = cb
+    if (typeof cb === 'function') {
+      closeCallbacks.current?.push(cb)
+    }
+
+    if (!isClosing.current) {
+      sheet.current?.close()
+      isClosing.current = true
     }
```
haileyok, line 117: Queue callbacks so that we call each one, incase there are multiple.

src/components/Dialog/index.tsx
```diff
@@ -105,19 +105,36 @@ export function Outer({
...
   const close = React.useCallback<DialogControlProps['close']>(cb => {
-    if (cb && typeof cb === 'function') {
-      if (closeCallback.current) {
-        logger.error(
-          `Dialog close was passed multiple callbacks, you shouldn't do that`,
-        )
-      }
-      closeCallback.current = cb
+    if (typeof cb === 'function') {
+      closeCallbacks.current?.push(cb)
```
gaearon, line 112: Why is it nullable?

src/components/Dialog/index.tsx
```diff
@@ -105,19 +105,36 @@ export function Outer({
...
+    }
+
+    if (!isClosing.current) {
+      sheet.current?.close()
+      isClosing.current = true
```
gaearon, line 117: nit: might want to move this line up in case close() throws for whatever reason

src/components/Dialog/index.tsx
```diff
@@ -105,19 +105,36 @@ export function Outer({
...
+    // This removes the dialog from our list of stored dialogs. Not super necessary on iOS, but on Android this
+    // tells us that we need to toggle the accessibility overlay setting
+    setDialogIsOpen(control.id, false)
+    setOpenIndex(-1)
+
+    if (closeCallbacks.current && closeCallbacks.current.length > 0) {
+      for (const cb of closeCallbacks.current) {
+        try {
+          cb()
+        } catch {}
```
gaearon: should not silence them — need to store at least the first one and rethrow at the end of the function

src/components/Dialog/index.tsx
haileyok: oops

src/components/Dialog/index.tsx
```diff
@@ -96,28 +97,55 @@ export function Outer({
...
+  // happen before we run this. It is passed to the `BottomSheet` component.
+  const onCloseAnimationComplete = React.useCallback(() => {
+    // This removes the dialog from our list of stored dialogs. Not super necessary on iOS, but on Android this
+    // tells us that we need to toggle the accessibility overlay setting
+    setDialogIsOpen(control.id, false)
+    setOpenIndex(-1)
+
+    callQueuedCallbacks()
+    onClose?.()
+    isClosing.current = false
```
gaearon, line 146: Set important flags before calling into external code so that our internal state doesn't get corrupted by external code throwing.

src/components/Dialog/index.web.tsx
```diff
@@ -33,24 +33,25 @@ export function Outer({
...
-      setIsVisible(true)
 
       try {
         if (cb && typeof cb === 'function') {
-          cb()
+          // This timeout ensures that the callback runs at the same time as it would on native. I.e.
+          // console.log('Step 1') -> close(() => console.log('Step 3')) -> console.log('Step 2')
+          // This should always output 'Step 1', 'Step 2', 'Step 3', but without the timeout it would output
+          // 'Step 1', 'Step 3', 'Step 2'.
+          setTimeout(cb)
```
haileyok, line 54: good callout on this one, `setTimeout` gives us the expected result.

src/components/Dialog/index.web.tsx
estrattonbailey: Oh ya know what, this is leftovers from an exit animation we removed, def safe to remove

src/components/Dialog/index.tsx
```diff
@@ -96,28 +96,49 @@ export function Outer({
...
+    closeCallbacks.current = []
+  }, [])
+
   const open = React.useCallback<DialogControlProps['open']>(
     ({index} = {}) => {
       setDialogIsOpen(control.id, true)
       // can be set to any index of `snapPoints`, but `0` is the first i.e. "open"
       setOpenIndex(index || 0)
+
+      callQueuedCallbacks()
```
estrattonbailey, line 117: Do we want to call all close callbacks here, or maybe just reset the an empty array on re-open?

haileyok: I think calling all of the old ones is the best move. I.e:

- Press save
- Close animation begins
- Somehow the dialog gets opened again before the animation ends
- ???

I think i'd expect the first thing I did to still complete personally.

PR #3450, opened 2024-04-08
[Embeds] Embed for single post

mozzius:
# Stacked on #3448 

This PR adds support for embedding a single post. It takes an AT URI in the search parameters, and supports rich text and embeds

Screenshot 2024-04-09 at 00 00 55

## Test plan

1. Start the dev server (`cd bskyembed` and `yarn dev`)
2. Find the AT URI of a post you want to test (I do this by opening the network tab, clicking on a post, then searching for getPostThread - it's passed as a query string)
3. Add `?uri=<at-uri>` to the query string (i.e. `http://localhost:5173/?uri=at://did:plc:wqwdm7xgynf7amclnjfhvncr/app.bsky.feed.post/3kpvyeesjry2o`)
4. Test all kinds of posts you can think of - different combinations of embeds, complicated facets etc

PR #3448, opened 2024-04-08
[Embeds] Create vite project and add to build pipeline 

mozzius:
This PR adds a vite/preact project in the `bskyembed` directory. It adds a new `build-embed` script which should create the build output and pass it to `bskyweb`, where a new route will serve the files via `/embed/post.html`

End result:

Screenshot 2024-04-08 at 20 12 52

## Test plan

`cd bskyembed`
`yarn dev`

PR #3454, opened 2024-04-09
[Embeds] Add CTA and add general polish - input needed!

mozzius:
Stacked on #3448 
and #3450 

We need to add some sort of visual appeal to the embeds - this is a very open question.

This is one approach, but happy to rework this PR in a completely different direction.

Screenshot 2024-04-09 at 17 46 45


pfrazee (approved):
Just reviewing the visual, looks great to me

haileyok (approved):
I really like the visual look! IMO feeling a bit Twittery isn't actually a problem for us at all.

PR #3455, opened 2024-04-09
[Statsig] Instrument feed display

gaearon:
Adds an event that fires when a feed is displayed, along with the index and the reason. It was a bit tricky to figure out the right place to thread these through. This is the minimally invasive way I've found.

## Test Plan

Add `console.log` to `logEvent` in `statsig.tsx`.

- Reload
- Use tabbar
- Use swipe
- On desktop, use right nav
- Switch between tabs

You should see events firing according to the schema.

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -2,8 +2,9 @@ import React from 'react'
 import {ActivityIndicator, AppState, StyleSheet, View} from 'react-native'
 import {useFocusEffect} from '@react-navigation/native'
 
+import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {useSetTitle} from '#/lib/hooks/useSetTitle'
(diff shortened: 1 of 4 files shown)


estrattonbailey (approved):
Works great!

src/view/screens/Home.tsx
```diff
@@ -96,6 +97,17 @@ function HomeScreenReady({
...
+  useFocusEffect(
+    useNonReactiveCallback(() => {
+      logEvent('home:feedDisplayed', {
+        index: selectedIndex,
+        feedType: selectedFeed.split('|')[0],
+        feedUrl: selectedFeed,
+        reason: 'focus',
+      })
+    }),
+  )
```
haileyok, line 109: cool. i remember you making this suggestion for these cases, glad to see it in action!

pfrazee (approved):
Tested on Web. Looks good!

PR #3456, opened 2024-04-09
[Statsig] Send prev route name

gaearon:
Sending previous only for navigation events. (We're already sending the current one with each event.)


pfrazee (approved):
Cool makes sense, didn't test

PR #3458, opened 2024-04-09
Use new button for PostThreadItem and list follow buttons

haileyok:
## Why

We changed the follow button on the profile but left the others in the app the same black/white color. This looks better and solves the issue of people thinking the white follow button is too bright.

Analytics remain the same, and the logic is simplified - we just use the single `FollowButton` component for both of these now.

I maintained the "Follow Back" gate as well, but only applied it to the follow button for `PostThread`. If we want to use it in all of these places, we just have to remove the `&& logContext === 'PostThreadItem'`

## Screenshots

Screenshot 2024-04-09 at 3 22 26 PM
Screenshot 2024-04-09 at 3 22 29 PM
Screenshot 2024-04-09 at 3 22 49 PM
Screenshot 2024-04-09 at 3 23 44 PM
Screenshot 2024-04-09 at 3 23 51 PM

## Test Plan

I went through each button to verify that they the functionality still works (so on PostThread and in the various lists of users, i.e. Liked By or Reposted By)

I also verified that events still get sent to Statsig, although I don't know how to tell if the context is properly being attached (i.e. `ProfilePostThread` as the location the follow is from). I didn't see it in network logs even on the main profile follow button, so unsure how to verify.

### DIFF
--- a/src/view/com/post-thread/PostThreadFollowBtn.tsx
+++ b/src/view/com/post-thread/PostThreadFollowBtn.tsx
@@ -1,24 +1,10 @@
 import React from 'react'
-import {View} from 'react-native'
 import {AppBskyActorDefs} from '@atproto/api'
-import {msg, Trans} from '@lingui/macro'
-import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
 
-import {logger} from '#/logger'
-import {track} from 'lib/analytics/analytics'
-import {isNative, isWeb} from 'platform/detection'
 import {Shadow, useProfileShadow} from 'state/cache/profile-shadow'
-import {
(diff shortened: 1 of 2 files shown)


haileyok:
We can also make the buttons on web smaller if they feel too big here. Not sure how I feel about them myself taking a second look at them.

gaearon:
Hmm. I'm not sure I dig the blue button in lists — it works well as a call to action when there's just one of those but IMO not so well when there's a whole lot.

haileyok:
Will leave this up for now, eventually i think we should hit these colors in some way but yea, i felt something was off at second glance too. I do think we want them to be the same as the follow button on the profile screen...i think.

#NotADesigner :)

PR #3462, opened 2024-04-10
Remove obsolete strings from `en` message catalog

surfdude29:
Supersedes #3280.

Similar to #2706, this is a cleanup PR that removes strings from the `en` message catalog that are obsolete and no longer used.


pfrazee (approved):
Thanks!

PR #3466, opened 2024-04-10
Update German translations 

imbstt:
There was some discussion in my circles on Bluesky, that the recently-added translation of "liked" → "likt" (and its variations) looked weird and uncommon. 

As far as I can tell, this originated from the last update by @surfdude29 using DeepL, although it is not *wrong*, but there's also no 100% fixed consensus about it in German grammar literature. I personally think "liked" works better than "likt" in German. 

While I was at it, I modified and added a bunch more translations, although I don't have time to complete any more currently.

### DIFF
--- a/src/locale/locales/de/messages.po
+++ b/src/locale/locales/de/messages.po
@@ -100,11 +100,11 @@ msgstr "Konto stummgeschaltet"
 #: src/components/moderation/ModerationDetailsDialog.tsx:93
 #: src/lib/moderation/useModerationCauseDescription.ts:91
 msgid "Account Muted"
-msgstr "Konto Stummgeschaltet"
+msgstr "Konto stummgeschaltet"
 
 #: src/components/moderation/ModerationDetailsDialog.tsx:82
 msgid "Account Muted by List"
-msgstr "Konto stummgeschaltet nach Liste"
+msgstr "Konto stummgeschaltet gemäß Liste"
 
 #: src/view/com/util/AccountDropdownBtn.tsx:41
 msgid "Account options"
@@ -125,7 +125,7 @@ msgstr ""
 
 #: src/view/com/profile/ProfileMenu.tsx:102
 msgid "Account unmuted"
-msgstr "Konto Stummschaltung aufgehoben"
+msgstr "Stummschaltung für Konto aufgehoben"
 
 #: src/components/dialogs/MutedWords.tsx:164
 #: src/view/com/auth/onboarding/RecommendedFeedsItem.tsx:150
@@ -480,7 +480,7 @@ msgstr "Blockierte Konten können nicht in deinen Threads antworten, dich erwäh
 
 #: src/view/com/post-thread/PostThread.tsx:313
 msgid "Blocked post."
-msgstr "Gesperrter Beitrag."
+msgstr "Blockierter Beitrag."
 
 #: src/screens/Profile/Sections/Labels.tsx:152
 msgid "Blocking does not prevent this labeler from placing labels on your account."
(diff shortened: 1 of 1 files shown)


src/locale/locales/de/messages.po
```diff
@@ -100,11 +100,11 @@ msgstr "Konto stummgeschaltet"
 #: src/components/moderation/ModerationDetailsDialog.tsx:93
 #: src/lib/moderation/useModerationCauseDescription.ts:91
 msgid "Account Muted"
-msgstr "Konto Stummgeschaltet"
+msgstr "Konto stummgeschaltet"
```
PythooonUser, line 103: Our initial thought here was that the English translation is uppercased, hence we did the same for the German translation. I guess that's open for debate? I'm fine with both versions.

src/locale/locales/de/messages.po
```diff
@@ -270,7 +270,7 @@ msgstr "Eine E-Mail wurde an deine vorherige Adresse {0} gesendet. Sie enthält
 
 #: src/lib/moderation/useReportOptions.ts:26
 msgid "An issue not included in these options"
-msgstr ""
+msgstr "Ein Problem, das hier nicht aufgelistet ist."
```
PythooonUser, line 273:
```suggestion
msgstr "Ein Problem, das hier nicht aufgelistet ist"
```

I'm not sure what the context of this translation is, but the English translation has no `.` at the end of the sentence, so maybe the actual sentence continues with something else?

src/locale/locales/de/messages.po
```diff
@@ -1350,27 +1350,27 @@ msgstr "Aufgrund der Apple-Richtlinien können Inhalte für Erwachsene erst nach
...
 
 #: src/lib/moderation/useGlobalLabelStrings.ts:43
 msgid "E.g. artistic nudes."
-msgstr ""
+msgstr "Z.B. künstlerische Nacktheit"
 
 #: src/view/com/modals/CreateOrEditList.tsx:284
 msgid "e.g. Great Posters"
-msgstr "z.B. Große Poster"
+msgstr "z.B. Großartige Poster"
```
PythooonUser, line 1373: Yeah, I guess that's what was meant? 😁

src/locale/locales/de/messages.po
```diff
@@ -1588,28 +1588,28 @@ msgstr "Alle"
...
 
 #: src/view/com/modals/DeleteAccount.tsx:230
 msgid "Exits account deletion process"
-msgstr ""
+msgstr "Verlässt den Vorgang der Accountlöschung"
 
 #: src/view/com/modals/ChangeHandle.tsx:151
 msgid "Exits handle change process"
-msgstr "Beendet den Prozess des Handle-Wechsels"
+msgstr "Verlässt den Vorgang des Handle-Wechsels"
```
PythooonUser, line 1599: 👍

src/locale/locales/de/messages.po
```diff
@@ -2380,35 +2380,35 @@ msgstr "Diesen Feed liken"
 #: src/Navigation.tsx:201
 #: src/Navigation.tsx:206
 msgid "Liked by"
-msgstr "Gelikt von"
+msgstr "Geliked von"
```
PythooonUser, line 2383: 🎉

PythooonUser (commented):
Looks good! Thanks for taking care!

src/locale/locales/de/messages.po
imbstt: I understood it as some sort of title for a dialog/modal, but titles aren't all uppercased in German like they often are in English (see e.g. Wikipedia article structures)

PythooonUser: Sure, I'm fine with your approach

pfrazee (approved):
Thanks all!!

PR #3467, opened 2024-04-10
[Statsig] Typecheck gates

gaearon:
@mozzius added `gates.ts` in #3408 to statically check the gate name. I'd prefer we do this using types so that we don't have to keep adding methods. This still gives us autocomplete etc.

I also added a lint rule that enforces we use the local (typed) version of `useGate`. This is important because ours also has a different signature — it returns the `value` directly rather than `{isLoading, value}` as the upstream one does.

## Test Plan

Gates are typechecked.

Screenshot 2024-04-10 at 16 35 56

Importing the wrong `useGate` errros.

Screenshot 2024-04-10 at 16 36 36

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,6 +31,7 @@ module.exports = {
         },
       },
     ],
+    'bsky-internal/use-typed-gates': 'error',
     'simple-import-sort/imports': [
       'warn',
       {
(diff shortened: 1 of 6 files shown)


mozzius (approved):
Looks good to me! The lint rule is great

PR #3469, opened 2024-04-10
Highlight text in composer if over the character limit

mozzius:
Native only for now. Splits the richtext in the composer on the character limit, and puts all characters over that into `<Text>` with a red background

# Test plan

- Type more than 300 characters into the composer and observe the red highlight
- Try and break it via putting multi-byte unicode characters in there. Flag emojis are a good candidate

image
image

### DIFF
--- a/src/view/com/composer/text-input/TextInput.tsx
+++ b/src/view/com/composer/text-input/TextInput.tsx
@@ -215,12 +215,14 @@ export const TextInput = forwardRef(function TextInputImpl(
   const splitter = useMemo(() => new Graphemer(), [])
   const textDecorated = useMemo(() => {
     let excess
+    let rt = richtext
     if (richtext.graphemeLength > MAX_GRAPHEME_LENGTH) {
+      rt = richtext.clone()
       const split = splitter.splitGraphemes(richtext.text)
       const validText = split.slice(0, MAX_GRAPHEME_LENGTH).join('')
       const validUnicode = new UnicodeString(validText)
       const excessText = split.slice(MAX_GRAPHEME_LENGTH).join('')
-      richtext.delete(validUnicode.utf8.byteLength, richtext.length)
+      rt.delete(validUnicode.utf8.byteLength, richtext.length)
       excess = (
         <RNText
           key="excess"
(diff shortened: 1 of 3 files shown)


mozzius:
Back in draft until we can figure out utf8/utf16 codepoints slice issue

mozzius:
Simulator Screenshot - iPhone 15 - 2024-05-09 at 14 25 12

I think I figured out how to split UTF-8 strings correctly

src/view/com/composer/text-input/TextInput.tsx
```diff
@@ -192,10 +194,32 @@ export const TextInput = forwardRef(function TextInputImpl(
...
 
+  const splitter = useMemo(() => new Graphemer(), [])
   const textDecorated = useMemo(() => {
+    let excess
+    if (richtext.graphemeLength > MAX_GRAPHEME_LENGTH) {
+      const split = splitter.splitGraphemes(richtext.text)
+      const validText = split.slice(0, MAX_GRAPHEME_LENGTH).join('')
+      const validUnicode = new UnicodeString(validText)
+      const excessText = split.slice(MAX_GRAPHEME_LENGTH).join('')
+      richtext.delete(validUnicode.utf8.byteLength, richtext.length)
```
gaearon, line 205: This looks problematic, `useMemo` should be pure and should not mutate things passed into it. Any other ideas for how to structure this code? You can create a new rich text, modify it, and return it from here.

gaearon (changes requested):
need to fix mutation

src/view/com/composer/text-input/TextInput.tsx
mozzius: Oh sure, no problem, `RichText` has a clone method which we can use here

gaearon:
on Android, this seems to hide the caret in the "over" segment. can we fix that somehow?

Screenshot 2024-05-24 at 00 52 23

mozzius:
Interesting, I'll see what can be done

mozzius:
Ok compromises, since [this React Native bug](https://github.com/facebook/react-native/issues/42093) has not been fixed:

iOS: mild red background, red underline
Android: red text, red underline

I would just do a red underline on Android, but of course `textDecorationColor` is iOS-only 🙃

Screenshot 2025-07-17 at 13 31 17
Screenshot 2025-07-17 at 13 31 46

mozzius:
Got it working on web. HOWEVER, I found that on both web and native, the link truncation behaviour really messes with it, so it's not ready for merge yet still :/

Screenshot 2025-07-17 at 14 29 54

PR #3470, opened 2024-04-10
[Statsig] Mark Testflight as staging tier

gaearon:
To make the separation clearer in the UI.

## Test Plan

Don't know how to test for sure but if we trust this constant we should be good.


haileyok (approved):
Yep this works. We're set up now to do both testflight and production builds solely on GitHub actions, so we can trust that env variable to always be set properly.

PR #3471, opened 2024-04-10
Search - only enable queries once tab is active

mozzius:
Keeps track of the active tab, and only enables the queries once the tab is active

## Test plan

Look at the network tab - tabs should only load once you click on them

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -204,7 +204,6 @@ function SearchScreenPostResults({
   const {_} = useLingui()
   const {currentAccount} = useSession()
   const [isPTR, setIsPTR] = React.useState(false)
-  const hasBeenViewed = useHasBeenTrue(active)
 
   const augmentedQuery = React.useMemo(() => {
     return augmentSearchQuery(query || '', {did: currentAccount?.did})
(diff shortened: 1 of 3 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -876,6 +914,17 @@ function scrollToTopWeb() {
   }
 }
 
+// takes a bool. returns true once the bool is true, and stays true forever
+function useHasBeenTrue(bool: boolean) {
+  const [isTrue, setIsTrue] = React.useState(bool)
+
+  React.useEffect(() => {
+    if (bool) setIsTrue(true)
```
gaearon, line 922: In general pattern like this should be avoided. See here: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

We could fix this, but I wonder if it makes sense to just leave `enabled` on only for the current one? Same approach we do in `Home` and `Profile`. I'd like to at least be consistent with behavior with those — IMO if we change it here, we should do it as a single change across all surfaces (in a separate PR).

mozzius: Yeah, it's actually fine if the other tabs' queries are disabled when inactive. Removed

gaearon (approved):
lgtm

gaearon:
needs TS fix

PR #3472, opened 2024-04-10
Update pt-BR localization to latest version

gildaswise:
I've removed all unused translations and added the few that were new from 1.76. I also fixed a translation that was being used in a button and I didn't check it after it was merged, so it wasn't making much sense.

I changed this translation a bit, removing the "as" in "Silenciar apenas as tags" to see if it properly works in the layout here: 

image


pfrazee (approved):
Awesome! Thank you @gildaswise !

PR #3473, opened 2024-04-10
Bluesky OAuth Client

haileyok:
I really want to avoid bloating the app with multiple libraries just to get the handful of functions we need for OAuth (there's no single library that offers everything). Instead, I'm using a bit of native code to perform the crypto operations that we need.


modules/expo-bluesky-oauth-client/src/oauth-client-react-native.ts
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error(LINKING_ERROR)
+  },
+
+  /**
+   * Create a private JWK for the given algorithm. The JWK should have a "use"
+   * an does not need a "kid" property.
+   *
+   * @throws if the algorithm is not supported ("ES256" must be supported)
+   */
+  async generateJwk(_algo: string): Promise<{publicKey: Key; privateKey: Key}> {
```
matthieusieben, line 27: @haileyok You don't need to generate a key pair in the native code as the `@atproto/jwk` lib already handles the conversion from private to public (and only requires a private key as input).

Also, the native code will generate a `Jwk`, not a `Key` instance. The `Key` instance needs to be instantiated using the jwk:

```suggestion 
  async generateJwk(_algo: string): Promise<Jwk> {
```

This can then be used as:

```ts
const jwk = await OauthClientReactNative.generateJwk(algo)
const kid = "random string"
return new ReactNativeKey({ ...jwk, kid })
```

modules/expo-bluesky-oauth-client/src/react-native-key.ts
```diff
@@ -0,0 +1,115 @@
...
+        // Note: OauthClientReactNative.generatePrivateJwk should throw if it
+        // doesn't support the algorithm.
+        const res = await OauthClientReactNative.generateJwk(algo)
+        const jwk = jwkValidator.parse({
+          ...res.privateKey,
+          key_ops: ['sign', 'verify'],
+          kid,
+        })
+        const use = jwk.use || 'sig'
+        return new ReactNativeKey(jwkValidator.parse({...jwk, use, kid}))
```
matthieusieben, line 29: note that `key_ops` and `use` are redundant and it's up to the implementation to decide which should be used. We settled for `use`.

```suggestion
        const res = await OauthClientReactNative.generateJwk(algo)
        const jwk = res.privateKey
        const use = jwk.use || 'sig'
        return new ReactNativeKey(jwkValidator.parse({...jwk, use, kid}))
```

So the code above should work. If you still get a validation error with that, it might be because of:
1) a bug in `jwkValidator`
2) the fact that the native module adds a `key_ops` that is not compatible with `use = "sig"`.

In case of #2, it would probably be best to remove `key_ops` in the native code, and make sure that the native code indeed creates signing keys.

modules/expo-bluesky-oauth-client/index.ts
```diff
@@ -1,6 +1,4 @@
-export * from './src/crypto-subtle'
-export * from './src/jose-key'
-export * from './src/rn-crypto-key'
-export * from './src/rn-oauth-client-factory'
-export * from './src/rn-oauth-database'
-export * from './src/util.web'
+export * from './src/oauth-client-react-native'
+export * from './src/react-native-crypto-implementation'
+export * from './src/react-native-key'
+export * from './src/react-native-store-with-key'
```
matthieusieben, line 4: Note, this is probably a good place to expose the "universal" module by having an `index.web.ts` that exposes the same interfaces/classes (but from `@atproto/oauth-client-browser`) ?

modules/expo-bluesky-oauth-client/src/oauth-client-react-native.ts
haileyok: Thanks for that tip! And yea, I'm just pushing up some code right now with the skeleton you have over in the oauth feature branch 👍

modules/expo-bluesky-oauth-client/src/react-native-key.ts
haileyok: Yea, I ran into something there. It looks like the keys are capable of signing (tested the outputs in an external validator as well as ran `verifyJwt` on the output JWT). Let me work off what you have there and dig deeper to see what might be the problem 👍

haileyok: Wonderful, this works! Thanks!

modules/expo-bluesky-oauth-client/src/react-native-oauth-client-factory.native.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export type RNOAuthClientOptions = {
+  responseMode?: OAuthResponseMode
+  responseType?: OAuthResponseType
+  clientMetadata: OAuthClientMetadata
+  fetch?: Fetch
+  crypto?: any
+}
+
+export class RNOAuthClientFactory extends OAuthClientFactory {
```
matthieusieben, line 25: Have a look at  `react-native-oauth-client-factory.ts` from the skeleton:
1) It contains a `signIn()` method that shows how this is expected to word
2) It uses a slightly different implementation for the stores that allows proper serialization of the `dpopKey` ;-)

PR #3482, opened 2024-04-11
Improve Android haptic, offer toggle for haptics in the app

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/3484

### Android Patch

We recently migrated to `expo-haptics`, mostly in preparation for using Fabric. There's a difference however in how the two libraries run an iOS "light" haptic config on Android.

The previous library used the `Vibration` API solely, which does not
have any configuration for intensity of vibration. The `Vibration` API has also been deprecated since SDK 26. See:
https://github.com/mkuczera/react-native-haptic-feedback/blob/master/android/src/main/java/com/mkuczera/vibrateFactory/VibrateWithDuration.java

Expo Haptics is using `VibrationManager` API on SDK >= 31. See: https://github.com/expo/expo/blob/main/packages/expo-haptics/android/src/main/java/expo/modules/haptics/HapticsModule.kt#L44
The timing and intensity of their haptic configurations though differs greatly from the original implementation. This
patch uses the new `VibrationManager` API to create the same vibration that would have been seen in the deprecated
`Vibration` API.

We could just revert this change if we'd rather not use this patch. However, given the deprecated API and the preference I have in using Expo-supported modules as we move into the Fabric migration, I would suggest this change. Given that the haptics were actually not even present at all on Android (many people say that the haptics are actually NEW, not just changed), that feels like more reason to move away from the deprecated API.

### Toggle

This issue has actually re-raised a lot of people's complaints about haptics in general. Even some iOS users would rather be able to toggle these off. It also isn't as common a behavior on Android, so some users there want to turn them off anyway. To that end, I've added a haptic toggle preference. The default is still to use haptics, but the user can disable them in the settings screen.

## Test Plan

To be honest, I can not tell much of a difference between the two haptics on Android. This is probably because of varying haptic generators in different devices. Regardless, this change is relatively straight forward so I am not worried about that.

Also I have tested the various places that we have haptics in the app, and verified that the toggle does indeed disable haptics when turned on.

### DIFF
--- a/src/lib/haptics.ts
+++ b/src/lib/haptics.ts
@@ -13,26 +13,30 @@ const hapticImpact: ImpactFeedbackStyle = isIOS
   : ImpactFeedbackStyle.Light // Users said the medium impact was too strong on Android; see APP-537s
 
 export class Haptics {
-  static default() {
-    if (isWeb) {
+  static default(enabled: boolean) {
+    if (!enabled || isWeb) {
       return
     }
+
     impactAsync(hapticImpact)
   }
-  static impact(type: ImpactFeedbackStyle = hapticImpact) {
-    if (isWeb) {
+  static impact(type: ImpactFeedbackStyle = hapticImpact, enabled: boolean) {
+    if (!enabled || isWeb) {
       return
     }
     impactAsync(type)
   }
-  static selection() {
-    if (isWeb) {
+  static selection(enabled: boolean) {
+    if (!enabled || isWeb) {
       return
     }
     selectionAsync()
   }
-  static notification = (type: 'success' | 'warning' | 'error') => {
-    if (isWeb) {
+  static notification = (
+    type: 'success' | 'warning' | 'error',
+    enabled: boolean,
+  ) => {
+    if (!enabled || isWeb) {
       return
     }
     switch (type) {
--- a/src/lib/haptics.ts
+++ b/src/lib/haptics.ts
@@ -1,14 +1,20 @@
+import React from 'react'
 import {impactAsync, ImpactFeedbackStyle} from 'expo-haptics'
 
 import {isIOS, isWeb} from 'platform/detection'
+import {useHapticsDisabled} from 'state/preferences/disable-haptics'
 
 const hapticImpact: ImpactFeedbackStyle = isIOS
   ? ImpactFeedbackStyle.Medium
   : ImpactFeedbackStyle.Light // Users said the medium impact was too strong on Android; see APP-537s
 
-export function playHaptic(disabled: boolean) {
-  if (disabled || isWeb) {
-    return
-  }
-  impactAsync(hapticImpact)
+export function useHaptics() {
+  const isHapticsDisabled = useHapticsDisabled()
+
+  return React.useCallback(() => {
+    if (isHapticsDisabled || isWeb) {
+      return
+    }
+    impactAsync(hapticImpact)
+  }, [isHapticsDisabled])
 }
--- a/src/lib/haptics.ts
+++ b/src/lib/haptics.ts
@@ -1,10 +1,4 @@
-import {
-  impactAsync,
-  ImpactFeedbackStyle,
-  notificationAsync,
-  NotificationFeedbackType,
-  selectionAsync,
-} from 'expo-haptics'
+import {impactAsync, ImpactFeedbackStyle} from 'expo-haptics'
 
 import {isIOS, isWeb} from 'platform/detection'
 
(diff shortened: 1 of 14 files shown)


src/lib/haptics.ts
```diff
@@ -13,26 +13,30 @@ const hapticImpact: ImpactFeedbackStyle = isIOS
...
+  static default(disabled: boolean) {
+    if (disabled || isWeb) {
       return
     }
+
     impactAsync(hapticImpact)
   }
-  static impact(type: ImpactFeedbackStyle = hapticImpact) {
-    if (isWeb) {
+  static impact(type: ImpactFeedbackStyle = hapticImpact, disabled: boolean) {
```
gaearon, line 23: Where are these being used? I can't find callsites. Are these here just in case to match the Expo API?

src/lib/haptics.ts
```diff
@@ -13,26 +13,30 @@ const hapticImpact: ImpactFeedbackStyle = isIOS
...
+    if (disabled || isWeb) {
       return
     }
     selectionAsync()
   }
-  static notification = (type: 'success' | 'warning' | 'error') => {
-    if (isWeb) {
+  static notification = (
+    type: 'success' | 'warning' | 'error',
+    enabled: boolean,
```
gaearon, line 37: Why do we have `enabled` here but `disabled` for others?

src/lib/haptics.ts
```diff
@@ -13,26 +13,30 @@ const hapticImpact: ImpactFeedbackStyle = isIOS
   : ImpactFeedbackStyle.Light // Users said the medium impact was too strong on Android; see APP-537s
```
gaearon, line 14: Could add

```js
export function useHaptics() {
  const isHapticsDisabled = useHapticsDisabled()
  return useCallback(() => {
    Haptics.default(isHapticsDisabled)
  }, [isHapticsDisabled])
}
```

Then move the callsites to

```js
const haptics = useHaptics()

// ...
haptics()
```

src/view/screens/ProfileFeed.tsx
```diff
@@ -221,18 +223,19 @@ export function ProfileFeedScreenInner({
       logger.error('Failed up update feeds', {message: err})
     }
   }, [
-    feedInfo,
+    isHapticsDisabled,
     isSaved,
-    saveFeed,
     removeFeed,
-    resetSaveFeed,
+    feedInfo.uri,
```
gaearon, line 229: Unintended change? I suppose it doesn't matter but I'm not sure

src/view/screens/ProfileFeed.tsx
```diff
@@ -245,7 +248,16 @@ export function ProfileFeedScreenInner({
       Toast.show(_(msg`There was an issue contacting the server`))
       logger.error('Failed to toggle pinned feed', {message: e})
     }
-  }, [isPinned, feedInfo, pinFeed, unpinFeed, resetPinFeed, resetUnpinFeed, _])
+  }, [
+    isHapticsDisabled,
+    isPinned,
+    unpinFeed,
+    feedInfo.uri,
```
gaearon, line 255: Same: unintended or intentional

src/view/screens/ProfileFeed.tsx
```diff
@@ -546,7 +559,17 @@ function AboutSection({
...
     }
-  }, [likeUri, isLiked, feedInfo, likeFeed, unlikeFeed, track, _])
+  }, [
+    isHapticsDisabled,
+    isLiked,
+    likeUri,
+    unlikeFeed,
+    track,
+    likeFeed,
+    feedInfo.uri,
```
gaearon, line 569: Same q

gaearon (approved):
readthrough looks good. small nits and suggestions. have not tried on device.

src/lib/haptics.ts
haileyok: I assume just to match, yea. That's how it was set up prior so I maintained the API but I'm fine with removing these.

src/view/screens/ProfileFeed.tsx
haileyok: Linter probably picked them up when I added to the deps. Is probably fine, I'm fine with reverting the lint changes though if we want.

haileyok: Same to the others below.

src/lib/haptics.ts
haileyok: Simplified this down to just `playHaptic(disabled: boolean)`. I don't think we have any intentions of using these other options, those originated from a while back and it does not look like they were ever used from what I can tell.

src/lib/haptics.ts
haileyok: forgot to invert this one, but removed anyway since we don't use it.

src/lib/haptics.ts
haileyok: nice.

PR #3285, opened 2024-03-20
Improve localization marks

quiple:
```ts
<Text style={[t.atoms.text, a.text_md, a.leading_snug, a.mt_lg]}>
  {modcause.source.type === 'user' ? (
    <Trans>This label was applied by the author.</Trans>
  ) : (
    <Trans>
      This label was applied by{' '}
      <InlineLinkText
        to={makeProfileLink({did: modcause.label.src, handle: ''})}
        onPress={() => control.close()}
        style={a.text_md}>
        {desc.source}
      </InlineLinkText>
      .
    </Trans>
  )}
</Text>
```

````diff
  <Trans>
    <Text type="xl-medium" style={pal.text}>
      {formatCountShortOnly(profile?.followersCount ?? 0)}
    </Text>{' '}
    <Plural
      value={profile?.followersCount || 0}
      one="follower"
      other="followers"
-   />{' '}
-   &middot;{' '}
- </Trans>
+   />
+ </Trans>{' '}
+ &middot;{' '}
````

Changes the position of `<Trans>` for easier localization.

---

```ts
<Trans>
  Delete Account{' '}
  <Text type="title-xl" style={[pal.text, s.bold]}>
    "
  </Text>
  <Text
    type="title-xl"
    numberOfLines={1}
    style={[
      isMobile ? styles.titleMobile : styles.titleDesktop,
      pal.text,
      s.bold,
    ]}>
    {currentAccount?.handle}
  </Text>
  <Text type="title-xl" style={[pal.text, s.bold]}>
    "
  </Text>
</Trans>
```

```ts
<Trans>
  Step {state.activeStep + 1} of{' '}
  {state.serviceDescription &&
  !state.serviceDescription.phoneVerificationRequired
    ? '2'
    : '3'}
</Trans>
```

Changes the position of `<Trans>` for localization in languages with reversed word order.

---

```diff
- <Text type="md-bold" style={[pal.textLight, s.mr5]}>
-   <Trans>Not Applicable.</Trans>
- </Text>
  <Trans>
+   <Text type="md-bold" style={[pal.textLight]}>
+     Not Applicable.
+   </Text>{' '}
    This warning is only available for posts with media attached.
  </Trans>
```

Replaces the right margin with a whitespace and changes the position of `<Trans>` for easier localization.

---

```diff
  label={
-   profileShadow.viewer?.following ? _('Following') : _('Follow')
+   profileShadow.viewer?.following ? _(msg`Following`) : _(msg`Follow`)
  }
```

```diff
[...]

PR #3491, opened 2024-04-12
Update Korean localization

quiple:
Update the Korean localization of added and changed strings on the fly.

I think it would be more helpful for localization if #3285 was merged.


pfrazee:
I'll try to get a look at that PR soonish

pfrazee (approved):
LGTM! Thank you quiple!

PR #3492, opened 2024-04-12
Dont apply the content-language filter if it will remove all content

pfrazee:
Right now, if you land on a feed that gives you content entirely in a different language than you've selected, it ends up showing the feed as empty. This is pretty confusing, especially when you're going to a feed like "Brazilian Supercluster."

This modifies the language filter to just give the original results-page if the filter would have removed everything in the page.


estrattonbailey (approved):
LGTM

gaearon (approved):

PR #3496, opened 2024-04-12
Enable updates for `production` behind `receive_updates` gate

haileyok:
## Why

We have been testing updates internally for a few weeks ago. So far we have not seen any issues with this on either platform amongst any of our team.

This PR will enable updates in production behind a Statsig feature gate. The update flow will be the same as it currently is in testflight, with two differences:

- Users will not receive an "Update Available" alert
- The update will be downloaded in the background upon app launch and will be installed automatically either:
	- After an app force quite and re-open
	- After fifteen minutes of inactivity in the background

The app will also check for updates every fifteen minutes while in the foreground. If an update becomes available, it will be downloaded in the background, and will be installed as described above.

However, as long as the `receive_update` gate returns `false`, none of this will happen. Using the gate will allow us to

- Only use this feature if we end up needing to
- Incrementally begin testing load against our updates service (i.e., ungate 5% at a time to start performing automatic checks)

## Test Plan

I have done some testing to ensure that manifest fetch failures do not cause problems. I.e., I have blocked `updates.bsky.app` and opened the app. I've also verified that the gate prevents me from checking for updates when set to fail.

### DIFF
--- a/src/lib/hooks/useOTAUpdates.web.ts
+++ b/src/lib/hooks/useOTAUpdates.web.ts
@@ -1,0 +1,1 @@
+export function useOTAUpdates() {}
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -13,6 +13,7 @@ import {UsePreferencesQueryResponse} from '#/state/queries/preferences/types'
 import {useSession} from '#/state/session'
 import {useSetDrawerSwipeDisabled, useSetMinimalShellMode} from '#/state/shell'
 import {useSelectedFeed, useSetSelectedFeed} from '#/state/shell/selected-feed'
+import {useOTAUpdates} from 'lib/hooks/useOTAUpdates'
 import {HomeTabNavigatorParams, NativeStackScreenProps} from 'lib/routes/types'
 import {FeedPage} from 'view/com/feeds/FeedPage'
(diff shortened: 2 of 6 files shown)


src/lib/hooks/useOTAUpdates.web.ts
```diff
@@ -0,0 +1 @@
+export function useOTAUpdates() {}
```
haileyok, line 1: Since we moved this to `Home.tsx`, this serves as a placeholder for web.

src/view/screens/Home.tsx
```diff
@@ -51,6 +52,8 @@ function HomeScreenReady({
   preferences: UsePreferencesQueryResponse
   pinnedFeedInfos: FeedSourceInfo[]
 }) {
+  useOTAUpdates()
```
haileyok, line 55: Moving this to `Home.tsx` because Statsig has not loaded if we call this hook immediately in `App.native.tsx`. Running it in `HomeScreenReady` ensures that it has loaded.

gaearon (approved):
code lgtm but i'd like Paul to make the call on whether we push this into the current release

haileyok:
Sounds good. We can touch base w/ Paul today during sync.

haileyok:
Don't merge this until we update the cert on Expo.

PR #3498, opened 2024-04-12
Implement FeedFeedback API

pfrazee:
This PR introduces a system for feedback to make it's way back to feeds. This feedback lets the feed know which posts have been seen (so it can stop showing the same stuff twice), requests for more/less of a kind of post, and interaction signals to help train the feed's model.

The permissions model for this is going to be important. For now, we're hard-coding it to the Discover feed only. Once we introduce a permission framework for asking user permission, we'll be able to send this feedback to other feeds as well.

Most of this is invisible, except for the show more / less menu items:

CleanShot 2024-04-19 at 12 39 51@2x

### Tasks

- [x] Add visibility tracking API on Web
- [x] Track seen events
- [x] Track all other interaction events
- [x] Add show more / less menu
- [x] Batch upload periodically


src/view/com/util/List.web.tsx
```diff
@@ -230,30 +243,119 @@ function useResizeObserver(
...
+  onViewableItemsChanged,
+  children,
+}: {
+  itemKey: string
+  item: ItemT
+  index: number
+  onViewableItemsChanged: FlatListProps<ItemT>['onViewableItemsChanged']
+  children: React.ReactNode
+}): React.ReactNode {
+  if (!onViewableItemsChanged) {
```
gaearon, line 293: Potential hazard if this is supplied conditionally. Might be better to just always wrap.

src/state/feed-feedback.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export function Provider({
+  feed,
+  children,
+}: React.PropsWithChildren<{feed: FeedDescriptor}>) {
+  const enabled = isDiscoverFeed(feed)
+  const queue = React.useRef<Set<string>>(new Set())
+
+  const sendToFeed = React.useRef(
+    debounce(
```
gaearon, line 35: Nit: this will create a new debounced function on every render (and discard it). There two nicer ways are:

- `useRef()` and then immediately `if (!ref.current) { ref.current = ... }`
- Or just `useState` with unused setter and lazy initializer

src/view/com/util/List.web.tsx
```diff
@@ -242,12 +250,47 @@ let Row = function RowImpl<ItemT>({
...
+      })
+    },
+  )
+
+  React.useEffect(() => {
+    if (!onItemSeen) {
+      return
+    }
+    const observer = new IntersectionObserver(handleIntersection)
+    const tail: Element | null = tailRef.current!
```
gaearon, line 281: nit: I only called this `tail` because it was literally tracking the tail of the list, this is more like `row` and `rowRef`.

gaearon (approved):
react parts make sense to me, didn't take a super close look at the product logic.

whyrusleeping:
this seems to be working, can we ship it behind a statsig gate for now?

whyrusleeping:
Actually, before shipping, we should make sure that when this api reports a post as seen, the user has definitely seen it.
If i scroll by a post quickly, or a post is partially off screen or something i dont want it reported as 'seen'

pfrazee:
@whyrusleeping done. I now make sure the post is substantively visible for at least 3 seconds

src/state/feed-feedback.tsx
```diff
@@ -0,0 +1,145 @@
...
+  flushAndReset: () => {},
+})
+
+export function useFeedFeedback(feed: FeedDescriptor, hasSession: boolean) {
+  const {getAgent} = useAgent()
+  const enabled = isDiscoverFeed(feed) && hasSession
+  const queue = React.useRef<Set<string>>(new Set())
+  const history = React.useRef<Set<string>>(new Set())
+
+  const [sendToFeed] = React.useState(() =>
```
gaearon, line 31: This is a subtle pitfall because if you use any state or props inside this function (or if you call any method that uses those), they'll be stale. Since the initial state isn't reactive.

I think I'll probably restructure this, need to think how.

vasi:
Hi! Sorry if it's rude to comment on an old PR, but wondering if there's somewhere I can follow along with the plans to generalize this to custom feeds. It seems necessary for any good algorithmic feeds.

PR #3499, opened 2024-04-12
Use getSuggestions endpoint behind the gate

gaearon:
When `use_new_suggestions_endpoint` is `true`, the Search tab uses the `getSuggestions` endpoint instead of what we're doing now (which hits `getSuggestedFollowsByActor` and then does the same for a few of them). The old behavior is preserved when the flag is off (which it currently is for everyone).

## Test Plan

Go to search tab. Observe roughly the same set (or at least after a few refreshes) as in prod.

Then pass yourself in overrides for the `use_new_suggestions_endpoint` gate.

Observe a difference set of results. (They're not currently very relevant but we'll be improving that on the backend.)

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -224,10 +224,11 @@ function SearchScreenSuggestedFollows() {
       keyExtractor={item => item.did}
       // @ts-ignore web only -prf
       desktopFixedHeight
-      contentContainerStyle={{paddingBottom: 1200}}
+      contentContainerStyle={{paddingBottom: 200}}
       keyboardShouldPersistTaps="handled"
       keyboardDismissMode="on-drag"
       onEndReached={onEndReached}
+      onEndReachedThreshold={2}
     />
   ) : (
     <CenteredView sideBorders style={[pal.border, s.hContentRegion]}>
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -118,8 +118,7 @@ function EmptyState({message, error}: {message: string; error?: string}) {
   )
 }
 
-function SearchScreenSuggestedFollows() {
-  const pal = usePalette('default')
+function useSuggestedFollows() {
   const {currentAccount} = useSession()
   const [suggestions, setSuggestions] = React.useState<
     AppBskyActorDefs.ProfileViewBasic[]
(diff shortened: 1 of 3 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -162,6 +167,56 @@ function SearchScreenSuggestedFollows() {
...
+      await fetchNextPage()
+    } catch (err) {
+      logger.error('Failed to load more suggested follows', {message: err})
+    }
+  }, [isFetchingNextPage, hasNextPage, isError, fetchNextPage])
+
+  const items: AppBskyActorDefs.ProfileViewBasic[] = []
+  if (suggestions) {
+    // Currently the responses contain duplicate items.
+    // Needs to be fixed on backend, but let's dedupe to be safe.
```
gaearon, line 197: I've added support for pagination to this view but it isn't currently noticeable because the `V1` version doesn't paginate, while `V2` does but most pages contain duplicates so it's not noticeable. It should work if the endpoint gets fixed though.

src/view/screens/Search/Search.tsx
```diff
@@ -172,6 +227,7 @@ function SearchScreenSuggestedFollows() {
       contentContainerStyle={{paddingBottom: 1200}}
```
haileyok, line 227: Need to remove this padding (or rather probably set it to like 200px or so), otherwise  you scroll down way too far.

src/view/screens/Search/Search.tsx
```diff
@@ -172,6 +227,7 @@ function SearchScreenSuggestedFollows() {
...
       keyboardShouldPersistTaps="handled"
       keyboardDismissMode="on-drag"
```
haileyok, line 229: Also if we add `onEndReachedThreshold={2}` (I think 3 might be a bit too soon since there are not that many items per page, unless we want to add more) we shouldn't hit the bottom as much.

src/view/screens/Search/Search.tsx
gaearon: yeah didn't want to touch it cause not sure why it's there

haileyok: Looks like it is there because we used to just show the keyboard overtop of the suggestions. now though (i think? it does for me anyway) we show a blank screen or list previous searches (only up to five) so we don't need to be able to scroll below the keyboard.

haileyok (approved):
so much better at the suggestions now

gaearon:
to be clear, the endpoint we're hitting right now serves old results (which may still be better in some cases than our current ones) but the plan is to also update that endpoint to be better — so we'll try that later!

PR #3500, opened 2024-04-12
add `likeCount` etal. to `embedViewRecordToPostView`

haileyok:
## Why

Fixes the QT navigation jumps that we experienced while investigating `FlatList` issues.

Because we are waiting on the `@atproto/api` update to land in npm, we are using `as number | undefined`. Will see if the update can land tomorrow before we cut a release, but if not I think we should ship ahead of time since this is a pretty janky experience.

## Test Plan

[screen-capture.webm](https://github.com/bluesky-social/social-app/assets/153161762/1706810a-dffb-40c8-9a55-6a050036bb29)

Tested on this post: https://bsky.app/profile/orpel.bsky.social/post/3kpvzzos2rz2u


src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -325,12 +325,6 @@ let PostThreadItemLoaded = ({
             {post.repostCount !== 0 || post.likeCount !== 0 ? (
               // Show this section unless we're *sure* it has no engagement.
               <View style={[styles.expandedInfo, pal.border]}>
-                {post.repostCount == null && post.likeCount == null && (
```
gaearon, line 328: Can this still be the case if you click through a nested chain of embeds really fast? If yes, might want to leave this in.

gaearon:
Some notes after testing this.

## Already good: Timeline -> embed

In prod, timeline -> embed already works for me with no other changes, presumably because like counts are present in the timeline response.

https://github.com/bluesky-social/social-app/assets/810438/d14c17ce-ee72-4c70-b3d2-ad5444cc85d7

So this PR doesn't change that case in either direction.

## Regression: Post (uncached) -> embed

In prod, I see post -> embed goes through these stages:

- Nothing at all :(
- Embed content and shimmer metrics
- Everything

https://github.com/bluesky-social/social-app/assets/810438/5134548a-3f3b-411e-9aee-fb711a76bce2

With this PR, it's not any better and in fact seems a bit worse:

- Nothing at all :(
- Embed content but no metrics and empty (broken) metrics
- Everything

https://github.com/bluesky-social/social-app/assets/810438/1378f008-2cf5-40b7-b00b-121647322e82

## Improvement: Post (cached) -> embed

In prod, post (cached) -> embed goes through these stages:

- Embed content and shimmer metrics
- Everything

https://github.com/bluesky-social/social-app/assets/810438/fe7e3547-8ea5-43ff-82c4-09258dab5f64

In this PR, we jump to everything instantly.

https://github.com/bluesky-social/social-app/assets/810438/e8199788-a7e7-48ad-a2d8-f3fcacb0810b

That's a clear improvement.

gaearon:
I reverted the shimmer change — this fixes the regression.

We still need to fix the `Post (uncached) -> Embed` case — it seems like we should at least be able to skip the big spinner. But ideally to have metrics too. It's out of scope of this PR but I'll add it to the checklist.

PR #3506, opened 2024-04-12
QT Jump Pt. 2 - Remove code duplication

haileyok:
## Why

https://github.com/bluesky-social/social-app/pull/3500 should have removed the need for the shimmer/placeholder portion of a post's metrics. However, in some cases - specifically when opening a QT from inside of another `PostThread` - there was still missing metrics.

We had duplicated the logic for creating a `PostView` from an embed record, which made it rather unobvious that we needed to update two places for converting an embed record.

With this, the expected result of no longer needing the shimmer is correct, so we can rm. The big spinner will be fixed in a separate PR, where we need to address caching the QP author's DID (right now we have to wait to resolve the handle to a DID which is causing the spinner).

## Test Plan

You can use the same post as before, but test in two different ways:

- From a profile or feed, as was fixed before
- From a `PostThread` - Once you get past the spinner, notice that the metrics are already available.

[screen-capture (1).webm](https://github.com/bluesky-social/social-app/assets/153161762/d2cee59a-1fbf-42d2-884e-d76bd1de04e5)

### DIFF
--- a/src/state/queries/post-thread.ts
+++ b/src/state/queries/post-thread.ts
@@ -11,7 +11,7 @@ import {getAgent} from '#/state/session'
 import {findAllPostsInQueryData as findAllPostsInNotifsQueryData} from './notifications/feed'
 import {findAllPostsInQueryData as findAllPostsInFeedQueryData} from './post-feed'
 import {precacheThreadPostProfiles} from './profile'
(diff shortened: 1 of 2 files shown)


gaearon (approved):
yay

PR #3508, opened 2024-04-12
Profile card hover preview

estrattonbailey:
Closes #2867 

Key things:
- added follow button (see new hook util)
- awaited profile prefetch so FloatingUI can correctly calculate position on first go
- did some debouncing to cut down on flicker + handle prefetch
- added press handlers to all links to close when clicked
  - added to `RichText` as well
- disabled on mobile web
- I decoupled the `Link` from the hover card itself, see usages

Re: closing on click of a link, I didn't think it was a good experience to close on any click inside e.g. the Follow button. Seemed to make sense to handle individually.

Also I chose not to enable hashtags in profiles, but we potentially could.

CleanShot 2024-04-12 at 15 26 27@2x
CleanShot 2024-04-12 at 15 26 40@2x
CleanShot 2024-04-12 at 15 26 52@2x
CleanShot 2024-04-12 at 15 27 05@2x


pfrazee (approved):
LFG

PR #3509, opened 2024-04-12
Store QP authors in the DID cache

haileyok:
## Why

This is another follow up to https://github.com/bluesky-social/social-app/pull/3500. We uncovered that quote post authors are not having their DIDs cached, which results in a full spinner when we navigate to the quoted post.

An existing solution to this would be to use `useResolveDid()`, however this results in an extra request for information that we already have. Instead, it makes more sense to update the cache with the author's handle and DID without performing a resolution, since we already have that information in the embed record.

## Testing

Using this post: http://localhost:19006/profile/kevinmkruse.bsky.social/post/3kpxbhmzcym2p

[screen-capture (2).webm](https://github.com/bluesky-social/social-app/assets/153161762/c186c063-0e58-439e-81bc-338f1415f7c2)

### DIFF
--- a/src/state/queries/resolve-uri.ts
+++ b/src/state/queries/resolve-uri.ts
@@ -49,13 +49,3 @@ export function useResolveDidQuery(didOrHandle: string | undefined) {
     enabled: !!didOrHandle,
   })
 }
-
-export function useAddDidToCache(handle: string, did: string) {
-  return useQuery<string, Error>({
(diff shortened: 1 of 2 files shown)


gaearon:
this feels like a pretty roundabout way of doing it. can we use some API to directly insert into the cache?

gaearon (approved):

PR #3511, opened 2024-04-12
rebased embedr

bnewbold:
Adds a new `embedr` (trailing r) golang web server, similar to `bskyweb`, which serves an oEmbed API endpoint, and embed HTML/assets for public embedding of Bluesky posts in other websites (aka, embedding a post in a news article or blog article).


bskyweb/cmd/embedr/server.go
```diff
@@ -0,0 +1,236 @@
...
+	e.IPExtractor = echo.ExtractIPFromXFFHeader()
+
+	// SECURITY: Do not modify without due consideration.
+	e.Use(middleware.SecureWithConfig(middleware.SecureConfig{
+		ContentTypeNosniff: "nosniff",
+		// diable XFrameOptions; we're embedding here!
+		HSTSMaxAge: 31536000, // 365 days
+		// TODO:
+		// ContentSecurityPolicy
+		// XSSProtection
```
pfrazee, line 103: Do we need to resolve these?

bnewbold: I think not, no, that code is copied from `bskyweb`

pfrazee (approved):
I think we're ready to go!

PR #3516, opened 2024-04-12
wait a sec before showing card

haileyok:
Small timeout before showing card so it doesn't show up immediately while moving cursor around.

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -62,22 +62,25 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
   const targetHovered = React.useRef(false)
   const cardHovered = React.useRef(false)
   const targetClicked = React.useRef(false)
+  const showTimeout = React.useRef<NodeJS.Timeout>()
 
   const onPointerEnterTarget = React.useCallback(() => {
-    targetHovered.current = true
+    showTimeout.current = setTimeout(async () => {
+      targetHovered.current = true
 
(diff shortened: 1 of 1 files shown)


src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -62,22 +62,25 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
...
+      } else {
+        await prefetchProfileQuery(props.did)
 
-    if (prefetchedProfile.current) {
-      // if we're navigating
-      if (targetClicked.current) return
-      setHovered(true)
-    } else {
-      prefetchProfileQuery(props.did).then(() => {
         if (targetHovered.current) {
```
haileyok, line 78: same thing, just the nested functions were getting a little out of hand after adding the timeout

src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -62,22 +62,25 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
...
-    targetHovered.current = true
+    showTimeout.current = setTimeout(async () => {
+      targetHovered.current = true
+
+      if (prefetchedProfile.current) {
+        // if we're navigating
+        if (targetClicked.current) return
+        setHovered(true)
+      } else {
+        await prefetchProfileQuery(props.did)
```
haileyok, line 76: just changed to await because the functions were getting a little out of hand after adding the timeout

pfrazee (approved):

PR #3520, opened 2024-04-13
cache fix for search post results

haileyok:
## Why

Similar to https://github.com/bluesky-social/social-app/pull/3509. We set the author's DID in the cache whenever we press on a post so that we remove all spinners.

We also add `findAllPostsInSearchQueryData()` to the `findAllPostsInQueryData()` so that we populate from search results (we already had this function ready to go, but it was just not being used).

## Test Plan

Search up something, see results

[screen-capture (4).webm](https://github.com/bluesky-social/social-app/assets/153161762/79493710-fb42-46f4-8e19-dde29e272703)


gaearon (approved):
tested on web, lgtm

PR #3521, opened 2024-04-13
Fade in animation for card

haileyok:
[screen-capture (5).webm](https://github.com/bluesky-social/social-app/assets/153161762/5d22442e-2bb8-4a69-9046-262342445363)

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -67,23 +67,23 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
   const unrenderTimeout = React.useRef<NodeJS.Timeout>()
 
   const animationStyle = {
-    animation: `${
-      isVisible ? 'avatarHoverFadeIn' : 'avatarHoverFadeOut'
-    } 0.1s forwards`,
+    animation: isVisible
+      ? 'avatarHoverFadeIn 0.3s forwards'
+      : 'avatarHoverFadeOut 0.2s forwards',
   }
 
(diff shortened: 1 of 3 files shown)


src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -63,52 +63,75 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
...
         prefetchedProfile.current = true
       }
     }, 350)
   }, [props.did, prefetchProfileQuery])
+
   const onPointerEnterCard = React.useCallback(() => {
     cardHovered.current = true
-    // if we're navigating
-    if (targetClicked.current) return
-    setHovered(true)
```
haileyok, line 89: setting state here doesn't actually do anything, the value is already true.

src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -119,19 +144,34 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
...
-            exiting={FadeOut.duration(80)}>
-            <div
-              ref={refs.setFloating}
-              style={floatingStyles}
-              onPointerEnter={onPointerEnterCard}
-              onPointerLeave={onPointerLeaveCard}>
-              <Card did={props.did} hide={hide} />
-            </div>
-          </Animated.View>
+          <style>
```
gaearon, line 149: Let's put this into the base HTML files inline

haileyok: Yea going to do that after figuring everything out 👍

gaearon (approved):

PR #3522, opened 2024-04-13
add oEmbed links to public post HTML meta headers

bnewbold:
We should probably wait on this one until we are confident the embeds are working the way we want.


pfrazee (approved):
LGTM

PR #3523, opened 2024-04-13
Cache DID and profile basic on profile card presses

haileyok:
## Why

Similar to other changes like https://github.com/bluesky-social/social-app/pull/3509. Caches the DID and profile basic view before navigating to the profile screen. This allows us to have some of the shell already visible to the user.

## Test Plan

You can see this change in action both from the suggestions screen as well as well as from the various "Liked By", "Reposted By", etc. screens.

[screen-capture (6).webm](https://github.com/bluesky-social/social-app/assets/153161762/8f4c6be4-a6a9-42a8-889b-374f6943183f)


pfrazee (approved):
LGTM

PR #3527, opened 2024-04-13
[Embeds] Align nested quote behaviour with the app

mozzius:
I had quotepost embeds only allow images and externals, however the app allows the image from a recordWithMedia. This PR changes the way we cap nesting by passing a "hideRecord" prop, so we can explicitly only hide quotes in a quote, allowing their image to appear

Current behaviour:

Screenshot 2024-04-13 at 11 12 14

New behaviour:

Screenshot 2024-04-13 at 11 08 12

App:

Screenshot 2024-04-13 at 11 08 21

### DIFF
--- a/bskyembed/src/components/embed.tsx
+++ b/bskyembed/src/components/embed.tsx
@@ -19,9 +19,11 @@ import {Link} from './link'
 export function Embed({
   content,
   labels,
+  hideRecord,
 }: {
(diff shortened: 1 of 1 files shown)


gaearon:
needs rebase / update on labels

gaearon (approved):

PR #3528, opened 2024-04-13
[Embeds] More obvious click area

gaearon:
Makes the author links only span their length instead of full width so it's easier to click on post and end up there.

Also gives cursor: pointer everywhere except post text to make it more obvious it's a link.

Gives underline to the timestamp so people can right-click it to copy url.

https://github.com/bluesky-social/social-app/assets/810438/601bfffc-63f8-4e05-ae90-c937c461cfff


mozzius (approved):
Good call, LGTM

PR #3534, opened 2024-04-13
Only close mobile web drawer if press happens on backdrop

mary-ext:
Was trying to figure out why the language picker on th drawer was near-impossible to open with touch, noticed that we're closing the drawer *even if* the tap happens inside the drawer.

Doesn't seem like the right behavior to do, this makes it so that it only happens on the backdrop instead.


gaearon (approved):

gaearon:
thank you!

PR #3535, opened 2024-04-13
Increase hit slop for web's app language picker

mary-ext:
14px seems hard to tap, so let's make it 32px instead.


pfrazee (approved):
Sure, looks good! Thanks Mary!

PR #3542, opened 2024-04-13
Add toggle in settings for hovers

haileyok:
Simple thing. Some users might find these annoying, so best to just give them a toggle to turn them off if they want to.


haileyok:
Better yet, let's just improve the damn thing.

PR #3544, opened 2024-04-13
[Session] V2

estrattonbailey:
This PR moves existing session handling to `v1.tsx` and adds the refactored version as `v2.tsx`. Then in `index.tsx` it conditionally exports, which should eventually be controlled by a flag.

Also here is some `catch` handling when switching accounts. With v1, `initSession` is swallowing some errors that it shouldn't be, and v2 fixes this. So adding this in now, even if v1 is being used, won't have any effect.


src/lib/hooks/useAccountSwitcher.ts
```diff
@@ -44,9 +45,14 @@ export function useAccountSwitcher() {
...
         }
-      } catch (e) {
-        Toast.show('Sorry! We need you to enter your password.')
+      } catch (e: any) {
+        logger.error(`switch account: selectAccount failed`, {
+          message: e.message,
+        })
         clearCurrentAccount() // back user out to login
+        setTimeout(() => {
+          Toast.show('Sorry! We need you to enter your password.')
```
surfdude29, line 54: Sorry to jump in here, would it be possible to mark this toast for localisation? And ideally the ones in L38 and L43-44 too? 🙏

estrattonbailey: Great call, thanks for the reminder!

PR #3545, opened 2024-04-13
Improve the language behaviors around the PWI

pfrazee:
- [x] Handle an overflow issue in the desktop web left nav on PWIs (longer button labels were overflowing)
- [x] When the language dropdown (used during logged out) changes, set the content and posting language in addition to the app ui language
- [x] Make sure switching the language in the PWI immediately updates the content

The goal is to get the PWI to a place where switching the language immediately has effects. This is currently blocked by some caching issues on Discover which we need to work with ops to fix. UPDATE: had to do a hackfix to workaround this (query parameter busting).

After this update:

https://github.com/bluesky-social/social-app/assets/1270099/5d2c2569-c024-41f9-bb27-8ad3a38010cb


gaearon (approved):
makes sense

PR #3547, opened 2024-04-13
Fix hover card animation with a reducer

gaearon:
I tried to be smart in https://github.com/bluesky-social/social-app/pull/3521 and it totally bit me. I think the impulse to express this as a state machine was correct, but the way I was trying to cheat my way out of React's model was wrong. Specifically, the problem is I tried to read `state` in a bunch of places but it wasn't updated by that point yet. So the logic was wrong, leading to wonky hover/unhover behavior.

This is a rewrite of that code in a way that I believe follows the React model. It might seem a bit convoluted at first but it worked correctly from the first try which is hopefully a sign that the principle is sound. The idea is to express this state machine as a reducer: a function that takes the current state (what state is the system in?) and the action (what has just happened?) and that returns the next state of the system.

One caveat is that reducers must be pure — i.e. we shouldn't kick off timeouts or do other things like this in a reducer. Luckily, we can use effects for that. The wrinkle in this solution is that we colocate the next state's effect _in the reducer_ but call it from `useEffect`. This sounds convoluted but I think the code tells a much better story, so do read through it.

I think this makes sense.

## Test Plan

https://github.com/bluesky-social/social-app/assets/810438/0e260fbe-540d-41e5-9f30-e367a1a0b83f

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -5,7 +5,6 @@ import {flip, offset, shift, size, useFloating} from '@floating-ui/react-dom'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
-import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {makeProfileLink} from '#/lib/routes/links'
 import {sanitizeDisplayName} from '#/lib/strings/display-names'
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Really interesting solution. Thanks so much for cleaning all of this up and making it readable!! 🙏 

Seeing the expected results now too, regardless of how fast I move my mouse 👍

PR #3548, opened 2024-04-13
Remove extra wrapper on notification user links

haileyok:
## Why

We added profile hovers to the users in a notification dropdown in #3508. When we did this, we introduced the usage of another wrapper, `NewLink` (the new `Link` component). Because of how flexbox and text works in React Native, sometimes these additional wrappers really mess with text layout. Here, we just need to remove the extra `View` and place the styles that were used in that view on `NewLink` instead.

## Test Plan

Open the notifications page on native and expand a notification.

Before:

Screenshot 2024-04-13 at 4 45 53 PM

After:

Screenshot 2024-04-13 at 4 45 42 PM

### DIFF
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -420,31 +420,30 @@ function ExpandedAuthorsList({
           to={makeProfileLink({
             did: author.did,
             handle: author.handle,
-          })}>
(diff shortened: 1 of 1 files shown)


gaearon (approved):
without whitespace: https://github.com/bluesky-social/social-app/pull/3548/files?w=1

confirmed this fixes it. checked ios android (fixed for those) and web (no change)

PR #3550, opened 2024-04-14
Restructure feed hover state machine code

gaearon:
Just playing with this pattern, kind of curious if it can be made maintainable for more complex pieces or it's mostly good for small things. I think I like this a bit more but I'm also not sure this structure would hold.


haileyok (approved):

pfrazee (approved):
Yeah I'm +1 this kind of pattern, pretty useful

PR #3551, opened 2024-04-14
Update Korean localization

quiple:
Update the Korean localization of added and changed strings on the fly.


pfrazee (approved):
Thanks quiple!!

PR #3552, opened 2024-04-14
Update Ukrainian translation

Dovgonosyk:
Added new strings and updated a couple of old ones that @imax9000 advised to improve.

Crowdin project for translation and discussion: https://crowdin.com/project/bsky-app-ua


pfrazee (approved):
Awesome, thanks all!

PR #3555, opened 2024-04-14
android: fix various places still using default Material Teal

kisaragi-hiu:
Some places on Android are currently using the default [Material Teal](https://m1.material.io/style/color.html#:~:text=Teal) color, including the TextInput cursor, TextInput selection, and the spinner (ActivityIndicator) --- it's quite out of place whenever it shows up.

This PR sets Expo's primaryColor to #1083fe (which is the color that defaultTheme.palette.default.brandText resolves to), then applies it as the native accent color via a plugin (because Expo doesn't apply the accent color).

Comparison:

| Before (composer & signup) | After (composer & signup) |
|--------|--------|
| before-composer | after-composer |
| before-signup | after-signup |

The `plugins/withAndroidStylesAccentColorPlugin.js` plugin is copied and modified from `plugins/withAndroidStylesWindowBackgroundPlugin.js` since that appears to be the idiomatic way of modifying Android's `res/values/styles.xml`.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -64,6 +64,8 @@ module.exports = function (config) {
       icon: './assets/icon.png',
       userInterfaceStyle: 'automatic',
       splash: SPLASH_CONFIG,
+      // hsl(211, 99%, 53%), same as palette.default.brandText
+      primaryColor: '#1083fe',
       ios: {
         supportsTablet: false,
         bundleIdentifier: 'xyz.blueskyweb.app',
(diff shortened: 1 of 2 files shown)


haileyok:
Another awesome PR! I'll test this out later this week, but looks solid! ❤️

haileyok (approved):
Tested this locally and it works great. Thank you so much!

PR #3559, opened 2024-04-15
embedr: improved HTML snippet, with links

bnewbold:
Adds post link (around datetime) and profile link (around display name and/or handle).

Also applies "sort-at" timestamp logic to the created at timestamp (https://docs.bsky.app/docs/advanced-guides/timestamps).

And removes some debug println.


gaearon (approved):
lg

PR #3560, opened 2024-04-15
Add follow button to feed item avatar

mozzius:
CleanShot 2024-05-29 at 16 21 12@2x
CleanShot 2024-05-29 at 16 21 30@2x


surfdude29:
Does this mean there would be a mini follow button as part of every avatar on every feed? Or have I misunderstood? 🤔

If I have understood correctly, I feel like this will generate an awful lot of unintended follows when people tap or click it by mistake, without actually meaning to follow the user in question.

Also, again just my personal opinion, but I feel like this will make the Bluesky interface look rather ugly and that it will be poorly received by users 😕

surfdude29:
As an alternative suggestion, what about adding a bottom sheet that opens when long-pressing the user's avatar? This could show exactly the same content as the hover cards on web do.

To follow a user, you would simply long-press the pfp and then tap the Follow button.

pfrazee:
> If I have understood correctly, I feel like this will generate an awful lot of unintended follows when people tap or click it by mistake, without actually meaning to follow the user in question.

We're talking about how to handle this properly. My personal preference would be an inline dropdown (even on mobile) that has "Go to profile" and "Follow" as options. We would need to cook up that UI behavior.

> As an alternative suggestion, what about adding a bottom sheet that opens when long-pressing the user's avatar? This could show exactly the same content as the hover cards on web do.

Part of the goal is also to introduce a visual indicator of whether you're following the person.

surfdude29:
Thank you for the clarification on what's planned and some of the thinking behind it.

pfrazee:
Sure. More than happy to hear more suggestions or reactions to it. Don't want to do something folks won't be into.

pfrazee:
A couple of tweaks:

|Smaller indicator|Confirmation modal|
|-|-|
|CleanShot 2024-04-19 at 13 33 07@2x|CleanShot 2024-04-19 at 13 34 09@2x|

We're also going to put this behind a feature gate so we can test it internally and then do an experimental rollout.

surfdude29:
> A couple of tweaks:

I think having a follow confirmation is essential, so thanks for adding that 🙏

And the smaller, less obtrusive indicator is welcome too.

> We're also going to put this behind a feature gate so we can test it internally and then do an experimental rollout.

That's a good way to proceed 👍

surfdude29:
> Your [Render](https://render.com) PR Server URL is https://social-app-pr-3560.onrender.com.

Should I be able to test this out by logging in here? When I do, it looks the same as Bluesky normally does, there's no indicator on pfps of users I don't follow.

Or maybe because it's behind a feature gate that doesn't work for this PR? 🤔

mozzius:
> Should I be able to test this out by logging in here? When I do, it looks the same as Bluesky normally does, there's no indicator on pfps of users I don't follow.
> 
> Or maybe because it's behind a feature gate that doesn't work for this PR? 🤔

This PR is iOS/Android only

surfdude29:
> > Should I be able to test this out by logging in here? When I do, it looks the same as Bluesky normally does, there's no indicator on pfps of users I don't follow.
> > Or maybe because it's behind a feature gate that doesn't work for this PR? 🤔
> 
> This PR is iOS/Android only

That would explain it 😅

mozzius:
Also added it to post thread items - only the secondary ones, since the main one will have a big "follow" button

Threaded replies do not have it because they're too small

Simulator Screenshot - iPhone 15 - 2024-04-24 at 23 28 09

src/view/com/posts/AviFollowButton.tsx
```diff
@@ -0,0 +1,115 @@
...
+        ios: {
+          name: 'person.badge.plus',
+        },
+        android: '',
+        web: ['far', 'user-plus'],
+      },
+    },
+  ]
+
+  return gate('show_avi_follow_button') && hasSession ? (
```
gaearon, line 87: note this generally should be the other way around — we should delay calling `gate` until we _have to_ make a decision. so it should be the last check

gaearon (approved):

src/view/com/posts/AviFollowButton.tsx
estrattonbailey: Ah ok yeah good call

PR #3577, opened 2024-04-16
Allow a post to contain only a quote

haileyok:
Simple, just check if there's a quote before rejecting the submission.


mozzius (approved):
lgtm

PR #3579, opened 2024-04-16
Use the existing `ExternalLinkEmbed` for embed previews in the composer

haileyok:
## Why

There's no real correlation between how our link cards look in the composer when seeing the preview and what you might see in your feed. To that end, we can use the existing `ExternalLinkEmbed` component to show previews to users.

## Test Plan

Adding various types of links to the composer gives us the expected results

## Normal Link

Screenshot 2024-04-16 at 11 17 26 AM

## Video Link

Screenshot 2024-04-16 at 11 11 08 AM

## Giphy Link

Screenshot 2024-04-16 at 11 18 52 AM

## Loading State

Screenshot 2024-04-16 at 11 20 01 AM

## Meta Extraction Error

Screenshot 2024-04-16 at 11 19 50 AM

### DIFF
--- a/src/view/com/composer/ExternalEmbed.tsx
+++ b/src/view/com/composer/ExternalEmbed.tsx
@@ -20,16 +20,19 @@ export const ExternalEmbed = ({
 }) => {
   const t = useTheme()
   const {_} = useLingui()
-  if (!link) {
-    return <View />
-  }
 
-  const linkInfo = {
-    title: link.meta?.title ?? link.uri,
-    uri: link.uri,
-    description: link.meta?.description ?? '',
-    thumb: link.localThumb?.path,
-  }
+  const linkInfo = React.useMemo(
+    () =>
+      link && {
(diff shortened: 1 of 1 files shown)


gaearon:
Seeing an add card button for an embed even after adding the card.

https://github.com/bluesky-social/social-app/assets/810438/41eb1f40-da55-41c2-aa86-6a3565e74840

gaearon:
I guess that's actually an unrelated bug already present on main, so disregard.

haileyok:
@gaearon Oops, it's RM'd in the other PR. I moved the base to main since they didn't relate, so missed that coming back in.

gaearon (approved):
works great in my testing

gaearon:
I think it happens in main just because we still technically detect the link, so we still "offer" the suggestion. I think your other PR will take care of that so let's ignore.

pfrazee (approved):
Works great, nice stuff Hailey

PR #3582, opened 2024-04-16
Always send ref_ params

gaearon:
Seems like conditionally including them isn't enough to get Statsig to show them in the UI. Also, let's decode the URL so it's more readable.

## Test Plan

Tried an existing link but changed to localhost.

Screenshot 2024-04-16 at 21 20 52

Also verified default values are empty strings.


estrattonbailey (approved):

pfrazee (approved):
Cool cool

PR #3583, opened 2024-04-16
Make ref_ always strings

gaearon:
My bad, missed it in #3582. Idk why TS didn't catch it.

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -15,8 +15,21 @@ import {useSession} from '../../state/session'
 import {LogEvents} from './events'
 import {Gate} from './gates'
 
-let refSrc: string = ''
-let refUrl: string = ''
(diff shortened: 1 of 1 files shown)


src/lib/statsig/statsig.tsx
```diff
@@ -15,8 +15,21 @@ import {useSession} from '../../state/session'
...
+    // This is the place where we can add our own stuff.
+    // Fields here have to be non-optional to be visible in the UI.
+    platform: 'ios' | 'android' | 'web'
+    refSrc: string
+    refUrl: string
+  }
+}
+
+let refSrc = ''
+let refUrl = ''
```
gaearon, line 32: This is the change.

gaearon: The rest is just hardening types.

mozzius (approved):
lgtm

PR #3584, opened 2024-04-16
Better dist strings for Sentry

haileyok:
## Why

This resolves two problems:

- We use environment variables now to determine the channel to use for Expo updates, so the `info.channel` variable is undefined. This results in the default of `development` showing up for our builds on Sentry.
- It can be difficult to determine when adoption actually started for a particular version, since we release internally sometimes a week or so prior to cutting the production release. Instead, now we will use `-tf` or `-dev` to differentiate the various internal builds. Production builds will have nothing appended.

Also did a check to see if we rely on this `info.channel` anywhere else in the codebase, and we do not 👍 

## Test Plan

We can check the logs going to Sentry in a Testflight client to ensure that the `-tf` string is being properly appended. We can also check in Sentry to see if the symbols are being uploaded correctly following this change.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -35,11 +35,6 @@ module.exports = function (config) {
    */
   const PLATFORM = process.env.EAS_BUILD_PLATFORM
 
-  const DIST_BUILD_NUMBER =
-    PLATFORM === 'android'
-      ? process.env.BSKY_ANDROID_VERSION_CODE
-      : process.env.BSKY_IOS_BUILD_NUMBER
-
   const IS_DEV = process.env.EXPO_PUBLIC_ENV === 'development'
   const IS_TESTFLIGHT = process.env.EXPO_PUBLIC_ENV === 'testflight'
   const IS_PRODUCTION = process.env.EXPO_PUBLIC_ENV === 'production'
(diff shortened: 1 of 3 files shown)


app.config.js
```diff
@@ -51,6 +51,10 @@ module.exports = function (config) {
     : undefined
   const UPDATES_ENABLED = !!UPDATES_CHANNEL
 
+  const SENTRY_DIST = `${PLATFORM}.${VERSION}.${DIST_BUILD_NUMBER ?? ''}${
```
gaearon, line 54: Can DIST_BUILD_NUMBER actually be not present? seems weird to have `1.2.`

gaearon (approved):

app.config.js
haileyok: Probably can remove it, since we don't ever release under the same 1.X more than once.

haileyok: Actually, that might not be a good idea. We upload Sentry symbols on every build, and I don't know if Sentry will overwrite old symbols.

gaearon: I mean more, does it have to have `?? ''`?

gaearon: I'd at least expect `DIST_BUILD_NUMBER ?? 0`

PR #3586, opened 2024-04-16
Scrolling while target is hovered and card is visible should hide the card

gaearon:
This is kind of subtle but pretty important.

If you're just scrolling, you don't want the hover popup to appear all the time due to your cursor position. In this PR, I'm adding a new behavior: scrolling away dismisses the popover (similar to unhovering).

There are some nuances.

Scrolling away _while the hover is on the card itself_ will not dismiss it. This is because, to end up in this situation, you must have moved the cursor on the card. So you might legitimately be trying to do something with it. 

Also, scrolling before the card is even visible (i.e. during the "might show" state) will reset it back to the hidden state. So this will restart the timer, reducing false positives.

## Implementation

I've needed to massage the abstraction a little bit to get it to be able to do this. In particular:

- The `delete state.effect` thing was unnecessary — [see here](https://twitter.com/dan_abramov2/status/1780230637397451185).
- I've made state function arguments named so that they're clearer when parameterized.
- I've given the existing arguments some default values cause they're actually always the same.
- I've added the ability for different states to store different information, i.e. taking more advantage of this being a discriminated union.
  - In particular, `'might-show'` and `'showing'` keep track of _why_ we're showing stuff in `state.reason` (either due to `'hovered-target'` or `'hovered-card'`). I didn't _have to_ literally store an action there, but seemed convenient.
  - This lets us conditionally decide whether scrolling should dismiss anything. (E.g. we want to do that in the `showing` state only when `state.reason` is `'hovered-target'`. If we're hovering a card, we don't want scrolling to dismiss it.)

Note that we're only listening to scroll when needed and not all the time. So this avoids firing lots of tiny state updates for every card on scroll. We only really need this for the one where something's currently happening.

You _might_ want to review individual commits for clarity but the whole diff isn't too bad.

## Test Plan

https://github.com/bluesky-social/social-app/assets/810438/a9c33b6f-f207-4aed-a1a6-07401b678780

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -57,8 +57,10 @@ type State = {
 
 type Action =
   | 'pressed'
-  | 'hovered'
-  | 'unhovered'
+  | 'hovered-target'
+  | 'unhovered-target'
+  | 'hovered-card'
+  | 'unhovered-card'
   | 'hovered-long-enough'
   | 'unhovered-long-enough'
   | 'finished-animating-hide'
@@ -90,7 +92,7 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
 
       // The user can kick things off by hovering a target.
       if (state.stage === 'hidden') {
-        if (action === 'hovered') {
+        if (action === 'hovered-target' || action === 'hovered-card') {
           return mightShow(SHOW_DELAY)
         }
       }
@@ -111,7 +113,7 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
 
       // We'll make a decision at the end of a grace period timeout.
       if (state.stage === 'might-show') {
-        if (action === 'unhovered') {
+        if (action === 'unhovered-target' || action === 'unhovered-card') {
           return hidden()
         }
         if (action === 'hovered-long-enough') {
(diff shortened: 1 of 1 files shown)


src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -108,33 +125,55 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
...
+        if (action === 'unhovered-target' || action === 'unhovered-card') {
+          return mightHide()
+        }
+        // Scrolling away if the hover is on the target instantly hides without a delay.
+        // If the hover is already on the card, we won't this.
+        if (
+          state.reason === 'hovered-target' &&
+          action === 'scrolled-while-showing'
+        ) {
+          return hiding()
```
haileyok, line 170: Really only mentioning this because you brought it up in the PR.

I don't feel super partial to either way of doing this - scrolling would almost always take your cursor off the card anyway so it's going to be hidden - but if you personally felt it would be nicer to disappear on any scrolling I think it would be safe. I think there might be something to letting scroll override anything - it's a different interaction than viewing the card.

Cards are never going to appear off screen and again, almost any amount of scroll will result in the card being hidden given how small they are. Up to you! (Or if I'm overlooking some reason that you'd definitely need to not hide on scroll...maybe I am)

haileyok (approved):
🙌

src/components/ProfileHoverCard/index.web.tsx
gaearon: I actually started with this, and then I felt uncomfortable that scrolling dismissed a card I was hovering. I then checked what Twitter does, and it special-cases this. So I think maybe it makes sense to special-case.

haileyok: cool, sgtm

PR #3588, opened 2024-04-16
Use `<Modal>` for Composer

mozzius:
Motivation: this lets us use native iOS sheets, which feel way more native

Currently, we use the animation API to create a modal. By using React Native's `<Modal>`, we can let React Native handle all of this, letting us drop the animation code. We can also then set `presentation` to `formSheet` so that it uses a native sheet on iOS.

`<Modal>` handles back button behaviour, so I removed this from the composer and instead added an imperative handle to trigger the cancellation behaviour, so that the `<Modal>` can trigger it.

This had the side effect of needing to rework how the composer bottom bar works - I added the library `react-native-keyboard-controller` to handle this. From my understanding, this is the best keyboard-handling library, and has worked well for me in the past. It should also let us add a more advanced keyboard-aware scrollview, so it's useful to have.

Note: this should look much better on iPad, when we add that.

https://github.com/bluesky-social/social-app/assets/10959775/ee1067ea-ab69-4684-a735-780b32eb2187

# Test plan

We need to make sure there's no regressions, especially accidental ones on web and android. Try and break the bottom bar, make the content of the composer super long etc etc.


gaearon:
Let's leave it till Monday so we can let it bake a bit. We don't know if the third-party dep is solid, so it would be good to let it bake in testflight for a week.

mozzius:
If we decide to use the `formSheet` presentation style (a la threads), revert [5fecef2](https://github.com/bluesky-social/social-app/pull/3588/commits/5fecef2ce5732e5f087c1395ff9911990d8b6dd9)

haileyok:
Bit of a margin that we will need to fix w/ the gifs.

Screenshot 2024-04-25 at 7 18 13 PM

src/alf/util/useColorModeTheme.ts
```diff
@@ -13,7 +13,11 @@ export function useColorModeTheme(): ThemeName {
   React.useLayoutEffect(() => {
     const theme = getThemeName(colorScheme, colorMode, darkTheme)
     updateDocument(theme)
-    SystemUI.setBackgroundColorAsync(getBackgroundColor(theme))
+
+    // must be default on iOS due to the formSheet modal presentation style
+    if (!isIOS) {
+      SystemUI.setBackgroundColorAsync(getBackgroundColor(theme))
+    }
```
haileyok, line 20: I forget where we landed on this, should check in a few places if this is problematic (off the top of my head I recall that modals were a problem. Might not be now though with our new ones 🤔). There's a few PRs we can go back and look at I think for context.

Worst case we can probably update this on the fly, but what you did here definitely might be a viable option.

src/view/com/composer/Composer.tsx
```diff
@@ -161,23 +174,8 @@ export const ComposePost = observer(function ComposePost({
...
-      () => {
-        onPressCancel()
-        return true
-      },
-    )
 
-    return () => {
-      backHandler.remove()
-    }
-  }, [onPressCancel])
```
haileyok, line 180: So happy we can get rid of this! it is giving us problems w/ the gif picker.

src/view/shell/Composer.tsx
```diff
@@ -1,77 +1,45 @@
...
-  const wrapperAnimStyle = {
-    transform: [
-      {
-        translateY: initInterp.interpolate({
-          inputRange: [0, 1],
-          outputRange: [winHeight, 0],
-        }),
-      },
-    ],
-  }
```
haileyok, line 39: Oh interesting! I never looked at this code before, but I think the reason the animation didn't work is because this wasn't wrapped in `useAnimatedStyle()` maybe? Not important now. Curious that it did work in dev mode though 🤔.

Wild guess is that this gets transpiled differently based on whether it is/isn't wrapped given the babel config but I honestly have no idea.

src/alf/util/useColorModeTheme.ts
haileyok: Oh, if we're going to use `fullscreen` I guess we can revert this part anyway? At least on the iPhone 14 it takes up the full screen so I don't see any borders.

src/view/shell/Composer.tsx
mozzius: I think this is React Native's Animation API, not Reanimated

haileyok: Ah I missed you used `useAnimatedValue`, heh. Back to being confused why it didn't work on web lol. Guess maybe the same reason `LayoutAnimation` doesn't either...

gaearon (approved):
the animation feels just a _little_ too slow to me but let's give it a go and adjust if needed

haileyok:
:shipit:

PR #3590, opened 2024-04-17
properly animate the profile hover card

haileyok:
## Why

It was pointed out that hover card animations on Firefox were not working very well. If you were scrolled down the page far enough, they would stutter. Turns out we had not properly implemented `useFloating()`.

See https://floating-ui.com/docs/useFloating#open and https://floating-ui.com/docs/useFloating#ispositioned

Edit: Well, we don't actually need to use `open` or `isPositioned`, but this seems to give us a hint as to what might be causing the jank. The card starts off at 0,0 when it gets rendered, then moves to wherever it's supposed to be positioned. Moving the animation _inside_ of the moving element fixes the jank.

## Test Plan

Using Firefox, scroll down the page about 20 posts or so, and hover over an avatar. The animations should now be smooth.

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -294,7 +294,7 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
             style={floatingStyles}
             onPointerEnter={onPointerEnterCard}
             onPointerLeave={onPointerLeaveCard}>
-            <div style={animationStyle}>
+            <div style={{willChange: 'transform', ...animationStyle}}>
               <Card did={props.did} hide={onPress} />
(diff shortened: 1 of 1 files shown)


mary-ext:
`will-change` CSS property can be quite a footgun I wouldn't really recommend it if it works well without the property :sweat_smile:

gaearon (approved):
perfect

gaearon:
>will-change CSS property can be quite a footgun I wouldn't really recommend it if it works well without the property 😅

the library already adds it anyway so we're just making it consistent between first and next renders (the library was adding it too late)

gaearon:
(thanks for the tip tho, will forcefully remove `willChange` if it's too wonky)

PR #3591, opened 2024-04-17
don't underline handle in post meta

haileyok:
## Why

It looks ugly.

But seriously. Right now, having the single underline for the entire text doesn't look great at all. We can separate them into two components, so they are individually underlined, but the @ doesn't look great.

image

Instead, let's just not underline the handle at all. It is still a link though.

## Test Plan

Hover the display name which should still be underlined, then the handle which should not.

Screenshot 2024-04-16 at 7 10 20 PM

### DIFF
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -2,34 +2,35 @@ import React, {ComponentProps, memo, useMemo} from 'react'
 import {
   GestureResponderEvent,
   Platform,
+  Pressable,
   StyleProp,
-  TextStyle,
   TextProps,
+  TextStyle,
+  TouchableOpacity,
   View,
   ViewStyle,
-  Pressable,
-  TouchableOpacity,
 } from 'react-native'
-import {useLinkProps, StackActions} from '@react-navigation/native'
-import {Text} from './text/Text'
-import {TypographyVariant} from 'lib/ThemeContext'
-import {router} from '../../../routes'
+import {sanitizeUrl} from '@braintree/sanitize-url'
+import {StackActions, useLinkProps} from '@react-navigation/native'
+
+import {useModalControls} from '#/state/modals'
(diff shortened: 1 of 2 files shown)


haileyok:
Ugh, breaks mobile. Hold on.

haileyok:
Okay, mobile before and after:

## Before

Screenshot 2024-04-16 at 7 19 00 PM

## After

Screenshot 2024-04-16 at 7 18 52 PM

gaearon:
There seem to be some unintended layout changes on native.

- Handle jumped a bit closer to top-right than baseline, need to move it back
- Timestamp jumped to the right, needs to stay in place


https://github.com/bluesky-social/social-app/assets/810438/0e74cb0f-8096-43d8-8168-607ef3792396

This PR:

Screenshot 2024-04-17 at 14 24 18

Desired (baseline):

Screenshot 2024-04-17 at 14 24 51

gaearon (changes requested):
needs native fix

haileyok:
Okay, the `flex` needed to be `flexShrink` for smaller handles. But for the life of me I can't see where this padding is coming from

image

image

haileyok:
You have got to be kidding me...it has to be wrapped in `Text`. Otherwise the padding is - seemingly - randomly applied (it's likely how iOS is calculating the text size in reality). Wouldn't even see the shift with some handles but then others would have a full space at the end...

There's still one shift though that I think we should keep. Right now the time since post isn't aligned on the bottom, so we want to either put that there (my preference visually) or center both the handle and the time.

## Before

Screenshot 2024-04-17 at 9 14 54 AM

## After

Screenshot 2024-04-17 at 9 15 18 AM

gaearon (approved):
i think this works. tested web, ios, android

gaearon:
i changed it back to nbsp (but on the other side) because the padding didn't work on mobile

PR #3593, opened 2024-04-17
Movable following feed

estrattonbailey:
Backend PR: https://github.com/bluesky-social/atproto/pull/2427

**To do**
- [x] Integrate new post-create-account handling, setting default feeds

### Description

This PR adds the ability to move your following feed and even un-pin it. To do so, it runs a migration on first load from v1 saved feeds prefs to v2. See linked PR above.

V2 preferences have different interfaces for reading and writing this data. So this PR touches all of those usages in order to make this work. I'll leave comments in the code with more details.

Testing:
- create prod test account with a few feeds, saved and/or pinned, on bsky.app
  - log in to branch deploy https://social-app-pr-3593.onrender.com/
  - should migrate smoothly, play around
- un-pin all feeds
- on ProfileFeed/List screens, test removing/pinning in all configurations, Feeds and SavedFeeds should stay in sync
  - note: you can remove saved lists now!
- note that your test profile on bsky.app is still fine and usable, but that changes to feeds aren't reflected on the branch deploy
- create an account (can't do from branch deploy)
  - create a new account, select some algos, including Discover and complete onboarding, feeds should look as expected
  - create another account, select NO algos, feeds should be Timeline/Discover

CleanShot 2024-04-24 at 12 47 48@2x
CleanShot 2024-04-24 at 12 48 10@2x
CleanShot 2024-04-24 at 12 48 12@2x
CleanShot 2024-04-24 at 12 48 21@2x
CleanShot 2024-04-24 at 12 48 25@2x
CleanShot 2024-04-24 at 12 50 55@2x


src/lib/statsig/gates.ts
```diff
@@ -8,5 +8,5 @@ export type Gate =
   | 'new_search'
   | 'receive_updates'
   | 'show_follow_back_label'
-  | 'start_session_with_following'
```
estrattonbailey, line 11: With this PR, we always default back to following or primary algo.

src/state/queries/feed.ts
```diff
@@ -199,10 +206,14 @@ export function useSearchPopularFeedsMutation() {
...
-const FOLLOWING_FEED_STUB: FeedSourceInfo = {
+/**
+ * The following feed, with fallbacks to Discover
+ */
+const HOME_FEED_STUB: FeedSourceInfo = {
+  isPrimaryAlgorithm: false,
   type: 'feed',
   displayName: 'Following',
-  uri: '',
+  uri: 'home',
```
estrattonbailey, line 216: All feeds are now driven by the `uri` value. For following feeds, it's `following` (unused atm) or `home` (which is Following + fallbacks). These two values have special handling elsewhere in the codebase to map to `getTimeline` instead of `getFeed`.

src/state/queries/feed.ts
```diff
@@ -216,10 +227,11 @@ const FOLLOWING_FEED_STUB: FeedSourceInfo = {
   likeCount: 0,
   likeUri: '',
 }
-const DISCOVER_FEED_STUB: FeedSourceInfo = {
+const PWI_DISCOVER_FEED_STUB: FeedSourceInfo = {
+  isPrimaryAlgorithm: true,
   type: 'feed',
   displayName: 'Discover',
-  uri: '',
+  uri: DISCOVER_FEED_URI,
```
estrattonbailey, line 234: Another example of feeds driven by URI. And in the case of PWI, Discover is for all intents and purposes, the primary algo.

src/state/queries/feed.ts
```diff
@@ -284,14 +309,35 @@ export function usePinnedFeedsInfos() {
...
-      for (let pinnedUri of pinnedUris) {
-        if (resolved.has(pinnedUri)) {
-          result.push(resolved.get(pinnedUri))
+
+      // if primary algo is enabled and was fetched, add it to the front of the list
+      if (primaryAlgo?.enabled && primaryAlgo?.uri) {
+        const feedInfo = resolved.get(primaryAlgo.uri)
+        if (feedInfo) {
+          feedInfo.isPrimaryAlgorithm = true
+          result = [feedInfo, ...result]
```
estrattonbailey, line 321: Fetched as a group, but pre-pended if present and marked as primary algo

src/state/queries/feed.ts
```diff
@@ -284,14 +309,34 @@ export function usePinnedFeedsInfos() {
...
-      for (let pinnedUri of pinnedUris) {
-        if (resolved.has(pinnedUri)) {
-          result.push(resolved.get(pinnedUri))
+
+      // if primary algo is enabled and was fetched, add it to the front of the list
+      if (primaryAlgo?.enabled && primaryAlgo?.uri) {
+        const feedInfo = resolved.get(primaryAlgo.uri)
+        if (feedInfo) {
+          feedInfo.isPrimaryAlgorithm = true
+          result.unshift(feedInfo)
```
estrattonbailey, line 321: Fetched as a group, but prepended if present and marked as primary algo.

src/state/queries/feed.ts
```diff
@@ -284,14 +309,34 @@ export function usePinnedFeedsInfos() {
...
         }
       }
+
+      const pinnedUrisSansPrimary = pinnedUris.filter(uri => {
+        if (primaryAlgo?.enabled) {
+          return uri !== primaryAlgo?.uri
+        }
+        return true
+      })
```
estrattonbailey, line 330: This is defensive, as are a couple other spots. Example: 3p app allows user to pin their primary algo.

src/view/screens/Home.tsx
```diff
@@ -70,6 +74,10 @@ function HomeScreenReady({
   const rawSelectedFeed = useSelectedFeed()
   const setSelectedFeed = useSetSelectedFeed()
   const maybeFoundIndex = allFeeds.indexOf(rawSelectedFeed as FeedDescriptor)
+  /*
+   * N.B. if `rawSelectedFeed` returns `DEFAULT_FEED_DESCRIPTOR`,
+   * `maybeFoundIndex` will be -1 and we'll fall back to left-most tab
+   */
```
estrattonbailey, line 80: See usage here. We pass `setSelectedFeed ` `FeedDescriptor`s, which can vary for the primary algo. So if we want to select the primary algo, we have this handling to fall back to index 0 (left-most tab).

src/view/com/pager/TabBar.tsx
```diff
@@ -139,7 +140,10 @@ export function TabBar({
                 <Text
                   type={isDesktop || isTablet ? 'xl-bold' : 'lg-bold'}
                   testID={testID ? `${testID}-${item}` : undefined}
-                  style={selected ? pal.text : pal.textLight}>
+                  style={[
+                    selected ? pal.text : pal.textLight,
+                    {lineHeight: 20},
```
estrattonbailey, line 145: `Feeds ✨ ` was increasing the line height slightly and resulting in a gap between the selected state and the top post, which looked bad. `20` appears to match well on desktop/mobile and fix the problem.

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -62,10 +62,13 @@ export function StepFinished() {
         // these must be serial
         (async () => {
           await getAgent().setInterestsPref({tags: selectedInterests})
-          await saveFeeds({
-            saved: selectedFeeds,
-            pinned: selectedFeeds,
-          })
+          await addSavedFeeds(
```
estrattonbailey, line 65: TODO

src/components/hooks/useHeaderOffset.ts
estrattonbailey, line 1: Pulled this out of `FeedPage`

src/lib/constants.ts
```diff
@@ -42,7 +45,7 @@ export function IS_TEST_USER(handle?: string) {
 }
 
 export function IS_PROD_SERVICE(url?: string) {
-  return url && url !== STAGING_SERVICE && url !== LOCAL_DEV_SERVICE
+  return url && url !== STAGING_SERVICE && !url.startsWith(LOCAL_DEV_SERVICE)
```
estrattonbailey, line 48: During account creation, this is fine, but once normalized via `URL.toString` it contains a slash so the exact match was failing.

src/screens/Feeds/NoFollowingFeed.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const {mutateAsync: addSavedFeeds} = useAddSavedFeedsMutation()
+
+  const addRecommendedFeeds = React.useCallback(
+    (e: any) => {
+      e.preventDefault()
+
+      addSavedFeeds([
+        {
+          ...TIMELINE_SAVED_FEED,
+          pinned: true,
```
estrattonbailey, line 24: Where `pinned` is overwritten in this PR it's bc I want to ensure the intent here is never modified by editing the source const

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -62,10 +64,39 @@ export function StepFinished() {
...
+          }
+
+          await overwriteSavedFeeds([
+            {
+              ...TIMELINE_SAVED_FEED,
+              pinned: true,
+              id: TID.nextStr(),
+            },
+            ...otherFeeds,
+          ])
```
estrattonbailey, line 99: After signup, an account has `timeline, discover` (has been like this). Previously, if a user also selected Discover from onboarding, it would be merged and deduped.

Duplicates are possible in v2, so the logic in this file preserves existing logic by overwriting feeds in a consistent order. All of this will change when reduced onboarding lands.

src/screens/Signup/state.ts
```diff
@@ -264,9 +267,12 @@ export function useSubmitSignup({
...
           verificationCode: verificationCode,
         })
+        // this initial set triggers v1->v2 feeds migration
         await setBirthDate({birthDate: state.dateOfBirth})
+        // add in default feed after migration has taken place
         if (IS_PROD_SERVICE(state.serviceUrl)) {
-          setSavedFeeds(DEFAULT_PROD_FEEDS)
+          // match existing handling
+          addSavedFeeds([DISCOVER_SAVED_FEED])
         }
```
estrattonbailey, line 276: We'll be able to remove this with reduced onboarding.

src/state/queries/post-feed.ts
```diff
@@ -43,8 +43,13 @@ type AuthorFilter =
...
 type FeedUri = string
 type ListUri = string
+/**
+ * Represents a "not found" state, that results in the left-most tab being
+ * selected on the Home screen.
+ */
+export const DEFAULT_FEED_DESCRIPTOR = '__default__'
 export type FeedDescriptor =
-  | 'home'
+  | '__default__'
```
estrattonbailey, line 52: This is here just to have a consistent fallback that forces the pager to select the first tab. See other comment further down.

src/state/queries/post-feed.ts
```diff
@@ -370,14 +375,12 @@ function createApi(
   params: FeedParams,
   feedTuners: FeedTunerFn[],
 ) {
-  if (feedDesc === 'home') {
+  if (feedDesc === 'following') {
```
estrattonbailey, line 378: This PR also normalizes how we refer to following. No longer `following` (the og) or `home`. It's just `following`, which receives `home`s treatment.

src/state/shell/selected-feed.tsx
```diff
@@ -1,47 +1,49 @@
...
-    }
+
+  const feedFromPersisted = persisted.get('lastSelectedHomeFeed')
+  if (feedFromPersisted) {
+    // Fall back to the last chosen one across all tabs.
+    return feedFromPersisted as FeedDescriptor
   }
-  return 'home'
+
+  return DEFAULT_FEED_DESCRIPTOR
```
estrattonbailey, line 40: This handling now returns whatever tab was last seen, OR this default value.

src/view/screens/Home.tsx
```diff
@@ -216,26 +225,39 @@ function HomeScreenReady({
...
+                feed={feed}
+                feedParams={homeFeedParams}
+                renderEmptyState={renderFollowingEmptyState}
+                renderEndOfFeed={FollowingEndOfFeed}
+              />
+            )
+          }
+          const savedFeedConfig = preferences.savedFeeds.find(f => {
+            return feed.includes(f.value)
+          })
```
estrattonbailey, line 245: This is drilled down to `FeedErrorMessage`

pfrazee (commented):
Visual skim of code looks good, will test next

src/view/screens/Home.tsx
```diff
@@ -152,12 +159,15 @@ function HomeScreenReady({
...
     (index: number, reason: LogEvents['home:feedDisplayed']['reason']) => {
       const feed = allFeeds[index]
-      logEvent('home:feedDisplayed', {
-        index,
-        feedType: feed.split('|')[0],
-        feedUrl: feed,
-        reason,
-      })
+
+      if (feed) {
```
gaearon, line 163: Can this ever be empty? Seems like we'd expect the index during selection to always be in bounds. Has something changed here?

src/view/screens/Home.tsx
```diff
@@ -198,12 +208,11 @@ function HomeScreenReady({
     return <CustomFeedEmptyState />
   }, [])
 
-  const [homeFeed, ...customFeeds] = allFeeds
   const homeFeedParams = React.useMemo<FeedParams>(() => {
     return {
       mergeFeedEnabled: Boolean(preferences.feedViewPrefs.lab_mergeFeedEnabled),
       mergeFeedSources: preferences.feedViewPrefs.lab_mergeFeedEnabled
-        ? preferences.feeds.saved
+        ? preferences.savedFeeds.map(f => f.value)
```
gaearon, line 215: Seems like we'd want to exclude `following` from this.

src/view/screens/Home.tsx
```diff
@@ -218,26 +227,39 @@ function HomeScreenReady({
...
+                isPageFocused={selectedFeed === feed}
+                feed={feed}
+                feedParams={homeFeedParams}
+                renderEmptyState={renderFollowingEmptyState}
+                renderEndOfFeed={FollowingEndOfFeed}
+              />
+            )
+          }
+          const savedFeedConfig = preferences.savedFeeds.find(f => {
+            return feed.includes(f.value)
```
gaearon, line 246: Feels off that we have to do a loop within a loop here. Not the end of the world but I think this means iterating over `allFeeds` might need some restructuring (should we just iterate over `preferences.savedFeeds` instead?)

gaearon (commented):
flagging some things that jumped out at me

src/state/queries/post-feed.ts
estrattonbailey: `null` is a great idea!

src/view/screens/Home.tsx
estrattonbailey: I don't remember why I added this but yeah I don't think it's needed 👍

src/view/screens/Home.tsx
estrattonbailey: Good call. Now that I'm thinking about it, should we exclude pinned feeds too? The current functionality does not, but does it make sense to sample in Discover if you also have it pinned?

src/view/screens/Home.tsx
estrattonbailey: Oh that's a good shout, it's 1:1 now, we might as well loop over `prefs.savedFeeds.filter(f => f.pinned)` and create the `FeedDescriptor`s in situ

src/view/screens/Home.tsx
gaearon: I wouldn't mess with that here yet, but IMO the whole "mergefeed" feature needs a rethink pass, it's kind of stuck as a v0.1 experiment

src/screens/Feeds/NoSavedFeedsOfAnyType.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * have no other feeds saved.
+ */
+export function NoSavedFeedsOfAnyType() {
+  const t = useTheme()
+  const {_} = useLingui()
+  const {isPending, mutateAsync: overwriteSavedFeeds} =
+    useOverwriteSavedFeedsMutation()
+
+  const addRecommendedFeeds = React.useCallback(async () => {
+    await overwriteSavedFeeds(
```
gaearon, line 26: Maybe silly question but is there a reason this doesn't use `addSavedFeeds`?

src/state/queries/feed.ts
```diff
@@ -243,59 +245,89 @@ export function usePinnedFeedsInfos() {
...
+  const pinnedLists = pinnedItems.filter(feed => feed.type === 'list')
 
   return useQuery({
     staleTime: STALE.INFINITY,
     enabled: !isLoadingPrefs,
     queryKey: [
       pinnedFeedInfosQueryKeyRoot,
-      (hasSession ? 'authed:' : 'unauthed:') + pinnedUris.join(','),
+      (hasSession ? 'authed:' : 'unauthed:') +
+        pinnedItems.map(f => f.value).join(','),
```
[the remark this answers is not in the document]
gaearon: Answer: `value` is probably a better key because in principle there's nothing preventing a saved feed's `value` changing on its own. Plus we don't really _need_ the ids here at all, the received data fully depends on the URLs.

src/view/screens/Home.tsx
gaearon: found a neat way to score this

PR #3594, opened 2024-04-17
[Statsig] Fix exposure logging by doing gate checks lazily

gaearon:
Our experiments are noisy. The hypothesis is that they're noisy in part because we're overcounting exposures. We only want an exposure if the GK result materially affects the outcome. However, the previous `useGate(...)` API forced us to do exposures early even if the GK results isn't even used (such as if it's only taken into account during some future event, or maybe it's used inside another condition).

To counter-act that, I'm changing the API to be:

```js
const gate = useGate()

if (gate('foo')) {
  // ...
}
```

The `gate('foo')` call will log an exposure so it must be put as late as possible — i.e. at the point where it actually affects the outcome in some user-perceptible way.

```js
// Bad ❌
if (gk('foo') && bar) {
  // ...
}

// Good ✅
if (bar && gk('foo')) {
  // ...
}
```

Although look inside!

```js
// Still bad ❌
if (bar && gk('foo')) {
  if (baz) {
    // ...
  }
}
```

(Can you guess how to fix this one?)

This also means that this is now an anti-pattern:

```js
const gate = useGate('foo')
const enableFoo = gate('foo') // Bad ❌

if (bar && enableFoo) {
  // ...
}
```

Instead, do:

```js
const gate = useGate('foo')

if (bar && gate('foo')) { // Good ✅
  // ...
}
```

A nice rule of thumb is imagine each `gate('foo')` call costs you money (it's also kind of true). How would you write the code? Write it that way.

## Test Plan

Manual scanning.

Check some test exposures in the dev console. Verify they hit more lazily than before (e.g. only when clicking Follow for the suggested follows gate).

Screenshot 2024-04-17 at 05 59 53

I need to also check that changing accounts works as expected — haven't done that yet.

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -2,11 +2,7 @@ import React from 'react'
 import {Platform} from 'react-native'
 import {AppState, AppStateStatus} from 'react-native'
 import {sha256} from 'js-sha256'
-import {
-  Statsig,
-  StatsigProvider,
-  useGate as useStatsigGate,
-} from 'statsig-react-native-expo'
+import {Statsig, StatsigProvider} from 'statsig-react-native-expo'
 
 import {logger} from '#/logger'
 import {isWeb} from '#/platform/detection'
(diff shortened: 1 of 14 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -210,7 +210,8 @@ function useSuggestedFollowsV2(): [
 
 function SearchScreenSuggestedFollows() {
   const pal = usePalette('default')
-  const useSuggestedFollows = useGate('use_new_suggestions_endpoint')
+  const gate = useGate()
+  const useSuggestedFollows = gate('use_new_suggestions_endpoint')
     ? // Conditional hook call here is *only* OK because useGate()
       // result won't change until a remount.
```
estrattonbailey, line 216: Nit: unused comment

src/lib/statsig/statsig.tsx
```diff
@@ -98,16 +94,23 @@ export function logEvent<E extends keyof LogEvents>(
...
-export function useGate(gateName: Gate): boolean {
-  const {isLoading, value} = useStatsigGate(gateName)
-  if (isLoading) {
-    // This should not happen because of waitForInitialization={true}.
-    console.error('Did not expected isLoading to ever be true.')
+export function useGate(): (gateName: Gate) => boolean {
+  const cache = React.useRef<Map<Gate, boolean>>()
+  if (cache.current === undefined) {
+    cache.current = new Map()
   }
```
estrattonbailey, line 101: Maybe silly question, but why not set the initial value in the hook call?

estrattonbailey (approved):
Nice, make sense to me

src/lib/statsig/statsig.tsx
estrattonbailey: Oh lawl just saw this comment https://github.com/bluesky-social/social-app/pull/3498/files#r1568218021

gaearon: Can you show the code you mean?

src/view/screens/Search/Search.tsx
gaearon: It's relevant to this:

```js
    ? // Conditional hook call here is *only* OK because useGate()
      // result won't change until a remount.
      useSuggestedFollowsV2
    : useSuggestedFollowsV1
```

gaearon:
I'm gonna move the cache stuff into another PR. There's more there and it's unrelated to this piece.

PR #3595, opened 2024-04-17
Update catalan messages.po

ivanbea:
New lines translated again, just 16 this time

You can check it @jordimas @darccio @ferranrego @MiquelAdell @mvdan @johnnydement @Mejans @surfdude29


pfrazee (approved):
Short enough, I was able spot check it. Looks good, thanks!

PR #3597, opened 2024-04-17
Cleanup files after upload on native

haileyok:
## Why

This kicks off a migration to `expo-file-system`. All we do here is remove files from the local storage after they are uploaded to prevent a growing cache.

## Test Plan

Attempting to upload an image or images in the composer still works, and a log lets us know that the image has been deleted. Similarly, after uploading an external link card, the downloaded image gets deleted.

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -1,6 +1,7 @@
+import {deleteAsync} from 'expo-file-system'
 import {
-  AppBskyEmbedImages,
   AppBskyEmbedExternal,
+  AppBskyEmbedImages,
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
   AppBskyFeedThreadgate,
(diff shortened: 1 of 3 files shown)


src/lib/api/index.ts
```diff
@@ -117,6 +119,13 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
       const {width, height} = image.compressed || image
       logger.debug(`Uploading image`)
       const res = await uploadBlob(agent, path, 'image/jpeg')
+
+      if (isNative) {
+        try {
+          deleteAsync(path)
```
gaearon, line 125: If this throws asynchronously, does that crash the app or not? E.g. imagine its implementation is

```js
async function deleteAsync() {
  await new Promise(resolve => setTimeout(resolve, 1000))
  throw Error('boom')
}
```

If this doesn't lead to a redbox, we're good. If it does, I think we need to handle Promise rejection too.

haileyok: No red box. I suppose we could actually remove the try/catch completely actually. That being said, I think we should just err on the side of caution in all cases. Updating.

gaearon (approved):

gaearon (approved):

PR #3599, opened 2024-04-17
Cleanup files after each iteration of compression and downloading

haileyok:
## Why

We currently do not remove files after each iteration of compression. This isn't ideal because we create a lot of extra copies of images that we don't need, bloating the document storage size

This also replaces a few current uses of `RNFS.unlink()` in favor of `deleteAsync()`, although there is no difference outside of the API change.

We also make a change to `moveToPermanentPath()` which ensures the move happens. It is possible for `moveAsync()` to fail (or in this case `RNFS.moveFile()`. If it does, then we may be in trouble later on. What we should instead do is _always_ copy the file, and then delete the original if possible. That way, we are guaranteed to have a copy that we can work with in the next step of the file upload process.

### Does this fix the error?

This PR might not solve the root issue at hand which is this:

Screenshot 2024-04-17 at 2 30 30 PM

However, I found something very suspect whenever I was working on this. The `RNFS.TemporaryDirectory` path is _not_ writeable using `copyAsync()` or `moveAsync()` from the `expo-file-system` package. I'm not certain what the distinction is there. I suspect there might be an API change, and RNFS is using some outdated API, or is doing something else bizarre under the hood. There are a lot of issues open in their GitHub repo over the past year regarding issues causing errors and crashes.

Rose should still be able to repro with the image that she has, so we should see if this resolves the issue for her.

## Test Plan

Testing that changing the avatar or changing the banner results in multiple iterations, all of which are properly cleaned up. We can still upload the avatar, since the final output is not deleted in this for loop.

https://github.com/bluesky-social/social-app/assets/153161762/c8547dc8-077f-468c-8f05-667d439cc249

### DIFF
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -113,7 +113,12 @@ export async function shareImageModal({uri}: {uri: string}) {
       UTI: 'image/png',
     })
   }
-  RNFS.unlink(imagePath)
+
+  try {
+    deleteAsync(imagePath)
+  } catch (e) {
+    console.error(e)
+  }
 }
 
 export async function saveImageToMediaLibrary({uri}: {uri: string}) {
@@ -131,7 +136,11 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
   // save
   await MediaLibrary.createAssetAsync(imagePath)
 
-  // TODO we should add a `deleteAsync()` here as well -hailey
+  try {
+    deleteAsync(imagePath)
+  } catch (e) {
+    console.error(e)
+  }
 }
 
 export function getImageDim(path: string): Promise<Dimensions> {
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -204,7 +204,7 @@ async function moveToPermanentPath(path: string, ext = 'jpg'): Promise<string> {
   return normalizePath(destinationPath)
 }
 
-async function safeDeleteAsync(path: string) {
+export async function safeDeleteAsync(path: string) {
   try {
     // Normalize is necessary for Android, otherwise it doesn't delete.
     await deleteAsync(normalizePath(path))
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -24,7 +24,10 @@ export async function compressIfNeeded(
     mode: 'stretch',
     maxSize,
   })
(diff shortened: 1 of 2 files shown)


src/lib/media/manip.ts
```diff
@@ -23,7 +24,10 @@ export async function compressIfNeeded(
     mode: 'stretch',
     maxSize,
   })
-  const finalImageMovedPath = await moveToPermanentPath(resizedImage.path)
+  const finalImageMovedPath = await moveToPermanentPath(
+    resizedImage.path,
+    '.jpg',
```
haileyok, line 29: This was missing a file extension 🤔

src/lib/media/manip.ts
```diff
@@ -168,26 +184,40 @@ async function doResize(localUri: string, opts: DoResizeOpts): Promise<Image> {
...
   https://github.com/ivpusic/react-native-image-crop-picker/issues/1199
   */
   const filename = uuid.v4()
 
-  const destinationPath = joinPath(
-    RNFS.TemporaryDirectoryPath,
-    `${filename}${ext}`,
-  )
-  await RNFS.moveFile(path, destinationPath)
+  const destinationPath = joinPath(cacheDirectory ?? '', `${filename}.${ext}`)
```
haileyok, line 208: `cacheDirectory` will not be null _unless_ we call this on web. This function only ever gets called on native though, so we know that `cacheDirectory` will always be available.

gaearon: If `cacheDirectory` can never be null, let's assert that? `?? ''` isn't good cause it obscures intent for future readers

src/lib/media/manip.ts
gaearon: is this always jpg?

gaearon:
Things somewhat improved in this PR, though a lot of stuff still isn't being cleaned up.

## Uploading an avatar

### Android: Before and After

upl_avatar_before

upl_avatar_after

### iOS: Before and After

ios_upl_avatar_before

ios_upload_avatar_after

## Uploading a post with an image

### Android: Before and After

post_image_before

post_image_after

### iOS: Before and After

ios_post_img_before

ios_upload_post_after

## Saving an image to a camera roll

### Android: Before and After

save_img_before

save_img_after

### iOS

No changes.

gaearon (approved):

PR #3600, opened 2024-04-17
Add GIF select to composer

mozzius:
Adds a simple way to add GIFs to your post, via the GIPHY API. This also updates the photo/camera icons in the composer to ALF icons

iOS (got rate limited while recording lol):

https://github.com/bluesky-social/social-app/assets/10959775/f0e417f1-db6d-4364-8091-ad19ce110922

Android:

https://github.com/bluesky-social/social-app/assets/10959775/6d9cc6fb-db90-480a-a9f1-b39d3c81eb74

Web:

https://github.com/bluesky-social/social-app/assets/10959775/450f549b-94df-4f53-b925-2d3622d91d26

Consent flow:

https://github.com/bluesky-social/social-app/assets/10959775/0dbc370e-dcad-4932-86fc-b61b5ddf0232

## Test plan

1. Check that you can add a GIF
2. Check that search works
3. Check that the consent flow works
    3a. Confirm that the copy is good
4. Check error/loading states


haileyok:
image

src/components/dialogs/GifSelect.tsx
```diff
@@ -0,0 +1,404 @@
...
+export function GifSelectDialog({
+  control,
+  onClose,
+  onSelectGif: onSelectGifProp,
+}: {
+  control: Dialog.DialogControlProps
+  onClose: () => void
+  onSelectGif: (url: string) => void
+}) {
+  const externalEmbedsPrefs = useExternalEmbedsPrefs()
```
gaearon, line 35: Can you move this into `DialogInner` like in some other places? So these Hooks don't run at all unless the dialog is visible.

src/lib/analytics/types.ts
```diff
@@ -11,6 +11,8 @@ export type TrackPropertiesMap = {
   'Composer:CameraOpened': {}
   'Composer:GalleryOpened': {}
   'Composer:ThreadgateOpened': {}
+  'Composer:GifSelectOpened': {}
+  'Composer:GifSelected': {}
```
gaearon, line 15: This is old (mixpanel) analytics, I think you want `statsig/events.ts` instead.

src/view/com/composer/photos/SelectGifBtn.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function SelectGifBtn({onClose, onSelectGif, disabled}: Props) {
+  const {track} = useAnalytics()
+  const {_} = useLingui()
+  const control = useDialogControl()
+  const t = useTheme()
+
+  const onPressSelectGif = useCallback(async () => {
+    track('Composer:GifSelectOpened')
```
gaearon, line 27: Should be `logEvent` (`track` is the old one).

src/components/dialogs/GifSelect.tsx
mozzius: The height of the sheet depends on this value, so I don't think it'd be possible to move down unless we rethink the consent flow approach

gaearon: I see. OK then

src/components/dialogs/GifSelect.tsx
```diff
@@ -0,0 +1,404 @@
...
+              shape="round"
+              onPress={() => control.close()}
+              label={_(msg`Close GIF dialog`)}>
+              <ButtonIcon icon={Arrow} size="md" />
+            </Button>
+          )}
+
+          <TextField.Root>
+            <TextField.Icon icon={Search} />
+            <TextField.Input
```
gaearon, line 164: What's up with our text input that allows this? Is this specific to our component or does the raw RN one have the same issue?

https://github.com/bluesky-social/social-app/assets/810438/905c1b0d-72c0-4551-b59a-7e65099bbf42

gaearon:
Seems a bit confusing that you can still press GIF or Photo if you already have a gif. Not super obvious that this would replace it. Maybe that's OK, but then it's confusing that the button _does_ get disabled in the opposite scenario (see last example).

https://github.com/bluesky-social/social-app/assets/810438/a844f0b4-44cf-4189-ac67-7b5b4e1f2e9e

gaearon:
1. Add a GIF
2. Add some photos (they currently replace the GIF)
3. Delete photos

The GIF pops back up as if it's still there. But actually if I didn't delete the photos, it wouldn't be there.

gaearon:
Play button being unclickable in the composer feels very confusing. I presume the new embed will fix it?

gaearon:
Pressing Escape on the web should dismiss the dialog.

mozzius:
> Play button being unclickable in the composer feels very confusing. I presume the new embed will fix it?

I think it would? Maybe it should allow taps

gaearon:
Is this legit?

Screenshot 2024-04-18 at 18 09 51

gaearon:
When I choose a gif, I see a few loading states:

https://github.com/bluesky-social/social-app/assets/810438/65fe1fec-9d21-4608-9d5f-15b3383e4aea

First: 

Screenshot 2024-04-18 at 18 12 19

Second:

Screenshot 2024-04-18 at 18 12 25

Third:

Screenshot 2024-04-18 at 18 12 36

Since we already know the aspect ratio (do we?), can we jump straight to the third one? Or, if we have to keep the spinner, can we specify the aspect ratio so that it doesn't jump multiple times?

Or would this be resolved by Hailey's changes?

gaearon (approved):
looks great, let's land assuming the things i mentioned have fixes

haileyok:
> Is this legit?

Missing type on the `.web`

> Play button being unclickable in the composer feels very confusing. I presume the new embed will fix it?

Yea. I don't think it's so bad to show it but 1. should follow the autoplay setting and 2. just needs to look better.

> What's up with our text input that allows this?

I don't know if this is just us, but it used to happen in our old login screens before we used our new text input API. So if it is us it's something deeper than just that API. (This could be that predictive text junk too, although I thought that was only enabled in a multiline text input. Looks similar though)

src/components/dialogs/GifSelect.tsx
mozzius: Removing the `value` prop fixes this - a little concerning. We should check the app for other places where this might happen

src/components/Lists.tsx
```diff
@@ -33,6 +35,7 @@ export function ListFooter({
         a.pb_lg,
         t.atoms.border_contrast_low,
         {height: height ?? 180, paddingTop: 30},
+        flatten(style),
```
gaearon, line 38: Why is `flatten` necessary here?

src/components/dialogs/GifSelect.tsx
```diff
@@ -0,0 +1,360 @@
...
+      control.close(() => onSelectGifProp(gif))
+    },
+    [control, onSelectGifProp],
+  )
+
+  let content = null
+  let snapPoints
+  switch (externalEmbedsPrefs?.giphy) {
+    case 'show':
+      content = <GifList control={control} onSelectGif={onSelectGif} />
```
haileyok, line 53: Small nit: you don't need to pass `control` here. Instead, you can access `close` from inside `GifList` with

```
const {close} = Dialog.useDialogContext()
```

See https://github.com/bluesky-social/social-app/pull/3605/files#diff-99a1e15777b23685eb4bff778798a79a17259a82785796933427fbc6c56734c3

This should be fine but I'm pretty scared of these things right now so trying to keep all of them using the same logic for now 😅

src/components/dialogs/GifSelect.tsx
```diff
@@ -0,0 +1,360 @@
...
+  const setExternalEmbedPref = useSetExternalEmbedPref()
+
+  const onShowPress = useCallback(() => {
+    setExternalEmbedPref('giphy', 'show')
+  }, [setExternalEmbedPref])
+
+  const onHidePress = useCallback(() => {
+    setExternalEmbedPref('giphy', 'hide')
+    control.close()
+  }, [control, setExternalEmbedPref])
```
haileyok, line 302: This one right now is definitely a no, again because we're doing stuff wrong elsewhere. We've been talking about how to make it safer so this won't break things but for now let's use the callback in `close()`

```
close(() => setExternalEmbedPref()
```

This one here might not actually cause problems, but it's possible to mess up the android back button if we don't use the callback here. See https://github.com/bluesky-social/social-app/pull/3451/files#r1556676290

Not your fault though, this hasn't been very clear and we're still figuring out what is and isn't okay and how to make sure it can't break (I'm liable to fuck this up again soon once I forget about the problem 😄)

haileyok: I fixed these two things btw in https://github.com/bluesky-social/social-app/pull/3605, so just leaving for reference 🙏

mozzius: Darn, yeah, my bad. I was copying code over from a legacy modal and forgot to account for this limitation. Thanks!

PR #3604, opened 2024-04-18
[Statsig] Isolate cache per session

gaearon:
## What

Ensures that gates report consistent values throughout a user's app session, and don't change abruptly in a running app.

## How

Replaces the per-component cache with a single cache ensuring gate reads are the same across the entire component trees for a session. I define a "session" as a period of time with the same `currentAccount?.did`. When the session changes, we throw away the old cache, and the checks start hitting the singleton `Statsig` object which also tracks exposures — getting cache (for the new session) along the way.

## Test Plan

Decrease the `setInterval` delay to ~10 seconds.

Open something that depends on a gate (e.g. search). Flip the gate in the tool. Stay on the page. Do some state updates (e.g. navigate to Home and Back). Verify that you still see the previous gate's behavior — it's as if the gate has not been touched. Ensure you've waited ~10s and refresh the page. You should immediately see the new gate behavior.

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -94,23 +94,30 @@ export function logEvent<E extends keyof LogEvents>(
   }
 }
 
+// We roll our own cache in front of Statsig because it is a singleton
+// and it's been difficult to get it to behave in a predictable way.
+// Our own cache ensures consistent evaluation within a single session.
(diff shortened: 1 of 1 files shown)


gaearon:
Extracting unrelated changes to another PR.

mozzius (approved):
Looks good. Gate state was constant for the whole session

PR #3605, opened 2024-04-18
GIF Viewer

haileyok:
## Why and How

The overall why is pretty obvious. We want to add GIFs. Let's break down the individual whys though.

### Permission/making requests

Following the current "permissions" scheme of telling users before outbound connections get made, the GIF composer as well as individual gifs will not be accessible/playable until the user consents. The default, "we have not asked for permission yet" display will be the same as if the GIF were just paused - with the exception that we are guaranteed to not be making a connection to GIPHY yet for a thumbnail but instead using the one hosted on our CDN:

Screenshot 2024-04-21 at 7 44 43 PM

The first time that user attempts to play a GIF, they will be prompted with a notice:

Screenshot 2024-04-21 at 7 45 25 PM

If they choose to accept the request, then they will not be asked again and the placeholder image will be replaced with a `<GifView>`. If they reject, then GIPHY links will appear as external link cards - as they do right now if a user rejects permission.

The same functionality applies on the PWI.

### Web Player

On the web, we can use the `<video>` element to play videos. We control the playback of the MP4 through a ref on the component and with the same API as the native views. We use `preload` for these components to ensure smooth playback.

### Native Players

The React portions of these views are pure. The only instance where their props should change at all is if the `disableAutoplay` setting is changed, in which case we will simply call `playAsync()`/`pauseAsync()` on the underlying native component. React is informed of playback state updates through the `onPlayerStateChange` event.

The native views will fill the size of their container, and as a result, dimensions should always be provided on the `<GifView>`. The best way to maintain appropriate sizing is through using the `aspectRatio` style on the view, i.e. `dimensions.width / dimensions.height`. That way, there are no async layout updates that cause jumps in the feed/thread.

The native players use WebP for playback, which is more performant on native platforms when rending large lists of animatable content. (There's a bit of nuance here, and if we had our own `UITableView` or `RecyclerView` on the native side, we could do this much more elegantly and with MP4s. However, because of some limitations on knowing when items are going to appear on screen with the `RCTScrollView`, there's some concerns with MP4s in this case).

Generally, on most connections WebPs should still load in under 1s. The average WebP from GIPHY appears to be 80KB. None the less, we also supply a `placeholderSource` to the `<GifView>` that will be displayed until the WebP has rendered. The intent here is that there will be almost no blank areas for loading while scrolling.

Additionally, the native views support `GifView.prefetchAsync([source1, source2, ...])`. We could use `FlatList`'s viewability callback to prefetch WebPs before they come into the viewport if we want to, but that logic would come in a separate PR. A (probably convoluted) example of this would be something like this: https://github.com/haileyok/social-app/blob/feat/feed-image-prefetch/src/lib/hooks/usePrefetchListImages.ts.

### Video?

Originally I had written this to use `AVPlayer` or `ExoPlayer` (depending on platform), however for the reasons above (regarding view lifecycle) I've opted to use the WebP instead. I've maintained a video branch though, which we can use for actual video playback at some point. The difference will be that we won't want to autoplay video all the time when we do make that move. Instead, we might have a WebP placeholder that would be "expanded" fullscreen to a full video - similar to how Twitter works. We'll discuss that topic when we get there.

Overall, performance is not affected much here on either platform. I've profiled large lists of GIFs on both platforms with a low-mid end device (iPhone X on iOS and an older Samsung A-series for Android). Scroll performance is no less performant than without the GIFs (although we certainly need to do overall work on Android...)

### Other Changes

I've had to make some changes to all of the embed cards to keep styles consistent - mainly top padding - while also allowing the GIF compoenent to expand to it's full size. I've gone through all of the link card cases that I can think of on both native and web and cannot see anything that stands out as a problem or regression, but just keep an eye on that.

## Test Plan

Like I said above, keep an eye on embed cards. There should be no differences in render appearance.

- Gifs in the composer
- Gifs in the feed
- Pause/play of a gif in the feed
- A paused gif should stay paused whenever you scroll past it and then scroll back to it
- Gifs should stay paused between app backgrounding to foregrounding (this was a problem on Android and needed a bit of extra logic to get right)
- Large lists of gifs shouldn't cause any performance/frame drops (the profile `orpel.bsky.social` is a good account to test this with, or maybe we should create our own thread of gifs)
- Checking the network logs on web in incognito should show:
  - Initial request for the profile `orpel.bsky.social` should show requests for the gif embeds going to our CDN for the placeholder
  - No connections to Giphy should occur until clicking play on one of the gifs and accepting the prompt
  - Declining the prompt should revert these embeds to using the normal link cards
- There shouldn't be any layout jumps with any of the new gif embeds (again, easy to check in a bigger thread or profile that has a lot of gifs)
- Test on mobile browsers as well. Aside from our composer not being great there in general, both the GIF selector and playback of gifs in the feeds works well.


modules/expo-bluesky-video-player/ios/ExpoBlueskyVideoPlayerView.swift
```diff
@@ -0,0 +1,65 @@
...
+  
+  public required init(appContext: AppContext? = nil) {
+    self.isPlaying = self.autoplay
+    self.controller = PlayerController()
+    super.init(appContext: appContext)
+    self.clipsToBounds = true
+    
+    self.controller.setFrame(rect: bounds)
+    self.addSubview(self.controller.view)
+  }
```
haileyok, line 26: This is the main native view. Here, we create a `AVPlayerViewController` that will host the actual `AVPlayer`. It's called `PlayerController` because we are subclassing `AVPlayerViewController` to add some useful functionality.

modules/expo-bluesky-video-player/ios/ExpoBlueskyVideoPlayerView.swift
```diff
@@ -0,0 +1,65 @@
...
+    if newWindow != nil {
+      guard let source = self.source else {
+        return
+      }
+
+      self.controller.prepare(source, view: self)
+    } else {
+      self.controller.release()
+    }
+  }
```
haileyok, line 38: As long as `removeClippedSubviews` on our `FlatList`s, only the visible items on the user's screen will ever be actively playing.

When the player leaves the viewport, we will clean up the player so as to not have a ton of players actively running. The `AVAsset` will be cached though.

modules/expo-bluesky-video-player/ios/PlayerItemManager.swift
```diff
@@ -0,0 +1,71 @@
...
+        return nil
+      }
+      self.assets[source] = asset
+      return asset
+    } else if let url = URL(string: source) {
+      guard let asset = AVAsset(url: url) as? AVURLAsset else {
+        return nil
+      }
+      self.assets[source] = asset
+      return asset
```
haileyok, line 29: These will always be of `AVURLAsset` but I really don't like doing `as!`, just in case.

modules/expo-bluesky-video-player/src/VideoPlayer.web.tsx
```diff
@@ -0,0 +1,56 @@
...
+    this.firePlayerStateChangeEvent(false)
+  }
+
+  async toggleAsync(): Promise<void> {
+    if (this.videoPlayerRef.current.paused) {
+      await this.playAsync()
+    } else {
+      await this.pauseAsync()
+    }
+  }
```
haileyok, line 41: even though none of these are actually async on web, let's just keep the same types as what we have on native.

modules/expo-bluesky-video-player/src/VideoPlayer.web.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <video
+        src={this.props.source}
+        autoPlay={this.props.autoplay ? 'autoplay' : undefined}
+        style={this.props.style}
+        preload={this.props.autoplay ? 'auto' : undefined}
+        loop="loop"
+        muted="muted"
+        ref={this.videoPlayerRef}
+      />
```
haileyok, line 53: Someone who has an idea should check if this is all we need

modules/expo-bluesky-video-player/ios/PlayerItemManager.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+import AVKit
+import CryptoKit
+
+class PlayerItemManager {
+  static let shared = PlayerItemManager()
+  
+  private var prefetchQueue: [String] = []
+  private var assets = NSMapTable<NSString, AVURLAsset>(keyOptions: .strongMemory, valueOptions: .weakMemory)
```
haileyok, line 9: Assets will be cached in memory as weak objects so GC will pick them up.

modules/expo-bluesky-video-player/android/src/main/java/expo/modules/blueskyvideoplayer/MediaItemManager.kt
```diff
@@ -0,0 +1,85 @@
...
+import java.io.File
+import java.io.FileOutputStream
+import java.lang.ref.WeakReference
+import java.security.MessageDigest
+import java.util.WeakHashMap
+
+class MediaItemManager(private val appContext: AppContext) {
+  companion object {
+    private val client = OkHttpClient()
+    private val mediaItems = WeakHashMap<String, WeakReference<MediaItem>>()
```
haileyok, line 19: Again using weak references for media items so they get removed by the GC

src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
```diff
@@ -35,51 +54,202 @@ export const ExternalLinkEmbed = ({
     embedPlayerParams.dimensions &&
     gate('new_gif_player')
```
haileyok, line 55: We can probably remove this gate, not necessary.

haileyok: Just removed it.

src/components/dialogs/GifSelect.tsx
```diff
@@ -45,12 +45,12 @@ export function GifSelectDialog({
   let snapPoints
   switch (externalEmbedsPrefs?.giphy) {
     case 'show':
-      content = <GifList control={control} onSelectGif={onSelectGif} />
+      content = <GifList onSelectGif={onSelectGif} />
```
haileyok, line 48: No changes in this file, just making some adjustments to make sure we use the `close(() => onClose())` pattern to avoid breaking things.

modules/expo-bluesky-gif-view/ios/GifView.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    if let url = URL(string: webpSource) {
+      self.webpOperation = imageManager.loadImage(
+        with: url,
+        options: [.retryFailed],
+        context: Util.createContext(),
+        progress: onProgress(_:_:_:),
+        completed: onLoaded(_:_:_:_:_:_:)
+      )
+    }
```
haileyok, line 93: We'll load both the placeholder and the WebP at the same time. The placeholder will only get used until the WebP has finished loading as well.

modules/expo-bluesky-gif-view/ios/GifView.swift
```diff
@@ -0,0 +1,178 @@
...
+      return
+    }
+
+    if let webpSource = self.webpSource,
+       imageUrl?.absoluteString == webpSource,
+       // UIImage perf suckssss if the image is animated
+       let data = data,
+       let animatedImage = SDAnimatedImage(data: data)
+    {
+      self.placeholderOperation?.cancel()
```
haileyok, line 143: In the event that we do load the WebP before the placeholder, we cancel the ongoing request for the placeholder.

gaearon:
Weird loading state on FF.

https://github.com/bluesky-social/social-app/assets/810438/b44cba88-dc7a-4fd0-aac1-440c898be3eb

gaearon:
If you don't give the permission, they currently revert to being embeds. That seems pretty confusing since you wouldn't know how to get back to the dialog if you dismissed it by accident.

Should we instead revert it to unplayed state (same before as giving permission)?

Screenshot 2024-04-22 at 23 21 03

haileyok:
> Firefox

Why Firefox why?

> Prompt

We shouldn't need to worry about this anyway since we're going to rm it sounds like w/ the proxy. If the plan changes will address.

surfdude29:
I just tested it out on the web and GIFs on Bluesky work so well, awesome work @haileyok! 🙌  (and @mozzius too ofc on the GIF picker :)

mozzius:
Fixed the FF loading issues

PR #3606, opened 2024-04-18
change default value of contentLanguage to all

omar0404:
[#3561](https://github.com/bluesky-social/social-app/issues/3561)


mary-ext:
You'd want to empty the array, not fill the array because that's going to require users to disable each and every language before being able to selectively filter the languages they want.

omar0404:
i agree, but why not to make all languages selected by default if the user is signed in ? @mary-ext

PR #3607, opened 2024-04-18
[Statsig] Prefetch configs for other accounts

gaearon:
Currently, we only fetch (and poll) the feature gate configuration for the _current_ user. So if the configuration has updated for another user, and then you switch accounts, we won't know that, and we'll use stale configs from the cache. It's only another refresh that you'll get the new config for the other user.

With this change, we're always fetching configs for all accounts:

- During init, we prefetch the "other" accounts.
- During polling, we prefetch all accounts (both current and "other").

While this doesn't by itself guarantee that an account switch would give you a fresh config (we'll improve on this in a follow-up PR), it would make it more likely because last poll could've grabbed it.

Note I've also changed the polling method from `updateUser(currentUser)` to `prefetchUsers(users)`. This seems to work just as fine but is more targeted (and unlike `updateUser` can be used for non-current users).

I also decreased the polling interval to one minute.

## Test Plan

**Tip: Decrease the polling time to 5 seconds for easier testing!**

Test that polling own account works:

1. Override some flag to be false for Alice.
1. Log into Alice's account. Observe the flag being false. Refresh if needed.
1. Override a flag to be true for Alice. Wait ~20 seconds for the change to deploy.
1. Wait for the poll. Verify the flag is still reported as false in the app.
1. Refresh the app. The flag should be reported as true.

Test that polling other accounts works:

1. Override some flag to be false for Alice.
1. Log into Alice's account. Observe the flag being false. Refresh if needed.
1. Switch to Bob's account. While being logged into Bob's account, override the flag to be true for Alice. Wait ~20 seconds for the change to deploy.
1. Wait for the poll (5 seconds).
1. Switch accounts to Alice. Verify the flag is already true for Alice (no refresh!)

For the initialization change:

1. Override some flag to be false for Alice.
1. Comment out polling completely in the source
1. Log into Alice's account. Observe the flag being false. Refresh if needed.
1. Switch to Bob's account. While being logged into Bob's account, override the flag to be true for Alice. Wait ~20 seconds for the change to deploy.
1. Refresh while being on Bob's account.
1. Refresh again while being on Bob's account.
1. Switch to Alice's account. Verify the flag is already true for Alice (no refresh!)

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -34,7 +34,7 @@ if (isWeb && typeof window !== 'undefined') {
 
 export type {LogEvents}
 
-function createStatsigOptions() {
+function createStatsigOptions(prefetchUsers: StatsigUser[]) {
   return {
     environment: {
       tier:
@@ -48,6 +48,8 @@ function createStatsigOptions() {
     // This ensures the UI is always consistent and doesn't update mid-session.
     // Note this makes cold load (no local storage) and private mode return `false` for all gates.
     initTimeoutMs: 1,
+    // Get fresh flags for other accounts as well, if any.
+    prefetchUsers,
   }
 }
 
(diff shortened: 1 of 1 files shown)


mozzius (approved):
LGTM!

PR #3612, opened 2024-04-18
adding notification filters

skiniks:
This PR is meant to be a jumping point for adding filters to the notifications screen to sort by type. 

Screenshot 2024-04-18 at 3 18 44 PM

### DIFF
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -60,7 +60,10 @@ export function Feed({
         let filteredItems = page.items
         if (filterType === 'Mentions') {
           let mentions = page.items.filter(
-            item => item.type === 'reply' || item.type === 'quote',
+            item =>
+              item.type === 'reply' ||
+              item.type === 'quote' ||
+              item.type === 'mention',
           )
           filteredItems = mentions
         }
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks for the PR! This is quite overwhelming and I think this level of granularity isn't that useful. How about we start with All and Mentions, similar to Twitter? Mentions would include any textual interactions, i.e. both replies and quotes.

skiniks:
Thanks @gaearon! I agree. I've gone ahead and updated the PR.

Screenshot 2024-04-19 at 9 51 58 AM

gaearon:
This implementation seems to just replace the contents of the same list, but I think we should treat these as actually separate tabs. E.g. you don't want scrolling one to affect the other. You can check the Search page for an example of how to make it work.

gaearon (changes requested):
need to use a real Pager

gaearon:
OK, so the experience here isn't super great because all filtering is done client-side, and so the "mentions" tab may require multiple roundtrips to get anything on the next page (and it may just be a few items). I think this makes it not super great since for large accounts, most notifications would probably be non-mentions anyway so the feature won't solve the problem they're having. We need to add a separate backend endpoint for this. I'll add it to a roadmap but probably not super soon.

skiniks:
Sounds good, thanks @gaearon!

surfdude29:
Following on from this, there's now a draft PR open to add a Conversations tab to the Notifications screen:
- #7044

skiniks:
Closing this as #7044 has been merged. Thanks for working on this with me, @gaearon!

PR #3615, opened 2024-04-19
align center post meta in threaded

haileyok:
Prior to adjustments on the left, this PR on the right. (`displayNameStyle` in this one spot, so does not affect styles elsewhere in the app)

Screenshot 2024-04-18 at 8 15 11 PM


haileyok:
Also looks good on web.

Screenshot 2024-04-18 at 8 16 45 PM

gaearon:
The post text is a bit too close to the avatar vertically. Can you push it down just a bit?

In prod it's a bit further down.

haileyok:
that might be it? i can't really tell if its supposed to be 4 or 5, and it looks like mobile needed 2px as well...

gaearon (approved):
ok

PR #3616, opened 2024-04-19
add dimensions to data

haileyok:
We don't have anywhere in this schema right now to place height and width measurements for GIFs, so we have to load the thumbnail _before_ we know the size of container which is janky. Putting it at the end of the URL like this works, and does not break the URL if the user does click through.

Screenshot 2024-04-18 at 9 00 08 PM


gaearon (approved):
sure. why `ww` and `hh`? i guess it would be nice if they won't ever clash with some params they might want to add.

haileyok:
Yea, I don't know if they add any params but I figure that `hh` or `ww` wouldn't be used.

PR #3620, opened 2024-04-19
Persist PDS URL for session resumption

mary-ext:
Fixes the pesky issue where non-blocking session resumption would cause API requests to go through `bsky.social` rather than the intended PDS URL.

This is a little more annoying than it needed to be because there isn't a single function for mapping the agent's state to the intended persisted schema. I *believe* I have caught all the cases (looked at where `upsertAccount` is being called)

**This fixes handle resolution issues during initial page load for _any and all_ accounts that are currently residing outside of bsky.social instances**. (only happens when authenticated, PWI works fine because it goes through AppView)

Should have a slight network improvement as well as it doesn't have to make connection requests to `bsky.social` first before going through the user's actual PDS.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -313,6 +313,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         refreshJwt: agent.session.refreshJwt,
         accessJwt: agent.session.accessJwt,
         deactivated: isSessionDeactivated(agent.session.accessJwt),
+        pdsUrl: agent.pdsUrl?.toString(),
       }
 
       await configureModeration(agent, account)
@@ -405,6 +406,10 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         logger.debug(`session: attempting to reuse previous session`)
 
         agent.session = prevSession
+        if (account.pdsUrl) {
+          agent.api.xrpc.uri = agent.pdsUrl = new URL(account.pdsUrl)
+        }
+
         __globalAgent = agent
         upsertAccount(account)
 
@@ -479,6 +484,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
           refreshJwt: agent.session.refreshJwt,
           accessJwt: agent.session.accessJwt,
           deactivated: isSessionDeactivated(agent.session.accessJwt),
+          pdsUrl: agent.pdsUrl?.toString(),
         }
       }
     },
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -369,7 +369,12 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     async account => {
       logger.debug(`session: initSession`, {}, logger.DebugContext.session)
 
-      const agent = new BskyAgent({service: account.pdsUrl || account.service})
+      const agent = new BskyAgent({service: account.service})
+
+      // restore the correct PDS URL if available
+      if (account.pdsUrl) {
+        agent.pdsUrl = agent.api.xrpc.uri = new URL(account.pdsUrl)
+      }
 
       agent.setPersistSessionHandler(
         createPersistSessionHandler(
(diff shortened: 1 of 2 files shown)


mary-ext:
I took a stab at fixing it in the API client itself so it doesn't have to do that mutation to both `.pdsUrl` and `.api.xrpc.uri` but I wasn't sure about the monkeypatching it necessitated.

gaearon:
Thank you! What's a good way to test it? Like before/after.

gaearon:
Well, I mean — is there a good way to test this assuming I'm not super familiar with the atproto network stuff. E.g. I'm imagining a test plan like

1. Load some user's profile (who should I use)
2. Before: (here's what was wrong in the UI)
3. After: (here's what's corrected now)

That would help me test it. (I can also wait until someone more knowledgeable than me wakes up :)

mary-ext:
Ah, sorry, misunderstood the question.

1. https://bsky.social/profile/genco.me  
  before: leads to on-plc.genco.me ❌  
  after: leads to genco.me ✅

2. https://bsky.app/profile/mary.mraow.party/post/3km5ymk4hhk2z  
  before: fails with handle resolution error ❌  
  after: should load the thread ✅

gaearon:
What happens if your PDS changes?

mary-ext:
with bsky.social being the only multi-instanced PDS, I'm not so sure how it'd work. my assumption is that once you're settled in one of the `*.host.bsky.network` PDS then you shouldn't be moved to anywhere else?

at least for other PDSes, if you've moved your account then your tokens are pretty much invalid and you shouldn't be able to refresh or login with the same PDS.

src/state/session/index.tsx
```diff
@@ -363,16 +369,19 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     async account => {
       logger.debug(`session: initSession`, {}, logger.DebugContext.session)
 
-      const agent = new BskyAgent({
-        service: account.service,
-        persistSession: createPersistSessionHandler(
+      const agent = new BskyAgent({service: account.pdsUrl || account.service})
```
devinivy, line 372: If we restore a session `{ service: A, pdsUrl: B }` into this agent, I believe it will end-up with a different session `{ service: B, pdsUrl: undefined }`.  It might be safer to assign the PDS url in the same way that it would happen via `getSession`, I believe something like this:
```ts
const agent = new BskyAgent({service: account.service})
agent.pdsUrl = account.pdsUrl ? new URL(account.pdsUrl) : undefined
agent.api.xrpc.uri = agent.pdsUrl || agent.service
```

I realize it's a little awkward 🙈 We will have a better way of thinking about this before long, it is in the works!

mary-ext: Ooh, thanks for spotting! Looks like I've forgotten that moving it to the constructor will make `pdsUrl` be undefined.

src/state/session/index.tsx
```diff
@@ -363,16 +369,24 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
       logger.debug(`session: initSession`, {}, logger.DebugContext.session)
 
-      const agent = new BskyAgent({
-        service: account.service,
-        persistSession: createPersistSessionHandler(
+      const agent = new BskyAgent({service: account.service})
+
+      // restore the correct PDS URL if available
+      if (account.pdsUrl) {
+        agent.pdsUrl = agent.api.xrpc.uri = new URL(account.pdsUrl)
```
pfrazee, line 376: We should probs make an API on the agent for this

gaearon:
Let's keep this until the next release to be safe.

src/state/session/index.tsx
mary-ext: I was thinking of moving the pdsUrl to the session object and patching the `.xrpc.api.uri` into a getter for that, wdyt? I think I still have that draft I could pull up

mary-ext: Opened PR https://github.com/bluesky-social/atproto/pull/2431

gaearon (approved):
lgtm

src/state/session/index.tsx
pfrazee: Appreciate it. Going to be some big changes with the OAuth update; will try to roll it into that

PR #3624, opened 2024-04-19
Animate search bar width when cancel appears

mozzius:
Uses the `useLayoutEffect`/`LayoutAnimation` trick on the search bar cancel button. I also put it in a custom hook, `useLayoutAnimation`, which ensures `configureNext` is not run on the first render.

https://github.com/bluesky-social/social-app/assets/10959775/b41e6127-e11a-4bd9-b2b9-d3f71a413d65

https://github.com/bluesky-social/social-app/assets/10959775/1e6282b2-78ac-4b2b-8cc9-fc7d3ea524f6

### DIFF
--- a/src/components/hooks/useLayoutAnimation.ts
+++ b/src/components/hooks/useLayoutAnimation.ts
@@ -1,0 +1,27 @@
+import {useLayoutEffect, useRef} from 'react'
+import {LayoutAnimation} from 'react-native'
+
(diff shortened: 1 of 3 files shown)


gaearon (changes requested):
as discussed out of band, let's do this without a custom hook — and if possible, move the configureNext call next to relevant setStates.

PR #3626, opened 2024-04-19
Add disable autoplay preference and group related settings into a dedicated page

mozzius:
Adds a `disableAutoplay` preference for disabling the autoplay of GIFs. Does not actually link into the GIF player yet, we'll have to wait for the new player to land in (I think) #3605

Also groups the alt text pref and disable haptics pref into one single Accessibility Settings screen, in an effort to reduce the amount of loose toggles in the settings.

https://github.com/bluesky-social/social-app/assets/10959775/c363156e-7745-41c3-8b0f-1751fa60a320

## Test plan

Ensure that the alt text pref and the disable haptics pref still work. Test the new disable autoplay pref by toggling it, I guess.


pfrazee (approved):
Just reviewing the code, LGTM

gaearon (approved):

PR #3628, opened 2024-04-19
Search in PWI

gaearon:
I kept the special empty state for PWI (since we have no suggested follows), the rest moved out of the `hasSession` codepath.

https://github.com/bluesky-social/social-app/assets/810438/8bfbfed6-b7ff-4dfc-ae2b-005920c99a02

https://github.com/bluesky-social/social-app/assets/810438/2757f150-c91d-443f-be7d-5ac87318bcad

## Test Plan

Tried search both logged in and logged out. Switched between tabs.

Also verified directly landing on `/search` works and shows a good empty state.

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -393,13 +393,7 @@ function SearchScreenUserResults({
   )
 }
 
-export function SearchScreenInner({
-  query,
-  primarySearch,
-}: {
-  query?: string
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Works well, still respects opt-outs as well and everything.

One thing we might want to do for later (not sure how we feel about it though) is make it more obvious where the search page is on native. I wouldn't really think to look in the drawer tbh.

PR #3630, opened 2024-04-20
Update Finnish translations 

jaoler:
Numerous new translations and minor fine-tunings. Also removed strings that are obsolete and no longer used.

### DIFF
--- a/src/locale/locales/fi/messages.po
+++ b/src/locale/locales/fi/messages.po
@@ -419,7 +419,7 @@ msgstr "Taiteellinen tai ei-eroottinen alastomuus."
 
 #: src/screens/Signup/StepHandle.tsx:119
 msgid "At least 3 characters"
-msgstr ""
+msgstr "Vähintään kolme merkkiä"
 
 #: src/components/moderation/LabelsOnMeDialog.tsx:246
 #: src/components/moderation/LabelsOnMeDialog.tsx:247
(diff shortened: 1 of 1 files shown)


src/locale/locales/fi/messages.po
```diff
@@ -1208,7 +1208,7 @@ msgstr "Muokkaa ulkoisten sivustojen mediasisältöjen asetuksia"
 
 #: src/view/screens/Settings.tsx:687
 #~ msgid "Danger Zone"
-#~ msgstr ""
+#~ msgstr "Vaaravyöhyke"
```
surfdude29, line 1211: Just a friendly tip, you don't need to bother translating any of the lines starting with `#~` because this means that the string is obsolete and no longer appears in the app (you can delete these in the file if you wish, or just leave them) :)

jaoler: Yes, thank you. 🙂 I knew about the # thing, but I thought that if there's a small chance they might be reinstated at some point, it's good to have them translated.

Why aren't they removed completely?

Yangbblood (commented):
Translations 👌

src/locale/locales/fi/messages.po
jaoler: If thats my job, ill start cleaning 😁

surfdude29: > Why aren't they removed completely?

They don't get removed automatically because sometimes there might just be a tiny change in the English string, like adding a punctuation mark, for example. So if they were removed automatically it would be less convenient for translators because they would either have to redo the translation from scratch for something where the meaning might not actually have changed, or else look up the old version of the file.

pfrazee (approved):
Thanks so much!!

PR #3643, opened 2024-04-22
[GIFs] Add error boundary to GIF picker

mozzius:
Adds an error boundary to the GIF picker. I couldn't reuse our existing `<ErrorBoundary>` since the error screen needs wrapping in a Dialog `<Inner>`, so I added a `renderError` prop to it so we can display a custom error screen.

Simulator Screenshot - iPhone 15 - 2024-04-22 at 14 24 16

Screenshot 2024-04-22 at 14 35 52

### Test plan

Throw some errors in the GIF picker, and see if it behaves as it should

### DIFF
--- a/src/components/dialogs/GifSelect.tsx
+++ b/src/components/dialogs/GifSelect.tsx
@@ -57,7 +57,7 @@ export function GifSelectDialog({
   }
 
   const renderErrorBoundary = useCallback(
-    (error: any) => <DialogError details={error.toString()} />,
+    (error: any) => <DialogError details={String(error)} />,
     [],
   )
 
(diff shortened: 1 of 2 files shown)


src/components/dialogs/GifSelect.tsx
```diff
@@ -54,13 +56,18 @@ export function GifSelectDialog({
       break
   }
 
+  const renderErrorBoundary = useCallback(
+    (error: any) => <DialogError details={error.toString()} />,
```
gaearon, line 60: This isn't safe because technically `error` itself could be `null` or `undefined`.

gaearon: I guess we have this broken in a few other places. I'll punt on fixing that for now.

gaearon (approved):
great, thanks

PR #3645, opened 2024-04-22
remove unnecessary padding from the right hand side of the post in threaded mode

haileyok:
## Why

Noticed over the weekend while testing that there's some strange padding on the right hand side of the post in threaded mode which makes photos or other embeds look skewed in the thread.

This should be safe, because we still have other padding that prevents the content from reaching the side of the screen:

Screenshot 2024-04-22 at 9 16 12 AM
Screenshot 2024-04-22 at 9 16 18 AM

## Test Plan

Open a post in threaded mode that has some embeds before and after to see the difference in alignment.

### Before

Screenshot 2024-04-22 at 9 11 59 AM

### After

Screenshot 2024-04-22 at 9 12 16 AM

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -675,7 +675,6 @@ const styles = StyleSheet.create({
   },
   layoutContentThreaded: {
     flex: 1,
-    paddingRight: 10,
   },
   meta: {
     flexDirection: 'row',


haileyok:
Screenshot 2024-04-22 at 9 25 20 AM

This looks fine on web as well.

gaearon (approved):
yea that looks better, thanks

PR #3647, opened 2024-04-22
remove gate for hiding vertical scroll indicators

haileyok:
Another small PR to remove gates for hiding vertical scroll indicators. I think we are okay with this, but if there's more discussion around it we can hold off. Some notes:

- All Meta apps remove the vertical scroll indicators
- Reddit hides the vertical scroll indicators
- Twitter does not hide it, although their scroll indicator is far less janky than ours (FlatList virtualization...)
- Our indicators actually are sometimes "hidden" as it is, because they don't necessarily conform to the theme. I.e. if your phone is in light mode but the app is in dark mode, you can't see the indicator on the dark background 😅 
- This only affects native clients - not web. Scroll indicators still appear on web.

### DIFF
--- a/src/lib/statsig/gates.ts
+++ b/src/lib/statsig/gates.ts
@@ -3,7 +3,6 @@ export type Gate =
   | 'autoexpand_suggestions_on_profile_follow_v2'
   | 'disable_min_shell_on_foregrounding_v2'
   | 'disable_poll_on_discover_v2'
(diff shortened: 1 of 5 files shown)


gaearon (changes requested):
we're seeing weird results in statsig so let's let this bake a bit more + wait until i run a separate test to verify confidence in the metric lifts feature

haileyok:
SGTM!

PR #3654, opened 2024-04-22
Move to Tenor for GIFs

haileyok:
Move to using Tenor through a proxy instead of GIPHY.

- Rewrite Tenor URLs to correct file format per platform
- Rewrite Tenor URLs to use the Bsky proxy
- Add GIF coder to iOS module


modules/expo-bluesky-gif-view/ios/GifView.swift
```diff
@@ -125,6 +125,7 @@ public class GifView: ExpoView, AVPlayerViewControllerDelegate {
     _ finished: Bool,
     _ imageUrl: URL?
   ) {
+    print(error)
```
gaearon, line 128: rm?

src/lib/strings/embed-player.ts
```diff
@@ -350,21 +337,34 @@ export function parseEmbedPlayerFromUrl(
...
+    const w = urlp.searchParams.get('ww')
+    let dimensions
+    if (h && w) {
+      dimensions = {
+        height: Number(h),
+        width: Number(w),
+      }
+    }
+
+    if (dimensions && id.includes('AAAAC')) {
```
gaearon, line 353: I think this can crash if `id` is `undefined`. I.e. if someone posts `media.tenor.com` with no path after, or if they ever introduce a new URL path format. Let's gracefully fallback if `id` or `filename` aren't strings?

gaearon (approved):

PR #3657, opened 2024-04-23
Mark `Copied to clipboard` toast for translation

surfdude29:
This small PR marks for translation the `Copied to clipboard` toast that is displayed on web when sharing the link to a post.

### DIFF
--- a/src/lib/sharing.ts
+++ b/src/lib/sharing.ts
@@ -1,6 +1,7 @@
 import {Share} from 'react-native'
 // import * as Sharing from 'expo-sharing'
 import {setStringAsync} from 'expo-clipboard'
+import {msg} from '@lingui/macro'
 
 import {isAndroid, isIOS} from 'platform/detection'
 import * as Toast from '#/view/com/util/Toast'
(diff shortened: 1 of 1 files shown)


gaearon:
`_` doesn't seem defined here.

surfdude29:
> `_` doesn't seem defined here.

Yup, just realised that from the lint error, hopefully I fixed it now?

gaearon:
`useLingui` is a hook, so we can only call it from a hook. You'd need to refactor this to `useShareUrl`, maybe like this:

```js
export async function useShareUrl(url: string) {
  const {_} = useLingui()
  return React.useCallback(() => {
    if (isAndroid) {
      await Share.share({message: url})
    } else if (isIOS) {
      await Share.share({url})
    } else {
      setStringAsync(url)
      Toast.show('Copied to clipboard')
    }    
  }, [_, url])
}

```

and then change callsites to do

```js
const shareUrl = useShareUrl(url)

// ...
```

surfdude29:
Sorry, I'm kind of in over my head here tbh, I usually just do translations and I thought this looked like an easy fix, but I'd probably better just close this 😬

PR #3663, opened 2024-04-23
Improve usability of search on web

haileyok:
## Why

- Inconsistent behavior on what the "X"/clear input button does compared to other websites/apps
- No browser history when navigating back through subsequent searches
- Don't dismiss autocomplete results on each keystroke
- Extreme, ultra, incredible jank on web search input

### Navigation History

We can use `navigation.push('Search, {q: term})` on web to create a stack of pages that the user can go "back" through.

Because we need to update the `searchText` state on web _only_ on navigation to the page, we use `useFocusEffect()` to update the value when necessary. Otherwise  the committed search term is _always_ set in the navigation params.

https://github.com/bluesky-social/social-app/assets/153161762/f9ba920e-eee0-4532-b541-880bf39b3467

### Clear Button

The clear button's behavior right now is only slightly changed to mimic other apps. Whenever we focus the search input, it will appear if there is text in the field. Clearing the field will not unfocus the input (something that happens on web right now).

https://github.com/bluesky-social/social-app/assets/153161762/93c823f2-e71e-4564-a049-b6b5d81df354

### Cancel Button

On native:

- When the input is focused and the keyboard is visible, it dismisses the keyboard and hides the autocomplete suggestions.
- When the input is unfocused, it returns the user to the main search screen (i.e. suggested follows)

This is basically the same as what we have right now, except that when the keyboard is visible and we press cancel, it actually clears out the entire query. We already have a button to do that (the X) so we should just be dismissing the keyboard here instead.

On web:

- Same functions as above except...
- If there is search history in the stack, we change the button to `Back` when the input is unfocused. That way, canceling a single search will not destroy the history. We use `navigation.pop()` for this.
- If there is no search history in the stack, we just unfocus the input and hide the autocomplete list.

https://github.com/bluesky-social/social-app/assets/153161762/12fbf514-2f1f-4a4a-9501-aaedfa2d3a87

### Autocomplete results

Instead of using a `setTimeout()` inside of `onChangeText()`, we use the `useThrottledInput()` hook instead to debounce the input. We then use a `useEffect()` to update the results each time the throttled input changes. This prevents the list of items from flashing in and out when going between loading states.

### Jank

Two things were problematic here.

1. We were autoselecting the entire text when we focused the text input. That wasn't really useful on web, so we scope that only to native now via the prop on `TextInput`.
2. We had some funky logic inside of a `useEffect()` that would run each time `queryParam` changed.

We are actually able to simplify the query state by storing the `searchText` and the `queryParam` separately. The only time we rely on `queryParam` is after committing - i.e. by either navigating to a search result or by updating the param with `navigation.setParams()`.

- On native, we use `navigation.setParams()` whenever we submit the search or select an item from the history. We also use this to clear the search when we press `Cancel`.
- On web, we use `navigation.push()` when we submit the search. Whenever we select an item in the handle history, however, we use `navigation.setParams()`. We could use `push()` here as well, but I think it makes more sense for selecting one of these to just update the current screen rather than create a new piece of history.

The TLDR of this is:

- `searchText` is used to perform the autocomplete queries and as the text to submit the next search
- `q` (from the query params) is used as our committed search term and source of truth on what the the search query is showing.

## Test Plan

Performing various searches, navigating back through history, etc. The things to look out for are:

- The input fields no longer act janky when selecting them
- The history stays accurate when navigating back on web
- The search text input stays in sync with the `q` param
- The autocomplete window hides/shows when it should


src/view/screens/Search/Search.tsx
```diff
@@ -671,19 +642,27 @@ export function SearchScreen(
...
-  }, [query, setShowAutocompleteResults, updateSearchHistory])
+    updateSearchHistory(searchText)
+    setQueryTerm(searchText)
+
+    if (isWeb) {
+      navigation.push('Search', {q: searchText})
+    } else {
+      navigation.setParams({q: searchText})
+    }
+  }, [navigation, searchText, updateSearchHistory])
```
haileyok, line 653: As mentioned above, we don't want to push on native because there isn't a back button. Whenever we rework the shell and potentially get rid of the drawer icon, we can revisit.

src/view/screens/Search/Search.tsx
```diff
@@ -786,23 +769,28 @@ export function SearchScreen(
...
               onPress={onPressCancelSearch}
               accessibilityRole="button"
               hitSlop={HITSLOP_10}>
               <Text style={[pal.text]}>
-                <Trans>Cancel</Trans>
+                {isWeb && !inputIsFocused && q ? (
+                  <Trans>Back</Trans>
+                ) : (
+                  <Trans>Cancel</Trans>
+                )}
```
haileyok, line 783: If there's history for us to go back into, we could show "Back" here instead or "Cancel". Not sure if we want to though.

surfdude29:
If I visit a link for a search (e.g. https://social-app-pr-3663.onrender.com/search?q=cats ), **Back** shows up instead of **Cancel**, when there is no history and tapping it doesn't do anything.

This is in Safari on iPadOS 16.6.

IMG_5058

surfdude29:
Sorry to jump in here again as I can see you two are both working hard on this, I wanted to give you a heads up that the user experience in Safari on iPad isn't great if you select the search bar after scrolling down the results a bit, the keyboard appearing causes the entire UI to move up:

https://github.com/bluesky-social/social-app/assets/149612116/24cf449e-f342-4640-a1e5-5454ec0498cf

gaearon (approved):
still not perfect but this a solid set of improvements, thank you.

gaearon:
@surfdude29 should be fixed, thanks.

gaearon:
Also included:

- Fix iOS keyboard jump
- Reuse previous results while typing
- Instantly show profile shell on autocomplete navigation
- Put last clicked recent search at the top

surfdude29:
> @surfdude29 should be fixed, thanks.

Confirmed fixed for me on [main.bsky.dev](https://main.bsky.dev/), thanks for these great search improvements @haileyok and @gaearon! 🙏

PR #3672, opened 2024-04-24
Remove unnecessary `.trimEnd()` and fix facet detection

haileyok:
## Why

See individual commits.


gaearon (approved):
i think this should work. needs more manual testing

gaearon (approved):
ok i'm actually gonna split this and land this part cause it's safe, the rest will come in another PR

Issue #3681, opened 2024-04-24
When the button closes the focus has a border


phieule2024:
**Describe the bug**

When the button closes the focus has a border
**To Reproduce**

1. Open modal register or Sign in 
2. Click at close button

**Expected behavior**

**Screenshots**
![image](https://github.com/bluesky-social/social-app/assets/69315534/4d5556a6-2d62-4ba9-bf7b-387314e09f33)

**Details**

 - Platform: desktop chrome
 - Platform version:
 - App version:

**Additional context**

Resolve
Cause svg is images

at social-app\bskyembed\src\index.css add

svg:focus {
  outline: none;
}

asahelcode:
Done

PR #3682, opened 2024-04-24
Fix issue #3681 (When the button closes the focus has a border)

henrynnamani:


haileyok:
Hey thanks for the PR! I think we need to add this to https://github.com/bluesky-social/social-app/blob/main/bskyweb/templates/base.html and https://github.com/bluesky-social/social-app/blob/main/web/index.html instead. Do you mind doing that? Thanks!

mary-ext:
Wouldn't this remove outlines entirely from the button? What happens if you tab into it now?

mary-ext:
image
Seems to me that the better option would be figuring out why `<FontAwesomeIcon>` here has its `tabindex` set to `-1` instead of undefined

asahelcode:
Okay, I will check that out

Issue #3679, opened 2024-04-24
Issue with the automatic link card generation


Zero3K:
It is trying to make the link card while the user is still typing out the URL which is resulting in an incorrect URL being used for it. It should make the link card after the user has finished typing it out.

Zero3K:
When will this be fixed?

haileyok:
today

Zero3K:
Thanks for fixing it 😄

Issue #3665, opened 2024-04-23
On web - embeds are re-added after being removed


Bossett:
**Describe the bug**
If I remove an embed in the composer, it is re-added as soon as I resume typing

**To Reproduce**

1. Create a post with a URL in it, wait for the embed to appear
2. Remove the embed
3. Add text to the composer

**Expected behavior**

Embeds should stay gone when removed

Bossett:
Chrome 124 on Win 10 - seems to be after every 'word'.

https://github.com/bluesky-social/social-app/assets/114112113/745117dc-1621-4ebc-8157-0e744f1db143

pfrazee:
Very much appreciate it. We think we've found a fix.

PR #3687, opened 2024-04-24
Rewrite the link detection

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/3672

Fixes https://github.com/bluesky-social/social-app/issues/3679
Fixes https://github.com/bluesky-social/social-app/issues/3665

This rewrites the logic that detects links. The main change is that we no longer rely on the cursor position — it works deterministically based on the previous and the next text content. I'm also removing regexes that try to match the URL inside the text, and instead rely on the information we already have from the facet. This also makes it work with newlines.

See inline comments for exact heuristics.

## Test Plan

This is pretty subtle. Just watch the whole thing.

https://github.com/bluesky-social/social-app/assets/810438/4e605352-1276-4026-8044-4eac355b179e

Did some of the same testing on mobile. Running into errors related to image resizer but they don't seem new so I assume it's some DEV-only thing.

### DIFF
--- a/src/view/com/composer/text-input/text-input-util.ts
+++ b/src/view/com/composer/text-input/text-input-util.ts
@@ -53,8 +53,8 @@ export function suggestLinkCardUri(
       continue
     }
     if (
-      /^[.,;!?](\s|$)/m.test(prevTextAfterUri) &&
-      /^[.,;!?]\s/m.test(nextTextAfterUri)
+      /^[)]?[.,:;!?)](\s|$)/m.test(prevTextAfterUri) &&
+      /^[)]?[.,:;!?)]\s/m.test(nextTextAfterUri)
     ) {
       // The link was *already* being followed by punctuation,
       // and now it's followed both by punctuation and a space.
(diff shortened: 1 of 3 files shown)


gaearon:
Some more test scenarios.

https://github.com/bluesky-social/social-app/assets/810438/3842843e-ba9f-4edd-8e9b-e76c6b8a7fb1

src/view/com/composer/text-input/text-input-util.ts
```diff
@@ -1,41 +1,85 @@
...
+    if (/^\s/m.test(nextTextAfterUri)) {
+      // The link is followed by a space, e.g.
+      // "google.com" -> "google.com " or
+      // "google.com." -> "google.com ".
+      // This is a clear indicator we can linkify it.
+      suggestedUris.add(uri)
+      continue
+    }
+    if (
+      /^[.,;!?](\s|$)/m.test(prevTextAfterUri) &&
```
haileyok, line 56: We should add `)` here, it isn't detected in the facets anymore for highlights

Screenshot 2024-04-24 at 9 09 02 AM

haileyok (approved):
Sickkkk 🙏

PR #3688, opened 2024-04-24
Fix Notifications screen header on tablet

mozzius:
`<ViewHeader>` only had a desktop and mobile layout, so for tablets there was no side borders.

Before:

Screenshot 2024-04-24 at 16 58 12

After:

Screenshot 2024-04-24 at 16 58 04


surfdude29:
Works well on iPadOS 16.6 👍

gaearon:
This adds double borders on Moderation. Let's audit existing usages of ViewHeader and fix them consistently?

Screenshot 2024-04-24 at 17 35 33

surfdude29:
FYI the **Liked By** and **Reposted By** pages also don't have side borders on tablet.

PR #3692, opened 2024-04-24
Remove Moderation from the sidebar

gaearon:
It's pretty busy in the left pane. Adjusting moderation settings isn't something you do every day, and it makes sense for them to just live in Settings. Let's remove the shortcut from the left pane. So you'd have to go Settings > Moderation. In Settings, I moved Moderation up so that it goes after Accessibility and Languages. So it's still easy to find.

## Left nav, desktop

Screenshot 2024-04-24 at 21 18 21

## Drawer, mobile web

Screenshot 2024-04-24 at 21 19 05

## Drawer, iOS

Screenshot 2024-04-24 at 21 20 02

## Drawer, Android

(Finally, I don't need to scroll to get to Settings.)

Screenshot 2024-04-24 at 21 19 52

## Settings

I reordered some stuff so these go first:

Screenshot 2024-04-24 at 21 20 41

### DIFF
--- a/src/view/screens/Settings/index.tsx
+++ b/src/view/screens/Settings/index.tsx
@@ -494,6 +494,49 @@ export function SettingsScreen({}: Props) {
           </Text>
         </TouchableOpacity>
         <TouchableOpacity
+          testID="languageSettingsBtn"
+          style={[
+            styles.linkCard,
+            pal.view,
+            isSwitchingAccounts && styles.dimmed,
+          ]}
+          onPress={isSwitchingAccounts ? undefined : onPressLanguageSettings}
+          accessibilityRole="button"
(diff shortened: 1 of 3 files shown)


src/view/screens/Settings/index.tsx
```diff
@@ -494,107 +494,107 @@ export function SettingsScreen({}: Props) {
           </Text>
         </TouchableOpacity>
         <TouchableOpacity
-          testID="preferencesHomeFeedButton"
+          testID="languageSettingsBtn"
```
gaearon, line 497: Changes in this file are just reorders.

Proposed new order:

Screenshot 2024-04-24 at 21 26 52

PR #3693, opened 2024-04-24
Ignore image responses on non-200 status

gaearon:
We currently crash in simulator if Cardy returns a non-200 response (as it does when an image is missing, e.g. for google.com). Let's instead check the response status.

## Test Plan

Verified callsites are resilient to returning undefined.

https://github.com/bluesky-social/social-app/assets/810438/43dda678-9e51-4325-afe7-f8eb062b525a

https://github.com/bluesky-social/social-app/assets/810438/419da788-213e-4dbd-9312-5b00f7e640fa

https://github.com/bluesky-social/social-app/assets/810438/d7f4a669-8383-40a6-b085-ee91658ccbe4

### DIFF
--- a/__tests__/lib/images.test.ts
+++ b/__tests__/lib/images.test.ts
@@ -1,9 +1,10 @@
+import ImageResizer from '@bam.tech/react-native-image-resizer'
+import RNFetchBlob from 'rn-fetch-blob'
+
 import {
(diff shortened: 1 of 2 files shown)


haileyok (approved):
Yep makes sense!

PR #3694, opened 2024-04-24
remove use of `pointerEvents` on iOS profile header

haileyok:
## Why

Using `pointerEvents` on iOS is preventing some areas of the bio from being "swipebackable". Since we now handle touch gestures in this region separately on iOS, we can set `pointerEvents` to `auto` in these areas.

## Test Plan

There were a few places that we know couldn't be swiped from before:

- Banner
- Under the following count
- Very close to the left edge of the screen

On a simulator, these areas appear to be good now. Just needs to be tested in client.


mozzius (approved):
LGTM on simulator, I can try and test on device if needed

gaearon (approved):
confirmed it works on simulator (and didn't before)

PR #3695, opened 2024-04-24
Send Bluesky feeds and suggested follows more data

estrattonbailey:
This PR sends `Accept-Language` to `getSuggestions`, and sends the interests collected during onboarding along with requests to `getSuggestions` and to `getFeed` using a header.

This data will only be shared with feeds owned and/or managed by Bluesky, for the same reasons outlined in #3498. When we have a better permissions model in place, the intention is provide feeds with the data they need to create a good experience.


pfrazee (approved):
LGTM

gaearon (approved):

PR #3696, opened 2024-04-25
Move Notifications up on desktop web

gaearon:
## Before

Screenshot 2024-04-25 at 04 19 13

## After

Screenshot 2024-04-25 at 04 19 48

I think this feels a bit better. Places Feeds and Lists closer together. Groups most commonly used ones at the top. Matches mobile which already uses this order.


pfrazee (approved):
Lmao I was looking at the screenshot and thinking "Honestly I think notifications should go even higher" and then I realized I was looking at the *Before* screenshot. So guess that's a 👍 from me!

PR #3707, opened 2024-04-25
[Session] Add a:a test

estrattonbailey:
This PR:
- copies and duplicates existing `session/index.tsx` to `v1.tsx` and `v2.tsx`
- updates `session/index.tsx` to export either version based on a flag

So the outcome is that the existing session logic is used whether the flag is enabled or not. The next step will be to drop the v2 code into the v2.tsx file.


gaearon (changes requested):
chatted offline, i think we might need a different strategy here because we're not gonna be able to gate this by the user. let's circle back early next week and decide on the approach.

PR #3710, opened 2024-04-25
Special-case bsky.app/download to open share sheet or copy to clipboard

mozzius:
We have a new bsky.app/download path that redirects users to the device-relevant app store. However, if you click the link in the app it tries to do a local navigation to a non-existent page and 404s. This PR opens the share sheet instead.

### Test plan

Click the link in this post: https://bsky.app/profile/bsky.app/post/3kqy4mjw5eo2t

### DIFF
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -150,7 +150,7 @@ export function isBskyDownloadUrl(url: string): boolean {
   if (isExternalUrl(url)) {
     return false
   }
-  return url.endsWith('/download')
+  return url === '/download' || url.startsWith('/download?')
 }
 
 export function convertBskyAppUrlIfNeeded(url: string): string {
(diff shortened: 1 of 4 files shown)


src/lib/strings/url-helpers.ts
```diff
@@ -145,6 +146,13 @@ export function isBskyListUrl(url: string): boolean {
   return false
 }
 
+export function isBskyDownloadUrl(url: string): boolean {
+  if (isExternalUrl(url)) {
+    return false
+  }
+  return url.endsWith('/download')
+}
```
estrattonbailey, line 154: Not sure that endsWith is super safe here since someone like us might toss some query params on there. What do you think about `url.startsWith(BSKY_DOWNLOAD_URL)`?

surfdude29:
When tapping the link in this post in Safari on my iPad it now copies the link to the clipboard 👍

https://social-app-pr-3710.onrender.com/profile/bsky.app/post/3kqy4mjw5eo2t

gaearon (approved):

PR #3713, opened 2024-04-25
embedr: handle out-of-range maxwidth; change default

bnewbold:
Looks like the default width is actually 600px, so I increased to that. Not sure if we should allow larger widths? Or whether that param is de facto even being used?

And don't return an error on out-of-range params, just return range-limited value.


mozzius (approved):
LGTM! Can I call myself a Go developer now?

PR #3719, opened 2024-04-26
Update horizontal padding for Saved feed header

quiple:
Sets the `paddingHorizontal` of the Saved feeds header to 18, just like the headers on other pages, such as Settings, Notifications, etc.


pfrazee (approved):
Thanks!

PR #3720, opened 2024-04-26
Update font size in WhoCanReply

quiple:
Before:

303183412-93034c09-5673-462f-b879-3dcee7ae1739

After:

303183382-f0166361-3a69-4c29-b922-ed899e3b60ab

Fixes an issue in `WhoCanReply.tsx` where a bigger font size was set for `TextLink` only.


pfrazee (approved):
Ah! Lovely, thank you

PR #3729, opened 2024-04-26
remove precacheThreadPostProfiles

haileyok:
## Test Plan

- Press user avatars in the feed
- Press user avatars in post threads

Same behavior of having a minimal shell should exist - the basic info like avatar and display name should be pre-populated on the push.


src/view/com/util/UserAvatar.tsx
```diff
@@ -375,14 +379,29 @@ let PreviewableUserAvatar = (
...
   const {_} = useLingui()
+  const queryClient = useQueryClient()
+
+  const onPress = React.useCallback(() => {
+    const profile: AppBskyActorDefs.ProfileViewBasic = {
+      handle: props.handle,
+      displayName: props.displayName,
+      did: props.did,
+      avatar: props.avatar ?? undefined,
+    }
```
haileyok, line 390: We need to build a little profile based on the data in props here.

gaearon (approved):

src/view/com/util/UserAvatar.tsx
haileyok: Nevermind, let's just pass profile in the props instead.

gaearon (approved):

Issue #3717, opened 2024-04-26
Visual issues with short dates over a year old


Bossett:
**Describe the bug**

The short date format is displayed with insufficient right padding when quoted on iOS (i.e. when posts are old):

![image](https://github.com/bluesky-social/social-app/assets/114112113/eb2517d3-a5dd-41ed-85d8-85b71216a5f3)

**To Reproduce**

1. Quote a post that is over a year old (so the date is the short format - DD/MM/YYY for me here)
2. Observe the post in a feed (profile feed or following feed)

**Expected behavior**

Date should have more padding to the right to give it balance similar to the X hours display:

![image](https://github.com/bluesky-social/social-app/assets/114112113/088bc23a-579e-44c4-b916-1b939ed0d5dd)

**Details**

 - Platform: iOS
 - Platform version: 17.4.1
 - App version: 1.79.0

haileyok:
Fixed in https://github.com/bluesky-social/social-app/pull/3982

PR #3732, opened 2024-04-27
fix timestamp overflow in quotes

haileyok:
## Why

Fixes https://github.com/bluesky-social/social-app/issues/3717

For some reason we were setting `maxWidth` on the text for web and iOS vs using `flexShrink`. We actually use `flex` on Android, so I'm unsure why we were adding additional space here on iOS.

In any case, using `maxWidth` rather than `flex`/`flexShrink` causes problems with `<Text>`. Without `flex`, the text component won't be able to properly break text when it should (or, at least, not well). This was never really a problem because it would break before we reached the end of the container, but now that we have year old posts, its' getting bad.

## Test Plan

So this bit us a tad bit before, so here's a video to display the differences on both iOS and on web with this change. The timestamp does move a bit on iOS, but it actually allows us to see more of the handle so I think that's a win here honestly.

Probably just need to think about if we like this or not. Maybe we want a bit more padding?

https://github.com/bluesky-social/social-app/assets/153161762/e918640d-e2f8-42dc-9f07-83b7c32511a9


estrattonbailey:
Seems net-positive overall. If it looks good on Android I'm +1

haileyok:
This also fixes this problem

image

haileyok:
So it looks like this now

Screenshot 2024-04-28 at 7 33 42 PM

haileyok:
Pushed one more change here:

- Android behaves the same as it always has
- Web stays at the 80% we had before as well
- iOS expands to 90% since we have room. We don't want to use 100% because it looks a little too close to the side of the screen, but 80% is giving a way a lot of free real estate too

Issue #760, opened 2023-05-19
Banner aspect ratio can not be changed on Linux/Firefox


smoyer64:
**Describe the bug**

On Firefox (Linux) 114.0b4, there is no way to change the aspect ration of a banner image.

**To Reproduce**

1. Attempt to load an image suitable for a banner (say 3:1).

**Expected behavior**

I expected to be able to change the aspect ratio.  On Chromium (Linux), there are in fact three buttons but the horizontal and vertical aspect ratios provide very slight changes from a plain square.  In the Android app, loading the same image results in the expected banner orientation (though my banner is a bit too large and does in fact get clippted

**Screenshots**

![bluesky-banner-square-linux-20230519](https://github.com/bluesky-social/social-app/assets/328333/ceb20810-b222-4d57-9ac4-a5247c291684)

**Details**

 - Platform: Xubuntu Linux with the latest Firefox and Chromium
 - Platform version: 22.04
 - App version: Build version TODO (TODO)

pfrazee:
Ah, yes indeed this is not the right behavior

pfrazee:
The cropper should be in the correct aspect ratio for the banner

wrthfl:
This is an issue on Windows too.
Firefox 114.0.2 (64-Bit)

mschwendener:
I used Chrome on Mac when I signed up a few days ago (because it seemed to work better than Firefox), but it's not behaving yet as it should.

mschwendener:
https://bsky.app/profile/marbules.bsky.social/post/3k3cto7gbvj2y

mschwendener:
Uploading / replacing the banner picture should have a cropping option that fits with the banner picture dimensions.

<img width="522" alt="screenshot_4597" src="https://github.com/bluesky-social/social-app/assets/4497705/41828aed-f217-4e3d-b906-efa49096c0d4">

<img width="531" alt="screenshot_4598" src="https://github.com/bluesky-social/social-app/assets/4497705/9098dda7-7c2d-4ed7-8fea-c928e3f3bb72">

mschwendener:
At least let me open somebody's banner picture in a new tab so I can see what's on it? 
 
<img width="586" alt="screenshot_4715" src="https://github.com/bluesky-social/social-app/assets/4497705/7f257114-817e-498c-bf0b-b9b363583b6f">

nicole-ashley:
[...]

PR #3735, opened 2024-04-27
[Issue 760] Use correct aspect ratio in banner cropper

piotrpalek:
This is a fix for the banner cropper issue closes https://github.com/bluesky-social/social-app/issues/760

Before:

image

After:

image

Note that in the "after" image the preview matches what you'd actually see ending up in the banner after clicking "Done"

Previously when uploading an image, you could choose aspect ratios, but none of them would fit the banner aspect ratio. I have modified the banner cropper on web to take into account the aspect ratio of the given width/height provided to the cropper [here](https://github.com/bluesky-social/social-app/blob/main/src/view/com/util/UserBanner.tsx#L67):

```
    onSelectNewBanner?.(
      await openCropper({
        mediaType: 'photo',
        path: items[0].path,
        width: 3000,
        height: 1000,
      }),
    )
```

I have also hidden the options to pick an aspect ratio in the cropper when the width / height is given, as I think it's confusing to pick a custom aspect ratio when we expect a specific width and height, right?

I've also removed the `reject(new Error('Canceled'))` call when pressing "Cancel" on the cropper preview, as it caused the app to crash with that error locally.

### DIFF
--- a/src/lib/media/picker.tsx
+++ b/src/lib/media/picker.tsx
@@ -1,8 +1,9 @@
 import {
+  Image as RNImage,
   openCamera as openCameraFn,
   openCropper as openCropperFn,
-  Image as RNImage,
 } from 'react-native-image-crop-picker'
+
 import {CameraOpts, CropperOptions} from './types'
 export {openPicker} from './picker.shared'
 
--- a/src/lib/media/picker.web.tsx
+++ b/src/lib/media/picker.web.tsx
@@ -1,7 +1,8 @@
 /// <reference lib="dom" />
 
-import {CameraOpts, CropperOptions} from './types'
 import {Image as RNImage} from 'react-native-image-crop-picker'
+
+import {CameraOpts, CropperOptions} from './types'
 export {openPicker} from './picker.shared'
 import {unstable__openModal} from '#/state/modals'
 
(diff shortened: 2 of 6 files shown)


piotrpalek:
@pfrazee not sure how it works here, but PR is ready for review ✅ (hope pinging isn't bad form here 😅)

pfrazee:
NW tagging is fine, we'll give it a look

pfrazee:
Okay this looks great. I had to change how you were dealing with the cancel rejection but otherwise this is perfect. Thanks so much @piotrpalek!

piotrpalek:
gotcha! I'll check it out 🙌

mschwendener:
Thank you! 

MUCH better – but not perfect yet: 

What I see when uploading 
screenshot_4926

what gets saved
screenshot_4925

I'm on a 13-inch MacBook Pro, in Forefox, if that's relevant.

pfrazee:
That might actually just be a mismatch in the ratio of the render vs the crop, and likely it's the render that's wrong

mschwendener:
Happy that what gets saved are **not** rounded corners - maybe also change that in the cropping screen?

piotrpalek:
@pfrazee Would it be OK to follow up on this with a PR removing the rounded corners and maybe fixing the mismatch between preview and rendered image?

PR #3736, opened 2024-04-27
YouTube Music as supported player

mary-ext:
Adds YouTube Music as a supported player


haileyok:
YouTube…why do you do this. Why do you have all these domains!!

Works just fine, thank you!

haileyok (approved):

PR #3737, opened 2024-04-27
Fix issue #3681 (When the button closes the focus has a border)

henrynnamani:


pfrazee:
Issue #3681 seems to have been deleted?

PR #3738, opened 2024-04-27
Mark `Text field` for translation in SubmitView.tsx

surfdude29:
This small PR marks the `Text field` placeholder text in the report dialog for translation.

IMG_5840


surfdude29:
IMG_6114

Is it even necessary to have placeholder text there, given that there's already a prompt `Optionally provide additional information below:` ?

pfrazee:
Uhhh yeah that shouldn't say "Text field." Agree that, not sure it needs a placeholder.

PR #3739, opened 2024-04-27
Don't trigger hovercards on scroll

gaearon:
Instead of activating on pointer enter, activate on the first pointer move within.

This avoids triggering them _just_ due to scroll. Intentionally moving the cursor still triggers it.

https://github.com/bluesky-social/social-app/assets/810438/584d219e-e3b8-40a9-89cc-228c94e72b28


gaearon:
Now that this decreases the unintentional trigger rate, I'm adding them to display names and handles.

I've also bumped the threshold delay to 500ms to further reduce false positives.


https://github.com/bluesky-social/social-app/assets/810438/07fed3f9-341e-4383-9e98-a248b976162c

src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -244,12 +244,20 @@ export function ProfileHoverCardInner(props: ProfileHoverCardProps) {
     }
   }, [prefetchProfileQuery, props.did])
 
-  const onPointerEnterTarget = React.useCallback(() => {
+  const didFireHover = React.useRef(false)
+  const onPointerMoveTarget = React.useCallback(() => {
     prefetchIfNeeded()
```
haileyok, line 249: Unrelated, but this is another good use case for getting a bit more info from the API rather than just `ProfileBasic`. Added to a little list of "why we want this" for Wednesday.

haileyok (approved):
Feels a lot better. The scroll issue was definitely noticeable when using a mouse w/ scroll wheel, but now I don't get those false positives. 🙏

PR #3740, opened 2024-04-28
Account for momentum when hiding minimal shell

gaearon:
Several tweaks to how we toggle the minimal shell based on scroll for native apps:

- If there is a velocity, don't snap scroll on drag end. Instead, wait for the momentum end.
- When snapping, always hide the bar if we scrolled down. (Previously, it was snap to closest position.)
- Bump the threshold for when we always show the shell at the top. (Avoids weird empty space at the top.)

No changes on web.

## Why

This is easier to understand on video (before/after below).

This aims to resolve several interaction problems:

- If you quickly flick the feed down, the bar would previously often snap back up — because it wouldn't take the velocity into account (so the calculation would think you haven't scrolled down "far enough"). With these changes, it will no longer snap "just" at the drag end, and instead will wait for the momentum events to settle. Note we're making an assumption that a non-zero velocity will be followed by momentum events.
- If you move the feed just a _little bit_ down and then release, it would previously snap back up — because we'd snap to the "closest" state, and if you only moved it a bit, the closest state is still "show the top bar". I'm changing this to always hide the bar on scroll down. (The only case in which you'd now ever see it "pop back up" during a drag down is at the very top of the feed where we don't want to show a blank space.)

Overall I think this makes it feels less obtrusive.

### Before

https://github.com/bluesky-social/social-app/assets/810438/632dc9f8-c557-4d14-932a-b0ad1d1d79fb

### After

https://github.com/bluesky-social/social-app/assets/810438/4d7a7fa5-cffd-4aff-a2f3-88d3a6c09bc7

### Before

https://github.com/bluesky-social/social-app/assets/810438/9b06e7bf-9aaa-4738-9148-ca865ce17aaa

### After

https://github.com/bluesky-social/social-app/assets/810438/e6a56c8a-845e-4437-8a98-77f780a915c4

### DIFF
--- a/src/lib/ScrollContext.tsx
+++ b/src/lib/ScrollContext.tsx
@@ -5,6 +5,8 @@ const ScrollContext = createContext<ScrollHandlers<any>>({
   onBeginDrag: undefined,
   onEndDrag: undefined,
   onScroll: undefined,
+  onMomentumBegin: undefined,
+  onMomentumEnd: undefined,
 })
 
 export function useScrollHandlers(): ScrollHandlers<any> {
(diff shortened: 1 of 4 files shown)


mozzius (approved):
Yeah, these feel good. Let's see how they feel in TF

PR #3741, opened 2024-04-28
Remove useEffect from TimeElapsed

gaearon:
It's not needed and forces an unnecessary second render pass.

See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

## Test Plan

Waited a minute, verified timestamps get updated.

### DIFF
--- a/src/view/com/util/TimeElapsed.tsx
+++ b/src/view/com/util/TimeElapsed.tsx
@@ -3,8 +3,6 @@ import React from 'react'
 import {useTickEveryMinute} from '#/state/shell'
 import {ago} from 'lib/strings/time'
 
-// FIXME(dan): Figure out why the false positives
-
 export function TimeElapsed({
   timestamp,
   children,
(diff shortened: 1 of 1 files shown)


src/view/com/util/TimeElapsed.tsx
```diff
@@ -3,8 +3,6 @@ import React from 'react'
 import {useTickEveryMinute} from '#/state/shell'
 import {ago} from 'lib/strings/time'
 
-// FIXME(dan): Figure out why the false positives
```
gaearon, line 6: This is something unrelated about lint and types that's since been addressed.

estrattonbailey (approved):
Nice one

src/view/com/util/TimeElapsed.tsx
```diff
@@ -15,9 +13,11 @@ export function TimeElapsed({
   const tick = useTickEveryMinute()
   const [timeElapsed, setTimeAgo] = React.useState(() => ago(timestamp))
 
-  React.useEffect(() => {
+  const [prevTick, setPrevTick] = React.useState(tick)
+  if (prevTick !== tick) {
+    setPrevTick(tick)
```
AhmedBaset, line 18: isn't that (bad setState in render)? 

I mean this error:
> `Warning: Cannot update a component (`App`) while rendering a different component (`Watever`). To locate the bad setState() call inside `Testing`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

gaearon: That error is about updating a _different_ component. Updating the same component is ok if it's following this specific pattern: https://react.dev/reference/react/useState#storing-information-from-previous-renders

PR #3745, opened 2024-04-28
Fix dropdown immediately closing on Enter

gaearon:
Web-only.

Previously, pressing Enter on a dropdown trigger by mistake caused it to immediately auto-close.

https://github.com/bluesky-social/social-app/assets/810438/75adac58-8925-4082-9c5e-45f73799f3ce

This is because we have a hack that delays the popover toggle until `onPress` (rather than `onPointerDown` as Radix would do by default — see https://github.com/radix-ui/primitives/issues/2418). The hack helps solve the mobile scroll problem, but it _also_ kicks in for presses caused by keyboard events. This doesn't make sense because Radix has already processed the keyboard event too — so we were flipping it right after Radix has flipped it, resulting in a reversal.

The fix is to disable the hack for presses caused by keyboard events.

```diff
          onPress={() => {
+            if (window.event instanceof KeyboardEvent) {
+              // The onPointerDown hack above is not relevant to this press, so don't do anything.
+              return
+            }
            // Compensate for the disabled onPointerDown above by triggering it manually.
            setOpen(o => !o)
          }}
```

I opted to use `window.event` and to not rely on specific kinds of events to try to be resilient to the `react-native-web` implementation — unless we're *sure* it's from *some* keyboard event, we'd still apply the hack.

## Reviewing

The first commit is just a restructure to make this component easier to edit. And to delay some computation until it's needed. The fix is in the second commit. I had to apply the fix both to legacy and ALF menus.

## Test Plan

For both types of popups (non-ALF like Repost and ALF like [...] menu) verify these.

Desktop (Chrome, FF, Safari):

- Can summon it by clicking
- Can summon it by pressing Enter
- Can summon it by pressing Space
- Can Escape out of it

Touch (Chrome mobile emulation, real mobile Safari):

- Can summon it by touching on touch device
- Can scroll past it (starting with a touch on it) without triggering it

### DIFF
--- a/src/components/Menu/index.web.tsx
+++ b/src/components/Menu/index.web.tsx
@@ -134,12 +134,10 @@ export function Trigger({children, label}: TriggerProps) {
             },
             props: {
               ...props,
-              onPointerDown: e => {
-                // Prevent false positive that interpret mobile scroll as a tap.
-                // This requires the custom onPress handler below to compensate.
-                // https://github.com/radix-ui/primitives/issues/1912
-                e.preventDefault()
-              },
+              // No-op override to prevent false positive that interprets mobile scroll as a tap.
+              // This requires the custom onPress handler below to compensate.
+              // https://github.com/radix-ui/primitives/issues/1912
+              onPointerDown: undefined,
               onPress: () => {
                 if (window.event instanceof KeyboardEvent) {
                   // The onPointerDown hack above is not relevant to this press, so don't do anything.
--- a/src/components/Menu/index.web.tsx
+++ b/src/components/Menu/index.web.tsx
@@ -1,27 +1,26 @@
 /* eslint-disable react/prop-types */
 
 import React from 'react'
-import {View, Pressable, ViewStyle, StyleProp} from 'react-native'
-import * as DropdownMenu from '@radix-ui/react-dropdown-menu'
+import {Pressable, StyleProp, View, ViewStyle} from 'react-native'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
+import * as DropdownMenu from '@radix-ui/react-dropdown-menu'
 
+import {atoms as a, flatten, useTheme, web} from '#/alf'
 import * as Dialog from '#/components/Dialog'
 import {useInteractionState} from '#/components/hooks/useInteractionState'
-import {atoms as a, useTheme, flatten, web} from '#/alf'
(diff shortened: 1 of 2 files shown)


src/components/Menu/index.web.tsx
```diff
@@ -135,10 +134,22 @@ export function Trigger({children, label}: TriggerProps) {
             },
             props: {
               ...props,
-              // disable on web, use `onPress`
-              onPointerDown: () => false,
```
gaearon, line 139: The `=> false` part was not doing anything here. React does not take return value of an event handler into account.

The reason this worked before is because this `onPointerDown` overrode `props.onPointerDown` due to spread. I replaced it with `undefined` to make that clearer.

Note that in the other implementation (pre-ALF), there is a `e.preventDefault` call. Shouldn't we add it here? The answer is no. We do not need to replicate it here. The only reason it existed in the non-ALF implementation is because we didn't have control over how props are spread (we used `asChild` there). So we called `e.preventDefault` _not even for the browser's sake_ but because Radix itself checks `e.defaultPrevented` before running its own logic. Since we have control over props in this version, we don't need `e.preventDefault` — and that's why it worked without it. (Again, `=> false` didn't do anything.)

estrattonbailey: Thanks for the comment, much clearer

src/components/Menu/index.web.tsx
```diff
@@ -135,10 +134,22 @@ export function Trigger({children, label}: TriggerProps) {
...
-                control.isOpen ? control.close() : control.open(),
+              // No-op override to prevent false positive that interprets mobile scroll as a tap.
+              // This requires the custom onPress handler below to compensate.
+              // https://github.com/radix-ui/primitives/issues/1912
+              onPointerDown: undefined,
+              onPress: () => {
+                if (window.event instanceof KeyboardEvent) {
+                  // The onPointerDown hack above is not relevant to this press, so don't do anything.
+                  return
+                }
```
estrattonbailey, line 145: Ah nice, good catch

estrattonbailey (approved):
Tested on web, much better

src/components/Menu/index.web.tsx
gaearon: np, this was a solid nerdsnipe — one had me wondering “but what does it do?” and the second one was “but how does it work?” 😅

PR #3746, opened 2024-04-28
[iOS] Fix selecting search input on focus

gaearon:
Makes this work on iOS.

https://github.com/bluesky-social/social-app/assets/810438/399c753d-f351-46e9-bd12-efd065c5ab1e

Doesn't touch Android or web.

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -25,7 +25,7 @@ import {NavigationProp} from '#/lib/routes/types'
 import {augmentSearchQuery} from '#/lib/strings/helpers'
 import {s} from '#/lib/styles'
 import {logger} from '#/logger'
-import {isNative, isWeb} from '#/platform/detection'
+import {isIOS, isNative, isWeb} from '#/platform/detection'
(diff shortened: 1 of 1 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -684,6 +684,12 @@ export function SearchScreen(
                 })
               } else {
                 setShowAutocomplete(true)
+                if (isIOS) {
+                  // We rely on selectTextOnFocus, but it's broken on iOS:
+                  // https://github.com/facebook/react-native/issues/41988
+                  textInput.current?.setSelection(0, searchText.length)
```
gaearon, line 690: It's not ideal that we rely on `searchText.length` here. I think in theory it could be stale. I haven't been able to trigger that tho. Seems fine.

haileyok (approved):
Verified that it works on both Android and iOS now. I might spend 30 minutes looking at the code here to see what it's supposed to be doing, been working on another PR for Text anyway.

Emmaccen:
LGTM 👍

PR #3748, opened 2024-04-29
Increase search `TextInput` hit area and improve the related UI

haileyok:
## Why

The hit area on the search text input is pretty small right now. What we can do instead is make the wrapper a `Pressable` and call `.focus()` whenever we press anywhere inside the gray pill.

## Test Plan

Open the app, tap around on the input. It should be a lot easier to focus now.

Also tested that the clear input button still works as well as our "select text on focus".

https://github.com/bluesky-social/social-app/assets/153161762/31c3dc26-79c4-41c2-8f3e-dc831e7908fe

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -11,6 +11,8 @@ import Animated, {
   FadeIn,
   FadeOut,
   LinearTransition,
+  useAnimatedStyle,
+  withSpring,
 } from 'react-native-reanimated'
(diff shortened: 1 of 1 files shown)


gaearon:
I also added a bunch of other tweaks to this interaction, let's see how that feels

gaearon:
- Select when clicking on icon too
- Only show "x" when you typed something
- Remove Clear button on rest (it's enough to focus it, on web you can press "x" in the field)
- Added input width animation

gaearon (approved):

gaearon (approved):

PR #3749, opened 2024-04-29
[Clipclops] Temp codegenerated lexicon

mozzius:
# Stacked on https://github.com/bluesky-social/social-app/pull/3754

Don't merge yet

PR #3750, opened 2024-04-29
[Clipclops] New clipclop dialog

mozzius:
# Stacked on https://github.com/bluesky-social/social-app/pull/3754 and https://github.com/bluesky-social/social-app/pull/3749

Need to figure out what the initial page looks like when there's no search term

simulator_screenshot_DE3C04B0-4814-4F5E-89D1-CAC42A5B4C61


src/components/dms/NewChat.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+  const isSearching = throttledSearch.length > 0
+
+  const searchQuery = useActorSearch({
+    query: throttledSearch || 'temp',
+  })
+
+  const {data, isLoading, isError, refetch} = isSearching
+    ? searchQuery
+    : searchQuery
```
haileyok, line 92: Oo, we recently used `keepPreviousData` for this in the search screen, I think it would be great here too. I just modified the query hook to take in a bool for if we want to keep the prev data, see: https://github.com/bluesky-social/social-app/blob/main/src/state/queries/actor-autocomplete.ts#L19

Then we implemented like this: https://github.com/bluesky-social/social-app/blob/main/src/view/screens/Search/Search.tsx#L811

We can probably just get rid of the throttle too here I think. So combined, we'll only get a spinner on the first character of the query instead of disappearing results.

haileyok: Oh, actually I think we can just use the existing `useActorAutocompleteQuery()` here?

haileyok: Huh, I did some testing and the `searchActors` query is really unreliable anyway. Sometimes I don't get any results for something that _definitely_ should return something, like `haileyok`. I do get the results though in the typeahead.

I think after we merge this https://github.com/bluesky-social/social-app/pull/3765 that the typeahead will be the most reliable here.

src/screens/Messages/List/index.tsx
```diff
@@ -79,33 +80,43 @@ export function MessagesListScreen({}: Props) {
...
+          onRetry={isError ? refetch : undefined}
+        />
+        <NewChat onNewChat={onNewChat} />
+      </>
     )
   }
 
   return (
-    <View>
+    <View style={a.flex_1}>
```
haileyok, line 111: Here I think we want `CenteredView` with `sideBorders`. (Pretty sure it still needs `flex: 1`)

haileyok:
No new chat button on web when on the desktop breakpoint.

Screenshot 2024-04-30 at 12 37 12 AM

haileyok:
So stoked for this!! Tested it out again tonight and was able to talk to myself from native and web 😆

src/screens/Messages/List/index.tsx
mozzius: This has a bunch of problems unfortunately, doesn't work on iPad.

mozzius:
Fixed the desktop breakpoint, I think it's ready to go

src/screens/Messages/List/index.tsx
haileyok: Oh well that's not good...now i'm curious how it looks everywhere else in the app 😳

haileyok (approved):

PR #3752, opened 2024-04-29
Improve search screen perf

gaearon:
The first few commits extract components with no changes to their logic and add some `memo`s.

The main real change is https://github.com/bluesky-social/social-app/commit/e46aac205c808caea2311c15b55455013eeec0f2. It renders both the autocomplete *and* the result list so that switching between those is less laggy. This is especially noticeable on iOS.

Finally, there's a few drive-by changes. https://github.com/bluesky-social/social-app/commit/16a13c5347e224a28bfe6c7612c94f9859da6a34 removes subdomain matching — it is not very useful now that we properly rank results, and it looks a bit out-of-place in the list. Also, https://github.com/bluesky-social/social-app/commit/36485ea7e9d7db5eddd5b95ac85c4d8cab2ec9bf adds some missing hitslops to the recent searches.

## Test Plan

The usual for this screen.

https://github.com/bluesky-social/social-app/assets/810438/735b2a72-6164-4395-9866-f11b452d5010

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -478,7 +478,6 @@ export function SearchScreen(
   const pal = usePalette('default')
   const {track} = useAnalytics()
   const setDrawerOpen = useSetDrawerOpen()
-  const moderationOpts = useModerationOpts()
   const setMinimalShellMode = useSetMinimalShellMode()
   const {isTabletOrDesktop, isTabletOrMobile} = useWebMediaQueries()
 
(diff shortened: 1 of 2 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -850,6 +864,7 @@ function AutocompleteResults({
     </>
   )
 }
+AutocompleteResults = React.memo(AutocompleteResults)
```
geobde, line 857: @gaearon why you dont just wrap the component with memo?
maybe I'm missing some context

gaearon: would need a Babel plugin to keep function names readable without adding an extra indentation level due to writing them twice

PR #3754, opened 2024-04-29
[Clipclops] Add screen to view and send clip clops

haileyok:
Let's not merge until we release update today since it has the new native code for keyboard controller.


mozzius (approved):
LGTM, `DmServiceUrlStorage` works well

src/screens/Messages/Conversation/MessageInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [message, setMessage] = React.useState('')
+
+  const inputRef = React.useRef<TextInput>(null)
+
+  const onSubmit = React.useCallback(() => {
+    onSendMessage(message)
+    setMessage('')
+    setTimeout(() => {
+      inputRef.current?.focus()
+    }, 100)
```
haileyok, line 26: note to self: maybe we don't need this timeout

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  // We use this to know if we should scroll after a new clop is added to the list
+  const isAtBottom = useRef(false)
+
+  // Because the viewableItemsChanged callback won't have access to the updated state, we use a ref to store the
+  // total number of clops
+  const totalMessages = useRef(10)
+
+  // TODO later
+  const [_, setShowSpinner] = useState(false)
```
haileyok, line 66: This works - has been tested against fake data - but the hacked state doesn't make it great to work with. Once we have some paginated results though it will be straight forward.

src/screens/Messages/Temp/useDmServiceUrlStorage.tsx
```diff
@@ -0,0 +1,64 @@
...
+  serviceUrl: '',
+  setServiceUrl: () => {},
+})
+
+export const useDmServiceUrlStorage = () =>
+  React.useContext(DmServiceUrlStorageContext)
+
+export function DmServiceUrlProvider({children}: {children: React.ReactNode}) {
+  const [serviceUrl, setServiceUrl] = React.useState<string>('')
+  const {getItem, setItem: setItemInner} = useAsyncStorage('dmServiceUrl')
```
haileyok, line 33: just using async storage directly here since we don't want to pollute zod schema

Issue #3724, opened 2024-04-26
Search Results on WebApp doesn't close when clicking other components


xtianpu:
**Describe the bug**
1. Search Results on Web doesn't close when clicking other components

**To Reproduce**

1. Type on search bar on the right
2. Click on other components or the result

**Expected behavior**

1. It should be closed, so user can be more focused like X did. Also, there are unneeded scrollbars on the right and bottom of the search results.

**Screenshots**
![image](https://github.com/bluesky-social/social-app/assets/8172235/869d5968-0c50-4e9b-8e4a-5c871b299e67)

**Details**

 - Platform: Chrome
 - Platform version: 124.0.6367.91
 - App version: 1.79.0

**Additional context**

mary-ext:
I believe that's completely intentional at the moment

haileyok:
Fixed here https://github.com/bluesky-social/social-app/pull/3663

gaearon:
Clicking outside still wouldn't hide it. This just needs to be turned into a popover imo.

PR #3758, opened 2024-04-29
Hide search suggestion on clicking outside search field

joonmo-you:
The current behavior, as @gaearon mentioned [here](https://github.com/bluesky-social/social-app/issues/3724#issuecomment-2080387930), allows the search suggestion open even when clicking outside the search field. Implementing an event handler for blurring would resolve this issue temporarily. Additionally, I concur with the suggestion to evolve the search suggestion into a popover.

### DIFF
--- a/src/view/shell/desktop/Search.tsx
+++ b/src/view/shell/desktop/Search.tsx
@@ -180,6 +180,10 @@ export function DesktopSearch() {
     setIsActive(false)
   }, [])
 
+  const onBlur = React.useCallback(() => {
+    setIsActive(false)
+  }, [])
+
   return (
     <View style={[styles.container, pal.view]}>
       <View
@@ -197,6 +201,7 @@ export function DesktopSearch() {
             returnKeyType="search"
             value={query}
             style={[pal.textLight, styles.input]}
+            onBlur={onBlur}
             onChangeText={onChangeText}
             onSubmitEditing={onSubmit}
             accessibilityRole="search"
--- a/src/view/shell/desktop/Search.tsx
+++ b/src/view/shell/desktop/Search.tsx
@@ -213,7 +213,7 @@ export function DesktopSearch() {
       document.removeEventListener('focusin', handleFocusIn)
       document.removeEventListener('focusout', handleFocusOut)
     }
-  }, [])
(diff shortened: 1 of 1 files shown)


gaearon (changes requested):
Doing this on blur wouldn't be sufficient because that would break it for keyboard users. Need to be able to navigate the list with keyboard without having it disappear.

joonmo-you:
> Doing this on blur wouldn't be sufficient because that would break it for keyboard users. Need to be able to navigate the list with keyboard without having it disappear.

I was naive, overlooking users who navigate using a keyboard. I resolved the issue by addressing two scenarios: Firstly, ensuring the search field remains open when focused elements remain within it. Secondly, closing the search field when focus shifts outside of it, either by clicking elsewhere or navigating away using the keyboard.

joonmo-you:
https://github.com/bluesky-social/social-app/assets/57599679/7d9ee692-5b57-4341-a75c-a16d99265789

zidanegimiga:
Hey @joonmo-you, were you also adding user's search history there. Referencing issue #3859

joonmo-you:
> Hey @joonmo-you, were you also adding user's search history there. Referencing issue #3859

I've just been through the issue you referred to. I am not sure the issue should be discussed here. I think the features related to search history would be better addressed separately.

src/view/shell/desktop/Search.tsx
```diff
@@ -180,8 +185,38 @@ export function DesktopSearch() {
...
+    }
+
+    document.addEventListener('focusin', handleFocusIn)
+    document.addEventListener('focusout', handleFocusOut)
+
+    return () => {
+      document.removeEventListener('focusin', handleFocusIn)
+      document.removeEventListener('focusout', handleFocusOut)
+    }
+  }, [])
```
pfrazee, line 216:
```suggestion
  }, [searchViewRef, setIsActive])
```

pfrazee: Update effects

pfrazee:
Thanks so much @joonmo-you for handling the requested change. It seems to work exactly right. @gaearon you good w/this?

PR #3760, opened 2024-04-29
switch branch for installing `react-native-bottom-sheet`

haileyok:
Just switches to using the main branch vs the `4.6.1` branch. I did:

- Merged Discord's 4.6.1 branch - what we were using - into main
- Reverted https://github.com/gorhom/react-native-bottom-sheet/pull/1288. See https://github.com/bluesky-social/react-native-bottom-sheet/commit/4af77ba9158fd2356650c952f6e8ff6ae9e8ec0b
- Tested that accessibility now works.
- Tested that closing the dialog w/ the keyboard open still works (this verifies that the 4.6.1 branch is now on main).

We shouldn't need to mess with this anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "@atproto/api": "^0.12.5",
     "@bam.tech/react-native-image-resizer": "^3.0.4",
     "@braintree/sanitize-url": "^6.0.2",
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
🙏

gaearon (approved):
lg

PR #3761, opened 2024-04-30
sentry errors for captcha web views and registration attempts

haileyok:
## Why

We should have some more robust error reporting around signups. This adds Sentry logs to the most sensitive areas of that flow:

- Captcha webview errors
- Other cases where a captcha isn't set (we shouldn't hit this one but just in case?)
- Any failure after submitting the registration request

## Test Plan

Not really sure how to test this one, aside from waiting for these to come through on Sentry. For what it's worth, we already had some of these coming through but this makes the errors a little easier to set up alerts for using the "Signup Flow Error:" prefix.

### DIFF
--- a/src/screens/Signup/StepCaptcha/CaptchaWebView.tsx
+++ b/src/screens/Signup/StepCaptcha/CaptchaWebView.tsx
@@ -3,7 +3,6 @@ import {StyleSheet} from 'react-native'
 import {WebView, WebViewNavigation} from 'react-native-webview'
 import {ShouldStartLoadRequest} from 'react-native-webview/lib/WebViewTypes'
 
-import {logger} from '#/logger'
 import {SignupState} from '#/screens/Signup/state'
 
 const ALLOWED_HOSTS = [
@@ -27,7 +26,7 @@ export function CaptchaWebView({
   stateParam: string
   state?: SignupState
   onSuccess: (code: string) => void
-  onError: () => void
+  onError: (error: object) => void
 }) {
   const redirectHost = React.useMemo(() => {
     if (!state?.serviceUrl) return 'bsky.app'
@@ -57,7 +56,7 @@ export function CaptchaWebView({
 
       const code = urlp.searchParams.get('code')
       if (urlp.searchParams.get('state') !== stateParam || !code) {
-        onError()
+        onError({error: 'Invalid state or code'})
         return
       }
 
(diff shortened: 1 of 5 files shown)


src/screens/Signup/StepCaptcha/CaptchaWebView.tsx
```diff
@@ -74,6 +74,13 @@ export function CaptchaWebView({
       onShouldStartLoadWithRequest={onShouldStartLoadWithRequest}
       onNavigationStateChange={onNavigationStateChange}
       scrollEnabled={false}
+      on
```
gaearon, line 77: ?

src/screens/Signup/StepCaptcha/CaptchaWebView.tsx
```diff
@@ -26,7 +26,7 @@ export function CaptchaWebView({
   stateParam: string
   state?: SignupState
   onSuccess: (code: string) => void
-  onError: () => void
+  onError: (error: object) => void
```
gaearon, line 29: Maybe not important in this particular case but in general we should get out of the habit of assuming that an error has a particular type. E.g. you technically can `throw null` or `throw undefined` or `throw 'foo'` and it's important that this doesn't trip or break error reporting.

gaearon (approved):
ok but test the error paths for real errors (e.g. make the iframe trigger `onError` for real etc)

src/screens/Signup/StepCaptcha/CaptchaWebView.tsx
haileyok: Ah, yea here these do have an `object` value, since they come from `nativeEvent`. But for consistency sake going forward, sure lets switch to `unknown` here 👍

gaearon: Hmm are passing the event itself? If so, does it turn into something reasonable when we stringify it later?

haileyok: The `nativeEvent` is just an object of various values that comes back from the native code when it fires the event. In this case, `nativeEvent` has the following values (depending on where the error comes from)

Screenshot 2024-04-29 at 10 25 04 PM
Screenshot 2024-04-29 at 10 25 12 PM

haileyok:
iframe error might not be reliable (I think onLoad will actually always fire here) so let's track with a few different things:

- User didn't complete the captcha in 20 seconds
- User pressed the back button on the captcha page (likely indicates they encountered some sort of error)

Screenshot 2024-04-29 at 10 23 36 PM
Screenshot 2024-04-29 at 10 23 51 PM

haileyok:
@gaearon Can you think of any other meaningful ways we could get an error from here, mainly on web? The `onHttpError` and `onError` do fire correctly on web, though we can't get errors from the iframe. I think these are pretty good indicators though of errors (over 20s to complete and pressing back), and if we see a spike we can take it as a problem.

PR #3728, opened 2024-04-26
[Session] V2 (V2)

estrattonbailey:

PR #3762, opened 2024-04-30
Session fixes, pt. 1

gaearon:

These changes are supposed to be 100% safe.
See individual commits. I'll comment inline on a few things.

## Test Plan

Careful reading. The point of these is that largely you can apply them without testing because they "obviously" work.

Also, some logging in / logging out / switching accounts.

Should be no observable changes.

### DIFF
--- a/src/lib/hooks/useAccountSwitcher.ts
+++ b/src/lib/hooks/useAccountSwitcher.ts
@@ -1,6 +1,9 @@
 import {useCallback} from 'react'
+import {msg} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
 
 import {useAnalytics} from '#/lib/analytics/analytics'
+import {logger} from '#/logger'
 import {isWeb} from '#/platform/detection'
 import {SessionAccount, useSessionApi} from '#/state/session'
 import {useLoggedOutViewControls} from '#/state/shell/logged-out'
@@ -8,6 +11,7 @@ import * as Toast from '#/view/com/util/Toast'
 import {LogEvents} from '../statsig/statsig'
 
 export function useAccountSwitcher() {
+  const {_} = useLingui()
   const {track} = useAnalytics()
   const {selectAccount, clearCurrentAccount} = useSessionApi()
   const {requestSwitchToAccount} = useLoggedOutViewControls()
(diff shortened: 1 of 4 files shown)


src/lib/hooks/useAccountSwitcher.ts
```diff
@@ -31,21 +35,26 @@ export function useAccountSwitcher() {
...
         }
-      } catch (e) {
-        Toast.show('Sorry! We need you to enter your password.')
+      } catch (e: any) {
+        logger.error(`switch account: selectAccount failed`, {
+          message: e.message,
+        })
         clearCurrentAccount() // back user out to login
+        setTimeout(() => {
+          Toast.show(_(msg`Sorry! We need you to enter your password.`))
```
gaearon, line 53: This is new but it just shows a toast.

This happens only if something fails (e.g. `await selectAccount`) above. I don't think we currently have a situation actually triggering this (`selectAccount` calls `initSession` but `initSession` doesn't yet have any codepaths that propagate errors up — so effectively I don't think this toast would ever show until we make more changes).

src/screens/Login/ChooseAccountForm.tsx
```diff
@@ -38,15 +39,22 @@ export const ChooseAccountForm = ({
...
+            })
+            track('Sign In', {resumedSession: true})
+            setTimeout(() => {
+              Toast.show(_(msg`Signed in as @${account.handle}`))
+            }, 100)
+          } catch (e: any) {
+            logger.error('choose account: initSession failed', {
+              message: e.message,
+            })
+            onSelectAccount(account)
```
gaearon, line 56: This is the main change here. It ensures that if there's an error from `initSession` (which, in my understanding, technically can't happen yet — but we could propagate errors in the future properly, which would active this codepath). If this codepath starts firing in the future, an error during `initSession` will now preselect the account in which you're trying to log in (and ask you to enter the password). Previously, you'd silently stay on the previous "select an account" form with no visual feedback about the failure.

(I'm actually not sure this is better. It's not ideal that we kick you to log in even if the problem isn't solved by a password — e.g. if it's a network issue. But let's capture the intent of who you were trying to log in as, and show some visual feedback.)

We know this state exists so we can iterate on it later.

src/state/persisted/schema.ts
```diff
@@ -19,12 +22,26 @@ const accountSchema = z.object({
...
+ * The current account. Stored in the `currentAccount` field.
+ *
+ * In previous versions, this included tokens and other info. Now, it's used
+ * only to reference the `did` field, and all other fields are marked as
+ * optional. They should be considered deprecated and not used, but are kept
+ * here for backwards compat.
+ */
+const currentAccountSchema = accountSchema.extend({
+  service: z.string().optional(),
+  handle: z.string().optional(),
```
gaearon, line 35: This relaxes the type a little bit so that `did` is the only required field. We should be careful because this means the app can write a less "complete" version of `currentAccount`, and if ever roll this back, the old version of the code would not be able to work with that (and would also believe falsehoods about it due to wrong types where these fields were required).

In this PR, there are no changes to how `currentAccount` is written though — so that should be fine.

src/state/session/index.tsx
```diff
@@ -618,20 +618,24 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
 
       logger.debug(`session: persisted onUpdate`, {})
 
-      if (session.currentAccount && session.currentAccount.refreshJwt) {
-        if (session.currentAccount?.did !== state.currentAccount?.did) {
+      const selectedAccount = session.accounts.find(
+        a => a.did === session.currentAccount?.did,
+      )
```
gaearon, line 623: This just effectively ensures we don't read `session.currentAccount` and instead read `session.accounts.find(a => a.did === session.currentAccount?.did)`.

So `session.accounts` is always the source of truth for the actual data. However, note that `session.currentAccount?.did` plays a decisive role in whether we're going to "find" an account or not. This keeps the existing code (e.g. logout relying on setting `currentAccount` to `undefined`) working.

pfrazee (commented):
All seems reasonable

PR #3765, opened 2024-04-30
remove trailing period from query term in `useActorAutocompleteQuery`

haileyok:
## Why

Right now, if there's a trailing period in the search term, then no results are returned for the query.

https://github.com/bluesky-social/social-app/assets/153161762/593ef5c1-9b42-4b4b-9fa2-c345f65f1c0a

Instead, we can remove the trailing period (after converting it to lowercase and trimming) to get the expected result. We should get this fixed on the backend as well.

## Test Plan

Test various handles. Adding a period to the end of the autocomplete query should still return the query with the suggested actor(s).

https://github.com/bluesky-social/social-app/assets/153161762/68c980b7-eecd-40dc-913c-c7f29250930b

### DIFF
--- a/src/state/queries/actor-autocomplete.ts
+++ b/src/state/queries/actor-autocomplete.ts
@@ -23,11 +23,17 @@ export function useActorAutocompleteQuery(
   const moderationOpts = useModerationOpts()
   const {getAgent} = useAgent()
 
-  prefix = prefix.toLowerCase()
+  const prefixTrimmed = React.useMemo(() => {
+    let trimmed = prefix.toLowerCase().trim()
+    if (trimmed.endsWith('.')) {
+      trimmed = trimmed.slice(0, -1)
+    }
+    return trimmed
+  }, [prefix])
 
   return useQuery<AppBskyActorDefs.ProfileViewBasic[]>({
     staleTime: STALE.MINUTES.ONE,
-    queryKey: RQKEY(prefix || ''),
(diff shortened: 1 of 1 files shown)


haileyok:
I flagged this in Slack for a backend fix as well.

I was going to implement a fix for the `searchActors` query but I noticed that `searchActors` gives some really weird results. Like `haileyok` doesn't return anything but `hailey` returns me as the first result 🤔 Flagged this as well. Seems like there's something else going on there.

src/state/queries/actor-autocomplete.ts
```diff
@@ -23,11 +23,17 @@ export function useActorAutocompleteQuery(
   const moderationOpts = useModerationOpts()
   const {getAgent} = useAgent()
 
-  prefix = prefix.toLowerCase()
+  const prefixTrimmed = React.useMemo(() => {
+    let trimmed = prefix.toLowerCase().trim()
+    if (trimmed.endsWith('.')) {
```
gaearon, line 28: Why special-case a single period? Does it not happen with other punctuation or more than one period?

haileyok: I think a trailing period is the only thing I'd expect to still have a valid result. I.e.

- `pfrazee..` would be invalid
- `pfrazee,` would be invalid

gaearon (approved):

Issue #3766, opened 2024-04-30
Notifications not removed when they are read in the app


bluelhf:
**Describe the bug**
Unread notifications are not cleared from the system notifications when they are read in the app

**To Reproduce**

1. Receive a notification
2. Open the app
3. Navigate to the notifications page
4. Tap the notification to read it
5. Notice that the notification has not disappeared from the system notification area

**Expected behavior**

When the notification is read in the app, the corresponding notification should be removed from the system tray

**Screenshots**

The notifications have been read:
![Screenshot showing un-highlighted, read notifications in the Bluesky app](https://github.com/bluesky-social/social-app/assets/52505120/13e6d883-f577-4048-a464-e2b1ca7c073a)

...yet they still appear in the system tray...
![Screenshot showing the same notifications still appearing in the system tray](https://github.com/bluesky-social/social-app/assets/52505120/e50bf44a-51c8-4331-b741-50b9388c3253)

...causing a frustrating red bubble :(
![Screenshot showing the Bluesky app icon with a red notification bubble](https://github.com/bluesky-social/social-app/assets/52505120/31b765f4-6f07-4a4c-9120-17af3f66968a)

**Details**

 - Platform: Android
 - Platform version: OxygenOS 14
 - App version: 1.79.0

**Additional context**

PR #3768, opened 2024-04-30
Notifications not removed when they are read in the app

omar0404:
[#3766](https://github.com/bluesky-social/social-app/issues/3766)


mary-ext:
This seems like the wrong approach? It only makes sense if the notification was recent. If there are multiple unread notifications you'd risk marking them all as read, and you'd have no idea which one is new.

omar0404:
@mary-ext i thought since we are switching to notifications tab ([here](https://github.com/bluesky-social/social-app/pull/3768/files#diff-95dc5c62bd776732e735749152d63d9cafeaa046c4bfcf367570640cdf05e207L138)) this means user unread all of them. i think facebook works this way.

PR #3773, opened 2024-04-30
Add kawaii mode

mozzius:
Adds kawaii mode. Activate by adding `?kawaii=true` and deactivate by adding `?kawaii=false` - kawaii mode is persisted in local storage. Web only

Kawaii mode swaps the logo and adds attribution

There are some minor style tweaks that should only apply to kawaii mode - please make sure that that is the case


pfrazee (approved):
Havent tested but code review lg

gaearon:
What about `?uwu=true`? Or is that a convention. React site does `uwu`

mozzius:
Hono went with `kawaii`, we could do `uwu` but kawaii is just what Emily suggested

haileyok (approved):
Glad we finally got around to doing this, it's been needed for a while!

haileyok:
Tested on renderio, looks good.

src/view/shell/Drawer.tsx
```diff
@@ -262,6 +264,17 @@ let DrawerContent = ({}: {}): React.ReactNode => {
               href="https://bsky.social/about/support/privacy-policy"
               text={_(msg`Privacy Policy`)}
             />
+            {kawaii && (
+              <Text type="md" style={pal.textLight}>
+                Logo by{' '}
```
surfdude29, line 269: Sorry to butt in, what about wrapping this section in `<Trans>` so that it can be translated into Japanese?

heikadog:
holy shit

lucidillusions:
love it ! 

can we have some different icons/shades for like/repost in kawaii mode please?

PR #3776, opened 2024-04-30
Add shutdown message to For You feed

pfrazee:
https://github.com/bluesky-social/social-app/assets/1270099/ecc8796a-de21-4bcb-a635-3ba1682c3b34

Note: performance is pretty rough when changing pinned feeds. It might be worth doing a deep dive on what gets triggered in response to it.

## Testing

- Scneario 1
  - Pin "For You" and "Discover"
  - Go to "For You." Click "Remove feed."
  - Feed should be removed
- Scenario 2
  - Pin "For You" and save but unpin "Discover"
  - Go to "For You." Click "Replace"
  - Feed should be removed, discover should be pinned
- Scenario 2
  - Pin "For You" and unsave "Discover"
  - Go to "For You." Click "Replace"
  - Feed should be removed, discover should be pinned

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -111,6 +111,10 @@ export const RECOMMENDED_SAVED_FEEDS: Pick<
   'type' | 'value' | 'pinned'
 >[] = [DISCOVER_SAVED_FEED, TIMELINE_SAVED_FEED]
 
+export const KNOWN_SHUTDOWN_FEEDS = [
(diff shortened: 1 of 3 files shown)


haileyok:
Checked the three different flows, they work fine.

When I first switched to the branch, I was in a weird state where I got stuck loading. Switching back to main would fix it, then switching to this again resulted in the same error.

I made another account though from the the 1.80 tag, then went straight to this branch and it worked out fine. Repeated that twice without problems. Someone else should try the same flow though from main -> this branch to see if there are any problems.

gaearon (approved):

PR #3779, opened 2024-05-01
[Session] Consistently copy SessionAgent fields

gaearon:

This aligns all places where we construct `SessionAgent` so that the fields are copied in the same way. In cases where we have `|| false` fallbacks, I removed them since we never compare them to `false` explicitly, and falsy is enough.

Also, this fixes a case where `emailAuthFactor` was being "lost" (when creating `refreshedAccount`).

In an upcoming PR, we'll actually unify these callsites, but for now it's just targeted fixes.

## Test Plan

Verified switching between accounts works.

For the `emailAuthFactor` bug, verified that refreshing the `/settings` page directly consistently shows "Require 2FA" toggled on for a 2FA account. Previously, it would show it on every second time.


pfrazee (approved):
Yeah lgtm

PR #3780, opened 2024-05-01
[Session] Move deactivated field off the session object

gaearon:
This didn't make sense because `prevSession` is only used as something we assign to `agent.session` (or pass to `agent.resumeSession` that does that assignment for us). But sessions have no `.deactivated` fields — only accounts do.

The only actual place where we depend on this field is right below, so I extracted it to a variable.

## Test Plan

Careful reading. Following types. Grepping `deactivated` to see usages.


estrattonbailey (approved):

pfrazee (approved):
Sure lgtm

PR #3781, opened 2024-05-01
[Session] Refactor to prepare for V2

gaearon:
This is a series of (supposedly safe) refactors extracted from https://github.com/bluesky-social/social-app/pull/3728 and moving us closer to its final state. They are intended to be safe and are not intended to change any functionality except one test-only case (https://github.com/bluesky-social/social-app/commit/c57da3a7f0c0de63233fb29e0375c10e73318326).

## Test Plan

Read individual commits. The usual login/logout/switch pass. I'll be doing more thorough testing when I finish the stack.


estrattonbailey (approved):
Tested changes on web/ios, minus the last commit, but that looks good to me too. Idk if you're ready for a stamp but here's one if it helps unblock! Happy to look again later.

PR #3784, opened 2024-05-01
Add PrivacyManifest info to comply with new App Store guidelines

haileyok:
## Why

Deadline is looming tomorrow. I believe this should be an automatic process in 0.74 (my early testing resulted in the build process generating one of these files for us!) However, there's various upgrades that need to happen and I think it would be safer for us to not prematurely upgrade these without a bit of testing.

Documentation for the app config can be found here: https://docs.expo.dev/guides/apple-privacy/

Apple's documentation can be found here: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

### `NSPrivacyAccessedAPICategoryFileTimestamp`

See https://github.com/react-native-async-storage/async-storage/pull/1075
See https://github.com/expo/expo/blob/c50b313944694af95c7ac22799eb33495332fb11/packages/expo-application/ios/PrivacyInfo.xcprivacy
See https://docs.sentry.io/platforms/react-native/data-management/apple-privacy-manifest/#create-privacy-manifest-in-expo

None of the other options listed here seem to apply to us https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393

### `NSPrivacyAccessedAPICategoryDiskSpace`

See https://github.com/expo/expo/blob/c50b313944694af95c7ac22799eb33495332fb11/packages/expo-application/ios/PrivacyInfo.xcprivacy

The other options listed here obviously do not apply to us. https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278397. I verified that Sentry does _not_ send along any disk space information on iOS which I thought maybe it did. `NSPrivacyAccessedAPICategoryDiskSpace` also is not listed in Sentry's docs.

### `NSPrivacyAccessedAPICategoryBootTime`

See https://github.com/expo/expo/blob/c50b313944694af95c7ac22799eb33495332fb11/packages/expo-device/ios/PrivacyInfo.xcprivacy
See https://docs.sentry.io/platforms/react-native/data-management/apple-privacy-manifest/#create-privacy-manifest-in-expo

I am pretty sure this covers the options here https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278394

### `NSPrivacyAccessedAPICategoryUserDefaults`

See https://github.com/expo/expo/blob/c50b313944694af95c7ac22799eb33495332fb11/packages/expo-constants/ios/PrivacyInfo.xcprivacy
See https://docs.sentry.io/platforms/react-native/data-management/apple-privacy-manifest/#create-privacy-manifest-in-expo

There's a separate option for accessing `NSUserDefaults` from apps, app clips, or extensions within the app group. I verified that our extension does _not_ access `NSUserDefaults` so we do not need to add that.

None of the other options in https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401 seem to apply to us.

### Step One

After checking out, you should be able to run a `yarn prebuild --clean` and find a `PrivacyManifest.xcprivacy` inside of `ios/Bluesky`. 

### Step Two

Hope that Apple doesn't find some reason that we need another one of these to be set. They have not said anything about those, so we should be okay.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2946,15 +2946,15 @@
     mv "~2"
     safe-json-stringify "~1"
 
-"@expo/cli@0.17.8":
-  version "0.17.8"
-  resolved "https://registry.yarnpkg.com/@expo/cli/-/cli-0.17.8.tgz#4abe0d8c604b73a6e1d0a10f34e993cbf1cbad42"
-  integrity sha512-yfkoghCltbGPDbRI71Qu3puInjXx4wO82+uhW82qbWLvosfIN7ep5Gr0Lq54liJpvlUG6M0IXM1GiGqcCyP12w==
+"@expo/cli@0.17.10":
+  version "0.17.10"
+  resolved "https://registry.yarnpkg.com/@expo/cli/-/cli-0.17.10.tgz#7dd5e2b4a01f5d29698c431729a19878fbd806f5"
+  integrity sha512-Jw2wY+lsavP9GRqwwLqF/SvB7w2GZ4sWBMcBKTZ8F0lWjwmLGAUt4WYquf20agdmnY/oZUHvWNkrz/t3SflhnA==
   dependencies:
     "@babel/runtime" "^7.20.0"
     "@expo/code-signing-certificates" "0.0.5"
     "@expo/config" "~8.5.0"
-    "@expo/config-plugins" "~7.8.0"
+    "@expo/config-plugins" "~7.9.0"
     "@expo/devcert" "^1.0.0"
     "@expo/env" "~0.2.2"
     "@expo/image-utils" "^0.4.0"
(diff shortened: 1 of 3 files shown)


gaearon (approved):
okay

yarn.lock
```diff
@@ -12078,24 +12066,24 @@ expo-web-browser@~12.8.2:
     compare-urls "^2.0.0"
     url "^0.11.0"
 
-expo@^50.0.8:
-  version "50.0.14"
-  resolved "https://registry.yarnpkg.com/expo/-/expo-50.0.14.tgz#ddcae86aa0ba8d1be3da9ad1bdda23fa539dc97d"
-  integrity sha512-yLPdxCMVAbmeEIpzzyAuJ79wvr6ToDDtQmuLDMAgWtjqP8x3CGddXxUe07PpKEQgzwJabdHvCLP5Bv94wMFIjQ==
+expo@^50.0.17:
```
gaearon, line 12069: I can't seem to find the changelog, do you know where it is?

PR #3785, opened 2024-05-01
Performance improvements: structural sharing & moderation opts context

pfrazee:
This PR does two things that work in conjunction to reduce needless re-renders.

## 1. Fix to structural sharing in the preferences object

"Structural sharing" is a technique in `useQuery` which ensures that the same JS object is returned when a response is `==` to the previous one. This stops needless hook triggers.

We've had this enabled for the preferences object for a while, but our SDK puts a `Date` object inside the `getPreferences` response and `useQuery` doesn't handle date comparisons. By replacing the method with a slightly modified version that handles `Date` objects, we get structural sharing back and therefore reduce a number of needless re-renders.

## 2. Move moderation opts into a context

The `useModerationOpts` hook is scattered throughout the codebase, often in leafs of the component tree. By moving it into a context, we remove a lot of object creations and reduce the size of the hook dependency-tree.

## Results

When rendering a thread screen on web, the react flamegraph gave the following results:

|Before|After|
|-|-|
|image|image|

The JS profiler gave the following:

|Before|After|
|-|-|
|image|image|

I'm not entirely sure I'm measuring these outcomes the most precise way, but it seems clear this yields some kind of benefit.


src/state/preferences/moderation-opts.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const override = useContext(moderationOptsOverrideContext)
+  const {currentAccount} = useSession()
+  const prefs = usePreferencesQuery()
+  const {labelDefs} = useLabelDefinitions()
+  const hiddenPosts = useHiddenPosts() // TODO move this into pds-stored prefs
+
+  const value = useMemo<ModerationOpts | undefined>(() => {
+    console.log('hit')
```
gaearon: log

src/state/preferences/moderation-opts.tsx
```diff
@@ -0,0 +1,59 @@
...
+              {
+                did: BSKY_LABELER_DID,
+                labels: DEFAULT_LOGGED_OUT_LABEL_PREFERENCES,
+              },
+            ],
+        hiddenPosts: hiddenPosts || [],
+      },
+      labelDefs,
+    }
+  }, [override, currentAccount, labelDefs, prefs.data, hiddenPosts])
```
gaearon, line 52: you only use `currentAccount?.did` so let's calculate that outside (or make it a dependency) so that other changes to `currentAccount` don't affect it

gaearon: oh, can we do this for `prefs.data?.moderationPrefs` too

src/state/preferences/moderation-opts.tsx
pfrazee: fixed

src/state/preferences/moderation-opts.tsx
pfrazee: Done

pfrazee: And done

gaearon (approved):
looking great

gaearon:
small related one, haven't checked if it makes any difference https://github.com/bluesky-social/social-app/pull/3786

PR #3789, opened 2024-05-01
[Clipclops] Improve message list behaviour

mozzius:
- Fix line height
- Bold unread last message
- Last message time elapsed indicator
- Own messages are prefixed with "You: "

### DIFF
--- a/src/screens/Messages/List/index.tsx
+++ b/src/screens/Messages/List/index.tsx
@@ -172,7 +172,7 @@ function ChatListItem({chat}: {chat: TempDmChatDefs.ChatView}) {
   const {currentAccount} = useSession()
 
   let lastMessage = _(msg`No messages yet`)
-  let lastMessageSentAt = null
+  let lastMessageSentAt: string | null = null
   if (TempDmChatDefs.isMessageView(chat.lastMessage)) {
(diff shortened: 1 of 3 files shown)


src/screens/Messages/List/index.tsx
```diff
@@ -171,8 +172,17 @@ function ChatListItem({chat}: {chat: TempDmChatDefs.ChatView}) {
   const {getAgent} = useAgent()
 
   let lastMessage = _(msg`No messages yet`)
+  let lastMessageSentAt = null
   if (TempDmChatDefs.isMessageView(chat.lastMessage)) {
-    lastMessage = chat.lastMessage.text
+    if (chat.lastMessage.sender?.did === getAgent().session?.did) {
```
gaearon, line 177: No reading mutable stuff during render pls. Use `currentSession` from `useSession`.

gaearon (approved):
lg with small change

src/screens/Messages/List/index.tsx
mozzius: Ohh ok I'll have to change that in a few places

gaearon (approved):

PR #3795, opened 2024-05-01
[Session] Derive currentAccount from accounts + currentAccountDid

gaearon:
We should be able to reconstruct `currentAccount` on demand from `accounts` and only track the current DID instead. Although we've wanted to make a bigger change to this in #3728 (and reconstruct it from agent), I'm doing a more incremental step here.

## Test Plan

Careful reading. Verify login/logout/switch still works. Do it in another tab too.


estrattonbailey (approved):
LGTM!

PR #3798, opened 2024-05-01
make smol a bit bigger

haileyok:
Screenshot 2024-05-01 at 9 31 45 AM


haileyok:
The fading header still operates fine with this larger logo on Chrome, Safari, and Firefox

PR #3800, opened 2024-05-01
Retry clops

estrattonbailey:


src/state/messages/convo.ts
```diff
@@ -66,6 +73,17 @@ export type ConvoState =
...
+  item: ConvoItem,
+): item is ConvoItem & {type: 'message'} {
+  if (!item) return false
+  return (
+    item.type === 'message' ||
+    item.type === 'deleted-message' ||
+    item.type === 'pending-message'
+  )
+}
+
```
haileyok, line 86: I think we want ConvoItem` to always include `sender` so that we can use a single `MessageItem` for rendering these. Then we can handle different logic based on the `type`.

haileyok (approved):
Tested this with intermittent connection, works great.

PR #3802, opened 2024-05-01
[Clipclops] Replace 🐴 with real icon

mozzius:
Sorry :(

Also replaces FAB button envelope with a plus


haileyok:
Won't fix.

haileyok (approved):
With much regret, approving.

mozzius:
RIP to 🐴

PR #3804, opened 2024-05-01
[Clipclops] Clop menu, leave clop, mute/unmute clop

mozzius:
- Convo menu
  - Go to profile
  - Mute/unmute notifications
  - Leave convo w/ confirm
  - Block/Report (TODO)
- Muted state on list view
- Long press list view item for menu

Leave convo confirm prompt:

Screenshot 2024-05-01 at 23 08 02


haileyok (approved):
👍 Might want to handle the `onMutate` states as well if we have these options (like leave chat) available from the list.

PR #3809, opened 2024-05-02
Update Korean localization

quiple:
Update the Korean localization of added and changed strings.

* Localizes strings related to messaging.

Thank you!


pfrazee (approved):
Thanks quiple!!

PR #3811, opened 2024-05-02
[Session] Extract resumeSession out

gaearon:
`resumeSession` is a thin wrapper over `initSession` which sets `setIsInitialLoad(false)` to hide the splash and then becomes effectively the same as `initSession`. I was wondering if I could get rid of it, and the only usage of `resumeSession` was from top-level `InnerApp` components which, conveniently, were the ones who needed `isInitialLoad` in the first place (to hide the splash). So I removed `resumeSession` from the Session API and removed `isInitialLoad` from its context.

## Why?

I want to pare it down, and `resumeSession` doesn't depend on any knowledge that isn't already available through the Session API. So it made sense to me to move it out to keep that code more focused. Conceptually we have one instance of the `InnerApp` component (it's just forked between two platforms). The local usage also makes sense because it lets us get rid of context for something that's really a very local thing that only one component cares about.

## Test Plan

Verify splash screen still gets hidden on session init. Verify switching users still works.

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -16,6 +16,7 @@ import {useLingui} from '@lingui/react'
 import {useQueryClient} from '@tanstack/react-query'
 
 import {Provider as StatsigProvider} from '#/lib/statsig/statsig'
+import {logger} from '#/logger'
 import {init as initPersistedState} from '#/state/persisted'
 import {Provider as LabelDefsProvider} from '#/state/preferences/label-defs'
 import {Provider as ModerationOptsProvider} from '#/state/preferences/moderation-opts'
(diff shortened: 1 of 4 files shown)


src/App.native.tsx
```diff
@@ -53,27 +55,41 @@ import {listenSessionDropped} from './state/events'
...
+        setIsReady(true)
+      }
+    }
     const account = readLastActiveAccount()
     resumeSession(account)
-  }, [resumeSession, _])
+  }, [initSession])
+
+  useEffect(() => {
+    return listenSessionDropped(() => {
```
gaearon, line 84: Fixed to return the unsub, I think this was previously leaking.

haileyok (approved):
Encountered the incorrect pinned feeds issue once when switching back and forth between accounts, but appears to be the existing issue rather than a new one. Switching accts works, and account is loaded before splash disappears (tested just setting `isReady` to false and watched the logs).

PR #3812, opened 2024-05-02
[Session] Extract selectAccount out

gaearon:
This is similar to https://github.com/bluesky-social/social-app/pull/3811 in spirit and is based on the same insight. `selectAccount` has nothing to do with session management per se, it's just a wrapper around `initSession` that tracks the progress of that call. It makes sense to make such state tracking local to where the action was triggered in the UI code. So I'm moving that state back into components.

I moved the `isSwitchingAccounts` state variable out of the Session API and into the calling components which now call `initSession` directly. For `AccountList`, I've plumbed it down from the component that owns the switch press handler.

## Test Plan

Use the account switcher, both the popover and the settings one. It feels a bit janky but I replicated the existing logic. This should make it a bit easier to change the UI logic if we'd like to later because it's all local state now.

### DIFF
--- a/src/components/AccountList.tsx
+++ b/src/components/AccountList.tsx
@@ -16,12 +16,14 @@ export function AccountList({
   onSelectAccount,
   onSelectOther,
   otherLabel,
+  isSwitchingAccounts,
 }: {
   onSelectAccount: (account: SessionAccount) => void
   onSelectOther: () => void
   otherLabel?: string
+  isSwitchingAccounts: boolean
 }) {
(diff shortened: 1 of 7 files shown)


haileyok:
Small thing when selecting an account that I noticed. My selection is highlighted as long as I am still pressing in, but as soon as I release the highlight moves down to "Other Account"

https://github.com/bluesky-social/social-app/assets/153161762/c25df29a-b087-47d1-a79a-612fd50253bf

haileyok (approved):

PR #3813, opened 2024-05-02
[WIP] Threaded composer

mary-ext:
image

image

## Checklist

- [x] The actual threading UI
- [x] Decent-ish keyboard navigation (I think we can do better, perhaps by making use of Ctrl/Meta-Up/Down to control focus?)
- [ ] External embeds
- [ ] Tenor GIFs
- [ ] Quote embeds
- [ ] Image embeds
- [ ] Post language
- [ ] Gated reply interactions
- [ ] Self-labeling
- [ ] Send the post ([demo](https://github.com/mary-ext/demo-bsky-thread-posting))

## Bonus checklist

- [x] ALF compliant composer

## Problem

1. This is the last component that makes use of MobX
    - Might be worth removing it here?
    - It doesn't seem worth for MobX removal to be in a separate PR, it'd potentially require double the work for old and new
2. The language picker button is now even more problematic for smaller displays thanks to this addition

<details>
<summary>Archived</summary>

https://github.com/bluesky-social/social-app/assets/148872143/4dd86c45-7851-48b4-812f-da16b690994f

</details>

### DIFF
--- a/src/state/shell/composer.tsx
+++ b/src/state/shell/composer.tsx
@@ -1,10 +1,11 @@
 import React from 'react'
 import {
+  AppBskyActorDefs,
   AppBskyEmbedRecord,
   AppBskyRichtextFacet,
   ModerationDecision,
-  AppBskyActorDefs,
 } from '@atproto/api'
+
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 
 export interface ComposerOptsPostRef {
@@ -34,6 +35,7 @@ export interface ComposerOpts {
   onPost?: () => void
   quote?: ComposerOptsQuote
   mention?: string // handle of user to mention
+  /** @deprecated */
   openPicker?: (pos: DOMRect | undefined) => void
   text?: string
   imageUris?: {uri: string; width: number; height: number}[]
--- a/src/view/com/composer-new/NewComposer.tsx
+++ b/src/view/com/composer-new/NewComposer.tsx
@@ -262,6 +262,7 @@ let Post = ({
                 : _(msg`What's up?`)
               : _(msg`Write another post`)
           }
+          grow={index === 0 && !hasNext}
           setRichText={onRichTextChange}
           onNewLink={onNewLink}
           onError={onError}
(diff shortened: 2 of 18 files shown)


mary-ext:
gauging interest :smile: cc @haileyok @pfrazee 

just need to know if this is something y'all want to pursue towards, because I have questions around the current image manipulation code and whether we could get rid of the last remaining MobX code on the app

Spacedesolate:
Worst part of threads is having to use unroll bot, or clicking through replies to get to next commented reply, esp on those with 10+ posts.

So I think it would be great if threads are marked as such, instead of just posting as tweets and replies. 

This way someone could post the entire tweet of 1000 words in one box, get to see how and where breakpoints would be applied, modify those if they wish.

And from readers side, one click option in 3 dot settings to unroll or view full post can make things lot easier.

mary-ext:
@Spacedesolate that's pretty much out of scope for this work irregardless if this makes it in or not (I'm just an outside contributor, I have no insight over what Bluesky intends to do)

consider making a new issue if this feature ever makes it in ^^

pfrazee:
Ooooh this is a very bold feature to pick up. We _are_ interested in it. I'm quite worried that it's going to take some extensive testing and edgecase handling, and that we'll end up letting it languish because we're focusing on other things. My worst-case outcome is that you do a bunch of work that we don't honor. Let me talk to the crew tomorrow.

Regarding the mobx, that's just there because we were afraid to mess with the code. We do want to ditch it, it's just a finicky system.

pfrazee:
Chatted with the team. We're pretty excited for this feature, and we think the approach you're taking -- putting the changes in a separate component -- makes this work pretty well. That keeps it safe from conflicts during its development.

So: if you're keen to keep working on this, we absolutely welcome it!

mary-ext:
Nice! I'll see if I can figure out a nice way to tackle the image manipulation, since that's really the only major thing blocking this at the moment

mary-ext:
Wrapped up with implementing DMs on my own Bluesky client, so I'm going to pick this up again, rebasing off https://github.com/bluesky-social/social-app/pull/3925/commits/8a12b606ee09b5e74fad243400f6279bbe99b99d

mary-ext:
Closing in favor of #4163

PR #3814, opened 2024-05-02
🚧 XRPC rework

matthieusieben:
This adapts the social-app to the changes introduced by https://github.com/bluesky-social/atproto/pull/2434

This PR should not be merged before those changes are published in an NPM package.


src/lib/link-meta/link-meta.ts
```diff
@@ -62,8 +63,9 @@ export async function getLinkMeta(
     const controller = new AbortController()
     const to = setTimeout(() => controller.abort(), timeout || 5e3)
 
+    const serciceUrl = await agent.getServiceUrl()
     const response = await fetch(
-      `${LINK_META_PROXY(agent.service.toString() || '')}${encodeURIComponent(
+      `${LINK_META_PROXY(serciceUrl.toString() || '')}${encodeURIComponent(
```
matthieusieben, line 68: Is there a way we could do this without having to have access to the `.getServiceUrl` ? This is the only place this method (from `@atproto/api`) is needed for.

PR #3816, opened 2024-05-02
✅ Fix "Download CAR file" on mobile

matthieusieben:
This change allows to download the CAR file in mobile apps.

### DIFF
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -273,23 +273,19 @@ export async function saveToDevice(
     const permissions =
       await StorageAccessFramework.requestDirectoryPermissionsAsync()
 
-    if (permissions.granted) {
-      const fileUrl = await StorageAccessFramework.createFileAsync(
-        permissions.directoryUri,
-        filename,
-        type,
-      )
-
-      await writeAsStringAsync(fileUrl, encoded, {
-        encoding: EncodingType.Base64,
-      })
-    } else {
-      // Permissions denied, fallback to sharing
-
-      await withTempFile(filename, encoded, type, async tmpFileUrl => {
-        await Sharing.shareAsync(tmpFileUrl, {mimeType: type})
-      })
+    if (!permissions.granted) {
+      throw new Error('Permission denied')
     }
+
+    const fileUrl = await StorageAccessFramework.createFileAsync(
+      permissions.directoryUri,
+      filename,
+      type,
+    )
+
+    await writeAsStringAsync(fileUrl, encoded, {
+      encoding: EncodingType.Base64,
+    })
   }
 }
 
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -17,6 +17,7 @@ import ImageResizer from '@bam.tech/react-native-image-resizer'
 import {Buffer} from 'buffer'
 import RNFetchBlob from 'rn-fetch-blob'
 
+import {logger} from '#/logger'
 import {isAndroid, isIOS} from 'platform/detection'
 import {Dimensions} from './types'
 
@@ -257,7 +258,7 @@ export async function saveBytesToDisk(
   type: string,
 ) {
   const encoded = Buffer.from(bytes).toString('base64')
-  await saveToDevice(filename, encoded, type)
+  return await saveToDevice(filename, encoded, type)
 }
 
 export async function saveToDevice(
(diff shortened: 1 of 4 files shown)


matthieusieben:
Capture d’écran 2024-05-02 à 10 25 39

haileyok:
I'll check this today if Eric doesn't get around to it 👍

matthieusieben:
FYI i was able to test on web & ios. I couldn't get my android emulator to work (for an unrelated reason I believe).

src/view/screens/Settings/ExportCarDialog.tsx
```diff
@@ -18,20 +19,17 @@ export function ExportCarDialog({
...
     }
-    // eg: https://bsky.social/xrpc/com.atproto.sync.getRepo?did=did:plc:ewvi7nxzyoun6zhxrhs64oiz
-    const url = new URL(agent.pdsUrl || agent.service)
-    url.pathname = '/xrpc/com.atproto.sync.getRepo'
-    url.searchParams.set('did', agent.session.did)
-    return url.toString()
-  }, [currentAccount, getAgent])
+    const did = agent.session.did
+    const res = await agent.com.atproto.sync.getRepo({did})
+    await saveBytesToDisk('repo.car', res.data, res.headers['content-type'])
```
haileyok, line 31: At least on a simulator the problem I've been having with this (or really, the underlying `Share`) is that it keeps opening in mail on Android with this mime type (I think it's ending up as `application/x-octet-stream`). I tried just removing the mime altogether worked to give me the full list of options, but there isn't a "save" option.

With that said, could be that I'm on a simulator. Do regular devices have this same problem?

src/view/screens/Settings/ExportCarDialog.tsx
```diff
@@ -18,20 +19,17 @@ export function ExportCarDialog({
...
+      return // shouldnt ever happen
     }
-    // eg: https://bsky.social/xrpc/com.atproto.sync.getRepo?did=did:plc:ewvi7nxzyoun6zhxrhs64oiz
-    const url = new URL(agent.pdsUrl || agent.service)
-    url.pathname = '/xrpc/com.atproto.sync.getRepo'
-    url.searchParams.set('did', agent.session.did)
-    return url.toString()
-  }, [currentAccount, getAgent])
+    const did = agent.session.did
+    const res = await agent.com.atproto.sync.getRepo({did})
```
haileyok, line 30: Aside from the issue with opening in mail on Android, this promise can take a while to resolve. For me it takes about 10 seconds or so for the download to complete (my CAR file is about 15 MB)

I think what we should do is

- Keep the old functionality on web. The browser should be handling the download there I think.
- Set the button to disabled with a spinner inside of it once you press download on native, so that the user knows something is going on.

I have some local changes that do all of that, but wondering about this mail issue right now 🙏

src/view/screens/Settings/ExportCarDialog.tsx
matthieusieben: I'll try on my native device.

src/view/screens/Settings/ExportCarDialog.tsx
matthieusieben: The main reason for me to change this behavior is that I want to remove the code that builds the download URL. Indeed, with the oauth refactor coming up, that URL won't be as straightforward to build and I don't want to add complex logic in the client, when the building of the url is already handled by the client when making the API call. For this reason, I would prefer if we could keep the same behavior on web.

I agree that having a loading state would be significantly better. I didn't add one because I wanted my changes to be minimal. If you have branch with these changes, can you link it here? Or simply add the commit with the loader on this branch?

haileyok: Ahh makes sense! Yes I can push those up here in a bit, would have the other day but didn't want to push to your branch w/o asking first 😄 will clean it up and push it up here a bit later today!

matthieusieben: I added a loader.

src/view/screens/Settings/ExportCarDialog.tsx
matthieusieben: The android share menu does not allow to "save file" as ios does (some custom file manager apps do enable that but the default one does not).

So to actually share on device, I had to change the code for android. There is still a fallback to regular sharing through apps if the user denied access to disk for the save operation.

I tested using my own (Google Pixel 8 Pro) phone and a lot of apps *do* appear in the sharing menu.

matthieusieben:
@haileyok Everything should be good now. Could you kindly review this again :-)

haileyok (approved):
Okay I fixed a few styling and localization nits. Tested this on all three platforms and it does work very well! Thank you for digging at the Expo docs to figure out how to do this correctly on Android (their new permissionless API is pretty interesting too!)

I also removed the fallback to sharing because it causes some issues. For example, if you open the file explorer and then press back instead of selecting a destination directory, it will try and open Gmail (at least on the simulator). I'll test this on a variety of devices today once it's merged to make sure this works on various Android versions, but I'm optimistic this will be okay (API 27 works fine for me on the simulator).

Also added some Sentry logging if errors are encountered at any point.

src/lib/media/manip.ts
```diff
@@ -240,3 +251,64 @@ function normalizePath(str: string, allPlatforms = false): string {
...
+export async function saveToDevice(
+  filename: string,
+  encoded: string,
+  type: string,
+) {
+  try {
+    if (isIOS) {
+      const tmpFileUrl = await withTempFile(filename, encoded)
+      await Sharing.shareAsync(tmpFileUrl, {UTI: type})
+      safeDeleteAsync(tmpFileUrl)
```
matthieusieben, line 273: @haileyok  Be careful here. although you do delete the tmpFile, the tmpDirUri will *not* be deleted!

The tmp file/dir will also not be deleted if Sharing.shareAsync fails.

You might want to do:

```ts
  const tmpDirUri = joinPath(documentDirectory as string, String(uuid.v4()))
  await makeDir
  try {
    await writeFileInDir
    await shareFile
    return true
  } finally {
    deleteDirAsync
  }
```

PR #3817, opened 2024-05-02
Allow unmuting and unblocking of curate lists, but not the opposite

estrattonbailey:
We were still allowing users to mute/block user lists. We made this mistake a long time ago, and at the time, backend actually allowed user lists to block and mute other users. We've since updated this so that only moderation lists work in this way, but some users still have user lists that are blocked or muted.

The intention is to NOT allow users to block/mute user lists. But if they have them blocked/muted bc of this mistake, we should allow them to un-do this. That's what this PR does.


pfrazee (approved):
Didnt test but at a glance the logic looks right

estrattonbailey:
Tested both scenarios, should be good

PR #3819, opened 2024-05-02
[Clipclop] Input polish

mozzius:
- Closes keyboard on drag
- Adjust padding
- Enable min shell mode


haileyok (approved):
Will be good to feel this on a real device but it feels fine in a simulator at least!

PR #3823, opened 2024-05-02
[Clipclops] Create `useKeyboardVerticalOffset` hook

mozzius:
The `KeyboardAvoidingView` requires adjustment to position the input correctly. Previously, I had trial-and-error'd the numbers requires for `keyboardVerticalOffset`. However, since the gap looks different on my base iPhone 14 vs the iPhone 15 simulator, I am fairly confident in saying that it's device specific.

On iOS: seems like it's the `top` safe area inset
On Android: seems like it's the navigation bar height

To handle this, I moved the logic out to a custom hook

### DIFF
--- a/src/screens/Messages/Conversation/MessagesList.tsx
+++ b/src/screens/Messages/Conversation/MessagesList.tsx
@@ -1,8 +1,10 @@
-import React, {useCallback, useRef} from 'react'
+import React, {useCallback, useEffect, useRef} from 'react'
 import {
+  Dimensions,
   FlatList,
   NativeScrollEvent,
   NativeSyntheticEvent,
+  Platform,
   View,
 } from 'react-native'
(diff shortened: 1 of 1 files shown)


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -175,3 +177,26 @@ export function MessagesList() {
...
+      },
+    )
+    return () => subscription.remove()
+  }, [])
+
+  return Platform.select({
+    ios: topInset,
+    android: screenWindowDifference,
+    default: 0,
+  })
```
haileyok, line 201: i remember us having a problem with a `Platform.select` before in prod but can't remember what exactly it ended up being linked to. Do you remember?

haileyok (approved):
Let's pay attention to if it works when available on TF, but it does work in the simulator.

PR #3825, opened 2024-05-02
Only run "disable min shell on foregrounding" behaviour if on the home screen

mozzius:
Currently, the "disable min shell on foregrounding" behaviour is in a `useEffect` in the home screen component. However, React Navigation does not unmount components even if you're on a different tab, so this means that this actually happens on every screen in the app, so long as you've loaded the home tab at some point (which you almost certainly have).

This wouldn't matter except that the new messages screen uses min shell mode, so if you unbackground it on the messages screen the tab bar overlaps the text input.

I've changed it to use `useFocusEffect` from React Navigation, which only runs the effect when the screen is active

## Test plan

Check that the behaviour is unchanged on the home tab
Unbackground on the messages screen, and see that there's no tab bar

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -126,7 +126,7 @@ function HomeScreenReady({
           if (
             isMobile &&
             mode.value === 1 &&
-            gate('disable_min_shell_on_foregrounding_v2')
+            gate('disable_min_shell_on_foregrounding_v3')
           ) {
             setMinimalShellMode(false)
           }
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -119,22 +119,24 @@ function HomeScreenReady({
   const gate = useGate()
   const mode = useMinimalShellMode()
   const {isMobile} = useWebMediaQueries()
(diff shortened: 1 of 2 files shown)


src/view/screens/Home.tsx
```diff
@@ -119,22 +119,24 @@ function HomeScreenReady({
...
-        ) {
-          setMinimalShellMode(false)
+  useFocusEffect(
+    React.useCallback(() => {
+      const listener = AppState.addEventListener('change', nextAppState => {
+        if (nextAppState === 'active') {
+          if (
+            isMobile &&
+            mode.value === 1 &&
+            gate('disable_min_shell_on_foregrounding_v2')
```
gaearon, line 129: Mind changing this to `disable_min_shell_on_foregrounding_v3`? That's my screwup but I think we'll want a clean experiment now that you found a flaw. I already cloned it.

gaearon (approved):
lg with rename

src/view/screens/Home.tsx
mozzius: Sure thing

PR #3827, opened 2024-05-02
Improve account switcher pending state

gaearon:
Addresses https://github.com/bluesky-social/social-app/pull/3812#issuecomment-2090771077.

- No longer jump the vertical height when switching accounts
- Fix potential races by ignoring switch requests while switching
- Provide visual feedback while switching (pending account remains selected)
- Ignore _all_ presses within that UI (even "Sign out" or "...") while switching

Still far from perfect but a little bit more sturdy. I kept the visual style mostly as-is.

## Test Plan

Recorded with an extra 1500ms artificial delay inside `initSession`.

https://github.com/bluesky-social/social-app/assets/810438/9c5ac207-412d-47e0-9e99-3f5c10443cdc

### DIFF
--- a/src/components/AccountList.tsx
+++ b/src/components/AccountList.tsx
@@ -16,12 +16,12 @@ export function AccountList({
   onSelectAccount,
   onSelectOther,
   otherLabel,
-  isSwitchingAccounts,
+  pendingDid,
 }: {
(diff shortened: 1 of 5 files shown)


haileyok (approved):
Works great now!

PR #3829, opened 2024-05-02
[Clipclops] External store, suspend/resume

estrattonbailey:
With this PR, `Convo` is hooked up to `useSyncExternalStore`.

- When React subscribes (goes from 0->1 listener) we init the `Convo`, and when React unsubscribes (1->0) the `Convo` enters a `suspended` status
  - `Convo` can resume from `suspended` status if it needs to
- When screen is backgrounded by React Navigation, `Convo` enters a `backgrounded` status, and polling drops to once every 10s (from every 2s by default)
  - When refocused, `Convo` resumes from `backgrounded`
- Snapshotting is lazy, only when React requests it
- Snapshots only change when a new message is received or sent
- Each `Convo` interface returned from `useSyncExternalStore` differs by `status`, but retains the same shape each time
  - This discriminated union might be a little much, but didn't want to totally change it in this PR, can fix-forward if need be
- Methods are bound to the instance once, never again
- Getters were replaced

### DIFF
--- a/src/screens/Messages/Conversation/index.tsx
+++ b/src/screens/Messages/Conversation/index.tsx
@@ -46,16 +46,16 @@ function Inner() {
   const myDid = currentAccount?.did
 
   const otherProfile = React.useMemo(() => {
-    if (chat.state.status !== ConvoStatus.Ready) return
-    return chat.state.convo.members.find(m => m.did !== myDid)
-  }, [chat.state, myDid])
+    if (chat.status !== ConvoStatus.Ready) return
+    return chat.convo.members.find(m => m.did !== myDid)
+  }, [chat, myDid])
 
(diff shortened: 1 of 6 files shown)


src/screens/Messages/Conversation/index.tsx
```diff
@@ -88,12 +87,9 @@ let Header = ({
...
 
-  const onUpdateConvo = useCallback(
-    (convo: ChatBskyConvoDefs.ConvoView) => {
-      service.convo = convo
-    },
-    [service],
-  )
+  const onUpdateConvo = useCallback(() => {
+    // TODO eric update muted state
+  }, [])
```
estrattonbailey, line 92: This was a hack, directly mutating state, will replace

gaearon (approved):

PR #3832, opened 2024-05-03
[Clipclops] Refactor message list

haileyok:
Turns out that `inverted` messeds up the accessibility - the accessibility is _also_ inverted on android and web. So this refactors to a normal `FlatList`

Did some explaining of the quirky things below. There are sure to be some things to tweak over the next week, but this feels like a _decent_ implementation of what we need. At some point, a better implementation - maybe one that fixes some of the issues present in `FlatList` as of right now - would be useful and probably feel better.

### DIFF
--- a/src/screens/Messages/Conversation/MessagesList.tsx
+++ b/src/screens/Messages/Conversation/MessagesList.tsx
@@ -16,7 +16,7 @@ import {useFocusEffect} from '@react-navigation/native'
 import {useChat} from '#/state/messages'
 import {ConvoItem, ConvoStatus} from '#/state/messages/convo'
 import {useSetMinimalShellMode} from '#/state/shell'
-import {isNative, isWeb} from 'platform/detection'
+import {isWeb} from 'platform/detection'
 import {MessageInput} from '#/screens/Messages/Conversation/MessageInput'
 import {useScrollToEndOnFocus} from '#/screens/Messages/Conversation/useScrollToEndOnFocus'
 import {atoms as a} from '#/alf'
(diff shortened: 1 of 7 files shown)


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -72,36 +74,63 @@ function keyExtractor(item: ConvoItem) {
...
 
-  const onEndReached = useCallback(() => {
-    if (chat.status === ConvoStatus.Ready) {
-      chat.fetchMessageHistory()
-    }
-  }, [chat])
+  // contentOpacity is a web-only hack. There is a slight delay between the rendering of the items and the user being
+  // scrolled to the bottom of the view. If we don't hide the items, there's a jarring flash whenever the scroll
+  // happens. Therefore, let's not show the items until that first scroll has actually happened
+  const [hasInitiallyScrolled, setHasInitiallyScrolled] = React.useState(false)
```
haileyok, line 93: Whenever the list was inverted, we didn't need to keep track of this state. We were guaranteed to always start at the "bottom" since...the bottom was really the top. We could also simply rely on `onEndReached` instead of `onStartReached` when adding items.

Now however, we need to keep track of this state for two reasons:

1. We do not want to call `onStartReached` on the initial render. Since the offset will be `50` (remember, `MaybeLoader` will always be rendered), the `FlatList` will fire `onStartReached` immediately. This prevents that from happening until some scroll has happened.

2. Unfortunately, we cannot rely on `initialScrollIndex` to properly place us at the right content offset - even if we let the content load before rendering the `FlatList`. `initialScrollIndex` like `scrollToIndex` is known to be unreliable unless used in combination with `getItemLayout`. In fact, the docs tell us that it cannot be used without `getItemLayout` at all https://reactnative.dev/docs/flatlist#initialscrollindex

Because of this limitation, we are going to need to rely on `scrollToOffset()` to put us at the bottom of the scroll view as soon as the content is loaded. This isn't a perfect solution, but unfortunately seems to be the one we need to rely on. Out of the box, this works fine on native. We simply use `hasInitiallyScrolled` to determine if we want to animate the scroll - we don't if it's the first scroll, then all subsequent ones should be animated.

On web however, we _additionally_ need to use this value to control whether the content is actually visible or not - using `opacity`. The content actually will appear _before_ the scroll is performed, which causes a janky flash. See below for what we do: scroll to the appropriate offset and then make the content visible.

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -72,36 +74,63 @@ function keyExtractor(item: ConvoItem) {
...
-    }
-  }, [])
+  // Used to keep track of the current content height. We'll need this in `onScroll` so we know when to start allowing
+  // onStartReached to fire.
+  const contentHeight = React.useRef(0)
+
+  // This is only used on native because `Keyboard` can't be imported on web. On web, an input focus will immediately
+  // trigger scrolling to the bottom. On native however, we need to wait for the keyboard to present before scrolling,
+  // which is what this hook listens for
+  useScrollToEndOnFocus(flatListRef)
```
haileyok, line 102: On web it's really simple to properly scroll to the bottom on input focus since there's no virtual keyboard. On native however, we don't want to scroll to the bottom until the keyboard has finished presenting. We use this hook to listen for the `keyboardDidShow` event, which will then trigger a scroll to bottom.

Have to put this in a hook because we cannot import `Keyboard` on web. Web is just an empty function.

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -141,39 +178,37 @@ export function MessagesList() {
...
-        onContentSizeChange={onContentSizeChange}
-        onScroll={onScroll}
-        // We don't really need to call this much since there are not any animations that rely on this
-        scrollEventThrottle={100}
+        keyExtractor={keyExtractor}
+        contentContainerStyle={{
+          paddingHorizontal: 10,
+          opacity: hasInitiallyScrolled ? 1 : 0,
+        }}
+        disableVirtualization={true}
```
haileyok, line 187: We should revisit if this prop is needed. It _does_ make the scrolling a little bit smoother if we disable virtualization, but it would be good to profile and see how this affects things.

haileyok:
Note this requires a backend change to work properly. The history is currently coming in the wrong order which is messing up things 😓. Going to talk to Devin tomorrow to get that done. Might also need something more from @estrattonbailey after Devin's change?

estrattonbailey (approved):

PR #3837, opened 2024-05-03
[Clipclops] Moar error

estrattonbailey:
Review without whitespace: https://github.com/bluesky-social/social-app/pull/3837/files?diff=split&w=1

This PR introduces some new `ConvoType`s, `error-fatal` (unused atm) and `error-recoverable`, which has a `retry` method and will probably be used differently from the fatal version (if we even need that). The idea is to insert a given `ConvoError` type into the flow of items.

This PR handles only failed history fetches. But imagine similar errors appearing above or below other messages.

CleanShot 2024-05-02 at 22 05 24@2x

### DIFF
--- a/src/state/messages/convo.ts
+++ b/src/state/messages/convo.ts
@@ -25,6 +25,10 @@ export enum ConvoStatus {
   Suspended = 'suspended',
 }
 
+export enum ConvoError {
+  HistoryFailed = 'historyFailed',
+}
+
 export type ConvoItem =
   | {
       type: 'message' | 'pending-message'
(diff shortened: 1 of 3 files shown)


src/state/messages/convo.ts
```diff
@@ -366,51 +382,72 @@ export class Convo {
...
     /*
-     * Delay if paginating while scrolled to prevent momentum scrolling from
-     * jerking the list around, plus makes it feel a little more human.
+     * If we've rendered a retry state for history fetching, exit. Upon retry,
+     * this will be removed and we'll try again.
      */
-    if (this.pastMessages.size > 0) {
-      await new Promise(y => setTimeout(y, 500))
-    }
+    if (this.headerItems.has(ConvoError.HistoryFailed)) return
```
estrattonbailey, line 389: There should only ever be one of these since I'm using the `ConvoError` as the key, so I'm thinking this should be safe. Alternatively we could add other booleans to the class.

haileyok (approved):
Styling parts LGTM! We'll probably want to change how the bottom error button looks too with this one in mind (right now it's just a button 😅)

PR #3843, opened 2024-05-03
Don't clear toasts when changing users

gaearon:
While working on the session code I've noticed we have a bunch of weird `setTimeout`s around toasts, and that toasts work poorly when switching accounts in general. Turns out, it's because we unmount them.

In this PR, I've relaxed it so that toasts are placed outside the area that gets remounted on user change. As a result, I was able to remove the fragile `setTimeout` hacks. See individual commits.

## Reviewing

- https://github.com/bluesky-social/social-app/commit/dcb14756808c813ef79176a27f8ebf09bb70fb8e?w=1: Puts legacy theme provider next to Alf provider.
- https://github.com/bluesky-social/social-app/commit/38a02b85ec0b15cff18b712cd626c16ce81cff55?w=1: Moves `RootSiblingParent` upwards. This is where toasts get mounted on native. I'll assume we _only_ use it for toasts so it's fine to bleed previous users' UI on account change.
- https://github.com/bluesky-social/social-app/commit/b54e3580b63d49fc71c90f45ebb287248e3980a7?w=1: On web, we use this to render toasts. Make sure it also doesn't get remounted.
- https://github.com/bluesky-social/social-app/commit/1d8ad442f943f3b8202abe9d8bbcb3dc56a9587e?w=1: We can remove the hacks now.
- https://github.com/bluesky-social/social-app/commit/06e20b20fb5bf50ff4fa1d131814456cf808cafe: Our session code does `emitSessionDropped` in the `onAgentSessionChange` handler. It seems like it makes sense to show a toast whenever that happens, regardless of the reason (since it can happen out of the blue too). The native `App.native.tsx` component already does this by `listenSessionDropped` so this copy-pastes the same approach into `App.web.tsx`. I presume it wasn't previously enabled due to the remounting issues.
  - This also lets us remove the toast in `useAccountSwitcher`. It will be taken care of by the `emitSessionDropped` logic. In fact, this toast isn't even being used yet anyway. It was added in https://github.com/bluesky-social/social-app/pull/3762/commits/c020a4c5c5266c678d1055470a10232c1d9e99a1#r1584118170 but not hooked up (cause `initSession` never throws atm), and I think we don't need it now either due to the new strategy.

## Test Plan

Switch accounts on iOS and web. Observe the toast showing up. Try doing it via Settings, via quick switcher, and via the Choose Account screen.

Force session expiry on web using the trick I described in the test plan for https://github.com/bluesky-social/social-app/pull/3838. Observe the expiry toast showing up.

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -91,34 +91,33 @@ function InnerApp() {
       <Alf theme={theme}>
         <ThemeProvider theme={theme}>
           <Splash isReady={isReady}>
-            <React.Fragment
-              // Resets the entire tree below when it changes:
-              key={currentAccount?.did}>
-              <QueryProvider currentDid={currentAccount?.did}>
-                <PushNotificationsListener>
-                  <StatsigProvider>
-                    {/* LabelDefsProvider MUST come before ModerationOptsProvider */}
-                    <LabelDefsProvider>
-                      <ModerationOptsProvider>
-                        <LoggedOutViewProvider>
-                          <SelectedFeedProvider>
-                            <UnreadNotifsProvider>
-                              {/* All components should be within this provider */}
-                              <RootSiblingParent>
+            <RootSiblingParent>
+              <React.Fragment
+                // Resets the entire tree below when it changes:
+                key={currentAccount?.did}>
+                <QueryProvider currentDid={currentAccount?.did}>
+                  <PushNotificationsListener>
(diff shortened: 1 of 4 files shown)


haileyok (approved):

mozzius (approved):
Makes sense, and looks good

pfrazee (commented):
Oh nice

PR #3849, opened 2024-05-03
[Clipclops] Delete message in dialog

haileyok:


src/components/dms/MessageMenu.tsx
```diff
@@ -28,18 +32,26 @@ export let MessageMenu = ({
...
   const deleteControl = usePromptControl()
 
   const isFromSelf = message.sender?.did === currentAccount?.did
 
   const onDelete = React.useCallback(() => {
-    // TODO delete the message
-  }, [])
+    if (chat.status !== ConvoStatus.Ready) return
+
+    chat.deleteMessage(message.id)
```
estrattonbailey, line 43: This can throw FYI. It optimistically removes the message immediately, but if it fails, it's re-inserted immediately too. My thinking is that we want to handle in situ and alert the user so they can retry right from the dialog.

haileyok: so delete message first and then close the dialog?

estrattonbailey: Yeah could just wait for it to delete, show error if fails, then close

estrattonbailey: Could consider other options for sure, but if someone wants to delete a message, we should make sure they know when it fails

haileyok: cool, sg. will think about it today.

mozzius:
Rebased, I did the copy message text in a different PR

mozzius:
Added a retry dialog

mozzius (approved):
"looks good" - Hailey

PR #3852, opened 2024-05-03
[Session] Remove global agent

gaearon:
A twist on #3728.

This finally kills `__globalAgent` and introduces a reducer for managing session state. I've tried a few other ways but as I was simplifying the code, it seemed to naturally gravitate towards this approach so I stopped resisting and extracted it.

## Background

In https://github.com/bluesky-social/social-app/pull/1946 and later https://github.com/bluesky-social/social-app/pull/2026, we've added a mutable global agent reference. This was required because RQ callbacks often access the current agent. Since RQ retains the `queryFn` closure unless you add something to the query key, when you switch an account, it would try to keep using the initial agent from mount — and fail.

The downside of reading a global is it depends on subtle timing. Also, if you switch accounts during an ongoing operation, it might see different agent references in the middle, potentially performing half of the operation on a wrong account. We've also seen things getting displayed incorrectly, e.g. showing wrong pinned feeds.

We've started moving towards a more disciplined approach where the query cache is rooted in the React tree, and the React tree is remounted when the current account changes. This allowed us to remove the fragile "clear the cache" logic in https://github.com/bluesky-social/social-app/pull/3333. This PR, which is extracted from #3728 with some tweaks, aims to complete this transition by moving the source of truth for `agent` into the React state. Here's why we think this will work:

- The state in `SessionProvider` determines the current account _and_ the corresponding agent.
- The entire React tree below "sees" that agent.
- When the current account changes, the entire tree below gets remounted.
  - This includes the React Query cache and any components referencing it.
- The new tree will "see" the new agent.

In situations where you switch account in the middle of an operation, the new approach should also be better. It will keep operating on the old RQ cache in memory (until it gets GC'd) and will keep using the old agent. No matter what happens there, it will remain independent from the newly mounted tree and the new agent.

## Managing state

In this PR, I'm moving state management from a bunch of `setState` calls to a reducer.

There are a few motivations for this:

- When there's a bug, reducers leave breadcrumbs (actions) which help differentiate whether _wrong things have happened_ (bad actions) or _we incorrectly updated state in response to them_ (bad state updates).
- We have a few "kinds" of mutations and it would be nice to describe them in a single place.
- **[Key insight]** The code is currently a mix of "preparing the agent" (async but not side-effectful) and "putting it into state" (obviously affects state). This draws a clearer distinction between these operations. I'll further separate these phases and make this clearer in #3854.
- It lets us easily add tests for the state update parts (will be done in https://github.com/bluesky-social/social-app/pull/3860).

## Consequences

The entire tree will see a consistent `getAgent()` result no matter when it's called. If you switch accounts, this will mount a new tree which will see _its own_ `getAgent()`, but if some code owned by the old tree is still executing, it will keep seeing the old agent. This should fix some latent bugs but can break cases where the code _expects_ the agent to be swapped out midway through the operation. I couldn't find any such code myself.


agentjabsco:
> This finally kills `__globalAgent`

oh no

gaearon:
oh yes

agentjabsco:
I've been looking for an excuse to use this bugout bag anyway

estrattonbailey (approved):

PR #3855, opened 2024-05-04
remove `hide_vertical_scrollbars` test

haileyok:
## Why

Since https://github.com/bluesky-social/social-app/pull/3429 we've been running some tests where we removed the vertical scroll indicators for some users. In the end - after figuring out the proper way to test this case - it seems that there is not a significant difference in usage between having these scrollbars visible and not having them visible.

I think that there would have been a time that I would have said "since there is not a significant difference in usage, maybe we should just keep the change since it appears cleaner". However, two things have changed my mind on this

1. I asked some users how they felt about this and - although a really small sample size - they were mostly in favor of keeping the scroll bars visible.
2. I actually discovered - thanks to [this post](https://bsky.app/profile/puffinomar.bsky.social/post/3krmzelklco25) - that at least on iOS there actually _is_ functionality in the scroll indicator. If you press and hold the indicator, it becomes draggable the same way that you can use it on web.

For both of these reasons, I think we should just leave the scroll bar. Especially whenever it comes to point two, it doesn't seem right to remove something that's actually a feature.

As long as no one else has any objections, lets rm.

## Test Plan

Not really much to test here if you have not been testing without indicators, but if you have, you should see the indicators again on the various screens.

### DIFF
--- a/src/lib/statsig/gates.ts
+++ b/src/lib/statsig/gates.ts
@@ -4,7 +4,6 @@ export type Gate =
   | 'disable_min_shell_on_foregrounding_v3'
   | 'disable_poll_on_discover_v2'
   | 'dms'
-  | 'hide_vertical_scroll_indicators'
   | 'show_follow_back_label_v2'
   | 'start_session_with_following_v2'
   | 'test_gate_1'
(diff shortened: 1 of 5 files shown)


haileyok:
Also, to this point Paul left in the PR

> Yeah I'm +1 to this, especially with the new profile header behavior it becomes super noticable

I've noticed that there's a ton of mainstream apps where this is the case. Given that it's not really an "us" thing and rather just a "this is the inconsistency that comes with this ecosystem" I think that part just "is what it is" :(

gaearon (approved):

gaearon (approved):
alrighty

PR #3857, opened 2024-05-04
[🐴] List Adjustments

haileyok:
- Prevent `scrollToEnd` from being called if a scroll is already in progress
- Animate the view up while keyboard is opening


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -222,15 +248,3 @@ export function MessagesList() {
...
 }
-
-function useKeyboardVerticalOffset() {
-  const {top: topInset} = useSafeAreaInsets()
-
-  return Platform.select({
-    ios: topInset,
-    // I thought this might be the navigation bar height, but not sure
-    // 25 is just trial and error
-    android: 25,
```
haileyok, line 233: @mozzius This `25` actually was giving me too much space 🤔 seems like zero is better?

haileyok:
@mozzius by no means is this necessarily the best way to do this! I was just messing around with it and got it working pretty smoothly but if there's a better way we can totally do that.

src/screens/Messages/Conversation/MessagesList.tsx
mozzius: Hmmm what on earth is that value supposed to be then. `25` is definitely right for my Pixel 3a emulator

mozzius (approved):
Looks good, I'll carry on investigating `keyboardVerticalOffset` next week in a different PR

mozzius:
> @mozzius by no means is this necessarily the best way to do this! I was just messing around with it and got it working pretty smoothly but if there's a better way we can totally do that.

If you got it working then it should be super simple to add interactive dismissal, nice work!

PR #3866, opened 2024-05-05
Fix `IntersectionObserver` `rootMargin` in web `List` implementation, add `onStartReached`

haileyok:
## Why

Right now, we use `IntersectionObserver`  to tell when we have reached the bottom of the page in our web `List` implementation. The currently implementation however has not worked correctly with the `onEndReachedThreshold` parameter. The parameter should fire `onEndReached` whenever we reach X pages away from the bottom of the scrollable content. That has not worked here, and we would have to hit the end to actually load new data.

The problem turned out to be that we were not configuring the `rootMargin` properly. We can think of the value we supply to `rootMargin` as a box surrounding the parent of the `Visibility` element. So, with our current logic of using `100%, 0px, 0px, 0px` this is what it looks like.

Screenshot 2024-05-04 at 9 10 19 PM

We will never end up firing `onEndReached` here until we reach the end!

Instead, we want to be setting the _bottom margin_. If we use `0px, 0px, 100%, 0px` our region looks like this:

Screenshot 2024-05-04 at 9 24 54 PM

What I have done here is modify the `Visibility` component to accept two different props: `topMargin` and `bottomMargin`. Whenever we add a `Visiblity` component to fire `onEndReached`, we will use `bottomMargin`. Then, when we want to use `onStartReached`, we can use `topMargin`.

## Test Plan

On the home feed, we currently load more items whenever we come within 2 pages of the end of the list. So, you should be seeing new content load far before you reach the end now.

We don't actually use `onStartReached` anywhere on web right now, but I have recorded a video of what I'm working on in the chat list to show this behavior.

## onEndReached - Watch the scroll indicator in this video, uses a threshold of 2

https://github.com/bluesky-social/social-app/assets/153161762/266f9245-662c-4106-82db-ef3f2a0ed23c

## onStartReached - No threshold

https://github.com/bluesky-social/social-app/assets/153161762/6625172e-324b-41e7-a1c1-12e13f0614ea

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -213,8 +213,8 @@ function ListImpl<ItemT>(
         ))}
         {onEndReached && (
           <Visibility
-            topMargin={(onEndReachedThreshold ?? 0) * 100 + '%'}
             onVisibleChange={onTailVisibilityChange}
+            bottomMargin={(onEndReachedThreshold ?? 0) * 100 + '%'}
           />
         )}
         {footer}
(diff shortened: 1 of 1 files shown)


gaearon (approved):
ahhh i was wondering why it wasn’t working. makes sense!

PR #3867, opened 2024-05-05
Further align web `List` with `FlatList`, add `contain` mode to web list implementation

haileyok:
## Why

**[Without whitespace](https://github.com/bluesky-social/social-app/pull/3867/files?w=1)**

Right now, our web list implementation only fully supports the the assumption that the list will only be used in a way that monitoring `window` scroll events will suffice. This means that, for example in our chat implementation, we cannot have a `List` that works inside of its own scrollable container within our window. This breaks usages that we could currently achieve with `FlatList`.

Of course, we could just use `FlatList` directly in these cases. This still is not a great solution though, because we lose the improvements that were made in the web list implementation like synchronous layout measurements and scroll events.

I am also adding the following `FlatList` props and functions to support our chat implementation:

- `scrollToEnd()` - This is a method that is available on the native `FlatList` but is missing from our implementation.
- Add `layoutMeasurement` and `contentSize` to the `onScroll` event.

### Container Mode

If we set the list to use container "mode" (using the `containWeb` prop), we add a `overflow-y` prop to the list's styles. The end result is that instead of our window becoming scrollable that the `<List>` itself becomes scrollable.

Screenshot 2024-05-05 at 12 49 04 AM

For the most part, the current implementation already supports everything when switched to this mode. However, there are a few additions we ned to make when it comes to handling scroll callbacks. Because we are now scrolling a specific element rather than the window, we need to attach the `scroll` listener to the div, which we already have a ref for called `nativeRef`.

Elsewhere, we have to fork logic depending on if we are in container mode or not. For example, the `contentSize` object in the `onScroll` callback is achieved this way in container mode:

```ts
          contentSize: {
            width: element?.scrollWidth,
            height: element?.scrollHeight,
          },
```

But when in "regular mode", we have to get this from the document:

```ts
          contentSize: {
            width: document.documentElement.scrollWidth,
            height: document.documentElement.scrollHeight,
          },
```

### `scrollToEnd()`

This is pretty basic. We just scroll to an offset of whatever the scroll height is. Again, depending on if we are using contain or "full" mode, we either use `element?.scrollHeight` or `document.documentElement.scrollHeight`.

## Test Plan

I have added a contained list to the storybook. To access it, scroll to the bottom of the storybook page. Observe the following in the console:

- `onStartReached` should fire immediately, since we start at the top of the list (this is normal `FlatList` behavior)
- `onEndReached` should fire within 2 pages - based on the height of the contained list - of the end. 
- `onScroll` should print throughout the scrolling you perform

Additionally, you can test using this PR: https://github.com/bluesky-social/social-app/pull/3866

Behavior in other lists in the app should remain the same. There have not been any logic changes to the old behavior, but just make sure that is the case.

https://github.com/bluesky-social/social-app/assets/153161762/5831ea70-b007-4f0c-a614-e2dc0bd5933b

Of course, the native behavior also remains the same:

https://github.com/bluesky-social/social-app/assets/153161762/0dca4288-8f50-45a7-a69d-ef1260747d96

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -137,28 +137,7 @@ function ListImpl<ItemT>(
 
   // --- onContentSizeChange, maintainVisibleContentPosition ---
   const containerRef = useRef(null)
-  const contentHeight = useRef(0)
-
-  const onResize = useNonReactiveCallback((w: number, h: number) => {
-    // If `minIndexForVisible` is set, we should maintain the scroll position any time the content size changes.
-    // In a real implementation, this number would be configurable. However, becuase we do not use it this way in our
-    // app I am opting to simply require it to be set to a number greater than or equal to 0. If the time comes that
-    // this is needed, we can make it configurable.
-    if (Number(maintainVisibleContentPosition?.minIndexForVisible) >= 0) {
-      const diff = h - contentHeight.current
-      if (diff > 0) {
-        const element = containWeb
-          ? (nativeRef.current as HTMLDivElement | null)
-          : window
-        element?.scrollBy(0, diff)
-      }
-    }
-
-    contentHeight.current = h
-    onContentSizeChange?.(w, h)
-  })
-
-  useResizeObserver(containerRef, onResize)
+  useResizeObserver(containerRef, onContentSizeChange)
 
   // --- onScroll ---
   const [isInsideVisibleTree, setIsInsideVisibleTree] = React.useState(false)
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -242,7 +242,7 @@ function ListImpl<ItemT>(
     <View
       {...props}
       // @ts-ignore web only
-      style={[style, containWeb && {'overflow-y': 'scroll'}]}
+      style={[style, containWeb && {flex: 1, 'overflow-y': 'scroll'}]}
       ref={nativeRef}>
       <Visibility
         onVisibleChange={setIsInsideVisibleTree}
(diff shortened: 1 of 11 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -303,14 +377,15 @@ let Visibility = ({
 
   React.useEffect(() => {
     const observer = new IntersectionObserver(handleIntersection, {
+      root,
       rootMargin: `${topMargin} 0px ${bottomMargin} 0px`,
     })
```
haileyok, line 382: The default for `IntersectionObserver` is for `root` to be the viewport. This works fine whenever we are not using container mode, but when we are we want our intersections to be determined based on the container's scroll area, not the full window's.

See: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/root

src/view/com/util/List.web.tsx
```diff
@@ -83,13 +86,60 @@ function ListImpl<ItemT>(
...
+        },
+        addEventListener(event: string, handler: any) {
+          window.addEventListener(event, handler)
+        },
+        removeEventListener(event: string, handler: any) {
+          window.removeEventListener(event, handler)
+        },
+      }
+    }
+  }, [containWeb])
```
haileyok, line 134: This is the main thing to pay attention to. It feels safer to create a a single source of truth on what a function does or a value returns rather than have a bunch of `if (containWeb) {} else {}` all over the component.

src/view/com/util/List.web.tsx
```diff
@@ -83,13 +86,62 @@ function ListImpl<ItemT>(
...
```
haileyok, line 136: This is the main thing to pay attention to. It feels safer to create a a single source of truth on what a function does or a value returns rather than have a bunch of if (containWeb) {} else {} all over the component.

estrattonbailey (approved):
LGTM

PR #3868, opened 2024-05-05
[🐴] Finalize web message screen

haileyok:
## Why

Web requires a little extra logic for maintaining scroll position whenever we load new items at the top of the list. This PR adds  a little check whenever the content size changes to see if:

1. We are at the top of the scroll view
2. We have already scrolled at least once (the first scroll takes us to the bottom of the page)

If we meet those criteria, we immediately scroll to the previous offset whenever new items get added to the list.

We also remove an unnecessary wrapper around the `List` that has become unnecessary after changes in the other two PRs.

**[Without whitespace](https://github.com/bluesky-social/social-app/pull/3868/files?w=1)**

## Test Plan

1. Load the messages in a large chat. Verify that you start at the bottom of the conversation.
2. Scroll up to the top. Verify that you see a spinner, and then new messages are added without causing your position to jump.
3. Verify that new messages can be added to the chat while you are scrolled away from the bottom without causing jumps in position.

https://github.com/bluesky-social/social-app/assets/153161762/d5ebb1a8-14d1-4dfe-8d7c-fba08c6964ae


estrattonbailey (approved):
Tested on web and Android, looks great

PR #3873, opened 2024-05-06
Update zh-TW translations

auroursa:
Special thanks to @KuwaLee for making extensive and very meticulous revisions to the translation, as well as adding a few translations for the new features in 1.80.


pfrazee (approved):
Thank you all so much!

Issue #3871, opened 2024-05-05
Appealing labels removes self-labels


Bossett:
**Describe the bug**
When a user self-labels a post, it is then moderated by the official service. If the official service and the user use the same label shortname, and the user appeals it, it effectively removes *both* labels - completely unlabelling the post.

This is a significant issue as the two labels appear to be different in the UI (see #3870) - and operate differently in terms of user expectation.

**To Reproduce**

1. Make a self-labelled post
2. Wait for the AI moderation to label it
3. Appeal the label successfully
4. The post is no longer labelled at all

**Expected behavior**

Appealing a label from the official service should not negate the label already on the content - it should have the effect of reverting to the self-label.

haileyok:
Seems like this part of the problem isn't actually caused by the label being removed, but by the fact that you can't actually see a self label. 

In this case the situation looks like this:

- You post an image and add the self-label nudity
- The post gets upgraded to the porn label immediately. To the user, this Bluesky upgraded label appears to be their self label
- If the user realizes that the label was upgraded, they appeal the upgrade
- Bluesky moderation accepts the appeal, and downgrades it back to nudity
- To the poster, the UI makes it look like all labels were removed from their post, even though they were not.

This should adjust that behavior so more tweaks can be made. https://github.com/bluesky-social/atproto/pull/2467

Bossett:
Ooooh - I misunderstood the pattern, yes that looks like it solves.

Would that then re-use the 'X labels has been applied...' pattern?

haileyok:
Needs a separate fix, but yea that's ideal. Even with this fix you still wouldn't see the label if you had the setting to "Show" so need to add the little indicator.

haileyok:
I linked the PR so it will close when it gets merged.

haileyok:
Part two is here: https://github.com/bluesky-social/social-app/pull/3874

PR #3874, opened 2024-05-06
Show "label has been placed..." even for self-labels

haileyok:
## Why

See: https://github.com/bluesky-social/social-app/issues/3871

> Seems like this part of the problem isn't actually caused by the label being removed, but by the fact that you can't actually see a self label.
> 
> In this case the situation looks like this:
> 
> You post an image and add the self-label nudity
> The post gets upgraded to the porn label immediately. To the user, this Bluesky upgraded label appears to be their self label
> If the user realizes that the label was upgraded, they appeal the upgrade
> Bluesky moderation accepts the appeal, and downgrades it back to nudity
> To the poster, the UI makes it look like all labels were removed from their post, even though they were not.

This PR displays the `label has been placed on this content...` notice and dialog even in cases where the label is a self-label. This can reduce some significant confusion that's going on with labels right now.

There was also an unnecessary `export` of the `DialogInner` here. Removed that in favor of only exporting the `Dialog`.

## Test Plan

This cannot be properly tested until API changes in https://github.com/bluesky-social/atproto/pull/2467. You can also apply a patch instead:

```
diff --git a/packages/api/src/moderation/decision.ts b/packages/api/src/moderation/decision.ts
--- a/packages/api/src/moderation/decision.ts	(revision 58f719cc1c8d0ebd5ad7cf11221372b671cd7857)
+++ b/packages/api/src/moderation/decision.ts	(revision 295d513b54c53a7e8a66f138e49f2df9b8e9f88b)
@@ -78,9 +78,27 @@
 
   ui(context: keyof ModerationBehavior): ModerationUI {
     const ui = new ModerationUI()
+    // Always apply labels on posts even if the user is the author. Also, we
+    // want to downgrade the `hide` setting to `warn` in these cases so the
+    // user is always aware of the label.
     if (this.isMe) {
+      if (context !== 'contentView' && context !== 'contentMedia') return ui
+      for (const cause of this.causes) {
+        if (cause.type !== 'label' || cause.target !== 'content') continue
+        if (!cause.downgraded) {
+          cause.setting = cause.setting === 'hide' ? 'warn' : cause.setting
+          if (cause.behavior[context] === 'blur') {
+            ui.blurs.push(cause)
+          } else if (cause.behavior[context] === 'alert') {
+            ui.alerts.push(cause)
+          } else if (cause.behavior[context] === 'inform') {
+            ui.informs.push(cause)
+          }
+        }
+      }
       return ui
     }
+
     for (const cause of this.causes) {
       if (
         cause.type === 'blocking' ||
diff --git a/packages/api/tests/moderation-behaviors.test.ts b/packages/api/tests/moderation-behaviors.test.ts
--- a/packages/api/tests/moderation-behaviors.test.ts	(revision 58f719cc1c8d0ebd5ad7cf11221372b671cd7857)
+++ b/packages/api/tests/moderation-behaviors.test.ts	(revision 295d513b54c53a7e8a66f138e49f2df9b8e9f88b)
@@ -551,7 +551,9 @@
     subject: 'post',
     author: 'self',
     labels: { post: ['!hide'] },
-    behaviors: {},
+    behaviors: {
+      contentView: ['blur'],
+    },
   },
   "Self-post: Imperative label ('!hide') on author profile": {
     cfg: 'none',
@@ -573,7 +575,9 @@
     subject: 'post',
     author: 'self',
     labels: { post: ['!warn'] },
-    behaviors: {},
+    behaviors: {
+      contentView: ['blur'],
+    },
   },
   "Self-post: Imperative label ('!warn') on author profile": {
     cfg: 'none',
```

Create a post and add a self-label to the post. You should see a notice about this label being added now.

https://github.com/bluesky-social/social-app/assets/153161762/c6d6db9f-5fec-4ca7-a3b8-c5f5aadc617e


pfrazee (approved):
This is good to go but the API bump means it depends on https://github.com/bluesky-social/social-app/pull/3593 now

PR #3876, opened 2024-05-06
Alt text for gifs

mozzius:
Adds a way to add alt text to GIFs, via the `description` field on the embed.

Simulator Screenshot - iPhone 15 - 2024-05-06 at 15 46 33
Simulator Screenshot - iPhone 15 - 2024-05-06 at 15 46 10
Simulator Screenshot - iPhone 15 - 2024-05-06 at 15 46 16


haileyok:
Back button does work after closing the dialog. There's an unrelated nit that should be resolved when pressing the back button, but outside this scope. (If you have the gif picker open and press the back button, it doesn't break anything anymore, but it might close the composer as well if there is no text in the composer, should be an easy fix)

PR #3877, opened 2024-05-06
A few `containWeb` List nits

haileyok:
- Pass the ref to `Visibility`
- Use getters for `contentHeight` et al.

cc @gaearon 

https://github.com/bluesky-social/social-app/assets/153161762/7009ce0c-58b4-4dda-8e45-692d082edc74

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -278,14 +278,18 @@ function ListImpl<ItemT>(
     },
   )
 
-  // Ignoring the error here because `nativeRef` will indeed be a `HTMLDivElement`. However, becuase we are using
-  // the RN `View`, it is not typed properly.
   return (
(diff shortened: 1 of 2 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -254,10 +278,12 @@ function ListImpl<ItemT>(
     },
   )
 
+  // Ignoring the error here because `nativeRef` will indeed be a `HTMLDivElement`. However, becuase we are using
+  // the RN `View`, it is not typed properly.
   return (
+    // @ts-expect-error
```
gaearon, line 284: Can we instead do something like `nativeRef as any as ...` so the rest gets checked?

gaearon (approved):
lg with nit

gaearon (approved):

PR #3881, opened 2024-05-06
Remove `getProfile` calls when loading feed

haileyok:
## Why

Currently, we make a ton of calls to `getProfile()` to display the "Reply to" label. However, since https://github.com/bluesky-social/atproto/pull/2461, we can now get this info from the `grandparentAuthor` field in the API response.

### Before

Screenshot 2024-05-06 at 4 20 52 PM

### After

Screenshot 2024-05-06 at 4 21 12 PM

Verify that the requests are no longer present when loading your timeline. Also verify that posts properly display the `Reply to` label when they should.

Check situations where:

- It appears normally: i.e. a post with one parent above. It should show the parent's parent's author.
- We remove the parent from the timeline. It should show the parent's author.
- Where we put a thread together, i.e. below

### A normal situation

Screenshot 2024-05-06 at 5 19 31 PM

### Removing the parent

Screenshot 2024-05-06 at 5 18 51 PM

### Us making a thread 

Screenshot 2024-05-06 at 5 16 31 PM

### DIFF
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -73,7 +73,7 @@ export interface FeedPostSliceItem {
   reason?: AppBskyFeedDefs.ReasonRepost | ReasonFeedSource
   feedContext: string | undefined
   moderation: ModerationDecision
-  grandparentAuthor?: AppBskyActorDefs.ProfileViewBasic
+  replyToAuthor?: AppBskyActorDefs.ProfileViewBasic
 }
 
 export interface FeedPostSlice {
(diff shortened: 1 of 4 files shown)


src/state/queries/preferences/const.ts
```diff
@@ -34,6 +34,7 @@ export const DEFAULT_LOGGED_OUT_PREFERENCES: UsePreferencesQueryResponse = {
     pinned: [],
     unpinned: [],
   },
+  savedFeeds: [],
```
haileyok, line 37: Not sure if this is the correct move here

package.json
```diff
@@ -51,7 +51,7 @@
   },
   "dependencies": {
     "@atproto-labs/api": "^0.12.8-clipclops.0",
-    "@atproto/api": "^0.12.5",
```
estrattonbailey, line 54: We should not merge this update unless we also plan to get movable following feed in in the same release. Once we merge this, anyone on the TF will also start diverging from v1->v2 feeds prefs.

If we need to get this in though, there are options, just need to discuss.

estrattonbailey (changes requested):
Just blocking for now on the SDK update, looks great otherwise

package.json
haileyok: Perfect, we were wondering what that was and were going to hold off anyway so good to know. Thank you!

gaearon (approved):

PR #3883, opened 2024-05-07
Align fetch polyfill with web behavior

gaearon:
This is the logic we use in the polyfill for native:

https://github.com/bluesky-social/social-app/blob/2ca4b74955244f23aa40ac2d98267e301a59793c/src/lib/api/api-polyfill.ts#L64-L74

And this is the logic on the web:

https://github.com/bluesky-social/atproto/blob/5226cbcc16e01d7076bd04e2cf3dc40d5ca9d2d3/packages/xrpc/src/util.ts#L157-L183

They were not aligned.

On the web, if we receive an OK response with no body at all, ` && data?.byteLength` is falsy so we fall through to [this branch](https://github.com/bluesky-social/atproto/blob/5226cbcc16e01d7076bd04e2cf3dc40d5ca9d2d3/packages/xrpc/src/util.ts#L180-L182). On native, we didn't do that, which led to an unhandled rejection with FeedFeedback API (which returns zero bytes).

## Test Plan

Browsed around both on iOS and Android. Seems fine.

### DIFF
--- a/src/lib/api/api-polyfill.ts
+++ b/src/lib/api/api-polyfill.ts
@@ -1,5 +1,5 @@
-import {BskyAgent, stringifyLex, jsonToLex} from '@atproto/api'
 import RNFS from 'react-native-fs'
+import {BskyAgent, jsonStringToLex, stringifyLex} from '@atproto/api'
 
 const GET_TIMEOUT = 15e3 // 15s
 const POST_TIMEOUT = 60e3 // 60s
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Okay cool. Might impact performance slightly but it's more correct so it lgtm

gaearon:
Not gonna risk regressing perf, in RN `.arrayBuffer()` seems to do a bunch of extra work compared to `.text()` because they both come from a blob, but `.arrayBuffer()` turns that blob into `base64` first on the native side and then it has to be turned back in JS.

PR #3884, opened 2024-05-07
Only close modal if directly clicking on backdrop

### LINKED ISSUE #1894: Ending the drag text select outside profile/list edit boxes closes it.
**Describe the bug**
Ending the drag text select outside profile/list edit boxes closes it. It also applies to few more popup boxes that let you edit text, probably all of them. Another example is the edit handle box.
I'm used to not care about ending the drag select precisely inside the box so this is really annoying.

**To Reproduce**

1. Edit your profile or a list
2. Drag select text and end it with cursor outside the editing box
3. The box closes

**Expected behavior**

The box does not close

**Details**

 - Platform: Chrome Windows

### LINKED ISSUE #3834: Dialogs close `onmouseup` when dragging cursor outside dialog bounds
**Describe the bug**

Dialogs like "Edit my profile" currently close if the user clicks within the dialog, drags their cursor outside the dialog, and then releases the click. This can commonly happen when attempting to select text within an input.

**To Reproduce**

1. Click and hold down within a dialog.
2. Drag the cursor outside of the dialog.
3. Release the click.
4. The dialog closes.

**Expected behavior**

The dialog remains open when an `onmouseup` event triggers outside of a dialog.

**Screenshots**

https://github.com/bluesky-social/social-app/assets/4117920/099b42e9-3265-42d8-b7bf-ed565903b80a

**Details**

- Build version: 1.80.0
- Platform: web
- Operating System: macOS
- Browser: Chrome v124

**Additional context**

⧽(•‿•)⧼ thank you

mary-ext:
Fixes #1894
Fixes #3834

This seemed to work for checking where the click/press initially started from, we don't want any clicks that initially started from the modal to be handled by the backdrop (which closes the modal)

### DIFF
--- a/src/view/com/modals/Modal.web.tsx
+++ b/src/view/com/modals/Modal.web.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import {StyleSheet, View} from 'react-native'
 import Animated, {FadeIn, FadeOut} from 'react-native-reanimated'
 
+import {useOutsideClick} from '#/lib/hooks/useOutsideClick'
 import {useWebBodyScrollLock} from '#/lib/hooks/useWebBodyScrollLock'
 import type {Modal as ModalIface} from '#/state/modals'
 import {useModalControls, useModals} from '#/state/modals'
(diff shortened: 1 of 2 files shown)


mary-ext:
Only deals with the old modals at the moment, which is fine, since the issue is with the profile edit modal. Let me see if there's a good way of doing this with the new modals, I think the right path might just be to expose a private API over the modal context

src/view/com/modals/Modal.web.tsx
```diff
@@ -51,23 +51,36 @@ function Modal({modal}: {modal: ModalIface}) {
...
+
   if (!isModalActive) {
     return null
   }
 
-  const onPressMask = () => {
+  const onPressMask = (ev: React.MouseEvent<HTMLDivElement>) => {
+    // 1. Ignore clicks from inside the modal
+    // 2. Ignore clicks that started inside the modal and ended in the mask
+    if (ev.target !== ev.currentTarget.firstChild || clickInsideRef.current) {
```
gaearon, line 63: Any solution that relies on timing here (like with tracking pointer down/up) seems suspicious to me. For this type of problem it's usually sufficient to check whether one node `contains` the other node.

What is special that requires this?

mary-ext: Yeah this doesn't seem the right approach I think, now that I've tested it again this actually swallows clicks to backdrop if you do the second case.

mary-ext: Hm, it seems that Headless UI does more or less the same thing to check where the click initially started.

https://github.com/tailwindlabs/headlessui/blob/2d5d35a5337154e823767291637634b69b09ab94/packages/%40headlessui-react/src/hooks/use-outside-click.ts#L106-L149

The problem this PR tries to fix can be illustrated here:

https://github.com/bluesky-social/social-app/assets/148872143/e5b292eb-71ff-4d5d-8740-6496f061e89d

It's very easy to overshoot the mouse cursor when selecting text, and while a confirmation modal would've easily done the same, it doesn't really remove the annoyance issue that overshooting causes

PR #3885, opened 2024-05-07
fix bad bool check in action

haileyok:
- Don't cancel still-running bundle deploys if we merge another PR. Feels better to just let them run (this is for internal use only)
- Replace a null check with what is supposed to just be an equality check
- Always fetch some commits so that we can get the commit hash
- Add `.env` to `.easignore`

Regarding the last point, if we don't add this to the `.easignore`, it won't pick up our environment variables. See `eas build:inspect --stage archive -p ios --output sampleBuildFiles`

### DIFF
--- a/.easignore
+++ b/.easignore
@@ -95,7 +95,6 @@ web-build/
 /ios/
 
 # environment variables
-.env
 .env.*
 
 # Firebase (Android) Google services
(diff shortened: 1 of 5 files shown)


gaearon:
>Add .env to .easignore

You mean remove it from ignore?

gaearon (approved):

haileyok:
oops yea. remove from easignore, _add_ to the build process.

PR #3890, opened 2024-05-07
Rewrite composer image state management, remove MobX

mary-ext:
Decided to do this separately from #3813, removes the only remaining MobX dependency from the app.

- Temporary image files are moved to `{cacheDirectory}/bsky-composer`, this folder is set to purge when the composer is closed. I feel this is a worthwhile approach compared to making sure that images are being purged properly every time it's being manipulated.
- As a side benefit, the new image models are immutable.
- ALF'd image alt text and edit dialog

| Alt text dialog | Edit image dialog |
| --- | --- |
|  |  |

- Working with image cropping libraries is just as wonderful as I remembered it to be, `react-avatar-editor` has been replaced with `react-image-crop` which is slightly more workable. Would've preferred one that also does fixed cropping but `react-easy-crop` is being finicky with initial states.
- The image cropper dialog for avatars and banners still exists, but it's been cleaned up slightly


mary-ext:
> Temporary image files are moved to {cacheDirectory}/bsky-composer

@haileyok just wondering if this is ok since you've dealt with making sure temp files from the composer are being purged properly, I do have concerns about calling `purgeTemporaryImageFiles` from within `useEffect` though which I might just have it do *after* the post has been sent

haileyok:
It seems fine to do the cleanup in the single pass that way. Maybe ideally they would get removed whenever the composer closes since the post should have been submitted by then and the files shouldn't be needed again? I haven't touched this dialog much, but iirc there is a way to tell the dialog what to do `onClose`.

mary-ext:
Ended up touching a lot more than expected (didn't realize that we had two cropper dialogs), but it should be good to go now.

src/view/com/composer/photos/EditImageDialog.web.tsx
```diff
@@ -0,0 +1,191 @@
...
+              <Path
+                d="M4 4H20V20H4V4Z"
+                stroke={
+                  aspect === 1 / 1 ? t.palette.primary_600 : t.palette.white
+                }
+                strokeWidth="2"
+                strokeLinejoin="round"
+              />
+            </Svg>
+          </ToolbarButton>
```
mary-ext, line 152: not sure how the icons are being exported here so i just plopped it off central's react native template

mary-ext, line 152: not sure how the svg icons are being exported so i just copied central's react native template code

PR #3895, opened 2024-05-07
Handle push notifications for DMs

haileyok:
### Edit

For now, let's merge this in with only the DM notification logic, since that is the primary thing we need right now. We should investigate other platforms nad determine how twe want to handle each individual case here. I.e. where do we want to push for a certain type of notification.

----

This needed a good bit of thinking to get right, but I think it is mostly there. We're tackling a few birds with one stone here!

1. Pressing a mention or reply notification takes you to that post rather than the notifications tab
2. Pressing a follow notification takes you to that profile
3. DM notifications will (currently) alway show up in the app. Unless...
4. Notifications will not be displayed for the current conversation you are viewing
5. When the app is foregrounded, you will not see notifications for messages on accounts _other_ than the one you're currently actively signed in to.
6. When in the background, you _will_ see notifications for all accounts. If you tap on a notification for an account that is not your currently signed-in account, you'll be automatically switched to the other account and then pushed to the conversation.

### Note about points 3 and 4

We need to consider one thing when it comes to points one and two here. When you open the app through one of these notifications, we won't actually "remove" these items from your "unread" notifications. This is a tricky thing to do, because we are currently making notifications as read with a "last seen" time. If we updated that time when you pressed a notification for a mention, we'd be wiping all your unreads up to that point - not just the one you're tapping.

With that said, we can comment out those handlers for now. It sort of comes down to "which is the better experience" I think right now. We'll talk about how to improve this as well, but it was a super easy thing to add in here where I needed to rework the handler anyway.

## Test Plan

Testing this is not fun. There isn't any way to receive notifications on the simulator, and even if you build for a device, you still cannot receive the notifications as the push notification token is a development token rather than a production one. With that said, you can send "fake" notifications.

1. Figure out the device name - `xcrun simctl list devices | grep Booted`
2. Save a payload to `testnotif.json`. An example payload would be
```json
{
    "aps": {
        "alert": {
            "title": "Mention Notification",
            "body": "A test Mention notification"
        }
    },
    "reason": "mention",
    "uri": "at://did:plc:2p25gewthftad2wefe7yyzhr/app.bsky.feed.post/3krxmsj45iw2x",
    "subject": ""
}
```
3. Send the notification - `xcrun simctl push "iPhone 15 Pro" xyz.blueskyweb.app testMention.json`

### Pressing a mention or reply notification takes you to the post

RocketSim_Recording_iPhone_15_Pro_6 1_2024-05-08_12 19 14

### Those types of notifications still do not show up when the app is foregrounded

https://github.com/bluesky-social/social-app/assets/153161762/f2cbf485-be9f-46bf-a1c5-b9fe48f54a9d

### Pressing a follow notification takes you to the profile

RocketSim_Recording_iPhone_15_Pro_6 1_2024-05-08_12 45 55

### Pressing a message notification takes you to the convo

RocketSim_Recording_iPhone_15_Pro_6 1_2024-05-08_12 47 48

### Pressing a notification for an account that isn't the current one switches your account

RocketSim_Recording_iPhone_15_Pro_6 1_2024-05-08_12 49 00


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -0,0 +1,207 @@
...
+            shouldPlaySound: false,
+            shouldSetBadge: false,
+          }
+        }
+
+        // Any notification other than a chat message should invalidate the unread page
+        invalidateCachedUnreadPage()
+        return DEFAULT_HANDLER_OPTIONS
+      },
+    })
```
haileyok, line 149: This replaces the old `addNotificationReceivedListener`, because it lets us also control visibility of the push notification.

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -0,0 +1,207 @@
...
+        }
+      })
+
+    if (
+      storedPayload?.reason === 'chat-message' &&
+      currentAccount?.did === storedPayload.recipientDid
+    ) {
+      handleNotification(storedPayload)
+      storedPayload = undefined
+    }
```
haileyok, line 191: When we encounter a case where you open a notification that doesn't belong to the currently signed in account, we switch the account above and store the payload. Once `currentAccount` changes, we will re-handle the notification.

This only applies to `chat-message` right now because we don't have a `recipientDid` yet on other notification payloads.

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -0,0 +1,207 @@
...
+        } else {
+          setTimeout(() => {
+            // @ts-expect-error types are weird here
+            navigation.navigate('MessagesTab', {
+              screen: 'MessagesConversation',
+              params: {
+                conversation: payload.convoId,
+              },
+            })
+          }, 500)
```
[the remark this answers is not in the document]
haileyok: Another thing that I'd like to look at here is handling adding the `Messages` screen onto this stack if it doesn't exist already. The react nav documentation is a bit unclear on how to do this, so maybe I can find an example elsewhere in the codebase.

src/screens/Messages/Conversation/index.tsx
```diff
@@ -109,7 +109,10 @@ let Header = ({
     if (isWeb) {
       navigation.replace('Messages')
     } else {
-      navigation.pop()
+      navigation.reset({
+        index: 0,
+        routes: [{name: 'Messages'}],
+      })
```
haileyok, line 115: For now, in cases where we got to this screen from a notification, adding the `Messages` route to the start of this creates the "pop" animation that we are looking for.

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -0,0 +1,207 @@
...
+      reason: Exclude<NotificationReason, 'chat-message'>
+      uri: string
+      subject: string
+    }
+  | {
+      reason: 'chat-message'
+      convoId: string
+      messageId: string
+      recipientDid: string
+    }
```
estrattonbailey, line 41: sickos-yes.png

src/lib/hooks/useNotificationHandler.ts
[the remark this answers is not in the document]
estrattonbailey: Yeah having Messages as the immediate screen when pressing "back" would be awesome

src/lib/hooks/useNotificationHandler.ts
estrattonbailey: Nice one

estrattonbailey (commented):
Code LGTM, pending discussion of behaviors

PR #3896, opened 2024-05-07
Fix color of new web splash logo

mozzius:
Seemed way too dark to me so I copy/pasted the hex code from the header svg


gaearon (approved):
nice catch

PR #3897, opened 2024-05-07
[🐴] Global event mgmt

estrattonbailey:
- [x] Trail logs for single convo


gaearon (approved):

estrattonbailey:
Merging for now, will continue to iterate, but I think the general contract here is solid

PR #3901, opened 2024-05-07
[🐴] Unread messages badge

mozzius:
Message badge + hooks for mutating the query cache. Hooks not yet in use except for optimistically updating unread state when navigating to the message page

Screenshot 2024-05-08 at 00 15 34


src/state/messages/index.tsx
```diff
@@ -37,15 +38,18 @@ export function ChatProvider({
...
+  const {mutate: markAsRead} = useMarkAsReadMutation()
 
   useFocusEffect(
     React.useCallback(() => {
       convo.resume()
+      markAsRead({convoId})
 
       return () => {
         convo.background()
+        markAsRead({convoId})
```
estrattonbailey, line 50: These handlers in this file work like this:
- when a chat is focused, it's marked as read because it's active
- upon blurring a chat, it's marked as read because we want to mark it read up to the last point it was active
- if the app is foregrounded and the screen is focused, it's marked as read because it's active
- if the app is backgrounded (or tab is unfocused) and the screen was focused, it's marked as read because we want to mark it read up to the last point it was active

src/state/queries/messages/list-converations.ts
```diff
@@ -27,3 +33,100 @@ export function useListConvos({refetchInterval}: {refetchInterval: number}) {
...
+export function useUnreadMessageCount() {
+  const {currentConvoId} = useCurrentConvoId()
+  const convos = useListConvos({
+    refetchInterval: 30_000,
+  })
+
+  const count =
+    convos.data?.pages
+      .flatMap(page => page.convos)
+      .filter(convo => convo.id !== currentConvoId)
```
estrattonbailey, line 46: This is optimistic, it immediate decrements the badge count when `currentConvoId` is set by the active chat screen.

src/state/queries/messages/list-converations.ts
```diff
@@ -27,3 +33,100 @@ export function useListConvos({refetchInterval}: {refetchInterval: number}) {
...
+    }
+  }, [count])
+}
+
+type ConvoListQueryData = {
+  pageParams: Array<string | undefined>
+  pages: Array<ChatBskyConvoListConvos.OutputSchema>
+}
+
+export function useOnDeleteMessage() {
```
estrattonbailey, line 64: These methods below are unused atm

estrattonbailey (approved):

PR #3904, opened 2024-05-07
[🐴] Integrate global event bus

estrattonbailey:
Integrates global clipclop log event bus + introduces a small bit of context at a top level for current convo id, so that we can suppress unread state updates and push notifications.

Double checked the conditional rendering of the event bus context, should not affect uses without clipclops enabled, or without the service URL.

### DIFF
--- a/src/state/messages/events/index.tsx
+++ b/src/state/messages/events/index.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import {AppState} from 'react-native'
 import {BskyAgent} from '@atproto-labs/api'
 
+import {useGate} from '#/lib/statsig/statsig'
 import {isWeb} from '#/platform/detection'
 import {MessagesEventBus} from '#/state/messages/events/agent'
 import {MessagesEventBusState} from '#/state/messages/events/types'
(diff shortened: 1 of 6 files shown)


src/state/messages/index.tsx
```diff
@@ -66,3 +68,11 @@ export function ChatProvider({
...
   return <ChatContext.Provider value={service}>{children}</ChatContext.Provider>
 }
+
+export function MessagesProvider({children}: {children: React.ReactNode}) {
+  return (
+    <CurrentConvoIdProvider>
+      <MessagesEventBusProvider>{children}</MessagesEventBusProvider>
+    </CurrentConvoIdProvider>
+  )
+}
```
estrattonbailey, line 78: Will clean up other stuff in this file in a later PR

src/state/messages/events/index.tsx
```diff
@@ -65,3 +66,18 @@ export function MessagesEventBusProvider({
...
+}) {
+  const gate = useGate()
+  const {serviceUrl} = useDmServiceUrlStorage()
+  if (gate('dms') && serviceUrl) {
+    return (
+      <Temp_MessagesEventBusProvider>{children}</Temp_MessagesEventBusProvider>
+    )
+  }
+  return children
+}
```
estrattonbailey, line 83: Temp conditional provider, only if DMs and the service URL are available to the user.

haileyok (approved):
Yea this looks like what we need. Thanks!

PR #3905, opened 2024-05-08
Localize datetime and numeric format

quiple:
Localizes datetime and numeric format strings.

`ago()` in `<TimeElapsed>` doesn't switch dynamically when switching languages within the app.


tkusano:
I've only tried the web version locally and it looks good.

quiple:
Replace with #4268 and #4273. Thanks!

PR #3912, opened 2024-05-08
Improve localization of {els} string on Create Account page

### LINKED ISSUE #3660: {labelTarget}, {els} won't work cross-linguistically
There are a couple of localizable strings that use the variable {labelTarget}.  Browsing the code, it looks like this can only take the values which are the localized versions of "account" or "content".  This won't work for Irish since the localizations of those words would need to change based on the wider context.

The general i18n principle is to avoid "sentence puzzles" like this where variables get plugged in from other localized strings, since they are sure to fail for some language.  Better to just offer separate strings for localization:

"label has been placed on this account"
"label has been placed on this content"

(NB this is in contrast with the need for *more* variables in the case of unlocalizable content like usernames, hashtags, URLs, etc.; cf Issue #3481)

Same issue with {els} in the string: "By creating an account you agree to the {els}."
Especially so given that (if I understand correctly), the value plugged into {els} could be either singular (a Privacy Policy) or plural (the Terms of Service).

quiple:
Makes it easier to localize by providing full context.

스크린샷 2024-09-18 09 51 32

```po
#: src/screens/Signup/StepInfo/Policies.tsx:74
#~ msgid "By creating an account you agree to the {els}."
#~ msgstr ""

#: src/screens/Signup/StepInfo/Policies.tsx:70
msgid "By creating an account you agree to the <0>Privacy Policy</0>."
msgstr ""

#: src/screens/Signup/StepInfo/Policies.tsx:46
msgid "By creating an account you agree to the <0>Terms of Service</0> and <1>Privacy Policy</1>."
msgstr ""

#: src/screens/Signup/StepInfo/Policies.tsx:60
msgid "By creating an account you agree to the <0>Terms of Service</0>."
msgstr ""
```
Resolves #3660.

Also, change the font color of "Having trouble?" to match the font color of the rest of the page.


haileyok (approved):
Confirmed this works on iOS and web. Thank you for doing this, and sorry we let it sit for so long 🙏

PR #3913, opened 2024-05-08
[🐴] Mark as read in convo menu

mozzius:


src/components/dms/ConvoMenu.tsx
```diff
@@ -107,6 +112,24 @@ let ConvoMenu = ({
...
           <Menu.Group>
+            {showMarkAsRead &&
+              ChatBskyConvoDefs.isMessageView(convo.lastMessage) && (
+                <Menu.Item
+                  label={_(msg`Mark as read`)}
+                  onPress={() => {
+                    if (ChatBskyConvoDefs.isMessageView(convo.lastMessage)) {
+                      markAsRead({
+                        convoId: convo.id,
+                        messageId: convo.lastMessage.id,
```
estrattonbailey, line 123: If you leave this off it uses the latest message by default, which is pry the right move in case the client is a split second behind the server

estrattonbailey (approved):

src/components/dms/ConvoMenu.tsx
mozzius: ah cool didn't realise

PR #3914, opened 2024-05-08
Move to requestPollInterval handling

estrattonbailey:
Different screens and views will require different poll intervals. Instead of implicitly setting and reverting poll interval directly for each case, this PR introduces `requestPollInterval`. This method maintains a reference to all requested intervals, and if the event bus is in the `Ready` state, we apply the lowest poll interval requested.

Components that request a poll interval will need to ensure they withdraw the request when unmounting, to `requestPollInterval` returns a function to do so.

```typescript
const events = useMessagesEventBus()

events.trailConvo(convoId, events => console.log(events))

useFocusEffect(
  useCallback(() => {
    const cleanup = events.requestPollInterval(1e3)
    return () => { cleanup() }
  }, [events])
)
```

### DIFF
--- a/src/state/messages/events/agent.ts
+++ b/src/state/messages/events/agent.ts
@@ -15,8 +15,7 @@ import {
 
 const LOGGER_CONTEXT = 'MessagesEventBus'
 
-const ACTIVE_POLL_INTERVAL = 60e3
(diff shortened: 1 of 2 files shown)


mozzius (approved):
beautiful

PR #3921, opened 2024-05-08
[🐴] Skeleton header

mozzius:
- skeleton header
- better error screen


haileyok (approved):
This feels good, we just need to adjust that little jump. Want to test on device though because maybe it's just the simulator/my shitty monitor.

PR #3923, opened 2024-05-09
[🐴] Add hover context menu for convo list on web

haileyok:


mozzius (approved):
looks great, works great

PR #3925, opened 2024-05-09
Rewrite composer image state management, remove MobX

pfrazee:
Supersedes https://github.com/bluesky-social/social-app/pull/3890 by @mary-ext.

---

*Original PR description:*

Decided to do this separately from #3813, removes the only remaining MobX dependency from the app.

- Temporary image files are moved to `{cacheDirectory}/bsky-composer`, this folder is set to purge when the composer is closed. I feel this is a worthwhile approach compared to making sure that images are being purged properly every time it's being manipulated.
- As a side benefit, the new image models are immutable.
- ALF'd image alt text and edit dialog

| Alt text dialog | Edit image dialog |
| --- | --- |
|  |  |

- Working with image cropping libraries is just as wonderful as I remembered it to be, `react-avatar-editor` has been replaced with `react-image-crop` which is slightly more workable. Would've preferred one that also does fixed cropping but `react-easy-crop` is being finicky with initial states.
- The image cropper dialog for avatars and banners still exists, but it's been cleaned up slightly

---

*Additions*

- Removed aspect ratio controls from the composer image editor. Not really needed now that there's a cropping box, and it wasn't working correctly.
- Fixes to the cropping modal used by UserAvatar and UserBanner


mary-ext:
The new modals might make it easy to accidentally lose your work if you manage to overshoot your mouse cursor while moving/resizing the crop area around.

I'll have to update https://github.com/bluesky-social/social-app/pull/3884 to get it to work for the new modals as well if we still want to close on backdrop but not on the case I just mentioned

PR #3926, opened 2024-05-09
[🐴] Rich text in messages

mozzius:
Adds support for links, mentions, tags. Not yet displayed while typing, it generates them just as you send and then the bubble can now display them

Screenshot 2024-05-09 at 13 42 07

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import {TextStyle} from 'react-native'
 import {AppBskyRichtextFacet, RichText as RichTextAPI} from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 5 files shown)


src/components/RichText.tsx
```diff
@@ -35,13 +36,22 @@ export function RichText({
...
-  const styles = [a.leading_snug, flatten(style)]
+
+  const flattenedStyle = flatten(style)
+  const plainStyles = [a.leading_snug, flattenedStyle]
+  const interactiveStyles = [
+    a.leading_snug,
+    a.pointer_events_auto,
+    display === 'underlines' && a.underline,
+    flattenedStyle,
+  ]
```
estrattonbailey, line 54: Wait why is this necessary here 🤔

mozzius: It was an attempt to simplify the code here, there was a lot of repetition of `[...styles, {pointerEvents: 'auto'}]` and I wanted to add `a.underline` to it

mozzius: I can't just pass in underline to the style prop because that would affect all the text

estrattonbailey: Oh of course, I got ya, my b. Kind of a nit, but we have `style`, can we just do `interactivesStyle` or something and pass in styles as an array that get spread here in place of the `display` check? Feels more stable than these other values.

mozzius: I like that better, will do

PR #3929, opened 2024-05-09
Don't eagerly load all JS assets

mary-ext:
Realized that Bluesky is currently eagerly loading all languages and some code that was meant to be lazily-loaded, then noticed that Expo? Webpack? generates an `asset-manifest.json` which contains the actual entrypoint scripts to put on the HTML file.

Anyway, that saved time from having to figure out plugins like `webpack-assets-manifest` which is what I was about to pull beforehand.

image

(top: this PR, bottom: main.bsky.dev)

This pull request shaves ~2 MB of JS assets and some precious processing time spent during initial page load. It's still not ideal, but better than where we were before.


mary-ext:
#3925 pulls in `react-image-crop` which requires us to import a CSS file. Thankfully it seems that the manifest adds in css files into the entrypoint array, I can account for that in this pull request if needed.

haileyok (approved):

PR #3930, opened 2024-05-09
[Reduced Onboarding] Add feature gate, a:a swap onboarding state

estrattonbailey:
This PR adds the feature gate, and copies the full state from existing -> reduced. So even with the flag enabled, onboarding should function the same as it does today.

Test:
- load locally and clear onboarding, should look the same


haileyok (approved):

haileyok:
Pretty sure this will only work for development so might have to remove gate when we ship.

PR #3933, opened 2024-05-09
[Reduced Onboarding] Add profile step

estrattonbailey:
Stacked on #3931

No change to Interests of Finished steps. Only affects profile step. See states below:
- image upload works
- only allows jpg/png
- avi creator replaces image if selected (and vice versa)
- avi is only uploaded if within the feature gate
- can return to avi step and uploaded image is retained in memory

CleanShot 2024-05-09 at 13 00 44@2x
CleanShot 2024-05-09 at 13 00 55@2x
CleanShot 2024-05-09 at 13 01 03@2x
CleanShot 2024-05-09 at 13 01 14@2x
CleanShot 2024-05-09 at 13 01 18@2x
CleanShot 2024-05-09 at 13 01 20@2x
CleanShot 2024-05-09 at 13 01 23@2x
CleanShot 2024-05-09 at 13 01 26@2x
CleanShot 2024-05-09 at 13 12 27@2x
CleanShot 2024-05-09 at 13 12 42@2x
CleanShot 2024-05-09 at 13 12 46@2x

### DIFF
--- a/src/screens/Onboarding/StepProfile/index.tsx
+++ b/src/screens/Onboarding/StepProfile/index.tsx
@@ -71,8 +71,9 @@ export function StepProfile() {
   const creatorControl = Dialog.useDialogControl()
   const [error, setError] = React.useState('')
 
-  const {dispatch} = React.useContext(Context)
+  const {state, dispatch} = React.useContext(Context)
   const [avatar, setAvatar] = React.useState<Avatar>({
+    image: state.profileStepResults?.image,
     placeholder: emojiItems.at,
     backgroundColor: randomColor,
     useCreatedAvatar: false,
@@ -127,6 +128,7 @@ export function StepProfile() {
     if (imageUri) {
       dispatch({
         type: 'setProfileStepResults',
+        image: avatar.image,
         imageUri,
         imageMime: avatar.image?.mime ?? 'image/jpeg',
       })
@@ -134,13 +136,7 @@ export function StepProfile() {
 
     dispatch({type: 'next'})
     track('OnboardingV2:StepProfile:End')
-  }, [
-    avatar.image?.mime,
-    avatar.image?.path,
-    avatar.useCreatedAvatar,
-    dispatch,
-    track,
-  ])
+  }, [avatar.image, avatar.useCreatedAvatar, dispatch, track])
 
   const onDoneCreating = React.useCallback(() => {
     setAvatar(prev => ({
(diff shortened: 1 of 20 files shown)


haileyok:
Can we make these a little bigger on web? Feels pretty small
Screenshot 2024-05-09 at 12 17 01 PM

haileyok:
These might be a bit small too but I also have not seen it on-device yet. That said, we definitely center the groups of items

Screenshot 2024-05-09 at 12 26 19 PM

haileyok:
Functionally works well on all platforms. Just the few nits about styles for the avatar picker but otherwise looks good. Didn't look at any code though.

src/screens/Onboarding/StepProfile/types.ts
```diff
@@ -0,0 +1,148 @@
...
+    component: Lab,
+  },
+  piggyBank: {
+    name: 'piggyBank',
+    component: PiggyBank,
+  },
+  poop: {
+    name: 'poop',
+    component: Poop,
+  },
```
haileyok, line 137: love seeing this in the codebase hahaha

src/screens/Onboarding/StepProfile/index.tsx
```diff
@@ -1,55 +1,317 @@
...
+    async (opts?: ImagePickerOptions) => {
+      const response = await launchImageLibraryAsync({
+        exif: false,
+        mediaTypes: MediaTypeOptions.Images,
+        quality: 1,
+        ...opts,
+      })
+
+      return (response.assets ?? [])
+        .slice(0, 1)
```
haileyok, line 98: Definitely fine to leave the slice here (Android is weird and _should_ respect the limit but every platform version has a different image picker implementation), but just want to note to anyone else reviewing that the picker only allows one image to be picked.

src/screens/Onboarding/StepProfile/index.tsx
```diff
@@ -1,55 +1,317 @@
...
+    let image = items[0]
+    if (!image) return
+
+    if (!isWeb) {
+      image = await openCropper({
+        mediaType: 'photo',
+        cropperCircleOverlay: true,
+        height: image.height,
+        width: image.width,
+        path: image.path,
```
haileyok, line 168: This is confusing to me (we use this same logic in `UserAvatar` as well so not specific to this PR, I think I copied it from UserAvatar before, so we should fix both).

A photo that I pick can be any size, but our avatar should always have a 1:1 aspect ratio - and probably should have some capped size like 250x250. With these options, I can't select any region of the photo.


https://github.com/bluesky-social/social-app/assets/153161762/98520c09-3430-4be1-8118-8520fb2c9901

If I set the height and with to 250 though, I can select anywhere (which is what we want for avatars)


https://github.com/bluesky-social/social-app/assets/153161762/3c17d731-928b-4ff4-8713-e1962152af53

gaearon (approved):

PR #3934, opened 2024-05-09
lowercase ALT text typo

mozzius:


haileyok (approved):

haileyok:
This is what we use everywhere else in the app, makes sense.

PR #3935, opened 2024-05-09
[🐴] Add retries to all handlers

estrattonbailey:
Retries all requests twice before failing and alerting the user.


mozzius (approved):
lg

PR #3937, opened 2024-05-09
[🐴] Handle errors, improve styling

estrattonbailey:
This PR:
- gets rid of original retry message UI and handling and replaces with the same handling as other errors
- makes sure all network errors are handled
- improves styling

CleanShot 2024-05-09 at 16 36 02@2x
CleanShot 2024-05-09 at 16 36 32@2x
CleanShot 2024-05-09 at 16 38 00@2x
CleanShot 2024-05-09 at 16 38 04@2x


mozzius (approved):
lgtm!

PR #3939, opened 2024-05-09
[Session] Restore emailAuthFactor and emailConfirmed from last session

gaearon:
Carries over https://github.com/bluesky-social/social-app/pull/3728/files#diff-ba67025ff0027278031f772f88a9ed2f7fa4b12421d0a423ad74b4937aa2ca53R281 from #3728.

My bad that I missed it.

This fixes the false positive "you have not verified your email" popup. The `emailConfirmed` flag is not required so I forgot to specify it. So when you resume session from storage, the first render will not have the full object.

The full object still arrives a bit later (and gets written to the persistent storage) but then the same problem happens on next refresh. (The reason people don't see the same problem on next refreshes now is due to the auto-snoozing flag for this popup.)

## Test Plan

Comment out all the other checks in `shouldRequestEmailConfirmation`. Verify that the full object comes through when you refresh the page, and the popup does not appear.

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -1,5 +1,4 @@
-import {BskyAgent} from '@atproto/api'
-import {AtpSessionEvent} from '@atproto-labs/api'
+import {AtpSessionData, AtpSessionEvent, BskyAgent} from '@atproto/api'
 
 import {networkRetry} from '#/lib/async/retry'
 import {PUBLIC_BSKY_SERVICE} from '#/lib/constants'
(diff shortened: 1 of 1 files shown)


pfrazee (approved):

alikatgh:
Screenshot 2024-05-10 at 16 53 11

Thank you for JJ. Enjoying it fully. 

Here is the console log. You should send your style over HTTPS. 

P.S. I can't come up with better ways to reach out and say "ty". 

All excuse me for off-top.

PR #3941, opened 2024-05-10
[🐴] Report message dialog

mozzius:
Simulator Screenshot - iPhone 15 - 2024-05-10 at 01 04 05
Simulator Screenshot - iPhone 15 - 2024-05-10 at 01 26 14

### DIFF
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -50,6 +50,7 @@ let ConvoMenu = ({
   const {_} = useLingui()
   const t = useTheme()
   const leaveConvoControl = Prompt.usePromptControl()
+  const reportControl = Prompt.usePromptControl()
   const {mutate: markAsRead} = useMarkAsReadMutation()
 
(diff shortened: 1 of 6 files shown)


src/components/dms/ConvoMenu.tsx
```diff
@@ -165,12 +166,9 @@ let ConvoMenu = ({
...
               />
             </Menu.Item>
-            <Menu.Item
-              label={_(msg`Report account`)}
-              onPress={() => {}}
-              disabled>
+            <Menu.Item label={_(msg`Report chat`)} onPress={reportControl.open}>
               <Menu.ItemText>
-                <Trans>Report account</Trans>
+                <Trans>Report chat</Trans>
```
haileyok, line 171: Maybe should use "message" for this. Even if we call the service "Chat" I think message sounds better here.

mozzius: Agreed

estrattonbailey (approved):

PR #3942, opened 2024-05-10
Let Expo/Webpack handle CSS assets

mary-ext:
Stacked on https://github.com/bluesky-social/social-app/pull/3929, see https://github.com/bluesky-social/social-app/pull/3929#issuecomment-2102953912

Manually sharing CSS code between `web/index.html` (dev builds) and `bskyweb/templates/base.html` (production) is brittle, let's have it handled as a built asset instead.

This also means hot-reloading on development, well, hopefully it should.

Small savings gained from minification (reduction of ~2kb uncompressed)


bskyweb/templates/base.html
```diff
@@ -13,248 +13,6 @@
...
-      from { opacity: 0; }
-      to { opacity: 1; }
-    }
-
-    @keyframes avatarHoverFadeOut {
-      from { opacity: 1; }
-      to { opacity: 0; }
-    }
-  </style>
-  </style>
```
mary-ext, line 257: Never even noticed the double closing tag until now :sweat_smile:

src/App.web.tsx
```diff
@@ -1,5 +1,6 @@
 import 'lib/sentry' // must be near top
 import 'view/icons'
+import './style.css'
```
mary-ext, line 3: This import isn't right, ideally this should be imported very last but the ESLint sort import plugin doesn't seem to like that, not sure what I'd do to make this right.

mozzius: you can edit the sort plugin's order to put css files at the end

mary-ext: Yeah, that's what I was unsure about, because according to the config it should be putting relative imports last?

https://github.com/bluesky-social/social-app/blob/195c9f10456a988a7b09369e41c8a8cf6c94431f/.eslintrc.js#L52-L64

gaearon (approved):
awesome, thanks

PR #3945, opened 2024-05-10
[🐴] Adjust messages list styles

haileyok:
## Why

**[without whitespace](https://github.com/bluesky-social/social-app/pull/3945/files?w=1)**

I'm a little concerned about obscuring handles in DMs behind long display names. Since it's a more personal conversation, I think there's benefit in making sure that the user can actually see who it is that they are talking to.

Open to ideas on this of course!

Also removes the extra spinner states that were jumpy, and prevents the list from flickering on initial render on native.

### Chat List

Screenshot 2024-05-09 at 10 52 25 PM

Screenshot 2024-05-09 at 11 20 30 PM

### Avatar Alignment

Use the same size and (almost) alignment as feeds so there isn't a noticeable shift. I say almost because we really need to move away from the odd-number old style system - or convert the new one to odd numbers - so there's an ever-so-slight difference here. (Note, spacing between handle and message is a little higher here, but the image above is what you should go off of. Made a change)

#### In the feed

Screenshot 2024-05-09 at 10 39 12 PM

#### In the chat

Screenshot 2024-05-09 at 10 40 35 PM

### Chat Header

Screenshot 2024-05-09 at 10 54 30 PM

Screenshot 2024-05-09 at 10 54 53 PM

### Chat Loading

https://github.com/bluesky-social/social-app/assets/153161762/be799c68-942a-4104-9594-c80f6c98e698

### DIFF
--- a/src/screens/Messages/Conversation/index.tsx
+++ b/src/screens/Messages/Conversation/index.tsx
@@ -144,10 +144,14 @@ let Header = ({
         {profile ? (
           <View style={[a.align_center]}>
             <PreviewableUserAvatar size={32} profile={profile} />
-            <Text style={[a.text_lg, a.font_bold, isWeb ? a.mt_md : a.mt_sm]}>
+            <Text
+              style={[a.text_lg, a.font_bold, isWeb ? a.mt_md : a.mt_sm]}
+              numberOfLines={1}>
               {profile.displayName}
             </Text>
-            <Text style={[t.atoms.text_contrast_medium, {fontSize: 15}]}>
+            <Text
+              style={[t.atoms.text_contrast_medium, {fontSize: 15}]}
+              numberOfLines={1}>
               @{profile.handle}
             </Text>
           </View>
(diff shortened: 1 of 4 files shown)


src/screens/Messages/Conversation/index.tsx
```diff
@@ -154,12 +142,19 @@ let Header = ({
       )}
       <View style={[a.align_center, a.gap_sm, a.flex_1]}>
         {profile ? (
-          <>
+          <View style={[a.align_center]}>
```
estrattonbailey, line 145: This whole header is getting a little cramped, we may want to consider doing a horizontal layout so it doesn't get too tall

src/screens/Messages/Conversation/index.tsx
```diff
@@ -56,18 +55,6 @@ function Inner() {
...
-    convo.status === ConvoStatus.Uninitialized ||
-    convo.status === ConvoStatus.Initializing
-  ) {
-    return (
-      <CenteredView style={a.flex_1} sideBorders>
-        <Header />
-        <ListMaybePlaceholder isLoading />
-      </CenteredView>
-    )
-  }
```
estrattonbailey, line 69: This check was intentional, since it would allow us to make a, like, `useSafeConvo` which returns data already in active states. If we remove this we gotta start doing nullish coalescing, like you had to do below. Any chance we can keep this or does it totally mess up loader?

estrattonbailey (commented):
Looking good! Made a PR with my thoughts here https://github.com/bluesky-social/social-app/pull/3949

estrattonbailey (approved):

PR #3946, opened 2024-05-10
[🐴] Mutate data instead of invalidating queries when muting or unmuting

haileyok:
### Mutations

Right now we are invalidating queries whenever we update mute state. This will be problematic if you have a large list of active chats and are scrolled down say 100 chats. If you were to mute a chat in this case, the list would reset and you'd lose your scroll position.

Instead, we can mutate the data `onSuccess`. (No need to be optimistic in these cases)

Additionally, simplifying the logic. We can use params on the `mutateFn` to eliminate the need for two separate hooks here.

### Using `initialData`

Since we are using the `ConvoMenu` component in two screens - and we would prefer to be using the state of the convo from RQ - we can use `useConvoQuery` inside the `ConvoMenu`. We will prefill the query with `initialData` in most cases with the `ConvoView` that we have access to. Optionally though, you can pass only a convo ID (not useful in this case, but maybe in the future)

### DIFF
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -7,7 +7,10 @@ import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
 
 import {NavigationProp} from '#/lib/routes/types'
-import {useMarkAsReadMutation} from '#/state/queries/messages/conversation'
+import {
+  useConvoQuery,
+  useMarkAsReadMutation,
+} from '#/state/queries/messages/conversation'
 import {useLeaveConvo} from '#/state/queries/messages/leave-conversation'
 import {useMuteConvo} from '#/state/queries/messages/mute-conversation'
 import * as Toast from '#/view/com/util/Toast'
@@ -25,7 +28,7 @@ import * as Prompt from '#/components/Prompt'
 import {Bubble_Stroke2_Corner2_Rounded as Bubble} from '../icons/Bubble'
 
 let ConvoMenu = ({
-  convo,
+  convoId,
   profile,
   control,
   currentScreen,
@@ -33,7 +36,7 @@ let ConvoMenu = ({
   hideTrigger,
   triggerOpacity,
 }: {
-  convo: ChatBskyConvoDefs.ConvoView
+  convoId: string
   profile: AppBskyActorDefs.ProfileViewBasic
   onUpdateConvo?: (convo: ChatBskyConvoDefs.ConvoView) => void
   control?: Menu.MenuControlProps
@@ -48,11 +51,13 @@ let ConvoMenu = ({
   const leaveConvoControl = Prompt.usePromptControl()
   const {mutate: markAsRead} = useMarkAsReadMutation()
 
+  const {data: convo} = useConvoQuery(convoId)
+
   const onNavigateToProfile = useCallback(() => {
     navigation.navigate('Profile', {name: profile.did})
   }, [navigation, profile.did])
 
-  const {mutate: muteConvo} = useMuteConvo(convo.id, {
+  const {mutate: muteConvo} = useMuteConvo(convoId, {
     onSuccess: data => {
       if (data.muted) {
         Toast.show(_(msg`Chat muted`))
@@ -65,7 +70,7 @@ let ConvoMenu = ({
     },
   })
 
-  const {mutate: leaveConvo} = useLeaveConvo(convo.id, {
+  const {mutate: leaveConvo} = useLeaveConvo(convoId, {
     onSuccess: () => {
       if (currentScreen === 'conversation') {
         navigation.replace('Messages')
@@ -110,7 +115,7 @@ let ConvoMenu = ({
                 label={_(msg`Mark as read`)}
                 onPress={() =>
                   markAsRead({
-                    convoId: convo.id,
+                    convoId: convoId,
                   })
                 }>
                 <Menu.ItemText>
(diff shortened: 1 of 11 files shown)


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -161,10 +161,10 @@ export function useNotificationsHandler() {
 
     const responseReceivedListener =
       Notifications.addNotificationResponseReceivedListener(e => {
-        if (e.notification.request.identifier === prevIdentifier.current) {
+        if (e.notification.date === prevDate.current) {
           return
         }
-        prevIdentifier.current = e.notification.request.identifier
+        prevDate.current = e.notification.date
```
haileyok, line 167: squeezing this in here. talked to devin, we're using `identifier` for another purpose it seems (I think gorush is messing this up, it's supposed to always be unique...) Date is also unique though (it is ridiculously precise...), so we will use that instead.

src/components/dms/ConvoMenu.tsx
```diff
@@ -28,16 +29,15 @@ import * as Prompt from '#/components/Prompt'
...
   profile,
-  onUpdateConvo,
   control,
   currentScreen,
   showMarkAsRead,
   hideTrigger,
   triggerOpacity,
 }: {
-  convo: ChatBskyConvoDefs.ConvoView
+  convoIdOrConvo: string | ConvoView
```
estrattonbailey, line 40: Do we need to accept both? If not, let's just use one or the other, preferably `id` I think.

src/state/queries/messages/conversation.ts
```diff
@@ -9,20 +10,22 @@ import {useHeaders} from './temp-headers'
 const RQKEY_ROOT = 'convo'
 export const RQKEY = (convoId: string) => [RQKEY_ROOT, convoId]
 
-export function useConvoQuery(convoId: string) {
+export function useConvoQuery(convoIdOrConvo: string | ConvoView) {
```
estrattonbailey, line 13: Would also update here if we don't need both

src/state/queries/messages/conversation.ts
```diff
@@ -37,9 +40,11 @@ export function useMarkAsReadMutation() {
       convoId,
       messageId,
     }: {
-      convoId: string
+      convoId?: string
```
estrattonbailey, line 43: If at all possible I'd like to avoid this. If we don't have this data, we probably shouldn't be rendering the components that call this mutation 🤔

src/state/queries/messages/leave-conversation.ts
```diff
@@ -11,7 +11,7 @@ import {RQKEY as CONVO_LIST_KEY} from './list-converations'
 import {useHeaders} from './temp-headers'
 
 export function useLeaveConvo(
-  convoId: string,
+  convoId: string | undefined,
```
estrattonbailey, line 14: Same here

src/state/queries/messages/mute-conversation.ts
```diff
@@ -1,18 +1,18 @@
...
 
-import {logger} from '#/logger'
 import {useDmServiceUrlStorage} from '#/screens/Messages/Temp/useDmServiceUrlStorage'
 import {RQKEY as CONVO_KEY} from './conversation'
 import {RQKEY as CONVO_LIST_KEY} from './list-converations'
 import {useHeaders} from './temp-headers'
 
 export function useMuteConvo(
-  convoId: string,
+  convoId: string | undefined,
```
estrattonbailey, line 15: +1

src/components/dms/ConvoMenu.tsx
```diff
@@ -52,31 +52,26 @@ let ConvoMenu = ({
   const leaveConvoControl = Prompt.usePromptControl()
   const {mutate: markAsRead} = useMarkAsReadMutation()
 
+  const {data: convo} = useConvoQuery(convoIdOrConvo)
```
estrattonbailey, line 55: This might be a reason to handle `ConvoView` in the `Convo` agent so that they're always in sync. Not sure it matters to much rn, but it might in the future

src/components/dms/ConvoMenu.tsx
haileyok: Probably will never happen, but i left it because do we ever anticipate it for something like viewing a report? We'd have to load it first. We actually do this with RQ in other areas as well fwiw.

estrattonbailey: I'm not sure I follow, seems like `ConvoMenu` should only be used in contexts where the convo is already loaded either way. And if we need to rely on server state, then fetching it within the component is probably the right call. Otherwise if we pass in a `ConvoView`, how do we know that it's revalidating on mutate?

It's a small thing, but I don't think we should build loose contracts from the get-go if we can help it.

This isn't a blocker though imo ✌️

haileyok: Right for sure. There is sense in doing that in the overall list - you have access to `ConvoListChats` (or...whatever the type is I forget so many names!) and can just pass in each value when you render the list.

The problem is when we're in an individual chat screen. We have `convo` from the chat service, but unless we add `mute()` to the chat service then we have to keep track of the mute state some other way.

```ts
const [muted, setMuted] = useState(convo.muted)

const onMutedChange = (muted) => setMuted(muted)

// Return dialog that calls onMutedChange
```

This to me feels more dangerous, since we definitely are not going to know of any changes that happen elsewhere. Even if we mutate with `setQueryData` or invalidate queries with `invalidateQuery`, our mute state in _this_ screen still wouldn't be updated.

estrattonbailey: Ok I see what you're saying. Yeah like I mentioned below maybe this needs to move into `Convo`.

haileyok: Will mark this one as a draft for now pending what we end up doing 👍

estrattonbailey (approved):

PR #3947, opened 2024-05-10
[🐴] Remove extra spinner states from chat screen

haileyok:
Cherry picking this out of https://github.com/bluesky-social/social-app/pull/3945, because it's really separate from that PR and I'd like to get this one in to test on device.

### DIFF
--- a/src/screens/Messages/Conversation/index.tsx
+++ b/src/screens/Messages/Conversation/index.tsx
@@ -21,7 +21,6 @@ import {MessagesList} from '#/screens/Messages/Conversation/MessagesList'
 import {atoms as a, useBreakpoints, useTheme} from '#/alf'
 import {ConvoMenu} from '#/components/dms/ConvoMenu'
 import {Error} from '#/components/Error'
-import {ListMaybePlaceholder} from '#/components/Lists'
 import {Text} from '#/components/Typography'
 import {ClipClopGate} from '../gate'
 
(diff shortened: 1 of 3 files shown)


src/state/messages/convo/agent.ts
```diff
@@ -553,7 +553,7 @@ export class Convo {
           {
             cursor: nextCursor,
             convoId: this.convoId,
-            limit: isNative ? 40 : 60,
+            limit: isNative ? 30 : 60,
```
haileyok, line 556: This needs to be the same as `initialNumToRender` in `messagesList`. If it's larger, there will be a scroll animation on render.

I'd like to keep this below 30 if we can, especially whenever we start considering adding other content types than text (honestly we might need to decrease the number when that happens, at least on native)

estrattonbailey: Heard chef 🫡

src/screens/Messages/Conversation/index.tsx
```diff
@@ -56,18 +55,6 @@ function Inner() {
...
-    convo.status === ConvoStatus.Uninitialized ||
-    convo.status === ConvoStatus.Initializing
-  ) {
-    return (
-      <CenteredView style={a.flex_1} sideBorders>
-        <Header />
-        <ListMaybePlaceholder isLoading />
-      </CenteredView>
-    )
-  }
```
estrattonbailey, line 69: Commented on other PR already, but want to discuss this a bit. Not necessarily a blocker but want to be sure we can't do this another way.

estrattonbailey (approved):

PR #3950, opened 2024-05-10
[🐴] Clean up after deleting message

estrattonbailey:
I noticed the other day that I was retaining deleted messages in the optimistic `deletedMessages` array. `pastMessages` are things loaded from history, and `newMessages` are things added from the firehose. So I need to check for existence in both those arrays and clean up once I get a committed event from the firehose.


mozzius (commented):
You left a console.log in there but otherwise lgtm

mozzius (approved):

PR #3952, opened 2024-05-10
[🐴] Simplify message passing, cleanup

estrattonbailey:
Simplifies the contract between the two agents and removes some unused types. It also ensures the events are torn down when the class is suspended.

### DIFF
--- a/src/state/messages/convo/agent.ts
+++ b/src/state/messages/convo/agent.ts
@@ -107,26 +107,6 @@ export class Convo {
     } else {
       DEBUG_ACTIVE_CHAT = this.convoId
     }
-
-    this.events.on(
-      event => {
-        switch (event.type) {
-          case 'connect': {
-            this.onFirehoseConnect()
-            break
-          }
-          case 'error': {
-            this.onFirehoseError(event.error)
-            break
-          }
-          case 'logs': {
-            this.ingestFirehose(event.logs)
-            break
-          }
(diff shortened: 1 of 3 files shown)


src/state/messages/convo/agent.ts
```diff
@@ -713,6 +741,8 @@ export class Convo {
       id: tempId,
       message,
     })
+    // remove on each send, it might go through now without user having to click
+    this.footerItems.delete(ConvoItemError.PendingFailed)
```
estrattonbailey, line 745: Noticed while testing this that if a send fails, and you send another that succeeds, you'll have no pending messages but the retry error will still be present. In this case, which is now fixed, if you click retry, it simply removes itself and does nothing, since there are no pending messages left.

src/state/messages/convo/agent.ts
```diff
@@ -601,6 +602,33 @@ export class Convo {
...
+            this.onFirehoseError(event.error)
+            break
+          }
+          case 'logs': {
+            this.ingestFirehose(event.logs)
+            break
+          }
+        }
+      },
+      {convoId: this.convoId},
```
estrattonbailey, line 628: Without this optional param, it will receive all `logs` events for all chats. For the purposes of `Convo` we obviously only want this particular convo.

haileyok (approved):

PR #3955, opened 2024-05-10
[🐴] Bump max width of message

haileyok:


mozzius (approved):
Lets see how this looks on a real device

PR #3957, opened 2024-05-10
Update zh-TW

KuwaLee:
@auroursa Please help me review again


auroursa (commented):
It looks like you ran intl:extract manually, so there are a lot of unexpected changes. Other than that, LGTM.

PR #3966, opened 2024-05-11
Onboarding fixes

gaearon:
See individual commits.

- https://github.com/bluesky-social/social-app/commit/d8ee3b630d2ae907df0a0e613cd3640ed683296a: Fixes a flash after new onboarding that caused Following to get selected instead of Discover. The problem is we were writing Discover into `savedFeeds` via `overwriteSavedFeeds`, but RQ has already cached initial `savedFeeds` (just the `following`), and we had a flash of that during the initial Home render. The fix is to invalidate RQ cache before going to Home.
- d8265e52650941a565e8593316a2326e9931b39a: There is no reason for this call to be serial from what I can see. It can be parallelized against bulk follows and setting prefs. Only setting prefs has to be serial (because lol, we designed it to be this way).
- https://github.com/bluesky-social/social-app/commit/03ecf4cce77b965b995965f8ec4283729cb1a1f7: Unnecessary await if we're not uploading anything. Let's remove that.
- https://github.com/bluesky-social/social-app/commit/cf29021070dcaa6bb42f2811000dbcd6d0900a4f: Unnecessary waterfall. Let the image upload begin in parallel with getting the profile.
- fd692948d842c4d82b7e371f394fda0f49ddda98: Drive-by fix for the same issue in regular profile editor. We can upload in parallel without waiting for account info.
- 528877bd012587b9dac8a72bc0a184a6946af0a0: It's weird to see an empty profile pic after finishing onboarding. Let's invalidate that query.
- https://github.com/bluesky-social/social-app/commit/a72bb815de7fd53df14704e44a65c6b7862533b9: The code in main adds the profile image step both to old and new onboarding, but only new onboarding used to persist this image. Oops. This makes it persist in both cases. If we don't want that step, we can gate this code again.

## Test Plan

Hardcode either `true` or `false` like this:

```diff
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -115,6 +115,9 @@ export function useGate(): (gateName: Gate) => boolean {
   }
   const gate = React.useCallback(
     (gateName: Gate): boolean => {
+      if (gateName === 'reduced_onboarding_and_home_algo') {
+        return false
+      }
```

Then go through both onboardings. Observe that:

- There is no flash of Following tab after onboard
- You reliably land on Discover
- Your profile pic is set

Also verify the normal "edit profile" workflow isn't borked (incl. avatar and cover upload).

New onboarding:

https://github.com/bluesky-social/social-app/assets/810438/bffbd1d0-7539-4a71-855a-e6260bccfc95

Old onboarding:

https://github.com/bluesky-social/social-app/assets/810438/341dcb0d-2dc1-4743-bda4-8ea4f8c62f70

### DIFF
--- a/src/screens/Onboarding/StepFinished.tsx
+++ b/src/screens/Onboarding/StepFinished.tsx
@@ -3,13 +3,17 @@ import {View} from 'react-native'
 import {TID} from '@atproto/common-web'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
+import {useQueryClient} from '@tanstack/react-query'
 
 import {useAnalytics} from '#/lib/analytics/analytics'
 import {BSKY_APP_ACCOUNT_DID, IS_PROD_SERVICE} from '#/lib/constants'
 import {DISCOVER_SAVED_FEED, TIMELINE_SAVED_FEED} from '#/lib/constants'
 import {logEvent, useGate} from '#/lib/statsig/statsig'
 import {logger} from '#/logger'
-import {useOverwriteSavedFeedsMutation} from '#/state/queries/preferences'
+import {
+  preferencesQueryKey,
+  useOverwriteSavedFeedsMutation,
+} from '#/state/queries/preferences'
 import {useAgent} from '#/state/session'
 import {useOnboardingDispatch} from '#/state/shell'
 import {uploadBlob} from 'lib/api'
@@ -41,6 +45,7 @@ export function StepFinished() {
   const onboardDispatch = useOnboardingDispatch()
   const [saving, setSaving] = React.useState(false)
   const {mutateAsync: overwriteSavedFeeds} = useOverwriteSavedFeedsMutation()
+  const queryClient = useQueryClient()
   const {getAgent} = useAgent()
   const gate = useGate()
 
@@ -139,6 +144,16 @@ export function StepFinished() {
       // don't alert the user, just let them into their account
     }
 
+    // Try to ensure that prefs are up-to-date by the time we render Home.
+    await queryClient
+      .invalidateQueries({
+        queryKey: preferencesQueryKey,
+      })
+      .catch(e => {
+        logger.error(e)
+        // Keep going.
+      })
+
     setSaving(false)
     dispatch({type: 'finish'})
     onboardDispatch({type: 'finish'})
@@ -154,6 +169,7 @@ export function StepFinished() {
     track,
     getAgent,
     gate,
+    queryClient,
   ])
 
   React.useEffect(() => {
--- a/src/screens/Onboarding/StepFinished.tsx
+++ b/src/screens/Onboarding/StepFinished.tsx
@@ -120,19 +120,17 @@ export function StepFinished() {
         })(),
 
         (async () => {
-          if (gate('reduced_onboarding_and_home_algo')) {
-            const {imageUri, imageMime} = profileStepResults
-            if (imageUri && imageMime) {
-              const blobPromise = uploadBlob(getAgent(), imageUri, imageMime)
-              await getAgent().upsertProfile(async existing => {
-                existing = existing ?? {}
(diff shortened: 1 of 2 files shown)


haileyok:
Okay going through each commit right now:

- [] https://github.com/bluesky-social/social-app/commit/d8ee3b630d2ae907df0a0e613cd3640ed683296a: Fixes a flash after new onboarding that caused Following to get selected instead of Discover. The problem is we were writing Discover into `savedFeeds` via `overwriteSavedFeeds`, but RQ has already cached initial `savedFeeds` (just the `following`), and we had a flash of that during the initial Home render. The fix is to invalidate RQ cache before going to Home.
- [] d8265e52650941a565e8593316a2326e9931b39a: There is no reason for this call to be serial from what I can see. It can be parallelized against bulk follows and setting prefs. Only setting prefs has to be serial (because lol, we designed it to be this way).
- [] https://github.com/bluesky-social/social-app/commit/03ecf4cce77b965b995965f8ec4283729cb1a1f7: Unnecessary await if we're not uploading anything. Let's remove that.
- [] https://github.com/bluesky-social/social-app/commit/cf29021070dcaa6bb42f2811000dbcd6d0900a4f: Unnecessary waterfall. Let the image upload begin in parallel with getting the profile.
- [] fd692948d842c4d82b7e371f394fda0f49ddda98: Drive-by fix for the same issue in regular profile editor. We can upload in parallel without waiting for account info.
- [] 528877bd012587b9dac8a72bc0a184a6946af0a0: It's weird to see an empty profile pic after finishing onboarding. Let's invalidate that query.
- [] https://github.com/bluesky-social/social-app/commit/a72bb815de7fd53df14704e44a65c6b7862533b9: The code in main adds the profile image step both to old and new onboarding, but only new onboarding used to persist this image. Oops. This makes it persist in both cases. If we don't want that step, we can gate this code again.

haileyok:
> There is no reason for this call to be serial from what I can see. It can be parallelized against bulk follows and setting prefs. Only setting prefs has to be serial (because lol, we designed it to be this way).

Okay I agree with this one, but I think we need to invalidate the profile query afterwards. If I upload an avatar during onboarding, then the profile icon in the bottom right-hand corner does not update until after I open up my profile, which could be a little confusing if I expect to have uploaded an avatar.

https://github.com/bluesky-social/social-app/assets/153161762/8da04cac-cf48-4677-9497-2c3b9380dbcb

gaearon:
>Okay I agree with this one, but I think we need to invalidate the profile query afterwards. If I upload an avatar during onboarding, then the profile icon in the bottom right-hand corner does not update until after I open up my profile, which could be a little confusing if I expect to have uploaded an avatar.

Yeah, https://github.com/bluesky-social/social-app/commit/528877bd012587b9dac8a72bc0a184a6946af0a0 should do it.

I'm not sure if it's not guaranteed to come through (e.g. `useProfileUpdateMutation` has `whenAppViewReady` call to be more rigorous). However I think we should also be mindful of the delay we're adding here — don't want to bounce. So I'd probably lean towards not adding more waiting here. Note the image you're using is very high res.

What we really should do (IMO) is upload the avatar in the earlier step. Of course that would have to be able to retry on failure here, so it would have to be more like an optimistic upload than a "let's just move this code there" kind of thing.

haileyok:
Yea, agree w/ those points. FWIW it also happens even if I upload the generic created avatar (I have not gotten to your mentioned commit yet though so will check against that here in a minute).

haileyok:
Okay yea, that does fix the issue (even w/ the high-res image). Seems like your idea would be a good one. And you're right about not wanting to make this take longer (especially on slower connections). Maybe:

- Attempt to upload the image as soon as next is pressed from that screen
- If for some reason it fails, attempt to do it again the same way we are right now
- If the second attempt fails...?

I think we can do that in a separate PR though.

haileyok (approved):

PR #3967, opened 2024-05-11
Fix flashes when replacing For You

gaearon:
We were triggering multiple mutations, each of which invalidates the prefs. So the transition felt very jumpy (two full screen refreshes one after another). This fixes it by introducing a special mutation just for this case, with one invalidation at the end.

This also changes "Remove" button to move you to Discover if you already have it pinned.

Note it's still a bit janky because there's an interrupting pager animation. But that already exists on main too.

## Test Plan

Press Remove when Discover is not saved:

https://github.com/bluesky-social/social-app/assets/810438/39c7a851-5188-487a-bee6-45ed58d9f0c7

Press Remove when Discover is saved (but not pinned):

https://github.com/bluesky-social/social-app/assets/810438/022a96da-dae2-444d-9382-793582a3b239

Press Remove when Discover is pinned:

https://github.com/bluesky-social/social-app/assets/810438/f15d035d-4602-40c8-90aa-e64bd261228d

Press Replace when Discover is not saved:

https://github.com/bluesky-social/social-app/assets/810438/5ffca8db-de66-47e4-af86-912bd2c98e91

Press Replace when Discover is saved (but not pinned):

https://github.com/bluesky-social/social-app/assets/810438/95f22709-45f8-4c03-b189-e6ecc0b61088

### DIFF
--- a/src/state/queries/preferences/index.ts
+++ b/src/state/queries/preferences/index.ts
@@ -6,6 +6,7 @@ import {
 import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query'
 
 import {track} from '#/lib/analytics/analytics'
+import {PROD_DEFAULT_FEED} from '#/lib/constants'
 import {replaceEqualDeep} from '#/lib/functions'
 import {getAge} from '#/lib/strings/time'
 import {STALE} from '#/state/queries'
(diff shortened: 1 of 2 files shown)


haileyok (approved):
Awesome, works as expected in all the cases. Will look at the animation jank later today after a few other things.

src/state/queries/preferences/index.ts
```diff
@@ -244,6 +245,45 @@ export function useRemoveFeedMutation() {
...
+          {
+            ...discoverFeedConfig,
+            pinned: true,
+          },
+        ])
+      }
+      // triggers a refetch
+      await queryClient.invalidateQueries({
+        queryKey: preferencesQueryKey,
+      })
```
yogiduzit, line 282: Why not do this in an `onSuccess` callback? https://tanstack.com/query/latest/docs/framework/react/guides/invalidations-from-mutations

PR #3968, opened 2024-05-11
FeedFeedback fixes

gaearon:
- https://github.com/bluesky-social/social-app/commit/3cc47165a9bd6f064ef269e4995838e4b6d97502: Lower the time threshold to 1500ms as discussed with @whyrusleeping.
- https://github.com/bluesky-social/social-app/commit/14f00d55cae9a8ed0168b30b3ac53c0388982e52: Replies were missing `feedContext` because the parent item (which was unshifted) didn't have it. For our purposes, the entire convo has the same context. The fix is mostly copypaste from how we handle `source`. There's probably a simpler way to write this but I don't understand the data structure for sure — and this will definitely work.

## Test Plan

Read the code for the first change. For the second change, added

```diff
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -294,6 +294,14 @@ export function usePostFeedQuery(
                           AppBskyFeedPost.validateRecord(item.post.record)
                             .success
                         ) {
+                          let before = item.feedContext
+                          let after =
+                            i === 0 && slice.feedContext
+                              ? slice.feedContext
+                              : item.feedContext
+                          if (before !== after) {
+                            console.log('before != after', before, after)
+                          }

```

and verified that the log shows up for replies

Screenshot 2024-05-11 at 20 32 32

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -71,6 +71,10 @@ export class FeedViewPostsSlice {
       ?.__source as ReasonFeedSource
   }
 
+  get feedContext() {
+    return this.items.find(item => item.feedContext)?.feedContext
+  }
+
   containsUri(uri: string) {
     return !!this.items.find(item => item.post.uri === uri)
   }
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Yep looks good. I went for a simpler feedContext fallback (don't need to pay attention to the `i===0` for this)

PR #3977, opened 2024-05-12
Move request for notifications permissions to `HomeReadyScreen`

haileyok:
## Why

Asking for permissions randomly at the start of onboarding can be pretty jarring and might result in more denials of permission. Instead, we can ask for permission after onboarding is complete.

We also want to be able to track this, so I have:

1. Added a feature gate on whether to show this at the start of onboarding or afterward
  - Display it on the interests screen of onboarding if the reduced onboarding gate is false
  - Display it on the profile screen of onboarding if the reduced onboarding gate is true
2. Added a new log type - `notifications:request`
  - Accepts a `status` with values `granted`, `denied`, or `undetermined` - the default values for Expo Notifications

On top of these changes, I am fixing/cleaning up some other logic. On main right now, we can see that we are registering the push token multiple times on app launch:

Screenshot 2024-05-11 at 11 11 46 PM

As noted below, that is because we are

1. Getting the device push token and registering it
2. Creating a listener that calls register token whenever the token changes. Whenever it changes, we register the new token (the one we just retrieved in step one!)

Instead, what I have done is call `getDevicePushToken()` on DID change _without_ registering it, then let the change listener handle the registration. This is the result:

Screenshot 2024-05-11 at 11 14 54 PM

## Test Plan

Different gate configurations should be tested:

### Reduced Onboarding: False, Request After Onboarding: False

https://github.com/bluesky-social/social-app/assets/153161762/bc741932-dfa6-4220-a9a0-18fc46c78548

### Reduced Onboarding: True, Request After Onboarding: False

https://github.com/bluesky-social/social-app/assets/153161762/4485930b-54c5-4e7f-920a-1a2a18acf1d3

### Reduced Onboarding: True/False, Request After Onboarding: True

https://github.com/bluesky-social/social-app/assets/153161762/90e7d40f-dfbe-4ae0-a765-dd1103b2e177

### DIFF
--- a/src/lib/notifications/notifications.ts
+++ b/src/lib/notifications/notifications.ts
@@ -1,9 +1,8 @@
 import React from 'react'
 import * as Notifications from 'expo-notifications'
-import {BskyAgent} from '@atproto/api'
 
 import {logger} from '#/logger'
-import {SessionAccount, useAgent, useSession} from '#/state/session'
+import {useAgent, useSession} from '#/state/session'
 import {logEvent, useGate} from 'lib/statsig/statsig'
 import {devicePlatform, isNative} from 'platform/detection'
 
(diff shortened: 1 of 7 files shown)


src/lib/notifications/notifications.ts
```diff
@@ -42,38 +41,52 @@ export async function requestPermissionsAndRegisterToken(
...
+  const {getAgent} = useAgent()
+  const {currentAccount} = useSession()
+
+  const prevAccountDid = React.useRef<string | undefined>(undefined)
+
+  React.useEffect(() => {
+    if (
+      !currentPermissions?.granted ||
+      !currentAccount ||
+      prevAccountDid.current === currentAccount.did
```
haileyok, line 55: I am not actually certain why we are using `prevAccountDid` (or before, it was called `currentAccountDid`). I assume that maybe `getAgent` can change? I couldn't get this to happen when I removed this check but decided to leave it incase there was a reason.

gaearon: We should remove this check, it’s from the old days before we started remounting the tree. You can assume `currentAccount` never changes from the tree’s perspective.

gaearon: So the whole notion of prevAccountDid should be deleted.

estrattonbailey (approved):
LGTM

PR #1051, opened 2023-07-25
Bump react-native-paste-input; remove hack

loganrosen:
Bumping `react-native-paste-input` to 0.6.4 to pull in [my fix for the blur issue on unmount](https://github.com/mattermost/react-native-paste-input/pull/20), which had led to a hack that's no longer needed.

PR #3980, opened 2024-05-12
Upgrade to Expo 51 and React Native 0.74

haileyok:
### Will put together some sort of test plan for this soon.

Testing branch for now. We should aim to upgrade to Expo 51/RN 0.74 as soon as possible as it provides a clear path to us being able to use the new architecture. I've bumped packages to latest SDK 51 versions. Notably:

- Upgrading `expo-image`. This also requires bumping versions of `SDWebImage` in our `ExpoGifView` module
- Upgrading `expo-image-picker`. This lets us remove the patch we were previous relying on, since these have now been merged:
	- https://github.com/expo/expo/pull/28514
	- https://github.com/expo/expo/pull/28509
- `hermes` now requires the following if you have not already done so (I have also updated `build.md` with this info):
	  - If you are using Apple Silicon and this is the first time you are building for RN 0.74+, you may need to run:
		    - `arch -arm64 brew install llvm`
		    - `sudo gem install ffi`
- One change to the `app.config.js`. See: https://docs.expo.dev/bare/upgrade/?fromSdk=50&toSdk=51
- Remove bulk of `react-native-paste-input` patch. This patch is extremely old and (from what I can see reading issues in their repo) has been resolved in later versions (that we are now upgrading to). We should check that whatever the problem before was is not happening now though!


src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -114,7 +114,9 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
       (scrollRef: AnimatedRef<any> | null, atIndex: number) => {
         scrollRefs.modify(refs => {
           'worklet'
-          refs[atIndex] = scrollRef
+          if (scrollRef != null) {
+            refs[atIndex] = scrollRef
+          }
           return refs
```
haileyok, line 120: this is one to look at. haven't looked into what actually changed here yet

src/view/com/util/Views.web.tsx
```diff
@@ -66,7 +68,7 @@ export const FlatList_INTERNAL = React.forwardRef(function FlatListImpl<ItemT>(
     desktopFixedHeight,
     ...props
   }: React.PropsWithChildren<FlatListProps<ItemT> & AddedProps>,
-  ref: React.Ref<Animated.FlatList<ItemT>>,
+  ref: React.Ref<FlatList<any>>,
```
haileyok, line 71: note to self that `any` should be `ItemT`

patches/@mattermost+react-native-paste-input+0.7.1.patch
```diff
@@ -0,0 +1,16 @@
...
+     _backedTextInputView.textInputDelegate = self;
+
++    // Disable inline predictions to prevent jank in the composer
++    if (@available(iOS 17.0, *)) {
++      _backedTextInputView.inlinePredictionType = UITextInlinePredictionTypeNo;
++    }
++
+     [self addSubview:_backedTextInputView];
+   }
+
```
haileyok, line 16: No changes to this patch, just had to re-create because line numbers changed.

patches/expo-notifications+0.28.1.patch
```diff
@@ -1,13 +1,13 @@
 diff --git a/node_modules/expo-notifications/android/build.gradle b/node_modules/expo-notifications/android/build.gradle
-index 97bf4f4..6e9d427 100644
+index d233e1f..cc2f856 100644
```
haileyok, line 2: Same here, just had to re-create the patch.

haileyok:
Another note:

I'm not sure why we still had such a large patch file for the paste input. The changes that I recently added were only these lines: https://github.com/bluesky-social/social-app/pull/3363/files

We also seem to have removed the hack introduced prior in this PR: https://github.com/bluesky-social/social-app/pull/1051

We could roll this back to the 0.6.4 version and keep the old patch most likely, though I'm still not sure what the point of that patch was. It seems to all be build intermediaries. There have not been any changes besides some dep updates between 0.6.4. and 0.7.1, so I don't feel strongly either way (unless I am missing something)

src/view/com/pager/PagerWithHeader.tsx
gaearon: Hmm I think we'd actually want to properly clear these.

gaearon: fixed differently

haileyok: ty for looking at that! wasn't really sure what changed here 🙏

gaearon (approved):
alright let's give this a go

PR #3981, opened 2024-05-13
get a little more accurate with month length

haileyok:
## Why

Because we are assuming months have 28 days, we're starting to see some cases like this:

image

Obviously, we know that a year cannot have 13 months in it, but we're starting to encounter cases like this because...well it was just Bluesky's "mainstream" birthday!

So, what do we want to do instead?

- Keep the same logic as before for anything under 30 days
	- 29 days should be `29d`
	- 30 days should be `1mo`
- Anything like 1 month and 15 days, 1 month and 20 days, etc. should display `1mo` in most cases.
- As we approach 28, 29, or 30 days, we can show the next month, i.e. 1 month and 19 days should show `2mo`
- As we reach the end of the year, we should not show `12mo` but instead begin showing the full date. I.e. 11 months and 29 days should show `4/20/24`
- As before, anything over a year old should also show the localized date string.

## Test Plan

Screenshot 2024-05-12 at 6 14 14 PM

I have added a variety of test cases to `string.test.ts`. Remember, this still is not _precise_, but it more closely resembles reality - especially as we get into the later months or as in this case the "13 months" case. If we truly want to be precise, 
we would either need to write some additional (annoying) logic or use a library like dayjs which is a little overkill for this purpose.

Of course feel free to tweak this however you feel like.

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -143,6 +143,10 @@ describe('makeRecordUri', () => {
 })
 
 describe('ago', () => {
+  const oneYearDate = new Date(
+    new Date().setMonth(new Date().getMonth() - 11),
+  ).setDate(new Date().getDate() - 28)
+
   const inputs = [
     1671461038,
     '04 Dec 1995 00:12:00 GMT',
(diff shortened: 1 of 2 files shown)


surfdude29:
> * Anything like 1 month and 15 days, 1 month and 20 days, etc. should display `1mo` in most cases.
> * As we approach 28, 29, or 30 days, we can show the next month, i.e. 1 month and 19 days should show `2mo`

I think there might be a typo in the second sentence above, I'm guessing it should be: "i.e. 1 month and **29** days should show `2mo`"?

__tests__/lib/string.test.ts
```diff
@@ -151,7 +155,32 @@ describe('ago', () => {
     new Date().setMinutes(new Date().getMinutes() - 10),
     new Date().setHours(new Date().getHours() - 1),
     new Date().setDate(new Date().getDate() - 1),
+    new Date().setDate(new Date().getDate() - 20),
```
estrattonbailey, line 158: Don't need to change in this PR, I realize it was already set up like this, but good tip for tests: each of these should be written as an `it` clause, with a description, so when one breaks we know exactly which failed. Written like this, one failure fails the whole thing, and determining which failed is harder.

estrattonbailey: Also it's hard to review bc idk which lines up with which :P

estrattonbailey (approved):
LGTM — we could consider a library for this also

haileyok:
> > * Anything like 1 month and 15 days, 1 month and 20 days, etc. should display `1mo` in most cases.
> > * As we approach 28, 29, or 30 days, we can show the next month, i.e. 1 month and 19 days should show `2mo`
> 
> I think there might be a typo in the second sentence above, I'm guessing it should be: "i.e. 1 month and **29** days should show `2mo`"?

ha yea, that's what i meant 😅

PR #3983, opened 2024-05-13
Replace e2e tests with Maestro

pfrazee:
This PR replaces all of the e2e tests with [Maestro](https://maestro.mobile.dev/)

The good news is, it's generally faster than detox, easier to run, easier to write tests for, and importantly it works (Detox had completely broken).


__e2e__/flows/create-account.yml
```diff
@@ -0,0 +1,37 @@
...
+- inputText: "hunter2"
+- pressKey: Enter
+- tapOn:
+    id: "nextBtn"
+- tapOn:
+     id: "handleInput"
+- inputText: "e2e-test"
+- tapOn:
+    id: "nextBtn"
+
```
haileyok, line 37: I had to add this to get it to work, otherwise it would end the test too soon and I wouldn't actually have a session when the next test ran

```yaml
- waitForAnimationToEnd:
    timeout: 5000
````

haileyok (approved):
All the tests passed! This is awesome, especially Maestro Studio. Thanks for this!

estrattonbailey:
Oh boy, can't wait for this!

PR #3984, opened 2024-05-13
Update catalan messages.po

ivanbea:
New lines update
You can check it @jordimas @darccio @ferranrego @MiquelAdell @johnnydement @surfdude29

### DIFF
--- a/src/locale/locales/ca/messages.po
+++ b/src/locale/locales/ca/messages.po
@@ -22,7 +22,7 @@ msgstr "(sense correu)"
 
 #: src/view/com/notifications/FeedItem.tsx:239
 msgid "{0, plural, one {{formattedCount} other} other {{formattedCount} others}}"
-msgstr ""
+msgstr "{0, plural, one {{formattedCount} other} other {{formattedCount} others}}"
 
 #: src/view/shell/desktop/RightNav.tsx:168
 #~ msgid "{0, plural, one {# invite code available} other {# invite codes available}}"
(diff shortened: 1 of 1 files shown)


src/locale/locales/ca/messages.po
```diff
@@ -298,7 +298,7 @@ msgstr "Afegeix text alternatiu"
 
 #: src/view/com/composer/GifAltText.tsx:175
 msgid "Add ALT text"
-msgstr ""
+msgstr "Afegeix ALT"
```
surfdude29, line 301: The English string was a typo that was fixed in #3934 to become `Add alt text` (the message catalogs haven't been updated with it yet). So I suggest you might want to change this to follow the translation below, something like `Afegeix text alternatiu`?

src/locale/locales/ca/messages.po
```diff
@@ -5724,7 +5724,7 @@ msgstr "Aquesta apel·lació s'enviarà a <0>{0}</0>."
 
 #: src/screens/Messages/Conversation/MessageListError.tsx:26
 msgid "This chat was disconnected due to a network error."
-msgstr ""
+msgstr Aquest xat s'ha desconnectat degut a un problema de xarxa.""
```
surfdude29, line 5727: There's a `"` missing at the beginning here and an extra `"` at the end.

ivanbea: Thanks! your corrections are so useful!

pfrazee (approved):
To make sure this gets into the release, I'm considering this sufficiently reviewed. Thanks all!

PR #3985, opened 2024-05-13
Delete the entire temporary directory instead of just the temp file, also use `cacheDirectory` over `documentDirectory`

haileyok:
Matthieu pointed out the temp directory itself might not get deleted, so let's delete the whole thing. Additionally though, let's use `cacheDirectory` just incase something does happen, since the cache directory will be cleared out by the OS eventually regardless of whether we are successful in deletion ourselves.


matthieusieben (approved):
LGTM

PR #3991, opened 2024-05-13
[Reduced Onboarding] Swap in base following

estrattonbailey:
Swaps in our "base" following feed, with no fallback to Discover. Below is what new users will see when they visit the following feed. These buttons take the user to Search and to Feeds, respectively.

CleanShot 2024-05-13 at 13 22 16@2x


haileyok (approved):

pfrazee (approved):
Yeah lgtm

PR #3993, opened 2024-05-13
[🐴] Message drafts

mozzius:
Saves message drafts, just in memory for now. Helpful because we unmount the text input aggressively.

### Test plan

Type a message then background the app and come back. Observes your message returns.
Type out multiple drafts to different people and see they're all preserved
Send a message and check that the draft is cleared


src/state/messages/message-drafts.tsx
```diff
@@ -0,0 +1,84 @@
...
+      if (currentConvoId) {
+        dispatch({
+          type: 'set',
+          convoId: currentConvoId,
+          draft: messageRef.current,
+        })
+      }
+    }
+  }, [currentConvoId, dispatch])
+}
```
mozzius, line 54: This is the sketchy part. Can't think of another way to stop `useEffect` firing every time the message changes (other than a completely different approach like event listeners)

estrattonbailey (approved):

src/state/messages/message-drafts.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export function useSaveMessageDraft(message: string) {
+  const {currentConvoId} = useCurrentConvoId()
+  const {dispatch} = useMessageDraftsContext()
+  const messageRef = useRef(message)
+  messageRef.current = message
+
+  useEffect(() => {
+    return () => {
+      if (currentConvoId) {
```
haileyok, line 44: Note that this won't save a draft if you go Chat -> The other person's profile -> Close the app. Not _huge_ but might cause confusion at some point.

PR #3999, opened 2024-05-13
[Reduced Onboarding] Add avatar metric

estrattonbailey:
Stacked on #3997 

Thought it'd be nice to know if the user created an avi, uploaded one, or took no action and used the default.

### DIFF
--- a/src/screens/Onboarding/StepFinished.tsx
+++ b/src/screens/Onboarding/StepFinished.tsx
@@ -160,7 +160,16 @@ export function StepFinished() {
     onboardDispatch({type: 'finish'})
     track('OnboardingV2:StepFinished:End')
     track('OnboardingV2:Complete')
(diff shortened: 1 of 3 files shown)


src/screens/Onboarding/StepFinished.tsx
```diff
@@ -160,7 +160,16 @@ export function StepFinished() {
...
+
+    let avatarResult = 'default'
+    if (profileStepResults.isCreatedAvatar) {
+      avatarResult = 'created'
+    } else if (profileStepResults.image) {
+      avatarResult = 'uploaded'
+    }
+    logEvent('onboarding:finished:nextPressed', {
+      avatarResult,
+    })
```
haileyok, line 172: we should only call `logEvent` if the gate is true I think, otherwise everyone including people who are in old onboarding will be logging this.

PR #4002, opened 2024-05-13
Tweak feed card styles

estrattonbailey:
Small tweaks to #3998 

No real changes to desktop, so just mobile screenshots.

This PR adjusts `FeedSourceCard` to truncate the display name and `Feed by` text to 1 line. This would affect rendering in other locations in the app.

However, `displayName` for a feed gen is limited to 24 graphemes, so this is really only an issue on small screens.

For the `Feed by` text, this could be much longer, but as you can see with a long handle, the wrapping is pretty bad. When clicking into a feed detail screen, this text is shown in full.

### After:
CleanShot 2024-05-13 at 17 17 15@2x
CleanShot 2024-05-13 at 17 22 55@2x
CleanShot 2024-05-13 at 17 23 22@2x

### Before:
CleanShot 2024-05-13 at 17 25 15@2x
CleanShot 2024-05-13 at 17 25 00@2x

### DIFF
--- a/src/view/com/feeds/FeedSourceCard.tsx
+++ b/src/view/com/feeds/FeedSourceCard.tsx
@@ -211,10 +211,10 @@ export function FeedSourceCardLoaded({
             <UserAvatar type="algo" size={36} avatar={feed.avatar} />
           </View>
           <View style={[styles.headerTextContainer]}>
-            <Text style={[pal.text, s.bold]} numberOfLines={3}>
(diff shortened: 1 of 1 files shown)


mozzius (approved):

pfrazee (approved):

estrattonbailey:
Just adding a test: top feed is at max length

CleanShot 2024-05-13 at 17 48 37@2x

PR #4005, opened 2024-05-14
Add push notification extensions

haileyok:
Test that it plays a sound when you're backgrounded
Test it doesn't play a sound when you're foregrounded
Test that other notification types are not changed
Test that the toggle (temporarily a little check box in the messages settings screen) turns off sounds


app.config.js
```diff
@@ -110,7 +110,7 @@ module.exports = function (config) {
             {
               NSPrivacyAccessedAPIType:
                 'NSPrivacyAccessedAPICategoryUserDefaults',
-              NSPrivacyAccessedAPITypeReasons: ['CA92.1'],
+              NSPrivacyAccessedAPITypeReasons: ['CA92.1', '1C8F.1'],
```
haileyok, line 113: Adding a new privacy type per https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401

The extension - and now the app through the module - access app group user defaults, which needs this additional privacy reason.

modules/BlueskyNSE/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    if reason == "chat-message" {
+      mutateWithChatMessage(bestAttempt)
+    }
+    
+    // The badge should always be incremented when in the background
+    mutateWithBadge(bestAttempt)
+    
+    contentHandler(bestAttempt)
+  }
```
haileyok, line 26: Incoming messages that have a `chat-message` reason should have a sound added to them. All of them should get a badge increment

modules/BlueskyNSE/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
...
+  
+  override func serviceExtensionTimeWillExpire() {
+    // If for some reason the alloted time expires, we don't actually want to display a notification
+  }
```
haileyok, line 30: When we add avatars, we'll want to use this block here to display the notification after roughly 30 seconds _if_ the network request fails or something else goes wrong. For now, there isn't a need to use this.

modules/expo-background-notification-handler/android/src/main/java/expo/modules/backgroundnotificationhandler/BackgroundNotificationHandler.kt
```diff
@@ -0,0 +1,31 @@
...
+      // We'll let expo-notifications handle the notification if the app is foregrounded
+      return
+    }
+
+    if (remoteMessage.data.getValue("reason") == "chat-message") {
+      mutateWithChatMessage(remoteMessage)
+    }
+
+    notifInterface.showMessage(remoteMessage)
+  }
```
haileyok, line 21: On Android, we already have an incrementing badge, so we do not need to do any mutations for that. The only thing we will mutate is the `channelId` which is based on the message's reason.

modules/expo-background-notification-handler/android/src/main/java/expo/modules/backgroundnotificationhandler/BackgroundNotificationHandlerInterface.kt
```diff
@@ -0,0 +1,7 @@
+package expo.modules.backgroundnotificationhandler
+
+import com.google.firebase.messaging.RemoteMessage
+
+interface BackgroundNotificationHandlerInterface {
+  fun showMessage(remoteMessage: RemoteMessage)
+}
```
haileyok: Just a stub so we can access `showMessage()` from `expo-notifications`

modules/expo-background-notification-handler/ios/ExpoBackgroundNotificationHandlerModule.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+    
+    AsyncFunction("removeManyFromStringArrayAsync") { (forKey: String, strings: [String]) in
+      if var curr = userDefaults?.stringArray(forKey: forKey) {
+        strings.forEach { s in
+          curr.removeAll(where: { $0 == s })
+        }
+        userDefaults?.setValue(curr, forKey: forKey)
+      }
+    }
```
haileyok, line 114: We likely won't need all of these options for updating an array, but I just made some different ones I think we could run into. No harm no foul. (All we use right now is `setBool`, this would be for like mute words/muted threads/etc.

modules/expo-background-notification-handler/src/ExpoBackgroundNotificationHandlerModule.web.ts
```diff
@@ -0,0 +1,27 @@
...
+  getStringArrayAsync: async (_: string) => {
+    return []
+  },
+  setBoolAsync: async (_: string, __: boolean) => {},
+  setStringAsync: async (_: string, __: string) => {},
+  setStringArrayAsync: async (_: string, __: string[]) => {},
+  addToStringArrayAsync: async (_: string, __: string) => {},
+  removeFromStringArrayAsync: async (_: string, __: string) => {},
+  addManyToStringArrayAsync: async (_: string, __: string[]) => {},
+  removeManyFromStringArrayAsync: async (_: string, __: string[]) => {},
```
haileyok, line 26: A web stub. None of these preferences are used on web, so the functions are all noop

patches/expo-notifications+0.27.6.patch
```diff
@@ -0,0 +1,197 @@
...
++      getBackgroundTasks().forEach {
++        it.scheduleJob(RemoteMessageSerializer.toBundle(remoteMessage))
++      }
+     }
+   }
+
++  override fun showMessage(remoteMessage: RemoteMessage) {
++    NotificationsService.receive(context, createNotification(remoteMessage))
++  }
++
```
haileyok, line 194: This part of the patch lets us handle incoming messages while the app is backgrounded or killed. We could have done it all in the patch but that would have been messy.

Good news: This code hasn't been updated in over 3 years, and probably won't any time soon 🤞

haileyok: Everything else in this patch (above the selection) is just adding `channelId` as something we can get from `data`, rather than relying on the OS to get it from `notification`. We don't get access to `data` when the app is backgrounded if there's a `notification` object on the payload, so we must get the channel from `data` unfortunately.

src/App.web.tsx
```diff
@@ -92,9 +93,11 @@ function InnerApp() {
...
                           <UnreadNotifsProvider>
-                            <SafeAreaProvider>
-                              <Shell />
-                            </SafeAreaProvider>
+                            <BackgroundNotificationPreferencesProvider>
+                              <SafeAreaProvider>
+                                <Shell />
+                              </SafeAreaProvider>
+                            </BackgroundNotificationPreferencesProvider>
                           </UnreadNotifsProvider>
```
haileyok, line 101: All the underlying functions are no-op on web of course, but having the context available prevents us from having to `if (isNative)` any time we update a setting (again, think when we have muted threads as an option).

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -60,6 +61,28 @@ export function useNotificationsHandler() {
...
+
+    Notifications.setNotificationChannelAsync('chat-messages-muted', {
+      name: 'Chat - Muted',
+      importance: Notifications.AndroidImportance.MAX,
+      sound: null,
+      showBadge: true,
+      vibrationPattern: [250],
+      lockscreenVisibility: Notifications.AndroidNotificationVisibility.PRIVATE,
+    })
+  }, [])
```
haileyok, line 84: Creating two new notification channels for the custom sounds. We'll have to create others for stuff like "Likes" and "Replies", etc. if we want to have more control over those in the future (regarding sound playback).

The user also gets some good configuration on Android through the different channels so we _should_ create separate ones for different notification reasons. Right now they all get grouped into one bucket.

src/screens/Messages/Settings.tsx
```diff
@@ -65,6 +68,18 @@ export function MessagesSettingsScreen({}: Props) {
...
+          name="a"
+          label="Click me"
+          value={preferences.playSoundChat}
+          onChange={() => {
+            setPref('playSoundChat', !preferences.playSoundChat)
+          }}>
+          <Toggle.Checkbox />
+          <Toggle.LabelText>Notification Sounds</Toggle.LabelText>
+        </Toggle.Item>
+      </View>
```
haileyok, line 82: Just for testing, will re-style later.

src/App.web.tsx
estrattonbailey: Good thinking

estrattonbailey (approved):

patches/expo-notifications+0.27.6.patch
Gcunhaa: Hey @haileyok , why would you want to handle incoming message instead of letting expo-notification handle it ?

haileyok: This lets us do things like this https://github.com/bluesky-social/social-app/blob/main/modules/expo-background-notification-handler/android/src/main/java/expo/modules/backgroundnotificationhandler/BackgroundNotificationHandler.kt

PR #4012, opened 2024-05-14
[🐴] Go back on account switches

haileyok:


src/screens/Messages/Conversation/index.tsx
```diff
@@ -31,10 +32,31 @@ type Props = NativeStackScreenProps<
   CommonNavigatorParams,
   'MessagesConversation'
 >
+
+let prevDid: string | undefined
```
mozzius, line 36: A bit odd having state outside of React? Shouldn't this be in a ref?

haileyok: It can't be because the whole tree is gettin nuked when see switch accounts. Ran into the same problem with notifications as well.

mozzius:
Talked to Hailey about this, this is fine-ish but it would be much better to have a `useWillSwitchAccounts(callback)` hook to run this logic

haileyok:
Yea this would be useful in some other notification logic too.

PR #4014, opened 2024-05-14
[🐴] show deleted account for `missing.invalid`

haileyok:
Screenshot 2024-05-14 at 11 11 34 AM
Screenshot 2024-05-14 at 11 11 38 AM

### DIFF
--- a/src/screens/Messages/List/ChatListItem.tsx
+++ b/src/screens/Messages/List/ChatListItem.tsx
@@ -29,6 +29,13 @@ export function ChatListItem({
   const {currentAccount} = useSession()
   const menuControl = useMenuControl()
   const {gtMobile} = useBreakpoints()
+  const otherUser = convo.members.find(
+    member => member.did !== currentAccount?.did,
+  )
+  const isDeletedAccount = otherUser?.handle === 'missing.invalid'
+  const displayName = isDeletedAccount
+    ? 'Deleted Account'
+    : otherUser?.displayName ?? otherUser?.handle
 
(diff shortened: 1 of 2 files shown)


src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -29,6 +29,13 @@ export function ChatListItem({
   const {currentAccount} = useSession()
   const menuControl = useMenuControl()
   const {gtMobile} = useBreakpoints()
+  const otherUser = convo.members.find(
+    member => member.did !== currentAccount?.did,
+  )
+  const isDeletedAccount = otherUser?.handle === 'missing.invalid'
+  const displayName = isDeletedAccount
+    ? 'Deleted Account'
+    : otherUser?.displayName ?? otherUser?.handle
```
mozzius, line 38: Should be
```js
  const displayName = isDeletedAccount
    ? 'Deleted Account'
    : otherUser?.displayName || otherUser?.handle
```

mozzius (changes requested):
Minor display name logic problem

src/screens/Messages/List/ChatListItem.tsx
mozzius: How do you get it to do the "accept suggested changes" button?

haileyok: Why bluesky why

mozzius (approved):
le sigh

PR #4018, opened 2024-05-14
Support routes in query string (`/?/`) to allow web app to be completely statically hosted.

elfprince13:
Draft of an idea discussed with @haileyok in a [thread](https://bsky.app/profile/haileyok.com/post/3kshiebv3zk26) this morning.

This lowers the barrier to entry for hosting forks of the app as no paths other than `/` are ever requested from the server, ergo no code needs to run on the server to support resolving deep links. Instead, the routes are encoded as query strings. For example `/profile/username.bsky.social/` can be accessed as `/?/profile/username.bsky.social`. This means, for example, you could host a forked bsky web client with GitHub Pages, or potentially even serve it over IPFS or other non-HTTP protocols.

This still supports parsing old-style links; however, all links generated in the UI should be in the new style. A slightly more complex solution would be to detect old vs new style links and switch the generation appropriately to allow both versions to exist in parallel in perpetuity. Creating as a draft PR for discussion as I'm not sure what the considerations are here from the perspective of the official client.

[edit]

you can test how this works with:

```
npx expo export:web
npx serve web-build --single
```


gaearon:
I think this isn't something we'd be able to commit to supporting so it might be easier to maintain as a fork. WDYT?

elfprince13:
The goal is to lower barrier to entry for experimenting in forks, so in that sense I see the attraction, but main downside of maintaining as a fork is it results in more time consumed trying to keep feature parity with upstream.

Is the hurdle here that you don't want to commit to pure client-side rendering in the future? Or just that the current implementation is breaking abstraction in a way that feels like it would be hard to maintain due to parallel code paths?

PR #4020, opened 2024-05-14
[🐴] Blocks/mutes and revalidation

estrattonbailey:
This PR introduces a method on the `Convo` agent called `revalidateConvo` that re-fetches the `ConvoView` and updates state for use in the React tree. It also adds an array of `blocks` which are basic profile views, inclusive of `viewer` state, if any members of the chat are blocking each other.


estrattonbailey:
Better solution in #4022 and #4024

PR #4021, opened 2024-05-14
[🐴] Show if user can be messaged in new chat search

mozzius:
Screenshot 2024-05-14 at 21 53 22

Also includes a bunch of other minor tweaks:

- Remove top border from loading state (here and in gif search)
- Show highlight if focused (convo list)
- Remove self from new chat search


src/components/dms/NewChat.tsx
```diff
@@ -253,3 +275,17 @@ function SearchablePeopleList({
...
+      return true
+    } else if (
+      profile.associated.chat.allowIncoming === 'following' &&
+      profile.viewer?.followedBy
+    ) {
+      return true
+    }
+  }
+  return false
+}
```
estrattonbailey, line 291: Something for the morning: will all users get the default declaration, or do we need to default to true in the client for users without it?

mozzius: Reworked this logic to default to following

estrattonbailey (approved):

PR #4027, opened 2024-05-15
[🐴] Only scroll down one "screen" in height when foregrounding

haileyok:


mozzius (approved):
Makes a lot more sense in combination with the other PR, I'm a fan now

PR #4032, opened 2024-05-15
[🐴] Better retry styling

estrattonbailey:
Instead of handling retries as a separate `ConvoItem`, I've added a prop to `pending-message` types so that we can style it within `MessageItem`.


mozzius (approved):
Good stuff

PR #4036, opened 2024-05-15
[🐴] Error recovery

estrattonbailey:
This PR started out as specifically block handling, but turned out to be easier to refactor everything at once.

**So included here is:**

Handling for when the user on the other end of the line blocks me and I don't know yet, so I send a message. It checks for block errors, and handles those as `unrecoverable` pending errors, whereas a network error (which I now check for) is `recoverable`. It also invalidates profile queries used elsewhere for shadowed optimistic block state using an emitter. Not super proud of this but it works and I think it's clear.

Better handling for network errors. Now if combinations of errors occur, the errors won't shift around, and resuming from one error will attempt to resume from other applicable errors.

It now attempts to recover when returning from background.

I also cleaned up the error components to be a little less invasive, since hopefully we'll clean stuff up before the user has a change to interact with them.

**Not in this PR**
The case of a user blocking me, I get into a blocked state, and then they unblock me, is not handled here. I expect that to be a rare occurrence, and it's only partially recoverable via the firehose e.g. if the blocker messages me again. Otherwise we'd need to poll for block state, and I don't think that's necessary atm. A simple refresh or pop/push back to the chat will reset it back to zero and the users can message again.

**Testing:**
- turn off network and send message, load history, etc
- turn network back on and recover
  - via interaction
  - via background->foreground
- block yourself from test account and then try to send a message, see block state in header

CleanShot 2024-05-16 at 12 41 01@2x
CleanShot 2024-05-16 at 12 41 11@2x


src/state/messages/convo/index.tsx
```diff
@@ -78,6 +82,23 @@ export function ConvoProvider({
...
+              queryKey: createProfileQueryKey(did),
+            })
+          }
+          queryClient.invalidateQueries({
+            queryKey: ListConvosQueryKey,
+          })
+        }
+      }
+    })
+  }, [convo, queryClient])
```
haileyok, line 100: Could we init `Convo` with `QueryClient` instead of the emitter for this?

haileyok (approved):

haileyok:
Slightly weird state where we have two error messages, but I think we can tweak these separately.

Screenshot 2024-05-16 at 11 20 06 AM

src/state/messages/convo/index.tsx
estrattonbailey: Since the life time of a queryClient should match the lifetime of a user, and therefore a `Convo`, yeah I think we could for now. But if in the future those lifetimes change, this would allow it to keep working.

Felt no-less stable to do it this way, so figured it's fine for now. What do you think?

PR #4038, opened 2024-05-15
[🐴] Remove keyboard controller lib

haileyok:
## Why

The keyboard controller library we were using for this was a little glitchy and was causing some side effects on Android that would have affected the composer 😮‍💨

What I've done here is taken a look at using Reanimated's `useAnimatedKeyboard` to achieve the same effect.

`useAnimatedKeyboard` has a shared value containing the current height of the keyboard at any given point during the animation. We can use this value to do two things:

1. Move a "sticky input" - an input with relative positioning and with a bottom that's always the height of the keyboard
2. Adjust the size of a `ListFooterComponent`, which will provide the needed "offset" and will fire a content size changed event.

The result is something that's actually a little closer to 60FPS than what we have right now, seemingly on both platforms (testing on Android simulator shows slightly lower FPS, but on a physical device I had good results)

## Test Plan

Really just use it. There might be more tweaks we have to make over time, but I think this is a much more solid solution than what we had before.

https://github.com/bluesky-social/social-app/assets/153161762/ce1e1161-9b51-4f29-898c-4c1c334733ac

https://github.com/bluesky-social/social-app/assets/153161762/c33d2af7-7719-488e-ada4-6e82913e6a0a

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,6 @@
     "react-native-get-random-values": "~1.11.0",
     "react-native-image-crop-picker": "^0.38.1",
     "react-native-ios-context-menu": "^1.15.3",
-    "react-native-keyboard-controller": "^1.11.7",
     "react-native-pager-view": "6.2.3",
     "react-native-picker-select": "^8.1.0",
     "react-native-progress": "bluesky-social/react-native-progress",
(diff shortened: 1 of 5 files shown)


mozzius (approved):
Incredible work. Lets get this on real devices to continue testing, but in the simulator it seems to behave wayyy better than before. I think you've managed to beat Discord's behaviour on Android 😈

PR #4039, opened 2024-05-16
Implement prefersReducedMotion on native

mozzius:
Currently, we only get the `prefersReducedMotion` value on web. Luckily, Reanimated has a `useReducedMotion` hook, but we need the value not in a hook. This PR uses the (internal) `isReducedMotion` function from Reanimated since it does exactly the same thing on web, and gets the value as it is at app startup on native. Check the simple implementation here

https://github.com/software-mansion/react-native-reanimated/blob/44d55dd3b6b92759fa820538d4f7c420bc8c2fa0/src/reanimated2/PlatformChecker.ts#L49

Note that the implementation of `useReducedMotion` is ultra simple and we could just call it here as not to rely on internal functions, but I think it's probably unsafe to break the rule of hooks here in case they change the implementation.

https://github.com/software-mansion/react-native-reanimated/blob/44d55dd3b6b92759fa820538d4f7c420bc8c2fa0/src/reanimated2/hook/useReducedMotion.ts#L4

### DIFF
--- a/patches/react-native-reanimated+3.6.0.patch
+++ b/patches/react-native-reanimated+3.6.0.patch
@@ -1,0 +1,30 @@
+diff --git a/node_modules/react-native-reanimated/lib/module/reanimated2/index.js b/node_modules/react-native-reanimated/lib/module/reanimated2/index.js
+index 91e49f4..c10d3fc 100644
+--- a/node_modules/react-native-reanimated/lib/module/reanimated2/index.js
++++ b/node_modules/react-native-reanimated/lib/module/reanimated2/index.js
+@@ -45,4 +45,5 @@ export { getUseOfValueInStyleWarning } from './pluginUtils';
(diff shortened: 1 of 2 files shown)


haileyok:
Oh holy shit, was wanting to do this before but realized it wasn't available in the RN function. This is great.

haileyok:
This is actually a global on native too, so if you wanted to you could drop the patch and access it directly through `global._REANIMATED_IS_REDUCED_MOTION` the same way that `_IS_FABRIC` is accessible that way.

https://github.com/software-mansion/react-native-reanimated/blob/44d55dd3b6b92759fa820538d4f7c420bc8c2fa0/Common/cpp/ReanimatedRuntime/RNRuntimeDecorator.cpp#L41
https://github.com/software-mansion/react-native-reanimated/blob/44d55dd3b6b92759fa820538d4f7c420bc8c2fa0/src/reanimated2/globals.d.ts#L24

haileyok (approved):

mozzius:
Yep, I think the patch is a good canary for if their internal implementation changes

PR #4040, opened 2024-05-16
[🐴] Decrement app badge when opening unread chat

haileyok:
- Decrement the badge when opening a chat from the messages list
- Decrement the badge when opening a chat from a push notification
- Increment the badge every time you receive a notification for a chat
	- EXCEPT for if it is the chat you are currently looking at

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -6,7 +6,6 @@ import {useQueryClient} from '@tanstack/react-query'
 import {logger} from '#/logger'
 import {track} from 'lib/analytics/analytics'
 import {useAccountSwitcher} from 'lib/hooks/useAccountSwitcher'
-import {decrementBadgeCount} from 'lib/notifications/notifications'
 import {NavigationProp} from 'lib/routes/types'
 import {logEvent} from 'lib/statsig/statsig'
 import {isAndroid} from 'platform/detection'
@@ -89,7 +88,6 @@ export function useNotificationsHandler() {
       if (!payload) return
 
       if (payload.reason === 'chat-message') {
-        decrementBadgeCount(1)
         if (payload.recipientDid !== currentAccount?.did && !storedPayload) {
           storedPayload = payload
           closeAllActiveElements()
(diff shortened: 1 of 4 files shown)


src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -103,11 +104,16 @@ function ChatListItemReady({
     setShowActions(true)
   }, [])
 
-  const onPress = useCallback(() => {
+  const onPress = useCallback(async () => {
+    if (convo.unreadCount > 0) {
+      // We only increment the badge once per conversation, so even if there are multiple unreads
+      // we only want to decrement by one.
+      decrementBadgeCount(1)
```
estrattonbailey, line 111: We're doing `markAsRead` within the `ConvoProvider`, could be a good spot for this too

estrattonbailey: Yeah pry good to do there in case you foreground the app directly to the convo page

haileyok: Ahhh okay perfect let me adjust to that, wasn't sure if we did it or the server did it for us.

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -169,10 +169,11 @@ export function useNotificationsHandler() {
           payload.reason === 'chat-message' &&
           payload.recipientDid === currentAccount?.did
         ) {
+          const isCurrentConvo = payload.convoId === currentConvoId
           return {
-            shouldShowAlert: payload.convoId !== currentConvoId,
+            shouldShowAlert: isCurrentConvo,
             shouldPlaySound: false,
-            shouldSetBadge: false,
+            shouldSetBadge: isCurrentConvo,
```
estrattonbailey, line 176: Should these be `!` inverted?

estrattonbailey (approved):

src/lib/hooks/useNotificationHandler.ts
haileyok: no that doesn't make sense!!! (thanks good catch)

PR #4043, opened 2024-05-16
[🐴] Pill for additional unreads when coming from background

haileyok:
Stacked on https://github.com/bluesky-social/social-app/pull/4027

Needs some styling work, but not sure how to make it look right now.

Go to about 2/3 of the way through the video to see effect when foregrounding.

https://github.com/bluesky-social/social-app/assets/153161762/e70500f1-281a-407a-89ac-47214b82c304

### DIFF
--- a/src/components/dms/NewMessagesPill.tsx
+++ b/src/components/dms/NewMessagesPill.tsx
@@ -1,48 +1,14 @@
 import React from 'react'
 import {View} from 'react-native'
-import Animated, {withTiming} from 'react-native-reanimated'
+import Animated from 'react-native-reanimated'
 import {Trans} from '@lingui/macro'
 
+import {
+  ScaleAndFadeIn,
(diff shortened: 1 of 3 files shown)


mozzius:
Looks super nice, but I'd expect something to happen when you click on it? Not sure if I'd expect it to scroll to bottom or scroll to the oldest unread message though

mozzius (approved):
Ok I understand more about how this is supposed to be working. This looks good (great in fact) but we should also consider showing this when any new message comes in and you're not scrolled to the bottom

haileyok:
Nice, as long as we're okay with the look then we can definitely do that!

haileyok:
Also I'm torn on the "should it scroll" too. On one hand it would let you get to the bottom quickly, on the other you'd end up missing the new content 🤔 

Still think I lean more toward making it scroll down

Issue #4054, opened 2024-05-16
Labels *and* label reveal being rendered together


Bossett:
**Describe the bug**

Labels are being rendered alongside 'X labels has been placed...' text (which should be 'X labels **have** been placed...' while I'm here):

![image](https://github.com/bluesky-social/social-app/assets/114112113/06f9bae1-bb39-4490-875d-0950c73c64ef)

**To Reproduce**

1. View your profile while it has been labelled

**Expected behavior**

Either show the labels, or show the reveal.

**Details**

 - Platform: Windows Chrome
 - Platform version: 124
 - App version: 1.83.0

haileyok:
Pretty sure this only applies to your own account, but can you see if it's happening for posts too? It shouldn't but want to make sure before I merge in this fix. Thank you!!!

Bossett:
I think that's right, I can't see it on posts

PR #4057, opened 2024-05-16
don't show individual labels on own profile, only "have been placed..."

haileyok:
## Why

Fixes https://github.com/bluesky-social/social-app/issues/4054

I think this showed up as a result of this: https://github.com/bluesky-social/atproto/commit/0b6f6f8636e6b9212568bdb508f8e6f61102b3e7#diff-1003892846e81d97c6492ad916b5379e6dac4e3b25c923d0a5007ec3eac08260

I think we only want to display individual labels on accounts which are _not_ our own, since on our own profile we show the appeal dialog.

## Test Plan

There's a labeler that has a applied labels to a lot of our accounts here: https://bsky.app/profile/yardcrow.com

Subscribe to that labeler and see if you have a label on your account. If you do, great. Test:

- On production, you should see both a list of labels _and_ the "placed on this account..." label
- On this branch, you should _only_ see the "placed on this account..." label
- On my profile https://bsky.app/profile/haileyok.com you should see _only_ the individual labels

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -84,9 +84,10 @@ let ProfileHeaderShell = ({
       <View
         style={[a.px_lg, a.pb_sm]}
         pointerEvents={isIOS ? 'auto' : 'box-none'}>
-        <ProfileHeaderAlerts moderation={moderation} />
-        {isMe && (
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Yep lgtm

PR #4062, opened 2024-05-16
[🐴] NUX

mozzius:
Appears based on the presence of the chat declaration record. Sets it to `following` as soon as it appears, and lets you change it in the dialog.

Placed in the messages list, so you'll see it the first time you click on the Chat tab

Simulator Screenshot - iPhone 15 - 2024-05-17 at 19 04 27


haileyok:
Unlikely this is related to this PR, but the first option has this highlighting by default on chrome

Screenshot 2024-05-16 at 2 27 31 PM

estrattonbailey (approved):

PR #4068, opened 2024-05-17
[🐴] Block Info

haileyok:
Stacked on https://github.com/bluesky-social/social-app/pull/4075

Honestly I don't really have enough context from other apps to know how this should look, but the functionality is there now at least so feel free to nit!

Screenshot 2024-05-17 at 12 32 25 AM


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -289,15 +313,110 @@ export function MessagesList({
...
+
+  return (
+    <View style={[hasMessages && a.py_md, a.gap_lg]}>
+      {hasMessages && (
+        <>
+          <Divider />
+        </>
+      )}
+      <Text style={[a.text_md, a.font_bold, a.text_center]}>
+        {_(msg`You have blocked this user`)}
```
estrattonbailey, line 369: Should also conditionally say "this user has blocked you"

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -289,15 +313,110 @@ export function MessagesList({
...
+      <View style={[a.flex_row, a.justify_between, a.gap_lg, a.px_md]}>
+        <Button
+          label={_(msg`Delete`)}
+          color="secondary"
+          variant="solid"
+          size="small"
+          style={[a.flex_1]}
+          onPress={leaveConvoControl.open}>
+          <ButtonText style={{color: t.palette.negative_500}}>
+            <Trans>Delete</Trans>
```
estrattonbailey, line 381: Should we keep this consistent and say "Leave chat"?

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -289,15 +313,110 @@ export function MessagesList({
...
+          </ButtonText>
+        </Button>
+        <Button
+          label={_(msg`Report`)}
+          color="secondary"
+          variant="solid"
+          size="small"
+          style={[a.flex_1]}
+          onPress={reportDialogControl.open}>
```
estrattonbailey, line 390: This needs to open the info prompt and direct the user to report a message

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -289,15 +313,110 @@ export function MessagesList({
...
+  const {_} = useLingui()
+  const [__, queueUnblock] = useProfileBlockMutationQueue(recipient)
+
+  const leaveConvoControl = useDialogControl()
+  const reportDialogControl = useDialogControl()
+
+  if (!isBlocking && !isBlockedBy) return null
+
+  return (
+    <View style={[hasMessages && a.py_md, a.gap_lg]}>
```
estrattonbailey, line 362: This could use just a little more padding on the bottom, little tight. Can be a size more on bottom than top if needed.

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -289,15 +313,110 @@ export function MessagesList({
...
+  isBlockedBy: boolean
+}) {
+  const t = useTheme()
+  const {_} = useLingui()
+  const [__, queueUnblock] = useProfileBlockMutationQueue(recipient)
+
+  const leaveConvoControl = useDialogControl()
+  const reportDialogControl = useDialogControl()
+
+  if (!isBlocking && !isBlockedBy) return null
```
estrattonbailey, line 359: Kinda odd to have this here? Could invert the logic above to `isBlocking || isBlockedBy`

src/screens/Messages/Conversation/index.tsx
```diff
@@ -71,15 +71,21 @@ function Inner() {
   const convoState = useConvo()
   const {_} = useLingui()
 
+  const moderationOpts = useModerationOpts()
+  const {data: recipient} = useProfileQuery({
+    did: convoState.recipients?.[0].did,
+  })
```
estrattonbailey, line 77: Nice

src/screens/Messages/Conversation/MessagesList.tsx
haileyok: Ah, hmm. I guess that makes sense since we have the same in the other dialog for "Report Conversation" and since the other option to "Delete" is performing the action on the conversation too.

Should we just add an option to the info prompt to report the user instead?

src/screens/Messages/Conversation/MessagesList.tsx
haileyok: Sure. I think I'll need to change the layout of the buttons a bit on native for this but like the idea.

src/screens/Messages/Conversation/MessagesList.tsx
haileyok: Ah no, not necessary to have it here at all 👍 missed it when moving the check up

estrattonbailey (approved):
Great work on this!

PR #4075, opened 2024-05-17
[🐴] Additional tweaks to the message list

haileyok:


src/screens/Messages/Conversation/index.tsx
```diff
@@ -71,23 +71,15 @@ function Inner() {
...
-    if (
-      !hasInitiallyRendered &&
-      isConvoActive(convoState) &&
-      !convoState.isFetchingHistory
-    ) {
-      setTimeout(() => {
-        setHasInitiallyRendered(true)
-      }, 15)
-    }
-  }, [convoState, hasInitiallyRendered])
```
haileyok, line 90: This was a nasty hack, time to rm

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -88,22 +92,9 @@ export function MessagesList() {
...
-    (offset: number, animated: boolean) => {
-      if (isWeb) {
-        flatListRef.current?.scrollToOffset({offset, animated})
-      } else {
-        runOnUI(scrollTo)(flatListRef, 0, offset, animated)
-      }
-    },
-    [flatListRef],
-  )
-
```
haileyok, line 106: In a galaxy far far away, developers have a Javascript ecosystem that interacts with native code fully synchronously, and they live very happily.

Alas, that is not our galaxy. There's some layouts that are not actually ready yet when we call `scrollTo` right now, particularly around sending a new message and expecting the view to scroll up. We've seen over the past week that the old way of doing things was solid, so let's return to that.

We _will_ still use `scrollTo` for keyboard animation scrolling, because its 🔥🥵🔥 and doesn't cause problems.

estrattonbailey (approved):

PR #4082, opened 2024-05-17
[🐴] don't include blocked convos in unread count

mozzius:
Unblocked/blocked
CleanShot 2024-05-17 at 14 34 56@2x
CleanShot 2024-05-17 at 14 34 47@2x

### DIFF
--- a/src/state/queries/messages/list-converations.ts
+++ b/src/state/queries/messages/list-converations.ts
@@ -1,5 +1,9 @@
 import {useCallback, useMemo} from 'react'
-import {ChatBskyConvoDefs, ChatBskyConvoListConvos} from '@atproto/api'
+import {
+  ChatBskyConvoDefs,
+  ChatBskyConvoListConvos,
+  moderateProfile,
+} from '@atproto/api'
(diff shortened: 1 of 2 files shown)


src/state/queries/messages/list-converations.ts
```diff
@@ -45,7 +46,12 @@ export function useUnreadMessageCount() {
       .flatMap(page => page.convos)
       .filter(convo => convo.id !== currentConvoId)
       .reduce((acc, convo) => {
-        return acc + (!convo.muted && convo.unreadCount > 0 ? 1 : 0)
+        const otherMember = convo.members.find(
+          member => member.did !== currentAccount?.did,
+        )
+        const blocked =
+          otherMember?.viewer?.blockedBy || otherMember?.viewer?.blocking
```
estrattonbailey, line 53: Should also check for `blockedByList` here. I think that's it, but peep the types for `ViewerState` just in case

estrattonbailey (approved):

estrattonbailey (approved):

PR #4088, opened 2024-05-17
Disable badge incrementing for DMs

haileyok:
## Why

Discussed this a bit, it will be difficult to handle this properly for DMs, so let's only increment for other notifications for now.

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -169,11 +169,10 @@ export function useNotificationsHandler() {
           payload.reason === 'chat-message' &&
           payload.recipientDid === currentAccount?.did
         ) {
-          const isCurrentConvo = payload.convoId === currentConvoId
           return {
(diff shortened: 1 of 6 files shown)


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -46,8 +46,9 @@ const DEFAULT_HANDLER_OPTIONS = {
   shouldSetBadge: true,
 }
 
-// This needs to stay outside the hook to persist between account switches
+// These need to stay outside the hook to persist between account switches
 let storedPayload: NotificationPayload | undefined
+let prevDate = 0
```
haileyok, line 51: Let's see if this helps with the notification issue, and if it does I'm going to look into what a more accurate fix is. Went over the docs again for expo-notifications and we're implementing the handler right, so not quite certain what is going on here. This should work though inbetween account switches.

estrattonbailey (approved):

PR #4090, opened 2024-05-18
[🐴] 🤞 This should be the final message list change - Use `dispatchCommand` so we don't need to know the content height

haileyok:
Also made it in this one so that as the message input increases in height, the content scrolls up with it, looks a lot better.

https://github.com/bluesky-social/social-app/assets/153161762/d0409d97-eb94-4096-8750-91d83b628b14


mozzius (approved):
lfggg

PR #4091, opened 2024-05-18
[🐴] Reset screen when pressing a DM notification while already in the conversation screen

haileyok:
When we are already in the conversation screen, we can't just do a `.navigate` - it won't work! Instead, we should use `reset`.


mozzius (approved):
neat

PR #4097, opened 2024-05-18
[🐴] DM button on profile

mozzius:
Calls `getProfileForMembers` to decide whether or not to show the button, since you can still chat someone if they have DMs closed if you already have an ongoing chat (I set mozzius.dev to `none` if you want to test). Means that it pops in after the rest of the profile but it doesn't cause layout shift. Buttons now wrap if needed, rather than flowing behind the avatar. Also includes a cleanup of the button sizes to make them all equal sizes (because before they were all different apparently????)

Opted not to show on labellers, you can still message them through the new chat dialog if you really need to but I think we want to avoid the perception of labellers as normal users.

Screenshot 2024-05-18 at 21 14 03

### DIFF
--- a/src/components/dms/MessageProfileButton.tsx
+++ b/src/components/dms/MessageProfileButton.tsx
@@ -1,44 +1,36 @@
-import React, {useCallback} from 'react'
+import React from 'react'
 import {AppBskyActorDefs} from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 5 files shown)


auroursa:
Sorry I'm not a developer, but I found an area that may could be improved.
After clicking the direct message button and then returning, it doesn't go back to the personal profile but to the direct messages list. To be honest, this is a bit confusing.

estrattonbailey (commented):
Works great! One last thing

estrattonbailey (approved):
LGTM

PR #4098, opened 2024-05-18
Native translation expo module

mozzius:
image


haileyok:
Let a little proposal here: https://github.com/bluesky-social/social-app/pull/4110.

https://github.com/bluesky-social/social-app/assets/153161762/87c65a18-a6ba-448e-a51b-7f8415f29712


We should also:
- Account for the fact the API might not support all the languages Google does. Can make a little list of the 18 or so languages the API does support and make sure it's one of those, probably. Wouldn't work in all contexts though.
- Include this in the translate button in the post dropdown/dialog.

Amazing work figuring this out!

haileyok:
Should make this one a library too 👀

mozzius:
> We should also:
> 
>     * Account for the fact the API might not support all the languages Google does. Can make a little list of the 18 or so languages the API does support and make sure it's one of those, probably. Wouldn't work in all contexts though.
> 
>     * Include this in the translate button in the post dropdown/dialog.

Added both of these. `isLanguageSupported` is just a best guess, but I don't think there's any better way to do it

haileyok (approved):

gaearon (approved):

PR #4101, opened 2024-05-18
Expand the background color to cover the overscroll area on settings screen on iOS

zetavg:
This is just a minor visual tweak to make the settings screen look better on iOS.

| Before | After |
| --- | --- |
|  |  |


gaearon (approved):
looks good, thanks!

PR #4102, opened 2024-05-18
[🐴] Minor nits

haileyok:
Few minor tweaks

- Memo the chat list item
- Set the initialNumToRender min height to something a little smaller (default is 100px but chat items are about 80px)
- Use `scrollTo` with an arbitrary height to get around async scroll jank on Android


mozzius (approved):
_sweeeet_

PR #4105, opened 2024-05-19
Update Spanish Localization

brodieavoult:
I corrected several strings, translated some untranslated ones, and cleaned things up. This is still not complete but I figured I'd push this out to fix the worst of it. I plan on continuing to work on this and to hopefully complete the localization.


pfrazee (approved):
Thanks!

PR #4110, opened 2024-05-20
Make iOS native translation view available globally

haileyok:
This is a little building on this amazing PR https://github.com/bluesky-social/social-app/pull/4098

I think however it might be more useful to make this a global API that can be called from anywhere, rather than a wrapper.

Currently, the wrapper presents the translation modal with a prop that needs to be set via state which is okay, however we probably would probably want to build on that.

The first thought I had was to just add a view function to the native view, like `presentAsync()`, that could be triggered with a press, maybe like so:

```tsx
const ref = React.useRef(null)
const onPress = React.useCallback(() => {
  // Text is a prop
  ref.presentAsync(text)
}, [text])

return (
  <Pressable onPress={onPress}>
  	<NativeView ref={ref} />
  </Pressable>
)
```

This also seems like a fine option - and we can definitely do this if it's preferred - though still I feel like this is an easier API to work with, as it can be dropped in with a single line wherever we want to do translations. Hopefully there's some available option on Android as well, though I have not done any digging on that.

Thank you @mozzius for figuring out the `UIHostingController`!

### DIFF
--- a/modules/expo-bluesky-translate/ios/ExpoBlueskyTranslateView.swift
+++ b/modules/expo-bluesky-translate/ios/ExpoBlueskyTranslateView.swift
@@ -2,36 +2,21 @@ import ExpoModulesCore
 import Foundation
 import SwiftUI
 
-class TranslateViewProps: ObservableObject {
-  @Published var text: String = ""
-  @Published var children: [UIView]?
-  @Published var onClose: EventDispatcher
-  @Published var isPresented: Bool = false {
-    didSet {
-      if !isPresented {
-        self.onClose()
-      }
-    }
-  }
-
-  init(onClose: EventDispatcher) {
-    self.onClose = onClose
-  }
+class TranslateViewState: ObservableObject {
+  static var shared = TranslateViewState()
+  
+  @Published var isPresented = false
+  @Published var text = ""
 }
 
 class ExpoBlueskyTranslateView: ExpoView {
-  let props: TranslateViewProps
-  let onClose = EventDispatcher()
-
-  override func didUpdateReactSubviews() {
-    let subChildren = self.reactSubviews()
-    props.children = subChildren
-  }
-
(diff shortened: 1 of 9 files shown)


modules/expo-bluesky-translate/ios/ExpoBlueskyTranslateView.swift
```diff
@@ -2,36 +2,21 @@ import ExpoModulesCore
...
-        self.onClose()
-      }
-    }
-  }
-
-  init(onClose: EventDispatcher) {
-    self.onClose = onClose
-  }
+class TranslateViewState: ObservableObject {
+  static var shared = TranslateViewState()
```
haileyok, line 6: If we were not targeting iOS 13.4, we could just use `@StateObject` from the top and not worry about making this a singleton. In any case, this results in the same thing.

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -637,32 +637,27 @@ function ExpandedPostDetails({
   const pal = usePalette('default')
   const {_} = useLingui()
   const openLink = useOpenLink()
-  const [presented, setPresented] = React.useState(false)
   const onTranslatePress = React.useCallback(() => {
     if (isNativeTranslateAvailable) {
-      setPresented(true)
+      NativeTranslationModule.presentAsync(text)
```
haileyok, line 642: We can call this from anywhere and have the dialog presented.

src/view/shell/index.tsx
```diff
@@ -93,6 +94,7 @@ function ShellInner() {
           </Drawer>
         </ErrorBoundary>
       </Animated.View>
+      <NativeTranslationView />
```
haileyok, line 97: Just sticking this in with our other dialogs.

mozzius (approved):
Much nicer. Will need to revisit for iPad

PR #4113, opened 2024-05-20
Update French localization

Signez:
Includes all the new strings for the future Chat feature (translated as `Discussions`, even though `chat` is also used in French for this purpose, but is more linked to "group chat" and "chatrooms" ; it also means _cat_ in French 🐈).

Pinging @surfdude29 and @pfrazee :)


surfdude29 (commented):
Excellent work @Signez, looks great to me! 🚀🥖

pfrazee (approved):
Awesome, thanks yall!

PR #4114, opened 2024-05-20
[🐴] Handle deleted accounts, restructure ChatListItem

estrattonbailey:
Review without whitespace: https://github.com/bluesky-social/social-app/pull/4114/files?diff=split&w=1

Deleted accounts delete all messages for both parties, so there's no need to load the Convo screen. This PR disables that link.

This PR also converts to a `Link` for better a11y, and restructures the `ChatListItem` so that the context menu is outside the `Link` for better a11y. It should also fix some alignment issues we had with deleted accounts.

CleanShot 2024-05-20 at 10 25 47@2x
CleanShot 2024-05-20 at 10 26 40@2x
CleanShot 2024-05-20 at 10 26 47@2x

### DIFF
--- a/src/screens/Messages/List/ChatListItem.tsx
+++ b/src/screens/Messages/List/ChatListItem.tsx
@@ -284,7 +284,6 @@ function ChatListItemReady({
         currentScreen="list"
         showMarkAsRead={convo.unreadCount > 0}
         hideTrigger={isNative}
-        triggerOpacity={!gtMobile || showActions || menuControl.isOpen ? 1 : 0}
         blockInfo={blockInfo}
         style={[
           a.absolute,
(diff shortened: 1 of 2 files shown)


src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -152,102 +144,119 @@ function ChatListItemReady({
...
-        onAccessibilityAction={onLongPress}>
+        onAccessibilityAction={onLongPress}
+        onPress={
+          isDeletedAccount
+            ? e => {
+                e.preventDefault()
+                return false
+              }
+            : undefined
+        }
```
haileyok, line 155: Cool for now, but we should make this a prop on our new link maybe, I've had to do this a few times before. Just something for later.

PR #4116, opened 2024-05-20
[🐴] Screenreader accessibility actions for individual chat items

haileyok:
Adding an accessibility "action" to each message. This will let the user hear what the option is and how to activate it.

https://github.com/bluesky-social/social-app/assets/153161762/52e3640a-90f1-4969-9891-af0f2a0d34bf


estrattonbailey (approved):
Nice one

PR #4119, opened 2024-05-20
[🐴] Fix load state for disabled chats

estrattonbailey:
Small, but noticed we never stopped the load state for users whose DM privileges have been revoked.

https://github.com/bluesky-social/social-app/assets/4732330/e9e9723e-7da8-4dd9-92cd-569a73438ab0


haileyok (approved):
oops, thanks!

PR #4121, opened 2024-05-20
Allow removal of feeds that fail to load

estrattonbailey:
Two issues here:
- if the feed fails to load, the `find` for `SavedFeedConfig` fails and we don't have anything to remove from prefs
- the warning dialog when removing a saved feed isn't part of the failed/loading state
  - fix here is to simply call `onUnsave`, don't need to warn the user if feed is failing to load and they want to remove


pfrazee (approved):
Yeah lgtm

PR #4126, opened 2024-05-20
[🐴] Appeal form for disabled DMs

mozzius:
mostly a straight copy of `LabelsOnMeDialog`, although I improved the onSubmit action by using `useMutation` and backported it to `LabelsOnMeDialog`

### DIFF
--- a/src/screens/Messages/Conversation/ChatDisabled.tsx
+++ b/src/screens/Messages/Conversation/ChatDisabled.tsx
@@ -1,8 +1,18 @@
-import React from 'react'
+import React, {useCallback, useState} from 'react'
 import {View} from 'react-native'
-import {Trans} from '@lingui/macro'
+import {ComAtprotoModerationDefs} from '@atproto/api'
+import {msg, Trans} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+import {useMutation} from '@tanstack/react-query'
 
-import {atoms as a, useTheme} from '#/alf'
+import {logger} from '#/logger'
+import {useAgent, useSession} from '#/state/session'
+import * as Toast from '#/view/com/util/Toast'
+import {atoms as a, useBreakpoints, useTheme, web} from '#/alf'
+import {Button, ButtonIcon, ButtonText} from '#/components/Button'
+import * as Dialog from '#/components/Dialog'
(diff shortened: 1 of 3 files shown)


src/components/dms/MessagesListHeader.tsx
```diff
@@ -176,6 +177,13 @@ function HeaderReady({
               ]}
               numberOfLines={1}>
               @{profile.handle}
+              {convoState.convo?.muted && (
+                <>
+                  {' '}
+                  &middot;{' '}
+                  <BellStroke size="xs" style={t.atoms.text_contrast_medium} />
+                </>
+              )}
```
estrattonbailey: Heh nice

src/screens/Messages/Conversation/ChatDisabled.tsx
```diff
@@ -20,7 +30,136 @@ export function ChatDisabled() {
...
+            ...web({outline: 0}),
+            textDecorationLine: 'underline',
+            textDecorationColor: t.palette.primary_500,
+          },
+        ]}
+        onPress={control.open}
+        onPressIn={onPressIn}
+        onPressOut={onPressOut}>
+        <Trans>Appeal this decision</Trans>
+      </Text>
```
estrattonbailey, line 67: Rn this isn't a tabbable element, so not good for a11y. This is the hacky way we have atm to add an inline link style button: https://github.com/bluesky-social/social-app/blob/main/src/components/dms/MessageItem.tsx#L222

estrattonbailey: Basically it's a link with an `onPress` that disables the link handling

estrattonbailey: Could alternatively just use a `small` sized `Button`

estrattonbailey (commented):
Looks great! Let's just do something more accessible for the trigger

src/screens/Messages/Conversation/ChatDisabled.tsx
mozzius: I think a small button is the way to go here

estrattonbailey (approved):

PR #4128, opened 2024-05-20
[🐴 Tweak appeal button styles

estrattonbailey:
CleanShot 2024-05-20 at 16 32 31@2x


mozzius (approved):
turns out it's actually quite helpful to see what you're working on

PR #4133, opened 2024-05-20
[🐴] Prevent flash upon first message

estrattonbailey:
Test with old version
- create fresh chat, leave one, or otherwise reset to no items in the MessageList
- send first message, see flash of load state

This PR fixes that, but probably breaks other stuff.


haileyok (approved):
this'll fix it. basically, since the _first_ message won't trigger the scroll until after it renders, we get a brief flash while `hasScrolled` is still false. Since we know we're about to "scroll" after sending that first message, this will prevent that.

kind of annoying hack all around, but for now it's what we have to do.

PR #4135, opened 2024-05-21
[🐴] set layout height in `onLayout`

haileyok:


estrattonbailey (approved):
Tested on ios/web, works great

PR #4146, opened 2024-05-21
Only fallback to Discover if Following is first pinned

gaearon:
Removes the gate check since it isn't correct — the population in the split that are eligible for new onboarding haven't necessarily gone through the new onboarding.

Instead, use a heuristic. If the Following feed is the first pinned feed, keep the old behavior with the fallback. If it has been moved (manually or due to new onboarding), don't mix in Discover in there. This should roughly map to the old vs new onboarding, but doesn't depend on the gate.

## Test Plan

Add console logs to function creation. Verify putting Following first will create the "mixed in" instance, while putting something in front of it would create the "raw" following.

### DIFF
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -17,7 +17,6 @@ import {
 import {HomeFeedAPI} from '#/lib/api/feed/home'
 import {aggregateUserInterests} from '#/lib/api/feed/utils'
 import {moderatePost_wrapped as moderatePost} from '#/lib/moderatePost_wrapped'
-import {useGate} from '#/lib/statsig/statsig'
 import {logger} from '#/logger'
 import {STALE} from '#/state/queries'
 import {DEFAULT_LOGGED_OUT_PREFERENCES} from '#/state/queries/preferences/const'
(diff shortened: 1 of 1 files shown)


src/state/queries/post-feed.ts
```diff
@@ -111,14 +110,18 @@ export function usePostFeedQuery(
   const enabled =
     opts?.enabled !== false && Boolean(moderationOpts) && Boolean(preferences)
   const userInterests = aggregateUserInterests(preferences)
+  const followingPinnedIndex =
+    preferences?.savedFeeds?.findIndex(
```
gaearon, line 114: I assume these are safe to query (it'll convert old prefs on the fly?)

src/state/queries/post-feed.ts
```diff
@@ -410,10 +409,10 @@ function createApi({
         userInterests,
       })
     } else {
-      if (useBaseFollowingFeed) {
-        return new FollowingFeedAPI({getAgent})
-      } else {
+      if (enableFollowingToDiscoverFallback) {
```
gaearon, line 412: I found flipping this a little easier to think about, like an extra feature whose natural state is "off". Then `=== 0` condition above seems pretty plain, and doesn't need to be inverted.

haileyok (approved):
Tested on device, works correctly. Switching to following first gives me discover fallback while if I move it to last I get the "Pick some follows" copy.

pfrazee (approved):
Yep, tested, looks good

PR #4156, opened 2024-05-21
More retries when resuming non-stale account

estrattonbailey:
A few of us have noticed getting signed out occasionally. For me anyway, this seems to happen when resuming from a background state e.g. foregrounding the app, or opening my laptop after being away.

The thinking in this PR is: if we have checked your JWT and it's not expired, we want to be _very_ sure the `resumeSession` call returns a real failure, and not some sort of network issue while the device reconnects to its wifi/cellular.

Other things that could be done here:
- check for network errors and retry those, but fail on actual auth errors
- add a back-off, so requests are spaced out a bit

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -53,7 +53,7 @@ export async function createAgentAndResume(
     agent.session = prevSession
     if (!storedAccount.deactivated) {
       // Intentionally not awaited to unblock the UI:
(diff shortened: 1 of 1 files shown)


gaearon (approved):

mozzius (approved):
Cant think of any downsides

estrattonbailey:
I'll have a look at improving this based on the things I mentioned above, but yeah for now this won't hurt

PR #4161, opened 2024-05-22
Add React Compiler

poteto:
In my ReactConf [talk](https://www.youtube.com/watch?v=T8TZQ6k4SLE&t=12020s) I demo'd adding React Compiler to bluesky, so I wanted to add it for real! 

Adding the compiler and eslint plugin was pretty straightforward. The only hiccup I ran into was an interesting interaction with @lingui/macro:

lingui recommends running their `macro` plugin [first in the pipeline](https://lingui.dev/ref/macro). Normally with the React Compiler, the compiler plugin should run first as we want to see the original code as it was written. However, this sometimes causes conflicts with other babel plugins.

In this case, it looks like the @lingui/macro plugin does some very light transformation that the compiler can still understand and compile correctly, so I made it run first.

Without letting @lingui/macro run first, the compiler would cause the @lingui/macro plugin to crash because it seems like it would strip off the `extra.raw` property off of StringLiterals which was being used [here](https://github.com/lingui/js-lingui/blob/1293412c5dcc565636403443788a5b5d4ca206c1/packages/macro/src/macroJsx.ts#L395). I need to figure out why the compiler is doing that, but for now this works and should be a safe change unless there were specific reasons
the macro plugin was placed 2nd to last.

Screenshot 2024-05-21 at 8 34 46 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     'lingui',
     'simple-import-sort',
     'bsky-internal',
+    'eslint-plugin-react-compiler',
   ],
   rules: {
     // Temporary until https://github.com/facebook/react-native/pull/43756 gets into a release.
@@ -67,6 +68,7 @@ module.exports = {
       },
     ],
     'simple-import-sort/exports': 'warn',
+    'react-compiler/react-compiler': 'error',
   },
   ignorePatterns: [
     '**/__mocks__/*.ts',
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,3 +12,4 @@
 android
 ios
 src/locale/locales
+lib/react-compiler-runtime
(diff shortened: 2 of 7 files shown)


gaearon (approved):
lfg

IgorGanapolsky:
I am curious if this work in production now, and how you upgraded the Babel plugin?  I tried this, and my Expo app completely broke...

gaearon:
yeah, i don't think we've had a single issue related to it

poteto:
@gaearon it might also be worth upgrading the compiler and letting us know if there are any issues! We've fixed a ton of bugs since then

gaearon:
yeah! will need to try that. what's a good hash that has been baked for a bit?

poteto:
@gaearon https://www.npmjs.com/package/babel-plugin-react-compiler/v/0.0.0-experimental-8d2bec7-20240812 has been in use internally since 8/9 and we haven't seen any issues yet

PR #4163, opened 2024-05-22
[WIP] Threaded composer v2

mary-ext:
Continuation of #3813, moved to a separate branch since this one is being based over https://github.com/bluesky-social/social-app/commit/8a12b606ee09b5e74fad243400f6279bbe99b99d

Actual composer-related changes starting from 36ec7fe3a739b8efb96da3fcd6373967ad24949a, with the newer changes starting bb355d0ce7697acca580df613ea591902b0b7113

UI looks more or less the same as before though

## Checklist

- [x] The actual threading UI
- [x] Decent-ish keyboard navigation (I think we can do better, perhaps by making use of Ctrl/Meta-Up/Down to control focus?)
- [x] External embeds
- [x] Tenor GIFs (and alt text)
- [x] Quote embeds
- [x] Image embeds
- [ ] Post language
  - Including the automatic language detection
- [x] Gated reply interactions
- [x] Self-labeling
- [x] Send the post ([demo](https://github.com/mary-ext/demo-bsky-thread-posting))
- [ ] Address the TODOs left in comments

## Notable changes

- Embeds are represented very differently now, with only a singular `embed` field
  - **Cons:**
    - The reducer code I've written to update the embeds might be rather fragile, though the dispatcher actions wouldn't have looked much different I think.
  - **Pros:**
    - Easy conversion to actual lexicons, since it's pretty much based around that
    - It's more clear what type of embeds can or can't be added or mixed given the current state
    - It should be easier to add new embed types, like videos or custom GIFs, possibly with a different set of constraints on what can be mixed alongside them
    - More control over styling, particularly around `recordWithMedia`
    - Easy memoization, embeds don't change often, applying it to the top-level component is enough
    - Tenor GIF embeds are easier to deal with now that they're no longer conflated with external embeds (while still being sent as one)
- `record` and `external` embeds are now handled by React Query, post submission no longer needs to be disabled as we try to retrieve details for the embed, since we can also retrieve them as part of the post submission process itself.
- `record` embed can now be added after adding `external` embed, previously this wasn't possible because `record` embeds had to go through the same process as `external` embeds, and doing so would just override the `external` embed before it settles as a `record` embed
- `record` embeds now show a proper preview for lists and feeds in particular.
- Labels are stored within `image` and `external` embeds, I think right now this makes a lot more sense since the only globals labels that exists are for media embeds only. This also means that labels are no longer persisted when embeds are being swapped for another, which might better match user expectations.
- Thread gating records are now created in the same operation

## Additional questions

- Should the language and labeling options be moved to the same place as reply gate currently is?
- How should post languages work across multiple posts? At least from the few I've asked, they want the ability to set different languages per-post, and I honestly agree with that


mary-ext:
There's still some work to be done here, but I think the additional work scope required (like how post languages should work) means that I should now pass this off to y'all. cc @pfrazee @haileyok

Happy to have worked on this :v:

pfrazee:
Sick! Look forward to giving it all a look.

mary-ext:
Between the many Indonesian users moving in, there are a decent amount of fanfic authors looking for threaded composer and drafts.

I think it might be worth making this happen soon, let me know if there's anything else I can help with on getting this pushed forwards.

pfrazee:
Nope! We will wrap this fairly soonOn Jun 8, 2024, at 6:07 AM, Mary ***@***.***> wrote:﻿
Between the many Indonesian users moving in, there are a decent amount of fanfic authors looking for threaded composer and drafts.
I think it might be worth making this happen soon, let me know if there's anything else I can help with on getting this pushed forwards.

koyuawsmbrtn:
It's been three months and now the Brazilians are flooding the platform. Guess it's time to finally review and hopefully merge. Please.

mary-ext:
There's been a lot of changes to the composer itself since this PR was made, and I need to find the time to work on rebasing the changes.

mary-ext:
Good news is that I should have some time to spare on rebasing this PR, problem is, we'd still need to get #3925 merged first (I'll need to rebase this as well, but if that's blocked, this one's blocked)

gaearon:
Ayy, we're finally getting to parity (aside from UI) in https://github.com/bluesky-social/social-app/pull/5962. Huge thanks for getting this started — no way I'd be able to figure out how to do it without your work here.

PR #4164, opened 2024-05-22
don't use `contentVisibility` on Firefox

haileyok:
## Why

The `content-visibility` style on Firefox is causing problems with the `List` implementation - at least for the list we use for chat conversations.

This PR just checks the useragent for `Firefox` and does not use this style - the same way we do not use it for Safari - if it matches.

## Test Plan

With this PR, the chat list should be fine on Firefox.

https://github.com/bluesky-social/social-app/assets/153161762/9e7732f9-7296-406d-a665-59aff3826b76

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -504,6 +504,7 @@ export const List = memo(React.forwardRef(ListImpl)) as <ItemT>(
 
 // https://stackoverflow.com/questions/7944460/detect-safari-browser
 const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
+const isFirefox = /firefox|fxios/i.test(navigator.userAgent)
 
 const styles = StyleSheet.create({
   sideBorders: {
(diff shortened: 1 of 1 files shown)


haileyok:
@gaearon I mentioned in Slack that I'm not certain what we're using this style for. It isn't (visibly anyway?) causing problems elsewhere in the app that we know of, so if it is useful for some reason we can scope this down further, maybe with a prop on `List`.

mary-ext:
`content-visibility` will pretty much ruin layout-based animations, since unless the browser knows it should render it, then the elements are assumed to have no height/width initially. I'd honestly recommend turning it off for chat specifically on all browsers, since users aren't probably going to go back and forth between pages

haileyok:
Ah that does make sense. Kind of wonder if this is what causes the ghost scroll position issues for some people (the content flashing up and down the screen). We still get reports of that occasionally though have never been able to reproduce.

mary-ext:
It's going to be hit-and-miss when it comes to a 100% reproduction I think, I've encountered it a few times but it wasn't anything I could do consistently.

mary-ext:
Hmm, it might be, and not sure if it's something that can be remedied easily too.

There is `contain-intrinsic-height: auto` which will make the element retain its previously-known size once it goes offscreen, but unfortunately that size measurement gets lost when it gets detached from DOM.

mozzius (approved):
Works like a charm

gaearon (approved):

PR #4167, opened 2024-05-22
[Embeds] stop adding tracking params to non-bsky.app links

mozzius:
The `ref_src` and `ref_url` params were being added to the embed cards within the embed, which is undesirable. This adds a `disableTracking` prop to the link component to try and avoid that

### DIFF
--- a/bskyembed/.eslintrc
+++ b/bskyembed/.eslintrc
@@ -15,6 +15,6 @@
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": "latest",
-    "project": "./tsconfig.json"
+    "project": "./bskyembed/tsconfig.json"
   }
 }
--- a/bskyembed/src/components/post.tsx
+++ b/bskyembed/src/components/post.tsx
@@ -1,4 +1,9 @@
-import {AppBskyFeedDefs, AppBskyFeedPost, RichText} from '@atproto/api'
+import {
+  AppBskyFeedDefs,
+  AppBskyFeedPost,
+  AppBskyRichtextFacet,
+  RichText,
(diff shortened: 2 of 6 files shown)


bskyembed/.eslintrc
```diff
@@ -15,6 +15,6 @@
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": "latest",
-    "project": "./tsconfig.json"
+    "project": "./bskyembed/tsconfig.json"
```
gaearon, line 18: ?

bskyembed/src/components/post.tsx
```diff
@@ -123,7 +123,8 @@ function PostContent({record}: {record: AppBskyFeedPost.Record | null}) {
         <Link
           key={counter}
           href={segment.link.uri}
-          className="text-blue-400 hover:underline">
+          className="text-blue-400 hover:underline"
+          disableTracking={!segment.link.uri.startsWith('https://bsky.app')}>
```
gaearon, line 127: is `segment.link.uri` guaranteed to exist?

gaearon (approved):
lgtm

bskyembed/.eslintrc
mozzius: Eslint wouldn't let me merge unless I did this - for some reason it looks at this path from the perspective of the repo root? Very odd, I don't know if there's something else funny going on

bskyembed/src/components/post.tsx
mozzius: Good shout, updated to use `validateLink` which is what the main app's richtext does

PR #4168, opened 2024-05-22
[🐴] Shadow profiles for unread count

mozzius:
This might be an insane way of doing it, let me know


estrattonbailey (approved):
LGTM, let's get Dan's thoughts too

mozzius:
spoke with Dan, best to avoid adding more hacks onto shadows. Closing

Issue #2513, opened 2024-01-12
Search history like Twitter (texts and users)


xtianpu:
**Is your feature request related to a problem? Please describe.**

I want a simple search history similar to Twitter

I’m always frustrated to retyping my frequent searches for users or some query that I need info from.

**Describe the solution you'd like**

search history similar to Twitter

**Describe alternatives you've considered**

i dont have any

**Additional context**

![IMG_2509](https://github.com/bluesky-social/social-app/assets/8172235/bb4c8ed4-79a0-4de0-ae56-d20f347f5207)

skiniks:
I started a basic implementation of this #2589.

xtianpu:
> I started a basic implementation of this #2589.

Thanks! Will you add recent searches for users too? Maybe it can be implemented when user's clicking suggestion profile searches.

pfrazee:
Thank you to @skiniks, we do now have this for searches that you fully run (but not autocompletes).

For the users, I wonder if what we do is keep a history of users you selected due to an autocomplete and show that as well

boly38:
maybe duplicate of #1677

Issue #1677, opened 2023-10-12
Show Recent Searches


MichaelGHUser:
**Describe the solution you'd like**

Show five most-recent searches upon interacting with the Search field.

**Describe alternatives you've considered**

Manually searching every time. 

**Additional context**

Not a must-have feature.  It would bring Bluesky closer to feature parity with other platforms, however.  More of a quality-of-life add.

![image](https://github.com/bluesky-social/social-app/assets/123146201/a217c2c0-b9b3-4efa-9bdb-d15b6bd69eee)

Zero3K:
Any news regarding this?

boly38:
seems duplicate of #2513

PR #4169, opened 2024-05-22
add profiles to search history

skiniks:
This PR satisfies #2513, a duplicate of #1677. The SearchHistory component is enhanced by fully implementing the display of recent profiles. Each recent profile now includes the avatar, display name, and handle. Additionally, the display name and handle are truncated with ellipses if they exceed a specified character limit to ensure they fit within the given space.

Screenshot 2024-05-22 at 3 32 30 PM


src/view/screens/Search/Search.tsx
```diff
@@ -562,22 +572,68 @@ export function SearchScreen(
...
+
   const navigateToItem = React.useCallback(
     (item: string) => {
       scrollToTopWeb()
       setShowAutocomplete(false)
       updateSearchHistory(item)
 
       if (isWeb) {
-        navigation.push('Search', {q: item})
+        navigation.replace('Search', {q: item})
```
gaearon, line 606: why replace? i think push was intentional

src/view/screens/Search/Search.tsx
```diff
@@ -880,12 +982,66 @@ function SearchHistory({
...
+                      accessibilityIgnoresInvertColors
+                    />
+                    <Text style={[pal.text, styles.profileName]}>
+                      {truncateText(profile.displayName || '', 12)}
+                    </Text>
+                    <Text style={[pal.textLight, styles.profileHandle]}>
+                      @{truncateText(profile.handle, 12)}
+                    </Text>
+                  </Pressable>
+                  <Pressable
```
gaearon, line 1026: Does this need accessibility label?

src/view/screens/Search/Search.tsx
skiniks: my mistake! i've reverted that.

src/view/screens/Search/Search.tsx
skiniks: definitely! i've added the label and hint.

gaearon:
A few issues.

Currently, pressing the avatar hides the dropdown before pushing the next view. That causes flicker:

https://github.com/bluesky-social/social-app/assets/810438/e4f8711e-d52f-4fbb-aa1e-a18728d6828c

I think we'd want to avoid that.

Another problem is that pushing an autocomplete result now seems to navigate _twice_.

https://github.com/bluesky-social/social-app/assets/810438/ca43166c-3877-4de5-a41a-1565d4d6af3b

It seems like you're adding some navigation logic — maybe existing logic could be reused? Whatever we're doing for the existing autocomplete. I presume it wouldn't have those issues? But I haven't looked closely.

gaearon (changes requested):
need to fix flicker and double nav

skiniks:
it looks like removing this fixes the issues:

```javascript
scrollToTopWeb()
setShowAutocomplete(false)
updateSelectedProfiles(profile)

gaearon:
seems like Back button is broken


https://github.com/bluesky-social/social-app/assets/810438/df5df32c-4f87-48fd-89fc-f0708712c44e

gaearon (approved):
thanks!

PR #4174, opened 2024-05-22
Log error statuses from failed resume session calls

estrattonbailey:
I want to better understand what failures occur here. We can pull status codes from our backend logs too, but network connectivity issues and local timeouts need to be captured here.

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -52,8 +52,17 @@ export async function createAgentAndResume(
   } else {
     agent.session = prevSession
     if (!storedAccount.deactivated) {
-      // Intentionally not awaited to unblock the UI:
-      networkRetry(3, () => agent.resumeSession(prevSession))
+      try {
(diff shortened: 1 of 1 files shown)


src/state/session/agent.ts
```diff
@@ -52,8 +52,18 @@ export async function createAgentAndResume(
   } else {
     agent.session = prevSession
     if (!storedAccount.deactivated) {
-      // Intentionally not awaited to unblock the UI:
-      networkRetry(3, () => agent.resumeSession(prevSession))
+      try {
+        // Intentionally not awaited to unblock the UI:
+        networkRetry(3, () => agent.resumeSession(prevSession))
+      } catch (e: any) {
```
gaearon, line 58: This isn't awaited so I think you want to do a `.catch()` instead — try/catch won't catch it

estrattonbailey: 🤦

gaearon (approved):

PR #4176, opened 2024-05-22
wrap web in disabled keyboard provider

haileyok:
## Why

`react-native-keyboard-controller` logs a warning - on every keystroke - if the app isn't wrapped in a `KeyboardProvider`. This makes sense of course on native, where it's a useful indication that something is wrong, but annoying when debugging on web.

This wraps the app in the keyboard provider on web as well to silence these warnings.

The context (https://github.com/kirillzyusko/react-native-keyboard-controller/blob/6da7f58bd692e0eeac26e44fd4960c1f4a31f3c0/src/context.ts#L20) contains just a few shared values and a couple functions. Hooks are no-op on web. https://github.com/kirillzyusko/react-native-keyboard-controller/blob/6da7f58bd692e0eeac26e44fd4960c1f4a31f3c0/src/reanimated.ts

### DIFF
--- a/src/App.web.tsx
+++ b/src/App.web.tsx
@@ -2,6 +2,7 @@ import 'lib/sentry' // must be near top
 import 'view/icons'
 
 import React, {useEffect, useState} from 'react'
+import {KeyboardProvider} from 'react-native-keyboard-controller'
 import {RootSiblingParent} from 'react-native-root-siblings'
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Booted web, no warnings that I see, seems to work great 👍

PR #4180, opened 2024-05-23
HomeMenuItem goes above SearchMenuItem in a sessioned Drawer

jcsalterego:
Unsure if this is wanted but in a Drawer with a session, move Home to the top to be consistent with other navigation menus.

### Before

before

### After

after

### DIFF
--- a/src/view/shell/Drawer.tsx
+++ b/src/view/shell/Drawer.tsx
@@ -246,8 +246,8 @@ let DrawerContent = ({}: {}): React.ReactNode => {
           {hasSession ? (
             <>
               <View style={{height: 16}} />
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Yea, that looks like an error when changing the icons. Thanks.

Edit: Actually, maybe it isn't...let me find out.

mozzius:
Yeah other apps often have search at the top?

haileyok (changes requested):
Removing approval for now just so we can talk about this tomorrow.

jcsalterego:
_Sits in the waiting room, flips through_ Poster's Monthly _Magazine_

jcsalterego:
It works now for some reason(?). I am closing this PR

PR #4181, opened 2024-05-23
[Session] Persist updates from inactive agent

gaearon:
If we have an agent in memory (but it isn't the active one), it's still possible that its session would get refreshed (e.g. if we have a dangling interval call somewhere that we didn't clean up). In that case, it's important that we actually persist the session since refreshing it has already consumed the stored refresh token.

It's possible this may be one of the causes of the reported unexpected logouts.

This PR changes the logic so that any agent's `update` event leads to updating persisted tokens — even if it's not the active agent. This should mean that, whenever we get a new session from the server (for any agent), it will get written into the persisted storage.

Note that in theory, we could have multiple agents for the same account "active" at the same time, which I think would mean that they might start "fighting" with each other. I'll think more about this case and <s>maybe send some follow-up</s> send https://github.com/bluesky-social/social-app/pull/4183.

## Test Plan

Mostly tests cause I'm not sure how to actually trigger refresh or expiry.

### DIFF
--- a/src/state/session/__tests__/session-test.ts
+++ b/src/state/session/__tests__/session-test.ts
@@ -872,7 +872,7 @@ describe('session', () => {
     expect(state.accounts[0].accessJwt).toBe('alice-access-jwt-3')
   })
 
-  it('ignores updates from a stale agent', () => {
+  it('accepts updates from a stale agent', () => {
     let state = getInitialState([])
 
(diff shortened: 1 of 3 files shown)


gaearon:
merged via https://github.com/bluesky-social/social-app/pull/4183

PR #4182, opened 2024-05-23
Add padding to dialogs when keyboard is open on Android

haileyok:
## Why

Because of the needed changes on Android for the keyboard to behave correctly for DMs (as well as, eventually, the composer), it seems to have affected some users ability to properly scroll down to the button for saving alt-text. This is also apparent on the report dialog. Although it doesn't _break_ functionality (the text can still be input) it isn't possible to see the input, since the container does not get resized correctly.

While it would be possible to use a `KeyboardAvoidingView` inside of the new dialogs fairly easily, this becomes complicated with dialogs that are presented inside of the composer itself. That is because our composer is actually a dialog in the _old_ system, and then we are presenting another _new_  dialog inside of that one. This results in a `KeyboardAvoidingView` that wrap other `KeyboardAvoidingView`s. Eventually we won't have this problem, if we move to a better dialog for the composer.

For now, we can use an animated view to apply the padding on Android as needed. This actually results in a better experience than what we had before - the input is automatically scrolled to now, something that previously only worked well on iOS.

## Test Plan

There are a variety of dialogs on Android where this is useful at. There is no change on iOS, since `null` is returned instead of a view on that platform.

### Report Dialog

https://github.com/bluesky-social/social-app/assets/153161762/8f7efede-d986-4d5a-9157-a18ace4901c9

### Alt Text

https://github.com/bluesky-social/social-app/assets/153161762/98ed02b8-5e11-42a8-98ef-937166211443

### Appeal Dialog

https://github.com/bluesky-social/social-app/assets/153161762/4ff18bd5-a6c2-450d-a5fc-8e19b6c7265b

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -1,5 +1,12 @@
 import React, {useImperativeHandle} from 'react'
-import {Dimensions, Pressable, StyleProp, View, ViewStyle} from 'react-native'
+import {
+  Dimensions,
+  Keyboard,
+  Pressable,
+  StyleProp,
+  View,
+  ViewStyle,
+} from 'react-native'
 import Animated, {useAnimatedStyle} from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import BottomSheet, {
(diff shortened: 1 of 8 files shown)


estrattonbailey (approved):
Looks like a slick solution to me

PR #4184, opened 2024-05-23
Handle zero bottom inset on iOS

haileyok:
## Why

It is fairly difficult on iOS for us to calculate our real safe area because our bottom is clamping the padding based on the device's inset. For example, on an iPhone 8 or lower, the padding gets set to `15` - because the real bottom inset is `0`. However, on other devices it becomes `30` because the inset _is_ set (an iPhone X or higher has a bottom inset of `34`).

this results in some bad behavior with the message input in the chat screen. We need to account for this _missing_ padding in these circumstances.

On most devices this offset is very small, even on Android devices. Newer iPhone devices have the tallest offsets, so for thos devices we want to set the max to 76. The minimum - when the offset is zero - should be 60.

## Test Plan

With the adjusted offset for these circumstances, all devices should appear to have the same style for the text input

Screenshot 2024-05-22 at 7 10 24 PM

### DIFF
--- a/src/screens/Messages/Conversation/MessagesList.tsx
+++ b/src/screens/Messages/Conversation/MessagesList.tsx
@@ -222,7 +222,7 @@ export function MessagesList({
 
   // -- Keyboard animation handling
   const {bottom: bottomInset} = useSafeAreaInsets()
-  const bottomOffset = isWeb ? 0 : clamp(60 + bottomInset, 60, 76)
+  const bottomOffset = isWeb ? 0 : clamp(60 + bottomInset, 60, 75)
 
   const keyboardHeight = useSharedValue(0)
   const keyboardIsOpening = useSharedValue(false)
--- a/src/screens/Messages/Conversation/MessagesList.tsx
+++ b/src/screens/Messages/Conversation/MessagesList.tsx
@@ -16,10 +16,11 @@ import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {AppBskyRichtextFacet, RichText} from '@atproto/api'
 
 import {shortenLinks} from '#/lib/strings/rich-text-manip'
-import {isIOS, isNative} from '#/platform/detection'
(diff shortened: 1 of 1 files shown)


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -221,8 +222,7 @@ export function MessagesList({
 
   // -- Keyboard animation handling
   const {bottom: bottomInset} = useSafeAreaInsets()
-  const nativeBottomBarHeight = isIOS ? 42 : 60
-  const bottomOffset = isWeb ? 0 : bottomInset + nativeBottomBarHeight
+  const bottomOffset = isWeb ? 0 : clamp(60 + bottomInset, 60, 76)
```
haileyok: This 15 pixel range comes from https://github.com/bluesky-social/social-app/blob/03655abb7cfbe8fea3f73fc4a298e54364c1de63/src/view/shell/bottom-bar/BottomBar.tsx#L134

src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -221,8 +222,7 @@ export function MessagesList({
 
   // -- Keyboard animation handling
   const {bottom: bottomInset} = useSafeAreaInsets()
-  const nativeBottomBarHeight = isIOS ? 42 : 60
-  const bottomOffset = isWeb ? 0 : bottomInset + nativeBottomBarHeight
+  const bottomOffset = isWeb ? 0 : clamp(60 + bottomInset, 60, 75)
```
haileyok, line 225: This 15 pixel range comes from https://github.com/bluesky-social/social-app/blob/03655abb7cfbe8fea3f73fc4a298e54364c1de63/src/view/shell/bottom-bar/BottomBar.tsx#L134

gaearon (approved):

PR #4187, opened 2024-05-23
✍️ Add OTA Docs

haileyok:
A little step by step for deploying an OTA release. I'll add another later for creating web bundles off of a branch, though those are fairly straight forward.

Added a scheme to follow for branch names - `1.X.0-ota-1`, then incrementing `1.X.0-ota-2` until the next client is tagged.


mozzius (approved):
great, tysm for writing this up

docs/deploy-ota.md
```diff
@@ -0,0 +1,62 @@
...
+- Identify the build number of the production app you want to deploy an update for. iOS and Android build numbers are divergent, so you will need to find both
+
+  ![app-build-number](./img/app-build-number.jpg)
+
+- Ensure that the commit the initial client was cut from is properly tagged. The tag should be in the format of 1.X.0
+
+### Preparation
+
+- Create a new branch from the tag that the initial release was cut from if no OTA deployment has been made yet for this client. Name this branch 1.X.0-ota-1
+- If a deployment has been made previously for this release, increment the tag, i.e. 1.x.0-ota-2
```
estrattonbailey, line 26: Should pry clarify, you mean literal git tags here. So we have a branch naming convention, plus we also tag those commits with tags following the same convention?

estrattonbailey: What purpose to tags serve here? When do we normally tag commits for release?

docs/deploy-ota.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Prerequisites
+
+- Remove any internal client from your device and download the client from the App Store/Google Play. This will help for testing as well as retrieving the build number.
+- You should have signed in to EAS locally through npx eas login. You will need to modify the build number in a subsequent step.
+- Identify the build number of the production app you want to deploy an update for. iOS and Android build numbers are divergent, so you will need to find both
+
+  ![app-build-number](./img/app-build-number.jpg)
+
+- Ensure that the commit the initial client was cut from is properly tagged. The tag should be in the format of 1.X.0
```
estrattonbailey, line 21: Nit: what if it's not

docs/deploy-ota.md
```diff
@@ -0,0 +1,62 @@
...
+
+Internal OTA deployments should be performed automatically upon all merges into main. In cases where the fingerprint diff results in incompatible native changes, a new client build will automatically be ran and deployed to TestFlight (iOS) or delivered in Slack (Android).
+
+## Production Deployments
+
+### Prerequisites
+
+- Remove any internal client from your device and download the client from the App Store/Google Play. This will help for testing as well as retrieving the build number.
+- You should have signed in to EAS locally through npx eas login. You will need to modify the build number in a subsequent step.
+- Identify the build number of the production app you want to deploy an update for. iOS and Android build numbers are divergent, so you will need to find both
```
estrattonbailey, line 17: Nit: the build number is just the (239) or whatever, right?

docs/deploy-ota.md
```diff
@@ -0,0 +1,62 @@
...
+    ![branch-selection](./img/branch-selection.png)
+  
+  - Double check the branch selection.
+  - Select the production channel
+  - Enter the version for the client you are releasing to, i.e. 1.80.0
+  
+    ![other-ota-options](./img/other-ota-options.png)
+  
+  - Triple check the branch selection.
+  - Press “Run Workflow”
```
estrattonbailey, line 54: Maybe add a recap here to tie it all together, like
- So you set the build number to 240
- You set the version to 1.8.0
- You're going to release `1.8.0.240` or whatever

docs/deploy-ota.md
haileyok: There's no need to tag the commits on those other branches, since we would not see them on the main branch anyway. But yea, I'll clarify that.

The tags serve the purpose of knowing which commit to fingerprint your proposed deployment commit against. The workflow will get whatever commit hash was tagged with your input version and use that for the comparison.

docs/deploy-ota.md
haileyok: Oh yea, there's a photo that you can't see here that highlights the number you're looking for. But yes, that is correct.

docs/deploy-ota.md
haileyok: That last line should have been "increment the branch name". Changed.

PR #4189, opened 2024-05-23
[🐴] update convo list from message bus

mozzius:
Should stop the convo list being outdated. We might want to increase the poll rate while looking at the messages page.


estrattonbailey (approved):
Did a bunch of testing here with Sam the last couple days we're good here

PR #4190, opened 2024-05-23
[🐴] Suspend event bus when switching accounts

estrattonbailey:
This provider is mounted in `App.*.tsx` and unmounts when the account changes. When unmounting, we need to call `suspend()` to deactivate the active event bus.

### DIFF
--- a/src/state/messages/events/index.tsx
+++ b/src/state/messages/events/index.tsx
@@ -36,6 +36,10 @@ export function MessagesEventBusProvider({
       // @ts-ignore
       window.bus = bus
     }
+
+    return () => {
+      bus.suspend()
+    }
   }, [bus])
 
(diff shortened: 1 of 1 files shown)


src/state/messages/events/index.tsx
```diff
@@ -36,6 +36,10 @@ export function MessagesEventBusProvider({
       // @ts-ignore
       window.bus = bus
     }
```
gaearon, line 38: Would it hurt to call `bus.resume()` here? Ideally effects should be symmetric so that running effect->cleanup->effect still works.

estrattonbailey: Good call, added! Also removed dev debug code, no need for it really.

Issue #4195, opened 2024-05-23
Please add a link to Messages settings in the Moderation settings


surfdude29:
**Is your feature request related to a problem? Please describe.**

It would be great if it were easier to access Messages settings by making a link to them available in Moderation settings.

**Describe the solution you'd like**

I'd like a link to Messages settings to be added to Moderation settings.

**Describe alternatives you've considered**

An alternative would be the status quo, where Messages settings are only accessible from the Messages screen.

**Additional context**

This was inspired by @kbtibbs.bsky.social (I haven't checked with them on wording of this issue just to be clear):

https://bsky.app/profile/kbtibbs.bsky.social/post/3kt4q3d2fws2x

PR #4197, opened 2024-05-23
[🐴] add link to chat settings from main settings

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/4195

Added under basic

image

### DIFF
--- a/src/screens/Messages/Settings.tsx
+++ b/src/screens/Messages/Settings.tsx
@@ -56,7 +56,7 @@ export function MessagesSettingsScreen({}: Props) {
 
   return (
     <CenteredView sideBorders style={a.h_full_vh}>
-      <ViewHeader title={_(msg`Settings`)} showOnDesktop showBorder />
+      <ViewHeader title={_(msg`Chat Settings`)} showOnDesktop showBorder />
       <View style={[a.p_lg, a.gap_md]}>
         <Text style={[a.text_lg, a.font_bold]}>
           <Trans>Allow new messages from</Trans>
--- a/src/view/screens/Settings/index.tsx
+++ b/src/view/screens/Settings/index.tsx
@@ -647,6 +647,31 @@ export function SettingsScreen({}: Props) {
             <Trans>External Media Preferences</Trans>
           </Text>
         </TouchableOpacity>
+        <TouchableOpacity
+          testID="linkToChatSettingsBtn"
+          style={[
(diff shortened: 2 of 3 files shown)


src/screens/Messages/Settings.tsx
```diff
@@ -56,7 +56,7 @@ export function MessagesSettingsScreen({}: Props) {
 
   return (
     <CenteredView sideBorders style={a.h_full_vh}>
-      <ViewHeader title={_(msg`Settings`)} showOnDesktop showBorder />
+      <ViewHeader title={_(msg`Chat Settings`)} showOnDesktop showBorder />
```
haileyok, line 59:
```suggestion
      <ViewHeader title={_(msg`Chat Preferences`)} showOnDesktop showBorder />
```

haileyok (approved):

src/view/screens/Settings/index.tsx
```diff
@@ -647,6 +647,31 @@ export function SettingsScreen({}: Props) {
...
+          accessibilityLabel={_(msg`Chat settings`)}
+          accessibilityHint={_(msg`Opens chat settings`)}>
+          <View style={[styles.iconContainer, pal.btn]}>
+            <FontAwesomeIcon
+              icon={['far', 'comment-dots']}
+              style={pal.text as FontAwesomeIconStyle}
+            />
+          </View>
+          <Text type="lg" style={pal.text}>
+            <Trans>Chat Settings</Trans>
```
haileyok, line 672:
```suggestion
            <Trans>Chat Preferences</Trans>
```

We use "Preferences" everywhere else here, but fwiw we _are_ using "Settings" on the actual settings screen itself...

haileyok: Don't feel strongly, we can change this whenever.

mozzius: I think Settings here, just so it's the same when you click through. Don't feel strongly either tho

haileyok: Yea I realized that after going into the screen. Another option might be to just drop the "Settings" all together. You're in "Settings" already so you know it takes you to more settings.

Again though, that doesn't follow the other behavior we have lol...probably just something to revisit when we rework this screen hopefully soon.

mozzius: can't wait to burn it all down fr

PR #4198, opened 2024-05-23
disable alt text auto focus on Android

haileyok:
## Why

On Android, focusing this input automatically causes problems with resizing the container correctly. Even prior to the latest release - where something else broke - this still happens occasionally. It isn't consistent, it works 80% of the time but the other 20% it breaks.

For now, let's disable the autofocus on Android until we get a better chance to rework these dialogs.


mozzius (approved):
sound good, I think Mary has replaced this in her huge threading PR, maybe we could ask her to split it out

PR #4199, opened 2024-05-23
Add `bundleDate`, `bundleIdentifier` to `StatsigUser`

haileyok:
## Why

We want to be able to identify versions incrementally in Statsig. This change includes a number that increments on each update which can be used easily in Statsig to include users above a certain version inside of a subgroup. It also includes the bundle identifier as another signal - although this is more useful for tracking a very specific version (down to a specific OTA release or hotfix) - which is the hash of the commit which the bundle was created from.

## Test Plan

These should be available in Statsig once it rolls out. We can verify that these values are set and will be available by:

- Go to the settings page in the app
- Click on the version string at the bottom of the page
- Paste what is on your clipboard into an input box.

You should see an output like this:

```
Build version: 1.83.0; Bundle info: 0c2fb13 (prod); Bundle date: 24052218; Platform: web
```

### DIFF
--- a/src/lib/hooks/useOTAUpdates.ts
+++ b/src/lib/hooks/useOTAUpdates.ts
@@ -12,6 +12,7 @@ import {
 
 import {logger} from '#/logger'
 import {IS_TESTFLIGHT} from 'lib/app-info'
+import {logEvent} from 'lib/statsig/statsig'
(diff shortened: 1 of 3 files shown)


gaearon (approved):
accept but move into custom

PR #4200, opened 2024-05-23
Privileged app passwords

mozzius:
# Don't merge yet, awaiting backend + new @atproto/api release

image

image


haileyok (approved):
Code LGTM. Modal padding needed a fix, got that in.

Overall for these PWs:

- Tested a bsky account against skeetdeck
  - Non-privileged did not work
  - Privileged worked
- Tested my own PDS after updating with the same results as above.

PR #4201, opened 2024-05-23
[🐴] Fix Firefox send button positioning

haileyok:
Buttons on Safari, Firefox, and Chrome - respectively.

Screenshot 2024-05-23 at 11 40 52 AM

### DIFF
--- a/src/screens/Messages/Conversation/MessageInput.web.tsx
+++ b/src/screens/Messages/Conversation/MessageInput.web.tsx
@@ -90,6 +90,7 @@ export function MessageInput({
       <View
         style={[
           a.flex_row,
+          a.align_center,
           t.atoms.bg_contrast_25,
           {
             paddingHorizontal: a.p_sm.padding - 2,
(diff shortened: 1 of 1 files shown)


src/screens/Messages/Conversation/MessageInput.web.tsx
```diff
@@ -92,11 +93,12 @@ export function MessageInput({
           a.flex_row,
           t.atoms.bg_contrast_25,
           {
-            paddingHorizontal: a.p_sm.padding - 2,
+            paddingRight: 6,
```
estrattonbailey, line 96: Same value :P

estrattonbailey: Using the atom is just to indicate that it's suppose to match the atom used elsewhere for horizontal spacing, but looks like it's diverged and it lgtm so not real reason to use the atom I suppose 🤔

estrattonbailey (approved):

src/screens/Messages/Conversation/MessageInput.web.tsx
haileyok: oh yea, i ended up going back and forth playing w/ the value and ended up forgetting to put this back. I actually prefer to show the atom usage as it clarifies _why_ we didn't use it (as I get more and more into the habit of using them especially). Added that back 👍

PR #4223, opened 2024-05-26
Update French localizations

Signez:
A few more strings that were introduced / changed just after my last pass on translations (it happens! 🤷‍♂️), mainly about the Chat feature.

Pinging @surfdude29 and @pfrazee as usual, but nothing should raise any eyebrows anyway :)

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -17,7 +17,7 @@ msgstr ""
 msgid "(no email)"
 msgstr "(pas d’e-mail)"
 
-#: src/view/com/notifications/FeedItem.tsx:239
+#: src/view/com/notifications/FeedItem.tsx:234
 msgid "{0, plural, one {{formattedCount} other} other {{formattedCount} others}}"
 msgstr "{0, plural, one {{formattedCount} autre} other {{formattedCount} autres}}"
 
@@ -29,7 +29,7 @@ msgstr "{0, plural, one {# étiquette a été placée sur ce compte} other {# é
 msgid "{0, plural, one {# label has been placed on this content} other {# labels have been placed on this content}}"
 msgstr "{0, plural, one {# étiquette a été placée sur ce contenu} other {# étiquettes ont été placées sur ce contenu}}"
 
-#: src/view/com/util/post-ctrls/RepostButton.tsx:62
+#: src/view/com/util/post-ctrls/RepostButton.tsx:64
 msgid "{0, plural, one {# repost} other {# reposts}}"
 msgstr "{0, plural, one {# repost} other {# reposts}}"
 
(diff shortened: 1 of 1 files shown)


src/locale/locales/fr/messages.po
```diff
@@ -5217,7 +5208,7 @@ msgstr "Cet appel sera envoyé à <0>{0}</0>."
 
 #: src/screens/Messages/Conversation/ChatDisabled.tsx:104
 msgid "This appeal will be sent to Bluesky's moderation service."
-msgstr ""
+msgstr "Cet appel sera envoyé aux services de modération de Bluesky."
```
surfdude29, line 5211:
```suggestion
msgstr "Cet appel sera envoyé au service de modération de Bluesky."
```
I think `service de modération` should be singular rather than plural here?

src/locale/locales/fr/messages.po
```diff
@@ -5387,7 +5378,7 @@ msgstr "Pour désactiver le 2FA par e-mail, veuillez vérifier votre accès à l
 
 #: src/components/dms/ReportConversationPrompt.tsx:20
 msgid "To report a conversation, please report one of its messages via the conversation screen. This lets our moderators understand the context of your issue."
-msgstr ""
+msgstr "Pour signaler une conversation, veuillez signaler un de ses messages via l’écran de conversation. Cela permettra à nos modérateurs de comprendre le contexte du problème."
```
surfdude29, line 5381:
```suggestion
msgstr "Pour signaler une conversation, veuillez signaler un de ses messages via l’écran de conversation. Cela permettra à la modération de comprendre le contexte du problème."
```
Would it make sense to use _la langue épicène_ here, using a phrasing something like this?

surfdude29 (commented):
Sorry for the delay, excellent work as always @Signez 👌

I just have two points I wanted to mention.

surfdude29:
Also, I have one change to suggest that's not in this PR; I think the second string below can be simplified to match the first string:
https://github.com/bluesky-social/social-app/blob/8b5547bdc0d8f89b4b6611a966dc0cf0eef8c551/src/locale/locales/fr/messages.po#L2035-L2045

surfdude29:
And the suggested change from a user in #4296 is here:

https://github.com/bluesky-social/social-app/blob/8b5547bdc0d8f89b4b6611a966dc0cf0eef8c551/src/locale/locales/fr/messages.po#L1391-L1393

src/locale/locales/fr/messages.po
Signez: Good catch again — yeah, I tried to never use _"modérateurs"_ for this exact reason. 👍

Signez:
@surfdude29 Thanks again for your review!

All the suggested changes were applied in [this commit](https://github.com/bluesky-social/social-app/pull/4223/commits/a13918a6da570afb6388526a6db27bf16a6b4921), including the fix for #4296 and that weird long string for Following Feed Settings (that bothered me too, but I forgot to fix it somehow before); I also took the time to [bump the locales again](https://github.com/bluesky-social/social-app/pull/4223/commits/c10b20006f5e18b7af8d711d97f2d40df7c298b3) and translate [three new, very basic strings](https://github.com/bluesky-social/social-app/pull/4223/commits/c4237d275080271867f1ce387dbc39a6b4782499) to keep this PR up-to-date.

@pfrazee Now that this was reviewed by a peer, I think you can merge this with confidence :)

pfrazee (approved):
Thanks!!

PR #4226, opened 2024-05-27
use custom github action for fingerprinting

haileyok:
## Why

There are a few race conditions right now with our fingerprint workflow for TestFlight/internal builds because of how the Expo fingerprint action caches previous fingerprints - and relies on those fingerprints. This results in a lot of false positives for fingerprint changes when merging PRs in quick succession.

This PR uses a custom action (found at https://github.com/bluesky-social/github-actions/tree/main/fingerprint-native) that does not use the cache.

## Test Plan

- [x] PR - Properly detects no changes
- [x] PR - Properly detect a native change against main
- [x] PR - Properly comments on PR when change is introduced against main
- [x] PR - Deletes the comment whenever diff differences are removed
- [x] Merge/TF - Properly detects no changes
- [x] Merge/TF - Properly detects a native change against last TF commit
- [x] Production - Properly detects no changes
- [x] Production - Properly detects a native change against supplied tag


.github/workflows/bundle-deploy-eas-update.yml
```diff
@@ -49,69 +49,22 @@ jobs:
...
-
       - name: 🔧 Setup Node
         uses: actions/setup-node@v4
         with:
           node-version-file: .nvmrc
           cache: yarn
 
-      - name: ⚙️ Install Dependencies
-        run: yarn install
-
```
haileyok, line 94: The custom action will handle installing dependencies, since it needs to install deps from both commits. Therefore, we are removing the `yarn install` step from this workflow.

The action will:

1. Checkout the previous commit and install deps
2. Fingerprint that commit
3. Checkout the current commit and install deps
4. Fingerprint the current commit
5. Diff the two fingerprints

Afterwards, we will have the proper deps installed, so there is no need to install them prior to running the fingerprint action.

estrattonbailey (approved):

PR #4227, opened 2024-05-27
[🐴] send record via link in text (Record DMs - base PR)

mozzius:
# Going to stack some PRs on this one

Extracts bsky.app links from messages and converts them to embeds before sending

Screenshot 2024-05-27 at 09 45 19

### DIFF
--- a/src/screens/Messages/Conversation/MessagesList.tsx
+++ b/src/screens/Messages/Conversation/MessagesList.tsx
@@ -13,12 +13,16 @@ import {
 } from 'react-native-reanimated'
 import {ReanimatedScrollEvent} from 'react-native-reanimated/lib/typescript/reanimated2/hook/commonTypes'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
-import {AppBskyRichtextFacet, RichText} from '@atproto/api'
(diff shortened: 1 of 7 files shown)


src/screens/Messages/Conversation/MessagesList.tsx
```diff
@@ -264,20 +269,71 @@ export function MessagesList({
...
+            // so trim on both sides
+            if (postLinkFacet.index.byteStart === 0) {
+              rt = new RichText({text: rt.text.trim()}, {cleanNewlines: true})
+            } else {
+              // otherwise just trim the end
+              rt = new RichText(
+                {text: rt.text.trimEnd()},
+                {cleanNewlines: true},
+              )
+            }
```
estrattonbailey, line 326: Hadn't thought hard about this before, but we were using `trimEnd` when sending, and `trim` inside `Convo`.

Do we want to allow leading whitespace, or should we just do `trim` in all cases?

estrattonbailey (approved):
Looking great!

src/screens/Messages/Conversation/MessagesList.tsx
mozzius: I was leaning towards `trimEnd` just to match how the main composer does it, but I'm not particularly fussed which way we do it

PR #4229, opened 2024-05-27
[🐴] add emoji multiplier prop to RichText and bump it up for DMs

mozzius:
Screenshot 2024-05-27 at 10 19 27

Emojis are now 1.85 times bigger than whatever the text size was meant to be, rather than a static 26px. This should mostly result in the same size.

In DMs, the multiplier is 3x

Before
Screenshot 2024-05-27 at 10 13 05

After
Screenshot 2024-05-27 at 10 12 53

### Test plan

Check it looks ok, including on existing areas that have rich text. Make sure to test it looks ok on Windows, since their emoji can be a bit funny looking


estrattonbailey (approved):

estrattonbailey (changes requested):
Needs tweaks, my b, sec

estrattonbailey (approved):

src/components/RichText.tsx
```diff
@@ -247,3 +246,10 @@ function RichTextTag({
     </React.Fragment>
   )
 }
+
+export function isOnlyEmoji(text: string) {
+  return (
+    text.length <= 15 &&
+    /^[\p{Emoji_Presentation}\p{Extended_Pictographic}]+$/u.test(text)
```
estrattonbailey, line 253: Supports large emojis like 🧑🏻‍❤️‍💋‍🧑🏼 which is `15` and captures a few more emoji like some flags 🇿🇼

estrattonbailey: If we wanted to do this right we'd count graphemes, but unsure it's worth the perf

estrattonbailey: Correction: 🧑🏻‍❤️‍💋‍🧑🏼 is not caught by the regex, but I'll leave the limit here anyway since even something like 🐴 is length 2.

PR #4230, opened 2024-05-27
[🐴] Record message

mozzius:
# Stacked on #4227 

The post is not hydrated by the server, so we have to fetch it clientside. Hopefully we can remove this in the future.

Screenshot 2024-05-27 at 11 33 44

### DIFF
--- a/src/components/dms/MessageItemEmbed.tsx
+++ b/src/components/dms/MessageItemEmbed.tsx
@@ -1,0 +1,108 @@
+import React, {useMemo} from 'react'
+import {View} from 'react-native'
+import {
+  AppBskyEmbedRecord,
+  AppBskyFeedPost,
+  AtUri,
+  RichText as RichTextAPI,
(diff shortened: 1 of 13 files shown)


src/components/dms/MessageItemEmbed.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  if (!post || !moderation || !rt || !record) {
+    return null
+  }
+
+  const itemUrip = new AtUri(post.uri)
+  const itemHref = makeProfileLink(post.author, 'post', itemUrip.rkey)
+
+  return (
+    <Link to={itemHref}>
```
estrattonbailey, line 63: If embed content in the post is hidden, and you click on it to show, it follows this link. May need a `EventStopper` somewhere.

estrattonbailey (commented):
Looking great!

estrattonbailey (approved):

PR #4233, opened 2024-05-27
Use the proper logic on iOS to increment the badge

haileyok:
## Why

In the past when sending a notification payload, I have simply set the badge number to whatever value I wanted to _increment_ the count by, however this doesn't seem to be the "correct" way to do this according to Apple docs:

https://developer.apple.com/documentation/usernotifications/generating-a-remote-notification#Payload-key-reference

This might be something that works in Expo's case with their logic, but it isn't the correct behavior when interacting directly with the native code.

To achieve this, we need to keep track of the badge count in `UserDefaults` within the Notification Service Extension. We cannot access the app's badge count from the NSE directly, which is why we need to store it in `UserDefaults`. Then, whenever we clear the notifications within the app, we will update the `UserDefaults` value with the new count.

## Test Plan

This needs to be deployed to test unfortunately - it cannot be tested locally. That said, to test we should:

- Send some notifications to yourself
- Update the number/clear the number by opening the app and visiting the notifications tab.


haileyok:
Marking this as a draft until we can test properly. We need to add `content-mutable` to the payload for all notification types, not just `chat-message`.

gaearon (approved):
code lgtm

PR #4236, opened 2024-05-27
add prop to ListImpl for disabling `content-visibility` style

haileyok:
## Why

We recently discovered that there was some problems with the `content-visibility` style in the `ListImpl` for chat on Firefox. We originally just removed this style all together on Firefox, however that came with what seems like some performance hits. It seems like a better idea to scope this down to specific cases where it causes problems instead.

I've added a prop to `ListProps` called `disableContentVisibility`, which will simply not add the style when set to `true`.

We could scope this down further to just set this prop to `true` if the browser is Firefox, though this might not always work in cases such as:

- Spoofed user agent
- Some older Chrome version that might not have good support for this style (?)

## Test Plan

Using the `MessagesList`, you can comment out the `disableContentVisibility` prop and see that on Firefox we return to having jank on message sending. If we uncomment the prop, we see that the jank disappears.

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -25,7 +25,7 @@ export type ListProps<ItemT> = Omit<
   desktopFixedHeight: any // TODO: Better types.
   containWeb?: boolean
   sideBorders?: boolean
-  useContentVisibility?: boolean
+  disableContentVisibility?: boolean
 }
 export type ListRef = React.MutableRefObject<any | null> // TODO: Better types.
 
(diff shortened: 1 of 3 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -445,7 +450,9 @@ let Row = function RowImpl<ItemT>({
   }
 
   return (
-    <View style={styles.row} ref={rowRef}>
+    <View
+      style={disableContentVisibility ? undefined : styles.row}
```
gaearon, line 454: This is a bit confusing since it's easy to add something to `styles.row` without realizing it's conditional on this prop. Something in naming needs to bind them better together.

gaearon (approved):

gaearon (approved):

PR #4248, opened 2024-05-29
Fix `hitSlop` type

haileyok:
CI is failing typecheck because it says that `hitSlop` on a `TextInput` cannot be `null`. However, this isn't the case. See: https://github.com/facebook/react-native/blob/a93a15aca3e380f2f8158465b1b315ac3fab8307/packages/react-native/Libraries/Components/View/ViewPropTypes.d.ts#L135

This is a problem with `react-native-gesture-detector`'s type for `hitSlop` and it's version of `TextInput`. We don't use this in our codebase at all - and don't have any need to either. Nor do we use the `BottomSheetTextInput`. We use our own UI kit's text input component.

I've asserted the type for now in the `react-native-bottom-sheet` library to silence this error, so need to update the `yarn.lock`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3113,7 +3113,7 @@
 
 "@discord/bottom-sheet@bluesky-social/react-native-bottom-sheet":
   version "4.6.1"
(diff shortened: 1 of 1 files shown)


gaearon (approved):
ok

PR #4250, opened 2024-05-29
bump more expo deps

haileyok:


patches/expo-notifications+0.28.1.patch
```diff
@@ -87,32 +87,32 @@ index f1fed19..166b34f 100644
...
      mTitle = (String) in.readObject();
      mText = (String) in.readObject();
      mSubtitle = (String) in.readObject();
 @@ -240,6 +250,7 @@ public class NotificationContent implements Parcelable, Serializable {
    }
- 
+
    public static class Builder {
-+    private string mChannelId;
++    private String mChannelId;
```
haileyok, line 108: reran the patch-package here because `string` was lowercased.

PR #4254, opened 2024-05-29
Always show the header on post threads on native

haileyok:
## Why

Whenever we introduced PRs for fixing layout jank in the post thread `List`, we inadvertently broke the header always being visible. Lots users - especially on iOS - miss not having the back button easily available.

This is just a rudimentary PR that adds back that header. It doesn't animate the header in/out like we do on some other screens. We may decide to do that, in which case we'll need a different approach.

I also did a bit of StyleSheet housekeeping while I was here by using ALFredo instead of the few uses of `palette`.

## Test Plan

Open a thread, and make sure the header stays at the top while also not breaking things in larger threads. This is a great thread to test in: https://bsky.app/profile/www.mozzius.dev/post/3kjqhblh6qk2o

https://github.com/bluesky-social/social-app/assets/153161762/93fa18f6-5a9b-4163-b39c-7cd9b2a92864

Screenshot 2024-05-28 at 11 49 26 PM

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -315,7 +315,8 @@ export function PostThread({
     skeleton?.highlightedPost?.type === 'post' &&
     (skeleton.highlightedPost.ctx.isParentLoading ||
       Boolean(skeleton?.parents && skeleton.parents.length > 0))
-  const showHeader = isNative || !hasParents || !isFetching
+  const showHeader =
+    isNative || (isTabletOrMobile && (!hasParents || !isFetching))
 
   const renderItem = React.useCallback(
     ({item, index}: {item: RowItem; index: number}) => {
@@ -448,7 +449,7 @@ export function PostThread({
 
   return (
     <CenteredView style={[a.flex_1]} sideBorders={true}>
-      {isTabletOrMobile && showHeader && (
+      {showHeader && (
         <ViewHeader
           title={_(msg({message: `Post`, context: 'description'}))}
           showBorder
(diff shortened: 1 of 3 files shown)


gaearon:
web :(

Screenshot 2024-05-30 at 02 37 35

src/view/com/post-thread/PostThread.tsx
```diff
@@ -449,39 +448,49 @@ export function PostThread({
...
-            // 300 is based on the minimum height of a post. This is enough extra height for the `maintainVisPos` to
-            // work without causing weird jumps on web or glitches on native
-            height={windowHeight - 200}
-          />
-        }
-        initialNumToRender={initialNumToRender}
-        windowSize={11}
-      />
-    </ScrollProvider>
+    <CenteredView style={[a.flex_1]} sideBorders={true}>
```
haileyok, line 451: We need to wrap this in a centered view and use _those_ side borders, rather than the ones on the `List`. Unless we tweaked `List` again to add a `header` prop, which just seems a little overkill since this is the only screen we need it on.

src/view/com/post-thread/PostThread.tsx
```diff
@@ -323,117 +311,100 @@ export function PostThread({
...
+        </View>
+      )
+    } else if (item === SHOW_HIDDEN_REPLIES || item === SHOW_MUTED_REPLIES) {
+      return (
+        <PostThreadShowHiddenReplies
+          type={item === SHOW_HIDDEN_REPLIES ? 'hidden' : 'muted'}
+          onPress={() =>
+            setHiddenRepliesState(HiddenRepliesState.ShowAndOverridePostHider)
+          }
+          hideTopBorder={index === 0}
```
haileyok, line 335: This was redundant, the only prop that changes is `type`

gaearon (approved):

PR #4255, opened 2024-05-29
Change Android's MainActivity launchMode to singleTop

### LINKED ISSUE #4212: In-app browser closes when app reopened from Android launcher
**Describe the bug**

When the in-app browser (Chrome Custom Tab) is open, leaving the app and returning to it via the home screen icon causes the browser to be closed.

**To Reproduce**

1. Open a link in the app using the in-app browser
2. Switch to another app or return to the home screen
3. Reopen the app by tapping on the home screen icon

**Expected behavior**

The in-app browser remains open.

**Screenshots**

https://github.com/bluesky-social/social-app/assets/6226254/5dbb4d2c-480f-4bf9-befe-5fed77dbbcf0

**Details**

 - Platform:  Android
 - Platform version: 14, 15 Beta 2.1
 - App version: 1.83.0.190 9eed70415 (prod)

**Additional context**

I am nearly certain the issue is identical to [this StackOverflow question](https://stackoverflow.com/questions/65479634/chrome-custom-tab-is-closed-when-app-is-re-opened-via-launcher). The posted solution describes the splash screen showing very briefly when reopening the app. It's only a fraction of a second, but the splash screen's blue background can be seen flashing on the screen when Bluesky is reopened from the home screen. However, I don't know if it's possible to implement the solution in React Native or if some other workaround is required.

npmanos:
Expo sets the `launchMode` property of  `.MainActivity` to `singleTask` in `AndroidManifest.xml` by default. This causes any activity above it in the app's stack to be destroyed any time the app is launched from the home screen. Since the app is a single activity, this normally isn't a problem since there aren't any activities above it in the stack. However, the in-app browser is a different activity, so reopening the app from the home screen when the browser had been open causes it to be closed and return the user to previous screen instead. The Android developer documentation cautions against using any mode other than `standard` or `singleTop`.

I've added a config plugin to alter the AndroidManifest to use `singleTop`. This seems to have resolved the issue.

Fixes #4212 

## Testing

I have built the release variant and installed it on my Pixel 6 Pro to verify the fix works. I have not tested for all possible side effects, though I haven't encountered any issues so far. If anyone can think of something specific that should be tested before merging, I'd be happy to try it.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -204,6 +204,7 @@ module.exports = function (config) {
         ],
         './plugins/withAndroidManifestPlugin.js',
         './plugins/withAndroidManifestFCMIconPlugin.js',
+        './plugins/withAndroidManifestLaunchModePlugin.js',
         './plugins/withAndroidStylesWindowBackgroundPlugin.js',
         './plugins/withAndroidStylesAccentColorPlugin.js',
         './plugins/withAndroidSplashScreenStatusBarTranslucentPlugin.js',
--- a/plugins/withAndroidManifestLaunchModePlugin.js
+++ b/plugins/withAndroidManifestLaunchModePlugin.js
@@ -8,7 +8,7 @@ module.exports = function withAndroidManifestLaunchModePlugin(appConfig) {
       const mainActivity = mainApplication.activity.find(
         elem => elem.$['android:name'] === '.MainActivity',
       )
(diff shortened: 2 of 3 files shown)


haileyok (approved):
Builds good for me locally. Thank you! This might actually help with another RN issue I've noticed in development.

PR #4262, opened 2024-05-29
Improve the visual clarity of labels on profiles and posts

pfrazee:
In https://github.com/bluesky-social/social-app/pull/3677, we updated the label rendering to show the origin of labels. Unfortunately this has been extremely visually noisy, making the labels difficult to digest.

Some folks smartly suggested we just show the avatar, so that's what this PR does if it's from a 3p labeler. In other cases, it'll show the icon.

## 3p labelers

|.|.|
|-|-|
|CleanShot 2024-05-29 at 08 01 22@2x|CleanShot 2024-05-29 at 08 01 46@2x|

|.|.|
|-|-|
|CleanShot 2024-05-29 at 08 02 40@2x|CleanShot 2024-05-29 at 08 03 33@2x|

## Baked-in labeler

|.|.|
|-|-|
|CleanShot 2024-05-29 at 08 04 32@2x|CleanShot 2024-05-29 at 08 04 47@2x|

### DIFF
--- a/src/components/moderation/PostAlerts.tsx
+++ b/src/components/moderation/PostAlerts.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import {StyleProp, View, ViewStyle} from 'react-native'
-import {ModerationCause, ModerationUI} from '@atproto/api'
+import {BSKY_LABELER_DID, ModerationCause, ModerationUI} from '@atproto/api'
 
 import {getModerationCauseKey} from '#/lib/moderation'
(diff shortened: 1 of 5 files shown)


str4d:
If a "blur" label is applied to a labeller's account by another labeller, and a user is subscribed to both, does the avatar in the first labeller's labels get blurred?

(I know very little about React, but I like tongue twisters!)

gaearon (approved):

PR #4265, opened 2024-05-29
PFP follow button: profile shadows

estrattonbailey:
Stacked on #3560


gaearon (approved):
a bit worried about added traversals but @haileyok tested on low end android on large threads and it seemed ok. let's give this a go

PR #4267, opened 2024-05-29
[🙅] Disambiguation of the deactivation

estrattonbailey:
We want to call user account deactivation... _deactivation_. What we previously called `deactivated` will now be called `signupQueued`, so I updated everything to reflect that to avoid confusion in upcoming PRs.

Because this functionality isn't used atm, it's safe to merge this and do a clean swap without waiting for backend.


KuwaLee:
Sorry to bother you, but I encountered difficulty while translating `left to go.`  string, since I am wondering it is place left so you can enroll or place where you line in (that you have to wait).

Plus, I think make it `{placeInQueue} left to go` instead of `left to go.` will let it more translatable since it will not fix `{placeInQueue}` at front

estrattonbailey:
@KuwaLee good question! Here `placeInQueue` means the number of users ahead of you in line. This is separate from the amount of time it will take. Example: `12 left to go` means there are 12 users ahead of you.

As for the formatting, I'm not sure we can because we use separate text components to style these string values. Is that OK? I will double check with folks also and maybe update.

gaearon (approved):

PR #4268, opened 2024-05-29
Consistent localized formatting

mary-ext:
Just realized Lingui offers convenience methods for language-aware number and date-time formatting via `i18n.number()` and `i18n.date()` respectively.

This pull request switches most of our formatting to use that, except for the `ago` function, I'll touch on that in a separate pull request (haven't figured out how to make unit formatting for months and minutes to not be the same `m` suffix, I think we might need a separate logic for English, unfortunately)

This seems like the better approach because currently we're very inconsistent as to how we're dealing with localized formatting. Here's how we've been doing it so far:
- pass `undefined` as the locale
- pass locale retrieved from `expo-localization`
- pass locale from our content language settings
- enforce `en-US` locale

I've removed the `formatCountShortOnly` function as it's only used in the drawer, I'm not sure why it should be using a different formatting than the one in `<Metrics>` component, but we can add that back if necessary.

### DIFF
--- a/src/screens/Profile/Header/Metrics.tsx
+++ b/src/screens/Profile/Header/Metrics.tsx
@@ -17,9 +17,9 @@ export function ProfileHeaderMetrics({
   profile: Shadow<AppBskyActorDefs.ProfileViewDetailed>
 }) {
   const t = useTheme()
-  const {_} = useLingui()
-  const following = formatCount(profile.followsCount || 0)
-  const followers = formatCount(profile.followersCount || 0)
+  const {_, i18n} = useLingui()
+  const following = formatCount(i18n, profile.followsCount || 0)
+  const followers = formatCount(i18n, profile.followersCount || 0)
   const pluralizedFollowers = plural(profile.followersCount || 0, {
     one: 'follower',
     other: 'followers',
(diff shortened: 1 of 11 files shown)


mary-ext:
I've just noticed that some locales doesn't seem to have any compact notation formatting whatsoever, like German and Italian, I'm not sure what to do in this case, should we do a fallback to English in those scenarios? or should we just not?

src/screens/Profile/Header/Metrics.tsx
```diff
@@ -54,7 +54,7 @@ export function ProfileHeaderMetrics({
         </Text>
       </InlineLinkText>
       <Text style={[a.font_bold, t.atoms.text, a.text_md]}>
-        {formatCount(profile.postsCount || 0)}{' '}
+        {formatCount(i18n, profile.postsCount || 0)}{' '}
         <Text style={[t.atoms.text_contrast_medium, a.font_normal, a.text_md]}>
           {plural(profile.postsCount || 0, {one: 'post', other: 'posts'})}
         </Text>
```
mary-ext, line 60: Food for thought, I feel like there's a better way to do this (where we have different styling for count and suffix) in a way that's more i18n-friendly (doesn't assume prefix/suffix)

```
{count, plural,
  one {# <0>post</0>}
  other {# <0>posts</0>}
}
```

Problem is, I don't understand ICU MessageFormat enough to know if like, there's a way to format the `#` sigil.

Perhaps this works as an alternative?

```
{count, plural,
  one {{fixedCount} <0>post</0>}
  other {{fixedCount} <0>posts</0>}
}
```

This isn't much of a priority right now though.

tkusano:
In Japanese, `i18n.number()` defines an compact notation for numbers in tens of thousands, but does not seem to define  for numbers in thousands. I think it's fine to leave it as it is, without falling back to English.

PR #4272, opened 2024-05-30
Use a margin of -6 instead of -5 for PostCtrls 😵‍💫

haileyok:
## Why

Because if I zoomed in to 280% it looked a little off.


gaearon (approved):
🫡

PR #4273, opened 2024-05-30
Localized date time strings

mary-ext:
Stacked on #4268, adds proper datetime localization.

There's an issue where memoized components might not necessarily reflect, as the I18n class instance is reused across locale changes. If that's a concern, we can do another change where we'll remake the instance when changing languages.

### DIFF
--- a/src/lib/strings/time.ts
+++ b/src/lib/strings/time.ts
@@ -1,10 +1,15 @@
+import {I18n} from '@lingui/core'
+import {msg} from '@lingui/macro'
+
+import {MONTH_FALLBACK_LOCALES} from '#/locale/constants'
+
 const NOW = 5
 const MINUTE = 60
 const HOUR = MINUTE * 60
 const DAY = HOUR * 24
 const MONTH_30 = DAY * 30
 const MONTH = DAY * 30.41675 // This results in 365.001 days in a year, which is close enough for nearly all cases
-export function ago(date: number | string | Date): string {
+export function ago(i18n: I18n, date: number | string | Date): string {
   let ts: number
   if (typeof date === 'string') {
     ts = Number(new Date(date))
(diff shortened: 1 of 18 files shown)


src/locale/constants.ts
```diff
@@ -0,0 +1,4 @@
+// The narrow unit display for `month` in English is `m`, which conflicts with
+// the one for `minute` (also `m`), this also goes for any locale that currently
+// falls back to English for narrow-unit display (missing localization?)
+export const MONTH_FALLBACK_LOCALES = ['en', 'ja', 'es', 'fr', 'tr']
```
mary-ext, line 4: I've checked all the locales we currently have and these are the ones that currently falls back to English as of Firefox 128 (nightly)

I'd have to recheck for Chrome as they could be on different versions of ICU. Curious how it'll be for Native as well.

tkusano: Since `i18n.number` does not work as expected for Japanese period expressions, it would be better to patch `ago()` as in #3905.

mary-ext: Hmm, you might be right, I was hoping I could avoid that since this approach would mean avoiding adding strings for something that the browser should already have (in its ICU data). I'll change it in just a bit

src/lib/strings/time.ts
```diff
@@ -33,23 +54,26 @@ export function ago(date: number | string | Date): string {
     }
 
     if (months < 12) {
-      return `${months}mo`
+      if (MONTH_FALLBACK_LOCALES.includes(i18n.locale)) return `${months}mo`
+
+      return i18n.number(months, {
+        style: 'unit',
+        unitDisplay: 'narrow',
+        unit: 'month',
```
tkusano, line 62:
```suggestion
      return i18n._(msg`${months}mo`)
```
...or, how about this?

src/locale/constants.ts
mary-ext: Replaced with strings, means that JP and other locales listed there can have proper localization for relative time

src/lib/strings/time.ts
```diff
@@ -13,17 +16,57 @@ export function ago(date: number | string | Date): string {
...
+      }),
+    )
   } else if (diffSeconds < MONTH_30) {
-    return `${Math.round(diffSeconds / DAY)}d`
+    const days = Math.round(diffSeconds / DAY)
+
+    return i18n._(
+      defineMessage({
+        message: `${days}d`,
+        comment: `How many minutes has passed, displayed in a narrow form`,
```
tkusano, line 67:
```suggestion
        comment: `How many days has passed, displayed in a narrow form`,
```

tkusano:
Tested on Firefox in Japanese. Looks good.

quiple:
The datetime strings seem to have been replaced by `useTimeAgo.ts`.

mary-ext:
Yeah I'll have to rebase this branch, should be able to get around to it on weekend

PR #4281, opened 2024-05-30
Remove copy post text option

mozzius:


mozzius:
Maybe come back to this later

PR #4290, opened 2024-05-30
[🙅] Add deactivate account dialog

estrattonbailey:
Adds the dialog as its own button on the Settings screen, and accessible from within the delete account dialog.

CleanShot 2024-05-31 at 14 48 21@2x
CleanShot 2024-05-31 at 15 05 15@2x
CleanShot 2024-05-31 at 15 05 12@2x
CleanShot 2024-05-31 at 15 03 14@2x
CleanShot 2024-05-31 at 14 54 01@2x

### DIFF
--- a/src/screens/Settings/components/DeactivateAccountDialog.tsx
+++ b/src/screens/Settings/components/DeactivateAccountDialog.tsx
@@ -1,27 +1,59 @@
 import React from 'react'
-import {msg} from '@lingui/macro'
+import {View} from 'react-native'
+import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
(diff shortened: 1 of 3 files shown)


surfdude29:
I realise it's a draft PR and the copy will change but I have one quick suggestion for you to consider – if feeds and lists from the account will also no longer be visible after a user has deactivated (which I'm assuming will be the case?) I think it would be helpful to mention that here as well:

https://github.com/bluesky-social/social-app/blob/eda34e61d3b8b87e6096370d93830b2182a84768/src/screens/Settings/components/DeactivateAccountDialog.tsx#L25-L26

gaearon (approved):

PR #4294, opened 2024-05-31
Change many border widths from `1` to `hairlineWidth`

haileyok:
## Why

Note: We are changing https://github.com/bluesky-social/social-app/pull/2090 to round to 0.5 instead, since hairline is 0.5 pixels.

Hairline width looks a lot better in many places.

I also adjusted a few more things while going through the codebase, see individual comments for details.

## Test Plan

Honestly, I think I've hit everything visually by clicking through the app on a simulator, though I'm not certain. Will need to continue using it.


src/view/screens/Notifications.tsx
```diff
@@ -145,7 +146,11 @@ export function NotificationsScreen({}: Props) {
 
   return (
     <View testID="notificationsScreen" style={s.hContentRegion}>
-      <ViewHeader title={_(msg`Notifications`)} canGoBack={false} />
+      <ViewHeader
+        title={_(msg`Notifications`)}
+        canGoBack={false}
+        showBorder={true}
```
haileyok, line 152: had to add border here, since previously it was on the top of the post item.

src/view/com/posts/FeedSlice.tsx
```diff
@@ -27,6 +33,7 @@ let FeedSlice = ({slice}: {slice: FeedPostSlice}): React.ReactNode => {
           moderation={slice.items[0].moderation}
           isThreadParent={isThreadParentAt(slice.items, 0)}
           isThreadChild={isThreadChildAt(slice.items, 0)}
+          hideTopBorder={hideTopBorder}
```
haileyok, line 36: In feeds, we use `FeedSlice` instead of `FeedItem` directly. So, we need to tell the first item in the feed slice to not display the border _if_ it is also the first item in `renderItem`.

src/view/com/posts/FeedSlice.tsx
```diff
@@ -75,6 +82,7 @@ let FeedSlice = ({slice}: {slice: FeedPostSlice}): React.ReactNode => {
           isThreadLastChild={
             isThreadChildAt(slice.items, i) && slice.items.length === i + 1
           }
+          hideTopBorder={hideTopBorder && i === 0}
```
haileyok, line 85: Same here

src/view/com/notifications/Feed.tsx
```diff
@@ -170,6 +181,8 @@ export function Feed({
         contentContainerStyle={s.contentContainer}
         // @ts-ignore our .web version only -prf
         desktopFixedHeight
+        initialNumToRender={initialNumToRender}
+        windowSize={11}
```
haileyok, line 185: We were missing some `FlatList` perf props here. Removed the above comment after adding this, since I wasn't even getting the warning previously anyway.

gaearon (approved):
need to fix

- search/hashtags (between posts)
- feeds page
- lists page header
- individual list page header
- mod list page header

i'm not sure i dig them on account switcher but this may be a spacing issue. not sure.

overall looks great to me

gaearon (approved):
lets gooooo

PR #4295, opened 2024-05-31
Disable non-deterministic flaky test

gaearon:
Skip test to unbreak main.

I tried fixing it properly but couldn't figure out how/why. This is a bit worrying. However it's still failing if I go back in time on `main` so this either points to the test already being flawed in some way, or at the implementation already having an issue. It's definitely worth another look but I need to get the CI fixed.

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -142,7 +142,8 @@ describe('makeRecordUri', () => {
   })
 })
 
-describe('ago', () => {
+// FIXME: Reenable after fixing non-deterministic test.
+describe.skip('ago', () => {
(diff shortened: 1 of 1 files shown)


haileyok (approved):

janaagaard75:
> Ah fantastic. Would you like to work on this change?

Sure. I will try to make a PR within the next week.

gaearon:
Thanks!

janaagaard75:
The pull request is here: https://github.com/bluesky-social/social-app/pull/4364

Unsure if it's acceptable to change the "months ago" calculation to always use 30 days, but I think this was the best tradeoff. Let me know if you prefer keeping the 30.41675 days.

PR #4297, opened 2024-05-31
remove resolution from post thread

haileyok:
## Why

In the continuous effort to remove waterfalls, we can now remove these requests from `getPostThread`. To do so however, we need to do a little bit of reworking on how we pull already-loaded posts from the cache.

Currently we load the posts from the cache just based off of their URI. However, since we now are not going to have the resolved URI, we instead need to check both the handle and the rkey instead in our various `findAllPostsInQueryData` functions. This is a pretty straight forward change, though it adds a little extra logic.

## Test Plan

Note in the debugger that there are no more requests for DID resolutions when opening a post on a fresh load. Also note that whenever you open a post from these areas, there is no spinner:

- Feed
- Profile
- Search page
- Notifications


mozzius (approved):
good stuff

src/state/queries/notifications/feed.ts
```diff
@@ -142,20 +142,37 @@ export function* findAllPostsInQueryData(
...
         }
+
         const quotedPost = getEmbeddedPost(item.subject?.embed)
-        if (quotedPost?.uri === uri) {
+        if (isDid && quotedPost?.uri === uri) {
+          yield embedViewRecordToPostView(quotedPost)
+        } else if (
+          !isDid &&
+          quotedPost?.author.handle === atUri.host &&
+          quotedPost.uri.endsWith(atUri.rkey)
```
gaearon, line 174: This feels complicated. Ok but wondering if there are easier ways. I do think we need to pull this into a utility though cause the copy pasta is not obvious.

gaearon: Let's at least extract `isSamePostUri` or something

gaearon (changes requested):
- let's reduce duplication of the URI check (it's non-trivial)
- placeholders are broken

haileyok:
@gaearon can you check now if placeholders work? I can't repro the spinner even off the first commit and still can't, so not sure what happened there? Left a video on Slack.

gaearon (approved):

PR #4300, opened 2024-05-31
[🐴] Embed records without removing them for backwards compat (DO NOT MERGE)

mozzius:
Intention is to send this branch as OTA ahead of the main release, so embeds start being generated


estrattonbailey:
Let's go with #4302 for right now

PR #4308, opened 2024-05-31
[🙅] Integrate deactivate

estrattonbailey:
- [x] double check migration, since we edited schema
- [x] double check 2fa
- [x] test with an app password


gaearon (approved):
amazing

PR #4309, opened 2024-05-31
invert the fab animation, play a haptic

haileyok:
## Why

While implementing a fab elsewhere, I realized that it's actually pretty hard to see the FAB animation right now. The fab is decreasing in size, so if your finger is over it you can't see the animation. Twitter's FAB does shrink as well, though to me it's still bizarre, since you can't actually see the animation at all. This PR inverts the animation, so it actually scales outward rather than inward.

I also added a small delay to the opening of the modal itself. There is still a haptic that immediately played so there's immediate feedback, but it gives us the chance to visually see the button shrink back to its original size. This delay does not occur if the user has haptic feedback disabled.

Additionally, I've added a haptic to the FAB (a light haptic though, not the default medium). The haptic gets you "excited to post" I think, though we can remove if we hate it. On device for me, it feels really nice.

The `Light` feedback here felt much better than the default `Medium`, so I updated `useHaptics()` to support passing in the wanted style - `Light`, `Medium`, or `Heavy`. On iOS, this defaults to `Medium`. On Android, it uses `Light` regardless of the passed style.

>  // Users said the medium impact was too strong on Android; see APP-537s

## Test Plan

On-device, you should be able to feel the light haptic once you press the button.

https://github.com/bluesky-social/social-app/assets/153161762/66644445-2caa-4686-b775-9cf5b8bcc009

### DIFF
--- a/src/lib/haptics.ts
+++ b/src/lib/haptics.ts
@@ -4,17 +4,21 @@ import {impactAsync, ImpactFeedbackStyle} from 'expo-haptics'
 import {isIOS, isWeb} from 'platform/detection'
 import {useHapticsDisabled} from 'state/preferences/disable-haptics'
 
-const hapticImpact: ImpactFeedbackStyle = isIOS
-  ? ImpactFeedbackStyle.Medium
-  : ImpactFeedbackStyle.Light // Users said the medium impact was too strong on Android; see APP-537s
-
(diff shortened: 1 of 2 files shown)


mozzius (approved):
I need to see how it feels on device

PR #4312, opened 2024-06-01
Add android on-device translation

fobos531:
Builds on top of https://github.com/bluesky-social/social-app/pull/4098

This PR aims to add on-device translation to Android via MLKit's [Translation API](https://developers.google.com/ml-kit/language/translation) to achieve feature parity with iOS. From my testing, Google's 1st party service (translate.google.com) still is the best, but if the aim is to have feature parity on Android, then I think this is a satisfactory solution.

Do note that translation _could_ break rich text formatting (link detection etc) when displaying the translated text, but that may even be expected when translating - the user could revert to the original text at any time if needed.
 
I'm not super familiar with the codebase, so I might have not done everything in the best possible way so bear with me on that, open to any feedback.

Todo: Add better Google attribution (according to their guidelines)

Quick preview:

https://github.com/bluesky-social/social-app/assets/19533289/0767fead-64d0-4481-a330-027f5b9fd02b

### DIFF
--- a/modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
+++ b/modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
@@ -1,0 +1,65 @@
+package expo.modules.blueskytranslate
+
+import com.google.mlkit.common.model.DownloadConditions
+import com.google.mlkit.nl.translate.Translation
+import com.google.mlkit.nl.translate.TranslatorOptions
+import expo.modules.kotlin.modules.Module
+import expo.modules.kotlin.modules.ModuleDefinition
+import expo.modules.kotlin.Promise
+
+class ExpoBlueskyTranslateModule : Module() {
+    var translationPromise: Promise? = null
+
+    override fun definition() = ModuleDefinition {
+        Name("ExpoBlueskyTranslate")
+
+        AsyncFunction("translateAsync") { sourceLanguage: String, targetLanguage: String, text: String, promise: Promise ->
+            if (translationPromise != null) {
+                promise.reject(TranslationInProgressException())
+                return@AsyncFunction
+            }
+
+            translationPromise = promise
+
+            val options = TranslatorOptions.Builder()
+                .setSourceLanguage(sourceLanguage)
+                .setTargetLanguage(targetLanguage)
+                .build()
+
+            val translator = Translation.getClient(options)
+
+            val conditions = DownloadConditions.Builder()
+                .requireWifi()
+                .build()
+
+
+            translator.downloadModelIfNeeded(conditions)
+                .addOnSuccessListener {
+                    translator.translate(encodeSpecialCharacters(text))
+                        .addOnSuccessListener { translatedText ->
+                            promise.resolve(decodeSpecialCharacters(translatedText))
+                            translationPromise = null
+                        }
+                        .addOnFailureListener { exception ->
+                            promise.reject(UnableToTranslateTextException())
+                            translationPromise = null
+                        }
+                }
+                .addOnFailureListener { exception ->
+                    promise.reject(UnableToTranslateTextException())
+                    translationPromise = null
+                }
+
+        }
+    }
+
+  private fun encodeSpecialCharacters(text: String): String {
+    return text.replace("\n", "␤")  // Use Unicode character for newline
+      .replace("\t", "␉")  // Use Unicode character for tab
+  }
+
+  private fun decodeSpecialCharacters(text: String): String {
+    return text.replace("␤", "\n")
+      .replace("␉", "\t")
+  }
+}
--- a/modules/expo-bluesky-translate/index.ts
+++ b/modules/expo-bluesky-translate/index.ts
@@ -1,6 +1,66 @@
+import {Platform} from 'react-native'
+
+import ExpoBlueskyTranslate from './src/ExpoBlueskyTranslate'
+
+// can be something like "17.5.1", so just take the first two parts
+const version = String(Platform.Version).split('.').slice(0, 2).join('.')
+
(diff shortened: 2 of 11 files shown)


modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
```diff
@@ -0,0 +1,65 @@
...
+
+class ExpoBlueskyTranslateModule : Module() {
+    var translationPromise: Promise? = null
+
+    override fun definition() = ModuleDefinition {
+        Name("ExpoBlueskyTranslate")
+
+        AsyncFunction("translateAsync") { sourceLanguage: String, targetLanguage: String, text: String, promise: Promise ->
+            if (translationPromise != null) {
+                promise.reject(TranslationInProgressException())
```
fobos531, line 18: Just a safeguard if one forgets to use async/await on JS side

modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
```diff
@@ -0,0 +1,65 @@
...
+
+            val options = TranslatorOptions.Builder()
+                .setSourceLanguage(sourceLanguage)
+                .setTargetLanguage(targetLanguage)
+                .build()
+
+            val translator = Translation.getClient(options)
+
+            val conditions = DownloadConditions.Builder()
+                .requireWifi()
```
fobos531, line 32: models are around 30MB, so the documentation advises to use WiFi before downloading the models

modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
```diff
@@ -0,0 +1,65 @@
...
+                .build()
+
+
+            translator.downloadModelIfNeeded(conditions)
+                .addOnSuccessListener {
+                    translator.translate(encodeSpecialCharacters(text))
```
fobos531, line 38: translation will only ever be performed if the model is downloaded first

modules/expo-bluesky-translate/android/src/main/java/expo/modules/blueskytranslate/ExpoBlueskyTranslateModule.kt
```diff
@@ -0,0 +1,65 @@
...
+                .addOnFailureListener { exception ->
+                    promise.reject(UnableToTranslateTextException())
+                    translationPromise = null
+                }
+
+        }
+    }
+
+  private fun encodeSpecialCharacters(text: String): String {
+    return text.replace("\n", "␤")  // Use Unicode character for newline
```
fobos531, line 57: in cases where the posts contain formatted text, this is a best-effort attempt to preserve this formatting after translation (the translator will strip whitespace characters otherwise)

modules/expo-bluesky-translate/index.ts
```diff
@@ -1,6 +1,66 @@
...
+import ExpoBlueskyTranslate from './src/ExpoBlueskyTranslate'
+
+// can be something like "17.5.1", so just take the first two parts
+const version = String(Platform.Version).split('.').slice(0, 2).join('.')
+
+export const isAvailableIOS = Number(version) >= 17.4
+export const isAvailableAndroid = true
+
+// https://en.wikipedia.org/wiki/Translate_(Apple)#Languages
+const SUPPORTED_LANGUAGES = [
```
fobos531, line 12: kept this array since Google's MLKit solution supports all of these, so we use the lowest common denominator in this case (Apple's translation service)

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -86,14 +90,18 @@ export function PostThreadItem({
   hideTopBorder?: boolean
 }) {
   const postShadowed = usePostShadow(post)
+
+  const [translatedText, setTranslatedText] = useState<string>(record.text)
```
fobos531, line 94: not sure if you prefer a different approach here other than useState and prop drilling

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -661,28 +673,68 @@ function ExpandedPostDetails({
...
+        NativeTranslationModule.presentAsync(text)
+      } else if (isAndroid) {
+        if (didTranslate) {
+          setTranslatedText(text)
+          setDidTranslate(false)
+        } else {
+          try {
+            setIsTranslating(true)
+            const translated = await NativeTranslationModule.translateAsync(
+              getPostLanguage(post) ?? 'en',
```
fobos531, line 711: worth noting that MLKit also has a "language identification" API that could be used to get the post language, but I figured its better to use the ATProto's API here

fobos531:
cc @mozzius

modules/expo-bluesky-translate/index.ts
danninov: Just a nitpick, but I think it's better to have sorted alphabetically.

mozzius:
Wow, amazing! I’ll take a deeper look on Monday, but this is great! Thanks!

With regards to rich text, the best way would probably be to re-detect the facets using our RichText API

modules/expo-bluesky-translate/index.ts
mozzius: Specifically opted to use Wikipedia’s order, which is by date added, so when they add new ones we can just add to the bottom.

Fwiw there’s no reason Android needs to limit itself to this!

haileyok:
as is tradition:

🚨🚨 S-tier PR Alert! 🚨🚨

ghost:
what happened here? or better why did nothing happen?

PR #4313, opened 2024-06-01
Fix code2 of several languages in LANGUAGES

tkusano:
`code2` of Luo and Punjabi entries in `LANGUAGES` containes invalid data.

PR #4314, opened 2024-06-01
Remove B code (ISO 639-2)

KuwaLee:
After seening https://github.com/bluesky-social/social-app/pull/4313, I found that language code contains both B and T codes (`ISO 639-2`), which means there are many duplicated langusge code, for example: 

`{code3: 'fra', code2: 'fr', name: 'French'}` 
`{code3: 'fre', code2: 'fr', name: 'French'}`

 It should use one standard, B or T, not both. (And UI fail to show both, I think it might because  `code2` property is duplicated )

According to [wikipedia](https://en.wikipedia.org/wiki/ISO_639-2#B_and_T_codes) T code are more favored, thus this PR remove B language code with [this reference](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes).
image

Plus, I would like to know if its support to use region in `code2` for example `zh-CN` and `zh-TW,` or it stick strictly only `ISO 639-2`?


KuwaLee:
And I find the ultimate answer why feed with lang filter malfunction, after removing duplicated language code, it work with chinese (which `chi/zho` are duplicated for `zh`)

KuwaLee:
This did fix to some degree in lang filter malfunction (For example Chinese which chi/zho is duplicated for zh), tested by me and @auroursa 

Before:
image
After:
image

surfdude29:
I think the goal of this PR is good but I think the current order of languages that the user sees in the interface should be retained.

For example, a user looking for Dutch; Flemish would expect to find it under D not N (because even though the list is sorted by `code3`, those correspond in large part to the language name in English).

So I suggest instead that the `code3` in the B code entry should be replaced with the equivalent T code, with the current list order maintained.

surfdude29:
To illustrate what I mean, instead of being deleted line 65 would become:
`{code3: 'sqi', code2: 'sq', name: 'Albanian'},`
and line 504/486 would be deleted instead

KuwaLee:
@surfdude29 by default it seems it order by `code3`, and I think it might need to discuss with bluesky team to reorder by `name` or reorder during runtime.

surfdude29:
> @surfdude29 by default it seems it sorted by `code3`, and I think it might need to discuss with bluesky team to reorder by `name` or reorder during runtime.

@KuwaLee Ah OK, I think that change should probably be merged before this one then as otherwise it could cause quite a bit of confusion if languages seem to disappear from the list.

Also, one other potential issue that occurred to me: for users that have one of these deleted `code3` languages set as their Primary Language or one of their Content Languages, how will this PR affect them? Will their language settings be automatically migrated?

KuwaLee:
@surfdude29 AFAIK bluesky seems are using code2 only now

surfdude29:
> (And UI fail to show both, I think it might because code2 property is duplicated )

Both are shown in the Primary Language dropdown for me (but not in the Content Languages dialog), so this PR would fix that issue too.

KuwaLee:
@surfdude29 Primary Language are only use when translating post

surfdude29:
> @surfdude29 AFAIK bluesky seems are using code2 only now

@KuwaLee code3 can be used for determining what languages are shown in feeds and for the Primary Language setting, right? Or else, why are they available as options? 🤔

KuwaLee:
@surfdude29 no, `code2` are actually added to post , and what used to filter post in feed also `code2`

surfdude29:
> @surfdude29 Primary Language are only use when translating post

@KuwaLee Yes, but I was meaning that this PR would fix the issue of languages being displayed twice in the UI in the Primary Language dropdown :)

For example:
IMG_6073

KuwaLee:
@surfdude29 that is simply because it have 2 object, Primary Language is used for parameters when calling google translation

And  what lang of post is what you choose when you posting, And they really added only code2
image
image
image

surfdude29:
@KuwaLee I apologise for my confusion, it seems I totally misunderstood how languages work on Bluesky, oops 😬

PR #4315, opened 2024-06-01
Unify profile tabs and lists screens placeholders

Simek:
# Why

Spotted that profile tabs and lists screen placeholders appearance varies. 

Also wanted to address an additional border on mobile platforms when refreshing an empty feed.

# How

Reuse `EmptyState` component inside profile tabs feeds and lists screen as a placeholder content. Go through all places which used the component before, to make sure that result after changes matches the appearance on PROD. Preform minor spacing tweaks.

I have also spotted that `EmptyState` relies on FA icons, with one custom icon introduced. I was wondering if similar exception should be done for "Feeds" and "Lists" placeholder icons to match them with one used in the drawer, but for the sake of changeset simplicity I have left it as-is. Happy to made an update or follow up PR, if you think it would a better for the presentation.

# Preview

Screenshot 2024-06-01 at 17 19 55

Screenshot 2024-06-01 at 17 47 21

### DIFF
--- a/src/view/com/feeds/ProfileFeedgens.tsx
+++ b/src/view/com/feeds/ProfileFeedgens.tsx
@@ -7,7 +7,7 @@ import {
   View,
   ViewStyle,
 } from 'react-native'
-import {msg, Trans} from '@lingui/macro'
+import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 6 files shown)


haileyok (approved):
Looks great, thank you!

We're working our way through removing FontAwesome, so these icons are fine! Good spot.

PR #4319, opened 2024-06-02
Update Spanish Localization

brodieavoult:
I've made some progress but it's still not complete yet. I'm going to continue working on this.


pfrazee:
Awesome. When you're ready just get another spanish speaker to +1

oledfish:
I would like to help translate the missing strings for the Spanish localization. I started working from the most recent .po file until I saw this open PR, so now I'm not sure how to proceed, or if the merge conflict from this file can be solved easily, as many strings have been added since. As for the PR itself, the changes here look good, especially the shift from "muteado/a" to "silenciado/a" and not translating "feeds" as "noticias/canales de noticias", which is confusing as it alludes to a type of content that is distinct from posts.

Issue #4301, opened 2024-05-31
Inserting a https-link in the web UI changes the link into the page title


jgarplind:
**Describe the bug**

When pasting a full link (e.g. https://reactrouter.com/en/main/upgrading/future#future-flags) in the post composer, the link text is not pasted, instead the page title appears, and there is no obvious "linkiness" such as a blue color or a link card implying that things are in order.

**To Reproduce**

1. Grab yourself a nice link, such as https://reactrouter.com/en/main/upgrading/future#future-flags
2. Open the post composer
3. Paste the link

**Expected behavior**

The link should be pasted as-is.

**Screenshots**

Bad: 
![image](https://github.com/bluesky-social/social-app/assets/10069144/9ad41b4b-446c-4610-9e08-8c5d8ad9fe14)

Good (see below for how to achieve):
![image](https://github.com/bluesky-social/social-app/assets/10069144/0bd732f7-f4b1-42b2-abb7-19c3626efd7b)

**Details**

 - Platform: Desktop Windows Edge
 - Platform version: Windows 11, Edge 125.0.2535.67
 - App version: Build version: 1.84.0; Bundle info: e6d2618 (prod); Bundle date: 24052420; Platform: web

**Additional context**

Simple enough workaround:
1. Split the link into 2 parts: "h" and "ttps://......"
2. Paste them individually
3. Ta-da, expected behavior

mary-ext:
seems to me the browser is copying *both* the URL and the page title, and because currently the composer isn't a plain text editor but rather a WYSIWYG editor pretending to be plain text, it has decided to paste the page title instead (in an attempt to get it to be a link)

I feel like we've had this issue before and the outcome was that we just don't know how to fix that yet, could you try seeing if Ctrl+Shift+V would give you the URL?

mary-ext:
> I feel like we've had this issue before

Ah yes! This is a duplicate of #1321

jgarplind:
Hey @mary-ext, thanks for getting back!

I'll close this and continue the discussion in the original issue.

Issue #3032, opened 2024-02-29
Can't create links to certain websites in new posts


qntm:
**To Reproduce**

1. In a desktop web browser (I'm using Microsoft Edge) go to a web page for which a link card could be created, such as https://github.com/bluesky-social.
2. Copy the URL out of the address bar.
3. Go to Bluesky, start creating a new post and paste the copied URL.

In this case, the text pasted into the new post is as follows:

> bluesky-social ([github.com)](https://github.com)

Note the hyperlink. Bluesky then provides a prompt, "Add a link card: [github.com](https://github.com)". Clicking on this prompt adds a link card to https://github.com - which is _not_ the specific page I wanted to link to.

Now try this with a Wikipedia page, such as https://en.wikipedia.org/wiki/Bluesky_(social_network). In this case, the pasted text is simply:

> Bluesky (social network) - Wikipedia

No hyperlink, no URL, no link card prompt, nothing. Linking to Wikipedia doesn't work at all.

**Expected behavior**

If I paste a hyperlink, the pasted text should be the URL and the link card prompt should create a link to that page (as opposed to some other page).

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/1735353/d5b2a762-89ab-4a30-af14-98301b561350)

![image](https://github.com/bluesky-social/social-app/assets/1735353/951163f5-1018-48b1-9193-8471cd468edd)

**Details**

 - Platform: Microsoft Edge on Windows

**Additional context**

This is probably happening because of some kind of special clipboard functionality either in Edge or in Bluesky, which replaces the copied string "https://github.com/bluesky-social" with a more advanced object with a special stringified value of some kind. When pasted, the Bluesky post does not receive the pasted URL, it receives the stringified value, which sometimes contains no URL at all, and sometimes contains a short URL for the website's main page, which is the wrong URL.

A workaround is to paste the copied URL into a text editor and then copy and paste it out of the text editor into the new post.

Zero3K:
This is a duplicate of https://github.com/bluesky-social/social-app/issues/1321. Please take a look at it for more information.

mary-ext:
I'll check in on this in a bit, will have to see if the clipboard data contains some additional metadata

Zero3K:
https://stackoverflow.com/questions/67560944/how-to-force-browser-to-paste-url-and-not-page-title-in-edge

Issue #1321, opened 2023-08-29
Issue when pasting a copied URL from the address bar of a browser


Zero3K:
When pasting a copied URL from the address bar of a browser into a new post, instead of the URL being inserted (like in Twitter), the title of the page and domain of the URL appear.

Zero3K:
If this is a bug, please mark it as such.

pfrazee:
Can you show a screenshot or video or give more details please?

Zero3K:
https://github.com/bluesky-social/social-app/assets/6194332/db9e87e8-b894-4c9c-9132-2ebb29f3cab5

pfrazee:
Okay thanks for the video. I decided to delete it so we don't share your desktop any longer than necessary.

I'm not sure what's causing that, but that's absolutely not bluesky. It might be a feature of your browser, your OS, or some software you installed on your desktop. As far as I know, our app wouldn't be able to produce that information even if we wanted to.

Sorry I can't help, good luck

Zero3K:
It doesn't happen in Twitter though.

Zero3K:
This should fix it:

https://stackoverflow.com/questions/67560944/how-to-force-browser-to-paste-url-and-not-page-title-in-edge

Zero3K:
What do you think?

pfrazee:
I'll be darned, you might be right. I apologize for dismissing this prematurely. I'll look into it.

Zero3K:
Any news?

Zero3K:
@estrattonbailey

mschwendener:
Just cosmetics, but maybe it coudl also be fixed: 

Chrome on Mac, first time I see this today: in the composer, copying an URL shows the last letter of the URL black, as if it was unliked. 

<img width="505" alt="screenshot_4499" src="https://github.com/bluesky-social/social-app/assets/4497705/ba6d3bf6-d5b7-4e01-938b-baf6713ef95f">

After posting, the links look normal https://bsky.app/profile/manuschwendener.bsky.social/post/3k6bgsmtfzo2v

mschwendener:
Found https://github.com/bluesky-social/social-app/issues/1293#issuecomment-1694682160, which seems related, but is closed 

"whether or not the underlining happens doesn't actually reflect what will be sent on the post"

mschwendener:
To reproduce (Chrome on Mac): 

[...]

PR #4327, opened 2024-06-03
Prevent rich-formatting paste

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/1321
Fixes https://github.com/bluesky-social/social-app/issues/3032
Fixes https://github.com/bluesky-social/social-app/issues/4301

We don't actually support rich WYSIWYG formatting in the composer, check if what's being pasted is one, and retrieve a plain-text version of it. This won't interfere with pasting HTML files (even though that's not supported either) as pasted files only have `Files` in the types array.

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -156,12 +156,13 @@ export const TextInput = React.forwardRef(function TextInputImpl(
           if (clipboardData) {
             if (clipboardData.types.includes('text/html')) {
               // Rich-text formatting is pasted, try retrieving plain text
-              event.preventDefault()
-
               const text = clipboardData.getData('text/plain')
 
-              // `pasteText` will invoke this handler again, but `clipboardData` will be empty.
+              // `pasteText` will invoke this handler again, but `clipboardData` will be null.
               view.pasteText(text)
+
+              // Return `true` to prevent ProseMirror's default paste behavior.
+              return true
             } else {
               // Otherwise, try retrieving images from the clipboard
 
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -150,16 +150,26 @@ export const TextInput = React.forwardRef(function TextInputImpl(
         attributes: {
           class: modeClass,
         },
-        handlePaste: (_, event) => {
-          const items = event.clipboardData?.items
+        handlePaste: (view, event) => {
+          const clipboardData = event.clipboardData
 
-          if (items === undefined) {
-            return
+          if (clipboardData) {
+            if (clipboardData.types.includes('text/html')) {
+              // Rich-text formatting is pasted, try retrieving plain text
+              event.preventDefault()
+
+              const text = clipboardData.getData('text/plain')
+
+              // `pasteText` will invoke this handler again, but `clipboardData` will be empty.
+              view.pasteText(text)
+            } else {
+              // Otherwise, try retrieving images from the clipboard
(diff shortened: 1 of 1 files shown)


mary-ext:
Need others to test this

gaearon:
would be good to verify that the fix works, then we can pull it on.

auroursa:
It seems to partially resolve the issue, but the link preview is not working (I'm not sure if this is the expected behavior). By the way, the link I tested is https://github.com/bluesky-social, and my Edge version is 125.0.2535.85.
屏幕截图 2024-06-04 105704

src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -150,16 +150,26 @@ export const TextInput = React.forwardRef(function TextInputImpl(
...
         },
-        handlePaste: (_, event) => {
-          const items = event.clipboardData?.items
+        handlePaste: (view, event) => {
+          const clipboardData = event.clipboardData
 
-          if (items === undefined) {
-            return
-          }
+          if (clipboardData) {
```
boly38, line 156: invert condition to adopt "return early" pattern and this to avoid one extra indent : for here, leave quickly when clipboardData is not defined (but IDK if it's possible to have undefined clipboardData on handlePaste event? doubleCheck ^^)

src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -150,16 +150,26 @@ export const TextInput = React.forwardRef(function TextInputImpl(
...
+
+              // `pasteText` will invoke this handler again, but `clipboardData` will be empty.
+              view.pasteText(text)
+            } else {
+              // Otherwise, try retrieving images from the clipboard
 
-          getImageFromUri(items, (uri: string) => {
-            textInputWebEmitter.emit('photo-pasted', uri)
-          })
+              getImageFromUri(clipboardData.items, (uri: string) => {
```
boly38, line 168: in legacy version, `clipboardData.items` where checked by a if that I cant see here: what if items is(are) undefined ?

boly38 (commented):
just some little note while reviewing your diff

src/view/com/composer/text-input/TextInput.web.tsx
mary-ext: `items` is not undefined, the previous check did it because of optional chaining in `event.clipboardData`, which is nullable.

mary-ext:
> It seems to partially resolve the issue

I'll take another look tomorrow to see if there's anything that can be done

Zero3K:
I wonder how Twitter/Upwork does it. Maybe someone could look into that and then use either of their methods.

mary-ext:
Looks like this is happening because `.preventDefault()` doesn't seem to prevent pasting, for some reason

mary-ext:
Yup. https://github.com/ProseMirror/prosemirror-view/blob/9da2e4ed026971e39555336d176bab7047df2529/src/input.ts#L620

```
// Rich-text formatting is pasted, try retrieving plain text
const text = clipboardData.getData('text/plain')

// `pasteText` will invoke this handler again, but `clipboardData` will be null.
view.pasteText(text)

// Return `true` to prevent ProseMirror's default paste behavior.
return true
```

mary-ext:
@auroursa Can you verify this again? I've confirmed that it works as expected on my end.

auroursa:
@mary-ext Yeah it works very well for me.

gaearon (approved):
didn't test Edge but this does result in more sensible paste output on macOS when pasting HTML from random places, let's give this a try.

PR #4328, opened 2024-06-03
Native `formSheet` for GIF select on iOS

mozzius:
Keeps existing ALF dialog for other platforms, just forks for iOS to provide a more native experience

https://github.com/bluesky-social/social-app/assets/10959775/7af7b1b7-ee34-412a-8c37-0ec5d8bf5d62

### Test plan

Make sure there's no regressions
- Selecting GIFs
- Searching
- Search errors
- Other errors (error boundary stuff)

### DIFF
--- a/src/components/dialogs/GifSelect.ios.tsx
+++ b/src/components/dialogs/GifSelect.ios.tsx
@@ -5,13 +5,11 @@ import React, {
   useRef,
   useState,
 } from 'react'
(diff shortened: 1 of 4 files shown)


gaearon (approved):
the duplication is a bit annoying and we can factor more shared stuff out but seems fine for now, let's see if it turns out a pita or not

PR #4339, opened 2024-06-03
[🐴] Post embeds polish

estrattonbailey:
This PR adds a few niceties on top of the existing handling:
- only strips the post link if it's at the start or end of the message string
- handles the last message view in the chat list

### DIFF
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -3,6 +3,7 @@ import psl from 'psl'
 import TLDs from 'tlds'
 
 import {BSKY_SERVICE} from 'lib/constants'
+import {isInvalidHandle} from 'lib/strings/handles'
 
 export const BSKY_APP_HOST = 'https://bsky.app'
 const BSKY_TRUSTED_HOSTS = [
@@ -81,6 +82,10 @@ export function toShareUrl(url: string): string {
     url = urlp.toString()
   }
   return url
+}
+
+export function toBskyAppUrl(url: string): string {
+  return new URL(url, BSKY_APP_HOST).toString()
 }
 
 export function isBskyAppUrl(url: string): boolean {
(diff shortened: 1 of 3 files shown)


src/lib/strings/url-helpers.ts
```diff
@@ -183,6 +188,22 @@ export function feedUriToHref(url: string): string {
   }
 }
 
+export function postUriToRelativePath(
```
haileyok, line 191: cool i think i realized im going to need this, thanks for adding it!

src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -98,11 +104,42 @@ function ChatListItemReady({
...
+          if (isFromMe) {
+            lastMessage = _(msg`You: ${short}`)
+          } else {
+            lastMessage = short
+          }
+        }
+      } else {
+        if (isFromMe) {
+          lastMessage = _(msg`You: ${defaultEmbeddedContentMessage}`)
+        } else {
```
haileyok: small nit, maybe we can just add the `isFromMe` check once at end of these other checks. i think we'll probably have cases here eventually with external links, photos, etc.

src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -95,21 +101,64 @@ function ChatListItemReady({
...
   const isDimStyle = convo.muted || moderation.blocked || isDeletedAccount
 
-  let lastMessage = _(msg`No messages yet`)
-  let lastMessageSentAt: string | null = null
-  if (ChatBskyConvoDefs.isMessageView(convo.lastMessage)) {
-    if (convo.lastMessage.sender?.did === currentAccount?.did) {
-      lastMessage = _(msg`You: ${convo.lastMessage.text}`)
-    } else {
-      lastMessage = convo.lastMessage.text
+  const {lastMessage, lastMessageSentAt} = React.useMemo(() => {
```
haileyok, line 104: could just drop the memo too if you wanted to

haileyok (approved):

haileyok:
Thanks for doing this, I gave myself a headache figuring out how exactly to use the `isViewRecord` stuff again when i was about to try it 😭

PR #4340, opened 2024-06-03
fix top border width for user list updates

haileyok:
## Why

While reviewing https://github.com/bluesky-social/social-app/pull/4315, I noticed a couple of existing nits to clean up:

- Still using some `borderTopWidth: 1` instead of `hairline` in the add/remove users to list screen
- Title was hitting the side of the screen and could take up multiple lines
- Border above the `Done` button was also using `1` instead of `hairline`.

### Before

Screenshot 2024-06-03 at 4 08 32 PM

### After

See Below

Screenshot 2024-06-03 at 4 17 23 PM


gaearon (approved):

haileyok:
Screenshot 2024-06-03 at 4 29 01 PM

Bad screenshot in the main comment, doesn't show the border between items. Here's what it actually looks like.

PR #4343, opened 2024-06-04
Composer - make bottom border more consistent when typing

mozzius:
- Floored values
- Changed `>=` to `>`
- Added 5px of leeway before showing the border
- Adjusted `keyboardVerticalOffset` to 162

> We should figure out *why* `keyboardVerticalOffset` needs to be 162


gaearon (approved):
still seeing a jump when typing too far down but it seems unrelated to the recent PR so let's go

PR #4345, opened 2024-06-04
[🐴] allow nested quotes in DMs

mozzius:
Since we were using the `PostEmbeds` component, we lost the ability to see quoted content of quote posts. This adds a prop which stops the component from showing nested quotes

Also removes unused text style prop


gaearon (approved):
kind of wonky but ok

PR #4347, opened 2024-06-04
[🐴] Maybe fix overfetching

mozzius:
Noticed that we were potentially refetching for each event, even though events are batched. This changes it to refetch max once per batch


gaearon:
shouldn't RQ dedupe a bunch of sync calls to refetch?

mozzius:
You're right, I'll keep looking

PR #4350, opened 2024-06-04
use rngh scrollview in search horizontal list

haileyok:
Need to use the `react-native-guesture-handler` `ScrollView` for android for https://github.com/bluesky-social/social-app/pull/4169


estrattonbailey (approved):
Confirmed works

PR #4351, opened 2024-06-04
Mark `accessibilityLabel` and `accessibilityHint` for translation

surfdude29:
As a small follow-up to #4169 this marks the `accessibilityLabel` and `accessibilityHint` for translation.


estrattonbailey (approved):

estrattonbailey (changes requested):
Oh missed the lint error, need to import `useLingui` and call it to get `_` here

Thanks for this btw, good catch!

surfdude29:
I think I put it in the right place 🤞

estrattonbailey (approved):
Thank you!

PR #4352, opened 2024-06-04
[🐴] Use recent convos for share via dialog

estrattonbailey:
Conditionally prepends recent convo user to beginning of Share Via DM dialog list, instead of just follows. Follows fill in remaining slots, and no change to search functionality. New Chat dialog is unchanged.

This uses the existing list convos query, which should be cached by the time this loads. So even though it loads the last 50 convos, we should be ok to put them all in there.


gaearon (approved):
ahh yea this is great

PR #4356, opened 2024-06-04
add types for `desktopFixedHeight` to `List`

haileyok:
Let's stop requiring `@ts-ignore` for this prop in our `List` implementation

### DIFF
--- a/src/view/com/util/List.tsx
+++ b/src/view/com/util/List.tsx
@@ -26,6 +26,7 @@ export type ListProps<ItemT> = Omit<
   onItemSeen?: (item: ItemT) => void
   containWeb?: boolean
   sideBorders?: boolean
+  desktopFixedHeight?: boolean
   // Web only prop to disable a perf optimization (which would otherwise be on).
   disableContentVisibility?: boolean
 }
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):

haileyok:
Oh, actually need to look closer. In `Views.web.tsx` it's typed as `boolean | number`

gaearon:
I never implemented it for web actually. I think it's maybe used for native?

gaearon:
I'm not sure what it's really supposed to be doing. I think feeds use it.

haileyok:
Yea, the `FlatList_INTERNAL` takes both, but that makes the prop name a little confusing. I'll have to look through later to see why it gets used on native or if it even does.

I see that `desktopFixedHeightOffset` is a prop for `Feed`, but it never gets used either as far as I can tell?

Screenshot 2024-06-04 at 11 42 22 AM

haileyok:
Will leave `number` as well for now, but made a note to revisit this and remove `number` if possible. Seems fine, though want to give it a more solid pass before assuming that's true.

PR #4362, opened 2024-06-04
Enforce named imports and remove unnecessary `React` imports

shubh73:
Changes

https://github.com/bluesky-social/social-app/pull/4362#pullrequestreview-2097354883


gaearon:
fwiw using default exports for React is kind of wrong and i think the plan is to eventually drop support for it. instead we should move to named imports for Hooks (and types). for JSX, we shouldn't need to import `React` at all — something is wrong in our setup if this is required. the new JSX transform should not need that.

gaearon (changes requested):
need to:

- use named imports everywhere (you can use a codemod for that)
- remove `React` imports altogether
- figure out why JSX requires `React` and fix that "why" — it shouldn't
- add a lint rule that enforces named usage going forward

haileyok:
Noting the JSX rule is coming from here: https://www.npmjs.com/package/@react-native-community/eslint-config?activeTab=code

Not sure what repo that's in anymore, doesn't look like it's in facebook/react-native or any of the react-native-community repos at a first glance, though it definitely might be in one of those - I didn't look very closely. The `package.json` still has facebook/react-native listed though.

shubh73:
Thanks for the feedback and additional context, @gaearon!

- `Use named imports everywhere (you can use a codemod for that)`:  Done using `update-react-imports` codemod. However, I want to ensure that I haven't missed any imports. Is there a way to test this properly?

- `Remove React imports altogether:` Done using `update-react-imports` codemod.

- `Figure out why JSX requires React and fix that "why" — it shouldn't`:  The `@react-native-community/eslint-config` package has been renamed to `@react-native/eslint-config` since [React Native 0.72](https://github.com/facebook/react-native/pull/37347). For now, I've disabled the `'react/jsx-uses-react'` rule.

- `Add a lint rule that enforces named usage going forward`: I've added the `'no-restricted-imports'` rule to the ESLint config, which restricts the default import of react. This should help enforce the usage of named imports.

PR #4364, opened 2024-06-04
Create a testable version of ago() and re-enable the disabled test

janaagaard75:
The [current tests for ago() function](https://github.com/bluesky-social/social-app/blob/22858192dc5848cf75fb48fa911c8a9bbf08dbaa/__tests__/lib/string.test.ts#L146) are not deterministic, so they are currently disabled.

Create a testable version of ago(), **dateDiff()**, that has two parameter, allowing for deterministic tests. ago() is then implemented as a special case of dateDiff(), where one of the dates is `new Date()`.

The implementation is changed so that months are now considered to a fixed 30 days instead of the current 30.41675 days. Using 30 days is arguably less precise, but it is a lot easier to reason about because the cut off from month to month is always a full day. I think this simpler solution was preferable. The root cause of the mess is that "a month ago" means a different number of days depending the current month.

This follows up on https://github.com/bluesky-social/social-app/pull/4295.

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -147,70 +147,67 @@ describe('ago', () => {
     new Date().setMonth(new Date().getMonth() - 11),
   ).setDate(new Date().getDate() - 28)
 
-  const inputs = [
-    1671461038,
-    '04 Dec 1995 00:12:00 GMT',
-    new Date(),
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Thank you so much! Really appreciate this.

I'm going to merge into another branch to make some edits to the tests and add a formatting option we've been kicking around. Will get this integrated soon.

PR #4367, opened 2024-06-04
Clarify some things in OTA docs

estrattonbailey:


haileyok (approved):
Thank you for adding to this! I've also made it easier to get this info from Slack, and updated the docs accordingly!

PR #4369, opened 2024-06-04
Fix: visually indicate when quoted content is labeled

pfrazee:
There's been some confusion with the current "blurring" control for quote posts, where it looks like the quoting post is the one labeled. This update fixes that issue by showing the border around the blur control.

(The example below is not actually labeled)

|Before|After|
|-|-|
|CleanShot 2024-06-04 at 16 59 06@2xCleanShot 2024-06-04 at 16 59 19@2x|CleanShot 2024-06-04 at 16 56 53@2xCleanShot 2024-06-04 at 16 58 30@2x|


gaearon (approved):
looking great! not sure if we should bring the quoted author in to be consistent with how a standalone is displayed, but this already is so much better

PR #4371, opened 2024-06-05
Remove `expo-constants`

haileyok:
## Why

We're not using this package at all anymore - we moved to using `expo-application` and `expo-device` as recommended in the Expo docs (https://docs.expo.dev/versions/latest/sdk/constants/), so lets rm.

## Test Plan

Shouldn't see a warning anymore in dev client, and should still see version info on the Settings screen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,6 @@
     "expo-build-properties": "^0.12.1",
     "expo-camera": "~15.0.9",
     "expo-clipboard": "^6.0.3",
-    "expo-constants": "~16.0.1",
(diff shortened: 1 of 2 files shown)


gaearon (approved):

haileyok:
Ah lol, typecheck fails now because something in the _tests_ is relying on `expo-constants`. Maybe our jest mocks. Will look later, not urgent.

haileyok:
Verified that tests don't actually fail, unsure what the case was in the session tests, though it is unrelated to `expo-constants` (it isn't used in any jest mocks either).

Issue #2683, opened 2024-01-30
Add 'Search my posts' link to dropdown menu on a user's own profile


surfdude29:
**Is your feature request related to a problem? Please describe.**

Searching your own posts has now been made easier by #2504 which added the 'from:me' search functionality. However, unless a user is aware of this they may not know how to search their own posts.

**Describe the solution you'd like**

Please add a 'Search my posts' link to the dropdown menu on a user's own profile, which would open the Search page with the query pre-filled with "from:me ". This would make searching your own posts a lot easier and more discoverable, especially for users new to Bluesky.

**Describe alternatives you've considered**

Those who know how to do it can simply type in "from:me" in the Search bar.

**Additional context**

The corresponding issue for adding a link in the dropdown menu to search another user's posts is https://github.com/bluesky-social/social-app/issues/2684.

Issue #2684, opened 2024-01-30
Add 'Search posts' link to dropdown menu on user's profile


surfdude29:
**Is your feature request related to a problem? Please describe.**

Searching a user's posts has now been made easier by recent improvements to the search service. However, unless someone is aware of this they may not know how to search another user's posts.

**Describe the solution you'd like**

Please add a 'Search posts' link to the dropdown menu on a user's profile, which would open the Search page with the query pre-filled with "from:[username] ". This would make searching another user's posts easier and more discoverable, especially for users new to Bluesky.

**Describe alternatives you've considered**

Those who know how to do it can simply type in "from:[username]" in the Search bar.

**Additional context**

The corresponding issue for adding a link in the dropdown menu to search a user's own posts is #2683.

PR #4373, opened 2024-06-05
Ability to search posts via profile dropdown

oracularhades:
Description: I've been using search tags to get posts from users, usually finding posts from 6 months ago that are going around in my head. I've found some usernames are long and annoying to type, or I can just never remember them (I cannot for the life of me remember Paul's "pfrazee.com" username [paul ur awesome, I'm just bad at some names]). Usernames have no autocomplete in the search-bar when using search tags. This makes it nice and easy.

You'll notice I've made another "<Menu.Group>", this is because otherwise users can't see the button when viewing their own account, or it's above the divider when logged in (if using the menu.group share is in). It's not visible when logged out because menu.group #2 and #3 are both under an "isSession" condition. I could have added it separately, but it would have made the code slightly harder to read, when there really isn't a use case for showing the button when logged out.

Limitations:
- "Search posts" is not translated.
- I have not tested this on mobile.

Screenshot 2024-06-05 at 3 46 07 PM

https://github.com/bluesky-social/social-app/assets/91714073/08ae003a-5780-4a15-aac3-8b1e5b080194

🦋

Thanks,
Josh

### DIFF
--- a/src/view/com/profile/ProfileMenu.tsx
+++ b/src/view/com/profile/ProfileMenu.tsx
@@ -234,6 +234,24 @@ let ProfileMenu = ({
           {hasSession && (
             <>
               <Menu.Divider />
+
+              {/* This group is seperated so users can use this button on their own account. */}
+              <Menu.Group>
+                <Menu.Item
+                  testID="profileHeaderDropdownSearchPostsBtn"
+                  label={_(msg`Search Posts`)}
+                  onPress={() => {
+                    navigate('Search', {
+                      q: `from:${profile.handle}`,
+                    })
+                  }}>
+                  <Menu.ItemText>
+                    <Trans>Search Posts</Trans>
+                  </Menu.ItemText>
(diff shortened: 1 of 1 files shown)


oracularhades:
Update. I have changed the position of the menu item, because before I didn't check what it looked like on someone else's profile (I'm very tired and it went over my head lol). It looked absurdly jarring before.

It could be nicer if it was in the middle of the other buttons, but that requires splitting the menu block and condition (so that someone can see it on their account), which I don't think is worth it unless the design team isn't happy with this.

**Before:** (jarring)
Screenshot 2024-06-05 at 4 07 42 PM

**After:**
Screenshot 2024-06-05 at 4 09 08 PM

Screenshot 2024-06-05 at 4 13 02 PM

surfdude29:
This would fix #2683 and #2684, thanks! 🙏

lucidillusions (commented):
this would be a good addition.

PR #4377, opened 2024-06-05
add `legacy` to picker for Android

haileyok:
## Why

Missed adding this flag in (big whoops on my part!). Android now _does_ show the `...` by default, but it _only_ allows browsing "Cloud media". The `legacy` flag is required for _local_ content.

## Test Plan

Build the app, see that the `...` also has the `Browse` option in the dropdown.


gaearon (approved):

gaearon:
is this the only callsite?

haileyok:
yea, one more in onboarding. otherwise we look good.

PR #4378, opened 2024-06-05
Fix undefined block

### LINKED ISSUE #4322: Reply to undefined when parent is blocked
**Describe the bug**
In a feed, replies to blocked users show "Reply to undefined".

This occurs the same way on web and in the Android app. (iOS untested)

**To Reproduce**

View a reply to a blocked user (whenever the post would appear as blocked, so for the viewer's own blocks, as well as blocks between the two other users), whenever the UI would add the "Reply to XYZ" context, such as when the reply is shown in a feed without the parent post.

**Expected behavior**

The reply context should read something like "Reply to a blocked user".

**Screenshots**

image

**Details**

 - Platforms:
  - Desktop: Chrome, Build version: 1.84.0; Bundle info: e6d2618 (prod); Bundle date: 24052420; Platform: web
  - Mobile: Android, 1.84.0.199 e6d2618 (prod)

oracularhades:
This fixes #4322

https://github.com/bluesky-social/social-app/assets/91714073/6a3048ad-d556-445f-8fca-8776d2eefa2d

### DIFF
--- a/src/view/com/posts/FeedItem.tsx
+++ b/src/view/com/posts/FeedItem.tsx
@@ -196,6 +196,12 @@ let FeedItemInner = ({
     },
   ]
 
+  let blockedParentAuthor: AppBskyActorDefs.ProfileViewBasic = {
+    did: '',
+    displayName: '[user blocked you]',
+    handle: '',
+  }
+
   return (
     <Link
       testID={`feedItem-by-${post.author.handle}`}
@@ -320,7 +326,13 @@ let FeedItemInner = ({
             onOpenAuthor={onOpenAuthor}
           />
           {!isThreadChild && showReplyTo && parentAuthor && (
-            <ReplyToLabel profile={parentAuthor} />
+            <ReplyToLabel
+              profile={
+                parentAuthor.viewer?.blockedBy
+                  ? blockedParentAuthor
+                  : parentAuthor
+              }
+            />
           )}
           <LabelsOnMyPost post={post} />
           <PostContent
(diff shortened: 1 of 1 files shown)


src/view/com/posts/FeedItem.tsx
```diff
@@ -196,6 +196,12 @@ let FeedItemInner = ({
     },
   ]
 
+  let blockedParentAuthor: AppBskyActorDefs.ProfileViewBasic = {
+    did: '',
+    displayName: '[user blocked you]',
```
estrattonbailey, line 201: If I block this user instead of them blocking me, this text might be confusing. Should probably do something more generic like `[blocked user]` and then checking for all forms of blocks on the viewer state when displaying it:

```
  blockedBy?: boolean
  blocking?: string
  blockingByList?: AppBskyGraphDefs.ListViewBasic
```

estrattonbailey (commented):
Thanks for this! Needs a small tweak, otherwise should be good pending other team feedback.

src/view/com/posts/FeedItem.tsx
oracularhades: I'm home in 30m and will adjust this then :)

oracularhades:
@estrattonbailey Thanks for letting me know! I didn't know ``blockedby`` was both me blocking them, and them blocking me. I thought me blocking them was handled by ``blocking``. Fixed now!

Screenshot 2024-06-06 at 12 47 22 PM

estrattonbailey (approved):
Looking good! Sorry for the delay. Gonna merge this into a branch and tweak it based on team thoughts and then we'll get it in. Thanks again!

PR #4379, opened 2024-06-05
Stringify path error

estrattonbailey:
We want to know what the object `path` is here.
CleanShot 2024-06-05 at 15 11 48@2x


AlirezaRahmani2004:
Let's see if we can swap it or not.

pfrazee (approved):

PR #4380, opened 2024-06-05
Sort profile feedgens by likes

estrattonbailey:
Sorts _each page_ of profile feedgens by likes. Because this is per-page, I upped the page size to `50` to lower the chances of not getting a good sort for users with many feeds.

I also considered appending results and re-sorting as it paginates, but this would mean that if a popular feed loads from page 2, it would get sorted to the top out of the user's view, and they _still_ may not see it. So best to retain per-page sorts for the moment.

### DIFF
--- a/src/state/queries/profile-feedgens.ts
+++ b/src/state/queries/profile-feedgens.ts
@@ -30,13 +30,24 @@ export function useProfileFeedgensQuery(
         limit: PAGE_SIZE,
         cursor: pageParam,
       })
-      res.data.feeds = res.data.feeds.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


src/state/queries/profile-feedgens.ts
```diff
@@ -30,6 +30,9 @@ export function useProfileFeedgensQuery(
         limit: PAGE_SIZE,
         cursor: pageParam,
       })
+      res.data.feeds = res.data.feeds.sort((a, b) => {
```
gaearon, line 33: `sort` is mutative so let's remove the reassignment for clarity

gaearon (approved):

PR #4386, opened 2024-06-06
cancel animations before updating value for min-shell animation

haileyok:
## Why

There seems to be a performance regression after updating Reanimated to 3.11.0. We were able to track this down to performing updates on `mode.value` while also running a `withTiming` animation on the same value. For example, if we were to remove this code:

```tsx
      mode.value = withTiming(v ? 1 : 0, {
        duration: 400,
        easing: Easing.bezier(0.25, 0.1, 0.25, 1),
      })
```

we would see the smoothness of the scroll be fixed.

## How

Any time that we update `mode` - whether it is to manually update the value or to animate the value - we should cancel an existing animation that for that value.

## Test Plan

Build this branch with `yarn android --device --variant release` and see if the performance is fixed. From my testing on a device, it does appear to be better. There doesn't appear to be any regression in the animation itself either.


gaearon (approved):
i'm not _sure_ if this does anything but it seems maybe marginally better. let's get it in too

gaearon:
(the main fix that did it for me was https://github.com/bluesky-social/social-app/pull/4391)

PR #4391, opened 2024-06-06
Switch to withSpring to work around Reanimated regression

gaearon:
When we switched from Reanimated 3.6 to 3.11, scrolling on Home has gotten incredibly choppy. @haileyok and I narrowed it down to this animated value — it's being set to both plain numbers and to `withTiming`, and for some reason it appears that doing both (or maybe even just `withTiming` by itself) makes it choppy even _after_ the animation has completed (i.e. while scrolling later). It seems like `withSpring` doesn't have this problem so this just switches us to do that.

I've looked at other places where we have `withTiming` and they don't seem negatively affected so I didn't touch it.

## Test Plan

Scroll Home feed on low-end Android. It was super choppy before, now it's okay (with some dropped frames).

### DIFF
--- a/src/state/shell/minimal-mode.tsx
+++ b/src/state/shell/minimal-mode.tsx
@@ -1,10 +1,5 @@
 import React from 'react'
-import {
-  Easing,
-  SharedValue,
-  useSharedValue,
-  withTiming,
-} from 'react-native-reanimated'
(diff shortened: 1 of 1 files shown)


mozzius (approved):
I wasn't able to reproduce the original bug to confirm it fixed it, but this change looks pretty safe and works the same on my sim

PR #4392, opened 2024-06-06
Fix prettier

gaearon:
oops


mozzius (approved):

wojtekmaj:
Speaking of Prettier, would switching to Biome, which is an order of magnitude faster than Prettier, enable some additional checks (even on pre-commit), to avoid such issues in the future?

PR #4400, opened 2024-06-06
Fix notifications pagination on web

pfrazee:
Reverts https://github.com/bluesky-social/social-app/pull/4235

It looks like the issue is that fetching the next page doesn't reliably add 2x list height to the scrollview. The intersection observer only tells us when the boundary is crossed -- it doesn't trigger if we're still in the boundary -- and so the next page successfully fires but then doesn't retrigger.

It's entirely possible, with client-side filtering, that even dropping the threshold down to 60% doesn't totally fix the issue.

- We could possibly solve this by calculating the number of new list items getting added post-grouping and, if it's too low, queue another fetch. That feels brittle though.
- We could look for some way to fire an event post-layout, perhaps by running [checkVisiblity](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) on the bottom of the screen.

### DIFF
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -182,7 +182,7 @@ export function Feed({
         refreshing={isPTRing}
         onRefresh={onRefresh}
         onEndReached={onEndReached}
-        onEndReachedThreshold={2}
+        onEndReachedThreshold={0.6}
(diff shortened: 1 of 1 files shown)


haileyok (approved):

gaearon:
We could maybe change the logic to keep trying until we either run out of pages or the tail runs out of viewport again.

PR #4403, opened 2024-06-06
add support for `ListEmptyComponent`, allow `undefined` data

haileyok:
### 1. `data` can be `undefined`

The native `FlatList` implementation allows `data` to be `undefined`. It does not _have_ to be an array. In our web implementation though, we assume that it is since all we do is assert `data as Array<ItemT>`.

To test this, use `Feed.tsx` and change the `data` prop to `undefined`. This appears fine - since the `FlatList` types allow it - but our causes an error.

Screenshot 2024-06-06 at 4 09 03 PM

### 2. `ListEmptyComponent` should be rendered when `data` is `undefined` or an empty array

There is a `ListEmptyComponent` prop on `FlatList` that will render the given element when the `data` is empty or `undefined`. I've added support for that view as well in this PR.

## Test Plan

- Use `Feeds.tsx` as an example
- Change `data` in the `List` to `undefined`
- There should no longer be an error 

Screenshot 2024-06-06 at 4 10 34 PM

### Empty Component 

- Add `ListEmptyComponent` prop to the `List`, with a value like:

```tsx
<View style={{alignItems: 'center', marginTop: 20}}>
	<Loader size="lg" />
</View>
```

- You should see a spinner

Screenshot 2024-06-06 at 4 12 47 PM

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -73,6 +73,8 @@ function ListImpl<ItemT>(
     )
   }
 
+  const isEmpty = !data || data.length === 0
+
   let headerComponent: JSX.Element | null = null
   if (ListHeaderComponent != null) {
     if (isValidElement(ListHeaderComponent)) {
@@ -334,7 +336,7 @@ function ListImpl<ItemT>(
           onVisibleChange={handleAboveTheFoldVisibleChange}
           style={[styles.aboveTheFoldDetector, {height: headerOffset}]}
         />
-        {onStartReached && (
+        {onStartReached && !isEmpty && (
           <Visibility
             root={containWeb ? nativeRef : null}
             onVisibleChange={onHeadVisibilityChange}
@@ -342,7 +344,7 @@ function ListImpl<ItemT>(
           />
         )}
         {headerComponent}
-        {!data || data.length === 0
+        {isEmpty
           ? emptyComponent
           : (data as Array<ItemT>)?.map((item, index) => {
               const key = keyExtractor!(item, index)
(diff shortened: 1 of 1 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -72,23 +73,33 @@ function ListImpl<ItemT>(
...
-      footer = ListFooterComponent
+      footerComponent = ListFooterComponent
     } else {
       // @ts-ignore Nah it's fine.
-      footer = <ListFooterComponent />
+      footerComponent = <ListFooterComponent />
+    }
+  }
+
+  let emptyComponent: JSX.Element | null = null
```
haileyok, line 96: I opted to call this `emptyComponent` even though our other elements are just `header` for `ListHeaderComponent` and `footer` for `ListFooterComponent`, since just `empty` is a little confusing at first glance. I renamed the other two accordingly.

src/view/com/util/List.web.tsx
```diff
@@ -323,36 +336,38 @@ function ListImpl<ItemT>(
...
+                  item={item}
+                  index={index}
+                  renderItem={renderItem}
+                  extraData={extraData}
+                  onItemSeen={onItemSeen}
+                  disableContentVisibility={disableContentVisibility}
+                />
+              )
+            })}
+        {onEndReached && !isEmpty && (
```
haileyok, line 363: This might not be strictly necessary, but I think it makes sense that we don't want to render these `Visibility` components if there is no data anyway.

gaearon (approved):

PR #4405, opened 2024-06-07
Add like animation

pfrazee:
https://github.com/bluesky-social/social-app/assets/1270099/86195973-dd56-4dc6-9016-57c2cd4c9ff8

Tested and works on all platforms


mozzius:
We should add it to the repost too!

gaearon:
Tweaked and merged in https://github.com/bluesky-social/social-app/pull/5033

PR #4417, opened 2024-06-07
Set profile hover prefetch stale time to 30s

estrattonbailey:
Two related things were happening in this PR:
- though `ProfileHoverCard` was using `onPointerMove` to prevent triggering prefetches while scrolling, `PostMeta` was still using `onPointerEnter`, which triggers while scrolling
- whether scroll or roll-over, profile prefetch was being called every time, so adding stale time of 30s effectively debounces those calls

So the result is: `getProfile` is no longer called on scroll, and it's called at most once every 30s if the user intentionally hovers over the triggers.

In `PostMeta`, `Text` doesn't support `onPointerMove`, so I wrapped it in a `View` and double checked overflow, should be OK (see screenshots)
CleanShot 2024-06-07 at 10 27 43@2x
CleanShot 2024-06-07 at 10 30 40@2x

### DIFF
--- a/src/view/com/util/PostMeta.tsx
+++ b/src/view/com/util/PostMeta.tsx
@@ -11,6 +11,7 @@ import {sanitizeHandle} from 'lib/strings/handles'
 import {niceDate} from 'lib/strings/time'
 import {TypographyVariant} from 'lib/ThemeContext'
 import {isAndroid, isWeb} from 'platform/detection'
+import {atoms as a} from '#/alf'
 import {ProfileHoverCard} from '#/components/ProfileHoverCard'
 import {TextLinkOnWebOnly} from './Link'
 import {Text} from './text/Text'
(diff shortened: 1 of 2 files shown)


src/view/com/util/PostMeta.tsx
```diff
@@ -66,37 +67,37 @@ let PostMeta = (opts: PostMetaOpts): React.ReactNode => {
...
-            disableMismatchWarning
-            style={[pal.textLight, {flexShrink: 4}]}
-            text={'\xa0' + sanitizeHandle(handle, '@')}
-            href={profileLink}
-            onBeforePress={onBeforePressAuthor}
-            onPointerEnter={onPointerEnter}
-            anchorNoUnderline
-          />
-        </Text>
+        <View onPointerMove={onPointerEnter} style={[a.flex_1]}>
```
gaearon, line 70: `onPointerMove={onPointerEnter}` is confusing, imo we should rename the prop if this is what it's doing now

gaearon: Oh it's not a prop, just a method. We should rename it. Also we should add a ref check because we don't want to hammer `prefetchQuery` on every pixel movement even if RQ can deduplicate those.

estrattonbailey: Ah good calls, done 👍

gaearon (approved):
lg but let's land after release

PR #4420, opened 2024-06-07
Add KnownFollowers component to standard profile header

estrattonbailey:
Adds known followers to profiles. Shows up to 3 avis, and 2 display names. Text wraps to 2 lines before overflowing via ellipsis. Link gets an underline on web when hovering, opacity on native when pressing.

If no known followers, this is hidden entirely and the profile header appears as it always has.

**I added an Oxford comma, didn't re-take screenshots tho** 🤷 

Known followers list page is a copy of our other Followers/Follows pages, no changes except data source and page title.

### Real data:
CleanShot 2024-06-11 at 15 40 15@2x
CleanShot 2024-06-11 at 15 40 11@2x
CleanShot 2024-06-11 at 15 37 29@2x
CleanShot 2024-06-11 at 15 36 50@2x
CleanShot 2024-06-11 at 15 36 41@2x
CleanShot 2024-06-11 at 15 37 10@2x

### Fake data for different states:
CleanShot 2024-06-07 at 16 05 34@2x
CleanShot 2024-06-07 at 16 05 49@2x
CleanShot 2024-06-07 at 16 05 56@2x
CleanShot 2024-06-07 at 16 06 38@2x
CleanShot 2024-06-07 at 16 10 40@2x
CleanShot 2024-06-07 at 16 08 49@2x
CleanShot 2024-06-07 at 16 09 00@2x
CleanShot 2024-06-07 at 16 09 06@2x
CleanShot 2024-06-07 at 16 09 16@2x
CleanShot 2024-06-07 at 16 09 29@2x


src/components/KnownFollowers.tsx
```diff
@@ -0,0 +1,154 @@
...
+export function KnownFollowers({
+  profile,
+  moderationOpts,
+}: {
+  profile: AppBskyActorDefs.ProfileViewDetailed
+  moderationOpts: ModerationOpts
+}) {
+  if (!profile.viewer?.knownFollowers) {
+    return null
+  }
```
estrattonbailey, line 27: Note to self, check for `count === 0` here too

estrattonbailey: Actually, we're going to check for `followers.length` instead, since blocks will still count towards `count`, see https://github.com/bluesky-social/atproto/pull/2557/files#r1634012991

mozzius (approved):
great stuff

src/components/KnownFollowers.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {makeProfileLink} from '#/lib/routes/links'
+import {isNative} from '#/platform/detection'
+import {sanitizeDisplayName} from 'lib/strings/display-names'
+import {UserAvatar} from '#/view/com/util/UserAvatar'
+import {atoms as a, useTheme} from '#/alf'
+import {Link} from '#/components/Link'
+import {Text} from '#/components/Typography'
+
+const AVI_SIZE = 30
+const AVI_BORDER = isNative ? 1 : 1
```
pfrazee, line 16: 1 or 1 eh

src/components/KnownFollowers.tsx
```diff
@@ -0,0 +1,201 @@
...
+    new Map(),
+  )
+
+  /*
+   * Results for `knownFollowers` are not sorted consistently, so when
+   * revalidating we can see a flash of this data updating. This cache prevents
+   * this happening for screens that remain in memory. When pushing a new
+   * screen, or once this one is popped, this cache is empty, so new data is
+   * displayed.
+   */
```
pfrazee, line 47: Wouldn't it be simpler to just sort `knownFollowers` ?

pfrazee: (in the client I mean)

estrattonbailey: We only get a slice back, there could be `n` results, so the slice we get may vary. We could do a better job of this on the backend, but we'd only be sorting DIDs that we can back as the skeleton. The dataplane _does_ return the full list atm though, so in that sense we _could_ add a stable sort by DID prior to hydrating just the slice of users we need for the request.

src/components/KnownFollowers.tsx
estrattonbailey: Whoops!

src/components/KnownFollowers.tsx
pfrazee: Slack conversation surfaced that we basically can't rely on the backend to ever give us the same results with every query. I'm comfortable with the logic of this approach.

pfrazee (approved):
Sick. Did a minor spacing tweak with labels underneath the social proof. Ship it!

PR #4427, opened 2024-06-08
Reliably focus keyboard on Android

gaearon:
My best attempt at making it work based on a real device. Requires some testing.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -405,13 +405,31 @@ export const ComposePost = observer(function ComposePost({
   // Backup focus on android, if the keyboard *still* refuses to show
   useEffect(() => {
     if (!isAndroid) return
-    if (isModalReady) {
-      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Tested on 3 devices from high to low end (lowest being a 2017 device). Keyboard presents in all cases and is generally smoother than before. This feels good as a guarantee that we don't run into problems. We should still revisit at some point I think but this is much more solid than what we currently have.

src/view/com/composer/Composer.tsx
```diff
@@ -405,13 +405,31 @@ export const ComposePost = observer(function ComposePost({
...
-          textInput.current?.blur()
-          textInput.current?.focus()
-        }
-      }, 300)
+    if (!isModalReady) return
+
+    function tryFocus() {
+      if (!Keyboard.isVisible()) {
+        textInput.current?.blur()
+        textInput.current?.focus()
```
har2008preet, line 413: we should clear interval here

PR #4429, opened 2024-06-08
Remove unnecessary Android code

gaearon:
Seems unnecessary after https://github.com/bluesky-social/social-app/pull/4427, let's remove to simplify.

Tested on my device, would be nice to test on more.


haileyok:
Seems fine, will test more in a bit. Should we also remove the `isModalReady` state and reactions from `shell/Composer.tsx`?

gaearon:
not relevant anymore

Issue #4423, opened 2024-06-07
Mini profile on hover allows following a blocker/blocked user


Juicysteak117:
**Describe the bug**
I was able to follow a user blocking me via the mini profile on hover. There's the big follow button still there. This persists in the case after blocking the user.

**To Reproduce**

1. Find a user blocking you or that you have blocked
2. Observe a reposted reply to the user which indicated "Reply to undefined", or perhaps someone tagging them
3. Notice the ability to follow

**Expected behavior**

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/20595808/94562e61-35dc-4c96-a39d-d1e8b83dbf93)
![image](https://github.com/bluesky-social/social-app/assets/20595808/7cd41cd1-aadc-40c3-89b8-dba36617c66e)

**Details**

 - Platform: Web
 - Platform version: Chromium
 - App version:

**Additional context**

I also noticed that when I blocked and unblocked this example user that for a brief flash after unblocking the user I could see their profile. Presumably the refresh after unblocking happens before the check for being blocked, allowing this momentary loophole.

PR #4440, opened 2024-06-08
Fix: mini profile on hover allows following a blocker/blocked user (#4423)

michalgolda:


estrattonbailey (approved):
Thanks for this! Gonna merge into a branch and make a tweak, then we'll get this in 👍

PR #4443, opened 2024-06-09
add support for `mark-read-generic` and `mark-read-messages`

haileyok:
## Why

We've added support for badge incrementing through https://github.com/bluesky-social/social-app/pull/4005 and https://github.com/bluesky-social/social-app/pull/4233. However, as of right now the only way to "decrement" the badge is to open the app and opening your notifications tab. Instead, we want the badge to decrement in every possible case:

- You open the notifications tab on the current device
- You open the notifications tab on _another_ device
- You read a message on the current device
- You read a message on _another_ device

## How

Right now, we use `updateSeen` to mark notifications having been "read". This works alright, but it doesn't give us any solid info or granularity. Nonetheless, for now we can just know that our "generic" notification badge count is going to drop to zero any time `updateSeen` is called.

In the case of messages, we again do not have a good way to "mark all as read". So, we will currently rely on just decrementing the messages badge count by one whenever we receive a `mark-read-messages`.

In both of these cases though, we should anticipate being able to provide a "count" of messages to decrement by. Therefore, in both cases I've added the ability to pass along a `decrementBy` field in the notification, which will tell the client to decrement the badge by that number of notifications. the current implementation makes this a little bit sketchy - sometimes we don't actually receive push notifications, I'm not sure why and that needs to get investigated - so this is solely a "looking forward" bit of logic rather than something we're going to do right now.

## Test Plan

We'll first need to deploy the backend changes (https://github.com/bluesky-social/atproto/pull/2567), at which point we'll start receiving `mark-read-generic` reason notifications. The good news is that we can test this at _any_ time, since we do not need to supply a `title` or `body` field for the notification, and as such iOS will not display it to the user anyway - even without the NSE "intercepting" the notification.

- Deploy https://github.com/bluesky-social/atproto/pull/2567
- Use this client _on device_ (`yarn intl:build && eas build -p ios --profile testflight --local --output build.ipa --non-interactive && eas submit -p ios --path build.ipa)
- Send yourself some notifications and observe the badge increment properly
- On another device or on the web app, open your notifications tab to clear the badge
- Test that the iOS badge also decrements whenever you do this

### Testing that notifications with empty `title` and `body` do not show up on iOS

https://github.com/bluesky-social/social-app/assets/153161762/70b0a26c-0f72-45b5-972b-f4db169dc89d


haileyok:
This should be ready for testing now as soon as we deploy the courier and appview changes.

PR #4446, opened 2024-06-09
[Session] Convert account to session data explicitly

gaearon:
We were previously suppressing a type error which was a legit mismatch in types. We were assigning `SessionAccount` (an app concept) directly into `agent.session` as if it were `AtpSessionData`.

Strictly saying, this isn't valid — for example, `refreshJwt` can be `undefined` in `SessionAcount` but is expected to be a `string` (even if empty) in `AtpSessionData`. We *do* have a check (per suppression comment) that should rule out this actually happening but I think it's better to ensure this statically. This also avoids unrelated account fields like `pdsUrl` ending up attached to the `session` object.

This PR removes the suppression and adds an explicit conversion based on existing logic from the resume session codepath.

**I don't expect this to change anything in practice.**

## Test Plan

Idk how to test this since it needs to trigger expiry. I think we need support from backend to force early expiry — with a header or something. So for now the test plan is just careful reading.

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -37,21 +37,7 @@ export async function createAgentAndResume(
   }
   const gates = tryFetchGates(storedAccount.did, 'prefer-low-latency')
   const moderation = configureModerationForAccount(agent, storedAccount)
-  const prevSession: AtpSessionData = {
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Oh yeah great idea

PR #4447, opened 2024-06-09
GIF previews in notifications

mozzius:
Currently there's no preview, which is especially odd if the post has no text. It's not animated, but I included a little label

Screenshot 2024-06-09 at 15 19 45

### DIFF
--- a/src/lib/strings/embed-player.ts
+++ b/src/lib/strings/embed-player.ts
@@ -1,7 +1,8 @@
-import {Dimensions, Platform} from 'react-native'
+import {Dimensions} from 'react-native'
 
 import {isSafari} from 'lib/browser'
 import {isWeb} from 'platform/detection'
+
 const {height: SCREEN_HEIGHT} = Dimensions.get('window')
 
 const IFRAME_HOST = isWeb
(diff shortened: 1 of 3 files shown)


src/lib/strings/embed-player.ts
```diff
@@ -516,3 +494,56 @@ export function getGiphyMetaUri(url: URL) {
...
+      id = id.replace('AAAAC', 'AAAAM')
+    }
+
+    return {
+      success: true,
+      playerUri: `https://t.gifs.bsky.app/${id}/${filename}`,
+      dimensions,
+    }
+  } catch {
+    return {success: false}
```
gaearon, line 547: I think this is a slightly risky way to approach this because it would swallow actual coding mistakes and silently start failing validation (in a way that seems like something's broken with the gif). 

Since we're already returning a first-class value representing failure and success, let's replace throws with `return { success: false }` and then remove try/catch (assuming we think our code for parsing is solid).

Is there any downside to that?

src/view/com/notifications/FeedItem.tsx
```diff
@@ -465,17 +467,41 @@ function AdditionalPostText({post}: {post?: AppBskyFeedDefs.PostView}) {
...
+      try {
+        const {success} = parseTenorGif(new URL(post.embed.external.uri))
+        if (success && post.embed.external.thumb) {
+          isGif = true
+          images = [
+            {
+              thumb: post.embed.external.thumb,
+              alt: post.embed.external.title,
+              fullsize: post.embed.external.thumb,
+            } satisfies AppBskyEmbedImages.ViewImage,
```
gaearon, line 490: Is this `satisfies` necessary? Just curious

src/view/com/util/images/ImageHorzList.tsx
```diff
@@ -2,39 +2,60 @@ import React from 'react'
...
   images: AppBskyEmbedImages.ViewImage[]
   style?: StyleProp<ViewStyle>
+  gif?: boolean
 }
 
-export function ImageHorzList({images, style}: Props) {
+export function ImageHorzList({images, style, gif}: Props) {
   return (
-    <View style={[styles.flexRow, style]}>
+    <View style={[a.flex_row, a.gap_xs, style]}>
```
gaearon, line 18: This is also used from DMs for the embed, are these style changes safe for that usage?

gaearon (approved):
the logic looks good, haven't tested the design yet. i'd prefer a version that didn't rely on try/catch for validation (see above comment), the rest should be good if you tested all usages of the changed component on all platforms

gaearon (changes requested):
i'd like to see https://github.com/bluesky-social/social-app/pull/4447#discussion_r1640625618 addressed so gonna flag for revision or i'll forget

src/view/com/util/images/ImageHorzList.tsx
mozzius: Yes, I made sure. Basically just replacing the StyleSheet styles with ALF atoms

gaearon (approved):

PR #4451, opened 2024-06-09
Fix (#4430): Use separate hooks for shell mode animated styles

### LINKED ISSUE #4430: Scroll performance on Profile page on Android
**Describe the bug**
When inside the profile screen, at least on my device (Xiaomi MI 9 SE) the UI stutters visibly while rendering the `LoadLatestBtn`.
What's weird: this happens on this device only (tried on Pixel 7 Emu and not able to repro) and exactly on the profile page (no visible stutter in home/feeds page or in notification screen, despite the logic for showing/hiding the button being identical)

**To Reproduce**

1. Go to the profile page on an Android device
2. Scroll until the button to go back to top is visible
3. Notice the lag

**Expected behavior**
The scroll should be smooth

**Screenshots**
(video)
Default behavior | Expected (read below)
:-: | :-:
<video src='https://github.com/bluesky-social/social-app/assets/2643961/ae8fd7bb-a266-4e60-bfcf-bc23d302191e' width=150/> | <video src='https://github.com/bluesky-social/social-app/assets/2643961/8822d6d2-f0a1-4187-ad34-65a705cb376f' width=150/>

**Details**

 - Platform: Android
 - Platform version: 14 (LineageOS)
 - App version: 1.185.1

**Additional context**

I explored the relevant logic here and experimented a bit, and I think the problem is with `useMinimalShellMode` creating and exporting the animatedStyle used by `LoadLatestBtn` together with other animatedStyles?

Apparently I managed to make it behave by exporting `fabMinimalShellTransform` from a dedicated hook instead of `useMinimalShellMode`.

This in theory avoids creating and returning all of the transforms altogether from the `useMinimalShellMode` hook, but I'm not very familiar with how Reanimated's `useAnimatedStyle` works under the hood and why this should impact performance on mid/low end devices.
Also, this doesn't explain why I can reproduce it ONLY on the Profile screen...

Default behaviour (React DevTools profiler, dev mode):

Separating the `animatedStyle` creation (React DevTools profiler, dev mode):

frncs-eu:
This is just a silly "move stuff" PR and is small step towards improving the perceived performance on low end devices.
This PR is based on the findings reported in #4430 and replaces the hook `useMinimalShellMode` with 3 separate hooks dedicated to building the animated style (for footer, header and Fab buttons)

This avoids creating 3 instances of the hook `useAnimatedStyle` of which only one is actually needed every time the hook is used, and reduces the first render time for components using the transitions on low end Android devices.

I tested for regressions for both iOS and web too, and found none, hence I updated the `.web` components aswell for consistency, but I'm not 100% familiar with simultaneous web/mobile development using RN so let me know if it's ok.

Also, I preferred to keep 3 hooks in the base `hooks` folder but I'm open to equivalent alternatives.

Fixes #4430


gaearon:
I think it would be a bit simpler if we kept them in the same file. Makes it easier to add/delete variants.

gaearon (approved):
this is a noticeable improvement on my galaxy a12 — not just on the profile screen but on Home as well. thank you!

frncs-eu:
Argh, I'm sorry you had to make the requested changes yourself @gaearon!
I was at work and planning to make them when at home but...you were faster 😛 

Anyway, I'm happy I helped improving the performance a bit! 🚀

gaearon:
no worries!

PR #4452, opened 2024-06-10
replace graphemer by unicode-segmenter

cometkim:
Same as https://github.com/bluesky-social/atproto/pull/2563

I also noticed that `graphemer` in social-app isn't specified in the `package.json`, but imported as a phantom dependency from the atproto package, which could be broken as the atproto package bumped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,6 +195,7 @@
     "statsig-react-native-expo": "^4.6.1",
     "tippy.js": "^6.3.7",
     "tlds": "^1.234.0",
+    "unicode-segmenter": "^0.7.0",
     "zeego": "^1.6.2",
     "zod": "^3.20.2"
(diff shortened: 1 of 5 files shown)


cometkim:
fyi the perf benchmark for React Native (hermes) target

https://github.com/cometkim/unicode-segmenter/pull/47

cometkim:
Ok, unicode-segmenter is in transitive dependencies, by `@atproto/lex-data`.

So, some of here can be replaced by `@atproto/lex-data` API, but some are not since it doesn't expose segmenting API as-is.

mozzius:
Superceded by https://github.com/bluesky-social/social-app/pull/9526 (which is basically identical) - thank you!! great library

PR #4456, opened 2024-06-10
Throttle instead of debounce

mozzius:


gaearon (approved):

gaearon (changes requested):
this looks like a noop to me on closer inspection

gaearon (approved):

PR #4457, opened 2024-06-10
Option to disable native translations

mozzius:
This PR does three things

- Falls back to google translate if the user's preferred language is not supported
- Cleans up duplicated translation logic into a neat little hook called `useTranslate`
- Adds an option to disable native translations altogether

simulator_screenshot_A9A6B60D-C9CA-4A8B-9129-AC75C1CDB6C7

### DIFF
--- a/src/view/screens/LanguageSettings.tsx
+++ b/src/view/screens/LanguageSettings.tsx
@@ -12,6 +12,10 @@ import {useFocusEffect} from '@react-navigation/native'
 import {sanitizeAppLanguageSetting} from '#/locale/helpers'
 import {useModalControls} from '#/state/modals'
 import {useLanguagePrefs, useLanguagePrefsApi} from '#/state/preferences'
+import {
+  useNativeTranslationDisabled,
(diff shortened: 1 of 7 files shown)


pfrazee:
If we _have_ to do a translation-service preference, it should be a selector between iOS and Google Translate

src/view/screens/LanguageSettings.tsx
```diff
@@ -302,7 +309,51 @@ export function LanguageSettingsScreen(_props: Props) {
...
+        {isNativeTranslationAvailable && (
+          <View style={{paddingBottom: 20}}>
+            <Divider style={{marginBottom: 20}} />
+
+            <Text type="title-sm" style={[pal.text, s.pb5]}>
+              <Trans>Native Translations</Trans>
+            </Text>
+            <Text style={[pal.text, s.pb10]}>
+              <Trans>
+                Translations are provided the operating system, where available.
```
surfdude29, line 323:
```suggestion
                Translations are provided by the operating system, where available.
```
I think **by** is missing here

PR #4458, opened 2024-06-10
Revert to old modal on android

mozzius:
Too many problems in the end, this just feels nicer


gaearon (changes requested):
seems a bit broken for me on device

https://github.com/bluesky-social/social-app/assets/810438/aa1624dc-cd4e-4df6-8f0c-a424a1a43e50

gaearon (approved):

PR #4461, opened 2024-06-10
Override legacy language codes for indonesian, hebrew, and yiddish

pfrazee:
Maps some old language codes that still get reported by Java (and thus Android)

https://stackoverflow.com/questions/55955641/correct-locale-for-indonesia-id-id-vs-in-id

> it is not a bug in Java, but "the problem" with backward compatibility. Java uses the first version of ISO 639. Later the standard has been updated, and some codes have been updated. Java was designed as fully backward compatible, so the authors decided to not update that codes. It is the cause why "id_ID" is changed to "in_ID". Indonesian is not the only code which is used in an old form. At least Hebrew and Yiddish are also used in the old form.

This is supported in https://xml.coverpages.org/iso639a.html

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -134,6 +134,7 @@ export function sanitizeAppLanguageSetting(appLanguage: string): AppLanguage {
       case 'hi':
         return AppLanguage.hi
       case 'id':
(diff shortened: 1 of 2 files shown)


gaearon (approved):
wow

PR #4462, opened 2024-06-10
Add support for new-tab clicks on feeds

pfrazee:
I noticed that meta-clicking on feed cards wasn't working. Factored out the logic from the generic Link component and added it to FeedSourceCard

### DIFF
--- a/src/view/com/feeds/FeedSourceCard.tsx
+++ b/src/view/com/feeds/FeedSourceCard.tsx
@@ -1,5 +1,12 @@
 import React from 'react'
-import {Pressable, StyleProp, StyleSheet, View, ViewStyle} from 'react-native'
+import {
+  Linking,
+  Pressable,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+} from 'react-native'
 import {AtUri} from '@atproto/api'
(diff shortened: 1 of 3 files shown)


src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -203,17 +211,30 @@ export function FeedSourceCardLoaded({
...
+        onPress={e => {
+          const shouldOpenInNewTab = shouldClickOpenNewTab(e)
           if (feed.type === 'feed') {
-            navigation.push('ProfileFeed', {
-              name: feed.creatorDid,
-              rkey: new AtUri(feed.uri).rkey,
-            })
+            if (shouldOpenInNewTab) {
+              Linking.openURL(
+                `/profile/${feed.creatorDid}/feed/${new AtUri(feed.uri).rkey}`,
```
gaearon, line 219: is there no one way to generate this from `ProfileFeed`? if not that's fine

gaearon (approved):

src/view/com/feeds/FeedSourceCard.tsx
pfrazee: We have a path -> route parser but not a route -> path builder

PR #4463, opened 2024-06-10
Fix: show known feed error UI more reliably

pfrazee:
Previously we were relying on the saved config to ensure the Known Feed Error UI was being given on feedgens. This updates it to work even if unsaved.

Can test with memgraph's authors feed


estrattonbailey (approved):
Oh good idea

PR #4467, opened 2024-06-10
stop using `addPushTokenListener`

haileyok:
## Why

The callback in `addPushTokenListener` is not fired once the token changes on Android. This appears to be a possible regression in `expo-notifications` - we are inquiring on that now.

What we know:
- Notifications.getDevicePushTokenAsync() still works on both iOS and Android
- The callback in addPushTokenListener() does get called on iOS
- The callback does not get called on Android

What is unclear:
- If this is actually expected behavior. I do not believe it is because:
  - It certainly worked in the past (this isn't a guarantee that it's expected of course though)
  - It continues to work on iOS
  - The documentation suggests to use this callback for registering your push token.
- What can happen if the listener is not added/it is broken. see https://docs.expo.dev/versions/latest/sdk/notifications/#addpushtokenlistenerlistener re the warning about it possibly changing.

## Test Plan

Merge this in and see if notifications start coming through on some production devices.

### DIFF
--- a/patches/expo-notifications+0.28.7.patch
+++ b/patches/expo-notifications+0.28.7.patch
@@ -63,7 +63,7 @@ index f1fed19..80afe9e 100644
 
 +  @Nullable
 +  public String getChannelId() {
-+    return mTitle;
++    return mChannelId;
 +  }
 +
    @Nullable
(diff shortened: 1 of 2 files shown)


gaearon (approved):

gaearon (approved):
verified notifs work on android now

PR #4470, opened 2024-06-11
Open convo menu when tapping on chat with deleted account

estrattonbailey:
As a better interim solution, this PR opens the `ConvoMenu` on click/press.

### DIFF
--- a/src/screens/Messages/List/ChatListItem.tsx
+++ b/src/screens/Messages/List/ChatListItem.tsx
@@ -208,7 +208,7 @@ function ChatListItemReady({
           !isDeletedAccount
             ? _(msg`Go to conversation with ${profile.handle}`)
             : _(
-                msg`The recipient of this chat has deleted their account. Press for options.`,
+                msg`This conversation is with a deleted or a deactivated account. Press for options.`,
               )
         }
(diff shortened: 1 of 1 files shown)


src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -208,7 +208,7 @@ function ChatListItemReady({
           !isDeletedAccount
             ? _(msg`Go to conversation with ${profile.handle}`)
             : _(
-                msg`The recipient of this chat has deleted their account. Click for options.`,
+                msg`The recipient of this chat has deleted their account. Press for options.`,
```
surfdude29, line 211: Apologies for being pedantic, but aren't there also cases where the account may have been taken down by moderators or deactivated (which I'm assuming has the same result in terms of the chat being inaccessible)?

What about something like:

"**This chat is no longer available as the recipient's account has been deleted or deactivated. Press for options.**"

gaearon (approved):

src/screens/Messages/List/ChatListItem.tsx
estrattonbailey: Good call!

PR #4473, opened 2024-06-11
Revert "Revert "Fix undefined block""

gaearon:
TODO for @estrattonbailey to revisit


estrattonbailey:
Re-did in #4479 to exclude the hover card stuff

PR #4474, opened 2024-06-11
Request permission to badge the app

haileyok:
## Why

Unclear why this works for me without allowing this permission - maybe there's a system default? - but Expo docs say you need to request permission to badge the app: https://docs.expo.dev/versions/latest/sdk/notifications/#setbadgecountasyncbadgecount-options

## Test Plan

@gaearon sees if he receives iOS badges after this PR.


haileyok:
Wasn't necessary it seems.

PR #4476, opened 2024-06-11
[Session] Logging

gaearon:
This adds enough logging to key parts of the session logic that we should be able to figure out what's going on if something goes wrong. Note that agents are numbered (using WeakMap), tokens are sent as hashes (so we can trace what happens to each of them over time).

- [x] Basic implementation
- [x] Add gate (by stable ID)
- [x] Send logs when the gate is on

## Test Plan

Allowlist your stable ID (`Statsig.getStableID()`) for the `debug_session` gate. We'll only enable it for the few users who reported problems. Observe events rolling in. (Don't forget to turn on "Show non-production logs"!)

Screenshot 2024-06-20 at 05 33 32

Screenshot 2024-06-20 at 05 33 40

Note that there's a notion of separate "realms" (`realmId`) which correspond to browser tabs or app instances, a notion of separate "messages" (which are JSON objects describing what happened), and "slices" (which is an artifact of the transport — Statsig forces us to break apart the logs into smaller chunks or it'll refuse to send them). Since this is all for manual debugging, I think that's fine. I might write a script to reconstruct it in a readable form if needed.

Also note that this is done by stable ID so it's able to debug across account switches and logouts.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -19,6 +19,7 @@ import {
 import {getInitialState, reducer} from './reducer'
 
 export {isSignupQueued} from './util'
+import {addSessionDebugLog} from './logging'
 export type {SessionAccount} from '#/state/session/types'
 import {SessionApiContext, SessionStateContext} from '#/state/session/types'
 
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
Looks great, haven't run yet though

pfrazee (approved):
Okay -- dont have strong opinions about how youve constructed the data you're sending up, but it does seem to function as intended. I was able to confirm the data is sending up by opting into the experiment.

PR #4478, opened 2024-06-11
`Reposted by you` and `Reply to you`

quiple:
If the reposted by the signed-in user or is a reply to the signed-in user, marks it as `Reposted by you` or `Reply to you`.


haileyok (approved):
Thanks! Just change naming of the variable to match (most) other places we do the same thing. Appreciate it!

PR #4479, opened 2024-06-11
Fix undefined block

estrattonbailey:
[Review without whitespace](https://github.com/bluesky-social/social-app/pull/4479/files?diff=split&w=1)

Simplifies this handling a bit after some discussion.

In production right now, blocks between users appear here as `Reply to undefined`, which is broken behavior. In post threads, posts display as `Blocked post` and they are non-interactive. You can't click into this user or otherwise follow the post through the block.

This PR improves upon this by replacing `undefined` with `a blocked post`, to mirror the post thread. It also removes the profile hover on web, and the link around the reply text. This way a user seeing this `Reply to a blocked post` cannot hover or navigate to this user: it's non-interactive. When clicking into the reply, they'll see our existing `Blocked post` handling in the thread view.

CleanShot 2024-06-11 at 09 51 45@2x

### DIFF
--- a/src/view/com/posts/FeedItem.tsx
+++ b/src/view/com/posts/FeedItem.tsx
@@ -204,8 +204,6 @@ let FeedItemInner = ({
     ) ||
     (!parentAuthor?.displayName && !parentAuthor?.handle)
 
-  console.log({isParentBlocked, parentAuthor})
-
   return (
     <Link
       testID={`feedItem-by-${post.author.handle}`}
--- a/src/view/com/posts/FeedItem.tsx
+++ b/src/view/com/posts/FeedItem.tsx
@@ -56,6 +56,7 @@ interface FeedItemProps {
   isThreadParent?: boolean
   feedContext: string | undefined
   hideTopBorder?: boolean
+  isParentBlocked?: boolean
 }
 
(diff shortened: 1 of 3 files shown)


src/view/com/posts/FeedItem.tsx
```diff
@@ -196,6 +196,14 @@ let FeedItemInner = ({
     },
   ]
 
+  const isParentBlocked =
+    Boolean(
+      parentAuthor?.viewer?.blockedBy ||
+        parentAuthor?.viewer?.blocking ||
+        parentAuthor?.viewer?.blockingByList,
+    ) ||
+    (!parentAuthor?.displayName && !parentAuthor?.handle)
```
gaearon, line 205: This check feels a bit weird, what does the server give us?

gaearon (approved):
ok but i don't love the check, is the server handing us a bad object?

src/view/com/posts/FeedItem.tsx
estrattonbailey: Good question, I should have looked deeper before. This was a hack for a 3p viewing a reply where the author and OP have a block between them. See latest commit, gonna ping in Slack.

PR #4481, opened 2024-06-11
register the push token after request has been approved on android

haileyok:
## Why

Temporary adjustment for now until we can mark messages read on the server and update the count. This will just decrement the badge count by the number of unread messages in the conversation when you open it.

Far from perfect, but given we're going to actually fix this in earnest in the nearish future, let's just make the badge clear for now and avoid digging too deeply at something that could cause more breaks.

I'm also adding one more fallback for push notification registration given the current expo bug. It shouldn't result in two push notification token registrations, but if it does that's fine too. Better to ensure their token registered right after onboarding.

## Test Plan

- Receive DM notification on Android
- Open the conversation
- See the badge has cleared


pfrazee (approved):
LGTM

PR #4491, opened 2024-06-11
Redo explore page

estrattonbailey:
Stacked on #4501 

CleanShot 2024-06-13 at 11 54 03@2x
CleanShot 2024-06-13 at 12 21 39@2x

CleanShot 2024-06-11 at 18 43 29@2x
CleanShot 2024-06-11 at 18 42 45@2x


src/state/queries/feed.ts
```diff
@@ -299,6 +299,34 @@ export function useSearchPopularFeedsMutation() {
   })
 }
 
+const popularFeedsSearchQueryKeyRoot = 'actor-search'
```
pfrazee, line 302: I think you copy-pasted the wrong key name

src/state/queries/suggested-follows.ts
```diff
@@ -31,7 +31,7 @@ const suggestedFollowsByActorQueryKey = (did: string) => [
   did,
 ]
 
-export function useSuggestedFollowsQuery() {
+export function useSuggestedFollowsQuery(options?: {limit?: number}) {
```
pfrazee, line 34: Probably gotta update the query key to include the param

pfrazee:
Hot dog! Got a couple UI thoughts:

CleanShot 2024-06-14 at 09 57 46@2x

On that avi spacing, since we're not gonna get text alignment, I figure why not just give it all more spacing? Could also drop from 4 items to 3 if needed.

CleanShot 2024-06-14 at 10 00 02@2x

I'm really not vibing on the blue border here. I find it really pulls my eye away from the content and doesn't help structure the IA for me.

src/state/queries/feed.ts
estrattonbailey: Ah good catch

src/state/queries/suggested-follows.ts
estrattonbailey: Good call

estrattonbailey:
CleanShot 2024-06-14 at 12 28 20@2x

pfrazee (approved):
Good stuff

PR #4493, opened 2024-06-12
Make the word order of the string for social proof changeable

tkusano:
In social-proof display, the list of accounts and number of people comes after "Followed by". But in Japanese, the translation equivalent to "Followed by" would be more natural to have it after the list of accounts, so I would like to see it changed so that the entire message is surrounded by `<Trans>...</Trans>`, not just `Followed by`.


tkusano:
PR #4494 is more better. Close this PR.

PR #4494, opened 2024-06-12
KnownFollowers localization

quiple:
Provides full context to the KnownFollowers string to allow for easier localization.

```po
#: src/components/KnownFollowers.tsx:169
#~ msgid "Followed by"
#~ msgstr ""

#: src/components/KnownFollowers.tsx:179
#~ msgid "{0, plural, one {and # other} other {and # others}}"
#~ msgstr ""

#: src/components/KnownFollowers.tsx:194
msgid "Followed by <0>{0}</0>"
msgstr ""

#: src/components/KnownFollowers.tsx:183
msgid "Followed by <0>{0}</0> and <1>{1}</1>"
msgstr ""

#: src/components/KnownFollowers.tsx:170
msgid "Followed by <0>{0}</0>, <1>{1}</1>, and {2, plural, one {# other} other {# others}}"
msgstr ""
```

### DIFF
--- a/src/components/KnownFollowers.tsx
+++ b/src/components/KnownFollowers.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import {View} from 'react-native'
 import {AppBskyActorDefs, moderateProfile, ModerationOpts} from '@atproto/api'
-import {msg, plural, Trans} from '@lingui/macro'
+import {msg, Plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
 import {makeProfileLink} from '#/lib/routes/links'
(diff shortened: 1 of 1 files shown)


mary-ext:
would be curious to know if [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) is supported on React Native since that seems like to be the better way of handling this situation

quiple:
> would be curious to know if [`Intl.ListFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) is supported on React Native since that seems like to be the better way of handling this situation (unless the following result isn't correct)
> 
> ```
> vehicles = ['Motorcycle', 'Bus', 'Car'];
> formatter = new Intl.ListFormat('en');
> formatter.format(vehicles);
> // en: "Motorcycle, Bus, and Car"
> // jp: "Motorcycle、Bus、Car"
> // ko: "Motorcycle, Bus 및 Car"
> ```

Wow, this looks good, but I don't think it would fit for Korean if the last item in the list is "# others".

mary-ext:
Ah I see, how would a Korean translation look like for the entire string?

quiple:
In Korean, this is fine if it's just listing account names, but it's a little different if you have "# others".

```po
#: src/components/KnownFollowers.tsx:194
msgid "Followed by <0>{0}</0>"
msgstr "<0>{0}</0> 님이 팔로우함"

#: src/components/KnownFollowers.tsx:183
msgid "Followed by <0>{0}</0> and <1>{1}</1>"
msgstr "<0>{0}</0> 및 <1>{1}</1> 님이 팔로우함"

#: src/components/KnownFollowers.tsx:170
msgid "Followed by <0>{0}</0>, <1>{1}</1>, and {2, plural, one {# other} other {# others}}"
msgstr "<0>{0}</0>, <1>{1}</1> 님 외 {2, plural, other {#}}명이 팔로우함"
```

gaearon (approved):

gaearon:
thank you!

PR #4496, opened 2024-06-12
Fix scrollbars on windows

mozzius:
- Fixes extra scrollbar on windows
- Stops the "Search for..." bit in the mini search results being hidden behind a spinner

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>


AdrianoCahete:
Why not put the vertical scrollbar only on list itself (ignoring the search box)?

And I think that you can customize the scrollbar visual in most recent browsers without big problem

PR #4499, opened 2024-06-12
Just use server count

estrattonbailey:
Was trying to be clever, but this logic was clearly flawed: the server only hydrates 5 profiles, and if any blocks exists, it filters those out. Adjusting the count here on the client isn't possible to do correctly, since there could still be 300 other follows in common that weren't hydrated.

Also, the count wasn't reflecting the few that show up as previews, so `and n others` was always wrong.

So we'll still rarely have the issue where we count 300 known followers, but all 300 are blocked. For this, `KnownFollowers` simply won't be rendered. But if say, 2 are not blocked, you'll get `followed by X, Y, and 298 others`, and when clicking into to view all, you'll see only those 2.

### DIFF
--- a/src/components/KnownFollowers.tsx
+++ b/src/components/KnownFollowers.tsx
@@ -82,7 +82,6 @@ function KnownFollowersInner({
     t.atoms.text_contrast_medium,
   ]
 
-  const count = cachedKnownFollowers.count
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
LGTM

PR #4500, opened 2024-06-12
Calculate correct keyboard offset in composer

mozzius:
`KeyboardAvoidingView` needs a an offset in the new composer modal. Currently, it's hardcoded with a value of 70, which works for dynamic island iPhones. However, it does not work for other models (see the SE below, it also doesn't quite work on my 14).

This is because it needs to be set to the height of this area, above the sheet:
Screenshot 2024-06-12 at 23 21 07

This PR calculates it by getting the height of the sheet and subtracting that value from the screen height. This produces 69 on the iPhone 15 and 40 on the iPhone SE, which just about matches the hardcoded value.

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
    <td>After</td>
  </tr>
</table>

# Test plan

Check it looks good on all 3 types of iOS device (no notch, notch, dynamic island)
Should be unchanged on android

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -14,6 +14,7 @@ import {
   LayoutChangeEvent,
   StyleSheet,
   TouchableOpacity,
+  useWindowDimensions,
   View,
 } from 'react-native'
 import Animated, {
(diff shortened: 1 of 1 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -741,6 +750,19 @@ function useAnimatedBorders() {
   }
 }
 
+function useKeyboardVerticalOffset() {
+  const {height: windowHeight} = useWindowDimensions()
+
+  const [viewHeight, setViewHeight] = useState(0)
+
+  return [
+    isIOS ? windowHeight - viewHeight : 0,
```
gaearon, line 759: `onLayout` is async so `viewHeight` will be `0` on first render. How are we handling this? Why doesn't this result in flicker?

mozzius: Good question - I think it's hidden by the animation, but we can guard against this potential problem by defaulting to `windowHeight - 69`, meaning the value is 69 on first render (same as if we'd hard coded it)

```js
const {height: windowHeight} = useWindowDimensions()
// estimate height to avoid potential jumping on first render
const [viewHeight, setViewHeight] = useState(windowHeight - 69)
```

mozzius: Screenshot 2024-06-13 at 01 05 10

mozzius: (it's really dumb, I know)

gaearon: Sorry I keep asking silly questions but I want to understand:

- Does this height depend on any UI elements in the app (e.g. our modal styling) or is it platform-specific?
- Is this height the same across all devices, or does it depend on device?
- On a single device, is it consistent for this screen or does it depend on something?
- If it doesn't depend on the actual UI and is specific to the device, is there any way to measure it beforehand so that by this point it's always available?

mozzius: 1. it's platform specific (specifically this native sheet presentation)
2. depends on device. the iPhone SE has no notch, so this area is small (40px). the notched iPhones are ~60px. the newest iPhones with the pill-shaped notch have the largest area of 69px
3. I'd expect it to stay the same. maybe rotating an iPad will change it?
4. I found this issue: https://github.com/react-navigation/react-navigation/issues/10317
Someone claims you can just add 10px to the top safe area inset to get the right value, so I'm gonna see if that works

mozzius: Update: I was able to simplify it to this:

```js
function useKeyboardVerticalOffset() {
  const {top} = useSafeAreaInsets()

  // Android etc
  if (!isIOS) return 0

  // iPhone SE
  if (top === 20) return 40

  // all other iPhones
  return top + 10
}
```

Weirdly the iPhone SE needs an extra 10px, but overall huge improvement.

Thanks for calling out my overly complex approach!

gaearon (approved):
code lgtm but i have not tested, pls merge after testing on each

mozzius:
tested on each, looks good

PR #4501, opened 2024-06-12
DRY up popular feeds query

estrattonbailey:
The popular feeds query returns a static list of feeds. Some are not available logged out, and it returns feeds even if you've already pinned them. We had logic on the Feeds screen to filter based on session state and by which feeds you have saved.

I needed to reuse the same logic in #4491 and seemed like we could push this logic into the query handler itself.

This means that **the popular feeds query will now return results with the feeds you've saved filtered out.** We don't use this for anything else at the moment, and if we do need this, it'd be easy enough to add a param to this query to conditionally filter.

The only semi-noticeable change that this PR introduces is over-fetching if the first page of results doesn't return enough feeds that aren't already pinned. It reuses the exact pattern we used in `post-feed` to accomplish this. This will be used on the Explore page to show ~3 feeds on the first page of results, but that first page is likely to contain some feeds the user already has pinned, thus the need for over-fetching to ensure we get at least 3 (it may be more e.g. 5 if first page has 2 un-pinned feeds and second page has a full 3).

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -1,3 +1,4 @@
+import {useCallback, useEffect, useMemo, useRef} from 'react'
 import {
   AppBskyActorDefs,
   AppBskyFeedDefs,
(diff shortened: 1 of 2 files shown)


src/view/screens/Feeds.tsx
```diff
@@ -327,37 +311,19 @@ export function FeedsScreen(_props: Props) {
             type: 'popularFeedsLoading',
           })
         } else {
-          if (
-            !popularFeeds?.pages ||
-            popularFeeds?.pages[0]?.feeds?.length === 0
-          ) {
+          if (!popularFeeds?.pages) {
```
estrattonbailey, line 314: First page _can_ be empty, but subsequent pages will have results

pfrazee (approved):
Reviewed code, tested on web and ios, looks good

src/state/queries/feed.ts
```diff
@@ -171,28 +172,117 @@ export function useFeedSourceInfoQuery({uri}: {uri: string}) {
...
       return res.data
     },
     initialPageParam: undefined,
     getNextPageParam: lastPage => lastPage.cursor,
+    select: useCallback(
+      (
+        data: InfiniteData<AppBskyUnspeccedGetPopularFeedGenerators.OutputSchema>,
+      ) => {
+        const {savedFeeds, hasSession: hasSessionInner} = selectArgs
+        data?.pages.map(page => {
```
gaearon, line 235: This is a mutation so it's not really selecting. Why is that?

PR #4502, opened 2024-06-13
Show social proof in hovercards

gaearon:
Just copy-paste from the profile. Web-only.

Screenshot 2024-06-13 at 03 51 57

Doesn't show up if there's no social proof.


estrattonbailey (approved):
Yesssss

PR #4504, opened 2024-06-13
Hide bio and social proof for blocked users

gaearon:
Twitter doesn't display them. I'm not sure we should either. Showing social proof definitely feels off on a block. For bio, I could see it either way, but I lean towards not displaying it either if there's a block relationship.

https://github.com/bluesky-social/social-app/assets/810438/c590d632-a6d9-443a-ba31-5aef0a2b4deb


pfrazee (approved):
I see no issue with this

FleetAdmiralJakob:
But the data is still sent to the client so someone still could use the data even if there is a block relationship right?

PR #4507, opened 2024-06-13
Add `native_pwi_disabled` feature gate experiment

estrattonbailey:
This adds an experiment to the logged-out view on native only. To do so, we need to fetch Statsig config immediately. So, if we have no user to resume from, we do a full load of Statsig data to ensure we have accurate config based on the stable device ID.

To do this, I tested calling `initialize` with `prefetchUsers` on the options object, but this failed all three tests I did, in that it didn't load the config in time. So I opted to adjust the API of `tryFetchGates` to allow `undefined` as the DID value, [which will then fall back](https://docs.statsig.com/client/jsClientSDK#statsig-user) to using the stable device ID.

This works in my testing every time:
- delete app to clear storage
- re-run `yarn ios`
- see that the config in Statsig dashboard is correct
  - I've been using 100% on or 100% off
- repeat

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -305,7 +305,7 @@ export function Provider({children}: {children: React.ReactNode}) {
     <GateCache.Provider value={gateCache}>
       <StatsigProvider
         key={did}
-        sdkKey="client-SXJakO39w9vIhl3D44u8UupyzFl4oZ2qPIkjwcvuPsV"
+        sdkKey={SDK_KEY}
         mountKey={currentStatsigUser.userID}
         user={currentStatsigUser}
         // This isn't really blocking due to short initTimeoutMs above.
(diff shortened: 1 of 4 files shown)


src/lib/statsig/statsig.tsx
```diff
@@ -14,6 +14,8 @@ import {useNonReactiveCallback} from '../hooks/useNonReactiveCallback'
 import {LogEvents} from './events'
 import {Gate} from './gates'
 
+const SDK_KEY = 'client-SXJakO39w9vIhl3D44u8UupyzFl4oZ2qPIkjwcvuPsV'
```
gaearon, line 17: mind using it below in the Provider? it's hardcoded there too

gaearon (approved):
code lgtm, small nit

src/lib/statsig/statsig.tsx
estrattonbailey: Ah yes, good call

PR #4509, opened 2024-06-13
add `document.referrer` to statsig custom

haileyok:
## Why

Want to add referrer to statsig events. We currently have `refUrl`, but that requires a URL parameter that doesn't get set outside of our embeds.

## How

On web, let's add the value of `document.referrer` to the statsig session `custom` object.

## Test Plan

I modified a link on google to direct me to `http://localhost:19006` with a console log inside of `statsig.tsx`.

Screenshot 2024-06-13 at 12 06 38 PM

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -27,6 +27,7 @@ type StatsigUser = {
     refSrc: string
     refUrl: string
     referrer: string
+    referrerHostname: string
     appLanguage: string
     contentLanguages: string[]
   }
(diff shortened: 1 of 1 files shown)


src/lib/statsig/statsig.tsx
```diff
@@ -26,19 +26,31 @@ type StatsigUser = {
...
 let refUrl = ''
+let referrer = ''
+let referrerHostname = ''
 if (isWeb && typeof window !== 'undefined') {
   const params = new URLSearchParams(window.location.search)
   refSrc = params.get('ref_src') ?? ''
   refUrl = decodeURIComponent(params.get('ref_url') ?? '')
 }
 
+if (isWeb && document !== 'undefined' && document != null) {
```
gaearon, line 46: You want `typeof document !== 'undefined`, not `document !== 'undefined'`

haileyok: sigh lol, nice

gaearon (approved):

PR #4510, opened 2024-06-13
Verify email reminders

estrattonbailey:
This PR adjusts the email verification reminder period to once per day, which is up from once all-time.

It does this by storing a date string with to-the-day resolution so that each date can be compared directly e.g. `'2024-6-13' === '2024-6-13'`. If the reminder needs to be shown, it returns `true` and snoozes, which sets the stored string to the current date.

As existing, once `account.emailConfirmed` is true, this process stops entirely.

Additionally, this PR snoozes the confirm check immediately upon account creation, so that users aren't prompted to verify until 1 calendar date later.


src/lib/strings/time.ts
```diff
@@ -61,3 +61,10 @@ export function getAge(birthDate: Date): number {
   }
   return age
 }
+
+export function toSimpleDateString(date: Date) {
+  const mm = date.getMonth() + 1 // 0-indexed
+  const dd = date.getDate()
+  const yyyy = date.getFullYear()
+  return `${yyyy}-${mm}-${dd}`
```
gaearon, line 69: This might end up being something like `2024-2-4` which is not universally supported when we feed it to `new Date` below. (E.g. `new Date(`2024-2-4`)` in Safari is `Invalid Date`.

src/state/shell/reminders.ts
```diff
@@ -1,36 +1,55 @@
...
-  const today = new Date()
-  // shard the users into 2 day of the week buckets
-  // (this is to avoid a sudden influx of email updates when
-  // this feature rolls out)
-  const code = toHashCode(account.did) % 7
-  if (code !== today.getDay() && code !== (today.getDay() + 1) % 7) {
-    return false
+
+  // snoozed recently
+  if (snoozedAt !== today) {
```
gaearon, line 36: Let's just remove this condition, the above early return rules the rest out

gaearon (changes requested):
Noted a potential bug in Safari stemming from custom string representation. It's possible to fix but IMO would be more sturdy if we relied on standard stringify/parse and then explicitly compared year/month/date between two Date objects. How do you feel about that?

gaearon (approved):

PR #4512, opened 2024-06-13
Feed source card

estrattonbailey:
Refactors `FeedSourceCard` into `FeedCard`.

Why? Because `FeedSourceCard` loads its own data, and in my cases (like on the Feeds screen) we already have data from the list endpoint. In fact, sometimes these `n+1` loads fail, and we leave the user with a loading state.

The old component is used a handful of places, so a re-implementation is the easiest way to replace.

This PR adds new "macro" components to replace `FeedSourceCard` that allow different compositions e.g. no likes data, or save instead of pin, without needed to bloat the main API contract.

However, this PR defines a `Default` export that is essentially a direct replacement for `FeedSourceCard` as it is today.

This PR does _not_ reimplement the placeholder states for the card itself, since it's not strictly needed in this PR. Future PRs will replace other varied usage of `FeedSourceCard` and create composable skeleton UI where appropriate.

CleanShot 2024-06-13 at 16 44 14@2x
CleanShot 2024-06-13 at 16 44 36@2x

### DIFF
--- a/src/components/FeedCard.tsx
+++ b/src/components/FeedCard.tsx
@@ -1,0 +1,219 @@
+import React from 'react'
+import {GestureResponderEvent, View} from 'react-native'
+import {AppBskyActorDefs, AppBskyFeedDefs, AtUri} from '@atproto/api'
+import {msg, plural, Trans} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+
+import {logger} from '#/logger'
+import {
+  useAddSavedFeedsMutation,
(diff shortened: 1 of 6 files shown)


pfrazee (approved):
Didnt test but code LGTM

haileyok (approved):
Heh, was just testing myself. If code looks good then testing looks good to me.

Side note: we should do this for the feeds page too. Each of the little feed items that I have pinned has this weird loading.

PR #4514, opened 2024-06-14
referrers for all platforms

haileyok:
## Why

We currently are not able to track referrers if the user is on native. This PR introduces a way for us to - when possible - see those referrals.

## How

There's a little bit of native code needed on each platform.

### iOS 

On iOS, we can get the `referrerURL` from `NSUserActivity`. See https://developer.apple.com/documentation/foundation/nsuseractivity/2875762-referrerurl

This requires a little plugin to add some code to our `AppDelegate.mm`. Whenever a user opens the app through a Universal Link, we can - when it is provided by the browser - record the referrer URL here. To make things simple and reduce the scope of the plugin, I am just storing the URL to `NSUserDefaults` and then retrieving it in our module.

We also are able to get the same for deep links, also in the `AppDelegate`. See https://developer.apple.com/documentation/uikit/uiapplicationopenurloptionssourceapplicationkey

The same process is used here: save the value to `NSUserDefaults` and retrieve it through our module.

The build plugin in this PR adds the necessary lines to the `AppDelegate.mm` file to achieve this.

### Android

On Android, we need to do two things to get this information:

- Check incoming intents for `Intent.EXTRA_REFERRER`, which will be the URI of the website that the request is coming from - if it is coming from one.
- Otherwise, check the current activity's `referrer` property, which will tell us the _application_ the request was made from. 

Since we can access these straight from the module code, all we do is store the intent whenever it comes in and return it to the JS side when calling `getReferrerAsync()`.

The `getGooglePlayReferrerInfoAsync()` call uses the Google Play Install Referrer API (see https://developer.android.com/google/play/installreferrer) to retrieve the URL that referred the user to the app store. This will be used for starter packs. I've already added this code to the starter packs PR, but once we merge that in and later review this, I'll just migrate the code to use this single module instead.

### Web

Same process as what exists in the app right now, but just made to follow the same API as iOS and Android.

### Sending to Statsig

On iOS and Android, we will send this info to Statsig in the `useIntentHandler` hook, whenever we receive a new link into the app.

On web, I've moved this from being a custom user property to being it's own event - again to align with how we handle this on iOS and Android. Whenever send the `init` event to Statsig, we will also send this event if needed.

## Test Plan

1. Ensure the build plugin functions. Check the `AppDelegate.mm` to make sure it looks good (there are commented blocks that signify where code is being added). App should still build on iOS.
2. Test web:
  - You can do this with a little console log and inspect-elementing a URL on another site to lead to Bluesky.
3. Test Android:
  - You can add a console log into `useIntentHandler` to check the status
Screenshot 2024-06-13 at 10 35 22 PM
Screenshot 2024-06-13 at 10 35 40 PM
4. Test iOS:
  - A good bit harder to test this one since you need to be using a physical device - or at least have it built with the correct provisioning profile - to test universal links. We should verify through Statsig that we are receiving good info from iOS devices.


gaearon (commented):
gonna wait with this until after sp

modules/expo-bluesky-swiss-army/android/src/main/java/expo/modules/blueskyswissarmy/referrer/ExpoBlueskyReferrerModule.kt
```diff
@@ -8,9 +11,49 @@ import expo.modules.kotlin.modules.ModuleDefinition
...
+        intent = null
+        return@AsyncFunction res
+      }
+
+      // In all other cases, we'll just record the app that sent the intent.
+      if (activityReferrer != null) {
+        // referrer could become null here. `.toString()` though can be called on null
+        val res = mapOf(
+          "referrer" to activityReferrer.toString(),
+          "hostname" to (activityReferrer?.host ?: ""),
```
haileyok, line 48: see https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to-string.html

gaearon (approved):

mary-ext:
seems like this broke web?

image

image

PR #4518, opened 2024-06-14
Server-side thread mutes

mozzius:
Thread mutes are now applied server-side and stop notifications from being sent out altogether.

Client implementation-wise, I added a new cache of `Record<string, boolean>` for optimistically setting muted threads. The "mute thread" button first looks there, and if no value is present looks at the record viewer state (this is because all posts in a thread share the same value, so I wasn't able to use the post shadow system). Mutations are queued, like likes are.

> Note: muted thread will no longer retroactively remove old notifications from the notification list


src/state/cache/thread-mutes.tsx
```diff
@@ -0,0 +1,43 @@
...
+type StateContext = Record<string, boolean>
+type SetStateContext = (uri: string, value: boolean) => void
+
+const stateContext = React.createContext<StateContext>({})
+const setStateContext = React.createContext<SetStateContext>(
+  (_: string) => false,
+)
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const [state, setState] = React.useState<StateContext>({})
```
gaearon, line 12: Any objections to using a Map? With lazy init, `useState(() => new Map())`. Generally Maps are better than objects when your keys are arbitrary.

src/state/cache/thread-mutes.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+type StateContext = Record<string, boolean>
+type SetStateContext = (uri: string, value: boolean) => void
+
+const stateContext = React.createContext<StateContext>({})
+const setStateContext = React.createContext<SetStateContext>(
+  (_: string) => false,
+)
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
```
gaearon, line 11: This provider is currently rendered _above_ the part that we key on `did`. Should it be moved to be below now? I assume the setting is per-account rather than per-device now?

gaearon (changes requested):
the code overall makes sense to me. have not QA'd.

i think the part about where the provider is placed is important, but lmk if not and i missed something

gaearon (approved):

PR #4520, opened 2024-06-14
Resolve patch-package warnings

gaearon:
Seems like straightforward bumps. Resolves warnings after yarn.

## Test Plan

👀🤔✅


mozzius (approved):
reading diffs of patch files is the worst

PR #4523, opened 2024-06-14
Migrate local thread mutes

mozzius:
# Stacked on https://github.com/bluesky-social/social-app/pull/4518

Simple. Uploads existing thread mutes then deletes them.

Does not account for multi-accounts, but I'm not sure if there is a neat solution here without making this far more complex than necessary

### DIFF
--- a/src/state/cache/thread-mutes.tsx
+++ b/src/state/cache/thread-mutes.tsx
@@ -1,8 +1,7 @@
 import React, {useEffect} from 'react'
-import {BskyAgent} from '@atproto/api'
 
 import * as persisted from '#/state/persisted'
-import {useAgent} from '../session'
+import {useAgent, useSession} from '../session'
 
 type StateContext = Map<string, boolean>
 type SetStateContext = (uri: string, value: boolean) => void
@@ -14,13 +13,6 @@ const setStateContext = React.createContext<SetStateContext>(
 
 export function Provider({children}: React.PropsWithChildren<{}>) {
   const [state, setState] = React.useState<StateContext>(() => new Map())
-  const agent = useAgent()
-
-  useEffect(() => {
-    if (agent.hasSession) {
-      migrateThreadMutes(agent)
-    }
-  }, [agent])
 
   const setThreadMute = React.useCallback(
     (uri: string, value: boolean) => {
@@ -32,6 +24,9 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     },
     [setState],
   )
+
+  useMigrateMutes(setThreadMute)
+
   return (
     <stateContext.Provider value={state}>
       <setStateContext.Provider value={setThreadMute}>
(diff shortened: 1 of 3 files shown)


src/state/cache/thread-mutes.tsx
```diff
@@ -41,3 +50,18 @@ export function useIsThreadMuted(uri: string, defaultValue = false) {
 export function useSetThreadMute() {
   return React.useContext(setStateContext)
 }
+
+function migrateThreadMutes(agent: BskyAgent) {
+  const threadMutes = persisted.get('mutedThreads')
+  if (threadMutes.length > 0) {
+    console.log('migrating', threadMutes.length, 'thread mutes')
```
gaearon, line 59: log

gaearon (changes requested):
if you have a thousand (as some older heavy users might), i think this version will fire off a thousand concurrent requests. would that result in a rate limit?

not sure how to solve but let's think about what we want to happen in this case

mary-ext:
Might be best to migrate the newest mutes first, and to update the array as the migration progresses? (if there's a lot of mutes then it's possible that the migration won't be finished by the time the user's done with the app and closes it)

src/state/cache/thread-mutes.tsx
```diff
@@ -0,0 +1,87 @@
...
+          if (!root) break
+
+          setThreadMute(root, true)
+
+          await agent.api.app.bsky.graph
+            .muteThread({root})
+            // not a big deal if this fails, since the post might have been deleted
+            .catch(console.error)
+
+          persisted.write('mutedThreads', threads)
```
gaearon, line 76: Are multiple browser tabs going to fight with each other and try to sync this value with each other? I wonder if we should just skip the "broadcast" here. cc @estrattonbailey for thoughts

src/state/cache/thread-mutes.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+      const migrate = async () => {
+        while (!cancelled) {
+          const threads = persisted.get('mutedThreads')
+
+          const root = threads.shift()
+
+          if (!root) break
+
+          persisted.write('mutedThreads', threads)
```
estrattonbailey, line 69: We could add a param here to not emit an update. But also, if a user reloads the app with 4 tabs open, this migration is going to fire 4 times, right?

src/state/cache/thread-mutes.tsx
```diff
@@ -0,0 +1,87 @@
...
+        }
+      }
+
+      migrate()
+
+      return () => {
+        cancelled = true
+      }
+    }
+  }, [agent, currentAccount, setThreadMute])
```
estrattonbailey, line 86: I don't _think_ `agent` or `currentAccount` can change here, but we might want to protect against this hook being called more than once?

src/state/persisted/schema.ts
```diff
@@ -88,6 +87,8 @@ export const schema = z.object({
   disableHaptics: z.boolean().optional(),
   disableAutoplay: z.boolean().optional(),
   kawaii: z.boolean().optional(),
+  // deprecated
```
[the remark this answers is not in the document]
mozzius: oh neat

src/state/cache/thread-mutes.tsx
mozzius: The loop cancels itself, so it's fine for the effect to fire multiple times

src/state/cache/thread-mutes.tsx
mozzius: The list will be processed in parallel by each active tab. It's fine if there's data races though, nothing happens if you mute the same thread twice. Also, writing to `persisted` doesn't refire the effect, so it doesn't cause multiple migrations or something

src/state/queries/post.ts
```diff
@@ -291,3 +292,72 @@ export function usePostDeleteMutation() {
...
+        await threadUnmuteMutation.mutateAsync({
+          uri: rootUri,
+        })
+        return false
+      }
+    },
+    onSuccess(finalIsMuted) {
+      // finalize
+      setThreadMute(rootUri, finalIsMuted)
+    },
```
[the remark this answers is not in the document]
mozzius: This resets it, just without feedback. Should I add feedback?

gaearon (commented):
lgtm!

gaearon (approved):
lgtm

PR #4527, opened 2024-06-15
Italian translation fix

marcomaroni-github:
Italian translation fix (feeds->feed)


src/locale/locales/it/messages.po
```diff
@@ -356,7 +356,7 @@ msgstr "Aggiunto alla lista"
 
 #: src/view/com/feeds/FeedSourceCard.tsx:126
 msgid "Added to my feeds"
-msgstr "Aggiunto ai miei feeds"
+msgstr "Aggiunto ai miei feed
```
surfdude29, line 359:
```suggestion
msgstr "Aggiunto ai miei feed"
```
I just noticed a small thing here, the end `"` is missing

src/locale/locales/it/messages.po
```diff
@@ -2202,7 +2202,7 @@ msgstr "Commenti"
 #: src/view/shell/Drawer.tsx:493
 #: src/view/shell/Drawer.tsx:494
 msgid "Feeds"
-msgstr "Feeds"
+msgstr "Feed
```
surfdude29, line 2205:
```suggestion
msgstr "Feed"
```
And here too

src/locale/locales/it/messages.po
```diff
@@ -2213,7 +2213,7 @@ msgstr "I feed sono algoritmi personalizzati che gli utenti creano con un minimo
 
 #: src/screens/Onboarding/StepTopicalFeeds.tsx:80
 #~ msgid "Feeds can be topical as well!"
-#~ msgstr "I feeds possono anche avere tematiche!"
+#~ msgstr "I feed possono anche avere tematiche!"
```
surfdude29, line 2216: Also just as a tip for the future, you don't need to bother changing any of the lines that start with `#~` as these are obsolete messages that have been removed from the codebase. So you can just ignore them or you can delete these lines if you prefer :)

pfrazee (approved):
Thanks!!

Issue #4529, opened 2024-06-16
RTL auto-detection makes PostMeta messy


jgarplind:
**Describe the bug**
RTL auto-detection causes `@`-character in PostMeta to be misplaced.

**To Reproduce**

See https://bsky.app/profile/pastures.bsky.social for example.

**Expected behavior**

`@`-character should always precede the handle, since the handle is inherently LTR-text, event if the display name is RTL-text.

**Screenshots**

![image](https://github.com/bluesky-social/social-app/assets/10069144/834bb1ab-91b8-4d43-8fea-83ceb79a3390)

**Details**

 - Platform: Android and Windows Edge. Presumably iOS also.
 - Platform version: 13 / 126
 - App version: 1.86

**Additional context**

This behavior appears to be the output of the BIDI algorithm, see [this post](https://bsky.app/profile/jgarplind.bsky.social/post/3kuyrou25bc23) for a link to the spec and a playground.

I consider this a minor inconvenience considering the overall state of Bluesky RTL support, but have a fix in mind that I will contribute soon.

mary-ext:
Here's how Signal handles Unicode bidi, seems worth taking a look
https://github.com/signalapp/Signal-Desktop/blob/ce0fb220411b97722e1e080c14faa65d23165784/ts/util/unicodeBidi.ts#L124

haileyok:
Merged this in, thank you!

PR #4531, opened 2024-06-16
Fix misplaced '@' in RTL post meta.

jgarplind:
Addresses https://github.com/bluesky-social/social-app/issues/4529

The '@' symbol was placed at the right-side of the user handle, but the correct placement is always at the left-side since the user handle is inherently LTR.

Uses techniques from the specification: https://www.unicode.org/reports/tr9/

## Detailed problem description

`PostMeta.tsx` combines two separate text-nodes (handle + display name) into one wrapping text node. The BIDI algorithm (see link above) interprets the combined text as mixed LTR and RTL and applies logic as per the specification. As part of this logic, the `@` sign is placed on the right side (i.e. "before" in RTL terms) of the user's handle, since `@` is a "neutral character", and the combined text is apparently classified as RTL due to the RTL nature of the initial characters of the text (the display name).

By telling the algorithm to treat the handle (including the `@` character) as LTR text, this problem is mitigated.

## Show and tell

Before:
image

After:
image

🗒️ Note: Initially, the "After" was still justified towards the profile picture. Upon forking this repo and setting things up again, it was suddenly justified towards the right despite identical code. Now, after rebasing on top of @haileyok's changes, it looks fine again. Hence, I suspect my local environment rather than the code change to have temporarily caused the post meta to be justified to the right.

## Testing
✅ Tested on Windows Edge (as shown above).

❌ Don't have the necessary equipment at hand to test on iOS.

❌ Ran into setup problems when trying Android emulator, and didn't spend time trying to set up a Development build.

in other words, it would be helpful to have someone with a fully functioning setup give throw a glance.


haileyok (approved):
Nice! Thanks for digging at that. This knocks out one of the more annoying problems with LTR, appreciate you looking into it.

PR #4542, opened 2024-06-17
Bsky short link service

devinivy:
Included here is a postgres-backed node service for generating short links.

### `POST /link`
Generates a short link for a given path.  A given path always produces the same short link, up to some path normalization.  Currently only paths to starter packs which contain the actor's DID (rather than handle) are supported, e.g. `/start/did:plc:example/3kv5iwj3csz2p`.
**Request**
```json
{ "path": "/start/did:plc:example/3kv5iwj3csz2p" }
```
**Response**
```json
{ "url": "https://link.bsky.app/6H8uqD" }
```
### `GET /:linkId`
This endpoint represents the short link itself.  It generates a 301 permanent redirect to the original link.  Query parameters are preserved.  If the link id is not familiar, a temporary 303 redirect to the app's homepage is served.


haileyok (approved):

bskylink/src/routes/create.ts
```diff
@@ -0,0 +1,111 @@
...
+          error: 'InvalidPath',
+          message:
+            '"path" parameter must be formatted as a path, starting with a "/"',
+        })
+      }
+      const parts = getPathParts(path)
+      if (parts.length === 3 && parts[0] === 'start') {
+        // link pattern: /start/{did}/{rkey}
+        if (!parts[1].startsWith('did:')) {
+          // enforce strong links
```
haileyok, line 36: makes sense, was wondering if we needed to enforce this but checks out.

bskyweb/cmd/bskyweb/server.go
```diff
@@ -292,6 +302,30 @@ func (srv *Server) Download(c echo.Context) error {
...
+			Balancer: middleware.NewRoundRobinBalancer(
+				[]*middleware.ProxyTarget{{URL: url}},
+			),
+			Skipper: func(c echo.Context) bool {
+				req := c.Request()
+				if req.Method == "GET" &&
+					strings.LastIndex(strings.TrimRight(req.URL.Path, "/"), "/") == 0 && // top-level path
+					!strings.HasPrefix(req.URL.Path, "/_") { // e.g. /_health endpoint
+					return false
+				}
```
haileyok, line 318: just to verify, we're okay with these not having some identifying prefix right? ik that came up in discussions though we definitely didn't want it to be a path like `/i/abcdef`

bskylink/src/routes/create.ts
devinivy: We could aim to eventually loosen this by resolving handles in this service.  But I do think we want the link service to track strong links.

pfrazee (commented):
Just gave it a glance without much opinion, seems good

PR #4544, opened 2024-06-17
Fix Android startup perf regression

gaearon:
Patch `supportedLocales` from  `@formatjs/intl-pluralrules` to only include the ones we actually use. Since this is easy to mess up, I also added a lint rule that enforces we don't add new locales without also tweaking that patch.

I'm not sure why the `@formatjs/intl-pluralrules` polyfill is so slow but I didn't have time to get into it. I filed https://github.com/formatjs/formatjs/issues/4463.

## Test Plan

Disable the `console`-removing Babel plugin in the Babel config.

Before (5 runs):

```
Time to first paint: 2675 ms
Time to first paint: 2926 ms
Time to first paint: 2762 ms
Time to first paint: 2928 ms
Time to first paint: 2853 ms
```

After (5 runs):

```
Time to first paint: 1905 ms
Time to first paint: 1921 ms
Time to first paint: 1926 ms
Time to first paint: 1929 ms
Time to first paint: 1901 ms
```

I haven't verified what this part of the `Intl` polyfill _actually_ does but it seems safe to assume that it's OK to exclude the locales we don't use. 

Switching languages seems to still work.

### DIFF
--- a/eslint/keep-i18n-patch-in-sync.js
+++ b/eslint/keep-i18n-patch-in-sync.js
@@ -1,0 +1,28 @@
+/* eslint-disable bsky-internal/keep-i18n-patch-in-sync */
+const LOCALE_DATA_FOLDER = '@formatjs/intl-pluralrules/locale-data/'
+const GEN_MODULE_PATH =
+  '@formatjs/intl-pluralrules/supported-locales.generated.js'
+
+exports.create = function create(context) {
+  delete require.cache[require.resolve(GEN_MODULE_PATH)]
+  const {supportedLocales} = require(GEN_MODULE_PATH)
+  return {
+    Literal(node) {
+      if (typeof node.value !== 'string') {
+        return
+      }
+      if (!node.value.startsWith(LOCALE_DATA_FOLDER)) {
+        return
+      }
+      const code = node.value.slice(LOCALE_DATA_FOLDER.length)
+      if (!supportedLocales.includes(code)) {
+        context.report({
+          node,
+          message:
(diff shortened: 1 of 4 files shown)


eslint/keep-i18n-patch-in-sync.js
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    Literal(node) {
+      if (typeof node.value !== 'string') {
+        return
+      }
+      if (!node.value.startsWith(LOCALE_DATA_FOLDER)) {
+        return
+      }
+      const code = node.value.slice(LOCALE_DATA_FOLDER.length)
+      if (!supportedLocales.includes(code)) {
```
gaearon, line 18: This will catch adding new lines like this:

https://github.com/bluesky-social/social-app/blob/077da0830924d4c040ef095d8349f13fdfdf4372/src/locale/i18n.ts#L81

without also amending the patch below.

pfrazee (approved):
Ah, interesting. Good to know

haileyok:
Three runs on a physical device w/ patch:

```
Time to first paint: 1235 ms
Time to first paint: 1166 ms
Time to first paint: 1207 ms
```

Three without the patch:

```
Time to first paint: 1888 ms
Time to first paint: 1893 ms
Time to first paint: 1924 ms
```

Also tested switching between languages, doesn't seem to affect anything there. Seems solid, great investigation! 🙌

haileyok (approved):

estrattonbailey (commented):
Just here to say: very nice 🔥

PR #4545, opened 2024-06-18
Implement thread locking

pfrazee:
- Tuned the `WhoCanReply` component visually (removed the icon, modified colors and spacing)
- Every thread created by the current user now shows the current threadgate settings in the PostThread screen
- When the user is the thread OP, they now have an "Edit" button
- The Threadgate modal now supports editing an existing post

## iOS

|Dark|Dark|Light|Light|
|-|-|-|-|
|image|image|image|image|

## Web

|Dark|Dark|Light|Light|
|-|-|-|-|
|CleanShot 2024-06-17 at 20 15 24@2x|CleanShot 2024-06-17 at 20 15 43@2x|CleanShot 2024-06-17 at 20 16 01@2x|CleanShot 2024-06-17 at 20 16 14@2x|


src/lib/analytics/types.ts
```diff
@@ -32,6 +32,8 @@ export type TrackPropertiesMap = {
   'Post:ThreadMute': {} // CAN BE SERVER
   'Post:ThreadUnmute': {} // CAN BE SERVER
   'Post:Reply': {} // CAN BE SERVER
+  'Post:EditThreadgateOpened': {}
```
gaearon, line 35: these are old mixpanel events, intentional? time to delete these maybe.

pfrazee: Oh -- I'm unsure, do we not still use `track()` ? That API depends on the declarations in `TrackPropertiesMap`

gaearon: yeah `track` is the old one, `logEvent` is the new one. we can probably just rename/remove at this point

PR #4547, opened 2024-06-18
Bsky link card service

devinivy:
This work implements a stateless node service for generating png link card images.  It currently supports generating link cards for starter packs at `GET /start/{actor}/{rkey}`.  Since it's compute-intensive to generate pngs, the service can be run in cluster mode by setting `CARD_CLUSTER_WORKER_COUNT`.


bskycard/src/routes/starter-pack.tsx
```diff
@@ -0,0 +1,98 @@
...
+        return res.status(404).end('not found')
+      }
+      const imageEntries = await Promise.all(
+        starterPack.listItemsSample
+          .map(li => li.subject)
+          .concat(starterPack.creator)
+          // has avatar
+          .filter(p => p.avatar)
+          // no bad labels
+          .filter(p => !p.labels.some(l => hideAvatarLabels.has(l.val)))
```
estrattonbailey, line 45: What does this look like if we don't have a full set of images 🤔

haileyok: We had some mockups of this are varying stages, it looked pretty good regardless of the count. Ideally we won't ever have less than 8 but...just in case it doesn't look horrible at all.

haileyok: (Assuming we have the same implementation here)

devinivy: Yeah, I tested it with all counts down to 1.  I think it might be nice to have a fallback design for 1-4, where the background tiling also looks a bit repetitive.

haileyok (approved):
express and react parts look good, assume docker file and workflow are also good

PR #4548, opened 2024-06-18
UI tweaks and nits for starter packs

haileyok:


__tests__/lib/string.test.ts
```diff
@@ -852,11 +856,15 @@ describe('parseStarterPackHttpUri', () => {
...
+    {name: 'did:123', rkey: 'rkey'},
+    null,
+    null,
+    null,
   ]
 
   it('returns the correct name and rkey when input is valid', () => {
     for (let i = 0; i < inputs.length; i++) {
-      const result = parseStarterPackHttpUri(inputs[i])
+      const result = parseStarterPackUri(inputs[i])
```
estrattonbailey, line 867: Kinda nit kinda not: should really use `it.each` and a nice interpolated string for this, or refactor into separate test cases like I did for `ago` tests. Reason being, it's hard to know which part failed in tests like this, why tests exist, if duplicate tests exist, etc, esp when the arrays of tests grow.

haileyok: Yea, I've been wanting to refactor a couple of the other tests in this thing for the same reason (that embed test is brutal). Might as clean this one up now.

__tests__/lib/string.test.ts
```diff
@@ -856,13 +856,13 @@ describe('createStarterPackLinkFromAndroidReferrer', () => {
...
       createStarterPackLinkFromAndroidReferrer(
-        'utm_content=starterpack-haileyok.com-rkey-more',
+        'utm_content=starterpack_haileyok.com_rkey_more',
       ),
     ).toEqual(null)
 
     expect(
       createStarterPackLinkFromAndroidReferrer(
-        'utm_content=notastarterpack-haileyok.com-rkey',
+        'utm_content=notastarterpack_haileyok.com_rkey',
```
estrattonbailey, line 865: Should add a test with a hyphen in the handle

src/lib/generate-starterpack.ts
```diff
@@ -111,7 +109,20 @@ export async function generateStarterpack({
       )
     ).uri
   } catch (e: unknown) {
-    logger.error('Failed to generate starter pack', {error: e})
+    logger.error('Failed to generate starter pack', {safeMessage: `${e}`})
```
estrattonbailey, line 112: Doesn't matter but FYI you actually don't need to coerce here, we check for `instanceof Error` in the breadcrumbs handling and call `Error.toString` https://github.com/bluesky-social/social-app/blob/main/src/logger/index.ts#L97

__tests__/lib/string.test.ts
estrattonbailey: Ope nvm I see you switched to underscores

src/lib/generate-starterpack.ts
haileyok: ah wasn't sure if we did it for each value or not, thanks.

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -72,8 +73,13 @@ export function StepFinished() {
           })
           listItems = listRes.data.items
         }
+      } catch (e) {
+        logger.error('Failed to fetch starter pack', {safeMessage: `${e}`})
+        // don't tell the user, just get them through onboarding.
```
estrattonbailey, line 78: Niiice

estrattonbailey (approved):

PR #4551, opened 2024-06-18
Dedupe Zod installation

gaearon:
Curious if it affects bundle size.

### DIFF
--- a/package.json
+++ b/package.json
@@ -271,6 +271,7 @@
   "resolutions": {
     "@types/react": "^18",
     "**/zeed-dom": "0.10.9",
+    "**/zod": "3.23.8",
     "**/expo-constants": "16.0.1",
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Nice, nothing iffy in the changelog either. I see in >3.3 [they say they made](https://zod.dev/CHANGELOG?id=_33) "HUGE speed improvements" 🤔 could maybe even test an upgrade if this one goes smoothly

gaearon:
Oh I think we're already on 3.22 — not 3.2, they just stopped publishing changelog on the docs site and moved next ones to GH

estrattonbailey:
Ah geez, brain fart 🙃 well I don't see any breaking changes that affect us in the actual changelog either. Nice fix!

PR #4553, opened 2024-06-18
Remove lodash dependency

mozzius:
We've been trying to use the version of lodash that's been split into individual modules, to reduce bundle size. While messing around with Expo Atlas, it turns out that was all for naught because `rn-fetch-blob` has it as a dependency. We were also accidentally importing `clamp` from the transient version of lodash, which I have fixed as well.

This adds Expo Atlas as a dependency - to use, follow the instructions here:
https://expo.dev/blog/introducing-expo-atlas#built-in-with-expo-sdk-51


mozzius:
well that didn't seem to work 🤨

PR #4332, opened 2024-06-03
Starter Packs

haileyok:
### Other PRs

This PR relies on two other PRs (which have been merged into this PR). Don't merge this in until we look at both of those and/or revert them being merged into this PR.

https://github.com/bluesky-social/social-app/pull/4403
https://github.com/bluesky-social/social-app/pull/4533

### Wizard

- [x] Details
- [x] Profiles
  - [x] Search
  - [x] List of those that have been added
  - [x] Use `typeahead` query over follows **Pending**
- [x] Feeds
- [x] Edit existing starter packs

### Logged In

- [x] Starter pack list on profiles
- [x] QR code generation for sharing starter packs
  - [x] Tweak style of this thing
  - [x] Dark mode
- [x] Starter pack screen showing name, description, profiles, and feeds
- [x] Sharing by a short link
- [x] Newskie badge on profiles
  - [x] Dialog that shows the starter pack used if applicable

### Logged Out

- [x] Landing
  - [x] iOS (App Clip) **Pending**
  - [x] Android (Android Referrer API)
  - [x] Web (Basic)
- [x] Follow suggested profiles on account creation
- [x] Pin suggested feeds
- [x] Set created at date
- [x] Land on the first-most feed if there were feeds in the pack
- [x] Land on the following feed if there were only profiles and there were 25 or more profiles in the pack

### Stats

- [x] Created an account w/ a starter pack
  - [x] Which one?
- [x] Followed accounts or pinned feeds from the starter pack page

### What the fucks

- [x] Make sure nothing weird happens if the starter pack still exists but the list is deleted
- [x] Make sure we only fetch 50 profiles even if the list has some other number (and restrict the follows to that limit)

PR #4556, opened 2024-06-18
Add `useGetTimeAgo` and utils

estrattonbailey:
This PR incorporates an excellent community PR to make testing the `ago` util deterministic. I don't think I've broken anything done in that PR.

This PR [adds in work](https://github.com/bluesky-social/social-app/blob/b5b8bf5b1e000431b7ae8db8d47b6b711d90bf50/src/lib/strings/time.ts) from #4332 (long formatting), and moves this functionality to a hook so that everything can be properly translated.

I also broke out all the tests so we know what we're testing.

**Testing**
Pull up your following using this branch and prod, check if timestamps match. Other special use cases are DMs timestamps and logs.


src/lib/strings/time.ts
```diff
@@ -1,45 +1,3 @@
...
-    return `${Math.floor(diffSeconds / HOUR)}h`
-  } else if (diffSeconds < MONTH_30) {
-    return `${Math.round(diffSeconds / DAY)}d`
-  } else {
-    let months = diffSeconds / MONTH
-    if (months % 1 >= 0.9) {
-      months = Math.ceil(months)
-    } else {
-      months = Math.floor(months)
-    }
```
estrattonbailey, line 33: Technically dropping this, but the output is essentially the same: values very close to 1year are rounded up and result in the full timestamp being shown.

haileyok (approved):

PR #4558, opened 2024-06-18
fix keyboard overlaying onboarding inputs

haileyok:
## Why

Small PR to address the keyboard being in the way of some inputs during onboarding on some devices/font sizes. Also makes sure that the keyboard gets dismissed when opening the DOB modal.

## Test Plan

Make sure the screen is scrollable, and that you can scroll all the way to the bottom with keyboard shown.

https://github.com/bluesky-social/social-app/assets/153161762/38017bd4-c1fe-4890-b961-079e2d62129f

### DIFF
--- a/src/components/forms/DateField/index.tsx
+++ b/src/components/forms/DateField/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {View} from 'react-native'
+import {Keyboard, View} from 'react-native'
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Tested on iOS, can't break it. Asked Hailey to test on Android and she says it works well there too 👍

PR #4560, opened 2024-06-18
Force callers of `getTimeAgo` to pass in the value for "now"

estrattonbailey:
Removed unused and problematic hook, and update the API of `getTimeAgo` to match `dateDiff` e.g. `earlier, later, options`.


gaearon (approved):
haven't tested but code lgtm!

src/components/NewskieDialog.tsx
```diff
@@ -30,12 +30,13 @@ export function NewskieDialog({
     const moderation = moderateProfile(profile, moderationOpts)
     return sanitizeDisplayName(name, moderation.ui('displayName'))
   }, [moderationOpts, profile])
+  const [now] = React.useState(Date.now())
```
gaearon, line 33: nit: `(() => Date.now())` to avoid unnecessary calls

gaearon: meh it's fine, this isn't commonly used

PR #4563, opened 2024-06-18
Better handling of blocks in `KnownFollowers`

estrattonbailey:
So the issue here is that there are two count values that matter, `length` or `count`:
- `knownFollowers.followers.length`
- `knownFollowers.count`

`followers[]` filters out blocks, and only hydrates 5 profiles server-side, so that array _could_ be `0`, though it's unlikely. If that ever happened, we'd just not render anything at all.

More likely is `followers.length < count` because 1 or more profiles are blocked and filtered out.

So rendering should be driven by `followers.length`, but `and N others` can only be driven by `count` since that can be much higher than the 5 max returned on `followers`.

So the fix here is:
- determine if we have 1, or more than 1 `followers[]` to render
- use the `count` to determine if we need a `, ` or ` and `
- use `count` to determine if we need `and N others`

Here's a test case illustrating that: we have 3 follows in common, but I block one of them. Note only 2 avis (one is blocked), but it still says `and 1 other`. Clicking through will only show 2 profiles.
CleanShot 2024-06-18 at 15 06 20@2x

### DIFF
--- a/src/components/KnownFollowers.tsx
+++ b/src/components/KnownFollowers.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import {View} from 'react-native'
 import {AppBskyActorDefs, moderateProfile, ModerationOpts} from '@atproto/api'
-import {msg, plural, Trans} from '@lingui/macro'
+import {msg, Plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
 import {makeProfileLink} from '#/lib/routes/links'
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Tested, looks good!

src/components/KnownFollowers.tsx
```diff
@@ -164,29 +165,22 @@ function KnownFollowersInner({
               },
             ]}
             numberOfLines={2}>
-            {count > 2 ? (
+            {renderableCount >= 2 ? (
               <Trans>
-                Followed by{' '}
+                {_(msg`Followed by`)}{' '}
```
gaearon, line 170: This kind of negates the point of using `Trans` above. The point was to determine the format first ("Followed by X", "Followed by X and Y" or "Followed by X, Y and Z" first), and then interpolate into one of the supported formats. Breaking `Followed by` into its own `msg` breaks that.

gaearon (changes requested):
let's fix `<Trans>`

src/components/KnownFollowers.tsx
gaearon: See https://github.com/bluesky-social/social-app/pull/4494 for context where this was introduced.

estrattonbailey: Oh ok, TIL, yeah can fix this up 👍

estrattonbailey: 2-for-1, realized when splitting out this logic that I was missing a case:
CleanShot 2024-06-18 at 15 42 46@2x

gaearon (approved):

PR #4566, opened 2024-06-18
Collection of moderation fixes

pfrazee:
- 5c2ef687567d398efbd5a4e8a6ed53fae79a8ea7 Fix to thread PostHider to ensure that blocked quote-posts do not cause the parent post to be blurred
- c0565393b91db0d79400efe4e4da5562b2812275 Change to "Show hidden replies" to subsequently show the post hiders with their reasons on each post, per some user requests
- c33b2dd2e6f3da4b862539603062da089e434e39 Fix to the fallback avatar used on labels (now shows the shield)

### DIFF
--- a/src/view/com/util/UserAvatar.tsx
+++ b/src/view/com/util/UserAvatar.tsx
@@ -35,6 +35,7 @@ export type UserAvatarType = 'user' | 'algo' | 'list' | 'labeler'
 
 interface BaseUserAvatarProps {
   type?: UserAvatarType
+  shape?: 'circle' | 'square'
   size: number
   avatar?: string | null
 }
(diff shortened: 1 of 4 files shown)


src/view/com/util/UserAvatar.tsx
```diff
@@ -159,16 +169,18 @@ export {DefaultAvatar}
...
   type = 'user',
+  shape,
   size,
   avatar,
   moderation,
   usePlainRNImage = false,
 }: UserAvatarProps): React.ReactNode => {
   const pal = usePalette('default')
   const backgroundColor = pal.colors.backgroundLight
+  shape = shape || type === 'user' ? 'circle' : 'square'
```
gaearon, line 180: I struggle to read this — is the precedence correct? I think the intent is

```js
shape = shape || (type === 'user' ? 'circle' : 'square')
```

with the part after `||` being a default value.

But the code is doing something more like `'foo' || true ? 1 : 0` (which gives `1` rather than `'foo'`)

Overall having two sources of truth for this is confusing, is the intent to allow sometimes specifying a non-type-default shape?

pfrazee: ah whoops

```
shape = shape || (type === 'user' ? 'circle' : 'square')
```

Correct that ^ is the intent. Essentially `shape` acts as an override of the shape that the different types default to. In this case, labelers are normally square, but when theyre on posts they need to be circles

gaearon (approved):

gaearon (approved):

PR #4567, opened 2024-06-19
Date and time localization

liberaldev:
Currently, Blue Sky has an issue where the date and time are output in English even though the app language is something other than English (ex. Korean), so I fixed it so that the date and time are output in the local language.


mary-ext:
Already being tackled in https://github.com/bluesky-social/social-app/pull/4273

mary-ext:
#4273 is merged now, so this PR seems obsolete

PR #4570, opened 2024-06-19
Fix avi placeholder layout

gaearon:
Seems like it grew from 80px to 90px at some point. Adjusting accordingly. Also tweaked the button placeholder padding to match the real one.

## Before

https://github.com/bluesky-social/social-app/assets/810438/e06c66f6-ba55-4514-a14a-17c67904a058

## After

https://github.com/bluesky-social/social-app/assets/810438/421e3bd6-da1c-4c94-87ad-09b70173c6f6

## Test Plan

Tested on web (above), verified it's also 1:1 on iOS and Android.


estrattonbailey (approved):
Good eye

PR #4572, opened 2024-06-19
Fix double border in PWI

mozzius:
Header size was being influenced by the presence of the feeds button. This sets a min-height


gaearon (approved):
works on my machine

PR #4574, opened 2024-06-19
Allow search params when converting bsky.app urls

mozzius:
Allows linking to specific searches, i.e. https://bsky.app/profile/jgarplind.bsky.social/post/3kuyshu7heu2a

### DIFF
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -158,11 +158,18 @@ export function isBskyDownloadUrl(url: string): boolean {
   return url === '/download' || url.startsWith('/download?')
 }
 
-export function convertBskyAppUrlIfNeeded(url: string): string {
+export function convertBskyAppUrlIfNeeded(
+  url: string,
+  includeSearch = false,
+): string {
   if (isBskyAppUrl(url)) {
     try {
       const urlp = new URL(url)
-      return urlp.pathname + urlp.search
+      if (includeSearch) {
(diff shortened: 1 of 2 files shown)


src/lib/strings/url-helpers.ts
```diff
@@ -162,7 +162,7 @@ export function convertBskyAppUrlIfNeeded(url: string): string {
   if (isBskyAppUrl(url)) {
     try {
       const urlp = new URL(url)
-      return urlp.pathname
+      return urlp.pathname + urlp.search
```
gaearon, line 165: I'm not sure this is a safe change — some code calling `convertBskyAppUrlIfNeeded` assumes there's no search. What do you think of the callsites? Could make a separate function for the case of handling in-app links. Or could return a richer data structure and let the caller decide which parts to use.

gaearon:
Doesn't seem to fix it for native btw

gaearon (changes requested):
Looks like this would fix it for web but it still doesn't work on native. I think for native we might want to have a pushable version of the Search. This might be useful in other places too, like for searches themselves. Let's maybe leave this for now.

mozzius:
I wish we were using Expo Router, this would *just work*

haileyok:
Don't we just need to add `Search` into `commonScreens` instead of only being inside the `Search` tab? @mozzius

pfrazee:
I'll take a stab at it

PR #4577, opened 2024-06-20
Fix threadgate read after write

gaearon:
- Fix read-after-write problem leading to threadgate update not propagating to the UI.
- Drive-by fix for React keys (feel free to ignore if you rewrote that in your branch).

## Test Plan

Verified with console logs the logic works as intended.

https://github.com/bluesky-social/social-app/assets/810438/6eb08c27-f2b9-4488-87ff-fa7dc728296b


pfrazee (approved):
LGTM, tested and confirmed works

PR #4582, opened 2024-06-20
Share dialog for starter packs

haileyok:


src/lib/strings/starter-pack.ts
```diff
@@ -77,3 +77,15 @@ export function httpStarterPackUriToAtUri(httpUri?: string): string | null {
 
   return `at://${parsed.name}/app.bsky.graph.starterpack/${parsed.rkey}`
 }
+
+export function getStarterPackOgCard(
+  didOrStarterPack: AppBskyGraphDefs.StarterPackView | string,
+  rkey?: string,
+) {
+  if (typeof didOrStarterPack === 'string') {
+    return `https://ogcard.bsky.app/start/${didOrStarterPack}/${rkey}`
```
estrattonbailey, line 86: These two URLs are different, that ok?

haileyok: oops yep, nice catch

estrattonbailey (approved):

PR #4584, opened 2024-06-20
Wait for AppView when posting

gaearon:
Similar to https://github.com/bluesky-social/social-app/pull/4577 but for posting.

## Before

https://github.com/bluesky-social/social-app/assets/810438/eb2f4403-8529-4a27-997d-7082ef41ccef

## After

https://github.com/bluesky-social/social-app/assets/810438/b86bdd31-f0c5-4c0b-8785-d9a0c70a624a

## Test Plan

See above videos. Notice that:

- Posts reliably show up now (in the "before" video, the first reply seemed to "disappear" until refresh/invalidation)
- When posts show up, parent posts reliably increment +1 reply count (in the "before" video, sometimes parents would show no replies)

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -24,12 +24,18 @@ import Animated, {
 } from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {LinearGradient} from 'expo-linear-gradient'
+import {
+  AppBskyFeedDefs,
+  AppBskyFeedGetPostThread,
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Code looks good. Tested on Web and with throttling.

The one tweak we could make to improve this is -- move the `whenAppViewReady` outside of the post flow and into the after-publish thread invalidation. That would let the user leave the composer while the publish finishes. Not important to do now but something to think about.

gaearon:
Sent https://github.com/bluesky-social/social-app/pull/4587 but not sure about it

PR #4586, opened 2024-06-20
Add a11y context

estrattonbailey:
Adds a couple a11y props to context so that they're available synchronously in the app. Right now we only check for reduced motion, so I've used that same value from Reanimated as the initial value here (that check IS sync, actually).

This also adds screen reader state for use in Starter Packs and eventually other infinite scroll cases.

Note case on web: screen reader always returns `true`. Since we're primarily concerned about mobile for this immediate need, I've opted to override on web to always be `false`. Eventually we may want to fork these props into something like `isScreenReaderEnabledWeb` and handle some cases differently.

### DIFF
--- a/src/state/a11y.tsx
+++ b/src/state/a11y.tsx
@@ -1,0 +1,65 @@
+import React from 'react'
+import {AccessibilityInfo} from 'react-native'
+import {isReducedMotion} from 'react-native-reanimated'
+
+import {isWeb} from '#/platform/detection'
+
+const Context = React.createContext({
+  reduceMotionEnabled: false,
+  screenReaderEnabled: false,
+})
+
+export function useA11y() {
+  return React.useContext(Context)
+}
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const [reduceMotionEnabled, setReduceMotionEnabled] = React.useState(
+    isReducedMotion(),
+  )
+  const [screenReaderEnabled, setScreenReaderEnabled] = React.useState(false)
+
+  React.useEffect(() => {
+    const reduceMotionChangedSubscription = AccessibilityInfo.addEventListener(
+      'reduceMotionChanged',
+      enabled => {
+        setReduceMotionEnabled(enabled)
+      },
+    )
+    const screenReaderChangedSubscription = AccessibilityInfo.addEventListener(
+      'screenReaderChanged',
+      enabled => {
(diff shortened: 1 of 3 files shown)


src/state/a11y.tsx
```diff
@@ -0,0 +1,65 @@
...
+  screenReaderEnabled: false,
+})
+
+export function useA11y() {
+  return React.useContext(Context)
+}
+
+export function Provider({children}: React.PropsWithChildren<{}>) {
+  const [reduceMotionEnabled, setReduceMotionEnabled] = React.useState(
+    isReducedMotion(),
```
haileyok, line 18: nit: should use `() => isReducedMotion()`

src/state/a11y.tsx
```diff
@@ -0,0 +1,65 @@
...
+      enabled => {
+        setScreenReaderEnabled(enabled)
+      },
+    )
+
+    ;(async () => {
+      const [_reduceMotionEnabled, _screenReaderEnabled] = await Promise.all([
+        AccessibilityInfo.isReduceMotionEnabled(),
+        AccessibilityInfo.isScreenReaderEnabled(),
+      ])
```
haileyok, line 40: im planning on adding some methods to get these synchronously in the future to the SwissArmyKnife (and dropping the patch we hacked into `reanimated`), but this seems fine for now.

src/state/a11y.tsx
```diff
@@ -0,0 +1,65 @@
...
+      reduceMotionChangedSubscription.remove()
+      screenReaderChangedSubscription.remove()
+    }
+  }, [])
+
+  const ctx = React.useMemo(() => {
+    return {
+      reduceMotionEnabled,
+      /**
+       * Always returns true on web. For now, we're using this for mobile a11y,
```
haileyok, line 55: really weird there isn't an api for this...maybe a fingerprinting concern ig

haileyok (approved):
one nit but lgtm

src/state/a11y.tsx
estrattonbailey: Always forget, thanks

pfrazee (approved):
LGTM

src/state/a11y.tsx
estrattonbailey: Oh awesome. Pry good to have on context anyway for testing, but would be nice to get these sync on first load.

PR #4589, opened 2024-06-20
Gate starter packs

haileyok:


src/view/screens/Profile.tsx
```diff
@@ -167,6 +170,9 @@ function ProfileScreenLoaded({
   const [currentPage, setCurrentPage] = React.useState(0)
   const {_} = useLingui()
   const setDrawerSwipeDisabled = useSetDrawerSwipeDisabled()
+  const gate = useGate()
+  const starterPacksEnabled =
+    IS_DEV || IS_TESTFLIGHT || (!isWeb && gate('starter_packs_enabled'))
```
haileyok, line 175: We will have to deploy `/start` to production for review. So gating based on `isWeb` as well as statsig here and we can remove that in a separate commit before releasing.

Native will remain ungated during review. Probably won't end up needing the statsig gate but just in case.

estrattonbailey: So in prod this is always false on web, right, and it's false on native too unless the gate is enabled. So are we going to ramp to 100% for review?

estrattonbailey (approved):

src/view/screens/Profile.tsx
haileyok: yup discussed on slack. will have this at 100% for review.

PR #4590, opened 2024-06-20
Don't render starter packs that are invalid, give option to delete for owner

haileyok:
As the title says.

### DIFF
--- a/src/screens/StarterPack/StarterPackScreen.tsx
+++ b/src/screens/StarterPack/StarterPackScreen.tsx
@@ -83,17 +83,11 @@ export function StarterPackScreen({route}: StarterPackScreeProps) {
 
   const isValid =
     starterPack &&
+    (starterPack.list || starterPack?.creator?.did === currentAccount?.did) &&
     AppBskyGraphDefs.validateStarterPackView(starterPack) &&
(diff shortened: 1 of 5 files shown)


src/screens/StarterPack/StarterPackScreen.tsx
```diff
@@ -73,26 +79,55 @@ export function StarterPackScreen({route}: StarterPackScreeProps) {
...
       />
     )
   }
 
+  if (
+    !listMembersQuery.data &&
+    starterPack.creator.did === currentAccount?.did
+  ) {
+    return <InvalidStarterPack rkey={rkey} />
+  }
```
estrattonbailey, line 124: Will this ever flash as the list loads? Basically, should we use the error state instead of data existence to determine if we should show this

estrattonbailey (approved):

src/screens/StarterPack/StarterPackScreen.tsx
haileyok: i actually tried that the first time, but turns out “list” isn’t even set on the starterPackView after it gets deleted. 

that said though let me switch this to “!starterPack?.list?.uri” so it’s more clear.

re: a flash, don’t see one on my end, though switching it to the above should be a better guarantee of that.

PR #4595, opened 2024-06-21
Fetch more than 3 suggested follows after first load

estrattonbailey:
New Explore page only loads 3 suggested follows up front in order to show feeds above the fold. But when loading more, we should load more than 3 at a time.

This PR loads 10 users _after_ the first page has been returned. Additionally, since feeds are already below the fold, we might as well fetch more than 3. So this PR updates to fetch 10 of those and let them extend below the fold.

### DIFF
--- a/src/state/queries/suggested-follows.ts
+++ b/src/state/queries/suggested-follows.ts
@@ -34,13 +34,14 @@ const suggestedFollowsByActorQueryKey = (did: string) => [
   did,
 ]
 
-type SuggestedFollowsOptions = {limit?: number}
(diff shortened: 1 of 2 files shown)


gaearon:
Why not always load 10? But then slice them in the UI according to which pages you have. This would also let us show the correct avatars (by peeking ahead).

pfrazee (approved):
Oh nice

pfrazee:
Ah, or dan's suggestion

gaearon (approved):
yesss

PR #4596, opened 2024-06-21
Stop feed pagination after we get dupes back

estrattonbailey:
Dataplane response is a static list atm, and is returning a few duplicates. Easiest protection is simply to filter those out until backend fix gets in.


estrattonbailey:
Backend rebuilt table to removed dupes, shouldn't be an issue again

PR #4600, opened 2024-06-22
Tweak feed card to prevent spinnerz when pushing to screen

haileyok:
## Why

There's some spinner states on feed cards - in prod, not related to the `FeedCard` change necessarily - that we'd like to remove. Easy little win.

We also didn't actually migrate to using the new `FeedCard` in the `ProfileFeedgens` list. I had done this work in the starterpacks branch, so I picked it out of there and made a couple tweaks after the recent `FeedCard` merge into main.

Finally, made a little types tweak inside of `FeedCard` so that `Link` knows whether something is a `ListView` or `GeneratorView`, supporting this new precaching.

## How

We already have a `precacheFeed` function that takes in `FeedSourceInfo` from inside one of our query functions. However, we also want to make it possible to run this with a `GeneratorView`.

Following a similar pattern that we use for posts and profiles, we're adding a `precacheFeedFromGeneratorView` that can be called with the `Link`'s `onPress`. This runs just before pushing to the screen - and lets the query already have the info it needs.

## Test Plan

Visit a profile that has some feeds, and notice the lack of a spinner state (easiest to see with throttling)

### Before

https://github.com/bluesky-social/social-app/assets/153161762/d1a9c122-e9d7-430e-9608-91babd558233 

### After

https://github.com/bluesky-social/social-app/assets/153161762/47194028-2246-423a-8e6a-4250e42dd9ea

### DIFF
--- a/src/components/FeedCard.tsx
+++ b/src/components/FeedCard.tsx
@@ -6,7 +6,6 @@ import {
   AppBskyGraphDefs,
   AtUri,
 } from '@atproto/api'
-import {ListView} from '@atproto/api/dist/client/types/app/bsky/graph/defs'
 import {msg, plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useQueryClient} from '@tanstack/react-query'
@@ -47,18 +46,19 @@ type Props =
 export function Default(props: Props) {
   const {type, view} = props
   const displayName = type === 'feed' ? view.displayName : view.name
+  const purpose = type === 'list' ? view.purpose : undefined
   return (
     <Link label={displayName} {...props}>
       <Outer>
         <Header>
           <Avatar src={view.avatar} />
-          <TitleAndByline title={displayName} creator={view.creator} />
-          <Action
-            uri={view.uri}
-            pin
+          <TitleAndByline
+            title={displayName}
+            creator={view.creator}
             type={type}
-            purpose={type === 'list' ? view.purpose : undefined}
+            purpose={purpose}
           />
+          <Action uri={view.uri} pin type={type} purpose={purpose} />
         </Header>
         <Description description={view.description} />
         {type === 'feed' && <Likes count={view.likeCount || 0} />}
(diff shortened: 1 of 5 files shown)


src/components/FeedCard.tsx
```diff
@@ -277,6 +295,7 @@ export function createProfileFeedHref({
 }: {
   feed: AppBskyFeedDefs.GeneratorView | AppBskyGraphDefs.ListView
 }) {
+  console.log(feed.creator)
```
gaearon, line 298: oops

src/view/screens/Feeds.tsx
```diff
@@ -644,7 +644,7 @@ function SavedFeed({
     savedFeed.type === 'feed' ? savedFeed.view.displayName : savedFeed.view.name
 
   return (
-    <FeedCard.Link testID={`saved-feed-${feed.displayName}`} feed={feed}>
+    <FeedCard.Link testID={`saved-feed-${feed.displayName}`} {...savedFeed}>
```
gaearon, line 647: Nit: Spread props become regrettable over time. Let's do it explicitly unless it's like 10 fields.

src/components/FeedCard.tsx
```diff
@@ -31,21 +33,21 @@ import * as Prompt from '#/components/Prompt'
...
       view: AppBskyGraphDefs.ListView
-    }) {
+    }
+
+export function Default(props: Props) {
+  const {type, view} = props
   const displayName = type === 'feed' ? view.displayName : view.name
   return (
-    <Link feed={view}>
+    <Link label={displayName} {...props}>
```
gaearon, line 50: Same, let's not spread here

src/view/screens/Feeds.tsx
haileyok: Oh yea, didn't like this either. Can you explain to me why the type is wonky though if I explicitly set `type` and `view` props

src/state/queries/feed.ts
```diff
@@ -588,3 +588,12 @@ function precacheList(
     list,
   )
 }
+
+export function precacheFeedFromGeneratorView(
+  queryClient: QueryClient,
+  view: AppBskyFeedDefs.GeneratorView,
+) {
+  const hydratedFeed = hydrateFeedGenerator(view)
+  console.log(hydratedFeed)
```
gaearon, line 597: oops

src/view/screens/Feeds.tsx
gaearon: Ok let's not screw around with types for now

gaearon (approved):
lgtm

PR #4602, opened 2024-06-22
Add social proof to suggested follows

mozzius:
Adds social proof to the profile cards on the Explore page. This is placed behind a gate so we can measure impact.

image


estrattonbailey (approved):
Nice one! LGTM

estrattonbailey:
I think we can hold off until next release since we're so far along on this one. I may touch this more for feed interstitials too, so we can continue to integrate there this week.

pfrazee (approved):

PR #4604, opened 2024-06-22
Couple of starter packs tweaks

haileyok:
**[without whitespace](https://github.com/bluesky-social/social-app/compare/hailey/sp-tweaks?w=1)**

- Back gesture in wizard
- Tweak copy on headers
- Tweak max length of display names in wizard (truncation)
- Trim display name and description of starter pack
- Use a "Remove" button in edit dialog rather than checkbox


src/components/StarterPack/Wizard/WizardListCard.tsx
```diff
@@ -85,17 +88,33 @@ function WizardListCard({
           {subtitle}
         </Text>
       </View>
-      <Checkbox />
+      {btnType === 'checkbox' ? (
+        <Checkbox />
+      ) : !disabled ? (
+        <Button
```
estrattonbailey, line 94: Hmmmm does work and appears to retain keyboard nav a11y so I think this is ok

estrattonbailey (approved):
LGTM

PR #4615, opened 2024-06-24
Remove the 'Who can reply' element except when viewing root, and add "edit"

pfrazee:
This PR:

- Adds an edit-pencil icon to WhoCanReply if you are the OP
- Removes the WhoCanReply element from the reply parentage when not viewing the OP

|Screenshot|Description|
|-|-|
|CleanShot 2024-06-24 at 08 32 55@2x|After spending the weekend QAing, I found this element somewhat jarring to see everywhere when viewing replies. Visually it scans like a moderated post (eg "Post by muted user") and it breaks up the conversation. It also just takes up more real-estate that the information deserves.|

|Screenshot|Description|
|-|-|
|CleanShot 2024-06-24 at 08 39 40@2x|When viewing the OP, the WhoCanReply remains, but I also felt it wasn't super obvious that thread authors could edit the threadgate. So I added a blue pencil icon.|


gaearon (approved):
lgtm

PR #4618, opened 2024-06-24
Wait for preferences before showing suggested feeds

haileyok:
## Why

The `useSavedFeeds` query needs the user's preferences to be loaded first. Generally a user's preferences should be loaded already and cached before we enter the wizard, but its possible for them to be invalidated when we enter. If they are invalidated, they will get refetched inside of `useSavedFeeds`.

While fetching preferences, the `useQuery` in `useSavedFeeds` is disabled. This means that `isLoading` will be _false_ for a moment - so just checking `isLoadingSavedFeeds` will not always work. Instead, we want to look at `isFetchedAfterMount`.

## Test Plan

You can add a test promise as shown in the video below and remove `staleTime: Infinity`.

### Before

https://github.com/bluesky-social/social-app/assets/153161762/b0ae5fd7-7dae-4aff-ac85-250fa8a9a246

### After

https://github.com/bluesky-social/social-app/assets/153161762/1b59e8a3-af7a-4395-8bcc-4777a9bfe5d2

### DIFF
--- a/src/screens/StarterPack/Wizard/StepFeeds.tsx
+++ b/src/screens/StarterPack/Wizard/StepFeeds.tsx
@@ -46,11 +46,16 @@ export function StepFeeds({moderationOpts}: {moderationOpts: ModerationOpts}) {
     limit: 30,
   })
   const popularFeeds = popularFeedsPages?.pages.flatMap(p => p.feeds) ?? []
+
+  // If we have saved feeds already loaded, display them immediately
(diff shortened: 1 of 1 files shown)


gaearon (approved):
since saved feeds are always earlier in concatenation, is there any downside to making just the concat conditional?

haileyok:
Yea that seems fine. Probably a good likelihood these will already be loaded so seems nice to have something show up immediately.

PR #4620, opened 2024-06-24
Disable facets in `FeedCard.Description` component

estrattonbailey:
This PR disables links within `FeedCard.Description`. It _does_ run `detectFacetsWithoutResolution` so that URLs look nice.

If we ever have need for them, we can add another export here called like `DescriptionWithFacets`, sorta like the first commit in this PR.

### DIFF
--- a/src/components/FeedCard.tsx
+++ b/src/components/FeedCard.tsx
@@ -199,12 +199,12 @@ export function TitleAndBylinePlaceholder({creator}: {creator?: boolean}) {
 }
 
 export function Description({description}: {description?: string}) {
-  const [rt] = React.useState(() => {
+  const rt = React.useMemo(() => {
     if (!description) return
     const rt = new RichTextApi({text: description || ''})
     rt.detectFacetsWithoutResolution()
     return rt
-  })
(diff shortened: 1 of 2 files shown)


estrattonbailey:
Wait one sec

src/components/FeedCard.tsx
```diff
@@ -199,13 +199,14 @@ export function TitleAndBylinePlaceholder({creator}: {creator?: boolean}) {
...
 
 export function Description({description}: {description?: string}) {
-  const [rt, isResolving] = useRichText(description || '')
-  if (!description) return null
-  return isResolving ? (
-    <RichText value={description} style={[a.leading_snug]} />
-  ) : (
-    <RichText value={rt} style={[a.leading_snug]} />
-  )
+  const [rt] = React.useState(() => {
```
gaearon, line 202: i think this should be a useMemo? or it won’t be reactive

estrattonbailey: Sure might as well 👍

haileyok (approved):

PR #4622, opened 2024-06-24
Refactor `ProfileCard` to be composable

estrattonbailey:
Refactors the new `ProfileCard` component to be flexible and composable. Also handles blocked account state, see screenshot, but we should really filter those accounts out from this view.

Main exports here are `Default` (which matches the `StarterPackCard` and `FeedCard`) and `Card` which does not include a `Link` that wraps the card by default.

We could integrate `KnownFollowers` into the default exports here, or we could integrate that compositionally, tbd.

Only used in starter packs in two places right now:
CleanShot 2024-06-27 at 12 17 05@2x
CleanShot 2024-06-27 at 12 17 14@2x
CleanShot 2024-06-27 at 12 17 22@2x
CleanShot 2024-06-27 at 12 18 12@2x

### DIFF
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -79,7 +79,7 @@ export function Outer({
 }: {
   children: React.ReactElement | React.ReactElement[]
 }) {
-  return <View style={[a.flex_1, a.gap_sm]}>{children}</View>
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
LGTM

PR #4626, opened 2024-06-24
Add borders around starter pack landing page when tablet or deskto

pfrazee:
Unchanged on mobile

CleanShot 2024-06-24 at 16 20 11@2x


haileyok (approved):
👨‍🎨

PR #4630, opened 2024-06-25
Maintain portrait in app clip

haileyok:
## Why

Make sure the app clip can't rotate to horizontal mode. Using the same values here as our app.


mozzius (approved):

haileyok:
ty. not going to merge yet, since will delay OTA for a bit

PR #4611, opened 2024-06-23
Update French localization

surfdude29:
Following on from [my previous PRs](https://github.com/bluesky-social/social-app/pulls?q=is%3Apr+is%3Aclosed+%22update+french%22+author%3Asurfdude29), I again used [DeepL](https://www.deepl.com/translator) and my own knowledge of French to translate the new strings.

All feedback from native French speakers is welcome – especially @Signez of course 🙏

Most of the new strings are related to the upcoming starter packs feature, and I went with `pack de démarrage` but I have no strong attachment to that so if there's a better translation please let me know :)

I didn't really have a clue what word would be best to use for `toggle` as a noun here:

https://github.com/bluesky-social/social-app/blob/0a0c7387905c7dc61fefd4f5f27d53b4797c00f6/src/components/StarterPack/Wizard/WizardListCard.tsx#L52

so I took a guess with `bouton` but that's almost certainly wrong. Please let me know what it should actually be 😄

PR #4637, opened 2024-06-25
Update French localization

surfdude29:
As a follow-up to #4611 this translates the new strings.

I'd be grateful if you could take a look @Signez 🙏

Sorry if I got my tenses mixed up again 😬


src/locale/locales/fr/messages.po
```diff
@@ -76,6 +76,10 @@ msgstr "{0, plural, one {repost} other {reposts}}"
 msgid "{0, plural, one {Unlike (# like)} other {Unlike (# likes)}}"
 msgstr "{0, plural, one {Déliker (# like)} other {Déliker (# likes)}}"
 
+#: src/screens/StarterPack/StarterPackLandingScreen.tsx:218
+msgid "{0} joined this week"
+msgstr "{0} ont été inscrites cette semaine"
```
surfdude29, line 81: I wasn't quite sure how to handle _écriture inclusive_ here, since it's talking about people is this OK?

Signez: I don't think it would work in French to simply says `<number> did something` unless it's part of an enumeration of some sort (i.e. it would work only in this case: `10 personnes sont allées à la boulangerie, dont 5 qui ont acheté des baguettes, hon hon`). We have to precise what we are talking about.

So here, using either `comptes` or `personnes` would side-step the problem:

```suggestion
msgstr "{0} personnes se sont inscrites cette semaine"
```

src/locale/locales/fr/messages.po
```diff
@@ -5853,6 +5890,10 @@ msgstr "Préférences des fils de discussion"
 msgid "Thread Preferences"
 msgstr "Préférences des fils de discussion"
 
+#: src/components/WhoCanReply.tsx:109
+msgid "Thread settings updated"
+msgstr "Mise à jour des paramètres du fil de discussion"
```
Signez, line 5895: Well, I think we want to inform the user that it did happen, so it should be this form:

```suggestion
msgstr "Paramètres du fil de discussion mis à jour"
```

src/locale/locales/fr/messages.po
```diff
@@ -6415,6 +6456,10 @@ msgstr "Nous sommes désolés ! Vous ne pouvez vous abonner qu’à vingt étiq
 msgid "Welcome back!"
 msgstr "Bienvenue !"
 
+#: src/components/NewskieDialog.tsx:103
+msgid "Welcome, friend!"
+msgstr "Bienvenue, mon ami !"
```
Signez, line 6461: `mon ami`, or `mon amie`? yeah, complicated.

We can workaround the problem by acknowledging the context: it's the string shown when we show an infobox about ourselves and we are a newsky. So any expression we use in that context should work.

```suggestion
msgstr "Bienvenue et enchanté !"
```

Signez (commented):
Again, great translations — a few nitpicks but awesome quality as usual.

src/locale/locales/fr/messages.po
surfdude29: Makes sense to me 👌

src/locale/locales/fr/messages.po
surfdude29: Oops, yes I should have rephrased this

src/locale/locales/fr/messages.po
surfdude29: D'oh, I completely forgot that _écriture inclusive_ would apply here, of course.

But you suggested a great workaround 👍

surfdude29 (commented):
All suggestions applied, thanks so much for your corrections to my silly mistakes, and of course nitpicks are *always* appreciated 😄

surfdude29:
If this second PR that translates the new strings could also be merged before the next release is cut, that would be much appreciated @pfrazee :)

pfrazee (approved):
Ended up having to resubmit so we can include this one! Thanks all!!

PR #4638, opened 2024-06-25
Video compression in composer

mozzius:
Gated, does not include upload


src/lib/media/video/compress.ts
```diff
@@ -0,0 +1,35 @@
...
+    {
+      getCancellationId,
+      compressionMethod: 'manual',
+      bitrate: 3_000_000, // 3mbps
+      maxSize: 1920,
+    },
+    onProgress,
+  )
+
+  await FileSystem.deleteAsync(file)
```
haileyok, line 29: i think we found that this can throw and is safer to wrap in a try/catch

src/lib/media/video/compress.ts
```diff
@@ -0,0 +1,35 @@
...
+  const info = await getVideoMetaData(compressed)
+  console.log('compressed size', (info.size / 1024).toFixed(2) + 'mb')
+  console.log(JSON.stringify(info, null, 2))
```
haileyok, line 32: log

src/view/com/composer/Composer.tsx
```diff
@@ -97,6 +99,18 @@ import {TextInput, TextInputRef} from './text-input/TextInput'
...
 import hairlineWidth = StyleSheet.hairlineWidth
+// temp, ensure globalThis.expo is defined. hopefully Expo will fix this upstream
+import {ensureNativeModulesAreInstalled} from 'expo-modules-core/build/ensureNativeModulesAreInstalled'
+
+// stupid method to ensure right order of module execution
+ensureNativeModulesAreInstalled()
+const VideoPreview = lazy(async () => {
+  const {VideoPreview} = await import('./videos/VideoPreview')
+  return {default: VideoPreview}
+})
```
haileyok, line 110: ik we were talking about not using `VideoPreview`. if we don't, do we still need the `ensureNativeModulesAreInstalled()`? fine to leave for now in this gated PR but just want to know if we'll have to leave that there.

src/view/com/composer/videos/SelectVideoBtn.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }, [onSelectVideo])
+
+  return (
+    <>
+      <Button
+        testID="openGifBtn"
+        onPress={onPressSelectVideo}
+        label={_(msg`Select GIF`)}
+        accessibilityHint={_(msg`Opens GIF select dialog`)}
```
haileyok, line 47: need to change label here

src/view/com/composer/videos/VideoPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+          size="small"
+          color="primary"
+          variant="solid">
+          <ButtonText>Share</ButtonText>
+        </Button>
+
+        <Button
+          onPress={async () => {
+            if (!(await MediaLibrary.getPermissionsAsync()).granted) {
+              await MediaLibrary.requestPermissionsAsync()
```
haileyok, line 65: we should pass in the granular permissions here for android, see https://docs.expo.dev/versions/latest/sdk/media-library/#medialibraryrequestpermissionsasyncwriteonly-granularpermissions

haileyok: looks like we should do it in the `getPermissionsAsync()` too https://docs.expo.dev/versions/latest/sdk/media-library/#medialibrarygetpermissionsasyncwriteonly-granularpermissions

src/view/com/composer/videos/VideoPreview.tsx
```diff
@@ -0,0 +1,78 @@
...
+            }
+            await MediaLibrary.createAssetAsync(video.uri)
```
haileyok, line 67: also i think we need to handle a denied state here. for now i think a toast is fine, that's what we do elsewhere i believe.

src/view/com/composer/videos/state.ts
```diff
@@ -0,0 +1,49 @@
...
+  const {_} = useLingui()
+  const [progress, setProgress] = useState(0)
+
+  const {mutate, data, isPending, isError, reset, variables} = useMutation({
+    mutationFn: async (asset: ImagePickerAsset) => {
+      if (asset.fileSize) {
+        console.log(
+          'uncompressed size',
+          (asset.fileSize / 1024 / 1024).toFixed(2) + 'mb',
+        )
```
haileyok, line 19: log (sry i know it's a draft just want to make sure i get these in while im here 😅)

src/view/com/composer/videos/state.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+      const compressed = await compressVideo(asset.uri, {
+        onProgress: num => setProgress(trunc2dp(num)),
+      })
+
+      return compressed
+    },
+    onError: error => {
+      console.error('error', error)
```
haileyok, line 28: log

haileyok: can this error for any reason other than the throw in `compressVideo`? if so would be nice to use `logger.error()` for things that are _not_ the `greater than 100mb` error just so we don't spam sentry. otherwise i think just toast is fine.

haileyok (commented):
LFG!!

have not tested, but things look good. couple questions and nits but otherwise think we could just put this in behind gate - seems that this only adds the button for video upload which would be fine to get in.

will leave it up to you if you want to strip out `VideoPreview` before merging.

src/view/com/composer/videos/state.ts
mozzius: I'll just show toast for now

src/view/com/composer/videos/VideoPreview.tsx
mozzius: This is just a temp thing while developing to export the compressed video to examine it. I'll remove it

haileyok:
hmm one error on web, let me see if i can figure out why. 

Screenshot 2024-07-05 at 4 39 37 PM

haileyok:
Getting an error whenever I try to compress on native.

Screenshot 2024-07-05 at 5 10 51 PM

haileyok:
On Android seems like compression works, but am getting this error instead

```
 ERROR  Your app just crashed. See the error below.
java.lang.IllegalStateException: setPictureInPictureParams: Current activity does not support picture-in-picture.
  android.os.Parcel.createExceptionOrNull(Parcel.java:3065)
  android.os.Parcel.createException(Parcel.java:3041)
  android.os.Parcel.readException(Parcel.java:3024)
  android.os.Parcel.readException(Parcel.java:2966)
  android.app.IActivityClientController$Stub$Proxy.setPictureInPictureParams(IActivityClientController.java:1966)
  android.app.ActivityClient.setPictureInPictureParams(ActivityClient.java:363)
  android.app.Activity.setPictureInPictureParams(Activity.java:3009)
  expo.modules.video.VideoView.applyRectHint(VideoView.kt:255)
  expo.modules.video.VideoView.onLayout(VideoView.kt:275)
  android.view.View.layout(View.java:24421)
  android.view.ViewGroup.layout(ViewGroup.java:6440)
  expo.modules.video.VideoView.mLayoutRunnable$lambda$3(VideoView.kt:125)
  expo.modules.video.VideoView.$r8$lambda$oOlWV_EIAOBus6c_5bDslSW12u4(Unknown Source:0)
  expo.modules.video.VideoView$$ExternalSyntheticLambda0.run(Unknown Source:2)
  android.os.Handler.handleCallback(Handler.java:958)
  android.os.Handler.dispatchMessage(Handler.java:99)
  android.os.Looper.loopOnce(Looper.java:205)
  android.os.Looper.loop(Looper.java:294)
  android.app.ActivityThread.main(ActivityThread.java:8177)
  java.lang.reflect.Method.invoke(Native Method)
  com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
  com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
  ```

File: src/lib/media/video/compress.ts [hunk 9]
```diff
@@ -0,0 +1,37 @@
...
+      maxSize: 1920,
+    },
+    onProgress,
+  )
+
+  try {
+    await FileSystem.deleteAsync(file)
+  } catch {
+    console.warn('Failed to delete original video')
+  }
```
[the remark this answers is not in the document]
haileyok: I RM'd this for now. Don't think we need it (and I don't think Android will actually let us delete it anyway so)

haileyok:
Works well on-device on all platforms! 🥳

efstathiosntonas:
@haileyok hi, why this was reverted? Thanks

surfdude29:
> @haileyok hi, why this was reverted? Thanks

Hailey can explain better than I can of course, but it looks like this was only reverted in the branch for the OTA (Over The Air) update for 1.88, which has a few important bug fixes that were merged after 1.88 was sent for app store review:

https://github.com/bluesky-social/social-app/commits/1.88.0-ota-1

efstathiosntonas:
thanks for the heads up @surfdude29. 

btw, I believe this should should use ffmpeg instead and create .m3u8 with `vod` flag so the video is streamed instead of a single mp4 file, unless I'm missing something 😅

haileyok:
@surfdude29 is right, it was reverted for an OTA deploy where the native change included in this PR would have prevented us from deploying it.

Regarding file type here, we are just compressing the file for uploading and processing on the server - the end file that users are streamed is not going to be an mp4, it will be a m3u8.

efstathiosntonas:
thanks @haileyok, I knew I was missing something :)

PR #4640, opened 2024-06-25
ensure we get navigated away to starter pack when signed in

haileyok:


pfrazee (approved):
Looks good, tested on web

PR #4641, opened 2024-06-25
tweak app clip AASA

haileyok:
AASA requires `applinks` to be present, even if the domain is only used for an app clip.

### DIFF
--- a/bskyweb/static/.well-known/apple-app-site-association
+++ b/bskyweb/static/.well-known/apple-app-site-association
@@ -1,8 +1,5 @@
 {
   "applinks": {
-    "appclips": {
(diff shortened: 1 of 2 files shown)


bskyweb/static/.well-known/apple-app-site-association
```diff
@@ -11,5 +8,8 @@
         ]
       }
     ]
-  }
+  },
+  "appclips": {
+    "apps": ["B3LX46C5HS.xyz.blueskyweb.app.AppClip"]
+  },
```
gaearon, line 14: Note this is invalid JSON (trailing comma), is that OK?

haileyok:
```suggestion
  }
```

PR #4644, opened 2024-06-25
FeedCard & ListCard cleanups

estrattonbailey:
This PR refactors `FeedCard` a bit and extracts a `ListCard` component for easier usage in either case. If you're rendering a feed, use `FeedCard`. List, use `ListCard`.

`ListCard` inherits from `FeedCard`, since they're very similar, but I think the clarity of usage like this makes this pattern worth it. If we need to further differentiate in the future we can pretty easily like this. Examples:
- lists don't have likes, no need for `Likes` component
- maybe one day we need a different action export for modlists like `SubscribeButton`
- `TitleAndByline` needs a `purpose` for lists, but not for feeds
  - maybe soon we need a `type` or `purpose` field for feeds  👀

### DIFF
--- a/src/components/FeedCard.tsx
+++ b/src/components/FeedCard.tsx
@@ -3,6 +3,7 @@ import {GestureResponderEvent, View} from 'react-native'
 import {
   AppBskyActorDefs,
   AppBskyFeedDefs,
+  AppBskyGraphDefs,
   AtUri,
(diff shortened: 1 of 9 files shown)


pfrazee (approved):
Reviewed code, tested on web, lgtm

PR #4650, opened 2024-06-26
Fix footer messages in starter pack wizard

tkusano:
At footer messages in starter pack wizard, when selected two feeds, first of selected items are displayed as 'You', not the feed's name. This PR fixes this problem.

This PR also contains companion Japanese message updates.


gaearon:
Can you pls separate the functional change and the translation into separate PRs? Also a screenshot of before/after would be very helpful. Thanks.

gaearon:
The problem with coupling a functional change to a translation change is that it's more difficult to tell how to modify other translations, and kind of leaves us in an inconsistent state. Ideally we'd just make the fix alone without invalidating translations if possible.

gaearon (approved):

gaearon:
verified English, French, Japanese


https://github.com/bluesky-social/social-app/assets/810438/49e524ca-c282-4ccc-b3f5-8f8dfb09a44a

tkusano:
Ah sorry to mix functional changes and translation updates. Thanks.

PR #4659, opened 2024-06-26
Better starterpack embed

haileyok:
**[without whitespace](https://github.com/bluesky-social/social-app/pull/4659/files?w=1)**

Temporarily using an external embed for this. The app view is actually returning a `NotFound` record for these, so the normal way of embedding a record won't do great. The image being displayed keeps the info fresh though.

Will add go.bsky.app resolution in a separate PR.

Note: resolution won't work on localhost, because of the origin header. Has to be on a bsky.app domain.


src/lib/link-meta/resolve-short-link.ts
```diff
@@ -0,0 +1,23 @@
+import {logger} from '#/logger'
+import {startUriToStarterPackUri} from 'lib/strings/starter-pack'
+
+export async function resolveShortLink(shortLink: string) {
+  const controller = new AbortController()
+  const to = setTimeout(() => controller.abort(), 2e3)
```
haileyok, line 6: these _should_ resolve in two seconds, but we can bump to maybe 5 if we need to.

pfrazee (approved):
Tested on iOS simulator, seems solid

PR #4660, opened 2024-06-26
[Starter Packs] Posts tab

gaearon:
Yay or nay?

https://github.com/bluesky-social/social-app/assets/810438/1e93a4be-b46c-4b48-b0c0-01da7ba7098b

### DIFF
--- a/src/screens/StarterPack/StarterPackScreen.tsx
+++ b/src/screens/StarterPack/StarterPackScreen.tsx
@@ -210,7 +210,7 @@ function StarterPackScreenInner({
                 />
               )
             : null}
-          {showPostsTab != null
+          {showPostsTab
             ? ({headerHeight, scrollElRef}) => (
                 <PostsList
                   // Validated above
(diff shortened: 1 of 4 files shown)


src/screens/StarterPack/StarterPackScreen.tsx
```diff
@@ -206,6 +210,18 @@ function StarterPackScreenInner({
                 />
               )
             : null}
+          {showPostsTab != null
```
gaearon, line 213: oops should be just boolean

estrattonbailey (approved):
NICE, just tested on web and works great, feeds tab or no

pfrazee (approved):
Tested on Web and iOS, works well

I'm +1 I think

estrattonbailey:
Tested query invalidation using iOS too e.g. remove all feeds and it removes the tab.

Yeah I'm +1, nice little preview of what you get.

Issue #4656, opened 2024-06-26
Copy/paste images


YuriGal:
**Describe the bug**

Copying a image from a Web page, and pasting it to a Bluesky post doesn't work in desktop browsers

**To Reproduce**

1. Using a Windows or a MacOS computer browse to any page with images
2. Copy an image from the page into clipboard
3. Paste image into a Bluesky post

**Expected behavior**

Image should appear in the post

**Details**

 - Platform: On desktop (Windows or Mac) Chrome or Firefox
 - App version: 1.87.0

**Additional context**

It used to work, but now is broken.

gaearon:
cc @mary-ext might be related to https://github.com/bluesky-social/social-app/pull/4327?

mary-ext:
oops thought I had covered this pretty well but it appears not, sorry!

PR #4670, opened 2024-06-27
Fix pasting images on web

gaearon:
Fixes https://github.com/bluesky-social/social-app/issues/4656.

This regressed in https://github.com/bluesky-social/social-app/pull/4327. Copying an image _does_ include `text/html` so the fix is to do both the `text/html` condition _and_ to run the code below searching for images.

## Test Plan

Copied an image, tried pasting it in Chrome, Safari, and FF. It works after the fix.

I haven't verified whether #4327 itself regressed (for Edge) but I don't see why it would. We're still taking plain text for the main paste, we just happen to _also_ look for images now. So it's probably fine.


estrattonbailey (approved):

pfrazee:
Nice thanks

PR #4673, opened 2024-06-27
Improve textinput performance in login and account creation

pfrazee:
There's noticeable flicker and input-lag in the login & account-creation forms. This PR resolves those issues.

- In the login form, switch to uncontrolled components
- ~In account creation, debounce dispatches to reduce renders (the components were already uncontrolled). I tested the debounce pretty extensively and have not felt the delay affects the it too heavily, though it is visible.~
- In account creation, moved validation into the "next" button and ditched the "can next" logic, removing the renders that were getting triggered by value changes.

Tested on all 3 platforms, with sim tests, and on-device with my iPhone.


gaearon (commented):
debouncing can lose user input if they click next too fast. feels like introducing more complexity into the logic than i'd be super comfortable with.

can we instead move fast-changing inputs to more local state (so only the form re-renders) or make them uncontrolled (and store the values in `onChange` callback into a ref) and then only put it into the reducer during the actual step transition? i.e. group it with the "next step" dispatch instead of dispatching on every keystroke.

pfrazee:
Okay, refactored to remove the debounce solution

gaearon (approved):
code lgtm. have not tested.

PR #4680, opened 2024-06-27
Handle pressing all go.bsky.app links in-app w/ resolution

haileyok:
## Why

We want to be able to resolve any short links in the app, even if they don't get resolved while posting. This is a little challenging because we have to reach out to `go.bsky.app` first. I added JSON response support in https://github.com/bluesky-social/social-app/pull/4671, which we can now use to perform resolution when landing on these.

## How

1. First, we click on a link in the app. We already check for if links start with `bsky.app` but we don't check `go.bsky.app` which we add here.
2. We then push to the `StarterPackScreenShort` route
3. The short route attempts to resolve the real URI and then render the starter pack screen with the resolved did and rkey

The nice thing here is this is extensible to other schemes whenever we want to. For example, post short links - which we are less worried about being _very_ short - could use `/p/123456` and we could create a resolve query for those types of links.

## Test Plan

RocketSim_Recording___RNIDE_TEMP_SIM_com apple CoreSimulator SimDeviceType iPhone-15-Pro_6 1_2024-06-27_12 43 20

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -325,12 +325,12 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
       <Stack.Screen
         name="StarterPack"
         getComponent={() => StarterPackScreen}
-        options={{title: title(msg`Starter Pack`), requireAuth: true}}
+        options={{title: title(msg`Starter Pack`)}}
       />
       <Stack.Screen
         name="StarterPackShort"
         getComponent={() => StarterPackScreenShort}
-        options={{title: title(msg`Starter Pack`), requireAuth: true}}
+        options={{title: title(msg`Starter Pack`)}}
       />
       <Stack.Screen
         name="StarterPackWizard"
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -43,7 +43,10 @@ import HashtagScreen from '#/screens/Hashtag'
 import {ModerationScreen} from '#/screens/Moderation'
 import {ProfileKnownFollowersScreen} from '#/screens/Profile/KnownFollowers'
 import {ProfileLabelerLikedByScreen} from '#/screens/Profile/ProfileLabelerLikedBy'
-import {StarterPackScreen} from '#/screens/StarterPack/StarterPackScreen'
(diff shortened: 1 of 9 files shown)


src/Navigation.tsx
```diff
@@ -324,6 +327,11 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
         getComponent={() => StarterPackScreen}
         options={{title: title(msg`Starter Pack`), requireAuth: true}}
       />
+      <Stack.Screen
+        name="StarterPackShort"
+        getComponent={() => StarterPackScreenShort}
+        options={{title: title(msg`Starter Pack`), requireAuth: true}}
+      />
```
haileyok, line 334: A new route that accepts a `code` parameter rather than name and rkey

src/lib/strings/url-helpers.ts
```diff
@@ -167,6 +168,9 @@ export function convertBskyAppUrlIfNeeded(url: string): string {
     } catch (e) {
       console.error('Unexpected error in convertBskyAppUrlIfNeeded()', e)
     }
+  } else if (!isWeb && isShortLink(url)) {
+    // We only want to do this on native, web handles the 301 for us
+    return shortLinkToHref(url)
```
haileyok, line 173: We don't need to do this for web, so only do it on native. Web can handle the 301 elegantly without weirdness.

src/screens/StarterPack/StarterPackLandingScreen.tsx
```diff
@@ -112,9 +115,6 @@ function LandingScreenLoaded({
   const listItemsCount = starterPack.list?.listItemCount ?? 0
 
   const onContinue = () => {
-    setActiveStarterPack({
-      uri: starterPack.uri,
-    })
```
haileyok, line 117: Just removing this since it is unneeded. No change in functionality.

gaearon (approved):
have not run, code lgtm (after discussed changes)

haileyok:
Retested signup flow w/ a starter pack, works good. This is a little funky as noted but is mostly a corner case.

PR #4682, opened 2024-06-27
Add custom user ID for starter pack signups

haileyok:
#### Android

- Connect a physical device - it's hard to access local host on Android emulators...
- Set `versionCode` in the `android` block of `app.config.js` to some big number, i.e. 500. Run a prebuild.
- Build the app with `cd android && ./gradlew assembleRelease`
- Open a starter pack URL in Chrome: `http://localhost:19006/start/pfrazee.com/3kvaphb2hpf2u`
- Press Download on Google Play
- Don't install the app through Play, but by running - from the `android` directory - `adb install app/build/outputs/apk/release/app-release.apk`
- Step through account creation a bit - you don't have to complete it, but just enough to send a few events to Statsig.

You should be able to check the event stream for recent `ctaPress` events and find your `starterPackId` in the event. You then should be able to search for other events by that ID - and should see events for the things you performed in the app. Just pressing the next button in sign up will be enough to send something to statsig with your SP id.

Note that you probably need to toggle "show non production logs"

#### iOS

- Open XCode, open the `.xcworkspace` file inside of `ios`. _Don't_ use the `.xcodeproj` file
- Make sure the simulator you are using does _not_ have Bluesky installed
- Start by running the App Clip
  - Select `Bluesky Clip` in the dropdown at the top
  - In the same dropdown, press `Edit Scheme`
  - Enter the localhost starter pack URL for `App Clip Invocation URL`
  - Press the build button button
  - Press the Join Bluesky button, then close the App Clip
- Select `Bluesky` from the dropdown at the top of Xcode and build
- Click through a bit of onboarding
- Do the same as Android to verify events are sent

Screenshot 2024-07-03 at 3 28 24 AM

Screenshot 2024-07-03 at 3 29 31 AM

Screenshot 2024-07-03 at 3 38 07 AM

Screenshot 2024-07-03 at 3 40 04 AM


modules/BlueskyClip/ViewController.swift
```diff
@@ -43,21 +43,14 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
...
 
       self.presentAppStoreOverlay()
       defaults?.setValue(url.absoluteString, forKey: "starterPackUri")
-
-    case .store:
-      guard let keyToStoreAs = payload.keyToStoreAs, let jsonToStore = payload.jsonToStore else {
-        return
-      }
-
-      self.defaults?.setValue(jsonToStore, forKey: keyToStoreAs)
```
haileyok, line 60: These were unused so just removing.

modules/BlueskyClip/ViewController.swift
```diff
@@ -128,6 +121,5 @@ struct WebViewActionPayload: Decodable {
   }
 
   let action: Action
-  let keyToStoreAs: String?
-  let jsonToStore: String?
```
haileyok, line 132: same here

PR #4692, opened 2024-06-28
Handle pushing to starterpack screen when unauthed

haileyok:


src/screens/StarterPack/StarterPackScreen.tsx
```diff
@@ -330,16 +298,41 @@ function Header({
...
+    }
+    const onBeforeRemove = () => {
+      setActiveStarterPack(undefined)
+    }
+
+    navigation.addListener('focus', onFocus)
+    navigation.addListener('beforeRemove', onBeforeRemove)
+
+    return () => {
+      console.log('remove listeners')
```
gaearon, line 330: oops

src/screens/StarterPack/StarterPackScreen.tsx
```diff
@@ -330,16 +298,41 @@ function Header({
...
+
+  React.useEffect(() => {
+    const onFocus = () => {
+      if (hasSession) return
+      setActiveStarterPack({
+        uri: starterPack.uri,
+      })
+    }
+    const onBeforeRemove = () => {
+      setActiveStarterPack(undefined)
```
gaearon, line 323: needs `hasSession` check or no?

src/screens/Signup/StepInfo/index.tsx
```diff
@@ -27,10 +27,18 @@ function sanitizeDate(date: Date): Date {
...
+export function StepInfo({
+  isLoadingStarterPack,
+}: {
+  isLoadingStarterPack: boolean
+}) {
   const {_} = useLingui()
   const {state, dispatch} = useSignupContext()
 
+  useEffect(() => {
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut)
```
gaearon, line 39: This is kind of a race condition, effects have no well-defined lifetime in RN right now so this call may align with some different paint. Can we move this closer to the setstate we actually want to animate?

gaearon (approved):

src/screens/Signup/StepInfo/index.tsx
haileyok: going to leave for now and see per conversation.

Issue #4433, opened 2024-06-08
Add Javanese localization


febrimuhnas:
Hello!
I would like to help translate the app to Javanese. I'm a native speaker of Javanese.

surfdude29:
Hello 👋  I'm a volunteer and I help out with translations a bit.

I can't speak for the Bluesky team but my understanding on adding new translations is that they'd be happy to add a Javanese translation as long as it has been approved by at least one other native speaker.

This is the English message catalog containing the strings that need to be translated:

https://github.com/bluesky-social/social-app/blob/main/src/locale/locales/en/messages.po

One tip: You don't need to bother translating the strings that start with `~#` as those are obsolete and are no longer present in the app (they'll be removed from the message catalog at some point).

Depending on your preferences, you (and any other Javanese language translators) may find it easier to set up a Crowdin project to collaborate and then submit a pull request once the translation is complete.

This is what the Indonesian and Ukrainian language teams have done:

https://crowdin.com/project/bluesky-id
https://crowdin.com/project/bsky-app-ua

I hope that's helpful, let me know if you have any questions :)

febrimuhnas:
Hello, @surfdude29 👋
_maturnuwun_ (thankyou) for your advice. I've made it.

PR #4694, opened 2024-06-28
Remove superseded strings from en

auroursa:
Similar to #3462
This PR intends to remove unused strings from en, to make it easier for other translators to translate into new languages based on the en file.

### DIFF
--- a/src/locale/locales/en/messages.po
+++ b/src/locale/locales/en/messages.po
@@ -26,16 +26,8 @@ msgid "{0, plural, one {{formattedCount} other} other {{formattedCount} others}}
 msgstr ""
 
 #: src/components/moderation/LabelsOnMe.tsx:55
-#~ msgid "{0, plural, one {# label has been placed on this account} other {# labels has been placed on this account}}"
-#~ msgstr ""
-
-#: src/components/moderation/LabelsOnMe.tsx:55
 msgid "{0, plural, one {# label has been placed on this account} other {# labels have been placed on this account}}"
 msgstr ""
-
-#: src/components/moderation/LabelsOnMe.tsx:61
-#~ msgid "{0, plural, one {# label has been placed on this content} other {# labels has been placed on this content}}"
-#~ msgstr ""
 
 #: src/components/moderation/LabelsOnMe.tsx:61
 msgid "{0, plural, one {# label has been placed on this content} other {# labels have been placed on this content}}"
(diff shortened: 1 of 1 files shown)


gaearon:
What's the methodology? Is there an automated command to do this?

surfdude29:
Sorry to jump in here, when I submitted my previous PRs to remove obsolete strings from message catalogs I just edited the files manually. However, I don't know what methodology @auroursa used of course.

There is a way to do it automatically using the `--clean` command:

https://lingui.dev/ref/cli

IMG_5111

auroursa:
@surfdude29 Thank you for letting me know. Indeed I cleaned it up manually, and it seems much simpler to do it via clean command.

auroursa:
Maybe we can run the clean command together with extract automatically each time. This way we can automate this process.

auroursa:
> Sorry to jump in here, when I submitted my previous PRs to remove obsolete strings from message catalogs I just edited the files manually. However, I don't know what methodology @auroursa used of course.
> 
> 
> 
> There is a way to do it automatically using the `--clean` command:
> 
> 
> 
> https://lingui.dev/ref/cli
> 
> 
> 
> IMG_5111

Maybe we can run the clean command together with extract automatically each time. This way we can automate this process.

auroursa:
> > Maybe we can run the clean command together with extract automatically each time. This way we can automate this process.
> 
> 
> 
> I think this would be a good idea, but my preference would be to do this only for the English message catalog. My reasoning is that sometimes there's only a very minor change to a string (e.g. adding or removing punctuation) and thus the already-agreed upon translation can easily be reused from the same file either in full or in part by simply making the minor change required.
> 
> 
> 
> If obsolete translations were removed automatically from all languages, it would mean that translators would have to refer back to the previous version of the file in order to reuse a translation that had been removed, which would be a bit annoying.
> 
> 
> 
> However, maybe I'm making a mountain out of a molehill and it wouldn't be that big a deal.
> 
> 
> 
> I'd be interested to hear other translators' views on this, as removing obsolete strings automatically would of course save translators from doing this themselves (as most seem to do) and thus save time and effort in that regard.

It's a good idea to automatically delete outdated strings in English only. It's better to leave the cleanup of other languages to the respective translators.

During the actual translation process, I've indeed encountered situations where changes in capitalization or punctuation generated new strings. In such cases, the outdated strings can still serve as a reference.

I know that running the extract command easily creates new language translations, making the automatic cleanup of outdated English strings somewhat redundant. But hey, when someone wants to contribute a new translation, they only need to copy the English file and make changes. This means they don't need a complete build environment, which could lower the barrier for contributing translation PRs.

surfdude29:
> But hey, when someone wants to contribute a new translation, they only need to copy the English file and make changes. This means they don't need a complete build environment, which could lower the barrier for contributing translation PRs.

This is exactly what I suggest to users who offer to contribute a new translation, e.g. when [a user offered to contribute a Javanese translation](https://github.com/bluesky-social/social-app/issues/4433#issuecomment-2156123015). That way, all the user needs in order to contribute is a text editor.

PR #4699, opened 2024-06-28
Add starter pack embeds to posts

haileyok:
Relies on https://github.com/bluesky-social/atproto/pull/2613 being deployed or using dev env from that branch.

Note that these will work with `go.bsky.app` links as well, though those records wouldn't work in the local dev env. Once ^ gets deployed we should just test again to make sure.

## Web

Screenshot 2024-06-28 at 12 46 41 PM

Screenshot 2024-06-28 at 12 46 47 PM

## Native

Screenshot 2024-06-28 at 12 56 24 PM

Screenshot 2024-06-28 at 12 56 28 PM

### DIFF
--- a/src/components/StarterPack/StarterPackCard.tsx
+++ b/src/components/StarterPack/StarterPackCard.tsx
@@ -136,6 +136,7 @@ export function Embed({starterPack}: {starterPack: StarterPackViewBasic}) {
   return (
     <View
       style={[
+        a.mt_xs,
         a.border,
         a.rounded_sm,
         a.overflow_hidden,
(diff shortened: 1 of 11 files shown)


src/components/StarterPack/StarterPackCard.tsx
```diff
@@ -104,14 +110,42 @@ export function Link({
   }
 
   return (
-    <InternalLink
```
gaearon, line 107: I think we actually want to keep this a link. E.g. cmd+click should open in new tab. Any reason for this change?

gaearon (changes requested):
let's figure out how to keep it a link, lgtm otherwise

gaearon (approved):
lgtm

PR #4701, opened 2024-06-29
Fix Gujarati language code

mary-ext:
As per title


mary-ext:
Nevermind there's already an existing pull request

PR #4711, opened 2024-07-01
Fix iOS native dropdowns to ignore the first press outside

gaearon:
Fixes the native dropdown on iOS to ignore the first press outside while active. This prevents misclicks when using the menu.

I didn't add any background to it — other apps I tested don't do that either.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/4711/files?w=1)

## Before

https://github.com/bluesky-social/social-app/assets/810438/ea5d2ed1-9cc6-463b-b630-014855e1ba00

## After

https://github.com/bluesky-social/social-app/assets/810438/11af8601-8567-4c78-9aa5-dfcd247fba1e

## Test Plan

See above for iOS. No change on other platforms.

Tested different callsites for `NativeDropdown`:

- Avi follow button
- Profile list context menu
- Account switcher

### DIFF
--- a/src/view/com/util/forms/NativeDropdown.tsx
+++ b/src/view/com/util/forms/NativeDropdown.tsx
@@ -1,13 +1,15 @@
 import React from 'react'
+import {Platform, Pressable, StyleSheet, View, ViewStyle} from 'react-native'
+import {IconProp} from '@fortawesome/fontawesome-svg-core'
 import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Seems fine but didnt test, we should test Voice Over once it lands just to make sure that's solid. iOS only so I think it should be okay. Easier to land in OTA to test that, otherwise have to make a build locally.

gaearon:
Tried building for device, can't seem to get pairing to work — let's test on OTA yeah

PR #4712, opened 2024-07-01
Add events to signup for captcha results

haileyok:
## Why

It would be nice to know whenever people successfully complete the captcha. There's some weird stats that are hard to understand if users are not actually failing captchas - which is the most likely 
reason we're seeing this. It would be good to have confirmation on this though.

## How

Adds a success and failure event that get logged upon completion of the captcha - whether it fails or not.

## Test Plan

Check network tab while stepping through onboarding to ensure events get logged where they should.

Screenshot 2024-07-01 at 2 25 26 PM

Screenshot 2024-07-01 at 2 25 42 PM

Screenshot 2024-07-01 at 2 29 56 PM

### DIFF
--- a/src/screens/Signup/StepCaptcha/index.tsx
+++ b/src/screens/Signup/StepCaptcha/index.tsx
@@ -6,6 +6,7 @@ import {nanoid} from 'nanoid/non-secure'
 
 import {createFullHandle} from '#/lib/strings/handles'
 import {logger} from '#/logger'
+import {logEvent} from 'lib/statsig/statsig'
 import {ScreenTransition} from '#/screens/Login/ScreenTransition'
 import {useSignupContext, useSubmitSignup} from '#/screens/Signup/state'
 import {CaptchaWebView} from '#/screens/Signup/StepCaptcha/CaptchaWebView'
@@ -39,6 +40,7 @@ export function StepCaptcha() {
   const onSuccess = React.useCallback(
     (code: string) => {
       setCompleted(true)
+      logEvent('signup:captchaSuccess', {})
       submit(code)
     },
     [submit],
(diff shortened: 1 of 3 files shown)


src/screens/Signup/StepCaptcha/index.tsx
```diff
@@ -50,6 +52,7 @@ export function StepCaptcha() {
         type: 'setError',
         value: _(msg`Error receiving captcha response.`),
       })
+      logEvent('signup:captchaFailure', {})
       logger.error('Signup Flow Error', {
```
estrattonbailey, line 56: Odd that we don't have any errors in Sentry for this

src/screens/Signup/index.tsx
```diff
@@ -112,6 +112,14 @@ export function Signup({onPressBack}: {onPressBack: () => void}) {
       }
     }
 
+    logEvent('signup:nextPressed', {
+      activeStep: state.activeStep,
+      phoneVerificationRequired:
+        state.activeStep === SignupStep.HANDLE
+          ? Boolean(state.serviceDescription?.phoneVerificationRequired)
+          : undefined,
```
estrattonbailey, line 120: It should be defined on the first step too right? Could always add it here in that case. I suppose if we're only checking for this value on `activeStep: 1` then it's same difference though.

src/screens/Signup/StepCaptcha/index.tsx
haileyok: Oh you know I probably filtered this out when we were getting slammed during the Indonesia wave. Let me check.

src/screens/Signup/index.tsx
haileyok: Oh yea, I just added it to this step so that we can explicitly see if it's undefined here for some reason. Seems unlikely but idk.

estrattonbailey (approved):

src/screens/Signup/StepCaptcha/index.tsx
haileyok: 👍 see slack

src/screens/Signup/index.tsx
haileyok: just made it always show up, can just determine which screen by `activeStep` like you said 👍

PR #4715, opened 2024-07-01
Don't follow self when following all

haileyok:
## Why

Simple, just don't follow yourself (still love that you can do this lol) whenever pressing Follow All in a starter pack

## Test Plan

Follow all on a starter pack, see that you no longer follow yourself.

### DIFF
--- a/src/screens/StarterPack/StarterPackScreen.tsx
+++ b/src/screens/StarterPack/StarterPackScreen.tsx
@@ -343,7 +343,11 @@ function Header({
         list: starterPack.list.uri,
       })
       const dids = list.data.items
-        .filter(li => !li.subject.viewer?.following)
+        .filter(
+          li =>
+            li.subject.did !== currentAccount?.did &&
+            !li.subject.viewer?.following,
+        )
         .map(li => li.subject.did)
 
       const followUris = await bulkWriteFollows(agent, dids)


src/components/StarterPack/Main/ProfilesList.tsx
```diff
@@ -47,7 +47,10 @@ export const ProfilesList = React.forwardRef<SectionRef, ProfilesListProps>(
     // The server returns these sorted by descending creation date, so we want to invert
     const profiles = data?.pages
       .flatMap(p => p.items.map(i => i.subject))
-      .filter(p => !p.associated?.labeler)
+      .filter(
+        p =>
+          !p.viewer?.blockedBy && !p.viewer?.blocking && !p.associated?.labeler,
```
gaearon, line 52: is this logic copypaste from somewhere? any reference points?

src/screens/StarterPack/StarterPackScreen.tsx
```diff
@@ -343,7 +343,15 @@ function Header({
         list: starterPack.list.uri,
       })
       const dids = list.data.items
-        .filter(li => !li.subject.viewer?.following)
+        .filter(
+          li =>
+            li.subject.did !== currentAccount?.did &&
```
gaearon, line 348: same q

src/components/StarterPack/Main/ProfilesList.tsx
haileyok: idt so (in either this case or the one below). this is the one weird place where we actually don't filter the blocks out on the server because of how other lists work. we could extract it to a function if we want to though even if it doesn't get used much/ever again.

haileyok: (probably something like `isBlockedOrMuted()`?)

haileyok: did that, though made them two separate things (blocked vs muted)

gaearon (approved):

PR #4717, opened 2024-07-02
Fix GIFs in DMs

mozzius:
Changes `maxWidth: 100%` to `width: 100%`. GIFs are full-width anyway so I don't believe this should cause any issues.

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Screenshot 2024-07-02 at 16 21 43</td>
    <td>Screenshot 2024-07-02 at 16 21 27</td>
  </tr>
</table>

## Test plan

Check GIFs still work in other places across the app


gaearon (approved):
sure ok

PR #4718, opened 2024-07-02
Allow rendering iframe from localhost 

gaearon:
Have not tested but this is my guess for how to do it.

The goal is to fix this on localhost:

Screenshot 2024-07-02 at 18 55 49


bnewbold:
This is old; is it still needed/helpful, or can we close?

PR #4719, opened 2024-07-02
[D1X] Pull out follow-backs for higher signal

estrattonbailey:
If someone follows you _back_, this PR ungroups that notification from the other follows and shows it as below. Multiple follow-backs are shown separately.

CleanShot 2024-07-02 at 15 00 46@2x

### DIFF
--- a/src/state/queries/notifications/feed.ts
+++ b/src/state/queries/notifications/feed.ts
@@ -26,6 +26,7 @@ import {
   useQueryClient,
 } from '@tanstack/react-query'
 
+import {useGate} from '#/lib/statsig/statsig'
 import {useAgent} from '#/state/session'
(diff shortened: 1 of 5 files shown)


src/state/queries/notifications/feed.ts
```diff
@@ -81,6 +83,7 @@ export function useNotificationFeedQuery(opts?: {enabled?: boolean}) {
             queryClient,
             moderationOpts,
             fetchAdditionalData: true,
+            ungroupFollowBacks: gate('ungroup_follow_backs'),
```
gaearon, line 86: This makes the check early — before we know if it would have any effect (it would only have effect if someone actually follows you back). So we're going to lose some statistical power if we do it this way.

gaearon (approved):

PR #4727, opened 2024-07-03
[D1X] Use user action and viewing history to inform suggested follows

estrattonbailey:
Keeps an in-memory store of liked posts, followed users, and post URIs seen on the Discover feed. This is only used in memory and is not stored or transmit anywhere.

This data is then used to grab up to 16 profiles, which are then filtered for any users already followed by the acting user. Up to 6 are then shown to the user as suggestions for others to follow.

### DIFF
--- a/src/state/userActionHistory.ts
+++ b/src/state/userActionHistory.ts
@@ -33,29 +33,29 @@ export function useActionHistorySnapshot() {
   return React.useState(() => getActionHistory())[0]
 }
 
-export function like(...postUris: string[]) {
+export function like(postUris: string[]) {
   userActionHistory.likes = userActionHistory.likes
     .concat(postUris)
     .slice(-LIKE_WINDOW)
 }
-export function unlike(...postUris: string[]) {
+export function unlike(postUris: string[]) {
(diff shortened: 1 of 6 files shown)


gaearon (approved):
have not tested, code lgtm. need to resolve conflict to land

src/state/userActionHistory.ts
```diff
@@ -0,0 +1,62 @@
...
+    .slice(-FOLLOW_WINDOW)
+}
+export function unfollow(...dids: string[]) {
+  userActionHistory.follows = userActionHistory.follows.filter(
+    uri => !dids.includes(uri),
+  )
+}
+
+export function seen(...postUris: string[]) {
+  userActionHistory.seen = userActionHistory.seen
```
gaearon, line 59: This is not super great because it'll re-allocate a growing array every time you see something (which is a lot of times). I think what we want to do here is to keep a mutable array, but snapshot it at the time `useActionHistorySnapshot` is called.

gaearon: ah nvm this only happens on pagination, i thought it's per-seen

PR #4728, opened 2024-07-04
Fix `onEndReached` not firing sometimes on web

haileyok:
## Why

Currently, if the tail intersection observer is intersecting, and then we fetch more items but not enough to push it offscreen, we won't get another `onEndReached` event (because we only fire `onEndReached` when the intersection status _changes_ but in this case it's `true` -> `true`). We need to retrigger `onEndReached` if it's _still_ intersecting after fetch.

## How

Add `key={data.length}` to the intersection observer at the tail. This remounts it if the number of items changes, so it re-executes the intersection callback if it's still close to the viewport.

## Test Plan

You can set the `onEndReachedThreshold` inside of `src/view/com/notifications/Feed.tsx` to some big number like 10, and see that we end up making a few calls for notifications as soon as we load the page. The behavior when the height is "normal" should remain the same.

Tested on janky wifi so pretty sure this works from some console logs but not 100% certain.

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -365,6 +365,7 @@ function ListImpl<ItemT>(
             root={containWeb ? nativeRef : null}
             onVisibleChange={onTailVisibilityChange}
             bottomMargin={(onEndReachedThreshold ?? 0) * 100 + '%'}
+            key={data?.length}
           />
         )}
         {footerComponent}
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -63,7 +63,53 @@ function ListImpl<ItemT>(
   }: ListProps<ItemT>,
   ref: React.Ref<ListMethods>,
 ) {
-  const [prevDataLength, setPrevDataLength] = React.useState(data?.length)
+  const contextScrollHandlers = useScrollHandlers()
+  const pal = usePalette('default')
+  const {isMobile} = useWebMediaQueries()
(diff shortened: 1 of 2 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -63,53 +63,7 @@ function ListImpl<ItemT>(
...
-      emptyComponent = <ListEmptyComponent />
-    }
-  }
-
-  if (headerOffset != null) {
-    style = addStyle(style, {
-      paddingTop: headerOffset,
-    })
-  }
+  const [prevDataLength, setPrevDataLength] = React.useState(data?.length)
```
pfrazee, line 66: Should this be a ref?

pfrazee (approved):
Tested on Web and works great, one question about the ref -- might reduce a needless rerender

src/view/com/util/List.web.tsx
```diff
@@ -184,6 +138,77 @@ function ListImpl<ItemT>(
...
+    const scrollPos = node.scrollY
+    const thresholdPx = onEndReachedThreshold * scrollViewHeight
+
+    if (scrollContainerHeight - thresholdPx <= scrollPos) {
+      onEndReached({distanceFromEnd: 0})
+    }
+    setPrevDataLength(data?.length)
+  }
+  if (data?.length !== prevDataLength) {
+    callOnEndReachedIfNeeded()
```
gaearon, line 161: There are side effects in it, so this isn't the right place to call it.

PR #4730, opened 2024-07-04
Fix feed feedback

gaearon:
We broke feed feedback when adding interstitials because feed list item is no longer a slice but a wrapper over it. Solving by adding a utility function that maps between the two types so that we don't break this again.

## Test Plan

Feedback is being sent again:

Screenshot 2024-07-04 at 19 49 19

Interstitials get filtered out:

Screenshot 2024-07-04 at 19 47 54


haileyok (approved):
glad we found this!

PR #4731, opened 2024-07-04
[Videos] Video player - PR #1 - basic player

mozzius:
Stacked on #4638 

Basic video player. Other stuff is stacked on it, probably don't merge until those are ready


src/view/com/util/post-embeds/VideoEmbedInner.tsx
```diff
@@ -0,0 +1,138 @@
...
+    player.pause()
+  }, [player])
+
+  const enterFullscreen = useCallback(() => {
+    if (ref.current) {
+      ref.current.enterFullscreen()
+    }
+  }, [])
+
+  useFrameCallback(() => {
```
haileyok, line 39: Are we going to be using this at all on web? I assume not with the intersection observer route it sounds like we're going to take.

I think I'm going to be using this a little in this implementation I cooked up on paper, but its fine to leave this logic here for now (or just empty this out, doesn't really matter)

src/view/com/util/post-embeds/VideoEmbedInner.web.tsx
```diff
@@ -0,0 +1,52 @@
+import React, {useEffect, useRef} from 'react'
+import Hls from 'hls.js'
+
+import {atoms as a} from '#/alf'
+
+export const VideoEmbedInner = ({source}: {source: string}) => {
```
haileyok, line 6: not reviewing this because i think we address it on part 2, correct me if wrong tho

haileyok (approved):
Quick glance this looks good. Gonna be tweaking it over time, so didn't super thoroughly test it.

src/view/com/posts/FeedItem.tsx
```diff
@@ -348,6 +349,7 @@ let FeedItemInner = ({
             postAuthor={post.author}
             onOpenEmbed={onOpenEmbed}
           />
+          <VideoEmbed source="https://vid.jazco.dev/watch/did:plc:q6gjnaw2blty4crticxkmujt/QmVdiZ6Bw6rTD98NTJzFWdmCTnWgiVcR7qxt127X5Km1JG/playlist.m3u8" />
```
haileyok, line 352: remember to remove this line before merge

src/view/com/util/post-embeds/VideoEmbedInner.web.tsx
mozzius: Correct!

src/view/com/posts/FeedItem.tsx
haileyok: or gate :)

src/view/com/util/post-embeds/VideoEmbedInner.tsx
mozzius: Nope, impl on web will be completely different

efstathiosntonas:
@mozzius did you encounter this on component unmount while testing it? 

```
[Reanimated] The view with tag XXXX returned an invalid measurement response. Please make sure the view is currently rendered.
```

I tried to use the `useFrameCallback` on my app and I've encountered the warning above.

While using this though the error went away:

```tsx
    const checkInView = useCallback(() => {
      'worklet';
      if (!aref) return;

      const measurement = measure(aref);
      if (measurement === null) {
        return;
      }
      if (
        measurement.pageY >= 0 &&
        measurement.pageY + measurement.height <= windowHeight
      ) {
        if (hasLeftView.value) {
          runOnJS(onEnterView)();
          hasLeftView.value = false;
        }
      } else {
        if (!hasLeftView.value) {
          runOnJS(onLeaveView)();
          hasLeftView.value = true;
        }
      }
    }, [aref, hasLeftView, onEnterView, onLeaveView, windowHeight]);

    useEffect(() => {
      const interval = setInterval(() => {
        runOnUI(checkInView)();
      }, 500);

      return () => clearInterval(interval);
    }, [checkInView]);
```

I believe this is ligther too since it calls every `500ms` instead on every single frame.

haileyok:
@efstathiosntonas We're not actually using that anyway in favor of a better solution (thankfully!) There is some context on why the interval trick doesn't work in all cases and why `useFrameCallback` is the hack we landed on (and have been using for a while) in https://github.com/bluesky-social/social-app/blob/c75bb65bef1671e493f10e06b51ee4d0cda98d83/src/view/com/util/post-embeds/ExternalPlayerEmbed.tsx#L139

If it's useful I don't mind sharing, gotta dig up the PR where I added it.

PR #4734, opened 2024-07-04
Fix order of checks in experiment

gaearon:
Same logic, but we're checking `gate()` only when it's actually gonna change the outcome.

## Test Plan

Careful inspection. It's just a bunch of booleans. (Famous last words)

Also try logging out.


estrattonbailey (approved):
Ok yep makes sense to me, haven't tested though

PR #4738, opened 2024-07-05
Show feedback for Follow button in interstitials

gaearon:
- Fix button feedback by adding another cache traversal for the shadow
- Add a toast

Note the toast is added to a shared component that's also used by the starter packs, so individual Follow buttons also show toasts there too.

## Test Plan

https://github.com/bluesky-social/social-app/assets/810438/de7ed0c5-f008-4d97-9b38-756b9aaf2168


pfrazee (approved):
LGTM

PR #4741, opened 2024-07-06
[Video] Visibility detection view

haileyok:


mozzius (commented):
Can't run on iOS yet, but for android:

1. looks good! feels smooth
2. It doesn't seem to autoplay initially
3. I feel like the amount it goes offscreen before switching could do with tweaking

mozzius (approved):
as is tradition: 💎

PR #4742, opened 2024-07-06
Update German localization (part 1)

surfdude29:
Supersedes #3664

Following the same process as #2917 (thanks [DeepL](https://www.deepl.com/translator)! 🙏) I have updated some of the German translations in this PR. I also made a few tweaks to existing strings, but of course let me know if I got something wrong there.

There are currently 715 missing strings in the German message catalog, but I thought  that was too many to tackle properly in one PR. So I've done the first 200+ and I'll move on to the rest in later PRs.

I signed up for a trial of DeepL Pro which allows you to specify an informal tone, so the translated strings in this PR should at least follow that.

I incorporated the corrections from @PythooonUser made in commit https://github.com/bluesky-social/social-app/pull/3664/commits/659bc5e78d0582551afb97cfc5d94a84063fc91a but otherwise I thought it was easiest to start from scratch with this PR.

All corrections from native German speakers are welcome, especially previous translators including @PythooonUser, @cdfzo and @imbstt.

Thanks! :)

EDIT: I've also merged in some updates to existing translations, thanks again @cdfzo! 🙏

PR #4743, opened 2024-07-06
Add missing `to` in StarterPackScreen.tsx string

surfdude29:
@cdfzo pointed out in a [review comment](https://github.com/bluesky-social/social-app/pull/4742/files#r1667428796) that there's a missing `to` in the string that the user is shown when they're deleting a starter pack. This PR adds it.


haileyok (approved):
Thanks!

PR #4744, opened 2024-07-07
Remove broken and void back button

oracularhades:
https://github.com/bluesky-social/social-app/assets/91714073/5b0f60a0-1654-4e8f-8596-fbe400093e67

This back button shouldn't be here. This button isn't programmed to go back, but instead updates "setActiveStarterPack(undefined)". Which causes a permanent 'processing' state.

However, this button causes an infinite loading spin and sucks for new users, especially for mobile users. Also, Because it's the same URL, browsers do not go back, my immediate instinct upon getting this infinite spin was to go back a page, but instead it will send the user to a new tab or what they were looking at, so they'd need to go back to their chat app (or whatever) and re-click the link. Horrifically bad for new user sign-ups. Because of browser caching, going back and forward (at least in FireFox) will display the infinite loading state again.

The starter-pack has different UIs for logged-in and logged-out view, thus we can just remove the entire button, as we don't need to support logged-in context.

(unrelated, but this is how this all started lol) https://bsky.app/profile/oracularhades.com/post/3kwnui4dg6f2z

🦋

### DIFF
--- a/src/screens/StarterPack/StarterPackLandingScreen.tsx
+++ b/src/screens/StarterPack/StarterPackLandingScreen.tsx
@@ -31,7 +31,6 @@ import {atoms as a, useTheme} from '#/alf'
 import {Button, ButtonText} from '#/components/Button'
 import {useDialogControl} from '#/components/Dialog'
 import * as FeedCard from '#/components/FeedCard'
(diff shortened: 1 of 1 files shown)


haileyok:
Nice, yea we don't need this now that we improved how tapping a SP link in the signed-out PWI works. Thanks for the find.

haileyok (approved):

PR #4752, opened 2024-07-08
change `contentVisibility` to `contain`

haileyok:
## Why

We've ran into another weird quirk when using `content-visiblity`. Scroll views often don't resize whenever new items get added to the list, which is causing problems with `onEndReached` in the web implementation of `FlatList`. We're switching to `contain: layout paint` instead.

## Test Plan

When opening the notifications tab on a large monitor, there shouldn't be chained requests for notifications.


gaearon (approved):

gaearon:
we should rename the property maybe?

haileyok:
Sure. I also changed `containWeb` to the more describe `disableFullWindow` with a little comment on usage so that it isn't redundant with this new prop. Checking real quick that it doesn't break anything.

PR #4753, opened 2024-07-08
Remove client filtering of starter packs

haileyok:
Removing the client filter once https://github.com/bluesky-social/atproto/pull/2629 gets merged and deployed.


mozzius (approved):
stamping that

PR #4754, opened 2024-07-08
[Video] Uploads

haileyok:


patches/expo-modules-core+1.12.11.patch
```diff
@@ -12,3 +12,15 @@ index bb74e80..0aa0202 100644
...
+--- a/node_modules/expo-modules-core/build/uuid/uuid.js
++++ b/node_modules/expo-modules-core/build/uuid/uuid.js
+@@ -1,5 +1,7 @@
+ import bytesToUuid from './lib/bytesToUuid';
+ import { Uuidv5Namespace } from './uuid.types';
++import { ensureNativeModulesAreInstalled } from '../ensureNativeModulesAreInstalled';
++ensureNativeModulesAreInstalled();
+ const nativeUuidv4 = globalThis?.expo?.uuidv4;
+ const nativeUuidv5 = globalThis?.expo?.uuidv5;
+ function uuidv4() {
```
haileyok, line 26: Temporary patch for now, see https://bluesky-builders.slack.com/archives/C05R8KJK1PC/p1720682497841229?thread_ts=1720651555.643739&cid=C05R8KJK1PC

src/state/queries/video/video-upload.web.ts
```diff
@@ -0,0 +1,61 @@
...
+  return useMutation({
+    mutationFn: async (video: CompressedVideo) => {
+      const uri = createVideoEndpointUrl('/upload', {
+        did: currentAccount!.did,
+        name: `hailey-${nanoid(12)}.mp4`,
+      })
+
+      const bytes = await fetch(video.uri).then(res => res.arrayBuffer())
+
+      const xhr = new XMLHttpRequest()
```
haileyok, line 30: Using XHR here because `ReadableStream` (which we'd need to use `fetch` here) is fairly new and might not be supported on some older browsers.

mozzius (approved):
Solid work! UI we can iterate on over time, let's get it in

PR #4757, opened 2024-07-09
Style labels like those old 80x15 web badges

rolandcrosby:
A modest proposal inspired by [these things](https://web.badges.world/)

Screenshot 2024-07-09 at 12 51 09 AM


pfrazee:
Heh. That's pretty incredible. We all got a chuckle.





but no

PR #4758, opened 2024-07-09
Log session events unconditionally

gaearon:
Adds a log when an agent emits an event (`update`, `expired`, and a few others). We want to get a baseline on how much expiration is occurring in practice, and we want to be able to target that with experiment fixes.

## Test Plan

Force early expiration using the short session debug header, verify it results in sending this event.

Screenshot 2024-07-09 at 20 11 02


haileyok (approved):
Code looks good, didn't test though since unsure how to set that header. I do see the metadata being added to the statsig event, just not sure how to trigger the `expired` event.

PR #4759, opened 2024-07-09
bskyweb: optional basic auth password middleware

bnewbold:
If this env var is configured, the golang web service will require HTTP Basic auth for access to the web interface.

Tested locally and this seems to work and not conflict with actual service auth. Not sure that will remain true in the future with OAuth, etc.

Not enabled by default, so should not impact prod deployment.


estrattonbailey (approved):
LGTM, render preview works

PR #4760, opened 2024-07-09
bskyweb: add robots disallow and 'canonical' config flags

bnewbold:
NOTE: it is critical to enable the "canonical" flag for our prod deployment before merging/deploying this branch.

Adds two config/envvar flags:

- one to allow swapping in a blank "Disallow: *" robots.txt to prevent crawling (eg, in dev/staging/informal deployments)
- another to only serve some endpoints in the "canonical" deployment (eg, bsky.app)

Normally we might try to detect what "environment" we are in and change these behaviors based on that, but we have kind of a sprawling menagerie of environments, and folks also might be forking this repo for their own use, so made these more granular flags. Open to changing path and doing this the traditional `ENVIRONMENT` variable way though.

### DIFF
--- a/bskyweb/cmd/bskyweb/main.go
+++ b/bskyweb/cmd/bskyweb/main.go
@@ -94,6 +94,20 @@ func run(args []string) {
 					Value:    "",
 					EnvVars:  []string{"STATIC_CDN_HOST"},
 				},
+				&cli.BoolFlag{
+					Name:     "bsky-canonical-instance",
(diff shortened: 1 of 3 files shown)


gaearon (approved):
lgtm but i don't go

haileyok (approved):
lgtm. do we not want to enable any of these before we merge/deploy? (i assume maybe from infra repo). as it stands i think we'd lose the routes you've put behind the flag now.

PR #4762, opened 2024-07-09
[Session] Experiment: Don't use withProxy

gaearon:
There's a theory that some of the logouts may be due to a bug in how `withProxy` interacts with the expiry logic (a cloned agent may expire and fork the token). I'm adding a gate that forks the logic to not use `withProxy`. We _could_ do this without the gate (the behavior should be equivalent) but I want to see if this actually reduces the number of `expired` events or not.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/4762/files?w=1)

## How

This is basically what `withProxy` does at the lower level:

https://github.com/bluesky-social/atproto/blob/7761a463b221717e5923dff77b1cc4c3755f3fd7/packages/api/src/agent.ts#L95-L99

which means this:

https://github.com/bluesky-social/atproto/blob/7761a463b221717e5923dff77b1cc4c3755f3fd7/packages/api/src/agent.ts#L128-L133

which means this:

https://github.com/bluesky-social/atproto/blob/7761a463b221717e5923dff77b1cc4c3755f3fd7/packages/api/src/agent.ts#L263-L268

So I'm doing the same but manually.

## Test Plan

Here's the feed feedback request with the gate off:

Screenshot 2024-07-09 at 23 17 50

Same request with the gate on:

Screenshot 2024-07-09 at 23 49 05

Also verified I can still appeal labels (200 OK):

Screenshot 2024-07-09 at 23 55 16

Also still able to submit reports (200 OK):

Screenshot 2024-07-10 at 00 00 10

Also checked these with the gate off.

### DIFF
--- a/src/components/ReportDialog/SubmitView.tsx
+++ b/src/components/ReportDialog/SubmitView.tsx
@@ -6,6 +6,7 @@ import {useLingui} from '@lingui/react'
 
 import {getLabelingServiceTitle} from '#/lib/moderation'
 import {ReportOption} from '#/lib/moderation/useReportOptions'
+import {useGate} from '#/lib/statsig/statsig'
 import {useAgent} from '#/state/session'
 import {CharProgress} from '#/view/com/composer/char-progress/CharProgress'
 import * as Toast from '#/view/com/util/Toast'
@@ -36,6 +37,7 @@ export function SubmitView({
   const t = useTheme()
   const {_} = useLingui()
   const agent = useAgent()
+  const gate = useGate()
   const [details, setDetails] = React.useState<string>('')
   const [submitting, setSubmitting] = React.useState<boolean>(false)
   const [selectedServices, setSelectedServices] = React.useState<string[]>([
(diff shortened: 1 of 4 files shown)


src/components/ReportDialog/SubmitView.tsx
```diff
@@ -60,15 +62,29 @@ export function SubmitView({
...
-          .then(
-            _ => true,
-            _ => false,
-          ),
-      ),
+      selectedServices.map(did => {
+        if (gate('session_withproxy_fix')) {
+          return agent
+            .createModerationReport(report, {
+              encoding: 'application/json',
```
gaearon, line 69: TS makes this required but this is the default (and only allowed) value

haileyok (approved):
Tested both on/off for appeal and create. lgtm, this is an interesting one!

src/components/ReportDialog/SubmitView.tsx
matthieusieben: This will be fixed in the next sdk release ;-)

PR #4767, opened 2024-07-10
ALF post button

haileyok:
## Why

I'm adding disabled state to the `Post` button inside of the video PR. Right now we just have a regular `TouchableOpacity` with a `LinearGradient` inside for the post button. This makes a disabled state pretty weird and is different than the rest of our button styles in the app now. Let's change it to be an ALF button instead.

## Test Plan

Just see how it looks and feels

RocketSim_Recording_iPhone_15_Pro_6 1_2024-07-10_12 37 27

https://github.com/bluesky-social/social-app/assets/153161762/653bbb03-8bc6-41c1-ae2a-acde9bf80e99

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -462,7 +462,11 @@ export const ComposePost = observer(function ComposePost({
               color="primary"
               shape="default"
               size="small"
-              style={[a.rounded_full, a.py_sm]}
+              style={[
+                a.rounded_full,
+                a.py_sm,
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
LGTM, didnt test

mozzius:
Text is no longer left aligned with the icon, maybe shift it over a bit? Idk if that'd then mean the background would go off the side of the screen when pressed

haileyok:
@mozzius mm yea didn't even think of that. i tweaked it a bit to make it look better and line up, ty!!

PR #4768, opened 2024-07-11
Lint native files

haileyok:
## Why

Finally got tired of Xcode's non-existent formatting and Android Studio's far-too-lax formatting rules.

## Test Plan

It builds. No real changes aside from formatting.


gaearon (approved):

package.json
```diff
@@ -29,6 +29,8 @@
     "test-ci": "NODE_ENV=test jest --ci --forceExit --reporters=default --reporters=jest-junit",
     "test-coverage": "NODE_ENV=test jest --coverage",
     "lint": "eslint --cache --ext .js,.jsx,.ts,.tsx src",
+    "lint-native": "swiftlint ./modules && ktlint ./modules",
```
gaearon, line 32: should this be added to the CI run?

haileyok: i was going to, but it wasn't immediately obvious how to get it working on ubuntu (saw something about docker 😦)

would like to do that though at some point, idk if we'd need to use a macos runner to get it to run or not though, so i'll figure it out later. isn't extremely vital now since this only deals with formatting anyway and errors can't cause problems.

PR #4769, opened 2024-07-11
[ALF] Theme & palette cleanup

estrattonbailey:
ALFs theming defines colors based on a scale, from `25` to `975`, where `25` is low in contrast to the application background color, and `975` is high in contrast. In light mode, this means the scale moves from light->dark, and conversely for dark mode.

However, only the grayscale behaved this way for all themes. The `primary`, `negative` and `positive` color ranges remained as they were defined for light mode. This meant that a dozen or so places in the app needed to do manual checks in order to select the correct color e.g. `t.name === 'light' ? t.palette.primary_25 : t.palette.primary_975`, where both color stops are equivalent relative to their theme.

This PR aligns the color scales so that that snippet above can be simply `t.palette.primary_25` and that `25` value the low-contrast value we expect in all color modes.

While doing this, I felt that it was a good idea to move theme generation into context so it can respond to state more readily, instead of having fixed themes defined ahead of time. We aren't utilizing this yet, but perhaps we will in the future.

The majority of this PR is moving things around, and there should be no noticeable changes.

I made a few subtle changes in the following files to avoid the manual `t.name` checks:
- `forms/Toggle` (compare using our Storybook)
- `Button` (Storybook)
- `TextField` (StoryBook)
- `MessageItem`
- `ThreadgateEditor`
- `IconCircle`
- `MutedWords`

The only noticeable change is in `Button` where I think I actually nicely improved the contrast of the solid/secondary button on dark/dim modes. Buttons are important so here's a detail, before->after:
buttons


haileyok (approved):
Clicked through the different themes in the storybook and through the app, only found one spot where we need to change something i think.

Thanks for doing this, big improvement!

### Unread notification:

Screenshot 2024-07-11 at 2 07 48 PM

### Feedback btn:

This isn't new actually, but wonder if we can just change this real quick.

Screenshot 2024-07-11 at 2 08 39 PM

estrattonbailey:
Gonna do a separately PR for the button bit

PR #4771, opened 2024-07-11
Remove invalid labelers when subscribing/unsubscribing

estrattonbailey:
Users who have subscribed to labelers who have subsequently shut down (or were taken down) cannot unsubscribe from them. Those values are stuck in their prefs. If the user is at the labeler limit, they also can't subscribe to new labelers.

Since these are user prefs and live on the PDS, and because we don't validate this endpoint, we can't very well do this on the server. So we need to do it here.

This PR checks for any invalid labelers and removes them every time a user subscribes or unsubscribes to a labeler. When subscribing, if the user is still at the limit after labeler validation, we throw an error and handle it by showing the informational dialog.

Note: this also means that the button to Subscribe to a labeler remains active, **even if the user might be at the limit already.**

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderLabeler.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderLabeler.tsx
@@ -10,6 +10,7 @@ import {
 import {msg, Plural, plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 import {MAX_LABELERS} from '#/lib/constants'
 import {isAppLabeler} from '#/lib/moderation'
 import {logger} from '#/logger'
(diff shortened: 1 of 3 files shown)


src/screens/Profile/Header/ProfileHeaderLabeler.tsx
```diff
@@ -328,8 +326,8 @@ function CantSubscribePrompt({
       <Prompt.TitleText>Unable to subscribe</Prompt.TitleText>
       <Prompt.DescriptionText>
         <Trans>
-          We're sorry! You can only subscribe to twenty labelers, and you've
-          reached your limit of twenty.
+          We're sorry! You can only subscribe to {MAX_LABELERS} labelers, and
+          you've reached your limit of {MAX_LABELERS}.
```
pfrazee, line 330: Is this going to jam up the translators?

pfrazee (approved):
Yeah seems reasonable, one small concern

src/screens/Profile/Header/ProfileHeaderLabeler.tsx
estrattonbailey: Yeah it might. I linked `MAX_LABELERS` in a comment instead, just wanted to keep a relationship there to avoid forgetting to update one or the other.

PR #4777, opened 2024-07-12
Return a promise for `getReferrerInfoAsync` on web

haileyok:
## Why

Forgot to actually return a promise here. We use a promise because that matches the way it works on native.

## Test Plan

The app will actually load now (main.bsky.dev ftw, thanks @mary-ext for catching it really fast! ❤️)


modules/expo-bluesky-swiss-army/src/Referrer/index.web.ts
```diff
@@ -8,27 +8,29 @@ export function getGooglePlayReferrerInfoAsync(): Promise<GooglePlayReferrerInfo
 }
 
 export function getReferrerInfoAsync(): Promise<ReferrerInfo | null> {
```
gaearon, line 10: alternatively i think we can mark this as async?

gaearon (approved):
ok but i think making the function `async` is simpler

haileyok:
lol...yea long day 🙃

PR #4662, opened 2024-06-27
Update French localization

surfdude29:
As a follow-up to #4611 and #4637 this PR adds translations for new strings, especially some high-visibility starter pack-related ones. It also includes one correction where I had missed changing `Packs` to `Kits`.

I'd be grateful if you could take a look @Signez 🙏

PR #4781, opened 2024-07-12
Update French localization

surfdude29:
Following on from #4662, this PR adds translations for new strings and makes a tweak to an existing string.

I'd be grateful if you could take a look @Signez 🙏


src/locale/locales/fr/messages.po
```diff
@@ -5030,6 +5049,10 @@ msgstr "Sélectionnez le(s) service(s) de modération destinataires du signaleme
 msgid "Select the service that hosts your data."
 msgstr "Sélectionnez le service qui héberge vos données."
 
+#: src/view/com/composer/videos/SelectVideoBtn.tsx:53
+msgid "Select video"
+msgstr "Sélectionne une vidéo"
```
Signez, line 5054: It's a label, so I would argue that it is, in fact, an infinitive we should use here.

```suggestion
msgstr "Sélectionner une vidéo"
```

Signez (commented):
Great work as usual — and good catch of the missing `d’actu` (even though it wasn’t actually a problem, as it could have been understood thanks to the context).

src/locale/locales/fr/messages.po
surfdude29: Oops, yes that makes sense 👍

surfdude29:
Correction applied, thanks yet again @Signez! :)

pfrazee (approved):
Thanks all!

PR #4782, opened 2024-07-12
Make all referrer info sync

haileyok:
We wanted to make this sync, but in some merges there were some mixups. This corrects those.


mozzius (approved):
Not tested but makes sense

PR #4784, opened 2024-07-12
Deployment of BlueSky using AWS as the cloud service provider.

surcodev:
The proposal is to create a clear and concise guide for deploying BlueSky on AWS, using GitHub Actions for continuous integration, ECR for Docker image management, and EC2 for hosting instances. This initiative will facilitate efficient use of AWS for the community, improving the deployment, management, and scalability of BlueSky.

### DIFF
--- a/docs/IaC deploy/Dockerfile
+++ b/docs/IaC deploy/Dockerfile
@@ -1,0 +1,8 @@
+# Usa la imagen base 
+FROM public.ecr.aws/t8b2r8w9/social-app:latest
+
+# Expone el puerto 8100 que es el puerto en el que el servidor escucha
+EXPOSE 8100
+
+# Comando que se ejecutará al iniciar el contenedor
+CMD ["/usr/bin/bskyweb", "serve"]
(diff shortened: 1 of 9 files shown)


bnewbold:
Hi! This seems like a great contribution, thanks for writing this up.

To date, we have mostly not included deployment scripts or instructions in our project repositories. We do include dev env setup scripts, build scripts, and container build scripts/instruction. But then we put self-hosting and deployment stuff in separate "distribution" repos. An example is the PDS distro: https://github.com/bluesky-social/pds has the deployment/self-hosting info, while the code mostly in the https://github.com/bluesky-social/pds repo.

Doesn't have to stay that way! But that is the pattern so far.

I'd encourage you to toss up a new github repo for these terraform/AWS scripts, and we can direct folks interested to that. My guess is that different folks will want different tools (kubernetes, nomad, just docker, systemd, nix, etc), and we'll end up pointing folks to multiple different sets of self-hosting/deployment guides over time.

PR #4786, opened 2024-07-12
Agrega switch para activar el uso de la IA

javierPerez900:


oracularhades:
This commit is spam.

#1, under src/view/screens/Search/AIModels.ts, for some reason, you have leaked your own API key, but you're also trying to make the social-app depend on Facebook/Meta's AI API and send queries to it. You have also included some nonsense prompt.
#2, you have changed line counts in the translation file, without actually changing the content of the files referenced.
#3, you've changed 28 files in 1 PR, with many changes not related to each other, and you have code comments in Spanish.

haileyok:
Thanks, but unsure of what the purpose of this is and we definitely won't be adding any models unnecessarily like this.

For the record, you did leave a huggingface token in this PR, so you probably will want to revoke that.

PR #4787, opened 2024-07-12
Add verification of known emails to the Register section

BrayanVall21:
A possible improvement focused on the Bluesky registration process was identified. Currently, the platform does not present any validation for the emails admitted during the registration section; and by mistake, it is possible that a user does not place the emails correctly. For this reason, a list of the most common email addresses is added that will help the user verify whether or not they made a mistake.

### DIFF
--- a/src/screens/Signup/StepInfo/Policies.tsx
+++ b/src/screens/Signup/StepInfo/Policies.tsx
@@ -13,10 +13,12 @@ export const Policies = ({
   serviceDescription,
   needsGuardian,
   under13,
+  verificatorEmail,
 }: {
   serviceDescription: ComAtprotoServerDescribeServer.OutputSchema
   needsGuardian: boolean
   under13: boolean
+  verificatorEmail : boolean
 }) => {
   const t = useTheme()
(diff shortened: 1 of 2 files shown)


oracularhades:
This isn't needed. You can update your email before you verify it. It's not like you have to verify to change it and your account becomes locked.

oracularhades:
Also. Your email "validation" restricts all possible email domains to 3 (Microsoft, Google, Yahoo), which is insane.

mary-ext:
i've no idea where you're getting this list of email provider domains from but it's certainly not very up to date and i'm doubting the legitimacy of some of them

oracularhades:
You're also removing basic validation checks and messing up library imports.

haileyok:
Thanks, but this isn't something that we're going to do, even with a list of valid domain names. For example, this would restrict me from signing up with a work email address, which is obviously legitimate but wouldn't pass this check.

The only check I think we could reasonably add here is to verify the user enters a valid TLD. Otherwise, email verification itself would need to take place (for example to address possible typos) outside of the client itself.

Issue #4775, opened 2024-07-12
Sign out for each signed in account


xtianpu:
**Is your feature request related to a problem? Please describe.**

It was inconvenient when signing out is signing out all of signed-in accounts

**Describe the solution you'd like**

Separation of sign out button for each signed in account

**Describe alternatives you've considered**

Signing out, then sign in again to specific account I wanted

gaearon:
not quite like this issue described, but https://github.com/bluesky-social/social-app/pull/4941 fixes this.

Issue #4683, opened 2024-06-27
Signing out an account is signing out all accounts


xtianpu:
**Describe the bug**
Signing out an account is signing out all accounts

**To Reproduce**

1. Press sign out button on settings

**Expected behavior**

Should be only one account to be signed out, and switch to another signed in account

**Details**

 - Platform: All platform
 - Platform version: -
 - App version: v1.87.0

swar00p:
Is this the expected behavior? I tried it on my iOS app and confirmed that I get signed out of both my accounts if I click on Sign Out on one of them. I'm a complete newbie but I cloned the code base on my laptop and took a quick look. I may be mistaken but it looks like the logout method in https://github.com/bluesky-social/social-app/blob/ea0586cd67427ce68f867ee25e03bb92169f23c3/src/state/session/index.tsx#L117 doesn't accept an account parameter.

I'm happy to dig in a bit more to see what's going on.

haileyok:
@swar00p Definitely is not expected behavior. Feel free to do some digging on it if you feel like it! ❤️

xtianpu:
@swar00p @haileyok oh sorry, this should be feature request, my bad assigned it as label bug. Should I close this and create a new one?

haileyok:
I think bug is fine - it feels like a bug to me 🙃

xtianpu:
Just create a new one with feature request label https://github.com/bluesky-social/social-app/issues/4775

swar00p:
> Just create a new one with feature request label #4775

@xtianpu I think @haileyok said that this should remain open. There was no need to create a feature request. Can you please close the feature request and reopen this bug?

swar00p:
@haileyok I think I was wrong. The way it's been implemented in https://github.com/bluesky-social/social-app/blob/7c1c24ef1badb746ecbe707d390bec4934940ced/src/state/session/reducer.ts#L145 and judging by the comment, it looks like it was deliberately implemented this way (to sign out of all accounts). Who makes the decision (one way or another) whether it should continue to work this way or if it should be implemented as a sign out for the selected account only?

haileyok:
@swar00p Yea, that is the way it was implemented, though we are fine with changing it. It's something that we'll change at some point, but do feel free to look into it yourself and PR if you want to! Also can just take the discussion to the other issue (either label really works here, it _does_ seem like a bug and I personally would consider it one, but I digress 😆)

haileyok:
[...]

PR #4788, opened 2024-07-13
 User is signed out of a selected account rather than all accounts

swar00p:
- Added the relevant account ID when dispatching the logout event (session/index.tsx)
- Consumed the account ID in the reducer and cleared the tokens for the specific account.

- Feature Request: https://github.com/bluesky-social/social-app/issues/4775
- Bug: https://github.com/bluesky-social/social-app/issues/4683

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -120,6 +120,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       cancelPendingTask()
       dispatch({
         type: 'logged-out',
+        accountDid: account.did,
       })
       logEvent('account:loggedOut', {logContext})
       addSessionDebugLog({type: 'method:end', method: 'logout'})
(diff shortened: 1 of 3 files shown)


swar00p:
Made necessary changes. The app should build now. I was able to build it locally and run it but the sign out functionality still seems to behave in the old way. I'm not sure if I'm missing some piece of the logic in some other part of the code base.

kulak91:
@swar00p I'm not familiar with this codebase either, but as far as i can see that you are only accepting `account` into the useCallback function, but i can't see any changes where do you actually invoke logout function (it still invokes with the context string only) 
image

Was your `loggedOutAccount` ever __not__ undefined?

gaearon:
thanks, we did a more extensive fix here https://github.com/bluesky-social/social-app/pull/4941

PR #4789, opened 2024-07-13
Add logging of selected feed preference when displaying the following feed

haileyok:
## Why

We want to do a little logging to ensure we don't have a bug in following feed preferences. There have been some reports that maybe it isn't working correctly, so let's test it out and see.

## How

We sample the feed displayed event now, so for the time being while we start debugging this, let's change that to always log these displays instead of sampling them.

## Test Plan

Whenever opening the app and displaying the following feed, we should be sending `feedDisplayed` events now instead of `feedDisplayed:sampled`. However, for all other feeds, we should still be using `feedDisplayed:sampled`.

There should also be two additional fields in the metadata now, showing what the preference is and how many likes are needed to actually display a reply in the following feed.

Screenshot 2024-07-13 at 10 55 56 AM

Screenshot 2024-07-13 at 10 57 47 AM

Screenshot 2024-07-13 at 10 58 15 AM

### DIFF
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -75,19 +75,6 @@ export type LogEvents = {
   'onboarding:finished:avatarResult': {
     avatarResult: 'default' | 'created' | 'uploaded'
   }
-  'home:feedDisplayed': {
-    feedUrl: string
-    feedType: string
-    index: number
-    followingShowRepliesFromPref: 'all' | 'following' | 'off'
-    followingRepliesMinLikePref: number
-    reason:
-      | 'focus'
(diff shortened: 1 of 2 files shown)


gaearon:
how do you feel about adding a separate event that’s only sent once? we sample these because they’re sent too many times. but realistically we don’t expect it to change within a session, plus we really only want to know the distribution of values by unique user rather than by unique feed display.

haileyok:
@gaearon changed now

Screenshot 2024-07-14 at 9 01 18 AM

Screenshot 2024-07-14 at 9 03 09 AM

gaearon (approved):

PR #4790, opened 2024-07-13
Only sign out of a single account instead of all

haileyok:
## Why

Updating https://github.com/bluesky-social/social-app/pull/4788

Right now, whenever a user logs out of an account, it signs them out of all their accounts rather than just the one they press "sign out" on. This means that even other accounts that still have a valid refresh token require a password to sign in again.

## Test Plan

Have not tested myself yet, just fixing some of the logic right now.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -120,6 +120,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       cancelPendingTask()
       dispatch({
         type: 'logged-out',
+        accountDid: account.did,
       })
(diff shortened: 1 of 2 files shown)


swar00p:
I'd like to revert this PR. It looks like I didn't fully understand the code base and the changes that were necessary and that's my bad. I'll do a better job next time around and set up the local environment for running lint & the test suite. I saw some instructions to do it but I was a bit impatient to submit the PR!

haileyok:
Sure, that's fine.

PR #4798, opened 2024-07-17
Priority notifications

mozzius:
Adds a new page, `/notifications/settings` with a toggle to turn priority notifications on. ~Backend is not yet deployed at time of writing~ It's ready to go!

Note on behaviour - we get what the current priority setting is from `listNotifications`. If we call this immediately after, I was running into read-after-write issues, so it attempts it a few times before giving feedback to the user on whether it has succeeded or not. It disables the input while waiting.

It shows a spinner while waiting for the initial value, if it's not already cached (which it should be 9/10 times, since it reuses the query from the main page and you're likely coming from that page)

Screenshot 2024-07-19 at 15 02 07

Screenshot 2024-07-24 at 16 08 58


src/view/screens/NotificationsSettings.tsx
```diff
@@ -0,0 +1,140 @@
...
+      />
+      <View style={[a.p_lg, a.gap_md]}>
+        <Text style={[a.text_lg, a.font_bold]}>
+          <FontAwesomeIcon icon="flask" style={t.atoms.text} />{' '}
+          <Trans>Notification filters</Trans>
+        </Text>
+        <Toggle.Group
+          label={_(msg`Priority notifications`)}
+          type="checkbox"
+          values={priority ? ['enabled'] : []}
```
haileyok, line 82: This doesn't revert back if the request fails which is a little weird.

src/view/screens/NotificationsSettings.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const priority = data?.pages.at(0)?.priority
+
+  const {_} = useLingui()
+  const t = useTheme()
+
+  const {mutate: onChangePriority, isPending} = useMutation({
+    mutationFn: async (keys: string[]) => {
+      const enabled = keys[0] === 'enabled'
+
+      eagerlySetCachedPriority(queryClient, enabled)
```
haileyok, line 39: I assume we should be doing the inverse of this in `onError`. A refetch might not return the correct state since a network error could have caused the failure.

haileyok:
Left two notes, but with a minimal test this works well.

I'm still concerned about one thing especially after using this but unrelated to the PR, I'll flag it in Slack.

src/view/screens/NotificationsSettings.tsx
```diff
@@ -0,0 +1,145 @@
...
+            a.px_xl,
+            a.py_lg,
+            a.rounded_md,
+            t.atoms.bg_contrast_25,
+          ]}>
+          <Text style={[t.atoms.text_contrast_high, a.leading_snug]}>
+            <Trans>
+              Experimental: Filter your notifications. Currently this means
+              you'll only receive notifications from users you follow — we'll
+              refine this over time.
```
haileyok, line 120:
```suggestion
              Experimental: When this preference is enabled, you'll only
              receive reply and quote notifications from users you follow.
              We'll continue to add more controls here over time.
```

haileyok (approved):
Tested a bunch of different states

- Intermitent network
- Going back before preference updates
- Slow network and spam change

Things seem really good here. Weighed pros/cons of that mutation, and for now this seems okay, since it's already a little bizarre how we're getting the preference. I think we will want to start returning a preference object whenever we add more preferences here (like granular filters).

Gonna have to build for TF to test since OTA is borked. Great work @mozzius!!! 🥳

haileyok (approved):

Issue #3311, opened 2024-03-21
Finnish, French, German, Portuguese and Ukrainian text in Muted words & tags dialog overflows on web


surfdude29:
**Describe the bug**

In the radio buttons (are they still called that?) in the Muted words & tags dialog, the text overflows on the web when the app language is set to Finnish, French, German, Portuguese or Ukrainian.

**To Reproduce**

1. Set App Language to Finnish, French, German, Portuguese or Ukrainian in Settings.
2. Open the Moderation > Muted words & tags dialog.
3. Observe that the text in the radio buttons overflows and often hides the icons.

**Expected behavior**

All of the text is readable and the icons are visible, perhaps with longer text wrapping on to the next line if that's possible?

**Screenshots**

Finnish:

![IMG_7809](https://github.com/bluesky-social/social-app/assets/149612116/9c576274-f933-4833-9176-917b48178593)

French:

![IMG_7808](https://github.com/bluesky-social/social-app/assets/149612116/280cc3f5-20f1-4e12-bcda-c27a9511d37b)

German:

![IMG_7805](https://github.com/bluesky-social/social-app/assets/149612116/789168ed-e30f-4158-bb49-0314983037a1)

Portuguese:

![IMG_7810](https://github.com/bluesky-social/social-app/assets/149612116/04cba038-ee26-4731-a9ff-cc2ff6426807)

Ukrainian:

![IMG_7811](https://github.com/bluesky-social/social-app/assets/149612116/72e65a61-432b-4585-a265-ab12940384fe)

**Details**

 - Platform: Firefox on iPadOS and Brave on Linux
 - App version: 1.73

**Additional context**

As far as I can see, this isn't an issue in the iOS app (and I assume not on Android either) due to the use of the bottom sheet.

surfdude29:
This has been fixed.

PR #4801, opened 2024-07-19
Update muted words dialog with `expiresAt` and `actorTarget`

estrattonbailey:
### Relies on https://github.com/bluesky-social/atproto/pull/2276

Adds support for timed mute words, as well as targeting only those users whom you don't follow. Also integrates new handlers into the `TagMenu` (when clicking or long-pressing on a tag).

This PR also adds a new hook `useFormatDistance`, with the intent that we can add more in the pattern of `use<method>` where `method` is a `date-fns` method that accepts localization. Right now, all our app languages are supported with the exception of Gaelic. In the interest of time and stability (I'd like to use a library for this) I think this is an OK trade-off. We can help facilitate a `ga` translation in the near future.

image

i18n works:
CleanShot 2024-07-19 at 14 24 27@2x


surfdude29:
While you're working on updating the muted words dialog, might it be possible to squeeze in a fix for #3311? 🙏

This issue affects the web interface in Catalan, Finnish, French, German, Irish, Portuguese and Ukrainian.

src/components/dialogs/MutedWords.tsx
```diff
@@ -53,45 +56,68 @@ function MutedWordsInner() {
...
     } catch (e: any) {
       logger.error(`Failed to save muted word`, {message: e.message})
       setError(e.message)
     }
-  }, [_, field, options, addMutedWord, setField])
+  }, [_, field, targets, addMutedWord, setField, durations, excludeFollowing])
 
   return (
     <Dialog.ScrollableInner label={_(msg`Manage your muted words and tags`)}>
-      <View onTouchStart={Keyboard.dismiss}>
```
haileyok, line 83: tested on ios/android to make sure this is okay to remove, seems good. think using the discord fork of bottomsheet made this unnecessary.

src/components/dialogs/MutedWords.tsx
```diff
@@ -107,30 +133,130 @@ function MutedWordsInner() {
             }}
             onSubmitEditing={submit}
           />
+        </View>
 
+        <View style={[a.pb_xl, a.gap_sm]}>
           <Toggle.Group
-            label={_(msg`Toggle between muted word options.`)}
+            label={_(msg`Select how long this mute word should apply for.`)}
```
haileyok, line 140:
```suggestion
            label={_(msg`Select how long to mute this word for.`)}
```

src/components/dialogs/MutedWords.tsx
```diff
@@ -289,53 +438,77 @@ function MutedWordRow({
...
-          {word.targets.map(target => (
-            <View
-              key={target}
-              style={[a.py_xs, a.px_sm, a.rounded_sm, t.atoms.bg_contrast_100]}>
-              <Text
-                style={[a.text_xs, a.font_bold, t.atoms.text_contrast_medium]}>
-                {target === 'content' ? _(msg`text`) : _(msg`tag`)}
+              <Text style={[a.font_normal, t.atoms.text_contrast_medium]}>
+                {' '}
+                <Trans>in</Trans>{' '}
```
haileyok, line 465: Can we move `<Trans>` to wrap the entire root `<Text>` here starting on L451?

haileyok (commented):
Cooooool

estrattonbailey:
Contrived example, but text can now handle wrapping for languages with longer transations.

CleanShot 2024-07-23 at 15 17 22@2x

estrattonbailey:
Not sure how ya saw those borked views, might be without proper types/validation if you didn't have the linked API SDK built and installed. Was able to fix the label alignment though, wonder when that got bumped.

surfdude29:
> Contrived example, but text can now handle wrapping for languages with longer transations.

Excellent, thanks! :)

pfrazee (approved):
Tested on all platforms, didn't test new logic correctness but the tests in the api seemed solid

LGTM!

PR #4803, opened 2024-07-19
Use appropriate icons for toasts

mozzius:
Despite toasts being used heavily for reporting errors, we just show a default tick icon which is a bit dissonant for a popup displaying bad news. This PR goes through and uses `xmark` for negative ones, with a few as `info` or `exclamation-circle` where appropriate. I also made the clipboard related ones use a clipboard icon.

Also, I shrunk the icon a little and applied `flex-shrink: 0` so it stays a consistent size

Screenshot 2024-07-19 at 14 56 57
Screenshot 2024-07-19 at 14 56 33

### DIFF
--- a/src/components/dms/dialogs/NewChatDialog.tsx
+++ b/src/components/dms/dialogs/NewChatDialog.tsx
@@ -32,7 +32,7 @@ export function NewChat({
       logEvent('chat:open', {logContext: 'NewChatDialog'})
     },
     onError: error => {
-      logger.error('Failed to create chat', {message: error})
+      logger.error('Failed to create chat', {safeMessage: error})
       Toast.show(_(msg`An issue occurred starting the chat`), 'xmark')
     },
   })
--- a/src/components/dms/dialogs/NewChatDialog.tsx
+++ b/src/components/dms/dialogs/NewChatDialog.tsx
@@ -2,6 +2,7 @@ import React, {useCallback} from 'react'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
+import {logger} from '#/logger'
 import {useGetConvoForMembers} from '#/state/queries/messages/get-convo-for-members'
 import {logEvent} from 'lib/statsig/statsig'
 import {FAB} from '#/view/com/util/fab/FAB'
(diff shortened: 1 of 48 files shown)


haileyok:
Do we want to do the same for native?

src/components/dms/dialogs/NewChatDialog.tsx
```diff
@@ -31,7 +32,8 @@ export function NewChat({
       logEvent('chat:open', {logContext: 'NewChatDialog'})
     },
     onError: error => {
-      Toast.show(error.message)
+      logger.error('Failed to create chat', {message: error})
```
haileyok, line 35: i think this needs to be `safeMessage`

mozzius:
> Do we want to do the same for native?

https://github.com/bluesky-social/social-app/pull/4808

haileyok (approved):

src/lib/media/picker.shared.ts
```diff
@@ -17,14 +17,14 @@ export async function openPicker(opts?: ImagePickerOptions) {
   })
 
   if (response.assets && response.assets.length > 4) {
-    Toast.show('You may only select up to 4 images')
+    Toast.show('You may only select up to 4 images', 'exclamation-circle')
```
surfdude29, line 20: While you're updating toasts, would it be possible to mark this for localization please? 🙏

src/lib/media/picker.shared.ts
```diff
@@ -17,14 +17,14 @@ export async function openPicker(opts?: ImagePickerOptions) {
...
   }
 
   return (response.assets ?? [])
     .slice(0, 4)
     .filter(asset => {
       if (asset.mimeType?.startsWith('image/')) return true
-      Toast.show('Only image files are supported')
+      Toast.show('Only image files are supported', 'exclamation-circle')
```
surfdude29, line 27: And here

src/lib/sharing.ts
```diff
@@ -20,6 +20,6 @@ export async function shareUrl(url: string) {
     // React Native Share is not supported by web. Web Share API
     // has increasing but not full support, so default to clipboard
     setStringAsync(url)
-    Toast.show('Copied to clipboard')
+    Toast.show('Copied to clipboard', 'clipboard-check')
```
surfdude29, line 23: Also here

src/components/dms/dialogs/NewChatDialog.tsx
mozzius: Seems like it's super inconsistent across the whole codebase, I'll change it here but we should look at it in other places too

PR #4808, opened 2024-07-21
Native toast rework

mozzius:
- New look
- Entry/exit animations
- Icons

Screenshot 2024-07-20 at 17 56 53
Screenshot 2024-07-20 at 17 58 03
Screenshot 2024-07-20 at 17 54 20
Screenshot 2024-07-20 at 17 55 47

### DIFF
--- a/src/alf/atoms.ts
+++ b/src/alf/atoms.ts
@@ -145,6 +145,9 @@ export const atoms = {
   flex_shrink: {
     flexShrink: 1,
   },
+  flex_shrink_0: {
+    flexShrink: 0,
+  },
   justify_start: {
     justifyContent: 'flex-start',
   },
--- a/src/view/com/util/Toast.tsx
+++ b/src/view/com/util/Toast.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useEffect, useState} from 'react'
 import {View} from 'react-native'
 import Animated, {FadeInUp, FadeOutUp} from 'react-native-reanimated'
 import RootSiblings from 'react-native-root-siblings'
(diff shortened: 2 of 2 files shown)


estrattonbailey (approved):
Nice! I'm +1, and we can continue to improve over time.

Exit animation on iOS isn't working for me though, if that's a quick fix we might as well do that here.

haileyok:
Don't know what the most elegant way to do this would be - first time I looked at this code, had no idea it worked this way before...but anyway, I think the problem @estrattonbailey is seeing is because the element needs to actually perform the animation before it gets destroyed. So i.e. if your exit animation runs for 300ms, you'd need to do something weird like:

- Add root element, add a conditional child with `isVisible` true
- Set `isVisible` to `false` at 3800 ms
- Destroy the root element at 4000ms

That feels pretty weird. At this point, you may as well just have the conditionally rendered element and leave the root always rendered. But we also have to make sure this API can remain usable outside React, unless we want to significantly change code (we use it outside react in quite a few places, which admittedly is nice and I think we should continue to allow if possible). I'm not the right person to provide an elegant and correct solution probably lol...but just some info based on problems I've had in the past.

estrattonbailey:
Oh yeah good call — seems like a fine approach to me. `RootSiblings` shouldn't care what's rendered there, so should work I think.

mozzius:
That's extremely odd, exit animation works on Android and Reanimated *should* keep it displayed while animating out

mozzius:
But yeah the old version faded out before unmounting so we can just use the same approach here. Not as neat but oh well

mozzius:
Fixed by making sure the Animated component is not at the root

PR #4814, opened 2024-07-23
Fix switch style and animate

mozzius:
After we changed all the border widths to hairline, the switch component broke, since the size of the container was expecting the border to be 1px. Setting the height to `20 + hairlineWidth * 2` fixes it

While I'm here, I added a layout animation, made it slightly wider, and made it use flexbox instead of absolute positioning

Before
Screenshot 2024-07-23 at 21 58 05
After
Screenshot 2024-07-23 at 21 53 04

### DIFF
--- a/src/components/forms/Toggle.tsx
+++ b/src/components/forms/Toggle.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {Pressable, StyleSheet, View, ViewStyle} from 'react-native'
+import {Pressable, View, ViewStyle} from 'react-native'
 import Animated, {LinearTransition} from 'react-native-reanimated'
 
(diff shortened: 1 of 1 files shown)


src/components/forms/Toggle.tsx
```diff
@@ -375,34 +377,35 @@ export function Switch() {
...
         a.relative,
-        a.border,
         a.rounded_full,
         t.atoms.bg,
         t.atoms.border_contrast_high,
         {
+          borderWidth: 1,
           height: 20,
-          width: 30,
+          width: 32,
```
estrattonbailey, line 386: I also felt that 35 was a little wide but can change it back nbd

estrattonbailey (approved):
Nice! Animation is hott

PR #4816, opened 2024-07-23
Hack-modify avi path to use smaller thumbs behind gate

pfrazee:
We now have smaller AVI thumbnails being served. We will eventually update our lexicons, or just the server responses, to give the new smaller AVI-thumb paths. For now, we're manually modifying the path.

Note: behind the `small_avi_thumb` gate so that we can roll this out progressively, since this change will result in a bunch of CDN cache misses and we need to control its rollout.

### DIFF
--- a/src/lib/statsig/gates.ts
+++ b/src/lib/statsig/gates.ts
@@ -14,3 +14,4 @@ export type Gate =
   | 'suggested_follows_interstitial'
   | 'ungroup_follow_backs'
   | 'videos'
+  | 'small_avi_thumb'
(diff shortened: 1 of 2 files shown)


gaearon (approved):
looks good, but let's add an exception for the one rendered from `ProfileHeaderShell`? like `isFullSize={true}` that defaults to `false` and overrides this

pfrazee:
> looks good, but let's add an exception for the one rendered from `ProfileHeaderShell`? like `isFullSize={true}` that defaults to `false` and overrides this

I added a size check which handles it well enough

PR #4818, opened 2024-07-24
Respect labels on feeds and lists

estrattonbailey:
Feeds and lists can now be labeled. This PR integrates the pre-existing `moderateFeedGenerator` and `moderateUserList` helpers.

If you have NSFW disabled, feeds with NSFW labels are filtered from:
  - Feeds screen
  - Explore page
  - Search page
  - Starter packs
  - Feeds tab on profiles

If a list has a non-overrideable label on it such as `!hide`, they are:
- filtered from the Lists tab on profiles
- hidden in posts using `ContentHider`
- hidden from direct view using `ScreenHider`

This PR opts to do the filtering within the queries themselves — so right at the source — as opposed to where the data is used. This is good for DRYness, and also atm we don't support other moderation on feeds or lists, such as avatar blurring, so there's no need to drill it down.

Also, this PR merges two popular feeds search queries used on the Search screen, and in the StepFeeds screen of starter packs. That way they can use the same filter logic. No functional changes, just chose the one that required the least amount of work to swap.

### Testing
- For feeds, just toggle your mod settings and hit all the surfaces listed above.
- For lists, use this post, click into the list, check out profile of owner: http://localhost:19006/profile/test.esb.lol/post/3ky26ll43us27


src/screens/StarterPack/Wizard/StepFeeds.tsx
```diff
@@ -59,7 +59,7 @@ export function StepFeeds({moderationOpts}: {moderationOpts: ModerationOpts}) {
       : undefined
 
   const {data: searchedFeeds, isFetching: isFetchingSearchedFeeds} =
-    useSearchPopularFeedsQuery({q: throttledQuery})
+    usePopularFeedsSearch({query: throttledQuery})
```
estrattonbailey, line 62: With the addition of `placeholderData: keepPreviousData,` there was no difference here, so should be a clean swap.

src/state/queries/feed.ts
```diff
@@ -225,6 +229,7 @@ export function useGetPopularFeedsQuery(options?: GetPopularFeedsOptions) {
     QueryKey,
     string | undefined
   >({
+    enabled: Boolean(moderationOpts),
```
estrattonbailey, line 232: This is copied from how we do it for post feeds.

src/state/queries/feed.ts
```diff
@@ -304,31 +314,24 @@ export function useGetPopularFeedsQuery(options?: GetPopularFeedsOptions) {
...
-      const res = await agent.app.bsky.unspecced.getPopularFeedGenerators({
-        limit: 15,
-        query: q,
-      })
+      if (moderationOpts) {
+        return res.data.feeds.filter(feed => {
+          const decision = moderateFeedGenerator(feed, moderationOpts)
+          return !decision.ui('contentList').filter
+        })
+      }
```
estrattonbailey, line 331: Mutations don't support `enabled`, but this method is called after load when the user types in the search input on Feeds screen. So even though we have a check, it should be present.

I spent a little time trying to swap for the other search query in here (not a mutation) but decided it wasn't worth the effort right now to refactor the Feeds screen to use it. This was a quick simple approach that works.

src/state/queries/feed.ts
```diff
@@ -346,17 +349,27 @@ export function usePopularFeedsSearch({
...
 
       return res.data.feeds
     },
+    placeholderData: keepPreviousData,
+    select(data) {
+      return data.filter(feed => {
+        const decision = moderateFeedGenerator(feed, moderationOpts!)
+        return !decision.ui('contentList').filter
+      })
+    },
```
estrattonbailey, line 372: So this is now used on Starter Packs as well. No functional changes. Upped `limit` to match SPs, Search screen is happy to show 15 instead of 10.

pfrazee (approved):
LGTM

src/state/queries/profile-feedgens.ts
```diff
@@ -38,5 +40,21 @@ export function useProfileFeedgensQuery(
     initialPageParam: undefined,
     getNextPageParam: lastPage => lastPage.cursor,
     enabled,
+    select(data) {
+      return {
+        ...data,
+        pages: data.pages.map(page => {
```
gaearon, line 46: Can you check if this is going to lead to re-renders for every pagination? See the hack we added to `select` in `post-feed` to avoid that there. I guess it's not super important here per se but would be good to verify.

src/state/queries/profile-lists.ts
```diff
@@ -27,17 +29,32 @@ export function useProfileListsQuery(did: string, opts?: {enabled?: boolean}) {
...
-      }
+      return res.data
     },
     initialPageParam: undefined,
     getNextPageParam: lastPage => lastPage.cursor,
     enabled,
+    select(data) {
+      return {
+        ...data,
+        pages: data.pages.map(page => {
```
gaearon, line 40: Same question about re-renders.

gaearon (commented):
not critical cause it's unlikely to paginate these

src/state/queries/profile-feedgens.ts
estrattonbailey: Did some digging on this and even did the approach from `post-feed`, but then realized it was really just `renderItem` on the `List` not being memoized 🙃

src/state/queries/profile-lists.ts
estrattonbailey: Similar solve to feedgens, the `item` objects passed to `renderItem` were being spread anew every time. I changed it to use the (slightly icky) approach we use for `ProfileFeedgens` and that seems to fix it up.

src/state/queries/profile-feedgens.ts
estrattonbailey: But yeah you were right, there were re-renders happening here, though not bc of the query 😆

PR #4819, opened 2024-07-24
make some settings screens scrollable for accessibility

haileyok:
## Why

Just using `CenteredView` on native causes some screens - like DM settings - to not allow the user to view all the options if they are using font scaling. Make them scrollviews instead.

## Test Plan

Test that the views are the same, just scrollable now.


pfrazee (approved):
LGTM

PR #4821, opened 2024-07-24
make toast shorter

haileyok:
## Why

Too long, want short.

## Test Plan

Goes away faster


estrattonbailey (approved):
Such short. Too short?

haileyok:
2.5 much long, 1.75 much short, 2 gud.

(Everyone can give an opinion i think once we get on device, but this feels a lot better. Less blocking of buttons like back)

PR #4828, opened 2024-07-25
ALF suggested follows in profile header

estrattonbailey:
Swaps in the components from the FeedInterstitial, otherwise no functional changes.

CleanShot 2024-07-25 at 11 05 03@2x


haileyok (approved):
Tested on all platforms, looks good!

PR #4830, opened 2024-07-25
Fix sloppy filter(Boolean) types

gaearon:
We were using this `as Foo[]` pattern which is not good. I got bit by this in my ongoing feed slice refactor because the types were lyring. I'm using [this approach](https://stackoverflow.com/questions/47632622/typescript-and-filter-boolean#comment93284037_47636222) to get them typed more strongly. It requires extracting variables instead of early returns and some manual annotations, but these actually _do_ get checked.

This surfaced some violations. They're not actually breaking the logic. These two are trivial:

- https://github.com/bluesky-social/social-app/commit/16c714a4ba64da90cff696b742691fa9f0e7d1d6
- https://github.com/bluesky-social/social-app/commit/d3189d3fe27819a82046c6079e4f0b2875150584

This one is more involved:

- https://github.com/bluesky-social/social-app/commit/f3caf3e921fb7c0e8dfaf449c1229346511bb6c5
- https://github.com/bluesky-social/social-app/commit/41f2dd7ebdaafd6522af28cb04c65e70b69e9720
- https://github.com/bluesky-social/social-app/commit/6dadf673d46f8cb2351f61effc41fa03d82fb9bf

It uncovered type errors stemming from `parentAuthor` being defined as `ProfileViewBasic` even though technically it could be anything (because we're reading `author` from a union). We also aren't able to use `isProfileViewBasic()` here because the object from the server does not have a `$type`. So instead I'm checking the `post` itself. If it's a `PostView`, we can read the `author` from it. I've had to add some extra logic to deal with blocked posts because the component types were lying.

## Test Plan

- Verified editing thread gates still work for each option.
- Verified "Load more" shows up and works as expected in Explore for both profiles and feeds.
- Verified "in reply to..." still shows up in feed
  - Placed breakpoints in both cases where we assign `parentAuthor` and verified we're able to hit both
  - Verified "Reply to you", "Reply to blocked post", "Reply to <...>" all work

### DIFF
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -78,7 +78,10 @@ export interface FeedPostSliceItem {
   uri: string
   post: AppBskyFeedDefs.PostView
   record: AppBskyFeedPost.Record
-  reason?: AppBskyFeedDefs.ReasonRepost | ReasonFeedSource
+  reason?:
+    | AppBskyFeedDefs.ReasonRepost
+    | ReasonFeedSource
+    | {[k: string]: unknown; $type: string}
   feedContext: string | undefined
   moderation: ModerationDecision
   parentAuthor?: AppBskyActorDefs.ProfileViewBasic
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -344,9 +344,17 @@ export function usePostFeedQuery(
                           AppBskyFeedPost.validateRecord(item.post.record)
                             .success
                         ) {
-                          const parentAuthor =
-                            item.reply?.parent?.author ??
-                            slice.items[i + 1]?.reply?.grandparentAuthor
+                          const parent = item.reply?.parent
+                          let parentAuthor:
+                            | AppBskyActorDefs.ProfileViewBasic
+                            | undefined
+                          if (AppBskyFeedDefs.isPostView(parent)) {
+                            parentAuthor = parent.author
+                          }
+                          if (!parentAuthor) {
+                            parentAuthor =
+                              slice.items[i + 1]?.reply?.grandparentAuthor
+                          }
                           const replyRef = item.reply
                           const isParentBlocked = AppBskyFeedDefs.isBlockedPost(
                             replyRef?.parent,
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -323,7 +323,7 @@ export function usePostFeedQuery(
                     )
                   }
 
-                  return {
+                  const feedPostSlice: FeedPostSlice = {
                     _reactKey: slice._reactKey,
                     _isFeedPostSlice: true,
                     rootUri: slice.rootItem.post.uri,
@@ -349,7 +349,7 @@ export function usePostFeedQuery(
                             replyRef?.parent,
                           )
 
-                          return {
+                          const feedPostSliceItem: FeedPostSliceItem = {
                             _reactKey: `${slice._reactKey}-${i}-${item.post.uri}`,
                             uri: item.post.uri,
                             post: item.post,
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Oh wow I'm disappointed in myself for not having thought of a type-assertion closure like that. Muuuuch better

Faithfinder:
Upgrade Typescript to 5.5, then you can just do [`x => !!x`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates) without any typing code around it =P

janaagaard75:
> Upgrade Typescript to 5.5, then you can just do [x => !!x](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates) without any typing code around it =P

(Unsure if you're being sarcastic.)

I think these kind of clever type conversions should be avoided, because it opens up minefield of edge cases that are hard to remember. See this comparison table: https://dorey.github.io/JavaScript-Equality-Table/. This TypeScript ESLint rule enforces using strict boolean expressions (kinda like in C#): https://typescript-eslint.io/rules/strict-boolean-expressions/. I think the added clarity outweighs the added verbosity.

Faithfinder:
> > Upgrade Typescript to 5.5, then you can just do [x => !!x](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#inferred-type-predicates) without any typing code around it =P
> 
> (Unsure if you're being sarcastic.)
> 
> I think these kind of clever type conversions should be avoided, because it opens up minefield of edge cases that are hard to remember. See this comparison table: [dorey.github.io/JavaScript-Equality-Table](https://dorey.github.io/JavaScript-Equality-Table/). This TypeScript ESLint rule enforces using strict boolean expressions (kinda like in C#): [typescript-eslint.io/rules/strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/). I think the added clarity outweighs the added verbosity.

No sarcasm at all. Not sure what that table has to do with anything, never proposed using `==` for anything.

As for the rule, didn't enjoy using it, personally, and I don't see it enabled in this codebase from a [cursory](https://github.com/bluesky-social/social-app/blob/main/.eslintrc.js) glance. `Boolean` and `!!` are just two ways to write the same thing, in the end. Due to that, typescript-eslint maintainers are [considering](https://github.com/typescript-eslint/typescript-eslint/issues/9060#issuecomment-2099361707) banning usage of `Boolean` for bypassing that rule, though probably won't really happen.

Here, my comment was just aiming at simplifying people's lives. Why write `<T>`, when it's possible to skip it? Alternatively you could use [`ts-reset`](https://github.com/total-typescript/ts-reset), which I bet Dan is familiar with, but that's potentially a more controversial change than just upgrading Typescript.

gaearon:
I just copied the first thing I found on SO, happy to merge if there’s some better fix.

Faithfinder:
> I just copied the first thing I found on SO, happy to merge if there’s some better fix.

@gaearon Chose whichever option strikes your fancy: #4840, #4841

janaagaard75:
> Not sure what that table has to do with anything, never proposed using `==` for anything.

Sorry. I meant to link to the [third tab with the truthy & falsy values in JavaScript](https://dorey.github.io/JavaScript-Equality-Table/#if-statement). It might just be me, but I find it very hard to remember things like an empty array is true while an empty string is false. It's my experience that forcing expressions to be explicit using [ESLint's strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/) avoids a surprising large amount of silly bugs, and that it's well worth the extra verbosity that this adds to the code.

My point is that it's worth avoiding code like `if (!!someArray) { ... }` and `if (!!someString) { ... }`.

PR #4831, opened 2024-07-25
Replace `import hairlineWidth =` with const

mozzius:
For whatever reason this was using the weird TypeScript typed CJS require import thing when it should've been a normal const. This was messing up the import sorting


haileyok:
rare intellij L

haileyok (approved):

PR #4834, opened 2024-07-25
Refactor feed slices

gaearon:
I'm refactoring feed slicing. This refactor should fully preserve the existing behavior but it makes strategic code changes that will prepare us for the next PR https://github.com/bluesky-social/social-app/pull/4835, which solves longstanding problems with how replies show up in the feed.

The best way to follow along, unfortunately, is to read commit-by-commit. Each change is meant to be atomic and safe (mostly manipulation). The main theme of this refactor is to decouple the shape of a "slice item" from `FeedPostView` so that https://github.com/bluesky-social/social-app/pull/4835 can construct it on-the-fly from parent and root with correct information. I'm tucking away low-level access of `slices[i].items.reply` and similar fields into the `slice` API itself so that I can later change the type of individual items.

I guess you could also just [read the entire diff](https://github.com/bluesky-social/social-app/pull/4834/files?w=1).

By the end of this refactor, the only code that reaches out into "internals" of a slice is the slice item manipulation logic which the next PR will get rid of.

## Test Plan

This is going to require some non-trivial testing. Things you'll want to look at:

- Mergefeed on/off
- Reposts of replies
- Self-threads
- Other people's threads
- Replies on/off
- Like threshold on replies
- Follow setting on replies
- Deduplication
- Reposts on/off
- Quotes on/off
- Replies to blocked/deleted posts

The goal in this PR is to stay aligned with the behavior on main.

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -64,10 +64,6 @@ export class FeedViewPostsSlice {
     )
   }
 
-  get isFullThread() {
-    return this.isThread && !this.items[0].reply
-  }
-
   get rootItem(): FeedViewPost {
     return this._originalFeedViewPost
   }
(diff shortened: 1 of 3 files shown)


haileyok (approved):
Did some testing, toggled the various settings and things seem to line up localhost vs. bsky.app.

PR #4839, opened 2024-07-25
[LinkLab] Add labels to mod details dialog

estrattonbailey:


src/components/moderation/ModerationDetailsDialog.tsx
```diff
@@ -127,10 +133,11 @@ function ModerationDetailsDialogInner({
               <Trans>
                 This label was applied by{' '}
                 <InlineLinkText
+                  label={desc.source || _(msg`an unknown labeler`)}
                   to={makeProfileLink({did: modcause.label.src, handle: ''})}
                   onPress={() => control.close()}
                   style={a.text_md}>
-                  {desc.source}
+                  {desc.source || _(msg`an unknown labeler`)}
```
estrattonbailey, line 140: This value can actually be `undefined`, so need a fallback here too.

pfrazee (approved):

PR #4841, opened 2024-07-25
Boolean filter improvement alternative: TS-reset

Faithfinder:
This is a follow up to #4830. This is a pull request with an alternative solution: see #4840. Feel free to merge both or either.

This version introduces [`ts-reset`](https://github.com/total-typescript/ts-reset) and uses it's type enchancements to simplify boolean filters.

Note that I had to enable only one enchancement from ts-reset, because enabling the whole of it fails the typecheck for the codebase due to `JSON.parse` returning `unknown` instead of `any`. You might want to look into enabling more, but it's out of scope for this PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -222,6 +222,7 @@
     "@react-native/typescript-config": "^0.74.1",
     "@testing-library/jest-native": "^5.4.1",
     "@testing-library/react-native": "^11.5.2",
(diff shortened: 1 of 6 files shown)


Faithfinder:
Closing cause other alternative was selected

PR #4849, opened 2024-07-26
Only show "followed you back" when appropriate

haileyok:
## Why

Right now, we will show the user "followed you back" in a follow notification if you follow the user. There's no other logic that checks if the follow occurred before or after you followed that person.
This means that if I follow you, then you follow me, your notification showing that I followed you will say "followed you back", even though I actually followed you first.

## How

Let's compare the rkey's timestamp with the new record's timestamp.

## Test Plan

See video

https://github.com/user-attachments/assets/f1961bfd-8169-44e3-9125-c154b8561bd8

### DIFF
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -194,13 +194,12 @@ let FeedItem = ({
       try {
         const rkey = new AtUri(item.notification.author.viewer.following).rkey
         const followingTimestamp = TID.fromStr(rkey).timestamp()
-        const followedTimestamp =
-          new Date(item.notification.record.createdAt).getTime() * 1000
-        isFollowBack = followedTimestamp > followingTimestamp
       } catch (e) {
         // For some reason the following URI was invalid. Default to it not being a follow back.
         console.error('Invalid following URI')
       }
+      const followedTimestamp = new Date(item.notification.record.createdAt).getTime() * 1000
+      isFollowBack = followedTimestamp > followingTimestamp
     }
 
     if (isFollowBack && gate('ungroup_follow_backs')) {
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -13,11 +13,13 @@ import {
   AppBskyEmbedRecordWithMedia,
   AppBskyFeedDefs,
   AppBskyFeedPost,
+  AppBskyGraphFollow,
   moderateProfile,
   ModerationDecision,
   ModerationOpts,
 } from '@atproto/api'
 import {AtUri} from '@atproto/api'
+import {TID} from '@atproto/common-web'
 import {msg, plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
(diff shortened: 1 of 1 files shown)


src/view/com/notifications/FeedItem.tsx
```diff
@@ -184,10 +186,21 @@ let FeedItem = ({
...
       item.notification.author.viewer?.following &&
-      gate('ungroup_follow_backs')
+      AppBskyGraphFollow.isRecord(item.notification.record)
     ) {
+      const rkey = new AtUri(item.notification.author.viewer.following).rkey
+      const followingTimestamp = TID.fromStr(rkey).timestamp()
+      const followedTimestamp =
+        new Date(item.notification.record.createdAt).getTime() * 1000
+
+      console.log(followedTimestamp, followingTimestamp)
```
gaearon, line 199: log

src/view/com/notifications/FeedItem.tsx
```diff
@@ -184,10 +186,24 @@ let FeedItem = ({
...
     ) {
+      try {
+        const rkey = new AtUri(item.notification.author.viewer.following).rkey
+        const followingTimestamp = TID.fromStr(rkey).timestamp()
+        const followedTimestamp =
+          new Date(item.notification.record.createdAt).getTime() * 1000
+        isFollowBack = followedTimestamp > followingTimestamp
+      } catch (e) {
+        // For some reason the following URI was invalid. Default to it not being a follow back.
+        console.error('Invalid following URI')
```
kulak91, line 202: Just curios about this.
Console error will be never be shown in production. Wouldn't it be more efficient to pass this error to the sentry to track possible issues later on?

src/view/com/notifications/FeedItem.tsx
```diff
@@ -184,10 +186,24 @@ let FeedItem = ({
...
-      gate('ungroup_follow_backs')
+      AppBskyGraphFollow.isRecord(item.notification.record)
     ) {
+      try {
+        const rkey = new AtUri(item.notification.author.viewer.following).rkey
+        const followingTimestamp = TID.fromStr(rkey).timestamp()
+        const followedTimestamp =
+          new Date(item.notification.record.createdAt).getTime() * 1000
+        isFollowBack = followedTimestamp > followingTimestamp
+      } catch (e) {
```
gaearon, line 200: If this catch exists solely for the AtUri constructor, let's move just the URI construction in it.

```
let rkey
try {
  rkey = ...
} catch (e) {
  // ..
}
if (rkey) {
  // ...
}
```

Otherwise we risk catching random other mistakes like typos or null references that we won't notice.

gaearon (changes requested):
let's make the try/catch more scoped

src/view/com/notifications/FeedItem.tsx
gaearon: In general, the rule of thumb is we should keep the minimal necessary code that we expect to throw in try/catches like this, and move all `foo.bar.baz` access or user code out of it. Otherwise we're risking swallowing legitimate code mistakes.

haileyok: Yea, this here is fine. Will change.

haileyok: Lets see if prettier complains about that...

gaearon (approved):
have not tested the date comparison but code lgtm

PR #4857, opened 2024-07-31
Upgrade API, implement XRPC rework

haileyok:
Bumping API, adding changes from https://github.com/bluesky-social/social-app/pull/3814


src/state/session/agent.ts
```diff
@@ -33,20 +33,30 @@ export async function createAgentAndResume(
...
+    persistSession: event => {
+      const {session} = agent
+      if (!session?.did) return // @TODO SESSION - ?
+
+      onSessionChange(agent, session.did, event)
+      if (event !== 'create' && event !== 'update') {
+        addSessionErrorLog(session.did, event)
+      }
+    },
+  })
```
haileyok, line 47: `persistSession` needs to get set inside of agent constructor now. Does this make sense?

src/state/session/agent.ts
```diff
@@ -81,41 +91,45 @@ export async function createAgentAndLogin(
...
```
haileyok, line 105: Same logic/question here.

src/state/session/agent.ts
```diff
@@ -33,20 +33,30 @@ export async function createAgentAndResume(
...
   const prevSession: AtpSessionData = sessionAccountToSession(storedAccount)
   if (isSessionExpired(storedAccount)) {
     await networkRetry(1, () => agent.resumeSession(prevSession))
   } else {
-    agent.session = prevSession
     if (!storedAccount.signupQueued) {
-      // Intentionally not awaited to unblock the UI:
-      networkRetry(3, () => agent.resumeSession(prevSession)).catch(
+      // @TODO SESSION - We were not awaiting this before. Why? And why is it now broken if we do not await?
+      await networkRetry(3, () => agent.resumeSession(prevSession)).catch(
```
haileyok, line 59: What UI gets blocked here if we await? (Also, what changed that we need to await this now for it to work?)

src/state/session/index.tsx
```diff
@@ -247,10 +235,6 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       const prevAgent = currentAgentRef.current
       currentAgentRef.current = agent
       addSessionDebugLog({type: 'agent:switch', prevAgent, nextAgent: agent})
-      // We never reuse agents so let's fully neutralize the previous one.
-      // This ensures it won't try to consume any refresh tokens.
-      prevAgent.session = undefined
-      prevAgent.setPersistSessionHandler(undefined)
```
haileyok, line 253: At a very quick glance, it seems like maybe what we want to do here is just destroy the agent? `session` no longer can be set, and `setPersistSessionHandler` isnt available anymore, so need something here (assuming we need this code)

src/lib/moderation.ts
```diff
@@ -60,10 +61,9 @@ export function isAppLabeler(
     | AppBskyLabelerDefs.LabelerView
     | AppBskyLabelerDefs.LabelerViewDetailed,
 ): boolean {
-  if (typeof labeler === 'string') {
-    return BskyAgent.appLabelers.includes(labeler)
-  }
-  return BskyAgent.appLabelers.includes(labeler.creator.did)
+  const did = typeof labeler === 'string' ? labeler : labeler.creator.did
+  // assertion here that `did` starts with `did:`
+  return did.startsWith('did:') && BskyAgent.appLabelers.includes(did as Did)
```
haileyok, line 66: This is maybe a little weird, but this `includes` has a type of `did:${string}` now. Feels maybe slight unnecessary - and it's equally weird imo that, i.e. `ProfileView`'s `creator.did` wouldn't be typed as `Did` if this is what we want to do.

haileyok: @devinivy for context

src/state/queries/preferences/moderation.ts
```diff
@@ -21,7 +22,7 @@ export function useMyLabelersQuery() {
   const dids = Array.from(
     new Set(
       BskyAgent.appLabelers.concat(
-        prefs.data?.moderationPrefs.labelers.map(l => l.did) || [],
+        prefs.data?.moderationPrefs.labelers.map(l => l.did as Did) || [],
```
haileyok, line 25: Same `as Did` here.

src/state/session/index.tsx
```diff
@@ -196,18 +193,20 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+          const nextSession = sessionAccountToSession(syncedAccount)
+          agent.resumeSession(nextSession)
           addSessionDebugLog({
             type: 'agent:patch',
             agent,
             prevSession,
-            nextSession: agent.session,
+            nextSession,
           })
         }
```
haileyok, line 211: Slightly modified since we can't set `agent.session` this way anymore (and don't want to rm since it looks like a lot of users are hitting this...). However, wondering about one thing...

`agent.session` can possibly be undefined - and I think likely would be here if we call `agent.resumeSession` without `await`. Would we have enough info for the log by just passing it the session from `sessionAccountToSession()`?

src/state/session/agent.ts
```diff
@@ -174,31 +188,20 @@ export async function createAgentAndCreateAccount(
...
-    event: AtpSessionEvent,
-  ) => void,
 ) {
   // There's nothing else left to do, so block on them here.
   await Promise.all([gates, moderation])
 
   // Now the agent is ready.
   const account = agentToSessionAccountOrThrow(agent)
-  agent.setPersistSessionHandler(event => {
-    onSessionChange(agent, account.did, event)
```
gaearon, line 197: The intention in the original code is to "decide" on a `did` by this point (note we're reading it from the account) and then to consider the account fully tied to that `did`. This is why we're using `account.did` (which is immutable) and not reading from the `session` object. I think we'd want to keep a similar guarantee after restructuring.

src/lib/moderation.ts
devinivy: I agree, it's probably worth loosening those types to match the types we get out of lexicon for DIDs.  If we want to use the `Did` type internally to the package, I see no issue with that— just a little funky in the public API since we don't use this type for DIDs elsewhere.  @matthieusieben what do you think of that?

src/state/session/agent.ts
haileyok: note to self: storedAccount

src/state/session/agent.ts
haileyok: Going to tweak with this in a separate PR per discussion at start of week. Left some context in slack, but here I've just duplicated the previous behavior (as far as I can tell) by using a weird hack.

src/lib/moderation.ts
matthieusieben: I changed the type here so this shouldn't be needed anymore

src/state/queries/preferences/moderation.ts
matthieusieben: can be reverted now

src/state/session/agent.ts
haileyok: no longer necessary

package.json
```diff
@@ -52,7 +52,7 @@
     "open-analyzer": "EXPO_PUBLIC_OPEN_ANALYZER=1 yarn build-web"
   },
   "dependencies": {
-    "@atproto/api": "0.12.29",
+    "@atproto/api": "^0.13.0",
```
mozzius, line 55: idk what your thoughts are on pinning the exact version here. We rely so closely on the specific version, I think it'd make sense to pin it

mozzius (approved):
all good, just one thought

package.json
haileyok: oh i didnt even realize we had it pinned before. yea lets.

PR #4858, opened 2024-07-31
remove `SignupQueued` screen

haileyok:
Doesn't seem necessary anymore. Has some stuff that will need to get changed w/ new XRPC rework, easier to just rm this screen.


mozzius (approved):

haileyok:
Leaving this in.

Issue #2426, opened 2024-01-04
Flag emoji not displaying on desktop windows


maisondasilva:
On my social network I like to use my country's flag, however on BlueSky I see that BR appears instead of the Flag, is it possible to correct this? I see this with my friends who use flags from other languages too!

Additionally, there could be an option to add the website and other information to the profile.

BlueSky

![image](https://github.com/bluesky-social/social-app/assets/11423362/c1a6b4ce-9d8f-4ad5-88fa-a3242d1b6516)

Another

![image](https://github.com/bluesky-social/social-app/assets/11423362/8e0e0f27-0fd5-41f0-92c0-1057fc4b4024)

pfrazee:
Marking this as a bug because that emoji should be rendering

haileyok:
MS doesn't actually support regional flag emoji. In Twitter's case they use their own Twemoji: https://abs-0.twimg.com/emoji/v2/svg/1f1e7-1f1f7.svg whereas if you used https://nitter.net/maisondasilva you'll see the "BR" as well. Firefox on Windows 
also uses Twemoji for whatever emoji are not supported by MS. Chrome I believe does not and will also display the BR.

Only solution would be to render hosted assets for these flags afaik.

stanjourdan:
I would add that when creating a new post (using chrome / windows), most flags would not even show up in the emoticon panel: 

![Image](https://github.com/user-attachments/assets/d7e3478a-c93d-4860-92ba-168d429188a6)

but it does work from the Android app.

PR #4862, opened 2024-07-31
Add twemoji

### LINKED ISSUE #1258: Wrap unicode flag codepoints in a flag supporting font.
**Problem**

Windows has for geo-political reasons avoided adding support for unicode flags, instead they render as the ISO country code. This means no Windows machines will display flags in messages if they're inserted via the unicode representation.

In order to display flags on Windows messages Twitter uses a custom font. BlueSky could consider the same.

Embeddable fonts that support flags include

[Noto](https://fonts.google.com/noto) from Google (no license indicated)
[twemoji](https://github.com/twitter/twemoji) from Twitter (MIT licensed)
[BabelStone](https://www.babelstone.co.uk/Fonts/Flags.html) ([SIL Open Font License](http://scripts.sil.org/OFL) 1.1,)

Rewrite the character to an image. This would disrupt screen readers, and would not scale if a user increases the font size in the browser)

**Additional context**

As a bonus the fonts may support newer flag additions faster than the operating system support, for example the LBGTQ pride flag.
Conversely, they may ignore certain flags altogether,

mozzius:
Closes https://github.com/bluesky-social/social-app/issues/1258, https://github.com/bluesky-social/social-app/issues/2426

Uses @discordapp/twemoji (discord's fork, I think the original is dead)

Uses jsdelivr CDN, but I'm guessing we'd want to host this ourselves - leaving in draft until we get the backend resource to do that

Implementation:
Made a twemoji component which takes a string, and if it's got emoji in it, splits the graphemes and replaces them with images. Scattered it all around the RichText component, wherever it's rendering a string

Screenshot 2024-07-31 at 23 29 32

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -253,6 +253,14 @@
       from { opacity: 1; }
       to { opacity: 0; }
     }
+
+    /* Twemoji */
+    img.twemoji {
+      height: 1em;
+      width: 1em;
+      margin: 0 .05em 0 .1em;
+      vertical-align: -0.1em;
+    }
   </style>
   </style>
   {% include "scripts.html" %}
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@bam.tech/react-native-image-resizer": "^3.0.4",
     "@braintree/sanitize-url": "^6.0.2",
     "@discord/bottom-sheet": "bluesky-social/react-native-bottom-sheet",
+    "@discordapp/twemoji": "^15.0.3",
     "@emoji-mart/react": "^1.1.1",
     "@expo/html-elements": "^0.4.2",
     "@expo/webpack-config": "^19.0.0",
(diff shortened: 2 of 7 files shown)


connie-e:
There is also this fork of Twemoji. https://github.com/jdecked/twemoji

mozzius:
Ah, thanks, that looks a little more up-to-date

j4ckxyz:
Not really sure how this all works (GitHub) - but there's this [fork of Twemoji](https://github.com/jdecked/twemoji) by a former Twitter employee, who is keeping the project going. From what I've seen, Discord is slower to add new emoji types than jdecked. This fork also supports the Unicode 15.0 spec. Might be more fitting than the Discord fork.

GlitchyPSIX:
jdecked's twemoji fork is the current fork everyone using twemoji uses, it is consistently updated

PR #4863, opened 2024-07-31
Fetch enough pages to fill a page's worth of items

gaearon:
We have a hack that fetches enough items to fill the first page's worth of items (30) if we're not getting enough items within the first few pages (6). This hack isn't reliable and doesn't handle similar cases further down the scroll (e.g. when there's an empty `[]` response in for further pages, or when too many items get filtered out later).

I'm adding a more extensive fix that should handle these cases consistently. The idea of the fix is to keep track of the "wanted" item count. When we start paginating, we bump the "wanted" item count by the page size (30). If we then don't receive enough items, we keep trying to paginate until those 30 items are filled. There are a few conditions where we reset that counter.

This mostly acts as a failsafe. In practice `FlatList` content size change triggering `onEndReached` (which was ported to web in https://github.com/bluesky-social/social-app/pull/4859) is accidentally sufficient in most cases because rendering a pagination spinner causes a content size change. However, that's not guaranteed to be reliable. So this is an attempt to be more comprehensive.

It also means that, for next pages, fetching the missing pages happens a bit earlier, similar to how we try to preemptively load the first 30 items without waiting for `onEndReached` in the existing hack.

## Test Plan

This branch is based before https://github.com/bluesky-social/social-app/pull/4859 to show that it solves the problem independently from https://github.com/bluesky-social/social-app/pull/4859. However, it should also work when rebased to `main`.

You can try adding some console logs to verify the behavior:

https://github.com/user-attachments/assets/c2d2afdf-1ba9-4620-9ca4-06dcb2fe22a2

### DIFF
--- a/src/state/queries/notifications/feed.ts
+++ b/src/state/queries/notifications/feed.ts
@@ -126,6 +126,7 @@ export function useNotificationFeedQuery(opts?: {
   // doesn't return enough items, we're going to continue asking for more items.
   const lastItemCount = useRef(0)
   const wantedItemCount = useRef(0)
+  const autoPaginationAttemptCount = useRef(0)
   useEffect(() => {
     const {data, isLoading, isRefetching, isFetchingNextPage, hasNextPage} =
       query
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
This works for me so long as we're sure we're okay with no failsafe abort condition. If there's ever some case that causes all pages to come back empty, then it's going to run infinitely

haileyok (approved):
LGTM! Re: Paul's question, I think the `hasNextPage` here should keep us safe since we ensured the appview doesn't return a cursor when it shouldn't (after the bug we saw earlier today).

gaearon:
>If there's ever some case that causes all pages to come back empty, then it's going to run infinitely

So I think if there's an appview bug like this, it would be an infinite loop regardless if the tail is in the viewport. But it's fair that this wouldn't be a repeated issue if you scroll up. I'll add a failsafe.

PR #4864, opened 2024-08-01
Display second-to-last rather than second post in a slice

gaearon:
## Why

Extracted from https://github.com/bluesky-social/social-app/pull/4835. When we start showing replies in the context of the threads they're in, they'll follow the same structure. So this makes a similar change for the case where we already display the context.

Overall I think this makes the slice feel more coherent. OP post sets the context for the thread, and the last two posts are an unbroken chain. As opposed to how it behaves now where the last post can be pretty far disconnected from the initial point.

## Before

Screenshot 2024-08-01 at 02 37 34

## After

Screenshot 2024-08-01 at 02 38 08


mozzius (approved):

pfrazee (approved):
Sure

PR #4865, opened 2024-08-01
Show quote posts

mozzius:
dependant on https://github.com/bluesky-social/atproto/pull/2658

<table>
  <tr>
    <td>Screenshot 1</td>
    <td>Screenshot 2</td>
  </tr>
  <tr>
    <td>Screenshot 2024-08-01 at 11 27 02</td>
    <td>Screenshot 2024-08-01 at 11 27 09</td>
  </tr>
</table>


mozzius:
@haileyok I moved the PostLikedBy and PostRepostedBy screens over to the new screens folder, I think we should do that whenever we successfully ALF a screen

haileyok:
image

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -199,6 +199,11 @@ let PostThreadItemLoaded = ({
     return makeProfileLink(post.author, 'post', urip.rkey, 'reposted-by')
   }, [post.uri, post.author])
   const repostsTitle = _(msg`Reposts of this post`)
+  const quotesHref = React.useMemo(() => {
+    const urip = new AtUri(post.uri)
+    return makeProfileLink(post.author, 'post', urip.rkey, 'quotes')
+  }, [post.uri, post.author])
```
haileyok, line 205: We really ought to simplify this down to like `uriToRkey` or something. I run into this all the time and hate having to make an `AtUri` every time 🙃. Especially since we could just throw it right into the prop now w/ compiler.

I.e. `makeProfileLink(post.author, 'post', getRkey(uri), 'quotes')`

Minor thing though don't have to do it rn.

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -199,6 +199,11 @@ let PostThreadItemLoaded = ({
...
+  const quotesTitle = _(msg`Quotes of this post`)
```
haileyok, line 206: Weird that we did this too lol but also minor unless you feel like moving these down directly to the props.

mozzius: ehh I feel like we should keep it consistent

haileyok: Oh, yea unless you redo the whole thing don't worry about it. But I'm just wondering why we did it this way in the first place. (Definitely fine to leave it alone though)

haileyok (approved):
Nice, great work! I added two things as well:

- `findAllPostsInQueryData` for the `queueLike`/`queueRepost` shadows (so the like count will update when viewing a list of quotes)
- Call `onPost` in the composer whenever the app view is ready (after the quote count changes from the original count). Makes it feel a little more responsive when on the post thread.

PR #4866, opened 2024-08-01
Move theme controls to its own screen

mozzius:
I'm incrementally replacing the settings screen. I think a good first step is to reorganise some of the options that are just lying around inline.

<table>
  <tr>
    <td>Main settings screen</td>
    <td>Appearance page</td>
  </tr>
  <tr>
    <td>Simulator Screenshot - iPhone 15 - 2024-08-01 at 14 18 55</td>
    <td>Simulator Screenshot - iPhone 15 - 2024-08-01 at 14 19 05</td>
  </tr>
</table>

https://github.com/user-attachments/assets/fc952c72-e0a5-46e7-b16d-d2949de742b3


surfdude29:
I think that looks good as a first step 👌

I have one suggestion on the wording: I think `Dark mode theme` would be clearer than just `Theme`. If a user has `Dark mode` set to `System` and they're currently in light mode, it might seem a bit strange to them that they have options for `Dim` and `Dark` which would have no effect on the appearance.

pfrazee (approved):

PR #4868, opened 2024-08-01
Always limit Following replies to the people you follow

gaearon:
This removes the granular reply settings in favor of the current defaults. Only replies *between people you already follow* will be displayed in the Following timeline. (This is already the current default, and has been for many months — but this change applies it retroactively for people who might have changed this setting before the "followed only" filter existed.)

I'm also removing the like-based filter. The original purpose of the global like-based filter for replies in the Following feed was to avoid a flurry of replies overwhelming it. However, that was mostly necessary because there was no filtering by whether you follow both people in the exchange. Now that we have that filtering (and have had it as a default for many months), the like-based filter seems much less useful. It's unclear what value to pick as a user (3? 10? how do you choose?) and it creates an inconsistency in how people consume the following feed. It's easy to set it to some arbitrary value and then forget about it, causing replies between mutuals to seemingly randomly fail to appear.

This change does not affect other clients or how they choose to present the Following feed. They may still offer different filtering options, including filtering by likes or showing all replies. However, we've had a lot of feedback that bigger accounts replying to smaller accounts causes undesired attention to small accounts. Removing this option helps ensure that the default client does not show out-of-context convos in the Following feed unless you follow both people in the conversation.

Note there's still an outstanding bug that causes *some* replies to people you _don't_ follow to appear in your Following feed (this happens for multiple self-replies in a row). That bug will be fixed separately. I'm frontloading this change because it makes other things a bit easier and reduces the explosion of configurations.

Of course, you can also still turn off replies completely, if you want. That setting is still there.

## Before

Screenshot 2024-08-01 at 20 10 42

## After

Screenshot 2024-08-01 at 20 11 44

(The actual settings are identical to the screenshot above; they're just no longer configurable.)

## Test Plan

Set your account settings in prod to show _all replies_. Then check out this branch. Verify that you only see conversations between people you already follow. Then set the like threshold on prod really high. Verify that it isn't taken into account locally.

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -299,15 +299,7 @@ export class FeedTuner {
     return slices
   }
 
-  static thresholdRepliesOnly({
-    userDid,
-    minLikes,
-    followedOnly,
-  }: {
-    userDid: string
-    minLikes: number
-    followedOnly: boolean
-  }) {
+  static followedRepliesOnly({userDid}: {userDid: string}) {
     return (
       tuner: FeedTuner,
       slices: FeedViewPostsSlice[],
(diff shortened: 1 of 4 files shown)


pfrazee (approved):
Open to more feedback but I'm +1 to this

PR #4871, opened 2024-08-03
Show replies in context of their threads

gaearon:
Resubmit of https://github.com/bluesky-social/social-app/pull/4835.

## What

This rewrites our feed slicing logic which determines how posts (and in particular replies) appear in the feed.

It aims to fix a few problems:

- Replies now appear within the context of their parent threads. This avoids out-of-context replies.
- Replies from the same thread don't appear again in your Following feed. In other words, replies from friends "bump" threads but you only see the most recent reply to any given thread. This avoids a single thread dominating the timeline.
- Replies no longer get "torn off" from their parent conversations due to a pagination boundary.
- Replies from friends no longer get "torn off" from the original "author threads" that they belong to. In fact we're removing special casing of an "author thread". This means https://github.com/bluesky-social/social-app/pull/2914 can be closed.
- <s>Lists now use the Following feed settings (and no longer show replies from non-follows).</s>
  - We probably still want to do this but I'll send a separate PR later — not part of this change.
- "Hide replies" no longer hides reposted replies (since someone you follow wanted to treat it as a top-level).

## How

This builds on changes from https://github.com/bluesky-social/social-app/pull/4834, https://github.com/bluesky-social/social-app/pull/4864, https://github.com/bluesky-social/social-app/pull/4868, and https://github.com/bluesky-social/social-app/pull/4869. I've made the "feed slice" encapsulate more of the "building up items" logic so that it's able to reuse `parent` and `root` from the response despite those not having a `reply`.

What we show on the screen is now much closer to the data shape returned by the server — essentially, we show `[root, ..., parent, reply]` for each item. In Following feeds, feedgens, and lists, we deduplicate by the `root` so that each thread acts as a forum thread — you only see it once (but replies from friends bump it). In author feeds (Posts and Replies), we use a different deduplication strategy: you want to see all relevant posts, so we instead hide contiguous leading sequences of posts that you have already seen. See comments in code for specific scenarios.

## Examples

Here's a broken up thread (before):

1_before

Here's how this PR fixes it (after):

1_after

Here's an out-of-context reply (before):

2_before

Here's how this PR adds context to it (after):

2_after

Here's a reply from somebody I follow to somebody I _don't_ follow (before):

3_before

There is no "after" for this example because we'd no longer show the above replies in Following in this PR.

Here is a "torn" reply to another person's thread (before):

5_before

Here is how this PR fixes it:

5_after

## Test Plan

Here's some things I encourage you to test:

- Following feed
  - Different settings (hide replies, hide reposts, hide quotes)
  - Mergefeed on and off
  - Move Following to first position on an account that doesn't follow anyone, verify Discover fallback works
  - Verify you only see people you follow for each root/parent/reply
  - Replies from a seen thread should not appear below
  - Repost of a non-leaf reply should not break the thread, both should remain visible
  - Reposts of same post by different follow should remain deduped
  - Replies with deleted and blocked parents or roots should not show up
  - Replies with blocked grandparents should not show up
  - When parent and grandparent authors are different in a >=3 post thread, "in reply to" is displayed
  - "Has new" polling and button works
- Starter Packs
  - Posts tab should work like Following
- Profile
  - Should deduplicate individual posts but should not deduplicate threads
  - Contiguous leading blocks should get deduplicated top-down (threads like `[ABC], [ABD], [AEF]` become `[ABC], [D], [EF]`). You should see "in reply to" for `[D]` and `[EF]` since their parent/roots got deduplicated.
  - Replies with deleted and blocked parents or roots _should_ (!) show up for Replies tab
  - Replies with blocked grandparents _should_ (!) show up for Replies tab
- Lists
  - Should work like Profile (deduplicate by leading posts, but not by threads)
  - We'll probably change it to work more like Following later, but out of scope of this PR
- Feedgens
  - Should work like Profile (deduplicate by leading posts, but not by threads)


src/state/queries/post-feed.ts
```diff
@@ -313,53 +313,22 @@ export function usePostFeedQuery(
                   const feedPostSlice: FeedPostSlice = {
```
gaearon, line 313: This is actually getting close to the shape of the "feed slice" class so I think we can unify these concepts in the future and maybe represent moderation as a feed tune.

src/lib/api/feed-manip.ts
```diff
@@ -127,168 +167,134 @@ export class FeedViewPostsSlice {
...
-          soonToBeSeenUris.add(uri)
+      // Most feeds dedupe by thread, but the author feed doesn't because it aims to make
+      // it possible to find all author's posts. This logic adds post-level dedupe for it.
+      for (let i = 0; i < slice.items.length; i++) {
+        const item = slice.items[i]
+        if (this.seenUris.has(item.post.uri)) {
+          if (i === 0) {
+            // Omit contiguous seen leading items.
+            // For example, [A -> B -> C], [A -> D -> E], [A -> D -> F]
+            // would turn into [A -> B -> C], [D -> E], [F].
```
gaearon, line 214: This heuristic is interesting and worth noticing. It only kicks in for author feeds because other feeds already get deduped by the root.

src/lib/api/feed-manip.ts
```diff
@@ -127,168 +167,134 @@ export class FeedViewPostsSlice {
...
+            slice.items.splice(0, 1)
+            i--
+          }
+          if (i === slice.items.length - 1) {
+            // If the last item in the slice was already seen, omit the whole slice.
+            // This means we'd miss its parents, but the user can "show more" to see it.
+            // For example, [A ... E -> F], [A ... D -> E], [A ... C -> D], [A -> B -> C]
+            // would get collapsed into [A ... E -> F], with B/C/D considered seen.
```
gaearon, line 222: This one is interesting too.

src/lib/api/feed-manip.ts
```diff
@@ -127,168 +167,134 @@ export class FeedViewPostsSlice {
...
-        if (item1.containsUri(item2.items[0].post.uri)) {
-          slices.splice(j, 1)
-          j--
-        }
+      const rootUri = slices[i].rootUri
+      if (!slices[i].isRepost && tuner.seenRootUris.has(rootUri)) {
+        slices.splice(i, 1)
+        i--
+      } else {
+        tuner.seenRootUris.add(rootUri)
```
gaearon, line 297: This is stateful which might be a problem for the "dry run". I'll look if we can remove dry runs as a concept — seems very fragile.

gaearon:
Pushed a commit that turns off thread deduping for feedgens. The motivation is that feedgens are already able to do thread deduping if they want to. So if a feedgen is serving a bunch of replies from the same thread, it must have a reason to do so and we shouldn't hide those. Instead we'll rely on contiguous leading post deduping for feedgens, same as for author feeds.

gaearon:
Pushed a revert to the Lists part of the deal. We probably still want to do it but decoupling from this already chunky change.

pfrazee (approved):
My review is pretty high level but the logic checks out!

PR #4872, opened 2024-08-03
[Persisted] Fork web and native, make it synchronous on the web

gaearon:
The way our persisted state syncing works does not seem completely reliable. At the very least, it's quite noisy — for example, merely opening a new tab causes all other tabs to sync to it (though nothing has really changed). Switching accounts causes tabs to send messages to each other — likely due to all of them trying to resume the session. This would be easier to debug if the `onUpdated` handlers only fired when there's an actual change to the slice of the state they cared about, but currently they're being called for every write from every tab.

I'd like to start simplifying this, but the cross-platform abstraction layering makes it hard to modify or think about. So this PR is just to start narrowing down the abstraction per platform. There's a few key ideas:

- Web is the only place where we need to sync between tabs or worry about concurrent storage access. So it would be nice to scope down all the related logic (like reacting to BroadcastChannel messages) to web only.
- Web offers synchronous local storage access. Synchronous access is a lot easier to think about when worrying about concurrency. So it's nice if we can get rid of internal `await`s in the logic for the web fork. (In the future, we probably should [use something like this](https://github.com/mrousavy/react-native-mmkv) for native too, but we're not going to do this now.)

In this PR, I remove the legacy storage migration code (it's been long enough that the tokens would've expired anyway), and then fork the implementation into native and web. I'm also removing `await`s in the internal implementation code from the web version. The port from `AsyncStorage` to `localStorage` for web is very straightforward — I checked with the source that the only thing `AsyncStorage` did on web is wrapping into Promises, which we explicitly do not need. There are no other changes for now, so this is mostly a refactor.

I'll layer on some more substantive changes in follow-up PRs.

## Test Plan

Change some settings, observe them propagate across tabs like before.

Read the code commit-by-commit and verify it's a straightforward transformation.


estrattonbailey (approved):
Hot dog ✨

PR #4873, opened 2024-08-04
[Persisted] Fix the race condition causing clobbered writes between tabs

gaearon:
Stacked on #4872.

This removes defensive programming in the original code, extracts some shared logic between platforms, and fixes a race condition that caused potential clobbered concurrent writes between tabs. I recommend to review commit by commit:

- https://github.com/bluesky-social/social-app/commit/2e70c0c95ba8d47dcca6c95d6206517d349375c8: This removes the `setTimeout(..., 0)` around the `broadcast.postMessage` call. I believe its premise is incorrect. The premise of this wrapper is that the other tab might read stale data because the broadcast message can arrive before the local storage update is flushed. I find this difficult to believe given that the `localStorage` contract on the web is synchronous while `postMessage` is _definitely_ asynchronous and definitely can lag behind under pressure. In other words, the inverse of what the comment says is likely to be true unless I'm missing something extraordinary (which is always possible). I think this wrapper was likely added because it makes a different race condition less common (but not impossible), which I describe in the last commit (see below). For now let's remove this wrapper.
-  https://github.com/bluesky-social/social-app/commit/08f7c3f298f8921e147a4132e4dab5dfb467616d?w=1: This removes the overly broad `try`/`catch`es around the parts of the code that we do not expect to throw. There are three kinds of things that can throw: `schema.parse()`, `JSON.parse()`, `JSON.stringify()`, and `localStorage`/`AsyncStorage` methods. Those remain wrapped in `try`/`catch`es that deal with those specific errors.
- ab3a1596d250d1115a6ad7c9b97417faa673e9f3: We do not actually need to write defaults into the storage on first initialization. It is enough to write them on first attempt to write. While writing this, I noticed that `_state` can end up getting set to `defaults`. This is no bueno because we mutate `_state` directly in `write` but we wouldn't want to mutate the `defaults`. We could fix this in a few different ways but I opted to just shallowly copy `_state` on write since we already treat values inside as immutable.
- https://github.com/bluesky-social/social-app/commit/24f0e8c55d9c068376f9463fdd5b5d0550c1ce9b: Pure refactor. Extracts common code into a `tryParse(obj): Schema | undefined` utility.
- https://github.com/bluesky-social/social-app/commit/d55e9f84beab8d60bc305c589d068d1642277417: Pure refactor. Changes `tryParse` to take a string, moves `JSON.parse` inside it. 
- https://github.com/bluesky-social/social-app/commit/fe59dd025dab511e50a03c36a776e51da13675a9: Introduces `tryStringify(value): string | undefined` that mirrors `tryParse`. This lets us separate out the parts we want to report to Sentry (e.g. invalid JSON or failed schema validation) from parts we don't (`localStorage` on the web is expected to fail in private mode). `writeToStorage` and `readToStorage` become thin storage wrappers.
- https://github.com/bluesky-social/social-app/commit/acb4094e4a950c2f0ab7c56fa12ad478782e0c1b: Adds a bit of memoization to the web version of `readToStorage` because I'll call it more often.
- c82fe5bea45f330f714e3beabab0ea85ff2f7269: Finally, the actual bugfix! If a tab writes to the storage but another tab hasn't received the event, the write from the second tab will overwrite the data put into the storage by the first tab. While there's no way to avoid the conflict if the keys are overlapping, we can easily resolve the conflict when they're not — we just need to _read_ the data from the storage _before_ applying the write (and we need to apply the write _on top of_ the fresh data).

## Test Plan

Normal things you'd do when testing storage.

For testing the last part of the fix, you can apply this diff (which _disables_ the fix):

```difff
diff --git a/src/state/persisted/index.web.ts b/src/state/persisted/index.web.ts
index d71b59096..35aac04d9 100644
--- a/src/state/persisted/index.web.ts
+++ b/src/state/persisted/index.web.ts
@@ -43,7 +43,16 @@ export async function write<K extends keyof Schema>(
   if (next) {
     // The storage could have been updated by a different tab before this tab is notified.
     // Make sure this write is applied on top of the latest data in the storage as long as it's valid.
-    _state = next
+    // _state = next
+    if (JSON.stringify(next) !== JSON.stringify(_state)) {
+      console.error('oh no')
+      console.error(' [memory]    invites:', _state?.invites.copiedInvites)
+      console.error(' [storage]    invites:', next?.invites.copiedInvites)
+      console.log()
+      console.error(' [memory]    posts:', _state?.hiddenPosts)
+      console.error(' [storage]    posts:', next?.hiddenPosts)
+
+    }
     // Don't fire the update listeners yet to avoid a loop.
     // If there was a change, we'll receive the broadcast event soon enough which will do that.
   }
@@ -51,6 +60,7 @@ export async function write<K extends keyof Schema>(
     ..._state,
     [key]: value,
   }
+  console.log('writing', key, value)
   writeToStorage(_state)
   broadcast.postMessage({event: UPDATE_EVENT})
 }
@@ -75,6 +85,9 @@ async function onBroadcastMessage({data}: MessageEvent) {
   if (typeof data === 'object' && data.event === UPDATE_EVENT) {
     // read next state, possibly updated by another tab
     const next = readFromStorage()
+    console.log('-- read from broadcast')
+    console.log('    invites:', next?.invites.copiedInvites)
+    console.log('    posts:', next?.hiddenPosts)
     if (next) {
       _state = next
       _emitter.emit('update')
diff --git a/src/state/preferences/hidden-posts.tsx b/src/state/preferences/hidden-posts.tsx
index 2c6a373e1..0e9bc7859 100644
--- a/src/state/preferences/hidden-posts.tsx
+++ b/src/state/preferences/hidden-posts.tsx
@@ -44,6 +44,22 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   React.useEffect(() => {
+    if (location.search === '?foo=') {
+      setInterval(() => {
+        persisted.write('invites', {
+          copiedInvites: [
+            Math.random().toString()
+          ]
+        })
+      }, 100)
+    } else {
+      setInterval(() => {
+        persisted.write('hiddenPosts', [
+          Math.random().toString()
+        ])
+      }, 100)
+    }
+
     return persisted.onUpdate(() => {
       setState(persisted.get('hiddenPosts'))
     })
```

Then open two tabs:

- http://localhost:19006/
- http://localhost:19006/?foo=

One tab will attempt to write to `invites` every 100ms, another tab will attempt to write to `hiddenPosts` every 100ms.

Observe that they clobber each other's writes:

Screenshot 2024-08-04 at 03 00 28

Of course, this just shows the symptom so you'd have to trust that the fix works. (It trivially works by making the `JSON.stringify(next) !== JSON.stringify(_state)` condition impossible to happen but I'm struggling to see how to demonstrate this more clearly.) So if this is not convincing, I can think a bit more, but I hope the idea should be clear.

### DIFF
--- a/src/state/persisted/index.ts
+++ b/src/state/persisted/index.ts
@@ -12,17 +12,11 @@ const BSKY_STORAGE = 'BSKY_STORAGE'
 let _state: Schema = defaults
 
 export async function init() {
-  try {
-    const stored = await readFromStorage()
-    if (!stored) {
-      await writeToStorage(defaults)
-    }
-    _state = stored || defaults
-  } catch (e) {
-    logger.error('persisted state: failed to load root state from storage', {
-      message: e,
-    })
+  const stored = await readFromStorage()
+  if (!stored) {
+    await writeToStorage(defaults)
   }
+  _state = stored || defaults
 }
 init satisfies PersistedApi['init']
 
@@ -35,14 +29,8 @@ export async function write<K extends keyof Schema>(
   key: K,
   value: Schema[K],
 ): Promise<void> {
-  try {
-    _state[key] = value
-    await writeToStorage(_state)
-  } catch (e) {
-    logger.error(`persisted state: failed writing root state to storage`, {
-      message: e,
-    })
-  }
+  _state[key] = value
+  await writeToStorage(_state)
 }
 write satisfies PersistedApi['write']
 
@@ -61,13 +49,26 @@ export async function clearStorage() {
 clearStorage satisfies PersistedApi['clearStorage']
 
 async function writeToStorage(value: Schema) {
-  schema.parse(value)
-  await AsyncStorage.setItem(BSKY_STORAGE, JSON.stringify(value))
+  try {
+    schema.parse(value)
+    await AsyncStorage.setItem(BSKY_STORAGE, JSON.stringify(value))
+  } catch (e) {
+    logger.error(`persisted state: failed writing root state to storage`, {
+      message: e,
+    })
+  }
 }
 
 async function readFromStorage(): Promise<Schema | undefined> {
-  const rawData = await AsyncStorage.getItem(BSKY_STORAGE)
-  const objData = rawData ? JSON.parse(rawData) : undefined
+  let objData
+  try {
+    const rawData = await AsyncStorage.getItem(BSKY_STORAGE)
+    objData = rawData ? JSON.parse(rawData) : undefined
+  } catch (e) {
+    logger.error('persisted state: failed to load root state from storage', {
+      message: e,
+    })
+  }
 
   // new user
   if (!objData) return undefined
--- a/src/state/persisted/index.ts
+++ b/src/state/persisted/index.ts
@@ -13,10 +13,9 @@ let _state: Schema = defaults
 
 export async function init() {
   const stored = await readFromStorage()
-  if (!stored) {
-    await writeToStorage(defaults)
+  if (stored) {
+    _state = stored
   }
-  _state = stored || defaults
 }
 init satisfies PersistedApi['init']
 
@@ -29,7 +28,10 @@ export async function write<K extends keyof Schema>(
   key: K,
   value: Schema[K],
 ): Promise<void> {
-  _state[key] = value
+  _state = {
+    ..._state,
+    [key]: value,
+  }
   await writeToStorage(_state)
 }
 write satisfies PersistedApi['write']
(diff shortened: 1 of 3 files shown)


gaearon:
> In other words, the inverse of what the comment says is likely to be true unless I'm missing something extraordinary (which is always possible).

Note that it _is_ possible I'm wrong about this one. In anecdotal testing I saw the update seeming to fail to propagate a few times. But I haven't had logging on to see what the sequence of events was. Maybe the write is synchronous within the tab but it's not seen by the other process if we don't let it spin first? If that's the problem it would be nice to have a reduced repro showing this browser behavior — I haven't seen it mentioned elsewhere and if this is a problem, I'm not sure `0` is solid.

This definitely requires some caution and investigation before merging.

gaearon:
Ah nvm, I think this might have just been Fast Refresh acting up due to the module scope variable after save.

estrattonbailey (approved):

PR #4877, opened 2024-08-05
Add `PlatformInfo` module

haileyok:
## Why

There's some various platform preferences we need from time to time (such as currently reduced motion preferences, or later font scale preferences, wifi status, etc). This is a little module that lets us grab values like that.

There's a patch for Reanimated we are using right now to get this value, though every time we change our Reanimated version we have to update the patch. It would be nice to get rid of that as well.

## Test Plan

I've added a button to the story book that you can get grabbing this value on various platforms with.

### Web

https://github.com/user-attachments/assets/9cf5521f-7325-448b-9e90-27a3fee5c67d

### Android

https://github.com/user-attachments/assets/7d10405a-6f41-4bf3-848f-11dd4041fb28

### iOS

https://github.com/user-attachments/assets/d1a42cea-2f20-4f82-8b6b-9a67cff6ac4c

### DIFF
--- a/jest/jestSetup.js
+++ b/jest/jestSetup.js
@@ -95,3 +95,13 @@ jest.mock('expo-application', () => ({
   nativeApplicationVersion: '1.0.0',
   nativeBuildVersion: '1',
 }))
+
+jest.mock('expo-modules-core', () => ({
+  requireNativeModule: jest.fn().mockImplementation(moduleName => {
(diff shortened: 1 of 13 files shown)


mozzius (approved):
Tested on all 3 platforms

gaearon (approved):

PR #4878, opened 2024-08-05
Detached QPs and hidden replies

estrattonbailey:
Adds quote gates, detached quotes, and hidden replies.


src/components/icons/X.tsx
estrattonbailey, line 1: Silly, will remove

src/state/queries/postgate/index.ts
```diff
@@ -0,0 +1,295 @@
...
+          repo: urip.host,
+          collection: POSTGATE_COLLECTION,
+          rkey: urip.rkey,
+        }),
+    )
+
+    if (data.value && AppBskyFeedPostgate.isRecord(data.value)) {
+      return data.value
+    } else {
+      return undefined
```
estrattonbailey, line 66: Could return null here instead (and at threadgate) and simplify some type checks

haileyok:
So far things are looking good:

- Had to push a commit to fix toast import, just be sure to pull next time you check out
- Wonder if we want to put your own reply behind `Show hidden replies` whenever the OP creator hides your post. I don't necessarily think that's a problem - in fact maybe that's preferred - but just something to think about
- The `Hide post for me` button might better if we say `Hide reply for me`, so it lines up with `Hide reply for everyone`
- Should check how difficult it is to either:
  - Have the respost/quote dialog use the new ALF system (swear we did this)
  - Make the disabled button grayed out

haileyok:
> @estrattonbailey
[Use a set for faster lookup](https://github.com/bluesky-social/social-app/pull/4878/commits/dc11707e54615262165b1c6f2e3b5f1e7d80bb89)

great minds think alike 🍻

src/view/com/composer/threadgate/ThreadgateBtn.tsx
```diff
@@ -59,16 +67,19 @@ export function ThreadgateBtn({
...
         </Button>
       </Animated.View>
-      <ThreadgateEditorDialog
+      <PostInteractionSettingsControlledDialog
         control={control}
-        threadgate={threadgate}
-        onChange={onChange}
+        onSave={() => {
+          control.close()
+        }}
```
haileyok, line 78: these always suss me out. i think its fine but lets make sure it doesnt break android back button.

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -77,7 +80,9 @@ export function PostThreadItem({
   overrideBlur: boolean
   onPostReply: (postUri: string | undefined) => void
   hideTopBorder?: boolean
+  threadgateRecord?: AppBskyFeedThreadgate.Record
 }) {
+  // TODO should load this for every post, ugh
```
haileyok, line 85: just making so we dont forget

src/view/com/posts/FeedItem.tsx
```diff
@@ -381,23 +400,63 @@ let FeedItemInner = ({
...
   const pal = usePalette('default')
   const {_} = useLingui()
+  const {currentAccount} = useSession()
   const [limitLines, setLimitLines] = useState(
     () => countLines(richText.text) >= MAX_POST_LINES,
   )
+  const {uris: hiddenReplyUris, recentlyUnhiddenUris} =
+    useThreadgateHiddenReplyUris()
+  const additionalPostAlerts: AppModerationCause[] = React.useMemo(() => {
+    const isPostHiddenByHiddenReplyCache = hiddenReplyUris.includes(post.uri)
```
haileyok, line 428: didn't see where this comes from but can this be a set too?

src/view/com/util/forms/PostDropdownBtn.tsx
```diff
@@ -104,17 +119,46 @@ let PostDropdownBtn = ({
...
+  const {uris: hiddenReplies, recentlyUnhiddenUris} =
+    useThreadgateHiddenReplyUris()
+
   const postUri = post.uri
   const postCid = post.cid
   const postAuthor = post.author
+  const quoteEmbed = React.useMemo(() => {
+    if (!currentAccount || !post.embed) return
+    return getMaybeDetachedQuoteEmbed({
+      viewerDid: currentAccount?.did,
```
haileyok, line 136: no need for `?` here, or we can remove `!currentAccount` above

src/view/com/posts/FeedItem.tsx
estrattonbailey: Oh yeah good call, might as well. It's an account-scoped context value, for optimistic state.

src/lib/api/index.ts
```diff
@@ -244,54 +253,43 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
 
   try {
     // TODO: this needs to be batch-created with the post!
```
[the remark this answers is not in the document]
estrattonbailey: I think we'd be looking at doing an `applyWrites` call, like in the thread composer PR. Would be a good win, but we don't need to address it right now. The user is notified if these writes fail, and they can edit the gating immediately after posting.

src/lib/moderation/useModerationCauseDescription.ts
```diff
@@ -111,6 +114,18 @@ export function useModerationCauseDescription(
         description: _(msg`You have hidden this post`),
       }
     }
+    if (cause.type === 'reply-hidden') {
+      const isYou = currentAccount?.did === cause.source.did
```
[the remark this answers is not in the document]
estrattonbailey: heh yeah totally

src/lib/api/index.ts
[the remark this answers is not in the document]
haileyok: ah nice, makes sense.

src/view/com/post-thread/PostThreadItem.tsx
estrattonbailey: oh lawl this is old, long done

src/view/com/composer/threadgate/ThreadgateBtn.tsx
estrattonbailey: Yep cool, let's leave this one unresolved for now

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -0,0 +1,539 @@
...
+  style?: StyleProp<ViewStyle>
+  disabled?: boolean
+}) {
+  const t = useTheme()
+  return (
+    <Button
+      disabled={disabled}
+      onPress={onPress}
+      label={label}
+      accessibilityHint="Select this option"
```
surfdude29, line 471: Would it be helpful to have this `accessibilityHint` marked for localization? (Tbh I'm unclear on what the role of `accessibilityHint`s is generally which is why I ask :)

src/lib/api/index.ts
```diff
@@ -244,54 +253,43 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
...
+    })
   } catch (e: any) {
-    console.error(`Failed to create threadgate: ${e.toString()}`)
+    logger.error(`Failed to create threadgate`, {
+      context: 'composer',
+      safeMessage: e.message,
+    })
     throw new Error(
-      'Post reply-controls failed to be set. Your post was created but anyone can reply to it.',
+      'Failed to save post interaction settings. Your post was created but users may be able to interact with it.',
```
surfdude29, line 270: Would it be feasible to mark this error string for localization?

src/lib/api/index.ts
```diff
@@ -244,54 +253,43 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
...
+        post: res.uri,
+      },
+    })
+  } catch (e: any) {
+    logger.error(`Failed to create postgate`, {
+      context: 'composer',
+      safeMessage: e.message,
+    })
+    throw new Error(
+      'Failed to save post interaction settings. Your post was created but users may be able to interact with it.',
```
surfdude29, line 290: Same question here

src/components/WhoCanReply.tsx
```diff
@@ -186,79 +160,84 @@ function WhoCanReplyDialog({
...
+  settings: ThreadgateAllowUISetting[]
+  embeddingDisabled: boolean
 }) {
+  const {_} = useLingui()
   return (
     <Dialog.Outer control={control}>
       <Dialog.Handle />
-      <WhoCanReplyDialogInner post={post} settings={settings} />
+      <Dialog.ScrollableInner
+        label={_(msg`Who can reply dialog`)}
```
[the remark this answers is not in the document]
estrattonbailey: Ah yeah we can do better here, good call

src/lib/api/index.ts
estrattonbailey: Love that suggestion. I'll make a separate PR to clean some of this up and include these once this PR is in 👍

haileyok:
Lets

Go

🥳

haileyok (approved):

src/lib/api/index.ts
surfdude29: @estrattonbailey Afaik the user-facing strings in `src/lib/api/index.ts` are some of the only ones in the app that can't yet be localized, so when you've got a moment sometime that PR would be appreciated :)

PR #4880, opened 2024-08-06
RFC: Show more replies in Following (different heuristic)

gaearon:
Alternative to the heuristic https://github.com/bluesky-social/social-app/pull/4871 based on community feedback.

This fixes the double-reply loophole and actually _opens up_ the default Following heuristic to be a bit more broad than the current default one on the website. We'd include not only replies _between_ people you follow (which is the current default), but _also_ replies where you follow only the replier and the thread OP, _or_ only the replier and the grandparent. In other words, we'd show the reply if there's *at least one other person* you'd recognize looking at the context above that reply.

In practice this means you'd see banter where you don't follow _some_ people as long as there's _one more reason_ to display them (e.g. you follow the OP, the parent, or the grandparent).

For example, I don't follow Jeremy but this still surfaces Stephanie's reply to Jeremy (because I follow Eustace):

Screenshot 2024-08-06 at 06 28 34

Here, I don't follow Anna, but the thread has Amy _and_ Rahaeli (both of whom I follow) so I see it too:

Screenshot 2024-08-06 at 06 30 08

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -25,6 +25,13 @@ type FeedSliceItem = {
   isParentBlocked: boolean
 }
 
+type AuthorContext = {
+  author: AppBskyActorDefs.ProfileViewBasic
+  parentAuthor: AppBskyActorDefs.ProfileViewBasic | undefined
+  grandparentAuthor: AppBskyActorDefs.ProfileViewBasic | undefined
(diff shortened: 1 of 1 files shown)


mozzius:
I like this. Seems like a good compromise. A one-off reply won't be shown, but if there's a whole subthread it probably will be.

pfrazee (approved):
Sure let's give it a shot

PR #4882, opened 2024-08-06
Show own replies before follows' replies in threads

gaearon:
We currently put OP replies first, then people you follow (for an option that's true by default), then other replies. However, we're not doing anything with _your own_ replies which [makes them a bit hard to find](https://bsky.app/profile/citizenkate.bsky.social/post/3kz2qmt5eac2k) in bigger threads.

This tweaks the heuristic so that your own replies go right after the OP (but above your follows).

## Test Plan

Reply to a thread, see your reply below the OP.

https://github.com/user-attachments/assets/794c7262-4bc4-474b-9602-dabc0c4f507b


pfrazee (approved):
Oh yeah def

PR #4883, opened 2024-08-06
Tweak feed manip to show cases of A -> B without further children

haileyok:
Fixes where situations like `[A] -> [B]` wouldn't show up, since we were returning false if `grandparentAuthor` was `null`.


gaearon (approved):
lg

PR #4886, opened 2024-08-07
[Video] Download videos

haileyok:
## Why

We want for users to be able to download videos that are uploaded to Bluesky within the app, however, we are very adverse to bloating the app bundle size (possibly doubling it) just to add `ffmpeg`. IT also seems that native APIs don't manage exporting HLS streams to MP4 (at least, not on iOS!)

What a conundrum!

## How

With probably one of the most cursed pieces of code I have ever written, here we are adding a page to `bsky.app` (`/video-download`) that accepts a `videoUrl` GET parameter. This should be the URL of an HLS stream master playlist (might in general, I don't know. It works with our playlist files). Once the page has been opened, `ffmpeg-wasm` will be loaded and each segment inside of the playlist (the `.ts` files) will be downloaded and pieced together with `ffmpeg` into a new MP4.

Next, we get a base64 URL of that file that can be downloaded. On native, we have to "trick" the `WebView` into receiving a "download file" event. To do that, we simply set the data URL as the `href` of an anchor tag, with a little ref callback that calls `.click()` on the element. This sends the data URL to the `WebView` download event handler. At this point, we would be finished on web. The user's browser will simply download the file to their computer. However, on native there's a bit more.

Generally, we would probably just pass the base64 url/string back to JS, but since these strings are _massive_ in size, attempting to send this over the bridge is _quite_ slow (freeze ups of around 2-3 seconds in iOS simulator). As a result, it's easier for us to just handle the download on the native side and save it to a temporary file - that of which we can pass back to the JS side to save to the user's library with `saveToMediaLibraryAsync`.

## Test Plan

For now, there isn't any real UI for this thing. There is a simple tester however inside of the storybook. You should be able to click on the "Test video download" button and observe in the console the download progress being logged. You should also a prompt for media library permissions (after the video has downloaded). If you accept, the video should be saved to your camera roll.


package.json
```diff
@@ -59,6 +59,9 @@
     "@emoji-mart/react": "^1.1.1",
     "@expo/html-elements": "^0.4.2",
     "@expo/webpack-config": "^19.0.0",
+    "@ffmpeg/core": "^0.12.6",
```
mozzius, line 62: do we need this one? looks like it's imported via `unpkg`

mozzius (approved):
Cursed, but works well!

package.json
haileyok: oh probably not.

PR #4888, opened 2024-08-08
Keep interstitial fresh on refresh

gaearon:
Refreshing Discover doesn't currently refresh the interstitial. This updates the content by keying it by first page's fetch time.

## Test Plan

Scroll down.

Screenshot 2024-08-08 at 05 53 09

Press Discover again to refresh. Scroll down again.

Observe different order/people:

Screenshot 2024-08-08 at 05 53 00


haileyok (approved):
yup, verified refreshes now

PR #4889, opened 2024-08-08
Use the suggested actors in the follow suggestion interstitial after a follow

gaearon:
This incorporates a signal: if you follow someone, we'll show their "suggested by actor" results next time you see the interstitial during this session.

## Test Plan

Verify people from the endpoint show up in the suggestions after next Discover refresh.

Note the result is better under the new gate, but the logic should work either way.

https://github.com/user-attachments/assets/9e54a659-7a67-4297-8505-5c35a196f508

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -222,6 +222,7 @@ export function useProfileFollowMutationQueue(
   logContext: LogEvents['profile:follow']['logContext'] &
     LogEvents['profile:unfollow']['logContext'],
 ) {
+  const agent = useAgent()
   const queryClient = useQueryClient()
   const did = profile.did
   const initialFollowingUri = profile.viewer?.following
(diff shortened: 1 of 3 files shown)


src/state/queries/profile.ts
```diff
@@ -253,6 +254,20 @@ export function useProfileFollowMutationQueue(
...
+          .getSuggestedFollowsByActor({
+            actor: did,
+          })
+          .then(res => {
+            const dids = res.data.suggestions
+              .filter(a => !a.viewer?.following)
+              .map(a => a.did)
+              .slice(0, 8)
+            userActionHistory.followSuggestion(dids)
+          })
```
estrattonbailey, line 269: Should we add a catch here and ignore the error?

estrattonbailey (approved):
Love this!

src/state/queries/profile.ts
gaearon: I think it'll be fine either way since it's async — we'll see an unhandled rejection at most (which should be fine on both web and RN). Is that correct?

gaearon: I'd prefer not to ignore in case we introduce a coding mistake somewhere. E.g. if there's a crash, I'd rather it get surfaced. And if we add a catch() that just calls the logger, that seems equivalent to not adding it at all. Lmk if this is wrong though

gaearon: Note that since it's not awaited, a network error wouldn't propagate to the calling code

PR #4892, opened 2024-08-08
Experiment: follow suggestions on profiles

estrattonbailey:
Adds an experiment to auto-show the follow suggestions on the profile page after you follow the viewed profile.

Also, the API response for the new follow suggestions endpoint is slow to filter out users you've recently followed, so this PR adds a quick filter to the client to remove any followed users. We don't currently limit or slice the response, so it contains enough results that we shouldn't need to worry about filtering out every result.


gaearon:
Not able to scroll on Android over the card — is this fixable? It's fine on iOS.

https://github.com/user-attachments/assets/3ccc8585-e0ee-422c-926f-97e25892b9a6

PR #4896, opened 2024-08-08
Remove native_pwi_disabled

estrattonbailey:


src/view/com/auth/LoggedOut.tsx
```diff
@@ -98,10 +96,7 @@ export function LoggedOut({onDismiss}: {onDismiss?: () => void}) {
               }}
             />
           </Pressable>
-        ) : isNative &&
-          !hasSession &&
-          isFirstScreen &&
-          !gate('native_pwi_disabled') ? (
+        ) : isNative && !hasSession && isFirstScreen ? (
```
gaearon, line 99: Doesn't this whole branch get `false`y now?

gaearon: `gate('native_pwi_disabled')` becomes `true`, so `!gate('native_pwi_disabled')` becomes `false`, so `isNative && !hasSession && isFirstScreen && !gate('native_pwi_disabled')` should be `false` i think

gaearon (approved):
lgtm

PR #4897, opened 2024-08-08
Remove unnecessary state update for reply gate

gaearon:
We currently set state on mount in post thread. This is only necessary because the mobile reply prompt is located above the post in the hierarchy so we don't have access to the post itself from the parent component.

This moves the mobile inline prompt into the main post thread component. This lets us remove this state and calculate it during render. This will also enable another simplification that I'll stack on top of this PR.

## Test Plan

Mobile "Write your reply" still shows up on all platforms in the right place.

Screenshot 2024-08-08 at 16 42 27
Screenshot 2024-08-08 at 16 42 42
Screenshot 2024-08-08 at 16 52 48

Reply gating still works.

Screenshot 2024-08-08 at 16 53 10


src/view/com/post-thread/PostThread.tsx
```diff
@@ -1,9 +1,12 @@
...
-import {useWindowDimensions, View} from 'react-native'
+import React, {useRef} from 'react'
+import {StyleSheet, useWindowDimensions, View} from 'react-native'
 import {runOnJS} from 'react-native-reanimated'
+import Animated from 'react-native-reanimated'
+import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {AppBskyFeedDefs} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
+import {clamp} from 'lodash'
```
mozzius, line 9: lodash is a transient dependency from rn-fetch-blob that we should avoid - it should be `import clamp from 'lodash.clamp'`

mozzius (changes requested):

src/view/com/post-thread/PostThread.tsx
mozzius: Oh, just released it's not new, it's just moved from another file. We should see about removing that transient dependency though, I have a PR here: https://github.com/bluesky-social/social-app/pull/4553

gaearon: yea but good callout

mozzius (approved):
Looks good!

PR #4898, opened 2024-08-08
Move onPressReply into child component

gaearon:
Similar in spirit to https://github.com/bluesky-social/social-app/pull/4897.

The parent screen component doesn't really "know" about the inner state so we had to read the thread from the cache. This is a bit convoluted cause the thread is "right there" in the child component. The only reason we did it was because the mobile prompt used to be at the parent component level. We've moved it to the child, letting us move the handler too.

I've also renamed `ComposePrompt` and moved the file to make it clearer it's only used in post-thread.

## Test Plan

Verify can still post via the mobile prompt. Verify it hits the `refetch` breakpoint when needed. New post shows up below.

https://github.com/user-attachments/assets/c8a92f74-dbe6-4417-8d8c-9fc138607c4a


haileyok (approved):
yep

PR #4899, opened 2024-08-08
Fix overflow on posts

mozzius:
Side effect of #4732 - I had to remove the overflow hidden on posts. This PR adds overflow hidden to more targeted areas within the post

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Screenshot 2024-08-08 at 16 44 56</td>
    <td>Screenshot 2024-08-08 at 16 44 38</td>
  </tr>
</table>

While I'm at it, I limited overflow in the profile area

<table>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Screenshot 2024-08-08 at 16 40 54</td>
    <td>Screenshot 2024-08-08 at 17 10 10</td>
  </tr>
</table>

As a bonus, I fixed overflow in displaynames in the `<ComposerReplyTo>` component

# Test plan

Test zalgo text in all contexts in which you can see a post - feed, profile feed, notifications, thread (main and reply) etc
Helpful post:
https://bsky.app/profile/samuel.bsky.team/post/3kz7qimdhjd27

### DIFF
--- a/src/components/moderation/ContentHider.tsx
+++ b/src/components/moderation/ContentHider.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {StyleProp, StyleSheet, View, ViewStyle} from 'react-native'
+import {StyleProp, View, ViewStyle} from 'react-native'
 import {ModerationUI} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
@@ -40,7 +40,7 @@ export function ContentHider({
 
   if (!blur || (ignoreMute && isJustAMute(modui))) {
     return (
-      <View testID={testID} style={[styles.outer, style]}>
+      <View testID={testID} style={[a.overflow_hidden, style]}>
         {children}
       </View>
     )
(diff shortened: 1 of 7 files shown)


src/components/moderation/ContentHider.tsx
```diff
@@ -163,21 +163,3 @@ export function ContentHider({
...
-    marginTop: 4,
-    paddingVertical: 14,
-    paddingLeft: 14,
-    paddingRight: 18,
-  },
-  showBtn: {
-    marginLeft: 'auto',
-    alignSelf: 'center',
-  },
-})
```
mozzius, line 183: Didn't end up touching this file, but I noticed that these styles are completely unused

haileyok (approved):
Looks good. Guess we didn't break quotes, composer, etc. so this should be the only place to fix.

Small thing, do we want to add some margin to the right hand side of the display name on profiles?

haileyok:
also do we want to fix this one? (tested on main, maybe its fixed here)

Screenshot 2024-08-08 at 9 51 49 AM

mozzius:
Yep I fixed that - composer reply to displayname has `numberOfLines={1}`

PR #4901, opened 2024-08-08
Show just-posted replies above OP replies

gaearon:
Follow-up to https://github.com/bluesky-social/social-app/pull/4882.

Normally we show OP replies above all so that threads work. However, if you have _just_ posted a reply, it is confusing to have it appear offscreen. Twitter handles this by prioritizing just-posted own replies above the OP's (but showing them below the OP's if you navigate away and back to the post). This accomplishes the same thing.

We can do this now because #4898 consolidated the logic for "after post" callback in the post thread component.

## Test Plan

Normal mode:

https://github.com/user-attachments/assets/0dc0bbc9-dd47-4357-a6ab-2433ac0839e4

Thread mode:

https://github.com/user-attachments/assets/94c3392d-9f48-48e0-8e24-c40a14c980d8

Note that we only do this when we're sure we won't completely _displace_ the OP post out of view:

https://github.com/user-attachments/assets/d6e1a9f6-1b97-40be-8a31-d285fe8d067f

Concretely, this means that we only do this in threaded more _or_ directly under the highlighted post. Otherwise, it feels too confusing to have the OP post disappear.

### DIFF
--- a/src/state/queries/post-thread.ts
+++ b/src/state/queries/post-thread.ts
@@ -137,6 +137,7 @@ export function sortThread(
   opts: UsePreferencesQueryResponse['threadViewPrefs'],
   modCache: ThreadModerationCache,
   currentDid: string | undefined,
+  justPostedUris: Set<string>,
 ): ThreadNode {
   if (node.type !== 'post') {
     return node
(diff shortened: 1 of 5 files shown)


mozzius (approved):
Seems solid in theory

PR #4902, opened 2024-08-08
Show Post header on desktop web

gaearon:
This thing:

https://github.com/user-attachments/assets/00879c47-7596-4936-927f-3bbda515d882

Previously we only showed it on mobile and tablet layouts, but not on desktop web. This wasn't intentional but was related to difficulties of scroll positioning. We've now resolved the underlying difficulties and can do this without much problem.

Arguably it feels a little large but we can fix that separately if we want — that's the same size as on other pages for desktop.

In addition to visual/behavior consistency, this also fixes a jump that appears when posting into a thread while desktop, which was caused by conditionally showing the view header based on `isFetching` when it should've been `isLoading`. If we revert this PR in the future, we need to apply the other fix then.

## Test Plan

https://github.com/user-attachments/assets/9e41c6d7-b3e1-4a47-bfbc-b92b3105cfc5

https://github.com/user-attachments/assets/81464ce9-af25-41dd-9401-9c78f028cdca

https://github.com/user-attachments/assets/06899518-6554-4fc8-8911-a789ac8c2787

No behavior difference on native (checked iOS and Android).

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -89,7 +89,7 @@ export function PostThread({uri}: {uri: string | undefined}) {
   const {hasSession, currentAccount} = useSession()
   const {_} = useLingui()
   const t = useTheme()
-  const {isMobile, isTabletOrMobile} = useWebMediaQueries()
+  const {isMobile} = useWebMediaQueries()
   const initialNumToRender = useInitialNumToRender()
   const {height: windowHeight} = useWindowDimensions()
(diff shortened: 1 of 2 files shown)


mozzius (approved):
Works! It does feel a little large though, we need to do a design pass on all the headers I think

Tanza3D:
i'd 100% say that as is this makes the ui worse, if it had some info like who posted the original post or how long ago the post was i think it'd be great, but at the moment it just reminds you that yes, you are indeed looking at a post, on a social media site with posts

gaearon:
yea we decided not to do this

PR #4908, opened 2024-08-09
[Video] Prevent pausing of background audio with `expo-video` on iOS

haileyok:
Requires https://github.com/bluesky-social/social-app/pull/4905 and https://github.com/bluesky-social/social-app/pull/4907

## Why

We need to do a little tweaking of `AVAudioSession`'s options depending on if we actually want to be playing audio or not for a video. We also have to ensure allow mixing on app start.

All of the commits here are from the other two PRs except for the final commit, see [`4aed6ce` (#4908)](https://github.com/bluesky-social/social-app/pull/4908/commits/4aed6ce69e0585d8f6f8cca9d82de8007a445ddc)

## Test Plan

Really requires a merge for good testing. Once built, we should play some music on an iPhone then open the app and see if the audio pauses or not. It shouldn't!


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -37,6 +38,12 @@ export function VideoEmbedInnerNative() {
         player={player}
         style={a.flex_1}
         nativeControls={true}
+        onEnterFullscreen={() => {
+          PlatformInfo.setAudioMixWithOthers(false)
+        }}
+        onExitFullscreen={() => {
+          PlatformInfo.setAudioMixWithOthers(true)
+        }}
```
haileyok, line 46: we can change audio level/mute/etc. here as well

mozzius (approved):

PR #4911, opened 2024-08-09
Don't kick to login screen on network error

gaearon:
With this fix, we will no longer kick you to the login screen when the agent emits a `network-error` event. The original motivation for kicking you to the login screen was twofold:

- The agent was losing the `.session` field so it literally couldn't continue, and we wanted to mirror that in app state.
- The app did not have a reliable way to track when the network comes back online on native, so requests would get stuck.

We're fixing this by:

- Putting the `.session` field back on the agent after a `network-error` event. (We should maybe fix this upstream?)
- Adding a mechanism to track whether the app is online that works across platforms.

Regarding the mechanism. I tried initially to use different network libraries. But it seemed pretty unreliable — for example, on the Android simulator I'd often get network reported as working even when I was turning it off. But not always. In the end I decided to just poll a backend endpoint. This should also handle cases like captive portals.

I initially tried to guard it by a StableID gate. But it turns out this doesn't currently work with out Statsig setup due to `initialize` being called too late. I've tried fixing it and it's a can of worms. It might be easier with their new SDK but I say let's for now just test it ourselves and in the worst case disable this over OTA.

## Test Plan

For web, verify that quick double refresh in Safari no longer kicks you out. Verify that the app can recover from toggling the WiFi.

For native, verify starting the app with WiFi off no longer kicks you to login screen. Verify turning on WiFi fixes stuck screens within five seconds. Toggle WiFi on and off, verify it's generally able to recover after being offline.

### DIFF
--- a/src/lib/react-query.tsx
+++ b/src/lib/react-query.tsx
@@ -20,15 +20,12 @@ async function checkIsOnline(): Promise<boolean> {
     setTimeout(() => {
       controller.abort()
     }, 15e3)
-    const res = await fetch(
-      'https://plc.directory/did:plc:z72i7hdynmk6r22z27h6tvur',
-      {
-        cache: 'no-store',
-        signal: controller.signal,
-      },
-    )
+    const res = await fetch('https://public.api.bsky.app/xrpc/_health', {
+      cache: 'no-store',
+      signal: controller.signal,
+    })
     const json = await res.json()
-    if (json.id) {
+    if (json.version) {
       return true
     } else {
       return false
--- a/src/lib/react-query.tsx
+++ b/src/lib/react-query.tsx
@@ -9,6 +9,7 @@ import {
 } from '@tanstack/react-query-persist-client'
 
 import {isNative} from '#/platform/detection'
+import {listenNetworkConfirmed, listenNetworkLost} from '#/state/events'
 
 // any query keys in this array will be persisted to AsyncStorage
 export const labelersDetailedInfoQueryKeyRoot = 'labelers-detailed-info'
@@ -35,6 +36,14 @@ async function checkIsOnline(): Promise<boolean> {
   }
 }
 
+listenNetworkLost(() => {
+  onlineManager.setOnline(false)
+})
+
+listenNetworkConfirmed(() => {
+  onlineManager.setOnline(true)
+})
+
 let checkPromise: Promise<void> | undefined
 function checkIsOnlineIfNeeded() {
   if (checkPromise) {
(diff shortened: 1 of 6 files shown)


src/lib/react-query.tsx
```diff
@@ -14,6 +14,44 @@ import {isNative} from '#/platform/detection'
...
+    checkPromise = undefined
+    onlineManager.setOnline(isOnline)
+  })
+}
+
+setInterval(() => {
+  if (AppState.currentState === 'active') {
+    checkIsOnlineIfNeeded()
+  }
+}, 5000)
```
pfrazee, line 53: Wait so you're doing this constantly?

pfrazee: Is there any way we could just trigger this when a request fails in a certain way?

src/state/session/agent.ts
```diff
@@ -193,7 +193,14 @@ async function prepareAgent(
 
   // Now the agent is ready.
   const account = agentToSessionAccountOrThrow(agent)
+  let lastSession = agent.session
   agent.setPersistSessionHandler(event => {
+    if (agent.session) {
+      lastSession = agent.session
+    } else if (event === 'network-error') {
+      // Put it back, we'll try again later.
+      agent.session = lastSession
```
pfrazee, line 202: You're right, upstream doesnt need to do this. Let's evaluate post oauth refactor -- that changes everything anyway

src/lib/react-query.tsx
pfrazee: Just looked at upstream. I think if you started these checks once `network-error` gets fired and stopped them once it resolved, that should be effective.

I'm going to approve so that we can test on TF but we definitely should _not_ be pinging any of our infra once every 5s for the entirely uptime of the app

pfrazee (approved):

src/lib/react-query.tsx
gaearon: The `network-error` event is unfortunately not enough. I am only seeing it being raised when the error happens during a session refresh. I’m not seeing it fire on regular failures so I don’t have a good way to observe them (unless we’re going to dig further into the change on the other side).

One possible alternative is to fork a fetch wrapper on the web too (currently we only do it on native) and add a handler there. I was hoping to avoid touching that part because @haileyok’s PR moves that around and at that point I think it would become more convenient. Maybe we should wait for that, or it’s worth forking it now?

gaearon: The other thing we could look at is the RQ onError callback but that wouldn’t cover everything, and I also couldn’t get it to work when I tried it last time.

pfrazee: Yeah unfortunately we really can't move forward with it as-is.

gaearon:
Updated to only poll after we have a failure (and stop polling after first successful response, whether via the poll mechanism or via the agent).

haileyok (approved):

Hussain-hamim:
Thanks for this change it really was a headache

PR #4912, opened 2024-08-09
[Videos] handle app backgrounding

mozzius:
1. Handles app backgrounding
2. Ensures video plays when fullscreening and unfullscreening, just in case it manages to get paused somehow

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
@@ -23,17 +23,17 @@ export function VideoEmbedInnerNative() {
   const ref = useRef<VideoView>(null)
   const isScreenFocused = useIsFocused()
   const isAppFocused = useAppState()
-  const [prevFocused, setPrevFocused] = useState(isAppFocused)
+  const prevFocusedRef = useRef(isAppFocused)
 
   // resume video when coming back from background
   useEffect(() => {
-    if (isAppFocused !== prevFocused) {
-      setPrevFocused(isAppFocused)
+    if (isAppFocused !== prevFocusedRef.current) {
+      prevFocusedRef.current = isAppFocused
       if (isAppFocused === 'active') {
         player.play()
       }
     }
-  }, [isAppFocused, prevFocused, player])
+  }, [isAppFocused, player])
 
   // pause the video when the screen is not focused
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -101,6 +114,24 @@ function Controls({
...
+      case 'idle':
+      case 'loading':
+      case 'readyToPlay': {
+        if (!player.playing) player.play()
+        enterFullscreen()
+        break
+      }
+      case 'error': {
+        // not exactly sure what to do here
+        // maybe replay the video?
```
haileyok, line 128: should be okay, can rm comment if you want

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -21,6 +22,18 @@ export function VideoEmbedInnerNative() {
...
+  const [prevFocused, setPrevFocused] = useState(isAppFocused)
+
+  // resume video when coming back from background
+  useEffect(() => {
+    if (isAppFocused !== prevFocused) {
+      setPrevFocused(isAppFocused)
+      if (isAppFocused === 'active') {
+        player.play()
+      }
+    }
```
haileyok, line 35: Might be wrong, but is there a reason that we need to store the previous state? Seems that we might be able to just do

```tsx
const appState = useAppState()

useEffect(() => {
  if (appState !== 'active') return
  player.play()
}, [appState, player])
```

Seems that `player` should be stable so the only thing that should change here is `appState`.

haileyok (approved):
One note but lgtm.

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
mozzius: I wanted to constrain when it fires specifically to when it's going from !active -> active. Your version will also fire on initial render. idk, do you think that matters?

haileyok: Ahh, right. Maybe let's just stick the prev value in a ref then?

mozzius: Yeah that'd work

PR #4913, opened 2024-08-09
[Video] Animate controls + fade out time after a while

mozzius:
Controls can be a bit obtrusive, and I noticed twitter has a nice behaviour where the countdown time gets removed after a little while. This adds a nice animation, and fades out the controls after 5 seconds

https://github.com/user-attachments/assets/79750cf2-1a6b-4e8b-9f0f-ac4dba6dc921


haileyok (approved):
sgtm

PR #4914, opened 2024-08-09
[Videos] Add error boundary to native

mozzius:
Copies over error boundary logic and turns web UI into a reusable component


haileyok:
I think we want to be logging these errors to Sentry (at least the native ones) so we can get some info if we run into anything. Maybe I'm missing where it happens, but let's add if it isn't already doing that.

mozzius:
`<ErrorBoundary>` handles it :)

image

haileyok:
swear i looked there and didnt see it...cool!

haileyok (approved):

PR #4915, opened 2024-08-09
[Video] Resume background audio whenever muting video audio

haileyok:
## Why

We want the background audio to properly resume once we either mute the video audio or exit fullscreen. Let's do that.

## How

See https://developer.apple.com/documentation/avfaudio/avaudiosession/setactiveoptions/1616603-notifyothersondeactivation

## Test Plan

Merge, build, test.


mozzius (approved):
let's see

PR #4917, opened 2024-08-10
Show handle in recent searches and fix truncation

rolandcrosby:
The profile view in "recent searches" used to not show a label if the user didn't have a display name set:

Screenshot 2024-08-10 at 12 06 54 AM

Also, when the name was truncated, the existing implementation was supposed to add "...", but the way it was implemented, the ellipsis often spilled onto a second line that overflowed and didn't render. This change switches the label to use the built-in ellipsis truncation behavior.

Screenshot 2024-08-10 at 12 06 42 AM

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -960,8 +960,7 @@ function SearchHistory({
                     />
                     <Text
                       style={[pal.text, styles.profileName]}
-                      numberOfLines={1}
-                      ellipsizeMode="tail">
+                      numberOfLines={1}>
                       {profile.displayName || profile.handle}
                     </Text>
                   </Link>
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -961,7 +961,7 @@ function SearchHistory({
                     <Text
                       style={[pal.text, styles.profileName]}
                       numberOfLines={1}
-                      ellipsizeMode="head">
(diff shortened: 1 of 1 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -965,8 +958,11 @@ function SearchHistory({
                       style={styles.profileAvatar as StyleProp<ImageStyle>}
                       accessibilityIgnoresInvertColors
                     />
-                    <Text style={[pal.text, styles.profileName]}>
-                      {truncateText(profile.displayName || '', 12)}
+                    <Text
+                      style={[pal.text, styles.profileName]}
+                      numberOfLines={1}
+                      ellipsizeMode="head">
```
haileyok, line 964: Is there a reason for setting this to `head`? My understanding is that (although your screenshot doesn't show this happening!) `head` will result in something like `helloworld123` being truncated like `...world123`. I think we'd rather the front of the string always be visible, so you'd see `hellowo...`.

haileyok (commented):
Thanks! Left one question.

src/view/screens/Search/Search.tsx
haileyok: Cool! Yea it's a bit weird to me as well sometimes. I think the differences in android and ios text rendering are fairly different so you see a lot of things that just do not line up...

haileyok (approved):
lgtm. appreciate you fixing this up for us!!

PR #4920, opened 2024-08-11
Mark string for localization

surfdude29:
This PR marks the `Compressing...` string in `src/view/com/composer/videos/VideoTranscodeProgress.tsx` for localization.


haileyok (approved):
thanks!

PR #4924, opened 2024-08-12
[Video] Invert usage of `setAudioActive`

haileyok:
Typo, should have been the other way around for `true`/`false`


mozzius (approved):
I feel so stupid because I noticed that when you first added it, but assumed you'd meant to do that!!

haileyok:
dont worry i feel just as stupid 😅

PR #4926, opened 2024-08-13
[Video] Audio duck off main thread

haileyok:
Seems that iOS doesn't like modifying the `AVAudioSession` on the main thread - this causes some lag. Let's run it on a background thread instead.


mozzius (approved):
Fancy

PR #4927, opened 2024-08-13
Fix session persistence handlers

gaearon:
There is a regression in https://github.com/bluesky-social/social-app/pull/4857 — although it appears to be setting up the session persistence handler, that handler (stored in a ref) is not being read. Additionally, it is a little fragile that merely calling `createAgent*` modifies a ref since `create*` functions were intended to be stateless. Otherwise it gets very difficult to reason about race conditions.

In this PR, I remove the ref and move the persistence handler to be scoped per-agent like before. The underlying `BskyAgent` signature changed so that the persistence handler has to be supplied during creation. However, we don't always know the `did` during its creation (e.g. we wouldn't know the `did` of an agent whose `createAccount` call is yet to follow). So we want late binding for the session handler. To do that, I store it in a local variable and swap it out when ready.

I ended up removing `prepareAgent` handler because it was obscuring what's actually happening. This adds a bit of duplication but should be clearer.

## Test Plan

Have not explicitly tested but @haileyok knows a few bugs that this should fix.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -1,9 +1,5 @@
 import React from 'react'
-import {
-  AtpPersistSessionHandler,
-  AtpSessionEvent,
-  BskyAgent,
-} from '@atproto/api'
+import {AtpSessionEvent, BskyAgent} from '@atproto/api'
 
 import {track} from '#/lib/analytics/analytics'
 import {logEvent} from '#/lib/statsig/statsig'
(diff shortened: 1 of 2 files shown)


src/state/session/index.tsx
```diff
@@ -266,7 +250,6 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
       // We never reuse agents so let's fully neutralize the previous one.
       // This ensures it won't try to consume any refresh tokens.
       prevAgent.sessionManager.session = undefined
-      setPersistSessionHandler(undefined)
```
gaearon, line 269: Note I am no longer cleaning up the persistence handler when the current agent gets swapped. I think this should be fine. It didn't used to be cleaned up in the past either, I only added this at some intermediate stage for peace of mind. But we already have a condition [here](https://github.com/bluesky-social/social-app/blob/99d1a881f2f5c16dddfc10550b39e379690c8135/src/state/session/reducer.ts#L72-L75) that prevents inactive agents from clearing persisted data. For inactive agents, `refreshedAccount` [should be](https://github.com/bluesky-social/social-app/blob/99d1a881f2f5c16dddfc10550b39e379690c8135/src/state/session/index.tsx#L65) `undefined` because we already set `prevAgent.sessionManager.session = undefined` [here](https://github.com/bluesky-social/social-app/blob/99d1a881f2f5c16dddfc10550b39e379690c8135/src/state/session/index.tsx#L268). And even if somehow it had a session object, the reducer always [matches events with state by the did](https://github.com/bluesky-social/social-app/blob/99d1a881f2f5c16dddfc10550b39e379690c8135/src/state/session/reducer.ts#L99) so it won't update the wrong thing.

gaearon:
let's go with https://github.com/bluesky-social/social-app/pull/4928

PR #4928, opened 2024-08-13
subclass agent to add setPersistSessionHandler

haileyok:
## Why

The API upgrade for OAuth deprecates `BskyAgent` as well as removes the ability to modify the persist session handler by calling `setPersistSessionHandler()` on the agent. Instead, the handler has moved to `SessionManager` and it can no longer be set.

To keep our previous logic in tact with minimal changes (and actually persist the session on change now 🙃) let's subclass `BskyAgent` to support this.

Also adding a `dispose()` method which unsets the session on the agent as well as unsets the persist handler (whereas before we made these two calls separately in `agent.ts`)

## Test Plan

Try changing your email on an alt, and it should immediately reflect the update.


src/state/session/agent-wrapper.ts
```diff
@@ -0,0 +1,31 @@
+import {AtpSessionEvent, BskyAgent} from '@atproto/api'
+
+/**
+ * This is a temporary class that wraps `BskyAgent` and allows us to update the session persist handler when needed,
+ * since the deprecated `BskyAgent` after the API upgrades for OAuth no longer support setting the handler in the
+ * `SessionManager` itself.
+ */
+export class BskyAgentWrapper extends BskyAgent {
```
gaearon, line 8: `BskyAppAgent`? To make it more legit.

src/state/session/agent-wrapper.ts
```diff
@@ -0,0 +1,31 @@
+import {AtpSessionEvent, BskyAgent} from '@atproto/api'
+
+/**
+ * This is a temporary class that wraps `BskyAgent` and allows us to update the session persist handler when needed,
```
gaearon, line 4: I wouldn't necessarily keep it temporary, it's nice to be able to preconfigure some things here. I might set up some other overrides for the online management.

gaearon (approved):

src/state/session/agent-wrapper.ts
```diff
@@ -0,0 +1,31 @@
...
+      service,
+      persistSession: (event: AtpSessionEvent) => {
+        if (this.persistSessionHandler) {
+          this.persistSessionHandler(event)
+        }
+      },
+    })
+  }
+
+  setPersistSessionHandler(handler?: (event: AtpSessionEvent) => void) {
```
gaearon, line 23: should probably be non-nullable here

haileyok: Ah, yea might as well just use `dispose()` for that when needed 👍

PR #4930, opened 2024-08-13
[Video] Try/catch video play/pause

mozzius:
Everything else should be caught and logged by the error boundary


haileyok (approved):
yea lets try and see if these do get caught. not sure if they will but worth a shot.

PR #4933, opened 2024-08-13
Hidden reply cache

estrattonbailey:
[Review w/o whitespace](https://github.com/bluesky-social/social-app/pull/4933/files?diff=split&w=1). Can ignore the CI failures, I'll fix on the main branch.

This PR adds the ability to hide a reply directly from a feed generator as well as from the notifications feed. To do so, it adds an in-memory cache of replies that were hidden, and uses that array to filter out those replies from feed generators and notifications feed.

So what happens here is: when the user hides a reply, then the in-memory cache and threadgate record are updated with that URI. Then when navigating to other screens where this reply may be present, the reply is filtered on the client. Upon a full reload, the server strips out the reply, and the in-memory cache is reset.

For feed items, we're able to hydrate the threadgate from the root post, since it's present in feed slices. So we use that for the initial state, and then defer to the in-memory cache.

**Why not use post shadow?**

Not every place where we want to be able to hide replies has a root post whose `threadgate` prop we can update e.g. notifications. We also lack a root post if the root post or a post above the reply was blocked or hidden. Plus, if we were to generate a `ThreadgateView` for a post without one already, we'd need to hack in a fake record and leave an empty CID. Not that big of deal but still meh.

### DIFF
--- a/src/App.web.tsx
+++ b/src/App.web.tsx
@@ -39,6 +39,7 @@ import {Provider as LoggedOutViewProvider} from '#/state/shell/logged-out'
 import {Provider as ProgressGuideProvider} from '#/state/shell/progress-guide'
 import {Provider as SelectedFeedProvider} from '#/state/shell/selected-feed'
 import {Provider as StarterPackProvider} from '#/state/shell/starter-pack'
+import {Provider as HiddenRepliesProvider} from '#/state/threadgate-hidden-replies'
 import {ActiveVideoProvider} from '#/view/com/util/post-embeds/ActiveVideoContext'
 import * as Toast from '#/view/com/util/Toast'
 import {ToastContainer} from '#/view/com/util/Toast.web'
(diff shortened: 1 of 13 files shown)


src/App.web.tsx
```diff
@@ -162,7 +163,9 @@ function App() {
                     <I18nProvider>
                       <PortalProvider>
                         <StarterPackProvider>
-                          <InnerApp />
+                          <HiddenRepliesProvider>
```
gaearon, line 166: This placement seems a bit wrong to me. Since hidden replies are owned by a user, I think we probably want to move this inside the user-local part keyed by the `did`. Then it would get destroyed between logouts. Does that make sense?

pfrazee (approved):
Didnt check for thoroughness or run any tests, but all the approaches here LGTM!

estrattonbailey:
Need to rebase the base branch to get new SDK updates and such, gonna merge this and fix the provider arrangement as my next TODO

PR #4934, opened 2024-08-14
[Video] set audio category to ambient every time a new player is made

mozzius:
Audio category could get out of sync with the player if you switch video


haileyok:
Nice. I made a small tweak on the native side so we avoid updating the session when unnecessary (it's a little slow). Looks good to me.

haileyok (approved):

mozzius:
Ah perfect, nice one

PR #4935, opened 2024-08-14
[Embed] Starter pack embed embed

mozzius:
Screenshot 2024-08-14 at 19 14 23

Starter pack posts you can test with

https://bsky.app/profile/sturdyalex.bsky.social/post/3kzn2d2yrz42q
https://bsky.app/profile/sturdyalex.bsky.social/post/3kzn2e6hgli2s


bskyembed/src/components/embed.tsx
```diff
@@ -350,3 +363,62 @@ function GenericWithImage({
...
+          </p>
+        )}
+      </div>
+    </Link>
+  )
+}
+
+// from #/lib/strings/starter-pack.ts
+function getStarterPackOgCard(
+  didOrStarterPack: AppBskyGraphDefs.StarterPackView,
```
haileyok, line 412: nit: could just be `starterPack`

bskyembed/src/components/embed.tsx
```diff
@@ -350,3 +363,62 @@ function GenericWithImage({
...
+            <p className="text-sm text-textLight line-clamp-2 leading-[18px]">
+              Starter pack by{' '}
+              {content.creator.displayName || content.creator.handle}
+            </p>
+          </div>
+        </div>
+        {content.record.description && (
+          <p className="text-sm mt-1">{content.record.description}</p>
+        )}
+        {!!content.joinedAllTimeCount && content.joinedAllTimeCount > 50 && (
```
haileyok, line 400: I wonder if we might want to always show this count for embeds. Or at least drop the count to something like 10 or 15?

bskyembed/src/components/embed.tsx
```diff
@@ -350,3 +363,62 @@ function GenericWithImage({
...
+      <div className="py-3 px-4">
+        <div className="flex space-x-2 items-center">
+          <img src={starterPackIcon} className="w-10 h-10" />
+          <div>
+            <p className="font-semibold leading-[21px]">
+              {content.record.name}
+            </p>
+            <p className="text-sm text-textLight line-clamp-2 leading-[18px]">
+              Starter pack by{' '}
+              {content.creator.displayName || content.creator.handle}
```
haileyok, line 393: nit: need and `@` in front of handle

bskyembed/src/components/embed.tsx
```diff
@@ -134,28 +137,37 @@ export function Embed({
...
-        )
+        // return (
+        //   <GenericWithImageEmbed
+        //     image={record.creator.avatar}
+        //     title={record.creator.displayName || record.creator.handle}
+        //     href={`/profile/${record.creator.did}`}
+        //     subtitle="Labeler"
+        //     description={`Liked by ${record.likeCount ?? 0} users`}
+        //   />
+        // )
```
estrattonbailey, line 148: Should this still be commented?

bskyembed/src/components/embed.tsx
haileyok: since you're also here @estrattonbailey, thoughts?

mozzius: I'm down to drop it to 10 or so. Should I do it for the in-app one too presumably?

haileyok: hmm, no strong opinions but i think leaving it alone in the app is fine. embeds feel more "immediately going to be seen" so think having the count show up there more easily is nice.

too bad we don't have like "joined in the last 24 hours" though.

haileyok (approved):

haileyok:
looks good, same question as eric re: the comment but i think you're right to comment it out.

bskyembed/src/components/embed.tsx
mozzius: imo it should be consistent, whichever way we do it. bare in mind the most popular one in this UK wave (the quiet riot one) has only 67 joins

PR #4941, opened 2024-08-15
RFC: Make "Sign out" log out current account, add "Sign out of all accounts"

gaearon:
Implements https://github.com/bluesky-social/social-app/issues/4775

We might need to think more about how this should work, but putting it out for initial discussion

### DIFF
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -14,6 +14,7 @@ export type LogEvents = {
   }
   'account:loggedOut': {
     logContext: 'SwitchAccount' | 'Settings' | 'SignupQueued' | 'Deactivated'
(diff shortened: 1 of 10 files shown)


gaearon:
Hm I guess this lost the indication which one you're signed into.

gaearon:
Added "Other accounts" grouping to make it very clear which one you're signed in.

Alternatively we could do the checkmark. I don't love the one we're using though, and we'd have to place it next to the dropdown — don't want to remove the ability to remove the last remaining account.

We could also have some text label on the right liked "Signed in".

mozzius (approved):
Checked for both single and multiple accounts, looks good

pfrazee (approved):
Added a small spacing tweak. LGTM!

PR #4942, opened 2024-08-15
Hack patch for testing OTA update crash behavior

haileyok:
## Why

There's a crash that's happening when `reloadAsync()` gets called by `expo-updates`. This is coming from attempting to destructure a null pointer, whenever removing all of the `SharedObject`s in `SharedObjectRegistry.clear()`.

A little digging shows that the `deinit` block of `VideoPlayer` runs _after_ `clear()` get's called. However, right now `clear()` calls `dict.removeAll()` _asynchronously_ inside of a separate queue from the main one. This is a small test to see if this helps to alievate the crashes. I don't consider this a production fix.

## Test Plan

Merge, let it build, and monitor for OTA crashes as we have been internally.


mozzius (approved):

haileyok:
we do want to revert this prior to cutting a release. added a note to release tracker.

PR #4943, opened 2024-08-15
[Video] add scrubber to the web player

mozzius:
https://github.com/user-attachments/assets/40f76d09-01da-4ee9-a3a9-0578ed148476

- [x] Mouse behaviour
- [x] aria attributes
- [x] Keyboard controls

# Test plan

- Test with mouse
- Test with keyboard (focus with tab, move left/right with arrow keys)
- Test different browsers - Firefox in particular

# ~Known issue~

~If you mouseUp while the mouse is over a post (but not over the video), it'll navigate to the post. `evt.preventDefault()` doesn't prevent it~

Solved! We make everything unclickable while scrubbing 😅


src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -210,123 +255,333 @@ export function Controls({
...
+      if (ref.current.contains(target)) {
+        evt.preventDefault()
+        isSeekingRef.current = true
+        seek(evt)
+        onSeekStart()
+        startScrubbing()
+      }
+    }
+
+    // Handle mouseup and touchend anywhere on the document
```
gaearon, line 440: This feels complicated. Can you use pointer events + setPointerCapture?

mozzius: didn't know that was a thing! thanks, will do

src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -210,123 +256,359 @@ export function Controls({
...
+        onSeekStart()
+
+        // HACK: there's divergent browser behaviour about what to do when
+        // a pointerUp event is fired outside the element that captured the
+        // pointer. Firefox clicks on the element the mouse is over, so we have
+        // to make everything unclickable while seeking. Remember to remove
+        // this class when you're done! -sfn
+        if (isFirefox) {
+          document.body.classList.add('force-no-clicks')
+        }
```
mozzius, line 449: I found the best way to solve the firefox click issue was to make everything unclickable while seeking. Any better ideas welcome

src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -210,123 +256,354 @@ export function Controls({
...
+    // pointer. Firefox clicks on the element the mouse is over, so we have
+    // to make everything unclickable while seeking -sfn
+    if (isFirefox && scrubbing) {
+      document.body.classList.add('force-no-clicks')
+
+      return () => {
+        document.body.classList.remove('force-no-clicks')
+      }
+    }
+  }, [scrubbing])
```
mozzius, line 486: This was the least worst solution I could come up with for the firefox click issue

haileyok (approved):
great work!

Left a little note in slack re: what pointers we should use (grabby hand? pointer? regular cursor?) We can tweak all of that in a different iteration.

src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
haileyok: I was actually able to break this if i take my cursor outside of the window while still pressing down. Can we prevent that from breaking?

src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -210,123 +256,359 @@ export function Controls({
...
+      const {signal} = abortController
+      document.documentElement.addEventListener(
+        'mouseleave',
+        () => {
+          isSeekingRef.current = false
+          onSeekEnd()
+          setScrubberActive(false)
+        },
+        {signal},
+      )
```
mozzius, line 484: @haileyok is the mouse leaving the window a firefox-only issue or should I apply this fix on all browsers?

haileyok: i assume it is just firefox yea. i couldn't get it to repro in chrome or safari.

mozzius: Ok cool, this fix should be good then

PR #4944, opened 2024-08-15
Fix orphaned feed slices, handle blocks

estrattonbailey:
### Relies on https://github.com/bluesky-social/atproto/pull/2721

Fixes a regression in #4871 where `A -> B -> C` and `A` blocks `B`, we'd see the entire `A -> B -> C` thread, whereas the intention was to filter out those "orphaned" threads. If we want to show orphaned threads, it's a one-line change in `feed-tuners`.

## Testing

Test the scenarios below using the backend PR. Otherwise test prod, nothing should look different using prod data. Only the backend PR sends down data needed to compute the correct view.

### `A` blocks `B`, or `B` blocks `A`. Viewing as `C`.

See in prod, thread is orphaned (and technically "Reply to test.esb.lol" should say "Reply to blocked post")
CleanShot 2024-08-15 at 17 19 01@2x

In main, you can see through the 3p blocks:
CleanShot 2024-08-15 at 17 20 28@2x

In this PR, `C` only sees `A` because the thread view would otherwise be "orphaned"
CleanShot 2024-08-16 at 16 09 08@2x

### `C` blocks `A`. Viewing as `C`
In all cases the entire feed slice is filtered out 👍 

### `C` blocks `B`. Viewing as `C`.
In all cases, only `A` is visible 👍 

## Not found posts

In case of `B` deleting their post, again only `A` is visible in feeds, and when viewing `C` author feed, this PR now notes that the post was removed with the following:

CleanShot 2024-08-16 at 16 28 45@2x

I chose "unknown post" instead of "deleted post" because there are multiple reasons a post might be "not found".

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -23,6 +23,7 @@ type FeedSliceItem = {
   record: AppBskyFeedPost.Record
   parentAuthor: AppBskyActorDefs.ProfileViewBasic | undefined
   isParentBlocked: boolean
+  isParentNotFound: boolean
 }
 
 type AuthorContext = {
@@ -68,6 +69,7 @@ export class FeedViewPostsSlice {
     }
     const parent = reply?.parent
     const isParentBlocked = AppBskyFeedDefs.isBlockedPost(parent)
+    const isParentNotFound = AppBskyFeedDefs.isNotFoundPost(parent)
     let parentAuthor: AppBskyActorDefs.ProfileViewBasic | undefined
     if (AppBskyFeedDefs.isPostView(parent)) {
       parentAuthor = parent.author
@@ -77,6 +79,7 @@ export class FeedViewPostsSlice {
       record: post.record,
       parentAuthor,
       isParentBlocked,
+      isParentNotFound,
     })
     if (!reply || reason) {
       return
@@ -94,11 +97,15 @@ export class FeedViewPostsSlice {
     const isGrandparentBlocked = Boolean(
       grandparent && AppBskyFeedDefs.isBlockedPost(grandparent),
     )
+    const isGrandparentNotFound = Boolean(
+      grandparent && AppBskyFeedDefs.isNotFoundPost(grandparent),
+    )
     this.items.unshift({
       post: parent,
       record: parent.record,
       parentAuthor: grandparentAuthor,
       isParentBlocked: isGrandparentBlocked,
+      isParentNotFound: isGrandparentNotFound,
     })
     if (isGrandparentBlocked) {
       this.isOrphan = true
@@ -121,6 +128,7 @@ export class FeedViewPostsSlice {
       post: root,
       record: root.record,
       isParentBlocked: false,
+      isParentNotFound: false,
       parentAuthor: undefined,
     })
     if (parent.record.reply?.parent.uri !== root.uri) {
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -92,7 +92,14 @@ export class FeedViewPostsSlice {
       this.isOrphan = true
       return
     }
-    const grandparent = reply?.grandparent
+    const root = reply.root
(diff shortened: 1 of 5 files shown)


src/state/preferences/feed-tuners.tsx
```diff
@@ -46,7 +46,7 @@ export function useFeedTuners(feedDesc: FeedDescriptor) {
       return feedTuners
     }
     if (feedDesc === 'following') {
-      const feedTuners = [FeedTuner.removeOrphans]
```
gaearon, line 49: afaik we don't want to do this, right?

src/lib/api/feed-manip.ts
```diff
@@ -103,7 +111,7 @@ export class FeedViewPostsSlice {
     })
     if (isGrandparentBlocked) {
       this.isOrphan = true
-      // Keep going, it might still have a root.
+      return
```
gaearon, line 114: i'm not sure this is correct — this would make it have no root which would e.g. break thread deduping. is this necessary?

src/lib/api/feed-manip.ts
```diff
@@ -205,8 +213,22 @@ export class FeedTuner {
...
+    const replyRefMap: ReplyRefMap = new Map()
+    for (const item of feed) {
+      for (const ancestor of [item.reply?.parent, item.reply?.root]) {
+        // trust server for blocked and not found reply parent/root refs
+        if (
+          AppBskyFeedDefs.isPostView(ancestor) ||
+          AppBskyFeedDefs.isBlockedPost(ancestor) ||
+          AppBskyFeedDefs.isNotFoundPost(ancestor)
+        ) {
+          replyRefMap.set(ancestor.uri, ancestor)
```
gaearon, line 225: i wonder if we can make this more scoped to clarify its purpose. e.g. `blockedPostUris` and `notFoundPostUris` sets get threaded through and then we check against those. we never really need the actual content of the post so this feels like it's exposing too much of the "not really consistent" picture to the slice logic. i'd like to keep it more limited.

src/lib/api/feed-manip.ts
```diff
@@ -89,23 +92,33 @@ export class FeedViewPostsSlice {
       this.isOrphan = true
       return
     }
+    const root = reply.root
+    const rootIsView =
```
gaearon, line 96: What does `rootIsView` mean? I thought maybe it's about it being `PostView` but the other condition is `BlockedPost`. Is this about detecting the ones that have a string `uri` property? I think `NotFound` would have it too but it's not in the condition. So I think I'm missing what this boolean means and why it's in the condition below.

src/lib/api/feed-manip.ts
```diff
@@ -89,23 +92,33 @@ export class FeedViewPostsSlice {
...
+      AppBskyFeedDefs.isPostView(root) || AppBskyFeedDefs.isBlockedPost(root)
+    // if parent is also the root, we have data!
+    const grandparent =
+      rootIsView && parent?.record?.reply?.parent?.uri === root?.uri
```
gaearon, line 100: Nit: This is overly defensive. Let's avoid using `?.` when we don't need it. For example, `root?.uri` is unnecessary because `root` always exists by now. This makes it obvious that we won't hit `undefined === undefined` by mistake — that's not obvious in the code as is.

`parent?.record?.reply` can be `parent.record.reply`.

gaearon (approved):
i think this looks good, see nits above. i don't understand what `rootIsView` is yet so it would be nice to confirm it works as intended (what's special about "post or blocked"?) before merging.

src/lib/api/feed-manip.ts
estrattonbailey: Ah, yeah it's exactly that. Added check for NotFoundPost too, good call 👍

gaearon (approved):
lg!

estrattonbailey:
Actually should have no effect on main for now, so should be clear to merge this ahead of app view changes

PR #4946, opened 2024-08-16
[Experiment] Always show bottom bar

gaearon:
This is where your DMs and notifications are. Let's experiment with always keeping it visible.

## Test Plan

No changes outside the gate.

With the gate:

https://github.com/user-attachments/assets/90d2974b-a151-4e68-916d-493baa0e7c87

https://github.com/user-attachments/assets/9e8096f6-ac33-40ba-b699-3a9888abc921

https://github.com/user-attachments/assets/49818daf-b849-46eb-9f60-6b2719e9f05e

### DIFF
--- a/src/lib/hooks/useMinimalShellTransform.ts
+++ b/src/lib/hooks/useMinimalShellTransform.ts
@@ -2,6 +2,7 @@ import {interpolate, useAnimatedStyle} from 'react-native-reanimated'
 
 import {useMinimalShellMode} from '#/state/shell/minimal-mode'
 import {useShellLayout} from '#/state/shell/shell-layout'
+import {useGate} from '../statsig/statsig'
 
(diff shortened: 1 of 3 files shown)


src/lib/hooks/useMinimalShellTransform.ts
```diff
@@ -27,8 +28,13 @@ export function useMinimalShellHeaderTransform() {
 export function useMinimalShellFooterTransform() {
   const mode = useMinimalShellMode()
   const {footerHeight} = useShellLayout()
+  const gate = useGate()
+  const isFixedBottomBar = gate('fixed_bottom_bar')
```
gaearon, line 32: Have to check here because can't do on another thread. It'll have behavior immediately on scroll so seems fine.

haileyok (approved):

PR #4958, opened 2024-08-20
Add `list hidden` screen

haileyok:
## Why

Whenever a list gets a `!hide` flag applied to it, the app should prevent users from viewing the list. However, users do still need to be able to remove the list from their pins or subscribed mod lists. Let's add a screen to do that.

We also were not hiding the list name from the title, since it was being set in `useSetTitle()`. I removed that, and it will show `List Hidden` in those cases.

## How

Whenever viewing the list, a screen will be shown telling the user the list has been hidden. If they are the owner, they will have the option to delete the list. If they are subscribed to the moderation list, they will be able to unsubscribe. And if they have the list saved or in their pins, they'll be able to remove it from there.

## Test Plan

Two test lists that I have made: http://localhost:19006/profile/haileyok.com/lists/3l247irgjn62r and http://localhost:19006/profile/haileyok.com/lists/3l24csobejw2r

You can also test deleting the list by changing the `isHidden` variable on line 151 in `ProfileList.tsx` to `true`.

Whenever you are the creator of the list, you should be able to still tap into the list with "Show anyway".

Screenshot 2024-08-20 at 11 29 05 AM

Non-creators should not be able to see the list.

Screenshot 2024-08-20 at 11 29 47 AM

However, if you are not the creator and you have saved or subscribed to the list, you _should_ be given the ability to unsave/unsubscribe from that list.

Screenshot 2024-08-20 at 11 32 14 AM

Screenshot 2024-08-20 at 11 33 21 AM


src/components/Error.tsx
```diff
@@ -27,31 +24,10 @@ export function Error({
...
-      // Checking the state for routes ensures that web doesn't encounter errors while going back
-      if (navigation.getState()?.routes) {
-        navigation.dispatch(StackActions.push(...router.matchPath('/')))
-      } else {
-        navigation.navigate('HomeTab')
-        navigation.dispatch(StackActions.popToTop())
-      }
-    }
-  }, [navigation, canGoBack, onGoBackProp])
+  const goBack = useGoBack(onGoBack)
```
haileyok, line 30: No change, just extracting this function to a reusable hook.

src/view/screens/ProfileList.tsx
```diff
@@ -142,12 +148,13 @@ function ProfileListScreenLoaded({
   const {openModal} = useModalControls()
   const isCurateList = list.purpose === 'app.bsky.graph.defs#curatelist'
   const isScreenFocused = useIsFocused()
+  const isHidden = list.labels?.findIndex(l => l.val === '!hide') !== -1
```
estrattonbailey, line 151: In the last PR I used the mod SDK and ScreenHider to do this. I like the new screen better, but we may want to borrow the mod SDK handling in case we want to respect other filter labels here. We can suss it out in the morning 👍

estrattonbailey (commented):
Looks so good!

src/view/screens/ProfileList.tsx
estrattonbailey: Yeah I think it'd be nice to keep this sorta functionality around CleanShot 2024-08-20 at 07 59 37@2x

estrattonbailey: Maybe the move is allow an additional component prop on ScreenHider and improve the look of that component if we want. There are a possibility of 2 CTAs right now, so I think instead of adding another button, maybe it'd be nice to allow for a lil alert section above the buttons that explains an additional action that can be taken. Can discuss this morning!

haileyok: Yep added this now.

haileyok:
> We should also hide the list name under Moderation > Lists in user settings for subscribers

I'm going to PR this separately.

src/screens/List/ListHiddenScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+            <Button
+              variant="solid"
+              color="secondary"
+              size="medium"
+              label={_(msg`Remove from saved feeds`)}
+              onPress={onRemoveList}
+              disabled={isProcessing}>
+              <ButtonText>
+                <Trans>Removed from saved feeds</Trans>
+              </ButtonText>
```
estrattonbailey, line 154: Nit: could add loader using `<ButtonIcon icon={Loader} />` to these buttons when processing

src/screens/List/ListHiddenScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+        logger.error('Failed to unblock list', {message: e})
+        Toast.show(
+          _(
+            msg`There was an issue. Please check your internet connection and try again.`,
+          ),
+        )
+        return
+      }
+    }
+    Toast.show(_(msg`Unsubscribed from list`))
```
estrattonbailey, line 79: Nit: should pry be inside the try blocks?

src/screens/List/ListHiddenScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+              </ButtonText>
+            </Button>
+          ) : null}
+          {isOwner ? (
+            <Button
+              variant="solid"
+              color="secondary"
+              size="medium"
+              label={_(msg`Show list anyway`)}
+              onPress={() => setIsContentVisible(true)}
```
estrattonbailey, line 163: This method won't do anything if `noOverride` is set. Kinda just future-proofing, but might be good to conditionally render this button. Alternatively, we remove that guard and force ourselves to check `meta.noOverride` on the hider context.

estrattonbailey (approved):
Nice! Couple of nits. I can hit them in a bit if you don't beat me 👍

src/screens/List/ListHiddenScreen.tsx
haileyok: Hrm, I think we just want it at the end here, since there are two separate try blocks and either of them could fail. (If both succeed we wouldn't see the toast inside of the first block)

src/screens/List/ListHiddenScreen.tsx
haileyok: yupyup

src/screens/List/ListHiddenScreen.tsx
estrattonbailey: Better safe than sorry, sgtm

estrattonbailey (approved):

PR #4960, opened 2024-08-20
Add Hider component

estrattonbailey:
Adds generic `Hider` component for use anywhere e.g.
- `ProfileList` entire screen
- list title on `Moderation > Lists` list of lists

Provides access to the result of `useModerationCauseDescription` and prepackages the `ModerationDetailsDialog`.

For lists right now we don't need to handle any other labels, so no need to have an override button for the moment.


haileyok:
ty!

PR #4964, opened 2024-08-20
List card hiding of title

haileyok:
## Why

We want to hide the title of a list from non-owners when the `!hide` label is applied. There was also a few remaining usages of the old `ListCard` component which I am removing here to simplify this change.

## Test Plan

Checking the moderation lists page should still show hidden lists even if you are not the owner. The name of the list however should be `Hidden list`.

If you are the owner, you should still see the list as well as its name.

Screenshot 2024-08-20 at 1 37 17 PM


src/components/moderation/Hider.tsx
```diff
@@ -29,22 +29,22 @@ export function Outer({
   modui,
   children,
 }: React.PropsWithChildren<{
-  modui: ModerationUI
+  modui?: ModerationUI
```
haileyok, line 32: I'm loosening these types since we often are using `useModerationOpts()`, which is possibly undefined - but really shouldn't be since the context is wrapping the entire app.

src/components/moderation/Hider.tsx
```diff
@@ -27,13 +27,17 @@ export const useHider = () => React.useContext(Context)
 
 export function Outer({
   modui,
+  override,
   children,
 }: React.PropsWithChildren<{
+  override?: boolean
```
haileyok, line 33: one additional prop that lets us supply some override, like if the hidden content belongs to the current session's account.

estrattonbailey (approved):

PR #4969, opened 2024-08-21
Fix fixed footer experiment

gaearon:
Haven't tested this very deeply but I think this should fix the `fixed_bottom_bar` experiment.

It takes an alternate approach to what I had before. The idea is to split the `mode` variable into `headerMode` and `footerMode`. Initially we always update them in unison (but read from `headerMode` unless we're specifically interested in the footer). Then we rip out the current conditional logic during read. Then we add some conditional logic during write, and only for the `MainScrollProvider` case which governs on-scroll showing/hiding. Other cases still set both.

Individual commits tell the story.

I think this should be enough.

## Test Plan

Verify the old mode (without gate) works like in prod.

Verify the experiment doesn't break starter pack wizard and other screens where we intentionally hide the bottom bar.


haileyok (approved):

haileyok:
Will want to see how it performs on-device (both platforms) once merged, but works good in simulator and code lgtm.

PR #4977, opened 2024-08-22
Optimistic hidden replies

estrattonbailey:
### Relies on https://github.com/bluesky-social/atproto/pull/2737

Merges local hidden reply cache (specific to user) with initially hydrated threadgate record, if we have it. Uses that as the source of truth for rendering logic.

Integrates new app view PR that returns `threadgate` on the `getPostThread` response, so that we always have an initial record to hydrate from, if it exists.

Clarifies relationship, so that only OP sees labels.


haileyok (approved):
Tested this locally, checks out. I also tested in production (which doesn't quite work while waiting on appview deploy) and the mutations look good. Things update immediately as expected 👍

PR #4979, opened 2024-08-22
Use RichText for sp description

estrattonbailey:
Before
image

After:
image


haileyok (approved):
moved the `isRecord` check to the component above, so we don't have to check twice in `LandingScreenLoaded`.

PR #4980, opened 2024-08-22
Make settings account buttons a little nicer

estrattonbailey:
Replaces the main line item wrapper with our new interactive elements, and replaces the ... menu with `Menu` and a larger click target. So we get nice hover/pressed states.

After:
CleanShot 2024-08-22 at 16 42 54@2x
CleanShot 2024-08-22 at 16 42 50@2x

Before
CleanShot 2024-08-22 at 16 43 30@2x
CleanShot 2024-08-22 at 16 43 27@2x

### DIFF
--- a/src/view/screens/Settings/index.tsx
+++ b/src/view/screens/Settings/index.tsx
@@ -55,8 +55,11 @@ import {UserAvatar} from 'view/com/util/UserAvatar'
 import {ScrollView} from 'view/com/util/Views'
 import {DeactivateAccountDialog} from '#/screens/Settings/components/DeactivateAccountDialog'
 import {atoms as a, useTheme} from '#/alf'
+import {Button, ButtonContext} from '#/components/Button'
 import {useDialogControl} from '#/components/Dialog'
 import {BirthDateSettingsDialog} from '#/components/dialogs/BirthDateSettings'
+import {Link as NewLink} from '#/components/Link'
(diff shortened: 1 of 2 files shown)


src/view/screens/Settings/index.tsx
```diff
@@ -72,62 +75,69 @@ function SettingsAccountCard({
...
           {profile?.displayName || account.handle}
-        </Text>
-        <Text type="sm" style={pal.textLight} numberOfLines={1}>
-          {account.handle}
-        </Text>
+        </NewText>
+        <NewText
+          style={[t.atoms.text_contrast_medium, a.leading_tight]}
+          numberOfLines={1}>
+          @{account.handle}
```
haileyok, line 114: do we want `sanitizeHandle` here?

haileyok (approved):

src/view/screens/Settings/index.tsx
estrattonbailey: Yeah that's a good idea here

haileyok:
One other small thing I initially missed. I think it's a little weird that i get this flash if i scroll in the right area of the list (when leaving the main link and entering the ... button. Maybe we just use the old link (no hover effect)?


https://github.com/user-attachments/assets/f4f2fcec-c6ac-4ca6-9568-1ab80893b3ab

PR #4982, opened 2024-08-23
Tweak rendering of top headers so they always appear even during load

haileyok:
## Why

On web, some of the headers don't render initially because of how we were rendering them. Web headers were being added inside of the `ListHeaderComponent`, however the `List` itself doesn't get rendered until the initial load has completed.

We also were displaying some top borders that were a little weird and didn't match how we did things elsewhere in the app. I've tweaked those so they look better.

## How

Moving the header to the outer `Screen` component means it will show up even while loading. This requires a little bit of refactoring (mostly wrapping the content in a `CenteredView` rather than just a `View`, but this will work for now. It would be nice to refactor these components entirely at some point, but that's more of a task.

## Test Plan

- On desktop web, check that the header shows up even on a Cmd+R
- Check that on web, no matter which breakpoint is being used results in only a single border at the top (no double borders!)
- Check that on native there is no visual change in how the header works.

### DIFF
--- a/src/components/Lists.tsx
+++ b/src/components/Lists.tsx
@@ -3,7 +3,6 @@ import {StyleProp, View, ViewStyle} from 'react-native'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
-import {useWebMediaQueries} from 'lib/hooks/useWebMediaQueries'
 import {cleanError} from 'lib/strings/errors'
 import {CenteredView} from 'view/com/util/Views'
 import {atoms as a, flatten, useBreakpoints, useTheme} from '#/alf'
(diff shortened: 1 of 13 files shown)


src/components/Lists.tsx
```diff
@@ -116,10 +117,10 @@ export function ListHeaderDesktop({
   title: string
   subtitle?: string
 }) {
-  const {gtTablet} = useBreakpoints()
+  const {isDesktop} = useWebMediaQueries()
   const t = useTheme()
 
-  if (!gtTablet) return null
+  if (!isDesktop) return null
```
haileyok, line 123: `gtTablet` and `isDesktop` still are not 1:1. The current prod behavior lets you hit a ~15px area where the header just won't show up because the `List` component uses `isDesktop` vs this using `gtTablet`

estrattonbailey: I thought we fixed this. Can we not align them?

haileyok: Fixed in [`f444f40` (#4982)](https://github.com/bluesky-social/social-app/pull/4982/commits/f444f40131447983f7d0ef6a672090442cdb1313)

estrattonbailey (approved):

PR #4983, opened 2024-08-23
fix hashtag screen side borders

haileyok:
## Why

There was some double border/half screen border weirdness going on in prod. Fixing.

## Test Plan

Hard refresh the hashtag screen (https://bsky.app/hashtag/test) and observe the jank. Then try on localhost.

Should be no change on native.


estrattonbailey (approved):
Nice looks great

PR #4992, opened 2024-08-25
Improvements to NSE

haileyok:
## Why

Doing a bit of cleanup around the NSE. Primarily here, I'm adding a dispatch queue for updating `UserDefaults`. See code comment:

```
// Because multiple instances of `NotificationService` may exist, we should
// be careful in accessing preferences that will be mutated _by the
// extension itself_. For example, we should not worry about `playChatSound`
// changing, since we never mutate that value within the extension itself.
// However, since we mutate `badgeCount` frequently, we should ensure that
// these updates always run sync with each other and that the have access
// to the most recent values.
```

## Test Plan

There shouldn't be any change in how things work most of the time. This may lead to some improvements in glitched counts, though.

### DIFF
--- a/modules/BlueskyNSE/NotificationService.swift
+++ b/modules/BlueskyNSE/NotificationService.swift
@@ -2,46 +2,80 @@ import UserNotifications
 import UIKit
 
 let APP_GROUP = "group.app.bsky"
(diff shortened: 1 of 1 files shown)


mozzius (approved):
Looks good, let's get it in the TF and see if it breaks things

PR #4996, opened 2024-08-26
[Video] Make compress/upload cancelable

mozzius:
Adds an abort controller so that the upload/compression is cancellable


haileyok (approved):

haileyok:
Noting this doesn't cancel uploads at least on native. Can handle separately.

PR #4997, opened 2024-08-26
[Video] Performance tweaks for Android

haileyok:
## Why

```
  // On Android, there are some performance concerns if we call `updateActiveViewAsync` too often while scrolling.
  // This isn't really a problem with how `updateActiveViewAsync` works, but rather that it will cause too many
  // ExoPlayer instances to be created and destroyed, which is pretty expensive and can cause some jank in the feed.
  // When comparing to Twitter, it seems they are much more conservative with when videos will begin playback than
  // they are on iOS, likely for the same reason.
  //
  // Because we're going to be calling this far less on Android, we will also be extra careful to make sure some video
  // starts playing when we reach the end of a scroll by calling `updateActiveViewAsync` at the end of a scroll (whether
  // the user ending the drag or the momentum ending). These extra calls will _not_ be excessively expensive, since it
  // is guaranteed to only happen once (React Native will only call either `onEndDrag` or `onMomentumEnd`, not both) and
  // any scroll performance jank will not be noticeable, since the scroll has already completed.
```


src/view/com/util/List.tsx
```diff
@@ -69,6 +83,9 @@ function ListImpl<ItemT>(
     },
     onEndDrag(e, ctx) {
       onEndDragFromContext?.(e, ctx)
+      if (isAndroid) {
+        runOnJS(updateActiveViewAsync)
```
mozzius, line 87: doesn't seem like you're actually calling the function here, shouldn't it be `runOnJS(updateActiveViewAsync)()`

mozzius: same below

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -91,22 +91,21 @@ function Controls({
...
-    })
-
+    // const interval = setInterval(() => {
+    //   // duration gets reset to 0 on loop
+    //   if (player.duration) setDuration(Math.floor(player.duration))
+    //   setCurrentTime(Math.floor(player.currentTime))
+    //
+    //   // how often should we update the time?
+    //   // 1000 gets out of sync with the video time
+    // }, 250)
```
mozzius, line 101: Any idea what we need to do to uncomment this? :/

mozzius (approved):
Looks good, although we do need to get the currentTime indicator back at some point

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
haileyok: Yea, I have a plan to fix this today.

PR #5005, opened 2024-08-28
Profile screen performance tweak - Adjust initial num to render based on header height

haileyok:
## Why

Whenever we first open a screen, we always want to initially render as few items as are necessary to fill the screen _and only fill the screen_. This results in faster tab switches and navigation pushes.

Right now, the profile tab will attempt to render around 9 posts initially on my profile, which is what would normally be needed to fill the screen. However, that number does not take into account the height of the profile header. Because the header often takes up half or more of the screen, we end up rendering a ton of extra posts that don't need to be there.

If we account for the header height, then the number of posts that are initially rendered when visiting my profile drops to four, cutting the initial render time of the screen by around half.

## Test Plan

Profile before and after

| Before | After |
|--------|--------|
| Screenshot 2024-08-27 at 5 45 51 PM | Screenshot 2024-08-27 at 5 49 29 PM |


src/screens/Profile/Sections/Feed.tsx
```diff
@@ -42,6 +43,10 @@ export const ProfileFeedSection = React.forwardRef<
   const queryClient = useQueryClient()
   const [hasNew, setHasNew] = React.useState(false)
   const [isScrolledDown, setIsScrolledDown] = React.useState(false)
+  const shouldUseAdjustedNumToRender = feed.endsWith('posts_and_author_threads')
```
haileyok, line 46: We are only concerned with doing this when it's the user's "Posts" tab. Subsequent tab switches:

1. May be done after the user has scrolled down the page, and so using the header offset wouldn't provide an accurate number
2. Won't be as noticeably expensive, since the rest of the screen's content has already been rendered.

PR #5008, opened 2024-08-28
Improve localization

quiple:
* Improves localization of starter pack names when automatically creating starter packs.
* Improves localization of `on Bluesky` at the bottom of the QR code sharing screen in the starter pack. Allows the string to be positioned after the logo.


estrattonbailey (approved):

estrattonbailey:
Thanks!

PR #5011, opened 2024-08-28
Hide quote counts for quotegated posts

estrattonbailey:
Test post: http://localhost:19006/profile/pfrazee.com/post/3l2shzxiref2u

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -398,7 +398,9 @@ let PostThreadItemLoaded = ({
                     </Text>
                   </Link>
                 ) : null}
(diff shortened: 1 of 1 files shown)


gaearon (approved):
have not tested but code lgtm

surfdude29:
The number next to the repost button still includes the number of quote posts... when quote posts aren't being displayed, shouldn't the numbers of reposts match?

E.g. on the test post it looks like there are 21 reposts if you go by the number next to the repost button (because it includes the 14 ghost quotes) but there are actually only 7

IMG_7923

https://main.bsky.dev/profile/pfrazee.com/post/3l2shzxiref2u

However, maybe it's not worth worrying about as more people upgrade to 1.90 and there are fewer ghost quotes being created 🤷🏻‍♂️

PR #5013, opened 2024-08-28
[Video] Add `timeRemainingChange` event to `player` in `expo-video`

haileyok:
## Why

We were originally doing this with a `setInterval` in https://github.com/bluesky-social/social-app/pull/4997, but it would cause some problems - particularly on Android - where `player` was being accessed after it was disposed of on the native side. This kind of exposed something that would actually be easier to just get as an event from `player` itself, rather than having to do the calculations on the JS side.

## Test Plan

Essentially, this should work the same as before. Worth checking that nothing broke in regards to time formatting and rounding, but it seems solid from my testing.


mozzius (approved):
lgtm! hard to review patch files but seems to work well

PR #5023, opened 2024-08-29
Remove new_user_guided_tour and tour code

gaearon:
We decided not to ship this. This PR removes both the callsite and the implementation.

I opted to do it this way (rather than keep dead code around) so that we aren't confused by worrying about updating these callsites if something else underneath changes. If/when we need a tour again, I think we can just revert this PR and readd it.

I don't feel super strongly about this and can see an argument for keeping it in repo though.

## Test Plan

Load app on all platforms. Walk between Search and Home. Create a new account and verify that works.


pfrazee (approved):
Yep yep this is 100% the move

didnt test but reviewed code

PR #5026, opened 2024-08-29
Fix language filtering for feeds

gaearon:
I broke this during a refactor — it was not ok to change iteration direction here. This is confusing enough that IMO we should just do a `filter`, especially given that we already clone the array anyway.

## Before

Screenshot 2024-08-29 at 20 05 02

## After

Screenshot 2024-08-29 at 20 05 11


estrattonbailey (approved):
LGTM, tested on web

PR #5028, opened 2024-08-29
Use `npx` for Expo

haileyok:
Smolbean, just using `npx` for all the `expo` scripts in `package.json`.


gaearon:
why is this better / what's the difference?

lukeacl:
As it stands this could be unnecessary clutter due to the local package being resolved by npx anyway. It could be handy if it was "npx expo@latest" to always use the latest CLI version, but could be unpredictable if latest versions have unexpected issues?

PR #5030, opened 2024-08-30
[Experiment] Suggest profiles in profile

gaearon:
A more prominent placement for the profile suggestions since they've gotten better recently. This adds an interstitial to profile feeds (Posts and Replies tabs only).

## Test Plan

Enable the gate. Observe this block after 5 slices:

Screenshot 2024-08-30 at 02 07 41

It should be profile-dependent. Check with different profiles. (Might work better with a fresh account if your cluster follows are already saturated.) Some profiles won't have it (if no suggestions were served). For some accounts you'll see generic suggestions (if we don't know their cluster).

Verify Discover interstitials work like before.

### DIFF
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import {View} from 'react-native'
 import {ScrollView} from 'react-native-gesture-handler'
-import {AppBskyFeedDefs, AtUri} from '@atproto/api'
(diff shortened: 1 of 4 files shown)


RoboLeader:
Hi! Providing some feedback: I hate this, I do not like having my timeline split with a carousel of random accounts. Please take this out of the app 👍

gaearon:
There's a new feedback thread in https://github.com/bluesky-social/social-app/issues/5214, would appreciate specific feedback there! Thanks.

Issue #4951, opened 2024-08-17
Show exact number of followers on hover over the rounded number


xanlpz:
A small thing, and perhaps this is lacking on purpose. But it would be nice to be able to see the exact number of followers after you hit 1k by hovering over the rounded number. Would only work on web, but I believe Twitter does (or did) the same.

mozzius:
You can now see this number by clicking on it and looking in the header of the page

PR #5032, opened 2024-08-30
Provide full tooltips on counts when truncated and rounded

lukeacl:
When viewing profiles, the following, followers, and post counts are truncated/rounded when they become larger numbers, such as **1,000** becoming **1K**. This change provides a tooltip on the web on hover to provide full information when truncation/rounding has occurred.

Resolves bluesky-social/social-app#4951


estrattonbailey (approved):
Nice! Dig this, gonna merge into a separate base and fix some types that we didn't port over with the new Text component 👍

PR #5048, opened 2024-08-30
Mark unhydrated replies as orphaned

gaearon:
Noticed this during appview/relay syncing issues.

(This logic existed in the old code but I dropped but because I didn't know why. Now I know.)

If we have `post.record.reply` but we didn't receive a `reply`, something went wrong (i.e. we don't know its parent or root). Let's mark it as orphaned in this case.

This will make it filtered out of Following. On other feeds, it will display as "reply to a post" (so it won't seem top-level).

## Test Plan

Don't know how to force it to happen, but I hardcoded the condition as `true` and verified the above.

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -81,7 +81,14 @@ export class FeedViewPostsSlice {
       isParentBlocked,
       isParentNotFound,
     })
-    if (!reply || reason) {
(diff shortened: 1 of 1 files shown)


gaearon:
Verified this works with a real post too after the relay restarted again

haileyok (approved):
nice find

PR #5049, opened 2024-08-31
Show some known OP replies in Following

gaearon:
This adjusts the heuristic to show more replies in Following.

Concretely, replies like `A -> B -> A` or `A -> ... -> B -> A` will now show even if you _only_ follow A (but not B). Previously, they were getting filtered out. Effectively, this means that an OP reply bumps the OP thread if you follow OP — but the condition is that B has to have >0 likes. This would filter out replies that didn't get anyone interested so far.

This doesn't contribute to the snowballing effect because you're _already_ following A anyway.

## Examples

I follow Mark but I don't follow Tim. With the change, this reply shows in Following.

Screenshot 2024-08-31 at 03 02 27

I follow Joyce but I don't follow David. With the change, this reply shows in Following.

Screenshot 2024-08-31 at 03 03 48

I follow Jerry but I don't follow Molly. With the change, this reply shows in Following.

Screenshot 2024-08-31 at 02 59 23

## Counter-examples

I follow Megan but this reply would still not show because the parent (whom I don't follow) has no likes.

Screenshot 2024-08-31 at 03 38 37

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -366,11 +366,7 @@ export class FeedTuner {
     ): FeedViewPostsSlice[] => {
       for (let i = 0; i < slices.length; i++) {
         const slice = slices[i]
-        if (
-          slice.isReply &&
-          !slice.isRepost &&
-          !shouldDisplayReplyInFollowing(slice.getAuthors(), userDid)
-        ) {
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Code looks good, seems like a good change, didn't test

gaearon:
let's give it a spin

PR #5056, opened 2024-09-01
Add some more brazilian portuguese translation

talesperes:
Hello bluesky team!
Here is the translation of some phrases into Brazilian Portuguese that I could easily find in the app. I hope this can be helpful :)


fabiohcnobre:
Hello @talesperes 
I have already added the translations to my PR #5057 which is more complete and is being reviewed by @pietrofxq 
Thanks

Issue #5039, opened 2024-08-30
Pasting text into the Bluesky web client post box deletes blank lines


mcclure:
**Describe the bug**

When pasting multiline text into the Bluesky New Post box, any blank lines are deleted. This is inconvenient because it is usual for multi-line skoots to include one blank line between paragraphs. I often write posts in another editor and then paste them into bluesky (reasons I might do this: I want to write the post ahead of time [bluesky has no], I want to write an entire thread all at once [bluesky has no autothreader]) and when I do, the post gets garbled in this way. It's inconvenient.

**To Reproduce**

Copy these four lines. Click "New Post" on bluesky. Paste.

> i think that i
>
> will never see
>
> a poem lovely
>
> as a tree

**Expected behavior**

![image](https://github.com/user-attachments/assets/3c776244-d7d3-47b3-88ec-01caf3a6fc43)

**Observed behavior**

![image](https://github.com/user-attachments/assets/380279fe-f516-4bce-818d-8863ef0ff4e7)

**Details**

Platform:
I see the problem on Chrome 127.0.6533.120 on Windows 10.
I see the problem on Firefox 192.0.2 on Linux (Ubuntu 24.04).
I *DO NOT* see the problem in the Bluesky Android app.

PR #5070, opened 2024-09-01
fix [5039]: Pasting text into the Bluesky web client post box deletes blank …

half144:
fixed the bug of pasting rich formatted text and it removing the white lines

fix this issue: https://github.com/bluesky-social/social-app/issues/5039

Captura de Tela 2024-09-01 às 19 14 17

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -155,13 +155,18 @@ export const TextInput = React.forwardRef(function TextInputImpl(
           let preventDefault = false
 
           if (clipboardData) {
-            if (clipboardData.types.includes('text/html')) {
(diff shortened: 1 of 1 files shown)


mary-ext:
This pull request undoes the fixes for the issues referenced here 

https://github.com/bluesky-social/social-app/pull/4327

half144:
qual a traducao? I tested the fixes you provided, and they continue to work on my end. Could you please help identify where it might be failing on your side so that I can address it? I also tested pasting the URLs directly, and they loaded normally. I was able to replicate the behavior reported in the issues and it worked as expected.

half144:
I tested the fixes you provided, and they continue to work on my end. Could you please help identify where it might be failing on your side so that I can address it? I also tested pasting the URLs directly, and they loaded normally. I was able to replicate the behavior reported in the issues and it worked as expected

Issue #5071, opened 2024-09-01
Post date can be altered by computer's date


mxsinister:
Changing your computer's date(or smartphone, haven't tested) lets you change the date of your post 

1. Write your post
2. Change your device date
3. Quickly press send

**Details**

 - Platform: Linux Mint

Brayozin:
this is intended behavior:

https://docs.bsky.app/docs/advanced-guides/timestamps

PR #5072, opened 2024-09-01
fix [5071]: Post date can be altered by computer's date

half144:
fix Post date can be altered by computer's date 

issue: https://github.com/bluesky-social/social-app/issues/5071

### DIFF
--- a/src/lib/api/get-server-date.ts
+++ b/src/lib/api/get-server-date.ts
@@ -1,0 +1,12 @@
+/**
+ * Will get the current time on the server
+ * because the client time is not accurate
+ * and can make user manipulate the time of the post
+ */
+export const getServerDate = async () => {
+  const createdAt = await fetch('https://worldtimeapi.org/api/timezone/Etc/UTC')
+  const response = await createdAt.json()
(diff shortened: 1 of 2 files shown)


src/lib/api/get-server-date.ts
PatrickalKhouri, line 1: but how will this addapt to the timezone of the client?

Brayozin:
this is intended behavior:

https://docs.bsky.app/docs/advanced-guides/timestamps

src/lib/api/get-server-date.ts
AdrianoCahete: Maybe get server date and timezone info from client, then apply the difference on server date?

ssg:
In addition to @Brayozin's note:

1) Anyone who can change the system date can also modify the date retrieved from the server just by using browser debugging tools. So, any fix for this on the client-side is useless.

2) This change also couples the client's stability to an external service not maintained by Bluesky.

3) Bluesky already supports non-chronological dates on posts such as https://bsky.app/profile/microsoft.com/post/3lcm2osud6f2n 

I suggest closing this PR, @half144 .

PR #5074, opened 2024-09-02
Improve animations for like button

haileyok:
## Why

The current animation makes use of `useAnimatedStyle` and `interpolate`, which causes some pretty significant render performance issues, particularly on older Android phones.

### Before

https://github.com/user-attachments/assets/ebf19aa8-ea45-4cb1-8583-f9a84391abb2

### After

https://github.com/user-attachments/assets/efde0975-23ad-4701-afe1-57e2e317d894


mozzius (approved):
Looks good, it also fixes the issue we were seeing on phones with large font sizes! Let's get this in asap

PR #5083, opened 2024-09-02
Make the counter more rounded

nucleartux:
I hope it wasn't intended design 😨 

Before:
Screenshot 2024-09-02 at 12 28 10

After:
Screenshot 2024-09-02 at 12 32 44
Screenshot 2024-09-02 at 12 32 48

PS. The circle still isn't perfectly round because the width is half a pixel more than the height. If no one minds, I can make a PR to make it look more like this:
Screenshot 2024-09-02 at 12 22 57

### DIFF
--- a/src/view/shell/desktop/LeftNav.tsx
+++ b/src/view/shell/desktop/LeftNav.tsx
@@ -470,7 +470,7 @@ const styles = StyleSheet.create({
     fontSize: 12,
     fontWeight: 'bold',
     paddingHorizontal: 4,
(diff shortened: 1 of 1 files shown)


mary-ext:
how does it look when the notification counter goes to 30+?

nucleartux:
Before:
image

After:
image

estrattonbailey (approved):
Nice, thanks! Doing some more in #5488 but happy to include this too ✨

PR #5095, opened 2024-09-02
Maintain like state when navigating between screens

haileyok:
## Why

We're using a `WeakMap` right now for our post thread shadow, which is a bit annoying. The way we use it, we use a `PostView` as the key for lookups, which is generally fine but causes problems when navigating between screens. For example:

https://github.com/user-attachments/assets/61a15586-fa91-409b-b16f-73df85a1c502

This happens because the `PostView` that we get in response to `getPostThread` is _not_ the same as the `PostView` that was seen in the feed.

## Test Plan

Since here we use just the post's URI as a key, we get the expected behavior of "keeping" the like whenever transition between screens

https://github.com/user-attachments/assets/66cb74ec-9f62-4eee-bfe3-82bee71b2e93

### DIFF
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -26,30 +26,28 @@ export interface PostShadow {
 export const POST_TOMBSTONE = Symbol('PostTombstone')
 
 const emitter = new EventEmitter()
-const shadows: WeakMap<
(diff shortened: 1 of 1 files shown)


thientran1707:
@CodiumAI-Agent /ask

PR #5101, opened 2024-09-03
[Video] Say No to Audio (Disable `expo-video` `AudioSession` management)

haileyok:
## Why

We're managing `AudioSession` state on our own through `PlatformInfo`. This lets us decide what to do when we want to do it, rather than relying on Expo's hardcoded management.

## Test Plan

Test different ways of starting/stopping video audio playback while background audio is playing on iOS. Things shouldn't get borked, even between backgrounding and foregrounding the app.


mozzius (approved):
looks good in theory, let's test on TestFlight

mozzius (approved):

PR #5104, opened 2024-09-03
[Video] Manage foreground/background playback on the native side

haileyok:
## Why

There's a `useEffect` right now that's trying to manage playing/pausing the video whenever we foreground/backgorund the app, but it's causing some problems. Tapping the sign, let's move this to the expo module lifecycle events instead.

## Test Plan

Videos in the feed should pause whenever you background, then they should resume whenever you foreground again.

This should also fix the problem where a video will try to mute/unmute itself over and over.

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
@@ -46,7 +46,7 @@ export function VideoEmbedInnerNative({
         player={player}
         style={[a.flex_1, a.rounded_sm]}
         contentFit="contain"
-        nativeControls={true}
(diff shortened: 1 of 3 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -69,7 +46,7 @@ export function VideoEmbedInnerNative({
         player={player}
         style={[a.flex_1, a.rounded_sm]}
         contentFit="contain"
-        nativeControls={true}
+        nativeControls={false}
```
haileyok, line 49: Also setting this to `false`, which is what it should be. Doesn't affect the visibility of native controls when in fullscreen.

Issue #5105, opened 2024-09-03
Accidentally deleting draft post when you mention an account


LukeberryPi:
**Describe the bug**

When creating a post and mentioning an account, it is possible to blur from the text area and lose your content.

**To Reproduce** 

1. Click on "New Post"
2. Mention someone (e.g.: write "@lukeberrypi.bsky.social")
3. Place your cursor after the last character of the mention
4. Press `cmd + ←`
5. You will navigate to the previous page (same as clicking the Back button on Chrome)
6. Your original post is now lost

**Expected behavior**

The expected behavior is to navigate to the beginning of the line, and preserve the post content without unintended navigation.

**Screenshots**

https://github.com/user-attachments/assets/c96de9c0-c637-4b2c-af87-c40d8d3c94a8

**Details**

 - Platform: macOS and Google Chrome
 - Platform version: Sonoma 14.3.1 (23D60) and Chrome Version 128.0.6613.113 (Official Build) (arm64)
 - App version: -

**Additional context**

Just for clarification, when the user is not focused on an text input element (as in <textarea> or <input>), `cmd + ←` acts as a "Back" on browser navigation. The issue here is that while composing a post, user intent is to go to the beginning of the line, not going to the previous page.

I noticed that mentions on posts are treated differently to normal text, because they can't even be selected via `shift + ←`, or navigated from within the mention itself. I believe these are valuable clues to locating and solving this bug.

half144:
hi, I created a PR to fix this, can you take a look? let's see if they approve https://github.com/bluesky-social/social-app/pull/5110

PR #5110, opened 2024-09-03
fix [5015]: Accidentally deleting draft post when you mention an account

half144:
Fixed: Accidentally deleting draft post when you mention an account https://github.com/bluesky-social/social-app/issues/5105

This PR fixes the cursor movement functionality to the beginning of the line when the Cmd + ArrowLeft (or Ctrl + ArrowLeft) key combination is pressed. Previously, the behavior was not working as expected due to issues with setting the selection in the editor.

Screenshots:

https://github.com/user-attachments/assets/6e04d5bd-e998-4852-8a67-d8d625456c8a

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -245,7 +245,10 @@ export const TextInput = React.forwardRef(function TextInputImpl(
   React.useEffect(() => {
     if (editor) {
       const handleKeyDown = (event: KeyboardEvent) => {
-        if ((event.metaKey || event.ctrlKey) && event.code === 'ArrowLeft') {
+        if (
+          (event.metaKey || event.ctrlKey || event.altKey) &&
+          event.code === 'ArrowLeft'
+        ) {
           const {state, dispatch} = editor.view
           const {selection} = state
           const {$anchor} = selection
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -246,7 +246,7 @@ export const TextInput = React.forwardRef(function TextInputImpl(
     if (editor) {
       const handleKeyDown = (event: KeyboardEvent) => {
         if (
-          (event.metaKey || event.ctrlKey || event.altKey) &&
+          (event.metaKey || event.altKey) &&
           event.code === 'ArrowLeft'
         ) {
           const {state, dispatch} = editor.view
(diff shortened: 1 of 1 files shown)


src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -240,6 +241,30 @@ export const TextInput = React.forwardRef(function TextInputImpl(
     }
   }, [onEmojiInserted])
 
+  // thats fix cmd + left navigates back when the editor is focused in a @mention
+  React.useEffect(() => {
+    if (editor) {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        if ((event.metaKey || event.ctrlKey) && event.code === 'ArrowLeft') {
```
LukeberryPi, line 248: in windows, the hotkey for navigating to the previous page on chrome is `alt + ←`, not `ctrl + ←`.

we should test if the bug is reproducible in windows and adjust the code accordingly.

half144: you're right! I don't have a windows machine nearby to test this, but the issue will very likely occur in that environment as well. i'll make the change.

LukeberryPi (commented):
we need to check if the bug happens on windows and adjust the code

half144 (commented):
Added Alt key to cover Windows machines.

src/view/com/composer/text-input/TextInput.web.tsx
half144: done

Vincentroland (commented):
Explict

half144:
> Explict

you can see the error in https://github.com/bluesky-social/social-app/issues/5105 that pr take care of this

src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -240,6 +241,33 @@ export const TextInput = React.forwardRef(function TextInputImpl(
     }
   }, [onEmojiInserted])
 
+  // thats fix cmd + left navigates back when the editor is focused in a @mention
+  React.useEffect(() => {
+    if (editor) {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        if (
+          (event.metaKey || event.ctrlKey || event.altKey) &&
```
LukeberryPi, line 249:
```suggestion
          (event.metaKey || event.altKey) &&
```

we can omit `ctrl` event key since it does not trigger any browser navigation, which is intrisic to the bug.

half144: done

Vincentroland (commented):
Expilct

PR #5117, opened 2024-09-03
hide the .bsky from profile, feed and posted mesages

Caitano09:
The username will be clearer and friendlier without breaking the link

https://github.com/user-attachments/assets/c298dc97-9cb0-4df1-8696-b856ddbabe60

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -109,7 +109,7 @@ export function RichText({
             // @ts-ignore TODO
             dataSet={WORD_WRAP}
             onPress={onLinkPress}>
-            {segment.text}
+            {segment.text.split('.bsky')[0]}
           </InlineLinkText>
         </ProfileHoverCard>,
       )
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -26,7 +26,7 @@ export function isInvalidHandle(handle: string): boolean {
 export function sanitizeHandle(handle: string, prefix = ''): string {
   return isInvalidHandle(handle)
     ? '⚠Invalid Handle'
-    : forceLTR(`${prefix}${handle}`)
+    : forceLTR(`${prefix}${handle.split('.bsky')[0]}`)
 }
 
 export interface IsValidHandle {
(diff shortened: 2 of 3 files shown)


Vincentroland (commented):
Explict

joncar:
This would cause significant impersonation problems. Consider a handle that was `official.gov.bsky.mydomain.tld`, it'd show up as `official.gov`!

wish13yt:
The .bsky.social part is intentional... to prevent impersonation..

knotbin:
Even disregarding the impersonation issues brought up, it's not a good idea to just hide .bsky.social. If this idea were to be considered, it should hide the (default) extension of whatever server the user is on, not just .bsky.social.

lsuperti:
In addition to the issues already pointed out by others, if this is ever implemented, it should, by all means, be configurable with a setting, probably disabled by default. There must also be an easy way to see what the actual handle is, regardless of the setting.

seboslaw:
In Skeets (@skeetsapp.com) I have made this optional (configurable in Settings) and it only works with .bsky.social handles. All non-bsky handles are always displayed in full length.

knotbin:
I don't think this is a good idea for the main app. It's already moderately difficult to understand the structure of handles for new users, and this would only make it more complicated and make it easier to fall for impersonation, even if it is an optional setting.

Caitano09:
I understand better now. Thanks guys.

mary-ext:
bsky.social is not the only PDS that exists, just because it is the premiere default PDS does not mean it's allowed to get special treatment

knotbin:
@mary-ext I mentioned this in a previous comment.

> Even disregarding the impersonation issues brought up, it's not a good idea to just hide .bsky.social. If this idea were to be considered, it should hide the (default) extension of whatever server the user is on, not just .bsky.social.

ShreyanJain9:
> @mary-ext I mentioned this in a previous comment.
> 
> > Even disregarding the impersonation issues brought up, it's not a good idea to just hide .bsky.social. If this idea were to be considered, it should hide the (default) extension of whatever server the user is on, not just .bsky.social.

Even that doesn't seem like a great idea to me, handles are intentionally independent from the server you're on.

knotbin:
Each server has a default handle extension though

PR #5123, opened 2024-09-03
Add quick access to “Quote” action via long press

coreh:
This is an established UX pattern/power user shortcut from other social networks. When long pressing the repost button, you bypass the intermediate dialog and are sent straight to the compose screen.


estrattonbailey (approved):
Nice one, thanks!

PR #5125, opened 2024-09-03
Add slight spacing between Post and CW button

coreh:
Found it easy to hit the wrong button by accident since they are so close... This PR adds a slight gap between them to help prevent someone accidentally hitting “Post” when they want to add a CW first.

Before / After

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -560,6 +560,7 @@ export const ComposePost = observer(function ComposePost({
                   onChange={setLabels}
                   hasMedia={hasMedia}
                 />
+                <View style={[styles.spacer]} />
                 {canPost ? (
                   <Button
                     testID="composerPublishBtn"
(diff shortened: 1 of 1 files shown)


coreh:
(It might make sense to also move the button somewhere else altogether. Not sure how the UX Design process works at Bluesky, so I went with the minimally disruptive change)

src/view/com/composer/Composer.tsx
```diff
@@ -560,6 +560,7 @@ export const ComposePost = observer(function ComposePost({
                   onChange={setLabels}
                   hasMedia={hasMedia}
                 />
+                <View style={[styles.spacer]} />
```
haileyok, line 563: I think it might make more sense here to just change the fragment to a `<View>` and use `style={{gap: 6}}`. Otherwise seems like a good change!

coreh: Done. Had to change the gap from 6 to 14, since for the existing gap in the `topbarInner` was making the 6px spacer act like a 14px spacer

haileyok (approved):
Thanks!

PR #5127, opened 2024-09-03
[Video] Tweak playback handling

haileyok:
## Why

Adjust the logic to start/stop playback whenever switching tabs.

## Test Plan

- Visit one tab, unmute a video
- Switch to another tab. Audio should stop playing
- Switch back to the tab, the video should resume playing but audio should be muted

https://github.com/user-attachments/assets/7aa6a043-47de-4d7b-a96e-2cc3f2c1839e


mozzius (approved):
Looking goooood

PR #5128, opened 2024-09-04
Fix number formatting for non-English locales

coreh:
Number formatting data was not being loaded along with the pluralization rule data when activating locales. This PR makes it so that numbers are now properly formatted across the different locales.

Before / After


estrattonbailey (approved):
Awesome, thank you

PR #5153, opened 2024-09-04
[Video] Check upload limits before uploading

mozzius:
Implementation of `getUploadLimits` API. This is the point where you'll recieve `uploads_disabled`/`uploads_forbidden` errors, so this includes nice messages for that case.

# Test plan

Check network request `getUploadLimits` for however many videos you have left, and upload more than that many

Comment out the upload gate then try with an account that does not pass the gate

### DIFF
--- a/src/state/queries/video/video.ts
+++ b/src/state/queries/video/video.ts
@@ -167,7 +167,7 @@ export function useUploadVideo({
             break
           case "Failed to get user's upload stats":
             message = _(
-              msg`Error: could not determine if you're allowed to upload videos. Please try again.`,
+              msg`We were unable to determine if you are allowed to upload videos. Please try again.`,
             )
             break
           case 'User has exceeded daily upload bytes limit':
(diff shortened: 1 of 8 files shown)


src/state/queries/video/video.ts
```diff
@@ -149,10 +153,40 @@ export function useUploadVideo({
...
+            message = _(msg`You are not allowed to upload videos.`)
+            break
+          case 'Uploading is disabled at the moment':
+            message = _(
+              msg`Hold up! We’re gradually giving access to video, and you’re still waiting in line. Check back soon!`,
+            )
+            break
+          case "Failed to get user's upload stats":
+            message = _(
+              msg`Error: could not determine if you're allowed to upload videos. Please try again.`,
```
haileyok, line 170:
```suggestion
              msg`We were unable to determine if you are allowed to upload videos. Please try again.`,
```

haileyok (approved):

PR #5154, opened 2024-09-04
Ensure profile labels can be appealed separately from account labels

estrattonbailey:
We were essentially hard-coding the `subject` as an account. I moved that logic down to the `AppealForm` and inferred the subject from the label itself.


pfrazee (approved):
Just did a scan of the code -- no testing -- but lgtm

estrattonbailey:
Confirming with Aaron, then will merge

PR #5155, opened 2024-09-04
Hides the '.bsky.social' while also addressing impersonation and security issues

igorcv88:
Addresses full handle handling to make the interface cleaner and more user friendly

Inspired by: @Caitano09 pull request https://github.com/bluesky-social/social-app/pull/5117

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -1,34 +1,20 @@
-import React from 'react'
-import {TextStyle} from 'react-native'
-import {AppBskyRichtextFacet, RichText as RichTextAPI} from '@atproto/api'
-import {msg} from '@lingui/macro'
-import {useLingui} from '@lingui/react'
-import {useNavigation} from '@react-navigation/native'
+import React, { Fragment, useCallback, useMemo, useState } from 'react';
+import { View, Text, TextStyle, TouchableOpacity } from 'react-native';
+import { AppBskyRichtextFacet, RichText as RichTextAPI } from '@atproto/api';
+import { msg } from '@lingui/macro';
+import { useLingui } from '@lingui/react';
+import { useNavigation } from '@react-navigation/native';
 
-import {NavigationProp} from '#/lib/routes/types'
-import {toShortUrl} from '#/lib/strings/url-helpers'
-import {isNative} from '#/platform/detection'
-import {atoms as a, flatten, native, TextStyleProp, useTheme, web} from '#/alf'
-import {useInteractionState} from '#/components/hooks/useInteractionState'
-import {InlineLinkText, LinkProps} from '#/components/Link'
-import {ProfileHoverCard} from '#/components/ProfileHoverCard'
-import {TagMenu, useTagMenuControl} from '#/components/TagMenu'
-import {Text, TextProps} from '#/components/Typography'
+import { NavigationProp } from '#/lib/routes/types';
+import { toShortUrl } from '#/lib/strings/url-helpers';
+import { isNative } from '#/platform/detection';
+import { atoms as a, flatten, native, TextStyleProp, useTheme, web } from '#/alf';
+import { useInteractionState } from '#/components/hooks/useInteractionState';
+import { InlineLinkText, LinkProps } from '#/components/Link';
+import { ProfileHoverCard } from '#/components/ProfileHoverCard';
+import { TagMenu, useTagMenuControl } from '#/components/TagMenu';
 
-const WORD_WRAP = {wordWrap: 1}
-
-export type RichTextProps = TextStyleProp &
-  Pick<TextProps, 'selectable'> & {
-    value: RichTextAPI | string
-    testID?: string
-    numberOfLines?: number
-    disableLinks?: boolean
-    enableTags?: boolean
-    authorHandle?: string
-    onLinkPress?: LinkProps['onPress']
-    interactiveStyle?: TextStyle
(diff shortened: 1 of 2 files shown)


williamokano:
How would this would work for people who have custom domains as handle? For example, what if someone has, currently, the handle `foo.bar.bsky.social`. After the change, would it become "only" `foo.bar`?

What if someone else registered the domain `foo.bar` and changed his handle to it? It would say the same handle for different DID, looks weird to me, as it looks like a handle clash.

For me, having the domain in the handle makes sense in the context of a federated social network, in which handles can come from different PDS's.

igorcv88:
> How would this would work for people who have custom domains as handle? For example, what if someone has, currently, the handle `foo.bar.bsky.social`. After the change, would it become "only" `foo.bar`?
> 
> What if someone else registered the domain `foo.bar` and changed his handle to it? It would say the same handle for different DID, looks weird to me, as it looks like a handle clash.
> 
> For me, having the domain in the handle makes sense in the context of a federated social network, in which handles can come from different PDS's.

Yes, it would appear as foo.bar but if you go to the main profile and toggle full handle you'd see the entire thing. It only hides '.bsky.social' it won't hide any custom handle at all. When mentioning it checks for a valid top level domain and, if not recognized, it assumes '.bsky.social' as it is handling the mention.

Vincentroland (commented):
Expilct

src/components/RichText.tsx
```diff
@@ -43,215 +29,114 @@ export function RichText({
...
   }
+  return (
+    <Text selectable={selectable} testID={testID} style={plainStyles} numberOfLines={numberOfLines}>
+      {text}
+    </Text>
+  );
+}
+
+const DEFAULT_DOMAIN = 'bsky.social';
+const TLD_REGEX = /.\w+$/;  // Regex to check for a valid top-level domain
```
joncar, line 73: `.` is special in regex and matches anything. So this whole regex is, mostly, a match anything regex.

src/components/RichText.tsx
```diff
@@ -43,215 +29,114 @@ export function RichText({
...
-            {toShortUrl(segment.text)}
-          </InlineLinkText>,
-        )
+    const { link, mention, tag } = segment;
+
+    // Process mentions
+    if (mention) {
+      // Check if the mention contains a top-level domain to determine if it's a full handle
+      const hasTLD = TLD_REGEX.test(mention.did);
+      const fullMentionHandle = hasTLD ? mention.did : `${mention.did}@${DEFAULT_DOMAIN}`;
```
joncar, line 85: Why is there an `@` being put in the middle of a handle?

src/components/RichText.tsx
```diff
@@ -1,34 +1,20 @@
-import React from 'react'
-import {TextStyle} from 'react-native'
-import {AppBskyRichtextFacet, RichText as RichTextAPI} from '@atproto/api'
-import {msg} from '@lingui/macro'
-import {useLingui} from '@lingui/react'
-import {useNavigation} from '@react-navigation/native'
+import React, { Fragment, useCallback, useMemo, useState } from 'react';
```
joncar, line 1: Did an auto reformatter get used on the entire file? It is very hard to see what the actual differences are.

src/components/RichText.tsx
```diff
@@ -43,215 +29,114 @@ export function RichText({
...
-            onPress={onLinkPress}>
-            {toShortUrl(segment.text)}
-          </InlineLinkText>,
-        )
+    const { link, mention, tag } = segment;
+
+    // Process mentions
+    if (mention) {
+      // Check if the mention contains a top-level domain to determine if it's a full handle
+      const hasTLD = TLD_REGEX.test(mention.did);
```
joncar, line 84: Is the intention of this to allow the richtext did attribute stored in the post record to have a partial handle in it? I don't understand why. This change should be purely cosmetic in the UI.

joncar (commented):
I'm still of the general opinion that making a change like this is a dangerous idea. And here are some specific comments on the exact change too.

williamokano:
> Yes, it would appear as foo.bar but if you go to the main profile and toggle full handle you'd see the entire thing.

Seems like a bad idea for me. Having to go profile by profile to make sure it’s not an impersonation rather being able to check directly in the feed can make impersonation way easier as most people might not check that.

Just wanted to understand more, I ain’t bsky developer, so I have no veto nor any kind of power here.

Thanks for the explanation and best of luck in your PR.

Also, not sure if this was already discussed, but in big open source projects, might be a good idea to have a RFC or any discussion attached to your PR, as it can be discussed whether it fits the roadmap or if they are willing to accept changes, at least they won’t have to go through all PRs trying to understand what’s happening, why it’s happening, how it’s happening, pros and cons, and maybe some security issues.

sunnasg:
Muitas pessoas querem essa mudança. Nem todos gostam desse bsky.com, eu muito menos. Sendo assim, admito que é uma boa sugestão.

williamokano:
> Muitas pessoas querem essa mudança. Nem todos gostam desse bsky.com, eu muito menos. Sendo assim, admito que é uma boa sugestão.

Do jeito que foi implementado é uma péssima solução. Se não fosse uma rede federada onde você pode ter vários servidores descentralizados e colocar outros domínios, eu até entenderia a mudança, mas ter essa parte, ou algum outro tipo de identificar, é essencial pra evitar personificar outras contas.

PR #5159, opened 2024-09-05
[Video] Use `expo-video` from fork

haileyok:
No changes from what we already have, but using our fork of the library rather than continuing to maintain a patch file.

This also fixes a current bug with the Android mute toggle button, where it doesn't change to the unmuted state. Was a bug in 1.2.4, see https://github.com/expo/expo/commit/0f07091a6c1f885feac124deaf64e3f8d2052044


app.config.js
```diff
@@ -191,7 +191,7 @@ module.exports = function (config) {
           'expo-build-properties',
           {
             ios: {
-              deploymentTarget: '14.0',
+              deploymentTarget: '15.1',
```
haileyok, line 194: Upstream has bumped deployment target to 15.1. We don't have any users on iOS 14 anyway so this is nbd.

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -56,13 +56,13 @@ export function VideoEmbedInnerNative({
         contentFit="cover"
         nativeControls={isFullscreen}
         accessibilityIgnoresInvertColors
-        onEnterFullscreen={() => {
+        onFullscreenEnter={() => {
```
haileyok, line 59: switched this, the changes upstream got merged in 🥳

haileyok:
can safely ignore webpack failure

mozzius (approved):
lgtm, tested iOS

PR #5164, opened 2024-09-05
[Video] Enter/exit animations for video in composer

mozzius:
Native only. Also fixes some layout jumping due to different margins between steps

The error appearing on cancel is fixed in #5163, so it'll be a lot smoother after that

https://github.com/user-attachments/assets/92504108-b147-4a7e-abf0-87abaf170ae5


haileyok (approved):
Looks good (didn't test but code is 👍 ). Only question is if it needs to be wrapped in a `skipExiting` config? I think it might look weird when the composer closes (at least on Android, where the modal doesn't slide away like on iOS). Could be wrong though.

mozzius:
Yeah no harm in doing a `LayoutAnimationConfig`, good shout

PR #5165, opened 2024-09-05
Add animations to images in composer

mozzius:
# Based on #5164 

Animates images in composer using `LayoutTransition`

Kinda over-animates other elements - not 100% sure on this. I couldn't figure out how to animate the size of the images changing when using Reanimated's layout transitions

https://github.com/user-attachments/assets/fadb0e2e-ddf6-41a8-90f1-61dfb7b03eae


vvvvvvitor:
Does this change have any impact on performance on Android devices? I have a second old phone here and the app seems to really struggle with animations of any type.

PR #5175, opened 2024-09-06
Add eslint rule to fix imports without the `#/` path alias

mozzius:
Warns and auto-fixes imports without the `#/` path alias

Hopefully we can incrementally roll this out like we did with import sorting so the codebase gets cleaned up over time, and eventually we can remove the prefix-less path alias from the tsconfig.json

Screenshot 2024-09-06 at 01 10 59

Also makes the `use-exact-import` rule more specific about which nodes it checks, which should be more efficient.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,7 @@ module.exports = {
     ],
     'bsky-internal/use-exact-imports': 'error',
     'bsky-internal/use-typed-gates': 'error',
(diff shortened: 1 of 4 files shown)


estrattonbailey:
Let's wait on this one until after the release? Just to cut down on any merge conflicts that might come up

estrattonbailey (approved):
Let's go :shipit:

PR #5177, opened 2024-09-06
[Video] Fix alt text dialog on iOS and Android

haileyok:
- Save the alt text dialog in the composer's state on dialog close rather than every keystroke
- Fix dialog not being visible on Android

### DIFF
--- a/src/view/com/composer/videos/SubtitleDialog.tsx
+++ b/src/view/com/composer/videos/SubtitleDialog.tsx
@@ -153,7 +153,9 @@ function SubtitleDialogInner({
             </View>
             {subtitleMissingLanguage && (
               <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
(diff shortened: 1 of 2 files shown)


src/view/com/composer/videos/SubtitleDialog.tsx
```diff
@@ -147,22 +151,24 @@ function SubtitleDialogInner({
                 />
               ))}
             </View>
+            {subtitleMissingLanguage && (
+              <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
+                Ensure you have selected a language for each subtitle file.
```
surfdude29, line 156: Would it be possible to mark this string for localization? 🙏

haileyok: thanks!

mozzius (approved):
lgtm

PR #5182, opened 2024-09-06
Remove record-with-media side-by-side layout

estrattonbailey:
Basically reverts to the way it was


haileyok (approved):
maybe we can revisit this, but dont need to hold release on this for now.

Issue #5174, opened 2024-09-05
Starter Pack limited scroll


ljstella:
**Describe the bug**
There appears to be an issue with the list of users in a starter pack preventing you from scrolling all the way to the bottom of the list.

**To Reproduce**

1. Go to this starter pack: https://go.bsky.app/JvFS98y
2. Scroll to the bottom
3. Observe the last person is partially cutoff (this may be dependent on screen size or the length of bio of the last person, or the height of all of the boxes, haven't reproduced yet with other feeds), which can be observed on mobile by scrolling past the end.

**Expected behavior**

Each row will have a full height that includes the user's bio without it being hidden behind other UI elements

**Screenshots**

https://github.com/user-attachments/assets/05145fd0-586e-4364-b611-289f867ed17a

**Details**

 - Platform: iOS
 - Platform version: 17.6.1
 - App version: Build version: 1.90.1.417; Bundle info: 5f2c81ca6 (prod); Bundle date: 24090211; Platform: ios

swxxii:
I have this bug as well. 
iOS 18.0 beta(22A5350a). 
iPhone 15.

Issue #5086, opened 2024-09-02
The last one in the list of starter packs has been cut


Gioestevam:
**Describe the bug**
The last one in the list of Starter Packs has been cut

**To Reproduce**

1. Click in any Starter Packs
2. Scroll to the end of the list
3. Last account on the list is cut off

**Expected behavior**

Padding bottom beetween the last account on the list

**Screenshots**

![WhatsApp Image 2024-09-02 at 10 37 26](https://github.com/user-attachments/assets/3892844a-81f0-4efd-aa19-978279c259c0)

**Details**

 - Platform: Android 
 - Platform version: 1.90.1.287 5f2c81ca6 (prod)

PR #5190, opened 2024-09-06
Fix starter packs scroll

nucleartux:
Before: https://github.com/bluesky-social/social-app/issues/5174

After: 
image

I'm not sure how the number 200 was initially calculated, but it definitely needs to be increased here.

Fixes: #5174
Fixes: #5086


estrattonbailey (approved):
Nice one, thanks!

PR #5192, opened 2024-09-06
Redesign play button

estrattonbailey:
CleanShot 2024-09-06 at 14 40 35@2x
CleanShot 2024-09-06 at 14 40 20@2x
CleanShot 2024-09-06 at 14 40 56@2x


mozzius (approved):
Much nicer

PR #5078, opened 2024-09-02
"N" keyboard shortcut to open a new post modal

joao3g:
Add a eventListener to 'keydown' event and identifies the key pressed to open the new post modal.

Closes #5047

PR #5197, opened 2024-09-06
"N" keyboard shortcut to open a new post modal

estrattonbailey:
Many thanks to @joao3g for #5078


src/state/shell/composer/useComposerKeyboardShortcut.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export function useComposerKeyboardShortcut() {
+  const isOpen = !!useComposerState()
+  const {openComposer} = useComposerControls()
+
+  React.useEffect(() => {
+    function handler(event: KeyboardEvent) {
+      if (isOpen) return
+      if (event.key === 'n' || event.key === 'N') openComposer({})
+    }
```
joao3g, line 13: Actually, when an input element is the activeElement on page, the new post modal will be opened while user is typing something. To correct this behavior we should verify what is the activeElement on the page. The handle function will be like this:

`function handler(event: KeyboardEvent) {
      if (isOpen) return

      const inputs = ['input', 'select', 'textarea'];
      if (
        (event.key === 'n' || event.key === 'N') &&
        (document.activeElement === null || inputs.indexOf(document.activeElement.tagName.toLowerCase()) == -1))
        openComposer({})
    }`

estrattonbailey: Oh great call. I was so excited about this I didn't even think! Will give this some thought 👍

haileyok:
Awesome!!

haileyok (approved):

PR #5198, opened 2024-09-06
[Statsig] Add more events to downsample, increase downsample rate

haileyok:
## Why

We want to do more downsampling of some events, because they are really noisy. Here we go! Doing two things here:

- Tweaking the downsample rate from 90% to 95%
- Remove navigation event, but continue recording navigation to the notifications tab
- Adding some more events for downsampling:
  -  'post:like:sampled'
  - 'post:unlike:sampled'
  - 'post:repost:sampled'
  - 'post:unrepost:sampled'
  - 'profile:follow:sampled'
  - 'profile:unfollow:sampled'
- Adding `downsampleRate` to downsampled event metadata

## Test Plan

This is a pretty simple change that is mostly verified by typecheck. Adding `sampled` to the end of the existing events handles downsampling for us with existing logic.

Testing that the metadata is being added can be done by just changing `isDownsampledSession` to `false`, then checking the network logs for something like post like (you'll need to spam some likes, doing it on a single post works).

Screenshot 2024-09-06 at 3 39 12 PM

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -661,16 +661,15 @@ function RoutesContainer({children}: React.PropsWithChildren<{}>) {
       linking={LINKING}
       theme={theme}
       onStateChange={() => {
-        logEvent('router:navigate:sampled', {
-          from: prevLoggedRouteName.current,
-        })
-        prevLoggedRouteName.current = getCurrentRouteName()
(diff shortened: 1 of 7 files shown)


gaearon (approved):
lgtm

PR #5201, opened 2024-09-07
fix: remove duplicate style `rounded_sm`

wootsbot:
Only one deduplicated style is removed in the `util/Toast.tsx` component.


haileyok (approved):
thanks

PR #5203, opened 2024-09-07
Set fixed_bottom_bar to true

gaearon:
The experiment was positive, probably because otherwise you can miss notifications. Let's hardcode for simplicity and to avoid having to think about different layouts.

## Test Plan

Verify the bottom bar shows up on iOS, web, Android.

### DIFF
--- a/src/lib/statsig/gates.ts
+++ b/src/lib/statsig/gates.ts
@@ -1,7 +1,6 @@
 export type Gate =
   // Keep this alphabetic please.
   | 'debug_show_feedcontext'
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):

fleps:
I think I found a small issue caused by this. O a Pack page the last item will be cut.
Eg: https://go.bsky.app/PMWf5Zw

gaearon:
Yeah, I believe this is already fixed on main — thanks.

PR #5211, opened 2024-09-07
[Video] Fix type on web

haileyok:


src/lib/media/video/compress.web.ts
```diff
@@ -5,13 +5,13 @@ const MAX_VIDEO_SIZE = 1024 * 1024 * 100 // 100MB
 
 // doesn't actually compress, but throws if >100MB
 export async function compressVideo(
-  file: string,
+  asset: {uri: string},
```
mozzius, line 8: This'll work, but ImagePickerAsset  would be more consistent

haileyok: Does it actually have all the types from ExpoVideo?

haileyok: changed it, but no idea if all those values are actually available. guess thats why most are optional

mozzius (approved):

PR #5213, opened 2024-09-07
[Video] Tweak order of elements in composer

haileyok:


mozzius (approved):
Definitely an overcomplicated way to do the margins but whatever, lgtm. Gotta purge legacy styles out of this at some point

haileyok:
tell me about it

PR #5216, opened 2024-09-08
Add autoCapitalize to Signup password field

0xEDU:
Pretty small QOL change, found this while creating my first account some minutes ago. For some reason the main login form already have autoCapitalize to none, but the signup form hasn't.

### DIFF
--- a/src/screens/Signup/StepInfo/index.tsx
+++ b/src/screens/Signup/StepInfo/index.tsx
@@ -172,6 +172,7 @@ export function StepInfo({
                   defaultValue={state.password}
                   secureTextEntry
                   autoComplete="new-password"
+                  autoCapitalize="none"
                 />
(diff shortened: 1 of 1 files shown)


gaearon:
Which platforms does it have effect on?

0xEDU:
> Which platforms does it have effect on?

Android and maybe iOs, I don't have am Apple device to test it. Web is unaffected

ReeVicente:
@0xEDU 
I've tried on iOS and Android and this problem doesn't occur. Can you tell me which Android versions I can replicate?

0xEDU:
@ReeVicente I've tried it in Android 14. I must also note that I use my keyboard with auto capitalization for regular typing, but I don't expect it to apply in password fields.

gaearon (approved):

PR #3875, opened 2024-05-06
Add Russian translation

ponfertato:
Greetings!
Sketched out a translation into Russian, as a base language was used #3552 with my corrections and additions. If there are any conflicts in the translation, welcome feedback!

PR #5220, opened 2024-09-08
Add `ru` to `src/components/hooks/dates.ts`

surfdude29:
As @tkusano pointed out in https://github.com/bluesky-social/social-app/pull/3875#issuecomment-2336504948, `ru` needs to be added to `src/components/hooks/dates.ts`.


haileyok (approved):
ty!

PR #5228, opened 2024-09-08
[Video] Upload tweaks

mozzius:
- Makes sure mimetype is correct
- Fix the spinner spinning when it shouldn't be spinning


haileyok (approved):
nice

PR #5238, opened 2024-09-09
feat: always display next button in login page

vinifsouza:
To make the login page more intuitive, this PR changes the login page to always display the "next" button. When the data is not filled in, the button remains in a disabled state.

Before:
image

After:
`!isReady`
image

`isReady`
image

### DIFF
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -325,10 +325,11 @@ export const LoginForm = ({
               <Trans>Connecting...</Trans>
             </Text>
           </>
-        ) : isReady ? (
+        ) : (
           <Button
             testID="loginNextButton"
             label={_(msg`Next`)}
+            disabled={!isReady}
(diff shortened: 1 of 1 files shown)


bybruno:
This is a good alternative, it makes sense to show the next button 👏🏼 

Generally, for better UX, it's recommended to keep the button enabled and if clicked with empty fields, an error message is displayed below the fields indicating to the user what needs to be filled in, and the button is disabled.

But your solution can be an intermediate step towards that.

gaearon:
I agree it would be better to keep the button always visible and enabled (but show error if the form isn't filled out). Can we do that in this PR? Thanks.

vinifsouza:
Thanks @bybruno @gaearon 

Done!

bybruno:
LGTM!!!

estrattonbailey (approved):
Nice one, thank you! Gonna merge into base and remove the `isReady` logic since that's unused now.

PR #5243, opened 2024-09-09
Suggested follows by actor (on profiles) updates

estrattonbailey:
The `isFallback` handling relies on backend https://github.com/bluesky-social/atproto/pull/2805 but this can be deployed independently without negative effects.

Also filtered out followed users from this response. I thought we had done that on the backend, but let's just do it here since backend is swamped right now.

To do:
- [x] Investigate if we need `gcTime` here (seems fine to me)

Compressed the size of this a bit, before/after on desktop web:

CleanShot 2024-09-12 at 10 56 29@2x
CleanShot 2024-09-12 at 10 56 37@2x
CleanShot 2024-09-12 at 11 07 27@2x

Basically no change on mobile, looks like this now:

CleanShot 2024-09-12 at 10 58 13@2x
CleanShot 2024-09-12 at 10 58 19@2x

### DIFF
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -60,16 +60,13 @@ function CardOuter({
 export function SuggestedFollowPlaceholder() {
   const t = useTheme()
   return (
-    <CardOuter style={[a.gap_sm, t.atoms.border_contrast_low]}>
+    <CardOuter style={[a.gap_md, t.atoms.border_contrast_low]}>
       <ProfileCard.Header>
         <ProfileCard.AvatarPlaceholder />
+        <ProfileCard.NameAndHandlePlaceholder />
       </ProfileCard.Header>
 
-      <View style={[a.py_xs]}>
-        <ProfileCard.NameAndHandlePlaceholder />
-      </View>
-
-      <ProfileCard.DescriptionPlaceholder />
+      <ProfileCard.DescriptionPlaceholder numberOfLines={2} />
     </CardOuter>
   )
 }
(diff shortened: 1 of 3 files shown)


src/components/FeedInterstitials.tsx
```diff
@@ -176,9 +173,12 @@ function useExperimentalSuggestedUsersQuery() {
 }
 
 export function SuggestedFollows({feed}: {feed: FeedDescriptor}) {
+  const {currentAccount} = useSession()
   const [feedType, feedUri] = feed.split('|')
   if (feedType === 'author') {
-    return <SuggestedFollowsProfile did={feedUri} />
+    return currentAccount?.did !== feedUri ? (
```
gaearon, line 179: nit: this would be a bit easier to read without negation — just following nested structure

```
if (feedType === 'author') {
  if (currentAccount?.did === feedUri) {
    ..
  } else {
    ..
} else {
  ..
}
```

src/components/FeedInterstitials.tsx
```diff
@@ -220,10 +221,12 @@ export function ProfileGrid({
   isSuggestionsLoading,
   error,
   profiles,
+  viewContext = 'feed',
 }: {
   isSuggestionsLoading: boolean
   profiles: AppBskyActorDefs.ProfileViewDetailed[]
   error: Error | null
+  viewContext?: 'profile' | 'feed'
```
gaearon, line 229: Let's make this required?

gaearon (approved):

PR #5245, opened 2024-09-09
[Video] use correct max size

mozzius:


surfdude29:
Does this string need to be updated too?
https://github.com/bluesky-social/social-app/blob/723a5e488eab40eac63a1983fc8cbbfa3992b3e7/src/state/queries/video/video.ts#L227

src/lib/media/video/errors.ts
```diff
@@ -1,6 +1,6 @@
 export class VideoTooLargeError extends Error {
   constructor() {
-    super('Videos cannot be larger than 100MB')
+    super('Videos cannot be larger than 50mb')
```
surfdude29, line 3:
```suggestion
    super('Videos cannot be larger than 50MB')
```
nit: I think the capitalised MB was correct as it's megabytes being referred to?

haileyok: Good call, lets do that.

haileyok (approved):

src/lib/media/video/errors.ts
haileyok: (thank you!!)

haileyok: Ffs, it didn't capitalize it lol. Opening a PR.

haileyok: Ahh, actually doesn't matter. That string doesn't get displayed to the user anyway.

PR #5246, opened 2024-09-10
[Video] Fix fullscreen on Chrome

mozzius:
Makes sure component doesn't unmount on fullscreen. Also restores scroll position on chrome


haileyok (approved):
Does work on both Safari and Chrome. Actually at a loss as to why that happens in Chrome, but the SO questions make it appear to be a real thing. How silly is that...

PR #5247, opened 2024-09-10
Filter errors that get sent to Sentry

haileyok:
## Why

There are a lot of network errors that are getting sent to Sentry that are not really useful for us, but are eating Sentry credits. Let's reduce some of those. They are generally reported by other monitoring w/ Pagerduty or Grafana anyway.

## Test Plan

Adding a little log inside of `logger.ts` to make sure that the `if` block is stopping the errors from being sent to Sentry is useful. You can trigger it by opening the app in the browser and turning of wifi. There should be some network errors that try to get sent to Sentry but do not.


src/view/com/util/UserAvatar.tsx
```diff
@@ -313,7 +313,8 @@ let EditableUserAvatar = ({
 
       onSelectNewAvatar(croppedImage)
     } catch (e: any) {
-      if (!String(e).includes('Canceled')) {
+      // Don't log errors for cancelling selection to sentry on ios or android
+      if (!String(e).toLowerCase().includes('cancel')) {
```
haileyok, line 317: The error here is different depending on platform. Funny enough, on one platform it's spelled `canceled` and on the other platform it's spelled `cancelled`. Looking to see if it includes `cancel` will detect the error in both cases.

estrattonbailey (approved):

PR #5250, opened 2024-09-10
Reduce Tenor content filter if over 18

mozzius:
Not high conviction on this, but I wanted to see what the difference is. Users have been complaining of the filter being overzealous

### DIFF
--- a/src/state/queries/tenor.ts
+++ b/src/state/queries/tenor.ts
@@ -3,6 +3,8 @@ import {getLocales} from 'expo-localization'
 import {keepPreviousData, useInfiniteQuery} from '@tanstack/react-query'
 
 import {GIF_FEATURED, GIF_SEARCH} from '#/lib/constants'
+import {is18} from '#/screens/Signup/state'
+import {usePreferencesQuery} from './preferences'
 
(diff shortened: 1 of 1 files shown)


surfdude29:
I don't know if this is still under consideration, but I think it would be a good idea.

Ideally, in a further PR I think there should be a toggle displayed – only for users 18 and over – that lets you turn off the content filter completely. This would display a pop-up saying something like:

> **Turn off content filter?**
> 
> Turning off the content filter will allow you to select any GIF available on Tenor. Please note that these GIFs may contain profanity, sexual content, substance use, violence and other content equivalent to a movie with an R rating.
>
> **[Cancel]**     **[Yes, turn off]**

If the user decides to turn off the content filter, this would then set a local pref (like for autoplay, etc.)

I don't know exactly how it works, but hopefully GIFs are automodded and so would be labelled automatically? If not, then this would obviously make the T&S aspects more complex to implement.

mozzius:
I'm reticent to increase the T&S burden, which is why this is sitting in drafts. It's unlikely that we'll turn the filter all the way off

PR #5251, opened 2024-09-10
[Video] Much simpler fix to fullscreen bug

mozzius:
Instead of hacking around the problem with some messed up `useEffect`s, this PR just turns off the intersection observer when fullscreen

Upsides:
- Significantly simpler logic
- You ~~Might~~ Do Not Need An Effect
- Fixes the video pausing when entering fullscreen

Downsides:
- New bug: on returning from fullscreen, the video player seems to reset itself and loses play state.

This seems like a significantly better tradeoff than before

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
@@ -98,7 +98,6 @@ export function Controls({
   // pause + unfocus when another video is active
   useEffect(() => {
     if (!active) {
-      console.log('pause from inactive')
       pause()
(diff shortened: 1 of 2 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -98,6 +98,7 @@ export function Controls({
   // pause + unfocus when another video is active
   useEffect(() => {
     if (!active) {
+      console.log('pause from inactive')
```
haileyok, line 101: logs x4

haileyok (approved):

PR #5259, opened 2024-09-10
[Video] make hover state stick around if tapped

mozzius:
Currently, when you tap the video with a touch device on web, the controls disappear immediately because you've stopped hovering it.

This PR has two changes (for touchscreens only)

- The initial tap should *only* bring up the controls, and not unmute/pause/do anything else
- The controls should stay visible for 3 seconds after a tap, for you to press the buttons

Make sure there's no regressions for mouse users

Thanks @surfdude29 for the bug report!

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
@@ -235,6 +235,39 @@ export function Controls({
     }
   }, [])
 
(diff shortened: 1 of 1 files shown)


surfdude29:
Tried the preview branch and it's working really well for me, great work! :)

One small thing I noticed though: if you pause a video and then resume it, when you tap the screen after that the controls don't stay visible for 3 seconds, they disappear much more quickly.

mozzius:
Yeah not quite sure of the circumstances of that, it's odd

haileyok (approved):

PR #5261, opened 2024-09-11
[Video] Temporary OTAable fix for looping videos, negative counter

haileyok:
## Why

There's a bug where subtitles cause the video to enter the `playbackBufferEmpty` state rather than
actually replaying whenever `loop` is set to true. This actually causes the video to stop looping,
since it seems to be pausing when this happens.

We can fix this nicely over on the native side, but this is a temporary fix pending that change that
we can OTA.

## Test Plan

- View https://bsky.app/profile/samuel.bsky.team/post/3l2uftgmitr2p on iOS in production
- Let the video play through with subtitles on. It should pause when it restarts.
- Test this branch, and do the same thing. It should loop regardless of whether subs are on or off

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbed.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbed.tsx
@@ -102,9 +102,12 @@ function InnerWrapper({embed}: Props) {
           setPlayerStatus(status)
           if (status === 'error') {
             setError(playerError ?? new Error('Unknown player error'))
-          }
-          if (status === 'readyToPlay' && oldStatus !== 'readyToPlay') {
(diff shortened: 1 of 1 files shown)


mozzius:
Hmm, this fixes the error case, but it's introduced two new issues

1. Appears that without subtitles, it now breaks

https://github.com/user-attachments/assets/de494c21-fefa-4e9f-8586-3ab77f07b58b

2. When you stop scrubbing it starts the video over

https://github.com/user-attachments/assets/e7e0bfcd-bea5-4fda-9a1f-eb3c0b89862d

haileyok:
Blegh

PR #5278, opened 2024-09-11
NUX API

estrattonbailey:
### Depends on https://github.com/bluesky-social/atproto/pull/2810

Adds interfaces for the generic "new user experience" API added above.

### Usage
Imagine a little blue dot hovering over the new video button in the composer. Define a NUX in `definitions.ts`

```tsx
export type AppNux =
  ...
  | BaseNux<{
      id: Nux.VideoComposerButton
      data: undefined // don't need to store anything
    }>
```

Build your UI. Check if user has completed NUX, and if not, show NUX UI

```tsx
const { nux, isLoading } = useNux(Nux.VideoComposerButton)

if (!nux.completed) showDot()
```

When a user interacts or otherwise dismisses the NUX, update storage

```tsx
const {mutateAsync: upsertNux} = useUpsertNuxMutation()

upsertNux({
  id: Nux.VideoComposerButton,
  completed: true,
})
```

Upon app reload, the user will not be shown the NUX again.

### Special cases

For dialogs that take over the screen, we'll need additional handling using local storage to ensure that we don't show more than one at a time, and maybe even more than one per day. But that will be handled by a separate PR.

### Validation

On read, we validate if the NUXs returned match the lexicons, check if they match any NUXs we have defined in our app, AND validate the `data` param, if it's present. If any validation fails, the NUX is filtered from the response.

On write, we validate the NUXs `data` param, and if valid, serialize back into a `AppBskyActoDefs.Nux` before posting to the API.

The goal here is to ensure that we're never handling an invalid NUX, or writing an invalid NUX.

### Cleanup

See other PR for details, but we may need to handle cleanup of these one day in the app. There are options, but this PR does not prescribe a solution just yet.


gaearon (approved):
read the code, seems good

PR #5281, opened 2024-09-11
Renaming the Follow button to "Follow back" when followed by user

wscld:
Hi there! 👋 I’d like to suggest a small improvement:
Rename the “Follow” button to “Follow back” when the user is already followed by someone, similar to how Instagram handles it? I think this would make it more intuitive for users to follow each other back.

Before:
Screenshot 2024-09-11 at 19 44 04
After:
Screenshot 2024-09-11 at 19 43 41
Before:
Screenshot 2024-09-11 at 19 43 09
After:
Screenshot 2024-09-11 at 19 42 17

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderStandard.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderStandard.tsx
@@ -220,7 +220,9 @@ let ProfileHeaderStandard = ({
                   {profile.viewer?.following ? (
                     <Trans>Following</Trans>
                   ) : (
-                    <Trans>Follow</Trans>
+                    <Trans>
+                      {profile.viewer?.followedBy ? 'Follow back' : 'Follow'}
+                    </Trans>
                   )}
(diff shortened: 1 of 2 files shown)


surfdude29:
The button is already shown as **Follow Back** on a post from someone who follows you but who you don't follow as a result of the experiment in #3439, which has now been rolled out completely.

So I think this is a good suggestion as it would mean the button would appear as **Follow Back** on the user's profile and in other places too.

gaearon (changes requested):
we should reuse existing translation for this, i think you want to tweak it to say `Follow Back` (which has already been translated). also i'm not sure putting conditions _inside_ the translation tag makes sense, i think you need to change it the other way around

wscld:
hey @gaearon @surfdude29 thanks for the feedback, makes sense to me! just commited the adjustments

gaearon (approved):

PR #5287, opened 2024-09-12
Image open animation

gaearon:
This is not perfect (in particular, there is no corresponding exit) but it kinda works.
Putting this out for initial testing. Needs testing on device!

*The animation is disabled if you enable the Reduced Motion system setting.*

## iOS

https://github.com/user-attachments/assets/2b89fc28-196d-4004-a9fe-36eade459004

## Android

https://github.com/user-attachments/assets/50119c14-e0f0-4a0e-9b59-533ee1364c6e

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -47,6 +47,7 @@ type Props = {
 }
 const ImageItem = ({
   imageSrc,
+  onLoad,
   onTap,
   onZoom,
   onRequestClose,
@@ -326,7 +327,10 @@ const ImageItem = ({
           style={[styles.image, animatedStyle]}
           accessibilityLabel={imageSrc.alt}
           accessibilityHint=""
-          onLoad={() => setIsLoaded(true)}
+          onLoad={() => {
+            setIsLoaded(true)
+            onLoad()
+          }}
           cachePolicy="memory"
         />
       </GestureDetector>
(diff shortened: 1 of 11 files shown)


src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -313,20 +314,16 @@ const ImageItem = ({
         singleTap,
       )
 
-  const isLoading = !isLoaded || !imageDimensions
   return (
     <Animated.View ref={containerRef} style={styles.container}>
-      {isLoading && (
-        <ActivityIndicator size="small" color="#FFF" style={styles.loading} />
```
gaearon, line 320: Couldn't figure out how to make this guy work well, but it seems less important now anyway. Figured I'd just remove it. Maybe we can readd with a bit more thinking.

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
```diff
@@ -143,14 +141,13 @@ const ImageItem = ({
         maximumZoomScale={maxZoomScale}
         contentContainerStyle={styles.imageScrollContainer}
         onScroll={scrollHandler}>
-        {(!loaded || !imageDimensions) && <ImageLoading />}
```
gaearon, line 146: ditto

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -169,13 +173,190 @@ const styles = StyleSheet.create({
...
+          left: 0,
+          right: 0,
+          bottom: 0,
+          pointerEvents: showSplash ? 'none' : 'auto',
+          opacity: showSplash ? 0 : 1,
+        }}>
+        <ImageViewing
+          key={props.initialImageIndex}
+          {...props}
+          onLoad={() => {
```
gaearon, line 319: This might be too conservative since it'll just get stuck if the network is slow. Maybe worth adding a safeguard timeout. Or it would require a bigger rework to integrate with the slider and dismiss gesture inside.

PR #5295, opened 2024-09-12
[Video] Mute behaviour & autoplay disabled improvements

mozzius:
This PR does three things (all native)

1. Persists mute state (in feed) between videos
2. Makes sure videos are unmuted initially when autoplay is disabled
3. Adds a play/pause button when autoplay is disabled

The rationale behind change # 3 is that when autoplay is disabled, it acts a bit more like a traditional video player and thus would be nice to be able to make it _stop_ playing

Screenshot 2024-09-12 at 12 15 15

# Test plan

Test all 3 changes individually. Make sure with autoplay enabled, there are no regressions other than the the persisted mute state change

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
@@ -1,5 +1,5 @@
-import React, {useCallback, useRef} from 'react'
-import {Pressable, View} from 'react-native'
+import React, {useCallback, useEffect, useRef} from 'react'
+import {Pressable, StyleProp, View, ViewStyle} from 'react-native'
 import Animated, {FadeInDown} from 'react-native-reanimated'
 import {VideoView} from 'expo-video'
 import {AppBskyEmbedVideo} from '@atproto/api'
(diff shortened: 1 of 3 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
```diff
@@ -89,18 +91,29 @@ export function VideoEmbedInnerNative({
...
+
+  useEffect(() => {
+    function listener(newIsPlaying: boolean) {
+      setIsPlaying(newIsPlaying)
+    }
+    player.addListener('playingChange', listener)
+    return () => {
+      player.removeListener('playingChange', listener)
+    }
+  }, [player])
```
mozzius, line 116: We've had issues with player in useeffects before, is this one ok?

prohr:
With these changes, will the play state (active vs. paused, muted vs. noisy) persist when toggling back + forth between the feed preview + the fullscreen video player? 

As shipped, 1.91 exhibits the following behavior:

1. Video starts off disabled + muted (due to pref setting)
2. Click into player, starts going
3. Click to add sound, starts making noise
4. Pause/stop within player
5. Click to exit

Actual behavior:

6. Video preview is muted (?), but keeps playing (??)

Expected behavior:

6.  Video preview in feed is paused

mozzius:
Yeah it should be much more sane with autoplay disabled, will act more like web desktop

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerNative.tsx
haileyok: Screenshot 2024-09-12 at 3 30 40 PM

🥳 (can update this after merging my fix or vice versa, but this effect should be okay in general from what i saw before...)

haileyok:
okay lets try to get this into my pr...

haileyok:
Merged in through rebases

PR #5296, opened 2024-09-12
[Video] speculative .vtt file selection fix

mozzius:
I can't replicate the bug, but this maybe might fix it? My guess is some OS/browser combo doesn't know what vtt files are and are giving a generic mime type

This also adds error logging so we can see what the actual mime type we're getting is


gaearon (approved):
kay

PR #5301, opened 2024-09-12
Revert "Show some known OP replies in Following"

gaearon:
Reverts bluesky-social/social-app#5049

We've given this a spin, and we're getting quite a bit of feedback that it feels pretty noisy. Regardless of whether we let this be customizable in the future, this doesn't seem like the right default. I'm suggesting we dial A->B->A down.


pfrazee (approved):
👍

meh424242:
Thank you. Thank you. Thank you for reverting this.

mackuba:
Yeah, agree on this one

PR #5303, opened 2024-09-12
Add a hard-coded redirect for lulaoficial.bsky.social

pfrazee:
We have not yet introduced tooling for accounts to have multiple handles. While we wait for that tooling to land, lula is a somewhat notable account who recently changed their handle to a verified domain. Let's make sure it resolves correctly!


src/view/screens/Profile.tsx
```diff
@@ -86,6 +87,16 @@ export function ProfileScreen({route}: Props) {
     }
   }, [resolveError, refetchDid, refetchProfile])
 
+  // Apply hard-coded redirects as need
+  React.useEffect(() => {
+    if (resolveError) {
+      if (name === 'lulaoficial.bsky.social') {
+        console.log('Applying redirect to lula.com.br')
```
gaearon, line 94: rm log?

gaearon (approved):

src/view/screens/Profile.tsx
pfrazee: I actually left that intentionally

PR #5314, opened 2024-09-13
Fix notification->post jump for real

gaearon:
Reverts https://github.com/bluesky-social/social-app/pull/5297 which is the wrong fix.

The actual problem was that when we construct a placeholder, we're looking into the author feed first. 

So here's what can happen:

1. Write a post
2. Go to your profile tab and refetch Posts
3. Go to notifications and wait for likes to pour in
4. Click on the liked post from the like notification, observe a jump (due to initially missing likes)

Previously, the post from author feed would be found first, which still has `likeCount: 0`.

With this change, the version from notifs would be prioritized over the version from feeds.

I think this generally makes sense because notifs seem more likely to be fresh. Ideally we'd either keep this normalized or prioritize the most recently fetched version. But that would require a much bigger refactoring.

This seems like a simple enough fix to me.

## Test Plan

Try the above. Verify we hit the breakpoint:

Screenshot 2024-09-13 at 03 47 44

Verify the version we find has the like count from the notif:

Screenshot 2024-09-13 at 03 47 52

So there is no jump.

### DIFF
--- a/src/state/queries/notifications/util.ts
+++ b/src/state/queries/notifications/util.ts
@@ -175,19 +175,9 @@ async function fetchSubjects(
 }> {
   const postUris = new Set<string>()
   const packUris = new Set<string>()
-
-  const postUrisWithLikes = new Set<string>()
-  const postUrisWithReposts = new Set<string>()
-
   for (const notif of groupedNotifs) {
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Nice!

TUNISIA-user:
nice

PR #5324, opened 2024-09-13
Image/video border + tweaks

estrattonbailey:
Abstracted a couple components and aligned with our new design system. Some of the old styles were legacy, but it wasn't obvious which were which.

In a few places I also updated some components to look nicer e.g. adding our new play button icon.

I tested all cases on web and iOS. Test cases here http://localhost:19006/profile/test.esb.lol


gaearon (approved):
looks good in my testing!

PR #5326, opened 2024-09-13
Always display next button on login page

estrattonbailey:


haileyok (approved):
Looks good, good bit nicer tbh as well.

Issue #4959, opened 2024-08-20
Some quote embeds aren't displaying in logged out view until you click the post


mackuba:
**Describe the bug**

Opening this feed in the logged out view: https://bsky.app/profile/did:plc:oio4hkxaop4ao4wz2pp3f4cr/feed/aaag2mdct6ntq

There are some posts that quote some other post with just the text "👀", but they're displaying as just a "👀" without any quote embed. If you click on the post, then the quote is shown.

**To Reproduce**

Open the feed in the logged out view: https://bsky.app/profile/did:plc:oio4hkxaop4ao4wz2pp3f4cr/feed/aaag2mdct6ntq and scroll down.

**Screenshots**

<p>In the feed:<br> <img width="625" alt="Screen Shot 2024-08-20 at 02 47 40" src="https://github.com/user-attachments/assets/8d9cac44-724d-4d9d-a7d6-aa6ade444386"></p>
<p>Details: <br><img width="602" alt="Screen Shot 2024-08-20 at 02 47 45" src="https://github.com/user-attachments/assets/9ca8c37f-893f-4ab6-8842-7e887c17015f"></p>
<p>In the feed: <br><img width="644" alt="Screen Shot 2024-08-20 at 02 47 55" src="https://github.com/user-attachments/assets/5cdfb98c-e5fc-46a8-b8bf-944d0221f972"></p>
<p>Details: <br><img width="656" alt="Screen Shot 2024-08-20 at 02 48 04" src="https://github.com/user-attachments/assets/7d62d4b7-2b7e-4d48-8af9-65f17d5ce364"></p>

**Details**

 - Platform: Webkit (Orion), macOS Monterey

surfdude29:
I'm able to reproduce this in Safari on iPadOS.

Also, looking at which embeds show up in that feed and which don't, it seems like the ones that show up are embeds of small to medium size (in terms of number of lines) rather than the larger ones? Maybe that has something to do with it? 🤔

mackuba:
I was thinking it were the ones where the quote text was literally one emoji, although I haven't spent much time analysing this (since like I mentioned it was 3 am ;)

surfdude29:
> I was thinking it were the ones where the quote text was literally one emoji, although I haven't spent much time analysing this (since like I mentioned it was 3 am ;)

That's a good working theory for 3am 😅

That might be part of the issue, but I think it might be wider than that. Because these two embeds don't show up, which aren't a single emoji:

[...]

PR #5329, opened 2024-09-13
Fix lexicon validation in PWI Discover

gaearon:
## Before

https://github.com/user-attachments/assets/917b8388-3eec-4e48-a8af-52959c308984

## After

https://github.com/user-attachments/assets/97cf6d7c-16f4-4a85-ba53-7ba2f52b03d7

## What?!

None of the media embeds were passing Lexicon validation in the logged out Discover because it did not go through the json-to-lex rehydration codepath, and so blobs were plain objects instead of `BlobRef` instances.

Screenshot 2024-09-13 at 19 50 58

The fix is to make it go through the same codepath as done in normal circumstances.

I verified we're not using this hack anywhere else so this should be enough.

### DIFF
--- a/src/lib/api/feed/custom.ts
+++ b/src/lib/api/feed/custom.ts
@@ -2,6 +2,7 @@ import {
   AppBskyFeedDefs,
   AppBskyFeedGetFeed as GetCustomFeed,
   BskyAgent,
+  jsonStringToLex,
 } from '@atproto/api'
 
 import {getContentLanguages} from '#/state/preferences/languages'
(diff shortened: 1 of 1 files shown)


mozzius (approved):
LGTM, nice find! Can't believe we never realised...

surfdude29:
> LGTM, nice find! Can't believe we never realised...

@mackuba noticed and opened #4959 👀

mackuba:
😏

gaearon:
Welp, guess we need to read issues more closely.

PR #5331, opened 2024-09-13
[Video] Remember mute state while scrolling

haileyok:


mozzius (approved):
good stuff, confirmed disable autoplay is working too

PR #5332, opened 2024-09-13
[Video] Don't require email verification on self-host

haileyok:


src/view/com/composer/videos/SelectVideoBtn.tsx
```diff
@@ -38,7 +40,14 @@ export function SelectVideoBtn({onSelectVideo, disabled, setError}: Props) {
       return
     }
 
-    if (!currentAccount?.emailConfirmed) {
+    console.log('currentAccount', currentAccount?.service)
```
mozzius: log

mozzius (approved):

src/view/com/composer/videos/SelectVideoBtn.tsx
haileyok: 👍

PR #5234, opened 2024-09-09
Add `context` to `Mute` and `Unmute` labels on video control

surfdude29:
Currently, the same `Mute` string is used in the context of the tag menu and as a label on a control for muting the sound of a video:

https://github.com/bluesky-social/social-app/blob/6c6a76b193edfd8bd46139b85fefd684ee557a8c/src/locale/locales/fr/messages.po#L3701-L3704

Given the different contexts, this isn't ideal for localization.

This PR suggests tweaking the labels to `Unmute video` and `Mute video`.

PR #5340, opened 2024-09-14
Add context to `Unmute` and `Mute` strings

surfdude29:
Similar to #5234, this PR adds `context` to the `Unmute` and `Mute` strings that refer to video, which enables better localization.

I think I got the syntax right 🤞

But I don't know how the code should be formatted to make Prettier happy?


haileyok (approved):
thanks!

surfdude29:
thanks for fixing it! :)

PR #5346, opened 2024-09-14
fix timeline appearance in browser on smartphone

tomo-x7:
If you view the timeline with the language setting set to Japanese in your smartphone browser, user posts with long handles and names look like this (probably all languages where the browser automatically does the wrapping without recognizing the space between words will be like this)
image
If the name and handle are short, it will look normal like this
image
If the language display is set to English, even if the name is in Japanese, it will be displayed correctly like this
image

### DIFF
--- a/src/view/com/util/PostMeta.tsx
+++ b/src/view/com/util/PostMeta.tsx
@@ -14,6 +14,7 @@ import {sanitizeHandle} from 'lib/strings/handles'
 import {niceDate} from 'lib/strings/time'
 import {TypographyVariant} from 'lib/ThemeContext'
 import {isAndroid} from 'platform/detection'
+import {web} from '#/alf'
 import {ProfileHoverCard} from '#/components/ProfileHoverCard'
 import {TextLinkOnWebOnly} from './Link'
 import {Text} from './text/Text'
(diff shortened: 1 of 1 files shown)


tomo-x7:
If you look at the timeline in Japanese, you will see posts from users with long handles and names like this (probably true for all languages where the browser automatically does the folding without recognizing the space between words)
image

gaearon:
Does this need the web-only condition?

tomo-x7:
> Does this need the web-only condition?

Sorry, I don't know.
However, I limited it to web only because native apps don't need to modify it and react native doesn't have whitespace property.

surfdude29:
See also #5322 by @quiple which I think is trying to fix the same issue

tomo-x7:
Oh!
I had missed that.
Sorry to hear that but I hope his 5322 works out.
Should I close this PR?

surfdude29:
> Oh!
> I had missed that.
> Sorry to hear that but I hope his 5322 works out.
> Should I close this PR?

I don't know what the etiquette is for situations like this, and to be clear I wasn't meaning any criticism, I just thought it would be helpful to mention the other PR as they both seem to be tackling the same issue.

I would suggest leaving this open and let the Bluesky team decide what to do :)

tomo-x7:
> I don't know what the etiquette is for situations like this, and to be clear I wasn't meaning any criticism, I just thought it would be helpful to mention the other PR as they both seem to be tackling the same issue.
Sorry for the misunderstanding. I don't feel criticized at all and I appreciate you telling me.

> I would suggest leaving this open and let the Bluesky team decide what to do :)
Okay, I will do that :)

Issue #1528, opened 2023-04-09
Prevent ability to enter an email address as a handle


arunsathiya:
At this time it’s possible to enter an email address in the “change my handle” flow but it shouldn’t be possible because, if I’m not mistaken, “@“ is not a valid character in the subdomain field.

devinivy:
@pfrazee doing some issue triage in the atproto, moved this one over here since it's more relevant to the app 👍

pfrazee:
Cool, assigning to @ansh

ssg:
@pfrazee & @arunsathiya  This doesn't repro anymore. When I type an email address into the handle field it invalidates the input field, and the Save button becomes disabled. 

![Screenshot](https://github.com/user-attachments/assets/fadceef1-817b-47ce-b4dd-85c3be58b793)

PR #5351, opened 2024-09-15
Add a Make Valid Handle Function for Custom Domain Handles

RetroSunstar:
Prevents the user from putting invalid domain characters in the Change Handle field for Custom Domains, mentioned in #1528 

Hopefully this is okay? 🩵🩷🩵🩷

### DIFF
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -13,6 +13,11 @@ export function makeValidHandle(str: string): string {
   return str.replace(/^[^a-z0-9]+/g, '').replace(/[^a-z0-9-]/g, '')
 }
 
+export function makeValidCustomDomainHandle(str: string): string {
+  str = str.toLowerCase()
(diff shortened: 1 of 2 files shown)


mary-ext:
this seems like the wrong approach? why not check if the handle is actually valid?
`/[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*(?:\.[a-zA-Z]{2,})/`

RetroSunstar:
> why not check if the handle is actually valid rather than stripping invalid characters?

Simply because that's how Bluesky handles it for .bsky.social handles, with the makeValidHandle function just above in handles.ts

I just put some characters that couldn't be used in a domain to begin with, esspecially the @ symbol as that's the subject of the original issue

Anything else I assume would be handled when the user goes to verify the DNS record or Text file? As the domain simply wouldn't exist


Alternatively, there's allowing the user to input these characters but show an error as soon as one is, which might affect typing less and allow the user to simply go back and change it? 🩵🩷🩵🩷

mozzius:
Superceded by #8658

PR #5352, opened 2024-09-15
[Video] Add volume slider to video embeds on web

### LINKED ISSUE #5285: Volume control for media player

Currently it is possible to mute and unmute videos embedded in posts, but it is not possible to adjust the playback volume.

Add a pop-out volume fader as is frequently found on other video player interfaces on the web.

Writing a GreaseMonkey script to add a volume control, but obviously this doesn't work on the native apps.

netux:
Adds an `<input type="range">` (better designs are welcome) next to the Toggle Mute button to adjust the volume of the video.

- The slider defaults to 50% volume.
  This is more a personal preference, but I think not maxing out the volume and blasting people's ears when focus on a video is probably a good idea. 50% is a comfortable default for me.
- Moving the slider draws focus and unmutes the video. This is inline with other places, like Twitter.
- The slider is logarithmic, meaning it adjusts better to human perception of sound.
  Basically, with a logarithmic scale, moving the slider from one point to another should feel like the slider value change more accurately resembles the change in volume perceived, even to audiophiles.
  I chose to use the conversion function from this blog post: <https://www.dr-lex.be/info-stuff/volumecontrols.html>.
- Volume is applied to all video embeds. This repurposes the video volume state from 843f9925f5d0773db321e617c1bd0be6a308ef7f.
- Even has a nice sliding animation, borrowing from YouTube's design.

GIF showcasing the volume slider animation

GIF of user interacting with the new slider and toggle mute button

Resolves #5285

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
@@ -751,6 +751,13 @@ function formatTime(time: number) {
   return `${minutes}:${seconds}`
 }
 
+function sliderVolumeToVideoVolume(value: number) {
+  return Math.pow(value, 4)
+}
+function videoVolumeToSliderVolume(value: number) {
+  return Math.pow(value, 1 / 4)
+}
+
 const INITIAL_VOLUME = 0.5
 
 function useVideoUtils(ref: React.RefObject<HTMLVideoElement>) {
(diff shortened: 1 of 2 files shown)


IsaacMarovitz:
I think the bar should be a bit longer and perhaps slightly thinner here to avoid feeling finicky. Also, I can see in your examples that the end of the control is slightly clipped.

> The slider defaults to 50% volume.

I think everyone would expect it to default to 100% not 50%

mozzius:
Thank you so much for this! I ended up redoing from scratch in #5363, but this PR was really helpful for doing the logarithmic volume controls. Much appreciated!

PR #5355, opened 2024-09-15
Typography tweaks

estrattonbailey:


src/App.web.tsx
```diff
@@ -96,50 +96,48 @@ function InnerApp() {
...
-        <ThemeProvider theme={theme}>
-          <RootSiblingParent>
-            <VideoVolumeProvider>
-              <ActiveVideoProvider>
-                <React.Fragment
-                  // Resets the entire tree below when it changes:
-                  key={currentAccount?.did}>
+      <StatsigProvider
+        // Resets the entire tree below when it changes:
+        key={currentAccount?.did}>
```
estrattonbailey, line 101: @gaearon special note here: I wanted to be able to use a feature flag in Alf, so I moved this up. I don't think there's any reason we can't have this `key` invalidation further up, but maybe you know of an edge case.

src/components/Button.tsx
```diff
@@ -635,14 +634,7 @@ export function ButtonText({children, style, ...rest}: ButtonTextProps) {
...
   return (
-    <Text
-      {...rest}
-      style={normalizeTextStyles([
-        a.font_bold,
-        a.text_center,
-        textStyles,
-        style,
-      ])}>
+    <Text {...rest} style={[a.font_bold, a.text_center, textStyles, style]}>
```
estrattonbailey, line 637: This was written before the new Typography components had settled. No apparent change in functionality.

PR #5360, opened 2024-09-16
[Video] Fix scrubber tap target

mozzius:
I put the padding on the wrong element - whoops!

Also shifts it down a little

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
@@ -259,6 +259,7 @@ export function Controls({
       if (evt.pointerType !== 'mouse' && !hovered) {
         evt.preventDefault()
       }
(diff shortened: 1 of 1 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
```diff
@@ -259,6 +259,7 @@ export function Controls({
       if (evt.pointerType !== 'mouse' && !hovered) {
         evt.preventDefault()
       }
+      clearTimeout(timeoutRef.current)
```
haileyok, line 262: this is supposed to be here right?

haileyok (approved):

src/view/com/util/post-embeds/VideoEmbedInner/VideoWebControls.tsx
mozzius: yep

PR #5362, opened 2024-09-16
add video playlist url to opengraph headers

Gargaj:
I'm not actually sure if M3U8 works as an embed, but it probably should since it's usually browser-to-browser?


Gargaj:
I guess video support was on a branch? 🤔

IsaacMarovitz:
Tried the same thing here #5320 more or less assuming that `m3u8` isn't supported

PR #5364, opened 2024-09-16
[Video] Fix safari showing spinner

mozzius:
Safari seems to emit `stalled` events when changing between HLS segments. a simple fix is to clear buffering state if the video is still playing

Also, unrelated: added a small animation to the scrubber


haileyok (approved):

mackuba:
Ooh nice

PR #5367, opened 2024-09-16
[Video] Revert safari hackfix

mozzius:
Don't merge until Jaz has fixed it server-side


haileyok (approved):

haileyok:
@mozzius I think we want to revert this now correct?

mozzius:
I need to check if it's been fixed server-side or not

haileyok:
Confirmed w/ Jaz that the fix got deployed. Will merge and ensure things are good on Safari.

PR #5375, opened 2024-09-17
Update messages.po

luan-u:
Change pt_BR text to be more natural

### DIFF
--- a/src/locale/locales/pt-BR/messages.po
+++ b/src/locale/locales/pt-BR/messages.po
@@ -5599,7 +5599,7 @@ msgstr "Resposta Oculta pelo Autor da Thread"
 #: src/components/moderation/ModerationDetailsDialog.tsx:114
 #: src/lib/moderation/useModerationCauseDescription.ts:122
 msgid "Reply Hidden by You"
-msgstr "Responder Oculto por Você"
+msgstr "Resposta oculta por você"
 
 #: src/components/dialogs/PostInteractionSettingsDialog.tsx:355
 msgid "Reply settings"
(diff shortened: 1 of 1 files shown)


src/locale/locales/pt-BR/messages.po
```diff
@@ -5599,7 +5599,7 @@ msgstr "Resposta Oculta pelo Autor da Thread"
 #: src/components/moderation/ModerationDetailsDialog.tsx:114
 #: src/lib/moderation/useModerationCauseDescription.ts:122
 msgid "Reply Hidden by You"
-msgstr "Responder Oculto por Você"
+msgstr "Resposta oculta por você"
```
gleydson, line 5602:
```suggestion
msgstr "Resposta ocultada por você"
```

This phrase is the better because it uses the past participle "ocultada," which suggests that the action of hiding has already been performed and completed

gleydson (commented):
LGTM

gleydson (commented):
LGTM

auroursa:
There are several translation improvement PRs for pt-BR. I suggest integrating your hard work into #5436.

estrattonbailey:
Is `Title Case` less common in Brazil, and do you prefer `Sentence case`? Similar edits in #5415, just want to confirm, since in other languages we seem to use `Title Case` most often.

luan-u:
Personally I prefer `Sentence case`, and also would say it's more common too. It feels more natural. `Title Case` is more used for, well, titles.

pfrazee (approved):
I don't feel strongly about this, so I'm +1

PR #5377, opened 2024-09-17
Pre-fill alt text with 10-million card post

clavin:
Adds a brief default alt text to the virtual card image generated for the 10-million user celebration post.

I’m not certain if this aligns with good accessibility practices, but I had the thought and found some others discussing it. I’m open to feedback, particularly from screen reader and other assistive technology users.

P.S. Congrats on 10 million users! 🥳

### DIFF
--- a/src/state/models/media/gallery.ts
+++ b/src/state/models/media/gallery.ts
@@ -1,19 +1,22 @@
 import {makeAutoObservable, runInAction} from 'mobx'
-import {ImageModel} from './image'
-import {Image as RNImage} from 'react-native-image-crop-picker'
-import {openPicker} from 'lib/media/picker'
+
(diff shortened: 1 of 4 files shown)


src/state/models/media/gallery.ts
```diff
@@ -100,10 +103,10 @@ export class GalleryModel {
   async addFromUris(uris: InitialImageUri[]) {
     for (const uriObj of uris) {
       this.add({
-        mime: 'image/jpeg',
```
clavin, line 103: This field was being silently ignored by the constructor to `ImageModel`, so I removed it here when I changed the constructor's options type

estrattonbailey (approved):
Thank you! Gonna merge into a base and workshop copy. Appreciate your help!

PR #5384, opened 2024-09-17
Language fixes

estrattonbailey:
The app currently only supports 2-letter BCP-47 language codes for most of its language handling. The exception is what we pass to Lingui for i18n, which _can_ include a region code e.g. `pt-BR`.

Using `pt-BR` locale as an example, we had two main issues in our handling:
1. When opening the app for the first time, `pt-BR` was normalized to `pt`, which matched none of our `AppLanguage` values, and so the app appeared translated in English.
2. Then, when adjusting the `appLanguage` preference to properly translate the app, we also updated the user's `primaryLanguage` and `contentLanguage` prefs, which resulted in region codes being inserted into those settings where they were not supported.

(1) is "fixed" by (2), but because of (2), many users who have fiddled with their UI translations may have bad data saved to their local storage.

So the main fix here is to sanitize that data on read and write, which covers both next app startup and next writes. Atm, good data will not be written until the next time the user updates these settings. I also de-coupled writing `appLanguage` from writing `primaryLanguage` and `contentLanguage` prefs to reduce confusion for users.

The other fix was to abstract and clean up initial locale handling so that on first load of the app for new users we try to translate the app using the full language tag first e.g. `pt-BR`, and if not found, fall back to `pt`. If that also isn't found, we fall back to `en` translations.

**This PR does not propose improvements for supporting region codes or ISO 639-3 3-letter codes.** It simply aims to fix the current handling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "await-lock": "^2.2.2",
     "babel-plugin-transform-remove-console": "^6.9.4",
     "base64-js": "^1.5.1",
+    "bcp-47": "^2.1.0",
     "bcp-47-match": "^2.0.3",
     "date-fns": "^2.30.0",
     "deprecated-react-native-prop-types": "^5.0.0",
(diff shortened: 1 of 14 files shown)


gaearon (approved):
i think this lgtm but let’s add uniqueness deduplication whenever we’re storing an array of those — i’ve seen duplicate entries in statsig data

PR #5386, opened 2024-09-17
Make it work with Xcode 16

mozzius:
- Updated `@sentry/react-native` to latest
- Updated `react-native-image-crop-picker` to latest


haileyok (approved):
Thanks for hopping on this so quickly.

Verified it still run on Xcode 15 so nothing should break with Actions.

PR #5395, opened 2024-09-17
[Neue] Base

estrattonbailey:
- [ ] Only show to users who joined prior to <date>


haileyok:
Nit: I think moving the description for `Font` to under the options would look a bit cleaner

Screenshot 2024-09-18 at 4 30 21 PM

src/App.native.tsx
```diff
@@ -106,62 +106,60 @@ function InnerApp() {
...
 function App() {
   const [isReady, setReady] = useState(false)
+  const [loaded] = useFonts()
 
   React.useEffect(() => {
     initPersistedState().then(() => setReady(true))
   }, [])
 
-  if (!isReady) {
+  if (!isReady || !loaded) {
```
haileyok, line 162: Question. Does `loaded` just default to true whenever we use the build plugin?

src/alf/tokens.ts
```diff
@@ -1,3 +1,7 @@
+import {Platform} from 'react-native'
+
+export const TRACKING = Platform.OS === 'android' ? 0.1 : 0
```
haileyok, line 3: Curious about why the difference?

src/components/dialogs/nuxs/index.tsx
```diff
@@ -83,7 +126,13 @@ function Inner() {
       if (nux && nux.completed) continue
 
       // then check gate (track exposure)
-      if (enabled && !enabled({gate})) continue
+      if (
+        enabled &&
+        !enabled({gate, currentAccount, currentProfile, preferences})
+      )
+        continue
```
haileyok, line 133: nit: can we wrap `continue` in braces?

src/components/dialogs/nuxs/index.tsx
```diff
@@ -38,12 +62,31 @@ export function useNuxDialogContext() {
...
-  const onboardingState = useOnboardingState()
-  return hasSession && !onboardingState.isActive ? <Inner /> : null
+  const {currentAccount} = useSession()
+  const {data: preferences} = usePreferencesQuery()
+  const {data: profile} = useProfileQuery({did: currentAccount?.did})
+  const onboardingActive = useOnboardingState().isActive
+
+  const isLoading =
+    !currentAccount || !preferences || !profile || onboardingActive
+  return isLoading ? null : (
```
haileyok, line 72: nit: can we invert this?

src/components/dialogs/nuxs/NeueTypography.tsx
```diff
@@ -0,0 +1,119 @@
...
+import {Text} from '#/components/Typography'
+
+export function NeueTypography() {
+  const t = useTheme()
+  const {_} = useLingui()
+  const nuxDialogs = useNuxDialogContext()
+  const control = Dialog.useDialogControl()
+  const {fonts} = useAlf()
+
+  Dialog.useAutoOpen(control, 3e3)
```
haileyok, line 22: Curiosity. What is the purpose of the `setTimeout`? I do recall having to do this for intents, but it was a lot shorter than three seconds if I remember correctly. Mainly:

- Why did we pick three seconds?
- Is there cases where this wouldn't be long enough?
- Is it possible the user could start scrolling the app before this gets presented to them, and the dialog would sort of get in the way?

src/App.native.tsx
haileyok: Nevermind thought `useFonts()` was imported from expo, i see what we're doing now 👍

src/components/dialogs/nuxs/NeueTypography.tsx
estrattonbailey: Opening right away is fine, I just liked the delay. Naturally any delay could cause it to pop up after scrolling. How do you feel about it?

haileyok: Lets see how it feels on device.

haileyok (approved):

haileyok:
Lets give this a spin on device. Things look good to me so far.

DavidBuchanan314:
I haven't actually tested this PR in any capacity, but my understanding is that the Inter font does "Contextual Alternates" by default. Here's their [demo](https://rsms.me/inter/):

image

Personally, I really don't like these. I find the modifications unexpected, potentially altering the meaning of text.

Anecdotally, I use the Element matrix client, which also uses Inter with Contextual Alternates - *every* time I've noticed it kicking in, it's been "incorrect" (e.g. inserting a multiplication symbol into something that was never a multiplication in the first place)

This CSS should fix it: `font-variant-ligatures: no-contextual`

src/components/dialogs/nuxs/NeueTypography.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+      <Dialog.ScrollableInner label={_(msg`Introducing new font settings`)}>
+        <View style={[a.gap_xl]}>
+          <View style={[a.gap_md]}>
+            <Text style={[a.text_3xl, {fontWeight: '900'}]}>
+              <Trans>Introducing new font settings ✨</Trans>
+            </Text>
+            <Text style={[a.text_lg, a.leading_snug]}>
+              <Trans>
+                To the ensure the best possible experience, we're introducing a
```
mackuba, line 56: @estrattonbailey "To the ensure" 😅

estrattonbailey: Wow nice one lol

mackuba: I opened main.bsky.dev and this just jumped at me 😉

PR #5396, opened 2024-09-17
Fix notification spacing

wscld:
This PR removes unnecessary spaces around the word "and" in the FeedItem component's notification items. Previously, there was an extra space before and after "and" due to the inclusion of {' '}. These have been removed to ensure proper spacing is handled.

Before:
Screenshot 2024-09-17 at 18 11 01

After
Screenshot 2024-09-17 at 18 11 07


estrattonbailey (approved):
Nice one! Thank you ✨

PR #5406, opened 2024-09-18
[Neue] Buttons

estrattonbailey:
Reduces button size options to `large` `small` and `tiny`
- merges `medium` into `large`, and makes `large` significantly smaller, only 2px taller than previous `medium`
- merges `xsmall` into `small`, `xsmall` grows by 4px in height and 8px in width, `small` shrinks by 2px height and 8px width
- tiny basically unchanged

In that process, I moved away from using spacing tokens for button padding, border radius, and gap, allowing us to better scale each button size. `small` especially had too much border radius for my taste.

I also adjusted the `ButtonIcon` component to align itself the same for each button size, no matter what icon size you pass to it. Previously a larger icon than the default would expand the button vertically.

Last commit is just migrating the size values. Very little overall change anywhere in the app.

New -> Old
CleanShot 2024-09-19 at 14 46 38@2x
CleanShot 2024-09-19 at 14 47 37@2x

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -30,7 +30,7 @@ export type ButtonColor =
   | 'gradient_sunset'
   | 'gradient_nordic'
   | 'gradient_bonfire'
-export type ButtonSize = 'tiny' | 'xsmall' | 'small' | 'medium' | 'large'
+export type ButtonSize = 'tiny' | 'small' | 'large'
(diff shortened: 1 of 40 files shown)


mozzius (approved):
Looks good! Not super noticeable change in the app but side-by-side they're a big improvement

estrattonbailey:
Let's give this a go and we can tweak

PR #5407, opened 2024-09-18
Fix for undefined ref on hot reload on web

estrattonbailey:


haileyok (approved):

gaearon:
hmmm we should figure out what’s actually going on here, this isnt supposed to happen

PR #5409, opened 2024-09-18
Update sentry sourcemaps upload

haileyok:
## Why

This changed, a lot. Simplest way to adjust this is to:

- Patch `@sentry/react-native` to allow passing additional args in the `expo-upload-sourcemaps` script.
- Create our own posthook that runs after builds, and includes the needed config

## Test Plan

Merge, let it build, check Sentry for sourcemaps after build


estrattonbailey (approved):

haileyok:
Verified this is working 👍

PR #5411, opened 2024-09-19
[Neue] Parallel font loading on web

estrattonbailey:
Expo statically extract font files specified in `useFonts` calls at build time, but font actually begin loading when `useFonts` is called. This means the entire JS bundle (4.2MB) has to load first, before we can load fonts.

This PR takes advantage of the static exporting, but manually loads the fonts using the DOM font API. This way, fonts are loaded parallel to the JS bundle. When the bundle boots up, we resolve `document.fonts.ready`, which resolves immediately if fonts are already loaded, or waits for them to load before we show the app.

The URLs of these fonts are hard-coded, but that's OK because they're content hashed. The hashes won't change unless the contents of the font files themselves change.

As before, on native, none of this is necessary and `expo-font` does all this for us.

Peep the waterfall in the deployment: https://social-app-pr-5411.onrender.com/


bskyweb/templates/base.html
```diff
@@ -315,6 +315,133 @@
   <meta property="og:site_name" content="Bluesky Social" />
   <link type="application/activity+json" href="" />
 
+  <script>
+    try {
+      const InterRegular = new FontFace(
```
gaearon, line 320: would it be sufficient to use `<link>` preload tags?

estrattonbailey: Discussed in Slack, I think we're gonna give preload a go in #5431

gaearon (approved):

Issue #5337, opened 2024-09-13
Tooltip says 'Follow' for labeler


kawarimidoll:
**Describe the bug**

When hovering over `@account.name` in the post, a tooltip is displayed.
I noticed that the tooltip says ‘Follow’ even if the account is a labeler. I think it should say ‘Subscribe’ instead.
Please see the screenshot.

**To Reproduce**

1. check this post https://bsky.app/profile/emilyliu.me/post/3l42z7nype52o and hover on `@newsdetective.bsky.social` (this is a labeler).

**Expected behavior**

**Screenshots**

<img width="527" alt="image" src="https://github.com/user-attachments/assets/d153f70d-b13a-41aa-8a92-774876c1c26d">

**Details**

 - Platform: web
 - Platform version: Build version: 1.91.0; Bundle info: 061683d (prod); Bundle date: 24091300; Platform: web
 - App version: N/A

**Additional context**

kawarimidoll:
fixed by #5414

PR #5414, opened 2024-09-19
Fixing hover card for labeler accounts

wscld:
Hey there! 👋 This PR fixes a bug in the ProfileHoverCard component where labeler status was not properly handled.
I have noticed this bug yesterday and it was also reported here #5337 

**Before:**
Screenshot 2024-09-18 at 23 37 29

**After:**
Screenshot 2024-09-18 at 23 36 58

**When hovering a user profile:**
Screenshot 2024-09-18 at 23 37 10

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -411,7 +411,7 @@ function Inner({
     () => currentAccount?.did === profile.did,
     [currentAccount, profile],
   )
-  const isLabeler = React.useMemo(() => profile.associated?.labeler, [profile])
+  const isLabeler = profile.associated?.labeler
 
   return (
     <View>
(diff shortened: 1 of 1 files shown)


src/components/ProfileHoverCard/index.web.tsx
```diff
@@ -411,6 +411,7 @@ function Inner({
     () => currentAccount?.did === profile.did,
     [currentAccount, profile],
   )
+  const isLabeler = React.useMemo(() => profile.associated?.labeler, [profile])
```
mary-ext, line 414: Memo is unnecessary

estrattonbailey (approved):
Nice one! Thank you 👍

src/components/ProfileHoverCard/index.web.tsx
estrattonbailey: Thanks, Mary ✨

PR #5415, opened 2024-09-19
Update to pt-br translation

geueds:


auroursa:
There are several translation improvement PRs for pt-BR. I suggest integrating your hard work into #5436.

estrattonbailey:
Is there a significant difference between `Chat` and `Conversa`? Why did we have `Chat` prior to this?

Also, we tend to use "Title Case" for buttons and titles. Is it more correct in `pt-BR` to use "sentence case" like `Seguir de volta` instead of ``Seguir De Volta`?

geueds:
> Is there a significant difference between `Chat` and `Conversa`? Why did we have `Chat` prior to this?

`Chat` is used in brazilian portuguese as a reference to old chat softwares, people do not use "Chat" as the word for what we're talking about here. "Chat" is not even used anymore on the web version for Bluesky app, only mobile, and that shall be fixed.

> Also, we tend to use "Title Case" for buttons and titles. Is it more correct in `pt-BR` to use "sentence case" like `Seguir de volta` instead of ``Seguir De Volta`?

At first, "Title Case" doesn't ever affect the words "de", "da", "do", "dos", "das", maybe `Seguir de Volta` could be the right way according to the title case option, but also we do not use that casing in `pt-BR` for buttons.

estrattonbailey:
@geueds thanks for the explanation! We can definitely include this if you want to re-open. Just wanted to make sure I understood :)

PR #5426, opened 2024-09-19
Validate TLD in signup

haileyok:
## Why

One part of preventing bad emails from being used during signup, like `whoever@haileyok.con`.

## Test Plan

There are different patterns that should get caught:

```
gnail.co/m
gmaill.co/m
gmai.co/m
gmail.co
gmal.co/m
iclod.co/m
icloud.co
outllok.co/m
outlok.co/m
outlook.co
yaoo.co/m
yaho.co/m
yahoo.co
yahooo.co/m
```

Also, any invalid TLD should get caught.

There is a list of test cases that includes the latest gTLDs for validation that they pass.

Also check that if you modify the email to another incorrect state it should continue to tell you that you entered the email incorrectly. However, if you modify it and it becomes correct, you should be able to proceed.

Lazy loading works here as well.

https://github.com/user-attachments/assets/318b0937-f626-435d-b9df-44f4880a5667

### DIFF
--- a/src/screens/Signup/StepInfo/index.tsx
+++ b/src/screens/Signup/StepInfo/index.tsx
@@ -7,7 +7,7 @@ import type tldts from 'tldts'
 
 import {logEvent} from '#/lib/statsig/statsig'
 import {logger} from '#/logger'
-import {emailTypoCheck} from 'lib/strings/email-typo'
+import {isEmailMaybeInvalid} from 'lib/strings/email'
 import {ScreenTransition} from '#/screens/Login/ScreenTransition'
 import {is13, is18, useSignupContext} from '#/screens/Signup/state'
 import {Policies} from '#/screens/Signup/StepInfo/Policies'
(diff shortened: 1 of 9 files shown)


src/screens/Signup/StepInfo/index.tsx
```diff
@@ -46,13 +48,43 @@ export function StepInfo({
...
+  const prevEmailValueRef = useRef<string>(state.email)
   const passwordValueRef = useRef<string>(state.password)
 
-  const onNextPress = React.useCallback(async () => {
+  const [hasWarnedEmail, setHasWarnedEmail] = React.useState<boolean>(false)
+
+  const tldtsRef = React.useRef<typeof tldts>()
+  React.useEffect(() => {
+    // @ts-expect-error - valid path
+    import('../../../../node_modules/tldts/dist/index.cjs.min.js').then(
```
gaearon, line 59: does `import('tldts/dist/index.cjs.min.js')` not work?

haileyok: will try again, but it did not. read in a SO that you actually have to supply the full path

haileyok: oh, okay without `node_modules` infront it does work. nice, ty

estrattonbailey (approved):
LGTM, works great!

PR #5427, opened 2024-09-19
[Neue] Ligatures & composer

estrattonbailey:
### Web before/after
Very subtle differences
CleanShot 2024-09-19 at 13 20 06@2x
CleanShot 2024-09-19 at 13 26 41@2x

### iOS before/after
Text is subtly smaller, tracking was previously `0.2` and with this update it's set to `0` to match our other text, that's the biggest difference.
IMG_3899
IMG_3898

### Android before/after
Same as iOS
Screenshot_20240919-141736
Screenshot_20240919-141635

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -22,7 +22,9 @@ import {
   LinkFacetMatch,
   suggestLinkCardUri,
 } from 'view/com/composer/text-input/text-input-util'
+import {atoms as a, useAlf} from '#/alf'
 import {Portal} from '#/components/Portal'
+import {normalizeTextStyles} from '#/components/Typography'
 import {Text} from '../../util/text/Text'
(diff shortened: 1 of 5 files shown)


src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -247,13 +250,31 @@ export const TextInput = React.forwardRef(function TextInputImpl(
...
+      },
+    )
+    /*
+     * TipTap component isn't a RN View and while it seems to convert
+     * `fontSize` to `px`, it doesn't convert `lineHeight`.
+     *
+     * `lineHeight` should always be defined here, this is defensive.
+     */
+    // @ts-ignore
+    style.lineHeight = style.lineHeight ? (style.lineHeight += 'px') : undefined
```
gaearon, line 269: this can just be `+` right?

gaearon (approved):

PR #5429, opened 2024-09-19
[Video] use dynamic import for hls.js

haileyok:


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
```diff
@@ -33,47 +34,57 @@ export function VideoEmbedInnerWeb({
   const hlsRef = useRef<Hls | undefined>(undefined)
 
   useEffect(() => {
-    if (!ref.current) return
-    if (!Hls.isSupported()) throw new HLSUnsupportedError()
+    let hls: Hls | undefined
+    // @ts-expect-error - dynamic import
+    import('hls.js/dist/hls.min').then(
```
gaearon, line 39: i actually think we want to kick it off early — say in the module scope — as long as it’s non blocking. since otherwise we’re delaying when the first video can be shown. 


consider playing with these: https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules

gaearon: basically it shouldn’t hold up the rest of the bundle but it also shouldn’t be deferred until render. you want to kick it off early on

haileyok: Just writing down so I don't forget:

Magic comment seems to not really be working. Better option might be to try and put this in a separate chunk which can get loaded in parallel maybe inside the webpack config. Docs are pretty mid unfortunately so need to play with this

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
```diff
@@ -33,47 +34,57 @@ export function VideoEmbedInnerWeb({
...
+      ({default: DynamicHls}: {default: typeof Hls}) => {
+        setHlsLoading(false)
```
gaearon, line 41: you probably want to exit here if cleanup has already ran (e.g. if `embed.playlist` changes faster than `Hls` loads

gaearon (changes requested):
- there's a race that needs to be fixed
- this is also a bit stale compared to main, and more refactoring would be needed on main before the same approach can be made to work here

gaearon (approved):

src/view/com/util/post-embeds/VideoEmbedInner/web-controls/VideoControls.tsx
```diff
@@ -413,7 +416,7 @@ export function Controls({
         <View
           pointerEvents="none"
           style={[a.absolute, a.inset_0, a.justify_center, a.align_center]}>
-          {buffering && <Loader fill={t.palette.white} size="lg" />}
+          {showSpinner && <Loader fill={t.palette.white} size="lg" />}
```
gaearon, line 419: should the above condition use `showSpinner` too instead of `buffering`? i have not tested visually closely

haileyok:
Tested on the video feed w/ deployment, seems to be working good now. Appreciate you taking a look @gaearon

PR #5430, opened 2024-09-19
[Video] Flush low quality segments once focused

### LINKED ISSUE #5315: First 9 seconds of video are always at a low streaming resolution.
**Describe the bug**
First 9 seconds of any video are always served at a lower image quality, I assume it's running a 320p or similar resolution. 

**To Reproduce**
1. Play any video, notice how after 9 seconds the image becomes better/crisper
https://bsky.app/profile/shvvhs.bsky.social/post/3l3yxnkds232n

**Expected behavior**
1. System should store if the user is able to stream videos at the best quality possibly, and then use these settings for subsequent videos. 
2. User should be able to select which resolution to watch the video. (suggestion)
3. Once the video loops, if any of the parts were offered at a lower resolution, they should be "updated" to the same resolution as all the later parts. (suggestion)

**Screenshots**
using download sites (like Cobalt), the video gets downloaded at the maximum quality available. Which confirms the issue is when the video is being streamed and not when it's being  encoded.
image

**Details**

 - Platform: Chrome, Firefox, With and Without VPN.
 - Platform version:
 - App version:

**Additional context**

mozzius:
Fixes #5315 

By default, once a fragment (~5 seconds of video) is loaded by HLS.js, it sticks around indefinitely. This means that the initial low-res fragments (which at minimum is the first 9 seconds of the video) stay low-res even after the video loops.

Solution:

1. Keep track of low-res fragments. We do this via the `FRAG_BUFFERED` event.
2. Once focused, we wait for the next fragment change, then:
3. If the auto-detected quality level at this time is greater than low-res, flush the low-res fragments
4. HLS.js will automatically reload the fragments on next loop, likely now in hi-res

This *shouldn't* result in unbounded loading of data - it only triggers the eviction if it has sufficient network conditions to load hi-res fragments, and once something is hi-res it should never be subsequently evicted.

This also fixes a bug which would result in the hi-res version loading regardless of whether the video was focused, so it might even reduce bandwidth use!

# Test plan

This is tricky to test. I would suggest adding logs to the flushing event trigger (VideoEmbedInnerWeb.tsx:109) and logging the `lowQualityFragments` state. Also keep an eye on the network tag, and just review the video to ensure that the first ~9 seconds are hi-res after looping

A good video to test with: https://bsky.app/profile/haileyok.com/post/3l4ioa5mudh25

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
@@ -51,6 +51,18 @@ export function VideoEmbedInnerWeb({
     // initial value, later on it's managed by Controls
     hls.autoLevelCapping = 0
 
+    // manually loop, so if we've flushed the first buffer it doesn't get confused
+    const abortController = new AbortController()
+    const {signal} = abortController
+    videoRef.current.addEventListener(
+      'ended',
+      function () {
+        this.currentTime = 0
+        this.play()
+      },
+      {signal},
+    )
+
     hls.on(Hls.Events.SUBTITLE_TRACKS_UPDATED, (_event, data) => {
       if (data.subtitleTracks.length > 0) {
         setHasSubtitleTrack(true)
@@ -82,6 +94,7 @@ export function VideoEmbedInnerWeb({
       hlsRef.current = undefined
       hls.detachMedia()
       hls.destroy()
+      abortController.abort()
     }
   }, [embed.playlist])
 
@@ -136,7 +149,6 @@ export function VideoEmbedInnerWeb({
             style={{width: '100%', height: '100%', objectFit: 'contain'}}
             playsInline
             preload="none"
-            loop
             muted={!focused}
             aria-labelledby={embed.alt ? figId : undefined}
           />
--- a/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
@@ -96,10 +96,8 @@ export function VideoEmbedInnerWeb({
         _event: Events.FRAG_CHANGED,
         {frag}: FragChangedData,
       ) {
-        if (!hlsRef.current) return
-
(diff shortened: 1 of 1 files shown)


src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
```diff
@@ -77,12 +85,54 @@ export function VideoEmbedInnerWeb({
...
+
+            flushed.push(lowQualFrag)
+          }
+
+          setLowQualityFragments(prev => prev.filter(f => !flushed.includes(f)))
+        }
+      }
+      hlsRef.current.on(Hls.Events.FRAG_CHANGED, fragChanged)
+
+      const current = hlsRef.current
```
gaearon, line 123: it would be best to capture this once at the top and then always use this reference instead of dereferencing during events. fewer things that can go wrong.

mozzius: Should I also use the captured version within the listener?

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
```diff
@@ -77,12 +85,55 @@ export function VideoEmbedInnerWeb({
...
+    if (!hlsRef.current) return
+
+    const current = hlsRef.current
+
+    if (focused) {
+      function fragChanged(
+        _event: Events.FRAG_CHANGED,
+        {frag}: FragChangedData,
+      ) {
+        if (!hlsRef.current) return
```
gaearon, line 99: i mean here, can you use the captured variable instead?

mozzius: can do, I've never been sure what the best way to go about it is

gaearon (approved):
react parts seem fine to me, have not tested video parts

haileyok:
This video is freezing whenever it replays. Audio continues but video does not https://social-app-pr-5430.onrender.com/profile/phantasmajones.net/post/3l4tkpes56w22

PR #5439, opened 2024-09-20
Add border to menu on web

futurGH:
In line with the new media borders, improves contrast a bit

| Before | After |
:------:|:------:
image  |  image


estrattonbailey (approved):
Dig it

PR #5449, opened 2024-09-22
[Neue] Handle emoji within custom font

estrattonbailey:
RN doesn't handle emojis within custom fonts very well. [See this issue.](https://github.com/facebook/react-native/issues/18559) The fix is to render just the emoji strings as `System` font.

Peep the first commit for the meat and potatoes, and 2-n commits that adds the `emoji` attribute where its needed.

While adding emoji support, I ported over a change I need in another PR as well, adding `dataSet` and `title` support to `Text`. We supported these on old `Text` component, and should here too.

Guards are in place to ensure that if `emoji` is specified, only `string`s are passed as children. We could technically expand this to include other elements, but they may unexpectedly obscure emoji in wrapping components that then won't get matched and styled as needed. This would likely be confusing. So at cost of boilerplate, this forces us to be specific about where we need to render emoji.

### What happens if we miss an `emoji` attribute somewhere?
Text may not align properly, that's it. So we should aim to cover our bases, but it's not catastrophic if we miss one on accident.

### Testing

Here's a good post to test: https://bsky.app/profile/test.esb.lol/post/3l4r55k332d2z

For other testing, click around and check out:
- post text
- "reposted by"
- "in reply to"
- labels & related dialogs
- feed cards
- profile cards
- list cards
- starter pack cards
- URL link cards
- profiles

### Missing
- [x] invalid handle


src/view/com/util/text/Text.tsx
```diff
@@ -35,6 +48,20 @@ export function Text({
...
 
+  if (IS_DEV) {
+    if (!emoji && childHasEmoji(children)) {
+      logger.warn(
+        `Text: emoji detected but emoji not enabled: "${children}"\n\nPlease add <Text emoji />'`,
+      )
+    }
+
+    if (emoji && !childIsString(children)) {
+      throw new Error(
```
gaearon, line 59: same nit

gaearon:
getting this on Search

Screenshot 2024-09-23 at 22 14 25

does it apply in author feed? this feels misaligned to me:

Screenshot 2024-09-23 at 22 15 52

getting this when clicking into a post from author feed:

Screenshot 2024-09-23 at 22 16 19

estrattonbailey:
Fixed a few more spots, can't repro the line-height misalignment 🤔

gaearon (approved):
lgtm but let's use logger.error to avoid prod/dev differences

Issue #1751, opened 2023-10-25
Design improvement: soften text and icon contrast in dark mode


scottdotjs:
At the moment, dark mode on web and in the app appears to be pure white on pure black and it's pretty jarring on the eyes. Could you please soften the contrast by darkening the text and icons? For example, have a look at Twitter in "lights out" mode.

Issue #647, opened 2023-05-16
Use an off-black background for dark mode and off-white for light mode


hs4man21:
**Describe the bug**
The current background colors on the web app are harsh, and this is actually an accessibility issue, as pure black on white text is known to [cause eye strain](https://uxmovement.com/content/why-you-should-never-use-pure-black-for-text-or-backgrounds/) whereas white on black can appear [blurred (this is called halation) especially to users with astigmatism](https://medium.com/@h_locke/why-dark-mode-causes-more-accessibility-issues-than-it-solves-d2f8359bb46a).

GitHub's color scheme might be a good model for how off-black and off-white backgrounds can be used for increased accessibility and general user comfort. Slack provides another example and is even more on the gray side of things than GitHub.

**To Reproduce**

**Expected behavior**

Pure black and white backgrounds should be avoided across the board. Slack, GitHub, Discord, VS Code and other apps use alternative colors for backgrounds to improve user experience.

If keeping pure black and white backgrounds is preferred, then another option is graying the text a bit to reduce the harshness.

**Screenshots**

#### Bluesky dark mode
<img width="600" alt="Screenshot 2023-05-16 at 12 30 25 AM" src="https://github.com/bluesky-social/social-app/assets/38192823/a55eec38-ab99-4da3-908c-aaec39b9f480">

#### GitHub dark mode (off-black)
<img width="237" alt="Screenshot 2023-05-16 at 12 31 13 AM" src="https://github.com/bluesky-social/social-app/assets/38192823/1b5b2cb6-990d-40bb-8037-7f36cb75e676">

#### Slack dark mode (further off-black/gray)
<img width="385" alt="Screenshot 2023-05-16 at 12 35 47 AM" src="https://github.com/bluesky-social/social-app/assets/38192823/e9f107fe-eff0-4fa9-8ad0-33dbf62ee03f">

**Details**

- Platform: Desktop Chrome MacOS
- Platform version: MacOS: Ventura 13.1, Chrome: 112.0.5615.137
- App version: 1.28(?), not sure where to find web app version number

PR #5453, opened 2024-09-23
Adjust dark mode text color to softer white

rimar1337:
related to #647 and #1751
helps users with sensitive eyes
changes foreground color in dark mode to a softer white (no pure white)
no changes to foreground color in light mode (light mode already uses soft black for foreground)
background colors are not changed

| Before | After |
|--------|--------|
| before-dim | after-dim |
| before |after |


estrattonbailey (approved):
Nice one, thanks!

PR #5456, opened 2024-09-23
Fix web splash

estrattonbailey:


mozzius (approved):
Looks good, I think? I see the splash, I don't know if there are any other considerations

PR #5457, opened 2024-09-23
Allow profile header to overscroll

mozzius:
This extracts the simple parts out of #2171 

Adds `allowHeaderOverScroll` prop to `PagerWithHeader` so that the profile header can scroll with the rest of the page

iOS only. does nothing on other platforms.

https://github.com/user-attachments/assets/44001625-268e-484a-bfd7-3a21d18f0543

# Test plan

Make sure android behaves the same, and make sure other list screens on iOS are unchanged (i.e. the PTR spinner is still visible)

### DIFF
--- a/src/screens/Profile/Sections/Feed.tsx
+++ b/src/screens/Profile/Sections/Feed.tsx
@@ -15,6 +15,7 @@ import {EmptyState} from '#/view/com/util/EmptyState'
 import {ListRef} from '#/view/com/util/List'
 import {LoadLatestBtn} from '#/view/com/util/load-latest/LoadLatestBtn'
 import {Text} from '#/view/com/util/text/Text'
+import {ios} from '#/alf'
 import {SectionRef} from './types'
 
 interface FeedSectionProps {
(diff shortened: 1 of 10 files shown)


src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -228,14 +233,20 @@ let PagerTabBar = ({
...
+    return {
+      transform: [
+        {
+          translateY: allowHeaderOverScroll
+            ? translateY
+            : Math.min(translateY, 0),
+        },
+      ],
+    }
+  })
```
mozzius, line 249: This is where the action happens, the rest is just prop drilling

src/screens/Profile/Sections/Feed.tsx
```diff
@@ -82,6 +83,7 @@ export const ProfileFeedSection = React.forwardRef<
         onScrolledDownChange={setIsScrolledDown}
         renderEmptyState={renderPostsEmpty}
         headerOffset={headerHeight}
+        progressViewOffset={ios(0)}
```
mozzius, line 86: this makes sure the spinner is at the top of the page, where it should be

mozzius:
I can easily feature flag it?

gaearon (approved):
lgtm

PR #5459, opened 2024-09-23
Add language filtering UI to search

estrattonbailey:
Adds language filtering to the search interface. Works like this:
- if `lang:*` param exists, it uses that, otherwise defaults to the user's first `contentLanguages[]` string
- when searching, selecting a different language from the dropdown overrides whats typed in the search box (if anything)

Also replaces basically the whole header section on the search screen. Functionality remains the same, except that when pressing "Cancel" it clears the search query as well. This felt like a bug to me with the existing handling.

Had to touch `TextField` in order to adjust its sizing to match new button sizing and keep alignment looking good. You can see the alignment on the search screen on mobile, or just check out the storybook.

## Web
CleanShot 2024-09-24 at 13 17 04@2x
CleanShot 2024-09-24 at 13 17 13@2x
CleanShot 2024-09-24 at 13 17 19@2x

## iOS
IMG_3911
IMG_3912
IMG_3913

## Android
Screenshot_20240924-131110
Screenshot_20240924-131125
Screenshot_20240924-131007


surfdude29:
I just tried it on the preview branch, very nice 👌

One minor point to mention (that is not in any way caused by this PR, of course): having languages in the dropdown does highlight the issue that some are repeated, e.g. French and German, which seems to be the issue referred to in #4314

IMG_4178

estrattonbailey:
@surfdude29 yeah we have some work we need to do here sometime. It's complicated by the fact that users have these values stored. So if we remove any, we need to remap old values to new values, among other things. Need to give it some thought, but it's on my list. Thanks for the bump!

src/view/screens/Search/Search.tsx
```diff
@@ -349,7 +491,7 @@ let SearchScreenInner = ({query}: {query?: string}): React.ReactNode => {
         title: _(msg`Top`),
         component: (
           <SearchScreenPostResults
-            query={query}
+            query={queryWithParams}
```
estrattonbailey, line 494: Post queries can have params like `lang`, but profiles and feeds don't support these and will return nothing. So those are passed the raw query without params.

src/view/screens/Storybook/Forms.tsx
```diff
@@ -32,7 +32,7 @@ export function Forms() {
           label="Text field"
         />
 
-        <View style={[a.flex_row, a.gap_sm]}>
+        <View style={[a.flex_row, a.align_start, a.gap_sm]}>
```
estrattonbailey, line 35: Prevents stretching, just for testing cross-platform

src/view/screens/Search/Search.tsx
```diff
@@ -747,91 +968,70 @@ let SearchInputBox = ({
...
+          onFocus={onFocus}
+          keyboardAppearance={t.scheme}
+          selectTextOnFocus={isNative}
+          autoFocus={false}
+          accessibilityRole="search"
+          autoCorrect={false}
+          autoComplete="off"
+          autoCapitalize="none"
+        />
+      </TextField.Root>
```
haileyok, line 1008: Very small question here that can be addressed separately, do we want to extract this out to our `SearchBox` (I think that's what it is called...) so we can use it in the right nav on web? Maybe with a configurable size since this might be too big for the sidebar. Just a thought.

src/view/screens/Search/Search.tsx
```diff
@@ -324,7 +333,138 @@ let SearchScreenFeedsResults = ({
...
+    paddingRight: 32,
+    paddingVertical: 8,
+    borderRadius: a.rounded_full.borderRadius,
+    borderWidth: a.border.borderWidth,
+    borderColor: t.atoms.border_contrast_low.borderColor,
+  }
+
+  return (
+    <RNPickerSelect
+      placeholder={{}}
```
haileyok, line 373: needed? just checking

haileyok (approved):
Feels really good, huge improvement.

Noted the hamburger question, worth considering I think going forward.

src/view/screens/Search/Search.tsx
estrattonbailey: Yeah that'd be cool. With the autocomplete and the operator hints we mentioned.

src/view/screens/Search/Search.tsx
estrattonbailey: Indeed it is not, just copy pasta from LanguageSettings

PR #5463, opened 2024-09-24
Remove `react-native-fs`

haileyok:
## Why

This has been lingering for a bit for almost no reason. We've pretty much completely moved to using `expo-file-system`, but there was a couple small usages left over.

## Test Plan

- Upload an image from both iOS and Android
- Verify the e2e `getFile()` still works (see video). It only needs to work on iOS

https://github.com/user-attachments/assets/ab7a4d93-ed59-4743-a0be-274c9616cde1

### DIFF
--- a/src/lib/api/upload-blob.ts
+++ b/src/lib/api/upload-blob.ts
@@ -1,4 +1,4 @@
-import RNFS from 'react-native-fs'
+import {copyAsync, deleteAsync} from 'expo-file-system'
 import {BskyAgent, ComAtprotoRepoUploadBlob} from '@atproto/api'
 
 /**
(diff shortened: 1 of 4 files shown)


src/lib/api/upload-blob.ts
```diff
@@ -74,7 +74,7 @@ async function withSafeFile<T>(
       return await fn(newPath)
     } finally {
       // Remove the temporary file
-      await RNFS.unlink(newPath)
+      await deleteAsync(newPath)
```
gaearon, line 77: does this need to be `safeDeleteAsync` instead?

gaearon (approved):
sounds good but see comment

src/lib/api/upload-blob.ts
haileyok: not in this case since we've already passed a good uri to `withSafeFile` (from above in `uploadBlob` where we prepend `file://` if needed)

ill use it though here just to promote the use of `safeDelete`

PR #5464, opened 2024-09-24
Remove image resizer

haileyok:
## Why

This library has caused problems with Fabric and isn't super well maintained. Can get rid of it in favor of Expo anyway.

## Test Plan

- Verify that compression does indeed work and that an appropriate max size is selected

Screenshot 2024-09-23 at 11 44 58 PM

- Verify that uploads still go through on both iOS and Android. The places we use compression are:
  - List avatar
  - User avatar
  - User banner


__tests__/lib/images.test.ts
```diff
@@ -100,47 +104,47 @@ describe('downloadAndResize', () => {
     }
 
     const result = await downloadAndResize(opts)
-    expect(result).toEqual(mockResizedImage)
```
haileyok, line 103: I'm removing this test, since it's not really possible to test this with jest. Notice that we were not even returning what the test says it should be returning, but instead just returning a valid `Image` object.

src/lib/media/picker.e2e.tsx
```diff
@@ -1,25 +1,39 @@
...
+    .split('/')
+    .slice(0, -6)
+    .concat(['Media', 'DCIM', '100APPLE'])
+    .join('/')
+
+  let files = await readDirectoryAsync(imagesDir)
+  files = files.filter(file => file.endsWith('.JPG'))
+  const file = `${imagesDir}/${files[0]}`
+
+  console.log(file)
```
gaearon, line 25: rm log?

src/lib/media/manip.ts
```diff
@@ -165,29 +167,46 @@ interface DoResizeOpts {
...
+  // Now instead, we have to supply the final dimensions to the manipulation function instead.
+  // Performing an "empty" manipulation lets us get the dimensions of the original image. React Native's Image.getSize()
+  // does not work for local files...
+  const imageRes = await manipulateAsync(localUri, [], {})
+  const newDimensions = getResizedDimensions({
+    width: imageRes.width,
+    height: imageRes.height,
+  })
+
+  for (let i = 0; i <= 9; i++) {
```
gaearon, line 181: why <= change?

src/lib/api/upload-blob.ts
```diff
@@ -74,7 +74,7 @@ async function withSafeFile<T>(
       return await fn(newPath)
     } finally {
       // Remove the temporary file
-      await RNFS.unlink(newPath)
+      await deleteAsync(newPath)
```
gaearon, line 77: does this need `safeDeleteAsync` for the normalization fix?

gaearon (approved):
seems ok but see comments

src/lib/media/manip.ts
haileyok: thanks

PR #5466, opened 2024-09-24
[Share Extension] Support on Android for sharing videos to app

haileyok:
## Why

Just adding support for "Share to bluesky" for videos on Android

## Test Plan

Try sharing a video into the app from photos app

https://github.com/user-attachments/assets/e48ff5c0-a532-4ae7-b9f5-40ad0c0180d7


mozzius (approved):
Looks sane, haven't tested

PR #5467, opened 2024-09-24
Automatically optimise SVG assets

mozzius:
Rather than copy/pasting into svgomg, this PR adds `svgo` as a dev dependency. I've tried to align the config with svgomg's output as much as possible, alas, there's still a chunky diff likely due to different svgo versions used. I'd advice ignoring commit 843c69dc5f2e8adbc5b370fd75a2d9e599bda1c0 when reviewing

run `yarn icons:optimize` to optimise all icons, or just dump svgs into the icons folder and the pre-commit hook will format it for you when you commit.


gaearon (approved):
sure

PR #5468, opened 2024-09-24
[Video] Refactor HLS logic

gaearon:
Gets rid of the second effect.

## Test Plan

Same as https://github.com/bluesky-social/social-app/pull/5430.

I have not actually tested myself properly.


Zero3K:
https://social-app-pr-5468.onrender.com/profile/phantasmajones.net/post/3l4tkpes56w22 works for me (it loops properly and the quality is good).

mozzius (approved):
turns out I made an elementary error (`return` instead of `continue` in the loop) but can confirm it works now

PR #5473, opened 2024-09-24
Add gate to increase post-feed page size

estrattonbailey:
Feeds that don't do language filtering themselves return all results to the client, where we do our best to filter by the user's `contentLanguages`. If we get no good results back, we "bail out" and return the raw results, which may all be in a different language.

This PR adds a flag so we can test upping this fetch limit to see if that improves how many posts we get back in the user's language.

### DIFF
--- a/src/state/queries/post-feed.ts
+++ b/src/state/queries/post-feed.ts
@@ -15,24 +15,25 @@ import {
   useInfiniteQuery,
 } from '@tanstack/react-query'
 
+import {AuthorFeedAPI} from '#/lib/api/feed/author'
+import {CustomFeedAPI} from '#/lib/api/feed/custom'
+import {FollowingFeedAPI} from '#/lib/api/feed/following'
 import {HomeFeedAPI} from '#/lib/api/feed/home'
+import {LikesFeedAPI} from '#/lib/api/feed/likes'
+import {ListFeedAPI} from '#/lib/api/feed/list'
+import {MergeFeedAPI} from '#/lib/api/feed/merge'
+import {FeedAPI, ReasonFeedSource} from '#/lib/api/feed/types'
 import {aggregateUserInterests} from '#/lib/api/feed/utils'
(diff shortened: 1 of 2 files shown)


gaearon (changes requested):
there's more logic that uses `PAGE_SIZE` — notably, the logic that ensures we have enough posts fetched in cases some of them get filtered out. i think you probably want to adjust that logic too.

i would suggest just removing the `PAGE_SIZE` constant altogether so that we're not tempted to use it accidentally.

estrattonbailey:
@gaearon right right, I should have explained this better or named things better. 

My thinking is is that we have two numbers here: the min number of posts we want to fill a screen (rn it's 30), and the fetch limit. That fetch limit is what I think we need to bump here, because that will ensure that we fill at least 30 posts for the view. If we increase the min posts we want to 100, we'll be doing likely about the same number of requests, and each request is now heavier than before.

That sound reasonable? My latest commit adds some clarification, good call on not wanting to confuse the const.

gaearon (approved):
ah ok thanks for explaining, makes sense

src/state/queries/post-feed.ts
```diff
@@ -135,6 +142,15 @@ export function usePostFeedQuery(
...
   const isDiscover = feedDesc.includes(DISCOVER_FEED_URI)
 
+  /**
+   * The number of posts to fetch in a single request. Because we filter
+   * unwanted content, we may over-fetch here to try and fill pages by
+   * `MIN_POSTS`.
+   */
+  const fetchLimit = React.useState(() => {
+    return gate('post_feed_lang_window') ? 100 : MIN_POSTS
+  })[0]
```
gaearon, line 152: this can just be 

```js
const fetchLimit = gate('post_feed_lang_window') ? 100 : MIN_POSTS
```

because gates are cached per-user in memory

estrattonbailey: Ah right, thanks!

PR #5474, opened 2024-09-24
Header blurred banner on overscroll (take 2)

mozzius:
# Stacked on #5457 

Supercedes #2171 

iOS only. Grows and blurs the banner on overscroll, with a spinner overlaid.

https://github.com/user-attachments/assets/1d8b7282-555c-4d85-b126-e0459b74c6a3

### DIFF
--- a/src/screens/Profile/Header/GrowableBanner.tsx
+++ b/src/screens/Profile/Header/GrowableBanner.tsx
@@ -59,7 +59,7 @@ function GrowableBannerInner({
   children: React.ReactNode
 }) {
   const isFetching = useIsProfileFetching()
-  const delayedIsFetching = useDelayedValue(isFetching, 500)
+  const delayedIsFetching = useStickyToggle(isFetching, 500)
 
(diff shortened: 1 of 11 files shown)


src/screens/Profile/Header/GrowableBanner.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }, [value, prevValue, delay])
+
+  useEffect(() => {
+    return () => {
+      clearTimeout(timeoutRef.current)
+    }
+  }, [])
+
+  return isDelayed ? true : prevValue
+}
```
mozzius, line 149: I really don't like this, is there a more elegant way to achieve this? I want the spinner to keep spinning for half a second while it animates back up

haileyok: might have a decent way of dealing with this, mentioned something in slack

gaearon (approved):
lgtm

PR #5476, opened 2024-09-24
Fix composer jumpiness

estrattonbailey:


haileyok (approved):
code lgtm, didn't test in sim yet but makes sense (using rn text over text seems like the right call here)

PR #5479, opened 2024-09-25
Add Cantonese Localization

auroursa:
When I see some of the Bluesky users complaining that they cannot understand either zh-CN or zh-TW and have to use English, I think it's necessary to complete the localization for Cantonese.

Cantonese (粵文) is the commonly used language in Hong Kong, Macau, and the local Chinese communities in Malaysia and Singapore.It's a unique language, with the majority of users being in Hong Kong, so this language file primarily follows Hong Kong's usage habits. 

For compatibility reasons, I categorized it as "zh-HK" instead of "yue" to match more browsers and OS. At the same time, I simplified zh-CN and zh-TW to 'Simplified Chinese' and 'Traditional Chinese, as their current names are too long.

I completed 100% of the translation, and welcome any reviews as well.


src/locale/languages.ts
```diff
@@ -47,8 +48,9 @@ export const APP_LANGUAGES: AppLanguageConfig[] = [
   {code2: AppLanguage.ru, name: 'Русский – Russian'},
   {code2: AppLanguage.tr, name: 'Türkçe – Turkish'},
   {code2: AppLanguage.uk, name: 'Українська – Ukrainian'},
-  {code2: AppLanguage.zh_CN, name: '简体中文（中国）– Chinese (Simplified)'},
-  {code2: AppLanguage.zh_TW, name: '繁體中文（臺灣）– Chinese (Traditional)'},
+  {code2: AppLanguage.zh_CN, name: '简体中文 – Simplified Chinese',},
```
surfdude29, line 51:
```suggestion
  {code2: AppLanguage.zh_CN, name: '简体中文 – Simplified Chinese'},
```
There's an extra `,` here which I think may be causing a lint error :)

auroursa: Thanks!

auroursa:
Thanks to Bluesky user @quicpasta.bsky.social for the review (commit 388f7df).

estrattonbailey (approved):
Amazing work! Thank you very much!

PR #5480, opened 2024-09-25
Avatar overscroll animation

mozzius:
# Stacked on #5457 and #5474 

Subtle grow animation for the avatar when overscrolling, to make it feel a bit more dynamic. Matches how large headers work on iOS (i.e. in Settings)

https://github.com/user-attachments/assets/79bc9419-3011-4c94-828a-ae04436c3ca3


gaearon (approved):
yay

PR #5481, opened 2024-09-25
Show a toast when a haptic is meant to fire while using simulator

mozzius:
Dev only. Should help test haptics while developing


haileyok (approved):
oh, crafty and nifty.

PR #5483, opened 2024-09-25
Use scale animation for tabs

mozzius:
Uses the new `PressableScale` for tabs

https://github.com/user-attachments/assets/fc2348b5-2bfb-47db-a938-98dc1bd7589d

### DIFF
--- a/src/lib/custom-animations/PressableScale.tsx
+++ b/src/lib/custom-animations/PressableScale.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {Pressable, PressableProps} from 'react-native'
+import {Pressable, PressableProps, StyleProp, ViewStyle} from 'react-native'
 import Animated, {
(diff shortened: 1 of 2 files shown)


src/lib/custom-animations/PressableScale.tsx
```diff
@@ -32,22 +35,24 @@ export function PressableScale({
...
         }
         cancelAnimation(scale)
         scale.value = withTiming(targetScale, {duration: 100})
       }}
       onPressOut={e => {
         'worklet'
-        if (rest.onPressOut) {
-          runOnJS(rest.onPressOut)(e)
+        if (onPressOut) {
+          runOnJS(onPressOut)(e)
```
haileyok, line 47: oops thanks for fixing that

haileyok (approved):
Nice! Let's see how it feels on device.

PR #5484, opened 2024-09-25
Messages list - make avatars link to profile

mozzius:
Makes the avatar link to the user's profile and replaces `UserAvatar` with `PreviewableUserAvatar`

### DIFF
--- a/src/screens/Messages/List/ChatListItem.tsx
+++ b/src/screens/Messages/List/ChatListItem.tsx
@@ -24,8 +24,9 @@ import {useProfileShadow} from '#/state/cache/profile-shadow'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
 import {useSession} from '#/state/session'
 import {TimeElapsed} from '#/view/com/util/TimeElapsed'
-import {UserAvatar} from '#/view/com/util/UserAvatar'
+import {PreviewableUserAvatar} from '#/view/com/util/UserAvatar'
 import {atoms as a, useBreakpoints, useTheme, web} from '#/alf'
+import * as tokens from '#/alf/tokens'
 import {ConvoMenu} from '#/components/dms/ConvoMenu'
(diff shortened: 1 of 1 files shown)


src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -203,6 +204,21 @@ function ChatListItemReady({
...
       onBlur={onMouseLeave}
       style={[a.relative]}>
+      <Link
+        to={`/messages/${convo.id}`}
+        label={displayName}
+        style={[
+          a.z_10,
+          a.absolute,
+          {top: tokens.space.md, left: tokens.space.lg},
+        ]}>
```
haileyok, line 214: Think we can remove this (and looks like it's going to the wrong place anyway haha)

src/screens/Messages/List/ChatListItem.tsx
```diff
@@ -236,11 +252,8 @@ function ChatListItemReady({
               (hovered || pressed || focused) && t.atoms.bg_contrast_25,
               t.atoms.border_contrast_low,
             ]}>
-            <UserAvatar
-              avatar={profile.avatar}
-              size={52}
-              moderation={moderation.ui('avatar')}
-            />
+            {/* Avatar goes here */}
+            <View style={{width: 52, height: 52}} />
```
haileyok, line 256: Do we need this? Why was the `<Link>` above absolute instead of just having the `PreviewableUserAvatar` here?

src/screens/Messages/List/ChatListItem.tsx
mozzius: lmao whoops I clicked it and it went to the right place so I never gave it a second thought 😭

src/screens/Messages/List/ChatListItem.tsx
mozzius: You're not allowed to nest anchor tags or touchables, so we have to absolutely position it

haileyok (approved):

PR #5489, opened 2024-09-25
Add back empty placeholder

estrattonbailey:
Without this we get a useless placeholder
CleanShot 2024-09-25 at 17 28 14@2x

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -370,6 +370,7 @@ function SearchLanguageDropdown({
 
   return (
     <RNPickerSelect
+      placeholder={{}}
       value={value}
       onValueChange={onChange}
       items={items}


src/view/screens/Search/Search.tsx
```diff
@@ -370,6 +370,7 @@ function SearchLanguageDropdown({
 
   return (
     <RNPickerSelect
+      placeholder={{}}
```
mozzius, line 373: You can use this to specify what the placeholder says, i.e.

```
placeholder={{
  label: _(msg`Select language...`),
  value: '',
}}
```

haileyok (approved):

src/view/screens/Search/Search.tsx
estrattonbailey: Talked in Slack, we may want one for "No language", but need to do a little more work

PR #5499, opened 2024-09-26
Ignore bogus onScroll values

mozzius:
The scroll listener is reporting strange values when the profile first loads (equal to negative header height).

This is a little bit hacky, but it appears safe to ignore those values, since they're typically way out of the range you can possibly achieve by just overscrolling.

Video showing the bogus values:

https://github.com/user-attachments/assets/d252d937-ed64-4043-82ac-0f1af2be4b66

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -166,7 +166,9 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
         // in fact, I actually can't do it, so let's just ignore those.
         // Let's only apply this hack if we've overscrolled by more than 300px
         // so that it's not a false positive from overscrolling -sfn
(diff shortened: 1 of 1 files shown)


gaearon (approved):

src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -161,10 +161,19 @@ export const PagerWithHeader = React.forwardRef<PagerRef, PagerWithHeaderProps>(
       (e: NativeScrollEvent) => {
         'worklet'
         const nextScrollY = e.contentOffset.y
-        scrollY.value = nextScrollY
-        runOnJS(queueThrottledOnScroll)()
+        // HACK: onScroll is reporting some strange values on load (negative header height).
+        // Highly improbable that you'd be overscrolled by over 400px -
```
gaearon, line 165: comment outdated

PR #5378, opened 2024-09-17
bring back kawaii mode

ovnanova:
i'm a kawaii girl in a formerly kawaii world and now i'm in shambles

PR #5503, opened 2024-09-26
Remove 10milly dialog, revert header logo changes

estrattonbailey:
Closes #5378 😅


haileyok (approved):
no harm ig in keeping some of the code around just inaccessible if we want to do more stuff like this in the future. can always revist the old commit though for that so nbd either way.

PR #5505, opened 2024-09-26
[Share Extension] Use the proper identifier for if statement on iOS

haileyok:
Missed this, originally thought it was `public.video` but its actually `public.movie`. Already had updated in the `Info.plist` but not here.


modules/Share-with-Bluesky/Info.plist
```diff
@@ -40,4 +40,4 @@
     <key>CFBundleShortVersionString</key>
     <string>$(MARKETING_VERSION)</string>
   </dict>
-</plist>
+</plist>
```
haileyok, line 43: Weirdly swiftlint removes this new line, maybe that's the standard here.

estrattonbailey (approved):

PR #5506, opened 2024-09-26
Make 2 images with varying aspect ratio look more presentable

mary-ext:
I used CSS grid for my web client but I wasn't sure how I'd translate it to something that works for React Native, so I don't think this is clamping the aspect ratio properly as it should have. (we're clamping the aspect ratio to between 3/4 and 4/3 so we don't end up with extreme crops)

| Before | After |
| --- | --- |
| image | image |
| image | image |
| image | image |


mary-ext:
My web client for point of reference

| Description | Image |
| --- | --- |
| 1000x563 and 667x1000 | image |
| two 1000x562 | image |
| two 667x1000 | image |

mary-ext:
I won't lie, this is actually worse than just presenting >=2 images in a horizontal scroller.

PR #5509, opened 2024-09-26
[Share Extension] Move away from deprecated API, implement JS side of things

haileyok:


modules/Share-with-Bluesky/ShareViewController.swift
```diff
@@ -150,7 +150,8 @@ class ShareViewController: UIViewController {
     var responder: UIResponder? = self
     while responder != nil {
       if let application = responder as? UIApplication {
-          return application.perform(#selector(openURL(_:)), with: url) != nil
+        application.open(url)
+        return true
       }
       responder = responder?.next
```
haileyok, line 156: `openURL` is deprecated now, we'll just use `application.open()` instead.

src/state/queries/video/video.ts
```diff
@@ -237,21 +239,24 @@ export function useUploadVideo({
...
-    onSelectVideo(asset)
-  }
+      dispatch({
+        type: 'SetAsset',
+        asset,
+      })
+      onSelectVideo(asset)
+    },
+    [_, onSelectVideo],
+  )
```
haileyok, line 259: No change here, just wrapping in `useCallback` for the effect below

mozzius (approved):
Not tested but read through it and seems ok

PR #5511, opened 2024-09-26
ALF text input for generic search input

mozzius:
Have not tested super thoroughly

Screenshot 2024-09-26 at 21 31 52


estrattonbailey (approved):
Nice one! Thanks for doing the legwork here. I just tweaked the new component and some spacing around its usages. Plus swapped into the search page now that the dust has settled there.

Issue #5492, opened 2024-09-26
Share Extension - some images show up as file URLs


jaronheard:
**Describe the bug**

Some images images show up as file URLs instead of images when shared through the share extension. Note, I have a fix shared in additional context.

**To Reproduce**

1. Open a photo in iMessage by tapping on it
2. Share to Bluesky from that view
3. Observe that the share content is file:///... and not an image

**Expected behavior**

The image shared is the image and not a file:/// link

**Screenshots**

<img src="https://github.com/user-attachments/assets/b53442f5-90b4-42b9-9797-d0b63a03d8d1" height="450">

**Details**

 - Platform: iOS
 - Platform version: 18.1
 - App version: 1.91.1.464 2feb5be (prod)

**Additional context**

I am using an adopted version of the bluesky share extension code for my app and fixed this by reordering the if statement that checks the content type to put the URL check last ([PR](https://github.com/jaronheard/soonlist-turbo/pull/219))

Tagging @haileyok since this is your work and saw in an in-flight PR [in-flight share extension PR](https://github.com/bluesky-social/social-app/pull/5466)... hope this is helpful since your code helped me!

surfdude29:
I can reproduce this on iOS 15.8.3 and iOS 17.7.

haileyok:
Lets try to fix...

haileyok:
Got it! Thanks for flagging that @jaronheard!

PR #5515, opened 2024-09-27
[Share Extension] Support images/movies from other apps like iMessage

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/5492

## Why

Some apps - when sharing an image or movie - actually end up being `public.url` rather than `public.image` or `public.movie`. As a result, we need to do a little bit of parsing for these.

## Test Plan

Hard to test this in simulator, since the easiest way to test is by sharing a movie or image from iMessage into Bluesky. However, you should be able to test that on a device. Should test:

- Sharing an image works
- Sharing a video works
- Sharing URLs (from something like Safari) still works as it did before

### DIFF
--- a/modules/Share-with-Bluesky/ShareViewController.swift
+++ b/modules/Share-with-Bluesky/ShareViewController.swift
@@ -180,12 +180,12 @@ class ShareViewController: UIViewController {
 extension URL {
   var type: URLType {
     get {
-      let urlStr = self.absoluteString
-      let ext = self.pathComponents.last?.split(separator: ".").last?.lowercased() ?? ""
-
-      if !urlStr.starts(with: "file://") {
+      guard self.absoluteString.starts(with: "file://"),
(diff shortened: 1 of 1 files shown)


modules/Share-with-Bluesky/ShareViewController.swift
```diff
@@ -43,9 +52,18 @@ class ShareViewController: UIViewController {
...
-      if let encoded = data.absoluteString.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed),
-         let url = URL(string: "\(self.appScheme)://intent/compose?text=\(encoded)") {
-        _ = self.openURL(url)
+      switch data.type {
+      case .image:
+        await handleImages(items: [item])
+        return
+      case .movie:
+        await handleVideos(items: [item])
+        return
```
haileyok, line 61: These will call `self.completeRequest()` on their own, so return after awaiting.

gaearon (approved):

PR #5517, opened 2024-09-27
Lazy load ViewShot

haileyok:
## Why

Now that we are able to, we should lazy-load `ViewShot`, since it saves us almost 80kb

## How

Just using `React.lazy` and `React.Suspense`. Extremely basic for the QR code dialog, we just use `React.Suspense` to show the spinner until `view-shot` has loaded.

For signup, we won't actually use `React.Suspense`, because we don't want the user to get blocked from signing up for any reason. However, we _will_ try to make sure it's ready to go by the time the user gets to that screen.

- On "continue" from the account info screen (very first step of registration), we'll call `import()` for `view-shot`
- By the time we get to the avatar creator, it should be loaded
- If it isn't loaded though, it will just result in an undefined avatar URL and the user will just get the default avatar.

## Test Plan

Go through an account creation, you should still get one of the "new" default avatars. Also verify that QR codes still work.

### DIFF
--- a/src/screens/Onboarding/StepProfile/PlaceholderCanvas.tsx
+++ b/src/screens/Onboarding/StepProfile/PlaceholderCanvas.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import {View} from 'react-native'
+import ViewShot from 'react-native-view-shot'
 
 import {useAvatar} from '#/screens/Onboarding/StepProfile/index'
 import {atoms as a} from '#/alf'
@@ -12,7 +13,7 @@ const LazyViewShot = React.lazy(
 const SIZE_MULTIPLIER = 5
 
 export interface PlaceholderCanvasRef {
-  capture: (() => Promise<string>) | undefined
+  capture: () => Promise<string | undefined>
 }
 
 // This component is supposed to be invisible to the user. We only need this for ViewShot to have something to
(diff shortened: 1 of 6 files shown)


src/screens/Onboarding/StepProfile/PlaceholderCanvas.tsx
```diff
@@ -33,12 +37,12 @@ export const PlaceholderCanvas = React.forwardRef<PlaceholderCanvasRef, {}>(
 
     React.useImperativeHandle(ref, () => ({
       // @ts-ignore this library doesn't have types
-      capture: viewshotRef.current.capture,
+      capture: viewshotRef.current?.capture,
```
gaearon, line 40: are we ok with this method being null in some cases? should we instead wrap to make it a noop?

haileyok: i tweaked it like this so it will show up as possibly being undefined [`3b220f6` (#5517)](https://github.com/bluesky-social/social-app/pull/5517/commits/3b220f6e82c108dfa451b65dd16243db5f7453fd)

im going to see if i can just get rid of that ts-ignore though, i think there actually is a type for this

haileyok: okay yea, this feels nicer [`1006d70` (#5517)](https://github.com/bluesky-social/social-app/pull/5517/commits/1006d70ed106fe59f1a026d1bf9180e14868722c)

gaearon (approved):

PR #5521, opened 2024-09-27
Rework native autocomplete

haileyok:
## Why

There was some clipping here that needed to get fixed. It also just kind of looked janky, so redid some design here.

- Add a border to the entire thing like there is on web
- Fix missing handles (`||` instead of `??`)
- Stop using `TouchableOpacity` and instead use `PressableScale`
- Change `PressableScale` to be a bit more in-line with stuff like `style` (remove `contentContainerStyle` prop)

## Test Plan

Everything should be the same as before functionality wise, with a couple of fixes.

https://github.com/user-attachments/assets/d51e7dcb-ce14-4e6a-a911-e86865a56bfa


mozzius (approved):
LGTM - I also stripped out the RN Animated animation and replaced it with Reanimated layout transitions, give it a counter-review before you merge

PR #5522, opened 2024-09-27
Reduce post reason icon size & offset

futurGH:
Makes the pin icon less aggressive, repost icon adjusted to match. Also edited the pin icon SVG because it visually appeared to have more space on the right side than on the left, causing right margin to be 1px off from the repost icon

| Before | After |
| ------ | ------ |
| image | image |


estrattonbailey (approved):
These icons are all exported on the same artboard size, 24x24. I verified that this one is correct, any adjustments should be made in CSS only.

Looks great though, gonna merge into a base to tweak this and get this included. Thanks!

futurGH:
Looks like the rounding when I optimized the SVG made the change barely perceptible anyways :P thanks!

PR #5534, opened 2024-09-30
Fix alt text - allow keyboard dismiss to be handled by scrollviews

mozzius:
Uses `keyboardDismissMode` et al props on the ScrollView for dismissing keyboard rather than the janky `onTouchMove={() => Keyboard.dismiss()}` on the modal containing element. Also made sure to fix it for the gif alt text modal too

To determine - why did we use this hack in the first place??


haileyok:
Don't need this anymore 🙌

PR #5535, opened 2024-09-30
Paralellize image uploads

gaearon:
I'm starting to bring in different pieces inspired by / taken from https://github.com/bluesky-social/social-app/pull/4163. The first of these pieces will be moving decoupling link meta / embed resolution from composer previews. I'm not ready to do that yet so I'm starting by doing little refactorings.

In this refactor, I'm doing a few things to the post creation method:

- Removed unsafe `as` coercions. They were only needed due to how the logic was structured with `embed` being redefined to wrap the previous `embed`.
- I found the recursive definition of `embed` a bit difficult to trace in general because it wasn't clear which cases are mutually exclusive. I've refactored it with stronger types to actually follow the codepaths that _can_ execute.
- While doing so, I've added parallelization for image uploads — similar to how we already do for video.

## Test Plan

Verify all of these still work.

- Posting a link (with preview)
- Posting one image
- Posting multiple images
- Posting a GIF
- Posting a video
- Posting a starter pack
- Posting a list
- Posting a feed
- Posting a quote
- Posting a quote with image(s)
- Posting a quote with video
- Posting a quote with GIF
- Posting a quote with link

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -201,7 +201,6 @@ async function resolveEmbed(agent: BskyAgent, opts: PostOpts) {
     embed = recordEmbed
   }
 
-  // add image embed if present
   if (opts.images?.length) {
     logger.debug(`Uploading images`, {
       count: opts.images.length,
@@ -239,10 +238,7 @@ async function resolveEmbed(agent: BskyAgent, opts: PostOpts) {
         images,
       }
     }
-  }
-
-  // add video embed if present
-  if (opts.video) {
+  } else if (opts.video) {
     const captions = await Promise.all(
       opts.video.captions
         .filter(caption => caption.lang !== '')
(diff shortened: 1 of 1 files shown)


src/lib/api/index.ts
```diff
@@ -313,3 +179,117 @@ export async function post(agent: BskyAgent, opts: PostOpts) {
...
+  | AppBskyEmbedImages.Main
+  | AppBskyEmbedVideo.Main
+  | AppBskyEmbedExternal.Main
+  | AppBskyEmbedRecord.Main
+  | AppBskyEmbedRecordWithMedia.Main
+  | undefined
+> {
+  if (opts.extLink?.embed) {
+    return opts.extLink.embed
+  }
```
estrattonbailey, line 196: This check (for no images) got lost here. So if you compose a post with an image(s), AND paste in a link for an embeddable entity like a Feed, the Feed is resolved and overwrites the images without any feedback to the user.

Existing functionality preserves the images, and the feed is left as a link facet.

estrattonbailey (commented):
So much better! Pinged in Slack for another issue, but if we wanna punt on that, we should just discuss my one comment here.

src/lib/api/index.ts
gaearon: makes sense, thanks for explaining. the original behavior only did this for photos, but it makes sense to do for videos too. so i pushed that.

estrattonbailey (approved):
A beaut!

PR #5540, opened 2024-09-30
Use Inter variable font

estrattonbailey:
A helpful user pointed out that we could take advantage of Inter's variable font and its optical sizing. This also cuts down requests from 6 fonts to 2 (though the files are 2x the size), and makes font handling quite a bit more simple. So, win win. We can continue to refine from here.


haileyok (approved):

haileyok:
We love users.

luciascarlet:
Thank you 🖤

Issue #5545, opened 2024-10-01
Starter Pack short links get a 404 error


mdruker:
### Steps to Reproduce

1. Add a share link to a starter pack on a profile, e.g. https://go.bsky.app/5XQRRF8
2. Click the link. It will go to https://bsky.app/starter-pack-short/5XQRRF8 instead
3. You get a 404 error or a "starter pack not found" error depending on whether you're logged in, though not even an actual 404 error page

### Attachments

![Screenshot 2024-09-30 at 10 58 57 PM](https://github.com/user-attachments/assets/f39e5e41-7e25-4f04-93c9-2b2f314464e8)
![Screenshot 2024-09-30 at 10 59 57 PM](https://github.com/user-attachments/assets/2fda3c19-78ea-495f-ab70-da78487e5e33)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Macbook Air, OS X 14.5, Chrome

### What version of the app are you using?

Build version: 1.91.1; Bundle info: 49ab486 (prod); Bundle date: 24092017; Platform: web

PR #5546, opened 2024-10-01
Handle /starter-pack-short/:code in Go

gaearon:
Maybe fixes https://github.com/bluesky-social/social-app/issues/5545. Have not verified.


gaearon:
Confirmed https://social-app-pr-5546.onrender.com/starter-pack-short/5XQRRF8 works but https://bsky.app/starter-pack-short/5XQRRF8 doesn't. So I think the fix is correct.

devinivy (approved):

PR #5547, opened 2024-10-01
Introduce a composer reducer and move image state there

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5535.

---

This continues taking apart / adapting https://github.com/bluesky-social/social-app/pull/4163.

Currently, all composer-related state lives in assorted state variables inside the `Composer` component. We're going to need to change this since we need to support composing multiple posts in a thread. So we're going to need to lift state up.

As a first step towards that future, I am adding a reducer for composer state. Reducers are much easier to lift up because we'll be able to index individual posts state by the ID/index of that post. For now, this reducer only manages the `images` state. But similarly to https://github.com/bluesky-social/social-app/pull/4163, I intend to move most composer state there, stacking individual behaviors on top of this PR.

This PR implements enough logic in the reducer that we can delete the `images` state variable completely. However, I haven't changed any surrounding logic to keep the diff small. So the `images` is still being re-derived directly in the component. I will be changing that later.

Note that I've also chosen to diverge from the data model in https://github.com/bluesky-social/social-app/pull/4163. In https://github.com/bluesky-social/social-app/pull/4163, the shape of the composer state tries to mimic the record shape. There are some benefits to that but I find it too clunky (and it's not truly the record shape anyway). Instead, I'm choosing a simpler data structure (an optional record + an optional media embed) which represents the same constraints ("can't have two record embeds") but is easier to update. (This PR does not use the `record` field yet.)

## Test Plan

- Verify adding, editing (alt text, crop), removing images works.
- Verify removing actually does remove the image (on post).
- Verify removing all images creates a post with no images.
- Verify the image limit is respected.
- Verify attaching videos and quotes works like before.
- Verify sharing image from intent works.

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -24,6 +24,7 @@ import {
   threadgateAllowUISettingToAllowRecordValue,
   writeThreadgateRecord,
 } from '#/state/queries/threadgate'
+import {ComposerState} from '#/view/com/composer/state'
 import {LinkMeta} from '../link-meta/link-meta'
 import {uploadBlob} from './upload-blob'
 
@@ -38,6 +39,7 @@ export interface ExternalEmbedDraft {
 }
 
 interface PostOpts {
+  composerState: ComposerState // TODO: Not used yet.
   rawText: string
   replyTo?: string
   quote?: {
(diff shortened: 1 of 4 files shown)


mary-ext:
Agreed on not fully mimicking the actual post records wrt embed and instead having a `media` and `record` field always available, it's rather awful.

estrattonbailey (approved):
Didn't test, code LGTM

PR #5548, opened 2024-10-01
Remove top padding from shell, move down into individual screens

mozzius:
Creates a new `<Layout.Screen>` which is responsible for a) setting the top inset and b) setting the correct height. Painful but it should clean up a lot of jank in lesser-touched screens

This lets us disable it selectively e.g. for the profile screen (PR to follow!)

The intent is to migrate more features, such as CenteredView and the various headers into these `<Layout.xyz>` components

However, since we have to remove it from the shell, we can't incrementally migrate it, so we also have to use it in every screen in the app in one giant PR. Sorry :(

# Test plan

Go to every single screen on the app 😅 and make sure it looks the same. Also needs to be done on web too. I'm so sorry

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -1,0 +1,25 @@
+import React from 'react'
+import {View} from 'react-native'
+import {useSafeAreaInsets} from 'react-native-safe-area-context'
+
+// Every screen should have a Layout component wrapping it.
+// This component provides a default padding for the top of the screen.
+// This allows certain screens to avoid
+//
+// In a future PR I will add a unified header component to this file and
+// things like a preconfigured scrollview.
+
+let Screen = ({
+  children,
+  disableTopPadding,
+}: {
+  children: React.ReactNode
+  disableTopPadding?: boolean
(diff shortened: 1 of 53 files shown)


src/components/Layout.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import {View, ViewStyle} from 'react-native'
+import {StyleProp} from 'react-native'
+import {useSafeAreaInsets} from 'react-native-safe-area-context'
+
+import {atoms as a} from '#/alf'
+
+// Every screen should have a Layout component wrapping it.
+// This component provides a default padding for the top of the screen.
+// This allows certain screens to avoid
```
surfdude29, line 10: I think there might be something missing in this sentence? 🤔

mozzius: lol yeah thanks

surfdude29: Also: this is one of those PRs where Hide whitespace **really** comes in handy 😅

estrattonbailey (approved):

PR #5550, opened 2024-10-01
Improve notification localization

estrattonbailey:
Many thanks to @quiple for the hard work on this!

Basically what this does is hoist up the construction of translated strings so that they can be constructed as a single unit. 

Logic has not changed aside from ensuring that follow-backs are only supported if ungrouped. In this branch, I added that guard and renamed a few variables to be a little more clear.


gaearon (approved):
have not tested but reading looks good

PR #5551, opened 2024-10-01
Tighten up video upload state management

gaearon:
This is related to the stream of work in https://github.com/bluesky-social/social-app/pull/5547. We want to aim towards making video state a part of the composer reducer. At least in some way. Syncing two different reducers is tricky so I'm hoping I can actually unify them in the future.

For now, I want to reduce the complexity of the video reducer a little bit. This is an assorted set of tweaks.

- c3789de6eb469aa7464d759b2c3d140fa5e83ba8: Removing dead code.
- a3aa41238b7807b85fa2ee24c5a6aed872039b55: Exposing the inner `dispatch` of the video reducer is unnecessary because we already expose an equivalent method called `clearVideo` which does `dispatch({ type: 'Reset' })` internally. Let's use that instead in the remaining callsite. This isn't strictly necessary now but I'm doing it so that in the future we can switch to composer's `dispatch`.
- [`bfd4874`?w=1](https://github.com/bluesky-social/social-app/pull/5551/commits/bfd4874a8197699d93b4c69b5fc767f0006bc466?w=1): We shouldn't write side effects in reducers. Now that there's only one place from which `dispatch({ type: 'Reset' })` can happen, we move this code into that method. The reducer no longer needs `queryClient`.
- ea254e132f78b43ada01d64a7b53ccf02ac34cd7: I want to move `jobId` into the reducer state later, so let's pull it out of the hook. I've added some logic to reset `setEnabled` when new `jobId` comes in which should be equivalent to the earlier method which did this by hand.
- 68e35bdfc5dbe53c7fbdd0ff51a134e208e2ab98: Notice this is the only `SetStatus` callsite. Let's make it more concrete.
- 50287fd40a9c89c78601cd131bd49d6aa91dfb10: Thanks to the above preparations, we can now move `jobId` itself into the video reducer state.
- d31c736eafb3ccae2a00f9fdf613ce8d849046d9: This makes query cancellation scoped. We'll need this for the thread composer in the future because it could have multiple videos (in multiple posts) uploading at the same time.

## Test Plan

Upload some videos, cancel upload at different points, there should be no changes in behavior. Verify you're able to remove an uploaded video and upload a different video instead. Turn off the internet during upload to trigger error cases.

This should basically work like before, but the code is tightened up a bit.

### DIFF
--- a/src/state/queries/video/video.ts
+++ b/src/state/queries/video/video.ts
@@ -3,7 +3,7 @@ import {ImagePickerAsset} from 'expo-image-picker'
 import {AppBskyVideoDefs, BlobRef} from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
-import {QueryClient, useQuery, useQueryClient} from '@tanstack/react-query'
+import {useQuery, useQueryClient} from '@tanstack/react-query'
 
 import {AbortError} from '#/lib/async/cancelable'
 import {SUPPORTED_MIME_TYPES, SupportedMimeTypes} from '#/lib/constants'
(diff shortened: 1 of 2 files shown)


gaearon:
https://github.com/bluesky-social/social-app/pull/5570

PR #5553, opened 2024-10-01
Include ghost hook to statically export fonts

estrattonbailey:
We use `expo export` when building for OTA updates. The although the `expo-font` plugin works for production builds, the `export` CLI command does not use the plugin config, and therefore does not output our font files for use in the OTA bundle.

The `useFonts` hook here ensures that the fonts are exported for native OTA, even though it is not otherwise used.

### DIFF
--- a/src/alf/fonts.ts
+++ b/src/alf/fonts.ts
@@ -1,3 +1,5 @@
+import {useFonts} from 'expo-font'
+
 import {isWeb} from '#/platform/detection'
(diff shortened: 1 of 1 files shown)


haileyok (approved):
Maybe should report it upstream too, curious as to why the build plugin isn't handling this.

estrattonbailey:
Gonna merge this for now to test, will update further if we hear back from Expo

PR #5558, opened 2024-10-01
Fix profile header buttons

estrattonbailey:
Noticed some misalignment here, and now that we're reigning in our button styles, this is one of the most important spots.

This makes a few small adjustments:
- removes custom styling from these buttons, except for the border radius
- replaces overflow menu button with `Button`
- removes the hacky `web({ top: -0.4 })` that I added, def not right, unneeded imo
- bumps up the small button size by 2px, was just def too small
- reduces size of small/round button by 4px, the "optical sizing" I was going for wasn't right here for small buttons, still looks good on `large` size though

CleanShot 2024-10-01 at 16 40 30@2x
CleanShot 2024-10-01 at 16 40 34@2x
CleanShot 2024-10-01 at 16 40 24@2x
CleanShot 2024-10-01 at 16 39 57@2x
CleanShot 2024-10-01 at 16 39 51@2x
CleanShot 2024-10-01 at 16 40 00@2x

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderStandard.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderStandard.tsx
@@ -153,8 +153,9 @@ let ProfileHeaderStandard = ({
           style={[
             {paddingLeft: 90},
             a.flex_row,
+            a.align_center,
             a.justify_end,
-            a.gap_sm,
+            a.gap_xs,
             a.pb_sm,
             a.flex_wrap,
           ]}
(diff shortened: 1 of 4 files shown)


src/screens/Profile/Header/ProfileHeaderStandard.tsx
```diff
@@ -150,11 +150,15 @@ let ProfileHeaderStandard = ({
         style={[a.px_lg, a.pt_md, a.pb_sm, a.overflow_hidden]}
         pointerEvents={isIOS ? 'auto' : 'box-none'}>
         <View
+          onLayout={e => {
+            console.log('height', e.nativeEvent.layout.height)
+          }}
```
mozzius, line 155: log

mozzius (approved):
Good stuff!

src/screens/Profile/Header/ProfileHeaderStandard.tsx
estrattonbailey: ope, hadn't saved the file

PR #5560, opened 2024-10-01
[Sheets] [Pt. 2] Remove a bunch of now-unnecessary things

haileyok:
- **remove**
- **close all dialogs**
- **rm handle component**
- **type**

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -21,7 +21,6 @@ import {
   DialogOuterProps,
 } from '#/components/Dialog/types'
 import {createInput} from '#/components/forms/TextField'
-import {FullWindowOverlay} from '#/components/FullWindowOverlay'
 import {Portal} from '#/components/Portal'
 
 export {useDialogContext, useDialogControl} from '#/components/Dialog/context'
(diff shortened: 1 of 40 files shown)


src/state/dialogs/index.tsx
```diff
@@ -42,11 +43,18 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+        const dialog = activeDialogs.current.get(id)
+        if (dialog) dialog.current.close()
+      })
+
+      return openDialogs.current.size > 0
+    } else {
+      // @TODO DIALOGS REFACTOR
+      console.error('HAILEY FIX THIS 🥺📋')
+      return false
+    }
```
haileyok, line 57: I still need to come back to this. I believe this will be unnecessary now (on all platforms?) since my understanding is it was used to fix the android back button (which is just handled natively now, no need for us to pay attention to it.)

src/view/com/composer/photos/ImageAltTextDialog.tsx
```diff
@@ -22,8 +22,6 @@ type Props = {
 export const ImageAltTextDialog = (props: Props): React.ReactNode => {
   return (
     <Dialog.Outer control={props.control}>
-      <Dialog.Handle />
```
haileyok, line 25: Removing all `Handle`s, because the native dialog includes a handle on both platforms.

src/components/Dialog/index.tsx
```diff
@@ -41,71 +29,20 @@ export * from '#/components/Dialog/utils'
...
-   * Used to manage open/closed, but index is otherwise handled internally by `BottomSheet`
-   */
-  const [openIndex, setOpenIndex] = React.useState(-1)
-
-  /*
-   * `openIndex` is the index of the snap point to open the bottom sheet to. If >0, the bottom sheet is open.
-   */
-  const isOpen = openIndex > -1
+  // @TODO DIALOG REFACTOR - can i get rid of this? seems pointless tbh
+  const [isOpen, setIsOpen] = React.useState(false)
```
haileyok, line 45: This ended up getting removed later since it isn't useful.

mozzius (approved):
Does not leave it in a working state (missing snapPoints), but that's fine

PR #5563, opened 2024-10-02
[Sheets] [Pt. 5] Update types and props

haileyok:
- **remove**
- **close all dialogs**
- **rm handle component**
- **type**
- **get scrollview working**
- **tidy**
- **bump**
- **bump gesture handler**
- **tweak types**
- **add props back**
- **update types**


mozzius (approved):
Looks pretty simple

Issue #5538, opened 2024-09-30
Round down follower count instead of up


vertigris:
**Is your feature request related to a problem? Please describe.**

Right now the app rounds up follower counts leading to situations where someone celebrates a milestone even though they have't reached it yet. As an example, my app says I have 15k followers but when I check on web it says 14.987. Which isn't a huge problem but it's kind of weird.

**Describe the solution you'd like**

Instead of rounding up, it should round down. So when I'm at 14.987, it should say 14.9k instead of 15k.

mary-ext:
This is supposed to have been fixed but for some reason the polyfill that's being used to get proper NumberFormat functionality on mobile is uh.... not working?

https://github.com/bluesky-social/social-app/commit/4d39ef2e19632eddc1a31a5b41078f6e40b92fbc

mary-ext:
This should be fixed with https://github.com/bluesky-social/social-app/pull/6384

PR #5568, opened 2024-10-02
Fix follower count truncation

mary-ext:
Fixes https://github.com/bluesky-social/social-app/issues/5538

On mobile, apply our own truncation since NumberFormat polyfill doesn't support roundingMode option.

This may look wrong on some locales like JP locale where numbers get displayed weirdly (e.g. 5083801 -> 5M / 508.3万), perhaps we should just apply it to web as well?

From brief testing, this truncation should work up to 999\_999\_999\_999\_997, I don't think anyone will reach this value.

### DIFF
--- a/src/view/com/util/numeric/format.ts
+++ b/src/view/com/util/numeric/format.ts
@@ -2,9 +2,7 @@ import type {I18n} from '@lingui/core'
 
 import {isWeb} from '#/platform/detection'
 
-// `@formatjs/intl-numberformat/polyfill` doesn't support `roundingMode`, and
-// we'd ideally want `roundingMode: trunc` to properly display shortened counts
-// without it being rounded up, which can be misleading for follower count.
+// `@formatjs/intl-numberformat/polyfill` doesn't support `roundingMode`
 const truncateRounding = (num: number): number => {
   if (num < 1_000) {
     return num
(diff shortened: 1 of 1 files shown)


aliceisjustplaying:
would this still leave a place to check somewhere the raw numbers on the UI?

mary-ext:
yes the `formatCount` function here is specifically for short numbers

gaearon:
what do we want to display for JP?

mary-ext:
4\_999\_999 should probably display 499万, right now it displays 490万, in `en` that'd be 4.9M

Actually, I wonder if other locales have their own scheme? Will see...

mary-ext:
I've checked and Chinese, Korean and Japanese seems to be the only ones problematic here, perhaps we should just special case those three?

aliceisjustplaying:
hindi has the whole lakh thing going for it i wonder if this should be 1.63 instead of just 1.6
IMG_1568

aliceisjustplaying:
this may be a separate thing tho

dolciss:
In Japanese (and probably also in Chinese and Korean) the name of the power of 10 changes every 10^4.
ex)9_999 -> 9999, 10_000 -> 1万, 12_345_000 -> 1234.5万, 100_000_000 -> 1億, 1_000_000_000_000 -> 1兆 ...

It is likely that where we currently calculate with 2, 5 or 8 digits for truncation,
we will need to calculate with 3, 7 or 11 in these languages.

estrattonbailey:
@mary-ext this seems like a bug in the FormatJS poly, right? I'm not seeing any open issues for this, and they state that it should be supported, so we may want to open an issue and see if we can get it fixed upstream.

mary-ext:
We should yeah, I think they're purposefully disabling tests for newer NumberFormat features and just never got round to it? I don't see anything wrt roundingMode on formatjs' repository other than the tests

PR #5570, opened 2024-10-02
Refactor video uploads

gaearon:
Supersedes https://github.com/bluesky-social/social-app/pull/5551.

I wanted to do a more minimal refactor at first but it ended up pretty fragile due to interdependency of various async hooks and how `useReducer` was interplaying with RQ. In this case, using RQ doesn't really give us anything and introduces a lot of complexity so I've flattened down the upload pipeline to a single `async` function.

I've also rewritten the video reducer in the last couple of commits.

This might be a bit tricky to review; the overall direction started changing from 0032646ea1ba06a8338913c37011fa1056222b45 and onward. You can read that commit by commit. Or try https://github.com/bluesky-social/social-app/pull/5570/files?w=1.

## Test Plan

Upload some videos. Verify success and error cases still work. Verify polling still works. Try turning off the wifi etc. Try uploading a video, canceling it, uploading another video. Verify post-while-uploading still works.

It would be good to check that specific error messages still work. I have not verified that.

Try different platforms.


mozzius (approved):
This is looking good. I can't find any regressions in my testing

PR #5573, opened 2024-10-02
Manage video reducer from composer reducer

gaearon:
In https://github.com/bluesky-social/social-app/pull/5547, I introduced a reducer for composer state management. Over time, we want to move the rest of meaningful composer state there so that we can later lift it up and make it work for multiple threaded posts. In https://github.com/bluesky-social/social-app/pull/5547 I started with managing the image state there. In this PR, I'm extending this approach to also handle videos.

Integrating videos into the reducer was a bit tricky because they were managed by a custom hook with a combination of React Query and a separate reducer. So there was no easy way to let that state be controlled. To fix this, I've refactored video state management in https://github.com/bluesky-social/social-app/pull/5570 to just be a plain reducer. Why? Because plain reducers compose.

In this PR, the way the video reducer is integrated has changed. Previously, it was given to `useReducer`. Now, the videos reducer is called/composed from the composer reducer. Note how it's not always "there": its state gets created when you add the video embed, and gets destroyed when you remove the video embed. (To do this, I've made the `idle` state represent the lack of a video, and excluded it from the state union.) It always starts its life in the `compressing` state now.

Note how the video actions are now "packed" inside the composer actions. That's composition again.

## Test Plan

Add/remove images. Add/remove video. Should work like before.

One consequence of better modeling is that now we can't add a video while we have images. (Previously the composer would let you do that but would silently omit images on post.) Since the action being (correctly) ignored is confusing, I've added greying out of the "add video" button when images are selected.

### DIFF
--- a/src/view/com/composer/state.ts
+++ b/src/view/com/composer/state.ts
@@ -17,24 +17,28 @@ type ImagesMedia = {
   labels: string[]
 }
 
+type VideoMedia = {
+  type: 'video'
+  video: VideoState
+}
+
 type ComposerEmbed = {
   // TODO: Other record types.
   record: PostRecord | undefined
   // TODO: Other media types.
-  media: ImagesMedia | undefined
+  media: ImagesMedia | VideoMedia | undefined
 }
 
 export type ComposerState = {
   // TODO: Other draft data.
   embed: ComposerEmbed
-  video: VideoState // TODO: Move into embed.
 }
 
 export type ComposerAction =
   | {type: 'embed_add_images'; images: ComposerImage[]}
   | {type: 'embed_update_image'; image: ComposerImage}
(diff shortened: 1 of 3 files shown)


src/view/com/composer/state.ts
```diff
@@ -104,6 +124,55 @@ export function composerReducer(
...
+      }
+    }
+    case 'embed_update_video': {
+      const videoAction = action.videoAction
+      const prevMedia = state.embed.media
+      let nextMedia = prevMedia
+      if (prevMedia?.type === 'video') {
+        nextMedia = {
+          ...prevMedia,
+          video: videoReducer(prevMedia.video, videoAction),
```
gaearon, line 151: composition :)

haileyok (approved):
Excellent stuff, this is coming together so nicely 🔥

PR #5575, opened 2024-10-02
[Sheets] [Pt. 9] Normalize RNGH pressable, use in remaining places

haileyok:
- **remove**
- **close all dialogs**
- **rm handle component**
- **type**
- **get scrollview working**
- **tidy**
- **bump**
- **bump gesture handler**
- **tweak types**
- **add props back**
- **update types**
- **update other dialog views**
- **replace all buttons with `BottomSheetButton`**
- **fix type**
- **wizard remove buttons**
- **fix**
- **fix embed consent**
- **buttons in account switcher**
- **custom bottom sheet link**
- **nesting support**
- **android support**
- **fix type**
- **simplify pressable changes**
- **rm**
- **rename for clairity**
- **fix other buttons in dialogs**
- **tweak**
- **more fixes**


src/components/Button.tsx
```diff
@@ -87,6 +87,7 @@ export type ButtonProps = Pick<
     style?: StyleProp<ViewStyle>
     hoverStyle?: StyleProp<ViewStyle>
     children: NonTextElements | ((context: ButtonContext) => NonTextElements)
+    PressableComponent?: React.ComponentType<PressableProps>
```
mozzius, line 90: Extremely neat, we can stick `PressableScale` in here now presumably?

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -303,7 +304,8 @@ export function PostInteractionSettingsForm({
               }
               value={quotesEnabled}
               onChange={onChangeQuotesEnabled}
-              style={[, a.justify_between, a.pt_xs]}>
+              style={[, a.justify_between, a.pt_xs]}
```
mozzius, line 307:
```suggestion
              style={[a.justify_between, a.pt_xs]}
```

src/view/com/util/post-ctrls/RepostButton.tsx
```diff
@@ -157,7 +157,6 @@ let RepostButton = ({
             </View>
             <BottomSheetButton
               label={_(msg`Cancel quote post`)}
-              onAccessibilityEscape={close}
```
mozzius, line 160: Is this necessary? (I never quite understood how this prop works)

mozzius (approved):
lgtm

src/view/com/util/post-ctrls/RepostButton.tsx
haileyok: It isn't supported on the RNGH pressable unfortunately (causes a crash). From what I read about it, I think the native dialog actually handles this for us (i think it lets the user use two finger presses to like dismiss or something...might look into the prop more later)

src/components/Button.tsx
haileyok: Yea that actually should work!

PR #5576, opened 2024-10-02
Clean up directory structure for DMs

mozzius:
This was a mess but we could never change it due to fear of merge conflicts. I think we're good now, so let's sort it out

This PR contains zero code changes, just file moves/renames and import fixes

There's still an awkward split between `/screens/Messages/components` and `/components/dms` but it's much more palatable 

# Test plan

Make sure lint passes


estrattonbailey (approved):
Works for me 👍

PR #5578, opened 2024-10-02
Invalidate author feeds after posting

mozzius:
Does not block composer closing, just invalidates

# Test plan

1. Go to profile screen
2. Make post
3. Watch post appear a second or two later


gaearon (changes requested):
we already have this so need to figure out what went wrong there and find some other solution

Screenshot 2024-10-03 at 11 33 43

PR #5592, opened 2024-10-03
Use new sheets for GIF select (delete platform branching)

mozzius:
I manually used a native formSheet for the GIF select dialog. No longer necessary! So we can delete that.

Also includes a fix for round buttons with RNGH


haileyok:
Seems like Screenshot 2024-10-03 at 5 04 38 PM is making the border for on-press square

haileyok:
Turns out we don't need RNGH afterall 😎

PR #5593, opened 2024-10-03
Track links and embeds in the composer reducer

gaearon:
This adds tracking of remaining embed/media types to the composer reducer.

They're not being used for anything yet, but you should roughly expect their state to mirror how the composer works today. (There are also some improvements. For example, pasting a quote link after a regular link will correctly mark a quote in the reducer state, whereas the old composer logic was not able to handle that.)

In future PRs, I will be making this the source of truth for two things:

- Composer preview
- Publishing a post

However we're missing some pieces to be able to do that, so for now it's not connected yet.

Note that in the draft, we don't distinguish between different types of links (other than quote/other distinction). This is because they're only relevant for preview and for posting. They're not important in the draft data model because the reducer behavior ("which field gets replaced by which field in which cases") doesn't care about the contents of the link beyond that.

## Test Plan

Add `console.log(composerState)` to `Composer.tsx`. Verify the `composerState.embed` fields evolve as you'd expect when you add GIF, add/remove media, paste links, remove GIF, remove links, etc.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -190,7 +190,7 @@ export const ComposePost = ({
   // TODO: Move more state here.
   const [composerState, dispatch] = useReducer(
     composerReducer,
-    {initImageUris},
+    {initImageUris, initQuoteUri: initQuote?.uri},
     createComposerState,
   )
 
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -337,9 +337,9 @@ export const ComposePost = ({
 
   const onNewLink = useCallback(
     (uri: string) => {
+      dispatch({type: 'embed_add_uri', uri})
       if (extLink != null) return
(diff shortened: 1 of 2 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -337,6 +337,7 @@ export const ComposePost = ({
 
   const onNewLink = useCallback(
     (uri: string) => {
+      dispatch({type: 'embed_add_uri', uri})
```
gaearon, line 340: Intentionally placed before the `return`. The logic for whether to handle this URI or ignore it is in the reducer now.

estrattonbailey: May want a comment here? Feels like something someone might run over in the future

estrattonbailey: Oh I get it, we're gonna remove the logic below 👍

estrattonbailey (approved):
Omg it's so much nicer

PR #5595, opened 2024-10-03
Make composer reducer source of truth for images/video when publishing

gaearon:
Stacked on #5593 although unrelated.

---

The first commit moves remaining video-related state into the video reducer.

The second commit finally makes the `composerState` source of truth for the publish function — although for now, in a limited way. Only images and videos are being read from there. We're going to be moving more things over there soon.

## Test Plan

For first commit, upload video, verify you can edit alt text. Add some subtitles (like [this file](https://gist.githubusercontent.com/samdutton/ca37f3adaf4e23679957b8083e061177/raw/e19399fbccbc069a2af4266e5120ae6bad62699a/sample.vtt)). Verify they show up after posting. Verify they clear after removing and readding a video.

For second commit, verify that generally posting images and videos works like before.

### DIFF
--- a/src/view/com/composer/videos/SubtitleDialog.tsx
+++ b/src/view/com/composer/videos/SubtitleDialog.tsx
@@ -22,13 +22,13 @@ import {SubtitleFilePicker} from './SubtitleFilePicker'
 
 const MAX_NUM_CAPTIONS = 1
 
+type CaptionsTrack = {lang: string; file: File}
+
 interface Props {
   defaultAltText: string
-  captions: {lang: string; file: File}[]
+  captions: CaptionsTrack[]
   saveAltText: (altText: string) => void
-  setCaptions: React.Dispatch<
-    React.SetStateAction<{lang: string; file: File}[]>
-  >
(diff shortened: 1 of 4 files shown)


src/view/com/composer/videos/SubtitleDialog.tsx
```diff
@@ -22,13 +22,13 @@ import {SubtitleFilePicker} from './SubtitleFilePicker'
...
+
 interface Props {
   defaultAltText: string
-  captions: {lang: string; file: File}[]
+  captions: CaptionsTrack[]
   saveAltText: (altText: string) => void
-  setCaptions: React.Dispatch<
-    React.SetStateAction<{lang: string; file: File}[]>
-  >
+  setCaptions: (updater: (prev: CaptionsTrack[]) => CaptionsTrack[]) => void
```
gaearon, line 31: this is slightly more narrow than the previous type because we only use the `setCaptions(state -> state)` signature.

estrattonbailey (approved):
All LGTM! This is ace 👍

PR #5597, opened 2024-10-03
Font tweaks

estrattonbailey:
Noticed that `InterVariable` renders a little light on Android. So this PR:
- renames font weight tokens to be agnostic of their actual file names
- bumps up those font weight tokens for Android vs iOS & web by one weight step
- uses those tokens for the source of truth for the old `Text` component via the old `theme`
- fixes a stray `fontWeight` in `ProfileHeaderDisplayName`, sets to `heavy` since it's large text
- fixes a stray `fontWeight` in the drawer
- removes the old `Html` file which has never been used afaik


haileyok (approved):
Looks solid

Issue #3234, opened 2024-03-18
Posts from muted users show up in search results


meicpalmer:
When searching, posts from muted users show up in the results, with the label "Account Muted" on them. The full content of the post is visible.

Expected behavior is for posts from muted users not to show up in the search results.

 - Platform version: Web (Firefox)
 - App version: Current? (don't know the version number)

surfdude29:
I just tested and I can confirm that I see this too. On web (1.73) the whole post is shown, just with the 'Account muted' label (as described above). On iOS (1.72) the post is hidden behind 'Post by muted user'. I don't know if this is how it's designed to work, but in search results I also would expect a post from a muted user not to be shown at all.

nouskjp:
Hi, I am a Japanese Bluesky user. I am experiencing the same problem with the browser version. It is not happening on the app version. I muted my sub-account as a test and took a screenshot of the result.

https://bsky.app/profile/nousk.jp/post/3ko44mecvuu2h

Here is my environment

Browser
Mac mini macos13 chrome version: 122.0.6261.129 safari version: 18615.3.12.11.2

Apps
iPhone12 IOS version 17.1.1 BlueSky app version 1.72

I do not understand English. All are machine translations by deepL. I hope there is no mistranslation.

nouskjp:
Hello. I updated the app version today and the posts of muted users were not filtered. The same was true for the browser version.

Browser
Mac mini macos13 chrome version: 122.0.6261.129 safari version: 18615.3.12.11.2

Apps
iPhone12 IOS version 17.1.1  App Version 1.73

mschwendener:
> posts from muted users show up in the results, with the label "Account Muted" on them. The full content of the post is 
> visible.

Also happens when I click on a hashtag.

meicpalmer:
Will anything be done about this? It's been two months.

surfdude29:
My opinion is that this is definitely a bug, especially on web where the full content of posts from muted users are shown. But I agree that posts from muted users should not appear at all in search results.

rotanen:
There are numerous posts on Bluesky of people who've noticed this and complain about it.

meicpalmer:
There are numerous posts on Bluesky of people complaining about this.

Nayelianne:
Being able to hide muted things from search results is very much needed, since when someone you don't want to see is spamming a particular tag, it makes it hard to filter the search results and find what you need.

[...]

PR #5599, opened 2024-10-03
Filter posts containing mute words from search and notifications

estrattonbailey:
Copies out the filtering and page reuse logic from `post-feed.ts` into these queries and applies post moderation to applicable data. This _entirely_ filters out posts containing muted words from both search and from your notifications.

### DIFF
--- a/src/state/queries/notifications/feed.ts
+++ b/src/state/queries/notifications/feed.ts
@@ -17,7 +17,12 @@
  */
 
 import {useCallback, useEffect, useMemo, useRef} from 'react'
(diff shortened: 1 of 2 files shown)


dolciss:
Let me ask you a question.
Since this change uses @atproto/api's ModerateAPI, 
will it filter not only mute words, but also labeling, account mutes and blocks?

surfdude29:
This is definitely a welcome change 🙏

Will it be a step towards fixing #3234? Or is that an entirely separate issue? 🤔

mozzius (approved):
Works well! Big improvement

PR #5603, opened 2024-10-04
fix event not firing

haileyok:


src/App.native.tsx
```diff
@@ -168,6 +169,7 @@ function App() {
   const [isReady, setReady] = useState(false)
 
   React.useEffect(() => {
+    setBackgroundColorAsync('#000')
```
mozzius, line 172: We don't actually need this - it's just that it doesn't get reset to the default black when you reload the dev build. If you reboot the sim it'll be correct

haileyok: Ah, so it'll be black in regular builds then? If so then it seems fine to rm this.

Actually, probably preferable. This causes the error screen in the dev client (like for a hard crash) to be black text on black background 🙃

mozzius (approved):
🔥

mozzius (changes requested):
Actually, found a screen which it doesn't work for

Simulator Screenshot - iPhone 16 - 2024-10-04 at 13 08 01

mozzius (approved):

PR #5605, opened 2024-10-04
Properly patch React Native to fix iOS `RefreshControl` bug

haileyok:
## Why

This is a followup to a hack patch introduced in https://github.com/bluesky-social/social-app/pull/3259. This patch does work - and we continue to use it right now - though there is actually a subtle difference in feel between what iOS actually does for a haptic and what we were emulating. Unfortunately it is still a bug on iOS 18 and no fix was introduced to RN. See https://github.com/facebook/react-native/issues/43388.

The good news is that there was some insight made on how to work around the bug in https://github.com/duckduckgo/iOS/pull/2942.

## How

Like the attached PR says, we don't want to actually set the tint color for the view until _after_ the `RCTRefreshControl` has been added to the `RCTScrollView` (`RCTScrollView` is the underlying component of `FlatList`, `ScrollView`, etc. so this will fix all cases).

We'll remap the `tintColor` prop to `customTintColor` so that we can store it, and override the setter for `tintColor` to only set it if the superview is a `UIScrollView`.

Note that we cannot just override the setter and be done with it, since the setter won't necessarily get called again after the superview changes. The setter for `customTintColor` will also call `setTintColor` so that the color is still reactive to any changes made by React (theme change, for example).

## Test Plan

- It builds
- Refreshing the feed results in a haptic

### DIFF
--- a/patches/react-native+0.74.1.patch
+++ b/patches/react-native+0.74.1.patch
@@ -38,37 +38,65 @@ index b0d71dc..41b9a0e 100644
 
  - (void)reactBlur
 diff --git a/node_modules/react-native/React/Views/RefreshControl/RCTRefreshControl.h b/node_modules/react-native/React/Views/RefreshControl/RCTRefreshControl.h
-index e9b330f..1ecdf0a 100644
+index e9b330f..ec5f58c 100644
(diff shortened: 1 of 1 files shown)


haileyok:
Webpack failure is okay 👍

estrattonbailey (approved):

PR #5606, opened 2024-10-04
Use composer state as source of truth for embeds/links on publish

gaearon:
This is the equivalent of #5595, but for embeds, links, and GIFs.

With this PR, we're not using the old `Composer` state variables like `quote`, `extLink`, or `extGif` for posting. Instead, we resolve links/records from the `composerState` managed by the new composer reducer.

This also means that `useExternalLinkFetch` is no longer used as a source of truth for posting. It's still used for previews though. In a follow up, I'll fix that up.

Note that this is a slight performance regression. That's because resolving is always done from scratch during post, and there is no caching applied. (Effectively, link meta and previews now get fetched twice — once on post and once on publish.) This will be fixed in a follow-up PR where I'll move resolving into RQ and hook it up to both places.

I won't merge this on its own without the follow-up PR that fixes perf. So this is mostly just for isolated review.

## Test Plan

Try combinations of quotes, images, GIFs, other records, and link embeds. Verify the posted result makes sense.


src/view/com/composer/state/composer.ts
```diff
@@ -304,9 +308,13 @@ export function createComposerState({
   }
   let quote: Link | undefined
   if (initQuoteUri) {
-    quote = {
-      type: 'link',
-      uri: initQuoteUri,
+    // TODO: Consider passing the app url directly.
```
haileyok, line 311: Seems fine as is, but leaving a note here in case you wanted to remember to come back here.

haileyok:
Tested composer + resulting post with:

- Gif by itself
- Gif with text in post
- Gif with quote (see note below)
- Video by itself
- Video with text in post
- Video with captions
- Video with quote
- Quote by itself
- Quote with text
- Quote with everything above ^

Bugs (not regressions):

- Trying to add a quote or link embed _after_ adding a gif does not work
  - Can try to fix this after merging this stack
- Adding a quote prevents you from adding any other media types
  - The combination of these two makes it _impossible_ to have a post with a gif and a link embed
  - You _can_ add a quote embed before adding a gif, so only links are impossible

haileyok (approved):
Works well from the testing above ^. A few bugs, only one of which we can actually fix and already exists in prod. That can come in a separate PR. Beautiful!

haileyok:
Ah, looks like in the final PR that bug is fixed. So really just the weird gif and link thing we can't solve with frontend alone.

src/view/com/composer/state/composer.ts
gaearon: this one actually seems like it's a bit more convenient to leave as is, we're relying on `initQuote.uri` being an `at://` uri in a different place

PR #5608, opened 2024-10-04
Fetch link previews from RQ

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5606

----

This makes the composer reducer a source of truth for link previews. Since the composer reducer itself doesn't actually store preview-related information, the preview is kept in RQ. That was the key insight from @mary-ext's work in #4163.

This unlocks a few things:

- We can kill the `extLink` / `setExtLink` / `extGif` / `setExtGif` / `useExternalLinkFetch` async monstrosity.
- We no longer prevent you from attaching media when a link card exists. (Media will just take priority.)
- We no longer force you to wait for the link card to upload. You can just press Post and it'll do the right thing.


surfdude29:
> Stacked on #5608

Little typo here as this is #5608, I think it should be #5606?

src/view/com/composer/useExternalLinkFetch.ts
```diff
@@ -1,187 +0,0 @@
...
-          newQuote => {
-            if (aborted) {
-              return
-            }
-            setQuote(newQuote)
-            setExtLink(undefined)
-          },
-          err => {
-            if (err instanceof EmbeddingDisabledError) {
-              setError(_(msg`This post's author has disabled quote posts.`))
```
surfdude29, line 67: Sorry if this is a stupid question... this error message shown when a quotegate is set doesn't seem to have been added back anywhere that I can see, should it be?

gaearon: yup, currently it’ll use the message from the error constructor but i need to fix this up

mozzius (approved):
I couldn't find a way to break it! The only thing I found is that if there's an external card and you paste a bsky.app link, it doesn't show a preview (but does post). However, this is happening on main and also in prod (so predates other composer changes) so I feel fine approving this, we should look into it though

haileyok (approved):
Ran a similar set of tests as the pr this is based on, looks solid.

src/view/com/composer/useExternalLinkFetch.ts
gaearon: @haileyok noting this ideally needs to be fixed — probably with a catch in Composer around post call and then rethrow with localized better error message

gaearon:
> The only thing I found is that if there's an external card and you paste a bsky.app link, it doesn't show a preview (but does post). However, this is happening on main and also in prod (so predates other composer changes) so I feel fine approving this, we should look into it though

Yea that’s an existing issue — everything is wired up to get this fixed now but we need to change the preview components in the composer to know how do display other embed types (and then return those embed types from RQ instead of just display names).

src/view/com/composer/useExternalLinkFetch.ts
gaearon: https://github.com/bluesky-social/social-app/pull/5644

haileyok: 👍 thanks for hitting it

gaearon:
[ok](https://github.com/bluesky-social/social-app/pull/5647)

PR #5611, opened 2024-10-04
Add alt text limit to image dialog

estrattonbailey:
Updates alt text limit to 2k characters for both images and GIFs + updates the UI with our char counter.

CleanShot 2024-10-04 at 15 51 39@2x
CleanShot 2024-10-04 at 15 51 34@2x
CleanShot 2024-10-04 at 15 55 41@2x
CleanShot 2024-10-04 at 15 56 56@2x
CleanShot 2024-10-04 at 16 01 45@2x

### DIFF
--- a/src/view/com/composer/GifAltText.tsx
+++ b/src/view/com/composer/GifAltText.tsx
@@ -14,11 +14,13 @@ import {
 import {enforceLen} from '#/lib/strings/helpers'
 import {isAndroid} from '#/platform/detection'
 import {Gif} from '#/state/queries/tenor'
+import {CharProgress} from '#/view/com/composer/char-progress/CharProgress'
 import {atoms as a, native, useTheme} from '#/alf'
 import {Button, ButtonText} from '#/components/Button'
 import * as Dialog from '#/components/Dialog'
 import * as TextField from '#/components/forms/TextField'
 import {Check_Stroke2_Corner0_Rounded as Check} from '#/components/icons/Check'
+import {CircleInfo_Stroke2_Corner0_Rounded as CircleInfo} from '#/components/icons/CircleInfo'
 import {PlusSmall_Stroke2_Corner0_Rounded as Plus} from '#/components/icons/Plus'
 import {PortalComponent} from '#/components/Portal'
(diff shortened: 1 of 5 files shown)


src/view/com/composer/GifAltText.tsx
```diff
@@ -114,61 +114,83 @@ export function GifAltText({
...
-  const {_} = useLingui()
-  const [altText, setAltText] = useState(initalValue)
-  const control = Dialog.useDialogContext()
-
-  const onPressSubmit = useCallback(() => {
-    onSubmit(altText)
-  }, [onSubmit, altText])
+  const t = useTheme()
+  const {_, i18n} = useLingui()
+  const [altText, setAltText] = useState(altTextRef.current)
```
gaearon, line 129: we shouldn’t read refs during render. 

this might be ok but why are we using a ref here? isn’t this a controlled input anyway?

src/view/com/composer/photos/ImageAltTextDialog.tsx
```diff
@@ -21,32 +25,48 @@ type Props = {
...
 }
 
-export const ImageAltTextDialog = (props: Props): React.ReactNode => {
+export const ImageAltTextDialog = ({
+  control,
+  image,
+  onChange,
+  Portal,
+}: Props): React.ReactNode => {
+  const altTextRef = React.useRef<string>(image.alt)
```
gaearon, line 34: hm, another ref. i’m worrying about proliferation of refs — why are we adding them? they should only be used as escape hatches and i don’t yet see what we’re escaping from. generally adding them for normal data flow needs to have some kind of explanation so we don’t miss this in reviews.

gaearon (changes requested):
let’s make sure we have a solid justification for adding refs — easy way to cause bugs if we’re not careful

haileyok:
example in gif dialog in https://github.com/bluesky-social/social-app/pull/5611/commits/13d195afc3c833dc6ee3fd6e6db66ff12974231b

gaearon (approved):

PR #5579, opened 2024-10-02
Swipe action to leave chat

mozzius:
Turns out `react-native-gesture-handler` has a simple component for this.

https://github.com/user-attachments/assets/3596230d-73b9-47d4-bbbb-b91db5e6575e

# Test plan

Make sure there's no perf implications for low end android

PR #5614, opened 2024-10-04
Swipeable to delete chat, custom swipeable

haileyok:
Same as https://github.com/bluesky-social/social-app/pull/5579, but using a custom swipeable that (imo) is a bit nicer as far as usability (perform action on release instead of swipe open and then tap). Acts similar to the iOS mail deletion.

Can rm other changes or leave them - if we want this.

https://github.com/user-attachments/assets/42bf7329-542a-447f-b813-51db07b2ff88


src/lib/custom-animations/GestureActionView.tsx
```diff
@@ -0,0 +1,404 @@
...
+  useAnimatedReaction,
+  useAnimatedStyle,
+  useDerivedValue,
+  useSharedValue,
+  withSequence,
+  withTiming,
+} from 'react-native-reanimated'
+
+import {useHaptics} from '#/lib/haptics'
+import absoluteFill = StyleSheet.absoluteFill
```
mozzius, line 18: not this again haha

mozzius (approved):
Looks good, although not tested on android. Idk if the mark read thing should have different behaviour if already read? current behaviour is fine though

I would maybe like some nicer icons but we can hit that in a follow-up

src/lib/custom-animations/GestureActionView.tsx
haileyok: Oh shit, thanks for pointing that out lol.

Issue #4918, opened 2024-08-10
Default alt-text provided by Tenor used to be automatically included if you didn't add it; no longer is


jdkruzr:
**Describe the bug**
Default alt-text provided by Tenor used to be automatically included if you didn't add it; no longer is. Not sure when this changed, possibly a couple months ago.

**To Reproduce**

1. Try to post a GIF from Tenor with no alt-text; the media will not have alt-text even though the suggested alt-text shows in the "+ ALT" dialogue.

All platforms.

neehhaa06:
Hello @jdkruzr 

Ideas for Default Alt-Text Missing:

```
    Fetch Suggested Alt-Text: See if Tenor offers suggested alt-text when you publish a GIF.
    Set Alt-Text: Use Tenor's suggested alt-text if no manual alt-text is supplied.
    Post-GIF: Send in the GIF together with the suggested or supplied alt-text.
```

```
function postGif(gifUrl, userAltText) {
    const suggestedAltText = getSuggestedAltTextFromTenor(gifUrl);

    // Use user-provided alt-text or default to suggested alt-text
    const altText = userAltText || suggestedAltText;

    // Post GIF with the selected alt-text
    postToServer(gifUrl, altText);
}

function getSuggestedAltTextFromTenor(gifUrl) {
    // Replace with actual API call to fetch suggested alt-text
    return "Default suggested alt-text";
}

function postToServer(gifUrl, altText) {
    // Implement the logic to post GIF to server
    console.log(`Posting GIF: ${gifUrl} with alt-text: ${altText}`);
}

```

mozzius:
Is this still an issue? If I remember correctly, we fixed this, but I could be wrong

Issue #4225, opened 2024-05-26
ALT compose window for GIFs on iOS 


JimHasFun:
**Describe the bug**

Compose window for ALT text for GIFs obscures composition after two lines.
**To Reproduce**

1. Select GIF
2. Click ALT
3. Type more than two lines of text

**Expected behavior**
Expect that window would move or expand to show text as it's being composed.

**Screenshots**

**Details**

 - Platform: 
 - Platform version:iOS 17.5.1
 - App version:Build version: 1.83.0.236; Bundle info: 9eed70415 (prod); Bundle date: 24052323; Platform: ios

**Additional context**

surfdude29:
I’m able to reproduce this on
Build version: 1.91.1.464; Bundle info: 49ab4867a (prod); Bundle date: 24091901; Platform: ios
running on iOS 17.7 on iPhone SE (3rd generation).

haileyok:
Fixed by https://github.com/bluesky-social/social-app/pull/5557. Should be out early this week.

PR #5615, opened 2024-10-05
Split GifEmbed props

gaearon:
This would help me in an upcoming refactor. Just splitting the props so it doesn't always accept the whole `link` object.

## Test Plan

Verify GIFs with explicit ALT and implicit ALT still show up (and both have alt in DOM, but only explicit one shows badge).

Verify editor still works.

### DIFF
--- a/src/view/com/composer/GifAltText.tsx
+++ b/src/view/com/composer/GifAltText.tsx
@@ -200,7 +200,9 @@ function AltTextInner({
           </Text>
           <View style={[a.align_center]}>
             <GifEmbed
-              link={link}
+              thumb={link.thumb}
+              altText={altText}
+              isPreferredAltText={true}
               params={params}
               hideAlt
               style={[native({maxHeight: 225})]}
--- a/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
+++ b/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
@@ -5,6 +5,7 @@ import {AppBskyEmbedExternal} from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
+import {parseAltFromGIFDescription} from '#/lib/gif-alt-text'
 import {usePalette} from '#/lib/hooks/usePalette'
 import {useWebMediaQueries} from '#/lib/hooks/useWebMediaQueries'
 import {shareUrl} from '#/lib/sharing'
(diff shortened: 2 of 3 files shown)


src/view/com/composer/GifAltText.tsx
```diff
@@ -200,7 +200,9 @@ function AltTextInner({
           </Text>
           <View style={[a.align_center]}>
             <GifEmbed
-              link={link}
+              thumb={link.thumb}
+              altText={altText}
+              isPreferredAltText={true}
```
gaearon, line 205: Note these can just be `altText` and `true` now because we're always passing down the "preferred" version. (In any case they don't really matter because it's a preview and has `hideAlt`. Arguably maybe it shouldn't get any alt at all.)

surfdude29:
> Verify GIFs with explicit ALT and implicit ALT still show up (and both have alt in DOM, but only explicit one shows badge).

Until I just saw this, I was under the mistaken impression that GIFs that didn't display the ALT badge didn't have alt in DOM, so it's good to know that they actually do.

Given that this might be confusing other people too (see #4918), is it planned as part of your refactor to show the badge for GIFs with implicit ALT too? I think this would be helpful and less confusing for users.

surfdude29:
Also, while you're working on the GIF alt composer, would it be possible to squeeze in a fix for #4225? 🙏

gaearon:
This PR is really a drive-by refactor for unrelated things I'm working on so I'm not taking a super close look at how it behaves.

>Given that this might be confusing other people too (see https://github.com/bluesky-social/social-app/issues/4918), is it planned as part of your refactor to show the badge for GIFs with implicit ALT too?

With the behavior currently in main, it won't show the badge until you open the "alt" popup, but once you do (and close it), that alt will be considered explicit. Which is slightly confusing in a different way but probably ok.

gaearon:
Meh I'll just roll it all into one PR.

surfdude29:
> This PR is really a drive-by refactor for unrelated things I'm working on so I'm not taking a super close look at how it behaves.

Ah fair enough 👍

> > Given that this might be confusing other people too (see #4918), is it planned as part of your refactor to show the badge for GIFs with implicit ALT too?
> 
> With the behavior currently in main, it won't show the badge until you open the "alt" popup, but once you do (and close it), that alt will be considered explicit. Which is slightly confusing in a different way but probably ok.

Yeah the behavior on main is certainly an improvement over 1.91.1.

PR #5557, opened 2024-10-01
[Sheets] [Pt. 1] Root PR

haileyok:
This is a base PR for each stacked Sheets PR. Instead of merging each into main, start from #5560 and merge them down into this PR. Once everything is reviewed and ready to go, this PR can get merged into main.

This PR does not add any new behavior, just removes a bit of code so not much testing is needed on this one. Check each PR for a test plan.

PR #5616, opened 2024-10-05
Enclose the entire message in Trans tags for translation

tkusano:
Recent modifications to the label-related code complicated the translation work for languages with word orders different from English. To address this, I made a patch to enclose the entire sentence within a Trans tag.


tkusano:
see comments from @quiple in #5557

haileyok (approved):
Oh! Definitely should have reverted that change on my end, appreciate you finding it and cleaning it up 🙏. Thought we were going to need to break this up but things worked out so we don't have to.

Works on both iOS and Android.

PR #5620, opened 2024-10-05
Fix composer char progress layout

haileyok:
## Why

This if a follow up to https://github.com/bluesky-social/social-app/pull/5611, but it also fixes some things that I realized are existent even in production right now.

For example, currently the language selector will jitter around as we type since the layout is shifting each time the character count updates (different chars take up different amounts of space). I'll be addressing that here as well as fixing the actual bug right now (bad layout, number shows above the wheel rather than to the left of it, oops!)

## Test Plan

Check each progress wheel/counter in each part of the app on both native and web.

#### Web

https://github.com/user-attachments/assets/0dd6906f-7706-417c-9247-c94309ac3623

#### Native

https://github.com/user-attachments/assets/2fe572d3-fc0e-4009-898c-24b00e156036

#### Web

https://github.com/user-attachments/assets/db19fd1c-3549-49b5-857b-697c5d622bbc

#### Native

https://github.com/user-attachments/assets/e7c770ae-a3d0-4f7c-a923-1e9b547737c3

#### Web

https://github.com/user-attachments/assets/4970f469-23ec-417e-b0c4-61092318b35b

#### Native

https://github.com/user-attachments/assets/f4a9ff15-d357-455a-a0dd-5f78dab9d0f6

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -867,48 +867,59 @@ export const ComposePost = ({
           )}
           <View
             style={[
+              a.flex_row,
+              a.py_xs,
+              {paddingLeft: 7, paddingRight: 16},
+              a.align_center,
+              a.border_t,
               t.atoms.bg,
               t.atoms.border_contrast_medium,
-              styles.bottomBar,
+              a.justify_between,
             ]}>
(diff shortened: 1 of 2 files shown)


gaearon (approved):
not tested, code lgtm

PR #5623, opened 2024-10-06
Move remaining composer state into reducer

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5606 and https://github.com/bluesky-social/social-app/pull/5608

---

This completes moving composer state to a reducer. It is now the source of truth for everything we're posting. This has no functional change, but it prepares us for a threaded composer in the future (which would compose the composer reducer).

I've also done a minor naming cleanup.

## Test Plan

Try all composer features.


src/lib/api/resolve.ts
```diff
@@ -68,50 +78,88 @@ export async function resolveLink(
...
+        title: res.data.list.name,
       },
     }
   }
   if (isBskyStartUrl(uri) || isBskyStarterPackUrl(uri)) {
-    const result = await getStarterPackAsEmbed(agent, fetchDid, uri)
+    const parsed = parseStarterPackUri(uri)
+    if (!parsed) {
+      throw new Error(
+        'Unexepectedly called getStarterPackAsEmbed with a non-starterpack url',
```
surfdude29, line 144:
```suggestion
        'Unexpectedly called getStarterPackAsEmbed with a non-starterpack url',
```

haileyok (approved):
Ran through this again, everything looks quite solid! Amazing work here and I'm super excited about the draft capabilities this introduces.

PR #5626, opened 2024-10-06
Fix padding in repost dialog

haileyok:
Swaps out `Inner` for `ScrollableInner` which fixes the padding problem. Going to go through and check the other few places we use `Inner` and - probably - replace them.


haileyok:
Oh heh, I guess the rest of the `Inner` uses are web only so nbd right now. Should still swap those later though just so its clear.

mozzius (approved):

PR #5627, opened 2024-10-06
Fix bounce when expanding sheet on ios

haileyok:


mozzius (approved):
Sounds about right, have not tested

haileyok:
Will let it run now so we can test on device. Worked in sim at least.

Issue #5624, opened 2024-10-06
Username ellipsis in Reposted by and Reply to truncates everything or overlaps text


MetaflameDragon:
### Steps to Reproduce

0. Use the webapp - tested on Chrome and Firefox Desktop
1. Go to any timeline that provides reasons for including posts (e.g. `resonRepost`) or shows the "Reply to" context
2. View one where the username is too long to fit (or resize your browser)
3. The username text should appear uniquely messed up in different browsers
  a. In Firefox, the entire username will be truncated to "..."
  b. In Chrome, the username will be clipped (as if `overflow: hidden`), but reasons beyond me the ellipsis appears _at the start of the line_ (that is, the start of the username), overlapping the text
  c. On Android, the native app displays the ellipsis okay, as does Firefox Android

### Firefox Desktop:
![image](https://github.com/user-attachments/assets/375c4f3b-f710-46cd-9dfc-a19225fa0588)

### Chrome Desktop:
![image](https://github.com/user-attachments/assets/1248b3d7-749e-4311-8807-93dc55aa8cba)
![image](https://github.com/user-attachments/assets/beea9c3a-14a9-4748-8765-6ab3806333c4)

### Reply to example
![image](https://github.com/user-attachments/assets/269035ec-0240-4439-ac76-4fd672a5417e)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Win11; Firefox 130.0.1 (64-bit); Chrome 129.0.6668.90 (Official Build) (64-bit)

### What version of the app are you using?

Build version: 1.91.1; Bundle info: 49ab486 (prod); Bundle date: 24092017; Platform: web

### Additional Information

The behavior for usernames in Reposted by and Reply to is the same.

Both the username and handle in the post header otherwise truncate correctly; ellipsis works okay elsewhere too.

Also tested on https://main.bsky.dev: Build version: 1.91.2; Bundle info: dev (prod); Bundle date: 24100605; Platform: web

The different font may require resizing the browser to cause overflow, but both Firefox and Chrome Desktop screw up the ellipsis the same way.

I tried messing around with disabling or changing some styles for the username element without success.

MetaflameDragon:
Okay so setting a concrete `width` so that it doesn't get limited by the parent container (e.g. `100px`) fixes the issue. `fit-content` etc. didn't work, it must be a concrete size. Weird...

PR #5630, opened 2024-10-06
Fix Repost by username ellipsis

MetaflameDragon:
This addresses #5624.

There was a missing flexbox around the "Reposted by" text, somehow that caused issues for ellipsis with the username. Unfortunately, a flexbox forcefully trims spaces, including the space after "Reposted by ". All of this also applies to "Reply to ".

I messed around with a few solutions and nothing else seemed to work properly (without possibly breaking something else). The ideal solution would be to give the parent `white-space: pre` instead of `nowrap`, but it seems to be controlled strictly by `numberOfLines` and React Native doesn't seem to like accepting it as a direct style. The resulting whitespace collapse behavior trims the space from "Reposted by " and its translations, and so far only nbsp worked.

A nbsp (from a constant) is already used in another place, but here it must be part of the translation string right now (actually, the whole "Reposted by <tag>" string), which... I'm not sure how to resolve correctly. Regardless, this fixes the behavior at least, maybe something else can be figured out with the translation strings.

I also suspect that this isn't an issue on mobile, as ProfileHoverCard seems to become passthrough (?), so the Text elements get inlined correctly. In this case, there are two divs in between, which is part of what causes this behavior. There's an `inline` props on the hover card that seems to go unused - maybe it could be related to this issue of inlining contained text?


MetaflameDragon:
Gonna try to poke this more tomorrow. I'll be honest, I'm not too experienced with the stuff used here, particularly how translation strings work, so I'll have to read up on that more first.

To clarify, __the ellipsis thing is fixed here__, but due to using a non-breaking space, __translation strings are broken instead__. I read about Trans tags providing custom components, which is necessary for using nbsp within translation strings since some of them use text on _both_ sides of the username (and so the username component must be referenced in the strings).

This would all also be much more straightforward if the _whole_ "Reposted by username" string were within the hover card tag instead of just the username (since the hover card breaking the string up is part of what causes the issue), but I'll see about getting the nbsp into the translation strings. Ditto for "Reply to username".

MetaflameDragon:
See the issue (#5624) for screenshots. A link to an existing example would be the current reposts [on my test account](https://bsky.app/profile/test.metaflame.dev) (yeah I made a test alt account just for this lol). Chrome & Firefox versions are also specified in the issue report. I don't believe I have any cosmetic filtering or other modifications enabled (uBlock Origin is disabled etc.), so that shouldn't be a factor.

Make sure to resize the browser window and/or zoom in if the username still doesn't get truncated. The new font on https://main.bsky.dev is significantly narrower, for example. This bug occurs both in "Reposted by" and "Reply to", both of which are in the linked example.

For a bit of technical detail, check the nearest flexbox around the username in the "Reposted by" text, you should see that neither the text element nor the flexbox shrink with the parent properly. This, for some reason paired with the fact that the Hover Card wrapper divs have flexboxes, causes the ellipsis to break. Removing the flex from the wrapper divs didn't seem like a clean solution though, so I tried making the child elements actually flex properly, which... eats leading/trailing spaces (unless they're a nbsp).

_Screenshot with a flexbox outlined_
image

Hope that answers the question!

MetaflameDragon:
All locales have been updated to work with the new string & tested in my browser. Ellipsis now behaves as expected for usernames in "Reposted by" and "Reply to". Locales where the username was at the start or in the middle also truncate as expected, only the name gets ellipsized.

Tested on: https://main.bsky.dev/profile/metaflame.dev/post/3l5tvlhwqwo2p (my & my test alt's usernames are significantly longer right now)

This was tested in the webapp in Firefox and Chrome on both desktop (see #5624 for device details) and on mobile.

Unrelated observed issue: Repost icon left of the text shrinks when window width decreases. This is previous behavior, really minor, and unrelated to this bug.

src/view/com/posts/FeedItem.tsx
```diff
@@ -304,14 +305,14 @@ let FeedItemInner = ({
...
-                style={pal.textLight}
+                style={[pal.textLight, {display: 'flex', flexDirection: 'row'}]}
                 lineHeight={1.2}
                 numberOfLines={1}>
                 {isOwner ? (
                   <Trans>Reposted by you</Trans>
                 ) : (
                   <Trans>
-                    Reposted by{' '}
+                    Reposted by{NON_BREAKING_SPACE}
```
MetaflameDragon, line 315: Maybe it's worth adding a comment to the two locations to explain the change, since these are the only two places where nbsp is used within localized messages?

src/view/com/posts/FeedItem.tsx
```diff
@@ -542,7 +543,7 @@ function ReplyToLabel({
     } else {
       label = (
         <Trans context="description">
-          Reply to{' '}
+          Reply to{NON_BREAKING_SPACE}
```
MetaflameDragon, line 546: Ditto for explanatory comment

surfdude29:
> All locales have been updated to work with the new string

FYI the Cantonese (`zh-HK`) localization was just recently merged in #5479 👀

MetaflameDragon:
Thanks for the heads-up! I didn't catch that in my conflict resolve (because, well, it wasn't a conflict lol)

Also, I think that all the translation files should be regenerated at some point, running the extract script locally made a ton of changes everywhere.

PR #5293, opened 2024-09-12
Tabular numbers to video timestamp

quiple:
Applies tabular-nums to vide timestamp.

| Before | After |
|-|-|
| <video src="https://github.com/user-attachments/assets/5b5f486d-f4a6-4abf-acb5-4ef763efb1dc" /> | <video src="https://github.com/user-attachments/assets/9e150038-27ae-4e3c-bbca-4c8e4d2a5d27" /> |

PR #5631, opened 2024-10-07
Use tabular numbers in `CharProgress`

surfdude29:
As a follow-up to #5620, I noticed that the character count doesn't use tabular numbers anymore. This adds back the style.

Inspired by #5293 🙏


estrattonbailey (approved):
Nice catch!

PR #5637, opened 2024-10-07
[Video] Prevent screen from dimming while in full screen

haileyok:
## Why

Adds new logic from https://github.com/bluesky-social/bluesky-video/pull/4 to the app.

## Test Plan

Full screen a video on both iOS and Android, observe that the screen does not dim while watching the video.


haileyok:
Webpack failure okay 👍

mozzius (approved):
open source ftw 🙌

PR #5638, opened 2024-10-07
Use Redix FocusTrap

estrattonbailey:


mozzius:
`@radix-ui/react-focus-guards` is required to make the focus loop correctly

PR #5647, opened 2024-10-08
Display non-post embeds in composer preview

gaearon:
Resolves https://github.com/bluesky-social/social-app/pull/5608#pullrequestreview-2352564290.

This finally adds previews for non-post records. Including fixing the special case (presence of quote will now correctly hide post record previews since they won't actually make it to the post.)

<s>The "X" button is not aligned perfectly for all cases, feel free to edit it up.</s> Fixed.

Note that https://go.bsky.app/* links don't currently go through this codepath in development because `resolveShortLink` doesn't work in development. So if you try that one, you should see a regular link preview. It looks almost the same tho.

## Test Plan

Screenshot 2024-10-21 at 20 07 11

Screenshot 2024-10-21 at 20 08 33

Screenshot 2024-10-21 at 20 09 14

### DIFF
--- a/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
+++ b/src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
@@ -8,10 +8,6 @@ import {useLingui} from '@lingui/react'
 import {parseAltFromGIFDescription} from '#/lib/gif-alt-text'
 import {shareUrl} from '#/lib/sharing'
 import {parseEmbedPlayerFromUrl} from '#/lib/strings/embed-player'
-import {
-  getStarterPackOgCard,
-  parseStarterPackUri,
-} from '#/lib/strings/starter-pack'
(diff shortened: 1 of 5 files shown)


src/view/com/util/post-embeds/ExternalLinkEmbed.tsx
```diff
@@ -81,10 +70,8 @@ export const ExternalLinkEmbed = ({
               }}
               source={{uri: imageUri}}
               accessibilityIgnoresInvertColors
-              accessibilityLabel={starterPackParsed ? link.title : undefined}
-              accessibilityHint={
-                starterPackParsed ? _(msg`Navigate to starter pack`) : undefined
-              }
+              accessibilityLabel={link.title}
```
gaearon, line 73: oh wait, this should now be `undefined` right? i’m confused about the purpose of this prop now

estrattonbailey (approved):

PR #5648, opened 2024-10-08
Move Dialogs to Radix

mozzius:
Adds Radix's `focus-scope` and `focus-guard` utilities to the dialogs and composer, so that they can compose properly with the dropdown menus. This means that dialogs and menus can now be nested, and the `Escape` key will work properly.

I also did the same to the composer, so now focus loops around when you get to the end

I also added nested sheet support to the menu component on native

# Test plan

1. Test the new `Dialog with menu in it` button in the Storybook (all platforms). Before these changes, this would've caused a recursion bug on firefox, so make sure that doesn't happen
2. Test the dialogs on web, esp. with the keyboard. They should act the same
3. Test the composer and the nested dialogs within it for regressions. again, should act the same but now with looping focus

### DIFF
--- a/src/components/Dialog/index.web.tsx
+++ b/src/components/Dialog/index.web.tsx
@@ -189,7 +189,11 @@ export function Inner({
           },
           flatten(style),
         ])}>
-        <DismissableLayer onFocusOutside={preventDefault} onDismiss={close}>
+        <DismissableLayer
+          onInteractOutside={preventDefault}
+          onFocusOutside={preventDefault}
+          onDismiss={close}
+          style={{display: 'flex', flexDirection: 'column'}}>
           {children}
         </DismissableLayer>
       </Animated.View>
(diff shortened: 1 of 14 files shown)


src/view/com/util/UserBanner.tsx
```diff
@@ -90,14 +93,11 @@ export function UserBanner({
 
   // setUserBanner is only passed as prop on the EditProfile component
   return onSelectNewBanner ? (
-    <EventStopper onKeyDown={false}>
+    <EventStopper onKeyDown={true}>
```
estrattonbailey, line 96: Not sure we even need this anymore

estrattonbailey (approved):
Beautiful

src/components/Dialog/index.web.tsx
```diff
@@ -10,7 +10,9 @@ import {
 import Animated, {FadeIn, FadeInDown} from 'react-native-reanimated'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
-import {FocusScope} from '@tamagui/focus-scope'
```
haileyok, line 13: Can we remove this now?

src/view/com/util/UserBanner.tsx
mozzius: oh cool! I'll follow up with a PR to remove those, wanna test thoroughly first so won't do it here

PR #5660, opened 2024-10-09
Allow nested sheets without boilerplate

mozzius:
Due to the strict necessity of iOS native sheets needing to be nested, we were creating new portal groups manually and then prop drilling to the sub-dialogs. This felt like a leaky abstraction, so this PR handles this automatically ✨

This has let me delete all of the unnecessary prop drilling in more cases. I've kept the prop around for the more general-purpose components for if we want to use it in future, but they're no longer used

How it works

1. Dialogs get the Portal they use from context. the default value of the context is same as before (thus no top-level provider needed) (also can be overridden by prop still)
2. Dialogs set a new default portal provider for their children via `<DefaultPortalOverride>`

# Test plan

1. Test normal dialogs work as before
2. Test nested dialogs work: within composer, prompt in muted words, "dialog with menu in it" in the storybook
3. Make sure it roughly works the same on web - I opted not to override the default portal for dialog content there

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -31,12 +31,12 @@ import {
   DialogOuterProps,
 } from '#/components/Dialog/types'
 import {createInput} from '#/components/forms/TextField'
-import {DefaultPortalOverride, useDefaultPortal} from '#/components/Portal'
 import {BottomSheet, BottomSheetSnapPoint} from '../../../modules/bottom-sheet'
 import {
   BottomSheetSnapPointChangeEvent,
   BottomSheetStateChangeEvent,
 } from '../../../modules/bottom-sheet/src/BottomSheet.types'
+import {BottomSheetNativeComponent} from '../../../modules/bottom-sheet/src/BottomSheetNativeComponent'
 
 export {useDialogContext, useDialogControl} from '#/components/Dialog/context'
 export * from '#/components/Dialog/types'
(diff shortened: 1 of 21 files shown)


src/components/Dialog/index.tsx
```diff
@@ -50,8 +50,10 @@ export function Outer({
   onClose,
   nativeOptions,
   testID,
-  Portal = DefaultPortal,
+  Portal: PortalProp,
```
estrattonbailey, line 53: Do we still need this?

mozzius: Hailey seemed to want to keep it around, I'd be fine to rm personally

src/components/Portal.tsx
```diff
@@ -72,3 +72,23 @@ const DefaultPortal = createPortalGroup()
 export const Provider = DefaultPortal.Provider
 export const Outlet = DefaultPortal.Outlet
 export const Portal = DefaultPortal.Portal
+
+// when in native sheets, we want the default Portal to move to inside the sheet for children of the sheet
+export const DefaultPortalContext = React.createContext(DefaultPortal.Portal)
```
estrattonbailey, line 77: Is "default" the right word here? I feel like we could just call it `PortalContext`. The default is `DefaultPortal`, which is aptly named in this file. But this context overrides, and in the spirit of "not thinking about it", I think `usePortal` or like `useParentPortal` might make more sense? No strong feelings, just spitballing.

mozzius: useDefaultPortal makes sense in the context of it maybe being passed in via prop instead. you shouldn't need to know that the default portal might get overridden if in a sheet

estrattonbailey (approved):
LGTM

src/components/Dialog/index.tsx
haileyok: Oh yea this is fine to RM now 👍

haileyok (approved):
Thank you for all the feedback on this 🙏

PR #5661, opened 2024-10-09
Move all portaling logic into `BottomSheet`, remove from our own `Dialog` wrapper

haileyok:


modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
```diff
@@ -0,0 +1,103 @@
...
+              flex: 1,
+              backgroundColor,
+            },
+            Platform.OS === 'android' && {
+              borderTopLeftRadius: cornerRadius,
+              borderTopRightRadius: cornerRadius,
+            },
+          ]}>
+          <View onLayout={this.updateLayout}>
+            <BottomSheetPortalProvider>{children}</BottomSheetPortalProvider>
```
haileyok, line 97: All the children get their own portal provider, so that any dialogs inside of here get put in the right place.

modules/bottom-sheet/src/BottomSheet.tsx
```diff
@@ -1,100 +1,24 @@
...
-      </NativeView>
+  if (__DEV__ && !Portal) {
+    throw new Error(
+      'BottomSheet: You need to wrap your component tree with a <BottomSheetPortalProvider> to use the bottom sheet.',
     )
   }
-}
+
+  return (
+    <Portal>
```
haileyok, line 20: Whatever the closest portal is we'll wrap the native component with.

modules/bottom-sheet/src/BottomSheet.tsx
```diff
@@ -1,100 +1,24 @@
...
-            },
-            Platform.OS === 'android' && {
-              borderTopLeftRadius: cornerRadius,
-              borderTopRightRadius: cornerRadius,
-            },
-          ]}>
-          <View onLayout={this.updateLayout}>{children}</View>
-        </View>
-      </NativeView>
+  if (__DEV__ && !Portal) {
```
gaearon, line 13: DEV-only throws are generally something to avoid because this significantly changes the code path in production. It means that if someone were to hit this in prod, they would see completely different behavior. So it would be hard to diagnose. In some cases, it can be _worse_ than crashing — for example if it accidentally uncovers a bug that causes some bigger problems (like losing or misinterpreting user data). I don’t mean that this would happen here but we should take it as a rule to throw both in PROD _and_ DEV when we do throw. Or it needs to change to be non-throwing.

PR #5663, opened 2024-10-09
Update the email verification dialog

haileyok:
### Web Desktop

Screenshot 2024-10-10 at 12 11 08 PM

Screenshot 2024-10-09 at 1 58 49 PM

## Web Mobile

Screenshot 2024-10-10 at 12 11 16 PM

Screenshot 2024-10-09 at 1 59 17 PM

## Native

RocketSim_Screenshot_iPhone_16_Pro_6 3_2024-10-10_12 07 49

Screenshot 2024-10-09 at 1 59 30 PM


src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+                <Text style={[a.text_md, a.leading_snug, a.font_bold]}>
+                  {currentAccount?.email}
+                </Text>{' '}
+                to verify it's you.{' '}
+                <Text
+                  onPress={() => {
+                    control.close(() => {
+                      openModal({name: 'change-email'})
+                    })
+                  }}
```
estrattonbailey, line 143: Can use `InlineLinkText` with a preventDefault and return false to give us underline on hover

src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+      return
+    }
+
+    setIsProcessing(false)
+    setDidVerify(true)
+    setCurrentStep('StepThree')
+  }
+
+  return (
+    <Dialog.ScrollableInner label={_(msg`Verify email dialog`)}>
```
estrattonbailey, line 119: I think on web we can do what we do with prompt and make this _not_ full width. Feels stretched rn.

src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+              size="large"
+              onPress={() => control.close()}>
+              <ButtonText>
+                <Trans>Close</Trans>
+              </ButtonText>
+            </Button>
+          ) : null}
+        </View>
+      </View>
+    </Dialog.ScrollableInner>
```
estrattonbailey, line 245: Can add a `Close` down here for web

src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+            </View>
+          ) : null}
+          <Text style={[a.text_md, a.leading_snug]}>
+            {currentStep === 'StepOne' ? (
+              <Trans>
+                You'll receive an email at{' '}
+                <Text style={[a.text_md, a.leading_snug, a.font_bold]}>
+                  {currentAccount?.email}
+                </Text>{' '}
+                to verify it's you.{' '}
```
surfdude29, line 137: Just wondering, might it make more sense to close the `<Trans>` tag here, rather than L151?

estrattonbailey (approved):
All works great, just a few design nitpicks

src/components/dialogs/VerifyEmailDialog.tsx
haileyok: This expects `to` as a prop :/

src/components/dialogs/VerifyEmailDialog.tsx
haileyok: 👍

PR #5664, opened 2024-10-09
Ensure app clip works even with `starter-pack`

haileyok:
An unlikely edge case - which probably doesn't explain the thing Apple is seeing - that is possible to hit. The user would need to actually be signed in on web for this to be possible, since otherwise they'd get redirected to `/start` rather than `/starter-pack`, but might as well fix it here.

### DIFF
--- a/modules/BlueskyClip/ViewController.swift
+++ b/modules/BlueskyClip/ViewController.swift
@@ -81,7 +81,7 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
   func isStarterPackUrl(_ url: URL) -> Bool {
     var host: String?
     if #available(iOS 16.0, *) {
-      host = url.host() ?? ""
+      host = url.host()
     } else {
       host = url.host
     }
--- a/modules/BlueskyClip/ViewController.swift
+++ b/modules/BlueskyClip/ViewController.swift
@@ -42,7 +42,7 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
           let payload = try? JSONDecoder().decode(WebViewActionPayload.self, from: data) else {
       return
     }
-
+    
     switch payload.action {
     case .present:
       guard let url = self.starterPackUrl else {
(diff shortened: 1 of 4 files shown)


src/view/screens/Home.tsx
```diff
@@ -4,35 +4,50 @@ import {useFocusEffect} from '@react-navigation/native'
...
 
   React.useEffect(() => {
+    if (isWeb && !currentAccount) {
+      const getParams = new URLSearchParams(window.location.search)
+      const splash = getParams.get('splash')
+      if (splash === 'true') {
+        setShowLoggedOut(true)
+        return
+      }
+    }
```
haileyok, line 49: Need a way to show the splash screen on web when signed out, so we'll just add a `splash` param that we can check.

src/view/com/auth/SplashScreen.web.tsx
```diff
@@ -28,6 +29,15 @@ export const SplashScreen = ({
...
   const {isTabletOrMobile: isMobileWeb} = useWebMediaQueries()
+  const [showClipOverlay, setShowClipOverlay] = React.useState(false)
+
+  React.useEffect(() => {
+    const getParams = new URLSearchParams(window.location.search)
+    const clip = getParams.get('clip')
+    if (clip === 'true') {
+      setShowClipOverlay(true)
+    }
+  }, [])
```
haileyok, line 40: If we use `clip=true` on the splash, we should show the app download overlay.

modules/BlueskyClip/ViewController.swift
```diff
@@ -66,23 +66,51 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
...
+      return false
+    case "go.bsky.app":
+      if url.pathComponents.count == 2 {
+        return true
+      }
+      return false
+    default:
+      return false
+    }
+  }
```
haileyok, line 104: This will catch all possibilities: go.bsky.app, bsky.app/start, and bsky.app/starter-pack.

modules/BlueskyClip/ViewController.swift
```diff
@@ -66,23 +66,51 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
...
-    let urlString = "\(url.absoluteString)?clip=true"
-    if let url = URL(string: urlString) {
-      self.webView?.load(URLRequest(url: url))
+    if isStarterPackUrl(url) {
+      let urlString = "\(url.absoluteString)?clip=true"
+      if let url = URL(string: urlString) {
+        self.webView?.load(URLRequest(url: url))
+      }
+    } else {
+      self.webView?.load(URLRequest(url: URL(string: "https://bsky.app/?splash=true&clip=true")!))
```
haileyok, line 113: If anything other than a real starterpack link comes in, we want to just send the user to the splash and present the appclip overlay

estrattonbailey (approved):
Makes sense to me!

PR #5669, opened 2024-10-10
[Sheets] Allow specifying an initial height

mozzius:
Adds a new `initialHeight` prop to the iOS sheets which allows them to open at half or full size initially

I wasn't able to see how you'd go about doing this on Android

https://github.com/user-attachments/assets/01ab832b-dadf-4ec1-8506-3cab329924d9

### DIFF
--- a/modules/bottom-sheet/ios/SheetViewController.swift
+++ b/modules/bottom-sheet/ios/SheetViewController.swift
@@ -20,42 +20,68 @@ class SheetViewController: UIViewController {
     }
   }
 
-  func setDetents(contentHeight: CGFloat, preventExpansion: Bool) {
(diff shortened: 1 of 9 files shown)


modules/bottom-sheet/ios/SheetViewController.swift
```diff
@@ -20,42 +20,68 @@ class SheetViewController: UIViewController {
...
+            }
+          ]
+
+          if initialHeight == "half" {
+            if (screenHeight - 100) / 2 > contentHeight {
+              sheet.detents.insert(.medium(), at: 0)
+            } else {
+              sheet.detents.append(.medium())
+            }
+          }
```
mozzius, line 63: This part is sus, I don't know the proper way to determine the real height of `.medium`

PR #5675, opened 2024-10-10
Add regional labeler for Germany

estrattonbailey:


mozzius (approved):
Sure, looks good I think?

PR #5676, opened 2024-10-10
Make default banner more palatable

estrattonbailey:
With our avatar generation (even without) this blue default banner color is pretty garish. This tones it back to look a little nicer out of the gate.

CleanShot 2024-10-10 at 12 10 28@2x
CleanShot 2024-10-10 at 12 10 26@2x

Before
CleanShot 2024-10-10 at 12 10 38@2x
CleanShot 2024-10-10 at 12 10 40@2x

### DIFF
--- a/src/view/com/util/UserBanner.tsx
+++ b/src/view/com/util/UserBanner.tsx
@@ -106,7 +106,7 @@ export function UserBanner({
               ) : (
                 <View
                   testID="userBannerFallback"
-                  style={[styles.bannerImage, styles.defaultBanner]}
+                  style={[styles.bannerImage, t.atoms.bg_contrast_25]}
                 />
               )}
               <View style={[styles.editButtonContainer, pal.btn]}>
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Sure I'm +1. What's it look like in light mode?

mozzius (approved):
yessss

estrattonbailey:
Good call good call
CleanShot 2024-10-10 at 15 34 16@2x
CleanShot 2024-10-10 at 15 34 23@2x

estrattonbailey:
Additional work on the Edit Profile dialog can happen in Sam's other PR imo

layeredstrange:
On a first load of a new account, you could display an image in the header with instructions on changing the header as a lot of peeps obviously don't know how or choose not to. Something like this.
Banner Test 1400x440

The image included doesn't line up perfectly as I still don't know the precise resolution of the profile header.

PR #5683, opened 2024-10-10
Add util for link static clicks

estrattonbailey:


src/components/TagMenu/index.tsx
```diff
@@ -4,7 +4,6 @@ import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
 
-import {makeSearchLink} from '#/lib/routes/links'
```
mozzius, line 7: You can delete this util now I suppose?

src/components/TagMenu/index.tsx
```diff
@@ -102,18 +101,13 @@ export function TagMenu({
                 ]}>
                 <Link
                   label={_(msg`Search for all posts with tag ${displayTag}`)}
```
mozzius, line 103: Maybe we should change the label to "View all posts with tag ${displaytag}" since it doesn't push to search anymore?

mozzius (approved):
Good idea

src/components/TagMenu/index.tsx
estrattonbailey: We may need it again in the future, might as well leave it imo

src/components/TagMenu/index.tsx
estrattonbailey: Good call, updated!

PR #5684, opened 2024-10-10
Move setup for Android sheet

haileyok:


modules/bottom-sheet/android/src/main/java/expo/modules/bottomsheet/BottomSheetView.kt
```diff
@@ -155,67 +155,63 @@ class BottomSheetView(
     if (this.isOpen || this.isOpening || this.isClosing) return
 
     val contentHeight = this.getContentHeight()
-
     val dialog = BottomSheetDialog(context)
     dialog.setContentView(dialogRootViewGroup)
     dialog.setCancelable(!preventDismiss)
-    dialog.setOnShowListener {
```
haileyok, line 162: Moving all of this out of the `setOnShowListener` to just all happen prior. If we _start_ in the right position (rather than move there from `Hidden`) it appears to "slide" how we expect.

estrattonbailey (approved):

PR #5685, opened 2024-10-10
Fix sheets to work nicely on ios 15

haileyok:
https://github.com/user-attachments/assets/53a616bc-3064-4ed4-b98f-04534eafe48b

### DIFF
--- a/modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
+++ b/modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
@@ -23,10 +23,13 @@ const NativeView: React.ComponentType<
 
 const NativeModule = requireNativeModule('BottomSheet')
 
+const isIOS15 = Platform.OS === 'ios' && Number(Platform.Version) < 16
+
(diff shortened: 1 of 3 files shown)


modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
```diff
@@ -67,6 +70,19 @@ export class BottomSheetNativeComponent extends React.Component<
...
+      const {viewHeight} = this.state
+      if (viewHeight < screenHeight / 2) {
+        extraStyles = {
+          height: viewHeight,
+          marginTop: screenHeight / 2 - viewHeight,
+          borderTopLeftRadius: cornerRadius,
+          borderTopRightRadius: cornerRadius,
+        }
+      }
+    }
```
haileyok, line 84: Because we cannot actually use a custom detent on the native side on iOS 15, we can do some JS style hacks to make it appear the same.

haileyok:
cc @surfdude29 thanks for the info!!

estrattonbailey (approved):

PR #5687, opened 2024-10-10
Bump icon size for small size buttons

estrattonbailey:
This was an oversight, should definitely be a size up.
CleanShot 2024-10-10 at 18 28 23@2x


pfrazee (approved):
Yeah checked around on web, looks right

PR #5690, opened 2024-10-10
Fix in pt-br locale

Mutrano:
example image
the goal is to change sentences in the muted words window from something like "Expirada em 4 dias" to "Expira em 4 dias"


Mutrano:
I'm closing this since it was already merged

PR #5692, opened 2024-10-11
Tweak threadgate button sizing in the composer on Web

pfrazee:
Reduces the pad on Web so the button isnt so dominant

|Before|After|
|-|-|
|CleanShot 2024-10-10 at 16 59 00@2x|CleanShot 2024-10-10 at 16 58 39@2x|

Mobile remains unchanged

|Mobile|
|-|
|CleanShot 2024-10-10 at 17 11 57@2x|


estrattonbailey (approved):
Nice yeah cool with me to override on a case by case

PR #5694, opened 2024-10-11
Disable contextual alternates only in Inter

quiple:
Disables contextual alternates only in Inter because non-common contextual alternates such as `0x10` and `=> ->` are characteristic of Inter.


estrattonbailey (approved):
Nice, good idea

PR #5695, opened 2024-10-11
Move intent handler to a child of `InnerApp`

haileyok:
## Why

We did some tweaking for the composer state, and that provider is now being used beneath the `useIntentHandler()` hook. We can move this to `Shell`, which actually gives us some additional benefits:

- Composer doesn't open until the app is seemingly "ready" (no longer opens with the splash screen as could happen right now)
- We ensure that the session is ready to go before attempting to use an intent

We should check that everything still works correctly though.

## Test Plan

- Compose intents work
  - Video
  - Images
- Verify email intent works


mozzius (approved):
Not tested, looks fine

PR #5713, opened 2024-10-11
Update TextField font family

quiple:
Enables the theme font Inter to be applied to TextField.

Before & after in Windows:
Screenshot 2024-10-11 at 14-15-52 검색 — Bluesky
Screenshot 2024-10-11 at 14-15-57 Search — Bluesky


quiple:
#6288 seems to be a better way to go.

PR #5714, opened 2024-10-11
Fix performance of feed reordering and add layout animations

pfrazee:
Changes the "Edit Feeds" screen to bundle all of the changes into one save action. Also adds animations. Tested on web, ios sim, and android sim.

|Mobile|Desktop|
|-|-|
|https://github.com/user-attachments/assets/9c8f6fe6-da36-47f0-871c-70f4d42a8be7|https://github.com/user-attachments/assets/19258a16-e5e6-438a-abfc-c8984a3b5370|

### DIFF
--- a/src/view/screens/SavedFeeds.tsx
+++ b/src/view/screens/SavedFeeds.tsx
@@ -5,18 +5,18 @@ import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useFocusEffect} from '@react-navigation/native'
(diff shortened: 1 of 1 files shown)


haileyok (approved):
lgtm, works on both ios and web. also checked reduced motion. could maybe play w/ animation duration (150?)

estrattonbailey:
Niiiiice. Long overdue.

New brain worm to fix one day: the non-centered `ViewHeader` title when there's a button in there lol.

Also one small thing, I feel like the save button may get missed. Unsure on options. Maybe instead of the button in the header we add a section below the header so we can add text like "When you're done press save" or something.

pfrazee:
@estrattonbailey agree, we should add a todo to replace that header

PR #5718, opened 2024-10-11
Present in-app browser as sheet

mozzius:
Presents in-app browser as sheet

Commit 1 has the meaty changes 2ae83fa6ed5c344c5bd3875a161a7b64b2dd6736
The other commits move the hook out of preferences, since that was a weird place for it to be


pfrazee (approved):
Works & feels great

PR #5723, opened 2024-10-11
Move composer open shortcut to shell

haileyok:
Moving the shortcut since we also moved the composer state provider, and they were both within the same component.


haileyok:
Needs an additional tweak for signed out.

pfrazee (approved):

PR #5730, opened 2024-10-11
shift hitslop of avi follow button

haileyok:
A smaller avatar makes it a little bit harder to "tap". The avi follow button's hitslop is currently extending into the avatar a bit, and because the overall avatar hit area is smaller, its harder to "correctly" hit it unless you tap the top of the avatar.

At least on a simulator, this lets us tap in the area the user probably expects and tries to tap in. Need to test on a physical device to be sure though, since simulator is easy to not realize problems with irt hit area.

### DIFF
--- a/src/view/com/posts/AviFollowButton.tsx
+++ b/src/view/com/posts/AviFollowButton.tsx
@@ -87,8 +87,8 @@ export function AviFollowButton({
           hitSlop={{
             top: 0,
             left: 0,
-            right: 3,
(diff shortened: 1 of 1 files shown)


rshigg:
Maybe it's worth it to extend the bottom/right hitslop a bit since the overall touch target is so much smaller now?

haileyok:
@rshigg good call!

pfrazee (approved):

PR #5732, opened 2024-10-11
Focus search input on Search tab soft reset

rshigg:
## Demo

https://github.com/user-attachments/assets/e3e557b4-8036-4dcc-aa4c-7a48763334ff

## Why?

This is the way the search tab works in many apps. Most notably this is something I miss from the bird site. It's not solving a huge problem but not having to stretch my finger to the top of the screen is nice when I'm feeling extra lazy.

## Considerations

- Is `soft-reset` the right event to listen to here? It seemed the most fitting since it's the event that fires when the tab is already selected.
- This is a pattern I see often on iOS but can't speak for Android, does it make sense there?
- The focused state does not get reset across page navigation, is this awkward?
  - Note: It seems that this is also the case if you manually focus the input field.
- If this is an addition for the lazy among us, would it make sense to also add a more convenient way to blur the input?
  - For example: If there are no search results and the user presses on the page body


mary-ext:
Dupe of https://github.com/bluesky-social/social-app/pull/3065 (although it's been a long time since this PR was made)

Yeah, soft-reset is the proper event, are you also checking if you're at the suggestions/explore page?

rshigg:
> Dupe of #3065 (although it's been a long time since this PR was made)

Good callout! I can close this PR if you think it's redundant 👍 

> Yeah, soft-reset is the proper event, are you also checking if you're at the suggestions/explore page?

To my understanding, the explore page is inside `SearchScreenInner` which is rendered by the `Search` screen where I focus the input, so that should have us covered. Unless you mean that we *shouldn't* focus the input when the explore page is visible?

Mattis142:
ik this adds nothing but I just wanted to confirm that yes indeed this is also an interaction that's pretty normal on android 😅

haileyok:
Ahh yes! There was _something_ that was blocking when #3065 was originally submitted that I didn't have time to look at, but we've refactored a lot of stuff around this screen and we might be able to do this now. I'll have to run the sim this coming week to see if I can suss out what the problem was last time...

In either case, I see that this change is just a one liner while there's a bit more stuff going on in #3065. Haven't looked closely at either of these yet, but have you tested this one liner against web as well? Looks like thats what some of the extra stuff was addressing in the other PR.

rshigg:
> Ahh yes! There was _something_ that was blocking when #3065 was originally submitted that I didn't have time to look at, but we've refactored a lot of stuff around this screen and we might be able to do this now. I'll have to run the sim this coming week to see if I can suss out what the problem was last time...

Awesome! Really excited to potentially get this merged if all goes well 😁


> In either case, I see that this change is just a one liner while there's a bit more stuff going on in #3065. Haven't looked closely at either of these yet, but have you tested this one liner against web as well? Looks like thats what some of the extra stuff was addressing in the other PR.

It only runs on mobile currently, since it felt like a mobile app pattern to me. There's definitely an argument to be made to add it web as well though.

rshigg:
@haileyok congrats on the 1.93 release! This feature seems to be getting a lot of traction lately, so might be worth potentially merging into the next release? 🤞😄 

No rush obviously, I appreciate all the hard work the team is putting in.

gaearon (approved):
looks good to me. let's do it for native for now, maybe we'll add to web too

hichemfantar:
looks like this is relevant https://github.com/bluesky-social/social-app/issues/6622

PR #5740, opened 2024-10-14
Translate some missing strings via global i18n instance

mozzius:
There are a few strings shown via toasts that are currently untranslated, since we have yet to restructure the places they are called to allow passing an i18n instance. however, in this circumstance, we can use the global `t` macro to translate them. This is not ideal, since it won't change when the language changes, however it's better than nothing. See the lingui docs:

https://lingui.dev/tutorials/react-patterns#translations-outside-react-components

My reasoning is that it's better for them to be translated but not change when changing language in the app, verses always being in english

# Test plan

Make up some translations for them by generating the i18n files, then check that they get translated after changing language and then restarting the app

### DIFF
--- a/src/lib/media/picker.shared.ts
+++ b/src/lib/media/picker.shared.ts
@@ -3,8 +3,10 @@ import {
   launchImageLibraryAsync,
   MediaTypeOptions,
 } from 'expo-image-picker'
+// TODO: replace global i18n instance with one returned from useLingui -sfn
+import {t} from '@lingui/macro'
 
-import * as Toast from 'view/com/util/Toast'
+import * as Toast from '#/view/com/util/Toast'
 import {getDataUriSize} from './util'
 
 export async function openPicker(opts?: ImagePickerOptions) {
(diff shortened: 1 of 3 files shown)


mary-ext:
I think Lingui's docs is expecting that for React, you'd want to instantiate your own i18n class rather than going for their global singleton.

Right now we've actually just been passing that singleton to the I18nProvider.

https://github.com/bluesky-social/social-app/blob/830b4bee9c738576769b75218dadc6b1de1ba3a3/src/locale/i18nProvider.tsx#L3

So for the picker it *should* be affected by the language changes as they are outside of React's rendering lifecycle.

mozzius:
Oh right I see, so it's perfect acceptable for stuff like toasts then?

mary-ext:
Should be good!

estrattonbailey (approved):
Ah good to know

surfdude29:
Could this method also be used to translate the untranslated strings in `src/lib/api/index.ts`? If so, would you be open to a PR for that?

mozzius:
@surfdude29 go for it! Feel free to PR any loose strings you find using this method, you probably know better than us where they are :)

surfdude29:
> @surfdude29 go for it! Feel free to PR any loose strings you find using this method, you probably know better than us where they are :)

I opened #5750.

Issue #4049, opened 2024-05-16
Show "Hide this post?" prompt after selecting "Show less like this" in the Discover feed


TheRosie689:
**Describe the solution you'd like**

 Could you make the "Hide this post?" prompt show up when selecting "Show less like this" in the Discover feed? 

<img width="1136" alt="Screenshot 2024-05-16 at 9 04 44 AM" src="https://github.com/bluesky-social/social-app/assets/74479954/85b323a0-52e8-449b-bee4-183624865ea6">

PR #5743, opened 2024-10-14
Hide post on show less

mozzius:
Closes https://github.com/bluesky-social/social-app/issues/4049

Shows the following prompt after pressing "show less". Alternatively, we could just automatically hide it, open to doing it either way


mary-ext:
Feels like the expected behavior is more of "I just don't want this to be recommended, remove this from the feed" rather than "I don't want to see it entirely, permanently even" which is what the hide post action does

mozzius:
Hmm yeah probably true

PR #5748, opened 2024-10-14
Move the profile header back-button in front of banner on Android

gpp-0:
On Android, the new GrowableBanner places the back button in profile headers behind the banner, making it inaccessible (i think this was unintentional).
I brought it to the front and made it pressable again


estrattonbailey (approved):
Nice one! Thanks :)

PR #5750, opened 2024-10-14
Translate strings in `src/lib/api/index.ts`

surfdude29:
Using the same method as #5740, this PR marks 7 strings for translation in `src/lib/api/index.ts`.


mozzius (approved):
Thanks dude!

PR #5757, opened 2024-10-15
Add more Android ripples

gpp-0:
I don't know if this was a deliberate design choice but the lack of ripple effects on Android makes the app feel very unresponsive to me, so I tried adding them in some places where I feel they make sense.

#### What I added
* Ripples were already implemented for posts that appear in the home tab, but they don't work on posts that appear in replies, or in the search and notification tabs. Quote embeds also don't have ripples even when the parent post does. I added all of those by setting the noFeedback property on the `view/com/util/Link` component but that looks like a hack and is probably not good
* I added ripples on user avatars that appear in profile headers
* I also added ripples on image/video/gif embeds. On images that are added in the composer, I replaced an invisible `TouchableOpacity` that only covered part of the image so now the entire image is pressable

#### What is missing
* I didn't add ripples on components that use the `components/Link` element because I couldn't figure out how to do it properly. Those include Cards (as in `ProfileCard`, `FeedCard` etc.), user avatars on posts, and embedded external links

### DIFF
--- a/src/components/moderation/PostHider.tsx
+++ b/src/components/moderation/PostHider.tsx
@@ -60,6 +60,7 @@ export function PostHider({
         style={style}
         href={href}
         accessible={false}
+        noFeedback
         onBeforePress={onBeforePress}
         {...props}>
         {children}
@@ -128,6 +129,7 @@ export function PostHider({
       testID={testID}
       style={addStyle(style, styles.child)}
       href={href}
+      noFeedback
       accessible={false}
       {...props}>
       {children}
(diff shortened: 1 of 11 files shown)


gaearon:
Would you mind bringing this up-to-date? Sounds like a good idea

gpp-0:
@gaearon Sorry, I missed that message, I've updated the PR!

Btw this is a super tiny detail but if this is merged then it makes the issue in https://github.com/bluesky-social/social-app/pull/5751 ever so slightly more visible

PR #5771, opened 2024-10-16
Logged out improvments

mozzius:
Adds missing safe area insets to the signup and login flows. Also:
- Replaces X icon with a nicer one
- Fetches all profiles for the account switch at once rather than individually
- Adds insets to the root error boundary

# Test plan

Make sure there's no content leaking into the safe area for all the logged out screens, i.e. auth, deactivation, onboarding, etc


src/components/AccountList.tsx
```diff
@@ -129,9 +135,7 @@ function AccountItem({
             <Text emoji style={[a.font_bold]}>
               {profile?.displayName || account.handle}{' '}
             </Text>
-            <Text emoji style={[t.atoms.text_contrast_medium]}>
-              {account.handle}
-            </Text>
+            <Text style={[t.atoms.text_contrast_medium]}>{account.handle}</Text>
```
estrattonbailey, line 138: Don't need to sanitize the display name here since it's the user's own account, but the `sanitizeHandle` util is used to insert the warning icon for invalid handles, and if that happens we need the `emoji` prop. Pry good to add that sanitization for consistency, idk why it's not here already.

src/screens/Deactivated.tsx
```diff
@@ -104,15 +104,15 @@ export function Deactivated() {
...
+    <View style={[a.util_screen_outer, a.flex_1, t.atoms.bg]}>
       <ScrollView
         style={[a.h_full, a.w_full]}
         contentContainerStyle={{borderWidth: 0}}>
         <View
           style={[
             a.px_2xl,
             {
-              paddingTop: isWeb ? 64 : insets.top,
+              paddingTop: isWeb ? 64 : insets.top + 16,
```
estrattonbailey, line 115: Nit: feel like we could use a lil more padding here on native, and for consistency, could do like `a.pt_2xl.paddingTop` or `tokens.space._2xl`

src/view/com/auth/LoggedOut.tsx
```diff
@@ -57,30 +57,32 @@ export function LoggedOut({onDismiss}: {onDismiss?: () => void}) {
...
+    <View
+      testID="noSessionView"
+      style={[
+        a.util_screen_outer,
+        t.atoms.bg,
+        {paddingTop: insets.top, paddingBottom: insets.bottom},
+      ]}>
       <ErrorBoundary>
         {onDismiss && screenState === ScreenState.S_LoginOrCreateAccount ? (
           <Pressable
```
estrattonbailey, line 69: Nit: could just swap in the `Button` here if you want

estrattonbailey (approved):
Nice nice

src/screens/Deactivated.tsx
estrattonbailey: Note: Android looks fine, iOS felt a little close to me but clearly it's gonna vary a bit, so not a huge deal

src/components/AccountList.tsx
mozzius: Ah ok good to know

src/screens/Deactivated.tsx
mozzius: bumped to `xl`, I think `2xl` might be a bit much even on iOS

PR #5774, opened 2024-10-16
[Settings] External media prefs revamp

mozzius:
# Based on #5773 

Simulator Screenshot - iPhone 16 - 2024-10-16 at 16 30 55

PR #5775, opened 2024-10-16
[Settings] Languages revamp (partial)

mozzius:
# Based on #5774 

Lazy restyling of existing components. I'd like to come back to this and make some nice universal <Select> components, but that's a fairly big job I'll leave for another time

### DIFF
--- a/src/screens/Settings/LanguageSettings.tsx
+++ b/src/screens/Settings/LanguageSettings.tsx
@@ -1,0 +1,274 @@
+import React, {useCallback, useMemo} from 'react'
+import {View} from 'react-native'
+import RNPickerSelect, {PickerSelectProps} from 'react-native-picker-select'
+import {msg, Trans} from '@lingui/macro'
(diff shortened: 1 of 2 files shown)


src/view/screens/Settings/index.tsx
```diff
@@ -28,6 +28,7 @@ import {HandIcon, HashtagIcon} from '#/lib/icons'
 import {makeProfileLink} from '#/lib/routes/links'
 import {CommonNavigatorParams, NativeStackScreenProps} from '#/lib/routes/types'
 import {NavigationProp} from '#/lib/routes/types'
+import {useGate} from '#/lib/statsig/statsig'
```
surfdude29, line 31: looks like this is causing a lint error, think it can be rm?

haileyok (approved):
The middle option needs to flex across the whole screen on native. Looks good on web, and all functions fine.

Screenshot 2024-10-31 at 12 17 43 PM

PR #5776, opened 2024-10-16
Fix font display on Android

estrattonbailey:
After a thorough investigation and a few hints here and there like [this one](https://stackoverflow.com/questions/43021540/custom-font-not-working-in-react-native/58765980#58765980), it would appear that RN for Android doesn't support `fontWeight` or `fontStyle` for custom fonts, and so in turn does not support variable fonts.

So the solve for this is to load separate font files for Android-only, and map weights and styles to the separate font files.

Android still appears to be using alternate glyphs for some characters, such as lowercase `a` (it's using `a.1`), but I haven't been able to determine how to turn those off on Android, and iOS and web still work correctly. So maybe that's an investigation for another time, since this seems to work fine otherwise and Inter technically supports these character alternates i.e. this shouldn't break anything afaik.

### DIFF
--- a/src/alf/fonts.ts
+++ b/src/alf/fonts.ts
@@ -1,7 +1,7 @@
 import {TextStyle} from 'react-native'
 import {useFonts} from 'expo-font'
 
-import {isWeb} from '#/platform/detection'
+import {isAndroid,isWeb} from '#/platform/detection'
 import {Device, device} from '#/storage'
 
 const WEB_FONT_FAMILIES = `system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji"`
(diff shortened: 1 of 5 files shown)


src/alf/fonts.ts
```diff
@@ -78,8 +104,19 @@ export function applyFonts(style: TextStyle, fontFamily: 'system' | 'theme') {
...
+          'Inter-BoldItalic': require('../../assets/fonts/inter/Inter-SemiBoldItalic.otf'),
+          'Inter-Black': require('../../assets/fonts/inter/Inter-ExtraBold.otf'),
+          'Inter-BlackItalic': require('../../assets/fonts/inter/Inter-ExtraBoldItalic.otf'),
+        }
+      : {
+          InterVariable: require('../../assets/fonts/inter/InterVariable.ttf'),
+          'InterVariable-Italic': require('../../assets/fonts/inter/InterVariable-Italic.ttf'),
+        },
+  )
 }
```
haileyok, line 122: Can we platform split this? Pretty sure this will bloat our bundle sizes on web.

haileyok (approved):

src/alf/fonts.ts
estrattonbailey: Fixed. Yeah the bot here was reporting totally assets, but those fonts wouldn't have actually loaded. Much better now though, thanks for the bump 👍

PR #5780, opened 2024-10-16
Move self-label button

haileyok:
Screenshot 2024-10-16 at 3 58 32 PM

Screenshot 2024-10-16 at 3 58 41 PM

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -559,38 +559,30 @@ export const ComposePost = ({
                     <ActivityIndicator />
                   </View>
                 </>
+              ) : canPost ? (
+                <Button
+                  testID="composerPublishBtn"
+                  label={replyTo ? _(msg`Publish reply`) : _(msg`Publish post`)}
+                  variant="solid"
+                  color="primary"
+                  shape="default"
+                  size="small"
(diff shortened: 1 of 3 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -758,22 +744,37 @@ export const ComposePost = ({
...
-                  type: 'update_threadgate',
-                  threadgate: nextThreadgate,
-                })
-              }}
-              style={bottomBarAnimatedStyle}
-            />
-          )}
+          <Animated.View
+            style={[a.flex_row, a.p_sm, t.atoms.bg, bottomBarAnimatedStyle]}>
+            <ScrollView
```
haileyok, line 749: A `ScrollView` because we want to handle accessibility font sizes gracefully.

surfdude29:
This looks great, thanks @haileyok! :)

mary-ext:
`Self-labels` seems non-descriptive, but I understand that there is not a lot of horizontal space for this, could we go for something like `Add labels` or `Add warning`?

haileyok:
Yea seems legit, I had some hesitation around that wording as well. Sorta think we could rephrase the interactions button too.

mozzius (approved):
lgtm. I put my suggestions in a PR, but this is good

PR #5777, opened 2024-10-16
[Settings] App passwords revamp

mozzius:
# Stacked on #5775 

Animations tested on iOS, Android, and Web

https://github.com/user-attachments/assets/21c9f4bd-0412-43df-bfec-dd1270feed28

PR #5781, opened 2024-10-16
[Settings] Change handle dialog

mozzius:
# Based on #5777

https://github.com/user-attachments/assets/35284427-db0c-4732-a7ba-08bc40f05aa5

### DIFF
--- a/src/screens/Settings/components/ChangeHandleDialog.tsx
+++ b/src/screens/Settings/components/ChangeHandleDialog.tsx
@@ -244,7 +244,7 @@ function ProvidedHandlePage({
           <Text style={[a.leading_snug]}>
             <Trans>
               If you have your own domain, you can use that as your handle. This
-              lets you self-verify your identity -{' '}
+              lets you self-verify your identity –{' '}
               <InlineLinkText
                 label={_(msg`learn more`)}
                 to="https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial"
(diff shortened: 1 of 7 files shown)


surfdude29 (commented):
These are two very pedantic suggested tweaks, so please feel free to ignore them: I think these two strings would look slightly better using an en dash rather than a hyphen. (I think most [style](https://www.thepunctuationguide.com/em-dash.html) [guides](https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use) would instead recommend an em dash in these cases, but I'm not sure that would fit Bluesky's style?)

mozzius:
I do love an em dash

src/screens/Settings/components/ChangeHandleDialog.tsx
```diff
@@ -0,0 +1,600 @@
...
+            <Trans>Change Handle</Trans>
+          </Dialog.HeaderText>
+        </Dialog.Header>
+      }
+      contentContainerStyle={[a.pt_0, a.px_0]}>
+      <View style={[a.flex_1, a.pt_lg, a.px_xl]}>
+        {serviceInfoError ? (
+          <ErrorScreen
+            title={_(msg`Oops!`)}
+            message={_(msg`There was an issue with fetching your service info`)}
```
surfdude29, line 102:
```suggestion
            message={_(msg`There was an issue fetching your service info`)}
```
slight suggested tweak

haileyok:
Large font sizes make it quite hard to input a handle here, might need to adjust that.

Also note the "Cancel" button gets kinda messed up w/ the header.

Screenshot 2024-10-31 at 12 52 27 PM

mozzius:
Fixed by setting a maxwidth. I'll fix the header separately

Simulator Screenshot - iPhone 16 - 2024-10-31 at 20 31 49

PR #5782, opened 2024-10-16
Remove double `an` from error message strings

surfdude29:
I noticed that there are [four files](https://github.com/search?q=repo%3Abluesky-social%2Fsocial-app+%22there+was+an+an%22+language%3ATSX&type=code&l=TSX) where `an` is repeated in some strings. This PR removes the repeated words.

Also, I noticed some logger error messages that say `Failed up` which – unless it's dev-speak or something? – I'm guessing should be `Failed to`, so I changed those too.


mozzius (approved):
Nice, yeah that's just an odd typo I think

PR #5785, opened 2024-10-17
Fix up drawer background

estrattonbailey:
Looks essentially the same on all platforms
CleanShot 2024-10-16 at 20 23 51@2x
CleanShot 2024-10-16 at 20 22 48@2x
CleanShot 2024-10-16 at 20 23 42@2x


mozzius (approved):
Pushed up a fix for some clipping I was seeing on web, looks good otherwise

PR #5761, opened 2024-10-15
Clean up drawer

mozzius:
- ALF'd most styles
- Made profile smaller, with extra bold display name
- Made slightly wider
- Made items closer together
- Added link to Chats screen (closes #5516)

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Simulator Screenshot - iPhone 16 - 2024-10-15 at 09 03 31</td>
      <td>Simulator Screenshot - iPhone 16 - 2024-10-15 at 09 00 23</td>
    </tr>
  </tbody>
</table>

PR #5788, opened 2024-10-17
Mark "Logo by ..." in RightNav translatable like it in Drawer

tkusano:
#5761 has marked 'Logo by ...' in Drawer translatable, so marked same message in RightNav translatable.


mozzius (approved):
Thanks!

PR #5790, opened 2024-10-17
Unified 'Chats' in drawers into 'Chat'

auroursa:
Based on #5761.

the desktop leftnav and the mobile bottom-bar accessibility hints used "Chat," so I have standardized it to "Chat" to allows for the use of previously translated strings in lingui.


mozzius (approved):
Good catch!

PR #5795, opened 2024-10-17
Tweak `de` string

### LINKED ISSUE #5559: Known followers line may contain "NaN" if app language is German
### Steps to Reproduce

In settings, select German as app language for default text. In someone's profile in the known followers info line
you should observe something like

    Gefolgt von X, Y and NaN andere

where it should, for example, be:

    Gefolgt von X, Y und 5 anderen

(for "Followed by X, Y, and 5 others")

In the translation the effective third value appears to be a non-number, which results in `NaN` instead of the expected `5`.

### Attachments

Screenshot from 2024-10-01 23-51-30

Android, Web (Desktop)

### Device Info

_No response_

Build version: 1.91.1; Bundle info: 49ab486 (prod); Bundle date: 24092017; Platform: web

### Additional Information

I suppose the bug is in line 2690, where it should read `und {2, plural` instead of `und {1, plural`:

https://github.com/bluesky-social/social-app/blob/5a98ea684d90526e80b8b3ef7554452c4f8db53b/src/locale/locales/de/messages.po#L2688-L2690

Probably after copying the line from 2682 and extending it, it has been forgotten to adjust `1` to `2`.

I'll provide a pull-request if needed.

surfdude29:
This is a tiny PR that cherry-picks a change from #5451 (which is in draft) that fixes a translation error in one of the German strings that I inadvertently introduced and partially fixes #5559.

### DIFF
--- a/src/locale/locales/de/messages.po
+++ b/src/locale/locales/de/messages.po
@@ -2747,7 +2747,7 @@ msgstr "Gefolgt von <0>{0}</0> und <1>{1}</1>"
 
 #: src/components/KnownFollowers.tsx:186
 msgid "Followed by <0>{0}</0>, <1>{1}</1>, and {2, plural, one {# other} other {# others}}"
-msgstr "Gefolgt von <0>{0}</0>, <1>{1}</1> und {1, plural, one {# anderer} other {# andere}}"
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):

estrattonbailey:
Thanks!

PR #5809, opened 2024-10-17
Make use of applyWrites when publishing posts

mary-ext:
From https://github.com/bluesky-social/social-app/pull/4163

Let's get composer back on track again.

Use applyWrites instead of creating post, threadgate and postgate records separately.

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -4,13 +4,17 @@ import {
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
   AppBskyEmbedVideo,
+  AppBskyFeedPost,
+  AppBskyFeedPostgate,
   AtUri,
   BlobRef,
   BskyAgent,
   ComAtprotoLabelDefs,
+  ComAtprotoRepoApplyWrites,
   ComAtprotoRepoStrongRef,
   RichText,
 } from '@atproto/api'
+import {TID} from '@atproto/common-web'
 import {t} from '@lingui/macro'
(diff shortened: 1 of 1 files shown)


src/lib/api/index.ts
```diff
@@ -101,78 +103,87 @@ export async function post(
...
+  const rkey = TID.nextStr()
+  const uri = `at://${agent.assertDid}/app.bsky.feed.post/${rkey}`
+  const date = new Date().toISOString()
+
+  const writes: ComAtprotoRepoApplyWrites.Create[] = []
+
+  // Create post record
+  {
+    const record: AppBskyFeedPost.Record = {
+      createdAt: date,
```
devinivy, line 115: I believe the PDS will fill-in `$type: 'app.bsky.feed.post'` field on the record, but I think it may be good hygiene to do it explicitly here.

gaearon (approved):
this is looking great to me. everything seems to work too. thanks

src/lib/api/index.ts
```diff
@@ -101,78 +103,87 @@ export async function post(
...
-    } catch (e: any) {
-      logger.error(`Failed to create postgate`, {
-        context: 'composer',
-        safeMessage: e.message,
-      })
+    const record: AppBskyFeedPostgate.Record = {
+      ...draft.postgate,
+      createdAt: date,
+      post: uri,
+    }
```
devinivy, line 157: Similar here— this should work as-is but I'd recommend setting `$type` on the record anyway.

PR #5810, opened 2024-10-17
Fix width handling for deactivated screen

estrattonbailey:
There were some crossed wires in the Deactivated screen, causing it to overflow mobile web. Mostly the fix here is ensuring the `#root` el that wraps our app has a width defined.


mozzius (approved):
Confirmed it fixes the Starter packs screen and the deactivated screen

PR #5811, opened 2024-10-17
Add back borders to feed liked-by screen

estrattonbailey:
To match `PostLikedBy` screen


mozzius (approved):
don't think ive ever been to this screen before lol

PR #5834, opened 2024-10-18
Add pwi as interest header for PWI

estrattonbailey:


mozzius (approved):
headers are case insensitive right? lgtm

PR #5836, opened 2024-10-18
tweak downsample rate to 1%

haileyok:


src/lib/statsig/statsig.tsx
```diff
@@ -89,7 +89,7 @@ export function toClout(n: number | null | undefined): number | undefined {
   }
 }
 
-const DOWNSAMPLE_RATE = 0.98 // 98% likely
+const DOWNSAMPLE_RATE = 0.99 // 98% likely
```
surfdude29, line 92:
```suggestion
const DOWNSAMPLE_RATE = 0.99 // 99% likely
```
Suggest updating comment to follow tweak

estrattonbailey (approved):

Issue #5719, opened 2024-10-11
Sharing feed does not work on iOS


yordiak:
### Steps to Reproduce

I use the latest version of the Bluesky app and I cannot share a feed on an iphone 11.

Steps to reproduce the problem:

1. Navigate to a feed (anything you want)
2. Click on the three dots and choose "share feed"
3. Nothing happens

I don't know if the problem appers in older versions, as welll

### What platform(s) does this occur on?

iOS

### What version of the app are you using?

1.92

haileyok:
What version of iOS are you using?

yordiak:
I am using the latest one (18.0.1)

haileyok:
Interesting! I'll have to try and play with it a bit, appreciate the report. By chance do you have reduced motion enabled? (shot in the dark, but there's one thing i can think of that might cause problems there)

yordiak:
I don't know... how can verify if i have enable such a setting?

haileyok:
@yordiak Hmm, no I tested with that setting enabled and it still works fine for me. Can you let me know if sharing a post works if you do so from the `...` on a post and hit "Share"?

yordiak:
@haileyok I tried that, and I cannot share a post either...

surfdude29:
I'm seeing this too sometimes on iOS 17.7, where after tapping **Share feed** the sharing sheet simply never opens, and also sometimes on iOS 15.8.3, where after tapping **Share feed** the Bluesky sheet never closes, the iOS sharing sheet opens and then quickly closes again.

However, sometimes on both iOS versions it works just fine; I haven't been able to find a way to reliably reproduce it so far.

surfdude29:
I’ve also run into this today on iOS 17.7 while trying to share a link to a labeller profile and while trying to share a link to a post.

PR #5849, opened 2024-10-19
Fix share button not working for some users

haileyok:
Fixes https://github.com/bluesky-social/social-app/issues/5719

## Why

This is one that I was unable to reproduce for some time, but was finally able to "accidentally" discover. If you press too long on the "share" button, the dialog actually ends up staying open a tad bit longer. It seems like when that happens, the app isn't able to present the share dialog because the menu sheet is still presented and animating away.

Notably here, we were not calling our menu item `onPress` inside of the `close()` function's callback, but rather in parallel. We've seen in the past where its pretty essential to do that, and it seems like this is another one of those cases.

## Test Plan

Before this PR, open the share sheet and take a bit longer on pressing the "Share" button. It shouldn't open the share sheet. Then, test this pr and do the same. It should still open.

Also worth checking other buttons like "Translate" to ensure they still work.

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -121,10 +121,9 @@ export function Item({children, label, style, onPress, ...rest}: ItemProps) {
       onFocus={onFocus}
       onBlur={onBlur}
       onPress={async e => {
-        await onPress(e)
-        if (!e.defaultPrevented) {
-          control?.close()
-        }
+        control?.close(() => {
+          onPress?.(e)
+        })
       }}
       onPressIn={e => {
         onPressIn()
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -121,9 +121,12 @@ export function Item({children, label, style, onPress, ...rest}: ItemProps) {
       onFocus={onFocus}
       onBlur={onBlur}
       onPress={async e => {
-        await onPress(e)
-        if (!e.defaultPrevented) {
-          control?.close()
+        if (e.defaultPrevented) {
+          await onPress(e)
+        } else {
+          control?.close(() => {
+            onPress(e)
+          })
         }
       }}
       onPressIn={e => {
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -121,12 +121,9 @@ export function Item({children, label, style, onPress, ...rest}: ItemProps) {
       onFocus={onFocus}
       onBlur={onBlur}
       onPress={async e => {
-        if (e.defaultPrevented) {
-          await onPress(e)
-        } else {
-          control?.close(() => {
-            onPress(e)
-          })
+        await onPress(e)
+        if (!e.defaultPrevented) {
+          control?.close()
         }
       }}
       onPressIn={e => {
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -9,7 +9,12 @@ import {atoms as a, useTheme} from '#/alf'
 import {Button, ButtonText} from '#/components/Button'
 import * as Dialog from '#/components/Dialog'
 import {useInteractionState} from '#/components/hooks/useInteractionState'
-import {Context, ItemContext} from '#/components/Menu/context'
+import {
+  Context,
+  ItemContext,
+  useMenuContext,
+  useMenuItemContext,
+} from '#/components/Menu/context'
 import {
   ContextType,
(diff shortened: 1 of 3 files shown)


src/components/Menu/index.tsx
```diff
@@ -120,9 +120,12 @@ export function Item({children, label, style, onPress, ...rest}: ItemProps) {
...
       onPress={async e => {
-        await onPress(e)
-        if (!e.defaultPrevented) {
-          control?.close()
+        if (e.defaultPrevented) {
+          await onPress(e)
+        } else {
+          control?.close(() => {
+            onPress(e)
+          })
```
haileyok, line 128: @estrattonbailey would it be okay for us to remove this `e.defaultPrevented()` entirely for native? Are there cases where we want something to run _before_ we close the sheet? I assume the `defaultPrevented()` is only useful on web but maybe I'm wrong.

src/components/Menu/index.tsx
```diff
@@ -120,10 +120,9 @@ export function Item({children, label, style, onPress, ...rest}: ItemProps) {
       onFocus={onFocus}
       onBlur={onBlur}
       onPress={async e => {
-        await onPress(e)
-        if (!e.defaultPrevented) {
-          control?.close()
-        }
+        control?.close(() => {
```
gaearon, line 123: note this is a change in behavior — if control is undefined, `onPress` won't fire at all. is this desirable? can control be undefined?

estrattonbailey: Should never be undefined, wonder why we're using optional chaining here

surfdude29:
If an OTA update happens for 1.93, could this be included please? I've run into it quite a few times by tapping too long 🤦🏻‍♂️

src/components/Menu/index.tsx
roni-castro, line 123: @haileyok What do you think about creating a custom hook, `useMenuControlContext`, instead of using `React.useContext` directly? This way, we could ensure that the control context for the Menu is always defined and never undefined.

```js
export const useMenuControlContext = () => {
  const context = React.useContext(Context)

  if (!context) {
    throw new Error('useMenuControlContext must be used within a Context.Provider')
  }

  return context
}
```
Also, this would remove this `@ts-ignore`

Screenshot 2024-11-12 at 14 12 46


`control?` is no longer needed in line 123
Screenshot 2024-11-12 at 14 13 57

Notes: 
* It exists [this function](https://github.com/bluesky-social/social-app/blob/main/src/components/Menu/index.tsx#L28-L30) that is supposed to be this custom hook I'm mentioning, but there's a typo in there `Memo`, and it is not handling undefined value. 
* The Menu/index.web.tsx file also have this unused custom hook, and in there's a [duplication of the context hook ](https://github.com/bluesky-social/social-app/blob/main/src/components/Menu/index.web.tsx#L45-L47), that's why I think it's better to export it from the `context.tsx`, to avoid unnecessary duplication.
* This custom hook is suggested by Kent C Dodds in his [blog post](https://kentcdodds.com/blog/how-to-use-react-context-effectively#conclusion)

roni-castro: This occurs because the initial value of the control must be [null](https://github.com/bluesky-social/social-app/blob/main/src/components/Menu/context.tsx#L7), as it's unknown. Refer to [this discussion](https://github.com/bluesky-social/social-app/pull/5849#discussion_r1838494512) for more details.

haileyok: Oop, yea need to make sure this PR doesn't get lost down in my stack... @estrattonbailey any thoughts on the suggested approach here? Think you have the most context on what we need out of this.

surfdude29:
@haileyok I still run into this occasionally, would it be possible to get this in for 1.96 please? 🙏

src/components/Menu/index.tsx
estrattonbailey: Yeah let's just remove this. I did another pass and I don't see any use cases atm, and even if we want something like this we can achieve it in other ways.

src/components/Menu/index.tsx
estrattonbailey: Yep this is a good suggestion, knocked that out. Appreciate the thoughts!

estrattonbailey (approved):
Tested on all platforms, looking good, will merge after checks clear 👍

surfdude29:
Thanks so much @estrattonbailey, @haileyok and @roni-castro! :)

PR #5881, opened 2024-10-21
Edit mode on switch account dialog

mozzius:
Required for settings redesign

https://github.com/user-attachments/assets/cd5ddeb3-fac6-4ce0-bddb-5b8df080fc37


mary-ext:
It seems like it'd be prone to accidental taps, e.g. wanting to remove X account but ended up removing Y.

mozzius:
The tap target is no different than the normal list? Do you think there should be a confirm prompt?

mary-ext:
Yeah, though perhaps adding a confirmation might add too many taps? Considering how annoying it is to sign-in again it seems appropriate.

mozzius:
Closed in favour of https://github.com/bluesky-social/social-app/pull/6131

PR #5886, opened 2024-10-21
FAB double haptic experiment

mozzius:
Buzzes on press in and on press out, with a stronger buzz on long press

Curious how this feels

### DIFF
--- a/src/view/com/util/fab/FABInner.tsx
+++ b/src/view/com/util/fab/FABInner.tsx
@@ -42,14 +42,14 @@ export function FABInner({testID, icon, onPress, ...props}: FABProps) {
       ]}>
       <PressableScale
         testID={testID}
+        onPressIn={() => playHaptic('Light')}
         onPress={e => {
           playHaptic('Light')
-          setTimeout(
-            () => {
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Tested on device, feels great

src/view/com/util/fab/FABInner.tsx
```diff
@@ -42,14 +42,14 @@ export function FABInner({testID, icon, onPress, ...props}: FABProps) {
...
+        onPressIn={() => playHaptic('Light')}
         onPress={e => {
           playHaptic('Light')
-          setTimeout(
-            () => {
-              onPress?.(e)
-            },
-            isHapticsDisabled ? 0 : 75,
-          )
+          setTimeout(() => onPress?.(e), isHapticsDisabled ? 0 : 75)
```
haileyok, line 48: Just as a note, we probably want to get rid of this `setTimeout` here. I believe it was originally used for the composer FAB, in which case the timeout lets you see the animation of the fab before the sheet begins to present. In general though, we want to prefer fast response rather than seeing the animation imo

PR #5677, opened 2024-10-10
Link cards

estrattonbailey:
CleanShot 2024-10-10 at 13 51 37@2x
CleanShot 2024-10-10 at 13 51 43@2x
CleanShot 2024-10-10 at 13 51 51@2x
CleanShot 2024-10-10 at 13 52 12@2x
CleanShot 2024-10-10 at 13 54 45@2x
CleanShot 2024-10-10 at 13 57 02@2x
CleanShot 2024-10-10 at 13 58 21@2x
CleanShot 2024-10-10 at 13 58 34@2x
CleanShot 2024-10-10 at 13 58 52@2x
CleanShot 2024-10-10 at 13 59 04@2x

PR #5887, opened 2024-10-21
Put ExternalEmbedRemoveBtn on top

estrattonbailey:
Evidently wasn't high enough in certain contexts ([video](https://github.com/bluesky-social/social-app/pull/5677#issuecomment-2423715433)). This seem to work not for all embed types.

### DIFF
--- a/src/view/com/composer/ExternalEmbedRemoveBtn.tsx
+++ b/src/view/com/composer/ExternalEmbedRemoveBtn.tsx
@@ -11,7 +11,7 @@ export function ExternalEmbedRemoveBtn({onRemove}: {onRemove: () => void}) {
   const {_} = useLingui()
 
   return (
-    <View style={[a.absolute, a.pt_sm, a.pr_sm, {top: 0, right: 0}]}>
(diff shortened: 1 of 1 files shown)


src/view/com/composer/ExternalEmbedRemoveBtn.tsx
```diff
@@ -11,7 +11,7 @@ export function ExternalEmbedRemoveBtn({onRemove}: {onRemove: () => void}) {
   const {_} = useLingui()
 
   return (
-    <View style={[a.absolute, a.pt_sm, a.pr_sm, {top: 0, right: 0}]}>
+    <View style={[a.absolute, a.pt_sm, a.pr_sm, a.z_50, {top: 0, right: 0}]}>
```
gaearon, line 14: this seems like it would be better fixed at the usage site, no? we can't guess the necessary z-order here, but we can be more certain in the place that defines the actual ordering (and it's possible we can also fix this by changing the DOM order itself)

gaearon (approved):

PR #5891, opened 2024-10-21
Fix starter pack text overflow

mozzius:
just a silly flexbox issue

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Simulator Screenshot - iPhone 16 - 2024-10-21 at 19 46 58</td>
      <td>Simulator Screenshot - iPhone 16 - 2024-10-21 at 19 46 47</td>
    </tr>
  </tbody>
</table>


haileyok (approved):
Just two small things we may want to hit:

- `numberOfLines` set to `2` for the title?
- `numberOfLines` set to `1` for the `Starter pack by...`

Think that follows what we do in other places.

PR #5893, opened 2024-10-21
Border on drawer

mozzius:


haileyok:
Does this make the already existing border problem on Android any worse?

mozzius:
good question lemme check (it shouldn't)

mozzius:
I can't see it on the sim. it might make the problem a little more obvious, but shouldn't make it worse

haileyok (approved):

PR #5898, opened 2024-10-22
React compiler beta and reenable rule

mozzius:
- update to `@beta`
- use `react-compiler-runtime`
- enable the eslint rule (set to warn)

A bunch of stuff breaks the rules (in particular Reanimated) so we'll have to use `warn` for now


poteto:
nice! you can also remove `lib/react-compiler-runtime` now and instead install `react-compiler-runtime@beta`, and replace `runtimeMOdule` with `target: '18'` to https://github.com/bluesky-social/social-app/blob/main/babel.config.js#L23

mozzius:
Thanks @poteto, much appreciated! 🙏

gaearon (approved):

gaearon:
let's try it

PR #5901, opened 2024-10-22
Animate search input

mozzius:
Layout animation on search input. also hides the drawer menu button when focused, to give more space to the input

https://github.com/user-attachments/assets/10e825bb-6da6-440a-9a5c-df1e051e3421


haileyok (changes requested):
Couple of things:

- Need to disable this when reduced motion is enabled
- On web, we can probably just not even have this animation at all. If we _do_ want the animation though, will have to fix it. Currently it slides in from the sides of the screen, not the bounds of the view. (Strong vibe from me is let's just not animate it on the web tbh)

mozzius:
yep, avoid layout animations on web other than fade

PR #5902, opened 2024-10-22
Add Hungarian localisation

smileyhead:
I have 8 years of localisation experience, but this is the first time I've tackled a social network, so feedback is very welcome.

I also have some questions/notes:
- Is there a way to get notified of any changes before build-time? I would like to avoid the localisation always being one version behind, with all the new text left untranslated, if possible.
- I have removed the special plural formatting, as Hungarian doesn't use plurals when counting. I hope this is acceptable.
- I have translated all ‘# minutes,’ ‘# hours,’ etc. strings as ‘# minutes ago,’ ‘# hours ago,‘ etc., since these alternate forms are achieved with suffixes in Hungarian and it would be impossible to separate them. Are they used in any other way? If so, would it be possible to make unique strings for other cases?
- Will there be a possibility to translate Bsky-associated content, like the name/description and (most importantly) categories in the Bluesky Moderation Service, as well as the names/descriptions of first-party feeds?

Thank you in advance.


auroursa:
It works well for me.



- It seems that after removing the plural, like time and others work properly.

- You can safely translate terms like “hours” to “hours ago” since, as far as I know, there aren’t other usages. I did the same in Chinese translations as well.

- Update the translation strings by running `yarn intl:extract --locale hu` each time manually, then you can get translation updates instantly. By the way, usually, there’s no need to translate strings starting with #~ as they have already been deprecated.

smileyhead:
Thank you very much for your feedback and your screenshot – I've made some small tweaks thanks to the latter.

Where and when exactly should I run the `yarn intl:extract --locale hu` command, if I may ask?

smileyhead:
Hmm, I've just looked closer at your screenshot and noticed that the ‘Liked by # users’ string did actually break. I wonder how to fix it…

image

auroursa:
> Thank you very much for your feedback and your screenshot – I've made some small tweaks thanks to the latter.
> 
> Where and when exactly should I run the `yarn intl:extract --locale hu` command, if I may ask?

First need to set up a local build environment, including tools like Node.js (following the [build instructions](https://github.com/bluesky-social/social-app/blob/main/docs/build.md)).

Typically, if you only need to test translations on the web, just need Node.js and Python, and make sure to add them to your system’s PATH.

After that, run `yarn install` and wait for a bit—long enough to enjoy a cup of coffee. Once the local setup is ready, execute `yarn intl:extract --locale hu`, and you’ll see the output like this:



It will update the messages.po file, and then you can use an appropriate tool to fill in any missing translation strings.

After completing the translation of the messages.po file, you can run `yarn intl:compile` to generate the correct target translation files. Then, execute `yarn web` to see the results.

That's it. Good luck and wishing you all the best.

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7651 @@
...
+msgid "{0, plural, one {Like (# like)} other {Like (# likes)}}"
+msgstr "Tetszik (# kedvelés)"
+
+#: src/view/com/post-thread/PostThreadItem.tsx:434
+msgid "{0, plural, one {like} other {likes}}"
+msgstr "kedvelés"
+
+#: src/components/FeedCard.tsx:213 src/view/com/feeds/FeedSourceCard.tsx:300
+msgid "{0, plural, one {Liked by # user} other {Liked by # users}}"
+msgstr "# felhasználó kedveli"
```
auroursa:
```suggestion
msgstr "{0, plural, other {# felhasználó kedveli}}"
```

auroursa: Maybe expressing the plural in this way will work, it works well in the Chinese translation.

smileyhead:
Thank you very much.

In the meantime, I believe to have deciphered the proper formatting of the plural strings and have applied a possible fix.

src/locale/locales/hu/messages.po
smileyhead: I have pushed a commit right before your answer with a different possible solution. Would you might checking if it works, please?

smileyhead:
Thank you very much!

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+msgstr "{0} kedvenc hírfolyamai és személyei – csatlakozz!"
+
+#: src/screens/StarterPack/Wizard/StepDetails.tsx:47
+msgid "{0}'s starter pack"
+msgstr "{0} kezdőcsomagja"
+
+#. How many days have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:158
+msgid "{0}d"
+msgstr "{0} n"
```
surfdude29, line 140:
```suggestion
msgstr "{0} n"
```

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+
+#. How many hours have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:148
+msgid "{0}h"
+msgstr "{0} ó"
```
surfdude29, line 145:
```suggestion
msgstr "{0} ó"
```

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+
+#. How many minutes have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:138
+msgid "{0}m"
+msgstr "{0} p"
```
surfdude29, line 150:
```suggestion
msgstr "{0} p"
```

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+
+#. How many months have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:169
+msgid "{0}mo"
+msgstr "{0} hó"
```
surfdude29, line 155:
```suggestion
msgstr "{0} hó"
```

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+
+#. How many seconds have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:128
+msgid "{0}s"
+msgstr "{0} mp"
```
surfdude29, line 160:
```suggestion
msgstr "{0} mp"
```

surfdude29 (commented):
If it's desirable to have a space in the translations of these narrow forms, I think it would be best to use a non-breaking space to make sure that the two elements are shown together :)

smileyhead:
Honestly, adding a space was just a stylistic choice at the time, but now I've checked Twitter and Facebook and it turns out they both do use spaces in Hungarian as well. The addition of non-breaking spaces is a good idea, though. Thank you for that.

surfdude29:
This looks like a great effort, nice work! 👌

I see that @auroursa has already been very helpful in answering questions and giving feedback :)

I just wanted to note a few other things:

> * Will there be a possibility to translate Bsky-associated content, like the name/description and (most importantly) categories in the Bluesky Moderation Service...

It looks like you've translated the four "global" content labels like Adult Content, Sexually Suggestive, etc, so if you're asking if you can translate all the other labels on the [profile page for the Bluesky Moderation Service](https://bsky.app/profile/moderation.bsky.app): unfortunately, that isn't possible yet.

> ...as well as the names/descriptions of first-party feeds?

Disappointing answer here too, I'm afraid, as **Discover** and **Following** are currently hardcoded in English.

And finally, one important thing to mention is that the devs usually ask to have at least one native speaker review a localization before merging it. Would it be possible to ask another Hungarian speaker to do that review?

Once they've done their review, my completely unofficial opinion as a fellow translator is that this PR would look in a good position to be merged 🚀

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+msgstr "{0} hó"
+
+#. How many seconds have passed, displayed in a narrow form
+#: src/lib/hooks/useTimeAgo.ts:128
+msgid "{0}s"
+msgstr "{0} mp"
+
+#: src/components/LabelingServiceCard/index.tsx:96
+msgid "{count, plural, one {Liked by # user} other {Liked by # users}}"
+msgstr "{0} felhasználó kedveli"
```
auroursa, line 164:
```suggestion
msgstr "{count} felhasználó kedveli"
```

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7803 @@
...
+
+#: src/components/dms/dialogs/SearchablePeopleList.tsx:384
+msgid "{handle} can't be messaged"
+msgstr "{handle} jelenleg nem fogad üzeneteket"
+
+#: src/screens/Profile/Header/ProfileHeaderLabeler.tsx:284
+#: src/screens/Profile/Header/ProfileHeaderLabeler.tsx:297
+#: src/view/screens/ProfileFeed.tsx:591
+msgid "{likeCount, plural, one {Liked by # user} other {Liked by # users}}"
+msgstr "{0} felhasználó kedveli"
```
auroursa, line 192:
```suggestion
msgstr "{likeCount} felhasználó kedveli"
```

smileyhead:
Thank you so much. I've asked my friend, @SonoSooS to help with proofreading – he should report back soon™.

auroursa (commented):
I found that specifying it as {0} does not display the count correctly.

auroursa:
Thanks @surfdude29 for the additional information. I feel frustrated about the inability to translate "Following," "Discover," and the Bluesky Moderation Service list, as this has been an issue affecting the Chinese translation as well.

But hey yes, I would say that technically this PR looks ready to merge🤗

Look forward to other Hungarian native speakers helping to refine the translation. You're doing a great job!

src/locale/locales/hu/messages.po
```diff
@@ -0,0 +1,7805 @@
...
+
+#: src/state/shell/progress-guide.tsx:166
+msgid "Task complete - 10 likes!"
+msgstr "Feladat teljesítve – 10 kedvelés!"
+
+#: src/components/ProgressGuide/List.tsx:49
+msgid "Teach our algorithm what you like"
+msgstr ""
+"Tanítsd meg az algoritmusnak,\n"
+"hogy mit szeretnél látni!"
```
surfdude29, line 6132:
```suggestion
msgstr "Tanítsd meg az algoritmusnak, hogy mit szeretnél látni!"
```
I think there was a bit of a glitch here

smileyhead: This was intentional. If I don't specify a line break there, the sidebar doesn't do it by itself and expands to the right, adding a horizontal scroll bar.

SonoSooS:
We went through it with @smileyhead, and correct some things that - while correct - felt really unnatural, like short "ago" units for example. Those may work in other languages, but not here. Otherwise yeah, he did a fantastic job, as usual.

However, there are some "dialect" issues that we couldn't agree on, so I'll list them here, just in case stylistic "overrides" are preferred over language correctness.
* pagination order, forwards "2/3" ("2nd page out of 3" (basically the same word order as in Hungarian)) vs. backwards "3/2" ("2nd out of 3 pages" (the spoken order is with the 3 first, then 2, so this one rolls off the tongue)). This one is very much so a mixed bag. Some people prefer "3/2", and others (like myself) utterly despise it, and prefer the "2/3" one. He even asked linguistics, and they said, that there is no rule for this. Media uses the backwards format, but most (but not all) people that I know around me use the forwards format. A lot of people will be displeased, no matter which one gets hardcoded.
* couldn't agree on whether to use "alsókötőjel" or "alulvonás" for the underscore (`_`). I have never heard of anyone using "alsókötőjel", and I've been from barely inhabited remote villages with "dialects" almost impossible to understand, to big cities, and I have never in my life heard the word "alsókötőjel", everyone I have ever met calls it an "alulvonás". However, according to Wikipedia, it is officially called "alsókötőjel", but in everyday language we say "alulvonás" for sure.
* this is not a problem with his translation, but there is no thousands separator for big numbers (for example 5-digit follower count). The official thousands separator here is the dot (`.`), but basically every website uses space (` `) as a thousands separator instead. Personally I prefer the comma (`,`) for readability though, but then there is a possibility of people confusing it for a decimal separator, so it sadly can not be used.

We also found a bunch of usability issues, but those have nothing to do with this PR, I'll report those separately.

src/locale/locales/hu/messages.po
surfdude29: Oh I see, I didn’t realise it was possible to add a line break like that

smileyhead:
> The official thousands separator here is the dot (.), but basically every website uses space ( ) as a thousands separator instead.

Correction: The main thousands separator is a space, and the second option is a dot (and we use a decimal comma).
So, yeah, a long number with digits would look like this: `12 345`. But for some reason, the site displays them as `12345`.

src/locale/locales/hu/messages.po
smileyhead: We've tested it and it works. 👍

surfdude29: That’s what counts 👌

pfrazee (approved):
Wow, thank you so much everybody for helping out so much, and thank you @smileyhead for this massive contribution!

PR #5908, opened 2024-10-23
Clean up English messages.po

tkusano:
Cleaned up en/messages.po to reduce size of messages.po for new language translators

- Removed old entries
- Removed one "translated" entry
- Added new messages since 1.92.1


surfdude29:
This will certainly be helpful, especially for new translators to help them avoid wasting their effort translating obsolete strings, thanks @tkusano! 🙏

pfrazee (approved):
Thanks!!

PR #5919, opened 2024-10-24
Change touch device detection method

### LINKED ISSUE #5856: Hover cards don't work on laptop with touch screen
### Steps to Reproduce

1. Use bluesky on a laptop with a touch screen
2. Hover over a profile name with the mouse

I did find the reason for this behavior:
https://github.com/bluesky-social/social-app/blob/72270fd5053f1132a7966197cd9dce17b5672d57/src/lib/browser.ts#L6-L7
On my computer `navigator.maxTouchPoints` is 10 because of the touch screen, even though I don't actually use it. This is true for both Chrome and Firefox.

### Attachments

_No response_

Web (Desktop)

### Device Info

_No response_

Build version: 1.92.1; Bundle info: 89cff31 (prod); Bundle date: 24101819; Platform: web

### Additional Information

_No response_

### LINKED ISSUE #5859: Desktop webapp on Windows device with a touchscreen does not send message when enter buttton is pressed
### Steps to Reproduce

1) Click "Chat" button on the left of the screen
2) Click a conversation to open it or begin a new message
3) Type out a message
4) Enter key sends the message in MacOS or on a Windows 10 desktop without a touchscreen / Enter does not send the composed message on Windows device with a touchscreen

### Attachments

[bksky_webapp_chat_enter-button_macos.webm](https://github.com/user-attachments/assets/d18bfee8-a8ed-426b-bf76-f5ff8becdb7d)

[bsky_webapp_chat_enter-button_win11.webm](https://github.com/user-attachments/assets/c6ec62ca-8eec-4640-b250-13d03ee250e9)

Web (Desktop), Microsoft Windows

### Device Info

Thinkpad Yoga x390 running Windows 11 Pro Version 23H2 [Firefox 131.0.3 (64-bit) and Microsoft Edge Version 130.0.2849.46 (Official build) (64-bit)]
M1 Macbook Air running MacOS 14.6.1 [Firefox 131.0.3 (aarch64) and Safari Version 17.6 (19618.3.11.11.5)]

Build version: 1.92.1; Bundle info: 89cff31 (prod); Bundle date: 24101819; Platform: web

### Additional Information

New to Github, sorry about all the edits

brw:
Using `navigator.maxTouchPoints` to check whether a device is a "touch device" also matches laptops with touch screen support. Checking for [`pointer: coarse`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer#coarse) instead should only detect whether the primary input mechanism is touch (e.g. smartphones and tablets but not laptops with touch screens).

It's not perfect though. No combination of `(any-)pointer` and `(any-)hover` seems to work for detecting when I plug a physical mouse into my Android phone for example. Might have to rely on the `touchstart` event for that.

Checking `pointer: coarse` is better than the current method at least as it does properly distinguish between mobile/tablet and pc, which the current method doesn't.

Resolves #5856 (hover cards and video volume slider not showing up), resolves #5859 (enter not sending chat messages)

### DIFF
--- a/src/lib/browser.ts
+++ b/src/lib/browser.ts
@@ -3,7 +3,6 @@ export const isSafari = /^((?!chrome|android).)*safari/i.test(
   navigator.userAgent,
 )
 export const isFirefox = /firefox|fxios/i.test(navigator.userAgent)
-export const isTouchDevice =
-  'ontouchstart' in window || navigator.maxTouchPoints > 1
+export const isTouchDevice = window.matchMedia('(pointer: coarse)').matches
 export const isAndroidWeb =
   /android/i.test(navigator.userAgent) && isTouchDevice


haileyok:
Thank you for pointing out this problem, was not aware!

To clarify, does the current implementation work whenever you plug in a mouse to your Android, or does neither solution work for that?

brw:
The current implementation doesn't work for that either no

brw:
Another thought I had was to maybe filter for `PointerEvent.nativeEvent.pointType === 'mouse'`. That seems to work for detecting whether the currently used pointing device is a physical mouse on both.

But in my testing it seems that even when using a physical mouse on Android the `mouse/pointerenter` and `mouse/pointerleave` events don't fire unless you actually click on an element, which already doesn't make the current video player very usable with a physical mouse on Android. I think Android (13) and/or mobile browsers might just not support mouse hover events at all (which would make sense with touch being the main input and physical mouse usage on Android being a very niche thing). Either that or I'm doing something wrong.

brw:
Small update, I've come across a few people who also encounter this issue because they have a display drawing tablet connected to their pc, which is also recognized as a touch screen. I imagine there's a sizable group of artists on the platform also not seeing a volume slider or hover cards because of this.

gaearon:
If you read through https://css-tricks.com/interaction-media-features-and-their-potential-for-incorrect-assumptions/, does the approach still hold up? (I haven't)

brw:
Thanks for the article! Great read. Whether it holds up depends on how you look at it I suppose. Even with all the pitfalls I still think that checking for `pointer: coarse` is likely good enough to at least determine the initial choice on whether or not to show hover elements (as that's primarily what `isTouchDevice` is currently being used for). The article also agrees.

`hover` and `any-pointer`/`any-hover` give far too many false positives from what I've seen. `pointer` is at least fairly consistent (phone/tablet/Surface in tablet mode: `coarse`, desktop/laptop/Surface with type cover: `fine`). Perhaps there's some golden combination of these that I'm not seeing, but I think that a simple `pointer: coarse` is probably the most reliable as far as media queries go.

I did already look at https://patrickhlauke.github.io/touch/pointer-hover-any-pointer-any-hover/results too, which is also mentioned in the article, to help determine what to use here (though it's slightly out of date).

It would probably be a good idea to supplement this using input events like https://github.com/ten1seven/what-input to determine the currently used input method aside from the primary one (package mentioned in the article, no idea if others exist for this, I haven't been able to find any). But that's quite a bit more involved of course. And you still run into issues like the fact that hover/enter/over events aren't really supported on Android web for example (in my testing at least, they just fire on click, even with a physical mouse).

It might also be a good idea to just give the user the option to switch manually as well (which the article also suggests). Especially when taking assistive technologies and browser bugs/differences into account. There doesn't seem to be a perfectly accurate way of determining this stuff.

Switching to `pointer: coarse` was mostly meant as a stepping stone solution because the current method has even more issues. I'm far from an expert here so I don't feel particularly confident enough to implement something more thorough right now (nor do I have enough devices to test), but if you have any suggestions or input I'd love to hear it.

brw:
Coming back to this after a while, is it possible to have someone review this? This bug still exists today and I've tried my best to outline above why this solution isn't perfect but already a big improvement. It would at least allow people with touch-enabled laptops and art tablets to view the volume slider.

I'm currently still using my [userscript](https://greasyfork.org/en/scripts/518378-spoof-bluesky-mobile-detection) that overrides `navigator.maxTouchPoints` to get around this issue.

sstoneb:
I don't really know what the norms/protocols are for participating in these threads as I am not a developer, but I would really like to see this issue resolved. It's a pretty big deal! As brw has already mentioned it's linked to multiple misbehaviors that presumably affect EVERY user of a computer with a touchscreen.

I regularly make use of four different windows machines, and three of the four have touchscreens as a secondary input method. On all three of them I cannot control video volumes or see profile cards on mouse hover. Here is another issue report for the same thing: [#6961 Profile hover cards are disabled on touchscreen Windows laptops](https://github.com/bluesky-social/social-app/issues/6961)

For now I'm going to install a browser extension so I can use brw's userscript but this resolves things only for me and not for thousands of other users, most of whom won't even know why the site is behaving strangely.

gaearon:
@mozzius

Issue #5871, opened 2024-10-20
Translations not working properly on the Welcome New Member Modal in Profiles


NaraDevStuff:
### Steps to Reproduce

1. Set system Settings to something other than english
2. Go to own newly created profile
3. Click on the half sun icon 
4. Look at Modal

### Attachments

Comparison: English vs German
![Screenshot 2024-10-20 at 19 28 22](https://github.com/user-attachments/assets/f980ecde-b49e-4fcf-89e6-9a90ddc332b2)
![IMG_5180](https://github.com/user-attachments/assets/6bc3f898-22d1-4bbd-b0b9-09855a474ca6)

### What platform(s) does this occur on?

iOS, Android, Web (Desktop), Web (Mobile)

### What version of the app are you using?

Build version: 1.92.1; Bundle info: 89cff31 (prod); Bundle date: 24101819; Platform: web

mozzius:
Fixed by https://github.com/bluesky-social/social-app/pull/7949

PR #5922, opened 2024-10-24
Update German translation

nandolawson:
The German translation is not finished yet. I would recommend giving Crowdin a chance to handle translations.


auroursa:
This helps to close #5871.

pfrazee (approved):
Thanks!!

PR #5924, opened 2024-10-24
Use ALF verify email dialog for reminder

mozzius:
# Stacked on #8106 

We already have a refreshed verify email dialog, but we weren't using it for the daily reminder. This fixes that.

- Uses the new dialog
- Adds an extra step to the dialog so it can replace the old one

<table>
  <thead>
    <tr>
      <th>iOS</th>
      <th>Web</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

Use an account with an unverified email. I added a button to the bottom of settings to unsnooze the reminder

### DIFF
--- a/src/components/dialogs/VerifyEmailDialog.tsx
+++ b/src/components/dialogs/VerifyEmailDialog.tsx
@@ -171,7 +171,6 @@ export function Inner({
               a.justify_center,
               {height: 150},
               t.atoms.bg_contrast_100,
-              web(a.mt_2xl),
             ]}>
             <EnvelopeIcon width={64} fill="white" />
           </View>
@@ -356,7 +355,6 @@ export function Inner({
           ) : null}
         </View>
       </View>
-      <Dialog.Close />
     </Dialog.ScrollableInner>
   )
 }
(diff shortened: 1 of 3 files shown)


src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -235,7 +260,32 @@ export function Inner({
...
+                variant="solid"
+                color="primary"
+                size="large"
+                onPress={() => setCurrentStep('StepOne')}>
+                <ButtonText>
+                  <Trans>Get started</Trans>
+                </ButtonText>
+              </Button>
+              <Button
+                label={_(msg`Maybe layer`)}
```
surfdude29, line 276:
```suggestion
                label={_(msg`Maybe later`)}
```
Typo in label here

src/components/dialogs/VerifyEmailDialog.tsx
```diff
@@ -235,7 +260,32 @@ export function Inner({
...
+                accessibilityHint={_(msg`Snooze reminder`)}
```
surfdude29, line 277:
```suggestion
                accessibilityHint={_(msg`Snoozes reminder`)}
```
I think `Snoozes` works better because IIRC `accessibilityHint` describes the result of the action? Not completely sure on that though tbh

estrattonbailey (approved):

PR #5931, opened 2024-10-24
Remove waterfalls from posting

gaearon:
Move unrelated asynchronous logic into separate functions and kick them off concurrently. Await results at the end.

## Before

Posting takes 10 seconds on a slow connection.

before

## After

Posting takes 6 seconds on a slow connection.

after

## Test Plan

Try posting plain post, post with facets, post with media, a reply, combinations of these.


estrattonbailey (approved):
Nice one!

PR #5942, opened 2024-10-25
Conditionally render labels button when media exists in post

haileyok:
## Why

We only want to show the "Labels" button right now whenever there is media in the post that a label can be added to.

## Test Plan

The labels button should show up when:

- A link embed is included in the post
- A video is added to the post
- One or more images are added to the post

The labels should be reset whenever you remove all of the media from your post as well. To test, add an image to the post, modify the labels, remove the images, then add another image. The labels should have reset.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -511,9 +511,6 @@ export const ComposePost = ({
 
           <ComposerPills
             isReply={!!replyTo}
-            containsMedia={
-              images.length > 0 || videoState.status !== 'idle' || !!extLink
-            }
             draft={draft}
             dispatch={dispatch}
             bottomBarAnimatedStyle={bottomBarAnimatedStyle}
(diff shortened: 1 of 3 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -896,13 +901,15 @@ function ComposerPills({
...
           />
         )}
-        <LabelsBtn
-          labels={draft.labels}
-          onChange={nextLabels => {
-            dispatch({type: 'update_labels', labels: nextLabels})
-          }}
-          hasMedia={hasMedia || hasLink}
-        />
+        {containsMedia ? (
```
gaearon, line 904: i think you can use `hasMedia || hasLink` condition that already exists

src/view/com/composer/Composer.tsx
```diff
@@ -896,13 +901,15 @@ function ComposerPills({
...
+          <LabelsBtn
+            labels={draft.labels}
+            onChange={nextLabels => {
+              dispatch({type: 'update_labels', labels: nextLabels})
+            }}
+            hasMedia={hasMedia || hasLink}
```
gaearon, line 910: i think this prop can be removed now? so then it would make sense to go into `LabelsBtn` and make all branches that depend on `hasMedia` truthy

gaearon (approved):
lg but a few nits

PR #5945, opened 2024-10-25
Refactor composer state for threads

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5953

---

This prepares the composer state structure to handle multiple posts in a thread. Previously we just had the `draft`. Now we have a `thread` made of `posts`.

Most of the reducer logic is now in `postReducer` which updates `PostState` in response to `PostAction`. The `postReducer` is being called for the current post from the outer `composerReducer`.

We're not actually taking advantage of this in the UI yet. The UI is the same and it only reads the current post's state (which is always the first post because we don't have any logic to add/remove posts yet).

Also, the publishing code only looks at the first post at the moment. That will be changed later.

## Test Plan

Same as usual, verify different composer features work.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -114,11 +114,13 @@ import {Text as NewText} from '#/components/Typography'
 import {BottomSheetPortalProvider} from '../../../../modules/bottom-sheet'
 import {
   ComposerAction,
-  ComposerDraft,
   composerReducer,
   createComposerState,
   EmbedDraft,
   MAX_IMAGES,
+  PostAction,
+  PostDraft,
+  ThreadDraft,
 } from './state/composer'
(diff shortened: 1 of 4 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -511,8 +523,9 @@ export const ComposePost = ({
 
           <ComposerPills
             isReply={!!replyTo}
-            draft={draft}
-            dispatch={dispatch}
+            post={draft}
+            thread={composerState.thread}
+            dispatch={composerDispatch}
```
gaearon, line 528: In most cases we pass the post dispatch but here it needs the broader one to set gates. I'm just calling it `dispatch` everywhere though. Might be a bit confusing but types clarify which one it needs.

haileyok (approved):
lgtm, tested range of composer features without issues

PR #5953, opened 2024-10-26
Disable Post button when empty

gaearon:
This in preparation for the threaded composer. It doesn't quite make sense to treat this as an error — we should just prevent you from posting, same as we do in other invalid cases (like too long post or missing ALT).

In the second commit, I'm also replacing ad-hoc grey disabled styling with our normal disabled button. We've already been using the disabled button for the case when you _pressed_ post so let's use it here too.

## Test Plan

Observe the initial state.

Screenshot 2024-10-26 at 20 45 16

Verify that typing makes the button enabled.

Screenshot 2024-10-26 at 20 45 20

Verify existing flows (adding image, filling alt, adding video, posting before it's uploaded) work like before.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -684,7 +684,7 @@ function ComposerTopBar({
               <ActivityIndicator />
             </View>
           </>
-        ) : canPost ? (
+        ) : (
           <Button
             testID="composerPublishBtn"
             label={isReply ? 'Publish reply' : 'Publish post'}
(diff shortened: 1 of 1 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -293,13 +293,26 @@ export const ComposePost = ({
...
+    !isEmptyPost
+
   const onPressPublish = React.useCallback(
     async (finishedUploading: boolean) => {
-      if (isPublishing || graphemeLength > MAX_GRAPHEME_LENGTH) {
+      if (isPublishing) {
         return
       }
 
-      if (isAltTextRequiredAndMissing) {
```
gaearon, line 302: both removed conditions are a part of `!canPost` now

mozzius:
Found a case where it's not disabled while empty - video errors

Screenshot 2024-10-28 at 20 34 16

gaearon:
I think we should probably disable posting if there's an error in general? Empty or not. Clear/ack the error, then post.

gaearon:
Pushed.

haileyok (approved):
lgtm

PR #5954, opened 2024-10-26
Remove indirection when rendering composer state

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5953 and https://github.com/bluesky-social/social-app/pull/5945

---

Continues refactoring to prepare for threads.

This is a refactor to remove some intermediate variables that no longer serve us well. We want to read from `draft` directly so that we can later adjust the conditions to deal with multiple posts.

The last commit moves a grapheme length calculation to the reducer because we won't be able to memoize it in the component anymore (since it will be per-post).

## Test Plan

Verify the discard prompt shows up when the post is non-empty (note that just a quote is considered empty).

Verify the grapheme calculations works like before, including with long URLs.

Verify missing ALT detection still works for images and for GIFs.

Verify Post button remains disabled on completely empty posts.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -59,7 +59,6 @@ import {usePalette} from '#/lib/hooks/usePalette'
 import {useWebMediaQueries} from '#/lib/hooks/useWebMediaQueries'
 import {logEvent} from '#/lib/statsig/statsig'
 import {cleanError} from '#/lib/strings/errors'
-import {shortenLinks} from '#/lib/strings/rich-text-manip'
 import {colors, s} from '#/lib/styles'
 import {logger} from '#/logger'
 import {isAndroid, isIOS, isNative, isWeb} from '#/platform/detection'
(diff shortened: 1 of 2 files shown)


haileyok:
this isn't a new issue, but im wondering why the "require alt text" warning doesn't let you post when adding a gif when there is already alt text set. shouldn't the alt text be getting automatically applied based on the default for those? wouldn't expect the warning to appear in those cases.

haileyok:
one other note, and probably not something that is new but didn't check.

- Add a video and press "Post" before the upload completes
- Reply button becomes disabled pending the completion of upload
- The video fails to process/upload. User gets an error
- User adjusts post content
  - At this point, the reply button never becomes "undisabled" _until a new video gets attached_
  - As soon as the new video uploads, the post automatically gets posted. There is no warning or "resetting" of the state when the new video gets added to the post

I think a failure should be resetting that "post when finished" state.

haileyok (approved):
lgtm with the two notes above, neither of which i believe are regressions but should get addressed somewhere in this rework

gaearon:
>this isn't a new issue, but im wondering why the "require alt text" warning doesn't let you post when adding a gif when there is already alt text set. shouldn't the alt text be getting automatically applied based on the default for those? wouldn't expect the warning to appear in those cases.


re: this, we settled on forcing you to open the dialog once to proofread the alt (even though it gets saved from that point on). since sometimes it's very low quality. otherwise the setting would just always pass things through which defeats the purpose of the setting.

gaearon:
>I think a failure should be resetting that "post when finished" state. (For a surefire way to get an upload to finish, can send you a video on Slack that will cause a failure)

yup, confirmed the refactor in https://github.com/bluesky-social/social-app/pull/5957 fixed that

PR #5957, opened 2024-10-26
Check video state for the entire thread

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5953, https://github.com/bluesky-social/social-app/pull/5945, https://github.com/bluesky-social/social-app/pull/5954, and https://github.com/bluesky-social/social-app/pull/5955

---

Same as https://github.com/bluesky-social/social-app/pull/5955 but for videos.

## Test Plan

Go through the usual flows with a single video. Upload it, change dimensions, post while uploading too, etc.

Also verify error path works too. And clearing errors.

PR #5960, opened 2024-10-27
Implement posting threads

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5953, https://github.com/bluesky-social/social-app/pull/5945, https://github.com/bluesky-social/social-app/pull/5954, https://github.com/bluesky-social/social-app/pull/5955, and https://github.com/bluesky-social/social-app/pull/5957

---

Implements posting a thread based on the strategy in #4163 by @mary-ext.

I've done some minor tweaks.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/5960/files?w=1)

## Test Plan

Verify posting one post has not regressed.

Verify postgates and threadgates work.

Multiple posts aren't wired up in the UI yet so this will be testable in a follow-up.


src/lib/api/index.ts
```diff
@@ -53,102 +56,118 @@ export async function post(
...
+    if (draft.labels.length) {
+      labels = {
+        $type: 'com.atproto.label.defs#selfLabels',
+        values: draft.labels.map(val => ({val})),
+      }
+    }
+
+    // The sorting behavior for multiple posts sharing the same createdAt time is
+    // undefined, so what we'll do here is increment the time by 1 for every post
+    now.setMilliseconds(now.getMilliseconds() + 1)
```
gaearon, line 104: @mary-ext i changed this from your original logic to increment from the actual posting timestamp — not that it matters but would avoid threads always having fake milliseconds. does this look good to you? mdn says they don't overflow and would be added up correctly.

src/lib/api/index.ts
```diff
@@ -53,102 +56,118 @@ export async function post(
...
-      rkey: rkey,
-      value: record,
-    })
+    // Prepare a ref to the current post for the next post in the thread.
+    const ref = {
+      cid: await computeCid(record),
+      uri,
+    }
+    replyPromise = {
+      root: reply?.root ?? ref,
```
gaearon, line 168: pretty subtle: if we're replying inside a thread, the root is the post we're replying to

src/lib/api/index.ts
mary-ext: Seems good, would give some variance into users' post time, because that sorting bug can end up affecting post visibility as well last I checked.

if two users were to post with the exact same timestamp then one of them would end up never show in the timeline.

haileyok (approved):
code lgtm, tested creating posts and such on ios which worked well for replies, gates, and media.

gaearon:
accidentally merged into wrong base, resubmitted as https://github.com/bluesky-social/social-app/pull/6049

PR #5962, opened 2024-10-27
Thread composer UI

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5953, https://github.com/bluesky-social/social-app/pull/5945, https://github.com/bluesky-social/social-app/pull/5954, https://github.com/bluesky-social/social-app/pull/5955, https://github.com/bluesky-social/social-app/pull/5957, https://github.com/bluesky-social/social-app/pull/5960, and #5976

---

This isn't pretty or anything. But it kind of works.

This is the culmination of porting pieces from https://github.com/bluesky-social/social-app/pull/4163 by @mary-ext.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/5962/files?w=1)

- [x] It works
- [x] Inserting posts in the middle
- [x] Don't allow creating >1 consecutive empty post
- [x] Removing posts from the thread
- [x] Collapse them in UI better, remove unnecessary gaps
- [x] Keyboard nav / a11y
- [x] Verify perf / re-renders (editing text shouldn't re-render other posts)
- [x] Scroll into view when adding to the end

https://github.com/user-attachments/assets/ec7e851c-ec7d-40c6-8c4c-4ee7ca76ddb5

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -322,131 +322,127 @@ export const ComposePost = ({
         ),
     )
 
-  const onPressPublish = React.useCallback(
-    async (finishedUploading: boolean) => {
-      if (isPublishing) {
-        return
-      }
+  const onPressPublish = React.useCallback(async () => {
+    if (isPublishing) {
+      return
+    }
 
-      if (!canPost) {
-        return
-      }
+    if (!canPost) {
+      return
+    }
 
-      if (
-        !finishedUploading &&
-        thread.posts.some(
-          post =>
-            post.embed.media?.type === 'video' &&
-            post.embed.media.video.asset &&
-            post.embed.media.video.status !== 'done',
-        )
-      ) {
-        setPublishOnUpload(true)
-        return
-      }
-
-      setError('')
-      setIsPublishing(true)
-
-      let postUri
-      try {
-        postUri = (
-          await apilib.post(agent, queryClient, {
-            thread,
-            replyTo: replyTo?.uri,
-            onStateChange: setPublishingStage,
-            langs: toPostLanguages(langPrefs.postLanguage),
-          })
-        ).uris[0]
-        try {
-          await whenAppViewReady(agent, postUri, res => {
-            const postedThread = res.data.thread
-            return AppBskyFeedDefs.isThreadViewPost(postedThread)
-          })
-        } catch (waitErr: any) {
-          logger.error(waitErr, {
-            message: `Waiting for app view failed`,
-          })
-          // Keep going because the post *was* published.
-        }
-      } catch (e: any) {
-        logger.error(e, {
-          message: `Composer: create post failed`,
-          hasImages: thread.posts.some(p => p.embed.media?.type === 'images'),
-        })
-
-        let err = cleanError(e.message)
-        if (err.includes('not locate record')) {
-          err = _(
-            msg`We're sorry! The post you are replying to has been deleted.`,
-          )
-        } else if (e instanceof EmbeddingDisabledError) {
-          err = _(msg`This post's author has disabled quote posts.`)
-        }
-        setError(err)
-        setIsPublishing(false)
-        return
-      } finally {
-        if (postUri) {
-          let index = 0
-          for (let post of thread.posts) {
-            logEvent('post:create', {
-              imageCount:
-                post.embed.media?.type === 'images'
-                  ? post.embed.media.images.length
-                  : 0,
-              isReply: index > 0 || !!replyTo,
-              hasLink: !!post.embed.link,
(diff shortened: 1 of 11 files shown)


robbwdoering:
Incredible work, can't wait! Love me a social media app with good commit hygiene.

I'd request a little `n/N` counter in the final Post render before you click in to indicate that it's a thread, potentially with a cute emoji or NerdIcon, but obviously that can wait until V2 (3? 4?) if it's tricky and/or controversial.  I'll try to remember to circle back and file this as a Feature Request if it doesn't make it in :)

src/view/com/composer/Composer.tsx
```diff
@@ -319,143 +318,140 @@ export const ComposePost = ({
...
-      if (!canPost) {
-        return
-      }
+  const onPressPublish = React.useCallback(async () => {
+    if (isPublishing) {
+      return
+    }
 
-      if (
-        !finishedUploading &&
```
gaearon, line 345: I believe this wasn't necessary. If the videos are actually uploaded, the condition below will let us through regardless. If not, we shouldn't get through regardless.

src/view/com/composer/Composer.tsx
```diff
@@ -539,44 +581,34 @@ export const ComposePost = ({
...
+                  isReply={index > 0 || !!replyTo}
+                  isActive={post.id === activePost.id}
+                  canRemovePost={thread.posts.length > 1}
+                  canRemoveQuote={index > 0 || !initQuote}
+                  onSelectVideo={selectVideo}
+                  onClearVideo={clearVideo}
+                  onPublish={onComposerPostPublish}
+                  onError={setError}
+                />
+                {isFooterSticky && post.id === activePost.id && footer}
```
gaearon, line 607: Took this from Twitter. IMO it makes sense. Web-only behavior.

src/view/com/composer/Composer.tsx
```diff
@@ -769,6 +856,8 @@ function ComposerTopBar({
             <ButtonText style={[a.text_md]}>
               {isReply ? (
                 <Trans context="action">Reply</Trans>
```
gaearon, line 858: theoretically Reply should also have something ("Reply All"?) but i imagine it would be too confusing to translate. meh.

src/view/com/composer/Composer.tsx
```diff
@@ -1261,9 +1379,14 @@ const styles = StyleSheet.create({
...
   },
+  post: {
+    marginHorizontal: 16,
+  },
+  inactivePost: {
+    opacity: 0.5,
+  },
   scrollView: {
     flex: 1,
-    paddingHorizontal: 16,
```
gaearon, line 1266: Replacing this with margins on inner content so we can let the footer go full bleed even when rendered within the post list.

src/view/com/composer/Composer.tsx
```diff
@@ -226,20 +230,15 @@ export const ComposePost = ({
 
   const clearVideo = React.useCallback(
     (postId: string) => {
-      const post = thread.posts.find(p => p.id === postId)
-      const postMedia = post?.embed.media
-      if (postMedia?.type === 'video') {
-        postMedia.video.abortController.abort()
```
gaearon, line 232: this moved into the reducer, can't find a more natural place

haileyok:
Pretty nitpicky, but maybe we want to change it from "one or more images" to "one or more items" (or some other word) for the alt text warning? Took me a second to realize my images all had alt but not the video, heh

src/view/com/composer/Composer.tsx
```diff
@@ -655,17 +704,17 @@ function ComposerPost({
...
         onImageAdd([res])
       }
     },
-    [onSelectVideo, onImageAdd],
+    [post.id, onSelectVideo, onImageAdd],
   )
 
   return (
-    <>
+    <View style={[styles.post, !isActive && styles.inactivePost]}>
```
haileyok, line 717: Just leaving a note for reference:

If we encounter problems with being able to scroll on Android over "disabled" post areas, I think the solution would be to actually set the text inputs as not-editable, essentially disabling them and preventing "interception" of touches.

Unclear though if that was just an emulator issue or a bigger problem. We'll know more once testing on physical device.

haileyok (approved):
There's some questions on Android, but I think we'll need to do some real "use" there to get a good sense of any actual problems. I'll start using my androids more exclusively after we get this in and try to notice any jank we can fix.

Overall things look great though! This is awesome stuff, and huge thanks to @mary-ext for all your early work on getting this done!

gaearon:
merged into wrong base, resubmit: https://github.com/bluesky-social/social-app/pull/6050

PR #5967, opened 2024-10-27
[WIP] Update react-navigation

gaearon:
Trying to get `fullScreenSwipeShadowEnabled` working but it doesn't seem to.

https://github.com/user-attachments/assets/596ce466-6283-4c05-81c8-3914950f753b


src/Navigation.tsx
```diff
@@ -506,6 +510,7 @@ function MessagesTabNavigator() {
         animationDuration: 285,
         gestureEnabled: true,
         fullScreenGestureEnabled: true,
+        fullScreenSwipeShadowEnabled: true,
```
hirbod, line 513:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

src/Navigation.tsx
```diff
@@ -480,6 +483,7 @@ function MyProfileTabNavigator() {
...
```
hirbod, line 486:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

src/Navigation.tsx
```diff
@@ -458,6 +460,7 @@ function NotificationsTabNavigator() {
...
```
hirbod, line 463:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

src/Navigation.tsx
```diff
@@ -421,6 +421,7 @@ function HomeTabNavigator() {
...
```
hirbod, line 424:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

hirbod (commented):
Should work after this change

src/Navigation.tsx
```diff
@@ -440,6 +441,7 @@ function SearchTabNavigator() {
...
```
hirbod, line 444:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

src/Navigation.tsx
```diff
@@ -540,6 +545,7 @@ const FlatNavigator = () => {
...
```
hirbod, line 548:
```suggestion
        fullScreenGestureShadowEnabled: true,
```

gaearon:
Fantastic, thank you!

https://github.com/user-attachments/assets/7758229d-f2bc-4717-aa09-209f5fba83da

hirbod:
You're welcome. Looks much better, adds the missing depth to the transition.

gaearon:
Now just need to fix 30 type errors 💀

hirbod:
I'll try to fix the weird looking snap in the react-native-screens repo next week. I really hate the cancel animation.

gaearon:
Ah yes, that would be lovely.

src/components/Link.tsx
```diff
@@ -92,10 +95,21 @@ export function useLink({
...
+      : !to.screen
+      ? convertBskyAppUrlIfNeeded(sanitizeUrl(to.href))
+      : buildHref(to.screen, to.params)
+  }, [to, buildHref])
+
+  if (!href) {
+    throw new Error(
+      'Link `to` prop must be a string or an object with `screen` and `params` properties',
+    )
+  }
```
mozzius: This is one of the significant changes I'm most unsure of

src/Navigation.tsx
```diff
@@ -507,29 +511,26 @@ function TabsNavigator() {
...
 }
 
+function screenOptions(t: Theme) {
+  return {
+    fullScreenGestureEnabled: true,
+    headerShown: false,
+    contentStyle: t.atoms.bg,
+  } as const
+}
+
```
mozzius, line 525: New default screen options - removed options that were the defaults anyway, and removed the animation duration (only affects back swipe gesture, and I want to try out the default (can change back to 285ms))

src/components/Link.tsx
```diff
@@ -92,10 +95,21 @@ export function useLink({
...
-    to:
-      typeof to === 'string' ? convertBskyAppUrlIfNeeded(sanitizeUrl(to)) : to,
-  })
+  const {buildHref} = useLinkBuilder()
+  const href = useMemo(() => {
+    return typeof to === 'string'
+      ? convertBskyAppUrlIfNeeded(sanitizeUrl(to))
+      : !to.screen
+      ? convertBskyAppUrlIfNeeded(sanitizeUrl(to.href))
+      : buildHref(to.screen, to.params)
```
estrattonbailey: Ah yeah kinda sticky. This suggestion is a little more explicit, but yeah I think the `throw` is a good move. I don't really expect any issues with any code we have in prod today.
```suggestion
      : to.screen
      ? buildHref(to.screen, to.params)
      : to.href
      ? convertBskyAppUrlIfNeeded(sanitizeUrl(to.href))
      : undefined
```

estrattonbailey: Thinking we want to prefer `screen` to `href`

mozzius: Yeah I'm assuming that the old code would've thrown anyway, just from inside the hook. Now we have to handle it explicitly

src/Navigation.tsx
```diff
@@ -507,29 +511,26 @@ function TabsNavigator() {
...
-        gestureEnabled: true,
-        fullScreenGestureEnabled: true,
-        headerShown: false,
-        contentStyle: t.atoms.bg,
-      }}>
+    <HomeTab.Navigator screenOptions={screenOptions(t)} initialRouteName="Home">
       <HomeTab.Screen name="Home" getComponent={() => HomeScreen} />
       <HomeTab.Screen name="Start" getComponent={() => HomeScreen} />
-      {commonScreens(HomeTab)}
+      {commonScreens(HomeTab as typeof Flat)}
```
haileyok, line 533: Flat? We call those apartments here, this isn't the UK.

estrattonbailey (approved):
Nothing jumping out to me here. I think we're awfully close to being able to remove the legacy behavior, but agree let's do that as a next step.

Thanks for picking this back up!

PR #5973, opened 2024-10-28
Limit number of lines in starter pack card

haileyok:
## Why

Quick follow up to https://github.com/bluesky-social/social-app/pull/5891. Just avoiding any UI breaks because of excessive line counts.

## Test Plan

Should be no change from https://github.com/bluesky-social/social-app/pull/5891.


mozzius (approved):
perf

PR #5974, opened 2024-10-28
FAB - delay followup haptic

mozzius:
this ensures you can feel both of them even if you tap fast. also adds to reply prompt


haileyok (approved):
Have not tested, timing seems probably okay though.

Should copy/paste this logic into the reply button on feed items though.

PR #5976, opened 2024-10-28
Make composer min height conditional and align attachments

gaearon:
There's a web-only min height on the composer, but having it always be on (even when there's attachments) is annoying. Twitter doesn't do that. This is the easiest way I found to adjust the behavior.

I've also adjusted the padding on mobile to be consistent between different attachment types.

## Before

https://github.com/user-attachments/assets/820ce96d-0c5e-4ecd-9af1-d8aa7b288518

## After

https://github.com/user-attachments/assets/5d624a8a-736b-4aca-8f15-c20c84bf362b

https://github.com/user-attachments/assets/425f3c60-25ef-4518-bdea-d99c203035d5

https://github.com/user-attachments/assets/dc18cb6e-81fa-4667-81c1-8267e4b62142

## Test Plan

Try all types of attachments on all platforms. (Including links, post embeds, list embeds, starter pack embeds.)

On the web, the expectation is that min-height is only enforced when there's no attachment.

Also verify the same for replies.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -92,9 +92,6 @@ a[role='link'][data-no-underline='1']:hover {
 }
 
 /* ProseMirror */
-.ProseMirror {
-  min-height: 140px;
-}
 .ProseMirror-dark {
   color: white;
 }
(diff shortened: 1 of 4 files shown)


haileyok (approved):
Tested all three platforms, seems fine on each!

luangjokaj:
ugh soo many spacing issues and tiny details to fix in this UI 😬 I'm tempted to make a PR.

gaearon:
A lot could be improved :) PRs welcome.

PR #5986, opened 2024-10-28
Fix double padding in `Dialog.InnerFlatList` on web

mozzius:
I changed the structure of `Dialog.Inner` around so padding works differently now. Unfortunately, `Dialog.InnerFlatList` tries to remove the horizontal padding, which then stopped working. This uses the correct prop to return the style to how it was

### DIFF
--- a/src/components/Dialog/index.web.tsx
+++ b/src/components/Dialog/index.web.tsx
@@ -217,14 +217,13 @@ export const InnerFlatList = React.forwardRef<
     <Inner
       label={label}
       style={[
+        a.overflow_hidden,
+        a.px_0,
         // @ts-ignore web only -sfn
-        {
-          paddingHorizontal: 0,
-          maxHeight: 'calc(-36px + 100vh)',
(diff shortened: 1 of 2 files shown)


src/components/Dialog/index.web.tsx
```diff
@@ -217,14 +217,13 @@ export const InnerFlatList = React.forwardRef<
...
-        {
-          paddingHorizontal: 0,
-          maxHeight: 'calc(-36px + 100vh)',
-          overflow: 'hidden',
-        },
+        {maxHeight: 'calc(-36px + 100vh)'},
         webInnerStyle,
-      ]}>
+      ]}
+      contentContainerStyle={[a.px_0]}>
```
mozzius, line 226: This is the main line, the rest are cleanups

haileyok: Seems good, I think we want to also pass in the `contentContainerStyle` from props here too though.

PR #5987, opened 2024-10-29
fix starterpack wizard screen

mozzius:
Same fix as DMs


haileyok (approved):
does fix. wonder if we can move this right into `Layout.Screen`. overall should do a pass at some point irt not needing styles like this or `a.util_screen_outer`. super easy to miss these styles 😭

PR #5996, opened 2024-10-29
Fix video quality for short videos

mozzius:
Two things which increase the quality

- Remove the requirement to click on the video to see the HD version
  - Not how it works on native, generally feels pretty horrible and leads to all the videos you watch being ugly
- Fix the edge case where low quality segments wouldn't get purged for short videos

# Test plan

Check that this one increases in quality after it loops (should be pretty obvious). You also should be able to see the increased quality without needing to click on it
https://bsky.app/profile/passivestar.bsky.social/post/3l7heqzikmi2j


gaearon (approved):
makes sense

PR #5998, opened 2024-10-29
add toki pona as a language a post can be tagged with

eightbitgiraffe:
i saw the previous pr was deleted so I made my own


eightbitgiraffe:
ive capitalized `toki pona` so that it fits with the exonyms that are used everywhere else in the list. (and also updated my fork, hopefully that doesnt mess up anything, idk im bad with git and github)

auroursa:
Apologies if I’m mistaken, but it seems that a language needs to have a code2 designation (ISO 639-1) to be used as a tagging language for posts.

eightbitgiraffe:
it does look like a lot of the other languages don't have a `code2` so im not sureOn Oct 29, 2024, at 10:54 PM, Frudrax Cheng ***@***.***> wrote:﻿
Apologies if I’m mistaken, but it seems that a language needs to have a code2 designation (ISO 639-1) to be used as a tagging language for posts.

Sobsz:
Skimming through the Post Languages list in the app, it seems to be a 1:1 match with the languages that have 2-character codes. [The source code](<https://github.com/bluesky-social/social-app/blob/1fc684aaef6dd851b4d6c9a23673fcf4e7170722/src/state/preferences/languages.tsx>) also refers to code2 often. I'm not sure of the reason; perhaps it's a way to coarsely trim the language list to somewhat-commonly used ones, in lieu of a search function?

tbodt:
2 character codes are not assigned anymore... if the software can't handle them that's a bug.

tbodt:
#6769 fixes 3 character codes

softastur:
Asturian language is also affected by this bug. Our locale is "ast"

PR #5999, opened 2024-10-29
Refactor lightbox model to plain object

gaearon:
This starts a stack of refactors to get the image open animation working.

In this PR, I'm just changing the model to be a plain object. There is no need for it to be a class. This aligns a bit better with the rest of the codebase and also fixes the unsound `as` casting in some of the code.

## Test Plan

- Verify can click on avatar to view it full screen
- Verify can click on an image in a post
- Verify can click on one of multiple images
  - Verify the gallery opens at the index that was clicked

Also, types.


mozzius (approved):
Couldn't find issues clicking about on web

PR #6000, opened 2024-10-29
Show almost-instant preview when opening lightbox

gaearon:
Stacked on #5999 

---

We have two kinds of images: the "previews" shown in feed and fullsize images.

Previously, the lightbox would always show the fullsize image. After this change, the lightbox will initially show the "preview" and *then* switch to the fullsize image once it's downloaded. This works via Expo Image `placeholder` prop. 

This means that, in most cases, opening lightbox should now feel *almost* instant. (No animation yet though.) It's not fully instant because there's still a delay to load it from the cache etc — we'll have more control over it after upgrading Expo.

This PR achieves that in several steps:

1. We need to plumb down the thumbnail URL — this one is easy.
2. This couldn't work before because we waited for `onLoad` event and forcibly displayed the spinner until then. The problem is that `onLoad` doesn't fire for placeholder images. So we couldn't use the `placeholder` prop before. To solve that, I'm changing the layout to place the spinner *beneath* the image. Then it can be rendered unconditionally.
  - I've had to resolve a minor layout hiccup on Android which was compositing opacity during dismiss incorrectly.
3. While the image is loading, its dimensions aren't available. There's no quick way to get placeholder dimensions either so instead we just rely on the gestures (like zoom and pinch) being disabled until dimensions load.
  - The logic to ignore missing dimensions was already present. However, I've noticed that with internet being offline, the dimensions were incorrectly reported as `0x0` which caused double-tap gesture to send `NaN`s to `zoomToRect`, causing a crash on iOS. I've made a driveby fix to treat `0x0` as missing dimensions and to avoid calculations based on those.

## Before

You always have to wait for the full image to load first. If there is no WiFi, you don't see anything at all on opening lightbox.

https://github.com/user-attachments/assets/885d57f6-780a-4387-863a-594626b3df7d

## After

Images will appear instantly (and progressively enhance as full size loads) thanks to the preview being cached via feed.

If you have no WiFi, you still see the lower quality image (almost) instantly:

https://github.com/user-attachments/assets/c7a5c4b5-ecf0-42d4-929c-e6ce58c8bdfc

However, you won't be able to interact with it (aside from dismissing) until the internet comes back online.

## Test Plan

Try with WiFi off and on.

Substitute `thumbUri` with some fake URI to verify the image *does* get swapped.

Verify that the spinner still works (e.g. by setting `thumbUri` to a completely unrelated URL that isn't cached and turning off WiFi). Verify you can dismiss the image before even the placeholder has loaded.

Verify that while dismissing the image, the spinner beneath it is not visible.


src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -313,20 +311,23 @@ const ImageItem = ({
...
           contentFit="contain"
           source={{uri: imageSrc.uri}}
-          style={[styles.image, animatedStyle]}
+          placeholderContentFit="contain"
+          placeholder={{uri: imageSrc.thumbUri}}
+          style={styles.image}
           accessibilityLabel={imageSrc.alt}
           accessibilityHint=""
-          onLoad={() => setIsLoaded(true)}
+          accessibilityIgnoresInvertColors
```
gaearon, line 330: was told to add this by the linter

src/view/com/lightbox/Lightbox.tsx
```diff
@@ -31,7 +31,9 @@ export function Lightbox() {
     const opts = activeLightbox
     return (
       <ImageView
-        images={[{uri: opts.profile.avatar || ''}]}
+        images={[
+          {uri: opts.profile.avatar || '', thumbUri: opts.profile.avatar || ''},
```
gaearon, line 35: could pass the low res one but it's maybe _too_ low res? can change tho

src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -72,7 +72,7 @@ export function ProfileSubpageHeader({
     ) {
       openLightbox({
         type: 'images',
-        images: [{uri: avatar}],
+        images: [{uri: avatar, thumbUri: avatar}],
```
gaearon, line 75: same

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
```diff
@@ -141,18 +135,21 @@ const ImageItem = ({
...
-          source={{uri: imageSrc.uri}}
-          style={[styles.image, animatedStyle]}
-          accessibilityLabel={imageSrc.alt}
-          accessibilityHint=""
-          onLoad={() => setLoaded(true)}
-          enableLiveTextInteraction={showControls && !scaled}
-        />
+        <Animated.View style={[styles.imageScrollContainer, animatedStyle]}>
+          <ActivityIndicator size="small" color="#FFF" style={styles.loading} />
+          <Image
```
haileyok, line 141: Nice. This is actually the _"right"_ way to do this as of right now (using an `Animated.View` around the `Image`, as there's a bug when animating the image directly. Didn't realize we were doing this the whole time...would be curious to profile before/after and see if it was indeed reloading when doing some of the resize operations (probably was eating CPU)

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
```diff
@@ -141,18 +135,21 @@ const ImageItem = ({
...
-          contentFit="contain"
-          source={{uri: imageSrc.uri}}
-          style={[styles.image, animatedStyle]}
-          accessibilityLabel={imageSrc.alt}
-          accessibilityHint=""
-          onLoad={() => setLoaded(true)}
-          enableLiveTextInteraction={showControls && !scaled}
-        />
+        <Animated.View style={[styles.imageScrollContainer, animatedStyle]}>
+          <ActivityIndicator size="small" color="#FFF" style={styles.loading} />
```
haileyok, line 140: Do we want to use `Loader` instead to keep consistency with what we do elsewhere? No strong opinion, whatever you think looks nicer in this context.

haileyok (approved):

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
gaearon: would rather not touch actual UI for now to stay safe, but can do later

PR #6001, opened 2024-10-29
Measure tapped image coordinates before opening lightbox

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/5999 and https://github.com/bluesky-social/social-app/pull/6000

---

When you click on an image preview in a feed, we need to know the screen coordinates of the image you clicked. This will later allow us to animate the image expanding into the lightbox.

In this PR, I'm measuring the screen coordinates for solo images and galleries in the feeds. The coordinates become a part of the lightbox state object and get plumbed into the lightbox component. The lightbox component doesn't use it yet.

## Test Plan

Add `console.log(_thumbDims)` in `ImageViewing/index.tsx` and verify that the coordinates make sense. Do it on iOS and Android.

It should be `null` for profile images cause I haven't bothered hooking it up there. In general the code is supposed to be resilient to the dimensions being `null` (in case something breaks with the measurement).

There's no user-visible behavior change in this PR.

### DIFF
--- a/src/view/com/util/post-embeds/index.tsx
+++ b/src/view/com/util/post-embeds/index.tsx
@@ -6,6 +6,14 @@ import {
   View,
   ViewStyle,
 } from 'react-native'
+import Animated, {
+  AnimatedRef,
+  measure,
+  MeasuredDimensions,
+  runOnJS,
+  runOnUI,
+  useAnimatedRef,
+} from 'react-native-reanimated'
 import {Image} from 'expo-image'
 import {
   AppBskyEmbedExternal,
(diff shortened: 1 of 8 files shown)


src/view/com/util/post-embeds/index.tsx
```diff
@@ -138,13 +147,27 @@ export function PostEmbeds({
...
         })
       }
+      const onPress = (
+        index: number,
+        ref: AnimatedRef<React.Component<{}, {}, any>>,
+      ) => {
+        runOnUI(() => {
+          'worklet'
+          const dims = measure(ref)
+          runOnJS(_openLightbox)(index, dims)
```
gaearon, line 168: This adds a bit of nested thread hopping. Not sure if a concern. Worth checking if it got slower. Not sure if there's another way to get measurements (I only want to do them lazily when actually needed, and not always).

mozzius (approved):
looks good. I don't think the thread hopping is a concern

PR #6012, opened 2024-10-30
fix text clipping around external embed's domain

benharri:
before:
image

after:
image


benharri:
fixed another case of clipped text on Handle.tsx

haileyok (approved):
thank you!

PR #6015, opened 2024-10-30

auroursa:
This PR includes the latest translations for zh-CN, zh-TW and zh-HK (Cantonese), and cleans up outdated strings.
Co-maintained by @KuwaLee @cirx1e


auroursa:
Here’s an explanation for why extra spaces were removed:

It seems that {profileName} could sometimes appear as "you", which makes adding spaces or quotation marks awkward and disrupts the flow of the sentence. 

image

By removing these extra spaces, we can ensure the text feels more natural and fluid, regardless of the placeholder’s content.

PR #6017, opened 2024-10-30
Fix: iOS Keyboard Persists When Opening Sidebar

yoleuh:
On iOS, when hitting the sidebar/menu icon while the search input is focused, the keyboard remains visible and the search input stays focused

Changes: 
Added blur() call on the search input reference when the sidebar icon is pressed, ensuring the keyboard dismisses properly

image


gaearon (approved):
lgtm

Issue #6026, opened 2024-10-31
Post Interaction Settings Selectables' text are clipped/overflowed


merojosa:
### Steps to Reproduce

1. Open modal to create new post
2. Click the post interaction settings

### Attachments

![image](https://github.com/user-attachments/assets/24a595fd-efde-4c03-85a7-5035c4daf1f1)

### What platform(s) does this occur on?

Android

### What version of the app are you using?

Build version: 1.92.1.355; Bundle info: 89cff31b6 (prod); Bundle date: 24101917; Platform: android

### Additional Information

Hi, I just saw that the selectable components ("Everybody", "Nobody", "Mentioned users", "Followed users") for the post interaction settings are cutting their text on Android due to the height or padding.

This PR https://github.com/bluesky-social/social-app/pull/6027 tries to fix this issue by increasing the height of the components. I chose 43 to have the same height as the "Save" button. Feel free to disregard the PR if you don't need it.

auroursa:
It has been fixed in #6527.

PR #6027, opened 2024-10-31
Fix PostInteractionSettingsForm Selectable height

merojosa:
Fixes https://github.com/bluesky-social/social-app/issues/6026


auroursa:
I was able to reproduce this issue when using a larger system font size.

But I hope it can adjust the height according to the actual size of the font. When the system font is smaller, it will not waste too much height space.

merojosa:
@auroursa huh, you are right. I didn't think about it because my font size settings are the default ones.

Checking the rest of the app with a larger font size, it seems every component adjuts its height as expected, so the issue is only on Post interaction settings.

Not sure if the Bluesky team is interested, but I'll try to fix it according to the font size when I have time, should be quick hopefully.

PR #6033, opened 2024-10-31
Add AT URIs as alternate links

tom-sherman:
> [!NOTE]
> I'm not sure how to test this or if this is where the change needs to actually go. I've pattern matched against the received HTML in prod and found these files, so let me know if it's incorrect and I'll rework this!

See https://github.com/likeandscribe/frontpage/pull/172 for the description of why I'd like to make this change.

Some extra discussion here: https://bsky.app/profile/did:plc:2xau7wbgdq4phuou2ypwuen7/post/3l7esaih44s2s

### DIFF
--- a/bskyweb/templates/post.html
+++ b/bskyweb/templates/post.html
@@ -38,6 +38,7 @@
   <meta name="article:published_time" content="{{ postView.IndexedAt }}">
   <meta name="article:published_time" content="{{ postView.IndexedAt }}">
   <link rel="alternate" type="application/json+oembed" href="https://embed.bsky.app/oembed?format=json&url={{ postView.Uri | urlencode }}" />
+  <link rel="alternate" href="{{ postView.Uri }}" />
 {% endif -%}
 {%- endblock %}
 
--- a/bskyweb/templates/profile.html
+++ b/bskyweb/templates/profile.html
@@ -37,6 +37,7 @@
   {%- if requestHost %}
   <link rel="alternate" type="application/rss+xml" href="https://{{ requestHost }}/profile/{{ profileView.Did }}/rss">
   {% endif %}
+  <link rel="alternate" href="at://{{ profileView.Did }}" />
 {% endif -%}
 {%- endblock %}
 


bnewbold:
I'm default ok with this, though it feels a bit under-specified. Should there be a `type` field, and if so what value?

There is a small related tension with AT URIs not quite being valid "URL" syntax, following WHATWG standards: https://url.spec.whatwg.org/

The issue is that there are multiple colons in the DID in the "authority" segment of the URI. I think there is a path to relaxing the spec to allow this specific pattern (DIDs in that part of URL), but it will be a delicate conversation: browsers have tons of noisy inbound requests and it is important to be diplomatic and make a clear case.

Anyways, probably none of that is a reason not to go ahead with this. Be bold!

bnewbold (approved):

mary-ext:
shouldn't profiles link to `app.bsky.actor.profile` record?

it seems very ambiguous, because it can mean many things:
1. am I linking to my account's repository?
2. am I linking to my Bluesky profile?
3. am I linking to my linkat.blue page?
4. am I linking to my <some other services here>'s profile?

bnewbold:
@mary-ext I think that `at://<identifier>` definitely does *not* imply anything `bsky` specific. An atproto account (identified by DID or handle) *definitely* could mean any or all of the things you mentioned.

I don't think that having a `<link rel="alternative">` pointing to an overall account in this way would imply an equivalency of the account with the page itself. It does make sense to me to link something like a bsky profile view page to the "overall account".

In practical terms, it does probably make more sense to tie `https://bsky.app/profile/<identifier>` to a profile record (so full AT URI), if nothing else because it includes the account identifier anyways.

(note that this codebase could be forked/white-labeled to run on other domains; in that case the relevant thing is still probably the `app.bsky.actor.profile` record)

bnewbold:
FWIW, I think it would also be fine/appropriate to have *both* a ref to the bsky profile record, and the atproto account identifier. Or even just the bare DID.

ngerakines:
Aren't brackets acceptable for URLs? Like, an ipv6 URL looks like `http://[::1]`. Personally, I don't see any issue with `at://[did:plc:tgudj2fjm77pzkuawquqhsxm]/events.smokesignal.calendar.event/3kxbvxj7blk2t`.

Issue #6035, opened 2024-10-31
Video currentTime is lost after scrolling away from video


tlaverdure:
### Steps to Reproduce

1. Go to this video: https://bsky.app/profile/tobyfox.undertale.com/post/3l7stsnpdnn2u
2. After a couple seconds scroll down to comments
3. Scroll back up to the video and you will notice the video starts over

### Attachments

https://github.com/user-attachments/assets/07378f93-a3a0-446f-b634-9816e59f5383

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Macbook Pro

### What version of the app are you using?

Latest

PR #6037, opened 2024-10-31
Fix video currentTime after scrolling away from video 

tlaverdure:
On web, when you scroll away from a video an `IntersectionObserver` is used to `play()` and `pause()` the video. As described in  #6035 the `currentTime` of the video is lost after scrolling away. It appears conditional rendering in the following code is causing this:

https://github.com/bluesky-social/social-app/blob/6bc5a05f4bdfd3bf9dea400b3a6b5d9ac356457a/src/view/com/util/post-embeds/VideoEmbed.web.tsx#L145-L159

**I've traced these changes back to two prior PRs for context:**
* https://github.com/bluesky-social/social-app/pull/5246
* https://github.com/bluesky-social/social-app/pull/5251

I didn't experience any issues playing a fullscreen video and the `currentTime` of the video was retained as expected.

### DIFF
--- a/src/view/com/util/post-embeds/VideoEmbed.web.tsx
+++ b/src/view/com/util/post-embeds/VideoEmbed.web.tsx
@@ -104,7 +104,6 @@ function ViewportObserver({
   isAnyViewActive: boolean
 }) {
   const ref = useRef<HTMLDivElement>(null)
-  const [nearScreen, setNearScreen] = useState(false)
   const [isFullscreen] = useFullscreen()
   const isWithinMessage = useIsWithinMessage()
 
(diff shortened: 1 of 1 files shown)


mozzius:
This was very much intentional, to ensure that the number of video players doesn't grow indefinitely as you scroll. I'd be open to a solution that caches/restores the play position but still unmounts the player when scrolling away

tlaverdure:
Thanks for the feedback. Can you share a little more about the resources you are concerned about consuming? I imagine a paused video should also pause the hls stream and once a user leaves a feed the video element should also be destroyed.

Since a video is unmounted each time you scroll past it, the current experience leads to multiple http requests being fired each time the video is remounted. I feel like this is an a more common performance concern as opposed to what may happen if a user loads thousands of videos and they are all paused. 

For example, just now on my following feed, which has already been loaded, I scrolled past like 80 posts. On the way down I re-downloaded 8MB, I then scrolled back up and re-downloaded another 8MB.

PR #6041, opened 2024-10-31
Show muted word in `ContentHider`

estrattonbailey:
Depends on backend PR: https://github.com/bluesky-social/atproto/pull/2934

I think we can improve how this looks, but solved for the base need to start.

CleanShot 2024-10-31 at 15 52 06@2x
CleanShot 2024-10-31 at 16 16 26@2x


cooljeanius:
Would be nice to have proper pluralization support based on localization or whatever, rather than the "(s)" method. Also a setting to control whether this additional info is shown or not would be useful.

simnaut:
This should be optional, because when a user mutes a word they don't want to see that word, even if its in a cause description

PR #6045, opened 2024-10-31
Fixes for thread composer on Android

gaearon:
Stacked on #5962 

---

Easier to review the whole diff.

Does a couple of things:

- Turns off `input.focus` into previous post after deleting a post on Android. This unfortunately risks doing something weird with the keyboard where the input ends up behind it and the scroll view becomes unaware there is more to scroll. I.e. it breaks `KeyboardAvoidingView`. I found no other fix than to disable the behavior.
- <s>Replaces `ScrollView` with `KeyboardAwareScrollView` which on its own (even without the above fix) seemed to significantly reduce how often the above issue occurred. Although I did disable the `focus` calls, I suspect they're not the only cause, so I've left this fix in too.</s>
  - Actually removed this — caused issues on iOS and didn't seem necessary after the fix above.
- Changes how "autoscroll to end" works. The old approach only triggers scroll to end if you pressed "+" on the last post. But it's also confusing if you add media and it doesn't scroll down. I realized that it's easier to do this in `onContentSizeChaged` instead. We already had a handler for this so I expanded its scope a little bit. Also did a minor naming cleanup and refactor there.

## Test Plan

Use a real Android device.

Create a long enough thread to go offscreen. Observe best effort attempts to keep you scrolled down (it doesn't always look aligned though, I suspect because input autofocus is fighting with explicit `scrollTo` — ah well).

Observe that adding media in the last post will scroll you down to it (at least after a bit of a delay) assuming you were scrolled down roughly to the end.

Try to spam "+" and deleting posts in the thread. Try deleting the last post, a post in the middle, etc. You should hopefully never enter a situation where the scroll view refuses to scroll to the last items.

Just check that it's ... usable?

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1205,11 +1205,21 @@ export function useComposerCancelRef() {
 
 function useAnimatedBorders() {
   const t = useTheme()
-  const hasScrolledToTop = useSharedValue(1)
-  const hasScrolledToBottom = useSharedValue(1)
   const contentOffset = useSharedValue(0)
   const scrollViewHeight = useSharedValue(Infinity)
   const contentHeight = useSharedValue(0)
+
+  const hasScrolledToTop = useDerivedValue(() =>
+    withTiming(contentOffset.value === 0 ? 1 : 0),
+  )
+
+  const hasScrolledToBottom = useDerivedValue(() =>
+    withTiming(
+      contentHeight.value - contentOffset.value - 5 <= scrollViewHeight.value
+        ? 1
+        : 0,
+    ),
+  )
 
   const showHideBottomBorder = useCallback(
     ({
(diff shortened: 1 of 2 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -1213,17 +1213,30 @@ export function useComposerCancelRef() {
...
+function useScrollTracker({
+  scrollViewRef,
+  stickyBottom,
+}: {
+  scrollViewRef: AnimatedRef<Animated.ScrollView>
+  stickyBottom: boolean
+}) {
   const t = useTheme()
-  const hasScrolledTop = useSharedValue(0)
-  const hasScrolledBottom = useSharedValue(0)
```
gaearon, line 1219: NB: I flipped their names/meaning because the old names were ambiguous. https://github.com/bluesky-social/social-app/pull/6045/commits/61e928b21394a52ca7b12fd04108f4516c30716f

src/view/com/composer/Composer.tsx
```diff
@@ -1213,17 +1213,30 @@ export function useComposerCancelRef() {
...
   const contentOffset = useSharedValue(0)
   const scrollViewHeight = useSharedValue(Infinity)
   const contentHeight = useSharedValue(0)
 
-  /**
-   * Make sure to run this on the UI thread!
```
gaearon, line 1225: This was already called from JS thread in some cases, I adjusted https://github.com/bluesky-social/social-app/pull/6045/commits/51c129fab5ab93e9ed0bcf9dfea8a30f0743b149 to reflect. It's also fine actually aside from how values are read, https://github.com/bluesky-social/social-app/pull/6045/commits/a1b6af01e4009914b828c3617b2633da32deca6c took care of the only bit that was off.

src/view/com/composer/Composer.tsx
```diff
@@ -1213,17 +1213,30 @@ export function useComposerCancelRef() {
   return useRef<CancelRef>(null)
 }
 
-function useAnimatedBorders() {
+function useScrollTracker({
+  scrollViewRef,
+  stickyBottom,
+}: {
+  scrollViewRef: AnimatedRef<Animated.ScrollView>
+  stickyBottom: boolean
```
gaearon, line 1221: made this a flag just to emphasize this hook is doing more things now

haileyok:
lfg

haileyok (approved):

PR #6046, opened 2024-11-01
Link to app.bsky.actor.profile record in link rel=alternate

char:
re: #6033

what do we think chat? this disambiguates Bluesky profiles from ATProto identities. 

after writing this, I've just seen [mary's comment](https://github.com/bluesky-social/social-app/pull/6033#issuecomment-2451109988) raising the same issue


bnewbold:
I replied over in the other thread. I think that linking to an atproto account identifier would be "fine" in this context, and would not set a norm or expectation that `at://<identifier>` "means" or "implies" a bsky profile.

But I agree that this is probably pragmatically better. Thanks for the contribution!

bnewbold (approved):

PR #6047, opened 2024-11-01
Unify dimensions cache between lightbox and feed

gaearon:
We have two pieces of code doing essentially the same thing but with separate caches. (The "thing" is measuring image sizes and caching them.) I'm unifying them into a single cache:

- Now it's always an LRU (previous feed cache grew unbounded for each item without known `aspectRatio`)
- The hook forces to specify `knownDimensions` which should catch cases where we know them beforehand
- Changes to `src` while rendering are now handled 
- Initial state is now reading from the cache if there's a hit
- The hook returns `undefined` when the dimensions aren't known yet (instead of returning `1` for ratio)
- The hook also now returns a tuple (aspect ratio first) where the ratio is calculated safely (for div 0).

In the final commit, I'm changing the lightbox code to use the dimensions plumbed down from the record when available. This means that for the vast majority of images, we won't trigger the "download and measure" codepath anymore.

## Test Plan

Add `setTimeout` inside the hook before resolving and add `knownDimensions = undefined` to force loading.

Verify the feed jumps a bit as image dimensions get resolved (this is the worst case, e.g. when the records are done by clients that don't fill in `aspectRatio`) but that it works. Verify you can open the lightbox still.

Add some logs to cache hits, misses, and active request tracking. Verify they work as expected.

Check solo lightbox, gallery lightbox, and profile avi lightbox.

Verify hardcoding `width: 0` or `height: 0` in the resize callback doesn't crash the lightbox.

Remove the `setTimeout` and `knownDimensions = undefined`. Verify we mostly use `knownDimensions` except for records with missing `aspectRatio` like https://bsky.app/profile/unseenjapan.com/post/3l5mkbqrqq72y. Verify the same for lightbox.


src/lib/media/image-sizes.ts
```diff
@@ -1,35 +1,85 @@
+import {useEffect, useState} from 'react'
 import {Image} from 'react-native'
 
 import type {Dimensions} from '#/lib/media/types'
 
-const sizes: Map<string, Dimensions> = new Map()
+type CacheStorageItem<T> = {key: string; value: T}
+const createCache = <T>(cacheSize: number) => ({
```
gaearon, line 7: this is copypasta from deleted file

mozzius (approved):
tested, looks good

PR #6048, opened 2024-11-01
Animate lightbox

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/6047 and https://github.com/bluesky-social/social-app/pull/6073

---

Animates the lightbox.

- [x] Animate in
- [x] Fix out-of-bound zooming on iOS
- [x] Animate out
  - [x] Swipe
  - [x] Close button
- [x] Prevent interactions during animation
- [x] Respect reduce motion
- [x] Fix "open 1, flick to 2, press close"
- [x] Sometimes initial bounds aren't calculated right

Known issues (addressed in https://github.com/bluesky-social/social-app/pull/6088)

- Cropped edges are visible from the first animation frames 
- Gestures outside the image (e.g. flick from black bar) don't work
- Swiping up sometimes fails after a quick swipe through gallery — requires to move the pager before it reactivates

I'll add some inline comments for the review but you kind of have to read the whole thing.

## Test Plan

Walk around the feed. Try opening images in different orientations. It's good to try some portrait and landscape ones. Try opening portrait and landscape _from_ different layouts (1 image, 2 images, 3 images, 4 images). Generally the first frame should be pixel-perfect (note edges of the image that don't fit in the thumbnail will "stick out" in the first frame — this is something I might be able to resolve in a follow-up but keeping out of scope for now).

Verify opening profile images still works, both for profile and feeds/SPs with custom avis. These aren't animated yet (that'll be added in https://github.com/bluesky-social/social-app/pull/6081).

Verify images that lack `aspectRatio`, like https://bsky.app/profile/unseenjapan.com/post/3l5mkbqrqq72y, open but without an animation. (It's possible to animate them too in principle but I want to keep the code a bit simpler. No dimension information — no animation. Encourages third party clients to supply this information too.)

Verify you can still swipe between multiple images. Verify panning and pinching still works.

Verify that on iOS, after you zoom into the image, you can no longer "go outside" to the black bars. Previously it would awkwardly let you do that.

Verify that double tap to zoom works.

Try different ways of closing the image. Closing it via the close button should animate out (if the image had `aspectRatio`). An image without an `aspectRatio` should close without animation. 

Verify you can open a post with multiple images, swipe to a different one, then press close. It should animate out to the image you were on last — i.e. possibly a different image than the one you opened.

Verify you can flick the image away up or down with a swipe. As the image goes offscreen, the lightbox should disappear. (In this case, there is no gradual "come back" animation.) Verify that the lightbox doesn't get stuck and you're able open it again after that.

Slow down the animation in code (set `stiffness` to `10`) and verify you can't break anything by spamming gestures.

https://github.com/user-attachments/assets/745021da-ef07-4a0a-86fc-047e50486a95


src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -232,6 +327,39 @@ function LightboxFooter({
...
+
+  const initialTranslateX = thumbnailCenterX - screenCenterX
+  const initialTranslateY = thumbnailCenterY - screenCenterY
+  let initialScale = thumbnailDims.width / screenSize.width
+
+  const imageAspect = imageDims.width / imageDims.height
+  const thumbAspect = thumbnailDims.width / thumbnailDims.height
+  if (imageAspect > 1 && thumbAspect < 1) {
+    initialScale = initialScale * imageAspect
+  }
```
gaearon, line 355: TODO: is inverted version of this also needed?

src/screens/Profile/Header/Shell.tsx
```diff
@@ -59,6 +59,7 @@ let ProfileHeaderShell = ({
           {
             uri: profile.avatar,
             thumbUri: profile.avatar,
+            thumbRect: null,
```
gaearon, line 62: this is now per-image because we need to support "open one image, swipe to next one, close" and animate out the other one. i.e. we need to know _each_ image's dimensions.

src/state/lightbox.tsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react'
-import type {MeasuredDimensions} from 'react-native-reanimated'
+import {nanoid} from 'nanoid/non-secure'
 
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {ImageSource} from '#/view/com/lightbox/ImageViewing/@types'
 
-type Lightbox = {
+export type Lightbox = {
+  id: string
```
gaearon, line 8: for react key, just in case

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -43,13 +46,14 @@ type Props = {
...
   showControls: boolean
+  dismissSwipePan: PanGesture
 }
 const ImageItem = ({
   imageSrc,
   onTap,
   onZoom,
-  onRequestClose,
   isScrollViewBeingDragged,
+  dismissSwipePan,
```
gaearon, line 56: we're moving the "swipe to dismiss" handling upwards because it needs to integrate with the opening animation which isn't forked by platform

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
```diff
@@ -290,33 +282,11 @@ const ImageItem = ({
...
-        })
-      }
-    })
-
   const composedGesture = isScrollViewBeingDragged
     ? // If the parent is not at rest, provide a no-op gesture.
       Gesture.Manual()
     : Gesture.Exclusive(
-        dismissSwipePan,
+        dismissSwipePan ?? Gesture.Manual(),
```
gaearon, line 289: this is just a noop

src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
```diff
@@ -128,50 +110,45 @@ const ImageItem = ({
...
+        centerContent>
+        <ActivityIndicator size="small" color="#FFF" style={styles.loading} />
+        <Image
+          contentFit="contain"
+          source={{uri: imageSrc.uri}}
+          placeholderContentFit="contain"
+          placeholder={{uri: imageSrc.thumbUri}}
+          style={{
+            width: SCREEN.width,
+            height: imageAspect ? SCREEN.width / imageAspect : undefined,
```
gaearon, line 139: new layout stuff. i needed to figure out how to position the scroll view so that its scrolling container doesn't stretch all the way to the top and bottom of the screen. this is important because it fixes an old iOS issue. previously, it was possible to zoom in and then vertically move past the black bars. so it was hard to zoom into corners. now this is fixed.

the price for that is that we won't show the image _until_ we've loaded its dimensions (because we don't know the height). this is generally ok now because https://github.com/bluesky-social/social-app/pull/6073 plumbs dimensions from `aspectRatio` in the majority of cases. (in other cases, we'll wait for the `Image.getSize` to call back — which means there'll be no nice transition and you'll see a spinner — but it's no worse than status quo.)

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -16,70 +16,148 @@ import {
...
+  onPressShare: (uri: string) => void
+}) {
+  const {images, index: initialImageIndex} = lightbox
   const [isScaled, setIsScaled] = useState(false)
   const [isDragging, setIsDragging] = useState(false)
   const [imageIndex, setImageIndex] = useState(initialImageIndex)
   const [showControls, setShowControls] = useState(true)
+  const dismissSwipeTranslateY = useSharedValue(0)
+
+  const animatedHeaderStyle = useAnimatedStyle(() => {
```
gaearon, line 76: controls hiding/showing is now integrated both with "swipe away" gesture and with open/close animation

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -16,70 +16,148 @@ import {
...
+  const activeImageStyle = useAnimatedStyle(() => {
+    if (openProgress.value === 1 || dismissSwipeTranslateY.value !== 0) {
+      return {
+        pointerEvents: 'auto',
+        transform: [{translateY: dismissSwipeTranslateY.value}],
+      }
+    }
+    const image = images[imageIndex]
+    if (image.thumbRect && image.dimensions) {
+      const interpolatedTransform = interpolateTransform(
```
gaearon, line 110: we're in the middle of the animation (opening or closing)

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -232,6 +327,48 @@ function LightboxFooter({
   )
 }
 
+function interpolateTransform(
```
gaearon, line 330: this little goblin took me several days to get right but it looks obvious in retrospect

src/view/com/lightbox/ImageViewing/index.tsx
```diff
@@ -277,13 +421,77 @@ const styles = StyleSheet.create({
...
+  onRequestClose,
+  onPressSave,
+  onPressShare,
+}: {
+  lightbox: Lightbox | null
+  onRequestClose: () => void
+  onPressSave: (uri: string) => void
+  onPressShare: (uri: string) => void
+}) {
+  const [activeLightbox, setActiveLightbox] = useState(nextLightbox)
```
gaearon, line 435: note! forking props into state. this is intentional because we want to treat new `lightbox` prop values as sort of "requests" — and trigger animations in response

src/view/com/lightbox/Lightbox.tsx
```diff
@@ -49,16 +49,9 @@ export function Lightbox() {
     [permissionResponse, requestPermission, _],
   )
 
-  if (!activeLightbox) {
-    return null
```
gaearon, line 53: we don't want that here anymore so that the props flow down and trigger leave animation

src/view/com/util/images/ImageLayoutGrid.tsx
```diff
@@ -54,35 +54,46 @@ function ImageLayoutGridInner(props: ImageLayoutGridInnerProps) {
   const gap = props.gap
   const count = props.images.length
 
+  const containerRef1 = useAnimatedRef()
+  const containerRef2 = useAnimatedRef()
+  const containerRef3 = useAnimatedRef()
+  const containerRef4 = useAnimatedRef()
```
gaearon, line 60: gotta keep track of all of them now

gaearon:
https://github.com/bluesky-social/social-app/pull/6159

PR #6051, opened 2024-11-01
Marked "Sort Replies" as translatable

auroursa:
Based on #5772

"Sort replies" was not translatable, so I marked it.

Also made a slight adjustment to "Sort Replies," as this term had already been translated previously and could be reused directly.


mozzius:
Thanks for catching this! Gonna close in favour of https://github.com/bluesky-social/social-app/pull/6110

PR #6053, opened 2024-11-01
Actually move images to the intended cache directory and ignore data URIs

mary-ext:
Need additional verification, cc @gaearon 

Just realized that the `moveIfNecessary` function wasn't working as intended, the images never actually gets moved to our intended cache directory because `from.startsWith(cacheDir)` will always return false for those images.

Added an extra check to not do it for data URIs as well, to handle images coming from share intents, and images coming from local drafts (which I'm still working on).

### DIFF
--- a/src/state/gallery.ts
+++ b/src/state/gallery.ts
@@ -248,7 +248,7 @@ export async function compressImage(img: ComposerImage): Promise<ImageMeta> {
 async function moveIfNecessary(from: string) {
   const cacheDir = isNative && getImageCacheDirectory()
 
-  if (cacheDir && from.startsWith(cacheDir)) {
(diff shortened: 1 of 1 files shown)


gaearon:
what’s the user visible consequence of the bug?

mary-ext:
Nothing user-visible, it just means that we weren't actually deleting the temporary images as they weren't ever moved there to begin with.

mary-ext:
will find some time to write up a test case tomorrow!

It is definitely harder to see what's going on, I'm not sure if file explorer apps on Android are able to peek into our cache directory so we might just have to check by putting console.logs on the before&after

gaearon:
i think i was able to inspect it with Android Studio in the past, just need to better understand what to look for

PR #6055, opened 2024-11-01
Local post thread drafts

mary-ext:
## To-dos

- [x] Validation schemas for saved drafts
- [x] Ability to save drafts
- [x] Modify composer reducer to accept the draft entry
- [ ] UI for user to view and restore drafts

### Native

- [ ] expo-sqlite persistence (the only place that seems to make sense)
- [x] Ensure that images are converted to data URIs

### Web

- [x] IndexedDB persistence

### DIFF
--- a/src/state/composer-drafts/database.ts
+++ b/src/state/composer-drafts/database.ts
@@ -1,0 +1,22 @@
+import {DraftEntry, UnknownDraftEntry} from './database.types'
+
+export const listDrafts = async ({
+  cursor: _cursor,
+  limit: _limit = 50,
+}: {
+  limit?: number
+  cursor?: string
+}): Promise<{
+  cursor: string | undefined
+  entries: UnknownDraftEntry[]
+}> => {
+  throw new Error(`unimplemented`)
+}
+
+export const putDraft = async (_entry: DraftEntry): Promise<void> => {
+  throw new Error(`unimplemented`)
+}
+
+export const removeDraft = async (_entryId: string): Promise<void> => {
+  throw new Error(`unimplemented`)
+}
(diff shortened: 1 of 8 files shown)


gaearon:
what's your thinking on schema changing etc? do we need to stay diligent with it? only evolve it in backwards-compatible way? where does validation happen?

mary-ext:
I'm still blank on these unfortunately.

I think the versioning is still good to have, but if things can be introduced in a backwards-compatible manner then we should go that route.

As for where validations should go... I was thinking it would be closer to the UI so we can present users the ability to delete them, although we could definitely just delete these ourselves (very unsure on this, at least for mobile you have the option to downgrade the app, but that's not much of a thing on web)

gaearon:
how do you want to handle media? it just gets serialized? what about videos?

mary-ext:
Right now media files are saved as data URIs:
- on native, we don't have the equivalent of IndexedDB, where it allows us to put JS' Blob types on the DB row. our closest is expo-sqlite, which only ticks the "it's a database" checkbox.
- on web, we're already dealing with data URIs, no question there.
- since we're forced to serialize to string, I'm not sure if there's a good way to ensure that no one tampers with the files that are referenced by the draft entries
- so data URI it is, since that means we're putting the entire file into the DB row.
- I don't think data URIs would do so well for larger files like videos, so drafting is disabled for videos.
- It also doesn't mesh well with the fact that we are uploading the video eagerly, prior to tapping the Post button, if the user decides to abort their draft or whatever, then that's bandwidth wasted, not good for users.
- Twitter doesn't do drafting for videos either.

GIFs could be supported but right now I'm not interested on finding out which fields from Tenor's API response we're actually using, so I decided to punt on it later.

PR #6059, opened 2024-11-01
Change post Follow button color on toggle

gaearon:
For consistency with other Follow buttons and for general legibility.

## Before

https://github.com/user-attachments/assets/6a323bd7-bc6d-49de-b024-173c783c53b1

## After

https://github.com/user-attachments/assets/9ec6bc6b-c301-4f19-8fdf-de341297a7e0

https://github.com/user-attachments/assets/1ddae2bc-519a-4259-96d5-85b7da630caf

https://github.com/user-attachments/assets/4665c5b3-250c-4e43-a307-791eba791c82

### DIFF
--- a/src/view/com/post-thread/PostThreadFollowBtn.tsx
+++ b/src/view/com/post-thread/PostThreadFollowBtn.tsx
@@ -115,7 +115,7 @@ function PostThreadFollowBtnLoaded({
       onPress={onPress}
       size="small"
       variant="solid"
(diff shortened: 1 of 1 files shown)


haileyok (approved):

geobde:
@danninov, I’m not entirely familiar with the codebase, but maybe this logic would fit better inside the component(like colouring based on button's state) rather than using conditional rendering. What do you think?

PR #6060, opened 2024-11-01
Override X-Bsky-Topics with a debugging param

gaearon:
For convenience when trying new flags.

## Test Plan

Not sent to non-bsky feeds:

Screenshot 2024-11-01 at 16 39 42

Bsky feeds still normally receive the topics header:

Screenshot 2024-11-01 at 16 39 17

But http://localhost:19006/?debug_topics=1234 overrides it:

Screenshot 2024-11-01 at 16 39 31


mozzius (approved):
I see the header, looks good

PR #6061, opened 2024-11-01
Improve focus-visible outlines in navigation

Tahul:
I noticed focus outlines on buttons are overflow most of the time.

The behavior replicates in currently deployed app on all browsers I tried (Safari, Firefox, Chrome)

Tried to improve it by adding `outline-offset`.

`outline-offset` do not seem supported by the Atoms convention, yet it seem [widely available](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset).

Here is tab navigation before the PR:

CleanShot 2024-11-01 at 19 27 25

You can notice almost all outlines overflows their parent container.

And here it is after the changes:

CleanShot 2024-11-01 at 19 28 03

I think we could also improve the outlines for avatar/collapse button by using `tabIndex` prop, the TypeScript errors made me question it.

### DIFF
--- a/src/alf/atoms.ts
+++ b/src/alf/atoms.ts
@@ -1,4 +1,4 @@
-import {Platform, StyleSheet, ViewStyle} from 'react-native'
+import {Platform, StyleProp, StyleSheet, ViewStyle} from 'react-native'
 
 import * as tokens from '#/alf/tokens'
 import {native, web} from '#/alf/util/platform'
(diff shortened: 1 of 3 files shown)


gaearon:
I don't think we want to apply this to every button — e.g. that squeezes the avatars in feed too tightly. Can you go through the UI and look for places where it's beneficial? E.g. I pushed the fix for left nav. If you could try to do it in more places where it definitely works well, that would be very appreciated.

gaearon (approved):
i'll just apply this fix to the left nav for now

Tahul:
hey @gaearon ; I have been continuing the work on fixing the outlines on some other places.

happy to send that soon in another branch :)

PR #6063, opened 2024-11-01
Add dids for discover debug

haileyok:


src/view/com/util/post-ctrls/PostCtrls.tsx
```diff
@@ -363,7 +363,7 @@ let PostCtrls = ({
           threadgateRecord={threadgateRecord}
         />
       </View>
-      {IS_INTERNAL && feedContext && (
```
gaearon, line 366: i think we should leave `IS_INTERNAL` in too, it's just another way to get it

haileyok: sure

gaearon (approved):

PR #6068, opened 2024-11-01
Fix duplicates in thread composer

gaearon:
Due to multiple subscriptions to the same global emitter, things were getting duplicated on the web.

Now there is only one active subscription, corresponding to the active input.

## Test Plan

- Draft a thread and press Cmd+Enter.
  - Before: Thread gets posted N times.
  - After: Thread gets posted once.
- Drag an image from desktop.
  - Before: Image gets inserted into each post N times.
  - After: Image gets inserted into the current post once.
- Paste an image from clipboard.
  - Before: Image gets pasted into each post.
  - After: Image gets pasted into the current post.
- Choose an emoji in the web picker.
  - Before: Emoji gets added to every post.
  - After: Emoji gets added to the current post.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -736,6 +736,7 @@ let ComposerPost = React.memo(function ComposerPost({
           placeholder={selectTextInputPlaceholder}
           autoFocus
           webForceMinHeight={forceMinHeight}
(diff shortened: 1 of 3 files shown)


mozzius (approved):
tested, works, lgtm!

PR #6070, opened 2024-11-01
thread composer: avoid emitting action on paste for inactive posts

frncs-eu:
@gaearon 
I found this little thing while [exploring the new shiny stuff](https://bsky.app/profile/frncs.eu/post/3l7vsn6cmlc24)!
Basically EVERY post inside the thread composer was listening for the paste event and emitting the `embed_add_images` action, so every reducer was updating with the embed data to display.
This should stop that.
I tested on web only but AFAIK this (copy/pasting image content) should not be possible?
Let me know if it's ok!

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -715,6 +715,9 @@ let ComposerPost = React.memo(function ComposerPost({
 
   const onPhotoPasted = useCallback(
     async (uri: string) => {
+      if (!isActive) {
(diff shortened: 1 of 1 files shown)


surfdude29:
Might be similar to #6068?

gaearon:
yup, https://github.com/bluesky-social/social-app/pull/6068 is a more correct fix because this event shouldn't have been firing in the first place. thanks for looking into it though!

PR #6073, opened 2024-11-01
Refactor lightbox prop drilling

gaearon:
Stacked on #6047

---

No changes to behavior.

This may be easier to read commit by commit.

- I get rid of the `cloneElement` and the weird data flow where the footer reads data via the hook. A more natural way to express this is to have the parent component pass a render prop to the child, closing over the lightbox data.
- Then I get rid of the two lightbox types. As far as the lightbox is concerned, everything here is images. It shouldn't be thinking about profile pictures etc. This makes it annoying to evolve the code. So I unified the types.
- Then I refactor `LightboxFooter` to take all `images`. This is a bit cleaner and allows it to make its own decisions about how to deal with them (right now it just reads the current one, but it doesn't have to).
- Finally, I decide to move `LightboxFooter` so it lives together with the actual lightbox shell. I keep some of the app-specific logic out of it but the visual parts are now moved. This will make adding animation later easier.

## Test Plan

Open lightbox with 1 image, or with multiple. Flick through them. Verify Save/Share buttons work and point at the right image. Verify the alt text shows up for the right image.

Open lightbox in profile too. Verify that works.

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -55,8 +55,18 @@ let ProfileHeaderShell = ({
     const modui = moderation.ui('avatar')
     if (profile.avatar && !(modui.blur && modui.noOverride)) {
       openLightbox({
-        type: 'profile-image',
-        profile: profile,
+        images: [
+          {
+            uri: profile.avatar,
+            thumbUri: profile.avatar,
+            dimensions: {
+              // It's fine if it's actually smaller but we know it's 1:1.
+              height: 1000,
+              width: 1000,
+            },
+          },
+        ],
+        index: 0,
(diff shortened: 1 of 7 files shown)


src/screens/Profile/Header/Shell.tsx
```diff
@@ -55,8 +55,18 @@ let ProfileHeaderShell = ({
     const modui = moderation.ui('avatar')
     if (profile.avatar && !(modui.blur && modui.noOverride)) {
       openLightbox({
-        type: 'profile-image',
-        profile: profile,
+        images: [
+          {
```
gaearon, line 59: this is copypasta moved from another file (see below)

mozzius (approved):
lgtm

PR #6084, opened 2024-11-02
fix(a11y): avoid plain `div`s as button or tabs

### LINKED ISSUE #4409: can't navigate to replies/reposts/likes buttons within keyboard on pc via some browser extensions
**Describe the bug**

(first i'm not sure this is a bug or not, it might be an intended behaviour, thus i'd like core devs to verify my claim. if it's intended, i'd like you to convert this issue to a `feature request`.)

since around v1.85 was out, i've been unable to click the said buttons when i navigate to them only with my keyboard input on pc.

**To Reproduce**

1. install a browser extension that allows you to click links and buttons without mouse clicks, such as Vimium (that's what i've been using), if you haven't yet.
2. open bluesky web app.
3. try to click one of the said buttons. in Vimium, you start with a `F` key to see clickable elements on a page.
4. you don't see the extension's clickable element indicator on those buttons, unlike the other elements such as someone's profile photo, name, post date, sidebar items, and hashtags and links in posts.

**Expected behavior**

the currently unclickable buttons explained above are clickable just as they were until a few days ago.

**Screenshots**

the current situation (after i pressed a `F` key with Vimium installed):

Capture d’écran 2024-06-07 à 13 11 16

you can click the elements that have yellow indicators by pressing the keys shown within an indicator, however the said buttons don't have any of them.

**Details**

 - Platform:  firefox desktop
 - Platform version: ?
 - App version: ?

**Additional context**

i checked what the html elements were used for the buttons and currently they're DIV elements. Vimium doesn't see elements that aren't A, BUTTON, or some form items clickable, even if it's got a click or such event handler or something like `aria-role="button"`.

cyyynthia:
This PR improves accessibility by turning skeet controls (Translate, Reply, Reskeet, Like, Share, More) actual buttons instead of `div`s with an event handler.

Same with tabs such as on profiles, they are now properly marked as tabs for assistive technologies to pick up.

Closes #4409

### DIFF
--- a/src/components/Menu/index.web.tsx
+++ b/src/components/Menu/index.web.tsx
@@ -111,7 +111,7 @@ const RadixTriggerPassThrough = React.forwardRef(
 )
 RadixTriggerPassThrough.displayName = 'RadixTriggerPassThrough'
 
-export function Trigger({children, label}: TriggerProps) {
+export function Trigger({children, label, role = 'button'}: TriggerProps) {
   const {control} = React.useContext(Context)
   const {
     state: hovered,
@@ -155,6 +155,7 @@ export function Trigger({children, label}: TriggerProps) {
               onMouseEnter,
               onMouseLeave,
               accessibilityLabel: label,
+              accessibilityRole: role,
             },
           })
         }
--- a/src/view/com/pager/TabBar.tsx
+++ b/src/view/com/pager/TabBar.tsx
@@ -118,7 +118,10 @@ export function TabBar({
   )
 
   return (
-    <View testID={testID} style={[pal.view, styles.outer]} accessibilityRole="tablist">
+    <View
+      testID={testID}
+      style={[pal.view, styles.outer]}
+      accessibilityRole="tablist">
       <DraggableScrollView
         testID={`${testID}-selector`}
         horizontal={true}
--- a/src/view/com/pager/TabBar.tsx
+++ b/src/view/com/pager/TabBar.tsx
@@ -118,7 +118,7 @@ export function TabBar({
   )
 
   return (
-    <View testID={testID} style={[pal.view, styles.outer]}>
+    <View testID={testID} style={[pal.view, styles.outer]} accessibilityRole="tablist">
       <DraggableScrollView
         testID={`${testID}-selector`}
         horizontal={true}
@@ -135,7 +135,8 @@ export function TabBar({
               onLayout={e => onItemLayout(e, i)}
               style={styles.item}
               hoverStyle={pal.viewLight}
-              onPress={() => onPressItem(i)}>
+              onPress={() => onPressItem(i)}
+              accessibilityRole="tab">
               <View style={[styles.itemInner, selected && indicatorStyle]}>
                 <Text
                   emoji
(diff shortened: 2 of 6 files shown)


haileyok (approved):
Thank you! Made a small tweak, will merge after I test it works...

src/components/Menu/index.web.tsx
rshigg, line 1: Might not be something we address in this PR but shouldn't the web implementation use a literal `button` element rather than relying on an aria role?

src/view/com/pager/TabBar.tsx
```diff
@@ -135,7 +138,8 @@ export function TabBar({
               onLayout={e => onItemLayout(e, i)}
               style={styles.item}
               hoverStyle={pal.viewLight}
-              onPress={() => onPressItem(i)}>
+              onPress={() => onPressItem(i)}
+              accessibilityRole="tab">
```
rshigg, line 142: Are there any tab panels we can associate with these tabs?

src/components/Menu/index.web.tsx
cyyynthia: React Native is quite smart here and makes it use a literal button on web (this also applies to the other places where a role has been added).

src/view/com/pager/TabBar.tsx
cyyynthia: Ideally the panel views would be associated with the tabs (as `tabpanel`s with `aria-labelledby`), but I didn't feel confident adding it (and afraid it'd require adapting multiple places where the component is used); so I kept it simple.

FWIW, not associating a tabpanel is not a violation of the WAI-ARIA specification -- it's a *should* -- and should not be an issue so long the semantics of the page still allow assistive technologies to let the user navigate. It's best practice to associate them, but getting rid of clickable plain divs is already a huge step up. 😅

I think (am pretty sure) there are many accessibility nits and issues throughout the app, so properly doing the association could be handled in future improvements

src/components/Menu/index.web.tsx
rshigg: Ah, I see. That's great to know 😄

src/view/com/pager/TabBar.tsx
rshigg: Agreed, it's definitely nice to get some semantics in there without having to do a huge refactor 👍 would be nice to eventually have it all hooked up.

haileyok:
Verified the tweak I made works, really appreciate you cleaning these problems up 🙏

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -744,12 +745,13 @@ function ExpandedPostDetails({
             &middot;
           </NewText>
 
-          <NewText
+          <InlineLinkText
```
SleeplessByte, line 748: This should not be a link but a button. 

The `label` is not necessary and adding it will actually break translation (e.g Google Translate) as it will not translate the `aria-label`.

SleeplessByte (commented):
I know this was merged, and definitely an improvement. I did leave a comment about something I ran into just now.

PR #6081, opened 2024-11-02
Enable lightbox animation for avis, make them rounded

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/6047, https://github.com/bluesky-social/social-app/pull/6073, and https://github.com/bluesky-social/social-app/pull/6048

---

Extends https://github.com/bluesky-social/social-app/pull/6048 to opening profile avatars. While at it, I've also made the lightbox display them as circles (or rounded rects for feeds/SPs) and limited displayed size on the web so they don't blow up.

[Review without spaces](https://github.com/bluesky-social/social-app/pull/6081/files?w=1)

## Known Issues

- Initial avi size is slightly off, might need to move the ref to the child (addressed in https://github.com/bluesky-social/social-app/pull/6088)

## Test Plan

https://github.com/user-attachments/assets/c1ca8eab-5456-43f5-b5fc-13d85b209aad

https://github.com/user-attachments/assets/25559c05-435c-438e-8f9e-5db191cdfef1

https://github.com/user-attachments/assets/aa882d7f-9d8c-4fc4-bbdf-c38c5d2dea29

Verify normal lightbox (regular images) still works too.

PR #6088, opened 2024-11-03
Fixes to animated lightbox

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/6047, https://github.com/bluesky-social/social-app/pull/6073, https://github.com/bluesky-social/social-app/pull/6048, and https://github.com/bluesky-social/social-app/pull/6081

---

## Known issues

- Slow in prod on low-end Android
  - I have a few ideas to try, otherwise might have to disable it on Android
- Incorrect positioning on Android device 
  - Doesn’t happen on emulator. 
- Flash of borders on the thumbnail after dismiss via swipe 
- Extra high aspect is behaving weirdly  https://bsky.app/profile/schlagteslinks.bsky.social/post/3l7y4l6yur72e
- I’ve had it lock up on iOS

---

Fixes for known issues in the preceding stack. It would've been a pain to rebase and I had to be very careful not to break other things. This seems to be an equilibrium of features working but it's quite fragile. See commit by commit.

- Fixes swipe-up gesture stopping working if you spam it too fast during a pager slide on iOS.
- Fixes being unable to drag or perform gestures with finger(s) on the black bars on Android.
- Fixes unintended ability to paginate while in the middle of an opening or closing animation.
- A single tap in zoomed state now always zooms out and shows controls. It was too hard to make the close animation (if you pressed close after bringing up controls in a zoomed state) work well. Now it just pops you back before showing controls. Twitter does the same.
- Cropped images no longer flash their cropped edges on the first animation frame. Now they continuously expand. This needs device testing because it definitely affects perf a bit. Mostly relevant for multi-image layouts.
- The avatar first frame is now precisely positioned.

## Test Plan

All of the above. Plus the usual. Try portraits/landscapes in different layouts (1, 2, 3, 4). Slow down if needed to verify animation.

https://github.com/user-attachments/assets/126f79c2-1967-4e54-9544-573cf6a5cc37

https://github.com/user-attachments/assets/cd6c1af6-8420-4035-b8d9-53032170b175

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -162,21 +162,22 @@ let ProfileHeaderShell = ({
           accessibilityRole="image"
           accessibilityLabel={_(msg`View ${profile.handle}'s avatar`)}
           accessibilityHint="">
-          <Animated.View
-            ref={aviRef}
+          <View
             style={[
               t.atoms.bg,
               {borderColor: t.atoms.bg.backgroundColor},
               styles.avi,
               profile.associated?.labeler && styles.aviLabeler,
             ]}>
-            <UserAvatar
-              type={profile.associated?.labeler ? 'labeler' : 'user'}
-              size={90}
-              avatar={profile.avatar}
-              moderation={moderation.ui('avatar')}
-            />
-          </Animated.View>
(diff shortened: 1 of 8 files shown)


gaearon:
https://github.com/bluesky-social/social-app/pull/6159

PR #3708, opened 2024-04-25
adding mute account and block account buttons

skiniks:
Reopening with adjustments because I accidentally closed #2610.

Screenshot 2024-04-25 at 3 40 29 PM

### Original PR Description
This PR adds Mute Account and Block Account buttons to the Post dropdown menu to close issue https://github.com/bluesky-social/social-app/issues/1017.

PR #6091, opened 2024-11-03
Add ability to block account from post dropdown menu

rshigg:
## Details

While #3708 already exists, it hasn't been updated sine April and I personally don't think including the mute option is necessary. If it is decided that a mute button is needed/wanted, I can update this PR or I could open a separate PR.

- I matched the blocking flow from the profile header but made it one-directional.
- I grouped the block option with the report option because they are both moderation actions.

Note: I don't consider this PR complete since I assume we will want some sort of visual feedback that a user was blocked while in a feed. My initial thinking was to hide the post after blocking but it felt a little janky. I would appreciate some feedback.

## Demo

(Sorry Abby and Jerry, it's not personal 😅 )

| From a feed | From a profile page |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/1b094b9c-41f9-4b3f-96c8-482e45f874c7" /> | <video  src="https://github.com/user-attachments/assets/52541dae-6fb2-4e75-9da5-284c46c01869" /> | 
| <video src="https://github.com/user-attachments/assets/3081cd87-b262-4383-83c4-aa50f3d68b26" /> | <video  src="https://github.com/user-attachments/assets/845f749b-98cc-4f4b-87a3-cc2593e48817" /> |


Enitoni:
Yes! Please add this. I was just about to make an issue for this. This makes blocking have less friction and makes the UX much better, especially in terms of a11y.

src/view/screens/Search/Search.tsx
```diff
@@ -805,6 +805,7 @@ export function SearchScreen(
     } else {
       setSearchText('')
       navigation.setParams({q: ''})
+      textInput.current?.focus()
```
haileyok, line 808: Is having this here intentional? Does it do anything in relation to this PR?

rshigg: Sorry this was a rebase mishap on my end, I'll remove this line from the PR as soon as I can 🫡

haileyok: No worries, I can remove!

haileyok (approved):
Nice! I made a small change to use `useProfileShadow` and just use `ProfileViewBasic` in the `profile` arg type (since all we ever need from the profile is did and viewer, which is available in all of the `ProfileViewX` types), no need for the `Pick`.

Really appreciate this one, been something on the todos for a while and just never had time to do it.

MrZaius:
Thank you!

PR #6092, opened 2024-11-03
Add tiktok embed

maxinne-dev:
## Details

I noticed that Tiktok links had no embed option, and even the link preview card looked very raw, then on a [comment](https://github.com/bluesky-social/social-app/issues/2314#issuecomment-1871070227) on #2314, @haileyok make some points about the Tiktok embeds. While searching a possible solution, I found out that now tiktok offers an embed player, and it was quite simple to add. While at it, I solved an issue with shortened Tiktok links that wouldn't return the page metadata due to an 301 redirect. This commits solves both the short `vm.tiktok.com` link and the embed to allow viewing Tiktok videos on the feed.

## Demo

https://github.com/user-attachments/assets/ee2bec2a-53ff-40e2-abf4-dfd3d8e6d5f5

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -1,11 +1,11 @@
 import {RichText} from '@atproto/api'
 
-import {parseEmbedPlayerFromUrl} from 'lib/strings/embed-player'
+import {parseEmbedPlayerFromUrl} from '#/lib/strings/embed-player'
 import {
   createStarterPackGooglePlayUri,
   createStarterPackLinkFromAndroidReferrer,
   parseStarterPackUri,
-} from 'lib/strings/starter-pack'
+} from '#/lib/strings/starter-pack'
 import {cleanError} from '../../src/lib/strings/errors'
 import {createFullHandle, makeValidHandle} from '../../src/lib/strings/handles'
 import {enforceLen} from '../../src/lib/strings/helpers'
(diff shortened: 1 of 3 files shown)


__tests__/lib/string.test.ts
```diff
@@ -1,11 +1,11 @@
 import {RichText} from '@atproto/api'
 
-import {parseEmbedPlayerFromUrl} from 'lib/strings/embed-player'
+import {parseEmbedPlayerFromUrl} from '#/lib/strings/embed-player'
```
auroursa, line 3:
```suggestion
import {parseEmbedPlayerFromUrl} from 'lib/strings/embed-player'
```

__tests__/lib/string.test.ts
```diff
@@ -1,11 +1,11 @@
...
 import {
   createStarterPackGooglePlayUri,
   createStarterPackLinkFromAndroidReferrer,
   parseStarterPackUri,
-} from 'lib/strings/starter-pack'
+} from '#/lib/strings/starter-pack'
```
auroursa, line 8: And this part
```suggestion
} from 'lib/strings/starter-pack'
```

auroursa (commented):
It might be the code consistency tool at work, but perhaps we should keep these areas the same as before?

__tests__/lib/string.test.ts
mary-ext: The imports are correct, there's now a lint error that enforces the #/ prefix for alias imports

__tests__/lib/string.test.ts
mary-ext: Same as my other comment

__tests__/lib/string.test.ts
auroursa: thanks for your patient explanation

PR #6095, opened 2024-11-04
Split Notifications into "All" & "Mentions" tabs

ericclemmons:
This PR solves the firehose that is Notifications by adding a filtered _Mentions_ tab, based on my prototype here: 

> https://bksy.app/profile/ericclemmons.com/post/3l7wti3xu7d2f

(I hope this effort doesn't negatively affect https://github.com/bluesky-social/social-app/pull/3612!)

## TODOs

- [x] Client-side tabs for _All_ and _Mentions_ (filtered to `reply` and `quote` types)
- [x] ~Update backend~

  > "you know, if it works decently in your experience, maybe we can give client-side solution a try even though it kinda sucks."
  > – https://bsky.app/profile/danabra.mov/post/3l7y3ssjlfk23
  - [ ] ~Add `NotificationType` filter to backend API~
  - [ ] ~Update `useNotificationFeedQuery` to support `NotificationType` allowlist~
  - [ ] ~Pass `tab.types` to `useNotificationFeedQuery`~

## Demo

| Platform | All Tab			| Mentions Tab			|
|----------|----------------|---------------------------|
| Android  | All		    | Mentions        				|
| iOS      | All | Mentions
| Web      | All | Mentions  |

### DIFF
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -77,7 +77,7 @@ export function Feed({
     overridePriorityNotifications,
   })
   const isEmpty = !isFetching && !data?.pages[0]?.items.length
-  const isTransitioning = tab.types !== deferredTab.types
+  const isTransitioning = tab !== deferredTab
 
   const items = React.useMemo(() => {
     if (isTransitioning) return [LOADING_ITEM]
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -33,15 +33,10 @@ const EMPTY_FEED_ITEM = {_reactKey: '__empty__'}
 const LOAD_MORE_ERROR_ITEM = {_reactKey: '__load_more_error__'}
 const LOADING_ITEM = {_reactKey: '__loading__'}
 
-type Tab = {
+const tabs: {
   label: string
   types?: NotificationType[]
-}
-
-const tabs: Tab[] = [
-  {label: 'All'},
-  {label: 'Mentions', types: ['reply', 'quote']},
-]
+}[] = [{label: 'All'}, {label: 'Mentions', types: ['reply', 'quote']}]
 
 export function Feed({
   scrollElRef,
--- a/src/view/com/notifications/Feed.tsx
+++ b/src/view/com/notifications/Feed.tsx
@@ -96,14 +96,6 @@ export function Feed({
     return arr
   }, [isFetched, isError, isEmpty, data])
 
-  const filteredItems = React.useMemo(() => {
-    if (!tab.types) {
-      return items
-    }
-
-    return items.filter(item => tab.types?.includes(item.type))
-  }, [items, tab])
-
   const onRefresh = React.useCallback(async () => {
     try {
       setIsPTRing(true)
@@ -167,16 +159,26 @@ export function Feed({
             <NotificationFeedLoadingPlaceholder />
           </View>
         )
+      } else if (!tab.types || tab.types.includes(item.type)) {
+        return (
+          <FeedItem
+            item={item}
+            moderationOpts={moderationOpts!}
+            hideTopBorder={index === 0 && isTabletOrMobile}
+          />
+        )
       }
-      return (
-        <FeedItem
-          item={item}
-          moderationOpts={moderationOpts!}
-          hideTopBorder={index === 0 && isTabletOrMobile}
-        />
-      )
+
+      return null
     },
-    [moderationOpts, isTabletOrMobile, _, onPressRetryLoadMore, pal.border],
+    [
+      moderationOpts,
+      isTabletOrMobile,
+      _,
+      onPressRetryLoadMore,
+      pal.border,
+      tab,
+    ],
   )
 
   const FeedFooter = React.useCallback(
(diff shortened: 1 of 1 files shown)


src/view/com/notifications/Feed.tsx
```diff
@@ -15,20 +15,29 @@ import {cleanError} from '#/lib/strings/errors'
...
 import {FeedItem} from './FeedItem'
 
 const EMPTY_FEED_ITEM = {_reactKey: '__empty__'}
 const LOAD_MORE_ERROR_ITEM = {_reactKey: '__load_more_error__'}
 const LOADING_ITEM = {_reactKey: '__loading__'}
 
+const tabs: {
+  label: string
+  types?: NotificationType[]
+}[] = [{label: 'All'}, {label: 'Mentions', types: ['reply', 'quote']}]
```
ericclemmons, line 39: 💅 Initially I had a specific `Tab` type, but I inlined it  and used `typeof tabs[number]` below.

src/view/com/notifications/Feed.tsx
```diff
@@ -112,6 +123,13 @@ export function Feed({
     fetchNextPage()
   }, [fetchNextPage])
 
+  const onSelectTab = React.useCallback(
+    (index: number) => setTab(tabs[index]),
+    [setTab],
+  )
+
+  const tabLabels = React.useMemo(() => tabs.map(({label}) => label), [])
```
ericclemmons, line 131: 💅  This can be moved out of the component (under `const tabs`) since it has no deps. _Or_, it could be generated at runtime below:

```js
<TabBar
  items={...}
```

Performance nit for an Array of size 2!

src/view/com/notifications/Feed.tsx
```diff
@@ -141,16 +159,26 @@ export function Feed({
             <NotificationFeedLoadingPlaceholder />
           </View>
         )
+      } else if (!tab.types || tab.types.includes(item.type)) {
```
ericclemmons, line 162: f174623ea42b974b3ad9e2707f3ad9371c8d150d changed my initial implementation that used `const filteredItems = React.useMemo(...)` to not rendering the filtered out types.

I figured since `<Feed item={items}>` already holds this list in memory, duplicating it to `filteredItems` would be wasting memory.

(I dont know if `return null` below negatively affects window virtualization, since approximate size is _even more variable_ now).

🏎️ There could be performance optimization by instead rendering the `FeedItem` and perhaps leveraging [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility).

ericclemmons:
Hmmm, on iOS I noticed 2 things:

1. It's nearly impossible to see the "bottom" of the infinite scroll on the _All_ tab.  However, on _Mentions_, you can fling to the bottom, hit ~20vh of empty space, see a loader, then need to scroll again.

  Loading indicator


  Knowing how this PR works, it's clear to me that this is from the "gaps" of filtered-out types.

2. Switching from _Mentions_ to _All_ has a ~1-3s delay before rendering. **This only happens after scrolling quite a ways, _then_ switching tabs**.  I'm curious if this was faster before f174623ea42b974b3ad9e2707f3ad9371c8d150d?

ericclemmons:
> I'm curious if this was faster before [f174623](https://github.com/bluesky-social/social-app/commit/f174623ea42b974b3ad9e2707f3ad9371c8d150d)?

Nope!  Same performance, but the _scroll position_ is saved when using `filteredItems` 🤨 Switching tabs _after_ f174623ea42b974b3ad9e2707f3ad9371c8d150d resets to the top (as expected).

src/view/com/notifications/Feed.tsx
```diff
@@ -51,6 +60,8 @@ export function Feed({
   const {_} = useLingui()
   const moderationOpts = useModerationOpts()
   const {checkUnread} = useUnreadNotificationsApi()
+  const [tab, setTab] = React.useState<(typeof tabs)[number]>(tabs[0])
```
ericclemmons, line 63: Oooo tab switching can be instant with:

```js
const notificationTypes = React.useDeferredValue(tab.types)
```

But the UI needs an indicator that the render is behind...

gaearon:
Yeah.

To set expectations, we can't merge this as is — there's quite a bit of exploratory work we need to do to see if it's feasible to make it behave nicely via a client-only approach. Right now with a moderate volume of notifications it feels very choppy because very few page fetches result in actually having mentions in them.

ericclemmons:
Sorry, I won't be able to continue this PR without sponsorship.  Property taxes are in and 🤯

For the future, I recommend updating the APIs per my origina suggestions:

- [ ] Add `NotificationType` filter to backend API
- [ ] Update `useNotificationFeedQuery` to support `NotificationType` allowlist
- [ ] Pass `tab.types` to `useNotificationFeedQuery`

gaearon:
Thanks for exploring this! I'm going to try to pick it up from work in https://github.com/bluesky-social/social-app/pull/3612. This PR strays a bit far behaviorally from other tabbed interfaces because we use the `PagerWithHeader` component elsewhere.

surfdude29:
Following on from this, there's now a draft PR open to add a Conversations tab to the Notifications screen:
- #7044

PR #6098, opened 2024-11-04
Add follower/following count to profile lists

hipstersmoothie:
There wasn't an easy way to see this info in app. I also noticed some people were making services just to see this info. It probably makes sense to just show the full count on this page

Mobile

CleanShot 2024-11-03 at 20 49 58@2x

Desktop

CleanShot 2024-11-03 at 20 50 03@2x


surfdude29:
This looks like it will be happening in #7057

Issue #5970, opened 2024-10-28
Convert Starter Packs into Lists


PauloCarrasco:
### Describe the Feature

A nice feature to have is a way to easily convert a starter pack into a list using a button like the "Follow All" on the top of the pack page (see image).

I think lots of people don't want to follow every account in a starter pack, and cherry picking accounts to follow or to manually add to a list is tedious. 

In my case, I like to add accounts associated to a subject to a list and scan that list for a while and then decide to cherry pick some accounts that create interesting content.

Hope my 2 cents of an idea can make the starter pack feature even more functional.

Kudos!

### Attachments

![Screenshot 2024-10-28 at 10 24 36](https://github.com/user-attachments/assets/4f7fdb8d-dab2-4423-b321-7d9617bf1cf3)

yoavg:
an even better version than "convert to list" would be to have "add to list" option from a starter pack. this covers the "convert to list" use case, but also allows more flexibility in list management.

svivian:
Honestly why are these two separate things in the first place? At the core they are identical - a collection of users. In the starter pack you can even look at the posts like a list.

Just make it so that you can either follow as a list, or follow all the accounts within.

PauloCarrasco:
Same concept, different wording, right? The final objective is to have the option to follow some/all users or add them easily to a new/existing list.

jbarciauskas:
Yeah I guess what this implies is that a Starter Pack should be a List plus a list of Feeds, where a List is a set of Accounts. Then you could just have an option to take a Starter Pack's List, or create a Starter Pack from a List (with an empty set of Feeds)

yoavg:
I think that, if one takes the notion that users should be editing and
curating their own lists, then the semantics of "take a Starter
Pack's List" should be "add all the accounts in it to a list" rather than
"convert it to a user-list" (and similarly, when a user "takes" a list from
another user, they should have the ability to "add all accounts" to their
own list, rather than just duplicate the list).

svivian:
> Starter Packs have both a group of `Accounts` and `Feeds`.

What’s the difference between a feed (within a starter pack) and a list? All I see is a list of recent posts by those users, which is exactly the same whether it’s a list or a feed in a starter pack.

monicaellerose:
is there a known way to turn a list into a starter pack?

monicaellerose:
> is there a known way to turn a list into a starter pack?

https://nws-bot.us/bskyStarterPack.php

[...]

Issue #5940, opened 2024-10-25
Proposal: Merge Starter Packs and Lists


knotbin:
Lists and starter packs offer the same functionality - adding accounts to a collection - for different purposes. However, their features are fragmented even though they are very similar. Lists can be pinned to the home feed while starter packs cannot, people can be easily added to lists from their profile, while they can't be easily added to starter packs, lists are searchable while starter packs are not.

This proposal argues for the merging of the two. Starter packs would be removed in favor of lists, which would now have an option to "Share as a Starter Pack"

Current starter packs would be converted to lists.
The only obstacle to this is the fact that lists can contain feeds. Two options for this: 
1. Get rid of feeds in current starter packs, as they are used in very few starter packs, and of the starter packs they are used on, they are a relatively unimportant addition.
2. Allow lists to contain feeds

This proposal will simplify an overcomplicated and confusing lineup of user collection features that are needlessly separated and fragmented. Please let me know your feedback.

wlonkly:
I am a fan of this plan. Lists feel native and dynamic, and starter packs bolted on and one-time, but the popularity of starter packs has been a great boost for the community. 

But I noticed what I think might be an error in the proposal:

> The only obstacle to this is the fact that lists can contain feeds

"that lists cannot contain feeds", right?

svivian:
Lists *are* feeds though. Or, nobody has managed to articulate any difference between them.

wlonkly:
> Lists _are_ feeds though. Or, nobody has managed to articulate any difference between them.

No, they're not really that related. A List is a pre-configured list (sorry) of users, which can be used as a "show me these users' posts" list like lists on Twitter did, or as a block list. 

A feed is an interface where a service, external to the main parts of Bluesky, can apply whatever logic it wants to include certain posts.  That might be "posts from these users", where it works roughly like a list does (but maybe "these users" changes somehow), or it can be based on some qualities of the posts themselves (like "[popular in your network](https://bsky.app/profile/bsky.app/feed/with-friends)" or "[contains the name of a city](https://bsky.app/profile/did:plc:3v55yprp3owcom57wrosvf5i/feed/aaai72gpp25t6)" or "[is getting ratio'd](https://bsky.app/profile/did:plc:pmyqirafcp3jqdhrl7crpq7t/feed/aaapfo6jminv2)".) 

Feeds are more like "pluggable algorithms", but they also require someone to run the software that implements the algorithm on their own server somewhere.

[...]

Issue #5932, opened 2024-10-25
Add new Start Pack menu item "Add to Lists" - Allows adding all users in a Starter Pack, to my own List.


adamatic-me:
### Describe the Feature

I follow more than 1k users. The Following feed already fails me with that number. I also know about feeds which can address this, like Quiet Posters, Best of Follows, etc. 

No, I want more granular control, I need sub-groups. That is why I love my curated Lists. I think that in the future, users will seek each other's lists much more. It could be a bigger thing on Bsky in the future. A user curating a List is superior to algorithmic feeds in some ways.

You guys know this, The other form of user curated lists on Bksy are Starter Packs. They are super popular. People have put a lot of time into creating these.

Let me combine these two powers, please!

In the Starter Pack drop down, please add the "Add to Lists" menu item.

![image](https://github.com/user-attachments/assets/ae04d94b-533c-4c47-b3fc-f8ccd4a613ee)

On click, the user is shown the same Add to List modal, as when clicked on a user, except instead of one user, all Starter Pack users are being manipulated.

![image](https://github.com/user-attachments/assets/14e24203-8138-403b-a682-4c9709d61384)

That's it! I would now have the ability to leverage all the work that Starter Pack creators did, and add their lists to the work that I did to create my own Lists!

Thank you for your time!

### Describe Alternatives

Open a Start Pack's user list, open each user, then click add to list. This is so annoying that I have only done it twice. I have wanted to do it to half of Starter Packs that I have seen.

adamatic-me:
Note on my recent edit: I realized that for a really powerful List management system, also adding the Add to Lists menu item to other Lists would allow me to combine my own Lists, and combine uses from another List into one of my own Lists.

@surfdude29 - mentioning you because you gave this request a like, and I have now changed it. Not sure if you still like it.

PR #6105, opened 2024-11-04
feat(starterpacks): convert to lists

joaomendoncaa:
# Convert Starter Packs to Lists!

image

This tackles #5970 (also #5932, and #5940). It might seem like a minor annoyance, but the number of likes on the issue(s) shows it's a concern for a lot of people. 

Due to its minor impact, I've added the feature tucked into the `OverflowMenu`. The motivation is simple; while there's currently no drawback to unreasonable following numbers, managing that following list will become extremely cumbersome in the future if it's flooded with starterpacks — the issue description goes well into detail on the problem.

Thanks to anyone who takes the time to chime in, and feel free to request any changes. 🍻 

### Feature demo

https://github.com/user-attachments/assets/282b0b00-d095-447b-b112-3e42e2c1132f

### DIFF
--- a/src/screens/StarterPack/StarterPackScreen.tsx
+++ b/src/screens/StarterPack/StarterPackScreen.tsx
@@ -26,7 +26,7 @@ import {getStarterPackOgCard} from '#/lib/strings/starter-pack'
 import {logger} from '#/logger'
 import {isWeb} from '#/platform/detection'
 import {updateProfileShadow} from '#/state/cache/profile-shadow'
-import {CreateOrEditListModal, useModalControls} from '#/state/modals'
+import {useModalControls} from '#/state/modals'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
 import {getAllListMembers} from '#/state/queries/list-members'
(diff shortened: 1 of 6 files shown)


haileyok:
Hey @joaomendoncaa! Thanks for the PR! I'll bring this up next week and see what direction we want to take this.

yoavg:
While this is great, I think the usability would be even better if we can change "convert to list" into "add all to list" (either an existing one or a new one, but for the first version it could be only existing ones).

This will allow more flexibility in how starter-packs are used, e.g. if I have a "tech" list and a "history" list, I could add several starter-packs to each one.

adamatic-me:
> While this is great, I think the usability would be even better if we can change "convert to list" into "add all to list" (either an existing one or a new one, but for the first version it could be only existing ones).
> 
> This will allow more flexibility in how starter-packs are used, e.g. if I have a "tech" list and a "history" list, I could add several starter-packs to each one.

This was my exact thinking when I made this request:  #5932. This required no novel screen design. 

In any case, really glad that someone is working on this! yay!

PR #6107, opened 2024-11-04
Hide 'Follows You' pill on own Followers page

jakebailey:
When looking at your own followers page, every account listed has the "Follows You" pill:

image

That seems a little redundant; after all, I am looking at my own followers page. Of course they all follow me...

This PR hides that pill when viewing your own page, like:

image

I will note that I really thought Twitter did this, and then I looked and it didn't. So, maybe this is a silly idea. Happy to close; it was fun to hack on the app locally! :smile: 

Also, have never worked with react-native, so wasn't sure how I was supposed to pass additional data down to `renderItem`, so just added a param instead of making each item be an object or something.

### DIFF
--- a/src/view/com/profile/ProfileFollowers.tsx
+++ b/src/view/com/profile/ProfileFollowers.tsx
@@ -14,18 +14,22 @@ import {ListFooter, ListMaybePlaceholder} from '#/components/Lists'
 import {List} from '../util/List'
 import {ProfileCardWithFollowBtn} from './ProfileCard'
 
-function renderItem({
-  item,
-  index,
-}: {
-  item: ActorDefs.ProfileViewBasic
-  index: number
-}) {
+function renderItem(
+  {
+    item,
+    index,
+  }: {
+    item: ActorDefs.ProfileViewBasic
+    index: number
+  },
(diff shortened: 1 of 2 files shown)


src/view/com/profile/ProfileFollowers.tsx
```diff
@@ -105,7 +109,7 @@ export function ProfileFollowers({name}: {name: string}) {
   return (
     <List
       data={followers}
-      renderItem={renderItem}
+      renderItem={info => renderItem(info, isMe)}
```
jakebailey, line 112: Not sure if this is the right way to do it, if I need to `useCallback` here to prevent this from changing, or if this method is just "all wrong".

gaearon:
I think for consistency between screens I'd keep the badge. Sure, it's a bit redundant here, but dropping the badge introduces a difference in behavior that imo contributes to mental load a bit.

jakebailey:
That's fair; happy to close this if it feels bad.

I'll also say that it'd be nice if pills weren't positioned in a a way that makes the cards even taller, e.g.

bluesky:

image

twitter:

image

But that's an entirely different kind of PR 😄

PR #6109, opened 2024-11-04
Fix notification time spacing

mozzius:
#5550 switched the spacing around the middot from whitespace characters to margins. Unfortunately this does not work, I'm unsure why - but it would definitely not work with text scaling, so probably best not to use it even if it did work.

This reverts it to just use whitespace again

### DIFF
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -451,8 +451,9 @@ let FeedItem = ({
             <TimeElapsed timestamp={item.notification.indexedAt}>
               {({timeElapsed}) => (
                 <>
-                  <Text style={[a.ml_xs, pal.textLight]}>&middot;</Text>
-                  <Text style={[a.ml_xs, pal.textLight]} title={niceTimestamp}>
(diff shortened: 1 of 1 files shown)


src/view/com/notifications/FeedItem.tsx
```diff
@@ -451,8 +451,9 @@ let FeedItem = ({
             <TimeElapsed timestamp={item.notification.indexedAt}>
               {({timeElapsed}) => (
                 <>
-                  <Text style={[a.ml_xs, pal.textLight]}>&middot;</Text>
-                  <Text style={[a.ml_xs, pal.textLight]} title={niceTimestamp}>
+                  {/* make sure there's whitespace around the middot -sfn */}
```
gaearon, line 454: i think you can `<Text style={[pal.textLight]}>{' '}&middot;{' '}</Text>` to make it explicit if you'd like

gaearon (approved):

src/view/com/notifications/FeedItem.tsx
mozzius: I wrote that - prettier removed it!

PR #6113, opened 2024-11-05
Fix start value in youtube embed by removing "s" from url param

webpro:
Added a youtube.com link (https://bsky.app/profile/webpro.nl/post/3la6ojjnymg2u) and noticed the iframed player didn't start from the expected `t`.

The url https://www.youtube.com/watch?v=U4_X2p6rPJI&t=168s contains the `s` suffix, whereas the `playerVars.start` should be an integer according to https://developers.google.com/youtube/player_parameters.

Should add that I didn't test fully end-to-end locally because of same origin issue and didn't run `server.go` to serve `/iframe/youtube.html`, but added tests.


gaearon (approved):
this looks good to me, thank u

PR #6117, opened 2024-11-05
Use megabytes rather than mebibytes for video limit

mozzius:
Backend uses 50 megabytes (1000 * 1000 * 50) but frontend was using 50 mebibytes (1024 * 1024 * 50)


gaearon (approved):
hm ok

PR #6122, opened 2024-11-05
Set useFetchDid staleTime to 5 minutes

aidenfoxivey:
https://github.com/bluesky-social/atproto/discussions/2909

I was having some issues as referenced above in this Github discussion. I have a suspicion that setting
the stale time as such is causing my issues, since changing the DNS record didn't seem to fix the issue.
If I'm off-base in this PR, please feel free to close or correct it. I sorted my issue out myself with `goat`
but figured it might be a good change to make.

Signed-off-by: Aiden Fox Ivey <aiden@aidenfoxivey.com>

### DIFF
--- a/src/state/queries/handle.ts
+++ b/src/state/queries/handle.ts
@@ -58,7 +58,7 @@ export function useFetchDid() {
   return React.useCallback(
     async (handleOrDid: string) => {
       return queryClient.fetchQuery({
-        staleTime: STALE.INFINITY,
+        staleTime: STALE.MINUTES.FIVE,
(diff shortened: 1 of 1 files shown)


aidenfoxivey:
Thinking about this a bit more, maybe something like 10 seconds is more reasonable.

A given user will likely only click to verify when they have actually changed something. Thus, it would make more sense to consider the old data as stale the moment after the query is first made.

aidenfoxivey:
I should add that I'm not much of a web programmer, but I like the platform and want to help fix bugs. 😃

PR #6123, opened 2024-11-05
Fix thin red circle around Like button on iOS

### LINKED ISSUE #6118: Like button has thin red ring around it on iOS
### Steps to Reproduce

1. Go to your profile.
2. Tap on the **Likes** tab.
3. Tap on any post that you have previously liked.
4. Observe the thin red ring around the **Like** button.

### Attachments

IMG_1982

iOS

### Device Info

iPhone SE (3rd generation), iOS 17.7.1

Build version: 1.92.1.530; Bundle info: 89cff31b6 (prod); Bundle date: 24101917; Platform: ios

### Additional Information

As noted, I only see this issue when looking at a post that I have already liked, it doesn't appear immediately after liking a post.

khuddite:
Fixes #6118 

Summary: We display an animation when the post is liked on the Post screen around `Like` button. However, when the post is already liked, we still display the animation UI with 0 opacity without an animation. That seems problematic on iOS because there is a noticeable thin red circle displayed even though its opacity is set to zero. I think it has something to do with how iOS handles UI rendering or how `react-native-reanimated` library handles keyframes.

Solution: The animation is completely removed from DOM tree when unnecessary. When necessary, it gets mounted and displays the animation as usual. It results in the reported issue being fixed while not impacting the current animation implementation.

Updated behavior:

https://github.com/user-attachments/assets/e32b4aa1-ea36-4b2f-999b-c0bf9e0759d3

### DIFF
--- a/src/lib/custom-animations/CountWheel.tsx
+++ b/src/lib/custom-animations/CountWheel.tsx
@@ -91,15 +91,15 @@ export function CountWheel({
   likeCount,
   big,
   isLiked,
-  isToggle,
+  hasBeenToggled,
 }: {
   likeCount: number
   big?: boolean
   isLiked: boolean
-  isToggle: boolean
+  hasBeenToggled: boolean
 }) {
   const t = useTheme()
-  const shouldAnimate = !useReducedMotion() && isToggle
+  const shouldAnimate = !useReducedMotion() && hasBeenToggled
   const shouldRoll = decideShouldRoll(isLiked, likeCount)
 
   // Incrementing the key will cause the `Animated.View` to re-render, with the newly selected entering/exiting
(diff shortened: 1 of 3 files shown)


layeredstrange:
It's not just iOS.

This is visible on 3 of my Android devices.

haileyok:
@khuddite Thank you! Some of the var names were a bit confusing, took me a bit to understand what was going on here. So I adjusted those and pushed into this branch. Appreciate you digging into this!

khuddite:
> It's not just iOS.
> 
> This is visible on 3 of my Android devices.

Thanks for confirming that, it's fixed on Android as well.

khuddite:
> @khuddite Thank you! Some of the var names were a bit confusing, took me a bit to understand what was going on here. So I adjusted those and pushed into this branch. Appreciate you digging into this!

@haileyok Thanks, new var names make more sense.

surfdude29:
Thank you so much @khuddite and @haileyok for looking into and fixing this! 🙏

A tiny thing but it was kind of annoying me lol

PR #6124, opened 2024-11-05
Remove SCREEN from lightbox layout

gaearon:
I want to get rid of these constants. They're subtly wrong in some cases, and it's very hard to think about the discrepancies due to safe insets. I want to change the code so that we rely on `SafeAreaView` and then everything is derived from its size.

I'm not entirely _sure_ this approach is the way to go, but this constant just sucks so I think it's a step forward.

The approach is to use alternatives to hardcoding `SCREEN`:

- On UI thread, use `measure(safeAreaRef)`. This always gives the height matching the area where we draw the image.
  - Curiously this _seems_ fine to do on every frame. So I'm just doing that unless proven wrong.
    - We might have to do something more fancy when adding orientation handling. But still better than before.
- On JS thread, prefer to use `<SafeAreaView>` where possible.
  - This works well when we don't need to know the value in the actual JS code. The `<SafeAreaView>` code is native.
  - When we absolutely *need* to know the value in JS itself, fall back to `useSafeAreaFrame()`.
    - Note it doesn't have the insets applied (which seems fine for the use case where we have it — measuring iOS width). We could always add `useSafeAreaInsets` if needed and subtract whichever ones we need manually.

## Test Plan

This fixes incorrect positioning on Android device when double-tapping into an image. Verify that double-tapping on main shows a black bar beneath, but double-tapping in this branch does not. (This doesn't repro on emulator for me.)

Verify controls (header, footer) work like before. Verify buttons are clickable. With no alt text, short alt text, and very long alt text. Verify very long alt text expands on tap but the scroll view doesn't overlap the header controls. Verify that controls still work and display correctly when invoking them while zoomed in.

Verify you can pan and pinch into the image. <s>Pinching into the image on device produces occasional flicker on Android, but this problem exists on main too. I'll need to look into that separately.</s> (Fixed in #6126)

It would be good to check what happens with super tall images.

### DIFF
--- a/src/state/lightbox.tsx
+++ b/src/state/lightbox.tsx
@@ -5,7 +5,7 @@ import {nanoid} from 'nanoid/non-secure'
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {ImageSource} from '#/view/com/lightbox/ImageViewing/@types'
 
-type Lightbox = {
+export type Lightbox = {
   id: string
   images: ImageSource[]
   thumbDims: MeasuredDimensions | null
--- a/src/state/lightbox.tsx
+++ b/src/state/lightbox.tsx
@@ -1,10 +1,12 @@
 import React from 'react'
 import type {MeasuredDimensions} from 'react-native-reanimated'
+import {nanoid} from 'nanoid/non-secure'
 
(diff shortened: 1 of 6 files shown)


mozzius (approved):
Bug confirmed fixed, could not see any regressions

PR #6125, opened 2024-11-05
feat: double tapping search should focus the search bar

uragirii:
- Adds functionality to focus the search input when user double presses the search icon on the bottom bar
- Also focuses input on long press


yoleuh:
similar to #5732

uragirii:
> similar to #5732

I'd close this PR then, thanks

PR #6129, opened 2024-11-06
Set Inter to use ss02 disambiguation styleset

benharri:
This adds ss02 to Inter which is the disambiguation set https://rsms.me/inter/#features
It makes ILli1 a lot easier to tell apart.

Haven't been able to get this working on web. Seems that some react-native-web styles are overriding it and the font-feature-settings aren't getting applied correctly.

Looks great on android though

Thread with some more info
https://bsky.app/profile/benharri.org/post/3l7r4sv42sk2q

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -49,6 +49,7 @@
       margin: 0px;
       padding: 0px;
       font-family: InterVariable, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Liberation Sans', Helvetica, Arial, sans-serif;
+      font-feature-settings: 'cv05' on, 'calt' off;
       text-rendering: optimizeLegibility;
       /* Platform-specific reset */
       -webkit-overflow-scrolling: touch;
--- a/src/alf/fonts.ts
+++ b/src/alf/fonts.ts
@@ -80,7 +80,9 @@ export function applyFonts(style: TextStyle, fontFamily: 'system' | 'theme') {
      * Disable contextual alternates in Inter
      * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant}
      */
-    style.fontVariant = (style.fontVariant || []).concat('no-contextual')
+    style.fontVariant = (style.fontVariant || [])
+      .concat('no-contextual')
+      .concat('character-variant-5')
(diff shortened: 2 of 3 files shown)


benharri:
Another less disruptive option is to use `cv05` which changes just the lowercase `l`

Gitlab used that for their version of Inter:
https://www.npmjs.com/package/@gitlab/fonts

character variants aren't in the react native fontVariant though...

gaearon:
this needs a look from @estrattonbailey

estrattonbailey:
Hey sorry for the delay. I like the idea, I tend to prefer these too. Other users may have differing opinions, but I'd be curious to try this. At the very least the lower L variant (I kinda think the I with bars is ugly heh).

I had to timebox this today, got some other stuff to get done this weekend, but did a little poking so I better understand.

It does appear that RNW doesn't support `stylistic-two`, which should map to `ss02` as you've already investigated. Searching around, it looks like [RN itself does some mapping for these](https://github.com/search?q=repo%3Afacebook%2Freact-native%20stylistic-two&type=code) values. I suppose RNW would need to do the same, probably prior to passing the raw values into the `styleq` library it uses internally (also authored by @ necolas). I see you have a PR with the types already, nice.

So it seems the full situation is:
- web needs a local patch in our codebase, or a PR to RNW
- iOS doesn't seem to be working either, might need a reduced test case and issue report in RN repo
- works fine on Android, nice

Not sure if/when we'll get time to dig deeper, but happy to review code or lend an ear if you want to keep looking into this!

benharri:
I think `cv05` is a better target as you mention here which only applies the curly lowercase `l`, but this runs into a similar issue in upstream RN. I'm not seeing any types/wrappers for any of the `cv` character variant font-features.

I will poke around a bit on patching RN and RNW for this when I find some time.

jasontenpenny:
I would love to see this get implemented, though in my mind it's really only important for hyperlinks.

In addition to the LlIi1 disambiguation, I think the O0 disambiguation is important also.

I don't know anything about React so I can't help with those peculiarities, but I know I was able to get this working on a website I'm building by using this CSS code (found in a forum somewhere): `font-feature-settings: 'liga' 1, 'ss02' 1, 'calt' 1;`.

PR #6135, opened 2024-11-07
[Lightbox] New dismiss gesture

gaearon:
Previously, iOS used a scroll listener on the ScrollView to determine the "swipe the image away" gesture. On Android, we had a gesture recognizer. This setup had a few downsides:

- The thresholds felt inconsistent between platforms
- The "fly away" gesture animation didn't really make sense
  - It was not "flying away" on iOS at all, just abruptly disappearing
  - We were changing the image opacity against a black background, whereas usually an animation like this changes the _backdrop_ opacity against the app itself
- The way the ScrollView was laid out on iOS caused dark bars to be part of the scrollable area when zoomed in

This is a rewrite of that gesture. I'm lifting it up to the parent component (which lets me animate the backdrop and controls based on it). I'm also adding a "flying away" state which waits for the image to go offscreen before unmounting the lightbox.

## Test Plan

Verify you can zoom in and out of images (with pinch and/or double-tap). Verify panning works.

Notice that on iOS, panning while zoomed in no longer includes the black bars (!)

Verify that you can flick the image away while it's not zoomed in. Verify you can navigate between images and flick them away. Verify you can flick in either direction, and that the animation doesn't lock up. Verify that if you just release the finger, it'll snap back in place. Try to spam gestures pretty hard and verify nothing gets stuck or confused. Especially galleries.

Verify that supertall images like https://bsky.app/profile/schlagteslinks.bsky.social/post/3l7y4l6yur72e don't mess up the layout.

https://github.com/user-attachments/assets/6bc2f0f9-cd2a-43e4-8a4a-3a922b596878

https://github.com/user-attachments/assets/36f5e752-3c40-4b7e-a3f0-5d1eb09b3355

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -1,6 +1,10 @@
 import React, {useState} from 'react'
-import {ActivityIndicator, StyleSheet, View} from 'react-native'
-import {Gesture, GestureDetector} from 'react-native-gesture-handler'
+import {ActivityIndicator, StyleProp, StyleSheet, View} from 'react-native'
+import {
+  Gesture,
+  GestureDetector,
+  PanGesture,
+} from 'react-native-gesture-handler'
 import Animated, {
   AnimatedRef,
   measure,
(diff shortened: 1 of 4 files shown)


mozzius (approved):
Tested as much as I could on the simulator. I'll test spamming gestures when its on device

Feels great! Big improvement

PR #6136, opened 2024-11-07
[Lightbox] Add border radius to avatars

gaearon:
Stacked on #6135

---

Self-explanatory.

On the web, I had to screw with the layout a bit so that it doesn't blow up the avi full screen cause that was also quite strange. Normal pictures display like before.

## Test Plan

https://github.com/user-attachments/assets/d0f2a985-d173-4aee-b7bc-2d843a93825f

https://github.com/user-attachments/assets/36e7d85c-3960-40e9-8a62-22cc5b963783

https://github.com/user-attachments/assets/fc7981ff-2f7a-42ac-83fa-81219c9e13c1

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -64,7 +64,7 @@ let ProfileHeaderShell = ({
               height: 1000,
               width: 1000,
             },
-            shape: 'circle',
+            type: 'circle-avi',
           },
         ],
         index: 0,
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -64,6 +64,7 @@ let ProfileHeaderShell = ({
               height: 1000,
               width: 1000,
             },
+            shape: 'circle',
           },
         ],
         index: 0,
--- a/src/view/com/lightbox/ImageViewing/@types/index.ts
+++ b/src/view/com/lightbox/ImageViewing/@types/index.ts
@@ -21,5 +21,5 @@ export type ImageSource = {
   thumbUri: string
   alt?: string
   dimensions: Dimensions | null
-  shape: 'circle' | 'rect' | 'rounded-rect'
+  type: 'image' | 'circle-avi' | 'rect-avi'
 }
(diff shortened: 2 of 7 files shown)


mozzius (approved):

cyrneko:
Actually, I'm not sure I agree with this change

Many will use avatars by artists, which often place watermarks in the corner
As a result, many will not explicitly mention artists in their bios - even when they should, being honest - because they know there is a watermark

The information is not lost as you can expand the image to see potential watermarks, which may be hidden with this change, even more so if they are crammed into a corner

gaearon:
Thanks for raising the concern! My arguments for this change are:

- Consistency with expectations from other social products
- Consistency within our own UI
- Enabling a further level of polish (animation)
- The full picture is still available
- This wasn't an intentional decision in the first place

Expanding on these:

1. Most other social products I know of with a "click to expand" interaction expands circle avatars into circles. This means that making this change would at least _not_ violate user expectations — and at best would fix something that does. (One notable exception to that is Facebook. However, Facebook expands avatars into _full original photos_ — they're high quality, not cropped to square, have a description, their own "likes"/"comments"/etc. In other words, they're actual posts. This is not what we're dealing with here, although something like this might be nice to add?)
3. From the in-product consistency perspective, the current behavior also feels inconsistent — avatars are represented as circles everywhere else in the user interface. Even when you choose it, it appears cropped to a circle. I associate my online presence with my circular avatar (like on other social products) and I found the default expansion a bit intrusive.
4. We can't (easily and performantly) animate from a circle to a square. It kinda sucks to not having the animation here. Again, other social products, especially on mobile, do have an animation for expanding. It's a pretty standard UX.
5. This isn't removing the ability to view the full picture for someone who's motivated, "Save Image" still works.
6. If we had not made this omission in the first place, it would not be a topic of conversation — so it feels unfortunate to consider it now as a blocker. It was not something intended to be relied on — just an unpolished screen.

I *do* agree that the part about attribution is interesting. I think having some native way to tag a creator would be a better way to address that. As you noted, people do that in the bio, but maybe it could be a part of this screen in the future. However, I don't think solving this should be a blocker for landing this change.

Let's give this a try?

Issue #6112, opened 2024-11-05
Proposal: Be smarter about quality level when converting images to JPGs


Novaenia:
When the Bluesky app converts an image to a JPG, it first starts at a quality level of 100%. If the resulting file is above the file size limit of 1MB, it drops the quality by 10% and tries again.
https://github.com/bluesky-social/social-app/blob/ab4e36115b390cb69b2dcd31e340898afcbef22a/src/lib/media/manip.ts#L181-L183
Each quality drop has diminishing returns - it is not linear. Often, going from 100% to 98-99% can shave off 25 to 30% of the size saved at 100% quality. Even just around 96% can result in a file that's less than half of the 100% size. Jumping from 100% to 90% sacrifices an unnecessary amount of quality. [Here's a chart from Wikipedia.](https://en.m.wikipedia.org/wiki/File:Quality_comparison_jpg_vs_saveforweb.jpg)

I think this loop should be changed, so that every single quality level from 100% down to 100% is tested.

Sidenote: The <2000x2000 & <1MB constraint on images sucks anyway! It would be swell [if it were to be raised](https://github.com/bluesky-social/atproto/discussions/2887).

PR #6137, opened 2024-11-07
Optimize doResize method with binary search 

JustSNguyen:
This pull request addresses issues discussed in #6112 . Currently, if an uploaded image exceeds a certain size threshold, BlueSky reduces the image quality by 10% and checks if the new size falls within the threshold. If not, BlueSky repeats this step until the image quality reaches 0. There are two issues with this approach:

1. As mentioned in #6112, even a 1-2% decrease in quality can significantly reduce the image size, so the 10% decrement is somewhat overkill.
2. The current solution could be optimized, especially for larger images. For these, the current approach requires multiple iterations to reach the target size, which is not ideal.

My proposed solution is to use binary search to find the maximum quality level that keeps the image size within the limit. This approach will prevent too much loss in image quality. Not only that, I also ran some benchmarks, and it appears that the current solution may perform slightly better for smaller images. However, for larger images, the binary search method does much better. That said, I only conducted a few tests here, so additional testing may be needed to confirm the difference.

| Image Size (KB) | Current Method (ms) | Binary Search Method (ms) |
|-----------------|---------------------|----------------------------|
| 1148            | 844                 | 1040                       |
| 1911            | 5454                | 6477                       |
| 3011            | 5530                | 3396                       |
| 3230            | 6079                | 5402                       |
| 10609           | 12210               | 4141                       |
| 14765           | 26390               | 10213                      |

### DIFF
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -178,15 +178,20 @@ async function doResize(localUri: string, opts: DoResizeOpts): Promise<Image> {
     height: imageRes.height,
   })
 
-  for (let i = 0; i < 9; i++) {
-    // nearest 10th
-    const quality = Math.round((1 - 0.1 * i) * 10) / 10
+  let minQualityPercentage = 0
+  let maxQualityPercentage = 101 // exclusive
+  let newDataUri
+
+  while (maxQualityPercentage - minQualityPercentage > 1) {
+    const qualityPercentage = Math.round(
+      (maxQualityPercentage + minQualityPercentage) / 2,
+    )
     const resizeRes = await manipulateAsync(
       localUri,
       [{resize: newDimensions}],
       {
         format: SaveFormat.JPEG,
-        compress: quality,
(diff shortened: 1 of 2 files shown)


gaearon:
This makes sense. Let's do it on all platforms?

JustSNguyen:
I have updated the PR to also support this for the mobile version @gaearon.

auroursa:
Excellent! Uploading photos taken with an iPhone（obviously compression is needed） clearly shows that the new algorithm improves image quality noticeably, with a shorter processing time as well.

gaearon:
This is changing `lib/media/manip` but I'm seeing some code go through a similar path in `src/state/gallery`. Would you mind checking how it works today? It seems like maybe it's forked between the image picker vs inline camera, but I'm not sure.

seekified:
Any news on this? Pretty high on a lot of people's wishlist imo.

mozzius:
Merged in #7490

Issue #6133, opened 2024-11-06
intent/compose & line break


jycouet:
### Steps to Reproduce

I'm trying to have 2 lines, like this when I prepare a message `https://bsky.app/intent/compose?text=`
```
line1
line2
```

But on Windows & Mac doens't have the same behavior.

## On Windows 

✅ `<br>` https://bsky.app/intent/compose?text=line1%3Cbr%3Eline2
```
line1
line2
```
❌ `\n` https://bsky.app/intent/compose?text=line1%0Aline2
```
line1 line2
```
✅ `<br>\n` https://bsky.app/intent/compose?text=line1%3Cbr%3E%0Aline2
```
line1
line2
```

## On Mac

❌ `<br>` https://bsky.app/intent/compose?text=line1%3Cbr%3Eline2
```
line1<br>line2
```
✅ `\n` https://bsky.app/intent/compose?text=line1%0Aline2
```
line1
line2
```
❌ `<br>\n` https://bsky.app/intent/compose?text=line1%3Cbr%3E%0Aline2
```
line1<br>
line2
```

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Windows & Mac

### What version of the app are you using?

Build version: 1.92.1; Bundle info: dfbfcdd (prod); Bundle date: 24110420; Platform: web

### Additional Information

Like to https://github.com/jycouet/jyc.dev/issues/2

saurabhdaware:
For me `\n` does not work in mac. `<br />` does 🤔 

Although `<br />` does not work when it's being shared to mobile app

dubzzz:
I tried to use `<br/>` but actually while being well displayed on the message. It does not submit properly.

This URL: https://bsky.app/intent/compose?text=Just%20solved%20the%20%23AdventOfPBT%20puzzle%20made%20by%20https%3A%2F%2Ffast-check.dev%2F%3Cbr%2F%3E%3Cbr%2F%3E%F0%9F%91%89%20Join%20the%20challenge%3A%20https%3A%2F%2Ffast-check.dev%2Fblog%2F2024%2F12%2F02%2Fadvent-of-pbt-day-2%2F

Appears nicely in the message prompt:

![image](https://github.com/user-attachments/assets/b5af6b01-03a6-4275-a9a9-a9fba513a407)

But the final message is:

![image](https://github.com/user-attachments/assets/3447a629-c431-4c0c-8b74-928748832336)

jycouet:
[...]

PR #6138, opened 2024-11-07
Fix line breaks in intent/compose URL

bigint:
Related to #6133


jycouet:
On windows browsers `\n` doesn't go to the next time :/

Issue #5722, opened 2024-10-11
System font stack doesn't actually use the system font


byemc:
### Steps to Reproduce

1. Set the system font to something non-standard
2. Go to "Settings" > "Appearance"
3. Change the "Font" setting to "System"
4. Check what font is being used (Firefox lets you see this in its devtools)

### Attachments

Bluesky using the installed font Segoe UI instead of the system font IBM Plex Sans: ![Screenshot showing Bluesky using Segoe UI as the "System" font instead of IBM Plex Sans, which is what is actually set as the system font.](https://github.com/user-attachments/assets/d11a2600-3d8e-470d-8da1-2a85f8d0ed81)

A boilerplate website showing that my browser can use the system font: 
![A webpage reading "Hello World! This is a paragraph" with Firefox devtools showing IBM Plex Sans as the font, under the name "system-ui"](https://github.com/user-attachments/assets/2e0c3957-d108-4967-8481-eb1dcfc40be1)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Florp 11.19 (Firefox 128 ESR) on Arch Linux using Plasma 6.2

### What version of the app are you using?

Build version: 1.92.0; Bundle info: 3b44ea2 (prod); Bundle date: 24101017; Platform: web

### Additional Information

You can select the system font in CSS using the generic font name [`system-ui`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#system-ui), which has been available in all major browsers since about 2021. Currently Bluesky uses the following as its "System" font stack: `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Liberation Sans", Helvetica, Arial, sans-serif`. Ideally `system-ui` would be the first item here.

PR #6139, opened 2024-11-07
Remove system-ui in font family

quiple:
The `system-ui` should be removed as it outputs the default font for the OS language, regardless of the app language.

Example on Korean Windows 11:
| Before | After |
|-|-|
|  |  |

Typically, Korean fonts don't contain Japanese kanji, etc., so you'll see a mix of Malgun Gothic and Meiryo.

Removing `system-ui` will output a language-specific font.

FYI:
<https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-family#system-ui>
<https://infinnie.github.io/blog/2017/systemui.html>
<https://github.com/pnpm/pnpm.io/pull/446>
<https://github.com/picturepan2/spectre/issues/666>
<https://github.com/facebook/docusaurus/issues/11727>
<https://github.com/facebook/docusaurus/pull/12125>


auroursa:
My system language is set to Chinese. After I removed “system-ui,” set the application language to Japanese, and reloaded the page, the UI did correctly use a Japanese font. But I noticed a minor issue: when the post content is in Chinese, it also tries to render it with the Japanese font, which doesn’t look as good as it did previously (even though the previous way was wrong).

Perhaps additional adjustments might help? For instance, matching the font language to the post content language could help prioritize the correct font. I’ve noticed that Twitter seems to handle this, but I don’t fully understand how complex this might be to implement.

surfdude29:
I don't know what the best solution is here, but FYI `system-ui` was added in #5749 to fix #5722.

quiple:
@auroursa 
It needs to set the `lang` attribute on the post content, which doesn't seem to be implemented yet. #1404 #2525

@surfdude29 
It would be nice if it didn't specify a specific font name in `font-family`...

haileyok:
Going to leave this one for @estrattonbailey as I think he has the most context on it.

estrattonbailey:
@quiple thanks for this 🙏  I'm not exactly clear on what needs to happen here tbh.

1. If we remove `system-ui`, which font-family definition is actually being used?
2. I noticed you re-ordered the definition as well. Is the key thing here falling back to `sans-serif` more readily?

quiple:
@estrattonbailey
* `system-ui`: Outputs system default fonts. `lang` attribute is ignored on Windows.
* `-apple-system, BlinkMacSystemFont`: Outputs language-specific system font on macOS.
* `sans-serif`: Outputs font set in the web browser. If a different language is specified in the `lang` attribute, outputs language-specific system font.

I think it can distinguish 2 cases.

1. If ignoring user-customized system/browser fonts:  
   `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";`
2. If respecting user-customized system/browser fonts (as desired in #5722):  
   `sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";`

Additionally, I've added an emoji font stack based on [tailwind's](https://tailwindcss.com/docs/font-family).

quiple:
The `system-ui` causes issues on all Chinese, Japanese, and Korean Windows. picturepan2/spectre#666 w3c/csswg-drafts#11074

auroursa:
I believe this PR should be continued, as it improves the font rendering for Korean, Japanese, and Chinese. I can reproduce this on Chrome with Chinese locale.

quiple:
@estrattonbailey Could you please review this?

> @estrattonbailey
> 
> * `system-ui`: Outputs system default fonts. `lang` attribute is ignored on Windows. Should not be used.
> * `-apple-system, BlinkMacSystemFont`: Outputs language-specific system font on macOS.
> * `sans-serif`: Outputs font set in the web browser. If a different language is specified in the `lang` attribute, outputs language-specific system font.
> 
> I think it can distinguish 2 cases.
> 
> 1. If ignoring user-customized system/browser fonts:
>    `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";`
> 2. If respecting user-customized system/browser fonts (as desired in [System font stack doesn't actually use the system font #5722](https://github.com/bluesky-social/social-app/issues/5722)):
>    `sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";`
> 
> Additionally, I've added an emoji font stack based on [tailwind's](https://tailwindcss.com/docs/font-family).

PR #6141, opened 2024-11-07
Increase memory on iOS

mozzius:
Adds increased memory limit entitlement

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -100,6 +100,7 @@ module.exports = function (config) {
         },
         entitlements: {
           'com.apple.developer.kernel.increased-memory-limit': true,
+          'com.apple.developer.kernel.extended-virtual-addressing': true,
           'com.apple.security.application-groups': 'group.app.bsky',
         },
         privacyManifests: {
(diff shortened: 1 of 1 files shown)


gaearon (approved):
let's try it

app.config.js
```diff
@@ -99,6 +99,8 @@ module.exports = function (config) {
           dark: DARK_SPLASH_CONFIG,
         },
         entitlements: {
+          'com.apple.developer.kernel.increased-memory-limit': true,
+          'com.apple.developer.kernel.extended-virtual-addressing': true,
```
hirbod, line 103: I don't think you need this, but it doesn't hurt

hirbod:
In case you wanna bump Android too, here is a config plugin (I highly recommend it):

```js
const { withAndroidManifest } = require('@expo/config-plugins')

module.exports = function withAndroidLargeHeap(config) {
  return withAndroidManifest(config, async (config) => {
    config.modResults.manifest.application[0].$['android:largeHeap'] = 'true'

    return config
  })
}
```

mozzius:
@hirbod I found we already have it! https://github.com/bluesky-social/social-app/blob/main/plugins/withAndroidManifestPlugin.js

PR #6143, opened 2024-11-07
Delete some unused code

mozzius:
# Based on #6144

I cannot find anywhere that imports these, so they should be safe to delete


src/components/dms/dialogs/TextInput.tsx
```diff
@@ -1 +0,0 @@
-export {BottomSheetTextInput as TextInput} from '@discord/bottom-sheet/src'
```
estrattonbailey: This had keyboard handling built in, at least for the original sheet implementation. Unsure of Discord's.

estrattonbailey: Like, maybe we _should_ be using this

mozzius: Yeah but this is for the JS-based bottom sheets, aka modals. No longer relevent with the new native ones

PR #6144, opened 2024-11-07
[Settings] Ungate, and remove old settings

mozzius:
Deletes all the old pages

# Test plan

I think CI should catch anything I missed. Still, check the new pages work, and that I didn't miss any files


haileyok (approved):
works on web :+1:

PR #5038, opened 2024-08-30
Add `dataSet` to `Text`, tweak metrics count tooltips on profiles

estrattonbailey:
CleanShot 2024-08-30 at 11 21 08@2x

PR #6146, opened 2024-11-07
Show full follower/following/post count metrics in tooltips

DavidBuchanan314:
Before: no way to see non-truncated metrics!

After: full metrics are visible on hover, like so (cursor not visible in screenshot)
image

I've only tested this on desktop firefox, and I assume it doesn't do anything on mobile at all (maybe you can tap and hold? idk)

### DIFF
--- a/src/screens/Profile/Header/Metrics.tsx
+++ b/src/screens/Profile/Header/Metrics.tsx
@@ -20,6 +20,10 @@ export function ProfileHeaderMetrics({
   const {_, i18n} = useLingui()
   const following = formatCount(i18n, profile.followsCount || 0)
   const followers = formatCount(i18n, profile.followersCount || 0)
+  const postCount = formatCount(i18n, profile.postsCount || 0)
+  const followingFull = Intl.NumberFormat().format(profile.followsCount || 0)
+  const followersFull = Intl.NumberFormat().format(profile.followersCount || 0)
+  const postCountFull = Intl.NumberFormat().format(profile.postsCount || 0)
   const pluralizedFollowers = plural(profile.followersCount || 0, {
     one: 'follower',
     other: 'followers',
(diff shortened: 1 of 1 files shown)


src/screens/Profile/Header/Metrics.tsx
```diff
@@ -28,16 +32,21 @@ export function ProfileHeaderMetrics({
...
   })
+  const pluralizedPostCount = plural(profile.postsCount || 0, {
+    one: 'post',
+    other: 'posts',
+  })
 
   return (
     <View
-      style={[a.flex_row, a.gap_sm, a.align_center, a.pb_md]}
+      style={[a.flex_row, a.gap_sm, a.align_center, a.pb_md, {zIndex: 1}]}
```
DavidBuchanan314, line 42: zIndex:1 was required to make the tooltip appear on top of the bio text below, idk if this is the "correct" place to put it

src/screens/Profile/Header/Metrics.tsx
```diff
@@ -47,17 +56,18 @@ export function ProfileHeaderMetrics({
...
-      <Text style={[a.font_bold, t.atoms.text, a.text_md]}>
-        {formatCount(i18n, profile.postsCount || 0)}{' '}
-        <Text style={[t.atoms.text_contrast_medium, a.font_normal, a.text_md]}>
-          {plural(profile.postsCount || 0, {one: 'post', other: 'posts'})}
-        </Text>
+      <Text
+        style={[t.atoms.text_contrast_medium, a.text_md]}
+        title={postCountFull}>
+        <Text style={[t.atoms.text, a.font_bold, a.text_md]}>{postCount} </Text>
+        {pluralizedPostCount}
```
DavidBuchanan314, line 70: I had to invert the nesting order of these <Text> elements so that the outer one was non-bold, and the inner one was bold, so that the tooltip would be non-bold when attached to the outer element. Hopefully I didn't break anything in the process.

DavidBuchanan314:
I think the styling of the tooltips could be improved a little (hairline border? shadow?), but I'm not sure where else they're used and didn't want to break anything. (existing styling is in src/style.css)

surfdude29:
See also #5038

DavidBuchanan314:
> See also https://github.com/bluesky-social/social-app/pull/5038

oops!

src/screens/Profile/Header/Metrics.tsx
haileyok:
```suggestion
      style={[a.flex_row, a.gap_sm, a.align_center, a.pb_md, a.z_10]}
```

Issue #5961, opened 2024-10-27
Text doesn't appear in full in starter pack


SpriterGors:
### Steps to Reproduce

1. On mobile, click on a starter pack
2. scroll to the end
3. notice that the text is cut and can't be scrolled up to be read.

### Attachments

Screenshot of my phone with the last text completely cut by the app buttons. It is impossible to scroll up.

![image](https://github.com/user-attachments/assets/e823d915-dea0-4e3f-a475-7cd8a0fcf697)

### What platform(s) does this occur on?

Android

### Device Info

Android 10, Samsung SM J6000GT

### What version of the app are you using?

Version 1.92.1.355 89cff31b6 (prod)

khuddite:
I am looking into the issue

khuddite:
![image](https://github.com/user-attachments/assets/4c3ac4f2-8f25-4317-af10-5c3d3e907129)

I could not recreate the issue on my phone or any Android emulators. It seems the issue is device-specific.

SpriterGors:
Thanks for looking for me.

I looked at other starter packs with this same device but they all behaved normally. That is, until I looked at the starter pack in OP again, and lo and behold, the problem is there:

Screenshot with user "jackal chan" cut off. It is impossible to scroll down.
![Sem título](https://github.com/user-attachments/assets/b3201aeb-a86b-4536-9054-6e4bf9d653c3)

This is the starter pack link: https://go.bsky.app/FKgUvos

This is what should've appeared under "jackal chan's" bio:
![image](https://github.com/user-attachments/assets/99243911-c3ea-41d4-a884-9bc4c9265b2c)

This is the starter pack's title. Could it be that the emoji in the list title is causing this issue?
![Sem título](https://github.com/user-attachments/assets/394bc555-1d6f-4836-a12c-56cc65fcf12c)

khuddite:
@SpriterGors Thanks for providing additional details.

I can reproduce the same issue on my device with this starter pack. It seems to be specific to this starter pack.

I will investigate what's going on there and fix it anyway.

PR #6152, opened 2024-11-07
Fix starter pack profiles list being cut off

khuddite:
Fixes: #5961 

- Summary
   The issue was the profiles list being cut off on some devices with specific starter packs. Actually, It is not specific to the devices, but it occurs whenever the starter pack header section is tall enough. So, in the current code, the bottom padding of the list is derived from an arbitrary number (300), which means it can be either bigger or smaller than necessary.
   
- Solution
    I updated the code so that the bottom padding is derived from the header height rather than an arbitrary number like 300. This ensures that the profiles list always has a consistent bottom space across all devices(30px). 
    
- Before
    Android - 
    [android-before.webm](https://github.com/user-attachments/assets/f8bb5535-4eae-462e-91d8-a1fb7b28f64c)

    iOS - 
    https://github.com/user-attachments/assets/3560756d-a6a4-4aea-a754-b8e4e99c0e5b

- After
    Android - 
    [android-after.webm](https://github.com/user-attachments/assets/d9a7a37d-00dd-4475-a5cf-76823056bc07)

    iOS - 
    https://github.com/user-attachments/assets/b92fa574-5d93-4e2b-bc3c-b882a1ce9f91

### DIFF
--- a/src/components/StarterPack/Main/ProfilesList.tsx
+++ b/src/components/StarterPack/Main/ProfilesList.tsx
@@ -40,7 +40,7 @@ export const ProfilesList = React.forwardRef<SectionRef, ProfilesListProps>(
     ref,
   ) {
     const t = useTheme()
-    const bottomBarOffset = useBottomBarOffset(300)
+    const bottomBarOffset = useBottomBarOffset(headerHeight)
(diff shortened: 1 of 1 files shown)


gaearon (approved):
lgtm

PR #6153, opened 2024-11-07
[Statsig] Remove client downsampling

haileyok:
## Why

We're moving downsampling to the server. Let's remove everything from the client that has to do with downsampling.

## Test Plan

It builds, it sends events.


gaearon (approved):
sounds good but maybe stacked with the change to use the proxy?

haileyok:
yep that's what i think too. also wondering if we want to just leave `:sampled` in the name for events we are sampling, so you can tell just from looking at this codebase, rather than having to reference the service.

PR #6157, opened 2024-11-08
Improve chat performance

mozzius:
Incorrect use of worklet functions in gestures was causing upwards of 7 seconds of delay when rendering chat items

Also fixed accessing UI thread values in render (in hook deps)

# Before

Screenshot 2024-11-08 at 03 14 09

# After

Screenshot 2024-11-08 at 03 12 39

# Test plan

Confirm chats open 2x as fast

Confirm gestures still work (long press + double tap)


gaearon (approved):
fantastic

PR #6159, opened 2024-11-08
[Lightbox] Open animation

gaearon:
- [x] Check why it doesn't work on Samuel's posts
  - [x] Looks like now it OOM's :(
    - [x] Look for leaks?
- [x] Flicker on https://bsky.app/profile/steveklabnik.com/post/3lafytssuhj27
- [x] iOS avi first frame jumps
- [x] I think I got stuck on iOS once
  - [x] Repro: quick open and close many times 
- [x] Animate tall images correctly

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -59,6 +59,7 @@ let ProfileHeaderShell = ({
           {
             uri: profile.avatar,
             thumbUri: profile.avatar,
+            thumbRect: null,
             dimensions: {
               // It's fine if it's actually smaller but we know it's 1:1.
               height: 1000,
@@ -68,7 +69,6 @@ let ProfileHeaderShell = ({
           },
         ],
         index: 0,
-        thumbDims: null,
       })
     }
   }, [openLightbox, profile, moderation])
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -1,5 +1,12 @@
 import React, {memo} from 'react'
 import {StyleSheet, TouchableWithoutFeedback, View} from 'react-native'
+import Animated, {
+  measure,
+  MeasuredDimensions,
+  runOnJS,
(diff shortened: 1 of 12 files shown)


src/screens/Profile/Header/Shell.tsx
```diff
@@ -68,10 +76,22 @@ let ProfileHeaderShell = ({
...
+    },
+    [openLightbox],
+  )
+
+  const onPressAvi = React.useCallback(() => {
+    const modui = moderation.ui('avatar')
+    const avatar = profile.avatar
+    if (avatar && !(modui.blur && modui.noOverride)) {
+      runOnUI(() => {
+        'worklet'
```
haileyok, line 89: just a question. is this one of those cases now where it will complain if you don't add in `worklet`?

src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -84,10 +91,22 @@ export function ProfileSubpageHeader({
...
         index: 0,
-        thumbDims: null,
       })
+    },
+    [openLightbox],
+  )
+
+  const onPressAvi = React.useCallback(() => {
+    if (
+      avatar // TODO && !(view.moderation.avatar.blur && view.moderation.avatar.noOverride)
```
haileyok, line 101: probably can remove this todo?

haileyok (approved):
from testing on simulators, this looks great!

src/screens/Profile/Header/Shell.tsx
gaearon: I believe so; it's only automatically added for reanimated's hooks

src/view/com/profile/ProfileSubpageHeader.tsx
gaearon: don't have context on it so don't wanna touch it. can do but separate PR?

gaearon:
Pushed a fix for supertall images. Should be good now.

PR #6164, opened 2024-11-08
show precise count in label/hover text

ljharb:
The intention is to expose the precise number of followers somewhere. Currently I have "1.1K", but until that drops to "1K" or rises to "1.2K", i'll have no idea where in that hundred-range the number falls.

I'm sure I'll need to do something with tests, and maybe still format the count in a less-approximate way - please give me some pointers since I'm very unfamiliar with the codebase.

I'd also like to expand this PR to show the full count on the `/follows` and `/followers` pages so it's easily visible on mobile, but i thought i'd start here :-)

### DIFF
--- a/src/screens/Profile/Header/Metrics.tsx
+++ b/src/screens/Profile/Header/Metrics.tsx
@@ -37,7 +37,7 @@ export function ProfileHeaderMetrics({
         testID="profileHeaderFollowersButton"
         style={[a.flex_row, t.atoms.text]}
         to={makeProfileLink(profile, 'followers')}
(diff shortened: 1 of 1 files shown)


mozzius:
We had a PR somewhere that showed the exact count on hover as a tooltip - I'll see if I can dig it up

surfdude29:
> We had a PR somewhere that showed the exact count on hover as a tooltip - I'll see if I can dig it up

There are two that I'm aware of: #6146 and #5038

ljharb:
awesome, if one of those is ready, then let's land one! :-) i'm attached to the capability, not to any given implementation.

PR #6166, opened 2024-11-08
Fix stuck lightbox

gaearon:
Fixes a regression I introduced in https://github.com/bluesky-social/social-app/pull/6135.

See commit by commit.

This is tricky to reproduce, but if we got `onEnd` (which "flying away") while `dismissSwipeTranslateY.value === 0`, the decay animation ("flying away") would not even start. This looks like a Reanimated bug. Simply shifting the initial value by pixel does the job.

Other commits are minor improvements.

## Test Plan

The easiest way to repro is to actually comment out `dismissSwipeTranslateY.value = e.translationY` in `onUpdate` or to hardcode it to `dismissSwipeTranslateY.value = 0`. Before the hack/fix, this would freeze up the lighbox. Now, it works.

### DIFF
--- a/src/view/com/lightbox/ImageViewing/index.tsx
+++ b/src/view/com/lightbox/ImageViewing/index.tsx
@@ -160,6 +160,23 @@ function ImageView({
     }
   }, [])
 
+  useAnimatedReaction(
+    () => {
+      const screenSize = measure(safeAreaRef)
+      return (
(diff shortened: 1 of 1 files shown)


mozzius (approved):
lgtm!

TUNISIA-user:
lgtm!

PR #6175, opened 2024-11-08
Switch label buttons to checkbox from radio

haileyok:
## Why

Radio buttons no. Check boxes yes. Nothing more to say.

## Test Plan

Things work as they did before.


mozzius (approved):
lgtm!

PR #6184, opened 2024-11-09
Follower sorting

hipstersmoothie:
I think putting followers at the top and bottom in certain situations leads to easier to use UI. I was initially going to do separate PRs but realized that I needed the sorting function. I applied it to some places where I think it makes sense. There may be more

> Note: For examples I follow luna

1. Starter packs get followed people sorted to the bottom

CleanShot 2024-11-09 at 01 45 16@2x

2. When mention a user, followed people are at the top

CleanShot 2024-11-09 at 01 46 28@2x

3. Mobile searches sort followed people to the top

CleanShot 2024-11-09 at 01 47 29@2x

4. Desktop searches sort followed people to the top

CleanShot 2024-11-09 at 01 48 08@2x

### DIFF
--- a/src/lib/functions.ts
+++ b/src/lib/functions.ts
@@ -1,3 +1,5 @@
+import {AppBskyActorDefs} from '@atproto/api'
+
 export function choose<U, T extends Record<string, U>>(
   value: keyof T,
   choices: T,
@@ -95,4 +97,20 @@ export function isPlainObject(o: any): o is Object {
 
 function hasObjectPrototype(o: any): boolean {
   return Object.prototype.toString.call(o) === '[object Object]'
+}
+
+export function sortProfiles(
+  first: AppBskyActorDefs.ProfileView,
+  second: AppBskyActorDefs.ProfileView,
+) {
+  const isFollowingFirst = Boolean(first.viewer?.following)
+  const isFollowingSecond = Boolean(second.viewer?.following)
+
+  if (isFollowingFirst && !isFollowingSecond) {
+    return -1
+  } else if (!isFollowingFirst && isFollowingSecond) {
+    return 1
+  }
+
+  return 0
 }
--- a/src/view/com/composer/text-input/web/Autocomplete.tsx
+++ b/src/view/com/composer/text-input/web/Autocomplete.tsx
@@ -5,6 +5,7 @@ import React, {
   useState,
 } from 'react'
 import {Pressable, StyleSheet, View} from 'react-native'
+import {AppBskyActorDefs} from '@atproto/api'
 import {Trans} from '@lingui/macro'
 import {ReactRenderer} from '@tiptap/react'
 import {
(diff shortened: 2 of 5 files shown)


src/lib/functions.ts
```diff
@@ -96,3 +98,19 @@ export function isPlainObject(o: any): o is Object {
 function hasObjectPrototype(o: any): boolean {
   return Object.prototype.toString.call(o) === '[object Object]'
 }
+
+export function sortProfiles(
```
hipstersmoothie, line 102: Wasn't sure where to put this function.

mary-ext:
are you sure 2-3 is actually necessary? in all my responses, the AppView is always returning profiles sorted by followers

hipstersmoothie:
I can move 2-4 to another pr but I would like to see them fixed

> the AppView always returns who you follow first, and it's been this way since the end of invite-only period

I'm unsure where the data is coming from but in the app none of these lists are sorted like that.

> in lists where pagination is involved

I don't think I touched any paginated list?

> if this was indeed a problem, this should've been made at the AppView-level, not the app, or at the very least the sorting should not be done after flattening the paginated list.

Can you point me towards where is should make the change?

mary-ext:
Ah, sorry for that, I thought the starter packs had paginated queries but I forgot that's not the case.

src/view/com/composer/text-input/web/Autocomplete.tsx
```diff
@@ -151,7 +156,7 @@ const MentionList = forwardRef<MentionListRef, SuggestionProps>(
       <div className="items">
         <View style={[pal.borderDark, pal.view, styles.container]}>
           {items.length > 0 ? (
-            items.map((item, index) => {
+            items.sort(sortProfiles).map((item, index) => {
```
gaearon, line 159: `sort` is mutative so we can't do it here.

ideally, if the server response isn't appropriately sorted but we _want_ it to be sorted, we'd place this logic closest to where the response is being parsed — i.e. in the query. mind doing that?

src/view/screens/Search/Search.tsx
```diff
@@ -1006,7 +1007,7 @@ let AutocompleteResults = ({
             }
             style={{borderBottomWidth: 1}}
           />
-          {autocompleteData?.map(item => (
+          {autocompleteData?.sort(sortProfiles).map(item => (
```
gaearon, line 1010: same, can't mutate while rendering. doing it in the query before returning is ok.

src/view/shell/desktop/Search.tsx
```diff
@@ -206,7 +207,7 @@ export function DesktopSearch() {
                     : undefined
                 }
               />
-              {autocompleteData?.map(item => (
+              {autocompleteData?.sort(sortProfiles).map(item => (
```
gaearon, line 210: same

gaearon (changes requested):
seems like a good idea, pls see comments

PR #6189, opened 2024-11-09
[Lightbox] Always rely on Expo Image cache

gaearon:
Removes our own caching layer for image dimensions. Instead, rely on `onLoad` callback from `expo-image` which tells us the dimensions. So if the image is cached by `expo-image`, these will come through via `onLoad`.

In the lightbox, until `onLoad` fires for the full image, we'll use either the embed record dimensions (if available), or fall back to the already fetched thumbnail dimensions (determined by `onLoad` in the feed).

What this gives us:

- No more duplicate requests (via RN Image vs Expo Image) for missing dimension embed records
- The lightbox is properly animated if `onLoad` in feed ran in time — since we know the aspect ratio early in that case

## Test Plan

Verify that images animate when opening. Notably, images on posts like https://bsky.app/profile/docdre.bsky.social/post/3l6lzzqhxwh2o (which has no image dimensions in embed record) will _also_ animate now.

Comment out the `onLoad` calls in `AutoSizedImage.tsx` and `Gallery.tsx`. Verify they aren't essential — the lightbox works without them. For posts with no dimensions in the embed (https://bsky.app/profile/docdre.bsky.social/post/3l6lzzqhxwh2o), the lightbox would not animate without those `onLoad` calls — similar as before this PR. It should still display though. Posts _with_ dimensions in embed records should continue animating even without those `onLoad` calls.

Verify images appear cropped in feed similarly as before. Verify the animation from the cropped ones works.

Not sure how to verify this, but I'd expect that we no longer fire duplicate requests for images with missing dimensions.


haileyok (approved):
makes sense, and works as described on the test image linked.

PR #6196, opened 2024-11-09
android scroll performance fixes pt. 1

haileyok:
## Why

Investigating Android performance problems led us to - at least in part - the lack of memoization here. There is a considerable improvement by memoizing here.

## Test Plan

Essentially, checkout main and test from a low end Android device. Compare that to what this branch feels like. There should be a marked difference in performance.

Also worth checking - perhaps side by side with a different device running on main - to make sure that no styles have changed by doing this, though they shouldn't have.

### DIFF
--- a/src/alf/index.tsx
+++ b/src/alf/index.tsx
@@ -103,35 +103,32 @@ export function ThemeProvider({
     })
   }, [])
 
-  return (
-    <Context.Provider
-      value={React.useMemo<Alf>(
-        () => ({
-          themes,
-          themeName: themeName,
-          theme: themes[themeName],
-          fonts: {
-            scale: fontScale,
-            scaleMultiplier: fontScaleMultiplier,
-            family: fontFamily,
-            setFontScale: setFontScaleAndPersist,
-            setFontFamily: setFontFamilyAndPersist,
-          },
-          flags: {},
-        }),
-        [
-          themeName,
-          themes,
-          fontScale,
-          setFontScaleAndPersist,
-          fontFamily,
-          setFontFamilyAndPersist,
-          fontScaleMultiplier,
-        ],
-      )}>
-      {children}
-    </Context.Provider>
+  const value = React.useMemo<Alf>(
+    () => ({
+      themes,
+      themeName: themeName,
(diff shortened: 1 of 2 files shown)


src/alf/index.tsx
```diff
@@ -103,35 +103,32 @@ export function ThemeProvider({
...
+      themes,
+      fontScale,
+      setFontScaleAndPersist,
+      fontFamily,
+      setFontFamilyAndPersist,
+      fontScaleMultiplier,
+    ],
   )
+
+  return <Context.Provider value={value}>{children}</Context.Provider>
```
haileyok, line 131: no change here, just moving this out of the jsx

src/view/com/util/text/Text.tsx
```diff
@@ -74,49 +75,45 @@ export function Text({
     // @ts-ignore
     if (flattened.fontSize) {
       // @ts-ignore
-      flattened.fontSize = flattened.fontSize * fonts.scaleMultiplier
+      flattened.fontSize = Math.round(
```
haileyok, line 78: Rounding here might not be strictly necessary, but it feels bad to have a chance of a value like `14.065`, which is one of the font sizes that we observe in the main feed with some font preferences.

src/view/com/util/text/Text.tsx
```diff
@@ -74,49 +75,45 @@ export function Text({
...
+      )
     }
 
-    const shared = {
-      uiTextView: true,
+    return {
+      uiTextView: selectable && isIOS,
       selectable,
       style: flattened,
+      dataSet: Object.assign({tooltip: title}, dataSet || {}),
```
gaearon, line 88: this is now always assigned but used to be skipped under iOS && selectable, right? seems fine either way cause it’s probably web only

gaearon (approved):

src/view/com/util/text/Text.tsx
haileyok: Yea, it should definitely be web only. Actually may make sense to _only_ add the prop for `isWeb` anyway.

Issue #6200, opened 2024-11-10
The body element remains locked when resizing with the navigation menu open.


seabeya:
### Steps to Reproduce

1. Go to the web app main page.
2. Resize the window down until you see the 'Open navigation' button (top left).
3. Open the navigation menu.
4. Resize the window up until the open navigation menu disappears.
5. Now, if you try scrolling, you will see it is not possible.

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 7731136 (prod); Bundle date: 24110818; Platform: web

### Additional Information

When we open the navigation menu it adds `style="overflow: hidden;"` to the body element. However, after resizing up, it doesn't remove the overflow style. Also, at this point we don't have an outside click handler, so clicking somewhere on the page doesn't help. The only way to restore scrolling is to go back to a small screen and close the menu manually.

PR #6201, opened 2024-11-10
Fix unscrollable body when returning to desktop mode from mobile mode with navigation drawer open.

JustSNguyen:
So, this PR is related to this issue: #6200. Currently, BlueSky blocks the body scroll when the navigation drawer is opened: https://github.com/bluesky-social/social-app/blob/89c93313a3fc3123a2e53856f53cc8cea8e693fa/src/view/shell/index.web.tsx#L36

The problem is that when users switch to mobile mode, open the drawer, and then switch back to desktop mode, the navigation drawer is still considered open, which is incorrect. I’ve implemented a quick fix here to automatically close the drawer whenever the user exits mobile mode.

### DIFF
--- a/src/view/shell/index.web.tsx
+++ b/src/view/shell/index.web.tsx
@@ -32,8 +32,8 @@ function ShellInner() {
   const navigator = useNavigation<NavigationProp>()
   const closeAllActiveElements = useCloseAllActiveElements()
   const {_} = useLingui()
-
   const showDrawer = !isDesktop && isDrawerOpen
+
   useWebBodyScrollLock(showDrawer)
   useComposerKeyboardShortcut()
   useIntentHandler()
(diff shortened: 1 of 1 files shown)


gaearon:
The data flow is kind of backwards here. We shouldn't be setting state in an effect based on some other state — we should instead make the value which _decides_ whether to render the drawer (presumably it's above in this case) the source of truth (or contribute to the source of truth). Is that possible?

gaearon:
I think you can declare `const showDrawer = !isDesktop && isDrawerOpen` and then use that for the body scroll hook.

JustSNguyen:
Oh thanks @gaearon for your feedback, I think that would be better. I have updated the PR.

gaearon (approved):

gaearon:
thank you!

PR #6202, opened 2024-11-10
Fix Retain Cycles in SheetViewController

DrHazemAli:
Fix Retain Cycles in `SheetViewController.swift`

This pull request addresses potential retain cycles in `SheetViewController.swift` to improve memory management. Key changes include:

- Added `[weak self]` in the closure of `sheet.animateChanges` in the `updateDetents` method to prevent `self` from being strongly captured.
- Also captured `sheet` weakly in the closure to avoid any potential strong references retained by `animateChanges`.

These changes ensure the `SheetViewController` and its `UISheetPresentationController` are properly deallocated when no longer needed, preventing memory leaks while maintaining the functionality of the class.

**Reference**
- [**Apple Documentation on Memory Management**](https://developer.apple.com/library/archive/documentation/Swift/Conceptual/Swift_Programming_Language/AutomaticReferenceCounting.html)  

Let me know if you need further clarification!

### DIFF
--- a/modules/bottom-sheet/ios/SheetViewController.swift
+++ b/modules/bottom-sheet/ios/SheetViewController.swift
@@ -64,14 +64,14 @@ class SheetViewController: UIViewController {
 
   func updateDetents(contentHeight: CGFloat, preventExpansion: Bool) {
     if let sheet = self.sheetPresentationController {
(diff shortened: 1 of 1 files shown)


DrHazemAli (commented):
Rechecked

haileyok (approved):
Ah, nice catch. Thanks!

PR #6203, opened 2024-11-10
Mark "Invalid Handle" translatable in sanitizeHandle()

tkusano:
Instead of `<Trans>saniizeHandle()</Trans>`, mark "`⚠Invalid Handle`" translatable in `sanitizeHandle()`. I'm not sure if this approach is correct for RTL languages.


oracularhades:
Why do you need `import {t} from '@lingui/macro`?

tkusano:
`t` macro is needed for...
- Mark this string is translatable for translators's work 
- Translate "invalid handle" to the user's display language
- Already used in other lib/\*/\*.ts

mozzius:
`t` macro is not the right tool here. you need to pass in an `i18n` instance and use `i18n._`, otherwise it won't update when you change languages

tkusano:
oh, I checked #4273, and realized the problem. hmm.

tkusano:
rewrote `sanitizeHandle()` to take `i18n` as 1st parameter and related functions and components.  I'm not sure if this is a right solution.

tkusano:
I caught up with main

estrattonbailey:
Thanks for all the hard work here! Considering the breadth of this, I think the simpler option in https://github.com/bluesky-social/social-app/pull/11071 is the right move here and will work for the vast majority of cases.

PR #6206, opened 2024-11-10
Fix the translation of the adjective "open"

CanGiante:
The word "open" can be a verb like in "Open avatar creator" or an adjective (e.g. "Bluesky is open"). This commit replace the verb "Apri" with the adjective "Aperto".


pfrazee (approved):
Thanks!!

PR #6207, opened 2024-11-10
Localize lang selectors according to the app language

Signez:
There are two main things that are not yet translatable in Bluesky's UI:

 - Default feed names and descriptions (but this may be a can of worms to open, apart from the specific case of _Following_);
 - Language selectors, in the composer and in the settings screens.

As the second one is quite ironic (don't you think? 🎶) and quite visible on a daily basis, nagging at me as the main translator of Bluesky's app into French with @surfdude29, I thought it would be great to help you out and fix it myself, so you don't have to take time away from your heavy roadmap 😅

---

In this pull request, I use `Intl.DisplayName` to offer a _best-effort_ translation of the language names, if the browser or target provides it.

Capture d’écran 2024-11-10 à 18 15 23
:arrow_up: _Here, notice the **Français** language name. It would show **French** on `main`._

Of course, I took the time to fix the same problem in the _Languages_ screen:

Capture d’écran 2024-11-10 à 18 15 14
:arrow_up: _This screen always offered a translation for the app language, and I don't touch it to help anyone revert an unwanted change of app language settings._

**Note:** This is a _best-effort_ fix: if Expo's runtime on the mobile apps do not support the `Intl.DisplayNames`, it will not do anything, and simply show English names as of today.

### DIFF
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -5,6 +5,7 @@ import lande from 'lande'
 import {hasProp} from '#/lib/type-guards'
 import {
   AppLanguage,
+  type Language,
   LANGUAGES_MAP_CODE2,
   LANGUAGES_MAP_CODE3,
 } from './languages'
@@ -31,9 +32,37 @@ export function code3ToCode2Strict(lang: string): string | undefined {
   return undefined
 }
 
-export function codeToLanguageName(lang: string): string {
-  const lang2 = code3ToCode2(lang)
-  return LANGUAGES_MAP_CODE2[lang2]?.name || lang
+function getLocalizedLanguage(
+  langCode: string,
+  appLang: string,
+): string | undefined {
+  const allNames = new Intl.DisplayNames([appLang], {
+    type: 'language',
+    fallback: 'none',
+    languageDisplay: 'standard',
+  })
+  const translatedName = allNames.of(langCode)
+
+  if (translatedName) {
+    // force simple title case (as languages do not always start with an uppercase in Unicode data)
+    return translatedName[0].toLocaleUpperCase() + translatedName.slice(1)
+  }
+}
+
+export function languageName(language: Language, appLang: string): string {
+  // if Intl.DisplayNames is unavailable on the target, display the English name
+  if (!(Intl as any).DisplayNames) {
+    return language.name
+  }
+
+  return getLocalizedLanguage(language.code2, appLang) || language.name
+}
+
+export function codeToLanguageName(lang2or3: string, appLang: string): string {
+  const code2 = code3ToCode2(lang2or3)
+  const knownLanguage = LANGUAGES_MAP_CODE2[code2]
+
+  return knownLanguage ? languageName(knownLanguage, appLang) : code2
 }
 
 export function getPostLanguage(
--- a/src/locale/helpers.ts
+++ b/src/locale/helpers.ts
@@ -36,16 +36,23 @@ function getLocalizedLanguage(
   langCode: string,
   appLang: string,
 ): string | undefined {
-  const allNames = new Intl.DisplayNames([appLang], {
-    type: 'language',
-    fallback: 'none',
-    languageDisplay: 'standard',
-  })
-  const translatedName = allNames.of(langCode)
+  try {
+    const allNames = new Intl.DisplayNames([appLang], {
+      type: 'language',
+      fallback: 'none',
+      languageDisplay: 'standard',
+    })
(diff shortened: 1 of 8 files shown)


auroursa:
I tested Chinese, Japanese and Korean and they work perfect for me👍
image

tkusano:
I haven't tested it exhaustively, but I gave it a quick run in Firefox and it looks good. Because Hermes engine ,iOS and Android use, doesn't seem to support `Intl.DisplayNames` API, polyfilling `@formatjs/intl-displaynames` might be a option.

surfdude29:
This is a great idea @Signez 👌

src/locale/helpers.ts
```diff
@@ -31,9 +32,37 @@ export function code3ToCode2Strict(lang: string): string | undefined {
...
+}
+
+export function languageName(language: Language, appLang: string): string {
+  // if Intl.DisplayNames is unavailable on the target, display the English name
+  if (!(Intl as any).DisplayNames) {
+    return language.name
+  }
+
+  return getLocalizedLanguage(language.code2, appLang) || language.name
+}
```
JoshMLeslie, line 59: Could this be rewritten as a hook, to avoid needing to pass in langPrefs everywhere?
e.g.
```
export function useLanguageName(language: Language): string {
  const {appLanguage} = useLangPrefs();
  ...
}
```

Signez: As hooks shall not be called conditionally, an hypothetical `useLanguageName` hook could not be used as-is inside `.map` calls… where most of the existing calls to `codeToLanguageName` are done.

So yeah, I am not convinced we would gain readability by using a hook here!

gaearon:
not gonna do right now but please remind me to look next week

Signez:
Just rebased on `main`, pinging you this week @gaearon as you suggested ✨

src/locale/helpers.ts
```diff
@@ -31,9 +32,37 @@ export function code3ToCode2Strict(lang: string): string | undefined {
...
+function getLocalizedLanguage(
+  langCode: string,
+  appLang: string,
+): string | undefined {
+  const allNames = new Intl.DisplayNames([appLang], {
+    type: 'language',
+    fallback: 'none',
+    languageDisplay: 'standard',
+  })
+  const translatedName = allNames.of(langCode)
```
c960657, line 44: [of()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/of) expects a language code matching the [unicode_language_id](https://unicode.org/reports/tr35/#Unicode_language_identifier) rule from CLDR which is only a subset of BCP 47, the format permitted in AT Protocol. E.g. `zh-YUE` is a valid BCP 47 code but will make the `of()` method throw a `RangeError`, so the call needs to be wrapped in a `try`/`catch` block.

Signez: Fixed, good catch! Thanks :)

Signez:
Rebased on `main`, with a nasty bug (caught by @c960657) fixed.

pfrazee (approved):
Nice! My main issue is that, at least in my sim tests, `Intl.DisplayNames` wasn't available on either android or ios. This only worked for me in my web test. You correctly handle that situation, and I want to go ahead and merge so we make progress. Hopefully that api comes to Hermes someday.

Signez:
Thanks! It was indeed what @tkusano spotted; they suggested that we could use [`@formatjs/intl-displaynames`](https://www.npmjs.com/package/@formatjs/intl-displaynames) (and it would work), but it would be a waste to package it in every installation when there is native API that would allow us to use the ones already stored by the system: [`getDisplayLanguage(Locale locale)`](https://developer.android.com/reference/java/util/Locale#getDisplayLanguage(java.util.Locale)) on a [`java.util.Locale`](https://developer.android.com/reference/java/util/Locale) on Android, and [`localizedString(forLanguageCode: string)`](https://developer.apple.com/documentation/foundation/locale/2293712-localizedstring) on `Fondation`'s [`Locale`](https://developer.apple.com/documentation/foundation/locale) on iOS.

I thought it would be a good follow-up PR in the future, maybe contributing an Expo package (or an improvement to [`expo-localization`](https://docs.expo.dev/versions/latest/sdk/localization/)) to offer this to other projects too!

Signez:
I think it would be a good compromise, indeed! Creating a blocklist for those few languages should be easy. (To be clear: it's a blocklist of the _user language to translate into_, i.e. `appLang`, not a blocklist of _language names to translate_.)

PR #6211, opened 2024-11-10
Upgrade RN to 0.76

haileyok:
Removing some patches:

- Metro patches, the changes we made got merged in upstream
- Some of the Expo Notifications patch (being able to reset the badge) got merged in upstream. Rming that change. Other EN changes are our own patch, but needed to be updated.
- modules-core patch got merged in upstream
- image-picker patch got merged in upstream
- React Native crash with empty body fix got merged in upstream, so rm'd that. The "select all on focus" fix also got merged in upstream so rm'd. The rest of the patch is for a few of our own hacks.


app.config.js
```diff
@@ -141,7 +141,6 @@ module.exports = function (config) {
         adaptiveIcon: {
           foregroundImage: './assets/icon-android-foreground.png',
           monochromeImage: './assets/icon-android-foreground.png',
-          backgroundImage: './assets/icon-android-background.png',
```
haileyok, line 144: This is the one thing I don't like. Going to ask in Expo channel what might have changed. Fails to build if I leave this here.

modules/expo-bluesky-gif-view/android/src/main/java/expo/modules/blueskygifview/GifView.kt
```diff
@@ -85,29 +86,27 @@ class GifView(
...
+            model: Any,
+            target: Target<Drawable>?,
+            dataSource: DataSource,
+            isFirstResource: Boolean
+          ): Boolean {
+            placeholderRequest?.let {
+              glide.clear(it)
             }
+            return false
+          }
```
haileyok, line 101: No changes here really, but the Glide bump made `Drawable` non-optional for `resource`

package.json
```diff
@@ -263,6 +261,7 @@
     "lint-staged": "^13.2.3",
     "metro-react-native-babel-preset": "^0.74.1",
     "prettier": "^2.8.3",
+    "react-compiler-runtime": "^19.0.0-beta-63b359f-20241101",
```
haileyok, line 264: Instead of the one we had inside `lib`

src/lib/hooks/usePermissions.ts
```diff
@@ -78,7 +78,7 @@ export function useVideoLibraryPermission() {
 }
 
 export function useCameraPermission() {
-  const [res, requestPermission] = Camera.useCameraPermissions()
+  const [res, requestPermission] = useExpoCameraPermissions()
```
haileyok, line 81: Needs to be tested

patches/react-native-compressor+1.9.0.patch
```diff
@@ -0,0 +1,16 @@
...
+@@ -197,11 +197,6 @@
+         "name": "RNCompressorSpec",
+         "type": "modules",
+         "jsSrcsDir": "src"
+-      },
+-      {
+-        "name": "RNVideoCompressorSpec",
+-        "type": "modules",
+-        "jsSrcsDir": "src"
+       }
```
haileyok, line 14: This is incorrect and causes build failure on iOS, removing one of the entries.

patches/react-native-screens+3.31.1.patch
```diff
@@ -0,0 +1,19 @@
...
+         stateWrapper: StateWrapper?
+     ): Any? {
+-        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
+-            // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
+-            view.fabricViewStateManager?.setStateWrapper(stateWrapper)
+-        }
++//        if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
++//            // fabricViewStateManager should never be null in Fabric. The null check is only for Paper's empty impl.
++//            view.fabricViewStateManager?.setStateWrapper(stateWrapper)
++//        }
```
haileyok, line 16: Upgrading screens to latest introduces a graphical bug which we do not want. However, this block fails on RN 0.76. It isn't useful anyway, since we are not using new arch yet (and the upgrade does not introduce the graphical bug in Fabric anyway, so we can upgrade when we go fabric)

patches/react-native+0.76.1.patch
```diff
@@ -152,19 +105,3 @@ index 40aaf9c..1c60164 100644
...
-
-       // If the JS thread is busy for multiple frames we cancel any other pending runnable.
--      if (mCurrentIdleCallbackRunnable != null) {
--        mCurrentIdleCallbackRunnable.cancel();
-+      IdleCallbackRunnable currentRunnable = mCurrentIdleCallbackRunnable;
-+      if (currentRunnable != null) {
-+        currentRunnable.cancel();
-       }
-
-       mCurrentIdleCallbackRunnable = new IdleCallbackRunnable(frameTimeNanos);
```
haileyok, line 170: Didn't look to see if this got fixed upstream (though I think it did). Regardless, this doesn't work anyway since we don't compile from source.

patches/react-native+0.76.1.patch
```diff
@@ -1,42 +1,3 @@
...
- - (void)reactFocus
- {
-   [self.backedTextInputView reactFocus];
-+
-+  if (_selectTextOnFocus) {
-+    [self.backedTextInputView selectAll:nil];
-+  }
- }
-
- - (void)reactBlur
```
haileyok, line 39: All of this has been merged in upstream and has been available since ~0.75

haileyok:
Lol okay, was able to upgrade without deleting yarn.lock from Linux, so shouldn't need those weird type patches anymore hopefully.

package.json
mozzius: I had done this, but we had to revert due to a crash bug (https://github.com/bluesky-social/social-app/pull/6071)

troZee:
FYI @haileyok @auroursa 

The new version of the pager has been released https://github.com/callstack/react-native-pager-view/releases/tag/v6.5.1 . In this version, we fixed issues regarding the new arch. Please let us know in case of any issues

package.json
```diff
@@ -166,36 +165,35 @@
...
     "react-native-get-random-values": "~1.11.0",
-    "react-native-image-crop-picker": "0.41.2",
+    "react-native-image-crop-picker": "0.41.4",
     "react-native-ios-context-menu": "^1.15.3",
-    "react-native-keyboard-controller": "^1.14.0",
+    "react-native-keyboard-controller": "^1.14.4",
     "react-native-mmkv": "^2.12.2",
-    "react-native-pager-view": "6.2.3",
-    "react-native-picker-select": "^9.1.3",
+    "react-native-pager-view": "6.5.0",
```
troZee, line 181: https://github.com/bluesky-social/social-app/pull/6211#issuecomment-2486747425

haileyok: Awesome, thank you for the heads up!

PR #6213, opened 2024-11-10
Fix render loop on Android

gaearon:
I'm not sure what the root cause is, but I'm seeing `onLoad -> setState -> onLoad` loops on Android in the feed even when the image has not changed. I'm also not sure whether this is a comprehensive fix, but it does seem to help.

## Test Plan

I haven't found a better way than to add `console.log('render')` to `AutoSizedImage.tsx`. As I scroll Discover, it would occasionally start repeatedly fire for the same image. After applying the fix (with Fast Refresh), it stops.

### Before

https://github.com/user-attachments/assets/d0f6d73e-0007-4128-9121-81d53e45b9af

### After

https://github.com/user-attachments/assets/0d19a482-6d69-434e-88d5-1da540d6b65b

The fix should generally be safe anyway so this shouldn't hurt anything.

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -396,10 +396,14 @@ const ImageItem = ({
                 placeholderContentFit="cover"
                 placeholder={{uri: imageSrc.thumbUri}}
                 accessibilityLabel={imageSrc.alt}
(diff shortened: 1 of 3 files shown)


haileyok (approved):

mzaien:
Hey @gaearon sorry for the random question, what is orange bar in the demo?

gaearon:
oh this is kind of unrelated, but it's the android GPU profiler https://developer.android.com/topic/performance/rendering/profile-gpu

PR #6215, opened 2024-11-10
Fix 'Show badge' button label being cut off

### LINKED ISSUE #6185: Label "Show Badge" is cut off
### Steps to Reproduce

1. Go to any labeler, e.g. https://bsky.app/profile/did:plc:zal76px7lfptnpgn4j3v6i7d
2. You'll see that "Show badge" is cut off

CleanShot 2024-11-09 at 13 48 19@2x

### Attachments

_No response_

Web (Desktop)

### Device Info

MacOS Sonoma 14.3.1 (23D60)

Chrome, Version 130.0.6723.70 (Official Build) (arm64)

### Additional Information

_No response_

khuddite:
Fixes #6185 
1. Summary
    `Show badge` label is cut off on the developer labels page.

2. Solution
    Before: 
image
    After: 
CleanShot 2024-11-10 at 15 26 03

The fix looks better than being cut off, but I am unsure if it is the best solution to the problem. I can think of different ways of solving it, and here are some of them
    - Updating the label to say `Show` instead of `Show badge`
    - Increasing the button width for all 3
    - Put the toggle group underneath `title/description` section like on mobile devices
    
I guess, I need some design inputs here on what is the desired behavior aesthetically.

### DIFF
--- a/src/components/forms/ToggleButton.tsx
+++ b/src/components/forms/ToggleButton.tsx
@@ -95,6 +95,9 @@ function ButtonInner({children}: React.PropsWithChildren<{}>) {
         native({
           paddingBottom: 10,
         }),
+        a.w_full,
+        a.h_full,
+        a.justify_center,
         a.px_md,
         t.atoms.bg,
         t.atoms.border_contrast_low,


nspilman:
Thanks for this fix! Came here to try to do the same thing :-)

JoshMLeslie:
+1 for "Show" instead of "Show Badge" as it's redundant. I'm more curious the distinction b/w "off" and "hide". The header description could be more verbose imo.

estrattonbailey (approved):
Thanks for the help! Much appreciated 🙏

victor-ponamariov:
+1 for just "Show".

But generally speaking, I'm very confused about the whole thing. We have labelers, that can add labels to our profiles via likes, right? 

> Labels are annotations on users and content. They can be used to hide, warn, and categorize the network.

So, what these three buttons actually do. Off, show, and hide. Is it about fine-tuning your feed, to hide people who have some labels? Then... why there are three buttons, not two (show/hide)?

And what happens if 2 labelers create a label with the same name (while in the database they'll have unique ids I believe), so we'll end up with two "Javascript" for example from two accounts/labelers.

PR #6219, opened 2024-11-11
Reduce Reanimated serialization traffic

gaearon:
This was a major WTF moment for me and I'm still not sure to what extent this is actually useful. However, the [Android GPU profiler](https://developer.android.com/topic/performance/rendering/profile-gpu) seems to be happy about this change.

There is some information about how Reanimated closures work here: https://github.com/software-mansion/react-native-reanimated/issues/3670#issuecomment-1303388096

We're closing over `withSpring` which I assume means that [this clone function](https://github.com/software-mansion/react-native-reanimated/blob/66137a6b084d787e5d98aa496d66b83cbd204d71/packages/react-native-reanimated/src/shareables.ts#L103-L291) will have to go into `withSpring` itself, deal with _its_ closures over top scope, and so on. I haven't tried logging what's going on there but I suspect there may be some serialization overhead.

However, we can just create initial and final values of the animation and expose _those_ to the worklet. I didn't expect this to be faster but it does seem to be. Or maybe the GPU debugger is being misleading.

## Test Plan

Enable GPU profiler, build for release on device, remove all embeds (they add too much other unpredictability), continuously scroll a little bit up and down with your finger without releasing on Home.

## Before

Overall we're way above the red line.

The green+blue stuff reaches the green line.

Screenshot 2024-11-11 at 01 27 26

## After

Overall we're a bit below the red line.

The green+blue stuff is halfway within the green line.

Screenshot 2024-11-11 at 01 27 16

### DIFF
--- a/src/view/com/util/MainScrollProvider.tsx
+++ b/src/view/com/util/MainScrollProvider.tsx
@@ -3,6 +3,7 @@ import {NativeScrollEvent} from 'react-native'
 import {
   cancelAnimation,
   interpolate,
+  makeMutable,
   useSharedValue,
   withSpring,
 } from 'react-native-reanimated'
@@ -20,6 +21,18 @@ function clamp(num: number, min: number, max: number) {
   return Math.min(Math.max(num, min), max)
 }
 
+const V0 = makeMutable(
+  withSpring(0, {
+    overshootClamping: true,
+  }),
+)
+
+const V1 = makeMutable(
+  withSpring(1, {
+    overshootClamping: true,
+  }),
+)
+
 export function MainScrollProvider({children}: {children: React.ReactNode}) {
   const {headerHeight} = useShellLayout()
   const {headerMode} = useMinimalShellMode()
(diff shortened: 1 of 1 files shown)


mozzius (approved):
Appears to double my framerate as far as I can tell! Great work

This is a major footgun with Reanimated, I'm shocked this makes such a difference. I *think* the `makeMutable` code is good, since it's outside of a component it's fine to not cancel/clean up

hirbod:
You only need to call cancelAnimation() when you actually animate with the value, can be skipped if not used for animation. Great find!

gaearon:
For posterity, Reanimated fixed this upstream in https://github.com/software-mansion/react-native-reanimated/pull/6758 which I assume went out in 3.17.0

mozzius:
Nice! I'll see about reverting this then

gaearon:
no worries, we already did the revert :) i just linked to this PR from somewhere and then realized it’s not very clear that the upstream issue was fixed

mozzius:
Ahh :) nice

Issue #6204, opened 2024-11-10
Lists tab on the profile page: broken pagination


imax9000:
### Steps to Reproduce

1. Open profile of an account with more than 30 lists
2. Open browser devtool and switch to "Network" tab to see the requests
3. In the page, switch to "Lists" tab
4. Observe that there's only one `app.bsky.graph.getLists` request sent
5. Scroll to the bottom of the page
6. Observe that no additional requests are sent to fetch the rest of the lists

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Firefox 132.0.1 on Windows

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 7731136 (prod); Bundle date: 24110818; Platform: web

imax9000:
Here are a few:

```
bluesky=# select count, 'https://bsky.app/profile/' || did from (select repo as id, count(*) from records where collection='app.bsky.graph.list' and not deleted group by repo) join repos using (id) order by count desc limit 10;
 count |                         ?column?                          
-------+-----------------------------------------------------------
   210 | https://bsky.app/profile/did:plc:pwlfo4w6auzwihryxik32t6d
   171 | https://bsky.app/profile/did:plc:tsfj4ybk5ke3gxfymzksnhwu
   161 | https://bsky.app/profile/did:plc:a3pshickwc2522q5l73vsweq
   109 | https://bsky.app/profile/did:plc:l3xkhxqujdpybq6uwxjgrkrz
   100 | https://bsky.app/profile/did:plc:sbzvs4amqvtmveswkrs34qon
    91 | https://bsky.app/profile/did:plc:ndywsez2becfh2hxp7ugrwwz
    85 | https://bsky.app/profile/did:plc:tdq2tzpiggojebao3mgwemqt
    81 | https://bsky.app/profile/did:plc:z5rc7zwgmfg6v35steefdeyb
    74 | https://bsky.app/profile/did:plc:7qbjf7s5trvjyy3qsw6ehcos
    73 | https://bsky.app/profile/did:plc:d4efji3r7iqvev73zc5kjelp
(10 rows)
```

I've checked two of them myself and it seemed to work fine =/ But breaks on this one: https://bsky.app/profile/did:plc:bmjomljebcsuxolnygfgqtap

PR #6221, opened 2024-11-11
Fix broken pagination for lists tab on profile page

khuddite:
Fixes: #6204 

1. Summary
`onEndReached` is not triggered for the `ProfileLists` component, and it seems `onEndReachedThreshold` must be set.
The issue is specific to the desktop.

2. Solution
Set `onEndReachedThreshold` to 2. I've tested and confirmed it's working across all different platforms (android, iOS and web)

3. Recording
https://github.com/user-attachments/assets/f94a2f55-c6d7-4152-bf88-d69e0d38fe20


gaearon:
> it seems onEndReachedThreshold must be set.

It's marked as optional, so if it's not working without it, we should figure out why and fix that instead.

khuddite:
> > it seems onEndReachedThreshold must be set.
> 
> It's marked as optional, so if it's not working without it, we should figure out why and fix that instead.

That's a good question. 
You're right - `onEndReachedThreshold` is optional on `FlatList` component, but we don't use it for desktop. It's used only for mobile. It seems we built our own list component for desktop - https://github.com/khuddite/social-app/blob/5da3f29498fda9ab1181df19a718e37099cb2cf6/src/view/com/util/List.web.tsx

On this component, `onEndReachedThreshold` is optional, but it defaults to zero when missing.

I think `onEndReachedThreshold` being zero is a good reason for this issue because it impacts `bottomMargin` of the tail visibility component. [bottomMargin={(onEndReachedThreshold ?? 0) * 100 + '%'}](https://github.com/khuddite/social-app/blob/4e44d44692aaa348a1736ed5d52eb31f78fc9af4/src/view/com/util/List.web.tsx#L370)

I came up with the number `2` because it's used in other list tabs that don't have the same problem (like replies, posts, media, etc). [onEndReachedThreshold={2} // number of posts left to trigger load more](https://github.com/khuddite/social-app/blob/4e44d44692aaa348a1736ed5d52eb31f78fc9af4/src/view/com/posts/Feed.tsx#L555-L556)

gaearon:
Hmm, is zero broken as in "it doesn't work even when reaching end directly"? We could maybe just change the default to `2` there (is that RN's default as well?) but I'd like to understand whether it's downright broken with `0` (not registering intersections at all), or whether this particular screen is just somehow harder to scroll _exactly_ to the end.

khuddite:
I investigated a bit more, and here are my findings. 

The lists tab has another issue - it's missing a footer component that displays a loading spinner when it reaches the end and pulls the next page. And it's related to the reported issue.

For other lists, when we don't have to display a loading spinner, we still display an empty view that has a certain [height](https://github.com/khuddite/social-app/blob/4e44d44692aaa348a1736ed5d52eb31f78fc9af4/src/view/com/posts/Feed.tsx#L533). And that seems required for the list to display the entire list without being cut off at the end and `onEndReached` to be triggered when it should be.

I've updated the `ProfilesList` component to include a footer and display a loading spinner while fetching the next page. This fixes the reported issue while making it consistent with other lists on the profile tab.

However, I think defaulting `onEndReachedThreshold` to `2` instead of `0` for the web list is still a good idea because that's how `FlatList` works on mobile. In the current implementation, if `onEndReachedThreshold`'s missing, it will default to `2` for mobile and `0` for web, which doesn't seem desired.

src/view/com/lists/ProfileLists.tsx
```diff
@@ -201,3 +218,7 @@ export const ProfileLists = React.forwardRef<SectionRef, ProfileListsProps>(
     )
   },
 )
+
+const styles = StyleSheet.create({
+  listsFooter: {paddingTop: 20},
```
JoshMLeslie, line 223: nit: is this worth creating separately for since it's only one style? Why not just inline it?

khuddite: That's how it's handled for feeds list - https://github.com/bluesky-social/social-app/blob/f68b15219fd02e23d965015201400138ed69d59d/src/view/com/posts/Feed.tsx#L575

gaearon (approved):

PR #6222, opened 2024-11-11
Fix repost button styling when logged out

rolandcrosby:
On the web, the logged-out variant of the repost button sets a white background with `variant="ghost"`, which conflicts with the background applied when a post is hovered:
Screenshot 2024-11-10 at 9 35 34 PM

Removing the button styling seems to make this button render like the others on the page:
Screenshot 2024-11-10 at 9 36 22 PM


mmraz (commented):
LGTM

gaearon (approved):
thanks

PR #6257, opened 2024-11-12
Use hairline border for feed like card

mozzius:
It had a lopsided border of hairline at the top and 1px on the other sides. Easy fix

Screenshot 2024-11-12 at 15 22 27

### DIFF
--- a/src/view/com/notifications/FeedItem.tsx
+++ b/src/view/com/notifications/FeedItem.tsx
@@ -467,7 +467,12 @@ let FeedItem = ({
         {item.type === 'feedgen-like' && item.subjectUri ? (
           <FeedSourceCard
             feedUri={item.subjectUri}
-            style={[pal.view, pal.border, styles.feedcard]}
+            style={[
+              t.atoms.bg,
+              t.atoms.border_contrast_low,
+              a.border,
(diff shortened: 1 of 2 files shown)


gaearon (approved):

src/view/com/notifications/FeedItem.tsx
```diff
@@ -467,7 +467,12 @@ let FeedItem = ({
         {item.type === 'feedgen-like' && item.subjectUri ? (
           <FeedSourceCard
             feedUri={item.subjectUri}
-            style={[pal.view, pal.border, styles.feedcard]}
+            style={[
+              t.atoms.bg,
+              t.atoms.border_contrast_low,
+              a.border,
+              styles.feedcard,
+            ]}
```
JoshMLeslie, line 475: Hi! I'm learning the codebase and am wondering why you switched style objects from `pal. ...` to `t.atoms. ...`?

mozzius: New design system. We're like halfway through migration to the ALF system

JoshMLeslie: Thanks!

PR #6260, opened 2024-11-12
fix: nothing selected bug

### LINKED ISSUE #6259: Reply settings: No default
### You're running a build that uses the React Native New Architecture, right?

- [X] I am running a build that uses the New Architecture

### Steps to Reproduce

1. Begin creating a new post 
2. Before posting, click on "Anybody can interact" (Interaction settings) 
3. Select "mentioned users" 
4. De-select "mentioned users" 

Now no setting will be visible. 

### Attachments

_No response_

iOS, Android

### Device Info

Simulator iPhone 16 Plus 

Latest/main branch

### Additional Information

_No response_

O4FDev:
Fixes #6259

https://github.com/user-attachments/assets/e5d9e34e-6355-4736-92b3-2d6f333dfff8

New behaviour 

vs. 

https://github.com/user-attachments/assets/2da6575e-8c88-46d2-a11d-a7b7b131edc4

Old behaviour

### DIFF
--- a/src/components/dialogs/PostInteractionSettingsDialog.tsx
+++ b/src/components/dialogs/PostInteractionSettingsDialog.tsx
@@ -249,17 +249,19 @@ export function PostInteractionSettingsForm({
       threadgateAllowUISettings.filter(
         v => v.type !== 'nobody' && v.type !== 'everybody',
       )
-    // toggle
+    /**
+     * Handle updating the selected audience settings:
+     * - If the setting doesn't exist yet, add it to the list
+     * - If it's the only setting and being removed, revert to default (everybody)
+     * - If it exists and there are other settings, remove it from the list
+     */
     const i = newSelected.findIndex(v => isEqual(v, setting))
     if (i === -1) {
       newSelected.push(setting)
+    } else if (newSelected.length === 1) {
+      newSelected = [{type: 'everybody'}]
     } else {
       newSelected.splice(i, 1)
-    }
-
-    // if nothing is selected, default to 'everybody'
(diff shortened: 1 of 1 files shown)


src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -257,6 +257,11 @@ export function PostInteractionSettingsForm({
       newSelected.splice(i, 1)
     }
 
+    // if nothing is selected, default to 'everybody'
+    if (newSelected.length === 0) {
+      newSelected.push({type: 'everybody'})
+    }
```
JoshMLeslie, line 263: Since newSelected.length === 0 is a consequence of  the logic starting above, at :253, imo it makes more sense to write this more inclusively with that block alike 

```
  /**
   * check if the incoming setting DNE -> add to the list
   * is the only setting + being removed -> revert to default
   * or is being removed from the others -> splice
   */
  const i = newSelected.findIndex(v => isEqual(v, setting));
  if (i === -1) {
    newSelected.push(setting);
  } else if (newSelected.length === 1) {
    newSelected = [{type: 'everybody'}];
  } else {
    newSelected.splice(i, 1);
  }
```

O4FDev: Good point! Updated :) 
Just slightly modified your comments to expand abbreviations

gaearon (approved):
thanks for the fix! i slightly tweaked the code, i think this makes it clearer what's going on

PR #6267, opened 2024-11-12
Dedupe date parsing in PostMeta

mozzius:
very minor, but since this is in the hot path there's no point calling the same function twice


haileyok (approved):
nice

PR #6272, opened 2024-11-12
feat: remove following toast

O4FDev:
## Why 

Removes the following toast when viewing profiles -

1. The back button is blocked when toasts are displayed (See: https://bsky.app/profile/hlimas.com/post/3la5let3xn42e) 
2. The interaction should be self explanatory (i.e. the button feedback)

## What it's not 

Not attempting to remove the toast elsewhere outside of the header of profiles


gaearon:
We made toasts draggable instead

O4FDev:
I don't believe this is a complete fix @gaearon - it makes sense from a user perspective still to not have this toast even if you can dismiss it.

PR #6273, opened 2024-11-12
[DO NOT MERGE] Change all instances of Reanimated `sv.value =` to `sv.set()`

mozzius:
# DO NOT MERGE: UNSURE IF THIS IS SAFE

Should make React Compiler happier and prevent optimisation bailouts

I've tried to be careful, but I can't be sure these are all correct

>[!NOTE]
>It seems as if `.set()` must always be called on the UI thread. I've had to move some assignments to within a worklet function

# Test plan

Temporarily reenable the lint rule, and see if it stops erroring.

Make sure all the components work the same

### DIFF
--- a/src/Splash.tsx
+++ b/src/Splash.tsx
@@ -10,6 +10,7 @@ import Animated, {
   Easing,
   interpolate,
   runOnJS,
+  runOnUI,
   useAnimatedStyle,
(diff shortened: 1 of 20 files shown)


gaearon:
merged separately

Issue #6269, opened 2024-11-12
Viewing a video in landscape mode causes layout shift.


BradEstey:
### Steps to Reproduce

1. Click into a video to view in full screen.
2. Rotate device to view in landscape.
3. Click the `×` to close video (While still in landscape).
4. The entire app will shift to the left by about 1/3 width.

### Attachments

https://github.com/user-attachments/assets/ffa45a72-5333-48a1-bc0b-1788060fb917

### What platform(s) does this occur on?

iOS

### Device Info

iPhone SE (MX992LL/A) - iOS v18.0.1

### What version of the app are you using?

1.93.0.548 0ac7c0962

### Additional Information

Note, there is a single frame of the video where the app is rotated 90 degrees before the layout shift occurs.

khuddite:
I can reproduce the issue, investigating what's happening

PR #6274, opened 2024-11-12
Fix exiting video player from full screen in the landscape mode shifts UI up

khuddite:
Fixes: #6269 

1. Summary
The issue is related to the drawer from `react-native-drawer-layout` library. I confirmed that by removing the drawer in the code.

2. Solution
I upgraded `react-native-drawer-layout` to the latest version(v4.0.1) to see if that fixed the issue, and apparently, it did. I tested the app on Android, iOS, and desktop, and it seemed it didn't introduce any additional bugs/issues.

3. Recording

https://github.com/user-attachments/assets/32300a0b-a86b-45e0-8ccb-18ef3e508492

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "react-native": "0.74.1",
     "react-native-compressor": "^1.8.24",
     "react-native-date-picker": "^4.4.2",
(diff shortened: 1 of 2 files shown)


mozzius (approved):
great, thanks!

PR #6282, opened 2024-11-13
Self-replies don't indent threads

ericclemmons:
## Problem

Self-replies are common for threading.  The issue is that a 🧵 can quickly nest to 10 levels (the current `REPLY_TREE_DEPTH`) – https://github.com/bluesky-social/social-app/blob/5faf8572bdd637dacf586ba47bdf106a7fff627a/src/state/queries/post-thread.ts#L36

## Proposal

When a `post.reply` is from the same author, don't indent.

## Examples

https://bsky.app/profile/notablyuncool.bsky.social/post/3lalusyp44c2q

> Screenshot 2024-11-12 at 7 02 46 PM

https://bsky.app/profile/wesbos.com/post/3laadg2ydjg25

> Screenshot 2024-11-12 at 7 03 35 PM

### DIFF
--- a/src/state/queries/post-thread.ts
+++ b/src/state/queries/post-thread.ts
@@ -8,6 +8,7 @@ import {
   ModerationDecision,
   ModerationOpts,
 } from '@atproto/api'
+import {PostView} from '@atproto/api/dist/client/types/app/bsky/feed/defs'
 import {QueryClient, useQuery, useQueryClient} from '@tanstack/react-query'
 
 import {moderatePost_wrapped as moderatePost} from '#/lib/moderatePost_wrapped'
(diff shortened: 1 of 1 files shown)


gaearon:
How do we differentiate multiple top-level replies vs a thread? Or multiple spawned threads, each from its own root reply.

ericclemmons:
@gaearon Can you provide some example URLs?

I haven't updated the UI to differentiate these use-cases since I've only had 2 examples to prototype from.

kevinmarks:
What phanpy.social does is well worth looking at here - it maintains the original posters' thread vertically and adds inline markers to expand replies, or switch to centering that post. see https://bsky.app/profile/kevinmarks.com/post/3lau6sym75s2q for screenshots

ericclemmons:
Sorry, I won't be able to continue this PR without sponsorship.  Property taxes are in and 🤯

Issue #6287, opened 2024-11-13
After clicking the 'Translate' button, Bluesky will returns to the home page


auroursa:
### Steps to Reproduce

1. Open a post that needs translation
2. Click the 'Translate' button
3. The browser opens a new tab to load the translation
4. Return to Bluesky's tab, it returns to the home page

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Windows 11 24H2 + Chrome 131.0.6778.70

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 7731136 (prod); Bundle date: 24110818; Platform: web

O4FDev:
Had a quick look - replicated. Going to have a look into fixing :)

surfdude29:
@auroursa @O4FDev Are either of you able to reproduce this in 1.94? I was previously seeing this on iOS, but now it's working fine for me.

auroursa:
> @auroursa @O4FDev Are either of you able to reproduce this in 1.94? I was previously seeing this on iOS, but now it's working fine for me.

Yes, unfortunately, I am still able to reproduce this issue in 1.94, whether on Web, Android, or iOS.

Strangely, I can’t always reproduce it on iOS. It seems that under certain conditions, iOS handles the translation properly without returning to the homepage. But on the Web, though, I can almost reproduce the issue 100% of the time. I’ve tried resetting all settings, switching the language to English, and changing browsers (from Chrome to Firefox).

surfdude29:
I just tried in Safari on my iPad, and yep, it's definitely still an issue in 1.94 🙃 Maybe I just got lucky when it didn't happen in the iOS app?

> I'm kind of afraid to click the translate button right now because I'm worried it will take me back to the homepage and I won't be able to find the post I was just looking at.

It's annoying to have to do because it's an extra click/tap, but as a workaround, does it happen when you use Translate from the post context menu? I've never had it happen when using that.

O4FDev:
Hahah... yeah let's not ignore the translate button 😆. 

I'm not quite sure how to get my PR reviewed and merged that fixes this, just hoping the team sees this before it gets too far back into the pages

auroursa:
> Hahah... yeah let's not ignore the translate button 😆.
> 
> I'm not quite sure how to get my PR reviewed and merged that fixes this, just hoping the team sees this before it gets too far back into the pages

Don't worry, things will be resolved. They just have many other important things to take care of.

PR #6289, opened 2024-11-13
fix: Translate button sends users to homepage

O4FDev:
Fixes: #6287 

## Why 

As reported by @auroursa, the translate button on extended posts will send the user to the homepage unintentionally 
See behaviour here: https://bsky.app/profile/youareexisting.bsky.social/post/3lasuvhylqk2j


gaearon:
where is that # coming from? we should remove it there instead.

O4FDev:
> where is that # coming from? we should remove it there instead.

Ah, my bad - didn't realise there was already a way to handle this in there. 
Done

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -753,10 +753,9 @@ function ExpandedPostDetails({
             </Text>
 
             <InlineLinkText
-              to="#"
```
gaearon, line 756: should this just be `to={translatorUrl}`? any reason not to do that?

O4FDev: Good point! Will update

auroursa:
I recently found that this issue isn’t limited to the web platform—it can occur across all platforms, including iOS. To be honest, it’s a bit annoying.

auroursa:
@gaearon Sorry to bother you, could you take another look at this PR?

O4FDev:
I believe there might've been an attempt to fix this - folks are reporting this as still an issue in 1.94 but more inconsistently. I think the version @gaearon came up with (implemented in this PR) might work the best, as I've not been able to replicate with it.

surfdude29:
Just wondering, is this ready to merge? There's a message saying that this branch has conflicts 🤔

O4FDev:
> Just wondering, is this ready to merge? There's a message saying that this branch has conflicts 🤔

I'll quickly fix the conflict so that it's not there but I suspect the solution will literally just be deleting what's there unfortunately, not sure if the maintainers can do a force merge w/ this version 🤷🏻

O4FDev:
Closed based on @auroursa fix above being merged.

PR #6290, opened 2024-11-13
Add Toki Pona to list of languages

tbodt:
https://iso639-3.sil.org/code/tok


quiple:
It's a duplicate of #5998

auroursa:
In the current code, a languages without have code2 will not be available as options for the post language (as I mentioned in #5998). Unfortunately, Toki Pona is not included in the ISO 639-1 list, so it does not have code2.

Some other popular languages also lack code2, and it seems we may need to make more changes, possibly involving server-side adjustments. But I’m not sure about Bluesky's future development direction, so it might be best for the developers to handle this.

PR #6298, opened 2024-11-13
Patch RCTScrollView to fix centerContent

gaearon:


haileyok (approved):

patches/react-native+0.74.1.patch
```diff
@@ -9,7 +9,7 @@ index caa5540..c5d4e67 100644
 -                                                  type != nil && [type length] > 0 ? type : @"application/octet-stream",
 +                                                  ![type isEqual:[NSNull null]] && [type length] > 0 ? type : @"application/octet-stream",
                                                    [data base64EncodedStringWithOptions:0]];
-
+
```
gaearon, line 12: i just regenerated the patch. don't know what's up with these. i suspect someone might have edited it by hand before and stripped whitespace in the editor

PR #6300, opened 2024-11-13
Add language detection for `Dockerfile.*` files

spenserblack:
This lets GitHub know that files with the path `Dockerfile.*` are `Dockerfile`s, enabling syntax highlighting and language search tools.


kankenny (commented):
lgtm

PR #6307, opened 2024-11-13
Increase strength of check for post stats guard

estrattonbailey:
Some folks noticed during the outage earlier today that `quoteCount` wasn't returned on responses, which caused this section to display with no contents. This PR checks for the existence of these values, and the only shows the section if any of the values are > 0.

This should be a very rare case.


kankenny (commented):
LGTM

estrattonbailey:
Closing in favor of fixing this on the backend https://github.com/bluesky-social/atproto/pull/2990

PR #6314, opened 2024-11-13
Cherry-pick from last release branch

gaearon:
Just two commits we need to carry over

### DIFF
--- a/__e2e__/flows/composer-self-label.yml
+++ b/__e2e__/flows/composer-self-label.yml
@@ -17,8 +17,8 @@ appId: xyz.blueskyweb.app
 - inputText: "Post with an image"
 - tapOn:
     id: "openGalleryBtn"
-- tapOn:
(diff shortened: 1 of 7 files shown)


haileyok (approved):

surfdude29:
@gaearon @pfrazee Given that `Porn` has been changed to `Adult` in the self-label dialog, does this line in the test need to be changed too?
https://github.com/bluesky-social/social-app/blob/5f7caf30105569fe02cda2b029d3e6cb1e4e4780/__e2e__/flows/composer-self-label.yml#L22

surfdude29:
Also, should `Porn` be changed to `Adult` in the label for accessibility?
https://github.com/bluesky-social/social-app/blob/5f7caf30105569fe02cda2b029d3e6cb1e4e4780/src/view/com/composer/labels/LabelsBtn.tsx#L158-L161

PR #6333, opened 2024-11-14
NewChatDialog: remove extra padding on web

Simek:
# Why

Spotted that "New Chat" dialog container includes an extra padding on the web, which seems unintended. It also causes a visual clipping at the end of the users list.

# How

Modify the default inner container style for web by `webInnerContentContainerStyle` prop.

This fix also applies to `SendViaChat` dialog.

# Test Plan

Tested the changes by running app locally, made sure that the new styles does not affect native platforms.

# Preview

Screenshot 2024-11-14 at 14 31 39


mozzius (approved):
Great, thanks!

Issue #6322, opened 2024-11-14
Accessibility bug:  All features relying on a toast (like the share button) are inaccessible


SleeplessByte:
### Steps to Reproduce

1. Go to https://bsky.app/
2. Go to any post
3. Click the share post button
4. A toast will appear in the bottom left corner. This toast is **not** announced to any screenreader, and thus, it is as if nothing happens

Before I go into detail, here are people, some of which way smarter than I am, talking about this subject:

- https://adrianroselli.com/2019/06/scraping-burned-toast.html
- https://www.erikkroes.nl/blog/burn-your-toast/
- https://www.scottohara.me/blog/2019/07/08/a-toast-to-a11y-toasts.html

In short: if the information conveyed in the toast is _necessary_ (and it is right now), then this fails on all fronts. You probably want to ensure the toast is not needed at all, but if you don't want to get rid of it, you want to use a [Live Region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) to announce the information displayed on the screen.

This is what happens in my screen reader:

```text
button share post

*click*

(nothing happens)
``` 

- [ ] Clicking the share button should announce the outcome (content copied to clipboard)

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### Device Info

Not Relevant

### What version of the app are you using?

Web

### Additional Information

I have not tested this on the apps. My guess is that any snackbar or toast that's native to the OS will work (to announce, but PLEASE test this), however most custom implementations do not, so again, please test this.

- [ ] Test a screenreader on iOS to see if the information is announced
- [ ] Test a screenreader on Android to see if the information is announced

stubbornella:
On VoiceOver I don’t hear the copy to clipboard announced after clicking control-option-space 

This is a good fix to make since the rest of the action menu is really well done accessibility-wise and meets the criteria for menus.

ashleemboyer:
Accessibility consultant here, chiming in to +1! I'm seeing this today with VoiceOver. Tiny screen recording here (no sound):
https://bsky.app/profile/ashleemboyer.com/post/3lcxdi5pjp22j

PR #6345, opened 2024-11-14
Dismissable toasts

mozzius:
Swipe to dismiss toasts. Also means you can keep it alive to read if you tap and hold

+ a11y improvements!

https://github.com/user-attachments/assets/fe649fce-8924-4617-a71a-8b51484fba2e

### DIFF
--- a/src/view/com/util/Toast.tsx
+++ b/src/view/com/util/Toast.tsx
@@ -151,7 +151,9 @@ function Toast({
     const translation = dismissSwipeTranslateY.get()
     return {
       transform: [
-        {translateY: translation > 0 ? translation / 2 : translation},
+        {
+          translateY: translation > 0 ? translation ** 0.7 : translation,
+        },
       ],
     }
   })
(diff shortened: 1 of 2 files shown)


SleeplessByte:
Related: #6322.

If you don't want to do the work to make these accessible now, you could add a fallback [accessibility action](https://reactnative.dev/docs/accessibility#accessibility-actions) so that people who cannot swipe, can still dismiss the toast sooner than the timeout. This does not make the toast accessible, but makes it a tiny bit less bad for a particular subset of users.

mozzius:
> Related: #6322.
> 
> If you don't want to do the work to make these accessible now, you could add a fallback [accessibility action](https://reactnative.dev/docs/accessibility#accessibility-actions) so that people who cannot swipe, can still dismiss the toast sooner than the timeout. This does not make the toast accessible, but makes it a tiny bit less bad for a particular subset of users. The TL;DR of the linked page is that swiping is not something that people universally can do, adding an `accessibilityAction` allows you to expose the dismiss via assistive technology as an alternative to the swipe.

good shout! pushed some a11y improvements too

gaearon (approved):
seems legit

PR #6349, opened 2024-11-14
Better error message (+ translations!)

mozzius:
- Added `NotEnoughResources` and `pipethrough network error` to the generic server error message
- Translated errors

imo it's fine to use the `t` macro here, since error messages should not be sticking around in the tree, and imo it's an acceptable tradeoff if maybe you see one in the wrong language immediately after switching langs, vs current behaviour

## Potential problems

Potential problems might arise from checking against the output of `cleanError` and expecting one of the now-translated strings to still be in english. however I did a cursory check of the references and it seems fine

### DIFF
--- a/src/lib/strings/errors.ts
+++ b/src/lib/strings/errors.ts
@@ -1,3 +1,5 @@
+import {t} from '@lingui/macro'
+
 export function cleanError(str: any): string {
   if (!str) {
     return ''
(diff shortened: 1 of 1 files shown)


gaearon (approved):

gaearon (approved):
have not tested

PR #6378, opened 2024-11-15
Reduce `listConvos` requests

estrattonbailey:
This PR:
1. Removes 60s polling of `listConvos` and replaces with 60s polling of `getLog`
2. Replaces query invalidation when marking as read with an optimistic cache update
3. Reduces page size of `listConvos` from default `50` to `20`, and reduces max unread count from `30+` to `10+`

# 1 has a small drawback in that if you read a conversation on another device, other devices won't reflect that read count until you refocus the chat list. However, the other direction works: if a new message comes in, unread count is incremented.


mozzius (approved):
lgtm not tested but looks safe

Issue #6365, opened 2024-11-15
Newline handling in web client is not correct.


olafurw:
### Steps to Reproduce

So, I cross post to Bluesky, Mastodon, LinkedIn and Threads. I do this by just writing the post in 1 client and copy/pasting it to the others.

But if I write the text like in the first image and paste it to Bluesky, then you remove the extra line

![image](https://github.com/user-attachments/assets/3c19b109-ee33-416e-a680-8354075f9294)

And if I type out the text in Bluesky, it will look like this if I paste it into other clients.

![image](https://github.com/user-attachments/assets/87a769fd-0e52-4c35-9a88-004286798fe5)

There is something you're doing to the newlines which makes the output incorrect.

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 7731136 (prod); Bundle date: 24110818; Platform: web

PR #6379, opened 2024-11-15
Paste html as html

airowe:
This fix addresses https://github.com/bluesky-social/social-app/issues/6365 by pasting incoming html text as html instead of parsing the raw text out of the pasted clipboard data.

If there was a reason for only wanting to paste raw text, please close this PR but pasting as HTML allows the @tiptap extensions to take over any formatting requirements (i.e. this one issues require the Hardbreak extension to be applied but since we're pasting as raw text, the newline characters (<br>) are stripped out.

This PR also adds the `sanitize-html` library to try to reduce insanity from parsed html.

***The addition of the above library allows us to restrict html tags to only `<br>` tags to minimize hacking vectors for potential script injection***


gaearon:
How much is this adding to the bundle size?

airowe:
> How much is this adding to the bundle size?

Per bundlephobia: https://bundlephobia.com/package/sanitize-html@2.13.1

194.8kB Minified
80.9kB Minified + Gzipped

I'll push up a change with a smaller bundle sized library shortly

airowe:
@gaearon Updated with `DomPurify` library, much smaller footprint: https://bundlephobia.com/package/dompurify@3.2.0

image

HorusGoul:
Hey! We don't need to introduce a new package to fix this. I tried using Tiptap configuration to configure parsing and serializing when copying and pasting, and it works! 

Opened another PR against main: #6658

mozzius:
I think this is superceded by https://github.com/bluesky-social/social-app/pull/6658, which got merged

PR #6381, opened 2024-11-15
Reduce calls to getLogs

estrattonbailey:
Stacked on #6378

Reduces poll interval for active chats from 3s to 4s, and polling on the messages list screen from 10s to 30s. Background polling doesn't matter here, all platforms pause polling when backgrounded atm.


mozzius (approved):
sure

PR #6391, opened 2024-11-15
Use compiler-safe Reanimated get/set APIs

gaearon:
Manually went through our codebase and fixed up all cases. I used find-all search for `useSharedValue`, `useDerivedValue`, and `.value` accessors.

Ran into a Reanimated bug with `set()` functions, but it was easy to work around: https://github.com/software-mansion/react-native-reanimated/issues/6613#issuecomment-2479870214. You just have to add `() =>` when setting it to animations, like `set(() => withStuff())`.

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/6391/files?w=1)

## Test Plan

Tried to find each of these components on iOS, verified common interactions work. Tried most of these Android, wouldn't expect platform differences with this API switch.

Look closely through the code: `set(withFoo())` would fail. It has to be `set(() => withFoo())`. However, `set(x)` is fine.

### DIFF
--- a/src/lib/custom-animations/PressableScale.tsx
+++ b/src/lib/custom-animations/PressableScale.tsx
@@ -2,7 +2,6 @@ import React from 'react'
 import {Pressable, PressableProps, StyleProp, ViewStyle} from 'react-native'
 import Animated, {
   cancelAnimation,
-  runOnJS,
   useAnimatedStyle,
   useReducedMotion,
   useSharedValue,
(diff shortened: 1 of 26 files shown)


src/lib/custom-animations/PressableScale.tsx
```diff
@@ -32,27 +31,25 @@ export function PressableScale({
...
   const animatedStyle = useAnimatedStyle(() => ({
-    transform: [{scale: scale.value}],
+    transform: [{scale: scale.get()}],
   }))
 
   return (
     <AnimatedPressable
       accessibilityRole="button"
       onPressIn={e => {
-        'worklet'
```
gaearon, line 42: This wasn't a worklet. onPressIn/onPressOut run on JS already.

src/view/com/composer/Composer.tsx
```diff
@@ -1308,21 +1308,22 @@ function useScrollTracker({
...
+      'worklet'
+      const oldContentHeight = contentHeight.get()
+      let shouldScrollToBottom = false
+      if (stickyBottom && newContentHeight > oldContentHeight) {
         const isFairlyCloseToBottom =
-          contentHeight.value - contentOffset.value - 100 <=
-          scrollViewHeight.value
+          oldContentHeight - contentOffset.get() - 100 <= scrollViewHeight.get()
         if (isFairlyCloseToBottom) {
-          runOnUI(() => {
```
gaearon, line 1318: I've refactored this to run the entire parent block on the UI thread since we've learned that otherwise every `.value` access does a thread hop. Might as well run everything on UI.

src/lib/custom-animations/PressableScale.tsx
```diff
@@ -32,27 +31,25 @@ export function PressableScale({
...
         if (onPressIn) {
-          runOnJS(onPressIn)(e)
+          onPressIn(e)
         }
         cancelAnimation(scale)
-        scale.value = withTiming(targetScale, {duration: 100})
+        scale.set(() => withTiming(targetScale, {duration: 100}))
```
mozzius, line 45: is `cancelAnimation` blocking? do they need to be grouped together into a single `runOnUI` call?

gaearon: i think it is. don't want to change this right now, i need to better understand what `cancelAnimation` does. we've heard we're not supposed to use it like this but i do remember some need for it. let's not touch for now?

mozzius (approved):
Changes look good. Clicked around the app a bit and didn't run into any problems

PR #6392, opened 2024-11-15
Use native modal for signup queue screen

mozzius:
The signup queue has slightly broken styles - no top inset, and the log out button is overlaid on top of the bottom bar, which shouldn't be visible at all

We can fix this by using a nice native modal, which also makes it feel a lot nicer - it includes entry/exit animations and feels more like a "gate"

Also includes minor style tweaks like a border on the gray box

# iOS

https://github.com/user-attachments/assets/b028b8fc-dd6d-4e4a-a0b4-bf233d8382b6

# Android

https://github.com/user-attachments/assets/dc0379e1-13d7-4e63-8ac8-56fededf6cac

# Test plan

in `src/state/session/util.ts` replace `isSignupQueued` with

```
let temp = true

export function isSignupQueued(accessJwt: string | undefined) {
  if (accessJwt) {
    setTimeout(() => {
      temp = false
    }, 10000)

    return temp
  }
  return false
}
```

This will show the gate initially and after 10 seconds, will let you past when you press the button. If you want a "real" test, sign out beforehand and then go through the signup flow

Test all 3 platforms. Make sure there's no weirdness of the modal on android (there shouldn't be, since everything is very simple here, no gesture/keyboard bullshit)

### DIFF
--- a/src/screens/SignupQueued.tsx
+++ b/src/screens/SignupQueued.tsx
@@ -95,7 +95,8 @@ export function SignupQueued() {
       {isIOS && <StatusBar style="light" />}
       <ScrollView
         style={[a.flex_1, t.atoms.bg]}
-        contentContainerStyle={{borderWidth: 0}}>
+        contentContainerStyle={{borderWidth: 0}}
+        bounces={false}>
         <View
           style={[
             a.flex_row,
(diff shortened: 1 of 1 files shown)


gaearon (approved):

gaearon:
works in my testing. it's a bit weird that on iOS you can drag the butterfly behind:

https://github.com/user-attachments/assets/9e9afd9f-22ca-4980-bfd1-ba57ee69afc4

no big deal so up to you whether to do something, feel free to merge

PR #6402, opened 2024-11-16
Remove aria-pressed attr from button. Closes #6325

### LINKED ISSUE #6325: Accessibility bug: buttons are spamming aria-pressed but they are not toggle buttons
### Steps to Reproduce

This is probably a bug in react-native-web, but that's no excuse.

1. Go to https://bsky.app/
2. Find any `<button>`
3. It will have `aria-pressed="false"`
4. Clicking it (mouse down / keyboard hold down) will switch it to `aria-pressed="true"`
5. Releasing will switch it back to `aria-pressed="false"`

This one is easy. It's wrong.

From https://w3c.github.io/aria/#aria-pressed and https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

> Adding `aria-pressed` to an element with a role of [button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) turns the `button` into a toggle button. The `aria-pressed` attribute is only relevant for toggle buttons. It represents the button's current "pressed" state.

These are not toggle buttons. 

- [ ] `aria-pressed` is not added to buttons that are not toggle buttons

### Attachments

Relevant html

```html
<button aria-pressed="false" role="button" tabindex="0"></button>
```

Web (Desktop), Web (Mobile)

### Device Info

Not Relevant

Web

### Additional Information

The problem here is that people may have turned on options that announce something when the button becomes "toggled-on". You don't want that.

PaulACoroneos:
Removes aria-pressed attr from button element as this attribute is [designed for usage with checkbox inputs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed). Closes #6325

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -458,7 +458,6 @@ export const Button = React.forwardRef<View, ButtonProps>(
         // @ts-ignore - this will always be a pressable
         ref={ref}
         aria-label={label}
-        aria-pressed={state.pressed}
         accessibilityLabel={label}
         disabled={disabled || false}
         accessibilityState={{


mozzius (approved):
Lovely, thank you!

Issue #6404, opened 2024-11-16
[Profile] Empty state on Feeds/Lists tabs flickers when swiping to them after they've loaded once


alyosha:
### Steps to Reproduce

1. Navigate to either Feeds/Lists tab on Profile with a user who has no relevant data
2. Swipe away from the tab once
3. Swipe back to the tab and observe empty state flicker

### Attachments

https://github.com/user-attachments/assets/32512bc1-b92d-4c44-9902-a9d5f800479f

### What platform(s) does this occur on?

iOS, Android, Web (Desktop), Web (Mobile)

### What version of the app are you using?

Build version: 1.93.0.373; Bundle info: ee184d2cd (prod); Bundle date: 24111518; Platform: android

### Additional Information

Should repro easily given the above steps. I tested on Android and web. 

Does not seem to be an issue on the Posts/Replies/Media/Starter Packs tabs. Likes tab has its own issues with the looping API calls, but that's being handled elsewhere.

Did my best to comb through the existing issues and confirm this isn't a dupe -- apologies in advance if I missed something.

PR #6406, opened 2024-11-16
[Profile] Prevent flicker on Lists/Feeds tabs

alyosha:
# Summary

Fixes https://github.com/bluesky-social/social-app/issues/6404

There is currently a flicker on both the `Feeds` and `Lists` tabs when swiping back to them after they've been loaded once (if they are empty)

Upon further investigation I realized the reason we don't see similar flickering for `Posts` / `Replies` / `Media` is because of `staleTime` specification in query options. Added for the queries on these two tabs as well.

# Visual

**BEFORE**

https://github.com/user-attachments/assets/653f07c5-d4dc-4fed-b226-44d4faa01495

**AFTER**

https://github.com/user-attachments/assets/1f29cc3f-f6d4-4034-9e8b-f2a1dd3fd67c

### DIFF
--- a/src/view/com/feeds/ProfileFeedgens.tsx
+++ b/src/view/com/feeds/ProfileFeedgens.tsx
@@ -1,4 +1,4 @@
-import React, {useRef} from 'react'
+import React from 'react'
 import {
   findNodeHandle,
   ListRenderItemInfo,
@@ -66,16 +66,12 @@ export const ProfileFeedgens = React.forwardRef<
   const isEmpty = !isFetching && !data?.pages[0]?.feeds.length
   const {data: preferences} = usePreferencesQuery()
 
-  const lastItemsState = useRef<any[]>([])
-
   const items = React.useMemo(() => {
     let items: any[] = []
     if (isError && isEmpty) {
       items = items.concat([ERROR_ITEM])
     }
-    if (isFetched && isFetching) {
-      return lastItemsState.current
-    } else if (!isFetched && isFetching) {
+    if (!isFetched && isFetching) {
       items = items.concat([LOADING])
     } else if (isEmpty) {
       items = items.concat([EMPTY])
@@ -89,7 +85,6 @@ export const ProfileFeedgens = React.forwardRef<
     return items
   }, [isError, isEmpty, isFetched, isFetching, data])
 
-  lastItemsState.current = items
   // events
   // =
 
--- a/src/view/com/feeds/ProfileFeedgens.tsx
+++ b/src/view/com/feeds/ProfileFeedgens.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useRef} from 'react'
 import {
   findNodeHandle,
   ListRenderItemInfo,
@@ -66,12 +66,16 @@ export const ProfileFeedgens = React.forwardRef<
   const isEmpty = !isFetching && !data?.pages[0]?.feeds.length
   const {data: preferences} = usePreferencesQuery()
 
+  const lastItemsState = useRef<any[]>([])
+
   const items = React.useMemo(() => {
     let items: any[] = []
     if (isError && isEmpty) {
       items = items.concat([ERROR_ITEM])
     }
-    if (!isFetched && isFetching) {
+    if (isFetched && isFetching) {
+      return lastItemsState.current
+    } else if (!isFetched && isFetching) {
       items = items.concat([LOADING])
     } else if (isEmpty) {
       items = items.concat([EMPTY])
@@ -85,6 +89,7 @@ export const ProfileFeedgens = React.forwardRef<
     return items
   }, [isError, isEmpty, isFetched, isFetching, data])
 
+  lastItemsState.current = items
   // events
   // =
 
--- a/src/view/com/lists/ProfileLists.tsx
+++ b/src/view/com/lists/ProfileLists.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useRef} from 'react'
 import {
   findNodeHandle,
   ListRenderItemInfo,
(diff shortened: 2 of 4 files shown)


src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -85,6 +89,7 @@ export const ProfileFeedgens = React.forwardRef<
     return items
   }, [isError, isEmpty, isFetched, isFetching, data])
 
+  lastItemsState.current = items
```
mozzius, line 92: Setting refs during render is not ideal - it will cause React Compiler to deoptimise the component

https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -63,17 +63,18 @@ export const ProfileFeedgens = React.forwardRef<
     error,
     refetch,
   } = useProfileFeedgensQuery(did, opts)
-  const isEmpty = !isFetching && !data?.pages[0]?.feeds.length
   const {data: preferences} = usePreferencesQuery()
 
   const items = React.useMemo(() => {
+    const noListEntries = !data?.pages[0]?.feeds.length
+    const isEmpty = !isFetching && noListEntries
```
alyosha, line 70: naming of `isEmpty` here is a bit confusing, will clean up later

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -63,27 +63,27 @@ export const ProfileFeedgens = React.forwardRef<
...
+    const hasListData = data?.pages[0]?.feeds.length
     let items: any[] = []
-    if (isError && isEmpty) {
+    if (isError && !isFetching && !hasListData) {
       items = items.concat([ERROR_ITEM])
     }
     if (!isFetched && isFetching) {
       items = items.concat([LOADING])
-    } else if (isEmpty) {
+    } else if (isFetched && !hasListData) {
```
alyosha, line 76: Hiding the component when fetching is the cause of the bug, so stop making display dependent on `isFetching`

One side effect of this change is that the empty state will not hide during re-fetch from `ErrorMessage`.  Can address that if it seems like an issue

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -63,27 +63,27 @@ export const ProfileFeedgens = React.forwardRef<
...
       items = items.concat([EMPTY])
     } else if (data?.pages) {
       for (const page of data?.pages) {
         items = items.concat(page.feeds)
       }
-    } else if (isError && !isEmpty) {
+    } else if (isError && !isFetching && hasListData) {
```
alyosha, line 82: The way this original logic was written, `LOAD_MORE_ERROR_ITEM` would be added whenever `isError` + `isFetching` are both true -- that did not seem like intentional behavior to me and seems like it could cause some weird bugs, for example if `isError` remained true while attempting refetch from `ErrorMessage`

Can revert this change if I'm missing something

src/view/com/lists/ProfileLists.tsx
```diff
@@ -60,27 +60,26 @@ export const ProfileLists = React.forwardRef<SectionRef, ProfileListsProps>(
...
-      } else if (isEmpty) {
+      } else if (isFetched && !hasListData) {
         items = items.concat([EMPTY])
       } else if (data?.pages) {
         for (const page of data?.pages) {
           items = items.concat(page.lists)
         }
       }
-      if (isError && !isEmpty) {
+      if (isError && !isFetching && hasListData) {
```
alyosha, line 78: The way this original logic was written, `LOAD_MORE_ERROR_ITEM` would be added whenever `isError` + `isFetching` were both true 

That behavior did not seem intentional to me as it has the potential to cause some weird bugs (for example if `isError` remains true while refetching from `ErrorMessage`, then `LoadMoreRetryBtn` will flash during the API call)

Can revert the change if I'm missing something

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -63,27 +63,27 @@ export const ProfileFeedgens = React.forwardRef<
...
     refetch,
   } = useProfileFeedgensQuery(did, opts)
-  const isEmpty = !isFetching && !data?.pages[0]?.feeds.length
   const {data: preferences} = usePreferencesQuery()
 
   const items = React.useMemo(() => {
+    const hasListData = data?.pages[0]?.feeds.length
     let items: any[] = []
-    if (isError && isEmpty) {
+    if (isError && !isFetching && !hasListData) {
```
alyosha, line 71: no change in behavior

src/view/com/feeds/ProfileFeedgens.tsx
alyosha, line 82: is it intentional we use `else if` here but not in `ProfileLists`?

alyosha:
@mozzius thanks for the feedback. I updated the implementation and would appreciate your thoughts whenever you have a second :pray: 

Sorry for the commit history, I can squash if that's better

alyosha:
Upon further investigation I realized the reason we don't see similar flickering for `Posts` / `Replies` / `Media` is because of `staleTime` specification. Seemed better to align with that for behavioral consistency, sorry again for the commit mess, I can squash whenever.

gaearon:
I wonder if we should actually be making more things "fresh"...

With this change, do they get refetched when you create/edit a list?

alyosha:
@gaearon thanks for the review :raised_hands:  

I tested locally and it seems that `/xrpc/app.bsky.graph.getLists` is called after creating or editing a list, meaning the content is already refretched the next time you check the profile tab. Should resolve the bug without impacting freshness in my understanding  :+1:  I'm a little less familiar with feeds, but it seems like it's not possible to create them in app? Maybe I missed something, please let me know I'm happy to test.

alyosha:
@mozzius @gaearon know you're super busy so no rush, but let me know if there's anything i can do on my end to move this one along. Also happy to close if you're thinking of a different approach 🙂

PR #6419, opened 2024-11-16
Manually implement `drawerType` logic

mozzius:
Seems like when upgrading react-native-drawer-layout there was a regression where `drawerType` is always `front` despite what the docs say. It should be `slide` on iOS

# Test plan

Confirm that it is indeed broken on iOS without this change
Confirm this change fixes it


haileyok (approved):
didnt test

mozzius:
Issue filed
https://github.com/react-navigation/react-navigation/issues/12266

PR #6420, opened 2024-11-16
Add manifest.json for basic PWA support

WesSouza:
This introduces a `manifest.json` to `bskyweb`, allowing the website to be installed as a "WebClip"/PWA application on iOS and iPadOS.

This also replaces the `apple-touch-icon.png` with the application icon, which then renders much better when added as a PWA.

Screen capture of the XCode Simulator showing Bluesky full screen on an iPad mini

Screen capture of the XCode Simulator showing Bluesky full screen on an iPhone

### Caveats

When running as a standalone PWA on iPhone, the safe area must be supported via the `env(safe-area-inset-*)` CSS variables. Unfortunately I wasn't able to quickly make that work together with `useSafeAreaInsets` from `react-native-safe-area-context`.

I was also very confused on how `yarn start` plays with the `bskyweb` app. I guess the team is overriding Expo's standard for public assets.

To test this, I built the website and started the Go server in `bskyweb`.

---

I will add that I am absolutely flabbergasted to have been able to have this working as a logged in user on my local machine. Amazing job on DX!


v42:
🚢

bskyweb/static/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "Bluesky",
+  "icons": [
+    {
+      "src": "/static/apple-touch-icon.png",
+      "type": "image/png",
+      "sizes": "1024x1024"
+    }
+  ],
+  "start_url": "/",
+  "display": "standalone",
```
rektide, line 12:
```suggestion
  "display": "standalone",
  "display_override": ["tabbed"],
```

Since this is a social network app and numberous links are likely to be clicked, my opinion is that we should open with tabs too.

https://developer.chrome.com/docs/capabilities/tabbed-application-mode#using_tabbed_application_mode

WesSouza: I'm not sure this is a good idea given this property is experimental:

https://developer.mozilla.org/en-US/docs/Web/Manifest/display_override

jswhisperer:
I installed Bluesky as a PWA no tabbed property, and really enjoy it 110% more 
image
image

I used overrides rather than this PR but think it's important a PWA does exist.

jswhisperer:
MacOS m1

PR #6424, opened 2024-11-16
Remove typography NUX and consistently call upsertProfile

gaearon:
We're removing the typography NUX (since enough people have seen it) and fixing some bugs with how NUXes are presented. In particular, this solves (going forward) the issue where a missing profile record could cause the NUX to appear for users who weren't supposed to get it. After this PR, profiles without a record never get any NUXes.

This is easier to review commit by commit.

- https://github.com/bluesky-social/social-app/commit/5540a6a51810c2565704622a9e1a33d2de7f2615?w=1: The upserting shouldn't be conditional, or we'll lose the `createdAt` and the `starterPack`. Let's just always do it. I don't know whether this can reproduce in practice but this is more correct anyway.
- 6771f5deaa4a370bc70af61b367c8801bd6677f7: We don't want this NUX anymore.
- 573f74c17b9320e1e903689d66ff6c2fd6da36e0: This expands the NUX condition to always exit early if `profile.createdAt` is missing. Since we don't know if the user is new or not. Note that due to an AppView bug (?), we're currently getting `0001-01-01T00:00:00.000Z` for accounts with no profile record instead of `undefined`. But @devinivy said this can/will be fixed on the AppView side. For now let's hardcode both.

## Test Plan

Add a `console.log(currentProfile.createdAt)` before `for (const {id, enabled} of queuedNuxs) {`.

Comment out `isSnoozed` implementation to return `false` for testing.

Sign up as a new user, verify no NUX. Verify the log does appear though (with today's date).

Verify sign up from both with custom avatar and with default avatar.

The comment out the `upsertAccount` call altogether and sign up again. Verify `console.log` does not run (because `createdAt` is bad).


src/components/dialogs/nuxs/index.tsx
```diff
@@ -66,7 +53,13 @@ export function NuxDialogs() {
   const onboardingActive = useOnboardingState().isActive
 
   const isLoading =
-    !currentAccount || !preferences || !profile || onboardingActive
+    onboardingActive ||
+    !currentAccount ||
+    !preferences ||
+    !profile ||
+    !profile.createdAt ||
+    profile.createdAt === '0001-01-01T00:00:00.000Z' // TODO: Fix this in AppView.
```
estrattonbailey, line 61: Nit, maybe add a comment here to the effect of: profile isn't legit ready until `createdAt` is a real date

estrattonbailey (approved):
Nice sleuthing

PR #6427, opened 2024-11-16
There’s no Moldavian/Moldovan language

claudiu-cristea:
The, so called, “Moldavian/Moldovan language” is actually Romanian language. The term is a Russian colonial relic. People from Republic of Moldova are speaking mostly Romanian along with other languages of national minorities (Russian, Ukrainian, Gagauz, etc).

Pretending that Moldavians are speaking Moldavian/Moldovan it’s like saying that Austrians are speaking “Austrian language”, not German or people from the United States are speaking mostly “American language”, not English. True, Republic of Moldova is an independent and sovereign state but it shares with Romania the same language of the majority of its citizens: Romanian language.

This is officially stated by the Constitution of the Republic of Moldova at Article 13, point 1:

> The state language of the Republic of Moldova is Romanian, operating on the basis of Latin script

Se also:

* https://en.wikipedia.org/wiki/Languages_of_Moldova


pfrazee (approved):
Cool, thanks for the clarificaiton!

PR #6433, opened 2024-11-16
Add gif support to web

mozzius:
Allows adding gifs on web - processed as video.

Also reimplements the video picker, since `expo-image-picker`'s implementation on web is absolutely terrible, does not support half the props, and does not support changing the accepted mime types. This change greatly simplifies the logic for normal videos too, since the preview no longer needs to pull double duty detecting the image dimensions or duration - this has been moved to `getVideoMetadata` which is called when the video is first picked. This can and should still be simplified further, since we're inefficiently passing dataurls around, but this is easy for now since it lets us maintain compatibility with the `ImagePickerAsset` type

>[!NOTE]
>I don't have a way of detecting how long a gif is (i.e. it could be not even animated, it could be >60s) so we just have to let the server reject these. Apparently it seems you can parse the bytes manually to determine this? Hopefully that's uncommon enough that we can get away with not rejecting on the client

TODO (but out of scope for this PR): different presentation for gifs. needs lexicon change.

https://github.com/user-attachments/assets/5d6833f0-438a-49a8-8436-c99030066328

# Test plan

- Upload a gif
- Drag and drop/paste a gif
- Upload a video
- Drag and drop/paste a video
- For above, check that constraints are still respected - 50mb, 60s (but not for gifs, see above)

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1023,17 +1023,6 @@ function ComposerEmbeds({
                   asset={video.asset}
                   video={video.video}
                   isActivePost={isActivePost}
-                  setDimensions={(width: number, height: number) => {
-                    dispatch({
-                      type: 'embed_update_video',
-                      videoAction: {
-                        type: 'update_dimensions',
-                        width,
-                        height,
-                        signal: video.abortController.signal,
-                      },
-                    })
-                  }}
                   clear={clearVideo}
                 />
               ) : null)}
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -56,13 +56,18 @@ import {useQueryClient} from '@tanstack/react-query'
 import * as apilib from '#/lib/api/index'
 import {EmbeddingDisabledError} from '#/lib/api/resolve'
 import {until} from '#/lib/async/until'
-import {MAX_GRAPHEME_LENGTH} from '#/lib/constants'
+import {
+  MAX_GRAPHEME_LENGTH,
+  SUPPORTED_MIME_TYPES,
+  SupportedMimeTypes,
+} from '#/lib/constants'
 import {useAnimatedScrollHandler} from '#/lib/hooks/useAnimatedScrollHandler_FIXED'
 import {useEmail} from '#/lib/hooks/useEmail'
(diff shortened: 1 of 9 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -744,14 +750,24 @@ let ComposerPost = React.memo(function ComposerPost({
 
   const onPhotoPasted = useCallback(
     async (uri: string) => {
-      if (uri.startsWith('data:video/')) {
-        onSelectVideo(post.id, {uri, type: 'video', height: 0, width: 0})
+      if (uri.startsWith('data:video/') || uri.startsWith('data:image/gif')) {
+        if (isNative) return // web only
```
mozzius, line 754: is it possible to paste videos on native? do we need to handle this? it will be broken in prod anyway since the old code sets the height/width to 0

auroursa:
It currently looks very much like uploading a normal video, with a playback control bar and all—which I believe GIFs don’t really need.

Additionally, slightly longer GIFs don’t seem to loop automatically and require clicking play again. Shouldn’t GIFs loop automatically?

mozzius:
>TODO (but out of scope for this PR): different presentation for gifs. needs lexicon change.

auroursa:
> >TODO (but out of scope for this PR): different presentation for gifs. needs lexicon change.

I sincerely apologize for misunderstood the meaning.

mozzius:
All good, don't worry about it

TodePond (commented):
looks fab lets go

gaearon (approved):
ok but needs cross browser testing

mozzius:
tested cross browser (ff, chrome, safari, mobile safari)

src/view/com/composer/videos/pickVideo.web.ts
```diff
@@ -0,0 +1,94 @@
...
+            width: img.width,
+            height: img.height,
+            // todo: calculate gif duration. seems possible if you read the bytes
+            // https://codepen.io/Ryman/pen/nZpYwY
+            // for now let's just let the server reject it, since that seems uncommon -sfn
+            duration: null,
+          })
+        }
+        img.onerror = (_ev, _source, _lineno, _colno, error) => {
+          console.log('Failed to grab GIF metadata', error)
```
BojanKogoj, line 63: Any reason why console.log is left in here?

PR #6453, opened 2024-11-17
Add profile hover cards to more screens

bdbch:
This PR adds ProfileHoverCards to the following screens:

- Notifications (On the first author & on each authors names in an expanded notification)
- Starter Pack profile list
- Starter Pack wizard profile list

Specially inside the Starter Pack wizard it was really hard to quickly see who a person is because you also couldn't open a profile in a new tab from there. This should help to quickly see what person you're adding to your Starter Pack.


bdbch:
Looking forward for feedback on this. I'd love to see more profile cards in a few places (as described above).

gaearon:
- Notification: Makes sense, if this was a separate PR I'd just merge that in.
- Starter Pack: Doesn't make sense to me, it's very duplicative. Would be better to extend the rows if we really need to.
- Starter Pack Wizard: I'm not sure it's worth doing it there. It doesn't work very well because it listens to the wrong scroll target, so it doesn't dismiss on scroll. This would have to be fixed. Still, what we *should* fix there is opening in a new tab.

gaearon (changes requested):
let's do it just in notifications. and let's instead fix starter pack wizard to use links / open in new tab

bdbch:
Sure thing, will do. Should I look into linking for new tabs for starter packs in a new PR?

gaearon:
>Should I look into linking for new tabs for starter packs in a new PR?

Yea probably best to separate.

bdbch:
Reverted my changes made to the starter pack wizard & the profile card component, but want to fix the anchor for the profile card. Right now the position is a bit off. Will update when I'm done.

image

bdbch:
Done! 👍

bdbch:
Happy new year!

Anything I could do to move this PR forward?

bdbch:
Already a year! Anything I can do to make progress on the PR? Otherwise I'd close it fit it's no longer needed.

bdbch:
Closing this for now - feel free to take over if wanted.

PR #6460, opened 2024-11-17
Upgrade React Compiler (take two)

gaearon:
Reapplying https://github.com/bluesky-social/social-app/pull/5898 but with a more recent version which should have the fix for https://github.com/facebook/react/issues/31402.

## Test Plan

TBD. I'll walk around the app and verify we don't have the same regression.


gaearon:
Confirmed the avatar problem we uncovered previously no longer reproduces.

mozzius (approved):
not tested extensively but seems ok

gaearon:
@poteto `npx react-devtools` (6.0.1-[c7c68ef842](https://github.com/facebook/react/commit/c7c68ef842)) does not show the memo badges for the React Native bundle. Is that expected? I do see the `Symbol.for("react.memo_cache_sentinel")` things in the actual bundle.

Malickmimd:
great work, admire you all specially @gaearon hope one day i'll be in your level to contribute

PR #6474, opened 2024-11-17
Fix layout jumps for images

gaearon:
I've noticed the `aspectRatio` value here fluctuates between user-submitted measurement vs measurement on fetch. The fluctuation is small but this leads to an extra setState and potentially to tiny layout jumps. And if the aspect ratio was not specified at all, the layout jump can be significant — per each such image.

Screenshot 2024-11-17 at 22 58 01

We should *never* layout jump in feed. So let's either use the provided aspect, or assume it's a square.

Also lets us avoid a setState.

## Test Plan

Verify images show up appropriately sized, as before (try tall, wide, square images).

Comment out the `= image.aspectRatio` part and verify that missing measurements are handled gracefully.

Verify lightbox works in all cases.

### DIFF
--- a/src/view/com/util/images/AutoSizedImage.tsx
+++ b/src/view/com/util/images/AutoSizedImage.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useRef} from 'react'
 import {DimensionValue, Pressable, View} from 'react-native'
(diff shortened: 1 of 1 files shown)


gaearon:
gave heads up to Paul, he's cool

PR #6488, opened 2024-11-18
Updated Irish translations

aindriu80:
Added missing translations, fixed some minor punctuation.  

** Regarding current translations - The current translation uses  "Mol" for "like" and "Molta" for "liked" however although not incorrect "Taitin" for "like" (verb) and "Taitnigh" for "liked" might be a better choice for a social media app such as Bluesky.

### DIFF
--- a/src/locale/locales/ga/messages.po
+++ b/src/locale/locales/ga/messages.po
@@ -103,7 +103,7 @@ msgstr "{0, plural, one {Dímhol (# mholadh)} two {Dímhol (# mholadh)} few {Dí
 
 #: src/screens/Settings/Settings.tsx:414
 msgid "{0}"
-msgstr ""
+msgstr "{0}"
 
 #. Pattern: {wordValue} in tags
 #: src/components/dialogs/MutedWords.tsx:475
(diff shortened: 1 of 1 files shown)


surfdude29:
Pinging lead Irish translator @kscanne for input

kscanne:
A Aindriú a chara (domhan beag!)

First, regarding "like"...  the three of us (me, @aonghuso, and Donncha King) working on the localization discussed the various possibilities and landed on "mol" in part because there are various places in the interface where "like" and "likes" are used nominally, and "moltaí" works nicely.  I wouldn't have a good solution if the verb is "taitin", or "is maith liom..." as they use on Facebook.

Regarding the translations in this pull request, many of them work, but many others conflict with the terminology we've established for the localization overall, and there are even some internal inconsistencies I see in the translations: "app password" becomes pasfhocal aip, pasfhocal aipe, pasfhocal aipeanna, pasfhocal feidhmchláir in different strings; similarly for "copy" as a verb, sometimes "cóip" and sometimes "cóipeáil"; "handle" should be "leasainm" not "láimhseáil", etc. — honestly it looks to me like these were Google Translated which isn't acceptable for user-facing localization.

The plural forms are also incorrect — Irish uses 5.... one, two, few, many, other.... which are needed to handle initial mutations in all cases.

@surfdude29  — given the above, I'd ask that these not be merged in.  As it happens, the three of us who've done the earlier work have most of these complete already and they're being proofread.  Can submit a patch in the next few days I hope.

surfdude29:
@kscanne I apologise if I gave the wrong impression, but to clarify: I'm a volunteer fellow translator like you; I don't have any decision-making power over what gets merged or doesn't. That's up to the Bluesky team.

I just wanted to let you know about the PR :)

PR #6490, opened 2024-11-18
Migrate from yarn classic to yarn berry.

### LINKED ISSUE #6478: Migrate from yarn classic to yarn berry (or even  pnpm)
### Describe the Feature

Hello. I was a little surprised to see bsky social app still uses yarn classic versus the more modern yarn berry or even pnpm. Is this something the team would be open to moving toward? Thank you!

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

PaulACoroneos:
Closes #6478 

References:

[yarn classic to berry migration guide](https://yarnpkg.com/migration/guide)
[why yarn berry over classic](https://yarnpkg.com/migration/overview)


PaulACoroneos:
I at least ran "web" locally and clicked around localhost and things seems fine. Unfortunately I am not having as much luck with native right now (getting errors I don't quite understand). I will have to try to figure that out in next several days but it was before this change so its likely unrelated since its failure is not during yarn install

gaearon:
Are there any other big projects that migrated and are happy with the experience? I never used Berry before

PaulACoroneos:
> Are there any other big projects that migrated and are happy with the experience? I never used Berry before

I can't speak for other OSS as I am fairly new to contributing but for the repo I implemented it came with some nice benefits. The biggest of course would by yarn PnP mode but that's significantly more complex. I found the investment was better spent eventually moving to pnpm. But this is a nice intermediate step since yarn classic gets no additional support currently fro. yarn time.

Since I have you here I think my blocker is the patch-package versions of several deps. I am not exactly sure what the process is expected here around that. It appears in one case the dependency may not even be used anymore. Is there documentation on how this should be handled? Otherwise I can try to stake a stab at manually updating the latest patch version of the packages then running patch package again.

Thank you for your time.

gaearon:
yeah tbh i don't think we can prioritize big infra changes like this at the moment, and i think we might as well adopt pnpm if we're making a change like this

PaulACoroneos:
> yeah tbh i don't think we can prioritize big infra changes like this at the moment, and i think we might as well adopt pnpm if we're making a change like this

okay that's reasonable. ill go ahead and close as won't do against considering pnpm. I don't think its easy for you to forecast when such an update will have bandwidth from team so may I can ask sometime later in 2025. take care.

Issue #6369, opened 2024-11-15
Accessibility bug: nearly impossible to login or sign up using a keyboard from the homepage


SleeplessByte:
### Steps to Reproduce

1. Go to https://bsky.app/
2. Try to reach the Sign in or Sign up buttons in the bottom bar
3. Good luck, you won't.

I don't think it's the expectation to tab over 400 times to get to these buttons. Additionally, content will load, so you potentially never reach it. The only way to currently sign up or login is to either know the URL or to _google_ (kagi search) the page.

- [ ] Allow users who are not signed in to sign up using a keyboard only without having to tab up to infinity

How to solve? You probably want a skip link, [understanding bypass blocks](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html). Usually these links are [added to skip the navigation](https://webaim.org/techniques/skipnav/), which bluesky lacks as well, but that's a different issue. When logged out, I would add a skip link at the top to "skip to navigation", which is what that is (sign up and login)

### Attachments

| Breakpoint | Focus order |
|--------|-----------------|
| Mobile | ![Focus order overlay on the logged out home page with a sign up reporting as 446, mobile viewport](https://github.com/user-attachments/assets/c439b03d-6f6b-4474-97cd-61e95f9caab5) |
| Desktop | ![Focus order overlay on the logged out home page with a sign up reporting as 416, desktop viewport](https://github.com/user-attachments/assets/e4d04981-9d9c-4ba2-94f2-54f3904e9322)

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### Device Info

Not Relevant

### What version of the app are you using?

Web

### Additional Information

If I recall correctly, Voiceover will not show a number for *each tabstop*, but only for each *visible tabstop*, so this _should not_ be an issue on Android/iOS in terms of using Voice Commands, but it is absolutely an issue for users who use something like a joystick or keyboard to tab through the interactive elements.

Issue #6370, opened 2024-11-15
Accessibility bug: Cannot reach the menu on the left or right unless I remove all feeds


SleeplessByte:
### Steps to Reproduce

1. Go to https://bsky.app/
2. Try to tab your way into the menu(s)
3. Good luck getting to the end of the feed

There are several issues here: 

- The focus order does not match the visual order.
- There is no way to quickly go past a post (`jk` navigation for example).
- There is no way to skip sections of the content (no skip links).

Solving the focus order solves part of it as the right sidebar on bluesky, just like on Twitter and LinkedIn is mostly useless.

- [ ] Ensure the [focus order matches visual order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order)
- [ ] Ensure the user can _skip_ the navigation
- [ ] Ensure the user can _skip_ the main content (to get to the sidebar on the right)

This is related to #6369 and the technique for skipping is the same.

You probably want several skip links, [understanding bypass blocks](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html). Usually these links are [added to skip the navigation to main content](https://webaim.org/techniques/skipnav/), which is probably what you want here as well, and you can add a second one to skip the feed to get to the right sidebar.

> [!CAUTION]
> There is no consensus on how to deal with a bottom navigation bar, but the people I test my software with are all okay with that taking focus precedence over the rest of the page, even though visually it comes "last". They say things like "yes, but a lot of apps have such a navigation so it's okay". My advice here is not leading and there is no WCAGuideline I can link to help here. You are just going to have to make a choice and defend it or be willing to adapt based on real-user-feedback.

The bottom navigation bar currently sits at the end of the focus order. I think this should move to the front, and in fact, that's how I always implement it on the products I work on. You don't want to have to scroll through the feed to get to the navigation.

### Attachments

| Breakpoint | Focus Order |
|------------|-------------|
| Desktop | ![Focus order on a logged-in home page showing to reach the navigation you need to tab over 600 times, desktop viewport](https://github.com/user-attachments/assets/16b62312-eab5-42e3-aa60-e2038386c016)
| Small laptop / Tablet | ![Focus order on a logged-in home page showing to reach the navigation you need to tab over 600 times, tablet viewport](https://github.com/user-attachments/assets/85c6ebbb-5cbf-4fe9-a8e9-b2aea1dd9889)
[...]

PR #6492, opened 2024-11-18
Fix Tab Order on Homepage

RichardNesbitt:
In response to -
Accessibility bug: nearly impossible to login or sign up using a keyboard from the homepage #6369

Reordered elements in createNativeStackNavigatorWithAuth.tsx and added z-index: 1 to the LeftNav and RightNav elements.


RichardNesbitt:
This will also address 
Accessibility bug: Cannot reach the menu on the left or right unless I remove all feeds #6370

src/view/shell/createNativeStackNavigatorWithAuth.tsx
```diff
@@ -138,6 +138,12 @@ function NativeStackNavigator({
 
   return (
     <NavigationContent>
+      {isWeb && !showBottomBar && (
+        <>
+          <DesktopLeftNav />
+          <DesktopRightNav routeName={activeRoute.name} />
```
SleeplessByte, line 144: Having the right sidebar before the main content would break the Focus Order guidelines though.

The left nav should definitely be first.

RichardNesbitt: Here's my justification for having it go LeftNav > RightNav > Feed.

[WCAG 2.2](https://www.w3.org/TR/WCAG22/)
**[2.1.2: No Keyboard Trap](https://www.w3.org/TR/WCAG22/#no-keyboard-trap)**
If the feed isn't last in the taborder, then the feed becomes a keyboard trap. With 'infinite-scroll' it's impossible to reliably reach the end of it.

**2.4.3: Focus Order**
[From the supplement](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html): 
**Goal**
Keyboard users navigate content in a correct order.
**What to do**
Elements receive focus in an order that preserves meaning.
**Why it's important**
Navigating a site or application with only the keyboard will make sense.

This last point is especially important. "Navigating a site or application with only the keyboard will make sense."

Right now: it doesn't make sense, since it can't be done.

If we do LeftNav > Feed > RightNav, it would make sense (logically) because of our propensity to read from left-to-right, however it's not _possible_ with the feeds configuration set to load infinitely.

I argue that with proper focus indicators, keyboard navigation in LeftNav > RightNav > Feed order would make the MOST sense given the purpose and nature of the feed block. 

Since we cannot disable this feature (the infinite-scroll) with compromising the UX (of everyone, not just sighted users) then LeftNav > RightNav > Feed is arguably the best possible solution.

SleeplessByte: Leave a comment to revert it when skiplinks are added? 

I personally agree with everything you've said.

RichardNesbitt: I'll do one better. I'll go add skiplinks into this PR.

RichardNesbitt: But of course now I'm thinking that "main content" is going to be the feed, which still renders that RightNav inaccessible. Thoughts? Ideas?

Honestly, why's it even there? 

We could just eliminate it like this...

Screenshot of BSky's homepage, showing that most of the content in RightNave is redundant, and the rest can easily be placed in other areas for better functionality.

PR #6494, opened 2024-11-18
DRAFT: migrate bsky.social to eslint v9 + flat config

PaulACoroneos:
I have started work to migrate the main eslint config for bsky.social to flat config + ESlint v9. I am probably going to need some help from core team because:

image

(but don't panic, a lot of errors are the same one over, and over again. We had options of course we can discuss in this PR).

OPEN TODO's

- [ ]  I have commented out bsky eslint plugin for now because TBH I have very little experience porting over a custom eslint plugin from extends to flat style. I can start doing reading but if someone more familiar in community wants to pitch in that would be awesome
- [ ] typescript-eslint is temporarily set to canary for now due to an ongoing bug with latest release. Once the release happens (later today I believe) I will update this to 8.15
- [x] I don't know what we want to do with the .eslintrc in bskyembed yet. It seems kind of out of date (and broken actually). It also doesn't really mirror the main one **complete**

Working on this migration has similarly been hard for my day to day (actually doing an even bigger migration now) so apologies if this is not a complete start. But I am hoping we can use it as a starting point. Please let me know what steps I can take to help move this forward.

Examples of broken stuff I think we can just fix (I am fine if we want to consider disabling this rule(s) if we don't want them, disabling for now, then re-enabling in follow up MR, etc):

image
image


PaulACoroneos:
Okay latest update. I have "disabled" the following rules with an intent to "quick follow" most of them:

```json
      //These should all be quick followed after eslint v9 upgrade
      'no-async-promise-executor': 'off',
      'no-case-declarations': 'off',
      'no-control-regex': 'off',
      'no-dupe-else-if': 'off',
      'no-empty': 'off',
      'no-empty-pattern': 'off',
      'no-irregular-whitespace': 'off',
      'no-prototype-builtins': 'off',
      'no-sparse-arrays': 'off',
      'no-unsafe-optional-chaining': 'off',
      'no-useless-escape': 'off',
      'no-var': 'off',
      'prefer-const': 'off',
      'prefer-rest-params': 'off',
      '@typescript-eslint/ban-ts-comment': "off",
      "@typescript-eslint/no-empty-object-type": ["error", { "allowObjectTypes": "always" }],
      '@typescript-eslint/no-explicit-any': 'warn',
      '@typescript-eslint/no-non-null-asserted-optional-chain': 'off',
      '@typescript-eslint/no-unsafe-function-type': 'off',
      '@typescript-eslint/no-unused-expressions':'off',
      "@typescript-eslint/no-require-imports": "off",
      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_', varsIgnorePattern: '^_'  }],
      '@typescript-eslint/no-wrapper-object-types': 'off'
      ```

I did go ahead and fix/enforce @typescript-eslint/no-unused variables and that is now present in the MR. I have found at least one logic bug so far using the above rules (not fixed for this MR, ideally id fix as part of turning on that rule, but I can split if we deem it higher priority:

image

https://github.com/bluesky-social/social-app/blob/0b38c77faa71dbc777d3d0162b86cb3e8695ff99/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx#L189

File: eslint.config.mjs [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+  recommendedConfig: eslint.configs.recommended,
+});
+
+export default tseslint.config(
+  eslint.configs.recommended,
+  ...tseslint.configs.recommended,
+  reactPlugin.configs.flat.recommended,
+  reactPlugin.configs.flat['jsx-runtime'],
```
haoqunjiang, line 25: IMO, this was not extended in the legacy config, so it shouldn't be extended in the flat config for now, at least not in this PR.

PaulACoroneos:
I am going to close this MR as simplifying it might as well start from scratch

eslint.config.mjs
haoqunjiang: I mean the `jsx-runtime` one… it's causing `import React` being removed everywhere. Removing it could significantly scale down the diff of this PR.

PaulACoroneos: Totally understand. Yeah I was thinking of doing JSX first since I can technically do it with existing config. It's a fairly low risk change anyways.

PaulACoroneos: @haoqunjiang I have broken the JSX transform + eslint config into #6516 6516

PR #6496, opened 2024-11-18
Help Siri to correctly pronounce Blue Sky.

m4p:
Help Siri to correctly pronounce Blue Sky. Instead of pronouncing it like “Lebowski”, i.e. “bluski”.


gaearon (approved):
sounds good to me, thank you

PR #6498, opened 2024-11-18
Extract RepostButton inner dialog

gaearon:
The usual refactor to avoid creating JSX eagerly.

## Test Plan

Repost button and dialog still works on iOS, Android.


mozzius (approved):
lgtm, pushed an additional simplification (we can use useDialogContext instead of passing control as prop)

wojtekmaj:
Curious, how does this prevent eager JSX creation? Since the new, extracted function does not contain e.g. an early return, I would expect identical JSX to be created before and after. 🤔

PR #6500, opened 2024-11-18
Add useHandleRef as a lighter alternative for useAnimatedRef

gaearon:
We use `useAnimatedRef` in a bunch of places just to get the measurements with `measure`. This is unfortunate because it contains [two shared values](https://github.com/software-mansion/react-native-reanimated/blob/36c291a15880c78a94dd125c51484630546ceb7c/packages/react-native-reanimated/src/hook/useAnimatedRef.ts#L46-L47), one of which stores an internal React node. I'm not sure whether that node goes through the serialization mechanism but the profiler was showing some overhead associated with each `useAnimatedRef` component.

In this PR, I'm taking just the part that we actually need — reading the component's internal handle. I'm doing the same thing [`findNodeHandle` does](https://github.com/facebook/react/blob/4beb1fd8bacacd3f4faeb180bca00ea0edcf27e4/packages/react-native-renderer/src/ReactNativePublicCompat.js#L118-L135) except for the [slow codepath we don't need (which only triggers for class components)](https://github.com/facebook/react/blob/4beb1fd8bacacd3f4faeb180bca00ea0edcf27e4/packages/react-native-renderer/src/ReactNativePublicCompat.js#L137-L145). I put that code inline so it's clearer how it works.

On the `measure` side, I realized that all Reanimated's `measure()` really wants is a function that returns the handle. (The `useAnimatedRef` API plays a trick here where it doubles as both a React callback ref which [receives the component](https://github.com/software-mansion/react-native-reanimated/blob/36c291a15880c78a94dd125c51484630546ceb7c/packages/react-native-reanimated/src/hook/useAnimatedRef.ts#L52-L56) and as a [function](https://github.com/software-mansion/react-native-reanimated/blob/36c291a15880c78a94dd125c51484630546ceb7c/packages/react-native-reanimated/src/hook/useAnimatedRef.ts#L82) that `measure` can [call](https://github.com/software-mansion/react-native-reanimated/blob/b2b58a68db2220b0f787448be5c7cc8eb099f3a8/packages/react-native-reanimated/src/platformFunctions/measure.ts#L40) to get the value.) So I'm just passing `() => handle` for that case.

I've also replaced `<Animated.View>` with plain `<View>` to reduce overhead since there's no need for Reanimated there at all.

## Test Plan

Verify you're still able to expand/collapse images:

- Single image (`AutoSizedImage`)
- Gallery
- User avatar
- Feed avatar

https://github.com/user-attachments/assets/b44d8e7e-f106-4ec5-bfdd-4b3076339707

https://github.com/user-attachments/assets/8afce036-72a0-47fa-bcab-820f8bdc54d2

## Before

`ImageLayoutGridInner` and `AutoSizedImage` were sticking out as having a bunch of self-time in traces.

Screenshot 2024-11-18 at 16 05 22
Screenshot 2024-11-18 at 16 09 41
Screenshot 2024-11-18 at 16 12 20

## After

`ImageLayoutGridInner` and `AutoSizedImage` no longer take disproportionate time in traces.

Screenshot 2024-11-18 at 16 17 52
Screenshot 2024-11-18 at 16 20 31
Screenshot 2024-11-18 at 16 24 45

### DIFF
--- a/src/lib/hooks/useHandleRef.ts
+++ b/src/lib/hooks/useHandleRef.ts
@@ -1,0 +1,39 @@
+import {useState} from 'react'
+import {AnimatedRef, measure, MeasuredDimensions} from 'react-native-reanimated'
+
+export type HandleRef = {
+  (node: any): void
+  current: null | number
+}
+
+// This is a lighterweight alternative to `useAnimatedRef()` for imperative UI thread actions.
+// Render it like <View ref={ref} />, then pass `ref.current` to `measureHandle()` and such.
+export function useHandleRef(): HandleRef {
+  return useState(() => {
+    const ref = (node: any) => {
+      if (node) {
+        ref.current =
+          node._nativeTag ??
+          node.__nativeTag ??
+          node.canonical?.nativeTag ??
+          null
+      } else {
+        ref.current = null
+      }
+    }
+    ref.current = null
+    return ref
+  })[0] as HandleRef
+}
+
+// When using this version, you need to read ref.current on the JS thread, and pass it to UI.
+export function measureHandle(
+  current: number | null,
+): MeasuredDimensions | null {
+  'worklet'
+  if (current !== null) {
+    return measure((() => current) as AnimatedRef<any>)
+  } else {
+    return null
+  }
+}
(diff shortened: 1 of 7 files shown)


mozzius (approved):
Tested on iOS+Android, lgtm

PR #6507, opened 2024-11-18
Split FeedSlice into FlatList rows

gaearon:
It was not very good that our `FeedSlice` is treated as a single FlatList row since it may contain multiple posts. Which, in the worst case, could also contain quotes and galleries, which would all have to be rendered at once.

This splits `FeedSlice` at the `FlatList` level so that its posts (and the "view full thread") are independent rows. There should be no observable visual changes — I've just moved the logic from `FeedSlice` into `renderItem`.

## Test Plan

Read through the `FeedSlice` code, see it mirrored in the `feedItems` loop and `renderItem` implementation.

I suggest to open localhost and prod side by side and scroll through both, verifying that different structures appear the same. Make sure to check a single post, a post with a reply, a three-item post, and a long (view full...) thread.

Verify "View full thread" goes to the right place (first post).

Try a few feeds: Following, Only Replies, and own profile "Replies". In own Replies, find some posts which starts with an omitted reply (the first post has "In reply to"), verify those appear the same.

Verify `getFeedPostSlice` gets called and successfully determines slices for Discover seen feed feedback.

## Before

A thread always had to render in a single chunk:

Screenshot 2024-11-18 at 21 13 46

## After

A thread can now be rendered in chunks:

Screenshot 2024-11-18 at 21 11 42


mozzius (approved):
looks good, I can't see any changes in behaviour

PR #6509, opened 2024-11-18
Use outline style for repost cancel button

mozzius:
I think it looks nicer, feel free to disagree

Screenshot 2024-11-18 at 21 42 54


gaearon (approved):
makes sense to me

PR #6510, opened 2024-11-19
Add Profiler build for Android

gaearon:
This adds `yarn android:profile` which produces a prod-like build that logs each React render phase. To read the log output, you'll need to connect with Android Studio and open logcat.

Example output:

Screenshot 2024-11-19 at 00 38 27

## Test Plan

Verify `yarn android:prod` does not produce these logs.

Verify `yarn android:profile` does.

<s>Verify that these two modes are independent and nothing's cached between them.</s> I'm actually having trouble with caching here, but assuming we always build real production builds from CI, this should be fine. I've asked Expo folks to look at this. It seems like editing the `metro.config.js` file is enough to bust the cache manually.

Does not affect iOS or web (on iOS, we don't set `BSKY_PROFILE`, and for web we don't use Metro).

### DIFF
--- a/src/AppProfiler.tsx
+++ b/src/AppProfiler.tsx
@@ -1,0 +1,29 @@
+import React, {Profiler} from 'react'
+
+// Don't let it get stripped out in profiling builds (which apply production Babel preset).
+const log = (global as any)['con' + 'sole'].log
+
+function onRender(id: string, phase: string, actualDuration: number) {
+  if (!__DEV__) {
+    // This block of code will exist in the production build of the app.
+    // However, only profiling builds of React call `onRender` so it's dead code in actual production.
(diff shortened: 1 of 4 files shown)


src/AppProfiler.tsx
```diff
@@ -0,0 +1,29 @@
+import React, {Profiler} from 'react'
+
+// Don't let it get stripped out in profiling builds (which apply production Babel preset).
+const log = (global as any)['con' + 'sole'].log
+
+function onRender(id: string, phase: string, actualDuration: number) {
+  if (!__DEV__) {
+    // This block of code will exist in the production build of the app.
+    // However, only profiling builds of React call `onRender` so it's dead code in actual production.
```
gaearon, line 9: !

haileyok (approved):

PR #6516, opened 2024-11-19
Leverage jsx transform to remove unnecessary react imports and update eslint config

PaulACoroneos:
Per feedback in #6494  we can break up migrating to ESLint v9 into smaller pieces. One thing I noticed is that bsky.social [has not yet implemented the JSX transform introduced by the React team.](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). It is no longer necessary to import React in JSX components if React itself is not being referenced.

The babel transform is built and in default as part of [babel-preset-expo and is turned to automatic by default](https://www.npmjs.com/package/babel-preset-expo).

In addition per react.dev blog, [the JSX transform is mandatory to move to React 19](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#new-jsx-transform-is-now-required).

What I did was go ahead and [extend the eslint config to include the jsx-runtime configuration](https://www.npmjs.com/package/eslint-plugin-react/) and I have also disabled prop-types since we use Typescript. I ran ```yarn web``` to validate the application runs after this change. Please let me know if any further details are required.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
   extends: [
     '@react-native-community',
     'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
     'plugin:react-native-a11y/ios',
     'prettier',
   ],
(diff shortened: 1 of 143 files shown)


PaulACoroneos:
@ haoqunjiang I went ahead and took your advice to break out the jsx transform piece. Thank you for your feedback

PR #6523, opened 2024-11-19
Update Korean localization

quiple:
Translates the new strings and unify the strings used for toasts into sentences.


pfrazee (approved):
Thank you as always quiple!

PR #6533, opened 2024-11-19
ListAddRemoveUsers: add profile links, hovercards, and new search

leonlarsson:
I was annoyed that I could not peek at profiles in the user list modal, so this PR aims to fix that

- Adds a ProfileHoverCard and Link around the avatar
- Adds the `<FollowsYou />` pill below the handle
- Replaces `<TextInput />` with the standard `<SearchInput />`
- Only shows "No results found for {query}" if there is a query (previously it showed immediately)

Preview:


leonlarsson:
Would love to get some feedback on this as more and more people are joining and creating lists for the first time :)

PR #6537, opened 2024-11-19
Dismiss keyboard on drag in starter pack screens

airowe:
Dismisses the keyboard from List component embedded within `StepProfiles` and `StepFeeds` with the `on-drag` mode provided by the native `ScrollView` component.

Prior to this PR, users would have to click a user to add them and then unclick, or hit the back button and come back to the screen in order to dismiss the keyboard as other controls are hidden behind the keyboard in a `BottomSheetDialog`.

- Follows pattern established in the `Explore` screen to dismiss keyboard on drag from the List component. -


gaearon:
I'm not able to reproduce this. Could you record a video of the problematic behavior?

gaearon:
Looks good. Doesn't seem like this does anything on Android, any idea why?

PR #6543, opened 2024-11-20
Task/remove prettier prettier eslint rule

PaulACoroneos:
Remove prettier/prettier override (to off) in .eslintrc.js in preparation of Eslint v9 upgrade. This was done by Dan because the previous config relied on usage of [eslint-config-prettier which is no longer recommended](https://typescript-eslint.io/users/what-about-formatting/). Dan's fix was [merged back in April and subsequently has been released](https://github.com/facebook/react-native/pull/43756).

Please let me know if there are any questions!

Edit: I reverted the package change after considering. Will keep this focused to removing prettier rule

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,8 +17,6 @@ module.exports = {
     'eslint-plugin-react-compiler',
   ],
   rules: {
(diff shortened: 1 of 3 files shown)


PaulACoroneos:
accidentally opened this too soon.

PR #6525, opened 2024-11-19
Bump typescript-eslint to 7.x

trivikr:
Fixes: https://github.com/bluesky-social/social-app/issues/6524

#### Before

Warning is shown about lack about support limited to `>=3.3.1 <5.2.0`.
```console
$ yarn lint
...
✖ 154 problems (0 errors, 154 warnings)
 
$ yarn list @typescript-eslint/parser
...
└─ @typescript-eslint/parser@5.62.0

```

#### After

No warning shown about the TypeScript version support.
```console
$ yarn lint
...
✖ 154 problems (0 errors, 154 warnings)

$ yarn list @typescript-eslint/parser
...
└─ @typescript-eslint/parser@7.18.0
```

PR #6544, opened 2024-11-20
Remove unnecessary lint rule suppression

PaulACoroneos:
Edit:

See updated comments from https://github.com/bluesky-social/social-app/pull/6544#issuecomment-2487273639. Thank you!

--- Original comments ---

Hello.

As part of the path to eslint v9 we need to [remove eslint-config-react-native-community which has not been updated in 2 years](https://www.npmjs.com/package/@react-native-community/eslint-config). This has [been replaced with react-native/eslint-plugin](https://www.npmjs.com/package/@react-native/eslint-plugin). As part of this migration eslint-plugin-react-hooks is no longer a dependency so I added it as well. (I also have added version detect so the react plugins know what version they are dealing with)

While doing this 2 rules-of-hooks violations were flagged. Both are valid as a hooks should not be called in an asynchronous function. That being said I don't really see a reason for the function to be marked asynchronous in first place. So I have remove the async from each declaration. This resolves the violation.

Finally we are able to formally remove references to prettier in the .eslintrc as the new plugins do not depend on eslint-plugin-prettier. Format concerns are now properly handled by prettier.

I ran yarn lint and validated we only get warnings back. Please let me know if you have any questions or need further details. Thank you!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,14 +1,14 @@
 module.exports = {
   root: true,
   extends: [
+    '@react-native',
     'plugin:react/recommended',
-    'plugin:react-hooks/recommended',
     'plugin:react/jsx-runtime',
     'plugin:react-native-a11y/ios',
+    'prettier',
   ],
   parser: '@typescript-eslint/parser',
   plugins: [
-    '@react-native',
     '@typescript-eslint',
     'react',
     'lingui',
@@ -112,9 +112,6 @@ module.exports = {
   ],
   settings: {
     componentWrapperFunctions: ['observer'],
-    react: {
-      version: 'detect',
-    },
   },
   parserOptions: {
     sourceType: 'module',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,14 +1,14 @@
 module.exports = {
   root: true,
   extends: [
-    '@react-native',
     'plugin:react/recommended',
+    'plugin:react-hooks/recommended',
     'plugin:react/jsx-runtime',
     'plugin:react-native-a11y/ios',
-    'prettier',
   ],
   parser: '@typescript-eslint/parser',
   plugins: [
+    '@react-native',
     '@typescript-eslint',
     'react',
     'lingui',
@@ -17,8 +17,6 @@ module.exports = {
     'eslint-plugin-react-compiler',
   ],
   rules: {
-    // Temporary until https://github.com/facebook/react-native/pull/43756 gets into a release.
-    'prettier/prettier': 0,
     'react/no-unescaped-entities': 0,
     'react/prop-types': 0,
     'react-native/no-inline-styles': 0,
@@ -114,6 +112,9 @@ module.exports = {
   ],
   settings: {
     componentWrapperFunctions: ['observer'],
+    react: {
+      version: 'detect',
+    },
   },
   parserOptions: {
     sourceType: 'module',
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -1,14 +1,14 @@
 module.exports = {
   root: true,
   extends: [
-    '@react-native-community',
     'plugin:react/recommended',
```
gaearon, line 4: do we want `@react-native` here too? for the config.

gaearon: nvm got it via https://github.com/bluesky-social/social-app/pull/6525

gaearon (approved):
let's try that

gaearon:
needs a rebase pls

.eslintrc.js
PaulACoroneos: Haha beat me to it no worries :)

PaulACoroneos:
@gaearon  rebase complete

.eslintrc.js
```diff
@@ -1,14 +1,14 @@
 module.exports = {
   root: true,
   extends: [
-    '@react-native',
```
trivikr, line 4: Is this removal intentional, or happened by mistake during rebase? This is the new ESLint config.

.eslintrc.js
```diff
@@ -1,14 +1,14 @@
...
     'plugin:react/recommended',
+    'plugin:react-hooks/recommended',
     'plugin:react/jsx-runtime',
     'plugin:react-native-a11y/ios',
-    'prettier',
```
trivikr, line 8: Please remove `eslint-config-prettier` dependency too from package.json

It's usually better to not mix multiple changes in one PR, but it's up to the maintainers to comment on it. The removal of `eslint-config-prettier` can be done independently of using `react-hooks`

PaulACoroneos: Yes it is unnecessary. I am sorting through some stuff around prettier right now so I have not had a chance to update. The inclusion was because moving to plugin the rule wasn't being extended anymore. Thank you and kindly give me a bit to sort things out.

PaulACoroneos:
@trivikr @gaearon Okay everything is fixed up (and I've now updated title after rebase) . Let me summarize:

I did not rebase tonight before starting work so I didn't see @trivikr beat me to the eslint package update. Thank you for doing that :).

When we made the update as @gaearon wanted we now relied on eslint-config-prettier which by configuration disables style rules such as semi, curly, etc. This is because we prefer to rely on the prettier config file instead. So now there is no need to disable prettier/prettier. I have kept the rules of hooks fixes because they do fail on my local and I think it's the correct thing per my understanding. I have run "yarn lint" locally and it only returns warnings.

My mistake was I accidentally read the docs for @react-natve/eslint-plugin instead of eslint-config. That is my fault and I sincerely apologize. Let me know if there are any further questions.

trivikr (commented):
LGTM

src/locale/i18n.web.ts
```diff
@@ -110,12 +110,9 @@ export async function dynamicActivate(locale: AppLanguage) {
   i18n.activate(locale)
 }
 
-export async function useLocaleLanguage() {
+export function useLocaleLanguage() {
   const {appLanguage} = useLanguagePrefs()
   useEffect(() => {
-    const sanitizedLanguage = sanitizeAppLanguageSetting(appLanguage)
-
-    document.documentElement.lang = sanitizedLanguage
```
gaearon, line 118: i think we still need this, no?

PaulACoroneos: shame on me for copy pasting haha. Yes let me fix thus

Issue #6471, opened 2024-11-17
Use native equivalents of lodash


trivikr:
### Describe the Feature

As per the source code, ten different utilities of lodash are being used
https://github.com/bluesky-social/social-app/blob/0b38c77faa71dbc777d3d0162b86cb3e8695ff99/package.json#L152-L161

Up to five of them can be replaced by native code
* `chunk`: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_chunk
* `debounce`: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_debounce
* `omit`: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_omit
* `random`: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_random
* `throttle`: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_throttle

These replacements can be used to reduce the number of lodash dependencies.

trivikr:
I noticed that 5 of the lodash dependencies weren't used.
Posted PR to remove them in https://github.com/bluesky-social/social-app/pull/6472

PaulACoroneos:
Left a comment on your PR but I think moving to lodash-es may be nice move as well

trivikr:
Closing as not planned, as example PR was closed https://github.com/bluesky-social/social-app/pull/6545

PR #6545, opened 2024-11-20
Use native equivalent of lodash.debounce

trivikr:
Refs: https://github.com/bluesky-social/social-app/issues/6471

Docs: https://you-dont-need.github.io/You-Dont-Need-Lodash-Underscore/#/?id=_debounce


gaearon:
While I know how to write a debounce, is there any benefit to this? The Lodash one works fine.

gaearon:
I think it's fine not to do this.

PR #6563, opened 2024-11-20
Fix change handle dialog on android

mozzius:
Caused by putting `a.overflow_hidden` on the ScrollView

### DIFF
--- a/src/components/Dialog/shared.tsx
+++ b/src/components/Dialog/shared.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import {StyleProp, TextStyle, View, ViewStyle} from 'react-native'
 
-import {atoms as a, useTheme, web} from '#/alf'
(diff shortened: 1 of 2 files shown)


src/components/Dialog/shared.tsx
```diff
@@ -29,10 +29,8 @@ export function Header({
         a.border_b,
         t.atoms.border_contrast_medium,
         t.atoms.bg,
-        web([
```
gaearon, line 32: what does this change? before/after?

gaearon (approved):
worked in my testing

src/components/Dialog/shared.tsx
mozzius: `a.overflow_hidden` was used to clip the corners of the header, but if we can't use that we need to use a border radius on the header itself instead

PR #6566, opened 2024-11-20
Updated Hindi UI for Version 1.94

FalKoopa:
I have translated the new text strings added in version 1.94 to Hindi. Also fixed some issues with my previous translation now that I was able to view the changes in-app.

(I also removed the old unused strings from the file, which is the reason for the large number in the red.)


trivikr (commented):
LGTM

pfrazee (approved):
Thanks everybody!!

PR #6577, opened 2024-11-20
Improve error name and message for signed-in only feeds

estrattonbailey:
When viewing a feed as a signed-out visitor to our web app, we check pages of posts to see if any posts _pass_ the default signed-out moderation settings. If none pass, we show an error screen in place of an entirely empty feed. We do this to prevent endless frontend fetching, trying to find a full page (30) posts that pass moderation and can be shown to the public.

Important to note: `no-unauthenticated` labels on users are also caught by this moderation check, meaning if a feed is full of folks who have opted out of public views, we will also show this error screen on the feed.

The previous error name and copy was a little misleading, and maybe concerning to some users. This PR aims to make it a little more clear what's happening. In the future we should revisit how we handle this "error" state.

CleanShot 2024-11-20 at 13 50 44@2x

### DIFF
--- a/src/view/com/posts/FeedErrorMessage.tsx
+++ b/src/view/com/posts/FeedErrorMessage.tsx
@@ -111,7 +111,7 @@ function FeedgenErrorMessage({
           msgLingui`Hmm, the feed server appears to be offline. Please let the feed owner know about this issue.`,
         ),
         [KnownError.FeedSignedInOnly]: _l(
-          msgLingui`This content is only available to signed-in users. Please sign in to view this feed.`,
+          msgLingui`This content is not viewable without a Bluesky account.`,
         ),
         [KnownError.FeedgenUnknown]: _l(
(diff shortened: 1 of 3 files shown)


src/view/com/posts/FeedErrorMessage.tsx
```diff
@@ -110,8 +110,8 @@ function FeedgenErrorMessage({
         [KnownError.FeedgenOffline]: _l(
           msgLingui`Hmm, the feed server appears to be offline. Please let the feed owner know about this issue.`,
         ),
-        [KnownError.FeedNSFPublic]: _l(
-          msgLingui`This content is not viewable without a Bluesky account.`,
+        [KnownError.FeedSignedInOnly]: _l(
+          msgLingui`This content is only available to signed-in users. Please sign in to view this feed.`,
```
pfrazee, line 114: Let's not change this message so that our translators dont have to redo it

PR #6583, opened 2024-11-21
[Android] Patch react-native-svg to cache parsed paths

gaearon:
Kind of a hackfix. See https://github.com/software-mansion/react-native-svg/pull/2548.

We don't want to parse these icons' paths every time we create a view with them — this is downright silly.

This caches them forever but it's ok because our SVGs are not dynamic and we have a limited number of them.

## Test Plan

Verify we're hitting the cache as expected. As you scroll feed, you have fewer cold parses and soon run out of them.

https://github.com/user-attachments/assets/b8b378d0-9898-441a-9406-5617887ca430

## Before

This function is a significant part of the `createViewInstance` trace.

Screenshot 2024-11-21 at 00 40 30

## After

It's a smaller part of the overall picture.

Screenshot 2024-11-21 at 01 14 22

## Proper fix

This doesn't fully solve the issue — this library is still showing up in `onDraw`, for example. I think ideally we'd move to https://github.com/zamplyy/react-native-vector-image-plugin or something like it, but let's land this for now.


haileyok (approved):
Nice. This makes sense, great find.

PR #6586, opened 2024-11-21
Disable vertical scrollbars on Android only

gaearon:
Revisiting https://github.com/bluesky-social/social-app/pull/3855 for two reasons:

- With infinite scroll, they don't really mean anything
- On Android specifically, they get pretty jumpy due to more choppy virtualization
- Them (re)appearing is showing up in traces — it's a tiny bit of extra work to render them
- On Android, they don't work as a navigational element anyway so the justification from https://github.com/bluesky-social/social-app/pull/3855 doesn't apply

So let's remove them but on Android only.

## Test Plan

Can still see them on iOS and web. Can't see them on Android.


estrattonbailey (approved):

mozzius (approved):
I'm fine with this

Not tested

PR #6587, opened 2024-11-21
Automatically switch app language based on device language.

auroursa:
Let’s introduce a package [react-native-localization-settings](https://www.npmjs.com/package/react-native-localization-settings) to make the app’s language selectable in the device settings（Update: It’s not required, but I think it could improve user experience).

Based on `expo-localization`, the app language can automatically follow system language. For example, if the device language is set to Japanese, the app will automatically switch to Japanese (instead of  English), similar to our web matches the browser language.

To compatibility with Android and iOS language codes, the Chinese language code needs to be modified to follow the IETF-BCP-47 format. I’ve tested it on Android versions 7 to 14, and all app language are working fine.

It should work well on iOS, but I haven't had the opportunity to test it.

https://github.com/user-attachments/assets/dadecbb2-ed16-4c00-af1e-b777f31ce08e


src/locale/i18n.ts
```diff
@@ -224,6 +225,25 @@ export async function dynamicActivate(locale: AppLanguage) {
   }
 }
 
+export function useSystemLanguage() {
+  const {appLanguage} = useLanguagePrefs()
+  useEffect(() => {
+    async function checkSystemLanguage() {
+      try {
+        const getSystemLanguage = await getLanguageAsync()
+        const systemLanguage = getSystemLanguage.replace('_', '-')
```
auroursa, line 234: I don't want to make too many changes, but the getLanguageAsync function returns a format like "zh_CN" on Android, which doesn't match the "zh-CN" format used in others code. So, I added this conversion, but it doesn't seem to work.

auroursa (commented):
I found an issue, the code works correctly for language codes like "ja" and "ko", but doesn't work when there's a regional variant, such as "zh-CN".

fobos531:
@auroursa I believe adding a new dependency is not needed, as the app already depends on `expo-localization`. You can use the `Localization.getLocales()` method to get the current language: https://docs.expo.dev/versions/latest/sdk/localization/#localizationgetlocales.

auroursa:
> @auroursa I believe adding a new dependency is not needed, as the app already depends on `expo-localization`. You can use the `Localization.getLocales()` method to get the current language: https://docs.expo.dev/versions/latest/sdk/localization/#localizationgetlocales.

Thank you! I've switched to using expo-localization, and it works well with the regional language codes (except for Chinese).

Maybe react-native-localization-settings still has value, as it allows users to switch languages through the system settings.


https://github.com/user-attachments/assets/e003b7c2-f8c7-4554-9d5a-1c352b5c7a78

auroursa:
Updated Chinese language code to follow IETF-BCP-47, and now all languages supported by the app are working correctly.

auroursa:
It will break the browser’s automatic language switching behavior. Closing this PR until ready.

PR #6588, opened 2024-11-21
Don't underline links on native hover

gaearon:
Underlining links on hover doesn't make sense on native because you can't see it under the hover anyway. iOS already applies a selected style anyway. And on Android we'd like to avoid extra work during touches.

Let's remove these handlers.

## Test Plan

Verify links still get underlined on web, but don't on native.

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -274,11 +274,6 @@ export function InlineLinkText({
     onOut: onHoverOut,
   } = useInteractionState()
   const {state: focused, onIn: onFocus, onOut: onBlur} = useInteractionState()
-  const {
-    state: pressed,
-    onIn: onPressIn,
-    onOut: onPressOut,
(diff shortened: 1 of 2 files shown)


src/components/Link.tsx
```diff
@@ -289,19 +284,20 @@ export function InlineLinkText({
       {...rest}
       style={[
         {color: t.palette.primary_500},
-        (hovered || focused || pressed) &&
+        (hovered || focused) &&
           !disableUnderline && {
-            ...web({outline: 0}),
-            textDecorationLine: 'underline',
-            textDecorationColor: flattenedStyle.color ?? t.palette.primary_500,
+            ...web({
```
mozzius, line 289: I guess the spread isn't necessary anymore

gaearon: i'll leave in for the next person who edits this

PR #6591, opened 2024-11-21
fix: reused getRkey function in embed component

MathurAditya724:
I've updated the `Embed` component to reuse the `getRkey` function instead of relying on `AtUri`. This change enhances reusability and helps reduce the bundle size slightly.

I came across this improvement while working on https://github.com/rhinobase/react-bluesky and thought it would be a helpful optimization.


mozzius (approved):
`getRkey` also uses `AtUri`, so it won't impact bundle size, but no harm in make it more consistent

PR #6094, opened 2024-11-03
Fix minor localization issues in Spanish:

elenatorro:
Minor things I've seen while using the app and then reviewing some of the texts:

- Use 'pack de inicio' instead of 'paquete de inicio'
- Use 'Inténtalo' instead of 'Intenta'
- Fix 'Por favor'
- Do not use 'una' as article for 'posts'
- Use 'Eliminado' instead of 'Removido'

PR #6599, opened 2024-11-21
Update Spanish localization

codehead:
Translate missing strings to hit full Spanish localization.
Going forward, I would take @elenatorro and @ecreeth 's very sensible suggestions in #6094 to the rest of the strings, but additional community input would be helpful to agree on the preservation (or not) of certain anglicisms (Is it a "post" or a "publicación"? In verbal form, "repost" or "republicar"? Once done, is it a "repost" or a "republicación"? ¿A un usuario "le gusta una publicación" o "le da «me gusta» a un post"? And so on with "Feed", "Discovery", etc.)


reoblauber:
Hey! I've also been working on the Spanish localization, but on my own, I'll publish it when I'm done.
I've been using some anglicisms like **Post**, **Repost**, **Feed**, etc because in my opinion trying to use native spanish words with those doesn't fit in well, but I did opt to not use anglicisms in a couple where they fit in, like **"Chat"** which in my project translated as **"Mensajes"**
Personally, I don't find the «» quotation marks appealing and instead use "". In the case of **"le da «me gusta» a un post"** I don't find it necessary to use quotation marks at all. In my localization I worded it as **"{user} le ha dado me gusta a tu post"**

elenatorro:
I agree that "Post," "Repost," and "Feed" are commonly used terms, as well as "Pack" versus "Paquete." I don't have a strong preference since both options are correct, but I think it would be helpful to know which terms people are more accustomed to using nowadays 👍🏽

codehead:
Hello @reoblauber . Of course, you're welcome to expand on this PR, it now has 100% coverage for the latest release.
All new strings are using "Publicación" but I do agree that "Post" feels more familiar, even in Spanish, and that "Chat" reads much better as "Mensajes".
@madkoding has done a massive effort before, some of their translations with `"me gusta"` use guillemets (`«»`) and some others don't; though it might seem trivial, we could start by standardizing these - in the new strings I went for regular double quotation marks because I feel quoting is necessary to describe someone else's action on one's post, but I'm not married to that idea. What I was wondering about was if it'd be more natural to say `a {user} le gusta tu post` rather than `{user} ha dado "me gusta" a tu post`. Do you have an opinion on that?

reoblauber:
In the PR I am working on I am mainly fixing some inconsistencies all around, just tidying up things, but I don't know how long it's gonna take so I might just make it its own PR. If I get it done soon enough I will consider just adding it to this one.

> some of their translations with "me gusta" use guillemets («») and some others don't

In my opinion, I don't think they are a good fit. I have of course seen them in the past but I don't think I've ever seen them used in the Spanish localization of any other site or app I've worked on, and personally they simply just don't feel natural to me to use them in this case.

> What I was wondering about was if it'd be more natural to say a {user} le gusta tu post rather than {user} ha dado "me gusta" a tu post. Do you have an opinion on that?

Both are correct, but I am more partial to the "le ha dado me gusta" (without the quotation marks around "me gusta", i personally don't find them necessary there) since it emphasizes more that the user pressed "Like" on the post, rather than the slightly ambiguous "le gusta tu post" which could mean the user simply thought your post was neat. This is of course just my opinion though and I wouldn't mind whichever you decide to choose.

pfrazee (approved):
Thanks!! Going to commit

codehead:
Absolutely right. It's missing @hogaza 's changes from #6691, I'll pull those. Thank you!
```
2740:msgstr ""
2839:msgstr ""
3767:msgstr ""
6685:msgstr ""
8222:msgstr ""
8477:msgstr ""
8481:msgstr ""
8896:msgstr ""
```

monty241:
I know that Bluesky wants to move to some CAT-tool, but for now the translations can be done in Localazy. Currently for sv, es, an and nl based upon en. I have learned that the Tools -> File management -> Export is only available to persons with role Owner. I will need to do the export; I am sorry, but just drop me an email or ping this issue. I will put the new version then in the branch i18n-es on https://github.com/monty241/social-app/tree/i18-es

monty241:
Yes, it is a little cumbersome. Hope that the alternative tool becomes available. In case it takes very long I can run a daily script to extract them using the Invantive SQL-driver for Localazy and commit & push them automatically, but a process integrated solution is better for bsky.

hogaza:
Thanks for adding my changes, @codehead ! I'm wondering how to proceed, as I wanted to update other strings but I'd like to do it in a coordinated way with the other contributors. @monty241 is it possible to join the Localazy team while we wait for an official Crowdin integration?

monty241:
Sure. Just enter request at https://localazy.com/p/bsky. Please put your GitHub name and language in the text.

monty241:
@hogaza I have added you to ES as translator and changed all first person to enter for a language to reviewer, in this case @codehead . I am currently the only one that can export it seems, but just drop me a note. It takes me just a few seconds and in case Crowdin takes very long I will change it into a SQL-script to dump daily to messages.po.

PR #6607, opened 2024-11-21
Pin exact version of React Compiler packages

poteto:
This PR pins the various React Compiler packages to their exact versions. While it is typically safe to upgrade the compiler, we recommend manual upgrades because changes in memoization may or may not be unexpected.

In the near future we will open source an upgrade script that automates comparing the current compiled output with the upgraded output and prints a diff for human confirmation prior to committing the upgrade.


gaearon (approved):
lg

PR #6609, opened 2024-11-21
Fix Fast Refresh in <Text> files

gaearon:
It was very annoying to iterate on the `Text` components because the app always reloaded. This was because those files export non-React functions, and so the edit was propagating up the module system and wasn't fully handled. The fix is to move non-React stuff to another file.

There are no logic changes.

## Test Plan

Save file in `Text.tsx` or `Typography.tsx`, the entire app no longer reloads.


estrattonbailey (approved):
Oh nice, good to know

PR #6610, opened 2024-11-21
Update bskyembed/yarn.lock to include only one Zod version

jviide:
This pull request updates the bskyembed/yarn.lock file so that only one version of Zod is included in the Bluesky embed bundle. This seems to reduce the resulting bundle size by about 9% (from 156kB to 142kB, gzipped).

While looking at the Bluesky embed code with @marvinhagemeister we spotted two versions of Zod in the bundle visualization treemap:

treemap

This could be solved by updating the embed bundle's lockfile. Previously both Zod 3.22.4 and 3.23.8 were included and the bundled code took around 156kB gzipped:

```
dist/static/post-BtQoyJ24.js   763.87 kB │ gzip: 156.07 kB
```

Updating the lockfile to only incldue Zod version 3.23.8 reduces the gzipped bundle size by about 9%:

```
dist/static/post-DOc5yEa0.js   700.56 kB │ gzip: 142.04 kB
```


gaearon (approved):
lgtm, thanks

PR #6611, opened 2024-11-21
Fix overflow issue on iOS autocomplete, among other things

mozzius:
Fixes a few issues with the autocomplete

- Fixes overflow issue on iOS (ripped out a dumb system based on number of characters and just figured out how to do the same behaviour with flexbox)
- Fixes missing display name issue on web (search for @zed.dev to see this)
- Remove use of reading refs in render in favour of the built-in flag on the hook
- Use unicode symbol for ellipsis instead of `...` (since this looks weird next to `numberOfLines`)

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Screenshot 2024-11-21 at 20 50 51</td>
      <td>Screenshot 2024-11-21 at 20 49 29</td>
    </tr>
  </tbody>
</table>

# Test plan

- Check that the overflow issue is indeed fixed (iOS) and check that I didn't break anything (Android)
- Check that it still maintains stale items as you type (this is what the ref did)
- Check that accounts where the displayname is >50% width and handles are short still take up as much space as they can (iOS + Android) - @atproto.com is a good example

### DIFF
--- a/src/view/com/composer/text-input/hooks/useGrapheme.tsx
+++ b/src/view/com/composer/text-input/hooks/useGrapheme.tsx
@@ -13,7 +13,7 @@ export const useGrapheme = () => {
 
         if (graphemes.length > length) {
           remainingCharacters = 0
-          name = `${graphemes.slice(0, length).join('')}...`
+          name = `${graphemes.slice(0, length).join('')}…`
         } else {
           remainingCharacters = length - graphemes.length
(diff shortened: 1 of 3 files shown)


gaearon (approved):
works well in my testing

PR #6615, opened 2024-11-21
Reduce <Text> nesting

gaearon:
This reduces the nesting we do for `Text` views on both platforms.

On iOS, we no longer nest multiple `UITextView`  inside each other in the emoji codepath. Only emojis themselves get wrapped. AFAIK the rest of the wrapping wasn't needed (or at least I can't tell why it is).

We're also changing `RichText` to pass through non-facet text fragments as plain strings instead of as `<Text>` subviews (this affects both iOS and Android). I suspect that the only reason they were `<Text>` subviews [was](https://github.com/bluesky-social/social-app/commit/5eb294488f08534abac3335acfa366cffea9259e#diff-74c35a105c369d2c177b8c4c3246c614ef34e52a449a12a707b9539087b0af07R153) to put the `emoji` flag on that `Text`. Instead, I've changed the `<Text emoji>` thing to support multiple children, as I don't see why it wouldn't work.

Also a few small driveby fixes.

Might be easier to review commit-by-commit to see the sequencing.

## Test Plan

Try a post like https://bsky.app/profile/shannonmattern.bsky.social/post/3ku7zasy6xz2e on all platforms, verify the emojis aren't "jumping".

Verify posts with a single emoji appear enlarged like before.

Verify posts with links work like before.


src/alf/typography.tsx
```diff
@@ -71,65 +72,55 @@ export type TextProps = Omit<RNTextProps, 'children'> & {
...
+  /**
+   * Whether the children could possibly contain emoji.
+   */
+  emoji?: boolean
+}
 
 const EMOJI = createEmojiRegex()
 
 export function childHasEmoji(children: React.ReactNode) {
-  return (Array.isArray(children) ? children : [children]).some(
```
gaearon, line 88: this is not the correct way to iterate over children because those could be in nested arrays

src/alf/typography.tsx
```diff
@@ -71,65 +72,55 @@ export type TextProps = Omit<RNTextProps, 'children'> & {
...
+  return hasEmoji
 }
 
 export function renderChildrenWithEmoji(
-  children: StringChild,
+  children: React.ReactNode,
   props: Omit<TextProps, 'children'> = {},
+  emoji: boolean,
 ) {
-  const normalized = Array.isArray(children) ? children : [children]
```
gaearon, line 107: same

src/components/RichText.tsx
```diff
@@ -150,17 +151,14 @@ export function RichText({
         />,
       )
     } else {
-      els.push(
-        <Text key={key} emoji style={plainStyles}>
-          {segment.text}
-        </Text>,
-      )
+      els.push(segment.text)
```
gaearon, line 154: this is the main reason for the entire change tbh

estrattonbailey (approved):

PR #6633, opened 2024-11-22
[Lightbox] Fix jump when zooming out on iOS

gaearon:
Currently, we disable bounce when the image is 1:1. This is to allow it being swiped away (instead of bouncing). However, this actually introduces a problem: if you start pinching in _first_, and _then_ pinch out, the image will jump as you cross the 1:1 threshold. That's because the `bounces` prop would change while dragging, and disabling bouncing would force-center the image. The fix is to only disable bouncing while both 1:1 _and_ at rest.

Implementation-wise, I wanted an excuse to try `useAnimatedProps`, so this is it. Seems like a convenient way to toggle native props without leaving the UI thread. An alternative would be to call `setNativeProps` imperatively.

## Test Plan

Verify you still can swipe away images when at rest.

Verify you still _cannot_ swipe away images while zoomed in.

Verify that if you pinch in and out (and in and out again), crossing the 1:1 threshold does not cause visual jumps. This is much easier to test on the device; if you do it on the Simulator, you'll have to _both pinch and pan_ to reproduce the jump.

## Before

https://github.com/user-attachments/assets/0c44e058-78f7-4066-b2b1-9f9a0f6e9a49

## After

https://github.com/user-attachments/assets/2cacba17-5979-4445-aa59-d2feb085ac60

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.ios.tsx
@@ -16,9 +16,11 @@ import {
 import Animated, {
   runOnJS,
   SharedValue,
+  useAnimatedProps,
   useAnimatedReaction,
   useAnimatedRef,
   useAnimatedStyle,
+  useSharedValue,
 } from 'react-native-reanimated'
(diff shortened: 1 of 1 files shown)


mozzius (approved):
yep, seems better

PR #6637, opened 2024-11-22
Lightbox status bar behaviour

mozzius:
- Use dark mode for status bar/nav bar
- (iOS) hide status bar when controls are hidden or when zoomed

https://github.com/user-attachments/assets/f247a7ec-9ada-464c-9b97-3c096146c08e

# test plan

Check it works on iOS and Android. Ensure the status bar/nav bar returns to previous state when closing the lightbox
Check light mode and dark mode


gaearon (approved):
lgtm

PR #6638, opened 2024-11-22
Add Aragonese language ('an') support and l10n

Juanpabl:
PR to add Aragonese language support and localization to Bluesky (code: an)


pfrazee:
Awesome! Is there a native speaker available to confirm this translation?

src/components/hooks/dates.ts
```diff
@@ -10,6 +10,7 @@
 import React from 'react'
 import {formatDistance, Locale} from 'date-fns'
 import {
+  an,
```
surfdude29, line 13: It looks like the `date-fns` package doesn't yet support the `an` locale, so I think this line should be removed.

https://github.com/date-fns/date-fns/tree/main/src/locale

src/components/hooks/dates.ts
```diff
@@ -41,6 +42,7 @@ import {useLanguagePrefs} from '#/state/preferences'
  */
 const locales: Record<AppLanguage, Locale | undefined> = {
   en: undefined,
+  an,
```
surfdude29, line 45:
```suggestion
  an: undefined,
```
As above, it looks like the `date-fns` package doesn't yet support the `an` locale, so I think this change is needed.

src/components/hooks/dates.ts
Juanpabl: Thanks! I commited your change

surfdude29: Nice! I think this line might need to be removed as well though; there is no entry for `ga` for instance, which is also not supported.

I didn't realise that you could suggest to remove a line like this, but apparently you can :)
```suggestion
```

src/locale/i18n.ts
```diff
@@ -41,6 +42,14 @@ import {useLanguagePrefs} from '#/state/preferences'
  */
 export async function dynamicActivate(locale: AppLanguage) {
   switch (locale) {
+    case AppLanguage.an: {
+      i18n.loadAndActivate({locale, messages: messagesAn})
+      await Promise.all([
+        import('@formatjs/intl-pluralrules/locale-data/an'),
+        import('@formatjs/intl-numberformat/locale-data/an'),
```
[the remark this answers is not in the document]
Juanpabl: I have replaced 'an' by 'es' as fallback language. That would be the most natural for most users.

Juanpabl:
> Awesome! Is there a native speaker available to confirm this translation?

Thanks @pfrazee. There are not many other Aragonese speakers with an account in Github, I'll try to have someone commenting here. 

In any case, I have quite some experience in localizing software to Aragonese (Firefox, Telegram, Mediawiki, etc.) and have followed common guidelines from Softaragonese in this project too. I am also member of the language official regulatory body (Academia Aragonesa de la Lengua) and I have therefore followed the official normative spelling. The translation should therefore be good enough for a first version. I'm sure there may be some issues identified upon deployment, and I commit to get feedback from the community to improve further versions.

src/locale/i18n.ts
```diff
@@ -44,6 +45,14 @@ import {useLanguagePrefs} from '#/state/preferences'
  */
 export async function dynamicActivate(locale: AppLanguage) {
   switch (locale) {
+    case AppLanguage.an: {
+      i18n.loadAndActivate({locale, messages: messagesAn})
+      await Promise.all([
+        import('@formatjs/intl-pluralrules/locale-data/es'),
```
surfdude29, line 51:
```suggestion
        import('@formatjs/intl-pluralrules/locale-data/an'),
```
That's great, although I would suggest that you can leave `intl-pluralrules` as `an` because the package supports it, it's only `intl-numberformat` that doesn't :)

https://www.npmjs.com/package/@formatjs/intl-pluralrules?activeTab=code

pfrazee (approved):
Okay! Thanks so much for the contribution!

PR #6641, opened 2024-11-22
Simplify play icon shadow

gaearon:
This solves a few things:

- The shadow was almost invisible on light theme because the `shadowColor` was theme-dependent (which doesn't make sense when overlaying a video).
- We were getting `Raised button causing error: View #... of type RCTView has a shadow set but cannot calculate shadow efficiently` which is only solveable by putting a solid background color on the element _with_ the shadow.

Should overall look the same-ish. Here it is against different backgrounds:

Screenshot 2024-11-22 at 19 18 43
Screenshot 2024-11-22 at 19 18 52
Screenshot 2024-11-22 at 19 19 38

### DIFF
--- a/src/components/video/PlayButtonIcon.tsx
+++ b/src/components/video/PlayButtonIcon.tsx
@@ -13,9 +13,12 @@ export function PlayButtonIcon({size = 32}: {size?: number}) {
       <View
         style={[
           a.rounded_full,
-          t.atoms.shadow_lg,
           {
             backgroundColor: bg,
+            shadowColor: 'black',
+            shadowRadius: 32,
+            shadowOpacity: 0.5,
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
hottt

mozzius (approved):
I don't exactly get the motivation but it works fine

gaearon:
>I don't exactly get the motivation

See the summary:

- It did not have the same color in dark and light theme, making it essentially invisible with the light theme
  - We do want it to always be visible so that the play button has a bit of contrast against light videos
- We were getting a warning from React Native whenever it rendered ("Raised button causing error: View #... of type RCTView has a shadow set but cannot calculate shadow efficiently") which is probably not good

PR #6617, opened 2024-11-21
Reduce overdraw

gaearon:
This probably isn't very important but I figured we might as well do it.

These don't need explicit backgrounds because there's a background above.

## Test Plan

Walk around the screens, everything looks the same visually. The home/profile tabbar still has a solid background. Same for the bottom bar.

## Before

Screenshot 2024-11-21 at 23 03 42

## After

Screenshot 2024-11-21 at 23 03 35

(The bottom bar isn't actually transparent, it's just how "Debug GPU Overdraw" visualizes it.)

PR #6642, opened 2024-11-22
Fix leaking background

gaearon:
Partial revert of https://github.com/bluesky-social/social-app/pull/6617. Noticed 1px of the white background is flashing during the drawer animation. Guess we need this after all.


src/view/shell/index.tsx
```diff
@@ -144,7 +146,7 @@ export const Shell: React.FC = function ShellImpl() {
     }
   }, [theme])
   return (
-    <View testID="mobileShellView" style={[styles.outerContainer]}>
+    <View testID="mobileShellView" style={[styles.outerContainer, pal.view]}>
```
estrattonbailey, line 149: FYI you can use `useTheme` from ALF and `t.atoms.bg` for this too 👍

estrattonbailey (approved):

src/view/shell/index.tsx
gaearon: ah well, next time :P

PR #6645, opened 2024-11-22
Fix interstitials to not interrupt feed slices

gaearon:
I regressed on this in https://github.com/bluesky-social/social-app/pull/6507. We never want to put an interstitial in between two items of the same slice. To fix this, I'm moving the logic to where we enumerate over the slices. While we're at it, I'm also ripping out a bit of abstraction that hasn't proven out to be useful and was making my mistake harder to spot. (Plus removing the experiment we didn't ship.)

## Test Plan

Reset onboarding state, observe the progress bar on Discover:

Screenshot 2024-11-22 at 20 13 05

Discover still shows suggestions:

Screenshot 2024-11-22 at 20 05 01

They change after Discover invalidation.

Profiles still show suggestions:

Screenshot 2024-11-22 at 20 12 57

### DIFF
--- a/src/lib/statsig/gates.ts
+++ b/src/lib/statsig/gates.ts
@@ -2,4 +2,3 @@ export type Gate =
   // Keep this alphabetic please.
   | 'debug_show_feedcontext' // DISABLED DUE TO EME
   | 'post_feed_lang_window' // DISABLED DUE TO EME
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
LGTM!

Issue #2540, opened 2024-01-16
Bsky app doesnt send referrer


jaoler:
You can't track traffic to your website from Bluesky app because of this. In Matomo or Google Analytics you only see visitors who use Bluesky in browser. 

Matomo dev team wrote:

_Our development confirmed the app does not send a referrer, so it is only possible to populate as direct entry unfortunately.

As BS is still in beta development, maybe they will add the information in the future.

Let me know if you have any questions please.

Regards,

--
Ralph Conti
support@matomo.cloud
Matomo Support Team_

jaoler:
Can you please fix this. It hard to show how important Bluesky is, if you cant track traffic to your site.

FairgroundTown:
This feels like it is massively in Bluesky's interests to fix.

fwojtan:
I agree. Although fixing this isn't free - to do it properly you'd need to host a proxy to add the header. There's got to be a strong commercial case for showing downstream sites that Bluesky is driving their traffic, but...
- it's not obvious to me that fixing this in the app is sufficient in the long run, if the network becomes properly federated people may click on a link you posted on the Bluesky app in a different client altogether, so we'd still be undercounting, or at least disaggregating, the impact of Bluesky on traffic
- whoever pays for the proxy would need to decide there's a business case and that this is a priority

spectre1989:
I think the business case for the proxy is that this will drive more users to the platform. People who monetise substacks and the like will be more active on bsky if they see more of their traffic coming from there

bthompson-365:
Sorry but this is not a bug - if you want to track visitors from Bluesky to your site put the appropriate information in the URL you post...

jaoler:
> Sorry but this is not a bug the lack of referrer information is a plus for an awful lot of people
> 
> And there is a simple workaround if you want to track visitors from Bluesky to your site put the appropriate information in the URL you post - in the same way you would when sending an email out..

No, you are wrong. Its totally differend thing put tracker urls to post you personally share.

FairgroundTown:
> Sorry but this is not a bug the lack of referrer information is a plus for an awful lot of people
>
> And there is a simple workaround if you want to track visitors from Bluesky to your site put the appropriate information in the URL you post - in the same way you would when sending an email out..

[...]

PR #6648, opened 2024-11-22
Pass referrer on native (with an opt out)

gaearon:
Attempts to fix https://github.com/bluesky-social/social-app/issues/2540.

See https://github.com/bluesky-social/social-app/issues/2540#issuecomment-2481304564 for tradeoffs of different solutions:

1. We don't want to degrade the user experience so replacing `SFSafariViewController` and `CustomTabsIntent` with web views seems out of question.
2. We'd rather not introduce more centralized points of failure, plus an extra redirect for every outbound link is a bit annoying, so we'd rather not introduce a service like `https://t.co`, at least for now.

This leaves us with passing referer information (just `utm_source=bluesky&utm_medium=social`) using UTM fields which tend to be hooked up to some kind of analytics system and have well-defined meanings. We will default to the passing being on _for native only_ (on web we don't need this because the browser passes the `Referer` field). There is an opt out. Also, we won't touch the links that already have the UTM fields to avoid messing with existing analytics campaigns. The records will remain clean; the rewriting will only be applied using the in-app browser or opening the system one.

Screenshot 2024-11-22 at 22 45 25

## Test Plan

Web: Verify no changes. The toggle doesn't exist, the links open without params. I've removed `noreferrer` on some links but I'm not sure it was actually getting taken into account anyway (I do see the domain `Referer` in production, and we do specify `<meta name="referrer" content="origin-when-cross-origin">` already which is the behavior we want).

Native: Verify there is a new settings toggle (on by default). Verify that opening a link in the in-app browser _or_ (if your setting is to use the system browser) will have the `?utm_source=bluesky&utm_medium=social` appended. Verify it doesn't break other search params in the URL. Verify it doesn't clobber existing `utm_source` or `utm_medium`. Verify that opt-out is respected both for in-app browser and for the system browser.


pfrazee (approved):

jaoler:
@gaearon question. Is this implemented? I dont see "send referrer" in settings. How this works?

gaearon:
We didn’t end up doing this. Felt too invasive to have the links messed with on share.

PR #6649, opened 2024-11-23
Add a hotness thread sort

gaearon:
Soft-depends on https://github.com/bluesky-social/atproto/pull/3082

---

This adds a new sorting mode ("hotness") and makes it a default for people who never updated their Thread Preferences (needs package bump). If you previously explicitly expressed a different preference, your old choice will be respected.

Currently, the default thread setting is "oldest first". This means that whoever comments first effectively owns the space above the fold, and there isn't much incentive to reply later — after a certain cutoff, unless you *already* have an existing follower base for whom your posts is prioritized, your reply will be seen by very few people.

The "hotness" ranking tries to fix this. It is not ideal (and we'll be tweaking it with time), but the goal is to give a chance to *recent* comments to get people's eyeballs, especially in larger threads — so that people can like your comment even if you haven't already built a significant following. (Most people who *already* follow you will see your comment at the top with either ranking because "prioritize follows" is true by default — so this doesn't change whether your followers see your posts.)

With the "hotness" ranking, recent comments (whether you have followers or not) will be seen by more users before being "weighed down" by the passage of time. However, likes will make some comments "bubble up" despite being old. The formula I'm using is inspired by [this one in Lemmy](https://join-lemmy.org/docs/contributors/07-ranking-algo.html) though it's worth noting we don't have a concept of downvotes. Since we need a mechanic to slightly weigh recent comparatively unpopular posts down, I've added another term to the exponential decay. Other than that, the formula is pretty similar. In short, it tries to balance showing newer stuff with surfacing some of the well-liked older stuff.

Again, the goal isn't to get this 100% right on the first try. However, we'd like conversations to be more dynamic, even in larger threads, and to give more people a chance to be seen — even if they haven't built a following. The exact heuristic will change over time and will likely incorporate other signals (e.g. whether OP liked the reply, whether OP is mutual, etc) so its name ("hotness") is intentionally fuzzy and non-committal. It's also worth emphasizing that this doesn't change much for smaller conversations because people who follow each other already see each other at the top of any comment sections, regardless of their chosen sorting function.

## Test Plan

Verify a user with no explicit past chosen preference gets the "hotness" ranking. (This part depends on https://github.com/bluesky-social/atproto/pull/3082.)

Verify a user with an existing past chosen preference keeps their existing preference.

Verify the hotness ranking prioritizes recent posts (comment out the "is self post" check, post something, and verify your comment appears at the top). Then verify that hotness decays over time — add `+ 24` or `+ 48` etc to the `hoursAgo` variable and verify that your post (with no likes) starts moving downwards (simulating time passing). Then hardcode some `likeCount` for your test post and verify that likes let it "bubble up" somewhat.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "icons:optimize": "svgo -f ./assets/icons"
   },
   "dependencies": {
-    "@atproto/api": "^0.13.11",
+    "@atproto/api": "^0.13.18",
     "@braintree/sanitize-url": "^6.0.2",
     "@discord/bottom-sheet": "bluesky-social/react-native-bottom-sheet",
     "@emoji-mart/react": "^1.1.1",
--- a/src/screens/Settings/ThreadPreferences.tsx
+++ b/src/screens/Settings/ThreadPreferences.tsx
@@ -56,6 +56,12 @@ export function ThreadPreferencesScreen({}: Props) {
                 values={sortReplies ? [sortReplies] : []}
                 onChange={values => setThreadViewPrefs({sort: values[0]})}>
                 <View style={[a.gap_sm, a.flex_1]}>
+                  <Toggle.Item name="hotness" label={_(msg`Hot replies first`)}>
+                    <Toggle.Radio />
+                    <Toggle.LabelText>
+                      <Trans>Hot replies first</Trans>
+                    </Toggle.LabelText>
+                  </Toggle.Item>
                   <Toggle.Item
                     name="oldest"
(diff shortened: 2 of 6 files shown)


haileyok (approved):

trivikr:
Posted PR to update `@atproto/dev-env` in https://github.com/bluesky-social/social-app/pull/6652

CrustyBumbler:
On the Android app, this got set as the default on update despite me having chosen Oldest First previously.

PR #6655, opened 2024-11-23
Expose "Copy link to post" menu item in Post Dropdown menu

### LINKED ISSUE #5912: Expose ‘Copy link to post’ list item on all platforms
### Describe the Feature

The ‘Copy link to post’ item in the post context menu is currently hidden / replaced on mobile by a ‘Share’ item which calls the system share sheet. While some share sheets do allow for the copying of the shared URL, this isn't always the case (some Android forks seem to make this more of an ordeal than it needs to be). It also requires at least one additional action compared to just having the option there.

I think the ‘Copy link to post’ list item should be displayed alongside the ‘Share’ item on mobile.

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

HorusGoul:
Fixes #5912.

TBD

### DIFF
--- a/src/lib/sharing.ts
+++ b/src/lib/sharing.ts
@@ -21,7 +21,11 @@ export async function shareUrl(url: string) {
   } else {
     // React Native Share is not supported by web. Web Share API
     // has increasing but not full support, so default to clipboard
-    setStringAsync(url)
-    Toast.show(t`Copied to clipboard`, 'clipboard-check')
+    copyUrl(url)
   }
+}
+
+export function copyUrl(url: string) {
+  setStringAsync(url)
+  Toast.show(t`Copied to clipboard`, 'clipboard-check')
 }
--- a/src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
+++ b/src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
@@ -21,7 +21,7 @@ import {useOpenLink} from '#/lib/hooks/useOpenLink'
 import {getCurrentRoute} from '#/lib/routes/helpers'
 import {makeProfileLink} from '#/lib/routes/links'
 import {CommonNavigatorParams, NavigationProp} from '#/lib/routes/types'
-import {shareUrl} from '#/lib/sharing'
+import {copyUrl, shareUrl} from '#/lib/sharing'
 import {logEvent} from '#/lib/statsig/statsig'
 import {richTextToString} from '#/lib/strings/rich-text-helpers'
 import {toShareUrl} from '#/lib/strings/url-helpers'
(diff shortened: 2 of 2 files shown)


src/lib/sharing.ts
```diff
@@ -21,7 +21,11 @@ export async function shareUrl(url: string) {
...
     // React Native Share is not supported by web. Web Share API
     // has increasing but not full support, so default to clipboard
-    setStringAsync(url)
-    Toast.show(t`Copied to clipboard`, 'clipboard-check')
+    copyUrl(url)
   }
 }
+
+export function copyUrl(url: string) {
+  setStringAsync(url)
```
HorusGoul, line 29: question: `setStringAsync` could fail. Should we make this function async and await the `setStringAsync(url)` call to ensure we actually copy the url before showing the toast?

src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
```diff
@@ -691,6 +712,16 @@ let PostDropdownMenuItems = ({
...
 
+      <Prompt.Basic
+        control={copyPostUrlLoggedOutWarningPromptControl}
+        title={_(msg`Note about sharing`)}
+        description={_(
+          msg`This post is only visible to logged-in users. It won't be visible to people who aren't logged in.`,
+        )}
+        onConfirm={onCopyPostUrl}
+        confirmButtonCta={_(msg`Share anyway`)}
+      />
```
HorusGoul, line 723: I duplicated the prompt for simplicity, but we might want to change the copy here. Copying a link might not always result in sharing, but it's most likely the case.

src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
```diff
@@ -429,20 +436,34 @@ let PostDropdownMenuItems = ({
...
-                onSharePost()
+                onCopyPostUrl()
               }
             }}>
-            <Menu.ItemText>
-              {isWeb ? _(msg`Copy link to post`) : _(msg`Share`)}
-            </Menu.ItemText>
-            <Menu.ItemIcon icon={Share} position="right" />
+            <Menu.ItemText>{_(msg`Copy link to post`)}</Menu.ItemText>
+            <Menu.ItemIcon icon={ChainLink} position="right" />
```
HorusGoul, line 466: question: I chose this icon because it's related to links, but Web used the `Share` icon before. Do we want to keep using the `Share` icon for web?

surfdude29:
As an iOS user, this would be great to have as it simply makes copying the URL easier and faster.

But of course more importantly, it seems like it would solve a big issue for some Android users.

I hope the team are able to take a look at this soon and hopefully get it merged 🙏

mozzius:
Thank you for this! We're actually planning on going a different route and adding a dedicated share menu, which would include the option to copy to clipboard

mozzius:
Superceded by #7840 - thank you for the contribution! Much appreciated 🙏

PR #6657, opened 2024-11-23
Fix translate button not working

### LINKED ISSUE #6287: After clicking the 'Translate' button, Bluesky will returns to the home page

### LINKED ISSUE #6446: Timeline scroll reset when going back from translation
### Steps to Reproduce

On the timeline translate a post, get back from the Google translate view, the timeline will be scrolled to the top thus "loosing" the original translated post

### Attachments

_No response_

Android

### Device Info

_No response_

Build version: 1.93.0.373; Bundle info: ee184d2cd (prod); Bundle date: 24111518; Platform: android

### Additional Information

Need to use the back button to come back to the app from Google translate opened inside the app (perhaps loss of a state)

auroursa:
Based on #6289

Fix #6287
Fix #6446

Sorry I know this is a bit repetitive, but I’m trying to provide more explanation in this PR. I've encountered this issue frequently over the past few days.

Thanks to @surfdude29’s [comment](https://github.com/bluesky-social/social-app/issues/6287#issuecomment-2487721453), made me realize that there is a behavioral difference between clicking "Translate" in the post context menu and directly clicking the translate button under the post, so I started investigating this.

In #6084, to fix the a11y issue, we replaced `<NewText>` with `<InlineLinkText>`, which introduced `to="#"`. But this breaks the translation and redirects the user to the homepage.

I absolutely respect the a11y fix, so I won’t revert this change. Replacing `onPress={onTranslatePress}` with `to={translatorUrl}` seems like the right approach. After removing `onPress` here, it will no longer be used elsewhere, so it can be safely removed import.

After testing, the translation button will work again.

https://github.com/user-attachments/assets/ab260274-005d-483c-b16c-0dfa11c2844c

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -1,5 +1,10 @@
 import React, {memo, useMemo} from 'react'
-import {StyleSheet, Text as RNText, View} from 'react-native'
+import {
+  GestureResponderEvent,
+  StyleSheet,
+  Text as RNText,
+  View,
+} from 'react-native'
 import {
   AppBskyFeedDefs,
   AppBskyFeedPost,
@@ -12,6 +17,7 @@ import {msg, Plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
 import {MAX_POST_LINES} from '#/lib/constants'
+import {useOpenLink} from '#/lib/hooks/useOpenLink'
 import {usePalette} from '#/lib/hooks/usePalette'
 import {makeProfileLink} from '#/lib/routes/links'
 import {sanitizeDisplayName} from '#/lib/strings/display-names'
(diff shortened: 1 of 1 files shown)


gaearon:
I'm still not able to reproduce this.

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -733,13 +732,8 @@ function ExpandedPostDetails({
   const t = useTheme()
   const pal = usePalette('default')
   const {_, i18n} = useLingui()
-  const openLink = useOpenLink()
   const isRootPost = !('reply' in post.record)
 
-  const onTranslatePress = React.useCallback(() => {
-    openLink(translatorUrl, true)
```
gaearon, line 740: This was here intentionally so that we can skip the consent modal for the translate flow

gaearon:
OK, I can repro it on web at least

gaearon:
Slightly tweaked the fix. Couldn't repro the original issue outside web but I think this should work.

gaearon (approved):

PR #6658, opened 2024-11-23
Fix Newline handling in Web's Composer

### LINKED ISSUE #6365: Newline handling in web client is not correct.

HorusGoul:
Fixes copying and pasting text that includes line breaks (`\n`) in the Web Composer.

Tiptap doesn't maintain newlines because this is a default from Prosemirror. Context: https://github.com/ueberdosis/tiptap/issues/775#issuecomment-920684174

Fixes #6365.

**Changes:**

- Setting the `blockSeparator` to `\n` instead of the default `\n\n`. Fixes the copying behavior.
- Adding a custom `clipboardTextParser`. The same as the code provided by @glavdir in https://github.com/ueberdosis/tiptap/issues/775#issuecomment-762971612, but with type annotations.

## Manual testing
https://github.com/user-attachments/assets/b3c6e4c1-0b5e-4805-aeae-72a5a13152cb

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -11,6 +11,7 @@ import {Paragraph} from '@tiptap/extension-paragraph'
 import {Placeholder} from '@tiptap/extension-placeholder'
 import {Text as TiptapText} from '@tiptap/extension-text'
 import {generateJSON} from '@tiptap/html'
+import {Fragment, Node, Slice} from '@tiptap/pm/model'
 import {EditorContent, JSONContent, useEditor} from '@tiptap/react'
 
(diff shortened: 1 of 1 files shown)


gaearon (approved):
thanks, this was very annoying but i couldn't tell where the mistake is

PR #6662, opened 2024-11-23
Fix avi button hitslops

gaearon:
We're trying to fix the avi button hitslops on Android. Currently, Android has significant dead zones where you can't click:

https://github.com/user-attachments/assets/15f4b502-211f-4ce2-a5ef-d59bae9cfa3d

https://github.com/bluesky-social/social-app/pull/6639 seemed like it would work but it did not. We're doing some wonky stuff with pressables and positioning so I tried to simplify it. See commit by commit.

In short:

- Remove `hitSlop` on a wrong Pressable that was swallowing taps (but not handling them) on Android
- Fix a clipping issue that caused touches outside the avi "column" layout getting dropped on Android
- Adjust hitslops for the avi button to get out of the way for pressing the pic (nudge them bottom-right)

Tested on a real device on Android and iOS.

## Test Plan

- Verify there are no dead zones (especially on Android). Tapping _always_ does something.
- Tapping in the center of the avi on a real device no longer brings up the menu.
- Tapping in the right bottom works reliably (especially on Android).
- There is still visual feedback (I tweaked the opacity though).
- Other `NativeDropdown` usages keep working ("..." dropdown in `ProfileList`).
- Other `AviFollowButton` usages keep working (post in search, post in notifications).

https://github.com/user-attachments/assets/eedcb085-e735-48fa-adf8-157008f6b271

https://github.com/user-attachments/assets/3cd3d1e0-0dc4-4715-9108-b4669606f658

### DIFF
--- a/src/view/com/posts/AviFollowButton.tsx
+++ b/src/view/com/posts/AviFollowButton.tsx
@@ -88,14 +88,21 @@ export function AviFollowButton({
             a.rounded_full,
             a.absolute,
             {
-              bottom: -4,
-              right: -4,
+              bottom: -7,
+              right: -7,
             },
           ]}>
           <NativeDropdown items={items}>
             <View
               style={[
-                {width: 24, height: 24},
+                {
+                  // An asymmetric hit slop
+                  // to prioritize bottom right taps.
+                  paddingTop: 2,
+                  paddingLeft: 2,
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
Oh very nice

PR #6676, opened 2024-11-23
Fix skeleton alignment

gaearon:
Fixes Notifications skeleton alignment to match the actual sizes (which changed at some point).

This gets the left top avatar to pixel-perfect alignment.

## Test Plan

https://github.com/user-attachments/assets/80afb515-720d-4d79-bd04-3c42856ffe2f

https://github.com/user-attachments/assets/05f9113f-d8c7-4af3-ad84-063569e3ef62

https://github.com/user-attachments/assets/ad06da3e-6d81-4a77-b765-0a51bd19ec5a

### DIFF
--- a/src/view/com/util/LoadingPlaceholder.tsx
+++ b/src/view/com/util/LoadingPlaceholder.tsx
@@ -139,7 +139,7 @@ export function NotificationLoadingPlaceholder({
   const pal = usePalette('default')
   return (
     <View style={[styles.notification, pal.view, style]}>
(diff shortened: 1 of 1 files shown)


src/view/com/util/LoadingPlaceholder.tsx
```diff
@@ -309,7 +309,7 @@ const styles = StyleSheet.create({
     padding: 5,
   },
   avatar: {
-    borderRadius: 26,
+    borderRadius: 999,
```
gaearon, line 312: so we don't have to keep updating this.

estrattonbailey (approved):

PR #6680, opened 2024-11-24
Fix format number polyfill for gl locale

gaearon:
Fixes the test that started failing on main with https://github.com/bluesky-social/social-app/pull/6463.

cf. https://unicode.org/cldr/charts/42/verify/numbers/gl.html


khuddite (commented):
Lgtm! Thanks for the fix!

mozzius (approved):

PR #6691, opened 2024-11-24
Spanish localization - completion and some fixes

hogaza:
Added some missing translations, corrected a few inconsistencies (paquete inicial; video) and edited to strings to use gender-neutral language. This PR complements the changes made in #6094


src/locale/locales/es/messages.po
```diff
@@ -1064,7 +1063,7 @@ msgstr "Las opciones de reproducción automática se han mudado a <0>Ajustes de
 #: src/screens/Settings/ContentAndMediaSettings.tsx:82
 #: src/screens/Settings/ContentAndMediaSettings.tsx:88
 msgid "Autoplay videos and GIFs"
-msgstr "Reproducir vídeos y GIFs automáticamente"
+msgstr "Reproducir videos y GIFs automáticamente"
```
Kloen, line 1066: `vídeos`

https://x.com/RAEinforma/status/1514273473794871303

*Same for the other cases as the translation is Spanish (es-ES) I'd say*

surfdude29: If there are enough differences to make it worthwhile, I wanted to mention that it would be possible to fork and make a Spanish (Latin America) (`es-419`) localization.

This was proposed earlier in #2472, but given the increased size in the user base since January and open approach to accepting other localizations I think the Bluesky devs would likely be open to that (although no guarantees of course).

src/locale/locales/es/messages.po
```diff
@@ -1,6 +1,5 @@
 msgid ""
-msgstr ""
-"POT-Creation-Date: 2023-12-22 01:56+0530\n"
+msgstr "POT-Creation-Date: 2023-12-22 01:56+0530\n"
```
surfdude29, line 2:
```suggestion
msgstr ""
"POT-Creation-Date: 2023-12-22 01:56+0530\n"
```
I think there's a little glitch here, these lines should remain the same :)

src/locale/locales/es/messages.po
```diff
@@ -8311,7 +8310,7 @@ msgstr "Demutear lista"
 
 #: src/lib/strings/errors.ts:11
 msgid "Unable to connect. Please check your internet connection and try again."
-msgstr "No es posible conectarse. Comprueba tu conexión a Internet e inténtalo de nuevo "
+msgstr "No es posible conectarse. Comprueba tu conexión a Internet e inténtalo de nuevo"
```
surfdude29, line 8313:
```suggestion
msgstr "No es posible conectarse. Comprueba tu conexión a Internet e inténtalo de nuevo."
```
nit: missing `.` at end here

src/locale/locales/es/messages.po
hogaza: Thanks for raising this @Kloen. I noticed inconsistencies between vídeo/video in the strings, hence this change. My approach focused on bringing consistency to the existing content. In thsi case "video" was the most frequent occurence. 

If indeed the translation is es-ES instead of a "neutral" approach, I agree with you that it should be "vídeo".

Kloen: @surfdude29 I think there's in fact enough differences to make it worthwhile to have a es-ES and es-419 localizations tho. While the base is the same, Latin America has a lot of subtle differences among all the countries the language is spoken in, and from the pov of a native spanish(es-ES) speaker it does feel odd a lot of times. *I find myself complaining about it on games or other applications, to a point I end up defaulting to English.*

auroursa: > @surfdude29 I think there's in fact enough differences to make it worthwhile to have a es-ES and es-419 localizations tho. While the base is the same, Latin America has a lot of subtle differences among all the countries the language is spoken in, and from the pov of a native spanish(es-ES) speaker it does feel odd a lot of times. _I find myself complaining about it on games or other applications, to a point I end up defaulting to English._

I think this is a sufficient reason to restart the Latin American Spanish localization, especially since I noticed that [Mastodon](https://crowdin.com/project/mastodon) also distinguishes between `es-ES`, `es-AR`, and `es-MX`.

But I think #2472 is already too outdated, perhaps we can use it as a partial foundation to start a new one.

hogaza: I would add that before adding new Spanish locales, it would be wise to first focus on polishing the current translations and resolve inconsistencies. This would facilitate then the adaptation to es-419.

dgruano:
Hi all!

Great to see there's so much movement trying to improve the Spanish localization. I found some typos, inconsistencies and translations worth discussing. Should I add them here or create a new PR with a separate discussion? Happy to contribute in any case!

surfdude29:
@dgruano I would suggest to add them here, my view is that ideally there should only be one PR open at a time for each localization to keep discussion and feedback consolidated.

pfrazee (approved):
Thanks everyone!

PR #6698, opened 2024-11-24
Don't re-sort already fetched post thread items

gaearon:
This took me a little bit to figure out but I think the result works well.
Might be easier to review the entire thing than commit by commit.

## The Problem

Currently, if your thread sorting preference is anything other than "oldest first", the sort order between refetches is not guaranteed to be stable. As a result, you'll often get layout shifts:

1. Start replying to somebody's reply from the parent post thread
2. The thread refetches
3. The order may be different, and now you're looking at a random comment

This is very easily observable with the "random" sort order:

https://github.com/user-attachments/assets/9d3ab603-0b1e-4eb3-9c2f-c6d54f6e948d

Although this was _most_ severe with the "random" sort order (where every attempt to reply reorders everything), it was also an issue with the "newest" or "liked" order. The new "hotness" mode also suffers from this, motivating the fix.

## A Solution

My proposal to fix this is to **avoid changing the order for the items that have previously been fetched and sorted.** Concretely, this means:

1. OP, Self, and Friend posts are still prioritized above everything
2. Other new items from each next refetch are conceptually grouped into a separate new "generation"
3. Items from older generations always sort above items from newer generations
4. Items within the same generation sort using the existing algorithms
5. Comparisons within the same generation are tweaked to be deterministic

All caching is local to the current post thread UI component and doesn't apply to pushes or pops.

Consequences of new behavior:

- On the first fetch, it's identical to the old behavior. The sorting order is applied as usual.
- On a refetch (e.g. due to replying), the previously unseen (newly fetched) items are placed at the bottom.
  - Prioritized items (OP post, self post, friends post) remain an exception. They're "bounded" so shifting a bit is OK.
  - Other than that, unseen items are placed at the bottom.
  - Within that new bottom slice, the unseen items are sorted among themselves with the selected order.
- If you exit the thread and push into it again, you'll see the pristine (everything sorted) state again.

## Test Plan

This is easiest to observe with the Random order. Previously, a newly fetched reply from an unrelated account would cause you to lose your position. With the change (recorded below), newly fetched non-priority replies are placed at the bottom:

https://github.com/user-attachments/assets/096483c3-43e9-4201-a955-ab31611c5a6a

You can also observe this with Hotness. Previously, a newly fetched reply from an unrelated account would cause you to lose your position. With the change (recorded below), newly fetched replies non-priority are placed at the bottom:

https://github.com/user-attachments/assets/6c92d563-7227-4ad3-8169-077ea07659ae

### DIFF
--- a/src/state/queries/post-thread.ts
+++ b/src/state/queries/post-thread.ts
@@ -242,11 +242,11 @@ export function sortThread(
 
       // Split items from different fetches into separate generations.
       let aFetchedAt = fetchedAtCache.get(a.uri)
-      let bFetchedAt = fetchedAtCache.get(b.uri)
       if (aFetchedAt === undefined) {
         fetchedAtCache.set(a.uri, fetchedAt)
         aFetchedAt = fetchedAt
       }
+      let bFetchedAt = fetchedAtCache.get(b.uri)
       if (bFetchedAt === undefined) {
         fetchedAtCache.set(b.uri, fetchedAt)
         bFetchedAt = fetchedAt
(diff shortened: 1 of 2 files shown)


pfrazee (approved):
Nice, that's a clever solve

PR #6682, opened 2024-11-24
Update catalan messages.po

ivanbea:
Back to 100% translated.
Applied pending corrections by @darcio and @jordimas https://static.softcatala.org/quality/bluesky_social_app.html

Please take a look at it @jordimas @darccio @surfdude29 @rortan134 @magarcia

PR #6699, opened 2024-11-24
Update messages.po in Catalan

aryiu:
Fixed and improved some Catalan strings


surfdude29:
There's already a PR open for updating the Catalan localization at #6682, would it be possible to discuss changes there?

PR #6701, opened 2024-11-24
Update messages.po in Catalan

aryiu:
Fixed and improved some Catalan strings:

- "starter pack" → "paquet inicial", to avoid anglicisms.
- "informar" → "denunciar", changed to avoid ambiguity. I think "denunciar" could be a better choice since it's a more specific term, while "informar" can be too general.
- Fixed the use of "tu" instead of "vos", to maintain the same voice in all strings.
- Added some "el dia" in strings with dates, for a better sense in Catalan. For example,`Arxivat del dia 11 d'octubre, 2024` instead of `Arxivat de 11 d'octubre, 2024`
- Added ":" when a name follows, to avoid wrong apostrophization. For example,`Paquet inicial de: Andreu` instead of `Paquet inicial de Andreu`. If it is the handle, I think this should be the same, because the "@" symbol is called "arroba" and should be apostrophed, so I think it's better to add a ":" before it.

### DIFF
--- a/src/locale/locales/ca/messages.po
+++ b/src/locale/locales/ca/messages.po
@@ -655,7 +655,7 @@ msgstr "Afegeix una contrasenya d'aplicació"
 
 #: src/view/com/modals/report/Modal.tsx:194
 #~ msgid "Add details to report"
-#~ msgstr "Afegeix detalls a l'informe"
+#~ msgstr "Afegeix detalls a la denúncia"
 
 #: src/view/com/composer/Composer.tsx:467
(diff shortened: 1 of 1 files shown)


ivanbea:
I have taken your suggestions into consideration on this PR https://github.com/bluesky-social/social-app/pull/6682

You can close this one

pfrazee:
Closing in favor of the PR, thank you so much @aryiu. If you think the other PR didn't capture it all, or need to discuss, feel free to reopen!

PR #6700, opened 2024-11-24
Remove broken translated strings

gaearon:
Something got borked with some translations during the last release: https://github.com/bluesky-social/social-app/pull/6463#issuecomment-2496116888

I'm removing things that give me build errors.

## Test Plan

Running `yarn` no longer errors

PR #6702, opened 2024-11-24
Fix glitches in gl/messages.po

surfdude29:
This should hopefully fix the [glitches](https://github.com/bluesky-social/social-app/pull/6463#issuecomment-2496116888) in the gl messages.po file.

Although I'm a bit baffled by the last error shown, as I can't see that string in the file? EDIT: last error is in the Italian file, which explains it

### DIFF
--- a/src/locale/locales/gl/messages.po
+++ b/src/locale/locales/gl/messages.po
@@ -191,7 +191,7 @@ msgstr ""
 
 #: src/components/LabelingServiceCard/index.tsx:96
 msgid "{count, plural, one {Liked by # user} other {Liked by # users}}"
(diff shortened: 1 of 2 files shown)


gaearon:
Ah nice, mind also fixing the Italian one? https://github.com/bluesky-social/social-app/pull/6700

gaearon:
Ahh so `other` here is syntax and translating it is the mistake. That makes sense.

gaearon (approved):

gaearon (approved):

surfdude29:
> Ahh so `other` here is syntax and translating it is the mistake. That makes sense.

@gaearon Just opened #6703 👍

PR #6487, opened 2024-11-18
Update Italian translations

mlocati:
I've added some missing translations, as well as fixed some ugly translations, and used always the same translation terms for some original translation terms (eg "handle" is now always "nome utente").

PR #6703, opened 2024-11-24
Fix `it` string

surfdude29:
This should fix the broken `it` string highlighted [here](https://github.com/bluesky-social/social-app/pull/6487#pullrequestreview-2456759449).


gaearon:
just did this in https://github.com/bluesky-social/social-app/pull/6702, sorry haha

surfdude29:
Np!

PR #6456, opened 2024-11-17
Romanian localization

claudiu-cristea:
This PR provides the Romanian localization

## Glossary

~~Some term localizations still need discussion.~~ EDIT: All glossary terms were discussed and we've agreed to this list

| Original       | Romanian                   | Not definitive | Comments |
|----------------|----------------------------|----------------|----------|
| account        | cont                       |                |          |
| chat           | conversație                |                |          |
| conversation   | conversație                |                |          |
| direct message | mesaj privat               |                |          |
| discover       | descoperă                  |                |          |
| feed           | flux                       |                |          |
| follower       | urmăritor                  |              |          |
| following      | urmărire                   |               |          |
| handle         | numele de utilizator              |               |          |
| label          | etichetă                   |                |          |
| like           | apreciere                  |                |          |
| message        | mesaj                      |                |          |
| mute (verb)          | a pune pe mut |               |          |
| mute          | pus pe mut |               |          |
| post           | postare                    |               |          |
| quote          | citare                      |                |          |
| reply          | răspuns                    |                |          |
| repost         | repostare                  |               |          |
| starter pack   | pachet de început            |                |          |
| tag            | etichetă                   |                |          |
| thread         | discuție              |               |          |
| unmute (verb)      | scoate de pe mut    |               |          |

PR #6712, opened 2024-11-25
Add Romanian language support

alextecplayz:
This PR adds Romanian (code: ro) localization to the Social app.


auroursa:
I found that another Romanian localization PR #6456

Would you mind integrating your contributions into that PR? I think having just one PR would help speed up the Romanian localization effort.

alextecplayz:
Thanks, I'll have a look. I haven't noticed it.

PR #6719, opened 2024-11-25
fix: keyboard handler memoization

kirillzyusko:
The array of dependencies is needed to prevent frequent re-attaching handlers. Internally `useKeyboardHandler` is based on `useEffect` and without dependencies `useEffect` will be fired on each re-render (which may affect performance).

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -250,30 +250,33 @@ export function MessagesList({
   // We use this value to keep track of when we want to disable the animation.
   const layoutScrollWithoutAnimation = useSharedValue(false)
 
-  useKeyboardHandler({
-    onStart: e => {
-      'worklet'
-      // Immediate updates - like opening the emoji picker - will have a duration of zero. In those cases, we should
-      // just update the height here instead of having the `onMove` event do it (that event will not fire!)
-      if (e.duration === 0) {
-        layoutScrollWithoutAnimation.set(true)
+  useKeyboardHandler(
+    {
+      onStart: e => {
+        'worklet'
+        // Immediate updates - like opening the emoji picker - will have a duration of zero. In those cases, we should
+        // just update the height here instead of having the `onMove` event do it (that event will not fire!)
+        if (e.duration === 0) {
+          layoutScrollWithoutAnimation.set(true)
+          keyboardHeight.set(e.height)
+        } else {
+          keyboardIsOpening.set(true)
+        }
+      },
+      onMove: e => {
+        'worklet'
         keyboardHeight.set(e.height)
-      } else {
-        keyboardIsOpening.set(true)
-      }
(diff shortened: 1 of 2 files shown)


gaearon (approved):

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -250,30 +250,33 @@ export function MessagesList({
...
-      'worklet'
-      keyboardHeight.set(e.height)
-      if (e.height > bottomOffset) {
-        scrollTo(flatListRef, 0, 1e7, false)
-      }
-    },
-    onEnd: () => {
-      'worklet'
-      keyboardIsOpening.set(false)
+        if (e.height > bottomOffset) {
```
gaearon, line 269: Doesn't `bottomOffset` need to be a dependency?

In general it's not great that `useKeyboardHandler` has a dependencies API — the general design guideline is that custom Hooks should not accept deps (because the linter would not check them). If you rename the Hook to end with `Effect` in the name they'll be checked though I guess.

kirillzyusko: > Doesn't bottomOffset need to be a dependency?

You're right, fixed it 👀 

> In general it's not great that useKeyboardHandler has a dependencies API — the general design guideline is that custom Hooks should not accept deps (because the linter would not check them). If you rename the Hook to end with Effect in the name they'll be checked though I guess.

Well, hard to disagree here 🙃 The name of hook was chosen in such way to match a signature/name from reanimated hooks (such as `useAnimatedScrollHandler`). But you are right - on top of what I remember hooks from reanimated have automemozation (they automatically capture all variables from the closure into a worklet via babel plugin - correct me if I'm wrong), while `useKeyboardHandler` doesn't make any magic by default...

gaearon: Would it make sense to always register the same handler, but have that handler proxy to the latest available version of the function? Like we do in [`useNonReactiveCallback`](https://github.com/bluesky-social/social-app/blob/main/src/lib/hooks/useNonReactiveCallback.ts). The downside of this is that it loses reactivity — i.e. if some variable changes, the cleanup/setup won't refire. I don't know whether this is OK or not for the needs of your API.

kirillzyusko: > Would it make sense to always register the same handler, but have that handler proxy to the latest available version of the function? Like we do in [useNonReactiveCallback](https://github.com/bluesky-social/social-app/blob/main/src/lib/hooks/useNonReactiveCallback.ts). The downside of this is that it loses reactivity — i.e. if some variable changes, the cleanup/setup won't refire. I don't know whether this is OK or not for the needs of your API.

I've been thinking about this, but since I'm registering `worklet` handlers I'm not sure whether it'll work as expected. But I'll check your suggestion, thanks!

Issue #6687, opened 2024-11-24
Post interaction settings UI is confusing


blowdart:
### Steps to Reproduce

1. Start to create a new post and bring up the post interaction settings dialog

![image](https://github.com/user-attachments/assets/2d771d26-89e9-438a-9b66-75293947f90b)

2. Note the default quote posts enabled switch

3. Click the switch

4. Be confused, as the switch now indicates that quote posts enabled is off, but the text has changed, so if the UI is to be believed the "Quote posts disabled" setting is off, meaning quote posts are enabled.

![image](https://github.com/user-attachments/assets/57401031-8c62-4adb-8fc9-76ef5cecba9f)

The text needs to stay the same, and not change with the switch status. It could also be simplified to just to "Enabled"

### What platform(s) does this occur on?

iOS, Web (Desktop)

### What version of the app are you using?

Build version: 1.94.0; Bundle info: 3c30bb1 (prod); Bundle date: 24111623; Platform: web; Platform version: undefined

LegendEffects:
I'm proposing to change it to "Allow quote posts", the state is reflected in the toggle, not the text.

PR #6720, opened 2024-11-25
Make quote post enabled switch label text static

LegendEffects:
Fixes https://github.com/bluesky-social/social-app/issues/6687

Switches should not change text to reflect state as the state is already reflected in the switch itself; changing the text makes it more confusing to the user.


gaearon (approved):

gaearon:
agreed

PR #6727, opened 2024-11-25
Ensure react-native-keyboard-controller enabled state doesn't get overwritten

mozzius:
Using a ref counting approach to ensure that opening and closing a dialog doesn't overwrite the previous state

# Before

https://github.com/user-attachments/assets/53180633-469a-4175-b87f-7bc7c7dc5fd8

# After

https://github.com/user-attachments/assets/fbf3d616-0eab-430e-8ebf-1d3140a30e43

## Test plan

- Check the bug is fixed: focus the DMs input, open a dialog, then focus the input again
- Log the keyboard controller enabled state, and ensure it's the same as before (barring the incorrect behaviour which we fixed, so like most screens except messages and starter packs edit should be disabled)


patches/react-native-keyboard-controller+1.14.0.patch
```diff
@@ -0,0 +1,41 @@
...
+   // animated values
+   const progress = useAnimatedValue(0);
+   const height = useAnimatedValue(0);
+@@ -189,6 +187,10 @@ export const KeyboardProvider = ({
+     }
+   }, [enabled]);
+ 
++  useEffect(() => {
++    setEnabled(enabledProp)
++  }, [enabledProp])
```
gaearon, line 37: this still does a hop, you don't need this now because you can use `enabled` from props directly

patches/react-native-keyboard-controller+1.14.0.patch
```diff
@@ -0,0 +1,41 @@
...
+   statusBarTranslucent,
+   navigationBarTranslucent,
+-  enabled: initiallyEnabled = true,
++  enabled: enabledProp = true,
+ }: KeyboardProviderProps) => {
+   // ref
+   const viewTagRef = useRef<React.Component<KeyboardControllerProps>>(null);
+   // state
+-  const [enabled, setEnabled] = useState(initiallyEnabled);
++  const [enabled, setEnabled] = useState(enabledProp);
```
gaearon, line 27: just delete this one. and maybe rename `enabled` to `enabled_PATCHED` to make it clear how it works

patches/react-native-keyboard-controller+1.14.0.patch
```diff
@@ -0,0 +1,74 @@
...
++   * A boolean prop indicating whether the module is enabled.
+    * Defaults to `true`.
+    */
+   enabled?: boolean;
+@@ -79,12 +77,10 @@ export const KeyboardProvider = ({
+   children,
+   statusBarTranslucent,
+   navigationBarTranslucent,
+-  enabled: initiallyEnabled = true,
++  enabled: enabled_PATCHED = true,
```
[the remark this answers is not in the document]
mozzius: done

src/lib/hooks/useEnableKeyboardController.tsx
```diff
@@ -0,0 +1,78 @@
...
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  const [refCount, setRef] = useState(0)
+
+  const value = useMemo(
+    () => ({
+      incrementRefCount: () => setRef(count => count + 1),
+      decrementRefCount: () => setRef(count => Math.max(count - 1, 0)),
```
gaearon, line 30: as long as we're doing them correctly, it should never dip below 0 so i would remove this check. instead, could console.error in render if it's negative

gaearon (approved):
code lgtm aside from a nit. have not tested that it resolves the original issue

src/App.web.tsx
```diff
@@ -102,7 +102,7 @@ function InnerApp() {
   if (!isReady || !hasCheckedReferrer) return null
 
   return (
-    <KeyboardProvider enabled={false}>
+    <KeyboardControllerProvider>
```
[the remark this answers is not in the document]
mozzius: I believe RNKC complains if the context is missing. we could maybe leave the `<KeyboardProvider>` as-is?

src/lib/hooks/useEnableKeyboardController.tsx
```diff
@@ -0,0 +1,87 @@
...
+          return next
+        }),
+    }),
+    [],
+  )
+
+  return (
+    <KeyboardControllerRefCountContext.Provider value={value}>
+      <KeyboardProvider
+        enabled_PATCHED={refCount > 0}
```
kirillzyusko, line 47: I slightly don't understand a full motivation of the patch 🤔 

In my head I'm confident that next solution should work:
- you just mount `KeyboardProvider` with default `enabled={false}` value (no patch needed here)
- you keep your code but instead of changing `enabled_PATCHED` property you are using `setEnabled()` from `useKeyboardController` and in `useEffect` call the same method? So the code will look like:

```tsx
const {setEnabled} = useKeyboardController();
// other code with reference count
useEffect(() => {
  setEnabled(refCount > 0);
}, [refCount])
```

gaearon: Setting state directly in effects is an anti-pattern because it causes an entire rendering cycle in vain. If it's possible to control a prop directly, it's always better.

kirillzyusko: > Setting state in useEffect is messy

Well, if you think about internals of `setEnabled`, the yes, it's a react-antipattern (because internally `setEnabled` is based on `useState`). But I highly doubt it can produce infinte/loop state updates or something like that 🤷‍♂️ 

But if you just think about `useEffect` as a thing that runs effect if props gets changes, then I think it's okay to have such code? Because this is exactly what you want to achieve (to disable/enable module based on ref count)

kirillzyusko: @gaearon okay, if you are concerned about double state updates, then can't you use `useRef` for ref counting and call `setEnabled` in `increment`/`decrement` methods? In this case you will not update state inside effect + you will have only a single state update? 🤔

gaearon: Yeah I think that would work. Just seems unfortunate that this is almost 1:1 mirror of a controlled component pattern (like input takes controlled value and uncontrolled initialValue)

mozzius: Will do 👍

kirillzyusko: > Yeah I think that would work. Just seems unfortunate that this is almost 1:1 mirror of a controlled component pattern (like input takes controlled value and uncontrolled initialValue)

Well, I understand your frustration here... But `setEnabled` functionality was added only to use in some rare cases. From what I remember you started to disable `keyboard-controller` in https://github.com/bluesky-social/social-app/pull/4399 and then started to disable it no more and more screens.

I don't know why you disabled lib in other cases, but the original problem https://github.com/bluesky-social/social-app/pull/4399 from has been fixed in latest versions of keyboard-controller (originally fixed in 1.12.0 and a better fix that handles even more corner cases landed in 1.14.1)

So if it was the only one issue (incorrectly working keyboard-controller in Modals on Android), then this problem should be fixed in the lib itself.

If you have more use-cases where enabled keyboard-controller caused more issues and you had to turn it on and off, then let me know - will be happy to fix it in the package 😊 

Since your codebase is open it's very easy for me to reproduce the problem on my end 👀

gaearon: Fair enough, we can give it another try soon!

mozzius: Yeah, I have another branch where I enable it globally. Ran into some issues but nothing major!

gaearon (approved):

PR #6729, opened 2024-11-25
Fix camera orientation on Android

gaearon:
Fixes a bug where using the camera button would produce a stretched image on Android. The issue was that the `width`/`height` were misreported by not taking the EXIF tags into account. The iOS version does this correctly.

Upstream PR: https://github.com/ivpusic/react-native-image-crop-picker/pull/2110

## Test Plan

Try taking a camera image in all orientations (landscape/rotate with both front/back camera), verify it appears correctly both before posting (in the composer preview) and after posting.


mozzius (approved):
good stuff. fixes the orientation that was broken before, unsure how to trigger all cases but I feel good enough to stamp

PR #6737, opened 2024-11-25
Edit string to avoid churning translations

gaearon:
Undoing part of https://github.com/bluesky-social/social-app/pull/6720 for this release. I do think the new string is better though so I'll revert this back later.


mozzius (approved):
Let's revert tomorrow

surfdude29:
@gaearon @mozzius Was it still intended to revert this?

PR #6739, opened 2024-11-26
Fix Android keyboard bug when opening a sheet

mozzius:
On Android, opening a menu in DMs would cause the keyboard input to enter a strange state. This attempts to fix it by setting the keyboard height on the `onEnd` event and reimplementing KeyboardStickyView manually (turns out it's a very simple component, and we basically are already doing same the logic anyway for the scrollview)

# Before

https://github.com/user-attachments/assets/f53e9374-1d81-48ce-8d0e-63401604825f

# After

https://github.com/user-attachments/assets/91940a5a-b1df-46bd-8c18-f54aa9b540ce

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -267,7 +267,7 @@ export function MessagesList({
           scrollTo(flatListRef, 0, 1e7, false)
         }
       },
-      onEnd: () => {
+      onEnd: e => {
         'worklet'
         keyboardHeight.set(e.height)
         if (e.height > bottomOffset) {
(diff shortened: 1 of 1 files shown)


gaearon:
what’s the bug upstream? is there something we can contribute?

kirillzyusko:
@mozzius thanks! I prepared a fix in repository: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/704

Would you mind to try to test it? 👀

kirillzyusko:
Also you can reach out to me via [discord](https://discordapp.com/users/973642097888489512) (nickname is `kiryl.ziusko`) 

Currently @hirbod sending me yours findings in Discord, so maybe if we could have a direct channel of communication it would be more effective 😊

mozzius:
Confirmed that that patch fixes the issue, however actually I think we want to go with this approach - `KeyboardStickyView` currently move slightly differently, this aligns the behaviour so that they move together

gaearon (approved):
code lgtm, let's verify it works well on device

PR #6742, opened 2024-11-26
Fix post time localization on Android

### LINKED ISSUE #6728: Post time in Chinese cannot be localized on Android
### Steps to Reproduce

### Attachments

Android:
IMG_2295

iOS:
IMG_2294

Android

### Device Info

Pixel 8, Android 15

### Additional Information

auroursa:
Fix #6728

It seems that the post date localization on Android is handled differently from Web and iOS. In some cases, this may cause the date localization to not work properly on Android. Here are some examples:

### Before

**Japanese (Web)**
screenshot 2024-11-26 131646

**Chinese (Web)**
screenshot 2024-11-26 131735

**Android**
| Chinese | Japanese |
| ----------- | ----------- |
|  |  |

For Chinese, this date format is like rolling back to English, so it needs to be changed. At the very least, it should match the date format used on the web. The example after the fix:

### After

**Android**
| Chinese | Japanese |
| ----------- | ----------- |
|  |  |


surfdude29:
This looks like a great solution, nice work 👌 (although I say that as someone who can't code or build the app to test it out, of course 😅)

One quick question: I can't read either Japanese or Chinese, are the correct date formats really exactly the same in the two languages? I must admit I never would have guessed that was the case 🤷🏻‍♂️

tkusano:
the screenshot attached as "After / Japanese (Android)" looks good to me.

auroursa:
It also fixes the localization of timestamps for archived posts.

### Chinese
| Before | After |
| ----------- | ----------- |
|  |  |


### Japanese
| Before | After |
| ----------- | ----------- |
|  |  |

src/lib/strings/time.ts
```diff
@@ -1,12 +1,36 @@
+import {DateTimeFormat} from '@formatjs/intl-datetimeformat'
```
gaearon, line 1: We'd like to avoid importing something that isn't used on the web. Let's instead fork this file like `time.ts` vs `time.web.ts` and keep this import in the native version. This also lets you remove the inline checks.

gaearon (changes requested):
let's tweak to fork the file instead

src/lib/strings/time.ts
gaearon: In fact, maybe it's worth moving the code into `time.android.ts` if we're sure the fix is only needed there

auroursa:
Reverted `time.ts`, and made the necessary changes to `PostThreadItem.tsx`

I believe the changes are only needed there for now, other time localizations seem fine.

surfdude29:
I think the entries for `an` and `ast` are missing (merged since you created this PR) 👀

https://github.com/bluesky-social/social-app/blob/b8b15f64d32f751186a50aa7df7826ef83ab2972/src/locale/i18n.ts#L51-L55

https://github.com/bluesky-social/social-app/blob/b8b15f64d32f751186a50aa7df7826ef83ab2972/src/locale/i18n.ts#L59-L63

`ast` is supported by [`@formatjs/intl-numberformat`](https://www.npmjs.com/package/@formatjs/intl-datetimeformat?activeTab=code) but not `an`.

For `an` I suggest `es` is used instead, like we do for `@formatjs/intl-numberformat`.

auroursa:
Spent some time checking the date localization for other languages, including the newly added `an` and `ast`, it should work fine.

Especially, I checked the time localization for English (UK) on Android to ensure it's in DDMMYYYY format. cc @surfdude29

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -740,6 +742,7 @@ function ExpandedPostDetails({
   const {_, i18n} = useLingui()
   const openLink = useOpenLink()
   const isRootPost = !('reply' in post.record)
+  const niceDate = isAndroid ? niceDateAndroid : niceDateWebOriOS
```
gaearon, line 745: We shouldn't need this. The build system will automatically pick up the `.android` file fork; you should import the main file like before.

auroursa: Sorry I didn't notice that before, revented `PostThreadItem.tsx`.

yarn.lock
```diff
@@ -4258,6 +4258,31 @@
     "@formatjs/intl-localematcher" "0.5.4"
     tslib "^2.4.0"
 
+"@formatjs/ecma402-abstract@2.2.4":
```
gaearon, line 4261: We have two copies of this in the lockfile now. I think we need to bump other related libraries in `package.json` so that we have everything in a lockstep and transitive dependencies are shared as much as possible.

src/locale/i18n.ts
```diff
@@ -1,8 +1,10 @@
 // Don't remove -force from these because detection is VERY slow on low-end Android.
 // https://github.com/formatjs/formatjs/issues/4463#issuecomment-2176070577
 import '@formatjs/intl-locale/polyfill-force'
+import '@formatjs/intl-datetimeformat/polyfill-force'
```
gaearon, line 4: Just flagging we need to measure the startup time impact of this on a real android device. We've had pretty bad regressions in the past due to related changes.

Another thing we might want to consider is forking this block between iOS and Android. If this is really android-only, it would be good to avoid activating this on iOS.

gaearon: Actually maybe you can move this import to `time.android.ts`? Since that's the only place that needs that API. I realize this makes it a bit inconsistent but it does avoid potentially regressing something on iOS.

gaearon: I guess I'm not sure whether that would break the dynamic imports below. Do they rely on ordering?

auroursa:
I will test it on Android and iOS later and update the `yarn.lock`.

About performance issues, will simulate a low-performance Android device to verify its impact.

Not entirely sure whether to introduce this on iOS as well since the behavior on iOS currently seems consistent with the Web.

Needs conduct more detailed tests over the next few days to carefully ensure nothing breaks.

gaearon:
I wonder what the source of discrepancy is. So far Android/iOS have been consistent with other 18n stuff.

auroursa:
> I wonder what the source of discrepancy is. So far Android/iOS have been consistent with other i18n stuff.

It seems that lingui has an issue when handling long-format dates, as this part is directly formatted through lingui.

Maybe I’ll create a minimal reproducible example to report this to upstream, as the issue consistently appears with both the previously used v4.5.0 and the latest v4.14.1 in RN 0.76. And it only affects certain asian languages on Android.

auroursa:
Just a personal thought, maybe we could try having iOS format the time using DateTimeFormat as well. We could still maintain the i18n-related imports in one place, which might help avoid additional maintenance effort.

After cleaning the lockfile, I retested on both iOS and Android, and the time formatting works correctly on iOS as well. I haven’t done rigorous flame graph testing on performance yet, but there doesn’t seem to be a significant impact on the app’s startup time on a simulated low-end Android device.

auroursa:
I submitted an [issue](https://github.com/lingui/js-lingui/issues/2123#issuecomment-2543394990) with a minimal reproducible example to Lingui, and this is the response I received:

> Lingui does not format dates by itself. It internally uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl.
> 
> If dates are not formatted correctly for you it's most likely because you need to set up the polyfill: https://lingui.dev/tutorials/react-native#polyfilling-intl-apis

Based on this I adjusted this PR. We still need to include `intl-datetimeformat`, but no longer need to fork `time.ts`.

timofei-iatsenko:
FYI, you also might be interested in that PR (not merged yet) in the lingui https://github.com/lingui/js-lingui/pull/2117

auroursa:
> FYI, you also might be interested in that PR (not merged yet) in the lingui [lingui/js-lingui#2117](https://github.com/lingui/js-lingui/pull/2117)

Thank you for pointing that out! 5.x also has some features that I'm particularly interested in, such as print placeholder values. I’ll check it again once it’s merged.

For now, I'd like to introduce polyfill to address the current issues, since it looks like migrating from 4.x to 5.x requires some changes beyond the scope of this PR. I admit I’ve been a bit eager to push this PR forward since it has been bothering me for some time.

Thanks again for clarifying my doubts so thoroughly!

gaearon:
Unrelated to this PR but noticed the DM conversation screen doesn't seem to translate dates:

Screenshot 2024-12-16 at 14 25 24

Wanna look into this?

auroursa:
> Wanna look into this?

Glad to look into this, but I’d like to learn more details. It seems to be working on Android when I tested. Does it happen when opening DM conversation after switching the language?

auroursa:
It hasn’t happened to me at this moment, but I have faced it before. Will look into it and see how to fix it when appropriate.

surfdude29:
Fwiw I see the DM issue too on iOS, e.g. here:
IMG_0872

PR #6746, opened 2024-11-26
Update favicon for development environment

clicktodev:
Change the favicon to reflect the development environment.

it's very easy for a developer to mistake which version is currently being used

expo docs on this https://docs.expo.dev/build-reference/variants/

we're not using variants, just changing the favicon on web

bug icon from lucide https://lucide.dev/icons/bug

cog icon from lucide https://lucide.dev/icons/cog

inspired by personal usage and @gaearon at https://bsky.app/profile/danabra.mov/post/3lbt4zicpuk27

image

included 2 designs for the favicon so i'd like your input on which one to use:

attempt 1:

image

image

attempt 2 - inverted colors:

image

image

attempt 3

image

image

attempt 4

image

attempt 5

image


auroursa:
Just a personal suggestion, perhaps we could try using caterpillar as the icon in the bottom right corner.

PR #6754, opened 2024-11-26

auroursa:

Meanwhile, significant updates have been made to Simplified Chinese, including standardizing the use of measure words, adopting more formal grammar, and reducing abbreviations.

Since introducing more Chinese variants is unlikely at this time, we aim to avoid incorporating region-specific terms into translations and strive to achieve regional neutrality.

Co-maintained by @KuwaLee @cirx1e @hakadao


surfdude29:
> Since introducing more Chinese variants is unlikely at this time, we aim to avoid incorporating region-specific terms into translations and strive to achieve regional neutrality.

Out of curiosity, which Chinese variants would you look to introduce sometime in the future?

While looking for something else I came across [this Mastodon issue](https://github.com/mastodon/mastodon/issues/18538#issuecomment-1142041630) which lists these:

```
zh	Chinese (zh)
Chinese (Simplified) (zh-Hans)
Chinese (Simplified, China) (zh-Hans-CN)
Chinese (Simplified, Hong Kong SAR China) (zh-Hans-HK)
Chinese (Simplified, Macau SAR China) (zh-Hans-MO)
Chinese (Simplified, Singapore) (zh-Hans-SG)
Chinese (Traditional) (zh-Hant)
Chinese (Traditional, Hong Kong SAR China) (zh-Hant-HK)
Chinese (Traditional, Macau SAR China) (zh-Hant-MO)
Chinese (Traditional, Taiwan) (zh-Hant-TW)
```

auroursa:
> > Since introducing more Chinese variants is unlikely at this time, we aim to avoid incorporating region-specific terms into translations and strive to achieve regional neutrality.
> 
> Out of curiosity, which Chinese variants would you look to introduce sometime in the future?
> 
> While looking for something else I came across [this Mastodon issue](https://github.com/mastodon/mastodon/issues/18538#issuecomment-1142041630) which lists these:
> 
> ```
> zh	Chinese (zh)
> Chinese (Simplified) (zh-Hans)
> Chinese (Simplified, China) (zh-Hans-CN)
> Chinese (Simplified, Hong Kong SAR China) (zh-Hans-HK)
> Chinese (Simplified, Macau SAR China) (zh-Hans-MO)
> Chinese (Simplified, Singapore) (zh-Hans-SG)
> Chinese (Traditional) (zh-Hant)
> Chinese (Traditional, Hong Kong SAR China) (zh-Hant-HK)
> Chinese (Traditional, Macau SAR China) (zh-Hant-MO)
> Chinese (Traditional, Taiwan) (zh-Hant-TW)
> ```

The Chinese variants we currently have:
- Chinese (Simplified, China) (zh-Hans-CN) aka. `zh-CN`
- Chinese (Traditional, Taiwan) (zh-Hant-TW) aka. `zh-TW`

Not included in the list, but maintained due to their many similarities with Chinese:
- Cantonese (Traditional, Hong Kong) (yue-Hant) aka. `zh-HK`

My long-term goal is to introduce the true `zh-HK`, and replace Cantonese in the correct position, i.e., `zh-yue` or `yue`.
- Chinese (Simplified, Hong Kong SAR China) (zh-Hans-HK)

Let me explain this point. For compatibility reasons, we are currently maintaining `yue` under the name `zh-HK`. So yeah, we gave up `zh-Hans-HK` / `zh-HK` to support Cantonese. Since the differences between `zh-Hant-HK` and `zh-Hant-TW` are small, I assume that people in Hong Kong can generally understand `zh-Hant-TW`. 

I’m not sure what issues might come up if we introduce ISO 639-3 language codes (three-letter codes) in AppLanguage, so it needs a lot of testing. I’d prefer to wait until Post Language supports these codes before considering renaming `zh-HK` back to `yue`.

If possible, I would also like to maintain Simplified Chinese for Singapore:
- Chinese (Simplified, Singapore) (zh-Hans-SG) aka. `zh-SG`

But with the current capabilities, I’m unable to achieve that, so the main goal of this PR is to ensure that users from Singapore who use Chinese can smoothly use `zh-CN`.

auroursa:
> I see... I appreciate the comprehensive explanation and thanks for indulging my linguistic curiosity 🤓

I'm happy to satisfy your linguistic curiosity.😊

By the way, Chinese rarely uses `zh` alone because it’s too broad. I believe the differences between `zh-CN` and `zh-TW` are greater than those between English (US) and English (UK).

As you can see, `zh-TW` is often labeled as "Traditional Chinese," while `zh-CN` is labeled as "Simplified Chinese." This distinction arises from China's reform in the last century, which altered the spelling of approximately 2,236 characters. The new spellings are more simplified compared to the traditional ones, while Taiwan chose to retain the original writing system. I believe the difference between `zh-CN` and `zh-TW` is more like the difference between Indonesian and Malaysian Malay. It’s worth maintaining both because they are quite distinct from each other.

Once we understand this, it's easy to distinguish between `zh-Hans` and `zh-Hant`. "s" is "simplified" and "t" is "traditional". This stems from a language movement, as many people believe that using zh-CN to represent Simplified Chinese is too simplistic. For example, I live in China, but I choose to use Traditional Chinese, so zh-CN doesn't accurately reflect this. But with the added distinction in the middle, we can easily make this differentiation, such as `zh-Hans-CN` and `zh-Hant-CN` (Yeah China still has a standard for Traditional Chinese).

But anyway, `zh-Hans` and `zh-Hant`, just like `zh`, are typically not used on their own because there are also vocabulary differences between regions. Using just `zh-Hans` / `zh-Hant` doesn't reflect this well (ah, the complexity of language).

For Hong Kong users, the written form of Chinese is very similar to `zh-TW` / `zh-Hant-TW`, with only minor differences in word usage and spelling conventions. But the language they speak is another language that is quite different from Chinese — Cantonese. Cantonese is more closely related to a spoken system of writing. This is why I chose to skip the true `zh-HK` and instead to maintain Cantonese for now. I believe this is kinder and more relatable for Hong Kong users, and this is what social media should aim to do.

auroursa:
Google's approach on Android: It tends to redirect `zh-Hans` to `zh-Hans-CN` / `zh-CN`, and `zh-Hant` to `zh-Hant-TW` / `zh-TW`. This is what I implemented in #6621. As for `zh`, Google seems to prefer redirecting it to `zh-Hans-CN`, but I disagree with this, and it would complicate the decision-making process. So I didn't include it in that PR.

References:
https://stackoverflow.com/a/48108664

pfrazee (approved):
Amazing as always. Thanks all

PR #6756, opened 2024-11-26
Fix `<List>` types

mozzius:
Cleans up the dodgy types we had here and correctly uses `Animated.FlatList` types instead of pretending to be a normal FlatList

# Test plan

See if CI passes


src/screens/Onboarding/Layout.tsx
```diff
@@ -35,7 +36,7 @@ export function Layout({children}: React.PropsWithChildren<{}>) {
   const {gtMobile} = useBreakpoints()
   const onboardDispatch = useOnboardingDispatch()
   const {state, dispatch} = React.useContext(Context)
-  const scrollview = React.useRef<ScrollView>(null)
+  const scrollview = React.useRef<Animated.ScrollView>(null)
```
mozzius, line 39: I have no idea why it won't let me use `ScrollView` here. Oh well, not a huge deal

PR #6762, opened 2024-11-26
[Nicer Tabs] Fork TabBar, simplify Pager

gaearon:
The goal is to resurrect https://github.com/bluesky-social/social-app/pull/1696. I've learned a bunch of things since I made that version, and I'm pretty sure I can get the indicator to work fast on Android now. Scrolling on demand may still be tricky but we'll see. Could be iOS-only.

In this PR, I'm just starting to clean things up a bit:

- `TabBar` is going to diverge even more between web and native, so I'll just fork it now.
- Remove `Pager`'s `onPageSelecting` event. It's kind of ill-defined (what does "selecting" mean exactly?), and the implementation encourages putting heavy work just as the deceleration is starting. This is not great for performance, especially if we start adding more visual flourishes to it. I believe the goal was to optimistically start some work earlier (maybe fetching the selected feed?) — but if that's the goal, we should just add some kind of query prewarming later. In either case, it gets in the way of refactoring, so let's kill it for now and revisit this decision at the end of the stack.

## Test Plan

Everything still works.

We're not sending the `reason` in the analytics event but that seems fine. Maybe I'll hook it up later, for now I want to simplify the code.

The "header catch up" animation after swipe now runs a bit later. We'll solve that in the subsequent PRs.

### DIFF
--- a/src/lib/statsig/events.ts
+++ b/src/lib/statsig/events.ts
@@ -80,12 +80,6 @@ export type LogEvents = {
     feedUrl: string
     feedType: string
     index: number
-    reason:
-      | 'focus'
-      | 'tabbar-click'
-      | 'pager-swipe'
-      | 'desktop-sidebar-click'
-      | 'starter-pack-initial-feed'
   }
   'feed:endReached': {
     feedUrl: string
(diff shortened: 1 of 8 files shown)


mozzius (approved):
nice simplification, works cross-platform

Issue #6511, opened 2024-11-19
Add toki pona to content languages


EtiTheSpirit:
### Describe the Feature

Currently, Bluesky supports a number of popular conlangs. One that has been making recent major strides in the number of speakers is toki pona, thanks to increased publicity, especially with presence in games (including VRChat and Minecraft, for example).

As a fluent speaker and an avid supporter of its community, I would really like to see it available as a language that can be used for content. This would help meet up with and find fellow community members, and potentially introduce new people to the language.

toki pona is [defined under ISO 639-3](https://iso639-3.sil.org/request/2021-043) as `tok`.

auroursa:
Related PR #5998.

Issue #6346, opened 2024-11-14
Language autodetected correctly but cannot be used on feeds or search


ivanbea:
### Steps to Reproduce

1- Post something in asturianu https://www.localeplanet.com/icu/ast-ES/index.html like "Ye cierto, ensin embargo nun sé por qué sí me sal automáticamente esta etiqueta:" and the language is recognized as "ast", also in the JSON "langs":["ast"]
2- Try to search something and filter by language, you cannot select Asturiano anywhere, even when the post is recognized on this language
Also with skyfeed you cannot select this language, I tried to edit the JSON like
 "id": "aaacx4f64znlo",
            "type": "remove",
            "subject": "language",
            "value": null,
            "language": "ast"
But it is not recognized.

I think this will happen with all the languages with a three-letter code (like ast), the language is recognized in the post but you can only select or search with two-letter codes.

### Attachments

See this post in asturianu

https://bsky.app/profile/iyanuvieu.bsky.social/post/3lawj5tvtvk2j

![bafkreicfa3btpisfe77gnqwrahqrgwpcy7gttarxlrvgthwuvmqv35svjy](https://github.com/user-attachments/assets/17739c9a-9e9e-480d-aea1-1862f1f27526)

### What platform(s) does this occur on?

iOS, Android, Web (Desktop), Web (Mobile)

### What version of the app are you using?

1.93.0.373

nuberlang:
I have the same problem, but I see that other languages ​​with three characters in the encoding do appear to select, for example Aragon.

auroursa:
> I have the same problem, but I see that other languages ​​with three characters in the encoding do appear to select, for example Aragon.

 Aragon has been assigned the code2 “an,” which is why it can be used as a Post Language. Currently, Post Language (as well as search language filters, etc.) is mapped one-to-one with code2. If a language isn’t included in the ISO 639-1 list, it can’t currently be used as a Post Language…

This is definitely an area that needs improvement, as many popular languages lack a code2.

iyanuvieu:
Asturian, ast, is a language that could be found forming parte of Iso 932

https://en.m.wikipedia.org/wiki/List_of_ISO_639-2_codes

auroursa:
> Asturian, ast, is a language that could be found forming parte of Iso 932
> 
> https://en.m.wikipedia.org/wiki/List_of_ISO_639-2_codes

[...]

PR #6769, opened 2024-11-27
Fully support languages with 3-character codes

tbodt:
Most languages do not have 2 character codes and will never get them, because the ISO committee no longer assigns them.

There are more things wrong with this language list, but I stopped here for now. Future work:

- Take the list from https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry, which is the list referenced by the HTML standard for possible values of `lang` attributes in HTML
- Get localized language names from CLDR
- Uh...figure out what to do about the fact that `lande` outputs `pes` for Persian, which actually means specifically Iranian Persian while `per` and `fas` and `fa` are Persian in general... man, no one really knows how to handle language families.

Have tested this on web on
- Language selector on post compose box
- Subtitle language selector
- Content language selector in settings
- Primary language selector in settings
- Language filter in search

### DIFF
--- a/src/locale/deviceLocales.ts
+++ b/src/locale/deviceLocales.ts
@@ -67,7 +67,7 @@ export function getLocales() {
 export const deviceLocales = getLocales()
 
 /**
- * BCP-47 language tag without region e.g. array of 2-char lang codes
+ * BCP-47 language tag without region
  *
  * {@link https://docs.expo.dev/versions/latest/sdk/localization/#locale}
  */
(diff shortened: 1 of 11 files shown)


tbodt:
@pfrazee I'd like to request a review from you as the author of #6207

pfrazee:
I'll get this on my tracker

jnpoJuwan:
@pfrazee Has there been any updates relating to this PR? There have been many people asking and asking for this in many threads. There is _no good justification_ for any platform to lack this level of basic language support when it prevents millions of people from interacting either reliably or even completely. These issues have persisted for over two years.

This issue may also be raised in the ATProto repo, which shares similar data and code.

----

Related issues:

* #1766
* #6346
* #6511

Related PRs:

* #1089
* #3204
* #4314
* #5998
* #6290
* #6774
* #7075
* #7174

tbodt:
Bump

PR #6774, opened 2024-11-27
Fix ast localization

auroursa:
[date-fns](https://www.npmjs.com/package/date-fns) does not define `ast`, we should remove the import, and set `ast` to `undefined`.

Although `ast` is a Code3 language code, the subsequent code requires a Code2 language code. Let's assign it as Code2 (even though it is not correct) so it can passes the check. This should not affect Post Language or other processing, as it is only related to the App Language.

### DIFF
--- a/src/components/hooks/dates.ts
+++ b/src/components/hooks/dates.ts
@@ -10,7 +10,6 @@
 import React from 'react'
 import {formatDistance, Locale} from 'date-fns'
 import {
-  ast,
   ca,
   de,
   enGB,
@@ -45,7 +44,7 @@ import {useLanguagePrefs} from '#/state/preferences'
  */
 const locales: Record<AppLanguage, Locale | undefined> = {
   en: undefined,
-  ast,
+  ast: undefined,
   an: undefined,
   ca,
   de,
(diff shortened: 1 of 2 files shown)


surfdude29:
cc: @estrattonbailey

gaearon (approved):

softastur:
This workaround could break web browser compatibility. I'll make the necessary changes on "date-fns" to work with "ast" locale. Please take a look to https://github.com/bluesky-social/social-app/issues/6346 issue

auroursa:
> This workaround could break web browser compatibility. I'll make the necessary changes on "date-fns" to work with "ast" locale. Please take a look at https://github.com/bluesky-social/social-app/issues/6346 issue

date-fns is a library used for showing countdowns in the “Muted words & Tags”. If it is set to undefined, it will fallback to English.

I’d like to know more details about your concerns, as this workaround won’t break the proper setting of the HTML language attribute.

image

softastur:
Interesting @auroursa . For some reason «ast» locale is not available on app language. content languages and primary language settings using pre-configured ast locale phone.

auroursa:
Hi @softastur I believe the current build of the App is based on a commit prior to this PR, or you can check the App version information to find the exact commit used for the build.

This should become fully available in later OTA update or next version 1.96.0.

softastur:
Hi @auroursa. Web client contains ast locale available only on app language settings since 1.95.0 version

auroursa:
> Hi @auroursa. Web client contains ast locale available only on app language settings since 1.95.0 version

If you’re referring to using `ast` as post language, we should wait for #6769 to be merged. This PR only ensures that `ast` can be used as app display language.

surfdude29:
I hadn't noticed earlier but the 1.95 OTA update rolled out a few days ago so the Asturian localization is now available on mobile as well as on web.

PR #6778, opened 2024-11-27
 Corrected the translation of "Back" to "Retour" for better accuracy in French

lsaudon:


auroursa:
cc @Signez

Signez:
Good catch here. (It's an accessibility label that was translated eons ago, and it slipped our checks when we did a review of the whole translation a few months ago.)

✅ Please consider this pull request reviewed positively!

Signez:
I can’t close it myself, but please consider this one actually merged (through #6991, which was merged).

Thanks @lsaudon! (If you can close it, it would be awesome!)

PR #6783, opened 2024-11-27
Fix gap between Write Your Reply and bottom bar on web

gaearon:
Mobile web only.

Does not affect native (it has a bottom safe inset that's larger), does not affect tablet or desktop.

## Before

Screenshot 2024-11-27 at 14 44 43

## After

Screenshot 2024-11-27 at 14 44 54


src/view/com/post-thread/PostThread.tsx
```diff
@@ -538,13 +538,14 @@ export function PostThread({uri}: {uri: string | undefined}) {
 function MobileComposePrompt({onPressReply}: {onPressReply: () => unknown}) {
   const safeAreaInsets = useSafeAreaInsets()
   const fabMinimalShellTransform = useMinimalShellFabTransform()
+  console.log()
```
mozzius, line 541: Log

mozzius (approved):

Issue #6803, opened 2024-11-28
Language picker doesn't respect device dark mode


skipness:
### Steps to Reproduce

1. go to splash screen / sign up / language setting page / Search
2. trigger the language picker

### Attachments
| Splash Screen  | Sign Up Page  | Languages Setting Page  | Search |
|---|---|---|---|
| <img src='https://github.com/user-attachments/assets/661c257f-7e82-41ec-8699-6bbd7d3b6d6c' width = 450 />  | <img src='https://github.com/user-attachments/assets/bdce127e-20a9-4505-b0d5-f56a746c5f3d' width = 450 />  | <img src='https://github.com/user-attachments/assets/25724447-2c83-4c35-bbee-be0c723f1b64' width = 450 />  | <img src='https://github.com/user-attachments/assets/52793d8b-0620-46f8-96f9-d1621aad7ae0' width = 450 />  | 

### What platform(s) does this occur on?

iOS

### Device Info

iPhone 14 Pro

### What version of the app are you using?

Build version: 1.95.0.634; Bundle info: 2a575ad (prod); Bundle date: 24112703; Platform: ios; Platform version: 18.0

skipness:
I can submit a PR to fix this if needed

auroursa:
> I can submit a PR to fix this if needed

I think this is a reasonable improvement, please do so

SilverPlate9938:
Same issue with Android as well![Screenshot_2024-11-29-14-21-47-94_cb2df8437d99d85560b8f74042fc78eb](https://github.com/user-attachments/assets/2dbe5337-cd79-4ebb-a9b3-b4417d505dea)

PR #6804, opened 2024-11-28
Fix language picker style in dark mode

skipness:
Fixes: #6803 
I don't have any Android devices nor can successfully test it with emulator so the picker in Android should still in wrong style under dark mode

### Before
| Splash Screen  | Sign Up Page | Languages Setting Page  | Search |
|---|---|---|---|
|   |   |   |   | 

### After
| Splash Screen  | Sign Up Page | Languages Setting Page  | Search |
|---|---|---|---|
|  |  |  |  |


mozzius (approved):
lgtm, thanks! doesn't affect android because that prop is iOS only (and that's fine)

Issue #6800, opened 2024-11-28
Recent regression in display of image embeds without pre-existing dimensions data


eramdam:
### Steps to Reproduce

Taking this post as an example 

https://bsky.app/profile/macthemes.bsky.social/post/3lbxkfazqdx2f

It showed "as expected" on 1.94 

![CleanShot 2024-11-27 at 15 48 19@2x](https://github.com/user-attachments/assets/1fa42995-ff92-444a-bc33-70b0940349e0)

But shows up with `objectFit: cover` on 1.95, which obstructs a bunch of the image 

![CleanShot 2024-11-27 at 15 49 29@2x](https://github.com/user-attachments/assets/ba8bb282-44dc-452e-b812-e8616afc658f)

I haven't checked if the issue is Web-specific because I don't have the tooling.

It seems like the regression was caused by the fallback square aspect ratio introduced in https://github.com/bluesky-social/social-app/pull/6474 

Which makes sense! Layout shifts suck, and it's an understandable fix, I just feel that retroactively changing the way images are displayed isn't great either?   
My bot has been posting images for way longer than the `aspectRatio` field has been available on `app.bsky.embed.images` records so it feels "unfair" 😛 

I fixed my posting code so it won't happen for future records, but I wonder if a middle ground "fix" would be to:
1. assume the aspect ratio is square
2. use the fetched dimensions to resize the image _inside_ the square aspect ratio? 

It might cause a slight flash/jump, but it's better than a layout shift and more importantly, wouldn't retroactively "hide" content for images that lacked aspect ratio data and displayed fine until 1.95

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### What version of the app are you using?

1.95

12Me21:
It wouldn't cause any flash/jump at all, just need to set css `object-fit` to `contain` (instead of `cover`), so all images with an incorrect/unknown aspect ratio will be scaled down instead of cropped

GlitchyPSIX:
> It wouldn't cause any flash/jump at all, just need to set css `object-fit` to `contain` (instead of `cover`), so all images with an incorrect/unknown aspect ratio will be scaled down instead of cropped

I think that would make images with aspect ratios > 1 get a _lot_ of unwarranted empty space though

12Me21:
[...]

PR #6828, opened 2024-11-28
"Contain" images with missing dimensions instead of cropping them

gaearon:
Addresses https://github.com/bluesky-social/social-app/issues/6800

As noted in https://github.com/bluesky-social/social-app/pull/6474, we don't want to use dynamically fetched dimensions on the client because it causes layout shifts for any post(s) after the image. In the worst case this can cause an entire cascade of relayouts which is absolutely terrible for performance.

However, as discussed in https://github.com/bluesky-social/social-app/issues/6800, cropping can feel pretty disruptive. Instead, let's try "containing" when the aspect ratio is unknown.

This only affects images with missing dimensions. (Currently, it's up to the client to specify the dimensions.) In longer term it would be nice if the server could somehow fill out the missing ones, but fixing that is out of scope of this PR.

I had to disable the lightbox animation for such images because the initial sizing is going to be different in the "contains" layout. While disabling the animation, I noticed flashes on first and last frames, which the two last commits take care of.

## Test Plan

Find an account with images that are missing dimensions, for example https://bsky.app/profile/wrtrstat.bsky.social

Verify that images with unknown aspect ratio are now "contained" instead of cropped.

Verify that opening them in the lightbox works, but skips the expanding animation.

Verify that opening images _with_ dimensions remains fluid and animated.

Test all platforms.

## Before

Screenshot 2024-11-28 at 17 36 26

Screenshot 2024-11-28 at 17 36 33

## After

Screenshot 2024-11-28 at 17 36 11
Screenshot 2024-11-28 at 17 36 17

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageItem/ImageItem.android.tsx
@@ -392,9 +392,9 @@ const ImageItem = ({
           <Animated.View style={imageCropStyle}>
             <Animated.View style={imageStyle}>
               <Image
-                contentFit="cover"
+                contentFit="contain"
                 source={{uri: imageSrc.uri}}
-                placeholderContentFit="cover"
+                placeholderContentFit="contain"
                 placeholder={{uri: imageSrc.thumbUri}}
                 accessibilityLabel={imageSrc.alt}
                 onLoad={
--- a/src/view/com/lightbox/ImageViewing/index.tsx
+++ b/src/view/com/lightbox/ImageViewing/index.tsx
@@ -8,7 +8,7 @@
 // Original code copied and simplified from the link below as the codebase is currently not maintained:
 // https://github.com/jobtoday/react-native-image-viewing
 
-import React, {useCallback, useEffect, useState} from 'react'
+import React, {useCallback, useEffect, useMemo, useState} from 'react'
 import {
   LayoutAnimation,
   PixelRatio,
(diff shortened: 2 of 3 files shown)


haileyok (approved):
Tested web and Android, works well.

WriterStat:
Hi, I like you all.

It works better than the previous system breaking fix, but it is not a good fix long term for lack of mandatory field communication. Temp fix at best. The changes from the last update need to be rolled back to the prior original working code before the breaking changes were made for now instead. And mandatory fields need to be communicated to your 3rd party and client post submission ecosystem, even if only in rule, then the code itself will basically age out fast, will no longer be used for any new posts. Problem goes away by itself.

Change instead needs to happen on image upload api/fields required for your client post submission ecosystem.

See the new posts on #6800

The previous breaking change made before this one made my page on Bluesky basically unusable, I wasn't the only one, with 20 million. This one will allow it to be read, but will make Bluesky less visually usable/likeable, a longer process to read/see and longer pages to interact with and you will be stuck with it. Make the change in image submission/image upload instead to fill the mandatory fields needed to work. That fixes your original problem.

Or get/calculate the data needed and store it on the first image request call, so it doesn't need to run the hack code more than one time. Also fixes your original problem.

ChrisArchitect:
Rolling this back yet? Going to squares with css contain breaking a long held standard for opengraph preview image dimensions to fix 'layout shift' or whatever is not a good look. And it not being emphasized in the API docs/examples about the aspectratio parameter means many many things have been thrown out of wack.

WriterStat:
The image aspect ratio can now be set/sent with send_image and send_images. The at-proto team is doing/did their thing. 

View: https://github.com/MarshalX/atproto/pull/480 

The breaking stuff should now be able to be rolled back to its first original working code, before the image/system breaking code was applied.

After rollback, all that would be needed would be to effectively communicate the mandatory API changes for image aspect ratios to your 3rd party and client post submission ecosystem for Bluesky.

Making the image aspect ratio mandatory(at least by rule/examples) during post submission/image upload like it is on the current bluesky website post button code should solve the original layout problem for new.

The layout shift problem will go away as the old code causing it will then age out because no new post/image submissions will need it after the ratio field is mandatory. Once the image aspect ratio is included on post submissions/uploads in the API.

Did I miss something? Again I like you and the platform, quite a lot. And want you to succeed. If I'm wrong let me know. Fixing the fields should fix the possible future layout issue, right? Or no?

For old posts image ratio, if really needed couldn't you calculate and then store it on first fetch? So it only used the old code only once at worst?

PR #6835, opened 2024-11-28
Add new languages to `CFBundleLocalizations` key

### LINKED ISSUE #6162: Add Localizations for Context Menus in TextInput(Copy, Paste, AutoFill, etc) in CFBundleLocalizations for iOS

surfdude29:
#6567 added the languages supported as `AppLanguage` in a `CFBundleLocalizations` key to `Info.plist`. This meant that with the release of 1.95, the supported languages now show up in the App Store listing, rather than just English being shown.

It also meant that #6162 was resolved and allowed the `TextInput` context menus to be translated into the device language (if it's supported as an `AppLanguage`).

| 1.94 | 1.95 |
|--------|--------|
| IMG_1134 | IMG_4117 |
| IMG_4120 | IMG_4124 |

This PR adds the codes for the localizations merged since #6567 – Aragonese, Asturian, Galician, Dutch and Vietnamese – to the `CFBundleLocalizations` key so that they'll also be displayed in the App Store listing and users with their device set to Dutch or Vietnamese will have the `TextInput` context menus translated. In future, the `code2` for a new language can simply be added as part of a localization PR.

This PR also makes a couple of minor, unrelated tweaks to keep language codes in alphabetical order and to replace a couple of hyphens with en-dashes in the list for selecting an app language.

**EDIT:** I also made some changes to the Chinese & Cantonese language tags following @auroursa's suggestions, and I changed `pt` to `pt-BR` in anticipation of the [forthcoming `pt-PT` localization](https://github.com/bluesky-social/social-app/pull/6836).


auroursa:
If you could additionally help me adjust the Chinese part, that would be greatly appreciated.

For [various reasons](https://github.com/bluesky-social/social-app/pull/6754#issuecomment-2502534764), the language tags for Chinese are quite complex. I think both iOS and Android are more prefer `zh-Hans` / `zh-Hant`. So could you help me change `zh-CN` to `zh-Hans` and `zh-TW` to `zh-Hant`? I’ll include reference links below.

`zh-HK` doesn’t seem to changes, as it refers to a specific language, rather than a language scope. It’s also quite frustrating that version 1.95.0 in the App Store doesn’t list Bluesky as supporting Cantonese. This currently seems to have no solution. The only thing I can do is make it fallback to Traditional Chinese instead of English whenever possible.

Reference:
https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html

https://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html

surfdude29:
> So could you help me change `zh-CN` to `zh-Hans` and `zh-TW` to `zh-Hant`?

Done ✅

> `zh-HK` doesn’t seem to changes, as it refers to a specific language, rather than a language scope. It’s also quite frustrating that version 1.95.0 in the App Store doesn’t list Bluesky as supporting Cantonese. This currently seems to have no solution. The only thing I can do is make it fallback to Traditional Chinese instead of English whenever possible.

From the first reference:

> Only use a script designator in a locale ID when there is ambiguity. For example, because Traditional Chinese is the default language in Hong Kong, use `zh_HK`, where `zh` is the code for Traditional Chinese and `HK` is the code for the Hong Kong region.

I think this is why; `zh_HK` is being mapped to `zh-Hant_HK`.

After doing some more googling and coming across some other sources like [this GitHub comment](https://github.com/mozilla/addons/issues/4515#issuecomment-2094115676) ([updated CLDR Likely subtags link](https://unicode.org/cldr/charts/46/supplemental/likely_subtags.html)) and seeing that these are the entries listed for Cantonese on my iPhone:

IMG_4148

what about changing the `zh_HK` entry to `yue-Hant` and seeing if that allows Cantonese to show up in the App Store listing?

auroursa:
> what about changing the zh_HK entry to yue-Hant and seeing if that allows Cantonese to show up in the App Store listing?

It's worth trying. 👍

I'm not sure if defining it directly as `yue` would have better compatibility, according to Apple's documentation, it is an allowed language code.

Also made changes related to Cantonese in #6621 to ensure that both `yue-Hant` and `yue-Hans` would be redirected to `zh-HK` before. (Yes, our localization team unanimously agrees that we should disregard the simplified form of Cantonese and redirect it to the traditional form haha.

surfdude29:
> > what about changing the zh_HK entry to yue-Hant and seeing if that allows Cantonese to show up in the App Store listing?
> 
> It's worth trying. 👍

Done ✅🤞

> I'm not sure if defining it directly as `yue` would have better compatibility, according to Apple's documentation, it is an allowed language code.

I went with `yue-Hant` for now and let's see what happens. If that doesn't work then we can try just `yue`.

> Also I before made changes related to Cantonese in #6621 to ensure that both `yue-Hant` and `yue-Hans` would be redirected to `zh-HK`. (Yes, our localization team unanimously agrees that we should disregard the simplified form of Cantonese and redirect it to the traditional form haha.

Oh yes I remember seeing that PR. Maintaining Cantonese localizations in both scripts might be a bit much, so probably the best solution 😅

auroursa:
OK let's see the results after the next App Store update. Thanks again!

gaearon (approved):

PR #6836, opened 2024-11-29
Add Portuguese-PT locale/translations support

joaomendoncaa:
wip  🫡 

image


src/locale/languages.ts
```diff
@@ -464,7 +466,8 @@ export const LANGUAGES: Language[] = [
   {code3: 'pli', code2: 'pi', name: 'Pali'},
   {code3: 'pol', code2: 'pl', name: 'Polish'},
   {code3: 'pon', code2: '', name: 'Pohnpeian'},
-  {code3: 'por', code2: 'pt', name: 'Portuguese'},
+  {code3: 'por', code2: 'pt-BR', name: 'Portuguese (BR)'},
```
auroursa, line 469: Please do not modify this part, as it refers to the post language, not the app display language. currently, post language does not support regional distinctions.

auroursa:
Just a personal suggestion here (I’m not a native Portuguese speaker), for compatibility reasons, I think using `pt` instead of `pt-PT` as the language code might be better. This would provide a fallback option for older devices that don’t support `pt-PT` and `pt-BR`.

There’s also a practical reason: as part of the localization, we’d need to modify `src\components\hooks\dates.ts` and add Portuguese (pt-PT) there. But [date-fns](https://www.npmjs.com/package/date-fns) doesn’t support `pt-PT`, only `pt` here, which might cause some complications.

src/components/hooks/dates.ts
```diff
@@ -63,6 +64,7 @@ const locales: Record<AppLanguage, Locale | undefined> = {
   nl,
   pl,
   ['pt-BR']: ptBR,
+  ['pt']: pt,
```
auroursa, line 67:
```suggestion
  pt,
```

src/locale/helpers.ts
```diff
@@ -157,6 +157,8 @@ export function sanitizeAppLanguageSetting(appLanguage: string): AppLanguage {
         return AppLanguage.pl
       case 'pt-BR':
         return AppLanguage.pt_BR
+      case 'pt-PT':
```
auroursa, line 160:
```suggestion
      case 'pt':
```

src/locale/i18n.ts
```diff
@@ -31,6 +31,7 @@ import {messages as messagesKo} from '#/locale/locales/ko/messages'
 import {messages as messagesNl} from '#/locale/locales/nl/messages'
 import {messages as messagesPl} from '#/locale/locales/pl/messages'
 import {messages as messagesPt_BR} from '#/locale/locales/pt-BR/messages'
+import {messages as messagesPt_PT} from '#/locale/locales/pt-PT/messages'
```
auroursa, line 34:
```suggestion
import {messages as messagesPt} from '#/locale/locales/pt/messages'
```

src/locale/i18n.ts
```diff
@@ -190,6 +191,14 @@ export async function dynamicActivate(locale: AppLanguage) {
       ])
       break
     }
+    case AppLanguage.pt: {
+      i18n.loadAndActivate({locale, messages: messagesPt_PT})
```
auroursa, line 195:
```suggestion
      i18n.loadAndActivate({locale, messages: messagesPt})
```

src/locale/i18n.web.ts
```diff
@@ -88,6 +88,10 @@ export async function dynamicActivate(locale: AppLanguage) {
       mod = await import(`./locales/pt-BR/messages`)
       break
     }
+    case AppLanguage.pt: {
+      mod = await import(`./locales/pt-PT/messages`)
```
auroursa, line 92:
```suggestion
      mod = await import(`./locales/pt/messages`)
```

auroursa (commented):
Let's update the remaining `pt-PT` (but you can still keep "Português (PT)" in the language list).

auroursa:
> @auroursa there's a lot of people involved in this translation, if I edit the top of `pt-PT/messages.po` to add a reference to all the people who are involved (but not necessarily commiting directly to the tree); will it be overridden at build time by the lingui cli?

Yes, it won’t be overwritten by the lingui cli.

It looks like you’ve updated all the language files. Please specify the language explicitly when running `intl:extract`:
`yarn intl:extract --locale pl`

auroursa:
Apologies, I forgot this step. You’ll also need to rename the `src\locale\locales\pt-PT` folder to `src\locale\locales\pt`.

surfdude29:
I actually think it would be better to mostly specify `pt-PT`. I may be understanding it wrong, but from what I've read the default region for the `pt` language code is `BR`, not `PT`, so just using `pt` rather than `pt-PT` might cause issues?

I read this somewhere else (can't remember where unfortunately) but one place that definitely says it is the [Unicode CLDR Likely Subtags list](https://unicode.org/cldr/charts/46/supplemental/likely_subtags.html).

auroursa:
> I actually think it would be better to mostly specify `pt-PT`. I may be understanding it wrong, but from what I've read the default region for the `pt` language code is `BR`, not `PT`, so just using `pt` rather than `pt-PT` might cause issues?
> 
> 
> 
> I read this somewhere else (can't remember where unfortunately) but one place that definitely says it is the [Unicode CLDR Likely Subtags list](https://unicode.org/cldr/charts/46/supplemental/likely_subtags.html).
> 
> 
> 
> edit: It's stated more explicitly [here](https://cldr.unicode.org/translation/translation-guide-general/default-content):
> 
> 
> 
> > Portuguese (Brazil) [pt_BR] is the default content locale for Portuguese [pt]

I am concerned about compatibility with some browsers and older devices. Their language list only includes Portuguese, so it will output `pt` by default. I’m not sure if the current AppLanguage matching mechanism can handle this situation. If it doesn’t match any regional suffix, it might fall back to English.

I noticed that the Portuguese government website uses both `pt` and `pt-PT`, but you are right. I still need to investigate further.

auroursa:
No, I think in web development, we should use `pt` instead of `pt_PT` to represent Portuguese (PT), because the module we use to localize dates [date-fns)(https://www.npmjs.com/package/date-fns) doesn’t have the `pt_PT` option, and it will cause the date format fallback to English.

surfdude29:
> No, I think in web development, we should use `pt` instead of `pt_PT` to represent Portuguese (PT), because the module we use to localize dates [date-fns](https://www.npmjs.com/package/date-fns) doesn’t have the `pt_PT` option, and it will cause the date format fallback to English.

I'm not sure that should be a particular issue, for `date-fns` we can just use `pt` while using `pt-PT` elsewhere :)

auroursa:
> > No, I think in web development, we should use `pt` instead of `pt_PT` to represent Portuguese (PT), because the module we use to localize dates [date-fns](https://www.npmjs.com/package/date-fns) doesn’t have the `pt_PT` option, and it will cause the date format fallback to English.
> 
> 
> 
> I'm not sure that should be a particular issue, for `date-fns` we can just use `pt` while using `pt-PT` elsewhere :)

Please give me some time to test it. I want to compile it on different platforms and check the results.

auroursa:
@joaomendoncaa May I ask about your browser language list and mobile language list settings? I want to make sure my settings are correct.

surfdude29:
> Please give me some time to test it. I want to compile it on different platforms and check the results.

Of course, there's no rush :)

I may well be wrong and `pt` on its own might be the better option, but my gut feeling from all the localization things I've read is that it will generally work better to include a region tag to distinguish between two locales with the same language. (It doesn't apply in the case of English of course because `en-US` is the universal default 😅)

auroursa:
If I’m mistaken, please forgive me, but I assume you have a local build environment. If you’re willing, could you test both pt and pt-PT under your current browser settings to see if Bluesky automatically switches to Portuguese on the first load?

It seems to work both on my end, but I’m not sure how it behaves in other browsers (I’m using Firefox).

auroursa:
I tested `pt` on both a Galaxy S22 and an Android emulator. When the system language is set to Portuguese (Portugal), it switches to the correct language (as shown in the video), both browsers and Android seem to redirect `pt` to `pt-PT` rather than `pt-BR`.

https://github.com/user-attachments/assets/08c42feb-8eea-4460-8488-07e63e960158

If it also works correctly in your browser tests, I think this comes down to a matter of personal preference. Either `pt` or `pt-PT` should be fine.

auroursa:
Okay let’s reset to bf430e9 because the current code is a bit messy…

As you mentioned, we should add a case for `pt` to `pt-BR` to ensure compatibility with older devices in `helpers.ts`. Additionally, we need to make some adjustments in `src/components/hooks/dates.fs` to allow `pt-PT` to properly utilize `date-fns`.

dates.fs:
```typescript
import {
  …
  pt,
  ptBR,
  …
} from 'date-fns/locale'

const locales: Record<AppLanguage, Locale | undefined> = {
  …
  pt: pt,
  ['pt-PT']: pt,
  ['pt-BR']: ptBR,
  …
}
```

surfdude29:
Would it be possible to add one more change to this PR? If you could add a `'pt-PT',` entry to `app.config.js` here please, that would be great:
https://github.com/bluesky-social/social-app/blob/770eeb59ca583c9bbc121790e0393195eda6b796/app.config.js#L115-L116

**Portuguese** is already displayed as a supported language in the App Store listing so it won't make any changes there. However, because both **Portuguese (Brazil)** and **Portuguese (Portugal)** are listed as iPhone Languages:

IMG_5023

I feel like it's probably best to have specific, separate entries for them both in the `CFBundleLocalizations` key rather than just one `pt` entry, to hopefully ensure that the `TextInput` context menus are translated properly for iOS users who have either locale set as their device language :)

src/locale/i18n.ts
```diff
@@ -198,6 +199,14 @@ export async function dynamicActivate(locale: AppLanguage) {
       ])
       break
     }
+    case AppLanguage.pt_PT: {
+      i18n.loadAndActivate({locale, messages: messagesPt_PT})
+      await Promise.all([
+        import('@formatjs/intl-pluralrules/locale-data/pt-PT'),
```
surfdude29, line 205:
```suggestion
        import('@formatjs/intl-datetimeformat/locale-data/pt-PT'),
        import('@formatjs/intl-pluralrules/locale-data/pt-PT'),
```

Following #6742 being merged, could you add this line please?

This will ensure that the date and time for posts are localized correctly on all platforms. Thanks!

Issue #6850, opened 2024-11-29
Duplicate profiles appearing in Starter Pack user selection


Mik1337:
### Steps to Reproduce

1. Create a new starter pack from the profile page
2. Create a fresh pack, add a name and proceed to the next screen
3. Scroll Down, you will notice duplicates

### Attachments

**iPhone 12**
https://github.com/user-attachments/assets/7c31fb19-c59f-4895-a05b-d6deb1f9c851
This issue seemed to be more pronounced on my phone.

**Firefox on Macbook**
https://github.com/user-attachments/assets/422bc479-f116-419f-947b-e6ffca40c594

### What platform(s) does this occur on?

iOS, Web (Desktop)

### Device Info

iPhone 12, Firefox

### What version of the app are you using?

Build version: 1.94.0.602; Bundle info: 3c30bb1 (prod); Bundle date: 24111623; Platform: ios; Platform version: 18.1.1

### Additional Information

I ran the project locally to see what was up, and i got these error logs

https://github.com/user-attachments/assets/f919a22b-d0dd-467d-b03f-999d4148453b

PR #6851, opened 2024-11-29
FIxes an issue where Duplicate profiles appear in Starter Pack user selection

Mik1337:
Seems to solve the issue...
https://github.com/user-attachments/assets/65812e5d-94fd-4fcb-b949-0f3d0b063425

PS: not sure if this should be handled on the clients device, but it's a quick fix.

### DIFF
--- a/src/screens/StarterPack/Wizard/StepProfiles.tsx
+++ b/src/screens/StarterPack/Wizard/StepProfiles.tsx
@@ -1,4 +1,4 @@
-import {useState} from 'react'
+import {useMemo,useState} from 'react'
 import {ListRenderItemInfo, View} from 'react-native'
 import {KeyboardAwareScrollView} from 'react-native-keyboard-controller'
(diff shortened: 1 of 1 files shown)


Mik1337:
fixed the issue I've created here

https://github.com/bluesky-social/social-app/issues/6850

src/screens/StarterPack/Wizard/StepProfiles.tsx
```diff
@@ -45,8 +45,9 @@ export function StepProfiles({
     const seen = new Set<string>()
     return topPages.pages
       .flatMap(p => p.actors)
+      .filter(p => !p.associated?.labeler)
       .filter(p => {
-        if (!p.associated?.labeler && !seen.has(p.did)) {
+        if (!seen.has(p.did)) {
```
Mik1337, line 50: i broke this out in a separate filter 'cause i don't really know what the `labeler` is and didn't want to break anything

PR #6857, opened 2024-11-29
Update Aragonese (an) localization

Juanpabl:
Update messages. L10n completed at https://localazy.com/p/bsky


pfrazee (commented):
Great, thanks!

pfrazee (approved):

PR #6866, opened 2024-11-30
Developer mode toggle

m-doescode:
Adds a developer mode toggle in Settings > About which lets you copy the at uri of any post/profile just like you could on discord

image

(Note that I am fairly new to contributions so please let me know if I've missed anything such as create any locale strings or if I've totally broken anything. Thank you!)

### DIFF
--- a/src/screens/Settings/AboutSettings.tsx
+++ b/src/screens/Settings/AboutSettings.tsx
@@ -7,17 +7,22 @@ import {NativeStackScreenProps} from '@react-navigation/native-stack'
 import {appVersion, BUNDLE_DATE, bundleInfo} from '#/lib/app-info'
 import {STATUS_PAGE_URL} from '#/lib/constants'
 import {CommonNavigatorParams} from '#/lib/routes/types'
+import {useDeveloperModeEnabled, useSetDeveloperModeEnabled} from '#/state/preferences'
 import * as Toast from '#/view/com/util/Toast'
 import * as SettingsList from '#/screens/Settings/components/SettingsList'
+import * as Toggle from '#/components/forms/Toggle'
 import {CodeLines_Stroke2_Corner2_Rounded as CodeLinesIcon} from '#/components/icons/CodeLines'
 import {Globe_Stroke2_Corner0_Rounded as GlobeIcon} from '#/components/icons/Globe'
 import {Newspaper_Stroke2_Corner2_Rounded as NewspaperIcon} from '#/components/icons/Newspaper'
 import {Wrench_Stroke2_Corner2_Rounded as WrenchIcon} from '#/components/icons/Wrench'
+import {Atom_Stroke2_Corner0_Rounded as AtomIcon} from '#/components/icons/Atom'
 import * as Layout from '#/components/Layout'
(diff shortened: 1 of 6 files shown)


DavidBuchanan314:
I think it makes the most sense to have the profile option just be "copy DID", whereas the menu on individual posts can have "Copy ATProto URI" (could be both, I suppose, but then the menus might get a bit crowded)

m-doescode:
> I think it makes the most sense to have the profile option just be "copy DID", whereas the menu on individual posts can have "Copy ATProto URI" (could be both, I suppose, but then the menus might get a bit crowded)

That's a good idea... There already isn't a profile.uri, i had to concatenate to produce an at:// uri. Plus DIDs are much more commonly needed than the at:// schema uri when it comes to accounts.

auroursa:
This seems like a nice feature, but I’d like to point out that we already have developer options. Currently, it requires setting an environment variable to appear.

Maybe you could explore making it easier to enable and integrating the "Copy DID" functionality into it? Additionally, I think Developer Options shouldn’t be too easy to enable—perhaps we could use a more complex method, like tapping the version number 10 times.

robpc:
> Additionally, I think Developer Options shouldn’t be too easy to enable—perhaps we could use a more complex method, like tapping the version number 10 times.

That would make sense for internal developer settings, but I think the intent of this PR is for the external hobby developers and in that case I like that it is accessible. 

If that seems confusing how about a different name for this setting? "Expose Atproto Details" might be more descriptive?

auroursa:
> If that seems confusing how about a different name for this setting? "Expose Atproto Details" might be more descriptive?

I believe this is a positive step toward improvement, as having two similar developer modes might lead to some confusion.

DavidBuchanan314:
> This seems like a nice feature, but I’d like to point out that we already have developer options. Currently, it requires setting an environment variable to appear.

IIUC, these are two different kinds of developer settings. The one you describe is for developers of bsky.app itself, whereas the one in this PR is for atproto developers at large. I think it makes sense to keep these as two different things, but maybe we should think about naming them less ambiguously. Perhaps "ATProto Developer Mode" 

I also don't think there's any need to "hide" them, developer features should be easily accessible for anyone who wants to use them. Android hides its developer mode because you can harm your device if you don't know what you're doing, but nobody will be harmed by copying an at:// URI.

auroursa:
I admit this is a meaningful new feature. My concern just that, since we already have a similar developer options here, it might lead to some confusion. That said, I completely agree with your point—it might just need a clearer name.

> I also don't think there's any need to "hide" them, developer features should be easily accessible for anyone who wants to use them.

Yeah I don’t think it should be hidden actually. My original idea was to integrate this feature into the existing developer options, considering the existing features there, it needs to be properly hidden. But if it’s just about copying at://URLs, it likely doesn’t need to be hidden at all.

bleonard252:
Could be "Power User Mode," containing more advanced things that most people don't need to use, mostly copying DIDs or AT URIs. (It could also be nice to have an integrated repo record browser/editor, or PLC editor, but that's another story.)

PR #6868, opened 2024-11-30
[Nicer Tabs] New native pager

gaearon:
This revives https://github.com/bluesky-social/social-app/pull/1696 but in a performant way.

## Test Plan

Try different interactions on iOS and Android:

- **Initial load:** The pager should autoscroll to the initially selected tab on reload. (Ideally this would not be animated, but since I couldn't figure out a reliable way to do it on the first frame on both platforms, I have to animate it.)
- **Swiping the pager drives tabbar scroll:** The swiping gesture should stay in sync with the tabbar scroll. A full page swipe should bring you to the same position as tapping the next item. Note that if you continuously swipe page after page, by the time you arrive at the last page, the blue line should naturally be at the right edge of the screen. This also means that by the time you're halfway through the feeds, the blue line should be _exactly_ centered in the middle. This is easy to see if you have an odd number of feeds.
- **Tapping predictably aligns the tabbar:** Tapping to select an item will adjust the tabbar to its "synced" position, same as a swipe. There is one exception to this, which I'll cover next.
- **Scrolling the tabbar itself temporarily unsyncs it:** If you start scrolling the tabbar and *then* tap on an item there, if that item is fully visible in view (with a margin of 20 on each side), that tap will _not_ change the tabbar scroll position. This is to avoid things shifting from "right under you" on tap. However, the *next* tap will be synced as usual (unless you scroll again). If the item was _not_ fully visible into view (with a margin of 20 on each side), we always sync.
- **Unsyncing, then swiping:** You can unsync the tabbar (by scrolling it) and _then_ swipe. In this case, we won't mess with the tabbar scroll position *during* the swipe — but it'll get re-sync immediately after you release the gesture.
- **Unsyncing, then tapping, then swiping:** Unsync the tabbar (by scrolling it as far as you want). Then tap on some item that's fully in view. The item should get selected but the tabbar shouldn't scroll "under" you. _Now_ start swiping. Notice that the gesture will not be jumpy — it'll gradually move as always but then _resync after you release_.

Verify non-primary cases:

- Just one tab: Check a feed's page
- Just two tabs: Delete feeds other than Following and Discover
- The profile: Verify profile tabs behave as expected
- Web: Should have no changes

Verify perf on Android and iOS devices.

Known issues:

- <s>Due to the change in https://github.com/bluesky-social/social-app/pull/6762, there's no prewarming during swipe, so each feed loads a bit "too late" (after swipe). I'll stack something on top of this PR to resolve that, but not in scope.</s>
  - Actually, this doesn't seem very relevant — we were only starting load when we got into a "settling" state so it was pretty late anyway. I'll look at improving this regardless but doesn't block land.

- On Android, swipes are sometimes interpreted as taps and bring up the lightbox or navigate into the post. This is not a new issue, and has always been broken — I'll likely look at it but not today.

https://github.com/user-attachments/assets/8a941234-fc98-4982-b652-9874151be3a5

### DIFF
--- a/src/view/com/pager/PagerWithHeader.web.tsx
+++ b/src/view/com/pager/PagerWithHeader.web.tsx
@@ -151,6 +151,8 @@ let PagerTabBar = ({
           selectedPage={currentPage}
           onSelect={onSelect}
           onPressSelected={onCurrentPageSelected}
+          dragProgress={undefined as any /* native-only */}
+          dragState={undefined as any /* native-only */}
         />
       </View>
     </>
--- a/src/view/com/home/HomeHeader.tsx
+++ b/src/view/com/home/HomeHeader.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import {useNavigation} from '@react-navigation/native'
 
-import {usePalette} from '#/lib/hooks/usePalette'
 import {NavigationProp} from '#/lib/routes/types'
 import {FeedSourceInfo} from '#/state/queries/feed'
 import {useSession} from '#/state/session'
@@ -19,7 +18,6 @@ export function HomeHeader(
   const {feeds} = props
   const {hasSession} = useSession()
   const navigation = useNavigation<NavigationProp>()
-  const pal = usePalette('default')
 
   const hasPinnedCustom = React.useMemo<boolean>(() => {
     if (!hasSession) return false
@@ -61,7 +59,6 @@ export function HomeHeader(
         onSelect={onSelect}
         testID={props.testID}
         items={items}
-        indicatorColor={pal.colors.link}
         dragGesture={props.dragGesture}
       />
     </HomeHeaderLayout>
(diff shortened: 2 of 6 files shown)


src/view/com/pager/PagerWithHeader.web.tsx
```diff
@@ -151,6 +151,8 @@ let PagerTabBar = ({
           selectedPage={currentPage}
           onSelect={onSelect}
           onPressSelected={onCurrentPageSelected}
+          dragProgress={undefined as any /* native-only */}
+          dragState={undefined as any /* native-only */}
```
mozzius, line 155: I wish there was a way to express platforms in the type system :/

mozzius: yeah, some sort of monad perhaps

haileyok (approved):
Tested for a full day and worked fine. Code lgtm

PR #6620, opened 2024-11-22
Update Japanese translation

tkusano:
@Hima-Zinn  @dolciss @middlingphys @hibiki , review please

- remove obsolete messages for font related nux
- translate some error messages

PR #6869, opened 2024-11-30
Update Hungarian translation to 1.95

smileyhead:
This one was a bit later than I wanted it to be – apologies.

Before the efforts get formalised through a service like Crowdin, could we set up a system where translators get notified about new strings *before* a build gets made? Pretty please?

(I have asked @SonoSooS to proofread.)


surfdude29:
> ... could we set up a system where translators get notified about new strings _before_ a build gets made? Pretty please?

Unfortunately this hasn't happened yet, but informally you can follow other translation PRs as they get updated to include new strings.

E.g. I think this should include all the new strings added since you created this PR? 🤞

https://github.com/bluesky-social/social-app/pull/6620/files/b5fd660fc179b0a0262c8f37382c2863fe89e10c..da39f90605f83a23d767709a4210bd7fa6ad5232

smileyhead:
Unfortunately, I will not be home for two more days, but I do plan on updating the translation again then. Thanks for the heads-up.

I've tried ‘updating’ the fork to hopefully match the main branch apart from my edits I hope I did it correctly.

pfrazee (approved):
Awesome, thanks!

PR #6873, opened 2024-11-30
feat: animate tab indicator position and size on native and web

AronBe:
I tried to make least invasive implementation and just basically extended the logic that was already there. Should work on all platforms including web. Feel free to get inspired for your implementation #6868 (which is very similar in base approach) or if you want, I can extend this one with your suggestions

https://github.com/user-attachments/assets/b14b4db7-9827-40b5-b87e-477c295a49ab

### DIFF
--- a/src/view/com/home/HomeHeader.tsx
+++ b/src/view/com/home/HomeHeader.tsx
@@ -16,7 +16,7 @@ export function HomeHeader(
     feeds: FeedSourceInfo[]
   },
 ) {
-  const {feeds} = props
+  const {feeds, pageOffset} = props
   const {hasSession} = useSession()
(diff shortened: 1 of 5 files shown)


gaearon:
oh nice, thanks for the PR! mine is a bit more extensive because i'm also synchronizing it to the ongoing gesture so i think i'd go with the approach in my PR. there's probably many other opportunities to contribute similar polish though.

gaearon:
for the web, we could do something — i'm explicitly forked the file to make it easier to play with. i'm personally in favor of avoiding Reanimated on the web as much as possible, but if we can do something nice with CSS, i'm down.

AronBe:
ok, understood, thanks for the great work on Bluesky 👍

PR #6887, opened 2024-12-02
clean rn 0.76 upgrade

haileyok:
## Why

Upgrade RN...made a clean PR because we've merged a lot and there was some yarn lock jank in the other PR.

## Test Plan

Use it use it use it


package.json
```diff
@@ -167,32 +167,32 @@
...
-    "react-native-reanimated": "^3.16.1",
+    "react-native-qrcode-styled": "^0.3.3",
+    "react-native-reanimated": "^3.16.3",
     "react-native-root-siblings": "^4.1.1",
-    "react-native-safe-area-context": "4.10.1",
-    "react-native-screens": "~3.31.1",
-    "react-native-svg": "^15.3.0",
-    "react-native-uitextview": "^1.3.0",
+    "react-native-safe-area-context": "4.14.0",
+    "react-native-screens": "~4.3.0",
```
WoLewicki, line 187: Hmm that is quite interesting that bumping to `react-native-screens` v4 works correctly for you if you use `native-stack`.  You are not using `@react-navigation` v7 and it is not recommended to bump one without the other: https://github.com/software-mansion/react-native-screens#usage-with-react-navigation.  But I mean, if it works, then 🤷‍♂️ 

What's more, `react-native-screens` v4 come with our own implementation of bottom-sheets connected to navigation: https://blog.swmansion.com/introducing-react-native-screens-4-0-0-1b833ff98a55. Maybe it would be good to research it too one day if you are unhappy with current bottom-sheet 😅

app.config.js
```diff
@@ -272,6 +262,18 @@ module.exports = function (config) {
             ],
           },
         ],
+        [
+          'expo-splash-screen',
+          PLATFORM === 'ios'
```
fobos531, line 267: @haileyok Worth noting that there are `android` and `ios` properties available in the config plugin (https://docs.expo.dev/versions/latest/sdk/splash-screen/#configurable-properties) - not sure, but maybe you can utilize this in place of the `PLATFORM` check

package.json
haileyok: Huh, yea this was actually a recommendation from the author!

https://bsky.app/profile/satya164.bsky.social/post/3lcawiutecs2g

Mostly I was just testing to see if it would work (and was sorta surprised it did!) Do you think it'd be best to just change this back?

app.config.js
haileyok: Perf, that makes sense :pinched_fingers:

package.json
haileyok: Also: https://bsky.app/profile/satya164.bsky.social/post/3lcaxcltl4m2j

haileyok: > What's more, react-native-screens v4 come with our own implementation of bottom-sheets connected to navigation: [blog.swmansion.com/introducing-react-native-screens-4-0-0-1b833ff98a55](https://blog.swmansion.com/introducing-react-native-screens-4-0-0-1b833ff98a55). Maybe it would be good to research it too one day if you are unhappy with current bottom-sheet 😅

Now _this_ is something i'd definitely be interested in trying!!

gaearon: i got wrecked by types there, would need some work to figure out what's going on

haileyok:
nice, i think this is ready now...lets run a build and test.

app.config.js
haileyok: hmm i guess i need to dig at what might be going on here, wondering if you have any context maybe.

seems like when i launch the app in dark mode now the splash is temporarily white instead of the proper color...do you see something obvious with the app config that is messed up?

package.json
WoLewicki: > Huh, yea this was actually a recommendation from the author!

Oh so probably `react-navigation` fixed all the problems with integration, nice!

WoLewicki: > Mostly I was just testing to see if it would work (and was sorta surprised it did!) Do you think it'd be best to just change this back?

I mean, if it works, then I think it's fine! I would just suggest bumping the navigation to v7 ASAP too. Then you'll be sure everything is up to date and supported.

app.config.js
haileyok: Ah no, it was present regardless of how I configured it (does look like the way you suggested works fine). I'll try again to be sure, but yea it was only on iOS. I suppose the fullscreen flag could be the problem, was hoping that would be "enough" for now, we'll see.

app.config.js
```diff
@@ -256,7 +251,7 @@ module.exports = function (config) {
         './plugins/starterPackAppClipExtension/withStarterPackAppClip.js',
         './plugins/withAndroidManifestPlugin.js',
         './plugins/withAndroidManifestFCMIconPlugin.js',
-        './plugins/withAndroidStylesWindowBackgroundPlugin.js',
+        // './plugins/withAndroidStylesWindowBackgroundPlugin.js',
```
gaearon, line 254: this was added in https://github.com/bluesky-social/social-app/pull/3105. is it no longer needed?

mozzius: from my understanding, this is now handled by the `expo-splash-screen` config plugin

gaearon (approved):

patches/react-native+0.76.3.patch.md
```diff
@@ -18,3 +18,7 @@ module.
 Patching `RCTBaseTextInputView.m` to fix an issue where `selectTextOnFocus` does not work as expected on iOS 17. This
 patch _only_ fixes the Paper version of `TextInput`. If we migrate to Fabric and the fix has not been made upstream,
 we can apply the same fix. See https://github.com/facebook/react-native/pull/44307.
```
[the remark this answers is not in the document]
gaearon: yeah

Issue #6888, opened 2024-12-02
Notification timestamp hover bubble can get cut off


maxfzz:
### Steps to Reproduce

1. Be followed by one person
2. Open Notifications
3. Hover over the follow notification's timestamp

### Attachments

<img width="543" alt="Screenshot 2024-12-01 at 22 18 07" src="https://github.com/user-attachments/assets/b02b69db-1303-4ff7-be59-b4d0cecae2f6">

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Build version: 1.95.0; Bundle info: c9d86de (prod); Bundle date: 24112719; Platform: web; Platform version: undefined

chloe-and-coffee:
It seems like this bug is an unintended consequence of this change: https://github.com/bluesky-social/social-app/pull/4899/files#diff-b386f6d1ddb83452d38afe308a2ec902dd0b0a07711264fa1ff4ae2bea171d51R258.  Post FeedItems don't get this issue because the timestamp is at the top of the post.  However, Notification FeedItems have the timestamps at the bottom.

PR #6890, opened 2024-12-02
6888 fix overflow issue for timestamp not being visible 

LavHinsu:
before:
image

after:
image

explanation: overflow was set to hidden in the FeedItem comp in notifications, this caused the timestamp in the notification item to be hidden. issue was described: in the https://github.com/bluesky-social/social-app/issues/6888


src/view/com/notifications/FeedItem.tsx
```diff
@@ -384,7 +384,7 @@ let FeedItem = ({
               borderColor: pal.colors.unreadNotifBorder,
             },
         {borderTopWidth: hideTopBorder ? 0 : StyleSheet.hairlineWidth},
-        a.overflow_hidden,
+        a.overflow_visible,
```
LavHinsu, line 387: this is not tested on mobile, fyi

PR #6896, opened 2024-12-02
Include avatar in RSS feed.

### LINKED ISSUE #3170: Broadcast User Avatar via RSS

I'm trying to set up webhooks for RSS, but I notice that unlike other services, bsky does not currently broadcast avatar image URLs via RSS, nor is there any clear way to link to them in a way that I can use one webhook for multiple users. Mastodon is able to do this and I think it'd be nice if bsky could as well.

Broadcast the user avatar URLs per post via RSS. Below is an example of how Mastodon handles it, and I believe bsky could use a similar approach without any major issues.

```xml
<image>
  <url>https://vt.social/media/accounts/avatars/109/308/259/723/197/265/original/e31c4176330dd79e.png</url>
  <title>熊柄秋子 (Akiko)</title>
  <link>https://vt.social/@Akiko</link>
</image>
```

I have considered using an alternative solution, such as using users' DID to access profile avatars, but this also isn't broadcasted in a clean way via RSS currently, so I'm very limited on options.

benubois:
Hello,

This adds an [`<image>`](https://www.rssboard.org/rss-specification#ltimagegtSubelementOfLtchannelgt) element containing the author's avatar. 

This way RSS readers would gain the option of displaying a nice avatar alongside the post. Maybe something like this:

bsky

This would close #3170.

Thanks!

### DIFF
--- a/bskyweb/cmd/bskyweb/rss.go
+++ b/bskyweb/cmd/bskyweb/rss.go
@@ -31,11 +31,19 @@ type Item struct {
 	GUID        ItemGUID
 }
 
+type Image struct {
+	XMLName xml.Name `xml:"image"`
+	URL     string   `xml:"url"`
+	Title   string   `xml:"title"`
+	Link    string   `xml:"link"`
+}
+
 type rss struct {
 	Version     string `xml:"version,attr"`
 	Description string `xml:"channel>description,omitempty"`
 	Link        string `xml:"channel>link"`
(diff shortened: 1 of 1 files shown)


mikekasprzak:
Silly question, but @gaearon's question about nullibility got me wondering: do all users have avatar URLs?

gaearon:
>do all users have avatar URLs?

From my understanding they don't necessarily, at least according to types in the app

PR #6899, opened 2024-12-02
Bump the initial fetch page size

gaearon:
Reverts https://github.com/bluesky-social/social-app/pull/5814 and hardcodes to essentially 90.

## Before

Screenshot 2024-12-02 at 17 44 43

## After

Screenshot 2024-12-02 at 17 44 50


gaearon:
will need to follow-up later

PR #6908, opened 2024-12-03
[Layout] Scrollbar gutters

estrattonbailey:
This PR resets us back to a base state in regards to scrollbar gutters. It then replaces our custom scroll-lock handling with that of `react-remove-scroll`, which is what Radix uses internally. This library provides utility classes and css vars to help us align things when scroll is locked. I'll document some things in a follow-up PR.


daniharo:
This change moves the left sidebar very close to the center section, which makes the UI feels a bit more crowded and less centered as the right sidebar takes a lot more width. Is that intended? (Maybe I'm just used to the current version)

| Before | After |
|--------|--------|
| image | image |

estrattonbailey:
Thanks @daniharo, I fixed that 👍

daniharo:
> Thanks @daniharo, I fixed that 👍

That's much better IMO.
However the buttons now have less width so in some languages when using "Larger" font size some options are overflowing the button, which is visible on hover:

| Language | Before | After |
|-|--------|--------|
| Polish | image | image |
| Ukranian | image | image |
| Spanish | image | image |

estrattonbailey:
@daniharo thanks for the double-check! That's the trouble with fixed width elements like this 😅 I've updated to allow a little more room, should be just about the same as what's in production. Clicked through the app languages and all looks good on my end now.

daniharo:
Thank you always! Now nothing overflows in the nav bar :)
Changing LeftBar's width to 260px instead of 240px would make it have the exact same space it has in prod, which I feel more comfortable, but as I said, maybe it's just personal preference.

I also noticed that the Search input width was before aligned with the separator lines below and now it's not, which also causes some overflow when reducing the viewport. This would be fixed setting the width to 300px for the sidebar:
| Before | After |
| -- | -- |
| image image | image image |


Sorry for so many comments, frontend eye can be a pain sometimes 😅

daniharo:
Also, when clicking on an image the feed scroll is reset. This doesn't happen on main so it seems to come from this PR:

https://github.com/user-attachments/assets/f9b4d2ce-381e-4fae-b680-7cd66b997081

estrattonbailey:
@daniharo not seeing the sidebar alignment issue, but have some follow-up code to improve that also, so nbd rn.

As for the jump thing, also had some separate code to fix that and merged it in here. Thanks for the help!

PR #6910, opened 2024-12-03
[Layout] Notifications Header

estrattonbailey:
Only notable changes here:
- When a soft-reset occurs, the gear icon is replaced with a loader icon (onPress unchanged)
- On mobile, the title is pressable as well — a11y is good, double checked on iOS/Android
- This new header has a bottom border, so I disabled on the notifications items and skeleton state first items

CleanShot 2024-12-02 at 19 08 30@2x
CleanShot 2024-12-02 at 19 08 17@2x
CleanShot 2024-12-02 at 19 01 30@2x
Screenshot_20241202-190138


mozzius (approved):
Pushed a fix for this, otherwise lgtm

Screenshot 2024-12-04 at 17 07 40

Issue #6897, opened 2024-12-02
Add Khmer Localization


roth-dev:
### Describe the Feature

Add Khmer Localization (ISO code:km)

The app currently lacks Khmer (km) localization. This issue aims to add Khmer translations for all UI strings and configure the app to support Khmer on both native (iOS/Android) and Web.

### Additional Context

**Why This is Needed:**
To improve accessibility and usability for Khmer-speaking users.

**Proposed Solution:**
Add Khmer translations for all existing UI strings.
Update localization files and configurations for both platforms.

roth-dev:
Merged: https://github.com/bluesky-social/social-app/pull/6914

PR #6283, opened 2024-11-13
Implement Polish translation

axiand:
## Prelude
I'm aware that this is already touched on in an earlier pull request (#2623), however that PR has been abandoned since late January of this year(!), and I don't believe it is going to be picked back up at this point. There is demand for an up-to-date localisation, as Bluesky's international userbase is rapidly growing and the app has evolved a lot since then.

## Description
This pull request adds Polish language support to the app. Please note that some strings may certainly be rough around the edges, as the immense amount of strings in the Bluesky app was a lot for me to do by myself, this is my first time tackling a translation project of this scale. With that in mind, proofreading and suggestions are both highly welcome.

PR #6914, opened 2024-12-03
Add Khmer Localization

roth-dev:


roth-dev:
This would close [[https://github.com/bluesky-social/social-app/issues/6897](url)](https://github.com/bluesky-social/social-app/issues/6897)

surfdude29:
It's always great to see another localization for Bluesky! 🦋🌐

Would it be possible to make three more changes in this PR to fully enable the Khmer localization please:

- In `src/components/hooks/dates.ts` please add a `km,` entry here:
https://github.com/bluesky-social/social-app/blob/147efb628a9eea755dd2901a82f39f14435f6629/src/components/hooks/dates.ts#L24-L25

- and here:
https://github.com/bluesky-social/social-app/blob/147efb628a9eea755dd2901a82f39f14435f6629/src/components/hooks/dates.ts#L61-L62

- And in `app.config.js` please add a `'km',` entry here:
https://github.com/bluesky-social/social-app/blob/147efb628a9eea755dd2901a82f39f14435f6629/app.config.js#L111-L112

surfdude29:
Also, the devs ask for localisation PRs to be reviewed by at least one other native speaker; is there someone you could ask to do that? They don't necessarily have to have a GitHub account :)

@/pfrazee said this about the initial review [on another localization PR](https://github.com/bluesky-social/social-app/pull/6283#issuecomment-2480806116):
> To get this initial one merged, if another native speaker could just give it an initial review. We don't have to get really detailed, but we want to make sure it's close enough!

src/locale/locales/km/messages.po
```diff
@@ -0,0 +1,8616 @@
...
+
+#: src/screens/Settings/AccessibilitySettings.tsx:42
+#: src/screens/Settings/Settings.tsx:183
+#: src/screens/Settings/Settings.tsx:186
+msgid "Accessibility"
+msgstr "ភាពងាយស្រួល"
+
+#: src/view/screens/Settings/index.tsx:455
+#~ msgid "Accessibility settings"
+#~ msgstr "ការកំណត់ភាពងាយស្រួល"
```
surfdude29, line 402: Also, one tip for the future: the lines starting with `#~` are obsolete strings that are no longer used in the app, so you can safely ignore them and there's no need to translate them. Some translators like to remove them to keep things tidy, but there's no need to do this as they'll simply be ignored by the app.

roth-dev:
@surfdude29 Thank you for reviewing my PR. Now I have added the missing key.

src/locale/locales/km/messages.po
roth-dev: Got it.

src/locale/locales/km/messages.po
```diff
@@ -0,0 +1,8616 @@
...
+
+#: src/Navigation.tsx:322
+msgid "Accessibility Settings"
+msgstr "ការកំណត់ភាពងាយស្រួល"
```
SokCheasrouch, line 406: Accessibility Settings would be better if translated to ការកំណត់មានភាពងាយស្រួល

src/locale/locales/km/messages.po
```diff
@@ -0,0 +1,8616 @@
...
+msgid "Account blocked"
+msgstr "គណនីត្រូវបានរារាំង"
+
+#: src/view/com/profile/ProfileMenu.tsx:145
+msgid "Account followed"
+msgstr "គណនីបានធ្វើតាម"
+
+#: src/view/com/profile/ProfileMenu.tsx:108
+msgid "Account muted"
+msgstr "គណនីត្រូវបានបិទ"
```
SokCheasrouch, line 427: Mute is different from Close and Block so I don't think the same translates as បិទ it's should be:
Mute = បិទសម្លេង 
Block = បញ្ឈប់

src/locale/locales/km/messages.po
surfdude29: A friendly tip in case it's helpful: you can suggest an alternate translation by clicking the ± button and then filling it in, e.g. like this:
```suggestion
msgstr "(suggested translation)"
```
If the PR author wants to accept the suggestion as it is, it's very easy for them to do so, and they can even accept many suggestions at once in a batch 💫

src/locale/locales/km/messages.po
roth-dev: So what would you suggest about `Account muted` in khmer?

roth-dev:
@SokCheasrouch I really appreciate your time and effort in helping me out. Please confirm if you're done with review. And everything should be ready to go🚀

SokCheasrouch:
@roth-dev I have just done my review

roth-dev:
@surfdude29 All done

src/locale/i18n.ts
```diff
@@ -166,6 +167,14 @@ export async function dynamicActivate(locale: AppLanguage) {
       ])
       break
     }
+    case AppLanguage.km: {
+      i18n.loadAndActivate({locale, messages: messagesKm})
+      await Promise.all([
+        import('@formatjs/intl-pluralrules/locale-data/km'),
```
surfdude29, line 173:
```suggestion
        import('@formatjs/intl-datetimeformat/locale-data/km'),
        import('@formatjs/intl-pluralrules/locale-data/km'),
```

Following #6742 being merged, could you add this line please?

This will ensure that the date and time for posts are localized correctly on all platforms. Thanks!

roth-dev:
All updated according to sugesstion.

pfrazee (approved):
Wow! Thanks so much!

PR #6919, opened 2024-12-03
[feat] change /profile/:name to /:name

callmearta:
This PR changes /profile/:name routes to /:name for easier route names.


MaxIsJoe:
won't this break existing profile links that are being shared/used around?
isn't there a better way to handle this, like a router that looks at a shortned `p` for profiles and points to the complete link?

callmearta:
I checked everything that came to my mind and nothing broke. There may be some issue somewhere that i'm unaware of but everything that I tried worked just fine so far. It required some changes in link generating functions as well. But I checked and history and everything worked just fine. I also changed some of the go functions but honestly I'm not a go developer and don't know if I did it right, so better check that part as well.

uncenter:
Thoughts on /@:name instead of just /:name? I prefer the more explicit @ symbol to clarify that this is a username.

bleonard252:
I think I'd like `/@handle` and `/did:...` to be the "preferred" versions since they're more direct and readable, with some redirects/aliases applied, particularly from `/profile/...` to `/@handle` or `/did:...`.

The main concern for breaking links to `/profile` routes is that _external_ tools and links exist. People posting links to Bluesky on chat apps or other social media platforms, or tools like [HAARP](https://haarp.blakeslabs.com) which compose profile links in some places -- these links break if the routes change. I support this change, but please make sure there's an alias/redirect to not completely break old links.

surfdude29:
Just my personal opinion, but I think `@` in URLs is ugly and should be avoided. Bluesky URLs have no need for it.

IIRC Twitter URLs used to use the `@` symbol early on, then they wisely ditched it.

callmearta:
If you are willing to merge this let me know so I can fix conflicts

PR #6921, opened 2024-12-03
Update messages.po - change "Back" translation to "Retour"

mc-zoi:


surfdude29:
Thanks for this, but there is actually already a PR open to do this :) #6778

PR #6924, opened 2024-12-03
ALF TabBar

mozzius:
Should broadly be the same as before, except I've reduced the height of the indicator by 1px (so it's now 2px tall)

# Test plan

Check it still all works on web and on native


estrattonbailey (approved):
LGTM, tested on iOS only

PR #6926, opened 2024-12-03
Simplify date picker theme

mozzius:


haileyok (approved):
as long as we are certain `dim` isn't one of the names (maybe we changed this or maybe it doesnt matter here?)

mozzius:
`dim` is a `t.name`, `t.scheme` is just `light`/`dark`

PR #6928, opened 2024-12-03
[Layout] Hashtag header

estrattonbailey:
Simple one. Only changes:
- back button exists on desktop now too
- share button exists on desktop now too (just copies to clipboard)

CleanShot 2024-12-03 at 13 11 22@2x
CleanShot 2024-12-03 at 13 11 27@2x
CleanShot 2024-12-03 at 13 11 07@2x
CleanShot 2024-12-03 at 13 11 03@2x
Screenshot_20241203-131039
Screenshot_20241203-131048


mozzius (approved):
lgtm

PR #6929, opened 2024-12-03
[Layout] ChatList header

estrattonbailey:
No changes to functionality

CleanShot 2024-12-03 at 13 38 31@2x
CleanShot 2024-12-03 at 13 38 58@2x
CleanShot 2024-12-03 at 13 38 35@2x
CleanShot 2024-12-03 at 13 39 06@2x
Screenshot_20241203-134018


mozzius (approved):
Also did chat settings, while we're here

Screenshot 2024-12-04 at 17 21 30

PR #6931, opened 2024-12-03
[ELI5] change sign in to create account

mozzius:
Also ALF'd some buttons

Screenshot 2024-12-03 at 21 29 45

Screenshot 2024-12-03 at 21 29 58

### DIFF
--- a/src/view/shell/NavSignupCard.tsx
+++ b/src/view/shell/NavSignupCard.tsx
@@ -44,12 +44,12 @@ let NavSignupCard = ({}: {}): React.ReactNode => {
       <View style={[a.flex_row, a.flex_wrap, a.gap_sm, a.pt_md]}>
         <Button
           onPress={showCreateAccount}
-          label={_(msg`Sign up`)}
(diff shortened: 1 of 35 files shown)


emilyliu7321 (approved):

src/view/shell/NavSignupCard.tsx
```diff
@@ -44,12 +44,12 @@ let NavSignupCard = ({}: {}): React.ReactNode => {
...
         <Button
           onPress={showCreateAccount}
-          label={_(msg`Sign up`)}
+          label={_(msg`Create account`)}
           size="small"
           variant="solid"
           color="primary">
           <ButtonText>
-            <Trans>Sign up</Trans>
+            <Trans>Create account</Trans>
```
estrattonbailey, line 52: Can we specify an `id` for Lingui so that this string doesn't change in other langs?

estrattonbailey (approved):
If we can save translators effort, should do that. But if we think this is more clear and can't avoid new translations, then that's a tradeoff we can take.

PR #6934, opened 2024-12-03
Add link-click metrics

pfrazee:
We're into driving clicks to people's sites, and we want to be able to tell creators & publishers about it. These metrics will help us convince more folks to make the switch.


haileyok (approved):
looks good to me

PR #6940, opened 2024-12-04
Adjust feed scroll bar top offset

skipness:
iOS only, add top offset to the feed scroll bar so the scroll bar won't cover by the tab bar

| Before | After |
|----|----|
| <video src='https://github.com/user-attachments/assets/ce7d4165-cbdc-4eb7-9d04-fb147383ae71' /> | <video src='https://github.com/user-attachments/assets/41fcb5f2-9d2d-4f5e-84a1-d55d3402950b' /> |


mozzius (approved):
Lovely, thank you

PR #6943, opened 2024-12-04
Add nepali translation

divyaswormakai:


surfdude29:
It's always great to see another localization for Bluesky! 🦋🌐

Would it be possible to make a couple more changes in this PR to fully enable the Nepali localization please:

- In `src/components/hooks/dates.ts` please add a `np: undefined,` entry here:
https://github.com/bluesky-social/social-app/blob/0ee20731d42bd58ac6a1020615e399b01767ce2d/src/components/hooks/dates.ts#L63-L64

This is because the [`date-fns`](https://www.npmjs.com/package/date-fns?activeTab=code) package unfortunately doesn't seem to support the `np` locale yet.

- In `app.config.js` please add a `'np',` entry here:
https://github.com/bluesky-social/social-app/blob/0ee20731d42bd58ac6a1020615e399b01767ce2d/app.config.js#L113-L114

This will enable Nepali to be listed as a supported language on the App Store page.

src/locale/i18n.ts
```diff
@@ -28,7 +28,8 @@ import {messages as messagesId} from '#/locale/locales/id/messages'
 import {messages as messagesIt} from '#/locale/locales/it/messages'
 import {messages as messagesJa} from '#/locale/locales/ja/messages'
 import {messages as messagesKo} from '#/locale/locales/ko/messages'
-import {messages as messagesNl} from '#/locale/locales/nl/messages'
+import { messages as messagesNl } from '#/locale/locales/nl/messages'
+import {messages as messagesNp} from '#/locale/locales/id/messages'
```
surfdude29, line 32:
```suggestion
import {messages as messagesNl} from '#/locale/locales/nl/messages'
import {messages as messagesNp} from '#/locale/locales/np/messages'
```
A couple of extra spaces seem to have been added here by accident, and also the language codes seem to have gotten a bit mixed up :)

src/locale/languages.ts
```diff
@@ -58,7 +59,8 @@ export const APP_LANGUAGES: AppLanguageConfig[] = [
   {code2: AppLanguage.it, name: 'Italiano – Italian'},
   {code2: AppLanguage.ja, name: '日本語 – Japanese'},
   {code2: AppLanguage.ko, name: '한국어 – Korean'},
-  {code2: AppLanguage.nl, name: 'Nederlands – Dutch'},
+  {code2: AppLanguage.nl, name: 'Nederlands – Dutch' },
```
surfdude29, line 62:
```suggestion
  {code2: AppLanguage.nl, name: 'Nederlands – Dutch'},
```
An extra space seems to have been added here by accident

surfdude29:
Also, the devs ask for localisation PRs to be reviewed by at least one other native speaker; is there someone you could ask to do that? They don't necessarily have to have a GitHub account :)

@/pfrazee said this about the initial review [on another localization PR](https://github.com/bluesky-social/social-app/pull/6283#issuecomment-2480806116):
> To get this initial one merged, if another native speaker could just give it an initial review. We don't have to get really detailed, but we want to make sure it's close enough!

src/locale/locales/np/messages.po
```diff
@@ -0,0 +1,9073 @@
...
+
+#: src/screens/Settings/AccountSettings.tsx:57
+#: src/view/com/modals/VerifyEmail.tsx:150
+msgid "(no email)"
+msgstr "इमेल छैन"
+
+#: src/view/com/notifications/FeedItem.tsx:232
+#: src/view/com/notifications/FeedItem.tsx:327
+#~ msgid "{0, plural, one {{formattedCount} other} other {{formattedCount} others}}"
+#~ msgstr "{0, plural, one {{formattedCount} अन्य} अन्य {{formattedCount} अन्यहरू}}"
```
surfdude29, line 27: A tip for the future: the lines starting with `#~` are obsolete strings that are no longer used in the app, so you can safely ignore them and there's no need to translate them. Some translators like to remove them to keep things tidy, but there's no need to do this if you'd prefer not to, as they'll simply be ignored by the app.

src/locale/locales/np/messages.po
```diff
@@ -0,0 +1,9073 @@
...
+
+#: src/lib/hooks/useTimeAgo.ts:156
+msgid "{0, plural, one {# day} other {# days}}"
+msgstr "{0, plural, one {# दिन} अन्य {# दिनहरू}}"
```
surfdude29, line 31:
```suggestion
msgstr "{0, plural, one {# दिन} other {# दिनहरू}}"
```
The formatting of the strings involving plurals is tricky, both `one` and `other` need to remain in English as they are syntax and not translatable strings.

This also applies to quite a few other strings that follow ⬇️

src/locale/locales/np/messages.po
divyaswormakai: Since, they have been translated, should we just let it be for now?

surfdude29: Yep, you can leave them as they are, that's no problem, I just wanted to let you know for the future :)

src/locale/locales/np/messages.po
divyaswormakai: updated the scripts likewise

divyaswormakai:
@surfdude29 
updated the messages files according to suggestions

whoisanku (commented):
Everything seems great but made some small changes into this PR #7033 & making this PR as final reviewed one and can be pushed to main. Every changes looks good now on Nepali language side from myself too as Nepali being a native language. 

Merge ready PR is #7033 !!!

divyaswormakai:
#7033  has some direct updates to it. so closing it as all the code here is copied there

PR #6953, opened 2024-12-04
[Layout] Every other screen

mozzius:
# Based on #6913

An attempt to go through every instance of `<ViewHeader>` and make sure there's no double borders. Most of the time it was just setting `topBorder={false}`, sometimes I needed to change the screen to use `<Layout.Center>` or something.

# Test plan

ctrl+f `<ViewHeader` and check every single one out, esp. on different web breakpoints

### DIFF
--- a/src/view/com/util/ViewHeader.tsx
+++ b/src/view/com/util/ViewHeader.tsx
@@ -1,20 +1,23 @@
 import {Header} from '#/components/Layout'
 
+/**
+ * Legacy ViewHeader component. Use Layout.Header going forward.
+ *
+ * @deprecated
+ */
 export function ViewHeader({
   title,
-  canGoBack,
   renderButton,
 }: {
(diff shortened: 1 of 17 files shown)


src/view/com/util/ViewHeader.tsx
```diff
@@ -1,20 +1,23 @@
...
   subtitle?: string
-  canGoBack?: boolean
   showOnDesktop?: boolean
   showBorder?: boolean
   renderButton?: () => JSX.Element
 }) {
   return (
     <Header.Outer>
-      {canGoBack ? <Header.BackButton /> : <Header.MenuButton />}
+      <Header.BackButton />
```
mozzius, line 20: Could not find an instance that *didn't* just need the back button

estrattonbailey (approved):

PR #6967, opened 2024-12-05
[Layout] Bleed profile banner into safe area

mozzius:
Best demonstrated via video

https://github.com/user-attachments/assets/158af9fb-d778-4f29-abb5-cad173559f90

## Old video with mini header (has been pulled out from this PR)

https://github.com/user-attachments/assets/0bc576bc-1806-42f8-9ec5-21a54684884b

# Test plan

Check it works good iOS and Android. Particular things to look for:
- Accessibility - is the minimal header accessible? is it correctly hidden from voiceover when invisible?
- Performance on low-end android. compare to current situation
- Check loading states, network conditions etc
Also check no regressions on web, other than adding in the back arrow on desktop

### DIFF
--- a/src/view/com/pager/PagerHeaderContext.tsx
+++ b/src/view/com/pager/PagerHeaderContext.tsx
@@ -1,40 +1,47 @@
 import React, {useContext} from 'react'
 import {SharedValue} from 'react-native-reanimated'
 
-import {isIOS} from '#/platform/detection'
+import {isNative} from '#/platform/detection'
 
-export const PagerHeaderContext =
-  React.createContext<SharedValue<number> | null>(null)
+export const PagerHeaderContext = React.createContext<{
+  scrollY: SharedValue<number>
+  headerHeight: number
+} | null>(null)
 
 /**
- * Passes the scrollY value to the pager header's banner, so it can grow on
- * overscroll on iOS. Not necessary to use this context provider on other platforms.
+ * Passes information about the
  *
- * @platform ios
+ * @platform ios, android
  */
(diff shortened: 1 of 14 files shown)


src/view/com/pager/PagerHeaderContext.tsx
```diff
@@ -1,40 +1,47 @@
...
-  React.createContext<SharedValue<number> | null>(null)
+export const PagerHeaderContext = React.createContext<{
+  scrollY: SharedValue<number>
+  headerHeight: number
+} | null>(null)
 
 /**
- * Passes the scrollY value to the pager header's banner, so it can grow on
- * overscroll on iOS. Not necessary to use this context provider on other platforms.
+ * Passes information about the
```
surfdude29, line 12: I think some of the comment might have gotten lost in a copy-paste here?

estrattonbailey (approved):
Excellent work dude! Tested on all devices and all seems well to me, the sped up animation def helps imo. No perf degradation that I can see 👍

src/view/shell/index.tsx
```diff
@@ -154,6 +155,7 @@ function ShellInner() {
 
 export const Shell: React.FC = function ShellImpl() {
   const {fullyExpandedCount} = useDialogStateControlContext()
+  const lightStatusBarRefCount = useLightStatusBarRefCount()
```
gaearon, line 158: don't pass the ref count in context, pass the boolean. so that it only changes between `true` and `false` and not on individual ref count changes

PR #6971, opened 2024-12-05
Refactor sidebar

estrattonbailey:
Extracted from subs branch. This cleans up the right nav to look better and fit incoming NUXs better. Looks especially better on the search page.

In the process, I made an improvement to the new `useGutters` hook so it's more ergonomic. 

CleanShot 2024-12-05 at 19 56 23@2x
CleanShot 2024-12-05 at 19 56 34@2x
CleanShot 2024-12-05 at 19 56 38@2x
CleanShot 2024-12-05 at 19 57 19@2x
CleanShot 2024-12-05 at 20 11 02@2x
(illustrating overflow with lots of feeds, works fine)
CleanShot 2024-12-05 at 20 15 23@2x

### DIFF
--- a/src/alf/util/useGutters.ts
+++ b/src/alf/util/useGutters.ts
@@ -30,10 +30,10 @@ const gutters: Record<
 }
 
 type Gutters = {
-  paddingTop?: number
-  paddingRight?: number
-  paddingBottom?: number
-  paddingLeft?: number
+  paddingTop: number
+  paddingRight: number
+  paddingBottom: number
+  paddingLeft: number
 }
 
 export function useGutters([all]: [Gutter]): Gutters
@@ -56,16 +56,13 @@ export function useGutters([top, right, bottom, left]: Gutter[]) {
     }
 
     return {
-      paddingTop:
-        top === 0 ? undefined : gutters[top][activeBreakpoint || 'default'],
+      paddingTop: top === 0 ? 0 : gutters[top][activeBreakpoint || 'default'],
       paddingRight:
-        right === 0 ? undefined : gutters[right][activeBreakpoint || 'default'],
+        right === 0 ? 0 : gutters[right][activeBreakpoint || 'default'],
       paddingBottom:
-        bottom === 0
-          ? undefined
-          : gutters[bottom][activeBreakpoint || 'default'],
+        bottom === 0 ? 0 : gutters[bottom][activeBreakpoint || 'default'],
       paddingLeft:
-        left === 0 ? undefined : gutters[left][activeBreakpoint || 'default'],
+        left === 0 ? 0 : gutters[left][activeBreakpoint || 'default'],
     }
   }, [activeBreakpoint, top, bottom, left, right])
 }
--- a/src/alf/util/useGutters.ts
+++ b/src/alf/util/useGutters.ts
@@ -44,11 +44,12 @@ export function useGutters([top, right, bottom, left]: [
   Gutter,
   Gutter,
 ]): Gutters
-export function useGutters([top, right, bottom, left]: Gutter[]) {
+export function useGutters(defs: Gutter[]) {
   const {activeBreakpoint} = useBreakpoints()
   return React.useMemo(() => {
+    let [top, right, bottom, left] = defs
+
     if (!right) {
-      // eslint-disable-next-line react-hooks/exhaustive-deps
       right = bottom = left = top
     } else if (!bottom || !left) {
       bottom = top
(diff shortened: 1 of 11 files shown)


src/components/Layout/Header/index.tsx
```diff
@@ -34,7 +34,7 @@ export function Outer({
   noBottomBorder?: boolean
 }) {
   const t = useTheme()
-  const gutter = useGutterStyles()
+  const gutters = useGutters([0, 'base'])
```
estrattonbailey, line 37: No change here from previous styles

src/view/com/home/HomeHeaderLayout.web.tsx
```diff
@@ -41,14 +41,14 @@ function HomeHeaderLayoutDesktopAndTablet({
   const {hasSession} = useSession()
   const {_} = useLingui()
   const kawaii = useKawaiiMode()
-  const gutter = useGutterStyles()
+  const gutters = useGutters([0, 'base'])
```
estrattonbailey, line 44: No change here from previous styles

Tanza3D:
image
these feel reaaally squished together now

estrattonbailey:
@Tanza3D I think you may be looking at an old version here. It's missing a link at the bottom for "More feeds" and the spacing around the sidebar is now wider on the left than on top, whereas here it looks event.

Or are you only referring to the line-height between the links?

Tanza3D:
> @Tanza3D I think you may be looking at an old version here. It's missing a link at the bottom for "More feeds" and the spacing around the sidebar is now wider on the left than on top, whereas here it looks event.
> 
> Or are you only referring to the line-height between the links?

mainly referring to the line height/padding between the feed links/buttons yeah; I feel they'd be better displayed as tabs similar to discords channel lists and such perhaps

src/alf/util/useGutters.ts
```diff
@@ -0,0 +1,68 @@
...
+  Gutter,
+  Gutter,
+  Gutter,
+  Gutter,
+]): Gutters
+export function useGutters([top, right, bottom, left]: Gutter[]) {
+  const {activeBreakpoint} = useBreakpoints()
+  return React.useMemo(() => {
+    if (!right) {
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
gaearon, line 51: nah, the rule is right, we shouldn't be doing this to function args.

gaearon: you can either extract these out of memo or declare new variables

src/alf/util/useGutters.ts
```diff
@@ -0,0 +1,69 @@
...
+    return {
+      paddingTop: top === 0 ? 0 : gutters[top][activeBreakpoint || 'default'],
+      paddingRight:
+        right === 0 ? 0 : gutters[right][activeBreakpoint || 'default'],
+      paddingBottom:
+        bottom === 0 ? 0 : gutters[bottom][activeBreakpoint || 'default'],
+      paddingLeft:
+        left === 0 ? 0 : gutters[left][activeBreakpoint || 'default'],
+    }
+  }, [activeBreakpoint, defs])
```
gaearon, line 68: this memoization seems useless because `defs` will be an inline array and will be different on every render. let's either solve this by destructuring early or remove the `useMemo`?

gaearon (changes requested):
thanks! a few things that jumped out to me that seem like regressions:

- pressing on a feed has gotten _much_ slower in this PR, something regressed there
- feed link staying underlined after navigation click feels weird to me, not sure why it does that
- something's wrong with the top bar, the butterfly and the # icon lost top padding

these are maybe more subjective:

- not sure about more narrow input, it's very noticeable how usernames rarely fit and display names don't fit more often too. i would suggest that if we were to narrow it, we should also turn the dropdown into an actual dropdown (and give more width to the list)
- not a fan of "external link" icon on more feeds, it looks like a link to another site
- overall they feel a bit too crowded to me, this is secondary navigation but they're very close to each other and easy to misclick. i'd prefer if there was a bit more space there

lmk if you disagree with some of these

gaearon (approved):
lfg

src/view/shell/desktop/RightNav.tsx
```diff
@@ -1,149 +1,106 @@
...
-            </Text>
-          )}
-        </View>
-      </View>
+      {kawaii && (
+        <Text style={[t.atoms.text_contrast_medium, {marginTop: 12}]}>
+          <Trans>
+            Logo by{' '}
+            <InlineLinkText
+              label={_(msg`Logo by @sawaratsuki.bsky.social`)}
```
mlocati, line 97: Should we translate `@sawaratsuki.bsky.social`? If not, what about using a placeholder instead of that username?

gaearon: Yeah there’s also already “logo by” around it, I think we can just remove the repetition to what it was before.

PR #6975, opened 2024-12-05
PT-BR Translation update

Felps3000:
Added translation to the new features in 1.95.0
Fixed inconsistencies


pfrazee:
A big part of this PR is changing the gender of "post" to "postagem." I've heard conflicting answers on whether that's correct

Felps3000:
The word "post" officially doesn't exist in brazilian portuguese, and I just followed the preexisting "postagem" that was being used in previous translations, one more accurate solution would be translate "post" to "publicação"

luan-u:
Right now there is mixed use of "post" and "postagem". I prefer the latter option because it feels more natural.

_Postagem_ (noun) and _postar_ (verb) are existing words with the sense of "putting a parcel in the mail [to be sent]", which got this new meaning of "publishing something in an online platform" with the rise of internet in our daily lives, and it is well established in any traditional dictionary. _Post_, on the other hand, is recognized as an anglicism with only the meaning of an online publication, but it has found widespread adoption. One could argue the average Portuguese-speaking netizen knows very well what a _post_ is.

I support the change from _post_ to _postagem_, but I must admit it stems from a perceived sense of language pureness. _post_ is borrowed from English, but so are many other words in Portuguese, coming from many other languages.

Felps3000:
This mixed use of "post" and "postagem" is something I was trying to get rid of since I started to collaborate in the translation, you brought some good commentaries about "postagem" being well spread among online users, in my opinion we should keep using "postagem" as a translation to "post".

pfrazee:
Okay I'm good w/that. Thanks yall for the explanation

pfrazee (approved):
Thanks!

Lucasofchirstms:
@luan-u Let's set it to Publicação or just Postagem, what do you prefer?

luan-u:
I prefer "postagem", which is already integrated (except for one or other string here and there).

Changing to "publicação" would require some rewriting, since we would also need to change "repostagem" to "republicação" to mantain consistency. So, unless there is a stronger reason, I prefer we stick with "postagem".

Lucasofchirstms:
Yeah, okay let just stick to Postagem

PR #6979, opened 2024-12-05
configure the bitdrift Capture SDK

snowp:
This wires up configuring the React Native version of the bitdrift Capture SDK and hooks it up as a logging transport


gaearon (approved):
lets try this

PR #6982, opened 2024-12-06
Minor boolean refactor for simplicity

seth-wood:
Tiny refactor of the booleans for simplicty and character-saving.


gaearon:
thanks for PR but we don't take purely stylistic changes, that's very subjective

seth-wood:
Makes sense - thanks for the feedback.

PR #6984, opened 2024-12-06
Fix spanish translation

ovniroto:
I have corrected the word "Agregar" (previously "Aregar") in the sentence "Agregar otra publicación".


pfrazee (approved):
Thanks!

PR #6991, opened 2024-12-06
French translations (for v1.96)

Signez:
It may be too early for the next version, but at least the pull request is already open, and we will be able to update it here.

- Bumped the strings;
- Translated the 18 new strings;
- Fixed a few typos / rewrote a few translations (got rid of `Feedback`, replacing it by `Commentaires` instead as I've seen this term used in other software for this matter);
- Deleted all the old, legacy strings that accumulated from a few versions now.

@surfdude29 Feel free to review! _Be careful to do it on a commit-by-commit basis_, as I took the opportunity to remove all the legacy commented strings, so it could be noisy to review it all at once. Those removals all take place in the last commit.

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -17,7 +17,7 @@ msgstr ""
 msgid "(contains embedded content)"
 msgstr "(contient du contenu intégré)"
 
-#: src/screens/Settings/AccountSettings.tsx:57
+#: src/screens/Settings/AccountSettings.tsx:65
 #: src/view/com/modals/VerifyEmail.tsx:150
 msgid "(no email)"
 msgstr "(pas d’e-mail)"
@@ -102,7 +102,7 @@ msgstr "{0, plural, one {repost} other {reposts}}"
 msgid "{0, plural, one {Unlike (# like)} other {Unlike (# likes)}}"
 msgstr "{0, plural, one {Retirer la mention « J’aime » (# ont aimé)} other {Retirer la mention « J’aime » (# ont aimé)}}"
 
-#: src/screens/Settings/Settings.tsx:414
+#: src/screens/Settings/Settings.tsx:422
 msgid "{0}"
 msgstr "{0}"
 
@@ -177,20 +177,20 @@ msgstr "{badge} éléments non lus"
 msgid "{count, plural, one {Liked by # user} other {Liked by # users}}"
 msgstr "{count, plural, one {Aimé par # compte} other {Aimé par # comptes}}"
 
-#: src/view/shell/desktop/LeftNav.tsx:223
+#: src/view/shell/desktop/LeftNav.tsx:186
 msgid "{count} unread items"
 msgstr "{count} éléments non lus"
 
 #: src/lib/generate-starterpack.ts:108
-#: src/screens/StarterPack/Wizard/index.tsx:180
+#: src/screens/StarterPack/Wizard/index.tsx:178
 msgid "{displayName}'s Starter Pack"
 msgstr "Kit de démarrage de {displayName}"
 
-#: src/screens/SignupQueued.tsx:220
+#: src/screens/SignupQueued.tsx:219
 msgid "{estimatedTimeHrs, plural, one {hour} other {hours}}"
 msgstr "{estimatedTimeHrs, plural, one {heure} other {heures}}"
 
-#: src/screens/SignupQueued.tsx:226
+#: src/screens/SignupQueued.tsx:225
 msgid "{estimatedTimeMins, plural, one {minute} other {minutes}}"
 msgstr "{estimatedTimeMins, plural, one {minute} other {minutes}}"
 
@@ -293,7 +293,7 @@ msgstr "{handle} ne peut être contacté par message"
 
 #: src/screens/Profile/Header/ProfileHeaderLabeler.tsx:294
 #: src/screens/Profile/Header/ProfileHeaderLabeler.tsx:307
-#: src/view/screens/ProfileFeed.tsx:591
+#: src/view/screens/ProfileFeed.tsx:590
 msgid "{likeCount, plural, one {Liked by # user} other {Liked by # users}}"
 msgstr "{likeCount, plural, one {Aimé par # compte} other {Aimé par # comptes}}"
 
@@ -313,12 +313,12 @@ msgstr "{profileName} a rejoint Bluesky il y a {0}"
 msgid "{profileName} joined Bluesky using a starter pack {0} ago"
 msgstr "{profileName} a rejoint Bluesky en utilisant un kit de démarrage il y a {0}"
 
-#: src/screens/StarterPack/Wizard/index.tsx:472
+#: src/screens/StarterPack/Wizard/index.tsx:449
 msgctxt "profiles"
 msgid "<0>{0}, </0><1>{1}, </1>and {2, plural, one {# other} other {# others}} are included in your starter pack"
 msgstr "<0>{0}, </0><1>{1} </1>et {2, plural, one {# autre} other {# autres}} font partie de votre kit de démarrage"
 
-#: src/screens/StarterPack/Wizard/index.tsx:525
+#: src/screens/StarterPack/Wizard/index.tsx:502
 msgctxt "feeds"
 msgid "<0>{0}, </0><1>{1}, </1>and {2, plural, one {# other} other {# others}} are included in your starter pack"
 msgstr "<0>{0}, </0><1>{1} </1>et {2, plural, one {# autre} other {# autres}} sont inclus dans votre kit de démarrage"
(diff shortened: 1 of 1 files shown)


src/locale/locales/fr/messages.po
```diff
@@ -1275,7 +1275,7 @@ msgstr "Modifier"
 
 #: src/screens/Settings/AppIconSettings.tsx:81
 msgid "Change app icon to \"{0}\""
-msgstr ""
+msgstr "Changer l’icône de l’application pour \"{0}\""
```
surfdude29, line 1225: Might it not make sense to use _guillemets_ here, as it's referring to the name of the app icon?

surfdude29 (commented):
Looks great to me! I just have one small query.

Also, would it make sense to incorporate the change in #6778 in this PR, so that there's only one French PR to merge?

src/locale/locales/fr/messages.po
Signez: Oh, good catch!

surfdude29:
Nice, looks ready to go to me 👌

src/locale/locales/fr/messages.po
```diff
@@ -841,43 +822,31 @@ msgstr "Appel soumis"
...
-
 #: src/screens/Feeds/NoSavedFeedsOfAnyType.tsx:47
 #: src/screens/Home/NoFeedsPinned.tsx:93
 msgid "Apply default recommended feeds"
 msgstr "Utiliser les fils d’actu recommandés par défaut"
 
 #: src/view/com/post-thread/PostThreadItem.tsx:835
 msgid "Archived from {0}"
-msgstr "Archivé par {0}"
+msgstr "Archivé comme du {0}"
```
CerryTsuki, line 839: this feels weird in canadian french. from the source code i think this is posts that are imported from another website? maybe using "**Archivé sur**" would be better, but it wouldn't be that big a deal since "Archivé" followed by a website + date should make it obvious to users what this information is

src/locale/locales/fr/messages.po
```diff
@@ -2277,7 +2168,7 @@ msgstr "ex. Artiste, amie des chiens et lectrice passionnée."
 
 #: src/lib/moderation/useGlobalLabelStrings.ts:43
 msgid "E.g. artistic nudes."
-msgstr "Ex. nus artistiques."
+msgstr "Nus artistiques par exemple."
```
CerryTsuki, line 2171: i feel like i've always heard "nudité artistique" over "nus artistiques", i think that'd be a better fit

src/locale/locales/fr/messages.po
```diff
@@ -4175,28 +3970,28 @@ msgstr "Masquer ce fil de discussion"
...
 
 #: src/lib/moderation/useModerationCauseDescription.ts:90
 msgid "Muted by \"{0}\""
 msgstr "Masqué par « {0} »"
 
-#: src/screens/Moderation/index.tsx:229
+#: src/screens/Moderation/index.tsx:208
 msgid "Muted words & tags"
-msgstr "Les mots et les mots-clés masqués"
+msgstr "Mots et les mots-clés masqués"
```
CerryTsuki, line 3992: i think you missed the second "les" here

src/locale/locales/fr/messages.po
```diff
@@ -6804,6 +6368,11 @@ msgstr "Menu de mot-clé : {displayTag}"
 msgid "Tags only"
 msgstr "Mots-clés seulement"
 
+#: src/screens/Settings/AppIconSettings.tsx:42
+#: src/screens/Settings/AppIconSettings.tsx:76
+msgid "Tap to change app icon"
+msgstr "Tapper pour changer l’icône de l’application"
```
CerryTsuki, line 6374: unsure about parisian french or other variants of french, but "tapper" in canadian french is similar to "to punch". "**Appuyez**" would be a better fit

CerryTsuki:
i threw in a few comments in translations that i think could be improved. i'm not sure how to suggest code changes directly in the PR though, sorry about that

surfdude29:
> i'm not sure how to add code changes directly in the PR

@CerryTsuki You can do that by clicking the ± symbol :)

(although annoyingly, sometimes it doesn't appear for me, not sure why)

src/locale/locales/fr/messages.po
Signez: Oh, good catch. I was so frustrated with the first `les` that I forgot the other one haha

src/locale/locales/fr/messages.po
Signez: `Tapper` in that sense in not unheard of on this side of the Atlantic ocean, but `Appuyez` is better anyway, so let's do that.

src/locale/locales/fr/messages.po
Signez: I understand your confusion here — but no, here `{0}` is a **date**:

image

I wrote it like `Archivé comme du 15 avril 1975`, as if `Archivé comme provenant du 15 avril 1975`. It's not _the best_, but clearly it's better than _Archivé par_ or _Archivé sur_, which would not work with a date.

Do you have any better idea, now that it's clearer `{0}` is filled with a date?

Signez:
Just rebased on `main`, [fixing three out of four errors](https://github.com/bluesky-social/social-app/pull/6991/commits/28e6dbab8d9eb4f511df99aa9f740b3e11a048e6) @CerryTsuki spotted (thanks again!).

Also bumped the translation with [two new translations](https://github.com/bluesky-social/social-app/pull/6991/commits/ee8de910103369297feecf1d679bd411dcb7b84c), where I had to deviate from the original English copy because we don't have a satisfactory translation for `timeline` — and surprisingly, this is the first occurrence of this noun in the UI!

(I also spotted the reference to the _Choose Your Own Adventure_ book series in that new string, but those book titles were translated as _Le livre dont vous êtes le héros_ in French… and I wasn't able to twist it in a satisfying manner. `La [timeline] dont vous êtes le héros ou héroïne !` sounds quite lame imho)

src/locale/locales/fr/messages.po
CerryTsuki: "_**Archivé le** 4 avril 1975 à 13:00_" sounds better
```
msgstr "Archivé le {0}"
```
admittedly though i'm not sure why "Archived from" is the original string, i don't think i've seen or heard "Archived from April 4th 1975"

Signez: It sounds better, but it would be a mistranslation in my opinion: it's not `Archived at` (as in, _this post was archived at that date_), but `Archived from`: _this post was archived as coming from that date_.

I think that's why `from` was used, and that's why I didn't simply chose `Archivé le`, which would have been the easiest choice: it doesn't mean the same thing!

Signez: We could sidestep the problem by writing it like this:

```
Archive datée du {0}
```

It's a bit longer, but way clearer. What do you think?

CerryTsuki: i like that way better, and it seems like it'd fit the UI, go for it

Signez:
Rebased again on `main`.

Translated 30+ new strings, mainly because of #7065 (but it's quick to fix), but also because of the new exact count headers introduced in #7057 and #7059. You can review them in [this commit](https://github.com/bluesky-social/social-app/pull/6991/commits/52ff6c4a58e257c905f440bfbcf043b5eaec5054).

src/locale/locales/fr/messages.po
```diff
@@ -3314,7 +3314,7 @@ msgstr "Si vous supprimez cette liste, vous ne pourrez pas la récupérer."
 
 #: src/screens/Settings/components/ChangeHandleDialog.tsx:250
 msgid "If you have your own domain, you can use that as your handle. This lets you self-verify your identity. <0>Learn more here.</0>"
-msgstr ""
+msgstr "Si vous possédez votre propre domaine, vous pouvez l’utiliser comme pseudonyme. Cela vous permet d’auto-vérifier votre identité. <0>En savoir plus</0>"
```
surfdude29, line 3245: We've previously translated `handle` as simply `pseudo` in all the other strings, so I think it might make sense to stick with that here for consistency?

Also, for my own education, I was wondering: in French does it work better not to have a `.` after `En savoir plus` because it's a sentence fragment, rather than a complete sentence? 🤔

src/locale/locales/fr/messages.po
```diff
@@ -8154,7 +8154,7 @@ msgstr "Votre choix sera enregistré, mais vous pourrez le modifier ultérieurem
 
 #: src/screens/Settings/components/ChangeHandleDialog.tsx:496
 msgid "Your current handle <0>{0}</0> will automatically remain reserved for you. You can switch back to it at any time from this account."
-msgstr ""
+msgstr "Votre pseudonyme actuel <0>{0}</0> sera automatiquement conservé et vous sera réservé. Vous pourrez y revenir à tout moment depuis ce compte."
```
surfdude29, line 8043: Same comment about the translation of `handle` here :)

surfdude29 (commented):
LGTM ✅ Great work @Signez! :)

I just have a couple of nits about the translation of `handle`.

src/locale/locales/fr/messages.po
Signez: Fixed ✅

And yeah, it's exactly for that reason I prefer not to keep "dangling full stops" at the end of such links. In my recollection, it's more common not to include them for `En savoir plus`, or to use ellipsis instead: `En savoir plus…` (but that's more of an old way of stating that _a dialog will follow_, as we used to do in UI copy in the old days!)

surfdude29: Makes sense to me 👍

pfrazee (approved):
Thanks all!!!

PR #6993, opened 2024-12-06
Font loading improvements

### LINKED ISSUE #6992: Fonts loading incorrectly and duplicating downloads
### Steps to Reproduce

As noted at: https://bsky.app/profile/screenspan.net/post/3lcnm3xpza225

1. Open DevTools
2. Got to https://bsky.app/
3. Notice the following errors in the console:

> A preload for 'https://bsky.app/static/media/InterVariable.c9f788f6e7ebaec75d7c.ttf' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.Understand this warningAI
3lcnm3xpza225:1
> A preload for 'https://bsky.app/static/media/InterVariable-Italic.55d6a3f35e9b605ba6f4.ttf' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.

4. Note in the Network track the Intervariable font is downloaded twice.
5. This is use to the lack of `crossorigin` in the preload lines ([needed despite it being same origin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#cors-enabled_fetches) because... fonts).

Additional issues:
- The fonts are loaded as OTF fonts even though woff2 is [widely supported](https://caniuse.com/woff2) and the only things you need to use. This is 520 KiB instead of 346 KiB for the normal font.
- The italic font is preloaded but then not used (at least not quickly) and that's another 558 KiB unnecessary download (and this italic font would automatically be downloaded later if needed).

### Attachments

_No response_

Web (Desktop), Web (Mobile)

### Device Info

MacOS 15.1.1 (24B91)

1.95.0

### Additional Information

_No response_

tunetheweb:
Fixes #6992

- Switches to WOFF2 fonts ~(note unsure of licensing issues in changing the font format?)~ - updated [they include WOFF2 in their release](https://github.com/rsms/inter/releases/download/v4.1/Inter-4.1.zip) so no licensing issues.
- Add `crossorigin` attribute to preloads
- Removes Italics preload as it doesn't appear to be used?


gaearon:
@estrattonbailey mind looking at this one?

tunetheweb:
Oh I'm missing these fonts as I ran out of my daily free credits on https://cloudconvert.com/otf-to-woff2 😔 :
- Inter-Black.woff2
- Inter-BlackItalic.woff2
- Inter-Bold.woff2
- Inter-BoldItalic.woff2
- Inter-ExtraBold.woff2
- Inter-ExtraBoldItalic.woff2

Will add to PR tomorrow after they reset as trying to convert locally is proving tricky, and that service was the easist.

tunetheweb:
Never mind. Finally got my Glyphhanger command to work so reconverted them all.

stoyan (commented):
these hashes e.g. `c9f788f6e7ebaec75d7c` look like artifacts of a build process, but otherwise the PR is spot on

bskyweb/templates/base.html
```diff
@@ -26,14 +25,14 @@
...
-      src: url("{{ staticCDNHost }}/static/media/InterVariable.c9f788f6e7ebaec75d7c.ttf") format('truetype');
+      src: url("{{ staticCDNHost }}/static/media/InterVariable.c9f788f6e7ebaec75d7c.woff2") format('truetype');
       font-weight: 300 1000;
       font-style: normal;
       font-display: swap;
     }
     @font-face {
       font-family: 'InterVariableItalic';
-      src: url("{{ staticCDNHost }}/static/media/InterVariable-Italic.55d6a3f35e9b605ba6f4.ttf") format('truetype');
+      src: url("{{ staticCDNHost }}/static/media/InterVariable-Italic.55d6a3f35e9b605ba6f4.woff2") format('truetype');
```
rviscomi, line 35:
```suggestion
      src: url("{{ staticCDNHost }}/static/media/InterVariable-Italic.55d6a3f35e9b605ba6f4.woff2") format('woff2');
```

bskyweb/templates/base.html
```diff
@@ -26,14 +25,14 @@
      */
     @font-face {
       font-family: 'InterVariable';
-      src: url("{{ staticCDNHost }}/static/media/InterVariable.c9f788f6e7ebaec75d7c.ttf") format('truetype');
+      src: url("{{ staticCDNHost }}/static/media/InterVariable.c9f788f6e7ebaec75d7c.woff2") format('truetype');
```
rviscomi, line 28:
```suggestion
      src: url("{{ staticCDNHost }}/static/media/InterVariable.c9f788f6e7ebaec75d7c.woff2") format('woff2');
```

rviscomi (commented):
Spotted a couple of mismatched formats

gaearon:
yea i think we need to figure out what's up with the hashes here, i doubt they'll match up?

tunetheweb:
I ran `npm run web-build` which generated the files with updated hashes.

estrattonbailey:
@tunetheweb thanks for this! I'm fixing some env issues locally on my machine, so haven't given this a good look yet.

We use Expo, and [according to their docs](https://github.com/bluesky-social/social-app/pull/7008), woff2 isn't supported on Android. So we'll need to double check if that's still the case, otherwise we probably want to just stick with TTF and OTF.

Good call on the preloads. We use italics very rarely right now, so should be fine to lazily load that when italic usages render.

tunetheweb:
Yeah should be easy enough to load all fonts but for web, only reference woff2. Will take a look tomorrow.

Alternatively we reduce the scope of this OR to just fixing the preloads (to add crossorigin to one, and remove th elther) and do the rest on another PR of you prefer?

FYI for a future PR I’d love to subset more as those Variable fonts are huge! But definitely let’s keep that out of this PR!

stoyan:
I agree, I think reducing the scope of this PR to just fixing the wasted downloads makes perfect sense. And leave the woff2-ification to another PR.

The difference between compressed otf/ttf vs woff2 is not that great anyway, way under 2x

Screenshot 2024-12-07 at 3 01 01 PM


OTOH subsetting will be a bigger win. Maybe using the [inverse subsetting](https://www.npmjs.com/package/inverse-subset) too for LATIN vs everything else. Here's what `Inter-LightItalic` file sizes look like when subset to LATIN and to everything but LATIN:

Screenshot 2024-12-07 at 3 20 33 PM

tunetheweb:
> Alternatively, we could use woff2 on web only. Should be pretty easy to fork logic in our `fonts` handling.

Looks like you already forked it in #5776 so this was super easy. Noe in https://github.com/bluesky-social/social-app/pull/6993/commits/1c7bc60b2e8358c2227f33765198fbe686db718e.

benharri:
Inter ships woff fonts in their release package https://github.com/rsms/inter/releases/download/v4.1/Inter-4.1.zip - no need to run them through a conversion.

Inverse subset is an interesting idea. I'd be curious to see how often the non-latin set would be loaded.

tunetheweb:
> Inter ships woff fonts in their release package https://github.com/rsms/inter/releases/download/v4.1/Inter-4.1.zip - no need to run them through a conversion.

Oh excellent! Updated to those instead. @estrattonbailey I think this is ready for another review?

> Inverse subset is an interesting idea. I'd be curious to see how often the non-latin set would be loaded.

Even if they are loaded frequently it's unlikely users would want multiple non-latin sets. So that's where the gain is. But it seems [inter have tried to subset and then pulled it back](https://github.com/rsms/inter/issues/322#issuecomment-811574107) so might be worth waiting for them to officially support this before we got to that step.

gaearon (approved):
let's give this a go

estrattonbailey:
Looks good to me, nice work all! Appreciate the help here 🙏

app.config.js
```diff
@@ -234,8 +234,8 @@ module.exports = function (config) {
           'expo-font',
           {
             fonts: [
-              './assets/fonts/inter/InterVariable.ttf',
-              './assets/fonts/inter/InterVariable-Italic.ttf',
+              './assets/fonts/inter/InterVariable.woff2',
+              './assets/fonts/inter/InterVariable-Italic.woff2',
```
gaearon, line 238: hm, this broke iOS build

gaearon: nvm, my mistake

PR #7008, opened 2024-12-07
Pin Ruby to 3.3.6

estrattonbailey:
Trying to solve an outdated dependency by updating my CocoaPods install sent me into a spiral. The previous default RN Ruby version 2.7.6 is EOL, as is openssl 1.1, which is required for these old ruby versions. So trying to rug my install and start fresh with system Ruby was failing miserably.

RN is [no longer pinned to a Ruby version](https://github.com/facebook/react-native/pull/36205#issuecomment-1446044004), which led mean to trying a fresh Ruby and CocoaPods install using Homebrew.

After resetting everything and reinstalling everything with Homebrew, all that was left was to update the `.ruby-version` file in our repo. Everything works fine afaict.

CleanShot 2024-12-07 at 14 27 47@2x


gaearon (approved):
sure lg

PR #7010, opened 2024-12-07
Add back button to desktop threads

estrattonbailey:
Noticed this got dropped in the layout refactor. This keeps the logic, but doesn't gate it based on breakpoint.

CleanShot 2024-12-07 at 16 43 17@2x


mozzius (approved):

mozzius:
Overdue - we should also think about what screens we want to make this sticky on web

PR #7015, opened 2024-12-08
add optimization

gwizxs:


gaearon:
Any particular reason you chose to optimize these specific components?

PR #7019, opened 2024-12-08
LoginForm: enable autocomplete on 2FA code field

rcombs:
This should (fingers crossed) work in Safari, at least.

### DIFF
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -257,7 +257,7 @@ export const LoginForm = ({
               autoCapitalize="none"
               autoFocus
               autoCorrect={false}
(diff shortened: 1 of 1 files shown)


gaearon:
How do we test this? Does Safari actually autofill them from email?

rcombs:
Generally Safari will autofill codes received in Mail.app. I haven't been able to get it to happen on bsky.app/bsky.social by modifying the page in the web inspector, but I'm not sure if that's because I'm modifying the element post-creation or if it doesn't like the formatting of the email somehow. In any case, having autocomplete enabled is necessary (even if it's not sufficient on its own) to get this working.

gaearon (approved):
doesn't hurt to try

PR #7028, opened 2024-12-09
Add `isSelfHosted` to `SessionAccount`

estrattonbailey:
In our current implementation this is correct, but this may need to change with OAuth. We'll need to confirm that [this getter](https://github.com/bluesky-social/atproto/blob/main/packages/api/src/atp-agent.ts#L95) will remain stable.

Why? This is needed for subscriptions to help us provide support. We could scatter about this logical check, but it'd be better to handle in once place where we can expand the logic to account for OAuth in the future.


gaearon (approved):
lgtm

src/state/session/agent.ts
```diff
@@ -204,6 +205,7 @@ export function agentToSessionAccount(
     active: agent.session.active,
     status: agent.session.status as SessionAccount['status'],
     pdsUrl: agent.pdsUrl?.toString(),
+    isSelfHosted: !agent.serviceUrl.toString().startsWith(BSKY_SERVICE),
```
DavidBuchanan314, line 208: The current value of `BSKY_SERVICE` is `https://bsky.social`. So services at `https://bsky.social.example.com` or even `https://bsky.socialblade.com` would be erroneously classified as not-self-hosted. Not sure if this matters much in practice.

estrattonbailey: Yeahhh kinda just chucked this in as a temp thing. I've since connected with Matthieu and have a ticket on the todo list to make this stable before we start relying on this value.

Issue #6951, opened 2024-12-04
"New Post" hotkey can trigger in "Reply" mode, detaching the reply.


WolfWings:
### Steps to Reproduce

1. Click 💬
2. Change focus away from the text-entry field; there are many ways to do this, clicking where the 💬 was (so an accidental double-click) or hitting tab/shift+tab or many, Many, MANY other things can cause this.
3. Press the 🅽 key on your keyboard, virtual or otherwise.

Suddenly you're no longer 'replying' but 'making a new post' instead.

This will even happen if you're several + posts deep trying to make a concise multi-part reply all at once, if focus EVER leaves active text-entry boxes the 🅽 key blows away the whole reply structure.

### Attachments

https://github.com/user-attachments/assets/023b5abe-dd48-4afb-a9a0-911bc7e04edf

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### What version of the app are you using?

Build version: 1.95.0; Bundle info: c9d86de (prod); Bundle date: 24112719; Platform: web; Platform version: undefined

monicaellerose:
I have the same issue on iOS and Firefox and it's annnoying

PR #7041, opened 2024-12-10
Fix composer shortcut detaching a reply

gaearon:
Fixes https://github.com/bluesky-social/social-app/issues/6951

1. Start typing a reply
2. Focus outside the text field
3. Press N

Before, it would detach the reply.

After the fix, it doesn't.

The fix itself is a bit more generic — basically, never auto-dismiss a reply in progress without a manual user action. This might apply to other cases (intents?) but I think the principle still holds. Dismiss the composer manually first, then something new can appear in there. If some piece of code wants to intentionally close the composer, it can call `closeComposer`.

## Before

https://github.com/user-attachments/assets/5f65d75a-56b7-4945-b92e-63f2e84c19a2

## After

https://github.com/user-attachments/assets/66d76769-2232-4d9d-af1c-63fc9e9cecd3

### DIFF
--- a/src/state/shell/composer/index.tsx
+++ b/src/state/shell/composer/index.tsx
@@ -83,7 +83,13 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         'exclamation-circle',
       )
     } else {
-      setState(opts)
+      setState(prevOpts => {
+        if (prevOpts) {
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Oh very nice, this sounds like a great fix imo

PR #7042, opened 2024-12-10
Fix iOS build

gaearon:
It got borked by #6993. I thought the change was web-only but it wasn't.

Using the fix from https://github.com/expo/expo/issues/4626#issuecomment-506563568. Seems to work.

iOS and Android build fine. Web not affected by this change.

## Before

Screenshot 2024-12-10 at 22 04 57

## After

Screenshot 2024-12-10 at 22 09 13

### DIFF
--- a/metro.config.js
+++ b/metro.config.js
@@ -14,6 +14,8 @@ if (process.env.BSKY_PROFILE) {
   cfg.cacheVersion += ':PROFILE'
 }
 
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):

tunetheweb:
Apologies and thanks for fixing my mistake!

Issue #6354, opened 2024-11-15
punycode failures, you can't @ a punycode handle, and handles render incorrectly


blowdart:
### Steps to Reproduce

Someone has registered a handle with a punycode domain -

bsky-bsky-bsky.p-256.移动 which, correctly you are mapping underneath to bsky-bsky-bsky.p-256.xn--6frz82g

However, a couple of problems occur in the web app.

1. View that profile (https://bsky.app/profile/bsky-bsky-bsky.p-256.xn--6frz82g)
2. Now @ them, either via the compose button on their profile, or by following them and then using the search ahead (which works)
3. Add some more text
4. Create the post
5. Marvel as you get a post but the @ is not active. There are no underlying facets in the post record (e.g. [at://did:plc:hfgp6pj3akhqxntgqwramlbg/app.bsky.feed.post/3laxkj2ww422c](https://atproto-browser.vercel.app/at/did:plc:hfgp6pj3akhqxntgqwramlbg/app.bsky.feed.post/3laxkj2ww422c))
6. You're displaying the punycode handle on their profile page and search ahead. You should probably convert it back.

![image](https://github.com/user-attachments/assets/3568e548-4f3c-4e0e-9ec5-fc1806922c5c)

### What platform(s) does this occur on?

iOS, Web (Desktop)

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 7731136 (prod); Bundle date: 24110818; Platform: web

pelifix:
I can confirm, same problem here.

nottucks:
Same problem for my handle. EX: https://pdsls.dev/at://did:plc:yx4aoiqhb73zvm3nykyzqa5e/app.bsky.feed.post/3me7vbcljjs2k

PR #7043, opened 2024-12-10
Add support for IDN (punycode) handles

drash-course:
Currently, all @handle.tld are displayed in their ASCII form. This is fine for most domains, but some TLDs accept IDN domains and it'd be nice to support them fully. I found at least one other issue https://github.com/bluesky-social/social-app/issues/6354 discussing this.

Of course, this opens the can of worms that are [IDN homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack).

I'm implementing a strategy similar to Firefox and Chrome : detect what "scripts" are being used in the IDN domain label and forbid mixing problematic scripts. In those cases, the punycode (ASCII) version of the label is shown instead. I've also added detection for the uppercase i, which can be mistaken for the lowercase l in many fonts. There are certainly more to add.

The security of this feature can't be taken lightly, and I hope this PR can create a discussion to progress this issue forward.


mary-ext:
this is really cool, are you basing this off Chrome/Firefox's exact behavior or just the general idea of what they're doing?

drash-course:
> this is really cool, are you basing this off Chrome/Firefox's exact behavior or just the general idea of what they're doing?

Just the general idea.
I've read https://bugzilla.mozilla.org/show_bug.cgi?id=1332714#c17 to understand the attack, and what Firefox considers their responsibility to filter.

Arguably it's the responsibility of Registrars to make sure people are not registering domains that are confusable. But in practice, those domains can be registered. So the status quo is that it's up to browsers and apps to carefully sanitize IDN domains and minimize the confusion potential with ad-hoc filters.

surfdude29:
It might be a good idea to open up a discussion about this in the atproto repo:

https://github.com/bluesky-social/atproto/discussions

drash-course:
> It might be a good idea to open up a discussion about this in the atproto repo:
> 
> https://github.com/bluesky-social/atproto/discussions

I'm reading the [AT Proto Handle spec](https://atproto.com/specs/handle#usage-and-implementation-guidelines) and it seems they already took a position on IDN handles.

> Internationalized Domain Names ("IDN", or "punycode") are not directly relevant to the low-level handle syntax. In their encoded form, IDNs are already valid hostnames, and thus valid handles. Such handles must be stored and transmitted in encoded ASCII form. Handles that "look like" IDNs, but do not parse as valid IDNs, are valid handles, just as they are valid hostnames. Applications may, optionally, parse and display IDN handles as Unicode.

@surfdude29 Do you think ATProto needs to standardize the filters apps should use to avoid IDN homograph attacks? Maybe ATProto can standardize what the security goals are. On the other hand, standardizing a specific implementation sounds unwise when you want to be able to react to novel attacks with new code / checks.

surfdude29:
@drash-course I didn't know about that part of the ATProto handle spec tbh, so I may well have been over-cautious in suggesting it would be a good idea to discuss it over at the atproto repository as well.

However, I think you have raised some excellent points in your post there, and I hope that others who know a lot more about it than me (including Bluesky team members like @bnewbold) see it and are able to give feedback.

drash-course:
I'm working on a more standard sanitization algo that follows https://wiki.mozilla.org/IDN_Display_Algorithm, and I'll add tests. I'll update this PR when it's ready, hopefully in a few days.

PR #7055, opened 2024-12-11
Remove rounding mode hack

mozzius:
https://github.com/formatjs/formatjs/issues/4678 was fixed upstream, so we can remove the hack

English:
Screenshot 2024-12-11 at 18 27 20

Japanese: 
Screenshot 2024-12-11 at 18 31 38

# Test plan
Round some numbers in various languages, and compare to the browser version


gaearon (approved):
lg

gaearon:
We can just delete those tests

PR #7056, opened 2024-12-11
New profile feed header

estrattonbailey:
Medium-sized step towards some cleaner UI for this screen. More work will be done here.

This PR aims to keep translation strings mostly the same, but some were added.

CleanShot 2024-12-11 at 14 27 53@2x
CleanShot 2024-12-11 at 14 27 56@2x
CleanShot 2024-12-11 at 14 28 01@2x
CleanShot 2024-12-11 at 14 28 03@2x


src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -0,0 +1,530 @@
...
+          {
+            ...savedFeedConfig,
+            pinned,
+          },
+        ])
+
+        if (pinned) {
+          Toast.show(_(msg`Pinned ${info.displayName} to Home`))
+        } else {
+          Toast.show(_(msg`Un-pinned ${info.displayName} from Home`))
```
surfdude29, line 129:
```suggestion
          Toast.show(_(msg`Unpinned ${info.displayName} from Home`))
```
I think it would work better without the hyphen, as currently all the `unpin*` strings in the app don't use one

src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -0,0 +1,530 @@
...
+            </Button>
+          </View>
+
+          <View style={[a.pt_xs, a.gap_lg]}>
+            <Divider />
+
+            <View
+              style={[a.flex_row, a.align_center, a.gap_sm, a.justify_between]}>
+              <Text style={[a.italic, t.atoms.text_contrast_medium]}>
+                Something wrong? Let us know.
```
surfdude29, line 501:
```suggestion
                <Trans>Something wrong? Let us know.</Trans>
```

surfdude29 (commented):
A couple of tiny suggestions 👀

src/screens/Profile/components/ProfileFeedHeader.tsx
surfdude29: Also, just wondering, will the text be able to wrap onto another line here at larger text sizes and in non-English localizations?

src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -0,0 +1,530 @@
...
+        <View style={[a.flex_1, a.gap_2xs]}>
+          <Text
+            style={[a.text_2xl, a.font_heavy, a.leading_tight]}
+            numberOfLines={2}>
+            {info.displayName}
+          </Text>
+          <Text
+            style={[a.text_sm, a.leading_tight, t.atoms.text_contrast_medium]}
+            numberOfLines={1}>
+            <Trans>By</Trans>{' '}
```
surfdude29, line 421: Also, I'm wondering if it might be desirable (or indeed feasible) to restructure this, so that `By` and the handle are both in the string to be translated, for languages where some flexibility is needed? 🤔

gaearon: Yes, `Trans` should be pulled outside here

gaearon (approved):

PR #7058, opened 2024-12-11
Reduce excessive empty space on profile with no bio

gaearon:
We had a structure like this:

- Parent view with `paddingBottom`
  - Metrics with `paddingBottom`
  - Conditional bio with no padding
  - Conditional social proof with `paddingTop`
 - Tabs 

As a result, when something conditional is missing, the padding were getting combined in an awkward way. For example:

- No bio or social proof: The metrics' `paddingBottom` was getting combined with the parent view's `paddingBottom`, resulting in double padding
- No bio, but has social proof: The metrics' `paddingBottom` was getting combined with the social proof's `paddingTop`, resulting in double padding

The fix is to treat these as three items with a gap, and to remove those paddings.

## Test Plan

Check that these cases now work well:

- No bio, no social proof (this used to be too spacey)
- No bio but has social proof (this used to be too spacey)

Check that if there _is_ a bio, layout hasn't changed.

## Before

Screenshot 2024-12-11 at 20 14 09

## After

Screenshot 2024-12-11 at 20 14 03

### DIFF
--- a/src/screens/Profile/Header/Metrics.tsx
+++ b/src/screens/Profile/Header/Metrics.tsx
@@ -30,7 +30,7 @@ export function ProfileHeaderMetrics({
 
   return (
     <View
-      style={[a.flex_row, a.gap_sm, a.align_center, a.pb_md]}
+      style={[a.flex_row, a.gap_sm, a.align_center]}
       pointerEvents="box-none">
       <InlineLinkText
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Didn't test, code lgtm, gap ftw

PR #7066, opened 2024-12-12
Make the pager take full width

gaearon:
This changes the pager to always fill the entire screen width by spacing out the items. If there are _many_ items (more than a screen's worth), then they are placed like before.

Easier to understand from a demo:

https://github.com/user-attachments/assets/7c28141f-c2c1-4be0-9b61-61c48a24f5c8

## Test Plan

Try these on iOS, Android, and web.

- Check Home the new user experience (Discover, Following)
- Check Home for a user with a lot of feeds
- Check profile tabs (own and others')
- Check hashtag and search screens
- Try it on top of https://github.com/bluesky-social/social-app/pull/7044

Centered item looks a bit weird when there's a single tab but the only case of that is being solved by https://github.com/bluesky-social/social-app/pull/7056.

### DIFF
--- a/src/screens/Hashtag.tsx
+++ b/src/screens/Hashtag.tsx
@@ -107,7 +107,7 @@ export default function HashtagScreen({
 
   return (
     <Layout.Screen>
-      <Layout.Header.Outer>
+      <Layout.Header.Outer noBottomBorder>
         <Layout.Header.BackButton />
         <Layout.Header.Content>
           <Layout.Header.TitleText>{headerTitle}</Layout.Header.TitleText>
--- a/src/view/com/pager/TabBar.tsx
+++ b/src/view/com/pager/TabBar.tsx
@@ -287,7 +287,7 @@ export function TabBar({
           onLayout={e => {
             contentSize.set(e.nativeEvent.layout.width)
           }}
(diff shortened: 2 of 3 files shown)


mozzius (approved):
Looks good, in stylistic terms I think making the blue indicator a little bit wider than the text would be nice

gaearon:
I'll get this in as is but can iterate

auroursa:
It’s possible that the shorter length of Chinese text caused the strange width gap that appeared after this PR. Sorry for the Chinese in the screenshot, but I wanted to more cleared show this change.

| Before | After |
| ----------- | ----------- |
|  |  |

gaearon:
Would it be desirable to remove the minimum width for Chinese characters?

auroursa:
Worth a try, this should not cause any other issues. It seems that Korean and Hindi also have the same issue, as their words are quite short. Perhaps we should reconsider the minimum width for this layout?

gaearon:
Maybe instead of minimum width we could do something else in styling? The intention is to avoid the line being _too_ short when it’s just a few items, like “All” and “Mentions” making “All” very short.

gaearon:
Fixed in https://github.com/bluesky-social/social-app/pull/7095

PR #7067, opened 2024-12-12
Make sure local web builds set debug environment flag

gaearon:
We shouldn't be relying on a developer copying the `.env.example` file into `.env` — so far we haven't needed that. The defaults just need to be correct.

Previously, if your `EXPO_PUBLIC_ENV` isn't explicitly set to `development`, you got a weird setup where `IS_DEV` is `false` in dev environment. With this change, if `EXPO_PUBLIC_ENV` isn't set, it's _assumed_ to be dev environment.

## Test Plan

Read the code. Verify that deleting `.env` still recognizes web localhost as being a dev environment. We're assuming that all production build systems already always set `EXPO_PUBLIC_ENV` since otherwise it wouldn't get set to `production` before either.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,6 @@
 # Copy this to `.env` and `.env.test` files
 
 SENTRY_AUTH_TOKEN=
-EXPO_PUBLIC_ENV=development
 EXPO_PUBLIC_LOG_LEVEL=debug
 EXPO_PUBLIC_LOG_DEBUG=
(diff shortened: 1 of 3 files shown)


haileyok (approved):

gaearon:
Eh, apparently https://bsky.app/ in prod has `EXPO_PUBLIC_ENV` set to `undefined` so the logic is wonky. We need to rethink this a bit more.

PR #7075, opened 2024-12-12
Add toki pona (tok) to content languages

itsvic-dev:
[**toki pona**](https://tokipona.org/) is a constructed language created by Sonja Lang in 2001. It is considered by some to have the second biggest conlang community online.

toki pona's ISO 639-3 code is `tok`. It does not have an ISO 639-1 code.

Bluesky already supports other constructed languages (Esperanto, Interlingua, Ido), so this feels like a pretty obvious addition. I'm not sure if this would need any extra support elsewhere.

### DIFF
--- a/src/locale/languages.ts
+++ b/src/locale/languages.ts
@@ -558,6 +558,7 @@ export const LANGUAGES: Language[] = [
   {code3: 'tli', code2: '', name: 'Tlingit'},
   {code3: 'tmh', code2: '', name: 'Tamashek'},
   {code3: 'tog', code2: '', name: 'Tonga (Nyasa)'},
+  {code3: 'tok', code2: '', name: 'toki pona'},
   {code3: 'ton', code2: 'to', name: 'Tonga (Tonga Islands)'},
   {code3: 'tpi', code2: '', name: 'Tok Pisin'},
   {code3: 'tsi', code2: '', name: 'Tsimshian'},


surfdude29:
This has been proposed several times before:
- #3204
- #5998
- #6290
- #6511

The issue is not that toki pona is a constructed language, but rather that the Bluesky app and, as I understand it, the Bluesky lexicon, only support two-letter ISO 639-1 codes for the language fields in posts, as @auroursa [explains here](https://github.com/bluesky-social/social-app/pull/6290#issuecomment-2472906386).

This limitation also affects other languages, such as Asturian (`ast`).

auroursa:
Supporting ISO 639-3 is definitely a step in the right direction. Many widely used languages are currently excluded simply because they lack ISO 639-1 codes. I’m looking forward to #6769 addressing this issue.

BTW I noticed [similar discussions](https://github.com/mastodon/mastodon/issues/18538) on Mastodon issues, which might be valuable references for developers.

Maybe we don’t need to support all ISO 639-3 languages but should focus on including a few highly requested ones (e.g., Toki Pona) to keep the Post Language List manageable and user-friendly.

As the maintainer of Cantonese localization, I naturally hope it can be used as a Post Language in the future (#1766). Currently, it faces the same issue as Toki Pona. :(

PR #6939, opened 2024-12-04
further increase target area of scrubber

SarcevicAntonio:
18px height -> 32px height.
[still](https://github.com/bluesky-social/social-app/pull/5265) too finicky on touchscreens for my fingers + [passes WCAG 2.2 AA](https://www.w3.org/WAI/WCAG22/quickref/?currentsidebar=%23col_overview&showtechniques=255#target-size-minimum) now 👍🏼

PR #7081, opened 2024-12-12
Increase scrubber height if touch device

mozzius:
Supercedes #6939 

Increases scrubber height if touch device, since 18px is too small for touchscreens


gaearon (approved):
didn't test, assume it looks the same?

PR #7088, opened 2024-12-13
Gate bitdrift integration

gaearon:
- Moves Statsig initialization much earlier — we'll now initialize unconditionally _before_ we know your user ID and then initialize again. In practice this means that we're able to check cached gates based on StableID almost on startup.
- Makes the Bitdrift integration conditional on an `enable_bitdrift` gate to derisk it.

Our test setup was getting borked by the changes in the init code. I tried to unbork it but something about `env.ts` was not doing it correctly. I ended up ripping out the custom `IS_TEST` thing and doing a raw `process.env.NODE_ENV` inline environment check. I think we should change the rest of the code to match that pattern (for `IS_DEV` and `IS_PROD`) to guarantee dead code elimination but that's out of scope of this PR.

## Test Plan

Verify that allowlisting your StableID for the `enable_bitdrift` gate leads to getting into the conditional block on next app reload. This should work whether you're logged in or not. No effect on web.

Verify that gates work in general.

### DIFF
--- a/src/lib/statsig/statsig.tsx
+++ b/src/lib/statsig/statsig.tsx
@@ -16,6 +16,8 @@ import {Gate} from './gates'
 
 const SDK_KEY = 'client-SXJakO39w9vIhl3D44u8UupyzFl4oZ2qPIkjwcvuPsV'
 
+export const initPromise = initialize()
+
 type StatsigUser = {
   userID: string | undefined
   // TODO: Remove when enough users have custom.platform:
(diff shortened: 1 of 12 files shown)


src/lib/statsig/statsig.tsx
```diff
@@ -16,6 +16,8 @@ import {Gate} from './gates'
 
 const SDK_KEY = 'client-SXJakO39w9vIhl3D44u8UupyzFl4oZ2qPIkjwcvuPsV'
 
+export const initPromise = initialize()
```
gaearon, line 19: The initialization now triggers here.

estrattonbailey (approved):

PR #7089, opened 2024-12-13
Remove the environment indirections

gaearon:
Stacked on https://github.com/bluesky-social/social-app/pull/7088

Supersedes https://github.com/bluesky-social/social-app/pull/7067

---

We have a bunch of custom indirection for environment checks like "is dev", "is prod", etc. This indirection is somewhat bothersome. It re-exports some of the underlying primitives but it's forked between platforms. Sometimes it's mocked. There are different conflicting sources of truth (`EXPO_PUBLIC_ENV`, `__DEV__`, `NODE_ENV`).

In https://github.com/bluesky-social/social-app/pull/7088 I've gotten rid of our custom `IS_TEST` in favor of `process.NODE_ENV === 'test'`.
In this PR, I'm doing the same for `IS_DEV` and `IS_PROD`.

Here's how to think about the change:

- If you want to know whether you're in development or production-ish mode, use `__DEV__`. It already works everywhere. It corresponds to development mode in React, and `if (!__DEV__)` on web gets dead-code eliminated by the bundler. Under the hood, the `__DEV__` constant always matches `production.env.NODE_ENV !== 'production'`.
- The `app.config.ts` file depends on `EXPO_PUBLIC_ENV` because it's running during the build (before the build, actually) and it's really about the Expo channel. We could maybe rename `EXPO_PUBLIC_ENV` to `EXPO_PUBLIC_CHANNEL`  to be clearer but it's fine as is. The app code should generally not be reading this aside from occasional "is TestFlight" check.

## Test Plan

Verify this fixes the same issue as https://github.com/bluesky-social/social-app/pull/7067. Namely, `localhost` now always shows feed debug segments in Discover without setting any env variables. (It will, however, *not* show them on `main.bsky.dev` because `__DEV__` is `false` there — which the fix in https://github.com/bluesky-social/social-app/pull/7067 would've regressed on.)

Scan through the code, confirm it makes sense. Make local iOS debug and release builds, confirm you get dev/prod in settings and that the debugging feed context is appropriately shown/hidden even for non-allowlisted accounts.


.env.example
```diff
@@ -2,7 +2,6 @@
 
 BITDRIFT_API_KEY=
 SENTRY_AUTH_TOKEN=
-EXPO_PUBLIC_ENV=development
```
gaearon, line 5: This will no longer be required to set. We can still set it for testflight and prod, otherwise it's assumed to be dev.

app.config.js
```diff
@@ -15,9 +15,9 @@ module.exports = function (config) {
    */
   const PLATFORM = process.env.EAS_BUILD_PLATFORM
 
-  const IS_DEV = process.env.EXPO_PUBLIC_ENV === 'development'
   const IS_TESTFLIGHT = process.env.EXPO_PUBLIC_ENV === 'testflight'
   const IS_PRODUCTION = process.env.EXPO_PUBLIC_ENV === 'production'
+  const IS_DEV = !IS_TESTFLIGHT || !IS_PRODUCTION
```
gaearon, line 20: This is now implied.

src/components/Typography.tsx
```diff
@@ -31,7 +30,7 @@ export function Text({
     flags,
   })
 
-  if (IS_DEV) {
+  if (__DEV__) {
```
gaearon, line 33: In the app code, you almost always want `__DEV__`.

src/lib/app-info.ts
```diff
@@ -1,9 +1,7 @@
 import {nativeApplicationVersion, nativeBuildVersion} from 'expo-application'
 
-export const BUILD_ENV = process.env.EXPO_PUBLIC_ENV
-export const IS_DEV = process.env.EXPO_PUBLIC_ENV === 'development'
 export const IS_TESTFLIGHT = process.env.EXPO_PUBLIC_ENV === 'testflight'
-export const IS_INTERNAL = IS_DEV || IS_TESTFLIGHT
+export const IS_INTERNAL = __DEV__ || IS_TESTFLIGHT
```
gaearon, line 4: These stay. They don't describe the JS environment, they describe the release channel.

src/lib/app-info.web.ts
```diff
@@ -1,9 +1,7 @@
 import {version} from '../../package.json'
 
-export const BUILD_ENV = process.env.EXPO_PUBLIC_ENV
-export const IS_DEV = process.env.EXPO_PUBLIC_ENV === 'development'
 export const IS_TESTFLIGHT = false
-export const IS_INTERNAL = IS_DEV
+export const IS_INTERNAL = __DEV__
```
gaearon, line 4: Web has no concept of internal builds so we only treat localhost as internal.

estrattonbailey (approved):
LGTM!

PR #7094, opened 2024-12-13
Uppercase the auth factor token input value to match the email casing.

BradEstey:
The 2fa token is uppercase in the email sent to users. This PR just adds `text-transform: uppercase` to the input element on the app to match the casing in the email. 

A boolean state is needed to only set the input to uppercase when there is an input, otherwise the placeholder gets uppercased as well.

### DIFF
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -60,6 +60,8 @@ export const LoginForm = ({
   const [isProcessing, setIsProcessing] = useState<boolean>(false)
   const [isAuthFactorTokenNeeded, setIsAuthFactorTokenNeeded] =
     useState<boolean>(false)
+  const [isAuthFactorTokenValueEmpty, setIsAuthFactorTokenValueEmpty] =
+    useState<boolean>(true)
   const identifierValueRef = useRef<string>(initialHandle || '')
   const passwordValueRef = useRef<string>('')
   const authFactorTokenValueRef = useRef<string>('')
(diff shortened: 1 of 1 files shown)


gaearon (approved):
makes sense

damz:
This doesn't seem to do anything on Android AND unfortunately it causes a refresh of the `<TextInput>` that makes React Native swallow the first input.

BradEstey:
I think the `isAuthFactorTokenValueEmpty` state can be removed entirely and just adding a `::placeholder` pseudo-element setting the text placeholder text to not uppercase should fix the rerender.

BradEstey:
Scratch that, it looks like React Native doesn't support `::placeholder` :(

damz:
It's a bit of an odd one: with the full application I can reproduce that changing the `textDecoration` or `color` style makes the `<TextInput>` lose its value. It's not even a race condition, it happens even after a long `setTimeout` inside the `onChangeText` handler. That would point to a bug in `react-native`.

But I cannot seem reproduce the behavior outside of the application at all, even with the same exact version of `react-native`.

damz:
> But I cannot seem reproduce the behavior outside of the application at all, even with the same exact version of react-native.

Ok, I tracked this down to `newArchEnabled`:

* With `newArchEnabled: false`, rerendering a `<TextInput>` component with some text-affecting style changed (I confirmed with `textTransform` and `color`) causes the component to lose its state.
* With `newArchEnabled: true`, the problem doesn't occur

damz:
Given that this PR only ever worked on the web, and caused the problem on native (at least on Android, I haven't been able to confirm this on iOS), I suggest we revert it @BradEstey.

PR #7095, opened 2024-12-13
Give minWidth to tab indicator only

gaearon:
Fixes https://github.com/bluesky-social/social-app/pull/7066#issuecomment-2540409035

Instead of giving min width to the tab item itself, instead give it to the indicator line. Also constrain it by the parent size so that it never extends beyond the tab item. As a result, the blue line will be 45px min, but won't inflate the tab bar item.

## Test Plan

Verify "All" in notifications still has some extra minimum width.

Verify profile tab bar items are not inflated anymore.

https://github.com/user-attachments/assets/58a6b313-588c-47a1-877a-195cb9edc4e3

https://github.com/user-attachments/assets/65f277b3-d307-439b-9327-11a9584092c3

Screenshot 2024-12-13 at 17 07 43

### DIFF
--- a/src/view/com/pager/TabBar.tsx
+++ b/src/view/com/pager/TabBar.tsx
@@ -237,12 +237,25 @@ export function TabBar({
         opacity: 0,
       }
     }
+
+    function getScaleX(index: number) {
+      const textWidth = textLayoutsValue[index].width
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Test on web/ios, lgtm!

PR #7096, opened 2024-12-13
Allow non-graphic emoji to inherit parent color

estrattonbailey:
iOS only thing

Before
CleanShot 2024-12-13 at 11 36 18@2x
After
CleanShot 2024-12-13 at 11 37 03@2x


gaearon (approved):
looks fine to me, but why was `black` there in the first place? seems like it was added in https://github.com/bluesky-social/social-app/commit/5eb294488f08534abac3335acfa366cffea9259e?

PR #7100, opened 2024-12-13
Tweak ProfileList design

gaearon:
This makes List and Moderation List screens closer in design to the Starter Pack screen:

- The description is moved inline into the header.
  - Also killed the "No description" label. If it has no description, it has no description, duh.
- The "About" tab is now just called "People" (for Lists) because it is now just a list of people.
- For mod lists, there's no tab header at all to avoid showing a single tab (which looked worse since #7066).

[Review without whitespace](https://github.com/bluesky-social/social-app/pull/7100/files?w=1)

## Test Plan

- Check lists, mod lists, starter packs
- Check lists with description and with no description

https://github.com/user-attachments/assets/0aa0b851-c50e-489e-a64d-495d1cb3bbe1

https://github.com/user-attachments/assets/952ff263-d063-40dd-ae1f-34f1d59efb9a

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -21,6 +21,7 @@ import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {ScrollProvider} from '#/lib/ScrollContext'
 import {isIOS} from '#/platform/detection'
 import {Pager, PagerRef, RenderTabBarFnProps} from '#/view/com/pager/Pager'
+import {useTheme} from '#/alf'
 import {ListMethods} from '../util/List'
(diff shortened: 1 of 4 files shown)


src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -277,7 +279,7 @@ let PagerTabBar = ({
   return (
     <Animated.View
       pointerEvents={isIOS ? 'auto' : 'box-none'}
-      style={[styles.tabBarMobile, headerTransform]}>
+      style={[styles.tabBarMobile, headerTransform, t.atoms.bg]}>
```
gaearon, line 282: i was noticing scrollview content sometimes appearing under the header, this makes the header always a solid color

estrattonbailey (approved):
Nice one

PR #7102, opened 2024-12-13
Fix z-indexes to make tabbar scroll on Android

gaearon:
Reverts this change: https://github.com/bluesky-social/social-app/pull/7044/commits/b3381b5180e5eff33342dde009d0918056e80df2#diff-65932679e7a40da6746980329ec9b70b756d28bfc7980c1ffe03c2e49f7a5bb3

I did it to make sure the tabbar is always drawn on top of the content (so the Notifications glimmer doesn't obscure the tabbar's bottom border) but this zIndex being there causes Android to break horizontal scroll on the tabbar itself (it is interpreted as a pager swipe).

This alternative set of changes (just plumbing zIndex through) seems to work.

## Test Plan

Dragging the tabbar works again on Android:

https://github.com/user-attachments/assets/b6a101aa-47b6-45ec-8b5a-5f171d82ffe5

On iOS, while Notifications are loading, the border is visible:

https://github.com/user-attachments/assets/c790ce25-b3b8-423c-9d01-741753b33f43

Without the zIndex hackery, it would not be visible.

Web works like before.

### DIFF
--- a/src/screens/Hashtag.tsx
+++ b/src/screens/Hashtag.tsx
@@ -134,7 +134,7 @@ export default function HashtagScreen({
       <Pager
         onPageSelected={onPageSelected}
         renderTabBar={props => (
-          <Layout.Center style={web([a.sticky, a.z_10, {top: 0}])}>
+          <Layout.Center style={[a.z_10, web([a.sticky, {top: 0}])]}>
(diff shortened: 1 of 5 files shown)


estrattonbailey (approved):
Works on my android device 👍

PR #7103, opened 2024-12-13
Add some helper text for custom domains

estrattonbailey:
CleanShot 2024-12-13 at 16 50 40@2x
CleanShot 2024-12-13 at 16 50 51@2x

### DIFF
--- a/src/screens/Settings/components/ChangeHandleDialog.tsx
+++ b/src/screens/Settings/components/ChangeHandleDialog.tsx
@@ -17,6 +17,7 @@ import {useMutation, useQueryClient} from '@tanstack/react-query'
 
 import {HITSLOP_10} from '#/lib/constants'
 import {cleanError} from '#/lib/strings/errors'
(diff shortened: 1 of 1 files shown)


src/screens/Settings/components/ChangeHandleDialog.tsx
```diff
@@ -488,6 +491,18 @@ function OwnHandlePage({goToServiceHandle}: {goToServiceHandle: () => void}) {
         </Animated.View>
       )}
       <Animated.View layout={native(LinearTransition)}>
+        {currentAccount?.handle?.includes('bsky.social') && (
```
gaearon, line 494: kind of a nit but this should probably be `endsWith`

gaearon (approved):
lgtm, small nit

src/screens/Settings/components/ChangeHandleDialog.tsx
estrattonbailey: ope yeah let's do that 👍

Issue #7097, opened 2024-12-13
Image upload broken on main


benharri:
### Steps to Reproduce

1. Try to post an image on https://main.bsky.dev or a local build
2. It fails with the message "Unable to compress image" or "Invalid uploadBlob input: string" depending on the size of the image file.

### Attachments

https://bsky.app/profile/benharri.org/post/3ld2bx43ips2l

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

firefox and chromium

### What version of the app are you using?

add7a91d5

### Additional Information

https://github.com/bluesky-social/social-app/blob/356dad1932c3404f581b747127ae251dbe165bb3/src/state/gallery.ts#L209-L246

benharri:
I did a bit of digging and logged the compressed size in each loop iteration and it's not changing at all with the new `factor`.

Relatedly, `manipulateAsync()` is marked as deprecated in expo-image-manipulator.

gaearon:
Any ideas when/how this broke?

benharri:
First thought is RN 0.76 but I haven't done a bisect yet.

gaearon:
Could be this commit, easy way to test is maybe to downgrade this package alone to previous version.

https://github.com/bluesky-social/social-app/blame/main/package.json#L134

gaearon:
We're pretty slammed right now with feature work, would appreciate if you could look into a fix.

benharri:
Initial testing looks good, I'll do a PR

gaearon:
Well I'm not sure we can just downgrade it safely, so I think we'd have to use the new API? https://github.com/expo/expo/pull/30194

benharri:
Oh right I noticed the deprecation warning in there. I will have a look at switching the usage to the new API as well

gaearon:
Would be good to understand what exactly broke,

gaearon:
Upgrading the API doesn't help. https://github.com/bluesky-social/social-app/pull/7104

gaearon:
https://github.com/bluesky-social/social-app/pull/7105

benharri:
Just got to the same spot as #7104 myself and found the `blob:` URLs. Will test the changes in #7105

PR #7105, opened 2024-12-13
Fix posting images on web

gaearon:
Fixes https://github.com/bluesky-social/social-app/issues/7097

If this isn't enough or we find some other issue, we'll fork the package by version instead, but let's try this first.

There were two issues:

- The library we use now emits `blob:` URIs rather than `data:` URIs
- It also accidentally lost support for the `compress` option on web, so it became a no-op

So this fixes both problems.

## Test Plan

Verify you can upload both _small_ images and _large_ images.

For small image, just use something small.

For large image, I was testing with https://apod.nasa.gov/apod/image/2412/M51_HaLRGB_APOD2048.jpg

No changes on native.

### DIFF
--- a/patches/expo-image-manipulator+13.0.5.patch
+++ b/patches/expo-image-manipulator+13.0.5.patch
@@ -25,7 +25,7 @@ index 120d8d3..f8aa49c 100644
  
  export default ExpoImageManipulator.Context as typeof ImageManipulatorContext;
 diff --git a/node_modules/expo-image-manipulator/src/web/ImageManipulatorContext.web.ts b/node_modules/expo-image-manipulator/src/web/ImageManipulatorContext.web.ts
-index 428848c..a851332 100644
+index 428848c..363a57a 100644
 --- a/node_modules/expo-image-manipulator/src/web/ImageManipulatorContext.web.ts
 +++ b/node_modules/expo-image-manipulator/src/web/ImageManipulatorContext.web.ts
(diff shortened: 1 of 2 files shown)


patches/expo-image-manipulator+13.0.5.patch
```diff
@@ -0,0 +1,48 @@
...
++  async renderAsync(compress?: number): Promise<ImageManipulatorImageRef> {
+     const canvas = await this.currentTask;
+ 
+     return new Promise((resolve) => {
+@@ -49,7 +49,7 @@ export default class ImageManipulatorContext extends SharedObject {
+         const url = blob ? URL.createObjectURL(blob) : canvas.toDataURL();
+ 
+         resolve(new ImageManipulatorImageRef(url, canvas.width, canvas.height));
+-      });
++      }, 'image/jpeg', compress);
```
gaearon, line 45: This is OK to hardcode in our patch because all callsites that specify `compress` also specify JPG as a format. Upstream they may want to be more judicious about this

gaearon:
not sure why the webpack analyzer task is failing but the patch applies fine to me locally. i think it's a false positive

benharri:
Looks good

surfdude29:
Works great for me on the preview branch 👍

gaearon:
Gonna get this in before the weekend

TUNISIA-user:
> Gonna get this in before the weekend
hahha coding is the weekend :)

TUNISIA-user (commented):
x

PR #7109, opened 2024-12-14
feat: use animation for entry of LoadLatestBtn

angelo-hub:
This adds a renimated animation for the displaying of the LoadLatestBtn, looks a bit cleaner than the current implementation where it can look jarring in the way it shows up

Video here:
https://github.com/user-attachments/assets/3967774d-e9f7-48cc-9971-bace375d0bdf


gaearon:
Tbh we'd just like to get rid of it, see https://github.com/bluesky-social/social-app/pull/7120

angelo-hub:
Thanks for the visibility I'll close it out!

Issue #7111, opened 2024-12-14
Left part of display name can be cut off for some accounts in Recent Searches history


surfdude29:
### Steps to Reproduce

1. Tap **Search** button in the bottom bar
2. Type in enough characters of **@atproto.com** so that you see the account in the list of suggestions
3. Tap the **AT Protocol Developers** account
4. Tap the **Search** button in the bottom bar twice
5. Observe that the beginning of the display name is cut off, as shown in the screenshot below

### Attachments

![IMG_9840](https://github.com/user-attachments/assets/e84e660b-fd76-40a5-a1ae-6003882eaf6b)

### What platform(s) does this occur on?

iOS

### Device Info

iPhone SE (3rd generation)

### What version of the app are you using?

Build version: 1.95.0.634; Bundle info: 41243e874 (prod); Bundle date: 24120701; Platform: ios; Platform version: 17.7.2

karolstawowski:
I managed to reproduce the problem on desktop web browsers Chrome 131.0.6778.140 and Firefox 133.0.3. The problem appears for screens under 800px width - `Mobile` breakpoint. I will prepare a PR with fix proposal

surfdude29:
Confirmed that this is fixed for me now in:

Build version: 1.96.4.818; Bundle info: 615bbc3 (prod); Bundle date: 25010300; Platform: ios; Platform version: 17.7.2

Thanks @karolstawowski! 🙏

PR #7119, opened 2024-12-15
feat: widen recent search profile link for mobile devices

karolstawowski:
Hi,

In response to issue #7111, I prepared a solution to set width of profile links' text in "Recent Searches" to 100% for mobile devices. Change does not affect current profiles layout. Below I place before and after screenshots.

I tested it on Chrome 131.0.6778.140 and Firefox 133.0.3 as problem appeared on web version for screens with width smaller than 900px.

Before:
chrome_3qxhet01Vr

After:
chrome_AeaqRlNvNM

Greetings
Karol

### DIFF
--- a/src/view/screens/Search/Search.tsx
+++ b/src/view/screens/Search/Search.tsx
@@ -1058,10 +1058,7 @@ function SearchHistory({
                     asAnchor
                     anchorNoUnderline
                     onBeforePress={() => onProfileClick(profile)}
-                    style={[
-                      styles.profilePressable,
-                      isMobile && styles.profilePressableMobile,
-                    ]}>
+                    style={styles.profilePressable}>
                     <Image
                       source={{uri: profile.avatar}}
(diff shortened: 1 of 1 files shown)


src/view/screens/Search/Search.tsx
```diff
@@ -1058,7 +1058,10 @@ function SearchHistory({
                     asAnchor
                     anchorNoUnderline
                     onBeforePress={() => onProfileClick(profile)}
-                    style={styles.profilePressable}>
+                    style={[
+                      styles.profilePressable,
+                      isMobile && styles.profilePressableMobile,
```
gaearon, line 1063: Is there a downside to make this style unconditional?

karolstawowski: It's okay to apply style to all screen sizes. I just pushed the change

gaearon (approved):

Issue #7002, opened 2024-12-07
Setting to hide the "Load more posts" arrow


kyleboas:
### Describe the Feature

Some may find the "Load more posts" arrow to be distracting when scrolling. I do. I don't like the blue dot that pops up when new posts are availble. It makes me want to tap the arrow and stop reading. It would be nice if we could disable the arrow completely with the choice to toggle it on or off in the settings.

### Attachments

![IMG_4213](https://github.com/user-attachments/assets/0236abb1-876f-4e9f-83cf-1d952fb17d3f)

### Describe Alternatives

A toggle in the settings to show or hide the arrow at the bottom left hand side when scrolling. 

If toggled off, users could tap the home button and then load the page manually.

gaearon:
We don't need the setting, we just need to kill this button. Its functionality should be rolled into the navbar/tabbar.

Issue #6902, opened 2024-12-02
Align "Load More Posts" button to the left edge of the main container


nickbrit:
### Steps to Reproduce

1. Go to the home page.
2. Scroll down.
3. Notice that the button is not aligned with the left edge of the main container.

### Attachments

<img width="1128" alt="Screen Shot 2024-12-02 at 8 57 29 AM" src="https://github.com/user-attachments/assets/61329dc7-62dc-4807-ad22-e675d6c93ff6">

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Mac

### What version of the app are you using?

Web

### Additional Information

Currently, the "Load More Posts" button does not align properly with the left edge of the main container.

Make sure the alignment adjusts correctly for different screen sizes and respects the container's padding/margin rules.

khuddite:
@nickbrit It seems your screenshot doesn't quite match your description, which button did you mean by "Load More Posts" button? Did you mean the "Scroll to top" button that has the up arrow icon in it?

nickbrit:
@khuddite yes, I meant this button.

I named it 'Load new posts' because that's what the aria-label says. But yes, it looks like a 'Scroll to top' button. However, it doesn't just scroll to the top; it also updates the content, which makes it confusing.

Issue #6443, opened 2024-11-17
LoadLatestBtn flickers when switching between discover and following when both feeds are scrolled to the top (web)


breauxna:
### Steps to Reproduce

1. Login to the app on desktop
2. Navigate back and forth between `Discover` and `Following` while `hasNew` is false and both feeds are scrolled to the top
3. See the load latest button flicker in the bottom left

I _think_ the `onScrolledDownChange` callback in the `List.web.tsx` component is being called twice on initial render of the feed setting `isScrolledDown` to true then false resulting in the flicker.

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Build version: 1.93.0; Bundle info: 5f449e3 (prod); Bundle date: 24111700; Platform: web

Issue #6319, opened 2024-11-14
Move the scroll up button to the right side?


victor-ponamariov:
### Describe the Feature

It's pretty common pattern to look for the button on the right side. 

I've even noticed it accidentally. 

![CleanShot 2024-11-14 at 04 58 45@2x](https://github.com/user-attachments/assets/36829b54-c8b4-47cc-a384-a9b6a48eb578)

SaySayTakamura:
A toggle would be good for this, leaving it to the user to choose which side they want the button to be on, as some will preffer it on the left side and others might preffer on the right side.

reetp:
On/off would be a help as well.

Forever hitting it accidentally. Grrrrr.

grndkntrl:
If you have the Stylus browser extension installed then you can create a userstyle.css for BS to shift it to wherever you want.

This is the rule I use for shifting it to the very left of the page:

```
button[aria-label="Load new posts"]
{
	left: 7px !important;
	right: auto !important;
	bottom: 7px !important;
}

button[aria-label="Scroll to top"]
{
	left: 7px !important;
	right: auto !important;
	bottom: 7px !important;
}
```

If you want it to move to the right-hand side, then just switch the left/right values.

reetp:
> If you have the Stylus browser extension installed 

I never use desktop.

Mobile only.

Still driving me insane!!

victor-ponamariov:
> If you have the Stylus browser extension installed then you can create a userstyle.css for BS to shift it to wherever you want.
> 
> This is the rule I use for shifting it to the very left of the page:
> 
> ```
> button[aria-label="Load new posts"]
> {
> 	left: 7px !important;
> 	right: auto !important;
> 	bottom: 7px !important;
> }
> 
> button[aria-label="Scroll to top"]
> {
> 	left: 7px !important;
> 	right: auto !important;
> 	bottom: 7px !important;
> }
> ```
> 
> If you want it to move to the right-hand side, then just switch the left/right values.

Well, sure there are many ways to change the view of an app, but Bluesky is used not only by those who knows what CSS is. 
I mean, No hard feelings, your solution will work.

I just proposed an idea when I noticed non-typical placement, maybe it'd better to move it to the right or do something else.

PR #7120, opened 2024-12-15
[Experiment] Remove "Load Latest" button

gaearon:
Fixes (eventually) https://github.com/bluesky-social/social-app/issues/7002, https://github.com/bluesky-social/social-app/issues/6902, https://github.com/bluesky-social/social-app/issues/6319, https://github.com/bluesky-social/social-app/issues/6443

In this experiment, I'm removing the "Load latest" floating button.

- On the Notifications tab, the tabbar item completely duplicates its functionality. So it was superfluous there.
- On the Home tab, the tabbar item was duplicating its functionality except for lack of a badge. So I added a badge.
- In other places, we don't always have a direct replacement (yet?)
  - You can always pull-to-refresh on mobile.
  - On profile, tab headers themselves act as a way to refresh. This is maybe not very discoverable.
  - On a standalone feed page on the web, there is currently no way to force a refresh.

I think I'd like to run an experiment regardless. This seems directionally better to me. If there is a drop in engagement, we can dig into that and maybe preserve the button for the "standalone feed page" case. Or come up with another UI there.

## Things to consider before a wider rollout

- The "individual feed" screen case described above, maybe it still needs something
- It's kind of annoying to get badged for Discover (and such) because they have no real notion of continuity— so the badge doesn't actually mean much (except that the polling timeout has run). Especially if you're somewhere deep inside a thread, and it's like we're pulling you out of the thread.
  - Let's try this on TF for a while and get a feel for what to do here.
    - Maybe we should disable polling _while_ you're in a thread but poll immediately on pop.

## Test Plan

Verify no changes with the `remove_show_latest_button` gate off.

Allowlist yourself into the gate.

The "Load latest" floating button should be gone everywhere.

But on Home, there will be now a new badge representing the currently selected feed.

- Verify the badge shows on all platforms
- Verify switching current feed updates the badge
- Verify pressing the badge clears it
- Verify the badge eventually shows up while you're on a different tab (e.g. Profile)

https://github.com/user-attachments/assets/0e11d07d-e605-4749-ac2b-2f0b52299e94

https://github.com/user-attachments/assets/d76cb169-349a-438a-95c2-45a54261db1a

https://github.com/user-attachments/assets/6e9fe021-c342-409d-a96f-5aed9c8a36be

### DIFF
--- a/src/view/com/feeds/FeedPage.tsx
+++ b/src/view/com/feeds/FeedPage.tsx
@@ -14,6 +14,7 @@ import {s} from '#/lib/styles'
 import {isNative} from '#/platform/detection'
 import {listenSoftReset} from '#/state/events'
 import {FeedFeedbackProvider, useFeedFeedback} from '#/state/feed-feedback'
+import {useSetHomeBadge} from '#/state/home-badge'
 import {RQKEY as FEED_RQKEY} from '#/state/queries/post-feed'
 import {FeedDescriptor, FeedParams} from '#/state/queries/post-feed'
 import {truncateAndInvalidate} from '#/state/queries/util'
@@ -59,6 +60,13 @@ export function FeedPage({
   const feedFeedback = useFeedFeedback(feed, hasSession)
   const scrollElRef = React.useRef<ListMethods>(null)
   const [hasNew, setHasNew] = React.useState(false)
+  const setHomeBadge = useSetHomeBadge()
+
+  React.useEffect(() => {
+    if (isPageFocused) {
+      setHomeBadge(hasNew)
+    }
+  }, [isPageFocused, hasNew, setHomeBadge])
 
   const scrollToTop = React.useCallback(() => {
     scrollElRef.current?.scrollToOffset({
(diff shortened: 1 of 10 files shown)


src/view/com/feeds/FeedPage.tsx
```diff
@@ -59,6 +60,13 @@ export function FeedPage({
   const feedFeedback = useFeedFeedback(feed, hasSession)
   const scrollElRef = React.useRef<ListMethods>(null)
   const [hasNew, setHasNew] = React.useState(false)
+  const setHomeBadge = useSetHomeBadge()
+
+  React.useEffect(() => {
+    if (isPageFocused) {
+      setHomeBadge(hasNew)
+    }
+  }, [isPageFocused, hasNew, setHomeBadge])
```
gaearon, line 69: The goal here is to react to both `isPageFocused` and `hasNew` so making it an effect is the easiest way to do that. We might want to refactor polling more generally and move some of this state around but I'm hesitant to do this now.

estrattonbailey (approved):
Worth a test, I agree

haileyok (approved):

reetp:
> The "Load latest" floating button should be gone everywhere.

Fantastic. 

> [Experiment]

Can we get a version to test?

PR #7132, opened 2024-12-16
Adjust gates

gaearon:
A few things:

- Remove the hardcoded `DISCOVER_DEBUG_DIDS` stuff. It sucks because it's not visible in production.
- Instead, reintroduce two different gates: `debug_subscriptions` and `debug_show_feedcontext`
  - They don't have Metric Lifts enabled so they're free to check
- Narrow down the build conditions
  - Only enable the app icon stuff when `IS_INTERNAL` is _also_ true (we never want it in public builds)
  - Enable the debug context if the gate is true _or_ we're in TestFlight (where we always enable it)

Basically the goal is to get rid of this hardcoded list.

## Test Plan

Probably easiest to test by going to an alt, verifying it has no feed feedback or app icon, then allowlisting it in each gate, and verifying both things show up. Note that feed context _also_ shows up for `IS_INTERNAL` so you'll want to comment that check out before testing.

### DIFF
--- a/src/view/com/util/post-ctrls/PostCtrls.tsx
+++ b/src/view/com/util/post-ctrls/PostCtrls.tsx
@@ -87,12 +87,7 @@ let PostCtrls = ({
   const {captureAction} = useProgressGuideControls()
   const playHaptic = useHaptics()
   const gate = useGate()
-  const isDiscoverDebugUser =
-    IS_INTERNAL ||
-    gate('debug_show_feedcontext', {
-      // Employee only
-      dangerouslyDisableExposureLogging: true,
-    })
+  const isDiscoverDebugUser = IS_INTERNAL || gate('debug_show_feedcontext')
   const isBlocked = Boolean(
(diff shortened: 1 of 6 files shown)


src/view/com/util/post-ctrls/PostCtrls.tsx
```diff
@@ -85,8 +86,8 @@ let PostCtrls = ({
   const {sendInteraction} = useFeedFeedbackContext()
   const {captureAction} = useProgressGuideControls()
   const playHaptic = useHaptics()
-  const isDiscoverDebugUser =
-    IS_INTERNAL || DISCOVER_DEBUG_DIDS[currentAccount?.did ?? '']
+  const gate = useGate()
+  const isDiscoverDebugUser = IS_INTERNAL || gate('debug_show_feedcontext')
```
gaearon, line 90: intentionally `||` here

haileyok (approved):

PR #7136, opened 2024-12-17
[ELI5] Tweak wording on the signup screen

gaearon:
People get confused here unsure what "handle" is. "Input your handle" is also confusing.

The first commit replaces the wordings.

Screenshot 2024-12-17 at 00 02 59

This matches our usage of "username" on the login screen.

The second commit is a bit unusual — I'm updating all translations as if this was the wording all along. That's to avoid regressions to translations just before the release. It's fine to update them later (if needed per language) since conceptually the phrasing is the same and the meaning has not changed.

It's a little inconsistent with what we have in Settings. Maybe in the future we'll change it there too. But for now let's make sure people are able to create an account.

## Test Plan

Verify above for English.

Then `yarn intl:build` and verify translations have not regressed and show old strings:

Screenshot 2024-12-17 at 00 01 50

### DIFF
--- a/src/locale/locales/an/messages.po
+++ b/src/locale/locales/an/messages.po
@@ -3613,7 +3613,7 @@ msgstr "Escribe la tuya clau"
 #~ msgstr "Escribe lo tuyo furnidor d'aloch preferiu"
 
 #: src/screens/Signup/StepHandle.tsx:114
-msgid "Input your user handle"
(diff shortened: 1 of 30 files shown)


gaearon:
got verbal ok from @pfrazee

mozzius (commented):
code lgtm for what it's worth

Issue #4703, opened 2024-06-29
Dedicated Trending Section


Mennaruuk:
**Is your feature request related to a problem? Please describe.**

Currently, Bluesky lacks an Explore tab. There are some users who want to curate their feeds and only follow a handful of accounts. However, there are others (like me) who want to know the chatter around the town and what's popular at a particular moment.

**Describe the solution you'd like**

Twitter had a cool Explore section until recently. It used to take multiple factors into account, such as the volume of tweets discussing a particular subject, the rate at which they are being posted, user engagement, and what new topics are gaining lots of traction all of a sudden. It can also be personalized, so if you're living in one city you may see what's trending in your city rather than another city.

![Top-1024x877](https://github.com/bluesky-social/social-app/assets/52135169/ea829f0f-a9da-4b23-8171-0cc3af41f47b)

**Describe alternatives you've considered**

Discover exists, although it's not topical. The strength of Twitter's Explore section is that it shows a list of the most discussed topics or hashtags. That way, I can press any of them and be redirected to tweets about that particular topic. None of Bluesky's feeds can do this at the moment. 

**Additional context**

I want to mention that Twitter's Explore tab isn't merely a collection of the most popular posts about a topic sorted from most liked or retweeted to least. It's more of a random amalgamation of what's most trending.

Also, there are different themes to choose, such as Sports, News, etc.:
![promoted-tweet-samsung](https://github.com/bluesky-social/social-app/assets/52135169/cad82132-d378-4493-ba47-f20c8d59ce49)

mmangopear:
Did the devs ever see this or add it to the roadmap? I feel like this is one of the few big missing holes

gcjbr:
Would be down to work on it if I had confirmation this is of interest of the maintainers

surfdude29:
FYI @pfrazee said earlier today:

> ... I've got a proposal for how to do Trending Topics. 🤞

https://bsky.app/profile/pfrazee.com/post/3l2zpqqyb762x

borysmarceloherrera:
Please implement

borysmarceloherrera:
How difficult is to create a tab and consume an api with a aggregated most commented words or a combination of 2 words.

It is very poor user experience without knowing what is going on. 

I think a World Wide tab and a Local tab based on users locations should be interesting.

ajalaca:
[...]

PR #7144, opened 2024-12-17
Trending (Beta)

estrattonbailey:


src/state/trending-config.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const langPrefs = useLanguagePrefs()
+  const {data: config, isLoading: isInitialLoad} = useServiceConfigQuery()
+  const ctx = React.useMemo<Context>(() => {
+    if (__DEV__) {
+      return {enabled: true}
+    }
+
+    /*
+     * Only English during beta period
+     */
```
haileyok, line 27: wonder if we're better off just adding this back to feature flags, so that way we can just change it live instead of requiring a whole new rollout? fwiw idt the change will happen super soon, but just thinking ahead.

src/state/trending-config.tsx
```diff
@@ -0,0 +1,70 @@
...
+      !!langPrefs.contentLanguages.length &&
+      !langPrefs.contentLanguages.includes('en')
+    ) {
+      return {enabled: false}
+    }
+
+    /*
+     * While loading, use cached value
+     */
+    const cachedEnabled = device.get(['trendingBetaEnabled'])
```
haileyok, line 38: nice

src/state/trending-config.tsx
```diff
@@ -0,0 +1,70 @@
...
+      // cache for next reload
+      device.set(['trendingBetaEnabled'], enabled)
+      return {enabled: false}
+    }
+
+    /*
+     * Service is enabled, but also check statsig in case we're rolling back.
+     */
+    const gateEnabled = gate('trending_topics_beta')
+    const _enabled = enabled && gateEnabled
```
haileyok, line 58: oh, are we also using a statsig gate? we need to keep in mind adblock is a thing, which is the main reason i was leaning on the appview for the flag rather than a featuregate.

src/components/TrendingTopics.tsx
```diff
@@ -0,0 +1,223 @@
...
+  size,
+  style,
+}: {topic: TrendingTopic; size?: 'large' | 'small'} & ViewStyleProp) {
+  const t = useTheme()
+  const topic = useTopic(raw)
+
+  const isSmall = size === 'small'
+  // const hasAvi = topic.type === 'feed' || topic.type === 'profile'
+  // const aviSize = isSmall ? 16 : 20
+  // const iconSize = isSmall ? 16 : 20
```
haileyok, line 28: i take it this would be for suggestions right?

src/state/trending-config.tsx
```diff
@@ -0,0 +1,70 @@
...
+     */
+    const cachedEnabled = device.get(['trendingBetaEnabled'])
+    if (isInitialLoad) {
+      return {enabled: Boolean(cachedEnabled)}
+    }
+
+    /*
+     * Doing an extra check here to reduce hits to statsig. If it's disabled on
+     * the server, we can exit early.
+     */
```
gaearon, line 46: if we disable metric lifts for the gate, there won't be a cost associated with it, so it's fine to hit it fwiw

src/components/TrendingTopics.tsx
estrattonbailey: trending feeds, profiles, starter packs, etc, just commented out all my WIP for that for now

src/state/trending-config.tsx
estrattonbailey: I was thinking that we'd want keep them, but happy to disable too

src/state/trending-config.tsx
estrattonbailey: Yeah trying to prevent that initial flash as much as possible

src/state/trending-config.tsx
estrattonbailey: We certainly can!

src/state/trending-config.tsx
estrattonbailey: Discussed in Slack — as we start to roll out this feature, this side effect is OK. We'll eventually remove the flag entirely, so adblockers won't be an issue 👍

src/state/trending-config.tsx
estrattonbailey: Discussed in Slack, gonna keep this during our initial rollout, will remove later on

pfrazee (approved):
Great work all. And away we go

stevenlafl:
https://github.com/bluesky-social/social-app/issues/4703

This requires closing for backlog grooming.

PR #7145, opened 2024-12-17
[Discover] Filter out globally seen posts client-side

gaearon:
We have a tension between two wants:

- We want Discover to avoid showing posts you've already seen
- Actually sending _every post you see_ (e.g. in Following) to Discover would be both too expensive on the server side to store and would have privacy implications if we ever wanted to extend this to third party feeds
  - But special-casing Discover forever, even though it is the primary feed in the app, isn't great either

The proposed solution here, which we arrived at after a discussion with @pfrazee and @whyrusleeping, is client-side deduplication. It's not perfect but it hits these points:

- It'll avoid showing posts you've already seen on the mobile app during the same session
  - Could be extended to multiple sessions if we serialize the buffer (it doesn't need to survive long)
  - Could be shared between browsers tabs in the future as well (for now it's not)
- It doesn't require any server computation
- In the longer run, we could imagine third party feeds declaring they want this capability on. Then Discover wouldn't need to be special-cased. Most third party feeds probably _don't_ want this so it would be opt-in.

## Test Plan

Add a log, verify posts get added to the buffer. Currently they get added:

- In any feeds, on linger
- In post thread, immediately for highlighted post + parent chain (we don't count linger there)

You can verify that with a log.

https://github.com/user-attachments/assets/4fbf5484-a0ee-44ac-ba28-e09cd9bdd27c

Then you can verify that it does the filtering. For that, add a log to where the filtering happens and log the post being filtered out. Then spend some time scrolling Following. Then go to Discover and repeatedly request it — if you're lucky, you'll trigger the behavior. Observe that the filtered post is not being displayed.

Verify logged-out Discover still works.

Verify the filtering is not being applied to other feeds.

## The Mechanism

I used a Bloom filter because it's cool. (?) No seriously, it's relatively light in memory (see the linked formula), otherwise a Set would be a lot larger. If we want to maintain continuity and serialize it or sync it between tabs, it'll get more complicated, but I think this is an OK start for now.

I'm doing this in the response handler rather than in the query or the feed tuner because it's not deterministic. The filter is mutable so we need to make sure we apply it once and consistently. Right before handing off the response seems like the perfect time.

### DIFF
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -20,7 +20,15 @@ type StateContext = {
 
 const stateContext = React.createContext<StateContext>({
   enabled: false,
-  onItemSeen: (_item: any) => {},
+  onItemSeen: (feedItem: any) => {
+    const slice = getFeedPostSlice(feedItem)
+    if (slice === null) {
+      return
+    }
+    for (const postItem of slice.items) {
+      markGloballySeenPost(postItem.uri)
+    }
+  },
   sendInteraction: (_interaction: AppBskyFeedDefs.Interaction) => {},
 })
 
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -265,7 +265,7 @@ const m = Math.ceil((n * Math.log(p)) / Math.log(1 / Math.pow(2, Math.log(2))))
 const k = Math.round((m / n) * Math.log(2))
 let globalBloomFilter = new BloomFilter(m, k)
 
-function markGloballySeenPost(uri: string) {
+export function markGloballySeenPost(uri: string) {
   if (globalBloomFilter.size >= n) {
     // If we ever get here, just restart to avoid saturation.
     globalBloomFilter = new BloomFilter(m, k)
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -274,7 +274,7 @@ const k = Math.round((m / n) * Math.log(2))
 let globalBloomFilter = new BloomFilter(m, k)
 
 export function markGloballySeenPost(uri: string) {
-  if (globalBloomFilter.size >= n) {
+  if (globalBloomFilter.size() >= n) {
     // If we ever get here, just restart to avoid saturation.
     globalBloomFilter = new BloomFilter(m, k)
   }
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -273,14 +273,18 @@ const m = Math.ceil((n * Math.log(p)) / Math.log(1 / Math.pow(2, Math.log(2))))
 const k = Math.round((m / n) * Math.log(2))
 let globalBloomFilter = new BloomFilter(m, k)
 
+const salt = Math.random().toString(36).slice(2)
+
 export function markGloballySeenPost(uri: string) {
   if (globalBloomFilter.size() >= n) {
     // If we ever get here, just restart to avoid saturation.
     globalBloomFilter = new BloomFilter(m, k)
   }
-  globalBloomFilter.add(uri)
+  const key = uri + salt
+  globalBloomFilter.add(key)
 }
 
 export function isLikelyGloballySeenPost(uri: string) {
-  return globalBloomFilter.test(uri)
+  const key = uri + salt
+  return globalBloomFilter.test(key)
 }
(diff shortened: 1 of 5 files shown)


src/state/feed-feedback.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 import {AppState, AppStateStatus} from 'react-native'
 import {AppBskyFeedDefs} from '@atproto/api'
+// @ts-ignore
```
gaearon, line 4: sorry

SleeplessByte: Maybe:

```typescript
// store in a .d.ts file
declare module 'bloomfilter' {
  export class BloomFilter {
    buckets: Int32Array;

    /**
     * Create a new empty bloom filter of size m with hashes k or
     * provide buckets as a number[] or Int32Array to deserialize a bloom filter
     * @param mOrBucketsArray number of bits (will be rounded up to nearest 32), or buckets
     * to deserialize into a filled bloomfilter
     * @param k number of hashes
     */
    constructor(m: number | number[] | Int32Array, k: number);

    /**
     * Add a value to a bloom filter
     * @param value
     */
    add(value: any): void;

    /**
     * Test whether a value exists in a bloom filter. (False positives are
     * possible, false negatives are not.)
     */
    test(value: any): boolean;

    /**
     * Estimated cardinality.
     */
    size(): number;
  }
}
```

Based on https://www.npmjs.com/package/@types/bloomfilter#additional-details and https://github.com/jasondavies/bloomfilter.js/blob/master/bloomfilter.js#L72-L77

nevercast:
This looks really good! Nice one! 

I did have one thought: the bloomfilter library (file) is only 100 lines long, has no dependencies, and hasn't been updated in 7 years. Likely abandoned. I wonder if this would be a good candiate for vendoring rather than an npm dependency. Only other requirement would be including the BSD-3 licence (MIT compatible) perhaps at the top of the file.

Hope you don't mind my drive by comment, saw the mention of bloom filter+ discover feed and definitely had to have a look.

Cheers,
Josh

src/state/feed-feedback.tsx
mozzius: can't we just use @types/bloomfilter?

gaearon:
This is only relevant for the Discover feed. It was already filtering out already seen posts on Discover so this change has no effect on the issue you're describing.

I agree re: it would be nice to maintain continuity; we want to do this too but it's a bigger project.

src/state/feed-feedback.tsx
SleeplessByte: You can, but then you miss `size(): number`.

SleeplessByte: Personally I think https://github.com/bluesky-social/social-app/pull/7145 is right, and instead you'll want to "just" create a local typescript file, even, porting the 7 year old code. Maintain it locally.

src/state/feed-feedback.tsx
```diff
@@ -93,24 +103,25 @@ export function useFeedFeedback(feed: FeedDescriptor, hasSession: boolean) {
...
-              item: postItem.uri,
-              event: 'app.bsky.feed.defs#interactionSeen',
-              feedContext: slice.feedContext,
-            }),
-          )
-          sendToFeed()
+        markGloballySeenPost(postItem.uri)
+
+        if (enabled) {
+          if (!history.current.has(postItem)) {
```
bel7aG, line 114: Would be better to keep a single compute like:

```JSX
if (enabled && !history.current.has(postItem))
```

src/view/com/post-thread/PostThread.tsx
```diff
@@ -391,6 +392,22 @@ export function PostThread({uri}: {uri: string | undefined}) {
...
+      // just the highlighted post and the parents. Presumably,
+      // if you're already midthread, you know what was above.
+      const {parents, highlightedPost} = skeleton
+      markGloballySeenPost(highlightedPost.uri)
+      for (let i = 0; i < parents.length; i++) {
+        const parent = parents[i]
+        if (isThreadPost(parent)) {
+          markGloballySeenPost(parent.uri)
+        }
+      }
```
bel7aG, line 407: Maybe `for..of` is more cleaner:
 
```JSX
 for (const parent of parents) {
      if (isThreadPost(parent)) {
        markGloballySeenPost(parent.uri)
      }
  }
```

PR #7146, opened 2024-12-18
Fix emoji picker position

estrattonbailey:
Just needed to be wrapped in a Portal, otherwise it was being positioned relative to the center column. Floating elements like this should always be moved out to the root so they never conflict with other UI.

Test: chat and composer on web

After, before
CleanShot 2024-12-17 at 18 18 27@2x
CleanShot 2024-12-17 at 18 18 43@2x


haileyok (approved):

gaearon:
tested in composer too?

PR #7148, opened 2024-12-18
[topics] use a screen for topics rather than search screen

haileyok:


src/screens/Topic.tsx
```diff
@@ -0,0 +1,204 @@
...
+    isFetched,
+    isFetchingNextPage,
+    isLoading,
+    isError,
+    error,
+    refetch,
+    fetchNextPage,
+    hasNextPage,
+  } = useSearchPostsQuery({
+    query: topic.replaceAll(' ', '+').replaceAll('%20', '+'),
```
estrattonbailey, line 150: Do we need to decode here first too? Haven't tested just stood out to me

haileyok: oh yea lets do that

estrattonbailey (approved):
Nice thanks! Little behind base but should be fine to merge in right?

haileyok:
made another one because this was missing a push or something, but same thing otherwise.

PR #6907, opened 2024-12-02
[Layout] Base

estrattonbailey:
See docs here: https://github.com/bluesky-social/social-app/blob/d3bd2a4d30de875e13771bda3d2b79e1e5dbfca3/src/components/Layout/README.md

New components:
- `Layout.Header.*` see docs for more info
- `Layout.Center` see docs, replaces `CenteredView`
- `Layout.Content` see docs, replaces `ScrollView` from `Views.tsx`
- `useGutterStyles` from ALF, start using this for gutters

Testing
- All screens
- Fixed headers on native, mobile web, animations
- Dropdowns, old modals, new modals, composer
  - Test with scrollbars enabled or disabled at the OS level

### Todo
- [x] Special header cases
  - [x] Notifications header #6910
  - [x] Hashtag screen header #6928
  - [x] ChatList header #6929
  - [x] Feeds screens headers (do not review yet) #6913
- [x] Clean up Pager and PagerWithHeader usage
- [x] Handle nested offsets
- [x] Decide on back button for chats
- [x] Consistent back buttons, do we have them or not on all screens
- [x] Feeds screen bug

PR #7153, opened 2024-12-18
Web sticky headers for most screens

mozzius:
Makes headers sticky on web, except for:

- Notifications (not needed for root tabs, the others don't use the header component anyway)
- Post thread (maintaining content position when loading replies puts the header on top of content)

https://github.com/user-attachments/assets/9573671f-5bac-45ea-8aca-124f449e8510

# Test plan

Look at all screens with headers, see if there's any where scrolling down looks janky

### DIFF
--- a/src/components/Layout/Header/index.tsx
+++ b/src/components/Layout/Header/index.tsx
@@ -30,10 +30,12 @@ import {Text} from '#/components/Typography'
 export function Outer({
   children,
   noBottomBorder,
+  headerRef,
   sticky = true,
 }: {
   children: React.ReactNode
   noBottomBorder?: boolean
+  headerRef?: React.MutableRefObject<View | null>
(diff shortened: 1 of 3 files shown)


auroursa:
In old layout, the back button was placed on the left sidebar, and I was accustomed to using it. 

Since #6907, it seems that the back button on the Web is placed in the header. If I click on a reply in the middle of a long post thread via the timeline, I have to scroll to the top to find the back button.

In contrast, on other screens, I can easily click the back button in the sticky header. While I can use the browser’s global back button to do this, it still feels a bit confusing to me.

What I understand for not adding a sticky header in the post thread is it takes up too much space. How about if scrolling stops, or if I start scrolling up, the header will appear, making it easier to quickly find the back button?

gaearon (approved):

PR #7156, opened 2024-12-18
Add reply sorting in-thread

gaearon:
Now that the header is always visible on all platforms, it seems like a nice place to put the sorting options.

The behavior is not perfect (e.g. it's a bit slower than I would have liked) but seems like we can ship and iterate.

## Test Plan

https://github.com/user-attachments/assets/15424c68-9d71-44b2-9067-5401eb4e4a35

https://github.com/user-attachments/assets/3df43707-d3fd-4508-a3b0-b5914973c725

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -190,6 +190,55 @@ export function ItemIcon({icon: Comp}: ItemIconProps) {
   )
 }
 
+export function ItemRadio({selected}: {selected: boolean}) {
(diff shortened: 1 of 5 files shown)


gaearon:
Pushed tree view setting as well.

Screenshot 2024-12-18 at 18 20 14
Screenshot 2024-12-18 at 18 20 21

estrattonbailey (approved):
HOTT

PR #7166, opened 2024-12-18
Don't hide recommended

gaearon:
We wanted to keep showing Recommended unconditionally.

## Test Plan

It shows even if you disable trending


estrattonbailey (approved):

0xGilda:
This sucks

PR #7174, opened 2024-12-19

### LINKED ISSUE #6728: Post time in Chinese cannot be localized on Android

auroursa:

**Update**: We are attempting to convert the language codes for Chinese and Cantonese from [RFC-1766](https://learn.microsoft.com/en-us/windows/win32/wmformat/language-strings) to [RFC-4646](https://www.iana.org/assignments/language-tags/language-tags.xhtml), to resolve language-related compatibility issues.

The following are the changes:

- `zh-CN` -> `zh-Hans-CN`
- `zh-TW` -> `zh-Hant-TW`
- `zh-HK` -> `yue`

Similar to `ast` #6774, `yue` has been proven to be a valid language code for application languages. Tests have shown that `zh-Hans-CN` and `zh-Hant-TW` can also be correctly used by Lingui and the application.

Note the impact of changing the language code is limited to the app display language, Cantonese still cannot be used as post language.

This will be an attempt to fix #6728 without introducing a polyfills.

Co-maintained by @cirx1e @hakadao @VentusUta


src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,23 @@ export function getLocales() {
     }
 
     if (typeof locale.languageTag === 'string') {
-      if (locale.languageTag.startsWith('zh-Hans')) {
-        // Simplified Chinese to zh-CN
-        locale.languageTag = 'zh-CN'
+      if (
+        locale.languageTag === 'zh-CN' ||
```
auroursa, line 45: To ensure browser compatibility, as browsers still widely use RFC-1766.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,23 @@ export function getLocales() {
...
+        locale.languageTag.startsWith('zh-Hans')
```
auroursa, line 46: We still need to do this, because iOS will not automatically fall back from `zh-Hans-CN` to `zh-Hans`, using `zh-Hans-CN` would break browser compatibility.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,23 @@ export function getLocales() {
...
+      ) {
+        // Simplified Chinese to zh-Hans
+        locale.languageTag = 'zh-Hans'
       }
-      if (locale.languageTag.startsWith('zh-Hant')) {
-        // Traditional Chinese to zh-TW
-        locale.languageTag = 'zh-TW'
+      if (
+        locale.languageTag === 'zh-TW' ||
```
auroursa, line 52: For the same reason as above—browser compatibility.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,23 @@ export function getLocales() {
...
+        locale.languageTag.startsWith('zh-Hant')
```
auroursa, line 53: Same here.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,23 @@ export function getLocales() {
...
+      ) {
+        // Traditional Chinese to zh-Hant
+        locale.languageTag = 'zh-Hant'
       }
       if (locale.languageTag.startsWith('yue')) {
-        // Cantonese (Yue) to zh-HK
-        locale.languageTag = 'zh-HK'
+        // Cantonese (Yue) to yue
+        locale.languageTag = 'yue'
```
auroursa, line 60: To match the two variants on Android and iOS:
`yue-Hans` and `yue-Hant`.

Not using `yue-Hant` directly due to poor browser compatibility.

auroursa (commented):
Let me explain some of the related changes to `deviceLocales.ts`, which are made to compatibility for both Web and Native platforms.

auroursa:
By converting the language tag from RFC-1766 to RFC-4646, it successfully fixed the post time issue. It’s still not perfect, at least it is on the same level as Japanese display without polyfill.

| Android | iOS |
|--- | ---|
|  |  |

PR #7175, opened 2024-12-19
Implement localization for ProfileList section titles

auroursa:
This makes `SECTION_TITLES_CURATE` localizable.

~~The modified variable name can be changed to something else that easier to understand.~~
Update: Change it to inline with `_(msg`.

### DIFF
--- a/src/view/screens/ProfileList.tsx
+++ b/src/view/screens/ProfileList.tsx
@@ -79,8 +79,6 @@ import * as Prompt from '#/components/Prompt'
 import {ReportDialog, useReportDialogControl} from '#/components/ReportDialog'
 import {RichText} from '#/components/RichText'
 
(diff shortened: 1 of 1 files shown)


src/view/screens/ProfileList.tsx
```diff
@@ -162,6 +162,7 @@ function ProfileListScreenLoaded({
   const isHidden = list.labels?.findIndex(l => l.val === '!hide') !== -1
   const isOwner = currentAccount?.did === list.creator.did
   const scrollElRef = useAnimatedRef()
+  const SECTION_TITLES_CURATE = SECTION_TITLES_CURATE_RAW.map(title => _(title))
```
gaearon, line 165: you can just define them inline with `_(msg` if we're doing this anyway

PR #7195, opened 2024-12-19
Date time format fix

estrattonbailey:
DMs were showing as being sent at UTC instead of the user's local time zone on iOS and Android. This seems to have been introduced in #6742.

This new polyfill should allow us to [specify a default timezone](https://formatjs.github.io/docs/polyfills/intl-datetimeformat/#default-timezone) on app load, but this does not appear to work on iOS or Android. See first commit for that attempt.

In the second commit, I applied the "fix", which is to use the built-in `Intl.DateTimeFormat` instead of the polyfill. **I have not yet confirmed if this regresses #6742.**

### DIFF
--- a/src/locale/i18n.ts
+++ b/src/locale/i18n.ts
@@ -1,7 +1,6 @@
 // Don't remove -force from these because detection is VERY slow on low-end Android.
 // https://github.com/formatjs/formatjs/issues/4463#issuecomment-2176070577
 import '@formatjs/intl-locale/polyfill-force'
-import '@formatjs/intl-datetimeformat/polyfill-force'
 import '@formatjs/intl-pluralrules/polyfill-force'
 import '@formatjs/intl-numberformat/polyfill-force'
 import '@formatjs/intl-datetimeformat/locale-data/en'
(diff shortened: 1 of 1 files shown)


gaearon:
did this in https://github.com/bluesky-social/social-app/pull/7196

surfdude29:
FYI several users in this thread reported times being shown incorrectly, apparently on posts too:

https://bsky.app/profile/pfrazee.com/post/3ldoxeka2uk2y

e.g.:
https://bsky.app/profile/magnolia2010.bsky.social/post/3ldpdxxjwpc2f
https://bsky.app/profile/mdruker.bsky.social/post/3ldpckwazks2s
https://bsky.app/profile/kelsee.bsky.social/post/3ldoxgaxrkk2l

auroursa:
I faced a timezone issue while exploring #6742. This was [my solution](https://github.com/bluesky-social/social-app/commit/c451b7be07f183dd32c0a749cda8e1817364f980) at the time (I hope this helps). Since no longer relied on datetimeformat to output time, I thought it was no longer an issue and removed it. I sincerely apologize for the consequences of not testing it thoroughly.

auroursa:
Upon rechecking the first commit, it seems that it was just missing `import '@formatjs/intl-datetimeformat/add-all-tz.js'`, adding it will make it work properly. 
image

auroursa:
According to the first commit, the missing import has been added. Please see #7204.

Issue #7125, opened 2024-12-16
Copy link icon should be 🔗 Link instead of ⬆️ Sharrow


brandonhaslegs:
### Steps to Reproduce

1. Go to any post
2. View row of icons under post
3. Click sharrow icon
4. Expect options for where to share or OS-level share sheet
5. Link is copied to clipboard
6. Click ••• icon
7. "Copy link to post" icon should be 🔗 instead of ⬆️

### Attachments

<img width="668" alt="Telegram 2024-12-16 14 52 54" src="https://github.com/user-attachments/assets/ee9cf44a-6ad6-4fa0-9773-bc2855fd3535" />

<img width="292" alt="image" src="https://github.com/user-attachments/assets/5b6bf85e-faf5-4f9f-bf1f-6364512ce042" />

<img width="272" alt="image" src="https://github.com/user-attachments/assets/bab052b2-ab0f-4261-b998-99f5df3b9d63" />

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### Device Info

2023 Mac Mini M2 Pro, Arc Version 1.72.1 (56511)

### What version of the app are you using?

Web

### Additional Information

This is not an issue on iOS because the share button brings up a share sheet.

PR #7203, opened 2024-12-20
Change copy link icons from ArrowOutOfBox to Link

ConnorBstill:
Brought up by this issue: https://github.com/bluesky-social/social-app/issues/7125
Screenshot 2024-12-19 at 9 56 28 PM


heikadog:
i think what the issue was asking for was the arrowoutofbox on the first icon to stay the same but the copy link icon to be a paperclip?

also the icon is too big

ConnorBstill:
> i think what the issue was asking for was the arrowoutofbox on the first icon to stay the same but the copy link icon to be a paperclip?
> 
> also the icon is too big

Yeah you're right it is a little big, I don't see where they said anything about a paperclip though

ConnorBstill:
Slightly smaller

Screenshot 2024-12-20 at 3 43 58 PM

mozzius (changes requested):
I like this! a couple things:

1. Can we keep this to web-only. ArrowOutOfBox is still better for native imo
2. There's several other menus were this needs to be done - sharing a feed or a list for example
3. Can we not do it for the PostCtrls? I think it works best just in the menus. I'll be changing how the PostCtrls one works soon anyway

ConnorBstill:
Sounds good, thanks for the feedback! Just to be completely sure, the components are all shared between web and native correct? So all I need to do is use `isWeb`?

ConnorBstill:
Added to other menus where I could find usage of `ArrowOutOfBox_Stroke2_Corner0_Rounded` except for in `src/screens/Topic` and `src/screens/Hashtag` which seems to me like just mobile layout screens. Let me know if I missed something

Screenshot 2025-01-07 at 11 39 08 PM
Screenshot 2025-01-07 at 11 59 39 PM
Screenshot 2025-01-08 at 12 06 13 AM

ConnorBstill:
@mozzius I missed the list share icon (which you mentioned, oops),  here's what that looks like now:
Screenshot 2025-01-10 at 12 46 22 AM

The icons that the `NativeDropdown` component uses look different (and worse imo) than the ones in the other menus, so if we want these to be consistent then NativeDropdown needs to be refactored, or its usage replaced with the other Menu component. Maybe it needs to stay the way it is for some reason unknown to me, but if it's something we want to change I can try to take that on in another issue/PR.

mozzius:
Superceded by #7840, but thank you for the PR!

PR #7138, opened 2024-12-17
Update Lingui to 5.1.1

auroursa:
There are some new features here that I'm really interested in, with [Print Placeholder Values](https://lingui.dev/blog/2024/11/28/announcing-lingui-5.0#print-placeholder-values-for-better-translation-context), this allows us to know exactly what the placeholders are.

Before this, `{0} of {1}` was difficult to translate.

image

After updates, translators can more easily understand `{0}` is `currentTime` and `{1}` is `duration` without having to check the source code.

image

The first commit is more like a regular upgrade, with no major changes.

The second commit removed `@lingui/macro` according to the [migration guide](https://lingui.dev/releases/migration-5) and updated all related imports. We can stick to the first commit for now, but since `@lingui/macro` is deprecated and will be removed in the next major version, we may need to address this later.

PR #7204, opened 2024-12-20
Fix timezone issue when using polyfill

auroursa:
Stack on [#7195(commit)](https://github.com/bluesky-social/social-app/commit/a09c5d6d8ea346a3a044dbed8aa9d8bbfeca9503).

According to the [documentation](https://formatjs.github.io/docs/polyfills/intl-datetimeformat/#default-timezone), I imported the missing `add-all-tz.js`, and it should work properly again.

Checked on the Galaxy S22 and iOS simulator, the timezone in the screenshot is based on Central Time (UTC-6).

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@expo/webpack-config": "^19.0.0",
     "@floating-ui/dom": "^1.6.3",
     "@floating-ui/react-dom": "^2.0.8",
+    "@formatjs/intl-datetimeformat": "^6.17.1",
     "@formatjs/intl-locale": "^4.2.8",
     "@formatjs/intl-numberformat": "^8.15.1",
     "@formatjs/intl-pluralrules": "^5.4.1",
--- a/src/locale/i18n.ts
+++ b/src/locale/i18n.ts
@@ -7,6 +7,7 @@ import '@formatjs/intl-numberformat/polyfill-force'
 import '@formatjs/intl-datetimeformat/locale-data/en'
 import '@formatjs/intl-pluralrules/locale-data/en'
 import '@formatjs/intl-numberformat/locale-data/en'
+import '@formatjs/intl-datetimeformat/add-all-tz.js'
 
 import {useEffect} from 'react'
 import {getCalendars} from 'expo-localization'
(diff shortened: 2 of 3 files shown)


surfdude29:
This may not be feasible, but for the `en-GB` locale would it be possible to still use the Lingui date formatting rather than the polyfill?

The polyfill sets the format as `[date], [time]` for `en-GB` on iOS (not sure about other platforms):
IMG_1678

Which I think doesn't look as good as Lingui displays it, as `[date] at [time]`:
IMG_1883

auroursa:
> This may not be feasible, but for the `en-GB` locale would it be possible to still use the Lingui date formatting rather than the polyfill?

As mentioned [here](https://github.com/lingui/js-lingui/issues/2123#issuecomment-2543394990)

> Lingui does not format dates by itself. It internally uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl.
> 
> If dates are not formatted correctly for you it's most likely because you need to set up the polyfill: https://lingui.dev/tutorials/react-native#polyfilling-intl-apis

Also can refer to the related discussion in #7138. After Lingui adds support for date formatting, polyfills will still need to be introduced.

After applying the polyfill, there is indeed some regression with en-GB, as you mentioned. But I believe it is still within an acceptable range compared to the issues with Chinese and Japanese (I also checked the other supported app languages, and there were no particularly severe regressions).

This does not mean en-GB has been compromised. I noticed that [/@formatjs/intl-datetimeformat/locale-data/en-GB.js](https://www.npmjs.com/package/@formatjs/intl-datetimeformat?activeTab=code) defines how to format date outputs. I will look for ways to improve it upstream to reach the level prior to the polyfill.

This is a improvement process, but first, let’s fix the date display for Chinese, as it currently uses the en-US date format to display post times. I’d like to hear your thoughts on this.

auroursa:
Referring to the documentation, I think I have a general understanding of why Chinese cannot display post time correctly without a polyfill.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

> "zh-Hans-CN": Chinese (language) written in simplified characters (script) as used in China (region)

Here Chinese is labeled using the RFC 4646 standard (zh-Hans-CN), but the application language currently uses RFC 1766 (zh-CN). I think that’s the root cause.

For some reason, Android only supports Intl with the RFC 4646 standard and not for RFC 1766. iOS and Web can correctly handle both RFC 4646 and RFC 1766. So, by introducing a polyfill, Android can correctly recognize zh-CN labeled with RFC 1766.

Without relying on a polyfill, post time issues on Android cannot be resolved in the short term unless we switch to a method other than Intl for formatting time. 

Alternatively, we could switch the app language for Chinese to RFC 4646 to stop introducing the polyfill, but this would make handling browser compatibility more challenging, as most browsers still follow RFC 1766...

The overly chaotic standards are the biggest challenge in maintaining Chinese localization, especially for cross-platform app. I’m completely exhausted.😢

auroursa:
No longer using polyfill, fixing this issue by modifying the language codes for Chinese. #7235

Issue #6861, opened 2024-11-29
Post Composer Cursor only reacts when clicking on "What's up?"


SpriterGors:
### Steps to Reproduce

1. Open up Post Composer
2. Notice the blinking cursor on default display text "What's up?"
3. Clicking anywhere inside the post composer disables the cursor, making typing impossible, you really need to deliberately click on the same line as "What's up" to resume typing.

### Attachments

Post composer, as it appears on the web.
![image](https://github.com/user-attachments/assets/38b140e0-3caa-4d94-98a0-5299a3b28563)

Clicking on this red area that looks like input area actually disables the cursor and typing.
![image](https://github.com/user-attachments/assets/fa7b00f0-099e-46c9-b1d3-9cdf2820fcd0)

This is unintuitive, and makes typing considerably difficult on web.

Video:

https://github.com/user-attachments/assets/bdb6725b-c056-4458-8f8d-417d137783fe

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

Windows 11 home; Mozilla Firefox 133.0 (64-bits)

### What version of the app are you using?

Build version: 1.95.0; Bundle info: c9d86de (prod); Bundle date: 24112719; Platform: web; Platform version: undefined

### Additional Information

At least make the input window visually different from the areas that would disable the cursor. Mockup:
![image](https://github.com/user-attachments/assets/1006ca00-7759-4306-91db-5e134221a126)

Though I doubt this is how it should look act. This mockup here is what I actually expect the post composer to behave normally:
![image](https://github.com/user-attachments/assets/7bf3352d-3a03-46bc-a315-acd9322bce4d)

GrizliK1988:
hi!

I've adjusted previous changes to expand composer input area for native and do not break other functionality.

Support for web was dropped from this issue because web now uses modal window and size for composer input is simplified.

One note to adjusted implementation:

There are 2 styles added to Composer.tsx: styles.avatarTopOffset and styles.textInputTopOffset. The reason of having them is to avoid content jumping, when keyboard is being hidden (Animated.ScrollView layout gets changed because of flexGrow is used and padding of 4 pixels causes view to go up 4px for a moment and then padding appears again).

PR #7207, opened 2024-12-20
#6861: Use available space to make focus on Composer TextInput easier

GrizliK1988:
Fixes issue #6861  that is also reproducible on Android and iOS and this PR provides a for Web + Native.

### What this fix does?

Essentially the same thing for Web and Native: utilize display flex to make TextInput to take full available height.

### Web Before and After (Desktop)

**New Post TextInput area before:** 
image

**New Post TextInput area after:** 
image

**Reply TextInput area before:** 
image

**Reply TextInput area after:** 
image

### Web Before and After (Mobile)

**New Post TextInput area before:** 
image

**New Post TextInput area after:** 
image

**Reply TextInput area before:** 
image

**Reply TextInput area after:** 
image

### iOS Before and After

**New Post TextInput area before:** 
image

**New Post TextInput area after:** 
image

**Reply TextInput area before:** 
image

**Reply TextInput area after:** 
image

### Android Before and After

**New Post TextInput area before:** 
image

**New Post TextInput area after:** 
image

**Reply TextInput area before:** 
image

**Reply TextInput area after:** 
image


gaearon:
Oh thanks! Last time I tried to do this I ran into some difficulty. Will give this a go.

c960657:
I wasn't aware of this PR when I filed #7248.

My fix does not address Native, so I think this approach is better.

gaearon:
OK, so one problem with this approach is this case:

Screenshot 2025-01-21 at 00 33 42

We want the text box to hug the last line of the text instead.

gaearon (changes requested):
need to fix the case where you have embeds

GrizliK1988:
Thanks for testing! Let me have a look

gaearon:
@mozzius

GrizliK1988:
Just realized that I missed to add a comment related to last changes.

Web: no changes needed because Composer component is in a modal and it has enforced min height in absolute value.

Native: here changes are still needed and implemented. Initial implementation was wrong because it blocked text input growth in height (and broke embeds styling with it). Current approach works well with embeds and keeps text input expansion in height. 

Why padding top changed: to avoid screen jump when keyboard is hidden. Keyboard hide causes layout change and padding top somehow causes jump in 4 pixels up for a moment and then padding appears again.

mozzius:
Moved to https://github.com/bluesky-social/social-app/pull/7836 so I could rebase it off of main - thank you so much for your great work though! Huge improvement!

GrizliK1988:
Thanks a lot for the great social app and for reviewing/testing :)

PR #7213, opened 2024-12-20
Moderate composer quote

mozzius:
# Test plan

Quote something with a label

### DIFF
--- a/src/view/com/util/post-embeds/QuoteEmbed.tsx
+++ b/src/view/com/util/post-embeds/QuoteEmbed.tsx
@@ -26,7 +26,7 @@ import {useQueryClient} from '@tanstack/react-query'
 import {HITSLOP_20} from '#/lib/constants'
 import {usePalette} from '#/lib/hooks/usePalette'
 import {InfoCircleIcon} from '#/lib/icons'
-import {moderatePost_wrapped} from '#/lib/moderatePost_wrapped'
+import {moderatePost_wrapped as moderatePost} from '#/lib/moderatePost_wrapped'
 import {makeProfileLink} from '#/lib/routes/links'
 import {s} from '#/lib/styles'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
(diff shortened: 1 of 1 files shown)


src/view/com/util/post-embeds/QuoteEmbed.tsx
```diff
@@ -295,10 +293,17 @@ export function QuoteX({onRemove}: {onRemove: () => void}) {
 
 export function LazyQuoteEmbed({uri}: {uri: string}) {
   const {data} = useResolveLinkQuery(uri)
-  if (!data || data.type !== 'record' || data.kind !== 'post') {
+  const moderationOpts = useModerationOpts()
+  if (
+    !data ||
+    data.type !== 'record' ||
+    data.kind !== 'post' ||
+    !moderationOpts
```
gaearon, line 301: is this correct? in other cases where this hook is used, we don't exit early if it returns `undefined`. we just leave the `moderation` undefined

mozzius: good question - there are also places where it is bundled in the loading state too. so would we prefer showing something unmoderated if the moderationOpts aren't ready, or would we rather show nothing? I'd lean toward waiting for both to load

gaearon: i'd rather show nothing but i'd also rather fix this consistently stacked on top of this PR than introduce inconsistency

gaearon: well i guess we wouldn't want layout to jump so we need to think more about this. when is it ever undefined?

mozzius: It's loading getPreferences I think. I think by the time you get to this point it should already be loaded

gaearon: it seems like realistically it can only be undefined before prefs have loaded (or if they never existed). if they have not loaded, the entire Home or Post screen anyway. so i think we should go with consistency here. in other words, it's not expected to change midway

mozzius: ok!

gaearon (approved):

PR #7215, opened 2024-12-20
Refine repost icon

quiple:
- **Before**
  스크린샷 2024-12-20 23 14 22 스크린샷 2024-12-20 23 14 07
  The weight of the repost icon in these two parts looks too light compared to the text.

- **After**
  스크린샷 2024-12-20 23 17 07 스크린샷 2024-12-20 23 17 24
  Uses the new icon with `stroke-weight` set to `2.5` in these two parts. And made some minor adjustments to the shape.

I've added the unoutlined path data to make it easier for others to modify:

```ts
export const Repost = createSinglePathSVG({
  path: 'M4.5 11V9C4.5 7.34315 5.84315 6 7.5 6H20M20 6L16.5 2.5M20 6L16.5 9.5M19.5 13V15C19.5 16.6569 18.1569 18 16.5 18H4M4 18L7.5 14.5M4 18L7.5 21.5',
})
```


quiple:
Added a `stroke-width: 2.5` version of the pin icon, making it exactly the same as the existing pin icon.

PR #7218, opened 2024-12-20
Don't add unused keydown listeners

gaearon:
Web-only.

We have hundreds of these listeners at any moment in time but usually they're not doing anything:

Screenshot 2024-12-20 at 17 15 21

They're only relevant when `open` is `true`, as seen in the code of the handlers.

So let's attach these listeners only if `open` is true.

## Test Plan

Place `console.log` into the listeners. Verify they run on every keystroke across the app.

Apply the fix.

Verify they no longer run.

Verify the hashtag link web dropdown menu still responds to "Escape".


estrattonbailey (approved):
Tested, nice one 👍

PR #7222, opened 2024-12-20
Align lightbox status bar behaviour on Android 15 with iOS

mozzius:
# Based on #7221 

Now that we have edge to edge mode, we can do what I always intended to do: hide the system bar with the other lightbox controls, like how iOS works

I don't know why the nav bar disappears so slowly - `react-native-edge-to-edge` claims that `expo-navigation-bars` uses depreciated android APIs, which might be it

https://github.com/user-attachments/assets/6c82d96e-2450-48b2-8622-86016a436b6e

### DIFF
--- a/src/view/com/lightbox/ImageViewing/components/ImageDefaultHeader.tsx
+++ b/src/view/com/lightbox/ImageViewing/components/ImageDefaultHeader.tsx
@@ -5,12 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  *
  */
-import {
-  SafeAreaView,
-  StyleSheet,
-  TouchableOpacity,
-  ViewStyle,
-} from 'react-native'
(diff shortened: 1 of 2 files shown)


gaearon (changes requested):
causes white flash on the bottom bar on my android 12

mozzius:
Let's put this on ice until we can figure it out. it's not urgent

zoontek:
It occurs because `expo-navigation-bar` is setting an opaque background, and the update doesn't occur in sync. By truly going edge-to-edge (transparent nav bar), there's no issue:

https://github.com/user-attachments/assets/a7a2508f-0db0-4d81-95f7-7562c0c17734

zoontek:
Preliminary work is here: https://github.com/bluesky-social/social-app/pull/7497

PR #7229, opened 2024-12-20
Apply avatar moderation to quote post

### LINKED ISSUE #7117: Avatars of users labelled for adult profile content aren't blurred in quotes
### Steps to Reproduce

1. Set adult content to 'Warn'.
2. Go to https://bsky.app/profile/juggermelon.bsky.social/post/3ld7yfriqgs22.

The outer avatar is blurred while the inner one is not.
This happens both in feed and detail view.

### Attachments

The example isn't explicit, but here's a screenshot of it (self-quote) with the irrelevant parts pixelated:

self-quote

Web (Desktop)

### Device Info

Vivaldi	7.0.3495.26 (Stable channel) (64-Bit)

Build version: 1.95.0; Bundle info: da56ef7 (prod); Bundle date: 24120823; Platform: web; Platform version: undefined

### Additional Information

The user in question is labelled `porn`, which interestingly doesn't collapse _all_ their posts unlike third-party labels applied to an account that I've set to 'Warn'.

That said, it would be nice if the labelling was a bit more specific and targeted avatar and banner image separately. I'd also like a function to still view the banner image like I can with the avatar if I click it.

c960657:
Avatars in quote posts are not blurred according to moderation.

The `avatarModeration` was introduced in #2325. Later the more general `moderation` attribute was introduced in #2550, which makes `avatarModeration` obsolete.

Fixes #7117.

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -567,7 +567,6 @@ let PostThreadItemLoaded = ({
                 timestamp={post.indexedAt}
                 postHref={postHref}
                 showAvatar={isThreadedChild}
-                avatarModeration={moderation.ui('avatar')}
                 avatarSize={24}
                 style={[a.pb_xs]}
               />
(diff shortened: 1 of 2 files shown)


gaearon:
I don't get it. Why aren't they equivalent? What is the source of divergence between the two calculations?

gaearon:
OK I got it, the issue isn't them being different, it's the old prop being optional (and not specified in a bunch of contexts). Makes sense.

gaearon (approved):
lgtm, thank you

PR #7235, opened 2024-12-21

### LINKED ISSUE #6728: Post time in Chinese cannot be localized on Android

auroursa:

**Update**: We are attempting to convert the language codes for Chinese and Cantonese from [RFC-1766](https://learn.microsoft.com/en-us/windows/win32/wmformat/language-strings) to [RFC-4646](https://www.iana.org/assignments/language-tags/language-tags.xhtml), to resolve language-related compatibility issues on native.

The following are the changes:

- `zh-CN` -> `zh-Hans-CN`
- `zh-TW` -> `zh-Hant-TW`
- `zh-HK` -> `zh-Hant-HK`

They are lengthy, but both Android and iOS are currently using them. and they are the recommended language code for Chinese by [IANA](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) and BCP 47.  The tricky issue with Chinese at the moment is that there are two conflicting standards, both of which are widely used.

The compatibility issue on Android stems from the differences in [Hermes](https://github.com/facebook/hermes/blob/main/doc/IntlAPIs.md) implementation of `Intl`. On Android, it is based on the [ICU4j](https://developer.android.com/guide/topics/resources/internationalization) project, which does not include RFC 1766 Chinese language codes like `zh-CN` since Android 7.0. This will be an attempt to fix #6728 without introducing polyfills.

Similar to #6774, tests have shown that `zh-Hans-CN`,  `zh-Hant-TW` and `zh-Hant-HK` can also be correctly used by app language.

Since these code changes are all related to Chinese adjustments (including modifications to `deviceLocales.ts`), I’ve included them in the regular Chinese localization update PR instead of opening a new one. This also helps our translators test the impact of the changes earlier.

The related changes to `deviceLocales.ts` can refer to #6621, except this time the conversion is reversed: from RFC 1766 to RFC 4646, as browsers still widely use RFC 1766. We still need them to correctly match the browser’s default language, and display Chinese when opening the web for the first time.

As far as I know, most open-source cross-platform apps that support Chinese have similar mapping mechanisms.I believe this is currently the best approach I’ve come up with for balancing Chinese display across platforms after repeated testing and research.

Note the impact of changing the language code is limited to the app display language, Cantonese still cannot be used as post language.

Co-maintained by @cirx1e @hakadao @VentusUta


src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,19 @@ export function getLocales() {
     }
 
     if (typeof locale.languageTag === 'string') {
-      if (locale.languageTag.startsWith('zh-Hans')) {
-        // Simplified Chinese to zh-CN
-        locale.languageTag = 'zh-CN'
+      if (
+        locale.languageTag === 'zh-CN' ||
```
auroursa, line 45: To ensure browser compatibility, as browsers still widely use RFC-1766.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,19 @@ export function getLocales() {
...
+        locale.languageTag.startsWith('zh-Hans')
```
auroursa, line 46: There are multiple `zh-Hans` / `zh-Hant` variants on iOS and Android, such as `zh-Hans-SG`. We want to match as many of these language variants as possible.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,19 @@ export function getLocales() {
...
+      ) {
+        // Simplified Chinese to zh-Hans-CN
+        locale.languageTag = 'zh-Hans-CN'
       }
-      if (locale.languageTag.startsWith('zh-Hant')) {
-        // Traditional Chinese to zh-TW
-        locale.languageTag = 'zh-TW'
-      }
-      if (locale.languageTag.startsWith('yue')) {
```
auroursa, line 52: After renaming `zh-HK` to `yue`, this is no longer needed.

auroursa:
I have considered directly using `zh-Hans` and `zh-Hant` instead of the current more complex expressions, but in my cross-testing, their compatibility wasn't good.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,13 @@ export function getLocales() {
     }
 
     if (typeof locale.languageTag === 'string') {
-      if (locale.languageTag.startsWith('zh-Hans')) {
-        // Simplified Chinese to zh-CN
-        locale.languageTag = 'zh-CN'
+      if (locale.languageTag === 'zh-CN') {
```
auroursa, line 44: To ensure browser compatibility, Chrome and Firefox still return `zh-CN`.

auroursa, line 44: To ensure browser compatibility, the current versions of Chrome and Firefox still use RFC 1766 and return `zh-CN`.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,13 @@ export function getLocales() {
...
+        locale.languageTag = 'zh-Hans-CN'
       }
-      if (locale.languageTag.startsWith('zh-Hant')) {
-        // Traditional Chinese to zh-TW
-        locale.languageTag = 'zh-TW'
-      }
-      if (locale.languageTag.startsWith('yue')) {
-        // Cantonese (Yue) to zh-HK
-        locale.languageTag = 'zh-HK'
+      if (locale.languageTag === 'zh-TW') {
```
auroursa, line 48: Same reason as above, we need to convert RFC 1766 to RFC 4646.

src/locale/deviceLocales.ts
```diff
@@ -41,17 +41,13 @@ export function getLocales() {
...
-        locale.languageTag = 'zh-CN'
+      if (locale.languageTag === 'zh-CN') {
+        // Simplified Chinese to zh-Hans-CN
+        locale.languageTag = 'zh-Hans-CN'
       }
-      if (locale.languageTag.startsWith('zh-Hant')) {
-        // Traditional Chinese to zh-TW
-        locale.languageTag = 'zh-TW'
-      }
-      if (locale.languageTag.startsWith('yue')) {
```
auroursa, line 52: After changing `zh-HK` to `yue`, this is no longer needed.

auroursa:
Over the past few months, I’ve been trying to understand why compatibility issues with Chinese keep occurring. Maybe this is unlikely to be implemented, but for Chinese, separating Web language code and Native App language code could greatly enhance the experience.

With the current code, I can only compromise on Web or Native. Previously I chose to compromise on Native, attempting to make iOS/Android match the Web language codes. Now I try to convert the Web language codes to match Native.

auroursa:
It looks good to me so far. Then will focus on the continuous improvement of the translation itself.  Ready to go.

For reference the most suitable Chinese language tags for the platforms:

**iOS**: `zh-Hans` / `zh-Hant`
`zh-Hans-CN` can be recognized, but it require a strict match with the system language set to Simplified Chinese + region set to China Mainland, if the locale is set to Canada for example, it will no longer match (outputting something unusual like `zh-Hans-CA`). With a similar rule for Traditional Chinese, this is why we still need **startsWith** matching. 

**Android**: `zh-Hans-CN` / `zh-Hant-TW`
It doesn't recognize `zh-Hans` at all, but uses `zh-Hans-CN`  doesn't have the strict matching like iOS, this will also solve the post time localization issue. So we decided to convert all platforms to this.

**Web**:  `zh-CN` / `zh-TW`
Almost all browsers have primitive support for RFC 4646 Chinese language tags. If the browser uses Simplified Chinese, it will still output `zh-CN`. This is why we need to map them after converting from RFC 1766 to RFC 4646. Edge briefly used `zh-Hans` as the default behavior during the EdgeHTML era, but after switching to the Blink engine, it reverted back to `zh-CN`.

Lastly, Merry Christmas and a Happy New Year!

pfrazee (approved):
Okay! Thanks so much!

PR #7248, opened 2024-12-22
Extend clickable area to height of container

c960657:
On web, if the text input looses focus, e.g. because you have used the media buttons in the composer footer, it is a bit tricky to refocus the text input in order to continue writing on the post text, because only a small strip at the top of the big white area of the compose dialog is clickable.

The green line shows the current clickable area. This PR extends the clickable area to the one indicated by red line.

image


mozzius (approved):
Great, tysm!

PR #7249, opened 2024-12-22
Correct font preload

tunetheweb:
Missed these updated from #6993.

This means the font is still being loaded twice:

https://www.webpagetest.org/result/241222_BiDcC9_64Z/1/details/cached/#waterfall_view_step1

Down from four times (as Italic is no longer preloaded), and smaller (as woff2), so it’s better, but still double what’s needed!


gaearon (approved):
lgtm

PR #7254, opened 2024-12-23
Cropper - opt out of edge-to-edge enforcement on Android 15

mozzius:
Breakdown of changes:

1. Changed theme to custom `UCropNoEdgeToEdge`
2. Added new theme which just extends the old theme
3. For API level 35+ (Android 15), add `<item name="android:windowOptOutEdgeToEdgeEnforcement">true</item>` to the theme

Screenshot 2024-12-23 at 13 44 40

# Test plan

Test main on Android 15 with 3-button navigation - observe overlap
Test this PR - overlap should be fixed
Test this PR on Android 14 or lower to check no regressions

### DIFF
--- a/patches/react-native-image-crop-picker+0.41.6.patch
+++ b/patches/react-native-image-crop-picker+0.41.6.patch
@@ -35,7 +35,7 @@ diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker
 index 9f20973..68d4766 100644
 --- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
 +++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-@@ -126,7 +126,8 @@ RCT_EXPORT_MODULE();
+@@ -126,7 +126,8 @@ - (void) setConfiguration:(NSDictionary *)options
  
  - (UIViewController*) getRootVC {
      UIViewController *root = [[[[UIApplication sharedApplication] delegate] window] rootViewController];
--- a/patches/react-native-image-crop-picker+0.41.6.patch
+++ b/patches/react-native-image-crop-picker+0.41.6.patch
@@ -1,8 +1,41 @@
+diff --git a/node_modules/react-native-image-crop-picker/android/src/main/AndroidManifest.xml b/node_modules/react-native-image-crop-picker/android/src/main/AndroidManifest.xml
+index 391f303..8e2c3db 100644
+--- a/node_modules/react-native-image-crop-picker/android/src/main/AndroidManifest.xml
++++ b/node_modules/react-native-image-crop-picker/android/src/main/AndroidManifest.xml
+@@ -24,7 +24,7 @@
+ 
+         <activity
+             android:name="com.yalantis.ucrop.UCropActivity"
+-            android:theme="@style/Theme.AppCompat.Light.NoActionBar" />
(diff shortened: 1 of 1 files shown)


patches/react-native-image-crop-picker+0.41.6.patch
```diff
@@ -1,8 +1,41 @@
...
++++ b/node_modules/react-native-image-crop-picker/android/src/main/res/values/styles.xml
+@@ -0,0 +1,3 @@
++<resources>
++    <style name="Theme.UCropNoEdgeToEdge" parent="Theme.AppCompat.Light.NoActionBar"/>
++</resources>
 diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
 index 9f20973..68d4766 100644
 --- a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
 +++ b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
-@@ -126,7 +126,8 @@ - (void) setConfiguration:(NSDictionary *)options
```
gaearon, line 5: is this accidental? doesn't seem related

gaearon (approved):
seems ok. are we going to be forever forked or do you have some plan for how to get this upstream?

also see unrelated change

patches/react-native-image-crop-picker+0.41.6.patch
mozzius: not sure why patch-packaged decided to change that. will rm

mozzius:
the edge-to-edge opt out will be removed in a future android version, so this is unfortunately just kicking the can down the road. I hope it gets patched upstream - there are issues open in both repos, but unfortunately both seem like pretty quiet projects. when the time comes we can assess our options.

gaearon:
might be good to send a PR upstream just to increase chances of this moving along

patches/react-native-image-crop-picker+0.41.6.patch
```diff
@@ -1,3 +1,36 @@
...
+diff --git a/node_modules/react-native-image-crop-picker/android/src/main/res/values/styles.xml b/node_modules/react-native-image-crop-picker/android/src/main/res/values/styles.xml
+new file mode 100644
+index 0000000..50129b6
+--- /dev/null
++++ b/node_modules/react-native-image-crop-picker/android/src/main/res/values/styles.xml
+@@ -0,0 +1,3 @@
++<resources>
++    <style name="Theme.UCropNoEdgeToEdge" parent="Theme.AppCompat.Light.NoActionBar"/>
++</resources>
 diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m
```
gaearon, line 34: this still seems unrelated, i think maybe it can be removed?

mozzius: No, that's needed, that's the default value of the theme. then /values-v35/styles.xml is the version just for SDK 35. android configs aren't forward compatible, so it needs to be forked explicitly

gaearon: no i mean this line `diff --git a/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m b/node_modules/react-native-image-crop-picker/ios/src/ImageCropPicker.m`, it seems unintentional

PR #7255, opened 2024-12-23
Localize embed component

c960657:
Just like the application itself, the embed component should be localized.

I suggest the component should match the language used in the embedding page, if available, with fallback to the users' preferred languages in the`Accept-Language` request header.


bskyembed/src/locale/locales/en/messages.ts
```diff
@@ -0,0 +1,4 @@
+/*eslint-disable*/ import type {Messages} from '@lingui/core'
```
auroursa, line 1: Maybe it should be added to `.gitignore` list just like [this](https://github.com/bluesky-social/social-app/blob/5020e4233a5a50be8fe7b8f360737da4d30ab74a/.gitignore#L112).

auroursa:
Just my opinion: `@lingui/macro` has been marked as deprecated in 5.0 and will be removed in the next major version. Since this is the first time localization is being introduced for the embed component, perhaps it’s worth trying the [new approach](https://lingui.dev/blog/2024/11/28/announcing-lingui-5.0#react-and-js-macros-were-split-to-separate-packages)?

There’s no rush, just a possible improvement direction that came to mind. Considering that the main project is still using lingui 4.x and actively relies on ·@lingui/macro·, it might make sense to stick with it for consistency.

c960657:
Thanks, I was not aware of that.

Switching from `@lingui/macro` to `@lingui/core/macro` was just a simple search-and-replace, so I don't think there is much consistency gained by staying on 4.x.

Issue #624, opened 2023-05-15
Rotate images on mobile.


changeling:
**Is your feature request related to a problem? Please describe.**

Opened images currently don't rotate upon device rotation on mobile devices. Without this ability, text can be difficult to read.

**Describe the solution you'd like**

I'd like images to rotate when expanded and mobile device is rotated.

williaia:
Very irritating to have to view a landscape image in portrait mode

chrismarquardt:
+1 
Especially when screenshots of text are involved, not being able to rotate them to landscape is a serious usability issue, I'd really love for Bluesky to be consistent with other social media apps in this regard

simonkbike:
+1
For reasons stated above. Don't understand why app is locked to portrait orientation

briankkeyser:
Many text images are too small to read unless rotated to landscape. Need this feature often.

drussellmrichie:
Just commenting that I still think this would be a great feature to add.

Viexi:
Commenting because I can't fathom how this hasn't been added already

mr-mac-chs:
Yup, still an issue. This is absolutely a serious useability issue that needs a fix.

chrismarquardt:
It has been over a year and images are still portrait only. Like many others, I consider image rotation essential and I don't get why it is being ignored. Feels as if BlueSky aren't using their own mobile app.

mr-mac-chs:
I also posted a "complaint" to @bsky.app about this. If you have a visual disability, this is really tough. I don't and I still find it challenging.

samuelgoff:
My only regret is that I have but one +1 to give for this PR.

In many cases, the **only** way to read landscape content in portrait mode is to zoom way in, cutting off 2/3 of the image which then necessitates scrolling 3 times per row, or save the image to smartphone so you can then open the image in a native image viewer so you can view it in landscape mode.

The is both an accessibility & just basic usability issue.

PR #7257, opened 2024-12-23
Unlock orientation when lightbox is open

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/624

Unlocks device rotation when lightbox is open

Details:
- Screen orientation is locked to portrait normally. on iOS this is acheived via the config plugin, on Android we just have to call `lockAsync` as soon as it starts up
- Locking/unlocked achieved via an effect in the lightbox component
- Safe area for iOS/Android 15 was changed to not include the sides either - this was 0 anyway for portrait mode, and now allows it to be full-screen when rotated
- Lightbox state (so zoom, position etc) is cleared via key whenever an orientation change happens to ensure that it doesn't get into a wonky state
- There's some sort of double nav bar inset when sideways on Android <15. It's entirely outside of the app, not anything we're doing as far as I can tell. Not sure what's up with that, I just painted it black.

https://github.com/user-attachments/assets/f81f8fe9-d654-4eaf-bfb4-5be916d956fb

https://github.com/user-attachments/assets/05fe6a97-c85c-4fae-bc66-405eb5e7817e
    
# Test plan

Test iOS, Android <=14, Android 15
Test rotating, test lightbox gesture behaviour while/after rotating
Try and test on device where possible

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -50,7 +50,6 @@ module.exports = function (config) {
       runtimeVersion: {
         policy: 'appVersion',
       },
-      orientation: 'portrait',
       icon: './assets/app-icons/ios_icon_default_light.png',
       userInterfaceStyle: 'automatic',
       primaryColor: '#1083fe',
@@ -346,6 +345,7 @@ module.exports = function (config) {
             },
           },
         ],
+        ['expo-screen-orientation', {initialOrientation: 'PORTRAIT_UP'}],
       ].filter(Boolean),
       extra: {
         eas: {
--- a/src/alf/util/navigationBar.ts
+++ b/src/alf/util/navigationBar.ts
@@ -1,4 +1,5 @@
 import * as NavigationBar from 'expo-navigation-bar'
+import * as SystemUI from 'expo-system-ui'
 
 import {isAndroid} from '#/platform/detection'
 import {Theme} from '../types'
@@ -9,10 +10,12 @@ export function setNavigationBar(themeType: 'theme' | 'lightbox', t: Theme) {
       NavigationBar.setBackgroundColorAsync(t.atoms.bg.backgroundColor)
       NavigationBar.setBorderColorAsync(t.atoms.bg.backgroundColor)
       NavigationBar.setButtonStyleAsync(t.name !== 'light' ? 'light' : 'dark')
+      SystemUI.setBackgroundColorAsync(t.atoms.bg.backgroundColor)
     } else {
       NavigationBar.setBackgroundColorAsync('black')
       NavigationBar.setBorderColorAsync('black')
       NavigationBar.setButtonStyleAsync('light')
+      SystemUI.setBackgroundColorAsync('black')
     }
   }
 }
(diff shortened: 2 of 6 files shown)


app.config.js
```diff
@@ -346,6 +345,7 @@ module.exports = function (config) {
             },
           },
         ],
+        ['expo-screen-orientation', {initialOrientation: 'PORTRAIT_UP'}],
```
mozzius, line 348: `initialOrientation` is iOS-only, annoyingly

src/alf/util/navigationBar.ts
```diff
@@ -9,10 +10,12 @@ export function setNavigationBar(themeType: 'theme' | 'lightbox', t: Theme) {
       NavigationBar.setBackgroundColorAsync(t.atoms.bg.backgroundColor)
       NavigationBar.setBorderColorAsync(t.atoms.bg.backgroundColor)
       NavigationBar.setButtonStyleAsync(t.name !== 'light' ? 'light' : 'dark')
+      SystemUI.setBackgroundColorAsync(t.atoms.bg.backgroundColor)
     } else {
       NavigationBar.setBackgroundColorAsync('black')
       NavigationBar.setBorderColorAsync('black')
       NavigationBar.setButtonStyleAsync('light')
+      SystemUI.setBackgroundColorAsync('black')
```
mozzius, line 18: Paints this red strip black when the lightbox is open

image

gaearon:
This transition feels very glitchy. Can we do something smarter here?

https://github.com/user-attachments/assets/6ef877ec-0124-49db-80a0-b7e5a75b5f7b

For example maybe we can always wait to update the lock state _before_ starting the animated transition.

gaearon:
And about this — I understand _why_ it happens but is there any way at all to make the rest of the app "keep the lock"? So that there's no "rotated feed" underneath it.

https://github.com/user-attachments/assets/47197a9f-614b-4e1b-996e-c167908d96aa

gaearon:
pushed a few fixes which should slightly help with cases above

gaearon (approved):

gaearon:
Still a little bit glitchy but let's get this in and iterate.

Skaldebane:
@gaearon you could maybe use the rotation sensor data? that way you could handle the rotation yourselves just for the lightbox, but keep the app in locked portrait mode in the OS's perspective

ofc while respecting the user's current rotation lock

Skaldebane:
as for the extra inset on android, i'm not sure how this works in RN but double insets (natively) usually mean that they were not consumed higher up the tree

e.g. if a parent applies nav bar insets (for whatever sides), it shall also consume them, so if a child does that again, it gets no insets applied for the consumed sides.

gaearon:
> ofc while respecting the user's current rotation lock preference

How would you read that? I don’t think it’s exposed.

Skaldebane:
@gaearon
yeah, doesn't seem to be exposed on iOS
it's exposed on Android though: https://stackoverflow.com/a/20728974/11055122

prohr:
Filed a separate bug #7452 to capture some of the remaining glitches

PS:  May be worth looking at how the custom video control deals with rotations, because that's much smoother

PR #7265, opened 2024-12-24
Fix Nepali language code

quiple:
The code for Nepali according to the ISO 639-1 language codes is `ne`.

[List of ISO 639 language codes - Wikipedia](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes)


whoisanku:
oh wow never knew that we shifted from np to ne! Great PR. @pfrazee this seems right two letter for Nepali language.

quiple:
> oh wow never knew that we shifted from np to ne! Great PR. @pfrazee this seems right two letter for Nepali language.

I believe the language code was originally `ne`, and the country code is `np`.

whoisanku:
Yes makes sense. Thanks for correcting :)

gaearon (approved):

Issue #7053, opened 2024-12-11
Change 'Delete Post' option text color to RED


pranavgoel29:
### Describe the Feature

**Current behaviour:**
The "Delete Post" option visually blends in with other options, making it harder for users to identify or distinguish it as a critical/destructive action.

**Proposed behaviour:** 
Change the text color of the "Delete Post" option to red.

#### Current UI:
<img width="249" alt="image" src="https://github.com/user-attachments/assets/6ab188dc-0da9-4e0d-8a76-79e061a4737a" />

#### Suggested UI:
<img width="242" alt="image" src="https://github.com/user-attachments/assets/dd35f38e-58bf-47fb-922c-ea433f440828" />

PR #7276, opened 2024-12-25
Change 'Delete Post' option text color to RED #7053

maharkus:
Add red text and icon color to delete own post.


maharkus (commented):
Add red text and icon color to delete own post.

maharkus (commented):
maharkus
wants to merge changes into
bluesky-social:main
from
maharkus:change-delete-post-color-to-red

PR #7281, opened 2024-12-26
Fix `moveToPermanentPath` ext Default Missing '.'

foohyfooh:


mozzius (approved):
nice spot

gaearon:
Verified it doesn’t break any related flows? Sometimes we accidentally depend on bugs.

PR #7293, opened 2024-12-27
Fix mute words in trending

### LINKED ISSUE #7284: Trending topics not always filtering out muted words due to apparent race condition
### Steps to Reproduce

1. Make sure you have at least one mute word configured that _should_ filter out a trending topic
2. Click the search icon to see trending topics
3. Refresh the page
4. Note that the trending topic that should be filtered out is still there (or at least _sometimes_ is still there depending on how close this race is)

You can additionally add a break point to the `useTrendingTopics` hook and note that the `muteWords` array is empty (see attachment).

This could be fixed by adding mutedWords to the query key, but I think a better solution would be to move the filtering outside the query since the query itself does not actually rely on the muted words. 

### Attachments

mute words list is empty but shouldn't be

Web (Desktop)

### Device Info

_No response_

Build version: 1.96.2; Bundle info: 3aa1a5f (prod); Bundle date: 24122022; Platform: web; Platform version: undefined

### Additional Information

_No response_

estrattonbailey:
Fixes #7284

Moves mute word filtering into the `select` handler. This matches other queries in the app.

### DIFF
--- a/src/state/queries/trending/useTrendingTopics.ts
+++ b/src/state/queries/trending/useTrendingTopics.ts
@@ -28,7 +28,10 @@ export function useTrendingTopics() {
       const {data} = await agent.api.app.bsky.unspecced.getTrendingTopics({
         limit: DEFAULT_LIMIT,
       })
-      return data
+      return {
+        topics: data.topics ?? [],
+        suggested: data.suggested ?? [],
+      }
     },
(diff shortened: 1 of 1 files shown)


haileyok (approved):
tested against repo listed in issue, works.

PR #7299, opened 2024-12-28
Update French translation (1.96.3 or 1.97)

Signez:
Trending topics are live! But the UI around it is not translated… let's fix that.

Very trivial translation work here (around 30 strings), should be quite trivial to review between two holiday meals — if you are inclined to do so! — dear @surfdude29 and/or @CerryTsuki ✨

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -15,7 +15,7 @@ msgstr ""
 
 #: src/components/ProgressGuide/FollowDialog.tsx:567
 msgid "(active)"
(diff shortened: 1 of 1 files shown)


pfrazee (approved):
Checked with google translate, looks good! Thanks!

surfdude29:
Oops, sorry @Signez I've been pretty much taking a break from GitHub over the holidays so I only just saw this 😬

But I just went through the changes and couldn't spot anything, so here's a late LGTM! ✅😅

PR #7312, opened 2024-12-31
Disable taps while swiping between feeds

gpp-0:
I've been encountering this behavior on mobile, where sometimes (a lot of the times!) swiping horizontally to go from feed A to feed B registers a tap on feed A. So it would move to the next feed and after a small delay it would unexpectedly expand a post or an image from the previous feed, which makes for a very annoying experience:

https://github.com/user-attachments/assets/ede1d500-1831-4c69-a2ed-6ae0d05722cb

This is most likely a bug with react-native-pager-view, callstack/react-native-pager-view#776.

As a workaround, I made it so that while the page is being dragged, touch events are captured by the corresponding view, so hopefully no presses are triggered further down. It seems to work, and I cannot reproduce the behavior anymore, **except** in the case where one is at the first/last page and swipes away from the edge, like so:

https://github.com/user-attachments/assets/3fe75717-9c02-4446-a581-2e0e2c19b588

### DIFF
--- a/src/screens/Hashtag.tsx
+++ b/src/screens/Hashtag.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import {ListRenderItemInfo, View} from 'react-native'
+import {runOnJS} from 'react-native-reanimated'
 import {PostView} from '@atproto/api/dist/client/types/app/bsky/feed/defs'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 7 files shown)


gaearon:
Thanks for looking into this. I agree it’s very annoying and we need to get some fix in.

Would you be open to looking into the underlying cause? We really should fix it at the pager level. If we know what the fix is, patching it in would be fine.

gpp-0:
I tried, but I don't really know enough about gesture stuff to make a proper fix, sorry. I'll look into it some more but I'm not very confident

gaearon:
I narrowed it down to a small repro case: https://github.com/callstack/react-native-pager-view/issues/960

PR #7340, opened 2025-01-03
feat: long press like button to see likes

clicktodev:
Implement long press action on the like button to navigate to the list of users who liked the post.

It's nice to be able to quickly see the likes without having to open the post first.

https://github.com/user-attachments/assets/6a74afa0-10cc-4bd7-922a-14054528afd5


hichemfantar:
@gaearon i found this pretty useful, what do you think?

gaearon:
interesting. any precedent for this in popular apps?

hichemfantar:
I can’t find other apps with the same behavior and I don’t remember where I got the idea from. It just felt natural to me.
Maybe try it out and let me know how it feels.
Who says bsky can’t be the first to introduce this :)
My only concern is if you plan to introduce multi reaction likes like in fb and linked in.

PR #7344, opened 2025-01-03
SDK@next migration

estrattonbailey:


src/state/queries/starter-packs.ts
```diff
@@ -131,7 +131,7 @@ export function useCreateStarterPackMutation({
 
       return await agent.app.bsky.graph.starterpack.create(
         {
-          repo: agent.session?.did,
+          repo: agent.session?.did!,
```
matthieusieben, line 134:
```suggestion
          repo: agent.assertDid,
```

src/state/queries/post-thread.ts
```diff
@@ -360,7 +359,7 @@ function responseToThreadNodes(
         depth,
         isHighlightedPost: depth === 0,
         hasMore:
-          direction === 'down' && !node.replies?.length && !!node.replyCount,
+          direction === 'down' && !node.replies?.length && !!post.replyCount,
```
estrattonbailey, line 362: I think this was actually a typo, which was masked by the `[key: string]: undefined` loose types we use to have here. `node` does not have a `replyCount` on it at all, but `node.post` does.

src/state/queries/notifications/util.ts
```diff
@@ -255,8 +256,14 @@ function getSubjectUri(
...
-      AppBskyFeedRepost.isRecord(notif.record) ||
-      AppBskyFeedLike.isRecord(notif.record)
+      atp.fastIsType<AppBskyFeedRepost.Record>(
+        notif.record,
+        AppBskyFeedRepost.isRecord,
+      ) ||
+      atp.fastIsType<AppBskyFeedLike.Record>(
+        notif.record,
+        AppBskyFeedLike.isRecord,
+      )
```
estrattonbailey, line 266: The alternative is casting these objects, or using `isValidRecord` for each, which runs full lexicon validation.

src/state/queries/post-thread.ts
```diff
@@ -329,8 +330,10 @@ function responseToThreadNodes(
 ): ThreadNode {
   if (
     AppBskyFeedDefs.isThreadViewPost(node) &&
-    AppBskyFeedPost.isRecord(node.post.record) &&
-    AppBskyFeedPost.validateRecord(node.post.record).success
+    atp.fastIsType<AppBskyFeedPost.Record>(
+      node.post.record,
+      AppBskyFeedPost.isRecord,
+    )
```
estrattonbailey, line 336: This effectively removes full record validation from this check. Double checking that we don't need this [here in Slack.](https://bluesky-builders.slack.com/archives/C04CY4895NK/p1736207759085869)

estrattonbailey: Yep we're good 👍

src/state/queries/messages/actor-declaration.ts
```diff
@@ -69,12 +69,10 @@ export function useDeleteActorDeclaration() {
   return useMutation({
     mutationFn: async () => {
       if (!currentAccount) throw new Error('Not logged in')
-      // TODO(sam): remove validate: false once PDSes have the new lexicon
```
estrattonbailey, line 72: @mozzius any reason this is still here?

src/types/atproto/index.ts
```diff
@@ -0,0 +1,25 @@
...
+ * and although the `isValid*` utils do, they also fully validate the object
+ * shape, which has a performance cost. This util allows us to prescribe the
+ * type we expect, while only checking the `$type` value of the record.
+ *
+ * Usage:
+ * ```ts
+ * import * as atp from '#/types/atproto'
+ *
+ * if (atp.dangerousIsType<AppBskyFeedPost.Record>(node.post.record, AppBskyFeedPost.isRecord)) {
+ * }
```
matthieusieben, line 17: Nit:

```suggestion
 * if (atp.dangerousIsType<AppBskyFeedPost.Record>(item, AppBskyFeedPost.isRecord)) {
 *   // `item` has type `$Typed<AppBskyFeedPost.Record>` here
 * }
```

src/types/atproto/post.ts
```diff
@@ -0,0 +1,136 @@
...
+  | {
+      type: 'labeler'
+      view: AppBskyLabelerDefs.LabelerView
+    }
+  | {
+      type: 'starter_pack'
+      view: AppBskyGraphDefs.StarterPackViewBasic
+    }
+  | {
+      type: 'images'
```
matthieusieben, line 46: You probably want singular here ?

src/types/atproto/post.ts
```diff
@@ -0,0 +1,136 @@
...
+      type: 'labeler',
+      view: view.record,
+    }
+  } else if (AppBskyGraphDefs.isStarterPackViewBasic(view.record)) {
+    return {
+      type: 'starter_pack',
+      view: view.record,
+    }
+  }
+}
```
matthieusieben, line 107: Very much a nit: you can slightly reduce the code footprint here by extracting the `record` into a var.

```suggestion
export function parseEmbedView(
  { record }: AppBskyEmbedRecord.View,
): View | undefined {
  if (AppBskyEmbedRecord.isViewRecord(record)) {
    return {
      type: 'post',
      view: record,
    }
  } else if (AppBskyEmbedRecord.isViewNotFound(record)) {
    return {
      type: 'post_not_found',
      view: record,
    }
  } else if (AppBskyEmbedRecord.isViewBlocked(record)) {
    return {
      type: 'post_blocked',
      view: record,
    }
  } else if (AppBskyEmbedRecord.isViewDetached(record)) {
    return {
      type: 'post_detached',
      view: record,
    }
  } else if (AppBskyFeedDefs.isGeneratorView(record)) {
    return {
      type: 'feed',
      view: record,
    }
  } else if (AppBskyGraphDefs.isListView(record)) {
    return {
      type: 'list',
      view: record,
    }
  } else if (AppBskyLabelerDefs.isLabelerView(record)) {
    return {
      type: 'labeler',
      view: record,
    }
  } else if (AppBskyGraphDefs.isStarterPackViewBasic(record)) {
    return {
      type: 'starter_pack',
      view: record,
    }
  }
}
```

src/types/atproto/starterPack.ts
```diff
@@ -0,0 +1,21 @@
...
+   * Renamed to clarify source of this util, but directly aliases the original.
+   * {@link AppBskyGraphDefs.isStarterPackViewBasic}
+   */
+  isStarterPackViewBasic as isBasicView,
+  /**
+   * Renamed to clarify source of this util, but directly aliases the original.
+   * {@link AppBskyGraphDefs.isStarterPackView}
+   */
+  isStarterPackView as isView,
+} from '@atproto/api/dist/client/types/app/bsky/graph/defs'
```
matthieusieben, line 14: Relying on the internal folder structure of a dependency is not ideal. Any reason why you are not doing the following instead ?

```suggestion
export const isBasicView = AppBskyGraphDefs.isStarterPackViewBasic
export const isView = AppBskyGraphDefs.isStarterPackView
```

src/lib/api/feed-manip.ts
```diff
@@ -55,16 +55,15 @@ export class FeedViewPostsSlice {
...
     }`
     if (feedPost.post.uri === FALLBACK_MARKER_POST.post.uri) {
       this.isFallbackMarker = true
       return
     }
-    if (
-      !AppBskyFeedPost.isRecord(post.record) ||
-      !AppBskyFeedPost.validateRecord(post.record).success
-    ) {
+    if (!AppBskyFeedPost.isValidRecord(post.record)) {
```
estrattonbailey, line 66: I don't think we need these validations here, but they were already in place and I wanted to double check if they're necessary. Left them for now, will revisit before this merges.

src/state/cache/thread-mutes.tsx
```diff
@@ -69,7 +69,9 @@ function useMigrateMutes(setThreadMute: SetStateContext) {
         while (!cancelled) {
           const threads = persisted.get('mutedThreads')
 
-          const root = threads.findLast(uri => uri.includes(currentAccount.did))
+          const root = threads
+            .reverse()
+            .find(uri => uri.includes(currentAccount.did))
```
estrattonbailey, line 74: I'm not sure why TS got bad about this suddenly, but decided to do this instead of rabbit holing any further on this PR.

src/state/messages/convo/agent.ts
```diff
@@ -511,17 +511,17 @@ export class Convo {
   private pendingFetchConvo:
     | Promise<{
         convo: ChatBskyConvoDefs.ConvoView
-        sender: AppBskyActorDefs.ProfileViewBasic | undefined
-        recipients: AppBskyActorDefs.ProfileViewBasic[]
+        sender: ChatBskyActorDefs.ProfileViewBasic | undefined
+        recipients: ChatBskyActorDefs.ProfileViewBasic[]
```
estrattonbailey, line 515: Good example of how we were actually using the wrong types here, but it was never an issue because they were loosely equal until now (and they're essentially the same object).

src/state/messages/convo/agent.ts
```diff
@@ -1031,7 +1031,10 @@ export class Convo {
            * `getItems` is only run in "active" status states, where
            * `this.sender` is defined
            */
-          sender: this.sender!,
+          sender: {
+            $type: 'chat.bsky.convo.defs#messageViewSender',
+            did: this.sender!.did,
+          },
```
estrattonbailey, line 1037: Here's an example of passing the full `ProfileView*` where only `did` was needed that was previously possible with the old loose types. This was only used to construct "pending" messages and never written to the user's repo, but it's good to be precise.

src/state/queries/messages/list-conversations.tsx
```diff
@@ -152,7 +161,11 @@ export function ListConvosProviderInner({
               function filterConvoFromPage(
                 convo: ChatBskyConvoDefs.ConvoView[],
               ) {
-                return convo.filter(c => c.id !== log.convoId)
+                return convo.filter(
+                  c =>
+                    c.id !==
+                    (log as ChatBskyConvoDefs.LogCreateMessage).convoId,
```
estrattonbailey, line 167: This is checked above, but within this nested closure TS loses it. This all runs synchronously, so I took a shortcut instead of doing the same checks again.

src/state/queries/messages/list-conversations.tsx
```diff
@@ -110,15 +110,24 @@ export function ListConvosProviderInner({
...
-                    }
-                  : convo,
-              ),
+              optimisticUpdate(log.convoId, old, convo => {
+                if (
+                  (ChatBskyConvoDefs.isDeletedMessageView(log.message) ||
+                    ChatBskyConvoDefs.isMessageView(log.message)) &&
+                  (ChatBskyConvoDefs.isDeletedMessageView(convo.lastMessage) ||
+                    ChatBskyConvoDefs.isMessageView(convo.lastMessage))
+                ) {
```
estrattonbailey, line 119: Just being overly cautious here

src/state/queries/notifications/util.ts
```diff
@@ -205,12 +206,9 @@ async function fetchSubjects(
...
   )
   const postsMap = new Map<string, AppBskyFeedDefs.PostView>()
-  const packsMap = new Map<string, AppBskyGraphDefs.StarterPackView>()
+  const packsMap = new Map<string, AppBskyGraphDefs.StarterPackViewBasic>()
   for (const post of postsChunks.flat()) {
-    if (
-      AppBskyFeedPost.isRecord(post.record) &&
-      AppBskyFeedPost.validateRecord(post.record).success
-    ) {
+    if (AppBskyFeedPost.isRecord(post.record)) {
```
estrattonbailey, line 211: Removed validation from critical path

src/state/shell/composer/index.tsx
```diff
@@ -21,7 +20,7 @@ export interface ComposerOptsPostRef {
   cid: string
   text: string
   author: AppBskyActorDefs.ProfileViewBasic
-  embed?: AppBskyEmbedRecord.ViewRecord['embed']
+  embed?: AppBskyFeedDefs.PostView['embed']
```
estrattonbailey, line 23: This type was just entirely wrong. `embed` does not exist on `AppBskyEmbedRecord.ViewRecord`. All the usages, however, were correct 😅

src/types/atproto/post.ts
estrattonbailey: Mayyyybe. `AppBskyEmbedImages` is plural, and `view.images[]` is an array, so I think `images` is probably fine here.

src/types/atproto/index.ts
```diff
@@ -0,0 +1,26 @@
+export * as profile from '#/types/atproto/profile'
```
matthieusieben, line 1: NIT: While `dangerousIsType` is indeed "Atprotoy" (has to do with the way atproto allow "open" data), the `profile` &  `starterPack` are not really part of "atproto".

Wouldn't it make more sense to name this something like `bsky` or `api` (contextualized by the app) ?

estrattonbailey: Oh yeah that's a solid idea 👍

src/components/StarterPack/QrCodeDialog.tsx
```diff
@@ -77,7 +81,11 @@ export function QrCodeDialog({
       } else {
         setIsProcessing(true)
 
-        if (!AppBskyGraphStarterpack.isRecord(starterPack.record)) {
+        if (
+          !asPredicate(AppBskyGraphStarterpack.validateRecord)(
+            starterPack.record,
+          )
```
matthieusieben, line 87: This might be more readable (& consistent with `dangerousIsType`) if you created another util in the `atp` namespace:

```suggestion
          !atp.safelyValidate(
            starterPack.record,
            AppBskyGraphStarterpack.validateRecord,
          )
```

```ts
// atp.ts

function safelyValidate<T, V = unknown>(
  value: V,
  validate: (value: unknown) => ValidationResult<T>,
): value is T & V {
  return validate(value).success
}
```


WDYT ?

src/types/atproto/post.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  } else {
+    return {
+      type: 'unknown',
+      view: null,
+    }
+  }
+}
+
+export function parseEmbed(embed: AppBskyFeedDefs.PostView['embed']): Embed {
```
matthieusieben, line 118: This function could support invalid records created with a `#main` suffix, transforming them into valid parsed embeds.

src/components/dms/MessagesListHeader.tsx
```diff
@@ -157,9 +157,12 @@ function HeaderReady({
...
 
-  const latestMessageFromOther = convoState.items.findLast(
-    item => item.type === 'message' && item.message.sender.did === profile.did,
-  )
+  const latestMessageFromOther = convoState.items
+    .reverse()
+    .find(
+      item =>
+        item.type === 'message' && item.message.sender.did === profile.did,
+    )
```
matthieusieben, line 165: Beware that `reverse()` will reverse the list in place. This could cause the list for be reversed on every render, causing the "found" item to be different every time.

If `findLast` doesn't work, you can clone the list before reversing it.

```suggestion
  const latestMessageFromOther = [...convoState.items]
    .reverse()
    .find(
      item =>
        item.type === 'message' && item.message.sender.did === profile.did,
    )
```

matthieusieben (commented):
Looks great. Some minor comments.

src/components/StarterPack/QrCodeDialog.tsx
estrattonbailey: Yeah I dig that. I want to go through and rename `#/types/atproto` to bsky in some places, as you suggested as well. Gonna do a pass on that in a sec, good idea to colocate this there too.

src/state/queries/messages/actor-declaration.ts
estrattonbailey: @mozzius same question heh

mozzius: Almost certainly can remove yep

mozzius (approved):
💎

The new utilities look great. Let's get it in so we can test

PR #7347, opened 2025-01-03
[@next] Profile handling, migrate `ProfileCard`

estrattonbailey:
The new SDK adds more strict types and now specifies the `$type` prop on all interfaces. Whereas previously, `ProfileViewBasic` as loosely equal to `ProfileView` and `ProfileViewDetailed`, these now result in type errors. And correctly so.

> NB: These new types still retain open-endedness e.g. `{ [key: string]: unknown }` to allow evolution.

Our app relies a lot of this loose equality e.g.:
- `FollowButton` is passed various views, and it can work with any of them
- `ProfileCard` is passed various views, and has truthy checks for things like `profile.description`, which isn't available on `Basic` views
- `precacheProfile` expects `Basic`, but we feed it `ProfileView` in some cases

In all these cases, it's _technically_ feasible to pass in any of these and handle specific cases downstream. However, we now get specific type errors, which we can use to more safely guard against improper references.

### So now what

This PR started with the old `ProfileCard` since it was a good example of this problem. That component references `profile?.description`, which is only available on `ProfileView` or `ProfileViewDetailed`. One option is to omit the new `$type` values like `Omit<AppBskyActorDefs.ProfileViewBasic, '$type'>`, which then falls back to the open-endedness and allows any profile view to be passed. This sorta works, but it's not clear what's happening to other devs, and it becomes unclear why we'd be using identity utils like `AppBskyActoDefs.isProfileViewDetailed()` where we need to. Example:
- Q: The type says it's a detailed view, why do I need to check again?
- A: Because it's not actually a detailed view in some cases, the types are just overly permissive.

So this PR introduces a new interface for profile view types, intended to be imported as `import * as Profile from "#/types/profile"`. It renames the identity utils for easier access, exports an `AnyView` for cases where we specifically allow multiple view types, and provides utils to _downgrade_ views to more simpler views. The opposite is not possible without lossy definitions e.g. `profile.description` doesn't exist on `ProfileViewBasic` and so couldn't be upgraded into `ProfileView`.

**The intention here** is to begin to use `Profile.AnyView` in place of separate profile views for cases where we allow multiple view types. Then, to use the identity methods to more safely fork logic based on the view type. And in cases where we need to, we can downgrade views to match what's expected by certain utils e.g. `precacheProfile(Profile.anyToBasic({...profileViewDetailed}))` which expects a `ProfileViewBasic` and can be derived from any profile view type.

> If it wasn't clear, profile views have varying levels of detail. In order of least -> most: `ProfileViewBasic`, `ProfileView`, and `ProfileViewDetailed`. Hence the "downgrade" or "upgrade" words I'm using here.

### Why not fix types everywhere to be specifically what we need?
We could, but I think this would be quite an undertaking and could cause some subtle errors.

Some valid use cases like `FollowButton` that only rely on a few props of a profile view _could_ be refactored to _only_ require those props instead of a profile view directly, but this would again require a lot of edits. This should probably be a best practice going forward where possible instead of passing around whole objects.

### Any similar issues to this?
Yep, `StarterPackView` and `StarterPackViewBasic`.

### DIFF
--- a/src/types/atproto/index.ts
+++ b/src/types/atproto/index.ts
@@ -1,0 +1,1 @@
+export * as profile from '#/types/atproto/profile'
--- a/src/types/atproto/profile.ts
+++ b/src/types/atproto/profile.ts
@@ -21,15 +21,17 @@ export {
 /**
  * Matches any profile view exported by our SDK
  */
-export type AnyView =
+export type AnyProfileView =
   | AppBskyActorDefs.ProfileViewBasic
   | AppBskyActorDefs.ProfileView
   | AppBskyActorDefs.ProfileViewDetailed
 
 /**
- * Downgrades any profile view to the most basic form.
+ * Maps any profile view type to `ProfileViewBasic`.
  */
-export function anyToBasic(view: AnyView): AppBskyActorDefs.ProfileViewBasic {
+export function anyToBasic(
+  view: AnyProfileView,
+): AppBskyActorDefs.ProfileViewBasic {
   return {
     $type: 'app.bsky.actor.defs#profileViewBasic',
     did: view.did,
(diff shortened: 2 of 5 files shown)


estrattonbailey:
Re-orged this a bit so that the idea way to use this is now `import * as atp from '#/types/atproto'`. Gonna keep workshopping from the base branch and we'll discuss async 👍

PR #806, opened 2023-05-30
Fix button to find people to follow

loganrosen:
Conditionally navigating to `Search` on Web instead of `SearchTab` (mobile-only) for the empty feeds. Closes #805.

PR #7360, opened 2025-01-04
Fix Explore link in feed interstitial

c960657:
For small screens, the link in the feed interstitial does not work. Same fix as in #806.

BTW, I am confused about the reference to the “Explore page”. AFAICT this term isn't used anywhere else. Should the button be relabeled?


c960657:
Fixed in #8733.

PR #7366, opened 2025-01-05
Update Romanian Translation

Laszlo19:
This PR updates all missing strings into romanian and corrects certain words (uses better terminology).

The translation was done on Crowdin and was checked by multiple people.

### DIFF
--- a/src/locale/locales/ro/messages.po
+++ b/src/locale/locales/ro/messages.po
@@ -1,21 +1,26 @@
 msgid ""
 msgstr ""
-"POT-Creation-Date: 2024-11-17 11:01+0200\n"
+"POT-Creation-Date: 2023-11-05 16:01-0800\n"
 "MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=utf-8\n"
+"Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: @lingui/cli\n"
 "Language: ro\n"
-"Project-Id-Version: \n"
+"Project-Id-Version: tempp\n"
 "Report-Msgid-Bugs-To: \n"
-"PO-Revision-Date: 2024-11-28 21:20+0000\n"
+"PO-Revision-Date: 2025-01-05 17:17\n"
 "Last-Translator: \n"
-"Language-Team: claudiucristea.ro; alextecplayz\n"
-"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : (n==0 || (n%100 > 0 && n%100 < 20)) ? 1 : 2);\n"
+"Language-Team: Romanian\n"
+"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : (n==0 || (n%100>0 && n%100<20)) ? 1 : 2);\n"
+"X-Crowdin-Project: tempp\n"
+"X-Crowdin-Project-ID: 737695\n"
+"X-Crowdin-Language: ro\n"
+"X-Crowdin-File: messages.po\n"
+"X-Crowdin-File-ID: 14\n"
 
-#: src/components/ProgressGuide/FollowDialog.tsx:567
+#: src/components/ProgressGuide/FollowDialog.tsx:563
 msgid "(active)"
-msgstr ""
+msgstr "(activ)"
 
 #: src/screens/Messages/components/ChatListItem.tsx:130
 msgid "(contains embedded content)"
@@ -24,7 +29,7 @@ msgstr "(conține conținut încorporat)"
 #: src/screens/Settings/AccountSettings.tsx:65
 #: src/view/com/modals/VerifyEmail.tsx:150
 msgid "(no email)"
-msgstr "(fără e-mail)"
+msgstr "(niciun e-mail)"
 
 #: src/view/com/notifications/FeedItem.tsx:232
 #: src/view/com/notifications/FeedItem.tsx:327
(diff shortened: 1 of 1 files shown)


surfdude29:
Thanks for this!

I'm assuming they're part of the Crowdin but just in case, pinging the two contributors to #6456: @claudiu-cristea & @alextecplayz

src/locale/locales/ro/messages.po
```diff
@@ -33,64 +38,64 @@ msgstr "(fără e-mail)"
 
 #: src/lib/hooks/useTimeAgo.ts:156
 msgid "{0, plural, one {# day} other {# days}}"
-msgstr "{0, plural, one {o zi} few {# zile} other {# de zile}}"
+msgstr "{0, plural, one {# zi} few {# zile} other {# zile}}"
```
claudiu-cristea, line 41: This change is wrong. The initial string was correct

src/locale/locales/ro/messages.po
```diff
@@ -33,64 +38,64 @@ msgstr "(fără e-mail)"
...
 
 #: src/screens/Profile/ProfileFollowers.tsx:40
 msgid "{0, plural, one {# follower} other {# followers}}"
-msgstr ""
+msgstr "{0, plural, one {# urmăritor}  other {# urmăritori}}"
```
claudiu-cristea, line 45:
```suggestion
msgstr "{0, plural, one {un urmăritor} few {# urmăritori} other {# de urmăritori}}"
```

claudiu-cristea:
I didn't cross the whole file but the changes to strings with plural are wrong and are not accounting for Romania 3 variants plurals

src/locale/locales/ro/messages.po
```diff
@@ -212,122 +217,122 @@ msgstr "{count} elemente necitite"
 #: src/lib/generate-starterpack.ts:108
 #: src/screens/StarterPack/Wizard/index.tsx:178
 msgid "{displayName}'s Starter Pack"
-msgstr "Pachetul de început al {displayName}"
+msgstr "Pachetul de pornire al lui {displayName}"
```
claudiu-cristea, line 220: It was decided in https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498121422 to use "pachet de început". See the [glossary](https://github.com/bluesky-social/social-app/pull/6456#issue-2665899849)

src/locale/locales/ro/messages.po
```diff
@@ -212,122 +217,122 @@ msgstr "{count} elemente necitite"
...
-msgstr "{firstAuthorLink} și <0>{additionalAuthorsCount, plural, one {un alt utilizator} few {alți {formattedAuthorsCount} utilizatori} other {alți {formattedAuthorsCount} de utilizatori}}</0> au redistribuit postarea ta"
+msgstr "{firstAuthorLink} și <0>{additionalAuthorsCount, plural, one {{formattedAuthorsCount} alți} few {{formattedAuthorsCount} alți} other {{formattedAuthorsCount} alți}}</0> v-au repostat postarea"
 
 #: src/view/com/notifications/NotificationFeedItem.tsx:351
 msgid "{firstAuthorLink} and <0>{additionalAuthorsCount, plural, one {{formattedAuthorsCount} other} other {{formattedAuthorsCount} others}}</0> signed up with your starter pack"
-msgstr "{firstAuthorLink} și <0>{additionalAuthorsCount, plural, one {un alt utilizator} few {alți {formattedAuthorsCount} utilizatori} other {alți {formattedAuthorsCount} de utilizatori}}</0> s-au înscris folosind pachetul tău de început"
+msgstr "{firstAuthorLink} și <0>{additionalAuthorsCount, plural, one {{formattedAuthorsCount} alți} few {{formattedAuthorsCount} alți} other {{formattedAuthorsCount} alți}}</0> s-au înregistrat cu pachetul dvs. de pornire"
```
claudiu-cristea, line 248: We are using the informal language ("tău", not "dumneavoastră" or "dvs."). See https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498593586

claudiu-cristea (commented):
Overall, this looks like a full rework of the Romanian localization. I think we should start from existing strings and rules established in #6456

Laszlo19:
Hi, sorry, I am new to Github and I did not see that PR

Overall my translation is based on official translation of Twitter (Using it for a few years their translation is engraved in my mind) and other platforms and companies like Microsoft, mainly fixing incorect terms like:

-mute/unmute are better translated as amuțire/dezamuțire, similarly to Discord

-handle that cannot be translated into nume de utilizator because "user name" alredy exists and is very confusing to a new user of the app, I wanted to sign up and I was very confused. As a person with a backround in translation, the perfectionist in me was pissed off and wanted to took the matters in my own hands, and it turned into a complete remake.

Im used to translations being as faithful to the original as posible, always refer to the user with Dvs., find translations that sound better hence "pachet de pornire".

Laszlo19:
For example, "# zi" is more correct as # is replaced by a number "1 zi", using "o zi" technically breaks and does not work as it can't show the number 1 because there is nothing to replace.
"de" from de zile is not necesary because "20 zile" and "20 de zile" mean the same thing, and overall is better to be consistent.

claudiu-cristea:
> (...) because "20 zile" and "20 de zile" mean the same thing, (...)

Of course they mean the same thing just that is not correct Romanian plural. The Romanian language has 3 plurals (singular + 2 plural variants). The variation is defined by the formula from the `.po` file header:

```
nplurals=3; plural=(n==1 ? 0 : (n==0 || (n%100 > 0 && n%100 < 20)) ? 1 : 2);
```

Using a single plural form is a relic of old times when there was no implementation of plural variants because 99.99% of the software was English and software localization was not yet an issue. The [gettext](https://www.gnu.org/software/gettext/) translation project is aware of this particularity of some languages and introduced the possibility of `.po` files to store multiple plural variants for languages that needs them. Then software projects know how to deal with such languages based on the _items count_ and show the correct string in UIs.

>  and overall is better to be consistent

Consistent with English? As is a Romanian localization we should be consistent with Romanian language rules.

> -mute/unmute are better translated as amuțire/dezamuțire, similarly to Discord

I don't have a strong opinion,  but "dezamuțire" doesn't sound good. I would let @alextecplayz discuss this as she proposed the actual form in https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498121422. 

> -handle that cannot be translated into nume de utilizator because "user name" alredy exists and is very confusing to a new user of the app

Quoting @alexandregz from the same  https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498121422: _"Romanian does not have an equivalent to a user handle specifically, so you'd have to use the Romanian equivalent of 'user name' instead, "Nume de utilizator". The literal meaning of "mâner" is the handle of a door, or the part of an object that you can grab and maneuver, and it certainly can't be understood in the context of a social media app. Don't use "identificator" (identifier", because it's different from the username."_ 

But, of course, we can discuss on this. However, my background of software developer, tells me that "user name" (often used as _username_, in a single word) refers what here is called _handle_, i.e., a combination of letters and digits, uniquely identifying a user. I the same time "identifier" of mostly used for an internal ID, such as a serial integer or an [universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier). User name is something that the user is choosing to represent them in that virtual space while the identifier is a value assigned by the system.

> Im used to translations being as faithful to the original as posible, always refer to the user with Dvs.,

The original (English) language has only one form of addressing. Romanian (and not only) has two forms: one formal (dumneavoastră, vă rog, etc) and the other informal/colloquial (tu, te rog, etc). Trying to be "faithful to the original" is not possible, you'll have to make a decision and pick one form or the other. Both are correct in Romanian language but the context is important. When addressing an official you'll use the formal form but with a friend you want to use the colloquial form. As this is a social app, I seemed to me more appropriate to use the informal way. We've discussed this in https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498121422, https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498445778 and https://github.com/bluesky-social/social-app/pull/6456#issuecomment-2498593586 and decided to continue with the informal form as make the app more familiar and friendly.

> For example, "# zi" is more correct as # is replaced by a number "1 zi", using "o zi" technically breaks and does not work as it can't show the number 1 because there is nothing to replace.

I'm a software developer and I know how this works. The "#" placeholder is replaced _only if exist_. We can discuss whether "1 zi" sounds better than "o zi" but the technical aspect should not be part of the discussion. Both forms are working correctly, w/o bugs (the current translation which is deployed shows it)

---

I do agree that the translation can and should be improved, but we need to decide on the process. Here's a proposal:

- Start a project on Crowdin by importing the _current_ `.po` file (existing translation)
- Maintain an official list of terms (glossary) in the Crowdin project to help with future translations. A start could be the list from https://github.com/bluesky-social/social-app/pull/6456#issue-2665899849
- Document in the Crowdin project the addressing form we're using (formal or informal). Any other guidelines should be added to documentation, e.g., using the correct Romanian diacritics (not Turkish)
- Start by fixing specific cases in the current translation, making the whole process transparent and subject to review by the community

PR #7367, opened 2025-01-05
Make post timestamp clickable

### LINKED ISSUE #6232: I think clicking datetime in a fullscreen post should open the post's url
### Describe the Feature

Right now, I can't click the datetime on full-screen posts like in other social media platforms. I find this useful when I want to open the post in a new tab for later.

### Attachments

post_date_full_screen

### Describe Alternatives

_No response_

### Additional Context

_No response_

c960657:
Make the timestamp in post thread view a clickable link, like on Facebook, Instagram, X, Mastdon, etc. 

Fixes #6232.

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -5,6 +5,7 @@ import {
   Text as RNText,
   View,
 } from 'react-native'
+import * as Clipboard from 'expo-clipboard'
 import {
   AppBskyFeedDefs,
   AppBskyFeedPost,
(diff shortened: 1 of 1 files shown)


pyxfluff:
Personally I think it should just copy the URL Instead of opening it in a new tab, much ore useful.

c960657:
It doesn't open a new tab.

pyxfluff:
My point for copying the link still stands, going to the menu is inconvenient

c960657:
Hmm, wouldn't it be confusing/non-standard if clicking a text that looks like a link writes to the clipboard?

auroursa:
> Hmm, wouldn't it be confusing/non-standard if clicking a text that looks like a link writes to the clipboard?

I think that sounds great! It might be even better to explicitly notify the user that the link has been copied to the clipboard, similar to when selecting "Copy Link" from the context menu.

pyxfluff:
:100: to the above reply. i want to be able to click it and have a "Link copied to clipboard!" alrert and move on without having to dig through submenus.

c960657:
Ok, let's do it :-)

PR #7378, opened 2025-01-07
Add unfollow prompt to avoid accidental unfollow

KevSlashNull:
This adds a prompt to ask before unfollowing accounts to avoid accidentally unfollowing someone.

This is useful because unfollowing is a reversible but destructive operation, so accidentally following someone can easily be reversed. But not vice versa because the account will be gone from your Following feed and your account's follows.

## Screenshots

image

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderStandard.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderStandard.tsx
@@ -69,6 +69,7 @@ let ProfileHeaderStandard = ({
   )
   const [_queueBlock, queueUnblock] = useProfileBlockMutationQueue(profile)
   const unblockPromptControl = Prompt.usePromptControl()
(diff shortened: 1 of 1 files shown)


KevSlashNull:
> If you confirm to unfollow someone, and then press "unfollow" again within a few minutes, I don't think it should show the dialog - otherwise it would be really annoying when clearing out your following list.

@oracularhades, thanks and i agree. though, and i should have clarified it in the description, this PR only touches the profile header. so, unless i'm missing something, you can still bulk unfollow from any account lists without ever seeing this prompt.

i would be careful to do this after the first confirmation because the user might misclick because the **Unfollow** confirmation button in the modal is close to the **Following** button. so perhaps after two confirms in the last minute?

i'll leave that to a follow-up and update the description here!

PR #7381, opened 2025-01-07
Fix double text input in composer

mozzius:
# Stacked on #7380 

We've got an issue with the paste input - there's two paste inputs being created, and for some reason the other one is now selectable/editable.

We can fix this by patching the library to remove the default one created by the React Native class we're inheriting from, so the real one created by `react-native-paste-input` is the only one used

https://github.com/user-attachments/assets/5b89cdd5-eac6-47b9-9387-8923b866b2cc

# Test plan

Try and repro the bug - involves clicking around at the very start of the composer text. Should be fixed 🤞
Check the input still works as it did
Check pasting still works

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@haileyok/bluesky-video": "0.2.6",
     "@ipld/dag-cbor": "^9.2.0",
     "@lingui/react": "^4.14.1",
-    "@mattermost/react-native-paste-input": "^0.7.1",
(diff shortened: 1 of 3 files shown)


haileyok (approved):
seems fine, left a note in slack with a question though. we should report this upstream as well just to put it on radar for a full fix (the maintainers of that repo have been pretty responsive previously)

PR #7383, opened 2025-01-07
Add bookmarks natively to Bluesky

ebwinters:
This PR adds native bookmarking to Bluesky. I understand it is likely this PR will implement Bluesky's bookmarking feature, so here are the:
### Groundwork for eventual bsky-team supported bookmarks implementation
- Get community feedback on initial bookmarking design
- Write code that can be used as a starting point for feature development
### Personal interest
- Introduce a [3rd party lexicon](https://github.com/lexicon-community/lexicon/blob/main/community/lexicon/bookmarks/bookmark.json) inside of Bluesky
- Bring a long-awaited feature to Bluesky that I’m also excited about

## Non-UI design choices

Implementation was done using `community.lexicon.bookmarks` for the following reasons:
1. Flexibility: Bluesky `at://` URIs are a subset of the `uri` type, meaning developers can more easily integrate bookmarks from Bluesky into other apps
2. Avoid duplication: Simply put, a bookmarking lexicon [already exists](https://github.com/lexicon-community/lexicon/pull/12) along with a [types package](https://www.npmjs.com/package/@lexicon-community/types), both of which have gone through a community design review. That lexicon will also be integrated into more 3rd party clients to come, so it makes sense to reuse it or extend it rather than make a new bookmarks schema
3. Open source collaboration: Adopting a third-party lexicon in Bluesky aligns with its goals of true decentralization and interoperability within the larger Atmosphere

### Public bookmarking
There is currently no way of making a private lexicon on `atproto`. This is generally a limitation within the protocol and I don't believe it should block development of bookmarks. There are workarounds for privacy, such as creating an anonymous account and bookmarking to that account, as well as extensions and labelers. 

### Putting bookmarks in the side bar
INSERT PICTURE HERE
I put bookmarks in the side bar for easier access. Another placement option is inside of the user profile, next to `your likes`.

### Hydrating the equivalent of an `isBookmarked` on every post
Showing a user if they've already bookmarked a post through icon highlighting is important so they don't make duplicates. However, since bookmarks are not build in to the response of `getPost()`, a post has no notion of a bookmark. To compensate, I fetch **all bookmarks by the current user**, cache them in memory, and then do a `bookmarks.contains(post.uri)` for each post rendered to the page. When a bookmark is removed/added, the cache is updated.
I can see this becoming a problem if a user has thousands of bookmarks, and am curious about other designs, tradeoffs, or approaches here. Some possibilities I can think of are
1. Building bookmarks into `PostView` at the protocol level, simiar to likes
2. Only caching bookmarks that have been created recently, which could fail when a user pulls up a post they've bookmarked from a while ago
3. Not highlighting bookmarks on posts in the feed view, and allowing duplicates

### Pagination of bookmarks
On the "my bookmarks" page, rather than fetching the entire list of user bookmarks, they should be paginated and shown on scroll. However, I was not able to figure out pagination, since the `getActorBookmarks` method does not exist on the Bluesky PDS. For this feature to move forward, this is a blocker that I'd need some help on.
    
### Filtering and searching bookmarks
The `community.lexicon.bookmarks.bookmark` allows for tagging bookmarks. This brings up more complicated UI designs that might not be worth fleshing out right now. However, in the future I can some popup when a bookmark is created for adding tags, and then searching/filtering on the "My Bookmarks" page.

## Screen shots and other links
Bookmarking on a feed:
400950259-00625ecd-3a71-48da-bf05-65a2cdffeace
Bookmarks page
400950248-8d481435-25c0-461f-96fc-19de7c6ef68d

[Discord thread with gif
](https://discord.com/channels/1097580399187738645/1326213989553803357)

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,7 @@
     "js-sha256": "^0.9.0",
     "jwt-decode": "^4.0.0",
     "lande": "^1.0.10",
+    "@lexicon-community/types": "^1.0.0",
     "lodash.chunk": "^4.2.0",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -21,6 +21,7 @@ import {useWebScrollRestoration} from '#/lib/hooks/useWebScrollRestoration'
 import {buildStateObject} from '#/lib/routes/helpers'
 import {
   AllNavigatorParams,
+  BookmarksTabNavigatorParams,
   BottomTabNavigatorParams,
   FlatNavigatorParams,
   HomeTabNavigatorParams,
@@ -40,6 +41,7 @@ import {
   shouldRequestEmailConfirmation,
   snoozeEmailConfirmationPrompt,
 } from '#/state/shell/reminders'
+import {BookmarksScreen} from '#/view/screens/Bookmarks'
 import {CommunityGuidelinesScreen} from '#/view/screens/CommunityGuidelines'
 import {CopyrightPolicyScreen} from '#/view/screens/CopyrightPolicy'
 import {DebugModScreen} from '#/view/screens/DebugMod'
@@ -108,6 +110,8 @@ const HomeTab = createNativeStackNavigatorWithAuth<HomeTabNavigatorParams>()
 const SearchTab = createNativeStackNavigatorWithAuth<SearchTabNavigatorParams>()
 const NotificationsTab =
   createNativeStackNavigatorWithAuth<NotificationsTabNavigatorParams>()
+const BookmarksTab =
+  createNativeStackNavigatorWithAuth<BookmarksTabNavigatorParams>()
 const MyProfileTab =
   createNativeStackNavigatorWithAuth<MyProfileTabNavigatorParams>()
 const MessagesTab =
@@ -128,6 +132,11 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
         name="NotFound"
         getComponent={() => NotFoundScreen}
         options={{title: title(msg`Not Found`)}}
+      />
+      <Stack.Screen
+        name="Bookmarks"
+        getComponent={() => BookmarksScreen}
+        options={{title: title(msg`Bookmarks`), requireAuth: true}}
       />
       <Stack.Screen
         name="Lists"
@@ -451,6 +460,10 @@ function TabsNavigator() {
         getComponent={() => NotificationsTabNavigator}
       />
       <Tab.Screen
+        name="BookmarksTab"
+        getComponent={() => BookmarksTabNavigator}
+      />
+      <Tab.Screen
         name="MyProfileTab"
         getComponent={() => MyProfileTabNavigator}
       />
@@ -516,6 +529,27 @@ function NotificationsTabNavigator() {
       />
       {commonScreens(NotificationsTab as typeof HomeTab)}
     </NotificationsTab.Navigator>
+  )
+}
+
+function BookmarksTabNavigator() {
+  const t = useTheme()
+  return (
+    <BookmarksTab.Navigator
+      screenOptions={{
+        animationDuration: 285,
+        gestureEnabled: true,
+        fullScreenGestureEnabled: true,
+        headerShown: false,
+        contentStyle: t.atoms.bg,
+      }}>
+      <BookmarksTab.Screen
+        name="Bookmarks"
+        getComponent={() => BookmarksScreen}
+        options={{requireAuth: true}}
+      />
+      {commonScreens(BookmarksTab as typeof HomeTab)}
+    </BookmarksTab.Navigator>
   )
 }
 
--- a/src/components/icons/Bookmark.tsx
+++ b/src/components/icons/Bookmark.tsx
@@ -1,0 +1,9 @@
+import {createSinglePathSVG} from './TEMPLATE'
+
+export const Bookmark_Stroke2_Corner0_Rounded = createSinglePathSVG({
+  path: 'M19.5 22.5c-.2 0-.3 0-.5-.1l-6.5-3.7L6 22.4c-.3.2-.7.2-1 0-.3-.2-.5-.5-.5-.9v-18c0-.6.4-1 1-1h14c.6 0 1 .4 1 1v18c0 .4-.2.7-.5.9-.2.1-.3.1-.5.1zm-7-6c.2 0 .3 0 .5.1l5.5 3.1V4.5h-12v15.3l5.5-3.1c.2-.2.3-.2.5-.2z',
+})
+
+export const Bookmark_Filled_Corner0_Rounded = createSinglePathSVG({
+  path: 'M19 22c-.2 0-.3 0-.5-.1L12 18.2l-6.5 3.7c-.3.2-.7.2-1 0-.3-.2-.5-.5-.5-.9V3c0-.6.4-1 1-1h14c.6 0 1 .4 1 1v18c0 .4-.2.7-.5.9-.2.1-.3.1-.5.1z',
+})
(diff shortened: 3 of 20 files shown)


pyxfluff:
Public everything isn't great and I wish my likes were private but thats more of a protocol issue. Bookmarks being public seems like a poor idea but unfortunately maybe it makes sense. Also, move the bookmark button to the far right with only a little big of padding with the menu button.

davidcelis:
I agree that it doesn't make sense for bookmarks to be public and there should be no need for that in the lexicon if it can be avoided

erlend-sh:
Count me in for ‘okay with public bookmarks’. Yeah it’s not ideal, but it’s simply the limitation of the platform as it exists right now, and it’s a great start for client convergence.

The notion of all-public-activity when using Bluesky is already deeply engrained for me, so my use of bookmarks won’t be any different.

Also, here’s a workaround that multi-account clients could easily accommodate:
1. make an anonymous alt-account
2. anything you want to privately bookmark, you bookmark with that account

This shouldn’t have to require constant account switching, as there’s already prior art (in Mastodon at least) for clients that have a UI utility like “repost/favorite from: acc1, acc2..”

ebwinters:
@oracularhades my problem with private bookmarks, rather than using an anonymous account, is that it takes away portability. The data for bookmarks would be stored on a private server, which seems wrong at a protocol level, rather than chats being stored on a private server. If anything, it made sense for blocks to be stored privately. Blocking someone can have real-world implications, and its better for those who are blocked to not know.

As for scraping, I don't see a reason why bookmarks are any different than likes or reposts. I could make the argument that reposts are actually what I'm truly interested in and support, and bookmarks are just a way of saving things like recipes to come back to later. That's just the nature of Bluesky, all the data is public.

ebwinters:
@surfdude29 I wouldn't be happy with public bookmarks per se, that's just the limits of the protocol _right now_. I'd rather bookmarks be stored in the PDS for portability, but if private state on the PDS is a thing, that is the ideal solution, and I'd imagine a lexicon/client change would be in store for the `community.lexicon.bookmarks` schema.

surfdude29:
@ebwinters Based on all the feedback I've seen from users on Bluesky, my view is that a fundamental requirement of any implementation of bookmarks on Bluesky is that they be kept private.

Therefore, I would suggest that we need to wait until this can be done in a way that's consistent with other goals – private state on the PDS seems a promising way to do that – instead of going ahead with public bookmarks as an interim solution, simply because that's all that can be managed at the moment.

I think doing this would be confusing for Bluesky users ("we're releasing a half-finished version of bookmarks, they're all public, but we hope to make them private at some point in the future") and would not meet their expectations.

erlend-sh:
What if the (public) bookmarks feature was behind a Labs 🧪 type setting, which also clearly pointed out the public nature of the bookmarks?

Wouldn’t that solve for the divergence in public/private expectations, while also satisfying all the people who want a better alternative than the noisy 📌 hack?

surfdude29:
> What if the (public) bookmarks feature was behind a Labs 🧪 type setting, which also clearly pointed out the public nature of the bookmarks?
> 
> Wouldn’t that solve for the divergence in public/private expectations, while also satisfying all the people who want a better alternative than the noisy 📌 hack?

That could work, and fwiw I would use it myself. It would solve the key concern I have of user confusion, which I believe would result from enabling public bookmarks by default. Users would have to opt in to use them and it would be made clear to them when they do so that bookmarks are public information, the same as likes are.

Of course, we'll have to wait for someone from the team to weigh in with what their thinking is as that's what will determine what happens with this, but I imagine they would consider issues like:

- Would releasing public bookmarks now as a Labs feature be useful to enough users?
- Is there likely to be a simple migration path to private bookmarks when those are released? Not just taking into account the Bluesky app, but third-party apps and services as well?
- Even taking into account that users are warned when they enable bookmarks that they're public information, are there Trust & Safety concerns over users being harassed, targeted or profiled based on posts that they've bookmarked? (I might be being naive but this point doesn't seem too likely to me, as I'm not aware of any significant cases of this happening at the moment with likes, despite those being public at the API level. However, for a comprehensive assessment I think it's worth mentioning as a possibility to consider.)

ebwinters:
Those are all good points. Well thought out

I personally agree that putting bookmarks behind an opt-in experience would be useful, and I'd think enough people want to bookmark recipes or articles which they wouldn't mind being public. As @surfdude29 called out, having a migration path to private state is crucial here. I can't think of how that would work technically, but I think there will be some solution developed since Bluesky wants types like mutes to live on the PDS. At worst there might be some manual migration tool

davidcelis:
If the Bluesky team themselves want mutes to live on the PDS, why not start with bookmarks implemented similarly to mutes as opposed to starting with them implemented publicly? Presumably they'll introduce some way to migrate mutes into the PDS while keeping them private if that's their goal, and we could do the same with bookmarks. If either scenario requires an eventual migration, it feels far more useful to me to start with a feature that behaves in the way people expect.

ebwinters:
@davidcelis correct me if I'm wrong but I don't think their app view code is public. There is a deprecated `bsky` package but it is not in use in prod

surfdude29:
> @davidcelis correct me if I'm wrong but I don't think their app view code is public. There is a deprecated `bsky` package but it is not in use in prod

@ebwinters This is the `bsky` (Bluesky AppView Service) package in the atproto repo, where have you seen that it's deprecated? I'm almost certain this code is running in prod:

https://github.com/bluesky-social/atproto/tree/main/packages/bsky

ebwinters:
It was mentioned on discord, so I'm not 100%

Invertex:
> * Even taking into account that users are warned when they enable bookmarks that they're public information, are there Trust & Safety concerns over users being harassed, targeted or profiled based on posts that they've bookmarked? (I might be being naive but this point doesn't seem too likely to me, as I'm not aware of any significant cases of this happening at the moment with likes, despite those being public at the API level. However, for a comprehensive assessment I think it's worth mentioning as a possibility to consider.)

To touch on this further, there actually is a bit of an issue there with likes. While it's not so much direct harassment (though I have seen it go on over snooped likes), it's people using data about likes to generate block-lists en-masse which others hop onto, trusting a singular user's judgement and resulting in greater potential platform segregation than realistically needs to be. It's not uncommon for the conditions to be added to such a list being "you liked one of this person's posts one time, so we're assuming you endorse that person" even though it's likely you don't know anything about them. This is pretty prevalent when it comes to the art space on BSKY as well where it becomes pretty messy with how varied the opinions are on things.

[Clearsky.app](https://clearsky.app/) has give people a pretty active awareness of this issue.

That's not to say it doesn't have upsides of course, I certainly enjoy being able to mass block bots that a list is detecting, or more clear examples of people I will not jive with. But the potential of information like this being misused to an egregious extent based on the whims of people with a decent following shouldn't be understated and there's a whole discussion to be had there about this that I'm trying not to get too into here haha.

So to circle back to "public" Bookmarks, it kind of worsens the potential bad behaviors of these systems, as people may be prone to a default assumption that if you bookmarked something, you endorse it/them similar to a "like", instead of the many reasons you might have for doing so that they don't know about.

I don't really see the purpose of a public Bookmarks system *alongside* a private one. I think it would be more accurate to call a public one "Favorites" instead of "Bookmarks". This would also help reduce user confusion and more clearly signal intent for that public information.
----------------------
It was also brought up to have public/curated sets of Bookmarks, but I would counter that as also unnecessary since we already have a system for this, which is "Lists". Whether you've bookmarked something or not would make no difference to adding posts to a list of "curated posts I want to display". So that discussion seems like it would be better suited to improving UX around making lists and displaying them, no?

ebwinters:
@Invertex I still think having a list of bookmarks/items that is specifically just for you in the UI (hidden to others in the UX but accessible on the protocol) is useful. The whole thing you mentioned about people abusing likes seems like a tradeoff when working with AT Protocol - the less data that is publicly accessible, the harder it is to move between apps. The more data that's accessible, the more people can abuse that data. Maybe there should be tooling around solving those problems instead of pushing more things into private state, which goes against the nature of the protocol IMO. I think you are in the majority opinion though, and bookmarks will come whenever private PDS state is delivered

mozzius:
Superceded by #8976. Thank you for you contribution!

PR #7388, opened 2025-01-07
Fix avatar shape in Account switch menu

mozzius:
Prop was missing in a couple of places, so labellers were showing up as circles

Screenshot 2025-01-07 at 22 12 31


gaearon (approved):

mozzius:
@gaearon made it required and turned out to be a few other places that needed it. I made sure to explicitly use `user` for the pills, since those need to be circles

PR #7392, opened 2025-01-08
Update Korean localization for v1.97.0

quiple:
Translates updated strings into Korean.


pfrazee (approved):
Thanks as always quiple!

PR #7405, opened 2025-01-08
update and re-enable the bitdrift SDK

snowp:
This reverts the PR to remove bitdrift and updates it to the latest version 0.6.1

Notable changes:
- The session management strategy has been exposed in the API. We configure it as activity-based which means that sessions will span multiple process lifetimes if they happen close together, and time out if the app is inactive for too long.
- `addField` has been added, which allows adding global fields that will be present on all logs going forward
- Network integration has been enabled for Android, which should automatic logs for every HTTP call made by the app. The iOS integration does not currently work with Bluesky due to some timing issues related to the bitdrift SDK being behind a feature flag, so this is disabled until we can implement a known fix.

### DIFF
--- a/src/lib/bitdrift.ts
+++ b/src/lib/bitdrift.ts
@@ -1,27 +1,23 @@
-// import {init} from '@bitdrift/react-native'
-// import {Statsig} from 'statsig-react-native-expo'
-// export {debug, error, info, warn} from '@bitdrift/react-native'
+import {init, SessionStrategy} from '@bitdrift/react-native'
+import {Statsig} from 'statsig-react-native-expo'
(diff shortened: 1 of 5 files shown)


src/lib/bitdrift.ts
```diff
@@ -1,27 +1,24 @@
...
-//   }
-// })
-
-// TODO: Reenable when the build issue is fixed.
-export function debug(_message: string) {}
-export function error(_message: string) {}
-export function info(_message: string) {}
-export function warn(_message: string) {}
+initPromise.then(() => {
+  let isEnabled = true
```
gaearon, line 10: This is hardcoded to true which is probably not intended?

gaearon (approved):
lets try it

PR #7409, opened 2025-01-09
Update `locales/uk/messages.po`

DanikVitek:
Better translation in this context


surfdude29:
Thanks for this! You might want to suggest this change at the [Crowdin project for the Ukrainian localization](https://crowdin.com/project/bsky-app-ua).

cc: @imax9000

PR #7410, opened 2025-01-09
Downgrade paste input dependency

mozzius:
v0.8.1 causes crashes on Android. To be figured out in future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@haileyok/bluesky-video": "0.2.6",
     "@ipld/dag-cbor": "^9.2.0",
     "@lingui/react": "^4.14.1",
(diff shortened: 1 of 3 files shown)


gaearon (approved):

WoLewicki:
@mozzius could you elaborate on what crash it causes? I have kind of close version to `0.8.1` in my new arch PR: https://github.com/bluesky-social/social-app/pull/6755 so not sure if it would happen there too?

gaearon:
Specifically when opening the composer

mozzius:
Ah I found the error

image

I haven't looked into it particularly deeply yet, I was working on something else and didn't actually need the upgrade for that

WoLewicki:
Ok so it will be nice to keep track if the same does not happen on new arch.

PR #7422, opened 2025-01-10
[ELI5] Validate too long handles in signup

mozzius:
Checks if a handle is too long in the signup flow. Handles can be up to 253 chars normally, but for PDS service handles the max is 30, including domain. I added an additional flag to the `validateHandle` function to lower the length check appropriately. I made it generic for other PDSes, rather than hard-coding 18 chars. I also corrected the logic in the change handle screen since that was hardcoded to 18.

Screenshot 2025-01-10 at 13 49 13

# Test plan

1. Confirm that the limit really is 30 chars - go through the signup process on main with a 19 char username and confirm it fails
2. Check that the PR change catches it, and check that the error message is correct

### DIFF
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -45,7 +45,7 @@ export interface IsValidHandle {
 export function validateHandle(
   str: string,
   userDomain: string,
-  serviceHandle?: boolean,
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
Yep nice one!

PR #7423, opened 2025-01-10
Conversation-level reporting by reporting most recent message available

mozzius:
Allows conversation-level reporting by reporting the most recent message available. In the chat list screen, this is the latest message that we get for the little preview, and in the chat screen itself we use the most recent message from the other person. Comes with a little note indicating that only the last few messages in the convo will be reported.

If it can't find a message to report, it falls back to the old prompt - for example if you're reporting from the chat list screen and you sent the last message

https://github.com/user-attachments/assets/eefa6cbf-94d1-48f1-91ab-e63aa230430e

# Test plan

Report different conversations from different menus. It should always only report the most recent message from the other person in the convo.

### DIFF
--- a/src/components/ReportDialog/SelectReportOptionView.tsx
+++ b/src/components/ReportDialog/SelectReportOptionView.tsx
@@ -23,15 +23,18 @@ import {
 } from '#/components/icons/Chevron'
 import {SquareArrowTopRight_Stroke2_Corner0_Rounded as SquareArrowTopRight} from '#/components/icons/SquareArrowTopRight'
 import {Text} from '#/components/Typography'
+import {Admonition} from '../Admonition'
 import {ReportDialogProps} from './types'
 
 export function SelectReportOptionView({
+  showConversationWarning,
   ...props
 }: {
   params: ReportDialogProps['params']
   labelers: AppBskyLabelerDefs.LabelerViewDetailed[]
   onSelectReportOption: (reportOption: ReportOption) => void
   goBack: () => void
+  showConversationWarning?: boolean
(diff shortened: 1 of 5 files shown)


src/components/ReportDialog/SelectReportOptionView.tsx
```diff
@@ -88,6 +91,14 @@ export function SelectReportOptionView({
         <Text style={[a.text_md, t.atoms.text_contrast_medium]}>
           {i18n.description}
         </Text>
+        {showConversationWarning && (
+          <Admonition type="info" style={a.mt_xs}>
+            <Trans>
+              Note: Only the most recent messages in this conversation will be
```
gaearon, line 97: "messages" or "message"?

gaearon: tbh i'm not sure if this notice is necessary. you're going to see the message being reported on the next screen anyway at which point you can infer this. i'm ok with this not being present.

gaearon (approved):

src/components/ReportDialog/SelectReportOptionView.tsx
mozzius: Mods can see a certain window of messages iirc. So despite reporting the conversation it is actually reporting only the most recent messages - thought we might want to specify, but on the other hand if you think it's fine then I'd be happy to leave it unmentioned

gaearon: let’s just drop the notice for simplicity here

PR #7425, opened 2025-01-10
Fix invalid password error message

mozzius:
1. Catches backend errors properly - seems this changed at some point
2. Better errors for empty username/password
3. Tweaks wording, and change .po files to avoid translation churn

Screenshot 2025-01-10 at 19 15 06


gaearon (approved):
lgtm

PR #7435, opened 2025-01-12
feat #1547 now threads, posts are collapsable

### LINKED ISSUE #1547: Thread Collapsing
A thread collapsing feature could be implemented to allow users to easily navigate through comment threads by collapsing and expanding them as needed. This feature should include:

1. A mechanism for collapsing and expanding threads, such as:
   - Clicking on the vertical line to the side of the comments, representing the thread hierarchy.
   - Clicking on a button, like an arrow icon, that toggles between an arrow pointing to the side (collapsed) and an arrow pointing down (expanded).
   - Maybe long tap for mobile.
2. A user setting for default thread collapsing behavior, allowing users to set a nesting level threshold. This would enable users to see only the least nested comments by default while still having the option to expand threads as needed.

X1Vi:
https://github.com/user-attachments/assets/61d6ef84-6ed1-4035-8590-f1d93dd4214c

I have made threads collapsible 
This closes #1547

### DIFF
--- a/src/view/com/post-thread/PostThreadItem.tsx
+++ b/src/view/com/post-thread/PostThreadItem.tsx
@@ -1,6 +1,7 @@
-import React, {memo, useMemo} from 'react'
+import React, {memo, useMemo, useState} from 'react'
 import {
   GestureResponderEvent,
+  Pressable,
   StyleSheet,
   Text as RNText,
   View,
@@ -268,6 +269,11 @@ let PostThreadItemLoaded = ({
       onPost: onPostReply,
     })
   }, [openComposer, post, record, onPostReply, moderation])
+  const [isCollapsed, setIsCollapsed] = useState(false)
+
+  const toggleCollapse = () => {
+    setIsCollapsed(!isCollapsed)
+  }
 
   const onPressShowMore = React.useCallback(() => {
     setLimitLines(false)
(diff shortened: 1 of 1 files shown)


X1Vi:
@gaearon can you please elaborate more as to what kind of icons do you want specifically (how should the icons look with a reference to a library that is being already being used in the bluesky codebase) ? And where do you want them ? A reference would be great, a figma would be awesome. Because I think logic is already there that we can use however we want it.

X1Vi:
I tried to make it like this https://github.com/bluesky-social/social-app/issues/1547#issuecomment-2506734375

X1Vi:
How bout show more and show less button attached to every item ?

gursuj:
> @gaearon can you please elaborate more as to what kind of icons do you want specifically (how should the icons look with a reference to a library that is being already being used in the bluesky codebase) ? And where do you want them ? A reference would be great. Because I think logic is already there that we can use however we want it.

Maybe icons for -/+ could be placed along the vertical bar that indicates the thread link, like how reddit does it?
image
This is how it looks when collapsed:
image
Noticed that if a collapsed comment has a parent, the vertical bar seems to connect to the plus icon
image

For extra reference, here's how old reddit shows collapse icons by default. Though, I think the above examples would be more suitable for Bluesky's UI than this.
image

gursuj:
I think the behavior for clicking on a comment should be reserved to opening that comment, i.e. how it works currently vs. how this PR makes it collapse? Usecases would be to view its replies easily, make it larger for sharing etc. and have be collapsing done when clicking the vertical bar / icons only. There have been recent changes to the vertical bar appearance. Might try making a PR for this idea.

PR #7448, opened 2025-01-14
[Android] Try to disambiguate taps from swipes

gaearon:
Hackfix for https://github.com/callstack/react-native-pager-view/issues/960.

Alternative to https://github.com/bluesky-social/social-app/pull/7312 but doesn't require copy paste and also seems to work more reliably.

Comments explain how it works.

Will need some testing.


mozzius (approved):
I'm fairly certain this is a solid fix. Tested on low-end android and it seems to eliminate the majority of rogue presses

PR #7449, opened 2025-01-14
Set webpack public path to auto to load from CDN (hopefully)

ericvolp12:
Webpack produces chunk and static files with references that aren't pointing to where the browser loaded the script from. When we serve the main script from a static asset CDN, the references to chunkfiles don't share the path of the CDN.

To fully leverage a static asset CDN, we should fix that hopefully with this patch.


ericvolp12:
Confirmed this does what we want and chunkfiles are being loaded from the CDN in staging woohoo!

gaearon (approved):
seems fine

Issue #7260, opened 2024-12-23
Bluesky can't detect a post's language reliably enough


ssg:
### Steps to Reproduce

1. Draft a new post while non-English is selected: type this poem by William Wordsworth for example, but it works with many other texts. "never did sun more beautifully steep, in its first splendour rock or valley or hill. ne'er saw i, never felt a calm so deep. the river glideth in its own sweet will. dear god! many houses seem asleep and all that mighty heart is lying still!"
3. No notification about "Are you writing this in English?"

### What platform(s) does this occur on?

iOS, Web (Desktop)

### What version of the app are you using?

1.96.2

### Additional Information

I understand that this is because "lande", the package used for language detection, is quite unreliable. But it's obviously insufficient for Bluesky. This is significant because picking the wrong language while posting means loss of most of the reach along with loss of followers (due to the noise created for people who don't speak the language). Maybe the whole multilingual UX needs to be worked on, I'm not sure.

SpriterGors:
I mentioned this issue here https://github.com/bluesky-social/social-app/issues/5764, and also is mentioned in https://github.com/bluesky-social/social-app/issues/6920 and https://github.com/bluesky-social/social-app/issues/5885

If the automated detector doesn't kick in reliably, at least we should have a more robust means to select the language before we hit "post"

KernelLeak:
Most of the time this happens to me, I'm actually replying to a post in German while language is set to English, or vice versa (and the detector not kicking in) - a very easy solution to fix most of these cases would be to default to the language of the post that's being replied to when replying.

surfdude29:
> […] a very easy solution to fix most of these cases would be to default to the language of the post that's being replied to when replying.

I like this idea, although I have a couple of thoughts:
- A toast or some kind of pop-up should let the user know that the post language has been changed
- I think it should only be changed for that reply, and should then return to the previous setting

> Oh, and actually being able to see the language a post is tagged as anywhere would help noticing it after posting...

This would definitely be useful, perhaps the Translate button could be changed to show the language, e.g. instead of just **Translate** it could say **Translate from [language name]**.

Also, in case you're curious about a particular post, you can look at the record on PDSls to check what language it's tagged as:

https://pdsls.dev/

KernelLeak:
> * A toast or some kind of pop-up should let the user know that the post language has been changed

Agreed.

> * I think it should only be changed for that reply, and should then return to the previous setting

[...]

PR #7455, opened 2025-01-14
Improve language detection accuracy

ssg:
Problem: when you're typing a post, "are you writing in X language?" prompt appears and disappears occasionally. That frequently happens when you press space and start writing the next word too as the half-word is probably unknown to the model. When you finish the word, the prompt appears again, but this makes the prompt unnecessarily blink, and lose opportunity to notify the user. That can cause posts appear with incorrect language settings and can be quite harmful for its reach (as mentioned in #7260).

This change aims to improve the UX of "are you writing in X language?" by discarding the last word from the detection set until a more reliable detection library replaces lande.

### DIFF
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -33,7 +33,14 @@ export function SuggestedLanguage({text}: {text: string}) {
   const {_} = useLingui()
 
   useEffect(() => {
-    const textTrimmed = text.trim()
+    let textTrimmed = text.trim()
+
+    // remove the last word before guessing to prevent a half-written word 
+    // from botching the confidence of language detection.
+    const lastSpace = textTrimmed.lastIndexOf(' ')
+    if (lastSpace > 0) {
+      textTrimmed = textTrimmed.slice(0, lastSpace)
+    }
 
     // Don't run the language model on small posts, the results are likely
     // to be inaccurate anyway.
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -35,8 +35,16 @@ export function SuggestedLanguage({text}: {text: string}) {
   useEffect(() => {
     let textTrimmed = text.trim()
 
-    // remove the last word before guessing to prevent a half-written word 
+    // Remove the last word before guessing to prevent a half-written word 
     // from botching the confidence of language detection.
+    // There are two gotchas with this approach:
+    // First, it might increase the practical minimum length for the language
+    // detection because removing the last word would eat away from the 
+    // 40 character min limit. I think it's worth it though. 
+    // Second, this will also discard the last word that has been typed fully 
+    // which might affect the outcome. One might consider detecting punctuation 
+    // at the end of the last word to include it in the language detection, 
+    // but it's quite hard to do that for all languages correctly.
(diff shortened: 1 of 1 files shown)


gaearon:
Can you please describe how you tested it? Videos and screenshots help.

ssg:
Thanks for asking for tests, @gaearon, and let me know if we need to have a test suite with broader scope for this. I'm attaching comparison videos for pre-change and post-change. As you can see, pre-change not only does blink, but fails to detect language properly too. Post-change kicks in slightly later, but works perfectly, stays solid, doesn't get confused.

Here is a Turkish text I'm writing on the current Bluesky on the web:

https://github.com/user-attachments/assets/75305d8a-b9fe-422a-bd34-91291fa21cef

Here is me typing the same text on my local server with the latest code changes:

https://github.com/user-attachments/assets/56c27b22-55d2-45a5-9d5d-46e8af1dd613

I also raised language detection threshold from `0.0002` to `0.02` because I've seen this a lot in detection output:

image

Older threshold was "0.0002", as you can see very low for the algoritm not to be confused. I suspect that someone might have confused 0.02 with 0.0002 due to mix up with decimal points. And because it can easily match multiple languages at that confidence level, the language detection would bail out, and turn off the language suggestion prompt causing either blinks or disappearance of the prompt completely.

I think a false positive is even better than not matching at all due to socially hard to recover problematic nature of posting in the wrong language (losing reach and followers). Lack of edit and invisibility of the post language in the UI make this even harder to tackle.

I only tested this manually using English and Turkish languages, and I understand certain languages have different detection characteristics and can behave erratically, but currently English and Turkish work much much better than how it was before. Let me know of your thoughts. Thanks!

ssg:
My greatest concern is obviously not being able to test this with a larger corpus. So, I don't know if I'm introducing a regression for all other languages while improving Turkish and English. A shot in the dark basically. Instinctively, I don't think I do, but I've been wrong before. :) 

But, this feature doesn't have any tests at the moment as I understand, so, maybe it would be safe to experiment on it. And, as I said, a false positive is much better than no prompt at all.

We can work on a test suite for reliable language detection, but it would require orders of magnitude more time, and effort and might stall any improvements to this feature indefinitely.

ssg:
Another example, the current language detection (fails to detect Turkish):

https://github.com/user-attachments/assets/ed476a9c-42b6-43f4-9bad-179bc9c22fc7

This is after my changes:

https://github.com/user-attachments/assets/de7f52eb-c142-45ca-82d7-087a7372eb8d

I'll post English examples too.

ssg:
An example of an English post. Before my changes:

https://github.com/user-attachments/assets/32e41812-354a-4a57-902d-c552b1b2f393

And after my changes:

https://github.com/user-attachments/assets/702181b0-599f-4fe2-be9d-5949e52eba73

ssg:
Another example of English before changes. Throughout all the text, English never seems to be detected:

https://github.com/user-attachments/assets/d875e103-c7bc-484e-b04e-32068e7fb519

And this is after changes. Still detected late, because model has less confidence in it (56% at the first two lines), but it gets detected eventually, and stays like that after a certain point:

https://github.com/user-attachments/assets/8de4790e-3de4-43bc-bc41-3d76feac8e4f

ssg:
Unfortunately, my attempt to fix the problem in https://github.com/bluesky-social/social-app/issues/8113#issuecomment-2777101878 destroyed this PR. I'll create a new one 😩

PR #7459, opened 2025-01-15
[Android] Fix taps triggering while swiping

gpp-0:
A more proper, native-level fix for the issue described in https://github.com/bluesky-social/social-app/pull/7312.
The root of the issue was https://github.com/callstack/react-native-pager-view/issues/954, ie. a cancel event wasn't dispatched to the pager's children when it started scrolling. The patch uses [notifyNativeGestureStarted](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/NativeGestureUtil.kt) to take care of that. It also disables the requestDisallowInterceptTouchEvent calls because they mess with the Drawer

### DIFF
--- a/patches/react-native-pager-view+6.6.1.patch
+++ b/patches/react-native-pager-view+6.6.1.patch
@@ -1,9 +1,9 @@
 diff --git a/node_modules/react-native-pager-view/android/build/tmp/kotlin-classes/debug/com/reactnativepagerview/NestedScrollableHost.class b/node_modules/react-native-pager-view/android/build/tmp/kotlin-classes/debug/com/reactnativepagerview/NestedScrollableHost.class
 new file mode 100644
-index 0000000..5129f0e
+index 0000000..b64fccc
 Binary files /dev/null and b/node_modules/react-native-pager-view/android/build/tmp/kotlin-classes/debug/com/reactnativepagerview/NestedScrollableHost.class differ
 diff --git a/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt b/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
-index 91d9946..294a3a8 100644
+index 91d9946..87b58d0 100644
 --- a/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
 +++ b/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
 @@ -8,6 +8,7 @@ import android.view.ViewConfiguration
@@ -14,7 +14,15 @@ index 91d9946..294a3a8 100644
  import kotlin.math.absoluteValue
  import kotlin.math.sign
  
-@@ -57,17 +58,14 @@ class NestedScrollableHost : FrameLayout {
+@@ -27,6 +28,7 @@ class NestedScrollableHost : FrameLayout {
+   private var touchSlop = 0
+   private var initialX = 0f
+   private var initialY = 0f
++  private var nativeGestureStarted: Boolean = false
+   private val parentViewPager: ViewPager2?
+     get() {
+       var v: View? = parent as? View
+@@ -57,17 +59,14 @@ class NestedScrollableHost : FrameLayout {
    }
  
    private fun handleInterceptTouchEvent(e: MotionEvent) {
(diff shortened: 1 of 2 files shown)


gaearon:
Oh nice! Mind reverting https://github.com/bluesky-social/social-app/pull/7448 and verifying it still works with your fix? Also, do you have an upstream PR to link to?

gaearon:
Note I filed https://github.com/callstack/react-native-pager-view/issues/960 with another minimal repro.

gpp-0:
Yep, I've tested https://github.com/callstack/react-native-pager-view/issues/960 and it works great, no stray taps! This branch already doesn't include https://github.com/bluesky-social/social-app/pull/7448, do you mean I should rebase and commit a revert of https://github.com/bluesky-social/social-app/pull/7448 here? Sorry I'm fairly new to this 😅
I'm not quite sure if I should make a PR upstream (at least not yet) because I haven't tested cases with multiple nested pagers that this class I modified is supposed to handle. Honestly I'm having trouble making sense of that class, it looks a bit dodgy and I'm not sure if it works as it's supposed to...

gaearon:
>I'm not quite sure if I should make a PR upstream (at least not yet) because I haven't tested cases with multiple nested pagers that this class I modified is supposed to handle.

Sometimes it's good to send an incomplete PR to start a discussion since the maintainers there started looking into the issue as well.

col3name:
> > I'm not quite sure if I should make a PR upstream (at least not yet) because I haven't tested cases with multiple nested pagers that this class I modified is supposed to handle.
> 
> Sometimes it's good to send an incomplete PR to start a discussion since the maintainers there started looking into the issue as well.



> > I'm not quite sure if I should make a PR upstream (at least not yet) because I haven't tested cases with multiple nested pagers that this class I modified is supposed to handle.
> 
> Sometimes it's good to send an incomplete PR to start a discussion since the maintainers there started looking into the issue as well.

Im find same behavior on ios 17.7.2, bsky 1.96.4.818


https://github.com/user-attachments/assets/5b44f51b-7026-4227-a0e1-8c2f9d3e5144

patches/react-native-pager-view+6.6.1.patch
```diff
@@ -0,0 +1,51 @@
...
+ import kotlin.math.absoluteValue
+ import kotlin.math.sign
+ 
+@@ -57,17 +59,12 @@ class NestedScrollableHost : FrameLayout {
+   }
+ 
+   private fun handleInterceptTouchEvent(e: MotionEvent) {
+-    val orientation = parentViewPager?.orientation ?: return
+-
+-    // Early return if child can't scroll in same direction as parent
```
gaearon, line 20: Can you tell me more about why this change is needed? You mentioned a drawer gesture conflict but I thought I already fixed that separately about a week ago. Is this change really necessary? If yes, which precise problem is it fixing?

gpp-0: The thing is that what is called the "parentViewPager" here is not the current pager A that is being considered, it's a potential parent pager B that wraps A. What is called "child" here is actually pager A. Originally, because of the early returns, this function only does anything when we have multiple nested pagers. With a single pager, like in the Bluesky app, parentViewPager will always be null so this used to be dead code in that case.

About the drawer gesture, this has to do with the parent.requestDisallowInterceptTouchEvent calls that I removed. Since I got rid of the early returns and this code is now executed, the parent view (different from parentViewPager) would not be allowed to intercept the right swipe gesture. Basically, it becomes impossible to open the drawer via swipe if those calls are included.

gaearon: If we _wanted_ to have multiple pagers, would this patch break them? Also, does this change mean that it’s unlikely for the code to get accepted upstream?

I think ideally we’d prefer something that’s likely to land there. 

If the problem is that the early return prevents the nested code from executing (which we now _do_ want to execute because there’s a line we  need there), can whatever the early returns were preventing be scoped more tightly? Presumably they’re preventing some concrete calls, maybe something inside should be wrapped with an if?

gpp-0: Yes, it turns out this completely breaks nested pagers, which is why I didn't end up making a PR upstream. I'm still looking into it, but I'll make a PR now at least suggesting the use of notifyNativeGestureStarted and we'll see how it goes

gpp-0:
Good news, I opened a slightly modified PR upstream which doesn't break nested pagers and doesn't break the drawer and overall should make more sense :)
https://github.com/callstack/react-native-pager-view/pull/961

gaearon:
Mind updating this patch to match your PR 1:1?

gaearon (approved):

gaearon:
OK, I'm also layering on https://github.com/bluesky-social/social-app/pull/7459/commits/88b187815d95e5e3f76e61af22e76e5715751236 for consistency with RN code though not sure it does anything for our case.

gaearon:
Reapplying this via upstream fix in https://github.com/bluesky-social/social-app/pull/7503

PR #7469, opened 2025-01-17
Do not serve cache headers on non-2XX for static assets

devinivy:
Need to use echo's before hook in order to observe the resulting status code, in order to decide whether to set the cache-control header.

Recommend peeking this ignoring whitespace changes!


ericvolp12 (approved):
Nice catch

PR #7481, opened 2025-01-18
Adding check for duplicated muted word

### LINKED ISSUE #7472: “Mute words & tags” menu allows you to add a word to your muted words list multiple times
### Steps to Reproduce

1. Open a post in the iOS app or the web app. 
2. Open the … menu and select “Mute words & tags”.
3. Enter a word to mute, such as “test”, and “Add +” it. 
4. Repeat step 3, entering the same word multiple times. 
5. You will see the word listed multiple times in “Your muted words” list. See the attached screenshot. 

This issue is more of a problem when muting a word that you forgot you already muted earlier because the two muted words won’t appear next to each other in the list. If you later remove one of those entries, you won’t know to remove the other entry. Then you will be confused why the word is still muted even though you removed it from the list. 

Instead of showing an error message to prevent muting of an already-muted word, I recommend silently moving the old muted word to the top of the list, pretending the word was entered successfully. The user didn’t need to know the truth because they got the result they wanted: their word was added to the top of the muted word list. 👍 

### Attachments

Image

iOS

### Device Info

iPhone 15 Plus running iOS 18.2.1

Build version: 1.96.4.818; Bundle info: 615bbc3 (prod); Bundle date: 25010300; Platform: ios; Platform version: 18.2.1

### Additional Information

_No response_

gabcvit:
Closes #7472 

1. Added a check to detect if the word already exists in the array of muted words. If yes, show an error message and do not save
2. Added a new translatable string for this specific scenario
3. Added translations in German and brazilian Portuguese

### Screen recording

https://github.com/user-attachments/assets/f751265a-61ec-4693-bb1d-5c5a3ada1e7f

### DIFF
--- a/src/components/dialogs/MutedWords.tsx
+++ b/src/components/dialogs/MutedWords.tsx
@@ -61,6 +61,12 @@ function MutedWordsInner() {
   const [durations, setDurations] = React.useState(['forever'])
   const [excludeFollowing, setExcludeFollowing] = React.useState(false)
 
+  const isDuplicated = (value: string) => Boolean(
+    (preferences?.moderationPrefs.mutedWords?.find(
+      m => m.value === value,
+    )
+  ))
+
   const submit = React.useCallback(async () => {
     const sanitizedValue = sanitizeMutedWordValue(field)
     const surfaces = ['tag', targets.includes('content') && 'content'].filter(
@@ -84,6 +90,12 @@ function MutedWordsInner() {
     if (!sanitizedValue || !surfaces.length) {
       setField('')
       setError(_(msg`Please enter a valid word, tag, or phrase to mute`))
+      return
+    }
+
+    if (isDuplicated(sanitizedValue)) {
+      setField('')
+      setError(_(msg`This word has already been muted`))
       return
     }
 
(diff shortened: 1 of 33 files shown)


gabcvit:
Would be nice to get some feedback on it.

cpeterso:
@gabcvit, I can't comment on the code change, but as a user, your fix is a much better user experience!

One suggestion: considering the "This word has already been muted" message from red to green. Entering a duplicated word is not an error and red suggests the user did something wrong or something bad happened when, in fact, the message is telling the user the word is muted, just let they wanted. 🙂

gabcvit:
Thanks @cpeterso!
Will look into it and see how I can improve the message look 😄 
Do you know how people can get more attention from the maintainers of the repo? I can imagine they're very busy people, but 2 weeks without any feedback feels strange to me.

cpeterso:
Sorry, I don't know. I've only filed a few issues here. Do the Bluesky developers have a Discord or other forum?

gabcvit:
@cpeterso, I checked around but didn't find anything. I guess I'll just leave the PR here until someone finally decides to give us some attention. Seems that they make fairly explicit that we should expect PRs to be ignored 😢 

Anyways. On a happier note, I extended the current "error message" component to support a more positive look, I renamed it to warning and added more configuration on whether it should be a negative or positive warning. This way people can reuse it for other features as well 🚀 

Here's a video on how it looks like at the moment:

https://github.com/user-attachments/assets/6d9ac58d-be66-492a-9ba4-941fbfb1a5e4

Issue #7486, opened 2025-01-18
popup on @ is inconsistent in web vs android when used with non-latin chars


oprudkyi:
### Steps to Reproduce

popup on @ is inconsistent in web vs android when used with non-latin chars 

1. web ![Image](https://github.com/user-attachments/assets/5f5976d0-632d-49f0-900f-146a3f544730)

2. android (doesn't work at all)

![Image](https://github.com/user-attachments/assets/37028901-bb8a-42ae-be52-c1a4eec6ab28)

random Cyrillic chars for testing 
йцугшщзфіплджєґюбьчя 

### What platform(s) does this occur on?

Android

### Device Info

Samsung Tab S7 FE

### What version of the app are you using?

Build version: 1.96.5.582; Bundle info: 5130d19 (prod); Bundle date: 25011704; Platform: android; Platform version: 34

surfdude29:
I can confirm this issue affects the iOS app too in my testing. You might want to change `android` to `native` :)

PR #7487, opened 2025-01-18
[Android] Fix @ popup for non-latin chars

oprudkyi:
- support for unicode chars
- ignore zero chars case
    
https://github.com/bluesky-social/social-app/issues/7486

### DIFF
--- a/__tests__/lib/strings/mention-manip.test.ts
+++ b/__tests__/lib/strings/mention-manip.test.ts
@@ -33,6 +33,7 @@ describe('getMentionAt', () => {
     ['alice@alice', 0, undefined],
     ['alice@alice', 6, undefined],
     ['hello @alice-com goodbye', 8, 'alice-com'],
+    ['@йцугшщзфіплджєґюбьчя hello', 6, 'йцугшщзфіплджєґюбьчя'],
   ]
 
   it.each(cases)(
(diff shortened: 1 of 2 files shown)


gaearon:
This seems a bit too broad?

oprudkyi:
@gaearon 
as far as I understand, `searchActorsTypeahead` now searches in profile titles , so there might be any character (at least on web ui it pass everything down to search fn )

gaearon:
Yea but we'd probably still want to exclude punctuation? Can we keep this to international alphanumeric?

oprudkyi:
@gaearon I have updated code
though as for me it is now differs from web ui (web supports punctuation as well)

oprudkyi:
@gaearon 
I forgot to add numerics. Fixed now
current version 
```
  let re = /(^|\s)@([\p{L}\d]*)/giu

```

oprudkyi:
rebased onto fresh main

Hi @pfrazee  @BinaryFiddler  
Could you please take a look ?  it fixes incompatibility of @ popup in android and web

surfdude29:
As I just mentioned in #7486, I can confirm that this issue affects the iOS app too in my testing. I can't build the app on iOS so I can't test this fix, but I'm hoping that this will also fix the issue on iOS. You might want to change `Android` to `Native` or `Android/iOS` in the title :)

PR #7488, opened 2025-01-18
Add collapsable threads and tweak "thread replies" appearance

vcarl:
"Thread view" for replies hasn't quite hit the mark for me — there's a lot of vertical screenspace reserved for controls and link embeds, which leaves room for improvement in my eyes on "skimmability". If I'm reading the replies, I'm usually looking for broad-strokes. 

This makes a couple of changes:

- Moves controls (like/repost/reply/menu) to the side of thread replies
- Tightens margins to compensate for reduced horizontal space
- Never collapses text under "shor more" links
- Never shows embeds for thread replies
- Adds a long-press interaction on any thread reply to collapse that post and all descendents

I'm opening this as a draft PR because there are some bugs I want to sort out before calling it fully ready, specifically

- Linear view collapses as well, which was not my intention
- "never collapse text" is not scoped to only thread view, which was my intention.

Here's a before/after comparison
Screenshot 2025-01-18 at 4 46 30 PM

And here's a recording of the collapse experience

https://github.com/user-attachments/assets/92ce2b87-b01a-48a1-9c50-435f27ea06a2


src/view/com/post-thread/PostThread.tsx
```diff
@@ -255,6 +255,28 @@ export function PostThread({uri}: {uri: string | undefined}) {
...
+    if (!skeleton) return set
+
+    for (let val of skeleton.replies) {
+      if (!isThreadPost(val)) continue
+      if (collapsedCIDs.has(val.post.cid) || set.has(val.post.cid)) {
+        val.replies?.forEach(r => {
+          isThreadPost(r) && set.add(r.post.cid)
+        })
+      }
+    }
```
vcarl, line 270: This traverses the tree of posts any time a post is collapsed, which may not be efficient enough in production

src/view/com/post-thread/PostThreadItem.tsx
```diff
@@ -270,13 +286,16 @@ let PostThreadItemLoaded = ({
   }, [openComposer, post, record, onPostReply, moderation])
 
   const onPressShowMore = React.useCallback(() => {
-    setLimitLines(false)
-  }, [setLimitLines])
+    // setLimitLines(false)
+    // }, [setLimitLines])
+  }, [])
```
vcarl, line 291: Flagging this as a blocking bug

vcarl:
Updated to tighten up some details, like de-emphasizing the post meta when collapsed (and removing the link+hovercard so the only interaction is "expand")
Screenshot 2025-01-18 at 10 22 23 PM

futurGH:
Might be easier to get this merged if you split the thread collapsing and actions rework into separate PRs

vcarl:
That's fair, but I do like how it currently preserves positioning of the controls between states, as a bonus on top of the additional post density. Can appreciate others may not see the density as an advantage, happy to split if it becomes a point of debate

vcarl:
Opted to split up the controls changes from the collapse behavior, after using it on a phone and feeling like deeply nested replies got too squished with the horizontal space taken up by controls

Screenshot 2025-01-20 at 6 46 19 PM

src/view/com/post-thread/PostThread.tsx
```diff
@@ -255,6 +257,28 @@ export function PostThread({uri}: {uri: string | undefined}) {
     randomCache,
   ])
 
+  const [collapsedCIDs, setCollapsedCIDs] = React.useState(new Set())
```
mozzius, line 260: I think you want to use URIs rather than CIDs here. URIs are what we use as unique identifiers for posts

vcarl: Done!

surfdude29:
I'd love to have this as an option alongside the regular **Threaded** mode. Maybe this could be called **Threaded (Compact)**?

IMG_2971

vcarl:
The failing checks are on files I didn't touch 👀

vcarl:
@surfdude29 I don't think the collapse mechanic needs to be under a new UI interaction mode. If you're referencing my previously proposed alternate controls experience, I want to table that suggestion for the purposes of this PR to keep it focused on the collapse behavior — I haven't yet opened a new PR for the UI change but I could if desired

mozzius:
Rebasing should fix the CI

surfdude29:
@vcarl Sorry if I misunderstood, I was assuming this would be as an alternate option to the existing Threaded mode?

Because otherwise this part:

> Never shows embeds for thread replies

would be a regression from the current Threaded mode (which would lead to a great deal of user confusion). Please correct me if I'm misunderstanding though.

vcarl:
"Regression" is a technical term regarding the reintroduction of a previously fixed bug; your complaint is about a UX decision I've made, not a regression.

This is a change to the rendering of replies to a selected post, which I believe to be correct on the hypothesis that people using "Threaded Mode" are seeking a social media experience akin to reddit. Through that lens, replies to a top-level post are more like "comments" than posts in their own right. As such, deemphasizing attachments in turn emphasizes the contents of the replies themselves in a way that I believe is more desired by those who use Threaded Mode — or, maybe I'm just speaking for myself, as this is my own preference.

It is a change to behavior though so if there's agreement that it needs its own PR I can make that alteration. This PR was the end result of playing in the codebase for an afternoon so I can accept if some of the changes aren't meant to be.

Here's a comparison of this PR:

Screenshot 2025-02-15 at 1 38 07 PM

Vs production appearance:

Screenshot 2025-02-15 at 1 38 27 PM
Screenshot 2025-02-15 at 1 38 33 PM

It's a pretty small change, and one I personally strongly prefer — I don't want some 5th-level reply taking up a third of my screen just because they shared a link. Links _above_ the selected post, i.e. post ancestors, will still show link previews.

surfdude29:
> This is a change to the rendering of replies to a selected post, which I believe to be correct on the hypothesis that people using "Threaded Mode" are seeking a social media experience akin to reddit. Through that lens, replies to a top-level post are more like "comments" than posts in their own right. As such, deemphasizing attachments in turn emphasizes the contents of the replies themselves in a way that I believe is more desired by those who use Threaded Mode — or, maybe I'm just speaking for myself, as this is my own preference.

Your line of reasoning does have a lot of merit.

To explain my thinking further, I'm of the view that link cards and external media embeds are an important part of the user experience, whether they're in a root post or a reply. Link cards give a great deal more context than the bare URL alone does – in showing the title, description and photo for a linked article, the link card conveys a lot of useful information and helps the user decide if they want to visit the link. And media embeds allow the user to actually view or listen to the content immediately, without having to open a browser or another app.

So I think that making embeds invisible in replies by default would be a significant loss of functionality and user value compared to the current Threaded mode.

I also think it would cause a lot of confusion. Consider the following reply chain where, as is often done, users have chosen to remove the links after the embed has been added:

> [root post]
>
> this goes hard
> [YouTube embed]
>>[1st reply]
>>
>>this one's great too
>>[Apple Music/Spotify/Bandcamp (hopefully soon?) embed]
>>>[2nd reply]
>>>
>>>also
>>>[YouTube embed]

If link cards and media embeds are made invisible by default in Threaded mode then those users would only see this:

> [root post]
>
> this goes hard
> [YouTube embed]
>>[1st reply]
>>
>>this one's great too
>>>[2nd reply]
>>>
>>>also

The YouTube embed in the root post still shows up fine, but in the replies only the text is visible. This would be pretty confusing, especially as users in Linear mode would be able to see the embeds fine but those in Threaded mode wouldn't.

So, as I'm sure you've gathered, my opinion is that making embeds invisible in replies _by default_ is really not a good idea.

I want to emphasise, though, that I think you've done a great job overall with this PR and I would really like to see a version of this merged in (I'm looking forward to using it myself.)

Therefore, instead of my previous assumption/suggestion of adding this as a separate mode alongside the existing Threaded mode – which, on reflection, would be less than ideal because it would add significant additional complexity – what about just making the behaviour of hiding embeds in replies optional?

Threaded mode is already classed as Experimental, so beneath the existing checkbox which enables it you could just have another checkbox saying something like:

> Hide link cards and external media in replies    [  ]

By doing it this way, users who like and want to retain the current behaviour of embeds being shown in replies could leave the setting off, and users who prefer Threaded mode in the most compact form possible could enable that option and choose to hide all external embeds in replies.

vcarl:
> instead of my previous assumption/suggestion of adding this as a separate mode alongside the existing Threaded mode […] what about just making the behaviour of hiding embeds in replies optional?

Sorry — I'm not really seeing the difference between these suggestions? Wouldn't either of these result in a new configuration option in the dropdown, and the difference in what you're suggesting is about precisely what behavior is being toggled by that option? Either suggestion requires pretty significant additional complexity by way of wiring this behavior up to a setting, and I strongly prefer not to block this PR on that additional code, to the point where I would prefer to remove this behavior before making it optional.

> The YouTube embed in the root post still shows up fine, but in the replies only the text is visible. This would be pretty confusing, especially as users in Linear mode would be able to see the embeds fine but those in Threaded mode wouldn't.

"confusing" mostly means "unfamiliar", confusing behaviors are introduced all the time — I didn't realize I could long-press on the profile image in the bottom bar to switch accounts on mobile for weeks, for instance. I don't believe this will be so confusing as to mislead people.

It has occurred to me, though, that it's possible for a post to delete the URL after an embed is added to the post, and that is an edge case that should be handled before merging this PR. I'll await further information about what requirements must be met to get this PR approved before investing further energy, though.

vcarl:
Checking in -- who has authority to approve this, and how can I get a final answer about what behaviors are okay vs need rework?

mozzius:
Hey - I'm tracking this, although I don't have a timeline for getting it in

Olaren15:
Having been a lot on reddit, I really miss not being able to collapse whole threads on bluesky. Hoping for this PR to get merged!

vcarl:
This absolutely has a ton of conflicts, and as it's been over 6 months without motion here, I'm going to close this.

PR #7496, opened 2025-01-19
Update messages.po

Snipufin:
Changed StepFinished "Avaa" to "Avoin": the StepFinished screen is using Open as an adjective and not as a verb like the Finnish version currently does.


Snipufin:
image

surfdude29:
cc: @valtlai

valtlai:
I confirm this is a correct fix and should be merged. Thank you, @Snipufin!

pfrazee (approved):
Thanks all!

PR #7500, opened 2025-01-20
Fix Discover posts showing up in Following with no explanation

gaearon:
When we run out of posts in Following, we fall back to showing posts from Discover.

This shouldn't generally happen unless you run out of posts from your follows. However, a few days ago we had a backend migration that caused the timelines to restart. So we were falling back into this mode even for who folks who follow people.

Unfortunately the conditions that caused the UI to show up got lost with one of the refactors so they were showing up without explanation. This restores the explanation.

## Test Plan

Screenshot 2025-01-20 at 01 58 52

Screenshot 2025-01-20 at 02 03 20

Screenshot 2025-01-20 at 01 59 41

### DIFF
--- a/src/view/com/posts/PostFeed.tsx
+++ b/src/view/com/posts/PostFeed.tsx
@@ -75,9 +75,8 @@ type FeedRow =
       key: string
     }
   | {
-      type: 'slice' // TODO can we remove?
+      type: 'fallbackMarker'
       key: string
-      slice: FeedPostSlice
     }
   | {
       type: 'sliceItem'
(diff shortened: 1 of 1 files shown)


src/view/com/posts/PostFeed.tsx
```diff
@@ -599,15 +604,14 @@ let PostFeed = ({
...
         return <TrendingVideosInterstitial />
+      } else if (row.type === 'fallbackMarker') {
+        // HACK
+        // tell the user we fell back to discover
+        // see home.ts (feed api) for more info
+        // -prf
+        return <DiscoverFallbackHeader />
       } else if (row.type === 'sliceItem') {
         const slice = row.slice
-        if (slice.isFallbackMarker) {
```
gaearon, line 604: this wasn't working because the row for this slice never got pushed (because it had no items). the fix it to make it a first-class row type now that is easy to do.

src/view/com/posts/PostFeed.tsx
```diff
@@ -75,9 +75,8 @@ type FeedRow =
       key: string
     }
   | {
-      type: 'slice' // TODO can we remove?
```
gaearon, line 78: unrelated driveby dead code removal

pfrazee (approved):

PR #7502, opened 2025-01-20
Revert "Revert "[Android] Fix taps triggering while swiping""

gaearon:
Reverts bluesky-social/social-app#7498

Let's merge after next release is cut


gaearon:
actually let's just use upstream https://github.com/bluesky-social/social-app/pull/7503

PR #7507, opened 2025-01-20
Update Russian phrase

IslamRustamov:
Problem:

Phrase "Начните одну!" is a word-by-word translation of "Start one!" which is not something that we usually say in Russian language, it _doesn't sound right_.

Solution:

A more appropriate ways to translate it would be:
1. Создайте первую!
2. Начните новую!
3. Создайте новую!


surfdude29:
cc: @Virenbar @ponfertato @DearFox

ponfertato:
It seems like something is missing here...
https://github.com/bluesky-social/social-app/blob/main/docs%2Flocalization.md

pfrazee (approved):
Looks legit in google translate. Thanks!

PR #7510, opened 2025-01-20
Fix crash when nothing is pinned

gaearon:
The issue was `.split` call on potentially `undefined` (when `allFeeds` is `[]`).

A proper way to protect against this would be to enable `"noUncheckedIndexedAccess": true` in the TS config. But this raises too many violations.

## Before

Screenshot 2025-01-20 at 19 22 30

## After

Screenshot 2025-01-20 at 19 22 38

## Test Plan

Unpin all feeds. Verify no crashes on all platforms. Re-add feeds. Verify you can switch tabs by swiping, tapping, or (on desktop) the right pane menu.

### DIFF
--- a/src/view/screens/Home.tsx
+++ b/src/view/screens/Home.tsx
@@ -15,7 +15,7 @@ import {logEvent} from '#/lib/statsig/statsig'
 import {isWeb} from '#/platform/detection'
 import {emitSoftReset} from '#/state/events'
 import {SavedFeedSourceInfo, usePinnedFeedsInfos} from '#/state/queries/feed'
-import {FeedParams} from '#/state/queries/post-feed'
+import {FeedDescriptor, FeedParams} from '#/state/queries/post-feed'
 import {usePreferencesQuery} from '#/state/queries/preferences'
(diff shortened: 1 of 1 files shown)


src/view/screens/Home.tsx
```diff
@@ -149,16 +150,20 @@ function HomeScreenReady({
...
       lastPagerReportedIndexRef.current = index
-      setSelectedFeed(feed)
-      logEvent('home:feedDisplayed', {
-        index,
-        feedType: feed.split('|')[0],
-        feedUrl: feed,
-      })
+      setSelectedFeed(maybeFeed)
+
+      if (maybeFeed) {
```
gaearon, line 160: Making this conditional is the fix.

src/view/screens/Home.tsx
```diff
@@ -135,11 +136,11 @@ function HomeScreenReady({
 
   useFocusEffect(
     useNonReactiveCallback(() => {
-      if (selectedFeed) {
+      if (maybeSelectedFeed) {
```
gaearon, line 139: this is just a var rename

pfrazee (approved):

PR #7516, opened 2025-01-20
show video feeds from posts even if they are replies

### LINKED ISSUE #7509: Videos which are replies do not appear in native video feeds
### Steps to Reproduce

0. On web, find a video in a video feed that is a reply to another post
1. Open that feed on mobile
2. video is not included in video feed

### Attachments

_No response_

iOS

### Device Info

_No response_

1.96.5.935

### Additional Information

When feeds get generated to send to the client, they get reordered by the feed tuner:
https://github.com/bluesky-social/social-app/blob/5090426f9c19f7c372b4426544790683b63b95cc/src/lib/api/feed-manip.ts#L51

This causes the first post in a reply chain becomes the root, and gets put into the item list first. The new code looks at the first post in the reordered list to see whether to include the slice in the list of items to show in the native video feed:
https://github.com/bluesky-social/social-app/blob/5090426f9c19f7c372b4426544790683b63b95cc/src/view/com/posts/PostFeed.tsx#L350

Since post context isn't shown in the mobile video view, it's probably worth it get rid of the context and just use the `_feedPost` as the only element in the slice, and I imagine that the cleanest place to do that would be in the feed tuners: https://github.com/bluesky-social/social-app/blob/bf95345b333c56876cabf4c5b8516c431cc8ce9b/src/state/preferences/feed-tuners.tsx#L14 as a new `FeedTuner.removeContext` can remove everything but the `_feedPost`, but I'm not quite sure where that tuner would be able to see the feed `contentMode`.

kaytwo:
fixes #7509

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -41,6 +41,7 @@ export class FeedViewPostsSlice {
   isFallbackMarker: boolean
   isOrphan: boolean
   rootUri: string
+  feedPostUri: string
 
   constructor(feedPost: FeedViewPost) {
     const {post, reply, reason} = feedPost
@@ -48,6 +49,7 @@ export class FeedViewPostsSlice {
     this.isIncompleteThread = false
     this.isFallbackMarker = false
     this.isOrphan = false
+    this.feedPostUri = post.uri
     if (AppBskyFeedDefs.isPostView(reply?.root)) {
       this.rootUri = reply.root.uri
     } else {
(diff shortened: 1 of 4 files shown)


gaearon (changes requested):
i think we'd also need to adjust the logic in `VideoFeed/index` to match this. currently it feels wrong in the opposite way — seems like it just enumerates all items that happen to be videos (whether root, parent, or leaf). would you like to update it to be 1:1 to this code?

gaearon (approved):

PR #7517, opened 2025-01-20
Add videos tab to profile

estrattonbailey:
Awaiting backend support


src/screens/Profile/Sections/Feed.tsx
```diff
@@ -45,6 +45,7 @@ export const ProfileFeedSection = React.forwardRef<
   const [hasNew, setHasNew] = React.useState(false)
   const [isScrolledDown, setIsScrolledDown] = React.useState(false)
   const shouldUseAdjustedNumToRender = feed.endsWith('posts_and_author_threads')
+  const isVideoFeed = isNative && feed.endsWith('posts_with_video')
```
ebwinters, line 48: These strings are sprinkled all over the code base, I don't know why they aren't in a `feedConstants.ts` file or something

estrattonbailey: Yep we're aware, our `FeedDescriptor` handling is a bit messy in general, and this is just a part of what we need to clean up at some point

gaearon (approved):

PR #7528, opened 2025-01-21
Some immersive video a11y fixes

estrattonbailey:
Handles a few simple cases:
- play/pause state on tap control
- disables expandable text control, shows and reads all text
- clarifies follow actions
- clarifies what video you're viewing


gaearon (approved):
code lgtm, did not test

PR #7534, opened 2025-01-21
make video posts which are quote posts show up in video feeds

kaytwo:
Because Video posts are fundamentally `EmbedRecordWithMedia` whereas plain video posts are `EmbedVideo`, the shape of the object and where the user's uploaded video exist within the post are different. This PR extracts the relevant `EmbedVideo` wherever relevant and passes it to the renderer.

# Testing

Check this feed on live mobile / dev mobile / desktop: https://bsky.app/profile/feedtester.bsky.social/feed/videos

On desktop it's got 3 videos, on this branch on iOS it has 3 videos, on live it only shows 1 video (the non-quote-post one).

### DIFF
--- a/src/lib/embeds.ts
+++ b/src/lib/embeds.ts
@@ -1,6 +1,7 @@
 import {
   AppBskyEmbedRecord,
   AppBskyEmbedRecordWithMedia,
+  AppBskyEmbedVideo,
   AppBskyFeedDefs,
 } from '@atproto/api'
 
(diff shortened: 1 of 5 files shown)


src/lib/embeds.ts
```diff
@@ -22,3 +23,25 @@ export function isEmbedByEmbedder(
...
+  if (AppBskyEmbedVideo.isView(v)) {
+    return v
+  }
+  if (
+    AppBskyEmbedRecordWithMedia.isView(v) &&
+    AppBskyEmbedVideo.isView(v.media)
+  ) {
+    return v.media
+  }
+  throw new Error(`Invalid video embed ${JSON.stringify(v)}`)
```
gaearon, line 46: Can we avoid this? Instead, it would be nice to do some kind of type refinement or to leave just this function, make it nullable, and then check for `null` instead of `is*`. Basically, let's make this codepath impossible.

kaytwo: I couldn't get the type refinement approach to work (TypeScript can't seem to understand that `v.media` is definitely a `AppBskyEmbedVideo.View` no matter how explicit I get with the parameter typing), so I went with the null approach instead. There are some conflicts with your refactor PR - I incorporated the change you made but it will still need to be reconciled.

gaearon:
apologies, this would need a rebase

kaytwo:
@gaearon PTAL, successfully rebased

Issue #7542, opened 2025-01-22
Release 1.97.0


pfrazee:
**Estimated build date: Friday Jan 24**

Sorry for all the patch builds lately, we've been YOLOing. This release will be a chance to catch up our translations with the video feed update. It also includes a smattering of quality-of-life improvements.

pfrazee:
QA has started!

surfdude29:
After the latest merges, Bluesky is now up to 33 localizations! 🦋🌐🎉

![Image](https://github.com/user-attachments/assets/a7627929-a4db-4d46-b53f-d50305aea631)

PR #7446, opened 2025-01-13
Post-report menu

mozzius:
Shows a menu offering to block and/or delete the conversation.

https://github.com/user-attachments/assets/e8aac306-8e0c-4358-b0dd-0dd00f42645d

# Test plan

You might want to comment out the line that sends the report (`ReportDialog.tsx:163`)

1. Test that the core functionality works (it blocks, it leaves)
2. Check the handling of when API calls is slow or fails - it fetches the profile view for example (`ReportDialog.tsx:75`)
3. Check web styles

PR #7546, opened 2025-01-22
French localization for v1.97

Signez:
Thanks to @pfrazee's new Bat-Signal 🦇 (i.e. #7542), here are 50 new translated strings in French for the upcoming v1.97. ✨

Some of them are not new translations, but simply a copy and paste of an old translation whose variable placeholder names have been changed (it happens sometimes...).

Ping @CerryTsuki and @surfdude29, as usual, for a review.  🙇

### DIFF
--- a/src/locale/locales/fr/messages.po
+++ b/src/locale/locales/fr/messages.po
@@ -8,7 +8,7 @@ msgstr ""
 "Language: fr\n"
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"PO-Revision-Date: 2024-12-14 17:17+0100\n"
+"PO-Revision-Date: 2025-01-22 20:43+0100\n"
 "Last-Translator: Stanislas Signoud (@signez.fr)\n"
 "Language-Team: Stanislas Signoud (@signez.fr), surfdude29, Cereza Auclair (@cereza.zone)\n"
 "Plural-Forms: \n"
@@ -893,7 +893,7 @@ msgstr "Vous confirmez ?"
 
 #: src/view/com/composer/select-language/SuggestedLanguage.tsx:69
 msgid "Are you writing in <0>{suggestedLanguageName}</0>?"
-msgstr ""
+msgstr "Écrivez-vous en <0>{suggestedLanguageName}</0> ?"
 
 #: src/screens/Onboarding/index.tsx:23
 #: src/screens/Onboarding/state.ts:95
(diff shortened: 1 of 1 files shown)


src/locale/locales/fr/messages.po
```diff
@@ -2462,11 +2462,11 @@ msgstr "Activer les tendances"
 
 #: src/screens/Settings/ContentAndMediaSettings.tsx:158
 msgid "Enable trending videos in your Discover feed"
-msgstr ""
+msgstr "Activer les vidéos tendances dans votre flux Discover"
```
surfdude29, line 2449: We've previously translated `feed` as `fil d’actu` so I think it would be good to stick to that here. (And fwiw we've actually used `flux` as the translation for `flow` elsewhere.)

src/locale/locales/fr/messages.po
```diff
@@ -2462,11 +2462,11 @@ msgstr "Activer les tendances"
...
 
 #: src/screens/Settings/ContentAndMediaSettings.tsx:144
 msgid "Enable trending videos in your Discover feed."
-msgstr ""
+msgstr "Activer les vidéos tendances dans votre flux Discover."
```
surfdude29, line 2453: And ditto here :)

src/locale/locales/fr/messages.po
Signez: Damn, I make this mistake all the time. Thanks for the head's up.

surfdude29: By all means, change the translation for `feed` to `flux` if that seems a better fit and makes sense! You're the native speaker after all 😅

Signez: I should think about it at some point yeah, but it would need its own pull request for a coherent change on the whole translation file, in the future :)

surfdude29:
@Signez FYI in case you have time and motivation to do translations of new strings, there were more added in:
- #7446 74bb65714b7c7b128ddb27438773b149bbe5ec6c
- #7566 f1d120d416bc262b5fd63e3407f9e0b09b13cd74

Or they can just wait until later ofc, neither group are that highly visible imo.

Signez:
_Let's strike the steel while it's hot._ Those 20-ish strings about appeals are now translated 👍

surfdude29:
Nice! LGTM ✅

pfrazee (approved):
Wonderful, thanks all!

PR #7548, opened 2025-01-22
Polish translations for 1.97

axiand:
Just a quick translation and maintenance pass to catch up to the most recent release. 

ping: #7542


pfrazee (approved):
🙏

PR #7554, opened 2025-01-23
Refine accessibilityHint

quiple:
I've made refinements to accessibilityHint based on [this document](https://developer.apple.com/documentation/objectivec/nsobject/1615093-accessibilityhint). English is not my native language, so please correct any awkwardness.

* Avoids beginning the phrase with the imperative form of a verb.
* Removes the phrases "Tap to", "Press to", etc. that precede the sentence. Should I also remove "Double tap to"?
* Removes periods at the end of sentences.

### DIFF
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -232,7 +232,7 @@ export const LoginForm = ({
               onSubmitEditing={onPressNext}
               blurOnSubmit={false} // HACK: https://github.com/facebook/react-native/issues/21911#issuecomment-558343069 Keyboard blur behavior is now handled in onSubmitEditing
               editable={!isProcessing}
-              accessibilityHint={_(msg`Input your password`)}
+              accessibilityHint={_(msg`Enter your password`)}
             />
             <Button
               testID="forgotPasswordButton"


src/view/screens/Search/Search.tsx
```diff
@@ -855,7 +855,7 @@ export function SearchScreen(
                   hitSlop={HITSLOP_10}
                   label={_(msg`Menu`)}
                   accessibilityHint={_(
-                    msg`Access navigation links and settings`,
+                    msg`Accesses to navigation links and settings`,
```
jason-me, line 858: Consider "Provides access to" instead.

src/components/ProgressGuide/FollowDialog.tsx
```diff
@@ -762,7 +762,7 @@ function SearchInput({
         autoComplete="off"
         autoCapitalize="none"
         accessibilityLabel={_(msg`Search profiles`)}
-        accessibilityHint={_(msg`Search profiles`)}
+        accessibilityHint={_(msg`Search for profiles`)}
```
jason-me, line 765: Consider "Searches for profiles" "Search for" is imperative.

src/components/dms/dialogs/SearchablePeopleList.tsx
```diff
@@ -505,7 +505,7 @@ function SearchInput({
         autoCapitalize="none"
         autoFocus
         accessibilityLabel={_(msg`Search profiles`)}
-        accessibilityHint={_(msg`Search profiles`)}
+        accessibilityHint={_(msg`Search for profiles`)}
```
jason-me, line 508: "Searches for"

jason-me (commented):
Everything else looks good!

src/components/ProgressGuide/FollowDialog.tsx
surfdude29: @jason-me In case it's helpful to know, when reviewing a PR you can click the button with the symbol of a page with a ± and this lets you provide a suggestion that can then be easily accepted by the author :)

E.g.:
```suggestion
        accessibilityHint={_(msg`suggested text here`)}
```

quiple:
@jason-me Thanks! I've updated your suggestions.

jason-me (commented):
Approve changes

pfrazee (approved):
Ah wow, thank you for this. We did not know!

SleeplessByte:
Note for future reader: whilst the `accessibilityHint` expects a action result such as plays, edits, removes, `aria-label` will expect an action to take, such as play, edit, remove. This is one of the reasons why using one property for both is sub-optimal. 

Given that there are way worse accessibility breaking bugs, perhaps something we can look into in the future.

src/components/ProgressGuide/FollowDialog.tsx
jason-me: Thanks for the tip!

PR #7555, opened 2025-01-23
Trim leading whitespace-only lines

gaearon:
This adds trimming of *leading whitespace-only lines* to posting.

The goal is to reduce cases where people accidentally linebreak before their replies.

This doesn't break ASCII art — we don't trim whitespace on leading lines that contain non-whitespace characters.

## Test Plan

Read the regex, convince yourself it's correct.

Try cases like this:

```
 
    

  test
test
  test  
  
```

Expected output:

```
  test
test
  test
```


pfrazee (approved):
Tested the regex, LGTM

PR #7557, opened 2025-01-23
Upgrade RN to 0.76.6

gaearon:
This seems like it has a few crash fixes and should be safe.

https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0766

The motivation is https://github.com/facebook/react-native/issues/47490#issuecomment-2604408116 — this fix only affects the new architecture but we might as well bump now and then rebase the Fabric PR on the same patch.

## Test Plan

Build, walk around the app.


estrattonbailey (approved):
Tested on web, iOS (sim and physical device), Android (ditto), found no regressions after using the app for a couple minutes in each build

PR #7319, opened 2024-12-31
New spanish translations

dgruano:
As promised, some more updates for the Spanish localization:

- Add new translations to reach 100% coverage
- Change example name from Alice to Alicia in all examples
- Make new strings translatable
  - "Posts" and "Publications" (lists page) and "Following" tab
  - Run lingui extract
  - Add translation to Spanish for those strings
- Some fixes:
  - Roll back to "Me gusta" for the "Likes" tab in profile (my bad)
  - Change "desmutear" and derivatives to "Dejar de silenciar"
  - Other minor fixes

- To validate:
  - src/screens/Profile/components/ProfileFeedHeader.tsx:353
  menú de feeds (menu "of feeds") or menú del feed (menu "of the feed")?

  - src/view/screens/ProfileList.tsx:683 (and others)
  Dejar de fijar?"

- TODO: Delete obsolete strings from locales. Is there any criteria to do this, or can they be deleted right away?

PR #7564, opened 2025-01-23
Spanish localization for 1.97.0

hogaza:
Following the announcement of the new version on #7542, here are newly translated strings and some corrections to previous ones. I have also taken into consideration the quotation marks issue mentioned on #7319 by @dgruano - I hope to look into the small things that can be improved once Crowdin is set up.


pfrazee (approved):
🙏

PR #7232, opened 2024-12-21
Add Swedish localization

nilaallj:
Following instructions from [#6244 (comment)](https://github.com/bluesky-social/social-app/issues/6244#issuecomment-2469391936).

Almost all the strings are translated (1755 out of 1810, 96%). It’d be awesome to get some help with the remaining ones and have someone review the whole thing to make tweaks or improvements if needed.

PR #7573, opened 2025-01-24
Updated Swedish translations for v.1.97

nilaallj:
Translations for new strings that were added after the initial PR (#7232).

Ping #7542


pfrazee (approved):
TYSM!

PR #7575, opened 2025-01-25
update bitdrift to 0.6.7 and enable iOS network integration

snowp:
This updates the bitdrift library to 0.6.7, which brings in a new version of the Capture SDK that addresses the limitations with the iOS networking integration that caused it to not play well with the bluesky app.

### DIFF
--- a/src/lib/bitdrift.ts
+++ b/src/lib/bitdrift.ts
@@ -18,6 +18,7 @@ initPromise.then(() => {
   if (isEnabled && BITDRIFT_API_KEY) {
     init(BITDRIFT_API_KEY, SessionStrategy.Activity, {
       url: 'https://api-bsky.bitdrift.io',
+      enableNetworkInstrumentation: true,
     })
   }
 })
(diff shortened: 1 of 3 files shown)


src/lib/bitdrift.ts
```diff
@@ -18,6 +18,7 @@ initPromise.then(() => {
   if (isEnabled && BITDRIFT_API_KEY) {
     init(BITDRIFT_API_KEY, SessionStrategy.Activity, {
       url: 'https://api-bsky.bitdrift.io',
+      enableNetworkInstrumentation: true,
```
Maker-Mark, line 21:
```suggestion
      enableNetworkInstrumentation: true, // Only effects iOS, Android instrumentation is set via Gradle Plugin
```

Maker-Mark (commented):
Nice to see bluesky using bitdrift :) I don't see the android plugin added for network instrumentation, so I left a small suggestion to make the init config more clear.

snowp:
I bumped this up to 0.6.7 since that includes a fix to the new iOS swizzling logic

PR #7220, opened 2024-12-20
Add Danish localization

c960657:
Add translation for Danish (da).

Demo site: https://bsky.aggemam.dk

PR #7579, opened 2025-01-25
 Update Danish localization for 1.97.0 

c960657:
Add missing strings to Danish (da) locale for v1.97.0.

While waiting for #7220 to be merged, I was monitoring the commit log of [src/locale/locales/en/messages.po](https://github.com/bluesky-social/social-app/blob/main/src/locale/locales/en/messages.po) to see if any new strings were added, but it seems this was not a good way to check for that :-)


pfrazee (approved):
TYSM!

PR #7587, opened 2025-01-26
feat(sign-up): Add tip about domain name handles

shrink:
A point of confusion when registering for Bluesky is the use of "username". Many people signing up only know of Bluesky as a Twitter alternative and do not have context around the protocol. During the sign up process, a user may be unable to find an available username and be discouraged from signing up[1].

In the most technically correct world, the sign-up process would not use the term "username" and would allow a user to choose between a handle using the `.bsky.social` domain or a handle using their own domain... but that is much more complicated ux -- most users do not understand domain names, nor do they want to.

A good middle-ground is informing users about support for domain names (with the term "verify" to plant the seeds that domain names can be used for verification) while allowing them to choose a username.

I don't love the phrasing but "did you know?" is intended to _inform_ without instructing, i.e: a user reads it and learns something but doesn't think they need to take any specific action. A user with an established online identity, or a user frustrated their chosen usernames are taken might think "ah! I know that domains can be used, I'll sign up with whatever as my username and then switch to my domain after" whereas a _normal_ user might think something on the spectrum of "neat, I know what domains are, it's cool that Bluesky uses them" to "I don't know what a domain is but verification sounds positive".

Screenshot 2025-01-26 at 16 28 18

[1] https://bsky.app/profile/luciferrari.tifosi.social/post/3lgl7s7n4ok2c

### DIFF
--- a/src/screens/Signup/StepHandle.tsx
+++ b/src/screens/Signup/StepHandle.tsx
@@ -13,6 +13,7 @@ import {useAgent} from '#/state/session'
 import {ScreenTransition} from '#/screens/Login/ScreenTransition'
 import {useSignupContext} from '#/screens/Signup/state'
 import {atoms as a, useTheme} from '#/alf'
+import {Admonition} from '#/components/Admonition'
 import * as TextField from '#/components/forms/TextField'
 import {useThrottledValue} from '#/components/hooks/useThrottledValue'
 import {At_Stroke2_Corner0_Rounded as At} from '#/components/icons/At'
@@ -183,6 +184,16 @@ export function StepHandle() {
             </View>
           </View>
         )}
+        <View>
+          <Admonition type="tip">
+            <Trans>
+              <Text style={[a.font_bold]}>Did you know?</Text> Bluesky supports
+              domain name handles like @example.com. Sign up with a username
+              first and then switch to your domain name when you are ready to
+              self-verify.
+            </Trans>
+          </Admonition>
+        </View>
       </View>
       <BackNextButtons
         isLoading={isLoading}


src/screens/Signup/StepHandle.tsx
```diff
@@ -183,6 +184,16 @@ export function StepHandle() {
             </View>
           </View>
         )}
+        <View>
+          <Admonition type="tip">
+            <Trans>
+              <Text style={[a.font_bold]}>Did you know?</Text> Bluesky supports
+              domain name handles like @example.com. Sign up with a username
+              first and then switch to your domain name when you are ready to
+              self-verify.
```
surfdude29, line 193:
```suggestion
              You can use a domain name as your username on Bluesky, for example
              <Text style={[a.font_bold]}>@mywebsite.com</Text>. Create an
              account with a <Text style={[a.font_bold]}>.bsky.social</Text>
              username first and then change it to your domain name when you're
              ready to self-verify your identity.{' '}
              <InlineLinkText
                label={_(msg`learn more`)}
                to="https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial"
                style={[a.font_bold]}
                disableMismatchWarning>
                Learn more here.
              </InlineLinkText>
```

I can't add this as a suggestion but this would have to be added too of course:

`import {InlineLinkText} from '#/components/Link'`

surfdude29 (commented):
I think this is a great idea!

I have a suggestion for slightly different wording that I think might be a bit clearer for new users and which also echoes the language used in the **Change Handle** dialog. Also, I think it would be good to have a **Learn more here** link to provide a more detailed explanation for those who would like it (also like in the **Change Handle** dialog).

https://github.com/bluesky-social/social-app/blob/355c50fc0fe97feb8b4ec4e29d47b725252088c7/src/screens/Settings/components/ChangeHandleDialog.tsx#L249-L257

src/screens/Signup/StepHandle.tsx
shrink: Thank you for the feedback :) I agree, a link to learn more would be great! That said, the domain handle tutorial blog post is almost 2 years old now and the language doesn't reflect the language used in the change handle flow: the change handle flow uses "handle" and "domain" whereas the blog post uses "username". 

I think "username" is fine in the context of sign-up (where a user can only choose the username part of a `.bsky.social` handle) but after sign up when a user does have full control over their handle the term "handle" is used exclusively (as far as I can tell).

I guess that means there's a separate action item for this issue: someone from the Bluesky team to update the blog post to use language consistent with the current Bluesky interface (unfortunately I don't think we can as the `bsky.social` website doesn't appear to be open-source).

Taking your suggestion but replacing "username" with "handle" (and assuming the blog post has been updated) would give us:

> You can use a domain name as your handle on Bluesky, for example **@example.com**. Create an account with a **.bsky.social** handle first and then change it to your domain name when you're ready to self-verify your identity. [Learn more here](https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial).

The reason to use `@example.com` instead of `@mywebsite.com` is that `mywebsite.com` exists whereas `example.com` is reserved for "illustrative examples in documents". That said, the repetitive appearance of the word "example" in "for example @example.com" does look a little weird. Maybe instead we could use real examples that are links themselves? e.g:

> You can use a domain name as your handle on Bluesky, for example [@nytimes.com](https://bsky.app/profile/nytimes.com) and [@jay.bsky.team](https://bsky.app/profile/jay.bsky.team). Create an account with a **.bsky.social** handle first and then change it to your domain name when you're ready to self-verify your identity. [Learn more here](https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial).

Let me know if you have any thoughts!

surfdude29: Thanks for the feedback on my feedback! :)

On the language used, there's a bit of backstory, as the Bluesky team have recently been making efforts to simplify things for new users (and fwiw I think the changes have been good ones).

For instance, as part of [a set of ELI5 PRs](https://github.com/bluesky-social/social-app/pulls?q=is%3Apr+ELI5+is%3Aclosed), in #7136 they changed the wording on the signup screen to refer to `username` rather than `handle`.

This followed an [update in November](https://bsky.app/profile/bsky.app/post/3lbic4dss222e) of the blog post explaining how to verify your account that changed the language to make it simpler and easier to understand for less technically-inclined users. For example, the title changed from **"[How to set your domain as your handle](https://web.archive.org/web/20241121124046/https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial)"** (Wayback link) to **"[How to verify your Bluesky account](https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial)"** (the date seems to only reflect when it was originally published).

As to the inconsistency between the language used in the signup flow and the language used in Settings, @/gaearon said in #7136:

> It's a little inconsistent with what we have in Settings. Maybe in the future we'll change it there too. But for now let's make sure people are able to create an account.

So I think for the language used in the signup flow we should take the lead from the recent language changes by the Bluesky team. I kind of tried to do that in my original suggestion, but after looking again at the updated blog post I realise that I didn't actually follow it in some ways. I've updated my suggestion below to follow it closely.

On what to use as the example handle, after considering it I do think your original suggestion of `@example.com` would work best. As you note, `example.com` is specifically reserved for this purpose. Using real accounts like `@nytimes.com` and `@jay.bsky.team` as examples is probably best avoided for several reasons imo (having to get permission, not wanting to appear to give endorsement to `@nytimes.com`, localisation, etc.)

So, taking all that into account (sorry it's such a long comment), here's my updated suggestion:

> You can set your website as your username on Bluesky, for instance **@example.com**. This is one form of verification and it's our version of a "blue check." Simply create an account with any placeholder username and then change it to your website when you're ready to self-verify your identity. **[Learn more here](https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial).**

It is admittedly rather wordy, but given the number of accounts – I'm thinking mainly of organisations – that still haven't set their domain name as their handle, I think it's worth including the key terms `verification` and `"blue check"` from the language used in the blog post to try to emphasise this point on signup as much as possible (well, at least for those who will read it 😅).

surfdude29: One more thing: I had forgotten that there's a PR already open on adding a tip on the handle screen, but it's not about domain names. Perhaps a combination of the two could be helpful? 🤔
- #6937

shrink: @surfdude29 thank you very much for the additional context! I wasn't aware of any of that. I agree that, with the recent changes made elsewhere in mind, your updated text looks great. I'll update the Pull Request.

(Regarding the use of "username" and "website" instead of "domain" and "handle" I think that's... I see where the idea is coming from, non-technical people will be more familiar with the terms "website" and "username" than "domain" and "handle" but I think there are opportunities to educate users for the greater good in a mission-driven project. Domain names are a fundamental building block of the world wide web, it feels weird to use the _wrong_ term just because it's more familiar. "username" is more ambiguous but in general you have _a_ username per platform (like "my instagram username") and atproto's usernames-across-many-platforms seems like the perfect opportunity to establish that these are _not_ usernames as you know it, these are a great new decentralized approach to identity. Granted, I named my project [handles](https://handles.net) so I have a clear bias towards "handle". Anyway, I'm bikeshedding, atproto and Bluesky are filled with great design decisions so I trust that this _is_ the right decision.)

> One more thing: I had forgotten that there's a PR already open on adding a tip on the handle screen, but it's not about domain names. Perhaps a combination of the two could be helpful? 🤔

I think the concept is great, an informative tip shown only when someone is creating their first account is a clever way to help teach new users without repeating information for existing users. I'd love to see that tip include information about ~domains~ websites! Should I close this Pull Request and add the suggestion to that PR? That'll probably save time and energy on reviewing and merging them separately.

(Something that reddit got right in their changes a few years back is automatically generating a username on sign up so that choosing a username isn't a part of the sign up flow. I imagine that's been discussed before though, and not on topic for this issue).

surfdude29: @shrink No worries!

> Regarding the use of "username" and "website" instead of "domain" and "handle"...

I agree with all of this. In the case of the language used in the signup flow and in the first parts of the blogpost, I think they decided that trying to ensure that less-technical users aren't put off or confused by the correct, but more technical, terminology was the more important consideration there; so that it isn't a barrier to them creating an account or changing to a domain handle (or perhaps, being inspired to ask their IT department to change to a domain handle :)

I'm pretty sure the Bluesky team have statistics on how many people access each step of the signup flow – which, I don't know if you know, used to be way longer – and maybe they were seeing more people than they would expect abandoning signup at the handle screen? (This is just speculation on my part, to be clear.) They also may have had user feedback that the language was confusing.

Anyway, imo they've arrived at a reasonable balance now between accessibility to less-technical users in places where that's important, while still using the more specific/correct terminology elsewhere.

For instance, comparing the [old version](https://web.archive.org/web/20241121124046/https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial) and the [current version](https://bsky.social/about/blog/4-28-2023-domain-handle-tutorial) of the blogpost, what used to be the introductory text that explains what domain names are, why Bluesky uses them as handles, etc hasn't been removed from the blogpost, it's just been moved to the end. So it now serves as a explanation of the background and context for users who are interested, while the step-by-step instructions of how to actually change to a domain handle – what most users are likely accessing the page to find out – have been moved to the beginning. (Admittedly, there's some inconsistency with `username` and `handle` both used interchangeably in the later sections.)

So I think there are still opportunities to educate users who want to know more, and thus hopefully help further Bluesky's mission, while not letting technical language get in the way of users who aren't interested in that side of things and just want to focus on having a good time on Bluesky :)

> I think the concept is great, an informative tip shown only when someone is creating their first account is a clever way to help teach new users without repeating information for existing users.

Agreed 👌

> I'd love to see that tip include information about ~~domains~~ websites!

Same 💯

> Should I close this Pull Request and add the suggestion to that PR? That'll probably save time and energy on reviewing and merging them separately.

Up to you! My suggestion would be to leave this PR open for now and write a short comment in #6937 pinging @/mozzius and briefly outlining the purpose/benefit of this PR (maybe including the suggested wording) and then he can take it from there, and hopefully discuss it with other team members :)

> Something that reddit got right in their changes a few years back is automatically generating a username on sign up so that choosing a username isn't a part of the sign up flow.

Yeah I personally hate picking usernames so I'm a big fan of reddit's approach! 😅

> I imagine that's been discussed before though

It might have been, although I can't say that I'm aware of it myself. It might be worth searching the issues to see if someone else has suggested it, and if not, maybe opening one yourself, as I think it's a good idea 👍

PR #7588, opened 2025-01-26
Per-user search history

mozzius:
Refactors the search history to partition the history by DID, so that history is unique to the logged-in account

I also changed it to use ~~react-query, since that's better for handling async state~~ the new MMKV storage!

# Test plan

Generate some search history
Switch to and from accounts to confirm it's partitioned

### DIFF
--- a/src/state/queries/profile.ts
+++ b/src/state/queries/profile.ts
@@ -10,6 +10,7 @@ import {
   ComAtprotoRepoUploadBlob,
 } from '@atproto/api'
 import {
+  keepPreviousData,
   QueryClient,
   useMutation,
   useQuery,
@@ -81,7 +82,13 @@ export function useProfileQuery({
   })
 }
 
-export function useProfilesQuery({handles}: {handles: string[]}) {
+export function useProfilesQuery({
+  handles,
+  maintainData,
+}: {
+  handles: string[]
+  maintainData?: boolean
+}) {
   const agent = useAgent()
   return useQuery({
     staleTime: STALE.MINUTES.FIVE,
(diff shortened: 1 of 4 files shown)


macrouch:
I just realized that Bluesky was on github and came to look into adding this exact feature. So happy to see a PR open already!

estrattonbailey (commented):
Nice, love this! Couple thoughts while we're at it:
- I'd like to avoid random strings for storage and converge around using `#/storage` interface if possible, lmk what you think of that, can chat in Slack
- The `#/storage` interface is sync, which doesn't help here much, but makes stuff a _little_ simpler
- Could pry pull this out into a hook just for code clarity, can live in this file for now

src/state/queries/profile.ts
```diff
@@ -90,6 +97,7 @@ export function useProfilesQuery({handles}: {handles: string[]}) {
       const res = await agent.getProfiles({actors: handles})
       return res.data
     },
+    placeholderData: maintainData ? keepPreviousData : undefined,
```
estrattonbailey, line 100: Oh didn't know they had a helper, v nice

estrattonbailey (approved):
Nice, works great, love the new utils

src/state/queries/profile.ts
mozzius: yeah - used to be a dedicated option, now it's a helper function https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#removed-keeppreviousdata-in-favor-of-placeholderdata-identity-function

PR #7589, opened 2025-01-26
Search - clean up refs in render

mozzius:
Should have zero functionality change, but fixes up bad React code

# Test plan

Confirm it works the same. I am pretty sure this case occurs when the query string is modified - i.e. when the search input is cleared


estrattonbailey (approved):
Yep nice one

PR #7590, opened 2025-01-26
Improved search page

mozzius:
Redesigned search input - it now sits below the header, and moves into place when focused. The language select has been moved to the top right

 

https://github.com/user-attachments/assets/f9f0f330-178d-4fcc-9302-c2840670e674


estrattonbailey (approved):
Very like

PR #7591, opened 2025-01-26
Improved search language select

mozzius:
# Stacked on #7590 

I hate react-native-picker-select with all my heart. Rewrites it to use a `Menu` component

Note: I had to add a hack to cap the dropdown height, to prevent it just going offscreen. I will look at fixing this properly - unfortunately seems there's a bunch of open issues regarding this problem (https://github.com/radix-ui/primitives/issues/2786)

# Native

https://github.com/user-attachments/assets/53416f89-0ef3-47e0-8ed2-6ffd1cc32672

# Web

https://github.com/user-attachments/assets/e78b30e5-5901-4a3d-a23d-f98f80e501cf

## Test plan

Confirm functionality is the same


estrattonbailey (commented):
Should throw a max width on here, or may be finally time to clean up our languages list.

CleanShot 2025-02-12 at 14 55 33@2x

estrattonbailey (approved):

surfdude29:
I noticed something while testing this that isn't ideal (although it's already present on main so isn't caused by this PR): Chinese and Portuguese aren't prioritized in the list along with the other app languages:

IMG_2683

IMG_2684

I think this may be because only app languages with two-letter codes are being prioritized? (which wouldn't include pt-BR, zh-CN, zh-HK and zh-TW)

https://github.com/bluesky-social/social-app/blob/aa1fa28442699f29ea23cb5cba0503be53b6c0ee/src/view/screens/Search/Search.tsx#L350-L356

mozzius:
Thanks @surfdude29, filed https://github.com/bluesky-social/social-app/pull/7731

PR #7594, opened 2025-01-26
Fix slow Hosting Provider dialog

mozzius:
Refactors the hosting provider dialog - it fell into the trap of the cross-portal state changes being really slow. Most state is moved down into the dialog

# Test plan

Confirm no functionality changes

### DIFF
--- a/src/view/com/auth/server-input/index.tsx
+++ b/src/view/com/auth/server-input/index.tsx
@@ -80,36 +80,40 @@ function DialogInner({
     persisted.get('pdsAddressHistory') || [],
   )
 
-  useImperativeHandle(formRef, () => ({
-    getFormState: () => {
-      let url
-      if (fixedOption === 'custom') {
-        url = customAddress.trim().toLowerCase()
-        if (!url) {
-          return null
-        }
-      } else {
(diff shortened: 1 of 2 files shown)


src/view/com/auth/server-input/index.tsx
```diff
@@ -24,177 +24,208 @@ export function ServerInputDialog({
...
-  }, [
-    fixedOption,
-    customAddress,
-    onSelect,
-    pdsAddressHistory,
-    setPdsAddressHistory,
-  ])
+      return url
+    },
+  }))
```
estrattonbailey, line 112: `useImperativeHandle` accepts a dependencies array too, right? May want to toss in the reactive things here

estrattonbailey: Does seem to work OK regardless though 👀

estrattonbailey (approved):
Nice, just one nit

src/view/com/auth/server-input/index.tsx
mozzius: whoopsie yeah :) added

PR #7603, opened 2025-01-27
Fix convo header loading state

mozzius:
Allows for seeding the Convo with placeholder data, which we can fetch from the RQ cache, so that we can skip the header skeleton state. The actual setup process of Convo is unchanged.

# Before

https://github.com/user-attachments/assets/d0f0bb61-c930-4506-8af5-dcd80c8c6399

# After

https://github.com/user-attachments/assets/3ea24ac7-7d97-4aa1-9b12-36d4956707c3

# Test plan

Test entering the convo in different ways, ensure the skeleton is skipped at least when entering from the chat list screen.
Check that the Convo system itself is broadly unchanged, other than allowing the convo and recipients to be defined early

### DIFF
--- a/src/state/queries/messages/conversation.ts
+++ b/src/state/queries/messages/conversation.ts
@@ -1,5 +1,10 @@
 import {ChatBskyConvoDefs} from '@atproto/api'
-import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query'
+import {
+  QueryClient,
+  useMutation,
+  useQuery,
+  useQueryClient,
+} from '@tanstack/react-query'
 
 import {STALE} from '#/state/queries'
 import {DM_SERVICE_HEADERS} from '#/state/queries/messages/const'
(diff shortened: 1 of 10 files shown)


src/state/queries/messages/conversation.ts
```diff
@@ -14,14 +19,17 @@ import {
 const RQKEY_ROOT = 'convo'
 export const RQKEY = (convoId: string) => [RQKEY_ROOT, convoId]
 
-export function useConvoQuery(convo: ChatBskyConvoDefs.ConvoView) {
+export function useConvoQuery(
+  convoId: string,
+  convo?: ChatBskyConvoDefs.ConvoView,
```
estrattonbailey, line 24: Did this need to become optional? `initialConvo` at the callsite should be defined, right, so this fn sig wouldn't necessarily need to change

estrattonbailey (approved):
LGTM, one question on the query fn

estrattonbailey (changes requested):
Actually, we lost profile shadow-based updates when blocking the user from the convo header 🤔

src/state/queries/messages/conversation.ts
mozzius: Ah yeah I can undo that. Motivation was to use it somewhere else, but since that's not needed I'll undo it

estrattonbailey (approved):

PR #7604, opened 2025-01-27
Make convo menu lazy

mozzius:
# Stacked on #7603 

An easy optimisation to make the chat list faster. Nesting the contents of the convo menu inside `Menu.Outer` makes it lazy, so it only renders once the dialog opens.

# Test plan

Ensure all the buttons work
Add a console log to `MenuInner` to confirm it's now lazy


estrattonbailey (commented):
This looks good to me! Noticed an issue in testing from the base branch, commented there.

estrattonbailey (approved):

Issue #7018, opened 2024-12-08
"Search this profile's tweets" option on profile page


simonw:
### Describe the Feature

I'd like to be able to navigate to a profile page (including my own profile page) and select an option to "search in these posts" which takes me to the search box prepopulated with `from:profile-ID`

This would both save me from having to type that, and also save me from having to remember the "from:x" syntax (which I just had to look up).

### Attachments

It could go in this menu:

![screenshot of a profile page, an arrow points to the ... button](https://github.com/user-attachments/assets/0c4e04b1-a220-406d-98d4-4c58f9c64214)

surfdude29:
There's a PR open which would add this: #4373

rednafi:
It’d be really cool if it also included the posts liked by a profile. This would make the search semantics look something as follows:

search term in (posts + comments + liked posts)

hexdotooo:
#7765 suggests that this may already have been implemented?

surfdude29:
Yup, this was fixed by #7622 which is now available in 1.98 💫

PR #7622, opened 2025-01-30
Screen for searching user's posts

mozzius:
# Based on https://github.com/bluesky-social/social-app/pull/7591

Fixes https://github.com/bluesky-social/social-app/issues/2683, https://github.com/bluesky-social/social-app/issues/2684, https://github.com/bluesky-social/social-app/issues/7018.

https://github.com/user-attachments/assets/1d9205cb-a96d-4a7a-ba52-51843380a3b7

# Test plan

- Test new screen on iOS/Android/Web (web especially)
- Ensure no regressions on the primary search page

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -91,6 +91,7 @@ import {VideoFeed} from '#/screens/VideoFeed'
 import {useTheme} from '#/alf'
 import {router} from '#/routes'
 import {Referrer} from '../modules/expo-bluesky-swiss-army'
+import {ProfileSearchScreen} from './screens/Profile/ProfileSearch'
 import {AboutSettingsScreen} from './screens/Settings/AboutSettings'
 import {AccessibilitySettingsScreen} from './screens/Settings/AccessibilitySettings'
 import {AccountSettingsScreen} from './screens/Settings/AccountSettings'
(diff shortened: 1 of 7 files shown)


surfdude29:
I just tried this out, it works really well, excellent job!

I have one small suggestion: what about making the string in the menu conditional on whether you're looking at your own profile (I think it's something like `isSelf`?) so that on your own profile it says `Search my posts` instead?

https://github.com/bluesky-social/social-app/blob/1e1154ab13f494d2c91fdf519805e507a3d03c3b/src/view/com/profile/ProfileMenu.tsx#L218-L221

Also, for reference, this would fix #2683, #2684 and the main request in #7018.

src/Navigation.tsx
```diff
@@ -207,6 +208,13 @@ function commonScreens(Stack: typeof HomeTab, unreadCountLabel?: string) {
         getComponent={() => ProfileListScreen}
         options={{title: title(msg`List`), requireAuth: true}}
       />
+      <Stack.Screen
+        name="ProfileSearch"
+        getComponent={() => ProfileSearchScreen}
+        options={{
+          title: title(msg`Search posts`),
```
haileyok, line 215: Minor nit, should we update the title from the screen itself so that we can make this like `Search @samuel.bsky.team's posts`?

src/view/com/profile/ProfileMenu.tsx
```diff
@@ -177,6 +181,10 @@ let ProfileMenu = ({
     shareText(profile.did)
   }, [profile.did])
 
+  const onPressSearch = React.useCallback(() => {
+    navigation.navigate('ProfileSearch', {name: profile.did})
```
haileyok, line 185: Can we use `profile.handle` instead? Cleaner URL

haileyok:
Functionally works well! Left two comments, and also would be nice if we could not "disappear" the posts on the screen when you focus the search input, but otherwise lgtm.

haileyok (approved):

PR #7624, opened 2025-01-30
Change search cancel behaviour

mozzius:
Before: cancel closes the input and restores previous search input (i.e. search "hello", delete input, press cancel restores input to "hello")
After: cancel clears input and returns to explore page

It is currently frustrating to return to the explore page from the search page. This change aligns behaviour with iOS search inputs (see apple music or whatever)


estrattonbailey (approved):
Yes, love it

PR #7627, opened 2025-01-31
Updated Hindi localisation for 1.97.0

FalKoopa:
The Hindi localisation was last updated in November and it's now been updated to the current version with all the new features added since like Trends.


pfrazee (approved):
Awesome, thanks!

PR #7635, opened 2025-02-02
Use range syntax for media queries

c960657:
In Firefox, if you set the browser zoom level to 90% and resize the window to 1299px wide (check using `document.body.clientWidth` in the browser console), you will see an empty white block above the main feed. It disappears if the window is made a tiny bit wider or narrower.

image

When zoom level is ≠ 100%, the page width be be non-integer. When two media queries are specified as `maxWidth=1299` and `minWidth=1300`, respectively , the width intervals are non-contiguous, and a non-integer width may fall between the two.

Specifying the intervals using range syntax fixes this issue. Range syntax is supported in [all modern browsers](https://caniuse.com/?search=Media%20Queries%3A%20Range%20Syntax).


c960657:
Hang on, this breaks native.

PR #7637, opened 2025-02-02
Typo fix

surfdude29:
Typo fix in string shown after DM report sent:
`recieved` ➡️ `received`

h/t @mackuba for [pointing it out](https://bsky.app/profile/mackuba.eu/post/3lha2lval2s2h)

EDIT: As suggested by @prohr, to avoid string churn this PR also amends the relevant strings in the .po files that have been updated so far (they don't seem to have all been updated as part of the 1.97 release).

### DIFF
--- a/src/components/dms/ReportDialog.tsx
+++ b/src/components/dms/ReportDialog.tsx
@@ -336,7 +336,7 @@ function DoneStep({
           <Trans>Report submitted</Trans>
         </Text>
         <Text style={[a.text_md, t.atoms.text_contrast_medium]}>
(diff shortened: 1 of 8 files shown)


prohr:
Looks like 7 .po files already picked up the typo.  Does the tooling autofix that, or should you do so manually?

surfdude29:
@prohr Good point, to avoid string churn it probably makes sense to fix those .po files as part of this PR, so that's what I've done 👍

mozzius (approved):
damn lol

PR #7639, opened 2025-02-03
Date input improvements

mozzius:
This PR introduces several changes related to the date input:

1. (Motivating this) adds a `maximumDate` field to `DateField`. Prevents entering a birthdate in the future
2. Changes the birthday input to use the new DateField. Results in nested dialogs on iOS, but otherwise much cleaner
3. Allows passing Date objects to `DateField`. Previously, it was the responsibility of the parent to stringify values to YYYY-MM-DD, but we can just accept both and run the util inside the component
4. Now that it's no longer in use, we can delete the legacy date input

https://github.com/user-attachments/assets/d43831ca-0e27-4beb-b6ad-72faa156592a

# Test plan

Test that the date inputs still work (birthday in settings, and signup form)
Test that you can no longer enter a date in the future
Check on all platforms, since implementation is forked significantly


estrattonbailey (approved):
LGTM! There are some comments in here about input/output formats. Mind updating those? Pry only need to note what the changed value format is, or whatever you feel is clearest 👍

mozzius:
Didn't see that - can do!

PR #7640, opened 2025-02-03
Spring cleaning

mozzius:
Deleting some unused files

1. Breakpoint layouts - completely unused and seems like a bad pattern
2. A literally empty file in the Starter Packs wizard
3. Old radio buttons - no longer used
4. Old selectable button - no longer used
5. `#/lib/storage.ts` - superceded by MMKV-based `#/storage`
6. `#/lib/type-assertions.ts` - unused


estrattonbailey (approved):
Looooove

Issue #7524, opened 2025-01-21
When jwt is malformed we show a NotificationScreen in a wrong place


NsdHSO:
### Steps to Reproduce

1. Go to login,
2. Destroy the jwt
3. Do an action that you need jwt

### Attachments

![Image](https://github.com/user-attachments/assets/aeb4a777-a118-451c-aacd-8d0eefb818dd)

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

latest

PR #7655, opened 2025-02-05
Added Portal to ErrorMessage to stack it on top of the view

andrefernandes-95:
## 📝 Summary

Added <Portal> to <ErrorMessage> so that it correctly positions it above all the other elements.

Right now, it seems every <View /> has a z-index: 0, which makes it very tricky to work with the children z-index, causing the issue described here https://github.com/bluesky-social/social-app/issues/7524

I think using a Portal could be the right idea here.

## 📝 Related Issues

https://github.com/bluesky-social/social-app/issues/7524

## 🚀 Demos

https://github.com/user-attachments/assets/fada6cfb-6f11-42f3-ba70-4ea9b15c6d60

(Testing the scroll to guarantee that the error message stays pinned)

https://github.com/user-attachments/assets/5fd6b38a-ef95-42ba-919a-3c7fd47f4087

### DIFF
--- a/src/view/com/util/error/ErrorMessage.tsx
+++ b/src/view/com/util/error/ErrorMessage.tsx
@@ -14,6 +14,7 @@ import {useLingui} from '@lingui/react'
 
 import {usePalette} from '#/lib/hooks/usePalette'
 import {useTheme} from '#/lib/ThemeContext'
+import {Portal} from '#/components/Portal'
 import {Text} from '../text/Text'
 
 export function ErrorMessage({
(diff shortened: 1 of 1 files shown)


mozzius:
Ah thanks for trying to fix it - however, we don't want stack it on top of everything, it just needs to fit in the center column, like it used to (and still does on native). I think wrapping it in <Layout.Center> will do the trick?

andrefernandes-95:
> Ah thanks for trying to fix it - however, we don't want stack it on top of everything, it just needs to fit in the center column, like it used to (and still does on native). I think wrapping it in <Layout.Center> will do the trick?

Thanks for the tip, I've gone back and followed the suggestion, you're spot on, it looks much better and it's within the layout like you mention. Took some screenshots from several ErrorMessage instances to check:

Layout Center
Other Screens 2
Other Screens 3
Other Screens

I'll commit the change and take new evidences. Thanks a lot!

mozzius:
Awesome, thank you so much 🙏 looks great!

mozzius (approved):
Beautiful

Issue #3578, opened 2024-04-16
Search bar and language selector are not visible in PWI in Safari on iPad


surfdude29:
**Describe the bug**

When using the public web interface in Safari on an iPad, the Search bar and the language selector are not visible at the 100% zoom setting.

**To Reproduce**

1. On an iPad, open [bsky.app](https://bsky.app/) in Safari when you are not logged in.
2. Note that you cannot see the Search bar or the language selector, as shown in this screenshot:

![IMG_3189](https://github.com/bluesky-social/social-app/assets/149612116/1179700a-ac56-48f0-a5ac-92f3a278e0bb)

**Expected behavior**

The Search bar and the language selector should be visible when viewing the PWI in Safari on an iPad.

They both do become visible when the zoom is reduced to 75% (screenshot below), but it is not obvious to the user that this will happen and it is not an ideal workaround.

![IMG_3190](https://github.com/bluesky-social/social-app/assets/149612116/0f0a4936-1f8e-4667-8f46-43defe5588f3)

**Details**

 - Platform: iPadOS
 - Platform version: 16.6, iPad (9th generation)
 - App version: 1.78

PR #7656, opened 2025-02-05
Better tablet layout

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/3578

Adds new breakpoints for web:

1. 1075px to 1200px - right nav visible, main content offset to side
2. 800px to 1075px - no right nav

I also moved the left nav on the tablet breakpoint to sit next to the content, rather than being stuck to the far left edge

# 1075px to 1200px

Screenshot 2025-02-06 at 14 27 21

# 800px to 1075px
Screenshot 2025-02-06 at 14 28 39

# Test plan

Check all the screens at the new breakpoints. I have tested:

1. All the common ones I can think of (timeline, feeds, profile, thread, settings etc)
2. Logged-out view
3. Signup queued
4. Deactivated
5. Takendown
6. Onboarding
7. Auth flow

And make sure there's no regressions on the mobile/desktop breakpoints

### DIFF
--- a/src/alf/breakpoints.ts
+++ b/src/alf/breakpoints.ts
@@ -31,8 +31,8 @@ export function useBreakpoints(): Record<Breakpoint, boolean> & {
  * Fine-tuned breakpoints for the shell layout
  */
 export function useLayoutBreakpoints() {
-  const rightNavVisible = useMediaQuery({minWidth: 1050})
-  const centerColumnOffset = useMediaQuery({minWidth: 1050, maxWidth: 1300})
+  const rightNavVisible = useMediaQuery({minWidth: 1075})
+  const centerColumnOffset = useMediaQuery({minWidth: 1075, maxWidth: 1300})
   const leftNavMinimal = useMediaQuery({maxWidth: 1300})
 
   return {
--- a/src/alf/breakpoints.ts
+++ b/src/alf/breakpoints.ts
@@ -26,3 +26,18 @@ export function useBreakpoints(): Record<Breakpoint, boolean> & {
     }
   }, [gtPhone, gtMobile, gtTablet])
 }
+
+/**
+ * Fine-tuned breakpoints for the shell layout
+ */
+export function useLayoutBreakpoints() {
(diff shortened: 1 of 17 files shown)


estrattonbailey (commented):
Clicked around a lot, only issue I found was on the labeler screen. There's some nesting going on here iirc, hence the offset context handling. Might be a clue?

CleanShot 2025-02-06 at 13 48 40@2x

mozzius:
Fixed a few more spots:
1. Labeller screen
2. Dialogs with centered views inside them (MaybeListPlaceholder in Gif dialog for instance)
3. Load latest button
4. Home screen loader

surfdude29:
This works really well on my iPad (9th gen), nice work @mozzius!

Having the search bar, feed list and Trending visible is a massive improvement, thanks! 🙏

I realise it would be out-of-scope for this PR, but just wondering, how technically feasible would it be in the future to move the PWI language selector to the right nav when in tablet layout? The app language is usually autodetected of course so it's not a big deal, but in case there is a desire to change it this would let tablet users do so when they're viewing the PWI (and without changing the zoom setting) and fix #3578.

mozzius:
Great idea @surfdude29! added!

surfdude29:
Oh nice, thanks! :)

surfdude29:
One other small point to mention: should Trending be showing up in the PWI? (Fwiw it should imho.)

Sometimes it shows up for me, sometimes not, so reloading the page can occasionally cause it to appear or disappear.

mozzius:
It should be showing up, yes. Hailey fixed a bug yesterday which was causing ~1% of requests to fail, so hopefully it should be a little more reliable - however, there's some sort of bug with en-GB which I'm still trying to get to the bottom of

surfdude29:
Ah ok, maybe I'm running into the en-GB bug on the preview branch.

mozzius:
Sure, I guess that is broken

estrattonbailey (approved):
New conflict but this LGTM 👍

Issue #7574, opened 2025-01-25
Cannot register longer handles anymore


drallgood:
### Steps to Reproduce

It used to be possible to register handles with longer domains, but as of #7422 this was blocked for "serviceHandles" for some reason, even though the PDS code does support longer handles no problem up to 18 characters for the front part (see https://github.com/bluesky-social/atproto/blob/a8f06939f94b3a5efee4b56c4de0e36502ff15fd/packages/pds/src/handle/index.ts#L89)

As I'm new to this codebase I'm not sure what this "serviceHandle" even is but the boolean for it is hardcoded to "true" during signup https://github.com/bluesky-social/social-app/blob/355c50fc0fe97feb8b4ec4e29d47b725252088c7/src/screens/Signup/StepHandle.tsx#L100

So essentially the max limit for a handle including the domain is now 30 characters

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Build version: 1.96.6; Bundle info: 5090426 (prod); Bundle date: 25012002; Platform: web; Platform version: undefined

mozzius:
Service handles are the ones provided by your PDS (for example, .bsky.social). The limit has always been 30 characters for that, including the full domain (which, for bsky.social, leaves 18 for the front part).

The additional frontend checks should not have actually changed anything, because if you try and register a 19 char bsky.social username you get a server error.

You can still use a custom domain, which has a 255 character limit (iirc)

drallgood:
@mozzius That isn't quite correct. I was able to register "retrorewatch.bsky.techhangout.social" a few weeks ago (subsequently changed to a domain but that doesn't matter)

And I was able to register my preferred handle (that is longer than the 30 character limit you reference!) via the pdsadmin tool (so via the API)

See also the code references I posted above. The server only enforces a 18 character limit for the front part

mozzius:
You were totally right! My bad, thanks for correcting me.

Fix here: https://github.com/bluesky-social/social-app/pull/7657

PR #7657, opened 2025-02-05
Improved service handle validation logic

mozzius:
fixes https://github.com/bluesky-social/social-app/issues/7574

I was incorrect in my understanding of how service handles work. The logic is:

1. Take the first domain segment
2. Check if it's >=3 and <=18

There is no *overall* length limit. The utility function has been updated to reflect this.

I also renamed the helper function to better reflect what it does, and removed the extraneous boolean flag which ended up always being used in all its callsites anyway.

#Test plan

If you wanna go all the way with testing this one:
1. Get a 3rd party PDS with a domain name longer than `bsky.social`
2. Change your handle to a name 18 characters long. The server will allow you to do it, but without this PR the frontend would've stopped you

If you don't wanna test I'm pretty sure a spot check of the code is fine


mozzius:
CI is broken - webpack one is using the initial commit, not the first one

estrattonbailey (approved):
Nice one! Extra credit could be adding a `lib/strings/__tests__/handles.test.ts` suite, or adding to the existing tests, but not a huge deal.

PR #7664, opened 2025-02-05
[APP-1013] Configure and apply default post interaction settings from user preferences

estrattonbailey:
This PR adds a new screen, accessible via the Moderation screen, that allows users to set default post interaction settings (postgate and threadgate settings). These settings are then used to pre-fill the interaction settings dialog in the composer.

The new screen actually uses the same components as the dialog itself. To accomplish this, I had to break out the `Header` from the `Form` component, but other changes were not necessary to the interaction settings form itself.

The new screen is pretty straightforward, basically just the form and a call to set these values in user preferences.

### Testing
1. Out of the gate, test that the interaction settings dialog works as normal. No defaults to pre-fill, works as before.
2. Then go set some default settings, and test that those are reflected in the composer

CleanShot 2025-02-05 at 15 48 57@2x
CleanShot 2025-02-05 at 15 49 08@2x

CleanShot 2025-02-05 at 17 23 23@2x
CleanShot 2025-02-05 at 17 28 47@2x

### DIFF
--- a/src/components/dialogs/PostInteractionSettingsDialog.tsx
+++ b/src/components/dialogs/PostInteractionSettingsDialog.tsx
@@ -7,7 +7,6 @@ import {useQueryClient} from '@tanstack/react-query'
 import isEqual from 'lodash.isequal'
 
 import {logger} from '#/logger'
-import {isWeb} from '#/platform/detection'
 import {STALE} from '#/state/queries'
 import {useMyListsQuery} from '#/state/queries/my-lists'
 import {
@@ -31,14 +30,12 @@ import {
 import {useAgent, useSession} from '#/state/session'
 import * as Toast from '#/view/com/util/Toast'
 import {atoms as a, useTheme} from '#/alf'
-import {Admonition} from '#/components/Admonition'
 import {Button, ButtonIcon, ButtonText} from '#/components/Button'
 import * as Dialog from '#/components/Dialog'
 import {Divider} from '#/components/Divider'
 import * as Toggle from '#/components/forms/Toggle'
 import {Check_Stroke2_Corner0_Rounded as Check} from '#/components/icons/Check'
 import {CircleInfo_Stroke2_Corner0_Rounded as CircleInfo} from '#/components/icons/CircleInfo'
-import {InlineLinkText} from '#/components/Link'
 import {Loader} from '#/components/Loader'
 import {Text} from '#/components/Typography'
 
@@ -62,6 +59,7 @@ export function PostInteractionSettingsControlledDialog({
 }: PostInteractionSettingsFormProps & {
   control: Dialog.DialogControlProps
 }) {
+  const t = useTheme()
   const {_} = useLingui()
 
   return (
@@ -73,25 +71,20 @@ export function PostInteractionSettingsControlledDialog({
         <View style={[a.gap_md]}>
           <Header />
           <PostInteractionSettingsForm {...rest} />
-          <Admonition type="tip" style={[a.mt_sm]}>
-            {isWeb ? (
-              <Trans>
-                You can configure defaults for these settings from within your{' '}
-                <InlineLinkText
-                  label={_(msg`Configure default interaction settings`)}
-                  to="/moderation/interaction-settings">
-                  moderation settings screen.
-                </InlineLinkText>
-              </Trans>
-            ) : (
-              <Trans>
-                You can configure defaults for these settings from within{' '}
-                <Text style={[a.font_bold]}>
-                  Settings &rarr; Moderation &rarr; Interaction settings
-                </Text>
-              </Trans>
-            )}
-          </Admonition>
+          <Text
+            style={[
+              a.pt_sm,
+              a.text_sm,
+              a.leading_snug,
+              t.atoms.text_contrast_medium,
+            ]}>
+            <Trans>
+              You can configure defaults for these settings in{' '}
+              <Text style={[a.font_bold, t.atoms.text_contrast_medium]}>
+                Settings &rarr; Moderation &rarr; Interaction settings.
+              </Text>
+            </Trans>
+          </Text>
         </View>
         <Dialog.Close />
       </Dialog.ScrollableInner>
--- a/src/components/dialogs/PostInteractionSettingsDialog.tsx
+++ b/src/components/dialogs/PostInteractionSettingsDialog.tsx
@@ -30,16 +30,19 @@ import {
 import {useAgent, useSession} from '#/state/session'
 import * as Toast from '#/view/com/util/Toast'
 import {atoms as a, useTheme} from '#/alf'
+import {Admonition} from '#/components/Admonition'
 import {Button, ButtonIcon, ButtonText} from '#/components/Button'
 import * as Dialog from '#/components/Dialog'
 import {Divider} from '#/components/Divider'
 import * as Toggle from '#/components/forms/Toggle'
 import {Check_Stroke2_Corner0_Rounded as Check} from '#/components/icons/Check'
 import {CircleInfo_Stroke2_Corner0_Rounded as CircleInfo} from '#/components/icons/CircleInfo'
+import {InlineLinkText} from '#/components/Link'
 import {Loader} from '#/components/Loader'
 import {Text} from '#/components/Typography'
 
 export type PostInteractionSettingsFormProps = {
+  canSave?: boolean
   onSave: () => void
   isSaving?: boolean
 
@@ -59,6 +62,7 @@ export function PostInteractionSettingsControlledDialog({
   control: Dialog.DialogControlProps
 }) {
   const {_} = useLingui()
+
   return (
     <Dialog.Outer control={control}>
       <Dialog.Handle />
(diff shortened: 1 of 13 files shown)


src/screens/Moderation/index.tsx
```diff
@@ -199,6 +200,21 @@ export function ModerationScreenInner({
           a.overflow_hidden,
           t.atoms.bg_contrast_25,
         ]}>
+        <Link
+          label={_(msg`View your default post interaction settings`)}
+          testID="interactionSettingsBtn"
+          to="/moderation/interaction-settings">
+          {state => (
+            <SubItem
+              title={_(msg`Interaction Settings`)}
```
surfdude29, line 209:
```suggestion
              title={_(msg`Interaction settings`)}
```
Hopefully I'm not being rude by giving feedback on a WIP PR, I had a look on the preview branch and I think it would fit in better with the other options with a lowercase `s`:

IMG_9475

src/screens/ModerationInteractionSettings/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+          </Layout.Header.TitleText>
+        </Layout.Header.Content>
+      </Layout.Header.Outer>
+
+      <Layout.Content>
+        <View style={[gutters, a.gap_xl]}>
+          <Admonition type="tip">
+            <Trans>
+              The following settings will be applied automatically to all new
+              posts you create.
```
surfdude29, line 44: I think it might be helpful to have a bit more explanation here (similar to in the screenshots).

What about adding something like:

`You can choose different interaction settings for a specific post from the composer.`

estrattonbailey: I dig it! Edited slightly to fit better on two lines. Lmk what you think 👍

surfdude29: Looks great 👌

pfrazee:
Great! One copy modification I'd suggest: in the settings screen

> The following settings will be used as your defaults when creating new posts. You can edit these for a specific post from the composer.

WDYT we change to:

> These settings will apply to new posts, and will not affect your existing posts.

surfdude29:
Throwing in two more cents even though I've already given an unsolicited opinion... 😅

I think this is a situation where a little extra ELI5 explanation could be very helpful for some users. So even though it's wordy, I think an extra sentence (tweaked from my earlier suggestion) would be beneficial here:

> These settings will apply to new posts, and will not affect your existing posts. You can always choose different settings for a post from the composer.

surfdude29:
One other point I just noticed in testing: on mobile web it's possible to lose the text of the post you're typing if you open **Post interaction settings** and then tap the link to the moderation settings screen. I'm not able to test on desktop web; I assume there it opens in a new tab and so it's not an issue?

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -59,19 +63,56 @@ export function PostInteractionSettingsControlledDialog({
...
+                </InlineLinkText>
+              </Trans>
+            ) : (
+              <Trans>
+                You can configure defaults for these settings from within{' '}
+                <Text style={[a.font_bold]}>
+                  Settings &rarr; Moderation &rarr; Interaction settings
+                </Text>
+              </Trans>
+            )}
```
estrattonbailey, line 93: Want to call out that this fork was because of issues closing the nested dialogs and navigating on mobile, hence the description of the navigation and not the nav itself.

But as @surfdude29 brings up (really good point btw, thank you), we don't have any good ways currently to _return_ to the composer, or otherwise save your post in progress. So taking the users out of this flow may be more confusing and frustrating than it's worth.

We could opt for what is the native option here, but for all platforms 🤔

estrattonbailey:
@surfdude29 hear you on the copy, I like your suggestion fwiw. Gonna let Sam weigh in when he's back online and we'll make a call in the morning on these couple things that you brought up. Appreciate your thoughts!

surfdude29:
@pfrazee I apologise for the cheeky tag here on something related to this feature but not directly about this PR; I have an idea for the @bsky.app release announcement post detailing this feature and I thought this would be a good place to mention it to you :)

What about including a direct link to the Post Interaction Settings screen as a facet? Users could then instantly access the new settings directly from the @bsky.app post where the feature is being explained to them. Also, in the future that post could then be quoted in reply to users for whom it might be helpful (e.g. "Did you know that you can set a default for all your posts to limit replies to only people you follow? This means you don't have to remember to do it each time, here's how") And instead of describing how to navigate there in Settings they would just need to tap/click the link 💫

Essentially, it would be a mini support article cunningly disguised as a @bsky.app post. And it would become even more accessible and easy to find when #7622 is merged and searching @bsky.app's posts is more intuitive.

E.g.:

> You can now set defaults for who can reply to and quote your posts.
>
> Keep in mind:
> - These settings will only apply to new posts and won't affect your existing posts
> - You can always adjust these settings for a single post from the composer
>
> Change your default post interaction settings [here](https://bsky.app/moderation/interaction-settings)

src/components/dialogs/PostInteractionSettingsDialog.tsx
mozzius: I vote going with how native does it, since we don't have drafts. It's fine, and linking out feels broken

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -59,19 +63,56 @@ export function PostInteractionSettingsControlledDialog({
...
     <Dialog.Outer control={control}>
       <Dialog.Handle />
       <Dialog.ScrollableInner
         label={_(msg`Edit post interaction settings`)}
         style={[{maxWidth: 500}, a.w_full]}>
-        <PostInteractionSettingsForm {...rest} />
+        <View style={[a.gap_md]}>
+          <Header />
+          <PostInteractionSettingsForm {...rest} />
+          <Admonition type="tip" style={[a.mt_sm]}>
```
mozzius, line 76: Can we have the option to hide this admonition, either via Nudge NUX, or just hiding it if you have a non-empty setting. Ideally both

mozzius (changes requested):
Couple notes but mostly great!

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -59,19 +63,56 @@ export function PostInteractionSettingsControlledDialog({
...
+                You can configure defaults for these settings from within your{' '}
+                <InlineLinkText
+                  label={_(msg`Configure default interaction settings`)}
+                  to="/moderation/interaction-settings">
+                  moderation settings screen.
+                </InlineLinkText>
+              </Trans>
+            ) : (
+              <Trans>
+                You can configure defaults for these settings from within{' '}
```
mozzius, line 88: Suggestion for copy change: `You can configure defaults for these settings from within Settings > Moderation > Interaction settings` to `You can set default options for these settings in Settings > Moderation > Interaction settings`

src/screens/ModerationInteractionSettings/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  return (
+    <Layout.Screen testID="ModerationInteractionSettingsScreen">
+      <Layout.Header.Outer>
+        <Layout.Header.BackButton />
+        <Layout.Header.Content>
+          <Layout.Header.TitleText>
+            <Trans>Post Interaction Settings</Trans>
+          </Layout.Header.TitleText>
+        </Layout.Header.Content>
+      </Layout.Header.Outer>
```
mozzius, line 37: Missing a `Layout.Header.Slot`!

mozzius (changes requested):

src/components/dialogs/PostInteractionSettingsDialog.tsx
surfdude29: The best of both maybe?

`You can set defaults for these settings in Settings > Moderation > Interaction settings`

src/components/dialogs/PostInteractionSettingsDialog.tsx
estrattonbailey: We'd have to do a nux, or just store on device, can't infer from prefs since they can be reset. I wonder if it's helpful enough that we should integrate it permanently but make it more subtle 🤔

src/screens/ModerationInteractionSettings/index.tsx
estrattonbailey: Oh wow I didn't even notice in the screenshot, thanks

src/components/dialogs/PostInteractionSettingsDialog.tsx
surfdude29: I deleted an earlier comment because I thought I might be unhelpfully bikeshedding, but fwiw another option might be:

`You can choose default interaction settings in Settings > Moderation > Interaction settings`

mozzius (approved):
lgtm! I'm ok with ignoring the lockfile edit warning, diff looks fine

pfrazee:
> @pfrazee I apologise for the cheeky tag here on something related to this feature but not directly about this PR; I have an idea for the @bsky.app release announcement post detailing this feature and I thought this would be a good place to mention it to you :)
> 
> What about including a direct link to the Post Interaction Settings screen as a facet? Users could then instantly access the new settings directly from the @bsky.app post where the feature is being explained to them. Also, in the future that post could then be quoted in reply to users for whom it might be helpful (e.g. "Did you know that you can set a default for all your posts to limit replies to only people you follow? This means you don't have to remember to do it each time, here's how") And instead of describing how to navigate there in Settings they would just need to tap/click the link 💫
> 
> Essentially, it would be a mini support article cunningly disguised as a @bsky.app post. And it would become even more accessible and easy to find when #7622 is merged and searching @bsky.app's posts is more intuitive.
> 
> E.g.:
> 
> > You can now set defaults for who can reply to and quote your posts.
> > Keep in mind:
> > 
> > * These settings will only apply to new posts and won't affect your existing posts
> > * You can always adjust these settings for a single post from the composer
> > 
> > Change your default post interaction settings [here](https://bsky.app/moderation/interaction-settings)

@surfdude29 Good suggestion! We'll give that a look

Issue #7443, opened 2025-01-13
Link cards should respect og:width and og:height for preview images, for visual reasons


JonFabritius:
### Describe the Feature

The automatically generated image preview in a link card currently renders a landscape format image regardless of source aspect ratio, resulting in cropping when the og:preview has been set to a square image, for example.

For photographers or illustrators sharing content from their own source URIs, this results in unintended visual presentation of images and looks broken.

Using the og:width and og:height metadata, or calculating the correct aspect ratio from the image file would solve this, resulting in a cleaner overall feed view.

mschwendener:
> looks broken

Yes!

https://bsky.app/profile/dianetucker.bsky.social/post/3loarxwwd7k26 

<img width="596" alt="Image" src="https://github.com/user-attachments/assets/56fde71d-7463-45cd-bbda-1b249611a921" />

mschwendener:
https://bsky.app/profile/artherstory.bsky.social/post/3m3fac4fka22p 

<img width="583" height="616" alt="Image" src="https://github.com/user-attachments/assets/b2e96c7a-f31e-4fd7-a981-3254ed2e6d5c" />

mschwendener:
Could this please be assigned to somebody?

mschwendener:
This https://github.com/bluesky-social/social-app/issues/1118 had at least gotten a "discussing" tag, but somebody closed the thread and I can't reopen it. 
  
 
<img width="595" height="494" alt="Image" src="https://github.com/user-attachments/assets/a3518412-ba45-4567-bb22-58cafb88b942" />

mschwendener:
@abenzer  

Examples of ugly crops, related to https://bsky.app/profile/alexbenzer.com/post/3mlvrfztuvu25 "why is post editing important to you?"

mschwendener:
See also here for examples https://github.com/bluesky-social/social-app/issues/5012

PR #7680, opened 2025-02-06
Enable dynamic aspect ratio for embed images

StephaneB1:
For some embed images, the default aspect ratio doesn't work and crops the preview image. In this PR I am setting the aspect ratio between a min-max value that prevents crazy aspect ratios to ruin the users' feed

https://github.com/bluesky-social/social-app/issues/7443

### Before
image

### After
Screenshot 2025-02-06 at 16 26 37


mozzius:
We'd really like to avoid dynamically changing the post height on render, it leads to degraded scroll performance and scroll jumps/jitters

I'd definitely be open to adding an aspectRatio field to external embeds like we did here for images: https://github.com/bluesky-social/atproto/pull/1306

PR #7681, opened 2025-02-06
[APP-1031] Add new followerRule to threadgate settings

estrattonbailey:
Adds `followersRule` support to interaction settings. App view and dataplane are already deployed and working, so all that's left is the UI.

CleanShot 2025-02-06 at 21 53 55@2x
CleanShot 2025-02-06 at 21 53 10@2x
CleanShot 2025-02-06 at 21 48 40@2x
CleanShot 2025-02-06 at 21 40 30@2x

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "icons:optimize": "svgo -f ./assets/icons"
   },
   "dependencies": {
-    "@atproto/api": "^0.13.33",
+    "@atproto/api": "^0.13.34",
     "@bitdrift/react-native": "^0.6.2",
     "@braintree/sanitize-url": "^6.0.2",
     "@discord/bottom-sheet": "bluesky-social/react-native-bottom-sheet",
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "icons:optimize": "svgo -f ./assets/icons"
   },
   "dependencies": {
-    "@atproto/api": "^0.13.34",
+    "@atproto/api": "^0.13.35",
     "@bitdrift/react-native": "^0.6.2",
     "@braintree/sanitize-url": "^6.0.2",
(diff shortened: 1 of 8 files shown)


mozzius (approved):
Awesome, works perfectly. Filed #7692 separately.

My only suggestion is maybe disable this enormous CTA when the threadgate is enabled? What do you think? It's probably fine to keep if you disagree

Screenshot 2025-02-07 at 22 51 08

pfrazee:
Yeah actually let's remove that in that case

estrattonbailey:
Discussed in Slack and decided to err on the side of increasing friction for following a user and then replying, if indeed the `followedBy` rule is in place.

@mozzius noting that since we show placeholder data from the cache here and get the `threadgate` from `getPostThread`, there is a slight flash of the Follow button rendering and then hiding.

I felt this was better than the alternative of not showing the Follow button and then showing it once the thread fully loaded, since that will affect the vast majority of posts, and the small regression here only affects posts with this threadgate enabled.

estrattonbailey:
Also noting that initially I had reverted the profile shadow use that was in place to ensure the reply button responded to a follow from the post thread. But I reverted that again so that post -> profile -> follow -> back to post enables replies. I think this would be confusing otherwise.

surfdude29:
Sorry for jumping in here with one small point that just occurred to me: if you view a post that has the `followedBy` threadgate set on a feed (e.g. Discover or on someone's profile) on native and you don't follow the author, would the follow option be unavailable when tapping on the author's avatar, too? It's not quite as big and obvious a CTA, of course.

mozzius:
I don't mind it being present on the hover popover etc - I think that's fine. the big CTA was enough

mozzius (approved):
Button flash is fine, not too bad

PR #7686, opened 2025-02-06
Post moderation menu

mozzius:


src/components/post-ctrls/PostModerationBtnMenuItems.tsx
```diff
@@ -0,0 +1,578 @@
...
+        description={_(
+          msg`This reply will be sorted into a hidden section at the bottom of your thread and will mute notifications for subsequent replies - both for yourself and others.`,
+        )}
+        onConfirm={onToggleReplyVisibility}
+        confirmButtonCta={_(msg`Yes, hide`)}
+      />
+
+      <Prompt.Basic
+        control={blockPromptControl}
+        title={_(msg`Block Account?`)}
```
surfdude29, line 566: Tiny suggested tweak here: the other prompt titles seem to be in sentence case, so what about `Block account?` here instead?

PR #7690, opened 2025-02-07
Unmark string for translation

surfdude29:
There's a string in Settings which is currently marked for translation but it consists only of the user's handle:

https://github.com/bluesky-social/social-app/blob/b12d39b4cff0fe1b91fb16496c3fea1ed22de5cf/src/screens/Settings/Settings.tsx#L423-L425

And thus it contains no translatable text:

https://github.com/bluesky-social/social-app/blob/b12d39b4cff0fe1b91fb16496c3fea1ed22de5cf/src/locale/locales/en/messages.po#L103-L105

So this PR removes the translation tags.

### DIFF
--- a/src/screens/Settings/Settings.tsx
+++ b/src/screens/Settings/Settings.tsx
@@ -421,7 +421,7 @@ function AccountRow({
           <View style={[{width: 28}]} />
         )}
         <SettingsList.ItemText>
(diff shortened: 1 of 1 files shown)


mozzius (approved):
nice, thanks

PR #7694, opened 2025-02-08
Fix quote & list padding not being pressable

mozzius:
Quote & list embeds have a problem with their padding, in that the pressable link area is contained within the padding, meaning the padding area is not clickable. Easiest to understand by just looking at the before/after of the pressable area

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

Ensure padding area is now clickable
Ensure styles for these elements is otherwise the same
Here's a thread of different embeds: https://bsky.app/profile/samuel.bsky.team/post/3lhmznav6j22d

### DIFF
--- a/src/view/com/util/post-embeds/index.tsx
+++ b/src/view/com/util/post-embeds/index.tsx
@@ -1,11 +1,5 @@
 import React from 'react'
-import {
-  InteractionManager,
-  StyleProp,
-  StyleSheet,
-  View,
-  ViewStyle,
-} from 'react-native'
+import {InteractionManager, StyleProp, View, ViewStyle} from 'react-native'
 import {MeasuredDimensions, runOnJS, runOnUI} from 'react-native-reanimated'
 import {Image} from 'expo-image'
 import {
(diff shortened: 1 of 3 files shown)


src/view/com/util/post-embeds/index.tsx
```diff
@@ -267,7 +260,7 @@ export function MaybeFeedCard({view}: {view: AppBskyFeedDefs.GeneratorView}) {
     <ContentHider modui={moderation?.ui('contentList')}>
       <FeedSourceCard
         feedUri={view.uri}
-        style={[pal.view, pal.border, styles.customFeedOuter]}
+        style={[a.border, t.atoms.border_contrast_medium, a.p_md, a.rounded_sm]}
```
mozzius, line 263: I don't think the `pal.view` was needed, correct me if I'm wrong. We need to avoid unnecessary android background colors for perf

haileyok: Yea let's get rid of that. We used it for background colors in the past, but otherwise unnecessary.

haileyok (approved):

PR #7701, opened 2025-02-08
Clicking "Post Sent" toast on web app will open the post

plasma-node:
This is QoL and mimics twitter behavior (for web) instead of having to go Sidebar => Profile => Scroll => Click tweet.

Text not adjusted, unsure how to handle localization.

Imo best solution would be parity between mobile & web with toasts appearing center top/bottom of screen, like twitter.


mozzius:
Superceded by #8878

PR #7706, opened 2025-02-08
Special treatment for recommended starter packs

mozzius:
Given that the content will (for the time being) usually mostly be all feeds, we should give starter packs a special treatment using the starter pack icon

I also changed the wording of the subheader when there's non-feed content in there

Screenshot 2025-02-08 at 23 38 33


estrattonbailey:
Yep nice, glad we're finally using the parsing logic haha

PR #7708, opened 2025-02-09
Update the ContentFilter setting for tenor gif search

jetpacktuxedo:
No other media "upload" (it's... Not really an upload here I guess, but still) limits content at post time, it is instead restricted at read-time by the client/viewer.

Per the [tenor docs](https://tenor.com/gifapi/documentation#contentfilter) the current "high" setting restricts it to the equivalent of "G-rated" gifs which quite frankly seems absurdly restrictive considering the other content that is permitted. This change as currently written bumps it to pg-13. 

I personally would just turn it "off" ("R-rated") but thought that was less likely to be accepted.

### DIFF
--- a/src/state/queries/tenor.ts
+++ b/src/state/queries/tenor.ts
@@ -54,7 +54,7 @@ function createTenorApi<Input extends object>(
     // 30 is divisible by 2 and 3, so both 2 and 3 column layouts can be used
     params.set('limit', '30')
 
-    params.set('contentfilter', 'high')
+    params.set('contentfilter', 'low')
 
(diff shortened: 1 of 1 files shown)


surfdude29:
FYI there is a similar PR from @mozzius at #5250.

jetpacktuxedo:
Oh I hadn't seen that, good catch! I'll leave this for now just in case the complexity in that one is seen as a barrier... I'm not sure I've seen any other platform that uses tenor ask me as an end user how I'd like to filter the available gifs, but it's a neat approach at least.

PR #7704, opened 2025-02-08
Animate dropdown menus

mozzius:
# Stacked on https://github.com/bluesky-social/social-app/pull/7703

Animates dropdown menus with a fade/zoom, with the origin being towards the direction of the trigger. Works in both directions thanks to a special radix css variable. Respects prefers-reduced-motion

https://github.com/user-attachments/assets/6caa5372-6772-4fee-95f5-0e5fa66d5da4

PR #7711, opened 2025-02-10
Animate drawer menu on mobile web

mozzius:
# Based on #7704 

Animates both ways. Animating in uses an exponential easing function

https://github.com/user-attachments/assets/a027ee65-3612-400d-8084-19ef7b537e20

### DIFF
--- a/src/view/shell/index.web.tsx
+++ b/src/view/shell/index.web.tsx
@@ -142,5 +142,7 @@ const styles = StyleSheet.create({
     top: 0,
     left: 0,
     height: '100%',
+    width: 330,
+    maxWidth: '80%',
   },
 })
--- a/src/view/shell/index.web.tsx
+++ b/src/view/shell/index.web.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect} from 'react'
+import {useEffect, useLayoutEffect, useState} from 'react'
 import {StyleSheet, TouchableWithoutFeedback, View} from 'react-native'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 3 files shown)


src/view/shell/index.web.tsx
```diff
@@ -66,20 +80,27 @@ function ShellInner() {
...
-                    dark: 'rgba(1, 82, 168, 0.1)',
-                    dim: 'rgba(10, 13, 16, 0.8)',
-                  }),
+                  backgroundColor: showDrawer
+                    ? select(t.name, {
+                        light: 'rgba(0, 57, 117, 0.1)',
+                        dark: 'rgba(1, 82, 168, 0.1)',
+                        dim: 'rgba(10, 13, 16, 0.8)',
+                      })
+                    : 'transparent',
```
estrattonbailey, line 95: I'm seeing the fade out, but not fade in. Maybe it doesn't like the transition from transparent to a defined color? Not sure tbh, css transitions are finicky with newly rendered content iirc

estrattonbailey (approved):
Nice, just one nit if you're able to fix it

src/view/shell/index.web.tsx
mozzius: haha I was hoping you wouldn't spot it - yeah, it's a transition rather than an animation, so the nothing -> color doesn't animate. Fixing this would require messing with the HTML structure which I was hoping to avoid

estrattonbailey: Hehe. Does it not work going from `rgba(0,0,0,0)` to the rgba values defined for the open state?

mozzius: I guess we'd need to render it once with that value, then on next tick change it again. totally doable but very messy!

estrattonbailey: Ok yeah that's what I was afraid of. LGTM, still net positive 👍

mozzius: We can come back to it!

Issue #6964, opened 2024-12-05
Add "Mute account" in pop-up menu


Shohreh:
### Describe the Feature

I didn't find [this suggestion](is:open label:feature-request mute account) in the list.

When right-clicking on an account in a thread, it shows "Block account", but no "Mute account".

Cheers,

![image](https://github.com/user-attachments/assets/269d92bf-e411-4bb7-8e45-df82ad7e6d8e)

Shohreh:
No. It's not listed in a _thread_, forcing the user to click more to mute the user by going through their profile.

If the menu has a "Block account", it should logically also have a "Mute account".

surfdude29:
Yes, that's what the issue I linked is about? 🤔

PR #7712, opened 2025-02-10
Mute account from post

### LINKED ISSUE #1017: add mute user button to posts

mozzius:
# Needs #7798 

Fixes #1017 and #6964.

Self-explanatory. Same as blocking.

# Test plan

Test muting and unmuting

### DIFF
--- a/src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
+++ b/src/view/com/util/forms/PostDropdownBtnMenuItems.tsx
@@ -41,7 +41,10 @@ import {
 } from '#/state/queries/post'
 import {useToggleQuoteDetachmentMutation} from '#/state/queries/postgate'
 import {getMaybeDetachedQuoteEmbed} from '#/state/queries/postgate/util'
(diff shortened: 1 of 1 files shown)


surfdude29:
This would be great to have 👍

It works fine in my testing, although I noticed one small thing I wanted to mention: with this PR, the order of **Block/Mute/Report** is different in the post context menu as compared to the account context menu. Not a big thing, but it might make sense to keep the order the same in both (whichever is considered better).

| Account menu | Post menu |
|--------|--------|
| IMG_2407 | IMG_2375 |

surfdude29:
Nice 👌

Also, I forgot to add, this will fix #1017 and #6964.

mozzius:
Nice, thank you :)

haileyok (approved):

PR #7723, opened 2025-02-11
Add dev mode for easy copying of at:// URIs and DIDs

tom-sherman:
Enabled by long pressing on the version number in the settings.


src/view/com/profile/ProfileMenu.tsx
```diff
@@ -167,6 +169,14 @@ let ProfileMenu = ({
     reportDialogControl.open()
   }, [reportDialogControl])
 
+  const onPressShareATUri = React.useCallback(() => {
+    shareUrl(`at://${profile.did}`)
+  }, [profile.did])
+
+  const onPressShareDID = React.useCallback(() => {
+    shareUrl(profile.did)
```
tom-sherman, line 177: I should probably just set the string to clipboard directly in these instances to be honest considering these aren't URLs and in some cases (like this one) not even URIs

tom-sherman:
This is an alternative to #6866, I think they may almost be identical actually

src/state/persisted/schema.ts
```diff
@@ -123,6 +123,7 @@ const schema = z.object({
   kawaii: z.boolean().optional(),
   hasCheckedForStarterPack: z.boolean().optional(),
   subtitlesEnabled: z.boolean().optional(),
+  devMode: z.boolean().optional(),
```
mozzius, line 126: Minor, but can we use the new `#/storage` API rather than `persisted` please?

mozzius (changes requested):

src/state/persisted/schema.ts
tom-sherman: Done. I went ahead and scoped it to the device, encapsulated that decision into a `useDevModeEnabled` hook that returns both the state and setter.

mozzius: Nice! Could you remove `devMode` from the persisted schema and I think it's good to go

src/screens/Settings/AboutSettings.tsx
```diff
@@ -66,6 +68,17 @@ export function AboutSettingsScreen({}: Props) {
...
             label={_(msg`Version ${appVersion}`)}
             accessibilityHint={_(msg`Copies build version to clipboard`)}
+            onLongPress={() => {
+              const newDevModeEnabled = !devModeEnabled
+              setDevModeEnabled(newDevModeEnabled)
+              Toast.show(
+                _(
+                  msg`Developer mode ${
+                    newDevModeEnabled ? 'enabled' : 'disabled'
+                  }`,
```
mozzius, line 78: Not sure the translation extraction is smart enough for this, probably best to simplify to this:
```suggestion
                newDevModeEnabled
                  ? _(msg`Developer mode enabled`)
                  : _(msg`Developer mode disabled`)
```

src/state/persisted/schema.ts
tom-sherman: Oops! Done

mozzius (approved):
Awesome stuff!

PR #7725, opened 2025-02-11
deps: Bumps bitdrift to 0.6.8 & fixes iOS network instrumentation bug

Maker-Mark:
This PR supersedes https://github.com/bluesky-social/social-app/pull/7575, the PR that added network instrumentation. 
It's important this PR is merged before adding `enableNetworkInstrumentation:true` to the bitdrift .init config since versions before 0.6.8 had a permission issue that caused the app to crash when backgrounded.

It may make sense to add the value of `enableNetworkInstrumentation` behind a feature flag. I'm going to leave that up to the maintainers if they want to enable the instrumentation, and if it should be feature flagged.

Take care :—)

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
   },
   "dependencies": {
     "@atproto/api": "^0.13.35",
-    "@bitdrift/react-native": "^0.6.2",
+    "@bitdrift/react-native": "^0.6.8",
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Thank you! Gonna merge into a base and build some tests from there 👍

PR #7728, opened 2025-02-12
Bitdrift integration

estrattonbailey:
Includes two PRs relating to version bumps, which puts us at the latest version.

This PR then integrates two new feature flags:
- `enable_bitdrift_v2` which replaces the initial flag
- `enable_bitdrift_v2_networking` which conditionally enables network logging (which caused a bug previously)

Decided to replace the initial flag so that as we ramp this new version, we don't ramp older clients that are still using the old version of the Bitdrift library.

Tested on TF and APK. Web should have no changes.

### DIFF
--- a/src/screens/Settings/AboutSettings.tsx
+++ b/src/screens/Settings/AboutSettings.tsx
@@ -1,8 +1,10 @@
+import {useMemo} from 'react'
 import {Platform} from 'react-native'
 import {setStringAsync} from 'expo-clipboard'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {NativeStackScreenProps} from '@react-navigation/native-stack'
+import {Statsig} from 'statsig-react-native-expo'
 
 import {appVersion, BUNDLE_DATE, bundleInfo} from '#/lib/app-info'
 import {STATUS_PAGE_URL} from '#/lib/constants'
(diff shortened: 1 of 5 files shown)


src/screens/Settings/AboutSettings.tsx
```diff
@@ -78,6 +81,21 @@ export function AboutSettingsScreen({}: Props) {
...
+            <SettingsList.PressableItem
+              label={_(msg`StatSig stable ID`)}
+              accessibilityHint={_(msg`Copies StatSig stable ID to clipboard`)}
+              onPress={() => {
+                setStringAsync(stableId)
+                Toast.show(_(msg`Copied user ID to clipboard`))
+              }}>
+              <SettingsList.ItemIcon icon={WrenchIcon} />
+              <SettingsList.ItemText numberOfLines={1}>
+                <Trans>{stableId}</Trans>
```
surfdude29, line 94: I'm not sure this line needs to be wrapped in `<Trans>` tags, as there doesn't seem to be any translatable text?

Also, I hope I'm not being undiplomatic in saying this, but I was wondering: is it necessary to mark for translation text strings that are gated behind `IS_INTERNAL`? As you've probably noticed, usually I'm a big fan of making sure that allll the text is translatable 😅

But I'm wondering if it makes sense to ask translators to translate these three extra strings when all the Bluesky team members seem to communicate in excellent English :)

estrattonbailey: Yeah this was quick copy pasta. This is still a draft PR, I'm not currently planning on shipping this new button, just added for debugging.

surfdude29: Oops, got it 👍

mozzius (approved):
Not up to date on bitdrift so idk what I'm testing, but it builds np so lgtm

PR #7729, opened 2025-02-12
Logger improvements

estrattonbailey:


src/logger/transports/bitdrift.ts
```diff
@@ -0,0 +1,26 @@
...
+} as const
+
+export const bitdriftTransport: Transport = (
+  level,
+  context,
+  message,
+  metadata,
+) => {
+  const log = logFunctions[level]
+  log(`(${context}) ${message}`, prepareMetadata(metadata))
```
murki, line 25: (hi from bitdrift 👋 ) not sure what context is but it could also be a good idea to also pass that as a field (aka metadata) so bitdrift can then group by/filter by that value /passerby comment

estrattonbailey: Thanks! Yeah was considering that also. Initially I wanted the logs to match our dev-only console transport, but that's not really a big deal and it will def work better with Bitdrift. Sentry I'm less sure on, but not too worried about Sentry right now.

haileyok (approved):

PR #7731, opened 2025-02-13
Prioritise language codes with regions

mozzius:
1. Prioritises language codes with regions (Chinese, Brazilian Portuguese)
2. More language name cleanups


surfdude29:
Nice, fixed on the preview branch for me 👌

Not a big deal, but for some reason Assamese is now showing up, even though we don't (yet) have that as an app language? 😅

mozzius:
ah dangit

estrattonbailey (approved):

PR #7735, opened 2025-02-13
Reenable router events

mozzius:
# DO NOT MERGE UNTIL MIDDLEMAN IS READY

Reenables `router:navigate`, except not sent to statsig using `lake:` prefix

Sends the previous and next route.

# Test plan

Replace `logEvent` with `console.log`, confirm it looks good


estrattonbailey (approved):

mozzius:
Realised `logEvent` adds a `routeName` field to all events anyway, so we don't need the `to` field

haileyok:
ready to go!

PR #7738, opened 2025-02-13
[APP-1049] show label expiration in frontend

estrattonbailey:
Show's label expiration (if set) on labels, for both the labeled user and viewers.

I also opted to align the UI a bit here from where it was, since I think the `LabelsOnMe` dialog treatment is more attractive and succinct, and works well for the expiry handling.

Also, `useGetTimeAgo` is really just a date diff, so tbh we could rename that to make it more obvious that it can be used for future dates too.

We should really build a real tooltip too for web hovers e.g. to show the precise expiry on hover.

CleanShot 2025-02-13 at 16 39 25@2x
CleanShot 2025-02-13 at 16 44 51@2x
CleanShot 2025-02-13 at 16 44 31@2x
CleanShot 2025-02-13 at 17 15 25@2x
CleanShot 2025-02-13 at 17 17 12@2x
CleanShot 2025-02-13 at 17 17 35@2x
CleanShot 2025-02-13 at 17 22 38@2x

### DIFF
--- a/src/components/moderation/ModerationDetailsDialog.tsx
+++ b/src/components/moderation/ModerationDetailsDialog.tsx
@@ -42,7 +42,7 @@ function ModerationDetailsDialogInner({
   const {_} = useLingui()
   const desc = useModerationCauseDescription(modcause)
   const {currentAccount} = useSession()
-  const timeDiff = useGetTimeAgo()
+  const timeDiff = useGetTimeAgo({future: true})
 
   let name
   let description
--- a/src/components/moderation/ModerationDetailsDialog.tsx
+++ b/src/components/moderation/ModerationDetailsDialog.tsx
@@ -169,7 +169,13 @@ function ModerationDetailsDialogInner({
             </Text>
           ) : (
             <>
-              <View style={[a.flex_row, a.justify_between, a.gap_xl]}>
+              <View
+                style={[
+                  a.flex_row,
+                  a.justify_between,
+                  a.gap_xl,
+                  {paddingBottom: 1},
+                ]}>
                 <Text
                   style={[
                     a.flex_1,
--- a/src/components/moderation/ModerationDetailsDialog.tsx
+++ b/src/components/moderation/ModerationDetailsDialog.tsx
@@ -3,14 +3,14 @@ import {ModerationCause} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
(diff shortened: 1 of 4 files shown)


src/lib/moderation/useModerationCauseDescription.ts
```diff
@@ -130,14 +132,16 @@ export function useModerationCauseDescription(
...
+      let sourceDisplayName = labeler?.creator.displayName
       if (!source) {
         if (cause.label.src === BSKY_LABELER_DID) {
-          source = 'Bluesky Moderation Service'
+          source = 'moderation.bsky.app'
+          sourceDisplayName = 'Bluesky Moderation Service'
         } else {
-          source = cause.label.src
+          source = _(msg`an unknown labeler`)
         }
```
estrattonbailey, line 145: Now prioritizes the handle over the display name

src/components/moderation/ModerationDetailsDialog.tsx
```diff
@@ -128,37 +130,82 @@ function ModerationDetailsDialogInner({
     description = ''
   }
 
+  const sourceName =
+    desc.source || desc.sourceDisplayName || _(msg`an unknown labeler`)
```
estrattonbailey, line 134: Source will always be defined based on the logic in `useModerationCauseDescription`, but did this for completeness

src/components/moderation/ModerationDetailsDialog.tsx
```diff
@@ -128,37 +130,82 @@ function ModerationDetailsDialogInner({
...
-                    label={desc.source || _(msg`an unknown labeler`)}
-                    to={makeProfileLink({did: modcause.label.src, handle: ''})}
-                    onPress={() => control.close()}
-                    style={a.text_md}>
-                    {desc.source || _(msg`an unknown labeler`)}
-                  </InlineLinkText>
-                  .
-                </Trans>
-              </Text>
+              <View style={[a.flex_row, a.justify_between, a.gap_xl]}>
```
mozzius, line 172: Visual alignment is a little off, would look better with +4px padding on the bottom

Screenshot 2025-02-18 at 13 19 38

mozzius (approved):
lgtm! one nit on visual alignment

src/components/moderation/ModerationDetailsDialog.tsx
estrattonbailey: Yeah anything more than 1px starts to look wrong to me, esp when there's not exp shown and it's just `Source: ...`. Copied this fix to `LabelsOnMe` as well since it's the same treatment there.

PR #7749, opened 2025-02-14
Consistent capitalisation on profile menu

mozzius:
Unlike all other menus, the profile menu uses Title Case. Changes it to use Sentence case ~~This is a find-and-replace to change it to Sentence case, including translations to prevent churn~~

### DIFF
--- a/src/view/com/profile/ProfileMenu.tsx
+++ b/src/view/com/profile/ProfileMenu.tsx
@@ -225,10 +225,10 @@ let ProfileMenu = ({
             </Menu.Item>
             <Menu.Item
               testID="profileHeaderDropdownSearchBtn"
-              label={_(msg`Search Posts`)}
+              label={_(msg`Search posts`)}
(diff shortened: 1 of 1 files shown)


surfdude29:
Given that translations are now managed on Crowdin which pulls the source English `messages.po` when it's updated, would updating the translations in this PR have any effect? 🤔

Fwiw I think the best strategy might be to not update any locales in this PR, and let us translators just select and approve the existing translations for the tweaked strings when they arrive on Crowdin :)

And for any locales where someone hasn't managed to approve them, I'm happy to approve the existing translations so that there's no churn (shouldn't take more than 10-15 mins tops to cover all locales).

mozzius:
Oh awesome, that's great. Crowdin ftw!

Pushed up just the source changes, no .po files

surfdude29:
Yup, Crowdin makes things so much easier all round 💫

haileyok (approved):

PR #7751, opened 2025-02-15
Make Android app start faster by disabling JS bundle compression

mrousavy:
Disables compression for the `.bundle` resource files in the .apk.

This makes the .apk a bit bigger, but improves startup time on Android - we measured a 400ms startup improvement (12%) in one of our apps.

Based on https://github.com/facebook/react-native/pull/49449


mrousavy:
Draft because I couldn't yet figure out how to make this work with Expo Config Plugins.

When running `expo config --type public`, my custom key properly appears:

```json
  android: {
    icon: './assets/app-icons/android_icon_default_light.png',
    googleServicesFile: './google-services.json',
    package: 'xyz.blueskyweb.app',
    intentFilters: [
      {
        action: 'VIEW',
        autoVerify: true,
        data: [
          {
            scheme: 'https',
            host: 'bsky.app'
          },
          {
            scheme: 'http',
            host: 'localhost:19006'
          }
        ],
        category: [
          'BROWSABLE',
          'DEFAULT'
        ]
      }
    ],
    permissions: [
      'android.permission.CAMERA',
      'android.permission.RECORD_AUDIO'
    ],
    adaptiveIcon: {
      foregroundImage: './assets/icon-android-foreground.png',
      monochromeImage: './assets/icon-android-foreground.png',
      backgroundImage: './assets/icon-android-background.png',
      backgroundColor: '#1185FE'
    },
    androidResources: {
      noCompress: [
        'bundle'
      ]
    }
```

..but it's not being added to `android/app/build.gradle`.... 🤔

mrousavy:
Okay now it works, I fixed it!

@gaearon do you guys have any TTI measurements in BlueSky? If yes, would be great if you could run those in release builds to see how much faster it got with my PR.

I can't seem to be able to build the app in release.

surfdude29:
FYI @gaearon isn't working at Bluesky anymore:

https://bsky.app/profile/danabra.mov/post/3lhcrrc76s22i

gaearon:
There's this:

https://github.com/bluesky-social/social-app/blob/40d05ac866a4783889f465458ffb5a02a169795a/src/Navigation.tsx#L860

You can get it to show up by commenting out the "remove console" plugin in the prod Babel config.

>I can't seem to be able to build the app in release.

Not sure if anything is broken, last time I tried `yarn android --variant release` was working fine.

mozzius:
Ok results are in:

**Current** (mean of 10 runs): 805ms

**With change** (mean of 10 runs): 773ms

Improvement: **-4%**

That's not nothing, and this was on a Pixel 7A. I'd imagine that lower-powered phones might see more benefit.

mozzius (approved):
🔥

mrousavy:
I haven't dug too deeply, but it looks like `initMs` is calculated from `__BUNDLE_START_TIME__`: https://github.com/bluesky-social/social-app/blob/40d05ac866a4783889f465458ffb5a02a169795a/src/Navigation.tsx#L856-L859

This sounds wrong to measure this specific improvement, since the performance gains already start a bit earlier than that; when Android starts to load the actual `.bundle` from resources (see [`_CompressedAsset::getBuffer`](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/libs/androidfw/Asset.cpp;l=903?q=Asset.cpp)).

So if you measured 4% with this change, you only measured the mmap difference. The real improvements might be much bigger because we didn't measure the uncompression part of this whole thing. Anyways; it's faster now 😄

gaearon:
Seems like you’d want to measure the time from the first native app log to when this console log comes in.

mrousavy:
yep

PR #7760, opened 2025-02-17
Add option to align web dialogs to top

estrattonbailey:
Needed for APP-1054, this PR just allows us to optionally align dialogs to the top of the screen on web. This is already helpful for the report dialog, which changes height otherwise and jumps around. 

CleanShot 2025-02-17 at 16 24 26@2x
CleanShot 2025-02-17 at 16 25 28@2x

### DIFF
--- a/src/components/Dialog/index.web.tsx
+++ b/src/components/Dialog/index.web.tsx
@@ -41,6 +41,7 @@ export function Outer({
   children,
   control,
   onClose,
+  webOptions,
 }: React.PropsWithChildren<DialogOuterProps>) {
   const {_} = useLingui()
   const {gtMobile} = useBreakpoints()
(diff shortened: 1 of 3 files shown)


mozzius (approved):
lgtm! off the top of my head, this would also be helpful for the threadgate dialog and the handle dialog

estrattonbailey:
@mozzius yeah was thinking about others but I didn't want to go too crazy with this off the bat. If we like this general idea, I actually have a better idea: just align all dialogs to the top, but give them a little breathing room on large screens so it doesn't look so cramped.

https://github.com/user-attachments/assets/50d109da-7e07-492c-8398-a38541e135c8

Not sure we even need center alignment then, since even small dialogs are relatively positioned to look nice on large screens and with other dialogs behind.

CleanShot 2025-02-18 at 09 13 15@2x

I think this is actually better, so gonna run with this, will push up in a sec.

PR #7766, opened 2025-02-18
Update i18n release process to solve missing translations

pfrazee:
In 1.98 we had some missing string translations. This was at minimum because the new release process didn't include a source-string extraction prior to cutting the build. This PR fixes the release process and also updates the `.po` files.


pfrazee:
Note that some translations got dropped with this run, but that's expected. See https://crowdin.com/project/bluesky-social/discussions/36 for info about why.

PR #7771, opened 2025-02-18
Add soft reset to chat list

mozzius:
Realised this was missing. Now works like other tabs

# Test plan

Check that chat list scrolls to top when pressing the tab

### DIFF
--- a/src/screens/Messages/ChatList.tsx
+++ b/src/screens/Messages/ChatList.tsx
@@ -137,20 +137,25 @@ export function MessagesScreen({navigation, route}: Props) {
     [navigation],
   )
 
-  const onScrollToTop = useCallback(() => {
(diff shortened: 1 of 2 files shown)


estrattonbailey (commented):
Scroll to top works, but I think we also want the listConvos query to refetch, right? That would need to be added to the `listenSoftReset` handler. Feed pages also truncate the infinite pagination cache first so that the query doesn't refetch all currently fetched pages, but that's probably less of an issue here.

mozzius:
I was thinking we don't need it to refetch since in theory it should be keep up to date via log polling

estrattonbailey:
Atm we poll every 60s, so could be a decent way get fresh data faster. I think we might as well add it here. What do you think?

estrattonbailey (approved):

Issue #7700, opened 2025-02-08
Feed search is very unflexible


thktkblt:
### Steps to Reproduce

For example, here's a three words art feed with punctuations.
If you skip the middle word or punctuation, it won't appear. It needs to be the exact same keywords in consecutive order.
I think it's such an unflexible search. Worse since only keywords in the title (not description) is searched. People will miss out a lot of fitting feeds. 

![Image](https://github.com/user-attachments/assets/40117385-d961-47d1-8c96-3841168ffb1e)
![Image](https://github.com/user-attachments/assets/e0d518a4-e527-49ce-a000-0692cc025bf4)
![Image](https://github.com/user-attachments/assets/63375803-8e35-4ffc-92f1-51cf4c229672)

### What platform(s) does this occur on?

Web (Mobile)

### What version of the app are you using?

-

thktkblt:
Sorry I posted a new issue but it's probably better to comment here instead.

The fix seems to be undone, the feed search is inflexible again.

PR #7779, opened 2025-02-18
Fixed flexibility of feed search

MichaScant:
Relates to issue #7700,

Searching through feeds no longer requires exact wording, even if your search skipped the middle word or punctuation, the result should still appear

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -345,6 +345,18 @@ export const createPopularFeedsSearchQueryKey = (query: string) => [
   query,
 ]
 
+function normalizeText(text: string): string {
+  return text
+    .toLowerCase()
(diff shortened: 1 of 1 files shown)


MichaScant:
Hello, as an update, even though the change I made to the front end works, I believe a better solution would be to modify the backend in atproto, specifically reformating the query to replace all spaces with '%' such that it doesn't perform strict matching. Below is a screenshot of a change I would make in the 'feed-gens.ts' file in atproto to achieve the solution. 

Please let me know if this is preferable, and I will open a pr in that repository.

Screenshot 2025-02-21 at 4 07 50 PM

mozzius:
Yeah, ideally we would like to avoid multiple requests per typeahead search

MichaScant:
pr opened on at-proto:

https://github.com/bluesky-social/atproto/pull/3593

Issue #7586, opened 2025-01-26
profile images arent burred in reply dialog


ImLunaHey:
### Steps to Reproduce

1. find a post that has a blurred profile image
2. click reply
3. see the profile image without the blur

### Attachments

<img width="603" alt="Image" src="https://github.com/user-attachments/assets/b516ed6c-4792-4184-b16a-34bc719f53be" />
<img width="633" alt="Image" src="https://github.com/user-attachments/assets/120c003f-9502-4fd8-a198-962a9a52f5b3" />

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile), Android, iOS

### What version of the app are you using?

Build version: 1.96.6; Bundle info: 5090426 (prod); Bundle date: 25012002; Platform: web; Platform version: undefined

khuddite:
I was not able to reproduce the issue on the latest main branch. I suppose it's already been fixed recently.

However, it was not blurred for the recent searches - 

![Image](https://github.com/user-attachments/assets/98bcef5d-3b93-44fe-bd78-ee5c0b549220)

![Image](https://github.com/user-attachments/assets/9c3e3446-6246-46e3-b05f-89c08ded8119)

![Image](https://github.com/user-attachments/assets/fc72e941-ac22-4430-a8a6-772b169058e7)

ImLunaHey:
it's not fixed. there are two views. one is by clicking on the reply button on the post and the other is clicking in the reply box at the bottom of the a screen. 

one of them shows it and one hides. 

https://github.com/user-attachments/assets/51872819-2cfe-4cf3-b6ed-dac83d9bd67e

seth-wood:
Direct fix PR for issue here - #7783

estrattonbailey:
Fixed in #7783, will go out in our next release

PR #7783, opened 2025-02-19
Profile Moderation Blur Bug Fix

seth-wood:
More direct fix for issue #7586 by adding moderation cache element to PostThread Reply.

Ensures moderation blur is present on PostThread as well as Reply Button for Composer.


estrattonbailey (approved):
Thank you!

PR #7796, opened 2025-02-20
Fix DMs input automatically accepting autocomplete suggestion instead of clearing

mozzius:
Explained in a comment:
```
// bit of a hack: iOS automatically accepts autocomplete suggestions when you tap anywhere on the screen
// including the button we just pressed - and this overrides clearing the input! so we watch for the
// next change and double make sure the input is cleared. It should *always* send an onChange event after
// clearing via setMessage('') that happens in onSubmit()
```
Basically, pressing a button will always accept the autocomplete suggestion on iOS - including the send button! and for some reason this takes priority over our code trying to clear the input.
Solution: wait for the next change event and clear it then.

Also includes storybook cleanup

# Test plan
- Type something that'll get autocomplete suggestions like "Hel" and confirm that this PR correctly clears the input on send
- Confirm that this does not impact future inputs you might make
- Ensure no regressions on android

### DIFF
--- a/src/screens/Messages/components/MessageInput.tsx
+++ b/src/screens/Messages/components/MessageInput.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import {useCallback, useState} from 'react'
 import {Pressable, TextInput, useWindowDimensions, View} from 'react-native'
 import {
   useFocusedInputHandler,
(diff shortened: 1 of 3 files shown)


src/screens/Messages/components/MessageInput.tsx
```diff
@@ -81,14 +82,11 @@ export function MessageInput({
...
     onSendMessage(message)
     playHaptic()
-    setMessage('')
     setEmbed(undefined)
-
-    // Pressing the send button causes the text input to lose focus, so we need to
-    // re-focus it after sending
-    setTimeout(() => {
-      inputRef.current?.focus()
-    }, 100)
```
mozzius, line 91: I think changing the text input to use `submitBehavior="submit"` means this is no longer necessary

haileyok: hmm, i didnt actually look up where they changed this, but seems kinda weird that this would have any effect. it seems like they just changed the name of the prop from the docs :thinking:

just have to make sure that pressing outside of the text input doesn't cause it to unhighlight

mozzius: ohhh this is for web lol, will restore it

haileyok (approved):

LovesWorking:
@mozzius This hack doesn't work in PROD or local for me. I put in a PR that solves this issue in a different way.

https://github.com/bluesky-social/social-app/pull/7849

PR #7799, opened 2025-02-21
Add crowdin nightly push

pfrazee:
With the switch to the crowdin enterprise, I'm going to make two updates:

1. Gotta update the core config file
2. Change the nightly action to push directly to crowdin after extracting source strings. Previously this was handled by the GitHub integration in crowdin, but that tool only supported bidirectional sync. All we want to do is regularly sync the source strings to crowdin.

### DIFF
--- a/.github/workflows/nightly-update-source-languages.yaml
+++ b/.github/workflows/nightly-update-source-languages.yaml
@@ -36,3 +36,13 @@ jobs:
         with:
           commit_message: Nightly source-language update
           file_pattern: ./src/locale/locales/en/messages.po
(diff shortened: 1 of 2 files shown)


surfdude29:
@estrattonbailey @mozzius The Crowdin nightly push action seems to have failed last night:

```
ERROR: The key you are authenticating with has been marked as read only.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

https://github.com/bluesky-social/social-app/actions/runs/20587553190/job/59126601591#step:8:1494

PR #7803, opened 2025-02-21
Better screen transitions for auth flow

mozzius:
Reusing the much nicer screen transition component from the starter pack flow, the auth flow now has animations that work backwards and forwards along with some added fades to smooth it all out.

# Before

https://github.com/user-attachments/assets/68a24a90-6c80-4ae5-9f23-5dbfdfa9648a

# After

https://github.com/user-attachments/assets/858d5a95-186f-4bae-83f8-00f2f0c5b405

## Test plan

Test out signup and login flows, confirm all still works and all transitions work as expected (right direction etc)
Confirm starter pack flow is unaffected
Confirm web is unchanged


mozzius:
Superceded by #8650

caidanw:
Web looks fine, no issues.

Tested on ios and seems like the page transition isn't quite the same, it's very quick and looks like more of a fade-in rather than a slide-in.

Most notable when transitioning from the "Your account" to the "Choose your username" page, the other pages seems to _slide_ a bit better. The back button slide seems more consistent.

https://github.com/user-attachments/assets/d2b2edcd-6cb0-49de-b626-5110aa283bc3

mozzius:
Hmm, I can't really repro. It could be just that it's rendering quite a heavy screen so there's a bit of lag

caidanw:
Alright, well if it's just a bit of lag then it looks good. Approved, feel free to merge

PR #7811, opened 2025-02-21
Basic minimum password requirements, plus field-specific errors

mozzius:
First and foremost: passwords are now minimum 8 characters.
Signup form errors can now be specific to a field, and go away once once the field passes the validity check so that the user knows when it's good.

Also rearranged the email validity check, because it would do the double check confirm then reject anyway if it failed the regex. Now it eagerly rejects really bad emails, doing the TLD check last

https://github.com/user-attachments/assets/6417c713-f707-45b4-b232-e618c7647313

# Test plan

Stress test the signup form. Enter errors in different orders etc, see that they clear properly

### DIFF
--- a/src/screens/Signup/StepInfo/index.tsx
+++ b/src/screens/Signup/StepInfo/index.tsx
@@ -119,6 +119,12 @@ export function StepInfo({
         value: _(msg`Please choose your password.`),
       })
     }
+    if (password.length < 8) {
+      return dispatch({
+        type: 'setError',
+        value: _(msg`Your password must be at least 8 characters long.`),
+      })
+    }
 
     dispatch({type: 'setInviteCode', value: inviteCode})
     dispatch({type: 'setEmail', value: email})
--- a/src/screens/Signup/StepInfo/index.tsx
+++ b/src/screens/Signup/StepInfo/index.tsx
@@ -107,7 +107,7 @@ export function StepInfo({
         return dispatch({
           type: 'setError',
           value: _(
-            msg`It looks like you may have entered your email address incorrectly. Are you sure it's right?`,
+            msg`Please double-check that you have entered your email address correctly.`,
           ),
         })
       }
--- a/src/screens/Signup/StepInfo/index.tsx
+++ b/src/screens/Signup/StepInfo/index.tsx
@@ -99,30 +99,35 @@ export function StepInfo({
       return dispatch({
         type: 'setError',
         value: _(msg`Please enter your invite code.`),
+        field: 'invite-code',
       })
     }
     if (!email) {
       return dispatch({
         type: 'setError',
         value: _(msg`Please enter your email.`),
+        field: 'email',
       })
     }
     if (!EmailValidator.validate(email)) {
       return dispatch({
         type: 'setError',
         value: _(msg`Your email appears to be invalid.`),
+        field: 'email',
       })
     }
     if (!password) {
       return dispatch({
         type: 'setError',
         value: _(msg`Please choose your password.`),
+        field: 'password',
       })
     }
     if (password.length < 8) {
       return dispatch({
         type: 'setError',
         value: _(msg`Your password must be at least 8 characters long.`),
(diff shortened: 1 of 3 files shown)


src/screens/Signup/StepInfo/index.tsx
```diff
@@ -99,24 +83,50 @@ export function StepInfo({
...
       })
     }
+    if (emailChanged && tldtsRef.current) {
+      if (isEmailMaybeInvalid(email, tldtsRef.current)) {
+        prevEmailValueRef.current = email
+        setHasWarnedEmail(true)
+        return dispatch({
+          type: 'setError',
+          value: _(
+            msg`It looks like you may have entered your email address incorrectly. Are you sure it's right?`,
```
haileyok, line 110: very minor nit and im sure we're breaking this elsewhere, but we should avoid contractions in the ui - particularly around any account management stuff.

```suggestion
            msg`It looks like you may have entered your email address incorrectly. Are you sure it is correct?`,
```

src/view/com/modals/ChangePassword.tsx
```diff
@@ -81,15 +81,26 @@ export function Component() {
...
         _(
           msg`You have entered an invalid code. It should look like XXXXX-XXXXX.`,
         ),
       )
       return
     }
+    if (!newPassword) {
+      setError(
+        _(
+          msg`Please enter a password. It should be at least 8 characters long.`,
```
haileyok, line 95:
```suggestion
          msg`Please enter a password. It must be at least 8 characters long.`,
```

src/screens/Signup/StepInfo/index.tsx
surfdude29: Any change in a string will need to be re-translated, but as I explained to @mozzius in another PR, that's not really a big deal imo, we translators are used to it. And with the translation memory feature that we have with Crowdin now, it's even easier and quicker 💫  So fwiw, my personal suggestion would be to always get the English string as good as it can be, and let us translators worry about keeping up :)

But I did also want to chime in with an alternative suggestion... what about rephrasing it as a request rather than a question?

E.g. something like:

> Please double-check that you've entered your email address correctly.

haileyok: Thanks for that explainer! Also yea, I'm also fine with that suggestion so long as we drop the contraction. Again, I think we've been pretty lose with that in the past but particularly around the sensitive bits of the app/ecosystem I think we should strive to avoid informal writing.

surfdude29: Oops, I didn't even realise that I'd included a contraction in my suggestion, d'oh 🤦🏻‍♂️😅

surfdude29: On that point, the button has the text `It's correct`, I'm not sure what would work better without a contraction there (`It is correct` somehow doesn't seem quite right)
IMG_7106

haileyok: (I am not a UI expert :upside_down_face:) I think in the actual buttons where we're dealing with space concerns or very brief statements it's fine? It's mostly just other copy around the app - i.e. small paragraphs, sentences, etc. - that we should be a bit more careful. The "It's Correct" seems right here, I agree. And also cannot come up with something better.

mozzius: Alternative would be "Confim email" I guess but I'm ok with "It's correct"

surfdude29:
@mozzius Suggestion for a follow-up tweak at some point: when there's an error state for the email or password field, the **Next** button could change to the disabled/inactive state (e.g. like the handle screen displays when the handle is too long)

IMG_3119

mozzius:
@surfdude29 hmm, not sure I'd want to imply that the form is disabled. Forms typically keep the submit button active at all times, even when there's validation errors

surfdude29:
@mozzius Ah got it 👍

PR #7814, opened 2025-02-22
Remove system-ui from font family when set to Theme in Font Settings

liberaldev:
The difference with #6139 is that the font family still includes system-ui when set to System in the font settings. If system-ui is included in the alternate fonts in the theme font settings, it can cause issues with font readability in CJK Windows environments, as pointed out in #6139. However, some people prefer the systme-ui font, so this PR ensures that system-ui is included in the font family when set to System.

### DIFF
--- a/src/alf/fonts.ts
+++ b/src/alf/fonts.ts
@@ -3,7 +3,7 @@ import {TextStyle} from 'react-native'
 import {isAndroid, isWeb} from '#/platform/detection'
 import {Device, device} from '#/storage'
 
-const WEB_FONT_FAMILIES = `system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji"`
+const WEB_FONT_FAMILIES = `-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji"`
 
 const factor = 0.0625 // 1 - (15/16)
 const fontScaleMultipliers: Record<Device['fontScale'], number> = {
--- a/src/alf/fonts.ts
+++ b/src/alf/fonts.ts
@@ -3,6 +3,9 @@ import {TextStyle} from 'react-native'
 import {isAndroid, isWeb} from '#/platform/detection'
 import {Device, device} from '#/storage'
 
+const WEB_FALLBACK_FONT_FAMILIES =
+  'sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"'
+
 const factor = 0.0625 // 1 - (15/16)
 const fontScaleMultipliers: Record<Device['fontScale'], number> = {
   '-2': 1 - factor * 3,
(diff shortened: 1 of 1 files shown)


liberaldev:
> But the issue in #6139 occurs regardless of the theme font.
> 
> It seems better to remove system-ui altogether, or to remove system-ui from CJK only.

Yes, so if this PR is merged, this will fix the issue if the font setting is set to the theme font. The reason we keep system-ui in the system font setting is because some people want the system font to come out, like in #5722.

quiple:
> > But the issue in #6139 occurs regardless of the theme font.
> > It seems better to remove system-ui altogether, or to remove system-ui from CJK only.
> 
> Yes, so if this PR is merged, this will fix the issue if the font setting is set to the theme font. The reason i keep system-ui in the system font setting is because some people want the system font to come out, like in #5722.

This is because the current `font-family` specifies fonts including `Segoe UI`. And if you remove all of them as shown below, the user-customized system fonts will be output.

```
sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
```

It doesn't have to use `system-ui` to respect user-customized system/browser fonts.

However, if this is not the intention of Bluesky and you need to ignore user fonts, you could probably set it up like this:

```
-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
```

liberaldev:
> > > But the issue in #6139 occurs regardless of the theme font.
> > > It seems better to remove system-ui altogether, or to remove system-ui from CJK only.
> > 
> > 
> > Yes, so if this PR is merged, this will fix the issue if the font setting is set to the theme font. The reason i keep system-ui in the system font setting is because some people want the system font to come out, like in #5722.
> 
> This is because the current `font-family` specifies fonts including `Segoe UI`. And if you remove all of them as shown below, the user-customized system fonts will be output.
> 
> ```
> sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
> ```
> 
> It doesn't have to use `system-ui` to respect user-customized system/browser fonts.
> 
> However, if this is not the intention of Bluesky and you need to ignore user-customized system/browser fonts, you could probably set it up like this:
> 
> ```
> -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
> ```

The `sans-serif` keyword does not necessarily display the system font. For example, on a Korean Mac system, browsers like Firefox or Chrome use Apple SD Gothic Neo even for Latin glyphs, while the actual system font on Mac uses Apple SD Gothic Neo only for Korean glyphs and uses the San Francisco typeface for Latin glyphs, creating a discrepancy with the true Mac system fonts. Therefore, if you want to call the system font without considering the Windows environment, `system-ui` is ideal.

However, when taking the Windows environment into account, there is an issue where using the `system-ui` keyword does not display the appropriate font based on the lang attribute. For this reason, in Windows environments (especially in CJK Windows environments), readability can sometimes be extremely poor. Considering this, I modified the font settings by prepending `-apple-system` and `BlinkMacSystemFont` before the system sans-serif, resulting in the following font-family: `apple-system, BlinkMacSystemFont, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"`.

Issue #7807, opened 2025-02-21
User can pre-load opening Replyee's profile while replying


himham-jak:
### Steps to Reproduce

1. Find a post to reply to and hit reply.
2. Then tap the profile image of the replyee.
3. Post the reply and the replyee's profile will instantly open.
4. Repeat steps 1 & 3 without step 2 and you will be returned to the feed after posting.

### What platform(s) does this occur on?

Android

### What version of the app are you using?

Build version: 1.98.0.615; Bundle info: bd1f15260 (prod); Bundle date: 25021921; Platform: android; Platform version: 31; Anonymous ID: 88ed40ca-35ac-49b2-aee9-77c5d63b6a65

### Additional Information

(this issue template has a typo in the app version question: it says visting instead of visiting.)

seth-wood:
PR submitted for fix here: #7815

mozzius:
Fixed in https://github.com/bluesky-social/social-app/pull/7816

PR #7815, opened 2025-02-22
Avatar profile preloading fix

seth-wood:
As mentioned in #7807 - this fixes the possible unintended preloading/profile switching when pressing the avatar on mobile and leaving the feed. It removes the Link wrapping of the avatar when the reply composer is activated.


seth-wood:
re-submitting with improved technique

PR #7825, opened 2025-02-23
Tweak app language string in Settings

surfdude29:
Following [feedback on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-gd/9#2667):
> This is a rather tortured English sentence for selecting the UI language and should be revisited. Something like "Select a language for the apps user-interface" would be much easier to understand. The way it's now, it first made me wonder if this was a language selector for some sort of dummy text in certain fields until I found it in the UI language selection menu.

This PR proposes slightly different wording, changing the text from:
`Select your app language for the default text to display in the app.`
to
`Select which language to use for the app's user interface.`

### DIFF
--- a/src/screens/Settings/LanguageSettings.tsx
+++ b/src/screens/Settings/LanguageSettings.tsx
@@ -82,8 +82,8 @@ export function LanguageSettingsScreen({}: Props) {
             <View style={[a.gap_md, a.w_full]}>
               <Text style={[a.leading_snug]}>
                 <Trans>
-                  Select your app language for the default text to display in
-                  the app.
+                  Select which language you'd like to use for the app's user
+                  interface.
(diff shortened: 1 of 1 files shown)


haileyok (approved):

surfdude29:
@haileyok I only just realised I suggested a contraction here, maybe just change the last part to `the user interface of the app.`?

haileyok (approved):
In this case, the apostrophe doesn't form a contraction since it's just making the noun possessive, so this is fine. Mostly we're worried about stuff like "cannot/can't", "were not/weren't", etc.

surfdude29:
@haileyok Oh yeah ofc, sorry my brain isn't working today apparently 🙃

PR #7832, opened 2025-02-24
New reporting flow

estrattonbailey:
This PR introduces a new reporting flow that will power new functionality defined by labelers:
- labelers will declare which report reasons they support e.g. spam, sexual, etc
- labelers will declare what types of subjects they support e.g. `account`, `record` and `chat`
- labelers will declare what NSID collections they support e.g. `app.bsky.feed.post` but not `app.bsky.feed.generator`

By default, and for backwards compat, if labelers do not declare settings for any of these, we assume they support them all. So if you run this PR right now, the report options should be exactly the same as they are in production, just in a new format.

And yes we opted to encourage users to submit to a single labeler at a time, since once these new lexicons roll out, users should have far fewer options to report to. Most atm will likely only route to Bluesky's moderation service.

Normal reporting flow:

https://github.com/user-attachments/assets/23e8a8e3-d1e6-4b2c-93d1-47469d9be6be

Reporting content with only one labeler available, auto-selects this labeler (in this case an account, but could be any record):

https://github.com/user-attachments/assets/34d8eca1-adc3-44b0-a9c2-f05c41440dab

Reporting other types of content e.g. feeds, lists:

https://github.com/user-attachments/assets/e31057ad-ce3a-4a56-a5a0-a3f0d9c4f372

Submission error state
CleanShot 2025-02-24 at 14 49 54@2x

Additional details expanded (also happens if Other is selected as report option):
CleanShot 2025-02-24 at 14 49 15@2x

Should-be-impossible scenario that there are no labelers available:
CleanShot 2025-02-24 at 14 48 49@2x

Initial loading state:
CleanShot 2025-02-24 at 14 47 59@2x

Initial load of labelers fails:
CleanShot 2025-02-24 at 14 48 17@2x

Developer error, passed in an object that can't be reported (should catch this in dev, but added a graceful fallback anyway)
CleanShot 2025-02-24 at 14 47 46@2x


mozzius:
Light mode is a little odd... maybe it's too monochrome? maybe the circle bit for the active one should be blue?


Screenshot 2025-02-24 at 21 33 25

src/components/moderation/ReportDialog/index.tsx
```diff
@@ -0,0 +1,626 @@
...
+              <OptionCardSkeleton />
+              <OptionCardSkeleton />
+              <OptionCardSkeleton />
+              {/* Here to capture focus for a hot sec to prevent flash */}
+              <Pressable accessible={false} />
+            </View>
+          ) : labelersLoadError || !allLabelers ? (
+            <Admonition type="error">
+              <Trans>Something went wrong, please try again.</Trans>
+            </Admonition>
```
mozzius, line 194: Can we put a retry button here? Given we're asking them to try again

mozzius:
Does the dialog need a close button on web?

src/components/moderation/ReportDialog/index.tsx
estrattonbailey: Gonna reply to all here:
- gonna poke at UI a bit more, I agree light mode is esp meh
- yeah a retry button here is a great idea
- close button, yeah lemme see about a better header for this dialog, feels naked rn

.eslintrc.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
           'H6',
           'P',
           'Admonition',
+          'Admonition.Admonition',
```
estrattonbailey, line 35: Helpful in cases where we import `* as Admonition`

src/lib/async/wait.ts
```diff
@@ -0,0 +1,5 @@
+export async function wait<T>(delay: number, fn: T): Promise<Awaited<T>> {
+  return await Promise.all([fn, new Promise(y => setTimeout(y, delay))]).then(
+    arr => arr[0],
+  )
+}
```
estrattonbailey, line 5: This is a util to provide a _minimum_ execution time. Helpful for load states and things.

src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -551,14 +554,15 @@ function DialogInner({
...
-            />
+            {info.view && (
+              <ReportDialog
+                control={reportDialogControl}
+                subject={{
+                  ...info.view,
+                  $type: 'app.bsky.feed.defs#generatorView',
+                }}
+              />
+            )}
```
estrattonbailey, line 565: Needed to pipe this through from feed info query. In this case, it should always be defined (unless the feed fails to load entirely).

It's optional in the types bc our PWI Discover feed config needs to conform to this too, and hard-coding a view here seems silly,  since you can't report Discover from PWI anyway.

mozzius:
Seems like it doesn't go away immediately on Android - not sure why

https://github.com/user-attachments/assets/7e9e949e-3f5a-440a-b0c8-467eaf1e5da0

mozzius (approved):
Lgtm! Other than the android issue (which is not the fault of this PR most likely), I haven't found any substantial issues. I'll do another round of QA once the appview rolls out

src/components/moderation/ReportDialog/index.tsx
```diff
@@ -0,0 +1,652 @@
...
+            <Admonition.Outer type="error">
+              <Admonition.Row>
+                <Admonition.Icon />
+                <Admonition.Text>
+                  <Trans>Something went wrong, please try again</Trans>
+                </Admonition.Text>
+                <Admonition.Button
+                  label={_(msg`Retry loading report options`)}
+                  onPress={() => refetchLabelers()}>
+                  <ButtonText>Retry</ButtonText>
```
surfdude29, line 200:
```suggestion
                  <ButtonText>
                    <Trans>Retry</Trans>
                  </ButtonText>
```
I think translation tags are missing here (not sure if the formatting is right in my suggestion though)

surfdude29 (commented):
This works great on the preview branch, fantastic work @estrattonbailey 👌

src/components/moderation/ReportDialog/index.tsx
estrattonbailey: Ah good call. Did it locally to ensure formatting, thanks for the bump!

PR #7833, opened 2025-02-24
Dedupe profile labels

mozzius:
# Test plan

Subscribe to https://bsky.app/profile/decentralised.goeo.lol
Check https://bsky.app/profile/retr0.id has only 1 label


estrattonbailey (approved):
LGTM, though we might want to apply this to `PostAlerts` (in this same directory) too. Not a problem atm bc we don't show both profile and account labels on posts, but conceivably it could happen that a single post receives duplicate labels applied.

In this case, the `unique` util could live here in this dir, or in `lib/moderation`.

PR #7834, opened 2025-02-24
Translate hosting provider text

surfdude29:
I noticed that the `You are creating an account on` hosting provider text shown when creating an account isn't currently translatable:

IMG_3121

This PR marks just that text for translation. I did try to mark the whole thing (including the button, so that the order could be changed for languages where that's appropriate), but I didn't realise that `<Trans>` needs to be inside `<Text>`. So there might be a better solution if it's possible to wrap the whole thing.


mozzius:
Yeah, if it's not bsky.social, the button will say `pds.mozzius.dev` (for example). so languages with difference sentence structures might need to get creative 😅

mozzius (approved):

PR #7835, opened 2025-02-24
[DMs] Optimise badge calculation

mozzius:


estrattonbailey (commented):
Nice nice!

- marking as read doesn't clear the `hasNew` indicator dot (with no number badge), however it does clear when the poll comes through
- once you reject and/or delete a chat, I think it gets stuck in some memory cache bc new messages from that same sender don't show up. I don't think this is actually a problem since you don't want to see that convo anyway, but FYI.
- the mark all as read button on web, not a fan of the animation personally, can discuss tho — on mobile, I didn't even notice it was there and actually clicked it to start a new chat, I don't know if that's the right call for the FAB, maybe we can fit it in the header?

mozzius:
1. easy fix, can do
2. yep, we have a cache - maybe should be pruned once the leave is successful
3. good feedback, I'm all ears - the core problem is that we don't have a good icon! I agree that the web thing is jank. We could maybe do a different style of FAB on mobile that is rectangular and has a label?

PR #7839, opened 2025-02-25
Fix max height of dialog flatlists

estrattonbailey:
Also noticed the `Close` button looked wack, and realized I had a wrapper around dialog content that is no longer needed with top-aligned dialogs.
CleanShot 2025-02-25 at 09 42 13@2x

### DIFF
--- a/src/components/Dialog/index.web.tsx
+++ b/src/components/Dialog/index.web.tsx
@@ -219,8 +219,8 @@ export const InnerFlatList = React.forwardRef<
       style={[
         a.overflow_hidden,
         a.px_0,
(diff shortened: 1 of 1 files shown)


mozzius (approved):
Normal dialogs no longer appear near the top (simple Prompts etc) - I assume that that's intentional? Nothing looks broken now at least, but make sure your top-aligned dialogs still work ig

estrattonbailey:
Ah yep good call. Didn't regress on what I did the other day, but made prompts look kinda bad, esp when stacked on top of other dialogs. Fixed the close button position a different way that doesn't regress anything else.

PR #7847, opened 2025-02-26
Modify hosting provider text for better localization

quiple:
I moved the `<Button>` to inside the `<Trans>` because some languages need to be able to move the position of the hosting provider name.

Also modified the margin style to eliminate line height differences.

### DIFF
--- a/src/components/forms/HostingProvider.tsx
+++ b/src/components/forms/HostingProvider.tsx
@@ -43,21 +43,23 @@ export function HostingProvider({
       {minimal ? (
         <View style={[a.flex_row, a.align_center, a.flex_wrap]}>
           <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
-            <Trans>You are creating an account on </Trans>
+            <Trans>
+              You are creating an account on{' '}
+              <Button
+                label={toNiceDomain(serviceUrl)}
+                accessibilityHint={_(msg`Changes hosting provider`)}
+                onPress={onPressSelectService}
+                variant="ghost"
+                color="secondary"
+                size="tiny"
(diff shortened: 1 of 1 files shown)


surfdude29:
Thanks for doing this in a better way than my PR @quiple 🙏

surfdude29:
This PR is also important because it fixes a bug I accidentally introduced in #7834 because I didn't realise that Lingui seems to automatically drop any spaces at the end of a string. So the hosting provider button now overlaps the end of `on` on main.bsky.dev:

IMG_7456

Oops, sorry @mozzius 😬

src/components/forms/HostingProvider.tsx
```diff
@@ -43,21 +43,23 @@ export function HostingProvider({
...
+            <Trans>
+              You are creating an account on{' '}
+              <Button
+                label={toNiceDomain(serviceUrl)}
+                accessibilityHint={_(msg`Changes hosting provider`)}
+                onPress={onPressSelectService}
+                variant="ghost"
+                color="secondary"
+                size="tiny"
+                style={[a.px_xs, {margin: tokens.space.xs * -1}]}>
```
mozzius, line 55:
```suggestion
                style={[a.px_xs, {marginHorizontal: tokens.space.xs * -1}]}>
```

mozzius (changes requested):
One nit. Thanks though, should've caught that space issue myself

pfrazee:
Unfortunately this doesn't render correctly on mobile:

CleanShot 2025-03-06 at 19 18 31@2x

I'm going to have to revert and we'll have to look for another solution, or a fix to the layout.

quiple:
@pfrazee Couldn't we just modify the style `marginHorizontal` to `margin`?

PR #7849, opened 2025-02-26
Fix iOS auto-correct race condition in message input

LovesWorking:


src/screens/Messages/components/MessageInput.tsx
```diff
@@ -61,32 +61,30 @@ export function MessageInput({
   const [isFocused, setIsFocused] = useState(false)
   const [message, setMessage] = useState(getDraft)
   const inputRef = useAnimatedRef<TextInput>()
-  const [shouldEnforceClear, setShouldEnforceClear] = useState(false)
+  const messageRef = useRef(getDraft || '');
```
mozzius, line 64: Why are we putting the message state in a ref now? why can't we use `message`?

If it because it causes `onSubmit` to change, then we should replace `useCallback` with `useNonReactiveCallback`

mozzius (changes requested):
Ohhhh ok this makes so much more sense, thank you! I wasn't aware of `InteractionManager.runAfterInteractions()`, this is super helpful

Just left one question for you

src/screens/Messages/components/MessageInput.tsx
LovesWorking: Using a ref here ensures that we always access the latest message value, even if React’s state has not fully updated yet. You can test this by logging both message and messageRef.current inside onSubmit. You’ll see that message still holds the pre-auto-correct value, while messageRef.current correctly reflects the latest input.

If there’s an alternative way to ensure message updates fully before onSubmit runs, I’m open to suggestions. However, based on testing, this approach seems to be the simplest without introducing additional complexity.

TL;DR: InteractionManager.runAfterInteractions() + useRef together is the only way I could make this work reliably while keeping it straightforward.

LovesWorking: My bad. Yea looks like useNonReactiveCallback works!

LovesWorking: Pushed the changes. Thanks!

mozzius: Nice, thanks! Yeah, it's nice - I think we're going to get a similar API in react core one day

mozzius (changes requested):
This still seems to be broken, at least on the simulator. runAfterInteractions still runs before the autocorrect comes in

https://github.com/user-attachments/assets/3f5c1450-b474-483f-a1db-ce8ae5c3ee99

LovesWorking:
> This still seems to be broken, at least on the simulator. runAfterInteractions still runs before the autocorrect comes in. Does it work differently on device?
> 
>  Simulator.Screen.Recording.-.iPhone.16.-.2025-02-27.at.12.56.13.mp4

Thanks for checking this. It seems to randomly fail. I'm seeing 75% of the time it's working fine and hte other times it's not. 

> This still seems to be broken, at least on the simulator. runAfterInteractions still runs before the autocorrect comes in. Does it work differently on device?
> 
>  Simulator.Screen.Recording.-.iPhone.16.-.2025-02-27.at.12.56.13.mp4

image
image
useNonReactiveCallback vs useCallback with useRef
It looks like useNonReactiveCallback isn’t very predictable.

I switched back to a plain useCallback with useRef for the value, and everything works perfectly again. useNonReactiveCallback seems inconsistent—it works great about 75% of the time, but sometimes it fails completely. Not sure why.

PR #7855, opened 2025-02-27
Disable Sentry on native

estrattonbailey:
This cleans up all the native Sentry config and sourcemap upload stuff, which is not working on web anyway. Sentry has a handy CLI that works for this once we get sourcemaps figured out, which may require some digging.

**NB:** this PR also swaps the Sentry DSN so that we get a fresh start with a new project.


src/logger/sentry/setup/index.web.ts
```diff
@@ -0,0 +1,10 @@
+import {init} from '@sentry/react-native'
+
+init({
+  enabled: !__DEV__,
+  autoSessionTracking: false,
+  dsn: 'https://8fb55ba4807fca137eedfc8403ee27ba@o4505071687041024.ingest.us.sentry.io/4508807082278912',
+  debug: false, // If `true`, Sentry will try to print out useful debugging information if something goes wrong with sending the event. Set it to `false` in production
+  environment: process.env.NODE_ENV,
+  ignoreErrors: ['t is not defined'],
```
estrattonbailey, line 9: We still don't know what this is, but it renders Sentry useless bc of its prevalence, so I'm going to ignore it for now.

estrattonbailey:
Yeah this one can wait a hot sec, nbd

PR #7857, opened 2025-02-27
tweak link handling

haileyok:


estrattonbailey (approved):
LGTM

voxpelli:
Anywhere where one can read the reasoning behind this change? See discussion at eg: https://social.coop/@eloquence/114247449195843015 and https://bsky.app/profile/ottocr.at/post/3lllfd6pbbc2w

haileyok:
Yep, what you've found here is correct. As for why we went with this particular implementation, there are a number of different reasons:

1. It isn't possible to open a browser session and have referrer headers set. See i.e. https://developer.apple.com/documentation/safariservices/sfsafariviewcontrollerdelegate#//apple_ref/occ/intfm/SFSafariViewControllerDelegate/safariViewController:didCompleteInitialLoad. Nothing in the delegate runs _before_ opening the session, so it isn't possible to set them there.
2. There is the possibility of using a web view to open links within the application, rather than use the native web browser sessions. However, there are multiple reasons to not do this, ranging from lack of a user's session/cookies being present in the webview (think clicking links on a site that you have a subscription to. You'd want to be signed in when you view the article) as well as any of the user's browser extensions not being available.
3. Suggestions that "publishers could just add their own UTM parameters" to links wouldn't generally provide anywhere near accurate counts on clickthroughs. The bulk of links shared on social sites are _not_ posted by the original publishers, but by third parties who probably wouldn't be including those parameters. Ontop of that, they generally create clutter in links that isn't great.

It should be noted that:

- This is a _client specific_ behavior, and anyone who does not use the main Bluesky client would not encounter this redirect service (unless the creator of that app decided to use it still)
- Also worth noting (I think you saw this already, but just to be thorough here) this is _not_ done in any source material (i.e. posts stored in a user's repository) nor is it rendered in the HTML (i.e. crawlers would still see the actual link, not the redirect service)

surfdude29:
> That's a reasonable justification for native apps, but doesn't explain why it needs to be done for users of the Bluesky web client (which already sends referrer information). "So we can always have go.bsky.app as the referrer" seems an unreasonable justification for a change that unnecessarily degrades the user experience of web client users.

I wonder if it would be feasible to have the redirect proxy on the main `bsky.app` domain, rather than on `go.bsky.app`? E.g. would it work to have the redirect links as something like `https://bsky.app/go/redirect?u=${url}` instead?

But then again, I'm guessing there was probably some good reason for hosting the redirect proxy on its own subdomain 🤔

PR #7859, opened 2025-02-28
Colocate bitdrift in logger, use logger for transport

estrattonbailey:
Moves Bitdrift over alongside Sentry in `#/logger/bitdrift`. Also, instead of doing `bitdrift.info()` directly, let's just use the logger now, which will send to Bitdrift via its transports on native.


haileyok (approved):

haileyok:
lint error that needs fixing btw

PR #7867, opened 2025-02-28
Logger metrics

estrattonbailey:
Adds a `logger.metric()` method for more convenience emission of metrics. By default, this new method sends events to the datalake and _not_ to StatSig. This method also calls all configured logger transports e.g. Bitdrift and relays the same information.

```typescript
logger.metric('router:navigate', {from: <route>})
// also send to StatSig
logger.metric('init', {...}, {statsig: true})
```

This is powered by our existing `logEvent` from StatSig. However, I made a small update there to replace the manual `lake:` prefixing with an additional param:

```typescript
logEvent('lake:router:navigate', {from: <route>})
// becomes
logEvent('router:navigate', {from: <route>}, {lake: true})
```

Notable change here: event types no longer have the explicit `lake:` prefix, but I don't anticipate that being an issue atm. Should discuss.

Since these events are now called via the logger, I moved their definitions over there. I left StatSig otherwise untouched to avoid too much churn. Eventually I think we could set StatSig up as just another logger transport, but for the same reason I haven't done that here.

### DIFF
--- a/src/logger/transports/bitdrift.ts
+++ b/src/logger/transports/bitdrift.ts
@@ -18,7 +18,6 @@ export const bitdriftTransport: Transport = (
 ) => {
   const log = logFunctions[level]
   log(message.toString(), {
-    // match Sentry payload
     __context__: context,
     ...prepareMetadata(metadata),
   })
--- a/src/logger/transports/bitdrift.ts
+++ b/src/logger/transports/bitdrift.ts
@@ -19,7 +19,7 @@ export const bitdriftTransport: Transport = (
   const log = logFunctions[level]
   log(message.toString(), {
     // match Sentry payload
-    context,
+    __context__: context,
     ...prepareMetadata(metadata),
   })
 }
--- a/src/logger/types.ts
+++ b/src/logger/types.ts
@@ -5,6 +5,7 @@
  */
 export enum LogContext {
   Default = 'logger',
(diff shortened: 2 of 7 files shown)


src/logger/transports/bitdrift.ts
```diff
@@ -18,8 +18,7 @@ export const bitdriftTransport: Transport = (
 ) => {
   const log = logFunctions[level]
   log(message.toString(), {
-    // match Sentry payload
-    context,
+    __context__: context,
```
estrattonbailey, line 21: Also noting I renamed this so that it's clearly "special" and doesn't conflict with some metadata we already have defined on some metric events.

src/logger/types.ts
```diff
@@ -9,6 +9,12 @@ export enum LogContext {
   Notifications = 'notifications',
   ConversationAgent = 'conversation-agent',
   DMsAgent = 'dms-agent',
+
+  /**
+   * METRIC IS FOR INTERNAL USE ONLY, don't create any other loggers using this
+   * context
+   */
+  Metric = 'metric',
```
estrattonbailey, line 17: This will be added as a Sentry tag and property `__context__` on all emitted events. In dev, it also shows in the console like `(metric) <name>`.

haileyok (approved):

PR #7871, opened 2025-03-01
Add metrics to reporting flow

estrattonbailey:


src/components/moderation/ReportDialog/state.ts
```diff
@@ -3,6 +3,7 @@ import {AppBskyLabelerDefs, ComAtprotoModerationDefs} from '@atproto/api'
 import {ReportOption} from './utils/useReportOptions'
 
 export type ReportState = {
+  userInteracted: boolean
```
mozzius, line 6: Is this used at all? I don't see it unless I'm missing something

estrattonbailey: Ope, correct. I was going to use it on close, but I gotta watch for close at the outer level, so don't have access to this. Will remove 👍

mozzius (approved):

PR #7874, opened 2025-03-01
Remove use left convos

DogPawHat:
Follow on from comments in the Tanstack discord made by @mozzius about [`useMutationState`](https://discord.com/channels/719702312431386674/1345072602292686950/1345072602292686950). I've attempted to give a more "straightforward" answer over there, but I'm doing this pr to tackle the problem here

## My understanding of the issue

I may be *horrifically* wrong here, but this is what my current reading of the code entails.

If a user leaves and re-joins the convo, the leave convo mutation is still filtering out the convo they just re-joined, and `useLeftConvos` will still have the successful mutation in the cache, filtering out the convo they just rejoined (this is probably the part I'm most unsure of lol). Hence the need to remove the mutation from the mutation cache, which is somewhat non-trivial.

## A modest proposal

However, it looks like `useLeaveConvo` is already using the cache optimistic update pattern described here: https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#via-the-cache . Tanstack docs only say to use either the UI *or* the cache approach, and hence **I think we can just delete `useLeftConvos` entirely and we'll be fine anyway.**

And that's what the PR does.

We also are disabling `useListConvosQuery` if there are any pending leave convos mutations in flight so that the optimistic updates will not be overwritten (and the convo flashes up again briefly). This was apparently most of the motivation behind `useLeftConvos` in the first place; it makes sense to move that mutation monitoring into the query hook instead.

~~(I've might have another solution as well if you *really* need to hold on to `useLeftConvos`, will make an alternative pr.)~~ Actually I won't make a pr, might comment on it back in the discord.

## Testing

Leave conversations and rejoin them with the same people. Only tested the happy path so far; chat with the local backend doesn't seem great to set up.

### DIFF
--- a/src/state/queries/messages/leave-conversation.ts
+++ b/src/state/queries/messages/leave-conversation.ts
@@ -12,7 +12,7 @@ import {logger} from '#/logger'
 import {useAgent} from '#/state/session'
 import {RQKEY_ROOT as CONVO_LIST_KEY} from './list-conversations'
 
-const RQKEY_ROOT = 'leave-convo'
+export const RQKEY_ROOT = 'leave-convo'
 export function RQKEY(convoId: string | undefined) {
   return [RQKEY_ROOT, convoId]
 }
@@ -42,7 +42,8 @@ export function useLeaveConvo(
 
       return data
     },
-    onMutate: () => {
+    onMutate: async () => {
+      await queryClient.cancelQueries({queryKey: CONVO_LIST_KEY})
       let prevPages: ChatBskyConvoListConvos.OutputSchema[] = []
       queryClient.setQueryData(
         [CONVO_LIST_KEY],
--- a/src/state/queries/messages/leave-conversation.ts
+++ b/src/state/queries/messages/leave-conversation.ts
@@ -1,11 +1,5 @@
-
-import {
-  type ChatBskyConvoListConvos,
-} from '@atproto/api'
-import {
-  useMutation,
-  useQueryClient,
-} from '@tanstack/react-query'
+import {type ChatBskyConvoListConvos} from '@atproto/api'
+import {useMutation, useQueryClient} from '@tanstack/react-query'
 
 import {DM_SERVICE_HEADERS} from '#/lib/constants'
 import {logger} from '#/logger'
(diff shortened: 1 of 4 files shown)


src/state/queries/messages/leave-conversation.ts
```diff
@@ -18,12 +14,10 @@ export function RQKEY(convoId: string | undefined) {
 export function useLeaveConvo(
   convoId: string | undefined,
   {
-    onSuccess,
     onMutate,
     onError,
   }: {
     onMutate?: () => void
-    onSuccess?: (data: ChatBskyConvoLeaveConvo.OutputSchema) => void
```
DogPawHat, line 26: Didn't see this option being used at all. Easy enough to add back in if you need it.

src/state/queries/messages/leave-conversation.ts
```diff
@@ -85,25 +75,10 @@ export function useLeaveConvo(
           }
         },
       )
-      queryClient.invalidateQueries({queryKey: CONVO_LIST_KEY})
       onError?.(error)
     },
+    onSettled: () => {
+      queryClient.invalidateQueries({queryKey: CONVO_LIST_KEY})
+    },
```
DogPawHat, line 82: `onSettled` is called regardless of success.

DogPawHat (commented):
Calling out two other minor changes I've made. If the context you have suggests there bad for some reason I don't know about, you can revert them.

src/state/queries/messages/leave-conversation.ts
```diff
@@ -42,7 +36,8 @@ export function useLeaveConvo(
 
       return data
     },
-    onMutate: () => {
+    onMutate: async () => {
+      await queryClient.cancelQueries({queryKey: CONVO_LIST_KEY})
```
DogPawHat, line 40: see example at https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo

DogPawHat:
@mozzius you still think this PR will work?

PR #7893, opened 2025-03-04
update actions to use latest macos/xcode

haileyok:


mozzius (approved):
lgtm, let's give it a try

PR #7924, opened 2025-03-07
Fix hosting position on native

mozzius:


mozzius:
cc @surfdude29 @quiple - unfortunately we have to revert https://github.com/bluesky-social/social-app/pull/7847 because nesting a button inside `<Text>` causes layout issues on native

image

surfdude29:
@mozzius Is it possible to fix the spacing bug? https://github.com/bluesky-social/social-app/pull/7847#issuecomment-2685105158

PR #7930, opened 2025-03-07
Fix mention autocomplete

mozzius:
# Before

https://github.com/user-attachments/assets/d2838f6c-342d-43d4-bb62-7e0f2fb99c61

# After

https://github.com/user-attachments/assets/74c13342-3187-43c1-a7a1-f109c6a7dc40

# Test plan

Confirm these are working on iOS and Android:
1. Autocomplete list is visible, with and without embeds/images/quotes
2. Tapping anywhere on the sheet focuses the input (open gif menu to blur the input)


estrattonbailey (approved):
Tested on iOS/Android

PR #7935, opened 2025-03-07
Add checks for invalid syntax post-i18n builds

estrattonbailey:
Lingui [does not support failing](https://lingui.dev/ref/cli#extract) if there is invalid syntax in the translation files ([callsite here](https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/compile.ts#L141-L143)). In lieu of multiple patches on the dependencies themselves, I've opted here to output i18n builds to a log file and grepping for known failures. If failures are detected, it'll `exit 1`.

Example failure: https://github.com/bluesky-social/social-app/actions/runs/13726616350/job/38394470053

### DIFF
--- a/.github/workflows/build-submit-android.yml
+++ b/.github/workflows/build-submit-android.yml
@@ -56,7 +56,10 @@ jobs:
         run: yarn install
 
       - name: 🔤 Compile translations
-        run: yarn intl:build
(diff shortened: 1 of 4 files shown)


estrattonbailey:
Render deploy failure is also a sign that this PR is working

mozzius (approved):
nice

PR #7943, opened 2025-03-09
Simplify content warning dialog text

surfdude29:
After considering [feedback on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-de/7?view=comfortable#595) about the second sentence of the current explanatory text displayed in the **Add a content warning** dialog being confusing:

> This may be overstepping a bit, but I believe if you want people to use this function, this is a confusing way to put it. ("If none are selected, this post is suitable for all audiences." is a technical description for how to interpret absence of labels *after* the fact, but runs counter-causality as guidance while authoring a post!)
>
> I would instead write "If the media is suitable for all audiences, select none." or something along those lines.

I think it might be clearer to simplify the whole string and turn it into a polite request to add a content warning if it's applicable.

So this PR proposes changing the text from:

> Choose self-labels that are applicable for the media you are posting. If none are selected, this post is suitable for all audiences.

⬇️

> Please add any content warning labels that are applicable for the media you are posting.

cc: @Tamschi


src/view/com/composer/labels/LabelsBtn.tsx
```diff
@@ -113,9 +113,8 @@ function DialogInner({
           </Text>
           <Text style={[t.atoms.text_contrast_medium, a.leading_snug]}>
             <Trans>
-              Choose self-labels that are applicable for the media you are
-              posting. If none are selected, this post is suitable for all
-              audiences.
+              Please add any content warning self-labels that are applicable for
```
haileyok, line 116: what if we don't even call them "self-labels" at all and just say "labels"?
```suggestion
              Please add any content warning labels that are applicable for
```

haileyok (approved):
question re: wording, but fine otherwise

haileyok:
thanks for the tweak! :blue_heart:

surfdude29:
Thanks for reviewing all these PRs! 😊

PR #7948, opened 2025-03-10
New toasts

mozzius:
Native toasts on iOS and Sonner on web via `burnt`. Still use old toasts on Android, but with a nicer API for triggering them


fobos531:
Note that [SPIndicator](https://github.com/ivanvorobei/SPIndicator), which `burnt` uses for iOS under the hood, is not particularly great for even slightly longer messages (truncates text, no proper multiline support etc), and seems to be unmaintained for a couple of years now.

mozzius:
Good to know! fwiw, I only just got an iOS build compiled (had unrelated build issues). Now I'm looking at it, yeah.... not my favourite

mozzius:
I'm thinking I'll switch this to a web-only change tbh!

fobos531:
> I'm thinking I'll switch this to a web-only change tbh!

Sounds good. If you ever find yourself needing to explore other JS-based implementations, this has a good reputation: https://github.com/gunnartorfis/sonner-native

PR #7974, opened 2025-03-12
Extract center column offset, reduce by 20px

mozzius:
After inspecting closely, I think that the tablet mode offset is a little too far to the left. I extracted the offset to a constant, then adjusted it slightly

# Test plan

See if on tablet breakpoint the whitespace to the sides is more even

### DIFF
--- a/src/alf/breakpoints.ts
+++ b/src/alf/breakpoints.ts
@@ -27,6 +27,8 @@ export function useBreakpoints(): Record<Breakpoint, boolean> & {
   }, [gtPhone, gtMobile, gtTablet])
 }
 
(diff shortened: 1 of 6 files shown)


src/alf/breakpoints.ts
```diff
@@ -27,6 +27,8 @@ export function useBreakpoints(): Record<Breakpoint, boolean> & {
   }, [gtPhone, gtMobile, gtTablet])
 }
 
+export const CENTER_COLUMN_OFFSET = -130
```
estrattonbailey, line 30: I kinda think this should live in `Layout/const.ts`?

mozzius: sure can

Issue #7957, opened 2025-03-11
Invalid colon in filenames prevent pulling on Windows


Septh:
#7913 introduced 3 files in `__e2e__/flows/report-dialog` that have a colon in their names:

![Image](https://github.com/user-attachments/assets/f6a6c139-4ecb-4537-a8c6-0a5ad5598866)

Unfortunately, these files cannot be created on a Windows machine since on this OS, the colon is not a valid character in filenames. This prevents me from updating my local clone of this repository: git fails with an "invalid path" error.

Any chance these files could be renamed upstream?

chrisjacob:
I've also experienced this same issue on Windows 11.

Github Desktop:

```
error: invalid path '__e2e__/flows/report-dialog/post:default.yml'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
```

CMD:
```
git restore --source=HEAD :/
error: invalid path '__e2e__/flows/report-dialog/post:default.yml'
error: invalid path '__e2e__/flows/report-dialog/post:edit-reason.yml'
error: invalid path '__e2e__/flows/report-dialog/post:reason-other.yml'
```

LoganDupont:
Hey @mozzius, can we have this small one fixed? This would resolve the problem where devs can't pull the repo on Windows and prevent them from contributing

mozzius:
Done

LoganDupont:
Thank you 🙏

PR #7976, opened 2025-03-12
fix git clone error on Windows (#7957)

icealtria:
This PR fixes a problem where Windows users are unable to `git clone` the repository due to invalid filenames that contain colons (`:`).

Colons are not allowed in filenames on Windows (NTFS), which causes checkout failure after cloning.  
See issue #7957.


chrisjacob:
I've also experienced this same issue on Windows 11.

Github Desktop:

```
error: invalid path '__e2e__/flows/report-dialog/post:default.yml'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
```

CMD:

```
git restore --source=HEAD :/
error: invalid path '__e2e__/flows/report-dialog/post:default.yml'
error: invalid path '__e2e__/flows/report-dialog/post:edit-reason.yml'
error: invalid path '__e2e__/flows/report-dialog/post:reason-other.yml'
```

ssg:
I unknowingly implemented the same patch in #8264 , but used dots instead of dashes as there is already a file called `account.default.yml` in the same folder to follow the existing conventions.

PR #7982, opened 2025-03-13
Make `Posts` and `People` tabs translatable for user lists

surfdude29:
When viewing a user list, the `Posts` and `People` tabs aren't currently translatable.

This PR (hopefully) makes them translatable by moving the constant to inside the `ProfileListScreenLoaded` function. However, I don't know if this is the right way of doing this.

EDIT: I also added commit https://github.com/bluesky-social/social-app/pull/7982/commits/4d223faf8fc99915a10cc216902d9121adbd74d0 which tweaks a few labels in this file, to use sentence case and to remove hyphens from `Un-mute` and `Un-block`, which aren't used in any other strings in the app:

`Delete List` ➡️ `Delete list`
`Report List` ➡️ `Report list`
`Un-mute list` ➡️ `Unmute list`
`Un-block list` ➡️ `Unblock list`


auroursa:
This will also close #7175, I'm not sure which is the right way.

surfdude29:
> This will also close #7175, I'm not sure which is the right way.

Oops, I had forgotten about that PR, sorry 😬

The method in #7175 does seem simpler, as I don't think the constant is used elsewhere?

auroursa:
Honestly I’m not entirely sure about the intention behind using Section Titles Curate here. @/grearon seems to prefer define them inline with _(msg, so I tried removing it.

Maybe keeping Section Titles Curate as this PR does is the better way to go. Either way, as long as it solves the issue, it’s worth it 👍.

surfdude29:
> […] Either way, as long as it solves the issue, it’s worth it 👍.

True that 💫

haileyok (approved):

PR #7984, opened 2025-03-13
Translate number joined starter pack string and add plural formatting

surfdude29:
The string shown on a starter pack showing how many people have used it to join Bluesky – when this is at least 50 – isn't currently translatable.

IMG_1917

This PR makes it translatable and, following advice from @akerbeltz on Crowdin that plural forms should be made available for all strings where it's relevant, uses the `Plural` macro too. Since this string is only shown when the number who have joined using a starter pack is at least 50, only the `other` prop is used (which is the only plural form required [according to the docs](https://lingui.dev/ref/macro#plural-1).)

EDIT: I also added a couple of commits (forgot to import `Plural` in the first one) adding plural formatting to the equivalent string in `StarterPackScreen.tsx`.


akerbeltz:
I may be misunderstanding something but from the linguistic point of view, we cannot assume that a number above 50 falls under `other` for all languages. It does for gd and many others but it's under `one` for Prussian and under `zero` for Latvian (https://www.unicode.org/cldr/charts/46/supplemental/language_plural_rules.html). It would be best to just apply the whole set of plural options, that way, going forward, all eventualities are covered, no matter what the number.

surfdude29:
The [Lingui docs](https://lingui.dev/ref/macro#plural) give a tip for the `plural` macro:

>💡 TIP
> Choose the plural forms used in your source code based on the pluralization rules of your source locale.

It seems reasonable to assume that this tip applies to the `Plural` macro too.

> […] we cannot assume that a number above 50 falls under other for all languages.

@akerbeltz Yes, I agree – and you've helpfully pointed to specific examples that prove that it doesn't – but the source locale is English and we know that in English, numbers in the range 50 and above all fall under `other`. Therefore, we only need to include `other` in the source code.

When a translator is translating a string involving plurals on Crowdin, they simply need to use the **Copy source to target** or **Copy source skeleton to target**/**Copy Source Skeleton** command and the plural forms relevant for their language will be automatically filled out for that string, and they can then add the correct translation for each plural form  💫

haileyok (approved):

Issue #7931, opened 2025-03-07
(web) Conversations don't refresh when going offline and back online


LovesWorking:
### Steps to Reproduce

Go offline while in a conversation on web.
Send message in app in the same conversation.
Go back online on web and click re-connect in the conversation.
Messages don't refresh. I never see the messages I sent from mobile.

![Image](https://github.com/user-attachments/assets/0e7dfa94-34d5-41f1-95eb-57ddccfaf805)

![Image](https://github.com/user-attachments/assets/1b3980ee-02e4-40df-83e6-83f3c44326bd)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

chrome

### What version of the app are you using?

web

PR #7985, opened 2025-03-13
fix:Conversations don't refresh when going offline and back online

omar0404:
#7931


gaearon:
cc @mozzius @estrattonbailey

PR #7987, opened 2025-03-13
Tweak copy to remove remaining uses of `sign up`

surfdude29:
This is a follow-up to #6931 which changed `Sign up` to `Create account`.

I noticed a few lower-profile uses of `sign up` and `signup`, so this PR tweaks the visible copy (and one `accessibilityHint` & one comment) in a few places.

### DIFF
--- a/src/screens/Deactivated.tsx
+++ b/src/screens/Deactivated.tsx
@@ -200,13 +200,13 @@ export function Deactivated() {
                 <Trans>Or, continue with another account.</Trans>
               </Text>
               <Button
-                label={_(msg`Sign in or sign up`)}
+                label={_(msg`Sign in or create account`)}
                 size="large"
                 variant="solid"
                 color="secondary"
                 onPress={() => setShowLoggedOut(true)}>
                 <ButtonText>
-                  <Trans>Sign in or sign up</Trans>
+                  <Trans>Sign in or create account</Trans>
                 </ButtonText>
               </Button>
             </>
--- a/src/screens/StarterPack/StarterPackLandingScreen.tsx
+++ b/src/screens/StarterPack/StarterPackLandingScreen.tsx
@@ -300,14 +300,14 @@ function LandingScreenLoaded({
             ) : null}
           </View>
           <Button
-            label={_(msg`Signup without a starter pack`)}
+            label={_(msg`Create account without a starter pack`)}
             variant="solid"
             color="secondary"
             size="large"
             style={[a.py_lg]}
             onPress={onJoinWithoutPress}>
             <ButtonText>
-              <Trans>Signup without a starter pack</Trans>
+              <Trans>Create account without a starter pack</Trans>
             </ButtonText>
           </Button>
         </View>
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -215,7 +215,7 @@ export const LoginForm = ({
               blurOnSubmit={false} // prevents flickering due to onSubmitEditing going to next field
               editable={!isProcessing}
               accessibilityHint={_(
-                msg`Input the username or email address you used at signup`,
+                msg`Enter the username or email address you used when you created your account`,
               )}
             />
           </TextField.Root>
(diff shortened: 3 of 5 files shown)


src/screens/Deactivated.tsx
```diff
@@ -200,13 +200,13 @@ export function Deactivated() {
                 <Trans>Or, continue with another account.</Trans>
               </Text>
               <Button
-                label={_(msg`Sign in or sign up`)}
+                label={_(msg`Sign in or create account`)}
```
haileyok, line 203: maybe?
```suggestion
                label={_(msg`Sign in or create an account`)}
```

src/screens/Deactivated.tsx
```diff
@@ -200,13 +200,13 @@ export function Deactivated() {
...
                 size="large"
                 variant="solid"
                 color="secondary"
                 onPress={() => setShowLoggedOut(true)}>
                 <ButtonText>
-                  <Trans>Sign in or sign up</Trans>
+                  <Trans>Sign in or create account</Trans>
```
haileyok, line 209: Same as above...maybe this doesn't match what we have elsewhere so if it doesn't ignore me (though I think we should change it. This doesn't sound right...)

src/screens/StarterPack/StarterPackLandingScreen.tsx
```diff
@@ -300,14 +300,14 @@ function LandingScreenLoaded({
...
-            label={_(msg`Signup without a starter pack`)}
+            label={_(msg`Create account without a starter pack`)}
             variant="solid"
             color="secondary"
             size="large"
             style={[a.py_lg]}
             onPress={onJoinWithoutPress}>
             <ButtonText>
-              <Trans>Signup without a starter pack</Trans>
+              <Trans>Create account without a starter pack</Trans>
```
haileyok, line 310:
```suggestion
              <Trans>Create an account without a starter pack</Trans>
```

src/screens/Deactivated.tsx
haileyok:
```suggestion
                  <Trans>Sign in or create an account</Trans>
```

haileyok:
Probably fine to ignore these suggestions, mostly left them to discuss internally. I assume we went with the shorter option because of space for the button, thats fine. But we should at least use `create AN account` when we do not care about string length (i.e. in comments, voiceover, or other areas where we already are using significant space)

surfdude29:
All suggestions applied 👍

As you say, when there are no space constraints or we're already using significant space, `create AN account` is definitely better.

src/screens/StarterPack/StarterPackLandingScreen.tsx
```diff
@@ -300,14 +300,14 @@ function LandingScreenLoaded({
             ) : null}
           </View>
           <Button
-            label={_(msg`Signup without a starter pack`)}
+            label={_(msg`Create an account without a starter pack`)}
```
surfdude29, line 303:
```suggestion
            label={_(msg`Create an account without using this starter pack`)}
```

src/screens/StarterPack/StarterPackLandingScreen.tsx
```diff
@@ -300,14 +300,14 @@ function LandingScreenLoaded({
...
             variant="solid"
             color="secondary"
             size="large"
             style={[a.py_lg]}
             onPress={onJoinWithoutPress}>
             <ButtonText>
-              <Trans>Signup without a starter pack</Trans>
+              <Trans>Create an account without a starter pack</Trans>
```
surfdude29, line 310:
```suggestion
              <Trans>Create an account without using this starter pack</Trans>
```

surfdude29 (commented):
One other thought I had is that the string for the button displayed at the bottom of a starter pack landing page could maybe be a bit clearer. I think `Create an account without using this starter pack` might be best, but perhaps that's too wordy and too long for the button?

IMG_5228

src/screens/StarterPack/StarterPackLandingScreen.tsx
haileyok: think this is fine personally

surfdude29:
@haileyok This now passes CI when you have a chance to take a look 👀

haileyok (approved):

PR #7988, opened 2025-03-13
Fix missing RecordWithMedia media embed preview in notifications view

estrattonbailey:
I missed [this ternary](https://github.com/bluesky-social/social-app/pull/7344/files#diff-fb34c2252d82c53bb675900d3b6b169e3eabd30e5cd37e0ed2c0303d2631283aL33) when I touched this last month. This PR fixes that by doing one recurse of the rendering here.

To do so, I had to adjust the types output of `parseEmbed`. The inputs to `parseEmbed` and `parseEmbedRecordView` both require the `$Typed` generic, and therefore their outputs should match and specify the `$Typed` variant of these views. This change has no effect on runtime code, it's just more type-correct.

### DIFF
--- a/src/components/MediaPreview.tsx
+++ b/src/components/MediaPreview.tsx
@@ -54,6 +54,14 @@ export function Embed({
         <VideoItem thumbnail={e.view.thumbnail} alt={e.view.alt} />
       </Outer>
     )
+  } else if (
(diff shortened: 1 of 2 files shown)


mozzius (approved):
looks good. fixes the issue, I need to understand the new system better but I think it's good

PR #7997, opened 2025-03-15
Add plural formatting, translate and tweak some accessibility strings in composer

surfdude29:
In response to [feedback on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-gd/7?view=comfortable#675), this PR adds plural formatting for the `accessibilityHint` for the text input in the composer.

Similar to #7984 and #7994, only the `other` prop is included as that is the only plural form relevant in the source locale, English. That will also be the only relevant plural form for translations in most other locales. However, adding plural formatting here means we can ensure that the string can be correctly localized by translators in all supported locales.

Also, I noticed that there were a few untranslatable accessibility strings nearby so I marked those for translation.

I also tweaked two of them to take into account the case where several posts or replies are being published at once. However, if that change would be confusing or clunky then I can easily remove it.

> [!NOTE]
> **I put this PR together partly using ChatGPT and it passes CI but I haven't tested it.**

cc: @akerbeltz

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -481,8 +481,8 @@ export const ComposePost = ({
       thread.posts.length > 1
         ? _(msg`Your posts have been published`)
         : replyTo
-          ? _(msg`Your reply has been published`)
-          : _(msg`Your post has been published`),
+        ? _(msg`Your reply has been published`)
+        : _(msg`Your post has been published`),
     )
   }, [
     _,
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -953,20 +953,36 @@ function ComposerTopBar({
             testID="composerPublishBtn"
             label={
               isReply
-                ? _(
-                    msg({
-                      message: 'Publish reply or replies',
-                      comment:
-                        'Accessibility label for button to publish a reply or thread of replies',
-                    }),
-                  )
-                : _(
-                    msg({
-                      message: 'Publish post or posts',
-                      comment:
-                        'Accessibility label for button to publish a post or thread of posts',
-                    }),
-                  )
+                ? isThread
+                  ? _(
+                      msg({
+                        message: 'Publish replies',
+                        comment:
+                          'Accessibility label for button to publish multiple replies in a thread',
+                      }),
+                    )
+                  : _(
+                      msg({
+                        message: 'Publish reply',
+                        comment:
+                          'Accessibility label for button to publish a single reply',
+                      }),
+                    )
+                : isThread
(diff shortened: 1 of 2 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -942,7 +947,23 @@ function ComposerTopBar({
         ) : (
           <Button
             testID="composerPublishBtn"
-            label={isReply ? 'Publish reply' : 'Publish post'}
+            label={
+              isReply
+                ? _(
+                    msg({
+                      message: 'Publish reply or replies',
```
haileyok, line 954: Should this be conditionally plural?

src/view/com/composer/Composer.tsx
```diff
@@ -942,7 +947,23 @@ function ComposerTopBar({
...
+                      comment:
+                        'Accessibility label for button to publish a reply or thread of replies',
+                    }),
+                  )
+                : _(
+                    msg({
+                      message: 'Publish post or posts',
```
haileyok, line 961: Same q here

haileyok (approved):
looks good, just q about plural

akerbeltz:
These shouldn't need plurals, nothing is being counted i.e. it's not going to show as _4 posts_ or _29 posts_ to the user.

src/view/com/composer/Composer.tsx
surfdude29: I originally considered doing that but I thought it was too complex for me to manage (even with ChatGPT's help). However, I think Claude has done it?

surfdude29:
> These shouldn't need plurals, nothing is being counted i.e. it's not going to show as _4 posts_ or _29 posts_ to the user.

@akerbeltz If I understood correctly what @haileyok was meaning, I think she's referring to the strings being conditional on whether or not it's a thread of replies/posts being published or just a single reply/post.

So as I've updated it now, the accessibility label should be `Publish reply` if only a single reply is being published, and `Publish replies` if more than one reply is being published, and the equivalent for posts.

surfdude29:
> looks good, just q about plural

@haileyok Updated to be conditionally plural (thanks Claude!), although CI seems to be failing now due to Prettier and I'm confused as to why 🤔

haileyok:
cool lemme take a peek, thanks!

surfdude29:
Prettier was flagging an issue from the string change in #7943, so I fixed that (I think) in 24434f18c83227a82c2b6f20d48e75bd8c3601cb.

haileyok:
thanks....even locally it wasn't complaining, so not quite sure what happened there

haileyok:
yay finally! thanks!

PR #7998, opened 2025-03-15
Add plural formatting for "too long" error messages when editing profile info

surfdude29:
In response to feedback for [two](https://bluesky.crowdin.com/editor/1/11/en-gd/7?view=comfortable#891) [strings](https://bluesky.crowdin.com/editor/1/11/en-gd/7?view=comfortable#949) on Crowdin, this PR adds plural formatting for the `Display name is too long` and `Description is too long` error messages that can be shown when a user is editing their profile info.

Similar to #7984, #7994 and #7997, only the `other` prop is included as that is the only plural form relevant in the source locale, English. That will also be the only relevant plural form for translations in most other locales. However, adding plural formatting here means we can ensure that the string can be correctly localized by translators in all supported locales.

> [!NOTE]
> **I put this PR together partly using ChatGPT and it passes CI but I haven't tested it.**

cc: @akerbeltz

### DIFF
--- a/src/screens/Profile/Header/EditProfileDialog.tsx
+++ b/src/screens/Profile/Header/EditProfileDialog.tsx
@@ -335,7 +335,7 @@ function DialogInner({
                 Display name is too long.{' '}
                 <Plural
                   value={DISPLAY_NAME_MAX_GRAPHEMES}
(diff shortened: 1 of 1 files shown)


haileyok (approved):
lgtm

Issue #7037, opened 2024-12-10
add Arabic language and RTL support


TajAlasfiyaa:
### Key Features:
1. **Arabic Translation**:  
   - Introduce Arabic localization using the Lingui tool, creating an `ar` catalog with Arabic translations for the app's strings.  
   - Follow the documented workflow for wrapping strings in `Trans` and using `yarn intl:extract` to generate catalogs.  

2. **RTL Layout Support**:  
   - Automatically enable RTL layout when an RTL language (like Arabic) is selected.  
   - Adjust UI components, such as text alignment, icons, navigation menus, and other visual elements, to fit RTL standards.  

3. **Testing**:  
   - Verify that the app works seamlessly with Arabic language and RTL layout.  
   - Ensure proper display of text and user interaction in all components.  

### Attachments

Here are some examples of RTL layout implementations in other apps, demonstrating how text alignment, navigation, and UI components adapt for RTL:  
![1_h9eBYKF0LEK7tZfTZk6gfg](https://github.com/user-attachments/assets/4f843b83-97fb-4860-af98-c8f4519f4e9b)

### Additional Context

- As a native Arabic speaker, I am willing to assist with translation and testing.  
- Arabic support would significantly expand the app’s user base and improve inclusivity, reflecting Bluesky's commitment to accessibility.  
- Resources such as Lingui's documentation ([https://lingui.dev/](https://lingui.dev/)) and React Native's RTL guide ([https://reactnative.dev/docs/rtl](https://reactnative.dev/docs/rtl)) can be helpful for implementation.

FiberAhmed:
ترجمته ولا باقي؟ إذا للحين، أقدر أساعدك في التدقيق والترجمة

surfdude29:
It would be great to have an Arabic localization for Bluesky! 🦋

Bluesky has recently moved to using Crowdin to manage translation, there's lots of info here:
- #7685

This is what it says on adding a new language:

> **Adding a new language**
>
> Create a new [Crowdin discussion](https://crowdin.com/project/bluesky-social/discussions) or [GitHub issue](https://github.com/bluesky-social/social-app/issues) requesting the new language be added to the project.
>
> Please only request a new language when you are certain you will be able to contribute a substantive portion of translations for the language.

So if you're still interested – I realise it's been a couple of months since you opened this issue – I would recommend opening a discussion on Crowdin as the best way forward :)

FiberAhmed:
[...]

PR #8001, opened 2025-03-15
Add language: Arabic

surfdude29:
Arabic is now at 100% translated [on Crowdin](https://bluesky.crowdin.com/bluesky-social) so this PR proposes making the necessary changes for when it's decided to add it to the app.

> [!IMPORTANT]
> Arabic would be the first RTL (Right-to-Left) language supported by Bluesky as an app language. Therefore, as described in #7037 opened by @TajAlasfiyaa, testing to ensure the app works properly with an RTL app language will be necessary before merging this PR, which is why it is currently in draft.

cc: @pfrazee @estrattonbailey


src/locale/languages.ts
```diff
@@ -54,6 +55,7 @@ interface AppLanguageConfig {
 export const APP_LANGUAGES: AppLanguageConfig[] = [
   {code2: AppLanguage.en, name: 'English'},
   {code2: AppLanguage.an, name: 'Aragonés – Aragonese'},
+  {code2: AppLanguage.ar, name: 'اَلْعَرَبِيَّةُ – Arabic'},
```
EyadMahm0ud, line 58:
```suggestion
  {code2: AppLanguage.ar, name: 'العربية – Arabic'},
```
No need to all these diacritics; keep it clean.

surfdude29: Thanks! :)

PR #8007, opened 2025-03-17
Generate separate strings for join messages in Newskie dialog

### LINKED ISSUE #8940: German localization: wrong verb form in "joined Bluesky" message when viewing own profile
**Severity**: Minor UI / localization bug

**Environment**: (web / iOS / Android). Screenshots attached.

**Description / actual behaviour**

When viewing other users' profiles the banner reads correctly in German, e.g.
"SnackpackWayne ist Bluesky vor 1 Tag bei beigetreten" (third-person).

When viewing my own profile the banner shows the same structure but with "Du" + wrong verb:
"Du ist Bluesky vor 1 Tag beigetreten" — the verb is ist (3rd person) instead of bist (2nd person).

**Expected behaviour**

When the subject is the current user (second person), the German text must use the correct 2nd-person conjugation:
"Du bist Bluesky vor {time} beigetreten" (or the localized equivalent).

When the subject is another person, use the 3rd-person form:
"{profileName} ist Bluesky vor {time} beigetreten".

**Root cause (likely)**

The UI appears to use a single translation/template like "{profileName} joined Bluesky {0} ago" and simply substitutes profileName with the display string "Du" when viewing your own profile. This works in English but breaks languages (like German) that require different verb forms for second vs third person.

**Notes / scope**

Affects German and any other languages with person/verb differences.
The same issue also applies to the string "{0} joined this week" when {0} is replaced with "Du" → it should be "Du bist diese Woche beigetreten", not "Du ist diese Woche beigetreten".

Image
Image

surfdude29:
In response to [feedback on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-gd/7?view=comfortable#3579):

> It's a really really bad idea to use pronouns for concatenation because they are heavily inflected in many languages depending on the context. For example, while you as the subject may be _thu_ in ScG, _from you_ becomes _uat_ and _to you_ _dhut_ or _thugad_ depending on context. UI strings with pronouns are best hard coded. Or need a detailed explanation of context.

This PR rewrites the code in the newskie dialog to make a few changes, so that separate strings with full sentences are generated for each of the four scenarios:

1. Removes the special "You" transformation logic from the `profileName` calculation, so it now just contains the sanitized display name.
2. Creates a new `getJoinMessage()` function that:
- Generates the appropriate message based on whether it's the current user (`isMe`) and if they joined via starter pack
- Uses the Lingui `msg` macro for each of the four distinct messages
- Each message has proper interpolation for the time ago string and profile name where appropriate
3. Simplifies the JSX by using this function directly in the `Text` component instead of conditionally rendering different `Trans` components.

> [!NOTE]
> **Claude wrote the code for this PR (I certainly couldn't 😅) and although it passes CI I have not tested it.**

Fixes #8940

### DIFF
--- a/src/components/NewskieDialog.tsx
+++ b/src/components/NewskieDialog.tsx
@@ -85,17 +85,32 @@ function DialogInner({
   const isMe = profile.did === currentAccount?.did
 
   const profileName = useMemo(() => {
-    const name = profile.displayName || profile.handle
-
-    if (isMe) {
-      return _(msg`You`)
-    }
-
-    if (!moderationOpts) return name
+    if (!moderationOpts) return profile.displayName || profile.handle
     const moderation = moderateProfile(profile, moderationOpts)
+    return sanitizeDisplayName(
+      profile.displayName || profile.handle, 
+      moderation.ui('displayName')
+    )
+  }, [moderationOpts, profile])
 
-    return sanitizeDisplayName(name, moderation.ui('displayName'))
-  }, [_, isMe, moderationOpts, profile])
+  // Generate the join message based on the conditions
+  const getJoinMessage = () => {
+    const timeAgoString = timeAgo(createdAt, now, {format: 'long'})
+    
+    if (isMe) {
+      if (profile.joinedViaStarterPack) {
+        return _(msg`You joined Bluesky using a starter pack ${timeAgoString} ago`)
+      } else {
+        return _(msg`You joined Bluesky ${timeAgoString} ago`)
+      }
+    } else {
+      if (profile.joinedViaStarterPack) {
+        return _(msg`${profileName} joined Bluesky using a starter pack ${timeAgoString} ago`)
+      } else {
+        return _(msg`${profileName} joined Bluesky ${timeAgoString} ago`)
+      }
+    }
+  }
 
   return (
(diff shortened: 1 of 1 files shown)


surfdude29:
@mozzius Thanks for taking a look at this. I'm not sure why the lockfile check was failing, but I've just synced the branch to main, so hopefully it will be good now 👍

surfdude29:
@mozzius Claude rebased the file for me, and after a few hiccups it now passes CI ✅

src/components/NewskieDialog.tsx
```diff
@@ -85,17 +85,36 @@ function DialogInner({
...
-    const name = profile.displayName || profile.handle
+    if (!moderationOpts) return profile.displayName || profile.handle
+    const moderation = moderateProfile(profile, moderationOpts)
+    return sanitizeDisplayName(
+      profile.displayName || profile.handle,
+      moderation.ui('displayName'),
+    )
+  }, [moderationOpts, profile])
+
+  // Generate the join message based on the conditions
```
mozzius:
```suggestion
```

surfdude29:
No worries, thanks for reviewing it! :)

PR #8014, opened 2025-03-17
New `ContextMenu` menu type for DM messages

mozzius:
Motivation: add a place to add emoji reactions (will be stacked on top of this PR)

This is a new kind of menu, similar to `Menu` on web (and in fact is just a `Menu` on web). On native, it presents a long press menu where the trigger is highlighted and a compact menu presented beneath. 

Double tap will eventually be used to _just_ show the reaction bar, but for now it also opens the menu.

## iOS

https://github.com/user-attachments/assets/35ce98d5-d3e9-4b70-95e9-f08a877ed9d6

## Android

https://github.com/user-attachments/assets/29347f05-78af-4d18-99a3-02b6ce608779

## How it works

The trigger has a RNGH gesture detector on it. When the gesture triggers, it measures the size/position of the trigger element, and renders a PNG of the element using `react-native-view-shot`. This image is then portal-ed to root, along with the menu itself (and a blur background). All elements are then positioned based on the measurements. A single shared value with a spring animation drives the fade/scale/translation animations of all the elements, so that everything is in sync.

## Stretch goal

It would be nice if the menu could be used by just keeping your finger pressed and then moving it over an option and lifting, like on native iOS menus. However, I don't see an obvious way to achieve this, will probably require RNGH hacks

## Test plan

Test iOS/Android. Test long press, double tap. Confirm you can use the menu and dismiss it. Confirm the menu is _always_ on screen, no matter the position/length of the message. Test android perf on device, and check safe areas on android (can be finnicky). Confirm a11y is ok.
Confirm web is unchanged.

### DIFF
--- a/src/components/ContextMenu/index.tsx
+++ b/src/components/ContextMenu/index.tsx
@@ -389,6 +389,8 @@ export function Outer({
               t.atoms.bg_contrast_25,
               a.w_full,
               // @ts-ignore react-native-web expects string, and this file is platform-split -sfn
+              // note: above @ts-ignore cannot be a @ts-expect-error because this does not cause an error
+              // in the typecheck CI - presumably because of RNW overriding the types
               {
                 transformOrigin:
                   align === 'left' ? [0, 0, 0] : [MENU_WIDTH, 0, 0],
--- a/src/components/ContextMenu/index.tsx
+++ b/src/components/ContextMenu/index.tsx
@@ -388,7 +388,7 @@ export function Outer({
               a.shadow_md,
               t.atoms.bg_contrast_25,
               a.w_full,
-              // @ts-expect-error react-native-web expects string, and this file is platform-split -sfn
+              // @ts-ignore react-native-web expects string, and this file is platform-split -sfn
               {
                 transformOrigin:
                   align === 'left' ? [0, 0, 0] : [MENU_WIDTH, 0, 0],
--- a/src/components/ContextMenu/index.tsx
+++ b/src/components/ContextMenu/index.tsx
@@ -1,0 +1,374 @@
+import React, {useMemo, useRef, useState} from 'react'
+import {
+  Keyboard,
+  Pressable,
+  StyleProp,
+  useWindowDimensions,
+  View,
+  ViewStyle,
+} from 'react-native'
+import {Gesture, GestureDetector} from 'react-native-gesture-handler'
+import Animated, {runOnJS} from 'react-native-reanimated'
+import {useIsFocused} from '@react-navigation/native'
+import flattenReactChildren from 'react-keyed-flatten-children'
+
+import {HITSLOP_10} from '#/lib/constants'
+import {useHaptics} from '#/lib/haptics'
+import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
+import {atoms as a, useTheme} from '#/alf'
+import {
+  Context,
+  ItemContext,
+  useContextMenuContext,
+  useContextMenuItemContext,
+} from '#/components/ContextMenu/context'
+import {
+  ContextType,
+  ItemIconProps,
+  ItemProps,
+  ItemTextProps,
+  Measurement,
+  TriggerProps,
+} from '#/components/ContextMenu/types'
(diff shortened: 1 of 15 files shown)


src/components/ContextMenu/index.tsx
```diff
@@ -0,0 +1,570 @@
...
+          left: measurement.x,
+          width: measurement.width,
+          height: measurement.height,
+        },
+        a.z_10,
+        a.pointer_events_none,
+        animatedStyles,
+      ]}>
+      <Image
+        onDisplay={onDisplay}
```
estrattonbailey, line 282: What would happen if the image fails to load for some reason? Is that possible?

src/components/ContextMenu/index.tsx
```diff
@@ -0,0 +1,570 @@
...
+              platform({
+                default: 0,
+                android: topInset, // not included in measurement
+              }),
+            width,
+            height,
+          }),
+        )
+      }),
+      captureRef(ref, {result: 'data-uri'}),
```
estrattonbailey, line 165: Similar question, any chance of anything in here throwing? Could we bail on the trigger clone view and just show the menu in that case?

src/components/ContextMenu/index.tsx
```diff
@@ -0,0 +1,570 @@
...
+  return (
+    <Pressable
+      {...rest}
+      accessibilityHint=""
+      accessibilityLabel={label}
+      onFocus={onFocus}
+      onBlur={onBlur}
+      onPress={e => {
+        context.close()
+        onPress?.(e)
```
estrattonbailey, line 456: I don't think [this](https://github.com/bluesky-social/social-app/blob/2038d229c869d1ff2adf259e868810df688d9f08/src/components/Menu/index.tsx#L131) is an issue here, just calling out in case we notice anything interfering with, like, the report dialog.

src/components/ContextMenu/index.tsx
```diff
@@ -0,0 +1,570 @@
...
+      fill={
+        disabled
+          ? t.atoms.text_contrast_low.color
+          : t.atoms.text_contrast_medium.color
+      }
+    />
+  )
+}
+
+export function ItemRadio({selected}: {selected: boolean}) {
```
estrattonbailey, line 524: Do you think we'll need this component in the future?

src/components/ContextMenu/types.ts
```diff
@@ -0,0 +1,97 @@
...
+   * For example, if the trigger is a message bubble, use the message content.
+   *
+   * @platform ios, android
+   */
+  contentLabel: string
+  hint?: string
+  role?: AccessibilityRole
+  style?: StyleProp<ViewStyle>
+}
+export type TriggerChildProps =
```
estrattonbailey, line 51: If any of these props aren't used or are leftover from the `Menu`, can pry remove for clarity.

estrattonbailey (commented):
This is excellent!

Couple comments. Also, posts have a little bit of space on top and bottom on Android, and top on iOS.
CleanShot 2025-03-19 at 10 52 20@2x
CleanShot 2025-03-19 at 10 54 02@2x

mozzius:
Ah that's fixable. I think it's because QPs have vertical margin

src/components/ContextMenu/index.tsx
mozzius: I can add handling for that, sure. I don't think it's possible but no harm

src/components/ContextMenu/index.tsx
mozzius: Yeah, this isn't the same `control` as the Dialogs, it's just simple useState state. should be fine

src/components/ContextMenu/index.tsx
mozzius: Unlikely, kept for parity with Menu

src/components/ContextMenu/types.ts
mozzius: I'll keep them, because it's a `Menu` on web. Therefore the types need to apply to both

estrattonbailey (approved):

PR #8020, opened 2025-03-18
Allow selecting `ContextMenu` options via press-and-hold

mozzius:
Stacked on #8014

If the context menu is opened via a long press, you can now keep your finger down, move it over a menu item, and release to select.

https://github.com/user-attachments/assets/4ca58633-de0c-488b-8fb6-ef90e8b375f0

## Test plan

Test long pressing to select stuff. Ensure other functionality is not impacted
Ensure it works when the menu moves around, ensure hover zone locations are accurate

### DIFF
--- a/src/components/ContextMenu/context.tsx
+++ b/src/components/ContextMenu/context.tsx
@@ -1,8 +1,14 @@
 import React from 'react'
 
-import type {ContextType, ItemContextType} from '#/components/ContextMenu/types'
+import {
+  type ContextType,
+  type ItemContextType,
+  type MenuContextType,
+} from '#/components/ContextMenu/types'
 
 export const Context = React.createContext<ContextType | null>(null)
+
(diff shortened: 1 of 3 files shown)


mozzius:
Tested on iOS and Android

src/components/ContextMenu/context.tsx
```diff
@@ -18,6 +24,18 @@ export function useContextMenuContext() {
   return context
 }
 
+export function useContextMenuMenuContext() {
```
estrattonbailey, line 27: how dare you

mozzius: 😅

estrattonbailey (approved):
This is inspired stuff my dude. Approach looks sound, poked around a bit to grok and tested on both my physical devices and simulators. LGTM!

PR #8022, opened 2025-03-19
Add video volume to persisted state

### LINKED ISSUE #7124: Videos: please store the selected volume instead of always setting it to 100%
### Describe the Feature

On other websites (Youtube, Twitter), video players set a cookie containing the volume of the video player. When the user watches a video and sets the volume to e.g. 40%, the volume of the next video that user will open is also 40%.

On bluesky, all videos always start with 100% volume which usually hurts my ears cause I have set my system up in a way which requires videos from browser to be at ~40%.

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

brw:
Store video volume state in localStorage, just like Twitter. Closes #7124 

~~Not entirely sure if it's necessary but I've also separated the volume and mute state Contexts and only included the mute state Provider on native as it doesn't use in-player volume controls. I've also moved them to `#/state/video` as that seems to be where everything that persists state is located, but let me know if that's not the right place.~~ No longer the case

Tested on web and Android.

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/VideoVolumeContext.tsx
+++ b/src/components/Post/Embed/VideoEmbed/VideoVolumeContext.tsx
@@ -1,17 +1,19 @@
 import {createContext, useContext, useMemo, useState} from 'react'
 
+import {device, useStorage} from '#/storage'
+
 const Context = createContext<{
   muted: boolean
-  setMuted: React.Dispatch<React.SetStateAction<boolean>>
+  setMuted: (v: boolean) => void
   // web
   volume: number
-  setVolume: React.Dispatch<React.SetStateAction<number>>
+  setVolume: (v: number) => void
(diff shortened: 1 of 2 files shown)


brw:
Ah I see. Something like this?

mozzius:
Yeah lovely, thanks. Will look at this when I have time

brw:
@mozzius hi! it's been a while and I've made the change much smaller now. Have you had any time yet to take a look at it?

Is `#/storage` still the place for this or has that since changed? I see in the recently added [CLAUDE.md](https://github.com/bluesky-social/social-app/blob/97083bb55bf919243e227f108f60d31aef777c2e/CLAUDE.md?plain=1#L55) for example that `#/state/persisted` is still defined as the persistent storage layer.

Issue #8019, opened 2025-03-18
Horizontal scroll on "similar accounts" carousel broken on Android


fongandrew:
### Steps to Reproduce

1. Go to profile page for someone
2. Scroll around until you see a "Similar Accounts" carousel
3. Try to swipe left or right

### Attachments

This doesn't scroll

![Image](https://github.com/user-attachments/assets/28a3d63d-6a38-4d68-afee-6d03782a8f9d)

### What platform(s) does this occur on?

Android

### Device Info

Pixel 7a, Android 15

### What version of the app are you using?

Build version: 1.99.0.621; Bundle info: 41d92f585 (prod); Bundle date: 25031722; Platform: android; Platform version: 35; Anonymous ID: b4dcef38-6e92-42bb-9713-233e34236096

PR #8036, opened 2025-03-21
Fix suggested follows scrolling

estrattonbailey:
We received a report that this section was not scrollable whatsoever on Android. This appears to fix that, just simply by wrapping the section in an additional `View`.

Additionally, in our prod app on iOS, it's possible to trigger the drawer gesture if you press, wait a sec to being you swipe, and then swipe. Immediate swipes work fine. This additional `View` also appears to fix that 🤔 

I haven't investigated why. Let's verify this works. To test, check out the tip of this branch (the fix), or the commit just prior (failure).

Fixes https://github.com/bluesky-social/social-app/issues/8019

### DIFF
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -360,35 +360,37 @@ export function ProfileGrid({
         </View>
       ) : (
         <BlockDrawerGesture>
-          <ScrollView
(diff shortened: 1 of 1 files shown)


haileyok (approved):
wtf

PR #8037, opened 2025-03-21
Instrument signup

mozzius:
New instrumentation for signup

  - Replaced logEvent with logger.metric across the signup flow
  - Added instrumentation for signup duration tracking
  - Added tracking for field-specific validation errors
  - Added detection of app backgrounding/foregrounding during signup
  - Added dedicated tracking for username-related errors
  - Updated metrics schema with new event types

# Test plan

Create account. make sure to make various errors in the fields as you go. then check the logs and ensure:
- Each time you made an error, it was logged
- On success, the time it took and the the number of backgrounds and errors should be logged

### DIFF
--- a/src/logger/metrics.ts
+++ b/src/logger/metrics.ts
@@ -51,6 +51,17 @@ export type MetricEvents = {
   }
   'signup:captchaSuccess': {}
   'signup:captchaFailure': {}
+  'signup:fieldError': {
+    field: string
+    errorCount: number
+    errorMessage: string
+    activeStep: number
+  }
+  'signup:backgrounded': {
+    activeStep: number
+    backgroundCount: number
+  }
+  'signup:handleTaken': {}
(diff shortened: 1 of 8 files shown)


estrattonbailey (commented):
Do any of these metrics need to continue being sent to StatSig? For instance, `account:create:success` we may want in a/b testing.

Looks good otherwise, ran it on iOS.

estrattonbailey (commented):
These seem like for sure should go to statsig:
- `signup:backgrounded`
- `account:create:success`

These maybe?
- `signup:fieldError`
- `signup:nextPressed`

mozzius:
Ah yeah - didn't consider that changing to `logger.metric` no longer sends this data to statsig. I've changed all of these to statsig, they're all important ones

haileyok (approved):
lgtm

Issue #5618, opened 2024-10-05
Increase Labeler Subscription Limit


Nayelianne:
### Describe the Feature

With the popularity of self-assigned labels for identification (eg. pronouns, professions and hobbies, game dev related labels, time zones, countries, zodiac, etc), many people are now subscribing to a bigger amount of them than before, to the point that the limitation of 20, when combined to the real moderation ones, might not be enough for some users anymore.

I understand that limit was already raised from 9 to 20, but it would be nice if it could be increased a little further so people can still have fun while still having space for the legit moderation ones.

Specially since many of the "fun" labelers, like the yardcrow and game dev ones, can be quite useful when browsing the feed and conversations, as they allow people to quickly recognize others that work in the same field, or fields they may be interested in.
I imagine as the service grows, many other useful labelers will also arise, putting more people in the position of having to decide which ones they'll have to sacrifice.

HappyDevs1:
Is this still a feature request, can I work on this?

surfdude29:
@HappyDevs1 I'm not sure there's too much to work on with this feature request tbh? The maximum number of labelers is controlled by a constant set here:

https://github.com/bluesky-social/social-app/blob/57d91373f713e9f13628128bd4db50908dcc3f97/src/lib/constants.ts#L176

So increasing the labeler limit is simply a matter of changing that constant. (The associated strings would need to be updated too, of course).

You could open a PR to do that. I think most of the work with this would probably be in persuading the Bluesky team that it's a beneficial change, rather than coding :)

(Fwiw, I do support an increase.)

PR #8044, opened 2025-03-23
Increase Labeler Subscription Limit

Teqed:
This is a draft PR for tracking the changes required for #5618 "Increase Labeler Subscription Limit".

Currently no decision has been made to increase the limit, or to what it would be increased to if it were, but these are the relevant files if the constant and strings need to be changed. I've used 30 for this draft.

During the build process the relevant `.po` localization files should be generated and ready for translation on Crowdin from what I've read in [the localization instructions.](https://github.com/bluesky-social/social-app/blob/main/docs/localization.md)

EDIT 03/28/25: For clarity, the included changes would allow the Bluesky client to subscribe to additional labels and include them on requests, but the Bluesky Appview will only respond with 20 labelers. E.g. this is only one half of the required changes.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -173,7 +173,7 @@ export const GIF_SEARCH = (params: string) =>
 export const GIF_FEATURED = (params: string) =>
   `${GIF_SERVICE}/tenor/v2/featured?${params}`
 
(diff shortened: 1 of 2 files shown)


mozzius:
Be aware this is a backend limit

PR #8052, opened 2025-03-24
Make the right sidebar smaller when offset, reduce total offset

mozzius:
Aim: reduce the intensity of the offset when in the tablet breakpoint
Solution: reduce the width of the sidebar when in that breakpoint, so less offset is required to keep things centered. Also, slightly reduce the range of the breakpoint, so that the min-sized right nav doesn't reach the edge of the screen

# Before

https://github.com/user-attachments/assets/03cb90d1-1379-4167-82f5-55a577b82142

# After

https://github.com/user-attachments/assets/6088b5a3-a22a-4640-9c6c-1a5a6735eabc


pfrazee:
Yeah much better

pfrazee (approved):

Issue #8050, opened 2025-03-24
BUG 🐜Notifications rendering timestamp


PSingletary:
### Steps to Reproduce

1.  go to bsky.app/notifications
2. look at top most notificaiton (grouped likes)
3. cursor over time (now / #s / #m / #h
4. the pop-up rendering the timestamp is hidden under the next notification below it

![Image](https://github.com/user-attachments/assets/09246aa7-f18b-46d0-8d4c-56a0c07a425d)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

macOS 15.3.2, Edge 134.0.3124.83 (Official build) (arm64)

### What version of the app are you using?

Build version: 1.99.0; Bundle info: 41d92f5 (prod); Bundle date: 25031722; Platform: web; Platform version: undefined; Anonymous ID: 6bb9cb9a-e0d3-4679-a53d-05f98480dd01

PR #8058, opened 2025-03-25
fix: notification timestamp tooltip is hidden

omar0404:
#8050


mozzius:
This is the issue https://github.com/bluesky-social/social-app/pull/4899

omar0404:
> This is the issue #4899

it's related to post overflow not notification so i think my fix will not revert it

mozzius:
well, that's the PR that added the line you're removing according to git blame. have you tested the issue from that old PR? is the fix still necessary?

omar0404:
> well, that's the PR that added the line you're removing according to git blame. have you tested the issue from that old PR? is the fix still necessary?

yes

PR #8061, opened 2025-03-26
The lab

pfrazee:
A new settings screen which

- Gives us the ability to describe feature gates to users
- Gives users the ability to enable or disable the feature gate on a device using a local-storage override
- Adds feedback tools for a given gate

Is built around featuregates (see changes to gates.ts, which is now gates.tsx)

**NOTE** before merging we need to alter `gates.tsx` so that the test gates aren't included in Labs

https://github.com/user-attachments/assets/effbef89-5c39-49b1-b59a-946371af41e2

### DIFF
--- a/src/screens/Settings/LabSettings/FeatureGateDialog.tsx
+++ b/src/screens/Settings/LabSettings/FeatureGateDialog.tsx
@@ -1,0 +1,175 @@
+import {useState} from 'react'
+import {View} from 'react-native'
+import {msg, Trans} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+
+import {type Gate, useGateDescriptions} from '#/lib/statsig/gates'
+import {logger} from '#/logger'
+import {atoms as a, useTheme} from '#/alf'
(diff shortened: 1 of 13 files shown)


src/screens/Settings/LabSettings/FeatureGateDialog.tsx
```diff
@@ -0,0 +1,215 @@
...
+          {enabled !== originalEnabled ? (
+            <Button
+              variant="solid"
+              color="primary"
+              size="large"
+              onPress={() => {
+                setIsRestarting(true)
+                if (isWeb) {
+                  /* @ts-ignore web only */
+                  navigation.reload()
```
mozzius, line 103: `window.navigation` is not available on Firefox/Safari. better to use the old fashioned `location.reload()`

```suggestion
                  location.reload()
```

mozzius:
Good stuff! UI-wise, I feel like the dialog needs a close/done button - there's nothing obvious to press once you've filled out the form

PR #8063, opened 2025-03-26
Fix loading jumps and footer on feeds tab

mozzius:
Use `isPending` rather than `isFetching` for skeleton state to prevent jumps. Also use proper `<ListFooter>`.

I also applied the same fix to the lists tab and the starter pack tab, but I don't know of any accounts with a ton of lists to test with.

# Before

https://github.com/user-attachments/assets/ec1f5c50-27db-4ecc-84f5-b901f8fad1e9

# After

https://github.com/user-attachments/assets/2c3f0a09-8d7e-44ec-b583-582251636061

# Test plan

Scroll the feeds on `@trending.bsky.app`'s profile

### DIFF
--- a/src/view/com/lists/ProfileLists.tsx
+++ b/src/view/com/lists/ProfileLists.tsx
@@ -1,29 +1,27 @@
 import React from 'react'
 import {
-  ActivityIndicator,
   findNodeHandle,
-  ListRenderItemInfo,
-  StyleProp,
-  StyleSheet,
+  type ListRenderItemInfo,
+  type StyleProp,
   View,
-  ViewStyle,
+  type ViewStyle,
 } from 'react-native'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useQueryClient} from '@tanstack/react-query'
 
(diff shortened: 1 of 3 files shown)


haileyok (approved):
looks good

src/view/com/lists/ProfileLists.tsx
```diff
@@ -212,6 +211,6 @@ export const ProfileLists = React.forwardRef<SectionRef, ProfileListsProps>(
   },
 )
 
-const styles = StyleSheet.create({
-  footer: {paddingTop: 20},
-})
+function keyExtractor(item: any) {
+  return item._reactKey || item.uri
+}
```
haileyok, line 216: yep yep yep

haileyok: (why do we even need the fallback to uri? mostly a question of _what is_ the react key and why would it ever not be present? this is fine just wondering)

haileyok: (i should know this but i haven't written app code in too long...)

PR #8072, opened 2025-03-27
[Explore] Suggested accounts module

mozzius:
# Stacked on #8067 

New suggested accounts module

Unlike the follow dialog, this one does not do more suggestions once you follow someone - just too complex for this first pass. Totally possible to do in a followup.

https://github.com/user-attachments/assets/a9b99f4f-ef68-45b8-a688-6a1547b066d0

# Test plan

Click on the tabs? Make sure the pagination works?

### DIFF
--- a/src/screens/Search/Explore.tsx
+++ b/src/screens/Search/Explore.tsx
@@ -1,13 +1,7 @@
 import {useCallback, useMemo, useRef} from 'react'
 import {View, type ViewabilityConfig, type ViewToken} from 'react-native'
-import {
-  type AppBskyActorDefs,
-  type AppBskyFeedDefs,
-  moderateProfile,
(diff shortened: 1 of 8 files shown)


src/screens/Search/Explore.tsx
```diff
@@ -199,6 +83,10 @@ function LoadMore({
 }
 
 type ExploreScreenItems =
+  | {
+      type: 'topBorder'
+      key: string
+    }
```
mozzius, line 89: Kinda extraneous, but this gives the search bar header a top border while it's over the trending topics area until the next sticky header replaces it. we'll have to remove the top border from the new trending topics module PR

estrattonbailey (approved):
LGTM

PR #8077, opened 2025-03-28
Design tweaks to applied reactions

estrattonbailey:
I felt it should have a lil more overlap, and stand out a little better, so I added a subtle bg and shadow.

CleanShot 2025-03-28 at 15 36 22@2x
CleanShot 2025-03-28 at 15 50 38@2x


mozzius (approved):
I'd maybe go with -10, but lgtm. I like the shadow/color

PR #8078, opened 2025-03-28
Put applied reactions inside context menu wrapper on native

estrattonbailey:
I think we can anticipate users trying to long-press on the applied reactions too.

No changes on web, since this would mis-align the menu buttons on hover if reactions are applied. Native now looks like this.
CleanShot 2025-03-28 at 15 55 14@2x


mozzius (approved):
Sure

src/components/dms/MessageItem.tsx
```diff
@@ -147,79 +216,11 @@ let MessageItem = ({
               />
             </View>
           )}
+
+          {isNative && appliedReactions}
```
mozzius, line 220: nit: maybe use the `native()` and `web()` utils instead?

estrattonbailey: I suppose we could do like `web(appliedReactions)`, but I don't think we've used them like that before. Gonna roll with this since I know it works 👍

PR #8093, opened 2025-04-01
[Reactions] Get reactions in chat list working again with new lexicon

mozzius:
For backwards compatibility, we removed `messageAndReactionView` from the union in `lastMessage` and added a new field `lastReaction`. We now have to compare the sentAt/createdAt dates to decide which one to show.

Also includes a minor android style fix on the reaction palette

# Test plan

Send messages and add/remove reactions in various orders and confirm the chat list is looking as it should
Confirm the log updates work correctly - esp. removing reactions

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "icons:optimize": "svgo -f ./assets/icons"
   },
   "dependencies": {
-    "@atproto/api": "^0.14.14",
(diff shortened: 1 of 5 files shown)


haileyok (approved):
sent/received a bunch of these on outdated and up-to-date clients. seems to work good.

PR #8095, opened 2025-04-01
[Explore] Starter Packs

estrattonbailey:
CleanShot 2025-04-02 at 15 35 38@2x
CleanShot 2025-04-02 at 15 35 54@2x

### DIFF
--- a/src/state/queries/useSuggestedStarterPacksQuery.ts
+++ b/src/state/queries/useSuggestedStarterPacksQuery.ts
@@ -1,5 +1,4 @@
 import {useQuery} from '@tanstack/react-query'
-import {AppBskyGraphDefs} from '@atproto/api'
 
 import {useAgent} from '#/state/session'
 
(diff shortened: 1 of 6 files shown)


src/state/queries/useSuggestedStarterPacksQuery.ts
```diff
@@ -0,0 +1,38 @@
...
+]
+
+export function useSuggestedStarterPacksQuery() {
+  const agent = useAgent()
+  const {data: preferences} = usePreferencesQuery()
+  const contentLangs = getContentLanguages().join(',')
+
+  return useQuery({
+    refetchOnWindowFocus: true,
+    staleTime: STALE.MINUTES.THREE,
```
haileyok, line 23: small note, the server caches this for 10 minutes. we can lower there if we want to (remember, it is random responses right now. if we want to drop the cache to like 1-3 minutes so the users sees "more" random sps that makes sense)

haileyok:
:pinched_fingers:

haileyok (approved):

src/state/queries/useSuggestedStarterPacksQuery.ts
estrattonbailey: Ah good call. Let's do a pass as a follow up 👍

estrattonbailey: Er well I gotta fix formatting anyway so I'll bop this

PR #8096, opened 2025-04-01
Move SENTRY_DSN to secrets

estrattonbailey:
Disables Sentry in dev to prevent accidental reporting by contributors who fork this repo, and moves the Sentry DSN into an environment variable to better avoid abuse or accidental misuse.

A new DSN will need to be populated here in GitHub for CI.


haileyok:
(should run a web build once we update secrets/merge to make sure that docker file is good)

PR #8110, opened 2025-04-03
Remove 'All' category and default to top interest

estrattonbailey:


estrattonbailey:
I think we're going to keep this and repurpose the tab with new search data

PR #8120, opened 2025-04-04
[Explore] Add flatlist perf-related props

mozzius:
Copied from `PostFeed`, I assume those numbers will works here too


src/screens/Search/components/ModuleHeader.tsx
```diff
@@ -41,6 +41,8 @@ export function Container({
         headerHeight && web({position: 'sticky', top: headerHeight}),
         style,
       ]}>
+      {/* Very non-scientific way to avoid small gap on scroll */}
+      <View style={[a.absolute, a.inset_0, t.atoms.bg, {top: -2}]} />
```
estrattonbailey, line 45: Meant to add this to a diff branch but seems harmless to include here too

estrattonbailey (approved):
Much better on my iPhone!

PR #8121, opened 2025-04-04
[Explore] Reduce scroll glitchiness on iOS

mozzius:
# Stacked on #8119 (that PR moves a file)

FlatList's sticky header implementation seems to struggle with having multiple sticky items onscreen at once. Removing use of `topBorder` from between feeds, which was only intended to be used at the very top of the Explore page, cuts down on scroll glitchiness significantly.

# Test plan

Compare to main:
- General scroll perf on iOS
- Whether it does the whiteout when fast scrolling
- Whether it gets jumpy when you manually drag the scrollbar
- Layout (I tried to get it looking the same as before)

### DIFF
--- a/src/screens/Search/Explore.tsx
+++ b/src/screens/Search/Explore.tsx
@@ -766,6 +766,9 @@ export function Explore({
           )
         }
         // feed previews
+        case 'preview:spacer': {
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Looks good just needs the rebase 👍

PR #8130, opened 2025-04-04
[Explore] New suggested follows endpoint

estrattonbailey:


src/screens/Search/Explore.tsx
```diff
@@ -203,33 +201,13 @@ export function Explore({
   const gate = useGate()
   const guide = useProgressGuide('follow-10')
   const [selectedInterest, setSelectedInterest] = useState<string | null>(null)
+  // TODO always get at least 10 back
```
estrattonbailey, line 204: Just want to test what happens when you follow a handful of users and refresh the page

haileyok (approved):
tested a bunch, this looks good. one thing we could probably do is limit number that come down in response, but that's a backend thing we can do.

src/screens/Search/Explore.tsx
haileyok: not quite sure what your todo was here. im going to merge though for now.

PR #8150, opened 2025-04-07
Update patches after #8133

estrattonbailey:
- The patch from #5605 doesn't seem to apply very clearly to RN 0.76.9 in this PR but the haptic still works and I haven't found any other degradations
- Not sure why picker was downgraded, but it was upgraded seemingly arbitrarily in #6887, so I don't know that we need that minor version


estrattonbailey:
Closing bc of a naming issue with `/patches` and this branch name

PR #8153, opened 2025-04-08
Various explore fixes: old type definition and rare header alignment issue

pfrazee:
1. Old type definition: this is a metric type definition change I missed in a previous PR 6e26670cf36e33a832347f2d5deda508dcf6d08a
2. Removed position: sticky from the feed headers (more below) 706db5d8920ca905e46102bb23f5e040512726d2

There is a case in Chrome, which we believe to be on certain pixel densities, where certain window widths will trigger an incorrect layout:

image

This can be triggered by a `position: sticky` inside of a `margin: 0 auto`. The one clue I found is that chrome uses integers for margin calculations, and so my theory is that `position: sticky` is causing the element positioning to get calculated at a different precision. This theory is backed up by the fact that the condition occurs when the margin calculation comes out to `.5` pixels.

Unfortunately this is an issue that hits us anywhere we use `position: sticky`, but for now I'm just fixing it here

### DIFF
--- a/src/logger/metrics.ts
+++ b/src/logger/metrics.ts
@@ -330,7 +330,6 @@ export type MetricEvents = {
       | 'suggestedFeeds'
       | 'suggestedStarterPacks'
       | `feed:${FeedDescriptor}`
-    index: number
   }
   'explore:module:searchButtonPress': {
     module: 'suggestedAccounts' | 'suggestedFeeds'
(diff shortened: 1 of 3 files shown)


estrattonbailey:
Closing in favor of #8154 — confirmed this half pixel issue only shows on low DPI screens, hence why most of us haven't noticed on our laptops or desktop setups (4k or higher). We will fix this later.

Issue #7685, opened 2025-02-06
Message to our translators: Crowdin is now integrated!


pfrazee:
Hey everybody!

Thanks so much for being patient with me. I've finally integrated Crowdin, and am ready to work this into our flow.

- **Crowdin Project: https://crowdin.com/project/bluesky-social**
- **Invite link to join: https://crowdin.com/project/bluesky-social/invite?h=caac48a68f4fcd62d285b403469336d02341845** (expires in 1 month)
- **Introduction to Crowdin: https://support.crowdin.com/for-translators/**

This update follows https://github.com/bluesky-social/social-app/issues/7317. Changes to the process can be found in this PR: https://github.com/bluesky-social/social-app/pull/7684.

Here's a summary of the changes:

## Using Crowdin

[Crowdin](https://crowdin.com/project/bluesky-social) will be our primary tool for managing translations. There are two roles:

- **Proof-readers**. Can create new translations and approve submitted translations.
- **Translators**. Can create new translations.

All translations must be approved by proof-readers before they are accepted into the app.

## Using other platforms

You may contribute PRs separately from Crowdin, however we strongly recommend using Crowdin to avoid conflicts.

## Code of conduct on Crowdin

Please treat everyone with respect. Proof-readers are given final say on translations. Translators who frequently come into conflict with other translators, or who contribute noticably incorrect translations, will have their membership to the Crowdin project revoked.

## Adding a new language

Create a new [Crowdin discussion](https://crowdin.com/project/bluesky-social/discussions) or [GitHub issue](https://github.com/bluesky-social/social-app/issues) requesting the new language be added to the project.

Please only request a new language when you are certain you will be able to contribute a substantive portion of translations for the language.

## Becoming a translator on Crowdin

[This invite link](https://crowdin.com/project/bluesky-social/invite?h=caac48a68f4fcd62d285b403469336d02341845) will add you to the project as a translator. The link will expire in 1 month. Not sure yet how it'll work after that link expires; still learning how crowdin works.

## Becoming a proof-reader on Crowdin

Proof-readers are community translators who have established their reliability through contributions. We already have quite a few who meet that expectation!

If you wish to be a proof-reader, please reply to this issue with your Crowdin username. I will evaluate your contribution history and make the call on whether I'm comfortable giving that permission.

Per the Code of Conduct, if I get the sense that a proof-reader is not behaving responsibly, I'll revoke their role.

## Kick-off

[...]

PR #8176, opened 2025-04-10
Corrected Turkish translations and added new translations

Yigit4416:
Generally in Turkish people use "feed" not "gönderi" and **most definitely not** "besleme" which used for sentence like "feeding the pet" or things like that so i used "feed". There are still lots of work to do.

### DIFF
--- a/src/locale/locales/tr/messages.po
+++ b/src/locale/locales/tr/messages.po
@@ -1992,7 +1992,7 @@ msgstr "Sohbetten ayrılınamadı"
 
 #: src/screens/Profile/ProfileFeed/index.tsx:80
 msgid "Could not load feed"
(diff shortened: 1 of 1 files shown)


surfdude29:
Hello, I'm a volunteer contributor and I help out with localization. I wanted to let you know that while localization PRs are still accepted, it's strongly recommended to use Crowdin:

> You may contribute PRs separately from Crowdin, however we strongly recommend using Crowdin to avoid conflicts.

from:

- #7685

This issue has lots of useful information for those interested in helping to localize Bluesky :)

You can join the Crowdin here, and suggest and vote for translations:

https://bluesky.crowdin.com/bluesky-social

Issue #898, opened 2023-06-24
[Android] F-Droid release


joe733:
### Is your feature request related to a problem? Please describe.

Yes, can't download android app without Google Playstore.

### Describe the solution you'd like

Bluesky is made available through [F-Droid](https://f-droid.org/).

### Describe alternatives you've considered

[IzzyOnDroid](https://apt.izzysoft.de/fdroid/index/info)?

### Additional context

`F-droid release === Peace of mind :)`

dmytri:
update?

gagzzz:
Hello,
I'd be also interested since I have a bluesky account but still cannot use it easily on my phone (no google things here). 

Also I tried with [Obtainium](https://github.com/ImranR98/Obtainium) but it requires published Releases, and bluesky-social/social-app only has one tag.

Please publish releases :)

Thank you!

RasheedAZ:
Are the developers okay with submitting this app for inclusion the F-Droid main repo? 

This is an open source app built on an open protocol... so would be great if it was available on FOSS only app stores like F-Droid.

Thanks :)

yozachar:
> F-droid release === Peace of mind :)

I'm re-thinking about that, after I read https://github.com/GrapheneOS/Camera/issues/227. It's been more than a year, since the last comment was made there. I do not know if the state of things has improved or not.

RasheedAZ:
This really is not the place for it and I wish you had not linked to that thread. Many of us are aware of all those issues, F-Droid's choices and reasons, how apps are packaged in Linux in general, and other personal issues surrounding GOS dev and would still like FOSS apps to be available on FOSS-only platforms and app stores if possible. Besides F-Droid, there are no FOSS-only app stores for Android (other than its forks which are not as well developed). 

For every single one of their criticisms of F-Droid there is a good reply whether they agree with it or not and the accusat

It is good that GOS prioritises security over privacy, decentralisation and avoidance of Big Tech, etc. but different people have different threat models and may prioritise things differently.

haileyok:
@codenyte I responded in the GitLab repo in this comment! https://gitlab.com/fdroid/rfp/-/issues/2732#note_1931850811

I'll also make sure Obtanium works as well in the meantime. You can (or the F-Droid maintainers) can reach out either on GitLab, here, or via email if anything else is needed.

haileyok:
[...]

PR #8178, opened 2025-04-11
Add Fastlane folder for F-Droid inclusion

chrrybmb64:
for #898 

F-Droid strongly recommends that upstream app repositories contain the app's metadata in Fastlane folder structure. This pull request adds that. Other languages can be added, but only English is required.

All of the images and descriptions were taken from this repo and Bluesky's page on the Google Play Store.


xty:
What's being proposed here is quite straightforward. Now let's see if Bluesky will lift a finger to help move this forward. 👀

PR #8190, opened 2025-04-14
Clear image cache button

mozzius:
`expo-image`'s image disk cache accumulates images indefinitely. On Android, this cache can be flushed in the app settings, but that's not possible on iOS without uninstalling the app. While the OS will flush it for you if it needs the space, this PR adds a button to manually flush it.

I put it in the about settings, next to the version number. if anyone has a better idea where to put it I'm all ears

Screenshot 2025-04-14 at 16 35 49

### DIFF
--- a/src/screens/Settings/AboutSettings.tsx
+++ b/src/screens/Settings/AboutSettings.tsx
@@ -1,29 +1,60 @@
 import {useMemo} from 'react'
 import {Platform} from 'react-native'
 import {setStringAsync} from 'expo-clipboard'
+import * as FileSystem from 'expo-file-system'
+import {Image} from 'expo-image'
 import {msg, Trans} from '@lingui/macro'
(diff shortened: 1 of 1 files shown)


layeredstrange:
Excellent work mozzius.

Conversations about space usage have been increasing.
This might be useful for space saving and might help boost performance.

Since it's a flush media function, wouldn't it be better in the area assigned to media? Just an idea. 😄

mozzius:
I tried it, it looks even more out of place. Its too technical of a button, hence why I put it next to the system log / version number

layeredstrange:
Better than no function to do so then. 😁👍🏻

haileyok (approved):

mozzius:
Better icon

Screenshot 2025-04-17 at 14 53 29

PR #8191, opened 2025-04-14
Modernize in-app browser consent dialog

mozzius:
ALF'd the in-app browser consent dialog. This required creating a "stateful dialog control" - just used for this one control so far, so I'm keeping it colocated. it only really makes sense for these global dialogs. I expect to have to use it for the sketchy link warning dialog too

<table>
  <thead>
    <tr>
      <th>iOS</th>
      <th>Android</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

Unset pref, and activate the dialog by pressing a link. Confirm functionality is the same, and it all works properly

If you can find an instance where we open a link from within another dialog, confirm that works too (I don't know if we do that anywhere, but I tried to ensure it would work)


estrattonbailey (approved):
Made a small tweak to make these buttons more obvious as to what they do. I think this is pretty nice.

CleanShot 2025-04-24 at 16 29 48@2x

PR #8195, opened 2025-04-14
Replace old ProfileCard with new

estrattonbailey:
Finally does away with the old `ProfileCard`. All that's left is `ProfileCardWithFollowButton`, which is just a convenience wrapper for the new `ProfileCard`. Also removes `LikesDialog`, which was never used and relied on an outdated API I don't think we need to port forward to the new card right now.

Testing
- lists
- muted/blocked accounts listing
- your follows, your following user listing screens
- post-liked-by
- search results


mozzius (approved):
Good stuff

PR #8211, opened 2025-04-17
Change `UK` country code to `GB` in additional-moderation-authorities.ts

### LINKED ISSUE #8206: Mandatory labeler for Great Britain is incorrectly applied to Ukraine
### Steps to Reproduce

https://github.com/bluesky-social/social-app/blob/main/src/state/session/additional-moderation-authorities.ts#L27

`UK` is Ukraine, Great Britain is `GB`. https://bsky.app/ipcc in fact does return `GB` for users in Great Britain.

EDIT: I'm not entirely correct, `UK` is *language* code for Ukrainian, while country code is `UA`. `UK` as a country code is reserved and is not in use.

### Attachments

_No response_

Web (Desktop)

### Device Info

_No response_

n/a

### Additional Information

_No response_

surfdude29:
Fixes #8206

As @imax9000 points out in that issue, the [ISO-3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) country code for the United Kingdom is `GB` rather than `UK` (although we only use `.uk` for domain names...)

It might also be a good idea to rename the [labeler account](https://bsky.app/profile/did:plc:gvkp7euswjjrctjmqwhhfzif) handle, to avoid any potential for confusion.

### DIFF
--- a/src/state/session/additional-moderation-authorities.ts
+++ b/src/state/session/additional-moderation-authorities.ts
@@ -6,7 +6,7 @@ import {device} from '#/storage'
 export const BR_LABELER = 'did:plc:ekitcvx7uwnauoqy5oest3hm' // Brazil
 export const DE_LABELER = 'did:plc:r55ow3tocux5kafs5dq445fy' // Germany
 export const RU_LABELER = 'did:plc:crm2agcxvvlj6hilnjdc4hox' // Russia
-export const UK_LABELER = 'did:plc:gvkp7euswjjrctjmqwhhfzif' // United Kingdom
+export const GB_LABELER = 'did:plc:gvkp7euswjjrctjmqwhhfzif' // United Kingdom
 export const AU_LABELER = 'did:plc:dsynw7isrf2eqlhcjx3ffnmt' // Australia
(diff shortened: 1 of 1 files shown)


blowdart:
The problem here is GB is Great Britain. Technically it excludes Northern Ireland.

pfrazee (approved):
Ah, thanks yall

pfrazee:
> The problem here is GB is Great Britain. Technically it excludes Northern Ireland.

We'll need to add that later then

surfdude29:
Although it's a bit counterintuitive, `GB` is in fact the [ISO-3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) code for the whole of the United Kingdom of Great Britain and Northern Ireland, so Northern Ireland isn't excluded :)

IMG_4430

blowdart:
Which is accurate but politically naive 😂

As long as it's not displayed anywhere in a UI it's fine, but if it ends up on screen the U.K. is much better than G.B. to display.

PR #8212, opened 2025-04-17
ALF searchable people list

mozzius:

PR #8213, opened 2025-04-17
Modernise list membership dialog

mozzius:
# Stacked on #8212 

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

- Dialog can be trigger from a user's profile, and from the list screen -> edit button on a member. Test both
- Test functionality add/remove
- Confirm dialog closes itself if you remove a member from the list you're currently on the screen for


src/components/dialogs/lists/UserAddRemoveListsDialog.tsx
```diff
@@ -0,0 +1,390 @@
...
+      if (starterPacks.length > 0) {
+        _items = _items.concat(
+          {
+            type: 'section-header',
+            title: _(msg`Starter packs`),
+            _reactKey: 'referencelist',
+          },
+          starterPacks.map(list => ({type: 'list', list})),
+        )
+      }
```
mozzius, line 130: We don't get reference lists from this query, but it would be nice to in the future! And we're prepared if they start appearing

surfdude29:
This looks so much better! 👌

I have one suggestion for the case where the user has no lists. As well as the `You have no lists.` message, would it be possible to add links to the screens where a user can create lists?

E.g. below the message, maybe it could say:

> You can create [a user list](https://bsky.app/lists) or [a moderation list](https://bsky.app/moderation/modlists)

and the user could tap/click the link and create a list, and then tap back to return to the profile and add the account to the list they've just created.

surfdude29:
Also, @luan-u gave feedback on Crowdin on the current string `Update {displayName} in Lists`:

IMG_5720

and I think he and you in this PR are both right that simply `Add {displayName} in lists` works better, as I suspect the dialog is mostly used to add a user to lists.

mozzius:
Dang it why has github started closing PRs like this rather than changing it to main like it used to!!

mozzius:
> I have one suggestion for the case where the user has no lists. As well as the `You have no lists.` message, would it be possible to add links to the screens where a user can create lists?

Not a bad idea, but out of scope for this PR. Maybe when I ALF the list create dialog

PR #8220, opened 2025-04-18
New Edit Profile dialog on web, use new Edit Image dialog everywhere

mozzius:
# Stacked on #8327 

Web was using the old edit profile modal for three reasons:
1. Nested radix components (dialog/dropdown menu) broke a11y/generally didn't work
2. The web image picker/cropper uses the old modal cropper, which is incompatible with the new dialogs
3. There was a bug where nested dialogs often lost the stacking order they were supposed to be in

#8219 fixes point 1. This PR fixes points 2 and 3. This required stopping the use of `openCropper` on web, and switching to `EditImageDialog`. `EditImageDialog` had to be enhanced to support aspect ratios and circle cropping, which it now does. I also made it much prettier. EditableUserAvatar and EditableBanner now use `EditImageDialog` on web, which was the last place that used the old cropper, which lets me delete it.

Note: I had to change the type signature of the returned image to `ImageMeta` from `RNImage`, and compresses it before returning. This makes it easier to consume, and the simpler type is much the same but works on web.

Bug 3 was fixed by making sure the order of dialogs was preserved when rerendering - the stacking order was implicit, by the order they were inserted. If an underlying dialog rerendered, it would jump to the top of the stack. Fixed by changing `delete map.current[id]` to `map.current[id] = null`.

All this lets us delete 2 legacy modals, and stop using `unstable__openModal` which was super sketchy!

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

- Test edit profile dialog on web. Confirm can change avi/banner
  - Confirm a11y/keyboard nav works
- Test image cropper - confirm can crop, confirm it uses the fixed aspect ratios
- Test cropping where you drag your mouse outside of the dialog before releasing - I hopefully fixed a bug that cause the dialog to close when you do this.
- Confirm native still works
- Test edit list modal - confirm can set/change list avi
- Confirm editing images in the composer still works


src/components/Portal.tsx
```diff
@@ -9,32 +19,32 @@ type ContextType = {
...
       if (map.current[id]) return
-      map.current[id] = <React.Fragment key={id}>{component}</React.Fragment>
+      map.current[id] = <Fragment key={id}>{component}</Fragment>
       setOutlet(<>{Object.values(map.current)}</>)
     }, [])
 
-    const remove = React.useCallback<ContextType['remove']>(id => {
-      delete map.current[id]
+    const remove = useCallback<ContextType['remove']>(id => {
+      map.current[id] = null
```
mozzius, line 43: This is the fix for the portal ordering bug

src/lib/media/picker.web.tsx
```diff
@@ -1,35 +1,18 @@
...
-        } else {
-          reject(new Error('Canceled'))
-        }
-      },
-    })
-  })
+export async function openCropper(_opts: CropperOptions): Promise<RNImage> {
+  throw new Error(
+    'openCropper is not supported on web. Use EditImageDialog instead.',
+  )
```
mozzius, line 17: Idk if we want to make it clearer somehow that these are native-only? Currently typescript can't tell

src/screens/Profile/Header/EditProfileDialog.tsx
```diff
@@ -137,17 +137,16 @@ function DialogInner({
...
       setImageError('')
       if (img === null) {
         setNewUserAvatar(null)
         setUserAvatar(null)
         return
       }
       try {
-        const finalImg = await compressIfNeeded(img, 1000000)
-        setNewUserAvatar(finalImg)
-        setUserAvatar(finalImg.path)
```
mozzius, line 150: This compression step moved to `EditableUserAvatar`

src/lib/media/manip.web.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference lib="dom" />
+
```
haileyok, line 2: jw what does this do

haileyok: guessing for zed types or something?

haileyok (approved):
good stuff. tested on safari, chrome, and firefox, all seem to work great. tested in composer and the various "edit profile/list" dialogs. also verified no regressions on ios/android.

src/lib/media/picker.web.tsx
haileyok: yea this is generally how i handle these in various modules. the only real way i believe to make it more obvious would be to actually change he names of the functions, but that's annoying. any amount of testing should reveal the errors if you do this in dev.

PR #8222, opened 2025-04-18
[Explore] Run refresh queries in parallel

mozzius:
The PTR mechanism has a Promise.all, but it accidentally awaits each of the promises, so it runs sequentially. Removing the awaits runs them in parallel


efstathiosntonas:
@mozzius hi, this is another place were Promise.all is misused: https://github.com/bluesky-social/social-app/blob/c16d5ce8213ac10f0c3530fe97c561c5262b0d27/src/screens/Settings/SettingsInterests.tsx#L112-L120

mozzius:
Nice thanks

Issue #5277, opened 2024-09-11
Seperate the single Autoplay option for Videos AND GIFs into two


RetroSunstar:
Looking at [this post](https://bsky.app/profile/caliburn-ex.bsky.social/post/3l3vqk3wg6c26), having the option to disable autoplay for videos OR gifs instead of both could be a good idea

There are times a user may not want to have videos Autoplay, which they can disable in [settings](https://bsky.app/settings/accessibility) but this causes GIFs to stop autoplaying as well, which seems fairly inconvenient to hit play on every gif just because you don't want videos to autoplay

It's not bad to give the option to users who want it, essepcially with certain types of video content that the user may not want to view which doesn't necessarily extend to gifs 🩵🩷🩵🩷

(If it's not too much to ask???)

pietrofxq:
I agree - videos also consume way more bandwidth than GIFs. I don't mind downloading some gifs but streaming every video passing by the timeline, especially in mobile networks, can be too much

ThatOneUnoriginal:
I would like this since it seems like auto playing videos seems to trip up last.fm when scrolling through the timeline (at least on the Android version.)

PR #8234, opened 2025-04-20
Split the singler autoplay option into two

RetroSunstar:
This is for #5277

It splits the single "Autoplay videos and GIFs" into their own, individual toggles 🐙

(Edited/Updated for the different paths since this originally opened, and made it a bit nicer 🩵🩷🩵🩷)


src/state/persisted/schema.ts
```diff
@@ -168,7 +169,8 @@ export const defaults: Schema = {
   lastSelectedHomeFeed: undefined,
   pdsAddressHistory: [],
   disableHaptics: false,
-  disableAutoplay: PlatformInfo.getIsReducedMotionEnabled(),
+  disableGifAutoplay: PlatformInfo.getIsReducedMotionEnabled(),
+  disableVideoAutoplay: PlatformInfo.getIsReducedMotionEnabled(),
```
haileyok, line 177: Hmm, I wonder if we can make this respect the user's current `disableAutoplay` preference - if it is set - so that we are not overwriting whatever they may have set right now. i think as this stands, it will turn on autoplay for everyone even if they have it turned off right now

haileyok (changes requested):
+1 to this, but we want to make sure we don't overrwrite user's current disabled autoplay preference

src/state/persisted/schema.ts
haileyok: (if i'm missing something pls let me know!!)

haileyok:
heya saw the notif that you pushed...

@mozzius when you have some time, this might be a worth pr (that i selfishly forgot to review when i was still around,...)

Issue #1558, opened 2023-09-27
Feedback on Threaded mode


mschwendener:
Hi

Threaded mode is easier to follow for me (on 13inch laptop) than the earlier version, thank you for that. 

Example https://bsky.app/profile/carlbergstrom.com/post/3k5e6btkslm2e 

But: After the 7th post the thread by the original poster stops, and only a very small "More -> "  leads to some more posts in the thread, which need another click on "More -> " several times (the thread consists of 42 posts). 

I would prefer to see the whole thread first, for example via the collapse/expand feature proposed [here](https://github.com/bluesky-social/social-app/issues/1547). 

If it stays how it is, "More -> " should be bigger, [get a hover state](https://github.com/bluesky-social/social-app/issues/1477), and – if possible – show a bigger number of posts from the original thread before the next "More > " than it does now. 

Thank you for working on this and asking for our opinion. 
 
<img width="555" alt="more" src="https://github.com/bluesky-social/social-app/assets/4497705/26b99b8d-26b3-4f57-bea8-76a698ab5412">

simonetthomas:
I think it's kind of the same thing I talked about here : https://github.com/bluesky-social/social-app/issues/1478
At the time it was not with the threaded mode, but unfortunately it didn't change the fact that only 7 posts are loaded at a time.

And for reading long threads, I actually think it's worse with the threaded mode because it shifts more and more to the right even when it's from the same person which makes no sense to me. I'd prefer to display all the thread aligned on the left, and only the replies by other people shifted to the right.

mschwendener:
Threaded mode got set to 'No' with the latest update? Could the "Yes" setting become sticky?

mschwendener:
The "More > " link is still much too easy to miss because it's so small.

surfdude29:
Does the "More > " link still show up for anyone else since 1.57 rolled out yesterday? I can no longer see it

mschwendener:
You are right: now "More > " has disappereared (web). 

Which is not better than before, because it means all the later posts be the original author of the [thread](https://bsky.app/profile/carlbergstrom.com/post/3k5e6btkslm2e) are no longer part of the thread at all. 

![no_more](https://github.com/bluesky-social/social-app/assets/4497705/c183785e-4262-4501-8241-6e0d4c264aa8)

surfdude29:
I'm guessing this is a bug rather than a design change. Could you clarify please @pfrazee ?

[...]

PR #8236, opened 2025-04-20
Remove "More" on PostThreadItem

gaearon:
I don't understand what this is for. It looks misaligned and it doesn't say anything that isn't already said by the number of replies below. I suspect it started showing up by accident due to some change in the data model.

Can we pleeease kill it.

I nuked both the UI and the viewmodel part for it because it wasn't used by anything else, and simpler = better.

## Before

Screenshot 2025-04-20 at 14 37 31

## After

Screenshot 2025-04-20 at 14 38 09

### DIFF
--- a/src/view/com/post-thread/PostThread.tsx
+++ b/src/view/com/post-thread/PostThread.tsx
@@ -5,7 +5,7 @@ import Animated from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {
   AppBskyFeedDefs,
-  AppBskyFeedThreadgate,
+  type AppBskyFeedThreadgate,
   moderatePost,
 } from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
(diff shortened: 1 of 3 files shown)


src/view/com/post-thread/PostThread.tsx
```diff
@@ -5,7 +5,7 @@ import Animated from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {
   AppBskyFeedDefs,
-  AppBskyFeedThreadgate,
+  type AppBskyFeedThreadgate,
```
gaearon, line 8: not sure what's up with this change, maybe ESLint rule autofix? or is it something on my side? regardless shouldn't hurt.

mozzius: Yep I added an eslint rule

mackuba:
I think the idea was that if someone makes a long self-thread like "1/20", it shows like 6 first posts and then it's unclear that the rest goes on below, so this was added to show that the rest of thread is there?

surfdude29:
I agree the More link doesn't serve a useful purpose when it shows up in a case where the reply/replies aren't actually visible (e.g. due to a block), but it serves a very useful purpose to indicate that there are more replies in threaded mode.

So fixing it rather than removing it entirely is surely a better option?

mschwendener:
> more replies in threaded mode 

Yes, and it is far too easy to overlook https://github.com/bluesky-social/social-app/issues/1558 

Old example https://bsky.app/profile/carlbergstrom.com/post/3k5e6btkslm2e 

New example https://bsky.app/profile/maryrosemuseum.bsky.social/post/3lmtngy3ulc2w

mozzius:
Superceded by #8443 (thank god)

PR #8248, opened 2025-04-21
Make verifier pressable

mozzius:
Also I found another double close button

# Test plan

Open verified dialog, press on a verifier to go to their profile


mozzius:
Superceded by #8249

PR #8256, opened 2025-04-21
Improve display name in Settings

estrattonbailey:
This improves the rendering so that long display names don't overflow the container on web, but does not resolve why on Android, _some_ display names are truncated more than necessary.

No change to verification checks if they exist.

CleanShot 2025-04-21 at 14 56 50@2x


haileyok (approved):
tested on web and android, works

PR #8262, opened 2025-04-22
introduce context for "Following" as the feed name.

ssg:
"Following" has different translations as the name of the main feed and the state of following someone in languages like Turkish. This allows "Following" as the feed name to be translated correctly.

This may introduce a regression if some languages are okay using the same translation for both uses of "Following", but since their purpose are different, I think the contextual split is warranted.

This change only affects the two strings in this patch, not the other uses of "Following".

### DIFF
--- a/src/view/screens/Feeds.tsx
+++ b/src/view/screens/Feeds.tsx
@@ -596,7 +596,7 @@ function FollowingFeed() {
             fill={t.palette.white}
           />
         </View>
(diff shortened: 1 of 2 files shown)


akerbeltz:
Fully agree, using the same string for these two contexts doesn't work for any of the 5 languages I speak.

surfdude29:
@estrattonbailey @mozzius This is a tiny change but would fix an annoying localization issue, would it be possible to have a look and maybe squeeze it into the next release? 🙏

Issue #8263, opened 2025-04-22
Regression: Bluesky Git repository always shows missing files on Windows


ssg:
### Steps to Reproduce

1. Clone `bluesky-social/social-app` git repository after March 5th, 2025.
2. Run `git status`
3. Git staging area shows certain files as deleted and they can't be rid of

### Attachments

Git status shows as:

```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    __e2e__/flows/report-dialog/post:default.yml
        deleted:    __e2e__/flows/report-dialog/post:edit-reason.yml
        deleted:    __e2e__/flows/report-dialog/post:reason-other.yml
```

Trying to restore these files fail as commas are used for NTFS data streams. So running this command:

```
git restore "__e2e__/flows/report-dialog/post:default.yml"
```

creates an empty file called `post` in the directory incorrectly and gives out this error:

```
error: unable to stat just-written file __e2e__/flows/report-dialog/post:default.yml: No such file or directory
```

Git status then shows:

```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    __e2e__/flows/report-dialog/post:default.yml
        deleted:    __e2e__/flows/report-dialog/post:edit-reason.yml
        deleted:    __e2e__/flows/report-dialog/post:reason-other.yml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        __e2e__/flows/report-dialog/post
```

### What platform(s) does this occur on?

Web (Mobile), Web (Desktop)

### Device Info

Windows 11 Pro

### What version of the app are you using?

1.101.0

### Additional Information

The changeset #7913 created three files with colons in their names. Git can't process these files on Windows properly and breaks.

It might be a good idea to put a pre-commit hook to prevent such changes that breaks cross-platform development for Bluesky.

surfdude29:
See also:
- #7957 
- #7976

ssg:
@surfdude29 Thanks! I already filed a PR for this.

ssg:
Closing this as duplicate.

PR #8264, opened 2025-04-22
rename file names with colons to the common notation

ssg:
The colons in file names break Windows compatibility and prevent Bluesky from being developed on Windows. (Regression: Bluesky Git repository always shows missing files on Windows #8263)


ssg:
Another issue: #7957

ssg:
There is another patch from March 12th to fix this, #7976. It uses dashes instead of dots. This PR might be more preferable because there is already a file called `account.default.yml` in the same folder, so they would look more uniform I guess.

glundgren93:
Any plans to merging this?

benharri:
Would be great to get this merged

mozzius (approved):

PR #8269, opened 2025-04-22
Modernise change password dialog

mozzius:
<table>
  <thead>
    <tr>
      <th>iOS</th>
      <th>Web</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>


mozzius:
Failing CI seems fine to ignore to me, second build is resetting to a dodgy state

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,310 @@
...
+
+  const [stage, setStage] = useState(Stages.RequestCode)
+  const [isProcessing, setIsProcessing] = useState(false)
+  const [resetCode, setResetCode] = useState('')
+  const [newPassword, setNewPassword] = useState('')
+  const [error, setError] = useState('')
+
+  const uiStrings = {
+    RequestCode: {
+      title: _(msg`Change Password`),
```
surfdude29, line 53:
```suggestion
      title: _(msg`Change password`),
```
nit: Suggest using title case here, as that seems to have become the recent standard for most dialog titles.

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,310 @@
...
+      message: _(
+        msg`If you want to change your password, we will send you a code to verify that this is your account.`,
+      ),
+    },
+    ChangePassword: {
+      title: _(msg`Enter Code`),
```
surfdude29, line 59:
```suggestion
      title: _(msg`Enter code`),
```
nit: Ditto on sentence case for dialog title

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,310 @@
...
+      message: _(msg`Enter the code you received to change your password.`),
```
surfdude29, line 60:
```suggestion
      message: _(msg`Please enter the code you received and the new password you would like to use.`),
```
nit: Suggest rewording using `Please` and as there are two fields for the user to enter info, also mentioning the new password itself.

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,310 @@
...
+    },
+    Done: {
+      title: _(msg`Password Changed`),
```
surfdude29, line 63:
```suggestion
      title: _(msg`Password changed`),
```
nit: As above on sentence case for dialog titles.

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,310 @@
...
+                variant="solid"
+                color="secondary"
+                size="large"
+                disabled={isProcessing}
+                onPress={() => {
+                  setResetCode('')
+                  setStage(Stages.RequestCode)
+                }}>
+                <ButtonText>
+                  <Trans>Back</Trans>
```
surfdude29, line 290: Just wondering, is there much point having a `Back` button here? If the user has already requested a code, why would they need to go back to the previous screen? I guess to request a new code if it hasn't arrived for some reason... so maybe it would make more sense to have a `Resend email` button instead?

surfdude29 (commented):
This looks so much better, very nice 👌

I left some suggestions on minor string changes, and I was also wondering if it might be feasible to reuse some of the code in #8296 to add some of the really nice UI touches in that PR to the change password flow here? E.g.:

- the nifty `Resend` button in `src/components/dialogs/EmailDialog/components/ResendEmailText.tsx`
- the cool new token validation logic in `src/components/dialogs/EmailDialog/components/TokenField.tsx`
- and I'm not completely sure if this would be relevant, but maybe the clean error hook added in `src/lib/hooks/useCleanError.ts`?

I think any (or all) of those would be great additions here :)

Also, it would be nice to have a clean (and thus also translatable) error for when the user enters the code wrong or if it has expired. Currently, the user gets a `Token is invalid` error which could be a little more helpful imo.

IMG_6311

elijaharita:
provided some detailed feedback [on linear](https://linear.app/blueskyweb/issue/APP-1306/review-change-password-modal-pr)

to summarize, there are a couple small issues on android that should probably be addressed

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+    setError('')
+    setIsProcessing(true)
+    try {
+      await agent.com.atproto.server.requestPasswordReset({
+        email: currentAccount.email,
+      })
+      setStage(Stages.ChangePassword)
+    } catch (e: any) {
+      logger.warn('Failed to request password reset', {error: e})
```
estrattonbailey: Let's move this down into the `!isNetworkError` block so we aren't reporting network failures to Sentry.

Also, not really documented anywhere, but `error` is auto-filtered from Sentry logs. For errors like this that are safe to send to Sentry (since this one won't have any PII in it), you can do `{safeMessage: e.message}`

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+    setError('')
+    setIsProcessing(true)
+    try {
+      await agent.com.atproto.server.resetPassword({
+        token: formattedCode,
+        password: newPassword,
+      })
+      setStage(Stages.Done)
+    } catch (e: any) {
+      logger.warn('Failed to set new password', {error: e})
```
estrattonbailey: Same here

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+        </View>
+
+        {stage === Stages.ChangePassword && (
+          <View style={[a.gap_md]}>
+            <View>
+              <TextField.LabelText>
+                <Trans>Confirmation code</Trans>
+              </TextField.LabelText>
+              <TextField.Root>
+                <TextField.Input
```
estrattonbailey, line 189: Doesn't need to be in this PR, but there is a `TokenField` component in the `EmailDialog` that handles validation and formatting of this code. Should be a drop-in here I think, but lmk if that's wrong.

Let's do that as a follow-up? Could generalize that component and move to `#/components/ConfirmationCodeField` or similar name and replace it here and in `EmailDialog`.

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+                  color="secondary"
+                  variant="solid"
+                  disabled={isProcessing}>
+                  <ButtonText>
+                    <Trans>Already have a code?</Trans>
+                  </ButtonText>
+                </Button>
+              ) : (
+                <Button
+                  label={_(msg`Cancel`)}
```
estrattonbailey, line 246: These cancel buttons can be limited to native only I think, since web will have the close button in the top right

src/screens/Settings/components/ChangePasswordDialog.tsx
```diff
@@ -0,0 +1,314 @@
...
+        <View style={[a.gap_sm, gtMobile && [a.flex_row_reverse, a.ml_auto]]}>
+          {stage === Stages.RequestCode ? (
+            <>
+              {!gtMobile && (
+                <Button
+                  label={_(msg`Already have a code?`)}
+                  onPress={() => setStage(Stages.ChangePassword)}
+                  size="large"
+                  color="primary"
+                  variant="ghost"
```
estrattonbailey: We can ditch the `ghost` (and all variants on these buttons, actually) and maybe use `primary_subtle`? We're considering `variant` props deprecated moving forward.

mozzius: Ha sure

mozzius:
Updated web layout, button color

Screenshot 2025-08-20 at 19 53 38
Screenshot 2025-08-20 at 19 53 44

src/screens/Settings/components/ChangePasswordDialog.tsx
mozzius: Oh cool! Yeah followup sounds good

src/screens/Settings/components/ChangePasswordDialog.tsx
mozzius: Already is via `gtMobile`

src/screens/Settings/components/ChangePasswordDialog.tsx
mozzius: Ohhhhh that's what it's for

src/screens/Settings/components/ChangePasswordDialog.tsx
estrattonbailey: On mobile web there's a Cancel button and an X, I think we can just use the X?

mozzius: Oh yeah sure

mozzius:
And for android - not perfect, but I just set a minHeight on the dialog

https://github.com/user-attachments/assets/f6519440-49bd-4065-8bc2-f58cb3ef6665

PR #8272, opened 2025-04-23
Sanitize handle in header

espeon:
This PR sanitizes the handle in the header (the one under the display name). It should match all the other times the handle is used.

### DIFF
--- a/src/screens/Profile/Header/Handle.tsx
+++ b/src/screens/Profile/Header/Handle.tsx
@@ -3,7 +3,7 @@ import {AppBskyActorDefs} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
-import {isInvalidHandle} from '#/lib/strings/handles'
+import {isInvalidHandle, sanitizeHandle} from '#/lib/strings/handles'
 import {isIOS} from '#/platform/detection'
 import {Shadow} from '#/state/cache/types'
 import {atoms as a, useTheme, web} from '#/alf'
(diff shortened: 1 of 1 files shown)


espeon:
Just a small note, this would lead to two `isValidHandle` checks, one inside the actual profile header handle component (we need it b/c of the specific styles associated with the invalid handle chip) and one more inside the `sanitizeHandle` function. 👍🏼

src/screens/Profile/Header/Handle.tsx
```diff
@@ -49,7 +49,9 @@ export function ProfileHeaderHandle({
             : [a.text_md, a.leading_snug, t.atoms.text_contrast_medium],
           web({wordBreak: 'break-all'}),
         ]}>
-        {invalidHandle ? _(msg`⚠Invalid Handle`) : `@${profile.handle}`}
+        {invalidHandle
+          ? _(msg`⚠Invalid Handle`)
```
mozzius, line 53: @surfdude29 is it valuable being able to translate this string? Should I try and get the same string inside of `sanitizeHandle` translated too?

surfdude29: In my view, yes, and there's actually a PR that I think is related to this?
- #6203

mozzius: Ah yeah nice. Super stale unfortunately :(

mozzius (approved):

src/screens/Profile/Header/Handle.tsx
surfdude29: @mozzius If @tkusano is able to rebase #6203 (I don't know how much work that might be), would you be open to reviewing it? It would be good to allow that string to be translatable :)

mozzius: Yep absolutely

tkusano: rebased #6203

PR #8275, opened 2025-04-23
Fix random app crashes caused by reanimated animations

notfirecc:
Small change, I was getting crashes when liking certain posts with the stacktrace pointing to [this](https://github.com/software-mansion/react-native-reanimated/pull/6897) being the issue. Updating reanimated to a newer version that includes a fix seems to have fixed this issue for me, though clearing the image cache is necessary for animations to appear correctly again. No other issues afaik after testing this.


mozzius:
Hopefully this upgrade should be included in #8281

mozzius:
it was

PR #8276, opened 2025-04-23
Verification UI tweaks

estrattonbailey:
Small tweaks and a little nudge for verifiers in case the user doesn't have a display name set.
CleanShot 2025-04-23 at 14 05 40@2x


pfrazee (approved):
Quick skim review, seems good!

PR #8286, opened 2025-04-26
Update mention of MobX in README

mozzius:


mozzius:
The plain React.createContext will continue until morale improves

PR #8287, opened 2025-04-26
Don't await deleting temp files

mozzius:
They get bulk deleted when the composer closes, so it just slows down the loop

### DIFF
--- a/src/state/gallery.ts
+++ b/src/state/gallery.ts
@@ -5,15 +5,15 @@ import {
   moveAsync,
 } from 'expo-file-system'
 import {
-  Action,
-  ActionCrop,
+  type Action,
+  type ActionCrop,
   manipulateAsync,
   SaveFormat,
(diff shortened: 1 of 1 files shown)


haileyok (approved):

mary-ext:
fyi this line is still wrong, the image never gets moved to that special dir
https://github.com/bluesky-social/social-app/blob/03eea0717c5accca008fab0973e3c7c69f2362b4/src/state/gallery.ts#L258

mozzius:
Oh hmm how on earth is that supposed to work then...

mary-ext:
it's supposed to check that the file currently *isn't* in the cache directory before moving it, but the check is reversed, so no files ever get moved to that directory

mozzius:
lmao

mary-ext:
it's embarassing because it's me that committed the code 🙂↕️

PR #8292, opened 2025-04-28
fix: use paginated user search for search screen

### LINKED ISSUE #6656: Increase number of results in People Search
### Describe the Feature

The People tab of search only shows me a maximum of 25 accounts, with seemingly no way to scroll or page to see more. I assume this is a deliberate decision but it makes people search less useful. Especially as it always favours showing accounts that I follow. So if I am looking for people to follow with a shared interest (by searching for a word or hashtag in bios) once I have followed 25 such accounts, people search is no longer usable for finding more of them

### Attachments

_No response_

### Describe Alternatives

- Increase length of results list of people search or scroll/paginate
- Allow search to be filtered to show only people followed/not followed

### Additional Context

_No response_

LoganDupont:
This is an important issue that would be fixed as a lot of users aren't found in the search results due to a current limit of 25 accounts. Due to the facts users aren't found they also don't get new followers and results in leaving the platform. This should be fixed as soon as possible! This will fix #6656. 

Also keep in mind it's my first time that I code with react so probably I made some mistakes.

This issue should actually also be fixed for feeds.


LoganDupont:
Closed in favor of #9009

PR #8294, opened 2025-04-28
[new arch] remove rn-fetch-blob dependency

haileyok:
- **basic bumps**
- **more tweaking**
- **fix rn patch**
- **fix crop picker patch**
- **fix media library patch**
- **rm unnecessary patch**
- **fix notifications patch**
- **update bottomsheet**
- **Update withAppDelegateReferrer.js**
- **Delete withNoBundleCompression.js**
- **rm withNoBundleCompression plugin**
- **rm findLast shim**
- **metro package exports is enabled by default**
- **update react/react-dom/react-compiler**
- **fix reanimated issue**
- **vendor expo-ized emoji popup**
- **fix types**
- **hackfix view full thread**
- **Update EmojiPickerModule.podspec**
- **more upgrades**
- **fix multiformats package version**
- **add baseurl**
- **bump mmkv**
- **bumps**
- **update react-keyed-flatten-children**
- **bump locale packages**
- **fix emoji picker dark mode**
- **rn upgrades**
- **Revert "bump locale packages"**
- **upgrade testing-library**
- **rm test renderer**
- **update patch name minors**
- **rm findNodeHandle from tabbar**
- **only do scrollview tag thing on ios**
- **disable package exports**
- **update expo notifications handler**
- **memoize emoji picker styles**
- **fix tests, mock multiformats**
- **bump some dev deps with RC versions**
- **completely rearchitect toasts**
- **rm logs**
- **layout animation config for composer footer**
- **disable autolinking for patched libs**
- **remove fetchblob dependency**


src/lib/media/manip.ts
```diff
@@ -69,28 +69,13 @@ export async function downloadAndResize(opts: DownloadAndResizeOpts) {
...
-    if (status !== 200) {
-      return
-    }
+  const path = createPath(appendExt)
 
-    const localUri = normalizePath(downloadRes.path(), true)
-    return await doResize(localUri, opts)
+  try {
+    await downloadImage(opts.uri, path, opts.timeout)
+    return await doResize(path, opts)
```
mozzius, line 76: should it used the returned path? since it's been passed to normalizePath

mozzius (approved):
Only tested on Android starter pack image/QR code, but it looks good (after pushing up a fix for the timeout 😉)

Let's test on iOS before we go ahead with it, but this should be good

src/lib/media/manip.ts
haileyok: not quite sure i understand here. (gonna merge so the base branch runs without crashing or anything else, but would like to clarify this)

PR #8296, opened 2025-04-28
[APP-1158] Refactor email-related dialogs

estrattonbailey:
Refactors multiple email related dialogs into a single dialog so that managing dependent state is more cohesive. Also improves on our existing functionality by finessing the design, handling errors, and doing things like validating email addresses.

This PR and the stacked PR do not touch the verification intent dialog in any way. It duplicates the flow added here, but we can address that as a separate PR later on.

### Stacked PRs 
- https://github.com/bluesky-social/social-app/pull/8305 (review this first and merge into here)

### Testing with the local dev env:
1. Change email
    1. Use the same email address already on file, see error
    2. Use invalid email address, see error
    3. Use new email address, ensure success state is reached
2. Verify your email
    1. Use code, either from the change email step (we fire that automatically now) or by sending a fresh email
    2. Use link from email (copy link from terminal, edit for localhost), verify that upon returning to previous tab that you see the success state
3. Change email from verified-email state
    1. Ensure it requests a code
    2. Try resending the email
    3. Enter code, ensure success state

### Testing with a fresh account in prod:
Some features like 2FA and blocking posts and messages don't work with the dev env.
1. Feature blocking, start in unverified state
    1. Try creating a post, ensure that closing without verifying closes the composer
    2. Try messaging a user, ensure closing without verifying redirects to home (do this multiple times, there's a small bug in prod)
    3 . Try creating a post and verifying your email, composer should remain open
    4. Try messaging someone and verifying your email, should leave you on that screen
2. Enable 2FA, start in unverified state
    1. Try to enable 2FA, should be blocked
    2. Try to enable 2FA, verify email using code/link, ensure in both cases that upon return you're redirected to the 2FA step
3. Disable 2FA
    1. You get the idea


src/components/dialogs/EmailDialog/screens/Manage2FA/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      ],
+      onVerify: () => {
+        showScreen({
+          id: ScreenID.Manage2FA,
+        })
+      },
+    })
+
+    return null
+  }
```
mozzius, line 34: This looks like it's setting state in render - probably need to move to an effect?

Screenshot 2025-04-30 at 18 29 47

src/components/dialogs/EmailDialog/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const {_} = useLingui()
+  const emailDialogControl = useEmailDialogControl()
+  const {isEmailVerified} = useAccountEmailState()
+  const onClose = useCallback(() => {
+    if (!isEmailVerified) {
+      if (emailDialogControl.value?.id === ScreenID.Verify) {
+        emailDialogControl.value.onCloseWithoutVerifying?.()
+      }
+    }
+  }, [isEmailVerified, emailDialogControl])
```
mozzius, line 32: The stateful global dialogs I think should have their state cleared on close. I believe this is part of what is causing the issues with the convo view in https://github.com/bluesky-social/social-app/pull/8305
```suggestion
  const onClose = useCallback(() => {
    if (!isEmailVerified) {
      if (emailDialogControl.value?.id === ScreenID.Verify) {
        emailDialogControl.value.onCloseWithoutVerifying?.()
      }
    }
    emailDialogControl.clear()
  }, [isEmailVerified, emailDialogControl])
```

mozzius:
Style nits:

Close button overlapping banner image

Screenshot 2025-04-30 at 17 54 23

Double @ is a bit weird, maybe pick a different icon

Screenshot 2025-05-01 at 10 12 05

The green tick is really pale. Maybe `positive_600` would be better?

Screenshot 2025-05-01 at 11 06 34

mozzius:
This is pretty minor but there's a bit of a disconnect between the wording in the dialog, and the wording in the email

Screenshot 2025-05-01 at 10 12 59
Screenshot 2025-05-01 at 10 13 18

mozzius:
If you have 2FA enabled, do you think we should warn users before changing their email that this will disable 2FA?

mozzius:
Other than that, it's looking great!

src/components/dialogs/EmailDialog/screens/Verify.tsx
```diff
@@ -0,0 +1,354 @@
...
+                value: token,
+              })
+            }}
+            onSubmitEditing={() => {}}
+          />
+
+          {state.error && <Admonition type="error">{state.error}</Admonition>}
+
+          <Button
+            label={_(msg`Verify`)}
```
surfdude29, line 320:
```suggestion
            label={_(msg`Verify code`)}
```
nit: Would it make sense to align the label and the `ButtonText` here?

src/components/dialogs/EmailDialog/screens/Verify.tsx
```diff
@@ -0,0 +1,354 @@
...
+          </Button>
+
+          <Divider />
+
+          <Text
+            style={[a.text_sm, a.leading_snug, t.atoms.text_contrast_medium]}>
+            <Trans>
+              Don't have a code?{' '}
+              <InlineLinkText
+                label={_(msg`Enter code`)}
```
surfdude29, line 339: I think something else might work better as the label here, as this is the case where the user doesn't have a code to enter, right?

src/components/dms/MessageProfileButton.tsx
```diff
@@ -59,14 +51,15 @@ export function MessageProfileButton({
...
-    profile.did,
-    initiateConvo,
-    convoAvailability,
-  ])
+  }, [navigation, profile.did, initiateConvo, convoAvailability])
+
+  const wrappedOnPress = requireEmailVerification(onPress, {
+    instructions: [
+      <Trans key="message">
+        Before you may message another user, you must first verify your email.
```
surfdude29, line 59:
```suggestion
        Before you can message another user, you must first verify your email.
```
nit: Slight suggested wording tweak that sounds a bit less formal imo

src/components/dms/dialogs/NewChatDialog.tsx
```diff
@@ -49,17 +46,22 @@ export function NewChat({
     [control, createChat],
   )
 
+  const onPress = useCallback(() => {
+    control.open()
+  }, [control])
+  const wrappedOnPress = requireEmailVerification(onPress, {
+    instructions: [
+      <Trans key="new-chat">
+        Before you may message another user, you must first verify your email.
```
surfdude29, line 55:
```suggestion
        Before you can message another user, you must first verify your email.
```
nit: As above

src/lib/hooks/useCleanError.ts
```diff
@@ -0,0 +1,91 @@
...
+            msg`The server appears to be experiencing issues. Please try again in a few moments.`,
+          ),
+        }
+      }
+
+      if (raw.includes('Bad token scope')) {
+        return {
+          raw,
+          clean: _(
+            msg`This feature is not available while using an App Password. Please sign in with your main password.`,
```
surfdude29, line 49:
```suggestion
            msg`This feature is not available while using an app password. Please sign in with your main password.`,
```
very small nit: In other parts of the UI I think we've standardized on lowercase for `app password`? But not too sure tbh

src/screens/Messages/ChatList.tsx
```diff
@@ -321,6 +322,18 @@ export function MessagesScreen({navigation, route}: Props) {
...
   const {gtMobile} = useBreakpoints()
+  const requireEmailVerification = useRequireEmailVerification()
+
+  const openChatControl = useCallback(() => {
+    newChatControl.open()
+  }, [newChatControl])
+  const wrappedOpenChatControl = requireEmailVerification(openChatControl, {
+    instructions: [
+      <Trans key="new-chat">
+        Before you may message another user, you must first verify your email.
```
surfdude29, line 333:
```suggestion
        Before you can message another user, you must first verify your email.
```
nit: As above, I think `can` sounds slightly less formal.

src/screens/Messages/Conversation.tsx
```diff
@@ -183,19 +186,50 @@ function InnerReady({
...
+       * Load bearing timeout, to bump this state update until the after the
+       * `navigator.addListener('state')` handler closes elements from
+       * `shell/index.*.tsx`  - sfn & esb
+       */
+      setTimeout(() =>
+        emailDialogControl.open({
+          id: EmailDialogScreenID.Verify,
+          instructions: [
+            <Trans key="pre-compose">
+              Before you may message another user, you must first verify your
```
surfdude29, line 214:
```suggestion
              Before you can message another user, you must first verify your
```
nit: ditto

src/screens/Settings/AccountSettings.tsx
```diff
@@ -97,7 +102,11 @@ export function AccountSettingsScreen({}: Props) {
           )}
           <SettingsList.PressableItem
             label={_(msg`Change email`)}
```
surfdude29, line 104:
```suggestion
            label={_(msg`Update email`)}
```
I suggest using `Update` here to be consistent with the language used in other strings.

src/screens/Settings/AccountSettings.tsx
```diff
@@ -97,7 +102,11 @@ export function AccountSettingsScreen({}: Props) {
...
-            onPress={() => changeEmailControl.open()}>
+            onPress={() =>
+              emailDialogControl.open({
+                id: EmailDialogScreenID.Update,
+              })
+            }>
             <SettingsList.ItemIcon icon={PencilIcon} />
             <SettingsList.ItemText>
               <Trans>Change email</Trans>
```
surfdude29, line 112:
```suggestion
              <Trans>Update email</Trans>
```
Ditto

src/components/dialogs/EmailDialog/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+    emailDialogControl.clear()
+  }, [isEmailVerified, emailDialogControl])
+
+  return (
+    <Dialog.Outer control={emailDialogControl.control} onClose={onClose}>
+      <Dialog.Handle />
+
+      <Dialog.ScrollableInner
+        label={_(msg`Make adjustments to your account email settings`)}
```
surfdude29, line 40:
```suggestion
        label={_(msg`Make adjustments to email settings for your account`)}
```
nit: This might read slightly clearer?

src/lib/hooks/useOpenComposer.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function useOpenComposer() {
+  const {openComposer} = rootUseOpenComposer()
+  const requireEmailVerification = useRequireEmailVerification()
+  return useMemo(() => {
+    return {
+      openComposer: requireEmailVerification(openComposer, {
+        instructions: [
+          <Trans key="pre-compose">
+            Before creating a post, you must first verify your email.
```
surfdude29, line 15:
```suggestion
            Before creating a post or replying, you must first verify your
            email.
```
nit: This message is shown when replying as well as creating a root post, so might be good to include that case too.

surfdude29 (commented):
I did quite a bit of testing on the preview branch and this works really well, very nice work! 👌

Sorry for leaving so many suggestions 😬  They're all string-related and mostly nits tbh.

src/components/dialogs/EmailDialog/components/TokenField.tsx
```diff
@@ -0,0 +1,39 @@
...
+import {msg} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+
+import * as TextField from '#/components/forms/TextField'
+import {Shield_Stroke2_Corner0_Rounded as Shield} from '#/components/icons/Shield'
+
+export function normalizeCode(value: string) {
+  const normalized = value.toUpperCase().replace(/[^A-Z2-7]/g, '')
+  if (normalized.length <= 5) return normalized
+  return `${normalized.slice(0, 5)}-${normalized.slice(5, 10)}`
```
surfdude29, line 11: Would it be worth adding some regex etc to check that the user has entered all 10 digits of the code before actually submitting it?

At the moment, if the user misses out a digit or more in their typing they'll get a generic `Failed to verify email, please try again.` error.

I asked Claude and he came up with this, but there's probably a simpler way of doing it 😅

https://claude.ai/public/artifacts/2307f345-534c-4963-bf87-9281c8bcbe64

src/components/dialogs/EmailDialog/screens/Verify.tsx
```diff
@@ -0,0 +1,354 @@
...
+            </ButtonText>
+            {state.mutationStatus === 'pending' && <ButtonIcon icon={Loader} />}
+          </Button>
+
+          <Divider />
+
+          <Text
+            style={[a.text_sm, a.leading_snug, t.atoms.text_contrast_medium]}>
+            <Trans>
+              Don't have a code?{' '}
```
surfdude29, line 337:
```suggestion
              Don't have a code or need a new one?{' '}
```
Maybe this is too much of an edge case, but if a user: requests a code, prefers to copy/paste or type it in rather than clicking the link in the email and then gets distracted/forgets etc, when they then go to enter the code and it's rejected because it's expired, it's perhaps not completely obvious to the user what they should do.

From their POV they might think "I *have* a code, it's just not working". So I think a little prompting here that they can easily get a new code might be helpful.

surfdude29:
Sorry for commenting so much on this PR, I have a few more thoughts that I hope might be helpful...

---

In prod, there's a `Change your email address` / `Need to change it?` link in the **Verify Your Email** dialog for people who have an email address on their account that they can't verify. However, there's no equivalent link in the new email verification flow that I can see.

| Verify email (gated feature) | Verify email (standard) 
|--------|--------|
| IMG_6265 | IMG_6286 | 

I think it might be quite a good idea to add a link – perhaps a simple `(change email)` after the user's email address? Or if a link isn't considered desirable, perhaps just a pointer like `If you need to update your email before verifying you can do that in Settings.`

Because otherwise, if a user is asked to verify their email as a result of the daily reminder dialog or when trying to post etc (i.e. not in the context of Settings where there's an obvious **Change email** option), I feel like they might be a bit confused as to how they can verify their email if the dialog is showing an email address that they don't have access to.

---

I noticed this while testing and it's obviously out of scope for this PR but I thought it's worth mentioning because it would be good to get fixed at some point: when you request to disable 2FA, you receive a somewhat confusing email with the subject **Email Update Requested**:

IMG_8563

"I'm trying to disable 2FA, so why is this email talking about 'your new email' when I'm not trying to update my email address?" 🤨

---

And lastly, I ran into something odd while I was testing. After I created a new test account in prod, signed in on the preview branch and then dismissed the verification prompt, I was surprised to be able to open the composer and create a post.

However, the **Verify email** dialog did pop up as it should every time I tried to open the composer after that.

I guess it must have been some kind of weird glitch because I tried creating two new test accounts after that and I wasn't able to reproduce it ¯\_(ツ)_/¯

mozzius (approved):
💎

src/components/dialogs/EmailDialog/screens/Update.tsx
```diff
@@ -0,0 +1,302 @@
...
+            <View style={[a.flex_row, a.gap_sm, a.align_center]}>
+              <Check fill={t.palette.positive_600} size="xs" />
+              <Text style={[a.text_md, a.font_heavy]}>
+                <Trans>Success!</Trans>
+              </Text>
+            </View>
+            <Text style={[a.leading_snug]}>
+              <Trans>
+                Click on the link in the email we just sent you to verify your
+                email.
```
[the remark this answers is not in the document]
estrattonbailey: If a user verifies their email, then changes it, they don't lose access to the features. But regardless this is a good shout in case a user changes their email immediately after signup, or if we adjust this handling in the future. More clarity is better than less!

src/components/dialogs/EmailDialog/screens/Verify.tsx
estrattonbailey: I like that! Will add 👍

estrattonbailey:
@surfdude29 thanks for all the comments! All seem fine to me, always good to tighten up copy 👍 

Added back the missed "Update email" link on the verification dialog, we had dropped it but I'm not sure "too many links on that screen" is a good enough reason to do so.

Re: email copy, yeah we need to clean that up too, that'll be a whole thing because we reuse that template for multiple purposes rn. Will get around to that eventually.

src/components/dialogs/EmailDialog/components/TokenField.tsx
estrattonbailey: Yeah this wasn't very good as it was, made some updates.

surfdude29: This works much better now, nicely done 👌

surfdude29:
@estrattonbailey No worries 👍

The "update email" link is great 👌

On the email templates: sounds good 👍 And for that particular email it's probably not even seen that often anyway, as I would guess that the flow where a user just disables 2FA (with no email update) is probably not that common.

src/components/dialogs/EmailDialog/screens/Update.tsx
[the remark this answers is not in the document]
estrattonbailey: Oh 🤦 I checked this using an account old enough that we don't do this check, forgot we handled it like that for older accounts but not new ones.

You are correct, so I'm doubly glad you called this out. Thanks!

src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
```diff
@@ -0,0 +1,79 @@
...
+    isEmailVerified: !!agent.session?.emailConfirmed,
+    email2FAEnabled: !!agent.session?.emailAuthFactor,
+  }
+  const query = useQuery<AccountEmailState>({
+    enabled: !!agent.session,
+    refetchOnWindowFocus: true,
+    queryKey: accountEmailStateQueryKey,
+    queryFn: async () => {
+      // will also trigger updates to `#/state/session` data
+      const {data} = await agent.resumeSession(agent.session!)
```
mackuba, line 55: @estrattonbailey is it expected that this fires every time the user switches back to the tab/window where the app is open? It feels kinda weird that something in "components/dialogs/EmailDialog" (I don't have any dialog open) is doing that…

mozzius: Yes, well it's determining whether it should open the email dialog...

estrattonbailey: Ah yeah we could probably improve this, good shout

mozzius: Woah... @mackuba you might've accidentally found what was causing our logout issue https://github.com/bluesky-social/social-app/pull/8670

surfdude29: I don't think it was an accident, he's been working on tracking it down 🕵️‍♂️😉

https://bsky.app/profile/mackuba.eu/post/3ltvaawceuk22

mozzius: Ah I stand corrected - thanks for drawing our attention to this nasty bit of code @mackuba!

mackuba: Yeah @mozzius, I've been looking at the logs I've added for that, and I was surprised that when I switched windows there and back I was already getting logs that seemed related to the session, so I started digging where it's coming from…

I'm not sure yet if *this* is what's causing the logouts (I haven't got the logout in the last 2-3 days since I added more logs to get more details).

I've also seen the DM chat.bsky.convo.getLog endpoint getting called several times during the night/day while the laptop was closed, causing the session refresh (while probably the browser might have been in some half-asleep mode).

mackuba: @mozzius 🙃

Screenshot 2025-07-20 at 00 09 23

mackuba: But I think this might be only a part of the problem, the other part is that I apparently had tabs open with 5 different access tokens at the same time

Screenshot 2025-07-20 at 00 18 07

mozzius: Wait so each tab has a different token; does any one of them failing cause the app to log you out?

estrattonbailey: Different access tokens active at the same time I think would mean they were generated from the same refresh token. But each of these new access tokens have their own refresh token, iirc, which effectively "forks" the refresh token. These forked refresh tokens are invalid and will fail the next time we try to refresh the session, hence leading to what appears to be a legitimate auth failure and log out.

mackuba: Not each tab different, but some tabs sometimes have a different token than others, in that case I had 5 different ones across some larger number of tabs. And I think at that point only one of them is valid and the rest is old/expired, so once a tab tries to use this one -> 💥 

@estrattonbailey yeah, I think that's one option, I think I've seen something like this. When viewing this log file, I'm replacing the access-request token pairs with e.g. "TOKEN A - TOKEN A", but there were some lines where one was TOKEN A and the other was different.

What I'm looking at right now:
- that `lastSession` I talked about with Dan yesterday, which I think might be caching an old token that's restored on connection error
- and I think it's also possible that when tabs refresh the token while on a half-asleep laptop, they might not always get the chance to properly broadcast/save it before they go to sleep again

PR #8295, opened 2025-04-28
new arch

haileyok:
- **basic bumps**
- **more tweaking**
- **fix rn patch**
- **fix crop picker patch**
- **fix media library patch**
- **rm unnecessary patch**
- **fix notifications patch**
- **update bottomsheet**
- **Update withAppDelegateReferrer.js**
- **Delete withNoBundleCompression.js**
- **rm withNoBundleCompression plugin**
- **rm findLast shim**
- **metro package exports is enabled by default**
- **update react/react-dom/react-compiler**
- **fix reanimated issue**
- **vendor expo-ized emoji popup**
- **fix types**
- **hackfix view full thread**
- **Update EmojiPickerModule.podspec**
- **more upgrades**
- **fix multiformats package version**
- **add baseurl**
- **bump mmkv**
- **bumps**
- **update react-keyed-flatten-children**
- **bump locale packages**
- **fix emoji picker dark mode**
- **rn upgrades**
- **Revert "bump locale packages"**
- **upgrade testing-library**
- **rm test renderer**
- **update patch name minors**
- **rm findNodeHandle from tabbar**
- **only do scrollview tag thing on ios**
- **disable package exports**
- **update expo notifications handler**
- **memoize emoji picker styles**
- **fix tests, mock multiformats**
- **bump some dev deps with RC versions**
- **completely rearchitect toasts**
- **rm logs**
- **layout animation config for composer footer**
- **disable autolinking for patched libs**
- **undo lingui changes**
- **enable new arch**

PR #8298, opened 2025-04-29
[new arch] experiment -  use animatedview/animatedref to measure rect for lightbox

haileyok:
**Stacked on #8295**

We originally used this method of measuring a ref based on findings in https://github.com/bluesky-social/social-app/pull/6500. See that PR for context. The TLDR is that this was much slower than the preferred method we switched to, however, it seems that might no longer be necessary.

The current implementation causes crashes unfortunately and will need to get reworked regardless, so may need to revisit again if this appears to not perform well. However, from the findings below things do seem better.

## Test Plan

We originally used this alternative method for measuring based on render times that we observed in https://github.com/bluesky-social/social-app/pull/6500. We'll compare timings to what we observed after the fact.

In addition, we still want to test this on some lower end devices.

The easiest way to get some easy perf testing is to go to something like the Cat Pics feed and "just scroll". Also should make sure that things still work as expected for:

- User Avatar
- Feed Avatar
- Images in feed 

Screenshot 2025-04-28 at 10 05 58 PM

Screenshot 2025-04-28 at 10 06 52 PM


src/view/com/util/post-embeds/index.tsx
```diff
@@ -158,13 +163,15 @@ export function PostEmbeds({
...
         fetchedDims: (Dimensions | null)[],
       ) => {
-        const handles = refs.map(r => r.current)
         runOnUI(() => {
           'worklet'
-          const rects = handles.map(measureHandle)
+          const rects: (MeasuredDimensions | null)[] = []
+          for (const r of refs) {
+            rects.push(measure(r))
+          }
```
haileyok, line 174: avoiding `.map` here since we need this to run on ui thread

mozzius:
Found a bug

https://github.com/user-attachments/assets/f01dbefb-0325-495a-921d-db33a29d48db

mozzius:
Also, the X button is often hidden behind the status bar, but it's unclear if it's caused by this

Screenshot 2025-04-29 at 17 50 24

PR #8305, opened 2025-04-29
[APP-1158] Gate features by email verification state

estrattonbailey:
Adds a new hook similar to `useRequireAuth` called `useRequireEmailVerification` which will similarly block the action until email verification has been completed. This hook uses the new `EmailDialog` added in #8296.

```ts
import {useRequireEmailVerification} from '#/lib/hooks/useRequireEmailVerification'

const requireEmailVerification = useRequireEmailVerification()
 
// The resulting callback inherits the types so it can be used pretty much anywhere.
const onPress = requireEmailVerification((e: GestureResponderEvent) => {...})
```

In this PR I've used this hook to wrap everything that we currently have gated using the existing `VerifyEmailDialog`, with a couple exceptions:
- for the composer, I opted to gate all calls to `openComposer` instead of doing it implicitly when the composer opens
- for the `Conversation` screen I needed to hack this a bit to prevent infinite renders, see comment below

This PR also adds a new hook `useOpenComposer`, which is now our canonical way to open the composer instead of `useComposerControls`. I attempted to use `useRequireEmailVerification` within the composer context, but could not get around circular import issues. So I ended up needing to handle it at the callsites. Thus, the new hook to prevent anyone from using the lower level API accidentally. **So most of the edits here are just swapping in this new method call.**

With this PR merged, we can remove `VerifyEmailDialog` and `ChangeEmailDialog` entirely.

## Testing
Don't bother testing the email flows here, just double check that features are blocked. You'll need to create a new account in prod to test with, don't verify your email, then click through a few key test cases here **on web and native:**
1. Composer
2. Conversation (message your new account from a diff account to test)
3. Lists
4. Etc

### DIFF
--- a/src/lib/hooks/useOpenComposer.tsx
+++ b/src/lib/hooks/useOpenComposer.tsx
@@ -1,0 +1,21 @@
+import {useMemo} from 'react'
+import {Trans} from '@lingui/macro'
+
+import {useRequireEmailVerification} from '#/lib/hooks/useRequireEmailVerification'
+import {useOpenComposer as rootUseOpenComposer} from '#/state/shell/composer'
+
+export function useOpenComposer() {
+  const {openComposer} = rootUseOpenComposer()
+  const requireEmailVerification = useRequireEmailVerification()
+  return useMemo(() => {
+    return {
+      openComposer: requireEmailVerification(openComposer, {
+        instructions: [
+          <Trans key="pre-compose">
+            Before creating a post, you must first verify your email.
+          </Trans>,
+        ],
+      }),
+    }
+  }, [openComposer, requireEmailVerification])
+}
--- a/src/screens/Messages/Conversation.tsx
+++ b/src/screens/Messages/Conversation.tsx
@@ -17,6 +17,7 @@ import {type NativeStackScreenProps} from '@react-navigation/native-stack'
 
 import {useEmail} from '#/lib/hooks/useEmail'
 import {useEnableKeyboardControllerScreen} from '#/lib/hooks/useEnableKeyboardController'
+import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {
   type CommonNavigatorParams,
   type NavigationProp,
@@ -192,7 +193,7 @@ function InnerReady({
   const {needsEmailVerification} = useEmail()
   const emailDialogControl = useEmailDialogControl()
 
-  React.useEffect(() => {
+  const maybeBlockForEmailVerification = useNonReactiveCallback(() => {
     if (needsEmailVerification) {
       emailDialogControl.open({
         id: EmailDialogScreenID.Verify,
@@ -207,7 +208,10 @@ function InnerReady({
         },
       })
     }
-  }, [needsEmailVerification, emailDialogControl, navigation])
+  })
+  React.useEffect(maybeBlockForEmailVerification, [
+    maybeBlockForEmailVerification,
+  ])
 
   return (
     <>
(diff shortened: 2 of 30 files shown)


src/state/shell/composer/index.tsx
```diff
@@ -125,5 +125,17 @@ export function useComposerState() {
...
+
+/**
+ * DO NOT USE DIRECTLY. The deprecation notice as a warning only, it's not
+ * actually deprecated.
+ *
+ * @deprecated use `#/lib/hooks/useOpenComposer` instead
+ */
+export function useOpenComposer() {
+  const {openComposer} = React.useContext(controlsContext)
+  return React.useMemo(() => ({openComposer}), [openComposer])
```
estrattonbailey, line 140: Split these apart so that it's harder for anyone to accidentally use the lower level open API.

src/view/com/composer/Composer.tsx
```diff
@@ -333,25 +328,6 @@ export const ComposePost = ({
     }
   }, [onPressCancel, closeAllDialogs, closeAllModals])
 
-  const {needsEmailVerification} = useEmail()
-  const emailDialogControl = useEmailDialogControl()
-
-  useEffect(() => {
-    if (needsEmailVerification) {
-      emailDialogControl.open({
```
estrattonbailey, line 341: Like the `Conversation` view, this will cause an infinite render loop. All access to the composer is now gated, so we don't need anything like this here. For the Conversation, access is more difficult to gate, so I left that use case.

src/view/com/composer/videos/SelectVideoBtn.tsx
estrattonbailey, line 1: We don't need to gate this anymore, the entire composer is gated

src/lib/hooks/useOpenComposer.tsx
```diff
@@ -0,0 +1,21 @@
+import {useMemo} from 'react'
+import {Trans} from '@lingui/macro'
+
+import {useRequireEmailVerification} from '#/lib/hooks/useRequireEmailVerification'
+import {useOpenComposer as rootUseOpenComposer} from '#/state/shell/composer'
+
+export function useOpenComposer() {
+  const {openComposer} = rootUseOpenComposer()
```
mozzius, line 8: nit: I'd prefer `useRootOpenComposer` or `useOpenComposer_root` or something. Hook not starting with `use` is confusing

mozzius:
DM gate doesn't seem to work when receiving inbound messages

https://github.com/user-attachments/assets/b6eabaed-2f54-44a8-8212-29f651bc4548

mozzius:
Ok pressing the accept button explicitly shows the dialog as it should, but subsequent attempts to open the chat don't show it

https://github.com/user-attachments/assets/9ca279a6-8311-43d7-9ba0-cabe52f81258

src/screens/Messages/Conversation.tsx
```diff
@@ -207,7 +212,10 @@ function InnerReady({
         },
       })
     }
-  }, [needsEmailVerification, emailDialogControl, navigation])
+  })
+  React.useEffect(maybeBlockForEmailVerification, [
+    maybeBlockForEmailVerification,
+  ])
```
mozzius, line 218: Yeah, this isn't working for me. It works when you refresh the page on the convo screen itself, but a page navigation to open this screen opens and immediately closes the dialog.

Screenshot 2025-05-01 at 10 38 15

mozzius:
Pushed up a hackfix (setTimeout does the trick, but I'd love to understand what's actually going on here)

Also added a prompt to the accept button:

Screenshot 2025-05-01 at 10 46 19

mozzius (approved):
Pushed up some tweaks but otherwise lgtm! The hook is very nice to use

The dialog not working unless you use a `setTimeout` is a bit concerning but I'm fine with letting it slide - might be worth making a ticket to look into more deeply

estrattonbailey:
Tried `useFocusEffect` too but not working on web, fine on native.

PR #8307, opened 2025-04-29
[APP-1161] Fix crash when closing open dialogs

estrattonbailey:
Potential fix for a crash when handling email verification intent links.

To repro from main using a dev build on a physical device:
- create a new account with an unverified email
- send yourself an email verification link
- open any dialog e.g. post dropdown menu
- click on link in email
- app will crash

**However,** it seems the issue is not necessarily `BottomSheetNativeComponent.dismissAll()`, bc I can independently trigger this method and close open sheets just fine. It seems only to be an issue when calling this native method after going from background-foreground.


haileyok:
Fixed in https://github.com/bluesky-social/social-app/pull/8348

PR #8308, opened 2025-04-29
Add backup / register rotation key button in settings

fossephate:
Wording and styles could probably be better / use some tweaks

demo video!
https://bsky.app/profile/did:plc:cn4gldkpxj43zpuqztnwyf6h/post/3lnybpypey22y

new demo!:
https://bsky.app/profile/fosse.co/post/3lochsdkfxs2d


fossephate:
some notes / todo
- keys are not stored on web, just shown for the user to copy to their clipboard
- UI has changed a bit since the first demo and now more closely matches the UI for the app passwords screen
- Should maybe remove the name of the and replace with the did when displaying it in the list, or give a way to name the key
- No way to actually use the backed up key, though that can come later / be a separate PR
- Maybe prompt to backup a key after logging in on mobile and no local key is found?
- Still need to enable face ID / biometrics / do permissions handling

bnewbold:
This is a big feature addition! It would land in our branded app, which means we would need to support it, document it, and take a degree of responsibility for outcomes (eg, if folks get hacked and lose identity without understanding what happened, that would be partially on us, disclaimers or not).

Have a lot of thoughts, will try to organize them.

Good things (more likely to merge some form of this):

- we definitely want everybody to have more material control and ownership over their identity, and this is a step in that direction
- we love the community pushing things forward and showing what is possible; thanks @fossephate!
- access to stored keys when logged out is a good touch (eg, if PDS is down or account has been takendown)

Downsides (less likely to merge some form of this):

- the first and largest piece of work on this feature is an analysis of all the security design options and trade-offs, for all the target platforms we support (Web, Android, iOS). I don't see that here, or linked; we'd need to do a research project before getting to code review
- if at all possible i'd like to avoid directly handing end users private cryptographic keys as a string in the user interface. would much rather have keys provisioned in hardware in a non-exportable manner (eg, CPU enclave or hardware dongle), or directly insert in to a system like passkeys.
- if we were blocking the overall ecosystem in way were something was just not possible, i'd be much more inclined to get something like this merged. but it is possible to experiment and implement something like this independently, eg as a separate mobile app or tool. The advantage of having this in the Bluesky app is the expected impact and accessibility; but those are exactly the factors which mean we need to think through all options and get it right.

Open questions:

- would the PDS account management interface be a better place for this? we haven't announced/documented it yet, but that is starting to roll out and we expect most account management stuff to happen there: account migration, password reset flow, account deletion, export backups, etc. that is the place I would put an "identity key management" interface, eg the ability to register new keys (as public keys) or remove existing keys. on the other hand, the Bluesky app is what a lot of folks use today, and on mobile might have more direct access to mobile OS or hardware security features.
- maybe something like this could be merged behind a feature gate, to make it accessible to community forks like deer.social
- should there be an independent "DID PLC app" for identity management? something app-neutral / vendor-neutral, and with a much smaller attack surface

haileyok:
> should there be an independent "DID PLC app" for identity management? something app-neutral / vendor-neutral, and with a much smaller attack surface

This is my inclination. In my view, the _optimal_ place for something like this is outside of the social application itself. I'm very not sold on the PDS account interface being a good place for this, and I'd really suggest we refrain from doing that. Managing e.g. sessions in the interface is fine, but this is a lot more potentially destructive and dangerous than that.

Another upside of making this a separate application would be the ability to handle other things e.g. repo backups and actual account migrations inside of said app, _regardless_ of what happens to the social app (i.e. removed from app stores, bugs, etc.)

haileyok:
> the first and largest piece of work on this feature is an analysis of all the security design options and trade-offs, for all the target platforms we support (Web, Android, iOS). I don't see that here, or linked; we'd need to do a research project before getting to code review

At least on the note of where and how this would work, I'd definitely say it should be Android and iOS only. On either platform, `expo-secure-store` under the hood would be using Android Keystore or Keychain Services respectively. Both systems make it difficult to extract keys from the device as well as restrict access to the keys within the device to _only_ the application that created the key (or, more accurately, to the set of applications that you actually allow to access the keychain item when you create the item). 

https://developer.apple.com/documentation/security/keychain-services
https://developer.android.com/privacy-and-security/keystore

It's definitely worth a write up and investigation on this though. I think I'd also be inclined to say that rather than use `expo-secure-store` itself that it might even be worth whoever would maintain such an application maintain the key access code on their own as well (or, if after review `expo-secure-store` looks to be sufficient in their e.g. defaults, that a fork of that library be maintained).

haileyok:
@fossephate I'd definitely like to chat more about this, so definitely feel free to bother me whenever about it!

haileyok:
Mostly unrelated, but I do think going for something like https://github.com/char/k256-paper-key would be cool too!

fossephate:
> @fossephate I'd definitely like to chat more about this, so definitely feel free to bother me whenever about it!

Currently in SF for the next week (maybe 2) but will take you up on this once I get back!

Some immediate thoughts and notes:

getting users to do anything “not technically necessary” is /really/ hard, and while a separate app for identity and key management I think makes sense I think it also already exists for many in the form of a password manager. That being said there’s unfortunately no easy way to add a button (on both ios and android) to “Store this in your password manager” (you can try to trick ios into thinking a text field is a password and you’re signing up for something but it’s pretty jank)

fossephate:
> Mostly unrelated, but I do think going for something like https://github.com/char/k256-paper-key would be cool too!

storing things as a passphrase definitely opens the door to a lot of pre-built solutions, as the crypto / wallet space has been trying to solve this problem (key management) for quite some time and it’s probably worth looking into some of the solutions already in the space

I’m not sure it’s a good fit for bsky but just to give an example: magic.link allows you to use other socials / email as a proof of identity in order to access keys (at least, iirc that’s how it works)

PR #8313, opened 2025-05-01
add expo env variables for appview and public api

a-viv-a:
This PR adds [expo env variables](https://docs.expo.dev/guides/environment-variables/) to configure the appview proxy and the public api url to use for requests at compile time. This makes it easier to selfhost social app as part of a 3rd party appview deployment, without maintaining any patches.

I referenced [futurs work](https://github.com/appview-wg-bsky/social-app) for this.

Worth highlighting that the way expo processes env files, the `#` in the did will be treated as an inline comment unless quoted.

There's a lot more that *could* be configured in env variables to try and make rebranded selfhosting easier, but I think that's a much bigger piece of work, given the lack of perceptual colorspace and the inlined assets and svgs throughout the code.

PRing this to avoid a parallel solution in deer.

## Testing

I tried this with `did:web:bsky.zeppelin.social#bsky_appview` and `https://bsky.zeppelin.social` on the web, which worked—something seems to be wrong with feeds on zeppelin though.

## Open questions

There are a couple places I didn't update for fear of breakage:
- [`src/state/queries/email-verification-required.ts`](https://github.com/bluesky-social/social-app/blob/main/src/state/queries/email-verification-required.ts)
  - not confident enough in what this is doing
- [`src/lib/notifications/notifications.ts`](https://github.com/bluesky-social/social-app/blob/main/src/lib/notifications/notifications.ts)
  - [this question](https://github.com/bluesky-social/atproto/discussions/2811) implies the dids used might be placeholders...

I updated [`src/lib/react-query.tsx`](https://github.com/bluesky-social/social-app/blob/main/src/lib/react-query.tsx) but zeppelin doesn't respond to the health xrpc route correctly... hard to test

Is it ok to be switching from `api.bsky.app` to `public.api.bsky.app` in [`src/lib/api/feed/custom.ts`](https://github.com/bluesky-social/social-app/blob/main/src/lib/api/feed/custom.ts) with the use of `PUBLIC_BSKY_SERVICE`?

The names are pretty awful because of the `EXPO_PUBLIC_` + `NAME` thing going on for `EXPO_PUBLIC_PUBLIC_BSKY_SERVICE`... But I worry about clarity otherwise? `EXPO_PUBLIC_BSKY_SERVICE` would theoretically refer to `BSKY_SERVICE` and not `PUBLIC_BSKY_SERVICE`. Ideally you could drop the expo prefix, but the docs are pretty vague on if this will cause issues or inhibit inlining.

I'm reading in the environment variables in [`src/env.ts`](https://github.com/bluesky-social/social-app/blob/main/src/env.ts) since the typecast is unwieldy to do in the constants file, and I chose to prefix with `ENV_` so that these values wouldn't conflict in autocompletion with their constants pair, since the constants version has a default value. Should `process.env` just be read in constants? I didn't see any existing uses of the values in env fwiw.


a-viv-a:
My new hunch is that https://bsky.zeppelin.social/xrpc/_health not returning a version is why feeds seem broken when using that appview, since react-query uses this to check connection status.

a-viv-a:
running prettier now, sorry about that

PR #8320, opened 2025-05-02
[new arch] Fix profile perf

mozzius:
**Stacked on #8295**

The culprit of the slow perf on the profile pages was this mechanism for keeping the scrollviews in sync:

Screenshot 2025-05-02 at 13 27 08

This syncs the other scrollviews in the pager every 80ms while the header is visible.

It seems like Reanimated's `scrollTo` is much much slower on Fabric. On Paper, we were able to do this 6x every 80ms. Now, it absolutely tanks performance.

This fixes it by changing the approach for keeping the scrollviews in sync. Rather than polling, we instead just sync on demand - either when the pager beings dragging, or when a tab is pressed

## Before

https://github.com/user-attachments/assets/e9c39c55-b687-4bc0-82c7-a831b186edb3

## After

https://github.com/user-attachments/assets/0c9507b6-7011-49e9-8f43-db3f0a71e1be

## Test plan

- Test perf on new arch.
- Confirm scrollviews are still synced properly

You probably want to review code with whitespace hidden, because I removed the `React.forwardRef` (I ❤️ React 19)

### DIFF
--- a/src/view/com/pager/Pager.tsx
+++ b/src/view/com/pager/Pager.tsx
@@ -1,16 +1,22 @@
-import React, {forwardRef, useCallback, useContext} from 'react'
+import {
+  useCallback,
+  useContext,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react'
 import {View} from 'react-native'
 import {DrawerGestureContext} from 'react-native-drawer-layout'
 import {Gesture, GestureDetector} from 'react-native-gesture-handler'
 import PagerView, {
-  PagerViewOnPageScrollEventData,
-  PagerViewOnPageSelectedEvent,
-  PagerViewOnPageSelectedEventData,
-  PageScrollStateChangedNativeEventData,
+  type PagerViewOnPageScrollEventData,
+  type PagerViewOnPageSelectedEvent,
+  type PagerViewOnPageSelectedEventData,
+  type PageScrollStateChangedNativeEventData,
 } from 'react-native-pager-view'
 import Animated, {
   runOnJS,
-  SharedValue,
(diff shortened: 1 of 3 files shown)


src/view/com/pager/Pager.tsx
```diff
@@ -47,114 +57,112 @@ interface Props {
 
 const AnimatedPagerView = Animated.createAnimatedComponent(PagerView)
 
-export const Pager = forwardRef<PagerRef, React.PropsWithChildren<Props>>(
```
haileyok, line 50: ohmygod no more forwardRef!!

src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -1,17 +1,16 @@
...
+  type ScrollView,
   StyleSheet,
   View,
 } from 'react-native'
 import Animated, {
-  AnimatedRef,
-  runOnJS,
+  type AnimatedRef,
   runOnUI,
   scrollTo,
```
haileyok, line 12: i do wonder if we could sop using this in general now...might be better to just use the scroll function on the view itself (but we can test this separately)

src/view/com/pager/Pager.tsx
mozzius: lets goooo

hirbod:
FYI, this is the upstream issue for it: https://github.com/software-mansion/react-native-reanimated/issues/6999

mozzius:
Good to know, thanks. Luckily our tab component's use of `scrollTo` was non-critical and easy to reduce

PR #8323, opened 2025-05-03
New `Select` component

mozzius:


src/components/Select/index.web.tsx
```diff
@@ -0,0 +1,252 @@
...
+export function Trigger({children, label}: TriggerProps) {
+  const t = useTheme()
+  const {
+    state: hovered,
+    onIn: onMouseEnter,
+    onOut: onMouseLeave,
+  } = useInteractionState()
+  const {state: focused, onIn: onFocus, onOut: onBlur} = useInteractionState()
+
+  if (typeof children === 'function') {
```
estrattonbailey, line 62: Do we need to support this? I would expect a `select` to be displayed the same in all contexts, with interaction effects built in.

I realize this is probably just a result of copying from from `Menu`, and it's not really a huge deal. But less is more sometimes!

src/components/Select/index.web.tsx
```diff
@@ -0,0 +1,252 @@
...
+    <RadixSelect.Icon>
+      <ChevronDownIcon style={[t.atoms.text, style]} size="xs" />
+    </RadixSelect.Icon>
+  )
+}
+
+export function Content<T>({items, renderItem}: ContentProps<T>) {
+  const t = useTheme()
+
+  const scrollBtnStyles = [
```
estrattonbailey, line 143: There's a little jump when these appear, they could probably be `absolute`.

src/components/Select/index.web.tsx
```diff
@@ -0,0 +1,252 @@
...
+            a.flex_1,
+            a.border,
+            t.atoms.border_contrast_low,
+            a.rounded_sm,
+          ]}>
+          <RadixSelect.ScrollUpButton style={flatten([scrollBtnStyles])}>
+            <ChevronUpIcon style={[t.atoms.text]} size="xs" />
+          </RadixSelect.ScrollUpButton>
+          <RadixSelect.Viewport style={flatten([a.p_xs])}>
+            {items.map((item, index) => renderItem(item, index))}
```
estrattonbailey, line 170: Might be good to pass in `selected` here too for easier custom `Item`s. `Item` can keep it too, since it's the default.

src/components/Select/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+  hovered: false,
+  focused: false,
+  pressed: false,
+})
+
+export function useItemContext() {
+  return useContext(ItemContext)
+}
+
+export function Item({children, value, label}: ItemProps) {
```
estrattonbailey, line 236: Probs want a `style` prop here too?

src/view/com/composer/videos/SubtitleDialog.tsx
```diff
@@ -240,19 +239,21 @@ function SubtitleFileRow({
             numberOfLines={1}>
             {file.name}
           </Text>
-          <RNPickerSelect
-            placeholder={{
-              label: _(msg`Select language...`),
-              value: '',
-            }}
+          <select
```
estrattonbailey, line 242: Why raw `select` here?

src/screens/Settings/LanguageSettings.tsx
```diff
@@ -171,77 +127,28 @@ export function LanguageSettingsScreen({}: Props) {
...
-                  placeholder={{}}
-                  value={langPrefs.primaryLanguage}
-                  onValueChange={onChangePrimaryLanguage}
-                  items={LANGUAGES.filter(l => Boolean(l.code2)).map(l => ({
+              <Select.Root
+                value={langPrefs.primaryLanguage}
+                onValueChange={onChangePrimaryLanguage}>
+                <Select.Trigger label={_(msg`Select primary language`)}>
+                  <Select.ValueText
+                    placeholder={_(msg`Select a primary language...`)}
```
estrattonbailey, line 135: Nit: here and above, could rm the ellipsis so the translations can be merged

src/components/Select/types.ts
```diff
@@ -0,0 +1,152 @@
...
+
+export type ContentProps<T> = {
+  /**
+   * Items to render. Recommended to be in the form {value: string, label: string} - if not,
+   * you need to provide a `valueExtractor` function to extract the value from an item and
+   * customise the `Select.ValueText` component.
+   */
+  items: T[]
+  renderItem: (item: T, index: number) => React.ReactElement
+  /*
```
estrattonbailey, line 133: Nit: docblock could be nice

src/components/Select/types.ts
```diff
@@ -0,0 +1,152 @@
...
+export type ValueProps = {
+  /**
+   * Only needed for native. Extracts the label from an item. Defaults to `item => item.label`
+   */
+  children?: (value: any) => string
+  placeholder?: string
+  style?: StyleProp<TextStyle>
+}
+
+/*
```
estrattonbailey, line 120: Nit: docblock could be nice

src/components/Select/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+      )}
+    </Button>
+  )
+}
+
+export function ItemText({children}: ItemTextProps) {
+  const {selected} = useItemContext()
+  const t = useTheme()
+
+  // eslint-disable-next-line bsky-internal/avoid-unwrapped-text
```
estrattonbailey, line 275: Unneeded?

estrattonbailey (commented):
Looks great! Haven't tested native yet but looks fine

src/components/Select/index.tsx
mozzius: Needed. Lint rule complains

src/components/Select/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+        color="secondary"
+        size="small"
+        variant="solid">
+        <>{children}</>
+      </Button>
+    )
+  }
+}
+
+export function ValueText({
```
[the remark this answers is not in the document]
mozzius: I thought I'd keep it customisable. i.e. if you pass a custom trigger, you can still use the text/icon

src/components/Select/index.web.tsx
mozzius: I use it here! I needed a smaller, subtler presentation with really precise spacing

https://github.com/bluesky-social/social-app/pull/8323/files#diff-6ab0d5a445d6566163c9e1fe87a3c852bad9add682115fef6928bf829a1d14d6R40-R66

src/view/com/composer/videos/SubtitleDialog.tsx
mozzius: Web only, and I want to keep the old behaviour

src/screens/Settings/LanguageSettings.tsx
mozzius: They actually don't need a placeholder here, since they will always have a value selected

src/components/Select/index.web.tsx
mozzius: Tried this, but this breaks the scrolling. Might be preferable to just get rid of them tbh

src/components/Select/index.web.tsx
estrattonbailey: Was thinking last night that we might as well just make it look the same, but yeah harmless enough to support this, sgtm

src/components/Select/index.web.tsx
estrattonbailey: Seems to work fine for me. If you hover over them they auto-scroll the options too, so might as well keep them. I'll take a quick stab at it since I have it up already.

src/view/com/composer/videos/SubtitleDialog.tsx
estrattonbailey: We should def make sure the select and dialogs play well together but yeah, fine for now

estrattonbailey (approved):

surfdude29:
When I tap on the language selector in the PWI in mobile Safari it doesn't open:

https://github.com/user-attachments/assets/8f81b3ee-b8db-4d4b-b375-ec0d338e1cb5

PR #8327, opened 2025-05-05
swap out cropper library

haileyok:
## Why

Cropper library we use right now is totally unmaintained and isn't likely to get updated. There is no new arch support and it's unlikely to get merged in. So here we are.

## How

Somehow no one else has ever made an image cropping library that has circle support. So, again, here we are. Just a new Expo module that offers that support (and the other things we need). See https://github.com/bluesky-social/expo-image-crop-tool

## Test Plan

Make sure that cropping works in:

- User avatar upload
- User banner upload
- Post image "editing"

Shouldn't be any changes to web, so need to check iOS and Android only (after verifying that the library we use on web still works, but it should)


mozzius:
Not a big deal but fyi I touch a bunch of this code in https://github.com/bluesky-social/social-app/pull/8220 (mainly on the web side)

mozzius (changes requested):
Looks good on iOS. Immediately crashes on Android lol

src/view/com/util/UserAvatar.tsx
```diff
@@ -335,17 +337,17 @@ let EditableUserAvatar = ({
...
+        async () => {
+          const croppedImage = await openCropper({
+            imageUri: item.path,
+            shape: 'circle',
+            aspectRatio: 1,
+          })
+          onSelectNewAvatar(croppedImage)
+        },
+        isIOS ? 500 : 0,
+      )
```
mozzius, line 350: What's with the timeout?

src/view/com/util/UserAvatar.tsx
```diff
@@ -312,9 +316,7 @@ let EditableUserAvatar = ({
 
     onSelectNewAvatar(
       await openCamera({
-        width: 1000,
-        height: 1000,
-        cropperCircleOverlay: true,
```
mozzius, line 317: Have we lost the ability to do the circle overlay on the camera? (nbd, just curious)

haileyok: yea, we could add it back but massive pita…

src/view/com/util/UserAvatar.tsx
haileyok: similar to the “sheet wrapper” problem, if it tries to open too fast and the sheet it is supposed to be presenting into closes…

src/view/com/util/UserAvatar.tsx
mozzius: really not worth it, dw about it lol

src/view/com/util/UserAvatar.tsx
mozzius: :/

gaearon: Isn't this the same problem as solved my this patch here? https://github.com/bluesky-social/social-app/pull/8327/files#diff-67d5bfa1b06f18170a5050a9eba50601b62d98e43587132f41188b2f1d3bfd68L45

I suppose you could edit this logic (https://github.com/bluesky-social/expo-image-crop-tool/blob/67ad0dbc1113e77de78e67f315108fedc007cd44/ios/Cropper.swift#L138-L140) to avoid recursing if `pvc.isBeingDismissed`?

haileyok: considered that as well. from testing though it didn’t work nicely (unless we actually block) because the VC still is presenting even while isBeingDismissed is true. blocking seems not great but so does this. i’ll see if claude has an answer lol

haileyok: let me try it again though, because you're right that the patch looks like the same thing

haileyok: this might do. it feels a bit safer since we'll present in the next loop. even without doing that it might be fine (again, seems that is what we were doing?), but feels better at least. https://github.com/bluesky-social/expo-image-crop-tool/commit/ef25f6c56f60ad3d94c4b99fc5a92303ccf03d27

haileyok:
okay all the weirdness should be fixed up now and actually rfr.

mozzius (approved):
Small android bug: rotate button also rotates the fixed crop area on the banner cropper

Other than that lgtm

Screenshot 2025-05-06 at 10 41 57

haileyok:
Fixed the rotation!

PR #8342, opened 2025-05-07
Android sheets edge to edge

mozzius:
Enables the sheets to present under the nav buttons, which looks much nicer.

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

# Test plan

Test as many sheets as you can find, ensure no functional regressions

The server input/hosting provider input seemed a bit dodgy, but also seems dodgy on main too. setting that sheet to 100% height improved it significantly

### DIFF
--- a/modules/bottom-sheet/android/src/main/java/expo/modules/bottomsheet/BottomSheetView.kt
+++ b/modules/bottom-sheet/android/src/main/java/expo/modules/bottomsheet/BottomSheetView.kt
@@ -37,7 +37,7 @@ class BottomSheetView(
       val resourceId = resources.getIdentifier("navigation_bar_height", "dimen", "android")
       return if (resourceId > 0) resources.getDimensionPixelSize(resourceId) else 0
   }
-  
+
   private fun getStatusBarHeight(): Int {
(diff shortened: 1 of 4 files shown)


mozzius:
The only one that seems a bit dodgy is this one

https://github.com/user-attachments/assets/99cdd455-0580-4aa3-8fda-e3dd5a115986

mozzius:
Changed to draft - I think this messes with the keyboard handling

src/components/Dialog/index.tsx
```diff
@@ -231,7 +229,6 @@ export const ScrollableInner = React.forwardRef<ScrollView, DialogInnerProps>(
       }
       paddingBottom = Math.max(paddingBottom, tokens.space._2xl)
     } else {
-      paddingBottom += keyboardHeight
```
mozzius, line 234: Turns out this is useless because useKeyboardHandler doesn't work on Android. We need to figure out how to do this though

mozzius:
Okayyy got it working :)

mozzius:
@claude please review :)

ds-boyce (approved):

PR #8347, opened 2025-05-07
Fix notifications badge on web

jcsalterego:
### Before

main-1

main-11

main-30plus

### After

exp-1

exp-10

exp-30plus


mozzius (approved):
jerry W

PR #8349, opened 2025-05-08
Add missing <Trans> to SetNewPasswordForm.tsx

quiple:
Makes “Reset code”, “New password” on the password reset page translatable.


pfrazee:
@mozzius let's make sure to get this in next release

PR #8354, opened 2025-05-09
Live

mozzius:


src/components/live/EditLiveDialog.tsx
```diff
@@ -0,0 +1,349 @@
...
+                  if (!definitelyUrl(liveLink)) {
+                    setLiveLinkError('Invalid URL')
+                  }
+                }}
+                returnKeyType="done"
+                autoCapitalize="none"
+                autoComplete="url"
+                autoCorrect={false}
+                onSubmitEditing={() => {
+                  console.log('Submit editing', isDirty, submitDisabled)
```
mozzius, line 181: log

estrattonbailey: rip

src/components/Button.tsx
estrattonbailey, line 1: More to do here but good enough for now

src/components/Dialog/index.web.tsx
```diff
@@ -195,13 +195,7 @@ export function Inner({
           onDismiss={close}
           style={{display: 'flex', flexDirection: 'column'}}>
           {header}
-          <View
-            style={[
-              gtMobile ? a.p_2xl : a.p_xl,
-              a.overflow_hidden,
-              a.rounded_md,
```
estrattonbailey, line 202: Noticed this the other day too. Unsure if we can easily remove this, some cases like this 🤔 gonna spot fix this one in this PR, I don't know of other places this would be an issue rn.
CleanShot 2025-05-09 at 15 10 41@2x

estrattonbailey (approved):
LIVE FROM CHICAGO ITS FRIDAY NIGHT

src/components/Dialog/index.web.tsx
mozzius: Darn, ok. This was breaking the new edit profile dialog too which is why I removed it

src/components/live/GoLiveDialog.tsx
```diff
@@ -0,0 +1,352 @@
...
+  const time = useCallback(
+    (offset: number) => {
+      tick!
+
+      const date = new Date()
+      date.setMinutes(date.getMinutes() + offset)
+      return i18n
+        .date(date, {hour: 'numeric', minute: '2-digit', hour12: true})
+        .toLocaleUpperCase()
+        .replace(' ', '')
```
quiple, line 71: 스크린샷 2025-05-10 18 05 22

This removes spaces to make it look like a typo occurred in the Korean.

Is there a reason to remove spaces in English?

mozzius: Oh ok. We can stop doing that then

PR #8362, opened 2025-05-10
Don't show NUXs to brand new users

estrattonbailey:
Adds a reusable util to gate NUXs to accounts older than `n` days old.


pfrazee (approved):

cyle:
nice

Issue #6823, opened 2024-11-28
Translations for standard feed-names like "Following"


monty241:
### Steps to Reproduce

There are several feed names that also occur in resources such as in:

> These settings only apply to the Following feed.

However, "Following" is currently not translateable, just like other standard feed-names as Quiet Posters. 

Standard feed names might better be translateable to ease adaptation of generic feed-names.

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

v1.95.0

Issue #6738, opened 2024-11-25
Feed aliases/alternative titles in other languages


luan-u:
### Describe the Feature

The idea is to provide alternative spelling names for the feeds provided by Bluesky team, such as "Following", "Popular With Friends" and others.

English is not a barrier for tech people and others with some proficiency in the language, but many don't have any level of English understanding (I speak that as a Brazilian but it may apply to many others as well). Since the Bluesky feeds are very central to the app experience, I feel this could help.

The overall idea is that:
- The feed has its main name
- The feed also has alternative names in other languages (not mandatory)
- An alternative name would be used based on app language setting
- Main and alternative names can be searchable

The suggestion is, in a certain way, based on pull request https://github.com/bluesky-social/social-app/pull/5769 and, specially, the following comment:

_Originally posted by @mozzius in https://github.com/bluesky-social/social-app/pull/5769#discussion_r1819901613_
> we should probably limit it to the ones we run? but other than that it's certainly not the worst idea, my only concern is it wouldn't be consistent across the app - you'd see a different name in the feeds screen, and if we did the same fix there then you still wouldn't be able to search for them

### Describe Alternatives

- Some sort of localization as already used by the social app, but applied only to the Bluesky feeds.
  - Name translation would be handled through .po files, as already used in the app.
  - I think all the possible feed names (main + alternatives) should be searchable to not hinder progress.
- Dummy feeds which would act as a copy of main feeds. Example:
  - There would be a "Seguindo" (Following) feed for Portuguese speakers. "Seguindo" would be a feed getting its content from Following, thus acting as a copy of it. Same for other languages.
  - App would have special code to swap these feeds if the user decide to change languages.

PR #8374, opened 2025-05-16
Make `Discover` and `Following` feed names translatable on Home screen

### LINKED ISSUE #2920: Cant' translate "Following" feed
Default feed is "Following". Seems you cant translate it currently. If i change to another language, its stays "Following".

image

### LINKED ISSUE #3477: Hardcoded texts - it confuses users
1. Profile page, bio. You can translate "following, but not followers or posts.
5e2713bb-f0cb-4cf0-8b7e-f1fd06fafd7b

2. Main page "Following" feed, cant transtale. Its stranslated in many other texts, so it should be translatable. If i say "In your Following feed" "Sinun Seuratut-syötteessäsi" and "Following feed is translated, it confuses people. 
e4d66fc2-b576-496f-b76a-33c2cd2be026

surfdude29:
Several users have raised the issue that the name of the **Following** feed isn't translatable (this also applies to **Discover**):
- #2920
- #3477
- #3493
- https://github.com/bluesky-social/social-app/pull/5769#discussion_r1819889721 (@/pfrazee and @/mozzius discussion on this)
- #6738
- #6823

These are some of the last hardcoded English strings in the UI.

This PR is an updated version of #3493 (credit to @quiple for the original PR 🙏) and goes part of the way towards a solution by making the default **Discover** and **Following** feed names translatable on the Home screen.

EDIT: This PR also makes translatable the **Feeds ✨** string that is shown in the tab bar in the PWI and when a user has no pinned feeds.

Fixes #2920
Fixes #3477
Closes #3493

> [!NOTE]
> **Claude wrote part of the code for this PR and although it passes CI I have not tested it.**


src/state/queries/feed.ts
```diff
@@ -428,7 +431,12 @@ export function usePinnedFeedsInfos() {
     ],
     queryFn: async () => {
       if (!hasSession) {
```
surfdude29, line 433: After I opened this PR, I asked Claude about this and realised that this code path only runs when the user is logged out, so I think this means that **Discover** would only show up as translated in the PWI, and not when logged in?

I asked Claude how to fix it but couldn't come up with anything that looked like it would work 🤔

PR #8377, opened 2025-05-16
Share bandwidth estimate between video instances

mozzius:
Alternative title: fix video quality once and for all 😎

Based on the fix mentioned in https://github.com/video-dev/hls.js/issues/5949

Currently, every single video starts at garbage quality, and will only switch to 720p after ~5 seconds. This is because it needs to load the first segment to get a bandwidth estimate.

The fix is to share the bandwidth estimate between player instances. We just get the latest value on each fragment load, stick it in a ref (so no more rerenders needed), then read that each time we create a new player instance. 

# Before - every video starts at potato quality 😭

https://github.com/user-attachments/assets/9de85af9-ae71-47b0-8448-064f983b8b32

# After - only the first video starts at garbage quality 😎

https://github.com/user-attachments/assets/cb68ddd2-d0c2-4b59-81eb-8c0c97d1bfc5


src/view/com/util/post-embeds/BandwidthEstimateContext.tsx
```diff
@@ -0,0 +1,12 @@
+type BandwidthEstimateContext = {
+  getBandwidthEstimate: () => number | undefined
+  setLastestEstimate: (estimate: number) => void
```
surfdude29, line 3:
```suggestion
  setLatestEstimate: (estimate: number) => void
```
lil typo, also in a few other places

mozzius: lol thanks

src/view/com/util/post-embeds/VideoEmbedInner/VideoEmbedInnerWeb.tsx
```diff
@@ -231,6 +233,11 @@ function useHLS({
     })
     hlsRef.current = hls
 
+    const latestEstimate = bandwidthEstimate.getBandwidthEstimate()
+    if (latestEstimate !== undefined) {
+      hls.bandwidthEstimate = latestEstimate
```
gaearon, line 238: Since this is never used for rendering and also doesn't need to be reactive (a changing estimate doesn’t need to reevaluate effects), it would be simpler to have a module with a top-level mutable variable and a get and a set method. The rest of the code would be similar but you wouldn’t need the provider or the context. You would call get here (not during render).

mozzius: Hmm yeah - it is a global property of the device/network conditions after all!

haileyok (approved):

PR #8380, opened 2025-05-16
Upgrade @types/react to 19 + run codemod

mozzius:
Chunky commit but it's mostly types. Just upgrade to `^19` and ran the codemod

https://react.dev/blog/2024/04/25/react-19-upgrade-guide

### DIFF
--- a/package.json
+++ b/package.json
@@ -277,7 +277,7 @@
     "@expo/image-utils": "0.6.3",
     "@react-native/babel-preset": "0.79.2",
     "@react-native/normalize-colors": "0.79.2",
-    "@types/react": "^18",
+    "@types/react": "^19",
     "**/expo-constants": "17.0.3",
     "**/expo-device": "7.0.1",
     "**/zod": "3.23.8",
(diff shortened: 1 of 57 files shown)


package.json
```diff
@@ -208,7 +208,7 @@
     "react-native-web-webview": "^1.0.2",
     "react-native-webview": "13.13.5",
     "react-remove-scroll-bar": "^2.3.8",
-    "react-responsive": "^9.0.2",
+    "react-responsive": "^10.0.1",
```
mozzius, line 211: for some reason `@types/react` wasn't in dev deps - it was a transitive dependency from `@types/react-responsive@8.0.5`, which we were then using `resolutions` to force to be the right version. worst part is `react-responsive` includes types from `v9` onwards so we don't even need `@types/react-responsive`

patches/@discord+bottom-sheet+4.6.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@discord/bottom-sheet/src/hooks/useStableCallback.ts b/node_modules/@discord/bottom-sheet/src/hooks/useStableCallback.ts
+index 1c788ab..d30f330 100644
+--- a/node_modules/@discord/bottom-sheet/src/hooks/useStableCallback.ts
++++ b/node_modules/@discord/bottom-sheet/src/hooks/useStableCallback.ts
+@@ -6,7 +6,7 @@ type Callback = (...args: any[]) => any;
+  * https://gist.github.com/JakeCoxon/c7ebf6e6496f8468226fd36b596e1985
+  */
+ export const useStableCallback = (callback: Callback) => {
+-  const callbackRef = useRef<Callback>();
++  const callbackRef = useRef<Callback>(undefined);
```
mozzius, line 10: using the correct types revealed this lib is using `useRef` wrong

mozzius:
This got stale - superceded by https://github.com/bluesky-social/social-app/pull/8918

PR #8381, opened 2025-05-17
add reason to notif event log

haileyok:


mozzius (approved):
not tested, seems harmless

Issue #8387, opened 2025-05-17
can't download image after enabling permission of camera usage


mikkimoose:
### Steps to Reproduce

1. download any images on bluesky app
2. error appears

### Attachments

![Image](https://github.com/user-attachments/assets/2cab498d-1159-4320-b78c-5943c1e54801)

### What platform(s) does this occur on?

Android

### Device Info

samsung galaxy a12

### What version of the app are you using?

1.102

mozzius:
Hey! Two dumb questions

1. You were able to save images before this, right?
2. What does this screen look like for you?

<img width="380" alt="Image" src="https://github.com/user-attachments/assets/5b4c58c2-f26c-46a9-9b9e-7a8d2af83b03" />

mikkimoose:
1. right now im using previous version of bluesky and downloading images
worked
2.

hope you can help 🙏

mozzius:
so sorry, but the image didn't come through! could you try again via the web ui rather than email?

mozzius:
@mikkimoose this might be stupid, but could you go to permissions, turn media permission off and on again, and then restart the app and try again?

mikkimoose:
after upgrading my app again, i think this time it worked.

PR #8397, opened 2025-05-21
Speculative fix to Android camera roll issue

mozzius:
There's an issue with camera roll permissions as of v1.102 - see https://github.com/bluesky-social/social-app/issues/8387

This is probably not a fix, but instead just tightens up the logic around permission handling. In the existing code, if `permissionResponse` returned from the hook is `null` (i.e. has not yet been fetched), we display the "permission was denied" toast. This PR instead refetches the permission status in that case.

Also, currently after granting permission to save, we do nothing, and the user has to press the save button again themselves. This PR checks the status after requesting, and saves immediately if granted.

The logic has been moved out of `Lightbox.tsx` and into `manip.ts`, so it can be shared in the Starter Pack image save dialog

# Test plan
Test saving images when:
- Permission has not been granted yet, and:
  - Allow (image should save immediately)
  - Deny (toast will show)
- Permission has been granted (should save as normal)
- Permission has been denied
  - Should re-request if possible, but will show a different toast if "deny & don't ask again" was selected

### DIFF
--- a/src/lib/media/save-image.ts
+++ b/src/lib/media/save-image.ts
@@ -48,7 +48,7 @@ export function useSaveImageToMediaLibrary() {
           }
         } else {
           Toast.show(
-            t`Permission to access camera roll was denied. Please enable it in your system settings.`,
+            t`Permission to access your photo library was denied. Please enable it in your system settings.`,
             'xmark',
           )
         }
--- a/src/lib/media/save-image.ts
+++ b/src/lib/media/save-image.ts
@@ -42,7 +42,7 @@ export function useSaveImageToMediaLibrary() {
           } else {
             // since we've been explicitly denied, show a toast.
             Toast.show(
-              t`Images cannot be saved unless permission is granted to access the camera roll.`,
+              t`Images cannot be saved unless permission is granted to access your photo library.`,
               'xmark',
             )
           }
--- a/src/lib/media/save-image.ts
+++ b/src/lib/media/save-image.ts
@@ -1,0 +1,59 @@
+import {useCallback} from 'react'
+import * as MediaLibrary from 'expo-media-library'
+import {t} from '@lingui/macro'
+
+import {isNative} from '#/platform/detection'
+import * as Toast from '#/view/com/util/Toast'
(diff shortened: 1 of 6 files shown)


src/lib/media/save-image.ts
```diff
@@ -0,0 +1,59 @@
...
+      } else {
+        if (permission.canAskAgain) {
+          // request again once
+          const askAgain = await requestPermission()
+          if (askAgain.granted) {
+            await save()
+          } else {
+            // since we've been explicitly denied, show a toast.
+            Toast.show(
+              t`Images cannot be saved unless permission is granted to access the camera roll.`,
```
surfdude29, line 45:
```suggestion
              t`Images cannot be saved unless permission is granted to access your photo library.`,
```

src/lib/media/save-image.ts
```diff
@@ -0,0 +1,59 @@
...
+              'xmark',
+            )
+          }
+        } else {
+          Toast.show(
+            t`Permission to access camera roll was denied. Please enable it in your system settings.`,
```
surfdude29, line 51:
```suggestion
            t`Permission to access your photo library was denied. Please enable it in your system settings.`,
```

surfdude29 (commented):
Nice, I've noticed this quirk in the iOS app too so this seems like it will be more intuitive and helpful 👌

From what I can tell, Android doesn't seem to use the "camera roll" terminology, and the iOS prompt refers to the user's photo library, so it might make sense to use `your photo library` in the toasts here?

IMG_1798

src/lib/media/save-image.ts
estrattonbailey, line 1: Hot take nit: we should name this (and other hooks and utils) `useSaveImageToMediaLibrary.ts` 🤷

Makes grepping so much easier and is an easy convention to enforce.

estrattonbailey (approved):

src/lib/media/save-image.ts
mozzius: Ya sure sg

mozzius: Will do in a follow up

Issue #8371, opened 2025-05-15
Incorrect Like Count on Feed Header


so-asano:
### Steps to Reproduce

1. Go to the feed screen.
2. Like a feed.
3. At this point, the like count in the header is correct.
4. Revisit the screen.
5. The like count is now incorrect, as if your like was double-counted.

### Attachments

28 likes is correct.

![Image](https://github.com/user-attachments/assets/8f5cd3f6-0f61-49db-9d90-ee03e3379639)

29 likes is incorrect.

![Image](https://github.com/user-attachments/assets/100a970d-7e8f-4c3b-ba58-e072c6499254)

### What platform(s) does this occur on?

iOS

### What version of the app are you using?

1.101

### Additional Information

I suspect the issue might be related to the logic around this part of the code:
https://github.com/bluesky-social/social-app/blob/main/src/screens/Profile/components/ProfileFeedHeader.tsx#L98-L99

PR #8398, opened 2025-05-21
Rectify a constant +1 to liked feeds

RetroSunstar:
This sees to rectify #8371

This could have been done by simply removing the +1 but that was for instantly reflecting a users like to them

So in trying to keep that, this seems to work without issue! 🙃

🩵🩷🩵🩷


haileyok (approved):
Nice, this logic looks good. Thanks for the fix!

PR #8404, opened 2025-05-22
Quiet some logs, fix a bug

estrattonbailey:
See comments in files

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -26,7 +26,13 @@ export type NotificationReason =
   | 'chat-message'
   | 'starterpack-joined'
 
+/**
+ * Manually overridden type, but retains the possibility of
+ * `notification.request.trigger.payload` being `undefined`, as specified in
+ * the source types.
+ */
 type NotificationPayload =
+  | undefined
   | {
       reason: Exclude<NotificationReason, 'chat-message'>
       uri: string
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -231,6 +242,8 @@ export function useNotificationsHandler() {
           const payload = e.notification.request.trigger
             .payload as NotificationPayload
```
estrattonbailey, line 243: This typecast was ignoring the possibility of `undefined` as specified in the types for `notification.request.trigger.payload`

src/logger/sentry/setup/index.ts
```diff
@@ -27,5 +27,15 @@ init({
...
+  ignoreErrors: [
+    /*
+     * Unknown internals errors
+     */
+    `t is not defined`,
+    `Can't find variable: t`,
+    /*
+     * Un-useful errors
+     */
+    `Network request failed`,
```
estrattonbailey, line 39: We got ~100k of these from throughout the app, just generic network failures. We should probably be better about gating errors using `isNetworkError`, but this should be safe. Any spikes in error rates should be detected in DataDog, and we should be able to safely ignore client connections e.g. bc of flakiness.

src/state/geolocation.tsx
```diff
@@ -127,10 +132,14 @@ export async function ensureGeolocationResolved() {
...
-      resolved: device.get(['geolocation']),
-    })
+    const {success} = await geolocationResolution
+    if (success) {
+      logger.debug(`geolocation: resolved`, {
+        resolved: device.get(['geolocation']),
+      })
+    } else {
+      logger.error(`geolocation: failed to resolve`)
+    }
```
estrattonbailey, line 142: Let's report once per session. If it's still too noisy we can drop, but this might help us detect new ad blockers and things that cause spikes in failures.

mozzius (approved):
not tested, all looks sensible

PR #8413, opened 2025-05-26
Notifications for likes on reposts

mozzius:
# Based on #8414

Screenshot 2025-05-26 at 13 34 28

# Test plan

Currently only testable in dev-env

- Test you get like/repost notifs when someone likes a post you reposted (from feed)
- Test that it still works when liking/reposting from the thread view (same as #8414)

### DIFF
--- a/src/screens/VideoFeed/index.tsx
+++ b/src/screens/VideoFeed/index.tsx
@@ -1023,7 +1023,12 @@ function PlayPauseTapArea({
   const {_} = useLingui()
   const doubleTapRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const playHaptic = useHaptics()
-  const [queueLike] = usePostLikeMutationQueue(post, 'ImmersiveVideo')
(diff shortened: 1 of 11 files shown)


Sininini:
Thrilled to see this, but every notification type should have a toggle with how many there can be.

src/screens/VideoFeed/index.tsx
```diff
@@ -1023,7 +1023,12 @@ function PlayPauseTapArea({
   const {_} = useLingui()
   const doubleTapRef = useRef<ReturnType<typeof setTimeout> | null>(null)
   const playHaptic = useHaptics()
-  const [queueLike] = usePostLikeMutationQueue(post, 'ImmersiveVideo')
+  // TODO: implement viaRepost -sfn
+  const [queueLike] = usePostLikeMutationQueue(
```
estrattonbailey, line 1027: We skipping this for now?

estrattonbailey (approved):

PR #8425, opened 2025-05-29
Add semantic markup for timestamps

knowtheory:
This is a minimalist change to add `time` HTML tags with a `datetime` attribute to posts when viewed on the web.

As of right now tools like Obsidian's Web Clipper have no way of reliably accessing the `indexedAt` timestamp when invoked on a `bsky.app` page.  The Clipper's default strategy is to read a page's `meta` tags to extract author and publication info.  However, the currently loaded post may have no relationship to the initially loaded meta tags, as the user navigates to new posts which load via javascript.  The only other solution is to extract post data from the DOM.  

Extracting post data from the DOM currently isn't possible because a post's timestamp has no structured elements or attributes allowing a user or the clipper to extract it using a meaningful CSS selector.

This PR wraps the `indexedAt` values displayed w/ posts w/ a `time` element.  I _think_ this doesn't impact React Native implementations, but would appreciate further guidance on that front!

This also fixes a meta tag in `post.html` that never actually displays the right content for `profile:username`.

### DIFF
--- a/bskyweb/templates/post.html
+++ b/bskyweb/templates/post.html
@@ -11,7 +11,7 @@
 {% block html_head_extra -%}
 {%- if postView -%}
   <meta property="og:type" content="article">
-  <meta property="profile:username" content="{{ profileView.Handle }}">
+  <meta property="profile:username" content="{{ postView.Author.Handle }}">
   {%- if requestURI %}
   <meta property="og:url" content="{{ requestURI }}">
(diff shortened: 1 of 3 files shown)


knowtheory:
@mary-ext awesome. thanks for the pointer, that's exactly what i needed.  I'll go have a think about how to refactor.

PR #8429, opened 2025-05-29
Make via-repost notifs groupable

mozzius:
Sent from my iPhone

image

### DIFF
--- a/src/state/queries/notifications/util.ts
+++ b/src/state/queries/notifications/util.ts
@@ -22,7 +22,13 @@ import {
   type NotificationType,
 } from './types'
 
-const GROUPABLE_REASONS = ['like', 'repost', 'follow', 'like-via-repost', 'repost-via-repost']
+const GROUPABLE_REASONS = [
+  'like',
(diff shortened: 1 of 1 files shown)


src/state/queries/notifications/util.ts
```diff
@@ -22,7 +22,7 @@ import {
   type NotificationType,
 } from './types'
 
-const GROUPABLE_REASONS = ['like', 'repost', 'follow']
+const GROUPABLE_REASONS = ['like', 'repost', 'follow', 'like-via-repost', 'repost-via-repost']
```
surfdude29, line 25:
```suggestion
const GROUPABLE_REASONS = [
  'like',
  'repost',
  'follow',
  'like-via-repost',
  'repost-via-repost',
]
```
I think there was a missing comma after `'repost-via-repost'` in your original commit

mozzius: Ahh thanks

haileyok (approved):

PR #8431, opened 2025-05-30
invert flag for sending to statsig

haileyok:


src/view/com/posts/PostFeed.tsx
```diff
@@ -796,10 +796,14 @@ let PostFeed = ({
...
-              subject: actor.did,
-              feed,
-            })
+            logger.metric(
+              'live:view:post',
+              {
+                subject: actor.did,
+                feed,
+              },
+              {statsig: false},
```
haileyok, line 805: This one could be fairly noisy so we'll keep it quiet

src/screens/Search/Explore.tsx
```diff
@@ -959,7 +959,7 @@ export function Explore({
     }
     if (!alreadyReportedRef.current.has(module)) {
       alreadyReportedRef.current.set(module, module)
-      logger.metric('explore:module:seen', {module})
+      logger.metric('explore:module:seen', {module}, {statsig: false})
```
haileyok, line 962: Also would be very noisy

PR #8432, opened 2025-05-30
tweak ordering of explore components

haileyok:
This ended up a little chonkier than expected, so here's what all we're doing:

- Removing the "new user" special ordering. We didn't see that change much
- Moving the "Discover Feeds" module directly under the "Trending" module for all users
- Bumping the number of feeds shown here to 6 from 3 _if they are eligible_
- Switching to use the unspecced `getSuggestedFeeds` _if they are eligible_
- Recording exposures to each various feed in the list
- Recording presses into individual feeds from the list

### DIFF
--- a/src/screens/Search/Explore.tsx
+++ b/src/screens/Search/Explore.tsx
@@ -492,7 +492,11 @@ export function Explore({
               if (hasPressedLoadMoreFeeds && index < 6) {
                 continue
               }
-              logger.metric('feed:suggestion:seen', {feedUrl: item.feed.uri})
+              logger.metric(
+                'feed:suggestion:seen',
+                {feedUrl: item.feed.uri},
+                {statsig: false},
+              )
             }
           }
           if (!hasPressedLoadMoreFeeds) {
--- a/src/screens/Search/Explore.tsx
+++ b/src/screens/Search/Explore.tsx
@@ -281,9 +281,10 @@ export function Explore({
     hasPressedLoadMoreFeeds,
   ])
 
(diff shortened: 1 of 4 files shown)


src/screens/Search/Explore.tsx
```diff
@@ -599,42 +595,24 @@ export function Explore({
...
-      } else {
-        i.push(trendingTopicsModule)
-        i.push(...suggestedFollowsModule)
-        i.push(...suggestedStarterPacksModule)
-      }
-      if (gate('explore_show_suggested_feeds')) {
-        i.push(...suggestedFeedsModule)
-      }
+      i.push(trendingTopicsModule)
+      i.push(...suggestedFeedsModule)
```
haileyok, line 599: suggested feeds now above follows/sps

src/screens/Search/Explore.tsx
```diff
@@ -458,78 +454,174 @@ export function Explore({
...
+          if (hasNextFeedsPage) {
+            i.push({
+              type: 'loadMore',
+              key: 'loadMoreFeeds',
+              message: _(msg`Load more suggested feeds`),
+              isLoadingMore: isLoadingMoreFeeds,
+              onLoadMore: onLoadMoreFeeds,
+            })
+          }
+        }
```
haileyok, line 588: All of this is the same as before.

estrattonbailey (approved):
Briefly ran on web, not thoroughly tested

pfrazee (approved):
lgtm

PR #8438, opened 2025-06-02
feat: migrate the app to native tabs

okwasniewski:
This PR migrates the app to use `react-native-bottom-tabs`. 

It makes tab switching feel quite a bit faster due to using native primitives underneath.

## Before

https://github.com/user-attachments/assets/2b533c4d-667b-41b1-9e73-d85b07639575

## After 

https://github.com/user-attachments/assets/f0dd4210-d07d-45b9-90f7-ff391895d300

Note: I had to change `src/view/shell/bottom-bar/BottomBar.tsx` because Bluesky still uses React Navigation v6 which was dropped by `react-native-bottom-tabs`. After updating this can be removed.

TODO: 

- [x] Test on Android

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -4,6 +4,7 @@ import * as Notifications from 'expo-notifications'
 import {type BottomTabBarProps} from '@bottom-tabs/react-navigation'
 import {i18n, type MessageDescriptor} from '@lingui/core'
 import {msg} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
 import {
   CommonActions,
   createNavigationContainerRef,
@@ -12,6 +13,7 @@ import {
   type LinkingOptions,
   NavigationContainer,
   StackActions,
+  useNavigation,
 } from '@react-navigation/native'
 
 import {timeout} from '#/lib/async/timeout'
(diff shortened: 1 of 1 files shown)


mozzius:
Nice! Regarding React Navigation v6 - we're hoping to get #5967 merged soon

mozzius:
Merged!

fobos531:
@okwasniewski Is it not necessary to add `useFrameworks: static` as per the install guide, or at least enable modular headers for SDWebImage?

mozzius:
Yeah, seems like that's needed in my testing. Works well on Android though! (although with animations off because wew)

okwasniewski:
@fobos531 @mozzius It's not necessary, it's also in the docs: 

CleanShot 2025-06-11 at 12 16 23@2x

The app wasn't building with static frameworks enabled, some third party deps are not handling it properly. 

I've tested iOS and it worked fine, I'll work on bumping the library today evening!

okwasniewski:
In the end I had to enable static frameworks, so I had to patch two packages which didn't support it. I didn't notice that Expo Image is another dependency (both Bottom Tabs and Expo Image use the same underlying library). I'm working on some solution for this to avoid forcing users to enable static frameworks. 

But I've tested and it works now!

PR #8446, opened 2025-06-05
Double the framerate

mozzius:
Enables 120fps animations on iOS

# Test plan

- Confirms it builds ok
- See if it feels smoother on a compatible iPhone. Requires Pro/Pro Max models iiuc


vineyardbovines (approved):
yeap, smoother on my 16 pro

PR #8452, opened 2025-06-06
Replace "Note about sharing" prompt with an inline hint

mozzius:
No longer triggers the annoying prompt - just shows an inline notice instead

<table>
<tr>
<th>iOS</th>
<th>Android</th>
<th>Web</th>
</tr>
  <tr>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

Also fixed PWI post menu - currently there's zero items that show up there since we moved everything to the share menu, so it looks broken. just added a simple item that triggers the sign in prompt

<table><tr>
<th>Before</th>
<th>After</th>
</tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -1,5 +1,11 @@
-import React from 'react'
-import {Pressable, StyleProp, View, ViewStyle} from 'react-native'
+import {cloneElement, Fragment, isValidElement, useMemo} from 'react'
+import {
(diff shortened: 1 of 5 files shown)


mozzius:
Yeah, that's easy enough, will do

mozzius:
Screenshot 2025-06-06 at 18 02 45

PR #8462, opened 2025-06-08
Guard 'Push source lang to Crowdin' based on existence of CROWDIN_PERSONAL_TOKEN

jcsalterego:
The Nightly Update Source Languages workflow triggers at 0200 UTC daily, and the last step fails if the repo does not have a `CROWDIN_PERSONAL_TOKEN` secret setup. I believe this means all forks with this workflow send out nightly failure notifications.

This guards the last step, checking for an existing `CROWDIN_PERSONAL_TOKEN` before attempting to push source lang to Crowdin.

### DIFF
--- a/.github/workflows/nightly-update-source-languages.yaml
+++ b/.github/workflows/nightly-update-source-languages.yaml
@@ -47,3 +47,4 @@ jobs:
           create_pull_request: false
         env:
           CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
+        if: env.CROWDIN_PERSONAL_TOKEN != ''


surfdude29:
@estrattonbailey @mozzius Would it be possible to take a look at this at some point? It should get rid of the annoying email notifications that those of us with forked `social-app` repos get :)

surfdude29:
@jcsalterego Actually, I asked Claude about this and he thinks that this might not work, as he doesn't think it's valid YAML syntax for GitHub Actions.

He suggested that you could just make the whole workflow conditional on `bluesky-social` being the repository owner – meaning it would simply be skipped in all forks – by adding this line:

`if: github.repository_owner == 'bluesky-social'`

after this:

https://github.com/bluesky-social/social-app/blob/eac02901435d7bc79a28e0bff665352b814f9508/.github/workflows/nightly-update-source-languages.yaml#L7-L10

jcsalterego:
@surfdude29 It worked for me - https://github.com/jcsalterego/social-app/actions/runs/15518612417/job/43688805648 -but the repository owner check makes sense as well.

surfdude29:
@jcsalterego Ah Claude may well be wrong then, apologies

Issue #2406, opened 2024-01-03
Muting/blocking users should remove them from feeds immediately


wtlgo:
**Is your feature request related to a problem? Please describe.**

When I mute a user or add them to the moderation list, I expect them to disappear from my feed at the moment when I do the action.
Instead, I need to manually reload the feed for this user to disappear from it. This is especially annoying if I'm very deep into the feed, and the user is very active and will be visible in the feed many times more. This is a minor inconvenience, but it is still an inconvenience.

**Describe the solution you'd like**

Instead of scrolling back to the beginning and manually reloading the feed every single time I want to mute someone, the app should remove muted users from the feed at the same time I muted them, and I should be able to continue scrolling without seeing the muted user again.

**Additional context**

https://github.com/bluesky-social/social-app/assets/50925676/a62b4aae-dd5b-4763-ac70-60a7f52e7bca

pfrazee:
Okay cool. This can be done by adding this state to the shadow cache.

PR #8463, opened 2025-06-08
Instant Feed Update on Mute or Moderation Action

DriftPT:
Implemented #2406:  Posts from muted or blocked users are now removed immediately from the feed. This is achieved by extending the ```usePostShadow``` hook to check if the post author is muted or blocked and return ```POST_TOMBSTONE``` accordingly.
Additionally, a confirmation prompt was added to the mute action, aligning it with the existing block flow for consistency. A unit test was also added to validate the new logic.

### DIFF
--- a/src/state/cache/post-shadow.ts
+++ b/src/state/cache/post-shadow.ts
@@ -14,6 +14,7 @@ import {findAllPostsInQueryData as findAllPostsInFeedQueryData} from '#/state/qu
 import {findAllPostsInQueryData as findAllPostsInQuoteQueryData} from '#/state/queries/post-quotes'
 import {findAllPostsInQueryData as findAllPostsInThreadQueryData} from '#/state/queries/post-thread'
 import {findAllPostsInQueryData as findAllPostsInSearchQueryData} from '#/state/queries/search-posts'
+import {useProfileShadow} from './profile-shadow'
 import {castAsShadow, type Shadow} from './types'
 export type {Shadow} from './types'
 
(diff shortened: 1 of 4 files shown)


src/state/cache/post-shadow.ts
```diff
@@ -43,6 +69,10 @@ export function usePostShadow(
     setShadow(shadows.get(post))
   }
 
+  const authorShadow = useProfileShadow(post.author)
+  const wasMuted = isAuthorMuted(authorShadow)
+  const wasBlocked = isAuthorBlocked(authorShadow)
```
mozzius: would be simpler to just do
```suggestion
  const wasMuted = !!authorShadow.viewer?.muted
  const wasBlocked = !!authorShadow.viewer?.blocking
```

DriftPT: Thanks for the suggestion! I've applied the change as you proposed and updated the original commit.

surfdude29:
My two cents: I think this would definitely be an improvement, but I'm wondering if it's necessary to have a confirmation prompt when muting (or unmuting) a user.

Given that muting is private and less harsh than blocking, I feel like it adds unnecessary friction. I really like the current behaviour where you simply mute a user and get a confirmation toast, and that's it.

mozzius:
Yep agree with @surfdude29 - the extra prompt isn't necessary

DriftPT:
Appreciate the feedback! I've updated the commit and removed the confirmation prompt as suggested.

mozzius (approved):
Good stuff, thank you

PR #8472, opened 2025-06-11
Post source handling updates

estrattonbailey:
In testing this with v2, I realized that we weren't tracking reply interactions originating from the reply composer list item itself, only the reply button in `PostControls`. I started implementing that and quickly realized that consuming the post source data in the outer component won't work, because the component can re-render and evacuate the previously cached post source value from state. Also, from the outer `PostThread` component, the URI generated from the URL path params is going to contain a handle (if we nav to a thread via any of our flows e.g. feeds).

So I set about making the post source cache stable between renders, and keying it off the handle.

These commits are in rough order, and essentially what they contain is:
- some debug logs which only appear if you have your `LOG_LEVEL` set to `debug`
- moving towards using the URI w/handle as the "key" of the cache
- persisting _already consumed_ post sources and keying that data by a value from `useId` so that we can always return a source for a given mounted component, if a source is available
- sending reply interactions
- passing down the post source as `anchorPostSource` to the thread components that need it
- realizing that we don't need `useRef`, and therefore don't need a context here at all
- cleaning up `consumedSources` on unmount
- cleaning up naming and adding comments

To give an idea of exactly what's happening here, here's an example:
1. user is browsing the Discover feed
2. user clicks on a post
3. `transientSources` cache is set with `key, source`
4. thread component renders
5. `transientSources` cache is cleared for `key`, `consumedSources` cache is set for `id, source`
6. `useUnstablePostSource` returns the same `source` until the component that rendered it is unmounted
7. user clicks on the avi of the author
8. user clicks on the same post, but from within the author's profile feed
9. (repeat steps 3-6) however this `source.feed` is undefined, since it came from the author feed
10. user clicks "back" in their browser to return to thread view (of the same post) from step 3
11. 2nd thread view runs cleanup and clears `consumedSources` map of its `id`
12. `useUnstablePostSource` _still returns_ the same `source`

To test, set your `LOG_LEVEL=debug` and watch for `feed-feedback` debug logs. After step 3, you should be able to reply/like/quote/etc and see `sendInteractions` fire with the correct content. Then proceed to step 9 and try again. You should not see any `sendInteraction` logs. Then proceed to step 11, and you should once again see `sendInteraction` logs.

If a user were to continue pushing screens onto the navigation stack and end up on the Discover feed via the Feeds screen or something, this will result in another post source entry under a different `useId` value, and it will function exactly the same way. And unwinding the history stack all the way back should _still work_ too.

Potential points of discussion:
- From a feed, do we ever link to a post using a URL with a DID instead of a handle? That will result in no source being found atm.

### DIFF
--- a/src/state/unstable-post-source.tsx
+++ b/src/state/unstable-post-source.tsx
@@ -18,7 +18,7 @@ export type PostSource = {
  * A cache of sources that will be consumed by the post thread view. This is
  * cleaned up any time a source is consumed.
  */
-const transientSourcesRef = new Map<string, PostSource>()
+const transientSources = new Map<string, PostSource>()
 
 /**
  * A cache of sources that have been consumed by the post thread view. This is
@@ -26,7 +26,7 @@ const transientSourcesRef = new Map<string, PostSource>()
  * consumes a source, this is never reused unless a user navigates back to a
  * post thread view that has not been dropped from memory.
  */
-const consumedSourcesRef = new Map<string, PostSource>()
+const consumedSources = new Map<string, PostSource>()
 
 /**
  * For stashing the feed that the user was browsing when they clicked on a post.
@@ -39,7 +39,7 @@ export function setUnstablePostSource(key: string, source: PostSource) {
     `setUnstablePostSource key should be a URI containing a handle, received ${key} — use buildPostSourceKey`,
   )
   logger.debug('set', {key, source})
-  transientSourcesRef.set(key, source)
+  transientSources.set(key, source)
 }
 
 /**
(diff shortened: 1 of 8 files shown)


src/state/unstable-post-source.tsx
```diff
@@ -1,62 +1,97 @@
...
+ * Just a lil dev helper
+ */
+function assertValid(key: string, message: string) {
+  if (__DEV__) {
+    const urip = new AtUri(key)
+    if (urip.host.startsWith('did:')) {
+      throw new Error(message)
+    }
+  }
 }
```
mozzius, line 97: nit: not a huge fan of dev-only errors, would be fine as a log

estrattonbailey: Yeah, but I wanna make really sure that if we encounter one of these, we know about it. Can downgrade later if we encounter none.

PR #8477, opened 2025-06-11
Reduce thread footer height where possible

mozzius:
The thread footer needs to "prop up" the anchor post if there's no replies. However, if we have replies, it results in an excessive footer height. This uses a heuristic of `replyCount * 100` to reduce the footer down to a sensible height


mozzius:
I get what you're going for with the bookend but the implementation makes me nervous. I need to think about it

PR #8479, opened 2025-06-12
Hover card on anchor displayName/handle

mozzius:
Added hover cards, and also updated link components to use the newer version

Screenshot 2025-06-12 at 13 23 31

### DIFF
--- a/src/components/ProfileHoverCard/types.ts
+++ b/src/components/ProfileHoverCard/types.ts
@@ -3,6 +3,5 @@ import type React from 'react'
 export type ProfileHoverCardProps = {
   children: React.ReactElement
   did: string
-  inline?: boolean
   disable?: boolean
 }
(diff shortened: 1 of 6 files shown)


src/components/ProfileHoverCard/types.ts
```diff
@@ -3,6 +3,5 @@ import type React from 'react'
 export type ProfileHoverCardProps = {
   children: React.ReactElement
   did: string
-  inline?: boolean
```
estrattonbailey, line 6: This isn't used anymore

estrattonbailey (approved):
Figured we could do a single hover over displayName/handle. Could really wrap the avi too, but would require some layout refactoring, so kept it simple for now. Also removed deprecated `inline` attr, why not.

Tested on web/iOS 👍

PR #8480, opened 2025-06-12
Use Button instead of TextLink for show more button

mozzius:
Use a proper Button instead of TextLink for show more. using links as buttons is inaccessible, and lets use DRY up some repeated code in the process. Also added a layout animation

In the processes of testing, I found notifications don't limit post height, so I added it 

Screenshot 2025-06-12 at 14 05 26

# Test plan

https://bsky.app/profile/did:plc:p2cp5gopk7mgjegy6wadk3ep/post/3lrfpt4xkhc24

View this post in thread, threads v2 post + treepost, profile, search


estrattonbailey (approved):
Great idea!

Made a couple changes:
- moved and renamed to be a little more specific
- passed in style so we can match the text style in each use case
- added a little extra space proportional to the text size

Tested on web/iOS 👍

PR #8486, opened 2025-06-13
Fix translations on Android using PROCESS_TEXT intent

mozzius:
Recently, the Google Translate app has started ignoring query params when you try and link to `https://translate.google.com`. Since it redirects to the app if you have it installed, this means that you lose the text and translation becomes impossible.

Instead what we can do is send a `PROCESS_TEXT` intent, which gives the option of opening it in Google Translate. The first time you open it it gives a list of options, but if you select "always" then it opens up a floating window every time which is really nice.

If the app isn't installed, it falls back to the browser, which works fine.

https://github.com/user-attachments/assets/6e3865b5-ecd9-448a-bc44-a610bc8ad2ae


mary-ext:
the option to use another translation app is nice, but with the code as it is, isn't it very much dependent on the existence of Google Translate being installed? if it's not installed, then the app would just skip to opening on browser

(would also have to check if it actually does what's advertised, DeepL, Microsoft, Yandex and Naver has an Android app)

mozzius:
That is true. However if you don't have any translation apps installed at all it would also break. I think the ideal solution would be to use `pm.queryIntentActivities` and see if there's any apps installed that we know would work, but that's a lot higher lift that requires writing kotlin

Choosing other translators isn't the primary goal of this PR, just a nice side effect. First and foremost we need to unbreak the feature. If someone else wants to write some native code in the `bluesky-swiss-army-knife` expo module to do that I'd definitely consider it

src/lib/hooks/useTranslate.ts
```diff
@@ -0,0 +1,47 @@
...
+          // use getApplicationIconAsync to determine if the translate app is installed
+          if (
+            !(await IntentLauncher.getApplicationIconAsync(
+              'com.google.android.apps.translate',
+            ))
+          ) {
+            throw new Error('Translate app not installed')
+          }
+
+          await IntentLauncher.startActivityAsync(
```
tom-sherman, line 27: Looks like this will throw if you call it while an existing promise is still pending, causing the browser to open. Maybe use a react query mutation with [`scope`](https://tanstack.com/query/latest/docs/framework/react/guides/mutations#mutation-scopes) to avoid that?

tom-sherman:
> However if you don't have any translation apps installed at all it would also break.

You can just rely on startActivity to throw if there are no apps to handle the activity, which will fall back to opening in the browser with your existing impl.

mozzius:
> You can just rely on startActivity to throw if there are no apps to handle the activity, which will fall back to opening in the browser with your existing impl.

Annoyingly the Android Accessibility Suite shows up so that doesn't work

src/lib/hooks/useTranslate.ts
mozzius: Which promise? Like if somehow happens to get fired twice you mean?

tom-sherman: The promise returned from `startActivityAsync`, yeah.

app.config.js
```diff
@@ -240,10 +240,10 @@ module.exports = function (_config) {
         ],
         './plugins/starterPackAppClipExtension/withStarterPackAppClip.js',
         './plugins/withGradleJVMHeapSizeIncrease.js',
-        './plugins/withAndroidManifestPlugin.js',
+        './plugins/withAndroidManifestLargeHeapPlugin.js',
```
haileyok, line 237: Who up manifesting they large heaps?

mozzius:
```
       🕯️
  🕯️        🕯️

🕯️ large heap 🕯️

  🕯️        🕯️
       🕯️
```

marcomaroni-github:
Non-English users are eagerly awaiting this fix. 😬

estrattonbailey (approved):
LGTM, needs conflict fix

src/lib/hooks/useTranslate.ts
mozzius: will add a todo because we're on an old version of RQ that doesn't have scope

PR #8491, opened 2025-06-14
Use post shadow within anchor post

estrattonbailey:
Sam noticed that expanded post stat details weren't optimistically updating.


mozzius (approved):
l g t m

PR #8500, opened 2025-06-16
Fix Typos in Comments and Documentation in Hooks

leopardracer:
**Description:**  
This pull request corrects minor typos in comments and documentation within the `useDraggableScrollView.ts` and `useTimeAgo.ts` hooks. No functional changes were made; only comment text was improved for clarity and accuracy.

### DIFF
--- a/src/lib/hooks/useDraggableScrollView.ts
+++ b/src/lib/hooks/useDraggableScrollView.ts
@@ -51,7 +51,7 @@ export function useDraggableScroll<Scrollable extends ScrollView = ScrollView>({
         return
       }
 
-      // Require n pixels momement before start of drag (3 in this case )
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useDraggableScrollView.ts
```diff
@@ -51,7 +51,7 @@ export function useDraggableScroll<Scrollable extends ScrollView = ScrollView>({
         return
       }
 
-      // Require n pixels momement before start of drag (3 in this case )
+      // Require n pixels moment before start of drag (3 in this case )
```
surfdude29:
```suggestion
      // Require n pixels movement before start of drag (3 in this case)
```
Given the context, I think it probably should be `movement` rather than `moment`?

PR #8503, opened 2025-06-16
Fix v2 tree view bug caused by moderation settings

estrattonbailey:
First commit: if posts are moderated (such as a mute word) then `repliesSeenCounter` will not match `repliesIndexCounter`. However, to accurately compute what reply is the last sibling in a chain, we should be comparing if the `replyIndex` matches the actual `replyIndexCounter`, that way any omitted replies further up the chain are not taken into account.

Second commit: what if the last sibling in a chain is moderated? If that's the case, then we need to use a more implicit check for `isLastSibling`, since we know the next item in the list is not a sibling of the current item.


mozzius (approved):
Nice, muting random words in the test thread words now

PR #8516, opened 2025-06-18
Mark translatable text in `PreferenceControls.tsx`

auroursa:


surfdude29:
Nice catch @auroursa!

src/screens/Settings/NotificationSettings/components/PreferenceControls.tsx
```diff
@@ -141,7 +141,9 @@ export function Inner({
       {'filter' in preference && (
         <>
           <Divider />
-          <Text style={[a.font_bold, a.text_md]}>From</Text>
+          <Text style={[a.font_bold, a.text_md]}>
+            <Trans>From</Trans>
+          </Text>
           <Toggle.Group
             type="radio"
             label={_('Filter who you receive notifications from')}
```
surfdude29:
```suggestion
            label={_(`Filter who you receive notifications from`)}
```
I noticed while I was reviewing the PR that this string was surrounded by single quotes `''` rather than backticks `` but I thought Lingui would still extract it.

However, it doesn't appear in the list of new strings added so I guess it does need to have backticks for it to be translatable:

IMG_0762

src/screens/Settings/NotificationSettings/components/PreferenceControls.tsx
```diff
@@ -141,10 +141,12 @@ export function Inner({
...
         <>
           <Divider />
-          <Text style={[a.font_bold, a.text_md]}>From</Text>
+          <Text style={[a.font_bold, a.text_md]}>
+            <Trans>From</Trans>
+          </Text>
           <Toggle.Group
             type="radio"
-            label={_('Filter who you receive notifications from')}
+            label={_(`Filter who you receive notifications from`)}
```
quiple:
```suggestion
            label={_(msg`Filter who you receive notifications from`)}
```

quiple:
```diff
-        label={_(`Select your preferred notification channels`)}
+        label={_(msg`Select your preferred notification channels`)}
```
And it seems to be missing on line 101 too.

auroursa: I compiled and checked the result, you are right.

auroursa:
Following @quiple's suggestion, I also modified L101 and added the `msg`.

surfdude29:
I completely missed that the `msg` macro was missing 🤦🏻‍♂️

Thanks for pointing that out @quiple! And catching the other missing usage :)

So I think my point about single quotes vs. backticks was probably irrelevant and it was just the missing `msg` macro?

Issue #4749, opened 2024-07-08
Links do not carry params for bsky.app domain


oracularhades:
**Describe the bug**
When clicking links, if the link domain is bsky.app, the params are removed. Is this some sort of attempted security fix, or a bug? If it's a bug I can PR a fix, but I wanted to check. It's annoying, because when a link something like a search query, people just get the default search page on web.

**To Reproduce**

1. Post any link to bluesky with params, it just has to be on the bsky.app domain, e.g https://bsky.app/?cool=params

**Expected behavior**

The link should work with full params

**Screenshots**

No screenshot, but example post https://bsky.app/profile/oracularhades.com/post/3kwrca7bvlm2w

**Details**

Build version: 1.87.0; Bundle info: fe01f32 (prod); Bundle date: 24062805; Platform: web

**Additional context**

gaearon:
It's a bug. There was a partial fix in https://github.com/bluesky-social/social-app/pull/4574 but it needs to be fixed to work on native too.

oracularhades:
There isn't an existing issue afaik, I'll leave this open, but if another issue is found, then I'm happy to close this to avoid inflating the 'economy', lol. https://bsky.app/profile/pfrazee.com/post/3kubgsbhjpc2b

PR #8520, opened 2025-06-18
Fix linking to specific search results

### LINKED ISSUE #1384: Bug when I link to a search result within bsky
What is happening here? Chrome and Firefox on Mac

https://bsky.app/profile/manuschwendener.bsky.social/post/3k6nxu63l2o23 

When I just click on the link in the post, I get to https://bsky.app/search = the URL gets cut. 

When I open the link with right click -> 'Open Link in New Tab', I get to https://bsky.app/search**?q=bookshop**, which is the URL I pasted. 

When I open the link in a new tab with cmd&click, I get to https://bsky.app/search  

(Even better would be if the separate tab _Users_ would lead to a separate URL, but that's not urgent).

mozzius:
Fix #1384, #4749

Part of it is stopping `convertBskyAppUrlIfNeeded` from stripping the search param. this is all that's needed for web

However, for native, the search screen is only available in the search tab. so we need to add some custom logic to the link component to support switching to a child route in a different tab, which is used if `screen` is not in `routeNames`.

I opted to only add this to the new link component since that's what we use within posts.

# Test plan

Test web + native. I can't think of any scenario where this code would run other than a link to another tab? so links to `/notifications` or `/chat` should also hit it, but that should be fine.
https://bsky.app/profile/samuel.bsky.team/post/3lruzfzrios2u

### DIFF
--- a/src/lib/strings/url-helpers.ts
+++ b/src/lib/strings/url-helpers.ts
@@ -193,6 +193,11 @@ export function convertBskyAppUrlIfNeeded(url: string): string {
         return startUriToStarterPackUri(urlp.pathname)
       }
 
+      // special-case search links
+      if (urlp.pathname === '/search') {
+        return `/search?q=${urlp.searchParams.get('q')}`
+      }
+
       return urlp.pathname
     } catch (e) {
       console.error('Unexpected error in convertBskyAppUrlIfNeeded()', e)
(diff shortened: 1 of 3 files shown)


src/lib/strings/url-helpers.ts
```diff
@@ -193,6 +193,11 @@ export function convertBskyAppUrlIfNeeded(url: string): string {
         return startUriToStarterPackUri(urlp.pathname)
       }
 
+      // special-case search links
+      if (urlp.pathname === '/search') {
+        return `/search?q=${urlp.searchParams.get('q')}`
+      }
```
mozzius, line 199: technically we could use the router parsing to support arbitrary params but
a) seems like overkill
b) better to be overzealous with stripping search params

estrattonbailey (approved):

estrattonbailey:
Tested with a link to a chat too

PR #8522, opened 2025-06-18
Fix anchor follow button alignment

estrattonbailey:
Review without whitespace https://github.com/bluesky-social/social-app/pull/8522/files?diff=split&w=1


mozzius (approved):
congrats at being better at css than me

Issue #8513, opened 2025-06-17
Android share extension not working if app not open in background


jparkrr:
### Steps to Reproduce

1. Make sure the app isn't running in the background on android
2. Share some content from another app, select Bluesky in the share menu
3. Notice that I am on the timeline, not composing a post.

It works correctly if the app was left running in the background on android

### What platform(s) does this occur on?

Android

### What version of the app are you using?

Build version: 1.103.0.653; Bundle info: 3ee093ea6 (prod); Bundle date: 25061318; Platform: android; Platform version: 35; Anonymous ID: 56cc1e5a-8926-4a9a-9d8f-069e995b2572

### Additional Information

Originally reported in https://github.com/bluesky-social/social-app/issues/4799#issuecomment-2271029690, but I'm creating a new issue because that one is seemingly different.

Looking through the code, the issue seems to be that `onNewIntent` doesn't run if the activity has to start https://github.com/bluesky-social/social-app/blob/9cf457acf8495c1287e366f677e1516847e4d739/modules/expo-receive-android-intents/android/src/main/java/xyz/blueskyweb/app/exporeceiveandroidintents/ExpoReceiveAndroidIntentsModule.kt#L26-L28 I found this StackOverflow answer that may be helpful: https://stackoverflow.com/a/36942185

jparkrr:
I've created a PR to fix this: https://github.com/bluesky-social/social-app/pull/8524

kazssym:
This fix seems to introduce another bug that opens the same share screen again when the app is opened later.

PR #8524, opened 2025-06-18
Add android intent handling for app launch

jparkrr:
Fixes https://github.com/bluesky-social/social-app/issues/8513

### What's broken
Right now if you share something on android to Bluesky and the app isn't already open, the app launches but your shared content disappears into the void.

### How it works
- Added an OnCreate handler that grabs the launch intent and processes it.

The new checkForInitialIntent() method:
- Grabs whatever intent launched the app
- Waits 1 second for the app to boot
- Hands it off to the existing intent handling code

#### About that 1-second delay...
I tried to find an easy way around it, but wasn't successful. The app seemingly needs time to boot up before it can handle deep links properly.

I considered storing the intent and having the app call back when it was ready, but I wanted to keep it simple. There might be other ways to solve it that I'm missing, I'm not an expert android developer. The 1-second delay is hacky but it is way better than losing people's shared content entirely.

### Testing
I was testing using `yarn android:prod` because otherwise I believe expo dev client will get in the way.

### DIFF
--- a/modules/expo-receive-android-intents/android/src/main/java/xyz/blueskyweb/app/exporeceiveandroidintents/ExpoReceiveAndroidIntentsModule.kt
+++ b/modules/expo-receive-android-intents/android/src/main/java/xyz/blueskyweb/app/exporeceiveandroidintents/ExpoReceiveAndroidIntentsModule.kt
@@ -23,8 +23,17 @@ class ExpoReceiveAndroidIntentsModule : Module() {
     ModuleDefinition {
       Name("ExpoReceiveAndroidIntents")
 
+      // Handle new intents received while the app is already running
+      // This is called when the app receives a share intent or other ACTION_SEND intent
+      // while it's already active in memory
       OnNewIntent {
         handleIntent(it)
+      }
+
+      // Handle the initial intent when the app is first created/launched
+      // This ensures we process share intents that were used to launch the app
+      OnCreate {
+        checkForInitialIntent()
       }
     }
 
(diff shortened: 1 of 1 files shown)


jparkrr:
@haileyok wondering if you have a chance to take a quick look and let me know your thoughts. thanks!

modules/expo-receive-android-intents/android/src/main/java/xyz/blueskyweb/app/exporeceiveandroidintents/ExpoReceiveAndroidIntentsModule.kt
```diff
@@ -53,6 +62,20 @@ class ExpoReceiveAndroidIntentsModule : Module() {
...
+  private fun checkForInitialIntent() {
+    val activity = appContext.currentActivity ?: return
+    val intent = activity.intent
+
+    if (intent != null) {
+      // We delay processing by 1 second to ensure the app has fully initialized
+      // before attempting to handle the share intent
+      android.os.Handler(android.os.Looper.getMainLooper()).postDelayed({
+        handleIntent(intent)
+      }, 1000)
```
haileyok, line 75: Do you know if this is something related to RN, our implementation, or just Android in general? Wondering if 1000ms is even enough in all cases (e.g. on lower end devices). I can look around, feel like there must be some lifecycle management we can do here to avoid this.

haileyok:
Thank you! Just the question about the delay, and I'll try to look around and see if there's an elegant way to do this. But does look good otherwise, appreciate you identifying and fixing this!

modules/expo-receive-android-intents/android/src/main/java/xyz/blueskyweb/app/exporeceiveandroidintents/ExpoReceiveAndroidIntentsModule.kt
jparkrr: I was trying to keep it very simple. I don't know much about the native lifecycle, but yeah, there might be a better way.

If you know someone to ping, feel free! If we can't get answers, I still kinda think this is worth merging because it's not worse than the current behavior.

jparkrr:
Let me know if there's any changes you'd like me to make @haileyok

kazssym:
According to this https://developer.android.com/training/basics/intents/filters#HandleIntent
either onCreate or onStart can be used to handle Intents. The major difference is that onStart is called after onNewIntent, and it is useful to unify intent handling IMO.

If onStart is used for intent handling, onNewIntent must call setIntent only to update the intent in the reused activity if I understand correctly. I hope this helps resolving the Android issue.

mozzius:
Thank you a bunch for this! I attempted to fix it independently and found the same thing, except that it was fine without the timeout on real devices. There's a good chance Expo/RN has made some sort of improvements since this PR was made.

I will close this in favour of https://github.com/bluesky-social/social-app/pull/9410. Thanks again!

PR #8532, opened 2025-06-18
Couple of Sentry cleanups

estrattonbailey:


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -242,7 +242,19 @@ export function useNotificationsHandler() {
...
+            })
+            return
+          }
+          if (!payload.reason) {
+            logger.error('useNotificationsHandler: received unknown payload', {
+              payload,
+              identifier: e.notification.request.identifier,
+            })
+            return
+          }
```
estrattonbailey, line 257: Really want to understand this one better so at least let's group by known issues.

src/state/messages/convo/agent.ts
```diff
@@ -130,7 +131,7 @@ export class Convo {
 
   getSnapshot(): ConvoState {
     if (!this.snapshot) this.snapshot = this.generateSnapshot()
-    // logger.debug('Convo: snapshotted', {})
```
estrattonbailey, line 133: We don't need this prefix, this logger instance has a `context` associated with it so they're already prefixed

src/state/messages/convo/agent.ts
```diff
@@ -484,7 +485,11 @@ export class Convo {
         this.dispatch({event: ConvoDispatchEvent.Ready})
       }
     } catch (e: any) {
-      logger.error(e, {message: 'Convo: setup failed'})
+      if (!isNetworkError(e)) {
+        logger.error('setup failed', {
+          safeMessage: e.message,
+        })
+      }
```
estrattonbailey, line 492: Wrapped all these with network checks. If it's just a network error, we don't care.

mozzius (approved):

PR #8534, opened 2025-06-18
[Prerelease Backend Only, Feature Flagged] Safelink implementation. Warns users about the content of links in DMS. 

BlueSkiesAndGreenPastures:
1. Supplies a blocking criteria, warning criteria, whitelist criteria
2. Operates on most to least restrictive or permissive, whitelist + block first, etc. domain before url etc.
3. feature flagged behind safelink flag.


bskylink/src/cache/cache.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (!existing || new Date(evt.createdAt) > new Date(existing.createdAt)) {
+      this.rules.set(key, evt)
+    }
+  }
+
+  smartUpdateDomain(event: ToolsOzoneSafelinkDefs.Event) {
+    try {
+      const domain = new URL(event.url).hostname
+      event.url = domain
+
```
haileyok, line 41: Best to make `try` blocks as scoped as possible. Instead, would be better to do

```
let url
try {
	url = new URL(event.url)
} catch(e) {
	// handle that error
}

// do whatever you wanted now with the hostname
```

bskylink/src/config.ts
```diff
@@ -10,6 +14,10 @@ export type ServiceConfig = {
   version?: string
   hostnames: string[]
   appHostname: string
+  safelink: number
```
haileyok: why `number`? is there not a `envBool`? or event just `envStr('SAFELINK_ENABLED') === 'true'`

bskylink/src/config.ts
```diff
@@ -52,6 +64,10 @@ export const readEnv = (): Environment => {
     dbPostgresPoolIdleTimeoutMs: envInt(
       'LINK_DB_POSTGRES_POOL_IDLE_TIMEOUT_MS',
     ),
+    safelink: envInt('SAFELINK'),
```
haileyok: maybe `safelinkEnabled`/`SAFELINK_ENABLED` instead, more obvious

bskylink/src/bin.ts
```diff
@@ -11,9 +11,17 @@ async function main() {
...
   const link = await LinkService.create(cfg)
+
+  if (cfg.service.safelink === 1) {
+    eventCache.init(cfg.service)
+    adaptiveFetchAndUpdate()
+  }
+
   await link.start()
   httpLogger.info('link service is running')
+  console.log('running service')
```
foysalit: probably not necessary anymore?

bskylink/src/cache/cache.ts
```diff
@@ -0,0 +1,255 @@
...
+    console.error(
+      '[eventCache:fetchAndUpdateEvents] No active config, skipping actions',
+    )
+    return
+  }
+  redirectLogger.info(
+    `[eventCache] Fetching events with cursor: ${cacheCursor}`,
+  )
+  // Use current session DID instead of env variable
+  const ozoneAgent = await new OzoneAgent(cfg)
```
foysalit: this reads a bit odd to me. i don't think you can `await` a class constructor.

bskylink/src/cache/cache.ts
```diff
@@ -0,0 +1,255 @@
...
+import {redirectLogger} from '../logger.js'
+import {OzoneAgent} from './ozoneAgent.js'
+let cacheCursor: string | undefined
+
+export class EventCache {
+  private rules = new Map<string, ToolsOzoneSafelinkDefs.Event>()
+  private cfg: ServiceConfig | undefined = undefined
+
+  constructor() {
+    this.cfg = undefined
```
foysalit: we could just make cfg a required param of the constructor instead of allowing it to be undefined at instantiation. I don't see a case where we would want to instantiate the event cache without a valid config.

```
export class EventCache {
  constructor(private cfg: ServiceConfig){}
  ...
}
```

should be enough to get rid of these lines 

```
  private cfg: ServiceConfig | undefined = undefined
  this.cfg = undefined
  async getConfig(): Promise<ServiceConfig | undefined> {
    return this.cfg
  }
```

and in addition, in the various places you're accessig config with `this.cfg?.prop` can become `this.cfg.prop`

bskylink/src/cache/cache.ts
```diff
@@ -0,0 +1,255 @@
...
+      if (event.action === ToolsOzoneSafelinkDefs.REMOVERULE) {
+        // If the action is to remove the rule, delete it from the cache
+        this.insert(domain, event)
+        redirectLogger.info(
+          `[EventCache] Removed rule for domain, adjusted audit log: ${domain}`,
+        )
+        return
+      }
+      // if update should happen
+      if (event.action === ToolsOzoneSafelinkDefs.WHITELIST) {
```
foysalit: these blocks look very identical, we can probably refactor them to 1 insert call?

foysalit (commented):
This is a partial review and mostly just nit-picks about typescript but otherwise looking good. 
I have to go through the logic a bit but was wondering if it would be hard to write some tests?

bskylink/src/cache/cache.ts
```diff
@@ -0,0 +1,255 @@
...
+  for (const event of res.data.events) {
+    eventCache.smartUpdate(event)
+  }
+
+  cacheCursor = res.data?.cursor
+
+  redirectLogger.info(
+    '[eventCache:fetchAndUpdateEvents] Current cache contents:',
+  )
+  redirectLogger.info(eventCache.list())
```
foysalit: we should probably get rid of this?

bskylink/src/routes/redirect.ts
```diff
@@ -41,12 +43,245 @@ export default function (ctx: AppContext, app: Express) {
 
       res.setHeader('Cache-Control', `max-age=${(7 * DAY) / SECOND}`)
       res.type('html')
-      res.status(200)
+      res.status(302)
```
haileyok: think foysal has some broader suggestions here on cleaning up the below, but noting that we probably shouldn't ever be serving a `302` here as we're doing redirects through meta tags rather than http headers

bskylink/src/routes/createShortLink.ts
```diff
@@ -78,13 +78,14 @@ const getUrl = (ctx: AppContext, req: Request, id: string) => {
   if (!ctx.cfg.service.hostnames.length) {
     assert(req.headers.host, 'request must be made with host header')
     const baseUrl =
-      req.protocol === 'http' && req.headers.host.startsWith('localhost:')
+      req.protocol === 'http' && req.headers.host!.startsWith('localhost:')
```
haileyok: don't need `!` (the `assert` above guarantees it will be set)

haileyok:
In general, not a huge fan of serving HTML like we are here:

- Scary, easy to mess things up and introduce problems...
- Doesn't have easy access to our styles, and we are adding _new_ styles that are hidden and not easily maintained
- Not localizable

All three of these issues _can_ get resolved while keeping the HTML within this service, but still not a fan. Localization is a big thing here, and I'm very opposed to displaying warnings/blockers like this without people actually being able to read them. If we instead could just have go.bsky.app redirect back to a page in the app, I'd feel better about this. If we really don't want to do that, then we're going to need to figure out the above problems imo before proceeding.

bskylink/src/cache/cache.ts
BlueSkiesAndGreenPastures: Absolutely can!

bskylink/src/routes/redirect.ts
```diff
@@ -36,17 +38,250 @@ export default function (ctx: AppContext, app: Express) {
       ) {
         res.setHeader('Cache-Control', 'no-store')
         res.setHeader('Location', `https://${ctx.cfg.service.appHostname}`)
-        return res.status(302).end()
+        return res.status(200).end()
```
haileyok: oop sorry, may have been misleading. _this_ one we do want a `302` since we are using `Location` to send the user back to the app. anything outside of this though we want to use `200` as we're actually serving html rather than redirecting

haileyok:
```suggestion
        return res.status(302).end()
```

BlueSkiesAndGreenPastures: Gotcha! no problem!

bskylink/src/config.ts
```diff
@@ -61,6 +77,10 @@ export const envToCfg = (env: Environment): Config => {
     version: env.version,
     hostnames: env.hostnames,
     appHostname: env.appHostname || 'bsky.app',
+    safelinkEnabled: env.safelinkEnabled || false,
```
haileyok, line 80: just a lil nit, `undefined` will also be `false`, so no need for the `||`

bskylink/src/config.ts
```diff
@@ -61,6 +77,10 @@ export const envToCfg = (env: Environment): Config => {
...
+    ozoneUrl: env.ozoneUrl || undefined,
+    ozoneAgentHandle: env.ozoneAgentHandle || undefined,
+    ozoneAgentPass: env.ozoneAgentPass || undefined,
```
haileyok: similarly for these, `undefined` is the default for all of these, so don't need the `|| undefined`

haileyok: (there is also a difference between `||` and `??`. worth knowing the differences. in this case would have been fine, but e.g. if you were using a number then a value of `0` would have given you `undefined`.

```
const whatever = 0
console.log(whatever || "no") // "no"
console.log(whatever ?? "no") // "0"
```

bskylink/src/routes/redirect.ts
```diff
@@ -41,247 +44,65 @@ export default function (ctx: AppContext, app: Express) {
...
-          )
-          res.writeHead(200, {
-            'Content-Type': 'text/html',
-            'Content-Length': Buffer.byteLength(html),
-          })
-          res.end(html) // Critical - must call end()
-          return
+        if (rulePresent) {
+          switch (rulePresent.action) {
+            case ToolsOzoneSafelinkDefs.BLOCK:
```
BlueSkiesAndGreenPastures, line 68: In spec + conversation paul and I discussed the previous ordering:

Whitelist || Remove, Block, Warn, Default.

I suppose we could in theory want to block certain pages of whitelisted content, but then if there are pages that are blockable on a site, it shouldn't be whitelisted right?

BlueSkiesAndGreenPastures: so I'd say in that case, 

Whitelist, Remove, Block, Warn, None

Would be the preferred order.

haileyok: Oh, can there be multiple rules set? If so, just have to move up the other two options.

haileyok: https://github.com/bluesky-social/social-app/pull/8534/commits/3db994ac52be8fd93a8a817740f6fcd91eb35ff7

haileyok:
There's some lint errors right now, will clean them up later as I want to just run eslint on this whole thing (theres a bunch of warnings to clean up in files we are not touching)

bskylink/src/routes/redirect.ts
BlueSkiesAndGreenPastures: Because of the way I've configured domain rules, yes. A domain rule will overwrite a url rule currently as opposed to allowing an exception.

BlueSkiesAndGreenPastures: it currently matches based on
domain -> ✅ 
url with no params -> ✅ 
url with parameters -> ✅ 

So that essentially makes that logic necessary.

BlueSkiesAndGreenPastures: When it builds again the tests do account for this. 👍 so we can just revisit

PR #8543, opened 2025-06-19
Make Scrollviews automatically adapt to footer height

### LINKED ISSUE #9270: Scroll bar on iOS goes behind the bottom nav
### Steps to Reproduce

On iOS when you're scrolling, you get a gray scroll bar on the right side of your screen

When scrolling far enough down and it'll go behind the tab nav at the bottom of the screen, or while dragging it downwards if you use that to scroll (which I do)

### Attachments

_No response_

iOS

### Device Info

_No response_

Build version: 1.108.0.1113; Bundle info: 332af2b (prod); Bundle date: 25101318; Platform: ios; Platform version: 26.0.1; Anonymous ID: c911b27d-b828-4207-9749-dc4fa33b566a

### Additional Information

Just a minor inconvenience, as you'd keep having to scroll upwards to bring it back to interact with it

mozzius:
Means that the scrollbar will no longer be obscured by the footer, and we no longer need overzealous list footer padding to ensure item visibility. `List` also has a `footerExtensionHeight` prop so that we can add the height of the composer reply prompt in the post thread.

Goal is to never have to worry about the footer again! We'll probably need to go and tone down the `ListFooter` component in a few places now that it's no longer loadbearing.

I've also fixed the bug on web where, as the footer shows/hides due to viewport breakpoints, the footerHeight would remain static, leading to broken UI on web.

Observe the scrollbar in these clips:

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
        <video src="https://github.com/user-attachments/assets/cae2029e-c3c0-469e-b80d-b05ad2defd5b" controls muted loop></video>
      </td>
      <td>
        <video src="https://github.com/user-attachments/assets/f4f9254e-0c74-473d-826a-88197b2301cb" controls muted loop></video>
      </td>
    </tr>
  </tbody>
</table>

# Test plan

Go to various screens on iOS and Android and see if the bottom of the screen looks good. There should be enough space, and (on iOS) the scroll bar should not be overlapped by the bottom tab bar

**How to review the code**: the critical changes are in `shell-layout.tsx`, changing the `footerHeight` from a shared value to a normal value. Then, I've gone through and changed `footerHeight.get()` to `footerHeight`, and removed Reanimated from components where it's no longer necessary

### DIFF
--- a/src/state/shell/shell-layout.tsx
+++ b/src/state/shell/shell-layout.tsx
@@ -1,6 +1,9 @@
 import {createContext, useContext, useMemo, useState} from 'react'
 import {type SharedValue, useSharedValue} from 'react-native-reanimated'
-import {useSafeAreaInsets} from 'react-native-safe-area-context'
+import {
+  type EdgeInsets,
+  useSafeAreaInsets,
+} from 'react-native-safe-area-context'
 
 import {clamp} from '#/lib/numbers'
 import {isWeb} from '#/platform/detection'
@@ -33,16 +36,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   const insets = useSafeAreaInsets()
   const {gtMobile} = useBreakpoints()
   const [footerHeight, setFooterHeight] = useState(() =>
-    platform({
-      // try and precisely guess the footer height, then round it to 4 decimal places
-      // to remove floating point imprecision. if we can guess it exactly,
-      // we get to skip a rerender
-      native: round4dp(
-        47 + a.border.borderWidth + clamp(insets.bottom, 15, 60),
-      ),
-      web: 58,
-      default: 0,
-    }),
+    estimateInitialHeight(insets),
   )
 
   const value = useMemo(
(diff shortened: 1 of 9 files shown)


src/state/shell/shell-layout.tsx
```diff
@@ -17,33 +23,45 @@ const stateContext = React.createContext<StateContext>({
...
-    addListener() {},
-    removeListener() {},
-    modify() {},
-    get() {
-      return 0
-    },
-    set() {},
-  },
+  footerHeight: 0,
+  setFooterHeight: () => {},
```
mozzius, line 30: We need to change this from a shared value to normal state. It never changes, so no point putting it on the UI thread

fobos531:
@mozzius In the video you shared, in the first example (the post details), is it possible to avoid having the erroneous bottom padding?

mozzius:
> @mozzius In the video you shared, in the first example (the post details), is it possible to avoid having the erroneous bottom padding?

That's an unrelated issue, see https://github.com/bluesky-social/social-app/pull/8477

src/state/shell/shell-layout.tsx
```diff
@@ -17,34 +26,48 @@ const stateContext = React.createContext<StateContext>({
...
+
+function estimateInitialHeight(insets: EdgeInsets): number {
+  return platform({
+    // try and precisely guess the footer height, then round it to 4 decimal places
+    // to remove floating point imprecision. if we can guess it exactly,
+    // we get to skip a rerender
+    native: round4dp(47 + a.border.borderWidth + clamp(insets.bottom, 15, 60)),
+    web: 58,
+    default: 0,
+  })
```
mozzius, line 68: This is kinda silly - might be better to just not do this

PR #8544, opened 2025-06-20
enable fabric on iso

haileyok:
- **fix paste input problems with a patch**
- **fix bottom sheet**


rodperottoni:
CleanShot 2025-06-27 at 13 24 38@2x

I seem to still run into dialog issues after enabling the new arch + making the changes to use `RCTSurfaceTouchHandler`. 🤔

mozzius:
@haileyok I rebased to be on top of `samuel/expo-54`. you may not be surprised to learn it no longer compiles

rodperottoni:
Hey guys, I really like the implementation of bottom sheets you have. I've been trying to make it work with the New Arch for a while and only now arrived to a solution (the code in this PR still results in bottom sheets taking up the entire screen and crashing when the app is reloaded). 

Here's a gist in case you want it for Bluesky: https://gist.github.com/rodperottoni/2f68ca45167b39cf6d68e2e6b69ecd35

Things to note:
- on iOS 26+, corner radius is left unset so the new liquid glass corner radius is used instead
- I had the help of our friend Claude for some of the things you see there - mostly around the layout of subviews, etc.

Hope it helps!

CleanShot 2026-02-02 at 20 00 42@2x

mozzius:
Hey - nice, thank you!!

PR #8547, opened 2025-06-20
Catch web browser errors

mozzius:
Only logs in dev, and falls back to opening in external browser

not tested, I don't know how to trigger the error

### DIFF
--- a/src/lib/hooks/useOpenLink.ts
+++ b/src/lib/hooks/useOpenLink.ts
@@ -11,6 +11,7 @@ import {
   isRelativeUrl,
   toNiceDomain,
 } from '#/lib/strings/url-helpers'
+import {logger} from '#/logger'
 import {isNative} from '#/platform/detection'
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useOpenLink.ts
```diff
@@ -64,6 +65,10 @@ export function useOpenLink() {
               toolbarColor: t.atoms.bg.backgroundColor,
               controlsColor: t.palette.primary_500,
               createTask: false,
+            }).catch(err => {
+              if (__DEV__)
+                logger.error('Could not open web browser', {message: err})
+              Linking.openURL(url)
```
mozzius, line 71: If we don't want to change behaviour we could simply only suppress logs here and not open the fallback

estrattonbailey: If the in-app browser fails to open, I think opening the default browser is a great fallback actually. I'm not sure we even care to log this tbh but dev-only can't hurt if you want it there

PR #8549, opened 2025-06-20
Add simple heuristic for reducing footer height in thread

mozzius:
Since #8477 is a little stuck, there's a simpler heuristic for the footer in some cases - if there's no parents, the anchor post won't get pushed down so we don't need to prop it up.

Will help with #8543, which causes *even more* footer height

# Test plan

Test with different amounts of parents/children


estrattonbailey (approved):
Nice one!

PR #8555, opened 2025-06-22
Tooltip

estrattonbailey:


src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,340 @@
...
+      return {
+        top: 0,
+        bottom: 0,
+        left: 0,
+        right: 0,
+        tipTop: 0,
+        tipLeft: 0,
+      }
+
+    const {width: ww, height: wh} = Dimensions.get('window')
```
mozzius: prefer `useWindowDimensions`

mozzius:
Love it

Added an animation:

https://github.com/user-attachments/assets/c23e7dae-c7ee-4ef7-99ec-23aa65a00874

src/components/Tooltip/index.tsx
estrattonbailey: Yep will do

mozzius (approved):
lgtm! my only style nit is the popup should be offset by the size of the arrow, so that it doesn't overlap the target

Screenshot 2025-06-24 at 13 14 01

PR #8563, opened 2025-06-24
Use lingui loaders instead of compile command

mozzius:
I'm eager to remove the compile step - it slows down all our CI


timofei-iatsenko:
Hi @mozzius, Lingui maintainer here.

I identified a major bottleneck in the catalog compiler recently and released a fix in [v5.5.0](https://github.com/lingui/js-lingui/releases/tag/v5.5.0). This release also introduces multithreading across all CLI commands.

Regarding the metro-loader: I suspect it won’t actually decrease total compilation time; it likely just shifts the overhead from a separate step into the app build step itself.

The fix in v5.5.0 should make a massive difference. In my local tests, compilation time dropped from over 10 minutes (at which point I previously had to kill the process) to just 17 seconds.

For projects with large catalogs, I recommend using the standalone compile command rather than bundler plugins. The CLI can leverage multithreading, whereas most bundler loaders cannot.

mozzius:
Hey @timofei-iatsenko - awesome, I'll check that out!

Re: compile vs bundler plugins - that's good to know. It's a bit more unwieldy, especially when writing CI workflows, but sounds like it's worth the tradeoff. I have another PR open to cache translation output in CI which should also help https://github.com/bluesky-social/social-app/pull/9412

timofei-iatsenko:
@mozzius, let me know if you need any help migrating the project to the latest version of Lingui.  I’d be happy to contribute.

I’m also open to diving into the codebase to suggest improvements or open PRs for Lingui-related best practices. Just let me know if the maintainers are interested in that.

mozzius:
Thanks @timofei-iatsenko! Luckily your migration guide is pretty thorough and the codemod worked well. Just made a PR here: https://github.com/bluesky-social/social-app/pull/9905

PR #8564, opened 2025-06-25
Notifications registration

estrattonbailey:
Fixes a small issue of the push token registration endpoint being called 2x, as well as clarifies the token update callback handling. I refactored this into hooks in anticipation of upcoming needs, and left a ton of comments to make it even more clear what's going on in here.

### Testing
Assuming you allow notifications (if you don't, then obviously the token won't be registered), these are the flows that can test call cases. You'll need to set your `LOG_LEVEL=debug` and `LOG_DEBUG=notifications` before running the build process (no need to prebuild).

- [x] iOS
  - [x] From fresh install
    - [x] Log into an existing account
    - [x] Create a new account
  - [x] From an install with one account already logged in
    - [x] Log into a different account
    - [x] Create a new account
- [x] Android
  - [x] From fresh install
    - [x] Log into an existing account
    - [x] Create a new account
  - [x] From an install with one account already logged in
    - [x] Log into a different account
    - [x] Create a new account

### DIFF
--- a/src/lib/notifications/notifications.ts
+++ b/src/lib/notifications/notifications.ts
@@ -1,4 +1,4 @@
-import React from 'react'
+import {useCallback, useEffect} from 'react'
 import {Platform} from 'react-native'
 import * as Notifications from 'expo-notifications'
 import {getBadgeCountAsync, setBadgeCountAsync} from 'expo-notifications'
@@ -13,16 +13,21 @@ import BackgroundNotificationHandler from '../../../modules/expo-background-noti
 const logger = Logger.create(Logger.Context.Notifications)
 
 /**
+ * @private
  * Registers the device's push notification token with the Bluesky server.
  */
-async function _registerPushToken(
-  agent: AtpAgent,
-  account: SessionAccount,
-  token: Notifications.DevicePushToken,
-) {
+async function _registerPushToken({
+  agent,
+  currentAccount,
+  token,
+}: {
+  agent: AtpAgent
+  currentAccount: SessionAccount
+  token: Notifications.DevicePushToken
+}) {
   try {
     await agent.app.bsky.notification.registerPush({
-      serviceDid: account.service?.includes('staging')
+      serviceDid: currentAccount.service?.includes('staging')
         ? 'did:web:api.staging.bsky.dev'
         : 'did:web:api.bsky.app',
       platform: Platform.select({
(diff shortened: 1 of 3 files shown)


src/lib/notifications/notifications.ts
```diff
@@ -1,91 +1,187 @@
...
+  token: Notifications.DevicePushToken
+}) {
   try {
-    await agent.api.app.bsky.notification.registerPush({
-      serviceDid: SERVICE_DID(account.service),
-      platform: devicePlatform,
+    await agent.app.bsky.notification.registerPush({
+      serviceDid: currentAccount.service?.includes('staging')
+        ? 'did:web:api.staging.bsky.dev'
+        : 'did:web:api.bsky.app',
```
haileyok: we have this btw
https://github.com/bluesky-social/social-app/blob/bb760400feaf4bab668fc2532a4de64e6833200a/src/lib/notifications/notifications.ts#L12-L15

haileyok: oh nvm you literally just moved it, disregard. thought this was in `constants` (maybe it should be?)

src/lib/notifications/notifications.ts
```diff
@@ -1,91 +1,187 @@
...
+      platform: Platform.select({
```
haileyok: btw `Platform.OS` will give you `'ios' | 'android' | 'web'`

src/lib/notifications/notifications.ts
```diff
@@ -1,91 +1,187 @@
...
+ * Should only be called after a user has logged-in, since registration is an
+ * authed endpoint.
+ *
+ * N.B. A previous regression in `expo-notifications` caused
+ * `addPushTokenListener` to not fire on Android after calling
+ * `getPushToken()`. Therefore, as insurance, we also call
+ * `registerPushToken` here.
+ *
+ * Because `registerPushToken` is debounced, we can safely call it on every
+ * platform and only a single call will be made to the server. This does race
```
haileyok: Unclear what you mean by "call it on every platform" here.

src/lib/notifications/notifications.ts
estrattonbailey: Good shout I'll put the strings in there

src/lib/notifications/notifications.ts
estrattonbailey: 🤦

PR #8566, opened 2025-06-25
Virtualise labeler list

mozzius:
The labeller profile list of labels is a ScrollView, rather than a FlatList. [Some labellers](https://bsky.app/profile/sonasky.app) define thousands of labels, which is problematic.

This PR switches it to a FlatList. This should see an increase in perf, especially for labelers with a lot of label definitions. This can be seen in the time it takes in initial load, but on android this'll almost certainly improve scroll perf quite significantly.

# Before

https://github.com/user-attachments/assets/8a4820d8-cbb9-4f36-8c1b-6343febe6541

# After

https://github.com/user-attachments/assets/2212ae93-6cc4-45b0-881c-12b94f2808de

# Test plan

Ensure all functionality is the same, including loading and error states.

### DIFF
--- a/src/screens/Profile/Sections/Labels.tsx
+++ b/src/screens/Profile/Sections/Labels.tsx
@@ -87,7 +87,6 @@ export function ProfileLabelsSection({
       if (!labelerInfo) return null
       return (
         <View
-          key={item.identifier}
           style={[
             t.atoms.bg_contrast_25,
             index === 0 && [
@@ -123,6 +122,7 @@ export function ProfileLabelsSection({
         ref={scrollElRef}
         data={labelValues}
         renderItem={renderItem}
+        keyExtractor={keyExtractor}
         contentContainerStyle={a.px_xl}
         headerOffset={headerHeight}
         progressViewOffset={ios(0)}
@@ -144,6 +144,10 @@ export function ProfileLabelsSection({
       />
     </View>
   )
+}
+
+function keyExtractor(item: InterpretedLabelValueDefinition) {
+  return item.identifier
 }
 
 export function LabelerListHeader({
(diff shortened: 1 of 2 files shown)


src/screens/Profile/Sections/Labels.tsx
```diff
@@ -36,194 +33,205 @@ interface LabelsSectionProps {
...
+      />
+    </View>
   )
-})
+}
 
-export function ProfileLabelsSectionInner({
-  moderationOpts,
+function keyExtractor(item: InterpretedLabelValueDefinition) {
+  return item.identifier
```
haileyok, line 151: are we certain these cant be duplicated? just making sure

src/screens/Profile/Sections/Labels.tsx
```diff
@@ -36,194 +33,205 @@ interface LabelsSectionProps {
...
+    return labelerInfo.policies.labelValues
+      .map(val => lookupLabelValueDefinition(val, customDefs))
+      .filter(
+        def => def && def?.configurable,
+      ) as InterpretedLabelValueDefinition[]
+  }, [labelerInfo, labelerError, isLabelerLoading])
+
+  const numItems = labelValues.length
+
+  const renderItem = useCallback(
```
haileyok, line 86: as someone who doesn't touch this codebase much anymore unfortunately (:sob:) do we really still need to keep using `useCallback`?

haileyok (approved):
Code looks good to me.

src/screens/Profile/Sections/Labels.tsx
mozzius: In this case probably not. I really wish I had an editor feature that tells me if it's going to get automemoized or not :/ I can't trust it until then

src/screens/Profile/Sections/Labels.tsx
mozzius: Good question! The old code keys off of identifier, so I assumed so - let me check

mozzius: Nothing stops duplication - pushed up a filter to dedupe `labelValues`

PR #8567, opened 2025-06-25
A few perf params for threads v2

estrattonbailey:


src/screens/PostThread/index.tsx
```diff
@@ -551,8 +552,18 @@ export function PostThread({uri}: {uri: string}) {
             />
           }
           initialNumToRender={initialNumToRender}
-          windowSize={11}
-          sideBorders={false}
+          /**
+           * Default: 21
+           */
+          windowSize={platform({android: 11})}
```
haileyok: I think we definitely want to keep this at `11` on both iOS and Android. The difference between 11 and 21 on iOS is about 500ms of render time in a large thread for me.

PR #8572, opened 2025-06-26
fix: preserve newlines from intent text query parameter

### LINKED ISSUE #6465: Compose intent does not include newlines
### Steps to Reproduce

1. Open Compose intent with a text containing newlines. For example: https://bsky.app/intent/compose?text=hello%0A%0Aworld
2. The composer opens, but the text says "hello world"

### Attachments

grafik

Web (Desktop)

### Device Info

Firefox 131.0.3 on macOS 13.7 Ventura

Build version: 1.93.0; Bundle info: 5f449e3 (prod); Bundle date: 24111700; Platform: web

### Additional Information

I am adding support for Bluesky to my sharing buttons (see https://github.com/kytta/shareon/issues/128), and I want the text in the composer to include some text and a URL with a blank line in-between. I am not sure if this is a supported/expected use case at all :)

BobbieGoede:
Resolves #6465 
Related #6133 ? I would think we don't want to support html tags
Supersedes #6138

This PR takes the parsed text query parameter, splits it by newlines and wraps each entry in a `p` tag to mimic the editors behavior of inserting `p` tags for each newline.

I'm not familiar with the codebase and react, it feels like we're missing additional step after opening the post composer in which mentions/urls and possible newlines are parsed and handled. Right now mentions are not highlighted and the preview/og for urls in the text are not picked up on.

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -6,6 +6,7 @@ import {
   createStarterPackLinkFromAndroidReferrer,
   parseStarterPackUri,
 } from '#/lib/strings/starter-pack'
+import {tenorUrlToBskyGifUrl} from '#/state/queries/tenor'
 import {cleanError} from '../../src/lib/strings/errors'
 import {createFullHandle, makeValidHandle} from '../../src/lib/strings/handles'
 import {enforceLen} from '../../src/lib/strings/helpers'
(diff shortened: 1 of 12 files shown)


mozzius:
I don't think that would work on native unfortunately

src/lib/hooks/useIntentHandler.ts
```diff
@@ -100,11 +100,17 @@ export function useComposeIntent() {
 
       closeAllActiveElements()
 
+      // Replace newlines with <p> tags to preserve formatting in composer
+      const formattedText = text
+        ?.split('\n')
+        .map(fragment => `<p>${fragment}</p>`)
+        .join('')
```
rfreebern: Maybe it's as simple as this?

```suggestion
      const formattedText = isNative ? text : text
        ?.split('\n')
        .map(fragment => `<p>${fragment}</p>`)
        .join('')
```

PR #8578, opened 2025-06-27
Fix unnecessary loading of feeds

mozzius:
There's currently a bug with feed embeds in posts - they flash a loading state, even though they have the requisite data already. I spend a bunch of time cleaning up the legacy feed card before realising we should just swap in the new `FeedCard` component in `/Post/Embed/FeedEmbed.tsx` - really simple. But all the other tweaks have now fixed layout shifts from the skeletons in notifications and in the feeds edit screen

Reduced layout shift in notifications:

https://github.com/user-attachments/assets/f11cd865-48a9-4c12-a9c0-5ad8c0654418

Zero layout shift in Edit feeds

https://github.com/user-attachments/assets/f9ba6656-f61b-4865-bbad-6ecfaf957fc0

# Test plan

See that there's no loading state on a post with a feed embed https://bsky.app/profile/samuel.bsky.team/post/3lsl5ptbwfc2z

See that there's no layout shifts in edit feeds

### DIFF
--- a/src/components/FeedCard.tsx
+++ b/src/components/FeedCard.tsx
@@ -214,7 +214,7 @@ export function DescriptionPlaceholder() {
 export function Likes({count}: {count: number}) {
   const t = useTheme()
   return (
-    <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
+    <Text style={[a.text_sm, t.atoms.text_contrast_medium, a.font_bold]}>
       <Trans>
         Liked by <Plural value={count || 0} one="# user" other="# users" />
       </Trans>
--- a/src/view/com/feeds/FeedSourceCard.tsx
+++ b/src/view/com/feeds/FeedSourceCard.tsx
@@ -2,12 +2,17 @@ import React from 'react'
 import {
   Linking,
   Pressable,
-  StyleProp,
+  type StyleProp,
   StyleSheet,
   View,
-  ViewStyle,
+  type ViewStyle,
 } from 'react-native'
-import {AtUri} from '@atproto/api'
+import {
+  type $Typed,
+  AppBskyFeedDefs,
+  type AppBskyGraphDefs,
+  AtUri,
+} from '@atproto/api'
 import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {msg, Plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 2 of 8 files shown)


src/components/FeedCard.tsx
```diff
@@ -214,7 +214,7 @@ export function DescriptionPlaceholder() {
 export function Likes({count}: {count: number}) {
   const t = useTheme()
   return (
-    <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
+    <Text style={[a.text_sm, t.atoms.text_contrast_medium, a.font_bold]}>
```
mozzius, line 217: now matches old card style - we can change it back to normal weight if this is not wanted

src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -1,355 +1,196 @@
...
   const t = useTheme()
-  const pal = usePalette('default')
   const {_} = useLingui()
-  const removePromptControl = Prompt.usePromptControl()
-  const navigation = useNavigationDeduped()
-
-  const {isPending: isAddSavedFeedPending, mutateAsync: addSavedFeeds} =
-    useAddSavedFeedsMutation()
-  const {isPending: isRemovePending, mutateAsync: removeFeed} =
-    useRemoveFeedMutation()
```
mozzius, line 107: All this button functionality is no longer used anywhere, and massively overcomplicates the component

src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -1,355 +1,196 @@
...
-                `/profile/${feed.creatorDid}/lists/${new AtUri(feed.uri).rkey}`,
-              )
-            } else {
-              navigation.push('ProfileList', {
-                name: feed.creatorDid,
-                rkey: new AtUri(feed.uri).rkey,
-              })
-            }
-          }
-        }}
```
mozzius, line 241: wtf is going on here lol. changed to a Link component

src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -1,355 +1,196 @@
...
   showSaveBtn?: boolean
   showDescription?: boolean
   showLikes?: boolean
   pinOnSave?: boolean
   showMinimalPlaceholder?: boolean
   hideTopBorder?: boolean
-}) {
-  const {data: preferences} = usePreferencesQuery()
-  const {data: feed} = useFeedSourceInfoQuery({uri: feedUri})
+  link?: boolean
```
estrattonbailey, line 38: General question: why even keep this component around?

Stuff like this `link` prop makes me think that it should just be `<Link><FeedCard /></Link>` at the call site 🤔

estrattonbailey (commented):
Just wonder how much work it'd be to replace `FeedSourceCard` usages entirely

src/view/com/feeds/FeedSourceCard.tsx
mozzius: It's quite ridiculous stretching this component into a completely different state. Probably made sense at one point when it had the buttons?

mozzius: I'll not do it here. Agree we should do it in future though

PR #8579, opened 2025-06-27
Better dead feed handling

mozzius:
# Stacked on #8578 

Currently we just show a perma-loading state for dead feeds, which sucks. This adds a little info popup so you can actually find out what the feed is

"click for information" is a bit awkward but since it's an edge case it's probably best to be clear and awkward rather than stylish and obtuse

<table>
  <thead>
    <tr>
      <th>Web</th>
      <th>Web</th>
      <th>Native</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

Also includes handling of deleted lists

Screenshot 2025-06-27 at 14 07 28

### DIFF
--- a/src/view/com/feeds/FeedSourceCard.tsx
+++ b/src/view/com/feeds/FeedSourceCard.tsx
@@ -8,6 +8,7 @@ import {
 import {msg, Plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
+import {cleanError} from '#/lib/strings/errors'
 import {sanitizeHandle} from '#/lib/strings/handles'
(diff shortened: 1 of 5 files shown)


src/view/com/feeds/FeedSourceCard.tsx
```diff
@@ -194,3 +227,193 @@ export function FeedSourceCardLoaded({
     )
   }
 }
+
+function MissingFeed({
```
estrattonbailey: Nit: eventually (even if not now) we wanna kill this `FeedSourceCard` component right, so maybe we put this elsewhere?

PR #8582, opened 2025-06-27
Tweak string in Content Languages modal

surfdude29:
As [mentioned on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-gd/9?view=comfortable#1629), the current string shown in the **Content Languages** modal could perhaps be a little clearer, as even on web they are toggle switches rather than checkboxes:

IMG_7348

This PR proposes tweaking it from `unchecked` ➡️ `unselected`

cc: @akerbeltz


estrattonbailey (approved):
Ah nice, yep makes sense

PR #8583, opened 2025-06-27
Update copy and link in reporting flow

estrattonbailey:
Awaiting confirmation on a question, but otherwise ready to go. See Linear ticket for details.


surfdude29:
This is very much a grammatical nit and would be slightly longer, but what about:

> Need to report a copyright violation, regulatory compliance issue, or make a legal request?

(edited to fix wording)

Also, I think the lint error might be caused by the recent upgrade to Prettier using a different style for spacing in some instances?

PR #8585, opened 2025-06-29
docs: correct increment spelling

bigint:


bigint:
@mozzius what's the reason of the closure?

PR #8593, opened 2025-07-01
Remove old_postonboarding gate

mozzius:
It's stale


surfdude29:
If the gate is being removed, would it make sense to also remove the code related to the [`like-10-and-follow-7` progress guide](https://github.com/search?q=repo%3Abluesky-social%2Fsocial-app%20like-10-and-follow-7&type=code)? 🤔

mozzius:
No, I'll keep it around, it's useful as a reference

PR #8596, opened 2025-07-02
Nicer error message for backend video too large error

mozzius:
The error message from the backend if the file is somehow too large is always `file size (100000001 bytes) is larger than the maximum allowed size (100000000 bytes)`, so we should show a nicer message

### DIFF
--- a/src/view/com/composer/state/video.ts
+++ b/src/view/com/composer/state/video.ts
@@ -392,7 +392,9 @@ function getCompressErrorMessage(e: unknown, _: I18n['_']): string | null {
     return null
   }
   if (e instanceof VideoTooLargeError) {
-    return _(msg`The selected video is larger than 100 MB.`)
+    return _(
+      msg`The selected video is larger than 100 MB. Please try again with a smaller file.`,
+    )
   }
   logger.error('Error compressing video', {safeMessage: e})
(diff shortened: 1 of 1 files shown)


src/view/com/composer/state/video.ts
```diff
@@ -428,6 +428,10 @@ function getUploadErrorMessage(e: unknown, _: I18n['_']): string | null {
         return _(
           msg`Your account is not yet old enough to upload videos. Please try again later.`,
         )
+      case 'file size (100000001 bytes) is larger than the maximum allowed size (100000000 bytes)':
+        return _(
+          msg`The file size is too large. Please try again with a smaller file.`,
```
surfdude29:
```suggestion
          msg`The selected video is larger than 100 MB. Please try again with a smaller file.`,
```
We have a very similar error string on L395 that's shown when trying to upload a video that's too large, what about combining the second sentence with that? (And ideally that string would also be updated with the combined version to reduce duplication :)

https://github.com/bluesky-social/social-app/blob/dbfb9dad2f69533c79d9104d3a584791e3cf0384/src/view/com/composer/state/video.ts#L394-L395

mozzius: Good idea!

PR #8598, opened 2025-07-02
Add mechanism for running navigation actions when it's ready

mozzius:
I actually can't find an area where this'd be useful yet. So I'll leave it in draft. Maybe we can use it somewhere in future


mozzius:
Superseded

PR #8601, opened 2025-07-03
Check handle as you type

mozzius:
- Checks handle availability as you type (debounced to 500ms)
- Full handle appears as ghost preview
- New API, now suggests alternatives if the username is taken. bsky.social only, old method is used for 3rd party PDSes
- New design (partial, will hold off on major changes to the shell/navigation for now)

## Caveats

1. Nav buttons bounce up and down - this requires changing the entire onboarding flow shell, will do in separate PR
2. Web: not a popover, just placed normally. Doesn't have the nice a11y stuff. Radix doesn't have anything for this - might need to use this? https://ariakit.org/examples/combobox-radix

https://github.com/user-attachments/assets/47ab4327-b9e3-470d-86ea-d20e6b97e47e

https://github.com/user-attachments/assets/54416db6-2b34-4cfb-96eb-f68b90723838

    

## Test plan

Try a variety of handles, confirm the logic all still works - esp length checks. Check error states. Confirm you can't just skip past the availability check by pressing when it's in an odd state. Confirm ghost text works crossplatform and doesn't mess with a11y

Test with 3rd party hosting provider

Check long handles overflow properly

Screenshot 2025-07-03 at 14 43 49


src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+                        }}
+                      />
+                    )}
+                </>
+              )}
+            {(!validCheck.handleChars || !validCheck.hyphenStartOrEnd) && (
+              <Requirement>
+                {!validCheck.hyphenStartOrEnd ? (
+                  <RequirementText>
+                    <Trans>Doesn't begin or end with a hyphen</Trans>
```
surfdude29:
```suggestion
                    <Trans>Username cannot begin or end with a hyphen</Trans>
```
Given the different context this error string now appears in, I suggest rewording it slightly so that it reads a bit better

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+                  </RequirementText>
+                ) : (
+                  <RequirementText>
+                    <Trans>Only contains letters, numbers, and hyphens</Trans>
```
surfdude29:
```suggestion
                    <Trans>
                      Username can only contain basic English letters (a–z),
                      numbers, and hyphens
                    </Trans>
```
Ditto here, and I also realised something: when this string is translated, it may not be clear to some users (until they encounter an error of course) that `letters` means only the 26 basic English letters. For a German-speaking user, for example, `ß` is a valid letter, but it isn't possible to use it in a username.

So I suggest a slight clarification here might be helpful, specifying that in terms of letters, it's only `basic English letters (a–z)` that are valid. I realise that it's a bit of a mouthful so it's not exactly ideal though 🤔

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+                  </RequirementText>
+                )}
+              </Requirement>
+            )}
+            <Requirement>
+              {(!validCheck.frontLengthNotTooLong ||
+                !validCheck.totalLength) && (
+                <RequirementText>
                   <Trans>
                     No longer than{' '}
```
surfdude29:
```suggestion
                    Username cannot be longer than{' '}
```
Ditto here on rewording the error string given the different context it appears in

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
                     <Plural
                       value={MAX_SERVICE_HANDLE_LENGTH}
                       other="# characters"
                     />
                   </Trans>
-                </Text>
-              ) : (
-                <Text style={[a.text_md, a.flex_1]}>
-                  <Trans>At least 3 characters</Trans>
```
surfdude29, line 194: Would it be worth considering adding a prompt/error string (shown in the same place as the error strings are shown, although it wouldn't necessarily be in red text) saying something like `Please keep typing, your username must be at least 3 characters` when someone has only typed 1 or 2 characters?

Or even simpler, maybe just `Please keep typing`?

As an example, this is how Fastmail does it when a user is creating a new alias:

IMG_6389

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+        t.atoms.bg,
+        t.atoms.border_contrast_low,
+        a.mt_xs,
+        a.z_50,
+        a.w_full,
+        a.zoom_fade_in,
+      ]}>
+      {suggestions.map((suggestion, index) => (
+        <Button
+          label={_(msg`Select ${suggestion.handle}`)}
```
surfdude29:
```suggestion
          label={_(
            msg({
              message: `Select ${suggestion.handle}`,
              comment: `Accessibility label for a username suggestion in the account creation flow`,
            }),
          )}
```
A comment for translators here might be helpful

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+              {
+                borderBottomStartRadius: borderRadius.sm,
+                borderBottomEndRadius: borderRadius.sm,
+              },
+              a.border_b_0,
+            ],
+          ]}>
+          <Text style={[a.text_md]}>{suggestion.handle}</Text>
+          <Text style={[a.text_sm, {color: t.palette.positive_700}]}>
+            <Trans>Available</Trans>
```
surfdude29:
```suggestion
            <Trans comment="Shown next to an available username suggestion in the account creation flow">
              Available
            </Trans>
```
Ditto :)

src/screens/Signup/StepHandle.tsx
```diff
@@ -103,115 +124,212 @@ export function StepHandle() {
...
+
+function RequirementText({children}: {children: React.ReactNode}) {
   const t = useTheme()
-  if (!valid) {
-    return <Times size="md" style={{color: t.palette.negative_500}} />
-  }
-  return <Check size="md" style={{color: t.palette.positive_700}} />
+  return (
+    <Text style={[a.text_sm, a.flex_1, {color: t.palette.negative_500}]}>
+      <Trans>{children}</Trans>
```
surfdude29:
```suggestion
      {children}
```
There would be no translatable text here, so I suggest removing `Trans` tags :)

surfdude29 (commented):
This works really well on the preview branch, very nicely done! (and kudos too to the backend crew for their work on the API) 👌

I left a few string-related suggestions that came to mind after trying it out, I hope they're helpful :)

surfdude29:
One tiny other thing I noticed, although it's hardly a big deal: the **Next** button becomes active as soon as suggestions are shown, even though it can't actually be tapped/clicked until a valid, available username is actually selected.

IMG_6381

surfdude29:
A few thoughts after doing some more testing:

- I think it would be a good idea to transform any uppercase letters that the user types into lowercase ones. Currently in prod, you can create an account having typed in a username such as `SurfDude29.bsky.social` and be under the impression that your new username will have the capital letters as you typed them. After all, Twitter allows capital letters in its usernames, so imo it's a reasonable assumption. However, after your account is created, you'll find out that your new username is actually `surfdude29.bsky.social`. A small thing, but it might be a little disappointing or confusing for some users?
- On the preview branch, is it expected that pressing **Next** after a valid username has been selected doesn't load the CAPTCHA screen?
- Hardly a big deal, but the **Next** button becomes active as soon as suggestions are shown, even though it can't actually be tapped/clicked until a valid, available username is actually selected.

IMG_6381

- Definitely an edge case but FYI: once you type in more than 63 characters, the `Only contains letters, numbers, and hyphens` error is shown as well, even when you've only typed those characters.

IMG_6394

mozzius:
Appreciate the feedback @surfdude29!

estrattonbailey (approved):

PR #8602, opened 2025-07-03
Improve "replied to a post" component

mozzius:
Couple things:

- Unified into single ALF'd component (had 2 separate implementations)
- New design courtesy of Darrin, w/ new icon
- Changed copy from "Reply to xyz" to "Replied to xyz"


haileyok:
I'd say let's hold off on merging this one until after release. If we do it now, it is going to break the copy for all other languages I believe.

haileyok (changes requested):
Looks fine, let's just hold off until after release. Marking request changes so we don't merge on accident.

mozzius:
Sounds good

estrattonbailey:
I feel like the new component could be 1-2px closer to the display name, or the post text could be 1-2px further away 🤔 feels just a _little_ unbalanced to me rn

mozzius:
@estrattonbailey here's what it looks like with `pb_sm` (4px, from 2px)

Screenshot 2025-07-14 at 20 16 54

PR #8611, opened 2025-07-07
Constraint video max height to 14/9

mozzius:
Slightly reduced from 16/9, but still keeps the same aspect ratio behaviour, so there's still no cropping

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

Note how before, it barely fits on screen without overlapping the FAB

### DIFF
--- a/src/components/Post/Embed/index.tsx
+++ b/src/components/Post/Embed/index.tsx
@@ -108,21 +108,11 @@ function MediaEmbed({
       )
     }
     case 'video': {
-      let crop
-      switch (rest.viewContext) {
-        case PostEmbedViewContext.Feed:
-        case PostEmbedViewContext.FeedEmbedRecordWithMedia:
-          crop = 'constrained' as const
-          break
-        case PostEmbedViewContext.ThreadHighlighted:
-          crop = 'none' as const
-          break
-      }
       return (
         <ContentHider
           modui={rest.moderation?.ui('contentMedia')}
           activeStyle={[a.mt_sm]}>
-          <VideoEmbed embed={embed.view} crop={crop} />
+          <VideoEmbed embed={embed.view} crop="constrained" />
         </ContentHider>
       )
     }


src/view/com/util/images/AutoSizedImage.tsx
```diff
@@ -33,12 +35,13 @@ export function ConstrainedImage({
    * the height of the image.
    */
   const outerAspectRatio = React.useMemo<DimensionValue>(() => {
-    const ratio =
-      isNative || !gtMobile
+    const ratio = boundingBox
+      ? Math.min(1 / aspectRatio, boundingBox)
+      : isNative || !gtMobile
```
estrattonbailey: This loses web handling, which apparently is 1:1. I don't recall why. Not a huge deal probably.

You could alternatively do `Math.min(1 / aspectRatio, boundingBox ?? 16 / 9)` here on line 41.

mozzius: Hmm, while it was 1:1 on web in the past it no longer is. I think that codepath is unused

src/components/Post/Embed/index.tsx
```diff
@@ -108,11 +108,21 @@ function MediaEmbed({
...
     case 'video': {
+      let crop
+      switch (rest.viewContext) {
+        case PostEmbedViewContext.Feed:
+        case PostEmbedViewContext.FeedEmbedRecordWithMedia:
+          crop = 'constrained' as const
+          break
+        case PostEmbedViewContext.ThreadHighlighted:
+          crop = 'none' as const
+          break
```
estrattonbailey: Do we actually want this? Trying to recall why we did it like this 🤔 this almost matches image handling, with the exception of `FeedEmbedRecordWithMedia` which should be square if we're trying to align images/video.

estrattonbailey: Also looks like border radius for no-crop on web is broken Image

estrattonbailey: Fixed the borders, curious about your thoughts on no-crop on the highlighted post. Not sure I like it for videos, personally 🤔

src/view/com/util/images/AutoSizedImage.tsx
estrattonbailey: See my commits below, I think it's a little more clear now

src/components/Post/Embed/index.tsx
mozzius: Actually yeah I think you're right. I'll make it always be constrained I think

mozzius:
LGTM! Let's do it

PR #8614, opened 2025-07-07
Follow redirects for soundcloud shortlinks

mozzius:
Allows embed card to work for their shortlinks

## Test plan

Post https://on.soundcloud.com/AgZAAwLT3iRjNrhqp3 and confirm the player now appears

### DIFF
--- a/src/lib/link-meta/link-meta.ts
+++ b/src/lib/link-meta/link-meta.ts
@@ -1,4 +1,4 @@
-import {BskyAgent} from '@atproto/api'
+import {type BskyAgent} from '@atproto/api'
 
 import {LINK_META_PROXY} from '#/lib/constants'
 import {getGiphyMetaUri} from '#/lib/strings/embed-player'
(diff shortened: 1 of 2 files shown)


src/lib/link-meta/link-meta.ts
```diff
@@ -46,6 +47,9 @@ export async function getLinkMeta(
       url = giphyMetaUri
       urlp = new URL(url)
     }
+    // follow redirects for soundcloud shortlinks
+    // QUESTION - do we want to follow redirects in other cases? -sfn
+    shouldFollowRedirect = urlp.hostname === 'on.soundcloud.com'
```
dpreussler, line 52: it would make sense it could be as easy as soundcloud.com/ABC/ redirecting to soundcloud.com/ABC or vice versa

mozzius: Would that make a difference w.r.t. to the widget? See here:

https://github.com/bluesky-social/social-app/blob/main/src/lib/strings/embed-player.ts#L208-L230

estrattonbailey (approved):

PR #8615, opened 2025-07-07
Improve search perf (maybe), ALF sidebar search

mozzius:
This started out profiling the search inputs on web because for some reason it's super laggy for my sometimes. It seems to be re-rendering previous pages or something? didn't quite get to the bottom of it, but I added `useDeferredValue` so that the rendering of the input is prioritised over the autocomplete etc 

Then I couldn't resist touching up the sidebar search/autocomplete. Improvements include
- absolutely positioned, no more moving the sidebar up and down
- top element no longer gets hidden when loading for no reason
- animated
- empty state when input focused
- clears on pressing escape
- ALF'd the hell up

https://github.com/user-attachments/assets/8314f8bc-20c9-46c5-9503-103f7137683d

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,8 +83,9 @@
     "@emoji-mart/react": "^1.1.1",
     "@expo/html-elements": "^0.12.5",
     "@expo/webpack-config": "^19.0.1",
-    "@floating-ui/dom": "^1.6.3",
-    "@floating-ui/react-dom": "^2.0.8",
+    "@floating-ui/dom": "^1.7.5",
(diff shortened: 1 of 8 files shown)


estrattonbailey:
Maybe too much for this PR, but the main issue I have with this search component is that it's not keyboard accessible

mozzius:
Yeah, great point. Maybe a good followup?

mozzius:
Added a ticket to linear with regards to making it a real, accessible popover

PR #8619, opened 2025-07-08
Specify `autoCapitalize` value

mozzius:
Android seems to default to "none" instead of "sentences" now. Probably an issue with the paste-input library, but let's just fix it the easy way for now


estrattonbailey (approved):

layeredstrange:
Strange thing about this is, the preinstalled or Gboard (Google Keyboard), seem to be the issue for me and most likely the problem affecting the users not having autocapitalisation. 

I had to replace it with a completely different keyboard entirely.

I did some keyboard tests on Android, only 2 out of 4 don't seem to be affected from my testing on 4 Android devices... 

Autocapitalising - Swiftkey and Heliboard
Not AC-ing - Default preinstalled and Gboard

PR #8620, opened 2025-07-08
Fix perf issue on web - restore pop behaviour to tabs

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/8632

React Navigation v7 includes this change:

https://reactnavigation.org/docs/upgrading-from-6.x#the-navigate-method-no-longer-goes-back-use-popto-instead

> The `navigate` method no longer goes back, use `popTo` instead

Since our web version is a giant stack navigator, this means that switching away from a tab and back used to `pop` back to the previous screen. This awkwardly means you lose the ability to use the back button, since technically it was a back action. However, this correctly reused the same screen that you're going back to, so it doesn't need to render the screen again from scratch.

However, since the new version _always pushes_, switching to and from tabs would add a new screen _every single time_. Not only does this mean each new screen has to render from scratch, which is slow, it also means continued use of the app would accumulate more and more dead screens in memory.

You can see in the flamegraph, going from home->search->home over and over just keeps on adding more and more screens

Screenshot 2025-07-08 at 09 44 48

This PR restores the old `pop` behaviour to the tab bars, meaning the old screens are reused.

Screenshot 2025-07-08 at 10 01 20

**IMO, this is still not ideal behaviour.** The push behaviour makes much more sense in terms of the browser history, and the browser back button behaviour is much more sane. _However_, to my knowledge there's no way to force it to reused old screens without evicting the previous screens via a `pop`, other than `getId` - but that drags the old screen to the top of the stack, breaking the history (see draft PR: #8621)

# Test plan

1. Observe the profiler, confirm that on `main` screens keep getting added and that in this PR it doesn't do that
2. Just use the app for a bit, click around, confirm it doesn't slow down over time

### DIFF
--- a/src/view/com/util/Link.tsx
+++ b/src/view/com/util/Link.tsx
@@ -410,10 +410,10 @@ function onPressInner(
       const [routeName, params] = router.matchPath(href)
       if (navigationAction === 'push') {
         // @ts-ignore we're not able to type check on this one -prf
-        navigation.push(routeName, params)
+        navigation.dispatch(StackActions.push(routeName, params))
       } else if (navigationAction === 'replace') {
         // @ts-ignore we're not able to type check on this one -prf
-        navigation.replace(routeName, params)
+        navigation.dispatch(StackActions.replace(routeName, params))
       } else if (navigationAction === 'navigate') {
         const state = navigation.getState()
         const tabState = getTabState(state, routeName)
@@ -421,9 +421,9 @@ function onPressInner(
           emitSoftReset()
         } else {
           // note: 'navigate' actually acts the same as 'push' nowadays
-          // therefore we need to use 'popTo' instead -sfn
+          // therefore we need to add 'pop' -sfn
           // @ts-ignore we're not able to type check on this one -prf
-          navigation.popTo(routeName, params)
+          navigation.navigate(routeName, params, {pop: true})
         }
       } else {
         throw Error('Unsupported navigator action.')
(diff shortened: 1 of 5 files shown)


src/view/com/util/Link.tsx
```diff
@@ -411,18 +410,20 @@ function onPressInner(
...
         const state = navigation.getState()
         const tabState = getTabState(state, routeName)
         if (tabState === TabState.InsideAtRoot) {
           emitSoftReset()
         } else {
+          // note: 'navigate' actually acts the same as 'push' nowadays
+          // therefore we need to use 'popTo' instead -sfn
           // @ts-ignore we're not able to type check on this one -prf
-          navigation.navigate(routeName, params)
+          navigation.popTo(routeName, params)
```
estrattonbailey: Gotta do this in `components/Link` too

gaearon:
Have you talked to someone at React Navigation about this? I wonder what’s their take on the right solution.

mozzius:
I've asked, yep

estrattonbailey (approved):

mackuba:
So, how do other React or RN based webapps handle this?… I don't think I've seen this kind of behavior anywhere, where clicking a section in a nav bar does not push a new page on the history stack, but instead pops some pages from it. This is really not web-like… Can't it, like, add a new page on the stack and drop the old screens from memory, and then rebuild them if you press the back button?

mozzius:
Not like this. It's basically because it's reimplementing native app patterns back into the web, but that's really not the desirable behaviour in this case. I could imagine other cases where you might want this

mackuba:
Ok that sounds like a… suboptimal approach 😅

PR #8621, opened 2025-07-08
Alternative fix to web perf issue

mozzius:
This is an attempt at alternative fix to #8620

Instead of popping the tabs, which is kind of a accidental hack to tame the stack size, we can instead specify an ID for each screen to uniquely identify it in the stack. Therefore, when we push to a screen that already exists in the stack, it'll reuse the old component. If you look at the flamegraph like in #8620, it successfully only mounts the two screens and just switches between them!

However, this is not perfect at all, since it drags the screen to the top of the stack, breaking the history. It also seems to not work with the browser back button at all. What we need is a way to specify an ID, such that it reuses the old screen, without messing with the history. Also, I noticed in the flamegraph that it reorders the components, which is weird - guess it's because the stack order is changing

Putting this in draft for now.

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -729,26 +729,31 @@ const FlatNavigator = () => {
         name="Home"
         getComponent={() => HomeScreen}
         options={{title: title(msg`Home`)}}
+        getId={() => 'home'}
       />
       <Flat.Screen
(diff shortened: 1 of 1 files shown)


rodperottoni:
super interesting. Following :)

PR #8629, opened 2025-07-09
Handle notifications on first load

mozzius:
Rather than using `getInitialURL`, we actually want to read the latest notification (i.e. the notification that caused the app to open) in `onReady`, then push to it. `getInitialURL` meant that you couldn't swipe to go back to home, because it becomes the root screen of the stack.

This _should_ also fix notification handling (for both cold start and backgrounded notifs) on android, as apparently the shape of the object changed at some point. Let's get it into prod and verify - no harm if it doesn't work since the current approach doesn't work either lol

This also now handles all types of notification properly!

- like, repost, *-via-repost: takes you to the subject post
- reply, mention, quote, activity notif: takes to the post
- follow, join via starterpack: takes you to the profile
- message: takes you to the convo, including account switching logic
- verification: takes you to the verification screen

# Test plan

1. Hard close the app
2. Tap on a notification
3. See that you're taken to the right place
4. Confirm account switching works for chat messages

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -7,8 +7,8 @@ import {CommonActions, useNavigation} from '@react-navigation/native'
 import {useQueryClient} from '@tanstack/react-query'
 
 import {useAccountSwitcher} from '#/lib/hooks/useAccountSwitcher'
+import {logger as notyLogger} from '#/lib/notifications/util'
 import {type NavigationProp} from '#/lib/routes/types'
-import {Logger} from '#/logger'
 import {isAndroid, isIOS} from '#/platform/detection'
 import {useCurrentConvoId} from '#/state/messages/current-convo-id'
 import {RQKEY as RQKEY_NOTIFS} from '#/state/queries/notifications/feed'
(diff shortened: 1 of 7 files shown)


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -418,3 +355,71 @@ export function useNotificationsHandler() {
...
+    typeof e.request.trigger !== 'object' ||
+    !('type' in e.request.trigger) ||
+    e.request.trigger.type !== 'push'
+  ) {
+    return null
+  }
+
+  const payload = (
+    isIOS ? e.request.trigger.payload : e.request.content.data
+  ) as NotificationPayload
```
estrattonbailey, line 394: So we're validating `trigger` existence, do we need to do so for `content.data`? Might be missing this in the docs but I don't see any mention of why we need to handle this differently between platforms. Might be worth a comment here if there's anything unique about this logic.

estrattonbailey (approved):
Code LGTM, charging my android and will test on APK before merging 👍

src/lib/hooks/useNotificationHandler.ts
mozzius: Yeah I'm thinking we should extremely distrust the types and straight up use Zod here

PR #8635, opened 2025-07-11
Patch in upstream fix for video rotation on Android

mozzius:
Cherrypicked https://github.com/expo/expo/pull/37849

# Test plan

Try uploading this video on android: https://github.com/mozzius/aspectratio-repro/blob/main/demo-portrait.mp4

Confirm it gets the correct aspect ratio


estrattonbailey (approved):
Nice, also tested, works fine

PR #8638, opened 2025-07-12
chore(bskyembed): ESLint flat configs

43081j:
Migrates the bskyembed sub-project to use ESLint flat configs.

Also bumps the preact plugin to 2.x (which itself is now flat config only).

We can probably do this change in the other packages too, but in their own PRs.

i imagine the lock file is a hefty change because TSESLint inside the preact plugin is newer than the one we use here? and over time, as we upgrade other projects, it'll de-dupe


mozzius:
Can you fix the prettier error?

43081j:
done 👍 

not sure why the lint-staged stuff didn't catch that!

PR #8652, opened 2025-07-15
UI for age assurance compliance

estrattonbailey:
_In the UK, the Online Safety Act requires that online platforms take specific steps to reduce the risk of children seeing harmful content. Since certain parts of that act come into effect later this month, we're making some changes for people in the UK._

_See our blog post for more information: https://bsky.social/about/blog/07-10-2025-age-assurance_


src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+  control: Dialog.DialogControlProps
+}) {
+  const {_} = useLingui()
+  return (
+    <Dialog.Outer control={control}>
+      <Dialog.Handle />
+
+      <Dialog.ScrollableInner
+        label={_(
+          'Begin the age assurance process by completing the fields below.',
```
surfdude29:
```suggestion
          msg`Begin the age assurance process by completing the fields below.`,
```
I think the `msg` macro might be missing here?

src/components/ageAssurance/AgeAssuranceAccountCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    {timeAgo === 'now' ? (
+                      <Trans>Last initiated just {timeAgo}</Trans>
+                    ) : (
+                      <Trans>Last initiated {timeAgo} ago</Trans>
+                    )}
+                  </Text>
+                ) : (
+                  <Text
+                    style={[a.text_sm, a.italic, t.atoms.text_contrast_medium]}>
+                    <Trans>Age verification takes only a few minutes</Trans>
```
surfdude29:
```suggestion
                    <Trans>Age assurance only takes a few minutes</Trans>
```
nit: Most strings use `age assurance` rather than `age verification`, so it might be good to stick to that here. Also, `only takes` word order reads slightly better to me

src/components/ageAssurance/AgeAssuranceRedirectDialog.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  // TODO for testing
+  // Dialog.useAutoOpen(control.control, 3e3)
+
+  return (
+    <Dialog.Outer control={control.control}>
+      <Dialog.Handle />
+
+      <Dialog.ScrollableInner
+        label={_(msg`Verifying your age verification status`)}
```
surfdude29:
```suggestion
        label={_(msg`Verifying your age assurance status`)}
```
nit: Ditto here, might be good to stay consistent with `age assurance`

src/screens/Moderation/index.tsx
```diff
@@ -306,8 +308,15 @@ export function ModerationScreenInner({
         <Trans>Content filters</Trans>
       </Text>
 
+      <AgeAssuranceAdmonition style={[a.pb_md]}>
+        <Trans>
+          You must complete age verification in order to access the settings
```
surfdude29:
```suggestion
          You must complete age assurance in order to access the settings
```
nit: Ditto here, also suggest sticking with `age assurance`

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+            <Divider />
+
+            <View style={[a.w_full, a.pt_xl, a.gap_lg, a.pb_lg]}>
+              {wasRecentlyInitiated && (
+                <Admonition type="warning">
+                  <Trans>
+                    You initiated this flow already,{' '}
+                    {getTimeAgo(lastInitiatedAt, new Date(), {format: 'long'})}{' '}
+                    ago. It may take up to 5 minutes for emails to reach your
+                    inbox. Consider waiting a few minutes before trying again.
```
surfdude29:
```suggestion
                    inbox. Please consider waiting a few minutes before trying
                    again.
```
nit: Suggest adding `Please`

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+                {emailError ? (
+                  <Admonition type="error" style={[a.mt_sm]}>
+                    {emailError}
+                  </Admonition>
+                ) : (
+                  <Admonition type="tip" style={[a.mt_sm]}>
+                    <Trans>
+                      Use your account email, or a real email you control, in
+                      case KWS or Bluesky needs to contact you.
```
surfdude29:
```suggestion
                      Use your account email address, or another real email
                      address you control, in case KWS or Bluesky needs to
                      contact you.
```
nit: I might be overthinking this, but for clarity, especially in localization, referring specifically to `email address` might be better?

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+            msg`Please double-check that you have entered your email address correctly.`,
+          ),
+        )
+        return {status: 'maybe'}
+      }
+
+      return {status: 'valid'}
+    }
+
+    setEmailError(_(msg`Please enter a valid email.`))
```
surfdude29:
```suggestion
    setEmailError(_(msg`Please enter a valid email address.`))
```
nit: We have a very similar existing string, so perhaps `address` could be added here to re-use existing translations and avoid duplication?
https://github.com/bluesky-social/social-app/blob/60c0689236cf40ea0411185aec16f916ebf40085/src/components/dialogs/EmailDialog/screens/Update.tsx#L148

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  <Trans>Your preferred language</Trans>
+                </TextField.LabelText>
+                <LanguageSelect value={language} onChange={setLanguage} />
+              </View>
+
+              {error && <Admonition type="error">{error}</Admonition>}
+
+              <Button
+                disabled={disabled}
+                label={_(msg`Get an email`)}
```
surfdude29:
```suggestion
                label={_(msg`Begin age assurance process`)}
```
nit: Would something like this maybe be a bit clearer? 🤔

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  overridePresentation
+                  disableMismatchWarning
+                  label={urls.kwsPrivacyPolicy}
+                  to={urls.kwsPrivacyPolicy}
+                  style={[a.text_xs, a.leading_snug]}>
+                  KWS Privacy Policy
+                </InlineLinkText>
+                . This means you won’t need to verify again the next time you
+                use this email for other apps, games, and services powered by
+                KWS technology, including Bluesky.
```
surfdude29:
```suggestion
                . This means you won’t need to verify your age again the next
                time you use this email address for other apps, games, or
                services powered by KWS technology, including Bluesky.
```
nit: Suggested wording tweaks.

Also, I might be misunderstanding, but as a user, the `including Bluesky` part would raise a question in my mind because it implies that I might or will have to go through the age assurance process again in the future for Bluesky, and it might make me hesitant to proceed: "Wait, you mean this isn't a one-off, I might have to do this whole thing again?" 🤨

If you have to agree to the privacy policy anyway to continue, does it make sense to specify `including Bluesky`?

src/components/ageAssurance/const.ts
```diff
@@ -0,0 +1,5 @@
+export const urls = {
+  kwsHome: 'https://www.kidswebservices.com/en-US',
```
surfdude29, line 2:
```suggestion
  kwsHome: 'https://parents.kidswebservices.com/',
```
I think it might make sense to link to the KWS subdomain targeted at end users, rather than to the main site which is for devs :)

surfdude29 (commented):
I left a few suggestions, mostly string-related, nice job on a chunky PR 👌

src/lib/notifications/notifications.ts
```diff
@@ -140,12 +169,15 @@ export function useNotificationsRegistration() {
...
+    useAgeAssuranceContext()
 
   useEffect(() => {
     /**
-     * We want this to init right away _after_ we have a logged in user.
+     * We want this to init right away _after_ we have a logged in user, and
+     * _after_ we've loaded their age assurance state.
      */
-    if (!currentAccount) return
+    if (!currentAccount || !isAgeRestrictionReady) return
```
estrattonbailey, line 180: Probably need to handle this differently, since we need to get that listener registered ASAP

mozzius:
Pushed a fix for this overlap

Screenshot 2025-07-16 at 11 58 04

bskyweb/cmd/bskyweb/server.go
```diff
@@ -302,6 +302,7 @@ func serve(cctx *cli.Context) error {
 	e.GET("/support/copyright", server.WebGeneric)
 	e.GET("/intent/compose", server.WebGeneric)
 	e.GET("/intent/verify-email", server.WebGeneric)
+	e.GET("/intent/age-assurance", server.WebGeneric)
```
mozzius, line 305: Added this missing route

src/components/ageAssurance/AgeAssuranceAppealDialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+          msg({
+            message: 'Age assurance inquiry was submitted',
+            context: 'toast',
+          }),
+        ),
+      )
+    },
+  })
+
+  const onSubmit = React.useCallback(() => mutate(), [mutate])
```
mozzius: nit: unnecessary useCallback

src/components/ageAssurance/AgeAssuranceDismissableNotice.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (!isReady) return null
+  if (isDeclaredUnderage) return null
+  if (!isAgeRestricted) return null
+  if (lastInitiatedAt) return null
+  if (hidden) return null
+  if (nux && nux.completed) return null
+
+  return (
+    <View style={style}>
+      <View style={[a.relative]}>
```
mozzius: nit: `View` has `position: relative` by default, so this is unnecessary.

I also think instead of the style prop, it would be clearer to just wrap in a view with padding where it's used, maybe?

src/state/queries/post-feed.ts
```diff
@@ -206,7 +218,11 @@ export function usePostFeedQuery(
          * some not.
          */
         if (!agent.session) {
-          assertSomePostsPassModeration(res.feed)
+          assertSomePostsPassModeration(
+            res.feed,
+            preferences?.moderationPrefs ||
+              DEFAULT_LOGGED_OUT_PREFERENCES.moderationPrefs,
+          )
```
mozzius, line 225: What's this for? Is this asserting different logged out prefs if you're in an age restricted geo?

src/components/ageAssurance/AgeAssuranceAccountCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+                ]}>
+                {hasInitiated ? (
+                  <Text
+                    style={[a.text_sm, a.italic, t.atoms.text_contrast_medium]}
+                    title={i18n.date(lastInitiatedAt, {
+                      dateStyle: 'medium',
+                      timeStyle: 'medium',
+                    })}>
+                    {timeAgo === 'now' ? (
+                      <Trans>Last initiated just {timeAgo}</Trans>
```
mozzius: nit: this is probably easier to translate
```suggestion
                      <Trans>Last initiated just now</Trans>
```

src/components/ageAssurance/AgeAssuranceAccountCard.tsx
```diff
@@ -0,0 +1,139 @@
...
+                  gtPhone ? [a.flex_row, a.gap_xl] : [a.gap_md],
+                ]}>
+                {hasInitiated ? (
+                  <Text
+                    style={[a.text_sm, a.italic, t.atoms.text_contrast_medium]}
+                    title={i18n.date(lastInitiatedAt, {
+                      dateStyle: 'medium',
+                      timeStyle: 'medium',
+                    })}>
+                    {timeAgo === 'now' ? (
```
mozzius: `timeAgo` is a translated string, so it's only going to be `now` in english

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,304 @@
...
+                  to={urls.kwsTermsOfUse}
+                  style={[a.text_xs, a.leading_snug]}>
+                  KWS Terms of Use
+                </InlineLinkText>{' '}
+                and acknowledge that KWS will store your verified status with
+                your hashed email address in accordance with the{' '}
+                <InlineLinkText
+                  overridePresentation
+                  disableMismatchWarning
+                  label={urls.kwsPrivacyPolicy}
```
mozzius: nit: shouldn't the label be equal to the text "KWS Privacy Policy", not the href?

mozzius (changes requested):
Looking good, just found one super minor issue

src/state/ageAssurance/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const {data, isFetched, refetch} = useQuery({
+    /**
+     * This is load bearing. We always want this query to run and end in a
+     * "fetched" state, even if we fall back to defaults. This lets the rest of
+     * the app know that we've at least attempted to load the AA state.
+     *
+     * However, it only needs to run if AA is enabled.
+     */
+    enabled: isAgeAssuranceEnabled,
+    queryKey: createAgeAssuranceQueryKey(agent.session?.did ?? 'never'),
```
mozzius, line 61: question - do we want to add the geolocation `isAgeRestrictedGeo` to the query key in case it changes?

mozzius:
nit: should hide DMs in share menu when DMs are otherwise disabled

bskyweb/cmd/bskyweb/server.go
estrattonbailey: Nice thanks!

src/components/ageAssurance/AgeAssuranceAppealDialog.tsx
estrattonbailey: Ah yeah copy pasta

src/components/ageAssurance/AgeAssuranceDismissableNotice.tsx
estrattonbailey: Love the idea, but I also need nothing to render here in some cases, and a wrapper view with padding would take up space. I generally avoid this pattern, but it's pretty OK imo if you isolate how the styles are applied so that they can't compete with internal styles.

src/state/queries/post-feed.ts
estrattonbailey: If you mean `assertSomePostsPassModeration`, that's for logged-out users viewing a feed where all content is filtered. It throws a `KnownError` (which wow, forgot about these, wish we used stuff like this more) and tells the user they need to log in (and edit their mod prefs, but yeah).

estrattonbailey: Oh and I'm passing in prefs since they might have been overridden by geo-age-restriction, yeah. In reality `preferences?.moderationPrefs` should already be defined as the `DEFAULT_LOGGED_OUT_PREFERENCES` for logged out users, the fallback is just for type safety.

src/components/ageAssurance/AgeAssuranceAccountCard.tsx
estrattonbailey: 🤦

src/state/ageAssurance/index.tsx
estrattonbailey: Good shout, but the response doesn't actually vary based on location. Returns your AA state, regardless of if you're in a region that requires it or not. I think it'd only matter were that the case, and if isAgeRestrictedGeo changed midway through a session, which is unlikely.

src/state/queries/post-feed.ts
mozzius: ok sounds good

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -0,0 +1,304 @@
...
+                    </Trans>
+                  </Admonition>
+                )}
+              </View>
+
+              <View>
+                <TextField.LabelText>
+                  <Trans>Your preferred language</Trans>
+                </TextField.LabelText>
+                <LanguageSelect value={language} onChange={setLanguage} />
```
mozzius: KWS only supports a limited set of languages - this dialog should probably only allow those in this list:

Screenshot 2025-07-16 at 19 08 05

mozzius (changes requested):

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
estrattonbailey: Gonna do this on the backend

src/components/ageAssurance/AgeAssuranceDismissableNotice.tsx
mozzius: Ok fair enough. The double view looked weird at first but I follow the reasoning

mozzius (approved):
quality work

estrattonbailey:
Got some more follow-on work but gonna merge this chonky PR and open a new more slim PR with some additional things 👍

Issue #8642, opened 2025-07-13
App starts always at Notifications tab, Android


ivanbea:
### Steps to Reproduce

1- Open Bluesky App
2- It always starts with the notifications tab

I already uninstalled the app and started anew with the same abnormal result, other bluesky users report the same behavior, since the last app update.

### What platform(s) does this occur on?

Android

### Device Info

Android 2

### What version of the app are you using?

1.104.0.672

PR #8657, opened 2025-07-17
Avoid handling notifications on entry if opened via deep link

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/8642

Turns out, when opening a deep link on android, `Notifications.getLastNotificationResponseAsync()` returns something like this:

```jsonc
{
  "notification": {
    "request": {
      "trigger": {
        "channelId": null,
        "type": "push"
      },
      "content": {
        "title": null,
        "data": {
          "create_new_tab": true,
          "com.android.browser.application_id": "com.android.chrome",
          "org.chromium.chrome.browser.request_metadata_token": {
            "0": 87,
            "1": 244,
            "2": 60,
            // etc
          },
          "org.chromium.chrome.browser.eenp": [
            "xyz.blueskyweb.app"
          ]
        }
      },
      "identifier": null
    },
    "date": 0
  },
  "actionIdentifier": "expo.modules.notifications.actions.DEFAULT"
}
```

This meant we were treating it like an unknown notification and redirecting to `/notifications`, overriding the deep link. The fix is to skip the push notification entry code if the app was opened via deep link, since it conflicts.

```tsx
  async function handlePushNotificationEntry() {
    if (!isNative) return

    // deep links take precedence - on android,
    // getLastNotificationResponseAsync returns a "notification"
    // that is actually a deep link. avoid handling it twice -sfn
    if (await Linking.getInitialURL()) {
      return
    }
   
   // rest of notification handling
```

# Before

Showing the deep link handling when the app is both alive and dead

https://github.com/user-attachments/assets/131b0725-148c-446a-8512-7e3e6e6fd087

# After

  

https://github.com/user-attachments/assets/b0c44dfe-3045-4c67-bb40-d73b50b6320a

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -1,4 +1,5 @@
 import {useCallback, useRef} from 'react'
+import {Linking} from 'react-native'
 import * as Notifications from 'expo-notifications'
 import {i18n, type MessageDescriptor} from '@lingui/core'
 import {msg} from '@lingui/macro'
@@ -884,6 +885,13 @@ function RoutesContainer({children}: React.PropsWithChildren<{}>) {
   async function handlePushNotificationEntry() {
     if (!isNative) return
 
+    // deep links take precedence - on android,
+    // getLastNotificationResponseAsync returns a "notification"
+    // that is actually a deep link. avoid handling it twice -sfn
+    if (await Linking.getInitialURL()) {
+      return
+    }
+
     /**
      * The notification that caused the app to open, if applicable
      */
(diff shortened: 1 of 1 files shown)


src/Navigation.tsx
```diff
@@ -1041,39 +1049,6 @@ function reset(): Promise<void> {
...
-      resetToTab('HomeTab')
-      // @ts-ignore matchPath doesnt give us type-checked output -prf
-      navigate(name, params)
-    }
-  } else {
-    // @ts-ignore matchPath doesnt give us type-checked output -prf
-    navigate(name, params)
-  }
-}
-
```
mozzius, line 1076: This code send me down a wild goose chase, but isn't actually used anyway - let's delete it

estrattonbailey (approved):
Nice, makes sense to me

Issue #8493, opened 2025-06-14
suspended accounts are unable to see the suspended message on login


ImLunaHey:
### Steps to Reproduce

1. get suspended
2. attempt to login again
3. see the message for a split second before it redirects to the login screen

### Attachments

<img width="479" alt="Image" src="https://github.com/user-attachments/assets/c3437310-6fad-4cf5-993f-1421cf9c78a8" />

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Build version: 1.103.0; Bundle info: 3ee093e (prod); Bundle date: 25061318; Platform: web; Platform version: undefined; Anonymous ID: e8f18f27-a030-426a-a8ad-bede013f1376

### Additional Information

my main account was suspended without telling me what caused it and now i can't even login to see the message telling me it's suspended. i had to check my emails.

Jaennaet:
Telling users _how_ they ostensibly violated the TOS and what the content in question was would be much more transparent than this "you broke a secret law and we're not telling how, but here's the entire TOS – you figure out which part we think you violated"

PR #8667, opened 2025-07-18
Disable email query if takendown

mozzius:
# Note - I think #8670 is the superior fix

Fixes https://github.com/bluesky-social/social-app/issues/8493

This was causing takendown accounts to be logged out instantly, because the email system was refreshing the session, failing, and causing a logout

# Test plan

Confirm you can see the suspended screen without getting immediately logged out

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -119,7 +119,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const logoutCurrentAccount = React.useCallback<
-    SessionApiContext['logoutEveryAccount']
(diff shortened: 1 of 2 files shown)


src/state/session/index.tsx
```diff
@@ -119,7 +119,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
   )
 
   const logoutCurrentAccount = React.useCallback<
-    SessionApiContext['logoutEveryAccount']
+    SessionApiContext['logoutCurrentAccount']
```
mozzius, line 122: just noticed this has the wrong type

PR #8670, opened 2025-07-18
Replace `resumeSession` with `getSession` in the email check

mozzius:
Alternative fix to https://github.com/bluesky-social/social-app/pull/8667 - fixes https://github.com/bluesky-social/social-app/issues/8493

`resumeSession` updates root session state in the app via `persistSessionHandler`. While this worked for the purposes of the email dialog, it's not ideal, because if `resumeSession` fails with a network error or other issue, our handling of `persistSessionHandler` will bump the user back out to log in again.

Admittedly, this fallback is poor UX. But it's not new, it has existed for over a year. And by re-running `resumeSession` on window focus via RQ, the frequency that users hit this fallback has increased recently.

This PR replaces `resumeSession` with a new method `partialRefreshSession` which does almost the same thing: calls `getSession` and updates `agent.session`. However, it does not call `persistSessionHandler` if anything goes wrong, avoiding the bad fallback case.

For now, the new method only updates the couple props we need for the email dialog. Added some comments to point this out, and mention that we'll want to be careful with this.

# Test plan

Confirm that this call failing doesn't cause a logout. Easiest way to test that is by signing into to a takendown account

### DIFF
--- a/src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
+++ b/src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
@@ -30,6 +30,9 @@ export function useAccountEmailState() {
    */
   useQuery({
     enabled: !!agent.session,
+    /**
+     * Only refetch if the email verification s incomplete.
+     */
     refetchOnWindowFocus: !prevIsEmailVerified,
     queryKey: accountEmailStateQueryKey,
     queryFn: async () => {
--- a/src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
+++ b/src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
@@ -1,7 +1,7 @@
-import {useCallback, useEffect, useState} from 'react'
-import {useQuery, useQueryClient} from '@tanstack/react-query'
+import {useEffect, useMemo, useState} from 'react'
+import {useQuery} from '@tanstack/react-query'
 
-import {useAgent} from '#/state/session'
+import {useAgent, useSessionApi} from '#/state/session'
 import {emitEmailVerified} from '#/components/dialogs/EmailDialog/events'
 
 export type AccountEmailState = {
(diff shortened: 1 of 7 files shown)


src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
```diff
@@ -51,8 +51,7 @@ export function useAccountEmailState() {
     refetchOnWindowFocus: true,
     queryKey: accountEmailStateQueryKey,
     queryFn: async () => {
-      // will also trigger updates to `#/state/session` data
-      const {data} = await agent.resumeSession(agent.session!)
```
estrattonbailey, line 55: What do we lose if session state doesn't update 🤔

mozzius: You know this code better than I do, but from the look of it the email dialog doesn't hinge off of the core session management

estrattonbailey: Seems that I did this for a reason, let's just run a quick test and confirm verify email link handling didn't regress https://github.com/bluesky-social/social-app/pull/8296/commits/ec7dfeafa26caf4494ba8451f9db41b9296c68c0

mozzius: Ah interesting 🤔

src/components/dialogs/EmailDialog/data/useConfirmEmail.ts
```diff
@@ -19,11 +16,8 @@ export function useConfirmEmail() {
         email: currentAccount.email,
         token: token.trim(),
       })
-      const {data} = await agent.resumeSession(agent.session!)
-      updateAccountEmailStateQueryCache({
-        isEmailVerified: !!data.emailConfirmed,
-        email2FAEnabled: !!data.emailAuthFactor,
-      })
```
estrattonbailey, line 26: This was actually kinda silly, since `resumeSession` already updates these values in state. This was only needed because this query would lag behind root state.

src/components/dialogs/EmailDialog/data/useAccountEmailState.ts
```diff
@@ -11,57 +11,36 @@ export type AccountEmailState = {
...
-    email2FAEnabled: !!agent.session?.emailAuthFactor,
-  }
-  const query = useQuery<AccountEmailState>({
+  const state: AccountEmailState = useMemo(
+    () => ({
+      isEmailVerified: !!agent.session?.emailConfirmed,
+      email2FAEnabled: !!agent.session?.emailAuthFactor,
+    }),
+    [agent.session],
+  )
```
estrattonbailey, line 26: Because `resumeSession` updates state via `persistSessionHandler`, this will be up to date in many cases. In the case below, now we call `partialRefreshSession` which avoids calling `persistSessionHandler` (and maybe logging the user out) and instead manually updates session state via the session reducer. That way, if `partialRefreshSession` fails, no harm done.

src/lib/api/feed/home.ts
```diff
@@ -93,7 +93,7 @@ export class HomeFeedAPI implements FeedAPI {
       }
     }
 
-    if (this.usingDiscover) {
+    if (this.usingDiscover && !__DEV__) {
```
estrattonbailey, line 96: This just avoids annoying errors when using the `dev-env`, since the `dev-env` can't contact Discover. They've been there forever but it made debugging this a pain.

haileyok: wait, by dev env do you mean like a dev PDS locally? is `__DEV__` the right flag for this? i thought `__DEV__` was true just when running the _app_ locally regardless of PDS

src/state/session/reducer.ts
```diff
@@ -180,6 +185,39 @@ let reducer = (state: State, action: Action): State => {
...
+        agent.session.emailAuthFactor =
+          patch.emailAuthFactor ?? agent.session.emailAuthFactor
+      }
+
+      return {
+        ...state,
+        currentAgentState: {
+          ...state.currentAgentState,
+          agent,
+        },
```
haileyok, line 207: So, just to verify the intention, we know that we are mutating the existing `agent`. Is there a reason then that we are still adding the agent here in the `currentAgentState`? Is this just for clarity or am I missing something?

src/lib/api/feed/home.ts
estrattonbailey: Ah yeah wasn't clear. This `usingDiscover` fallback was only for old accounts who have `Following` pinned first, post-onboarding. When running the app locally we just don't need this fallback handling at all.

src/state/session/reducer.ts
estrattonbailey: For clarity, and in case anything checks equality of `currentAgentState`, seemed safer to just make it more intentional.

src/lib/api/feed/home.ts
haileyok: ah got it, thanks

haileyok (approved):

estrattonbailey:
Just gonna wait until release goes out to keep things simpler 👍

PR #8671, opened 2025-07-18
Age assurance tweaks

estrattonbailey:
Refetch user's status on window focus if (1) they've initiated a flow before and (2) their `status = pending`.

Also adds a success state to the polling dialog they see after a successful deeplink. Before the state updates and closing dialog was kinda janky. This updates state in the bg and lets the user close the dialog.


mozzius (approved):
Not tested but looks good at a glance

estrattonbailey:
Tested e2e in prod on device, should be good 👍

Issue #7285, opened 2024-12-26
Send "show less/more" interactions to third-party feed generators


rolandcrosby:
### Describe the Feature

When a third-party feed generator sets `acceptsInteractions` to true, the Bluesky app should show the "show less/more like this" menu items for posts in the generator's feeds. When a user selects one of those menu items, a corresponding `app.bsky.feed.sendInteractions` request should be sent to the feed generator.

The real request here: support for this should be added independently from enabling sharing of UI event data with third-party feed generators. #3498 doesn't distinguish between these types of feedback and doesn't support either for third-party feeds. While the "show less/more" interactions and UI interaction events do use the same API endpoint, the user expectations and necessary privacy controls are very different. A user clicking "show less like this" is directly trying to send a signal to the software generating the feed. There would be no reason to provide an "opt out" of sharing "show more/less like this" data with feed generators - if you don't want the feed generator to receive the data point "this user wants to see less like this", just don't click the "tell the feed generator I want to see less like this" menu item.

### Additional Context

Feed generators can set the `acceptsInteractions` boolean to indicate that clients can send feed interaction data to the [`app.bsky.feed.sendInteractions`](https://docs.bsky.app/docs/api/app-bsky-feed-send-interactions) endpoint. The [interactions implemented today](https://github.com/bluesky-social/atproto/blob/22a96d6b7459dbfd88c289ace88ec823e9adc6ee/lexicons/app/bsky/feed/defs.json#L241-L287) fall into a few categories:
- direct user interactions with the feed itself (`requestLess`, `requestMore`)
- events tracking user interactions with posts in the feed (`interactionLike`, `interactionReply`, etc)
- events tracking what the user viewed in the feed and where they navigated (`interactionSeen`, `clickthroughAuthor`, etc)

This has not yet been implemented for third party feed generators. #3498 explains that this is because it's important to get the permissions right. That seems like a big project: sending user UI interaction events to third parties in this way is a novel and sensitive problem.

andrioid:
I think the direct interactions bring most value to those of us who are creating custom feed services and pose the least privacy threat.

The user-tracking ones could be limited to official services, or require an opt-in from the user.

kaytwo:
Those three categories seem to fall pretty cleanly into three different sensitivity bins:

* Public user interactions that can be pulled off the firehose (with a little information loss regarding whether they engaged with it while in the feed or elsewhere) - not sending these interactions back is more of a privacy speed bump than a privacy protection.[^public]
[...]

PR #8672, opened 2025-07-18
Enable show less / more buttons for third party feeds

kindgracekind:
Addresses #7285 by setting `feedfeedBack.enabled` when a feed generator has `acceptsInteractions: true`. Adds a `enabledInteractions` array to filter out passive actions from third-party feeds, since these will require more user permissions.

Code details:
* Updated `useFeedFeedback()` to accept a `FeedSourceInfo` object (needed to get `acceptsInteractions` and `proxyDid` for feed)
* Updated `PostSource` to attach a full `FeedSourceInfo` object instead of a `FeedDescriptor`

Questions / blockers:
* Is there a way to access `FeedSourceInfo` from the video feed? If so, the code can be simplified quite a bit.
* I added a new piece of text ("Tell feed provider:"). Should I add translations for this?
* It looks like `acceptsInteractions` isn't being passed through the API. Here's an example of a record with it set: [generator record](https://pdsls.dev/at://did:plc:p572wxnsuoogcrhlfrlizlrb/app.bsky.feed.generator/undiscover). Is there a place I can make this change in the atproto codebase?
* When I try to send feedback to a third-party generator DID, I get a 400 (Bad Request). Does this also require an API change?

Thanks, and please let me know if there's anything I've missed!


src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -490,13 +491,16 @@ let PostMenuItems = ({
...
         {isDiscoverDebugUser && (
-          <Menu.Item
-            testID="postDropdownReportMisclassificationBtn"
-            label={_(msg`Assign topic for algo`)}
-            onPress={onReportMisclassification}>
-            <Menu.ItemText>{_(msg`Assign topic for algo`)}</Menu.ItemText>
-            <Menu.ItemIcon icon={AtomIcon} position="right" />
-          </Menu.Item>
+          <>
+            <Menu.Divider />
```
kindgracekind, line 494: Added a divider since "Tell feed provider" doesn't apply to this.

src/state/feed-feedback.tsx
```diff
@@ -184,6 +215,50 @@ function isDiscoverFeed(feed?: FeedDescriptor) {
   return !!feed && FEEDBACK_FEEDS.includes(feed)
 }
 
+function buildFeedInfo(feed: FeedSourceInfo | FeedDescriptor): FeedInfo | null {
```
kindgracekind: Coalesce FeedSourceInfo and FeedDescriptor into a FeedInfo object by adding some defaults. I added this because I couldn't find a way to access FeedSourceInfo from the videos feed. If that's possible, this can be removed in favor of using FeedSourceInfo for everything!

kabs9000:
I would very much like this to happen, please

haileyok:
I've added this to my list of stuff to look at, thank you for doing this! Definitely interested in getting this in.

kindgracekind:
Great, thank you! Let me know about the 2 potential API changes (`acceptsInteractions` and proxying), if those are changes I can make I can do that in parallel to make sure everything's ready to this to merge!

haileyok:
@github-actions ota

haileyok:
@kindgracekind do you happen to have a feed you set up?

kindgracekind:
Yes, here! https://bsky.app/profile/gracekind.net/feed/undiscover

haileyok:
> It looks like acceptsInteractions isn't being passed through the API. Here's an example of a record with it set: [generator record](https://pdsls.dev/at://did:plc:p572wxnsuoogcrhlfrlizlrb/app.bsky.feed.generator/undiscover). Is there a place I can make this change in the atproto codebase?

I'll go ahead and PR this now.

src/components/PostControls/PostMenu/PostMenuItems.tsx
haileyok: >  Added a divider since "Tell feed provider" doesn't apply to this.

Yep, makes sense.

mary-ext:
IIRC there was a past discussion over how feeds would intend to use the data provided from the feed interaction endpoint.

I think it would be good if feeds were to have a standardized field for privacy policy (and for the app to possibly warn? you get a surprising amount of data from this)

kindgracekind:
There’s a field for this in the describeFieldGenerator response: https://docs.bsky.app/docs/api/app-bsky-feed-describe-feed-generator

I didn’t add a UI element to view it because I didn’t think it was necessary for show more/less, but I can add it if the team decides it is necessary!

haileyok:
Yeah, the separation of the show more/less here is a good first step for us to start allowing this, then dig more at what we want to do with the passive data.

haileyok:
I've went ahead and created that PR here: https://github.com/bluesky-social/atproto/pull/4071

src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -470,6 +470,7 @@ let PostMenuItems = ({
           <>
             <Menu.Divider />
             <Menu.Group>
+              <Menu.LabelText>{_(msg`Tell feed provider:`)}</Menu.LabelText>
```
haileyok: We might want to change this copy, but it's fine for now. @estrattonbailey re: his opinion on this, will also ask internally

src/lib/constants.ts
```diff
@@ -91,6 +91,24 @@ export const STAGING_FEEDS = [
 
 export const FEEDBACK_FEEDS = [...PROD_FEEDS, ...STAGING_FEEDS]
 
+export const PASSIVE_INTERACTIONS = [
+  'app.bsky.feed.defs#clickthroughItem',
+  'app.bsky.feed.defs#clickthroughAuthor',
+  'app.bsky.feed.defs#clickthroughReposter',
+  'app.bsky.feed.defs#clickthroughEmbed',
+  'app.bsky.feed.defs#interactionSeen',
+]
```
haileyok: Might be nice for these to be typed, could prove useful

```suggestion
export const PASSIVE_FEEDBACK_INTERACTIONS = [
  'app.bsky.feed.defs#clickthroughItem',
  'app.bsky.feed.defs#clickthroughAuthor',
  'app.bsky.feed.defs#clickthroughReposter',
  'app.bsky.feed.defs#clickthroughEmbed',
  'app.bsky.feed.defs#interactionSeen',
] as const
export type PassiveFeedbackInteraction = (typeof PASSIVE_FEEDBACK_INTERACTIONS)[number]
```

src/lib/constants.ts
```diff
@@ -91,6 +91,24 @@ export const STAGING_FEEDS = [
...
+
+export const DIRECT_INTERACTIONS = [
+  'app.bsky.feed.defs#requestLess',
+  'app.bsky.feed.defs#requestMore',
+]
+
+export const ALL_INTERACTIONS = [
+  ...PASSIVE_INTERACTIONS,
+  ...DIRECT_INTERACTIONS,
+]
```
haileyok: Same on these

```suggestion
export const DIRECT_FEEDBACK_INTERACTIONS = [
  'app.bsky.feed.defs#requestLess',
  'app.bsky.feed.defs#requestMore',
] as const
export type DirectFeedbackInteraction = (typeof DIRECT_FEEDBACK_INTERACTIONS)[number]

export const FEEDBACK_INTERACTIONS = [
  ...PASSIVE_INTERACTIONS,
  ...DIRECT_INTERACTIONS,
] as const
export type FeedbackInteraction = (typeof FEEDBACK_INTERACTIONS)[number]
```

src/state/feed-feedback.tsx
```diff
@@ -27,21 +33,40 @@ export type StateContext = {
...
+  feed: FeedSourceInfo | FeedDescriptor | undefined,
   hasSession: boolean,
 ) {
   const agent = useAgent()
-  const enabled = isDiscoverFeed(feed) && hasSession
+
+  const feedInfo = feed ? buildFeedInfo(feed) : null
+  const enabled = !!feedInfo && feedInfo.acceptsInteractions && hasSession
+
+  const enabledInteractions = useMemo(() => {
```
haileyok: Can skip using `useMemo` now that we are using react compiler!

haileyok (commented):
Did a quick pass over this, mostly all looks good. Going to do another pass in a bit to dig more at the `feedSourceInfo` question, see if we can't clean that up a bit.

mackuba:
> There’s a field for this in the describeFieldGenerator response: https://docs.bsky.app/docs/api/app-bsky-feed-describe-feed-generator

I don't think this is used anywhere at the moment? Looks like it was there since the beginning when feed API was added in 2023.

haileyok:
> I don't think this is used anywhere at the moment? Looks like it was there since the beginning when feed API was added in 2023.

By this do you mean that just no one is using it or that it doesn't come down in responses? We definitely don't display it in the app (so maybe we also don't hydrate it) but could particularly once we start letting people collect the passive feedback.

mackuba:
Yeah, I mean that it doesn't seem to be used anywhere in the app (and likely not used by anyone anywhere for anything), so something that could be possibly changed/replaced if needed because it's more of a stub at this point I think

DGaffney:
hell yeah dogg - @haileyok do you or @kindgracekind  have a PDSLS link I can see for where acceptsInteractions needs to get set / how would we... get this feedback? Happy to enable this Graze wide asap

haileyok:
@DGaffney [this one ](https://pdsls.dev/at://did:plc:p572wxnsuoogcrhlfrlizlrb/app.bsky.feed.generator/undiscover) should do! i'll try and get the appview change deployed soon as well so you can start seeing it come back in the appview response.

src/lib/constants.ts
kindgracekind: Added!

src/state/feed-feedback.tsx
kindgracekind: Updated 👍

src/lib/constants.ts
kindgracekind: Added

kindgracekind:
Btw I've confirmed that the 400 from `app.bsky.feed.sendInteractions` was just me not handling the request correctly from my API - I'll mark that as resolved in the description!

kindgracekind:
Added a query to pull feed source info for video feeds! This request shouldn't block anything other than feed feedback on that page. This allowed me to simplify the feed feedback code, since the same feed source obj is now being used everywhere.

Are there any more blockers other than translations and the `acceptsInteractions` PR?

haileyok:
@kindgracekind appview change should be deployed now, so this should be testable without any tweaking/dev environment.

i'll give it some more looks now that we can do that soon.

kindgracekind:
Just tested locally and it's working for me! Although, I haven't been able to test native video feeds because I can't get the mobile build to work :/

src/components/PostControls/PostMenu/PostMenuItems.tsx
surfdude29:
```suggestion
              <Menu.LabelText>
                {_(
                  msg({
                    message: `Tell feed provider:`,
                    comment:
                      `Visible label for group of show more/show less options in post context menu in certain feeds which enable feedback`,
                  }),
                )}
              </Menu.LabelText>
```
A comment for translators might be helpful here? Although, maybe it's not too necessary as the string will be visible in the UI 🤔

haileyok:
> I added a new piece of text ("Tell feed provider:"). Should I add translations for this?

No need, we'll let translators update that as they please.

src/components/PostControls/PostMenu/PostMenuItems.tsx
surfdude29:
```suggestion
              <Menu.LabelText>
                {_(
                  msg({
                    message: `Tell feed provider:`,
                    comment:
                      `Visible label for group of show more/show less options in post context menu in certain feeds which enable feedback`,
                  }),
                )}
              </Menu.LabelText>
```
A comment for translators might be helpful here? Although, maybe it's not too necessary as the string will be visible in the UI 🤔

haileyok:
@github-actions ota

src/state/feed-feedback.tsx
```diff
@@ -28,22 +39,33 @@ export type StateContext = {
...
   const agent = useAgent()
-  const enabled = isDiscoverFeed(feed) && hasSession
+
+  const feed =
+    !!feedSourceInfo && isFeedSourceFeedInfo(feedSourceInfo)
+      ? feedSourceInfo
+      : undefined
+
+  const isDiscover = isDiscoverFeed(feed?.feedDescriptor)
+  const acceptsInteractions = isDiscover || (feed?.acceptsInteractions ?? false)
```
haileyok:
```suggestion
  const acceptsInteractions = Boolean(isDiscover || feed?.acceptsInteractions)
```

src/state/feed-feedback.tsx
```diff
@@ -28,22 +39,33 @@ export type StateContext = {
...
 export function useFeedFeedback(
-  feed: FeedDescriptor | undefined,
+  feedSourceInfo: FeedSourceInfo | undefined,
   hasSession: boolean,
 ) {
   const agent = useAgent()
-  const enabled = isDiscoverFeed(feed) && hasSession
+
+  const feed =
+    !!feedSourceInfo && isFeedSourceFeedInfo(feedSourceInfo)
```
haileyok, line 91:
```suggestion
    Boolean(feedSourceInfo) && isFeedSourceFeedInfo(feedSourceInfo)
```

src/state/feed-feedback.tsx
```diff
@@ -28,22 +39,33 @@ export type StateContext = {
...
+      ? feedSourceInfo
+      : undefined
+
+  const isDiscover = isDiscoverFeed(feed?.feedDescriptor)
+  const acceptsInteractions = isDiscover || (feed?.acceptsInteractions ?? false)
+  const enabled = !!feed && acceptsInteractions && hasSession
```
haileyok:
```suggestion
  const enabled = Boolean(feed) && acceptsInteractions && hasSession
```

src/state/feed-feedback.tsx
```diff
@@ -66,19 +88,26 @@ export function useFeedFeedback(
...
+        interaction.event &&
+        isFeedbackInteraction(interaction.event) &&
+        enabledInteractions.includes(interaction.event),
+    )
+
+    if (interactionsToSend.length === 0) {
+      return
     }
 
+    const proxyDid = feed?.view?.did
```
haileyok: Should we move this up and only enable if the did is set? I think it will always be set, but little weird using it when it's optional like this

src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -466,6 +466,7 @@ let PostMenuItems = ({
...
```
haileyok: Might suggest only showing this message if the feed is _not_ discover (no strong preference, but keeping things the same for normies...)

haileyok: @estrattonbailey @mozzius suggestions on copy for this?

kindgracekind: Made this change!

kindgracekind:
No problem, thanks for reviewing and testing!

src/state/feed-feedback.tsx
kindgracekind: I reverted this because it was making typescript unhappy (it thought feedSourceInfo would still be undefined in the second clause). Let me know if there's some way around this, or if I should update the other instances for consistency. Otherwise I'll leave it as-is!

kindgracekind:
Made the requested changes (with one exception, see the comment). Let me know if there's anything more you need from me before merging!

DGaffney:
Thank you for making this happen! Very excited to see how this helps feed operators.

src/state/feed-feedback.tsx
haileyok: oh weird, no worries. it's extremely nitpicky anyway :p. also could have been a linter error

kindgracekind:
This is just awaiting approval on copy, right? Would it be possible to merge as-is and iterate on copy later?

haileyok:
I think we actually landed on something anyway. What if we just go ahead and don't show the text in the dropdown/modal at all, but in the toast that shows up after sending feedback change the copy to "Feedback sent to feed operator"?

estrattonbailey:
Ope, sorry, I caused the conflicts. Those files are now deleted, you can just remove them from this PR.

src/lib/constants.ts
```diff
@@ -92,6 +92,40 @@ export const STAGING_FEEDS = [
 
 export const FEEDBACK_FEEDS = [...PROD_FEEDS, ...STAGING_FEEDS]
 
+export const PASSIVE_FEEDBACK_INTERACTIONS = [
```
estrattonbailey: Loving how organized this is!

I realize you're just following our convention, and thanks for that, but I think we should just put these in `feed-feedback.tsx` though. Can move `FEEDBACK_FEEDS` too, idk why that's in here.

src/components/PostControls/PostMenu/PostMenuItems.tsx
estrattonbailey: Hmmm yeah not sure on this one. Starts to look ugly if it gets too long, may be tough in other langs so maybe is what it is.

Some quick alts?
- `Notify feed provider`
- `Instruct feed provider`

estrattonbailey: Ah ok I missed some convos yesterday, I see Hailey posted about dropping this text entirely, ignore me!

src/lib/constants.ts
haileyok: moved

haileyok:
Made those changes. If CI passes will merge. Thanks @kindgracekind!!!

DGaffney:
image

PR #8679, opened 2025-07-20
Maybe fix logouts

gaearon:
This is written on GH, I haven't actually tried it.

See @mackuba's theory here: https://bsky.app/profile/mackuba.eu/post/3luecguaugc2h

The thought behind this PR is that we're already doing something super awkward, we might as well do it consistently

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -295,7 +295,8 @@ class BskyAppAgent extends BskyAgent {
           _lastNonEmptySession = val
         }
       },
-      configurable: true,  
+      configurable: true,
+      enumerable: true
     })
 
     this.persistSessionHandler = event => {
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -283,13 +283,35 @@ class BskyAppAgent extends BskyAgent {
 
     // Now the agent is ready.
     const account = agentToSessionAccountOrThrow(this)
-    let lastSession = this.sessionManager.session
+
+    // HACK (part 1/2): Hijack all assignments of `session`.
+    let _session = this.sessionManager.session
+    let _lastNonEmptySession = this.sessionManager.session
+    Object.defineProperty(this.sessionManager, 'session', {
+      get() { return _session; },
+      set(val) {
+        _session = val;
+        if (val) {
+          _lastNonEmptySession = val
+        }
+      },
+      configurable: true,  
+    })
+
     this.persistSessionHandler = event => {
-      if (this.sessionManager.session) {
-        lastSession = this.sessionManager.session
-      } else if (event === 'network-error') {
-        // Put it back, we'll try again later.
-        this.sessionManager.session = lastSession
+      if (
+        !this.sessionManager.session &&
+        _lastNonEmptySession
+      ) {
+        // HACK (part 2/2): Patch the last known one onto the object.
+        // We think this is OK to do because it was *just* set to undefined up the callstack.
+        // This is a fragile assumption though and should really be fixed upstream.
+        if (event === 'network-error') {
+          this.sessionManager.session = _lastNonEmptySession
+        } else if (event === 'expired') {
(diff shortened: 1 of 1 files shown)


mackuba:
Ngl, this looks cursed

gaearon:
cursed code *should* look cursed

haileyok:
(have not yet tested, but god i cannot wait to be able to rip out this session manager crap one day...)

gaearon:
I wouldn’t take this PR extremely seriously but I also wouldn’t exclude the possibility that it actually fixes the issue.

A more proper fix would be upstream to rip out the `.session = undefined` assignment that happens on network error. Then we won’t need to try to patch it back here. And if we don’t patch it back here, we don’t have a risk of patching a stale one and bringing it back to life (which presumably may cause problems?)

Note re: bringing a stale one back to life, I don’t actually have a 100% confident scenario in mind. Describing that might need @mackuba’s help based on the logs. It’s not obvious to me why patching a stale one is a problem for other tabs. I haven’t thought this through. It would be good to write it down explicitly.

And then the question is how much to fix at once. Eg it would be nice to first validate the exact theory. For example by getting a reliable repro case. Then this lets us verify this specific issue as the root cause of the problem. Which would also let us compare different fixes and how extensive they are.

The most conservative fix is probably just to make sure this closed-over “last working session” is always up-to-date. This means either “catching” the places where other code patches `.session` (like attempted in this PR), thus deepening the hack (but keeping it isolated from other logic). Or on the contrary making this mechanism a bit more explicit, like with some custom field that has to be updated whenever you manually patch session. That seems like we do it only in one location anyway (when data from another tab is picked up). 

So the options are to fix it upstream (but maybe break other things), make the hack more invasive without changing callsites (this PR), or change the callsites and add some coordination mechanism for this hack.

gaearon:
Basically the problem is that listening for persist handler and keeping track of “last non-empty session” there is not enough because this code won’t fire if `.session` was edited but “persist” was not called. And that’s exactly how we’re patching `.session` ourselves in the other file! So this code could close over a stale session. And this “bring it back” after the next network error.

gaearon:
One possible fix in that case could just be to add a persist event after we patch the session ourselves. Since that’s the pattern in the atproto client already. My only concern about that is that we’re listening to this handler ourselves so we need to make sure not to enter some kind of recursion. Or try to broadcast the thing we just received. So we need to closely audit what happens during that persist. If it’s fine then maybe that’s most elegant.

haileyok:
I'm going to let @estrattonbailey be the one to look at this in full (believe it is already on his plate), but it's worth noting that I've been dry running OAuth session management and think we're going to be ripping out a lot of this stuff here soon once we do get around to landing that. (I know we have been saying that for some time, but the reality is actually quite close now). That said, I think exploring this in general is still good, and appreciate both you and especially @mackuba for the investigation :pray:

estrattonbailey:
Oh hey yeah missed the convo here. Yeah I spent a good chunk of yesterday looking at this again and writing up a detailed doc in Linear to try to draw a circle around all these symptoms.

I think the base hypothesis here that we accidentally use a stale session is probably correct and we need to find a way to reliably get the freshest session at any given time. @mackuba may have [pointed out something here](https://github.com/bluesky-social/atproto/commit/1d6bc979a6d606f0d3b8b011f46ddc3d7bce5762#r162560638) that's been one of my theories: that browsers may throttle events/async handlers in stale tabs, which could cause a given tab to fall behind. If that's the case, then we need a reliable source of truth for what constitutes the latest session once that tab comes back online, which I'm thinking atm means a synchronous read to `localStorage`. May try something like that.

gaearon:
I'm motivated to actively work on this but I need someone to be on point for fast reviews.

This particular PR is superseded by https://github.com/bluesky-social/atproto/pull/4238. If we can get this reviewed and merged, I can clean up this part of the social app and then do the "write to storage synchronously" fix separately.

PR #8689, opened 2025-07-22
feat(ui): remove like count from trending videos

caidanw:
Summary
---

Hide the like count on trending videos. Implemented as a hardcoded feature flag so that we can toggle it back on if desired.

Based on feedback and discussion in the linear ticket, I've included some minor adjustments to hide the gradient (where the like count was) and update the profile picture to be slightly bigger and have a small dropshadow.

Linear: https://linear.app/blueskyweb/issue/APP-965/remove-like-count-from-trending-videos

Copilot Generated
---

This pull request introduces a minor change to the `CompactVideoPostCard` component in the `src/components/VideoPostCard.tsx` file. The change involves adding a `showLikeCount` flag to control whether the like count is displayed.

### Changes to `CompactVideoPostCard`:

* Added a new constant, `showLikeCount`, initialized to `false`, to determine whether the like count should be shown.
* Updated the conditional rendering logic for the like count to include the `showLikeCount` flag, ensuring the like count is only displayed when both `likeCount > 0` and `showLikeCount` are true.

### DIFF
--- a/src/components/VideoPostCard.tsx
+++ b/src/components/VideoPostCard.tsx
@@ -390,6 +390,7 @@ export function CompactVideoPostCard({
   if (!AppBskyEmbedVideo.isView(embed)) return null
 
   const likeCount = post?.likeCount ?? 0
+  const showLikeCount = false
   const {thumbnail} = embed
   const black = getBlackColor(t)
 
@@ -506,7 +507,7 @@ export function CompactVideoPostCard({
 
                 <View
                   style={[a.relative, a.z_10, a.p_sm, a.flex_row, a.gap_md]}>
-                  {likeCount > 0 && (
+                  {likeCount > 0 && showLikeCount && (
                     <View style={[a.flex_row, a.align_center, a.gap_xs]}>
                       <Heart size="sm" fill="white" />
                       <Text style={[a.text_sm, a.font_bold, {color: 'white'}]}>
--- a/src/components/VideoPostCard.tsx
+++ b/src/components/VideoPostCard.tsx
@@ -476,7 +476,7 @@ export function CompactVideoPostCard({
             />
             <MediaInsetBorder />
 
-            <View style={[a.absolute, a.inset_0]}>
+            <View style={[a.absolute, a.inset_0, a.shadow_sm]}>
               <View style={[a.absolute, a.inset_0, a.p_sm, {bottom: 'auto'}]}>
                 <View
                   style={[a.relative, a.rounded_full, {width: 24, height: 24}]}>
--- a/src/components/VideoPostCard.tsx
+++ b/src/components/VideoPostCard.tsx
@@ -479,10 +479,10 @@ export function CompactVideoPostCard({
             <View style={[a.absolute, a.inset_0]}>
               <View style={[a.absolute, a.inset_0, a.p_sm, {bottom: 'auto'}]}>
                 <View
-                  style={[a.relative, a.rounded_full, {width: 20, height: 20}]}>
+                  style={[a.relative, a.rounded_full, {width: 24, height: 24}]}>
                   <UserAvatar
                     type="user"
-                    size={20}
+                    size={24}
                     avatar={post.author.avatar}
                   />
                   <MediaInsetBorder />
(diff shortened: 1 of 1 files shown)


haileyok (approved):

estrattonbailey:
@haileyok there was some discussion on this one https://linear.app/blueskyweb/issue/APP-965/remove-like-count-from-trending-videos

src/components/VideoPostCard.tsx
```diff
@@ -475,47 +476,51 @@ export function CompactVideoPostCard({
             />
             <MediaInsetBorder />
 
-            <View style={[a.absolute, a.inset_0]}>
+            <View style={[a.absolute, a.inset_0, a.shadow_sm]}>
               <View style={[a.absolute, a.inset_0, a.p_sm, {bottom: 'auto'}]}>
                 <View
-                  style={[a.relative, a.rounded_full, {width: 20, height: 20}]}>
+                  style={[a.relative, a.rounded_full, {width: 24, height: 24}]}>
```
jimmylee, line 482: Thought: Is there a place where these values should live? Do they also follow the design?

haileyok: we do have e.g. `a.w_full`, `a.p_xl`, etc. but for things like this we do generally use numbers like this when needed. also doing in-line styles is fine (and we should use them unless we update the atoms. kind of a small note, but we avoid using e.g. `Stylesheet.create()` these days).

that said, i feel like avatar sizes are something we should start being very consistent on throughout the app, and something like `a.avatar_md` `a.avatar_lg` etc would be good. let's see what @estrattonbailey thinks though as that is mostly his space

haileyok: (also worth seeing with darrin re: if we have any ideas on a set of avatar sizes to begin using)

estrattonbailey: If Darrin has preferences and ideas for fixed avatar sizes, cool with me. Would be great to get to that point, but until then I'm not personally concerned with numbers like this.

src/components/VideoPostCard.tsx
```diff
@@ -475,47 +476,51 @@ export function CompactVideoPostCard({
             />
             <MediaInsetBorder />
 
-            <View style={[a.absolute, a.inset_0]}>
+            <View style={[a.absolute, a.inset_0, a.shadow_sm]}>
```
estrattonbailey: The base shadow atoms don't have colors yet, anything color related comes from the `Theme` object from `useTheme` e.g. `const {t} = useTheme()` and `t.atoms.shadow_sm`. So gotta use the shadow atom from the theme itself.

Our system is a little weird, apologies 😅 you can think of `t.atoms` as being "color themed atoms". The base `a` atoms are agnostic of theme (and therefore colors) entirely.

src/components/VideoPostCard.tsx
```diff
@@ -390,6 +390,7 @@ export function CompactVideoPostCard({
   if (!AppBskyEmbedVideo.isView(embed)) return null
 
   const likeCount = post?.likeCount ?? 0
+  const showLikeCount = false
```
estrattonbailey, line 393: I'm cool with this. Clear spot to extend later 👍

src/components/VideoPostCard.tsx
caidanw: Great question, I have no idea on where these should live, maybe @haileyok has an idea?

They follow the design, was advised directly by Darrin L. to update these values to match the his new design.

src/components/VideoPostCard.tsx
caidanw: Ah ok thanks, yeah have not used this specific styling system before.

With the `t.atoms.shadow_sm` using the colored theme, does this mean that in dark mode the shadow will be white or will it remain black?

estrattonbailey: Yeah we investigated a few options and landed on this in early Jan 2024. It roughly follows Tailwind naming conventions.

An yeah, `t.atoms` always match the active color theme.

caidanw: Cool thanks for the context, I've updated to use the theme atoms. Is the new change what you expected?

estrattonbailey (approved):
Nice one, thanks!

jimmylee:
Thanks @caidanw

PR #8691, opened 2025-07-22
[APP-1281] Add margin between messages in the composer thread 

BinaryFiddler:
Change log:
- Adding 8px of margin to the last in active thread in the composer which is always the last one at the bottom of the compose window

https://github.com/user-attachments/assets/2c2e9f01-b3c8-4b37-b4ad-ad4363dded1f

https://linear.app/blueskyweb/issue/APP-1281/post-composer-thread-ui-issues

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -699,6 +699,7 @@ export const ComposePost = ({
                   dispatch={composerDispatch}
                   textInput={post.id === activePost.id ? textInput : null}
                   isFirstPost={index === 0}
+                  isLastPost={index === thread.posts.length - 1}
                   isPartOfThread={thread.posts.length > 1}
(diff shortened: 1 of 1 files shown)


jimmylee (commented):
LGTM, great work.

It looks like you followed convention and added another stylistic case. Thinking aloud: I don’t love this component overall. It feels like there are a bunch of conditional states that have to be tested optically, but I’m glad we were able to get into the codebase quickly.

src/view/com/composer/Composer.tsx
```diff
@@ -830,6 +833,8 @@ let ComposerPost = React.memo(function ComposerPost({
     <View
       style={[
         a.mx_lg,
+        a.mb_sm,
+        !isActive && isLastPost && a.mb_lg,
```
haileyok, line 837: perfect

haileyok (approved):

PR #8692, opened 2025-07-22
[APP-1083] bug fix: videos not accurately autoplaying on web

APiligrim:
This PR addresses [this ticket ](https://linear.app/blueskyweb/issue/APP-1083/videos-not-autoplaying-on-web ) and updates the logic of activating a video to autoplay when it is most closest to the center of the viewport. To replicate the issue, please visit https://bsky.app/profile/apolloarmando.bsky.social/post/3lj4avp6ftk2d to see the production view. 

**Video of the problem with old implementation:** 
This original implementation used an "ideal position" concept but at times the video closest to the viewport center wasn't reliably getting activated when the most visible video was on the screen.

https://github.com/user-attachments/assets/180f418b-5b32-49d7-ad3f-f79edb3564bf

**Video of the new implementation:**
This updated approach uses a safe zone that represents where videos should be active because they're most visible to the user. It covers 35% of the viewport height, centered around the viewport center.

https://github.com/user-attachments/assets/e3a14beb-7383-4d49-a434-5a8bef7541c7

https://github.com/user-attachments/assets/25e1fc49-2dba-49a0-ab03-82576fd003b6

**Checklist**
- [x]  If a video is in the safe zone, it should be active
- [x] If multiple videos are in the safe zone, the one closest to the viewport center gets priority
- [x] Fallback: If no videos are in the safe zone, use the closest-to-center logic

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/index.web.tsx
+++ b/src/components/Post/Embed/VideoEmbed/index.web.tsx
@@ -36,22 +36,20 @@ export function VideoEmbed({
   useEffect(() => {
     if (!ref.current) return
     if (isFullscreen && !isFirefox) return
-
     const observer = new IntersectionObserver(
       entries => {
         const entry = entries[0]
         if (!entry) return
         setOnScreen(entry.isIntersecting)
+        sendPosition(
+          entry.boundingClientRect.y + entry.boundingClientRect.height / 2,
+        )
       },
       {threshold: 0.5},
     )
-
     observer.observe(ref.current)
-
-    return () => {
-      observer.disconnect()
-    }
-  }, [isFullscreen])
+    return () => observer.disconnect()
+  }, [sendPosition, isFullscreen])
 
   const [key, setKey] = useState(0)
   const renderError = useCallback(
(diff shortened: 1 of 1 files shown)


jimmylee (commented):
LGTM from my side, maybe someone on Bluesky may have some comments about convention, I like that we fixed the bug here without changing too much.

src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -133,24 +135,35 @@ function ViewportObserver({
...
+      const entry = entries[0]
+      if (!entry) return
+      setNearScreen(entry.isIntersecting)
+      updatePosition()
+    })
+
     observer.observe(ref.current)
-    return () => observer.disconnect()
+
+    window.addEventListener('scroll', updatePosition, {passive: true})
```
estrattonbailey: Let's just be sure of the performance implications of adding a scroll listener in addition to the existing observers. Slightly concerned about the additional `getBoundingClientRect` call, since observers already do that, but might not be a huge deal here.

APiligrim:
@estrattonbailey Thank you for your comment and the note about optimization!

In the original implementation the `IntersectionObserver` shows a bit of delays in tracking if the video's position is in the viewport when the user scrolls back and forth as seen in this video https://intdevstucom.slack.com/archives/C096H7JRYM9/p1753144371449629?thread_ts=1753142971.608749&cid=C096H7JRYM9. The main reason for the delay is probably due to the number of thresholds `Array.from({length: 101}, (_, i) => i / 100)` resulting in up to 101 potential position recalculations as a video scrolls through the viewport and positions not updating in time.

The `getBoundingClientRect` and scroll event listener were added to give real-time position updates of the video in the viewport during scrolling. But after your comment about optimization, I removed the `getBoundingClientRect`  and added a throttled scroll listener so that its only activating it during active scrolling. Also since the scroll event listener provides real time  updates when its active, I also updated the `Array.from({length: 101}, (_, i) => i / 100)` to only show a few options `{threshold: [0, 0.5, 1.0]}`  to detect basic visibility states if the video is entering, in the center, or fully visible in the viewport.

APiligrim:
I updated my implementation solution to make it more easy to read and only use the intersection observer 

The original implementation had a viewport had an ideal zone but the video was not getting activated in that line , videos 
https://github.com/user-attachments/assets/1c4063e4-0566-4acb-a389-0f69d3e68191 

New implementation is to use safe zones:
- the green safe zone is where a video should be active at all times since its the most visible on the screen
- if that active video starts moving away from that safe zone and a new video is closer to or is in the safe zone, then it would make the new video active

https://github.com/user-attachments/assets/b3d2f324-b86c-4d95-a8f6-bfcb4a255a90

src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -138,22 +138,53 @@ function ViewportObserver({
...
-          entry.boundingClientRect.y + entry.boundingClientRect.height / 2
-        sendPosition(position)
+        lastObserverEntry = entry
         setNearScreen(entry.isIntersecting)
+        const rect = entry.boundingClientRect
+        const position = rect.y + rect.height / 2
+        sendPosition(position)
       },
-      {threshold: Array.from({length: 101}, (_, i) => i / 100)},
+      {threshold: [0, 0.1, 0.25, 0.5, 0.75, 1.0]},
```
jimmylee, line 169: Nice

jimmylee (approved):
LGTM, needs @mozzius to confirm

haileyok:
@github-actions ota

haileyok:
(don't mind me, just testing something)

haileyok:
@github-actions ota

haileyok (approved):
From my eye this looks good, though I don't know much about this code in general and would like to wait for @mozzius to take a look (he gets back on Monday)

haileyok:
@github-actions ota

mozzius:
Ok I'm here, let's take a look 👀

jimmylee:
@mozzius any thoughts about this? Safe to land?

Issue #8608, opened 2025-07-05
Cannot search anything without logging in, getting NetworkError when attempting to fetch resource


IgorArnaut:
### Steps to Reproduce

1. Go to Bluesky website
2. Do not log in
3. Search something, expecting to get results
4. Get `Error: Error: NetworkError when attempting to fetch resource`

### Attachments

<img width="1876" height="931" alt="Image" src="https://github.com/user-attachments/assets/25919122-319a-491c-9472-60896006163d" />

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

Web

### Additional Information

I cannot search for anything on BlueSky without an account.

mozzius:
This has been intentionally disabled, but we need to update the app to reflect this

IArnaut2:
> This has been intentionally disabled, but we need to update the app to reflect this

This is a very counter-intuitive decision. You are blocking out guests from searching posts. What other site blocks out guests from searching the posts? Probably X.

mozzius:
Unfortunately unsociable people were abusing it. We'd like to provide it, but repeated abuse has made it impossible. Sorry :(

IArnaut2:
> Unfortunately unsociable people were abusing it. We'd like to provide it, but repeated abuse has made it impossible. Sorry :(

How do you even abuse a search function?

mozzius:
> How do you even abuse a search function?

By spamming hundreds of thousands of requests at it

PR #8649, opened 2025-07-15
Remove broken search screen in PWI

mozzius:
Fixes #8608 

We don't support logged-out search anymore, so let's replace the tab. I made the actual 2 post search tabs dependent on being logged in, then added a 3rd tab labelled "Posts" that just prompts the user to sign in.

Screenshot 2025-07-15 at 13 36 29

# Test plan

Make sure search screen works as expected, both logged in and logged out

PR #8697, opened 2025-07-22
[APP-1157] disable searches for logged out users, prompt them to sign in

BinaryFiddler:
Change log:
- Disable searches for top and latest results for logged out users
- Implemented a new search error component
Screenshot 2025-07-23 at 4 54 49 PM
Screenshot 2025-07-23 at 4 54 57 PM

Figma ref: https://www.figma.com/design/TqFguaWhkvAPnCth5u7cvE/Everything?node-id=3392-964&m=dev

Linear ticket: https://linear.app/blueskyweb/issue/APP-1157/logged-out-search-is-broken

### DIFF
--- a/src/components/SearchError.tsx
+++ b/src/components/SearchError.tsx
@@ -16,7 +16,6 @@ export function SearchError({
   const {gtMobile} = useBreakpoints()
   const pal = usePalette('default')
 
-  // NOTE (chenyu): asking Hannah about the new fill color for the SVG
   return (
     <Layout.Content>
       <View
--- a/src/components/SearchError.tsx
+++ b/src/components/SearchError.tsx
@@ -23,22 +23,19 @@ export function SearchError({
         style={[
           a.align_center,
           a.gap_4xl,
+          a.px_xl,
           {
-            paddingHorizontal: 10,
             paddingVertical: 150,
           },
         ]}>
         <TimesLarge_Stroke2_Corner0_Rounded width={32} fill={pal.colors.icon} />
         <View
           style={[
-            a.flex,
-            a.flex_col,
             a.align_center,
             {maxWidth: gtMobile ? 394 : 294},
             gtMobile ? a.gap_md : a.gap_sm,
           ]}>
-          <Text
-            style={[a.font_bold, a.text_lg, a.text_center, {lineHeight: 1.3}]}>
+          <Text style={[a.font_bold, a.text_lg, a.text_center, a.leading_snug]}>
             {title}
           </Text>
           {children}
--- a/src/components/SearchError.tsx
+++ b/src/components/SearchError.tsx
@@ -1,0 +1,49 @@
+import {View} from 'react-native'
+
+import {usePalette} from '#/lib/hooks/usePalette'
+import {atoms as a, useBreakpoints} from '#/alf'
+import * as Layout from '#/components/Layout'
+import {Text} from '#/components/Typography'
+import {TimesLarge_Stroke2_Corner0_Rounded} from './icons/Times'
+
+export function SearchError({
+  title,
+  children,
+}: {
+  title?: string
+  children?: React.ReactNode
+}) {
+  const {gtMobile} = useBreakpoints()
+  const pal = usePalette('default')
+
+  // NOTE (chenyu): asking Hannah about the new fill color for the SVG
+  return (
+    <Layout.Content>
+      <View
+        style={[
+          a.align_center,
+          a.gap_4xl,
+          {
+            paddingHorizontal: 10,
+            paddingVertical: 150,
+          },
+        ]}>
+        <TimesLarge_Stroke2_Corner0_Rounded width={32} fill={pal.colors.icon} />
+        <View
+          style={[
+            a.flex,
+            a.flex_col,
+            a.align_center,
+            {maxWidth: gtMobile ? 394 : 294},
+            gtMobile ? a.gap_md : a.gap_sm,
+          ]}>
+          <Text
+            style={[a.font_bold, a.text_lg, a.text_center, {lineHeight: 1.3}]}>
+            {title}
+          </Text>
+          {children}
+        </View>
+      </View>
+    </Layout.Content>
+  )
+}
--- a/src/screens/Search/SearchResults.tsx
+++ b/src/screens/Search/SearchResults.tsx
@@ -16,11 +16,11 @@ import {Pager} from '#/view/com/pager/Pager'
 import {TabBar} from '#/view/com/pager/TabBar'
 import {Post} from '#/view/com/post/Post'
 import {ProfileCardWithFollowBtn} from '#/view/com/profile/ProfileCard'
-import {TextLink} from '#/view/com/util/Link'
 import {List} from '#/view/com/util/List'
 import {atoms as a, useTheme, web} from '#/alf'
 import * as FeedCard from '#/components/FeedCard'
 import * as Layout from '#/components/Layout'
+import {InlineLinkText} from '#/components/Link'
 import {SearchError} from '#/components/SearchError'
 import {Text} from '#/components/Typography'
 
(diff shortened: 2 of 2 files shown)


src/screens/Search/SearchResults.tsx
```diff
@@ -161,6 +174,7 @@ let SearchScreenPostResults = ({
   const {_} = useLingui()
   const {currentAccount} = useSession()
   const [isPTR, setIsPTR] = useState(false)
+  const isLoggedin = currentAccount && currentAccount.did
```
haileyok: nit: we mostly use `Boolean(currentAccount?.did)` for these sorts of things

src/screens/Search/SearchResults.tsx
```diff
@@ -216,6 +230,51 @@ let SearchScreenPostResults = ({
...
+
+  const showSignIn = useCallback(() => {
+    closeAllActiveElements()
+    requestSwitchToAccount({requestedAccount: 'none'})
+  }, [requestSwitchToAccount, closeAllActiveElements])
+
+  const showCreateAccount = useCallback(() => {
+    closeAllActiveElements()
+    requestSwitchToAccount({requestedAccount: 'new'})
+  }, [requestSwitchToAccount, closeAllActiveElements])
```
haileyok: generally do not need to use `useCallback` anymore since we're now using react compiler. can just do `const showSignIn = () => {}`

we still use `useCallback` all over the place, but with new code generally feels good to stop using it

haileyok (commented):
two small nits, left a q in slack https://bluesky-builders.slack.com/archives/C096H7JRYM9/p1753212497012309?thread_ts=1753208786.683469&cid=C096H7JRYM9

surfdude29:
Fixes #8608 

Also, sorry to jump in here as I realise it might be out-of-scope for this PR, but like in #8649 I just wanted to flag up that the hashtag search pages are also broken in the PWI, and a similar approach would be great there too :)

IMG_3463

src/screens/Search/SearchResults.tsx
```diff
@@ -9,12 +9,15 @@ import {useActorSearch} from '#/state/queries/actor-search'
 import {usePopularFeedsSearch} from '#/state/queries/feed'
 import {useSearchPostsQuery} from '#/state/queries/search-posts'
 import {useSession} from '#/state/session'
+import {useLoggedOutViewControls} from '#/state/shell/logged-out'
```
jimmylee, line 13: Thought: Does the Bsky team care that these are alphabetized?

BinaryFiddler:
> Fixes #8608
> 
> Also, sorry to jump in here as I realise it might be out-of-scope for this PR, but like in #8649 I just wanted to flag up that the hashtag search pages are also broken in the PWI, and a similar approach would be great there too :)
> 
> IMG_3463

I'll either roll that into this one or create a separate linear ticket

src/screens/Search/SearchResults.tsx
```diff
@@ -216,6 +232,54 @@ let SearchScreenPostResults = ({
...
+      <SearchError
+        title={_(msg`Search is currently unavailable when logged out`)}>
+        <Text
+          style={[
+            a.text_md,
+            a.text_center,
+            t.atoms.text_contrast_medium,
+            {lineHeight: 1.3},
+          ]}>
+          <TextLink
```
BinaryFiddler: I'm actually unsure if this is the idiomatic way, please do let me know if there is a better approach.

src/components/SearchError.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // NOTE (chenyu): asking Hannah about the new fill color for the SVG
+  return (
+    <Layout.Content>
+      <View
+        style={[
+          a.align_center,
+          a.gap_4xl,
+          {
+            paddingHorizontal: 10,
```
jimmylee: I'm guessing these are just values that look nice and we nudged them around to feel it out.

src/components/SearchError.tsx
```diff
@@ -0,0 +1,47 @@
...
+            paddingVertical: 150,
+          },
+        ]}>
+        <Cross width={32} fill={'#D1D8E0'} />
+        <View
+          style={[
+            a.flex,
+            a.flex_col,
+            a.align_center,
+            {maxWidth: gtMobile ? 394 : 294},
```
jimmylee, line 34: Same comment about values here as I made above, how do we know `394` and `294` are values we want to work with? I'm sure they satisfy the requirement of fixing this issue/UI

cc: @estrattonbailey

jimmylee (commented):
LGTM, @BinaryFiddler, there just seems to be some magic values that I would want to consult @estrattonbailey on.

src/screens/Search/SearchResults.tsx
BinaryFiddler: I'll pretend they do if no one answers 😁

src/components/SearchError.tsx
BinaryFiddler: Yes, I post some questions about these values on Figma. I'd love to hear more about how we can approach this systematically.

src/components/SearchError.tsx
BinaryFiddler: Same above

src/components/SearchError.tsx
estrattonbailey: `a.px_md` is 12px, pry fine to use the design system value here

src/components/SearchError.tsx
```diff
@@ -0,0 +1,47 @@
...
+  // NOTE (chenyu): asking Hannah about the new fill color for the SVG
+  return (
+    <Layout.Content>
+      <View
+        style={[
+          a.align_center,
+          a.gap_4xl,
+          {
+            paddingHorizontal: 10,
+            paddingVertical: 150,
```
estrattonbailey, line 27: Might be nice to do this proportionally to screen height using `vh` units if we can

src/components/SearchError.tsx
estrattonbailey: Max width I typically consider vibes based. If Darrin is happy with it, I think whatever value here is fine 👍

src/components/SearchError.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <View
+          style={[
+            a.flex,
+            a.flex_col,
+            a.align_center,
+            {maxWidth: gtMobile ? 394 : 294},
+            gtMobile ? a.gap_md : a.gap_sm,
+          ]}>
+          <Text
+            style={[a.font_bold, a.text_lg, a.text_center, {lineHeight: 1.3}]}>
```
estrattonbailey: Can use `a.leading_snug` for lineHeight

src/components/SearchError.tsx
```diff
@@ -0,0 +1,47 @@
...
+          {
+            paddingHorizontal: 10,
+            paddingVertical: 150,
+          },
+        ]}>
+        <Cross width={32} fill={'#D1D8E0'} />
+        <View
+          style={[
+            a.flex,
+            a.flex_col,
```
estrattonbailey: Views actually default to `flex` and `flex-direction: column` in React Native, so you can drop these

src/components/icons/Cross.tsx
```diff
@@ -0,0 +1,5 @@
+import {createSinglePathSVG} from './TEMPLATE'
+
+export const Cross = createSinglePathSVG({
```
estrattonbailey: We have this icon already actually! Called `TimesLarge_Stroke2_Corner0_Rounded`. I'll link the icon file in Slack, has more info in there.

src/screens/Search/SearchResults.tsx
estrattonbailey: Nah. We have some import sorting, but haven't gone as far as alpha yet

src/screens/Search/SearchResults.tsx
estrattonbailey: Good question, the more up to date component is `InlineLinkText`, let's use that

estrattonbailey (commented):
Looking good!

src/screens/Search/SearchResults.tsx
surfdude29: Apologies for jumping in here again, would it make sense to wrap this in `Trans` tags? I think this would allow the `Text` here to be fully localized, e.g. similar to how it's done here:
https://github.com/bluesky-social/social-app/blob/6a7b8a279b6113def5d6330ef40801a5484a0b7c/src/view/com/posts/FeedShutdownMsg.tsx#L98-L109

src/components/SearchError.tsx
BinaryFiddler: Darrin mentioned since this is a temp solution, this should be fine

src/screens/Search/SearchResults.tsx
```diff
@@ -216,6 +232,54 @@ let SearchScreenPostResults = ({
...
+    requestSwitchToAccount({requestedAccount: 'none'})
+  }
+
+  const showCreateAccount = () => {
+    closeAllActiveElements()
+    requestSwitchToAccount({requestedAccount: 'new'})
+  }
+
+  const pal = usePalette('default')
+  const t = useTheme()
```
haileyok: nit: one thing that at least I try to always do (but i think we all do) is keep hooks and variables at the top of components when possible. its easy to sneak in a `return` somewhere above hooks when they are not placed up there and break rules of react

src/components/SearchError.tsx
```diff
@@ -0,0 +1,46 @@
...
+  title,
+  children,
+}: {
+  title?: string
+  children?: React.ReactNode
+}) {
+  const {gtMobile} = useBreakpoints()
+  const pal = usePalette('default')
+
+  // NOTE (chenyu): asking Hannah about the new fill color for the SVG
```
haileyok: just a quick q on if we want to hold off on merging before we get an answer to this, no pref either way personally.

haileyok (approved):
this is a significant improvement! good stuff, thanks for working through all the questions with us 🙏 

will hold off to merge on the nit and just verifying you want to merge w/o addressing that note

src/components/SearchError.tsx
BinaryFiddler: removing this since i switched to use `pal.colors.icon`

src/screens/Search/SearchResults.tsx
```diff
@@ -216,6 +234,51 @@ let SearchScreenPostResults = ({
...
+  if (!isLoggedin) {
+    return (
+      <SearchError
+        title={_(msg`Search is currently unavailable when logged out`)}>
+        <Text style={[a.text_md, a.text_center, a.leading_snug]}>
+          <Trans>
+            <InlineLinkText
+              style={[pal.link]}
+              label={_(msg`sign in`)}
+              to={'#'}
```
estrattonbailey, line 259: @BinaryFiddler @jimmylee @haileyok doesn't really matter here since we're sending the user to a new screen and therefore overriding the `#` that will get pushed to the URL, but FYI for cases like this where we want the link to behave like a button we have `createStaticClick` from `Link.tsx`. There are examples in the codebase ✌️

IArnaut2:
This is idiotic. You can't just restrict searching for logged out users.

PR #8699, opened 2025-07-22
Include app version in metrics

estrattonbailey:
`nativeApplicationVersion` from `expo-application` only works on native. Importing `package.json` should work on all platforms afaik.


haileyok (approved):
yep

PR #8701, opened 2025-07-22
[APP-1173] Clean up env

estrattonbailey:
Best reviewed commit-by-commit. Pretty self explanatory. The gist is:
- Renamed bundled variables to follow `EXPO_PUBLIC_*` convention
- All scripts now set env vars in `.env` where possible, or in the shell command where necessary (atm this is only Sentry)
- New var `EXPO_PUBLIC_RELEASE_VERSION` which is just the package.json version
  - Can't get this on REnder.com deploys, so there are fallbacks in `webpack.config.js` and `env/common.ts`
- `EXPO_PUBLIC_BUNDLE_IDENTIFIER` is now the full commit hash, and we just truncate when rendering in the About Settings
- Sentry env vars are all consistent now, and re-use existing values:
  - `SENTRY_DIST=EXPO_PUBLIC_BUNDLE_IDENTIFIER`
  - `SENTRY_RELEASE=EXPO_PUBLIC_RELEASE_VERSION`
- Fixed the missing bundle ID in `main.bsky.dev` deploys (which used to say `dev (prod)` which was funny
- Replaced `NODE_ENV` where sensible, since really this should match `EXPO_PUBLIC_ENV` (aliased to `env.ENV` now)

Should be good, but to test this before we consider merging, I'm going to run builds and check the `main.bsky.dev` deployment, etc.


src/env.ts
```diff
@@ -1,6 +1,70 @@
...
+/**
+ * Currently unused. Enable debug for a subset of the debug logs.
+ */
+export const LOG_DEBUG: string = process.env.EXPO_PUBLIC_LOG_DEBUG || ''
+
+/**
+ * The DID of the chat service to proxy to
+ */
+export const CHAT_PROXY_DID: Did =
+  process.env.EXPO_PUBLIC_CHAT_PROXY_DID || 'did:web:api.bsky.chat'
```
estrattonbailey: Adding this in the spirit of #8313, which I'd like to revisit as a follow-on to this work

src/logger/sentry/setup/index.ts
```diff
@@ -1,32 +1,16 @@
-/**
- * Importing these separately from `platform/detection` and `lib/app-info` to
- * avoid future conflicts and/or circular deps
- */
```
estrattonbailey, line 4: Shouldn't be an issue now

.env.example
```diff
@@ -1,8 +1,49 @@
...
+EXPO_PUBLIC_ENV=development
 
-BITDRIFT_API_KEY=
-SENTRY_AUTH_TOKEN=
-EXPO_PUBLIC_LOG_LEVEL=debug
-EXPO_PUBLIC_LOG_DEBUG=
+# This is the short commit hash that the current bundle was made from.
 EXPO_PUBLIC_BUNDLE_IDENTIFIER=
+
+# Should be formatted YYMMDD so that it increases for each build.
```
haileyok: i think this needs hour/minute too fyi, though its a CI thing and doesn't matter much in the example. might as well mention tho (if its just `yymmdd` you could have duplicate bundle idents)

.env.example
```diff
@@ -1,8 +1,49 @@
...
 EXPO_PUBLIC_BUNDLE_DATE=0
+
+# Optional app view proxy DID
+EXPO_PUBLIC_APPVIEW_PROXY_DID=did:web:api.cozy.bsky.dev
```
haileyok: will this be around forever? might not want to set this/set it to the default appview so that people who copy `.env.example` to `.env` don't bork things (unless we plan for cozy to remain online indefniitely?)

haileyok: nvm i see you rm'd it later. do you want to still leave it though as just an undefined env var for others to see/notice? e.g. zeppelin/deer

haileyok (approved):
all lgtm, just the couple questions. great cleanup though

src/env.ts
```diff
@@ -1,6 +1,71 @@
...
+/**
+ * The env the app is running in e.g. development, testflight, production
+ */
+export const ENV: string = process.env.EXPO_PUBLIC_ENV
+
+/**
+ * Indicates whether the app is running in TestFlight
+ */
+export const IS_TESTFLIGHT =
+  ENV === 'testflight' || ENV.startsWith('testflight')
```
haileyok: this should always be `testflight` even if it's the android version. the `testflight-android` is just for the github action, as we want to run a separate EAS profile.

https://github.com/bluesky-social/social-app/blob/main/eas.json#L63-L88

haileyok: (note the `      "distribution": "internal",` line, which is why we have this separate profile. no other differences)

estrattonbailey: Oh right! Forgot about this

estrattonbailey: Ok bopped this here, will be sure to backport https://github.com/bluesky-social/social-app/pull/8704/commits/bab44d8ac4a7636b320ff09ba5be28f79535ca7b

.github/workflows/bundle-deploy-eas-update.yml
```diff
@@ -101,25 +101,30 @@ jobs:
...
+          echo "EXPO_PUBLIC_BITDRIFT_API_KEY=${{ secrets.BITDRIFT_API_KEY }}" >> .env
           echo "$json" > google-services.json
 
-      - name: Setup Sentry vars for build-time injection
-        id: sentry
-        run: |
-          echo "SENTRY_DIST=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
-          echo "SENTRY_RELEASE=$(jq -r '.version' package.json)" >> $GITHUB_OUTPUT
-
+      # TODO do we need EXPO_PUBLIC_ENV here?
```
estrattonbailey: TODO

.github/workflows/build-and-push-bskyweb-aws.yaml
```diff
@@ -43,12 +43,11 @@ jobs:
           tags: |
             type=sha,enable=true,priority=100,prefix=,suffix=,format=long
 
-      - name: Set outputs
-        id: vars
+      - name: Setup env
```
estrattonbailey:
```suggestion
      - name: Env
```

estrattonbailey:
@github-actions ota

estrattonbailey:
Gonna merge this and do the remaining testing that requires this in `main` 👍

PR #8705, opened 2025-07-23
[APP-1291] prevent GestureActionView from hijacking swipe in a direction which no actions are specified

elijaharita:
For example, this PR makes it so that if the user swipes right on a GestureActionView with no right actions, GestureActionView will ignore the swipe. This allows the swipe to bubble up and be handled by a parent instead of being swallowed.

### DIFF
--- a/src/lib/custom-animations/GestureActionView.tsx
+++ b/src/lib/custom-animations/GestureActionView.tsx
@@ -114,6 +114,7 @@ export function GestureActionView({
     },
   )
 
+  // NOTE(haileyok):
   // Absurdly high value so it doesn't interfere with the pan gestures above (i.e., scroll)
   // reanimated doesn't offer great support for disabling y/x axes :/
   const effectivelyDisabledOffset = 200
--- a/src/lib/custom-animations/GestureActionView.tsx
+++ b/src/lib/custom-animations/GestureActionView.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import {ColorValue, Dimensions, StyleSheet, View} from 'react-native'
+import {type ColorValue, Dimensions, StyleSheet, View} from 'react-native'
 import {Gesture, GestureDetector} from 'react-native-gesture-handler'
 import Animated, {
   clamp,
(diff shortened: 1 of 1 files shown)


src/lib/custom-animations/GestureActionView.tsx
```diff
@@ -114,11 +114,15 @@ export function GestureActionView({
     },
   )
 
+  // Absurdly high value so it doesn't interfere with the pan gestures above (i.e., scroll)
```
jimmylee, line 118: @haileyok 

Internally we do a comment style

```
// NOTE(githubUsername):
// Some english about what we did
```

Is there a particular convention we should follow, is it okay to start that convention here? Let me know.

jimmylee (commented):
LGTM, will want @haileyok to look at this but nice win @elijaharita

src/lib/custom-animations/GestureActionView.tsx
elijaharita: i just moved this comment from its original position without editing it. imo it would be misleading to add my name to it since i didn't actually write it.

haileyok: +1 to using that convention going forward fwiw! think its good to attribute them back w/o needing to check git blame (e.g. in this case it'll at first look like elijah made the comment since he moved it!)

haileyok: (if you feel like it you can stick `haileyok` in there following that convention)

src/lib/custom-animations/GestureActionView.tsx
```diff
@@ -114,11 +114,15 @@ export function GestureActionView({
...
+  // reanimated doesn't offer great support for disabling y/x axes :/
+  const effectivelyDisabledOffset = 200
```
haileyok, line 120: just noting that i like this naming, nice stuff

haileyok (approved):

jimmylee (approved):

src/lib/custom-animations/GestureActionView.tsx
haileyok, line 117:
```suggestion
  // NOTE(haileyok):
  // Absurdly high value so it doesn't interfere with the pan gestures above (i.e., scroll)
```

haileyok:
```suggestion
  // NOTE(haileyok):
  // Absurdly high value so it doesn't interfere with the pan gestures above (i.e., scroll)
```

haileyok: ~~wtf did github get rid of the ability to just commit a suggestion lol~~

nvm just don't have perm to commit to ur branch

elijaharita: committed! thanks!

PR #8708, opened 2025-07-23
[APP-1317] update: chat request ui to hide chat requests when there are no messages

APiligrim:
This PR addresses this [linear ticket](https://linear.app/blueskyweb/issue/APP-1317/chat-request-ui-is-confusing) about hiding `chat requests` when there are no messages in the inbox. This implementation updates the logic to only render the "Inbox" request types if at least 1 request is present.

Before:
Screenshot 2025-07-23 at 2 50 42 PM

After: 
Screenshot 2025-07-23 at 2 52 57 PM
Screenshot 2025-07-23 at 4 28 17 PM

### DIFF
--- a/src/screens/Messages/ChatList.tsx
+++ b/src/screens/Messages/ChatList.tsx
@@ -229,6 +229,7 @@ export function MessagesScreenInner({navigation, route}: Props) {
     return listenSoftReset(onSoftReset)
   }, [onSoftReset, isScreenFocused])
 
+  // NOTE(APiligrim)
   // Show empty state only if there are no conversations at all
   const actualConversations = conversations.filter(
     item => item.type === 'CONVERSATION',
--- a/src/screens/Messages/ChatList.tsx
+++ b/src/screens/Messages/ChatList.tsx
@@ -164,12 +164,18 @@ export function MessagesScreenInner({navigation, route}: Props) {
         // filter out convos that are actively being left
         .filter(convo => !leftConvos.includes(convo.id))
 
+      const hasInboxRequests = inboxPreviewConvos?.length > 0
+
       return [
-        {
-          type: 'INBOX',
-          count: inboxPreviewConvos.length,
-          profiles: inboxPreviewConvos.slice(0, 3),
-        },
+        ...(hasInboxRequests
+          ? [
+              {
+                type: 'INBOX' as const,
+                count: inboxPreviewConvos.length,
+                profiles: inboxPreviewConvos.slice(0, 3),
+              },
+            ]
+          : []),
         ...conversations.map(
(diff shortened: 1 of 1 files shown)


src/screens/Messages/ChatList.tsx
```diff
@@ -223,16 +229,23 @@ export function MessagesScreenInner({navigation, route}: Props) {
     return listenSoftReset(onSoftReset)
   }, [onSoftReset, isScreenFocused])
 
-  // Will always have 1 item - the inbox button
-  if (conversations.length < 2) {
+  // Show empty state only if there are no conversations at all
```
jimmylee, line 233: Note

```
// NOTE(APiligrim)
// Show empty state only if there are no conversations at all
```

jimmylee (commented):
Nice work, I have a comment though

I don't like these variables:

```
actualConversations
hasInboxRequests
inboxPreviewConvos
conversations
```

I know that @APiligrim inherited some of this and did a good job improving it, but can we consult @haileyok on maybe a more concise/better naming for these? Otherwise if its okay to land it incrementally thats fine too.

src/screens/Messages/ChatList.tsx
APiligrim: updated it 🙏

jimmylee:
In a brief sidebar with @haileyok we're going to merge once tests are fixed to fix this bug and resolve the linear ticket, and then we should spin up another ticket to fix variable names when we have spare time.

haileyok (approved):
awesome. appreciate the callout @jimmylee, thank you @APiligrim!

src/screens/Messages/ChatList.tsx
```diff
@@ -164,12 +164,18 @@ export function MessagesScreenInner({navigation, route}: Props) {
         // filter out convos that are actively being left
         .filter(convo => !leftConvos.includes(convo.id))
 
+      const hasInboxRequests = inboxPreviewConvos?.length > 0
```
mozzius, line 167: inboxPreviewConvos only has _unread_ requests. I have lots of requests but I can't see the button anymore because they're all  read.

APiligrim: @mozzius thank you for catching this! For the button currently when I go on my local or the live app, if I get a request and read it, it still shows the "Chat Requests" bar at the top. Does it not get shown for you?
Screenshot 2025-07-28 at 3 25 12 PM


Screenshot 2025-07-28 at 3 35 16 PM
Screenshot 2025-07-28 at 3 35 25 PM
Screenshot 2025-07-28 at 3 35 51 PM

haileyok: (yeah it does for me too fwiw!)

PR #8713, opened 2025-07-24
OTA deployments on PR comment action

haileyok:


haileyok:
@github-actions ota

src/lib/hooks/useOTAUpdates.ts
```diff
@@ -29,14 +29,137 @@ async function setExtraParams() {
...
+              style: 'default',
+              onPress: async () => {
+                await reloadAsync()
+              },
+            },
+          ],
+        )
+      } else {
+        Alert.alert(
+          'No Deployment Availalbe',
```
mozzius: typo

src/screens/Settings/Settings.tsx
```diff
@@ -392,6 +394,41 @@ function DevOptions() {
...
 
+  const {
+    onTryApplyUpdate,
+    onRevertToEmbedded,
+    isCurrentlyRunningPullRequestDeployment,
+    currentChannel,
+  } = useApplyPullRequestOTAUpdate()
+
+  const onPressApplyPullRequest = async () => {
+    Alert.prompt(
```
mozzius: iOS-only FYI

src/lib/hooks/useOTAUpdates.ts
surfdude29: also, three other of `availalbe` and two of `availalble` 👀

src/lib/hooks/useOTAUpdates.ts
```diff
@@ -60,36 +183,18 @@ export function useOTAUpdates() {
...
-        )
-      }
+      await updateTestflight()
     } catch (e: any) {
       logger.error('Internal OTA Update Error', {error: `${e}`})
     }
   }, [])
 
   React.useEffect(() => {
+    // We don't need to check anything if the curernt update is a PR update
```
surfdude29: also typo here in `current`

src/lib/hooks/useOTAUpdates.ts
haileyok: this is how you spell availalbe

src/screens/Settings/Settings.tsx
haileyok: ugh really? that's annoying. let me just do the intent ig and avoid this entirely

mozzius: Yah so annoying, would be so nice to have a simple way to prompt

haileyok: okay, i just rm'd that entirely and opted for having to use an intent

src/lib/hooks/useIntentHandler.ts
```diff
@@ -92,6 +96,18 @@ export function useIntentHandler() {
           }
           return
         }
+        case 'apply-ota': {
+          if (!IS_TESTFLIGHT) {
```
haileyok: tbh i'm not opposed to letting people other than us do this necessarily, but i'd rather make sure things dont break for now. i think it'd be useful though to actually let people do it!

app.config.js
```diff
@@ -190,7 +190,6 @@ module.exports = function (_config) {
             }
           : undefined,
         checkAutomatically: 'NEVER',
-        channel: UPDATES_CHANNEL,
```
haileyok, line 193: not supplying a channel here lets us update from _any_ channel, which is what we want. there are still checks for e.g. "will this update run on this version of the native client"

mozzius (approved):
have not tested and looks weird and strange. yolo

PR #8720, opened 2025-07-25
simplify logic for pull request OTAs

haileyok:
now that we have this working, cleaning up some of the logic and removing duplicated stuff.

also, github doesn't let you have links with deeplinks as the href, so we'll just have to render it in the browser and have you copy/paste it into the url bar.

### DIFF
--- a/src/lib/hooks/useOTAUpdates.ts
+++ b/src/lib/hooks/useOTAUpdates.ts
@@ -75,7 +75,6 @@ export function useApplyPullRequestOTAUpdate() {
     await setExtraParamsPullRequest(channel)
     const res = await checkForUpdateAsync()
     if (res.isAvailable) {
-      await fetchUpdateAsync()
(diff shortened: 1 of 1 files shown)


app.config.js
```diff
@@ -26,8 +26,7 @@ module.exports = function (_config) {
     ...(IS_DEV || IS_TESTFLIGHT ? [] : []),
   ]
 
-  const UPDATES_ENABLED =
-    IS_TESTFLIGHT !== undefined || IS_PRODUCTION !== undefined
+  const UPDATES_ENABLED = IS_TESTFLIGHT || IS_PRODUCTION
```
haileyok, line 29: Neither of these were ever going to be `undefined`, but we just want to enable OTAs whenever either of the values is true. (This prevents the weird error you can get when doing local dev about missing certificates)

mozzius (approved):

PR #8724, opened 2025-07-25
Make sure assurance request is fetched before returning true

haileyok:


haileyok:
@github-actions ota

estrattonbailey (commented):
Just be sure to check the load state of like, blocked DMs screen. Thanks for tackling this!

PR #8735, opened 2025-07-29
fix: small profile followup fixes

elijaharita:
Follow up to #8711

I'm not super sure why, but the visual spacing between the display name and handle on the profile component is different on web and ios, so I set up a native-only compensation.

Additionally, I undid the change to the left offset of the profile picture by suggestion of Samuel.

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderStandard.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderStandard.tsx
@@ -251,7 +251,7 @@ let ProfileHeaderStandard = ({
           <ProfileMenu profile={profile} />
         </View>
         <View
-          style={[a.flex_col, a.gap_sm, a.pb_sm, live ? a.pt_sm : a.pt_2xs]}>
+          style={[a.flex_col, a.gap_xs, a.pb_sm, live ? a.pt_sm : a.pt_2xs]}>
           <View style={[a.flex_row, a.align_center, a.gap_xs, a.flex_1]}>
             <Text
               emoji
@@ -261,9 +261,7 @@ let ProfileHeaderStandard = ({
                 gtMobile ? a.text_4xl : a.text_3xl,
                 a.self_start,
                 a.font_heavy,
-                {
-                  marginBottom: platform({native: -4}),
-                },
+                a.leading_tight,
               ]}>
               {sanitizeDisplayName(
                 profile.displayName || sanitizeHandle(profile.handle),


src/screens/Profile/Header/ProfileHeaderStandard.tsx
```diff
@@ -261,6 +261,9 @@ let ProfileHeaderStandard = ({
                 gtMobile ? a.text_4xl : a.text_3xl,
                 a.self_start,
                 a.font_heavy,
+                {
+                  marginBottom: platform({native: -4}),
+                },
```
estrattonbailey: The issue here is the `a.gap_sm` on line 254 above. That gap value of 8px is in addition to the line height applied to the display name and the handle, which ends up being >8px.

Not having a explicit lineHeight here is also our bad, RN handles line heights inconsistently across platforms, sometimes even cutting off descenders. So an explicit `a.leading_tight` on the display name would also help to keep all platforms consistent.

estrattonbailey (commented):
So I think the move here is actually to

(A) Ensure line heights are consistent cross-platform — the handle already has one, and `a.leading_tight` should replace the `marginBottom: -4` on the display name.

(B) Then, reduce the `gap_sm` on line 254 to `gap_xs`. Might not be pixel perfect, but it reduces decision fatigue and balances the spacing, at least to my eye in my testing.

src/screens/Profile/Header/ProfileHeaderStandard.tsx
haileyok: We can make this a later problem for sure, but something that's going to be really difficult is having e.g. Darrin come with designs that have specific pixel values then trying to translate them into this. We can chat async, I just wonder if there's some easier way to manage this (though I'm assuming you've given this a lot of thought before lol)

estrattonbailey: Yeah it's not that we haven't made efforts to do this, but we'd need to do a lot more if we truly wanted to allow for pixel perfection. Even if line heights were perfectly consistent, they inherently add a small amount of whitespace around text that needs to be compensated for. There are ways we could bake this into the component, but we just haven't invested the time to do it yet. So we're left with vibes-based spacing.

There are options though, and pry a good time to be thinking about them. I'll try to dig up something we did at a previous job and see what it would take.

elijaharita: this all makes sense, i'll update the PR

elijaharita: thanks for the learnings, i'll keep this in mind next time i'm working with vertical type margins! committed that fix

PR #8737, opened 2025-07-29
Rework variable naming, logic for rendering chat requests

haileyok:
The naming of a lot of variables here was a little confusing, so I've updated the names to make it a bit more clear and reduced the amount of work what we do to get the right results.

cc @APiligrim

### DIFF
--- a/src/screens/Messages/ChatList.tsx
+++ b/src/screens/Messages/ChatList.tsx
@@ -140,22 +140,24 @@ export function MessagesScreenInner({navigation, route}: Props) {
 
   const leftConvos = useLeftConvos()
 
-  const inboxPreviewConvos = useMemo(() => {
-    const inbox =
-      inboxData?.pages
-        .flatMap(page => page.convos)
-        .filter(
-          convo =>
-            !leftConvos.includes(convo.id) &&
-            !convo.muted &&
-            convo.unreadCount > 0 &&
-            convo.members.every(member => member.handle !== 'missing.invalid'),
-        ) ?? []
+  const inboxAllConvos =
+    inboxData?.pages
+      .flatMap(page => page.convos)
+      .filter(
+        convo =>
+          !leftConvos.includes(convo.id) &&
+          !convo.muted &&
+          convo.members.every(member => member.handle !== 'missing.invalid'),
+      ) ?? []
+  const hasInboxConvos = inboxAllConvos?.length > 0
 
-    return inbox
-      .map(x => x.members.find(y => y.did !== currentAccount?.did))
-      .filter(x => !!x)
-  }, [inboxData, leftConvos, currentAccount?.did])
+  const inboxUnreadConvos = inboxAllConvos.filter(
+    convo => convo.unreadCount > 0,
+  )
+
+  const inboxUnreadConvoMembers = inboxUnreadConvos
+    .map(x => x.members.find(y => y.did !== currentAccount?.did))
+    .filter(x => !!x)
 
   const conversations = useMemo(() => {
     if (data?.pages) {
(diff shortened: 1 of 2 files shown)


src/screens/Messages/components/InboxPreview.tsx
```diff
@@ -12,10 +12,8 @@ import {Link} from '#/components/Link'
 
 export function InboxPreview({
   profiles,
-}: // count,
-{
+}: {
   profiles: ChatBskyActorDefs.ProfileViewBasic[]
-  count: number
```
haileyok, line 18: rming, we don't use this prop

src/screens/Messages/ChatList.tsx
```diff
@@ -140,22 +140,24 @@ export function MessagesScreenInner({navigation, route}: Props) {
...
+      ) ?? []
+  const hasInboxConvos = inboxAllConvos?.length > 0
+
+  const inboxUnreadConvos = inboxAllConvos.filter(
+    convo => convo.unreadCount > 0,
+  )
+
+  const inboxUnreadConvoMembers = inboxUnreadConvos
+    .map(x => x.members.find(y => y.did !== currentAccount?.did))
+    .filter(x => !!x)
```
haileyok, line 160: Working our way down here from most to least. This was previously `inboxPreviewConvos`, but that isn't clear on:

- Is it all or just unread?
- Is it really the _convo_ or is it a _profile_

src/screens/Messages/ChatList.tsx
```diff
@@ -164,15 +166,13 @@ export function MessagesScreenInner({navigation, route}: Props) {
         // filter out convos that are actively being left
         .filter(convo => !leftConvos.includes(convo.id))
 
-      const hasInboxRequests = inboxPreviewConvos?.length > 0
-
       return [
-        ...(hasInboxRequests
+        ...(hasInboxConvos
```
haileyok, line 170: removing the second variable here with a different name and just sticking with `hasInboxConvos`

src/screens/Messages/ChatList.tsx
```diff
@@ -234,18 +234,14 @@ export function MessagesScreenInner({navigation, route}: Props) {
...
       <Layout.Screen>
         <Header newChatControl={newChatControl} />
         <Layout.Center>
-          {hasInboxRequests && (
-            <InboxPreview
-              count={inboxPreviewConvos.length}
-              profiles={inboxPreviewConvos}
-            />
+          {!isLoading && hasInboxConvos && (
+            <InboxPreview profiles={inboxUnreadConvoMembers} />
```
haileyok, line 244: The `!isLoading` just prevents a flash of the thing while we pull in the conversations. Might not need to be rendered, e.g. if you have no convos. Not a bug in prod atm since we just _always_ show the button.

src/screens/Messages/ChatList.tsx
```diff
@@ -140,22 +140,24 @@ export function MessagesScreenInner({navigation, route}: Props) {
 
   const leftConvos = useLeftConvos()
 
-  const inboxPreviewConvos = useMemo(() => {
```
haileyok, line 143: rming some unnecessary `useMemo`, no actual change though

src/screens/Messages/ChatList.tsx
```diff
@@ -234,18 +234,14 @@ export function MessagesScreenInner({navigation, route}: Props) {
   const actualConversations = conversations.filter(
```
APiligrim: could actual conversations be `conversations`  if it relates to all conversations including read and unread ones

src/screens/Messages/ChatList.tsx
APiligrim: **inboxUnreadConvoMembers** is a lot clearer! 

Do we also have another variable to check if a conversation was read or does that have to be filtered out further from `inboxAllConvos` ?

src/screens/Messages/ChatList.tsx
haileyok: hmm. what about `activeConversations`? This I _believe_ is still not technically _all_ conversations....just the ones that you accepted from the inbox :thinking:

src/screens/Messages/ChatList.tsx
haileyok: Only way I believe you can tell is by checking the `unreadyCount` on the convo

src/screens/Messages/ChatList.tsx
haileyok: (pushed a commit that does that!)

PR #8740, opened 2025-07-30
Add metrics to embedr service

amckinley:
Add support for prometheus metrics scraping, exposes new HTTP listener which defaults on port 9090 for metrics/profiling.

```
% curl localhost:9090/metrics | grep -v # | grep request
echo_requests_total{code="200",host="localhost:8100",method="GET",url="/"} 1
echo_requests_total{code="404",host="localhost:8100",method="GET",url="/foobar"} 1
promhttp_metric_handler_requests_in_flight 1
promhttp_metric_handler_requests_total{code="200"} 4
promhttp_metric_handler_requests_total{code="500"} 0
promhttp_metric_handler_requests_total{code="503"} 0
<snip>
```

```
% curl localhost:9090/debug/pprof/heap > heap.out
% file heap.out
heap.out: gzip compressed data, max speed, original size modulo 2^32 4683
```


bskyweb/cmd/embedr/server.go
```diff
@@ -65,13 +70,28 @@ func serve(cctx *cli.Context) error {
 		return err
 	}
 
+	metricsMux := http.NewServeMux()
```
jazware: Let's use a `DefaultServeMux` here and import `net/http/pprof` anonymously at the top of the file.

ericvolp12 (approved):

PR #8746, opened 2025-07-30
Convert old toast types to new ones, mark as deprecated

mozzius:
A bad typecast hid the fact that old usages of `Toast.show()` that used the fontawesome icons were not actually valid and broke at runtime. This PR:

1. Splits the type into `ToastType` and `LegacyToastType`, allowing both
2. Maps legacy types into the new ones
3. Removes the bad typecasts, making it typesafe
4. Shows deprecation warnings for usages of the legacy types

### DIFF
--- a/src/view/com/util/Toast.tsx
+++ b/src/view/com/util/Toast.tsx
@@ -32,6 +32,13 @@ import {Text} from '#/components/Typography'
 
 const TIMEOUT = 2e3
 
+// Use type overloading to mark certain types as deprecated -sfn
+// https://stackoverflow.com/a/78325851/13325987
+export function show(message: string, type?: ToastType): void
+/**
+ * @deprecated type is deprecated - use one of `'default' | 'success' | 'error' | 'warning' | 'info'`
+ */
+export function show(message: string, type?: LegacyToastType): void
 export function show(
   message: string,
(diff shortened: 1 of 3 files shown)


src/view/com/util/Toast.tsx
```diff
@@ -57,9 +75,8 @@ function Toast({
   const [cardHeight, setCardHeight] = useState(0)
 
   const toastStyles = getToastTypeStyles(t)
-  const colors = toastStyles[type as keyof typeof toastStyles]
-  const IconComponent =
-    TOAST_TYPE_TO_ICON[type as keyof typeof TOAST_TYPE_TO_ICON]
```
mozzius, line 62: I try and avoid these `as keyof typeof` casts, bit of a smell. And in fact all typecasts where it can be helped. In this case it was covering up the fact that using `xmark` would be undefined here, which causes a crash when trying to access `colors.backgroundColor`

estrattonbailey (approved):

mozzius:
Gonna merge this, we can make more tweaks in a later PR

jimmylee:
Thanks for doing this @mozzius, and welcome back, It is really appreciated of you to jump in and help us patch this.

PR #8748, opened 2025-07-30
Some toasts cleanup and reorg

estrattonbailey:
Wanted to implement some things we've talked about for a long time now that we have proper designs and all the code has been written. So the bulk of the work was already done, and this PR just:
- re-orgs files into `#/components` alongside our "new" components
- aligns styles/layout into a `Toast` component that is shared cross-platform, platform-specific wrappers remain unchanged
- leaves existing `#/view/com/util/Toast` API in place, marked as deprecated
- updates new `import {toast} from '#/components/Toast'` API to allow for other markup and custom duration

https://github.com/user-attachments/assets/c47ce67f-3ce4-420c-b67e-5f155730fcfb

https://github.com/user-attachments/assets/e4dadc1b-eaf9-4bf7-8cdd-6ddcacab88e1


estrattonbailey:
@github-actions ota

estrattonbailey:
@github-actions ota

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,204 @@
...
+import {Warning_Stroke2_Corner0_Rounded as WarningIcon} from '#/components/icons/Warning'
+import {type ToastType} from '#/components/Toast/types'
+import {Text} from '#/components/Typography'
+
+type ContextType = {
+  type: ToastType
+}
+
+export const ICONS = {
+  default: SuccessIcon,
```
APiligrim, line 17: for future reference, do we also have icons with background colors yet or if this kind of icon needs to be made from scratch? 

I was going to add support for this kind of icon in the toast too for the default option in the near future.

Screenshot 2025-07-30 at 2 51 25 PM

src/components/Toast/index.e2e.tsx
```diff
@@ -0,0 +1,5 @@
+export function ToastContainer() {
+  return null
+}
```
APiligrim, line 3: could you explain what the `.e2e.tsx` file stands for so that I also make it in the future when a new component is made 👀  is it mainly displaying the component in storybook?

src/components/Toast/types.ts
```diff
@@ -0,0 +1,24 @@
+export type ToastType = 'default' | 'success' | 'error' | 'warning' | 'info'
+
```
APiligrim, line 2: this is great for me to note that I should put types into /component/types.ts to make it clean 👍

src/style.css
```diff
@@ -369,3 +369,23 @@ input[type='range'][orient='vertical']::-moz-range-thumb {
     transform: translateY(0);
   }
 }
+
+/*
+ * #/components/Toast/index.web.tsx
+ */
+@keyframes toastFadeIn {
+  from {
```
APiligrim, line 377: are we adding these toast animations here instead of the /Toast file so that this is the animation that would be used by default globally ?

APiligrim (commented):
This looks good to me! Thank you for explaining and organizing it into the right place!

src/components/Toast/Toast.tsx
estrattonbailey: Not baked into the icons themselves, no. We use an icon set, and we try to keep the icons as close to the default set as possible. There might be a variant in the icon set like this, but if not and we need one in the future, we can def create a custom one.

src/components/Toast/index.e2e.tsx
estrattonbailey: "e2e" is "end-to-end", used in end-to-end testing, powered by Maestro. This file is just here so that toasts don't show up when running e2e tests. We only rarely need to create `.e2e.tsx` specific files.

src/components/Toast/types.ts
estrattonbailey: Not always necessary, just when it's shared between platform files. We don't have a strict pattern here, but felt natural to me in this case.

src/style.css
estrattonbailey: Mostly just moved it here so that were didn't have to append a `<style>` tag every time we open a toast on web. Might as well be a global that can be referenced as needed.

Note in this css file, these types of things are namespaced to the specific use case. We don't have a ton of styles in here, so to avoid maintenance burden we've just keep things entirely separate instead of coming up with generic animations and reusing between components.

haileyok (approved):

jimmylee (commented):
LGTM, thanks for doing this @estrattonbailey

mozzius:
Going to push up some slide in/out animations - thoughts? Can revert 

https://github.com/user-attachments/assets/397fe549-3205-495f-a64a-33d39ec0173e

PR #8758, opened 2025-07-31
[APP-1310] Button refresh

estrattonbailey:
Stacked on #8754

This PR updates the buttons to match the refreshed designs.


mozzius:
@github-actions ota

src/components/Button.tsx
```diff
@@ -442,52 +434,58 @@ export const Button = React.forwardRef<View, ButtonProps>(
       if (shape === 'default') {
         if (size === 'large') {
           baseStyles.push({
-            paddingVertical: 13,
-            paddingHorizontal: 20,
-            borderRadius: 8,
```
mozzius, line 447: Figma designs have continuous corners - we can do that with the `borderCurve` property

https://reactnative.dev/docs/view-style-props#bordercurve-ios

src/components/Button.tsx
```diff
@@ -442,52 +434,58 @@ export const Button = React.forwardRef<View, ButtonProps>(
...
-            gap: 8,
+            paddingVertical: 14,
+            paddingHorizontal: 24,
+            borderRadius: 10,
+            gap: 5,
```
mozzius: This seems quite narrow

src/components/Button.tsx
estrattonbailey: Oh yeah they do, missed that

src/components/Button.tsx
estrattonbailey: Designs call for 4, I felt 5 was better. I'll mock up some more icons, since those vary in the amount of whitespace they contain, and eyeball a value.

estrattonbailey:
Talked to Sam, we good here

PR #8764, opened 2025-07-31
setup i18n

BinaryFiddler:


haileyok:
At a quick glance this should be fine. I'm used to `.po` files for this, but the JSON is pretty readable and considering this is a really small thing that will likely not get updated much, I'm pretty fine with this (and it's easier to either A. generate some translations for or B. have third parties generate locales for), so totally okay with this approach.

I'll try to get the dev-env working to verify everything is good, but code looks correct from the glance.

Thank you so much :pray:

jimmylee:
@BinaryFiddler needs a lint pass through with prettier

Signez:
Hey there! Just two cents from Bluesky's translators community: it should be really easy to simply add those files to Crowdin: it should literally be a matter of updating [`crowdin.yml`](https://github.com/bluesky-social/social-app/blob/main/crowdin.yml#L7-L20) to add this, and everything should work automagically after that! Crowdin supports this kind of translations in JSON files, so it should be really that simple.

PR #8767, opened 2025-08-01
Tweak accessibility label

surfdude29:
@nilaallj raised the following issue [on Crowdin](https://bluesky.crowdin.com/editor/1/11/en-sv/7#8734):

> This is an accessibility label, right? I’m not sure if “toast” is a good term to use outside of a developer context. A blind user might not understand what is being dismissed here. Wouldn’t it be better to use something like “message” and maybe “alert” for important toasts?

I'm not too familiar with accessibility guidelines and practices, so I decided to ask Claude, and he did [some helpful research](https://claude.ai/share/99264d01-077f-4ee4-ab28-b2a795e76348).

Given what Claude found, I think referring to a toast as an `alert` might be the most user-friendly option. While either `message` or `notification` could work, given that we already use those words extensively in the UI (and accessibility labels) to refer to other things, I think it might not be clear – to both screen-reader users and translators – what's being referred to in this context.

### DIFF
--- a/src/components/Toast/index.web.tsx
+++ b/src/components/Toast/index.web.tsx
@@ -79,7 +79,10 @@ export const ToastContainer: React.FC<ToastContainerProps> = ({}) => {
           <Pressable
             style={[a.absolute, a.inset_0]}
             accessibilityLabel={_(
-              msg({message: 'Dismiss alert', context: 'toast'}),
+              msg({
+                message: `Dismiss message`,
+                comment: `Refers to dismissing a toast notification`,
+              }),
             )}
             accessibilityHint=""
             onPress={() => setActiveToast(undefined)}
--- a/src/components/Toast/index.web.tsx
+++ b/src/components/Toast/index.web.tsx
@@ -81,7 +81,7 @@ export const ToastContainer: React.FC<ToastContainerProps> = ({}) => {
             accessibilityLabel={_(
               msg({
                 message: `Dismiss message`,
-                comment: `Refers to dismissing a toast notification`,
+                comment: `Accessibility label for dismissing a toast notification`,
               }),
             )}
             accessibilityHint=""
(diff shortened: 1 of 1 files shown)


nilaallj:
Thanks for raising this.

For screen reader users, the word “alert” carries a stronger sense of urgency and interruption than something like “message”. Alerts are typically used for critical things that need immediate attention, and they can interrupt what the screen reader is currently doing. That can be frustrating for the user unless the interruption is justified.

Most toasts don’t behave that way. They’re usually informational, non-disruptive and more like messages the user can read when they’re ready. That’s why I think a label like “Dismiss message” would be a better fit in most cases. It more accurately reflects the nature of the content and avoids suggesting urgency where there is none. In my opinion, “Dismiss alert” should only be used on toasts that are actual alerts (with the behaviour I described above). 🙂

APiligrim:
Would it make sense to say `Dismiss _(name of the toastType)__`  so that its descriptive of the toast? For the error ones we could say `alert` since it's something that alerts the user of an error but for other more informative ones as @nilaallj mentioned, we could use other common names such as `Dismiss message` or `Dismiss warning`

surfdude29:
If it would be fairly straightforward to implement, the solution proposed by @APiligrim seems like it might be ideal.

However, if it's felt that it would add too much complexity, I think @nilaallj does make a strong argument for simply using `message` instead of `alert`, so for the time being I've changed it to say that. (I also added a comment for translators, to make it clear where this string is used, instead of the context which I used erroneously.)

nilaallj:
@surfdude29 I agree “Dismiss message” probably is the best solution for now.

I like @APiligrim’s suggestion too. Maybe it’s better to simply use “error message” instead of “alert” if we’re going for that? I don’t think the alert behaviour must imply any particular toast type or vice versa. It depends on how that functionality is implemented, if implemented at all. 🙂

PR #8768, opened 2025-08-02
feat: resolve pds url from handles in login screen

Turtlepaw:
### What's in this PR?

- Implements `resolvePdsServiceUrl` based on [a-viv-a/deer-social](https://github.com/a-viv-a/deer-social/blob/main/src/state/queries/direct-fetch-record.ts#L125) to fetch the DID document using plc.directory or did:web
- In logins screen, after a `.` is detected in the identifier, it will attempts to resolve the PDS URL from the identifier

> [!NOTE]
> Requests are debounced to prevent excessive requests

### Preview

https://github.com/user-attachments/assets/7154ac21-40ea-4479-b534-7b07ac3f1223

### DIFF
--- a/src/screens/Login/LoginForm.tsx
+++ b/src/screens/Login/LoginForm.tsx
@@ -47,6 +47,8 @@ export const LoginForm = ({
   onPressForgotPassword,
   onAttemptSuccess,
   onAttemptFailed,
+  debouncedResolveService,
+  isResolvingService,
 }: {
   error: string
   serviceUrl: string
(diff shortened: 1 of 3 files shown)


Turtlepaw:
related: https://github.com/zeppelin-social/social-app/pull/2

Turtlepaw:
Also, it could also instead "Bluesky Social" it could say "Automatic" to signal that the PDS will be resolved automatically.


https://github.com/user-attachments/assets/2bb35d65-d698-4260-b28c-00de5fa041af

PR #8774, opened 2025-08-04
Fix video thumbnails on native

mozzius:
https://github.com/bluesky-social/social-app/pull/8754 caused a background to be added to the overlay button. Simply adding `a.bg_transparent` fixes it.

While I'm here, I removed the `forwardRef` (React 19 wooo) and removed an unnecessary `useEffect`.

I also bumped `expo-image` while figuring out the fix - not necessary as it turns out, but it seems to have some good fixes as per the [changelog](https://github.com/expo/expo/blob/main/packages/expo-image/CHANGELOG.md). Feel free to revert that commit if you're rather not

https://github.com/user-attachments/assets/d57e7591-972a-40d8-b15a-ac1917730cc5

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/index.tsx
+++ b/src/components/Post/Embed/VideoEmbed/index.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useState} from 'react'
+import {useCallback, useRef, useState} from 'react'
 import {ActivityIndicator, View} from 'react-native'
 import {ImageBackground} from 'expo-image'
 import {type AppBskyEmbedVideo} from '@atproto/api'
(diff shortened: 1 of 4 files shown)


src/components/Post/Embed/VideoEmbed/index.tsx
```diff
@@ -127,7 +125,12 @@ function InnerWrapper({embed}: Props) {
       >
         {showOverlay && (
           <Button
-            style={[a.flex_1, a.align_center, a.justify_center]}
+            style={[
+              a.flex_1,
+              a.align_center,
+              a.justify_center,
+              a.bg_transparent,
```
estrattonbailey: I think all we gotta do is remove the stray `color="secondary"` on line 138. After my first button PR, if `color` is set, we assume we want a "button" style. Here this button really might as well be a `Pressable`. I'll test this out quick and push a fix since I caused the confusion 👍

estrattonbailey (approved):
Nice catch!

PR #8775, opened 2025-08-04
Fix android starting at notif screen

mozzius:
Not tested, because I can't repro, but we're getting reports of something similar to https://github.com/bluesky-social/social-app/pull/8657

The cause last time was deep links showing up as "notification" when fetching them at launch. Clearly, some devices have something similar going on, where it shows some sort of notification at boot. Our fallback behaviour is to redirect to `/notifications` by default, but clearly this is a bad default. Let's redirect to `/notifications` only if it seems to be one our _our_ notifications, by checking for `uri` in the payload

# Test plan

If your android device launches to the notif screen, confirm this fixes it. Otherwise idk. yolo i guess

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -400,9 +400,7 @@ export function getNotificationPayload(
   }
 }
 
-export function notificationToURL(
-  payload: NotificationPayload,
-): string | undefined {
+export function notificationToURL(payload: NotificationPayload): string | null {
   switch (payload?.reason) {
     case 'like':
     case 'repost':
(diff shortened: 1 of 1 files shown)


src/lib/hooks/useNotificationHandler.ts
```diff
@@ -433,10 +431,19 @@ export function notificationToURL(
...
     case 'verified':
     case 'unverified':
-    default:
       return '/notifications'
+    default:
+      // little heuristic - if it's an unknown notification type with a `uri`,
+      // assume it's a valid notification that we don't know about yet.
+      if (payload && (payload as any)?.uri) {
+        return '/notifications'
+      } else {
```
mozzius: Alternatively, we could just do nothing

estrattonbailey (approved):
I did notice this once, but can't repro right now. Iirc, it opened to Notifications and I didn't even have new notifications, so I think you might be right on this one and it might have been an Android thing. Reasonable imo to assume `payload.uri` is from us, and otherwise do nothing. If this doesn't work then yeah we could probably just do nothing.

estrattonbailey:
Although, is there any harm in doing nothing regardless? We _know_ what notifications we send, and we handle those. Do we even need a happy path for forward compatibility? I assume any new notifications we add to backend we will also to the client

mozzius:
More than happy to just do nothing if not recognised. I'll push that up (was too focused on maintaining prior behaviour)

PR #8776, opened 2025-08-04
Fix web video ViewportObserver component

mozzius:
Reverts #8692 

Deeper investigation reveals that when we changed the video cropping logic, we were cutting off the "sensor" div that sticks out of the post. This is the root cause of #8692, so we can revert.

Code to review is https://github.com/bluesky-social/social-app/pull/8776/commits/86757f004e43c72009a8d101f2ffdaf719794487

# Before

Position only updates when the video touches the edge of the screen

https://github.com/user-attachments/assets/3cb96857-d256-4d70-8c5d-7e82193ee523

# After

https://github.com/user-attachments/assets/cf2a11da-300e-4add-9fbd-4f45e92dc344

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/index.web.tsx
+++ b/src/components/Post/Embed/VideoEmbed/index.web.tsx
@@ -1,4 +1,11 @@
-import {useCallback, useEffect, useRef, useState} from 'react'
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react'
 import {View} from 'react-native'
 import {type AppBskyEmbedVideo} from '@atproto/api'
 import {msg} from '@lingui/macro'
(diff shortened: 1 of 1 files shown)


src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -83,41 +90,47 @@ export function VideoEmbed({
       style={{display: 'flex', flex: 1, cursor: 'default'}}
       onClick={evt => evt.stopPropagation()}>
       <ErrorBoundary renderError={renderError} key={key}>
-        <ViewportObserver
-          sendPosition={sendPosition}
-          isAnyViewActive={currentActiveView !== null}>
+        <OnlyNearScreen>
```
mozzius, line 93: We need to hide the children of this component, but the ViewportObserver needs to be lifted out. So we pass context and hide based on that. A little convoluted

src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -213,6 +197,18 @@ function ViewportObserver({
   )
 }
 
+/**
+ * Awkward data flow here, but we need to hide the video when it's not near the screen.
```
APiligrim, line 201: thank you for these comments 🙏

PR #8779, opened 2025-08-05
[APP-1365]: Ensure toast text meets a11y guidelines

APiligrim:
This PR addresses this [linear ticket ](https://linear.app/blueskyweb/issue/APP-1365/ensure-toast-text-meets-a11y-guidelines) to ensure toast text meets a11y guidelines as well as improves the success icon to match other icon styles in the toast module. 

The latest toast design in Figma:

Screenshot 2025-08-04 at 11 48 13 PM

### **Before:**

https://github.com/user-attachments/assets/d6bee570-4c48-404d-9970-4bd2dc118a84

### **After:**

`Note` in Figma the borderWidth is 0.5px but according to @mozzius , since we are using a variable that is platform specific, it’s 0.333px width on native, but that’s not possible to achieve on web so it’s 1px instead. 

Screenshot 2025-08-06 at 12 41 07 PM
Screenshot 2025-08-06 at 12 41 22 PM
Screenshot 2025-08-06 at 12 41 33 PM

### DIFF
--- a/src/components/Toast/Toast.tsx
+++ b/src/components/Toast/Toast.tsx
@@ -116,15 +116,15 @@ function useToastStyles({type}: {type: ToastType}) {
         backgroundColor: select(t.name, {
           light: t.palette.primary_25,
           dim: t.palette.primary_25,
-          dark: t.palette.primary_50,
+          dark: t.palette.primary_25,
         }),
         borderColor: select(t.name, {
           light: t.palette.primary_300,
           dim: t.palette.primary_600,
-          dark: t.palette.primary_500,
+          dark: t.palette.primary_300,
         }),
         iconColor: select(t.name, {
-          light: t.palette.primary_500,
+          light: t.palette.primary_600,
           dim: t.palette.primary_600,
           dark: t.palette.primary_600,
         }),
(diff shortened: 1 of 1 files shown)


src/components/Toast/Toast.tsx
```diff
@@ -93,34 +93,34 @@ function useToastStyles({type}: {type: ToastType}) {
...
+          dark: t.atoms.text.color,
         }),
       },
       success: {
         backgroundColor: select(t.name, {
-          light: t.palette.primary_100,
-          dim: t.palette.primary_100,
+          light: t.palette.primary_25,
+          dim: t.palette.primary_25,
           dark: t.palette.primary_50,
```
estrattonbailey: Realize this isn't marked as ready, but just wanted to note that I think a few values don't match Figma, including this one, which I think should be `primary_25` as well.

If all these values are the same for all themes, we can get rid of all the `select` calls here 👍

APiligrim:
@estrattonbailey do you happened to know what this variable is in code? I searched for a variable that ends with `1000` but could not find it in `themes.ts`. I'm currently using `t.atoms.text.color` for it 👀 

Also `text.default` in Figma is a regular `t.atoms.text.color` in code right?

Screenshot 2025-08-06 at 12 33 55 PM

estrattonbailey (approved):
Thanks!

estrattonbailey:
@APiligrim yep yep, you handled this correctly, good question

PR #8782, opened 2025-08-05
[APP-1356] Policy update dialog

estrattonbailey:
Bluesky will soon be updating our Terms of Service, Privacy Policy, Copyright Policy, and Community Guidelines. This PR adds an announcement dialog that users will need to complete before using the app. More info coming soon!


src/components/PolicyUpdateOverlay/context.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+  return context
+}
+
+export function Provider({children}: {children?: ReactNode}) {
+  const state = usePolicyUpdateState()
+
+  return (
+    <PortalProvider>
+      <Context.Provider value={useMemo(() => state, [state])}>
```
haileyok: fwiw you should be able to stick `state` directly in this, don't need the `useMemo`

estrattonbailey: Ah true, was initially passing in other props too, will clean up

src/screens/Settings/AboutSettings.tsx
```diff
@@ -179,10 +185,57 @@ export function AboutSettingsScreen({}: Props) {
                   </SettingsList.ItemText>
                 </SettingsList.PressableItem>
               )}
+
+              <PolicyUpdateDebug />
```
mozzius: smallest nit of all time but I was expecting to find it with these options on the settings home screen

Screenshot 2025-08-06 at 21 17 37

estrattonbailey: Ah yeah lemme move it there

src/view/shell/index.tsx
```diff
@@ -160,8 +166,16 @@ function ShellInner() {
...
+
+      {/* Until policy update has been completed by the user, don't render anything that is portaled */}
+      {policyUpdateState.completed && (
+        <>
+          <PortalOutlet />
+          <BottomSheetOutlet />
+        </>
+      )}
+
+      <PolicyUpdateOverlayPortalOutlet />
```
mozzius, line 178: Maybe do this on web too? I got the activity notifs NUX underneath

src/components/FocusScope/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  return (
+    <>
+      <Pressable
+        accessible
+        accessibilityLabel={_(
+          msg`You've reached the start of the active content.`,
+        )}
+        accessibilityHint={_(
+          msg`Please go back, or activate this element to return to the start of the active content.`,
```
surfdude29, line 93: I don't quite understand what's going on here, so apologies if I'm off track, but I'm wondering if the `accessibilityHint`s here and on line 117 should be the same, even though the `accessibilityLabel`s are different?

src/view/shell/index.tsx
estrattonbailey: Yeah tbh even safer on web to do this, pry the right move

src/components/FocusScope/index.tsx
estrattonbailey: Good question. These "guards" are focusable by screen readers, and only rendered when a screen reader is enabled. The idea is that if you hit one, you (a) know if you're at the start or end of the "active content" and (b) we tell you to go back the way you came or activate (double tap) to go back to the start of the active content.

Really only used on Android atm, iOS never hits these, and web implementation is totally separate.

surfdude29: Got it, appreciate the explanation 🙏

PR #8784, opened 2025-08-06
[APP-1357] profile header follow recommendations

caidanw:
Fixes https://github.com/bluesky-social/social-app/issues/8849

Summary
---

Add a new feature to show suggested follow recommendations when you follow an account.

New components:
- `ProfileHeaderSuggestedFollows` is the main component that implements the layout of the suggested profiles.
- `AnimatedProfileHeaderSuggestedFollows` is a wrapper that uses `AccordionAnimation` to wrap the `ProfileHeaderSuggestedFollows`, also handles the gated feature for A/B testing (prevents any data loading or component rendering early to avoid extra work).
- `AccordionAnimation` is a generic animation component that is reusable should we need to re-use the same animation somewhere.

Other changes:
- Add new `statsig` gate name, `post_follow_profile_suggested_accounts`
- Some changes were cherry-picked from @mozzius branch.
- Disable preview on card Avatar to avoid nested `<a>` tags.
- Add label for `SuggestedFolllowPlaceholder` button.
- There are two separate animations for web and mobile, as mobile couldn't handle the constant amount of layout calculations during animation.

This is a gated feature for use with A/B testing, if you pull down this branch and use it locally you may need to comment out the gated feature logic.

Due to issues stemming from Androids gesture handling, we can not support this feature on Android yet. It's been discussed and agreed upon that we should omit this feature on Android for now, and use iOS and web to gather metrics about engagement. Hopefully in the future when the profile header is re-architected it'll be in a better state to handle this on Android.

Screenshots
---

Web:

https://github.com/user-attachments/assets/32d69708-67f9-4b63-b96b-8025fb22949c

Mobile:

https://github.com/user-attachments/assets/cba28d83-57b2-48e4-a081-97f56e225174

### DIFF
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -82,7 +82,6 @@ export function SuggestedFollowPlaceholder() {
         <Button
           label={_(msg`Follow account placeholder`)}
           size="small"
-          variant="solid"
           color="secondary"
           disabled
           style={[a.w_full, a.rounded_sm]}>
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -8,6 +8,7 @@ import {useNavigation} from '@react-navigation/native'
 import {type NavigationProp} from '#/lib/routes/types'
 import {logEvent} from '#/lib/statsig/statsig'
 import {logger} from '#/logger'
+import {isIOS} from '#/platform/detection'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
 import {useGetPopularFeedsQuery} from '#/state/queries/feed'
 import {type FeedDescriptor} from '#/state/queries/post-feed'
@@ -367,7 +368,8 @@ export function ProfileGrid({
         !isProfileHeaderContext && a.border_t,
         t.atoms.border_contrast_low,
         t.atoms.bg_contrast_25,
-      ]}>
+      ]}
+      pointerEvents={isIOS ? 'auto' : 'box-none'}>
       <View
         style={[
           a.px_lg,
@@ -375,7 +377,8 @@ export function ProfileGrid({
           a.flex_row,
           a.align_center,
           a.justify_between,
-        ]}>
+        ]}
+        pointerEvents={isIOS ? 'auto' : 'box-none'}>
         <Text style={[a.text_sm, a.font_bold, t.atoms.text]}>
           {isFeedContext ? (
             <Trans>Suggested for you</Trans>
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -256,107 +256,105 @@ export function ProfileGrid({
   const maxLength = gtMobile ? 3 : isProfileHeaderContext ? 12 : 6
   const minLength = gtMobile ? 3 : 4
 
-  const content = isLoading ? (
-    Array(maxLength)
-      .fill(0)
-      .map((_, i) => (
-        <View
-          key={i}
-          style={[
-            a.flex_1,
-            gtMobile &&
-              web([
-                a.flex_0,
-                a.flex_grow,
-                {width: `calc(30% - ${a.gap_md.gap / 2}px)`},
-              ]),
-          ]}>
-          <SuggestedFollowPlaceholder />
-        </View>
-      ))
-  ) : error || !profiles.length ? null : (
-    <>
-      {profiles.slice(0, maxLength).map((profile, index) => (
-        <ProfileCard.Link
-          key={profile.did}
-          profile={profile}
-          onPress={() => {
-            logEvent('suggestedUser:press', {
-              logContext: isFeedContext
-                ? 'InterstitialDiscover'
-                : 'InterstitialProfile',
-              recId,
-              position: index,
-            })
-          }}
-          style={[
-            a.flex_1,
-            gtMobile &&
-              web([
-                a.flex_0,
-                a.flex_grow,
-                {width: `calc(30% - ${a.gap_md.gap / 2}px)`},
-              ]),
-          ]}>
-          {({hovered, pressed}) => (
-            <CardOuter
-              style={[(hovered || pressed) && t.atoms.border_contrast_high]}>
-              <ProfileCard.Outer>
-                <View
-                  style={[
-                    a.flex_col,
-                    a.align_center,
-                    a.gap_sm,
-                    a.pb_sm,
-                    a.mb_auto,
-                  ]}>
-                  <ProfileCard.Avatar
-                    profile={profile}
-                    moderationOpts={moderationOpts}
-                    disabledPreview
-                    size={88}
-                  />
-                  <View style={[a.flex_col, a.align_center, a.max_w_full]}>
-                    <ProfileCard.Name
+  const content = isLoading
+    ? Array(maxLength)
+        .fill(0)
+        .map((_, i) => (
+          <View
+            key={i}
+            style={[
+              a.flex_1,
+              gtMobile &&
+                web([
+                  a.flex_0,
+                  a.flex_grow,
+                  {width: `calc(30% - ${a.gap_md.gap / 2}px)`},
+                ]),
+            ]}>
+            <SuggestedFollowPlaceholder />
+          </View>
+        ))
+    : error || !profiles.length
+      ? null
+      : profiles.slice(0, maxLength).map((profile, index) => (
+          <ProfileCard.Link
+            key={profile.did}
+            profile={profile}
+            onPress={() => {
+              logEvent('suggestedUser:press', {
+                logContext: isFeedContext
+                  ? 'InterstitialDiscover'
+                  : 'InterstitialProfile',
+                recId,
+                position: index,
+              })
+            }}
+            style={[
+              a.flex_1,
+              gtMobile &&
+                web([
+                  a.flex_0,
+                  a.flex_grow,
+                  {width: `calc(30% - ${a.gap_md.gap / 2}px)`},
+                ]),
+            ]}>
+            {({hovered, pressed}) => (
+              <CardOuter
+                style={[(hovered || pressed) && t.atoms.border_contrast_high]}>
+                <ProfileCard.Outer>
+                  <View
+                    style={[
+                      a.flex_col,
+                      a.align_center,
+                      a.gap_sm,
+                      a.pb_sm,
+                      a.mb_auto,
+                    ]}>
+                    <ProfileCard.Avatar
                       profile={profile}
                       moderationOpts={moderationOpts}
+                      disabledPreview
+                      size={88}
                     />
-                    <ProfileCard.Description
-                      profile={profile}
-                      numberOfLines={2}
-                      style={[
-                        t.atoms.text_contrast_medium,
-                        a.text_center,
-                        a.text_xs,
-                      ]}
-                    />
+                    <View style={[a.flex_col, a.align_center, a.max_w_full]}>
+                      <ProfileCard.Name
+                        profile={profile}
+                        moderationOpts={moderationOpts}
+                      />
+                      <ProfileCard.Description
+                        profile={profile}
+                        numberOfLines={2}
+                        style={[
+                          t.atoms.text_contrast_medium,
+                          a.text_center,
+                          a.text_xs,
+                        ]}
+                      />
+                    </View>
                   </View>
-                </View>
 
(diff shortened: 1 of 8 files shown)


src/components/Button.tsx
```diff
@@ -531,6 +531,7 @@ export const Button = React.forwardRef<View, ButtonProps>(
           a.flex_row,
           a.align_center,
           a.justify_center,
+          a.curve_continuous,
```
mozzius: I think leave this off for now, we'll put it in #8758

src/lib/custom-animations/AccordionAnimation.tsx
```diff
@@ -0,0 +1,42 @@
...
+  children,
+}: AccordionAnimationProps) {
+  const height = useSharedValue(0)
+  const derivedHeight = useDerivedValue(() => {
+    const targetHeight = isExpanded ? height.value : 0
+    return withTiming(targetHeight, {duration})
+  }, [isExpanded, duration])
+
+  const bodyStyle = useAnimatedStyle(() => ({
+    height: derivedHeight.value,
```
mozzius: nit: prefer `sharedValue.get()` instead of `sharedValue.value` - it makes React Compiler sad

```suggestion
    height: derivedHeight.get(),
```

mozzius: Animating height causes Yoga to relayout every frame, which is slow - especially in this case where it'll then have to relayout everything below it.

Generally, you should prefer animating GPU-accelerate-able properties such as transforms (similar to CSS)

src/components/FeedInterstitials.tsx
```diff
@@ -385,17 +390,19 @@ export function ProfileGrid({
...
-          to="/search">
-          <Trans>See more</Trans>
-        </InlineLinkText>
+        {viewContext !== 'profileHeader' && (
+          <InlineLinkText
+            label={_(msg`See more suggested profiles on the Explore page`)}
+            to="/search">
+            <Trans>See more</Trans>
+          </InlineLinkText>
+        )}
```
mozzius: Missing the Follow All button (also I don't think it should have a gray background?)

Screenshot 2025-08-06 at 08 05 02

src/components/Button.tsx
caidanw: Done, thanks!

src/lib/custom-animations/AccordionAnimation.tsx
caidanw: Got it, wasn't aware of this (the docs used `.value`), updating now.

caidanw: Yeah ran into that myself, ended up using a different animation for the mobile view. The height animation on web looks and runs great.

src/components/FeedInterstitials.tsx
caidanw: Darrin created a new design, this was the one he gave me (link in the linear ticket). Main changes being the background and drop shadow on the cards. There was a previous conversation (on slack, link to it in linear ticket) that discussed excluding the "Follow all" functionality, though it may fit better in this context.

image

src/components/FeedInterstitials.tsx
```diff
@@ -79,9 +80,8 @@ export function SuggestedFollowPlaceholder() {
         </View>
 
         <Button
-          label=""
+          label={_(msg`Follow account placeholder`)}
```
jimmylee: Do you want this to say `Follow account placeholder`?

jimmylee (commented):
You got a pretty good grip on the styling code, I need to confirm that is the right placeholder with you and I think Darrin should probably confirm the design with @mozzius but other than that this code looks good!

src/screens/Profile/Header/SuggestedFollows.tsx
```diff
@@ -0,0 +1,37 @@
+import {AccordionAnimation} from '#/lib/custom-animations/AccordionAnimation'
+import {useGate} from '#/lib/statsig/statsig'
+import {useSuggestedFollowsByActorQuery} from '#/state/queries/suggested-follows'
+import {ProfileGrid} from '#/components/FeedInterstitials'
+
+export function ProfileHeaderSuggestedFollows({actorDid}: {actorDid: string}) {
+  const {isLoading, data, error} = useSuggestedFollowsByActorQuery({
+    did: actorDid,
+  })
```
mozzius, line 10: You'll notice this flashes with new content every time you open it. I think we need to add `staleTime: STALE.MINUTES.FIVE` so that it doesn't need to refetch

src/lib/custom-animations/AccordionAnimation.tsx
```diff
@@ -0,0 +1,76 @@
...
+  style,
+  children,
+}: AccordionAnimationProps) {
+  if (!isExpanded) return null
+
+  return (
+    <Animated.View
+      style={style}
+      entering={FadeInUp.duration(duration)}
+      exiting={FadeOutDown.duration(duration / 2)}>
```
mozzius: I think `FadeOutUp` would work better here

src/components/FeedInterstitials.tsx
mozzius: ah, ok 👍 I'll check with Darrin because it's not clear which ones are the latest in the "Project Activation" file

src/components/FeedInterstitials.tsx
```diff
@@ -385,17 +390,19 @@ export function ProfileGrid({
...
-          label={_(msg`See more suggested profiles on the Explore page`)}
-          to="/search">
-          <Trans>See more</Trans>
-        </InlineLinkText>
+        {viewContext !== 'profileHeader' && (
+          <InlineLinkText
+            label={_(msg`See more suggested profiles on the Explore page`)}
+            to="/search">
+            <Trans>See more</Trans>
+          </InlineLinkText>
```
mozzius, line 394: Maybe it's something for another PR but it doesn't feel great just linking to the search page, especially given it doesn't show the follow suggestions at the top so it seems unrelated

It would be great to implement this page:

Screenshot 2025-08-08 at 11 08 56

src/components/FeedInterstitials.tsx
caidanw: Confirming that this is the correct placeholder, I built it myself in a previous PR. It matches all the same structure as the actual profile card, but in skeleton form. So that there is no content layout shift once the content is loaded in.

This button previously had no label due to it being disabled and un-clickable, but I felt that for folks who use screen readers it would be useful for them to know why they can't click the button. I think "Follow account placeholder" is an apt description, but I'm open to changing it.

src/screens/Profile/Header/SuggestedFollows.tsx
caidanw: Ooh good catch, I was noticing that too. The problem is that this call is being made 2 times, once for the this component. And one time (as the page loads) down on the user's "Posts feed", since there is another interstitial with the same content that existed there before this PR.

caidanw: Fixed, added it to the `useSuggestedFollowsByActorQuery` hook with a default value.

src/components/FeedInterstitials.tsx
caidanw: Yeah the code you highlighted was implemented in a previous PR. The decision was to keep existing functionality of linking to the search page for now, since that was the existing behavior.

From what I remember, we are waiting on an API or other feature (memory is a bit fuzzy) before we can fully build out the "Suggested follow" screen.

I'm not sure if we have a linear ticket for this screen or not, but it was planned to be implemented later.

src/components/FeedInterstitials.tsx
mozzius: Does it have to be a button? Why not make a same-sized View instead?

src/components/FeedInterstitials.tsx
mozzius: Ok sounds good! Let's make sure we're tracking this in Linear

src/components/FeedInterstitials.tsx
surfdude29:
```suggestion
          label={_(
            msg({
              message: `Follow button placeholder`,
              comment: `Accessibility label for the unclickable Follow button placeholder shown while content is loading`,
            }),
          )}
```
I think something like `Follow button placeholder` might be a bit clearer?

Also, a comment for translators would be helpful to add here 🙏

caidanw: @mozzius definitely can make a view, just figured it would be more accurate to use the disabled button in the event of changes to button styling (we wouldn't need to update the skeleton view).

caidanw: @surfdude29 thanks, that does seem a bit clearer

mozzius:
Before we merge - reminder to self to check scroll behaviour

mozzius (changes requested):
Yeah sorry this is broken on Android and really glitchy on iOS unfortunately. I think we need to find a fix before we can merge this - at least for android. or just gate to iOS or web

# Android

Can't scroll past it, just swallows swipes. Also can't scroll through the suggestions either - it's just a dead zone


https://github.com/user-attachments/assets/c4f246b5-db86-4be5-ab46-90ea6cd8e003

# iOS

Freaks out when you swipe on it whilst it's momentum scrolling

Note: this is probably caused by the native Scroll Forwarder module in
`modules/expo-scroll-forwarder`

https://github.com/user-attachments/assets/8e3d7a93-3b25-427e-a55a-13788af7d9f2

 I think this is why this component was removed in the first place :(

jimmylee:
Thanks for helping to test @mozzius - It is a good reminder for me to make sure I run Android as well

caidanw:
> Yeah sorry this is broken on Android and really glitchy on iOS unfortunately. I think we need to find a fix before we can merge this - at least for android. or just gate to iOS or web
> 
> # Android
> Can't scroll past it, just swallows swipes. Also can't scroll through the suggestions either - it's just a dead zone
> 
>  Screen.Recording.2025-08-13.at.10.56.07.mov 
> # iOS
> Freaks out when you swipe on it whilst it's momentum scrolling
> 
> Note: this is probably caused by the native Scroll Forwarder module in `modules/expo-scroll-forwarder`
> 
>  Simulator.Screen.Recording.-.iPhone.16.-.2025-08-13.at.10.54.54.mp4 
> I think this is why this component was removed in the first place :(

Thanks for catching this, I'll double check on iOS and Android. Working on finding a solution to this problem.

Edit: I just tried it on my iOS simulator. I'm using an iPhone 16 with iOS 18.5 and I can't reproduce this issue. Can you provide more details about the specific simulator you are using.

mozzius:
Nice work fixing the scrolling! Good find

I don't know if this is an inherent Android limitation, but it's a shame that you can't scroll the page up and down when swiping on it. It works on the other interstitials - I guess because it's absolutely positioned over the scrollview? I think this is acceptable if we can't find a fix but might be worth having a look

https://github.com/user-attachments/assets/8ab42a1b-3431-4d75-b3b3-cdaf774b8071

src/screens/Profile/Header/ProfileHeaderStandard.tsx
```diff
@@ -155,174 +159,185 @@ let ProfileHeaderStandard = ({
...
+                    />
+                  </View>
+                )}
+            </View>
+          )}
+        </View>
+
+        <Prompt.Basic
+          control={unblockPromptControl}
+          title={_(msg`Unblock Account?`)}
```
surfdude29, line 322:
```suggestion
          title={_(msg`Unblock account?`)}
```
micronit: I know this isn't exactly related to this PR, but as we pretty much always use sentence case for dialog titles nowadays, it might be nice to tweak this :)

surfdude29 (commented):
I can't test it, but it looks really nice from the screen recordings!

Just one tiny suggestion (not too important), otherwise, for the (non-code) parts that I understand, LGTM 👌

src/screens/Profile/Header/ProfileHeaderStandard.tsx
caidanw: If you turn on "hide whitespace" in the github diff view, you'll notice that I didn't actually touch this line. I've searched through the code base for uses of "Unblock Account" vs "Unblock account" and it seems pretty mixed. So I'll leave this out of this PR for now, but we'll likely want to standardize on it in a separate ticket.

surfdude29: Oops, my bad, yeah that can be followed up on sometime later 👍

PR #8791, opened 2025-08-07
Toast tweaks

estrattonbailey:
Made some small adjustments and fixed the e2e test error.

### DIFF
--- a/src/components/Toast/Toast.tsx
+++ b/src/components/Toast/Toast.tsx
@@ -1,7 +1,7 @@
 import {createContext, useContext, useMemo} from 'react'
 import {View} from 'react-native'
 
-import {atoms as a, select, useTheme} from '#/alf'
+import {atoms as a, select, useAlf, useTheme} from '#/alf'
 import {CircleInfo_Stroke2_Corner0_Rounded as CircleInfo} from '#/components/icons/CircleInfo'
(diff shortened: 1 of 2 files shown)


src/components/Toast/Toast.tsx
```diff
@@ -32,9 +32,19 @@ export function Toast({
...
   const Icon = ICONS[type]
+  /**
+   * Vibes-based number, adjusts `top` of `View` that wraps the text to
+   * compensate for different type sizes and keep the first line of text
+   * aligned with the icon. - esb
+   */
+  const fontScaleCompensation = useMemo(
+    () => parseInt(fonts.scale) * -1 * 0.65,
+    [fonts.scale],
+  )
```
mozzius, line 47: This is fine for now but we should come up with some sort of generic utility for this. Esp if we can figure out a way to match the system font scale at the same time

mozzius (approved):
lgtm, tested all 3 sizes

src/components/Toast/Toast.tsx
estrattonbailey: Yeah this is def a hack for our font scaling, which I'd like to remove entirely. System font scaling is a whole 'nother beast. Honestly can't picture a solution that does both in a generic way, this is a pretty specific use case, but if you've got ideas would love to hear!

PR #8793, opened 2025-08-07
Fix policy overlay logic

estrattonbailey:
We needed to hoist up state so that it could be used in the `shell` to unmount the portal outlets. This meant prior `completed === true`, no dialogs would render. This broke signup and onboarding.

This PR hoists up another piece of state `isReadyToShowOverlay` that is `false` until we actually render the `PolicyUpdateOverlay` in `createNativeStackNavigatorWithAuth`, which only happens _after_ login/signup/onboarding etc.

### DIFF
--- a/src/components/PolicyUpdateOverlay/context.tsx
+++ b/src/components/PolicyUpdateOverlay/context.tsx
@@ -20,6 +20,11 @@ const Context = createContext<{
     completed: true,
     complete: () => {},
   },
+  /**
+   * Although our data will be ready to go when the app shell mounts, we don't
+   * want to show the overlay until we actually render it, which happens after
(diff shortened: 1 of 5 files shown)


src/components/PolicyUpdateOverlay/context.tsx
```diff
@@ -1,32 +1,71 @@
...
 }
 
 export function Provider({children}: {children?: ReactNode}) {
-  const state = usePolicyUpdateState()
+  const {hasSession} = useSession()
+  const [isReadyToShowOverlay, setIsReadyToShowOverlay] = useState(false)
+  console.log({
+    isReadyToShowOverlay,
+    hasSession,
+  })
```
mozzius: log

PR #8795, opened 2025-08-07
Profile back button tweaks for web

Simek:
# Why

Spotted that profile back button on web does not match position of other back buttons and lack the interaction effect.

https://github.com/user-attachments/assets/7d7b1fd0-31b4-4951-b432-83bd9fe1cc71

# How

Alter the profile back button position, align border radius with other back buttons, replace `TouchableWithoutFeedback` with `Button` component to be able to use hover state, rearrange and simplify styles.

# Preview

https://github.com/user-attachments/assets/a23b4aaf-9d1a-4da5-8da0-da120f2ff0be

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -26,6 +26,7 @@ import {LoadingPlaceholder} from '#/view/com/util/LoadingPlaceholder'
 import {UserAvatar} from '#/view/com/util/UserAvatar'
 import {UserBanner} from '#/view/com/util/UserBanner'
 import {atoms as a, platform, useTheme} from '#/alf'
+import {Button} from '#/components/Button'
 import {useDialogControl} from '#/components/Dialog'
 import {ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeftIcon} from '#/components/icons/Arrow'
 import {EditLiveDialog} from '#/components/live/EditLiveDialog'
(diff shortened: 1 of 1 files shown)


src/screens/Profile/Header/Shell.tsx
```diff
@@ -261,20 +271,12 @@ export {ProfileHeaderShell}
 const styles = StyleSheet.create({
   backBtnWrapper: {
     position: 'absolute',
-    left: 12,
-    width: 30,
-    height: 30,
-    overflow: 'hidden',
-    borderRadius: 15,
```
mozzius, line 268: While we're at it, can we remove the rest of the `StyleSheet` and just use ALF?

Simek: Sure! Is there a way to define size/position using ALF, or those styles need to remain as inline objects?

mozzius: Just keep those inline yeah

src/screens/Profile/Header/Shell.tsx
```diff
@@ -148,30 +150,46 @@ let ProfileHeaderShell = ({
...
-                        top: platform({
-                          web: 10,
-                          default: topInset,
-                        }),
+                        width: 31,
+                        height: 31,
+                        backgroundColor: transparentifyColor('#000', 0.5),
+                      },
+                      hovered && {
+                        backgroundColor: '#000',
```
mozzius:
```suggestion
                        backgroundColor: transparentifyColor('#000', 0.75),
```

PR #8797, opened 2025-08-07
Experiment: add backdrop blur to sticky headers on the web

Simek:
# Why

A small experiment involving adding backdrop blur for sticky headers on the web.

# How

Add `color` dependency to enable theme color modification, apply transformed background color and backdrop blur to the sticky headers on the web.

I wonder what do you think about this addition? I'm also happy to tweak chosen values (alpha and blur intensity) if you have any preference.

# Preview

https://github.com/user-attachments/assets/0374eedc-0123-4ab6-a7f6-cb79441a68d6

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "base64-js": "^1.5.1",
     "bcp-47": "^2.1.0",
     "bcp-47-match": "^2.0.3",
+    "color": "^5.0.0",
     "date-fns": "^2.30.0",
     "deprecated-react-native-prop-types": "^5.0.0",
     "email-validator": "^2.0.4",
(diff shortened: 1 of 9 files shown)


mozzius:
we have a `transparentifyColor` utility that should handle this btw

mozzius:
Also, I had a similar attempt here: https://github.com/bluesky-social/social-app/pull/7890

I think I prefer your approach though!

Simek:
Thanks for the suggestion and reference @mozzius! Removed the `color` dependency in favour of `transparentifyColor` util. Also applied the effect on Explore and Notifications screens.

Screenshot 2025-08-08 at 11 06 38

Screenshot 2025-08-08 at 11 03 13

mishl-dev:
waow there is already a pr for this

PR #8800, opened 2025-08-07
Update links

estrattonbailey:


pfrazee (approved):

surfdude29:
@estrattonbailey @pfrazee This is not about this PR (but it is sort of related), and the code isn't in social-app, but hopefully it's OK to ping you guys about this here:

I just noticed that the links to the Terms of Service and Privacy Policy given on the signup screen (which I think come from data fetched from the entryway?) are still linking to the `blueskyweb.xyz` versions, rather than to the `bsky.social` versions:

https://blueskyweb.xyz/support/tos
https://blueskyweb.xyz/support/privacy-policy

They just redirect to `bsky.social` so they still work of course, but given that the new domain launched back in early 2024 and seeing a different domain linked could potentially be confusing for some users, it might be good to update the links at some point :)

estrattonbailey:
@surfdude29 great call! I'll update those 👍

PR #8805, opened 2025-08-08
Add `english-only-resource` context and tweak labels

surfdude29:
This PR proposes adding an `english-only-resource` context to certain strings that link to a blog post, help article, etc which is – for now, at least – only available in English. I think this would provide useful information to translators so that, if they want, they can let the user know that the resource they're about to access has not been localized and is only available in English.

For example, in the French localization, [we have been adding `(en anglais)`](https://bluesky.crowdin.com/editor/1/all/en-fr/7#q=anglais) (meaning _in English_) to the end of translations where we feel it's helpful to let the user know that the blog post or article being linked to is only available in English. Having the `english-only-resource` context on relevant strings would make this easier.

To avoid excessive string churn, this PR only touches some of the relevant strings. If feedback from translators is positive that this is useful and outweighs any inconvenience from string churn, then a future PR could add `english-only-resource` context to other relevant strings.

If there is another context that would be clearer, then I'm open to changing it to something that works better. Here are a few other possible options:
- `english-only`
- `non-localized-resource`
- `not-localized`
- `untranslated-resource`

Also included in this PR are a couple of very minor string tweaks to reduce duplication:
- Removing a `.` from the end of a label on L218 in `src/components/moderation/ContentHider.tsx` to align it with the `accessibilityHint` on L152
- Changing `learn more` ➡️ `Learn more` on L271 of `src/screens/Settings/components/ChangeHandleDialog.tsx`

### DIFF
--- a/src/components/moderation/ContentHider.tsx
+++ b/src/components/moderation/ContentHider.tsx
@@ -215,7 +215,7 @@ function ContentHiderActive({
             control.open()
           }}
           label={_(
-            msg`Learn more about the moderation applied to this content.`,
+            msg`Learn more about the moderation applied to this content`,
           )}
           style={[a.pt_sm]}>
           {state => (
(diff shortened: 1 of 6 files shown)


mozzius:
Great idea!

estrattonbailey (approved):
Sure, yeah I like this idea. We will try to remember to add this in the future 👍

PR #8766, opened 2025-08-01
Update starter pack wizard screen

BinaryFiddler:
Change log:
- Update the layout changes in starter pack creation screens

### "Choose people" page
 - Edit button is moved to the top right
 - The "Next" button takes the full width
 - Profile pictures of selected people now overlap

### "Choose feeds" page
 - Edit button is moved to the top right
 - The "Next" button takes the full width

### What's not in this PR:
- System wide styling updates for checkboxes and buttons 
- New modals for the starter pack flow initiated from the profile menu options

simulator_screenshot_4A6A99C4-DB95-48D0-A274-26D05589AB72

simulator_screenshot_A7650E91-3A48-465B-8C39-0BE4D563D695

simulator_screenshot_25E9C68D-77BE-46F0-8AA7-3BFBE623DFF9

PR #8806, opened 2025-08-08
Binaryfiddler/starter pack part2

BinaryFiddler:
Follow up to https://github.com/bluesky-social/social-app/pull/8766

- The rest of the starter pack work flow
- The initial selected profile in a new starter pack is now the profile the user is viewing instead of always being the currentUser's 
- Omit the success Toast per Samuel's suggestion since we don't support Toast at the bottom for now

Known todos: (still working on it)
- The dialog sometimes shows the Empty view during the initial load
- Remove the Toast.show when profiles are added/removed from starter packs


jimmylee:
Some conflicts were picked up here

`src/components/StarterPack/Wizard/WizardEditListDialog.tsx`
`src/screens/StarterPack/Wizard/index.tsx`

estrattonbailey (commented):
Haven't looked deeply at the code yet, but did a cursory test and seems to work well! Couple initial notes:
- Some missing/incorrect aspects compared to the Figma design
- Needs some design finesse on desktop especially

Sam and I can do a deeper look with fresh eyes in the morning. Thanks!

mozzius:
Yeah biggest issue is the nested scrollviews. Adding a scrollview within the dialog is really janky, see here:

https://github.com/user-attachments/assets/839aefc8-860c-4816-85f6-da18f971d21b

We need to use `Dialog.InnerFlatList` instead where the entire dialog is scrollable, with a header at the top - see the gif select dialog for an example

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,388 @@
...
+                    style={[
+                      {
+                        zIndex: 1 - index,
+                        marginLeft: index > 0 ? -2 : 0,
+                        borderWidth: 0.5,
+                        borderColor: t.atoms.bg.backgroundColor,
+                      },
+                    ]}
+                  />
+                ))}
```
mozzius: You can just use the `AvatarStack` component here I think!

BinaryFiddler: The AvatarStack has a different overlap and, and borderWidth (it overlaps 1/3 of the icon, and has a border width of 1). Is that ok?

BinaryFiddler: The AvartarStack vs desgin: 
Screenshot 2025-08-20 at 11 59 14 AM

Screenshot 2025-08-20 at 12 00 05 PM

BinaryFiddler:
Refactored a bit to use the `Dialog.InnerFlatList` inside the dialog instead of the base FlatList

src/components/dialogs/StarterPackDialog.tsx
mozzius: Oh good catch - I say let's just use AvatarStack for consistency

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+      </View>
+
+      <View style={[a.align_center]}>
+        <Button
+          label={_(msg`Create starter pack`)}
+          color="secondary_inverted"
+          size="small"
+          onPress={onStartWizard}>
+          <ButtonText>
+            <Trans>Create</Trans>
```
surfdude29:
```suggestion
            <Trans comment="Text on button to create a new starter pack">
              Create
            </Trans>
```
Since this is a single word and is afaik only used here in the UI, I think a comment for future translators might be helpful :)

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+            <Text style={[a.text_md, a.font_bold]}>
+              <Trans>New starter pack</Trans>
+            </Text>
+            <Button
+              label={_(msg`Create starter pack`)}
+              color="secondary_inverted"
+              size="small"
+              onPress={onStartWizard}>
+              <ButtonText>
+                <Trans>Create</Trans>
```
surfdude29:
```suggestion
                <Trans comment="Text on button to create a new starter pack">
                  Create
                </Trans>
```
And would similarly be good here too :)

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+
+                {starterPack.list?.listItemCount &&
+                  starterPack.list.listItemCount > 4 && (
+                    <Text
+                      style={[
+                        a.text_sm,
+                        t.atoms.text_contrast_medium,
+                        a.ml_xs,
+                      ]}>
+                      {`+${starterPack.list.listItemCount - 4} more`}
```
surfdude29: Just checking, would this string be translatable? 🤔

src/screens/StarterPack/Wizard/index.tsx
```diff
@@ -472,12 +484,23 @@ function Footer({
             {
               items.length < 2 ? (
                 <Trans>
-                  It's just you right now! Add more people to your starter pack
-                  by searching above.
+                  It's just{' '}
+                  <Text style={[a.font_bold, textStyles]} emoji>
+                    {currentAccount?.did === items[0].did
+                      ? 'you'
+                      : getName(items[0])}{' '}
```
surfdude29: IIUC, feedback from translators is that this type of concatenated string is less than ideal as it can be very difficult to translate in some languages (e.g. see #7827). Would it be possible to have separate strings for each case? 🙏

src/screens/StarterPack/Wizard/index.tsx
```diff
@@ -472,12 +484,23 @@ function Footer({
...
+                  right now! Add more people to your starter pack by searching
+                  above.
                 </Trans>
               ) : items.length === 2 ? (
                 <Trans>
-                  <Text style={[a.font_bold, textStyles]}>You</Text> and
+                  <Text style={[a.font_bold, textStyles]}>
+                    {currentAccount?.did === items[0].did
+                      ? 'you'
+                      : getName(items[0])}
```
surfdude29: And similarly here

src/components/dialogs/StarterPackDialog.tsx
mozzius: It should probably be a `<Plural>`

src/components/dialogs/StarterPackDialog.tsx
BinaryFiddler: sounds good!

src/components/dialogs/StarterPackDialog.tsx
BinaryFiddler: Sounds good

src/screens/StarterPack/Wizard/index.tsx
surfdude29: Perfect, thanks @BinaryFiddler! :)

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,394 @@
...
+          a.my_lg,
+          a.align_center,
+        ]}>
+        <Text style={[a.text_lg, a.font_bold]}>
+          <Trans>Add to starter packs</Trans>
+        </Text>
+        <TimesLarge_Stroke2_Corner0_Rounded
+          onPress={onClose}
+          fill={t.atoms.text_contrast_medium.color}
+        />
```
mozzius: Need to use a `<Button>` here.

```tsx
<Button label={_(msg`Close`)} onPress={onClose}>
  <ButtonIcon icon={XIcon} />
</Button>
```

Also, it's nicer to `import {TimesLarge_Stroke2_Corner0_Rounded as XIcon}`

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,394 @@
...
+      <>
+        <Dialog.Close />
+        <Dialog.ScrollableInner
+          contentContainerStyle={[a.align_center, a.p_xl]}
+          label={_(msg`Add to starter packs`)}>
+          <Loader size="xl" />
+        </Dialog.ScrollableInner>
+      </>
+    )
+  }
```
mozzius: Need to give this a min-height, otherwise the sheet is way too small initially. Also, I would probably prefer to not conditionally render a different component, and instead just always render the FlatList - if it's loading, make it render only a single item (the spinner)

https://github.com/user-attachments/assets/42fcb217-d973-48df-b2d8-eb744934944a

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+  return (
+    <>
+      <Dialog.Close />
```
mozzius: Remove this extra close button - `<Dialog.Close />` is web only and you already have one in the header

Screenshot 2025-08-21 at 12 34 02

src/components/dialogs/StarterPackDialog.tsx
```diff
@@ -0,0 +1,394 @@
...
+      <Dialog.InnerFlatList
+        data={membershipItems}
+        renderItem={renderItem}
+        keyExtractor={(item: StarterPackWithMembership, index: number) =>
+          item.starterPack.uri || index.toString()
+        }
```
mozzius: will `item.starterPack.uri` ever be falsy? I'm not sure about giving the index as a key

mozzius (commented):
Looking *much* improved on native, nice work. The padding on web needs some love however

Screenshot 2025-08-21 at 12 29 59

src/components/dialogs/StarterPackDialog.tsx
BinaryFiddler: Sounds good

src/components/dialogs/StarterPackDialog.tsx
BinaryFiddler: good catch, `item.starterPack.uri ` should be sufficient and can never be falsy

src/components/dialogs/StarterPackDialog.tsx
BinaryFiddler: Removed this one, and followed the advice to render a single loader component in the list.

BinaryFiddler:
> Looking _much_ improved on native, nice work. The padding on web needs some love however - this is my fault haha, the InnerFlatList doesn't really work out of the box, something I need to fix
> Screenshot 2025-08-21 at 12 29 59
Screenshot 2025-08-21 at 5 52 53 PM

Updated for the web. Also fixed an issue where the dialog doesn't open back up no the web upon successful creation of new starter pack

BinaryFiddler:
I noticed in figma, the web version is supposed to wrap the entire wizard flow into the dialog. It seems like a pretty sizable change on how the wizard works. I left a comment in figma, thinking it's probably better to defer it in a separate ticket. Would love to know your thoughts too.

https://www.figma.com/design/TqFguaWhkvAPnCth5u7cvE?node-id=3822-29734&m=dev#1393538162

src/screens/StarterPack/Wizard/index.tsx
```diff
@@ -260,6 +283,7 @@ function WizardInner({
         currentListItems: currentListItems,
       })
     } else {
+      console.log('Creating new starter pack: ', state.profiles)
```
surfdude29: Just checking, should this log be here? 🤔

BinaryFiddler: Good catch, out here during testing. Going to remove it

jimmylee (approved):
LGTM, nice one @BinaryFiddler !!

PR #8810, opened 2025-08-10
Add context and comments to button text, label and title in email verification dialog

surfdude29:
I think it might be helpful to translators to make it a bit clearer that the `Verify code` button text and accessibility label in the email dialog – when the user is entering a code to verify their account email address – represents an action.

Therefore, this PR proposes adding `action` context and a comment for translators to the `Verify code` string.

The suggested comment is:

`Button text and accessibility label for action to verify the user's email address using the code entered`

This is a bit wordy but hopefully will make it very clear to translators what the string means.

___

**EDIT:** Also, I think the `Verify email code` title in the dialog might be difficult for some translators to parse and translate without context as to when it's shown. Therefore, I suggest also adding this comment for translators to that string:

`Dialog title when a user is verifying their email address by entering a code they have been sent`

IMG_7361


estrattonbailey (approved):
Great idea, thanks!

surfdude29:
Oops, not sure why the lockfile check is failing, should I sync it to main?

edit: hopefully should be sorted now 👍

PR #8815, opened 2025-08-11
[Perf - part 3] Stop every dialog control in the entire app rerendering when opening a dialog

mozzius:
**This PR is not stacked on anything**

This is the primary driver of the composer being so slow to open. Because we'd stuffed `fullyExpandedCount` into the dialog control context, every consumer of the context (i.e. every single `useDialogControl`) would rerender when `fullyExpandedCount` changed, such as when the composer opens on iOS, or when a full-height dialog opens.

# Before

On the profiler the commits are the following:

Commits 1 & 2: debug toast, ignore

Commit 3: actually rendering the composer. This takes ~30ms

Commit 4: by far the largest one, this is updating the `fullyExpandedCount` from 0 -> 1. This causes a rerender of every component with a dialog control, of which there are multiple in every post. In the simulator this takes 180ms.

Screenshot 2025-08-11 at 15 59 52

# After

There is only one spike, which is the composer rendering. Note the long tail is fixed by #8813 

Screenshot 2025-08-11 at 15 55 52

# Solution

Just move `fullyExpandedCount` to a separate context. This then only triggers a rerender of the component that needs it, which is only 1 place (the shell).

# Test plan

Confirm that the profiler no longer shows everything rerendering when opening the composer. Confirm that the status bar still is the correct color when opening the sheet to max.

### DIFF
--- a/src/state/dialogs/index.tsx
+++ b/src/state/dialogs/index.tsx
@@ -25,25 +25,18 @@ interface IDialogControlContext {
   setFullyExpandedCount: React.Dispatch<React.SetStateAction<number>>
 }
 
-interface IDialogFullyExpandedContext {
-  /**
-   * The number of dialogs that are fully expanded. This is used to determine the background color of the status bar
-   * on iOS.
-   */
-  fullyExpandedCount: number
-}
-
 const DialogContext = React.createContext<IDialogContext>({} as IDialogContext)
 
 const DialogControlContext = React.createContext<IDialogControlContext>(
   {} as IDialogControlContext,
 )
 
-const DialogFullyExpandedContext =
-  React.createContext<IDialogFullyExpandedContext>({
-    fullyExpandedCount: 0,
-  } as IDialogFullyExpandedContext)
-DialogFullyExpandedContext.displayName = 'DialogFullyExpandedContext'
+/**
+ * The number of dialogs that are fully expanded. This is used to determine the background color of the status bar
+ * on iOS.
+ */
+const DialogFullyExpandedCountContext = React.createContext<number>(0)
+DialogFullyExpandedCountContext.displayName = 'DialogFullyExpandedCountContext'
 
 export function useDialogStateContext() {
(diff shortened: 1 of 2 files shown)


src/state/dialogs/index.tsx
```diff
@@ -22,12 +22,15 @@ interface IDialogContext {
 interface IDialogControlContext {
   closeAllDialogs(): boolean
   setDialogIsOpen(id: string, isOpen: boolean): void
+  setFullyExpandedCount: React.Dispatch<React.SetStateAction<number>>
+}
+
+interface IDialogFullyExpandedContext {
   /**
-   * The number of dialogs that are fully expanded. This is used to determine the backgground color of the status bar
+   * The number of dialogs that are fully expanded. This is used to determine the background color of the status bar
```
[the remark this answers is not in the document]
mozzius: Yes, you're commenting on the line where I fixed the typo. Are you for real

src/state/dialogs/index.tsx
```diff
@@ -85,21 +98,24 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+    [closeAllDialogs, setDialogIsOpen, setFullyExpandedCount],
+  )
+
+  const fullyExpanded = React.useMemo(
+    () => ({
       fullyExpandedCount,
-      setFullyExpandedCount,
-    ],
+    }),
+    [fullyExpandedCount],
```
TkDodo: wouldn’t it make more sense to just pass the `fullyExpandedCount` number directly to the context as a value now, without wrapping it in an object? I guess a context named `DialogFullyExpandedContext` will not / should not contain any more information other than that `fullyExpandedCount`, and we wouldn’t need to memoize the value if it were just a number because that’s always referentially stable.

mozzius: You know what, that's a good call - we *really don't want anything in here*, since that's what lead to the perf problem to begin with

estrattonbailey (approved):

PR #8821, opened 2025-08-11
Fix policy state

estrattonbailey:
There seems to be a subtle race condition when marking `completedForDevice = true` _after_ account creation, due to the app re-rendering with the new account in place.

Really, these policies can be marked as complete after the first step of the create account flow, which removes any chance of the race condition when the app re-renders after the account is created.


mozzius (approved):
Terrible horrible disgusting gross but sure lgtm

Will dig into the root cause tomorrow

PR #8825, opened 2025-08-12
Preconnect to go.bsky.app

mozzius:
Might make the link proxy service a little bit faster?

Also, should we use `preconnect` or `dns-prefetch` here? I don't know

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -8,6 +8,7 @@
     Preconnect to essential domains
   -->
   <link rel="preconnect" href="https://bsky.social">
+  <link rel="preconnect" href="https://go.bsky.app">
   <title>{%- block head_title -%}Bluesky{%- endblock -%}</title>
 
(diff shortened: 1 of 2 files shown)


surfdude29:
I'm obviously no expert on these matters, but my understanding from reading things like [this StackOverflow answer](https://stackoverflow.com/a/51059127) is that `preconnect` might be more helpful than `dns-prefetch` here, because it would also set up the TLS connection to `go.bsky.app`, which could be pretty useful in speeding up loading links.

However, would `preconnect` cause a lot of extra requests to `go.bsky.app` that end up being unnecessary, because the user doesn't click a link? That sounds like it would be the main consideration 🤔

Also, would it help to add either `preconnect` or `dns-prefetch` here as well?

https://github.com/bluesky-social/social-app/blob/683d7bf5a7b34e925a4976f14b7dcee5804855fd/web/index.html#L14-L18

PR #8826, opened 2025-08-12
Remove unicode LTR chars and replace with CSS on profile on web

mozzius:
Currently copy pasting a handle from a profile doesn't work because of the extra unicode control chars we jam in there.

We can replace this with CSS. gpt5 tells me the most direct equivalent is:

```
direction: 'ltr',
unicodeBidi: 'isolate',
```

I only did this for the profile handle to minimise impact

# Test plan

Confirm you can copy paste handles without getting extra chars
Confirm the LTR behaviour is the same

### DIFF
--- a/src/lib/strings/handles.ts
+++ b/src/lib/strings/handles.ts
@@ -25,10 +25,17 @@ export function isInvalidHandle(handle: string): boolean {
   return handle === 'handle.invalid'
 }
 
-export function sanitizeHandle(handle: string, prefix = ''): string {
+export function sanitizeHandle(
+  handle: string,
+  prefix = '',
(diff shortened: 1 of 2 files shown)


DavidBuchanan314:
I can confirm that copy-pasting handles from profiles works properly now.

"Confirm the LTR behaviour is the same" is a little tricky because I don't know the full set of scenarios that would need testing.

I've tested viewing an account with an Arabic display name in desktop Firefox, and desktop Chromium with browser language set to Arabic - for both of those, the layout and direction of all text is the same as before.

DavidBuchanan314 (approved):

PR #8710, opened 2025-07-24
[APP-1318] Post composer: combine image & video buttons 

APiligrim:
This PR addresses this linear ticket about combining image and video into one button and reusing an image icon. These are the cases that are supported:

**To Do:**

- [x] Combine image/video buttons into a single button, using the image icon
- [x] Allow selecting both images and videos using APIs supported by the picker library
- [x] Limit selectionLimit: 4 using APIs supported by the picker library

Validate selection. Determine how many assets were selected by the user and:
  - [x] If images, and more than 4 images were selected, alert user with a Toast.show that they can only upload 4 images at a time right now
  - [x] If videos, and more than 1 video selected, toast the user and tell them only one video at a time

If a mix of videos/images:

- [x] If video was selected first, take the video and discard the selected images
- [x] If image(s) were selected and a video selected later, discard video and take up to 4 images
- [x] In both cases, toast user that they cannot upload a mix of images and videos at this time

Handle on Web
- [x] Make sure the user is able to select video files on web (currently it is not allowed so the user needs to use a GIF)

## Browser Testing:

- [x] If images, and more than 4 images were selected, alert user with a Toast.show that they can only upload 4 images at a time right now

https://github.com/user-attachments/assets/dc949465-6f93-4f14-bbd6-bf534d52f838

- [x] If videos, and more than 1 video selected, toast the user and tell them only one video at a time

https://github.com/user-attachments/assets/8c7145ed-8019-455a-ab01-4a7519cfdbbe

**Selecting images and videos at the same time** 

- [x] If video was selected first, take the video and discard the selected images & toast the user

https://github.com/user-attachments/assets/be0b8437-a669-4ba0-ad93-91ac6fbc199f

- [x] If image(s) were selected and a video selected later, discard video and take up to 4 images & toast the user

https://github.com/user-attachments/assets/8e394178-9088-4329-9ffa-98efdd0afd56

## IOS Testing

- [x]  No more than 4 items are allowed to be selected, showing if 4 images are picked , there is a limit that is set

https://github.com/user-attachments/assets/d62ec176-6047-4bd5-9d4c-3c60e6e179a0

- [x] If image(s) were selected and a video selected later, discard video and take up to 4 images & toast the user

https://github.com/user-attachments/assets/664c6526-76fb-4437-8bd1-89541d09ae86

- [x] If video was selected first, take the video and discard the selected images 

https://github.com/user-attachments/assets/cc8d9b0f-e316-45ed-a1e2-3cd6e7b25bb8

[...]

PR #8828, opened 2025-08-12
[APP-1318] `SelectMediaButton`

estrattonbailey:
Stacked on #8839

Combines the image/video buttons into a single button. Provides the following validations:
- limits selection of images to 4
  - on iOS we can do this in the image picker itself
  - on Android, we do this validation manually
- limits selection of videos and GIFs to 1
  - provides separate errors for video and GIF files
- disallows videos that are longer than 3 minutes
- prevents selecting a mix of media types
  - uses whatever selected item is returned first as the "type" to constrain against
    - example: select 5 images and a video, we'll only take the first 4 images
  - fixes a prod bug that allows you to select an image first, then a GIF (which will be rasterized)

### DIFF
--- a/src/view/com/composer/SelectMediaButton.tsx
+++ b/src/view/com/composer/SelectMediaButton.tsx
@@ -410,7 +410,7 @@ export function SelectMediaButton({
             msg`You can only select one GIF at a time.`,
           ),
           [SelectedAssetError.FileTooBig]: _(
-            msg`One or more of your selected files is too large. Maximum size is 100MB.`,
+            msg`One or more of your selected files is too large. Maximum size is 100 MB.`,
           ),
         }[error]
       })


src/view/com/composer/Composer.tsx
```diff
@@ -514,13 +522,6 @@ export const ComposePost = ({
       onPostSuccess?.(postSuccessData)
     }
     onClose()
-    Toast.show(
-      thread.posts.length > 1
-        ? _(msg`Your posts have been published`)
-        : replyTo
-          ? _(msg`Your reply has been published`)
-          : _(msg`Your post has been published`),
-    )
```
surfdude29: I think these toasts give a useful confirmation to the user, just wondering why they're being removed? 🤔

src/view/com/composer/Composer.tsx
```diff
@@ -811,7 +816,9 @@ let ComposerPost = React.memo(function ComposerPost({
         if (isNative) return // web only
         const [mimeType] = uri.slice('data:'.length).split(';')
         if (!SUPPORTED_MIME_TYPES.includes(mimeType as SupportedMimeTypes)) {
-          Toast.show(_(msg`Unsupported video type`), 'xmark')
+          toast.show(_(msg`Unsupported video type`), {
```
surfdude29: The current error toast gives the mime type of the video in the error message, would that be useful to include here?
https://github.com/bluesky-social/social-app/blob/f001af2fe3ca977beb22fefefb2f98ba2c076d49/src/view/com/composer/videos/SelectVideoBtn.tsx#L48

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,501 @@
...
+
+    if (typeof selectedVideo.duration !== 'number') {
+      try {
+        const metadata = await getAdditionalVideoMetadata(selectedVideo)
+        selectedVideo.duration = metadata.duration
+        selectedVideo.width = metadata.width
+        selectedVideo.height = metadata.height
+      } catch (e: any) {
+        logger.error(`processSelectedAssets: failed to get video metadata`, {
+          safeMessage: e.message,
```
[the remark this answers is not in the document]
mozzius: ok so replace the type assertion with a different, entirely the same kind of assertion.

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,501 @@
...
+  } else if (mimeType?.startsWith('image/')) {
+    type = 'image'
+  }
+
+  // console.log({
+  //   asset,
+  //   type,
+  //   mimeType,
+  // })
+
```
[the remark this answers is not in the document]
mozzius: What

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,501 @@
...
+      const errors = Array.from(errorCodes).map(error => {
+        return {
+          [SelectedAssetError.Unsupported]: _(
+            msg`One or more of your selected files are not supported.`,
+          ),
+          [SelectedAssetError.MixedTypes]: _(
+            msg`Selecting multiple media types is not supported.`,
+          ),
+          [SelectedAssetError.MaxImages]: _(
+            msg`You can select up to ${MAX_IMAGES} total images.`,
```
surfdude29:
```suggestion
            msg`You can select up to 4 total images.`,
```

We currently hardcode the maximum number of images in this string (I'm not clear on when it's shown):
https://github.com/bluesky-social/social-app/blob/f001af2fe3ca977beb22fefefb2f98ba2c076d49/src/lib/media/picker.shared.ts#L25
Would it make sense to do that here as well? (And if so, the other string could perhaps be aligned with this one to avoid duplication.)

src/view/com/composer/text-input/TextInput.tsx
```diff
@@ -154,6 +154,7 @@ export const TextInput = forwardRef(function TextInputImpl(
       }
 
       const uris = files.map(f => f.uri)
+      // TODO
```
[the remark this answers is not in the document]
estrattonbailey: Oh thanks copilot, mean to rm this

src/view/com/composer/SelectMediaButton.tsx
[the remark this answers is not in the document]
estrattonbailey: nah

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,501 @@
...
+    }
+  } else if (selectableAssetType === 'gif') {
+    if (supportedAssets.length > 1) {
+      errors.add(SelectedAssetError.MaxGIFs)
+      supportedAssets = supportedAssets.slice(0, 1)
+    }
+  }
+
+  return {
+    type: selectableAssetType!, // set above
```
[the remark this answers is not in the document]
estrattonbailey: validated above, as comment states

src/view/com/composer/SelectMediaButton.tsx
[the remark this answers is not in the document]
estrattonbailey: yep, debug code, will remove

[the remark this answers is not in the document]
mozzius: Oh I think the app isn't showing me the diff properly

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+type ValidatedImagePickerAsset = Omit<ImagePickerAsset, 'mimeType'> & {
+  mimeType: string
+}
+
+/**
+ * Codes for known validation states
+ */
+enum SelectedAssetError {
+  Unsupported = 'Unsupported',
+  MixedTypes = 'MixedTypes',
```
APiligrim, line 56: this is a good idea! Makes it a lot easier to track / add on new edge cases

mozzius:
@github-actions ota

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+  gif: 'image/gif',
+  jpg: 'image/jpeg',
+  jpeg: 'image/jpeg',
+  png: 'image/png',
+  svg: 'image/svg+xml',
+  heic: 'image/heic',
+}
+
+/**
+ * Attemps to bucket the given asset into one of our known types based on its
```
APiligrim: thank you for these comments! They are super helpful!

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+    sheetWrapper,
+    processSelectedAssets,
+    selectionCountRemaining,
+  ])
+
+  return (
+    <Button
+      testID="openMediaBtn"
+      onPress={onPressSelectMedia}
+      label={_(msg`Media`)}
```
surfdude29:
```suggestion
      label={_(
        msg({
          message: `Add media to post`,
          comment: `Accessibility label for button in composer to add photos or a video to a post`,
        }),
      )}
```
Would a slightly longer label for the button be helpful here? (I'm genuinely not sure how much info it's good to include for screen reader users in a button label like this.)

Also, a comment for translators might be helpful (especially if the label is just `Media`, but perhaps also if it's more descriptive.)

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+        errors: errorCodes,
+      } = await processImagePickerAssets(rawAssets, {
+        selectionCountRemaining,
+        allowedAssetTypes,
+      })
+
+      /*
+       * Convert error codes to user-friendly messages.
+       */
+      const errors = Array.from(errorCodes).map(error => {
```
APiligrim, line 387: would these errors show Toast messages to the user to say why their files failed to upload? 👀

APiligrim (commented):
the written types for edge cases and separated functions for each of the processes look great and make the code readable too. Thank you for all the detailed explanations!

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+  selectedAssetsCount,
+  onSelectAssets,
+}: SelectMediaButtonProps) {
+  const {_} = useLingui()
+  const {requestPhotoAccessIfNeeded} = usePhotoLibraryPermission()
+  const {requestVideoAccessIfNeeded} = useVideoLibraryPermission()
+  const sheetWrapper = useSheetWrapper()
+  const t = useTheme()
+
+  const selectionCountRemaining = MAX_IMAGES - selectedAssetsCount
```
APiligrim, line 371: I like that MAX_IMAGES is a variable in cases later this variable gets updated, we only need to update it once

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,495 @@
...
+ * Codes for known validation states
+ */
+enum SelectedAssetError {
+  Unsupported = 'Unsupported',
+  MixedTypes = 'MixedTypes',
+  MaxImages = 'MaxImages',
+  MaxVideos = 'MaxVideos',
+  VideoTooLong = 'VideoTooLong',
+  MaxGIFs = 'MaxGIFs',
+  NoGifsOnNative = 'NoGifsOnNative',
```
APiligrim: for `NoGifsOnNative` are we not going to support GIFs for now at all for native? Just wondering why 👀

APiligrim: Also we are still keeping the GIF button on native then right?

src/view/com/composer/Composer.tsx
estrattonbailey: Ooof, thanks, noticed this in #8710 and totally forgot to add back here, thanks!

src/view/com/composer/SelectMediaButton.tsx
surfdude29:
```suggestion
 * Attempts to bucket the given asset into one of our known types based on its
```
micronit

src/view/com/composer/Composer.tsx
estrattonbailey: Ah good idea!

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: Good question! The GIF button is actually for selecting a GIF from Tenor. On web, we allow you to upload a `.gif` file, which we then convert to a video before posting. We haven't implemented this logic yet on native, that's what this error is for.

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: Uploading happens in a later step, these errors only pertain to the selection of assets.

src/view/com/composer/SelectMediaButton.tsx
mozzius: Is that an error though? Currently they're just handled as images (because they are also valid images)

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: Oh ya know what, this is just old sloppy logic that we don't need anymore, thanks for mentioning this, I'm gonna just clean this up 👍

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: Yes that's perfect, thanks!

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: Oh this is true, maybe this is too heavy handed. No reason we couldn't also toast the user to tell them it will upload as a static image 🤔

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,506 @@
...
+      const errors = Array.from(errorCodes).map(error => {
+        return {
+          [SelectedAssetError.Unsupported]: _(
+            msg`One or more of your selected files are not supported.`,
+          ),
+          [SelectedAssetError.MixedTypes]: _(
+            msg`Selecting multiple media types is not supported.`,
+          ),
+          [SelectedAssetError.MaxImages]: _(
+            msg`You can select up to ${MAX_IMAGES} total images.`,
```
surfdude29:
```suggestion
          [SelectedAssetError.MaxImages]: _(
            msg({
              message: `You can select up to ${plural(MAX_IMAGES, {
                other: '# images',
              })} in total.`,
              comment: `Error message for maximum number of images that can be selected to add to a post, currently 4 but may change.`,
            }),
          ),
```
Would it be possible to add plural formatting and a comment for translators here please, for languages with complex plurals? I also suggested a slight wording tweak to `… ${MAX_IMAGES} images in total.`

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,506 @@
...
+      )}
+      accessibilityHint={
+        isNative
+          ? _(
+              msg`Opens device gallery to select up to ${MAX_IMAGES} images, or a single video.`,
+            )
+          : _(
+              msg`Opens device gallery to select up to ${MAX_IMAGES} images, or a single video or GIF.`,
+            )
+      }
```
surfdude29:
```suggestion
      accessibilityHint={
        isNative
          ? _(
              msg({
                message: `Opens device gallery to select up to ${plural(
                  MAX_IMAGES,
                  {
                    other: '# images',
                  },
                )}, or a single video.`,
                comment: `Accessibility hint on native for button in composer to add images or a video to a post. Maximum number of images that can be selected is currently 4 but may change.`,
              }),
            )
          : _(
              msg({
                message: `Opens device gallery to select up to ${plural(
                  MAX_IMAGES,
                  {
                    other: '# images',
                  },
                )}, or a single video or GIF.`,
                comment: `Accessibility hint on web for button in composer to add images, a video, or a GIF to a post. Maximum number of images that can be selected is currently 4 but may change.`,
              }),
            )
      }
```
And if plural formatting and comments could be added here too that would be great 🙏

src/view/com/composer/SelectMediaButton.tsx
mozzius: Decided to just support gifs, see below

mozzius:
Regarding `NoGifsOnNative` - screw it, let's just support gifs on native

https://github.com/user-attachments/assets/5ea732a5-c50f-42f1-be6d-f996d92e3c6c

Cavaets:

- it has to skip the compression step, because the library does not support compressing gifs. however, I think most gifs are gonna be below 100mb (!!) so that's fine. we can just let the server reject them. basically works like web does.
- had to implement a custom preview (also fine)

src/view/com/composer/Composer.tsx
```diff
@@ -650,84 +659,88 @@ export const ComposePost = ({
...
-                  isPartOfThread={thread.posts.length > 1}
-                  isReply={index > 0 || !!replyTo}
-                  isActive={post.id === activePost.id}
-                  canRemovePost={thread.posts.length > 1}
-                  canRemoveQuote={index > 0 || !initQuote}
-                  onSelectVideo={selectVideo}
-                  onClearVideo={clearVideo}
-                  onPublish={onComposerPostPublish}
-                  onError={setError}
+      <RootSiblingParent>
```
mozzius, line 662: Is this still necessary with Sonner toasts?

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,500 @@
...
+      style={a.p_sm}
+      variant="ghost"
+      shape="round"
+      color="primary"
+      disabled={disabled}>
+      <ImageIcon
+        size="lg"
+        style={disabled && t.atoms.text_contrast_low}
+        accessibilityIgnoresInvertColors={true}
+      />
```
mozzius, line 521: Do we want a new icon here?

mozzius (changes requested):
Looking pretty solid! Only one noteworthy bug - can't seem to select additional images after selecting some

  

https://github.com/user-attachments/assets/3aef42f1-264b-45d2-9fb1-10d3d017c1fb

surfdude29:
I did some testing on the preview branch and it's working well, very nice work @APiligrim, @estrattonbailey and @mozzius  👌

One thing I noticed is that too large videos on web are treated differently – not sure about native, I know everything over 25 MB is compressed there anyway, so maybe it's not an issue – and I think you might want to look again at that.

In prod, when you select a video bigger than 100 MB, you get a specific error message that tells you exactly what the problem is (I think [from here](https://github.com/bluesky-social/social-app/blob/221623f55aa6c1bbe699c8d409832da110923c76/src/view/com/composer/state/video.ts#L395)? because the other version looks like an upload error).

However, when you select a video bigger than 100 MB on the preview branch, you only get the generic `Unsupported` error toast. Tbh this feels quite a lot less useful in understanding why the video can't be added to the post, so if it's possible to show the `VideoTooLargeError` string instead, I think that would be ideal.

| error shown in prod | error shown on preview branch |
|--------|--------|
| IMG_8093 | IMG_8094 |

___

Also, on web, now that some video errors are presented as toasts popping up in the lower-left corner of the screen rather than being shown directly in the composer, I think it might be worth considering using the longer, 6-second toast, to give users more of a chance to spot it and read it.

src/view/com/composer/SelectMediaButton.tsx
estrattonbailey: I think the image is fine, did you have an idea?

src/view/com/composer/Composer.tsx
estrattonbailey: Ah no, thanks

src/view/com/composer/SelectMediaButton.tsx
mozzius: Looking at our options, there's this one? I think I'm fine with photo but I don't know if it'll be confusing for users if it looks like the video button has disappeared

mozzius: There's also the Cat Videos icon lmao

Screenshot 2025-08-14 at 18 07 11

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,522 @@
...
+          ),
+          [SelectedAssetError.MaxImages]: _(
+            msg({
+              message: `You can select up to ${plural(MAX_IMAGES, {
+                other: '# images',
+              })} in total.`,
+              comment: `Error message for maximum number of images that can be selected to add to a post, currently 4 but may change.`,
+            }),
+          ),
+          ),
```
surfdude29:
```suggestion
```
Oops, I added a duplicate line at the end here in my suggestion, sorry 😬

estrattonbailey: Np at all!

estrattonbailey:
@surfdude29 when you were testing with a large video on web, what file format was it? If I try to upload a 150mb `.mp4` I get the correct error. So I'm wondering if you tried to upload a video that this branch considers unsupported, but that prod supports.

surfdude29:
@estrattonbailey Ah interesting, I'll do a bit more testing and report back if I can reproduce, and if so with what formats 👍

surfdude29:
@estrattonbailey OK sorry for the delay in getting back to you, I tried several times using Big Buck Bunny as the test file ([from here](https://download.blender.org/peach/bigbuckbunny_movies/)).

I should mention btw that I'm testing in Safari on iPadOS 18.6, so it's possible there's some iOS weirdness going on that makes what I'm seeing an edge case?

Anyway, using the raw unmodified file (`big_buck_bunny_1080p_h264.mov`):
9 minutes 56 seconds long, 725 MB

**On preview branch:** gets `Unsupported` error
**In prod:** gets `Video must be less than 3 minutes long` error

I then took the same file and trimmed it in iPadOS Photos app (no re-encoding):
1 minute 13 seconds long, 110 MB ([can be downloaded here](https://transfer.it/t/6KlPX84hGBRB)):

**On preview branch:** gets `Unsupported` error
**In prod:** gets `VideoTooLarge` error

And I see the same pattern with other video formats (e.g. HEVC), so maybe it's an iOS (or Expo?) thing if it works fine on desktop?

surfdude29:
Excellent, it now triggers the new `FileTooBig` error for me every time I select a video larger than 100 MB 🙏

IMG_8337

Also, this isn't a criticism, and I imagine that the scenario where a user tries to upload a GIF larger than 100 MB will be a rare edge case, but I was interested to see that when I try selecting a gargantuan 180 MB GIF that I made to test with, it triggers the `VideoTooLargeError` rather than the `FileTooBig` error. I guess even though an uploaded GIF is (I think) mostly treated as a video, it goes down a different code path that ends up at a different error?

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,609 @@
...
+            msg`You can only select one video at a time.`,
+          ),
+          [SelectedAssetError.VideoTooLong]: _(
+            msg`Videos must be less than 3 minutes long.`,
+          ),
+          [SelectedAssetError.MaxGIFs]: _(
+            msg`You can only select one GIF at a time.`,
+          ),
+          [SelectedAssetError.FileTooBig]: _(
+            msg`This file is too large. Maximum size is 100mb.`,
```
surfdude29, line 494:
```suggestion
            msg`This file is too large. Maximum size is 100 MB.`,
```
nit

estrattonbailey:
@surfdude29 I think my latest commit should fix the GIF error, since it checks the size as an early step.

surfdude29:
@estrattonbailey It does indeed fix the GIF error for me, and it's so much speedier in general at showing the `FileTooBig` error message, especially for very large videos. Really nice work! 💫

yarn.lock
```diff
@@ -11295,7 +11295,7 @@ expo-image-manipulator@~13.1.7:
   dependencies:
     expo-image-loader "~5.1.0"
 
-expo-image-picker@~16.1.4:
+expo-image-picker@^16.1.4:
   version "16.1.4"
```
mozzius: Doesn't look like it actually got bumped? lockfile still points to v16

estrattonbailey: Heh whoops, fixed now

surfdude29:
> … it's so much speedier in general at showing the `FileTooBig` error message, especially for very large videos …

And with the new version of `expo-image-picker` it shows up instantly now, wow! 🚀

src/view/com/composer/SelectMediaButton.tsx
```diff
@@ -0,0 +1,514 @@
...
+            msg`You can only select one video at a time.`,
+          ),
+          [SelectedAssetError.VideoTooLong]: _(
+            msg`Videos must be less than 3 minutes long.`,
+          ),
+          [SelectedAssetError.MaxGIFs]: _(
+            msg`You can only select one GIF at a time.`,
+          ),
+          [SelectedAssetError.FileTooBig]: _(
+            msg`One or more of your selected files is too large. Maximum size is 100MB.`,
```
surfdude29:
```suggestion
            msg`One or more of your selected files is too large. Maximum size is 100 MB.`,
```
nit: maybe with a no-break space between value and unit?

estrattonbailey:
Let's give this a go

PR #8829, opened 2025-08-13
Fix deletion of composed emojis in web composer

### LINKED ISSUE #8660: Web composer - composite emoji get deleted one unicode character at a time
### Steps to Reproduce

1. Put a country flag into the web post composer
2. Press backspace
3. Observe it turns into a black flag, instead of deleting the entire flag

### Attachments

https://github.com/user-attachments/assets/e8a0a4ec-6c7d-4b44-a395-c779b9f8dd6c

Web (Desktop)

### Device Info

Firefox

1.104.0

### Additional Information

_No response_

arturo32:
This PR fixes #8660 where the deletion of composed emojis (such as 🏴󠁧󠁢󠁥󠁮󠁧󠁿, 🏴󠁧󠁢󠁳󠁣󠁴󠁿, 🏴󠁧󠁢󠁷󠁬󠁳󠁿, 🏳️‍⚧️, 🏳️‍🌈) by pressing the `backspace` key on the web composer don't work as expected, i.e., it leaves behind "root" emojis (such as 🏳️ or 🏴󠁧󠁢󠁷󠁬󠁳).

This is a [common problem](https://github.com/Expensify/App/issues/51336#issuecomment-2438017697) that happens in [many projects](https://github.com/element-hq/element-web/issues/22196) that have text input. You can even test this on this PR's comments. And a common solution is to use a library that splits the graphemes of the text, to delete the text-units that the user sees.

Bluesky's social-app already uses a library that does that: Graphemer.

This fix uses the `keyDown` event to recognize presses on the `backspace` key and splits all text before the cursor in graphemes. It then removes only the last grapheme by using a [TipTap command](https://tiptap.dev/docs/editor/api/commands/selection/delete-range).

If the text is selected, the default behavior already deletes the graphemes correctly. The same goes for the `delete` key.

**Before:**

https://github.com/user-attachments/assets/79f67208-a1cb-4d3c-a690-189491e74386

**After:**

[screncapture showing flag emojis being erased normally on the bluesky web composer](https://github.com/user-attachments/assets/182b80d0-c8cb-411b-9753-859eb6101806)

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -13,6 +13,7 @@ import {Text as TiptapText} from '@tiptap/extension-text'
 import {generateJSON} from '@tiptap/html'
 import {Fragment, Node, Slice} from '@tiptap/pm/model'
 import {EditorContent, type JSONContent, useEditor} from '@tiptap/react'
+import Graphemer from 'graphemer'
 
 import {useColorSchemeStyle} from '#/lib/hooks/useColorSchemeStyle'
 import {usePalette} from '#/lib/hooks/usePalette'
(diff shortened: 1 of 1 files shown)


mozzius (approved):
Awesome stuff, thank you!

PR #8830, opened 2025-08-13
Cross Platform Syntax Highlighting

skiniks:
Introduces cross-platform syntax highlighting via Shiki (native and web), integrates provider/context, and renders code blocks in RichText. Requires New Architecture enabled for the native engine (`react-native-shiki-engine`). On web this uses `@shikijs/engine-oniguruma` with the inlined WASM.

- Introduced Shiki for syntax highlighting.
- Added new components: HighlighterProvider, HighlighterContext, and CodeBlockText.
- Updated App components to include HighlighterProvider.
- Enhanced RichText component to render code blocks using CodeBlockText.
- Added support for multiple programming languages and themes.
- Updated package.json with new dependencies for Shiki.

### Open questions for review
- Languages set (size vs coverage)
  - Current definitions live in `src/lib/shiki/langs.ts`.
  - Preference?
    - Option A: Bundle the full Shiki language set (or web's bundled set) for zero config.
    - Option B (current): Curated set to minimize size and memory.
- Theme selection
  - Current definitions in `src/lib/shiki/themes.ts` with default `nord`.
  - Preference?
    - Option A: Nord (current default).
    - Option B: GitHub Dark/Light pair.
    - Option C: Other.

### Blockers
- New Architecture must be enabled to ship the native engine.

### Related/Blocking work
- Code embed type proposal:
  - Code embed PR: [PR #3780](https://github.com/bluesky-social/atproto/pull/3780)
  - Summary:
    - Adds `app.bsky.embed.code` lexicon with `code` (≤8000 chars) and `lang` (≤20 chars).
    - Integrates with existing embed/view system.

### Request for decisions
- Approve curated language list vs full bundle.
- Choose default theme strategy (single default vs light/dark pair).
- Confirm proceeding with the code embed PR as the integration path.


src/App.native.tsx
```diff
@@ -19,6 +19,7 @@ import * as Sentry from '@sentry/react-native'
 import {KeyboardControllerProvider} from '#/lib/hooks/useEnableKeyboardController'
 import {Provider as HideBottomBarBorderProvider} from '#/lib/hooks/useHideBottomBarBorder'
 import {QueryProvider} from '#/lib/react-query'
+import {HighlighterProvider} from '#/lib/shiki/HighlighterProvider.native'
```
mozzius:
```suggestion
import {HighlighterProvider} from '#/lib/shiki/HighlighterProvider'
```

skiniks:
Thanks for taking a look this morning, @mozzius. I noticed the New Architecture was rolled back, so the native Shiki engine is blocked on iOS for now. In the meantime, when you have a moment, could you confirm whether the code-embed approach in [PR #3780](https://github.com/bluesky-social/atproto/pull/3780) is the right direction, and give it a review so we can move this new PR forward once the New Arch is re-enabled?

src/components/RichText.tsx
```diff
@@ -59,6 +60,41 @@ export function RichText({
   const {text, facets} = richText
 
   if (!facets?.length) {
+    const fencedRegex = /```([a-zA-Z0-9_-]+)?\n([\s\S]*?)```/g
```
isuffix: This is a pretty limited set of characters for the language tag. I don’t know what languages Shiki does or doesn’t include, but this should be more careful about allowing non-Latin language names. I would recommend changing to use the Unicode XID property class with `\p{XID_Continue}`. Note that it includes underscore, but not hyphen.

I’d then also include some symbols like `+`, `#`, `*`, and `!` for safety (each of those can be found in various language names on Rosetta code). And I’d allow trailing whitespace in case people are sloppy or editing on mobile/etc.

So I would change the regex to:
````suggestion
    const fencedRegex = /```([\p{XID_Continue}+#*!-]+)?[^\n\S]*\n([\s\S]*?)```/ug
````

isuffix: Ope. I just now looked at surrounding context and realized this is not quite recent. Apologies if this draft isn’t still relevant.

skiniks: Great catch. Just applied it to the draft!

skiniks:
Hey @mozzius & @estrattonbailey, circling back on this now that the New Architecture is enabled.

I've merged the latest `main` into this branch, and brought the Shiki stack current. Also addressed the fenced language-tag regex feedback brought up by @isuffix.

If it's useful as a reference, `react-native-shiki-engine` is also in production use in [t3-code](https://github.com/pingdotgg/t3code)'s mobile app.

If syntax highlighting is something you still want to ship, I'm happy to own getting this across the finish line. I'll need direction on the remaining product calls (language set / theme, lexicon / code embed, and anything else on your side).

PR #8831, opened 2025-08-13
fix: don't open user profile when clicking edit button in user list

elijaharita:
Just prevents default on the edit button to avoid activating the link when clicked.


mozzius (approved):
lovely, thank you!

PR #8832, opened 2025-08-13
feat: use OP's language as recommendation when replying

elijaharita:
This immediately suggests the language of the post you are replying to. Once you start typing and the language model kicks in, it will update with the computed language based on the content of your reply like it would have before.

Since langs is an array, I am just taking the first entry as the suggested language. Not sure if this makes sense since I'm not sure in which situations multiple languages can be applied to a post.

image

### DIFF
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -19,10 +19,16 @@ import {Text} from '#/components/Typography'
 const onIdle = globalThis.requestIdleCallback || (cb => setTimeout(cb, 1))
 const cancelIdle = globalThis.cancelIdleCallback || clearTimeout
 
-export function SuggestedLanguage({text}: {text: string}) {
+export function SuggestedLanguage({
+  text,
+  fallback,
+}: {
+  text: string
+  fallback?: string
+}) {
   const [suggestedLanguage, setSuggestedLanguage] = useState<
     string | undefined
-  >()
+  >(fallback)
   const langPrefs = useLanguagePrefs()
   const setLangPrefs = useLanguagePrefsApi()
   const t = useTheme()
@@ -34,7 +40,7 @@ export function SuggestedLanguage({text}: {text: string}) {
     // Don't run the language model on small posts, the results are likely
     // to be inaccurate anyway.
     if (textTrimmed.length < 40) {
-      setSuggestedLanguage(undefined)
+      setSuggestedLanguage(fallback)
       return
     }
 
@@ -43,7 +49,7 @@ export function SuggestedLanguage({text}: {text: string}) {
     })
 
     return () => cancelIdle(idle)
-  }, [text])
+  }, [text, fallback])
 
   if (
     suggestedLanguage &&
--- a/src/screens/PostThread/components/ThreadItemAnchor.tsx
+++ b/src/screens/PostThread/components/ThreadItemAnchor.tsx
@@ -254,6 +254,7 @@ const ThreadItemAnchorInner = memo(function ThreadItemAnchorInner({
         author: post.author,
         embed: post.embed,
         moderation,
+        langs: record.langs,
       },
       onPostSuccess: onPostSuccess,
     })
--- a/src/screens/PostThread/components/ThreadItemPost.tsx
+++ b/src/screens/PostThread/components/ThreadItemPost.tsx
@@ -234,6 +234,7 @@ const ThreadItemPostInner = memo(function ThreadItemPostInner({
         author: post.author,
         embed: post.embed,
         moderation,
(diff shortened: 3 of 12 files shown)


src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -19,10 +19,16 @@ import {Text} from '#/components/Typography'
 const onIdle = globalThis.requestIdleCallback || (cb => setTimeout(cb, 1))
 const cancelIdle = globalThis.cancelIdleCallback || clearTimeout
 
-export function SuggestedLanguage({text}: {text: string}) {
+export function SuggestedLanguage({
+  text,
+  fallback,
+}: {
+  text: string
+  fallback?: string
```
mozzius: nit: maybe rename to `replyToLanguage` for clarity?

src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -34,7 +40,7 @@ export function SuggestedLanguage({text}: {text: string}) {
     // Don't run the language model on small posts, the results are likely
     // to be inaccurate anyway.
     if (textTrimmed.length < 40) {
-      setSuggestedLanguage(undefined)
+      setSuggestedLanguage(fallback)
```
mozzius: nit: rather than setting it to the parent post language when less than 40 characters, I think we should only show it before they start typing. I think we're going to get a fair number of false positives so let's not make it too overbearing

Maybe revert that back to `setSuggestedLanguage(undefined)` and then add this before:

```ts
    // before they start typing, suggest the language of the post
    // they're replying to, if applicable
    if (replyToLanguage && text.length === 0) {
      setSuggestedLanguage(replyToLanguage)
      return
    }
```

src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -19,10 +19,16 @@ import {Text} from '#/components/Typography'
...
+  text,
+  fallback,
+}: {
+  text: string
+  fallback?: string
+}) {
   const [suggestedLanguage, setSuggestedLanguage] = useState<
     string | undefined
-  >()
+  >(fallback)
```
mozzius: nit, if we're doing the other change

```suggestion
  >(text.length === 0 ? replyToLanguage : undefined)
```

surfdude29:
This would be a very welcome improvement 🙏

When a similar potential feature [was discussed previously](https://github.com/bluesky-social/social-app/issues/7260#issuecomment-2661413779), one point was raised that I believe would be worth thinking about here.

Following the scenario illustrated in the screenshot above, if the user clicks **Yes** and the post language setting for this reply is switched to Japanese, what happens to the setting after they've made this reply? If they then create a new post, would the language for that post be automatically set to Japanese (as it's the last setting used), or would it revert to the language setting that the user previously had (English, in the screenshot example)?

If the user clicks **Yes** to accept the recommendation, I think it would be desirable for the post language setting change to only be temporary, taking effect for the reply only, and then reverting to the post language setting that the user previously had.

mozzius:
Agree 👍

KernelLeak:
While that's one prompt more than I'd like it's still way less error-prone than forgetting to change the language and less clicking than always manually switching languages and back immediately after replying...

elijaharita:
I agree as well. I actually thought that this setting was local to the post that was being edited when I first wrote this. I'll think about how I can implement this nicely. I'm feeling like this should probably be saved as a component state instead of a context, as that would automatically clear the setting as soon as we unmount, which sounds like the desirable behavior to me. Thoughts?

elijaharita:
I guess this raises the question of if the user accepts the recommendation, and then goes and manually adds more languages to the post. Does it stay temporary in this case, or would this lock in the recommendation as a permanent thing?

I guess this is leading me to feel like it might be nice to just allow the user to choose whether they are setting the global or local setting? Set it to be post-specific by default, and if they want to change their global setting, they just press a toggle and then do the same action since that would conceivably be less frequent?

Maybe we respond to this problem in a followup? Unless we consider this to be a blocking problem to solve first

mozzius:
I think this solution is fine for now. But we should definitely consider temporary vs permanent changes, because the current system is not great

elijaharita:
@mozzius when you say this solution is fine for now, are you saying the current PR is fine as-is?

mozzius:
Yeah I think this is good! Will re-review and stamp this one, and we should follow up with rethinking the language switching behaviour in a separate ticket

surfdude29:
@elijaharita @mozzius Tried this out on the dev server, very nice 👌

I have one suggestion for a small follow-up tweak: if the user selects a post language different to the suggested `replyToLanguage`, I think that should dismiss the prompt, because then they've expressly made a choice.

PR #8847, opened 2025-08-14
Add suggested follows experiment to onboarding

mozzius:
Suggested follows screen in onboarding. Includes some tweaks to the rest of onboarding

# Native

# Web

Screenshot 2025-08-21 at 16 08 02

# Test plan

- Confirm gate works, confirm there's the right number of steps in both A and B groups
- Test screen
  - Interests are passed through to the suggested user query and revalidate if changed
  - Following users
  - Follow all
  - Tabs
  - Language behaviour - same as explore, non-english users should get the search-based fallback
  - Relevant actions have metrics

Designs have follow-up suggestions after each follow - will do a in a follow-up PR


src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  const [selectedInterest, setSelectedInterest] = useState<string | null>(null)
+  // keeping track of who was followed via the follow all button
+  // so we can enable/disable the button without having to dig through the shadow cache
+  const [followedUsers, setFollowedUsers] = useState<string[]>([])
+
+  /*
+   * Special language handling copied wholesale from the Explore screen
+   */
+  // FIGURE OUT BEFORE MERGE: IS THIS POPULATED BEFORE FIRST ONBOARDING
```
haileyok: figure out

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+    overrideInterests: state.interestsStepResults.selectedInterests,
+  })
+
+  const isError = !!error
+
+  const saveInterests = useCallback(async () => {
+    setSaving(true)
+
+    try {
+      setSaving(false)
```
haileyok: I don't think this `setSaving(true)` will have any effect?

src/screens/Onboarding/index.tsx
```diff
@@ -46,13 +54,16 @@ export function Onboarding() {
     <Portal>
       <OnboardingControls.Provider>
         <Context.Provider
-          value={React.useMemo(
+          value={useMemo(
             () => ({state, dispatch, interestsDisplayNames}),
             [state, dispatch, interestsDisplayNames],
           )}>
```
haileyok: We don't need memo at all here, can rm

haileyok:
This is the view i get on web 
image

haileyok:
Logic seems to work fine, though the `followAll()` button doesn't do anything re: processing state as noted. Plus the web UI problem (DPI?)

src/screens/Onboarding/index.tsx
mozzius: This is actually the sort of case I'd be worried about removing it from because if it doesn't get automatically added it would lead to perf degradation

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
mozzius: Whoops lol

src/screens/Onboarding/index.tsx
haileyok: /gen, are we worried about that? in that, is there a possible scenario where this would break with compiler?

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+    refetch,
+  } = useSuggestedUsers({
+    category: selectedInterest || (useFullExperience ? null : interests[0]),
+    search: !useFullExperience,
+    overrideInterests: state.interestsStepResults.selectedInterests,
+  })
+
+  const isError = !!error
+
+  const saveInterests = useCallback(async () => {
```
estrattonbailey: Nit: outdated name

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+      for (const did of followableDids) {
+        const uri = uris.get(did)
+        updateProfileShadow(queryClient, did, {
+          followingUri: uri,
+        })
+      }
+      return followableDids
+    },
+    onSuccess: newlyFollowed => {
+      toast.show(_(msg`Done!`), {type: 'success'})
```
estrattonbailey: Nit: might as well say more like `Followed all accounts!`

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+              {minHeight: 400},
+            ]}>
+            <Loader size="xl" />
+          </View>
+        ) : isError ? (
+          <View style={[a.flex_1, a.px_xl, a.pt_5xl]}>
+            <Text
+              style={[a.text_center, a.text_md, t.atoms.text_contrast_medium]}>
+              <Trans>
+                An error occurred while fetching suggested accounts.
```
estrattonbailey, line 177: Similar flex issue here. Also, might as well use an `Admonition`?
Image

estrattonbailey (commented):
Yeah just some wacky flex stuff going on, seems like. Logic seems to work though 👍

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+      )
+    },
+  })
+
+  const canFollowAll = followableDids.length > 0 && !isFollowingAll
+
+  return (
+    <View style={[a.align_start]} testID="onboardingInterests">
+      <Text style={[a.font_heavy, a.text_3xl]}>
+        <Trans>Suggested for you</Trans>
```
surfdude29:
```suggestion
        <Trans comment="Accounts suggested to the user for them to follow">
          Suggested for you
        </Trans>
```
nit: Suggest adding comment for translators here

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+          a.flex_1,
+          a.justify_start,
+        ]}>
+        <SuggestedAccountsTabBar
+          selectedInterest={selectedInterest}
+          onSelectInterest={setSelectedInterest}
+          defaultTabLabel={_(
+            msg({
+              message: 'All',
+              comment: 'the default tab in the interests tab bar',
```
surfdude29, line 156: Appreciate the comment here 🙏

src/screens/Onboarding/index.tsx
mozzius: well I don't think so. the problem with Compiler is that there's no easy way to tell what it's actually doing without looking at the generated code as far as I can tell?

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
mozzius: Done!

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
mozzius: Done: changed sim language to french and hard reinstalled, looked good

Simulator Screenshot - iPhone 16 - 2025-08-26 at 15 26 46

estrattonbailey (approved):
Beautiful work sir 🫡

PR #8850, opened 2025-08-14
[APP-1303] Redesign/refactor post language select

APiligrim:
This PR addresses [this ticket ](https://linear.app/blueskyweb/issue/APP-1303/redesignrefactor-post-language-select) and updates the post language select to the latest design. It also removes some depreciated dependencies that were used to newer components.

Design:
Screenshot 2025-08-20 at 11 21 31 AM
Screenshot 2025-08-21 at 10 09 34 AM

Screenshots:

**IOS**

https://github.com/user-attachments/assets/95a21e8c-75e6-486b-afc8-8ffd5de22fd3

**Android**

https://github.com/user-attachments/assets/590a47f3-5c63-4c97-9272-5f7f7b919269

**Desktop**

https://github.com/user-attachments/assets/2c2dd86f-4330-4388-882d-66d3418708b3


src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -1,8 +1,8 @@
-import {useCallback, useMemo} from 'react'
+import {useCallback} from 'react'
 import {Keyboard, StyleSheet} from 'react-native'
 import {
   FontAwesomeIcon,
-  FontAwesomeIconStyle,
+  type FontAwesomeIconStyle,
 } from '@fortawesome/react-native-fontawesome'
```
mozzius: Can we replace the font awesome icon with one from the design system?

mozzius:
Looking good, but we need to move it from a Modal to a Dialog (explained more in slack)

src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -1,133 +1,73 @@
...
+        )}
+      </Button>
+
+      <Dialog.Outer control={control}>
+        <Dialog.ScrollableInner
+          accessibilityLabelledBy="dialog-title"
+          accessibilityDescribedBy="dialog-description">
+          <PostLanguagesSettingsDialog onClose={control.close} />
+        </Dialog.ScrollableInner>
+      </Dialog.Outer>
```
mozzius: Easy to forget, but needs a close button on web!

```suggestion
        </Dialog.ScrollableInner>
        <Dialog.Close />
      </Dialog.Outer>
```

src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -1,133 +1,73 @@
...
+            {postLanguagesPref
+              .map(lang => codeToLanguageName(lang, langPrefs.appLanguage))
+              .join(', ')}
+          </Text>
+        ) : (
+          <GlobeIcon size="xs" style={[{color: t.palette.primary_500}]} />
+        )}
+      </Button>
+
+      <Dialog.Outer control={control}>
```
mozzius: Nit: it's nicer and neater to encapsulate the entire dialog in one component rather than putting the Outer part somewhere else.

So rough pseudocode:

```tsx
function SomethingBtn() {
  const control = Dialog.useDialogControl()

  return (
    <>
       <Button onPress={control.open} />
       <SomethingDialog control={control} />
    </>
  )
}

// maybe in a different file
function SomethingDialog({control}) {
  return (
      <Dialog.Handle />
      <SomethingDialogInner />
    </Dialog.Outer>
  )
}

// perf optimisation: if we move the Inner part to its own component,
// it only gets added to the React tree when the dialog is open
function SomethingDialogInner({}) {
  return (
    <Dialog.ScrollableInner>
      // dialog goes here
    </Dialog.ScrollableInner>
  )
}
```

src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -1,133 +1,73 @@
...
-    </DropdownButton>
+    <>
+      <Button
+        testID="selectLangBtn"
+        onPress={onPressMore}
+        variant="ghost"
+        size="small"
+        hitSlop={LANG_DROPDOWN_HITSLOP}
+        label={_(msg`Language selection`)}
+        accessibilityLabel={_(msg`Language selection`)}
```
surfdude29: Just curious, does `accessibilityLabel` do something different to `label`? I kind of assumed they did the same thing, but then I don't know that much about accessibility 😅

src/view/com/composer/select-language/SelectLangBtn.tsx
```diff
@@ -1,133 +1,73 @@
...
+        accessibilityHint={_(msg`Opens language settings`)}
```
surfdude29:
```suggestion
        accessibilityHint={_(msg`Opens post languages settings`)}
```
Maybe this could be tweaked to make it clearer that it refers to settings for the language(s) of the post, rather than general language settings for the app? :)

src/view/com/composer/select-language/SelectLangBtn.tsx
mozzius: `label` just sets `accessibilityLabel`, the difference is that it's a required type. Good catch, the latter one should be removed
```suggestion
```

mozzius:
One thing I'm not so sure about is the nested ScrollView - generally something to avoid. Will play around with it though, maybe we can make it work

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,210 @@
...
+    </>
+  )
+}
+
+function LanguageDialog({control}: {control: Dialog.DialogControlProps}) {
+  return (
+    <Dialog.Outer control={control}>
+      <Dialog.Handle />
+      <PostLanguagesSettingsDialogInner onClose={control.close} />
+      <Dialog.Close />
```
mozzius: Annoyingly you have to put it *inside* the ScrollableInner as the last element, otherwise it floats around outside of the dialog box

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const postLanguagesPref = toPostLanguages(langPrefs.postLanguage)
+
+  return (
+    <>
+      <Button
+        testID="selectLangBtn"
+        onPress={onPressMore}
+        size="small"
+        hitSlop={LANG_DROPDOWN_HITSLOP}
+        label={_(msg`Post language selection`)}
```
surfdude29:
```suggestion
        label={_(
          msg({
            message: `Post language selection`,
            comment: `Accessibility label for button that opens dialog to choose post language settings`,
          }),
        )}
```
A comment for translators might be helpful here, although I wouldn't say it's essential.

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
APiligrim: updated it!

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+      accessibilityLabelledBy="dialog-title"
+      accessibilityDescribedBy="dialog-description">
+      <View style={[a.relative, a.flex_1]}>
+        <View
+          testID="postLanguagesDialog"
+          style={[
+            t.atoms.bg,
+            a.flex_1,
+            a.px_sm,
+            isMobile ? [a.pt_xl] : [a.pt_lg, {maxHeight: deviceHeight * 0.7}],
```
estrattonbailey, line 148: The dialog already has built-in gutters, can remove padding here

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+            <Trans>Which languages are used in this post?</Trans>
+          </Text>
+
+          <View style={[a.flex_1, {maxHeight: isMobile ? '72%' : '75%'}]}>
+            <ScrollView
+              style={[
+                a.flex_1,
+                a.px_0,
+                a.gap_sm,
+                {maxHeight: deviceHeight * 0.8},
```
mozzius, line 182: This approach doesn't work at all unfortunately

Simulator Screenshot - iPhone 16 - 2025-08-19 at 16 34 06

mozzius, line 182: This approach doesn't work on different screen sizes

Screenshot 2025-08-19 at 16 36 55

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+      <PostLanguagesSettingsDialogInner onClose={control.close} />
+    </Dialog.Outer>
+  )
+}
+
+export function PostLanguagesSettingsDialogInner({
+  onClose,
+}: {
+  onClose: () => void
+}) {
```
mozzius, line 94: You don't need to pass this down! Instead, in `PostLanguagesSettingsDialogInner`, use

```ts
const control = Dialog.useDialogContext()
control.close()
```

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+        toggleList.includes(langB.code2) ||
+        deviceLanguageCodes.includes(langB.code2)
+      if (hasA === hasB) return langA.name.localeCompare(langB.name)
+      if (hasA) return -1
+      return 1
+    })
+    return langs
+  }, [toggleList])
+
+  const deviceHeight = Dimensions.get('window').height
```
mozzius, line 135: If you're have to get the device height, prefer `useWindowDimensions()`

src/view/com/composer/select-language/SelectPostLanguagesDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  const onPressDone = React.useCallback(() => {
+    let langsString = toggleList.join(',')
+
+    if (!langsString) {
+      langsString = langPrefs.primaryLanguage
+    }
+
+    setLangPrefs.setPostLanguage(langsString)
+    onClose()
```
mozzius, line 111: Subtle one, but generally speaking if a dialog "does something" when it closes, you should do it *after* the dialog closes by passing a callback to `control.close`. This ensures it doesn't get into a weird state

```suggestion
    control.close(() => {
      setLangPrefs.setPostLanguage(langsString)
    })
```

PR #8852, opened 2025-08-15
Remove "that's all, folks!" message

mozzius:
I don't know why I added it. It's stupid. Let's remove it.

### DIFF
--- a/src/view/com/post-thread/PostQuotes.tsx
+++ b/src/view/com/post-thread/PostQuotes.tsx
@@ -1,9 +1,9 @@
 import {useCallback, useState} from 'react'
 import {
-  AppBskyFeedDefs,
+  type AppBskyFeedDefs,
   AppBskyFeedPost,
   moderatePost,
-  ModerationDecision,
+  type ModerationDecision,
 } from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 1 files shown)


surfdude29:
I thought @estrattonbailey was [to blame](https://bsky.app/profile/pfrazee.com/post/3l2s7gjafno2b)?

Or [imho](https://bsky.app/profile/surfdude29.ispost.ing/post/3l2s7wg7ebo2b) (and [some](https://bsky.app/profile/aghastronaut.com/post/3l2s75a25jh23) [others](https://bsky.app/profile/rockyartz.bsky.social/post/3lvfdz3aiwc26)), thank 😄

mozzius:
Oh haha so it was, git blame points at me but it was Eric

sorry Eric but its gotta go 😤

estrattonbailey:
image

mozzius:
image

ioletsgo:
If you remove it, you have to replace it with something else.
"It's quiet. Too quiet."
"I admire its purity."
"Luke, I am your Quote Posts."
These are just a few that I thought about off the top of my head. If you consider removing one silly thing, please consider if replacing it with something even sillier would be a good move, if not, don't remove it.

There's no need for bluesky to be a completely sanitized corporate environment.

ioletsgo:
if you would like an alternative proposal, one that might keep the whimsy of the site in check, could I maybe suggest putting flavor text like this up to an easily editable external file where several entries could be added to have a chance to appear. 

I think it might prove to be really engaging having people write non-sequiturs, could help make people feel like they're being heard, while also making them feel like bluesky is a platform for them if it was made up to be a friendly competition on the site. Imagine if, instead of "Post Deleted" it was "Sorry I ated it. 404 Error.", it would definitely give bluesky its own specific touch that would distinguish it from competition.

Before Elon Musk took over X dot com, it actually had a kind of similar vibe? Bluesky isn't really trying as much to recapture such a vibe and I think that's kind of miss potential, especially as Elon continues to strip X dot Com of all of its previous identity, those goofy esoteric error illustrations and all.

PR #8856, opened 2025-08-15
Fix `isPartOfLastBranchFromDepth` assignment

estrattonbailey:
Fixes a bug where `isPartOfLastBranchFromDepth` was not updated if further replies down the tree were in fact _also_ "the last branch from `depth`".

See screenshot, where the posts should have the following values:

A. `isPartOfLastBranchFromDepth: 1`, since it's the last sibling, and its depth is `1`
B. `isPartOfLastBranchFromDepth: undefined` since it's not the last sibling (there's a post further down)
C. `isPartOfLastBranchFromDepth: 3` since it's a last sibling and its depth is `3`
D. `isPartOfLastBranchFromDepth: 4` since it _itself_ is also a last sibling and its depth is `4`
E. `isPartOfLastBranchFromDepth: undefined` since it's not a last sibling (it's the first)

Case `D` is what this PR fixes.

CleanShot 2025-08-15 at 16 07 00@2x

### DIFF
--- a/src/state/queries/usePostThread/traversal.ts
+++ b/src/state/queries/usePostThread/traversal.ts
@@ -307,9 +307,16 @@ export function sortAndAnnotateThreadItems(
               metadata.isPartOfLastBranchFromDepth = metadata.depth
 
               /**
-               * If the parent is part of the last branch of the sub-tree, so is the child.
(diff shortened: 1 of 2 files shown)


mozzius (approved):
ok I thiiiink I understand, that's very subtle

PR #8863, opened 2025-08-18
use an env var for configurable proxy header

haileyok:
## Why

A few things motivate this:
- A bug with some account migrations would benefit from this
- It's too easy right now to not know when a proxy header is set
- We should have done this a long time ago...the PDS shouldn't be required to fallback for requests to work, the app should always be telling the PDS what service to proxy to if it's supposed to be proxying

## How

Relatively easy, though there are a few footguns we want to avoid.

One is to swap out all of our usages of `AtpAgent` or `Agent` with our own `Agent` (which just extends the `Agent` we were already using from `@atproto/api`) to include a parameter for a proxy header that is required. Passing `null` in for this is fine, but we want to make it clear whenever we instantiate a new client that there is or is not a header set.

The other is to update our `BskyAppAgent` to also set the Bluesky appview proxy header as soon as we finish resuming a session, creating one, or creating an account. Again, we want to avoid the possibility of forgetting at some later point.

## Test Plan

1. Feeds still load properly, including ones that need the browser's DID (e.g. Discover)
2. Changing password, email, etc. still works fine
3. Chat service still works correctly
4. General test of all the rest of the app, just use it.

### DIFF
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -259,7 +259,7 @@ export function sessionAccountToSession(
 
 export class Agent extends BaseAgent {
   constructor(
-    proxyHeader: string | null,
+    proxyHeader: ProxyHeaderValue | null,
     options: SessionManager | FetchHandler | FetchHandlerOptions,
   ) {
     super(options)


.env.example
```diff
@@ -16,6 +16,9 @@ EXPO_PUBLIC_LOG_LEVEL=debug
 # Enable debug logs for specific logger instances
 EXPO_PUBLIC_LOG_DEBUG=session
 
+# Bluesky appview DID
+EXPO_PUBLIC_BLUESKY_PROXY_DID=
```
estrattonbailey, line 20: App Views are always `did:web` right? Perhaps it'd be helpful to call this `EXPO_PUBLIC_APPVIEW_DID_WEB`, which is more self documenting.

Also maybe our default value would be good here as an example?

haileyok:
> Can we use Agent.prototype.configureProxy here? 

Avoided this since it is deprecated and no longer available on the new `Agent` class.

.env.example
haileyok: I don't actually know if that is the case. A `did:plc:` could have an appview in it's DID doc (same with, e.g., a feed. you'd just list it under the services object)

haileyok: Example of a feedgen inside a `did:plc:` doc is https://web.plc.directory/did/did:plc:4uz2445cjiw7w4nobfgnu35f

estrattonbailey: Oh ok sg

estrattonbailey:
@haileyok `configureProxy` looks fine, `configureProxyHeader` is deprecated

haileyok:
Oh nice, missed that. Switched to that then.

.env.example
devinivy: Accurate 👍

PR #8866, opened 2025-08-18
[APP-1398] Fix post shadow/like shadow state when replying

estrattonbailey:


surfdude29:
Would this be about fixing the bug where if you like a post and then reply, the like disappears after your reply is published? I've been running into that a lot recently, been meaning to report it.

estrattonbailey:
@surfdude29 yep exactly 👍

src/state/queries/usePostThread/queryCache.ts
```diff
@@ -85,10 +89,13 @@ export function createCacheMutator({
...
-            ...parent.value.post,
-            replyCount: (parent.value.post.replyCount || 0) + 1,
-          }
+          const shadow = dangerousGetPostShadow(parent.value.post)
+          const prevOptimisticCount = shadow?.optimisticReplyCount || 0
+          const prevReplyCount = parent.value.post.replyCount || 0
+          updatePostShadow(queryClient, parent.value.post.uri, {
+            // prefer optimistic count, if we already have some
+            optimisticReplyCount: (prevOptimisticCount || prevReplyCount) + 1,
+          })
```
mozzius: Would this make the priority a bit clearer?

```suggestion
          const prevOptimisticCount = shadow?.optimisticReplyCount
          const prevReplyCount = parent.value.post.replyCount
          updatePostShadow(queryClient, parent.value.post.uri, {
            // prefer optimistic count, if we already have some
            optimisticReplyCount: (prevOptimisticCount ?? prevReplyCount ?? 0) + 1,
          })
```

mozzius (changes requested):
This is the right approach, but I think something within the thread system isn't consuming from the shadow cache because it's missing the line:

Screenshot 2025-08-19 at 16 16 26

I was able to fix it by adding the line:

```ts
parent.value.post.replyCount = (parent.value.post.replyCount || 0) + 1
````

But that seems _really_ sketchy

**Comment by n0vedad** [comment]
Great work on identifying and fixing this issue! This is funny because I've spent the last few days investigating the same bug too. Some people were very confused by "vanishing" likes in a very specific flow. 

After adding some logging and analyzing, I wanted to share some insights how I understand them:

The core problem is introduced by V2's thread handling at https://github.com/bluesky-social/social-app/pull/8443 when **post_threads_v2_unspecced** start breaking things. So when`insertReplies()` processes new replies, it can create fresh Post object references, breaking WeakMap-based shadow state lookups AND break React query cache --> Like seems to "vanish" until refresh.

Your fix correctly addresses reply counts by using `dangerousGetPostShadow()` to read existing shadow state before `insertReplies()` creates new Post references. V2's `insertReplies()` doesn't aggressively replace the parent Post reference in this specific flow, `optimisticReplyCount` gets tracked and surviving the WeakMap breaks.

But I'm afraid the underlying WeakMap reference fragmentation could still affect more shadow properties and still break things thourgh memory pressure. So I experimented with a hybrid approach for comprehensive protection and use a hybrid WeakMap (for backwards compatibility) + URI-based shadow fallback system which works too.

```
//src/state/cache/post-shadow.ts
const uriShadows: Map<string, Partial<PostShadow>> = new Map()
```

Theoretically there must be a periodically cleanup for URI-Maps than and I don't know how this scale at production, but it shouldn't be a problem through time-limited user sessions while a user can see a limited amount of posts and should have a mimimal "memory footprint". But maybe I'm missing something because this is my first deep try with Bluesky developing.

Nevertheless the current fix is great for the immediate bug fixing!

surfdude29:
I'm trying this out on the preview branch, but it's put me back on threads v1 which I'm a little confused by, as I thought the threads v2 gate was an account-level thing? 🤔

mozzius:
I haven't checked but there's a good chance the feature gates don't work on the preview branches

src/state/queries/usePostThread/queryCache.ts
estrattonbailey: Dig it 👍

PR #8867, opened 2025-08-18
Add profile link to switcher menu

Inbestigator:
This is an annoying experience I've run into a bunch where I instinctively click the top avatar button to go to my profile.

This also has more parity with the mobile version, as clicking the avatar on mobile will take you to your profile instead of opening a menu.

Screenshot 2025-08-18 at 3 32 44 PM
Screenshot 2025-08-18 at 3 41 39 PM

### DIFF
--- a/src/view/shell/desktop/LeftNav.tsx
+++ b/src/view/shell/desktop/LeftNav.tsx
@@ -193,7 +193,6 @@ function ProfileCard() {
           </Menu.Trigger>
           <SwitchMenuItems
             accounts={otherAccounts}
-            currentAccount={currentAccount}
             signOutPromptControl={signOutPromptControl}
           />
         </Menu.Root>
@@ -219,7 +218,6 @@ function ProfileCard() {
 
 function SwitchMenuItems({
   accounts,
-  currentAccount,
   signOutPromptControl,
 }: {
   accounts:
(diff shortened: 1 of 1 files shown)


src/view/shell/desktop/LeftNav.tsx
```diff
@@ -255,6 +296,16 @@ function SwitchMenuItems({
           <Menu.Divider />
         </>
       )}
+      <Menu.Item
+        label={_(msg`Go to profile`)}
+        // @ts-expect-error the function signature differs on web -prf
```
mozzius: I don't think you're Paul R. Frazee 😉

src/view/shell/desktop/LeftNav.tsx
```diff
@@ -226,16 +228,55 @@ function SwitchMenuItems({
...
+      if (isCurrent) {
+        emitSoftReset()
+      } else {
+        const [screen, params] = router.matchPath(profileLink)
+        // @ts-expect-error TODO: type matchPath well enough that it can be plugged into navigation.navigate directly
+        navigation.navigate(screen, params, {pop: true})
+      }
+    },
+    [navigation, profileLink, isCurrent],
+  )
```
mozzius: I think you can drop all of this. `e.preventDefault()` means that the menu doesn't close once you click it

```suggestion
  const onProfilePress = useCallback(() => {
    if (isCurrent) {
      emitSoftReset()
    } else {
      const [screen, params] = router.matchPath(profileLink)
      // @ts-expect-error TODO: type matchPath well enough that it can be plugged into navigation.navigate directly
      navigation.navigate(screen, params, {pop: true})
    }
  }, [navigation, profileLink, isCurrent])
```

mozzius (changes requested):
Much appreciated, just one comment

src/view/shell/desktop/LeftNav.tsx
Inbestigator: The `preventDefault` is because technically it is a link, it has an href. That's why using the modkeys disables the navigation logic

src/view/shell/desktop/LeftNav.tsx
Inbestigator: 🤦‍♂️ I copied the prevention comments lol

src/view/shell/desktop/LeftNav.tsx
Inbestigator: I've pushed a fix that uses the menu context to close it on navigation

mozzius (approved):
thank you!

PR #8868, opened 2025-08-19
Add underage notice to Moderation screen

estrattonbailey:
This is a small bug left over from Age Assurance. For underage users, we hide adult content settings entirely. However, some adults may mistakenly entered the wrong birthdate, which previously left them in a "underage" state with no instructions on how to fix this. This adds a little admonition to the top of the screen for this case.

CleanShot 2025-08-18 at 19 01 31@2x

### DIFF
--- a/src/screens/Moderation/index.tsx
+++ b/src/screens/Moderation/index.tsx
@@ -22,6 +22,7 @@ import {
 import {isNonConfigurableModerationAuthority} from '#/state/session/additional-moderation-authorities'
 import {useSetMinimalShellMode} from '#/state/shell'
 import {atoms as a, useBreakpoints, useTheme, type ViewStyleProp} from '#/alf'
+import {Admonition} from '#/components/Admonition'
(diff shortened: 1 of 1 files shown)


src/screens/Moderation/index.tsx
```diff
@@ -201,6 +202,24 @@ export function ModerationScreenInner({
...
+      {isDeclaredUnderage && (
+        <View style={[a.pb_2xl]}>
+          <Admonition type="tip" style={[a.pb_md]}>
+            <Trans>
+              Your declared age is under 18. Some settings below may be
+              disabled. If this was a mistake, you may edit your bithdate in
+              your{' '}
+              <InlineLinkText
+                to="/settings/account"
+                label={_(msg`Account Settings`)}>
```
surfdude29:
```suggestion
                label={_(msg`Go to account settings`)}>
```
nit: Suggest adjusting label to match one used in similar copy, reducing string duplication:

https://github.com/bluesky-social/social-app/blob/122a46891a7f912c8e2777bae00c4b1f64154257/src/components/ageAssurance/AgeAssuranceAdmonition.tsx#L86-L97

PR #8869, opened 2025-08-19
fix typo

cirx1e:
`bithdate` -> `birthdate`

https://github.com/bluesky-social/social-app/blob/48baf7277df4b447521d372db9ca3f7f3ca7b73b/src/screens/Moderation/index.tsx#L208-L218


surfdude29:
Good catch! 👌

mozzius:
son of a bith...

PR #8872, opened 2025-08-20
Fix default opengraph

estrattonbailey:
This PR does a few things:
1. Includes the author avi on post links
2. Adds `twitter:site` at a top level — we can replace this one day if we ever link user accounts to their Twitter accounts
3. Sets the Twitter image and Twitter desc on posts

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -99,9 +99,6 @@
   <link type="application/activity+json" href="" />
 
   {% block html_head_extra -%}{%- endblock %}
-
-  <meta property="og:image" content="https://bsky.app/static/social-card-default-gradient.png" />
-  <meta property="twitter:image" content="https://bsky.app/static/social-card-default-gradient.png"  />
 </head>
 <body>
 {%- block body_all %}


mozzius (approved):

Gargaj:
This made discord embeds look awful.
image
Semantically `twitter:image` should not be used for the user avatar.

estrattonbailey:
@Gargaj the intention here was to fix cases where posts without media were shared in apps like WhatsApp and iMessage. As you can see in the code, if the post contains media, we use the `summary_large_card` variant. If it does not, and we fall back to the avatar, we use the `summary` card.

This allows posts with media to show the media in a large format, and posts without media to display the avatar in a small format, as visualized in the screenshot. WhatsApp treats this the same way, in our testing. I don't know how Discord handles this, or what attributes they use to determine their display, but the usage of `twitter:image` or `og:image` is correct iiuc.
CleanShot 2025-09-04 at 09 25 18@2x

Gargaj:
@estrattonbailey Let's go to the specs then:

OpenGraph spec (https://ogp.me/):

> og:image - An image URL which should represent your object within the graph.

This depends on the reading, but if the "object" is the post then the user avatar definitely does not represent the post, it represents the user who created the post. The `profile:*` type is where you could put an avatar, but an explicit avatar field is not part of the spec currently.

Twitter docs: (https://developer.x.com/en/docs/x-for-websites/cards/overview/summary-card-with-large-image)

> twitter:image - A URL to a unique image representing the content of the page. **You should not use** a generic image such as your website logo, **author photo**, or other image that spans multiple pages. 

"Author photo" is **explicitly** flagged as something you should NOT use in that field.

mary-ext:
I think it'd be fine to do UA sniffing here, crawlers from chat apps like Discord's have discernable user agents

PR #8874, opened 2025-08-21
Fix Gif Select dialog search bar scrolling away

mozzius:
Extracted out of #8218 

The GifSelect dialog has an issue where the header can scroll away on web. I would really like to get this dialog component to a state where it doesn't need excessive configuration to use OOTB, hopefully this takes it a step closer

# Test plan

Confirm fix worked for gif select dialog

Ensure no regressions in other dialogs (starter pack wizard edit, searchable people list, progress guide find follows)


surfdude29:
I don't use the GifSelect dialog all that often, but when I do I notice this, so I'm glad it's getting fixed :)

It works well on the preview branch for me 👌

PR #8875, opened 2025-08-21
[APP-1406] follow suggestions card to show follow back

caidanw:
Summary
---

When a user already follows you, the profile cards "Follow" button now shows "Follow back" instead. This helps keep the follow button consistent across the app.

In addition, I noticed that since this component is used in a few other places among the app we also have it updated there for free. Such as on the search page under the "Suggested Accounts" section.

Edit: updated all uses of "Follow Back" through out the app (only a handful of spots) to be "Follow back"

Screenshots
---
_Note: I manually swapped the "Follow" and "Follow back" to get these screenshots, as it's tricky to find it naturally._

**Profile header suggested accounts**
image

**Search page suggested accounts section**
image

### DIFF
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -515,7 +515,7 @@ export function FollowButtonInner({
   )
   const followLabel = _(
     msg({
-      message: profile.viewer?.followedBy ? 'Follow back' : 'Follow',
+      message: profile.viewer?.followedBy ? 'Follow Back' : 'Follow',
       comment: 'User is not following this account, click to follow',
     }),
   )
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -515,7 +515,7 @@ export function FollowButtonInner({
   )
   const followLabel = _(
     msg({
-      message: 'Follow',
+      message: profile.viewer?.followedBy ? 'Follow back' : 'Follow',
       comment: 'User is not following this account, click to follow',
     }),
   )
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -515,7 +515,7 @@ export function FollowButtonInner({
   )
   const followLabel = _(
     msg({
-      message: profile.viewer?.followedBy ? 'Follow Back' : 'Follow',
+      message: profile.viewer?.followedBy ? 'Follow back' : 'Follow',
       comment: 'User is not following this account, click to follow',
     }),
   )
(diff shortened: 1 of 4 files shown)


surfdude29:
FYI we currently use `Follow Back` elsewhere on buttons, fwiw I don't have a strong view on which it should be, I just think it should be consistent :)

https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/view/com/profile/FollowButton.tsx#L81

https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/view/com/post-thread/PostThreadFollowBtn.tsx#L129

https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/screens/Profile/Header/ProfileHeaderStandard.tsx#L250

EDIT: Some history in case it's helpful – `Follow Back` was initially added to Post Thread as an experiment in #3439, and was then added to the two other places in #5281.

Although it should be noted that since the move to Crowdin, the fact that an existing string has already been translated is now less important in my view, as the translation memory feature makes it incredibly easy for us translators to reuse an existing translation where there's only been a small change.

caidanw:
> FYI we currently use `Follow Back` elsewhere on buttons, fwiw I don't have a strong view on which it should be, I just think it should be consistent :)
> 
> https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/view/com/profile/FollowButton.tsx#L81
> 
> https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/view/com/post-thread/PostThreadFollowBtn.tsx#L129
> 
> https://github.com/bluesky-social/social-app/blob/eabcd9150d3513988f5b3c47b95a601d5f1bf738/src/screens/Profile/Header/ProfileHeaderStandard.tsx#L250
> 
> EDIT: Some history in case it's helpful – `Follow Back` was initially added to Post Thread as an experiment in #3439, and was then added to the two other places in #5281.
> 
> Although it should be noted that since the move to Crowdin, the fact that an existing string has already been translated is now less important in my view, as the translation memory feature makes it incredibly easy for us translators to reuse an existing string where there's only been a small change.

Thanks for the context and for the catch! I'll update it now!

surfdude29:
Sorry if I caused any confusion with my comment, I think either is fine, it's just consistency that's important.

So fwiw I think the latest commit changing all usages to `Follow back` works great 👍

jimmylee (approved):
LGTM, thanks for doing that.

src/components/ProfileCard.tsx
```diff
@@ -515,7 +515,7 @@ export function FollowButtonInner({
   )
   const followLabel = _(
     msg({
-      message: 'Follow',
+      message: profile.viewer?.followedBy ? 'Follow back' : 'Follow',
```
surfdude29, line 518: @caidanw @estrattonbailey I think this might need to be coded a bit differently, as Lingui seems to be having trouble here:

IMG_9706

surfdude29:
I opened #8880

Issue #8873, opened 2025-08-21
Trending video slider missing an icon


vertigris:
### Steps to Reproduce

1. Go to Discover
2. Scroll to trending video slider
3. Swipe to the end
4. The icon is either missing or wrong as it just appears a blue circle

### Attachments

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/380bf4bb-36a9-4407-9bfa-46dcd660fd9b" />

![Image](https://github.com/user-attachments/assets/a951ac76-e343-4f5c-a7c1-88926676b1a7)

### What platform(s) does this occur on?

iOS

### Device Info

IPhone 12

### What version of the app are you using?

Build version: 1.106.0.1093; Bundle info: 683d7bf (prod); Bundle date: 25081202; Platform: ios; Platform version: 18.5; Anonymous ID: 74c49169-ea65-453a-a69c-221aa755b300

PR #8876, opened 2025-08-21
[APP-1399] trending videos icon missing

caidanw:
Summary
---
The icon was "missing" but really it was just the same color as the background. I've updated the component to use the actual `Button` and `ButtonIcon` instead of a `View`.

During my investigation of the code I realized no one had introduced any changes specifically for that icon (according to git blame). So after poking around a bit I found out that the issue was that the `ButtonIcon` was updated to use these hooks:
```typescript
  const {size: buttonSize} = useButtonContext()
  const textStyles = useSharedButtonTextStyles()
```

Which means that the icon was using the primary color because there was no present button context and thus thought it had no background.

I've also extracted a new `ViewMoreCard` component out of the `VideoCards` component to help clean up the structure and provide a better separation of concerns (`VideoCards` had lots of hooks that were solely for the `ViewMoreCard`.

Additionally I've refactored some of the code, mostly the imports, to be in alignment with the new code guidelines. Replacing `import React from 'react'` with `import {useState, useCallback} from 'react'`.

*Hint for reviewers, turn on the "Hide whitespace" setting in the github diff view to better see the changes that I made (as the refactor shows a lot more than what was actually changed).*

Screenshots
---
Tested on all themes.

image

image

image

### DIFF
--- a/src/components/interstitials/TrendingVideos.tsx
+++ b/src/components/interstitials/TrendingVideos.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect} from 'react'
+import {useCallback, useEffect, useMemo} from 'react'
 import {ScrollView, View} from 'react-native'
 import {AppBskyEmbedVideo, AtUri} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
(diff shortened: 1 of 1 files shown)


surfdude29:
Fixes #8873 👌

PR #8883, opened 2025-08-22
feat: add canonical URLs to improve SEO

caidanw:
## Summary
- Add canonical URLs to home page, post pages, and profile pages to improve SEO and prevent duplicate content issues
- Use existing `requestURI` template variable for dynamic canonical URLs on post and profile pages
- Set static canonical URL for home page

## Changes
- **Home page**: Added static canonical URL pointing to `https://bsky.app`
- **Post pages**: Added dynamic canonical URL using `requestURI` template variable
- **Profile pages**: Added dynamic canonical URL using `requestURI` template variable

### DIFF
--- a/bskyweb/templates/home.html
+++ b/bskyweb/templates/home.html
@@ -12,6 +12,7 @@
 
   <meta property="og:url" content="https://bsky.app" />
   <meta name="twitter:url" content="https://bsky.app" />
+  <link rel="canonical" href="https://bsky.app" />
 
   <meta property="og:image" content="https://bsky.app/static/social-card-default-gradient.png" />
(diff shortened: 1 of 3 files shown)


softplus:
FWIW I don't think this does much with regards to SEO. If you're setting the requestURI as canonical, it's basically a no-op. If you want to set a rel=canonical, I'd recommend cleaning the URL and always using a cleaned URL instead (at least, remove any URL parameters). That would help clean things up like https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fbsky.app%2Fprofile%2F+inurl%3Autm_source

jimmylee:
Thanks John, that is a useful suggestion to make sure that the canonical URL is the clean and preferred URL, and we can ensure removal of any unnecessary query parameters.

PR #8885, opened 2025-08-23
feat: prevent search engines from using Discover feed content as page descriptions

caidanw:
## Summary

- ~~Adds NoSnippetWrapper component that conditionally applies `data-nosnippet` attribute to prevent search engines from indexing content as page descriptions~~
- ~~Wraps Discover feed content with NoSnippetWrapper to solve SEO issue where Google was displaying dynamic feed content as the website description in search results~~
- Add the `data-nosnippet` attribute to the Discover feed

## Technical Details

- ~~**NoSnippetWrapper component**: Uses `unstable_createElement` from `react-native-web` to create DOM elements with data attributes (React Native View components don't support `data-*` attributes)~~
- ~~**Platform handling**: Only applies wrapper on web platform, returns unwrapped children on native~~
- **Feed detection**: Uses existing `DISCOVER_FEED_URI` constant to identify when to apply the data attribute
- **Google compliance**: Uses `data-nosnippet` format per Google's documentation for preventing snippet indexing

## Testing

- TypeScript compilation passes
- Linting passes
- Confirmed DOM shows `data-nosnippet` attribute on Discover feed pages
- Verified no impact on non-Discover feeds or native platforms


mozzius (changes requested):
There's an easier way to add data attributes - it's not typed, but you can use the `dataSet` prop. no need to faff with the unstable_createElement stuff :)

```tsx
return (
    <View
      testID={testID}
      // @ts-expect-error web only -sfn
      dataSet={{nosnippet: ''}}>
         // ...
```

caidanw:
> There's an easier way to add data attributes - it's not typed, but you can use the `dataSet` prop. no need to faff with the unstable_createElement stuff :)
> 
> ```tsx
> return (
>     <View
>       testID={testID}
>       // @ts-expect-error web only -sfn
>       dataSet={{nosnippet: ''}}>
>          // ...
> ```

Oh man haha thanks, I searched the docs and web and couldn't find any native way to add it. I'll update now.

jimmylee (approved):
LGTM, landing this to give it a shot, thanks @caidanw.

src/view/com/feeds/FeedPage.tsx
```diff
@@ -73,6 +77,10 @@ export function FeedPage({
     return isNative && _isVideoFeed
   }, [feedInfo])
 
+  const isDiscoverFeed = useMemo(() => {
+    return feedInfo.uri === DISCOVER_FEED_URI
+  }, [feedInfo.uri])
+
```
mozzius: `useMemo` is massively overkill for a string comparison lol

mozzius (approved):
Confirmed working 👍

Screenshot 2025-08-27 at 12 03 17

src/view/com/feeds/FeedPage.tsx
caidanw: haha true, was following the existing pattern as the `isVideoFeed` but I'll update now

caidanw: done!

jimmylee (approved):
LGTM!!

src/view/com/feeds/FeedPage.tsx
mozzius: thanks! yeah, that other check is probably also overkill too haha

PR #8886, opened 2025-08-23
feat: don't retain accepted language suggestion after finishing or exiting post

elijaharita:
For https://linear.app/blueskyweb/issue/APP-1400/post-specific-language-selection

This makes language suggestions temporary. After the pending post is published or cancelled, the value will no longer be saved to langPrefs. Only manually set post language will be saved to langPrefs.

### DIFF
--- a/src/view/com/composer/select-language/PostLanguageSelectDialog.tsx
+++ b/src/view/com/composer/select-language/PostLanguageSelectDialog.tsx
@@ -28,7 +28,7 @@ export function PostLanguageSelectDialog({
   currentLanguages,
   control,
 }: {
-  currentLanguages?: string[],
+  currentLanguages?: string[]
   control: Dialog.DialogControlProps
 }) {
   const {height} = useWindowDimensions()
@@ -51,11 +51,7 @@ export function PostLanguageSelectDialog({
   )
 }
 
-export function DialogInner({
-  currentLanguages,
-}: {
-  currentLanguages?: string[]
-}) {
+export function DialogInner({currentLanguages}: {currentLanguages?: string[]}) {
   const control = Dialog.useDialogContext()
   const [headerHeight, setHeaderHeight] = useState(0)
 


src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -94,7 +90,7 @@ export function SuggestedLanguage({
           color="secondary"
           size="small"
           variant="solid"
-          onPress={() => setLangPrefs.setPostLanguage(suggestedLanguage)}
```
elijaharita, line 99: instead of changing langPrefs, suggested language now just forwards the new value to a change handler, where it can be stored in a state variable

jimmylee (approved):
LGTM, I'd like @haileyok to help me take a look at this as well.

src/view/com/composer/Composer.tsx
```diff
@@ -194,6 +195,19 @@ export const ComposePost = ({
...
+
+  // NOTE(@elijaharita): if a temporary language suggestion has been accepted,
+  // show that as the post language instead of the one from langPrefs.
+  const currentLanguages = useMemo(
+    () =>
+      acceptedLanguageSuggestion
+        ? [acceptedLanguageSuggestion]
+        : toPostLanguages(langPrefs.postLanguage),
+    [acceptedLanguageSuggestion, langPrefs.postLanguage],
+  )
```
estrattonbailey, line 212: Once I opt to accept a language, I can't change it to anything other than the suggestion. Should probably clear this value if the user chooses something else.

Also, once a suggestion is accepted, the LangSelectBtn doesn't reflect that selected state.

estrattonbailey (changes requested):
Nice solution 👍 almost there. Mind rebasing while you're in here? We merged a bug fix for web the other day.

src/view/com/composer/Composer.tsx
elijaharita: interesting, it works fine for me to change the language after accepting the suggestion. which platform are you experiencing this on?

elijaharita: i'm also not having the issue you're experiencing with the SelectLangBtn not updating.

i'll investigate and see if i can reproduce though.

elijaharita: oops false alarm on failing repro, i was on wrong branch haha. retrying

elijaharita: ok nice was able to repro and pushed a fix!

estrattonbailey:
@elijaharita ah shoot, ya know what, #8914 introduced the new lang selector, so this needs to integrate with the new one 🙃 my bad, I knew that was coming down the pipe and didn't connect the dots

elijaharita:
haha no worries, i also know it was coming, i helped out a lot with that one! wasn't sure which one would merge first. the rebase will be pretty easy.

estrattonbailey:
We needed to hotfix something in main this morning, need to rebase this again in order to test it out. Gonna merge into my base branch and test it there before merging. Thanks for this!

PR #8889, opened 2025-08-23
Update origin

estrattonbailey:


WinterPhoenix:
Can I just say, I’m not a fan of these commit messages? It’s not even tied to a feature or anything, it’s just “Debug” or “Update origin”.

Debug what? Update the origin of what? It’s going to make finding these commits, which are related to censorship enforced by the official app, difficult in the future.

PR #8839, opened 2025-08-13
Integrate Sonner for toasts

estrattonbailey:
Integrates `sonner` and `sonner-native` for our toasts.

https://github.com/user-attachments/assets/9ce45907-1708-40b0-8e60-25aa01f0a302

Android looks exactly the same. Real W for this library tbh.
https://github.com/user-attachments/assets/21615c80-98e6-412f-9de8-7b925f567ee2

PR #8897, opened 2025-08-25
Fix language popup not appearing

mozzius:
#8828 added `<RootSiblingParent>` to the composer (two in fact??) in order to try and get toasts to work in the composer.

However this broke the language popup on web - locally I see this and on main.bsky.dev it just shows nothing

Screenshot 2025-08-25 at 21 02 49

Luckily this is not necessary anymore thanks to #8839 so we can just remove them

## After

Screenshot 2025-08-25 at 21 05 33


mozzius:
all good!

Issue #6594, opened 2024-11-21
No option to opt-out of Tracking


celenityy:
I made a thread [here](https://bsky.app/profile/celenity.dev/post/3lbdcrtz4q22o) going into more detail, *(which I would recommend reading for some context)*, but TL;DR:

* `bsky.app` & Bluesky's mobile apps _(which this repo pertains to)_, contain tracking code from Sentry & Statsig.

* `bsky.social` contains tracking code from Google.

# **There is currently no option to opt out of or disable this functionality.**

**My proposal is as follows:**

* Remove Google tracking from `bsky.social`, **period.**

* Ideally, remove Sentry & Statsig from `bsky.app` as well; but as a compromise, at least **add some kind of setting or toggle to disable it.**

# Why?

* **Privacy is a fundamental human right.** Sharing data with third parties who can (and in Google's case, will **certainly**) use that data to profile & track users poses clear privacy concerns.

* To quote [Bluesky's website](https://bsky.social/about):

> **Bluesky is social media as it should be.** Find your community among millions of users, unleash your creativity, and have some fun again.

Evidently, unwanted data collection & tracking is **not** how social media should be. Among other reasons, **it is exactly why myself and millions of other people have left platforms like Facebook & Twitter/X in the first place.**

* Another quote from Bluesky's website:

> **Control your scroll**
>
> **Use powerful moderation lists, content filters, and more to control what ends up in your timeline.** In the future, people and organizations will be able to use open tools to contribute even more directly to moderation.

**If Bluesky is designed to give control back to users, why would that not extend to other areas, such as privacy & data collection?**

# Additionally:

* IANAL, so I won't go too deep here, but this could *potentially* raises GDPR concerns, as this data is being collected 1: without **clear** user consent, 2: without a way to toggle or opt out, and 3: is **not** necessary for Bluesky to function.

* _(With the obvious exception of Threads)_, other widely adopted open social media platforms similar to Bluesky, such as Mastodon and Nostr, **have no tracking or data collection, period.** This does not reflect well on Bluesky when compared to alternatives, and it **will** hurt adoption from users.

* **Above all else, it is simply the right thing to do.**

# Conclusion:

[Rose from Bluesky has made the claim that Bluesky won't sell user data](https://bsky.app/profile/rose.bsky.team/post/3l7biwvrpbo23). I find this stance odd and think it may be worth reconsidering. After all, if you're just going to hand out data to third party companies, why not get something out of it??? 😬 *(/s)*

[...]

PR #8898, opened 2025-08-25
Add value proposition screen experiment

mozzius:
## Stacked on #8847 

https://github.com/user-attachments/assets/04b3d8ef-4819-46c4-8676-fb47af12c6f7

# Test plan

Test different platforms
Check they look like designs
Ensure other side of gate still works


LoganDupont:
Can we also provide something so people actually update their display name? Most people search users based on their name but because most people don't update their display name they can't be found in the search results.

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -245,6 +253,249 @@ export function StepFinished() {
...
+        msg`Ditch the trolls and clickbait. Find real people and conversations that matter to you.`,
+      ),
+      alt: _(
+        msg`Your profile picture surrounded by concentric circles of other users' profile pictures`,
+      ),
+    },
+    {
+      title: _(msg`Forget the noise`),
+      description: _(
+        msg`No ads, no tracking, no engagement traps. Bluesky respects your time and attention.`,
```
surfdude29:
```suggestion
        msg`No ads, no invasive tracking, no engagement traps. Bluesky respects your time and attention.`,
```
This seems like a great experiment to try!

I have a little concern about the wording here though. I personally don't have a problem with the metrics and data that the app collects, but it can reasonably be interpreted as tracking (e.g. see #6594). So I understand that the intention here is to communicate "no ad-related tracking that tracks you all over the web and invades your privacy, like Facebook etc.", but I feel like simply claiming that the app has `no tracking` might be overstating things a bit.

And maybe I'm being over cautious, but I think that claiming "no tracking" in a highly visible place (especially if the experiment becomes permanent), while the app collects quite a lot of metrics and data, could potentially open PBC up to regulatory issues (and maybe bad PR if a journalist thinks it's misleading).

I think if you qualified the claim by saying something like `no invasive tracking` it would be clearer. However, I'm certainly no lawyer or regulatory policy expert :)

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -245,6 +253,249 @@ export function StepFinished() {
...
+            style={[a.bg_transparent]}>
+            <ButtonText>
+              <Trans>Skip</Trans>
+            </ButtonText>
+          </Button>
+        </OnboardingHeaderSlot.Portal>
+      )}
+
+      <LayoutAnimationConfig skipEntering skipExiting>
+        <Animated.View
```
estrattonbailey, line 386: Let's remove pointer events here. It looks like something that should be swipeable, but it's not, and you end up dragging the image on web.

src/screens/Onboarding/StepFinished.tsx
```diff
@@ -245,6 +253,249 @@ export function StepFinished() {
...
+          key={subStep}
+          entering={native(
+            SlideInRight.easing(Easing.out(Easing.exp)).duration(500),
+          )}
```
estrattonbailey, line 390: If you wanna make this work on web, could use Legend's Motion lib 🤔

PR #8900, opened 2025-08-25
feat: implement canonical URL filter to clean UTM parameters for SEO

caidanw:
## Summary

- Implement custom Pongo2 filter to remove query parameters and fragments from canonical URLs
- Update post and profile templates to use cleaned canonical URLs while preserving original URLs for Open Graph metadata
- Add comprehensive test coverage for URL cleaning scenarios

## Problem Addressed

Community feedback indicated that canonical URLs containing UTM parameters and query strings violate SEO best practices. Different URLs with tracking parameters should point to the same canonical version for proper search engine indexing.

## Solution

Created a `canonical` filter that:
- Removes query parameters (`?utm_source=twitter`, etc.)
- Removes URL fragments (`#section`)
- Gracefully handles malformed URLs by returning the original
- Preserves full URLs in `og:url` for social media analytics

## Changes

- **New**: `bskyweb/cmd/bskyweb/filters.go` - Custom Pongo2 filter implementation
- **New**: `bskyweb/cmd/bskyweb/filters_test.go` - Comprehensive test coverage
- **Modified**: `bskyweb/templates/post.html` - Apply canonical filter to canonical link
- **Modified**: `bskyweb/templates/profile.html` - Apply canonical filter to canonical link

## Testing

- ✅ All unit tests pass (6 test scenarios)
- ✅ Build verification successful
- ✅ No regressions in existing functionality
- ✅ Filter registration confirmed via init() function

## Before/After

**Before:**
```html
<meta property="og:url" content="https://bsky.app/profile/user?utm_source=twitter">
<link rel="canonical" href="https://bsky.app/profile/user?utm_source=twitter" />
```

**After:**
```html
<meta property="og:url" content="https://bsky.app/profile/user?utm_source=twitter">
<link rel="canonical" href="https://bsky.app/profile/user" />
```

This implementation addresses community SEO concerns while maintaining full functionality for social media sharing and analytics tracking.


bskyweb/cmd/bskyweb/filters.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"net/url"
+
+	"github.com/flosch/pongo2/v6"
+)
+
+func init() {
+	pongo2.RegisterFilter("canonical", filterCanonical)
```
bnewbold: I would rename this to `canonicalurl` to make it a bit clearer what it is doing.

(the pongo2 filter casing norm seems to be lower-case with words concatenated)

caidanw: Good call out, I was thinking something similar but wanted to get feedback on the naming.

caidanw: I've updated the filter name to be `canonicalize_url`. I took a look at the built-in filters and this felt like the most appropriate name.

mary-ext:
fyi Chrome Android grabs the canonical URL when you attempt to share the page (but not when copying from address bar), so there could be some complaints about the long URLs if you canonicalize it to DID

caidanw:
> fyi Chrome Android grabs the canonical URL when you attempt to share the page (but not when copying from address bar), so there could be some complaints about the long URLs if you canonicalize it to DID

Thanks for the info, didn't know Chrome Android would do this. Luckily we aren't changing the urls, so if the requested url is using the username then it'll keep the username. Which I believe the client should always redirect using the username over DID. I could be wrong though.

jimmylee (approved):
LGTM, thanks for doing this @caidanw

PR #8902, opened 2025-08-26
Threads v2 cleanup

estrattonbailey:
Deletes the gate and unused components, and re-homes the couple old components that were untouched in the V2 project to live alongside the new thread components. Required swapping out the old/new components in DebugMod, no functional change.


mozzius (approved):
lgtm

PR #8904, opened 2025-08-26
Add disk build cache provider

mozzius:
This video says it all.

https://github.com/user-attachments/assets/e27e0a9f-63f5-4da9-8997-832d7cbef840

We can also configure a remote build cache provider, like EAS or github actions, so that we can share build outputs between us which would be crazy. Disk cache is free and fast and easy so let's do this first

Builds are stored in `node_modules` so you can use `yarn nuke` to clear. Works via fingerprints

### DIFF
--- a/package.json
+++ b/package.json
@@ -282,7 +282,6 @@
     "**/@expo/image-utils": "0.8.7",
     "**/expo-constants": "18.0.8",
     "**/expo-device": "7.1.4",
-    "**/zod": "3.23.8",
     "**/multiformats": "9.9.0"
   },
   "jest": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -286,7 +287,6 @@
     "@types/react": "^18",
     "**/expo-constants": "17.0.3",
     "**/expo-device": "7.1.4",
-    "**/zod": "3.23.8",
```
mozzius: Was resolving both Zod 3 and Zod 4 to 3 - we can safely remove this as all the Zod 3s resolve to 3.23.8 anyway:

Screenshot 2025-08-26 at 16 07 54

PR #8909, opened 2025-08-26
Fix toast type

estrattonbailey:
I realized that the simpler way to do this is just to _not_ allow passing in `type` manually, and instead relying on a single source of truth: context.

Also, the only real use case for the previous `Default` export was the Storybook, and it's trivial to create that use case ad hoc in the Storybook itself. That'll help prevent us from using that directly, since the preferred API is `toast.show()`.

### DIFF
--- a/src/components/Toast/Toast.tsx
+++ b/src/components/Toast/Toast.tsx
@@ -115,7 +115,7 @@ export function Text({children}: {children: React.ReactNode}) {
 
 export function Action(
   props: Omit<ButtonProps, UninheritableButtonProps | 'children'> & {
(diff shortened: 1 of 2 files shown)


surfdude29:
Not directly related to this PR, but while you're updating the usage in the Composer would it be possible to mark the `View` string for translation please? :)
https://github.com/bluesky-social/social-app/blob/d19e8c702ec2342868cafafadc164c328ea03b12/src/view/com/composer/Composer.tsx#L545

surfdude29:
Thanks @estrattonbailey 🙏

PR #8910, opened 2025-08-26
Update birthday setting dialog

BinaryFiddler:
https://linear.app/blueskyweb/issue/APP-1402/update-birthdate-dialog-wvalidation-and-copy

- Set max width of the dialog to 400px on the web
- Updating the text and spacing inline with the UpdateEmail dialog
- Default to 18 years ago if the birthday is not set
- Show warning for users between 13-18
- Show error for users under 13, disable saving button

### Web
Screenshot 2025-08-26 at 2 32 56 PM
Screenshot 2025-08-26 at 2 31 02 PM

### Mobile
Simulator Screenshot - iPhone 15 Pro - 2025-08-26 at 14 43 16
Simulator Screenshot - iPhone 15 Pro - 2025-08-26 at 14 40 47

### DIFF
--- a/src/components/dialogs/BirthDateSettings.tsx
+++ b/src/components/dialogs/BirthDateSettings.tsx
@@ -117,7 +117,7 @@ function BirthdayInner({
         />
       </View>
 
-      {isUnder18 && (
+      {isUnder18 && hasChanged && (
         <Admonition type="info">
           <Trans>
             The birthdate you've entered means you are under 18 years old.
--- a/src/components/dialogs/BirthDateSettings.tsx
+++ b/src/components/dialogs/BirthDateSettings.tsx
@@ -4,21 +4,23 @@ import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 
 import {cleanError} from '#/lib/strings/errors'
-import {getDateAgo} from '#/lib/strings/time'
+import {getAge, getDateAgo} from '#/lib/strings/time'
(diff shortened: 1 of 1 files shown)


BinaryFiddler:
Screenshot 2025-08-27 at 1 49 37 PM

Per Eric's suggestion, changed the admonition level from warning to info for users under 18

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -102,10 +114,32 @@ function BirthdayInner({
           onChangeDate={newDate => setDate(new Date(newDate))}
           label={_(msg`Birthday`)}
           accessibilityHint={_(msg`Enter your birth date`)}
-          maximumDate={getDateAgo(13)}
         />
       </View>
 
+      {isUnder18 && (
```
estrattonbailey: Didn't think about this until I tested it, but I think if the user already has set their bday to be <18, we don't need to show the admonition. But if they edit the date to be <18, then we should show it. Tested this locally 👍 
```suggestion
      {isUnder18 && hasChanged && (
```

estrattonbailey (approved):
LGTM, thanks! Just left the one suggestion, what do you think?

src/components/dialogs/BirthDateSettings.tsx
BinaryFiddler: yeah, i agree. totally makes sense.

PR #8917, opened 2025-08-27
safelink

haileyok:


bskylink/locales/fr.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Potentially Dangerous Link": "Lien Potentiellement Dangereux",
+  "Blocked Link": "Lien Bloqué",
+  "This link may be malicious. You should proceed at your own risk.": "Ce lien peut être malveillant. Vous devriez procéder à vos propres risques.",
+  "This link has been identified as malicious and has blocked for your safety.": "Ce lien a été identifié comme malveillant et a été bloqué pour votre sécurité.",
+  "Continue Anyway": "Continuer Quand Même",
+  "Return to Bluesky": "Retourner à Bluesky"
+}
```
Signez: Here are some better translations in French 🥖

```suggestion
{
  "Potentially Dangerous Link": "Lien potentiellement dangereux",
  "Blocked Link": "Lien bloqué",
  "This link may be malicious. You should proceed at your own risk.": "Ce lien est peut-être malveillant. Ne continuez qu’à vos risques et périls.",
  "This link has been identified as malicious and has blocked for your safety.": "Ce lien a été identifié comme malveillant et a été bloqué pour votre sécurité.",
  "Continue Anyway": "Continuer quand même",
  "Return to Bluesky": "Retourner sur Bluesky"
}
```

surfdude29: I'm not a native French speaker like @Signez, but his translations seem spot-on to me 👌

bskylink/src/routes/createShortLink.ts
```diff
@@ -83,8 +83,9 @@ const getUrl = (ctx: AppContext, req: Request, id: string) => {
         : `https://${req.headers.host}`
     return `${baseUrl}/${id}`
   }
-  const baseUrl = ctx.cfg.service.hostnames.includes(req.headers.host)
-    ? `https://${req.headers.host}`
+  const host = req.headers.host ?? ''
+  const baseUrl = ctx.cfg.service.hostnames.includes(host)
```
ghost: Depending on what's the access load here, you could store the hostnames as a set for faster lookup.

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+        )
+        await this.setCursor(res.data.cursor)
+      }
+      setTimeout(() => this.runFetchEvents(), SAFELINK_MIN_FETCH_INTERVAL)
+    }
+  }
+
+  private async getCursor() {
+    if (this.cursor === '') {
+      // TODO: catch err
```
ghost: Did you forget to do this one, or you want to keep it with the TODO?

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+      return
+    }
+
+    if (res.data.events.length === 0) {
+      redirectLogger.info('received no new safelink events from ozone')
+      setTimeout(() => this.runFetchEvents(), SAFELINK_MAX_FETCH_INTERVAL)
+    } else {
+      await this.db.transaction(async db => {
+        for (const rule of res.data.events) {
+          if (rule.eventType === 'removeRule') {
```
ghost: Maybe better to do a switch here? With a default of noop, just logging. In case we want to expand the event types, it wouldn't cause errors when trying to add when it shouldn't.

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+        .execute()
+      this.cursor = cursor
+    } catch (err) {
+      redirectLogger.error({error: err}, 'failed to update safelink cursor')
+    }
+  }
+
+  private static normalizeUrl(input: string) {
+    if (!input.startsWith('https://')) {
+      input = `https://${input}`
```
ghost, line 283: This assumes we either get no protocol at all, or https. Is it possible that we get `http://`? Then this logic would add `https` on top of `http`.

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  constructor(pdsHost: string, identifier: string, password: string) {
+    this.identifier = identifier
+    this.password = password
+
+    this.session = new CredentialSession(new URL(pdsHost))
+    this.agent = new AtpAgent(this.session)
+  }
+
+  public async getAgent(): Promise<AtpAgent> {
```
foysalit: this looks good but wondering how reliable the 1hr refresh token approach is? could we make a low latency request like getSession() and catch 401 error from that call to log in more reliably?

bskylink/src/db/schema.ts
```diff
@@ -14,4 +16,38 @@ export enum LinkType {
   StarterPack = 1,
 }
 
+export type RuleEventType =
```
foysalit: this isn't auto generated, right? could we use values exported from lexicons here? or does that add extra dep?

bskylink/src/routes/createShortLink.ts
```diff
@@ -94,7 +95,7 @@ const normalizedPathFromParts = (parts: string[]): string => {
     '/' +
     parts
       .map(encodeURIComponent)
-      .map(part => part.replaceAll('%3A', ':')) // preserve colons
+      .map(part => part.replace(/%3A/g, ':')) // preserve colons
```
foysalit, line 100: with string manipulation and concatenation like this, i like to add an example in comments so that we can easily visualize for thwat kind of input, what kind of output is expected.

bskylink/src/bin.ts
```diff
@@ -11,7 +10,13 @@ async function main() {
     await migrateDb.migrateToLatestOrThrow()
     await migrateDb.close()
   }
+
   const link = await LinkService.create(cfg)
+
+  if (link.ctx.cfg.service.safelinkEnabled) {
+    link.ctx.safelinkClient.runFetchEvents()
```
foysalit, line 17: do we need an `await` here?

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    if (res.data.events.length === 0) {
+      redirectLogger.info('received no new safelink events from ozone')
+      setTimeout(() => this.runFetchEvents(), SAFELINK_MAX_FETCH_INTERVAL)
+    } else {
+      await this.db.transaction(async db => {
+        for (const rule of res.data.events) {
+          if (rule.eventType === 'removeRule') {
+            await this.removeRule(db, rule)
+          } else {
```
foysalit: do we need a handler for `updateRule` ?

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+    }
+  }
+
+  private async getCursor() {
+    if (this.cursor === '') {
+      // TODO: catch err
+      const res = await this.db.db
+        .selectFrom('safelink_cursor')
+        .selectAll()
+        .orderBy('createdAt desc')
```
foysalit: this syntax may work but i've mostly used `.orderBy('createdAt', 'desc')` style

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+        return ''
+      }
+      this.cursor = res.cursor
+    }
+    return this.cursor
+  }
+
+  private async setCursor(cursor: string) {
+    try {
+      await this.db.db
```
foysalit, line 266: could we not just keep 1 row for cursors and just update the same row? feels a bit redundant to insert cursor values for every page.

bskylink/src/cache/safelinkClient.ts
```diff
@@ -0,0 +1,317 @@
...
+      return 'ok'
+    }
+
+    const urlRule = this.urlCache.get(url)
+    if (urlRule && urlRule !== 'ok') {
+      return urlRule
+    }
+
+    const domainRule = this.domainCache.get(domain)
+    if (domainRule) {
```
foysalit, line 73: a bit awkward that we're checking `urlRule !== 'ok'` but not for domainRule, is this expected?

bskylink/src/db/migrations/002-safelink.ts
```diff
@@ -0,0 +1,37 @@
+import {type Kysely, sql} from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
```
foysalit, line 4: a couple notes here, having designed the ozone side of the schema, it feels like this table is combining the structure of 2 tables from ozone which is why it looks fairly different so ignore if some of this doesn't apply

- we probably should have a unique index on pattern and url columns to avoid multiple inserts for the same rule where we could end up with multiple rules for same pattern and url but with different action.
- technically, we could also make the pattern and url columns the composite primary key for the table and not require id but i personally like having auto incrementing ids so leaving it is fine.
- most likely wont be applicable today but with larger data, we may want to drop stuff like comment column or createdBy since all of this is stored on ozone and this app doesn't need them IMO

bskylink/src/bin.ts
haileyok: this just runs periodically in the background at an interval, so its fine to continue onward here

bskylink/src/db/schema.ts
haileyok: oh yea we could try to do that. i think this was here before the api was available probably?

bskylink/src/db/migrations/002-safelink.ts
haileyok: i'm not really opposed to how we want to do it here, but i'll offer reasoning from my side

- i kind of like that you'll get a history of updates in the db rather than updating the existing one. my assumption is that there should only ever be a single "active" rule for a specific pattern and url, right? in that case, my thought was to fetch the latest one from the DB every time, while maintaining the history of the changes. of course, it's not totally necessary and we could just upsert if it exists
- yea agree, i also like to have the incremented id so if we're fine with that i'll leave it be
- definitely fine with dropping those off

bskylink/src/cache/safelinkClient.ts
haileyok: so, here we want to fall through to the domain rule check even if `urlRule` is `ok`, since we don't know if the domain rule is blocking but the URL rule is fine. if `domainRule` is `ok` though, we don't have to fall through further since there isn't any other ruleset to check.

i'll stick a comment in here to explain that

bskylink/src/cache/safelinkClient.ts
haileyok: i'd just handle both `addRule` and `updateRule` inside the `this.addRule()` handler. either A. just by inserting a new column every time or B. by upserting the existing unique key (as discussed above, either works)

(made this a switch so it's a little more clear)

bskylink/src/cache/safelinkClient.ts
haileyok: ah no, can rm. went with just returning an empty `''` cursor rather than throwing with `executeTakeFirstOrThrow()`

bskylink/src/routes/createShortLink.ts
haileyok: i actually have no idea what the load is here, but agree good idea

bskylink/src/cache/safelinkClient.ts
haileyok: yea that seems good. i added a check for `ExpiredTokenError` which will attempt to refresh the token then try the request again...that assumes our agent will throw that error when it's expired, which I assume it does?

bskylink/src/db/migrations/002-safelink.ts
foysalit: won't oppose on keeping history but ozone is already keeping the change history and if things go south, we would probably want to debug one source instead of trying to match diff between 2 different tables/databases? again, not a blocker, just wanted to note.
sounds good on the rest of it!

bskylink/src/cache/safelinkClient.ts
foysalit: oh right ok that sounds good!

bskylink/src/cache/safelinkClient.ts
foysalit: i think so? not super sure though lol.

bskylink/src/db/migrations/002-safelink.ts
haileyok: oh yea yea, for sure. in here, i just want to keep the history so that there's a log of "we actually fetched these events from ozone", rather than use it as an audit log for updates that people take or whatever. i'd actually be fine dropping it later even, really its just a "check here in case of bugs" thing

PR #8918, opened 2025-08-27
Upgrade `@types/react` to 19 and run codemod

mozzius:
#8380 got stale, so reran the codemods off of main

# test plan

confirm ci passes


package.json
```diff
@@ -212,7 +212,7 @@
     "react-native-web-webview": "^1.0.2",
     "react-native-webview": "^13.13.5",
     "react-remove-scroll-bar": "^2.3.8",
-    "react-responsive": "^9.0.2",
+    "react-responsive": "^10.0.1",
```
mozzius, line 216: for some reason `@types/react` wasn't in dev deps - it was a transitive dependency from `@types/react-responsive@8.0.5`, which we were then using `resolutions` to force to be the right version. worst part is `react-responsive` includes types from `v9` onwards so we don't even need `@types/react-responsive`

estrattonbailey (approved):

PR #8921, opened 2025-08-27
Update dev env

estrattonbailey:


mozzius (approved):
seems to work as far as I can tell

PR #8928, opened 2025-08-28
Fall back to v16 picker

estrattonbailey:
Reverts back `expo-image-picker@16` and re-adds the patch we had prior to #8828. Should be no functional changes.


mozzius:
Closing this, we can reopen if needed

PR #8933, opened 2025-08-29
[APP-1418] - Create testbed for embed service

APiligrim:
This PR creates a testbed for embed service to test different states of embeds for [this ticket](https://linear.app/blueskyweb/issue/APP-1418/create-testbed-for-embed-service). 

https://github.com/user-attachments/assets/0b4a3fbd-75a0-4f91-a84d-560f0accabae

**Screenshots**
Screenshot 2025-08-29 at 5 09 29 PM
Screenshot 2025-08-29 at 5 13 58 PM

### DIFF
--- a/src/locale/locales/en/messages.po
+++ b/src/locale/locales/en/messages.po
@@ -2805,10 +2805,6 @@ msgstr ""
 msgid "Dismiss this section"
 msgstr ""
 
-#: src/components/Toast/index.web.tsx:81
-msgid "Dismiss toast"
-msgstr ""
-
 #: src/screens/Settings/AccessibilitySettings.tsx:69
 #: src/screens/Settings/AccessibilitySettings.tsx:74
 msgid "Display larger alt text badges"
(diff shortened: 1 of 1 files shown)


bskyembed/test.html
```diff
@@ -0,0 +1,1252 @@
...
+              >@test-all-embeds-b.bsky.social</a
+            >)
+            <a
+              href="https://bsky.app/profile/did:plc:yaqf6zpqlko6hkm5zukoilix/post/3lkcj4xgmdc2l?ref_src=embed"
+              >Mar 13, 2025 at 6:58 PM</a
+            >
+          </blockquote>
+          <script
+            async
+            src="https://embed.bsky.app/static/embed.js"
```
mozzius: this uses the production code. it would be much more helpful if we could see the output of the dev server instead

mozzius:
Seems there was an image committed by accident? Also, no need to run the `intl:extract` script, that happens automatically

jimmylee (approved):
LGTM, this looks like we can finally test Bluesky embeds. I'll wait for @mozzius to also give a thumbs up just in case I missed something.

mozzius:
Close, but it points to a locally running `bskyweb` instance, whereas ideally it's pointing at the vite devserver.

I managed to get that working in #9115, so I will close this PR.

PR #8942, opened 2025-08-29
Reduce checkmark size in thread anchor

mozzius:
It was looking a little large at 16px. `ProfileCard` has the same font size and uses 14px, so switched it to that. It's the only case that uses `md`

# Before

Screenshot 2025-08-29 at 23 31 10

# After

Screenshot 2025-08-29 at 23 30 42


estrattonbailey:
Let's just test this on all platforms to be sure

mozzius:
Tested android

PR #8944, opened 2025-08-30
Welcome modal on logged-out homepage

abenzer:
Shows a soft welcome modal to logged-out visitors on the homepage.


src/components/WelcomeModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+    ...a.fixed,
+    ...a.inset_0,
+    ...a.justify_center,
+    ...a.align_center,
+    zIndex: 9999,
+    backgroundColor: 'rgba(0,0,0,0.2)',
+    ...web({
+      backdropFilter: 'blur(15px)',
+    }),
+  },
```
mozzius: rather than using `StyleSheet`, we usually just inline the styles like this:

```jsx
<View style={[a.fixed, a.inset_0, a.justify_center, a.align_center, a.z_50, {backgroundColor: 'rgba(0,0,0,0.2)'} />
```

src/components/WelcomeModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+        toValue: 0,
+        duration: 100,
+        useNativeDriver: true,
+      }).start(() => {
+        control.close()
+        if (callback) callback()
+      })
+    },
+    [fadeAnim, control],
+  )
```
mozzius: We typically avoid React Native's `Animated` API - since this is web only, we can just use CSS

src/components/hooks/useWelcomeModal.ts
```diff
@@ -0,0 +1,41 @@
...
+  useEffect(() => {
+    // Only show modal if:
+    // 1. User is not logged in
+    // 2. We haven't shown it yet in this session
+    // 3. We're on the web (this is a web-only feature)
+    // 4. We're on the homepage (path is '/' or '/home')
+    if (
+      isWeb &&
+      !hasSession &&
+      sessionStorage.getItem('welcomeModalShown') !== 'true' &&
```
mozzius: decided to use the SessionStorage API so it doesn't immediately pop up again if you reload. However it will show again in a new tab

mozzius (approved):
Looks good! Looking forward to seeing how the experiment performs

Pushed up some tweaks:
- aligned code with our typical style (i.e. inlined styles)
- replaced JS animation with CSS one
- fix a11y (add focus guards/focus scope, add `aria-modal` etc)

abenzer:
@mozzius thanks for the tweaks! i made a few more:

1. fixed a mismatch in text color between the "Sign in" and "Explore the app" button
2. reduced the gap between the two buttons to match the Figma design
3. updated your logic of only showing the modal once per session → showing the modal until the user closes it

PR #8947, opened 2025-09-01
[Perf] Fix drawer jank issue (maybe)

mozzius:
The cause of the drawer jank is due to our use of the drawer gesture, as we use it to set `requireExternalGestureToFail` on the main home screen pager.

Unfortunately, the drawer gesture is in a `useMemo` that depends on the open/close state of the drawer, so currently it causes a rerender of the `GestureDetector` around the pager view, which then rerenders the pager and all of it's contents.

See the flamegraph here, each time the drawer opens/closes there's a 25ms commit which leads to jank

Screenshot 2025-09-01 at 21 14 50

One fix is to keep the gesture we pass to the gesture detector the same, and add the drawer gesture to it using an effect.

```tsx
function DrawerGestureRequireFail({children}: {children: React.ReactNode}) {
  const drawerGesture = useContext(DrawerGestureContext)

  const nativeGesture = useMemo(() => Gesture.Native(), [])

  useEffect(() => {
    if (drawerGesture) {
      nativeGesture.requireExternalGestureToFail(drawerGesture)
    }
  }, [drawerGesture, nativeGesture])

  return <GestureDetector gesture={nativeGesture}>{children}</GestureDetector>
}
```

Here's the flamegraph after:

Screenshot 2025-09-01 at 21 14 15

I don't love this fix though. Seems suspicious to me to be endlessly adding gestures to the other one without cleaning them up - is this allowed?

PR also includes some other cleanups to try and reduce the size of renders slightly

# Test plan

Confirm perf improvement
Confirm the drawer gesture synchronisation still works
Vibe check the fix

### DIFF
--- a/src/state/shell/drawer-open.tsx
+++ b/src/state/shell/drawer-open.tsx
@@ -1,15 +1,15 @@
-import React from 'react'
+import {createContext, useContext, useState} from 'react'
 
 type StateContext = boolean
 type SetContext = (v: boolean) => void
 
-const stateContext = React.createContext<StateContext>(false)
+const stateContext = createContext<StateContext>(false)
 stateContext.displayName = 'DrawerOpenStateContext'
-const setContext = React.createContext<SetContext>((_: boolean) => {})
+const setContext = createContext<SetContext>((_: boolean) => {})
 setContext.displayName = 'DrawerOpenSetContext'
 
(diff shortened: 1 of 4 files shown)


mozzius:
@github-actions ota

PR #8948, opened 2025-09-01
Add QR code to github action

mozzius:
code for the QR generator here: https://github.com/mozzius/bsky-qr/blob/main/src/app/route.tsx

Hopefully should look like this if it work!

Screenshot 2025-09-01 at 22 29 04


mozzius:
@github-actions ota

haileyok (approved):
:shipit:

PR #8949, opened 2025-09-01
[Perf] Better fix for drawer jitter (maybe)

mozzius:
Better fix for #8947 

I found a better fix for the jitter issue! While #8947 improves things by reducing the amount of work the device is doing and thus hiding the problem better, the real issue is that the final translation value from the gesture isn't being used, and thus it jumps back slightly to an old position when the animation ends. A one-line patch seems to fix it(?)

Both PRs together would probably be the best fix - there is still a big rerender problem

# Before

https://github.com/user-attachments/assets/650b9c52-9e8b-44bf-90f6-59584e31f829

# After

https://github.com/user-attachments/assets/03a08090-9658-498e-8bb2-f0f902eedd68

# Test plan

Confirm jitter is gone


mozzius:
@github-actions ota

mozzius:
@github-actions ota

PR #8952, opened 2025-09-02
[Perf] Drawer jitter fix (final) (final)

mozzius:
Supercedes #8947 and #8949 

Ok this time I think I have it 🙃

Turns out the _true_ cause was `toggleDrawer()` (which starts the animation) fires twice - first when the gesture ends, and then a second time in an effect when `open` changes.

Luckily, this is already fixed upstream! https://github.com/react-navigation/react-navigation/blob/main/packages/react-native-drawer-layout/src/views/Drawer.native.tsx#L241C7-L243C8

It's not yet released though, so here's a patch of just the relevant changes


mozzius:
@github-actions ota

estrattonbailey (approved):

PR #8960, opened 2025-09-02
Dev build QOL improvements

elijaharita:
Allows dev build to be installed alongside the production app on Android and iOS by giving them different app bundle IDs. Rebase of npmanos' branch.


mozzius:
ahhh lovely, good stuff

haileyok:
yaaaa thats the stuff
image

app.config.js
```diff
@@ -148,8 +152,10 @@ module.exports = function (_config) {
           backgroundImage: './assets/icon-android-background.png',
           backgroundColor: '#1185FE',
         },
-        googleServicesFile: './google-services.json',
-        package: 'xyz.blueskyweb.app',
+        googleServicesFile: IS_DEV
+          ? './google-services.json.example'
+          : './google-services.json',
```
mozzius, line 162: I can see this being a problem - Testflight builds would then not get push notifications

elijaharita: discussed previously, this is not an issue because IS_DEV is false on testflight builds

PR #8963, opened 2025-09-03
Use Automatic representation for all image pickers

estrattonbailey:
This fixes a regression caused by our [picker library upgrade](https://github.com/expo/expo/blob/main/packages/expo-image-picker/CHANGELOG.md#-breaking-changes-1). Although `Current` seems to work for most use cases, our handling of images is inconsistent. `Automatic` (the old default) is the safer route. It also fixes the regression we noticed where `.heif` images were no longer supported.

The only difference between this and what's in production is that `Automatic` applies to picked videos as well. I don't anticipate that causing issues, since `Automatic` "[chooses the appropriate asset representation](https://developer.apple.com/documentation/photokit/phpickerconfigurationassetrepresentationmode/automatic)", and the later compression step will convert further if needed.

### DIFF
--- a/src/screens/Onboarding/StepProfile/index.tsx
+++ b/src/screens/Onboarding/StepProfile/index.tsx
@@ -4,7 +4,7 @@ import {Image as ExpoImage} from 'expo-image'
 import {
   type ImagePickerOptions,
   launchImageLibraryAsync,
-  MediaTypeOptions,
+  UIImagePickerPreferredAssetRepresentationMode,
 } from 'expo-image-picker'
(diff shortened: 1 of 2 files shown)


src/screens/Onboarding/StepProfile/index.tsx
```diff
@@ -97,10 +97,12 @@ export function StepProfile() {
       const response = await sheetWrapper(
         launchImageLibraryAsync({
           exif: false,
-          mediaTypes: MediaTypeOptions.Images,
```
estrattonbailey, line 100: This was deprecated

PR #8965, opened 2025-09-03
Tweak spacing in composer bar

mozzius:
Composer lang select has a bunch of horizontal margin for some reason. It already has padding, so we can cut it down a bit. Includes a bit of right margin for visual alignment

# Before

Screenshot 2025-09-03 at 16 42 50

# After

Screenshot 2025-09-03 at 16 41 40


mozzius:
Superceded by #8981

PR #8966, opened 2025-09-03
add cursor constraint to safelink

haileyok:


bskylink/src/db/migrations/003-safelink-cursor-constraint.ts
```diff
@@ -0,0 +1,31 @@
...
+): Promise<void> {
+  // Remove existing items from safelink_rule that were duplicated due to broken cursor
+  await db.deleteFrom(['safelink_rule']).execute()
+
+  // Delete the old cursor
+  await db.deleteFrom(['safelink_cursor']).execute()
+
+  await db.schema
+    .alterTable('safelink_cursor')
+    .addUniqueConstraint('uc_id', ['id'])
```
ghost: Should you make it a PK instead?

PR #8969, opened 2025-09-04
[APP-1403] update: Empty state component

APiligrim:


src/view/com/util/EmptyState.tsx
```diff
@@ -1,76 +1,115 @@
...
   const pal = usePalette('default')
   const {isTabletOrDesktop} = useWebMediaQueries()
   const iconSize = isTabletOrDesktop ? 64 : 48
+
+  const renderIcon = () => {
+    if (typeof icon === 'object' && React.isValidElement(icon)) {
+      return icon
+    }
+
+    if (icon === 'user-group') {
```
BinaryFiddler: what are these used for? can't the icon object passed into here handle everything already?

src/view/com/util/EmptyState.tsx
```diff
@@ -1,76 +1,115 @@
...
-          <FontAwesomeIcon
-            icon={icon}
-            size={iconSize}
-            style={[{color: pal.colors.emptyStateIcon} as FontAwesomeIconStyle]}
-          />
-        )}
+        {renderIcon()}
       </View>
-      <Text type="xl" style={[{color: pal.colors.textLight}, styles.text]}>
+      <Text
```
BinaryFiddler, line 86: Should this be wrapped in `<Trans>`?

src/view/com/util/EmptyState.tsx
APiligrim: It looks like the `user-group` is no longer being used even though it was used in the past so I can remove it, and for the "growth" we can remove it as well once the Notifications empty state is being worked on either in this or a follow up PR.

src/view/com/util/EmptyState.tsx
APiligrim: thank you! Updated it!

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -135,9 +138,21 @@ export const ProfileFeedgens = React.forwardRef<
...
-            icon="hashtag"
-            message={_(msg`You have no feeds.`)}
-            testID="listsEmpty"
+            icon={
+              <HashtagWideIcon
+                size={64}
+                color={t.atoms.text_contrast_medium.color}
+              />
+            }
+            message="You haven't made any custom feeds yet."
```
mozzius: Needs translation

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -135,9 +138,21 @@ export const ProfileFeedgens = React.forwardRef<
...
+            textStyle={[t.atoms.text_contrast_medium, a.font_medium]}
+            button={{
+              label: 'Browse custom feeds',
+              text: 'Browse custom feeds',
```
mozzius: Same here

src/view/com/util/EmptyState.tsx
mozzius: That won't work unfortunately. When lingui statically extracts it, it will just see:

`<0></0>`

You need to mark the actual strings for translation. So for this, it would be the responsibility of the caller to do

```jsx
<EmptyState message={_(msg`This text is now translatable`)} />
```

src/view/com/util/EmptyState.tsx
```diff
@@ -1,76 +1,101 @@
...
+      return <Growth width={iconSize} fill={pal.colors.emptyStateIcon} />
+    }
+
+    return (
+      <FontAwesomeIcon
+        icon={icon as IconProp}
+        size={iconSize}
+        style={[{color: pal.colors.emptyStateIcon} as FontAwesomeIconStyle]}
+      />
+    )
```
mozzius: We need to get rid of fontawesome

mozzius (changes requested):
Left some comments about translations

src/view/com/util/EmptyState.tsx
APiligrim: I updated it and removed  the <Trans> and am passing messages like this to the component`<EmptyState message={_(msg`This text is now translatable`)} />`

src/components/icons/CircleAndSquare.tsx
```diff
@@ -0,0 +1,40 @@
...
+        />
+        <Circle
+          cx="44.1195"
+          cy="39.0366"
+          r="18.9633"
+          stroke={fill}
+          strokeWidth="2"
+          fill="none"
+        />
+      </Svg>
```
mozzius: Can we just mush it into a single path in figma? should be possible

src/components/icons/CircleAndSquare.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import Svg, {Circle, Rect} from 'react-native-svg'
+
+import {type Props, useCommonSVGProps} from '#/components/icons/common'
+
+export const Circle_And_Square_Stroke1_Corner0_Rounded_Filled =
+  React.forwardRef<Svg, Props>(function CircleAndSquareImpl(props, ref) {
```
mozzius: just fyi in general, forwardRef is no longer necessary since React 19

https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop

src/components/StarterPack/ProfileStarterPacks.tsx
```diff
@@ -47,12 +51,38 @@ interface ProfileFeedgensProps {
...
+//         icon={icon ?? 'growth'}
+//         message={
+//           message ??
+//           'Starter packs let you share your favorite feeds and people with your friends.'
+//         }
+//         button={button}
+//       />
+//     </View>
+//   )
+// }
```
mozzius: delete

src/lib/icons.tsx
```diff
@@ -304,6 +304,28 @@ export function HashtagIcon({
...
+        d="M23.8333 10L18.5 54M46.5 10L41.1667 54M10.5 20.6667H54.5M54.5 43.3333H10.5"
+        stroke="currentColor"
+        strokeWidth={strokeWidth}
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </Svg>
+  )
+}
+
```
mozzius: I think this should be with the other icon files?

src/view/com/util/EmptyState.tsx
```diff
@@ -1,76 +1,96 @@
...
+  )
+
+  const renderIcon = () => {
+    if (typeof icon === 'object' && React.isValidElement(icon)) {
+      return icon
+    }
+
+    if (icon === 'growth') {
+      return <Growth width={iconSize} fill={pal.colors.emptyStateIcon} />
+    }
```
mozzius: Let's get rid of this special case, it's silly. `icon` should just be a `React.ReactElement`

src/view/com/util/EmptyState.tsx
mozzius: Yeah perfect! 👌

src/view/com/feeds/ProfileFeedgens.tsx
```diff
@@ -147,9 +150,21 @@ export function ProfileFeedgens({
...
           <EmptyState
-            icon="hashtag"
-            message={_(msg`You have no feeds.`)}
-            testID="listsEmpty"
+            icon={
+              <HashtagWideIcon
+                size={64}
+                style={{color: t.atoms.text_contrast_low.color}}
+              />
+            }
```
mozzius: Ok I don't really like the API here. I think it should work more like our Buttons:

```jsx
<ButtonIcon icon={ExampleIcon} />
```

That way, `EmptyState` can take care of the color and the size, and it's a lot easier to use/think about.

src/components/icons/VideoClip.tsx
```diff
@@ -1,5 +1,13 @@
 import {createSinglePathSVG} from './TEMPLATE'
 
+export const VideoClip_Stroke1_Corner0_Rounded = createSinglePathSVG({
+  viewBox: '0 0 47 46',
+  strokeWidth: 2,
+  strokeLinecap: 'square',
+  strokeLinejoin: 'round',
```
mozzius: Is there a reason these are so custom? Is it possible for them to just work like the other normal icons?

mozzius (changes requested):
Okay, it's looking pretty nice! I think my one piece of feedback is about the API of the `EmptyState` component - the `icon` prop should look more like this:

```jsx
<EmptyState
   icon={ExampleIcon}
   // etc
/>
```

That way, we can just slot in icons much more easily and not have to think about styles or sizes or colors.

src/components/icons/CircleAndSquare.tsx
APiligrim: updated it!

src/components/icons/BulletList.tsx
```diff
@@ -1,5 +1,10 @@
...
+  viewBox: '0 0 47 38',
+  path: 'M6.83301 25.333C10.3308 25.333 13.167 28.1691 13.167 31.667C13.1668 35.1647 10.3307 38 6.83301 38C3.33545 37.9998 0.500158 35.1645 0.5 31.667C0.5 28.1692 3.33534 25.3332 6.83301 25.333ZM6.83301 27.333C4.43993 27.3332 2.5 29.2738 2.5 31.667C2.50016 34.06 4.44001 35.9998 6.83301 36C9.22616 36 11.1668 34.0601 11.167 31.667C11.167 29.2737 9.22625 27.333 6.83301 27.333ZM45.5 30.667C46.0523 30.667 46.5 31.1147 46.5 31.667C46.4998 32.2191 46.0522 32.667 45.5 32.667H22.833C22.281 32.6668 21.8332 32.219 21.833 31.667C21.833 31.1148 22.2809 30.6671 22.833 30.667H45.5ZM6.83301 0C10.3307 0 13.1668 2.83534 13.167 6.33301C13.167 9.83081 10.3308 12.667 6.83301 12.667C3.33534 12.6668 0.5 9.8307 0.5 6.33301C0.500174 2.83546 3.33546 0.000174194 6.83301 0ZM6.83301 2C4.44003 2.00017 2.50017 3.94003 2.5 6.33301C2.5 8.72615 4.43993 10.6668 6.83301 10.667C9.22624 10.667 11.167 8.72624 11.167 6.33301C11.1668 3.93993 9.22615 2 6.83301 2ZM45.5 5.33301C46.0522 5.33301 46.4998 5.78087 46.5 6.33301C46.5 6.88529 46.0523 7.33301 45.5 7.33301H22.833C22.2809 7.33285 21.833 6.8852 21.833 6.33301C21.8332 5.78097 22.281 5.33317 22.833 5.33301H45.5Z',
```
mozzius: All of these new icons - can you also add them to `/assets/icons`?

There's a script we can run to then optimise them - `yarn icons:optimize` which helps reduce bundle size

src/view/com/lists/MyLists.tsx
```diff
@@ -102,34 +85,13 @@ export function MyLists({
...
-                },
-              ]}>
-              {emptyText}
-            </Text>
-          </View>
+          <EmptyState
+            icon={ListIcon}
+            message={_(
+              msg`Lists allow you to see content from your favorite people.`,
+            )}
```
mozzius: Not quite right - this screen is used for all types of list, including moderation lists, which is the opposite of what the text says!

I would keep the old logic above with the switch statement, and just update the text for the `curate` case.

Image

src/view/com/lists/ProfileLists.tsx
```diff
@@ -50,7 +53,7 @@ interface ProfileListsProps {
   setScrollViewTag: (tag: number | null) => void
 }
 
-export function ProfileLists({
+export function ProfileFeedgens({
```
mozzius: Uhhh what's going on here. Did you copy/paste `ProfileFeedgens.tsx` in here by accident? I now see my Feeds in my Lists tab

Image

src/view/com/lists/ProfileLists.tsx
```diff
@@ -233,3 +250,6 @@ export function ProfileLists({
 function keyExtractor(item: any) {
   return item._reactKey || item.uri
 }
+
+// Export as ProfileLists for compatibility
+export const ProfileLists = ProfileFeedgens
```
mozzius: ???

src/view/com/lists/MyLists.tsx
APiligrim: Updated it to only the _curate_ case!

abenzer: @mozzius can we get a final review from you please?

src/components/Button.tsx
```diff
@@ -798,13 +798,13 @@ export function ButtonIcon({
...
        */
       const iconSize = {
-        '2xs': 8,
         xs: 12,
         sm: 16,
         md: 18,
         lg: 24,
         xl: 28,
         '2xl': 32,
+        '3xl': 40,
```
mozzius, line 808:
```suggestion
        '2xs': 8,
        xs: 12,
        sm: 16,
        md: 18,
        lg: 24,
        xl: 28,
        '2xl': 32,
        '3xl': 40,
```

mozzius: Missing 2xs

mozzius:
@github-actions ota

PR #8976, opened 2025-09-04
📓 Bookmarks

estrattonbailey:
Well folks, we finally did it. Bookmarks coming at you soon 🎉 

> [!NOTE]
> ATproto currently only supports public data. On the other hand, we've designed bookmarks as a private feature, in which only the user can see their own bookmarks. We might expand the feature to allow for making some bookmarks public (e.g., create collections and define visibility), but that's outside the initial scope.
>
> [We have the intention to extend the protocol to add support for private data](https://github.com/bluesky-social/atproto/discussions/121#discussioncomment-6065949), but it is a complex topic, and we haven't yet built support for it.
> 
> So for this feature, we rely on private storage. We have modeled our private storage on lexicon definitions to facilitate a future migration to on-protocol, once we have that capability.

CleanShot 2025-09-04 at 17 16 21@2x
CleanShot 2025-09-04 at 17 16 49@2x

### DIFF
--- a/src/components/PostControls/index.tsx
+++ b/src/components/PostControls/index.tsx
@@ -188,7 +188,8 @@ let PostControls = ({
   }
 
   const secondaryControlSpacingStyles = flatten([
-    a.gap_xs,
+    {gap: 0}, // default, we want `gap` to be defined on the resulting object
+    variant !== 'compact' && a.gap_xs,
     (big || gtPhone) && a.gap_sm,
   ])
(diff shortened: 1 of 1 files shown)


surfdude29:
For anyone interested in the technical details of the bookmarks implementation on the backend, please also see the PR in the `atproto` repo:

- https://github.com/bluesky-social/atproto/pull/4163

PR #8981, opened 2025-09-05
Restore quick language select

### LINKED ISSUE #8971: New Language dialog is *worse*
### Describe the Feature

I see how the idea that I can just flag my posts with the language I am using seems like a reasonable idea on the surface, but to me it seems like it sprung from the minds of people who don't really use different languages all that much.

I am constantly moving between German and English all day. My brain has gotten to a point where I can't even remember which language I was handling a certain interaction in the very recent past. 

It seems that an automatic language detection would be more precise than me forgetting to set it all the time. 

Mind you, up to now that was a simple drop-down and not the multi-choice dialog monstrosity we have now.

Not a fan.

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

mozzius:
Fixes #8971 

With the revamp of the language select, we lost the language quick switcher. On consideration, the full multi-language picker is many more taps than the old one

This PR restores this functionality using modern components

# Web

This showcases what happens if you have no languages in your language history

https://github.com/user-attachments/assets/b96bd9a2-f8e2-4da1-94fd-9f071093b8cc

# iOS

https://github.com/user-attachments/assets/0a5c7360-4ee4-419e-b705-3c78936949db

# Android

https://github.com/user-attachments/assets/8f26fb2c-3535-43ef-8d2f-b617fec491e9

# Test plan

Test the case of having no languages in your history, and then lots of languages

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -110,7 +110,6 @@ import {LabelsBtn} from '#/view/com/composer/labels/LabelsBtn'
 import {Gallery} from '#/view/com/composer/photos/Gallery'
 import {OpenCameraBtn} from '#/view/com/composer/photos/OpenCameraBtn'
 import {SelectGifBtn} from '#/view/com/composer/photos/SelectGifBtn'
-import {SelectPostLanguagesBtn} from '#/view/com/composer/select-language/SelectPostLanguagesDialog'
 import {SuggestedLanguage} from '#/view/com/composer/select-language/SuggestedLanguage'
 // TODO: Prevent naming components that coincide with RN primitives
 // due to linting false positives
@@ -131,6 +130,7 @@ import * as Prompt from '#/components/Prompt'
 import * as Toast from '#/components/Toast'
 import {Text as NewText} from '#/components/Typography'
 import {BottomSheetPortalProvider} from '../../../../modules/bottom-sheet'
+import {PostLanguageSelect} from './select-language/PostLanguageSelect'
 import {
   type AssetType,
   SelectMediaButton,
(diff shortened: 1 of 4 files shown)


src/view/com/composer/select-language/PostLanguageSelect.tsx
```diff
@@ -0,0 +1,129 @@
...
+    dedupedHistory.length === 1 &&
+    dedupedHistory[0] === langPrefs.postLanguage
+  ) {
+    return (
+      <>
+        <LanguageBtn onPress={languageDialogControl.open} />
+        <PostLanguageSelectDialog control={languageDialogControl} />
+      </>
+    )
+  }
```
mozzius, line 40: if there's no history, jump straight to the full dialog

src/view/com/composer/select-language/PostLanguageSelect.tsx
```diff
@@ -0,0 +1,131 @@
...
+            })}
+          </Menu.Group>
+          <Menu.Divider />
+          <Menu.Item
+            label={_(msg`More languages...`)}
+            onPress={languageDialogControl.open}>
+            <Menu.ItemText>
+              <Trans>More languages...</Trans>
+            </Menu.ItemText>
+            <Menu.ItemIcon icon={ChevronRightIcon} />
```
mozzius, line 75: I don't love this text+icon combo. any suggestions?

Screenshot 2025-09-05 at 13 05 46

src/view/com/composer/Composer.tsx
```diff
@@ -1440,17 +1441,13 @@ function ComposerFooter({
       <View style={[a.flex_row, a.align_center, a.justify_between]}>
         {showAddButton && (
           <Button
-            label={_(msg`Add new post`)}
+            label={_(msg`Add another post to thread`)}
```
surfdude29, line 1444: That's much clearer 👌

surfdude29:
Looks and works great on the preview branch! :)

estrattonbailey (approved):
So much nicer

PR #8982, opened 2025-09-05
Add translation context for "Saved"

mozzius:


estrattonbailey (approved):

estrattonbailey (changes requested):
Wait let's use `comment` so it doesn't result in new translations

mozzius:
ah shit

Signez:
> ah shit

No, I really think it **should** be a different `context` (and not simply a `comment`) if you don't want the problem I described earlier on Discord:

> Please consider adding context tags for very common, short words with a lot of meanings like _"Saved"_, especially when they are used in contexts like the `Drawer` or `LeftNav`.
> 
> I know it can feel tedious to add context tags, but for such overloaded terms, we may have to translate them with very different words in other languages — and you don't want a word used for _"kept somewhere to be found again quickly"_ to show up in the future during a profile save for example, without anybody noticing it 😅 

 If you only use `comment`, the string won't be different and it _will_ cause problems in the long run.

surfdude29:
Also, it's very easy for me to carry over translations for `Saved` that have already been done, I'm happy to do that to ensure that no translation gets left out and there's no delay to cutting the release.

PR #8987, opened 2025-09-05
Use comment instead of context

mozzius:


Signez:
Sorry, but please reconsider this PR and keep it as a `context` tag. Yeah, it causes string churn, but it's because it will prevent other usage of `Saved` in the future to be translated automatically with the same terms when it could clearly not be the same. (See this comment [for more context](https://github.com/bluesky-social/social-app/pull/8982#issuecomment-3258622414)).

mozzius:
Oh damn it I keep merging stuff without reading the comments. Will revert

estrattonbailey:
@Signez ah ok now we got it, thanks for the clarification!

PR #8991, opened 2025-09-05
Make logs more clear

estrattonbailey:


src/state/geolocation/config.ts
```diff
@@ -85,7 +85,7 @@ export function beginResolveGeolocationConfig() {
     } catch (e: any) {
       success = false
 
-      logger.debug(`geolocation config: failed initial request`, {
+      logger.debug(`config: failed initial request`, {
```
estrattonbailey, line 88: The logger's `context` prop already prefixes these logs, so this was redundant

src/state/geolocation/index.tsx
```diff
@@ -91,11 +91,13 @@ export function GeolocationStatusProvider({
 
   const configContext = React.useMemo(() => ({config}), [config])
   const statusContext = React.useMemo(() => {
-    if (deviceGeolocation) {
-      logger.debug('geolocation: has device geolocation available')
+    if (deviceGeolocation?.countryCode) {
```
estrattonbailey, line 94: Can be an empty object, this is the more appropriate check

src/state/geolocation/index.tsx
```diff
@@ -91,11 +91,13 @@ export function GeolocationStatusProvider({
...
+      logger.debug('has device geolocation available')
     }
     const geolocation = mergeGeolocation(deviceGeolocation, config)
     const status = computeGeolocationStatus(geolocation, config)
+    // ensure this remains debug and never leaves device
+    logger.debug('result', {deviceGeolocation, geolocation, status, config})
```
estrattonbailey, line 100: Debug logs are explicitly disabled from going off device

mozzius (approved):

PR #8994, opened 2025-09-05
Safi/design forgot password

melodylumen:
- rebase on current develop
- optimize imports (more work to do here to wrap long lines)


melodylumen:
opened in error

PR #8996, opened 2025-09-06
Fix check for whether string is only emoji using more accurate unicode groups.

addiesh:
This breaks up the original check into two separate regexes, but the original failed on like 90% of the emoji I tried. The culprit, nine times out of ten, was either a variation selector 16 or a zwj, which (as it turns out) are used a lot!!! This should account for that. Anyway I'm not getting paid nor do I enjoy writing typescript, I'm just chronically online, so please accept my PR.
(also, thanks to xan.lol for bringing this onto my feed, I love a good nerd sniping)

### DIFF
--- a/src/alf/typography.tsx
+++ b/src/alf/typography.tsx
@@ -122,7 +122,11 @@ export function renderChildrenWithEmoji(
   })
 }
 
-const SINGLE_EMOJI_RE = /^[\p{Emoji_Presentation}\p{Extended_Pictographic}]+$/u
+const CONTAINS_EMOJI_RE = /[\p{Extended_Pictographic}]/u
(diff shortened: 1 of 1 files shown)


addiesh:
this is the code I hacked together to test it:
```js

[
    // only a V16
    ["\uFE0F", false],
    ["bwah", false],
    ["#", false],
    ["🙀", true],
    ["🏳️‍⚧️", true],
    ["🐈‍⬛", true],
    ["👍", true],
    ["👍", true],
    ["🫃🏿", true],
    ["🏴‍☠️", true],
].forEach(e => {
    console.info(`${e[0]}`);
    let ioe = isOnlyEmoji(e[0]);
    console.info(`${e[0]}: ${ioe}`);
    console.assert(ioe == e[1]);
});
```

PR #9006, opened 2025-09-08
Clean up language before suggesting it in composer

mozzius:
Some clients/bots use non-2 letter language codes. We should parse it before doing the suggestion thing

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

### DIFF
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -1,5 +1,6 @@
 import {useEffect, useState} from 'react'
 import {View} from 'react-native'
+import {parseLanguage} from '@atproto/api'
 import {msg, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import lande from 'lande'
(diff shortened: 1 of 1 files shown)


src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -125,3 +127,11 @@ function guessLanguage(text: string): string | undefined {
   }
   return code3ToCode2Strict(lang)
 }
+
+function cleanUpLanguage(text: string | undefined): string | undefined {
+  if (!text) {
+    return undefined
+  }
+
+  return parseLanguage(text)?.language
```
estrattonbailey, line 136: Huh, TIL

mozzius: same lol

mozzius: atproto/api is full of wonders

PR #8931, opened 2025-08-28
Expo 54

mozzius:
- [x] Update deps
- [x] Update patches
- [x] Patch https://github.com/henninghall/react-native-date-picker/issues/937
- [x] Patch https://github.com/bluesky-social/react-native-paste-input/pull/5
- [x] ~~Fix `paste-input`~~ _I accidentally enabled new arch and got confused by the error message lol, all good here_
- [x] Fix expo modules
- [x] `expo-file-system` -> `expo-file-system/legacy`
- [x] Remove Zeego and its dependencies - no longer needed

Fixes https://github.com/bluesky-social/social-app/issues/8386 🎉

### To do:

- [x] 💥 Crash at startup
  Related issues:
  - https://github.com/expo/expo/issues/39597
  - https://github.com/mrousavy/react-native-mmkv/issues/688

PR #9027, opened 2025-09-11
Allow swipe dismissal of the composer on iOS

mozzius:
# Stacked on #8931 

The latest version of React Native comes with a new prop on the built-in Modal - [allowSwipeDismissal](https://reactnative.dev/docs/modal#allowswipedismissal-ios)

This allows much smooth and natural swipe dismiss gesture of the composer. In this PR, I use it for when the composer is empty, i.e. for when we don't show the "discard draft" prompt. if we are going to show the prompt, we disable the prop so the swipe gesture has "resistance" - I think it feels really nice

~~This requires some awkward useEffect usage, but it's pretty limited in scope and fairly obvious what's going on at least.~~

https://github.com/user-attachments/assets/bf885c20-5d01-4854-9d87-a0245f626797

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -173,8 +173,10 @@ export const ComposePost = ({
   imageUris: initImageUris,
   videoUri: initVideoUri,
   cancelRef,
+  setIsDirty,
 }: Props & {
   cancelRef?: React.RefObject<CancelRef | null>
+  setIsDirty?: React.Dispatch<React.SetStateAction<boolean>>
 }) => {
   const {currentAccount} = useSession()
   const agent = useAgent()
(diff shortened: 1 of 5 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -304,24 +306,37 @@ export const ComposePost = ({
     [insets, isKeyboardVisible],
   )
 
-  const onPressCancel = useCallback(() => {
-    if (textInput.current?.maybeClosePopup()) {
+  const isDirty = thread.posts.some(
+    post =>
+      post.shortenedGraphemeLength > 0 || post.embed.media || post.embed.link,
+  )
```
mozzius: we could consider changing this to comparing the current state to the initial state - i.e. if it was opened with prefilled text

src/view/com/composer/Composer.tsx
```diff
@@ -304,24 +306,37 @@ export const ComposePost = ({
...
+  // WARNING - if the Modal on iOS thinks it's not dirty,
+  // `allowSwipeDismissal` will be true, and if we don't then close the composer
+  // when `onPressCancel` is called it will be bad (might even softlock)
+  // so we need to keep the parent state and the behaviour of onPressCancel in
+  // tight sync. do NOT force the modal to stay open without marking it as dirty! -sfn
+  useEffect(() => {
+    if (isIOS) {
+      setIsDirty?.(isDirty)
+    }
+  }, [isDirty, setIsDirty])
```
estrattonbailey: Can we not compute `isDirty` up a level and pass it in here?

mozzius: We'd have to move the composer state up to the parent level, which is platform split. Doable but potentially quite verbose - I can give it a go if you think it'd be better?

mozzius: This current approach is the minimal LoC but I agree it's not the best

mozzius:
@estrattonbailey I attempted to hoist the state to the parent. Seems to work, but I'm not a huge fan of how we then have to keep the state around even when the composer is closed. Will think more on it because there was a nice simplicity to how it worked before

PR #9029, opened 2025-09-11
Log cleanup

mozzius:
Noticed sentry is dying again


src/state/queries/notifications/unread.tsx
```diff
@@ -190,7 +191,13 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
           }
           broadcast.postMessage({event: unreadCountStr})
         } catch (e) {
-          logger.warn('Failed to check unread notifications', {error: e})
+          if (!isNetworkError(e)) {
+            logger.warn('Failed to check unread notifications', {error: e})
+          } else {
+            logger.log(
+              'Failed to check unread notifications due to network error',
+            )
```
estrattonbailey: `log` will still get sent as `captureMessage`. If you don't want this to reach Sentry, use either `debug` or remove entirely. I think we can do the latter here.

src/lib/hooks/useNotificationHandler.ts
```diff
@@ -386,7 +386,7 @@ export function getNotificationPayload(
     return payload
   } else {
     if (payload) {
-      notyLogger.warn('getNotificationPayload: received unknown payload', {
+      notyLogger.log('getNotificationPayload: received unknown payload', {
```
estrattonbailey: Same as other comment. If these logs aren't useful we can remove.

mozzius:
Will rm!

src/state/queries/notifications/unread.tsx
mozzius: copy that

PR #9030, opened 2025-09-11
Migrate to `@bsky.app/alf` package

estrattonbailey:
In this PR `@bsky.app/alf` acts as the "core" library. That library is intended to be agnostic of JS runtime, and contain only the building blocks needed for more use-case-specific integrations.

In that library, a few changes where made, notably:
- aligned font weights with Inter e.g. where we used `font_bold` previously to refer to the "semi bold" weight, we now use `font_semi_bold`
- font sizes are now defined as the scaled decimals we were computing in this codebase, therefore the scaling here has been adjusted down one notch to ensure no changes to this codebase
- I didn't migrate over all atoms, since some are codebase specific. Those are just extended here to no external effect.

Other than that, this PR basically contains some refreshed colors. There should be no noticeable changes to font sizes/weights or other tokens.

### DIFF
--- a/src/components/forms/TextField.tsx
+++ b/src/components/forms/TextField.tsx
@@ -15,7 +15,6 @@ import {
   android,
   applyFonts,
   atoms as a,
-  ios,
   platform,
   type TextStyleProp,
   tokens,
@@ -202,17 +201,23 @@ export function createInput(Component: typeof TextInput) {
       a.px_xs,
       {
         // paddingVertical doesn't work w/multiline - esb
-        lineHeight: a.text_md.fontSize * 1.1875,
+        lineHeight: a.text_md.fontSize * 1.2,
         textAlignVertical: rest.multiline ? 'top' : undefined,
         minHeight: rest.multiline ? 80 : undefined,
         minWidth: 0,
+        paddingTop: 13,
+        paddingBottom: 13,
       },
-      ios({paddingTop: 12, paddingBottom: 13}),
-      // Needs to be sm on Paper, md on Fabric for some godforsaken reason -sfn
-      android(a.py_sm),
-      // fix for autofill styles covering border
+      android({
+        paddingTop: 8,
+        paddingBottom: 9,
+      }),
+      /*
+       * Margins are needed here to avoid autofill background overlapping the
+       * top and bottom borders - esb
+       */
       web({
-        paddingTop: 10,
+        paddingTop: 11,
         paddingBottom: 11,
         marginTop: 2,
         marginBottom: 2,
@@ -262,7 +267,7 @@ export function createInput(Component: typeof TextInput) {
             a.absolute,
             a.inset_0,
             a.rounded_sm,
-            t.atoms.bg_contrast_25,
+            t.atoms.bg_contrast_50,
             {borderColor: 'transparent', borderWidth: 2},
             ctx.hovered ? chromeHover : {},
             ctx.focused ? chromeFocus : {},


mozzius:
@github-actions ota

src/alf/util/systemUI.ts
```diff
@@ -1,7 +1,7 @@
 import * as SystemUI from 'expo-system-ui'
+import {type Theme} from '@bsky.app/alf'
```
mozzius, line 2: `@bsky.app/alf` makes me think... what if atproto package manager...

src/alf/index.tsx
```diff
@@ -96,13 +89,7 @@ export function ThemeProvider({
...
   )
   const themes = React.useMemo(() => {
-    return createThemes({
-      hues: {
-        primary: BLUE_HUE,
-        negative: RED_HUE,
-        positive: GREEN_HUE,
-      },
-    })
+    return createThemes()
```
mozzius: any reason we do this in a useMemo and not just define it outside of the component tree? That way we don't need to do it twice (once when we create the provider, once here)

We can still pass it through context so you can override it, but we could just make the default one once

src/alf/util/systemUI.ts
estrattonbailey: Imagine that

src/alf/index.tsx
estrattonbailey: Nah no reason, was just refactoring in place. Good call 👍

mozzius: Nice nice, yeah looks much cleaner now 👍

src/alf/fonts.ts
```diff
@@ -7,11 +7,11 @@ const WEB_FONT_FAMILIES = `system-ui, -apple-system, BlinkMacSystemFont, "Segoe
...
-  '-2': 1 - factor * 3,
-  '-1': 1 - factor * 2,
-  '0': 1 - factor * 1, // default
-  '1': 1,
-  '2': 1 + factor * 1,
+  '-2': 1 - factor * 1, // unused
+  '-1': 1 - factor * 1,
+  '0': 1, // default
+  '1': 1 + factor * 1,
+  '2': 1 + factor * 1, // unused
```
mozzius, line 14: any reason we even have the -2 and 2?

src/alf/tokens.ts
```diff
@@ -1,61 +1,10 @@
...
-/**
- * These correspond to Inter font files we actually load.
- */
-export const fontWeight = {
-  normal: '400',
-  medium: '500',
-  bold: '600',
-  heavy: '800',
+  temp_purple: tokens.labelerColor.purple,
+  temp_purple_dark: tokens.labelerColor.purple_dark,
```
mozzius, line 7: ah we should add Adminition orange to here

mozzius (approved):
This looks great, I also read through the `@bsky.app/alf` code and it looks very solid and clean. Great work!

src/alf/fonts.ts
estrattonbailey: Nah just leftovers, just didn't clean up here

src/alf/tokens.ts
estrattonbailey: Oh yeah that's a good idea too

davidfraser:
@estrattonbailey is there a public git repo for @bsky.app/alf? I couldn't see one but would be useful to have...

mozzius:
It's hosted here on Tangled: https://tangled.org/esb.lol/alf/

PR #9032, opened 2025-09-11
PDS Resolution From Handle

ETJeanMachine:
# Feature: PDS Resolution From Handle

This adds PDS resolution to the log-in flow, while preserving existing options to manually enter your PDS from the previous dialogue. This handles both PLC and web DID's in the resolution flow.

plc self hosted account

This feature supports did:web, did:plc, as well as emails (if the user manually inputs a domain). By default, it shows bsky.social as the login source, and will change depending on what the user enters.

The feature supports changing between handles in the login flow as well, including reverting back to a bsky.social domain. In the future, this feature should likely be replaced by a oauth implementation. 

Additionally, there isnt any buffer animation when you revert to a bsky.social domain (like how the other have), but since no API calls are necessary here, the slowness of the page to make the visual change isn't relevant to messing up the login flow.


ETJeanMachine:
I am wondering if changing the UI of the login screen here at all makes sense - such as presenting a checkbox instead of the existing "edit" box for automatically finding the proper PDS, and then when unchecking it you get the ability to manually input the PDS. Open to suggestions here, but I don't want to make substantial UI changes (hence the current approach) w/o asking, and I also know Oauth 2.0 is likely in the works so I don't want to massively shake things up prior to that.

ETJeanMachine:
Converting this to a draft because I want to confirm either that this isnt a feature in active consideration due to oauth/oauth scopes being in development, and also I've had some additional considerations about how to implement it as is, and I'm not 100% satisified with my current approach.

As an aside, I'm editing this page to remove the examples because its crashing my web browser every time I open up the PR, lol.

PR #9036, opened 2025-09-12
Wrap interaction counts with `Trans` to add translation flexibility

surfdude29:
After a question from @nilaallj on Discord about adding flexibility for translators in the interaction counts displayed at the bottom of a post, I asked Claude and he (eventually) came up with a very simple solution that seems plausible to me: simply wrap the whole string in a `Trans` macro, with the `Plural` component nested inside.

This should give translators the flexibility to change the order in the string if it would work better in their language.

> [!NOTE]
> **Claude wrote the code for this PR, and although it passes CI I have not tested it.**
>
> **Also, this PR would cause churn in highly-visible strings. So if it were to be approved, ideally merging would happen in the period after a release has been cut, to give translators the maximum amount of time to update translations in time for the subsequent release.**

[Link to review without whitespace](https://github.com/bluesky-social/social-app/pull/9036/files?diff=unified&w=1)


surfdude29:
@mozzius Conflicts fixed, should be good now

mozzius:
@claude /review

surfdude29:
@mozzius Ditto, sorry the lockfile check failed, now resynced to main. I don't think @claude's suggestion here is necessary, I think it will be clear to translators that they can reorder the elements.

mozzius (approved):

PR #9050, opened 2025-09-18
feat(composer): add an account switcher

### LINKED ISSUE #5531: Allow switching accounts from the post composer

Sometimes I start writing a post and then realize I'm logged into the wrong account. This can be particularly annoying if it's a reply, since then I have to exit the composer, switch accounts, find the post I wanted to reply to, and then re-type my post. It'd be nice to be able to quickly switch accounts from the post composer without losing what I've typed so far.

Perhaps tapping my avatar in the composer could open the account switcher so I could quickly select which account to post from.

1. Copy the post I wrote to the clipboard
2. Switch accounts
3. If I was replying to someone, go find the post I wanted to reply to
4. Paste the text I had written (or write it from scratch)
5. Publish the post

**Additional context**

N/A

### LINKED ISSUE #6708: Quick account picker in posts and responses
### Describe the Feature

When you have multiple accounts logged in, it would be very helpful if you could click your user avatar and choose the account that you want the post or response to come from.

This is particularly useful for responses, because this is how the workflow goes right now:

1. you see a great post from paul frazee on your timeline
2. you attempt to reply, but you notice that you're logged into your NSFW alt-account
3. you close out the reply
4. you go to the account switcher, switch to your work account
5. you then search for the account that made the post, scroll through their timeline to try to find it
6. once you find it, you type up your response and click submit
7. you then go back into the account switcher and choose your NSFW alt-account again

This is what I would propose -- it's exactly how the Twitter iOS / Android clients behave:

1. you see yet another great post from paul frazee on your timeline
2. you attempt to reply, but notice that you're logged into your NSFW alt-account
3. you click your avatar, the account picker pops up, and you choose your work account
4. you type your response and click submit
5. (you're done, and remain logged into your NSFW alt-account)

This would save me several minutes each time I do this, and is a very important feature for people who split their lives into multiple online personas. It's pretty easy to test this with the Twitter client -- simply log in to multiple accounts, go to make a response, and click your avatar.

### Attachments

Step 1: click your user icon
IMG_20091004C608-1

Step 2: it pops up the user switcher, you choose
IMG_6059F5ABBF29-1

Step 3: it now shows where the reply is coming from, keeping you in your original account once the reply happens
IMG_B52C3EC7D4D7-1

### Describe Alternatives

_No response_

### Additional Context

_No response_

SapphoSys:
This pull request adds an account switcher to the post composer. Closes #6708 and closes #5531.

## Technical details
When the user clicks on their user avatar, the account switcher is triggered. If there is only one account, the switcher is disabled.

It temporarily assumes an identity just for the composer and goes away if the modal is dismissed. The global site identity remains unaffected.

It uses the same strings as [src/components/dialogs/SwitchAccount.tsx](https://github.com/bluesky-social/social-app/blob/main/src/components/dialogs/SwitchAccount.tsx). The expired token logic is more or less the same as [src/lib/hooks/useAccountSwitcher.ts](https://github.com/bluesky-social/social-app/blob/main/src/lib/hooks/useAccountSwitcher.ts), only that I temporarily spawn an agent beforehand to check if the identity is valid.

Feature tested on Web and Android. I don't have an iOS device at hand, so I was unable to test the changes there.

### Web

msedge_hFg5pBmZBG

### Android

<details>
<summary>(long screenshot)</summary>
Screenshot_20250918-224358_Bluesky
</details>

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -57,7 +57,7 @@ import {FontAwesomeIcon} from '@fortawesome/react-native-fontawesome'
 import {msg, plural, Trans} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
-import {useQueryClient} from '@tanstack/react-query'
+import {useQuery, useQueryClient} from '@tanstack/react-query'
 
 import * as apilib from '#/lib/api/index'
 import {EmbeddingDisabledError} from '#/lib/api/resolve'
@@ -187,26 +187,29 @@ export const ComposePost = ({
   const defaultAgent = useAgent()
   const [selectedAccount, setSelectedAccount] = useState(currentAccount!)
 
-  const agent = useMemo(() => {
-    if (selectedAccount.did === currentAccount!.did) {
-      return defaultAgent
-    }
-    const session = new CredentialSession(new URL(selectedAccount.service))
-    if (
-      selectedAccount.refreshJwt &&
-      selectedAccount.accessJwt &&
-      selectedAccount.active
-    ) {
-      session.resumeSession({
-        ...selectedAccount,
-        accessJwt: selectedAccount.accessJwt,
-        refreshJwt: selectedAccount.refreshJwt,
-        active: selectedAccount.active,
-      })
-    }
-    const newAgent = new AtpAgent(session)
-    return newAgent
-  }, [selectedAccount, currentAccount, defaultAgent])
+  const {data: agent = defaultAgent} = useQuery({
+    queryKey: ['composer-agent', currentAccount, selectedAccount],
+    queryFn: async () => {
+      if (selectedAccount.did === currentAccount!.did) {
+        return defaultAgent
+      }
+      const session = new CredentialSession(new URL(selectedAccount.service))
+      if (
+        selectedAccount.refreshJwt &&
+        selectedAccount.accessJwt &&
+        selectedAccount.active
+      ) {
+        await session.resumeSession({
+          ...selectedAccount,
+          accessJwt: selectedAccount.accessJwt,
+          refreshJwt: selectedAccount.refreshJwt,
+          active: selectedAccount.active,
+        })
+      }
+      return new AtpAgent(session)
+    },
+    staleTime: 5 * 60 * 1000,
+  })
 
   const queryClient = useQueryClient()
   const currentDid = selectedAccount.did
@@ -264,6 +267,11 @@ export const ComposePost = ({
       try {
         // try a simple request to check if the session is valid
         await tempAgent.getProfile({actor: account.did})
+        // lets cache the agent for this account to avoid double resume
+        queryClient.setQueryData(
+          ['composer-agent', currentAccount, selectedAccount],
+          tempAgent,
+        )
         // if it succeeds, update the selected account
         setSelectedAccount(account)
       } catch (e: any) {
@@ -283,11 +291,12 @@ export const ComposePost = ({
       }
     },
     [
-      selectedAccount.did,
+      selectedAccount,
       closeComposer,
       requestSwitchToAccount,
       _,
-      setSelectedAccount,
+      currentAccount,
+      queryClient,
     ],
   )
 
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -188,7 +188,15 @@ export const ComposePost = ({
   const [selectedAccount, setSelectedAccount] = useState(currentAccount!)
 
   const {data: agent = defaultAgent} = useQuery({
-    queryKey: ['composer-agent', currentAccount, selectedAccount],
+    queryKey: [
+      'composer-agent',
+      currentAccount?.did,
+      currentAccount?.service,
+      selectedAccount?.did,
+      selectedAccount?.service,
+      selectedAccount?.accessJwt,
+      selectedAccount?.refreshJwt,
+    ],
     queryFn: async () => {
       if (selectedAccount.did === currentAccount!.did) {
         return defaultAgent
@@ -269,7 +277,15 @@ export const ComposePost = ({
         await tempAgent.getProfile({actor: account.did})
         // lets cache the agent for this account to avoid double resume
         queryClient.setQueryData(
-          ['composer-agent', currentAccount, selectedAccount],
+          [
+            'composer-agent',
+            currentAccount?.did,
+            currentAccount?.service,
+            account.did,
+            account.service,
+            account.accessJwt,
+            account.refreshJwt,
+          ],
           tempAgent,
         )
         // if it succeeds, update the selected account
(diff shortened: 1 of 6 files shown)


mozzius:
holy shit

src/view/com/composer/Composer.tsx
```diff
@@ -176,11 +183,35 @@ export const ComposePost = ({
...
+      selectedAccount.refreshJwt &&
+      selectedAccount.accessJwt &&
+      selectedAccount.active
+    ) {
+      session.resumeSession({
+        ...selectedAccount,
+        accessJwt: selectedAccount.accessJwt,
+        refreshJwt: selectedAccount.refreshJwt,
+        active: selectedAccount.active,
+      })
```
mozzius: `session.resumeSession()` is an async API call - not sure how I feel about doing this in a `useMemo`. Maybe `useQuery` might be more appropriate? We should probably await this

src/view/com/composer/Composer.tsx
```diff
@@ -209,6 +243,54 @@ export const ComposePost = ({
...
+          active: account.active,
+        })
+      }
+      const tempAgent = new AtpAgent(session)
+
+      try {
+        // try a simple request to check if the session is valid
+        await tempAgent.getProfile({actor: account.did})
+        // if it succeeds, update the selected account
+        setSelectedAccount(account)
```
mozzius: if we use a query as mentioned above, maybe we could use `setQueryData` here to reuse the Agent directly so we don't need to resume session twice

src/view/com/composer/Composer.tsx
```diff
@@ -878,13 +975,82 @@ let ComposerPost = React.memo(function ComposerPost({
...
-          style={[a.mt_xs]}
-        />
+      <View style={[a.flex_row, a.align_start, isNative && a.flex_1]}>
+        {isWeb ? (
+          <Menu.Root>
+            <Menu.Trigger label={_(msg`Switch account`)}>
+              {({props}) => (
+                <Button
+                  {...props}
+                  disabled={otherAccounts.length === 0}
```
mozzius: Would be much neater to just have a single component that is just platform split at the file level for this button+dialog

src/view/com/composer/Composer.tsx
gaearon: yeah that's a side effect and should not be in useMemo

src/view/com/composer/Composer.tsx
```diff
@@ -209,6 +243,54 @@ export const ComposePost = ({
     createComposerState,
   )
 
+  const onSelectAccount = React.useCallback(
```
gaearon, line 285: make sure to check what happens if the async code below is slowed down to trigger a race condition of resolving in the opposite order

haileyok:
just wanted to say

based

src/view/com/composer/Composer.tsx
SapphoSys: @mozzius switched over to tanstack's useQuery if you wanna take a look :)

src/view/com/composer/Composer.tsx
SapphoSys: @mozzius done and done in the commits below

src/view/com/composer/Composer.tsx
SapphoSys: done in 49130caa54e468f7b9f7b6103d1525c93406b4ef

src/view/com/composer/Composer.tsx
SapphoSys: I quickly switched from account A to B to C, then back to A. Switching accounts has a (close to) 1 second delay (might be quicker with your connection, it varied for me), with subsequent switches being a lot faster
(this is with the useQuery & useQueryClient logic added in order to cache sessions)

I also decided to add a 1 second setTimeout promise delay just to check if the rapid fire switch requests would be out of order. I haven't noticed that being the case

SapphoSys: Though I could probably add a condition guard or something just in case... hm
@mozzius let me know if it'd be necessary here

src/view/com/composer/Composer.tsx
```diff
@@ -187,26 +187,29 @@ export const ComposePost = ({
...
-    const newAgent = new AtpAgent(session)
-    return newAgent
-  }, [selectedAccount, currentAccount, defaultAgent])
+  const {data: agent = defaultAgent} = useQuery({
+    queryKey: ['composer-agent', currentAccount, selectedAccount],
+    queryFn: async () => {
+      if (selectedAccount.did === currentAccount!.did) {
+        return defaultAgent
+      }
+      const session = new CredentialSession(new URL(selectedAccount.service))
```
mary-ext, line 225: shouldn't this have a listener for session updates, any session expiry here is going to result in the session getting marked as invalid (due to stale refresh token) when you do another post (with that account), or actually switch to that account later

selectedAccount should probably be a DID, instead of passing the entire Account object in

SapphoSys: ya good point, i just quickly put things together to get the switcher working in the composer but now that i look at it its redundant

src/view/com/composer/Composer.tsx
```diff
@@ -254,8 +276,15 @@ export const ComposePost = ({
...
-      if (account.did === selectedAccount.did) {
+    async (accountDid: string) => {
+      if (accountDid === selectedAccountDid) {
+        return
+      }
+
+      // get fresh account data from session store
+      const account = accounts.find(acc => acc.did === accountDid)
+      if (!account) {
+        setError('Account not found')
```
SapphoSys, line 294: i'm not 100% sure about the setError statement here tbh since the codebase only uses setError with either translated strings using `msg`, or to pass errors just as is (`e.message`, `data.error`, etc)
i originally wanted to use the throw statement, but wasn't sure it fit here. this feels a little hacky... hm

src/view/com/composer/Composer.tsx
SapphoSys: @mary-ext done. you can check out commits 1ccdbcc and 06aec44

src/view/com/composer/Composer.tsx
gaearon: to clarify, i mean that race conditions are always possible — so it’s good to find a way to trigger one reliably (eg by simulating alternating delays) and then to write the code in a way that it’s resilient to the race

src/view/com/composer/Composer.tsx
```diff
@@ -176,11 +180,70 @@ export const ComposePost = ({
...
+  const [selectedAccountDid, setSelectedAccountDid] = useState(
+    currentAccount!.did,
+  )
+
+  const {data: agent = defaultAgent} = useQuery({
+    queryKey: [
+      'composer-agent',
+      currentAccount?.did,
+      currentAccount?.service,
+      currentAccount?.active,
```
mozzius, line 194: is this necessary? queryClients get reset when you change account anyway

PR #9063, opened 2025-09-22
Explicitly declare worklets for useAnimatedScrollHandler_FIXED.web.ts

estrattonbailey:
Worklet validation was added for `3.19.0` in https://github.com/software-mansion/react-native-reanimated/pull/7604

Though the docblocks for this hook state that the callbacks are automatically workletized, the actual docs site for the hook states that the callbacks should be workletized manually. According to [this conversation](https://github.com/software-mansion/react-native-reanimated/issues/7493#issuecomment-3112869389) we need to workletize these callbacks.

For us, this seems to only affect web. And the nested worklets in this case don't seem to have any negative effect. You can just toss a log into the `FIXED` hook or it's higher-level usages and see that it fires just fine.

We added this due to https://github.com/software-mansion/react-native-reanimated/issues/5360, which I think is also fixed by https://github.com/software-mansion/react-native-reanimated/pull/7604. I'll investigate further.

### DIFF
--- a/src/lib/hooks/useAnimatedScrollHandler_FIXED.web.ts
+++ b/src/lib/hooks/useAnimatedScrollHandler_FIXED.web.ts
@@ -12,26 +12,31 @@ export const useAnimatedScrollHandler: typeof useAnimatedScrollHandler_BUGGY = (
   return useAnimatedScrollHandler_BUGGY(
     {
       onBeginDrag(e, ctx) {
+        'worklet'
         if (typeof ref.current !== 'function' && ref.current.onBeginDrag) {
           ref.current.onBeginDrag(e, ctx)
         }
       },
       onEndDrag(e, ctx) {
+        'worklet'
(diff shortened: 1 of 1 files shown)


estrattonbailey:
Yolo deleted this hook because I don't think we need it, will do some more testing here and using the repro branch linked above.

mozzius:
The specific issue/repro is this one: https://github.com/software-mansion/react-native-reanimated/issues/5360

You can see here the stale closure issue on web has indeed been fixed 🥳 this is the bug we had the "fixed" hook for so we can indeed remove it safely

https://github.com/user-attachments/assets/5b6716fb-f536-4eec-b733-1bd014ac1040

mozzius:
And native is also fixed, yay!

https://github.com/user-attachments/assets/726f19a2-1b89-49a4-bf61-62226cd2efdf

PR #9067, opened 2025-09-23
Language selection and suggestion UX improvements

estrattonbailey:
Builds on the excellent work in #8886. This PR does a few additional things:
- Ensures that the reply suggestion is only shown when the composer opens. Once the user begins typing, has selected a suggestion, or selected a different language manually, it will not show again.
- Edited the text when replying to a post to make it more clear why we're suggesting a language
- Ensures that for reply parents with multiple languages specified, if you're replying in any of those languages the prompt will not show
  - We still only suggest the first language specified, in any case
- Clarifies some things using named variables and callbacks

Post marked as `italian,spanish`:
CleanShot 2025-10-01 at 18 04 20@2x
Replying in `spanish`:
CleanShot 2025-10-01 at 18 05 10@2x

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -208,11 +208,10 @@ export const ComposePost = ({
   >(null)
 
   /**
-   * The language of the post being replied to, if any. We just use the first
-   * language available, for now.
+   * The language(s) of the post being replied to.
    */
-  const [replyToLanguage, setReplyToLanguage] = useState<string | undefined>(
-    replyTo?.langs?.[0],
+  const [replyToLanguages, setReplyToLanguages] = useState<string[]>(
+    replyTo?.langs || [],
   )
 
   /**
@@ -234,7 +233,7 @@ export const ComposePost = ({
    */
   const onSelectLanguage = () => {
     setAcceptedLanguageSuggestion(null)
-    setReplyToLanguage(undefined)
+    setReplyToLanguages([])
   }
 
   const [composerState, composerDispatch] = useReducer(
@@ -694,7 +693,7 @@ export const ComposePost = ({
     <>
       <SuggestedLanguage
         text={activePost.richtext.text}
-        replyToLanguage={replyToLanguage}
+        replyToLanguages={replyToLanguages}
         currentLanguages={currentLanguages}
         onAcceptSuggestedLanguage={setAcceptedLanguageSuggestion}
       />
--- a/src/view/com/composer/select-language/SuggestedLanguage.tsx
+++ b/src/view/com/composer/select-language/SuggestedLanguage.tsx
@@ -18,17 +18,30 @@ const cancelIdle = globalThis.cancelIdleCallback || clearTimeout
 
 export function SuggestedLanguage({
   text,
-  replyToLanguage: replyToLanguageProp,
+  replyToLanguages: replyToLanguagesProp,
   currentLanguages,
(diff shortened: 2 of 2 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -197,6 +198,45 @@ export const ComposePost = ({
   const [publishingStage, setPublishingStage] = useState('')
   const [error, setError] = useState('')
 
+  /**
+   * A temporarly local reference to a language suggestion that the user has
```
surfdude29:
```suggestion
   * A temporary local reference to a language suggestion that the user has
```
nit

src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -58,55 +55,121 @@ export function SuggestedLanguage({
...
+      replyToLanguage,
+      langPrefs.appLanguage,
+    )
+
+    return (
+      <LanguageSuggestionButton
+        label={
+          <RNText>
+            <Trans>
+              The post you're replying to is written in {suggestedLanguageName}.
```
surfdude29:
```suggestion
              The post you're replying to has this language tag: {suggestedLanguageName}.
```

Given that posts can have incorrect language tags, perhaps a string that refers more precisely to the data that the app knows – what the language tag is on the post being replied to – would work better here?

This would avoid situations where the UI displays mistaken information because a post has an incorrect language tag. For example, if the user is replying to a post that is written in English but has a language tag of German, the UI string would only correctly state that the language tag of the post is German, rather than claiming that the post is actually written in German, which the user could clearly see is false.

estrattonbailey: Yeah that's a good point. Is "language tag" the best way to describe this... will think on this. We're also only looking at the first language specified, I think we could improve that in the event of a post that is `english, german` and the user is replying in either of those languages. In that case I would think we don't need this nudge at all.

surfdude29: "has this language code:" or "is marked as being written in" are two other possibilities, I also considered "is tagged as" but I thought that might cause confusion with hashtags.

That's a good thought about when a user's replying to a post with two (or three?) language tags/codes.

Obviously completely out-of-scope for this PR, but on the subject of language tagging of posts: the suggestion that some users have made of providing an option to show the language setting right next to the **Post** button in the composer might be worth investigating at some point.

For users who regularly switch between posting languages, it's possible that the increased visibility of the setting could help significantly cut down on incorrectly tagged posts.

src/view/com/composer/Composer.tsx
```diff
@@ -211,11 +227,15 @@ export const ComposePost = ({
...
 
-  // This effect clears the temporary language suggestion if the post language
-  // is manually changed, so the user doesn't get stuck with the suggestion.
-  useEffect(() => {
+  /**
+   * When the user selects a language from the composer language selector,
+   * clear any temporary language suggestions they may have selected
+   * previously, and any we might try to suggest to them.
+   */
+  const onSelectLanguage = () => {
```
elijaharita, line 234: using a callback instead of an effect here is a lot cleaner, nice

estrattonbailey: `useEffect` def worked and was less code, but I tend to prefer making control flow as obvious as possible

mozzius (approved):

PR #9069, opened 2025-09-23
Start migrating blink. Also add better shutdown handling for ogcard

amckinley:


.github/workflows/build-and-push-link-aws.yaml
```diff
@@ -1,6 +1,11 @@
 name: build-and-push-link-aws
 on:
   workflow_dispatch:
+  pull_request:
+    paths:
+      - 'bskylink/**'
+      - 'Dockerfile.bskylink'
+      - '.github/workflows/build-and-push-link-aws.yaml'
```
haileyok, line 8: ooo nice

bskylink/src/bin.ts
```diff
@@ -1,29 +1,62 @@
...
-  const cfg = envToCfg(env)
-  if (cfg.db.migrationUrl) {
-    const migrateDb = Database.postgres({
-      url: cfg.db.migrationUrl,
-      schema: cfg.db.schema,
-    })
-    await migrateDb.migrateToLatestOrThrow()
-    await migrateDb.close()
-  }
+  try {
```
haileyok, line 4: since we are restructuring this, i generally like to scope down `try/catch` blocks in typescript as much as possible, so you can get the most descriptive errors. e.g. if `readEnv()` throws (i have no idea if it does), i'd make that have its own try/catch

```ts
let env: AppEnvironment
try {
	env = readEnv()
} catch(error) {
	// log error or whatever
	// exit process
}
```

but for now this is all fine, we can adjust more in a separate PR particularly as i have no idea what even to suggest we scope them down to, since again i have no idea what can actually throw here

bskylink/src/bin.ts
```diff
@@ -1,29 +1,62 @@
...
+      'Configuration loaded',
+    )
+
+    if (cfg.db.migrationUrl) {
+      httpLogger.info('Running database migrations')
+      const migrateDb = Database.postgres({
+        url: cfg.db.migrationUrl,
+        schema: cfg.db.schema,
+      })
+      await migrateDb.migrateToLatestOrThrow()
```
haileyok, line 26: like this for example i'd probably scope it its own try/catch so that you can do like

```ts
try{
	await migrateDb.migrateToLatestOrThrow()
} catch(error) {
	httpLogger.error({error}, 'Error migrating database')
	process.exit(1)
}
```

PR #9078, opened 2025-09-24
Sync threaded reply between apps/instances

ETJeanMachine:
Previously, the app was not actually storing the user's option to use a threaded/linear view when the option was selected in the modal available above replies; instead, just opting for whatever was present in the server, while not actively updating it. This fixes this issue by calling the function correctly with the save boolean as true, syncing the user's change to the server.

### DIFF
--- a/src/state/queries/usePostThread/index.ts
+++ b/src/state/queries/usePostThread/index.ts
@@ -47,7 +47,7 @@ export function usePostThread({anchor}: {anchor?: string}) {
     view,
     setView: baseSetView,
     prioritizeFollowedUsers,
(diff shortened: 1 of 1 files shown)


mozzius:
Hmm, I think this was the intended behaviour, so you could just switch quickly back and forth - you'd go to settings to change it permanently. do you think it should be sticky here?

ETJeanMachine:
I think it should - I don't think it's immediately obvious on the user end it isn't sticky, which can be confusing. Either it can be sticky here or it could be changed to be more obvious you should go to settings for the permanent change?

Issue #4605, opened 2024-06-23
User setting to toggle "like" visibility on posts


jetpacktuxedo:
**Is your feature request related to a problem? Please describe.**

Sometimes thinking about people liking your posts and worrying about counts can be stressful. 

**Describe the solution you'd like**

A toggle somewhere in the settings menu to hide likes. When like visibility is disabled it should hide the counter next to the like button (at a minimum on the user's own posts, but alternatively hiding it on all posts would work too, maybe a sub-option?) as well as preventing notifications about likes.

**Describe alternatives you've considered**

One could simply log off I suppose

**Additional context**

Very janky mockup I made in my phone:

![Screenshot_20240622-204555~3](https://github.com/bluesky-social/social-app/assets/3813932/132da621-3838-447f-bcdc-2dc099cd7284)

surfdude29:
This is a good idea to have as an option. It could also be extended perhaps to include repost and follower numbers as well (maybe with a configurable option), like the third-party [Skeets](https://www.skeetsapp.com/) app does as part of one of its 'digital health' settings:

![IMG_2413](https://github.com/bluesky-social/social-app/assets/149612116/65905e0c-add4-4d4f-8a52-574ce5cb7c4b)
![IMG_2414](https://github.com/bluesky-social/social-app/assets/149612116/3be873f9-9b4a-4d1e-920e-e6c316fd6ec6)

Although afaik Skeets doesn't hide notifications about likes.

pedipanol:
Seconding this, I think Skeet's approach is pretty good, similar to Cohost which I loved.

I don't have an apple device so I'd love for it to be integrated into the main app (if anyone know another third party app that does it please lmk)

Claromale:
> I have made a simple implementation of this over at https://github.com/Fuzzyzilla/social-app/tree/hide-numbers, it adds an option in the accessibility tab. Currently it only hides Like and Repost counts on posts and comments, as profiles followers/like notifications/etc. require design work that I am not comfortable making decisions on without more discussion :3
> 
> Some examples:
> 
> 1. It would be nice to hide follower counts on profiles without removing the functionality of viewing the followers list. While the "50 followers" text serves as a fairly obvious link to take you to that page, removing the count makes it far less clear that it is clickable, so a visual rework is necessary.
[...]

PR #9080, opened 2025-09-25
Add Metric/Impression Visibility Options

RetroSunstar:
This is a rough draft for adding the options to hide certain metrics, such as: Likes, Replies, Saves, Reposts, and Quote Posts 🐙

The request I found for this is: #4605 (Sorry if I missed any others!)

It currently lacks handling for Feed/Labeller likes or the Following/Followers/Posts value on profiles, in an attempt to keep the page cleaner and not overburden it with options 😇

It also doesn't touch the "Read # more replies" section, for 2 reasons, finding a middle ground for something I don't fully understand and the state of a feature-request for changing its naming to begin with.

For reposts, the option hides the count next to the button entirely at the moment, instead of following the quote post option and simply returning 0 in the equation. I felt that this made more sense than requiring both be off? (quotes return zero otherwise the number next to the repost button and "# Reposts" would be mismatched)

Any suggestions, prefered solutions, or changes would be great! 🩵🩷🩵🩷

(Found under Content & Media)
Screenshot 2025-09-25 052941

Screenshot 2025-09-25 054600


RetroSunstar:
Major rework on this since this feature's initial conception! 

Entirely rewritten on the backend to be much cleaner, with a complete menu redesign to allow for 4 variants in options with only 2 visible settings for the end-user. This differs from the original 3, allowing users to individually change each value from one of 4 states: Show, Hide on own posts, Hide on others' posts, and Hide on all posts

Though this menu is up for modifications if need be 🐙

image

PR #9086, opened 2025-09-25
Add `StackedButton` component

estrattonbailey:
Separate component because we won't have multiple sizes of this one, and it will always be displayed with both and icon and text.

In storybook and mocked it up in a sheet to show what it could look like.

CleanShot 2025-09-30 at 16 09 15@2x
CleanShot 2025-09-30 at 16 08 11@2x

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -837,3 +837,47 @@ export function ButtonIcon({
     </View>
   )
 }
+
+export type StackedButtonProps = Omit<
+  ButtonProps,
+  keyof VariantProps | 'children'
+> &
+  Pick<VariantProps, 'color'> & {
+    children: React.ReactNode
+    icon: React.ComponentType<SVGIconProps>
+  }
+
+export function StackedButton({children, ...props}: StackedButtonProps) {
+  return (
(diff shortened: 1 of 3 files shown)


mozzius:
Bop it _forever_??

src/components/Button.tsx
```diff
@@ -751,22 +771,29 @@ export function useSharedButtonTextStyles() {
...
 
     if (size === 'large') {
-      baseStyles.push(a.text_md, a.leading_snug, a.font_medium)
+      baseStyles.push(a.text_md, a.leading_snug, a.font_semi_bold)
     } else if (size === 'small') {
-      baseStyles.push(a.text_sm, a.leading_snug, a.font_medium)
+      baseStyles.push(a.text_sm, a.leading_snug, a.font_semi_bold)
     } else if (size === 'tiny') {
-      baseStyles.push(a.text_xs, a.leading_snug, a.font_medium)
+      baseStyles.push(a.text_xs, a.leading_snug, a.font_semi_bold)
```
mozzius: this font size change should only be for the stacked buttons, no?

PR #9091, opened 2025-09-27
Add embed support for Scaniverse 3D scans

elfprince13:
Recognizes and embeds Scaniverse scan links so that users can share and view 3D content in the Bluesky app.

Scaniverse's moderation policies can be found here: https://scaniverse.com/about/guidelines Scaniverse is available through the US Apple App Store, suggesting it is compliant with the same Apple T&Cs that Bluesky must comply with it. Scaniverse is a product of Niantic, a well known tech company.

### DIFF
--- a/__tests__/lib/string.test.ts
+++ b/__tests__/lib/string.test.ts
@@ -436,6 +436,13 @@ describe('parseEmbedPlayerFromUrl', () => {
 
     'https://www.flickr.com/groups/898944@N23/',
     'https://www.flickr.com/groups',
+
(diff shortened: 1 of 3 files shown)


elfprince13:
Okay - took some fiddling to find a node version where `yarn` would just work out of the box, but I've confirmed that this works via `yarn web` on a local build.

https://bsky.app/profile/elfprince13.mumak.app/post/3lztkso4sss2d equivalent on local client build if you want to test.

elfprince13:
@mozzius Ran `yarn prettier --write` should be good to go

PR #9094, opened 2025-09-29
Send inferrable interactions to third-party feeds

gaearon:
Currently, non-Bluesky algorithmic feeds are operating at an enormous disadvantage because they don't get the "seen" interaction. So they either have to aggressively prune all posts they've already served (which causes them to run out of content too quickly) or use [convoluted workarounds](https://bsky.app/profile/spacecowboy17.bsky.social/post/3ly4qrpprfs2q) that don't work well to approximate it.

The "For You" feed in particular is gaining popularity as a Discover alternative (anecdotally, Discover still doesn't seem to be very good) but this limitation is [hurting it](http://bsky.app/profile/spacecowboy17.bsky.social/post/3lzwuholkvc2n). This PR removes the limitation.

## Why this is good and fine

The original concern was that sending "seen" events back to the feed could potentially be privacy-sensitive and thus requires opt-in UI. I don't believe the premise of this argument is sound.

The thing is, if you scroll down more than ~30 items (i.e. if you're actually using the feed), it's going to request the next page regardless. Since the Bluesky client only operates in the "scroll" modality and not something more unusual, it implicitly means that you've scrolled past the items the feed has already served. It is logically inconsistent to say that "you scrolled past these posts" is sensitive information *to the feed that served them* to you if that information is obviously inferrable *by that same feed* from the subsequent pagination requests. The only items for which you have this "privacy" today is the tail of the feed, i.e. the last page. This is not a meaningful distinction to the user and therefore the whole limitation kind of doesn't make sense.

Note there is NO "linger" information being sent like how long you looked at something. This is also only being sent in the flat list of the feed itself, and NOT for navigations within (e.g. clicking into thread or profile).

On the other hand, the gains from this are significant. Third-party feeds no longer need to "guess" which items are actually OK to mark as read. Which is *particularly* important for the first page since that's where many people do refreshes. And refreshes not working feel broken, at least compared to Discover. (There's a separate question about UX of refreshes and allowing continuity there but we *should* have it separately—the fact that the Bluesky client can't maintain continuity across refreshes is a client-side issue, and is fixable on its own.)

## Test Plan

Reduce the delay, set breakpoints, verify only allowlisted entries go through. Verify the same via network tab.

Verify clickthroughs (click link, click author) still do NOT get sent. Verify "share" does NOT get sent. Verify clicking into some thread and navigating around does NOT send anything new. Verify it does NOT get sent for feeds that don't have feedback turned on.

### DIFF
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -28,9 +28,21 @@ import {useAgent} from './session'
 
 export const FEEDBACK_FEEDS = [...PROD_FEEDS, ...STAGING_FEEDS]
 
-export const DIRECT_FEEDBACK_INTERACTIONS = new Set<
+export const THIRD_PARTY_ALLOWED_INTERACTIONS = new Set<
   AppBskyFeedDefs.Interaction['event']
->(['app.bsky.feed.defs#requestLess', 'app.bsky.feed.defs#requestMore'])
+>([
+  // These are explicit actions and are therefore fine to send.
+  'app.bsky.feed.defs#requestLess',
+  'app.bsky.feed.defs#requestMore',
+  // These can be inferred from the firehose and are therefore fine to send.
+  'app.bsky.feed.defs#interactionLike',
+  'app.bsky.feed.defs#interactionQuote',
+  'app.bsky.feed.defs#interactionReply',
+  'app.bsky.feed.defs#interactionRepost',
+  // This can be inferred from pagination requests for everything except the very last page
+  // so it is fine to send. It is crucial for third party algorithmic feeds to receive these.
+  'app.bsky.feed.defs#interactionSeen',
+])
 
 const logger = Logger.create(Logger.Context.FeedFeedback)
 
@@ -228,7 +240,7 @@ function isInteractionAllowed(
     return false
   }
   const isDiscover = isDiscoverFeed(feed.feedDescriptor)
-  return isDiscover ? true : DIRECT_FEEDBACK_INTERACTIONS.has(interaction)
+  return isDiscover ? true : THIRD_PARTY_ALLOWED_INTERACTIONS.has(interaction)
 }
 
 function toString(interaction: AppBskyFeedDefs.Interaction): string {


src/state/feed-feedback.tsx
```diff
@@ -28,9 +28,21 @@ import {useAgent} from './session'
...
 
-export const DIRECT_FEEDBACK_INTERACTIONS = new Set<
+export const THIRD_PARTY_ALLOWED_INTERACTIONS = new Set<
   AppBskyFeedDefs.Interaction['event']
->(['app.bsky.feed.defs#requestLess', 'app.bsky.feed.defs#requestMore'])
+>([
+  // These are explicit actions and are therefore fine to send.
+  'app.bsky.feed.defs#requestLess',
+  'app.bsky.feed.defs#requestMore',
+  // These can be inferred from the firehose and are therefore fine to send.
```
gaearon, line 37: I added these for good measure because they're always obviously inferrable as public information, and correlating "this was served + there was an interaction" is practically close enough to just getting it directly. (But getting it directly is nicer for feeds that aren't tracking everything on the firehose.)

src/state/feed-feedback.tsx
```diff
@@ -28,9 +28,21 @@ import {useAgent} from './session'
...
+  'app.bsky.feed.defs#interactionLike',
+  'app.bsky.feed.defs#interactionQuote',
+  'app.bsky.feed.defs#interactionReply',
+  'app.bsky.feed.defs#interactionRepost',
```
gaearon, line 41: I added these for good measure because they're always obviously inferrable as public information, and correlating "this was served + there was an interaction" is practically close enough to just getting it directly. (But getting it directly is nicer for feeds that aren't tracking everything on the firehose.)

estrattonbailey (approved):
Thanks, Dan! This actually came up recently but we hadn't picked up the thread yet, so perfect timing. Your logic is sound, as always, we're all +1

mozzius:
Yay!

PR #9099, opened 2025-09-29
Fix feedfeedback metrics not distinguishing which feed it's from

mozzius:
Since #8672, we now allow show more/show less on third party feeds. However, we didn't update the metrics and thus that's all being logged as `discover:showMore`/`discover:showLess`.

I've added some metrics - generic `feed:showMore` etc for all the previously `discover:`-prefixed events, which now includes the feed descriptor. ~~I've opted to still send `discover:` ones when using discover for backwards compatibility, but not to statsig~~. For forwards compatibility I have made sure to include seen/interacted events in case we merge #9094 

## test plan

look at logs, confirm new events are being sent. confirm `discover:` ones are only sent when on discover

### DIFF
--- a/src/state/feed-feedback.tsx
+++ b/src/state/feed-feedback.tsx
@@ -82,7 +82,6 @@ export function useFeedFeedback(
           flushToStatsig(
             aggregatedStats.current,
             feed?.feedDescriptor ?? 'unknown',
-            isDiscover,
           ),
         45e3,
         {
@@ -90,7 +89,7 @@ export function useFeedFeedback(
           trailing: true,
         },
       ),
-    [feed?.feedDescriptor, isDiscover],
+    [feed?.feedDescriptor],
   )
 
   const sendToFeedNoDelay = useCallback(() => {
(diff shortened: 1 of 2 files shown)


src/state/feed-feedback.tsx
```diff
@@ -301,28 +336,52 @@ function sendOrAggregateInteractionsForStats(
...
 
   if (stats.clickthroughCount > 0) {
-    logEvent('discover:clickthrough', {
+    if (isDiscover) {
+      logger.metric(
+        'discover:clickthrough',
+        {count: stats.clickthroughCount},
+        {statsig: false},
+      )
+    }
```
estrattonbailey: I think the backwards compat layer is the right move for now, let's us gradually move to the new agnostic metric

estrattonbailey (approved):
I'm +1. Let's make a ticket or stage another draft PR to remove the compat layer later on?

mozzius:
Ok confirmed with Ian that he does not use these metrics at all, so I think we should just skip the backwards compat step and do a hard switch to the new metric. What do you think?

mozzius:
Okay removed

PR #9108, opened 2025-09-29
[Fix Logouts] Remove buggy hackfix

gaearon:
## Depends on https://github.com/bluesky-social/atproto/pull/4238

Once https://github.com/bluesky-social/atproto/pull/4238 is merged, this removes the hackfix working around it. That resolves one of the potential issues that @mackuba identified in https://github.com/bluesky-social/social-app/issues/8695#issuecomment-3103433871, namely the `lastSession` variable getting stale.

I don't know if there's an actual repro case for that problem, but removing clearly buggy code in favor of upstream fix seems like a good idea anyway.

## Test Plan

Apply this PR first. We should see bad behavior on network errors, same as before the hackfix was added in https://github.com/bluesky-social/social-app/pull/4911.

Then apply https://github.com/bluesky-social/atproto/pull/4238. The test plan from https://github.com/bluesky-social/social-app/pull/4911 should pass although the hackfix is no longer present.

I have not done this myself yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "icons:optimize": "svgo -f ./assets/icons"
   },
   "dependencies": {
-    "@atproto/api": "^0.16.7",
+    "@atproto/api": "^0.17.0",
     "@bitdrift/react-native": "^0.6.8",
     "@braintree/sanitize-url": "^6.0.2",
     "@bsky.app/alf": "^0.1.2",
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -321,15 +321,7 @@ class BskyAppAgent extends BskyAgent {
 
     // Now the agent is ready.
     const account = agentToSessionAccountOrThrow(this)
-    let lastSession = this.sessionManager.session
     this.persistSessionHandler = event => {
-      if (this.sessionManager.session) {
-        lastSession = this.sessionManager.session
-      } else if (event === 'network-error') {
-        // Put it back, we'll try again later.
-        this.sessionManager.session = lastSession
-      }
-
       onSessionChange(this, account.did, event)
       if (event !== 'create' && event !== 'update') {
         addSessionErrorLog(account.did, event)
--- a/src/state/session/agent.ts
+++ b/src/state/session/agent.ts
@@ -322,6 +322,7 @@ class BskyAppAgent extends BskyAgent {
     // Now the agent is ready.
     const account = agentToSessionAccountOrThrow(this)
     this.persistSessionHandler = event => {
+      console.log('persistSessionHandler', {event})
       onSessionChange(this, account.did, event)
       if (event !== 'create' && event !== 'update') {
         addSessionErrorLog(account.did, event)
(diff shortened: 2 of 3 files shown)


gaearon:
I'm actually not even sure how to reproduce the original issue because I'm not sure how to forcefully get a `network-error`. I'm sure there was some way (maybe captive wifi? broken wifi?) but turning off wifi doesn't do the job.

I think this fix is correct but this will need someone from the team to shepherd as I'm unlikely to test this in depth.

ThisIsMissEm:
> I'm actually not even sure how to reproduce the original issue because I'm not sure how to forcefully get a `network-error`. I'm sure there was some way (maybe captive wifi? broken wifi?) but turning off wifi doesn't do the job.
> 
> I think this fix is correct but this will need someone from the team to shepherd as I'm unlikely to test this in depth.

Yeah, probably using the developer tools and setting the network mode to something lossy would do the trick?

mackuba:
You could try this thing: https://www.avanderlee.com/debugging/network-link-conditioner-utility/ (just don't forget to turn it off…)

estrattonbailey:
> You could try this thing: https://www.avanderlee.com/debugging/network-link-conditioner-utility/ (just don't forget to turn it off…)

Yeah this is what I've used in the past

mackuba:
Tbh, I haven't been able to manually trigger this condition where it would restore an expired session in that spot because of a network error, even with the Link Conditioner… so I can't confirm it. But also I remember being at the Ahoy conference in Hamburg in the spring on a bad hotel WiFi, and being logged out then all the time…

estrattonbailey:
I'm struggling as well. Maybe I haven't used Network Link Conditioner for this in the past 🤔 

As noted [here](https://github.com/bluesky-social/atproto/pull/4238#issuecomment-3352973267), this handling was originally added for unknown and server errors. Gonna spend a little more time trying to repro using the local `dev-env` package, but in any case, the logic here looks sound to me and I believe it should accomplish the same goals as the previous implementation.

estrattonbailey:
Ok partial repro here https://github.com/bluesky-social/atproto/pull/4242

Appears that `BskyAgent.prepare` races and may not be called in time, so on a cold load we actually end up hitting the debugger before the `persistSessionHandler` is attached. That means there's a case where this handling wasn't even used. Even prior to Dan's upstream patch the session shouldn't have been unset.

I gotta pause on this, would be curious to know if anyone can confirm this finding.

gaearon:
Yeah I wouldn’t expect this to work reliably for the initial load, just for some cases of losing network midway.

estrattonbailey:
> Yeah I wouldn’t expect this to work reliably for the initial load, just for some cases of losing network midway.

Oh then I should point out that `network-error` is only emitted during a `resumeSession` call. Other network requests internally call `refreshSession`, which will emit `expired` events and rug the session if the token is legit expired, but otherwise won't emit anything.

estrattonbailey (approved):
With the fix upstream we can merge this. I'm not positive of the impact, but it's a nice cleanup regardless, and should match existing handling. Thanks again for everyone's help on this 🙏

estrattonbailey:
> Appears that BskyAgent.prepare races and may not be called in time

This could also be a clue. If `prepare` isn't called in time, then a legit session update (like from the `isSessionExpired`) sent to `persistSessionHandler` may not be handled, and therefore may not be broadcast to other tabs.

PR #9109, opened 2025-09-29
[Fix Logouts] Persist accounts synchronously

gaearon:
This should address the second problem @mackuba identified in https://github.com/bluesky-social/social-app/issues/8695.

Concretely, this one: https://github.com/bluesky-social/social-app/issues/8695#issuecomment-3123273071.

*Note: Although this PR complements https://github.com/bluesky-social/social-app/pull/9108, it doesn't depend on it.*

## The Problem

There may be an async gap between a network callback from a backgrounded tab's agent event (which calls `dispatch`) and the corresponding `useReducer` execution + subsequent `useEffect` that writes the `state` into persisted state.

## The Fix

There may be some simpler way to write this. But for now, I'd suggest to keep the existing logic (for safety) but drop `useReducer` entirely. Instead, we'll add a custom store class we hook up to React via `useSyncExternalStore`.

The store calls our reducer logic but makes sure to persist to local storage immediately. So we don't need to wait for a React render cycle or batching or any effects. It gets written right away.

Note that `dispatch` is referentially stable (thanks to the store being kept in state) so we shouldn't be changing anything about memoization etc. This should be a purely internal change for what `dispatch` does.

## Test Plan

The usual with accounts — log in, log out, open many tabs, verify they're updating in unison, etc. @estrattonbailey would know what to do. 

This does need careful testing. I haven't tested it myself yet.

### DIFF
--- a/src/state/session/index.tsx
+++ b/src/state/session/index.tsx
@@ -51,6 +51,7 @@ class SessionStore {
   private listeners = new Set<() => void>()
 
   constructor() {
+    // Careful: By the time this runs, `persisted` needs to already be filled.
     const initialState = getInitialState(persisted.get('session').accounts)
     addSessionDebugLog({type: 'reducer:init', state: initialState})
     this.state = initialState
@@ -69,6 +70,8 @@ class SessionStore {
 
   dispatch = (action: Action) => {
     const nextState = reducer(this.state, action)
+    this.state = nextState
+    // Persist synchronously without waiting for the React render cycle.
     if (nextState.needsPersist) {
       nextState.needsPersist = false
       const persistedData = {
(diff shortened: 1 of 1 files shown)


gaearon:
OK, I think this should be ready for review/testing. I've done some casual testing on the web and it seemed to work. I have not tested on iOS/Android and probably won't be able to (apparently my system is out of date).

src/state/session/index.tsx
```diff
@@ -46,29 +46,67 @@ const ApiContext = React.createContext<SessionApiContext>({
...
-export function Provider({children}: React.PropsWithChildren<{}>) {
-  const cancelPendingTask = useOneTaskAtATime()
-  const [state, dispatch] = React.useReducer(reducer, null, () => {
+class SessionStore {
+  private state: State
+  private listeners = new Set<() => void>()
+
+  constructor() {
+    // Careful: By the time this runs, `persisted` needs to already be filled.
     const initialState = getInitialState(persisted.get('session').accounts)
```
ThisIsMissEm, line 55: What happens in the case where there is no persisted session or no accounts? Could this result in an error?

gaearon: The `persisted` abstraction fills that in with default values (i.e. we get `[]` here). This part is similar to how it worked before, I just moved it to a class.

gaearon: The timing caveat is just that the initial read is async. We delay the component tree from mounting until then but reading in module scope would've read too early.

estrattonbailey:
Dig it, will look in the morning

mackuba:
I'll try to get these changes merged into my test instance that does all that logging, and then test it there for a week or two

mackuba:
Oh no I said "instance" :(

Issue #9095, opened 2025-09-29
Navigating between posts in the thread often loses scroll position


gaearon:
### Steps to Reproduce

See here. It's starting in the right place and then scrolls to top. This is annoying and bad.

https://github.com/user-attachments/assets/f1be4c28-f0bb-4a8b-ba3f-cd4a135ac8d9

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

main

gaearon:
I suspect the problem is that this logic in `PostThread`:

https://github.com/bluesky-social/social-app/blob/d3dbb9468949e5122a18c1e07e28fe76cbeb5d36/src/screens/PostThread/index.tsx#L161-L215

might have been written primarily for the "hard refresh" use case than for navigation between replies.

For hard refresh it works fine:

https://github.com/user-attachments/assets/e2f68c3e-0897-46dd-96f5-9d31e9b2486f

But for navigation in-thread, it jumps around because it disables `shouldHandleScroll.current`:

https://github.com/user-attachments/assets/462e36db-a841-4cef-a0fa-e07c64e290e9

Maybe it shouldn't disable it in some cases?

gaearon:
Btw previously we didn't have `deferParents` on the web at all:

https://github.com/bluesky-social/social-app/blob/e6d26186a9e529069b392dbe909dd059295c615b/src/view/com/post-thread/PostThread.tsx#L151-L155

I wonder if it's necessary now...

gaearon:
https://github.com/bluesky-social/social-app/pull/9111

PR #9111, opened 2025-09-30
[Web] Fix thread jumps

gaearon:
Fixes https://github.com/bluesky-social/social-app/issues/9095

This logic is luckily web-only.

To repro, you have to navigate to a reply *and then* navigate to another one (e.g. upwards).

I don't understand the logic quite enough to tell what's happening, but for some reason we were setting `shouldHandleScroll.current = false` a bit too early. This function runs an extra time in this scenario and we end up missing the last event when we actually *should* scroll.

I changed the condition to what seems to make it work: only stop attempting after one positive scroll. I don't know if this is a perfect heuristic but seems to work at least on Chrome. You'll probably want to test FF, Safari, and Mobile Safari.

## Test Plan

I was doing stuff like this. Previously it would lose scroll position.

https://github.com/user-attachments/assets/6b4e9699-3990-4034-9186-2f6692b6f0bd

### DIFF
--- a/src/screens/PostThread/index.tsx
+++ b/src/screens/PostThread/index.tsx
@@ -148,7 +148,9 @@ export function PostThread({uri}: {uri: string}) {
    */
   const shouldHandleScroll = useRef(true)
   /**
-   * Called any time the content size of the list changes, _just_ before paint.
+   * Called any time the content size of the list changes. Could be a fresh
+   * render, items being added to the list, or any resize that changes the
+   * scrollable size of the content.
    *
    * We want this to fire every time we change params (which will reset
    * `deferParents` via `onLayout` on the anchor post, due to the key change),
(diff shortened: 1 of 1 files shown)


gaearon:
Curiously the bug repro's on main but not in prod. Idk what's up with that.

estrattonbailey:
> Curiously the bug repro's on main but not in prod. Idk what's up with that.

Yeah wth, that's odd

gaearon:
It's gotta be either dev/prod difference or a regression since last deploy. Would be good to confirm which is that, and why. I'm not entirely comfortable with the `> 0` condition because if it never happens (for one or the other reason), the scroll will keep readjusting on each resize.

estrattonbailey:
Must be something in https://github.com/bluesky-social/social-app/pull/8931. RN-for-web [doesn't seem to be the culprit](https://github.com/necolas/react-native-web/releases). Lot of changes in RN itself of course.

It seems that after `setDeferParents(true)` runs, `onContentSizeChange` fires 2x now, whereas before it ran 1x. Basically, I think it's running both before and after the DOM has updated to give us a positive offset for the anchor, and we're disabling scroll handling after the first pass.

The intent here is that if `shouldHandleScroll.current === true`, then we want to ensure that the anchor offset is reset back to `0`. So you nailed that.

I think your concern about the `> 0` condition is a good instinct. If there are _any_ parents of a post — including deleted tombstones and such — that offset value will be `> 0`. It will be `< 0` for the root post, but that's handled by the `isRoot` check.

I could also theoretically happen if we're expecting parents to be prepended and after `setDeferParents(true)` runs, and the API returns nothing or nothing is prepended for some reason. Then yeah `onContentSizeChange` will not run and we won't reset `shouldHandleScroll`. I don't think this case would blow us up though — actually it mostly works if we _never_ reset it, except for the resize case. Subsequent navigations or a content resize should recover the broken state.

I don't want to claim there aren't other edge cases, but I think this this one is unlikely to happen and recoverable in the case it does. I think your fix is good and we can roll with it for now 👍

PR #9115, opened 2025-09-30
Embed testbed with tweaks

mozzius:
Supercedes #8933 

Points the iframes to the actual Vite dev server, so you can dev with hot reloading

https://github.com/user-attachments/assets/507d4a29-92dc-421d-a9eb-747ea2d81902

### DIFF
--- a/bskyembed/README.md
+++ b/bskyembed/README.md
@@ -23,4 +23,4 @@ In another terminal window, run the snippet dev script:
 yarn dev-snippet
 ```
 
-Then, open `file:/<your files>/social-app/bskyembed/test.html` in your browser.
+You can then see the testbed page at http://localhost:5173/test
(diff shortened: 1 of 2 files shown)


bskyembed/test.html
```diff
@@ -0,0 +1,1083 @@
...
+          <div>Post from labeler</div>
+        </div>
+        <div class="item">
+          <div>Long name</div>
+        </div>
+        <div class="item">
+          <div>Long handle</div>
+        </div>
+      </div>
+    </section>
```
elijaharita: i missed this during my pass, but we can probably get rid of this section. otherwise looks good! was able to run and test locally

mozzius:
```suggestion
```

mozzius (commented):
will remove!

estrattonbailey (approved):

PR #9116, opened 2025-09-30
starter pack metrics

BinaryFiddler:
Add new starter pack related event metrics:

- Existing 'starterPack:create' event in already place
- Adding profile:addToStarterPack when click the the “Add to starter pack” item in the menu
- Adding starterPack:addUser when the user adds another user while editing the pack during the pack’s creation flow
- Adding starterPack:removeUser when the user removes another user while editing the pack during the pack’s creation flow

https://linear.app/blueskyweb/issue/APP-1425/metrics-event-add-to-starter-pack

### DIFF
--- a/src/logger/metrics.ts
+++ b/src/logger/metrics.ts
@@ -322,10 +322,10 @@ export type MetricEvents = {
       | 'SendViaChatDialog'
   }
   'starterPack:addUser': {
-    starterPackUri?: string
+    starterPack?: string
   }
   'starterPack:removeUser': {
-    starterPackUri?: string
+    starterPack?: string
   }
   'starterPack:share': {
     starterPack: string
--- a/src/components/dialogs/StarterPackDialog.tsx
+++ b/src/components/dialogs/StarterPackDialog.tsx
@@ -305,7 +305,7 @@ function StarterPackItem({
         listUri: listUri,
         actorDid: targetDid,
       })
-      logger.metric('starterPack:addUser', {starterPackUri})
+      logger.metric('starterPack:addUser', {starterPack: starterPackUri})
     } else {
       if (!starterPackWithMembership.listItem?.uri) {
         console.error('Cannot remove: missing membership URI')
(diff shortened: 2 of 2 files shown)


src/logger/metrics.ts
```diff
@@ -321,6 +321,8 @@ export type MetricEvents = {
       | 'ChatsList'
       | 'SendViaChatDialog'
   }
+  'starterPack:addUser': {}
+  'starterPack:removeUser': {}
```
mozzius: All these other ones include the `starterPark` uri, I think add that here too for consistency. You can make it optional, in case the starter pack doesn't exist yet

mozzius (changes requested):
These are fine but from what I understand from the linear ticket we need to be adding metrics to `src/components/dialogs/StarterPackDialog.tsx` too for the add/remove user part

src/logger/metrics.ts
BinaryFiddler: Alex said these are events emitted during the creation flow of a new starter pack which means there will never be any starterpack uri. Do we want to emit this when adding/removing user to existing starter pack too?

src/logger/metrics.ts
```diff
@@ -321,8 +321,12 @@ export type MetricEvents = {
       | 'ChatsList'
       | 'SendViaChatDialog'
   }
-  'starterPack:addUser': {}
-  'starterPack:removeUser': {}
+  'starterPack:addUser': {
+    starterPackUri?: string
+  }
+  'starterPack:removeUser': {
+    starterPackUri?: string
```
mozzius: nit

```suggestion
    starterPack?: string
```

src/logger/metrics.ts
```diff
@@ -321,8 +321,12 @@ export type MetricEvents = {
       | 'ChatsList'
       | 'SendViaChatDialog'
   }
-  'starterPack:addUser': {}
-  'starterPack:removeUser': {}
+  'starterPack:addUser': {
+    starterPackUri?: string
```
mozzius: nit

```suggestion
    starterPack?: string
```

PR #9119, opened 2025-10-02
[WIP] Lifecycle and intent handling improvements

### LINKED ISSUE #8513: Android share extension not working if app not open in background

kazssym:
This pull request updates the lifecycle handling in the `ExpoReceiveAndroidIntentsModule` to ensure that the initial intent is processed when the module starts, and that new intents are stored for later handling rather than processed immediately.

Lifecycle and intent handling improvements:

* Added an `OnStart` block to process the current intent (if any) when the module starts, ensuring that initial intents are handled properly.
* Modified the `OnNewIntent` block to store the new intent in the `intent` property instead of immediately processing it, allowing for deferred intent handling.

Probably fixes #8513 but not test.


kazssym:
I'm setting up a dev container to build and test the app for Android locally.

PR #9122, opened 2025-10-02
Tighten up eslint rule to catch unused `const {_} = useLingui()`

mozzius:
`_` is configured to be the exemption from the `no-unused-vars` rule. However, Lingui uses `_` as a variable name, which means it gets ignored from the rule.

As a result, we have a bunch of useless extra `useLingui()` calls all over the place. This PR:

1. Changes the exemption to exclude just a single `_` character (i.e. must be `_ignore` instead)
2. Fixes all the places where this now causes an error

Also, upgraded our React Compiler plugin version - we actually had two versions floating around in there which doesn't seem good. And removed an unused function which had like 10 eslint warnings


yarn.lock
```diff
@@ -8603,14 +8603,7 @@ babel-plugin-polyfill-regenerator@^0.6.1:
...
-  resolved "https://registry.yarnpkg.com/babel-plugin-react-compiler/-/babel-plugin-react-compiler-19.1.0-rc.1.tgz#99d131be61017e40abbaedd98321069bf8b7e54a"
-  integrity sha512-M4fpG+Hfq5gWzsJeeMErdRokzg0fdJ8IAk+JDhfB/WLT+U3WwJWR8edphypJrk447/JEvYu6DBFwsTn10bMW4Q==
-  dependencies:
-    "@babel/types" "^7.26.0"
-
-babel-plugin-react-compiler@^19.1.0-rc.2:
+babel-plugin-react-compiler@^19.1.0-rc.2, babel-plugin-react-compiler@^19.1.0-rc.3:
   version "19.1.0-rc.3"
   resolved "https://registry.yarnpkg.com/babel-plugin-react-compiler/-/babel-plugin-react-compiler-19.1.0-rc.3.tgz#45e5a282a2460b3701971e5eb8310a90a7919022"
   integrity sha512-mjRn69WuTz4adL0bXGx8Rsyk1086zFJeKmes6aK0xPuK3aaXmDJdLHqwKKMrpm6KAI1MCoUK72d2VeqQbu8YIA==
```
mozzius, line 8616: Looks like we had 2 of them 🙃

estrattonbailey (approved):

PR #9133, opened 2025-10-03
Value prop screen - polish,  convert to pager

mozzius:
This is what it was always supposed to be, but I skipped that in favour of shipping faster. I think the experiment is looking promising so let's polish it

- Platform splits the value prop, turns native into a pager
- Splits up the monolithic `StepFinished.tsx` file
- Couple style tweaks to the overall shell (i.e. removing scrollbar on android)

If we return to this in future the dots probably shouldn't scroll with the pager but for now that's not a needle-mover

https://github.com/user-attachments/assets/e4ecd64b-4e62-4345-b41e-c05cbff24a12

## Test plan

Go through onboarding on all platforms


src/screens/Onboarding/Layout.tsx
mozzius, line 1: Adding `height: 100%` and `min-height: 100%` here were required to get the pager to work. Need to make sure they don't break other screens but as far as I can tell it's harmless

PR #9138, opened 2025-10-04
Force refresh camera permission status after user returns from settings

### LINKED ISSUE #9088: Camera not Opening for Posting ( Functionality, Permissions and State Issues )
### Steps to Reproduce

Device / Operating System: Google Pixel 9a, Android 16
Bluesky Release: 1.108.0
( unsure if this is reproducible on other devices such as iPhone etc )

Description of issue.
WHEN on feed, click the button to write a new skeet. 
Click the Camera Icon.

Image

Case 1 ( OS permissions not yet granted ): 
- Prompted to grant app permissions 
- Clicking CTA on prompt does not open correct Camera setting on OS settings 
- After granting permission and going back to app, no change, Camera still not open ( see Case 2)

SHOULD: 
- If Permissions not granted, show prompt, display CTA button
- Clicking CTA button opens correct permission setting in OS

Case 2 ( After OS permissions granted ): 
- Nothing happens, Camera does not open to take a photo

SHOULD: 
- Open Camera to take a photo

Hack to open Camera FOLLOWING Case 2: 
Click "Gallery" button, then hit back and THEN Camera opens
Possibly related to a state management problem in whatever component or page manages this functionality. 

SHOULD: 
- Not behave like this. Clicking on gallery opens gallery. Backing out closes gallery and does not open Camera. Clicking camera opens camera. Backing out takes you back to the post and doesn't open anything else. 

FINALLY. If you follow the hack, you will find that a number of Camera "instances" have opened corresponding to the number of times you clicked the camera icon ( which is intuitive behaviour after clicking the icon and seeing no changes). This seems to indicate some kind of state management issue again. Suggest an investigation using a singleton to ensure only one instance of Camera can be opened ( and this instance can be destroyed on exit and created when required, or can be stored in a store using something like zustand and shared by any other components that might need an instance depending on the state of the code ). 

Image

### Attachments

_No response_

Android

### Device Info

 Google Pixel 9a, Android 16

 1.108.0

### Additional Information

_No response_

EvvTim:
Improves the camera permission handling by ensuring the permission status is re-checked when the user attempts to access the camera. Previously, if the permission was denied, the app would only show an alert and not attempt to refresh the status.

### Changes
- Updated `useCameraPermission` hook to handle `DENIED` permissions by re-requesting permission.
- Returns the updated permission status, allowing proper handling on both iOS and Android.
- Ensures `openPermissionAlert` is only shown if permission remains denied after re-request.

### Issue
Fixes: [#9088](https://github.com/bluesky-social/social-app/issues/9088)

### Testing:
1. Deny camera permission in the app.
2. Attempt to access the camera.
3. Hook now re-requests permission.
4. Alert is displayed only if permission remains denied.

### DIFF
--- a/src/lib/hooks/usePermissions.ts
+++ b/src/lib/hooks/usePermissions.ts
@@ -1,5 +1,8 @@
 import {Linking} from 'react-native'
-import {useCameraPermissions as useExpoCameraPermissions} from 'expo-camera'
+import {
+  PermissionStatus,
+  useCameraPermissions as useExpoCameraPermissions,
+} from 'expo-camera'
 import * as MediaLibrary from 'expo-media-library'
 
 import {isWeb} from '#/platform/detection'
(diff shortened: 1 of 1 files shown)


src/lib/hooks/usePermissions.ts
```diff
@@ -78,18 +81,29 @@ export function useVideoLibraryPermission() {
...
+      !permission ||
+      permission.status === PermissionStatus.UNDETERMINED ||
+      permission.canAskAgain
+    ) {
+      const updated = await requestPermission()
+      return updated?.granted ?? false
     }
+
+    if (permission.status === PermissionStatus.DENIED) {
+      const updated = await requestPermission()
```
mozzius, line 99: is this necessary since we've already checked `permission.canAskAgain`?

EvvTim: I know it looks like a duplicate check, but we call `requestPermission()` here to  get the latest status in case the use manually enabled the permission in the device settings. If it's still denied, we show the alert. This ensueres we handle both denial and manual changes reliably

mozzius: Surely we should just refresh the permission state when the app comes back from the background

nathanorm:
**question:** does this fix everything mentioned in ( https://github.com/bluesky-social/social-app/issues/9088 )? 

In particular, wondering about the case where multiple instances of the camera seem to open.

PR #9139, opened 2025-10-04
feat(link-meta): Tweaks `getLinkMeta` to whitelist trusted hostnames

### LINKED ISSUE #1672: Blue Sky's composer should gather social graph data for PDFs and other non-HTML content too
Right now, if you put a link ending in `.html` into the composer (on Web), and ask the website to generate a card, you can watch the network panel make a request to https://cardyb.bsky.app/v1/extract.

For example, this URL makes a request in the network panel when you ask to make a card:

https://foo.com/foo.html

But these URLs, ending in `.jpeg`, `.png`, `.pdf` and `.xml` do not:

https://foo.com/foo.jpeg

https://foo.com/foo.png

https://foo.com/foo.xml

https://foo.com/foo.pdf

I understand the reasoning: In theory, those file endings are telling Blue Sky that they will not have Social Graph information, since that information only exists in HTML content.

That theory is correct, but at [the website I run](https://www.courtlistener.com), we share millions of PDFs, and we have a neat hack in place to help fight misinformation and provide better details to our users. When we detect an open graph crawler, we redirect the crawler to an HTML page with open graph data (if it's not a crawler, we serve the PDF). I know that DocumentCloud also uses this trick.

This works on Twitter, Facebook, Slack, Mastodon and a bunch of other sites. As far as I know, Blue Sky is the only one where it doesn't work. 

**To Reproduce**

1. Paste this link into the web composer: https://foo.com/f.html
2. Open the browser's network panel.
3. Press the button in the composer to get the card.
4. Note that it returns an error (the link doesn't work), and that you see a request in the network panel:

    image

5. Change the URL to https://foo.com/f.pdf
6. Press the button in the composer to get the card.
7. Note that it made no requests and throws no error.

**Expected behavior**

Blue Sky should go to the URL, regardless of the file ending, and test if it's actually HTML or a PDF. Heck, some horribly misconfigured website might end links with .pdf even when serving HTML. :)

**Details**

 - Platform: Firefox on Linux
 - Platform version: 118

**Additional context**

This bug is a bit of a bummer because one of the things that drove me to Blue Sky is that Twitter removed headlines. This bug means that the links from my website don't have twitter cards or headlines either. Darn! 

I took a look around the code, but couldn't find where this is done. If somebody sends a pointer, we've got technical folks and volunteers that could help with this.

ERosendo:
### What’s changed

Previously, the `getLinkMeta` function would return early for all non-HTML URLs, skipping metadata fetching. This PR introduces a whitelist of trusted hostnames (e.g., `storage.courtlistener.com`) allowing the preview fetcher to process non-HTML links from these hosts and continue fetching metadata.

Additional hosts can be easily added to the whitelist if similar needs arise in the future. This PR closes #1672 .

### DIFF
--- a/src/lib/link-meta/link-meta.ts
+++ b/src/lib/link-meta/link-meta.ts
@@ -62,7 +62,10 @@ export async function getLinkMeta(
     likelyType,
     url,
   }
-  if (likelyType !== LikelyType.HTML) {
+  const trustedHostnames: string[] = ['storage.courtlistener.com']
+  const isTrusted = trustedHostnames.includes(urlp.hostname)
+  // Only allow non-HTML links to continue if they're from trusted sources
+  if (likelyType !== LikelyType.HTML && !isTrusted) {
     return meta
   }
 


mlissner:
Just chiming in to note that the special case we added here was suggested by Paul. Notes here:

https://github.com/bluesky-social/social-app/issues/1672#issuecomment-3366829788

And short chat here: https://bsky.app/profile/michaeljaylissner.com/post/3m2chmso2sk2v

Thank you! We're really excited to have our links supported on Bluesky.

src/lib/link-meta/link-meta.ts
```diff
@@ -62,7 +62,10 @@ export async function getLinkMeta(
     likelyType,
     url,
   }
-  if (likelyType !== LikelyType.HTML) {
+  const trustedHostnames: string[] = ['storage.courtlistener.com']
+  const isTrusted = trustedHostnames.includes(urlp.hostname)
```
mozzius: maybe call these variables `htmlExemptedHostnames` and `isExemptedFromHtmlCheck` or something for clarity?

mozzius (approved):
Works well! Nice

src/lib/link-meta/link-meta.ts
ERosendo: Sure! I'll rename the variables. Thanks for your suggestion.

ERosendo:
@mozziusm thanks for your review.  I've applied the requested changes and renamed the variables for better clarity:

- `trustedHostnames` is now `htmlExemptedHostnames`
- `isTrusted` is now `isExemptedFromHtmlCheck`

PR #9143, opened 2025-10-05
reply button in feeds opens thread

easrng:
<a href="https://bsky.app/profile/hailey.at/post/3m2g3t7ohtc2d">image</a>

### DIFF
--- a/src/view/com/posts/PostFeedItem.tsx
+++ b/src/view/com/posts/PostFeedItem.tsx
@@ -17,9 +17,11 @@ import {useQueryClient} from '@tanstack/react-query'
 import {useActorStatus} from '#/lib/actor-status'
 import {isReasonFeedSource, type ReasonFeedSource} from '#/lib/api/feed/types'
 import {MAX_POST_LINES} from '#/lib/constants'
+import {useOpenComposer} from '#/lib/hooks/useOpenComposer'
 import {usePalette} from '#/lib/hooks/usePalette'
 import {makeProfileLink} from '#/lib/routes/links'
 import {type NavigationProp} from '#/lib/routes/types'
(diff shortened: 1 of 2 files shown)


src/view/com/posts/PostFeedItem.tsx
```diff
@@ -185,16 +186,9 @@ let FeedItemInner = ({
       feedContext,
       reqId,
     })
```
mozzius: We should send the clickthrough event, not the reply event

mozzius:
We should probably add the accessibility properties (`role` etc) so it's accessible as a link

src/view/com/posts/PostFeedItem.tsx
```diff
@@ -185,16 +186,9 @@ let FeedItemInner = ({
...
-    openComposer({
-      replyTo: {
-        uri: post.uri,
-        cid: post.cid,
-        text: record.text || '',
-        author: post.author,
-        embed: post.embed,
-        moderation,
-        langs: record.langs,
-      },
```
mozzius, line 197: Hmm, I guess we still want to do this for the main post in a thread

And actually, maybe we still want it in a thread? And only make it a link in a feed? @haileyok thoughts?

mozzius:
~~Also, we need to consider the immersive video feed~~

src/view/com/posts/PostFeedItem.tsx
easrng: there are different components for thread vs feed vs notification vs video feed, I only changed the button in feeds

mozzius: Whoops sorry that was a very stupid comment, I didn't read the file name 🤦

PR #9150, opened 2025-10-06
[16KB] use 16kb-compatible fork of react-native-mmkv

mozzius:
Fork here: https://github.com/mozzius/react-native-mmkv/

Updated to the latest MMKV 1.3 LTS which supports 16KB page sizes

# Test plan

Confirm it compiles and mmkv still works on iOS+Android

You might want to check my commits in the fork: https://github.com/mozzius/react-native-mmkv/commits/main/


mozzius:
Hmm, doesn't work yet - I believe we need to update the NDK?

mozzius:
Got it working

PR #9153, opened 2025-10-07
check for self labels when determining if we should hide embeds

haileyok:
same thing as https://github.com/bluesky-social/social-app/commit/7f4e302b856e638dcfd39cf9f6f924ae3b3a85c0, but also checking in the self labels

cc @elijaharita


jimmylee (approved):
LGTM

PR #9163, opened 2025-10-08
Update react-related eslint, add `eslint-plugin-react-hooks`

mozzius:
React Compiler is 1.0.0! This PR:

1. Updates the babel plugin to 1.0.0, removes the unnecessary `react-compiler-runtime` package
2. Adds the `react-hooks` eslint config - the React Compiler eslint rules are now bundled into this
3. Fixes the eslint config since obviously that broke everything
4. Tries to fix as many new eslint errors as is reasonable

I recommend going through one commit at a time

- *Unused variables* - updating typescript-eslint meant that the `e` in `catch (e) {}` counts as an unused variable.
- *Purity errors* - `Math.random()` and `Date.now()` cannot be used in render anymore. I moved the random calls to the top level (since they don't need to change) and replaced `Date.now()` with our tick every minute hook. Also, `useState(Date.now())` needs to be `useState(() => Date.now())`, and I replaced some refs/memos that didn't actually change over time with state.
- *Immutability* - Mutating stuff in render is not allowed and it seems like the new lint rules started catching some cases that had slipped through before
- *Hoisting* - dumb but had to reorder some hook calls
- *Set state in effect* - other than one easy one, the cases we're doing this I don't really see how we'd do any other way. I set this rule to warn.

I ended up setting a few of the rules to `warn` because there's a fair few false positives / things that are too big to untangle unfortunately. A few examples:

### Using dispatch in reducer

Dan wrote this and I think it's fine? `dispatch` should be a stable function, but it complains "`dispatch` is accessed before it is declared"

```tsx
const [state, dispatch] = useReducer((state, action) => {
  switch (action) {
    case 'example':
      return {
        foo: 'bar',
        effect() {
          const id = setTimeout(() => dispatch('do-something-later'), waitMs)
          return () => {
            clearTimeout(id)
          }
        },
      }
  }
})

useEffect(() => {
  if (currentState.effect) {
    const effect = currentState.effect
    return effect()
  }
}, [currentState])
```

### Render functions that pass in event handlers that use refs

it doesn't like how `renderTabBar` is passed a function that uses a ref, even though it's used as an `onPress` handler and not used in render.

```tsx
const pagerView = useRef<PagerView>(null)

const onTabBarSelect = useCallback((index: number) => {
    pagerView.current?.setPage(index)
  },
  [pagerView],
)

return (
  <View>
    {renderTabBar({
      selectedPage,
      onSelect: onTabBarSelect,
    })}
    <PagerView ref={pagerView}>
      {children}
    </PagerView>
  </View>
)
```

### useMemo that applies a throttle to a callback

sendToFeedNoDelay uses a ref, so this useMemo gets flagged, even though it's only used as a callback

```tsx
  const sendToFeed = useMemo(
    () =>
      throttle(sendToFeedNoDelay, 10e3, {
        leading: false,
        trailing: true,
      }),
    [sendToFeedNoDelay],
  )
```


src/components/FeedInterstitials.tsx
```diff
@@ -140,6 +140,11 @@ function sortSeenPosts(postA: SeenPost, postB: SeenPost): 0 | 1 | -1 {
   }
 }
 
+const random1 = Math.random()
+const random2 = Math.random()
+const random3 = Math.random()
+const random4 = Math.random()
```
mozzius, line 146: I think it's ok that it doesn't change between hooks?

PR #9173, opened 2025-10-09
[Threadgate] Tweak threadgate buttons

mozzius:
Make them a little less subtle

- Added little chevrons
- In composer, made it a little larger
- In thread, made it blue if it's editable
- Gave it press states
- Made the timestamp on the thread use the `medium` style on small breakpoints

This will be the base PR for threadgate improvements

- https://github.com/bluesky-social/social-app/pull/9173
  - https://github.com/bluesky-social/social-app/pull/9350
  - https://github.com/bluesky-social/social-app/pull/9342
    - https://github.com/bluesky-social/social-app/pull/9343
    - https://github.com/bluesky-social/social-app/pull/9341

Screenshot 2025-11-03 at 16 29 25
Screenshot 2025-11-03 at 16 29 36

### DIFF
--- a/src/components/Select/index.web.tsx
+++ b/src/components/Select/index.web.tsx
@@ -1,4 +1,4 @@
-import {createContext, forwardRef, useContext, useMemo} from 'react'
+import {createContext, forwardRef, Fragment, useContext, useMemo} from 'react'
 import {View} from 'react-native'
 import {Select as RadixSelect} from 'radix-ui'
 
@@ -138,7 +138,11 @@ export function Icon({style}: IconProps) {
   )
 }
 
-export function Content<T>({items, renderItem}: ContentProps<T>) {
+export function Content<T>({
+  items,
+  renderItem,
+  valueExtractor = defaultItemValueExtractor,
+}: ContentProps<T>) {
   const t = useTheme()
   const selectedValue = useContext(SelectedValueContext)
 
@@ -196,7 +200,11 @@ export function Content<T>({items, renderItem}: ContentProps<T>) {
             <ChevronUpIcon style={[t.atoms.text]} size="xs" />
           </RadixSelect.ScrollUpButton>
           <RadixSelect.Viewport style={flatten([a.p_xs])}>
-            {items.map((item, index) => renderItem(item, index, selectedValue))}
+            {items.map((item, index) => (
+              <Fragment key={valueExtractor(item)}>
+                {renderItem(item, index, selectedValue)}
+              </Fragment>
+            ))}
           </RadixSelect.Viewport>
           <RadixSelect.ScrollDownButton style={flatten(down)}>
             <ChevronDownIcon style={[t.atoms.text]} size="xs" />
@@ -205,6 +213,10 @@ export function Content<T>({items, renderItem}: ContentProps<T>) {
       </RadixSelect.Content>
     </RadixSelect.Portal>
   )
+}
+
+function defaultItemValueExtractor(item: any) {
+  return item.value
 }
 
 const ItemContext = createContext<{


mozzius:
@github-actions ota

src/components/WhoCanReply.tsx
```diff
@@ -84,7 +89,14 @@ export function WhoCanReply({post, isThreadAuthor, style}: WhoCanReplyProps) {
       Keyboard.dismiss()
     }
     if (isThreadAuthor) {
-      editDialogControl.open()
+      // wait on prefetch if it manages to resolve in under 200ms
+      // otherwise, proceed immediately and show the spinner -sfn
```
estrattonbailey, line 93: Nicely done

src/storage/hooks/threadgate-nudged.ts
```diff
@@ -0,0 +1,9 @@
+import {device, useStorage} from '#/storage'
+
+export function useThreadgateNudged() {
```
estrattonbailey, line 3: These hooks don't share state if they're used in multiple places, which would lead to them going out of sync. Do we really need this as a separate file or can we just use the `useStorage` hook in `ThreadgateBtn` directly?

src/view/com/composer/threadgate/ThreadgateBtn.tsx
```diff
@@ -29,15 +43,82 @@ export function ThreadgateBtn({
...
       Keyboard.dismiss()
     }
 
+    setShowTooltip(false)
+    setThreadgateNudged(true)
+
     control.open()
   }
 
+  const defaultThreadgateAllowUISettings = threadgateRecordToAllowUISetting({
```
estrattonbailey: Tiny naming nit: `default*` on this and the next are more like `prefs*`. The defaults are `everybody` and `[]`. Ignore this if you want, just a tiny intent clarification thing ✌️

src/components/dialogs/PostInteractionSettingsDialog.tsx
```diff
@@ -330,229 +326,347 @@ export function PostInteractionSettingsForm({
...
+                  <Panel.Panel active={selected} adjacent="both">
+                    <Toggle.Checkbox />
+                    <Panel.PanelText>
+                      <Trans>People you mention</Trans>
+                    </Panel.PanelText>
+                  </Panel.Panel>
+                )}
+              </Toggle.Item>
+
+              <Button
```
estrattonbailey, line 510: Need a `disabled` here when you're not the OP 😛

estrattonbailey: Oh unless the OP has lists selected, this lets you open it to see those lists... so maybe fine

src/components/forms/Toggle/Panel.tsx
estrattonbailey, line 1: Not a big deal at all, but yeah we could put this alongside Toggle. Just not sure where else we'd use this. If we ever do, pry best to have a couple use cases ready before we abstract it. Your call!

src/components/forms/Toggle/Panel.tsx
```diff
@@ -0,0 +1,120 @@
...
+    />
+  )
+}
+
+/**
+ * A group of panels. TODO: auto-leading/trailing
+ */
+export function PanelGroup({children}: {children: React.ReactNode}) {
+  return <View style={[a.w_full, a.gap_2xs]}>{children}</View>
+}
```
estrattonbailey, line 120: There's a similar pattern in `Menu.Group`

estrattonbailey (approved):
Couple small things that don't really matter, great work!

src/storage/hooks/threadgate-nudged.ts
mozzius: Hmm, was just following convention, but yeah raw useStorage is simpler

src/components/forms/Toggle/Panel.tsx
mozzius: Done - I actually quite like `Toggle.Panel`

src/components/forms/Toggle/Panel.tsx
mozzius: Will return to this in future

src/view/com/composer/threadgate/ThreadgateBtn.tsx
mozzius: Good shout

src/components/dialogs/PostInteractionSettingsDialog.tsx
mozzius: Yeah I think this is fine (also, it's impossible to get this screen if you're not OP, unsure why we have this logic...)

src/storage/hooks/threadgate-nudged.ts
mozzius: I'll leave it for now and give some thought as to how we want this to work holistically

src/components/dialogs/PostInteractionSettingsDialog.tsx
estrattonbailey: It's possible because this dialog does double duty for postgate settings for a user's (non-OP) reply

src/storage/hooks/threadgate-nudged.ts
estrattonbailey: Yeah these hooks are kind of a footgun. Like this hook shouldn't be used in multiple places, basically, since different instances won't stay in sync. Easy to get confused. Raw hook has similar issues though. Really we need like a Zustand atom or something here lol

mozzius: Yeah - I mean I tried to make them sync I think? But react-native-mmkv doesn't quite work right

PR #9188, opened 2025-10-10
Combine SubtleWebHover and SubtleHover, default to web only for all cases

estrattonbailey:
This was WIP during threads v2 project. This fixes the immediate color contrast need and combines the component into one that applies to web by default, and optionally on native. I removed all native support, which in hindsight wasn't the right call imo.


mozzius:
Hmm, can we restore it on the trending topics cards? Quite good to have tap feedback there

src/components/SubtleHover.tsx
```diff
@@ -18,17 +25,34 @@ export function SubtleHover({style, hover}: ViewStyleProp & {hover: boolean}) {
...
     />
   )
+
+  if (hover) {
+    console.log({
+      isWeb,
+      web,
+      isNative,
+      native,
+    })
```
mozzius, line 48: Log

PR #9189, opened 2025-10-10
prompt user to log in on hashtag screen

BinaryFiddler:
- Using the same design and wording as on the search screen

Screenshot 2025-10-10 at 2 16 50 PM

### DIFF
--- a/src/screens/Hashtag.tsx
+++ b/src/screens/Hashtag.tsx
@@ -168,8 +168,7 @@ function HashtagScreenTab({
   const initialNumToRender = useInitialNumToRender()
   const [isPTR, setIsPTR] = React.useState(false)
   const t = useTheme()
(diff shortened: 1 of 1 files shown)


src/screens/Hashtag.tsx
```diff
@@ -161,6 +167,9 @@ function HashtagScreenTab({
   const {_} = useLingui()
   const initialNumToRender = useInitialNumToRender()
   const [isPTR, setIsPTR] = React.useState(false)
+  const t = useTheme()
+  const {currentAccount} = useSession()
+  const isLoggedin = Boolean(currentAccount?.did)
```
mozzius: You can just get the bool directly from `useSession`
```suggestion
  const {hasSession} = useSession()
```

PR #9195, opened 2025-10-13
Fix label spacing

mozzius:
Makes the label component have bottom margin only

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>


mozzius:
Screenshot 2025-10-13 at 17 35 02

Fine imo, no worse than before

PR #9205, opened 2025-10-15
feat: PostControlsSkeleton

elijaharita:
This is the skeleton version of the PostControls element, shown as a placeholder when that element is loading. ThreadItemPost, ThreadItemTreePost, and ThreadItemAnchor have been updated to use the new component in this PR.

Layout shift is demonstrated below by placing the skeleton component inline below the real PostControls component.

Desktop:
image

Mobile (Android):
image


elijaharita:
Updated skele item sizes, now appears as follows

image

estrattonbailey (approved):
Nice looks great!

PR #9209, opened 2025-10-15
Remove notice in create account flow

estrattonbailey:
The current versions of these policies are now in effect, we can remove this notice.

### DIFF
--- a/src/screens/PostThread/index.tsx
+++ b/src/screens/PostThread/index.tsx
@@ -428,7 +428,6 @@ export function PostThread({uri}: {uri: string}) {
               />
             )
           } else {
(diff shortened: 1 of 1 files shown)


src/screens/PostThread/index.tsx
```diff
@@ -428,6 +428,7 @@ export function PostThread({uri}: {uri: string}) {
               />
             )
           } else {
+            // return <ThreadItemPostSkeleton index={index} />
```
mozzius: nit: comment

estrattonbailey: oh lol leftover from reviewing that PR

PR #9213, opened 2025-10-15
[Reporting] Integrate new `ReportDialog` into DMs

estrattonbailey:
_Stacked on #9079_

Migrates chat reporting from the old, separately maintained report dialog, to the new combined `ReportDialog`, which includes all the new reporting reasons.

This PR:
- replaces and deletes old `/dms/ReportDialog`
- forks `ReportDialog` title depending on if it's a message being reported or the convo itself
- re-purposes the `DoneStep` from `/dms/ReportDialog` as `BlockOrDeleteDialog`, which opens after a report us successfully submitted
  - also added an error state in case we can't load the profile, which is just a "success" message that uses the same copy as the dialog were it to load properly

### New dialog
CleanShot 2025-10-15 at 18 35 38@2x
### Error state
CleanShot 2025-10-15 at 18 35 54@2x

### DIFF
--- a/src/components/dms/AfterReportDialog.tsx
+++ b/src/components/dms/AfterReportDialog.tsx
@@ -27,7 +27,11 @@ type ReportDialogParams = {
   message: ChatBskyConvoDefs.MessageView
 }
 
-export const BlockOrDeleteDialog = memo(function BlockOrDeleteDialogInner({
+/**
+ * Dialog shown after a report is submitted, allowing the user to block the
+ * reporter and/or leave the conversation.
+ */
+export const AfterReportDialog = memo(function BlockOrDeleteDialogInner({
   control,
   params,
   currentScreen,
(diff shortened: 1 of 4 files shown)


src/components/dms/BlockOrDeleteDialog.tsx
```diff
@@ -0,0 +1,215 @@
...
+import * as Toggle from '#/components/forms/Toggle'
+import {Loader} from '#/components/Loader'
+import {Text} from '#/components/Typography'
+
+type ReportDialogParams = {
+  convoId: string
+  message: ChatBskyConvoDefs.MessageView
+}
+
+export const BlockOrDeleteDialog = memo(function BlockOrDeleteDialogInner({
```
mozzius: nit: maybe it should be called `PostReportDialog` or `ReportFollowupDialog`

Issue #3161, opened 2024-03-09
Android image upload doesn't preserve selection order for multiple images


Juicysteak117:
**Describe the bug**
On Android (unsure if this is also an IOS issue), selecting multiple images does not upload them in the order that they are selected. This is especially tedious given that there's no way to rearrange images in the composer either, so images must be deleted and re-selected.

**To Reproduce**

1. On Android, open the post composer
2. Attempt to attach four images in a post
3. Observe.

**Expected behavior**

I expect the order to be consistent.

**Screenshots**

Selection order example:

![image](https://github.com/bluesky-social/social-app/assets/20595808/b059544f-78cc-4d67-990d-ba3c78fc1c29)

![image](https://github.com/bluesky-social/social-app/assets/20595808/00fc8e72-5ebc-4613-ac20-42fcb3c6f061)

Permutations are consistent. Appears to be nearly-cyclic with the only deviation being the 4-cycle as the (2 3) sub cycle doesn't permute. Issue is only when doing multiple image selections, as four single selections is obviously fine, but for two sets of two image selections the issue persists. Issue does not persist for a single followed by a double, however it does for a single followed by a triple, although upon testing, the image size seems like it may play a role given that in my test images (BENEDICT MATCHA) always becomes (MATCHA BENEDICT) but (MATCHA CHILI) remains (MATCHA CHILI). Further testing is required.

(1) -> (1)
(1 2) -> (2 1)
(1 2 3) -> (2 3 1)
(1 2 3 4) -> (4 2 3 1)
(1 2)(3 4) -> (2 1)(4 3)
(1 2 3)(4) -> (2 3 1)(4)
(1)(2 3 4) -> (1)(4 3 2)
(1)(2 3) -> (1)(2 3) !

**Details**

 - Platform: Android 
 - Platform version: 13
 - App version: Latest

**Additional context**

As a side note, #2794 is a bug that still persists which made this testing a bit more tedious.

haileyok:
I will peek at it, but I'm not actually sure it's fixable because of https://github.com/bluesky-social/social-app/pull/2384. The images come back in a whacky order and we don't get any info regarding the order they were selected in (nor can we even limit the number of photos you select in that screen...) Might be some solution though!

Re: #2794, I've been going through and migrating modals to the new UI, so that will be a focus for sure. Agreed that it's really subpar during my android testing 😓

Juicysteak117:
[...]

PR #9227, opened 2025-10-18
fix(composer): preserve selected assets order on image selection

tomo-x7:
# Summary
Preserve original assets order when selecting images for the composer so the order of selected photos is retained when attaching multiple images to a post.
Japan has many users who post manga with storylines. For them, disrupting the sequence is a critical issue.
# Example
When attaching multiple photos in the following order,
image
In this way, the order may become disrupted.
image
# Cause of the problem
## Problem code
```TS
const images: ComposerImage[] = []

await Promise.all(
  assets.map(async image => {
    const composerImage = await createComposerImage({
      path: image.uri,
      width: image.width,
      height: image.height,
      mime: image.mimeType!,
    })
    images.push(composerImage)
  }),
).catch(e => {
  logger.error(`createComposerImage failed`, {
    safeMessage: e.message,
  })
})
```
Concurrent pushes into a shared array cause a race condition that breaks the selected assets order.
## Fixed code
```TS
const images: ComposerImage[] = await Promise.allSettled(
  assets.map(async image => {
    const composerImage = await createComposerImage({
      path: image.uri,
      width: image.width,
      height: image.height,
      mime: image.mimeType!,
    })
    return composerImage
  }),
).then(res =>
  res
    .map<ComposerImage | undefined>(result => {
      if (result.status === 'fulfilled') return result.value
      logger.error(`createComposerImage failed`, {
        safeMessage: result.reason.message,
      })
    })
    .filter((img): img is ComposerImage => img != null),
)
```
It preserves the original assets order while keeping image processing concurrent (no serialization added).
# Environment
- Mac, Chrome
- Windows, Chrome, 1.108.0

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1405,9 +1405,7 @@ function ComposerFooter({
 
       if (assets.length) {
         if (type === 'image') {
-          const images: ComposerImage[] = []
-
-          await Promise.all(
+          const images: ComposerImage[] = await Promise.allSettled(
             assets.map(async image => {
               const composerImage = await createComposerImage({
                 path: image.uri,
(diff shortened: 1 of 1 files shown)


tomo-x7:
I don't have a native test environment, so I'm not sure, but this might also fix #3161.
@haileyok What do you think?

PR #9232, opened 2025-10-18
Add data-at-uri attributes for identifying posts and profiles on pages

ThisIsMissEm:
This provides extensions the ability to find the relevant area of the page for a post or profile, allowing for extensions to add additional features to those elements.

My primary use-case is around adding community notes or other data associated with posts based on their `at://` URIs. For instance, an extension could do the following:

```ts
const rkey = window.location.pathname.split('/').at(-1)
const postNode = document.querySelector(`[data-at-uri$="/${rkey}"]`) 
```

And then the extension can use both the value of the `data-at-uri` via `postNode.dataset.atUri` and identify the area on page where an element may be anchored to.

The extension would need to probably handle re-renders of the subtree and navigation events, but it's possibly better than nothing & would enable third-party extensions to do interesting layered behavior on top of the existing experience.

It is perhaps a bit more complicated for profiles, since you'd need to do the handle to did resolution in order to know the `at://` URI to look for.

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -8,7 +8,11 @@ import Animated, {
   useAnimatedRef,
 } from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
-import {type AppBskyActorDefs, type ModerationDecision} from '@atproto/api'
+import {
+  type AppBskyActorDefs,
+  AtUri,
+  type ModerationDecision,
+} from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
 import {useNavigation} from '@react-navigation/native'
@@ -63,6 +67,10 @@ let ProfileHeaderShell = ({
   const liveStatusControl = useDialogControl()
 
   const aviRef = useAnimatedRef()
+
+  const atUri = useMemo(() => {
+    return AtUri.make(profile.did).href
+  }, [profile.did])
 
   const onPressBack = useCallback(() => {
     if (navigation.canGoBack()) {
@@ -143,7 +151,10 @@ let ProfileHeaderShell = ({
   ])
 
   return (
-    <View style={t.atoms.bg} pointerEvents={isIOS ? 'auto' : 'box-none'}>
+    <View
+      style={t.atoms.bg}
+      pointerEvents={isIOS ? 'auto' : 'box-none'}
+      dataSet={{'at-uri': atUri}}>
       <View
         pointerEvents={isIOS ? 'auto' : 'box-none'}
         style={[a.relative, {height: 150}]}>
--- a/src/screens/PostThread/components/ThreadItemAnchor.tsx
+++ b/src/screens/PostThread/components/ThreadItemAnchor.tsx
@@ -306,6 +306,7 @@ const ThreadItemAnchorInner = memo(function ThreadItemAnchorInner({
       <ThreadItemAnchorParentReplyLine isRoot={isRoot} />
 
       <View
+        dataSet={{'at-uri': post.uri}}
         testID={`postThreadItem-by-${post.author.handle}`}
         style={[
           {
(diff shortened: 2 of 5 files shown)


ThisIsMissEm:
Perhaps interestingly here, if the page is loaded from the server, then there is a `link[rel=alternate]` node with the full `at://` URI, but if it's from a user navigation, then this link attribute is missing. I'm not entirely sure why that is, but does mean that you can't use that consistently.

src/screens/Profile/Header/Shell.tsx
```diff
@@ -64,6 +68,10 @@ let ProfileHeaderShell = ({
 
   const aviRef = useAnimatedRef()
 
+  const atUri = useMemo(() => {
+    return AtUri.make(profile.did).href
```
ThisIsMissEm, line 72: This could arguably be better as: 
```ts
return new AtUri(`at://${profile.did}/app.bsky.actor.profile/self`).href
```

Which would match the alternate link: https://github.com/bluesky-social/social-app/blob/028ed2f8e2edd1d35e62246e7e6df2cea09fdb07/bskyweb/templates/profile.html#L27

ThisIsMissEm:
Apparently the way to do correct link alternate on navigation would be to use a library like react-helmet: https://github.com/necolas/react-native-web/discussions/1974

mozzius:
I think this should _just work_ in React 19 now

https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags

ThisIsMissEm:
@mozzius so whilst this does work for adding the link tags, the rendering logic in the app makes things get a bit confused: multiple link elements end up stacking up:
Screenshot 2025-10-20 at 20 28 00

This seems to be due to how old views are just hidden rather than unrendered, and I'm not sure if that's a state I can access from within a component deeper in the tree.

ThisIsMissEm:
<details>
<summary>Looks like I can maybe use the isFocused state from react-navigation to get the right effect:</summary>

```diff
diff --git i/src/screens/Profile/Header/Shell.tsx w/src/screens/Profile/Header/Shell.tsx
index 20fb595eb..cdde48603 100644
--- i/src/screens/Profile/Header/Shell.tsx
+++ w/src/screens/Profile/Header/Shell.tsx
@@ -15,14 +15,14 @@ import {
 } from '@atproto/api'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
-import {useNavigation} from '@react-navigation/native'
+import {useIsFocused, useNavigation} from '@react-navigation/native'
 
 import {useActorStatus} from '#/lib/actor-status'
 import {BACK_HITSLOP} from '#/lib/constants'
 import {useHaptics} from '#/lib/haptics'
 import {type NavigationProp} from '#/lib/routes/types'
 import {logger} from '#/logger'
-import {isIOS} from '#/platform/detection'
+import {isIOS, isNative} from '#/platform/detection'
 import {type Shadow} from '#/state/cache/types'
 import {useLightboxControls} from '#/state/lightbox'
 import {useSession} from '#/state/session'
@@ -63,13 +63,14 @@ let ProfileHeaderShell = ({
   const {openLightbox} = useLightboxControls()
   const navigation = useNavigation<NavigationProp>()
   const {top: topInset} = useSafeAreaInsets()
+  const isFocused = useIsFocused()
   const playHaptic = useHaptics()
   const liveStatusControl = useDialogControl()
 
   const aviRef = useAnimatedRef()
 
   const atUri = useMemo(() => {
-    return AtUri.make(profile.did).href
+    return AtUri.make(profile.did, 'app.bsky.actor.profile', 'self').href
   }, [profile.did])
 
   const onPressBack = useCallback(() => {
@@ -155,6 +156,7 @@ let ProfileHeaderShell = ({
       style={t.atoms.bg}
       pointerEvents={isIOS ? 'auto' : 'box-none'}
       dataSet={{'at-uri': atUri}}>
+      {!isNative && isFocused ? <link rel="alternate" href={atUri} /> : null}
       <View
         pointerEvents={isIOS ? 'auto' : 'box-none'}
         style={[a.relative, {height: 150}]}>
```

Although, I do note I'd not necessarily _need_ to use `useIsFocused` as we have navigation in scope which has the `isFocused`, though it's docs say to use the hook instead: https://reactnavigation.org/docs/navigation-object#isfocused
</details>

I'm also assuming in the above that rendering `<link>` elements would not be possibly/useful in react-native on non-web platforms.

PR #9237, opened 2025-10-20
Maybe fix broken step in android testflight CI

mozzius:
Noticed that part of the Android CI referenced the "testflight" profile, when it should probably be "testflight-android". Is this correct?


estrattonbailey (approved):
Oh yeah good catch, idk if this will fix our current issue but this is correct

mozzius:
Yeah no definitely won't fix the current issue

Issue #8920, opened 2025-08-27
Show by how many characters you are currently over the character limit for DMs


LabbaDev:
### Describe the Feature

As a frequent user of BlueSky, one who loves this website and who uses it a lot to interact with my friends and the communities I'm part of, there's one thing that often frustrates me: the lack of transparency in whether or not my DM is currently over the character limit.

The current behavior is triggered at the click of the _Send_ button, at which point (on the web version) a red pop-up message will appear in the bottom-left corner of the screen, saying `Message is too long`. This can crearly be problematic, as sometimes I end up having to break my messages in many different ways and try various combinations even just to discover at which point BlueSky will finally allow the message to get through.

Therefore, I suggest adding a small text below the main DM text area which is dynamically displayed whenever the content of the text area is currently over the character limit.

I have taken the time to fork the repository and run it locally, where I implemented a solution through React Native's `<Text>`. My intervention, both in [`src/screens/Messages/components/MessageInput.tsx`](https://github.com/bluesky-social/social-app/blob/17cbf75747da83392a4347d98d4c72c537e1ad88/src/screens/Messages/components/MessageInput.tsx) and in [`src/screens/Messages/components/MessageInput.web.tsx`](https://github.com/bluesky-social/social-app/blob/17cbf75747da83392a4347d98d4c72c537e1ad88/src/screens/Messages/components/MessageInput.web.tsx), consists of adding this code at the very bottom of the outermost `<View>` element:

```tsx
{new Graphemer().countGraphemes(message) > MAX_DM_GRAPHEME_LENGTH ? (
    <Text
        style={[
            a.py_sm,
            {
                color: t.palette.negative_500,
            },
        ]}>{`Message is too long by ${new Graphemer().countGraphemes(message) - MAX_DM_GRAPHEME_LENGTH} characters`}</Text>
) : null}
```
I am aware that optimizations of this code are undoubtedly possible, mainly related to the multiple callings of `new Graphemer().countGraphemes(message)`; but I assume that those who are more accustomed with the style paradigms of the BlueSky codebase know how to make it pretty. This code here is just a blueprint that, in my case as I tested it locally, seems to work just fine.

In the attachments to this issue I am uploading a screenshow of what this new feature looks like on my computer, in the web version of BlueSky as seen from Chrome.

I hope the code, or at the very least this issue that I am opening, can help in making BlueSky even better!

### Attachments

[...]

PR #9240, opened 2025-10-20
Display character overflow count in DMs

LabbaDev:
This is my first time attempting a contribution to an open-source project, so I will try my best. Following the repository's guidelines, I first opened an issue on the topic (#8920) and let about a couple months pass in case some discussion was needed.

To sum it up, it is a very small quality-of-life improvement on something that had always bothered me (as in, actively making my use of the website harder and more frustrating), which is the lack of clarity in when a DM is over the character limit. As it stands now, the only indication that this has happened occurs upon submitting the DM, which fails with a `Message is too long` popup if you wrote too much; but without anything showing up while you type the message, it then becomes a guessing game of cutting up your DM at random points until the site is satisfied with the length of each piece, and will allow it to go through.

I have therefore thought of adding a small message that appears immediately below the DM text box, telling you by how many characters your message is currently too long whenever you are over the limit. I tried my best to stay compliant with BlueSky's current styling rules, and so I simply used a small vertical padding (`py_sm`) and the brand color for error messages (`palette.negative_500`), which I thought was just the minimal styling possible to make it functional.

temp11


LabbaDev:
Some time has passed and thus merge conflicts had arisen, so I solved them by adapting to the new structure. I used

```js
import {Text} from '#/components/Typography'
```
as seems to be preferable, and most importantly I have adapted to the new `countGraphemes(message)` in place of the old `Graphemer().countGraphemes(message)`.

I know it's been a while since this pull request was submitted, and you are certainly very busy yourselves; but I hope it can be useful, and it should be incredibly short to see, review and test too!

PR #9241, opened 2025-10-20
Add dark/tinted icon variants

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/9295
Fixes https://github.com/bluesky-social/social-app/issues/9285

Uses the new Icon Composer `.icon` format for the iOS icon

https://github.com/user-attachments/assets/0e68052e-ff00-4549-a8b0-8e3a73e78ee0


mozzius:
Oh no lol we need to wait for https://github.com/expo/expo/pull/40403

assets/app-icons/ios_icon_default.icon/icon.json
mozzius, line 1: `.icon` files are actually a `json` and some PNGs in a trenchcoat, pretty neat

app.config.js
```diff
@@ -50,6 +50,10 @@ module.exports = function (_config) {
         config: {
           usesNonExemptEncryption: false,
         },
+        icon:
+          PLATFORM === 'web' // web build doesn't like .icon files
+            ? './assets/app-icons/ios_icon_default_next.png'
+            : '/assets/app-icons/ios_icon_default.icon',
```
estrattonbailey: Web build uses iOS settings? No issue with missing `.` at the start of the path I take it 🤔

app.config.js
```diff
@@ -323,18 +327,13 @@ module.exports = function (_config) {
              * Default set
              */
             default_light: {
-              ios: './assets/app-icons/ios_icon_default_light.png',
-              android: './assets/app-icons/android_icon_default_light.png',
+              ios: './assets/app-icons/ios_icon_default_old_light.png',
```
estrattonbailey: Can a user revert to the new default icons after selecting an old one?

estrattonbailey: Little confused by the naming maybe. The default set here points to the old ones, and the new default is called default too.

mozzius: This is all for the app icons page which is inaccessible. We'll need to add a button to reset to default (which is doable). Since that'll need a lot of work to get nice anyway, I think it's best to leave it all for when we do it properly

I can rename the files to make it a little less confusing though

app.config.js
mozzius: Yeah for some reason the web build wants to run it through `@expo/image-utils` and fails :/

PR #9242, opened 2025-10-20
Get iOS debug files into Sentry

estrattonbailey:
Our crash reports in Sentry for iOS builds are currently missing debug symbols for the main `Bluesky` package.

It's clear when viewing the Debug Files that _are_ uploaded to Sentry, that for the main `Bluesky` package, all we're getting before this PR was the `Mach-O executable`, which appears to the ~35MB source code of the app. _After_ this PR, we're now getting a ~145MB `Mach-O debug companion` which replaces the executable. As [stated in the docs](https://docs.sentry.io/platforms/apple/guides/macos/data-management/debug-files/file-formats/), this file has a `debug` tag, which makes me think it's what we were looking for.

### Apple Privacy Manifest
The Sentry docs state that [these changes](https://docs.sentry.io/platforms/react-native/data-management/apple-privacy-manifest/#create-privacy-manifest-in-expo) are required for the plugin to work. We already had some of these in place previously, but some where missing.

This alone does not seem to fix the fact that Sentry's Expo plugin on its own does not upload the main bundle's debug file.

### Build updates
Sentry definitely does its [own work internally](https://github.com/getsentry/sentry-react-native/blob/main/packages/core/scripts/sentry-xcode-debug-files.sh) to get all the debug files. This happens during the `Upload Debug Symbols to Sentry`, which is added by the Sentry Expo plugin to XCode's Build Phases. But it does not appear to find the `.dSYM` files for the main `Bluesky` bundle on its own.

To output the `.dSYM` files, our `eas build` command has been updated from `--output build.ipa` (which outputs just the app bundle) to `--output build.tar.gz`. This produces a gzipped tarball that includes the `.ipa` and other artifacts. FYI, the default if you don't supply a `--output` flag is to tarball everything into a `build-<some_number_string>.tar.gz` file.

This tarball alone does not include the `.dSYM` files. To include those, we need to add `buildArtifactPaths` to the `eas.json` config. NOW after running `eas build`, the tarball contains a `Bluesky.app.dSYM.zip` file.

### Manual upload
So now our build outputs a tarball, which we uncompress. We can then `eas submit` the included `.ipa`, and separately upload the `.dSYM.zip` to Sentry manually.

### Misc notes
- If you include `buildArtifactPaths`, and keep the `--output build.ipa`, the resulting `.ipa` file size is equivalent to the `.ipa` + `.dSYM` file sizes. You can submit this directly to TestFlight and it works fine. The resulting app bundle size, as stated in App Store Connect, remains just the size of the `.ipa`. I'm not sure what happens to the rest of the file, which I assumed contains the `.dSYM` files.
  - If you inspect the `Build Metadata` of the app build in ASC, it always states `Includes Symbols: Yes`, regardless of if the `build.ipa` includes the debug files. It looks like this for older builds too. But clearly XCode isn't getting symbols either atm. Fixing that isn't nearly as important as fixing Sentry atm.

### DIFF
--- a/.github/workflows/build-submit-ios.yml
+++ b/.github/workflows/build-submit-ios.yml
@@ -115,7 +115,7 @@ jobs:
       - name: 🚀 Deploy
         run: eas submit -p ios --non-interactive --path ios-build/ios/build/Bluesky.ipa
 
-      - name: Upload dSYM to Sentry
+      - name: 🪲 Upload dSYM to Sentry
         run: >
           SENTRY_ORG=blueskyweb
           SENTRY_PROJECT=app


.github/workflows/build-submit-ios.yml
```diff
@@ -98,10 +98,29 @@ jobs:
...
+          else
+            echo "Archive file not found!"
+            exit 1
+          fi
 
       - name: 🚀 Deploy
-        run: eas submit -p ios --non-interactive --path build.ipa
+        run: eas submit -p ios --non-interactive --path ios-build/ios/build/Bluesky.ipa
+
+      - name: Upload dSYM to Sentry
```
mozzius: Need an emoji‼️

mozzius:
```suggestion
      - name: 🪲 Upload dSYM to Sentry
```

.github/workflows/build-submit-ios.yml
```diff
@@ -98,10 +98,29 @@ jobs:
           yarn use-build-number-with-bump
           eas build -p ios
           --profile ${{ inputs.profile || 'testflight' }}
-          --local --output build.ipa --non-interactive
+          --local --output build.tar.gz --non-interactive
+
+      - name: Extract build artifact
```
mozzius:
```suggestion
      - name: 📂 Extract build artifact
```

PR #9259, opened 2025-10-22
Remove error log when updating email

estrattonbailey:
Since this is an endpoint the user interacts with and it provides validation, and the handling in here is quite simple, the majority of these logs are gonna be validation issues like `Email already taken` or `Rate limited`. We don't have a good way of filtering known vs unknown errors here. We already provide feedback to the user, I think we can remove this to free up some capacity.

https://blueskyweb.sentry.io/issues/6732146906/events/ad95f3a98659454587e83db7562aefe8/?project=4508807082278912&query=is:unresolved%20update%20email%20failed


estrattonbailey:
@mozzius ya know what, this is a small fry anyway, so just gonna close. We really need to start to use lexicon error types more so we can differentiate between known/unknown errors. The 10 I clicked through were all "known" errors and could be ignored, but we can't keep matching error strings in the client.

mozzius:
Definitely agree

PR #9262, opened 2025-10-22
Disable Sentry sampling for now

estrattonbailey:


efstathiosntonas:
hey folks, just for academic knowledge, why sampling is disabled? Thanks

mozzius:
We're tracking down a bug. We're going to reenable it once we've caught it

PR #9266, opened 2025-10-23
Patch out `PerformanceMonitor` component

mozzius:
This Reanimated debug component that we don't even seem to use is causing the crashes with copying text in DMs.

This only happens when using the `--no-dev` flag, hence why we missed it in development

Screenshot 2025-10-23 at 11 57 05


mozzius:
@github-actions ota

patches/react-native-reanimated+3.19.1.patch
```diff
@@ -0,0 +1,420 @@
...
+-};
+-
+-/**
+- * A component that lets you measure fps values on JS and UI threads on both the
+- * Paper and Fabric architectures.
+- *
+- * @param smoothingFrames - Determines amount of saved frames which will be used
+- *   for fps value smoothing.
+- */
+-export function PerformanceMonitor({
```
estrattonbailey, line 351: Might as well leave this export and return `null`?

mozzius: Done!

PR #9272, opened 2025-10-24
use admonition instead of toast for errors when appealing labels

BinaryFiddler:
https://github.com/user-attachments/assets/5289f41c-8e07-4b00-a1d9-7960395bddcf

### DIFF
--- a/src/components/moderation/LabelsOnMeDialog.tsx
+++ b/src/components/moderation/LabelsOnMeDialog.tsx
@@ -230,7 +230,7 @@ function AppealForm({
   const sourceName = labeler
     ? sanitizeHandle(labeler.creator.handle, '@')
     : label.src
-  const [error, setError] = useState<React.ReactNode>(null)
+  const [error, setError] = useState<string | null>(null)
 
   const {mutate, isPending} = useMutation({
     mutationFn: async () => {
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Oooh, this looks nice, curious how it'd look on web though?

src/components/moderation/LabelsOnMeDialog.tsx
```diff
@@ -285,6 +298,11 @@ function AppealForm({
           </Trans>
         </Text>
       </View>
+      {error && (
+        <Admonition type="error" style={[a.mt_sm]}>
+          <Trans>{error}</Trans>
```
mozzius: `<Trans>` doesn't do anything here, can remove

src/components/moderation/LabelsOnMeDialog.tsx
```diff
@@ -228,6 +230,7 @@ function AppealForm({
   const sourceName = labeler
     ? sanitizeHandle(labeler.creator.handle, '@')
     : label.src
+  const [error, setError] = useState<React.ReactNode>(null)
```
mozzius: `_(msg``)` returns a string

```suggestion
  const [error, setError] = useState<string | null>(null)
```

src/components/moderation/LabelsOnMeDialog.tsx
```diff
@@ -252,8 +255,18 @@ function AppealForm({
       )
     },
     onError: err => {
+      if (err instanceof XRPCError) {
+        if (err.error === 'AlreadyAppealed') {
```
mozzius: if `err instanceof XRPCError` and `err.error !== 'AlreadyAppealed'`, it won't display anything - need to handle this case

BinaryFiddler:
> Oooh, this looks nice, curious how it'd look on web though?

Like this:

Screenshot 2025-10-27 at 1 10 26 PM

mozzius:
Lovely

PR #9273, opened 2025-10-26
Add android ripple effect to images in feed

### LINKED ISSUE #3821: Add an ink-ripple-on-touch effect to images

On mobile, pressing on an image in a post to expand it feels very awkward, there's no feedback and since there's usually a big delay until the image opens I'm usually not sure if I actually pressed it or not.

I think a Material style Ink ripple or something similar would make a big difference, that's how it is on twitter and it feels much better

mozzius:
Fixes #3821 

A bit of style :)

https://github.com/user-attachments/assets/b025c9f9-e9d2-4099-959b-032d23a5a7af


ledlamp:
This is terrible!!! It keeps doing the effect when touching images while scrolling! So annoying!

Issue #7452, opened 2025-01-14
Rotate images cleanly on mobile


prohr:
### Steps to Reproduce

To feel fully immersive, rotating an image should smoothly toggle only the size + orientation (in a single step).  Unlike the simple fix which was landed to close #624, please don't also reset other things about the view state (like whether the alt text + controls are displayed)

1. Tap on a wide image to display fullscreen (portrait)
2. Tap once to get rid of alt text + controls
3. Rotate device
4. Rotate device back

Expected behavior:

3a. Immediately see fullscreen landscape image
4a. Immediately see fullscreen portrait image

Actual behavior:

3a. Brief peek at missized landscape image
3b. Image resizes correctly, but underneath alt text + controls 
3c. Tap to get rid of the overlay

4a, 4b, 4c.  Similar multistep process for portrait

### Attachments

https://github.com/user-attachments/assets/3d69cdad-365c-48c2-91fc-4451c5e17c8d

Rotating the image here:

https://bsky.app/profile/carlbergstrom.com/post/3lfh2vt36ys2e

### What platform(s) does this occur on?

Android

### Device Info

Pixel 6a

### What version of the app are you using?

Build version: 1.96.3.507; Bundle info: 9f075b1 (prod); Bundle date: 24123121; Platform: android; Platform version: 34

### Additional Information

The two-step resizing effect seems to be more noticeable for images which don't have alt text+/or which have a less square aspect ratio

It's also worth considering how to handle rotation when the image is already zoomed in.  Simplest implementation would rezoom to fit the new viewport (fullscreen).  More immersive implementations might try to preserve the zoom level +/or focal point (but have special cases to consider when that doesn't "fit")

prohr:
Update:  Figured out at least one root cause for the flicker. 

1. Choose a post with multiple images
2. Tap twice to make first image full screen 
3. Scroll to say the third one 
4. Rotate your device

Expected behavior:

5.  Uncaptioned 3rd image rotates + resizes smoothly 

Actual behavior:

5a. Uncaptioned 3rd image rotates without resizing 
5b. Captioned 1st image displays at the correct size 

Whoops!

prohr:
Video of the second repro case[](url)

https://github.com/user-attachments/assets/e8825655-5f85-4330-8799-36dbe92c88b4

PR #9279, opened 2025-10-27
Keep image index when rotating lightbox

mozzius:
Lifts state out of the inner part of the lightbox that gets recreated via key when rotating


prohr:
When testing, can you check whether this will also resolve #7452?

( Specifically, toggling the caption +/or zoom states )

PR #9283, opened 2025-10-27
Add display of pronouns when composing replies

ThisIsMissEm:
When replying to a post where the account that posted it has a pronouns field, display their pronouns next to their display name / handle.

Screenshot 2025-10-27 at 19 59 53

This is the most minimal step towards integrating the [new pronouns field](https://github.com/bluesky-social/atproto/pull/4232) into the social-app, it doesn't allow adding a pronoun yet, just displaying them if present.

There is a case where this would be hard to support, which is when you click the `ComposeBtn` from a profile page. Ideally there we'd show something above the composer saying "Mentioning [display name / handle] (pronouns)" and then hide that if they backspaced out the mention text. But that would be much more difficult to implement, since the `ComposeBtn` doesn't actually pass through a mention to the actor, [just their handle](https://github.com/bluesky-social/social-app/blob/bd17f0481054f23aa2e0bb5aecf88566fdfe9d6a/src/view/shell/desktop/LeftNav.tsx#L563) as a string, if their handle is valid.

Someone's also mentioned to me that it'd be useful when composing quote posts, but I'm not sure there's a good way to add it without redesigning the quote post in the composer to split display name and handle to separate lines?
Existing quote post composer

### DIFF
--- a/src/view/com/composer/ComposerReplyTo.tsx
+++ b/src/view/com/composer/ComposerReplyTo.tsx
@@ -108,6 +108,18 @@ export function ComposerReplyTo({replyTo}: {replyTo: ComposerOptsPostRef}) {
               replyTo.author.displayName ||
                 sanitizeHandle(replyTo.author.handle),
             )}
+            {typeof replyTo.author.pronouns === 'string' && (
+              <Text
+                style={[
+                  a.pl_xs,
+                  a.font_normal,
+                  a.flex_shrink,
+                  t.atoms.text_contrast_medium,
+                ]}
+                numberOfLines={1}>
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Interestingly, the work that would be needed to enable displaying pronouns in the composer when mentioning from profile would be the same, or at least similar to, that which would be needed for getting that initial mention to be rendered as an `@mention` correctly in the compose, I think?

Currently it displays that first mention as text rather than a highlighted mention, though I'm not 100% sure why. (I did try adding a mention facet to the initial richtext but that still didn't result in correct rendering)

Screenshot 2025-10-27 at 21 13 50

PR #9287, opened 2025-10-27
Adds client event for empty feed error

abenzer:


src/view/com/posts/CustomFeedEmptyState.tsx
```diff
@@ -11,11 +11,34 @@ import {usePalette} from '#/lib/hooks/usePalette'
...
+  useEffect(() => {
+    // Log the empty feed error event
+    if (feedFeedback.feedSourceInfo && currentAccount?.did) {
+      const feedDescriptor = feedFeedback.feedDescriptor
+      const [feedType, feedId] = feedDescriptor?.split('|') || [
+        'unknown',
+        'unknown',
+      ]
+
+      logger.metric('feed:emptyError', {
```
estrattonbailey: This may not be an error in the case of other feeds or feeds users just created that don't have posts yet, so could skew the data. If we only care about Discover right now, we could check if the feed URI is `DISCOVER_FEED_URI` from `lib/constants` and only then log it.

And if it's that specific, no harm in changing the event name to like `discover:emptyError` or something.

Issue #8457, opened 2025-06-07
Images Not Saving


RedFoxNightfox:
### Steps to Reproduce

1. Go to image 
2. Click on save
3. See "image saved"
4. Goto file manager and look at recent files
5. No image downloaded

### What platform(s) does this occur on?

Android

### Device Info

Samsung A14

### What version of the app are you using?

Build version: 1.103.0.653; Bundle info: 6fbd6605d (prod); Bundle date: 25060521; Platform: android; Platform version: 34; Anonymous ID: cde6b8f7-66e8-4b4b-a4bc-afdf3f9c580e

### Additional Information

I recently moved the bluesky folder, along with a bunch of other program image save folders, drom internal to external to free up some space, and that's when I noticed this bug. Recreating the bluesky folder did not fix tne issue.

I have tried clearing cash, clearing data, uninstalling and reinstalling, and the bug persists.

RedFoxNightfox:
After uninstalling and reinstalling a few times again and it still not working, I set permissions to limited instead of full, it started working again. No idea why.

Issue #1360, opened 2023-09-02
Save images to Bluesky folder


thktkblt:
For now, the images downloaded from Bluesky are saved to general DCIM folder (I use Android), so it would get mixed with pics from other sources and feel cluttered.

pfrazee:
I originally did this, but it required more permissions than folks were comfortable with :(

fuelflo:
I'd also appreciate this. What permissions does that need, normally just access to storage and or photos/videos and you could only ask for permission when saving, right?

pfrazee:
IIRC it forced me to ask for full photo library access (eat me). It's way out of scope for the team's current workstream but as a PR I'd accept some settings for configuring where your images are saved and dealing with the permissions.

fuelflo:
Yeah, granular photo/video library access is only available from Android 14, which will probably be released in two days....

https://developer.android.com/about/versions/14/changes/partial-photo-video-access

But as Android manufacturer's upgrade slowly to newer versions, most of the users will be stuck for quite some time to have to grant full access.

SaphireLattice:
Sigh. This one is getting extremely annoying. `/DCIM` folder gets very enthusiastically synced by Google Photos. Which uh, is extremely not what I want, and there's no way to exclude that without also excluding actual camera photos, I think?

So I end up with a mess of a gallery thanks to this.

kairusds:
i wish they could just have the photos saved in Download/BlueSky or Picture/BlueSky

dbrian57:
Came here to add my +1 for this change. The current setting really gunks up my camera roll. I like to keep my real life and weirdo online life separate, thank you very much.

marksteward:
No extra permission should be needed to download files to the Downloads directory on Android.

Nonplayable:
Begging y'all to let us save images downloaded from Bluesky to literally any folder other than DCIM. I do not need my camera folder or Google backups cluttered with dumb memes I want to send to a non Bluesky using friend

RedFoxNightfox:
I too find it very annoying that I cannot change the download folder to not be DCIM. Every other andriod app I have ever used either allows me to change the download folder, or at least defaults to .\pictures\{app} or .\downloads\{app} folders.

holistic-developer:
Since DCIM = digital camera images, this folder is not intended to be used for this purpose.

I understand the issue of asking too many permissions upfront, so ideally the required permissions could be requested right when the user tries to save an image.
[...]

PR #9293, opened 2025-10-28
Android - add fallbacks when saving image to gallery

mozzius:
Fixes https://github.com/bluesky-social/social-app/issues/8457

Since #1360, we've been saving images to a "Bluesky" folder on Android. However, this has resulted in an issue where a small % of users cannot save images.

Turns out that you're only allowed to save images to directories if they reside in the `Pictures` or `DCIM` folders. If a user creates their own folder, say in `Downloads` or on an SD card, we will successfully fetch the album but be unable to save to it - `expo-media-library` queries albums by name, so turns out this can happen. Even worse, some bad defensive programming on my part meant that the success toast got triggered even if it failed :/ whoops

## Issue reproduction

On Android:

1. Save an initial image. This creates `Pictures` > `Bluesky` > `[name].jpg`
2. Create a folder in `Downloads`, move the `Bluesky` folder to `Downloads` > `Example` > `Bluesky`
3. Trying to save another image. It finds the album but it can no longer write to it

## The fix

The fix is to add a couple of fallback paths

1. Add album migration step. Not the immediate problem, but can't hurt - have wrapped in try/catch
2. If the `createAssetAsync` call fails, create a new Bluesky folder instead
3. If creating a new folder fails, save the image to DCIM

This ensures the image will always save, even if it has to fall back to DCIM

Also rethrows the error in the outer try/catch so it correctly shows a failure toast

# Test plan

Follow the issue reproduction steps. confirm that either a new Bluesky folder is created _or_ the image is saved to DCIM

### DIFF
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -120,7 +120,7 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
             await MediaLibrary.migrateAlbumIfNeededAsync(album)
           }
         } catch (err) {
-          logger.debug('Attempted and failed to migrate album', {
+          logger.info('Attempted and failed to migrate album', {
             safeMessage: err,
           })
         }
@@ -129,7 +129,7 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
           // if album exists, put the image straight in there
           await MediaLibrary.createAssetAsync(imagePath, album)
         } catch (err) {
-          logger.debug('Failed to create asset', {safeMessage: err})
+          logger.info('Failed to create asset', {safeMessage: err})
           // however, it's possible that we don't have write permission to the album
           // try making a new one!
           try {
@@ -140,7 +140,7 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
               imagePath,
             )
           } catch (err2) {
-            logger.debug('Failed to create asset in a fresh album', {
+            logger.info('Failed to create asset in a fresh album', {
               safeMessage: err2,
             })
             // ... and if all else fails, just put it in DCIM


src/lib/media/manip.ts
```diff
@@ -114,8 +114,39 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
...
       if (album) {
-        // if album exists, put the image straight in there
-        await MediaLibrary.createAssetAsync(imagePath, album)
+        // try and migrate if needed
+        try {
+          if (await MediaLibrary.albumNeedsMigrationAsync(album)) {
+            await MediaLibrary.migrateAlbumIfNeededAsync(album)
+          }
+        } catch (err) {
+          logger.debug('Attempted and failed to migrate album', {
```
estrattonbailey: Do we not care to get these into Sentry? `info` would add a breadcrumb at least if that's helpful

src/lib/media/manip.ts
```diff
@@ -114,8 +114,39 @@ export async function saveImageToMediaLibrary({uri}: {uri: string}) {
...
+            safeMessage: err,
+          })
+        }
+
+        try {
+          // if album exists, put the image straight in there
+          await MediaLibrary.createAssetAsync(imagePath, album)
+        } catch (err) {
```
estrattonbailey, line 131: Do we need to check for a specific error here?

mozzius: I haven't come across any other error that this might throw, so no I don't think so. No harm trying the fallback in all cases

src/lib/media/manip.ts
mozzius: Honestly no tbh. They're mostly expected errors, no point wasting sentry usage on them - the important one that we re-throw gets logged

mozzius: Actually, no, I'll make them `info` - will make the breadcrumb more helpful. Thanks for the tip!

estrattonbailey (approved):

PR #9297, opened 2025-10-29
Avoid asking for read access to photo library when saving images on iOS

mozzius:
Currently we ask for the full read/write permission before saving an image on iOS. Turns out this is unnecessary, as there is a dedicated `MediaLibrary.saveToLibraryAsync(uri)` function that will simply ask for "Add only" permission.

I opted to platform split the file since the iOS version is so much simpler

## Before

https://github.com/user-attachments/assets/d3fdf846-53d7-4be8-9c98-e814c53d6d9d

## After

https://github.com/user-attachments/assets/0bd8ccc5-1b47-4586-a5cd-4fdef311fa26

### DIFF
--- a/src/lib/media/save-image.ios.ts
+++ b/src/lib/media/save-image.ios.ts
@@ -1,5 +1,6 @@
 import {useCallback} from 'react'
-import {t} from '@lingui/macro'
+import {msg} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
 
 import {isNative} from '#/platform/detection'
 import * as Toast from '#/components/Toast'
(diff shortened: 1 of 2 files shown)


src/lib/media/save-image.ios.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function useSaveImageToMediaLibrary() {
+  return useCallback(async (uri: string) => {
+    if (!isNative) {
+      throw new Error('useSaveImageToMediaLibrary is native only')
+    }
+
+    try {
+      await saveImageToMediaLibrary({uri})
+      Toast.show(t`Image saved`)
```
estrattonbailey: Might as well replace `t` with `_` while were in here, in both files

estrattonbailey (approved):
Nice nice

PR #9305, opened 2025-10-29
fix: do not render external embed if the domain is not valid

caidanw:
Summary
---

Ensure that for errors returned from the link metadata service, we should not show a embed card.

Part of the issue here was that the errors were not properly being propagated through the react query hook. I've updated the code to properly throw errors so that it works as expected.

Created some custom error classes so that we can better handle the various cases of failure. I've started simply with the existing errors we were throwing. ~~This allows us to handle the different types of failures, and to ensure that we can show an error embed card when we have an unexpected error, otherwise when we encounter specific errors from the service we can remove the card entirely.~~ Updated to not show any error cards and to remove the embed if we encounter any error.

### DIFF
--- a/src/lib/link-meta/errors.ts
+++ b/src/lib/link-meta/errors.ts
@@ -1,0 +1,35 @@
+/**
+ * Custom error classes for specific link metadata errors we need to handle differently
+ */
+
+/**
+ * Thrown when a URL contains a host IP address, which is not allowed
+ * This is a security restriction to prevent access to internal/private IPs
+ */
+export class InvalidUrlHostIPError extends Error {
+  constructor(
+    message: string = 'Invalid URL: host IP addresses are not allowed',
+  ) {
+    super(message)
+    this.name = 'InvalidUrlHostIPError'
+  }
+}
+
+/**
+ * Thrown when a URL is malformed or invalid (but not specifically due to host IP)
+ */
+export class InvalidUrlError extends Error {
+  constructor(message: string = 'Invalid URL') {
+    super(message)
+    this.name = 'InvalidUrlError'
+  }
+}
+
+/**
+ * Type guard to check if an error is specifically an InvalidUrlHostIPError
+ */
+export function isInvalidUrlHostIPError(
+  error: unknown,
+): error is InvalidUrlHostIPError {
+  return error instanceof InvalidUrlHostIPError
+}
--- a/src/lib/link-meta/errors.ts
+++ b/src/lib/link-meta/errors.ts
@@ -24,12 +24,3 @@ export class InvalidUrlError extends Error {
     this.name = 'InvalidUrlError'
   }
 }
-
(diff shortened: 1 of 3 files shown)


mozzius:
Hmm, I think I disagree with this approach. Wouldn't it make more sense to not include it in the post in the first place if we get nothing back from the api

mary-ext:
I'd also note that this might be somewhat disruptive unless it checks for the presence of said link on the richtext

mozzius:
Yeah, this is why I'd prefer the change to be in the composer I think. That way deleting the link isn't misleading

Also, worth making sure it's not one of the whitelisted special embed types

jimmylee:
I think it is nice that users receive feedback when they try to embed a link and discover that the link has missing meta tag data.

caidanw:
@mozzius
> Wouldn't it make more sense to not include it in the post in the first place if we get nothing back from the api

I could be wrong, but this approach is essentially not including the embed in the post. At least at the closest point to where we process the uri and get the link metadata.

@mary-ext
> I'd also note that this might be somewhat disruptive unless it checks for the presence of said link on the richtext

Can you elaborate on how this would be disruptive for richtext? I looked through the `ComposerPost` component code, and it doesn't seem like my change would affect rich text, it only prevents rendering of the embed if there isn't enough data.


Other than going up the tree to the `ExternalEmbedLink` component, we can't really place this code anywhere else without doing an additional call to the hook `useResolveLinkQuery(uri)` or a large refactor.

We could do an additional call to that hook in the `ComposerEmbeds` component, but I'm not sure if it's a good solution either. 

---

When I try to move the check further up the component tree I get an infinite loader for the embed. This is because the logic expects `linkComponent` to be a valid component otherwise it renders the loader. I can in theory return an empty `div` here but that feels wrong. We could a check for the `linkComponent` being `null` and do an early exit from the component, but it doesn't really do anything different than my current solution. 

image

---

If either of you have suggestions for solutions, I'd love to hear them but I'm struggling to understand how I'd approach this without a much larger refactor.

mary-ext:
it's possible to omit the links in the post text after attaching it as an embed, leaving the link embed as the sole external link, I think this trick has been widely shared by posters to squeeze a little bit more text into the 300 character limit.


https://github.com/user-attachments/assets/2258f0b3-52b5-4e3a-aef7-921213e8f481

caidanw:
> it's possible to omit the links in the post text after attaching it as an embed, leaving the link embed as the sole external link, I think this trick has been widely shared by posters to squeeze a little bit more text into the 300 character limit.
> 
>  Record_2025-10-31-06-38-59_cb2df8437d99d85560b8f74042fc78eb.mp4

Oh interesting! I didn't know about this. Well I only opened this PR because it was assigned to me as a ticket internally. I'll talk with the team and let them know about these potential impacts.

PR #9307, opened 2025-10-30
Update web dim `<html>` background color

mozzius:
These weren't updated with the ALF colour things

I can do the main jarring one, but @estrattonbailey can you help update the others? I don't know what these values do


mozzius:
Thanks @estrattonbailey

PR #9308, opened 2025-10-30
Update android splash screen with new icon

mozzius:
https://github.com/user-attachments/assets/59d311d8-9abc-4dd3-8c8a-39022cedf393


orual:
thanks, I hate it. the old one looked *really* nice, sorry sam.

mozzius:
It's ok, it's ok to be wrong

estrattonbailey:
I think the drop shadow here is a little too dark

PR #9311, opened 2025-10-31
Use 0.5px media border on high DPI screens on web

mozzius:
<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

### DIFF
--- a/src/components/MediaInsetBorder.tsx
+++ b/src/components/MediaInsetBorder.tsx
@@ -1,6 +1,8 @@
+import {StyleSheet} from 'react-native'
 import type React from 'react'
 
-import {atoms as a, useTheme, type ViewStyleProp} from '#/alf'
+import {isHighDPI} from '#/lib/browser'
+import {atoms as a, platform, useTheme, type ViewStyleProp} from '#/alf'
 import {Fill} from '#/components/Fill'
 
 /**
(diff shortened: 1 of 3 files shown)


src/components/MediaInsetBorder.tsx
```diff
@@ -25,7 +27,15 @@ export function MediaInsetBorder({
...
       style={[
         a.rounded_md,
-        a.border,
+        {
+          borderWidth: platform({
+            native: StyleSheet.hairlineWidth,
+            // while we generally use hairlineWidth (aka 1px),
+            // we make an exception here for high DPI screens
+            // as the 1px border is very noticeable -sfn
+            web: isHighDPI ? 0.5 : StyleSheet.hairlineWidth,
```
rafaeleyng, line 36: Maybe want to do `StyleSheet.hairlineWidth / 2` to keep both things tied?

mozzius: 🤔 - could do. However, I think CSS may add hairline borders to the spec at some point, so it might not be 1px forever

estrattonbailey (approved):

Issue #7311, opened 2024-12-31
CORS headers are missing from oEmbed endpoint


eblu3:
### Steps to Reproduce

1. Fetch any data from https://embed.bsky.app/oembed.
2. Client returns a CORS error, as the `Access-Control-Allow-Origin` header is missing.

OR

1. Attempt to fetch a Bluesky post page, in order to find either the `Link` HTTP header, or the `link` element with a `type` of `application/json+oembed` in the `head` section of the document.
2. Client returns a CORS error, as the `Access-Control-Allow-Origin` header is missing. The web server does not return a `Link` header in the response either.

### What platform(s) does this occur on?

Web (Desktop), Web (Mobile)

### What version of the app are you using?

The latest version currently hosted on https://bsky.app (I do not use Bluesky, so the settings menu is not visible to me)

QuinnyPig:
Seeing this as well. CORS is the bane of all things...

tiagoamaro:
I'm facing the same problem while trying to use the `oEmbed` endpoint, documented in the ["oEmbed and Post Embed Widget" page](https://docs.bsky.app/docs/advanced-guides/oembed#oembed-endpoint).

Fetching its resource through the client side is impossible without proxying and adding the `Access-Control-Allow-Origin` header ourselves.

I don't want to use a third party library or try to mimic Bluesky design since the official oEmbed endpoint offers its HTML format (with the embed JS included).

blue82dc:
I am having this exact same problem. It is preventing me from offering users of my web forum the option of embedding Bluesky posts in their forum posts. Please fix it.

jcink:
+1 to the issue regarding forum embeds, we offer forum software as a service, and an example request client-side like:

```
      $.ajax({
           url: 'https://embed.bsky.app/oembed?maxwidth=300&maxheight=400&url=' + blueskyPost,
           dataType: 'json',
           success: function(data) {
               $this.html(data.html);
           }
```

Which could be useful for implementing a bluesky bbcode can't succeed because of CORS...

izalea:
Hi, I am working with a news website, and our editors would like to embed bluesky posts in their articles, just like they do all other social media. This CORS issue is a real constraint... I could try to fix it if anyone could provide me a hint of where you have the code that handles the embeds, cheers.

PR #9312, opened 2025-10-31
[embedr] check for AtIdentifier in sanity check, allowing `handle` in addition to `did`

DarthTealc:
This PR updates the embed iframe url `https://embed.bsky.app/embed/:atid/app.bsky.feed.post/:post` sanity check to accept either `handle` or `did`, instead of only accepting `did`.

**Current behaviour:**
`https://embed.bsky.app/embed/did:plc:p2cp5gopk7mgjegy6wadk3ep/app.bsky.feed.post/3lxciw4rf4c2p` = functional embed iframe
`https://embed.bsky.app/embed/samuel.fm/app.bsky.feed.post/3lxciw4rf4c2p` = "Invalid DID: DID syntax didn't validate via regex"

**PR behaviour:**
`https://embed.bsky.app/embed/did:plc:p2cp5gopk7mgjegy6wadk3ep/app.bsky.feed.post/3lxciw4rf4c2p` = functional embed iframe
`https://embed.bsky.app/embed/samuel.fm/app.bsky.feed.post/3lxciw4rf4c2p` = functional embed iframe

### Details

Bluesky's embed iframe page is actually already fully functional with handles. You can test this by building the current version of [bskyembed](https://github.com/bluesky-social/social-app/tree/main/bskyembed) locally and you'll see that `http://localhost:5173/embed/samuel.fm/app.bsky.feed.post/3lxciw4rf4c2p` loads without issue.
The reason it doesn't currently work on `https://embed.bsky.app/embed/*` is because the sanity check on generating the page explicitly checks for `syntax.ParseDID`. This PR changes that sanity check to use `syntax.ParseAtIdentifier` instead. That single change fixes it, but I've made additional changes to rename the variables to reflect how they now represent an At ID instead of just a did.

### DIFF
--- a/bskyembed/test.html
+++ b/bskyembed/test.html
@@ -1050,5 +1050,27 @@
         </div>
       </div>
     </section>
+	
+    <!-- AT ID types - with data-bluesky-uri only -->
+    <section>
+      <h1>AT ID types</h1>
+      <div class="grid">
+        <div class="item">
+          <blockquote class="bluesky-embed" data-bluesky-uri="at://did:plc:p2cp5gopk7mgjegy6wadk3ep/app.bsky.feed.post/3lxciw4rf4c2p">
+            <p lang="en">static html to be replaced with generated iframe</p>
+          </blockquote>
+
+          <div>did:plc</div>
+        </div>
+
+        <div class="item">
+          <blockquote class="bluesky-embed" data-bluesky-uri="at://samuel.fm/app.bsky.feed.post/3lxciw4rf4c2p">
+            <p lang="en">static html to be replaced with generated iframe</p>
+          </blockquote>
+
+          <div>handle</div>
+        </div>
+      </div>
+    </section>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


DarthTealc:
Issue #7311 relates to the oEmbed endpoint, which accepts requests for either `did:plc` or `handle` identifiers, but can't be used for clientside embedding due to CORS restrictions. 
The iframe endpoint has no issues with cross-domain embedding, but isn't a suitable alternative as it only accepts `did:plc`. 

This PR gives the iframe endpoint the same acceptance of both `did:plc` and `handle` as oEmbed does, and could be a solution for website developers in the event CORS restrictions are unable to be loosened on the oEmbed endpoint.

bnewbold (commented):
I approved and merged the linked CORS PR, which I think resolves your need. Note that it hasn't necessarily been deployed yet.

For this PR, I think putting a handle in the embed is a bit of an anti-pattern. Handles are not permanent account identifiers, and can change over time, potentially breaking embeds. I think we should encourage resolving to a DID as much as possible.

PR #9318, opened 2025-11-01
Fix DM message input placement on web

jcsalterego:
| Before | After |
| -- | -- |
| Screenshot 2025-11-01 at 8 37 14 AM | Screenshot 2025-11-01 at 8 37 17 AM |

On web, `animatedStickyViewStyle` has been leaving space for the footer nav looking at `footerHeight.get()`. In my browsers, this looks like `58px`, so the message input box gets elevated by `58px`. The problem is, the footer nav is hidden in this view, so the message input box looks elevated and actually covers the last message.

I don't have a working setup for testing iOS so I've scoped this to `isWeb`. It's at least a starting point.

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -301,7 +301,11 @@ export function MessagesList({
 
   const animatedStickyViewStyle = useAnimatedStyle(() => ({
     transform: [
-      {translateY: -Math.max(keyboardHeight.get(), footerHeight.get())},
+      {
+        translateY: isWeb
+          ? 0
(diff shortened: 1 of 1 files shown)


mozzius:
This is caused by changing your browser window width, as when the bottom bar gets hidden it doesn't update footerHeight to be 0.

I will fix this properly in #8543

jcsalterego:
Thanks @mozzius! You're a saint and a scholar

jcsalterego:
Reopening this to see if we can get this merged ahead of the long-term fix https://github.com/bluesky-social/social-app/pull/8543

PR #9323, opened 2025-11-03
Fix DMs on mobile web by using `dvh`

### LINKED ISSUE #9331: Chrome on Android, chat message cut off by text input
### Steps to Reproduce

1. log into Bsky on Chrome on Android, using your phone vertically.
2. Head to a private chat message with a thread long enough to require scrolling.

Bottom of message is cut off by text input (see screenshot). There is no way to get it un-overlapped, you can't scroll the messages anymore. The text input always obscures part of the last message.

Likely you need to enable the on screen buttons for this bug to occur, which I know is not a common setting these days. Overall this bug is very minor (why would you use Chrome and not the dedicated app?), but figured I'd report it anyway.

### Attachments

Image

Web (Mobile)

### Device Info

Google Pixel 6a

bksy.app website on mobile

### Additional Information

Android 16; Pixel 6a Builf/BP3A.250905.014

Chrome 141.0.7390.122

mozzius:
# Stacked on #8543 

Changes `a.util_screen_outer` to be `100dvh` instead of `100vh`

Using a dynamic viewport height means that the DM screen now fits within the dynamic viewport area, and the input no longer gets stuck below the browser UI, and there's no double scrollbars.

This is a separate PR from the base because the implications may be a little far-reaching, as this globally changes every screen. However, just using the feed etc seems to work just fine. On a desktop browser, `dvh === vh`, and just using the app normally on my phone doesn't seem to have broken anything.

https://github.com/user-attachments/assets/b5fc68ea-7b9e-4075-af2a-0ec39cfade63

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -227,40 +227,39 @@ function App() {
    * that is set up in the InnerApp component above.
    */
   return (
-    <Geo.Provider>
-      <A11yProvider>
-        <KeyboardControllerProvider>
-          <OnboardingProvider>
-            <AnalyticsContext>
-              <SessionProvider>
-                <PrefsStateProvider>
-                  <I18nProvider>
-                    <ShellStateProvider>
-                      <ModalStateProvider>
-                        <DialogStateProvider>
-                          <LightboxStateProvider>
-                            <PortalProvider>
-                              <BottomSheetProvider>
-                                <StarterPackProvider>
-                                  <SafeAreaProvider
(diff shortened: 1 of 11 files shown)


estrattonbailey (approved):

remcohaszing:
@estrattonbailey I see this was approved in november, but it’s still not merged. Is anything blocking this? Chats are pretty much unusable on mobile because of this.

mozzius:
Superseded by https://github.com/bluesky-social/social-app/pull/10230

PR #9334, opened 2025-11-05
Restore flatten to a place that is needed

mozzius:


src/components/Dialog/index.web.tsx
```diff
@@ -180,7 +180,8 @@ export function Inner({
         onClick={stopPropagation}
         onStartShouldSetResponder={_ => true}
         onTouchEnd={stopPropagation}
-        style={[
+        // note: flatten is required for some reason -sfn
+        style={flatten([
```
estrattonbailey, line 184: Ope, yep, it's because of the `asChild` on the `FocusScope` wrapper. It clones the child, and Radix expects objects.

mozzius: Ahhh ok

PR #9335, opened 2025-11-05
Fix Android crash by try/catch dialog open

mozzius:
We're seeing 100 crashes/day on Android from "Cannot read property 'open' of null" in `/Dialog/context.tsx`

It seems that the replication is to make a post and then immediately use the system back button. If you time it right, it'll catch it at just the wrong moment - the composer tries to handle the back press by opening the "discard post?" prompt, but the composer is dismissing so the dialog isn't available.

This seems like a flaw in our dialog API, but I don't have sufficient context to do a deeper fix beyond try/catching it. This specific call shouldn't crash the app, so a try/catch is probably good? But clearly something deeper is wrong with the way we're doing things

See [Sentry](https://blueskyweb.sentry.io/issues/6537272327/events/9a8ae09237834df7882d39f401995f01/?project=4508807082278912&query=is%3Aunresolved%20Cannot%20read%20property%20%27open%27%20of%20null&referrer=previous-event)

### DIFF
--- a/src/components/Dialog/context.ts
+++ b/src/components/Dialog/context.ts
@@ -7,13 +7,13 @@ import {
   useRef,
 } from 'react'
 
-import {logger} from '#/logger'
 import {useDialogStateContext} from '#/state/dialogs'
 import {
   type DialogContextProps,
   type DialogControlRefProps,
   type DialogOuterProps,
 } from '#/components/Dialog/types'
+import {IS_DEV} from '#/env'
 import {BottomSheetSnapPoint} from '../../../modules/bottom-sheet/src/BottomSheet.types'
 
 export const Context = createContext<DialogContextProps>({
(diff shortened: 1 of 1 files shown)


src/components/Dialog/context.ts
```diff
@@ -50,7 +51,15 @@ export function useDialogControl(): DialogOuterProps['control'] {
       id,
       ref: control,
       open: () => {
-        control.current.open()
+        try {
+          control.current.open()
+        } catch (err) {
```
estrattonbailey: Hmmm, so yeah `useImperativeHandle` hasn't registered the ref yet 🤔 this could also happen if you tried to call `open()` on a control that hasn't been passed to a `Dialog.Outer`.

So it's actually kinda expected, maybe optional chaining is fine here. Like if the Dialog hasn't even rendered yet, then calling `open` should be a no-op.

mozzius: I can make it `console.warn` in DEV? good compromise?

estrattonbailey: Sure that sounds good

mozzius: Done!

PR #9337, opened 2025-11-05
Move yt embed script to its own source file

DavidBuchanan314:
I'm looking into configuring a [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) for bsky.app.

As far as I can tell, this was the only "inline" script tag in the codebase (i.e. no `src` URL), and moving it to its own source file it will allow us to set a stricter CSP (no `unsafe-inline`).

I put the script alongside the HTML file rather than in `static/js/` because that path seems to be reserved for generated source and is handled specially in `server.go` - but I don't understand the build/bundling process very well so maybe there was something better I could have done.

I've tested this on localhost (via `go run ./cmd/bskyweb serve`) and it seems to work.

### DIFF
--- a/bskyweb/cmd/bskyweb/server.go
+++ b/bskyweb/cmd/bskyweb/server.go
@@ -223,7 +223,7 @@ func serve(cctx *cli.Context) error {
 		e.GET("/robots.txt", echo.WrapHandler(staticHandler))
 	}
 
-	e.GET("/iframe/youtube.html", echo.WrapHandler(staticHandler))
+	e.GET("/iframe/*", echo.WrapHandler(staticHandler))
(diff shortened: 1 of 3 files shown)


DavidBuchanan314:
Relatedly, I wonder if `IFRAME_HOST` could be `https://web-cdn.bsky.app/` instead of just `https://bsky.app/`, allowing both youtube.html and youtube.js to be served via CDN https://github.com/bluesky-social/social-app/blob/531ff6c43300e84edc8ca3ed15be834c82f4062a/src/lib/strings/embed-player.ts#L8-L15

ncsmatias (commented):
You might consider:

Adding error handling for missing or malformed parameters (e.g., when videoId or start aren’t provided).

Optionally setting responsive width/height for the player instead of fixed 1000px values, to improve layout flexibility.

Including comments inside the script to explain each section (API loading, parameter parsing, player creation) for maintainability.

PR #9339, opened 2025-11-06
Fix video letterboxing issue

mozzius:
In https://github.com/bluesky-social/social-app/pull/9019, we gave videos a background image of the thumbnail, as the actual player itself doesn't load until it gets closer to the screen. However, for videos with tall aspect ratios where we letterbox them slightly, this was visible around the edge of the player

   
Screenshot 2025-11-06 at 10 51 07

Simple fix is to put a background on the actual player. The player itself, once loaded, has it's own thumbnail

Screenshot 2025-11-06 at 10 50 17


estrattonbailey:
@mozzius whatcha think of this https://github.com/bluesky-social/social-app/pull/9371

PR #9342, opened 2025-11-06
[Threadgate] Refresh dialog

mozzius:
# Stacked on #9173 

Stacked on this:

- #9341
- https://github.com/bluesky-social/social-app/pull/9343

Screenshot 2025-11-06 at 17 25 49

PR #9343, opened 2025-11-06
[Threadgate] Refreshed toggle styles

mozzius:
# Stacked on #9342 

Screenshot 2025-11-06 at 19 23 06

### DIFF
--- a/src/components/forms/Toggle.tsx
+++ b/src/components/forms/Toggle.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import {createContext, useCallback, useContext, useMemo} from 'react'
 import {
   Pressable,
   type PressableProps,
@@ -6,13 +6,15 @@ import {
   View,
   type ViewStyle,
 } from 'react-native'
-import Animated, {LinearTransition} from 'react-native-reanimated'
+import Animated, {Easing, LinearTransition} from 'react-native-reanimated'
 
 import {HITSLOP_10} from '#/lib/constants'
+import {useHaptics} from '#/lib/haptics'
 import {isNative} from '#/platform/detection'
 import {
   atoms as a,
   native,
+  platform,
   type TextStyleProp,
   useTheme,
   type ViewStyleProp,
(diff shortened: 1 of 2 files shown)


src/components/forms/Toggle.tsx
estrattonbailey, line 1: Oddly seeing misalignment on the red ones?

CleanShot 2025-11-10 at 18 33 57@2x

estrattonbailey: Wait weird, I think it's a pixel rounding issue? Shows fine here, but when rendered to my screen the circle isn't aligned. Happens on the blue too, just more noticeable on red.

src/components/forms/Toggle.tsx
```diff
@@ -445,29 +521,27 @@ export function Radio() {
...
         a.rounded_full,
         t.atoms.border_contrast_high,
+        a.transition_color,
         {
           borderWidth: 1,
-          height: 24,
-          width: 24,
+          height: 25,
+          width: 25,
+          margin: -1,
```
estrattonbailey, line 528: Is this for visual alignment? Do we need on all sides or just the left?

src/components/forms/Toggle.tsx
```diff
@@ -445,29 +521,27 @@ export function Radio() {
         a.align_center,
         a.rounded_full,
         t.atoms.border_contrast_high,
+        a.transition_color,
         {
           borderWidth: 1,
-          height: 24,
-          width: 24,
+          height: 25,
```
estrattonbailey, line 526: Designs have this at 26px, and I asked Darrin if it should match the switch at 28px.

src/components/forms/Toggle.tsx
estrattonbailey: Weird! Not really actionable afaict, just looks off _sometimes_ on my screen.

src/components/forms/Toggle.tsx
```diff
@@ -113,25 +114,13 @@ export function Group({
     [type, onChange, values],
   )
 
-  React.useEffect(() => {
```
estrattonbailey, line 115: What was I on when I wrote this 😅

src/components/forms/Toggle.tsx
mozzius: My thinking was it allows it to be visually larger while remaining the same size as the checkbox. This is moot though if we end up making it 26px

PR #9346, opened 2025-11-07
fix: feed title line heights

caidanw:
Summary
---

Tall emojis like the box 📦 were being cut off due to too small of a line-height (leading).

So I've adjusted them to fit these emojis. Additionally I've adjusted some of the leading for the creator handle line below the feed title, such that it's a close as possible to the existing visual spacing.

Screenshots
---
image

image

<strong>After adding the emoji and adjusting leading</strong>
image

### DIFF
--- a/src/screens/Profile/components/ProfileFeedHeader.tsx
+++ b/src/screens/Profile/components/ProfileFeedHeader.tsx
@@ -442,12 +442,13 @@ function DialogInner({
 
         <View style={[a.flex_1, a.gap_2xs]}>
           <Text
-            style={[a.text_2xl, a.font_bold, a.leading_snug]}
-            numberOfLines={2}>
+            style={[a.text_2xl, a.font_bold, a.leading_tight]}
+            numberOfLines={2}
+            emoji>
             {info.displayName}
           </Text>
           <Text
-            style={[a.text_sm, a.leading_snug, t.atoms.text_contrast_medium]}
+            style={[a.text_sm, a.leading_relaxed, t.atoms.text_contrast_medium]}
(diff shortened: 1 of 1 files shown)


src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -235,7 +235,7 @@ export function ProfileFeedHeader({info}: {info: FeedSourceFeedInfo}) {
                         style={[
                           a.text_md,
                           a.font_bold,
-                          a.leading_tight,
+                          a.leading_relaxed,
                           gtMobile && a.text_lg,
                         ]}
                         numberOfLines={2}>
```
mozzius:
```suggestion
                        numberOfLines={2}
                        emoji>
```

src/screens/Profile/components/ProfileFeedHeader.tsx
```diff
@@ -441,12 +441,12 @@ function DialogInner({
 
         <View style={[a.flex_1, a.gap_2xs]}>
           <Text
-            style={[a.text_2xl, a.font_bold, a.leading_tight]}
+            style={[a.text_2xl, a.font_bold, a.leading_snug]}
             numberOfLines={2}>
```
mozzius:
```suggestion
            numberOfLines={2}
            emoji>
```

src/screens/Profile/components/ProfileFeedHeader.tsx
caidanw: Oh interesting, I didn't know there was an emoji prop. I'll add it if you think it's useful.

PR #9347, opened 2025-11-07
Tweak accessibility labels

surfdude29:
This PR suggests a few minor tweaks to a couple of accessibility labels added in #9213:

Change `account` to `user` for consistency with other usages in the dialog, and change `or` to `and/or` as both options may be selected:

`Would you like to block this account or delete this conversation?`
⬇️
`Would you like to block this user and/or delete this conversation?`

___

Add `user` to this string to improve clarity and for consistency with other strings in the dialog:

`Block and/or delete this conversation`
⬇️
`Block user and/or delete this conversation`

cc: @smileyhead


estrattonbailey:
Thanks!

PR #9350, opened 2025-11-07
[Threadgate] Add hint

mozzius:
# Stacked on #9173 

Reworked and refined the tooltip component. It now works within the composer, and works more consistently (currently there's a race condition which stops it appearing on mount - I had to add a delay to the profile button because it opens too fast now!).

# iOS

Had to add a special provider which lets it measure it's position relative to the sheet bounds rather than the screen bounds.

https://github.com/user-attachments/assets/6b3e2235-3205-4105-ab86-1fef1a1ef1af

# Android

https://github.com/user-attachments/assets/2a64c961-7905-4ed3-af3d-6f50a42af6f4

# Web

Note: one concern I have with this is it defocuses the keyboard, so you can't just start typing. Would maybe appreciate a hand getting this working?

https://github.com/user-attachments/assets/d95feceb-f4e8-41d2-b359-19dc1a3d46f8

### DIFF
--- a/src/components/Tooltip/index.tsx
+++ b/src/components/Tooltip/index.tsx
@@ -12,6 +12,7 @@ import {useWindowDimensions, View} from 'react-native'
 import Animated, {Easing, ZoomIn} from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 
+import {useIsKeyboardVisible} from '#/lib/hooks/useIsKeyboardVisible'
 import {GlobalGestureEventsProvider} from '#/state/global-gesture-events'
 import {atoms as a, select, useTheme} from '#/alf'
 import {useOnGesture} from '#/components/hooks/useOnGesture'
@@ -151,6 +152,7 @@ export function Target({children}: {children: React.ReactNode}) {
   const [hasLayedOut, setHasLayedOut] = useState(false)
   const targetRef = useRef<View>(null)
   const containerRef = useContext(TooltipProviderContext)
+  const keyboardIsOpen = useIsKeyboardVisible()
 
   useEffect(() => {
     if (!shouldMeasure || !hasLayedOut) return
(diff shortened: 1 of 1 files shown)


src/view/com/composer/threadgate/ThreadgateBtn.tsx
```diff
@@ -50,19 +74,32 @@ export function ThreadgateBtn({
...
+        position="top">
+        <Tooltip.Target>
+          <Button
+            color={showTooltip ? 'primary_subtle' : 'secondary'}
+            size="small"
+            testID="openReplyGateButton"
+            onPress={onPress}
+            label={label}
+            accessibilityHint={_(
+              msg`Opens a dialog to choose who can reply to this thread`,
```
surfdude29:
```suggestion
              msg`Opens a dialog to choose who can interact with this post`,
```
nit: Suggest updating this string to reflect that the user can control who quotes the post too

src/view/com/composer/threadgate/ThreadgateBtn.tsx
```diff
@@ -50,19 +74,32 @@ export function ThreadgateBtn({
...
+            )}>
+            <ButtonIcon icon={anyoneCanInteract ? EarthIcon : GroupIcon} />
+            <ButtonText numberOfLines={1}>{label}</ButtonText>
+            <ButtonIcon icon={ChevronDownIcon} size="xs" />
+          </Button>
+        </Tooltip.Target>
+        <Tooltip.TextBubble>
+          <Text>
+            <Trans>Psst! You can edit who can interact with this post.</Trans>
```
surfdude29, line 98:
```suggestion
            <Trans>Psst! You can choose who can interact with this post.</Trans>
```
~~nit: I think `choose` might be a bit clearer here~~

nvm, I forgot it's **Edit interaction settings** in the context menu

PR #9357, opened 2025-11-09
Add location attribute to profile:follow client event

abenzer:


abenzer:
Just noticed that we have a "routeName" attribute on the event, so we don't need this change.

PR #9364, opened 2025-11-10
Try and guard against play/pause function being called after player is destroyed

mozzius:
In the video feed, due to the double-tap feature, we call `player.pause()` etc after a short timeout. We can see this is causing errors due to the player getting destroyed while the timeout is in flight, and then we have a stale value for player meaning `player.playing` throws.

[Sentry here](https://blueskyweb.sentry.io/issues/6534817820/events/recommended/?project=4508807082278912&query=release%3A1.109.0%20level%3Afatal&referrer=recommended-event&sort=freq) - affects iOS and Android

The fix is two-pronged - we use `useNonReactiveCallback` and track if the component is mounted in an effect and use `isMounted.current` to try and avoid calling stale players, and then just to be safe I try/catch'd it, to stop this being a fatal error.


estrattonbailey (approved):
Seems reasonable to me!

PR #9366, opened 2025-11-10
Use new async method for `player.replace`

mozzius:
In the video feed, we use `player.replace` to replace the video source as you swipe. on iOS this is synchronous and leads to main thread hangs. `expo-video` has since added a new async method which does the loading on a different thread, with the sync method deprecated and soon to be removed, so we should use this new, better method.


tom-sherman:
What happens when you call replaceAsync twice in quick succession? Does `replaceAsync` handle cancellation/discarding of the previous result automatically?

mozzius:
Should do. Maybe we should wait for [this](https://github.com/expo/expo/pull/40265) to land

tom-sherman:
You're right, just checked the implementation and on both ios and android it's queued.

PR #9371, opened 2025-11-10
Video crop on web tweaks

estrattonbailey:
If we're gonna do a letterbox, we should just make it full-width imo. This PR:
- updates the thumbnail to be centered and without a repeat
- moves the bg and border out to the outer component so they apply to the thumb and the video itself
- removes `crop` from `VideoEmbed`, it's not being used and it's caused us confusion in the past
- removed `gtMobile` handling on web so that awkward screen sizes don't end up with massive videos playing

What do you think?

CleanShot 2025-11-10 at 17 40 02@2x
CleanShot 2025-11-10 at 17 40 19@2x


estrattonbailey:
@mozzius also, you can see that even on my 4k screen, the corners of the `MediaInsetBorder` get clipped. I think we may need to increase what constitutes "high DPI" or just use 1px.

PR #9383, opened 2025-11-13
Update button spacing

mozzius:
Figma: Bluesky Design System / Buttons (darrin left a comment)

Tweaks the horizontal padding for the new capsule buttons. Now button icons have some slight negative padding and a larger gap in the parent so that they are close to the edge than the text is, while keeping the same total gap. This should let us have the asymmetric padding without needing a manual `position` prop, which is annoying.

For large buttons with a button on the left:
Left side: 24px padding + -2px margin = 22px
Middle: 6px gap + -2px margin = 4px
Right side: 24px padding = 24px

Screenshot 2025-11-13 at 16 16 17

# Before / after

https://github.com/user-attachments/assets/75db8041-c1ad-48c3-b135-d402716603b6

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -442,20 +442,20 @@ export const Button = React.forwardRef<View, ButtonProps>(
         if (size === 'large') {
           baseStyles.push(a.rounded_full, {
             paddingVertical: 12,
-            paddingHorizontal: 25,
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Oh nice very clever!

Issue #9395, opened 2025-11-16
Android share intents should receive text with images


tomsqrd:
### Describe the Feature

With Android share intents, images can be sent to applications in combination with text. The bluesky Android application can receive such intents, however it only adds the shared image to the post

`intent/compose` links are able to receive both images and text, so it should be possible to receive both the image and (optional) text extras for share intents and combine them to an `intent/compose` link.

### Attachments

Here is what such an intent looks like in the share sheet, and how the bluesky app currently receives it:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/a19e41f0-0c9d-4b87-a05f-412a59a8ad0e" /> <img width="300" alt="Image" src="https://github.com/user-attachments/assets/6bf9e1b3-fceb-48a6-8c19-22652853d94c" />

mozzius:
Do you have an example of an app that sends these kinds of intents? I'm guessing that that's Letterboxd?

tomsqrd:
Unfortunately that's not letterboxd, the example is my own app that's still in development.

You can create an intent like this using Google's "Keep Notes" app, saving an image with a description, then sharing that via "Send" -> "Send via other apps".

Sharing it to WhatsApp for example, the text is then prefilled.

mozzius:
Ah perfect thank you!

PR #9396, opened 2025-11-16
Reading the optional String extras from attachment share intents

tomsqrd:
and adding them to the compose intents for images and videos if they exist.

This is a suggestion for solving #9395, enabling images to be sharable with text on the Android application.


mozzius:
this is great, thanks. will have a look at this when I get the chance

mozzius (approved):
Works well, great PR thank you!

PR #9400, opened 2025-11-17
[1.110.0 release prep] Translations

mozzius:


quiple:
It seems that it contains an untranslationd string that has not yet been reflected in Crowdin project.

mozzius:
Thanks @quiple - I'll re-sync it

surfdude29:
@quiple Oops, I think that was my bad (from the PRs I submitted), sorry 😬

@mozzius After you re-sync, would there be a bit of time before you pull the translations again? I could carry over the translations for the strings that my PRs tweaked to minimise untranslated strings.

mozzius:
Going to be doing it right about now, if that's ok. We can always OTA in the strings in a day or two

surfdude29:
Ah OK, that'll work 👌

PR #9406, opened 2025-11-18
Instantly remove posts from a feed when mute/blocking the author

### LINKED ISSUE #9446: Hide post instantly from search results when muting/blocking its poster
### Describe the Feature

When I search for something and see a post that I don't like, I mute/block its poster, but it still remains in the search results. I have to refresh just to have that post gone.

BlueSky should instantly hide posts from search results when muting/blocking an user.

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

mozzius:
`PostFeed` now watches the shadow cache for changes, and filters out newly mute/blocked post authors from feeds.

It shouldn't need to read the existing feed cache, as this is effectively only for optimistic updates. Refreshing the feed will obviously also strip out muted posts by default.

I've tried to make it as efficient as possible and reduce iterating over all the data constantly, but let me know if you can see any optimisations.

We had a previous PR which used the normal profile shadow inside the component and returned null if it was muted/blocked - however this had a huge amount of unintended side effects and weird/unexpected breakages. This approach doesn't catch everything, but I think we need to be really focused and targetted about where we apply this. The filter approach I think is best, because it only happens in specific screens. I also exempted author feeds, as they have their own special handling.

We could look at applying this to the notifications feed in future

https://github.com/user-attachments/assets/92524c17-3939-44fc-8665-5f33216b9377


estrattonbailey (approved):

IArnaut2:
It still does not work.

PR #9410, opened 2025-11-19
Android: Handle share intents when app first launches

### LINKED ISSUE #8513: Android share extension not working if app not open in background

mozzius:
Fixes #8513

The share intent handler only handles new incoming intents - if the app was initialised with an intent, it doesn't get picked up. Fix is to read the current intent in `OnCreate` and handle that

This is similar to #8513 but that one also adds a timeout - in my testing the timeout is not necessary. I checked it on a mid-range and low-end real device and it works fine.

# Test plan

Create a production build on a real device. Make sure the app is fully dead. Share something (i.e. a link) to Bluesky and confirm it works.

- [x] I have tested thoroughly myself

### DIFF
--- a/src/lib/hooks/useIntentHandler.ts
+++ b/src/lib/hooks/useIntentHandler.ts
@@ -24,7 +24,7 @@ const VALID_IMAGE_REGEX = /^[\w.:\-_/]+\|\d+(\.\d+)?\|\d+(\.\d+)?$/
 let previousIntentUrl = ''
 
 export function useIntentHandler() {
-  const incomingUrl = Linking.useURL()
+  const incomingUrl = Linking.useLinkingURL()
   const composeIntent = useComposeIntent()
   const verifyEmailIntent = useVerifyEmailIntent()
   const ageAssuranceRedirectDialogControl =
@@ -108,6 +108,7 @@ export function useIntentHandler() {
           } else {
             tryApplyUpdate(channel)
           }
+          return
         }
         default: {
           return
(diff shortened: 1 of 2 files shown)


src/lib/hooks/useIntentHandler.ts
```diff
@@ -24,7 +24,7 @@ const VALID_IMAGE_REGEX = /^[\w.:\-_/]+\|\d+(\.\d+)?\|\d+(\.\d+)?$/
 let previousIntentUrl = ''
 
 export function useIntentHandler() {
-  const incomingUrl = Linking.useURL()
+  const incomingUrl = Linking.useLinkingURL()
```
mozzius, line 27: `useURL` is deprecated in favour of `useLinkingURL`. I found that they act similarly, except `useLinkingURL` initialises with the correct value whereas `useURL` always initialises with `null`. Seems better.

PR #9412, opened 2025-11-19
Cache translation compilation action step

mozzius:
The translation step takes almost a full minute, and is getting slower. We can optimise this by caching the output, and only compiling/checking for errors on cache miss. We do this step in nearly all the different actions, so this way they can share the result between them.

~~Idk if it will work. Let's see!~~ It worked! In the CI for this PR, we can see it went from 54s in Run tests to 12s in Run linters

Screenshot 2025-11-19 at 14 52 35


mozzius:
#9905 makes compilation so fast, it's faster to compile than to cache 🤯 no need for this anymore

PR #9413, opened 2025-11-19
Use useConfirmEmail hook in deep link verification flow

mozzius:
Resolves the issue where verifying your email via deeplink wouldn't unlock the email gate straight away. The hook from the new flow refreshes the session after verifying, so the gate gets unlocked immediately.

Also cleaned up the nasty async effect into a nice mutation 

# Test plan

Use the deep link verification method, then press the composer button - the email gate should now let you in without needing to refresh


estrattonbailey (approved):
Nice, didn't test but looks safe to me

PR #9439, opened 2025-11-23
Add URL prewarming to `onPressIn` on links

mozzius:
Uses [`SFSafariViewController.prewarmConnections(to:)`](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller/prewarmconnections(to:)) to prewarm connections to both `go.bsky.app` and the destination URL. I'm hoping we might see a minor perceived speed increase?

Test plan: click on some links, seem if they feel faster I suppose. I wasn't able to get any concrete data but it does feel like it lingers on `go.bsky.app` a little less

### DIFF
--- a/modules/url-prewarm/expo-module.config.json
+++ b/modules/url-prewarm/expo-module.config.json
@@ -1,0 +1,6 @@
+{
+  "platforms": ["ios"],
+  "ios": {
+    "modules": ["UrlPrewarmModule"]
+  }
+}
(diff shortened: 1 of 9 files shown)


mary-ext:
wouldn't this introduce the fake clicks issue that X had also introduced with their iOS webview thingy

https://twitter.com/bluesky/status/1985795354009485826
https://twitter.com/bluesky/status/1985877210465714424

EDIT: did not see that it was onPressIn

mozzius:
@mary-ext I assume it's more like `<link rel="preconnect">`. it doesn't actually *preload* the page, just prewarms the connection

PR #9443, opened 2025-11-25
Fix messages input box visibilty

devxan:
I noticed when sending messages from the Bluesky website, the input box covered messages. 
image
This is due to the navigation bar, which is shown under certain viewport widths, but is still taken into account after hiding for that input's position. This PR should fix that for the website!
image
I've tested this on web and Android through [my Bluesky fork](https://bitchsky.app), but it should still work on iOS.

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -299,11 +299,18 @@ export function MessagesList({
     marginBottom: Math.max(keyboardHeight.get(), footerHeight.get()),
   }))
 
-  const animatedStickyViewStyle = useAnimatedStyle(() => ({
-    transform: [
-      {translateY: -Math.max(keyboardHeight.get(), footerHeight.get())},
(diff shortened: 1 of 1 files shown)


src/screens/Messages/components/MessagesList.tsx
```diff
@@ -299,11 +299,18 @@ export function MessagesList({
...
-  const animatedStickyViewStyle = useAnimatedStyle(() => ({
-    transform: [
-      {translateY: -Math.max(keyboardHeight.get(), footerHeight.get())},
-    ],
-  }))
+  const animatedStickyViewStyle = useAnimatedStyle(() => {
+    if (!isNative) {
+      return {
+        transform: [{translateY: 0}],
+      }
```
mozzius:
```suggestion
      return {}
```

mozzius:
I fixed this in #8543 - if that one doesn't get merged soon, I'll consider using this one :)

PR #9447, opened 2025-11-25
Patch in proguard rules to fix video fullscreen

### LINKED ISSUE #9437: Video fullscreen broken on Android
### Steps to Reproduce

Videos play normally, but tapping on it to enlarge it causes it to fade to black without fullscreening.

Everything else remains responsive - scrolling works, tab-switching works, even the non-fullscreen video controls work (pause/play and sound mute). The video continues playing (as indicated by the time left) and loops.

Scrolling away pauses the video (but it remains black), and resuming playback (by tapping the small play button) causes the video to restart. At this point, the video player shows the standard desktop controls (though attempting to interact with them causes the video to fade to black again), and the time remaining begins alternating between the previously shown and actual value.

### Attachments

Here's a video, though it sadly didn't record my tap locations. I interacted with the central part of the video (to try to fullscreen) and the play/sound controls in the corner, which caused different behaviors as described.

https://github.com/user-attachments/assets/959dc73e-96bc-4088-941d-80399adbbac3

Android

### Device Info

Fairphone 5G, Android 15

Build version: 1.110.1.741; Bundle info: 5d220e4 (prod); Bundle date: 25112018; Platform: android; Platform version: 35; Anonymous ID: 2945b823-4d0c-41f3-832e-314c6c92b62b

### Additional Information

_No response_

mozzius:
Maybe fixes #9437 

Video fullscreen mode has an issue with sometimes it not being able to find the activity. My theory is that when google play generates the builds from the AAB, it accidentally strips out the activity. I'm hoping explicitly whitelisting the file will do... something.

## Test plan

If you have the issue already, see if it fixes it.

If not...
¯\\_(ツ)_/¯

### DIFF
--- a/patches/@haileyok+bluesky-video+0.3.2.patch
+++ b/patches/@haileyok+bluesky-video+0.3.2.patch
@@ -1,0 +1,20 @@
+diff --git a/node_modules/@haileyok/bluesky-video/android/build.gradle b/node_modules/@haileyok/bluesky-video/android/build.gradle
+index b988d3f..7743421 100644
+--- a/node_modules/@haileyok/bluesky-video/android/build.gradle
++++ b/node_modules/@haileyok/bluesky-video/android/build.gradle
+@@ -36,6 +36,7 @@ android {
+   defaultConfig {
+     versionCode 1
+     versionName "0.1.0"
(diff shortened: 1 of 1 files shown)


mozzius:
This is actually maybe not it

PR #9448, opened 2025-11-26
feat: convert saved feeds list to be draggable

espeon:
Attempted a rework of the active saved feeds list to have the elements in said list to be draggable. Note that it introduces a new JS dependency (react-native-sortables) (no native deps afaik) and replaces the real border with a fake one that fades out and scales in a bit on drag. The vertical handle icon icon is clearly temporary and taken from Lucide.

https://github.com/user-attachments/assets/73d8f17c-7eb0-44d7-931b-caf33d75582e


plugins/starterPackAppClipExtension/withAppEntitlements.js
```diff
@@ -6,9 +6,9 @@ const withAppEntitlements = config => {
     config.modResults['com.apple.security.application-groups'] = [
       `group.app.bsky`,
     ]
-    config.modResults[
-      'com.apple.developer.associated-appclip-app-identifiers'
-    ] = [`$(AppIdentifierPrefix)${config.ios.bundleIdentifier}.AppClip`]
+    // config.modResults[
+    //   'com.apple.developer.associated-appclip-app-identifiers'
+    // ] = [`$(AppIdentifierPrefix)${config.ios.bundleIdentifier}.AppClip`]
```
mozzius: pls rm this change :)

src/screens/SavedFeeds.tsx
```diff
@@ -356,6 +340,29 @@ function ListItem({
...
+      ),
+    }
+  })
+
+  return (
+    <Animated.View style={[a.border_b, t.atoms.bg, borderOpacityStyle]}>
+      {children}
     </Animated.View>
   )
 }
```
mozzius: Changed it so the border animation is linked to the activation animation :)

mozzius:
Looking good! I made some tweaks.

This library seems much higher quality than any we'd tried before (I guess the ecosystem is improving?). That said, the only real issue I've found is the insanely bad perf it has on initial render, especially on Android. Like what is this lmao

https://github.com/user-attachments/assets/36be05cb-a371-4cde-aed5-ae2c3bdd0f14

src/screens/SavedFeeds.tsx
```diff
@@ -356,6 +342,29 @@ function ListItem({
...
+  const transparentBorderColor = t.atoms.bg.backgroundColor
+  const borderOpacityStyle = useAnimatedStyle(() => {
+    return {
+      borderColor: interpolateColor(
+        activationAnimationProgress.get(),
+        [0, 1],
+        [borderColor, transparentBorderColor],
+      ),
+    }
+  })
```
mozzius, line 369: Linked the border color animation to the actual activation animation so it's way smoother

espeon: ah cool, TIL you can do this in reanimated

PR #9449, opened 2025-11-26
Use FULL_SCREEN presentation for IAB

mozzius:
The sheet presentation is quite annoying to dismiss. Also, Apple says we shouldn't use it lol:
  
Screenshot 2025-11-26 at 10 17 05

Let's use the FULL_SCREEN presentation like how every other app does:

https://github.com/user-attachments/assets/bd23b209-1af0-4260-825d-f2958271917c

### DIFF
--- a/src/lib/hooks/useOpenLink.ts
+++ b/src/lib/hooks/useOpenLink.ts
@@ -16,13 +16,11 @@ import {isNative} from '#/platform/detection'
 import {useInAppBrowser} from '#/state/preferences/in-app-browser'
 import {useTheme} from '#/alf'
 import {useDialogContext} from '#/components/Dialog'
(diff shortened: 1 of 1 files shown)


surfdude29:
> The sheet presentation is quite annoying to dismiss. Also, Apple says we shouldn't use it lol:

And yet, Apple Maps uses a sheet presentation for its IAB 🧐

(I think this is a good change for social-app, I've accidentally swiped down and closed the IAB sheet by mistake a couple of times in the last few days alone 🙃)

mozzius:
It's situation-dependent. Apple Maps is like, all sheets, so it's a sensible choice there. social-app has a more standard stack-based navigation ui so the more standard presentation works better

estrattonbailey (approved):

PR #9451, opened 2025-11-26
Catch crop cancelled errors

mozzius:
We can safely ignore them. At some point the error message changed from "User cancelled" to "Crop cancelled" so I changed it to catch both cases

### DIFF
--- a/src/lib/strings/errors.ts
+++ b/src/lib/strings/errors.ts
@@ -53,6 +53,13 @@ export function isErrorMaybeAppPasswordPermissions(e: unknown) {
   return str.includes('Bad token scope') || str.includes('Bad token method')
 }
 
+/**
+ * Intended to capture "User cancelled" or "Crop cancelled" errors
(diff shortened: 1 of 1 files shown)


src/lib/strings/errors.ts
```diff
@@ -52,3 +52,8 @@ export function isErrorMaybeAppPasswordPermissions(e: unknown) {
   const str = String(e)
   return str.includes('Bad token scope') || str.includes('Bad token method')
 }
+
+export function isCancelledError(e: unknown) {
+  const str = String(e).toLowerCase()
+  return str.includes('cancel')
+}
```
estrattonbailey, line 66: Let's at least name this more specifically or leave a comment or something about the intent of this method 👍

estrattonbailey (approved):

src/lib/strings/errors.ts
mozzius: Added a comment 👍

PR #9456, opened 2025-11-27
Process videos on web with WebCodecs

mozzius:
Using the excellent [MediaBunny](https://mediabunny.dev/) library we can finally do client-side video compression efficiently. The existing system is maintained as a fallback as browser support is most-of-the-way there but not complete (Firefox for Android 👀).

This means that videos >25mb can get compressed before upload, allowing much greater upload speeds and an overall reduction in the time it takes to upload a video. I also cleaned up the processing pipeline, meaning we no longer rely on Data URIs to pass data around - big win!
  

https://github.com/user-attachments/assets/27d83104-56a0-450e-a2bb-8e93d8d135d6

# Test plan

Upload videos with various audio/video codecs and file types in different sizes. It should be able to handle normal MP4s just fine, and will bail out if there's a codec it doesn't recognise.

I found that it can cope with ~300mb videos max before the browser starts to fall apart.


mary-ext:
Mediabunny doesn't support GIFs? I think you might still be able to jerry-rig something with WebCodecs itself

here's how I've converted GIFs to WEBM via the older webm-muxer library (same author, predates Mediabunny, I haven't upgraded yet)

https://github.com/mary-ext/aglais/blob/95a70f438922d00b147243b5971cdc7ed999610d/src/components/composer/workers/gif-conversion.ts#L2

mozzius:
Oooh interesting thank you! It would also be super useful to be able to tell how many frames the gif has, so we can just use it as an image if it only has one

mozzius:
Oooh interesting thank you! It would also be super useful to be able to tell how many frames the gif has, so we can just use it as an image if it only has one

mary-ext:
updated to use Mediabunny, so hopefully it's better as a reference
https://github.com/mary-ext/aglais/blob/095847d4a343d928251ca39411975c669178f049/src/components/composer/workers/gif-conversion.ts

estrattonbailey:
Superseded by #10955

PR #9458, opened 2025-11-27
optimize rendering of fake verification symbols

easrng:
image


mozzius:
oh jeez I guess we should put that Unicode symbol in sanitizeHandle 😭

PR #9464, opened 2025-11-30
Adds a composer prompt on the home screen to invite users to post

abenzer:


src/view/com/posts/PostFeed.tsx
```diff
@@ -501,6 +507,18 @@ let PostFeed = ({
                           'interstitial2-' + sliceIndex + '-' + lastFetchedAt,
                       })
                     }
+                    // Show composer prompt for Discover and Following feeds
+                    if (
+                      hasSession &&
+                      (feedUriOrActorDid === DISCOVER_FEED_URI ||
+                        feed === 'following') &&
+                      gate('show_composer_prompt')
+                    ) {
```
estrattonbailey, line 526: @abenzer for cases like this, we want to put the gate check last so that we don't count exposures for logged out users, or users of other feeds. This makes it so that ONLY logged in users of Discover/Following are part of the experiment.

estrattonbailey (approved):
Nice one!

PR #9471, opened 2025-12-01
ALF content language dialog

mozzius:
Extracts out the post language dialog into a generic language select dialog, then uses it for the content language select in settings.

I used the new `Panel` style - lemme know if that's too jarring. It's quite chunky. I made the Selects larger to match

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

Web:

Screenshot 2025-12-01 at 14 38 47

### DIFF
--- a/src/components/dialogs/LanguageSelectDialog.tsx
+++ b/src/components/dialogs/LanguageSelectDialog.tsx
@@ -186,7 +186,7 @@ export function DialogInner({
               a.text_xl,
               a.mb_sm,
             ]}>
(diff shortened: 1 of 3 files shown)


modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
estrattonbailey, line 1: Caught a stray when merging main, but so minor I'll just keep it

Issue #6825, opened 2024-11-28
Translate picklist for primary language


monty241:
### Steps to Reproduce

See picture of /appsettings/language. The primary language is always English. In Italy only 13% of the people can read this.

Please translate content to current UI language as already done in app-language directly above it.

![image](https://github.com/user-attachments/assets/f44b874b-c69f-4de3-b064-0e43fe87454e)

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

v1.95.0

surfdude29:
I think this would be fixed – partially, at least – by #6207.

surfdude29:
Now that #6207 has been merged I think this should be fixed on web, but from the [comment here](https://github.com/bluesky-social/social-app/pull/6207#issuecomment-2566735583) further work would be needed so that it works on iOS and Android.

PR #7033, opened 2024-12-10
Nepali Translation (divyaswormakai's reviewed & some changes) 

whoisanku:
I looked over the PR #6943 from @divyaswormakai , looks good on his side most of the things but added some additional changes to make it better on language side.

Issue #6826, opened 2024-11-28
Translate content languages (see also 6824 and 6825)


monty241:
### Steps to Reproduce

See picture. Only 13% of Italy speaks English, so they can not read this. Please translate to current UI language.

![image](https://github.com/user-attachments/assets/3cb32924-d352-46a8-8735-8d816c804453)

### What platform(s) does this occur on?

Web (Desktop)

### What version of the app are you using?

1.95.0

auroursa:
#6207 will fixed this issue.

surfdude29:
Similar to #6825, now that #6207 has been merged I think this should be fixed on web, but from the [comment here](https://github.com/bluesky-social/social-app/pull/6207#issuecomment-2566735583) further work will be needed so that it works on iOS and Android.

PR #9473, opened 2025-12-01
Telephone country code select

mozzius:
Redesigned select input + added telephone-specific input. Automatically adapts to user's locale to autofill

Simulator Screenshot - iPhone 17 - 2025-12-01 at 19 25 59

https://github.com/user-attachments/assets/3262f342-45fe-4ffa-ad83-798b8f034cc7

### DIFF
--- a/src/lib/international-telephone-codes.ts
+++ b/src/lib/international-telephone-codes.ts
@@ -1254,11 +1254,9 @@ export const INTERNATIONAL_TELEPHONE_CODES = {
 const DEFAULT_PHONE_COUNTRY = 'US'
 
 export function getDefaultCountry(location?: {countryCode?: string}) {
-  const supportedCountries = Object.keys(INTERNATIONAL_TELEPHONE_CODES)
-
   if (
     location?.countryCode &&
-    supportedCountries.includes(location.countryCode.toUpperCase())
+    location.countryCode.toUpperCase() in INTERNATIONAL_TELEPHONE_CODES
   ) {
     return location.countryCode.toUpperCase()
   }


src/lib/international-telephone-codes.ts
```diff
@@ -0,0 +1,1530 @@
...
+    svgFlag: require('../../assets/icons/flags/SD.svg'),
+  },
+  LA: {
+    name: (i18n: I18n) => i18n._(msg`Lao People's Democratic Republic`),
+    code: '+856',
+    unicodeFlag: '🇱🇦',
+    svgFlag: require('../../assets/icons/flags/LA.svg'),
+  },
+  TW: {
+    name: (i18n: I18n) => i18n._(msg`Taiwan, Province of China`),
```
surfdude29:
```suggestion
    name: (i18n: I18n) => i18n._(msg`Taiwan`),
```
I think it would be better to refer simply to `Taiwan`, as in #2797 :)

mozzius: Good catch, missed that one

surfdude29:
_(this is a copy of [a message I posted to the Bluesky Translators Discord](https://discord.com/channels/1338165055992107061/1338165056596082731/1446515043490336848))_

If I've understood correctly, this PR (a part of the upcoming opt-in [contacts matching feature](https://docs.bsky.app/blog/contact-import-rfc)) will add a translatable string for each country/territory name.

I was wondering if it might be feasible to implement something that uses the Unicode CLDR data for territories – [this is the data for English](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-localenames-full/main/en/territories.json), for example – that would mean that country and territory names would be automatically translated, rather than translators having to translate each string individually on Crowdin (even if they're just copying and pasting from the CLDR data).

Like something similar to what @Signez implemented in #6207 to localize language selectors using `Intl.DisplayName`? 🤔

surfdude29:
[According to Claude](https://claude.ai/share/ae787cdc-0aa6-4fe0-8579-03ed0d3bcc1c), the `Intl.DisplayNames` API would cover this too, although he also mentioned the option of a polyfill.

mozzius:
Great suggestion @surfdude29! Switched to use `Intl.DisplayNames` and added a polyfill for native

src/locale/i18n.ts
```diff
@@ -256,18 +282,25 @@ export async function dynamicActivate(locale: AppLanguage) {
...
     }
     case AppLanguage.ne: {
       i18n.loadAndActivate({locale, messages: messagesNe})
+      await Promise.all([
+        import('@formatjs/intl-pluralrules/locale-data/ne'),
+        import('@formatjs/intl-numberformat/locale-data/ne'),
+        import('@formatjs/intl-displaynames/locale-data/ne'),
+      ])
       break
     }
```
mozzius, line 297: @surfdude29 any idea why Nepali was missing locale data? Seems available so I added it

surfdude29: @mozzius Good catch!

Looking at [the PR that added support for the Nepali translation](https://github.com/bluesky-social/social-app/pull/7033), I think I remember what may have happened: I checked the `localedata` directories for the `@formatjs/intl-pluralrules` and `@formatjs/intl-numberformat` packages on npmjs.com, saw that there was no entry for `np`, and assumed that Nepali wasn't supported.

However, as @/quiple pointed out in #7265, the correct ISO code for the Nepali language is actually `ne` – `np` is the ISO code for the country Nepal (similar to how `ja` is the code for the Japanese language, while `jp` is the code for Japan... although of course sometimes the codes are the same, e.g. `de`, `es` and `fr` are both country and language codes 😅). So the language code got fixed, but I didn't then think to recheck the `localedata` directories for `ne` – which, as you say, is supported.

Thanks for spotting it and fixing it now :)

surfdude29:
> Great suggestion @surfdude29! Switched to use `Intl.DisplayNames` and added a polyfill for native

Nice! As a side benefit of adding the polyfill, I think that means that language names will also now be translated on native, meaning that this will fix #6825 and #6826?

src/components/Select/types.ts
```diff
@@ -123,9 +123,16 @@ export type ValueProps = {
   /**
    * Only needed for native. Extracts the label from an item. Defaults to `item => item.label`
    */
-  children?: (value: any) => string
+  children?: (value: any) => React.ReactNode
   placeholder?: string
   style?: StyleProp<TextStyle>
+  /**
+   * By default, web just extracts the component from inside the dropdown and portals it in here.
```
estrattonbailey, line 130: In where?

src/lib/international-telephone-codes.ts
```diff
@@ -0,0 +1,1275 @@
...
+}
+
+const DEFAULT_PHONE_COUNTRY = 'US'
+
+export function getDefaultCountry(location?: {countryCode?: string}) {
+  const supportedCountries = Object.keys(INTERNATIONAL_TELEPHONE_CODES)
+
+  if (
+    location?.countryCode &&
+    supportedCountries.includes(location.countryCode.toUpperCase())
```
estrattonbailey: Could just do `location.countryCode.toUpperCase() in INTERNATIONAL_TELEPHONE_CODES` or `INTERNATIONAL_TELEPHONE_CODES[cc]` and skip the array scan. You do this below already :P

src/components/Select/types.ts
mozzius: Into the <Select.Trigger> component, it's default Radix behaviour thats hard to replicate on native

src/lib/international-telephone-codes.ts
mozzius: Smart!

surfdude29:
Tested on preview branch, works really well. I think it's especially cool how the countries are sorted by their translated name, not their English name or ISO code. Very nicely done! 👌

PR #9479, opened 2025-12-04
Age Assurance V2

estrattonbailey:
Since adding Age Assurance and regional access controls earlier this year, more countries and jurisdictions have enacted Age Assurance laws that Bluesky must follow in order to remain legally compliant. For example, Australia and Virginia both have new requirements on the horizon that our V1 implementation cannot account for. And previously, we made the difficult decision to block Mississippi entirely due to the additional complexity required to support that region at the time.

Because of these new requirements, and with the prospect of more to come in the future, we've decided to dedicate some time to building out a V2 of Age Assurance, which can handled these more complex rulesets.

This PR unifies the separate systems we have been using into a single "V2" system that will work for all regions. This includes Australia, Virginia, and even Mississippi.

Appview companion PR: https://github.com/bluesky-social/atproto/pull/4407

### DIFF
--- a/src/components/dialogs/BirthDateSettings.tsx
+++ b/src/components/dialogs/BirthDateSettings.tsx
@@ -40,11 +40,11 @@ export function BirthDateSettingsDialog({
       <Dialog.Handle />
       {isBirthdateUpdateAllowed ? (
         <Dialog.ScrollableInner
-          label={_(msg`My Birthday`)}
+          label={_(msg`My Birthdate`)}
           style={web({maxWidth: 400})}>
           <View style={[a.gap_md]}>
             <Text style={[a.text_xl, a.font_semi_bold]}>
-              <Trans>My Birthday</Trans>
+              <Trans>My Birthdate</Trans>
             </Text>
             <Text
               style={[a.text_md, a.leading_snug, t.atoms.text_contrast_medium]}>
@@ -60,7 +60,7 @@ export function BirthDateSettingsDialog({
                 message={
                   error?.toString() ||
                   _(
-                    msg`We were unable to load your birth date preferences. Please try again.`,
+                    msg`We were unable to load your birthdate preferences. Please try again.`,
                   )
                 }
                 style={[a.rounded_sm]}
@@ -74,7 +74,7 @@ export function BirthDateSettingsDialog({
         </Dialog.ScrollableInner>
       ) : (
         <Dialog.ScrollableInner
-          label={_(msg`You recently changed your birthday`)}
+          label={_(msg`You recently changed your birthdate`)}
           style={web({maxWidth: 400})}>
           <View style={[a.gap_sm]}>
             <Text
@@ -84,12 +84,12 @@ export function BirthDateSettingsDialog({
                 a.leading_snug,
                 {paddingRight: 32},
               ]}>
-              <Trans>You recently changed your birthday</Trans>
+              <Trans>You recently changed your birthdate</Trans>
             </Text>
             <Text
               style={[a.text_md, a.leading_snug, t.atoms.text_contrast_medium]}>
               <Trans>
-                There is a limit to how often you can change your birthday. You
+                There is a limit to how often you can change your birthdate. You
                 may need to wait a day or two before updating it again.
               </Trans>
             </Text>
@@ -144,8 +144,8 @@ function BirthdayInner({
           testID="birthdayInput"
           value={date}
           onChangeDate={newDate => setDate(new Date(newDate))}
-          label={_(msg`Birthday`)}
-          accessibilityHint={_(msg`Enter your birth date`)}
+          label={_(msg`Birthdate`)}
+          accessibilityHint={_(msg`Enter your birthdate`)}
         />
       </View>
 
@@ -178,7 +178,7 @@ function BirthdayInner({
 
       <View style={isWeb && [a.flex_row, a.justify_end]}>
         <Button
-          label={hasChanged ? _(msg`Save birthday`) : _(msg`Done`)}
+          label={hasChanged ? _(msg`Save birthdate`) : _(msg`Done`)}
           size="large"
           onPress={onSave}
           variant="solid"


src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+      // So we change the URL ourselves. The navigator will pick it up on remount.
+      history.pushState(null, '', '/')
+    }
+    logoutCurrentAccount('AgeAssuranceNoAccessScreen')
+  }, [logoutCurrentAccount])
+
+  const birthdayUpdateText = canUpdateBirthday ? (
+    <Text style={[textStyles]}>
+      <Trans>
+        If your birth date is not accurate, you can update it by{' '}
```
surfdude29:
```suggestion
        If your birthdate is not accurate, you can update it by{' '}
```
~~nit: The strings are currently inconsistent, but I suggest it might be a good idea to standardize on either `birthdate` or `date of birth`.~~

**EDIT:** I apologize for getting this wrong, after looking it up I found out that `birth date` is actually the correct usage in US English, it's only `birthdate` in British English. I'll delete or amend my suggestions as appropriate.

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+        <SimpleInlineLinkText
+          label={_(msg`Click here to update your birthday`)}
```
surfdude29:
```suggestion
          label={_(msg`Click here to update your birth date`)}
```
~~nit: as above~~

**EDIT:** As explained above, I suggest standardizing on `birth date`.

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+                    {birthdayUpdateText}
+                  </View>
+
+                  <AccessSection />
+                </>
+              ) : (
+                <View style={[a.gap_lg]}>
+                  <Text style={[textStyles]}>
+                    <Trans>
+                      Unfortunately, the birth date you have saved to your
```
surfdude29:
```suggestion
                      Unfortunately, the birthdate you have saved to your
```
nit: as above

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+                  {birthdayUpdateText}
+                </View>
+              )}
+            </>
+          ) : (
+            <View style={[a.gap_lg]}>
+              <Text style={[textStyles]}>
+                <Trans>
+                  Looks like you haven't added your birth date. You must provide
```
surfdude29:
```suggestion
                  It looks like you haven't added your birth date. You must provide
```
nit: suggest slight rewording

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+                  an accurate date of birth to use Bluesky.
+                </Trans>
+              </Text>
+              <Button
+                color="primary"
+                size="large"
+                label={_(msg`Click here to update your birthday`)}
```
surfdude29:
```suggestion
                label={_(msg`Click here to update your birth date`)}
```
nit: as above

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+                onPress={() => birthdayControl.open()}>
+                <ButtonText>
+                  <Trans>Add your birth date</Trans>
```
surfdude29:
```suggestion
                  <Trans>Add your birthdate</Trans>
```
nit: as above

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+                </ButtonText>
+              </Button>
+            </View>
+          )}
+
+          <View style={[a.pt_lg, a.gap_xl]}>
+            <Logo width={120} textFill={t.atoms.text.color} />
+            <Text style={[a.text_sm, a.italic, t.atoms.text_contrast_medium]}>
+              To log out,{' '}
```
surfdude29: `<Trans>` tags are missing for this string

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,65 @@ export function BirthDateSettingsDialog({
...
-            <Trans>My Birthday</Trans>
-          </Text>
-          <Text style={[a.leading_snug, t.atoms.text_contrast_medium]}>
-            <Trans>
-              This information is private and not shared with other users.
-            </Trans>
-          </Text>
+      {isBirthdateUpdateAllowed ? (
+        <Dialog.ScrollableInner
+          label={_(msg`My Birthday`)}
```
surfdude29:
```suggestion
          label={_(msg`My Birth date`)}
```
nit: As mentioned previously, suggest standardizing on `birth date`

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,65 @@ export function BirthDateSettingsDialog({
...
+          style={web({maxWidth: 400})}>
+          <View style={[a.gap_md]}>
+            <Text style={[a.text_xl, a.font_semi_bold]}>
+              <Trans>My Birthday</Trans>
```
surfdude29:
```suggestion
              <Trans>My Birth date</Trans>
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,65 @@ export function BirthDateSettingsDialog({
...
-            />
-          ) : (
-            <BirthdayInner control={control} preferences={preferences} />
-          )}
-        </View>
+          <Dialog.Close />
+        </Dialog.ScrollableInner>
+      ) : (
+        <Dialog.ScrollableInner
+          label={_(msg`My Birthday`)}
```
surfdude29:
```suggestion
          label={_(msg`My Birth date`)}
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,65 @@ export function BirthDateSettingsDialog({
...
+          style={web({maxWidth: 400})}>
+          <View style={[a.gap_sm]}>
+            <Text style={[a.text_xl, a.font_semi_bold]}>
+              <Trans>You recently changed your birthday</Trans>
```
surfdude29:
```suggestion
              <Trans>You recently changed your birth date</Trans>
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,65 @@ export function BirthDateSettingsDialog({
...
+            </Text>
+            <Text
+              style={[a.text_md, a.leading_snug, t.atoms.text_contrast_medium]}>
+              <Trans>
+                There is a limit to how often you can change your birth date.
```
surfdude29:
```suggestion
                There is a limit to how often you can change your birthdate.
```
nit: as above

surfdude29:
IMG_2853

The title text doesn't seem to wrap here 👀

src/ageAssurance/components/NoAccessScreen.tsx
estrattonbailey: I've gone back and forth on this a couple times actually. I prefer `birthdate` (one word) as well, and it seems to be generally accepted as a valid spelling, maybe even preferred for UI contexts based on other apps. I think I'll go with `birthdate` 👍

surfdude29:
@estrattonbailey Sorry about getting `birth date` vs. `birthdate` wrong in my original suggestions, I've now amended/deleted as appropriate to use `birth date`.

EDIT: D'oh, I only just saw your comment after I'd finished editing my suggestions 🤦🏻‍♂️🫠

src/ageAssurance/components/NoAccessScreen.tsx
estrattonbailey: Eh, I think it's fine

estrattonbailey:
@surfdude29 lol np at all. I like birthdate better, and it seems common enough. Not gonna make or break anything imo, so let's roll with it. Bigger fish to fry on this PR 🫠 

Edit: also thank you for your help, as always!

__tests__/lib/string.test.ts
```diff
@@ -957,20 +957,20 @@ describe('parseStarterPackHttpUri', () => {
   })
 
   it('returns the at uri when the input is a valid starterpack at uri', () => {
-    const validAtUri = 'at://did:123/app.bsky.graph.starterpack/rkey'
+    const validAtUri = 'at://did:plc:123/app.bsky.graph.starterpack/rkey'
     expect(parseStarterPackUri(validAtUri)).toEqual({
-      name: 'did:123',
+      name: 'did:plc:123',
```
estrattonbailey, line 962: Matthieu's latest updates made validation of `new AtUri()` more strict

src/ageAssurance/components/RedirectOverlay.tsx
estrattonbailey, line 1: @mozzius had to reimplement the redirect dialog because the `NoAccessScreen` blocks the component tree. I had the existing `AgeAssuranceRedirectDialog` mounted in the `NoAccessScreen` AND within the shell, controlled by `useIntentHandler`. When a redirect came in, `AgeAssuranceRedirectDialog` would flash open, AA state would update in the bg and the `NoAccessScreen` would unmount, the main tree would mount, and `useIntentHandler` had already run, so the redirect dialog would not open again.

So the fix is to hoist this up a level, but that makes using dialog portals a mess. So I opted to just make it a fullscreen overlay. Doesn't animate, but it allows for the component tree to unmount/remount stuff in the bg while the user gets a smooth experience. The logic in this component is the same, just the outer UI has changed to overlay everything.

src/ageAssurance/components/RedirectOverlay.tsx
```diff
@@ -0,0 +1,334 @@
...
+  return isOpen ? (
+    <FullWindowOverlay>
+      <View
+        style={[
+          a.fixed,
+          a.inset_0,
+          // setting a zIndex when using FullWindowOverlay on iOS
+          // means the taps pass straight through to the underlying content (???)
+          // so don't set it on iOS. FullWindowOverlay already does the job.
+          !isIOS && {zIndex: 9999},
```
estrattonbailey, line 148: Copied some things from the PolicyUpdateOverlay component I built a while back

src/ageAssurance/components/RedirectOverlay.tsx
```diff
@@ -0,0 +1,334 @@
...
+          a.align_center,
+          // @ts-ignore
+          platform({
+            web: {
+              paddingTop: '35vh',
+            },
+            default: {
+              paddingTop: Dimensions.get('window').height * 0.35,
+            },
+          }),
```
estrattonbailey, line 160: Iffy centering logic, but works

src/ageAssurance/data.tsx
```diff
@@ -0,0 +1,496 @@
...
+  defaultOptions: {
+    queries: {
+      /**
+       * We clear this manually, so disable automatic garbage collection.
+       * @see https://tanstack.com/query/latest/docs/framework/react/plugins/persistQueryClient#how-it-works
+       */
+      gcTime: Infinity,
+    },
+  },
+})
```
estrattonbailey, line 40: Note custom `QueryClient` here

package.json
```diff
@@ -103,7 +103,7 @@
     "@react-navigation/native-stack": "^7.3.13",
     "@sentry/react-native": "~6.20.0",
     "@tanstack/query-async-storage-persister": "^5.25.0",
-    "@tanstack/react-query": "^5.8.1",
+    "@tanstack/react-query": "5.25.0",
```
mozzius, line 106: Any particular reason for this specific version? Latest is 5.90.12

src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -150,11 +151,11 @@ function Inner() {
               <Trans>
                 We're having issues initializing the age assurance process for
                 your account. Please{' '}
-                <InlineLinkText
+                <SimpleInlineLinkText
```
estrattonbailey, line 154: Needed to swap these out since this dialog is used in the `NoAccessScreen`, which is outside the nav context.

src/components/ageAssurance/AgeAssuranceRedirectDialog.tsx
estrattonbailey, line 1: This file isn't used atm, but it's been updated to support V2.

src/components/ageAssurance/useAgeAssuranceCopy.ts
```diff
@@ -2,20 +2,28 @@ import {useMemo} from 'react'
...
-        msg`The laws in your location require you to verify you're an adult before accessing certain features on Bluesky, like adult content and direct messaging.`,
-      ),
+      notice:
+        aa.state.access === aa.Access.Safe
+          ? _(
+              msg`Due to laws in your region, certain features on Bluesky are currently restricted until you're able to verify you're an adult.`,
+            )
+          : _(
+              msg`The laws in your location require you to verify you're an adult before accessing certain features on Bluesky, like adult content and direct messaging.`,
+            ),
```
estrattonbailey, line 20: New fork in copy for users in `safe` mode

src/Splash.web.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const size = 100
+const ratio = 57 / 64
+
+export function Splash() {
+  return (
+    <View style={[a.fixed, a.inset_0, a.align_center, a.justify_center]}>
+      <Svg
+        fill="none"
+        viewBox="0 0 64 57"
```
mozzius, line 20: Nice, this makes things way smoother. We could add an exit animation to this potentially

src/components/dialogs/DeviceLocationRequestDialog.tsx
```diff
@@ -76,9 +70,8 @@ function DeviceLocationRequestDialogInner({onLocationAcquired}: Props) {
         const location = req.location
 
         if (location && location.countryCode) {
-          const geolocationStatus = computeGeolocationStatus(location, config)
           onLocationAcquired?.({
-            geolocationStatus,
+            geolocation: location,
```
estrattonbailey, line 74: This dialog was intended to be generic, but I didn't write it completely generic previously. This check has just moved to the callsites and out of this dialog.

src/components/Link.tsx
```diff
@@ -448,7 +448,9 @@ export function SimpleInlineLinkText({
     href = createProxiedUrl(href)
   }
 
-  const onPress = () => {
+  const onPress = (e: GestureResponderEvent) => {
+    const exitEarlyIfFalse = outerOnPress?.(e)
+    if (exitEarlyIfFalse === false) return
     Linking.openURL(href)
   }
```
estrattonbailey, line 455: Copied logic from other link components to the "simple" one, same handling

src/components/Link.tsx
```diff
@@ -517,7 +519,7 @@ export function WebOnlyInlineLinkText({
 export function createStaticClick(
   onPressHandler: Exclude<BaseLinkProps['onPress'], undefined>,
 ): {
-  to: BaseLinkProps['to']
+  to: string
```
estrattonbailey, line 522: Just here for types for compat with SimpleInlineLinkText

src/lib/__tests__/parseLinkingUrl.test.ts
estrattonbailey, line 1: Needed this for new `RedirectOverlay` URL parsing, plus we should write tests for weird stuff like this

src/lib/api/resolve.ts
```diff
@@ -163,6 +163,7 @@ export async function resolveLink(
       const res = await agent.resolveHandle({
         handle: urip.host,
       })
+      // @ts-expect-error TODO new-sdk-migration
```
estrattonbailey, line 166: A handful of these caused by new SDK work too, should really validate these (there's a util) but not gonna do that here

src/lib/hooks/useIntentHandler.ts
estrattonbailey, line 1: No longer used for AA `RedirectOverlay`, that component handles its own state now.

src/lib/strings/url-helpers.ts
```diff
@@ -41,7 +41,8 @@ export function makeRecordUri(
   collection: string,
   rkey: string,
 ) {
-  const urip = new AtUri('at://host/')
+  const urip = new AtUri('at://placeholder.placeholder/')
+  // @ts-expect-error TODO new-sdk-migration
```
estrattonbailey, line 45: Both of these caused by new SDK work

src/state/queries/post-feed.ts
```diff
@@ -141,12 +140,8 @@ export function usePostFeedQuery(
    * available for the remainder of the session, so this delay only affects cold
    * loads. -esb
    */
-  const {isReady: isAgeAssuranceReady} = useAgeAssuranceContext()
   const enabled =
-    opts?.enabled !== false &&
-    Boolean(moderationOpts) &&
-    Boolean(preferences) &&
-    isAgeAssuranceReady
+    opts?.enabled !== false && Boolean(moderationOpts) && Boolean(preferences)
```
estrattonbailey, line 144: Post feeds can load immediately now since we'll have prefs overridden if necessary

src/state/session/index.tsx
```diff
@@ -194,12 +206,15 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         {statsig: true},
       )
       addSessionDebugLog({type: 'method:end', method: 'logout'})
+      clearAgeAssuranceData()
+      // reset onboarding flow on logout
+      onboardingDispatch({type: 'skip'})
```
estrattonbailey, line 211: This is needed to ensure we don't leak info, and to ensure that if a user sees the `NoAccessScreen` after account creation and chooses to log out instead of completing AA, that the next account they log into isn't stuck in onboarding.

src/state/session/index.tsx
```diff
@@ -220,8 +235,12 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         newAccount: account,
       })
       addSessionDebugLog({type: 'method:end', method: 'resumeSession', account})
+      if (isSwitchingAccounts) {
+        // reset onboarding flow on switch account
+        onboardingDispatch({type: 'skip'})
+      }
```
estrattonbailey, line 241: This may actually fix the longstanding web issue of users wiping their prefs due to getting stuck back in onboarding 😅

src/state/shell/index.tsx
```diff
@@ -23,9 +22,7 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
         <DrawerSwipableProvider>
           <MinimalModeProvider>
             <ColorModeProvider>
-              <OnboardingProvider>
```
estrattonbailey, line 26: Need to hoist this out above the `SessionProvider`, it's now in the root `App.tsx` files 🙃

src/view/shell/index.tsx
```diff
@@ -213,13 +214,15 @@ export function Shell() {
...
-        <BlockedGeoOverlay />
+      {aa.state.access === aa.Access.None ? (
+        <NoAccessScreen />
       ) : (
         <RoutesContainer>
           <ShellInner />
         </RoutesContainer>
       )}
+
+      <RedirectOverlay />
```
estrattonbailey, line 225: As mentioned previously, hoisted this up a level so it blocks everything while the status is being verified.

src/App.web.tsx
```diff
@@ -174,44 +174,45 @@ function App() {
...
-      ensureGeolocationConfigIsResolved(),
-    ]).then(() => setReady(true))
+    Promise.all([initPersistedState(), Geo.resolve()]).then(() =>
+      setReady(true),
+    )
   }, [])
 
   if (!isReady) {
-    return null
+    return <Splash isReady />
```
estrattonbailey, line 183: New splash, looks a lot better, still not perfect but can figure that out another time

src/components/Link.tsx
mozzius: we could/should use `e.defaultPrevented`? returning `false` is a little subtle for my liking

package.json
estrattonbailey: Had some type issues, versions need to match evidently so they resolve to the same QueryClient impl. Figured 17 minor versions was risky enough. Their changelogs apparently didn't exist until ~5.90, and GitHub releases are such ass that I haven't even checked for any glaring issues, but our usage is pretty simple and everything seems to work fine.

src/components/Link.tsx
estrattonbailey: Can't recall why I did that originally, tbh with you. But it's what we've been using, should investigate and fix later imo

src/components/ageAssurance/AgeAssuranceAdmonition.tsx
```diff
@@ -2,25 +2,23 @@ import {View} from 'react-native'
...
   style,
 }: ViewStyleProp & {children: React.ReactNode}) {
   const control = useDialogControl()
-  const {isReady, isDeclaredUnderage, isAgeRestricted} = useAgeAssurance()
+  const aa = useAgeAssurance()
 
-  if (!isReady) return null
-  if (isDeclaredUnderage) return null
-  if (!isAgeRestricted) return null
+  if (aa.state.access === aa.Access.Full) return null
```
mozzius, line 21: so am I correct in understanding that we no long have "isDeclaredUnderage" concept, and instead this is an internal detail of the overall access?

package.json
mozzius: Sounds good, I didn't realise we were so far behind. I'll see about getting that updated at some point

src/Splash.web.tsx
mozzius: There's a bit of flash from dark->dim mode, but I'm sure we can figure that out later

src/components/Link.tsx
mozzius: sg, not a big deal, just a minor preference

src/components/ageAssurance/AgeAssuranceAdmonition.tsx
estrattonbailey: Exactly correct

src/ageAssurance/types.ts
```diff
@@ -0,0 +1,65 @@
...
+      break
+    case 'blocked':
+      status = AgeAssuranceStatus.Blocked
+      break
+    default:
+      logger.error(`parseStatusFromString: unknown status value: ${raw}`)
+      status = AgeAssuranceStatus.Unknown
+  }
+  return status
+}
```
mozzius, line 37: a bit neater maybe?

```suggestion
export function parseStatusFromString(raw: string) {
  switch (raw) {
    case 'unknown':
      return AgeAssuranceStatus.Unknown
    case 'pending':
      return AgeAssuranceStatus.Pending
    case 'assured':
      return AgeAssuranceStatus.Assured
    case 'blocked':
      return AgeAssuranceStatus.Blocked
    default:
      logger.error(`parseStatusFromString: unknown status value: ${raw}`)
      return AgeAssuranceStatus.Unknown
  }
}
```

src/Splash.web.tsx
estrattonbailey: Yeah should cache that or something, didn't want to eff with that on this PR, but could pry squeeze it in today/tomorrow

src/ageAssurance/types.ts
```diff
@@ -0,0 +1,65 @@
...
+      break
+    case 'full':
+      access = AgeAssuranceAccess.Full
+      break
+    default:
+      logger.error(`parseAccessFromString: unknown access value: ${raw}`)
+      access = AgeAssuranceAccess.Full
+  }
+  return access
+}
```
mozzius:
```suggestion
export function parseAccessFromString(raw: string) {
  switch (raw) {
    case 'unknown':
      return AgeAssuranceAccess.Unknown
    case 'none':
      return AgeAssuranceAccess.None
    case 'safe':
      return AgeAssuranceAccess.Safe
    case 'full':
      return AgeAssuranceAccess.Full
    default:
      logger.error(`parseAccessFromString: unknown access value: ${raw}`)
      return AgeAssuranceAccess.Full
  }
}
```

src/geolocation/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const geolocationService = useGeolocationServiceResponse()
+  const [deviceGeolocation, setDeviceGeolocation] = useStorage(device, [
+    'deviceGeolocation',
+  ])
+  const geolocation = useMemo(() => {
+    return mergeGeolocations(deviceGeolocation, geolocationService)
+  }, [deviceGeolocation, geolocationService])
+
+  useEffect(() => {
+    device.set(['mergedGeolocation'], geolocation)
```
mozzius, line 52: a little odd. could you explain this? Looks like it's saving the resolved value to a cache, then reading from it in a future session?

estrattonbailey: I'll add a comment. This saves the combined geo service + device GPS location data so we can read it elsewhere out of band of React. Needs to be cached so that it can be read during the `prefetch*` calls, which happen prior to other stuff being ready.

mozzius: Ok that makes sense. So it's for initialising future sessions?

src/ageAssurance/components/RedirectOverlay.tsx
mozzius: Makes sense

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
-            <Trans>My Birthday</Trans>
-          </Text>
-          <Text style={[a.leading_snug, t.atoms.text_contrast_medium]}>
-            <Trans>
-              This information is private and not shared with other users.
-            </Trans>
-          </Text>
+      {isBirthdateUpdateAllowed ? (
+        <Dialog.ScrollableInner
+          label={_(msg`My Birthday`)}
```
surfdude29:
```suggestion
          label={_(msg`My Birthdate`)}
```
nit: `Birthday` ➡️ `Birthdate`

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
+          style={web({maxWidth: 400})}>
+          <View style={[a.gap_md]}>
+            <Text style={[a.text_xl, a.font_semi_bold]}>
+              <Trans>My Birthday</Trans>
```
surfdude29:
```suggestion
              <Trans>My Birthdate</Trans>
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
+            </Text>
+
+            {isLoading ? (
+              <Loader size="xl" />
+            ) : error || !preferences ? (
+              <ErrorMessage
+                message={
+                  error?.toString() ||
+                  _(
+                    msg`We were unable to load your birth date preferences. Please try again.`,
```
surfdude29:
```suggestion
                    msg`We were unable to load your birthdate preferences. Please try again.`,
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
-            />
-          ) : (
-            <BirthdayInner control={control} preferences={preferences} />
-          )}
-        </View>
+          <Dialog.Close />
+        </Dialog.ScrollableInner>
+      ) : (
+        <Dialog.ScrollableInner
+          label={_(msg`You recently changed your birthday`)}
```
surfdude29:
```suggestion
          label={_(msg`You recently changed your birthdate`)}
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
+          style={web({maxWidth: 400})}>
+          <View style={[a.gap_sm]}>
+            <Text
+              style={[
+                a.text_xl,
+                a.font_semi_bold,
+                a.leading_snug,
+                {paddingRight: 32},
+              ]}>
+              <Trans>You recently changed your birthday</Trans>
```
surfdude29:
```suggestion
              <Trans>You recently changed your birthdate</Trans>
```
nit: as above

src/components/dialogs/BirthDateSettings.tsx
```diff
@@ -30,42 +33,71 @@ export function BirthDateSettingsDialog({
...
+            </Text>
+            <Text
+              style={[a.text_md, a.leading_snug, t.atoms.text_contrast_medium]}>
+              <Trans>
+                There is a limit to how often you can change your birthday. You
```
surfdude29:
```suggestion
                There is a limit to how often you can change your birthdate. You
```
nit: as above

src/state/session/index.tsx
mozzius: We _really_ need to scope this by DID lol, this feels really sketchy

src/view/shell/index.web.tsx
```diff
@@ -139,16 +139,19 @@ function ShellInner() {
 
 export function Shell() {
   const t = useTheme()
-  const {status: geolocation} = useGeolocationStatus()
+  const aa = useAgeAssurance()
+  useIntentHandler()
```
mozzius: Seems ok, I'm suprised it doesn't actually need nav context

src/geolocation/service.ts
```diff
@@ -0,0 +1,137 @@
...
+          },
+        )
+
+        // retry 3 times, but don't await, proceed with default
+        networkRetry(3, () =>
+          fetchGeolocationServiceData(GEOLOCATION_SERVICE_URL),
+        )
+          .then(config => {
+            cacheResponseOrThrow(config)
+            success = true
```
mozzius: this is pointless as it will have already resolved

src/state/shell/index.tsx
mozzius: :/ this does feel a little sketchy, I don't love messing with the order of this. That said, theoretically it should be fine

mozzius: Actually sorry yeah totally fine, it's not the actual components being moved, just the state provider. All good 👍

mozzius (approved):
as a great man once said: 💎

src/view/shell/index.web.tsx
estrattonbailey: This can actually go back to where it was now. The native file has it here, but web had it in the inner shell component. I don't think it matters, but can't hurt to let sleeping dogs lie.

src/geolocation/service.ts
estrattonbailey: heh yep

src/view/shell/index.web.tsx
mozzius: that would be preferable!

src/components/dialogs/BirthDateSettings.tsx
estrattonbailey: Sure might as well, thanks

PR #9480, opened 2025-12-04
Add `loading="lazy"` to expo-image on web

mozzius:
Supercedes #9477 by patching the prop in directly. TODO: upstream

Screenshot 2025-12-04 at 12 42 44

Added `loading="lazy"` to:
- Image embeds
- Link embed thumbnails
- Avatars

The last couple commits are mostly cleanup, removing indirection, and moving to the new directory structure

# Before

https://github.com/user-attachments/assets/9978dcca-265c-4500-838f-8c27f10059e0

# After

https://github.com/user-attachments/assets/836255ef-2d95-4336-a8df-c5c1606278bc

### DIFF
--- a/src/view/com/util/UserAvatar.tsx
+++ b/src/view/com/util/UserAvatar.tsx
@@ -310,7 +310,6 @@ let UserAvatar = ({
           }}
           blurRadius={moderation?.blur ? BLUR_AMOUNT : 0}
           onLoad={onLoad}
-          loading="lazy"
         />
       )}
       {!noBorder && <MediaInsetBorder style={borderStyle} />}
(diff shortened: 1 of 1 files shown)


src/view/com/util/images/Image.web.tsx
mozzius, line 1: This looks like a hold-over from when we used `react-native-fast-image` and thus needed platform splitting. we can just use expo-image for both now, and thus I just went and removed the indirection. It didn't even use the priority prop lol

mozzius: We *can* change the priority fwiw

estrattonbailey (approved):
Nice, LGTM. Left the avatars eager for now.

PR #9481, opened 2025-12-04
Redo web lightbox, add animations and keyboard a11y

mozzius:
Redesigned web lightbox to use new components, add Radix stuff so that it's keyboard accessible, and added nice animations.

I also made the loading process nicer. Now, it waits for the image to load before playing the entry animation (ensuring that the animation looks nice, with the image all loaded) - and if loading takes more than a second, we show a spinner.

https://github.com/user-attachments/assets/e4a41fea-750a-4f02-a2e1-841d1c93af99

# Test plan

Test:
- Light/dark mode
- Using it with a keyboard (note the current one doesn't use a focus scope at all)
- Highlighting text making sure we don't regress https://github.com/bluesky-social/social-app/pull/1571
- Slow internet behaviour
  - With and without multiple images

### DIFF
--- a/src/view/com/lightbox/Lightbox.web.tsx
+++ b/src/view/com/lightbox/Lightbox.web.tsx
@@ -155,7 +155,7 @@ function LightboxGallery({
             ]}
             hoverStyle={styles.blurredBackdropHover}
             color="secondary"
-            label={_('Previous image')}
+            label={_(msg`Previous image`)}
             shape="round"
             size={gtPhone ? 'large' : 'small'}>
             <ChevronLeftIcon
@@ -176,7 +176,7 @@ function LightboxGallery({
             ]}
             hoverStyle={styles.blurredBackdropHover}
             color="secondary"
-            label={_('Next image')}
+            label={_(msg`Next image`)}
             shape="round"
             size={gtPhone ? 'large' : 'small'}>
             <ChevronRightIcon
@@ -216,7 +216,7 @@ function LightboxGallery({
         ]}
         hoverStyle={styles.blurredBackdropHover}
         color="secondary"
-        label={_('Close lightbox')}
+        label={_(msg`Close lightbox`)}
         shape="round"
         size={gtPhone ? 'large' : 'small'}>
         <XIcon size={gtPhone ? 'md' : 'sm'} style={{color: t.palette.white}} />


src/view/com/lightbox/Lightbox.web.tsx
```diff
@@ -92,95 +125,69 @@ function LightboxInner({
...
+            style={[
+              a.absolute,
+              styles.leftBtn,
+              styles.blurredBackdrop,
+              a.transition_color,
+              delayedFadeInAnim,
+            ]}
+            hoverStyle={styles.blurredBackdropHover}
+            color="secondary"
+            label={_('Previous image')}
```
vineyardbovines: do we need lingui here?

src/view/com/lightbox/Lightbox.web.tsx
```diff
@@ -92,95 +125,69 @@ function LightboxInner({
...
+            style={[
+              a.absolute,
+              styles.rightBtn,
+              styles.blurredBackdrop,
+              a.transition_color,
+              delayedFadeInAnim,
+            ]}
+            hoverStyle={styles.blurredBackdropHover}
+            color="secondary"
+            label={_('Next image')}
```
vineyardbovines: (same lingui question here)

src/view/com/lightbox/Lightbox.web.tsx
```diff
@@ -190,46 +197,133 @@ function LightboxInner({
...
+        style={[
+          a.absolute,
+          styles.closeBtn,
+          styles.blurredBackdrop,
+          a.transition_color,
+          delayedFadeInAnim,
+        ]}
+        hoverStyle={styles.blurredBackdropHover}
+        color="secondary"
+        label={_('Close lightbox')}
```
vineyardbovines: (same lingui question here)

vineyardbovines:
CleanShot 2026-01-28 at 09 53 00
CleanShot 2026-01-28 at 10 04 09
CleanShot 2026-01-28 at 09 57 05

matches on my end!

src/view/com/lightbox/Lightbox.web.tsx
mozzius: Oops nice catch! Annoying that it doesn't give a type error here tbh 🤔

PR #9482, opened 2025-12-04
Clicking "see more" on user suggestions opens suggested users modal

abenzer:


src/components/ProgressGuide/FollowDialog.tsx
```diff
@@ -84,11 +84,26 @@ export function FollowDialog({guide}: {guide: Follow10ProgressGuide}) {
...
+  return {
+    control,
+    Dialog: ({guide}: {guide?: Follow10ProgressGuide}) => (
+      <Dialog.Outer control={control} nativeOptions={{minHeight}}>
+        <Dialog.Handle />
+        <DialogInner guide={guide} />
+      </Dialog.Outer>
+    ),
+  }
+}
```
mozzius: Hmm, not sure I'm a fan of bundling the dialog into the hook like this. or at least, we don't typically do this, it's an interesting idea

src/components/ProgressGuide/FollowDialog.tsx
```diff
@@ -84,11 +84,26 @@ export function FollowDialog({guide}: {guide: Follow10ProgressGuide}) {
...
 }
 
+export function useFollowDialogControl() {
+  const control = Dialog.useDialogControl()
+  const {height: minHeight} = useWindowDimensions()
+
+  return {
+    control,
+    Dialog: ({guide}: {guide?: Follow10ProgressGuide}) => (
+      <Dialog.Outer control={control} nativeOptions={{minHeight}}>
```
estrattonbailey: @mozzius we shouldn't need to force this anymore, right?

mozzius: On the flatlists, no, but I really need to get around to adding a prop to just force it to be full height because this workaround is ridiculous

PR #9485, opened 2025-12-05
Fix authentication when signing in with a DID

rolandcrosby:
Currently, if you enter a DID as the identifier in the login form, you won't be able to authenticate, because the app adds `.bsky.social` (or other configured domain) to any identifier that doesn't contain `.` or `@`. With this change, identifiers starting with `did:` will also be left alone.


ThisIsMissEm:
@matthieusieben this seems reasonable to me, but not exactly a normal user behaviour (since normal users almost certainly won't know their did.

rolandcrosby:
dunno what to make of that CI error, I didn't touch `yarn.lock`... rebasing on latest main to see if that fixes it

PR #9486, opened 2025-12-05
[Contacts] Contacts matching flow

mozzius:
# Stacked on #9473 

# [Read the RFC for this work here](https://docs.bsky.app/blog/contact-import-rfc)

PR Stack:

- [x] https://github.com/bluesky-social/social-app/pull/9486
  - [x] https://github.com/bluesky-social/social-app/pull/9515
  - [x] https://github.com/bluesky-social/social-app/pull/9487
  - [x] https://github.com/bluesky-social/social-app/pull/9489
    - [x] https://github.com/bluesky-social/social-app/pull/9491
  - [x] https://github.com/bluesky-social/social-app/pull/9519

## This PR

This PR adds the basic flow accessible via settings, with mocked data. Stacked on this are two branches: one which integrates with the API (WIP, requires dev-env and running rolodex locally), and one which integrates this flow into onboarding

PR #9487, opened 2025-12-05
[Contacts] API integration

mozzius:
# Stacked on #9486


src/components/contacts/contacts.ts
```diff
@@ -0,0 +1,73 @@
...
+import {type CountryCode} from '#/lib/international-telephone-codes'
+import {normalizePhoneNumber} from './phone-number'
+import {type Contact} from './state'
+
+/**
+ * Takes the raw contact book and returns a plain list of numbers in E.164 format, along
+ * with a mapping to retrieve the contact ID when we get the results back.
+ *
+ * `countryCode` is used as a fallback for local numbers that don't have a country code associated with them.
+ * I'm making the assumption that most local numbers in someone's phone book will be the same as theirs.
```
estrattonbailey, line 12: Is there an alternative to this, perhaps using libphonenumber to infer the country? Just wondering if we encounter false positives, if there we have a plan b

src/components/contacts/contacts.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const normalized = normalizePhoneNumber(
+        rawNumber,
+        number.countryCode,
+        countryCode,
+      )
+      if (normalized === null) continue
+
+      // skip if it's your own number
+      if (normalized === ownNumber) continue
```
estrattonbailey, line 45: Just double checking if this doesn't throw off the index later down the line

src/components/contacts/phone-number.ts
```diff
@@ -89,3 +123,37 @@ export function getCountryCodeFromPastedNumber(
...
+): string | null {
+  try {
+    const result = parsePhoneNumber(rawNumber, {
+      defaultCountry:
+        countryCode && isSupportedCountry(countryCode)
+          ? countryCode
+          : fallbackCountryCode,
+    })
+
+    if (!result.isValid()) return null
```
estrattonbailey, line 143: Oh I see, so if it fails validation then the fallback country code was pry incorrect. Seems reasonable to me

src/screens/Settings/FindContactsSettings.tsx
```diff
@@ -50,16 +62,27 @@ export function FindContactsSettingsScreen({}: Props) {
         <Layout.Header.Slot />
       </Layout.Header.Outer>
       {isNative ? (
-        !hasInitiated ? (
-          <Intro />
+        data ? (
+          !data.syncStatus ? (
```
estrattonbailey, line 66: If it resolves without an error, will `data` be truthy? Usually we check `isLoading` and `error` first, and the final ternary means it resolved.

estrattonbailey: Ok yeah looks like an empty object 👍

src/components/contacts/contacts.ts
mozzius: yeah so we're using libphonenumber inside `normalizePhoneNumber`. it will infer the country from the number if possible - but if it's not, we can give a fallback number. that's what I've (tried) to do here

src/components/contacts/contacts.ts
mozzius: should be good - the index is based on the "hits", not off of every number

src/components/contacts/phone-number.ts
mozzius: Yeah or if for whatever reason we get something that isn't a number back

PR #9491, opened 2025-12-05
[Contacts] More onboarding changes

mozzius:
# Stacked on #9486 

#9489 was about adding the new steps, this one is about making style tweaks to the onboarding as a whole so the old screens match the new ones. The stepper progress bar has been removed.

I changed web to have content centered, which I think looks a little nicer

https://github.com/user-attachments/assets/8fb6b097-a598-4d18-8eb3-b2bdc1248266

### DIFF
--- a/src/components/contacts/screens/PhoneInput.tsx
+++ b/src/components/contacts/screens/PhoneInput.tsx
@@ -13,6 +13,7 @@ import {
 import {cleanError, isNetworkError} from '#/lib/strings/errors'
 import {logger} from '#/logger'
 import {useAgent} from '#/state/session'
+import {OnboardingPosition} from '#/screens/Onboarding/Layout'
 import {
   android,
   atoms as a,
@@ -21,7 +22,6 @@ import {
   useGutters,
   useTheme,
 } from '#/alf'
-import {OnboardingPosition} from '#/screens/Onboarding/Layout'
 import {Button, ButtonIcon, ButtonText} from '#/components/Button'
 import * as TextField from '#/components/forms/TextField'
 import {InternationalPhoneCodeSelect} from '#/components/InternationalPhoneCodeSelect'


src/screens/Onboarding/Layout.tsx
```diff
@@ -59,119 +59,85 @@ export function Layout({children}: React.PropsWithChildren<{}>) {
       accessibilityLabel={dialogLabel}
       accessibilityHint={_(msg`Customizes your Bluesky experience`)}
       style={[isWeb ? a.fixed : a.absolute, a.inset_0, a.flex_1, t.atoms.bg]}>
-      {__DEV__ && (
```
estrattonbailey, line 62: Oops, can we keep this?

mozzius: It's just below - can I use `IS_INTERNAL` instead?

estrattonbailey: Oh it exists on mobile web only, not desktop. Maybe just tweak that then

src/screens/Onboarding/StepInterests/index.tsx
```diff
@@ -93,7 +90,7 @@ export function StepInterests() {
           <ButtonText>
             <Trans>Continue</Trans>
           </ButtonText>
-          <ButtonIcon icon={saving ? Loader : ChevronRight} />
+          {saving && <ButtonIcon icon={Loader} />}
```
estrattonbailey, line 93: Why no chevron? Nice to not have the button shift imo

estrattonbailey (approved):
LGTM

src/screens/Onboarding/Layout.tsx
mozzius: Ohhh yeah obviously, can fix

src/screens/Onboarding/StepInterests/index.tsx
mozzius: Designs don't have one. Agree though, we should raise with Darrin

PR #9497, opened 2025-12-06
Restructure notification/linking handling to use synchronous APIs

mozzius:
Switches notification and linking handling from async to synchronous APIs:

- `Linking` from `react-native` → `expo-linking` with `useLinkingURL()`
- `getLastNotificationResponseAsync()` → `getLastNotificationResponse()` (synchronous)
- Dedup via `clearLastNotificationResponse()` instead of a module-level date variable

This makes the notification entry handler synchronous and simpler, and is a prerequisite for #2 (nav state persistence) which needs these values available synchronously at render time.

## Test plan

- Ensure opening the app via a push notification still navigates to the correct screen
- Ensure opening the app via a deep link still works
- Ensure chat message notifications still trigger account switch when needed


mozzius:
@github-actions ota

M21H:
I feel like there should be an option for users to toggle this feature🤔

PR #9498, opened 2025-12-06
Save current composer draft in localstorage

abenzer:
Saves anything the user has entered into the composer to localstorage, including text, images, and threaded posts.

### DIFF
--- a/src/view/com/composer/useComposerDraft.ts
+++ b/src/view/com/composer/useComposerDraft.ts
@@ -1,385 +1,59 @@
-import {useCallback, useEffect, useMemo, useRef} from 'react'
-import {RichText} from '@atproto/api'
+import {useCallback, useMemo} from 'react'
 
-import {type SelfLabel} from '#/lib/moderation'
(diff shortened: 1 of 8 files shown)


src/view/com/composer/useComposerDraft.ts
```diff
@@ -0,0 +1,301 @@
...
+            type: 'gif',
+            gif: post.embed.gif,
+            alt: post.embed.gif.alt,
+          }
+        }
+
+        return {
+          id: post.id,
+          richtext: new RichText({text: post.text}),
+          shortenedGraphemeLength: post.text.length, // Will be recalculated
```
mozzius:
```suggestion
        const rt = new RichText({text: post.text})
        return {
          id: post.id,
          richtext: rt,
          shortenedGraphemeLength: rt.graphemeLength,
```

mozzius:
Note: localstorage is only available on web. We'll have to use MMKV on native - see `#/storage`

mozzius:
Superceded by #9691 :)

PR #9499, opened 2025-12-06
Add nicer loading screen while OTA is loading

mozzius:
`expo-updates` lets you specify an image etc. I tried to replicate the splash screen to make that transition more seamless


mozzius:
@github-actions ota

PR #9502, opened 2025-12-06
Fixed an error causing the Lists tab to show feeds instead

RetroSunstar:
In #8969, all instances of lists were swapped with feeds instead, this just patches those back to a working state 🐙


surfdude29:
Can confirm, I see this bug on main.

This seems like a serious issue and 1.111 is due for release soon (although hopefully this can be fixed in an OTA?), so I hope the ping is OK: @estrattonbailey @mozzius @abenzer

mozzius (approved):
Sigh. Thank you.

PR #9515, opened 2025-12-10
[Contacts] NUXes

mozzius:
# Stacked on #9486 

Screenshot 2025-12-10 at 15 14 38

https://github.com/user-attachments/assets/3b21c3e7-6c78-48f2-b611-ae34affd89c3

### DIFF
--- a/src/components/dialogs/nuxs/index.tsx
+++ b/src/components/dialogs/nuxs/index.tsx
@@ -44,9 +44,9 @@ const queuedNuxs: {
   {
     id: Nux.FindContactsAnnouncement,
     enabled: ({currentProfile}) => {
-      return isExistingUserAsOf(
-        '2025-12-16T00:00:00.000Z',
-        currentProfile.createdAt,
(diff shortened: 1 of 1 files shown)


src/components/dialogs/nuxs/index.tsx
```diff
@@ -34,24 +42,24 @@ const queuedNuxs: {
   }) => boolean
 }[] = [
   {
-    id: Nux.BookmarksAnnouncement,
+    id: Nux.FindContactsAnnouncement,
     enabled: ({currentProfile}) => {
       return isExistingUserAsOf(
-        '2025-09-08T00:00:00.000Z',
+        '2025-12-16T00:00:00.000Z',
```
estrattonbailey: Should move the `isNative` check here, otherwise it's gonna technically activate on web even if it doesn't render.

mozzius: Done

PR #9519, opened 2025-12-10
[Contacts] Add `contact-match` notification type to feed

mozzius:
# Stacked on #9486 

Copy TBD

Simulator Screenshot - iPhone 17 - 2025-12-10 at 18 40 10

### DIFF
--- a/src/view/com/notifications/NotificationFeedItem.tsx
+++ b/src/view/com/notifications/NotificationFeedItem.tsx
@@ -814,6 +814,7 @@ function FollowBackButton({profile}: {profile: AppBskyActorDefs.ProfileView}) {
   }
 
   const isFollowing = profileShadow.viewer.following
+  const isFollowedBy = profileShadow.viewer.followedBy
   const followingLabel = _(
     msg({
       message: 'Following',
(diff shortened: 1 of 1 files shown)


src/view/com/notifications/NotificationFeedItem.tsx
```diff
@@ -670,7 +673,9 @@ let NotificationFeedItem = ({
                 </TimeElapsed>
               </Text>
             </ExpandListPressable>
-            {item.type === 'follow' && !hasMultipleAuthors && !isFollowBack ? (
+            {(item.type === 'follow' && !hasMultipleAuthors && !isFollowBack) ||
+            (item.type === 'contact-match' &&
+              !item.notification.author.viewer?.following) ? (
```
estrattonbailey, line 678: Do we need to check if the new user is following you yet? Or is that what triggers the notification type?

estrattonbailey (approved):

src/view/com/notifications/NotificationFeedItem.tsx
mozzius: Good catch, updated the button to allow both states

PR #9522, opened 2025-12-10
[AAv2] Init dialog copy update

estrattonbailey:


src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -202,11 +202,11 @@ function Inner() {
                     style={[a.text_sm, a.leading_snug]}>
                     KWS
                   </SimpleInlineLinkText>{' '}
-                  to verify that you’re an adult. When you click "Begin" below,
-                  KWS will check if you have previously verified your age using
-                  this email address for other games/services powered by KWS
-                  technology. If not, KWS will email you instructions for
-                  verifying your age. When you’re done, you'll be brought back
+                  to handle age verification. When you click "Begin" below, KWS
```
surfdude29, line 205: nit: `age assurance` is – I believe – used in all other strings referring to the process, so it might make sense to use it here?

although admittedly `verify your age` works fine later on in the string ¯\_(ツ)_/¯

estrattonbailey: Thanks, we're gonna roll with it

PR #9526, opened 2025-12-11
Replace `graphemer` with `unicode-segmenter`

mozzius:
We're using `unicode-segmenter` in @atproto/api now, so no need to have two libraries doing the same thing.

We really should consider using this library to its full potential - it returns iterators, and there's a built-in way to tell if each segment is an emoji.

`graphemer` was also a transitive dep from eslint 😖 gross

Note: yarn.lock lint *will* fail - intentional, yarn was refusing to dedupe them so we had a dual package thing going on. Fixed via `yarn upgrade`

### DIFF
--- a/package.json
+++ b/package.json
@@ -222,7 +222,7 @@
     "tippy.js": "^6.3.7",
     "tlds": "^1.234.0",
     "tldts": "^6.1.46",
-    "unicode-segmenter": "^0.14.1",
+    "unicode-segmenter": "^0.14.5",
     "zod": "^3.20.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


src/screens/Profile/Header/EditProfileDialog.tsx
```diff
@@ -203,11 +203,11 @@ function DialogInner({
     _,
   ])
 
-  const displayNameTooLong = useWarnMaxGraphemeCount({
+  const displayNameTooLong = isOverMaxGraphemeCount({
```
cometkim, line 206: Unicode segmentation is still expensive job even with the faster library. It would be better to memoize its result unless Bluesky is using React compiler.

mozzius: We are using react compiler :)

cometkim:
Please update unicode-segmenter to v0.14.3+

I just reduced the size significantly :sunglasses: 

https://github.com/cometkim/unicode-segmenter/releases/tag/unicode-segmenter%400.14.3

mozzius:
Yooo nice one 😎

PR #9527, opened 2025-12-11
Lazy load `psl`

mozzius:
Should reduce bundle size by a good few kb

We only use it in the link warning component to split the apex domain from the subdomain. I made a fallback component that just doesn't do that part, for while it's loading


mozzius:
lol wtf bundle size went *up*!

estrattonbailey (changes requested):
Ah crap, I recently used this for liveNow stuff, we'll need to rebase and handle that case too if we wanna do this

mozzius:
Sure np

mozzius:
Ah, live now seems like a lot harder to lazyload. Putting into draft for now

PR #9528, opened 2025-12-11
Remove RNGH from web bundle

mozzius:
We don't use it at all on web, but there were a couple of stray imports that kept it in the bundle. Easy enough to fix.

As a happy side effect it also fixes https://bsky.app/profile/goose.art/post/3m7hfnd7f7k2s


gaearon:
maybe you could add a nonsensial `alias` or some explicit throw for it to the webpack config so that the build fails if it’s ever accidentally readded

mozzius:
Good idea!

PR #9532, opened 2025-12-11
Force resolution of ESM version of `unicode-segmenter` on web

mozzius:
The fallback for old browsers that don't have Intl.Segmenter wasn't working, as it was resolving to the `.cjs` version of `unicode-segmenter` which got lost in the build pipeline (it got added as an image lmao). We can force a resolution to the ESM version to fix.

# Before 
Screenshot 2025-12-11 at 14 41 37

# After

Screenshot 2025-12-11 at 14 40 00

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -23,6 +23,10 @@ module.exports = async function (env, argv) {
   config = withAlias(config, {
     'react-native$': 'react-native-web',
     'react-native-webview': 'react-native-web-webview',
+    // Force ESM version
+    'unicode-segmenter/grapheme': require
+      .resolve('unicode-segmenter/grapheme')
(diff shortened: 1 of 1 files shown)


cometkim:
Interesting... I thought I've fixed all resolution issues for most bundlers. Do you think the resolution issue is specific to the Bluesky project? Or do I need to fix something on my package?

mozzius:
I'm really not sure - it could just be us. @expo/webpack-config is unmaintained, our atproto packages are CJS and really need upgrading to ESM, which I hope can happen soon, and the overall build pipeline is quite custom and janky. It could just be an us thing. It'll all be fixed when we upgrade the atproto ones to ESM at least :)

Your exports config does seem to be correct for what it's worth

PR #9538, opened 2025-12-11
[AAv2] Handle deactivated and suspended accounts more deliberately

estrattonbailey:
Currently accounts that should see the AA `NoAccessScreen` who are also deactivated or suspended are handled like this:
1. Deactivated accounts see the AA screen first. Once they get access, they see the `Deactivated` screen. This is OK-ish.
2. Suspended accounts see the `Takendown` screen first because `ageassurance.getState` fails entirely and we can't compute state for the user. This is a bug. Regardless. Once a user is restored, they'll see the AA screen.

This PR makes the handling of these screens more deliberate, so that the order of operations is not left up to failed requests, and makes more sense to a user.

This PR aligns the markup of these full-screen components ([review without whitespace](https://github.com/bluesky-social/social-app/pull/9538/changes?w=1)). And now:
1. Deactivated users see `Deactivated` first, and once re-activated, see the AA screen (if applicable of course)
2. Suspended users see the `Takendown` screen first, and once re-instated, see the AA screen (ditto)

### DIFF
--- a/src/ageAssurance/components/NoAccessScreen.tsx
+++ b/src/ageAssurance/components/NoAccessScreen.tsx
@@ -114,123 +114,127 @@ export function NoAccessScreen() {
 
   return (
     <>
-      <ScrollView
-        contentContainerStyle={[
-          a.px_2xl,
-          {
-            paddingTop: isWeb ? a.p_5xl.padding : insets.top + a.p_2xl.padding,
-            paddingBottom: 100,
-          },
-        ]}>
-        <View
-          style={[
-            a.mx_auto,
-            a.w_full,
-            web({
-              maxWidth: 380,
-              paddingTop: gtPhone ? '8vh' : undefined,
-            }),
+      <View style={[a.util_screen_outer, a.flex_1]}>
+        <ScrollView
+          contentContainerStyle={[
+            a.px_2xl,
             {
-              gap: 32,
+              paddingTop: isWeb
+                ? a.p_5xl.padding
+                : insets.top + a.p_2xl.padding,
+              paddingBottom: 100,
             },
           ]}>
-          <View style={[a.align_start]}>
-            <AgeAssuranceBadge />
-          </View>
+          <View
+            style={[
+              a.mx_auto,
+              a.w_full,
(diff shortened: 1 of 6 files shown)


src/screens/Deactivated.tsx
```diff
@@ -37,19 +35,12 @@ export function Deactivated() {
...
   const agent = useAgent()
   const [pending, setPending] = React.useState(false)
   const [error, setError] = React.useState<string | undefined>()
   const queryClient = useQueryClient()
 
-  useFocusEffect(
-    React.useCallback(() => {
-      setMinimalShellMode(true)
-    }, [setMinimalShellMode]),
-  )
```
estrattonbailey, line 51: Don't need this anymore, there's no shell around it

src/screens/Takendown.tsx
```diff
@@ -130,12 +129,7 @@ export function Takendown() {
   useEnableKeyboardController(true)
 
   return (
-    <Modal
-      visible
-      animationType={native('slide')}
-      presentationStyle="formSheet"
-      style={[web(a.util_screen_outer)]}>
-      {isIOS && <SystemBars style={{statusBar: 'light'}} />}
```
estrattonbailey, line 138: Don't need this anymore either

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -114,123 +114,127 @@ export function NoAccessScreen() {
...
-        ]}>
-        <View
-          style={[
-            a.mx_auto,
-            a.w_full,
-            web({
-              maxWidth: 380,
-              paddingTop: gtPhone ? '8vh' : undefined,
-            }),
+      <View style={[a.util_screen_outer, a.flex_1]}>
```
estrattonbailey, line 117: Matched this for consistency

src/screens/Takendown.tsx
mozzius: Heh yeah this was more or less a stylistic thing, it didn't actually serve any purpose. Fine to remove

PR #9555, opened 2025-12-15
[APP-1684] Some contact import tweaks

estrattonbailey:


src/components/InternationalPhoneCodeSelect.tsx
```diff
@@ -113,5 +114,5 @@ function Flag({unicodeFlag, svgFlag}: {unicodeFlag: string; svgFlag: any}) {
       />
     )
   }
-  return unicodeFlag + ' '
+  return <RNText style={[{lineHeight: 21}]}>{unicodeFlag + ' '}</RNText>
```
estrattonbailey, line 117: Helps alignment on iOS and Android, tested on both platforms

PR #9559, opened 2025-12-16
[bskylink] URL-encode apostrophes before HTML escaping

surfdude29:
> [!WARNING]
> **This bug was diagnosed and the PR and description [put together by Claude](https://claude.ai/share/6c310fc1-f885-4331-8e29-c3f60fe71c01). I have not tested the fix.**

## Bug Summary

I noticed a bug when tapping the Wikipedia link in [this post](https://bsky.app/profile/lasertrapperkeeper.bsky.social/post/3ma2zw6ufls2g): URLs containing unencoded apostrophes (e.g., `They're`) are truncated when redirecting through the `bskylink` service, causing users to be redirected to incomplete URLs.

**EDIT:** I see this bug in the iOS app and in mobile Safari, but @layeredstrange [reported](https://bsky.app/profile/mikejay.link/post/3ma3b2fa7l22x) that he is unable to reproduce it in the Android app.

## Example

**Original URL:**
```
https://en.wikipedia.org/wiki/They're_Coming_to_Take_Me_Away%2C_Ha-Haaa
```

**Actual redirect destination:**
```
https://en.wikipedia.org/wiki/They
```

## Root Cause

The bug is in `src/html/linkRedirectContents.ts`. The code uses `escapeHTML()` to sanitize the URL before embedding it in a meta refresh tag:

```html
<meta http-equiv="refresh" content="0; URL='${escapeHTML(link)}'" />
```

The problem is a **two-stage parsing issue**:

1. **HTML Entity Escaping**: `escapeHTML()` converts `'` to `&#39;`
2. **Browser HTML Parsing**: The browser decodes `&#39;` back to `'` when reading the `content` attribute
3. **Meta Refresh Parsing**: The meta refresh parser then sees the raw apostrophe and interprets it as the closing quote for the URL

### Step-by-step breakdown

| Stage | Value |
|-------|-------|
| Input URL | `https://en.wikipedia.org/wiki/They're_Coming...` |
| After `escapeHTML()` | `https://en.wikipedia.org/wiki/They&#39;re_Coming...` |
| HTML output | `content="0; URL='https://en.wikipedia.org/wiki/They&#39;re_Coming...'"` |
| After browser HTML decode | `content="0; URL='https://en.wikipedia.org/wiki/They're_Coming...'"` |
| Meta refresh sees | URL starts with `'`, ends at next `'` → `https://en.wikipedia.org/wiki/They` |

## The Fix

URL-encode the apostrophe as `%27` **before** HTML escaping. This ensures the meta refresh parser never sees a raw apostrophe:

```typescript
export function linkRedirectContents(link: string): string {
  // Encode apostrophes to prevent breaking out of single-quoted URL
  const safeLink = link.replace(/'/g, '%27')

  return `
    <html>
      <head>
        <meta http-equiv="refresh" content="0; URL='${escapeHTML(safeLink)}'" />
        ...
      </head>
    </html>
  `
}
```

### Result after fix

| Stage | Value |
|-------|-------|
| Input URL | `https://en.wikipedia.org/wiki/They're_Coming...` |
| After `replace()` | `https://en.wikipedia.org/wiki/They%27re_Coming...` |
| HTML output | `content="0; URL='https://en.wikipedia.org/wiki/They%27re_Coming...'"` |
| After browser HTML decode | `content="0; URL='https://en.wikipedia.org/wiki/They%27re_Coming...'"` |
| Meta refresh sees | Full URL with `%27` preserved ✓ |

## Files Changed

- `bskylink/src/html/linkRedirectContents.ts` – Add apostrophe encoding before HTML escaping

## Testing

1. Create a post with a link containing an apostrophe in the path
2. Click the link on Bluesky
3. Verify you're redirected to the correct full URL, not a truncated version

### DIFF
--- a/bskylink/src/html/linkRedirectContents.ts
+++ b/bskylink/src/html/linkRedirectContents.ts
@@ -1,10 +1,20 @@
 import escapeHTML from 'escape-html'
 
 export function linkRedirectContents(link: string): string {
+  // Encode characters that could break out of the single-quoted URL in meta refresh.
+  // HTML entity escaping (&#39;) is insufficient because the browser decodes entities
+  // before the meta refresh parser processes the URL, allowing apostrophes to
+  // prematurely terminate the URL string.
+  //
+  // Example: "They're" with HTML escaping becomes "They&#39;re" in HTML, but after
+  // the browser decodes the content attribute, the meta refresh parser sees "They're"
+  // and interprets the apostrophe as the closing quote, truncating the URL to "They".
+  const safeLink = link.replace(/'/g, '%27')
+
   return `
     <html>
       <head>
-        <meta http-equiv="refresh" content="0; URL='${escapeHTML(link)}'" />
+        <meta http-equiv="refresh" content="0; URL='${escapeHTML(safeLink)}'" />
         <meta
           http-equiv="Cache-Control"
           content="no-store, no-cache, must-revalidate, max-age=0" />


DavidBuchanan314:
LGTM (but I have also not tested)

Relevant bit of HTML spec: https://html.spec.whatwg.org/#shared-declarative-refresh-steps

surfdude29:
I asked a couple of follow-up questions based on the spec, and Claude said that just encoding single-quotes should be fine for the current implementation, but maybe it wouldn't hurt to also encode `"`.

___

**Yes, for the current `bskylink` implementation, only `'` (U+0027) needs to be encoded.**

Here's why:

1. **bskylink uses single quotes**: `URL='${url}'`
2. **The meta refresh parser** sees `URL='` and sets `quote = '`
3. **Step 11.11** only truncates when it finds a character **equal to `quote`**
4. A `"` in the URL won't match `'`, so no truncation occurs

However, for **defence in depth**, you could argue for encoding both:

```typescript
// Minimal fix (sufficient for current implementation)
const safeLink = link.replace(/'/g, '%27')

// Defence in depth (future-proof if template ever changes to double quotes)
const safeLink = link.replace(/'/g, '%27').replace(/"/g, '%22')
```

The second approach would be harmless (both `%27` and `%22` are valid URL percent-encodings) and would protect against someone accidentally changing the template to use double quotes in the future.

But strictly speaking, **only `'` is required** given the current code uses `URL='...'`.

ds-boyce (approved):
Verified this locally via tests. It appears to be functioning as described.

In the generated HTML, I saw this before and after the change in this PR:
```diff
- <meta http-equiv="refresh" content="0; URL='https://en.wikipedia.org/wiki/Singin&#39;_in_the_Rain'" />
+ <meta http-equiv="refresh" content="0; URL='https://en.wikipedia.org/wiki/Singin%27_in_the_Rain'" />
```

The input was `https://en.wikipedia.org/wiki/Singin'_in_the_Rain`.

surfdude29:
@ds-boyce Thanks for reviewing this 🙏

PR #9561, opened 2025-12-16
Fix DNS TXT record host generation for handles with subdomains

0x00evil:
Previously, the DNS TXT record Host field was hardcoded to always display _atproto, regardless of whether the user’s input domain included a subdomain. This caused verification failures for handles using subdomains (e.g., user.example.com), because the correct DNS record must be placed at _atproto.user.example.com, not _atproto.example.com.

This PR fixes the host generation logic to follow the Atproto specification:
- If the user enters a domain with a subdomain (e.g., user.example.com), the Host should be _atproto.user (so the full record becomes _atproto.user.example.com).
- If the user enters an apex domain (e.g., example.com), the Host should be _atproto (resulting in _atproto.example.com).

The change ensures that users receive the correct DNS instruction based on their input, enabling successful domain verification for both apex and subdomain-based handles.


ThisIsMissEm:
You'd probably want to use the public suffix list for this, since ccTLDs often have things like domain.co.nz which would trigger the subdomain logic but then sub.domain.co.nz may not (depending on the split used)

PR #9564, opened 2025-12-16
[AAv2] Integrate on-device APIs

estrattonbailey:
Integrates on-device age APIs from Apple and Google to power Age Assurance verification in regions that support/allow this method of verification.


estrattonbailey:
> device thresholds are hardcoded and decoupled from region rule ages
Not an issue at the moment and can be adjusted in the future if we ever need to.

estrattonbailey:
Gone back and forth in slack a bunch and done a lot of laborious on-device testing, going to say this one is ready

surfdude29:
@estrattonbailey Great work on this! 👌

Because this is the first time many users will have encountered an app asking them to share their age range, I wonder if the team might be open to expanding the first part of the string that's shown when a user has the opportunity to share it?

I feel like more clarity and reassurance that only an age range is being shared – and not any more specific personal data – might be quite helpful for users in deciding whether they even want to tap the **Share age data** button that opens the native age range sharing flow.

Current strings:
https://github.com/bluesky-social/social-app/blob/2eddbf8eb0ab40ed7f8bb551f994e2ecad60cab5/src/components/ageAssurance/AgeAssuranceAccountCard.tsx#L169-L171

https://github.com/bluesky-social/social-app/blob/2eddbf8eb0ab40ed7f8bb551f994e2ecad60cab5/src/ageAssurance/components/NoAccessScreen.tsx#L433-L435

Based on [my experience with the iOS flow and the wording used in that](https://bsky.app/profile/surfdude29.ispost.ing/post/3mgbfqtbul22q), I came up with this suggestion (ideally the strings would be platform-split):

Proposed iOS string:

> Sharing the age range associated with your Apple Account reveals only that, for example, you’re at least 18. `<bold>`Your age and birthday are not shared.`</bold>` Bluesky only applies your age range on the device where you share it. Alternatively, [KWS link sentence unchanged].

Also, I think it would be a good idea to change the button CTA on iOS  to **Share age range** to match the native flow. (As noted below, I don't know what wording is used in the Android flow, the same might be appropriate there or another string might be better.)

Proposed Android string:

> Sharing the age range associated with your Google Account reveals only that, for example, you’re at least 18. `<bold>`Your age and birthday are not shared.`</bold>` Bluesky only applies your age range on the device where you share it. Alternatively, [KWS link sentence unchanged].

(I haven't tested the Android flow, some tweaks may be necessary to more closely align this with the wording used there.)

EDITED: Various tweaks and polish (thanks Claude!)

estrattonbailey:
@surfdude29 hmmm yeah could pry clarify this a bit. I'll have a quick look, good call.

estrattonbailey:
Integrated here https://github.com/bluesky-social/social-app/pull/11106

PR #9571, opened 2025-12-18
[AAv2] Improve instructions on NoAccessScreen and bday dialog

estrattonbailey:
Copy didn't change, just the logic for where it's shown.


surfdude29:
@estrattonbailey Just wondering, would it make sense to add `orgAdmonition` to the birthdate dialog as well? If so, I could try my hand at putting a PR together.

And maybe the signup flow too? Or perhaps that would be off-putting or confusing to users signing up for personal accounts (the vast majority ofc)?

PR #9574, opened 2025-12-18
Temporarily use English for Esperanto `intl-displaynames` polyfill

### LINKED ISSUE #9580: Regression: change language page broken
### Steps to Reproduce

If you go to the "change language" page on Android with not-english, the app will break.

### Attachments

Image

Android

### Device Info

Samsung Galaxy S25 Ultra

1.112.0.774

### Additional Information

Looking at the i18n files, your English i18n was updated 4 days ago, and is 18 lines longer than the others, so there is likely a key that was added, and none of the other i18n files were re-generated.

Future mitigation steps could include checking the i18n file lengths to make sure they match (easy solution) or have the same key and file associations, before releasing to prod.

I guess you could also diff the output of `tail -n+21 <file> | grep -v msgstr`

surfdude29:
> [!IMPORTANT]
> This fix seems logical to me, but I haven't tested it.

In #9473, the `formatjs/intl-displaynames` polyfill was added to provide localization of territory and language display names where `Intl.DisplayNames` isn't supported (especially on native).

This seems to work well for all the app languages we support, with the exception of Esperanto. There seems to be a bug in the [package](https://www.npmjs.com/package/@formatjs/intl-displaynames?activeTab=code), because there isn't actually any data in the `locale-data/eo.js` file:

IMG_9971

So when the app tries to load the polyfill's locale data for Esperanto, it crashes:

IMG_9969

A point to note is that CLDR data is indeed available for both [languages](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-localenames-full/main/eo/languages.json) and [territories](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-localenames-full/main/eo/territories.json) in Esperanto.

Until the bug is fixed in `formatjs/intl-displaynames`, this PR proposes simply using English display names instead.

@szafranek13

Fixes #9580

___

## Test Plan

1. On native, open [Settings > Languages](https://bsky.app/settings/language) and change the **App Language** to **Esperanto**
2. Tap the **Content Languages** selector
3. List should open with language names in English

### DIFF
--- a/.github/workflows/build-submit-android.yml
+++ b/.github/workflows/build-submit-android.yml
@@ -79,13 +79,15 @@ jobs:
           echo "$json" > google-services.json
 
       - name: 🏗️ EAS Build
+        env:
+          PROFILE: ${{ inputs.profile || 'testflight-android' }}
         run: >
           SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
           SENTRY_RELEASE=${{ steps.env.outputs.EXPO_PUBLIC_RELEASE_VERSION }}
           SENTRY_DIST=${{ steps.env.outputs.EXPO_PUBLIC_BUNDLE_IDENTIFIER }}
           yarn use-build-number-with-bump
           eas build -p android
-          --profile ${{ inputs.profile || 'testflight-android' }}
+          --profile $PROFILE
           --local --output build.aab --non-interactive
 
       - name: ✍️ Rename Testflight bundle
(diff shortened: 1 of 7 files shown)


src/locale/i18n.ts
```diff
@@ -138,7 +138,7 @@ export async function dynamicActivate(locale: AppLanguage) {
       await Promise.all([
         import('@formatjs/intl-pluralrules/locale-data/eo'),
         import('@formatjs/intl-numberformat/locale-data/eo'),
-        import('@formatjs/intl-displaynames/locale-data/eo'),
+        import('@formatjs/intl-displaynames/locale-data/en'),
```
mozzius: no need to use the `en` version as it's imported at the top
```suggestion
        // borked, see https://github.com/bluesky-social/social-app/pull/9574
        // import('@formatjs/intl-displaynames/locale-data/eo'),
```

surfdude29: Ah of course, thanks

PR #9577, opened 2025-12-20
Prevent flash of wrong theme on startup (web)

mozzius:
Adds a tiny script to read the theme prefs and sets the correct bg color on the splash screen

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -64,7 +64,7 @@
       background-color: white;
       --text: black;
       --background: white;
-      --backgroundLight: #EFF2F6;
+      --backgroundLight: #DCE2EA;
     }
     html.theme--dark,
     html.theme--dark body,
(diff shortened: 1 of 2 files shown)


src/state/shell/color-mode.tsx
```diff
@@ -74,3 +77,21 @@ export function useThemePrefs() {
...
+      colorMode,
+      darkTheme,
+    }: {
+      colorMode: persisted.Schema['colorMode']
+      darkTheme: persisted.Schema['darkTheme']
+    }) => {
+      React.useEffect(() => {
+        if (window !== undefined) {
+          window.localStorage.setItem('SPLASH_DARK_THEME', darkTheme ?? 'dim')
+          window.localStorage.setItem('SPLASH_COLOR_MODE', colorMode)
```
mozzius: I chose to save it to localstorage separately, just so we have more control over the key. We could just read the main storage object but then we'd have to figure out the key, decode into json, worry about the schema, blah blah blah. Dedicated item seemed the way to go

PR #9578, opened 2025-12-22
chore: fix some typos in comments

xibeiyoumian:
fix some typos in comments


src/lib/hooks/useDraggableScrollView.ts
```diff
@@ -48,7 +48,7 @@ export function useDraggableScroll<Scrollable extends ScrollView = ScrollView>({
         return
       }
 
-      // Require n pixels momement before start of drag (3 in this case )
+      // Require n pixels moment before start of drag (3 in this case )
```
surfdude29, line 51:
```suggestion
      // Require n pixels movement before start of drag (3 in this case )
```
I think this should be `movement`

PR #9579, opened 2025-12-22
Set up first theme overrides in settings folder

hilkeros:


hilkeros:
this was supposed to be in the indie-app fork by northsky...

PR #9594, opened 2025-12-26
Speed up startup by persisting some queries

mozzius:
Persists the preferences query and the feed info query. This lets us skip a stage of the initial startup waterfall, as you can see in this video: we no longer see the circular spinner and skip right to the skeletons

# Before

https://github.com/user-attachments/assets/2e3e787b-27e5-489c-a7e5-8724238dd8fd

# After

https://github.com/user-attachments/assets/70a07999-9585-4c23-9b90-57bcd0d52d30

Two details to be aware of:
- I added [`superjson`](https://www.npmjs.com/package/superjson) in order to serialise/deserialise the dates
- I added a `PERSIST_VERSION` to the persist key as this will be a backwards-incompatible change. bit of a blunt way to do this as it will wipe out the whole cache each time we increment this. open to suggestions for a neater way to do it

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -22,7 +22,11 @@ import {
 import {DISCOVER_FEED_URI, DISCOVER_SAVED_FEED} from '#/lib/constants'
 import {sanitizeDisplayName} from '#/lib/strings/display-names'
 import {sanitizeHandle} from '#/lib/strings/handles'
-import {persist, STALE} from '#/state/queries'
+import {
+  PERSISTED_QUERY_GCTIME,
+  PERSISTED_QUERY_ROOT,
+  STALE,
+} from '#/state/queries'
 import {RQKEY as listQueryKey} from '#/state/queries/list'
 import {usePreferencesQuery} from '#/state/queries/preferences'
 import {useAgent, useSession} from '#/state/session'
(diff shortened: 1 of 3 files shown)


src/lib/react-query.tsx
```diff
@@ -7,14 +7,22 @@ import {
...
 } from '@tanstack/react-query-persist-client'
-import type React from 'react'
+import {SuperJSON} from 'superjson'
 
 import {isNative} from '#/platform/detection'
 import {listenNetworkConfirmed, listenNetworkLost} from '#/state/events'
 
-// any query keys in this array will be persisted to AsyncStorage
+// any query keys in this array will be persisted to AsyncStorage.
+// keys are here for require cycle reasons ¯\_(ツ)_/¯
```
estrattonbailey: :| damn, yeah

Maybe the move is to use a structured key with a prefix that indicates the query should be cached, like I [outlined a while back](https://linear.app/blueskyweb/issue/APP-1010/improve-react-query-setup). Could just start with like `persist` or something.

mozzius: Mmmm, that might be neater?

Or maybe have it as a whole segment: `[PERSIST, 'getPreferences']`

estrattonbailey (approved):

PR #9603, opened 2025-12-27
Cleaner sidebar layout

abenzer:
Took a first pass at a cleaner layout for the sidebar, attempting to make each section more visually distinct. 

Before: 
before

After: 
after

### DIFF
--- a/src/components/ProgressGuide/FollowDialog.tsx
+++ b/src/components/ProgressGuide/FollowDialog.tsx
@@ -31,6 +31,7 @@ import {
 import {Button, ButtonIcon, ButtonText} from '#/components/Button'
 import * as Dialog from '#/components/Dialog'
 import {useInteractionState} from '#/components/hooks/useInteractionState'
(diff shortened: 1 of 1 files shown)


mackuba:
Not bad 🤔 I think I like the way these "boxes" look. Though regarding the Trending one, I think the "Trending" heading is a bit too hard to distinguish from the actual trend links - I'd work a bit more on the font for that label, maybe consider adding back the icon?

abenzer:
Screenshot 2025-12-31 at 3 08 39 PM

Think we're getting closer — @mackuba how do you like this design?

mozzius:
@claude can ya give this a review?

mozzius:
This part is not so pretty, but passable. just noting

Screenshot 2026-01-09 at 20 19 45

mozzius (approved):
Made a couple tweaks but BIG improvement, love it. Looks beautiful

PR #9612, opened 2025-12-30
Lazy load storybook

mozzius:
Couple reasons:
1. It's quite a big file which is never used by normal users. Removing from the main bundle makes sense
2. It includes the international tel code select component, which on web inlines all the svg flags. Lazy loading it mean the flags go into the lazy bundle.


mozzius:
Not sure why the render deploy failed :/

PR #9613, opened 2025-12-30
Optimise `getLikelyType` - replace giant ext map with switch case

mozzius:
Previously, we had a giant file:

```ts
const EXT_MIME_TYPES: Record<string, string> = {
  '123': 'application/vnd.lotus-1-2-3',
  '1km': 'application/vnd.1000minds.decision-model+xml',
  '3dml': 'text/vnd.in3d.3dml',
  '3ds': 'image/x-3ds',
  '3g2': 'video/3gpp2',
  '3gp': 'video/3gpp',
  '3gpp': 'video/3gpp',
  // etc for a thousand lines
```

To figure out if something is probably html, text, image, or video, we got the mime type from the extension then checked to see what the string started with.

This is obviously insanely inefficient as we have all this data in the bundle and then in memory, only for it to be discarded as we only care about the first bit (`image/` etc).

I made claude use metaprogramming to generate a giant switch case to handle this instead. This halves the bundle size of the file, and will be faster to run and less memory-demanding.

# Before

Screenshot 2025-12-30 at 13 24 39

# After

Screenshot 2025-12-30 at 13 24 24

## Test plan

I added unit tests - it compares the old implementation to the new one, and also adds some additional tests

Ensure it passes


mozzius:
I think something is wrong with the bundle size CI thingy - should be 23kb smaller

PR #9614, opened 2025-12-30
Lazy load web composer input

mozzius:
This might be a bit crazy, but the tiptap-based rich text input on web requires some pretty chunky dependencies (3rd biggest vendored dep). We can lazy load them to only require them after the composer is opened for the first time.

Look at the size of the chunk!

Screenshot 2025-12-30 at 16 04 31

You might say "wait a sec but what do they see while it's loading?". My solution is a lightweight plain-text version implemented with a simple `textinput`. This lets you type while it loads, and more or less seamlessly passes over to the proper version once loaded.

Here's with an artificial 2 second delay (placeholder is red, real is blue)

https://github.com/user-attachments/assets/dba05089-6edf-4a3b-8d62-b6d7dec196d3

### DIFF
--- a/src/view/com/composer/text-input/TextInput.web.tsx
+++ b/src/view/com/composer/text-input/TextInput.web.tsx
@@ -1,13 +1,24 @@
-import {lazy, Suspense, useEffect, useState} from 'react'
+import {
+  lazy,
+  Suspense,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
 import {StyleSheet, View} from 'react-native'
 import Animated, {FadeIn, FadeOut} from 'react-native-reanimated'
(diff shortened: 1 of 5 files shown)


src/view/com/composer/text-input/TextInput.web.tsx
```diff
@@ -146,210 +94,12 @@ export function TextInput({
...
-  }, [t, fonts, webForceMinHeight])
-
   return (
     <>
       <View style={[styles.container, hasRightPadding && styles.rightPadding]}>
-        {/* @ts-ignore inputStyle is fine */}
-        <EditorContent editor={editor} style={inputStyle} />
+        <Suspense fallback={<FallbackTextInput {...props} />}>
+          <RichTextInput {...props} />
+        </Suspense>
```
mozzius, line 102: React is so good

PR #9628, opened 2026-01-03
Add 'Clear All' button to search history

darnfish:
Introduces a 'Clear All' button in the search history component, allowing users to clear both term and profile search histories at once. The button is wired to a new callback that resets the relevant state arrays.

Simulator Screenshot - iPhone 16 Pro - 2026-01-03 at 15 36 12


src/screens/Search/components/SearchHistory.tsx
```diff
@@ -44,10 +46,28 @@ export function SearchHistory({
...
+              a.pt_sm,
+              a.flex_row,
+              a.align_center,
+              a.justify_between,
+            ]}>
             <Text style={[a.text_md, a.font_semi_bold]}>
               <Trans>Recent Searches</Trans>
             </Text>
+            <Button
+              label={_(msg`Clear all`)}
```
surfdude29:
```suggestion
              label={_(msg`Clear All`)}
```
nit: It would be good to use the same capitalization here, to reduce string duplication for translators :)

PR #9631, opened 2026-01-04
feat(settings): add search bar to filter settings

ezramechaber:
Adds a search input to the Settings screen that filters settings in real-time, similar to X or iPhone settings. 

Users can search by setting title or keywords (actual UI labels from settings screens). Requires 2+ characters to trigger filtering. Shows matched keywords below results.

- Add settingsSearchConfig.ts with searchable settings categories
- Add useSettingsSearch hook for filtering logic
- Add SettingsSearchEmpty component for no-results state
- Add SettingsSearchResultItem component for search results
- Integrate search bar into Settings.tsx

One note on maintainability: This does require a static list of all the search items. It's a simpler implementation but if settings change, this would add overhead for a dev to update search.

### DIFF
--- a/src/locale/locales/en/messages.po
+++ b/src/locale/locales/en/messages.po
@@ -574,10 +574,11 @@ msgstr ""
 msgid "A screenshot of a profile page with a bell icon next to the follow button, indicating the new activity notifications feature."
 msgstr ""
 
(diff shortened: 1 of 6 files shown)


ezramechaber:
CleanShot 2026-01-04 at 17 54 13

PR #9635, opened 2026-01-05
implement collapsible tree

turbocrime:
Add collapsible thread branches. Tap the minus icon on a reply line to collapse, plus icon to expand.

Collapse state lives in the `PostThread` component.

State is not cleared by app navigation, but won't apply anywhere but the original view anchor.

State is reset on refresh or browser navigation.

https://github.com/user-attachments/assets/2e88ae89-7cfc-469c-ae71-2d05fc6482d7

### DIFF
--- a/src/components/icons/CircleMinus.tsx
+++ b/src/components/icons/CircleMinus.tsx
@@ -1,0 +1,5 @@
+import {createSinglePathSVG} from './TEMPLATE'
+
+export const CircleMinus_Stroke2_Corner0_Rounded = createSinglePathSVG({
+  path: 'M12 2c5.523 0 10 4.477 10 10s-4.477 10-10 10S2 17.523 2 12 6.477 2 12 2Zm0 2a8 8 0 1 0 0 16 8 8 0 0 0 0-16ZM8 12a1 1 0 0 1 1-1h6a1 1 0 0 1 0 2H9a1 1 0 0 1-1-1Z',
+})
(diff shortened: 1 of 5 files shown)


turbocrime:
simple things that might be good to change:
- make (more of? entire?) collapsed meta clickable to expand, rather than navigate to the post
- change the toggle icon hover cursor to the pointer cursor (button doesn't do this automatically?)
- collapse animation

some possible improvements that would raise visual design questions:
- display item count in collapsed meta. not very complex to implement.
- spacing between the 'collapse' icon and the reply icon, to match controls spacing. probably means pushing post controls component to the right, which would be inconsistent with the linear thread view.

turbocrime:
~~ideally, tree state might want to be in `thread.state` like other parts of thread ui state, handled by `usePostThread` hook. this could simplify downstream components but it seems more complex overall, and i'm uncertain if it would create rerender churn.~~

~~keeping a set of post uris feels slightly incorrect, but it is effective and performant.~~

after exploring inclusion within `thread.state` and `usePostThread`, it looks
like presentational state doesn't fit there. other state there is totally
derived, or directly concerns queries.

turbocrime:
it is possible to store tree state persistently, and apply it across all views. i chose not to do that because:
- state logic would be more complex
- persistent state might be confusing to users
- expiring that state is probably desirable, but that raises a lot of questions.

i think this simple component state implementation is preferable.

PR #9637, opened 2026-01-05
Watermark posts in screenshots

mozzius:
Adds a watermark to screenshots of posts on iOS.

When using the app, the watermark is not visible at all

## How it works

I made an expo module [`expo-privacy-sensitive`](https://www.npmjs.com/package/expo-privacy-sensitive). How it works is it takes the child view, and stuffs it into the password field. iOS automatically hides the password content (the dots) in screenshots, so by stuffing the view into there it gets hidden in screenshots.

I then added the logo, and absolutely positioned the follow button on top of it. when the button is hidden, the watermark is revealed!

## Test plan

on iOS, test posts with:
- Follow button
- Follow back button
- No button (already following)
- Long handle/display name (such as [deepfates](https://bsky.app/profile/deepfates.com.deepfates.com.deepfates.com.deepfates.com.deepfates.com))

### DIFF
--- a/src/screens/PostThread/components/GrowthHack.tsx
+++ b/src/screens/PostThread/components/GrowthHack.tsx
@@ -5,7 +5,10 @@ import {PrivacySensitive} from 'expo-privacy-sensitive'
 import {useAppState} from '#/lib/hooks/useAppState'
 import {isIOS} from '#/platform/detection'
 import {atoms as a, useTheme} from '#/alf'
+import {sizes as iconSizes} from '#/components/icons/common'
 import {Mark as Logo} from '#/components/icons/Logo'
+
+const ICON_SIZE = 'xl' as const
 
 export function GrowthHack({
   children,
@@ -48,11 +51,15 @@ export function GrowthHack({
         ]}>
         <View
           onLayout={evt => setWidth(evt.nativeEvent.layout.width)}
-          style={t.atoms.bg}>
+          style={[
+            t.atoms.bg,
+            // make sure it covers the icon! the won't always be a button
+            {minWidth: iconSizes[ICON_SIZE], minHeight: iconSizes[ICON_SIZE]},
+          ]}>
           {children}
         </View>
       </PrivacySensitive>
-      <Logo size="xl" />
+      <Logo size={ICON_SIZE} />
     </View>
   )
 }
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Well that is just terrific

quiple:
It doesn't seem to apply to one's own posts.

mozzius:
1. It's only on the iOS app, as it's not possible on other platforms
2. It's only present on the main post of a thread

mcclure:
I'm exempt from this due to using Android, but I *really* don't like this. Like this is basically malware behavior as far as I'm concerned.

dannypage:
> How it works is it takes the child view, and stuffs it into the password field. iOS automatically hides the password content (the dots) in screenshots, so by stuffing the view into there it gets hidden in screenshots.

Exploiting a privacy feature to make screenshots prettier is very short-sighted. Given the feature branch was called `growth-hack`, I assume there wasn’t a security reason for this change?

roxaloxa:
Absolutely horrendous and hostile user experience to alter screenshots.

dannypage:
Follow up: If a watermark feature like this must exist, an alternative that doesn’t break completely user expectation and doesn’t exploit a privacy feature would be to have a “copy post as image” option, similar to Reddit when you copy an image on their site.

Meorge:
This seems to me like it should be opt-in at the least, given that the expected behavior of a screenshot is capturing exactly what's on the user's screen at that moment in time. dannypage's suggestion of implementing a dedicated "copy post as image" option that shows this watermark, while keeping screenshots unchanged, is also a good alternative.

brian-farrell:
> Adds a watermark to screenshots of posts on iOS.
> 
> 
> When using the app, the watermark is not visible at all
> ## How it works
> 
> I made an expo module [`expo-privacy-sensitive`](https://www.npmjs.com/package/expo-privacy-sensitive). How it works is it takes the child view, and stuffs it into the password field. iOS automatically hides the password content (the dots) in screenshots, so by stuffing the view into there it gets hidden in screenshots.
> 
> I then added the logo, and absolutely positioned the follow button on top of it. when the button is hidden, the watermark is revealed!
> ## Test plan
> 
> on iOS, test posts with:
> 
>     * Follow button
> 
>     * Follow back button
> 
>     * No button (already following)
> 
>     * Long handle/display name (such as [deepfates](https://bsky.app/profile/deepfates.com.deepfates.com.deepfates.com.deepfates.com.deepfates.com))

WHY would you implement such a ridiculous thing? NOT WANTED.

PR #9643, opened 2026-01-06
Enable lightbox for profile banner

mozzius:
Makes the profile banner pressable, opening the image lightbox

https://github.com/user-attachments/assets/ccd25c13-2460-4202-b4bf-b18b5402e15b

### DIFF
--- a/src/screens/Profile/Header/Shell.tsx
+++ b/src/screens/Profile/Header/Shell.tsx
@@ -77,7 +77,7 @@ let ProfileHeaderShell = ({
     (
       uri: string,
       thumbRect: MeasuredDimensions | null,
-      type: 'circle-avi' | 'profile-banner' = 'circle-avi',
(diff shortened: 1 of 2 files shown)


src/screens/Profile/Header/Shell.tsx
```diff
@@ -73,20 +74,31 @@ let ProfileHeaderShell = ({
...
-            },
+            dimensions:
+              type === 'circle-avi'
+                ? {
+                    // It's fine if it's actually smaller but we know it's 1:1.
+                    height: 1000,
+                    width: 1000,
+                  }
+                : {
+                    // Banner aspect ratio is 3:1
```
estrattonbailey, line 96: Is it exactly? Does it matter?

mozzius: That's what it crops to on upload - I tested with setting a different value and it seems to be fine. It gets replaced with the real values once the image loads

PR #9644, opened 2026-01-06
Add recId from suggested users

mozzius:


estrattonbailey:
See Slack thread, not sure we care about the recId on other follows, may just confuse us later on. For the events we care about, we should probably just add the `suggestedUser:follow` event to all places we care about

mozzius:
Copy that 🫡

PR #9648, opened 2026-01-07
Profile minimal header better threshold

M21H:
Before:

https://github.com/user-attachments/assets/684b8a9b-4338-4993-bf20-e018dbaf8829

After:

https://github.com/user-attachments/assets/785d2e56-a233-4a86-854b-bdea671864a5

### DIFF
--- a/src/screens/Profile/Header/index.tsx
+++ b/src/screens/Profile/Header/index.tsx
@@ -98,6 +98,8 @@ let ProfileHeader = ({setMinimumHeight, ...props}: Props): React.ReactNode => {
 ProfileHeader = memo(ProfileHeader)
 export {ProfileHeader}
 
+const headerThreshold = 97
+
 const MinimalHeader = memo(function MinimalHeader({
   onLayout,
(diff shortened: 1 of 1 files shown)


src/screens/Profile/Header/index.tsx
```diff
@@ -98,6 +98,8 @@ let ProfileHeader = ({setMinimumHeight, ...props}: Props): React.ReactNode => {
 ProfileHeader = memo(ProfileHeader)
 export {ProfileHeader}
 
+const headerThreshold = 97
```
mozzius: could you make this all uppercase? like `HEADER_THRESHOLD` (makes it clearer it's just a constant)

mozzius (commented):
lol nice thanks for figuring this one out

mozzius:
Hmm, strange: for me, there's still a bit of a jump. Something like 95 looks closer on my simulator

https://github.com/user-attachments/assets/9f204785-d018-4c3b-8cdc-19ce1bee3abc

PR #9650, opened 2026-01-07
[AAv2] Improve minimum age handling

estrattonbailey:
We already require a minimum age of 13 in order to create an account. In some regions, this age value is different e.g. in Australia it's 16. This PR integrates https://github.com/bluesky-social/atproto/pull/4513 so that this new `minAccessAge` value can be configured by region. We also use this value to determine if a user is able to initiate the Age Assurance flow.


estrattonbailey:
This one is a pain to test and I'm confident in it, so just gonna merge to rebase some other things on top

mozzius:
🫡 I did give it a read-through, nothing jumped out

PR #9653, opened 2026-01-08
Standardize metadata for client events in feeds

abenzer:


src/components/PostControls/ShareMenu/index.tsx
```diff
@@ -45,9 +47,11 @@ let ShareMenuButton = ({
   threadgateRecord?: AppBskyFeedThreadgate.Record
   onShare: () => void
   hitSlop?: Insets
+  logContext: 'FeedItem' | 'PostThreadItem' | 'Post' | 'ImmersiveVideo'
```
ThisIsMissEm, line 50: This union looks repeated, should it be an exported type?

mozzius:
@claude can you review this

PR #9660, opened 2026-01-08
Handle AA config failure

estrattonbailey:
If the config fails to load, we can't really do anything. But users would be very confused and think they were being age restricted. Hopefully this clarifies a bit.

CleanShot 2026-01-08 at 12 07 08@2x
CleanShot 2026-01-08 at 12 07 17@2x
CleanShot 2026-01-08 at 12 07 21@2x
CleanShot 2026-01-08 at 12 07 28@2x
CleanShot 2026-01-08 at 12 07 35@2x

### DIFF
--- a/src/components/ageAssurance/AgeAssuranceErrors.tsx
+++ b/src/components/ageAssurance/AgeAssuranceErrors.tsx
@@ -1,16 +1,37 @@
-import {Trans} from '@lingui/macro'
+import {msg, Trans} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
 
 import {type ViewStyleProp} from '#/alf'
-import {Admonition} from '#/components/Admonition'
+import * as Admonition from '#/components/Admonition'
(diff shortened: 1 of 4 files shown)


src/components/ageAssurance/AgeAssuranceErrors.tsx
```diff
@@ -0,0 +1,16 @@
+import {Trans} from '@lingui/macro'
+
+import {type ViewStyleProp} from '#/alf'
+import {Admonition} from '#/components/Admonition'
+
+export function AgeAssuranceConfigUnavailableError(props: ViewStyleProp) {
+  return (
+    <Admonition type="error" style={props.style}>
+      <Trans>
+        We were unable to load age assurance configuration for your region,
```
surfdude29: nit: "… unable to load **the** age assurance configuration for your region …" would read a little better imo

EDIT: or maybe "… unable to load age assurance configuration **data** for your region …"

mozzius (approved):
Tested, works well. Nice one!

PR #9663, opened 2026-01-08
Do not render links if uri is invalid

estrattonbailey:
Lifted validation we use in `RichText.detectFacets` so that any `link` facet that comes down the pipe must contain a valid URL in order to be rendered as a URL in our app. No reason to be rendering stuff like this https://bsky.app/profile/scoiattolo.mountainherder.xyz/post/3mbuvztjijs2y

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -11,7 +11,7 @@ import {RichTextTag} from '#/components/RichTextTag'
 import {Text, type TextProps} from '#/components/Typography'
 
 const WORD_WRAP = {wordWrap: 1}
-// lifted from facet detection in `RichText` impl
+// lifted from facet detection in `RichText` impl, _without_ `gm` flags
 const URL_REGEX =
   /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/i
 


src/components/RichText.tsx
```diff
@@ -11,6 +11,9 @@ import {RichTextTag} from '#/components/RichTextTag'
 import {Text, type TextProps} from '#/components/Typography'
 
 const WORD_WRAP = {wordWrap: 1}
+// lifted from facet detection in `RichText` impl
+const URL_REGEX =
+  /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/i
```
mozzius: I think this is actually exported from `@atproto/api`, can just do
```ts
import {URL_REGEX} from '@atproto/api'
```

mozzius (approved):

src/components/RichText.tsx
estrattonbailey: The `g` and `m` flags on the original are messing with the matching. Lemme add that to the comment.

src/components/RichText.tsx
```diff
@@ -11,6 +11,9 @@ import {RichTextTag} from '#/components/RichTextTag'
 import {Text, type TextProps} from '#/components/Typography'
 
 const WORD_WRAP = {wordWrap: 1}
+// lifted from facet detection in `RichText` impl
```
estrattonbailey:
```suggestion
// lifted from facet detection in `RichText` impl, _without_ `gm` flags
```

src/components/RichText.tsx
mozzius: ah ok cool that's fine, just leave a comment otherwise I will totally accidentally refactor it in 6 months lol

ThisIsMissEm:
Why not use URL.canParse instead of the regex? Regex looks like it's for matching links in text, when uri should be well formed?

PR #9665, opened 2026-01-09
chore: react native compressor package to remove crash in specific case

numandev1:
updated react-native-compressor package to remove crash + audio patch


mozzius:
Awesome, thank you @numandev1! Will review properly when I get the chance

PR #9670, opened 2026-01-09
Reduce startup hang by not querying fonts

mozzius:
Patch of https://github.com/expo/expo/pull/42033 which contains more details. tl;dr `expo-font` queries all the available fonts at startup. This can be quite expensive (I can see in Xcode it's 26% of our hangs), and we don't even use it because we use the config plugin.

Screenshot 2026-01-09 at 13 23 19

Easy fix: make it a `lazy var`, so it only gets queried when it's used.  

## Test plan

- Ensure iOS app builds

I confirmed it does not query fonts until needed (I called `getLoadedFonts` after 10s):

Screenshot 2026-01-09 at 13 56 34


mozzius:
What do you think @claude my boy

mozzius:
Good enough for Expo means good enough for us, will merge

PR #9674, opened 2026-01-09
Send interests to all feeds

gaearon:
## Why

Given that Bluesky's algorithmic feeds are still a bit "meh", third-party feeds such as [For You](https://bsky.app/profile/did:plc:3guzzweuqraryl3rdkimjamk/feed/for-you) have become indispensable to the ecosystem. Although they're now competitive for users with existing data like Follows and Likes, they're still at a disadvantage compared to Discover for the cold start case where the only known information about the user is their selected list of interests.

This PR makes the selected list of interests available to all feeds, solving the problem.

## Why this is fine

The current list of interests itself is pretty tame:

```
  animals, art, books, comedy, comics, culture, dev, education,
  finance, food, gaming, journalism, movies, music, nature, news,
  pets, photography, politics, science, sports, tech, tv, writers
```

There's nothing embarrassing in either of those. It is also stored per-account so there is no risk of identifying alts by a unique combination or something like that.

I don't think there is any meaningful reason to keep it only available to the Bluesky-operated feeds. And there is a significant upside to opening it up.

For simplicity, I opted to open it up to all feeds (rather than, say, just to Saved or Pinned feeds). This is to avoid thinking about weird race conditions where you might see different things depending on whether you just pinned it or not. This also avoids the problem where a feed preview is inaccurate and doesn't match your experience if you actually started using it. It seems important to serve relevant content right away.

## When to reconsider this

If the list of interests is significantly expanded to include something that people might prefer to keep private, this approach may no longer make sense. In that case something separate could be added. I'd also say any strategy that relies on privileging Bluesky's feeds is probably bad anyway though.

## Implementation

I kept the existing `x-bsky-topics` header name. Yes it's "non-standard" but the point of headers is you can tack on things that you don't have a better place for. It would be a one-line change to read it from some more official place if this gets "standardized" later. I think it's fine.

## Test Plan

Observe it being sent (when present):

Screenshot 2026-01-09 at 18 10 20

### DIFF
--- a/src/lib/api/feed/custom.ts
+++ b/src/lib/api/feed/custom.ts
@@ -10,7 +10,7 @@ import {
   getContentLanguages,
 } from '#/state/preferences/languages'
 import {type FeedAPI, type FeedAPIResponse} from './types'
-import {createBskyTopicsHeader, isBlueskyOwnedFeed} from './utils'
+import {createBskyTopicsHeader} from './utils'
 
 export class CustomFeedAPI implements FeedAPI {
   agent: BskyAgent
@@ -52,7 +52,6 @@ export class CustomFeedAPI implements FeedAPI {
   }): Promise<FeedAPIResponse> {
     const contentLangs = getContentLanguages().join(',')
     const agent = this.agent
-    const isBlueskyOwned = isBlueskyOwnedFeed(this.params.feed)
 
     const res = agent.did
       ? await this.agent.app.bsky.feed.getFeed(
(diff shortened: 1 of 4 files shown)


ThisIsMissEm:
Don't feedgens usually get parameters via the URL? Also, would this X-Bsky-Topics header be correctly proxied back to the feedgen by the pds?

gaearon:
>Don't feedgens usually get parameters via the URL?

For officially documented params, I think so, but this isn't in the Lexicon. It's how Discover works today though. So this is just opening up the same pattern to others.

>Also, would this X-Bsky-Topics header be correctly proxied back to the feedgen by the pds?

I presume PDS probably doesn't special-case Discover (I hope) so I'd expect it would work with other feeds. But I haven't checked.

thepudds:
Sorry for quick drive-by comment, but if there are concerns about doing this, maybe the onboarding UI could in theory also be tweaked to mention that the interests can be shared with third-party feed generators.

That said (and as @gaearon said above), the current topic list does seem fairly uninteresting, so maybe not needed.

In any event, sending interests to third-part feeds in some way does seem valuable.

PR #9677, opened 2026-01-10
Replace Add User to Lists Modal with modern Dialog

mozzius:
Replace deprecated `UserAddRemoveLists.tsx` modal with a modern Dialog component that uses the efficient `getListsWithMembership()` API.

## Implementation details

- Created `useListsWithMembershipQuery()` hook that fetches lists with membership status in a single efficient call (instead of fetching all 100k+ list items)
- Created new `UserAddRemoveListsDialog` component following the `StarterPackDialog` pattern
- Removed the dangerous `useDangerousListMembershipsQuery()` entirely
- Removed the shared `MyLists` component dependency - dialog owns its data fetching directly
- Added optimistic updates so membership changes appear instantly in the UI
- Updated `ProfileMenu.tsx` and `ListMembers.tsx` to use the new dialog instead of modal

## Test plan

- Open user profile, tap menu → "Add to Lists", verify dialog opens with user's lists
- Add/remove user from a list, verify membership status updates immediately
- Verify close button works
- Check for empty state when user has no lists
- Verify error toast appears on failed mutations
- Inspect network tab to confirm `getListsWithMembership` is being called instead of large list item queries


src/state/queries/list-memberships.ts
```diff
@@ -133,43 +46,18 @@ export function useListMembershipAddMutation({
...
         },
       )
-      // TODO
-      // we need to wait for appview to update, but there's not an efficient
-      // query for that, so we use a timeout below
-      // -prf
       return res
     },
     onSuccess: (data, variables) => {
-      // manually update the cache; a refetch is too expensive
```
estrattonbailey, line 143: We lost optimistic updates here, same for removal. Quick glance at the query cache tells me we don't have the data we need to optimistically update the existing `useAllListMembersQuery` query that powers the `getMembership` check above. That query is also gross bc it tries to fetch up to 6 pages of results.

I don't have immediate ideas and I tried to one-shot with Claude but it broke profile shadows and still didn't optimistically update, so just gonna punt on this one so we can focus on other stuff.

src/state/queries/lists-with-membership.ts
```diff
@@ -0,0 +1,120 @@
...
+  useInfiniteQuery,
+} from '@tanstack/react-query'
+
+import {useAgent} from '#/state/session'
+
+export type ListWithMembership =
+  AppBskyGraphGetListsWithMembership.ListWithMembership
+
+const RQKEY_ROOT = 'lists-with-membership'
+export const RQKEY = (actor: string) => [RQKEY_ROOT, actor]
```
estrattonbailey: We should stop using these non-uniquely named query keys

mozzius: Yeah agreed

src/state/queries/list-memberships.ts
mozzius: Hmmmm that's weird, I'll double check it because I thought it was working. Could be wrong

mozzius: No okay this was updating `useDangerousListMembershipsQuery()`, which no longer exists. We should probably optimistically update the other query too,but this isn't a regression iirc

You can see we still invalidate the other query below

mozzius:
@claude please review

mozzius:
@claude rereview

mozzius:
@estrattonbailey couple rounds of claudereview have left this in a good state I think

estrattonbailey (approved):

PR #9695, opened 2026-01-14
Improved client events for feed interactions

abenzer:


src/state/queries/post.ts
```diff
@@ -111,7 +111,7 @@ export function usePostLikeMutationQueue(
   const postCid = post.cid
   const initialLikeUri = post.viewer?.like
   const likeMutation = usePostLikeMutation(feedDescriptor, logContext, post)
-  const unlikeMutation = usePostUnlikeMutation(feedDescriptor, logContext)
+  const unlikeMutation = usePostUnlikeMutation(feedDescriptor, logContext, post)
```
mozzius, line 114: just a thought, not for this PR: we're stuffing more and more params into hooks like this. I think we should probably use a pattern like:

```ts
function useWhateverMutation(feedDescription: string, metrics: { logContext: string, post: etc }) {}
```

basically bundle all the metrics-related data into a single param, otherwise the code is really hard to parse / inconsistent between different hooks imo

src/state/queries/post.ts
```diff
@@ -288,7 +301,12 @@ function usePostRepostMutation(
     {uri: string; cid: string; via?: {uri: string; cid: string}} // the post's uri and cid, and the repost uri/cid if present
   >({
     mutationFn: ({uri, cid, via}) => {
-      logger.metric('post:repost', {logContext, feedDescriptor})
+      logger.metric('post:repost', {
+        uri,
+        authorDid: post.author.did,
```
mozzius, line 306: Technically you can extract the `did` from the AT URI

```ts
const atUri = new AtUri(uri)
const authorDid = atUri.host
```

mozzius (approved):

mozzius:
@claude please take a look at this, are we missing any spots?

PR #9696, opened 2026-01-14
[APP-1767] Live event feeds

estrattonbailey:
We're going to be experimenting with promoting feeds tailored to live events. We'll surface these in a few places in the app, and provide some controls over how they display. More to come here! This is just the start.

CleanShot 2026-01-13 at 20 55 49@2x
CleanShot 2026-01-13 at 20 55 52@2x
CleanShot 2026-01-13 at 20 55 59@2x
CleanShot 2026-01-13 at 21 17 48@2x
CleanShot 2026-01-14 at 11 27 07@2x

### DIFF
--- a/src/features/liveEvents/preferences.ts
+++ b/src/features/liveEvents/preferences.ts
@@ -9,6 +9,7 @@ import {
   usePreferencesQuery,
 } from '#/state/queries/preferences'
 import {useAgent} from '#/state/session'
+import * as env from '#/env'
 import {
   type LiveEventFeed,
   type LiveEventFeedMetricContext,
@@ -40,7 +41,7 @@ function useWebOnlyDebugLiveEventPreferences() {
   const agent = useAgent()
 
   useEffect(() => {
-    if (isWeb && typeof window !== 'undefined') {
+    if (env.IS_DEV && isWeb && typeof window !== 'undefined') {
       // @ts-ignore
       window.__updateLiveEventPreferences = async (
         action: LiveEventPreferencesAction,


src/features/liveEvents/components/LiveEventFeedOptionsMenu.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+        <Text style={[a.text_md, a.leading_snug]}>
+          <Trans>
+            If you choose to hide all events, you can always re-enable them from{' '}
+            <Span style={[a.font_semi_bold]}>Settings → Content & Media</Span>.
+          </Trans>
+        </Text>
+      </View>
+
+      <View style={[a.gap_sm]}>
```
mozzius, line 126: We **really** need to decide on a consistent amount of spacing between our buttons. Some places use `md`, some places use `sm`

src/features/liveEvents/components/LiveEventFeedOptionsMenu.tsx
```diff
@@ -0,0 +1,167 @@
...
+        <Button
+          label={_(msg`Hide this event`)}
+          size="large"
+          color="primary_subtle"
+          onPress={() => {
+            update({type: 'hideFeed', id: feed.id})
+            undoAction.current = {type: 'unhideFeed', id: feed.id}
```
mozzius: cool pattern - maybe could put this in the `onMutate` section of the mutation? That way you could move that logic into the hook

estrattonbailey: Nice idea, I think I sorted an OK solution, will push in a bit

src/features/liveEvents/components/LiveEventFeedOptionsMenu.tsx
estrattonbailey: Yeah we need to lock down dialog components in general. We did some designs but will need to make some stable/flexible base components and migrate everything at some point

src/features/liveEvents/preferences.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  return useMutation<
+    AppBskyActorDefs.LiveEventPreferences,
+    Error,
+    LiveEventPreferencesAction,
+    {undoAction: LiveEventPreferencesAction | null}
+  >({
+    onSettled(data, error, variables) {
+      // If __canUndo is not explicitly set to false, we allow undo
+      const canUndo = variables.__canUndo === undefined ? true : false
```
mozzius: Confused by this... if `__canUndo` is `true`, does that mean `canUndo` becomes `false`?

src/features/liveEvents/preferences.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      switch (action.type) {
+        case 'hideFeed':
+        case 'unhideFeed': {
+          if (!props.feed) {
+            if (env.IS_DEV) {
+              throw new Error(
+                'props.feed is required when calling hideFeed or unhideFeed',
+              )
+            }
```
mozzius: dev-only throw. should be an error log or throw every time

src/features/liveEvents/preferences.ts
estrattonbailey: Good call f7edfafb20e511b7c5c67e3e174fd2b09b8357ce

src/features/liveEvents/preferences.ts
estrattonbailey: Good call f7edfafb20e511b7c5c67e3e174fd2b09b8357ce

src/features/liveEvents/preferences.ts
```diff
@@ -73,7 +72,18 @@ export function useUpdateLiveEventPreferences(props: {
...
+       * here is that we want to invert the action that was just passed in, and
+       * provide it as an `undoAction` to the `onUpdateSuccess` callback.
+       *
+       * If the operation was not a success, we don't provide the `undoAction`.
+       *
+       * Upon the first call of the mutation, the `__canUndo` flag is undefined,
+       * so we allow the undo. However, when we create the `undoAction`, we
+       * set its `__canUndo` flag to false, so that if the user were to call
+       * the undo action, we would not provide another undo for that.
+       */
```
mozzius: nice thank you 👍

PR #9710, opened 2026-01-16
Add Russian translation for bookmarks and saved feeds

keyboy21:


keyboy21:
I will add more translation later

PR #9714, opened 2026-01-16
Make live now gate fail-open

mozzius:
Allows adblock users to use Live Now

**NOT TESTED**


mozzius:
@claude did I do this correctly

PR #9734, opened 2026-01-21
[APP-1782] Analytics migration

estrattonbailey:


src/analytics/identifiers/session.ts
```diff
@@ -0,0 +1,40 @@
...
+  return sessionId
+}
+
+export function useSessionId() {
+  const [id, setId] = useState(() => sessionId)
+
+  useEffect(() => {
+    const sub = onAppStateChange(state => {
+      if (state === 'active') {
+        const lastEvent = device.get(['nativeSessionIdLastEventAt'])
```
estrattonbailey, line 27: If not expired I need to bump the last event up

src/analytics/metrics/types.ts
```diff
@@ -374,7 +383,6 @@ export type MetricEvents = {
       | 'AvatarButton'
       | 'StarterPackProfilesList'
       | 'FeedInterstitial'
-      | 'ProfileHeaderSuggestedFollows'
```
estrattonbailey, line 377: whoops

src/state/feed-feedback.tsx
```diff
@@ -85,22 +85,55 @@ export function useFeedFeedback(
...
     () =>
       throttle(
         () =>
-          flushToStatsig(
+          flushEvents(
             aggregatedStats.current,
             feed?.feedDescriptor ?? 'unknown',
           ),
-        45e3,
+        5e3,
```
estrattonbailey: Ope debug slipped through

src/Navigation.tsx
```diff
@@ -973,47 +980,101 @@ function RoutesContainer({children}: React.PropsWithChildren<{}>) {
...
+      // See BottomBar.tsx for an example of how to handle nested navigators in the tabs correctly.
+      // I'm scared of missing a spot (esp. with push notifications etc) so let's enable this legacy behaviour for now.
+      // We will need to confirm we handle nested navigators correctly by the time we migrate to React Navigation 8.x
+      // -sfn
+      navigationInChildEnabled>
+      <AnalyticsContext
+        metadata={utils.useMeta({
+          navigation: {
+            previousScreen: prevLoggedRouteName.current,
+            currentScreen: getCurrentRouteName(),
```
estrattonbailey: Should see if we actually get the right values here

PR #9739, opened 2026-01-23
Changing functionality of "developer mode"

ETJeanMachine:
Right now, on bsky, there's two things about "developer mode" (enabled by holding down the version number in settings) I find extremely odd/confusing:

1. How you enter it.
2. How it works.

Currently, dev mode allows you to "share" DID's and post AT URI's. This, at least on android, results in a *pop up* modal showing to send an AT URI/DID via messaging apps. This is, extremely confusing, lol? I cannot think of the last time I thought to myself: "man, you know what I want to do: text my mom my account DID for my bluesky account." This PR changes the "share" button into a "copy" button that just plainly copies it to your clipboard instead.

The other thing is how you enter into dev mode - in almost all system's I've used that have developer options you can access via interacting with a version number, it's usually via *tapping* a version number multiple times; not holding it. This PR changes the holding action into a tapping action, and gives you a toast indicating how many more times you have to tap in order to enable/disable developer mode.

For clarity, I had Claude mostly write this, but considering it's just existing functionality (you copy links to your clipboard in the mobile apple via the share modal), I figured this was a totally fine use of it. My lookover/testing of the code seems to indicate this seems mostly accurate for the feature, it's not exactly complex. It also updated locales which like, uh, sure. Seems fine lol.

### DIFF
--- a/src/components/PostControls/PostMenu/PostMenuItems.tsx
+++ b/src/components/PostControls/PostMenu/PostMenuItems.tsx
@@ -95,7 +95,6 @@ import {
 import * as Prompt from '#/components/Prompt'
 import {useAnalytics} from '#/analytics'
 import {IS_INTERNAL} from '#/env'
-import {useDevMode} from '#/storage/hooks/dev-mode'
 import * as bsky from '#/types/bsky'
 
 let PostMenuItems = ({
@@ -125,7 +124,6 @@ let PostMenuItems = ({
   const {hasSession, currentAccount} = useSession()
   const {_} = useLingui()
   const ax = useAnalytics()
-  const [devModeEnabled] = useDevMode()
   const langPrefs = useLanguagePrefs()
   const {mutateAsync: deletePostMutate} = usePostDeleteMutation()
   const {mutateAsync: pinPostMutate, isPending: isPinPending} =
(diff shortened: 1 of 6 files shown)


AaronDewes:
Changing the way developer mode is activated has no benefit in my opinion. Also, the "Share" modal on Android has a copy button. This PR just makes it harder to share a DID without really adding an benefit.

ETJeanMachine:
> Changing the way developer mode is activated has no benefit in my opinion. Also, the "Share" modal on Android has a copy button. This PR just makes it harder to share a DID without really adding an benefit.

I think the reason why I want the change is precisely because I cannot think of a situation where you would want to "share" a DID rather than copy it? The main reason I personally grab DID's is to paste them into specific places - PDSls, a .env file for a project I'm working on, etc (these are not places you are going to "share" to). This saves a click in the process of doing that. If this doesn't align with the way the bsky dev team sees things, sure, but I do think it's worth noting that in desktop already it's a "copy" button, on sites like Discord their dev mode does copying (not sharing), because this more accurately reflects the actual needs of developers.

The other change is just to be consistent with other platforms that have dev modes. I'm open to removing that from the PR.

AaronDewes:
In general, it's good not to do two different things in one PR.

Apart from that, removing the share feature just because you don't think it is useful is not a good idea in my opinion, because you're not the only person using this app. Copying the data now is much simpler than it would be to share after this PR.

PR #9758, opened 2026-01-26
Fix: yarn install by pining the package.json

leafwind:
Hi, this is my first PR to this repo, and please feel free to let me know if this is not the best way to go.

### Problem

We have these patch files which require these modules need to be EXACTLY the version, otherwise `yarn install` will fail.

Many of the patches can tolerate the version drift if they don't conflict, but some of them are not, and this is the list of them:
- `"expo": "=54.0.27"`: `patch/expo-modules-core+3.0.28.patch`
- `"react-native-compressor": "=1.13.0"`: `patch/react-native-compressor+1.13.0.patch`
- `"sonner-native": "=0.21.0"`: `patch/sonner-native+0.21.0.patch`

For example, if you remove the `yarn.lock` and do `yarn install` again:

```
**ERROR** Failed to apply patch for package expo-modules-core at path
  
    node_modules/expo-modules-core

  This error was caused because expo-modules-core has changed since you
  made the patch file for it. This introduced conflicts with your patch,
  just like a merge conflict in Git when separate incompatible changes are
  made to the same piece of code.

  Maybe this means your patch file is no longer necessary, in which case
  hooray! Just delete it!

  Otherwise, you need to generate a new patch file.

  To generate a new one, just repeat the steps you made to generate the first
  one.

  i.e. manually make the appropriate file changes, then run 

    patch-package expo-modules-core

  Info:
    Patch file: patches/expo-modules-core+3.0.28.patch
    Patch was made for version: 3.0.28
    Installed version: 3.0.29
```

### Proposed temporary solution

I understand that `yarn.lock` is made for this purpose, but for soft-forked downstream developers like me, if I remove some of the packages/modules on my own build, I need to update `yarn.lock` anyway, and the current `package.json` is now blocking build and will cause all the soft-downstream forks drift.

The long term solution is to fix those patches first, but now this can unblock the build.

Let me know if this makes sense to you, and I appreciate your time to review it. 🙇‍♂️


leafwind:
nvm, I gave up it, just use `yarn remove {package}` can keep the minimum change while able to remove packages I don't need anymore.

PR #9769, opened 2026-01-27
Fix `IS_DEV` in app config

mozzius:
Currently `IS_DEV` is defined as:

```
const IS_TESTFLIGHT = process.env.EXPO_PUBLIC_ENV === 'testflight'
const IS_PRODUCTION = process.env.EXPO_PUBLIC_ENV === 'production'
const IS_DEV = !IS_TESTFLIGHT || !IS_PRODUCTION
```

This is going to be always true, as it's a logical contradiction

Fix is to use `&&`


mozzius:
@claude is this correct?

PR #9773, opened 2026-01-27
Fix Live Now not showing for non-streamer

estrattonbailey:
When validating the actor `status` record, we were comparing the domain against the allowed domains _for the viewing account_. Which in the case of some domains that are only available to certain users, mean that livestreams with those domains were showing as _not_ live.


estrattonbailey:
Tested all cases locally 👍

mozzius:
@claude take a look please!

PR #9776, opened 2026-01-27
[APP-1503] Add content inset to modlist

vineyardbovines:


src/view/com/lists/ListMembers.tsx
```diff
@@ -256,6 +256,7 @@ export function ListMembers({
         contentContainerStyle={{
           minHeight: Dimensions.get('window').height * 1.5,
         }}
+        contentInset={{bottom: 200}}
```
mozzius: I think that that prop is iOS only?
https://reactnative.dev/docs/next/scrollview#contentinset-ios

I think the move here is increasing the height of the ListFooter component perhaps?

vineyardbovines: shows how much i've been in iOS lately 📴 

the pattern i'm seeing in other spots sets `paddingBottom: 200` in `contentContainerStyle` (just updated to this). we also have `ListFooterComponent` applied here, it's just `null` if the list is empty so i could set a spacer view. any specific thought either way?

mozzius: haha all good!

yeah, that's also a totally reasonable fix, but tbh I'm surprised ListFooter doesn't do the job. It should be rendering when the list is long enough to go off the screen. any idea what's causing the list to go so far off the screen? I can see we've already specified a custom height for the footer, which should cover it

vineyardbovines: the root of it going off screen in the first place is from the tab bar being an overlay and so the inner list scrollview can't take it into account when calculating the available height. ListFooter is within that so it has the same 'height lgtm'

in writing that i remembered `useBottomTabBarHeight` exists, so i actually swapped out the hardcoded `200` with bottomTabBar * 2 to account for tab bar + padding. slightly better.

mozzius:
Unfortunately `useBottomTabBarHeight` doesn't work on web :(

Screenshot 2026-02-11 at 16 39 03

mozzius:
Hey, I had a proper look at this. Seems there's more going on than just missing padding, it's a nasty quirk of how the list expects to be a child of the PagerWithHeader component.

This is honestly a much nastier bug than I realised, sorry for giving it to you as your first ticket, I wouldn't have done if I'd known!

Check out the fix I found here: https://github.com/bluesky-social/social-app/pull/9858

PR #9779, opened 2026-01-27
[APP-1784] Proper fix for Live Now status not showing

estrattonbailey:
Improves on #9773

### DIFF
--- a/src/state/service-config.tsx
+++ b/src/state/service-config.tsx
@@ -97,28 +97,30 @@ export function useLiveNowConfig(): LiveNowConfig {
   const ctx = useContext(LiveNowContext)
   const canGoLive = useCanGoLive()
   const {currentAccount} = useSession()
-  const defaultAllowedHosts = new Set(DEFAULT_LIVE_ALLOWED_DOMAINS)
-  const allowedHostsExceptionsByDid = new Map<string, Set<string>>()
-  for (const live of ctx) {
-    allowedHostsExceptionsByDid.set(
-      live.did,
-      new Set(DEFAULT_LIVE_ALLOWED_DOMAINS.concat(live.domains)),
-    )
-  }
-  if (!currentAccount?.did || !canGoLive)
+  return useMemo(() => {
+    const defaultAllowedHosts = new Set(DEFAULT_LIVE_ALLOWED_DOMAINS)
+    const allowedHostsExceptionsByDid = new Map<string, Set<string>>()
+    for (const live of ctx) {
+      allowedHostsExceptionsByDid.set(
+        live.did,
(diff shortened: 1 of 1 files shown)


src/lib/actor-status.ts
```diff
@@ -73,10 +74,15 @@ export function isStatusValidForViewers(
   config: LiveNowConfig,
 ) {
   if (status.status !== 'app.bsky.actor.status#live') return false
+  if (!status.uri) return false // should not happen, just backwards compat
```
estrattonbailey, line 77: I added this field to the lexicon as an optional field, but tbh we could make this required and remove this in the future since it's a view and not a record lex.

src/lib/actor-status.ts
```diff
@@ -73,10 +74,15 @@ export function isStatusValidForViewers(
...
   try {
+    const {host: liveDid} = new AtUri(status.uri)
```
estrattonbailey, line 79: Since this should always be a DID coming back on our `statusView`, this should be safe

src/state/service-config.tsx
```diff
@@ -89,28 +89,35 @@ const DEFAULT_LIVE_ALLOWED_DOMAINS = [
...
-  const allVipDomains = new Set(ctx.flatMap(live => live.domains))
-  const allSupportedDomains = new Set(
-    DEFAULT_LIVE_ALLOWED_DOMAINS.concat(Array.from(allVipDomains)),
-  )
+  const defaultAllowedHosts = new Set(DEFAULT_LIVE_ALLOWED_DOMAINS)
+  const allowedHostsExceptionsByDid = new Map<string, Set<string>>()
+  for (const live of ctx) {
+    allowedHostsExceptionsByDid.set(
+      live.did,
+      new Set(DEFAULT_LIVE_ALLOWED_DOMAINS.concat(live.domains)),
```
estrattonbailey: I realize it's redundant to add the defaults here too, but until we refactor this I want to avoid issues of failing to check for both `defaultAllowedHosts` and `allowedHostsExceptionsByDid` since that's implicit and not totally clear. I think we can improve this, just don't want to blow up this PR.

src/state/service-config.tsx
```diff
@@ -89,28 +89,35 @@ const DEFAULT_LIVE_ALLOWED_DOMAINS = [
   'www.bluecast.app',
 ]
 export type LiveNowConfig = {
-  allowedDomains: Set<string>
-  allSupportedDomains: Set<string>
+  currentAccountAllowedHosts: Set<string>
```
estrattonbailey, line 92: Renamed this too to make sure it's extremely clear what we're checking for here.

estrattonbailey:
@claude my boy, I got 5 hours of sleep last night, can you plz double check this one too for shits and gigs

PR #9785, opened 2026-01-28
[APP-1774] Handle z-index for FAB

vineyardbovines:
This is a bit of a weird one. tl;dr react-native-web automatically applies `zIndex: 0` to plain views/divs. The FAB element already had the appropriate z-index, but the hierarchical contains all contain `zIndex: 0` and so the FAB can't break out of the stacking context.

I tried the [isolation property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/isolation) to try and force the FAB to be in its own stacking context, but made no difference. The only thing that ended up working was applying the workaround that's mentioned in the linked discussion thread, and then excluding the animation container from that. So, this way, the `zIndex: 1` will be respected.

I don't super love this because I don't know the downstream implications of making this change- on smoke testing around on web, it _seems_ fine, but I'd want someone to check me on if the juice is worth the squeeze here.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -58,8 +58,8 @@ a[role='link'][data-no-underline='1']:hover {
 }
 /* Remove auto z-index applied by react-native-web for FAB */
 /* see: https://github.com/necolas/react-native-web/discussions/2547 */
-[class^="css-view-"]:not([data-testid="fab-outer"]) {
-    z-index: auto;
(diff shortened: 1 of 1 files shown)


vineyardbovines:
closing since its a pretty niche edge case, will re-visit if it comes up more frequently

mozzius:
yeah definitely open to revisiting this if we find the need for it

PR #9786, opened 2026-01-28
[APP-1761] Prevent welcome modal from reappearing

vineyardbovines:


src/components/hooks/useWelcomeModal.ts
```diff
@@ -21,14 +21,18 @@ export function useWelcomeModal() {
...
       const hasUserClosedModal =
-        sessionStorage.getItem('welcomeModalClosed') === 'true'
+        localStorage.getItem('welcomeModalClosed') === 'true'
+      const hasModalBeenShown =
+        localStorage.getItem('welcomeModalShown') === 'true'
 
-      if (isHomePage && !hasUserClosedModal) {
+      if (isHomePage && !hasUserClosedModal && !hasModalBeenShown) {
+        // Mark that the modal has been shown, don't show again
+        localStorage.setItem('welcomeModalShown', 'true')
```
mozzius: Lgtm, but logically, do we need the `welcomeModalClosed` one now? I can't think of a case where that would be false and `welcomeModalShown` would be true, I think it's redundant

vineyardbovines: this is why we don't let cursor make changes. fixed!

PR #9789, opened 2026-01-29
Add ESLint rule to enforce Lingui msg usage

mozzius:
Adds a custom ESLint rule 'lingui-msg-rule' that ensures the Lingui _()
function is called with msg`` template literals or plural/select macros,
preventing accidental misuse like _('string') which bypasses i18n.

https://claude.ai/code/session_01JMXXPUgAHiSBGmfGwUojKy


vineyardbovines (approved):
🤑

PR #9793, opened 2026-01-29
add expo/metro-runtime for web hot reloading

vineyardbovines:
image
 
IIRC Expo 54 disables hot reloading for web by default and this enables it

### DIFF
--- a/index.web.js
+++ b/index.web.js
@@ -1,3 +1,4 @@
+import '@expo/metro-runtime'
 import '#/platform/markBundleStartTime'
 import '#/platform/polyfills'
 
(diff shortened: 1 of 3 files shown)


index.web.js
```diff
@@ -1,3 +1,4 @@
+import '@expo/metro-runtime'
```
mozzius, line 1: I'm guessing we don't want this in the prod bundle?

```suggestion
if (__DEV__) require('@expo/metro-runtime')
```

mozzius: also might wanna leave a comment for posterity

PR #9800, opened 2026-01-30
show rich text in drafts list

mozzius:
Screenshot 2026-01-30 at 18 27 56


estrattonbailey (changes requested):
Gonna cherry pick this commit into the new preview 👍

PR #9801, opened 2026-01-30
remove webpack in favor of metro for web

vineyardbovines:
clearly a big one, but hopefully a big improvement. running web locally is a lot faster now, as is building it. random smoke testing around the local builds, everything seems in tact so far. 

**test steps**

- run `yarn web`: dev server starts and HMR works
  - dev server startup time is significantly faster. change any component to see HMR
- run `yarn build-web`: exports production, should populate `dist/` and `bskyweb/static`
- verify `scripts.html` is generated with the correct CDN paths
- run app from the go server

**misc todo**
- [ ] Replace `generate-webpack-stats-file`

CleanShot 2026-01-30 at 11 53 07


src/components/Dialog/index.web.tsx
```diff
@@ -10,7 +10,9 @@ import {
 } from 'react-native'
 import {msg} from '@lingui/macro'
 import {useLingui} from '@lingui/react'
-import {DismissableLayer, FocusGuards, FocusScope} from 'radix-ui/internal'
+import {DismissableLayer} from '@radix-ui/react-dismissable-layer'
```
vineyardbovines, line 21: because we have `cfg.resolver.unstable_enablePackageExports = false` in the metro config, i didn't really want to see what broke if i removed it. instead, i swapped out the package imports for radix-ui

mozzius: see here for context on why we added it

https://github.com/bluesky-social/social-app/pull/8219

package.json
mozzius, line 1: Looks like a bunch of indentation was added, maybe set your editor to "detect current indentation" or w/e

package.json
```diff
@@ -1,383 +1,381 @@
...
+        "intl:compile-if-needed": "is-ci || [ -f src/locale/locales/en/messages.js ] || yarn intl:compile",
+        "intl:pull": "crowdin download translations --verbose -b main",
+        "intl:push": "crowdin push translations --verbose -b main",
+        "intl:push-sources": "crowdin push sources --verbose -b main",
+        "nuke": "rm -rf ./node_modules && rm -rf ./ios && rm -rf ./android",
+        "update-extensions": "bash scripts/updateExtensions.sh",
+        "export": "npx expo export --dump-sourcemap && yarn upload-native-sourcemaps",
+        "upload-native-sourcemaps": "npx sentry-expo-upload-sourcemaps dist",
+        "make-deploy-bundle": "bash scripts/bundleUpdate.sh",
+        "generate-web-stats": "EXPO_ATLAS=1 yarn build-web",
```
mozzius: Oh cool, we should definitely configure expo-atlas to work for all platforms

package.json
vineyardbovines: me the other day: i'm gonna fix my dev env because its trashed
also me: fix the app dev env instead

mozzius:
@claude review

mozzius:
@claude re-review

PR #9809, opened 2026-02-02
Make GIFs look like GIFs

mozzius:
Tags `.gif` file uploads with a special presentation, which then changes the appearance of the video controls (simplifies them to match other gifs, mutes). This is via the new `presentation` field in the record.

I also made the web video player cope better with short videos (see the changes to looping)

https://github.com/user-attachments/assets/99302a22-fc4a-4e70-8c82-5226c70bd7a1

https://github.com/user-attachments/assets/86fcea27-71b2-4c9f-a1ee-5f2e7bc7ec08

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/index.web.tsx
+++ b/src/components/Post/Embed/VideoEmbed/index.web.tsx
@@ -98,7 +98,6 @@ export function VideoEmbed({embed}: {embed: AppBskyEmbedVideo.View}) {
             setActive={setActive}
             onScreen={onScreen}
             lastKnownTime={lastKnownTime}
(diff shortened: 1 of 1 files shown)


src/lib/api/index.ts
```diff
@@ -354,6 +354,8 @@ async function resolveMedia(
       alt: videoDraft.altText || undefined,
       captions: captions.length === 0 ? undefined : captions,
       aspectRatio,
+      presentation:
+        videoDraft.video.mimeType === 'image/gif' ? 'gif' : 'default',
```
mozzius, line 358: This is where we set the `presentation` field

PR #9814, opened 2026-02-03
Only treat animated gifs as videos, leave static gifs as images

mozzius:
Checks if a `image/gif` file is actually animated, and only treat it as a video if it has multiple frames. It will fail to upload as a video if it only has one frame with "Failed to get duration"

You can validate the `isAnimatedGif` function in isolation here: 
https://codesandbox.io/p/devbox/t8sjj7

These are my test gifs: [gif-examples.zip](https://github.com/user-attachments/files/25051855/gif-examples.zip)

https://github.com/user-attachments/assets/79bb11a7-f870-4dba-b3b9-27435f12efa8

https://github.com/user-attachments/assets/efc6c775-da92-4e4c-9f0b-4567933f46c2

# Test plan

Try out animated and non-animated gifs, on web and native

### DIFF
--- a/src/view/com/composer/SelectMediaButton.tsx
+++ b/src/view/com/composer/SelectMediaButton.tsx
@@ -130,9 +130,7 @@ const extensionToMimeType: Record<
  * `mimeType`. If `mimeType` is not available, we try to infer it through
  * various means.
  */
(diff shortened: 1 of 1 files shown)


vineyardbovines (approved):
schweet

PR #9831, opened 2026-02-05
[APP-1794] respect mute words for live event feed banners

vineyardbovines:
Utilize `hasMutedWord` to ensure live event feed banners with muted words don't display (with the exception of bsky team and dev mode). 

Tested with both a mock event (porsche) and the dev live event (ncaa) on web and mobile.

CleanShot 2026-02-05 at 10 03 26
CleanShot 2026-02-05 at 10 11 23


estrattonbailey (approved):
Nailed it 😎

PR #9838, opened 2026-02-06
Use post context to generate consistent filenames for saved images (fixes #7280)

### LINKED ISSUE #7280: Save Images With Consistent Names
### Describe the Feature

Instead of generating a random UUID for the image being being saved, save the image with a consistent timestamped name eg `BSKY_YYYYmmdd_HHMMSS` so that it is easier to search through them in a filesystem.

### Attachments

_No response_

### Describe Alternatives

_No response_

### Additional Context

_No response_

Igetin:
# What / Why

When saving images from the lightbox view, the filename was previously generic/non-deterministic (UUID), which made it hard to relate saved files back to the original post and search for them later in a filesystem.

This change makes saved post-image filenames deterministic based on post metadata.

This was originally requested in #7280.

# Changes

- derives a stable save name for image lightbox items in `ImageEmbed` from the owning post (`CommonProps.post`) that the embed layer already receives, so no new props or call-site changes are needed
- generates a `baseSaveName` for post images using the format: `handle_postRkey_imgN_postCreatedAtUtc`
- threads that optional `baseSaveName`  through the current lightbox save flow and media save pipeline
- preserves UUID fallback behavior when post context is unavailable
- applies the same filename behavior to the current web lightbox download path as well as native save flows

# Filename format

Example filename:

```
15delicious.bsky.social_3me7qo3xdyc23_img1_2026-02-06T223014.jpg
```

If the required post context is unavailable when determining the save name, the save flow falls back to a generated UUID filename. That still covers cases like saving a profile picture from a non-post lightbox entry point.

# Testing

I have tested this on my Android 15 device. It would be helpful if someone could also test this on an iOS device.

- Post with multiple images:
  1. Open the post
  2. Tap the first image to open lightbox
  3. Tap Save → swipe right (repeat for each image)
  4. Save one of the images a second time
- User profile picture:
  1. Navigate to a user's profile
  2. Tap the profile picture
  3. Tap Save
- Verify filenames:
  1. Open your gallery app and navigate to the Bluesky album
  2. Verify the post filenames match the format `${handle}_${postRkey}_img{N}_{postCreatedAtUtc}`
  3. On Android, verify that the post image that was saved twice has two files, with the newer one having the suffix ` (1)`
  4. Verify that the saved profile picture uses a UUID filename (such as `f94ff601-ff56-4bf1-9758-84fc49666ed3.jpg`)

### Other notes

Let me know if this should be behind a configuration option rather than default behavior.

---

Closes #7280.

### DIFF
--- a/src/components/Lightbox/Lightbox.tsx
+++ b/src/components/Lightbox/Lightbox.tsx
@@ -19,7 +19,7 @@ export function Lightbox() {
     <ImageView
       lightbox={activeLightbox}
       onRequestClose={onClose}
-      onPressSave={uri => void saveImageToAlbum(uri)}
+      onPressSave={opts => void saveImageToAlbum(opts)}
       onPressShare={uri => void shareImageModal({uri})}
     />
   )
--- a/src/components/Lightbox/Lightbox.web.tsx
+++ b/src/components/Lightbox/Lightbox.web.tsx
@@ -311,7 +311,10 @@ function LightboxGallery({
             <Menu.Item
               label={l`Download image`}
               onPress={() => {
-                saveImageToMediaLibrary({uri: img.uri}).then(
+                saveImageToMediaLibrary({
+                  uri: img.uri,
+                  baseSaveName: img.baseSaveName,
+                }).then(
                   () => {
                     Toast.show(l`Image saved`)
                   },
--- a/src/components/Lightbox/pager/ImagePager.tsx
+++ b/src/components/Lightbox/pager/ImagePager.tsx
@@ -85,7 +85,7 @@ export default function ImageViewRoot({
 }: {
   lightbox: Lightbox | null
   onRequestClose: () => void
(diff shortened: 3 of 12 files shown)


Igetin:
Rebased onto latest upstream (`bluesky-social:main`) and resolved conflicts by preserving the upstream save pipeline changes while keeping this PR’s deterministic filename behavior. I've also retested the new functionality as per the issue description.

Igetin:
Once again refreshed this branch onto current `main` and reapplied the feature through the current lightbox/embed architecture. I also reran the local typecheck and hook validation.

Igetin:
Rebased, tweaked, and retested the implementation for the 1.127.0 release. The embed layer now receives the owning post directly (`CommonProps.post`), so the save name is derived from it inside `ImageEmbed` instead of threading custom props through every call site. The diff is now several files smaller, but the behavior remains the same.

PR #9840, opened 2026-02-08
Drag image to start post on web

ezramechaber:
Enable users to drag an image onto a target to begin a post, rather than needing to open the "What's up" window first.

### DIFF
--- a/src/lib/api/feed-manip.ts
+++ b/src/lib/api/feed-manip.ts
@@ -335,6 +335,27 @@ export class FeedTuner {
     return slices
   }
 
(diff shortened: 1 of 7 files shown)


mozzius:
hey! thanks for the PR, couple things:

1. I think you've accidentally commited two things into this branch (hide reposts from user stuff?)
2. there's an existing system for handling this when the composer is open, ideally we'd just move that existing system out of the composer and open it if the composer isn't open already

ezramechaber:
yep, thanks @mozzius wrong branch. lemme close/clean.

PR #9850, opened 2026-02-10
[APP-1833] Handle errors on saving a draft over char limit

vineyardbovines:
Since we already show/display the character limit to the user, I figured best UX here would be to prompt the user to discard or go back and edit the post instead of offering them the option of saving a draft we know can't be saved, so the sheet has been updated to reflect that.

As a fallback, the existing error display will still get populated with the right message, on some off chance that someone finds a way to try to save an over-limit draft.

Bumps draft limit to 1k chars to allow for workshopping!

CleanShot 2026-02-10 at 12 01 51

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -569,7 +569,7 @@ export const ComposePost = ({
     if (tooLong) {
       setError(
         _(
-          msg`Post is too long to save as a draft. The maximum number of characters is ${MAX_DRAFT_GRAPHEME_LENGTH}.`,
+          msg`One or more posts are too long to save as a draft. ${plural(MAX_DRAFT_GRAPHEME_LENGTH, {one: 'The maximum number of characters is # character.', other: 'The maximum number of characters is # characters.'})}`,
         ),
       )
       return false


src/view/com/composer/Composer.tsx
```diff
@@ -543,7 +557,36 @@ export const ComposePost = ({
...
+  )
+
+  const checkDraftTextLength = React.useCallback((): boolean => {
+    const tooLong = composerState.thread.posts.some(
+      post => post.shortenedGraphemeLength > MAX_GRAPHEME_LENGTH,
+    )
+    if (tooLong) {
+      setError(
+        _(
+          msg`Post is too long to save as a draft. The maximum number of characters is ${MAX_GRAPHEME_LENGTH}.`,
```
surfdude29:
```suggestion
          msg`Your post is too long to save as a draft. ${plural(
            MAX_GRAPHEME_LENGTH,
            {
              other: 'The maximum number of characters is # characters.',
            },
          )}`,
```
nit: `Your post` might align better with the new string on line 1249? Also, plural formatting would be appreciated here to ensure that the string can be correctly localized in all languages 🙏

vineyardbovines:
CleanShot 2026-02-11 at 13 00 06
CleanShot 2026-02-11 at 13 00 13

Updated discard sheet for 1k limit

src/view/com/composer/drafts/DraftsButton.tsx
```diff
@@ -17,14 +17,16 @@ export function DraftsButton({
   isEmpty,
   isDirty,
   isEditingDraft,
+  canSaveDraft,
   textLength,
 }: {
   onSelectDraft: (draft: DraftSummary) => void
-  onSaveDraft: () => Promise<void>
+  onSaveDraft: () => Promise<boolean>
```
estrattonbailey: Nit: dig the thinking here, but I think it'd be a little more clear what the boolean means if the save method returned a result object like `{ success: boolean }`. Whatcha think?

src/view/com/composer/Composer.tsx
```diff
@@ -275,6 +277,19 @@ export const ComposePost = ({
...
+  // Clear error when composer content changes, but only if all posts are
+  // back within the character limit.
+  const allPostsWithinLimit = thread.posts.every(
+    post => post.richtext.graphemeLength <= MAX_DRAFT_GRAPHEME_LENGTH,
+  )
+  useEffect(() => {
+    if (error && allPostsWithinLimit) {
+      setError('')
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
estrattonbailey: They truthy error check should handle this right? Works fine like this too just looks sketchy :P

src/view/com/composer/Composer.tsx
```diff
@@ -543,7 +558,36 @@ export const ComposePost = ({
     revokeAllMediaUrls()
   }, [closeComposer, queryClient])
 
+  const getDraftSaveError = React.useCallback(
+    (e: unknown): string => {
+      if (e instanceof AppBskyDraftCreateDraft.DraftLimitReachedError) {
+        return _(msg`You've reached the maximum number of drafts`)
+      }
```
estrattonbailey, line 559: Very nice ✨

src/view/com/composer/Composer.tsx
```diff
@@ -543,7 +558,36 @@ export const ComposePost = ({
...
+      return _(msg`Failed to save draft`)
+    },
+    [_],
+  )
+
+  const checkDraftTextLength = React.useCallback((): boolean => {
```
estrattonbailey: Nit: I read the use cases first and was confused for a sec. Maybe a better name for this would be `checkDraftTextLengthOrError` or `validateDraftTextOrError` to make it super obvious.

src/view/com/composer/Composer.tsx
```diff
@@ -569,18 +613,43 @@ export const ComposePost = ({
...
+        composerState,
+        existingDraftId: composerState.draftId,
+      })
+      composerDispatch({type: 'mark_saved', draftId: result.draftId})
+      return true
+    } catch (e) {
+      logger.error('Failed to save draft', {error: e})
+      setError(getDraftSaveError(e))
+      return false
+    }
```
estrattonbailey, line 640: Great add! Though actually for this case and the one above (which we missed before), you can drop the `logger.error` since it looks like the mutation handles this already.

src/view/com/composer/Composer.tsx
estrattonbailey: Hmmm actually, this clears any error, not just draft errors. I don't see any cases where that's a real issue right now, but since there's an X on the error message to clear it, and it clears on next submit or on next draft save (of a valid draft), I think we could remove this effect entirely. What do you think?

estrattonbailey:
Oh btw had to rebase bc I had some long drafts saved and hit the validation issue from earlier ✌️

src/view/com/composer/drafts/DraftsButton.tsx
vineyardbovines: agreed!

src/view/com/composer/Composer.tsx
vineyardbovines: went with `validate` - easier to reason with 👌

src/view/com/composer/Composer.tsx
vineyardbovines: yeah i think it's fine to remove, i can't think of a good enough edge case to have it hang around

src/view/com/composer/Composer.tsx
estrattonbailey: Ah yeah good call. Also maybe `One or more posts are too long to save as a draft.` would be a little more accurate, since it could be a thread 🤔

surfdude29:
@vineyardbovines @estrattonbailey Thanks for adding the pluralization 🙏

One tiny thing to mention for the future: my reading of Lingui docs is that when the context means that only the `other` prop is required in English – e.g. in this case, because `MAX_GRAPHEME_LENGTH` will logically always be greater than 1 – the `one` prop can simply be omitted :)

When a translator for a language with more complex plural forms is translating the string, Crowdin will automatically populate the translation with the plural forms for that language.

My understanding on this is set out in this comment: https://github.com/bluesky-social/social-app/pull/7984#issuecomment-2722259409

PR #9853, opened 2026-02-10
Use Tanstack Query isPending over isLoading

ds-boyce:
The `isLoading` value from the `useQuery` hook is initially `false` if `initialData` is provided, as we do in `useResolveDidQuery`, resulting in reaching the `should never happen` case in `Profile.tsx` on the initial render.

Conversely, the `isPending` field is `true` even if `initialData` is provided. By switching from `isLoading` to `isPending` we avoid the single render of the `should never happen` case.

(You’ll have to scrub the videos to see the change as it happens very fast.)

https://github.com/user-attachments/assets/aac94fa3-9cab-4095-8b32-1a106c0cfe16

https://github.com/user-attachments/assets/5d39b17a-429b-4244-89a3-279b52677053

### DIFF
--- a/src/view/screens/Profile.tsx
+++ b/src/view/screens/Profile.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useMemo} from 'react'
+import {useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import {StyleSheet} from 'react-native'
 import {SafeAreaView} from 'react-native-safe-area-context'
 import {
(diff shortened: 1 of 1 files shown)


src/view/screens/Profile.tsx
```diff
@@ -79,14 +79,14 @@ function ProfileScreenInner({route}: Props) {
     data: resolvedDid,
     error: resolveError,
     refetch: refetchDid,
-    isLoading: isLoadingDid,
+    status: didStatus,
```
mozzius: can we use `isPending` instead of the status enum? that's what we typically use elsewhere

ds-boyce: yep, will do

PR #9855, opened 2026-02-11
Simplify Android build: bundletool + eas submit

mozzius:
This simplifies the Android build pipeline to make it more like iOS. We now auto-submit to google play - `android-testflight` profile submits automatically to the google play internal track and ditto for `production`, and we now use `bundletool` to produce the APK that gets used for github releases which means we don't have to run the build process twice.

One concern: the QA team uses the internal track. We might want to consider moving them to a different track (maybe the "alpha" track?) so that we can control better which build they get

## Summary
- Remove redundant second EAS build (`production-apk` profile) — use `bundletool` to generate a signed universal APK from the AAB instead
- Add `eas submit` to auto-submit to Google Play for both `production` and `testflight-android` profiles
- Unify workflow so both profiles follow the same path: **Build AAB → eas submit → bundletool APK → upload artifact → Slack**

## Changes
**`eas.json`:**
- Remove `production-apk` build profile
- Remove `distribution: internal` from `testflight-android` (so it produces an AAB)
- Add Android submit configs (`production` → production track, `testflight-android` → internal track)

**`.github/workflows/build-submit-android.yml`:**
- Remove: rename step, conditional artifact uploads, conditional Slack notifications, Metro cache clear, entire second APK build + its upload + notification
- Add: `eas submit`, bundletool setup (via `amyu/setup-bundletool`), keystore decode, signed universal APK generation, single artifact upload, unified Slack notifications

## Setup required
4 new GitHub Actions secrets must be added before merging:
- `ANDROID_KEYSTORE_BASE64` — base64-encoded keystore
- `ANDROID_KEYSTORE_PASSWORD`
- `ANDROID_KEY_ALIAS`
- `ANDROID_KEY_PASSWORD`

`eas submit` also requires a Google Play service account key configured in EAS.

## Test plan
- [x] Add the 4 keystore secrets to the repo
- [x] Trigger workflow with `testflight-android` on this branch to verify:
  - AAB is built (not APK)
  - `eas submit` succeeds (internal track)
  - bundletool produces a valid signed universal APK
  - Slack notifications fire with correct info
- [x] Trigger workflow with `production` to verify production track submission


mozzius:
I tried running it and it failed. Claude recommends this as a fix:

## Fix for `eas submit` Google Service Account error

The `eas submit -p android` step fails because EAS has no Google Play Service Account credentials configured, and can't prompt in `--non-interactive` mode.

### Prerequisites (manual)

Add a GitHub secret `GOOGLE_SERVICE_ACCOUNT_KEY` containing the raw JSON of a Google Cloud service account key with Google Play Developer API access.

### Code changes needed

**1. `eas.json`** — add `serviceAccountKeyPath` to both Android submit profiles:

```json
"production": {
  "android": {
    "track": "production",
    "serviceAccountKeyPath": "./google-service-account-key.json"
  }
},
"testflight-android": {
  "android": {
    "track": "internal",
    "serviceAccountKeyPath": "./google-service-account-key.json"
  }
}
```

**2. `.github/workflows/build-submit-android.yml`** — add this step before "Submit to Google Play":

```yaml
- name: 🔑 Decode Google Service Account Key
  run: echo '${{ secrets.GOOGLE_SERVICE_ACCOUNT_KEY }}' > google-service-account-key.json
```

**3. `.github/workflows/bundle-deploy-eas-update.yml`** — same step before "Submit to Google Play" in the `buildIfNecessaryAndroid` job.

**4. `.gitignore`** — add `google-service-account-key.json` (currently only `google-services.json` is ignored).

estrattonbailey (approved):
Tested both and production track worked just fine!

estrattonbailey (approved):
🚀

PR #9861, opened 2026-02-11
[APP-1805] Analytics tweaks and tracking callbacks

estrattonbailey:


src/geolocation/util.ts
```diff
@@ -125,3 +126,21 @@ export function mergeGeolocations(
   })
   return geolocation
 }
+
+/**
+ * Get's the merged geolocation as a string in the format of
+ * "countryCode-regionCode", or just "countryCode" if regionCode is not
+ * available.
+ */
+export function getGeolocationString() {
```
estrattonbailey: Unused artifact I copied from another codebase, just left it here bc I bet we use it at some point

src/geolocation/util.ts
```diff
@@ -125,3 +126,26 @@ export function mergeGeolocations(
   })
   return geolocation
 }
+
+/**
+ * Get's the IP-based geolocation as a string in the format of
```
surfdude29:
```suggestion
 * Gets the IP-based geolocation as a string in the format of
```
nit

src/analytics/index.tsx
```diff
@@ -181,20 +181,25 @@ export function AnalyticsFeaturesContext({
...
+      variationId: result.key,
+    })
+  })
+  feats.setFeatureUsageCallback((feature, result) => {
+    parentContext.metric('feature:viewed', {
+      featureId: feature,
+      featureResultValue: result.value,
+      experimentId: result.experiment?.key,
+      variationId: result.experimentResult?.key,
     })
```
mozzius, line 201: Is there a reason to not put this in a useEffect?

I'd be slightly concerned react compiler might try and optimise this, as it's an unexpected side effect

estrattonbailey: Yeah there's a comment to this effect just above — will double check on the compiler

estrattonbailey: Should be good here in terms of the compiler. Agreed it's a little icky in general, but in the grand scheme it's alright and let's us eval gates with all the correct metadata on first render without need to set a `ready` flag or something and re-eval on a second pass.

PR #9863, opened 2026-02-11
Refactor delete account dialog using ALF

ds-boyce:
Replaces the old delete account dialog with an ALF one, including new UX.

| iOS    | Web |
| - | - |
| email | email-web |
| code | code-web |
| code-error | code-error-web |
| confirm | confirm-web |

### DIFF
--- a/src/screens/Settings/components/DeleteAccountDialog.tsx
+++ b/src/screens/Settings/components/DeleteAccountDialog.tsx
@@ -325,7 +325,7 @@ function DeleteAccountDialogInner({
             <Prompt.Cancel />
           </Prompt.Actions>
           {error && (
-            <Admonition style={[a.mt_lg, a.mb_lg]} type="error">
+            <Admonition style={[a.mt_lg]} type="error">
               <Text style={[a.flex_1, a.leading_snug]}>{error}</Text>
             </Admonition>
           )}


src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
+import {useCallback, useRef, useState} from 'react'
+import {Text as RNText, type TextInput, View} from 'react-native'
+import {msg, Trans} from '@lingui/macro'
+import {useLingui} from '@lingui/react'
+
+import {DM_SERVICE_HEADERS} from '#/lib/constants'
+import {cleanError} from '#/lib/strings/errors'
```
estrattonbailey: Nit: we have a new-ish hook for this that works better with our i18n library, `#/lib/hooks/useCleanError`. Same API, just updates when language setting changes.

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+import {createStaticClick, InlineLinkText} from '#/components/Link'
+import {Loader} from '#/components/Loader'
+import * as Prompt from '#/components/Prompt'
+import * as toast from '#/components/Toast'
+import {Span, Text} from '#/components/Typography'
+import {resetToTab} from '#/Navigation'
+import {Divider} from './SettingsList'
+
+const WHITESPACE_RE = /\s/gu
+const CONFIRM_CODE_LENGTH = 11
```
estrattonbailey: I see you've got `TokenField` in here, which is perf. I think we might as well use the utils from that file too for normalization and validity checks.

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+              <Span style={[a.font_semi_bold, t.atoms.text]}>
+                {currentEmail}
+              </Span>
+              .
+            </Trans>
+          </Prompt.DescriptionText>
+          <View style={[a.gap_lg, a.mb_lg]}>
+            <Button
+              color="primary"
+              size={gtMobile ? 'small' : 'large'}
```
estrattonbailey: We can use `Prompt.Action` for all buttons other than the `Cancel`. And the actions can be wrapped with `Prompt.Actions` component. Some of the older dialogs probably haven't been updated, apologies for the bad references :P

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+                a.gap_sm,
+                a.p_md,
+                a.rounded_sm,
+                t.atoms.bg_contrast_25,
+                a.mb_lg,
+              ]}>
+              <CircleInfo size="md" fill={t.palette.negative_400} />
+              <Text style={[a.flex_1, a.leading_snug]}>{error}</Text>
+            </View>
+          )}
```
estrattonbailey, line 216: We can use an `Admonition` here too, like you have below

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+          <Admonition type="tip">
+            <Trans>
+              You can also{' '}
+              <RNText
+                style={[{color: t.palette.primary_500}, web(a.underline)]}
+                onPress={handleDeactivate}>
+                temporarily deactivate
+              </RNText>{' '}
```
estrattonbailey: No reason you'd know this, but we have a little-used `Span` export from `#/components/Typography` for cases like this. Just a convenience thing to be aware of 👍

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+                !isConfirmCodeValid(confirmCode) || !isPasswordValid(password)
+              }
+              size={gtMobile ? 'small' : 'large'}
+              label={_(msg`Delete My Account`)}
+              onPress={handleDeleteAccount}>
+              <ButtonText>{_(msg`Delete My Account`)}</ButtonText>
+            </Button>
+            <Prompt.Cancel />
+          </View>
+          {error && (
```
estrattonbailey, line 327: Admonition here too 👍

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+
+  const sendEmail = useCallback(async () => {
+    if (emailState === EmailState.PENDING) {
+      return
+    }
+    try {
+      setError('')
+      setEmailState(EmailState.PENDING)
+      await agent.com.atproto.server.requestAccountDelete()
+      setEmailState(EmailState.DEFAULT)
```
estrattonbailey: Nit: could chuck these in the `finally` block

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+                {currentEmail}
+              </Span>{' '}
+              for an email with the confirmation code to enter below:
+            </Trans>
+          </Prompt.DescriptionText>
+          <View style={[a.mb_xs]}>
+            <TextField.LabelText>
+              <Trans>Confirmation Code</Trans>
+            </TextField.LabelText>
+            <TokenField
```
estrattonbailey, line 253: Let's make sure these fields don't autocomplete. `TokenField` should probably never autofill (rn it seems to use my email). Password could maybe autofill, maybe something to discuss, but not a bad idea to ask for the password again for this use case to ensure they don't make a mistake imo.

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+              t.atoms.text_contrast_medium,
+            ]}>
+            <Trans>
+              <InlineLinkText
+                label={_(msg`Resend`)}
+                {...createStaticClick(() => {
+                  void handleSendEmail()
+                })}>
+                Click here to resend.
+              </InlineLinkText>
```
estrattonbailey: Would be rad to include a little success state like we have for the `EmailDialog`. Kinda hard to tell what's happened here. Could also add a `Toast`, but since this is already inline, would be nicer imo to do the success state inline too.

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+              <TextField.Icon icon={Lock} />
+              <TextField.Input
+                inputRef={passwordRef}
+                testID="newPasswordInput"
+                label={_(msg`Enter your password`)}
+                autoCapitalize="none"
+                autoCorrect={false}
+                returnKeyType="done"
+                secureTextEntry={true}
+                autoComplete="off"
```
estrattonbailey, line 307: Weird that this isn't working 🤔 maybe a RN-web issue?

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+                clearButtonMode="while-editing"
+                passwordRules={`minlength: ${PASSWORD_MIN_LENGTH}};`}
+                value={password}
+                onChangeText={setPassword}
+                onSubmitEditing={handleDeleteAccount}
+                accessibilityHint={_(msg`Input your password`)}
```
estrattonbailey: `hint` here is kinda redundant. Not required here imo unless we wanna add additional details.

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+                a.p_md,
+                a.rounded_sm,
+                t.atoms.bg_contrast_25,
+                a.mb_lg,
+              ]}>
+              <CircleInfo size="md" fill={t.palette.negative_400} />
+              <Text style={[a.flex_1, a.leading_snug]}>{error}</Text>
+            </View>
+          )}
+          <Divider />
```
estrattonbailey: Looks like a stray component here

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,363 @@
...
+  const currentEmail = currentAccount?.email ?? '(no email)'
+
+  switch (step) {
+    case Step.SEND_CODE:
+      return (
+        <>
+          <Prompt.TitleText>
+            {_(msg`Delete account “${currentHandle}”`)}
+          </Prompt.TitleText>
+          <Prompt.DescriptionText>
```
estrattonbailey: We actually have a new component now to wrap these two components, `Prompt.Content`. Adds a little spacing is all. Should do for all of these.

estrattonbailey (commented):
Looking good! Mostly small stuff and patterns that aren't entirely obvious from our old code 😅 

I think we should also try to squeeze in a "Already have a code?" link on the initial screen. Rn if you close the dialog accidentally, you have to send another email. `EmailDialog` has an example of this. May need to consult Darrin to make it look good, I realize there's a lot on this initial screen lol.

surfdude29:
As part of this PR, could the copy please be updated to make clear to users that the deletion process on Bluesky affects their entire ATProto account, not just their presence on Bluesky.

Please see the discussion about this issue raised by @oppiliappan here:

https://bsky.app/profile/oppi.li/post/3menfmvakcc2x

Maybe something like a warning admonition that says:

**"This will delete all records and data associated with your ATProto account, including from non-Bluesky services."**

heikadog:
> As part of this PR, could the copy please be updated to make clear to users that the deletion process on Bluesky affects their entire ATProto account, not just their presence on Bluesky.
> 
> Please see the discussion about this issue raised by @oppiliappan here:
> 
> https://bsky.app/profile/oppi.li/post/3menfmvakcc2x
> 
> Maybe something like a warning admonition that says:
> 
> **"This will delete all records and data associated with your ATProto account, including from non-Bluesky services."**

i'm saying this as someone commonly seen as an ATProto dev: people might not know what ATProto means in this context? a lot of people know Bluesky The App and not the bit that makes it decentralized, or they don't care. 

i'd change it to say something like 
> "This will delete your Bluesky account and all associated data. Because Bluesky uses the AT Protocol, this will also affect any other services you use with this account." 

that way people who don't use extra atproto services have the information they need and people who do have a clarification

src/screens/Settings/components/DeleteAccountDialog.tsx
ds-boyce: looks like `Prompt.Action` only accepts a string label, and i have a `ButtonIcon` here to display the  `Loader` or `Envelope` icon based on state.

ok to use this one as-is (and update the other two) and maybe consider an `icon` prompt for `Prompt.Action` as a follow-up?

ds-boyce: ah, similarly the second button has a `disabled` state until the fields are valid. same question. 🙂

src/screens/Settings/components/DeleteAccountDialog.tsx
ds-boyce: this appears to swallow the error since the `finally` block always executes.

surfdude29:
@heikadog

> i'd change it to say something like
> 
> > "This will delete your Bluesky account and all associated data. Because Bluesky uses the AT Protocol, this will also affect any other services you use with this account."
> 
> that way people who don't use extra atproto services have the information they need and people who do have a clarification

I think you're on the right track with that. I thought some more about this, asked Claude for some feedback on my suggestion (especially in terms of how easy less-technically inclined users would be able to understand it), and I've ended up with this suggested wording:

> This will permanently delete your account and all associated user data – not just on Bluesky.
> 
> Your Bluesky account is also your identity on other services that use the same AT Protocol network. Deleting your account here will permanently and irreversibly delete your data on those services too.

One point I'd like to highlight: I think it's especially important to emphasize the permanence of account deletion, as @bnewbold said [here](https://bsky.app/profile/bnewbold.net/post/3mengkqw4nk2i) that this aspect surprises some people:

> I suspect another aspect to some of these situations is that many folks are accustomed to account "deletion" being more like "deep hibernation"; while PDS actually just does real deletion. this isn't everybody, or the main issues, but it does surprise a fraction of people

src/screens/Settings/components/DeleteAccountDialog.tsx
estrattonbailey: Oh my bad, sleepy brain last night. Yeah sg, let's just match up the sizing then 👍

heikadog:
oh i love that suggestion

src/screens/Settings/components/DeleteAccountDialog.tsx
ds-boyce: ha for case three, `Prompt.Action` calls `close()` internally, so if there’s an error it closes the dialog instead of returning to `Step.VERIFY_CODE` to display the error message.

maybe `onPress` in `Prompt.Action` should conditionally do that? something we can talk about.

src/screens/Settings/components/DeleteAccountDialog.tsx
estrattonbailey: Oh sorry again, not clear at all, just meant the two calls to `setEmailState(EmailState.DEFAULT)`. Really nit picky though nbd 👍

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,373 @@
...
+              onPress={handleSendEmail}>
+              <ButtonText>{_(msg`Send Email`)}</ButtonText>
+              <ButtonIcon
+                icon={emailState === EmailState.PENDING ? Loader : Envelope}
+              />
+            </Button>
+            <Prompt.Cancel />
+          </Prompt.Actions>
+          {error && (
+            <Admonition style={[a.mt_lg, a.mb_lg]} type="error">
```
estrattonbailey: Nit: can drop the `mb_lg` here since it compounds with the `mt_lg` on the Admonition below

src/screens/Settings/components/DeleteAccountDialog.tsx
ds-boyce: ah yeah i cleaned that one up 🙂

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,373 @@
...
+        message: 'Failed to delete account',
+      })
+      setConfirmCode('')
+      setPassword('')
+      setStep(Step.VERIFY_CODE)
+    }
+  }, [
+    _,
+    agent.api.chat.bsky.actor,
+    agent.com.atproto.server,
```
mozzius: should just be `agent` here

ds-boyce: good catch

src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -0,0 +1,371 @@
...
+              onPress={handleSendEmail}>
+              <ButtonText>{_(msg`Send email`)}</ButtonText>
+              <ButtonIcon
+                icon={emailState === EmailState.PENDING ? Loader : Envelope}
+              />
+            </Button>
+            <Prompt.Cancel />
+          </Prompt.Actions>
+          {error && (
+            <Admonition style={[a.mt_lg, a.mb_lg]} type="error">
```
estrattonbailey: Nit: could rm the `mb_lg` here since it stacks on top of `mt_lg` below

estrattonbailey (approved):
Nice! LGTM!

src/screens/Settings/components/DeleteAccountDialog.tsx
ds-boyce: 🦅 👀

PR #9864, opened 2026-02-12
Fix keyboard flicker when selecting a draft in composer

mozzius:
When selecting a draft, the drafts sheet closes and the underlying text
input regains focus, showing the keyboard. Then the state update causes
the ComposerPost to remount (key changes), hiding the keyboard. Finally,
the new component mounts with autoFocus, showing it again. This
show-hide-show cycle is visible as a flicker.

Fix by calling Keyboard.dismiss() at the start of handleSelectDraft so
the keyboard stays hidden during the transition and only appears once
when the new component mounts.

https://claude.ai/code/session_01MXNaTWzQNCx8fKR1fxKi6v

## Before

https://github.com/user-attachments/assets/62652c63-8b86-4a17-80bf-b1341ef1370b

## After

https://github.com/user-attachments/assets/f943fbf8-8c14-4b8f-9d78-3477315fe088

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -477,12 +477,6 @@ export const ComposePost = ({
 
   const handleSelectDraft = React.useCallback(
     async (draftSummary: DraftSummary) => {
-      // Dismiss keyboard immediately to prevent flicker. Without this,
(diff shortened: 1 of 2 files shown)


mozzius:
@github-actions ota

vineyardbovines (approved):

PR #9866, opened 2026-02-12
[APP-1829] Show "media on other device" for threaded media

vineyardbovines:
`hasMissingMedia` depends on `storage.mediaExists()`, which checks whether media files exist locally. using `hasMedia` instead ensures that if it's not the originating device and the draft has media, by definition it's on the other device.

### DIFF
--- a/src/view/com/composer/drafts/DraftItem.tsx
+++ b/src/view/com/composer/drafts/DraftItem.tsx
@@ -37,7 +37,7 @@ export function DraftItem({
   const post = draft.posts[0]
 
   const mediaExistsOnOtherDevice =
-    !draft.meta.isOriginatingDevice && draft.meta.hasMissingMedia
+    !draft.meta.isOriginatingDevice && draft.meta.hasMedia
   const mediaIsMissing =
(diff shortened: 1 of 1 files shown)


mozzius:
This makes sense to me

@claude am I missing anything here

mozzius:
@claude hmm this explanation doesn't work for me, why doesn't the existing approach work? and does hasMedia apply to the whole thread?

vineyardbovines:
@claude here is the ticket context:

"Drafts show a "Media exists on another device" if there's media in the first post of the draft. If a threaded post (with media) exists further down, we don't label the draft as having media, which is not ideal. Just need to check if any posts contain media on another device and use that to show the notice."

mozzius:
@claude are you *absolutely* sure there's no other circumstance where that message could go missing

mozzius:
Actually that means this PR is wrong then, as we definitely don't want GIFs to count here

estrattonbailey:
I thought we reproduced this, but I can't seem to now and yeah the logic seems fine as is 🤔

PR #9869, opened 2026-02-13
Persist navigation state between app opens

mozzius:
# Stacked on #9497 

https://reactnavigation.org/docs/state-persistence

https://github.com/user-attachments/assets/ffb90f95-635a-4929-9d7d-9ae9328dc373

# Test plan

- Ensure it works
- Ensure if you quit twice in a row, it clears it properly (important, in case the screen errors and it restores to the same screen. don't want a bootloop situation 😖

### DIFF
--- a/src/Navigation.tsx
+++ b/src/Navigation.tsx
@@ -887,6 +887,11 @@ function RoutesContainer({children}: React.PropsWithChildren<{}>) {
   const emailDialogControl = useEmailDialogControl()
   const closeAllActiveElements = useCloseAllActiveElements()
   const linkingUrl = Linking.useLinkingURL()
+  const [initialNotificationResponse, setInitialNotificationResponse] =
+    useState(() => {
+      if (!IS_NATIVE) return null
+      return Notifications.getLastNotificationResponse()
+    })
 
   const [initialState] = useState(() => {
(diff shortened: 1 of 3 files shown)


estrattonbailey (commented):
Should we put an expiry on this? I think we probably should. Coming back a day later and re-opening the app that was evicted from memory to the same screen actually seems wrong to me, but that may be subjective.

surfdude29:
My unasked for two cents: yes, some kind of expiry seems like a good idea. I would suggest starting with something like 6 hours and then potentially adjusting if there's user feedback that it's too long. Expiry after 6 hours would cover most of those occasions where you move onto doing other stuff, get distracted etc but then want to return to where you were on Bluesky.

However, I can equally see a case that an expiry in the region of 2-4 hours would probably be fine for most people.

estrattonbailey:
Yeah there's probably something reasonable here. My personal preference would be like 30 min maybe?

One thing I _loathe_ is how Instagram resets after like 60 seconds away. Abominable user experience imo.

surfdude29:
> One thing I _loathe_ is how Instagram resets after like 60 seconds away. Abominable user experience imo.

Yes, truly awful

PR #9871, opened 2026-02-13
[APP-1776] Refactor Live Now config, reorg files

estrattonbailey:
This is a pure refactor up until [Sanitize URLs to compare just apex domain](https://github.com/bluesky-social/social-app/commit/43d825b32275f2da3163d4cb308f58e1e55a1af1). I recommend reviewing by stepping through each commit.

The new config doesn't specify subdomains, so any subdomain under say, `youtube.com`, is valid. Special carve out for `nba.smart.link` since we aren't sure who owns that. That's the only real functional change here, aside from the actual loading of the config, which follows patterns we already use for Live Events and geolocation data.

### DIFF
--- a/src/state/appConfig.tsx
+++ b/src/state/appConfig.tsx
@@ -1,7 +1,6 @@
 import {createContext, useContext} from 'react'
 import {QueryClient, useQuery} from '@tanstack/react-query'
 
-import {networkRetry} from '#/lib/async/retry'
 import {APP_CONFIG_URL} from '#/env'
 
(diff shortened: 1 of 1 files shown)


mozzius:
@claude can you give this a look? the restructuring lgtm though, I like it

estrattonbailey:
@mozzius Claude-o made a fair point on the response reuse. I was double checking the retry handling and decided to just rely on RQ for that and added a custom retry rule. Fuzz tested in browser, should be better now all around.

PR #9877, opened 2026-02-15
Use video thumbnail as og:image for posts with video embeds

mozzius:
Since HLS streams can't use og:video, this uses the video's thumbnail image as the og:image instead, giving video posts a proper preview card. Handles both standalone video embeds and videos in recordWithMedia embeds.

https://claude.ai/code/session_01QCXR2u2dVFUAqdKmnryrZX


mozzius:
Well, I say they can't use og:video - we could totally put them in there and just mark the mime type

mozzius:
Seems to work in dev

Screenshot 2026-02-16 at 12 51 01

PR #9882, opened 2026-02-16
Fix missing pixel row in text rendering on iOS

### LINKED ISSUE #9836: Content rendered with missing horizontal lines (iOS app)
### Steps to Reproduce

1. View this post in iOS app https://bsky.app/profile/gingerbeardman.com/post/3me4oyfx5sk2d
2. Look at the line containing "Tower Street"
3. There is a missing row of pixels that causes the bottom of the characters in that line to be shortened (as if the image is compressed vertically but not horizontally) 
4. Compare with website 

I'm not sure how often this happens, I've not looked for further instances 

### Attachments

Zoomed image of the issue 

Image

iOS app 
Image

Website 

Image

iOS

### Device Info

iPhone 16 Pro (iOS 18.7.3)

Build version: 1.115.0.1280; Bundle info: d6b6b18 (prod); Bundle date: 26012722; Platform: ios; Platform version: 18.7.3; Device ID: 909d495a-38bd-4b25-b21e-10c853017fec

### Additional Information

_No response_

mozzius:
## Summary

Fixes #9836

`react-native-uitextview` calculates the text container height using `boundingRect`, which returns fractional values. Apple's docs recommend ceiling the result. Without it, `clipsToBounds` can clip through a line of text, producing a missing row of pixels.

This patch ceils `finalHeight` in `getNeededSize()` before it's stored and converted to `Float` (a lossy `CGFloat` → `Float` conversion that can further reduce the value).

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>Screenshot 2026-02-16 at 13 49 27</td>
<td>Screenshot 2026-02-16 at 13 52 29</td>
</tr>
</table>

## Test plan

- [ ] Open a post with multi-line text on iOS (e.g. [this post](https://bsky.app/profile/samuel.fm/post/3mexxt34vk22r))
- [ ] Verify no lines have missing pixel rows
- [ ] Verify text layout is otherwise unchanged

### DIFF
--- a/patches/react-native-uitextview+1.4.0.patch
+++ b/patches/react-native-uitextview+1.4.0.patch
@@ -1,11 +1,11 @@
 diff --git a/node_modules/react-native-uitextview/ios/RNUITextViewShadow.swift b/node_modules/react-native-uitextview/ios/RNUITextViewShadow.swift
-index c34ba71..13d576a 100644
+index c34ba71..3602856 100644
 --- a/node_modules/react-native-uitextview/ios/RNUITextViewShadow.swift
 +++ b/node_modules/react-native-uitextview/ios/RNUITextViewShadow.swift
-@@ -159,13 +159,23 @@ class RNUITextViewShadow: RCTShadowView {
+@@ -159,13 +159,25 @@ class RNUITextViewShadow: RCTShadowView {
      let maxSize = CGSize(width: CGFloat(maxWidth), height: CGFloat(MAXFLOAT))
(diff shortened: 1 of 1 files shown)


mozzius:
Ignore CI failure, false negative

vineyardbovines (approved):
i might be too blind to see the difference here but logically this makes sense

mozzius:
It very much might depend on the device model - I was able to repro and confirm by making the iPhone 17 sim really really big

Screenshot 2026-02-16 at 15 40 41

PR #9888, opened 2026-02-16
Fix typo in `DeleteAccountDialog` message

surfdude29:
As a follow-up to #9863, a small typo fix:

`irreversably` ➡️ `irreversibly`

PR #9889, opened 2026-02-16
Use Prompt.Action for delete account buttons

ds-boyce:
Switches from custom `Button` components to `Prompt.Action` using new props from #9887.


src/screens/Settings/components/DeleteAccountDialog.tsx
surfdude29:
```suggestion
                This will irreversibly delete your Bluesky account{' '}
```
nit: Maybe the typo fix from #9888 could be squeezed into this PR, and then I'll close that one? 🙏

PR #9890, opened 2026-02-16
Improve error handling on Explore page

ds-boyce:
The Explore screen would indefinitely display the placeholder for the **Discover new feeds** section in some error cases (see **Before** screenshot). This adds [a check](https://github.com/bluesky-social/social-app/pull/9890/changes/3bdb80f9fb62302d5df0d904c04c6b2a67c9da0a) for `suggestedFeedsError` to ensure we display an appropriate error message instead.

| Before | After |
| - | - |
| before | after |

### DIFF
--- a/src/screens/Search/Explore.tsx
+++ b/src/screens/Search/Explore.tsx
@@ -362,7 +362,7 @@ export function Explore({
     i.push({
       type: 'tabbedHeader',
       key: 'suggested-accounts-header',
-      title: _(msg`Suggested Accounts`),
+      title: _(msg`Suggested accounts`),
       icon: Person,
       searchButton: {
         label: _(msg`Search for more accounts`),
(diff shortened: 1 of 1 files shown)


src/screens/Search/Explore.tsx
```diff
@@ -78,10 +78,14 @@ function LoadMore({item}: {item: ExploreScreenItems & {type: 'loadMore'}}) {
   const t = useTheme()
   const {_} = useLingui()
 
+  const handleOnPress = () => {
+    void item.onLoadMore()
+  }
```
estrattonbailey, line 83: Not a problem, but what's your thinking when pulling out these handlers?

ds-boyce: lint was complaining about the unhandled promise. i could have inlined it, i’m just in the practice of not inlining functions, objects, and arrays as prop values.

PR #9892, opened 2026-02-17
Log onboarding errors for suggested accounts

ds-boyce:
This will give us insight to errors when fetching and following suggested accounts.

### DIFF
--- a/src/screens/Onboarding/StepSuggestedAccounts/index.tsx
+++ b/src/screens/Onboarding/StepSuggestedAccounts/index.tsx
@@ -7,7 +7,6 @@ import {useMutation, useQueryClient} from '@tanstack/react-query'
 import * as bcp47Match from 'bcp-47-match'
 
 import {wait} from '#/lib/async/wait'
-import {useCleanError} from '#/lib/hooks/useCleanError'
 import {popularInterests, useInterestsDisplayNames} from '#/lib/interests'
 import {isBlockedOrBlocking, isMuted} from '#/lib/moderation/blocked-and-muted'
 import {logger} from '#/logger'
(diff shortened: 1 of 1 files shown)


src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -75,6 +78,7 @@ export function StepSuggestedAccounts() {
     overrideInterests: state.interestsStepResults.selectedInterests,
   })
 
+  const {raw: rawError} = cleanError(error)
```
mozzius: I think(?) this isn't doing anything, it's just stringifying it.

I'm not a huge fan of the `useCleanError` hook in general, see here: https://linear.app/blueskyweb/issue/APP-1259/clean-up-cleanerror-handling-and-network-error-handling#comment-f376dbcd

src/screens/Onboarding/StepSuggestedAccounts/index.tsx
```diff
@@ -128,7 +132,11 @@ export function StepSuggestedAccounts() {
       toast.show(_(msg`Followed all accounts!`), {type: 'success'})
       setFollowedUsers(followed => [...followed, ...newlyFollowed])
     },
-    onError: () => {
+    onError: e => {
+      const {raw} = cleanError(e)
+      logger.error(raw || e, {
+        message: 'Failed to follow all suggested accounts during onboarding',
+      })
```
mozzius: In general I think we tend to do it the other way around:

```ts
logger.error('Failed to follow all suggested accounts during onboarding', {
  safeMessage: e
})
```

PR #9893, opened 2026-02-17
[APP-1859] pinned feed drag n drop

abenzer:
Updated prototype demonstrating drag-and-drop for reordering pinned feeds. Tested on all platforms. 

https://github.com/user-attachments/assets/eabec86f-69be-4388-a52d-58854eaea83f

### DIFF
--- a/src/alf/atoms.ts
+++ b/src/alf/atoms.ts
@@ -10,10 +10,6 @@ const EXP_CURVE = 'cubic-bezier(0.16, 1, 0.3, 1)'
 export const atoms = {
   ...baseAtoms,
 
-  rounded_sheet: {
-    borderRadius: 40,
-  },
-
   h_full_vh: web({
     height: '100vh',
   }),
--- a/src/components/Post/Embed/VideoEmbed/VideoEmbedInner/web-controls/VideoControls.tsx
+++ b/src/components/Post/Embed/VideoEmbed/VideoEmbedInner/web-controls/VideoControls.tsx
@@ -402,8 +402,8 @@ export function Controls({
           {hasSubtitleTrack && (
             <ControlButton
               active={subtitlesEnabled}
-              activeLabel={_(msg`Disable subtitles`)}
-              inactiveLabel={_(msg`Enable subtitles`)}
+              activeLabel={_(msg`Disable captions`)}
+              inactiveLabel={_(msg`Enable captions`)}
               activeIcon={CCActiveIcon}
               inactiveIcon={CCInactiveIcon}
               onPress={onPressSubtitles}
--- a/src/components/RichTextTag.tsx
+++ b/src/components/RichTextTag.tsx
@@ -148,7 +148,11 @@ export function RichTextTag({
         </Menu.Group>
         <Menu.Divider />
         <Menu.Item
-          label={isMuted ? _(msg`Unmute ${tag}`) : _(msg`Mute ${tag}`)}
+          label={
+            isMuted
+              ? _(msg`Unmute ${isCashtag ? tag : `#${tag}`}`)
+              : _(msg`Mute ${isCashtag ? tag : `#${tag}`}`)
+          }
           onPress={() => {
             if (isMuted) {
               resetUpsert()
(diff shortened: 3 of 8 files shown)


vineyardbovines:
dropping some thoughts here as well before i continue with this: the biggest thing for me is thinking about new arch migration down the line which has reanimated going from v3 -> v4 and introducing react-native-worklets, so this would become just-one-more-thing to make sure migrates cleanly when that happens.

that being said, an existing lib solution is pretty few and far between. i think the most up to date / widely used one is [react-native-draggable-flatlist](https://github.com/computerjazz/react-native-draggable-flatlist). there's also [react-native-reorderable-list](https://github.com/omahili/react-native-reorderable-list) which is a bit more actively maintained it seems. but this has always been one of those things that in the past i've ended up just building myself.

tl;dr i'd be for keeping it since its pretty narrow scope, so i'll pick it up from here and test it as much as possible on all the things you mentioned @estrattonbailey and then maybe even pull in one of those libs to see how it compares

vineyardbovines:
https://github.com/user-attachments/assets/5a2151b0-4334-4ee9-9a16-74c61c3f5829

auto scroll demo

vineyardbovines:
alright, ran through a bunch of tests with this:

| Manual test | Result |
| -- | -- |
| Swap adjacent feeds | ✅  |
| Drag to restore order | ✅  |
| Unpin/re-pin | ✅  |
| Multi-position drag | ✅  |
| Bottom-to-top drag | ✅  |
| Zoomed in drag | ✅  |
| Save change persistence | ✅  |

really not much to add here native-wise tbh! works as advertised. 
if anyone wants to give it a test, pls do 🙏

mozzius:
@github-actions ota

mozzius:
Feels great on device!

I think it could be like 25% more satisfying if it had a very very soft haptic as you move it over each row (would have to be iOS only)

mozzius:
@github-actions ota

mozzius:
@claude please review!

mozzius (changes requested):
The latest commit appears to introduce a crash when I start dragging :(

Also, note Claude's feedback about the unused code. `useAutoScroll` appears unused

Image

vineyardbovines:
damnit @claude did something get reverted that broke this?

vineyardbovines:
as an side, i need to check my claude code setup since i def encountered that crash while developing and then applied the same fix at one point 🤔 
either way @mozzius wanna give it another shot?

mozzius:
@github-actions ota

mozzius:
Hmm no that's not it, now I'm getting weird gaps and it's still not working

https://github.com/user-attachments/assets/e400f7f6-71bc-4c53-b5a9-33a7959c7952

vineyardbovines:
yeah seeing the same- gonna keep going on it

vineyardbovines:
alright, _now_ it should be all set to manually test. looking correct on my end

CleanShot 2026-02-20 at 17 42 34

src/components/DraggableList/index.web.tsx
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  const newPos = dragState ? getNewPosition(dragState) : -1
+
+  return (
+    <View
+      style={{height: data.length * itemHeight, position: 'relative'}}
+      // @ts-expect-error web-only pointer events
+      onPointerMove={handlePointerMove}
+      onPointerUp={handlePointerUp}>
```
mozzius: I think `onPointerCancel` is needed here too?

src/components/DraggableList/useAutoScroll.ts
```diff
@@ -0,0 +1,77 @@
...
+  const move = (clientY: number) => {
+    clientYRef.current = clientY
+  }
+
+  const stop = () => {
+    activeRef.current = false
+    cancelAnimationFrame(rafRef.current)
+    scrollableRef.current = null
+    onDeltaRef.current = null
+  }
```
mozzius: Feels a bit sketchy to me that this has to be manually cleaned up. Maybe clean up on unmount?

```ts
useEffect(() => {
  return () => {
    cancelAnimationFrame(rafRef.current)
  }
}, [])
```

vineyardbovines:
@mozzius yeah i ended up just removing it altogether, its what i had started with for autoscroll on hitting viewport edges but its handled now

PR #9895, opened 2026-02-17
ALF feed error screen

mozzius:
Get rid of custom error screen and replace it with the component

Screenshot 2026-02-17 at 08 40 21


src/screens/Profile/ProfileFeed/index.tsx
```diff
@@ -55,54 +48,30 @@ export function ProfileFeedScreen(props: Props) {
...
-                  <Trans>Go Back</Trans>
-                </Text>
-              </Button>
-            </View>
-          </View>
-        </Layout.Content>
+        <ErrorScreen
+          showHeader
+          title={_(msg`Could not load feed`)}
+          message={error.toString()}
```
ds-boyce: is this a good use case for `useCleanError`?

mozzius: Oh absolutely, good catch!

PR #9897, opened 2026-02-17
Add dynamic scroll indicator insets to profile tabs

mozzius:
Fixes https://bsky.app/profile/logic-denier.hellbhoy.net/post/3meyb4hu3jc25

## Summary
- Hoists `minimumHeaderHeight` from `PagerTabBar` to `PagerWithHeader` and exposes it as `collapsedHeaderHeight` (minimumHeaderHeight + tabBarHeight) through child params
- Adds animated iOS scroll indicator insets to `PostFeed` that track the collapsing header, gated behind an `adjustScrollIndicators` prop so it only activates on profile pages
- Applies the same pattern to Labels, Lists, Feeds, and Starter Packs profile tabs

### Before

See the big old gap when I scroll, and also that it goes behind the profile when I drag down

https://github.com/user-attachments/assets/587e5d0d-7042-4133-ba0f-8814897658a1

### After

https://github.com/user-attachments/assets/f6b4f240-11e3-468d-8662-5f1481862da6

## Test plan
- [ ] Open a profile on iOS, scroll down — scroll indicator should start below the full header and track down to the collapsed header height
- [ ] Verify the scroll indicator clamps at the collapsed header (minimal header + tab bar) and doesn't go above it
- [ ] Verify non-profile feeds (Following, Discover) are unaffected — static scroll indicator positioning
- [ ] Verify web is unaffected (collapsedHeaderHeight defaults to 0)
- [ ] Test on all profile tabs: Posts, Replies, Media, Videos, Likes, Feeds, Lists, Starter Packs, Labels

### DIFF
--- a/src/components/StarterPack/ProfileStarterPacks.tsx
+++ b/src/components/StarterPack/ProfileStarterPacks.tsx
@@ -201,40 +201,48 @@ export function ProfileStarterPacks({
     [isTabletOrDesktop, t.atoms.border_contrast_low],
   )
 
+  const list = (
+    <List
+      testID={testID ? `${testID}-flatlist` : undefined}
+      ref={scrollElRef}
+      data={items}
+      renderItem={renderItem}
+      keyExtractor={keyExtractor}
+      refreshing={isPTRing}
+      headerOffset={headerOffset}
+      progressViewOffset={ios(0)}
+      contentContainerStyle={{
+        minHeight: height + headerOffset,
+        paddingBottom: bottomBarOffset,
+      }}
+      removeClippedSubviews={true}
+      desktopFixedHeight
+      onEndReached={onEndReached}
+      onRefresh={onRefresh}
+      ListEmptyComponent={
+        data ? (isMe ? EmptyComponent : undefined) : FeedLoadingPlaceholder
+      }
+      ListFooterComponent={
+        !!data && items?.length !== 0 && isMe ? CreateAnother : undefined
+      }
+      animatedProps={IS_IOS ? animatedProps : undefined}
+    />
+  )
+
   return (
     <View testID={testID} style={style}>
-      <ScrollProvider
-        onScroll={onScrollWorklet}
-        onBeginDrag={onBeginDragFromContext}
-        onEndDrag={onEndDragFromContext}
-        onMomentumBegin={onMomentumEndFromContext}
-        onMomentumEnd={onMomentumEndFromContext}>
-        <List
-          testID={testID ? `${testID}-flatlist` : undefined}
-          ref={scrollElRef}
-          data={items}
-          renderItem={renderItem}
-          keyExtractor={keyExtractor}
-          refreshing={isPTRing}
-          headerOffset={headerOffset}
-          progressViewOffset={ios(0)}
-          contentContainerStyle={{
-            minHeight: height + headerOffset,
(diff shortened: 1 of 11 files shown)


helle253:
I'm the one that filed the bug report - huge h/t to @mozzius for being on the ball, i wasn't even expecting it to get fixed tbh

src/view/com/posts/PostFeed.tsx
```diff
@@ -985,37 +996,88 @@ let PostFeed = ({
...
+      'worklet'
+      onScrollFromContext?.(e, ctx)
+      scrollY.set(e.contentOffset.y)
+    },
+    [onScrollFromContext, scrollY],
+  )
+
+  const {footerHeight} = useShellLayout()
+
+  const animatedProps = useAnimatedProps(() => {
```
vineyardbovines: seems like it might make sense to extract this out and then reuse it since we have it in 5 places?

src/components/StarterPack/ProfileStarterPacks.tsx
```diff
@@ -158,32 +200,48 @@ export function ProfileStarterPacks({
...
-        ListFooterComponent={
-          !!data && items?.length !== 0 && isMe ? CreateAnother : undefined
-        }
-      />
+      {IS_IOS ? (
+        <ScrollProvider
+          onScroll={onScrollWorklet}
+          onBeginDrag={onBeginDragFromContext}
+          onEndDrag={onEndDragFromContext}
+          onMomentumBegin={onMomentumEndFromContext}
```
vineyardbovines: do we want `onMomentumEndFromContext` to fire on both begin and end? is that just a naming thing or would there actually be collision here

src/view/com/posts/PostFeed.tsx
```diff
@@ -985,37 +996,88 @@ let PostFeed = ({
...
-        desktopFixedHeight={
-          desktopFixedHeightOffset ? desktopFixedHeightOffset : true
-        }
-        initialNumToRender={initialNumToRenderOverride ?? initialNumToRender}
-        windowSize={9}
-        maxToRenderPerBatch={IS_IOS ? 5 : 1}
-        updateCellsBatchingPeriod={40}
-        onItemSeen={onItemSeen}
-      />
+      {IS_IOS ? (
```
vineyardbovines: wouldn't this run on every postfeed? i.e. discover, etc. and not just profile? probably wouldn't cause jank since it's running in worklets anyway but it would be tracking `scrollY` on every tab

vineyardbovines (commented):
coupla questions

src/view/com/posts/PostFeed.tsx
mozzius: Good call!

src/components/StarterPack/ProfileStarterPacks.tsx
mozzius: Good catch 👀

src/view/com/posts/PostFeed.tsx
mozzius: Yeah I think this is fixable

mozzius:
@vineyardbovines thanks for prompting me to abstract into a hook, it's now a lot more cleaner and avoids work on other platforms by platform-splitting the hook. The diff with whitespace hidden is a lot cleaner now too

PR #9898, opened 2026-02-17
Clear link and gif caches to avoid pointing at missing files

vineyardbovines:
Clear resolve-link and resolve-gif query caches when the composer closes, since purgeTemporaryImageFiles() deletes the cached thumbnails those queries reference. Without this, restoring a draft serves stale ResolvedLink data pointing at missing files.

### DIFF
--- a/src/state/shell/composer/index.tsx
+++ b/src/state/shell/composer/index.tsx
@@ -12,7 +12,11 @@ import {useQueryClient} from '@tanstack/react-query'
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {postUriToRelativePath, toBskyAppUrl} from '#/lib/strings/url-helpers'
 import {purgeTemporaryImageFiles} from '#/state/gallery'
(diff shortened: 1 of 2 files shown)


src/state/shell/composer/index.tsx
```diff
@@ -120,6 +124,12 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
     if (wasOpen) {
       setState(undefined)
       purgeTemporaryImageFiles()
+      // Purging deletes cached thumbnails on disk, so remove the query
+      // caches that may hold references to those now-deleted file paths.
+      // Without this, restoring a draft would serve stale ResolvedLink
+      // data pointing at missing files, causing "Failed to load blob".
+      queryClient.removeQueries({queryKey: [RQKEY_LINK_ROOT]})
+      queryClient.removeQueries({queryKey: [RQKEY_GIF_ROOT]})
```
mozzius, line 132: hmm I think it would make more sense to put this inside of `purgeTemporaryImageFiles()`

mozzius: actually nah it's fine, changed my mind

PR #9900, opened 2026-02-18
Add chat data export to Export My Data dialog

mozzius:
Not the prettiest but it gets the job done

Screenshot 2026-02-18 at 10 37 12

## Summary
- Adds a secondary "Download chat data" button to the Export My Data dialog, using the `chat.bsky.actor.exportAccountData` endpoint
- Uses raw fetch via `agent.sessionManager.fetchHandler` instead of the typed XRPC client because the client's `httpResponseBodyParse` incorrectly matches `application/jsonl` as `application/json` (substring match), causing a JSON parse error
- Tracks loading state per-button (`'repo' | 'chat' | false`) so each button shows its own spinner while both are disabled during a download

## Test plan
- [ ] Open Settings > Account > Export My Data
- [ ] Tap "Download CAR file" — verify it still works as before
- [ ] Tap "Download chat data" — verify it downloads a `.jsonl` file
- [ ] Verify both buttons are disabled while either download is in progress
- [ ] Test on web + native

### DIFF
--- a/src/screens/Settings/components/ExportCarDialog.tsx
+++ b/src/screens/Settings/components/ExportCarDialog.tsx
@@ -121,6 +121,14 @@ export function ExportCarDialog({
             {loading === 'repo' && <ButtonIcon icon={Loader} />}
           </Button>
 
+          <Text style={[a.text_sm, a.leading_snug, t.atoms.text_contrast_high]}>
+            <Trans>
+              You can also download your chat data as a "JSONL" file. This file
+              only includes chat messages that you have sent and does not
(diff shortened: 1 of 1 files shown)


surfdude29:
I just tested this on the preview branch and it works well, nicely done 👌

For the copy shown in the dialog, adding something like this might be helpful:

> You can also download your chat data as a "JSONL" file. This file only includes chat messages that you have sent and does not include chat messages that you have received.

mozzius:
Good idea @surfdude29

PR #9901, opened 2026-02-18
[APP-1809] add cleanup button for unavailable moderation services

vineyardbovines:
Adds a tip if there are any unavailable mod services and allows for removal. 

Someone may have to check me here, but I don't think unavailable services actually get returned/listed (other than the DID) so there's no other visual indicators here.

image


src/screens/Moderation/index.tsx
```diff
@@ -42,6 +43,7 @@ import {InlineLinkText, Link} from '#/components/Link'
 import {ListMaybePlaceholder} from '#/components/Lists'
 import {Loader} from '#/components/Loader'
 import {GlobalLabelPreference} from '#/components/moderation/LabelPreference'
+import * as toast from '#/components/Toast'
```
mozzius: nit: we're a bit inconsistent with this, but I think usually we go with `Toast` (uppercase)

src/screens/Moderation/index.tsx
```diff
@@ -439,6 +465,34 @@ export function ModerationScreenInner({
...
+              label={_(msg`Remove unavailable moderation services`)}
+              onPress={handleCleanup}
+              disabled={isRemovingLabelers}>
+              {isRemovingLabelers ? (
+                <Loader size="xs" />
+              ) : (
+                <ButtonText>
+                  <Trans>Remove</Trans>
+                </ButtonText>
+              )}
```
mozzius: nit: this is a bit cleaner imo - also make sure to use `ButtonIcon`!
```suggestion
              <ButtonText>
                <Trans>Remove</Trans>
              </ButtonText>
              {isRemovingLabelers && <ButtonIcon icon={Loader} />}
```

src/screens/Moderation/index.tsx
vineyardbovines: 👌

mozzius:
@vineyardbovines CI is failing :(

vineyardbovines:
@mozzius looks like its all set now? just needed to re-run the linter

PR #9925, opened 2026-02-21
Fix reply failures caused by fetching posts from PDS

mozzius:
Fixes https://bsky.app/profile/numb.comfortab.ly/post/3mfdbrispuk2z

What I think is happening:

- the client calls `agent.getPost()` to get the replyRef, which under the hood is `com.atproto.repo.getRecord`
- the PDS checks if the DID is local to that PDS
  - if yes: fetch the record from it's own DB
  - if no: forward request to appview

what's happened is `numb.comfortab.ly` has migrated from Amanita to Northsky. when I try and reply to their post, Amanita looks up their DID, thinks it's one of theirs, and fails to find the post so errors

## The fix

use `getPosts` instead of `getPost`, since that always goes straight through to the AppView

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -5,8 +5,7 @@ import {
   type AppBskyEmbedRecord,
   type AppBskyEmbedRecordWithMedia,
   type AppBskyEmbedVideo,
-  type AppBskyFeedPost,
-  AtUri,
+  AppBskyFeedPost,
   BlobRef,
   type BskyAgent,
   type ComAtprotoLabelDefs,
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):

heikadog:
Seeing this. Can't reply to Rudy after he switched PDSes at one point to blackstar.quest and incorporated it into Blacksky.
Screenshot_20260404-151753

The reply window says the post was deleted, but I can still interact in all the other ways (quoting, reposting, liking).

Seeing this cross-platform.

spookalucca:
users are still waiting on this!

mozzius:
@claude double check the logic pls

spookalucca:
I'm glad this has been merged, has it been deployed yet? We're still seeing some reports from our users about this.

PR #9930, opened 2026-02-23
Use on-device translation on mobile when available

ds-boyce:
On Android and iOS, if on-device translation fails, it silently falls back to Google Translate.

If language detection is inaccurate, a language override can be selected. Note that available native options differ by platform; other languages fall back to Google Translate.

Web always uses Google Translate.

https://github.com/user-attachments/assets/9a565b6c-f89e-44d8-a4c6-a7631083d87c

https://github.com/user-attachments/assets/d71bda0d-1934-4e81-bcb4-e3c5e4cab8c5

**Test plan**

* **Web** – no change, tested and verified in a desktop browser
* **Android** – tested and verified on a device, fallback to Google Translate tested and verified in an emulator and with unsupported languages
* **iOS** – tested and verified on a device, fallback to google translate tested and verified in a simulator and with unsupported languages

### DIFF
--- a/src/components/ContextMenu/context.tsx
+++ b/src/components/ContextMenu/context.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import {createContext, useContext} from 'react'
 
 import {
   type ContextType,
@@ -6,17 +6,17 @@ import {
   type MenuContextType,
 } from '#/components/ContextMenu/types'
 
-export const Context = React.createContext<ContextType | null>(null)
+export const Context = createContext<ContextType | null>(null)
 Context.displayName = 'ContextMenuContext'
 
-export const MenuContext = React.createContext<MenuContextType | null>(null)
+export const MenuContext = createContext<MenuContextType | null>(null)
 MenuContext.displayName = 'ContextMenuMenuContext'
 
-export const ItemContext = React.createContext<ItemContextType | null>(null)
+export const ItemContext = createContext<ItemContextType | null>(null)
 ItemContext.displayName = 'ContextMenuItemContext'
 
 export function useContextMenuContext() {
-  const context = React.useContext(Context)
+  const context = useContext(Context)
 
   if (!context) {
     throw new Error(
@@ -28,7 +28,7 @@ export function useContextMenuContext() {
 }
 
 export function useContextMenuMenuContext() {
-  const context = React.useContext(MenuContext)
+  const context = useContext(MenuContext)
 
   if (!context) {
     throw new Error(
@@ -40,7 +40,7 @@ export function useContextMenuMenuContext() {
 }
 
 export function useContextMenuItemContext() {
-  const context = React.useContext(ItemContext)
+  const context = useContext(ItemContext)
 
   if (!context) {
     throw new Error(
--- a/src/components/ContextMenu/index.tsx
+++ b/src/components/ContextMenu/index.tsx
@@ -23,6 +23,7 @@ import {
   type GestureUpdateEvent,
   type PanGestureHandlerEventPayload,
 } from 'react-native-gesture-handler'
+import {KeyboardEvents} from 'react-native-keyboard-controller'
 import Animated, {
   clamp,
   interpolate,
@@ -35,6 +36,7 @@ import Animated, {
   type WithSpringConfig,
 } from 'react-native-reanimated'
 import {
+  type EdgeInsets,
   useSafeAreaFrame,
   useSafeAreaInsets,
 } from 'react-native-safe-area-context'
@@ -81,9 +83,9 @@ export {
 const {Provider: PortalProvider, Outlet, Portal} = createPortalGroup()
 
 const SPRING_IN: WithSpringConfig = {
-  mass: IS_IOS ? 1.25 : 0.75,
-  damping: 50,
-  stiffness: 1100,
+  mass: 0.75,
+  damping: 300,
+  stiffness: 1200,
   restDisplacementThreshold: 0.01,
 }
 
@@ -110,6 +112,7 @@ export function Root({children}: {children: React.ReactNode}) {
   const playHaptic = useHaptics()
   const [mode, setMode] = useState<'full' | 'auxiliary-only'>('full')
   const [measurement, setMeasurement] = useState<Measurement | null>(null)
+  const returnLocationSV = useSharedValue<{x: number; y: number} | null>(null)
   const animationSV = useSharedValue(0)
   const translationSV = useSharedValue(0)
   const isFocused = useIsFocused()
@@ -142,6 +145,7 @@ export function Root({children}: {children: React.ReactNode}) {
       ({
         isOpen: !!measurement && isFocused,
         measurement,
+        returnLocationSV,
         animationSV,
         translationSV,
         mode,
@@ -149,6 +153,8 @@ export function Root({children}: {children: React.ReactNode}) {
           setMeasurement(evt)
           setMode(mode)
           animationSV.set(withSpring(1, SPRING_IN))
+          // reset return location
+          returnLocationSV.set(null)
         },
         close: () => {
           animationSV.set(
@@ -156,6 +162,9 @@ export function Root({children}: {children: React.ReactNode}) {
               if (finished) {
                 hoverablesSV.set({})
                 translationSV.set(0)
+                // note: return location has to be reset on open,
+                // rather than on close, otherwise there's a flicker
+                // where the reanimated update is faster than the react render
                 runOnJS(onCompletedClose)()
               }
             }),
@@ -194,6 +203,7 @@ export function Root({children}: {children: React.ReactNode}) {
       }) satisfies ContextType,
     [
       measurement,
+      returnLocationSV,
       setMeasurement,
       onCompletedClose,
       isFocused,
@@ -225,7 +235,7 @@ export function Root({children}: {children: React.ReactNode}) {
 export function Trigger({children, label, contentLabel, style}: TriggerProps) {
   const context = useContextMenuContext()
   const playHaptic = useHaptics()
-  const {top: topInset} = useSafeAreaInsets()
+  const insets = useSafeAreaInsets()
   const ref = useRef<View>(null)
   const isFocused = useIsFocused()
   const [image, setImage] = useState<string | null>(null)
@@ -237,23 +247,8 @@ export function Trigger({children, label, contentLabel, style}: TriggerProps) {
   const open = useNonReactiveCallback(
     async (mode: 'full' | 'auxiliary-only') => {
       playHaptic()
-      Keyboard.dismiss()
       const [measurement, capture] = await Promise.all([
-        new Promise<Measurement>(resolve => {
-          ref.current?.measureInWindow((x, y, width, height) =>
-            resolve({
-              x,
-              y:
-                y +
-                platform({
-                  default: 0,
-                  android: topInset, // not included in measurement
-                }),
-              width,
-              height,
-            }),
-          )
-        }),
+        measureView(ref.current, insets),
         captureRef(ref, {result: 'data-uri'}).catch(err => {
           logger.error(err instanceof Error ? err : String(err), {
             message: 'Failed to capture image of context menu trigger',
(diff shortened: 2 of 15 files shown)


mozzius:
@ds-boyce this is looking really good! excited for it. initial thoughts before I look deeply at the code:

1. Really minor, but I have my app language set to British English, so it was surprising to me on iOS that it wanted to download the English (US) language pack instead of English (UK). maybe we could special-case it? (100% guarantee people will complain lol)
2. The "Edit" button... confusing what it does. I take it it lets you override the *source* language (not the destination language). I think it would help to set a more descriptive label in the Select menu, currently it just says "Select language"
3. For both this edit select menu, and the content languages menu, maybe an indication that machine translation is available on this device? I'm thinking something like:

English — _inline translation available_

4. Not sure how to reproduce, but at one point I got it into a state where it would flicker idle->loading->idle on press without opening google translate. I think maybe we need to `key` the whole component based on content language?
5. Probably a pre-existing problem, but can we sort the content languages menu A-Z? It's currently scrambled for me on iOS
6. Need a bit of `hitSlop` on those buttons! very hard to press :)

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -608,18 +643,32 @@ function ExpandedPostDetails({
...
-              )}
-              label={_(msg`Translate`)}
-              style={[a.text_sm]}
-              onPress={onTranslatePress}>
-              <Trans>Translate</Trans>
-            </InlineLinkText>
+            {translationState.status === 'loading' ? (
+              <View style={[a.flex_row, a.align_center, a.gap_xs]}>
+                <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
+                  <Loader size="xs" /> <Trans>Translating…</Trans>
```
mozzius: nit: nesting Views inside of Text is generally a bad idea, from the looks of it you were intending to put it outside anyway

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -589,9 +608,25 @@ function ExpandedPostDetails({
...
+    },
+    [clearTranslation],
+  )
+
+  useEffect(() => {
+    return () => {
+      // Clear the translation when leaving the post
+      clearTranslation()
+    }
+  }, [clearTranslation])
```
mozzius: screens generally don't get unmounted when you leave them - I think `useFocusEffect` from React Navigation would better here

src/screens/Settings/LanguageSettings.tsx
```diff
@@ -144,7 +144,10 @@ export function LanguageSettingsScreen({}: Props) {
                       <Select.ItemText>{label}</Select.ItemText>
                     </Select.Item>
                   )}
-                  items={DEDUPED_LANGUAGES.map(l => ({
+                  items={DEDUPED_LANGUAGES.sort(
+                    (a, b) =>
+                      a.name.localeCompare(b.name, langPrefs.primaryLanguage), // Localized sort
+                  ).map(l => ({
```
mozzius, line 150: oh huh you did try and sort it. weirdly, it made the sorting all funky on iOS - this is the end of the list:

Image

patches/expo-translate-text+0.1.0.patch
mozzius, line 1: this is a substantial patch - do you think it's maybe worth maintaining our own fork of this library?

If so-
- fork into `bluesky-social` org
- publish on npm as `@bsky.app/expo-translate-text` (I can invite you to the npm org)

src/screens/Settings/LanguageSettings.tsx
ds-boyce: In my opinion, these should be the localized names regardless of app language, e.g., “Hindi” should always be “हिन्दी”, since that’s what a native speaker would be looking for. The localized sort would make more sense then.

I had started down that path, but overlooked this line when backing it out. Should I revisit it?

ds-boyce: This is what the end of the list looks like with the above proposal (everything above is A–Z):
languages

ds-boyce: Actually, for this specific case it makes more sense to use the current app language to localize this list, since you’re selecting a language you may not be familiar with. I’ll take another look at the sort here.

ds-boyce:
Thanks for taking a look! I really appreciate the feedback.

1. I added some special cases for `"en"`, `"es"`, `"pt"`, and `"zh"`: It now uses the iOS language preference to further refine the language selection, e.g., from `"en"` to `"en-GB"`.
2. I updated the label to read “Select the source language”.
3. Two concerns:
   * There may not be enough room for the suggested text, especially when localized, and
   * the supported languages may change after an OS update.
4. I haven’t been able to reproduce this but I’ll keep trying.
5. I fixed sorting. 🙂 
6. Added a `hitSlop` to the “Edit” button. No change to the existing “Translate” link since it’s a `Text` component.

Please take another look and let me know if I missed anything.

patches/expo-translate-text+0.1.0.patch
ds-boyce: Switched to [@bsky.app/expo-translate-text](https://www.npmjs.com/package/@bsky.app/expo-translate-text).

mozzius:
fyi @ds-boyce I just rebased on main

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -411,6 +418,11 @@ const ThreadItemAnchorInner = memo(function ThreadItemAnchorInner({
                 shouldProxyLinks={true}
               />
             ) : undefined}
+            <TranslatedPost
+              postText={record.text}
+              postUri={post.uri}
+              hideLoading
+            />
             {post.embed && (
```
mozzius: Imo the Translate button should go here, because at the moment it appears on the other side of any embeds which is a bit confusing. It would make sense for the result of the button to appear near to the button

Image

src/components/Post/Translated/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+              </Text>{' '}
+              <TranslationLanguageSelect
+                postUri={postUri}
+                sourceLanguage={sourceLanguage}
+                postText={postText}
+              />
+            </>
+          )}
+        </Text>
+        <Text emoji selectable style={[a.text_md]}>
```
mozzius: nit: `a.leading_snug` tends to look a little btter

```suggestion
        <Text emoji selectable style={[a.text_md, a.leading_snug]}>
```

src/translation/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+const IDLE: TranslationState = {status: 'idle'}
+
+const emitter = new Emitter()
+
+// Note: Since we’re storing this in memory without clearing it, e.g., via LRU,
+// this could get large over time if we don't clear the translation when the
+// component is unmounted. Something to keep an eye out for.
+const translations = new Map<string, TranslationState>()
```
mozzius: Okay this is my big bit of feedback: I don't think the EventEmitter+useSyncExternalStore is the right approach here. The translation state only really has to live in one place (the thread anchor), so I'd just hoist it up to there, hold the logic in a hook, and then just pass data/callbacks down to the relevant display components. This means we don't have to worry about the cache growing, or keeping track of events, because we can just keep it all inside React.

Generally, we tend to lean towards React Context over event emitters (we do use both though) - and in this case, since it's quite local, I think you can just do it by passing props directly. I don't think there'll be a situation where this translation state is consumed outside of this thread anchor?

(if there's a specific reason for this architecture I'm not seeing please lemme know!)

src/screens/PostThread/components/ThreadItemAnchor.tsx
ds-boyce: Relocated the “Translate” link to above the embedded content.

translate

src/translation/index.ts
ds-boyce: The `translationState` is consumed in three places:

* `PostMenuItems` which displays a “Translate” / “Translating…” / “Hide Translation” link,
* `ThreadItemAnchor` which also displays a “Translate” link, and
* `TranslatedPost` which displays the actual translation result.

These three things need to be kept in sync; simply putting `useState` in the `useTranslateOnDevice` hook doesn’t trigger a re-render of all three components, only the `ThreadItemAnchor` or `PostMenuItems`; whichever makes the call.

An event-driven approach seemed cleaner to me than adding to context (I was trying to be mindful of cascading re-renders, which `useSyncExternalStore` avoids), but I could investigate the latter if it’s preferred.

mozzius: I would lean towards Context/prop drilling - I think it's cleanest imo. Cascading renders is a reasonable concern, but the state isn't going to be changing very much, and with React Compiler  it's not such a big deal.

I would've though you can have a single useTranslateOnDevice() in the anchor component, and pass the callback+pending state from that down to the two buttons, and pass the output to TranslatedPost

ds-boyce: I could indeed! I was also trying to avoid prop drilling. 😅 

I’ll take another pass at this.

ds-boyce: FWIW My other concern with calling the hook in the anchor component is that changes to translation state will cause the entire component to re-render, vs the pieces that care about that state. But I don’t want to belabor this (too much), so take this as voicing my two cents. 🙂

ds-boyce: Ok switched to the Context API. Please take another tour around and let me know what you think.

src/App.native.tsx
```diff
@@ -179,9 +180,11 @@ function InnerApp() {
...
                                                           <GlobalGestureEventsProvider>
                                                             <IntentDialogProvider>
-                                                              <TestCtrls />
-                                                              <Shell />
-                                                              <ToastOutlet />
+                                                              <TranslationProvider>
+                                                                <TestCtrls />
+                                                                <Shell />
+                                                                <ToastOutlet />
+                                                              </TranslationProvider>
```
mozzius: What's the reasoning behind wrapping the whole app in a `TranslationProvider`? Can't it be *just* the thread anchor? This leads to some really weird behaviour - PostMenuItems doesn't just appear in the anchor, it's also in other places such as the feed. Using one single global state for this connects all the menus together, meaning that
a) the button activates something on a different screen
b) the state is shared between the different posts - see this video, where I translate one post, then go the next post and the button says "Hide Translation"
  
https://github.com/user-attachments/assets/5df9a7a5-1395-4b65-bf38-ce0224573d01

The API I'd like to see would be more like: 

```tsx
function ThreadItemAnchor() {
  
  return (
    <TranslateOnDeviceProvider postText={post.record.text}>
      <BlahBlahBlah />   
    
      <TranslateDisplay />
      <TranslateButton />
    
      <BlahBlahBlah /> 
      <PostControls />
      <BlahBlahBlah />   
    </TranslateOnDeviceProvider>
  )
}

// then in PostControls
const translate = useTranslateOnDevice()
if (translate) {
  // we're within the context!
  translate.show()
} else {
  // we're not in the context, so just open in browser
  // since you wouldn't be able to see the result
  fallbackTranslation()
}
```

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -589,9 +618,68 @@ function ExpandedPostDetails({
...
+    },
+    [clearTranslation],
+  )
+
+  useFocusEffect(() => {
+    return () => {
+      // Clear the translation when leaving the post
+      clearTranslation()
+    }
+  })
```
mozzius: I thin you can delete this now that it's scoped per-post 👍

src/App.native.tsx
ds-boyce: Ok, I scoped the `TranslateOnDeviceProvider` to `ThreadItemAnchor`. Additionally, `PostMenuItems` always links out to Google Translate.

src/translation/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+      case 'zh': // zh-Hans-CN, zh-Hant-HK, zh-Hant-TW
+        targetLangCode = primaryLanguageTag ?? targetLangCodeOriginal
+        break
+    }
+  }
+
+  const {onTranslateTask} =
+    // Needed in order to type check the dynamically imported module.
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    (await require('@bsky.app/expo-translate-text')) as typeof import('@bsky.app/expo-translate-text')
```
mozzius: pretty sure `require` is synchronous, so you don't need the await

src/translation/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+          os: Platform.OS,
+          sourceLanguage: result.sourceLanguage,
+          targetLanguage: result.targetLanguage,
+        })
+        setTranslationState({
+          status: 'success',
+          translatedText: result.translatedText,
+          sourceLanguage: result.sourceLanguage,
+          targetLanguage: result.targetLanguage,
+        })
```
mozzius, line 159: You could add `LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut)` next to this setState to make this animate nicely - same for `clearTranslation`

src/components/Post/Translated/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+}) {
+  const t = useTheme()
+  const {i18n} = useLingui()
+
+  const langName = sourceLanguage
+    ? codeToLanguageName(sourceLanguage, i18n.locale)
+    : undefined
+
+  return (
+    <>
```
mozzius: nit: unnecessary fragment

mozzius (approved):
This is so great, nice work! Sorry for the back-and-forth but I think it's much better for it.

Let's get it on the TestFlight so we can stress-test it!

(stamped so merge at your pleasure, but left some little nits)

src/translation/index.tsx
ds-boyce: Ah, leftover from when i attempted `await import()`.

PR #9938, opened 2026-02-24
iOS 26 tweaks #2

mozzius:
fix some missing padding


Ichicoro:
Is there any way to test this build via testflight?

mozzius:
we do not have a public testflight at this time, no. you'll just have to wait ;)

PR #9945, opened 2026-02-25
[Chat] Blurred chat header

mozzius:
## Summary

Adds a Liquid Glass blur effect to the chat conversation header on iOS 26, using a new `expo-scroll-edge-interaction` native module that wraps Apple's `UIScrollEdgeElementContainerInteraction` API.

### What changed

- **New native module** `expo-scroll-edge-interaction` (iOS-only) — wraps `UIScrollEdgeElementContainerInteraction`. Receives the header view tag and scroll view tag from JS, sets up the interaction on the native side. Supports configurable edge direction.

- **New React context** `ScrollEdgeInteractionProvider` + hooks:
  - `useTransparentHeaderProps()` — returns ref/onLayout for the header view, or `undefined` when not in a provider
  - `useScrollEdgeScrollView()` — returns a ref callback for the scroll view, used automatically by `List`
  - `useTransparentHeaderHeight()` — returns the measured header height (`null` outside provider)

- **`Layout.Screen`** gains a `transparentHeader` prop that wraps children in the provider and skips the top safe area inset. No-op on non-Liquid-Glass devices.

- **`Layout.Header.Outer`** (iOS) auto-detects the provider and renders as a transparent absolutely-positioned overlay when present.

- **`List`** automatically registers its scroll view with the provider and applies `contentInset` for the header height.

### Limitations

- The blur effect currently only works with `List` (native FlatList), not `Layout.Content` (Animated.ScrollView). We haven't figured out how to get iOS 26's scroll edge effects working on `Animated.ScrollView` yet.

## Test plan

- [ ] iOS 26 simulator: open a chat conversation — header should have progressive Liquid Glass blur as messages scroll underneath
- [ ] iOS < 26: should fall back to normal opaque header
- [ ] Web: no change, passthrough
- [ ] Content inset: messages should start below the header, scroll smoothly behind it
- [ ] `yarn typecheck` and `yarn lint` pass


mozzius:
I want to make a nice generic API for this before we merge

PR #9949, opened 2026-02-25
Add search event analytics

ds-boyce:
Adds new `search:*` analytics events to provide insights into search behavior. Notably, queries themselves are not logged.

Look for `ax.metric` calls; the rest is just cleaning up lint warnings.

### DIFF
--- a/src/screens/Search/SearchResults.tsx
+++ b/src/screens/Search/SearchResults.tsx
@@ -523,7 +523,7 @@ function SearchScreenProfileButton({
       tab: 'people',
       resultType: 'profile',
       position,
-      uri: `at://${profile.did}`,
+      uri: profile.did,
     })
   }
   return <ProfileCardWithFollowBtn profile={profile} onPress={handlePress} />


mozzius:
@ds-boyce this has picked up some merge conflicts - mind fixing them? they look pretty trivial luckily

metrics changes look good though I think

estrattonbailey (approved):

src/screens/Search/SearchResults.tsx
```diff
@@ -185,7 +192,7 @@ function NoResultsText({query}: {query: string}) {
             })}
             to={urls.website.blog.searchTipsAndTricks}
             style={[a.text_md, a.leading_snug]}>
-            read about how to use search filters
+            read about how to use search filters.
```
mlocati, line 195: @ds-boyce why did you add that period? We already have it after the `</InlineLinkText>` closing tag.

In crowdin we see this translatable string (please remark the two periods, one before `</0>` and one after it):

```
Try a different search term, or <0>read about how to use search filters.</0>.
```

estrattonbailey: Ope sorry, we just missed that. We'll get an OTA out with translations and some tweaks in the next day or so, can include this.

ds-boyce: Ah, I overlooked that. Thanks for flagging.

ds-boyce: Fixed in #0004. Sorry about that!

PR #9950, opened 2026-02-25
Fix infinite loading spinner when changing search terms

ds-boyce:
**Test plan**
1. Open the Bluesky app
2. Search for a term, e.g., “test”
3. Switch to “Latest” tab
4. Search for a new term, e.g., “test2”

**Previously**
An infinite loading spinner until switching tabs.

**Now**
New results load as expected.

https://github.com/user-attachments/assets/0646df5b-1a55-40a4-b718-f744b14430aa

**Test plan**
1. Open the Bluesky app
2. Search for a term, e.g., “test”
3. Change the search results language

**Previously**
An infinite loading spinner until switching tabs.

**Now**
New results load as expected.

https://github.com/user-attachments/assets/67133431-8e31-433e-82ef-eb30e2487b28

### DIFF
--- a/src/screens/Search/Shell.tsx
+++ b/src/screens/Search/Shell.tsx
@@ -467,7 +467,7 @@ let SearchScreenInner = ({
   query: string
   queryWithParams: string
   headerHeight: number
-  focusSearchInput: (tab?: 'user' | 'profile' | 'feed') => void
+  focusSearchInput: (tab?: TabParam) => void
 }): React.ReactNode => {
   const t = useTheme()
   const setMinimalShellMode = useSetMinimalShellMode()


ds-boyce:
Ooh, good catch, will fix.

src/screens/Search/SearchResults.tsx
```diff
@@ -1,9 +1,8 @@
 import {memo, useCallback, useMemo, useState} from 'react'
 import {ActivityIndicator, View} from 'react-native'
 import {type AppBskyFeedDefs} from '@atproto/api'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
-import {Trans} from '@lingui/react/macro'
+import {type ProfileView} from '@atproto/api/dist/client/types/app/bsky/actor/defs'
```
mozzius: nit: IDE auto-import always does this annoying deep import, but instead you want to do 

```ts
import {AppBskyActorDefs} from '@atproto/api'

let user: AppBskyActorDefs.ProfileView
```

double nit: we actually have a type helper here, as there's a few different kinds of profile view, so it's usually best to just use the union version:

```ts
import type * as bsky from '#/types/bsky'

let user: bsky.profile.AnyProfileView
```

ds-boyce: The former seems to work, but the latter doesn’t match the type in `ProfileCard.tsx`, so I updated the type on the `profile` prop there as well.

src/screens/Search/SearchResults.tsx
```diff
@@ -423,10 +417,14 @@ let SearchScreenUserResults = ({
...
-          keyExtractor={item => item.did}
+          renderItem={({item}: {item: bsky.profile.AnyProfileView}) => (
+            <ProfileCardWithFollowBtn profile={item} />
+          )}
+          keyExtractor={(item: bsky.profile.AnyProfileView) => item.did}
           refreshing={isPTR}
-          onRefresh={onPullToRefresh}
+          onRefresh={() => {
+            void onPullToRefresh()
+          }}
```
mozzius:
```suggestion
          onRefresh={() => void onPullToRefresh()}
```

mozzius (approved):
Nice work! No more crashes 💆

PR #9951, opened 2026-02-25
[APP-1876] Fix pull-to-refresh getting stuck on tab switch

vineyardbovines:
`isPTRing` controls the `refreshing` prop, if you pull to refresh and the component becomes inactive, `isPTRing` remains true until the async refresh completes.


mozzius:
@github-actions ota

PR #9952, opened 2026-02-25
Give prompts a little more breathing room on web

ds-boyce:
| Before | After |
| - | - |
| before | after |

### DIFF
--- a/src/components/Prompt.tsx
+++ b/src/components/Prompt.tsx
@@ -61,7 +61,7 @@ export function Outer({
         <Dialog.ScrollableInner
           accessibilityLabelledBy={titleId}
           accessibilityDescribedBy={descriptionId}
-          style={web([{maxWidth: 320, borderRadius: 36}])}>
+          style={web([{maxWidth: 360, borderRadius: 36}])}>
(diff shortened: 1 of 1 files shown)


mozzius:
Hmm, in this case, would it make more sense to use a plain `Dialog` instead of a `Prompt`? There's a lot more going on here than we would normally stuff into a `Prompt`.

With `Dialog`s, we generally just set whatever max-width feels right for that specific case

ds-boyce:
> There's a lot more going on here than we would normally stuff into a Prompt.

Ah, ok. The presentation of `Prompt` on web seems to be very close to a `Dialog`. Is the difference just a matter of content?

mozzius:
I don't know what the original idea behind it was, but it basically is just for simple cases - we have a `Prompt.Basic` wrapper too. I think if you're mixing in non-Prompt components, at that point you probably want a Dialog

ds-boyce:
Gotcha. The `Admonition` is the only non-`Prompt` piece, but I see your point. I’ll close this out in favor of switching to a Dialog.

Thanks!

PR #9953, opened 2026-02-25
[APP-1882] fix email not updating in session state after email change

vineyardbovines:
`agent.resumeSession()` silently discards the `getSession` response when the refresh token matches the current session, so email/email confirmed changes never propagate to local state


mozzius:
@claude please review

PR #9955, opened 2026-02-26
Remove JPG hardcoding from image processing systems

mozzius:
Removes assumptions about the file type that the CDN returns.

- bskyogcard now sniffs mimetype from the magic bytes
- downloading images is now agnostic of mime type
- saving images re-encodes as JPG (because I know people will complain lol)

and slightly unrelated but:

- in onboarding, rather than rejecting non-jpg/png images, it just attempts to reencode as jpg. should allow way more files at this critical step (gif, webp, heic etc) and only rejects if absolutely necessary.

## Test plan

- in `saveImageToMediaLibrary`, add `uri = uri.replace('@jpeg', '@webp')` to make CDN serve webp. confirm it still works
- go through onboarding, test out uploading various images as a pfp
- test uploading/posting images etc


mozzius:
@claude pls review

PR #9959, opened 2026-02-26
Fix iOS sheet keyboard handling using native scrollview features

mozzius:
Improves our sheet handling, attempting to de-jank them as much as possible. The current implementation is a pile of magic numbers and terribleness.

- Removes `KeyboardAwareScrollView` - this was causing strange and unexpected padding. Now sheets don't expand all the way on iOS when the keyboard opens, only as much as necessary. It didn't even really work anyway and iOS has this built in (although, with a little less space between the input and the keyboard - acceptable compromise imo)
- Removes padding guesswork where possible and adopts native scrollview props of `automaticallyAdjustKeyboardInsets` and `contentInsetAdjustmentBehavior`. This leads to *far* more consistent behaviour, especially for the keyboard case. Previously we were just guessing random numbers for the bottom padding, and now iOS just handles it automatically.
- Fixes the `headerOffset` prop on the `Dialog.FlatListInner` causing unexpected padding.
- cleaned up the alt text dialog a little

What's remaining:
- Android still doesn't automatically adjust the scrollview inset when the keyboard opens. It would be nice if it did, but this isn't a regression

<table>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

## Test plan

we need to make sure that all the sheets work (i.e. nothing goes offscreen, can use keyboard)

on iOS and android:
- test a partial-height sheet (such as the repost dialog)
- test a partial-height sheet with keyboard (such as live now)
- test the report flow (IMPORTANT)
- test a full-height sheet (such as mutewords)
- test a full-height sheet with keyboard (such as alt text)
- test the flatlist sheets:
  - app language select
  - composer language select

### DIFF
--- a/src/view/com/composer/videos/SubtitleDialog.tsx
+++ b/src/view/com/composer/videos/SubtitleDialog.tsx
@@ -60,7 +60,7 @@ export function SubtitleDialogBtn(props: Props) {
           )}
         </ButtonText>
       </Button>
-      <Dialog.Outer control={control}>
+      <Dialog.Outer control={control} nativeOptions={{preventExpansion: true}}>
         <Dialog.Handle />
         <SubtitleDialogInner {...props} />
       </Dialog.Outer>


src/components/Dialog/index.tsx
```diff
@@ -212,33 +208,17 @@ export const ScrollableInner = React.forwardRef<ScrollView, DialogInnerProps>(
...
-    )
+    const isAtMaxSnapPoint = nativeSnapPoint === BottomSheetSnapPoint.Full
 
     let paddingBottom = 0
     if (IS_IOS) {
-      paddingBottom += keyboardHeight / 4
-      if (nativeSnapPoint === BottomSheetSnapPoint.Full) {
-        paddingBottom += insets.bottom + tokens.space.md
-      }
-      paddingBottom = Math.max(paddingBottom, tokens.space._2xl)
```
mozzius, line 235: sooo much jank, gone

vineyardbovines: YEET

mozzius:
Lemme check iOS 18 real quick actually...

mozzius:
Yeah okay as expected.

God, after using 26 for a while, iOS 18 really does feel 8 versions behind...

Simulator Screenshot - iPhone 16 - 2026-02-27 at 23 57 28

PR #9960, opened 2026-02-26
[APP-1819] fix progress guide flash and extra gap

vineyardbovines:
stale server state causes the progress guide to render before the follow query resolves, so it causes a blip and then leaves a UI gap behind


vineyardbovines:
@github-actions ota

mozzius:
Aha, nice. Simple and makes sense

Issue #8086, opened 2025-03-30
[feature request] Copy post and url


juju4:
### Describe the Feature

Currently, can either "Copy post text", either "Share" which gives bluesky post link. It would be great to be able to do both in one action.
Tested on app 1.99.0.983 (ios)
Render something like
```
TEXT with attachments/links url
https://bsky.app/profile/link
```

PR #9961, opened 2026-02-26
Add 'Copy link to post' to post overflow menu

orpheuslummis:
## Summary

- Adds a "Copy link to post" item to the post three-dot (overflow) menu, placed right after "Copy post text"
- Uses the same URL construction (`toShareUrl`) and chain-link icon as the share menu's existing "Copy link" action
- Gives users quicker access to copying the post URL without going through the share sub-menu

This action already exists in the share menu (arrow icon), but users looking for copy actions naturally check the overflow menu where "Copy post text" lives. Grouping both copy actions together improves discoverability.

Related: #8086

## Test plan

- [ ] Open the three-dot menu on any post → "Copy link to post" appears below "Copy post text" with a chain-link icon
- [ ] Tap "Copy link to post" → clipboard contains the full `https://bsky.app/profile/.../post/...` URL
- [ ] Toast "Copied to clipboard" appears
- [ ] Verify on web and native (iOS/Android)
- [ ] Verify the item does not appear for logged-out users viewing `!no-unauthenticated` posts (same gate as "Copy post text")

### DIFF
--- a/src/components/PostControls/PostMenu/PostMenuItems.tsx
+++ b/src/components/PostControls/PostMenu/PostMenuItems.tsx
@@ -65,6 +65,7 @@ import {
 } from '#/components/dialogs/PostInteractionSettingsDialog'
 import {Atom_Stroke2_Corner0_Rounded as AtomIcon} from '#/components/icons/Atom'
 import {BubbleQuestion_Stroke2_Corner0_Rounded as Translate} from '#/components/icons/Bubble'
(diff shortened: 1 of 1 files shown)


surfdude29:
> Gives users quicker access to copying the post URL without going through the share sub-menu

I don't understand this, wouldn't it be the same number of taps/clicks (two)? 🤔

IMHO if adding an option to the post context menu, it would probably make more sense to add **Copy post text and link** as #8086 asks for, because then it's a unique option that adds value :)

PR #9962, opened 2026-02-27
Cache profile view before opening AfterReportDialog

mozzius:
Prevents a loading spinner flicker, which can cause our sheets to go to the wrong height in certain circumstances

## Before (see loading state flicker)

https://github.com/user-attachments/assets/0db22046-e09e-4d53-869c-626981c1e26c

## After

https://github.com/user-attachments/assets/640b28af-35bf-4d4a-abd6-136718d25f5e

### DIFF
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -5,6 +5,7 @@ import {msg} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react'
 import {Trans} from '@lingui/react/macro'
 import {useNavigation} from '@react-navigation/native'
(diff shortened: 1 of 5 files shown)


src/components/dms/ConvoMenu.tsx
```diff
@@ -125,6 +130,12 @@ let ConvoMenu = ({
             }}
             control={reportControl}
             onAfterSubmit={() => {
+              const sender = convo.members.find(
+                member => member.did === latestReportableMessage.sender.did,
+              )
+              if (sender) {
+                unstableCacheProfileView(queryClient, sender)
```
estrattonbailey, line 137: sickos yes

PR #9964, opened 2026-02-27
[APP-1917] fix header measurement for android content disappearing on switch

vineyardbovines:
when `showLoggedOut` toggles, the entire NativeStackView is replaced with `<LoggedOut />` which unmounts all screens, so when the user returns the nav stack remounts but needs to re-measure its header on Android after native view reconstruction. it calls `.measure()` but this can return 0 or never callback, so then `headerOnlyHeight` gets stuck at 0 and then all tab content ends up returning null.

this should also be new-arch / fabric proof since it'll get the height directly from the native layout system.

would definitely love some more tests on this android-wise

### DIFF
--- a/src/view/com/pager/PagerWithHeader.tsx
+++ b/src/view/com/pager/PagerWithHeader.tsx
@@ -16,7 +16,6 @@ import Animated, {
   useSharedValue,
 } from 'react-native-reanimated'
 
-import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {ScrollProvider} from '#/lib/ScrollContext'
 import {
   Pager,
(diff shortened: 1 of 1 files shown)


src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -278,14 +278,24 @@ let PagerTabBar = ({
       <View
         ref={headerRef}
         pointerEvents={IS_IOS ? 'auto' : 'box-none'}
-        collapsable={false}>
+        collapsable={false}
+        onLayout={(e: LayoutChangeEvent) => {
+          // Fallback measurement using onLayout directly on the header wrapper.
+          // This is more reliable than .measure() on Android after certain
+          // navigation transitions (e.g. returning from the logged-out view)
+          // where .measure() can fail to return a height.
```
mozzius: can ya add a link to this PR for context?

src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -284,6 +284,7 @@ let PagerTabBar = ({
           // This is more reliable than .measure() on Android after certain
           // navigation transitions (e.g. returning from the logged-out view)
           // where .measure() can fail to return a height.
+          // ref: https://github.com/bluesky-social/social-app/pull/9964 -sfp
```
mozzius, line 289: wait you're `sfp`? I'm `sfn`, this is gonna get confusing

mozzius (approved):
@vineyardbovines definitely worked, but when I added logging it seems it does fire too early and trigger an extra render with the wrong height, so I pushed up a slightly more conservative approach where it saves it to a ref and only uses it for when `height` is `undefined`. Lemme know what you think!

src/view/com/pager/PagerWithHeader.tsx
vineyardbovines: lmaooooooooooo YES

vineyardbovines:
ooo yep, yeah i see that on my commit. i think a ref is ez- not seeing it on latest

PR #9972, opened 2026-02-28
Fix crash with translate link on web

ds-boyce:
The `index.web.tsx` file was lacking the context provider, so it would always crash.


mozzius (approved):
Tested, works!

PR #9973, opened 2026-03-01
The Great Unjanking of the Sheets

### LINKED ISSUE #6418: Bottom sheets are resizable even if they shouldn't be
### Steps to Reproduce

1. Tap the repost button/the meatball menu on posts/the meatball menu on feed overviews/the meatball menu on accounts/the Muted words & tags setting/the Export My Data setting
2. drag up the sheet
3. It'll stay there for a couple seconds before snapping back to the bottom, often stopping halfway 

### Attachments

_No response_

Android

### Device Info

Google Pixel 7 Pro 

Build version: 1.93.0.373

### Additional Information

I think ideally the bottom sheets shouldn't be resizable at all, unless the point was to make them more reachable, but then you should certainly fix the fact that they keep snapping back down to the bottom of the screen

### LINKED ISSUE #9943: Go Live sheet stuck behind navbar on Android
### Steps to Reproduce

1. Go to profile on Android app
2. Open "Go Live" prompt
3. Type in "live link" field.
4. Keyboard will cover the field, but even if typed correctly, cannot move sheet up at all and will be stuck behind so cannot save live status.

### Attachments

Bottom half of screen showing 'Go Live' section behind android navbar

Android

### Device Info

Google Pixel 7, Android 16

Build version: 1.117.0.825; Bundle info: cb7e2ab (prod); Bundle date: 26022020; Platform: android; Platform version: 36; Device ID: 68473ce2-0471-43eb-b51b-4a7a2cc8284d

### Additional Information

_No response_

### LINKED ISSUE #9969: Pulling up on the new Report sheet resets it twice
### Steps to Reproduce

1. Go to report anything 
2. Pull the bottom sheet up
3. It resets heights twice

### Attachments

https://github.com/user-attachments/assets/b21f5ae3-0729-47c9-b7f6-2f19181e660f

Android

### Device Info

Pixel 8, Android 16

Build version: 1.117.0.825; Bundle info: c6bcc9a (prod); Bundle date: 26022614; Platform: android; Platform version: 36; Device ID: a8102095-ac3e-4842-83f7-b2b0acddf889

### Additional Information

Ideally, I shouldn't be able to pull up on the Report sheet at all to expand it up, especially if it'll just reset itself (twice). The other sheets (blocking, etc.) from what I've tested don't have this behavior.

mozzius:
I finally did it. Sheets are no longer janky.

image

Fixes #6418 
Fixes #9943 
Fixes #9969

Most importantly, the bottom padding is now _completely predictable_. We can now just set a consistent `a.pb_2xl` at the bottom edge, and it will be guaranteed to be respected regardless of sheet height, keyboard state, etc etc. The keyboard also now just pushes the sheet content up on Android (matching iOS), so we no longer need to fear about text inputs being inaccessible.

## Summary

Major overhaul of the bottom sheet module: moves content height detection from JS to native, adds a `fullHeight` prop, fixes several Android `BottomSheetBehavior` issues, and replaces native keyboard handling with JS-side keyboard padding.

### Native content height observation

Previously, content height changes were detected via JS `onLayout` callbacks, which required a round-trip through the JS bridge before the sheet could resize. This caused visible lag and, on Android, required a hacky `setTimeout` workaround because `updateLayout()` wouldn't fire promptly.

Now both platforms observe content height changes purely on the native side:
- **iOS**: KVO observation on the content view's `bounds`. Calls `updateDetents` directly when the height changes, bypassing the old `prevLayoutDetentIdentifier` guard that could block legitimate updates during animations.
- **Android**: `OnLayoutChangeListener` on each direct child of the inner view. We use `OnLayoutChangeListener` (not `OnGlobalLayoutListener`) because React Native calls `view.layout()` directly via Yoga, bypassing `requestLayout()`/`performTraversals()`. `OnLayoutChangeListener` fires from `setFrame()` which IS called by `layout()`, so it catches RN-driven updates.

The JS-side `onLayout` → `updateLayout()` bridge path is removed entirely. The `BottomSheetNativeComponent` `onLayout` is now only used for the iOS 15 `viewHeight` workaround.

Also fixes `getContentHeight()` on Android to use the tallest direct child's height rather than a fragile `allViews` iterator — the handle is absolutely positioned and overlaps the content, so summing would double-count.

### `fullHeight` prop (Android + iOS)

Sheets that need to be full-height previously used `minHeight: screenHeight`, which forced `getTargetHeight()` to return the screen height. On Android this caused a gap at the top (the sheet couldn't quite reach the screen edge due to `isFitToContents=true` positioning based on measured content), and the gap swallowed the first swipe.

The new `fullHeight` prop skips dynamic content height entirely:
- **iOS**: Sets detent to `.large`, skips content height observation
- **Android**: Uses `isFitToContents=false` with `expandedOffset=statusBarHeight`, skips content observation. The `expandedOffset` API directly controls where the sheet sits when expanded, avoiding all the `maxHeight` gymnastics

### Android behavior overhaul

**3-state snapping for normal sheets**: Switched from `isFitToContents=true` to `isFitToContents=false` with `expandedOffset=statusBarHeight`. Previously, `isFitToContents=true` prevented the `BottomSheetBehavior` from settling at `STATE_HALF_EXPANDED` during gestures — once a user swiped a sheet to full height, they could never swipe it back down to the content-sized detent. With `isFitToContents=false`, all three snap points (half-expanded ↔ expanded ↔ hidden) work properly.

**Deferred layout updates during gestures**: `updateLayout()` now defers ratio/state changes while the behavior is in `STATE_DRAGGING` or `STATE_SETTLING`. The updated values are applied once the gesture completes. Without this, forcing a state change mid-drag interrupts dismiss swipes and causes visual glitches.

**`preventExpansion` safety net**: `preventExpansion` sheets keep `isFitToContents=true` with `maxHeight` to physically prevent drag-up. A state callback bounces `EXPANDED→HALF_EXPANDED` as a safety net for an edge case where rapid content resizes during the opening animation can invert the expanded/half-expanded offsets (they end up ~11px apart but in the wrong order), causing the behavior to cascade to `HIDDEN` and dismiss the sheet.

**`requestLayout()` after `maxHeight` changes**: When `updateLayout()` changes `maxHeight` for `preventExpansion` sheets, the `FrameLayout` needs to re-measure. Without `requestLayout()`, the sheet's physical height stays at its old value even though `maxHeight` changed — this was preventing keyboard padding from actually resizing the sheet.

**Transparent `backgroundTint`**: The Material3 `BottomSheetBehavior` applies a `MaterialShapeDrawable` with a surface tint during `onLayoutChild`, overriding our `setBackgroundColor(0)` call. Setting `backgroundTint=transparent` in the theme removes the tint that was visible in gaps between the sheet and screen edge.

### Keyboard handling

Removed all native keyboard handling (insets listener, `isKeyboardVisible` tracking, manual `STATE_EXPANDED` on keyboard show). The old approach had several issues:
- Race condition: if a sheet opened while the keyboard was hiding, `imeVisible` was still `true` during the animation, causing a spurious expand→collapse
- Fighting with content observation: expanding the sheet on keyboard show triggered content re-measurement, which triggered `updateLayout()`, which sometimes fought the expansion

Replaced with JS-side keyboard padding: `ScrollableInner` listens for RN `Keyboard` events and adds bottom padding equal to the keyboard height on Android. The content observer detects the height change and resizes the sheet naturally through the existing `updateLayout()` flow. For `preventExpansion` sheets, the `requestLayout()` call ensures the `FrameLayout` physically grows to accommodate the padding.

Generalized `useOnKeyboardDidShow` → `useOnKeyboard(eventName, cb)` to support both show and hide events.

### Callers

Replaced `minHeight: screenHeight` with `fullHeight: true` in 16 dialogs and removed the now-unnecessary `useWindowDimensions` imports. Cleaned up a few other things:
- `ServerInputDialog`: Simplified from platform-specific `{android: {minHeight: height/2}, ios: {preventExpansion: true}}` to just `{preventExpansion: true}`
- `ImageAltTextDialog`: Removed old `{height: 300}` keyboard spacer hack

## Test plan

tl;dr test every sheet 😥

- [ ] Open Edit Profile, Change Handle, Add App Password, Image Alt Text, GIF Alt Text dialogs on **Android** — should open at full height with no gap at the top, dismissible by swipe
- [ ] Same on **iOS** — should open at `.large` detent
- [ ] Open a non-fullHeight sheet (e.g. post report menu) on **Android** — should open at content height. Swipe up to full height, then swipe back down — should snap back to content height (not just dismiss)
- [ ] Open a `preventExpansion` sheet (e.g. post interaction settings) on **Android** — should NOT be swipeable to full height
- [ ] Open a sheet with a text input on **Android** — keyboard should appear and the sheet should resize to accommodate it. Content should be scrollable above the keyboard
- [ ] Dismiss keyboard, then immediately open a sheet — should NOT briefly expand then collapse
- [ ] Open a sheet with rapidly changing content (e.g. post interaction settings with spinner) on **Android** — should not dismiss unexpectedly
- [ ] On **iOS**, sheets should resize smoothly when content changes (e.g. loading states)
- [ ] All existing sheet behavior on **iOS** should be unchanged

### DIFF
--- a/modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
+++ b/modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
@@ -34,6 +34,10 @@ const IS_IOS15 =
   Platform.OS === 'ios' &&
   // semvar - can be 3 segments, so can't use Number(Platform.Version)
   Number(Platform.Version.split('.').at(0)) < 16
+// older android versions (15 and below) aren't naturally edge-to-edge
+// and behave a little differently
+const IS_NON_E2E_ANDROID =
+  Platform.OS === 'android' && Number(Platform.Version) < 35
 
 export class BottomSheetNativeComponent extends React.Component<
   BottomSheetViewProps,
@@ -142,6 +146,13 @@ function BottomSheetNativeComponentInner({
   const cornerRadius = rest.cornerRadius ?? 0
   const {height: screenHeight} = useWindowDimensions()
 
+  // sigh... on older Android versions, screenHeight does not include safe area insets
+  // on newer Androids + iOS, it does. we need to find the inner bit + the bottom inset
+  // for the sheet content
+  const sheetHeight = IS_NON_E2E_ANDROID
+    ? screenHeight + insets.bottom
+    : screenHeight - insets.top
+
   return (
     <NativeView
       {...rest}
@@ -149,7 +160,7 @@ function BottomSheetNativeComponentInner({
       ref={nativeViewRef}
       style={{
         position: 'absolute',
-        height: screenHeight - insets.top,
+        height: sheetHeight,
         width: '100%',
       }}
       containerBackgroundColor={backgroundColor}>
--- a/modules/bottom-sheet/android/src/main/java/expo/modules/bottomsheet/BottomSheetView.kt
+++ b/modules/bottom-sheet/android/src/main/java/expo/modules/bottomsheet/BottomSheetView.kt
@@ -38,9 +38,13 @@ class BottomSheetView(
   private var lastObservedContentHeight: Float = 0f
   private var pendingLayoutUpdate: Boolean = false
 
-  private val screenHeight =
-    context.resources.displayMetrics.heightPixels
-      .toFloat()
+  private val screenHeight: Float =
+    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+      context.resources.displayMetrics.heightPixels.toFloat()
+    } else {
+      val wm = context.getSystemService(Context.WINDOW_SERVICE) as android.view.WindowManager
+      wm.currentWindowMetrics.bounds.height().toFloat()
+    }
 
   private fun getNavigationBarHeight(): Int {
     val resourceId = resources.getIdentifier("navigation_bar_height", "dimen", "android")
@@ -203,6 +207,7 @@ class BottomSheetView(
     val bottomSheet = dialog.findViewById<FrameLayout>(com.google.android.material.R.id.design_bottom_sheet)
     bottomSheet?.let {
       it.setBackgroundColor(0)
+      it.elevation = 0f
 
       val behavior = BottomSheetBehavior.from(it)
       behavior.state = BottomSheetBehavior.STATE_HIDDEN
--- a/modules/bottom-sheet/android/src/main/res/values/styles.xml
+++ b/modules/bottom-sheet/android/src/main/res/values/styles.xml
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <style name="EdgeToEdgeBottomSheetDialogTheme" parent="Theme.Material3.DayNight.BottomSheetDialog">
-        <!-- Enable edge-to-edge -->
+    <style name="EdgeToEdgeBottomSheetDialogTheme" parent="ThemeOverlay.Material3.DayNight.BottomSheetDialog">
+        <!-- Enable edge-to-edge, matching react-native-edge-to-edge's setup -->
         <item name="android:navigationBarColor">@android:color/transparent</item>
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:windowIsFloating">false</item>
+        <item name="android:windowDrawsSystemBarBackgrounds">true</item>
+        <item name="android:fitsSystemWindows">false</item>
         <item name="enableEdgeToEdge">true</item>
 
         <!-- Configure bottom sheet to respect system window insets -->


modules/bottom-sheet/src/BottomSheetNativeComponent.tsx
```diff
@@ -150,22 +138,20 @@ function BottomSheetNativeComponentInner({
...
   ) => void
   nativeViewRef: React.RefObject<View>
-  onLayout: (event: LayoutChangeEvent) => void
+  onLayout?: (event: LayoutChangeEvent) => void
 }) {
   const insets = useSafeAreaInsets()
   const cornerRadius = rest.cornerRadius ?? 0
   const {height: screenHeight} = useWindowDimensions()
 
-  const sheetHeight = IS_IOS ? screenHeight - insets.top : screenHeight
```
mozzius, line 159: turns out a *ton* of the positioning problems were due to not doing this on Android too

src/view/com/composer/photos/ImageAltTextDialog.tsx
```diff
@@ -177,8 +173,6 @@ const ImageAltTextInner = ({
           </Button>
         </AltTextCounterWrapper>
       </View>
-      {/* Maybe fix this later -h */}
-      {IS_ANDROID && isKeyboardVisible ? <View style={{height: 300}} /> : null}
```
mozzius, line 183: 1 year later...

estrattonbailey:
I think what I'm seeing here is:
- all sheets have some edge-to-edge issue at the bottom when first opening
- fixed height sheets snap down to correct position when interacted with
- flex height sheets expand just fine, but are lacking a background

For the last point, an early version of this component just had a 250% height bg element that extends off the screen. Could be an option.

https://github.com/user-attachments/assets/86320124-599d-4093-9dc2-44bc1aaa81a4

https://github.com/user-attachments/assets/e0dea60d-1423-4c55-8a03-ae760b382b17

estrattonbailey:
May not be able to do anything about this, but the mid-snap point on Android is verrrrry sensitive. Basically don't know that it's there.

https://github.com/user-attachments/assets/b80c5e18-ae02-4d99-96c5-f89207a8e923

estrattonbailey:
Had a crash on iOS (built to device locally):
- opened report dialog and clicked around
- closed, immediately opened Mute Words
- crashed before it opened all the way

mozzius:
Whaaaaat okay what you see on Android is very different to what I see. I was getting it at one point but I thought I fixed it

Lemme take a look and I'll let you know what I find!

src/components/Dialog/index.tsx
```diff
@@ -238,7 +246,12 @@ export const ScrollableInner = React.forwardRef<ScrollView, DialogInnerProps>(
         contentContainerStyle={[
           a.pt_2xl,
           IS_LIQUID_GLASS ? a.px_2xl : a.px_xl,
-          {paddingBottom},
+          platform({
+            ios: a.pb_2xl,
+            android: {
+              paddingBottom: keyboardHeight + insets.bottom + tokens.space.xl,
+            },
```
estrattonbailey, line 253: (note partially for myself) why don't we handle keyboard the same on iOS?

mozzius: iOS has built-in handling via automaticallyAdjustKeyboardInsets. If we insert the padding ourselves, it will make the partial-height sheets too tall

PR #9974, opened 2026-03-01
Replace lande with native language detection

mozzius:
## Summary
- Replace the JS-only `lande` library with `@bsky.app/expo-guess-language`, which uses on-device ML Kit (Android) / NLLanguageRecognizer (iOS) for language detection in the composer, with lande as a sync fallback for feed filtering
- Composer language suggestion now uses the async native API (`guessLanguageAsync`) for better accuracy on mobile
- Per-platform confidence thresholds: web stays strict (0.97), native drops to 0.9 (0.8 for device locales) since native models are significantly more accurate
- Device-locale advantage: languages installed on the user's device get a lower noise floor, making them less likely to be pruned as candidates

## Test plan
- [ ] Type 30+ characters in different languages on iOS, Android, and web
- [ ] Verify language suggestions appear at expected thresholds on each platform
- [ ] Verify feed language filtering still works correctly (posts are not incorrectly hidden)
- [ ] Verify reply-language suggestion still appears when replying to a post in a different language


src/locale/helpers.ts
```diff
@@ -79,6 +80,7 @@ export function codeToLanguageName(lang2or3: string, appLang: string): string {
 
 export function getPostLanguage(
   post: AppBskyFeedDefs.PostView,
+  enableNativeDetection?: boolean,
```
ds-boyce: Can we make this an `options` object to avoid boolean params? Something like this:

```ts
export function getPostLanguage(
  post: AppBskyFeedDefs.PostView,
  options?: {
    enableNativeDetection?: boolean
  },
): string | undefined {
  const enableNativeDetection = options?.enableNativeDetection ?? false;
```

ds-boyce (approved):
Let’s land this, then we can update the composer’s language detection to use `guessLanguageSync` and improve the result at write-time. The goal being to have more confidence in a post’s language tags going forward, so that we can remove detection from `getPostLanguage` at read-time.

We’ll need to monitor that, so I’ll update #10014 after this is merged.

src/locale/helpers.ts
```diff
@@ -105,27 +107,41 @@ export function getPostLanguage(
...
+        ([lang, _probability]: [string, number]) => {
+          return candidates.includes(code3ToCode2(lang))
+        },
+      )
+    }
+
+    if (langsProbabilityMap[0]) {
+      return code3ToCode2(langsProbabilityMap[0][0])
+    }
   }
```
ds-boyce: What do you think of removing this block and trusting the language tags 100%?

estrattonbailey (commented):
I can't get this to build to my iPhone locally, just getting a generic `APIInternalError`. Might be my env so not blocking, but FYI before this gets merged.

ds-boyce:
I was able to build iOS locally and run in a simulator and on-device. @mozzius want to confirm on your end as well?

src/locale/helpers.ts
ds-boyce: I guess we still need Lande for web until `LanguageDetector` is available. 😞

estrattonbailey (changes requested):
Ok this looks solid to me, seems to Just Work™. I think we should do a few things before this lands though:
- In addition to "Yes", we should have a "No". Or perhaps a ✅ and a ❌  to save space.
- If user selects "No", the prompt should not show up again
- Add metrics with similar clarity to #10061, tracking success rates of this prompt (when shown)
- Invert the feature flag so we can slow roll this

ds-boyce:
@estrattonbailey How’s this look?

suggestion

src/analytics/metrics/types.ts
```diff
@@ -796,6 +796,60 @@ export type Events = {
...
+    os: Platform['OS']
+    /**
+     * The language we detected and suggested to the user as an override for the
+     * expected target language.
+     */
+    suggestedLanguage: string | undefined
+    /**
+     * This is the user's configured primary language, which is always defined.
+     */
+    expectedTargetLanguage: string
```
estrattonbailey: Do we need this prop? Not sure it makes sense here. If anything, it could be like, `currentTargetLanguage`, which can vary depending on what the user last posted in.

src/analytics/metrics/types.ts
```diff
@@ -796,6 +796,60 @@ export type Events = {
...
```
estrattonbailey: Same here and below

src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -48,29 +79,70 @@ export function SuggestedLanguage({
...
+      // eslint-disable-next-line react-hooks/set-state-in-effect
       setSuggestedLanguage(undefined)
       return
     }
 
-    const idle = onIdle(() => {
-      setSuggestedLanguage(guessLanguage(textTrimmed))
-    })
+    const idle = onIdle(
+      () =>
```
estrattonbailey: If they've declined, we could either not run this again to save performance, or not run it again for the same language

src/view/com/composer/select-language/SuggestedLanguage.tsx
```diff
@@ -175,41 +255,80 @@ function LanguageSuggestionButton({
...
-          <ButtonText>
-            <Trans>Yes</Trans>
-          </ButtonText>
+          label={l`Accept this language suggestion`}>
+          <ButtonIcon icon={CheckIcon} size="sm" />
+        </Button>
+
+        <Button
+          size="small"
+          color="secondary"
```
estrattonbailey, line 476: Check out `shape='round'`

ds-boyce:
Updated.

composer

estrattonbailey (changes requested):
We're pretty close here! @ds-boyce have a look at my last commit and verify that works alright. Still not psyched on running the language model so often, maybe we can think of something to prevent that 🤔 

Also this doesn't run on web right now.

ds-boyce:
I added a 500ms debounce for the input text before we attempt to guess the language.

Also put up https://github.com/bluesky-social/expo-guess-language/pull/1 to fix web.

estrattonbailey:
@claude give this another pass with a keen eye for bugs and cases where we might annoy the user by showing the prompt too often.

estrattonbailey (approved):

PR #9979, opened 2026-03-02
Force relayout to fix stuck header blur

mozzius:
The header blur introduced by #9935 has a bug where scrolling down, then swiping to the side causes the text in the tab bar to be affected by the blur. This is because when invisible, it sits in the blur area and gets blurred. Transforming the element over into place doesn't consistently clear the blur, especially when swiping between pages.

The fix is not great, but forcing a relayout by changing `top` by 1px seems to do it. 

If you have a better idea for forcing a relayout, I'm all ears. zIndex did not work.

## Before

https://github.com/user-attachments/assets/149cc722-7ad3-42d1-babf-d0fd4e21591f

## After

https://github.com/user-attachments/assets/800af664-6263-4d54-882e-6cde6525970a

### DIFF
--- a/src/lib/hooks/useMinimalShellTransform.ts
+++ b/src/lib/hooks/useMinimalShellTransform.ts
@@ -16,16 +16,39 @@ export function useMinimalShellHeaderTransform() {
 
   const headerTransform = useAnimatedStyle(() => {
     const headerModeValue = headerMode.get()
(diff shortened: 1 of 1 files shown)


vineyardbovines (approved):
i can't think of a diff way to do it if zindex wasn't working ¯\\_(ツ)_/¯

PR #9980, opened 2026-03-02
Translation followups

mozzius:
Followups to #9930 

Few changes:
- change "Edit" to "Change" - bit clearer imo
- revert the change to use localised language names in favour of `Intl.DisplayNames`, make sure it's used consistently.
- (not related to PR) optimise `Intl.DisplayNames` usage by caching instances

The PR changed the language names to be the name for a language in its own name. This makes sense when looking for your *own* language, and we do indeed do that in the App Language dropdown. However, for translate source/content language/post language etc etc, we want to show the name for each language in the _users'_ language - for which we have a utility already, `languageName`, which uses the `Intl.DisplayNames` API.

However, I don't think we were using it consistently enough before, so I went ahead and made sure we it everywhere thats needed.

## Before / After

Simulator Screenshot - iPhone 17 - 2026-03-02 at 13 55 06
Simulator Screenshot - iPhone 17 - 2026-03-02 at 13 52 07

### DIFF
--- a/src/components/dialogs/LanguageSelectDialog.tsx
+++ b/src/components/dialogs/LanguageSelectDialog.tsx
@@ -302,6 +302,7 @@ export function DialogInner({
             )
           }
           const lang = item.lang
+          const name = languageName(lang, langPrefs.appLanguage)
 
           const isLastItem = index === numItems - 1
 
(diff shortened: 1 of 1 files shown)


surfdude29:
Not directly related to this PR, but I just noticed that the header text in the Content Languages and Post Languages dialogs doesn't wrap on mobile web:

IMG_2243
IMG_2244
IMG_2255

PR #9982, opened 2026-03-03
Fix translate hitslop by using Button instead of Text

mozzius:
`<Text>` does not have the `hitSlop` prop, so we need to use a real pressable instead. I also increased it to 30px, because the tap target is _really_ small.

## Test plan

Try tapping the Translate/Change/Hide buttons - should be more forgiving


estrattonbailey:
D hit this in #9987

PR #9985, opened 2026-03-03
Add on-device translation to feeds/replies/etc via PostMenu

mozzius:
This PR implements my vision for the provider system that I don't think I explained properly. Basically, we add this hook to the Translate option in `PostMenuItems`

```ts
/**
 * Triggers the on-device translation, if within the TranslateOnDeviceProvider.
 * Otherwise, falls back to linking out to Google Translate.
 */
export function useMaybeTranslateOnDevice() {
  const onDeviceTranslation = useContext(Context)
  const fallbackTranslation = useTranslate()

  return useCallback(
    (text: string, targetLangCode: string, sourceLangCode?: string) => {
      if (onDeviceTranslation) {
        return onDeviceTranslation.translate(
          text,
          targetLangCode,
          sourceLangCode,
        )
      }
      return fallbackTranslation(text, targetLangCode, sourceLangCode)
    },
    [onDeviceTranslation, fallbackTranslation],
  )
}
```

We do conditional logic depending on whether the component is within a `TranslateOnDeviceProvider`. Then, in the places we want the Post Menu to trigger an on-device translation, we wrap the post component in the provider and add an outlet. Simple as doing this:

Screenshot 2026-03-03 at 16 25 50

This will then cause the post menu item to trigger the on-device version. But if not (i.e. in the video feed feed) it gracefully falls back to googie translate

https://github.com/user-attachments/assets/8cfbbffc-ad1b-4b18-877b-0ce3cc5a1aea

### DIFF
--- a/src/view/com/post/Post.tsx
+++ b/src/view/com/post/Post.tsx
@@ -33,9 +33,11 @@ import {PostAlerts} from '#/components/moderation/PostAlerts'
 import {Embed, PostEmbedViewContext} from '#/components/Post/Embed'
 import {PostRepliedTo} from '#/components/Post/PostRepliedTo'
 import {ShowMoreTextButton} from '#/components/Post/ShowMoreTextButton'
+import {TranslatedPost} from '#/components/Post/Translated'
 import {PostControls} from '#/components/PostControls'
 import {RichText} from '#/components/RichText'
 import {SubtleHover} from '#/components/SubtleHover'
+import {Provider as OnDeviceTranslationProvider} from '#/translation'
 import * as bsky from '#/types/bsky'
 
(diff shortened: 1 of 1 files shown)


src/translation/index.tsx
```diff
@@ -95,11 +96,7 @@ const Context = createContext<{
     sourceLangCode?: string,
   ) => Promise<void>
   clearTranslation: () => void
-}>({
-  translationState: IDLE,
-  translate: async () => {},
-  clearTranslation: () => {},
-})
```
mozzius, line 102: Needed so we can distinguish between being inside or outside the provider, and react accordingly

PR #9986, opened 2026-03-03
[APP-1932] fix scroll header jumping on profiles with large headers

vineyardbovines:
i think there are a few things at play here:

1. in `ExpoScrollForwarder`, every pan gesture on the header set `contentOffset.y` to `0` when it was negative. since the list's position "at top" is a negative contentOffset, it would snap past the entire header on profiles with large headers. 
   - updated this to record `max(contentOffset.y, 0) as the gesture starting point without mutating scroll position

2. the previous heuristic in `PagerWithHeader` rejected scroll events at exactly `-headerHeight`, but with large headers we can get in a position where >400px is a valid position. 
    - updated this to keep filtering the first event on initial load, and then accept all subsequent events

3. the header height was measured once via `onLayout` and then never updated and the parent `onLayout` fires when the header resizes (from suggested follows guide, etc.) 
    - updated this to also call `onHeaderOnlyLayout` when the height differs from the current value so `headerOnlyHeight` is in sync

### DIFF
--- a/modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
+++ b/modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
@@ -77,8 +77,9 @@ class ExpoScrollForwarderView: ExpoView, UIGestureRecognizerDelegate {
 
     if sender.state == .began {
       // Use max(0) so gesture deltas compute correctly, but don't mutate
-      // contentOffset — snapping it to 0 causes a visible jump on profiles
-      // with large headers where the user is scrolled into negative territory.
+      // contentOffset
+      // snapping it to 0 causes a visible jump on profiles with large headers
+      // where the user is scrolled into negative territory. - sfp
       self.initialOffset = max(sv.contentOffset.y, 0)
     }
 
--- a/modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
+++ b/modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
@@ -76,11 +76,10 @@ class ExpoScrollForwarderView: ExpoView, UIGestureRecognizerDelegate {
     let translation = sender.translation(in: self).y
 
     if sender.state == .began {
-      if sv.contentOffset.y < 0 {
-        sv.contentOffset.y = 0
-      }
-
-      self.initialOffset = sv.contentOffset.y
+      // Use max(0) so gesture deltas compute correctly, but don't mutate
+      // contentOffset — snapping it to 0 causes a visible jump on profiles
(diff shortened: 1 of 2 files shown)


src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -288,7 +297,15 @@ let PagerTabBar = ({
...
           if (isHeaderReady) {
-            fallbackHeaderOnlyHeight.current = e.nativeEvent.layout.height
+            fallbackHeaderOnlyHeight.current = height
+            // Re-measure when the header content changes size (e.g.
+            // SuggestedFollows accordion expanding/collapsing). The sentinel
+            // view below only fires onLayout once on mount, so without this
+            // the headerOnlyHeight goes stale. - sfp
+            const rounded = Math.round(height * 2) / 2
+            if (rounded > 0 && rounded !== headerOnlyHeight) {
+              onHeaderOnlyLayout(height)
```
mozzius: I have a counterproposal in https://github.com/bluesky-social/social-app/pull/9990 - you're correct that we need to monitor for changes, but the iteration above fires too early. I think we can greatly simplify things, lemme know what you think (it's stacked on this PR)

src/view/com/pager/PagerWithHeader.tsx
```diff
@@ -156,20 +156,28 @@ export function PagerWithHeader({
...
+        if (isPossiblyInvalid) {
+          return
+        }
+        hasReceivedScroll.set(true)
       }
+      scrollY.set(nextScrollY)
     },
-    [scrollY, headerHeight],
+    [scrollY, headerHeight, hasReceivedScroll],
   )
```
mozzius, line 187: Not sure sure this part works. The problem is that it's not just the first render, I get a few bogus events back to back. It also seems to happen again after the header height changes.
Image

modules/expo-scroll-forwarder/ios/ExpoScrollForwarderView.swift
```diff
@@ -76,11 +76,11 @@ class ExpoScrollForwarderView: ExpoView, UIGestureRecognizerDelegate {
...
-        sv.contentOffset.y = 0
-      }
-
-      self.initialOffset = sv.contentOffset.y
+      // Use max(0) so gesture deltas compute correctly, but don't mutate
+      // contentOffset
+      // snapping it to 0 causes a visible jump on profiles with large headers
+      // where the user is scrolled into negative territory. - sfp
+      self.initialOffset = max(sv.contentOffset.y, 0)
     }
```
mozzius, line 84: need to think about this but it makes sense

mozzius (changes requested):

PR #9987, opened 2026-03-03
Refine UX for on-device translation for posts

ds-boyce:
* Adds error state with Google Translate link (vs opening Google Translate automatically).
* Adds ability to use on-device translations from feeds.
* Adds on-device translation for replies and threaded replies.
* Translations are cleaned up automatically when the component rendering them loses focus.
* Updates presentation of translated text.

https://github.com/user-attachments/assets/6e51a36c-a6f9-41ab-9dd7-8488c9e4abd0

error

### DIFF
--- a/src/components/Post/Translated/index.tsx
+++ b/src/components/Post/Translated/index.tsx
@@ -70,6 +70,7 @@ export function TranslatedPost({
           <TranslationLink
             postText={postText}
             primaryLanguage={langPrefs.primaryLanguage}
+            sourceLanguage={postLanguage}
             translate={translate}
           />
         )
@@ -95,10 +96,12 @@ function TranslationLoading() {
 function TranslationLink({
   postText,
   primaryLanguage,
+  sourceLanguage,
   translate,
 }: {
   postText: string
   primaryLanguage: string
+  sourceLanguage: string | null
   translate: TranslationFunction
 }) {
   const t = useTheme()
@@ -112,11 +115,11 @@ function TranslationLink({
     })
 
     ax.metric('translate', {
-      sourceLanguages: [],
+      sourceLanguages: sourceLanguage ? [sourceLanguage] : [],
       targetLanguage: primaryLanguage,
       textLength: postText.length,
     })
-  }, [ax, postText, primaryLanguage, translate])
+  }, [ax, postText, primaryLanguage, translate, sourceLanguage])
 
   return (
     <View


src/view/shell/index.tsx
```diff
@@ -240,7 +241,9 @@ export function Shell() {
             <NoAccessScreen />
           ) : (
             <RoutesContainer>
-              <ShellInner />
+              <TranslateOnDeviceProvider>
+                <ShellInner />
+              </TranslateOnDeviceProvider>
```
estrattonbailey: Let's just put these in `App.<platform>.tsx` and wrap the `Shell`

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -547,6 +538,7 @@ const ThreadItemAnchorInner = memo(function ThreadItemAnchorInner({
                 feedContext={postSource?.post?.feedContext}
                 reqId={postSource?.post?.reqId}
                 viaRepost={viaRepost}
+                googleTranslate={false}
```
estrattonbailey: Nit: this defaults to `false` already

src/lib/translation/types.ts
```diff
@@ -0,0 +1,18 @@
...
+      translatedText: string
+      sourceLanguage: TranslationTaskResult['sourceLanguage']
+      targetLanguage: TranslationTaskResult['targetLanguage']
+    }
+
+export type Options = {
+  /**
+   * Whether to force the use of Google Translate. Default is false.
+   */
+  googleTranslate?: boolean
```
estrattonbailey: I think a better name for this would be `forceGoogleTranslate`. That indicates a little better how it's intended to be used imo, and makes it more clear when applied to a component has a value-less boolean attr.

Should match that everywhere we use this prop 👍

src/components/Post/Translated/index.tsx
```diff
@@ -1,38 +1,41 @@
...
+  translationKey,
   postText,
-  hideLoading = false,
 }: {
+  translationKey: string
   postText: string
-  hideLoading: boolean
 }) {
   const {translationState} = useTranslateOnDevice()
+  // Register this component as using this translation key with focus-based cleanup
```
estrattonbailey: What do you think about combining these hooks? This component doesn't need to react to state changes not associated with the `translationKey`. And you could do the ref counting automatically.

Most basic idea: a new hook that wraps both of these like `useTranslationByKey(uri)` and just returns the translation state of this post.

src/components/Post/Translated/index.tsx
```diff
@@ -98,14 +108,16 @@ function TranslationResult({
...
 }: {
+  translationKey: string
   postText: string
   sourceLanguage: string
 }) {
   const ax = useAnalytics()
-  const {_} = useLingui()
+  const {t: l} = useLingui()
   const langPrefs = useLanguagePrefs()
   const {translate} = useTranslateOnDevice()
```
estrattonbailey: Really, that same hook could provide a shadowed `translate` method so that you could just create one instance of this hook and pass the `translate` method down. Something to noodle on anyway 🤔

src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -255,11 +254,19 @@ let PostMenuItems = ({
...
-    void translate(record.text, langPrefs.primaryLanguage)
+    void translate(
+      translationKey,
+      record.text,
+      langPrefs.primaryLanguage,
+      undefined,
+      {
+        googleTranslate,
+      },
+    )
```
estrattonbailey: Nit: I think we should avoid positional args in favor of object destructuring. Makes for easier to maintain APIs and clearer property tracing.

src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -133,7 +135,7 @@ let PostMenuItems = ({
   const {hidePost} = useHiddenPostsApi()
   const feedFeedback = useFeedFeedbackContext()
   const openLink = useOpenLink()
-  const translate = useTranslate()
+  const {clearTranslation, translate, translationState} = useTranslateOnDevice()
```
estrattonbailey: Yeah would be nice imo to have this hook already know what `translationKey` we're using instead of having to pass that around

src/lib/translation/index.tsx
```diff
@@ -120,25 +89,108 @@ export function useTranslateOnDevice() {
...
+  )
+}
+
 export function Provider({children}: React.PropsWithChildren<unknown>) {
-  const [translationState, setTranslationState] =
-    useState<TranslationState>(IDLE)
+  const [translationState, setTranslationState] = useState<
+    Record<string, TranslationState>
+  >({})
+  const [refCounts, setRefCounts] = useState<Record<string, number>>({})
```
estrattonbailey, line 149: As a last thing, we should look at re-renders for this. I think we could get away with a `useRef` here, but let's tackle that last.

src/screens/PostThread/components/ThreadItemAnchor.tsx
```diff
@@ -420,8 +409,10 @@ const ThreadItemAnchorInner = memo(function ThreadItemAnchorInner({
                 shouldProxyLinks={true}
               />
             ) : undefined}
-            <TranslatedPost postText={record.text} hideLoading />
-            <TranslateLink post={item.value.post} />
+            <TranslatedPost
+              translationKey={translationKey}
```
estrattonbailey: Nit: since this is a clearly named property and we don't use `translationKey` elsewhere in this file, I think we can just inline it

src/view/com/post/Post.tsx
estrattonbailey, line 1: Same nits in this file

src/lib/translation/index.tsx
```diff
@@ -120,25 +89,108 @@ export function useTranslateOnDevice() {
...
       sourceLangCode?: string,
+      options?: Options,
     ) => {
-      setTranslationState({status: 'loading'})
+      const translateUrl = getTranslatorLink(
+        text,
+        targetLangCode,
+        sourceLangCode,
+      )
+      if (options?.googleTranslate) {
```
estrattonbailey: We should include the activity launcher handling from `useTranslate` here too. Really, you could probably just use the renamed `useGoogleTranslate` within this logical fork.

src/lib/translation/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+  // language packs must be downloaded and installed.
+  if (Platform.OS === 'ios') {
+    const deviceLocales = getLocales()
+    const primaryLanguageTag = deviceLocales[0]?.languageTag
+    switch (targetLangCodeOriginal) {
+      case 'en': // en-US, en-GB
+      case 'es': // es-419, es-ES
+      case 'pt': // pt-BR, pt-PT
+      case 'zh': // zh-Hans-CN, zh-Hant-HK, zh-Hant-TW
+        targetLangCode = primaryLanguageTag ?? targetLangCodeOriginal
```
mozzius: nit: they might have changed their app language

```suggestion
        if (primaryLanguageTag && primaryLanguageTag.startsWith(targetLangCodeOriginal)) {
          targetLangCode = primaryLanguageTag
        }
```

src/lib/translation/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+        break
+    }
+  }
+
+  const {onTranslateTask} =
+    // Needed in order to type check the dynamically imported module.
+    // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+    require('@bsky.app/expo-translate-text') as typeof import('@bsky.app/expo-translate-text')
```
mozzius: I don't think we really need to dynamically import this. it's not very heavy

src/components/Post/Translated/index.tsx
```diff
@@ -1,113 +1,331 @@
...
+        </View>
+      </View>
+      <View style={[a.flex_row, a.align_center]}>
+        <Text>
+          <InlineLinkText
+            to={getTranslatorLink(postText, primaryLanguage)}
+            label={l`Try Google Translate`}
+            style={[a.text_xs, a.font_medium]}>
+            <Trans>Try Google Translate</Trans>
+          </InlineLinkText>
```
mozzius: This doesn't work on android, we need to use the `useGoogleTranslate` hook (pushed a fix up)

mozzius:
Suggestion: https://github.com/bluesky-social/social-app/pull/9991

src/components/Post/Translated/index.tsx
```diff
@@ -1,113 +1,358 @@
...
+            hoverStyle={{opacity: 0.5}}
+            onPress={clearTranslation}>
+            <XIcon size="sm" fill={t.atoms.text_contrast_medium.color} />
+          </Button>
+        </View>
+      </View>
+      <View style={[a.flex_row, a.align_center]}>
+        <Link
+          to={getTranslatorLink(postText, primaryLanguage)}
+          onPress={handleFallback}
```
estrattonbailey: Nice! Tbh, probably not needed, but dig this double-fallback. We also have a `createStaticClick` util exported from `#/components/Link` for cases like this where you just want to do an `onPress`

ds-boyce:
Last commit incorporates #9991.

src/env/index.ts
```diff
@@ -49,3 +49,5 @@ export const IS_WEB_FIREFOX: boolean = false
 export const IS_HIGH_DPI: boolean = true
 // ideally we'd use isLiquidGlassAvailable() from expo-glass-effect but checking iOS version is good enough for now
 export const IS_LIQUID_GLASS: boolean = iOSMajorVersion >= 26
+// So we can avoid attempting on-device translation when we know it's unsupported.
+export const HAS_ON_DEVICE_TRANSLATION: boolean = iOSMajorVersion >= 18
```
estrattonbailey: Should add this to the web file too just for parity, set to `false`

src/lib/translation/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+  // Special cases for regional languages since iOS differentiates and missing
+  // language packs must be downloaded and installed.
+  if (Platform.OS === 'ios') {
+    const deviceLocales = getLocales()
+    const primaryLanguageTag = deviceLocales[0]?.languageTag
+    switch (targetLangCodeOriginal) {
+      case 'en': // en-US, en-GB
+      case 'es': // es-419, es-ES
+      case 'pt': // pt-BR, pt-PT
+      case 'zh': // zh-Hans-CN, zh-Hant-HK, zh-Hant-TW
```
estrattonbailey, line 55: `targetLangCodeOriginal` could have the regional suffix right?

src/lib/translation/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+    },
+    [context, forceGoogleTranslate, key],
+  )
+
+  const clearTranslation = useCallback(
+    () => context.clearTranslation(key),
+    [context, key],
+  )
+
+  return {
```
estrattonbailey: Do we need to memoize this or is the compiler gonna help us here?

src/lib/translation/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+            ...prev,
+            [key]: {status: 'error', message: errorMessage},
+          }))
+        }
+      }
+
+      // Translate after the next state change.
+      LayoutAnimation.configureNext(
+        LayoutAnimation.Presets.easeInEaseOut,
+        () => void translateAfterAnimation(),
```
estrattonbailey: Is there any chance of this not firing?

estrattonbailey: Actually, do we need this delay now that we handle errors in situ? Previously it was showing the pending state for a split second, then hiding. Now, it should immediately expand to show the error state, right?

ds-boyce: It won’t fire if the component is unmounted, but that should be ok, right?

src/lib/translation/index.tsx
ds-boyce: If it does, it will just pass through as-is. If it doesn’t, we want to make sure to catch it here.

And since the `langPrefs.primaryLanguage` uses the `code2` value from `LANGUAGES`, the latter should always be the case. This addresses the issue we saw in testing where if your iOS language was `en-GB` you’d be prompted to download the `en-US` language pack (since apparently iOS interprets `en` as `en-US`).

The Android check above is just future-proofing in case we start appending region/script codes to the `code2` values in `LANGUAGES`.

src/lib/translation/index.tsx
ds-boyce: Compiler should be able to catch this but a little manual memoization wouldn’t hurt.

src/lib/translation/index.tsx
ds-boyce: I’ll check regarding the error state.

src/view/com/posts/PostFeedItem.tsx
```diff
@@ -466,21 +475,33 @@ let PostContent = ({
...
-            numberOfLines={limitLines ? MAX_POST_LINES : undefined}
-            style={[a.flex_1, a.text_md]}
-            authorHandle={postAuthor.handle}
-            shouldProxyLinks={true}
-          />
-          {limitLines && (
-            <ShowMoreTextButton style={[a.text_md]} onPress={onPressShowMore} />
-          )}
-        </View>
+        <>
```
estrattonbailey: Nit: don't need this fragment anymore? Also, could reuse `richText.text` maybe? Nbd on the latter

src/lib/translation/index.tsx
ds-boyce: Ok, removed it from the callback.

ds-boyce: Hm, with it removed there’s no more empty up-and-down animation on error, but the “Loading…” spinner can appear and disappear very quickly in some cases.

Putting it back in the callback introduces a bit of an artificial delay, but it looks smoother in my opinion. What do you think?

estrattonbailey: Cool sounds good, thanks for investigating!

PR #9988, opened 2026-03-04
Remove getSuggestedFollowsByActor fallbacks and use recIdStr

ds-boyce:
Uses the `getSuggestedFollowsByActor` endpoint everywhere and trusts the results vs falling back to `getSuggestions` and munging data.
suggestions

Also switches to a string `recId` from numeric.
id

Fixes an issue where the “Suggested for you” section would remain visible after dismissing all profiles and another where the two sections weren’t in sync.

https://github.com/user-attachments/assets/8a7b0fac-8554-467c-8c35-0e4861ca9003

Finally, addresses a bunch of lint warnings.

### DIFF
--- a/src/components/FeedInterstitials.tsx
+++ b/src/components/FeedInterstitials.tsx
@@ -232,11 +232,13 @@ export function SuggestedFollowsProfile({did}: {did: string}) {
     (dismissedDid: string) => {
       queryClient.setQueryData(
         suggestedFollowsByActorQueryKey(did),
-        (old: typeof data) => {
-          if (!old) return old
+        (previous: typeof data) => {
+          if (!previous) return previous
           return {
-            ...old,
-            suggestions: old.suggestions.filter(s => s.did !== dismissedDid),
+            ...previous,
+            suggestions: previous.suggestions.filter(
+              s => s.did !== dismissedDid,
+            ),
           }
         },
       )
@@ -567,10 +569,10 @@ export function ProfileGrid({
   const profileCountForMinCheck = totalProfileCount ?? profiles.length
 
   useEffect(() => {
-    if (error || (!isLoading && profiles.length < 1)) {
+    if (error || (!isLoading && profileCountForMinCheck < minLength)) {
       onRequestHide?.()
     }
-  }, [error, isLoading, onRequestHide, profiles.length])
+  }, [error, isLoading, onRequestHide, profileCountForMinCheck, minLength])
 
   if (error || (!isLoading && profileCountForMinCheck < minLength)) {
     ax.logger.debug(`Not enough profiles to show suggested follows`)
--- a/src/screens/Profile/Header/SuggestedFollows.tsx
+++ b/src/screens/Profile/Header/SuggestedFollows.tsx
@@ -22,8 +22,6 @@ export function ProfileHeaderSuggestedFollows({
   const {profiles, onDismiss, isLoading, error} =
     useProfileHeaderSuggestions(actorDid)
 
-  if (!profiles.length && !isLoading) return null
-
   /* NOTE (caidanw):
    * Android does not work well with this feature yet.
    * This issue stems from Android not allowing dragging on clickable elements in the profile header.
@@ -57,11 +55,13 @@ function useProfileHeaderSuggestions(actorDid: string) {
     (dismissedDid: string) => {
       queryClient.setQueryData(
         suggestedFollowsByActorQueryKey(actorDid),
-        (old: typeof data) => {
-          if (!old) return old
+        (previous: typeof data) => {
+          if (!previous) return previous
           return {
-            ...old,
-            suggestions: old.suggestions.filter(s => s.did !== dismissedDid),
+            ...previous,
+            suggestions: previous.suggestions.filter(
+              s => s.did !== dismissedDid,
+            ),
           }
         },
       )


src/components/FeedInterstitials.tsx
```diff
@@ -157,7 +154,7 @@ function sortSeenPosts(postA: SeenPost, postB: SeenPost): 0 | 1 | -1 {
   }
 }
 
-function useExperimentalSuggestedUsersQuery() {
+function useExperimentalSuggestedUsersQuery(randomizer = Math.random) {
```
estrattonbailey: 🤔 do we need this?

src/screens/Profile/Header/ProfileHeaderStandard.tsx
```diff
@@ -254,7 +257,8 @@ export function HeaderStandardButtons({
             )}`,
           ),
         )
-      } catch (e: any) {
+      } catch (err) {
+        const e = err as Error
```
estrattonbailey, line 269: Did linting do these? Np really, but kinda noisy

src/view/com/posts/PostFeed.tsx
```diff
@@ -204,6 +203,7 @@ let PostFeed = ({
   savedFeedConfig,
   initialNumToRender: initialNumToRenderOverride,
   isVideoFeed = false,
+  lastFetchDate = Date.now,
```
estrattonbailey: Hmm, same question. What's the benefit here?

src/components/FeedInterstitials.tsx
```diff
@@ -216,81 +213,42 @@ export function SuggestedFollows({feed}: {feed: FeedDescriptor}) {
...
-      ) ?? []
 
     // Dedupe by did, preferring actor-specific profiles
     const seen = new Set<string>()
-    const combined: {actor: bsky.profile.AnyProfileView; recId?: number}[] = []
+    const filtered: {actor: bsky.profile.AnyProfileView; recId?: string}[] = []
 
     for (const profile of actorProfiles) {
       if (!seen.has(profile.did)) {
         seen.add(profile.did)
```
estrattonbailey: Now that we're not munging data, we could get rid of this extra deduping in any cases reading from the new endpoint

src/components/FeedInterstitials.tsx
ds-boyce: Calling `Math.random` in render makes it an impure function, which lint caught.

src/screens/Profile/Header/ProfileHeaderStandard.tsx
ds-boyce: Yep, error is typed as `unknown` so lint flagged it.

src/view/com/posts/PostFeed.tsx
ds-boyce: Ensures that render is a pure function. Flagged by lint.

src/components/FeedInterstitials.tsx
ds-boyce: > Error: Cannot call impure function during render
>
> `Math.random` is an impure function. Calling an impure function can produce unstable results that update unpredictably when the component happens to re-render. (https://react.dev/reference/rules/components-and-hooks-must-be-pure#components-and-hooks-must-be-idempotent).

estrattonbailey: I don't think this changes the result, and this is a valid use case, so let's just ignore this lint

src/screens/Profile/Header/SuggestedFollows.tsx
```diff
@@ -1,22 +1,19 @@
...
 export function ProfileHeaderSuggestedFollows({
   isExpanded,
   actorDid,
+  onAllResultsDismissed,
 }: {
   isExpanded: boolean
   actorDid: string
+  onAllResultsDismissed: () => void
 }) {
   const {allProfiles, filteredProfiles, onDismiss, isLoading, error} =
```
estrattonbailey: We might as well guard against `error` here too instead of relying on the callback, right?

src/view/com/posts/PostFeed.tsx
estrattonbailey: Does this actually change the behavior? If not, let's revert, do the function inline, or make it a global function in the file.

src/components/FeedInterstitials.tsx
```diff
@@ -400,14 +304,16 @@ export function ProfileGrid({
...
 }: {
   isSuggestionsLoading: boolean
-  profiles: {actor: bsky.profile.AnyProfileView; recId?: number}[]
+  profiles: {actor: bsky.profile.AnyProfileView; recId?: string}[]
   totalProfileCount?: number
   error: Error | null
   viewContext: 'profile' | 'profileHeader' | 'feed'
   onDismiss?: (did: string) => void
   isVisible?: boolean
+  onAllResultsDismissed?: () => void
```
estrattonbailey: Optional nit: maybe a better name is like `onShouldHide` since it can technically occur if `error` or no results? Another common pattern I've seen is `onRequestHide`, in the case where state is external and there's an `X` close button or something within a component.

src/screens/Profile/Header/SuggestedFollows.tsx
ds-boyce: Don’t we still need the callback in order to close the accordion? I was trying to centralize the logic in the `ProfileGrid`, but maybe I’m misunderstanding your suggestion.

src/components/FeedInterstitials.tsx
```diff
@@ -216,86 +219,44 @@ export function SuggestedFollows({feed}: {feed: FeedDescriptor}) {
...
-      }
-    }
-
-    return combined
-  }, [data?.suggestions, moreSuggestions?.pages, did, data?.recId])
-
-  const filteredProfiles = useMemo(() => {
-    return allProfiles.filter(p => !dismissedDids.has(p.actor.did))
-  }, [allProfiles, dismissedDids])
+  const queryClient = useQueryClient()
```
estrattonbailey, line 229: Would be kinda nice to create a new hook — basically `useProfileHeaderSuggestions` — called like `useSuggestedFollowsByActorWithDismiss` so we can share this logic here and in the profile header. Could live alongside the `useSuggestedFollowsByActorQuery`.

ds-boyce: Good idea!

ds-boyce: I’m going to do that as a follow-up to this.

ds-boyce: #10048

PR #9993, opened 2026-03-04
Update React Native Screens and React Navigation to latest

mozzius:
I want this bug fix: https://github.com/software-mansion/react-native-screens/pull/3584

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -264,7 +264,7 @@ module.exports = function (_config) {
               ],
             },
             android: {
-              compileSdkVersion: 35,
+              compileSdkVersion: 36,
               targetSdkVersion: 35,
               buildToolsVersion: '35.0.0',
               buildReactNativeFromSource: IS_PRODUCTION,


estrattonbailey:
Looks like the bug fix is behind a feature flag `iosPreventReattachmentOfDismissedScreens`

estrattonbailey:
Have you tested the dep updates on all platforms?

mozzius:
@estrattonbailey looks like there's a later commit that unflags it.

No, I have not tested on Android, good shout - will do that

estrattonbailey:
Sharing here for posterity
CleanShot 2026-03-10 at 14 13 55@2x

mozzius:
Interesting! We might need to target SDK 36

estrattonbailey:
Yeah works locally. Had to brush up on this. Looks like these deps support a minSdK of 24, so we should be ok to bump this.

PR #9997, opened 2026-03-04
feat: add per-user mute reposts from profile menu

cidoliveira:
## Summary
- Adds "Mute reposts" / "Unmute reposts" toggle to the profile three-dot menu
- Muted users' reposts are filtered from the Following and List feeds (original posts still appear)
- Uses local persisted storage (same pattern as `hiddenPosts`)

## Implementation
- **State layer:** `mutedReposts` field in persisted schema + `MutedRepostsProvider` React Context (follows `hidden-posts.tsx` pattern)
- **Feed filtering:** `FeedTuner.removeRepostsFromMuted()` factory method, skipped when global "hide reposts" is ON
- **Profile menu:** Toggle item with `RepostSlash` icon, toast confirmation, inside `!isSelf` guard

## Known Limitations
- Mute state is device-local (localStorage/AsyncStorage), not synced across devices. Server-side sync would require a new AT Protocol lexicon, which is outside the scope of a client-side contribution.

## Test Plan
- [x] Menu item appears on other users' profiles with repost-slash icon
- [x] Menu item hidden on own profile
- [x] Tapping toggles label between "Mute reposts" / "Unmute reposts"
- [x] Toast confirms action
- [x] Muted user's reposts filtered from Following feed
- [x] Original posts from muted user still appear
- [x] Unmuting restores reposts
- [x] State persists across page reload
- [x] TypeScript typecheck passes (no new errors)
- [x] ESLint passes (no new warnings)


cidoliveira:
Hey! 👋

I built this as a client-side implementation of per-user repost muting, which is a frequently requested feature. I wanted to get feedback before going further.

The current approach uses local persisted storage (same pattern as `hiddenPosts`), which means it works per-device but doesn't sync across clients. I'm aware this might not be the direction the team prefers — if server-side sync via a new AT Protocol lexicon is the desired approach, I'm happy to adapt this PR or close it in favor of a server-driven solution.

A few notes:
- The translation strings (`Mute reposts` / `Unmute reposts`) will need to go through the nightly `intl:extract` / `intl:compile` CI job
- The `FeedTuner.removeRepostsFromMuted` filter is skipped when the global "hide reposts" preference is already ON, so there's no redundant filtering

Would love to hear if this is something you'd consider merging as a local-storage MVP, or if you'd prefer to wait for a protocol-level solution. Either way, happy to help move it forward.

cidoliveira:
giving this a bump to see if it gets some attention, i've been using this for the last 2 weeks to mute repost-heavy mutuals and it's been great. since it doesn't have lexicon support i'll probably just make it a chrome extension

mozzius:
Superceded by #11223

PR #10003, opened 2026-03-05
add alert variant to dialog

vineyardbovines:
CleanShot 2026-03-05 at 12 18 14

since the alert and sheet are both modals with the same control APIs, i've extended the dialog system to support a `type` component for different presentation styles (`sheet` and `alert`). internally, alerts just have a basic view for inner content and the sheet variant remains unchanged (with handle, scrollable, etc.) 

definitely open to making an entirely separate component here if we want, but wanted to first propose extending dialogs in this way just from the functional perspective.


mozzius:
Rather than wrapping Dialog, what about wrapping Prompt instead? Dialogs is a bit catch-all but Prompt is generally meant to be more intentional/locked down. Seems like the API would match up nicely if we added an icon component to it?

vineyardbovines:
@mozzius yeah i think thats a lot better actually. just swapped it over

estrattonbailey:
Can you chuck an example into the `Dialogs` `Storybook` file?

PR #10008, opened 2026-03-05
[APP-1928] add bot/automated account badge and self-labeling settings

vineyardbovines:
**Summary**

- Adds a robot icon badge next to account names that have self-labeled as automated/bot accounts
- New settings screen (AutomationLabelSettings) where users can toggle the bot label on their own account
- Bot badge is tappable, shows an info sheet explaining the account is automated, with a link to settings if viewing your own profile
- Badge appears across the app: profiles, post meta, notifications, DMs, search, composer, hover cards, and embed

**Test plan**

- Toggle the bot label on/off in Settings > Account > Automation label
- Verify the robot badge appears next to your display name across surfaces (feed, profile, notifications, DMs, search)
- Tap the badge on your own profile — should show sheet with "Open settings" button
- Tap the badge on another bot account — should show sheet without settings button
- Test on iOS, Android, and web

**Not included here**
This implementation is not a bi-directional bot <-> owner link. It's simply a self label, which will allow for filtering and hiding content, like any other label. This is an initial pass at this, more work to come.

### DIFF
--- a/src/components/BotBadge.tsx
+++ b/src/components/BotBadge.tsx
@@ -1,7 +1,6 @@
 import {View} from 'react-native'
 import {type ComAtprotoLabelDefs} from '@atproto/api'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
+import {useLingui} from '@lingui/react/macro'
 
 import {atoms as a, useTheme} from '#/alf'
 import {BotAccountAlert} from '#/components/BotAccountAlert'
@@ -51,7 +50,7 @@ export function BotBadgeButton({
 }) {
   const t = useTheme()
   const ax = useAnalytics()
-  const {_} = useLingui()
+  const {t: l} = useLingui()
   const control = useDialogControl()
 
   if (!isBotAccount(profile)) {
@@ -61,7 +60,7 @@ export function BotBadgeButton({
   return (
     <>
       <Button
-        label={_(msg`Automated account`)}
+        label={l`Automated account`}
         hitSlop={20}
         onPress={evt => {
           evt.preventDefault()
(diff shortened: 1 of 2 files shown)


launchpad001:
I personally feel that this implementation is incomplete.

I think maybe we should implement a new lexicon, such as `com.atproto.agent.declaration`, and include `ownerDid` so that we can quickly find the owner of a bot?

rektide:
To be clear, this is both new UI and new lexicon field?

Can you help causals readers (like me) understand this better? Did `com.atproto.label.defs#selfLabels` exist & you just added `bot`?  Did the NSID exist and you add a field? Is it all net new?

I tend to agree with @rdp-studio  that a more complete handling of this issue would also include an owner did field.

surfdude29:
> Did `com.atproto.label.defs#selfLabels` exist & you just added `bot`?

(I'm not a bsky team member just a volunteer contributor)

Yes, `com.atproto.label.defs#selfLabels` is already in the lexicon:

https://github.com/bluesky-social/atproto/blob/main/lexicons/com/atproto/label/defs.json#L54

src/screens/Settings/Settings.tsx
```diff
@@ -374,6 +375,14 @@ function ProfilePreview({
             <VerificationCheckButton profile={shadow} size="lg" />
           </View>
         )}
+        <View
+          style={[
+            {
+              marginTop: platform({web: 8, ios: 8, android: 10}),
+            },
+          ]}>
```
estrattonbailey: Note to self: check for these cases and avoid extra space showing at all times

estrattonbailey:
@rdp-studio @rektide we've considered options around bi-directional link records (to ensure that a bot truly is owned by another account), but this is not that. It's just a label, like any other label, and will enable content filtering/hiding like any other label. Basically our first pass at this type of thing.

A bi-directional system could power many other things, and we need to think more about it. But good shout, we agree, more is needed here!

estrattonbailey (approved):

PR #10013, opened 2026-03-05
Disable animations for translation events on Android

ds-boyce:
These were causing layout issues.

| Before | After |
| - | - |
| before | after |


src/lib/translation/index.tsx
```diff
@@ -192,7 +192,9 @@ export function Provider({children}: React.PropsWithChildren<unknown>) {
   }, [])
 
   const clearTranslation = useCallback((key: string) => {
-    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut)
+    if (Platform.OS !== 'android') {
```
mozzius: nit: use `IS_ANDROID`

PR #10014, opened 2026-03-05
Log languages a post is tagged with after translating

ds-boyce:
**Android**

```json
"event": "translate:result",
"payload": {
  "method": "on-device",
  "os": "android",
  "sourceSelection": "automatic",
  "sourceLanguage": "ja",
  "targetLanguage": "en",
  "postLanguages": [
    "ja"
  ]
}
```

**iOS**
```json
"event": "translate:result",
"payload": {
  "method": "fallback-alert",
  "os": "ios",
  "sourceSelection": "automatic",
  "sourceLanguage": "fi",
  "targetLanguage": "en",
  "postLanguages": [
    "en"
  ]
}
```

**Web (success)**

```json
"event": "translate:result",
"payload": {
    "method": "google-translate",
    "os": "web",
    "sourceSelection": "automatic",
    "sourceLanguage": "ja",
    "targetLanguage": "en",
    "postLanguages": [
        "ja"
    ]
}
```

### DIFF
--- a/src/lib/translation/types.ts
+++ b/src/lib/translation/types.ts
@@ -35,7 +35,7 @@ export type TranslationFunctionParams = {
 
 export type TranslationOptions = {
   key: string
-  forceGoogleTranslate: boolean
+  forceGoogleTranslate?: boolean
   /**
    * The language(s) of the post being translated. Used for analytics purposes
    * to understand translation usage patterns better. Optional because it may


src/lib/translation/index.tsx
```diff
@@ -242,33 +251,41 @@ export function Provider({children}: React.PropsWithChildren<unknown>) {
...
             status: 'success',
             translatedText: result.translatedText,
             sourceLanguage: result.sourceLanguage,
             targetLanguage: result.targetLanguage,
+            postLanguages: postLangCodes,
           },
         }))
       } catch (e) {
-        logger.error('Failed to translate post on device', {safeMessage: e})
+        const sourceLanguage = sourceLangCode ?? guessLanguage(text)
```
estrattonbailey: Since we fall back to `guessLanguage` here and in `attemptTranslation`, we've got an additional variable at play that will affect the results of `isPostLanguageAccurate`.  Or if the user changes language to translate `en` to `en`, we'll get `true` even if the post wasn't tagged correctly.

Now that I'm looking at this I'm wondering if this data is good enough to both tracking 🤔 what do you think?

ds-boyce: I added `isPostLanguageAccurate` to make viewing and filtering the metrics easier, but you make a good point.

Maybe we need to also track when a `translate` call is made via the “Translate” link vs the “Change” link, so that we only include this additional data for the former case.

Something like `sourceSelection: 'automatic' | 'manual'`.

ds-boyce: I think even if we drop `isPostLanguageAccurate` the source selection would be useful to track.

ds-boyce: Ok, dropped `isPostLanguageAccurate` and added `sourceSelection`. So if someone changes the source language to `en` on a posted tagged with `en`, we’ll be able to differentiate that from auto-detection.

PR #10015, opened 2026-03-06
Refine presentation of translation results

ds-boyce:
* Translated text size matches post text size.
* Top line aligns with the X button and wraps if space runs out.
* X button placement is consistent between success and error states.
* Margins and padding have been tweaked for greater consistency.
* Reduced font weight for non-interactive text.

| Success | Error |
| - | - |
| translated | error |

### DIFF
--- a/src/components/Post/Translated/index.tsx
+++ b/src/components/Post/Translated/index.tsx
@@ -96,7 +96,7 @@ function TranslationLoading() {
         <Text style={[a.text_sm, t.atoms.text_contrast_medium]}>
           <Trans>Translating</Trans>
         </Text>
-        <Loader size="xs" />
+        <Loader size="xs" fill={t.atoms.text_contrast_medium.color} />
       </View>
     </View>
   )


src/components/Post/Translated/index.tsx
```diff
@@ -175,31 +182,38 @@ function TranslationError({
     <View
       style={[
         a.px_lg,
-        a.pt_sm,
-        a.pb_md,
+        a.py_lg,
         a.mt_sm,
         a.border,
         a.rounded_lg,
+        a.flex_nowrap,
```
estrattonbailey: A better solve for this (to ensure we don't overlap UI) is pry absolutely position the button in the top right, and add paddingRight to the content so that the content can wrap normally and never interfere with the button.

src/components/Post/Translated/index.tsx
```diff
@@ -175,31 +182,38 @@ function TranslationError({
...
+          hitSlop={HITSLOP_30}
+          hoverStyle={{opacity: 0.5}}
+          style={[a.self_end]}
+          onPress={clearTranslation}>
+          <XIcon size="sm" fill={t.atoms.text_contrast_medium.color} />
+        </Button>
       </View>
-      <View style={[a.flex_row, a.align_center]}>
+      <View style={[a.flex_row, a.align_center, a.mt_sm]}>
         <Link
```
estrattonbailey, line 219: Ah didn't think of this before, but we have `InlineLinkText` that applies link styles to text, could swap that in here instead and drop the inner `Text` element entirely

src/components/Post/Translated/index.tsx
```diff
@@ -241,89 +264,101 @@ function TranslationResult({
     ? codeToLanguageName(sourceLanguage, i18n.locale)
     : undefined
 
+  const flattenedStyle = flatten(postTextStyle) ?? {}
+  const fontSize = isOnlyEmoji(postText)
```
estrattonbailey: Good thinking here! Not sure we need to support translation of emoji-only posts though, could pry drop the `emojiMultiplier` stuff 🤔

src/components/Post/Translated/index.tsx
ds-boyce: First attempt did this for the error state; I should listen to my gut. Will update!

src/components/Post/Translated/index.tsx
ds-boyce: Nice, ty!

src/components/Post/Translated/index.tsx
ds-boyce: Yeah I only included it since the “Translate” option is always in the “…” menu, regardless of post content. Just wanted to ensure the presentation was consistent.

estrattonbailey: True, yeah that's fair. What happens there on iOS? Same language translations always error out right?

ds-boyce: Ah, right we currently throw an error if the translated text matches the source text. I can yank this out to simplify things here.

src/components/Post/Translated/index.tsx
ds-boyce: `InlineLinkText` doesn’t appear to have a `hitSlop` property since it’s a `Text` element, whereas `Link` is a `Button`. I think I’m going to leave this as-is for the accessibility boost.

estrattonbailey: Right right, good call!

PR #10025, opened 2026-03-07
Add “download image” option to web, simplify downloads

mozzius:
Two parts to this
1. Adds a download button to the web lightbox to have parity with native. Currently a menu with a single option in it, presumably we can add more options to this down the line
2. Rather than re-encoding the images we get from the CDN as JPEGs, we just ask the CDN for JPEGs directly. Should make downloads faster

Test plan

- download an image (web + native)


estrattonbailey (approved):
LGTM too!

PR #10029, opened 2026-03-09
Fix the Go Live dialog getting stuck behind elements on Android

RetroSunstar:
Currently the Go Live dialog gets stuck behind the keyboard on android devices

When it has an embed, the menu buttons also become inaccessable (so you can never actually start a live event)

This PR fixes that by simply removing the preventExpansion limiter. Additionally it adds preventDismiss as this would now be a problem 🐙


mozzius:
Hey! This'll work - will merge this if we don't get #9973 in before the next release

mozzius:
We merged it :) thanks for this PR though, good fix!

PR #10033, opened 2026-03-09
Update Lingui syntax to v5 in `.ts` files

ds-boyce:
Doing this manually ahead of the `.tsx` codemod in #10032.

### DIFF
--- a/src/view/com/composer/state/video.ts
+++ b/src/view/com/composer/state/video.ts
@@ -413,15 +413,15 @@ function getUploadErrorMessage(
       case 'User is not allowed to upload videos':
         return l`You are not allowed to upload videos.`
       case 'Uploading is disabled at the moment':
-        return `Hold up! We’re gradually giving access to video, and you’re still waiting in line. Check back soon!`
+        return l`Hold up! We’re gradually giving access to video, and you’re still waiting in line. Check back soon!`
       case 'Failed to get user’s upload stats':
         return l`We were unable to determine if you are allowed to upload videos. Please try again.`
       case 'User has exceeded daily upload bytes limit':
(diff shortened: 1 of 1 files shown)


mozzius:
I don't think this does the same thing. If you import the global `i18n` instance, it won't update when the language changes.

Why do we need to remove these ones?

ds-boyce:
Ah, that’s good to know. I’ll just ignore these when running the codemod then.

ds-boyce:
Ok, updated these to pass the `t` from `useLingui` to the helper functions, as it was.

src/view/com/composer/state/video.ts
```diff
@@ -387,66 +386,56 @@ export async function processVideo(
...
       case 'User has exceeded daily upload bytes limit':
-        return _(
-          msg`You've reached your daily limit for video uploads (too many bytes)`,
-        )
+        return `You’ve reached your daily limit for video uploads (too many bytes)`
       case 'User has exceeded daily upload videos limit':
-        return _(
-          msg`You've reached your daily limit for video uploads (too many videos)`,
-        )
+        return `You’ve reached your daily limit for video uploads (too many videos)`
```
estrattonbailey: Couple missing in there :P

src/view/com/composer/state/video.ts
```diff
@@ -387,66 +386,56 @@ export async function processVideo(
...
       case 'Account is not old enough to upload videos':
-        return _(
-          msg`Your account is not yet old enough to upload videos. Please try again later.`,
-        )
+        return `Your account is not yet old enough to upload videos. Please try again later.`
```
estrattonbailey: Couple missing in there :P

ds-boyce:
Can’t pass the `t` macro around.

PR #10036, opened 2026-03-10
Update issue templates to take uploads, and fix typo

heikadog:
Updated the issue templates to take the new `upload` type, and fixed a typo in the bug report template:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#upload

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,5 +1,5 @@
 name: "Bug Report"
-description: "Create a report for an issue you have experience in the app."
+description: "Create a report for an issue you have experienced in the app."
 labels: ["bug"]
 body:
   - type: markdown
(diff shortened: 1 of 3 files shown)


estrattonbailey:
Is this needed? The text area has an upload button beneath it, and has the added bonus of being able to add comments adjacent to any uploads that are added to the text area.

heikadog:
> [...]The text area has an upload button beneath it

It's clearer that you can upload something this way as a button rather than a smaller link below the text area, and has more explicit file type restrictions *just* to images and videos rather than accepting additional filetypes, which the main issue templates don't do.

> [..]and has the added bonus of being able to add comments adjacent to any uploads that are added to the text area.

The benefit of having adjacent comments is already accomplished by the Additional Information box—users can use that to add comments to uploads if needed. Plus, looking at recent issues with media, it doesn't seem like people use the Attachments field to add comments when they do add attachments anyway, so that benefit seems more theoretical in practice.

PR #10037, opened 2026-03-10
First pass at fixing large type by adding `maxFontSizeMultiplier` 

mozzius:
Really really rough first pass at fixing the most broken areas when using super large type. Notably, adding the draft button pushed the Post button offscreen, which broke the composer for people using large type.

The fix is to sprinkle in `maxFontSizeMultiplier` where appropriate

<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>


estrattonbailey (approved):
LGTM

PR #10040, opened 2026-03-11
Run react imports codemod

estrattonbailey:
[It's old code](https://github.com/reactjs/react-codemod?tab=readme-ov-file#update-react-imports) but it checks out. This codemod uses an old version of `codemod`, but since it's just imports, seems to work just fine.

Ran this on all platforms and ran the Jest tests locally. E2E boots up, but I didn't actually run through them all. Poking around the app should do the trick for imports.

### DIFF
--- a/bskyogcard/src/components/Butterfly.tsx
+++ b/bskyogcard/src/components/Butterfly.tsx
@@ -1,6 +1,6 @@
-import React from 'react'
+import {type SVGAttributes} from 'react'
 
-export function Butterfly(props: React.SVGAttributes<SVGSVGElement>) {
+export function Butterfly(props: SVGAttributes<SVGSVGElement>) {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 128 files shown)


bskyogcard/src/components/Butterfly.tsx
```diff
@@ -1,6 +1,6 @@
-import React from 'react'
+import {type SVGAttributes} from 'react'
 
-export function Butterfly(props: React.SVGAttributes<SVGSVGElement>) {
```
ds-boyce, line 3: nit: we don’t need to import `type` from `React` anywhere, since the `React` type is globally available. We only need to address runtime cases, e.g., `useState`.

ds-boyce: The codemod in #10032 considers that case. 🙂

estrattonbailey:
@ds-boyce as discussed, let's just trim down to the single codemod and do the modification in your other PR 👍

PR #10041, opened 2026-03-11
Provide age-restricted viewers the option to delete their account

ds-boyce:
See 7306e61344c88710bdadf9293440eb820a9cd32c for the relevant changes.

| Android | iOS |
| - | - |
| android | ios |

### DIFF
--- a/src/view/shell/index.tsx
+++ b/src/view/shell/index.tsx
@@ -236,7 +236,7 @@ export function Shell() {
         <Deactivated />
       ) : (
         <>
-          {aa.state.access !== aa.Access.None ? (
+          {aa.state.access === aa.Access.None ? (
             <NoAccessScreen />
           ) : (
             <RoutesContainer>


src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -216,7 +239,7 @@ export function NoAccessScreen() {
                 </Text>
                 <Text style={[textStyles]}>
                   <Trans>
-                    Set your birthdate below and we'll get you back to posting
+                    Set your birthdate below and we’ll get you back to posting
```
estrattonbailey: Let's not churn translations with these edits on this one. An edit like this would mean we need to re-translate the string. Or maybe the system is smart enough to ignore this. If that's the case, nbd, but otherwise, let's drop the apostrophe edits here. AA is just such a touchy one.

ds-boyce: Noted, I’ll revert this.

src/ageAssurance/components/NoAccessScreen.tsx
```diff
@@ -112,6 +117,24 @@ export function NoAccessScreen() {
           .
         </Trans>
       </Text>
+      <Text style={[textStyles]}>
```
estrattonbailey: Let's move this into its own variable so we can put it in both spots it's in already, and in the `!hasDeclaredAge` block as well. So 3 total spots. 

Currently the `!isBlocked &&` is incorrect for this too, we want this visible for every possible state.

estrattonbailey (changes requested):
Small tweak to make this more agnostic of other logic.

Alternative: what do you think about putting it down here? This sorta thing, but with a link to open the dialog.

Image

```diff
diff --git i/src/ageAssurance/components/NoAccessScreen.tsx w/src/ageAssurance/components/NoAccessScreen.tsx
index 8d307bb47..3de6dc4b8 100644
--- i/src/ageAssurance/components/NoAccessScreen.tsx
+++ w/src/ageAssurance/components/NoAccessScreen.tsx
@@ -257,9 +257,9 @@ export function NoAccessScreen() {
               </View>
             )}
 
-            <View style={[a.pt_lg, a.gap_xl]}>
+            <View style={[a.pt_lg, a.gap_xl, {maxWidth: 280}]}>
               <Logo width={120} textFill={t.atoms.text.color} />
-              <Text style={[a.text_sm, a.italic, t.atoms.text_contrast_medium]}>
+              <Text style={[a.text_sm, a.italic, a.leading_snug, t.atoms.text_contrast_medium]}>
                 <Trans>
                   To log out,{' '}
                   <SimpleInlineLinkText
@@ -269,7 +269,7 @@ export function NoAccessScreen() {
                     })}>
                     click here
                   </SimpleInlineLinkText>
-                  .
+                  . Of if you'd prefer, you can delete your account.
                 </Trans>
               </Text>
             </View>
```

estrattonbailey (approved):

PR #10046, opened 2026-03-12
Address lint warnings in List.web.tsx

ds-boyce:
Omitted this in #10032 in favor of doing it manually, since it revealed a small issue with refs.

Before, the `React.useRef` use wasn’t typed:
before

Switching to `import {useRef} from 'react'` revealed the type issue:
after

To address, I put a guard around the `observe` calls. Tested on web and everything appears to be working as intended.

### DIFF
--- a/src/view/com/util/List.web.tsx
+++ b/src/view/com/util/List.web.tsx
@@ -1,9 +1,13 @@
-import React, {
+import {
+  forwardRef,
   isValidElement,
-  type JSX,
   memo,
   startTransition,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
   useRef,
+  useState,
 } from 'react'
 import {
   type FlatListProps,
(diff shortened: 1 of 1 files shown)


src/view/com/util/List.web.tsx
```diff
@@ -113,9 +117,9 @@ function ListImpl<ItemT>(
     })
   }
 
-  const getScrollableNode = React.useCallback(() => {
+  const getScrollableNode = useCallback(() => {
     if (disableFullWindowScroll) {
-      const element = nativeRef.current as HTMLDivElement | null
+      const element = nativeRef.current
```
ds-boyce, line 122: > This assertion is unnecessary since it does not change the type of the expression.

src/view/com/util/List.web.tsx
```diff
@@ -468,12 +472,12 @@ let Row = function RowImpl<ItemT>({
             if (!intersectionTimeout.current) {
               intersectionTimeout.current = setTimeout(() => {
                 intersectionTimeout.current = undefined
-                onItemSeen!(item)
+                onItemSeen(item)
```
ds-boyce, line 475: Same here as above. ESLint fixed these automatically.

src/view/com/util/List.web.tsx
```diff
@@ -482,18 +486,22 @@ let Row = function RowImpl<ItemT>({
...
     const observer = new IntersectionObserver(
       handleIntersection,
       ON_ITEM_SEEN_INTERSECTION_OPTS,
     )
-    const row: Element | null = rowRef.current!
-    observer.observe(row)
+    const row: Element | null = rowRef.current
+    if (row) {
+      observer.observe(row)
+    }
```
ds-boyce, line 500: The null check.

PR #10047, opened 2026-03-12
Fix video/audio desync on Firefox in web video player

mozzius:
Two issues were causing audio to drift out of sync with video on Firefox:

1. fastSeek() seeks video to the nearest keyframe but keeps audio at the
   precise timestamp, causing progressive drift. Disable fastSeek on Firefox
   and use currentTime for precise seeking instead.

2. Low-quality segment flushing only flushed video buffers, leaving stale
   audio data that could desync from newly-loaded video. Flush both audio
   and video buffers together when upgrading quality.

https://claude.ai/code/session_01LGFSY2GgQJ3iw5Cp4XwiuL


estrattonbailey:
How do we test this?

mozzius:
Tbh not sure, could never repro consistently. Leave it for now, I'll take another look

PR #10053, opened 2026-03-13
Require age assurance to access chat settings

ds-boyce:
See sister PR for additional related handling: https://github.com/bluesky-social/social-app/pull/10216

Removes the link to **Chat settings** in the age assurance header if chat is disabled for that account.

chat

Also makes `/messages/settings` an age-restricted screen in case someone hits the URL directly on web.

settings


Rexogamer:
am I missing something here? do these settings not work without age assurance? because otherwise it seems like a negative to prevent people from blocking messages they can't read

estrattonbailey (approved):

PR #10055, opened 2026-03-15
feat(web): drag media onto feed to start a new post

ezramechaber:
## Summary

Enables dragging an image or video onto the feed to start a new post on web.

- **Unified drag-and-drop handling**: Moved the body-level drag/drop listeners out of the composer's `TextInput.web.tsx` into a shared `useImageDropZone` hook. This eliminates duplicate listeners and creates a single system that works whether the composer is open or not.
  - **Composer open**: dropped media routes through the existing `textInputWebEmitter` → `onPhotoPasted` pipeline (existing behavior, unchanged)
  - **Composer closed**: dropped media opens the composer with images/video attached
- **Video support**: The composer-closed drop path handles both images and videos. Uses object URLs for efficient video dimension extraction.
- **Bug fix**: Wired the missing `videoUri` prop through `Composer.web.tsx` — it was defined in `ComposerOpts` but never passed to `ComposePost`
- **Visual**: When dragging over the feed, the "What's up?" composer prompt transforms into a dashed drop target

Addresses feedback from #9840 to reuse the existing composer drag-and-drop system rather than building a parallel one.

## Test plan

- [ ] Drag an image onto the feed → composer opens with image attached
- [ ] Drag a video onto the feed → composer opens with video attached  
- [ ] Drag an image while composer is already open → image added to open composer
- [ ] Drag a video while composer is already open → video added to open composer
- [ ] Drop zone overlay appears on both the feed prompt and in the composer
- [ ] Non-media files are ignored
- [ ] Paste images/videos into the composer still works (no regression)

Supersedes #9840.


ezramechaber:
@mozzius this covers your feedback from https://github.com/bluesky-social/social-app/pull/9840 (pulling the existing method out and re-using it for composer and feed). Tested on desktop web with Chromium. 

I don't have a good way to bundle and test react native on my phone - if there's a good workflow for this to test effectively, I'm happy to try!

PR #10059, opened 2026-03-16
Refactor chat list implementation

mozzius:
By using the new `KeyboardChatScrollView` from `react-native-keyboard-controller`, we can significantly simplify our code and make it way smoother, including interactive keyboard dismissal on both platforms.

https://github.com/user-attachments/assets/bec975fc-9588-4354-b04f-9daa917b0015

### Test plan

For all platforms:
- Test keyboard opening/closing, confirm position is correct
- Test keyboard dismiss behaviour - should be "interactive" on iOS and Android
- Test opening keyboard when scrolled upwards
- Test chat with very few messages
- Test behaviour with incoming messages (i.e. scroll to bottom or not)
- Test Liquid Glass, confirm fallback works okay
- General stress testing


webpack.config.js
```diff
@@ -19,6 +19,9 @@ const reactNativeWebWebviewConfiguration = {
 }
 
 module.exports = async function (env, argv) {
+  env.babel = {
+    dangerouslyAddModulePathsToTranspile: ['@bsky.app/expo'],
+  }
```
mozzius, line 24: turns out @expo/webpack-config has a hardcoded allowlist for packages that start with "expo", so we need to add our `@bsky.app/expo-*` packages too

ds-boyce:
Looks really slick on iOS. 🙂 

On Android, the send button hides while scrolling, which can result in a flicker:

https://github.com/user-attachments/assets/792620f8-0812-42cc-a4f0-df37678c80cb

…to avoid it, should we persist that button while the keyboard is open?

mozzius:
Oh, that's bizarre, will fix

mozzius:
@ds-boyce okay fixed by deviating from the design a bit - I think we should just keep the button visible when the keyboard is closed, I can't figure out how to make it look good to have it appear during the animation

How about we assume this is good enough for now, and I can follow up in another PR if we decide we really need it to hide when closed

ds-boyce (approved):
Yep, send it.

ds-boyce:
Tested this again on an Android device running 16, an iOS simulator running 15.5, another running 26.4, and on web. Looks good all around.

patches/expo-glass-effect+55.0.8.patch
estrattonbailey, line 1: What's this patch for? Worth adding a md?

mozzius: Oh, Expo 54 support. Lemme add a md file

src/screens/Messages/components/MessageComposer.tsx
```diff
@@ -91,152 +120,111 @@ export function MessageComposer({
...
+            glassEffectStyle="regular"
+            style={[a.flex_1, a.rounded_xl, {minHeight: MIN_HEIGHT}]}
+            tintColor={t.palette.contrast_50}
+            fallbackStyle={[t.atoms.bg_contrast_50]}>
+            {IS_WEB && (
+              <Pressable
+                onPress={e => {
+                  e.currentTarget.measure(
+                    (_fx, _fy, _width, _height, px, py) => {
+                      // TODO: rip this horrible system out
```
estrattonbailey, line 149: Indeed, that library I sent you a while back looks perfect tbh

M21H:
Using Xcode 16.1 (16B40), I encounter this error when building the project. Does this indicate that an upgrade to a newer Xcode version is required?

```(/Users/max/projects/open-source/social-app/node_modules/@bsky.app/expo-scroll-edge-effect/ios/ExpoScrollEdgeEffectView.swift:120:56)

  118 |
  119 |   @available(iOS 26, *)
> 120 |   private func resolveEffectStyle(_ effect: String) -> UIScrollEdgeEffect.Style? {
      |                                                        ^ cannot find type 'UIScrollEdgeEffect' in scope
  121 |     switch effect {
  122 |     case "automatic": return .automatic
  123 |     case "hard": return .hard```

PR #10063, opened 2026-03-18
web: LRU screen cache to bound memory in long sessions

mozzius:
## Summary

- On web, React Navigation's `NativeStackView` keeps every visited screen fully mounted (`display: none`). Over long sessions this leads to unbounded memory growth — every screen's React tree, hooks, and query subscriptions stay alive.
- Adds an LRU cache to the custom navigator: Home is always pinned, the focused screen is always mounted, and the 5 most recently visited screens are kept. Everything else renders a lightweight `<View />` placeholder — the route stays in navigation state so browser back/forward works, but the React tree is torn down.
- Defers scroll restoration to `requestAnimationFrame` so that screens re-mounted after LRU eviction have their content in the DOM before `scrollTo` runs.

## Test plan

- [ ] Navigate through 10+ screens on web, verify only ~7 are mounted (Home + focused + 5 cached) via React DevTools
- [ ] Press browser back to an evicted screen — should re-mount with data from query cache and restore scroll position
- [ ] Verify native is unaffected (LRU logic is gated behind `IS_WEB`)
- [ ] Monitor query cache size over a long session to confirm it stabilizes

### DIFF
--- a/src/lib/hooks/useWebScrollRestoration.ts
+++ b/src/lib/hooks/useWebScrollRestoration.ts
@@ -42,7 +42,11 @@ export function useWebScrollRestoration() {
     () => ({
       focus(e: EventArg<'focus', boolean | undefined, unknown>) {
         const scrollY = state.scrollYs.get(e.target) ?? 0
-        window.scrollTo(0, scrollY)
+        // Deferred so that screens re-mounted by LRU eviction have
+        // time to render their content before we scroll.
+        requestAnimationFrame(() => {
+          window.scrollTo(0, scrollY)
+        })
         state.focusedKey = e.target ?? null
       },
     }),


src/view/shell/createNativeStackNavigatorWithAuth.tsx
```diff
@@ -42,6 +42,10 @@ import {BottomBarWeb} from './bottom-bar/BottomBarWeb'
 import {DesktopLeftNav} from './desktop/LeftNav'
 import {DesktopRightNav} from './desktop/RightNav'
 
+// On web, only this many screens (beyond Home + focused) stay mounted.
+// Older screens are unmounted to prevent memory growth during long sessions.
+const WEB_MAX_CACHED_SCREENS = 5
```
estrattonbailey, line 47: I wonder if we could push this to 10 or so. The larger backlog we allow, the less a chance of someone experiencing a bug relating to unwinding the stack. I haven't encountered anything yet though.

Little bit of fuzz testing: 7 screens, 5 more screens, 6 more screens (just navigating between accounts and threads using a feed as a starting point).

Image

Initial snap was home screen at rest. Maybe we start with 5 (or even less) to increase our chances of identifying issues. Will make a note for next release to revisit this.

mozzius: Yeah I genuinely have no idea at all how to systematically debug memory issues like this. Would appreciate your eyes on it

PR #10067, opened 2026-03-18
Ensure mute state stays in sync after mutation

ds-boyce:
The bell icon in the chat list and header wasn’t updating when muting/unmuting a conversation until after a hard refresh.

This [optimistically updates](https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#via-the-cache) the query cache after the mutation, with rollback in case it fails. It then uses the optimistic value in `ChatContext`, keeping that as the source of truth for the `mute` value via `useConvo()`. Notably, this means no changes were required to `MessagesListHeader.tsx` or `ChatListItem.tsx`.

The icon now updates properly:

https://github.com/user-attachments/assets/52dd4573-6474-4c11-b099-fff6766f9d8c

Tested on Android (device), iOS (simulator), and web.

### DIFF
--- a/src/state/messages/convo/index.tsx
+++ b/src/state/messages/convo/index.tsx
@@ -119,5 +119,20 @@ export function ConvoProvider({
     })
   }, [convo, queryClient])
 
+  useEffect(() => {
+    const [root, id] = getConvoKey(convoId)
+    return queryClient.getQueryCache().subscribe(event => {
+      const queryKey = event.query.queryKey as string[]
+      if (queryKey[0] === root && queryKey[1] === id) {
+        const data = event.query.state.data as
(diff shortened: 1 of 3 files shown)


src/state/messages/convo/index.tsx
```diff
@@ -119,5 +119,20 @@ export function ConvoProvider({
...
+        const data = event.query.state.data as
+          | ChatBskyConvoDefs.ConvoView
+          | undefined
+        if (data && convo.convo && data.muted !== convo.convo.muted) {
+          convo.updateMuted(data.muted)
+        }
+      }
+    })
+  }, [convo, convoId, queryClient])
+
```
mozzius, line 136: Huh, interesting pattern, didn't know you could do that!

PR #10072, opened 2026-03-19
Support links cleanup

mozzius:
- Moved all hardcoded `bsky.social/about` urls to `#/lib/constants`
- Cleaned up the support redirect screens to a simple reusable component


estrattonbailey:
Will come back to this soon, I need to triage all the Zendesk links and such too to understand what FAQs and help blogs we link to

PR #10074, opened 2026-03-19
Pass correct log context to suggested user metrics

ds-boyce:
The `suggestedUser:seen` event was passing `ProgressGuide` as the log context in all cases.

### DIFF
--- a/src/analytics/metrics/types.ts
+++ b/src/analytics/metrics/types.ts
@@ -483,8 +483,6 @@ export type Events = {
       | 'DiscoverInterstitial'
       | 'ProfileInterstitial'
       | 'ProfileHeader'
-      | 'FollowDialogProfile'
-      | 'FollowDialogGuide'
       | 'Onboarding'
       | 'SeeMoreSuggestedUsers'
       | 'ProgressGuide'
@@ -500,8 +498,6 @@ export type Events = {
       | 'DiscoverInterstitial'
       | 'ProfileInterstitial'
       | 'ProfileHeader'
-      | 'FollowDialogProfile'
-      | 'FollowDialogGuide'
       | 'Onboarding'
       | 'SeeMoreSuggestedUsers'
     recId?: number | string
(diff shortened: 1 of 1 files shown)


estrattonbailey:
Yeah interesting, good shout. These use different endpoints to fetch data, do we want to conflate follows that happen on data from one endpoint with data from another? I think for Ian's current work the answer is probably no, but maybe we should check with him?

iwsmith:
Eric is right that I want the LogContext to be unique to the surface an event happened on. Ideally there is a 1:1:1 between discover endpoints : UI surface : client Log Context

src/analytics/metrics/types.ts
```diff
@@ -480,11 +480,14 @@ export type Events = {
...
-      | 'Profile'
+      | 'DiscoverInterstitial'
+      | 'ProfileInterstitial'
+      | 'ProfileHeader'
+      | 'FollowDialogFeed'
+      | 'FollowDialogProfile'
+      | 'FollowDialogGuide'
       | 'Onboarding'
-    location: 'Card' | 'Profile' | 'FollowAll'
+    location: 'Card' | 'ProfileHeader' | 'FollowAll'
```
iwsmith: Q: would a profile interstitial and profile header (i.e. click follow) both have the same location? That is fine/expected, since it is the profile page and the log context can disambiguate which modal did it.

src/analytics/metrics/types.ts
```diff
@@ -504,27 +511,24 @@ export type Events = {
...
-      | 'InterstitialProfile'
-      | 'Profile'
+      | 'DiscoverInterstitial'
+      | 'ProfileInterstitial'
+      | 'ProfileHeader'
+      | 'FollowDialogFeed'
+      | 'FollowDialogProfile'
+      | 'FollowDialogGuide'
       | 'Onboarding'
-      | 'ProgressGuide'
```
iwsmith, line 511: Q: what is this `ProgressGuide` and has it gone away?

iwsmith (approved):
minor Q, but looks right to me.

src/analytics/metrics/types.ts
ds-boyce: That’s the old hard-coded value. I restored it for `FollowDialog` vs `FollowDialogWithoutGuide`.

PR #10077, opened 2026-03-19
Fix issue with site loading on older versions of Safari

ds-boyce:
Lookbehind assertions [weren’t supported in WebKit](https://bugs.webkit.org/show_bug.cgi?id=174931) until 2023.

This PR updates the `IMGPROXY_PRESET_RE` to use capturing groups instead.

* [Before](https://regex101.com/r/FBq7mG/1)
* [After](https://regex101.com/r/EHTwQn/1)

The site works again on older versions of Safari. Tested on a simulator with iOS 15.5.

### DIFF
--- a/src/lib/media/util.ts
+++ b/src/lib/media/util.ts
@@ -36,12 +36,14 @@ export type ImgproxyPreset =
   | 'feed_thumbnail'
   | 'download'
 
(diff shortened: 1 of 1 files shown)


mozzius:
@claude this look good to you?

PR #10079, opened 2026-03-20
Fix PostThread scroll positioning on web

mozzius:
## Summary

Fixes two web-only bugs with thread scroll positioning when viewing threads with >5 parent posts:

- **Scroll compensation for parent pagination**: When `onStartReached` paginates in more parents (5→10→15→...), they render above the viewport with no scroll adjustment, pushing the anchor post down. Borrows the content-height delta pattern from MessagesList — tracks `prevContentHeight` and uses `window.scrollBy(delta)` to keep the viewport in place after each pagination.

- **Pagination blocked during background refetch**: The `onStartReached`/`onEndReached` guards checked `isFetching`, which is `true` during background refetches of cached data. When navigating to a previously-visited thread, cached data renders immediately but `isFetching` stays true during the refetch, blocking pagination. Changed to `isPending` which correctly targets the case we care about: no real data has been fetched yet.

Neither issue affects native, which uses `maintainVisibleContentPosition` for scroll compensation.

## Test plan

- [ ] Find a thread with >5 parents on web
- [ ] Cold load → verify anchor pins to top, first 5 parents above
- [ ] Scroll up → verify remaining parents load without anchor jumping
- [ ] Navigate to another thread → back to first thread → verify parents load (not blocked by refetch)
- [ ] Change sort/view params → verify anchor re-pins correctly
- [ ] Test root post (no parents) → no regressions
- [ ] Quick sanity check on native

### DIFF
--- a/src/screens/PostThread/index.tsx
+++ b/src/screens/PostThread/index.tsx
@@ -363,7 +363,10 @@ export function PostThread({uri}: {uri: string}) {
   )
 
   const onStartReached = () => {
-    if (thread.state.isFetching) return
+    // Don't paginate until we have real data. We intentionally don't check
+    // `isFetching` — background refetches shouldn't block pagination when
+    // we already have real cached data.
+    if (thread.state.isPending) return
     // can be true after `prepareForParamsUpdate` is called
     if (deferParents) return
     // prevent any state mutations if we know we're done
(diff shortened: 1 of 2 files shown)


estrattonbailey:
See #10086

PR #10081, opened 2026-03-20
Add Unfollow Confirmation Dialog to profiles, search, notifications and suggested accounts.

oracularhades:
I keep accidentally unfollowing mutuals, it's awkward and involves me DM'ing to explain the unfollow was accidental.

This PR implements an unfollow confirmation dialog to profiles, search, notifications and suggested accounts with the exception of:
- [Irrelevant] The follow button on posts: Irrelevant because it doesn't show if you already follow the author.
- [Annoying] The video feed: It's largely unused and is annoying to test.

Screenshot 2026-03-21 at 1 23 41 AM

_A note about src/components/ProfileCard.tsx:527 (onPressPop): I removed it as it doesn't appear to be used and GestureResponderEvent is no longer passed due to the dialog. Instead e.preventDefault() and e.stopPropagation() are passed before opening the dialog to prevent the double-click behaviour. This seemed like the simplest fix without overcomplicating the unfollow logic._

### DIFF
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -595,7 +595,7 @@ export function FollowButtonInner({
 
       <Prompt.Basic
         control={unfollowPromptControl}
-        title={l(msg`Unfollow account?`)}
+        title={l(msg`Unfollow Account?`)}
         description={l(
           msg`You will no longer follow this account. If you have a prior DM conversation, this account can still message you.`,
         )}
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -11,6 +11,7 @@ import {
   type ModerationOpts,
   RichText as RichTextApi,
 } from '@atproto/api'
+import {msg} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react/macro'
 
 import {getModerationCauseKey} from '#/lib/moderation'
@@ -41,6 +42,7 @@ import {PlusLarge_Stroke2_Corner0_Rounded as Plus} from '#/components/icons/Plus
 import {Link as InternalLink, type LinkProps} from '#/components/Link'
 import * as Pills from '#/components/Pills'
 import {ProfileBadges} from '#/components/ProfileBadges'
+import * as Prompt from '#/components/Prompt'
 import {RichText} from '#/components/RichText'
 import * as Toast from '#/components/Toast'
 import {Text} from '#/components/Typography'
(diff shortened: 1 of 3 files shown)


surfdude29:
I think this is a good idea, and I have one suggested improvement.

For times when someone has decided to prune their Following feed and so will be unfollowing multiple people in a short amount of time, it might be really annoying to have that process now involve two clicks/taps rather than just one at present.

What about having an option in the dialog with a checkbox saying something like:

[  ] `Don't show this confirmation dialog for the next hour`

Then someone who's doing follow pruning could check that and unfollowing would be a one-click action for the next hour. After that one-hour period has elapsed, the confirmation dialog would start appearing again on Unfollow actions.

There are existing timers for things in the app, so hopefully a one-hour timer of that kind wouldn't be too complex to implement?

surfdude29:
Also, I have some suggestions for the dialog copy.

`You will no longer follow this account.` – This seems a little redundant to point out imo, and doesn't convey useful information that isn't already conveyed by the dialog title and the fact that the user clicked Unfollow.

`If you have a prior DM conversation, this account can still message you.` – I think this could be clearer.

So what about:

`You will no longer see posts from this account in your Following feed.` – This is one of the key consequences that unfollowing an account has for the user's experience, so it might be helpful to spell it out?

`Unfollowing this account will not affect any ongoing chat conversations.` – I think this would offer reassurance while hopefully not causing any confusion, and the use of `ongoing` echoes the copy on the chat settings screen.

I suggest `chat` rather than `DMs`, because `DM` is currently only used in UI strings to refer to the Germ DM integration; Bluesky DMs are always referred to as `chat`/`messages`.*

On the dialog title: one of the devs will correct me if I'm wrong on this, but I think dialog titles usually use sentence case rather than title case these days, so I would suggest `Unfollow account?` instead :)

(I think the `Block Account?` and `Unblock Account?` strings just haven't been updated yet.)

*While checking this I noticed that there is actually an exception, I'll open a PR to suggest aligning those strings.

surfdude29:
Also also (sorry for all the comments), while discussing this with Claude he came up with one alternative suggestion to tackle the accidental unfollow problem that might be worth considering, although it might be more technically complex to implement:

> **A brief undo toast instead of a confirmation dialog** – rather than asking before the action, just do it immediately and show an “Unfollowed. Undo?” toast for a few seconds (similar to Gmail’s “Message deleted / Undo” pattern). This is zero extra taps in the happy path, and one tap to recover from a mistake. It’s arguably a better UX pattern than a confirmation dialog in general, though it would be a bigger architectural change since it requires the unfollow to be reversible within a short window.
>
> …
>
> So the viable version would be the delayed approach – buffer the unfollow locally during the toast window, update the UI optimistically, and only fire the API call once the toast expires without the user tapping undo. If they do tap undo, you just discard the pending action and revert the UI state. No network request ever happens.
>
> That fits reasonably well with how the app already handles optimistic updates, but it does add a layer of state management – you’d need to track pending unfollows and handle edge cases like the user navigating away or closing the app during the toast window.

PR #10066, opened 2026-03-18
Add button for creating a new group clip clop

ds-boyce:
This is a fork of `SearchablePeopleList` with logic specific to group clip clops (more to come).

The new component is behind a feature flag; no changes were made to the existing UI.

https://github.com/user-attachments/assets/ff6f6dc2-2eba-4a37-964b-5ef36de610ac

| Android | iOS | Web |
| - | - | - |
| android | ios | web |

PR #10089, opened 2026-03-20
Add list of selectable group clip clop members

ds-boyce:
Step two after #10066.

https://github.com/user-attachments/assets/94f42f3a-dd82-4701-a3bd-0863301d53c1

| Android | iOS | Web |
| - | - | - |
| android | ios | web |

### DIFF
--- a/src/components/dms/InitiateChatFlow.tsx
+++ b/src/components/dms/InitiateChatFlow.tsx
@@ -1,4 +1,11 @@
-import {useCallback, useLayoutEffect, useMemo, useRef, useState} from 'react'
+import {
+  useCallback,
+  useLayoutEffect,
+  useMemo,
+  useReducer,
+  useRef,
+  useState,
+} from 'react'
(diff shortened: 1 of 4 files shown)


src/components/dms/InitiateChatFlow.tsx
```diff
@@ -182,6 +184,7 @@ export function InitiateChatFlow({
         setChatTitle(title)
         LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut)
         setChatState(ChatState.NEW_CHAT)
+        setGroupChatDids([])
```
mozzius: I wonder if it would be neater to bundle these up into a Reducer? What do you think?

ds-boyce: I was considering it. I’ll take a pass at it.

ds-boyce: I did this in #10109 since it was a pretty big change and I wanted to avoid the merge conflicts.

PR #10091, opened 2026-03-20
Display scrollable list of group clip clop members

ds-boyce:
Step three after #10066 and #10089.

https://github.com/user-attachments/assets/2972b9e6-845b-4825-8018-93c0d42fe7f6

| Android | iOS | Web |
| - | - | - |
| android | ios | web |

### DIFF
--- a/src/components/dms/InitiateChatFlow.tsx
+++ b/src/components/dms/InitiateChatFlow.tsx
@@ -1,4 +1,11 @@
-import {useCallback, useLayoutEffect, useMemo, useRef, useState} from 'react'
+import {
+  useCallback,
+  useLayoutEffect,
+  useMemo,
+  useReducer,
+  useRef,
+  useState,
+} from 'react'
 import {LayoutAnimation, TextInput, View} from 'react-native'
 import {moderateProfile, type ModerationOpts} from '@atproto/api'
 import {Trans, useLingui} from '@lingui/react/macro'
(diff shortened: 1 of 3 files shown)


src/components/dms/ChatProfileTabs.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  const [containerWidth, setContainerWidth] = useState(0)
+
+  useAnimatedReaction(
+    () => containerSize.get(),
+    latestValue => {
+      runOnJS(setContainerWidth)(latestValue)
+    },
+    [containerWidth],
+  )
```
mozzius: We don't actually use this on the UI thread, and it's set in an `onLayout` - I think we can simplify to just setting `setContainerWidth` directly in `onLayout` and skip the thread hopping

mozzius: In fact, we don't even use `containerWidth` at all? I think we can delete it

src/components/dms/ChatProfileTabs.tsx
```diff
@@ -0,0 +1,210 @@
...
+        onLayout={e => {
+          containerSize.set(e.nativeEvent.layout.width)
+        }}
+        onScroll={e => {
+          scrollX.set(Math.round(e.nativeEvent.contentOffset.x))
+        }}
+        onScrollBeginDrag={() => {
+          // Remember that you've manually messed with the tabbar scroll.
+          // This will disable auto-adjustment until after next pager swipe or item tap.
+          syncScrollState.set('unsynced')
```
mozzius: I also this this can be deleted as well, seems left over from other component

src/components/dms/ChatProfileTabs.tsx
```diff
@@ -0,0 +1,210 @@
...
+    },
+    [layouts],
+  )
+
+  const handleLayout = useCallback(
+    (index: number) => (e: LayoutChangeEvent) => {
+      runOnUI(onItemLayout)(index, e.nativeEvent.layout)
+    },
+    [onItemLayout],
+  )
```
mozzius: also seems unused

src/components/dms/ChatProfileTabs.tsx
```diff
@@ -0,0 +1,210 @@
...
+  useAnimatedReaction(
+    () => containerSize.get(),
+    latestValue => {
+      runOnJS(setContainerWidth)(latestValue)
+    },
+    [containerWidth],
+  )
+
+  return (
+    <View testID={testID} accessibilityRole="tablist" style={[t.atoms.bg]}>
```
mozzius: nit: tablist I believe is incorrect here

PR #10095, opened 2026-03-21
Prevent render loop when toggling feed preference

ds-boyce:
Using state to manage the previous data comparison here was creating a render loop: Every time there was a new `data` reference, state was mutated during render via `setPrevData`, which triggered a new render, which resulted in a new `data` reference, and so on.

Using a ref to track the previous `data` avoids the render loop.

This was identified by toggling the experimental setting in “Following feed preferences”.

**Before**

https://github.com/user-attachments/assets/d3639b47-589f-4343-b37d-469773f9bb1e

**After**

https://github.com/user-attachments/assets/f44e8d6d-d784-460a-8a24-26917f1bfa16


estrattonbailey:
Confirmed this works too but let's go with the `useEffect` approach as discussed in Slack https://github.com/bluesky-social/social-app/pull/10114#pullrequestreview-4001928811

PR #10096, opened 2026-03-21
Avoid modifying a value returned by a hook

ds-boyce:
Noticed lint flagged this while I was investigating #10095. React Compiler doesn’t like it.

### DIFF
--- a/src/state/queries/preferences/index.ts
+++ b/src/state/queries/preferences/index.ts
@@ -52,7 +52,7 @@ export function usePreferencesQuery() {
         const res = await agent.getPreferences()
 
         // save to local storage to ensure there are labels on initial requests
-        saveLabelers(
+        void saveLabelers(
           agent.did,
           res.moderationPrefs.labelers.map(l => l.did),
         )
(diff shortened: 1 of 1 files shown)


src/state/queries/preferences/index.ts
```diff
@@ -97,13 +104,6 @@ export function usePreferencesQuery() {
     ),
   })
 
-  if (query.data?.birthDate) {
-    /**
-     * The persisted query cache stores dates as strings, but our code expects a `Date`.
-     */
-    query.data.birthDate = new Date(query.data.birthDate)
```
estrattonbailey, line 104: Removing this may break query persistence https://github.com/bluesky-social/social-app/pull/9594#pullrequestreview-3704328369

estrattonbailey: We could improve this comment for sure tho

mozzius: Why don't we just change it to a Date in the select()? That way we don't have to mutate the state

PR #10102, opened 2026-03-23
Fix search typeahead performance with absolute overlay

mozzius:
Note this is mostly for native. Web remains broadly the same

## Summary
- **SearchScreenInner stays always mounted** instead of being toggled with `display: none/flex`, eliminating costly unmount/remount cycles when focusing the search input
- **Typeahead panel is now absolutely positioned** over the content with `a.absolute, a.inset_0` and an opaque background
- **Reanimated layout animations** (`FadeInUp`/`FadeOutDown`) on native for a smooth overlay transition (skipped on web via `native()` wrapper)
- **Accessible**: `accessibilityViewIsModal` and `accessibilityRole="list"` so screen readers treat the overlay properly

## Test plan
- [ ] Focus/unfocus the search input on iOS and Android — should be noticeably faster with no layout jank
- [ ] Verify typeahead suggestions and search history render correctly over the content
- [ ] Verify fade-in-up / fade-out-down animation is smooth on native
- [ ] Verify web has no animation but instant show/hide works correctly
- [ ] Test with VoiceOver (iOS) / TalkBack (Android) — overlay should be announced as a modal list, content behind should be hidden
- [ ] Verify search results / Explore content is not re-mounted when toggling autocomplete

https://claude.ai/code/session_01BhE3QBg6XZwXTwgTbQP9L9

### DIFF
--- a/src/screens/Search/Shell.tsx
+++ b/src/screens/Search/Shell.tsx
@@ -463,7 +463,8 @@ export function SearchScreenShell({
               web: [a.flex_1],
               native: [t.atoms.bg, a.absolute, a.inset_0],
             })}
-            accessibilityViewIsModal>
+            accessibilityViewIsModal
+            accessibilityRole="list">
             {searchText.length > 0 ? (
               <AutocompleteResults
                 isAutocompleteFetching={isAutocompleteFetching}


estrattonbailey:
This is superseded by https://github.com/bluesky-social/social-app/pull/10146 right?

mozzius:
Sure is!

PR #10117, opened 2026-03-24
Add resolution downsizing to image compression

estrattonbailey:
For images that don't compress well, such as those with high levels of detail, we need to size down the max resolution and re-try the compression. Otherwise some images won't compress to <1MB at 4k px, and fail.

This _could_ be a bug even prior to the bump to 4000px^2 resolution. We basically want this util to compress iteratively until it succeeds.

The result of this is that images will be uploaded at the highest resolution and quality possible, but some images will not be uploaded at the max resolution we allow, if they're unable to compress well.

### DIFF
--- a/src/state/gallery.ts
+++ b/src/state/gallery.ts
@@ -204,23 +204,22 @@ export async function compressImage(
   img: ComposerImage,
   options?: {
     highResolution?: boolean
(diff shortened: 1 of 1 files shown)


mozzius:
@claude pls review

PR #10120, opened 2026-03-24
Add some more clarity to the RQ docs

estrattonbailey:


src/state/queries/util.ts
```diff
@@ -14,6 +14,30 @@ import {
...
+  root: string,
+  /**
+   * Any arguments the query depends on, and if changed, should result in the query being refetched.
+   */
+  args?: T,
+) {
+  if (args === undefined) {
+    return [root] as const
+  }
+  return [root, args] as const
```
mozzius: maybe we could add make persisting queries smoother this way? Include a flag to prefix the persistence key?

estrattonbailey: Yeah good call. I'm taking it this far, might as well go all the way.

I got rid of the overrides for clarity. Dead simple now, just migrated the other cases.

PR #10121, opened 2026-03-24
Update web styles for group clip clops

ds-boyce:
Updates chat dialog styles for web.

https://github.com/user-attachments/assets/62aadf0d-e1d2-4d89-b5b6-82025e995fad

### DIFF
--- a/src/components/dms/InitiateChatFlow.tsx
+++ b/src/components/dms/InitiateChatFlow.tsx
@@ -25,13 +25,13 @@ import * as TextField from '#/components/forms/TextField'
 import * as Toggle from '#/components/forms/Toggle'
 import {useInteractionState} from '#/components/hooks/useInteractionState'
 import {
-  ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeft,
-  ArrowRight_Stroke2_Corner0_Rounded as ArrowRight,
+  ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeftIcon,
+  ArrowRight_Stroke2_Corner0_Rounded as ArrowRightIcon,
 } from '#/components/icons/Arrow'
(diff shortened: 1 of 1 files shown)


mozzius:
🔥

src/components/dms/InitiateChatFlow.tsx
```diff
@@ -24,10 +24,14 @@ import {canBeMessaged} from '#/components/dms/util'
 import * as TextField from '#/components/forms/TextField'
 import * as Toggle from '#/components/forms/Toggle'
 import {useInteractionState} from '#/components/hooks/useInteractionState'
-import {ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeft} from '#/components/icons/Arrow'
+import {
+  ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeft,
+  ArrowRight_Stroke2_Corner0_Rounded as ArrowRight,
```
mozzius: extremely small nit: we're really inconsistent on this, but I personally think it would be clearer if we add an `Icon` postfix to these imports, e.g. `ArrowLeft_Stroke2_Corner0_Rounded as ArrowLeftIcon`

mozzius:
Looks good, only one issue I could see: there's an extra item hidden at the bottom! I think we need to add extra padding to the list content to compensate. Unfortunately `contentInsets` is iOS only so you'll probably need to just add bottom padding to `contentContainerStyle`

https://github.com/user-attachments/assets/edcb6ce3-2725-4be2-a9ba-d14bfffd5e26

ds-boyce:
Fixed the footer issue.

src/components/dms/InitiateChatFlow.tsx
ds-boyce: Updated. 🙂

PR #10122, opened 2026-03-24
Fix text overflow in post meta on web

ds-boyce:
Setting `minWidth: 0` here overrides the default flexbox behavior, ensuring the content’s box shrinks to avoid overflow.

| Before | After |
| - | - |
| before | after |

Focused these to web-only since native was already working correctly. Tested on an Android device and iOS simulator.

### DIFF
--- a/src/components/ProfileHoverCard/index.web.tsx
+++ b/src/components/ProfileHoverCard/index.web.tsx
@@ -519,6 +519,7 @@ function Inner({
                 a.leading_snug,
                 a.font_semi_bold,
                 a.self_start,
+                {minWidth: 0},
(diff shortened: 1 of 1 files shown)


LoganDupont:
i'm not sure, but I think this one is also related to it #10125

ds-boyce:
Looks like this issue was introduced with #10105. I think this is the right fix given that change.

I need to look at #10125 as well, since this fix doesn’t address that.

I’ll make a new atom vs inlining style here.

ds-boyce:
Ok, applied the same fix to `ProfileHoverCard`.

hover

PR #10130, opened 2026-03-25
Prevent overflow-x on mobile browsers

ds-boyce:
This was flagged on the Samsung Internet app, although it may affect other browsers as well.

Uses the same breakpoint as [`isMobile`](https://github.com/bluesky-social/social-app/blob/main/src/lib/hooks/useWebMediaQueries.tsx#L11).

| Before | After |
| - | - |
| before | after |

Tested in Samsung Internet and Chrome on a Pixel 8 (Android 16.0), and Safari on an iPhone 17 (iOS 26.3).

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -78,6 +78,13 @@
         min-height: 100%;
         width: 100%;
       }
+      @media (max-width: 799px) {
+        html,
+        body,
+        #root {
+          overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


estrattonbailey:
Also will be superseded by https://github.com/bluesky-social/social-app/pull/10131

PR #10137, opened 2026-03-27
fix: show unfollow button in minimal profile header

### LINKED ISSUE #10124: No way to unfollow from the minimised profile view
### Steps to Reproduce

1. Open a profile page on mobile
2. Scroll down to show the minimised profile view

### Attachments

_No response_

Web (Mobile), Android

### Device Info

Google Pixel 7

Build version 1.119.1; Bundle info: fc402f1 (prod); Bundle date: 26032421; Platform: web; Platform version: 0.0.0; Device ID: 40df6a1f-cbb9-4092-a88a-4801a8768ee6

### Additional Information

On the Android app, there's the minimised view which shows the handle and display name as well as an option to follow if not already. However, when following, the only option is to change post notifications, rather than showing an unfollow button as well. The menu button also doesn't show any option to unfollow either in this view.

Additionally, web on mobile doesn't even have this much; all that's visible is just the profile tabs at the top.

Banner on mobile Android view

Android<br>

Banner on the mobile web view

Web (Mobile)<br>

cyphercodes:
## Problem

When viewing a profile on mobile and scrolling down to show the minimized profile header, users who are already following the account have no way to unfollow from this view. The follow/unfollow button was hidden when `minimal=true` and `profile.viewer?.following=true`.

Fixes #10124

## Solution

Remove the conditional wrapper `(!minimal || !profile.viewer?.following) &&` so the Follow/Unfollow button is always rendered in the minimal header, regardless of follow state.

## Changes

- Always show Follow/Unfollow button in minimal header
- Users can now unfollow directly from the minimized profile view
- No changes to non-minimal header behavior

## Testing

- [x] Verified button renders in minimal header when following
- [x] Verified button renders in minimal header when not following
- [x] Verified button functionality (follow/unfollow) works correctly

### DIFF
--- a/src/screens/Profile/Header/ProfileHeaderStandard.tsx
+++ b/src/screens/Profile/Header/ProfileHeaderStandard.tsx
@@ -379,31 +379,29 @@ export function HeaderStandardButtons({
             </>
           )}
 
-          {(!minimal || !profile.viewer?.following) && (
-            <Button
-              testID={profile.viewer?.following ? 'unfollowBtn' : 'followBtn'}
-              size="small"
-              color={profile.viewer?.following ? 'secondary' : 'primary'}
-              label={
-                profile.viewer?.following
-                  ? _(msg`Unfollow ${profile.handle}`)
-                  : _(msg`Follow ${profile.handle}`)
-              }
-              onPress={
-                profile.viewer?.following ? onPressUnfollow : onPressFollow
-              }>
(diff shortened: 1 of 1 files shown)


cyphercodes:
Closing this PR. After further review, I see the maintainers indicated this behavior is intentional. Apologies for not catching that in the issue discussion earlier.

PR #10138, opened 2026-03-27
Update emoji regex to detect variations and sequences

ds-boyce:
Updates the `SINGLE_EMOJI_RE` regex to support emoji with variations like ⏹️ and multi-part emoji like 👨‍👩‍👧‍👧.

| Before | After |
| - | - |
| before | after |


mozzius:
@claude does this look right to you

PR #10145, opened 2026-03-30
Create global keyboard shortcut handler

ds-boyce:
Uses [`react-hotkeys-hook`](https://npmx.dev/package/react-hotkeys-hook) to replace `useComposerKeyboardShortcut` with `useKeyboardShortcuts` and adds an additional `/` shortcut for focusing search inputs.

### DIFF
--- a/src/App.web.tsx
+++ b/src/App.web.tsx
@@ -5,11 +5,10 @@ import './style.css'
 import {Fragment, useEffect, useState} from 'react'
 import {KeyboardProvider as KeyboardControllerProvider} from 'react-native-keyboard-controller'
 import {SafeAreaProvider} from 'react-native-safe-area-context'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
+import {useLingui} from '@lingui/react/macro'
 import * as Sentry from '@sentry/react-native'
-import {HotkeysProvider} from 'react-hotkeys-hook'
 
+import {Provider as HotkeysProvider} from '#/lib/hotkeys'
 import {QueryProvider} from '#/lib/react-query'
 import {ThemeProvider} from '#/lib/ThemeContext'
 import {Provider as TranslateOnDeviceProvider} from '#/lib/translation'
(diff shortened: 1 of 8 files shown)


src/App.web.tsx
```diff
@@ -156,8 +157,13 @@ function InnerApp() {
                                                           <HideBottomBarBorderProvider>
                                                             <IntentDialogProvider>
                                                               <TranslateOnDeviceProvider>
-                                                                <Shell />
-                                                                <ToastOutlet />
+                                                                <HotkeysProvider
+                                                                  initiallyActiveScopes={[
+                                                                    'global',
+                                                                  ]}>
```
estrattonbailey: Whatcha think about just tossing this in `#/lib/hotkeys` and wrapping the `useGlobalKeyboardShortcuts` within a re-exported `HotkeysProvider`? Could just do both at once basically.

src/lib/hotkeys/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+        },
+        {
+          requiresSession: true,
+        },
+      ),
+    {scopes: ['global'], description: l`Compose new post`},
+    [openComposer],
+  )
+
+  useHotkeys(Hotkeys.FOCUS_SEARCH, () => handleKey(emitFocusSearch), {
```
estrattonbailey, line 71: This one could really be allowed for un-authed users (and we should work towards allowing search for un-authed users soon)

juliusv:
@ds-boyce FYI, since this got rolled out today, I noticed that it doesn't work on international keyboard layouts: I have a German QWERTZ layout, and I have to press "-" instead of "/" to get the search to activate. I noticed this PR is using `react-hotkeys-hook`, which uses `event.code` instead of `event.key` to determine which key was pressed: https://github.com/JohannesKlauss/react-hotkeys-hook/blob/6d988e3571c25d2e9fb02d9e1ec1ddfd54eff99a/packages/react-hotkeys-hook/src/lib/isHotkeyPressed.ts#L10

Going by https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code, `event.key` would be the right thing to query for:

> To determine what character corresponds with the key event, use the [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property instead.

PR #10149, opened 2026-03-31
Append profile to recent search history

oracularhades:
If you search a profile your query will appear in "Recent searches". However, this feels useless if tapping a previously profile-specific query searches all of Bluesky. This PR updates this inconvenience.

Before:
Screenshot 2026-03-31 at 7 01 01 PM

After:
Screenshot 2026-03-31 at 7 01 49 PM


estrattonbailey:
I'll look at this soon!

PR #10151, opened 2026-03-31
Add `no-extraneous-dependencies` and `no-nodejs-modules` eslint rules

mozzius:
`import-x/no-extraneous-dependencies` disallows importing transient dependencies, which I am strongly in favour of. I listed 3 exceptions: a jest one (who cares), a very specific case where we use a util from `expo-modules-core`, and `@atproto/common-web` because I don't want those to accidentally go out of sync with the version from `@atproto/api`.

This caught:
- usage of `@react-navigation/core` instead of `@react-navigation/native`
- usage of `@atproto/xrpc`. Luckily @matthieusieben recently reexported XRPCError via `@atproto/api`, so adopted that.

`import-x/no-nodejs-modules` disallows importing nodejs builtins, which makes sense because we're not in a node environment.

This caught:
- `import {Buffer} from 'buffer'` - we can just use `window.Buffer` I think.

# Test plan

The only functional change is possibly the Buffer change. We only use it for the downloading a CAR file on native - check that that still works.

### DIFF
--- a/src/lib/media/manip.ts
+++ b/src/lib/media/manip.ts
@@ -321,6 +321,7 @@ export async function saveBytesToDisk(
   bytes: Uint8Array,
   type: string,
 ) {
+  // ideally we'd use `bytes.toBase64()`, but that's only baseline newly available
   let binary = ''
   for (const byte of bytes) {
     binary += String.fromCharCode(byte)


mozzius:
Note: I had to manually deduplicate our atproto packages, so ignore the lockfile edit CI failure

mozzius:
Yep on iOS, will test Android tomorrow

mozzius:
Okay no I was really dumb here. `window.Buffer` doesn't exist, it's a NodeJS thing. `import "buffer"` was actually pulling in a polyfill that was a transitive dependency from `whatwg-url-without-unicode`, which the eslint rule mistook for a nodejs import.

I think we should still get rid of it, so I just replaced it with good old `btoa`

estrattonbailey:
If you've tested it, sgtm 👍

mozzius:
Tested on all platforms

PR #10152, opened 2026-03-31
Remove direct dependency on `@expo/config-plugins`

mozzius:
According to expo-doctor, we should be using `expo/config-plugins` instead. Updates our config plugins to do this

### DIFF
--- a/package.json
+++ b/package.json
@@ -245,7 +245,6 @@
     "@babel/runtime": "^7.26.0",
     "@crowdin/cli": "^4.14.1",
     "@eslint/js": "^9.39.2",
-    "@expo/config-plugins": "~54.0.4",
     "@lingui/babel-plugin-lingui-macro": "^5.9.2",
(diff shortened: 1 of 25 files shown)


plugins/withAndroidDayNightThemePlugin.js
```diff
@@ -1,6 +1,6 @@
 // Based on https://github.com/expo/expo/pull/33957
 // Could be removed once the app has been updated to Expo 53
```
mozzius: we should do this

estrattonbailey:
@mozzius merged the other PR and it created a conflict here. Sounds worth doing, maybe we can knock this out next week?

mozzius:
Ya will rebase, np

mozzius:
@estrattonbailey okie rebased

estrattonbailey (approved):

PR #10157, opened 2026-03-31
[APP-1934] replace image grid layout with carousel

vineyardbovines:
https://github.com/user-attachments/assets/38be98a5-3a23-4d6c-9b58-0be2ab58be4d


src/components/images/Gallery/index.web.tsx
```diff
@@ -0,0 +1,311 @@
...
+            <Text
+              style={[
+                a.font_bold,
+                {fontSize: 11, color: t.atoms.text_inverted.color},
+              ]}>
+              {currentPage + 1}/{images.length}
+            </Text>
+          </View>
+          {currentPage > 0 && (
+            <Pressable
```
ds-boyce: Could these two be merged into a single component, e.g., `GalleryNav`?

src/components/images/Gallery/index.web.tsx
```diff
@@ -0,0 +1,311 @@
...
+}
+
+export function Gallery({
+  images,
+  onPress,
+  onPressIn,
+  viewContext,
+}: GalleryProps) {
+  const t = useTheme()
+  const {_} = useLingui()
```
ds-boyce: nit: We can use `useLingui(t: l)` from `@lingui/react/macro` for new code.

```diff
- _(msg`Opens full image`)
+ l`Opens full image`
```

ds-boyce (approved):
Looks good! Just a couple of nits.

estrattonbailey (changes requested):
Looking solid! Couple things left to hit, some we've talked about already but just summarizing here:
- web
  - should be full-bleed to match native
  - needs click-and-drag, drop the arrows
    - will want to double check this works ok in the composer
  - composer isn't full bleed rn, but we can skip that if unless it's really simple to adjust the composer padding
  - all images are the same aspect ratio atm
- all
  - scroll snapping
  - overall container aspect ratio can pry be more like 3/5 to match designs
  - let's try to lazy load the images
  - image aspect ratio limit is 2/3 or 3/2 in the designs, I think it's not quite that rn in the code

Here's a test case for us http://localhost:19006/profile/test.esb.lol/post/3mihgijiacs2j

src/components/images/Gallery/index.web.tsx
```diff
@@ -0,0 +1,334 @@
...
+    const ar = image.aspectRatio
+    if (ar && ar.width > 0 && ar.height > 0) {
+      const ratio = ar.width / ar.height
+      const w = containerHeight * ratio
+      return Math.max(containerWidth * 0.4, Math.min(w, containerWidth))
+    }
+    return containerWidth
+  }
+
+  // Click-and-drag scrolling
```
mozzius: are you aware of the `useDraggableScroll` hook we have? did you try that?

vineyardbovines: CleanShot 2026-04-02 at 09 05 06

yeah- wasn't quite cutting it. i pivoted to something different and then figured i'd compare the two afterwards to see what the overlap is. pretty sure its because `e.preventDefault()` isn't/can't be called on mousedown here so the browser image drag takes over?

vineyardbovines:
@estrattonbailey made some updates here that should cover your notes- the only thing on web that hasn't been adjusted is scroll snapping. wanted to get your thoughts on the current state before adding it, it was slightly janky when i tossed it in alongside the click+drag. IMO its a bit tighter now but if you/anyone thinks it should be adjusted more lmk

vineyardbovines:
looking good on chrome, firefox, safari 🚀

estrattonbailey (approved):
Let's do it :extreme_teamwork:

catofnineswords:
I wonder if anyone did UAT before releasing this change, as I'm seeing nothing but horrified responses on bsky.

PR #10159, opened 2026-04-01
New rich text composer + autocomplete

estrattonbailey:
Adds two new Bluesky libraries, `@bsky.app/tapper` and `@bsky.app/sift` for rich text and autocomplete, respectively.

Included in this PR is a new `MessageComposer` component for DMs, which is still alpha and will be gated to the Bluesky team while we iron out any kinks.

This PR also contains a new autocomplete view for the search bar in the right nav on desktop. This autocomplete is fully accessible and supports keyboard navigation following the `combobox` pattern.

### DIFF
--- a/src/screens/Search/components/AutocompleteResults.tsx
+++ b/src/screens/Search/components/AutocompleteResults.tsx
@@ -132,5 +132,3 @@ let SearchLinkCard = ({
     </Link>
   )
 }
-SearchLinkCard = memo(SearchLinkCard)
-export {SearchLinkCard}


src/components/Autocomplete/useAutocomplete/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      'autocomplete',
+      {
+        type,
+        query: q,
+      },
+    ],
+    async queryFn() {
+      if (type === 'profile') {
+        // TODO return recents
+        if (!query) return []
```
vineyardbovines: stale closure maybe, wouldn't we want this to be `if (!q)` vs query?

estrattonbailey: Ah great call, did some bad renaming in here

ds-boyce:
Our lint configuration still flags it when it’s not explicit, e.g., when a callback is in the dependency array of a `useEffect`.

We should probably just upgrade to `eslint-plugin-react-hooks@latest` since it’s been updated to work with React Compiler.

src/screens/Search/components/AutocompleteResults.tsx
```diff
@@ -76,3 +83,54 @@ let AutocompleteResults = ({
 }
 AutocompleteResults = memo(AutocompleteResults)
 export {AutocompleteResults}
+
+let SearchLinkCard = ({
```
estrattonbailey, line 87: I just moved this old component here since it's now unused in the `Search` component

estrattonbailey:
@vineyardbovines removed 4 unneeded callbacks 👍

src/screens/Search/components/AutocompleteResults.tsx
ds-boyce: Can we remove the `export` then?

estrattonbailey: Sure, done

src/components/Composer/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+      })}
+    </Text>
+  )
+
+  return (
+    <>
+      <View style={[a.relative, outerStyle]}>
+        {IS_WEB && (
+          <View
+            pointerEvents="none"
```
mary-ext, line 308: this should ideally be marked with the `inert` HTML attribute, not only would it prevent mouse events, it also prevents focus events, hides the text from screen readers, and hides it from Ctrl+F find bar so any text matches inside the composer doesn't end up getting duplicated

PR #10165, opened 2026-04-02
[APP-2039] refactor lightbox orientation transform

vineyardbovines:
locks back to portrait mode before starting the close animation (instead of after), and then added scroll positioning save/restoration to prevent the feed from jumping to the top when orientation changes

CleanShot 2026-04-02 at 13 03 51


src/components/ageAssurance/AgeAssuranceInitDialog.tsx
```diff
@@ -1,6 +1,6 @@
 import {useState} from 'react'
 import {View} from 'react-native'
-import {XRPCError} from '@atproto/api'
+import {XRPCError} from '@atproto/xrpc'
```
mozzius, line 3: No this was correct. You may need to install the latest version of @atproto/api

src/view/com/util/List.tsx
```diff
@@ -174,8 +185,11 @@ let List = forwardRef<ListMethods, ListProps>(
         scrollsToTop={scrollsToTop}
         scrollEventThrottle={1}
         style={style}
-        // @ts-expect-error FlatList_INTERNAL ref type is wrong -sfn
-        ref={ref}
+        ref={node => {
+          localRef.current = node as FlatList_INTERNAL | null
+          if (typeof ref === 'function') ref(node as FlatList_INTERNAL | null)
+          else if (ref) (ref as React.MutableRefObject<any>).current = node
+        }}
```
mozzius, line 192: we have a mergeRefs util which I think should do the trick here?

estrattonbailey (changes requested):
Let's hold off on this for now and discuss what we want to do with the new lightbox this week. If we decide to pick this solution back up, we'll need to update this with Sam's feedback 👍

mozzius:
Partially superceded by #10737

PR #10169, opened 2026-04-03
Fix OOM crash - disable tanstack extension integration

mozzius:
After upgrading react query to latest, it started OOMing on web. I tracked this down to https://github.com/TanStack/query/pull/7326 (v5.32.0)

Seems like something in the tanstack query browser extension is broken for the main feed query. Old RQ versions are immune because the relevant internals are private, and throw instead:

```
Uncaught TypeError: can't access property "length", e.observers is undefined
    m <anonymous code>:1
    s <anonymous code>:1
    N <anonymous code>:1
    s <anonymous code>:1
    r helpers.ts:116
    setInterval handler*14261/Ro/< browserapierrors.ts:107
    main <anonymous code>:1
    sync <anonymous code>:1
    sync <anonymous code>:1
    <anonymous> <anonymous code>:1
```

v5.32.0 unprivates that property, which causes it to explode in some bizarre way

Let's just disable the extension because I think it's the one at fault here

## Test plan

Confirm no OOM on web with the tanstack query browser extension installed


dborysov:
@mozzius, ah, thanks for the research!

Now the extension is using [getObserversCount](https://github.com/TanStack/query/blob/main/packages/query-core/src/query.ts#L381-L383) instead of `observers.length`, so now it works properly on old versions of TQ.

Also I optimized the serialization algorithm (it was recursive, now it's not), so now it works fast on deep structures.

I checked the prod version of bluesky with the latest version of the extension (1.0.2) and it works good. Please let me know if something's still off

mozzius:
Great, thank you!! Will give it a try

PR #10208, opened 2026-04-08
bskyweb: update to golang v1.26

bnewbold:
Keeping up with Go language updates, per our usual process/policy (v1.26.1 has been out for a while).

Ran `go fix`, which applies basic style/lint tweaks, and the changes look safe/trivial, so I rolled them up here.


mozzius:
@claude what do you think

mozzius (approved):
thanks Bryan!

PR #10214, opened 2026-04-09
Disable keyboard preload on iOS to prevent keyboard flash

art9mid:
## Scope
Disable keyboard preload on iOS to prevent the keyboard from briefly flashing on screen during app initialization.

## Implementation
Set `preload={false}` on `KeyboardControllerProvider` in `App.native.tsx`. The preload feature is unnecessary since the app's landing screen has no text inputs — the keyboard will be naturally initialized by the time a user navigates to a screen with inputs (composer, search, messages, etc.).

Reference: https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/keyboard-provider#preload-

## Issue

https://github.com/user-attachments/assets/9c2f3094-dab5-4731-b2ee-23a391e3134e

### DIFF
--- a/src/App.native.tsx
+++ b/src/App.native.tsx
@@ -235,7 +235,7 @@ function App() {
     <Geo.Provider>
       <AppConfigProvider>
         <A11yProvider>
-          <KeyboardControllerProvider>
+          <KeyboardControllerProvider preload={false}>
             <OnboardingProvider>
               <AnalyticsContext>
                 <SessionProvider>


mozzius:
Oh lol I thought I'd already done that: https://github.com/bluesky-social/social-app/pull/9778

Seems it got lost somewhere along the way. Good catch!

mozzius (approved):

efstathiosntonas:
Hey guys, according to the docs: 

> And don't forget to manually [preload](https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/keyboard-controller#preload-) the keyboard later via KeyboardController when appropriate.

So something like this is missing later down the tree? 

```tsx
  useEffect(() => {                                                                                                                                                               
    if (Platform.OS === 'ios') {                                                                                                                                                                                                                                                                                        
      KeyboardController.preload();                                                                                                                                                                                                                                                                                   
    }                                                                                                             
  }, []); 
```

art9mid:
> Hey guys, according to the docs:
> 
> > And don't forget to manually [preload](https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/keyboard-controller#preload-) the keyboard later via KeyboardController when appropriate.
> 
> So something like this is missing later down the tree?
> 
> ```tsx
>   useEffect(() => {                                                                                                                                                               
>     if (Platform.OS === 'ios') {                                                                                                                                                                                                                                                                                        
>       KeyboardController.preload();                                                                                                                                                                                                                                                                                   
>     }                                                                                                             
>   }, []); 
> ```

Makes sense in general, but do we actually have any screens that require the keyboard right on first launch?
As far as I know, we don’t, so I’m not sure there’s a real need to manually preload it at this point.
Let me know what you think

efstathiosntonas:
matters only on the first keyboard open, not a big deal imo. 

> This method preloads the keyboard to prevent noticeable delay when the user focuses the first input after launching the app.

mozzius:
Preloading delays initial launch, which we want to be as fast as possible. So no, we don't want to preload, we want to defer this work until it's actually needed, which is the iOS default behaviour.

efstathiosntonas:
@mozzius on my app I got it on `navigationReady` callback. 

Without it the keyboard delays about 500ms.  

Again, not a big deal, just pointing out the numbers.

PR #10216, opened 2026-04-09
Write chat declaration record in response to different events

ds-boyce:
This PR automatically restricts incoming chats:
- upon signup, if the user is a minor or subject to age assurance laws in their region
- upon editing birthdate to an age <18
- upon any change to age assurance state that results in access to chats being restricted


estrattonbailey:
@claude my boy, please review this one

estrattonbailey (approved):

PR #10217, opened 2026-04-09
Start integrating group clip clop settings

ds-boyce:
* Opens group settings directly from the “…” button.
* Populates avatar bubbles and member list from convo data.
* Adds ability to mute and unmute.
* Adds ability to block and unblock individual members.
* Removes hard-coded “Added by“ line until we have the requisite data.

More to come later.

muted

### DIFF
--- a/src/screens/Messages/ConversationSettings.tsx
+++ b/src/screens/Messages/ConversationSettings.tsx
@@ -659,6 +659,8 @@ function SettingsHeader({
   const t = useTheme()
   const {t: l} = useLingui()
 
+  const convoState = useConvo()
+
   const {mutate: muteConvo} = useMuteConvo(convo.id, {
     onSuccess: data => {
       if (data.convo.muted) {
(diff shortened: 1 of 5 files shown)


src/screens/Messages/ConversationSettings.tsx
```diff
@@ -252,13 +311,14 @@ function Member({
   profile,
   status,
 }: {
-  profile: bsky.profile.AnyProfileView
+  profile: Shadow<bsky.profile.AnyProfileView>
   status: 'admin' | 'member' | 'invited'
```
mozzius: Ideally let's use `owner` instead of `admin` for clarity

src/screens/Messages/ConversationSettings.tsx
```diff
@@ -541,39 +630,80 @@ function MemberMenu({
...
   const t = useTheme()
   const {t: l} = useLingui()
+
+  const convoState = useConvo()
+  const {currentAccount} = useSession()
+
+  const isAdmin =
+    currentAccount?.did == null
+      ? false
+      : convoState.getPrimaryMember?.()?.did === currentAccount.did
```
mozzius: This doesn't seem to be working for me? Might be shitty code on my part in `getPrimaryMember` lol

mozzius (approved):
:chefs_kiss:

src/screens/Messages/ConversationSettings.tsx
ds-boyce: Yeah I saw some weird behavior with it when I tried to create my own group chat. Going to leave this as-is for now and we can follow up.

ds-boyce: Not “yeah” to shitty code, incidentally. 😅 I’m sure we’ll find the root cause.

PR #10218, opened 2026-04-09
Optionally animate avatar bubbles

ds-boyce:
Not currently enabled (defaults to `false`).

https://github.com/user-attachments/assets/030dcdc2-178d-44ca-9ce8-10124fc4d8d0

### DIFF
--- a/src/components/AvatarBubbles.tsx
+++ b/src/components/AvatarBubbles.tsx
@@ -1,4 +1,13 @@
+import {useCallback, useEffect} from 'react'
 import {type StyleProp, View, type ViewStyle} from 'react-native'
+import Animated, {
+  Easing,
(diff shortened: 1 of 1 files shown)


mozzius:
It might be neater to do this declaratively through layout animations?

ds-boyce:
As long as we don’t change the `animate` value after mount. I can add a note.

ds-boyce:
Actually, looks like the declarative animation overrides `transform`, so we lose the positioning of the avatars. I’d like to keep the single `transform` vs using absolute positioning. So that’s a second caveat for the declarative approach.

What do you think?

mozzius:
Hmm no okay that's a good reason I think! let's keep this approach!

PR #10231, opened 2026-04-10
Remove own reaction from reactions dialog on tap

ds-boyce:
Your reaction is sorted to the top of the list, and removed if tapped.

https://github.com/user-attachments/assets/5e321a83-070a-4c43-bf08-d70fe79c7d4d

web

### DIFF
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -467,7 +467,15 @@ let MessageItem = ({
                   squaredTopCorner={squaredTopCorner || hasEmbedAndText}
                 />
               )}
-              {appliedReactions}
+              <View
+                style={[
+                  !isFromSelf &&
+                    isGroupChat && {
+                      paddingLeft: AVATAR_SIZE,
+                    },
+                ]}>
+                {appliedReactions}
+              </View>
             </ActionsWrapper>
           </View>
         </View>
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -588,6 +588,7 @@ function ReactionsDialog({
         totalReactions={reactions?.length ?? 0}
         onFilter={handleFilter}
       />
+      <Dialog.Close />
     </>
   )
 
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -327,8 +327,8 @@ let MessageItem = ({
               t.atoms.bg_contrast_25,
               t.atoms.shadow_sm,
               {
-                paddingTop: 3,
-                paddingBottom: 3,
+                paddingTop: platform({android: 2, default: 3}),
+                paddingBottom: platform({android: 2, default: 3}),
                 transform: [{translateY: -8}],
               },
             ]}
@@ -342,7 +342,12 @@ let MessageItem = ({
                 layout={native(LinearTransition.delay(300))}
                 key={group.value}
                 style={[a.py_2xs]}>
-                <Text emoji style={[a.text_xs]}>
+                <Text
+                  emoji
+                  style={[
+                    a.text_xs,
+                    {textAlignVertical: 'center', includeFontPadding: false},
+                  ]}>
                   {group.value}
                 </Text>
               </Animated.View>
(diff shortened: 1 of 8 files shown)


mozzius:
Could you add `Dialog.Close` to the web version?

mozzius:
Also, I think the names need to be center-aligned, rather than start-aligned, if you compare to the designs. The text is also very small/too close together, at least on web

Screenshot 2026-04-13 at 13 43 47

Also, real nitpicky but looks like the border of the active tab should blend with the background

Screenshot 2026-04-13 at 13 43 50

src/components/dms/MessageItem.tsx
```diff
@@ -606,7 +640,7 @@ function ReactionsDialog({
                     {displayName}
                   </Text>
                   <Text style={[a.text_xs, t.atoms.text_contrast_medium]}>
-                    {handle}
+                    {isFromSelf ? l`Tap to remove` : handle}
                   </Text>
```
mozzius: These two `Text`s should probably use the `numberOfLines` prop

src/components/dms/MessageItem.tsx
```diff
@@ -568,11 +559,24 @@ function ReactionsDialog({
...
       <Dialog.ScrollableInner
         label={l`Reactions`}
         contentContainerStyle={[a.pt_0]}
+        header={IS_WEB ? header : null}
         style={[web({maxWidth: 400})]}>
         {filteredMembers.map(profile => {
+          const isFromSelf = currentAccount?.did === profile.did
+
           const displayName = sanitizeDisplayName(
             profile?.displayName || sanitizeHandle(profile?.handle ?? ''),
```
mozzius: `createSanitizedDisplayName` is a little more ergonomic :)

src/components/dms/MessageItem.tsx
```diff
@@ -584,16 +588,46 @@ function ReactionsDialog({
             ? new RichTextAPI({text: reaction.value})
             : undefined
 
-          return rt ? (
-            <View
+          return reaction && rt ? (
+            <Pressable
+              accessibilityRole="button"
+              accessibilityHint={l`Tap to remove your ${reaction.value} reaction`}
```
mozzius: Not ideal that if it's not from self, this is a no-op. I'd prefer if it was abstracted out to a component where if self, it's a Button, and if not self, it's a View

mozzius:
The reaction bubble needs to be smaller (24px high), shifted in a bit, and the avatar should be aligned with the message bubble, not the reaction bubble (I think you mentioned this already)

Screenshot 2026-04-13 at 14 09 12
Screenshot 2026-04-13 at 14 08 59

src/components/dms/MessageItem.tsx
```diff
@@ -498,17 +503,21 @@ export {MessageItemMetadata}
 function ReactionsDialog({
```
mozzius: Can we move this to it's own file? MessageItem.tsx is getting chunky

src/components/dms/MessageItem.tsx
```diff
@@ -635,10 +768,17 @@ function ReactionTabs({
   return (
     <View accessibilityRole="list" style={[t.atoms.bg]}>
       <DraggableScrollView
+        ref={scrollViewRef}
         horizontal={true}
         showsHorizontalScrollIndicator={false}
         onScroll={e => {
```
mozzius: do we need to enable a scroll throttle (`scrollEventThrottle`) here? not that it should be a huge issue either way

mozzius (changes requested):
Looking sooo good! Great work, looks so much better esp. on web

I'm not 100% sold of the half-sheet however; seems less than ideal, and doesn't appear to work properly on Android.

https://github.com/user-attachments/assets/ce42af23-cdc6-4a82-b76c-831610560895

Suggestion: can we just use a normal auto-sized scrollview sheet, and set a `minHeight` on the content at `screenHeight / 2`? That way, it can grow naturally which might just be the simpler and nicer way to do it. Wouldn't completely match designs but I suspect it be the pragmatic path

mozzius:
Ah found a bug: it doesn't support the case where you add multiple reactions (you can add up to 5). I would expect "all" here to have 7 items

Fix here: https://github.com/bluesky-social/social-app/pull/10267

Screenshot 2026-04-16 at 14 31 19

mozzius (approved):
Good stuff

PR #10233, opened 2026-04-11
Disable viewport zoom in prod

### LINKED ISSUE #8494: input font-size causes zoom on mobile web (iOS)
### Steps to Reproduce

1. Browse the app in mobile web on iOS with system default font sizes
2. Use the search input

This will cause the browser to zoom because the font size is less than **16px** (currently **15px**). The zoom remains after navigating, so you constantly have to zoom back out after using search (and probably other inputs).

Web (Mobile)

### Device Info

iPhone 14

1.103.0

### Additional Information

This is a long time accessibility thing for iOS, the only solution is to make the font-size on inputs 16px or larger. 

https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/

mozzius:
Adds `maximum-scale=1` to the viewport meta tag. We were using this in dev but not prod. Motivation is that it fixes the DM input on mobile safari

Fixes #8494

Downside is that this disables zooming in on the lightbox. We should probably add support to the lightbox directly rather than leaving this to the browser, and it greatly improves the sign-in/search/dm experience so imo it's a reasonable compromise. The app already is primarily designed for small screens already

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -9,7 +9,7 @@
      -->
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1.00001, viewport-fit=cover"
+      content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1, viewport-fit=cover"
     />
     <!--
       Preconnect to essential domains
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1, viewport-fit=cover">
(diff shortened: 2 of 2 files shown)


web/index.html
```diff
@@ -9,7 +9,7 @@
      -->
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, minimum-scale=1, maximum-scale=1.00001, viewport-fit=cover"
```
mozzius, line 12: `maximum-scale=1.00001` is weird, presumably motivated by something or other but let's just simplify to 1

vineyardbovines (approved):

web/index.html
rebecaruiz: Thank you. You just disabled zooming with a simple gesture on mobile web for visually impaired users. Accessibility champions.

mozzius:
Feedback noted. I'll find a better fix for what I'm trying to solve here

mozzius:
The fix has been deployed. Sorry about this!

PR #10240, opened 2026-04-13
[APP-2066] Migrate from Tenor to KLIPY

vineyardbovines:
~needs https://github.com/bluesky-social/tango/pull/878~

basically a 1:1 swap from tenor to KLIPY, which provides a drop-in migration with v2 endpoints


src/components/dialogs/GifSelect.tsx
```diff
@@ -93,11 +98,14 @@ function GifList({
   const [undeferredSearch, setSearch] = useState('')
   const search = useThrottledValue(undeferredSearch, 500)
   const {height} = useWindowDimensions()
+  const useKlipy = ax.features.enabled(ax.features.KlipyGifProviderEnable)
```
ds-boyce: The `use` prefix is generally reserved for hooks, but I don’t think we have a lint rule to flag it. Maybe `klippyEnabled`?

src/components/dialogs/GifSelect.tsx
```diff
@@ -213,13 +227,21 @@ function GifList({
                 sideBorders={false}
                 topBorder={false}
                 errorTitle={_(msg`Failed to load GIFs`)}
-                errorMessage={_(msg`There was an issue connecting to Tenor.`)}
+                errorMessage={
+                  useKlipy
+                    ? _(msg`There was an issue connecting to KLIPY.`)
```
ds-boyce: Nit: Can use `{t: l}` from `import {useLingui} from '@lingui/react/macro'`

src/state/queries/klipy.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    const res = await fetch(urlFn(params.toString()), {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    })
+    if (!res.ok) {
+      throw new Error('Failed to fetch KLIPY API')
```
ds-boyce: Nit: This is consistent with `tenor.ts` but we could add `res.status` to this error message.

ds-boyce (commented):
Looks great! Nice work.

Tried to test locally, but I got an error message: “There was an issue connecting to KLIPY.” Something on my end?

vineyardbovines:
CORS-related PR was merged in tango, once deployed it should resolve klipy not loading

estrattonbailey (changes requested):
Nice, looks like the API is working now!

Display should match what we have in prod for Tenor, something must be off with the "display as GIF check".

Image

Image

src/analytics/features/types.ts
```diff
@@ -13,6 +13,7 @@ export enum Features {
   ImageUploadsBlobSize2mbEnabled = 'image_uploads:blob_size_2mb:enabled',
   GroupChatsEnable = 'group_chats:enable',
   DmsNewMessageComposerEnable = 'dms:new_message_composer:enable',
+  KlipyGifProviderEnable = 'klipy_gif_provider:enable',
```
estrattonbailey, line 16: I added this to Growthbook btw 👍

estrattonbailey:
Just noting (shared video is Slack) the video preview and posts still don't seem to work for me

src/lib/strings/embed-player.ts
```diff
@@ -628,3 +645,86 @@ export function isTenorGifUri(url: URL | string) {
...
+    const webmSlug = playerUrl.searchParams.get('webm')
+    const mp4Slug = playerUrl.searchParams.get('mp4')
+    const slug = IS_WEB_SAFARI ? mp4Slug : webmSlug
+    const ext = IS_WEB_SAFARI ? 'mp4' : 'webm'
+
+    if (slug) {
+      const parts = playerUrl.pathname.split('/')
+      parts[parts.length - 1] = `${slug}.${ext}`
+      playerUrl.pathname = parts.join('/')
+    }
```
estrattonbailey, line 703: If we don't get a slug here we can't play it on web, right? Should we add an `else` here and return `success` false? Falling back to the link card is A-OK in this case

estrattonbailey (approved):
LGTM! Just the one question

src/lib/strings/embed-player.ts
vineyardbovines: ooh yes thank you, good catch. added

PR #10251, opened 2026-04-14
Fix stale searchText in search submit handler

estrattonbailey:
The onSubmit handler captured searchText from the render closure, so if the user typed fast and hit enter before React committed the last keystroke's re-render, the search would use a stale value. Use a ref instead so the handler always reads the latest searchText at call time.

### DIFF
--- a/src/screens/Search/Shell.tsx
+++ b/src/screens/Search/Shell.tsx
@@ -96,9 +96,12 @@ export function SearchScreenShell({
   const [activeTab, setActiveTab] = useState(() => getTabIndex(tabParam))
 
   // Query terms
-  const [searchText, setSearchText] = useState<string>(queryParam)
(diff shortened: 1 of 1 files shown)


src/screens/Search/Shell.tsx
```diff
@@ -97,6 +97,8 @@ export function SearchScreenShell({
 
   // Query terms
   const [searchText, setSearchText] = useState<string>(queryParam)
+  const searchTextRef = useRef(searchText)
+  searchTextRef.current = searchText
```
mozzius: surely that gets set at the same time as the render? you'd need to be setting it in the onChange handler no?

estrattonbailey: good catch, that definitely seems right to me

PR #10252, opened 2026-04-14
[Chat] Transparent blur header for chat (iOS 26)

mozzius:
## Summary

- Wraps the chat header in `<ScrollEdgeEffect edge="top">` for a progressive blur effect on iOS 26, using the existing `@bsky.app/expo-scroll-edge-effect` library
- Removes the screen's safe area top padding on liquid glass so the header owns the full top edge
- Refactors the chat list's scroll-to-bottom system to be more robust:
  - Initial scroll is no longer gated on `isAtBottom` (which `contentInset` could invalidate via early scroll events)
  - Replaces the 100ms `setTimeout` hack with a footer sentinel `onLayout` + `requestAnimationFrame` — deterministic and faster

https://github.com/user-attachments/assets/748962ec-6d38-44f1-8282-eb6bd379c71f

## Test plan

- [ ] iOS 26: open a chat — header should have progressive blur as messages scroll behind it
- [ ] iOS < 26: normal opaque header, no visual change
- [ ] Web: no change
- [ ] Chat with history loading: overlay stays until history finishes, then reveals smoothly
- [ ] Empty chat: shows immediately
- [ ] Send message while at bottom: auto-scrolls
- [ ] Scroll up, receive message: does NOT auto-scroll
- [ ] Background and return to same chat: re-initializes correctly

### DIFF
--- a/src/components/AvatarBubbles.tsx
+++ b/src/components/AvatarBubbles.tsx
@@ -26,6 +26,7 @@ export function AvatarBubbles({profiles: allProfiles, size = 'large'}: Props) {
         x={-2}
         y={-2}
         style={[a.z_20]}
+        includeProfileBorder
       />
       <AvatarBubble
         profile={profiles[1]}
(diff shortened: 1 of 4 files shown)


ds-boyce (approved):
🥂

PR #10254, opened 2026-04-14
Add header to group clip clops

ds-boyce:
Implements the UI for the group chat header. Functionality will be added in a follow-up PR.

header

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -61,6 +61,7 @@ import {useAnalytics} from '#/analytics'
 import {IS_ANDROID, IS_NATIVE, IS_WEB} from '#/env'
 import {ChatStatusInfo} from './ChatStatusInfo'
 import {MessageInputEmbed, useMessageEmbed} from './MessageInputEmbed'
+import {MessagesListHeader} from './MessagesListHeader'
 import {KeyboardStickyView} from './vendor/KeyboardStickyView'
 
 function MaybeLoader({isLoading}: {isLoading: boolean}) {
@@ -383,7 +384,7 @@ export function MessagesList({
           profile={convoState.convo.members.find(
             member => member.did === item.message.sender.did,
           )}
-          isGroupChat={convoState.getGroupInfo?.() != null}
+          isGroupChat={convoState.isGroup()}
         />
       )
     } else if (item.type === 'deleted-message') {
@@ -448,7 +449,12 @@ export function MessagesList({
             showsVerticalScrollIndicator={!IS_ANDROID}
             scrollEventThrottle={100}
             ListHeaderComponent={
-              <MaybeLoader isLoading={convoState.isFetchingHistory} />
+              <>
+                <MaybeLoader isLoading={convoState.isFetchingHistory} />
+                {convoState.isGroup() ? (
+                  <MessagesListHeader convoState={convoState} />
+                ) : null}
+              </>
             }
             // native only (prop is not supported on web)
             renderScrollComponent={renderScrollComponent}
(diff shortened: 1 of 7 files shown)


src/screens/Messages/components/MessagesListHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+import {type ConvoState} from '#/state/messages/convo/types'
+import {useSession} from '#/state/session'
+import {atoms as a, useTheme} from '#/alf'
+import {AvatarBubbles} from '#/components/AvatarBubbles'
+import {Button, ButtonIcon, ButtonText} from '#/components/Button'
+import {ChainLink_Stroke2_Corner0_Rounded as ChainLinkIcon} from '#/components/icons/ChainLink'
+import {PersonPlus_Stroke2_Corner0_Rounded as PersonPlusIcon} from '#/components/icons/Person'
+import {Text} from '#/components/Typography'
+
+export function MessagesListHeader({convoState}: {convoState: ConvoState}) {
```
mozzius: `Header` is maybe not the right word... info panel?

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -448,7 +449,12 @@ export function MessagesList({
             showsVerticalScrollIndicator={!IS_ANDROID}
             scrollEventThrottle={100}
             ListHeaderComponent={
-              <MaybeLoader isLoading={convoState.isFetchingHistory} />
+              <>
+                <MaybeLoader isLoading={convoState.isFetchingHistory} />
+                {convoState.isGroup() ? (
+                  <MessagesListHeader convoState={convoState} />
+                ) : null}
```
mozzius: We only want to show this once we've reached the start of the chat. So I think we need to retrieve from the convo state whether we've run out of history to fetch

mozzius (changes requested):

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -416,6 +426,8 @@ export function MessagesList({
     [inputHeightUI],
   )
 
+  console.log('DEBUG >>>', 'convoState.hasAllHistory', convoState.hasAllHistory)
```
mozzius, line 429: log

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -448,7 +460,12 @@ export function MessagesList({
             showsVerticalScrollIndicator={!IS_ANDROID}
             scrollEventThrottle={100}
             ListHeaderComponent={
-              <MaybeLoader isLoading={convoState.isFetchingHistory} />
+              <>
+                <MaybeLoader isLoading={convoState.isFetchingHistory} />
+                {convoState.isGroup() && convoState.hasAllHistory ? (
```
mozzius, line 465: I think we should show this for all convos. but let's do it in a followup

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -150,10 +158,12 @@ export function MessagesList({
...
-  if (prevHasScrolled.current && !hasScrolled) {
-    hasInitiallyScrolled.current = false
-  }
-  prevHasScrolled.current = hasScrolled
+  useLayoutEffect(() => {
+    if (prevHasScrolled.current && !hasScrolled) {
+      hasInitiallyScrolled.current = false
+    }
+    prevHasScrolled.current = hasScrolled
+  }, [hasScrolled])
```
mozzius, line 166: I'm getting an intermittent problem with an infinite spinner when opening the chat. might be due to this? but equally could just be the outage

mozzius: Let's assume it's fine. I'll keep an eye out for it

mozzius (approved):
Nice!

PR #10256, opened 2026-04-15
Stop passing `_` around in composer functions

mozzius:
We pass around `_` a bunch - we should be passing around `i18n` instead. This lets use use the macro hook.

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -54,9 +54,8 @@ import {
   type BskyAgent,
   type RichText,
 } from '@atproto/api'
-import {msg, plural} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
-import {Trans} from '@lingui/react/macro'
+import {plural} from '@lingui/core/macro'
+import {Trans, useLingui} from '@lingui/react/macro'
 import {useNavigation} from '@react-navigation/native'
 import {useQueryClient} from '@tanstack/react-query'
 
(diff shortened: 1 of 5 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -1189,7 +1185,13 @@ export const ComposePost = ({
...
             onScroll={scrollHandler}
             contentContainerStyle={a.flex_grow}
-            style={a.flex_1}
+            style={[
+              a.flex_1,
+              web({
+                scrollbarGutter: 'stable',
+                scrollbarColor: `${t.palette.contrast_200} transparent`,
+              }),
+            ]}
```
mozzius, line 1194: Also made the scrollbar nicer on web :)

Image

src/view/com/composer/Composer.tsx
```diff
@@ -197,12 +196,13 @@ export const ComposePost = ({
...
 }) => {
   const {currentAccount} = useSession()
+  const t = useTheme()
   const ax = useAnalytics()
   const agent = useAgent()
   const queryClient = useQueryClient()
   const currentDid = currentAccount!.did
   const {closeComposer} = useComposerControls()
-  const {_} = useLingui()
+  const {t: l, i18n} = useLingui()
```
estrattonbailey, line 205: So we know that passing `l` around won't work since it's evaled as a macro. We're sure i18n works OK for passing around like this?

mozzius: Yes, because `i18n` isn't a macro. `msg`, the message descriptor, is the macro, so `i18n._(msg``)` works fine

estrattonbailey: Cool that was my assumption, haven't had a chance to deeply grok v5

estrattonbailey (approved):

PR #10258, opened 2026-04-15
[Chat] Split view layout for web

mozzius:
This is a chunky one. Sorry in advance

- Adds a "split-view" layout to chats
- Reworks empty states for all(?) the chat screens
- Reworks request button

### Messages screen

Screenshot 2026-05-11 at 17 24 07

### Convo screen

Screenshot 2026-05-11 at 17 24 32

### Requests screen

Screenshot 2026-05-11 at 17 29 14

### Empty states

Screenshot 2026-05-11 at 17 23 11


ds-boyce:
Should we block the left column for age assurance? In some cases, you can still see who’s messaged you and the most recent message, e.g., “Here’s a reply”. Also, the new message composer is still accessible.

aa

mozzius:
Great catch, I'll need to take a look

ds-boyce:
“Chats” breaks awkwardly at mid resolution (common for tablets). Maybe we can set a breakpoint for the requests button that removes the text label so we have more breathing room.

break

mozzius:
@claude review please!

ds-boyce (approved):
LGTM! This is a huge improvement.

PR #10259, opened 2026-04-15
Enforce using named imports from 'react'

ds-boyce:
There were only four new cases of this since the codemod in #10032, so this just prevents us from adding more.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -250,7 +250,7 @@ export default defineConfig(
       '@typescript-eslint/prefer-promise-reject-errors': 'warn',
       '@typescript-eslint/await-thenable': 'warn',
 
-      "no-restricted-imports": ["warn", {
+      "no-restricted-imports": ["error", {
         "paths": [{
           "name": "react",
           "importNames": ["React", "default"],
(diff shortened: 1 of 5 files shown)


src/lib/hooks/useAnimatedValue.ts
```diff
@@ -1,12 +1,12 @@
...
 export function useAnimatedValue(initialValue: number) {
-  const lazyRef = React.useRef<Animated.Value>(undefined)
+  const lazyRef = useRef<Animated.Value>(undefined)
 
   if (lazyRef.current === undefined) {
     lazyRef.current = new Animated.Value(initialValue)
   }
 
-  return lazyRef.current as Animated.Value
+  return lazyRef.current
```
ds-boyce, line 11: ESLint auto-fixed this since the type assertion is redundant.

eslint.config.mjs
```diff
@@ -250,6 +250,14 @@ export default defineConfig(
       '@typescript-eslint/prefer-promise-reject-errors': 'warn',
       '@typescript-eslint/await-thenable': 'warn',
 
+      "no-restricted-imports": ["warn", {
```
mozzius: maybe make this an error?

mozzius (approved):

eslint.config.mjs
ds-boyce: Ok! I’ll have to update some files outside of `src` then.

ds-boyce: Actually, looks like the `lint` script only targets `src`, so I’ll just ignore those.

PR #10261, opened 2026-04-15
[APP-2067] Rebuild GIF Dialog

vineyardbovines:
rebuilds the GIF dialog; enables masonry layout and categories. search autocompletes and shows relevant gifs via typeahead

image


src/features/gifPicker/components/GifCategoryPills.tsx
```diff
@@ -0,0 +1,96 @@
...
+  hasRecents,
+}: {
+  activeId: string
+  onSelect: (category: GifCategory) => void
+  hasRecents: boolean
+}) {
+  // Subscribe to locale changes so i18n._() returns the current translation.
+  // The lingui-msg-rule lint rule forbids _() with variables, so we use
+  // i18n._() directly to translate the MessageDescriptor from GIF_CATEGORIES.
+  useLingui()
```
estrattonbailey: Hmmm interesting, I see the exported config used elsewhere that makes this necessary. Leaving this as a note to myself.

src/features/gifPicker/components/GifCategoryPills.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (category.id === 'recents' && !hasRecents) return null
+        const isActive = category.id === activeId
+        return (
+          <Button
+            key={category.id}
+            label={i18n._(category.label)}
+            aria-current={isActive ? 'true' : undefined}
+            onPress={() => onSelect(category)}
+            size="small"
+            variant={isActive ? 'solid' : 'ghost'}
```
estrattonbailey: We're trying to move away from using `variant` hence the deprecation notice on it. `color=secondary_inverted` might be the move here?

src/features/gifPicker/components/GifCategoryPills.tsx
```diff
@@ -0,0 +1,96 @@
...
+            color="secondary"
+            shape="round">
+            <PillIcon icon={category.icon} />
```
estrattonbailey: Can we just use `ButtonIcon` here?

src/features/gifPicker/components/GifPickerPlaceholder.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const {_} = useLingui()
+
+  const emptyMessage = isSearching
+    ? _(msg`No GIFs found for "${query}".`)
+    : isRecentsEmpty
+      ? _(msg`No recent GIFs yet. Pick one to see it here.`)
+      : _(msg`No GIFs to show right now. Try again in a moment.`)
+
+  return (
+    <ListMaybePlaceholder
```
estrattonbailey, line 43: Note to self: just wanna see what this look like for this use case.

Edit: oh I see we already have this in the existing dialog.

src/features/gifPicker/hooks/useGifPickerData.ts
```diff
@@ -0,0 +1,51 @@
...
+  useTenorGifSearchQuery,
+} from '#/state/queries/tenor'
+import {useAnalytics} from '#/analytics'
+
+/**
+ * Single entry point for the GIF picker's data layer. Wraps the Klipy/Tenor
+ * feature-flag split and the featured-vs-search branching so the UI only ever
+ * consumes one query result.
+ *
+ * The Tenor path is kept alive until the Klipy rollout is complete.
```
estrattonbailey: I like your thinking here but tbh we can pry ditch Tenor in this PR if it's any help to ya. We'll roll out Klipy and ensure we're happy with it before we launch this, so could cut scope to keep things simple.

src/state/queries/gif.ts
estrattonbailey, line 1: Really like this refactor! I actually think we could consider going further and putting this in a `utils` file in the `gifPicker` dir and even moving the queries into that dir 🤔 whatcha think of that?

src/features/gifPicker/components/GifCategoryPills.tsx
estrattonbailey: Now that I'm playing with it I think the inverted color could be good to make the selected one more prominent too. That or primary could work imo.

estrattonbailey (commented):
Looking great!

One major issue: with a couple recents saved, if I search for something, clear the search, and go back to recents, I see leftover results in addition to my recents. Maybe a key issue?

One general note: for new code we should use the new Lingui macro imports and like `const {t: l} = useLingui()` stuff 👍

surfdude29:
This looks very nice 👌

A few small points I noticed from testing on the preview branch, which are demonstrated in the screen recording below:

(1) Switching between categories seems to reload the dialog or something, which causes a bit of a strange visual effect?

(2) When scrolling, it looks a little odd to me that the category icons float over the GIFs – would it make sense to make them part of the search bar/header with a solid background?

(3) When I type a letter in the search bar, it causes the keyboard to disappear, and the character doesn't appear in the search bar

(4) When I'm in the search results screen, how do I return to the default view, with categories etc?

https://github.com/user-attachments/assets/a238c206-fd5f-4a03-a532-4a741f4283a5

src/features/gifPicker/GifPickerDialog.tsx
```diff
@@ -0,0 +1,186 @@
...
+}: {
+  controlRef: React.RefObject<{open: () => void} | null>
+  onClose?: () => void
+  onSelectGif: (gif: Gif) => void
+}) {
+  const control = Dialog.useDialogControl()
+
+  useImperativeHandle(controlRef, () => ({
+    open: () => control.open(),
+  }))
```
estrattonbailey: I realize this is leftover from the prev impl, but we might as well just hoist `useDialogControl` up to the `SelectGifBtn` instead of doing the imperative handling

src/features/gifPicker/components/GifCategoryPills.tsx
```diff
@@ -0,0 +1,81 @@
...
+  {id: 'love', icon: Heart, label: msg`Love`, searchterm: 'love'},
+  {id: 'happy', icon: EmojiSmile, label: msg`Happy`, searchterm: 'happy'},
+  {id: 'sad', icon: EmojiSad, label: msg`Sad`, searchterm: 'cry'},
+  {
+    id: 'party',
+    icon: Celebrate,
+    label: msg`Party`,
+    searchterm: 'congratulations',
+  },
+  {id: 'yes', icon: Shaka, label: msg`Yes`, searchterm: 'yes'},
```
estrattonbailey: Nit: while we're here, let's add `comment` to each translation, and maybe tweak these slightly to be a little more descriptive of what they search for for visually impaired users.

estrattonbailey (approved):
LGTM!

estrattonbailey (approved):

surfdude29:
Thank you for the comments for translators, those are much appreciated 🙏

PR #10262, opened 2026-04-15
Reveal chat timestamp on tap

ds-boyce:
Tapping a message without a timestamp reveals that message’s timestamp.

This was added to the `ContextMenu` component so that it wouldn’t interfere with the existing double-tap and tap-and-hold gestures.

https://github.com/user-attachments/assets/a5c36185-37c9-4e8c-b60d-1573f481dc68


mozzius:
Code looks good. Will ask a design question in the slack

mozzius (approved):

PR #10266, opened 2026-04-16
Support group chats across other surfaces

mozzius:
I made a utility to parse a raw `ConvoView` into something more ergonomic that supports proper type discrimination so that it's easier to work with. No more `bsky.dangerousIsType` everywhere! It also nicely extracts either the convo owner or the other participant as the "primary member", which let me delete a bunch of duplicated code

I then used it in:
- ChatListItem
- RequestListItem
- ShareMenu
- SearchablePeopleList

### DIFF
--- a/src/components/dms/util.ts
+++ b/src/components/dms/util.ts
@@ -1,7 +1,7 @@
-import {type ChatBskyConvoDefs} from '@atproto/api'
+import {type $Typed, ChatBskyActorDefs, ChatBskyConvoDefs} from '@atproto/api'
 
 import {EMOJI_REACTION_LIMIT} from '#/lib/constants'
-import type * as bsky from '#/types/bsky'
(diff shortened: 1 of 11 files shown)


src/components/dms/util.ts
```diff
@@ -54,3 +55,99 @@ export function hasReachedReactionLimit(
...
+    if (!otherUser) {
+      throw new Error('No other user found in direct convo')
+    }
+
+    return {
+      view: convoView,
+      kind: 'direct',
+      details: convoView.kind,
+      primaryMember: otherUser as DirectConvoMember,
+      members: convoView.members as Array<DirectConvoMember>,
```
mozzius, line 147: just taking this at face value, but since this adds no additional information to the ProfileViewBasic it shouldnt matter

ds-boyce (approved):
Love it.

PR #10270, opened 2026-04-16
Lazily measure lightbox thumbnails

mozzius:
## Summary

- Only measure the tapped image's thumbnail when opening the lightbox, instead of all images. This fixes the gallery carousel case where offscreen images (recycled by FlatList) fail to measure and block the open animation entirely.
- Measurement logic moved into `openLightbox` so callsites just pass refs — no more `runOnUI`/`measure`/`runOnJS` boilerplate at each callsite.
- A `thumbRects` SharedValue map caches measured rects for the animation worklet. Seeded on open, updated on close (when the user swiped to a different image and presses X).

You can see at the end of the video the final image is offscreen, so there's no animation. Bit janky, not ideal. I think the ideal behaviour is that we scroll the underlying scrollview along with the pager

https://github.com/user-attachments/assets/ca96d25b-d295-430d-82c9-f740b53926db

## Test plan

- [ ] Single image post: open/close animation works
- [ ] Multi-image gallery (4+ images): open animation works on tapped image (previously broken — snapped open)
- [ ] Swipe to different image in lightbox, close via X: animated close
- [ ] Swipe-to-dismiss: fly-away works as before
- [ ] Profile avatar lightbox: open/close animation works
- [ ] Profile banner lightbox: open/close animation works
- [ ] No Reanimated measurement warnings on gallery open

### DIFF
--- a/src/view/com/profile/ProfileSubpageHeader.tsx
+++ b/src/view/com/profile/ProfileSubpageHeader.tsx
@@ -1,12 +1,6 @@
 import {useCallback} from 'react'
 import {Pressable, View} from 'react-native'
-import Animated, {
-  measure,
-  type MeasuredDimensions,
-  runOnJS,
-  runOnUI,
-  useAnimatedRef,
-} from 'react-native-reanimated'
+import Animated, {useAnimatedRef} from 'react-native-reanimated'
 import {type AppBskyGraphDefs} from '@atproto/api'
 import {msg} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react'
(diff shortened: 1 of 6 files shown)


src/view/com/profile/ProfileSubpageHeader.tsx
```diff
@@ -60,14 +54,17 @@ export function ProfileSubpageHeader({
   const canGoBack = navigation.canGoBack()
   const aviRef = useAnimatedRef()
 
-  const _openLightbox = useCallback(
-    (uri: string, thumbRect: MeasuredDimensions | null) => {
+  const onPressAvi = useCallback(() => {
+    if (
+      avatar // TODO && !(view.moderation.avatar.blur && view.moderation.avatar.noOverride)
```
vineyardbovines, line 59: needed?

vineyardbovines (approved):

src/view/com/profile/ProfileSubpageHeader.tsx
mozzius: No idea what it is, but it's pre-existing

PR #10272, opened 2026-04-16
Animate border radius in lightbox transition

mozzius:
## Summary

- Adds a border radius animation to the lightbox shared element transition. The source thumbnail's border radius (e.g. `rounded_md` = 12px for post images) interpolates to 0 as the image opens to full screen, and back on close.
- The radius is compensated for the crop frame scale and overall scale so it visually matches the source thumbnail at `progress=0`.
- Only affects `type: 'image'` — avatar and banner lightboxes don't set `thumbBorderRadius` so they're unaffected.
- Extracts `LightboxTransforms` type to reduce duplication across the three ImageItem platform files.

https://github.com/user-attachments/assets/ab94ca51-da66-4c3b-b458-d9bdbe35da62

## Test plan

- [ ] Single image: border radius animates from rounded to square on open, back on close
- [ ] Multi-image grid: same behavior
- [ ] Profile avatar lightbox: no border radius animation (circle clip unchanged)
- [ ] Profile banner lightbox: no border radius animation

### DIFF
--- a/src/components/Post/Embed/ImageEmbed.tsx
+++ b/src/components/Post/Embed/ImageEmbed.tsx
@@ -4,7 +4,7 @@ import {Image} from 'expo-image'
 
 import {useLightboxControls} from '#/state/lightbox'
 import {type Dimensions} from '#/view/com/lightbox/ImageViewing/@types'
-import {atoms as a} from '#/alf'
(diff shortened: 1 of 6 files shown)


vineyardbovines (approved):
sweeet, yes

PR #10274, opened 2026-04-16
Don't toggle the date divider when tapping a reaction

ds-boyce:
This was a little distracting and unexpected.


mozzius (approved):
Sure makes sense

PR #10275, opened 2026-04-16
Add ability to edit group clip clop name

ds-boyce:
Owners can now edit group clip clop names after creation.


mozzius (approved):
:spongebob_money:

we really need to get those system messages going

PR #10278, opened 2026-04-17
Upstream merge

Ashex:


Ashex:
Woops, that went the wrong way

PR #10279, opened 2026-04-17
Delete "non-standard" styles from `pal`

mozzius:
We had a bunch of random one-off colors in the old theme that were either unused or used in one place. Let's inline all these one-off styles to make moving off of `usePalette` easier. The only difficult one was `ViewFullThread.tsx` which I just ALF'd

`pal.borderDark` and `pal.icon` were also unused, so deleted those too


src/view/com/posts/ViewFullThread.tsx
```diff
@@ -1,75 +1,70 @@
...
+              <Circle cx="2" cy="34" r="1.5" fill={t.palette.contrast_200} />
+            </Svg>
+          </View>
+          <Text
+            style={[
+              a.text_md,
+              {color: t.palette.primary_500, paddingTop: 18, paddingBottom: 4},
+            ]}>
+            {/* HACKFIX: Trans isn't working after SDK 53 upgrade -sfn */}
+            {l`View full thread`}
```
ds-boyce, line 64: Is this still true?

ds-boyce (approved):
So much beautiful red.

src/view/com/posts/ViewFullThread.tsx
mozzius: Good question! I cannot remember why this was...

PR #10283, opened 2026-04-17
Fix crash when deleting account from NoAccessScreen

ds-boyce:
> Couldn't find a navigation object. Is your component inside NavigationContainer

This was because `NoAccessScreen` is not a child of `RoutesContainer`, since the latter contains navigation elements we necessarily want to block for AA.

This PR switches to `SimpleInlineLinkText` instead.

### DIFF
--- a/src/screens/Settings/components/DeleteAccountDialog.tsx
+++ b/src/screens/Settings/components/DeleteAccountDialog.tsx
@@ -6,11 +6,10 @@ import {Trans} from '@lingui/react/macro'
 
 import {DM_SERVICE_HEADERS} from '#/lib/constants'
 import {useCleanError} from '#/lib/hooks/useCleanError'
-import {useOpenLink} from '#/lib/hooks/useOpenLink'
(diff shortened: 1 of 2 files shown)


src/screens/Settings/components/DeleteAccountDialog.tsx
```diff
@@ -262,24 +260,24 @@ function DeleteAccountDialogInner({
             {emailSentCount > 1 ? (
               <Trans>
                 Email sent!{' '}
-                <InlineLinkText
-                  label={_(msg`Resend`)}
-                  {...createStaticClick(() => {
+                <Span
```
estrattonbailey: Oh! Of course that's it. We actually have a `SimpleInlineLinkText` for this exact use case, should be a drop-in replacement for `InlineLinkText`

estrattonbailey (approved):
LGTM! Let's just use the existing component, should be a 1:1 swap

PR #10286, opened 2026-04-17
Squash carousel in chat messages, fix rounding

estrattonbailey:
The quote embed context was unused, probs something I was gonna do something with but never did


estrattonbailey:
This one should be safe, tested on web and ios, gonna merge

PR #10287, opened 2026-04-17
Add ability to remove members from a group clip clop

ds-boyce:
Members can now be removed from existing group clip clops.

https://github.com/user-attachments/assets/945df519-6ff5-45f9-9caf-5c77513521b8

### DIFF
--- a/src/screens/Messages/ConversationSettings.tsx
+++ b/src/screens/Messages/ConversationSettings.tsx
@@ -433,14 +433,14 @@ function Member({
         break
     }
   } else {
-    statusBadge = (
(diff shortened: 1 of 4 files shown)


src/screens/Messages/ConversationSettings.tsx
```diff
@@ -502,10 +511,12 @@ function StatusButton({
 }
 
 function MemberMenu({
+  convo,
   profile,
   type,
   isOwner,
 }: {
+  convo: ConvoWithDetails | null
```
mozzius: Would be cleaner if this wasn't nullable, if possible

PR #10290, opened 2026-04-17
Use convo view instead of convo state for driving the UI

ds-boyce:
This is a refactor for performance reasons; no changes to existing behavior.

### DIFF
--- a/src/screens/Messages/Conversation.tsx
+++ b/src/screens/Messages/Conversation.tsx
@@ -3,6 +3,7 @@ import {type LayoutChangeEvent, View} from 'react-native'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
 import {
   type AppBskyActorDefs,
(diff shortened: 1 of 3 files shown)


src/screens/Messages/Conversation.tsx
```diff
@@ -166,9 +171,13 @@ function Inner() {
       {!readyToShow && (
         <View style={IS_LIQUID_GLASS && {paddingTop: topInset}}>
           {moderation ? (
-            <MessagesListHeader profile={recipient} moderation={moderation} />
+            <MessagesListHeader
+              convo={convoState.convo}
+              profile={recipient}
```
mozzius, line 182: I wonder if we should just get this from ConvoWithDetails maybe?

mozzius (approved):
Muuuch cleaner; haven't tested though. Up to you if you want to merge it, I can return to it on Monday if you prefer

PR #10291, opened 2026-04-17
Fix missing recId values for suggested user events

ds-boyce:
Confirmed locally that all `suggestedUser:` events now include a `recId` value: Explore, Discover, profile header, profile interstitial, onboarding, and progress guide (“Follow 10 people to get started”).

### DIFF
--- a/src/state/queries/trending/useGetSuggestedOnboardingUsersQuery.ts
+++ b/src/state/queries/trending/useGetSuggestedOnboardingUsersQuery.ts
@@ -5,6 +5,7 @@ import {
 import {type QueryClient, useQuery} from '@tanstack/react-query'
 
 import {createBskyTopicsHeader} from '#/lib/api/feed/utils'
+import {logger} from '#/logger'
 import {getContentLanguages} from '#/state/preferences/languages'
 import {STALE} from '#/state/queries'
 import {usePreferencesQuery} from '#/state/queries/preferences'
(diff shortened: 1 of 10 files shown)


src/state/queries/trending/useGetSuggestedOnboardingUsersQuery.ts
```diff
@@ -54,7 +54,7 @@ export function useGetSuggestedOnboardingUsersQuery(props: QueryProps) {
         },
       )
 
-      return {...data, recId: data.recIdStr}
+      return {...data, recId: data.recIdStr ?? data.recId}
```
estrattonbailey: This is a single endpoint and should use one or the other, right? Do we need to clarify in the atproto repo which is which?

iwsmith: yeah let's just user `recIdStr` going forward. `recId` is deprecated. LMK if its missing from any endpoints and I can update that on my end (probably).

ds-boyce: It appears to be missing from this endpoint, which is why I added the fallback.

(This is also one of the cases where `recId` was already a string.)

estrattonbailey: Looks like we have both and [both are strings](https://github.com/bluesky-social/atproto/pull/4683/changes). So we don't need this nullish fallback then 👍 I'll hit that quick I've got the code loaded already

estrattonbailey (approved):

PR #10293, opened 2026-04-18
[Chat] Remove convoState dependency from header

mozzius:
Makes the source of truth for the header to be React Query instead of the convo state. My plan is to make the `Convo` class entirely devoted to powering the list of messages - currently the data dependency is very convoluted and weird. This is part 1 of that.


ds-boyce (approved):
Clean.

PR #10300, opened 2026-04-19
Add iOS peek long-press context menu for image embeds

### LINKED ISSUE #7274: Support for Peek and Pop interactions on iOS/iPadOS
### Describe the Feature

A useful feature of iOS is the [Peek and Pop](https://developer.apple.com/documentation/uikit/implementing-peek-and-pop) feature that can be activated with a "Haptic touch" (i.e. a long press). This interaction facilitates contextual menu style actions for some elements and even entire content previews for other elements.

The most common use case, for example, is Haptic touching a URL which brings up a mini preview of the website behind the URL. Swiping up on the preview at this point can provide options like copying the previewed URL. Tapping on the preview brings a full WebView of the linked URL.

This is really useful when browsing links in posts as it lets users take a peek at what's behind the link before deciding whether they want to interact with it more fully. And there are other benefits of peek and pop too, like peeking on a tall image that has been cropped to fit on the feed to see the entire image.

It does seem like there's some level of React Native support for this feature via this library: https://github.com/osdnk/peek-and-pop

### Attachments

https://github.com/user-attachments/assets/2c199e8d-1199-4334-b608-24a93c1bc3f2

In the first example, I long pressed on the image in the post which brings up a focused preview of the image as well as some contextual menu options.

In the second example, I long pressed on the post itself which brings up a focused preview of the post (including the replies to the post) as well as some contextual menu options.

In the third example, I long pressed the URL in the post's body which brings up a focused preview of the webpage behind the URL as well as some contextual menu options.

In the fourth and final example, I long pressed on the poster's profile photo which brings up a focused preview of the poster's account as well as some contextual menu options.

So as we can see, implementing peek and pop behavior can provide some very useful ways for users to navigate through the app quickly and utilize some quick actions from contextual menus.

### Describe Alternatives

_No response_

### Additional Context

iOS peek and pop developer documentation: https://developer.apple.com/documentation/uikit/implementing-peek-and-pop

React Native peek and pop library: https://github.com/osdnk/peek-and-pop

mozzius:
Long-pressing an image on iOS now shows a native context menu with a full-size preview (peek), along with Save and Share actions. Tapping the preview opens the lightbox.

This works across all image surfaces: single images, gallery carousels, image grids, and the small thumbnails in notification items.

https://github.com/user-attachments/assets/ca546ed4-9f7e-4e59-ba51-94b535b6f2d9

### Implementation

New native Expo module (`expo-bluesky-context-menu`) that wraps `UIContextMenuInteraction`. The preview controller shares SDWebImage's cache with expo-image, so the fullsize is usually instant if it's been prefetched — otherwise it shows the thumbnail as a placeholder.

On the JS side, a `PeekMenu` component re-exports the module on iOS and noops on other platforms. `ImageContextMenu` wraps each image embed with the peek interaction, and `MediaPreview` gains a `peekable` prop for smaller thumbnail surfaces like notifications.

One fun bug: React Native's layout engine sometimes produces sub-pixel widths (e.g. `150.00001525878906` instead of `150.0`), which caused iOS's context menu dismiss animation to glitch. Fixed by snapping `bounds` to pixel boundaries in the native view.

### Test plan

- [ ] **Single image post:** long-press → preview appears with correct aspect ratio → dismiss animates back cleanly → tap preview opens lightbox → close lightbox animates back to thumbnail
- [ ] **Gallery carousel:** long-press each image in a multi-image carousel → preview + menu work for all positions (first, middle, last)
- [ ] **Notification thumbnails:** long-press a small image thumbnail in notifications → peek works, Save/Share actions function
- [ ] **Save image:** tap "Save image" in the menu → image saved to photo library
- [ ] **Share:** tap "Share" → share sheet appears with the fullsize image
- [ ] **Portrait, landscape, square images:** preview sizes correctly for all aspect ratios
- [ ] **Quoted posts with media:** long-press images inside quote embeds
- [ ] **Android/Web:** no regressions — PeekMenu noops, images behave as before
- [ ] **Normal tap still works:** tapping (not long-pressing) an image still opens the lightbox normally
- [ ] **Lightbox close animation:** after opening via peek commit, closing the lightbox animates back to the thumbnail (not the preview position)


modules/expo-bluesky-context-menu/ios/SVGPathParser.swift
```diff
@@ -0,0 +1,319 @@
...
+  mutating func peek() -> Character? {
+    skipSeparators()
+    return index < chars.count ? chars[index] : nil
+  }
+
+  mutating func consume() {
+    if index < chars.count { index += 1 }
+  }
+
+  mutating func readNumber() -> CGFloat {
```
ds84182: I normally don't do this but seeing as you used Claude to generate this code I feel an obligation to comment:

This parse function is genuinely awful.

* You have `peek` and `consume` available. Swift has guards. Use them.
    * A character `peek` function in a parser should not implicitly consume separators (whitespace).
* There is no need to shove everything into a singular while loop. Checking if you're at the first iteration inside the loop is a code smell.

Generally the flow of this function should be:

* peek, if the char is the sign then consume it.
* while loop. while the char is a digit, consume it.
* peek, if the char is a period, consume it. then while loop, while the char is a digit, consume it.
* peek, if the char is an 'E', consume it. then while loop, while the char is a digit, consume it.

I don't write Swift, but from glancing at the language documentation for a tiny bit it should roughly be:

```swift
if let c = peek(), c == "+" || c == "-" {
  consume()
}
while let c = peek(), c.isNumber {
  consume()
}
if let c = peek(), c == "." {
  consume()
  while let c = peek(), c.isNumber {
    consume()
  }
}
if let c = peek(), c == "e" || c == "E" {
  consume()
  if let c = peek(), c == "+" || c == "-" {
    consume()
  }
  while let c = peek(), c.isNumber {
    consume()
  }
}
```

Which is better than juggling two booleans and manually incrementing an index counter while bounds checking. Much simpler too.

mozzius:
@claude please review

surfdude29:
This looks great!

One small suggestion from watching the video: Whichever icon of the two is preferred, I think it would make sense to use the same one for the Save actions in the context menu and the lightbox.

mozzius:
We will be redesigning the lightbox soon. It uses the old design language/icon pack

mozzius:
@claude I've moved it out to an npm package. can you re-review? do the changes in social-app make sense?

estrattonbailey (approved):
Sick as

PR #10302, opened 2026-04-19
nit: fix perf regression in hotkeys

WamWooWam:
Fixes a pretty gnarly performance regression causing a fat re-render whenever you open the drawer or lightbox on Android/iOS.

Backported from https://tangled.org/jollywhoppers.com/witchsky.app/pulls/106/round/0

### DIFF
--- a/src/lib/hotkeys/index.native.tsx
+++ b/src/lib/hotkeys/index.native.tsx
@@ -9,8 +9,8 @@ const noop = () => {}
 export function useHotkeysContext() {
   return useMemo(
     () => ({
-      registerHotkey: noop,
-      unregisterHotkey: noop,
+      enableScope: noop,
+      disableScope: noop,
     }),
     [],
   )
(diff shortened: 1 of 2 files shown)


mozzius:
~~I don't see this happening. Looks like React Compiler is (correctly) catching it, unless I don't understand the repro steps.~~

~~Are you able to repro this in `social-app`? My guess is that RC is broken in your fork~~

No wait got it. Good catch! Thank you

src/lib/hotkeys/index.native.tsx
```diff
@@ -1,10 +1,17 @@
...
+
 export function useHotkeysContext() {
-  return {
-    enableScope: () => {},
-    disableScope: () => {},
-  }
+  return useMemo(
+    () => ({
+      registerHotkey: noop,
+      unregisterHotkey: noop,
```
mozzius:
```suggestion
      enableScope: noop,
      disableScope: noop,
```

mozzius (approved):

PR #10303, opened 2026-04-20
Hides pillarboxing on vertical videos on web

abenzer:
Hides pillarboxing in the video player for vertical videos on web, so it matches mobile. 

Before:
before

After:
after


estrattonbailey:
Superseded by #11118

PR #10308, opened 2026-04-20
fix hang on starterpack creation

vineyardbovines:
removes the chained loading dialog that was triggering sheet presentation race, causing starter pack creation to get stuck / unable to go back.

## Test plan

- [ ] Create list flow completes without loading dialog getting stuck
- [ ] Navigation to ProfileList happens immediately after save
- [ ] Members populate in the list (background fire-and-forget)
- [ ] Back arrow responsive


src/components/dialogs/lists/CreateListFromStarterPackDialog.tsx
```diff
@@ -46,95 +44,77 @@ export function CreateListFromStarterPackDialog({
...
 
   const onListCreated = (listUri: string) => {
-    loadingDialogControl.open()
-    addMembersAndNavigate(listUri)
+    const urip = new AtUri(listUri)
+    navigation.navigate('ProfileList', {
+      name: urip.hostname,
+      rkey: urip.rkey,
+    })
+    addMembersToList(listUri)
```
estrattonbailey, line 117: Ah yeah I see the need here. Seems to work fine, but I think ideally we'd invalidate the feed of posts on that list as soon as the members are finished being added?

src/components/dialogs/lists/CreateListFromStarterPackDialog.tsx
```diff
@@ -46,95 +44,77 @@ export function CreateListFromStarterPackDialog({
...
+          (res: {data: {items: unknown[]}}) => res.data.items.length > 0,
+          () =>
+            agent.app.bsky.graph.getList({
+              list: listUri,
+              limit: 1,
+            }),
+        )
+      }
 
       queryClient.invalidateQueries({queryKey: ['list-members', listUri]})
```
estrattonbailey, line 97: Yeah can we invalidate the list feed query here too? Also I think we can re-use the query key creator exported from the `list-members` file.

estrattonbailey (commented):
Nice one, I like the simplification 👍

PR #10312, opened 2026-04-20
Display system messages in chats

ds-boyce:
System messages now appear in the chat list and in message lists.

~Please note the TODO in `src/state/messages/convo/agent.ts`; it will be addressed in a follow-up.~ TODO addressed here.

https://github.com/user-attachments/assets/985e3010-eaa1-4ac1-8ecd-bf6c8dca6c8f


mozzius:
Pushed up the rest of the icons :)

src/components/dms/DateDivider.tsx
```diff
@@ -56,12 +56,11 @@ let DateDivider = ({date: dateStr}: {date: string}): React.ReactNode => {
   }
 
   return (
-    <View style={[a.w_full, a.my_sm]}>
```
mozzius, line 59: does this even need a background color?

src/components/dms/SystemMessageItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+        a.justify_center,
+        a.px_md,
+        a.mt_md,
+      ]}>
+      <Icon size="xs" style={[a.mr_2xs, t.atoms.text_contrast_medium]} />
+      <Text
+        style={[
+          a.text_xs,
+          a.text_center,
+          t.atoms.text_contrast_medium,
```
mozzius, line 37: fwiw the figma has this at `text_contrast_high`. `medium` does look a little better to me though

mozzius (approved):
Good stuff! I made some stylistic tweaks but otherwise good

src/components/dms/SystemMessageItem.tsx
```diff
@@ -0,0 +1,44 @@
+import {View} from 'react-native'
+import {useLingui} from '@lingui/react/macro'
+
+import {type ConvoItem} from '#/state/messages/convo/types'
+import {atoms as a, useTheme} from '#/alf'
+import {getSystemMessageInfo} from '#/components/dms/systemMessage'
+import {Text} from '#/components/Typography'
+
+export function SystemMessageItem({
```
mozzius, line 9: We can do this in a followup, but these need to be clickable in certain cases

Image

src/components/dms/DateDivider.tsx
ds-boyce: Nope!

src/components/dms/SystemMessageItem.tsx
ds-boyce: Ah, I used the same styles as the date divider here.

mozzius: That's also `high` in the figma too

ds-boyce: Checked with Darrin and we’re going with `text_contrast_medium`. 🙂

PR #10313, opened 2026-04-20
don't screenshot dictation placeholders as pasted images

### LINKED ISSUE #10301: iOS: Dictation generates phantom image attachments when dictation stops
### Steps to Reproduce

Title:
iOS: Dictation generates phantom image attachments when dictation stops

Body:
### Summary
When using dictation in the Bluesky iOS app, stopping dictation — either manually or automatically — can generate one or more unintended image attachments in the post composer.

### Impact
This can result in users accidentally posting phantom images. For blind and low-vision users, this is especially problematic because the phantom images may not be properly announced, which can lead to posts containing images without intended alt text.

### Environment
- App: Bluesky iOS app
- Platform: iPhone / iOS
- Occurs on: multiple iPhone models and iOS versions
- VoiceOver: issue also occurs with VoiceOver turned off

### Steps to reproduce
1. Open the Bluesky app on iOS.
2. Start creating a new post.
3. Activate the iOS dictation feature.
4. Speak some text.
5. Stop dictation manually, or wait for dictation to stop automatically.
6. Observe the post composer.

### Expected behavior
Stopping dictation should only end speech input and should not create any media attachments.

### Actual behavior
One or more unintended image attachments appear in the composer after dictation stops. In testing, up to four phantom images were generated.

### Frequency
Reproducible in repeated tests.

### Additional notes
- The issue appears to be in the iOS composer flow and not limited to accessibility mode, because it also happens when VoiceOver is disabled.
- A screen recording documenting the behavior is available.

### Reference
Public example / report:
https://bsky.app/profile/blindleben.bsky.social/post/3mjofw5k75c2f

### Attachments

_No response_

iOS

### Device Info

Various iPhone models and various iOS versions

Version 1.120.0.1449

### Additional Information

Additional notes
- The issue appears to be in the iOS composer flow and not limited to accessibility mode, because it also happens when VoiceOver is disabled.
- A screen recording documenting the behavior is available:  
  https://bsky.app/profile/blindleben.bsky.social/post/3mjofw5k75c2f

vineyardbovines:
Test Plan:

- Dictate text
- There is no screenshot in the composer

CleanShot 2026-04-20 at 16 52 03

### DIFF
--- a/patches/expo-paste-input+0.1.15.patch
+++ b/patches/expo-paste-input+0.1.15.patch
@@ -1,0 +1,136 @@
+diff --git a/node_modules/expo-paste-input/ios/ExpoPasteInputView.swift b/node_modules/expo-paste-input/ios/ExpoPasteInputView.swift
+index 2164aec4ec1d..d216db6d2927 100644
+--- a/node_modules/expo-paste-input/ios/ExpoPasteInputView.swift
++++ b/node_modules/expo-paste-input/ios/ExpoPasteInputView.swift
+@@ -511,14 +511,17 @@ class ExpoPasteInputView: ExpoView {
+     var attachmentRanges: [NSRange] = []
+     var mediaPayloads: [MediaPayload] = []
+ 
++    // Only track ranges for attachments we successfully extract a real payload
++    // from. Attachments without a payload (e.g. iOS dictation placeholders)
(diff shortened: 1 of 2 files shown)


mozzius (commented):
Wh

Works good on the sim

mozzius (approved):

PR #10319, opened 2026-04-21
Refactor minimal shell mode to refcounting

mozzius:
## Summary
- Replaces the old boolean setter (`useSetMinimalShellMode(true/false)`) with a ref-counted system. Multiple consumers can independently request minimal mode without stepping on each other.
- Screens that just showed the shell on focus (~25 files) had their boilerplate deleted — the shell is now visible by default (count=0).
- Screens that hide the shell (Conversation, VideoFeed, StarterPack Wizard) now use `<Layout.Screen minimalShell>`.
- Non-screen components (LoggedOut, FindContactsFlow) use `useEnableMinimalShellMode()`.
- Home pager directly manipulates `headerMode` sharedvalue

Net: -216 lines.

## Test plan
- [ ] Home feed: scroll down (header hides), switch tabs (header snaps back), drag pager (header snaps back)
- [ ] Messages conversation: header/footer hidden, navigate back → header/footer visible
  - [ ] Deep links hide the shell correctly (can test by going to a DM in safari in the sim, then pressing "Open" in the banner) 
- [ ] Video feed: header/footer hidden, navigate back → visible
- [ ] StarterPack wizard: same as above
- [ ] Logged out state: shell stays hidden
- [ ] General navigation between screens: shell always visible

### DIFF
--- a/src/state/shell/minimal-mode.tsx
+++ b/src/state/shell/minimal-mode.tsx
@@ -104,41 +104,6 @@ export function useMinimalShellModeSetters() {
   return context
 }
 
-/**
- * Bypasses the refcount and directly animates the shell. Only for use in
- * scroll/pager callbacks where the refcount is 0 but MainScrollProvider has
- * hidden the header via direct shared-value manipulation.
- * @deprecated Prefer `useEnableMinimalShellMode` or the `minimalShell` prop.
- */
-export function useDangerouslyImperativelySetMinimalShellMode() {
-  const {headerMode, footerMode} = useMinimalShellMode()
-
-  const setMode = useCallback(
-    (v: boolean) => {
-      'worklet'
-      headerMode.set(() =>
-        withSpring(v ? 1 : 0, {
-          overshootClamping: true,
-        }),
-      )
-      footerMode.set(() =>
-        withSpring(v ? 1 : 0, {
-          overshootClamping: true,
-        }),
-      )
-    },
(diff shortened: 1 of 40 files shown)


src/state/shell/minimal-mode.tsx
```diff
@@ -65,15 +81,84 @@ export function Provider({children}: React.PropsWithChildren<{}>) {
...
+          overshootClamping: true,
+        }),
+      )
+    },
+    [headerMode, footerMode],
+  )
+
+  useFocusEffect(
+    useCallback(() => {
+      return () => setMode(false)
```
estrattonbailey: The naming here is a little unclear. When this un-focuses, it sets to `false` which should show the shell, right? Nit, but maybe `setShow` and `setShowWorklet` would be more clear?

estrattonbailey (approved):
Beautiful refactor! Really appreciate the comments, very clear 👍

src/state/shell/minimal-mode.tsx
mozzius: Yeah that tripped me up. I will change it to "show" | "hide" instead of bools

mozzius: actually I got rid of that entirely and just made `Home` directly manipulate `headerMode`, as it shouldn't be touching the bottom bar show/hide state at all

PR #10320, opened 2026-04-21
Display sender name in last message for group clip clops

ds-boyce:
Also did some cleanup for `ChatListItem`.

system


src/components/dms/getReactionInfo.ts
```diff
@@ -0,0 +1,59 @@
...
+  )
+  const target = lastMessageText ? `"${lastMessageText}"` : fallbackMessage
+
+  let message: string
+  if (isFromMe) {
+    message = i18n._(msg`You reacted ${reaction.value} to ${target}`)
+  } else if (isGroup) {
+    message = i18n._(msg`${name} reacted ${reaction.value} to ${target}`)
+  } else if (sender) {
+    const directName = sanitizeDisplayName(sender.displayName || sender.handle)
```
mozzius: can't we just use `name` here?

src/components/dms/getReactionInfo.ts
```diff
@@ -0,0 +1,59 @@
...
+}): UserReactionInfo | null {
+  if (!ChatBskyConvoDefs.isMessageAndReactionView(convo.lastReaction)) {
+    return null
+  }
+
+  const {reaction, message: reactedTo} = convo.lastReaction
+  const isFromMe = reaction.sender.did === currentAccountDid
+  const senderDid = reaction.sender.did
+  const sender = convo.members.find(m => m.did === senderDid)
+  const name = sender ? createSanitizedDisplayName(sender) : senderDid
```
mozzius: should probably fall back to something generic like `Someone` rather than the raw DID - we never want to show that in the UI

mozzius (approved):
Looking good!

mozzius (approved):

PR #10322, opened 2026-04-21
Add ability to create and modify join links

ds-boyce:
Builds on #10318. Group clip clop invite links can now be created, edited, disabled, and re-enabled. Sharing is enabled in #10323.

Posting and sharing will come in a follow-up PR.

https://github.com/user-attachments/assets/c940220e-75d3-41a4-aa7b-77b52bdf9e67


src/screens/Messages/components/InviteLinkDialog.tsx
```diff
@@ -145,68 +213,119 @@ export function InviteLinkDialog({
...
+            <View style={[a.flex_row, a.justify_between, a.gap_sm, a.mt_lg]}>
+              <StackedButton
+                label={l`Disable`}
+                icon={ChainLinkBrokenIcon}
+                color="negative_subtle"
+                style={[a.flex_1, a.rounded_full]}
+                onPress={() => {
+                  disableJoinLink()
+                }}>
+                Disable
```
surfdude29: I think the `<Trans>` tags might be missing here? And also on L290 and L302

PR #10328, opened 2026-04-22
Use latest react-textarea-autosize, force bump transitive deps to support React 19

mozzius:
The version of `react-textarea-autosize` we're on doesn't have React 19 marked as a peerDependency. I upgraded it, and had to tamper with the lockfile to make its dependencies also support React 19 (they would otherwise be stuck on an old one)

This will cause the lockfile CI to fail: this is expected

Reduces the number of peer dependency warnings we get when installing:

## Before

Screenshot 2026-04-22 at 13 03 28

## After

Screenshot 2026-04-22 at 13 03 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,7 +233,7 @@
     "react-native-webview": "^13.13.5",
     "react-remove-scroll-bar": "^2.3.8",
     "react-responsive": "^10.0.1",
-    "react-textarea-autosize": "^8.5.3",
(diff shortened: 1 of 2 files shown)


estrattonbailey:
We can ditch this one with the new composer input right?

mozzius:
Pretty sure yeah!

PR #10330, opened 2026-04-22
Refresh lightbox designs

vineyardbovines:
https://github.com/user-attachments/assets/22f05a89-ee90-4c80-b58a-dd47f7ab6a76

https://github.com/user-attachments/assets/916819fb-7d5d-4202-8404-3238ab9c5f6c

unrelated but useful- added superpower-generated stuff to gitignore. it gets really large / bloated.


src/components/Lightbox/chrome/CircleChromeButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+  label: string
+  onPress?: PressableProps['onPress']
+  testID?: string
+}
+
+const SIZE = 44
+const RADIUS = 24
+const ICON = 24
+
+export function CircleChromeButton({
```
vineyardbovines, line 26: not married to the word/naming 'chrome' here, if anyone has differing opinions

src/components/images/ImageLayoutGrid.tsx
```diff
@@ -3,9 +3,9 @@ import {type StyleProp, View, type ViewStyle} from 'react-native'
 import {type AnimatedRef, useAnimatedRef} from 'react-native-reanimated'
 import {type AppBskyEmbedImages} from '@atproto/api'
 
-import {type Dimensions} from '#/view/com/lightbox/ImageViewing/@types'
 import {atoms as a, useBreakpoints} from '#/alf'
 import {PostEmbedViewContext} from '#/components/Post/Embed/types'
+import {type Dimensions} from '#/features/lightbox/types'
```
estrattonbailey: Nit: love the reorg here, though this seems more like a component to me since it contains a single component. I think we should instead put this in `#/components/Lightbox`?

estrattonbailey (approved):
Nice LGTM!

SaphireLattice:
Just got this update and I'm really unsure about the UX of moving the download and share buttons into a hidden away menu on mobile.

The previous actions were reachable easily with a thumb, and this change feels like it's purpose built to restrict and discourage users from using these. 

If it were a different site doing this without being open sourced, I would put a tinfoil hat of "This is made to force people to share posts instead of images" 😆

PR #10333, opened 2026-04-22
Display actual number of group clip clop requests

ds-boyce:
Removed hard-coded value. Re-renders when new invites are received, so the count is relatively fresh.

Only fetches the first 50 (the default), so if there are more than that it just displays “50+ requests”.


src/state/queries/messages/list-join-requests.ts
```diff
@@ -0,0 +1,65 @@
...
+            void queryClient.invalidateQueries({
+              queryKey: createListJoinRequestsQueryKey({convoId}),
+            })
+            return
+          }
+        }
+      },
+      {convoId},
+    )
+  }, [isEnabled, convoId, messagesBus, queryClient])
```
mozzius, line 49: Ahh very nice, smart

mozzius (approved):

PR #10335, opened 2026-04-22
[Chat] Convo refactor - Move convo/member data fetching into React Query

mozzius:
## Summary

- Moves conversation and member list fetching out of the `Convo` class and into React Query (`useConvoQuery` + `useListConvoMembersQuery`), with the `ConvoProvider` syncing data in via `setConvoData`
- `Convo` stores `ConvoWithDetails` instead of raw `ConvoView`, so consumers get parsed group/direct info directly (e.g. `convoState.convo.kind === 'group'`)
- Removes `sender`, `recipients`, `isGroup()`, `getGroupInfo()`, `getPrimaryMember()` from `ConvoState` — consumers use `convoState.convo.primaryMember`, `convoState.convo.members`, etc.
- `relatedProfiles` map on items accumulates profiles from members, message history, and reactions — handles people who've left the chat
- `parseConvoView` accepts optional full member list from `getConvoMembers`, hardening against truncated `convo.members`

## Test plan

- [ ] Open a DM conversation — messages load, header renders
- [ ] Open a group chat — group info panel shows correctly
- [ ] Scroll back in message history — profiles from former members resolve
- [ ] Background and resume the app — convo refreshes via React Query
- [ ] System messages display member names (or "Someone" fallback)
- [ ] Send a message, verify pending/sent flow works
- [ ] Reactions show correct sender profiles

### DIFF
--- a/src/state/messages/convo/agent.ts
+++ b/src/state/messages/convo/agent.ts
@@ -1,6 +1,6 @@
 import {
   type AtpAgent,
-  ChatBskyActorDefs,
+  type ChatBskyActorDefs,
   ChatBskyConvoDefs,
   type ChatBskyConvoGetLog,
   type ChatBskyConvoSendMessage,
@@ -19,6 +19,7 @@ import {Logger} from '#/logger'
 import {
   ACTIVE_POLL_INTERVAL,
   BACKGROUND_POLL_INTERVAL,
+  GET_MESSAGE_HISTORY_LIMIT,
   INACTIVE_TIMEOUT,
   NETWORK_FAILURE_STATUSES,
 } from '#/state/messages/convo/const'
@@ -26,7 +27,6 @@ import {
   type ConvoDispatch,
   ConvoDispatchEvent,
   type ConvoError,
-  ConvoErrorCode,
   type ConvoEvent,
   type ConvoItem,
   ConvoItemError,
@@ -36,8 +36,8 @@ import {
 } from '#/state/messages/convo/types'
 import {type MessagesEventBus} from '#/state/messages/events/agent'
 import {type MessagesEventBusError} from '#/state/messages/events/types'
-import {IS_NATIVE} from '#/env'
-import * as bsky from '#/types/bsky'
+import {type ConvoWithDetails, parseConvoView} from '#/components/dms/util'
+import type * as bsky from '#/types/bsky'
 
 const logger = Logger.create(Logger.Context.ConversationAgent)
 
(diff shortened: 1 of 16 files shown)


src/state/queries/messages/list-convo-members.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const RQKEY_SEGMENT = 'members'
+export const RQKEY = (convoId: string) => [
+  GET_CONVOS_KEY,
+  convoId,
+  RQKEY_SEGMENT,
+]
+
+// group chat size is 50, so should fetch the whole list in one go
+const LIMIT = 50
```
ds-boyce, line 16: We should export this and use it in `MessagesConversationSettingsScreen` as well.

src/state/messages/convo/types.ts
```diff
@@ -6,13 +6,16 @@ import {
...
 
 export type ConvoParams = {
   convoId: string
   agent: BskyAgent
   events: MessagesEventBus
-  placeholderData?: {
-    convo: ChatBskyConvoDefs.ConvoView
+  initialData?: {
+    convo?: ChatBskyConvoDefs.ConvoView
+    members?: ChatBskyActorDefs.ProfileViewBasic[]
```
ds-boyce, line 18: Good naming change here.

src/state/messages/convo/agent.ts
```diff
@@ -215,19 +199,14 @@ export class Convo {
           markConvoAccepted: this.markConvoAccepted,
           addReaction: this.addReaction,
           removeReaction: this.removeReaction,
-          isGroup: this.isGroup,
-          getGroupInfo: this.getGroupInfo,
-          getPrimaryMember: this.getPrimaryMember,
-        }
+        } as ConvoState
```
ds-boyce, line 202: Nit: The cast here may not hold true if we ever add a new status, but i can’t think of what one would be so it’s probably fine. The alternative is kind of redundant.

src/state/messages/convo/index.tsx
```diff
@@ -72,76 +74,86 @@ export function ConvoProvider({
...
     })
   })
   const service = useSyncExternalStore(convo.subscribe, convo.getSnapshot)
   const {mutate: markAsRead} = useMarkAsReadMutation()
 
+  useEffect(() => {
+    if (convoData && membersData) {
+      convo.setConvoData(convoData, membersData)
+    }
+  }, [convo, convoData, membersData])
```
ds-boyce, line 105: I think this effect is going to fire on every refetch, regardless of whether the data changed. Is there an additional check we can make here to guard it?

ds-boyce (commented):
Good stuff so far! Yay for the separation of concerns between React Query and the state machine.

PR #10340, opened 2026-04-23
Only lock viewport zoom while text inputs are focused

mozzius:
Revert #10233 and instead toggle maximum-scale=1 on iOS Safari during focusin/focusout for text inputs. This stops the auto-zoom when focusing an input with font-size < 16px while preserving pinch-to-zoom for accessibility at all other times.

**Edit: applied only to the DMs screen**

https://github.com/user-attachments/assets/768a280d-c3c7-4699-87cd-ba321aae5d54

## Test plan

- [ ] Confirm iOS safari doesn't zooms in when focusing text inputs
- [ ] Confirm you can still pinch to zoom at other times

### DIFF
--- a/bskyweb/templates/base.html
+++ b/bskyweb/templates/base.html
@@ -2,7 +2,7 @@
 <html>
 <head>
   <meta charset="UTF-8">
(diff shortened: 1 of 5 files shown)


vineyardbovines (approved):
oooo nice

PR #10349, opened 2026-04-23
Add ability to lock and unlock group clip clops

ds-boyce:
Also fixes the unlock icon, which had an invalid `path`.

https://github.com/user-attachments/assets/8b0e2dac-4317-4b11-9d82-66430821e8cf


mozzius (approved):
Looking good!

PR #10353, opened 2026-04-24
[Chat] Remove sender/recipeient state from Convo agent

mozzius:
## Stacked on #10352 

Turns out we didn't really use this anywhere, and it's very boilerplate-y to deal with - not to mention completely superceded by `ConvoWithDetails`

PR #10354, opened 2026-04-24
[Chat] Keep track of `relatedProfiles` in Convo agent

mozzius:
## Stacked on #10353 

Any time the convo sees a profile view, it will dump it into `relatedProfiles`, which it uses when it needs a DID->profile

### DIFF
--- a/src/components/dms/ActionsWrapper.tsx
+++ b/src/components/dms/ActionsWrapper.tsx
@@ -4,23 +4,29 @@ import {useLingui} from '@lingui/react/macro'
 
 import {atoms as a} from '#/alf'
 import {MessageContextMenu} from '#/components/dms/MessageContextMenu'
+import type * as bsky from '#/types/bsky'
 
 export function ActionsWrapper({
   message,
   isFromSelf,
+  senderProfile,
   children,
   onTap,
 }: {
   message: ChatBskyConvoDefs.MessageView
   hasReactions?: boolean
   isFromSelf: boolean
+  senderProfile?: bsky.profile.AnyProfileView
   children: React.ReactNode
   onTap?: () => void
 }) {
   const {t: l} = useLingui()
 
   return (
-    <MessageContextMenu message={message} onTap={onTap}>
+    <MessageContextMenu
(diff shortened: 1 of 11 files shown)


src/components/dms/ActionsWrapper.tsx
```diff
@@ -4,23 +4,29 @@ import {useLingui} from '@lingui/react/macro'
...
   message,
   isFromSelf,
+  senderProfile,
   children,
   onTap,
 }: {
   message: ChatBskyConvoDefs.MessageView
   hasReactions?: boolean
   isFromSelf: boolean
+  senderProfile?: bsky.profile.AnyProfileView
```
estrattonbailey, line 19: Is this actually AnyProfileView or will it just be a chat view?

src/state/messages/convo/agent.ts
```diff
@@ -1168,6 +1172,44 @@ export class Convo {
...
   }
 
+  private getRelatedProfilesForItem(
+    m:
+      | ChatBskyConvoDefs.MessageView
+      | ChatBskyConvoDefs.DeletedMessageView
+      | ChatBskyConvoDefs.SystemMessageView,
+  ): bsky.profile.AnyProfileView[] {
+    const seen = new Set<string>()
+    const profiles: bsky.profile.AnyProfileView[] = []
```
estrattonbailey: Instead of building new arrays for every message, we might as well just attach the reference to `this.relatedProfiles` to each one since it's already a map for quick lookups. Save some memory that way

estrattonbailey (approved):
Memory thing could do imo, lgtm otherwise!

src/components/dms/ActionsWrapper.tsx
mozzius: Hmm, I suppose a chatview

src/state/messages/convo/agent.ts
mozzius: Right, sure, it's just passing by reference so no harm to pass the big old Map around

PR #10357, opened 2026-04-24
[Chat] Fetch full member list

mozzius:
- Fetches full member list
- Mutates itself when the message comes in via messagebus

Does not integrate with convo agent yet

### DIFF
--- a/src/state/queries/messages/list-convo-members.ts
+++ b/src/state/queries/messages/list-convo-members.ts
@@ -1,0 +1,48 @@
+import {type ChatBskyActorDefs} from '@atproto/api'
+import {useQuery} from '@tanstack/react-query'
+
+import {DM_SERVICE_HEADERS} from '#/lib/constants'
+import {STALE} from '#/state/queries'
+import {useAgent} from '#/state/session'
+import {RQKEY_ROOT as GET_CONVOS_KEY} from './conversation'
+
+export const RQKEY_SEGMENT = 'members'
+export const RQKEY = (convoId: string) => [
+  GET_CONVOS_KEY,
+  convoId,
+  RQKEY_SEGMENT,
+]
+
+// group chat size is 50, so should fetch the whole list in one go
+const LIMIT = 50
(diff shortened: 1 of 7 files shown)


src/state/queries/messages/list-convo-members.ts
```diff
@@ -0,0 +1,106 @@
...
+import {useAgent} from '#/state/session'
+import * as bsky from '#/types/bsky'
+import {RQKEY_ROOT as GET_CONVOS_KEY} from './conversation'
+
+export const RQKEY_SEGMENT = 'members'
+export const RQKEY = (convoId: string) => [
+  GET_CONVOS_KEY,
+  convoId,
+  RQKEY_SEGMENT,
+]
```
estrattonbailey: Really need to add some `do not`s to the claude file so it follows the new query key pattern, nbd rn tho

src/state/queries/messages/list-convo-members.ts
```diff
@@ -0,0 +1,106 @@
...
+}: {
+  convoId: string
+  placeholderData?: ChatBskyActorDefs.ProfileViewBasic[]
+}) {
+  const agent = useAgent()
+  const queryClient = useQueryClient()
+  const messagesBus = useMessagesEventBus()
+
+  useEffect(() => {
+    const unsub = messagesBus.on(
```
estrattonbailey, line 31: This is a neat workaround, good thinking

src/state/queries/messages/list-convo-members.ts
mozzius: This was handwritten :/

estrattonbailey (approved):

src/state/queries/messages/list-convo-members.ts
mozzius: You're right tho

estrattonbailey: Lmao np np

PR #10366, opened 2026-04-24
Optimistically remove members from chat

ds-boyce:
Also disables the “Add members” button when the mutation is pending.


mozzius (approved):
works good

PR #10373, opened 2026-04-25
[Chat] Fix assumption that every chat has an owner

mozzius:
Previously, there was an implicit assumption every chat has an owner. This is not the case if the owner leaves and puts it in the perma-locked state

This makes `convo.primaryMember` optional, and then goes and fixes the place that relied on it being always set

Also excludes locked chats from incrementing the badge count

### DIFF
--- a/src/screens/Messages/components/ChatListItem.tsx
+++ b/src/screens/Messages/components/ChatListItem.tsx
@@ -124,7 +124,7 @@ function DirectChatItem({
 
   return (
     <BaseChatItem
-      convo={convo.view}
+      convo={convo}
       avatar={
         <PreviewableUserAvatar
           profile={profile}
@@ -183,7 +183,7 @@ function GroupChatItem({
 
   return (
     <BaseChatItem
-      convo={convo.view}
+      convo={convo}
       avatar={<AvatarBubbles profiles={convo.members} size={52} />}
       title={chatName}
       accessibilityHint={l`Go to the group chat named "${chatName}"`}
@@ -213,7 +213,7 @@ function BaseChatItem({
   postAlerts,
   children,
 }: {
-  convo: ChatBskyConvoDefs.ConvoView
+  convo: ConvoWithDetails
   avatar: React.ReactNode
   title: string
   subtitle?: string
@@ -238,7 +238,13 @@ function BaseChatItem({
 
   const playHaptic = useHaptics()
   const queryClient = useQueryClient()
-  const hasUnread = convo.unreadCount > 0 && !isDeletedAccount
+  const hasUnread =
+    convo.view.unreadCount > 0 &&
+    !isDeletedAccount &&
+    !(
+      convo.kind === 'group' &&
+      convo.details.lockStatus === 'locked-permanently'
+    )
 
   const blockInfo = useMemo(() => {
     if (!primaryProfileModeration) return {listBlocks: [], userBlock: undefined}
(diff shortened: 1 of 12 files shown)


src/screens/Messages/components/ChatListItem.tsx
```diff
@@ -233,9 +238,16 @@ function BaseChatItem({
 
   const playHaptic = useHaptics()
   const queryClient = useQueryClient()
-  const hasUnread = convo.unreadCount > 0 && !isDeletedAccount
+  const hasUnread =
+    convo.view.unreadCount > 0 &&
+    !isDeletedAccount &&
+    !(
+      convo.kind === 'group' &&
+      convo.details.lockStatus === 'locked-permanently'
```
estrattonbailey, line 246: Why was this occurring in the first place? Can we not update unread state of a locked convo? Maybe we should also allow that up to the last rev

src/screens/Messages/components/ChatListItem.tsx
```diff
@@ -550,13 +566,14 @@ function BaseChatItem({
 
           <ChatListItemPortal.Outlet />
 
-          {showMenu && (
+          {/* TODO: Allow showing menu for groups where the owner has left! */}
```
estrattonbailey, line 569: Note: let's track this in Linear too if you haven't already

src/screens/Messages/components/MessagesListInfoPanel.tsx
```diff
@@ -139,18 +144,24 @@ export function MessagesListInfoPanel({
...
       <Dialog.Outer
         control={addMembersControl}
         testID="addChatMembersDialog"
         nativeOptions={{fullHeight: true}}>
         <Dialog.Handle />
         <AddMembersFlow
-          members={members.map(profile => profile.did)}
+          members={(fullMemberList ?? convo.members).map(
+            profile => profile.did,
+          )}
```
estrattonbailey: Should we just build the query into this flow? Or could we replace `members` above with `fullMemberList`?

src/state/queries/messages/list-conversations.tsx
```diff
@@ -408,6 +409,8 @@ export function useUnreadMessageCount() {
     numUnread?: string
     hasNew: boolean
   }>(() => {
+    console.log('accepted', accepted)
+    console.log('request', request)
```
estrattonbailey: 😛

estrattonbailey (commented):
Generally looks good, have not run yet

src/screens/Messages/components/ChatListItem.tsx
mozzius: Yeah actually I'm going to remove this

mozzius: Actually no, so what this is is just the display of the unread dot on a chat list item. Imo when a chat is permanently ended, we don't need a notification dot

src/screens/Messages/components/ChatListItem.tsx
mozzius: https://linear.app/blueskyweb/issue/APP-2079/make-convomenu-aware-of-group-chats

src/screens/Messages/components/MessagesListInfoPanel.tsx
mozzius: imo let's build this into the flow

mozzius: Made a ticket

src/screens/Messages/components/MessagesListInfoPanel.tsx
```diff
@@ -139,11 +141,15 @@ export function MessagesListInfoPanel({
           </View>
         ) : null}
       </View>
-      <InviteLinkDialog
-        isOwner={isOwner}
-        convo={convo}
-        control={inviteLinkControl}
-      />
+      {convo.primaryMember && moderationOpts && (
```
ds-boyce, line 144: Could this get out of sync with `isJoinLinkEnabled`? Just wondering if there’s a case where we show the button but the dialog isn’t rendered.

mozzius: Hmm I don't think so. If there's no primary member, it means it's in a locked permanently state

ds-boyce (approved):

PR #10376, opened 2026-04-26
[Search] Debounce autocomplete network query

haileyok:
## Human Note

Context here is I keep running into a problem on web where I'll type some search text and press enter, but I get navigated away missing the last ~3-5 characters. Have tried a few different things but the best way seemed to be to just debounce here. Have tested locally and this does fix my problem! Might end up saving a bunch of network requests too. The suggested results are slightly slower to update as a result of this change, so that's something to consider. 

## Summary

Every keystroke in the right-side desktop search and the composer's mention/emoji autocomplete was firing a fresh `searchActorsTypeahead` request and re-rendering the full dropdown. Under that per-keystroke render pressure, fast-typing-then-Enter on the desktop search could submit a truncated `query` to the Search screen because the urgent state-update flush wasn't completing between keystrokes.

Fix: debounce the value that drives the network query inside `useAutocomplete`.

## Related Issues/Tasks

N/A

## Changes Made

- `src/components/Autocomplete/useAutocomplete/index.ts`: wrap incoming `q` in `useDebouncedValue(q, 150)` and use the debounced value for the TanStack `queryKey` and `queryFn`.
- Live `q` continues to flow through the `select` callback so the `search-${q}` fallback row and the moderation `isExactMatch` check stay in sync with what the user just typed.
- `keepPreviousData` on the query means previous results stay visible during the debounce window — no flicker.

Affects both consumers of `useAutocomplete`:
- `DesktopSearch` (right-side search box, web only)
- `Composer` (mention/emoji autocomplete)

## Confidence Level

Confidence Level: Claude

## Testing

- `yarn typecheck` clean
- `yarn lint` clean (auto-applied via lint-staged)
- Needs manual verification on web: type fast in the right-side search and confirm the URL on the Search screen matches the full typed string. Also worth a quick check on composer mention/emoji autocomplete to confirm the 150ms delay feels right and doesn't break the emoji auto-replace path in `Composer/index.tsx`.

### DIFF
--- a/src/components/Autocomplete/useAutocomplete/index.ts
+++ b/src/components/Autocomplete/useAutocomplete/index.ts
@@ -2,6 +2,7 @@ import {useCallback} from 'react'
 import {moderateProfile, type ModerationOpts} from '@atproto/api'
 import {keepPreviousData, useQuery} from '@tanstack/react-query'
 
+import {useDebouncedValue} from '#/lib/hooks/useDebouncedValue'
 import {isJustAMute, moduiContainsHideableOffense} from '#/lib/moderation'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
 import {STALE} from '#/state/queries'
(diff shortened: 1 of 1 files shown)


mozzius:
Well well well, look who it is... :)

I think #10381 should fix the root cause of the bug, but this is also a good addition! Thanks

haileyok:
please @mozzius whatever must be done to prevent me from typing "opus benchmarks" but getting "op" in the text box and getting unclothed men without labeled posts in my results 😭

mozzius:
lmfaooo

PR #10381, opened 2026-04-27
Fix search query autocomplete lagging behind

### LINKED ISSUE #10296: Search Bar Suggestions lagging behind text being entered and causing searches to get cut off
### Steps to Reproduce

1. Using the search bar in the right nav
2. Start writing something to search in the text box
3. Pause for a moment (allows the suggestions popup to appear)
4. Continue writing

As you continue, you should notice that the element just blow the text field isn't updating 1-1 with what's being entered into the search box, it takes a second to catch up

Which causes incomplete searches if you hit enter at any point before it updates to properly reflect the entered text

### Attachments

[Who_Knows_But_Shrimps_Is_Bugs.mp4](https://github.com/user-attachments/assets/876b5778-5d49-49c3-ba23-03e21112492c)

Web (Desktop)

### Device Info

_No response_

Build version: 1.120.0; Bundle info: 1bb7398 (prod); Bundle date: 26041015; Platform: web; Platform version: 0.0.0; Device ID: a52f3498-b0c9-4b0d-b2f0-4a2ac07e42f5

### Additional Information

This doesn't seem to work if you keep entering the same sentance over and over in the same session

This only occurs when the popup under the search bar appears, as if you paste something or have a shorter/quicker search and don't give it time to show, all searches seem to work entirely fine without being cut off 

Sometimes the suggestion is fast enough to not experience this, but it does seem to happen frequently enough to be a real problem 🐙

LegendEffects:
Fixes https://github.com/bluesky-social/social-app/issues/10296

This was being caused by the literal query completion being added in the react query selector, this only runs after the query fn promise is resolved.

It may be worth debouncing this query but outside of the scope of this.

### DIFF
--- a/src/components/Autocomplete/useAutocomplete/index.ts
+++ b/src/components/Autocomplete/useAutocomplete/index.ts
@@ -1,4 +1,4 @@
-import {useCallback} from 'react'
+import {useCallback, useMemo} from 'react'
 import {moderateProfile, type ModerationOpts} from '@atproto/api'
 import {keepPreviousData, useQuery} from '@tanstack/react-query'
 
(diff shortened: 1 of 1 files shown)


mozzius (approved):
great catch

PR #10383, opened 2026-04-28
Enable @typescript-eslint/no-explicit-any

ds-boyce:
Suppresses all existing violations while ensuring we don’t introduce new ones (existing can be addressed incrementally).


mozzius (approved):
💎

PR #10388, opened 2026-04-28
Group runs of system messages under a single toggle

ds-boyce:
Long runs of multiple system messages can take up a lot of real estate, e.g., multiple people being added or removed. This groups them under a toggle to reduce the noise.

https://github.com/user-attachments/assets/35e44ca7-8070-472f-8202-a204ce2aa1a9


mozzius (approved):
Works good in practice, @claude please review the logic

PR #10389, opened 2026-04-28
Show date dividers between system messages

ds-boyce:
Makes it clearer when system events occurred. Built on #10388.

timestamps


mozzius (approved):
Looks good I think - @claude do you agree?

PR #10395, opened 2026-04-29
Open invite link dialog from system messages

ds-boyce:
The following system messages now open the invite link dialog:

* Invite link created
* Invite link edited
* Invite link enabled
* Invite link disabled

Stacked on #10394.

### DIFF
--- a/src/components/dms/SystemMessageItem.tsx
+++ b/src/components/dms/SystemMessageItem.tsx
@@ -1,8 +1,9 @@
-import {View} from 'react-native'
+import {Pressable, View} from 'react-native'
 import {useLingui} from '@lingui/react/macro'
 
 import {makeProfileLink} from '#/lib/routes/links'
 import {type ConvoItem} from '#/state/messages/convo/types'
+import {useInviteLinkDialog} from '#/screens/Messages/components/InviteLinkDialogProvider'
 import {atoms as a, useTheme} from '#/alf'
 import {getSystemMessageInfo} from '#/components/dms/getSystemMessageInfo'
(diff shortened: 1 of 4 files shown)


src/components/dms/SystemMessageItem.tsx
```diff
@@ -57,8 +59,17 @@ export function SystemMessageItem({
         </Link>
       )
     case 'inviteLink':
-      // TODO Need to invoke InviteLinkDialog for this case. -dsb
-      return row
+      if (!inviteLinkControl) return row
+      return (
+        <Pressable
```
mozzius, line 64: fwiw we can use `Button` for these cases if you just leave off the `size`/`color` props

mozzius (approved):

src/components/dms/SystemMessageItem.tsx
ds-boyce: Oh sweet, TIL. In this case, I think it would result in a little more duplication since I would need to use and style `ButtonText` and `ButtonIcon` vs reusing the `row` component.

mozzius: you don't *have* to use `ButtonText`/`ButtonIcon` inside a `Button` - you can use it like a pressable

ds-boyce: Ah, also TIL. I’ll update in [d/chat-system-message-links](https://github.com/bluesky-social/social-app/tree/d/chat-system-message-links).

ds-boyce: Done!

PR #10404, opened 2026-05-04
Fix image cropping not saving on iOS

### LINKED ISSUE #10421: Image editing broken on IOS
### Steps to Reproduce

1. Create a post and add an image
2. Click the edit pencil 
3. Resize the active area by dragging the corner
4. Click the ok tick
5. Note that the image has not changed in the post composer
6. Post the post and note that the trimmed image was not posted, the full image was

### Attachments

_No response_

iOS

### Device Info

iOS 26.4.2, iPhone 17 Pro

1.121.0.1504 361578e (prod)

### Additional Information

_No response_

mozzius:
## Summary

- Bump `@bsky.app/expo-image-crop-tool` to `0.5.1` ([PR](https://github.com/bluesky-social/expo-image-crop-tool/pull/20)), which fixes the iOS cropper to write to `cachesDirectory` instead of `NSTemporaryDirectory`.
- Log non-cancellation errors from `cropImage` via `logger.error` so the next breakage in this path doesn't hide behind the silent fallback.

## Why

On iOS, cropping an image in the composer produced the original uncropped preview. The root cause was in the cropper library: it wrote the cropped JPEG into `NSTemporaryDirectory`, but `expo-file-system`'s `moveAsync` only grants write permission on `caches`, `documents`, `applicationSupport`, and app-group directories. The subsequent `moveAsync` call in `moveIfNecessary` threw `File '…/tmp/…/..' is not writable`, which our catch block in `cropImage` swallowed and returned the original `img`.

Android was unaffected because the Kotlin side already wrote into the app's `cacheDir`.

The `logger.error` addition is a secondary mitigation — `cropImage` still falls back to the original image on non-cancellation errors (to avoid breaking the compose flow), but failures now surface in Sentry instead of silently producing the reported symptom.

## Test plan

- [ ] iOS: open composer → add photo → tap edit → crop → confirm. Cropped preview replaces original.
- [ ] iOS: cancel the crop. Composer still shows the original, no error.
- [ ] Android: regression check — crop still works end-to-end.
- [ ] Web: regression check — `EditImageDialog.web.tsx` path untouched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "@braintree/sanitize-url": "^6.0.2",
     "@bsky.app/alf": "^0.1.9",
     "@bsky.app/expo-guess-language": "^0.2.8",
-    "@bsky.app/expo-image-crop-tool": "^0.5.0",
+    "@bsky.app/expo-image-crop-tool": "^0.5.1",
     "@bsky.app/expo-scroll-edge-effect": "^0.1.4",
     "@bsky.app/expo-translate-text": "^0.2.9",
     "@bsky.app/react-native-mmkv": "2.12.5",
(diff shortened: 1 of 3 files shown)


mozzius:
@claude please review

PR #10405, opened 2026-05-04
[Chat] Defer message send until after textinput resizes

mozzius:
This causes scroll bugginess when we resize + add a new message at the same time

## Before

https://github.com/user-attachments/assets/5498eb58-b37b-4ad3-9e3a-05d20452afad

## After

https://github.com/user-attachments/assets/93f810cb-80a6-46e9-9fea-0369b5f4b29e


ds-boyce (approved):
Nice.

PR #10413, opened 2026-05-04
Show message when chat is locked or ended

ds-boyce:
This component appears in place of the composer when a chat is locked or permanently locked.

locked


mozzius (approved):
looks good!

PR #10414, opened 2026-05-04
Reduce intensity of drop shadow on chat reactions

ds-boyce:
**Before**

before

**After**

after

### DIFF
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -36,6 +36,7 @@ import {unstableCacheProfileView} from '#/state/queries/unstable-profile-cache'
 import {useSession} from '#/state/session'
 import {atoms as a, native, platform, useTheme} from '#/alf'
 import {isOnlyEmoji} from '#/alf/typography'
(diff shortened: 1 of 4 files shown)


src/components/dms/MessageItem.tsx
```diff
@@ -323,12 +324,25 @@ let MessageItem = ({
...
+                  boxShadow: `0 2px 8px 0 ${alpha(t.palette.black, 0.1)}`,
+                },
+                ios: {
+                  shadowColor: t.palette.black,
+                  shadowOffset: {width: 0, height: 2},
+                  shadowOpacity: 0.1,
+                  shadowRadius: 8,
+                },
+                android: {elevation: 0.5},
+              }),
```
mozzius: maybe we should as `shadow_xs` to ALF?

ds-boyce: Good call.

ds-boyce: https://github.com/bluesky-social/toolbox/pull/34

mozzius (approved):

PR #10415, opened 2026-05-04
Add ability to follow from chat settings

ds-boyce:
Displays a follow link for chat members you’re not already following.

https://github.com/user-attachments/assets/781b2ae4-cece-4395-8915-4f55782667fe

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -24,11 +24,6 @@
       "count": 5
     }
   },
-  "src/analytics/metrics/types.ts": {
-    "@typescript-eslint/no-explicit-any": {
-      "count": 2
-    }
-  },
   "src/components/Button.tsx": {
     "@typescript-eslint/no-explicit-any": {
       "count": 2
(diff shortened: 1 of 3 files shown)


src/screens/Messages/ConversationSettings/Member.tsx
```diff
@@ -109,6 +138,14 @@ export function Member({
             </ProfileCard.Header>
           </ProfileCard.Outer>
         </ProfileCard.Link>
+        {isSelf || isFollowing ? null : (
+          <SimpleInlineLinkText
+            label={l`Follow ${displayName}`}
+            {...createStaticClick(handleFollow)}
```
mozzius, line 144: as an aside: not a fan of this pattern. imo we should have an `InlineButton` or somesuch instead

eslint-suppressions.json
```diff
@@ -24,11 +24,6 @@
       "count": 5
     }
   },
-  "src/analytics/metrics/types.ts": {
-    "@typescript-eslint/no-explicit-any": {
-      "count": 2
-    }
-  },
```
mozzius, line 31: awesome

mozzius (approved):

PR #10417, opened 2026-05-04
Update presentation of chat requests

ds-boyce:
* Reduces width and increases size of buttons
* Increases margin above buttons
* Uses primary button color for “Accept”
* Adds icons to buttons on chat screen
* Adds inviter information to chat screen

**Before**

https://github.com/user-attachments/assets/44f19e47-d7bb-44a8-a31f-b73ac20dbff7

**After**

https://github.com/user-attachments/assets/9f479767-e456-4dff-9450-12a3d5d5afab

### DIFF
--- a/src/components/KnownFollowers.tsx
+++ b/src/components/KnownFollowers.tsx
@@ -5,9 +5,7 @@ import {
   moderateProfile,
   type ModerationOpts,
 } from '@atproto/api'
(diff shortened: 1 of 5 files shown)


mozzius (approved):
lgtm, will need to touch up the other place where those buttons are used as dicussed

PR #10419, opened 2026-05-05
Update to Node.js v24

ds-boyce:
Node.js v20 is now EOL.

* Updates to Node.js v24.15.0 across the board (latest LTS release).
* Updates `@types/node` to v24.12.2.
* For `bskyogcard`:
   * Updates TypeScript and `tsconfig.json` for compatibility.
   * Removes deprecated `ts-node` package in favor of `tsx` and running directly with `node`.
   * Adds missing `@types/react` dependency.
* Adds a small script to GitHub Actions to ensure pinned Node.js versions stay in sync with `.nvmrc`.


bskyogcard/package.json
```diff
@@ -3,11 +3,14 @@
   "version": "0.0.0",
   "type": "module",
   "main": "src/index.ts",
+  "engines": {
+    "node": ">=24.15.0"
+  },
   "scripts": {
-    "start": "node --loader ts-node/esm ./src/bin.ts",
-    "dev": "node --watch-path ./src --loader ts-node/esm ./src/bin.ts",
+    "start": "tsx ./src/bin.ts",
```
mozzius, line 10: Node 24 supports typescript syntax! We just need to enable the erasable syntax only mode in tsconfig

ds-boyce: Oh, right! I’ll update.

ds-boyce: Ah, a couple things:
* Can’t just update the `start` and `dev` scripts since they use JSX transforms.
* The `erasableSyntaxOnly` option was added in TypeScript v5.8, but we’re on v5.4.5, so I’d have to update that as well.

ds-boyce: I’m going to go ahead and update TypeScript to match the root.

estrattonbailey (approved):
LGTM

PR #10423, opened 2026-05-05
Display chat message underneath embed, not above

ds-boyce:
Just updates the order of things per the latest design.

embed


mozzius (approved):
much better imo

PR #10425, opened 2026-05-05
Update presentation of settings menus

ds-boyce:
Just updates the presentation per the latest design.

menu


src/screens/Messages/ConversationSettings/MemberMenu.tsx
```diff
@@ -213,31 +214,35 @@ export function MemberMenu({
                     ? handleBlockMember
                     : blockMemberPrompt.open
                 }>
-                <Menu.ItemText>
-                  <Trans>Block</Trans>
+                {profile.viewer?.blocking ? (
+                  <PersonCheck fill={t.palette.negative_500} size="lg" />
```
mozzius: would rather keep using `Menu.ItemIcon` etc - can we add a `destructive` variant to `Menu.Item` instead that turns text + icon red automatically?

ds-boyce: Yes indeed.

mozzius (approved):

PR #10426, opened 2026-05-05
Update context menu presentation

ds-boyce:
* Removes border between items
* Moves icon to left side
* Adjusts center alignment of text
* Swaps Translate and Report icons

menu


mozzius (approved):
lgtm

PR #10429, opened 2026-05-05
Display message timestamp in context menu

ds-boyce:
Simplifies the logic required to display a timestamp, avoids accidental taps, and reduces clutter in the UI.

| iOS | Web |
| - | - |
| emoji | web |


src/components/ContextMenu/index.tsx
```diff
@@ -696,6 +701,7 @@ export function Outer({
                   a.shadow_md,
                   t.atoms.bg_contrast_25,
                   a.w_full,
+                  a.p_sm,
```
mozzius: Unfortunately this breaks the hitbox of the press-and-hold system. We have to manually account for positioning changes here.

Relatedly:
- the hover state of the items don't look so great anymore if we go with the padding, maybe make them all rounded?
- imo we should keep the same padding as before on web

https://github.com/user-attachments/assets/6d5e3bec-1c25-424d-bc19-914ced38d53e

mozzius (changes requested):
Two things

src/components/ContextMenu/index.tsx
ds-boyce: Good callouts, I’ll take another pass.

mozzius (approved):
goodbye fun easter egg

PR #10434, opened 2026-05-06
Preserve memoization of message list

ds-boyce:
`Convo.getItems()` in `agent.ts` was rebuilding the full `ConvoItem[]` on every state commit, and the final `map()` would spread every item. As a result, every `ConvoItem` got a brand-new object identity on every commit, even when only one message changed.

This was most evident when adding a reaction to a message. Despite none of the message content changing for the rest of the thread, every single `MessageItem` would re-render. Since `disableVirtualization` and `removeClippedSubviews` are both `true` on the `List`, this could result in hundreds of re-renders for a single reaction.

For example, in a 60-message thread this would yield 120 re-renders: Every message times two commits (optimistic and server response).

This PR removes the `map()` and gets the associated data within `renderItem` instead. With this change, there are only 6 re-renders: The reacted-to message and its immediate neighbors times the two commits.

Verified this locally by logging `MessageItem` renders in a 60-message thread on this branch vs `main`.

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -484,11 +484,6 @@
       "count": 3
     }
   },
-  "src/state/messages/convo/agent.ts": {
-    "@typescript-eslint/no-explicit-any": {
-      "count": 2
-    }
-  },
   "src/state/messages/events/agent.ts": {
     "@typescript-eslint/no-explicit-any": {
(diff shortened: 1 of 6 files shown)


mozzius (approved):
*extremely* nice, great find

PR #10439, opened 2026-05-07
Add region notice to age assurance card

vineyardbovines:
Add more information to explain to the user why they might be seeing an age assurance requirement, and denote that their location is an estimate with potential external influence (VPN, etc.)

<figure>
CleanShot 2026-05-07 at 12 05 51
<figcaption>native</figcaption>
</figure>

<figure>
CleanShot 2026-05-07 at 12 05 55
<figcaption>web</figcaption>
</figure>

### DIFF
--- a/src/components/ageAssurance/AgeAssuranceAccountCard.tsx
+++ b/src/components/ageAssurance/AgeAssuranceAccountCard.tsx
@@ -9,8 +9,8 @@ import {AgeAssuranceAppealDialog} from '#/components/ageAssurance/AgeAssuranceAp
 import {AgeAssuranceBadge} from '#/components/ageAssurance/AgeAssuranceBadge'
 import {AgeAssuranceConfigUnavailableError} from '#/components/ageAssurance/AgeAssuranceErrors'
 import {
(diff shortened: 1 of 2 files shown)


src/components/ageAssurance/AgeAssuranceAccountCard.tsx
```diff
@@ -228,3 +215,81 @@ function Inner({style}: ViewStyleProp & {}) {
...
+        />
+      )}
+
+      {region && (
+        <Text style={[a.text_sm, a.leading_snug]}>
+          {isGPS ? (
+            <Trans>
+              Based on your device's location, we think you're in{' '}
+              <Text style={[a.text_sm, a.font_bold]}>{region}</Text>. Some
+              features are limited because of regional age laws.
```
estrattonbailey: The 2nd half here is somewhat redundant next to the first line of copy that pre-existed this PR. We could probably combine `Based on your device's location, we think you're in <location>. his estimate may be inaccurate if you're using a VPN.` and just have two lines of copy on this card.

estrattonbailey (approved):

PR #10441, opened 2026-05-07
Fix issue where last `RecentProfileItem` can't be removed

ds-boyce:
Noticed this while testing search. Previously, tapping the **X** on the last `RecentProfileItem` had no effect. The entry was removed from storage, but render was using stale data. Filtering against the source of truth resolves the issue.

### DIFF
--- a/src/screens/Search/Shell.tsx
+++ b/src/screens/Search/Shell.tsx
@@ -94,12 +94,12 @@ export function SearchScreenShell({
   const [activeTab, setActiveTab] = useState(() => getTabIndex(tabParam))
 
   // Query terms
-  const [searchText, _setSearchText] = useState<string>(queryParam)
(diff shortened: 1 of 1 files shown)


src/screens/Search/Shell.tsx
```diff
@@ -435,7 +444,11 @@ export function SearchScreenShell({
         ) : (
           <SearchHistory
             searchHistory={termHistory}
-            selectedProfiles={accountHistoryProfiles?.profiles || []}
+            selectedProfiles={
+              accountHistoryProfiles?.profiles.filter(p =>
+                accountHistory.includes(p.did),
+              ) ?? []
+            }
```
ds-boyce, line 451: This is the fix, the rest of the changes are to address lint warnings.

mozzius (approved):

PR #10443, opened 2026-05-07
[APP-2146] Update GIF content filtering and [APP-2152] harden web GIFs

vineyardbovines:
- Allow up to PG-13 rated GIFs via `low` content filter param (APP-2146)
- Swap browser detection via UA sniffing with `canPlayType` and multiple source tags
  - For KLIPY we already get both mp4 and webm in the response (for Tenor we could derive both from the same shared URL)
- Resume playback on `visibilitychange` so GIFs unfreeze with a backgrounded tab returns to the foreground (mostly affects Safari)
- Dropped unused feature flag enum entry

### DIFF
--- a/src/features/gifPicker/queries.ts
+++ b/src/features/gifPicker/queries.ts
@@ -56,7 +56,7 @@ function createKlipyApi<Input extends object>(
     // 30 is divisible by 2 and 3, so both 2 and 3 column layouts can be used
     params.set('limit', '30')
 
-    params.set('contentfilter', 'high')
+    params.set('contentfilter', 'low')
 
     const locale = getLocales?.()?.[0]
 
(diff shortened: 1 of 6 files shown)


src/features/gifPicker/queries.ts
```diff
@@ -56,7 +56,7 @@ function createKlipyApi<Input extends object>(
     // 30 is divisible by 2 and 3, so both 2 and 3 column layouts can be used
     params.set('limit', '30')
 
-    params.set('contentfilter', 'high')
+    params.set('contentfilter', 'low')
```
ds-boyce: Can we leave a comment here indicating this is (or should be) the equivalent of PG-13?

ds-boyce (approved):
One minor comment but nice work!

src/features/gifPicker/queries.ts
vineyardbovines: done!

PR #10455, opened 2026-05-11
Update chat menu styles and options

ds-boyce:
Updates presentation to match latest design and hides options not relevant to group chats.

| Chat | Group chat |
| - | - |
| chat | group |

https://github.com/user-attachments/assets/c2226bae-8dea-4c36-8b49-acc469c7d168

### DIFF
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -1,9 +1,7 @@
 import {memo, useCallback} from 'react'
 import {Keyboard, View} from 'react-native'
-import {type ChatBskyConvoDefs, type ModerationCause} from '@atproto/api'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
-import {Trans} from '@lingui/react/macro'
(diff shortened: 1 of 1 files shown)


src/components/dms/ConvoMenu.tsx
```diff
@@ -228,75 +225,77 @@ function MenuContent({
...
               <Trans>Unmute conversation</Trans>
             ) : (
               <Trans>Mute conversation</Trans>
             )}
           </Menu.ItemText>
-          <Menu.ItemIcon icon={convo?.muted ? Unmute : Mute} />
-        </Menu.Item>
-      </Menu.Group>
-      <Menu.Divider />
-      <Menu.Group>
```
mozzius, line 273: I think it would be nice to keep the group (which is native only btw) what do you think?

mozzius (approved):

src/components/dms/ConvoMenu.tsx
ds-boyce: Yeah, since it’s native only let’s keep it.

PR #10466, opened 2026-05-12
Fix stale display name for renamed pinned lists

fahadahmed32:
### **Summary**
When a pinned list's name is changed, the updated name wasn't reflected in the pinned feeds sidebar/tab bar until the app was restarted.

usePinnedFeedsInfos and useSavedFeeds both cache list metadata with staleTime: INFINITY, keyed only by the list URIs. Since renaming a list doesn't change its URI, the cache was never invalidated on edit — leaving the old name displayed indefinitely.

Fix: invalidate the feed-info query cache in useListEditMutation.onSuccess, so pinned and saved feed display names refresh immediately after a list is renamed.

### **Test plan**

- Pin a list
- Rename the list
- Confirm the new name appears in the pinned feeds sidebar (desktop) and tab bar (mobile) without restarting

### DIFF
--- a/src/state/queries/feed.ts
+++ b/src/state/queries/feed.ts
@@ -413,12 +413,14 @@ const PWI_DISCOVER_FEED_STUB: SavedFeedSourceInfo = {
   contentMode: undefined,
 }
 
+export const FEED_INFO_RQKEY_ROOT = 'feed-info'
+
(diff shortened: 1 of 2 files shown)


vineyardbovines (approved):
nice find!

PR #10470, opened 2026-05-12
Render chat skeleton instead of loading spinner

ds-boyce:
Adds a chat skeleton that appears immediately and fades into the conversation once the latter is ready.

To facilitate this, I moved the composer/chat footer to avoid duplicating it or creating a separate placeholder (both of which could get out of sync). Once I pulled that thread, I moved a few more things out as well. Most of this was to reduce the scope of `ChatBody`; now its sole function is to render the layout.

Note that there’s some logic specific to Liquid Glass, both for positioning the last message in the skeleton and for ensuring the top blur is applied there as well.

**Android**

https://github.com/user-attachments/assets/2aea4e67-7ed0-4a18-ba12-26224d0e17ce

**iOS**

https://github.com/user-attachments/assets/028d4edc-9549-4334-a014-16b1bfcc2aee

**Web**

https://github.com/user-attachments/assets/41e9ebdf-f7be-447e-8ea2-0a6645eb33a6

### DIFF
--- a/src/components/dms/ChatEmptyPill.tsx
+++ b/src/components/dms/ChatEmptyPill.tsx
@@ -6,8 +6,7 @@ import Animated, {
   useSharedValue,
   withTiming,
 } from 'react-native-reanimated'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
+import {useLingui} from '@lingui/react/macro'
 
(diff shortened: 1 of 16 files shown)


ds-boyce:
This is pretty brittle and the result isn’t as nice as I hoped, so closing this in favor of a smaller-scoped solution.

PR #10477, opened 2026-05-12
Improve birthdate handling

ds-boyce:
This PR addresses an issue where if AA state updated while the dialog was open, the `hasChanged` value would become `true` due to a `Date` reference comparison (even though the underlying birthdate was the same), which would in turn enable the save button.

Pressing the button at that point would save the same birthdate (effectively a no-op) but also trigger the 48-hour lockout. Here we compare timestamps instead, ensuring the button is only enabled when the user actually changes the value, avoiding unnecessary lockouts.

Also:

* Uses `useCleanError` in `BirthdateSettingsDialog` for consistency with `BirthdayInner`.
* Avoids casting via `error.toString()`, which can result in `[object Object]` in some cases.
* ~Avoids modifying the `birthDate` value returned from the `usePreferencesQuery` hook (violates the rules of hooks).~

### DIFF
--- a/src/state/queries/preferences/index.ts
+++ b/src/state/queries/preferences/index.ts
@@ -53,7 +53,7 @@ export function usePreferencesQuery() {
         const res = await agent.getPreferences()
 
         // save to local storage to ensure there are labels on initial requests
-        saveLabelers(
+        void saveLabelers(
           agent.did,
           res.moderationPrefs.labelers.map(l => l.did),
         )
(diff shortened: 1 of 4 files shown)


src/state/queries/preferences/index.ts
```diff
@@ -80,6 +80,14 @@ export function usePreferencesQuery() {
     },
     select: useCallback(
       (data: UsePreferencesQueryResponse) => {
+        /**
+         * The persisted query cache stores dates as strings, but our code
+         * expects a `Date`.
+         */
+        if (data.birthDate && !(data.birthDate instanceof Date)) {
+          data = {...data, birthDate: new Date(data.birthDate)}
+        }
```
estrattonbailey: This comment is load bearing, perhaps it could be clearer. @mozzius double check me, but the returned data from `select` is cached and persisted to disk, and as the comment states, storing a `Date` will break the persisted layer. That's why the mutation was outside the query even though it's technically naughty.

mozzius: Well, sorta - I never verified it, but that's why we kept it in outside the query yeah

ds-boyce: Yeah I’m not entirely happy with this approach, although it seems we don’t access this value from the cache anywhere atm (obviously that could change). I can revert this bit and/or dig into it more.

ds-boyce: Reverted for now.

estrattonbailey (approved):
LGTM!

PR #10482, opened 2026-05-13
[pnpm] Migrate `bskyembed`

mozzius:
Migrates to pnpm, upgrades various deps. In particular, bumps the vite version


ds-boyce (approved):
📦

PR #10483, opened 2026-05-13
[bskyembed] Migrate to `@atproto/lex`+`@bsky/sdk`

mozzius:
uncompressed: 851.21 kB -> 167.00 kB = **80% smaller**
gzipped: 171.15 kB -> 57.12 kB = **67% smaller**

## Before

Screenshot 2026-05-13 at 15 16 20

## After

Screenshot 2026-07-17 at 11 54 13

## Bundle

Screenshot 2026-07-17 at 11 54 33


bskyembed/src/util/facet-renderer.ts
mozzius, line 1: claudeslopped this minimal version of RichText which just does facets -> segment array. we should think about what an ideal richtext library looks like

bskyembed/src/util/facet-renderer.ts
```diff
@@ -0,0 +1,87 @@
...
+  get link(): Link | undefined {
+    return this.facet?.features.find(
+      f => f.$type === 'app.bsky.richtext.facet#link',
+    ) as Link | undefined
+  }
+
+  get mention(): Mention | undefined {
+    return this.facet?.features.find(
+      f => f.$type === 'app.bsky.richtext.facet#mention',
+    ) as Mention | undefined
```
matthieusieben: You should use this for `$type` checking:

```suggestion
    return this.facet?.features.find(facetDefs.mention.isTypeOf)
```

bskyembed/.gitignore
```diff
@@ -3,3 +3,6 @@ node_modules
 dist
 dist-ssr
 *.local
+
+lexicons
```
matthieusieben: The `lexicons` folder should be checked out to avoid build failures if resolution fails during build.

bskyembed/package.json
```diff
@@ -10,10 +10,17 @@
...
     "lint": "eslint --cache src",
     "typecheck": "tsc --noEmit",
-    "format": "prettier -w src"
+    "format": "prettier -w src",
+    "lex": "lex",
+    "lex:install": "lex install",
+    "lex:update": "lex install --update",
+    "lex:ci": "lex install --ci",
+    "lex:build": "lex build --lexicons ./lexicons --out ./src/lexicons --clear",
+    "postinstall": "pnpm lex:install && pnpm lex:build"
```
matthieusieben: "postinstall" should run `lex install --ci` (with **`--ci`** !) to ensure that the downloaded lexicons match the `cid` of the lexicons.json manifest.

Note that nothing will be downloaded if the lexicons json files are checkedout in git.

bskyembed/src/screens/post.tsx
```diff
@@ -1,21 +1,20 @@
...
 const root = document.getElementById('app')
 if (!root) throw new Error('No root element')
 
-const agent = new AtpAgent({
-  service: 'https://public.api.bsky.app',
-})
+const client = new Client('https://public.api.bsky.app')
 
 const uri = `at://${window.location.pathname.slice('/embed/'.length)}`
 if (!uri) {
```
matthieusieben: you can use a type guard here (instead of a type cast later):

```suggestion
if (!isAtUriString(uri)) {
```

bskyembed/.gitignore
mozzius: Ah okay, I wasn't sure which was best

PR #10491, opened 2026-05-13
Don't show Follow link if user is blocked

ds-boyce:
Missed this case in #10415.


src/screens/Messages/ConversationSettings/Member.tsx
```diff
@@ -138,7 +138,7 @@ export function Member({
             </ProfileCard.Header>
           </ProfileCard.Outer>
         </ProfileCard.Link>
-        {isSelf || isFollowing ? null : (
+        {isSelf || isFollowing || profile.viewer?.blocking ? null : (
```
mozzius: We should make this bidirectional. iirc there's a isBlockedOrBlocking  util

mozzius (approved):

mozzius (approved):

PR #10500, opened 2026-05-15
Reduce lightbox button size on native

mozzius:
The current size feels way too chunky and draws attention away from the image.

### Before

image

### After

Simulator Screenshot - iPhone 17 - 2026-05-15 at 18 01 33


mozzius:
@claude please review

PR #10508, opened 2026-05-15
 Fix links, mentions, and tags not rendering in composer reply-to preview

tolik518:
The reply-to preview in the composer rendered post text as plain text, so links, mentions, and hashtags appeared unstyled and weren't interactive.

The change is pretty small, the plain `<Text>` was jsut replaced with `RichText` (passing through the post facets)

Before: 
image

After:
image


tolik518:
Could you take a look @pfrazee

PR #10518, opened 2026-05-17
Tweak location strings

surfdude29:
In response to [feedback on Discord](https://discord.com/channels/1338165055992107061/1338165056596082731/1503747994640715907), this PR proposes tweaking five age assurance strings related to location from `Confirm your location` to `Update your location`, as I think this would be clearer for users.

As an example of somewhere else that similar copy is used, Google has an `Update location` link in the footer of search results pages on mobile which performs a similar function: it prompts the user to grant access to device geolocation.

IMG_2170

Also, this PR proposes adding `with GPS` to the end of strings that don't currently include it, because I believe it would be helpful to make that clear to users before the dialog is shown.

cc: @nilaallj


estrattonbailey (approved):

surfdude29:
@estrattonbailey Thanks for taking a look, I just synced to main so hopefully the lockfile check will pass 🤞

Issue #8465, opened 2025-06-09
Some posts with text and an image attached have the last bit of the text cut off


diabler-ie:
### Steps to Reproduce

Unsure how to reproduce this issue for others, but for me the issue persists for the same posts after restarting the app and navigating back to the post in question. In other words It's not all image+text posts, but the ones it effects it effects consistently.

### Attachments

This is an example post that is effected by this big (at least for me)

https://bsky.app/profile/nathangrayson.bsky.social/post/3lr4uvw62fc27

![Image](https://github.com/user-attachments/assets/cb80018c-d383-4944-b4f9-0d13353ea609)

![Image](https://github.com/user-attachments/assets/aaf4e1b5-62d3-4e86-8df5-a7512a6f33d7)

### What platform(s) does this occur on?

Android

### Device Info

Google Pixel 9 Pro XL, GrapheneOS

### What version of the app are you using?

Build version: 1.103.0.653; Bundle info: 6fbd6605d (prod); Bundle date: 25060521; Platform: android; Platform version: 35; Anonymous ID: 5493eb43-ce40-41ba-944a-d6928448c7d8

diabler-ie:
In case it's unclear from the above, the full text of the post is visible once I tap into the specific post to view replies—but there is no indication that there's additional text I'm not seeing until I've done that

Necco-Z:
Can confirm this, and it doesn't happen only to posts with images. From what I can gather, randomly the client decides to not include the words that would form the last line if it's too short. This also only happens in the timeline - if you open the post the text is complete.

![Image](https://github.com/user-attachments/assets/37542a72-8f87-4587-b33d-723e9ffeed79)
![Image](https://github.com/user-attachments/assets/c8d5ffc4-dcbe-4ce2-b306-3d67ccb34a06)

arturo32:
Does anyone in version 1.130.0.1493 or later can reproduce this anymore? Because I can't!! 

I have no idea what/who fixed it

diabler-ie:
Fixed on the newer versions for me as well! 🥳

Issue #8161, opened 2025-04-08
Final line of post missing in Android app when non-focused


mcclure:
### Steps to Reproduce

Warning, my example thread here is "political"! (The subject is historical struggles to unionize industries in the US and Europe.)

To reproduce:

1. Look at this post https://bsky.app/profile/myg3l.bsky.social/post/3lmdfnv4q422r
2. Click EITHER one post up https://bsky.app/profile/dryad.technology/post/3lmdfhq5b622a or one post down https://bsky.app/profile/dryad.technology/post/3lmdgayoki22a
3. After "focusing" on one of those other two posts, scroll to the post from step 1. Notice the final sentence is missing.

### Attachments

On my phone, it looks like this viewing "post 1"

![Image](https://github.com/user-attachments/assets/86cbf1f9-3034-4fef-a9a2-064979cdd0eb)

And like this viewing "post 2".

![Image](https://github.com/user-attachments/assets/9b40a8cf-8073-4d26-86cc-6d50cefda923)

Notice the second sentence, "where did those Americans go?", is missing. 

### What platform(s) does this occur on?

Android

### Device Info

Sony Xperia 5 III, Android 13

### What version of the app are you using?

Build version: 1.99.0.621; Bundle info: 7f52ba1b0 (prod); Bundle date: 25040112; Platform: android; Platform version: 33; Anonymous ID: 4e736962-ce94-4245-8979-7dc418bf25fe

### Additional Information

Maybe related to

https://github.com/bluesky-social/social-app/issues/7122

?

I think this is a pretty serious bug. Editing content out of a post can change its meaning, and there is nothing to indicate words were removed. In the case of the above thread, my reply ("You mean Thomas Paine?") appears completely nonsensical because the question it is responding to was edited out.

mikekasprzak:
Yeah, I also suspect it's related to https://github.com/bluesky-social/social-app/issues/7122. Looking at my own version over there, initially I thought it was the emoji I used messing with my extents, but none of the other posts use emoji.

I suspected blank line is somehow the culprit, as this bug is slightly different in that a whole line is displaced (versus one word/token), but it could be a red haring.

If I was to guess, I expect the code that draws a post and the code that measures it are out of sync, one of of them rounding or flooring the text extents while the other doesn't. That would explain #7122 (which I'll consider the "single word wrap" bug), but unless there's another problematic glyph or combination glyph+font size that rounds/floors differently. 🤔

[...]

Issue #7122, opened 2024-12-15
[Android] Word wrap cuts off final word of post in some views


mcclure:
### Steps to Reproduce

I open this post on my Android phone. It is the second post of a thread.

https://bsky.app/profile/dryad.technology/post/3ldemwed4ws2j

I see this

![Screenshot_20241215-155905](https://github.com/user-attachments/assets/016bfb04-89ef-473d-a22f-e91eac624a55)

The final two words of the first post of the thread are "chill music". The word "music" is cut off. There is no … or anything else to indicate missing content.

I tap the first post of the thread, to focus it. Now I see this

![Screenshot_20241215-155909](https://github.com/user-attachments/assets/4a9cd540-6f4f-482c-b8bf-6e8edcbab53a)

Now the word "music" is back. The final word is only cut off under some specific conditions.

### What platform(s) does this occur on?

Android

### Device Info

Song XPeria 5 III, Android 13

### What version of the app are you using?

Build version: 1.95.0.441; Bundle info: 41243e874 (prod); Bundle date: 24120701; Platform: android; Platform version: 33

mikekasprzak:
lol, hi mcc. 😉 

I've seen the "missing last line" bug a few times myself, but here's an example on the main feed where the last word (`crime.`) got cut off. Screenshot is below. I'm on a Pixel 8 Pro.

Link: <https://bsky.app/profile/did:plc:svpym4ujks7qxczscyzq7fuy/post/3lnnlayhdh22v>

⚠ WARNING ⚠: Political, also someone's steak snuck into my timeline.

.
.
.
.
.
.

![Image](https://github.com/user-attachments/assets/2d02ec73-973f-46e6-952d-ee4f6c369dcd)

arturo32:
For now I'm having trouble reproducing this error in the developing mode (I don't know how to run this app on Expo Go and the emulator does not show the error. I will have to figure out how to download the apk from source...).

But I have seen this error many times in my Xiomi Redmi phone (android 14).
The example bellow is a random phrase that should end with "(não respondam)". 
I wrote it inspired by another post with the same bug. I used the same number of characters but it didn't work initially. It is very finicky to reproduce: **the length of each character matters**. Only when I changed "l" and "i" to "b" I could reproduce the bug with my post:

![Image](https://github.com/user-attachments/assets/2a0ea355-87d7-4e10-8676-7bb94d8d2ad2)

[...]

Issue #5835, opened 2024-10-18
Post with too many lines gets cut off


beholdnec:
### Steps to Reproduce

1. Go to https://bsky.app/profile/dotticus.bsky.social/post/3l6pize2jef2t
2. Notice in the reply by tudsworthington, the final line "loser" is cut off. There is no indication that the text was cut off.
3. Tap on the post. The full text becomes visible.

### Attachments

![1000063466](https://github.com/user-attachments/assets/49fb0a4d-9741-46fc-a50d-96b48571b061)
![1000063467](https://github.com/user-attachments/assets/95163ebc-3f10-4307-98e7-2704572475dc)

### What platform(s) does this occur on?

Android

### Device Info

Android 15

### What version of the app are you using?

Build version: 1.92.1.355; Bundle info: 3fd8c0613 (prod); Bundle date: 24101721; Platform: android

lyndonhood:
I'm getting lines cut off regularly. I believe it can happen anywhere but more lines make it more likely. 

E.g. for me my reply to https://bsky.app/profile/ghet.bsky.social/post/3lrfbablcps2b has the last word cut off when viewing that post (not, in this case, in the timeline or my replies screen) but when expanded https://bsky.app/profile/lyndonhood.bsky.social/post/3lrfbhqg5z22n shows the whole text, as one line.

Redmi Note 9T; Android 12 MUIU 14.0.4; version 1.103.0.653

PR #10523, opened 2026-05-17
Workaround Android text widow truncation in RichText

### LINKED ISSUE #5235: Truncated Widowed Words in Feed View on Android
**Describe the bug**
The last word of a post is sometimes truncated on the feed view when a single word is 'widowed' to a new line.

**To Reproduce**

1. I don't have an exact steps to reproduce, but it seems to happen most often when the widowed word would push the post from 2 lines to 3.

**Expected behavior**

The post height would expand to make all the words in the post visible

**Screenshots**

Here's a screen recording demonstrating the issue with the final word in the first post in Cooper Lund's feed ("Trump")

**Details**

 - Platform: Android 
 - Platform version: 14
 - App version: 1.90.1, although I've seen it with basically every version going all the way back to the beginning

**Additional context**

I'm on a Pixel 8 Pro, if that matters.

diabler-ie:
## Summary

Fixes #5235.

On Android, the last word of a post is sometimes silently dropped in the feed view when the final word would widow onto a new line. Tapping into the post detail view shows the full text, so users routinely miss content — particularly painful since the missing word often changes the meaning of the post.

## Root cause

This is an upstream React Native Android text-layout bug:

- facebook/react-native#53286 — Text rendering cut off on Android 15 & 16
- facebook/react-native#53344 — minimal reproducer (lineHeight + row)

The bug triggers when a `<Text>` with an explicit `lineHeight` is rendered inside a row-flex container and a single word would widow onto a new line. The Android text engine miscalculates the wrap measurement and the last line is never rendered. RichText hits this because post text is rendered with `a.leading_snug` (lineHeight) inside the feed item's avatar-content row.

## Fix

Append a non-breaking space (`U+00A0`) to the rendered post text on Android. The trailing NBSP shifts the wrap measurement past the buggy code path.

NBSP rather than a regular space because trailing ASCII whitespace gets trimmed by Android's text engine before measurement — I verified that a regular trailing space does **not** fix the bug, but NBSP does.

The constant is `''` on iOS and web, so this is Android-only and has no effect on those platforms.

## Test plan

- [x] Built a patched debug + release APK on a Pixel 9 Pro XL (Android 16), confirmed previously-truncated posts (e.g. ones with widow-prone trailing words) now render in full.
- [x] Scrolled home feed and Likes feed; could not find a reproducer of the original bug after the patch.
- [x] Sanity-checked text content with the patch — NBSP at end of text is invisible and copy-paste behavior is unchanged for short and long posts.
- [ ] Cross-device verification (only tested on one device — the bug's intermittent nature makes broader confirmation valuable).

## Notes

- The mitigation is one trailing character. The real fix needs to land upstream in RN's Android Text component; this is intended as a workaround until then.
- RichText is the shared rendering point for feed, detail, thread, and other surfaces — fixing it here covers all call sites without requiring per-site changes.

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -9,8 +9,20 @@ import {InlineLinkText, type LinkProps} from '#/components/Link'
 import {ProfileHoverCard} from '#/components/ProfileHoverCard'
 import {RichTextTag} from '#/components/RichTextTag'
 import {Text, type TextProps} from '#/components/Typography'
+import {IS_ANDROID} from '#/env'
 
 const WORD_WRAP = {wordWrap: 1}
+/*
+ * Workaround for an RN Android text-layout bug (see facebook/react-native#53286
+ * and #53344) that drops the last wrapped line when a single word would widow
+ * inside a row-flex container with `lineHeight` set. Appending a non-breaking
+ * space at the end of the text shifts the wrap measurement past the buggy
+ * path. NBSP is used over a regular space because trailing ASCII whitespace
+ * gets trimmed by Android's text engine before measurement.
+ *
+ * Bluesky issue: bluesky-social/social-app#5235
+ */
+const ANDROID_WIDOW_FIX = IS_ANDROID ? '\u00A0' : ''
 // lifted from facet detection in `RichText` impl, _without_ `gm` flags
 const URL_REGEX =
   /(^|\s|\()((https?:\/\/[\S]+)|((?<domain>[a-z][a-z0-9]*(\.[a-z0-9]+)+)[\S]*))/i
@@ -101,7 +113,7 @@ export function RichText({
         onTextLayout={onTextLayout}
         // @ts-ignore web only -prf
         dataSet={WORD_WRAP}>
-        {text}
+        {text + ANDROID_WIDOW_FIX}
       </Text>
     )
   }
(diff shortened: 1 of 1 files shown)


surfdude29:
If this PR fixes the problem, it will also close these issues, which are duplicates of #5235 (thanks to @arturo32 for collating them):
- #5835
- #7122
- #8161
- #8465

arturo32:
Thank you @surfdude29 for the recognition!

Just to be sure: while trying to reproduce the error with my own posts I discovered that [the size of each character present in the post matters](https://github.com/bluesky-social/social-app/issues/7122#issuecomment-3793746441). What I'm afraid of this fix is that it may fix the issue for phones with this error right now, but it may start to appear in different models with different screen sizes, shifting the problem.

I do not fully understand the cause of this bug nor your fix. Not because your description wasn't good, just because I'm tired and I'm not a mobile developer.

What I get of the fix is: add a small invisible special space at the end of the post. What I'm afraid it will happen: some other combinations of word and characters will trigger this error again in the same phones that  had this error (or even phones that never had this issue).

I just wanted to highlight this behavior (which you may already know) and expose my worries here. I may have interpreted your fix in a completely wrong way.

diabler-ie:
No worries, and in looking into this further I actually found what I think is the root cause and have a better fix. Closing this one, and opening a new PR: [https://github.com/bluesky-social/social-app/pull/11127](https://github.com/bluesky-social/social-app/pull/11127)

PR #10530, opened 2026-05-18
[Chat] Handle new chat notification reasons

mozzius:
## Summary

Adds client support for three new push notification reasons emitted by the chat service:

- `chat-added-to-group`
- `chat-removed-from-group`
- `chat-join-request-rejected`

Also improves the iOS Notification Service Extension for group chats:

- Sets the notification subtitle to the group name for `chat-message`/`chat-reaction` in group convos, and passes `speakableGroupName` so iOS treats it as a group communication notification.
- Attaches the composite group avatar (`convoAvatarUrl` from the chat service push payload, rendered by the ogcard `/avatar-bubbles` endpoint) to the `speakableGroupName` parameter on the `INSendMessageIntent`, so iOS shows the group avatar alongside the sender's avatar on the lock screen.
- For `chat-message` notifications with a non-`message` `messageKind` (system events like `add_member`, `convo_locked`, `edit_group`), copies `title` into `body` before applying `INSendMessageIntent` (otherwise iOS replaces `title` with the sender's display name and the system description is lost). This is scoped to `chat-message` only - on `chat-reaction`, `messageKind` refers to the reacted-to message, so applying the swap there would clobber the descriptive reaction body for reactions on system messages.
- New branch in the NSE for the three system reasons: plays the chat sound (if `playSoundChat` is enabled), sets the group-name subtitle, updates the badge, no `INSendMessageIntent` since there's no real sender.

The three new reasons all route to the Messages list (no `pushToConversation`), since the convo may no longer be accessible to the recipient. They reuse the existing `chat-messages` Android channel.

While here, generalized the `chat-message`-only handling for cold-start notifications to cover all `chat-*` reasons (previously cold-start tap on a `chat-reaction` notification did nothing).

## Test plan

- [ ] On iOS, send a `chat-message` to a group convo and verify the group name appears as a subtitle on the lock screen.
- [ ] On iOS, in a group convo, send a `chat-message` and verify the lock screen shows both the sender avatar and the composite group avatar.
- [ ] On iOS, in a group convo, send a `chat-reaction` and verify the lock screen shows both the sender avatar and the composite group avatar.
- [ ] On iOS, send a `chat-message` with `messageKind=add_member` (or other system kind) in a group and verify the lock screen shows sender / group / system event description.
- [ ] On iOS, react to a system message (e.g. an `add_member` event) and verify the notification body still reads `Reacted with <emoji> to "..."` rather than the system event title.
- [ ] On iOS, send each of `chat-added-to-group`, `chat-removed-from-group`, `chat-join-request-rejected` and verify the notification renders with the group-name subtitle and respects the chat-sound preference.
- [ ] Tap each of the three new notifications (foreground, background, cold-start) and verify it opens the Messages list.
- [ ] Verify account-switch handling still works: send one of the new chat notifications to an account that isn't the current one, tap it, and confirm the app switches accounts before opening Messages.
- [ ] On Android, confirm the three new reasons still notify (they use the existing `chat-messages` channel) and tapping opens the Messages list.

### DIFF
--- a/modules/BlueskyNSE/NotificationService.swift
+++ b/modules/BlueskyNSE/NotificationService.swift
@@ -120,6 +120,15 @@ class NotificationService: UNNotificationServiceExtension {
       attachments: nil
     )
 
+    // For group convos, attach the composite group avatar (rendered by the
+    // ogcard service) to the `speakableGroupName` parameter so iOS shows it
+    // alongside the sender on the Communication Notification.
+    if userInfo["convoKind"] as? String == "group",
+       let convoAvatarUrlString = userInfo["convoAvatarUrl"] as? String,
+       let groupImage = downloadAvatarImage(from: convoAvatarUrlString) {
+      intent.setImage(groupImage, forParameterNamed: \.speakableGroupName)
+    }
+
     let interaction = INInteraction(intent: intent, response: nil)
     interaction.direction = .incoming
     interaction.donate(completion: nil)
--- a/modules/BlueskyNSE/NotificationService.swift
+++ b/modules/BlueskyNSE/NotificationService.swift
@@ -44,11 +44,20 @@ class NotificationService: UNNotificationServiceExtension {
 
     if reason == "chat-message" || reason == "chat-reaction" {
       mutateWithChatMessage(bestAttempt)
+      mutateChatMessageBody(bestAttempt, userInfo: request.content.userInfo)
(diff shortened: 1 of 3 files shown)


mozzius:
@claude review

mozzius:
@claude re-review

mozzius:
@claude please re-review

PR #10534, opened 2026-05-18
Tweak reaction size and padding

ds-boyce:
Updates appearance per latest design.


mozzius (approved):
Will need a rebase sorry!

PR #10536, opened 2026-05-18
Exclude @atproto and @bsky.app packages from minimum age exclusion

ds-boyce:
We manage these ourselves.

(Ran Prettier on the `.yaml` file.)


mozzius (approved):
lgtm, I wonder why my prettier was setting the yaml to use "

ds-boyce:
Default Prettier uses double quotes… maybe it didn’t pick up the config file.

mozzius:
Oh, it's probably because I haven't configured Zed to use prettier for .yaml, so it uses the built-in formatter

PR #10542, opened 2026-05-19
Fix up prettier config, apply to more files

mozzius:
Figured out why Prettier wasn't working on certain files - we had an overzealous allowlist. Added `mjs`/`cjs` files plus the `pnpm-workspace.yaml`


ds-boyce (approved):
Defaults when?

PR #10543, opened 2026-05-19
ESM atproto packages

mozzius:
Tried to remove as much janky config hacks as I could


ds-boyce (approved):
Took a look around on Android, iOS, and web and everything seems to be in order. Good stuff!

estrattonbailey (approved):
Nice one! Tested prod builds, linting, testing, etc. Let's get it in :shipit:

PR #10544, opened 2026-05-19
Fix Image Options not opening on web

RetroSunstar:
After a recent refresh of the lightbox, the image options button on web has since stopped opening its menu

This adds an onPress prop to the menu trigger to restore functionality 🐙

### DIFF
--- a/src/components/Lightbox/Lightbox.web.tsx
+++ b/src/components/Lightbox/Lightbox.web.tsx
@@ -268,6 +268,7 @@ function LightboxGallery({
                 icon={EllipsisIcon}
                 iconStyle={{transform: [{rotate: '90deg'}]}}
                 label={_(msg`Image options`)}
+                onPress={props.onPress}
               />
             </Pressable>
           )}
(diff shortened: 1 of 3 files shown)


src/components/Lightbox/Lightbox.web.tsx
```diff
@@ -268,6 +268,7 @@ function LightboxGallery({
                 icon={EllipsisIcon}
                 iconStyle={{transform: [{rotate: '90deg'}]}}
                 label={_(msg`Image options`)}
+                onPress={props.onPress}
               />
             </Pressable>
```
mozzius: I think we need to remove the wrapping Pressable instead, and spread `prop` into CircleChromeButton

RetroSunstar: > I think we need to remove the wrapping Pressable instead, and spread `prop` into CircleChromeButton

Something like the new commit? 🐙

mozzius: Perfect!

mozzius (approved):
Beautiful 🫡

PR #10563, opened 2026-05-20
[Chat] New post attachment style in composer

mozzius:
Screenshot 2026-05-20 at 17 10 11

### DIFF
--- a/src/screens/Messages/components/MessageInputEmbed.tsx
+++ b/src/screens/Messages/components/MessageInputEmbed.tsx
@@ -144,7 +144,6 @@ export function MessageInputEmbed({
   }
 
   switch (status) {
-    // case 'success':
     case 'pending': {
       return (
         <SimpleContainer onRemove={onRemove}>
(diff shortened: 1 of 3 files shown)


src/screens/Messages/components/MessageInputEmbed.tsx
```diff
@@ -138,88 +138,126 @@ export function MessageInputEmbed({
...
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut)
+    setEmbed(undefined)
+  }
+
   switch (status) {
-    case 'pending':
-      content = (
-        <View
-          style={[a.flex_1, {minHeight: 64}, a.justify_center, a.align_center]}>
+    // case 'success':
```
ds-boyce: This can be removed.

ds-boyce (approved):

src/screens/Messages/components/MessageInputEmbed.tsx
mozzius: Good catch!

PR #10564, opened 2026-05-20
[Chat] Add embeds to the optimistic message

mozzius:
https://github.com/user-attachments/assets/062d1bf3-8d62-4877-b155-0f096210d9c8


ds-boyce (approved):
Nice.

PR #10566, opened 2026-05-20
Skip installing expo cli in CI

mozzius:
Might work? Might not


ds-boyce (approved):
Only one way to find out…

PR #10576, opened 2026-05-21
Add Info Panel to direct chats

ds-boyce:
Similar to the group chat info panel.

ios

### DIFF
--- a/src/screens/Messages/components/MessagesListInfoPanel.tsx
+++ b/src/screens/Messages/components/MessagesListInfoPanel.tsx
@@ -1,176 +1,81 @@
 import {View} from 'react-native'
-import {Plural, Trans, useLingui} from '@lingui/react/macro'
+import {Trans, useLingui} from '@lingui/react/macro'
+import {useNavigation} from '@react-navigation/native'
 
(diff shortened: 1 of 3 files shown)


src/screens/Messages/components/MessagesListInfoPanel.tsx
```diff
@@ -1,173 +1,81 @@
...
+        ]}>
+        <Text style={[a.text_2xl, a.font_bold, t.atoms.text]}>
+          {displayName}
+        </Text>
+        <ProfileBadges profile={profile} size="lg" />
+      </View>
+      <Text style={[a.text_sm, a.mt_xs, t.atoms.text_contrast_high]}>
+        {handle}
+      </Text>
+      <View
```
mozzius, line 75: I think we should add <PostAlerts> (aka profile labels) in here

mozzius (approved):
LGTM, just needs to resolve the merge conflict

src/screens/Messages/components/MessagesListInfoPanel.tsx
ds-boyce: They appear after the handle now.

PR #10579, opened 2026-05-21
Fix chat message spacing and alignment

ds-boyce:
* Centralizes message gap logic in `MessageItem`.
* Fixes left alignment of embeds for direct chats.


mozzius (approved):

mozzius:
fwiw I'm still seeing the gap here

Screenshot 2026-05-22 at 11 56 28

PR #10581, opened 2026-05-21
Fix issue where suggested language prompt could not be dismissed

ds-boyce:
The decline handler was only tracking detected languages on new posts, not suggested reply languages. Now `onDecline` handles both cases.


mozzius:
@claude review

mozzius (approved):

PR #10589, opened 2026-05-22
[pnpm] Attempt to remove `eas-cli` from devDeps

mozzius:
This didn't work before, but we really don't want it in our lockfile. Also cleaned up `engines`/`devEngines`

The secret for getting `expo-github-action` to work with pnpm v11 is to pass
```
packager: 'pnpm --allow-build=dtrace-provider'
```


ds-boyce (approved):
Beautiful.

PR #10595, opened 2026-05-22
Add loading state to message composer and input

ds-boyce:
This will be used in a follow up PR (#10596) to replace the full-screen loading spinner.


src/screens/Messages/components/MessageInput.tsx
```diff
@@ -226,7 +229,7 @@ export function MessageInput({
           }}>
           <Pressable
             accessibilityRole="button"
-            accessibilityLabel={l`Send message`}
+            accessibilityLabel={loading ? l`Loading chat…` : l`Message`}
```
surfdude29: nit: Would it be possible to add the `action` context to the `Message` string here please? This allows translators to use a separate translation for this context as compared to the other uses of the string that refer to `Message` as a noun :)

Similar to:
https://github.com/bluesky-social/social-app/blob/4bdedc55be1d6b2f7d25c9e79b93e1dcbbea5d70/src/screens/Messages/ConversationSettings/MemberMenu.tsx#L199

mozzius (approved):

PR #10596, opened 2026-05-22
Update loading state for chats

ds-boyce:
Implements the new loading state for the composer/input from #10595.

https://github.com/user-attachments/assets/bb896912-8047-404f-b43c-425446cc0b9f

https://github.com/user-attachments/assets/03935fba-11e9-4e55-86ac-5abfb6492a5b

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -1,5 +1,4 @@
 import {
-  cloneElement,
   useCallback,
   useEffect,
   useId,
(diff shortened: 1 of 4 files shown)


src/screens/Messages/components/MessagesList.tsx
```diff
@@ -663,7 +685,12 @@ function ConversationFooter({
 
   switch (footerState) {
     case 'loading':
-      return null
+      // Render children with loading prop instead of null
+      return children
+        ? cloneElement(children as React.ReactElement<{loading?: boolean}>, {
+            loading: true,
+          })
+        : null
```
mozzius: This seems kinda nasty to me. Is there a better way to thread this data through?

ds-boyce: Yep, switched to a render prop.

mozzius (approved):
Beautiful

PR #10598, opened 2026-05-22
Create logged-out view for group chat invites

ds-boyce:
* Reworks `StarterPackProvider` as `LandingProvider` to support chat join requests (and other future types).
* Fixes behavior of back button within the sign in and create account screens so that it returns to the landing screen.
* Adds a new `LogomarkWithType` component.
* Uses the above for a new `JoinRequest` landing screen for logged-out users who follow a chat join request link.

https://github.com/user-attachments/assets/9ed37f71-2b5b-42c7-82b3-4db417835f92

https://github.com/user-attachments/assets/8ae50ae0-a5a4-497a-b161-72cfcc3caa5c


src/screens/Messages/JoinRequest.tsx
```diff
@@ -0,0 +1,253 @@
...
+              <View>
+                <Text
+                  style={[
+                    a.mb_2xs,
+                    a.text_center,
+                    a.text_sm,
+                    a.font_semi_bold,
+                    t.atoms.text,
+                  ]}>
+                  By{' '}
```
surfdude29: nit: I think the `<Trans>` tags are missing here

mozzius:
@claude please review

mozzius (changes requested):
Looking great on Chrome, however it's a bit broken on firefox. probably just needs fiddling with flexbox, I had a look at the code and nothing jumped out as obviously wrong

Also, I think I agree with Claude's comment about a11y - it's not a modal, even if it visually resembles one

Image

mozzius (approved):
Absolute 🔥

really small nit: this spacing is quite tight, leading needs adjusting maybe? or a 2px gap.

Image

PR #10548, opened 2026-05-19
[Chat] Gate group chats for under-18 users

mozzius:
## Summary

Per the [group chats spec](https://linear.app/blueskyweb/document/group-chats-spec-a68081160496), under-18 users cannot be added to or participate in group chats.

- New u18 accounts get `allowGroupInvites: 'none'` written to their chat actor declaration on signup. Existing accounts get the same write on birthdate-change-to-u18.
- The "New group chat" entry in `InitiateChatFlow` is hidden for u18.
- The "Allow group chat invites from" toggle in chat settings is locked to "No one" for u18, with explanatory copy.
- `listConvos` is called with `kind: 'direct'` for u18 so groups don't appear in the requests inbox or contribute to the requests/unread badges.

Closes APP-2165.

## Test plan

- [x] `pnpm typecheck` and `pnpm lint` pass
- [x] `pnpm test` passes (existing session/agent test still green)
- [x] Sign up a fresh u18 account: chat declaration record has `allowGroupInvites: 'none'`
- [x] As u18: New Chat dialog has no group-creation entry; chat settings show the group-invites section locked to "No one" with helper copy
- [ ] As u18: requests inbox shows no group convos and the badge in the chat header / bottom-bar doesn't count groups
- [x] As 18+: all four behaviors work as before
- [x] Birthdate change 19 → 17: `allowGroupInvites` flips to `none` and the UI updates immediately

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10609, opened 2026-05-26
AA reshuffling for group chats

estrattonbailey:
A little reshuffling of logic could make this easier to read and implement.
- [x] break out flag computation into reusable util
- [ ] DRY things up
- [ ] expose flags through the `getAndComputeAgeAssuranceState` needed for account creation
- [ ] sort out "chat disabled" vs "group chat disabled" and the interactions therein
- [ ] use semantic name `groupChatsDisabled`
- [ ] implement same `maybe*` disable chats shape

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ bskyweb/static/media/*.svg
 
 # superpowers plugin plans/specs — local-only workspace
 docs/superpowers/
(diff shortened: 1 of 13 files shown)


estrattonbailey:
Superseded by #10548

PR #10619, opened 2026-05-26
Update presentation of DM header

ds-boyce:
Primarily adds the handle after the display name.

web

ios


mozzius (approved):
Made a small tweak to handle long names, lgtm

PR #10623, opened 2026-05-26
Disable other chat settings when a chat is locked

ds-boyce:
These options error when used on a locked conversation anyway, so we remove or disable them in the UI here.

### DIFF
--- a/src/screens/Messages/components/MessagesListGroupInfoPanel.tsx
+++ b/src/screens/Messages/components/MessagesListGroupInfoPanel.tsx
@@ -83,9 +83,7 @@ export function MessagesListGroupInfoPanel({
     )
   }
 
-  const isLocked =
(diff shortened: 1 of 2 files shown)


src/screens/Messages/components/MessagesListGroupInfoPanel.tsx
```diff
@@ -83,7 +83,11 @@ export function MessagesListGroupInfoPanel({
     )
   }
 
-  const showButtons = isOwner || isJoinLinkEnabled
+  const isLocked =
+    convo.details.lockStatus === 'locked' ||
+    convo.details.lockStatus === 'locked-permanently'
```
mozzius: `!== 'unlocked'` might be clearer

mozzius (approved):

PR #10624, opened 2026-05-26
Handle new group chat log events

ds-boyce:
Also removes some unnecessary casts.


src/state/queries/messages/list-convo-members.ts
mozzius, line 1: I think we should move this into the main gigantic function in `list-conversation`, as that one is always mounted

mozzius (approved):

src/state/queries/messages/list-convo-members.ts
ds-boyce: Agreed, I’m going to take a pass at merging these as a follow-up.

ds-boyce: #10632

PR #10625, opened 2026-05-26
Remove hard-coded chat member limit

ds-boyce:
We get this from the response now so that we can potentially increase it without requiring a client update.

### DIFF
--- a/src/screens/Messages/ConversationSettings/index.tsx
+++ b/src/screens/Messages/ConversationSettings/index.tsx
@@ -246,6 +246,7 @@ function GroupSettings({
         return (
           <MembersAndRequests
             memberCount={convo.details.memberCount}
+            memberLimit={convo.details.memberLimit ?? convo.details.memberCount}
             requestCount={requestCount}
             hasMoreRequests={!!hasMoreRequests}
             isOwner={isOwner}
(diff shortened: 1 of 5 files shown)


src/screens/Messages/ConversationSettings/index.tsx
```diff
@@ -246,6 +246,7 @@ function GroupSettings({
         return (
           <MembersAndRequests
             memberCount={convo.details.memberCount}
+            memberLimit={convo.details.memberLimit ?? convo.details.memberCount}
```
mozzius: Is this nullable?

ds-boyce: limit

@rafaeleyng should `memberLimit` always be defined?

rafaeleyng: It should. I don't know exactly the shape of the object you are dealing with here, but on the lexicons, we have 3 objects that have `memberLimit`, and it is required on all of them

ds-boyce: Ok, looks like it was optional in `0.20.5` but required in `0.20.6`, so I’m going to bump `@atproto/api` and update this.

mozzius (approved):

PR #10626, opened 2026-05-26
Create logged-in dialog for group chat invites

ds-boyce:
Builds on #10598 to define a `/c/{code}` intent that handles logged-out and logged-in links, including cold-start links.

https://github.com/user-attachments/assets/f8a1ec61-be84-4d6c-9430-87184abc97d9

### DIFF
--- a/src/components/intents/GroupChatJoinDialog.tsx
+++ b/src/components/intents/GroupChatJoinDialog.tsx
@@ -390,8 +390,10 @@ function GroupChatJoinDialogContent({code}: {code?: string}) {
         <Button
           testID="openButton"
           onPress={() => {
-            navigation.navigate('MessagesConversation', {
-              conversation: convoId,
+            control.close(() => {
+              navigation.navigate('MessagesConversation', {
+                conversation: convoId,
+              })
             })
           }}
           label={l`Open group chat`}
--- a/src/components/intents/GroupChatJoinDialog.tsx
+++ b/src/components/intents/GroupChatJoinDialog.tsx
@@ -1,5 +1,9 @@
 import {View} from 'react-native'
-import {ChatBskyGroupRequestJoin, moderateProfile} from '@atproto/api'
+import {
+  ChatBskyGroupRequestJoin,
+  ChatBskyGroupWithdrawJoinRequest,
+  moderateProfile,
+} from '@atproto/api'
 import {Trans, useLingui} from '@lingui/react/macro'
 import {useNavigation} from '@react-navigation/native'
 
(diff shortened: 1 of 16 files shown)


src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,341 @@
...
+import {InlineLinkText} from '#/components/Link'
+import {Loader} from '#/components/Loader'
+import * as Toast from '#/components/Toast'
+import {Text} from '#/components/Typography'
+
+export function GroupChatJoinDialog() {
+  const {groupChatJoinDialogControl, groupChatJoinState} = useIntentDialogs()
+
+  return (
+    <Dialog.Outer control={groupChatJoinDialogControl}>
```
mozzius: We should make this the default value for this prop, but:

```suggestion
    <Dialog.Outer control={groupChatJoinDialogControl} nativeOptions={{preventExpansion: true}}>
```

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,341 @@
...
+    joinGroupChat({code})
+  }
+
+  if (isLoading || !data || !moderationOpts) {
+    return (
+      <View style={[a.p_2xl]}>
+        <Loader size="xl" />
+      </View>
+    )
+  }
```
mozzius, line 186: This spinner doesn't feel great, esp. on web. Would it be possible to preload the query instead and delay opening the dialog until it's ready?

I used this pattern here, using `Promise.race`:
https://github.com/bluesky-social/social-app/blob/main/src/components/WhoCanReply.tsx#L97-L104

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,341 @@
...
+                label={`@${joinLinkPreview.owner.handle}`}
+                to={makeProfileLink(joinLinkPreview.owner)}
+                style={[a.mb_2xs, a.text_sm, a.font_semi_bold]}>
+                {createSanitizedDisplayName(
+                  joinLinkPreview.owner,
+                  true,
+                  moderateProfile(joinLinkPreview.owner, moderationOpts).ui(
+                    'displayName',
+                  ),
+                )}{' '}
```
mozzius: Ah this probably applies to the other PR, but we need to make sure we get checkmarks in whereever we show a displayName. Super annoying when it's inline with some text tho :/

mozzius (commented):
I need to do a second pass on this to test all the combinations, but it's looking great so far! 

Can you rebase it?

mozzius:
At least on web, the checkmark seems a bit off. Perhaps `size="xs"` would be better?

Screenshot 2026-06-02 at 14 52 46

mozzius:
Question: is this expected to work on native via a deep link?


https://github.com/user-attachments/assets/8b23b6da-c36e-4b2d-aa55-9ab6d53d7dbb

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,361 @@
...
+    ButtonIconImage = WarningIcon
+    buttonText = l`Chat invite link no longer available`
+  } else if (joinLinkPreview.memberCount >= joinLinkPreview.memberLimit) {
+    canJoin = false
+    ButtonIconImage = HandIcon
+    buttonText = l`This chat is full`
+  } else if (joinLinkPreview.joinRule === 'followedByOwner' && !isFollowing) {
+    canJoin = false
+    ButtonIconImage = HandIcon
+    buttonText = l`Only people the chat owner follows can join`
```
mozzius, line 268: Using a disabled button for this state results in pretty poor contrast unfortunately. Might need to ask Darrin

Image

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,361 @@
...
+          color="secondary"
+          style={[a.w_full]}>
+          <ButtonText>{l`Close`}</ButtonText>
+        </Button>
+      </>
+    )
+  }
+
+  const convoId = joinLinkPreview.convo?.id
+  const isFollowing = joinLinkPreview.owner.viewer?.following ?? false
```
mozzius, line 246: This doesn't appear to work, I think it's a backend issue (viewer is missing)

mozzius (approved):
This is great!

I did try the end-to-end flow (going to the logged-out state and logging in) but this doesn't appear - is that expected to work or will we have to tackle it separately?

src/components/intents/GroupChatJoinDialog.tsx
ds-boyce: Is this a problem for all disabled buttons? Just wondering if we should tweak the contrast in the `Button` component vs a one-off here.

ds-boyce:
> I did try the end-to-end flow (going to the logged-out state and logging in) but this doesn't appear - is that expected to work or will we have to tackle it separately?

Fixed! It should open the dialog as intended now.

ds-boyce:
> At least on web, the checkmark seems a bit off. Perhaps `size="xs"` would be better?

Tweaked this as well.

src/components/intents/GroupChatJoinDialog.tsx
mozzius: Fixed!

src/components/intents/GroupChatJoinDialog.tsx
mozzius: Hmm fair point - let's defer this discussion for another time, it's not blocking

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -0,0 +1,430 @@
...
+          </View>
+        </View>
+      </View>
+      {convoId ? (
+        <Button
+          testID="openButton"
+          onPress={() => {
+            navigation.navigate('MessagesConversation', {
+              conversation: convoId,
+            })
```
mozzius:
```suggestion
            control.close(() => {
              navigation.navigate('MessagesConversation', {
                conversation: convoId,
              })
            })
```

mozzius (approved):
Great stuff

PR #10627, opened 2026-05-27
[ALF] Set text to `leading_snug` by default

mozzius:


estrattonbailey (approved):
Let's give it a shot and be on the lookout for issues

PR #10628, opened 2026-05-27
Use correct pnpm version in dockerfile

mozzius:


ThisIsMissEm:
Why not use corepack which will automatically read packageManager from package.json?

mozzius:
Oh yeah this is terrible I know, I just don't want to touch it right now.

ThisIsMissEm:
@mozzius ha, fair enough!

ThisIsMissEm:
oh goodness, I just saw the nvm inside docker thing. That should probably just be corepack + node version

mozzius:
As I said, terrible 😫

PR #10629, opened 2026-05-27
[Chat] Add OpenGraph card for chat invite links

mozzius:
Example URL: http://localhost:3000/chat-invite/bh79DWW

Also migrated bskyogcard to pnpm and did some various dependency upgrades

bh79DWW

# Test plan

- Test long chat names

bh79DWW overflow

- Test long handles

image

- Test emoji/non-latin script - should work


rafaeleyng (approved):
Nice!

PR #10632, opened 2026-05-27
Colocate all chat log event handlers

ds-boyce:
Moves the member events out of `list-convo-members.ts` and into `list-conversations.tsx`, which is always mounted. Follow-up to #10624.

Tested by adding and removing a member from a group chat in one window (owner), and confirming the events were handled in another (member).

### DIFF
--- a/src/state/queries/messages/list-conversations.tsx
+++ b/src/state/queries/messages/list-conversations.tsx
@@ -173,6 +173,49 @@ export function ListConvosProviderInner({
           )
         }
 
+        function handleMemberAdded(
+          convoId: string,
+          did: string,
+          relatedProfiles: ChatBskyActorDefs.ProfileViewBasic[],
+          rev: string,
+        ) {
+          const newMember = relatedProfiles.find(r => r.did === did)
+          if (!newMember) return
+          // If the optimistic add already added them, skip the
+          // memberCount bump to avoid double-counting.
+          const alreadyKnownMember =
+            queryClient
+              .getQueryData<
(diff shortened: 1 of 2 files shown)


src/state/queries/messages/list-conversations.tsx
```diff
@@ -138,6 +141,21 @@ export function ListConvosProviderInner({
...
+        ) {
+          queryClient.setQueryData<ChatBskyActorDefs.ProfileViewBasic[]>(
+            listConvoMembersQueryKey(convoId),
+            old => {
+              if (!old) return // query doesn't exist yet, skip
+              return fn(old)
+            },
+          )
+        }
+
```
mozzius, line 159: We should mutate `members` in `getConvo`/`listConvos` too

mozzius (approved):

src/state/queries/messages/list-conversations.tsx
ds-boyce: One thing to note: Since `members` is a curated subset of all conversation members, we can optimistically update it but the result may be different from what we get on the next server update.

ds-boyce:
@mozzius This added some complexity… can you take another look?

src/state/queries/messages/list-conversations.tsx
mozzius: I think that is fine. I don't think it affects anything if there's *too many*

mozzius:
Ouch yeah. @claude can I get a second set of eyes on this please

mozzius:
I think this is good feedback from claude. I can't see anything else

Issue #9708, opened 2026-01-16
Some issues with the Live Indicator beta


LooneyH:
### Steps to Reproduce

1. Go to your profile
2. Go to the "Go Live" composer.

### Attachments

<img width="491" height="283" alt="Error for 'Twitch.tv'" src="https://github.com/user-attachments/assets/5f60115c-ce0c-4d29-96de-13c60ed8f883" />
Android linking to 'Twitch.tv' 

<img width="491" height="477" alt="Working as expected for 'twitch.tv'" src="https://github.com/user-attachments/assets/78776cf9-8d95-41ac-a95b-4cc475147d14" />
Android linking to 'twitch.tv'

<img width="491" height="1024" alt="Duration falling behind navigation bar" src="https://github.com/user-attachments/assets/7d681a34-ffe4-4e42-8383-3e63bf35e0c4" />
Bottom of the duration list falling behind navigation bar

### What platform(s) does this occur on?

Web (Desktop), Android

### Device Info

Firefox 147.0 on macOS 26.2; Pixel 7 on Android 16

### What version of the app are you using?

1.114

### Additional Information

Currently there are a handful of notable issues with the Live Indicator beta that I thought I'd make note of (some of these I imagine are known and others not, and some are bug-related and user-experience related). I don't view any of these as critical, but improving these would definitely make the user experience nicer.

- ~~Currently on Android, if you type Twitch.tv instead of twitch.tv, it will refuse to work (this is not present on web, at least on desktop).~~
- ~~The lack of support for other major platforms such as YouTube is an obvious limitation for now, but the lack of integration to the Twitch API makes this feel very artificial.~~
- ~~Additionally, on Android the list of times is currently both lengthy and ends up falling behind the gesture bar at the bottom.~~
- ~~Not something I believe that can be helped and I know this is something that you're aware of already, but adblockers currently seem to block events.bsky.app and it isn't really well stated to users. I think a simple post on the main Bluesky account addressing this would help on that front, at least until it is whitelisted by adblockers more widely.~~
- The system for setting how long a stream lasts is janky and artificially limits the indicator to 4 hours (when people regularly stream for longer). I assume this is in place to limit abusing the indicator to constantly advertise your channel, but API integration would be able to resolve this issue by only showing when you are actually live.
- The lack of any oAuth with Twitch also presents an issue of users presenting links to channels that are not theirs.
[...]

PR #10636, opened 2026-05-28
Restore link in Go Live prompt if status record present

LooneyH:
Currently, every time a user wants to go live on Bluesky, they have to type in the link every single time. This PR checks if there is a valid record status record present for the user, restoring the last used link in the "Go live" prompt. This will help reduce friction for users and help increase adoption.


LooneyH:
Almost all other issues listed in https://github.com/bluesky-social/social-app/issues/9708 have since been fixed, so this would be another great one and a small but impactful QoL improvement

LooneyH:
Thought I'd add I tested this on accounts both with previous expired records and no records at all, and both cases worked fine (the former now autofilling as intended and the latter working as it currently does). Would love to see this merged as someone who can speak to how annoying it is to have to re-type the link every single time I go live

mozzius:
@claude review

mozzius:
Please can you remove the second commit. i18n compilation is done automatically

LooneyH:
Yeah sorry, didn't realise it had added the second commit until now, my bad! Should be sorted

LooneyH:
Lemme know if there's anything else that needs sorting at all

LooneyH:
@mozzius sorry to be a pain but just pinging in case this has been lost in all the recent feature updates (as it's passed all the tests)

estrattonbailey (approved):

PR #10642, opened 2026-05-28
Use `pnpm` to handle node/pnpm versions in Dockerfile

mozzius:
Bit of a proof-of-concept, but makes the Dockerfile much simpler. Also using `devEngines.runtime` for CI node version: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file


ds-boyce (approved):
Sick. Built and ran this locally and everything seems fine.

PR #10638, opened 2026-05-28
[Chat] Tweak styles in invite link dialog

mozzius:
- Tidy up styles
- Add utility functions for the joinLink stringification/unstringifications
- Only edit link when options change
- Replace `useEffect` with `useState`

PR #10644, opened 2026-05-28
Update copy for the join link description

ds-boyce:
Adds info about the current and total number of members.


mozzius (approved):

ds-boyce:
Looks like these changes are in #10638.

PR #10646, opened 2026-05-28
Refactor notification settings as dialogs

ds-boyce:
Primarily so we can invoke these from chat settings, but overall this feels like a better fit.

https://github.com/user-attachments/assets/36964110-095d-4df1-9e68-88aa3a238f80


mozzius:
I think on native, a nice `large` `secondary` "Done" button at the bottom of the sheet would make it feel nice/more balanced

ds-boyce:
@mozzius Added!

src/components/dialogs/NotificationSettingsDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+              allowDisableInApp={allowDisableInApp}
+            />
+          )}
+        </SettingsList.Container>
+        <Dialog.Close />
+        {IS_NATIVE ? (
+          <Button
+            color="secondary"
+            size="large"
+            label={l`Click here to close this dialog`}
```
mozzius:
```suggestion
            label={l`Close dialog`}
```

mozzius (approved):
LGTM! Looks good.

Minor cleanup thing would be to remove the routes from bskyweb too

PR #10650, opened 2026-05-29
respect autoplay settings in gif dialog

vineyardbovines:
If a user has autoplay gifs turned off in settings, ensure the GIF dialog isn't autoplaying them. 

Instead, play GIFs on hover (web) or on long press (native). Tap/click still selects a GIF for the composer.

https://github.com/user-attachments/assets/6b7b3c91-82c4-4550-bac8-48d9e8854103

https://github.com/user-attachments/assets/a5747a22-331c-45df-8b86-b8f9f16d8760

### DIFF
--- a/src/features/gifPicker/components/GifPickerItem.tsx
+++ b/src/features/gifPicker/components/GifPickerItem.tsx
@@ -1,9 +1,12 @@
+import {useState} from 'react'
 import {Image} from 'expo-image'
 import {useLingui} from '@lingui/react/macro'
 
+import {useAutoplayDisabled} from '#/state/preferences'
 import {atoms as a, useTheme} from '#/alf'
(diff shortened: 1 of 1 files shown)


surfdude29:
I have autoplay turned off because I don't want videos autoplaying in my feed, but that doesn't mean I want to see a wall of frozen GIFs when I open the picker 😅

Also, I think some users who have autoplay turned off might assume the app is broken if GIFs no longer animate in the picker, because they might not expect the autoplay setting to affect that.

However, I agree that of course there does need to be a way of preventing GIFs from autoplaying in the dialog for those who need it.

Maybe the best way forward is to provide separate autoplay prefs for videos and GIFs, as this PR suggests? 🙏

- #8234

PR #10651, opened 2026-05-29
Fix stale display names for lists/feeds in top bar across restarts/devices

### LINKED ISSUE #9939: Top Bar stuck using cached List names with no way to trigger a refresh
### Steps to Reproduce

1. Make a list
2. Pin that list to your feeds
3. Update the name of that list to something else

In the Feeds tab and top bar, the name of the list gets stuck with the old name instead of updating to the new value

### Attachments

_No response_

iOS, Web (Desktop)

### Device Info

_No response_

Build version: 1.117.0.1317; Bundle info: cb7e2ab (prod); Bundle date: 26022020; Platform: ios; Platform version: 26.3; Device ID: c911b27d-b828-4207-9749-dc4fa33b566a

### Additional Information

On mobile, getting the Feeds tab to recognise the new name is as simple as manually refreshing it. (closing the app doesn't do anything)

But there's no way to get the Top Bar to refresh in the same way. Adding a new feed to your saved feeds does cause the name to update, but upon removing that new feed, it changes back (and if you do this multiple times, the affect infects the combination of feeds you used with that version of the name too)

The same issue occurs on web as well, but you can't refresh the Feeds tab like you can on mobile and you have the right nav on top of that

### LINKED ISSUE #10440: Changing a List name is not reflected in pinned Feeds atop Home
### Steps to Reproduce

1. Create a List with at least one account.
2. Name it 'Test List 1'.
3. Pin it to Feeds.
4. Note the name in Feeds at top of Home.
5. Change the name of this List to 'Test List 001' via Settings-> Lists or Home->Feeds(#)->Settings (Pinned Feeds).
6. Note that the name remains 'Test List 1' in Feeds at top of Home and Home->Feeds(#)->Settings (Pinned Feeds) while appearing as 'Test List 001' in Settings->Lists.
7. Un-pinning and re-pinning does not resolve, nor does restarting the app.

### Attachments

_No response_

iOS

### Device Info

iPhone 16 Pro Max

1.121.0

### Additional Information

_No response_

RetroSunstar:
Currently the names of pinned items in the top bar are not refreshed or refreshable by the end user*

#10466 made a change to invalidate the cache but this only happens when you change the name of a list on the same device, causing cross-devices to remain stale

This PR invalidates ``FEED_INFO_RQKEY_ROOT`` once the app finishes loading and sets ``staleTime`` to one hour, stopping Feed/List names from being indefinitely cached when using ``persistedVersion`` with ``staleTime: INFINITY``

Originally I assumed this to be intentional and was considering only providing a way to trigger a refresh by opening the Feeds menu (as it already retrieves up-to-date names), but the first line of #10466 reads as if the cache does update on restarts. This is not the behaviour I could see when testing it myself, but I'm giving it a shot at being implemented before anything else 😇

I'd be more than happy to try another approach! 🐙

Fixes #9939 & combined with #10466 entirely Fixes #10440


mozzius:
No, this is a load-bearing query for the initial startup of the app. Why not just reduce the staletime to something finite?

RetroSunstar:
> No, this is a load-bearing query for the initial startup of the app. Why not just reduce the staletime to something finite?

Because this gives the user a way to trigger an infomation update in an expected way, setting the time to a finite value would work in fixing it eventually but still lets the values remain stale for that period of time without a way to trigger it faster, no? 

If it has to be a persistent piece of data in the initial startup, couldn't you simply invalidate the cache once the app finishes loading? 

Otherwise sure, we can just set it to something finite! 🐙

estrattonbailey:
Once the data is cached, it'll load fast and can invalidate after a reasonable period of time. I think the simpler approach here is better: just a short finite stale time, maybe 15 min?

estrattonbailey:
@RetroSunstar mind just merging upstream so we can let CI run? Hardly need it for this but it otherwise can gum things up :P

PR #10654, opened 2026-05-29
Fix text selection for multiline alt text in lightbox

mozzius:
Just a clash with the press interaction

https://github.com/user-attachments/assets/49fa6465-7ffa-4819-8456-cd8b0b820c4f


estrattonbailey (commented):
Doesn't appear to work on my Android device

mozzius:
Superceded by #10698

PR #10664, opened 2026-05-31
New layout for expanded notification bundles

abenzer:
## What
  
Reworks the layout for bundled notifications (likes, reposts, follows, etc.),  gated behind the `notifications:expanded_profile_card:enable` feature flag, plus a couple of bug/performance fixes that are not gated.
  
### Expanded profile card treatment (gated)
- When you expand a grouped notification, authors now render as profile cards with follow buttons.
- On web, only show the cell hover highlight when hovering an area that actually opens the post, so hovering the authors list no longer implies the whole cell is clickable. 
    
### iOS expand performance (gated)
- The native expand/collapse animation now uses a native-driver opacity fade instead of a JS-thread height animation (`useNativeDriver: false`). This fixes the jittery/janky animation when expanding notifications in the current iOS app, since the animation no longer runs on the JS thread.
    
### Timestamp tooltip clipping (not gated) 
- Fixes the hover timestamp tooltip getting cut off on notifications. The cell previously used `overflow: hidden` (added in #4899 to contain long handles), which also clipped the tooltip that renders below the row. We now clip only horizontal overflow on web (`overflowX: 'clip'`) so the tooltip can spill below while keeping the long-handle guard. Native is unchanged. 
- Here's how it looked before. We don't want it to clip like this:
Screenshot 2026-05-31 at 12 17 59 AM
    
## Test plan
- [x] Web and native: expand a bundled notification — you should see follow buttons on the profile cards. 
- [x] Web: hover a timestamp on a short notification (e.g. a follow) — full date tooltip shows fully, not clipped. 
- [x] Web: hover the authors list vs. the rest of the cell — the bg color highlight should only happen on areas where clicking would take you to the post.  
- [x] iOS: expand a bundled notification — smooth fade, no jitter.


surfdude29:
Regarding the notification timestamp tooltip being cut off, it looks like this PR will fix these issues:
- #6888
- #8050

And supersede these PRs:
- #6892
- #8058

src/view/com/notifications/NotificationFeedItem.tsx
```diff
@@ -981,40 +1013,120 @@ function CondensedAuthorsList({
...
+  const targetHeight = profileCardEnabled
+    ? measuredHeight
+    : authors.length * (EXPANDED_AUTHOR_EL_HEIGHT + 10) /*10=margin*/
+  const heightStyle = {
+    height: Animated.multiply(heightInterp, targetHeight),
+  }
   return (
     <Animated.View style={[a.overflow_hidden, heightStyle]}>
-      {visible &&
+      {profileCardEnabled ? (
```
ds-boyce: Should we keep the `visible` check here so we’re not unconditionally fetching profile information?

estrattonbailey: Good call, doesn't hurt here so let's keep it. Added 👍

estrattonbailey (approved):
LGTM

PR #10671, opened 2026-06-01
ensure alt text is scrollable

vineyardbovines:
at larger text sizes, alt text on images wasn't scrollable / was taking the entire screen. now, we ensure it can be scrolled to keep the same visual pattern


estrattonbailey (commented):
Can we fix web too while we're at it? Works on native (tested on iOS)

vineyardbovines:
CleanShot 2026-06-04 at 11 22 59@2x

estrattonbailey (approved):

PR #10672, opened 2026-06-01
fix(ios): fix Hermes startup race in ExpoBridgeModule (patch)

mozzius:
## Summary

Backports [expo/expo#45374](https://github.com/expo/expo/pull/45374) into our `expo-modules-core@3.0.30` patch to fix a recurring intermittent **launch crash** on iOS.

TestFlight shows a cluster of `EXC_BAD_ACCESS (SIGSEGV)` crashes on cold start, landing at varying Hermes frames (`HadesGC::mark`, `prepareRuntime` → `defineProperty`, `initializeRuntime` → `bindNativePerformanceNow`). The varying crash site is the signature of heap corruption from a data race, not a single null-deref.

## Root cause

We run the **legacy bridge** (`newArchEnabled: false`, so `RCTRuntimeExecutor.h` is absent and the `#if !__has_include(...)` branch compiles):

1. `EXNativeModulesProxy` defaults to `requiresMainQueueSetup = YES`, so its `setBridge:` runs on the **main thread** and lazily loads `ExpoBridgeModule` there. RN's lazy module-load path ignores `ExpoBridgeModule`'s own `+requiresMainQueueSetup = NO`.
2. `ExpoBridgeModule.setBridge:` assigns `_appContext._runtime`, whose Swift `didSet` synchronously calls `prepareRuntime()` — JSI mutations, **on the main thread**.
3. Meanwhile the **JS thread** is inside `JSIExecutor::initializeRuntime()` doing its own JSI mutations.

Two threads mutating the same Hermes runtime with no synchronization → Hades GC write barrier sees an inconsistent pointer → crash wherever the GC/allocator next touches the poisoned heap.

## Fix

Hop the runtime install (and the `prepareRuntime()` chain it triggers) onto `RCTJSThread` via `dispatchBlock:queue:`, so all JSI mutation is serialized on the JS thread. Weak captures avoid extending object lifetimes; the `_runtime == nil` guard prevents a double-install if the blocking-sync `installModules` path (also on the JS thread) got there first.

Scoped to the `#if !__has_include(<ReactCommon/RCTRuntimeExecutor.h>)` branch, so it's a no-op under new arch / bridgeless.

## Notes / risk

- `dispatchBlock:queue:` and `RCTJSThread` confirmed present on our RN 0.81 (`RCTCxxBridge.mm`).
- Upstream's real fix ([#44351](https://github.com/expo/expo/pull/44351)) removes `ExpoBridgeModule` entirely in SDK 55+ — **drop this patch on that upgrade**.
- One observable behavior change: code reading `_appContext._runtime` synchronously inside a `setBridge:` chain now sees `nil`. No standard/our modules appear to do this.

## Testing

- [ ] Builds on iOS (needs a real build — patched pod, not covered by JS checks)
- [ ] Optional: reproduce via the semaphore harness from #45374 (gate `initializeRuntime()` with a `usleep`) to confirm before/after
- [ ] Watch TestFlight launch-crash rate on next beta

Refs: expo/expo#43003, expo/expo#45374, expo/expo#44351

### DIFF
--- a/patches/expo-modules-core@3.0.30.patch
+++ b/patches/expo-modules-core@3.0.30.patch
@@ -13,3 +13,29 @@ index 47c4d15f6b10bbd77858cfff425cda9a618735b9..afe138d22d566244482498a7be0e14b8
    // Check for Content-Type
    val skipContentTypes = listOf(
      "text/event-stream", // Server Sent Events
+diff --git a/ios/Core/ExpoBridgeModule.mm b/ios/Core/ExpoBridgeModule.mm
+index 2ed1c00f47406e109750cc27ace7e0d88e42c00e..2d6ba8a0323581f05ac5a83089c64ab69211e038 100644
+--- a/ios/Core/ExpoBridgeModule.mm
++++ b/ios/Core/ExpoBridgeModule.mm
+@@ -46,7 +46,20 @@ - (void)setBridge:(RCTBridge *)bridge
+   _appContext.reactBridge = bridge;
+ 
+ #if !__has_include(<ReactCommon/RCTRuntimeExecutor.h>)
+-  _appContext._runtime = [EXJavaScriptRuntimeManager runtimeFromBridge:bridge];
++  // Hop the runtime install (and the prepareRuntime() chain it triggers via
++  // _runtime.didSet) onto RCTJSThread. The original line ran synchronously
++  // on whatever thread called setBridge: - typically the main thread - and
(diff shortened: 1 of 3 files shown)


mozzius:
@claude review

vineyardbovines (approved):
word, good find. think it might be worth dropping in a comment or noting somewhere that this patch can be removed once we get onto SDK 55

PR #10673, opened 2026-06-01
Fix Android crash in chat keyboard scroll

mozzius:
## Summary

Fixes an intermittent Android production crash in the DM message list:

```
com.facebook.jni.CppException: Exception in HostFunction: Value is null, expected a number
  at _scrollToPaper (native)
  ...
  at handleAndFlushAnimationFrame_reactNativeReanimated_coreTs1
  at KeyboardAnimationCallback.onProgress (KeyboardAnimationCallback.kt:274)
```

## Root cause

`react-native-keyboard-controller`'s `useExtraContentPadding` (used by `KeyboardChatScrollView` in `MessagesList.tsx`) defers its `scrollTo` via `requestAnimationFrame`:

```ts
requestAnimationFrame(() => {
  scrollTo(scrollViewRef, 0, target, false)
})
```

There's a gap between scheduling the frame and it firing. If the `ScrollView` detaches in that window (navigating away from the convo, or the list re-creating its `renderScrollComponent`), `scrollViewRef()` resolves to `null` when the frame finally runs.

On the old architecture (`newArchEnabled=false`), reanimated binds `scrollTo` to `scrollToPaper`, which hands the resolved tag straight to the native `_scrollToPaper` HostFunction. That function expects a `number`, so a `null` tag throws `Value is null, expected a number`. The `handleAndFlushAnimationFrame` frame in the trace confirms it's the deferred path.

## Fix

Re-check the ref inside the rAF callback, right before scrolling. This is done in the existing `react-native-keyboard-controller` patch (the same `useExtraContentPadding` file we already patch).

Scoped to the deferred path that actually crashes, rather than swallowing nulls in reanimated's shared `scrollTo` primitive.

## Test plan

- [ ] Open a DM conversation on Android, focus the input, then navigate back / switch conversations while the keyboard is animating - no crash
- [x] `pnpm typecheck`

### DIFF
--- a/patches/react-native-keyboard-controller@1.21.8.patch
+++ b/patches/react-native-keyboard-controller@1.21.8.patch
@@ -1,5 +1,5 @@
 diff --git a/src/components/KeyboardChatScrollView/useExtraContentPadding/index.ts b/src/components/KeyboardChatScrollView/useExtraContentPadding/index.ts
-index 0f6d7c67a307885310ab184fdf9e7a5c7b296825..1e0bdd5b1d3b1eceb47bfb0050da84061fd7f77c 100644
+index 0f6d7c67a307885310ab184fdf9e7a5c7b296825..1a01093e7909973cef5268f999158df389e77634 100644
 --- a/src/components/KeyboardChatScrollView/useExtraContentPadding/index.ts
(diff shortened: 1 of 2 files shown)


ds-boyce (approved):
Seems legit.

PR #10676, opened 2026-06-01
Fix video size error showing 100 MB for users with 300 MB upload limit

surfdude29:
> [!IMPORTANT]
> **Claude Code analyzed this issue and wrote the PR and much of the description.**
>

## Issue

If a user has the `LargeVideoUploads` flag enabled and tries to upload a video file larger than 300 MB on web, they get this confusing error message referring to the 100 MB limit:

IMG_8975

A selected video's file size is validated in three places:

1. **Compression** (`getCompressErrorMessage`) ✅ flag-aware
2. **Upload** (`getUploadErrorMessage`) ✅ flag-aware
3. **Selection time** (`SelectMediaButton.tsx`) ❌ **not** flag-aware

This PR is a follow-up to #10497 to update the third place.

The confusing error message comes from the earliest check, which runs the moment you pick a file on web. In `src/view/com/composer/SelectMediaButton.tsx`, `processImagePickerAssets` had two problems:

- The size threshold compared against `VIDEO_MAX_SIZE` (300 MB) unconditionally, ignoring the feature flag entirely.
- The `FileTooBig` error string was **hardcoded** to `Maximum size is 100 MB.`

So a flag-enabled user picking a video over 300 MB correctly tripped the 300 MB threshold but got a message that said 100 MB. (As a bonus bug, users *without* the flag were allowed to select up to 300 MB at this stage, only failing later during compression/upload.)

The root cause is that `SelectMediaButton` had no access to the `large_video_uploads` flag — it was only passed to the later compression/upload steps.

## Fix

Threaded the existing `LargeVideoUploads` flag into the selection-time validation:

- **`SelectMediaButton.tsx`** — added an `enableLargeVideoUploads` prop, used it to pick `VIDEO_MAX_SIZE` (300 MB) vs `VIDEO_MAX_SIZE_REDUCED` (100 MB) for the threshold, and interpolated the matching number into the error message (`Maximum size is ${videoSize} MB.`), mirroring the wording approach already used in `video.ts`.
- **`Composer.tsx`** — `ComposerFooter` (which renders the button) now reads the flag via the existing `useAnalytics()` hook and passes it down.

This fixes both the misleading message and the threshold mismatch for non-flagged users.

## Test plan (on web)

- With the flag **off**, a 100–300 MB video is now rejected at selection with "Maximum size is 100 MB"
- With the flag **on**, a 100–300 MB video is accepted and a >300 MB video shows "Maximum size is 300 MB".

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1917,6 +1917,10 @@ function ComposerFooter({
   const t = useTheme()
   const {t: l} = useLingui()
   const {gtPhone} = useBreakpoints()
+  const ax = useAnalytics()
+  const enableLargeVideoUploads = ax.features.enabled(
+    ax.features.LargeVideoUploads,
+  )
   /*
    * Once we've allowed a certain type of asset to be selected, we don't allow
    * other types of media to be selected.
@@ -2033,6 +2037,7 @@ function ComposerFooter({
                 selectedAssetsCount={selectedAssetsCount}
                 onSelectAssets={onSelectAssets}
                 autoOpen={openGallery}
+                enableLargeVideoUploads={enableLargeVideoUploads}
               />
               <OpenCameraBtn
                 disabled={media?.type === 'images' ? isMaxImages : !!media}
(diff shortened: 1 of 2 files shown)


mozzius:
Superceded by #10683

PR #10689, opened 2026-06-02
Use expo/expo-github-action v9

ds-boyce:
Targeting `@main` seems fragile. Switches to `@v9` to get the [Node.js v24 update](https://github.com/expo/expo-github-action/issues/351#issuecomment-4600812901).


mozzius (approved):
Ah nice, finally. iirc they had a bunch of unpublished changes sitting on main

PR #10690, opened 2026-06-02
Apply per-use-case image size configs

### LINKED ISSUE #10725: Profile banner upload fails with raw validation error when image exceeds size limit
### Steps to Reproduce

1. Open Bluesky profile.
2. Click "Edit profile".
3. Upload a banner image larger than 1 MB.
4. Save the profile.

Result:
A raw validation error is displayed:

"Invalid app.bsky.actor.profile record: blob too big (maximum 1000000, got 1991913) at $.record.banner"

Expected:
The app should either:
- Prevent selecting oversized images before upload, or
- Automatically compress/resize the image, or
- Display a user-friendly error message explaining that the banner exceeds the 1 MB limit and how to fix it.

### Attachments

[Screenshot 2026-06-04 at 11.15.05 AM.png](https://github.com/user-attachments/assets/59480a36-c00f-4448-ac64-3d568e0d34dd)

Web (Desktop)

### Device Info

macOS, Google Chrome

Bluesky Web App (accessed through Google Chrome on macOS)

### Additional Information

The upload itself appears to be rejected correctly because the image exceeds the allowed size limit.

The issue is that the application surfaces an internal validation error directly to the user:

> Invalid app.bsky.actor.profile record: blob too big (maximum 1000000, got 1991913) at $.record.banner

This message exposes implementation details such as the record type and schema path (`$.record.banner`).

A more user-friendly error message would improve the experience, for example:

> Banner image is too large. Please upload an image smaller than 1 MB.

Ideally, the client should validate file size before upload and clearly communicate the limit to the user.

estrattonbailey:
Replaces the hardcoded image limits in the compression/resize helpers with two explicit, named size configs passed in at each call site:

- `IMAGE_SIZE_CONFIG_POSTS` (4000px / 2MB) for post images uploaded via the composer and images pasted/typed as a URL. The pasted-URL path previously pre-shrank to 2000px / 1MB before `compressImage` ran, capping quality; it now matches the post limits.
- `IMAGE_SIZE_CONFIG_2K_1MB` (2000px / 1MB) for avatars, banners, onboarding, link-card thumbnails, and the e2e picker.

`compressImage`, `compressIfNeeded`, and `getResizedDimensions` no longer carry default constants - callers pass the appropriate config directly so the size policy lives at the call site.

**Web parity:** web `doResize` previously didn't clamp dimensions (`mode: 'stretch'`, quality-only re-encode). It now clamps via the shared `getResizedDimensions` like native does. This affects web callers of `compressIfNeeded`/`downloadAndResize` - onboarding profile image, link-card thumbnails, and pasted-URL downloads - which now downscale oversized images instead of only re-encoding.

Note: web avatar/banner edits go through the shared `compressImage` and remain at 2000px / 1MB, same as production (`main`).

Fixes #10725


src/lib/media/util.ts
```diff
@@ -2,6 +2,28 @@ export function extractDataUriMime(uri: string): string {
   return uri.substring(uri.indexOf(':') + 1, uri.indexOf(';'))
 }
 
+export function getResizedDimensions(
```
estrattonbailey, line 5: Just moved this to a shared file

src/lib/constants.ts
```diff
@@ -96,7 +96,13 @@ export const STAGING_FEEDS = [
...
 ]
 
-export const POST_IMG_MAX = {
+export const IMAGE_SIZE_CONFIG_POSTS = {
+  width: 4000,
+  height: 4000,
+  size: 2000000,
+}
+
+export const IMAGE_SIZE_CONFIG_2K_1MB = {
```
estrattonbailey, line 105: Left "posts" as a semantic suffix, but the other is shared between concerns so used a spec suffix

mozzius:
@claude please review

mozzius:
@claude please rereview

surfdude29:
I think this fixes #10725

estrattonbailey:
@surfdude29 yes it should!

PR #10697, opened 2026-06-02
Paginate group chat members list

ds-boyce:
This avoids hard-coding the limit to 50 in the client so that we can potentially increase it later without an app update.

### DIFF
--- a/src/state/queries/messages/list-convo-members.ts
+++ b/src/state/queries/messages/list-convo-members.ts
@@ -1,5 +1,12 @@
-import {type ChatBskyActorDefs} from '@atproto/api'
-import {type QueryClient, useQuery} from '@tanstack/react-query'
+import {
+  type ChatBskyActorDefs,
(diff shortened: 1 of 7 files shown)


src/state/queries/messages/list-convo-members.ts
```diff
@@ -22,42 +30,41 @@ export function useListConvoMembersQuery({
...
-      } while (cursor)
-
-      return members
+    queryFn: async ({pageParam}) => {
+      const {data} = await agent.chat.bsky.convo.getConvoMembers(
+        {convoId, cursor: pageParam as string | undefined, limit: LIMIT},
+        {headers: DM_SERVICE_HEADERS},
+      )
+      return data
     },
```
mozzius, line 41: Give we already handled >50 members by eagerly paginating, is it worth the extra complexity?

ds-boyce: Eh, I suppose not. Closing.

PR #10698, opened 2026-06-02
Fix alt text selection in lightbox on Android (and iOS)

vineyardbovines:
## Why

Selecting alt text in the image lightbox regressed in #10330 (lightbox redesign). That refactor moved the tap-to-expand `onPress` off the `<Text>` and onto a wrapping `<Pressable>`, leaving the alt text a bare `<Text selectable>` inside a touchable.

#10654 added an empty `onLongPress` to the `Pressable`, which fixes selection on **iOS only** - iOS's native `UITextView` selection recognizer can coexist with the parent touchable. On **Android**, selection is the native `TextView`'s long-press, and the parent touchable consumes that gesture before the `TextView` ever sees it, so the noop on the `Pressable` does nothing. Broken on a real Android device.

## Fix

Restore the pre-#10330 structure: put `onPress` + the (intentional, documented) noop `onLongPress` back on the selectable `<Text>` and demote the wrapper to a plain `<View accessibilityRole="button">`. ALF's `Text` forwards these to the native node - `UITextView` on iOS (whose Swift handles `onPress`/`onLongPress` on its own recognizers) and plain `RNText` on Android - so one node owns both gestures on both platforms.

RN has no public API to arbitrate tap vs. native text selection on a single node, so the empty `onLongPress` is the supported workaround; a comment now explains why so it doesn't read as dead code.

## Test plan
- [ ] iOS: open image lightbox, tap alt text to expand, long-press to select/copy
- [ ] Android: same - confirm long-press now selects text
- [ ] Both: tap still toggles expand/collapse; scroll-to-dismiss unaffected


estrattonbailey (approved):
Tested on all platforms, works for me. Just needs a rebase 👍

PR #10701, opened 2026-06-03
Use square icons for labellers in Automation Label preview card

RetroSunstar:
Currently the Automation Label settings page will always use a circular icon for the example display

This doesn't really fit as every other instance across the app will use square borders for labellers instead 🐙


mozzius (approved):
Good catch

PR #10704, opened 2026-06-03
Peek preview for external link embeds

mozzius:
## What

Long-pressing an external link embed on iOS now peeks a **live preview of the page** in an in-app browser. When the user has "open links in app" enabled, tapping the peek morphs seamlessly into the full browser; otherwise it hands off to the normal link flow.

Built on the new `link` preview type in `@bsky.app/peek-menu` (bumped to `^0.3.0`). See https://github.com/bluesky-social/toolbox/pull/49

## Video

https://github.com/user-attachments/assets/035ced90-4f2a-4590-b102-2e28b65df8bf

## How it works

- New `peek?: boolean` prop on the modern `Link` component (iOS-only, external web links only). When set, `Link` wraps its `Button` in a peek menu.
- The in-app-browser preference drives the commit behavior:
  - **on** → native morph into the in-app browser (`useInAppBrowser: true` on the peek)
  - **off / unset** → the tap fires `onPreviewPress`, which runs the existing `openLink` flow (so the consent dialog still shows when the pref is unset)
- The RN `onLongPress` is disabled while peeking so it doesn't fight the native context-menu gesture (that conflict was cancelling the tap and breaking the lift animation).
- URL is proxied (`createProxiedUrl`) when `shouldProxy`, browser tints come from the theme, and the peek's lift `borderRadius` is derived from the link's own style.
- The peek menu carries a **Share** action.

## Scope

- Enabled on `ExternalEmbed`. Also made that card opaque (`t.atoms.bg`) — it was transparent, which made the peek lift look like a floating shadow.
- **Not** enabled on `StandardSiteEmbed`: its tap target is an absolutely-positioned overlay rather than a wrapper, so peeking it would require restructuring the card layout. Left for a follow-up.
- iOS only; no-op on Android/web (peek-menu falls through to a plain wrapper).


estrattonbailey:
Should we add this to standard site cards too?

mozzius:
Would be nice to, however I wasn't sure which element to wrap, seems like the actual links is absolutely positioned over the card?

estrattonbailey (approved):

PR #10706, opened 2026-06-03
Remove `react-native-dotenv`

mozzius:
[It's incompatible with `react-native-worklets`](https://github.com/goatandsheep/react-native-dotenv/issues/565), and we don't even use it (we just use the build-in Expo env var support), so let's bin it

## Test plan

Sanity check me and confirm we don't use any env vars that arent' prefixed `EXPO_PUBLIC_`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -19,19 +19,6 @@ module.exports = function (api) {
       '@lingui/babel-plugin-lingui-macro',
       ['babel-plugin-react-compiler', {target: '19'}],
       [
-        'module:react-native-dotenv',
-        {
-          envName: 'APP_ENV',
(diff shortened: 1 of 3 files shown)


estrattonbailey (commented):
Pry fine but let's wait for early next release cycle

mozzius:
Ye no problem, just extracting it out of the larger New Arch PR

tjzel:
fyi:

- https://github.com/software-mansion/react-native-reanimated/pull/9703

mozzius:
nice, thanks @tjzel!

vineyardbovines (approved):

PR #10709, opened 2026-06-03
Reduce iOS memory pressure in lightbox, composer, GIF picker

vineyardbovines:
## Summary

Three changes aimed at reducing iOS `WatchdogTermination` (jetsam) crashes. Hypothesis-driven from breadcrumb analysis of 10 recent events; needs TestFlight validation before claiming rate impact.

**Lightbox pager windowing** (`src/components/Lightbox/pager/ImagePager.tsx`)
Multi-image post lightbox previously mounted every page's `LightboxImage` simultaneously, so opening a 4-image post would kick off 4 fullsize decodes at once. Now renders only the current ±1 pages; far-away pages are placeholders. The current page is always rendered, so swipe-to-close and animations are unaffected. The neighbor at ±1 is pre-loaded so adjacent swipes are still instant; only swiping >1 page incurs a brief load.

Targets the pattern seen in events ending with `UIPageViewController` mount → 3-5 `feed_fullsize` fetches in <1s → kill.

**Global memory warning handler** (`src/App.native.tsx`)
Subscribes to RN's `memoryWarning` event (backed by `UIApplicationDidReceiveMemoryWarningNotification` on iOS) and calls `Image.clearMemoryCache()`. Currently we ignore these warnings entirely — one captured event shows 5 LOW_MEMORY breadcrumbs over 14 seconds before jetsam with no response from us. This gives the GC an explicit chance to reclaim before the OS does it for us.

**GIF picker item Image tweaks** (`src/features/gifPicker/components/GifPickerItem.tsx`)
Minor: `priority=\"low\"` + `recyclingKey={gif.id}`. This is _not_ the real fix the GIF picker needs — that requires restructuring the masonry from one big FlatList row into per-row virtualization, which is bigger than this PR. Worth its own follow-up issue.

## Known follow-ups

Things I deliberately left out of this PR:
- **GIF picker masonry virtualization** — the real fix for the GIF picker OOM pattern (event with ~30 animated GIFs in a BottomSheet).
- **Composer image downsampling** — PHPicker returns 12MP photos and we render the preview at ~250px; decoding the full resolution is wasteful.

## Test plan

- [ ] Open a 4-image post lightbox, swipe through all images, close. Confirm no visible regression.
- [ ] Open multi-image post lightbox on a low-memory device (e.g. iPhone 8/SE) — should no longer crash on open.
- [ ] Use composer with a picked image + a link card, leave open for 30s. No regression in preview rendering.
- [ ] Open GIF picker, scroll through search results. No regression in scroll perf or load times.

### DIFF
--- a/src/components/Lightbox/pager/ImagePager.tsx
+++ b/src/components/Lightbox/pager/ImagePager.tsx
@@ -249,8 +249,21 @@ function ImageView({
   const [isDragging, setIsDragging] = useState(false)
   const [showControls, setShowControls] = useState(true)
   const [isAltExpanded, setIsAltExpanded] = useState(false)
+  const [hasFullyOpened, setHasFullyOpened] = useState(false)
   const dismissSwipeTranslateY = useSharedValue(0)
   const isFlyingAway = useSharedValue(false)
+
+  // Hold the neighbor window at 0 until the open animation completes, so we
+  // only decode the active image during the most memory-constrained moment.
+  // Mirrors the orientation-unlock gate below.
+  useAnimatedReaction(
+    () => openProgress.get() === 1,
+    isOpen => {
+      if (isOpen) {
+        runOnJS(setHasFullyOpened)(true)
+      }
+    },
+  )
 
   const containerStyle = useAnimatedStyle(() => {
     if (openProgress.get() < 1) {
@@ -415,7 +428,7 @@ function ImageView({
         style={styles.pager}>
         {images.map((imageSrc, i) => (
           <View key={`${i}-${imageSrc.uri}`}>
-            {Math.abs(i - imageIndex) <= 1 ? (
+            {Math.abs(i - imageIndex) <= (hasFullyOpened ? 1 : 0) ? (
               <LightboxImage
                 onTap={onTap}
                 onZoom={onZoom}
--- a/src/components/Lightbox/pager/ImagePager.tsx
+++ b/src/components/Lightbox/pager/ImagePager.tsx
@@ -256,6 +256,9 @@ function ImageView({
   // Hold the neighbor window at 0 until the open animation completes, so we
   // only decode the active image during the most memory-constrained moment.
   // Mirrors the orientation-unlock gate below.
+  // Hold the neighbor window at 0 until the open animation completes, so we
+  // only decode the active image during the most memory-constrained moment.
+  // Mirrors the orientation-unlock gate below.
   useAnimatedReaction(
     () => openProgress.get() === 1,
     isOpen => {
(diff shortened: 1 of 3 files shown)


src/components/Lightbox/pager/ImagePager.tsx
```diff
@@ -383,22 +383,24 @@ function ImageView({
...
-              safeAreaRef={safeAreaRef}
-              isScaled={isScaled}
-              isFlyingAway={isFlyingAway}
-              isActive={i === imageIndex}
-              dismissSwipeTranslateY={dismissSwipeTranslateY}
-              openProgress={openProgress}
-              thumbRects={thumbRects}
-              imageIndex={i}
-            />
+            {Math.abs(i - imageIndex) <= 1 ? (
```
mozzius: Worth noting that there is already some sort of logic in there to prevent other images loading as the lightbox is animating open, might be cleaner to integrate with that system instead?

src/features/gifPicker/components/GifPickerItem.tsx
mozzius, line 50: Maybe we could use animated `webp` instead of `gif` to lower memory usage? GIFs are famously inefficient

From the Klipy docs:

Image

estrattonbailey: Solid idea

src/components/Lightbox/pager/ImagePager.tsx
vineyardbovines: Want to make sure I integrate with the right thing — which existing system did you have in mind? Looking through `LightboxImage`/`ImageItem.ios`, the `transforms.isHidden`/`isResting` signals track open-progress but only feed into `opacity` and the loader. The `<Image source>` itself still mounts as soon as PagerView pre-renders the neighbor page, so I don't see a pre-existing gate that prevents the decode during the open animation. If you're thinking of a specific signal I could thread into `LightboxImage` or `ImageItem` to gate the mount instead of the parent-side `hasFullyOpened` flag, happy to refactor — just want to make sure I'm hooking into what you meant.

src/features/gifPicker/components/GifPickerItem.tsx
vineyardbovines: Down to try this — could you drop the Klipy URL pattern for the animated webp variant? The image with the docs snippet isn't accessible to me here. The Klipy responses we get are normalized to the Tenor shape (`preview` / `gif` / `tinygif` / `mp4` / `webm`), so there's no `webp` field; I'd need to know whether we transform via an extension swap on the CDN path, a query param, or a different format slug. Once I know the pattern I can wire up a helper alongside `gifPreviewUrl` and switch the picker source over.

mozzius: https://static.klipy.com/ii/935d7ab9d8c6202580a668421940ec81/14/af/SE72470w.webp

PR #10711, opened 2026-06-03
Compile i18n before running lint

ds-boyce:
Fixes job failures like [this one](https://github.com/bluesky-social/social-app/actions/runs/26904787531/job/79366584557).


ds-boyce:
Reordering steps breaks the check. Interesting.

estrattonbailey (approved):

PR #10712, opened 2026-06-03
[Android] Fix horizontal swipes being cancelled when leaving bounds

mozzius:
All scrollviews wrapped in `BlockDrawerGesture` on android were really finnicky, because they'd stop the moment the swipe goes out of bounds. in particular, this is really painful for the pager tab bar.

The solution: turns out the RNGH docs are incorrect, `shouldCancelWhenOutside()` defaults to `true` on Android for some godforsaken reason. Solution: manually set it to `false`

https://docs.swmansion.com/react-native-gesture-handler/docs/legacy-gestures/pan-gesture/#shouldcancelwhenoutsidevalue-boolean


estrattonbailey (approved):
Image

PR #10713, opened 2026-06-03
Make extra sure the drawer gesture doesn't accidentally trigger on Android

mozzius:
Increasing the `activeOffsetX` makes extra sure it doesn't override the horizontal scrollviews

## Test plan

Scroll a horizontal scrollview very slowly. With this fix, the drawer should no longer open


estrattonbailey (approved):
Nice find!

PR #10715, opened 2026-06-03
Generate app icon types as postinstall step

ds-boyce:
Fixes failed jobs like [this one](https://github.com/bluesky-social/social-app/actions/runs/26910692412).


mozzius:
Any idea what changed to make this become a problem?

mozzius:
I think I have a better solution in #10724 by just loosening the types

PR #10718, opened 2026-06-03
Fix web carousel horizontal swipe triggering browser back/forward

vineyardbovines:
## Summary

On web, a horizontal trackpad swipe over the in-feed image carousel could trigger the browser's back/forward navigation instead of scrolling the carousel. This fixes that while leaving normal vertical page scroll and back/forward elsewhere untouched.

- **Chrome / Firefox:** add `overscroll-behavior-x: contain` to the carousel scroller, so horizontal overscroll no longer chains into the history-nav gesture (zero JS overhead).
- **Safari:** Safari ignores `overscroll-behavior`, so add a non-passive `wheel` handler on the scroller in `usePointerHandlers.web.ts`. It intercepts predominantly-horizontal wheel events, scrolls the carousel manually (clamped to bounds), and `preventDefault`s the history-nav gesture. Vertical-dominant events return early so page scroll is never swallowed. The active index/focus stays in sync, settling only when the index actually changes.

## Test plan

- [ ] **Chrome:** horizontal trackpad swipe over a >4-image carousel scrolls images, no browser back/forward (including swiping past the first/last image).
- [ ] **Safari:** same as above (this is the platform the wheel handler targets, including at the carousel boundaries).
- [ ] **Vertical scroll (both):** vertical two-finger scroll over the carousel still scrolls the page.
- [ ] **Back/forward elsewhere (both):** horizontal swipe outside a carousel still triggers browser back/forward.
- [ ] **Carousel still works (both):** mouse click-and-drag, keyboard arrows, and tap-to-open-lightbox all function; active image/focus correct after wheel scrolling.

> Note: the wheel handler is not Safari-gated, so on Chrome horizontal trackpad scrolling now routes through the manual scroll path (the `overscroll-behavior` style alone would suffice there). Behavior is correct and consistent cross-browser; flagging in case the scroll *feel* warrants Safari-gating later.

[APP-2287](https://linear.app/blueskyweb/issue/APP-2287/carousel-web-trackpad-horizontal-swipe-triggers-browser-backforward)

### DIFF
--- a/src/components/images/Gallery/index.tsx
+++ b/src/components/images/Gallery/index.tsx
@@ -340,6 +340,11 @@ export function Gallery({
               marginLeft: -insetLeft,
               width,
             },
+            // Prevent horizontal trackpad/wheel swipes from triggering the
+            // browser's back/forward overscroll-navigation gesture. Handles
+            // Chrome and Firefox; Safari is handled via the wheel listener in
+            // usePointerHandlers.web.ts since it ignores overscroll-behavior.
+            web({overscrollBehaviorX: 'contain'}),
           ]}
           contentContainerStyle={{
             gap: ITEM_GAP,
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):
Nice one! TIL `overscrollBehaviorX: 'contain'`. Good shout on Safari gating, I think that's the move.

PR #10719, opened 2026-06-03
Show image count badge on in-feed carousel

vineyardbovines:
## Summary

A top complaint about the carousel rollout was that it's hard to tell how many images are in a set, especially when the next image isn't peeking from the right edge. This adds a persistent "{n}/{total}" badge (e.g. "3/10") to the top-right of the in-feed carousel, visible on all platforms.

- Mirrors the existing `currentIndexRef` into reactive state, updated only when the index actually changes — the badge stays live without re-rendering on every scroll frame.
- Styling follows the Lightbox `PagerDots` pattern (white bold text, rounded). It sits on a translucent dark pill so it stays legible over arbitrary in-feed images (PagerDots itself can skip the backing because the lightbox provides a dark scrim).
- Hidden in quote posts (`hideBadges`) and for single images.
- Accessibility: a web `aria-live="polite"` region announces position changes; native screen-reader users continue to get the existing per-image stacked layout (each image already labelled "Image N of M"), so the carousel branch isn't reached there.

## Test plan

- [ ] iOS / Android / web: post with >1 image shows the "1/10" badge top-right, updating to "2/10", "3/10"… as you swipe/scroll.
- [ ] Badge does not appear on single-image posts or quote-post galleries.
- [ ] Web: count updates via keyboard arrows and click-drag.
- [ ] VoiceOver/TalkBack: per-image position is announced.

> No Figma design exists for the badge yet; styling follows the PagerDots pattern by decision. Placement/treatment can be refined when designs land.

[APP-2285](https://linear.app/blueskyweb/issue/APP-2285/carousel-show-numbering-badges-eg-310-on-photos)

### DIFF
--- a/src/components/images/Gallery/index.tsx
+++ b/src/components/images/Gallery/index.tsx
@@ -490,21 +490,22 @@ function GalleryImage({
               style={[
                 a.absolute,
                 a.justify_center,
+                a.rounded_sm,
+                a.p_xs,
+                t.atoms.bg_contrast_25,
                 {
-                  top: a.p_sm.padding,
-                  right: a.p_sm.padding,
-                  paddingHorizontal: 6,
-                  paddingVertical: 2,
-                  borderRadius: 999,
-                  backgroundColor: 'rgba(0, 0, 0, 0.6)',
+                  top: a.p_xs.padding,
+                  right: a.p_xs.padding,
+                  opacity: 0.8,
+                },
+                largeAltBadge && {
+                  padding: 6,
                 },
               ]}>
               <Text
                 style={[
                   a.font_bold,
-                  a.text_xs,
-                  a.leading_tight,
-                  {color: '#fff'},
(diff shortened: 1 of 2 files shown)


src/components/images/Gallery/index.tsx
```diff
@@ -483,6 +483,34 @@ function GalleryImage({
...
+              style={[
+                a.absolute,
+                a.justify_center,
+                {
+                  top: a.p_sm.padding,
+                  right: a.p_sm.padding,
+                  paddingHorizontal: 6,
+                  paddingVertical: 2,
+                  borderRadius: 999,
+                  backgroundColor: 'rgba(0, 0, 0, 0.6)',
```
estrattonbailey: Let's match the color/size spec from the crop/alt badges
Image

estrattonbailey (approved):
Otherwise LGTM!

PR #10726, opened 2026-06-04
[Chat] Surface blocked members to top of group member list for owner

mozzius:
Closes APP-2246.

When you own a group chat, members you've blocked (or who have blocked you) now sort to the top of the member list, are tagged with a "Blocked" indicator, and get an inline **Remove** button to kick them quickly.

Simulator Screenshot - iPhone 17 - 2026-06-04 at 16 20 36

## Changes
- `ConversationSettings/index.tsx` — owner-only sort tier that floats blocked-or-blocking members to the top (below the owner/self rows).
- `Member.tsx` — for the owner, blocked rows show a destructive inline **Remove** button (in place of the Follow link) that opens a confirmation prompt before calling the existing `useRemoveFromGroupChat` mutation.
- `prompts.tsx` — new `RemoveMemberPrompt` confirmation dialog.

The existing "Remove from chat" 3-dot menu item is unchanged.

## Notes
- First pass intentionally sidesteps the profile-shadow cache: the parent re-sort runs over raw query data, so it settles on refetch/remount rather than instantly when you block someone mid-session. The inline button itself is reactive (the row's profile is shadowed). A follow-up could subscribe the list to shadow block updates (à la `usePostAuthorShadowFilter`) for live re-sorting.

## Test plan
- [ ] As group owner, block a member -> they move to the top with a "Blocked" tag and a Remove button
- [ ] Tap Remove -> confirmation prompt -> member is removed
- [ ] Non-owners see no change

### DIFF
--- a/src/screens/Messages/ConversationSettings/Member.tsx
+++ b/src/screens/Messages/ConversationSettings/Member.tsx
@@ -7,18 +7,22 @@ import {createSanitizedDisplayName} from '#/lib/moderation/create-sanitized-disp
 import {logger} from '#/logger'
 import {useProfileShadow} from '#/state/cache/profile-shadow'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
+import {useRemoveFromGroupChat} from '#/state/queries/messages/remove-from-group'
 import {useProfileFollowMutationQueue} from '#/state/queries/profile'
 import {useRequireAuth, useSession} from '#/state/session'
 import {atoms as a, native, useTheme, web} from '#/alf'
+import {Button, ButtonText} from '#/components/Button'
 import {
   type ConvoWithDetails,
   type GroupConvoMember,
 } from '#/components/dms/util'
 import {createStaticClick, SimpleInlineLinkText} from '#/components/Link'
 import * as ProfileCard from '#/components/ProfileCard'
+import * as Prompt from '#/components/Prompt'
 import * as Toast from '#/components/Toast'
 import {Text} from '#/components/Typography'
 import {MemberMenu} from './MemberMenu'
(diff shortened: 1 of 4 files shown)


src/screens/Messages/ConversationSettings/prompts.tsx
```diff
@@ -97,6 +97,30 @@ export function LeaveChatPrompt({
   )
 }
 
+export function RemoveMemberPrompt({
```
ds-boyce, line 154: Maybe outside this PR, but we don’t show a prompt from the context menu’s “Remove from chat” option. Should we add this there too?

src/screens/Messages/ConversationSettings/Member.tsx
```diff
@@ -147,6 +172,11 @@ export function Member({
         )}
         {statusBadge}
       </View>
+      <RemoveMemberPrompt
+        control={removeMemberPrompt}
+        displayName={displayName}
+        onConfirm={() => removeMembers({members: [profile.did]})}
+      />
```
ds-boyce: Should we also gate this behind `showRemoveButton`?

src/screens/Messages/ConversationSettings/prompts.tsx
mozzius: Good shout - claude added this because it was concerned it was too easy to misclick as a top-level button, but we can certainly add to both

mozzius:
@claude please review

ds-boyce (approved):
Re: 2 looks like Claude disagreed with me. 😅 I’ll leave the decision to you.

mozzius:
@claude please review

PR #10727, opened 2026-06-04
Add list of mutual chats to block dialog

ds-boyce:
Replaces all block prompts with a new block dialog that includes a list of mutual group chats, if any.

* If you’re the owner, you can kick the member from the chat.
* If not, you can leave the chat.

list

### DIFF
--- a/src/components/moderation/BlockDialog.tsx
+++ b/src/components/moderation/BlockDialog.tsx
@@ -6,17 +6,21 @@ import {
   ChatBskyGroupRemoveMembers,
 } from '@atproto/api'
 import {Trans, useLingui} from '@lingui/react/macro'
+import {useQueryClient} from '@tanstack/react-query'
 
 import {isNetworkError} from '#/lib/strings/errors'
 import {logger} from '#/logger'
 import {type Shadow} from '#/state/cache/types'
 import {useLeaveConvo} from '#/state/queries/messages/leave-conversation'
-import {useListMutualGroupsQuery} from '#/state/queries/messages/list-mutual-groups'
+import {
+  createListMutualGroupsQueryKey,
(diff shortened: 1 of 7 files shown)


mozzius:
@claude please review

ds-boyce:
@claude please review

ds-boyce:
> For the kick path you probably don't want to close the dialog — the viewer may want to keep going — so this only applies to leave.

Claude **really** wants to close the dialog on leave, but same as kick you can potentially leave multiple group chats from the dialog so this suggestion doesn’t make sense.

src/components/moderation/BlockDialog.tsx
```diff
@@ -0,0 +1,383 @@
...
+      footer={
+        <Dialog.FlatListFooter
+          onLayout={evt => setFooterHeight(evt.nativeEvent.layout.height)}>
+          {footer}
+        </Dialog.FlatListFooter>
+      }
+      style={[web(a.px_md)]}
+      contentContainerStyle={[a.gap_0, {paddingBottom: footerHeight}]}
+      scrollIndicatorInsets={{top: headerHeight, bottom: footerHeight}}
+      onRefresh={() => void onRefresh()}
```
mozzius: definitely do not want a refresh control here - I want the swipe down to close the sheet

ds-boyce:
@claude please review

ds-boyce:
@mozzius Ok, finally satisfied Claude. 😅

mozzius (approved):
Looking good to me. @claude please give it one final review

estrattonbailey (changes requested):
Making some UI tweaks on web

estrattonbailey (approved):
Nice

PR #10454, opened 2026-05-11
Add group chat join links to supported embed types

ds-boyce:
Pulls the `code` from the URL and passes it to `getJoinLinkPreviews` to get the data for the embed card.

https://github.com/user-attachments/assets/eec07b3d-4a7f-43da-ad28-9659011febce

PR #10728, opened 2026-06-04
[Chat] Add chat invite as message embed type

mozzius:
- Adds chat invites as a message embed type
- Refactors the component into a compound component to share across different locations
- Did a pass on making it match designs exactly
  - Added a `medium` size button - not pixel-perfect, will need a closer look some other time
 
Simulator Screenshot - iPhone 17 - 2026-06-04 at 18 21 04

## Test plan

- Test #10454 still works
- Test sending an invite link as a DM
- Test sending an invite link in the same chat - should show a copy button
- Test sending an invite link with additional text - should extract and remove the url from the text

### DIFF
--- a/src/components/dms/ChatInvite/Card.tsx
+++ b/src/components/dms/ChatInvite/Card.tsx
@@ -1,0 +1,90 @@
+import {View} from 'react-native'
+import {Plural, Trans} from '@lingui/react/macro'
+
+import {createSanitizedDisplayName} from '#/lib/moderation/create-sanitized-display-name'
+import {sanitizeHandle} from '#/lib/strings/handles'
+import {atoms as a, useTheme} from '#/alf'
+import {AvatarBubbles} from '#/components/AvatarBubbles'
+import {ProfileBadges} from '#/components/ProfileBadges'
+import {Text} from '#/components/Typography'
+import {useChatInvite} from './Context'
+
+/**
+ * Presentational preview of a chat invite: member avatars, group name, member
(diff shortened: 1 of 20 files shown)


src/components/dms/ChatInvite/Card.tsx
```diff
@@ -0,0 +1,90 @@
...
+  if (!preview) return null
+
+  const ownerDisplayName = createSanitizedDisplayName(preview.owner)
+  const ownerHandle = sanitizeHandle(preview.owner.handle, '@')
+  const avatarProfiles = preview.convo?.members ?? [preview.owner]
+
+  return (
+    <View style={[a.flex_row, a.gap_md, a.align_center]}>
+      <AvatarBubbles size={56} self profiles={avatarProfiles} />
+      <View style={[a.flex_1, size == 'large' ? a.gap_2xs : a.gap_xs]}>
```
ds-boyce: Nit: `===`

src/state/queries/join-links.ts
```diff
@@ -1,4 +1,10 @@
-import {AtpAgent} from '@atproto/api'
+import {useCallback} from 'react'
+import {
+  AtpAgent,
+  type BskyAgent,
```
ds-boyce: Nit: `BskyAgent` is deprecated in favor of `AtpAgent`.

ds-boyce (approved):
Beautiful. ✨ Can’t wait for all this to come together.

src/state/queries/join-links.ts
mozzius: God we gotta sort this out once and for all

PR #10730, opened 2026-06-04
Fix useJoinRequestMutation

ds-boyce:
Tried to DRY this up but failed to bind the call, so it would fail.

```
TypeError: Cannot read property '_client' of undefined {}
```


rafaeleyng (approved):
What is `this`? 😛

PR #10735, opened 2026-06-04
Gate StandardSite subscribe button custom theme on AAA contrast

estrattonbailey:
Only apply the publication's custom accent colors to the subscribe button when the accent/accentForeground pairing clears WCAG AAA for large text (4.5:1). Otherwise render the button with the default secondary_inverted styling, which is guaranteed legible.

Adds a contrastRatio util to colorGeneration and exposes it via alf utils.


mozzius:
Wow cooking

PR #10744, opened 2026-06-05
[Chat] Disable font scaling on fixed-height chat invite cards

mozzius:
## Summary

Chat invite cards rendered in fixed-height containers (post embeds in feeds and the composer) could overflow their container when the user had a large font scale set, because the text inside continued to scale.

This threads a `hasFixedHeight` flag through `ChatInvite.Root` and its context, so the card, join button, and profile badges disable font scaling when they live in a fixed-height container. Surfaces that aren't height-constrained (DM message embeds, the message input embed) pass `hasFixedHeight={false}` and keep scaling.

## Screenshot

Simulator Screenshot - iPhone 17 - 2026-06-05 at 14 36 44

## Test plan

- [ ] Set a large font scale in device accessibility settings
- [ ] View a chat invite embed in a feed - text should fit within the card
- [ ] View a chat invite in the composer - text should fit within the card
- [ ] Confirm DM message invite embeds still scale normally

### DIFF
--- a/src/components/Post/Embed/ChatInviteEmbed.tsx
+++ b/src/components/Post/Embed/ChatInviteEmbed.tsx
@@ -23,7 +23,7 @@ export function ChatInviteEmbed({
   style?: StyleProp<ViewStyle>
 }) {
   return (
-    <ChatInvite.Root code={code}>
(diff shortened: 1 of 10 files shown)


mozzius:
@claude please review

PR #10748, opened 2026-06-05
Align group invite settings with eligibility logic

mozzius:
## Summary

The chat settings screen handled an unset `allowGroupInvites` differently from how `canBeAddedToGroup` actually interprets it, causing the UI to mismatch the enforced behavior.

`canBeAddedToGroup` (`src/components/dms/util.ts`) treats an unset `allowGroupInvites` as a fall-through to the general DM preference (`allowIncoming`). But the settings screen and save logic didn't:

- **Display** (`src/screens/Messages/Settings.tsx`): an unset `allowGroupInvites` was hardcoded to show as "Users I follow", ignoring the DM preference. A user with `allowIncoming: 'all'` and no group-invite setting actually allows group invites from everyone, but the UI showed "Users I follow".
- **Save** (`src/state/queries/messages/actor-declaration.ts`): `allowGroupInvites` had no default and was only written to the record when truthy, so saving other settings left it omitted/undefined.

## Changes

- Added `resolveAllowGroupInvites` next to `canBeAddedToGroup` as the single source of truth for the `allowGroupInvites ?? allowIncoming ?? 'following'` fallback chain.
- Settings screen, the save path, and the optimistic `onMutate` cache update all use the helper, so display / persisted record / optimistic cache can't drift.
- On save, `allowGroupInvites` is always written to the declaration record as a concrete value.

## Behavioral change worth flagging for QA

Before this PR, a user with `allowIncoming: 'all'` and `allowGroupInvites: undefined` had group invites *implicitly* track their DM setting (via the unset fallback in `canBeAddedToGroup`). After this PR, the first time a user saves **any** chat setting, `allowGroupInvites` is locked to a concrete value, so later changes to `allowIncoming` no longer carry it along. This matches the user's currently-displayed intent, but it's a one-way decoupling.

## Test plan

- [ ] With no `allowGroupInvites` set and DM setting = Everyone, the group-invite radio shows "Everyone"
- [ ] Saving any setting persists a concrete `allowGroupInvites` value matching the displayed one
- [ ] Changing only the DM setting updates the optimistic cache's `allowGroupInvites` to the resolved value (no flicker on refetch)

### DIFF
--- a/src/components/dms/util.ts
+++ b/src/components/dms/util.ts
@@ -47,6 +47,22 @@ export function canBeAddedToGroup(profile: bsky.profile.AnyProfileView) {
   }
 }
 
+/**
+ * Resolves the effective `allowGroupInvites` value for a chat declaration.
+ * When unset, group invites follow the general DM preference
+ * (`allowIncoming`), which itself defaults to `following`. This mirrors the
+ * `undefined` fallthrough in canBeAddedToGroup, and is the single source of
+ * truth for both displaying and persisting the setting.
+ */
+export function resolveAllowGroupInvites(
+  chat: {allowIncoming?: string; allowGroupInvites?: string} | undefined,
+): 'all' | 'none' | 'following' {
+  return (chat?.allowGroupInvites ?? chat?.allowIncoming ?? 'following') as
+    | 'all'
+    | 'none'
+    | 'following'
+}
+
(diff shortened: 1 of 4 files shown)


mozzius:
@claude please review

mozzius:
@claude please take a second look

PR #10753, opened 2026-06-05
[Chat] Enforce limits for group size

mozzius:
- Group member count in create/add flows
- Group name length in create/edit name flows


mozzius:
@claude please review

ds-boyce (approved):
Better all the time! I like adding a “Group is full” message but that’s not blocking.

mozzius:
Ah yeah we can add that as a follow-on

PR #10754, opened 2026-06-05
[Chat] Centralize conversation report-subject resolution

mozzius:
## Summary

Conversation-level reporting (the "Report conversation" action, distinct from reporting an individual message) was wired up independently at every surface, and each derived its report subject slightly differently. This started as "let empty group invites be reported" and grew into unifying the logic.

Adds a single helper, `getConvoReportSubject(convo, ownDid)` in `src/components/dms/util.ts`, used at every conversation-level report surface:

- **Group** → always report the whole convo, targeting the **owner** (`convoRef`). Returns `null` if the owner has left (nothing to report against), in which case the Report item is hidden.
- **Direct** → report the last **reportable** message if there is one (exists + not sent by us) as a `messageRef`; otherwise report the whole convo targeting the other user (`convoRef`).
- **Message context menu** (reporting a specific message) is unchanged.

The after-report dialog is now routed by convo kind everywhere: group → "Block and leave" (`AfterReportConversationDialog`), direct → "Block and delete" (`AfterReportDialog`).

## Why

- Removes the `lastMessage` gate so empty group invites can be reported.
- Fixes the inconsistency where `ConvoMenu` blocked `profile.did` on the no-message path but `latestReportableMessage.sender.did` on the message path — in a group opened from the chat-list row those differed, causing a wasted profile cache-prime and a brief loading flicker on the post-report block step. The block target is now uniformly `primaryMember.did`.
- `RejectMenu` previously reported your own last message as the subject (no `isFromMe` exclusion) and always showed the "delete" dialog even for group invites. Both are now consistent with the rest of the app.

## Surfaces touched

- `ConvoMenu` (chat-list row + direct convo header) — now takes `ConvoWithDetails`, drops the `latestReportableMessage` prop, resolves the subject internally.
- `RejectMenu` (request list + invite-in-conversation) — uses the helper, routes the after-dialog by kind.
- `ChatListItem` / `MessagesListHeader` — pass the parsed convo, drop the now-unused `latestReportableMessage` plumbing.

`ConversationSettings` (group-only) already did exactly the helper's group-branch behavior, so it's left as-is.

## Test plan

- `pnpm typecheck` + `pnpm lint` pass.
- Manual matrix:
  - Direct, last message from other user → "Report this conversation", `messageRef`, "Block and delete".
  - Direct, last message is mine / no messages → `convoRef` targeting other user, "Block and delete".
  - Group, any state → `convoRef` targeting owner, "Block and leave".
  - Group, owner has left → Report item hidden.
  - No profile-loading flicker on the post-report block step for group chats opened from the chat-list row.

### DIFF
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -129,12 +129,7 @@ let ConvoMenu = ({
             }}
             control={reportControl}
             onAfterSubmit={() => {
-              const sender = convo.members.find(
-                member => member.did === latestReportableMessage.sender.did,
-              )
-              if (sender) {
-                unstableCacheProfileView(queryClient, sender)
-              }
+              unstableCacheProfileView(queryClient, profile)
               blockOrDeleteControl.open()
             }}
           />
--- a/src/components/dms/ConvoMenu.tsx
+++ b/src/components/dms/ConvoMenu.tsx
@@ -1,6 +1,6 @@
 import {memo, useCallback} from 'react'
 import {Keyboard, View} from 'react-native'
-import {ChatBskyConvoDefs, type ModerationCause} from '@atproto/api'
+import {type ModerationCause} from '@atproto/api'
 import {Trans, useLingui} from '@lingui/react/macro'
 import {useNavigation} from '@react-navigation/native'
 import {useQueryClient} from '@tanstack/react-query'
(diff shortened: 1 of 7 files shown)


mozzius:
@claude please review

ds-boyce (approved):

mozzius:
@claude please review again

mozzius:
@claude the scope has grown significantly - please review again

ds-boyce (approved):
Merge conflicts 😞 but LGTM.

PR #10757, opened 2026-06-05
Block posting of invalid chat invites

ds-boyce:
Looks for invalid `chat-invite` links in the `resolve-link` query cache and disables the post button if one is found. Uses the cache so we can differentiate between the `undefined` loading state and the `null` resolved state.

https://github.com/user-attachments/assets/264af81a-a8aa-4b0e-bd88-e48e4301992a


ds-boyce:
@claude please review

mozzius (approved):
good stuff

PR #10768, opened 2026-06-06
fix: avoid mutating query cache for birthDate conversion in usePreferencesQuery

ashish-khankari:
The birthDate string-to-Date conversion was happening outside the select 
function by directly mutating query.data.birthDate. This ran on every render 
and violated react-compiler immutability rules.

Moved the conversion into the select function where a new object is returned 
instead of mutating the cached data.


estrattonbailey:
This is unfortunately required for the moment because of how we serialize persisted queries

PR #9948, opened 2026-02-25
Don't strip query parameters on in-app navigation

surfdude29:
Might partially fix #6400 

> [!WARNING]
> **This PR and its description [were put together by Claude](https://claude.ai/share/013162ff-7579-46d2-a459-d832093c47c0). Although it passes CI, I have not tested the fix.**

## Summary

When a user taps a link like `https://bsky.app/hashtag/muzika?author=did:plc:nr4a7ddupzvwpfv5sxklmkqh` from within a post in the app, the `?author=` query parameter is silently dropped, and the hashtag screen shows unfiltered results instead of results from that specific author.

## Root Cause

The function `convertBskyAppUrlIfNeeded()` in `src/lib/strings/url-helpers.ts` converts `bsky.app` URLs into internal navigation paths for the app's router. On **line 202**, it returns only `urlp.pathname`, discarding all query parameters:

```typescript
// BEFORE (buggy)
export function convertBskyAppUrlIfNeeded(url: string): string {
  if (isBskyAppUrl(url)) {
    try {
      const urlp = new URL(url)

      if (isBskyStartUrl(url)) {
        return startUriToStarterPackUri(urlp.pathname)
      }

      // special-case search links
      if (urlp.pathname === '/search') {
        return `/search?q=${urlp.searchParams.get('q')}`
      }

      return urlp.pathname  // ← BUG: drops all query parameters
    } catch (e) { ... }
  }
  ...
}
```

A `/search` special case existed (lines 197–200) which preserved the `q` parameter for search URLs, but no equivalent existed for hashtag URLs or any other route.

### Call Chain

1. User taps `https://bsky.app/hashtag/muzika?author=did:plc:xxx` in a post
2. `Link.tsx` calls `convertBskyAppUrlIfNeeded(url)` → returns `/hashtag/muzika` (query stripped!)
3. `router.matchPath('/hashtag/muzika')` → `{tag: 'muzika'}` (no `author`)
4. `HashtagScreen` receives `route.params` without `author`, shows unfiltered results

## Fix

**File:** `src/lib/strings/url-helpers.ts`

Replace the search-specific special case and the bare `urlp.pathname` return with a single line that preserves query parameters for all bsky.app URLs:

```typescript
// AFTER (fixed)
export function convertBskyAppUrlIfNeeded(url: string): string {
  if (isBskyAppUrl(url)) {
    try {
      const urlp = new URL(url)

      if (isBskyStartUrl(url)) {
        return startUriToStarterPackUri(urlp.pathname)
[...]

Issue #6400, opened 2024-11-16
Link for hashtag from someone


sutnistj:
### Steps to Reproduce

Before a link like https://bsky.app/hashtag/muzika?author=did:plc:nr4a7ddupzvwpfv5sxklmkqh worked as should but from some time it stopped or itʼs just my imagination that worked. For compare, a link with similar type (with did:plc) https://bsky.app/profile/did:plc:nr4a7ddupzvwpfv5sxklmkqh still works as expected. 

### Attachments

![image](https://github.com/user-attachments/assets/b72b666f-4749-4eaa-afad-9e0434a1ec5f)

As you can see, the top had a part: 
> #muzika
> Від @‪did:plc:nr4a7ddupzvwpfv5sxklmkqh‬

But result ignore the part «from». 

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

firefox 132.0.2 from official mozilla repo; linux mint cinnamon

### What version of the app are you using?

Version 1.93.0 6bfcdeb (prod)

bmann:
Same report for handle.

Posting the filtered by author link in a post — eg https://bsky.app/hashtag/Vancouver?author=bmann.ca

doesn’t apply the author filter query in either iOS app or iOS mobile Safari

mozzius:
Looks like this got fixed

sutnistj:
@mozzius  Not, itʼs not fixed. 

<img width="599" height="802" alt="Image" src="https://github.com/user-attachments/assets/b082aec8-13ee-4765-8d1b-e921e568050a" />

What expected:

<img width="605" height="793" alt="Image" src="https://github.com/user-attachments/assets/390dde0d-d2e1-4092-b0c1-2ff65a32cb49" />

surfdude29:
I just tested these links in the iOS app:

https://bsky.app/hashtag/muzika?author=did:plc:nr4a7ddupzvwpfv5sxklmkqh

https://bsky.app/hashtag/Vancouver?author=bmann.ca

**Tapping the links from a post in the app:**
Neither one correctly filters by author

**Tapping the links from outside the app:**
The DID link doesn't filter by author but the handle link does

PR #10772, opened 2026-06-07
Fix hashtag author filter failing for DIDs

surfdude29:
> [!IMPORTANT]
> **Claude Code analyzed this issue and wrote the PR and much of the description.**
>

## Context

A related problem in #6400 was fixed by #9948 but the core issue that was originally described still exists. This PR aims to fix the originally described issue by allowing hashtag search pages to be filtered by DID-based author.

## The Problem

When navigating to a hashtag URL with a DID-based author filter (e.g. `https://bsky.app/hashtag/muzika?author=did:plc:nr4a7ddupzvwpfv5sxklmkqh`), the author filter is silently ignored – the header shows `From @did:plc:...` but the search results are unfiltered, showing posts from all authors.

The header also incorrectly shows an `@` before the DID.

## Root Cause

The `HashtagScreenTab` component constructs a freeform query string like:

```
#muzika from:did:plc:nr4a7ddupzvwpfv5sxklmkqh
```

and passes it as the `q` parameter to `app.bsky.feed.searchPosts`. The search backend uses Lucene-style query parsing, and the colons in the DID (`did:plc:xxx`) cause the parser to misinterpret the `from:` operator (it appears to read `from:did` as the value and discard the rest), so the filter is dropped.

Handles don't contain colons, so `from:sutnistj.bsky.social` parses correctly – which is why the handle case works.

### Use the dedicated `author` API parameter

The `app.bsky.feed.searchPosts` endpoint already has a dedicated `author` parameter (type: `at-identifier`) that properly accepts both handles and DIDs. From the [lexicon](https://github.com/bluesky-social/atproto/blob/main/lexicons/app/bsky/feed/searchPosts.json):

> Filter to posts by the given account. Handles are resolved to DID before query-time.

Instead of stuffing `from:{author}` into `q`, this PR passes `author` as a first-class API parameter, letting the backend handle filtering for both handles and DIDs.

### Omit `@` prefix for DIDs in the header

The subtitle now checks whether `author` starts with `did:` and conditionally omits the `@` prefix, so DIDs display as `From did:plc:xxx` instead of `From @did:plc:xxx`.

## Key Changes

- **Updated `searchPostsQueryKey`**: Added `author` as an optional parameter to the query key function so each author filter gets a distinct cache entry
- **Extended `useSearchPostsQuery` hook**: Added optional `author` parameter that gets passed directly to the API call
- **Updated query dependency tracking**: Added `author` to the `selectArgs` useMemo dependency array and updated the condition for detecting user-specific searches to check for the `author` parameter
- **Refactored `HashtagScreenTab`**: 
  - Removed the logic that appended `from:${author}` to the search query string
  - Now passes `author` as a separate parameter to `useSearchPostsQuery`
  - Simplified `queryParam` to only handle the hashtag/cashtag formatting
- **Improved author display**: Updated the subtitle text in `HashtagScreen` to conditionally format the author display so that the `@` prefix is only shown for handles and not for DIDs

## Testing

| Scenario | Before | After |
|---|---|---|
| `?author=did:plc:xxx` – results | Unfiltered | ✅ Correctly filtered |
| `?author=did:plc:xxx` – header | `From @did:plc:xxx` | ✅ `From did:plc:xxx` |
| `?author=sutnistj.bsky.social` – results | Works | ✅ Works (now via dedicated param) |


ds-boyce:
👀

ds-boyce (approved):
Thanks!

PR #10773, opened 2026-06-07
Fix incorrect background rendering when re-enabling automation toggle

ashish-khankari:
### Description:
This fixes an issue where the background color does not correctly restore when the automation toggle is re-enabled. After turning the feature back on, only partial styling is applied, resulting in an incomplete dark background (visible as a white circular area only instead of a correct background). Disabling the toggle restores the correct appearance, indicating inconsistent state handling during reactivation.

The update ensures that background styles are fully reapplied whenever the automation toggle is enabled, preventing stale or partially applied styles from persisting across state changes.

### UX Impact:

- Fixes inconsistent visual state when toggling automation back on
- Ensures background renders correctly 
- Improves perceived stability and polish of automation switching behavior

### Reporduction (Before Fix)

image

### Result (After Fix)

image

### DIFF
--- a/src/components/forms/Toggle/index.tsx
+++ b/src/components/forms/Toggle/index.tsx
@@ -461,7 +461,7 @@ export function Switch() {
 
     if (disabled) {
       base.push({
-        backgroundColor: t.palette.contrast_50,
+        backgroundColor: t.palette.contrast_100,
       })
 
       if (selected) {


estrattonbailey:
Would rather not touch the base design system components at this time, but thanks for the PR!

Issue #8003, opened 2025-03-16
Add time-to-live (TTL) to blocks and mutes


OneDeuxTriSeiGo:
## Informal Description

Add the ability to "put users in the timeout corner" instead of blocking or muting them permanently (which is easy to forget you've done).

## Formal Description

Introduce a TTL (time-to-live) parameter to mutes and blocks. The TTL is a date-time after which the block/mute would expire and be dropped from the list of mutes/blocks.

The UX for blocking/muting would be to add a prompt to specify a timeout duration (from which the TTL would be computed) after clicking block/mute. 

Additionally a setting of "default timeout" should be added to the user's moderation settings.

The UX for being the recipient of a mute with TTL should not change from the current experience. i.e. The user should not be made aware of the mute.

The UX for being the recipient of a block with TTL should ideally specify how long the user has been blocked for/when the block expires so that it is clear to the user that they've received a "warning" and may not get a temporary block the next time around.

The UX for other users interacting with the block (i.e. side-effects of the nuclear block) should not differ from the current experience (as the "warning" is not for them).

### Attachments

Facebook Temporary Mute/Snooze: https://www.howtogeek.com/674715/how-to-mute-someone-on-facebook/ (includes pictures of the UX)

### Describe Alternatives

An alternative (suggested in the bsky dev discord) would be to introduce a generic TTL mechanism that could apply to actions in general and allow those actions to automatically revert after the TTL expires. This would apply to creating posts, changing profile icon, display name, profile description, pinning a post, etc. 

Example use cases:

- Auto-deleting posts after 1 year/1 month, etc (instead of running a bot to delete posts manually and hammering the API)
- Temporary icon, profile, or pin changes for events. ex: Display name @ Conference/convention, ex: pinning booth location for conventions for the duration of the convention.
- Pinning updated weekly schedules or "next stream/event" posts for streamers, etc

This more general use case would require significantly more work however it would provide a lot of power and flexibility to users.

### Additional Context

This post was the inspiration for making this issue: https://bsky.app/profile/did:plc:nn5vetzjmmbqm53jukhg5hpi/post/3lkgznfstcs2e The replies also all indicate support for the feature from various users.

cooljeanius:
some additional possible synonyms, for searchability:
- "timed blocks"
- "temporary blocks"
- "short-term blocks"
- "ephemeral blocks"

PR #10777, opened 2026-06-07
Add time-to-live (TTL) to blocks and mutes

alvesykes:
### Feature: 

Allows users to choose whether a mute or block should be permanent or temporary (24 hours, 7 days, or 30 days). 

### Implementation details:
-  Architecture: App-level implementation requiring no changes to the AT Protocol. Continues to use existing APIs while storing local expiration metadata (target DID, expiration time, and block record URI).

-  Storage: Metadata is stored in account-scoped storage to ensure temporary moderation actions from one account do not affect another account on the same device.

-  Logic & Hooks: Introduced src/state/moderation-timeouts.ts and the useModerationTimeoutCleanup() hook, which runs periodically in the app shell to remove expired actions and invalidate UI queries. Profile mutation queues were extended to accept an optional expiresAt value.

-  UI: Added TimeoutSelectDialog to profile menus, post menus, and the after-report flow. Moderation settings screens now read local timeout metadata, hide expired entries, and display expiration info.

-  Testing: Added comprehensive unit tests in src/state/__tests__/moderation-timeouts.test.ts covering expiration logic, storage, and account separation.

### Linked Issue:

-  #8003 

### Co-Author:

-  @Miguel-Alfaiao

### DIFF
--- a/src/components/PostControls/PostMenu/PostMenuItems.tsx
+++ b/src/components/PostControls/PostMenu/PostMenuItems.tsx
@@ -1,4 +1,4 @@
-import {memo, useMemo} from 'react'
+import {memo, useMemo, useState} from 'react'
 import {
   Platform,
   type PressableProps,
@@ -33,6 +33,10 @@ import {logger} from '#/logger'
 import {type Shadow} from '#/state/cache/post-shadow'
 import {useProfileShadow} from '#/state/cache/profile-shadow'
 import {useFeedFeedbackContext} from '#/state/feed-feedback'
+import {
+  calculateExpiresAt,
+  type ModerationTimeoutDuration,
+} from '#/state/moderation-timeouts'
 import {
   useHiddenPosts,
   useHiddenPostsApi,
@@ -92,6 +96,7 @@ import {
   ReportDialog,
   useReportDialogControl,
 } from '#/components/moderation/ReportDialog'
+import {TimeoutSelectDialog} from '#/components/moderation/TimeoutSelectDialog'
 import * as Prompt from '#/components/Prompt'
 import * as Toast from '#/components/Toast'
 import {useAnalytics} from '#/analytics'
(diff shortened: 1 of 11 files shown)


vikanezrimaya:
While I am a mere passer-by: why store on device-side? Why not extend the block lexicon with an optional expiration key? Does it make sense to keep the expiry date private?

My use-case: I switch between a phone, tablet and web browser to read my skyline. I would prefer that temporary block info actually followed me no matter which of my devices I'm using. I don't mind revealing expiry dates for my blocks, as blocks are already public info. In fact, I think this would be healthy for the social ecosystem.

Miguel-Alfaiao:
Good point. The main downside of the current approach is exactly that the expiry metadata does not sync across devices.

We did consider a server-side approach, but the Bluesky API does not currently support expiration timestamps for block/mute actions natively and adding that would require protocol-level changes. So this implementation is a first version that delivers the feature without any breaking changes.

About the privacy question, keeping the expiry date local was more of a side effect of the device-side approach than an intentional decision. You make a good point that blocks are already public, so a server-side solution would not introduce new privacy concerns either. That would definitely be the ideal long-term direction and could be built on top of this.

alvesykes:
Hi! Just pinging this PR in case it got overlooked.

Tagging @pfrazee and @estrattonbailey in case you’ve had an opportunity to take a look or could help direct it to the right person. Let me know if there's anything else you'd like for us to adjust or improve.

estrattonbailey:
Hey all, love the intent here and thanks for the PR. As noted, the real place this should live is on the record, and in fact, block handling goes far far deeper than the client. In may cases, content is filtered at the database level before it even reaches our appview or chat services.

To implement timed blocks would be a large undertaking that we don't have capacity for at this time. We've been aware of this ask for a while though, and there's always a chance we'll get to it in the future 🙏

Miguel-Alfaiao:
Thanks for the explanation! That makes total sense. Good to know it's on the radar for the future.

PR #10781, opened 2026-06-08
Show accept/reject footer for empty request convos

mozzius:
## What

Fixes a bug where a pending chat request with no messages yet showed the message composer instead of the accept/reject footer.

## Cause

In `getFooterState()` (`MessagesList.tsx`), the `items.length === 0` check ran before the `request` status check, so an empty request convo returned `new-chat` (composer) and never reached the `request` branch (`ChatStatusInfo`).

## Fix

Check the `request` status first, so a pending request renders its accept/reject footer regardless of message count.


mozzius:
@claude please review

PR #10783, opened 2026-06-08
Allow multiline specific fields to expand based on input content

RetroSunstar:
Currently text fields made with the intent of hosting content with multiple lines are locked to 80px (3 lines)

This will often come up in at least 3 places: the description fields for lists/profiles and the reason box in the report dialog

The max ``280`` height was based on how description fields collapse down to ~15 lines on profiles, and makes sense to have an upper limit instead of expanding to the edge of the page like the composer

Tested and works on both chrome and firefox 🐙 

Though there may be a cleaner approach all the same 😅


estrattonbailey:
We've got an `AutosizedTextarea` component now, would that work here?

RetroSunstar:
> We've got an `AutosizedTextarea` component now, would that work here?

Something a little more like that? 🤔

PR #10791, opened 2026-06-08
[Chat] Try and resolve loading bug once and for all

mozzius:
Has a weird slide animation when you enter from a push noty but whatever

https://github.com/user-attachments/assets/0115d94e-368c-4044-b2a3-25d9cab1008b


ds-boyce (approved):
This was a pain, I’m sure.

PR #10801, opened 2026-06-08
Fix back button behavior on conversation settings screen

ds-boyce:
After a full refresh, the back button here now takes you back to the conversation.


src/screens/Messages/ConversationSettings/index.tsx
```diff
@@ -86,12 +86,23 @@ type Props = NativeStackScreenProps<
...
       <Layout.Header.Outer>
-        <Layout.Header.BackButton />
+        <Layout.Header.BackButton
+          onPress={evt => {
+            if (!navigation.canGoBack()) {
+              evt.preventDefault()
+              navigation.navigate('MessagesConversation', {
+                conversation: convoId,
+              })
+            }
```
mozzius: I think we should make this web only. Otherwise, if you ended up here somehow on native, it would _push_ the screen, the back button on that screen would take you back, and you'd be stuck in a loop

```suggestion
            if (IS_WEB && !navigation.canGoBack()) {
              evt.preventDefault()
              navigation.navigate('MessagesConversation', {
                conversation: convoId,
              })
            }
```

ds-boyce: Good call.

mozzius (approved):

PR #10803, opened 2026-06-08
bskyweb: meta noindex tag for non-indexable pages

blackmichael:
Per Google's SEO feedback, web pages for action URLs or otherwise non-content-contributing URLs should have `noindex` directives in the HTML meta. There's also value in adding `nofollow` directives for our `/intent/*` routes as well, to prevent crawlers from unnecessarily following links found in an action route.


mozzius:
@claude please review

vineyardbovines (approved):

PR #10812, opened 2026-06-09
Fix composer horizontal swipe triggering browser back/forward

vineyardbovines:
## Summary

- Two-finger horizontal trackpad swipes over the composer were escaping to the page's history-nav gesture, particularly noticeable when a quote post was attached.
- Apply `overscroll-behavior-x: contain` to the composer's `Animated.ScrollView` (Chrome/Firefox) and add a wheel listener that `preventDefault`s horizontal-dominant wheels for Safari, mirroring the existing carousel fix.

## Test plan

- [ ] In the composer (with and without a quote attached), two-finger horizontal trackpad swipe over the body does not trigger browser back/forward
- [ ] Gallery inside a quote post still scrolls horizontally between images
- [ ] Vertical scrolling inside the composer still works
- [ ] Verify in Chrome and Safari

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -134,7 +134,14 @@ import * as Prompt from '#/components/Prompt'
 import * as Toast from '#/components/Toast'
 import {Text} from '#/components/Typography'
 import {useAnalytics} from '#/analytics'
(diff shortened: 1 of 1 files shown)


mozzius:
@claude please review

estrattonbailey (approved):
Yep nice.

Noting for posteriy: carousels aren't swipeable in the composer anyway, but this prevents nav from happening for anyone who tries to swipe. If we notice any issues with this, we can walk this back.

PR #10813, opened 2026-06-09
bskyweb: extract gallery media for SEO

blackmichael:
Posts using `app.bsky.embed.gallery` (Photos v2) were rendering with no `og:image` and an empty JSON-LD `image[]`. `extractPostMedia` in `bskyweb/cmd/bskyweb/jsonld.go` only covered the legacy `app.bsky.embed.images` and `app.bsky.embed.video` shapes, plus the same two as media in a record-with-media embed. Anything authored via the new gallery embed fell through to `nil` and shipped without preview imagery.

This adds gallery support in both the top-level embed slot and the `record-with-media.media` slot, mirroring the existing images/video paths so the byte-equality contract between `og:image` and the JSON-LD `image[]` / `thumbnailUrl` fields is preserved (Google requires this; the existing render test guards it).

## Changes

- Bumps indigo version

**`bskyweb/cmd/bskyweb/jsonld.go`**
- `extractPostMedia` now branches on `EmbedGallery_View` in both slots.
- New `galleryThumbs` helper extracts `Thumbnail` from each `EmbedGallery_View_Items_Elem`. Items are a union type, so nil entries and unknown (future) variants are skipped rather than emitting empty strings - this keeps the SEO path forward-compatible if non-image gallery item types land later.

**Tests**
- New `withGallery` and `withRecordWithMediaGallery` helpers.
- `TestBuildPostJSONLD_WithGallery` mirrors the existing images test (length, order, `thumbnailUrl == image[0]`).
- `TestBuildPostJSONLD_GalleryInRecordWithMedia` covers the nested case and confirms `isBasedOn` still emits for the quoted record.
- `TestExtractPostMedia_GallerySkipsUnknownItems` covers nil items, empty-union items, and an all-unknown gallery (returns nil, not \`[\"\"]\`).
- `TestRenderPost_OGImageMatchesJSONLD_Gallery` enforces the byte-equality contract end-to-end. Skips locally without `templates/scripts.html`, same as the existing image variant; runs in CI after `yarn build-web`.

## Out of scope

- og:video for gallery. Today's gallery union only contains image items; if/when video items land, that's a separate ticket.
- Alt text in JSON-LD image entries (would require shifting from `[]string` to `ImageObject`; not gallery-specific).

## Tests

- Post with 10 images in a gallery embed renders proper jsonld data https://search.google.com/test/rich-results/result?id=8yXNsv9Zg1FW_57ojG-ozA
- Post with 1 image in classic images embed has no regressions with jsonld data https://search.google.com/test/rich-results/result?id=kn5QRZdgUdKNDcAETpaFfA

### DIFF
--- a/bskyweb/cmd/bskyweb/jsonld.go
+++ b/bskyweb/cmd/bskyweb/jsonld.go
@@ -183,7 +183,8 @@ func imageThumbs(images []*appbsky.EmbedImages_ViewImage) []string {
 // galleryThumbs returns the thumbnail URLs of image items in a gallery
 // embed, or nil if empty. Items_Elem is a union; non-image variants and
 // nil entries are skipped so future gallery item types don't break SEO
-// extraction.
+// extraction. Empty Thumbnail strings are also skipped to avoid emitting
+// <meta property="og:image" content=""> if the appview ever returns one.
 func galleryThumbs(items []*appbsky.EmbedGallery_View_Items_Elem) []string {
 	if len(items) == 0 {
 		return nil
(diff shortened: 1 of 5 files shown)


blackmichael:
@claude please review

estrattonbailey (approved):

PR #10814, opened 2026-06-09
bskyweb: add video objects to jsonld SEO

blackmichael:
Adds schema.org `VideoObject` to the JSON-LD blob bskyweb already emits for posts and profiles. When a post has a video embed, the `DiscussionForumPosting` (and `Comment`, and profile `hasPart[]`) entries now include a `video` field with `@type: VideoObject`, name, description, thumbnailUrl, uploadDate, contentUrl (HLS playlist), embedUrl (canonical post URL), and width/height when the aspect ratio is available.

- Refactors video-embed lookup into a shared `findVideoEmbed` helper used by both `extractVideoMeta` (og:video) and `buildVideoObject` (JSON-LD), so the two outputs cannot drift.
- `name` falls back to `Video by @<handle>` when alt text is empty, then to `Video on Bluesky` for `handle.invalid`. `description` falls back to `name` when post text is empty, so Google rich-result requirements are always satisfied.
- Honors the existing `embedHidden` gate (self/post-view labels) so labeled posts do not surface video metadata.
- `duration` is omitted because the appview does not expose it.

## Tests

- Post with video now renders video object jsonld https://search.google.com/test/rich-results/result?id=MOVAWOretWmwtm-VKnhAQw

### DIFF
--- a/bskyweb/cmd/bskyweb/jsonld.go
+++ b/bskyweb/cmd/bskyweb/jsonld.go
@@ -493,14 +493,8 @@ func buildPostNode(pv *appbsky.FeedDefs_PostView, replies []*appbsky.FeedDefs_Th
 		thumb = images[0]
 	}
 
-	postURL := bskyPostURLFromATURI(pv.Author.Handle, pv.Uri)
+	postURL := bskyPostURLFromATURIWithDIDFallback(pv.Author.Handle, pv.Uri)
 	postText := postRecordText(pv)
-	// videoEmbedURL prefers the handle-form URL but falls back to DID-form
-	// for handle.invalid authors so VideoObject.embedUrl is never empty.
-	videoEmbedURL := postURL
-	if videoEmbedURL == "" {
-		videoEmbedURL = bskyPostURLFromATURIWithDIDFallback(pv.Author.Handle, pv.Uri)
-	}
 
 	node := discussionForumPosting{
 		Type:            "DiscussionForumPosting",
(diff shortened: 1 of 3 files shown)


blackmichael:
@claude plz review

bskyweb/cmd/bskyweb/jsonld.go
```diff
@@ -394,14 +493,24 @@ func buildPostNode(pv *appbsky.FeedDefs_PostView, replies []*appbsky.FeedDefs_Th
...
 	}
 
+	postURL := bskyPostURLFromATURI(pv.Author.Handle, pv.Uri)
+	postText := postRecordText(pv)
+	// videoEmbedURL prefers the handle-form URL but falls back to DID-form
+	// for handle.invalid authors so VideoObject.embedUrl is never empty.
+	videoEmbedURL := postURL
+	if videoEmbedURL == "" {
+		videoEmbedURL = bskyPostURLFromATURIWithDIDFallback(pv.Author.Handle, pv.Uri)
+	}
```
estrattonbailey: `postUrl` and `videoEmbedUrl` are kinda redundant here right? Should `postUrl` use the DID-fallback method too?

bskyweb/cmd/bskyweb/jsonld.go
```diff
@@ -464,14 +573,21 @@ func buildReplyNode(pv *appbsky.FeedDefs_PostView, hideLabels map[string]bool) c
 	if len(images) > 0 {
 		thumb = images[0]
 	}
+	postURL := bskyPostURLFromATURI(pv.Author.Handle, pv.Uri)
+	postText := postRecordText(pv)
+	videoEmbedURL := postURL
+	if videoEmbedURL == "" {
+		videoEmbedURL = bskyPostURLFromATURIWithDIDFallback(pv.Author.Handle, pv.Uri)
```
estrattonbailey: Same here

estrattonbailey (approved):
LGTM, one question

bskyweb/cmd/bskyweb/jsonld.go
blackmichael: Great callout, we should definitely just be using the same url in both cases

PR #10819, opened 2026-06-09
[Chat] Reusable loading/unavailable states for chat invites

mozzius:
## What

The `ChatInvite` primitive's loading and error/disabled handling was awkward: `JoinRequestEmbed` and `MessageInputEmbed` each reimplemented their own loading spinner and "no longer available" markup, and `MessageItemInviteEmbed` (the DM message embed) had no unavailable state at all — a disabled or invalid link rendered an empty bubble.

This extracts those states into the headless primitive so every surface shares them.

## Changes

- **`ChatInvite.Loading`** and **`ChatInvite.Unavailable`** — new content-only presentational parts. They own the spinner / warning markup but take the outer container `style` from the consumer, since the embed box and the chat bubble are styled differently.
- **`status` enum** — `Root` now exposes a single `status: 'loading' | 'error' | 'unavailable' | 'available'` from context, replacing the loose `loading`/`error` booleans. Consumers branch on one value.
- Updated all four consumers (`JoinRequestEmbed`, `ChatInviteEmbed`, `MessageItemInviteEmbed`, `MessageInputEmbed`) to consume the shared parts.
- **`MessageItemInviteEmbed` gains the "no longer available" state** it was previously missing.

Screenshot 2026-06-09 at 20 12 53

## Test plan

- [ ] Disabled/invalid invite link in a DM shows "no longer available" instead of an empty bubble
- [ ] Invite link in feed/composer still shows loading → card, and falls back to plain external embed on fetch error
- [ ] Logged-out and logged-in both render correctly

### DIFF
--- a/src/components/Post/Embed/ChatInviteEmbed.tsx
+++ b/src/components/Post/Embed/ChatInviteEmbed.tsx
@@ -38,9 +38,9 @@ function ChatInviteEmbedBody({
   onOpen?: () => void
   style?: StyleProp<ViewStyle>
 }) {
-  const {error} = ChatInvite.useChatInvite()
+  const {status} = ChatInvite.useChatInvite()
 
-  if (error) {
+  if (status === 'error') {
     return <ExternalEmbed link={link} onOpen={onOpen} style={style} />
   }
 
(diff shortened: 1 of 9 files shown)


mozzius:
@claude please review

PR #10820, opened 2026-06-09
Don't suggest users who can't be messaged/added

ds-boyce:
Was odd to put these under “Suggestions”. Note this only filters them underneath that heading; they still appear when you search.


mozzius (approved):
I think looks good, @claude please confirm

PR #10822, opened 2026-06-09
Fix link to profile from chat invite embed

ds-boyce:
This only worked on web.


mozzius (approved):
`ComplicatedInlineTextLink`

PR #10827, opened 2026-06-09
[Chat] Sync convo agent with shadow cache

estrattonbailey:


src/state/cache/profile-shadow.ts
```diff
@@ -53,6 +53,18 @@ const shadows: WeakMap<
...
+type ShadowUpdateEventPayload = {did: string; shadow: Partial<ProfileShadow>}
+export function useOnUpdateProfileShadow(
+  onUpdate: (payload: ShadowUpdateEventPayload) => void,
+) {
+  useEffect(() => {
+    emitter.addListener('shadow-update', onUpdate)
+    return () => {
+      emitter.removeListener('shadow-update', onUpdate)
+    }
+  }, [onUpdate])
```
ds-boyce: I think this is tearing down and adding the listener every render. Could store the `onUpdate` param in a ref and avoid the dependency array entirely.

estrattonbailey: Great call

ds-boyce (approved):

ds-boyce (approved):
Nice.

src/state/queries/profile.ts
```diff
@@ -528,7 +527,6 @@ export function useProfileBlockMutationQueue(
       updateProfileShadow(queryClient, did, {
         blockingUri: finalBlockingUri,
       })
-      void queryClient.invalidateQueries({queryKey: [RQKEY_LIST_CONVOS]})
```
estrattonbailey, line 531: This was causing a re-fetch which broke the identity checks the profile shadow does. I don't think we need this?

src/components/AvatarBubbles.tsx
```diff
@@ -100,45 +90,50 @@ export function AvatarBubbles({
...
   includeProfileBorder?: boolean
-  moderation?: ModerationUI
 }) {
   const t = useTheme()
-
+  const profile = useProfileShadow(profileUnshadowed)
+  const moderationOpts = useModerationOpts()
+  const moderation = moderationOpts
+    ? moderateProfile(profile, moderationOpts)
+    : undefined
```
estrattonbailey: `moderation.ui('avatar').blur === true` and yet the `UserAvatar` component doesn't update and so never gets blurred/unblurred on block. Not a massive deal I suppose.

I wasn't able to figure this out, need to run.

ds-boyce: > The root cause: `useProfileShadow`’s listener was calling `shadows.get(profile)` which only works when the profile object reference has been registered in the `WeakMap` via `findProfilesInCache`.

Confirmed switching to `setShadow(prev => ({...prev, ...value}))` fixes the blur issue in `AvatarBubbles`, but that means the `WeakMap` is no longer the single source of truth.

ds-boyce:
```diff
diff --git a/src/state/cache/profile-shadow.ts b/src/state/cache/profile-shadow.ts
index 430a03577..46f3f9aae 100644
--- a/src/state/cache/profile-shadow.ts
+++ b/src/state/cache/profile-shadow.ts
@@ -83,8 +83,8 @@ export function useProfileShadow<
   }

   useEffect(() => {
-    function onUpdate() {
-      setShadow(shadows.get(profile))
+    function onUpdate(value: Partial<ProfileShadow>) {
+      setShadow(prev => ({...prev, ...value}))
     }
     emitter.addListener(profile.did, onUpdate)
     return () => {
@@ -119,9 +119,8 @@ export function useMaybeProfileShadow<

   useEffect(() => {
     if (!profile) return
-    function onUpdate() {
-      if (!profile) return
-      setShadow(shadows.get(profile))
+    function onUpdate(value: Partial<ProfileShadow>) {
+      setShadow(prev => ({...prev, ...value}))
     }
     emitter.addListener(profile.did, onUpdate)
     return () => {
```

mozzius:
Ran a code review over this branch – found a few gaps in the shadow integration, fixes are stacked in #10841:

1. **Placeholder bubbles regression**: `if (!profile) return null` in `AvatarBubbles` broke the callers that pad `profiles` with `undefined` to render placeholder bubbles (`OutgoingRequestListItem`, `JoinRequest`, `GroupChatJoinDialog`).
2. **Shadow merge clobbered by server writes**: `mergeProfileShadow` mutated `relatedProfiles` one-shot, so any `setConvo`/`fetchMemberList`/`fetchMessageHistory`/firehose write reverted the optimistic state (these objects never enter the react-query cache, so the WeakMap shadows can't reach them either). Also only covered `relatedProfiles` – `convo.members`/`primaryMember` consumers (ConversationSettings, group info panel, ChatStatusInfo) stayed stale.
3. **Convo list staleness**: the removed `RQKEY_LIST_CONVOS` invalidation was still load-bearing – `calculateCount` (unread badge) and `getMessageInfo` read the cached members raw, outside any shadow hook, and blocks emit no chat log event, so nothing else refreshes the list.

Things reviewed and deliberately left alone: the `get-status` staleTime change (consistent with sibling queries, refetchOnWindowFocus is off) and the per-bubble `useProfileShadow` subscriptions (same cost already paid per-post in feeds).

mozzius (approved):

PR #10830, opened 2026-06-09
Add new analytics events for group chats

ds-boyce:
See `src/analytics/metrics/types.ts` for the new event types.

### DIFF
--- a/src/components/intents/GroupChatJoinDialog.tsx
+++ b/src/components/intents/GroupChatJoinDialog.tsx
@@ -1,3 +1,4 @@
+import {useEffect} from 'react'
 import {View} from 'react-native'
 import {
   ChatBskyGroupDefs,
@@ -45,6 +46,7 @@ import {InlineLinkText} from '#/components/Link'
 import {Loader} from '#/components/Loader'
 import * as Toast from '#/components/Toast'
 import {Text} from '#/components/Typography'
+import {useAnalytics} from '#/analytics'
 import {ProfileBadges} from '../ProfileBadges'
 
 export function GroupChatJoinDialog() {
@@ -83,6 +85,11 @@ function GroupChatJoinDialogContent({code}: {code?: string}) {
   const moderationOpts = useModerationOpts()
   const navigation = useNavigation<NavigationProp>()
   const queryClient = useQueryClient()
+  const ax = useAnalytics()
+
+  useEffect(() => {
+    ax.metric('groupchat:landingPage:view', {hasSession})
+  }, [ax, hasSession])
 
   const {data, error, isLoading} = useJoinLinkPreviewsQuery({
     codes: code ? [code] : undefined,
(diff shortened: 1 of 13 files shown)


mozzius:
@claude please review

mozzius (approved):
lgtm at a glance

ds-boyce:
3 is intended but I’m taking a look at 1 and 2.

src/components/intents/GroupChatJoinDialog.tsx
```diff
@@ -78,11 +81,21 @@ function GroupChatJoinDialogInner({code}: {code?: string}) {
...
   const navigation = useNavigation<NavigationProp>()
   const queryClient = useQueryClient()
+  const ax = useAnalytics()
+
+  const logViewOnce = useCallOnce()
+  useEffect(() => {
+    logViewOnce(() => {
+      ax.metric('groupchat:landingPage:view', {hasSession})
+    })
+  }, [ax, hasSession, logViewOnce])
```
estrattonbailey, line 96: Don't bother changing this but you can also just do `useCallOnce(() => {})()`

src/screens/Messages/components/MessagesList.tsx
```diff
@@ -471,6 +472,23 @@ export function MessagesList({
...
+
+      if (
+        embedView?.$type === 'chat.bsky.embed.joinLink#view' &&
+        ChatBskyGroupDefs.isJoinLinkPreviewView(embedView.joinLinkPreview)
+      ) {
+        ax.metric('groupchat:inviteLink:shared', {
+          convoId: embedView.joinLinkPreview.convoId,
+          method: 'dm',
+        })
+      }
```
estrattonbailey: Metrics per message? That'll be very noisy, we should be able to create dashboards for this without metrics. Asked in Slack.

ds-boyce: I’ll remove it. We’re tracking the clicks on the share buttons; the idea here was to compare that with actual shares.

src/screens/Messages/JoinRequest.tsx
```diff
@@ -44,10 +49,15 @@ type Props = {
...
   const t = useTheme()
   const {t: l} = useLingui()
+  const ax = useAnalytics()
 
   const {gtMobile, gtTablet} = useBreakpoints()
   const moderationOpts = useModerationOpts()
 
+  useEffect(() => {
+    ax.metric('groupchat:landingPage:view', {hasSession: false})
+  }, [ax])
```
estrattonbailey, line 56: This screen is always logged out, right?

ds-boyce: Yep, I’m removing it now.

ds-boyce: …actually wait. This event is used for both the logged-out landing page and the logged-in dialog (both via `/chat/{code}` URLs). So yes, this is always false here but it’s dynamic in the other context.

src/screens/Messages/components/MessagesList.tsx
ds-boyce: Removed.

PR #10831, opened 2026-06-09
Clear search term after adding chat member

ds-boyce:
Better than having to manually delete the search term every time after adding a new member.

https://github.com/user-attachments/assets/f7f9b3c3-d936-4a91-8331-7e0a39fd726f


mozzius (approved):
LGTM. @claude thoughts?

PR #10833, opened 2026-06-10
[Chat] revalidate join links on leave

mozzius:
## Test plan

- Have a join link pointing to a convo you're in
- Leave the convo
- Confirm card updates state immediately


mozzius:
@claude please review

PR #10834, opened 2026-06-10
Remove leave-convo suppression, use real optimistic update

mozzius:
## Why

When you leave a chat, we currently suppress it from the frontend in a heavy-handed way: `useLeftConvos()` reads React Query mutation state (pending *and* success) and every list surface (`ListConvosProvider`, `ChatList`, `Header`, `Inbox`) filters those convo IDs out. This was a workaround for a slow backend `leaveConvo` query (30+ seconds, sometimes timing out), where the convo had to stay hidden long after refetches would have clobbered an optimistic cache edit.

The backend query is now fast, so this can all go.

## The bug the workaround was hiding

`useLeaveConvo`'s `onMutate` already attempted an optimistic update, but it called `setQueryData([CONVO_LIST_KEY])` — an **exact** key match against just the root `'convo-list'`. Real queries are keyed `['convo-list', 'accepted', ...]`, `['convo-list', 'request', ...]`, etc., so the update never matched anything and was a no-op. That's why the suppression layer was needed in the first place.

## What changed

- `leave-conversation.ts`:
  - `onMutate` now uses `setQueriesData` (prefix match) to optimistically remove the convo from **all** convo-list queries, and also from the requests cache (`optimisticDelete` from `list-conversation-requests`), which the old code never touched
  - snapshots both caches and restores them in `onError`, then invalidates both roots — same pattern as `useAcceptConversation`
  - deleted `useLeftConvos()`
- `list-conversations.tsx`, `ChatList.tsx`, `Inbox.tsx`: removed the `leftConvos` filtering

No changes needed at the 8 `useLeaveConvo` call sites — the hook's signature is unchanged.

## Test plan

- [x] `pnpm typecheck`, `pnpm lint`, `pnpm test`
- [ ] Leave a 1:1 convo via the chat menu — disappears immediately, stays gone after refetch
- [x] Leave a group via conversation settings — navigates back, gone from list
- [ ] Reject/delete an incoming request from the inbox — disappears from requests list
- [ ] Failure case (offline/throttled): convo reappears in the correct list and error toast shows

### DIFF
--- a/src/screens/Messages/ChatList.tsx
+++ b/src/screens/Messages/ChatList.tsx
@@ -16,7 +16,6 @@ import {listenSoftReset} from '#/state/events'
 import {MESSAGE_SCREEN_POLL_INTERVAL} from '#/state/messages/convo/const'
 import {useMessagesEventBus} from '#/state/messages/events'
 import {useChatActorStatusQuery} from '#/state/queries/messages/get-status'
-import {useLeftConvos} from '#/state/queries/messages/leave-conversation'
 import {useListConvosQuery} from '#/state/queries/messages/list-conversations'
 import {EmptyState} from '#/view/com/util/EmptyState'
 import {List, type ListRef} from '#/view/com/util/List'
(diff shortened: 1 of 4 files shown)


mozzius:
@claude please review

PR #10838, opened 2026-06-10
[Chat] Fix optimistic updates by using partial keys

mozzius:
Optimistic updates from the chat firehose were broken because `RQKEY` was applying default values to the key, when it was intended to produce a partial key (to target multiple queries).

Fixed by using a proper partial key.

## Test plan

Send messages in a chat, confirm the chatlistitem updates with the lastmessage live


mozzius:
@claude please review

mozzius:
@claude please review

PR #10841, opened 2026-06-10
[Chat] Fix shadow overlay gaps + placeholder bubbles (stacked on #10827)

mozzius:
Stacked on #10827. Fixes issues found in code review of that branch:

### Placeholder bubbles regression (`AvatarBubbles.tsx`)
`if (!profile) return null` dropped the `AvatarPlaceholder` path for `undefined` entries, but `OutgoingRequestListItem`, `JoinRequest`, and `GroupChatJoinDialog` deliberately pad `profiles` with `Array(n).fill(undefined)` to render placeholder bubbles for unknown group members – they were rendering a lone avatar positioned for a 4-bubble layout. Switched to `useMaybeProfileShadow` so the profile can stay optional, and collapsed the two `Animated.View` returns back into one.

### Convo agent shadow merge could be clobbered (`agent.ts`)
`mergeProfileShadow` baked the shadow into `relatedProfiles` as a one-shot mutation. Any subsequent server write (`setConvo`/`updateConvo`/`fetchMemberList`/`fetchMessageHistory`/firehose ingestion) replaced the merged entry with a raw server profile – and since these objects come from direct service fetches, they're invisible to the shadow cache's react-query scan, so e.g. an optimistic block would visually revert on resume/refresh while the chat server's `viewer` state lagged. Now the agent keeps an accumulating `profileShadows: Map<did, Partial<ProfileShadow>>` overlay and re-applies it after every server write. Also:

- shadows now apply to `convo.members`/`primaryMember` too (previously only message rows updated; ConversationSettings, the group info panel, and ChatStatusInfo kept stale state)
- updates for dids not yet in `relatedProfiles` are no longer silently dropped
- no-op merges (e.g. the finalize after an optimistic update writes the same value) skip `commit()`, avoiding a full items-snapshot rebuild that re-renders every message row
- the subscription moved out of React: `ConvoProvider`'s `useOnUpdateProfileShadow` hook (with its render-time ref write + eslint suppression) is replaced by a plain `listenProfileShadowUpdate` that the agent subscribes to in its own lifecycle, alongside its first subscriber

### Reinstated list-conversations invalidation on block (`profile.ts`)
The shadow path only reaches components that read profiles through shadow hooks. The convo list is also read raw – `calculateCount` (unread badge) and `getMessageInfo` moderate the cached `primaryMember` directly – and blocks emit no chat log event, so with the invalidation removed that data stayed stale indefinitely. Re-added with a comment explaining why.

### DIFF
--- a/src/components/AvatarBubbles.tsx
+++ b/src/components/AvatarBubbles.tsx
@@ -10,7 +10,7 @@ import Animated, {
 } from 'react-native-reanimated'
 import {moderateProfile} from '@atproto/api'
 
-import {useProfileShadow} from '#/state/cache/profile-shadow'
+import {useMaybeProfileShadow} from '#/state/cache/profile-shadow'
 import {useModerationOpts} from '#/state/preferences/moderation-opts'
 import {useSession} from '#/state/session'
 import {UserAvatar} from '#/view/com/util/UserAvatar'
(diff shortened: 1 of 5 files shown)


estrattonbailey (approved):
Very nice

PR #10849, opened 2026-06-10
Enable Google Translate in post overflow menu

ds-boyce:
Flips the default from `false` to `true`. This is just an alternative to inline translation, which still uses on-device APIs (when available).

### DIFF
--- a/src/components/PostControls/PostMenu/PostMenuItems.tsx
+++ b/src/components/PostControls/PostMenu/PostMenuItems.tsx
@@ -551,9 +551,13 @@ let PostMenuItems = ({
               ) : (
                 <Menu.Item
                   testID="postDropdownTranslateBtn"
-                  label={l`Translate`}
+                  label={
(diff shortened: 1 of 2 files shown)


src/components/PostControls/PostMenu/PostMenuItems.tsx
```diff
@@ -551,9 +551,13 @@ let PostMenuItems = ({
...
                 <Menu.Item
                   testID="postDropdownTranslateBtn"
-                  label={l`Translate`}
+                  label={
+                    forceGoogleTranslate ? l`Google Translate` : l`Translate`
+                  }
                   onPress={onPressTranslate}>
-                  <Menu.ItemText>{l`Translate`}</Menu.ItemText>
+                  <Menu.ItemText>
+                    {forceGoogleTranslate ? l`Google Translate` : l`Translate`}
```
mozzius: maybe "Open in Google Translate"?

mozzius (approved):

PR #10854, opened 2026-06-10
Use add-only photo permission when saving QR codes

vineyardbovines:
## Problem

On a fresh iOS install, pressing Download in the invite friends dialog showed the full photo library read permission prompt and a failure toast at the same time, and the save did not complete.

The handler already requested the add-only permission (`requestPermissionsAsync(true)`), but then called `createAssetAsync`. In expo-media-library's iOS module, `createAssetAsync` fetches the created asset back via `PHAsset.fetchAssets` to resolve the promise with asset metadata. That fetch is a photo library read: with read-write authorization still undetermined, the Photos framework auto-prompts for full library access, the fetch returns nil, and the promise rejects, which produced the simultaneous prompt and error toast.

## Fix

Use `saveToLibraryAsync` instead, which writes via `UIImageWriteToSavedPhotosAlbum` and never reads the library. This matches the existing lightbox image download flow (`save-image.ios.ts`).

The starter pack QR dialog had the same pattern, plus it requested the read-scoped media library permission from expo-image-picker. It now uses the same add-only request and `saveToLibraryAsync`.

## Android

Checked against expo-media-library's Android source: `saveToLibraryAsync` uses the same `createAssetWithAlbumId` write path as `createAssetAsync`, minus resolving the asset info back to JS. Its permission gate passes unconditionally on API 33+, and on older API levels it needs `WRITE_EXTERNAL_STORAGE`, which is exactly what `requestPermissionsAsync(true)` requests there. The starter pack dialog no longer prompts for read access on Android either.

## Testing

`pnpm typecheck`, `pnpm lint`, and `prettier --check` pass. Permission state is sticky per install, so this should get a QA pass on a fresh install via TestFlight to confirm only the "Add Photos Only" prompt appears and the save completes.

### DIFF
--- a/src/components/StarterPack/QrCodeDialog.tsx
+++ b/src/components/StarterPack/QrCodeDialog.tsx
@@ -1,8 +1,7 @@
 import {Suspense, useRef, useState} from 'react'
 import {View} from 'react-native'
 import type ViewShot from 'react-native-view-shot'
-import {requestMediaLibraryPermissionsAsync} from 'expo-image-picker'
-import {createAssetAsync} from 'expo-media-library'
+import {requestPermissionsAsync, saveToLibraryAsync} from 'expo-media-library'
 import * as Sharing from 'expo-sharing'
 import {type AppBskyGraphDefs, AppBskyGraphStarterpack} from '@atproto/api'
 import {msg} from '@lingui/core/macro'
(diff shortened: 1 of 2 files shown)


estrattonbailey (approved):

vineyardbovines:
@claude quick pass

PR #10863, opened 2026-06-11
Add a Metrics Server to bskyweb

jcalabro:
I added metrics w/ echo in #10370, but I never actually rigged that up to a real server! oops!

We now expose this on an internal server so this doesn't leak to the public internet. This will be scraped in to our internal metrics servers for a nice usage dashboard.


mozzius:
@claude pls review

mozzius (approved):

PR #10867, opened 2026-06-11
Keep lightbox image index on device rotation

vineyardbovines:
## Why

Rotating the device while the lightbox is open snaps the pager back to the originally tapped image. Repro: open a post with multiple photos, tap the first photo, swipe to a later one, rotate the device. The lightbox jumps back to the first photo.

## Root cause

`ImageView` is rendered with `key={activeLightbox.id + '-' + orientation}`, so rotation intentionally remounts it to re-layout the transition transforms. The current image index lived inside `ImageView` as `useState(lightbox.index)`, so the remount reset it to the opening index and `PagerView` re-opened at `initialPage = lightbox.index`.

## Fix

Lift `imageIndex` up into `ImageViewRoot`, which survives rotation, and have `ImageView` capture the pager's `initialPage` from the preserved index at mount. The orientation-keyed remount itself is unchanged.

This takes the same approach as the earlier `samuel/lightbox-rotate-keep-index` branch, which fixed this in the pre-redesign lightbox but never landed; the redesign carried the bug forward. That branch can likely be deleted once this merges.

Notes from checking side effects: the swipe analytic does not fire on the rotation remount (prev and next are equal), the close animation measures the correct thumbnail after rotating, and Android gets the fix too since this is the shared native pager.

### DIFF
--- a/src/components/Lightbox/pager/ImagePager.tsx
+++ b/src/components/Lightbox/pager/ImagePager.tsx
@@ -91,6 +91,9 @@ export default function ImageViewRoot({
   'use no memo'
   const ref = useAnimatedRef<View>()
   const [activeLightbox, setActiveLightbox] = useState(nextLightbox)
+  // Lives here rather than in ImageView so it survives the remount
+  // when the orientation-based key below changes on rotation.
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):
Niiice

PR #10870, opened 2026-06-11
Optimistically update chat invite button status

ds-boyce:
The “Request to join” vs “Requested” status is based on `requestedAt`, so we optimistically update it so the button displays the correct status.


mozzius (approved):

mozzius:
@claude please review

PR #10873, opened 2026-06-11
Fix followed by chat invites

ds-boyce:
Got this exactly backwards.


mozzius (approved):
I think that's right now. @claude please confirm

PR #10875, opened 2026-06-11
[Chat] Use relatedprofiles for reactions/new message logs

mozzius:
Currently we were just ignoring the field, when we should've been appending them to the `members` list


mozzius:
@claude please review

PR #10877, opened 2026-06-11
Fix over-eager URL faceting in DMs composer

estrattonbailey:
Stuff like `P.S.` and `e.g.` no longer highlight as URLs. They're matched against `tlds` package as we do in `RichText`, and a subtle match group stripping bug was fixed by avoiding `matchAll` and just using indexed match groups.


ds-boyce (approved):
Nice, working as intended.

PR #10879, opened 2026-06-11
Upgrade to pnpm 11.5.3

ds-boyce:
[Changelog](https://github.com/pnpm/pnpm/releases/tag/v10.34.3)

> [!NOTE]
> ~I didn’t update `Dockerfile.embedr` this time (and we shouldn’t need to going forward) since I changed `"onFail"` to `"download"` which works with pnpm v10.21.0 and newer.~
> Had to keep `"warn"` until we can update scripts that use `npx`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": "11.5.2",
-      "onFail": "warn"
+      "version": "11.5.3",
+      "onFail": "download"
     },
     "runtime": {
       "name": "node",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,8 +8,8 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": "11.5.2",
-      "onFail": "warn"
```
mozzius, line 12: Hmm, does npx still work if we do this? Annoyingly we still have a bunch of scripts that use npx

ds-boyce: In that case I’ll have to revert this and just bump the version in `Dockerfile.embedr`. 😞

mozzius (approved):

PR #10888, opened 2026-06-12
[Chat] Restructure MessageItem to prevent alignment issues

mozzius:
Currently the avatar is just absolutely positioned, which leads to mismatches when the reaction bar changes height

Screenshot 2026-06-12 at 14 20 45

I changed it so it adds a fixed amount of bottom padding, and absolutely positions the reactions themselves

Screenshot 2026-06-12 at 14 20 11

This was of course a complete nightmare to restructure, so the diff is quite big. But while I was there, I capped the max number of emojis displayed to 10 rather than wrapping them

## Test plan

Every combination of:

- Platform
- Chat kind (direct/group)
- `isFromSelf`
- Multiple emoji
- Long press menu behaviour

### DIFF
--- a/src/components/dms/ActionsWrapper.tsx
+++ b/src/components/dms/ActionsWrapper.tsx
@@ -14,7 +14,6 @@ export function ActionsWrapper({
   children,
 }: {
   message: ChatBskyConvoDefs.MessageView
-  hasReactions?: boolean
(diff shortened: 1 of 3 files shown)


ds-boyce (approved):
Thanks for tackling this!

PR #10889, opened 2026-06-12
[Chat] Hide blocked accounts' reactions in chats

mozzius:
## Summary

Extends the existing "this message is hidden" blocking treatment to message reactions, in both directions (the reactor and the reactee).

Today a message from a blocked account gets hidden, but its reactions don't, and a blocked account's *reaction* on someone else's message still leaks their identity (and the reacted-to message's text) into the UI. This closes both gaps.

### Chat list (last-message preview)
- `getReactionInfo` now reports `isBlocked`, set when **either** the reactor **or** the author of the reacted-to message is blocked (in either direction). When the last reaction is blocked, `ChatListItem` skips it entirely and falls back to the last real message, rather than letting a blocked account's reaction bump the preview. (A last *message* from a blocked sender still shows the existing `This message is hidden` placeholder, as before.)

#### Before

Screenshot 2026-06-12 at 14 38 22

#### After

Screenshot 2026-06-12 at 14 39 01

### In-conversation
- New `filterBlockedReactions` helper drops reactions from blocked accounts. Applied to the reaction pills under a message (`MessageItem`) and to the reactions dialog (`ReactionsDialog`), so a blocked reactor's avatar/handle/emoji never render. `relatedProfiles` is shadow-synced by the convo agent, so this reflects optimistic blocks.

#### Before

Screenshot 2026-06-12 at 14 35 40

#### After

Screenshot 2026-06-12 at 14 35 58

### Shadow-cache sync
- The chat-list block checks (`getMessageInfo` / `getReactionInfo`) previously read `viewer.blocking` off the raw `convo.members`, which the profile shadow cache can't see - so an optimistic block didn't take effect in the preview until the convo list refetched. They now resolve through the already-shadowed `primaryProfile` via a shared `isDidBlockedInConvo` helper. Covers 1-1 convos and group messages/reactions from the owner immediately; other group members fall back to the raw member (unchanged behavior).

## Test plan
- [ ] Block a user you share a 1-1 chat with -> their last message preview reads "This message is hidden" immediately (no refetch wait), and a reaction from them no longer bumps the preview
- [ ] In a group chat, a blocked member's reactions disappear from the reaction pills and the reactions dialog
- [ ] A blocked account reacting to your message no longer surfaces their name in the chat list preview (the previous real message is shown instead)
- [ ] `pnpm typecheck` and `pnpm lint` pass

### DIFF
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -53,7 +53,11 @@ import {DateDivider} from './DateDivider'
 import {MessageItemEmbed} from './MessageItemEmbed'
 import {MessageItemInviteEmbed} from './MessageItemInviteEmbed'
 import {groupReactions} from './ReactionsDialog'
-import {CLUSTERED_MESSAGE_THRESHOLD_MS, MESSAGE_GAP_THRESHOLD_MS} from './util'
+import {
+  CLUSTERED_MESSAGE_THRESHOLD_MS,
+  filterBlockedReactions,
+  MESSAGE_GAP_THRESHOLD_MS,
+} from './util'
 
 const AVATAR_SIZE = 28
 const CLUSTERED_MESSAGE_GAP = 2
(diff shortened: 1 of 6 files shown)


ds-boyce (approved):
Good fix.

PR #10890, opened 2026-06-12
Add long-press menu to Chat tab for marking all as read

mozzius:
## Summary

Long-pressing the Chat tab in the native bottom bar now opens a menu with two options:

- **Mark all chats as read** - clears unread accepted conversations
- **Mark all requests as read** - clears unread message requests

Both reuse the existing `useUpdateAllRead` mutation that already powers the Inbox's "Mark all as read" button, including its optimistic updates and toasts.

Screenshot 2026-06-12 at 15 26 17

## Details

- Native-only by construction (web uses `BottomBarWeb`).
- Long-press is gated on `aa.flags.chatDisabled`, matching how the unread badge is already gated for age assurance.
- Uses `Menu.useMenuControl()` with an external control to open the menu on long-press, the same approach `ChatListItem` uses for `ConvoMenu`.
- Plays a haptic on long-press, consistent with the profile tab's long-press behavior.

## Test plan

- [ ] Long-press the Chat tab on iOS/Android with unread chats and confirm "Mark all chats as read" clears them
- [ ] Long-press with unread requests and confirm "Mark all requests as read" clears them
- [ ] Confirm the menu does not open when chat is disabled (age assurance)

### DIFF
--- a/src/view/shell/bottom-bar/BottomBar.tsx
+++ b/src/view/shell/bottom-bar/BottomBar.tsx
@@ -2,9 +2,8 @@ import {type JSX, useCallback} from 'react'
 import {type GestureResponderEvent, View} from 'react-native'
 import Animated from 'react-native-reanimated'
 import {useSafeAreaInsets} from 'react-native-safe-area-context'
-import {msg, plural} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
-import {Trans} from '@lingui/react/macro'
+import {plural} from '@lingui/core/macro'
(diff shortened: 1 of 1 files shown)


gaearon:
> that already powers the Inbox's "Mark all as read" button

Wait where’s that? I don’t have that on iOS in the inbox.

ds-boyce:
> Wait where’s that? I don’t have that on iOS in the inbox.

At the moment the “Mark all as read” button is only on web, and only for the inbox.

read

ds-boyce (approved):

PR #10892, opened 2026-06-12
Persist selected QR code color theme in invite friends dialog

vineyardbovines:
## Why

The color picker in the invite friends dialog kept its selection in component state, which resets whenever the sheet closes. Reopening the dialog always showed the default blue (day) theme regardless of what was previously picked.

## How

- Added an optional `inviteFriendsThemeKey` to the `Device` storage schema
- Added a `useInviteThemeKey` hook (same pattern as `useFollowersPromoDismissed`) that persists the selection on device and falls back to `day` when nothing is stored or the stored key is no longer a valid theme
- Swapped the dialog's `useState` for the hook

## Test plan

- `pnpm typecheck`, `pnpm lint`, prettier, and invite friends test suites pass
- Manual: open invite friends, pick a non-default color, close and reopen the sheet, and the selection is retained (including across app restarts)

### DIFF
--- a/src/features/inviteFriends/InviteFriendsDialogInner.tsx
+++ b/src/features/inviteFriends/InviteFriendsDialogInner.tsx
@@ -1,4 +1,4 @@
-import {Suspense, useRef, useState} from 'react'
+import {Suspense, useRef} from 'react'
 import {Pressable, View} from 'react-native'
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):

mozzius:
@vineyardbovines we good to merge this?

PR #10893, opened 2026-06-12
Reduce Sentry noise from logger transport and expected-failure call sites

vineyardbovines:
## Why

The vast majority of events in our Sentry project are strings forwarded by the logger transport rather than actual crashes or exceptions. Most of the top issues by volume are expected operational conditions (offline devices, session/token churn, third-party timeouts) reported at error level, which buries the real crash signal and makes the issue stream unusable. This PR cuts that noise at the source.

## What

**Transport (`src/logger/transports/sentry.ts`):**

- Only `error`-level string messages are sent to Sentry as events. `warn` and `log` strings are still recorded as breadcrumbs (and attached to any subsequent event), but no longer create issues of their own.
- The network-error filter now also checks the common metadata keys (`safeMessage`, `message`, `error`), since the underlying cause is usually passed there rather than in the message string. Previously, messages like "poll events failed" passed the filter even when the root cause was a network error.
- The same network-error filter now applies to the `captureException` path for `Error` objects.

**Call sites demoted from `logger.error` to `logger.warn`** (all are expected conditions, and most already excluded network errors but still produced large volumes from auth churn and upstream failures):

- session resume on launch (`App.native.tsx`, `App.web.tsx`)
- account switch (`ChooseAccountForm`)
- chat event polling (`state/messages/events/agent.ts`)
- feed latest polling (`PostFeed`)
- OTA update check (`useOTAUpdates`)
- push token registration (`notifications`)
- on-device translation failure (`lib/translation`)
- web video `play()` rejection (`VideoEmbed` web controls)

**Grouping fix:** `ContextMenu` logged a warning with a dynamic element id interpolated into the message, which created a new issue group per id. The id now goes in metadata.

Also prunes eslint suppressions made stale by the enum comparison fix in the transport.

## Test plan

- Updated `src/logger/__tests__/logger.test.ts`: warn/log strings assert breadcrumb-only behavior, added coverage for error-level strings being captured and for network errors being filtered via message, metadata, and exception paths.
- `pnpm test`, `pnpm typecheck`, `pnpm lint`, `prettier --check` all pass.

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -1058,11 +1058,6 @@
       "count": 1
     }
   },
-  "src/logger/transports/sentry.ts": {
-    "@typescript-eslint/no-unsafe-enum-comparison": {
-      "count": 3
-    }
-  },
   "src/logger/types.ts": {
     "@typescript-eslint/no-redundant-type-constituents": {
       "count": 1
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -124,7 +124,7 @@ function InnerApp() {
           await features.init
         }
       } catch (e) {
-        logger.error(`session: resume failed`, {message: e})
+        logger.warn(`session: resume failed`, {message: e})
       } finally {
         setIsReady(true)
       }
--- a/src/App.web.tsx
+++ b/src/App.web.tsx
@@ -103,7 +103,7 @@ function InnerApp() {
           await features.init
         }
       } catch (e) {
-        logger.error('session: resumeSession failed', {message: e})
+        logger.warn('session: resumeSession failed', {message: e})
       } finally {
         setIsReady(true)
       }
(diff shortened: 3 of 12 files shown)


src/lib/hooks/useOTAUpdates.ts
```diff
@@ -147,7 +147,7 @@ export function useOTAUpdates() {
         }
       } catch (err) {
         if (!isNetworkError(err)) {
-          logger.error('OTA Update Error', {safeMessage: err})
```
mozzius, line 150: I feel like I'd want to know about this one?

estrattonbailey: Yeah maybe let's keep this one

estrattonbailey (approved):
Nice 👍 yeah maybe let's keep the one error Sam mentioned. Just needs a conflict fix, then feel free to merge!

Issue #6815, opened 2024-11-28
User preference to automatically open the content warning (self-label) dialog when attaching media


nasso:
### Describe the Feature

TL;DR: Add a user preference checkbox that makes the content warning (self-label) modal dialog open automatically when media is added in the composer.

## Motivation

Content creators and artists who often post adult or graphic content complain that they tend to forget to set labels before publishing their work. Because self-labels can't be applied after the post was published, the only way is to wait on moderation services to catch that mistake, adding extra load on the shoulders of the moderation team.

After discussing with the relevant users, they've expressed their desire to have a more aggressive reminder that could be enabled in the user preferences.

## Proposed solution

I think the best way to address this is to make self-labeling part of the user flow when composing a post with media. I believe the best way to do that is to open the label selection dialog as soon as media is attached. Because this is opt-in, it shouldn't be annoying, especially for users who post a lot of media that needs self-labelling.

### Attachments

https://github.com/user-attachments/assets/f1250f41-1cd3-427f-82c2-279543d7ca8e

### Describe Alternatives

1. Show the dialog when clicking "Post". The dialog would need a "Cancel" button though, to go back to the composer. It's unclear whether "Cancel" would keep the label selection though.
2. Show a confirmation dialog when hitting "Post" if the post has media attached without any label.
3. Show a warning in the composer when the post has media attached without a label. This would be similar to the existing warning for missing ALT text, but without enforcing labels to be added. This is probably less effective.
4. Same as (3) but add a "safe" option in the label selection dialog form, and enforce it to be explicitly selected. This is Confusing™ and Weird™.

### Additional Context

Related: https://github.com/bluesky-social/social-app/issues/5948, https://github.com/bluesky-social/social-app/issues/5628

foxyoreos:
Copying over from #5628 

Just being able to set a default that I can override would be a large usability improvement, and from my experience on other social networks, NSFW artists would use the feature.

If I could set a default label, it would make posting on this network a lot less stressful, because if I post something SFW and accidentally forget to remove the label - oops, but no big deal. If I post something NSFW and forget to add the label, I have to wait for auto-moderation to catch it (if it even does), and in the meantime all of my followers get spammed with unlabeled porn. That's assuming I even notice the mistake.

In the meantime, because tensions on Bluesky are relatively high right now about NSFW art and labeling, I'm much more likely to get added to a bunch of "doesn't label their posts" blocklists if I ever make that mistake. >w<

[...]

Issue #5948, opened 2024-10-26
Add a setting for users to assign default content warning labels to all of their media posts


harkebark:
### Describe the Feature

Some accounts primarily share media with adult content labels. These accounts may accidentally share media without the proper content warnings, causing viewers to see content they've chosen to opt out of seeing. To help prevent this, there could be an account-wide setting to choose a set of default labels for every post a user makes. So for example, a user could choose to mark all of their posts as "suggestive" by default, and then any media they post would automatically have that label checked by default. If a user has this setting enabled and wants to post media without this label they would manually uncheck the label at post time. This would be a Bluesky equivalent to Twitter's "Mark all my media as sensitive media" setting.

Furthermore, if a user gets multiple reports for not properly labeling their content, this setting could automatically be applied as a precautionary measure and the user could be notified that they need to either tag their content or change their settings.

surfdude29:
The second part of this is similar to #5628.

Issue #5628, opened 2024-10-06
Situational adult content labelling reminders


qazmlp:
### Describe the Feature

If a user regularly posts adult content without self-labelling it (as determined by the moderation service), please nudge them into doing so.

What I would like to see is an informative popup in a supportive tone that explains the feature, why it's important (keeping sexually explicit content away minors, letting users set their own boundaries) _and that the label does not affect visibility in Bluesky's own algorithmic feeds_.

### Describe Alternatives

You could apply the adult content label to accounts who regularly don't self-label porn. I personally would like to avoid that becoming necessary, and if you do end up doing that as a measure of last resort, please do so with much higher threshold than the reminder (and ideally only after manual review, if that doesn't affect your moderators' well-being).

### Additional Context

Too many content creators are used to Twitter and other networks, where a sensitive content label significantly damages their reach, and if applied to their account comes with a search ban. I really want to avoid a similar situation coming about here.

The moderation service here on Bluesky seems to apply labels with some delay, especially on art where the content quite often can't be detected automatically. This means that depending on someone's social graph, a significant amount of nudity and porn can slip through even if they've hidden such content or set it to 'Warn'.

Personally I have set these labels to 'Warn', but I still have to be careful when browsing Bluesky chronologically in public, which makes for an inconvenient user experience.

(This feature request is written with some frustration. I apologise if that shines through.
I appreciate and support adult content creators, I'm just not always in the mood to view their works.)

qazmlp:
This is related to #5383. I think the measures would compliment each other. I think the reminder is necessary even if #5383 is implemented since some creators actually avoid the labels intentionally.

It hasn't happened to me, probably because I'm always super polite and usually indirect when reminding someone to label their posts, but some creators react aggressively when someone asks them to self-label.

foxyoreos:
This would be a large usability improvement.

Getting labels right on Bluesky and not forgetting them is more important because I can't edit labels after the fact. The current UI is so easy to miss. I haven't missed it yet because I'm paranoid about checking, but I can 100% tell you that at some point in the future I will be tired and I will miss the label. It's not a matter of if, it's a matter of when.

If I could set a default label, it would make posting on this network a lot less stressful, because if I post something SFW and accidentally forget to remove the label - oops, but no big deal. If I post something NSFW and forget to add the label, I have to wait for auto-moderation to catch it (if it even does), and in the meantime all of my followers get spammed with unlabeled porn. That's assuming I even notice the mistake.

[...]

PR #10898, opened 2026-06-13
Add reminder to label posts with media

### LINKED ISSUE #8406: "Label Post Reminder" Option
### Describe the Feature

A feature that would raise a pop-up warning the moment you hit "Post", reminding you to label your image before uploading it. 
Would help a ton with remembering that labeling exists, as the current UI places the option in a easy to forget/miss spot.

### Attachments

_No response_

### Require Labeling before posting
Similar to the "Require alt text before posting" Accessibility Setting, this option would not allow you to submit your post before adding a label to it. For this one to work, there would need to be a redundant "No Labels" option to choose from, so that you could still post images that don't require content labels.

### Additional Context

After seeing some discussion in Bluesky and some Discord servers I'm in, I decided to submit this feature request.

ThisIsMissEm:
I was having a discussion [with someone](https://bsky.app/profile/thisismissem.social/post/3mo5nnwa64k25) about issues facing adult creators using Bluesky and the "forgetting to add labels" issue came up, so here's an early implementation of adding a reminder to add labels before posting media (sans-making server-side changes to persist the setting across devices)

This fixes #8406, #5628, #6815 (in a slightly different way). Currently I've implemented the preference for this reminder using the persisted state, but if we like this feature, then it should probably be moved into the server-side backed `src/state/preferences/moderation-opts.tsx` (this requires API changes, so I wanted to experiment with the UI side first)

By allowing people using the app to receive a reminder to add labels to media they're posting, it should significantly cut down of the number of posts with media where adult content is not correctly labeled, and acts as a nudge to push people towards the behavior that we want (adult content is correctly labeled).

https://github.com/user-attachments/assets/87de2784-9743-4cd4-b73d-8ac348f650da

#5948 could be a nice follow-up to this, potentially, though I'm less certain of this.

## Open Questions
- is this something we want? (I think so)
- Should this also apply to links? (Since they're the other thing that triggers the labels visibility in the composer)

### DIFF
--- a/src/screens/Moderation/index.tsx
+++ b/src/screens/Moderation/index.tsx
@@ -12,6 +12,10 @@ import {
 } from '#/lib/routes/types'
 import {logger} from '#/logger'
 import {useIsBirthdateUpdateAllowed} from '#/state/birthdate'
+import {
+  useLabelReminderEnabled,
+  useSetLabelReminderEnabled,
+} from '#/state/preferences'
 import {useRemoveLabelersMutation} from '#/state/queries/labeler'
 import {
   useMyLabelersQuery,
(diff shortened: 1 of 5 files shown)


estrattonbailey:
I dig this, thanks for the work here! Will add this to the team agenda discuss early next week. I think you're right, would be ideal to save to server prefs 👍

ThisIsMissEm:
> I dig this, thanks for the work here! Will add this to the team agenda discuss early next week. I think you're right, would be ideal to save to server prefs 👍

More than happy to move it across to prefs, but figured it might be worth getting this into people's hands first to check the UX is right and helpful, before actually investing in the server settings

PR #10903, opened 2026-06-13
Message replies in chat

mozzius:
## What

Adds the ability to reply to a specific message in a chat conversation, a la Signal/WhatsApp/Discord. Linear: [Message replies](https://linear.app/blueskyweb/project/message-replies-080d468d3386/overview).

- **Start a reply** — "Reply" is the first item in the message context menu (native long-press + web hover).
- **Compose** — a quoted preview (sender + snippet + dismiss X) shows above the composer.
- **Send** — `replyTo: {messageId}` flows through the existing `sendMessage` path. An optimistic hydrated `replyTo` view is plumbed through the convo agent (mirroring `optimisticEmbedView`) so the quote renders instantly and survives the firehose round-trip.
- **In bubble** — a caption line ("↪ You replied to X") plus a tappable nested quote at the top of the bubble. Deleted originals show an italic "Message deleted" placeholder.
- **Tap to scroll** — tapping the quote scrolls to the original *if it's currently loaded*; no-op otherwise (no historical fetch, per scope).
- **Analytics** — `chat:message:reply:send` and `chat:message:reply:tap`.

Built on the new `chat.bsky.convo` reply lexicon (`@atproto/api` 0.20.15, bumped in a separate commit).

## Implementation notes

- Reply target lives in `MessageOverlays` (the existing `useMessageDialogs` context), so it's convo-scoped for free and can't leak across conversations. A small `Composer` bridge in `MessagesList` reads it inside the provider and injects it into the send call.

## Design

Matches the four Figma frames linked in the project. Nested-quote colors/spacing were styled from chat conventions rather than exact tokens — worth an eyeball on device.

## Test plan

- [x] `pnpm typecheck`, `pnpm lint`, chat test suite pass
- [ ] Manual: DM + group chat — long-press/hover → Reply → preview → send → bubble shows caption + quote + text → tap quote scrolls up
- [ ] Reply to a then-deleted message shows the deleted placeholder
- [ ] Tapping a reply to an out-of-window message is a no-op

## Open questions

- Metric names (`chat:message:reply:send` / `:tap`) — picked to match the existing `chat:`/`groupchat:` convention; sanity-check against what the data team expects.


ds-boyce:
> Reply to a then-deleted message shows the deleted placeholder

The message is deleted immediately, but the reply still contains the message text until the conversation is refreshed.

Everything else looks great! This is excellent.

ds-boyce (approved):
Confirmed the deletion case is working as intended now.

:shipit:

src/components/dms/MessageItem.tsx
```diff
@@ -648,3 +741,181 @@ function BlockedPlaceholder({
...
+    text = l`Deleted message`
+    subtle = true
+  }
+
+  return (
+    <Button
+      label={
+        senderName
+          ? l`Replied-to message from ${senderName}, tap to scroll to it`
+          : l`Replied-to message, tap to scroll to it`
```
mlocati, line 892: A "Replied-to message" is a reply to a message?

mozzius: It's the preview of the message that is being replied to

mlocati: So, it's something like
> Message from {senderName} that you're replying to; tap to scroll to it

(I'm asking just to understand how I can translate it into Italian)

mozzius: yep

mlocati: Thanks!

Issue #10681, opened 2026-06-01
App passwords not generating


jamtur01:
### Steps to Reproduce

1. Go to Settings -> Privacy & Security -> App Password.
2. Click + Add App Password
3. Click next - app password name box goes red and refuses to create app password. Have tried manual names, autogenerated names, nothing.

<img width="1366" height="706" alt="Image" src="https://github.com/user-attachments/assets/ce4ee25c-f47e-4f3d-b227-655bc2174de1" />

In log see:

```
enoki.us-east.host.bsky.network/xrpc/com.atproto.server.createAppPassword:1  Failed to load resource: the server responded with a status of 400 ()
```

Tried in incognito mode in both Chrome and Safari.

Version 148.0.7778.179 (Official Build) (arm64)

### What platform(s) does this occur on?

Web (Desktop)

### Device Info

MacOS 26.5 Chrome Version 148.0.7778.179 (Official Build) (arm64)

### What version of the app are you using?

Web

wingerter:
Hi, this looks like it may be the same underlying issue as [#7569](https://github.com/bluesky-social/social-app/issues/7569):

- identical symptom (name field border turns red, no error message, fails with both manual and auto-generated names)

---

**Two things that could help narrow it down:**

1. The server's actual error would help a lot: the 400 from `com.atproto.server.createAppPassword` should include a response body with an atproto error name + message (visible in the Network tab on that request)
_Could you paste it?_ That distinguishes input validation vs. a per-account app-password limit vs. something else server-side.

2. There also seems to be a client-side bug that hides the reason. In `AddAppPasswordDialog.tsx`, on a server error the input is marked invalid (`isInvalid={!!error}`) but the message is never shown: server errors aren't a DisplayableError, so the first `<Admonition>` is skipped, and the fallback `{!!apiError || (error && …)}` short-circuits to true (rendering nothing) whenever apiError is set.

So you get a red border with no explanation, similar to #7569. Surfacing the server's message there would at least make the failure self-explanatory.

---

Happy to open a PR for the client-side error-surfacing part if that'd be useful.

jamtur01:
```json
{"error":"InvalidToken","message":"Bad token scope"}
```

Same error for `privileged` true or false.

wingerter:
Thanks @jamtur01, that pins it down.

Two separate things are going on:

### 1. The server error. 
[...]

Issue #7569, opened 2025-01-24
Can not create new app passwords


techmagus:
### Steps to Reproduce

When creating a new app password, the form simply turns the field's box-border to red. No error messages. No app password created. I only have 8 currently.

Tried:
* Android app
* Web browser
* Using app generated name
* Random names
* Refreshed
* Re-logged
* Clear cache

![Image](https://github.com/user-attachments/assets/3a9dd3bd-8e49-4abe-a063-9fc29ec6863e)

![Image](https://github.com/user-attachments/assets/8266c412-2cf2-4a8d-919f-ccd283c20506)

### What platform(s) does this occur on?

Web (Mobile), Android

### What version of the app are you using?

1.96.7

krz42:
I had the same issue, and using incognito mode helped me create an app password. However, something is definitely wrong here.

PR #10904, opened 2026-06-13
Surface the server error when app password creation fails

wingerter:
## Summary

When creating an app password fails, the dialog marks the input invalid
(red border) but never shows *why*. Two things hide the reason:

- Server errors aren't a `DisplayableError`, so the first `<Admonition>`
  (gated on `error instanceof DisplayableError`) is skipped.
- The fallback was written as
  `{!!apiError || (error && !(error instanceof DisplayableError) && (…))}`.
  Because of `||`, once `apiError` is set this short-circuits to `true`
  and React renders the boolean, i.e. nothing.

The result is a red box with no explanation. That is what made #10681
(and the older #7569) hard to diagnose: the underlying `InvalidToken: Bad
token scope`, for example, was completely invisible to the user.

## Change

Render the actual error via `useCleanError`: prefer the cleaned message,
fall back to the raw atproto error, then to the existing generic string.
The dead branch (`error && !(error instanceof DisplayableError)`) is
dropped, since both `validationError` and `regexFailError` are always
`DisplayableError`s and are already handled by the admonition above.

Single-file change, no behavior change on the success path.

## Test plan

- Trigger a server-side failure (for example, create an app password from
  a session lacking full-access scope). The dialog now shows the server's
  message instead of a silent red border.
- Validation errors (empty / duplicate / invalid-char name) still render
  via the existing `DisplayableError` admonition.

Related: #10681, #7569


wingerter:
Hello @mozzius , since you tested my other fix, maybe if you have time, you could have a look into this fix? Thank you very much!

wingerter:
Hi @mozzius, gentle nudge on this one whenever you have a moment. It's a small single-file fix that surfaces the real server error in the Add App Password dialog instead of just an empty red border. Happy to adjust anything if needed. Thanks!

PR #10908, opened 2026-06-14
Fix missing chat notification avatar on Apple Watch

mozzius:
Chat notifications are upgraded to Communication Notifications and attach
the sender's avatar to the INPerson via downloadAvatarImage. The image was
built with INImage(imageData:), which renders on the iPhone but is not
relayed to paired devices like the Apple Watch. The Watch then falls back to
drawing a monogram from the sender's initials instead of the avatar.

Write the downloaded avatar bytes to a file in the shared App Group
container and back the INImage with INImage(url:) so the system can resolve
the image lazily across devices. Files are named by a stable hash of the
content-addressed thumbnail URL so notifications from the same sender reuse
one file.

### DIFF
--- a/modules/BlueskyNSE/NotificationService.swift
+++ b/modules/BlueskyNSE/NotificationService.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Intents
 import UIKit
 import UserNotifications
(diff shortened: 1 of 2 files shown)


mozzius:
Unfortunately this was complete nonsense

PR #10911, opened 2026-06-15
Reset textarea height after clear

ds-boyce:
`Composer`’s `clear()` calls `tapper.inputProps.onChangeText('')`, which writes directly to the `tapper` store and bypasses `AutosizedTextarea`’s `onChangeText`/resize logic. On Android the input height is pinned to `nativeHeight` state, and that state only updates from `onContentSizeChange` – which Android doesn’t fire on a programmatic clear. So the expanded height sticks.

To address it, we reset the height in `clear()` via a new imperative handle on `AutosizedTextarea` (`resetHeight`).

### DIFF
--- a/src/components/Composer/index.tsx
+++ b/src/components/Composer/index.tsx
@@ -34,6 +34,7 @@ import {
 } from '#/components/Autocomplete'
 import {
   AutosizedTextarea,
+  type AutosizedTextareaHeightApi,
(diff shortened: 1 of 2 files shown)


ds-boyce:
@claude please review

ds-boyce:
Confirmed working as intended on web.

mozzius (approved):
Seems to work!

PR #10913, opened 2026-06-15
Always scroll to bottom of chat after sending a message

ds-boyce:
Due to `maintainVisibleContentPosition={{minIndexForVisible: 0}}` anchoring item 0, appending content below the viewport doesn’t report a content-size change, so sending a new message while scrolled up never triggers a scroll down.

To ensure the desired behavior, a ref is used to track the send (`pendingSendScroll`). Then we scroll to the bottom of the chat if that ref is set when a new `'pending-message'` is appended.

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -201,11 +201,13 @@ export function MessagesList({
   const isAtBottom = useSharedValue(true)
 
   // Set when the local user sends a message so we follow it to the end even
-  // from a scrolled-up position. We can't rely on onContentSizeChange here:
-  // with maintainVisibleContentPosition anchored to item 0, appending a message
-  // below the viewport does not report a content-size change on native, so that
-  // callback never fires. Instead an effect watches the rendered item count and
-  // scrolls imperatively once the pending message lands (APP-2223).
+  // from a scrolled-up position. On native, onContentSizeChange can't be relied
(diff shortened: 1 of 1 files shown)


ds-boyce:
@claude please review

ds-boyce:
1. `pending items` are local-only optimistic sends so this should be fine.
2. This is also fine.
3. Updated the comment.
4. Added the teardown.

mozzius (changes requested):
This didn't seem to work for me, on iOS it doesn't scroll all the way to the bottom - seems like it's off by exactly the keyboard height

https://github.com/user-attachments/assets/c5d0acd5-7ee4-47b2-b7a9-4ef9e408639d

mozzius (changes requested):
This is just bizarre, but alas: it now ~works for single line text, but does not work at all for multi line text.

this is probably a race condition with react-native-keyboard-controller, if i had to guess?

is there any other way to approach this? maybe do it an event handler kinda way when the user sends it, rather than relying on useEffect? and stick a timeout on it?

https://github.com/user-attachments/assets/3b2d3de3-e019-4b31-9c12-4c23775eacf2

ds-boyce:
@claude please review

ds-boyce:
Ok the timeout isn’t perfect, but it should be more consistent than the previous approach.

@claude please review

mozzius (approved):
Works consistently now! nice one

PR #10915, opened 2026-06-16
Style follow buttons in expanded bundles, add client event on expand

abenzer:
Applies a more subtle style to the follow buttons that appear in expanded notification bundles. 

Also measures bundle expansions with a new client event `notifications:bundleExpand`

### DIFF
--- a/src/view/com/notifications/NotificationFeedItem.tsx
+++ b/src/view/com/notifications/NotificationFeedItem.tsx
@@ -147,14 +147,6 @@ let NotificationFeedItem = ({
     return ''
   }, [item])
 
-  const onToggleAuthorsExpanded = (e?: GestureResponderEvent) => {
-    if (e) {
-      e.preventDefault()
-      e.stopPropagation()
-    }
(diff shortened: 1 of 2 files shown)


src/view/com/notifications/NotificationFeedItem.tsx
```diff
@@ -1114,7 +1122,11 @@ function ExpandedAuthorProfileCard({
           <ProfileCard.FollowButton
             profile={author.profile}
             moderationOpts={moderationOpts}
-            logContext="ProfileCard"
+            logContext="NotificationExpandedProfileCard"
+            size="tiny"
+            variant={isFollowing ? 'ghost' : 'solid'}
+            color={isFollowing ? 'secondary' : 'primary_subtle'}
+            withIcon={isFollowing}
           />
```
mozzius, line 1130: imo we should hide this entirely if `!!author.profile.viewer?.following` is true - the _unshadowed_ version that is. this way, it'll be hidden when you're already following them to start with, but the button will still show the ghost state when you click on it

mozzius (approved):

PR #10919, opened 2026-06-16
[Chat] Poll for messages on chat push notification

mozzius:
Resolves APP-2415.

## What

Two changes to the DM short-polling system (`MessagesEventBus`):

1. **Stop polling entirely in the background.** Previously we kept a 5-minute background poll (`BACKGROUND_POLL_INTERVAL`), which the OS freezes on native anyway and just burns requests on an unfocused web tab. Every `Background` transition now calls `stopPoll()` instead of `resetPoll()`. Resuming already triggers an immediate catch-up poll (`resume()` -> `resetPoll()` -> `startPoll()` polls right away), so we don't miss anything on unbackground.

2. **Force an immediate poll on chat push.** New `pollNow()` method on the bus (dispatches the existing `UpdatePoll` -> `resetPoll()`), called from the foreground notification handler when a chat push arrives for the current account. The convo list / open convo now update on the next round-trip instead of waiting up to a full interval.

Both triggers reset the interval clock.

## Notes

- `pollNow()` is a no-op unless the bus is actively polling (`Ready`); in init/resume states those paths poll on their own.
- Added `useMaybeMessagesEventBus()` (null-tolerant) since the notification handler is always mounted, including when logged out, where the regular hook throws.
- The push trigger is native-only in practice (expo foreground handler). Web relies on refocus -> `resume()` -> immediate poll.
- Dropped the now-unused `BACKGROUND_POLL_INTERVAL` constant and the dead `Backgrounded` branches in the state machine / `getPollInterval`.

## Test plan

- [x] `pnpm typecheck`, `pnpm lint`, `pnpm test` all pass
- [ ] Manual (native, two accounts): with the app foregrounded on a non-Messages tab, send a DM from device B and confirm the list/unread update promptly (watch `DMsAgent` logs for `updatePoll` -> poll)
- [ ] Manual: background the app >1min, send a DM, foreground -> immediate poll fires
- [ ] Manual: confirm no `getLog` requests fire while backgrounded

### DIFF
--- a/src/lib/hooks/useNotificationHandler.ts
+++ b/src/lib/hooks/useNotificationHandler.ts
@@ -10,6 +10,7 @@ import {useAccountSwitcher} from '#/lib/hooks/useAccountSwitcher'
 import {logger as notyLogger} from '#/lib/notifications/util'
 import {type NavigationProp} from '#/lib/routes/types'
 import {useCurrentConvoId} from '#/state/messages/current-convo-id'
+import {useMaybeMessagesEventBus} from '#/state/messages/events'
 import {RQKEY as RQKEY_NOTIFS} from '#/state/queries/notifications/feed'
 import {invalidateCachedUnreadPage} from '#/state/queries/notifications/unread'
 import {truncateAndInvalidate} from '#/state/queries/util'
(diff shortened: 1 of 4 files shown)


mozzius:
@claude please review

PR #10927, opened 2026-06-16
Fix for `AutosizedTextarea` not resetting height after clear on Android 

estrattonbailey:
`onContentSizeChange` fires on Android, but since we set `{height: nativeHeight}` to force it to grow/shrink on Android, `nextHeight` will always be equal to `nativeHeight` and we'll never reset.

This adds a new prop `rawValue` which we check on Android. If it get's cleared out, we reset back to the pre-calculated `minInputHeight`.


ds-boyce (approved):
Better all the time.

PR #10929, opened 2026-06-16
Flash message bubble when tapping a reply

ds-boyce:
Flashes the message bubble instead of the background behind it.

https://github.com/user-attachments/assets/3abe3fe2-9680-470a-baf7-129a24da952c


mozzius (approved):
🔥

PR #10930, opened 2026-06-16
Increase chat bubble max width to 85% (was 80%)

ds-boyce:
| Before | After |
| - | - |
| before | after |


src/components/dms/util.ts
```diff
@@ -17,6 +17,8 @@ import * as bsky from '#/types/bsky'
 export const MESSAGE_GAP_THRESHOLD_MS = 60 * 60 * 1000
 export const CLUSTERED_MESSAGE_THRESHOLD_MS = 5 * 60 * 1000
 
+export const MESSAGE_BUBBLE_MAX_WIDTH = '85%'
```
mozzius: We should keep this at 80% on web. I picked this number to ensure there's space for the context menu/emoji button

mozzius (approved):

PR #10945, opened 2026-06-18
bskyweb: add isPartOf jsonld post attr

blackmichael:
## Summary

Adds schema.org `isPartOf` to post JSON-LD, and reduces the SSR post handler to a single appview call.

## `isPartOf`

Replies now emit `isPartOf` in their JSON-LD, pointing at the thread root's canonical URL so search engines know the reply belongs to a larger thread.

- Only on the top-level post, never on `comment[]` entries.
- Handle-form only: if the root's handle is unusable, `isPartOf` is omitted rather than pointing at a non-canonical DID URL.

## Fewer appview calls

The handler used to make up to three calls (profile + thread + root lookup); now it makes one in the common case.

- **Root from the thread.** `FeedGetPostThread` now uses `parentHeight=80`, so the ancestor chain comes back with the post. We walk it to resolve `isPartOf`, falling back to `FeedGetPosts` only for very deep or blocked threads. Note that this is functionally equivalent to what already exists in the code, because the `FeedGetPostThread` function would omit the `parentHeight` param when the value was zero, resulting in the default value of `80` being used.
- **No more `ActorGetProfile`.** AT-URIs accept a handle or DID, so we fetch the thread directly from the URL and read the canonical URL, handle, and display name from its author view.
- **Auth gating from the thread.** The appview includes the author's `!no-unauthenticated` label on the post, so we detect logged-out-restricted posts without a profile fetch. The check runs before any content is rendered, so nothing leaks.

## Tests

- `findRootPostInParents`: direct/multi-level/truncated/blocked chains, empty root.
- `postAuthorRequiresAuth`: nil view, nil author, self-label, foreign src, other val, negated.
- `isPartOf`: present on main post, omitted when empty, never on `comment[]`.

The handler itself isn't unit-tested (needs network) - worth a manual check against a `!no-unauthenticated` account before shipping.

### DIFF
--- a/bskyweb/cmd/bskyweb/jsonld.go
+++ b/bskyweb/cmd/bskyweb/jsonld.go
@@ -65,6 +65,7 @@ type discussionForumPosting struct {
 	CommentCount    *int64            `json:"commentCount,omitempty"`
 	Comment         []comment         `json:"comment,omitempty"`
 	IsBasedOn       string            `json:"isBasedOn,omitempty"`
+	IsPartOf        string            `json:"isPartOf,omitempty"`
 	SharedContent   *sharedContent    `json:"sharedContent,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


blackmichael:
@claude plz review

rafaeleyng:
Hey @blackmichael , I believe you can achieve this without the extra call.

You already have a call to `appbsky.FeedGetPostThread`. That gives you the full thread, with parents above (up to the root, see more on this later) and the children below.

You would have to do `threadView.Parent` then check whether it is `FeedDefs_ThreadViewPost`, and continue getting the `.Parent` recursively.

A couple of notes:
1. Would need to check whether this goes all the way to the root, unbounded (e.g., it doesn't stop early on very long threads. We have some limits we can pass to `app.bsky.feed.getPostThread` (`parentHeight`). The default of 80 is probably enough for >99.99% of the cases. You could also have 2 strategies: try to recurse all the way to the parent, and if not there, use the approach you have here.
2. If any of the intermediate parents is `notFoundPost` (e.g., a post that was deleted), then it would also break the parent chain. In that case, you would also fallback to this approach you have here.

So yeah, in general, you could try the optimized case, and it doesn't work, do the fallback.

rafaeleyng:
One note, I realized now that we call `appbsky.FeedGetPostThread` with `parentHeight` 0. I kind of assumed we used the default value. So in that case we would probably be increasing the cost of every call. But still, I believe it wouldn't cause any more work for the base case (no reply), and batching it on the same call would still be cheaper than doing the separate call for the reply case.

rafaeleyng (approved):
🚀

PR #10949, opened 2026-06-18
Fix embeds overlapping each other in chat

ds-boyce:
This negative margin is no longer needed and causes layout issues. Tested on Android, iOS, and web.

| Before | After |
| - | - |
| before | after |


estrattonbailey (approved):

estrattonbailey:
There's still some extra padding up top though, right? CleanShot 2026-06-18 at 12 25 44

ds-boyce:
It appears to be consistent between embeds and plain messages.

embed

message

ds-boyce:
Ah, I see. The embed has uniform padding inside it, but `QuoteEmbed` applies `a.mt_sm` to the content, which is unnecessary in chat, so “Bluesky” appears further from the top edge than intended.

ds-boyce:
Fixed.

margin

PR #10950, opened 2026-06-18
Display more information in chat replies

ds-boyce:
Rather than a catch-all “(contains embedded content)” for replies:
* Displays chat links.
* Displays external links.
* For all other embeds, displays the type of embed.

Also slightly reduces the padding above the original senders name and tweaks position of the X icon on reply previews, per design.

### DIFF
--- a/src/components/dms/replyPreview.ts
+++ b/src/components/dms/replyPreview.ts
@@ -1,0 +1,126 @@
+import {
+  AppBskyEmbedExternal,
+  AppBskyEmbedRecord,
+  AppBskyFeedDefs,
+  AppBskyGraphDefs,
+  AppBskyLabelerDefs,
+  type ChatBskyConvoDefs,
+  ChatBskyEmbedJoinLink,
+} from '@atproto/api'
+import {useLingui} from '@lingui/react/macro'
+
+import {BSKY_APP_HOST, toShortUrl} from '#/lib/strings/url-helpers'
+import {isKnownJoinLinkPreview} from '#/state/queries/join-links'
+
+/**
+ * Describes the embed of a quoted message that has no text of its own, so the
+ * reply preview can show what was shared instead of a generic placeholder. For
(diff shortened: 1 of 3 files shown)


src/components/dms/replyPreview.ts
```diff
@@ -0,0 +1,126 @@
...
+  return (message: ChatBskyConvoDefs.MessageView) => {
+    const text = message.text
+    if (text.trim()) {
+      return {text, subtle: false}
+    }
+
+    if (ChatBskyEmbedJoinLink.isView(message.embed)) {
+      const {joinLinkPreview} = message.embed
+      return {
+        text: isKnownJoinLinkPreview(joinLinkPreview)
```
estrattonbailey: Could handle expired links here too, `(expired chat invite link)`

src/components/dms/replyPreview.ts
```diff
@@ -0,0 +1,126 @@
...
+            message: '(starter pack)',
+            comment: 'A reply summary in chat',
+          }),
+          subtle: true,
+        }
+      case 'labeler':
+        return {
+          text: l({message: '(labeler)', comment: 'A reply summary in chat'}),
+          subtle: true,
+        }
```
estrattonbailey: Do we support any of these embeds right now?

src/components/dms/replyPreview.ts
ds-boyce: `isKnownJoinLinkPreview` checks for disabled and invalid links, but I can split on those types here with more specific messages.

src/components/dms/replyPreview.ts
ds-boyce: Nah, I just had them here for completeness. Removed.

ds-boyce:
Going to change to `(no text)` and `(deleted message)` for consistency.

estrattonbailey (approved):

PR #10953, opened 2026-06-18
Fix issue with stale unread message counts

ds-boyce:
The only thing that refreshes the new `getUnreadCounts` query is `debouncedInvalidateUnreadCounts()` inside the log-event handler, which means the count is only corrected if/when a log event happens to arrive, and even then it's gated by a 15-second stale time. As a result, the count could end up forever stale in some cases.

By optimistically updating the count (with rollback on error) we can ensure the badge stays in sync when conversations and requests are read.

Also updates the max “10+” badge value to use the same cap as the API (31, so “30+”).

### DIFF
--- a/src/state/queries/messages/conversation.ts
+++ b/src/state/queries/messages/conversation.ts
@@ -15,7 +15,11 @@ import {DM_SERVICE_HEADERS} from '#/lib/constants'
 import {STALE} from '#/state/queries'
 import {useOnMarkAsRead} from '#/state/queries/messages/list-conversations'
 import {useAgent} from '#/state/session'
-import {RQKEY_PARTIAL as UNREAD_COUNTS_PARTIAL_KEY} from './get-unread-counts'
+import {
+  RQKEY_PARTIAL as UNREAD_COUNTS_PARTIAL_KEY,
+  UNREAD_ACCEPTED_CAP,
+  UNREAD_REQUEST_CAP,
+} from './get-unread-counts'
 import {
   type ConvoListQueryData,
   getConvoFromQueryData,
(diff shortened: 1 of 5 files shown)


ds-boyce:
@claude please review

src/state/queries/messages/conversation.ts
```diff
@@ -74,11 +80,81 @@ export function useMarkAsReadMutation() {
...
+          old => {
+            if (!old) return old
+            return {
+              ...old,
+              ...(unreadStatus === 'request'
+                ? {
+                    unreadRequestConvos:
+                      old.unreadRequestConvos >= UNREAD_REQUEST_CAP
+                        ? old.unreadRequestConvos
+                        : Math.max(0, old.unreadRequestConvos - 1),
```
estrattonbailey, line 123: Why is this needed instead of just decrementing by 1?

src/state/queries/messages/get-unread-counts.ts
```diff
@@ -10,6 +10,14 @@ export const RQKEY = (includeGroupChats: boolean) =>
   [RQKEY_ROOT, includeGroupChats] as const
 export const RQKEY_PARTIAL = [RQKEY_ROOT] as const
 
+// the server sentinel-caps the badge counts: unreadAcceptedConvos maxes at 31
+// (meaning "more than 30") and unreadRequestConvos at 11 (meaning "more than
+// 10"). at the cap the value is no longer an exact count, so consumers must not
+// treat it as one - both the optimistic decrement and the badge display ceiling
+// key off these.
```
estrattonbailey, line 17: Oh I see so we need to expect optimistic updates pushing the counts higher

estrattonbailey (approved):
Ok makes sense to me!

PR #10038, opened 2026-03-11
Replace react-native-compressor with expo-bluesky-video-compress

mozzius:
## Summary
- New `expo-bluesky-video-compress` native module for video compression on iOS and Android
- Android: GL pipeline (decoder → SurfaceTexture → OpenGL ES → EGL → encoder) for proper hardware-accelerated scaling, frame rate capping to 30fps, reduced codec timeouts (10ms → 100μs), and encoder speed hints
- iOS: AVAssetReader/Writer pipeline with hardware VideoToolbox encoding, AAC audio passthrough
- Smart compression: probes video metadata and skips compression when bitrate ≤5Mbps, resolution ≤1920px, and file size ≤100MB
- Removes `react-native-compressor` dependency and its patch file

## Test plan
- [ ] Compress a high-res video (4K) on Android — verify correct orientation, reasonable file size (~3Mbps × duration), and faster than before
- [ ] Compress a portrait video (rotation metadata) on both platforms — verify orientation is preserved
- [ ] Compress a video already under thresholds — verify it's passed through without re-encoding
- [ ] Compress a 60fps video — verify output is 30fps
- [ ] Compress a .MOV file on iOS — verify it's converted to .mp4
- [ ] Cancel compression mid-way — verify cleanup
- [ ] Web video upload still works (no regressions from removing react-native-compressor)

> **Note**: Revert [`dfed94e`](https://github.com/bluesky-social/social-app/pull/10038/commits/dfed94ea67604c38282fce929a1eaf3afc42e41d) to see a comparison between old and new

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10954, opened 2026-06-18
Replace react-native-compressor video path with native expo-bluesky-video-compress module [APP-2428]

vineyardbovines:
## Summary

Adds a local Expo module that handles the video compression path previously served by `react-native-compressor`. `react-native-compressor` stays installed for now since `pickVideo.ts` and `VideoTranscodeBackdrop.tsx` still use its non-compression helpers (`getVideoMetaData`, `createVideoThumbnail`, `clearCache`) — fully removing the dep is a follow-up.

### tl;dr

- **Input**: anything ios/anroid can decode
- **Output**: always H.264, but a flip for HEVC is available.
- **Changes**: Compression on native will now skip for videos meeting these conditions:
  - Format is mp4/webm
  - Not HDR
  - File size <= 100 mb
  - Bitrate <= 5 mbps
  - Max dimensions <= 1920px 

### Codec selection

- `'auto'` (default) → h264, on both platforms. HLS pipeline plus HEVC patent licensing make h264 the safe choice for now.
- `'h264'` → h264 explicit.
- `'hevc'` → HEVC opt-in. Code path is fully implemented (hardware HEVC selection, software-HEVC ban on Android) and can be enabled later via feature flag in social-app without touching native code.

### Changes

- Drops `react-native-compressor`'s `Video.compress` / `getVideoMetaData` for the compression path.
- Adds probe-based smart-skip via `shouldCompress()`: clips already at or under 5 Mbps / 1920 px / 100 MB pass through unmodified. Matches the thresholds from the earlier PR #10038 attempt.
- Forces `codec: 'h264'` at the call site to make the production choice explicit (defends against accidental HEVC selection if defaults change).
- Same `CompressedVideo` return shape; no breaking change for callers.

## Test plan

- [ ] `pnpm install` and `cd ios && pod install` — verify the new module is picked up by `expo-modules-autolinking`.
- [ ] `pnpm ios` and `pnpm android` — full build/install on a physical device.
- [ ] Compose video upload, h264 source: verify output is h264, orientation correct, audio present, server accepts.
- [ ] Compose video upload, HDR/Dolby Vision source: confirm `probe()` returns metadata and `compress()` runs without crashing. Output should be SDR BT.709.
- [ ] Compose video upload, GIF: verify the `image/gif` early-return path still works.
- [ ] Compose video upload, already-compressed clip (≤5 Mbps, ≤1920 px, ≤100 MB): verify `shouldCompress` returns false and the source passes through.
- [ ] Cancellation: cancel mid-compression, verify cleanup (no leaked tmp files, no zombie encoders).
- [ ] Web upload: unchanged from before (uses `compress.web.ts`), but confirm no regression.


surfdude29:
Apologies for jumping in here, this looks like a very good PR, I just have a little suggestion as a result of f40ff6250a5b1bf44e9a9624b512ca968824d287.

Given that the explicit size gate of 100 MB that triggers compression on native has now been removed, and that the 5 Mbps threshold combined with the duration limit of 180 seconds mathematically means that any in-duration video larger than ~112.5 MB won't be passed through, I was thinking it might be a good idea to update the PR description to remove the references to the 100 MB size gate that is no longer present, to avoid any potential confusion from anyone who refers to this PR in the future :)

> **Changes**: Compression on native will now skip for videos meeting these conditions:
>  - Format is mp4/webm
>  - Not HDR
>  - File size <= 100 mb

> Adds probe-based smart-skip via `shouldCompress()`: clips already at or under 5 Mbps / 1920 px / 100 MB pass through unmodified.

> Compose video upload, already-compressed clip (≤5 Mbps, ≤1920 px, ≤100 MB): verify `shouldCompress` returns false and the source passes through.

estrattonbailey (approved):
Sick, that's perf, but yeah Claude might be right on the web import

PR #10955, opened 2026-06-18
Process videos on web with WebCodecs

vineyardbovines:
## Summary

- Use the WebCodecs API (via `mediabunny`) for web video transcoding when the browser supports it, with a fallback to the existing browser-API path.
- When WebCodecs is available, skip the upfront filesize gate so larger source videos can be processed in-browser.
- Replace the old `pickVideo.ts` / `pickVideo.web.ts` modules with a slimmer `metadata.ts` / `metadata.web.ts` split focused on getting video metadata; native still uses `react-native-compressor` so draft-restore keeps working.

## Test plan

- [x] Supported codecs: mp4/h264, mp4/avi, hevc, webm (some [test videos here](https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/1080/Big_Buck_Bunny_1080_10s_5MB.mp4))
- [x] Web (WebCodecs-capable browser, e.g. recent Chrome): pick a video larger than the legacy filesize limit and confirm transcoding + preview work.
- [x] Web (no WebCodecs, here's a link to [Firefox 129](https://ftp.mozilla.org/pub/firefox/releases/129.0/mac/en-US/)): confirm fallback to browser APIs still produces a valid preview and the legacy filesize gate still rejects oversized videos.
- [x] Web: GIF picker still works and the GIF preview renders.
- [x] Web: paste-a-video flow still produces a preview.
- [x] Web: restore a draft that contains a video.
- [ ] Native (iOS + Android): pick a video, confirm transcode/preview unchanged.
- [ ] Native: restore a draft that contains a video.


src/lib/media/video/compress.web.ts
```diff
@@ -1,56 +1,316 @@
...
 import {type CompressedVideo} from './types'
 
-// doesn't actually compress, converts to ArrayBuffer
+const TARGET_BITRATE = 3_000_000 // 3mbps, matches native
+const MAX_DIMENSION = 1920 // matches native
+const MIN_SIZE_FOR_COMPRESSION = 25 * 1000 * 1000 // 25mb, matches native
+
+// Codecs to try in order of preference
+// avc (H.264) is most compatible, vp9/vp8 are fallbacks for WebM
+const VIDEO_CODECS: VideoCodec[] = ['avc', 'hevc', 'vp9', 'vp8']
```
estrattonbailey: If these consts are shared, let's move them into a video specific const file someplace logical

src/lib/media/video/compress.web.ts
```diff
@@ -1,56 +1,316 @@
...
+    await conversion.execute()
+  } finally {
+    input.dispose()
+  }
+
+  const elapsed = performance.now() - startTime
+  const bytes = target.buffer
+
+  if (!bytes) {
+    throw new Error('MediaBunny compression produced no output')
```
estrattonbailey: When does this happen? If this is surfaced to the user we may want to use a better error message and log this to Sentry

src/lib/media/video/compress.web.ts
estrattonbailey: Should include the other consts in this file too probably? Or maybe the codecs and the container formats we want live next to these utils, that's fine too since they're kinda tied together.

estrattonbailey (approved):
LGTM

src/lib/media/video/compress.web.ts
vineyardbovines: Moved the three shared knobs (TARGET_BITRATE, MAX_DIMENSION, MIN_SIZE_FOR_COMPRESSION) into `src/lib/media/video/constants.ts` as `COMPRESSION_TARGET_BITRATE`, `COMPRESSION_MAX_DIMENSION`, `COMPRESSION_MIN_SIZE_BYTES` — same names #10954 will need so the file unions cleanly when both PRs land. Codec lists stayed local per your follow-up.

src/lib/media/video/compress.web.ts
vineyardbovines: This shouldn't happen — `BufferTarget.buffer` is null after `conversion.execute()` resolves cleanly, which is an internal-state edge in mediabunny. Not user-surfaced (the outer `compressVideo` catches and falls back to the original blob, line 62-66), but added a `logger.error` so Sentry captures the condition if it ever fires in the wild. Throw message now says "Compression produced empty output" — no implementation name in the user path.

PR #10956, opened 2026-06-18
Set a min height on native Menus

ds-boyce:
Stops single-item menus from being clipped on iOS.

| Before | After |
| - | - |
| before | after |

### DIFF
--- a/src/components/Menu/index.tsx
+++ b/src/components/Menu/index.tsx
@@ -6,12 +6,10 @@ import {
   View,
   type ViewStyle,
 } from 'react-native'
-import {msg} from '@lingui/core/macro'
-import {useLingui} from '@lingui/react'
-import {Trans} from '@lingui/react/macro'
+import {Trans, useLingui} from '@lingui/react/macro'
 import flattenReactChildren from 'react-keyed-flatten-children'
 
-import {atoms as a, useTheme} from '#/alf'
+import {atoms as a, native, useTheme} from '#/alf'
 import {Button, ButtonText} from '#/components/Button'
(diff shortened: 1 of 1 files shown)


src/components/Menu/index.tsx
```diff
@@ -110,7 +110,9 @@ export function Outer({
       <Dialog.Handle />
       {/* Re-wrap with context since Dialogs are portal-ed to root */}
       <Context.Provider value={context}>
-        <Dialog.ScrollableInner label={_(msg`Menu`)}>
+        <Dialog.ScrollableInner
+          label={l`Menu`}
+          contentContainerStyle={native({minHeight: NATIVE_MENU_MIN_HEIGHT})}>
```
ds-boyce: The core change is here.

estrattonbailey:
This is pry fine, but do we know why this menu doesn't just adapt to the height of the single button in the first place?

ds-boyce:
I don’t think so. In the past, we’ve added a “Cancel” or “Done” button to avoid the issue. I can spend some time trying to figure out the root cause.

ds-boyce:
Ok so it’s an issue specific to iOS 26 and how it [calculates the safe area for the bottom sheet](https://github.com/bluesky-social/social-app/blob/main/modules/bottom-sheet/ios/SheetViewController.swift#L36-L45). Setting `minHeight` is the smallest change necessary to address it without mucking around in the native module, which would affect every bottom sheet.

I further scoped the change to `ios`-only so at least we’re only applying it where it’s needed.

estrattonbailey (approved):
Sweet sg

PR #10961, opened 2026-06-19
(Prototype) Auto-detect hosting provider from handle on sign in

abenzer:
## What

When signing in, resolve the handle to its PDS and route the login to the correct server automatically, so people on non-Bluesky servers no longer have to manually pick a custom server.

This is a prototype meant to demonstrate the potential UX. Code changes and polish may be needed. Feedback welcome!

## Demo

https://github.com/user-attachments/assets/fa1fac1f-6614-4fab-87b1-a7cca6b0d478

## Changes

- **Auto-resolve on the login form.** As the user enters their handle, resolve it (handle → DID → DID doc → PDS) in the background and use that server on sign in. Bluesky-hosted accounts (bsky.social, `*.host.bsky.network`) are the default and not surfaced.
- **Status next to Sign in.** Shows "You're signing in to {host}" for non-Bluesky hosts, a spinner while resolving (only if it's actually slow), and a "Couldn't find your server → Use a custom server" fallback on failure.
- **Resolve on submit too.** Best-effort and timeout-bounded, so a fast submit or password-manager autofill can't beat the background lookup and wrongly default to bsky.social. Falls back to the default service on timeout/error.
- **Custom-server dialog hardening.** Validate the address and block "Done" on invalid input; truncate + cap the recent-server shortcuts.
- **Display safety.** Middle-truncate shown hosts so long/unparseable values can't blow out the layout.
- **Adds client events** for resolve success/failure and custom-server usage.

## Testing

1. **Bluesky account** — sign in with a normal `*.bsky.social` handle: no custom host message appears, signs in works exactly like it did before.
2. **Non-Bluesky account** — enter a handle hosted elsewhere: after a beat, "You're signing in to {host}" text appears and login routes to that PDS on submit. 
3. **Fast submit / autofill** — use a saved non-Bluesky login with auto-fill and auto-submit (e.g. with a password manager) and it should still route to the correct host and sign in correctly. 
4. **Slow / failed resolution** — throttle the network or use a bogus handle: spinner appears after ~600ms, then "Couldn't find your server" text appears with the custom server link.
5. **Custom-server dialog** (via "Use a custom server" / "Change"):
   - Try inputting something invalid like `localhost:2583asdasd` — clicking "Done" shows a validation error and the dialog doesn't close.
   - Valid input (`example.com`, `localhost:2583`) is accepted and reflected in the status message.
   - Recent-server shortcuts are displayed, capped at 5. 
6. **Long host names** — confirm long hostnames in the status message on the sign in form and shortcut chips in the custom server dialog truncate instead of overflowing.
7. **Email login** — entering an email skips resolution and uses the default service.


surfdude29:
Looking very nice!

Two UX suggestions to consider:
- For us slow typers, is it possible to increase the time before the "Couldn't find your server" message appears? Otherwise it might be confusing to users if it pops up while they're typing, e.g.:

https://github.com/user-attachments/assets/8a29b191-d69d-4d51-aeb8-e5fe6a4fc028

- I think showing the hosting provider in bold would be helpful :)

mozzius:
@github-actions ota

PR #10970, opened 2026-06-21
WSOD when trying to create account from starter pack on non-default provider

c960657:
An infinite update loop in `useSetActiveStarterPack` causes WSOD when trying to create account from starter pack on non-default provider.

### Steps to reproduce
1. When logged out, go to a starter packer, e.g. https://bsky.app/starter-pack/bsky.app/3livawpa7d326
2. Click the _Join Bluesky_ button.
3. Next to _You are creating an account on_, click on _Bluesky Social_.

## Expected result
The _Choose your account provider_ dialogue opens.

## Actual result
White screen of death.

## Fix
`useSetActiveStarterPacks` returned a callback that closed over the current landing (for debugging logging) and wrote a fresh object on every call. Each call changed the landing, which gave the callback a new identity, which re-ran `useLandingEntry`'s effect, which called it again – a loop React aborts with "Maximum update depth exceeded".

Read the previous landing via the functional updater instead of closing over it, so the callback only depends on the stable `setLanding` and keeps a stable identity.

### DIFF
--- a/src/state/shell/landing.tsx
+++ b/src/state/shell/landing.tsx
@@ -1,4 +1,10 @@
-import {createContext, useContext, useState} from 'react'
+import {
+  createContext,
+  type Dispatch,
+  type SetStateAction,
+  useContext,
+  useState,
+} from 'react'
 
 import {logger} from '#/logger'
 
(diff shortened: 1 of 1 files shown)


c960657:
A simpler fix would be to just remove the `logger.debug` call.

mozzius (approved):

PR #10972, opened 2026-06-22
Add new chat notification preferences

ds-boyce:
Two new app settings for message and message request notifications:

https://github.com/user-attachments/assets/3a2005f5-b240-40f6-a1ae-8197c5bdaeba

Both are also available via chat settings for convenience:

https://github.com/user-attachments/assets/290d4711-532c-48bf-8e2f-527936330e94

### DIFF
--- a/src/components/dialogs/NotificationSettingsDialog.tsx
+++ b/src/components/dialogs/NotificationSettingsDialog.tsx
@@ -1,8 +1,10 @@
 import {View} from 'react-native'
-import {type AppBskyNotificationDefs} from '@atproto/api'
 import {Trans, useLingui} from '@lingui/react/macro'
 
-import {useNotificationSettingsQuery} from '#/state/queries/notifications/settings'
+import {
+  type NotificationSettingsPreferenceName,
+  useNotificationSettingsQuery,
+} from '#/state/queries/notifications/settings'
 import * as SettingsList from '#/screens/Settings/components/SettingsList'
 import {PreferenceControls} from '#/screens/Settings/NotificationSettings/components/PreferenceControls'
 import {atoms as a, useTheme, web} from '#/alf'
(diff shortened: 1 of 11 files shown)


ds-boyce:
@claude please review

ds-boyce:
> Chat service failure leaves the dialogs spinning forever (medium)

Addressed by splitting the queries.

> Does the chat putPreferences endpoint accept partial updates? (please confirm)

Confirmed.

> App-level chat preference is now silently dropped

Intentional, added a comment.

src/components/dialogs/NotificationSettingsDialog.tsx
```diff
@@ -55,7 +59,13 @@ function NotificationSettingsDialogInner({
 }: Omit<NotificationSettingsDialogProps, 'icon'>) {
   const t = useTheme()
   const {t: l} = useLingui()
-  const {data: preferences, isError} = useNotificationSettingsQuery()
+  const isChat = isChatPreferenceName(name)
+  const appQuery = useNotificationSettingsQuery({enabled: !isChat})
+  const chatQuery = useChatNotificationSettingsQuery({enabled: isChat})
+  const isError = isChat ? chatQuery.isError : appQuery.isError
+  const preference = isChatPreferenceName(name)
```
estrattonbailey: Duplicate check here :P

src/screens/Messages/Settings.tsx
```diff
@@ -237,6 +250,54 @@ export function MessagesSettingsScreenInner({}: Props) {
...
+                </Text>
+                <NotificationPreferenceSubtitle
+                  preference={chatNotificationSettings?.chat}
+                  isLoading={!chatNotificationSettings}
+                  isError={chatSettingsError}
+                />
+              </View>
+              <ChevronRightIcon style={[a.ml_2xs, t.atoms.text]} size="lg" />
+            </Toggle.Item>
+            <Toggle.Item
```
estrattonbailey: Wait, these shouldn't be `Toggle.Item`, they should just be a pressable. Their aria type is a checkbox rn, just needs to be a button. Same for data export, which predates this PR.

estrattonbailey (approved):
Looks great, couple of small things

PR #10977, opened 2026-06-24
Add icon to AppLanguageDropdown

iLynxcat:
Related to #10978 

Adds the Earth icon (from the Languages page in Settings) to the language picker.

The goal is to signify that "this is the language picker" without requiring the user to parse "English"

**Also bundled in:** Improves the language picker to use `cursor: pointer` for hovered items in the list. Right now, it uses the default cursor and switches to the I-beam when hovering the text of an item.

### RightNav

Logged-out feed with expanded right nav

### LeftNav

Logged-out feed

### Signup

Signup page

### SplashScreen

Splash screen


estrattonbailey (approved):
Nice I dig 👍

PR #10978, opened 2026-06-24
signup: align "Contact support" to the right on larger displays

iLynxcat:
Related to #10977

Realigns the "Contact support" text to the right on the signup flow. Improves clarity versus squishing it against the language picker

**Only tested on web**

## Screenshots

Before:
Signup flow with left aligned support hint, next to language picker

After (this PR):
Signup flow with right aligned support hint, below Next button


iLynxcat:
updated to only change the alignment when gtMobile breakpoint passes, preserves design on mobile & smaller screens

estrattonbailey (approved):
Thanks!

PR #10982, opened 2026-06-24
Focus composer input when editable

ds-boyce:
The `autoFocus` prop only has an effect when the component is first mounted, so if `editable` is `false` initially, as it is with the chat composer while we show the loading state, then the input never receives focus.

Using an effect that focuses the input when `editable` is `true` addresses the issue.


estrattonbailey (approved):

ds-boyce:
> One thing worth a manual confirm (non-blocking, web keyboard nav):

Tested this locally and focus returns to the trigger as intended.

PR #10985, opened 2026-06-24
Increase unread cap to 99+

ds-boyce:
The cap was increased here: https://github.com/bluesky-social/atproto/pull/5145

Also no longer artificially bumps the count by 1 if there are one or more incoming chat requests. The count simply and accurately reflects the number of unread messages.

Also fixes the `hasUnreadConvos` condition for the “Mark all as read” button for requests, which was using the first page of results instead of the new count query.

### DIFF
--- a/src/screens/Messages/components/InboxRequests.tsx
+++ b/src/screens/Messages/components/InboxRequests.tsx
@@ -1,11 +1,13 @@
 import {plural} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react/macro'
 
-import {UNREAD_REQUEST_CAP} from '#/state/queries/messages/get-unread-counts'
 import {atoms as a} from '#/alf'
 import {ButtonIcon, ButtonText} from '#/components/Button'
 import {Inbox_Stroke2_Corner2_Rounded as InboxIcon} from '#/components/icons/Inbox'
 import {Link} from '#/components/Link'
+
+// The server caps unreadRequestConvos at 11, where 11 means "any more than 10".
+const REQUEST_COUNT_CAP = 11
 
 export function InboxRequests({
   count,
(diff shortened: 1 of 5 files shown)


src/screens/Messages/components/InboxRequests.tsx
```diff
@@ -30,8 +28,8 @@ export function InboxRequests({
       })
     : overflow
       ? l({
-          message: `10+ requests`,
-          comment: 'Displayed when the number of requests is greater than 10',
+          message: `${UNREAD_REQUEST_CAP - 1}+ requests`,
+          comment: 'Displayed when the number of requests exceeds the cap',
```
surfdude29, line 32:
```suggestion
          comment: 'Displayed when the number of requests exceeds the cap – for example, 99+ requests',
```
nit: Suggest amending the comment for translators here to give an example of what the string will be with the cap set at its current level

src/view/shell/bottom-bar/BottomBar.tsx
```diff
@@ -242,12 +242,17 @@ export function BottomBar({navigation}: BottomTabBarProps) {
...
-                  ? l({
-                      message: plural(numUnreadMessages.numUnread ?? 0, {
-                        one: '# unread item',
-                        other: '# unread items',
-                      }),
-                    })
+                  ? numUnreadMessages.numUnread?.includes('+')
+                    ? l({
+                        message: `${numUnreadMessages.numUnread} unread items`,
+                        comment: 'For example, 99+ unread items',
```
surfdude29:
```suggestion
                        comment: 'Accessibility hint for the bottom bar chat icon when the number of unread messages exceeds the cap, with the + symbol already included – for example, 99+ unread items',
```
nit: Suggest expanding this comment for translators to add clarity on context and mention explicitly that the `+` symbol is already included

estrattonbailey (approved):

Issue #10966, opened 2026-06-20
Rename the new 'Invite Friends' dialog on native to 'Share Profile'


RetroSunstar:
### Describe the Feature

Currently the new share button on the header of a users own profile will open a menu entirely referrred to in the format of inviting a friend

Every function in this new menu can be used outside of a friends context, but more importantly nothing about this menu is of an "inviting" nature

All this menu does is give you a link to your profile, and nothing about a link to the users profile would trigger a response that's on-par with inviting them to something

The 'Share Profile' naming scheme is just simply accurate for this

### Attachments

[Bluesky.png](https://github.com/user-attachments/assets/c28aa034-c897-492e-8888-a9ec39e4ef7e)
[Example.png](https://github.com/user-attachments/assets/53c1edb3-e7e2-4e18-a823-77138adf4639)

### Additional Context

Sharing a QR code can be useful for content creators, and thus the scan QR code function works to scan things on a screen quickly (not just in person scanning)

The 'Share Link' button has the exact same function as the 'Share via...' menu item and takes the same number of clicks to get to

The copy link field is just a box with the link to your profile

There is no prompt to follow the user, there is no prompt to make an account, and there is nothing outside of the experience of clicking the task bar and putting that link somewhere else

Even the icon for this feature is usually understood to mean sharing something

While changing the name of single piece of text in a header is super simple, it'd be more about the filenames and if those should be changed too 🐙

PR #10987, opened 2026-06-25
Change dialog label and header from 'Invite Friends' to 'Share Profile'

gabecodez:
Related to #10966 

Here is how I went about changing it, but if this were to be a change that the team wants to do, should all references to the InviteFriends be changed to ShareProfile or something similar as OP mentioned or should only the on-screen text be changed?

(I am very new to contributing on GitHub and am trying to learn how to be better. Please do not hesitate to offer critique or close this Pull Request. Thank you!)


abenzer:
@estrattonbailey i agree with this change

PR #10989, opened 2026-06-25
Hide autocomplete suggestions when input loses focus

ds-boyce:
Fixes an issue where it would remain visible if you opened the composer after entering a search term.

https://github.com/user-attachments/assets/fc28051a-4857-41f5-b562-5ec079d5cc22


ds-boyce:
> One thing worth verifying before merge (non-blocking)

Verified this locally.

estrattonbailey (approved):

PR #10994, opened 2026-06-25
Add app language: Czech (`cs`)

surfdude29:
Czech (`cs`) is now at 100% translated [on Crowdin](https://bluesky.crowdin.com/bluesky-social), so this PR proposes adding it as an app language.

[KWS docs](https://dev.epicgames.com/docs/kids-web-services/kws-localization#languages-supported-by-kws-apis) say that Czech is supported by KWS APIs so this PR adds it to `KWS_SUPPORTED_LANGS` to enable localization of web UI screens and user-facing emails from KWS.


estrattonbailey (approved):
Nice! Good call on KWS too 👍

PR #10995, opened 2026-06-26
mitigação conexão tls mal configurada

enzolda:


estrattonbailey:
@enzolda thanks for looking into this! This is an internal service and we're in the process of deprecating it, so we'll leave this be for now.

PR #10997, opened 2026-06-26
Use RNKC for composer keyboard avoiding view

mozzius:
Couple of motivations here:

- Current system seems flaky on Android on initial load
- Current one isn't animated
- Has janky conditional logic based on `isKeyboardVisible` which is opposite per platform
- Seems to get flakier on new arch

Adopts `react-native-keyboard-controller`'s `KeyboardAvoidingView` instead, which is much less janky and allows me to remove the conditional logic around padding


src/view/com/composer/Composer.tsx
```diff
@@ -798,17 +796,9 @@ export const ComposePost = ({
...
-        // keyboard is closed, so we subtract that in the offset and add it back
-        // here when the keyboard is open
-        (IS_ANDROID && isKeyboardVisible)
-          ? insets.bottom
-          : 0,
+      paddingBottom: insets.bottom,
     }),
-    [insets, isKeyboardVisible],
+    [insets.top, insets.bottom],
   )
```
mozzius, line 870: goodbye horrible horrible code

src/view/com/composer/Composer.tsx
```diff
@@ -2361,24 +2351,31 @@ function useScrollTracker({
...
+  const insets = useSafeAreaInsets()
 
-  // Android etc
-  if (!IS_IOS) {
-    // need to account for the edge-to-edge nav bar
-    return bottom * -1
-  }
+  // the keyboardavoidingview has bottom padding to avoid being obscured by the safe area when keyboard is closed.
+  // however, this leads to a gap when the keyboard is open. we account for that by subtracting the bottom inset when open.
+  let keyboardVerticalOffset = insets.bottom * -1
```
mozzius, line 2422: same behaviour for all platforms now, much easier

vineyardbovines (approved):
heck yes, love this lib. looks good- no regressions on my iPhone 17 Pro or Samsung Galaxy

PR #10999, opened 2026-06-26
Fail video duration check before compression

vineyardbovines:
- Probe video duration in `selectVideo` when missing (share-extension deeplinks ship only `uri|width|height`)
- Reject before `processVideo` when duration exceeds `VIDEO_MAX_DURATION_MS`, surfacing the error in the composer inline error state
- Apply the same gate to `restoreVideo`
- Normalize native `getVideoMetadata` duration to milliseconds to match `expo-image-picker`'s convention (was seconds)

Without this, a long video selected through the share extension or web paste would run through compression first and only get rejected afterward.

### DIFF
--- a/src/view/com/composer/videos/metadata.ts
+++ b/src/view/com/composer/videos/metadata.ts
@@ -16,7 +16,11 @@ export async function getVideoMetadata(
     mimeType: extToMime(metadata.extension),
     width: metadata.width,
     height: metadata.height,
-    duration: metadata.duration,
+    /*
+     * react-native-compressor reports seconds; the rest of the app treats
+     * `ImagePickerAsset.duration` as milliseconds (matching expo-image-picker).
+     */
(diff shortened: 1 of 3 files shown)


src/view/com/composer/videos/metadata.ts
```diff
@@ -16,7 +16,11 @@ export async function getVideoMetadata(
     mimeType: extToMime(metadata.extension),
     width: metadata.width,
     height: metadata.height,
-    duration: metadata.duration,
+    /*
+     * react-native-compressor reports seconds; the rest of the app treats
+     * `ImagePickerAsset.duration` as milliseconds (matching expo-image-picker).
+     */
+    duration: metadata.duration * 1000,
```
estrattonbailey, line 23: rip

estrattonbailey (approved):
bango 🎉

PR #11001, opened 2026-06-27
Fix 149/150 user limit for starter packs

RetroSunstar:
Currently users can only have 149 people in a starter pack, this number includes the pack creator themselves

This looks to be a simple math issue where the disabled const is based on a ``STARTER_PACK_MAX_SIZE - 1`` comparison 🐙


mozzius:
@claude is this true

mozzius (approved):

PR #11007, opened 2026-06-29
Clean up media-upload temp files on iOS

### LINKED ISSUE #11006: On iOS, media-upload temp files seem to build up in Caches and aren't freed by "Clear image cache"
### Steps to Reproduce

1. On iOS, attach some images/videos to posts and post (or cancel) a few times.
2. Go to Settings -> About and tap "Clear image cache".

What I'd expect is for that button to free the space used while preparing the attached media. From what I can tell, though, it only clears the small `expo-image` cache - the larger items (full-resolution copies of the picked photos/videos and the intermediate files from image processing) seem to stick around, and I couldn't find an in-app way to clear them other than reinstalling.

Here's what I measured on an iOS Simulator (dev build) after a handful of uploads:

```
 24M  Library/Caches/ImageManipulator/            (looks like it's never cleaned - expo-image-manipulator)
9.9M  Library/Caches/ImagePicker/                 (copies of picked photos/videos - also seems to stay)
860K  Library/Caches/com.hackemist.SDImageCache/  (this is what the button clears)
```

After tapping "Clear image cache", `SDImageCache` went from ~860K to ~8K, but `ImageManipulator/` and `ImagePicker/` were unchanged. They also seem to remain after the composer is closed (`bsky-composer/` was 0 B at the time), so it looks like nothing is removing them.

I haven't been able to confirm this directly on the App Store build (the container isn't inspectable there), but since this is the same JS path in both builds, I suspect the same thing happens in production - which might explain installs that grow quite large over time.

### Attachments

_No response_

iOS

### Device Info

iOS Simulator (iPhone 17, iOS 26.x), dev build of `main`

1.125.0 (dev build of `main`)

### Additional Information

A bit of digging into where this might come from:

- `purgeTemporaryImageFiles()` in `src/state/gallery.ts` only deletes `bsky-composer/`. The full-resolution files that `expo-image-manipulator` / `expo-image-picker` leave in `Library/Caches/{ImageManipulator,ImagePicker}/` don't appear to be cleaned anywhere.
- The "Clear image cache" button calls `Image.clearDiskCache()`, which only covers `com.hackemist.SDImageCache/`.

A couple of ideas that might help (happy to be wrong on any of this):

1. Extend `purgeTemporaryImageFiles()` (it already runs when the composer closes) to also clear `ImageManipulator/` and `ImagePicker/`, so things don't keep accumulating.
2. Have the existing "Clear image cache" button also call that, and maybe rename it to "Clear cache" - that would give people whose install is already large a way to recover the space.
3. (smaller follow-up) Compressed video output for videos over 25 MB is written to `tmp/` by `react-native-compressor` and doesn't seem to be removed after upload.

so-asano:
Fixes #11006

## What

When attaching media to a post, `expo-image-picker` copies the picked photos/videos into `Library/Caches/ImagePicker/` and `expo-image-manipulator` leaves intermediate full-resolution outputs in `Library/Caches/ImageManipulator/`. Nothing cleaned these up, so on iOS - where the OS exposes no "clear cache" - they accumulated indefinitely, one full-resolution copy per attached item. The existing "Clear image cache" button only flushed the `expo-image` disk cache (`SDImageCache`), so users had no way to reclaim this space short of reinstalling.

## Changes

- Extend `purgeTemporaryImageFiles()` (already runs on composer close) to also clear the `ImagePicker/` and `ImageManipulator/` cache dirs, stopping the leak going forward.
- Make the settings button also call `purgeTemporaryImageFiles()` and rename it from "Clear image cache" to "Clear cache", so already-bloated installs can recover the space.

## Testing

Verified on iOS Simulator:

- A few image/video posts left ~34 MB in `ImageManipulator/` + `ImagePicker/` that the old button could not free.
- With this change, tapping the button dropped total cache from ~44 MB to ~5 MB (the remainder is OS-managed `NSURLCache`).
- After posting, `ImageManipulator/` and `ImagePicker/` are empty once the composer closes.

## Notes

- iOS only; Android already exposes a system "clear cache".
- Follow-up (not in this PR): compressed video output for videos > 25 MB is written to `tmp/` by `react-native-compressor` and isn't removed after upload. See #11006.


src/screens/Settings/AboutSettings.tsx
```diff
@@ -124,11 +131,11 @@ export function AboutSettingsScreen({}: Props) {
           {IS_NATIVE && (
             <SettingsList.PressableItem
               onPress={() => onClearImageCache()}
-              label={_(msg`Clear image cache`)}
+              label={_(msg`Clear cache`)}
```
estrattonbailey: Do we need to change this strings? It creates churn for the translators.

vineyardbovines (changes requested):
looks good to me! only request is reverting the translation changes, i agree with @estrattonbailey that its churn for translators. actual code itself is all good

src/screens/Settings/AboutSettings.tsx
so-asano: Thank you, reverted to the existing "image cache" strings.

vineyardbovines (approved):
🚀

so-asano:
Hi @estrattonbailey, I reverted the string change as discussed, and @vineyardbovines has approved the PR.

Is there anything else needed from my side?

vineyardbovines:
@so-asano i just approved the workflows for you- once they pass you can hit that big green button!

so-asano:
@vineyardbovines Pushed an update — could you approve the workflows again?

so-asano:
Thanks @vineyardbovines! All checks are green and there are no conflicts. I don't have merge access though - could you merge it?

PR #11017, opened 2026-06-29
Avoid leaking previous messages via replies

ds-boyce:
Replies to messages sent before a user joined a chat should not include the previous message content.

https://github.com/bluesky-social/atproto/pull/5137

### DIFF
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -78,6 +78,7 @@ function messageIsReply(
   message:
     | ChatBskyConvoDefs.MessageView
     | ChatBskyConvoDefs.DeletedMessageView
+    | ChatBskyConvoDefs.MessageBeforeUserJoinedGroupView
     | null,
 ): boolean {
   return (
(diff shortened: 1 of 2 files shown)


src/components/dms/MessageItem.tsx
```diff
@@ -74,16 +74,19 @@ const BORDER_RADIUS = 20
...
+  | ChatBskyConvoDefs.MessageView
+  | ChatBskyConvoDefs.DeletedMessageView
+  | null
+
 function messageIsReply(
-  message:
-    | ChatBskyConvoDefs.MessageView
-    | ChatBskyConvoDefs.DeletedMessageView
-    | null,
+  message: MessageItemNeighbor | ChatBskyConvoDefs.MessageView,
```
estrattonbailey: Dupe type here?

src/components/dms/MessageItem.tsx
```diff
@@ -751,28 +753,56 @@ function ReplyCaption({
...
+    caption = l`Someone replied`
+  }
 
   return (
     <Button
-      label={l`Scroll to the message this is replying to`}
+      label={
+        onPress
+          ? l`Scroll to the message this is replying to`
+          : l`Reply to a message sent before you joined`
```
estrattonbailey: Sounds like the verb form of `reply`. Maybe `A reply...` is better?

estrattonbailey (approved):
LGTM, couple of nits

PR #11026, opened 2026-06-30
Add featured feed tab to profiles

DGaffney:
Let accounts designate one of their feeds as a "featured feed" that greets visitors on their profile. When set, the feed appears as the first tab (to the left of Posts), labeled with the feed's name, so visitors land on a curated feed instead of the reverse-chronological post list. Posts/Replies/Media remain one tap away, and the feature is fully opt-in.

The selection is stored as a custom `featuredFeed` field on the app.bsky.actor.profile record (alongside pinnedPost). Since the AppView does not surface custom fields, viewers read it from the repo record directly via com.atproto.repo.getRecord.

- profile-featured-feed.ts: query the raw profile record and derive the featured feed uri
- profile.ts: persist the featuredFeed field through the profile update mutation and invalidate the record query on success
- EditProfileDialog + FeaturedFeedPickerDialog: pick a featured feed from the feeds the account has created
- Profile.tsx: render the featured feed as the first profile tab
- feed.ts: allow useFeedSourceInfoQuery to be disabled

Claude-Session: https://claude.ai/code/session_01EF2br6faLvA5G6pScFiSzg

### DIFF
--- a/src/screens/Profile/Header/EditProfileDialog.tsx
+++ b/src/screens/Profile/Header/EditProfileDialog.tsx
@@ -10,7 +10,12 @@ import {cleanError} from '#/lib/strings/errors'
 import {isOverMaxGraphemeCount} from '#/lib/strings/helpers'
 import {logger} from '#/logger'
 import {type ImageMeta} from '#/state/gallery'
+import {useFeedSourceInfoQuery} from '#/state/queries/feed'
 import {useProfileUpdateMutation} from '#/state/queries/profile'
+import {
+  useFeaturedFeedUri,
+  useProfileRecordQuery,
+} from '#/state/queries/profile-featured-feed'
 import {ErrorMessage} from '#/view/com/util/error/ErrorMessage'
 import {EditableUserAvatar} from '#/view/com/util/UserAvatar'
 import {UserBanner} from '#/view/com/util/UserBanner'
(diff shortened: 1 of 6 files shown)


kindgracekind:
You’d probably want a separate record for this instead of a new field, see: https://docs.bsky.app/blog/pinned-posts#current-recommendations

surfdude29:
I'm no branding expert, but what about using **Showcase** for the profile tab? This might be more easily understood and easier to market as a Bluesky USP ("You can easily showcase your best content to visitors to your Bluesky profile with a prominently-displayed custom feed that you control.")

Ex-Twitter has a **Highlights** tab only for paid users, but that's of course much more limited because it's only for manually selected posts as they don't have custom feeds.

https://techcrunch.com/2023/08/15/paid-users-of-x-formerly-twitter-gets-a-new-tab-to-highlight-their-posts/

estrattonbailey:
Hey! Thanks for this, dig the idea. We'll def consider this, would love to provide more customization for profiles! Gonna move to draft and label this so we can more easily keep track of it 🙏

DGaffney:
:sickosyes: thank you @estrattonbailey - obviously feel free to throw this version out / rewrite it whatever this is mostly just a claude'd placeholder to be as productive as possible in negging you to build this thing <3

DGaffney:
@estrattonbailey its been a bit - is this anywhere on the roadmap?

DGaffney:
thread


bump

PR #11027, opened 2026-06-30
Added Automation label to labeller display names

RetroSunstar:
Currently labellers declared with the 'bot' label do not show the automation icon beside their display names like regular profiles would

This creates an inconsistency between the two profile types and other elements across the app, such as profile hover cards or display names on posts which _do_ show this badge for labellers

But while looking into this, another inconsistency popped up between ``ProfileHeaderStandard`` and ``ProfileHeaderLabeler`` in how they handle these displays. Labellers will use ``DisplayName.tsx`` while Standard will redefine the whole block itself, as this didn't seem necessary it makes more sense to update it now than contributing to this duplication 🐙


estrattonbailey:
@RetroSunstar looks good, thanks. Do you have a labeler on hand I can test with?

RetroSunstar:
> @RetroSunstar looks good, thanks. Do you have a labeler on hand I can test with?

[firewatcher.bsky.social](https://bsky.app/profile/firewatcher.bsky.social) is a blank subject with the bot label, if that works? 🐙

estrattonbailey (approved):

Issue #5913, opened 2024-10-23
Post character limits


dkletter:
### Describe the Feature

Please consider increasing the character limit from 300 to 500 and make all links equal 12 chars, no matter how long the URL is.

We aren’t living in an SMS-restricted era anymore. It can be difficult to write a thoughtful critique of books, art, music, live performances, etc without resorting to clumsy abbreviations and whatnot. When I’m writing about an upcoming album, for example, I like to include a Bandcamp link or similar. I know there’s an option to delete the link after the preview is generated, but it feels clunky and I’m worried about how future proof it is, y’know?

jabberwockxeno:
I would also like to express my support for this, though I'd prefer 2000 (with everything beyond 300 hidden behind a "show more") over merely 500, which is still not enough to really say anything of real nuance or depth

Erics1337:
Please increase the character limit! Bluesky is difficult to use for those coming from Instagram due to this arbitrary restriction. Twitter originally had a short limit because it operated through SMS, but that shouldn't apply to Bluesky. Let users choose whether they want to read longer posts. There’s a reason Instagram allows for longer posts—they're necessary! Dont make thisa blocker for people wishing for an alternative platform from Meta.

JaredTweed:
Please increase the character limit substantially! I love nuance and the character limit eliminates that from all Bluesky discussion.

Issue #4952, opened 2024-08-17
Invalid app.bsky.feed.post record: Recordtext must not be longer than 3000 characters


enn-nafnlaus:
**Describe the bug**
Received the above error when trying to post in the webapp.

**To Reproduce**
Try to post this:

![image](https://github.com/user-attachments/assets/b8d00584-374c-4cf8-a30d-bb1747545b21)

**Expected behavior**

It either posts without error, or it doesn't say that I'm under the character limit (note that it says I have 59 characters left)

**Screenshots**

See above.

**Details**

 - Platform: Chrome for Linux
 - Platform version:
 - App version:

**Additional context**

None

PR #11031, opened 2026-06-30
Raise composer character limit

### LINKED ISSUE #10960: Constrained 300-character limit causes high user friction
### Steps to Reproduce

Open the Bluesky composer to draft a post.

Attempt to communicate a nuanced thought, detailed technical explanation, or contextualized statement.

Observe the hard 300-character limit.

Users are forced to either heavily edit out important context, string together clunky threads, or abandon the post entirely.

### Attachments

_No response_

iOS, Web (Desktop), Web (Mobile), Android

### Device Info

_No response_

Latest

### Additional Information

The current 300-character hard limit creates an unnecessary conflict between AtProto's potential as a universal social layer and Bluesky's specific design choices as a microblogging client. I am proposing a structural shift: decouple the protocol's capacity from the client's visual presentation by raising the protocol limit to 2000 characters.

While the microblogging modality is highly effective for reading, enforcing it strictly at the protocol level introduces severe UX friction during creation:

- Cognitive Overhead (The Editing Tax): Users consistently abandon high-value, nuanced content because adapting a complete thought into a strict 300-character box—or artificially fracturing it into a thread—requires too much editing effort.

- Accessibility & Context: Arbitrary short limits structurally disadvantage users, particularly neurodivergent individuals, who rely on precise phrasing and detailed context to communicate clearly and avoid misunderstanding.

- The Federation Bottleneck: Constraining AtProto itself to 300 characters forces independent developers and federated platforms to inherit Bluesky's exact UI paradigm. If the protocol is to thrive as a foundational layer, it must be flexible enough to support varied content lengths.

The Solution: The "Show More" Compromise
We can fully preserve the fast-paced, scannable timeline without throttling user expression. The protocol should accept 2000-character records, while the official Bluesky client handles the constraint visually—automatically truncating text past 300 characters behind a native "Show More" button.

By making this change, Bluesky can eliminate friction at the point of creation while perfectly maintaining its crisp, microblogging aesthetic at the point of consumption.

ThirdToneProject:
## Summary

- raise the composer post grapheme limit from 300 to 2000
- raise the draft-save grapheme limit to match, so longer posts can still be saved as drafts
- update draft-limit prompts to render from the shared draft limit constant

## Research / caveat

This is the social-app side of the change. The current `app.bsky.feed.post` lexicon in `bluesky-social/atproto` still declares `text.maxGraphemes: 300` and `text.maxLength: 3000`, so this client patch depends on a protocol/server validation change before it can be merge-ready.

Relevant issue threads:

- Closes #10960
- Refs #5913
- Refs #4952

## Test plan

- `git diff --check`
- Not run locally: `pnpm lint` and `pnpm typecheck` both require installed dependencies, and this checkout has no `node_modules`. The local runtime also reports Node `v20.19.4`, while the repo requires `>=24.15.0`.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -61,9 +61,9 @@ export function FEEDBACK_FORM_URL({
 export const MAX_DISPLAY_NAME = 64
 export const MAX_DESCRIPTION = 256
 
-export const MAX_GRAPHEME_LENGTH = 300
+export const MAX_GRAPHEME_LENGTH = 2000
 
-export const MAX_DRAFT_GRAPHEME_LENGTH = 1000
+export const MAX_DRAFT_GRAPHEME_LENGTH = 2000
 
 export const MAX_DM_GRAPHEME_LENGTH = 1000
 
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1375,7 +1375,14 @@ export const ComposePost = ({
                     </Trans>
                   )
                 ) : (
-                  <Trans>You can only save drafts up to 1000 characters.</Trans>
+                  <Trans>
+                    You can only save drafts up to{' '}
+                    {plural(MAX_DRAFT_GRAPHEME_LENGTH, {
+                      one: '# character',
+                      other: '# characters',
+                    })}
+                    .
+                  </Trans>
                 )}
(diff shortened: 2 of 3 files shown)


ThirdToneProject:
Closing because this PR was opened from the wrong GitHub account. I will reopen from the correct account.

PR #11032, opened 2026-06-30
Raise composer character limit

### LINKED ISSUE #10960: Constrained 300-character limit causes high user friction

Erics1337:
## Summary

- raise the composer post grapheme limit from 300 to 2000
- raise the draft-save grapheme limit to match, so longer posts can still be saved as drafts
- update draft-limit prompts to render from the shared draft limit constant

## Research / caveat

This is the social-app side of the change. The current `app.bsky.feed.post` lexicon in `bluesky-social/atproto` still declares `text.maxGraphemes: 300` and `text.maxLength: 3000`, so this client patch depends on a protocol/server validation change before it can be merge-ready.

Upstream protocol discussion:

- https://github.com/bluesky-social/atproto/discussions/5171

Relevant issue threads:

- Closes #10960
- Refs #5913
- Refs #4952

## Test plan

- `git diff --check`
- Not run locally: `pnpm lint` and `pnpm typecheck` both require installed dependencies, and this checkout has no `node_modules`. The local runtime also reports Node `v20.19.4`, while the repo requires `>=24.15.0`.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -61,9 +61,9 @@ export function FEEDBACK_FORM_URL({
 export const MAX_DISPLAY_NAME = 64
 export const MAX_DESCRIPTION = 256
 
-export const MAX_GRAPHEME_LENGTH = 300
+export const MAX_GRAPHEME_LENGTH = 2000
 
-export const MAX_DRAFT_GRAPHEME_LENGTH = 1000
+export const MAX_DRAFT_GRAPHEME_LENGTH = 2000
 
 export const MAX_DM_GRAPHEME_LENGTH = 1000
 
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -1375,7 +1375,14 @@ export const ComposePost = ({
                     </Trans>
                   )
                 ) : (
-                  <Trans>You can only save drafts up to 1000 characters.</Trans>
+                  <Trans>
+                    You can only save drafts up to{' '}
+                    {plural(MAX_DRAFT_GRAPHEME_LENGTH, {
+                      one: '# character',
+                      other: '# characters',
+                    })}
+                    .
+                  </Trans>
                 )}
               </Prompt.DescriptionText>
             </Prompt.Content>
--- a/src/view/com/composer/drafts/DraftsButton.tsx
+++ b/src/view/com/composer/drafts/DraftsButton.tsx
@@ -1,4 +1,4 @@
-import {msg} from '@lingui/core/macro'
+import {msg, plural} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react'
 import {Trans} from '@lingui/react/macro'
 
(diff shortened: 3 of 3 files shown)


estrattonbailey:
@Erics1337 hey appreciate you working on this ❤️ but we aren't currently interested in raising the character limit on posts. It would introduce a whole new paradigm that we're not ready to support on frontend, backend, or on the moderation side of things.

There are lots of great apps in the atmosphere that support long-form content, like Leaflet, PCKT, Offprint, Lemma, etc.

Erics1337:
@estrattonbailey thanks for taking a look, and I appreciate the direct answer.

Would you be willing to expand a bit on what parts of this make it a new paradigm that Bluesky is not ready to support? I opened a protocol-level discussion here with more context and research: https://github.com/bluesky-social/atproto/discussions/5171

The distinction I am trying to make is between protocol capacity and client presentation. I agree that the main Bluesky timeline should stay fast and scannable, and I am not proposing that 2000-character posts render inline everywhere. The proposal is closer to: allow a larger `app.bsky.feed.post` ceiling, while clients preserve the current feel through truncation / “Show more” / compact rendering on feeds, quotes, notifications, embeds, etc.

I would really like to understand the specific blockers on frontend, backend, and moderation:

- Is the concern mostly product identity / feed readability?
- Are there known AppView, PDS, indexing, relay, or storage costs that make 2000 meaningfully risky?
- Is moderation tooling currently built around the assumption that posts are short?
- Would a staged or flag-gated approach change the calculus?

I am pushing on this because the 300-character limit is driving away people who otherwise want to use Bluesky, especially people trying to write with context, accessibility, technical detail, or nuance. That feels counterproductive to Bluesky’s broader mission as an open social layer: the protocol-level limit forces all `app.bsky` clients into one very specific microblogging composition model, even if individual clients could choose stricter local limits or collapsed presentation.

If the answer is “not now,” I understand. But it would be helpful to document what would need to change before this could become viable.

PR #11035, opened 2026-06-30
Log additional context for post:* events

ds-boyce:
Within a thread that was entered from a feed, only the anchor post inherited the source. So quoting any other post in the thread yielded a missing `feedDescriptor`. This resulted in a large number of `post*` events with an undefined `feedDescriptor`.

### DIFF
--- a/src/screens/PostThread/index.tsx
+++ b/src/screens/PostThread/index.tsx
@@ -14,7 +14,11 @@ import {useInitialNumToRender} from '#/lib/hooks/useInitialNumToRender'
 import {useNonReactiveCallback} from '#/lib/hooks/useNonReactiveCallback'
 import {useOpenComposer} from '#/lib/hooks/useOpenComposer'
 import {usePostViewTracking} from '#/lib/hooks/usePostViewTracking'
-import {useFeedFeedback} from '#/state/feed-feedback'
+import {
+  FeedFeedbackProvider,
+  type StateContext as FeedFeedbackStateContext,
+  useFeedFeedback,
+} from '#/state/feed-feedback'
 import {type ThreadViewOption} from '#/state/queries/preferences/useThreadPreferences'
 import {
   PostThreadContextProvider,
@@ -68,6 +72,24 @@ export function PostThread({uri}: {uri: string}) {
     anchorPostSource?.feedSourceInfo,
     hasSession,
   )
+
+  /*
+   * Non-anchor posts (parents and replies) are not the post the user tapped
+   * from a feed, so they should not report feed interactions. But they should
+   * still carry the originating feedDescriptor for analytics, so events like
+   * post:clickQuotePost can be attributed to the feed the thread was opened
+   * from. This inert context value provides feedDescriptor while keeping
+   * interaction reporting disabled (enabled: false makes sendInteraction and
+   * onItemSeen no-ops). The anchor renders its own full provider that nests
+   * inside and overrides this for its subtree.
+   */
+  const analyticsOnlyFeedFeedback: FeedFeedbackStateContext = {
+    enabled: false,
+    onItemSeen: () => {},
+    sendInteraction: () => {},
+    feedDescriptor: feedFeedback.feedDescriptor,
+    feedSourceInfo: undefined,
+  }
 
   /*
    * One query to rule them all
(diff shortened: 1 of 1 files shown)


src/screens/PostThread/index.tsx
```diff
@@ -640,6 +650,36 @@ export function PostThread({uri}: {uri: string}) {
...
+  /*
+   * Non-anchor posts (parents and replies) are not the post the user tapped
+   * from a feed, so they should not report feed interactions. But they should
+   * still carry the originating feedDescriptor for analytics, so events like
+   * post:clickQuotePost can be attributed to the feed the thread was opened
+   * from. This inert context value provides feedDescriptor while keeping
+   * interaction reporting disabled (enabled: false makes sendInteraction and
+   * onItemSeen no-ops). The anchor renders its own full provider that nests
+   * inside and overrides this for its subtree.
+   */
```
estrattonbailey, line 668: Are we sure we don't want to send interactions for other posts in the thread? This fixes the root of the ticket, but we'll be dropping all the `sendInteraction` calls. Idk if that's a problem, but it is a change that this PR is making, iiuc.

ds-boyce: Non-anchor posts never sent them in the first place (no `FeedFeedbackProvider`); the only post that actually sent interactions was the [anchor](https://github.com/bluesky-social/social-app/blob/main/src/screens/PostThread/components/ThreadItemAnchor.tsx#L532). For the rest it’s an explicit no-op now.

estrattonbailey: Ok right right. Should we send likes and such for replies too? A "like" or a "quote post click" seem similar. All good if not, up to algo team I suppose

estrattonbailey (approved):
SGTM if algo team is +1!

src/screens/PostThread/index.tsx
ds-boyce: @ocseckin What do you think?

ocseckin: Checking if I understand correctly:

- An anchor is the post the user clicks on, from any feed surface. E.g., I'm on Discover. I see post X, which quotes post Y. I click post X, and it opens thread T. Here, post X is the anchor and post Y is the parent.
The other posts in thread T (everything except post X) don't carry a `feedDescriptor`. So if I quote post Y, that quote won't have a `feedDescriptor` either.
- This is likely why so many quotes on [this post](https://bsky.app/profile/brendelbored.bsky.social/post/3moegbsvyzk2g) had a null `feedDescriptor`. Because most people see a post like X that quotes Y and they quoted Y after entering thread Y.
- With this PR, `post:*` events will inherit the anchor's `feedDescriptor`. So we won't miss these cases anymore.

One thing I'm unclear on: what does it mean to "send likes and such for replies"? A reply is already a `post:*` event (e.g. `post:Reply`), right? If so, wouldn't this PR already cover it? Or do you mean something different: for example, a "like" event on thread T isn't a `post:*` event (and is something like `interaction:Like`), so it wouldn't inherit the anchor's `feedDescriptor` unless we add that separately?

ds-boyce: Think of it as two pipelines:

* Analytics pipeline: Every `post:*` event fired on a non-anchor post now inherits the anchor's `feedDescriptor` instead of `undefined`. That’s the change here, which covers everything because it all goes through `ax.metric` with the same context.
* Feed feedback pipeline: Stays a no-op (`enabled: false`), so we deliberately do not report `interactionLike` etc. to the feed generator for non-anchor posts. Those posts weren't actually served by the feed.

So “send likes for replies” means analytics attribution: Any like/repost/quote on a reply in the thread now carries the anchor's `feedDescriptor`.

It doesn’t mean feed-generator interaction feedback: This PR intentionally doesn’t do that (it keeps `enabled: false`). Turning that on would be a separate change we probably don’t want, since those replies weren’t served by the feed.

ocseckin: I see, there are two pipelines and the feed feedback side will stay a no-op in this PR as of now.

One more question: Let's say I see the post X that quotes post Y on thread T. If I click directly on post Y (not on post X) and quote post Y, would it inherit the anchor's `feedDescriptor` T now?

ds-boyce: Something like this, where you’re Z?

```
[T]
  [A]
  [B]
  ...
  [Y] => [X[Y]] => [Z[Y]]
```

In this case, no, your (Z’s) post would not inherit T’s `feedDescriptor`, it would be `undefined`.

ocseckin: Ok, all sound good to me

PR #11042, opened 2026-07-01
Fix blank message list after submit on Android

ds-boyce:
We were passing `Number.MAX_SAFE_INTEGER` to oversaturate the scroll, but `View.scrollTo` takes a 32-bit `int`. It worked on iOS bc `UIScrollView.setContentOffset` takes a `CGFloat`.

This PR switches to `1e7` which is more than large enough and fits for both platforms. It also reformats the multi-line comments per convention.

### DIFF
--- a/src/screens/Messages/components/MessagesList.tsx
+++ b/src/screens/Messages/components/MessagesList.tsx
@@ -86,6 +86,8 @@ import {MessagesListGroupInfoPanel} from './MessagesListGroupInfoPanel'
 import {MessagesListInfoPanel} from './MessagesListInfoPanel'
 import {KeyboardStickyView} from './vendor/KeyboardStickyView'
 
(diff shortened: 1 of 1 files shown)


src/screens/Messages/components/MessagesList.tsx
```diff
@@ -277,29 +297,42 @@ export function MessagesList({
...
+   * fits comfortably in int32. (APP-2223)
+   */
+  const SATURATING_SCROLL_OFFSET = 1e7
   const scrollSendToBottom = useCallback(() => {
     flatListRef.current?.scrollToOffset({
-      offset: Number.MAX_SAFE_INTEGER,
+      offset: SATURATING_SCROLL_OFFSET,
       animated: true,
     })
   }, [flatListRef])
```
ds-boyce: The relevant change.

mozzius (approved):

PR #11043, opened 2026-07-01
Special-case 'me' values in search queries

ds-boyce:
For parity with search v1.

### DIFF
--- a/src/lib/strings/helpers.ts
+++ b/src/lib/strings/helpers.ts
@@ -47,30 +47,12 @@ export function countLines(str: string | undefined): number {
   return str.match(/\n/g)?.length ?? 0
 }
 
-// Augments search query with additional syntax like `from:me`
-export function augmentSearchQuery(query: string, {did}: {did?: string}) {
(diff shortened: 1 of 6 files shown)


mozzius (approved):

src/lib/strings/helpers.ts
```diff
@@ -47,30 +47,12 @@ export function countLines(str: string | undefined): number {
   return str.match(/\n/g)?.length ?? 0
 }
 
-// Augments search query with additional syntax like `from:me`
-export function augmentSearchQuery(query: string, {did}: {did?: string}) {
```
estrattonbailey, line 51: Does this affect v1 at all? Does v1 resolve `from:me` on the backend too?

estrattonbailey (approved):
Nice nice

src/lib/strings/helpers.ts
ds-boyce: Yep v1 resolves `from:me` as well.

PR #11044, opened 2026-07-02
add version bump workflow triggered by app store release webhook

mozzius:
## What

Adds `.github/workflows/version-bump-on-release.yml`, which listens for a `repository_dispatch` event (`asc-app-released`) and opens a PR bumping the minor version in `package.json` (e.g. 1.127.0 → 1.128.0).

## Why

App Store Connect now supports webhooks for app version state changes. A small Cloudflare Worker relay receives those webhooks, posts review-status updates to Slack, and — when a release reaches `READY_FOR_DISTRIBUTION` (live on the App Store) — fires the dispatch event that triggers this workflow. This automates the manual "bump package.json after release" step of the iOS release flow.

## Details

- Dedupes duplicate webhook deliveries by checking for an existing open PR on the target branch (`bot/version-bump-<next-version>`)
- Version write-back uses `jq --indent 2` with a `git diff` guard that aborts if anything other than the version line changed
- Least-privilege permissions (`contents: write`, `pull-requests: write`), zizmor-clean, no `${{ }}` interpolation in run blocks
- Also has `workflow_dispatch` for manual testing
- Requires the repo/org setting "Allow GitHub Actions to create and approve pull requests" to be enabled for `gh pr create` with the default `GITHUB_TOKEN` to work

### DIFF
--- a/.github/workflows/version-bump-on-release.yml
+++ b/.github/workflows/version-bump-on-release.yml
@@ -9,6 +9,13 @@ on:
   # Allow manual runs for testing the bump/PR flow without a real release.
   workflow_dispatch:
 
+# Serialize runs: Apple delivers the release webhook at least once, so two
+# near-simultaneous deliveries could otherwise race the dedup check against
+# branch creation. A single concurrency group forces them to run one at a time.
+concurrency:
+  group: version-bump-on-release
+  cancel-in-progress: false
+
 # Least privilege: we only need to push a branch (contents) and open a PR.
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


mozzius:
Thanks for the review. Addressed in 390456b:

- **Fix 1** (race): added a workflow-level `concurrency` group (`version-bump-on-release`, `cancel-in-progress: false`) so at-least-once webhook deliveries serialize instead of racing the dedup check against branch creation.
- **Fix 2** (diff guard): replaced the `grep -c '"version"'` count with a `git diff --numstat` check asserting exactly `1+1` (one line added, one removed), which catches jq reformatting anywhere in the file and isn't fooled by other `"version"` keys.

On suggestion 3 (gate on a payload version): intentionally skipped. Apple's webhook payload doesn't include the version string, only an `appStoreVersions` resource ID, so the relay's `client_payload` can't carry it without an extra ASC API callback from the Worker. The re-delivery-after-merge window is small and now further shrunk by the concurrency group; worst case is a spurious PR that gets closed manually.

mozzius:
CI failure fixed in https://github.com/bluesky-social/social-app/pull/11046

PR #11045, opened 2026-07-02
fix: auto-follow @coseeker.com during onboarding instead of Bluesky

### LINKED ISSUE #61: Failed image upload runs into cache issue on reupload attempt
There's some state that's created during the process of publishing a post with images attached. If the post fails to actually upload, that state conflicts on subsequent attempts to upload.

To reproduce, modify the publishing flow to throw an error right before the post record is created, then try to reupload. You will see an error about a file already existing.

achalagrawal:
## Summary
- Onboarding auto-follows a hardcoded account via `BSKY_APP_ACCOUNT_DID` (`src/screens/Onboarding/StepFinished/index.tsx`), sourced from each brand's `appAccountDid`.
- `coseeker`, `k4m2a`, and `mdparivaar` brand configs still had the real Bluesky team DID (`did:plc:z72i7hdynmk6r22z27h6tvur`), copy-pasted from upstream.
- Updated all three to `@coseeker.com`'s DID (`did:plc:ieyfjh6ystyufa3a7pi3jw5q`) so new users follow the right account on signup.
- `brands/bluesky/brand.ts` is intentionally left unchanged — its docblock states it's meant to be an exact clone of upstream Bluesky behavior (used as the fallback when no brand override is set).

## Test plan
- [x] `pnpm typecheck` passes
- [ ] Manually verify by completing onboarding on a coseeker build and confirming the new account follows @coseeker.com

Fixes #61


achalagrawal:
Opened against the wrong repo by mistake — closing. Correct PR is against k4m2a/social-app.

PR #11046, opened 2026-07-02
Fix zizmor findings in build workflows

mozzius:
Fixes the remaining `zizmor .github/workflows/* --min-confidence=medium` findings. It now reports zero.

**Fix action pin (bundle-deploy-eas-update.yml, pull-request-commit.yml)** — the pinned `bluesky-social/github-actions/fingerprint-native` SHA didn't correspond to any published tag. Published a `v0.2.0` tag for it in the github-actions repo and updated the pin comments to match.

**`secrets-inherit` (nightly-build.yml × 2)** — the nightly `ios`/`android` jobs passed `secrets: inherit` to the reusable build workflows, handing them the entire repo secret store. Both reusable workflows now declare exactly the secrets they use in their `workflow_call` trigger (all `required: true`), and `nightly-build.yml` forwards them explicitly. `workflow_dispatch` runs are unaffected — they read repo secrets directly.

**`superfluous-actions` (build-submit-android.yml)** — replaced `softprops/action-gh-release` with a plain `gh release upload` script step. The tag/filename are passed via `env:` (no template interpolation in the script body), `--clobber` makes re-runs idempotent, and the step still emits `url` (via `gh release view`) so the downstream Slack notification is unchanged.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -78,7 +78,20 @@ jobs:
     with:
       profile: testflight
       testFlightGroup: "QA Team"
-    secrets: inherit
+    # Pass only the secrets the reusable workflow declares, rather than `secrets: inherit`,
+    # so the nightly build never hands the reusable workflow the entire repo secret store.
+    secrets:
+      EXPO_TOKEN: ${{ secrets.EXPO_TOKEN }}
+      ENV_TOKEN: ${{ secrets.ENV_TOKEN }}
+      SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+      BITDRIFT_API_KEY: ${{ secrets.BITDRIFT_API_KEY }}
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly-build.yml
```diff
@@ -91,7 +104,21 @@ jobs:
     uses: ./.github/workflows/build-submit-android.yml
     with:
       profile: testflight-android
-    secrets: inherit
+    # Pass only the secrets the reusable workflow declares, rather than `secrets: inherit`,
+    # so the nightly build never hands the reusable workflow the entire repo secret store.
+    secrets:
```
estrattonbailey, line 109: The pre-existing workflows already have full access to secrets when run via workflow dispatch, so this is basically a no-op, right? We're just tightening access for an action that has always had access to all secrets. This could be a little annoying if we need add a new secret to the build workflows, so not a huge deal since we rarely need to update these workflows.

estrattonbailey (approved):

PR #11051, opened 2026-07-02
Use Sift for username suggestions during signup

ds-boyce:
https://github.com/user-attachments/assets/a9a609ff-8e94-41aa-b694-b4b02ec11fe5

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -1492,11 +1492,6 @@
       "count": 1
     }
   },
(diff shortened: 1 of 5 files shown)


ds-boyce:
> This is fine **as long as @bsky.app/sift's useSift is native-safe**

Confirmed signup works as before on native.

mozzius (approved):
Good stuff

PR #11054, opened 2026-07-02
Remove success toast from Bookmark Button

RetroSunstar:
Currently when bookmarking a post, a toast appears informing the user that the post has been bookmarked, alongside an undo button

This is entirely unnecessary, as the icon for bookmarks that the user must press in order to do so already updates its appearance to signify a post has been bookmarked. Undoing a bookmark can be done by simply reinteracting with the bookmark button, and a dedicated undo button is not required here

This is also inconsistent when compared to other post interation buttons, for example likes don't show a toast at all

The best move would be to remove both toasts for saving and removing a bookmark, BUT the toast used for removals does serve one purpose: The undo button is insanely helpful for accidential removals of bookmarks in the 'Saves' tab because this menu removes entires instantly. Without an undo button for this action, this would become infuriating from a user perspective 🐙


estrattonbailey (approved):
Yep seems reasonable, thanks!

PR #11059, opened 2026-07-03
Fix support link cutoff in signup footer on mobile

surfdude29:
> [!NOTE]
> **Claude Code analyzed this issue and wrote the PR and the description.**
>
> **I tested this on web with the window shrunk and the zoom increased, but not on Android or iOS.**

## Context

In the Create account flow, the footer row holds the app-language selector on the left and a "Having trouble? Contact support" text on the right. On native and mobile web, for locales with long strings — or in English at larger text sizes — the support text overflows off the right edge of the screen and gets clipped.

| German – iOS – 100% | Spanish – iOS – 100% | 
|--------|--------|
| IMG_7038 | IMG_7039 | 

| Aragonese – iOS – larger text | English – iOS – larger text |
|--------|--------|
| IMG_7024 | IMG_7032 |

**Root cause** (`src/screens/Signup/index.tsx:217-247`): the footer is a `flex_row`, and on mobile (`!gtMobile`) the `View` wrapping the support text has **no flex styles at all** (`gtMobile ? [a.flex_1, ...] : []`). In Yoga, `flexShrink` defaults to `0`, so the text view renders at its intrinsic single-line width and simply overflows the row — the text never wraps because it never receives a width constraint.

**Desired behavior**: if the string can't fit beside the language selector, the whole text block wraps onto its own line underneath the selector.

## Change

Single file: `src/screens/Signup/index.tsx`

1. **Add `a.flex_wrap` to the footer row container** (the `View` at line 217 with `a.w_full, a.py_lg, a.flex_row, a.gap_md, a.align_center`).
   - Yoga places flex items on lines using their intrinsic (flex-basis) size before shrinking, so when the text block doesn't fit next to the dropdown, it wraps to a second line below it.
   - The existing `a.gap_md` applies to both axes in Yoga, so wrapped lines get sensible vertical spacing for free.
   - On `gtMobile` this is a no-op: the text wrapper there uses `a.flex_1` (flex-basis 0), which always fits on the first line, so desktop layout is unchanged.

2. **Give the mobile text wrapper `a.flex_shrink`** — change the ternary at line 226-231 from `gtMobile ? [a.flex_1, a.flex, a.flex_row, a.justify_end] : []` to use `[a.flex_shrink]` for the mobile branch.
   - This handles the extreme case where the string is longer than the *full* row width even on its own line (very long locales and/or large font scale): shrinking constrains the view's width, letting the `Text` wrap internally onto multiple lines instead of clipping.
   - Shrink only kicks in within a line after wrapping, so it doesn't prevent the wrap-below-the-selector behavior from step 1.

No other screens share this pattern — the language-dropdown + support-link footer exists only in the Signup screen (`AppLanguageDropdown` is used elsewhere, but not in this row layout).


surfdude29:
@mozzius Conflict has been resolved

surfdude29:
@estrattonbailey @mozzius Would it be possible to take a look at this please? The bottom of several screens of the account creation flow currently looks a bit broken in some languages on iOS (I assume on Android too) and on mobile web, and I tested this on web with the window shrunk and the zoom increased and it does seem to fix the issue and allow the text to wrap:

Screenshot 2026-07-16 at 13 53 34

mozzius (approved):

surfdude29:
@mozzius Anything else I need to do here? :)

mozzius:
Heya. I got massively sidetracked while reviewing this and ended up doing a massive bug hunt for a reanimated bug lol. PR looks good!

surfdude29:
Ah fair enough! Nice, thanks for taking a look 🙏

surfdude29:
@mozzius Maybe this could make it into 1.132? 🤞

mozzius:
Omg sorry haha

surfdude29:
No worries, thanks! :)

PR #11065, opened 2026-07-05
Fix back swipe opening the drawer after a cold-start deep link

mozzius:
On a cold start from a deep link, getStateFromPath returns a partial
navigation state with no `index`. React Navigation rehydrates it
correctly (the deep-linked screen sits on top of Home), but child
navigators don't write the rehydrated state back to the container on
the initial render, so the state observed via useNavigationState keeps
the index-less child state until the first navigation action.

getCurrentRoute() only descended into nested state while `index` was a
number, so it stopped at the tab route and isStateAtTabRoot() reported
true. DrawerLayout then kept the drawer swipe gesture enabled on top of
the deep-linked screen, and a back swipe opened the drawer instead of
popping back to Home.

Fix by mirroring React Navigation's rehydration in getCurrentRoute()
(a missing `index` focuses the last route) and by setting `index`
explicitly in buildStateObject(). Also reuse getCurrentRoute() in
getPathFromState(), which had the same descent logic inlined.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_017jQFxubQ8JepHTPRrjC452

### DIFF
--- a/__tests__/lib/routes-helpers.test.ts
+++ b/__tests__/lib/routes-helpers.test.ts
@@ -1,0 +1,107 @@
+import {
+  buildStateObject,
+  getCurrentRoute,
+  isStateAtTabRoot,
+} from '#/lib/routes/helpers'
+import {type State} from '#/lib/routes/types'
+
+describe('getCurrentRoute', () => {
+  it('returns Home when there is no state', () => {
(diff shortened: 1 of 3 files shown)


mozzius:
Confirmed working on device

PR #11066, opened 2026-07-05
Fix invisible bidi characters copied with handles on web

### LINKED ISSUE #8451: web: copying and pasting handles from profile pages includes invisible Unicode bidi characters
### Steps to Reproduce

Hi all! When you select and copy someone's handle from their bsky.app profile page, you also get invisible Unicode bidi characters, specifically `\u202a` at the beginning and `\u202c` at the end. These characters often trip up other Bluesky apps and tools when you paste that handle into them. Any chance there's a different way to do bidi handling that wouldn't cause this problem?

### Attachments

_No response_

Web (Desktop)

### Device Info

Multiple

latest

### Additional Information

_No response_

a-y-ibrahim:
## Summary
Selecting and copying a handle (or display name) from a profile, feed, or notification on web also copies invisible Unicode bidi characters (U+202A and U+202C). They break handle lookups and interop when pasted into other Bluesky/ATProto apps and tools. Closes #8451.

## Cause
`forceLTR()` in `src/lib/strings/bidi.ts` wraps the string in U+202A ... U+202C to keep it left-to-right. On web those control characters are part of the rendered DOM text, so they end up in the copied text. `sanitizeHandle()` applies `forceLTR` by default, and a few display-name render sites call it directly.

## Change
On web, isolate the direction with CSS instead of injecting characters:
- `forceLTR` returns the string unchanged on web (native keeps the manual Unicode wrapping, which has no CSS equivalent there).
- The handle/name render sites get `direction: 'ltr'` + `unicode-bidi: 'isolate'`.

This mirrors the approach already used for the profile header handle in `src/screens/Profile/Header/Handle.tsx`. Handles are ASCII so they stay visually LTR, and display names that may be RTL keep their isolation via the CSS, while the copied text stays clean.

Updated sites: `ProfileCard`, `PostMeta`, `NotificationFeedItem`. The central `forceLTR` change also covers every `sanitizeHandle` call site on web.

## Notes
This follows the existing inline CSS pattern in `Handle.tsx`. Happy to extract a shared `<Bdi>` component instead, as suggested in the thread, if you prefer.

## Testing
Ran locally against the repo toolchain: `typecheck` (tsgo), `lint` (eslint), `prettier --check`, and the full `jest` suite all pass.


a-y-ibrahim:
Addressed both review comments in 4c64eb0:

- `jest` was not imported from `@jest/globals` even though `jest.mock(...)` is used: added it to the import list.
- The directional-formatting constants used literal invisible characters (U+202A, U+202C) instead of escapes: switched both to numeric escapes, matching the style already used in `bidi.ts`.

Local jest run is green after the change.

estrattonbailey (approved):
Looks like a good solution, thank you

PR #11068, opened 2026-07-06
Clean up feature gate for follow buttons in expanded notifications

abenzer:
Ships the new expanded notification bundle layout to everyone and removes the `notifications:expanded_profile_card:enable` feature gate, making it the only code path.

## Testing

Use an account with bundled notifications (e.g. a post with multiple likes or several new followers), then:

1. Go to the Notifications tab and tap/click the chevron on a bundled notification to expand it
2. Confirm each author row shows an avatar, name/handle, and a follow button
3. Tap Follow on one author - it should flip to "Following", and back if tapped again
4. Collapse and re-expand the bundle to check the animation (fade on native, height expand on web)
5. Web only: hover over the expanded authors list - the row's background hover highlight should turn off while the cursor is over the list
6. Tapping an author row should still open their profile

Check on both web and native, since the expand animation differs per platform.

### DIFF
--- a/src/analytics/features/types.ts
+++ b/src/analytics/features/types.ts
@@ -12,7 +12,6 @@ export enum Features {
   GroupChatsDisable = 'group_chats:disable',
   ComposerLanguageDetectionEnable = 'composer:language_detection:enable',
   PostGalleryEmbedEnable = 'post_gallery_embed:enable',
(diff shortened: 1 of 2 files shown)


vineyardbovines (approved):
noice @claude quick sweep to triple-check please

ds-boyce (approved):

PR #11080, opened 2026-07-07
Prototype: Show known likers on post thread page

abenzer:
> **Note: this is a prototype.** It's gated behind `post_thread:known_likers:enable` while we evaluate the UX. Not intended for a general rollout in its current form.

## What this does

On the post thread page, the post's plain "N likes" stat is replaced with a row showing which people you follow liked the post: a few stacked avatars plus a sentence like **"Jack, Amy, and 34 others like this."** Names link to profiles, but the rest of the row links to the liked-by list. Viewers who don't follow any of the likers (or aren't signed in) see the existing "N likes" text unchanged.

## Demo
Screenshot 2026-07-06 at 6 16 00 PM

## How it works

- Fetches a single page of `app.bsky.feed.getLikes` (`limit: 100`, the API max) via a new `useLikedBySampleQuery`, kept separate from the liked-by screen's paginated query.
- Filters to likers the viewer follows, excluding self, muted, blocked, and blocking accounts. Display names and avatars go through the standard moderation pipeline.
- Gated behind `post_thread:known_likers:enable`. The gate is evaluated only for signed-in viewers of posts with likes, and when it's off, no `getLikes` request fires.

## Known limitations

- Known likers are sampled from the 100 most recent likes, so a followed liker who liked a popular post early may not surface. Going deeper than 100 would need server-side support.
- The row swaps in when the sample loads, which shifts layout slightly.
- Viewing a post and then opening its likes list fetches likers twice (separate caches).


estrattonbailey:
Let's discuss in Slack. Just concerned about the feasibility of actually shipping this to production for all posts. Huge +1 on the feature though 👍

estrattonbailey (approved):

PR #11083, opened 2026-07-07
Update SVGO

mozzius:


estrattonbailey:
Do we even need this? Forgot it was here tbh. The hardest part of exporting icons is collapsing everything into a single path. The minification is helpful, but using https://jakearchibald.github.io/svgomg/ is actually good for previewing too.

mozzius:
I prefer it being in the repo because it standardizes the changes it applies. svgomg just uses svgo under the hood, so you can just dump svgs straight from figma into the assets folder and it will process it

estrattonbailey (approved):

mozzius:
@claude can you resolve conflicts

PR #11085, opened 2026-07-07
Update KWS supported languages and mapping

surfdude29:
I noticed that `KWS_SUPPORTED_LANGS` doesn't currently include all of the languages that KWS supports according to [their docs](https://dev.epicgames.com/docs/kids-web-services/kws-localization#languages-supported-by-kws-apis).

This PR proposes updating `KWS_SUPPORTED_LANGS` to include all 34 languages, reordered by ISO code – as we do in `src/locale/languages.ts` – and tweaking endonyms. This will enable more users who are verifying their age through KWS to receive emails and view web UI screens in the supported language of their choice.

Also, as a consequence of Chinese (Traditional) (`zh-Hant`) now being supported by KWS, this PR updates the `convertToKWSSupportedLanguage` function to map the `zh-Hant-HK` and `zh-Hant-TW` app languages to `zh-Hant` for KWS.


estrattonbailey (approved):
Thank you!

PR #11091, opened 2026-07-07
Collapse label appeal flow into ModerationDetailsDialog

amckinley:
## Why

Labels on your own posts currently render twice: once as the normal label chips, and again as a "1 content label" pill whose only reason to exist is that its dialog carries the Appeal button. Clicking the label chip itself opens a read-only details popup with no appeal path, which is non-obvious ([slack thread](https://bluesky-builders.slack.com/archives/C063SSE1Z7H/p1783452555015729)). The extra pill also reads vaguely ominous for benign third-party labels.

## What

- `ModerationDetailsDialog` (opened by clicking any label chip) now shows an **Appeal** button in the source footer when the label targets the viewer's own account or content (`label.uri` is their DID or an `at://<did>/...` URI), was not self-applied, and there's a session. Pressing it swaps the dialog body to the appeal form; Back returns to details.
- Extracted `AppealForm` from `LabelsOnMeDialog.tsx` into `src/components/moderation/AppealForm.tsx` (behavior unchanged) so both dialogs share it.
- Removed the per-post `LabelsOnMyPost` pill from all 5 render sites (feed item, thread post/anchor/tree post, generic post). The appeal path is now the chip itself.
- `LabelsOnMe` is now account-only (its sole remaining caller is the profile header, which keeps the "N account labels" pill — account labels don't render as chips on your own profile, so that pill isn't redundant).

## Notes for review

- Eligibility keys off the label subject URI, so the Appeal button also appears when viewing an account-level label's details via regular chips — previously only reachable through the profile-header pill. Believed to be an improvement rather than a regression, but flagging as a deliberate behavior expansion.
- `LabelsOnMeDialog` keeps its `type: 'account' | 'content'` prop as-is; only the pill component was narrowed.
- Deliberate narrowing (flagged by review): the old pill read `post.labels` directly, so it surfaced an appeal path even for labels you've configured to “off” for a labeler you subscribe to. Those labels no longer render a chip, so they lose their per-post appeal entry point. (Labels from *unsubscribed* labelers are unaffected — they never arrive in `post.labels`, since the `atproto-accept-labelers` header is built from your subscribed labelers.) Accepting this: the user explicitly hid the label, and re-adding chrome for that case defeats the purpose of the PR. If it matters, the labeler settings surface is a better future home for it.

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm test` (34 suites / 623 tests) all green.
- No existing tests cover these components specifically.

### Before Screenshots
before-content-label-pill
before-details-no-appeal
before-labels-on-me-dialog

### After Screenshots

after-appeal-form
after-details-appeal
after-post-no-pill

### DIFF
--- a/src/components/Pills.tsx
+++ b/src/components/Pills.tsx
@@ -151,6 +151,83 @@ export function Label({
   )
 }
 
+export type LabelBaseProps = {
+  label: string
+  onPress: () => void
+  noBg?: boolean
+} & CommonProps
+
+export function LabelBase({label, onPress, size = 'sm', noBg}: LabelBaseProps) {
+  const t = useTheme()
+
+  const {outer, text} = useMemo(() => {
+    switch (size) {
+      case 'lg': {
+        return {
+          outer: [
+            t.atoms.bg_contrast_25,
+            {
+              gap: 5,
+              paddingHorizontal: 5,
+              paddingVertical: 5,
+            },
+          ],
+          text: [a.text_sm],
+        }
+      }
+      case 'sm':
+      default: {
+        return {
+          outer: [
+            !noBg && t.atoms.bg_contrast_25,
+            {
+              gap: 3,
+              paddingHorizontal: 3,
+              paddingVertical: 3,
+            },
+          ],
+          text: [a.text_xs],
+        }
+      }
+    }
+  }, [t, size, noBg])
+
+  return (
+    <>
+      <Button
+        label={label}
+        onPress={e => {
+          e.preventDefault()
+          e.stopPropagation()
+          onPress()
+        }}>
+        {({hovered, pressed}) => (
+          <View
+            style={[
+              a.flex_row,
+              a.align_center,
+              a.rounded_full,
+              outer,
+              (hovered || pressed) && t.atoms.bg_contrast_50,
+            ]}>
+            <Text
+              emoji
+              style={[
+                text,
+                a.font_semi_bold,
+                a.leading_tight,
+                t.atoms.text_contrast_medium,
+                {paddingRight: 3},
+              ]}>
+              {label}
+            </Text>
+          </View>
+        )}
+      </Button>
+    </>
+  )
+}
+
 export function FollowsYou({size = 'sm'}: CommonProps) {
   const t = useTheme()
 
--- a/src/components/Pills.tsx
+++ b/src/components/Pills.tsx
@@ -93,7 +93,15 @@ export function Label({
 }
 
 export type LabelBaseProps = {
+  /**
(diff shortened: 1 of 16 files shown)


amckinley:
Re the review's main finding (appeal access narrowing for content labels): verified against `@atproto/api`'s `moderation/decision.ts` — the reviewer is right about the mechanism but slightly overstates the scope. `addLabel` skips labels from unconfigured labelers *and* labels set to `ignore`, so those never produce a chip. However, labels from labelers you don't subscribe to generally never arrive in `post.labels` in the first place (the `atproto-accept-labelers` header is derived from your subscribed labelers), so the old pill mostly couldn't show them either. The real delta is labels you've explicitly set to "off" for a subscribed labeler — those used to get a pill, and now have no per-post appeal path.

Accepting that narrowing as intended: if the user configured a label to be invisible, re-adding a pill just for its appeal path recreates the chrome this PR removes. Noted in the PR description; if there's demand, the labeler settings screen is a better home for appealing hidden labels.

mozzius:
I think we might want to keep the weird "1 label on your post" thing but filter out labels that appear as labels - so it only appears for "off" labels

estrattonbailey:
Post in question here has `Spam` `Rude` `Bluesky Elder` and `test` labels applied.

## As author

#### Before
CleanShot 2026-07-09 at 20 49 39

#### After
CleanShot 2026-07-09 at 20 49 46

<hr />

## As viewer
No change

<hr />

## As author, with all labels off

#### Before
CleanShot 2026-07-09 at 20 51 43

#### After
This now includes labels applied on the account as well as the post, to be consistent.
CleanShot 2026-07-09 at 20 51 47

<hr />

## With all labels off
Both viewer/author views are the same.
CleanShot 2026-07-09 at 20 54 42

<hr />

## As author, only info/alert labels

#### Before
CleanShot 2026-07-09 at 21 01 42

#### After
CleanShot 2026-07-09 at 21 01 46


<hr />

## Only info/alert labels
Both viewer/author views are the same
CleanShot 2026-07-09 at 20 59 30

<hr />

## Appeal button now in all label buttons
Improved the layout from the initial commit on this PR
CleanShot 2026-07-09 at 21 05 54

Also included the admonition stating which labels are applied to the account (this was a recent add elsewhere, parity here)
CleanShot 2026-07-09 at 21 06 36

<hr />

## Additional labels
This is the `+n` or `4 labels applied` pill upon click. Labels applied to accounts carry the same admonition.

CleanShot 2026-07-09 at 21 07 23

estrattonbailey (approved):
Made my proposed changes, if all looks good we can merge this

PR #11094, opened 2026-07-08
Lazy-load mediabunny on web

mozzius:
## Why

`mediabunny` is only ever executed when a user picks, drops, or restores a video in the composer, but it was statically imported by `compress.web.ts` and `metadata.web.ts` — both pulled in via the composer's static import chain from `App.web.tsx` — so it landed in the main webpack bundle and was parsed on every page load.

## What

- New `src/lib/media/video/mediabunny.ts` with a `loadMediaBunny()` helper: a cached dynamic `import('mediabunny')` split into its own chunk (`webpackChunkName: "mediabunny"`). The cache resets on failure so a transient chunk-load error doesn't poison the session.
- `compress.web.ts`: mediabunny imports are now type-only; runtime symbols are destructured from `await loadMediaBunny()` inside `probeWithMediaBunny`, `doCompression`, `findEncodableVideoCodec`, and `findEncodableAudioCodec`. The load deliberately happens inside functions already wrapped in try/catch, so a chunk-load failure falls back to the raw passthrough upload rather than failing the post.
- `metadata.web.ts`: same treatment in `getMetadataWithWebCodecs()`; a chunk failure takes the existing WebCodecs-failed path and falls back to the browser `<video>` API. `hasWebCodecs()` is untouched, so `SelectMediaButton`'s import no longer drags mediabunny into the main bundle.

## Test plan

- `pnpm typecheck`, `pnpm lint` pass; `pnpm test` has 2 pre-existing failures in `src/logger/__tests__/logger.test.ts` that fail identically on main.
- Manual: attach an .mp4 in the web composer — the mediabunny chunk loads only at that moment (DevTools Network) and compression/upload work normally. A GIF takes the passthrough path without loading the chunk. Blocking the chunk URL falls back to an uncompressed upload with the existing warning log.

### DIFF
--- a/src/lib/media/video/compress.web.ts
+++ b/src/lib/media/video/compress.web.ts
@@ -1,17 +1,8 @@
 import {type ImagePickerAsset} from 'expo-image-picker'
 import {
-  ALL_FORMATS,
   type AudioCodec,
-  BlobSource,
-  BufferTarget,
-  canEncodeAudio,
-  canEncodeVideo,
-  Conversion,
-  Input,
-  Mp4OutputFormat,
-  Output,
+  type Input as MediaBunnyInput,
   type VideoCodec,
-  WebMOutputFormat,
 } from 'mediabunny'
 
 import {VIDEO_MAX_SIZE} from '#/lib/constants'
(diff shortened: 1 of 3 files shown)


vineyardbovines (approved):
sweeet

PR #11098, opened 2026-07-08
Replace @tiptap/html generateJSON with native DOMParser equivalent

mozzius:
## Why

`@tiptap/html` exists to support SSR: it parses HTML via `zeed-dom` instead of the browser's DOM. That pulls ~250KB (stat size) of parser dependencies into the initial web bundle – `entities` (~211KB), `css-what` (~39KB), plus `zeed-dom` itself. Our only usage is in the composer's `TextInput.web.tsx`, which always runs in a real browser. We can drop the dependency entirely

## How

Replaced `generateJSON` from `@tiptap/html` with a local `richTextToEditorJson` helper that does the same thing using the native `DOMParser` plus ProseMirror's `DOMParser.fromSchema` – which is exactly what `@tiptap/core`'s own browser version does. The HTML is wrapped in `<body>` before parsing to preserve leading whitespace (same trick `@tiptap/core` uses internally).

## Verification

- Diff-tested the old and new implementations side by side (jsdom) across mentions, multiline text, leading/trailing whitespace, HTML entities, emoji, and empty strings – byte-identical ProseMirror JSON output for all cases.
- Rebuilt with `EXPO_PUBLIC_GENERATE_STATS=1`: `zeed-dom`, `entities`, `css-what`, and `htmlparser2` are gone from the bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,7 +140,6 @@
     "@tiptap/extension-paragraph": "^2.9.1",
     "@tiptap/extension-placeholder": "^2.9.1",
     "@tiptap/extension-text": "^2.9.1",
-    "@tiptap/html": "^2.9.1",
     "@tiptap/pm": "^2.9.1",
     "@tiptap/react": "^2.9.1",
     "@tiptap/suggestion": "^2.9.1",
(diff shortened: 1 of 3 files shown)


estrattonbailey (approved):
If you've tested it, SGTM!

PR #11099, opened 2026-07-08
Lazy load @ipld/dag-cbor in computeCid

mozzius:
## Why

`@ipld/dag-cbor` and its `cborg` dependency (~190KB stat size combined) were in the initial web bundle, but they have exactly one consumer in the whole module graph: `computeCid()` in `src/lib/api/index.ts`, which is only called when posting a thread (to compute the CID of the just-created post so the next post can reference it as its reply parent).

## How

Switched the static import to a dynamic `await import('@ipld/dag-cbor')` inside `computeCid`. The function is already async and its only call site already awaits it, so there's no behavior change – the chunk downloads once, during a network-bound flow.

## Verification

Built with `EXPO_PUBLIC_GENERATE_STATS=1` and confirmed via `stats.json`: `@ipld/dag-cbor` and all `cborg` modules moved from the initial chunk to an async chunk.

### DIFF
--- a/src/lib/api/index.ts
+++ b/src/lib/api/index.ts
@@ -16,7 +16,6 @@ import {
   RichText,
 } from '@atproto/api'
 import {TID} from '@atproto/common-web'
-import * as dcbor from '@ipld/dag-cbor'
 import {t} from '@lingui/core/macro'
 import {type QueryClient} from '@tanstack/react-query'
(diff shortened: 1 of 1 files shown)


estrattonbailey (approved):

estrattonbailey:
Didn't test

PR #11101, opened 2026-07-08
Revert "Add more search placeholders (#11089)"

ds-boyce:
The “Advanced search” button only appears after searching, so in practice the “all of these words” field is always populated and you only see one half of these pairings. They don’t make a lot of sense on their own.

This reverts commit a689322f4aeec38a58a0fd8fb84284ef2d341c40.


mozzius (approved):
Lol sorry, should've looked closer

PR #11106, opened 2026-07-08
Adjust AA device signals copy

estrattonbailey:
Clarifies what sharing device data means for users, inspired by Surfdude's comment here https://github.com/bluesky-social/social-app/pull/9564#issuecomment-4870981554 CleanShot 2026-07-08 at 11 22 26


surfdude29:
@estrattonbailey Looks great, thanks! 🙏

PR #11111, opened 2026-07-08
Add video playback observability to Sentry

vineyardbovines:
Adds Sentry performance spans around video playback in feeds, to establish old-arch baselines for scroll + autoplay performance ahead of the new architecture upgrade.

## What

- Enables `tracesSampleRate` (1.0 internal, 1% prod). This is what makes spans send at all, and it auto-enables the SDK's JS stall tracking and slow/frozen frame tracking, whose measurements attach to every root span. Also adds a `new_arch` tag for filtering.
- New `createPlaybackTelemetry` (`src/lib/media/video/playbackTelemetry.ts`): opens one `video.playback` root span per activation window (video becomes the active autoplaying one, ends on deactivation) with `surface`, `presentation`, `timeToReadyMs`, `timeToFirstPlayMs`, `outcome`, and error info. Since these are root spans, stall and frame measurements ride along, capturing scroll smoothness while a video is on screen. Sentry-only; nothing goes to the analytics pipeline.
- Wires it into both native surfaces:
  - in-feed embeds, via the `BlueskyVideoView` event callbacks in `VideoEmbed/index.tsx` (plus an optional `onError` prop on `VideoEmbedInnerNative`)
  - the immersive video feed, via a `usePlaybackTelemetry` hook on the expo-video players, with a `preloaded` attribute to separate load time from swipe latency

Web is untouched. Side effect: the existing `video.upload` spans, which were created but never sampled, now flow at the same rates.

## Test plan

- Unit tests for the telemetry module (span lifecycle, idempotent end, error outcome, re-activation).
- `pnpm typecheck`, `pnpm lint`, `pnpm test`, `pnpm prettier` pass.
- Runtime spans verified via the next internal build (Sentry is disabled in dev).

### DIFF
--- a/src/components/Post/Embed/VideoEmbed/index.tsx
+++ b/src/components/Post/Embed/VideoEmbed/index.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useRef, useState} from 'react'
+import {useCallback, useEffect, useRef, useState} from 'react'
 import {ActivityIndicator, View} from 'react-native'
 import {ImageBackground} from 'expo-image'
 import {type AppBskyEmbedVideo} from '@atproto/api'
@@ -6,6 +6,10 @@ import {msg} from '@lingui/core/macro'
 import {useLingui} from '@lingui/react'
 import {Trans} from '@lingui/react/macro'
 
+import {
+  createPlaybackTelemetry,
+  type PlaybackTelemetry,
+} from '#/lib/media/video/playbackTelemetry'
 import {ErrorBoundary} from '#/view/com/util/ErrorBoundary'
 import {atoms as a, platform} from '#/alf'
 import {Button} from '#/components/Button'
@@ -75,6 +79,17 @@ function InnerWrapper({embed}: Props) {
   const [isActive, setIsActive] = useState(false)
   const showSpinner = useThrottledValue(isActive && isLoading, 100)
 
+  /*
+   * Created lazily on first activation so videos that are never scrolled into
+   * the active position cost nothing.
+   */
+  const telemetryRef = useRef<PlaybackTelemetry | null>(null)
+  useEffect(() => {
+    return () => {
+      telemetryRef.current?.deactivated()
+    }
+  }, [])
+
   const showOverlay =
     !isActive ||
     isLoading ||
@@ -89,9 +104,33 @@ function InnerWrapper({embed}: Props) {
     <>
       <VideoEmbedInnerNative
         embed={embed}
-        setStatus={setStatus}
-        setIsLoading={setIsLoading}
-        setIsActive={setIsActive}
+        setStatus={s => {
+          setStatus(s)
+          if (s === 'playing') {
+            telemetryRef.current?.playing()
+          }
+        }}
+        setIsLoading={loading => {
+          setIsLoading(loading)
+          if (!loading) {
+            telemetryRef.current?.ready()
+          }
+        }}
+        setIsActive={active => {
+          setIsActive(active)
+          if (active) {
+            telemetryRef.current ??= createPlaybackTelemetry({
+              surface: 'feed',
+              presentation: embed.presentation === 'gif' ? 'gif' : 'video',
+            })
+            telemetryRef.current.activated()
+          } else {
+            telemetryRef.current?.deactivated()
+          }
+        }}
+        onError={error => {
+          telemetryRef.current?.error(error)
+        }}
         ref={ref}
       />
       <ImageBackground
--- a/src/lib/media/video/playbackTelemetry.ts
+++ b/src/lib/media/video/playbackTelemetry.ts
@@ -1,0 +1,83 @@
+import {Sentry} from '#/logger/sentry/lib'
+
+/**
+ * Where the video is being played.
+ */
+export type PlaybackSurface = 'feed' | 'immersiveFeed'
+
+export type PlaybackTelemetry = {
+  activated: (opts?: {preloaded?: boolean}) => void
+  ready: () => void
+  playing: () => void
+  error: (e: unknown) => void
+  deactivated: () => void
+}
+
+/**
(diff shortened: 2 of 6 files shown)


src/components/Post/Embed/VideoEmbed/index.tsx
```diff
@@ -89,9 +104,33 @@ function InnerWrapper({embed}: Props) {
...
+          setIsActive(active)
+          if (active) {
+            telemetryRef.current ??= createPlaybackTelemetry({
+              surface: 'feed',
+              presentation: embed.presentation === 'gif' ? 'gif' : 'video',
+            })
+            telemetryRef.current.activated()
+          } else {
+            telemetryRef.current?.deactivated()
+          }
```
estrattonbailey, line 129: So if it becomes active again, will we re-open the span or start a new one, or can it only report telemetry once in it's lifecycle? Thinking about a user like scrolling back up to a previous video.

src/lib/media/video/playbackTelemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+    span = undefined
+  }
+
+  return {
+    activated(opts) {
+      if (span) return
+      activatedAt = Date.now()
+      sawReady = false
+      sawPlaying = false
+      span = Sentry.startInactiveSpan({
```
estrattonbailey, line 50: Ok looks like it will create a new one, nice

estrattonbailey (approved):
LGTM

mozzius:
Reviewed the diff plus the vendored `@sentry/react-native` 8.18 integration code, since the risky part of this PR is what `tracesSampleRate` turns on implicitly. Tests pass, typecheck/lint clean. The telemetry module itself is solid — the findings are all side effects of enabling tracing.

### 1. The first playback span per app run gets hijacked as the app-start transaction (high, native)

Setting `tracesSampleRate` auto-enables `appStartIntegration` (default `enableAppStartTracking: true`, non-standalone). In non-standalone mode the SDK locks onto **the first sampled root span of the app run** (`recordFirstStartedActiveRootSpanId` in `dist/js/tracing/integrations/appStart.js`) and, when that span's transaction is processed, rewrites it: `op` becomes `ui.load` and `start_timestamp` is re-anchored to process start.

We register no navigation instrumentation (no `reactNavigationIntegration` / `registerNavigationContainer` anywhere), so the first root span of a run will always be a `video.playback` or `video.upload` span. Concretely: user cold-starts the app, scrolls to a video within 60s → that playback span is reported with `op: ui.load` and a duration measured from process launch. It vanishes from `op:video.playback` queries — and it's exactly the "cold start → first video" case you most want in an old-arch baseline. On internal builds (100% sampling) this corrupts **every** app run's first video span.

**Fix:** add `enableAppStartTracking: false` to `init()`, or opt into standalone app-start tracing (`_experiments.enableStandaloneAppStartTracing: true`) if you want app-start data as its own transaction.

### 2. Tracing enables `sentry-trace`/`baggage` header injection on every outgoing request (medium, native)

`hasTracingEnabled` also pulls in `reactNativeTracingIntegration` (default `enableAutoPerformanceTracing: true`), which calls `instrumentOutgoingRequests` with `tracePropagationTargets` defaulting to `[/.*/]` on native. Every XHR/fetch from the app — including requests to arbitrary third-party PDSes, which is routine on atproto — now carries `sentry-trace` and `baggage` headers (baggage includes release, environment, and the Sentry public key). That's new metadata leakage to servers we don't control, and it's not needed for this PR's goal (no distributed tracing backend to continue into).

**Fix:** set `tracePropagationTargets: []` in `init()` to keep tracing on but disable header injection.

### 3. "Web is untouched" isn't accurate (low)

`src/logger/sentry/setup/index.ts` has no `.web` variant and is imported by `App.web.tsx`, so web also gets `tracesSampleRate`. Consequences: web `video.upload` spans (from the shared `telemetry.ts`, used by `upload.web.ts`) now flow at 1%/100%; the stall-tracking integration is installed on web too (it's not gated by `notWeb()`), running its 50ms loop while any span is open; and browser request instrumentation activates. None of this looks harmful, but it contradicts the PR description and deserves a deliberate call. Also, the `new_arch` tag is set on web where it's always `false` — harmless but pollutes the dashboard filter; could be gated to native.

### 4. Feed embeds mostly won't record `timeToReadyMs` for already-loaded videos (low, data quality)

In `VideoEmbed/index.tsx`, `ready()` is driven by `onLoadingChange(false)`. If the video finished loading before it becomes active (or was previously active — a scroll-away-and-back reopens a span but no new loading transition occurs), `ready()` either fires before the span exists (no-op) or never fires, so the attribute is simply absent. The immersive feed handles this explicitly with the `preloaded` check; the feed surface has no equivalent, so `timeToReadyMs` coverage will be biased toward slow loads. Worth either mirroring the `preloaded` idea (e.g. mark spans where no load transition was observed) or just documenting that the attribute means "load completed during this activation window."

### Minor notes

- A span stays open while the app is backgrounded mid-playback (deactivation never fires), so durations for those windows include background time. Stall tracking pauses in background, but the span duration itself is inflated. Consider ending on `AppState` background, or accept the noise knowingly.
- GIFs create a span per activation window, so scrolling a GIF-heavy feed on an internal build (100% sampling) will be chatty. The `presentation` attribute makes them filterable, so probably fine — just flagging volume.

### What's good

`createPlaybackTelemetry` is clean: idempotent `end()`, guarded re-activation, error path matches the existing `telemetry.ts` conventions, lazy creation so never-activated videos cost nothing, and the tests cover the lifecycle well. The two-surface wiring is sensible and the `preloaded` distinction in the immersive feed is a nice touch. The claim that stall/frame measurements attach to these root spans checks out in the SDK source — both integrations key off any sampled root span.

**Verdict:** findings 1 and 2 are one-line `init()` fixes and worth making before merge — #1 in particular undermines the baseline data this PR exists to collect.

PR #11113, opened 2026-07-08
Potentially fix disappearing chat emoji on some devices

ds-boyce:
This was reported on a Samsung S22 Ultra:

https://github.com/user-attachments/assets/c4624c5d-de33-4406-b2eb-5bde320d938b

I’ve been unable to repro it on other Android devices, so this is a potential fix:

* Don’t bother animating the border radius on emoji-only messages since we don’t show them in bubbles anyway (avoids a second render).

### DIFF
--- a/src/components/dms/MessageItem.tsx
+++ b/src/components/dms/MessageItem.tsx
@@ -580,14 +580,6 @@ let MessageItem = ({
                                 android: {marginTop: a.mt_2xs.marginTop},
                                 default: {marginBottom: -a.mb_sm.marginBottom},
                               }),
-                            /*
-                             * On Android, give the enlarged glyph run a
-                             * self-contained draw box so a re-measure can't clip
-                             * the trailing emoji. `includeFontPadding` restores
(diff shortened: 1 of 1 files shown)


src/components/dms/MessageItem.tsx
```diff
@@ -553,14 +572,22 @@ let MessageItem = ({
...
                                 default: {marginBottom: -a.mb_sm.marginBottom},
                               }),
+                            /*
+                             * On Android, give the enlarged glyph run a
+                             * self-contained draw box so a re-measure can't clip
+                             * the trailing emoji. `includeFontPadding` restores
+                             * the ascent/descent slack that `leading_tight`
+                             * removes.
+                             */
+                            platform({android: {includeFontPadding: true}}),
```
mozzius: I think this is the default value?

ds-boyce: Yep, removed.

estrattonbailey (approved):
Love it

PR #11114, opened 2026-07-09
bskyogcard: downscale avatars before rendering to stop OOM cycle

amckinley:
## Problem

`bskyogcard` in prod OOMs against its 2Gi limit roughly hourly (~20 OOMKills/day/pod, chronic for weeks — surfaced by a new sustained-restart-loop alert on the infra side). Not a JS heap leak: heap stays ~30-60Mi. The growth is native memory.

The card routes fetch avatars from `cdn.bsky.app/img/avatar/`, which serves up to 1000×1000, while cards display avatars at ≤210px. Both `satori` (image parsing) and `resvg` (rasterization) operate at intrinsic image size — ~4MB of decoded RGBA per avatar per render, ×7 avatars per starter-pack card. satori's internal 500-entry image LRU also retains the oversized data URIs (~100Mi of heap per cluster worker at steady state). Sustained unique-image traffic plus allocator fragmentation ratchets RSS until the kernel kills the pod. Because workers survive ~10min between kills, the pod never enters CrashLoopBackOff and this was invisible until cadence-based alerting.

## Fix

Downscale fetched avatars with `sharp` to 256px max (`fit: inside`, no enlargement, JPEG q95) before handing bytes to the render pipeline, and disable libvips' operation cache (50MB/process, useless for unique-image traffic). Source pixels are the same full-res CDN asset as today and display sizes never exceed the 256px cap, so rendered output is visually identical (verified by rendering identical starter packs before/after and comparing).

## Numbers

Load test: 74 real starter packs on loop, concurrency 8, 6 rounds (444 renders):

| | peak RSS | JS heap during load | heap floor after idle |
|---|---|---|---|
| before | 1441Mi, climbing round-over-round | 130-160Mi | 126Mi (LRU retention) |
| after | ~1000Mi, flat by round 3 | 45-70Mi | 47Mi |

Decoded pixel volume per avatar drops ~15×. The resize itself is ~5-10ms per image, amortized inside the CDN fetch round-trip.

## Notes

- `pnpm-lock.yaml` shows ~40 lines of pnpm `11.5.2 → 11.10.0` pin churn in the env document (first YAML doc): `pnpm add sharp` synced it to the `devEngines` pin in `package.json`, which had drifted. Legit fallout of touching the lockfile with the correct pnpm version.
- `sharp` ships prebuilt `linuxmusl` binaries, so the alpine Docker build needs no toolchain changes.

### DIFF
--- a/bskyogcard/package.json
+++ b/bskyogcard/package.json
@@ -28,6 +28,7 @@
     "prom-client": "^15.1.3",
     "react": "^19.2.6",
     "satori": "^0.26.0",
+    "sharp": "^0.35.3",
     "twemoji": "^14.0.2"
   },
   "devDependencies": {
--- a/bskyogcard/pnpm-lock.yaml
+++ b/bskyogcard/pnpm-lock.yaml
@@ -7,52 +7,52 @@ importers:
     configDependencies: {}
     packageManagerDependencies:
       '@pnpm/exe':
-        specifier: 11.5.2
-        version: 11.5.2
+        specifier: 11.10.0
+        version: 11.10.0
       pnpm:
-        specifier: 11.5.2
-        version: 11.5.2
+        specifier: 11.10.0
+        version: 11.10.0
 
 packages:
 
-  '@pnpm/exe@11.5.2':
-    resolution: {integrity: sha512-4UFnP2rhNu1xjAQ+I1GdIUUEtCJuTYJlbpiWSFA4POAID3Lpt+2vrjImWO7eOJ7iCY3vpc4TFe2IW3sAolW4Kg==}
+  '@pnpm/exe@11.10.0':
+    resolution: {integrity: sha512-mrmfi2C7LpZkyq0voKKye6MzrK8/K7tYQRiSB/jqOiwnFtQxxcA3xGTY9cEsrGpNl2ClPecQofLuj+BO8AIsxw==}
     hasBin: true
 
-  '@pnpm/linux-arm64@11.5.2':
-    resolution: {integrity: sha512-MbJySnu2y9cCBqlODLjUlZ87JnRC3Inq40rvGHWJSrSQ0PnuHeSw2NDMnLI8Hf9hCY+ooussRc5iiR4IAkjUvg==}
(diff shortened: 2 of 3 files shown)


mozzius (changes requested):
I think it would be a heck of a lot easier to just fetch the thumbnail variant of the avatar, no?

amckinley:
lol good call, i’ll do that instead

amckinley:
yeah so the thumbnail variant was actually the first thing we tried last night, and i vetoed it on quality: `avatar_thumbnail` is 128px, but the starter-pack foreground circles render at ~202px, so it gets *upscaled* ~1.6x and you can see it. the sharp version fetches the same full-res source as prod today and only ever downscales (256 > 202), which is why it's indistinguishable from baseline.

side-by-side crops from identical renders (2x zoom, nearest-neighbor so github doesn't smooth over the evidence):

pack4 avatar comparison

pack2 avatar comparison

B is the thumbnail approach — look at the hair strands and the linework edges. A vs C is the diff this PR actually ships, i cant tell them apart.

memory-wise both approaches fix the OOM about equally (the killer was decoding 1000px sources at render time; whether the small image comes pre-shrunk from the CDN or gets shrunk in-process doesnt change the render-side math much). so its purely a "one line vs a dep" tradeoff against visible avatar quality. im happy either way but wanted you deciding with the crops in front of you rather than from the diff alone

(comparison images live on a throwaway `ogcard-comparison-assets` orphan branch, delete after this merges)

estrattonbailey (approved):
Another alternative would be to introduce a fresh size at the CDN layer, but idk if we wanna absorb that stampede rn.

This seems fine as a first pass. If it Just Works, then great. If not, we can swap to the one liner and consider a middle-tier CDN transform.

amckinley:
@mozzius @estrattonbailey these pods are still regulaly crash looping in prod, so im just going to double their already-crazy memory limits. i still think this^^ or something similar to it is better than throwing gigs of memory at this workload but ill stop the bleeding now

estrattonbailey:
@amckinley yeah I say let's try this, you're good to merge

PR #11118, opened 2026-07-09
Hide video pillarboxing on web when the card fits

vineyardbovines:
On web, videos always rendered as a full-width card with black pillarbox bars around the video. This renders portrait videos at their own aspect ratio (matching native) when the resulting card is wide enough, and softens the bars when it isn't.

- The card keeps the full-width pillarbox when it would be narrower than 280px (measured via onLayout), when the video is taller than 1:2, or when the aspect ratio is missing - a narrow card can't fit the overlay controls, and a ratio-fit card would still show bars for very tall videos (see #9371).
- When the pillarbox is shown, a blurred cover-fit copy of the thumbnail fills the bars instead of flat black.

Supersedes #10303.

Tested on the web dev server: portrait 1052x1712 renders at its own ratio on desktop and flips to the blurred full-width card below the 280px floor (both directions on window resize), 381x800 (taller than 1:2) keeps the blurred full-width card, landscape unchanged.

before and after

CleanShot 2026-07-10 at 10 52 27@2x


src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -89,20 +96,58 @@ export function VideoEmbed({embed}: {embed: AppBskyEmbedVideo.View}) {
     constrained = Math.max(aspectRatio, ratio)
   }
 
+  const [containerWidth, setContainerWidth] = useState(0)
```
estrattonbailey, line 99: This is a nit, but can we use the aspect ratio to calculate width (constrained by a max height, which I think we do on native) so that we don't need to do a FOUC? Could use a heuristic min window width and fall back to letterbox in those cases.

src/components/Post/Embed/VideoEmbed/index.web.tsx
```diff
@@ -89,20 +96,58 @@ export function VideoEmbed({embed}: {embed: AppBskyEmbedVideo.View}) {
...
+            inset: 0,
+            backgroundImage: `url(${embed.thumbnail})`,
+            backgroundSize: 'cover',
+            backgroundPosition: 'center',
+            filter: 'blur(32px)',
+            // hide the transparent fade the blur creates at the edges
+            transform: 'scale(1.2)',
+          }}
+        />
+      )}
```
estrattonbailey, line 158: Love this, looks great imo

src/components/Post/Embed/VideoEmbed/index.web.tsx
vineyardbovines: Done in 35ee7d5 - dropped the onLayout measurement and estimate the card width from window width capped at CENTER_COLUMN_WIDTH minus ~100px of feed chrome, so the decision is right on first paint. Narrower containers (quotes, DMs) can overestimate slightly, but the ratio-fit card is bounded by the same max-height math either way.

estrattonbailey (changes requested):
Ok I was wrong on this, it already works like native as much as it can, and the heuristic check breaks DMs as Claude rightfully pointed out.

The measurement is pry the best path forward. Give all embed locations a quick fuzz on native/web and diff screen sizes and that should do the trick.

Also, I'm missing the thumbnail as the video bg on web, shows a flash of black before the video shows.

vineyardbovines:
Addressed in 6cac311:

- Dropped the window-width heuristic and went back to measuring the card container with onLayout. The measured width falls back to the full-width pillarbox in narrow containers, so DMs and quote embeds are covered. The decision lands a frame after first paint; only a horizontal snap, no vertical layout shift.
- Restored the thumbnail as the container background, same as main (contain over black), so it shows behind/before the video in both the full-bleed and ratio-fit presentations. The blurred backdrop plus a sharp redraw above it still render in the full-bleed path only.

Still needs the manual fuzz pass across embed locations (feed, thread, quotes, DMs) on native/web at different screen sizes.

vineyardbovines:
https://github.com/user-attachments/assets/9016f49b-ae91-4651-858a-7a5844cb1102

fixed that flickering

vineyardbovines:
CleanShot 2026-07-21 at 11 38 32@2x
CleanShot 2026-07-21 at 11 38 38@2x

estrattonbailey (approved):
Sick as, looks great 👍

PR #11120, opened 2026-07-09
Submit beta feature feedback via Sentry

ds-boyce:
https://github.com/user-attachments/assets/630b1cfc-f4ea-43c6-a5c7-2d36b493d1fb


ThisIsMissEm:
I guess this makes sense, but part of me also wants to give a shout out to https://userinput.app  (AT Protocol native!)

PR #11126, opened 2026-07-09
Add From: Me to advanced search filters

ds-boyce:
In addition to searching via `from:me` in the query, this allows selecting yourself from the “Filter by author” dropdown in the UI.

Existing URLs are backwards compatible, so no links or bookmarks are broken.

ios

### DIFF
--- a/src/screens/Search/components/AdvancedSearchDialog/utils.ts
+++ b/src/screens/Search/components/AdvancedSearchDialog/utils.ts
@@ -14,10 +14,12 @@ export type RepliesFilter = 'all' | 'none' | 'only'
 export type MediaFilter = 'all' | 'media' | 'video'
 
 /**
- * Whether to limit results to authors the user follows. Serializes into the
- * `following` sibling param ('following' -> following:true, anyone -> unset).
+ * Which authors to limit results to. Serializes into the `from` sibling param:
+ * 'following' -> from:true (people you follow), 'you' -> from:me (your own
+ * posts, reconstructed into `from:me` at the v2 API boundary), 'anyone' ->
+ * unset.
  */
-export type FollowingFilter = 'anyone' | 'following'
+export type FollowingFilter = 'anyone' | 'following' | 'you'
 
 export type FilterField = 'authors' | 'mentions' | 'domains' | 'urls' | 'tags'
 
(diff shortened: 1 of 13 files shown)


src/screens/Search/components/AdvancedSearchDialog/utils.ts
```diff
@@ -14,10 +14,12 @@ export type RepliesFilter = 'all' | 'none' | 'only'
...
 /**
- * Whether to limit results to authors the user follows. Serializes into the
- * `following` sibling param ('following' -> following:true, anyone -> unset).
+ * Which authors to limit results to. Serializes into the `from` sibling param:
+ * 'following' -> from=following (people you follow), 'you' -> from=me (your own
+ * posts, reconstructed into `from:me` at the v2 API boundary), 'anyone' ->
+ * unset.
  */
-export type FollowingFilter = 'anyone' | 'following'
+export type FollowingFilter = 'anyone' | 'following' | 'you'
```
estrattonbailey: I'm very confused that this is called "following filter", but it's legacy, and "you" means "me". Is there a way to clarify this relationship more?

src/screens/Search/searchParams.ts
```diff
@@ -50,9 +54,17 @@ export const FILTER_PARAM_KEYS = [
...
+  'from',
 ] as const
 
+/**
+ * Filter param keys that are no longer written but may still be present in old
+ * URLs/history. `readSearchFilters` maps these onto their current equivalents;
+ * the param-rebuild helpers strip them so a cleared filter can't be resurrected
+ * from a stale legacy key. `following=true` was the old form of `from=following`.
+ */
+export const LEGACY_FILTER_PARAM_KEYS = ['following'] as const
```
estrattonbailey: I thought we only wrote lexicon params to URL query params for sharing. `following: true` is still a valid param in the lexicon, I don't think we wanna write non-standard params to the URL if we can help it.

estrattonbailey: Or I guess this is a product of adding unrelated props to the dropdown formerly labeled as "From". The params we use for state in the form and the URL query params we write should be decoupled if possible

src/screens/Search/components/AdvancedSearchDialog/utils.ts
ds-boyce: This was a case of failure to rename the type.

src/screens/Search/searchParams.ts
ds-boyce: Ah, I didn’t know that was the case, or I would have remapped `q` to `query` as well. 😅 

In either case, keeping both could yield a technically invalid query `?following=true&from=me`, although the UI itself would never construct it. So maybe that’s fine.

estrattonbailey: Yeah I think `q` is fine, but good shout. That one preexists this work, so we can keep that one. We just wanna avoid having to write param remapping as much as we can. If it stays basically 1:1 with the lexicon, in theory we don't have backwards compat issues in the future.

ds-boyce: Makes sense to me, I’ll revise this accordingly.

estrattonbailey (approved):

PR #11127, opened 2026-07-10
Fix Android silently dropping the last line of wrapped text

### LINKED ISSUE #5235: Truncated Widowed Words in Feed View on Android

diabler-ie:
## Summary

Fixes #5235 (and duplicates #5835, #7122, #8161, #8465).

Replaces #10523, which I am closing after finding the root cause.

On Android, the last word or words of a post are sometimes missing in the feed. The post detail view shows the full text. There is no error and no `…` — the text just ends early. Users have reported this since 2024, on many Android versions, and changing the system font size makes it appear or disappear. The root cause explains all of this.

## Root cause

React Native (old architecture, which this app uses) wraps every `<Text>` into lines **twice**, in two separate places:

1. **Measurement.** `ReactTextShadowNode.measureSpannedText` builds a `StaticLayout` with width `(int) Math.ceil(availableWidth)`. The number of lines from this step decides the height of the text's box. ([RN 0.81 source](https://github.com/facebook/react-native/blob/v0.81.5/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/text/ReactTextShadowNode.kt))
2. **Render.** The `ReactTextView` (a normal Android `TextView`) receives the same text and wraps it again on its own, using its final frame width. Because of rounding (the `ceil` above, plus Yoga rounding positions to whole pixels), this width can be **1-2px smaller** than the width used in step 1.

So the two steps can disagree. If a line is almost exactly as wide as the text box, it fits in step 1 but does not fit in step 2. The renderer then needs one more line than the box has room for. The extra line is drawn below the bottom of the box, where it is cut off and invisible.

To confirm this, I built a debug version of the app that logs the internal text layout of every `TextView`, and tested two live posts that show the bug (Pixel 9 Pro XL):

- On normal lines, the two steps produce **identical line widths** (within 0.03px). Fonts and sizes agree perfectly. Only the wrap width differs.
- Post 1: one line was 816.57px wide. Measurement wrapped the text at width 817, so this line fit. The renderer wrapped at width 816, so the last word (`eyes.`) moved to a 10th line — below a box with room for only 9 lines.
- Post 2: the two steps disagreed already at **line 1**. Every line after it changed too, and the renderer needed 4 lines for a 3-line box.

Post 2 shows why the NBSP idea in #10523 cannot fully work: it only adds width to the *last* line, but the problem can start at any line. @arturo32's observation in #7122 — "the length of each character matters" — was exactly right. Credit to him, and to @surfdude29 for collecting the duplicate issues.

I also ruled out other suspects: `useBoundsForWidth` (turned it off at runtime; the line breaks did not change), font scale (it only changes which lines are near the limit), and `elegantTextHeight` / `letterSpacing` (identical in both steps). A JavaScript fix is not possible: the `onTextLayout` event is produced by the measurement step, so JavaScript can never see what the renderer did differently. This is also why the bug was so hard to find.

## Fix

A config plugin (same pattern as our other plugins that add native code) registers a `ReactPackage` whose view manager for `RCTText` replaces the default one. Its `TextView` subclass reports 2px less right padding. `TextView` computes its wrap width as `view width - padding`, so this makes the render-side wrap width 2px wider. That covers the full rounding difference: the renderer can now always fit what measurement fit, and the two steps agree on the line count again.

Worst case after this change: a near-limit line can draw up to 2px of its last letter outside the view. In practice this is not visible, because the outer 2px of a letter are almost always empty space. Very rarely, a text box can be one line taller than needed. Both outcomes are harmless compared to losing words.

This fix is for the old architecture only. It should be removed when the app moves to the new architecture: Fabric's prepared text layouts (`PreparedLayoutTextView`) draw the measured layout directly, so the two steps cannot disagree there.

## Test plan

- [x] Built this branch (release, Pixel 9 Pro XL, Android 17, font scale 1.15). A live post that reproduces the bug now shows its full text in the feed, and the line breaks match the measurement step within 0.1px. Screenshots below.
- [x] Used the debug probe to confirm the fix is active and works as designed. Same post, control build vs. this branch (`WrapWidthTextView` running, wrap width 818 vs. the 816px frame, layout height == view height):

  ```
  control: viewWH=816x150 layoutWH=816x200 lines=4 [.. |IS going to be w=756.7 |different w=154.7]
  fixed:   viewWH=816x150 layoutWH=818x150 lines=3 [.. |to be different w=788.3]  (WrapWidthTextView)
  ```

  The second reproducer post (the one where only the last line is affected) was checked the same way: on the control build the renderer produced 10 lines for a 9-line box; with the fix, the layout fits the box.
- [x] Scrolled Following, Likes, profile feeds, and post threads. No visual regressions, no changed line wraps outside the near-limit cases, spacing unchanged.
- [x] Text selection and copy work the same as before (the fix does not change the text).
- [ ] Testing on more devices and screen densities is welcome. The mechanism does not depend on density (the difference is a fixed ~2px from pixel rounding), but more coverage is always good.

| Before (control build of main) | After (this branch) |
|---|---|
| Post ends "this time it IS going to be" - the last word is missing | Full text: "this time it IS going to be different" |

Same post, same feed position. Note that line 1 also wraps differently in the fixed build: the renderer now uses the same line breaks as the measurement step.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -303,6 +303,7 @@ module.exports = function (_config) {
         './plugins/withAndroidManifestFCMIconPlugin.js',
         './plugins/withAndroidManifestIntentQueriesPlugin.js',
         './plugins/withAndroidStylesAccentColorPlugin.js',
+        './plugins/androidTextWrapWidthFix/withAndroidTextWrapWidthFix.js',
         './plugins/withAndroidNoJitpackPlugin.js',
         './plugins/shareExtension/withShareExtensions.js',
         './plugins/notificationsExtension/withNotificationsExtension.js',
--- a/plugins/androidTextWrapWidthFix/WrapWidthTextFix.kt
+++ b/plugins/androidTextWrapWidthFix/WrapWidthTextFix.kt
@@ -1,0 +1,57 @@
+package xyz.blueskyweb.app.textwrapfix
+
+import android.content.Context
+import com.facebook.react.ReactPackage
+import com.facebook.react.bridge.NativeModule
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.uimanager.ThemedReactContext
+import com.facebook.react.uimanager.ViewManager
+import com.facebook.react.views.text.ReactTextView
+import com.facebook.react.views.text.ReactTextViewManager
+
+/**
+ * Works around Android silently dropping the last line of wrapped post text
+ * (#5235).
+ *
+ * React Native (old architecture) measures text with a StaticLayout built at
+ * ceil(availableWidth), but the TextView later re-breaks the same text at the
+ * Yoga-rounded integer frame width, which can be up to ~2px narrower
+ * (measurement ceil plus frame position rounding). A line whose advance width
+ * falls inside that window fits during measurement but wraps during render,
+ * pushing the last line below the measured bounds where it is clipped.
+ *
+ * TextView computes its wrap width as viewWidth - compoundPadding, so
+ * reporting slightly less right padding widens the render-side wrap width to
+ * cover the window: the renderer can always fit what measurement fit. Worst
+ * case, a knife-edge line paints up to 2px of a glyph edge past the view
+ * bounds instead of dropping a whole word.
+ *
+ * Old architecture only. Remove after migrating to the new architecture:
+ * Fabric's prepared text layouts draw the measured layout directly, so the
+ * two passes cannot disagree.
+ */
+private const val RENDER_WRAP_SLACK_PX = 2
+
+internal class WrapWidthTextView(context: Context) : ReactTextView(context) {
+  override fun getCompoundPaddingRight(): Int =
+      super.getCompoundPaddingRight() - RENDER_WRAP_SLACK_PX
+}
+
+internal class WrapWidthTextViewManager : ReactTextViewManager() {
+  override fun createViewInstance(context: ThemedReactContext): ReactTextView =
+      WrapWidthTextView(context)
+}
+
+/**
+ * Must be registered after the autolinked packages so that its "RCTText"
+ * view manager replaces the stock one (last registration wins).
+ */
+class WrapWidthTextPackage : ReactPackage {
+  override fun createNativeModules(
+      reactContext: ReactApplicationContext
+  ): List<NativeModule> = emptyList()
+
+  override fun createViewManagers(
+      reactContext: ReactApplicationContext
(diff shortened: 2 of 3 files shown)


mozzius:
Wow, thanks for the deep research. We're soon about to upgrade to the new architecture, does the same issue happen there?

diabler-ie:
Good timing --- I checked this in the RN 0.81 source.

**The new architecture will fix it, but only with a feature flag enabled:** on Fabric, text still goes through the same two steps by default --- Fabric measures the text, and `ReactTextView` re-wraps it at render time --- so the same disagreement can still happen there (there are reports of this bug from new-architecture apps too). The fix is the `enablePreparedTextLayout` flag: with it, Android draws the exact layout that measurement produced (`PreparedLayoutTextViewManager`), so the two steps cannot disagree by design. In RN 0.81 the flag exists but is off by default ([MainReactPackage.kt#L152](https://github.com/facebook/react-native/blob/v0.81.5/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/shell/MainReactPackage.kt#L152)). I have not tested that path myself, and I do not know how battle-tested it is, so it is worth validating during your migration.

Since you're moving to the new architecture, up to you if this is worth merging --- I pushed a guard so the workaround steps aside automatically when `enablePreparedTextLayout` is on, so enabling the flag can never be blocked or broken by this PR, and deleting the plugin afterwards is a one-line cleanup --- but if you're migrating soon then I'm equally happy with you skipping this PR to go straight to the new architecture + the flag, accepting the bug until then (it's patched on my device already anyway!)

Thanks!

arturo32:
In "Build version: 1.130.0.1493" I cannot reproduce the issue anymore in my android device!!!

diabler-ie:
This is also fixed for me! 🥳

PR #11128, opened 2026-07-10
Add context comments to "From" strings for translation clarity

quiple:
This PR adds translation context comments to the `From` strings in the Advanced Search dialog and Notification Settings.

The word `From` has completely different meanings and grammatical forms depending on the context in target languages. Providing these comments ensures translators can provide accurate localized strings.

### DIFF
--- a/src/screens/Search/components/AdvancedSearchDialog/index.tsx
+++ b/src/screens/Search/components/AdvancedSearchDialog/index.tsx
@@ -403,7 +403,7 @@ function DialogInner({
                 t.atoms.text_contrast_medium,
                 a.mb_sm,
               ]}>
-              <Trans comment="Filter search results by a specific post author">
+              <Trans context="Filter search results by a specific post author">
                 From
               </Trans>
             </Text>
--- a/src/screens/Search/components/AdvancedSearchDialog/index.tsx
+++ b/src/screens/Search/components/AdvancedSearchDialog/index.tsx
@@ -403,7 +403,9 @@ function DialogInner({
                 t.atoms.text_contrast_medium,
                 a.mb_sm,
               ]}>
-              <Trans>From</Trans>
+              <Trans comment="Filter search results by a specific post author">
+                From
+              </Trans>
             </Text>
             <View style={[a.flex_row]}>
               <FollowingDropdown value={following} onChange={setFollowing} />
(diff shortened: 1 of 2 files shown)


tkusano:
I also agree with this PR.

In English, the word “From” appears before a word indicating a source or provenance. However, when translated literally into Japanese, it becomes 「から」, a case particle that is placed after a noun. If 「から」 appears on its own before another word, it becomes incomprehensible to Japanese speakers.

Until now, this label had only been used on the notification settings screen, where “From” was translated into the more specific phrase 「通知対象」 (“notification targets”). As a result, it did not seem unnatural to Japanese speakers.

However, that phrase is not appropriate on the search settings screen, so its use there is confusing for Japanese speakers. This could be avoided by translating it as something like 「検索対象」 (“search targets” or “search scope”).

For that reason, I think adding a `comment` is a very good solution.

mozzius (approved):

src/screens/Search/components/AdvancedSearchDialog/index.tsx
```diff
@@ -403,7 +403,9 @@ function DialogInner({
                 t.atoms.text_contrast_medium,
                 a.mb_sm,
               ]}>
-              <Trans>From</Trans>
+              <Trans comment="Filter search results by a specific post author">
```
surfdude29:
```suggestion
              <Trans context="Filter search results by a specific post author">
```

To get Lingui to generate separate catalog entries for the same source string used in two different contexts, we need `context` rather than `comment` here :)

`context` is folded into the generated message ID, so the two usages become two distinct entries that can be translated independently on Crowdin. A `comment` doesn’t affect the ID – it’s just a translator note on the (still-shared) entry, so both usages would collapse to one string and one translation.

src/screens/Settings/NotificationSettings/components/PreferenceControls.tsx
```diff
@@ -156,7 +156,9 @@ export function Inner({
         <>
           <Divider />
           <Text style={[a.font_semi_bold, a.text_md]}>
-            <Trans>From</Trans>
+            <Trans comment="Filter who you receive notifications from">
```
surfdude29:
```suggestion
            <Trans context="Filter who you receive notifications from">
```

mozzius (approved):

estrattonbailey:
@ds-boyce you may need to revise this after #11126 lands

ds-boyce:
These changes were included in #11126

PR #11133, opened 2026-07-10
Split mobile build/submit pipelines and de-duplicate OTA workflow

mozzius:
## Why

The three mobile CI entrypoints had grown into monolithic jobs:

- **iOS**: one job on `macos-26-xlarge` did build → `eas submit` → dSYM upload → fastlane TestFlight-group assignment. The fastlane step polls Apple processing for a long time, idling the most expensive runner in the fleet.
- **Android**: one job on `Linux-x64-32core` did build → submit → bundletool universal APK → artifact upload → Slack — everything after the build only needs `build.aab` + secrets.
- **OTA** (`bundle-deploy-eas-update.yml`): contained verbatim copy-pastes of both build jobs ("GitHub actions are horrible so let's just copy paste this in"), which had already drifted from the originals.

### New composite actions (`.github/actions/`)
- `setup-expo-project` — pnpm/node/jq/install/EAS CLI prologue (was duplicated 4×)
- `write-env` — writes `.env` + `google-services.json`, outputs `release-version`/`bundle-identifier`
- `compile-i18n` — `intl:build` + the "invalid syntax" error gate

### iOS: 1 job → 3
- `build` (`macos-26-xlarge`, overridable via new `runner` input): EAS build, read build number from the IPA (PlistBuddy is macOS-only), upload IPA+dSYM as a 1-day handoff artifact
- `submit` (`macos-26`): `eas submit`, dSYM Sentry upload, Slack notify, testflight-commit cache
- `distribute` (`macos-26`): fastlane TestFlight-group assignment — the long Apple-processing poll now idles a normal-size runner instead of the xlarge

### Android: 2 jobs → 4
- `build` (`Linux-x64-32core`, overridable via `runner` input): EAS build, upload `build.aab` as a 1-day handoff artifact
- `submit` (`ubuntu-latest`): Play Store submission — skippable via a new `submit` boolean input, so a build can be dispatched that only produces the APK artifact
- `universalApk` (`ubuntu-latest`): bundletool signed universal APK + artifact + Slack, running in parallel with `submit`
- `attachToRelease`: unchanged, now consumes the APK from `universalApk`

### OTA de-dup (487 → 233 lines)
Both copy-pasted `buildIfNecessary*` jobs are replaced with `uses:` calls to the reusable workflows (passing the smaller runners the OTA path already used), with explicit secret passthrough matching `nightly-build.yml`'s convention. `bundleDeploy` also adopts the `write-env` and `compile-i18n` actions.

`nightly-build.yml` is untouched — workflow outputs still map to the `build` jobs.

## Intentional behavior changes

1. **Android APK generation no longer blocked by Play submission**: `universalApk` runs in parallel with `submit`, so QA gets its APK even if the Play upload fails.
2. **OTA builds regain the i18n error grep** (dropped in the copy-pastes).
3. **OTA Android testflight-commit cache actually records now**: the copied condition (`channel != 'testflight' && != 'production'`) was never true. Both platforms now record on their testflight profiles via the reusable workflows.
4. **OTA Android APK artifact compression 0 → 6** (matches the standalone workflow).
5. **New handoff artifacts** (`ios-build-*`, `android-aab-*`, 1-day retention) are downloadable by anyone with repo read, like the existing APK artifact. The IPA is store-signed/encrypted.

## Verification

- `zizmor --min-confidence medium` clean across all workflows + new actions (matches CI config)
- `actionlint`: only pre-existing info-level SC2086 notes (same class as on main)
- Manual end-to-end testing to follow before marking ready:
  ```
  gh workflow run build-submit-android.yml --ref samuel/split-build-pipelines -f profile=testflight-android -f submit=false
  gh workflow run build-submit-ios.yml --ref samuel/split-build-pipelines -f profile=testflight -f testFlightGroup=none
  ```

### DIFF
--- a/.github/actions/setup-expo-project/action.yml
+++ b/.github/actions/setup-expo-project/action.yml
@@ -1,0 +1,47 @@
+---
+name: Setup Expo Project
+description: Install dependencies and set up the Expo/EAS CLI for a build. Does not check out the repo.
+
+inputs:
+  expo-token:
+    description: Expo token (EXPO_TOKEN secret)
+    required: true
+  eas-version:
+    description: EAS CLI version to install
+    required: false
+    default: '19.0.5'
+
+runs:
+  using: composite
+  steps:
+    - name: Check for EXPO_TOKEN
+      shell: bash
+      env:
+        EXPO_TOKEN: ${{ inputs.expo-token }}
+      run: >
+        if [ -z "$EXPO_TOKEN" ]; then
+          echo "You must provide an EXPO_TOKEN secret linked to this project's Expo account in this repo's secrets. Learn more: https://docs.expo.dev/eas-update/github-actions"
+          exit 1
+        fi
+
+    - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
+
+    - name: 🔧 Setup Node
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version-file: package.json
+        cache: pnpm
+
+    - name: 🪛 Setup jq
+      uses: dcarbone/install-jq-action@4fcb5062d7ce9bc4382d1a352d19ba3ba2c317c1 # v4.0.1
+
+    - name: ⚙️ Install dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
+
+    - name: 🔨 Setup Expo CLI
+      uses: expo/expo-github-action@eab7a230208c952974db8c3245cfd78402c7b385 # 9.0.0
+      with:
+        eas-version: ${{ inputs.eas-version }}
+        packager: 'pnpm --allow-build=dtrace-provider'
+        token: ${{ inputs.expo-token }}
--- a/.github/workflows/build-submit-android.yml
+++ b/.github/workflows/build-submit-android.yml
@@ -10,12 +10,25 @@ on:
         options:
           - testflight-android
           - production
+      submit:
+        type: boolean
+        description: Submit the build to Google Play (disable to only produce the APK artifact)
+        default: true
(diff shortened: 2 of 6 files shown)


.github/actions/setup-expo-project/action.yml
```diff
@@ -0,0 +1,47 @@
...
+      shell: bash
+      env:
+        EXPO_TOKEN: ${{ inputs.expo-token }}
+      run: >
+        if [ -z "$EXPO_TOKEN" ]; then
+          echo "You must provide an EXPO_TOKEN secret linked to this project's Expo account in this repo's secrets. Learn more: https://docs.expo.dev/eas-update/github-actions"
+          exit 1
+        fi
+
+    - uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
```
vineyardbovines, line 27: interesting that this can come before node lol

.github/workflows/build-submit-android.yml
```diff
@@ -56,84 +69,51 @@ permissions:
 jobs:
   build:
     if: github.repository == 'bluesky-social/social-app'
-    name: Build and Submit Android
-    runs-on: Linux-x64-32core
+    name: Build Android
+    runs-on: ${{ inputs.runner || 'Linux-x64-32core' }}
```
vineyardbovines, line 73: [just a question] why do we need this machine vs. just plain ubuntu?

mozzius: It's a custom runner that's actually good and not running on a potato. cuts build times down massively

.github/actions/setup-expo-project/action.yml
mozzius: yeah lol, seems pnpm is positioning itself as a meta-tool that sits above the runtime

vineyardbovines (approved):

PR #11144, opened 2026-07-11
Implement Bluvy Messenger Badge like GermDM

Thomasfds:
Add Bluvy Messenger Badge inspired from GermDM badge on profil.

Displayed from a custom lexicon created in Bluvy Messenger client. When we click on badge that send user to Bluvy for open a conversation with the user selected. After login in Bluvy the user is redirect on the conversation system directly.

We can't do like GermDM. We need one another api call for check thé lexicon and the good url.

Lexicon have visibility data (public, mutual, nothing) that this parameter who allow us to displayed the badge or not.


Thomasfds:
https://github.com/user-attachments/assets/9759e546-26e3-4d3b-bd6d-2585df0f51f0

Video about how the button work.

PR #11146, opened 2026-07-12
Add moduleSuffixes to tsconfig

mozzius:
## Per-platform typechecking via `moduleSuffixes`

The typechecker previously resolved every import to the base (non-suffixed) file, so platform-specific variants (`.ios`, `.android`, `.native`, `.web`) were only ever compiled standalone — never checked against the code that imports them. Type drift between a platform variant and its consumers was invisible until runtime.

This PR splits typechecking into three passes, each using `moduleSuffixes` to mirror the corresponding bundler's real resolution order:

| Pass | `moduleSuffixes` | Program |
| --- | --- | --- |
| `typecheck:ios` | `[".ios", ".native", ""]` | whole codebase |
| `typecheck:android` | `[".android", ".native", ""]` | whole codebase |
| `typecheck:web` | `[".web", ""]` | the web module graph, rooted at `index.web.ts` + all `.web.*` files |

`pnpm typecheck` runs all three; CI runs them as separate matrix jobs. The web pass is graph-rooted so that native-only implementations (which webpack never bundles) aren't checked against web-resolved imports — a combination that never ships. Every file is still covered: the iOS/Android passes include everything.

### Real bugs this caught

- **Android referrer tracking was broken**: `Referrer.getReferrerInfo()` is async on Android (the Play referrer API only exposes a promise) but was sync on iOS/web, and shared callers used it synchronously — so on Android the `deepLink:referrerReceived` metric fired with a Promise instead of data. The API is now uniformly promise-returning.
- **Drawer highlighting on web**: the web `useNavigationTabState` was missing `isAtFeeds`/`isAtBookmarks`, so those drawer items never showed as active.

### Drift fixes

The rest of the diff is aligning platform variants with the shared contracts their consumers are written against, e.g.:

- Web `List` exported `ListRef = RefObject<View>` even though the ref exposes scroll methods; web `Pager`/`TabBar`/`PagerWithHeader`/`Dialog`/`Menu`/`ContextMenu` were missing props/exports their native twins have (accepted-and-ignored where native-only, documented inline).
- Platform stubs that existed only to throw (`SubtitleFilePicker.native`, `VideoEmbedInnerWeb.native`, `FindContactsFlow.web`, …) had untyped signatures; they now carry real ones. Prop types that were duplicated across variants moved into sibling `.shared.ts` files so they can't drift.
- A handful of genuine call-site type errors (keyExtractors typed for the wrong item type, etc.).

### Package type quirks

- **react-native-svg**: its published `.web.d.ts` files describe a different API than the native declarations (no `SvgProps`/`PathProps`, react-native-web style types) even though the web runtime accepts native-typed props. The existing pnpm patch now deletes those two files so resolution falls through to the native declarations. To be raised with Software Mansion.
- **expo-file-system / expo-image-manipulator / device-attest**: similar divergences are handled by per-pass ambient shims in `src/platform/*.{web,android}-check.d.ts` (type-only, excluded from the passes they aren't written for).

### Runtime impact

Near zero by design: stubs still throw/no-op, ignored props are ignored. The exceptions: the `getReferrerInfo` fix above, web `Pager` now honors `onTabPressed`, and `BottomSheetNativeComponent.dismissAll()` is a safe no-op on web instead of a crash.


estrattonbailey (approved):
Rad. Will pry wanna merge main and typecheck before merging

PR #7897, opened 2025-03-04
A few good metrics

estrattonbailey:
This PR largely follows the migration steps from [this document](https://github.com/expo/fyi/blob/main/sentry-expo-migration.md).

A few key changes:
- Sentry's `release` value is now consistent across platforms e.g. `1.99.0`
- Sentry's `dist` value is now consistent across platforms, using the full commit hash
- I made a new Sentry project so that we get a clean break from the old

Sentry sourcemaps are now uploaded two separate ways:
- For native builds, via the Expo plugin
- For web builds, via the Webpack plugin
- For OTA builds, via the internal `@sentry/react-native` script ([source](https://github.com/getsentry/sentry-react-native/pull/3491)) and a new patch on that package

**TODO**
- [x] Maybe clean up env var injection
- [x] Remove debug logs
- [x] Add to OTA workflow
- [x] Remove old source map script

**Future work**
There's some work we could do to clean up env vars, but what I have here is clear imo (if verbose), so gonna leave as is for now.

**Additional notes on Sentry**
Source map script can be used like this: https://github.com/getsentry/sentry-react-native/pull/3491

Need to generate source maps for OTA. I got the following error when testing locally:
```
Uploaded 0 of 2 bundles and sourcemaps. Ensure you are running `expo export` with the `--dump-sourcemap` flag.
```

PR #11147, opened 2026-07-13
Update Sentry to 8.x

mozzius:
We were a bit behind with the Sentry SDK. Updating it let us drop a patch and adopt some cool new sentry features.

- Removing web replay from the bundle (we already do this in Webpack) https://docs.sentry.io/platforms/react-native/manual-setup/metro/#reduce-bundle-size 
- Capturing component names for breakcrumbs https://docs.sentry.io/platforms/react-native/integrations/component-names

I also cleaned up the metro config and switched it to a `.ts` file, fixing all the type errors

## Sourcemap upload changes

Dropping the patch is a deliberate behavior change on the OTA path, not just cleanup. The patch injected `--dist $SENTRY_DIST` into the sourcemap upload so artifacts matched the runtime's `dist` (the commit SHA) under Sentry's legacy release/dist matching. That's no longer needed: the pipeline now uses [debug IDs](https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/debug-ids/) end-to-end (the Metro serializer stamps a `debugId` into bundle + map, the upload passes `--debug-id-reference`, the runtime attaches `event.debug_meta`), and Symbolicator matches on debug ID first – release/dist is only the legacy fallback. This matches [Sentry's recommended Expo flow](https://docs.sentry.io/platforms/react-native/sourcemaps/uploading/expo/), which uploads with no dist at all.

Accordingly, this PR also:

- Calls `@sentry/expo-upload-sourcemaps` directly (now an explicit devDependency, pinned in lockstep with `@sentry/react-native`) instead of the deprecated `sentry-expo-upload-sourcemaps` stub. Org/project/url resolve from the plugin config in `app.config.js`.
- Drops the now-dead `SENTRY_DIST` from the OTA "Create Bundle" steps in `bundle-deploy-eas-update.yml` and `pull-request-comment.yml`. `SENTRY_RELEASE` is kept so artifacts stay browsable under the release in the Sentry UI (weak association; doesn't affect matching).
- Intentionally **keeps** `SENTRY_DIST` everywhere it's still read: the native `eas build` steps (the Sentry Xcode build phase and Gradle plugin consume it) and the web build (Dockerfile → `@sentry/webpack-plugin`).
- Rewrites the stale "Sentry sourcemaps" section in `docs/build.md` (it referenced a `postPublish` hook that no longer exists).

## Test plan

- confirm it still runs
- read the migration guides
  - https://docs.sentry.io/platforms/react-native/migration/v6-to-v7/
  - https://docs.sentry.io/platforms/react-native/migration/v7-to-v8/
  - https://docs.sentry.io/platforms/javascript/guides/react/migration/v8-to-v9/
  - https://docs.sentry.io/platforms/javascript/guides/react/migration/v9-to-v10/
- verified locally: `expo export --dump-sourcemap` produces a bundle + map sharing the same `debugId`, and the upload CLI resolves org/project from `app.config.js` and reaches the upload step (401 with a dummy token, as expected)
- after the first OTA deploy off this branch: confirm a JS error symbolicates and the event shows a **debug ID** match (not release/dist)

## Related work
https://github.com/bluesky-social/social-app/pull/7897


docs/build.md
```diff
@@ -169,27 +169,13 @@ See [testing.md](./testing.md).
...
-  `node_modules/@sentry/cli/bin/sentry-cli releases \
-files <release name> \
-upload-sourcemaps \
---dist <iOS Update ID> \
---rewrite \
-dist/bundles/main.jsbundle dist/bundles/ios-<hash>.map`
+Sourcemaps upload automatically as part of each build path; there are no manual steps.
+
+- Native builds (`eas build` for iOS/Android): the `@sentry/react-native` Xcode build phase and Gradle plugin upload sourcemaps during the build. `SENTRY_AUTH_TOKEN`, `SENTRY_RELEASE`, and `SENTRY_DIST` are provided by the build workflows and read natively.
+- OTA updates: `pnpm export` runs `expo export --dump-sourcemap` and then `pnpm upload-native-sourcemaps`, which invokes `@sentry/expo-upload-sourcemaps` on the `dist/` folder. The org and project are read from the Sentry plugin config in `app.config.js`; `SENTRY_AUTH_TOKEN` is required and `SENTRY_RELEASE` is optional (it only lets you browse the artifacts under a release in the Sentry UI).
```
estrattonbailey: So we set a `dist` tag on builds but not on OTA? The `dist` is just the hash, which can be useful for finding the commit that introduced a bug. I think `release` is also important, nice to see if a release is crashy vs the previous.

estrattonbailey (approved):
All seems fine but the dist handling seems off to me rn. Might as well keep it around for commit hashes imo.

Also, I'd recommend a button to trigger an error report so we can test this once it goes out.

PR #11149, opened 2026-07-14
Add DM button to labeller profiles

RetroSunstar:
Currently users have to navigate to the Chat tab and initiate conversations with labellers from there, unlike regular users who have a dedicated DM button on the profile itself

This is unintuitive, inconsistent, and often needs to be explained by labellers themselves if they use DMs in any way

It doesn't make a ton of sense to not include it, as if a labeller didn't want DMs then this would be disabled on the account level. Logically, at no point should the app itself attempt to hide away the functionality, as it hinders the labellers that do use them 🐙️


estrattonbailey (approved):
Nice nice

PR #11150, opened 2026-07-14
Add TIDA form URL for adult content report

foysalit:
This PR adds an interstitial to link out to TIDA intake form when `Adult content` reporting category is selected, in order to nudge users to a better way of reporting NCII content.

https://github.com/user-attachments/assets/eb75adc5-d7be-4d88-9a29-b3aa70cbde61


src/components/moderation/ReportDialog/index.tsx
```diff
@@ -780,6 +792,124 @@ function OptionCard({
...
+                a.px_md,
+                a.py_sm,
+                a.rounded_sm,
+                a.border,
+                hovered || pressed
+                  ? [t.atoms.border_contrast_high]
+                  : [t.atoms.border_contrast_low],
+              ]}>
+              <Text style={[a.flex_1, a.italic, a.leading_snug]}>
+                <Trans>
```
surfdude29: nit: As the NCII form is currently only available in English, I wonder if it might be good to add the `english-only-resource` `context` (#8805) to this string, to highlight this to translators so they can mention it if they feel it's helpful? Or perhaps that might not be appropriate in this instance? 🤔

src/components/moderation/ReportDialog/index.tsx
```diff
@@ -780,6 +792,124 @@ function OptionCard({
...
+            type: 'answerNciiQuestion',
+            question: 'isDepicted',
+            answer,
+          })
+        }}
+      />
+      {outcome === 'externalForm' && (
+        <Link
+          to={NCII_FORM}
+          label={l`Submit your report through the Report non-consensual intimate imagery (NCII) form`}>
```
surfdude29: nit: ditto here for the accessibility label

estrattonbailey (approved):
Thanks for tackling this!

PR #11151, opened 2026-07-14
fix web feed footer clearing the bottom bar

vineyardbovines:
## Summary
- Adds `useBottomBarOffset()` to `PostFeed`'s `FeedFooter` spacer so the last post is no longer covered by the fixed mobile-web bottom bar
- No-op on desktop web (hook returns 0 above the 800px breakpoint) and on native (still uses the doubled `headerOffset` path)

## Test plan

hard to get a screenshot of it but:

- [ ] Mobile web (<800px): scroll to end of Following/Discover; last post fully visible
- [ ] Desktop web (≥800px): no visual change at feed end
- [ ] Native iOS/Android: no visual change at feed end


mozzius (approved):
This sucks, we should fix it at the root eventually (set scrollview content insets properly to account for the tab bar). This is fine for now

PR #11160, opened 2026-07-15
Display beta badge when isBetaUser is true

ds-boyce:
Only displays for your own profile to indicate you’ve opted in to beta features.

beta

### DIFF
--- a/src/components/ProfileBadges.tsx
+++ b/src/components/ProfileBadges.tsx
@@ -29,19 +29,19 @@ const botIconSizes: Record<Size, number> = {
 } as const
 
 const betaIconSizes: Record<Size, number> = {
-  xs: 10,
-  sm: 11,
-  md: 12,
-  lg: 16,
-  xl: 20,
+  xs: 8,
+  sm: 8,
+  md: 8,
+  lg: 10,
+  xl: 12,
 } as const
 
 const betaBadgePadding: Record<Size, number> = {
-  xs: 2,
-  sm: 3,
-  md: 4,
-  lg: 6,
-  xl: 8,
+  xs: 1,
+  sm: 2,
+  md: 3,
+  lg: 4,
+  xl: 5,
 } as const
 
 export function ProfileBadges({
--- a/src/components/ProfileBadges.tsx
+++ b/src/components/ProfileBadges.tsx
@@ -88,30 +88,30 @@ export function ProfileBadges({
       ]}>
       {interactive ? (
         <>
+          <VerificationCheckButton
+            profile={shadowed}
+            width={verificationIconWidth}
+          />
           <BetaBadgeButton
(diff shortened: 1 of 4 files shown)


ds-boyce:
> `BetaBadgeButton` tooltip likely won't open on web (double-toggle) — non-blocking, please verify on web

Works on web:

https://github.com/user-attachments/assets/cb70ae57-45d7-402c-a7ba-4ef5b39d492f

src/components/ProfileBadges.tsx
```diff
@@ -68,6 +88,11 @@ export function ProfileBadges({
       ]}>
       {interactive ? (
         <>
+          <BetaBadgeButton
+            profile={shadowed}
+            width={betaIconWidth}
+            padding={betaBadgeScaledPadding}
+          />
```
estrattonbailey, line 120: Does the padding match the VerificationCheckButton? Screenshot looks a little off but might just be like it is

ds-boyce: `VerifierCheck` is an icon, but the beta badge is an icon with a container. The padding around the icon is a single value and matches the Figma, so it should be centered in the circle/disc. What looks off?

estrattonbailey: CleanShot 2026-07-15 at 17 10 10

They're just not the same size. We should just match the size of the Verification icon, and there are multiple sizes of it depending on where the profile appears in the UI.

ds-boyce: Ah, right. I was following the Figma here but I’ll loop back.

estrattonbailey:
For some reason the alignment is off on web now, not sure why from the changes on this PR 🤔 also might be nice to match the hover on the verification check on web.

Also, I kinda think we should reverse the order of these, check first.

CleanShot 2026-07-20 at 10 46 24

ds-boyce:
Alignment is technically pixel-perfect as far as I can tell, the effect seems to be purely visual. Not sure if we want to compensate for it, or maybe tweak the colors so it’s less pronounced.

alignment

I’ll add the hover and switch the order.

ds-boyce:
https://github.com/user-attachments/assets/ea6951fa-33ce-4ab9-b4cf-e1264af40aa4

ds-boyce:
I’ll add the missing transition and hitslop.

estrattonbailey (approved):
Nice! Looks good to me now, maybe was a me-problem.

Last thing, the hitslop on these badges, esp bot badge, is preventing tapping on each cleanly on native. If there are multiple, we should adjust the x-axis hitslop between the badges so they don't overlap. Badges at the beginning or end can keep a wider hitslop though.

ds-boyce:
| Before | After |
| - | - |
| before | after |

PR #11164, opened 2026-07-15
Lighten blue (link) text for dim and dark themes

ds-boyce:
Increases contrast for better legibility. Covers post links, handles, and hashtags in rich text components, e.g., posts, messages, and bios.

### DIFF
--- a/oxlint-suppressions.json
+++ b/oxlint-suppressions.json
@@ -1617,14 +1617,6 @@
       "count": 1
     }
   },
-  "src/view/com/composer/text-input/TextInput.tsx": {
-    "typescript/no-floating-promises": {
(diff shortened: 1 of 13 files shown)


mozzius:
I think the inline Link component also sets itself as blue, need to change it there too. And there's almost certainly other random places

ds-boyce:
I guess the question is how extensive do we want to make this? There are other aspects of the UI that are still intended to use `primary_500`.

estrattonbailey:
I think we want a `t.atoms.text_link` atom and it should apply to all links, but we'll check with design on that

estrattonbailey (approved):
Nice

PR #11178, opened 2026-07-16
Add height constraint to full-height bottom sheets

ds-boyce:
This notably caused an issue where the edit profile screen was not scrollable for long bios.

https://github.com/user-attachments/assets/2a9af83b-5006-4445-ab4f-303b0ac59c28

With `flex: 1`, the sheet becomes scrollable as intended.

https://github.com/user-attachments/assets/2df0e896-b604-42c9-a079-88efc3b0e0de


estrattonbailey (approved):
And no change on iOS or for non-full height bottom sheets?

ds-boyce:
Yep, iOS and non-full height sheets both work as intended.

https://github.com/user-attachments/assets/8927f992-2260-4370-aa55-1e0c04b23962

https://github.com/user-attachments/assets/e02f0a6c-a70f-4b92-97df-ed3d3de82487

PR #11180, opened 2026-07-17
Fix isAppLabelers being counted towards max labeller counter

RetroSunstar:
Currently users can only subscribe to a max of 19 labellers

This is because the check used doesn't remove isAppLabeler declared labellers from the labelerCount like it does with invalidLabelers

So ``moderation.bsky.app`` will always reduce this maximum by 1, and any additional moderation authorities will reduce this number further 

Which seems unfair to the user when informing them of a 20 labeller limit but taking up those slots with entries outside of their control 🐙

### DIFF
--- a/src/state/queries/labeler.ts
+++ b/src/state/queries/labeler.ts
@@ -3,6 +3,7 @@ import {useMutation, useQuery, useQueryClient} from '@tanstack/react-query'
 import {z} from 'zod'
 
 import {MAX_LABELERS} from '#/lib/constants'
+import {isAppLabeler} from '#/lib/moderation'
 import {GCTIME, STALE} from '#/state/queries'
 import {
   preferencesQueryKey,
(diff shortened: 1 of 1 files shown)


mozzius:
No, this is limited on the backend too. There has to be a total limit of 20, _including_ the fixed ones

RetroSunstar:
> No, this is limited on the backend too. There has to be a total limit of 20, _including_ the fixed ones

If we include the fixed labellers, the max of 20 already falls apart when it comes to the additional services.

If you subscribe to 19 + the base labeller, than move to somewhere that requires an additional serivce: 21

If the app fails to obtain a country code: all, or 25 out of the dids that are actual labellers and not regular accounts

If a user subscribes to 20 labellers using an external platform that doesn't declare or count any base labelers, that's 21-22

There is a failure at a certain stage when you have all the moderation authorities applied, but this issue seems to already exist and the addition of 1 labeller doesn't have a major impact in that specifically

PR #11182, opened 2026-07-17
Migrate from `@atproto/api` to `@bsky/sdk`

### LINKED ISSUE #11195: Spontaneous logouts: session-expiry handling destroys valid credentials and can be resurrected by cross-tab races
## Summary

Users have reported random logouts since at least early 2024 (tracked internally in the "Logout issues" Linear project). After tracing the session layer, the PDS token semantics, and Sentry data, we believe the dominant cause is not tokens genuinely expiring - it is the client's expiry handling amplifying one stale tab (or one stale persisted write) into a full logout that destroys valid credentials. Sentry shows this at enormous scale: the `session: resume failed` / `choose account: initSession failed` / `session: resumeSession failed` / `Token has expired` cluster accounts for ~5.9M events and >900k affected users in the current retention window, with breadcrumbs showing `refreshSession` returning 400 within ~1s of cold start - i.e. the *stored* token was already dead before the user did anything.

## Background: token semantics

- Refresh tokens are single-use but rotation grants the used token a **2-hour grace period** with a stable successor id (`account-manager.ts` in the PDS). Concurrent refreshes within 2h converge; a token **more than 2h behind** the current generation gets `ExpiredToken`.
- The app **never calls `com.atproto.server.deleteSession`** - logout is purely local token clearing. So nothing the app does revokes tokens server-side; a token can only die by falling >2h behind.

## Root cause 1: expiry handling destroys newer valid tokens (all platforms, worst on web)

When any session's refresh fails with `ExpiredToken`, the `'expired'` event flows to the reducer (`src/state/session/reducer.ts` `'received-agent-event'`), which clears the account's `accessJwt`/`refreshJwt`, switches to the logged-out state, and persists + broadcasts. The write **wholesale-replaces** the persisted session blob.

The problem: the dying session's token is not necessarily the latest generation. Ways a client legitimately ends up holding a >2h-stale token while a *newer valid generation exists*:

- **Frozen/discarded background tabs** (Chrome Memory Saver etc.): the tab misses hours of rotations; on unfreeze, an interval (e.g. the chat event-bus poll) or queued fetch can fire with the ancient token *before* the queued BroadcastChannel messages are processed.
- **Failed localStorage writes** (private mode, quota): `writeToStorage` swallows the error while the broadcast still fires - tabs silently diverge.
- **Old Safari (<15.4)**: the BroadcastChannel shim is a stub; there is no cross-tab sync at all.
- **Native**: persistence is async (`AsyncStorage.setItem`); a rotation that commits server-side but dies before the write flushes (app killed, crash, OOM) leaves a stale token on disk. Next cold start >2h later fails resume.

In all of these, the stale client's `ExpiredToken` triggers a token clear that **overwrites the healthy generation in storage** and (on web) broadcasts logout to every tab. One zombie tab logs out the whole browser and destroys credentials that were perfectly valid - they were never revoked, just deleted locally.

This also explains the observed follow-on: after a failed resume, the account entry retains the dead tokens (or none), so re-selecting the account from the switcher fails again and the user is forced to re-enter their password.

### Proposed fix

Compare-and-clear: when handling `'expired'`, compare the dying session's `refreshJwt` against the latest persisted state for that DID. If storage holds a *different* (newer) generation, do not clear or log out - rebuild the session from the newer tokens instead (with a loop guard so repeated failures still fall through to a real logout). Only when the dying token *is* the latest generation should the logout proceed.

## Root cause 2: logout doesn't stick on web (follower-tab resurrection)

In the cross-tab sync handler (`src/state/session/index.tsx`, `persisted.onUpdate('session')`), a follower tab that sees a different current DID runs `void resumeSession(syncedAccount)`. Nothing cancels this task when a subsequent broadcast logs the tab out. Sequence:

1. Tab A switches accounts -> broadcast -> tab B starts `resumeSession` (async, may refresh over the network).
2. Tab A logs out -> broadcast -> tab B goes logged-out.
3. Tab B's pending resume completes anyway, dispatches `'switched-to-account'` with freshly rotated tokens, persists, and broadcasts - **signing the whole browser back in.**

### Proposed fix

Cancel the pending one-task when a synced update logs the tab out or changes DID, and re-check the store after the resume resolves - bail and dispose the built session if the account is no longer the intended current account.

## Related weaknesses found during the same investigation

- **`'update'` events lack a bundle-identity check** in the reducer (the guard only covers the no-`refreshedAccount` path): a refresh completing in the window between logout and agent disposal writes fresh tokens back into a soft-logged-out account (the account picker silently regains one-tap access).
- **Forced logout never unregisters push tokens** - the unregister side effect exists only on explicit logout/remove paths, so a device that hit the expiry path keeps receiving pushes for the logged-out account.
- **No server-side revocation on logout**: soft logout leaves the refresh token valid for up to 90 days. Anything that captured it (extension, backup of localStorage) can resume the session. Worth calling `deleteSession` fire-and-forget after clearing local state.

## Sentry references

- APP-S37V `session: resume failed` (native)
- APP-S3VC `choose account: initSession failed` (mixed; breadcrumbs show resume-fail -> account-picker-fail with the same dead token)
- APP-S3X8 `session: resumeSession failed` (web only)
- APP-SDAV `Error: Token has expired`
- Smaller corroborating groups: APP-SPVQ/APP-SQ2C `Token has been revoked`, APP-STPE `Concurrent session update detected`, APP-S3X7 `networkRetry failed to resume session`

mozzius:
Migrates the app from `@atproto/api` to **`@bsky/sdk`** + `@atproto/lex`. The `@atproto/api` dependency is **fully removed** from the root package.

## What changed

**Session layer** (`src/state/session/`)
- `PasswordSession` (from `@atproto/lex-password-session`) is the single auth core; the old `AtpAgent`/`SessionAgent` bridge is deleted.
- `SessionBundle = {session, accountClient, appviewClient, chatClient, service}` — three lex `Client`s per account plus a dedicated chat client (`did:web:api.bsky.chat#bsky_chat` proxy; `DM_SERVICE_HEADERS` and its 36 call sites are gone).
- Hook contracts: `usePdsClient()`/`useChatClient()` **throw a typed `NotAuthenticatedError` when logged out** rather than silently falling back to the public appview (writes must never hit public infra); read hooks (`useAppviewClient()`/`useLexClient()`) keep the public fallback. `useMaybe*` variants return `null` for logged-out-aware callers.
- `useSessionApi().refreshSession()` replaces `agent.resumeSession()`; it returns the refreshed `SessionAccount` snapshot so callers (e.g. SignupQueued) don't race the reducer.

**API calls**
- All call sites use `client.call(schema, params, options?)` with generated lexicon schemas from `#/lexicons` (vendored JSON in `lexicons/`, generated by `lex build` on postinstall). Sugar methods became SDK actions (`like`, `upsertProfile`, `overwriteSavedFeeds`, `upsertNux`, ...).
- Per-call proxying via the `{service}` option (push notifications, feed-feedback, labeler reports) instead of manual `atproto-proxy` headers — covered by a header-emission unit test.
- Six previously-missing lexicons vendored locally (`com.atproto.temp.checkHandleAvailability`/`checkSignupQueue`, the unspecced suggested-users family).

**Types & moderation**
- `#/types/bsky` is single-world (`#/lexicons`); the `toLex`/`dangerousIsType`/`validate` migration escape hatches are deleted.
- SDK moderation (`@bsky/sdk/moderation`) and richtext (`@bsky/sdk/richtext`) replace the transplanted `@atproto/api` versions.
- `lib/xrpc-error.ts` provides lex-native error helpers (`isXrpcError`/`getErrorStatus`/`getErrorName`/`getErrorHeader`).
- Report subjects go through typed builders (`ReportDialog/utils/reportSubject.ts`); chat message/convo refs are the one documented assertion (wire-accepted but not in the lexicon union).

**Client-side CID computation** extracted to `lib/api/computeCid.ts` with golden-CID regression fixtures (byte-identical throughout the migration — this gates the reply-chain CID path).

## Not in this PR
- `bskyembed/`, `bskyogcard/`, `bskylink/`, `dev-env/` — standalone packages with their own lockfiles and `@atproto/api` pins; they migrate independently.
- Runtime smoke testing (see warning above).

## Verification
- `pnpm typecheck` 0 errors, `pnpm test` 40/40 suites (606 passed + 28 todo, 23 snapshots), `pnpm lint` clean.
- `grep -rn "@atproto/api" src/ __tests__/ package.json pnpm-lock.yaml` → zero.
- Golden CIDs byte-identical; session reducer scenarios (background-account expiry, network-error transience, arm-latch event mapping) pass unchanged.


mozzius:
@claude please review

mozzius:
@claude another round of review please

src/components/dms/LeaveConvoPrompt.tsx
```diff
@@ -36,11 +36,9 @@ export function LeaveConvoPrompt({
...
       if (isNetworkError(error)) {
         errorMessage = l`A network error occurred. Please check your internet connection.`
-      } else if (error instanceof ChatBskyConvoLeaveConvo.InvalidConvoError) {
+      } else if (getErrorName(error) === 'InvalidConvo') {
         errorMessage = l`Conversation not found.`
-      } else if (
-        error instanceof ChatBskyConvoLeaveConvo.OwnerCannotLeaveError
-      ) {
+      } else if (getErrorName(error) === 'OwnerCannotLeave') {
         errorMessage = l`Owner must lock the group before leaving.`
```
mozzius: don't like how it downgraded these instanceofs to untyped string comparisons. fixing...

PR #11184, opened 2026-07-17
Add canonical OP thread numbering

estrattonbailey:
An OP thread is the oldest contiguous chain of replies starting from the root, all authored by OP.

| Android | iOS | Web |
| - | - | - |
| android | ios | web |

### DIFF
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -1,4 +1,4 @@
-import {type ReactNode, useMemo} from 'react'
+import {useMemo} from 'react'
 import {type StyleProp, type TextStyle} from 'react-native'
 import {AppBskyRichtextFacet, RichText as RichTextAPI} from '@atproto/api'
 
@@ -27,7 +27,7 @@ export type RichTextProps = TextStyleProp &
     interactiveStyle?: StyleProp<TextStyle>
     emojiMultiplier?: number
     shouldProxyLinks?: boolean
-    suffix?: ReactNode
+    suffix?: React.ReactNode
     /**
      * DANGEROUS: Disable facet lexicon validation
      *
--- a/src/components/RichText.tsx
+++ b/src/components/RichText.tsx
@@ -1,4 +1,4 @@
-import {useMemo} from 'react'
+import {type ReactNode, useMemo} from 'react'
 import {type StyleProp, type TextStyle} from 'react-native'
 import {AppBskyRichtextFacet, RichText as RichTextAPI} from '@atproto/api'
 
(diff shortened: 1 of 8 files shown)


src/screens/PostThread/components/ThreadItemPostNumber.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <Text
+        style={[
+          a.text_xs,
+          a.font_medium,
+          t.atoms.text_contrast_high,
+          {
+            fontVariant: ['tabular-nums'],
+          },
+        ]}>
+        {index}/{count}
```
surfdude29: nit: Could this be wrapped for translation please? There are occasionally languages where the reverse order for similar strings seems to be preferred, e.g. for the group chat members string in Hungarian:

Image

Also, maybe adding an `accessibilityLabel` would be helpful for screen reader users? I haven't tested so this may be wrong, but Claude said that otherwise it could be read out like this:

> VoiceOver and TalkBack will read “3/7” as something like “three slash seven” or possibly as a date, depending on the platform’s heuristics.

For the accessibility label, I'm no expert on these matters but my instinct is that a concise `Post ${index} of ${count}` might be best? Although if more precision were needed, `Post ${index} of ${count} in this thread` or `Post ${index} of ${count} in the author's thread` could work.

```
import {Trans, useLingui} from '@lingui/react/macro'

// ...

const t = useTheme()
const {t: l} = useLingui()

// ...

<Text
  accessibilityLabel={l({
    message: `Post ${index} of ${count}`,
    context: 'post-number-in-thread',
    comment:
      "Screen reader label for the badge showing a post's position in the original poster's thread — the badge itself displays the short form, e.g. '3/7'.",
  })}
  style={[a.text_xs, a.font_semi_bold, t.atoms.text_contrast_medium]}>
  <Trans
    context="post-number-in-thread"
    comment="Compact badge showing where this post sits in the original poster's thread — e.g. '3/7' is the third of seven posts. {index} is this post's position and {count} is the total.">
    {index}/{count}
  </Trans>
</Text>
```

src/components/RichText.tsx
```diff
@@ -27,6 +27,7 @@ export type RichTextProps = TextStyleProp &
     interactiveStyle?: StyleProp<TextStyle>
     emojiMultiplier?: number
     shouldProxyLinks?: boolean
+    suffix?: ReactNode
```
mozzius: nit: typically use the implicit `React.ReactNode`, although that is pretty weird now I come to think about it

mozzius (approved):
Awesome!

PR #11192, opened 2026-07-18
Move known likers below stats row; restore original order in stats row

abenzer:
- Restores the original order of items in the stats row on posts (they should mirror the interaction buttons below)
- Known likers row now wraps to up to two lines when needed
- Known likers row now has a smaller text size

Screenshot 2026-07-17 at 9 08 38 PM


estrattonbailey (approved):
Nice polish!

PR #11204, opened 2026-07-20
Don't roll up account labels into additional post alerts

estrattonbailey:
From a viewer standpoint, the logic is pretty simple. If you've configured labels to "Show" or "Warn", they'll appear in `PostAlerts` regardless of if they apply to your post or your account as a whole

From an author standpoint, it's similar but subtle. If you've configured labels to "Show" or "Warn", they'll appear in `PostAlerts` regardless of if they apply to your post or your account as a whole. However, labels that you've configured to be hidden will show _if they're applied to the post itself_, because we want to give you a path to appeal them. Labels applied to your account that are not configured to be visible do not show, since that appeal path remains accessible on your profile.

The `+n` pill only shows if there are some labels configured to be visible on the post, and `n` others configured to be hidden, after the logic above is applied.

### DIFF
--- a/src/components/moderation/PostAlerts.tsx
+++ b/src/components/moderation/PostAlerts.tsx
@@ -54,7 +54,15 @@ export function PostAlerts({
   const isOwnPost = !!post && post.author.did === currentAccount?.did
   const allLabels: ComAtprotoLabelDefs.Label[] =
     isOwnPost && view === 'expanded'
-      ? [...(post.labels ?? []), ...(post.author.labels ?? [])]
+      ? [
+          ...(post.labels ?? []),
+          /*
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
Will this also handle the like whole-account label applied by bluesky moderation which rolled out and then rolled back or nah?

estrattonbailey:
@ThisIsMissEm this doesn't change anything about how account labels are applied when making UI decisions like hiding/showing content. These are just badges on the posts, and this change is oriented towards the author side of things, making it more or less in line with where things were prior to our release last week when we made some small UI tweaks to these badges.

ThisIsMissEm:
@estrattonbailey ah, okay!

PR #11206, opened 2026-07-20
Update presentation of trending feeds

ds-boyce:
It doesn’t make much sense to have pins or likes for trending feeds since they’re ephemeral. The description is more important as well, particularly for breaking news.

| Feed | Trending feed |
| - | - |
| feed | trending |

### DIFF
--- a/src/screens/Profile/ProfileFeed/index.tsx
+++ b/src/screens/Profile/ProfileFeed/index.tsx
@@ -25,7 +25,6 @@ import {
   type UsePreferencesQueryResponse,
 } from '#/state/queries/preferences'
 import {useResolveUriQuery} from '#/state/queries/resolve-uri'
-import {useTrendingTopics} from '#/state/queries/trending/useTrendingTopics'
 import {truncateAndInvalidate} from '#/state/queries/util'
 import {useSession} from '#/state/session'
 import {PostFeed} from '#/view/com/posts/PostFeed'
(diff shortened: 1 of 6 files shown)


mozzius (changes requested):
Looking real good, but for these feeds the header becomes one long thin button. I think it would make more sense to drop the custom header for trending feeds and just use the standard header components? And just make the ... menu open the dialog

Image

src/screens/Profile/ProfileFeed/index.tsx
```diff
@@ -181,13 +183,34 @@ export function ProfileFeedScreenInner({
...
+
+        return (
+          params.rkey === feedInfo.route.params.rkey &&
+          (owner === feedInfo.creatorDid.toLowerCase() ||
+            owner === feedInfo.creatorHandle.toLowerCase())
+        )
+      }) ?? false,
+    [trending?.topics, feedInfo],
+  )
+
```
mozzius: Would it not be simpler to just hardcode this to be all feeds by @trending.bsky.app? Otherwise when it stops trending it will show the old header

ds-boyce: Yep! @estrattonbailey and I came to the same conclusion yesterday I just haven’t had a chance to update this yet.

src/screens/Profile/ProfileFeed/index.tsx
```diff
@@ -45,14 +44,17 @@ import {HashtagWide_Stroke1_Corner0_Rounded as HashtagWideIcon} from '#/componen
 import * as Layout from '#/components/Layout'
 import {IS_NATIVE} from '#/env'
 
+const TRENDING_DID = 'did:plc:qrz3lhbyuxbeilrc6nekdqme'
+const TRENDING_HANDLE = 'trending.bsky.app'
```
mozzius: nit: can move these to the constants file

mozzius (approved):

src/screens/Profile/components/ProfileFeedHeader.tsx
estrattonbailey, line 516: I think if this is from `trending.bsky.app`, we might as well ditch the link and just show static text `by Bluesky`

estrattonbailey (approved):
LGTM!

PR #11220, opened 2026-07-21
Fix WelcomeModal sign in button keyboard activation

### LINKED ISSUE #11214: Cannot press `WelcomeModal` sign in button using the keyboard
### Steps to Reproduce

1. Go to https://bsky.app/ as a guest
2. In the welcome modal that appears, press the tab key until the sign in button is focused
3. Attempt to activate this button using the space or enter key
4. Nothing happens

### Attachments

Image

Web (Desktop)

### Device Info

N/A

Build version: 1.128.0; Bundle info: 10aaa0b (prod); Bundle date: 26072023; Platform: web; Platform version: 0.0.0

### Additional Information

This issue occurs in [WelcomeModal.tsx](https://github.com/bluesky-social/social-app/blob/daa11b63e719780a16c9ecf399e9bdf1d25c18d0/src/components/WelcomeModal.tsx#L190-L208), where the `onPress` attribute for the sign in button has been incorrectly placed on the `Text` component rather than the `Pressable`.

I could have created a pull request to fix this myself, but I've declined to do so since I can see from other PRs that you use bot code review tools, which I greatly dislike interacting with.

I've confirmed on React Native's [example page](https://reactnative.dev/docs/pressable) for `Pressable` that if implemented correctly with the `onPress` attribute on the `Pressable` itself, keyboard support is partially correctly implemented. The enter key correctly activates the button, although the space key does not, which is likely a bug in React Native.

MonsterPi13:
## Summary

In the welcome modal, the "Sign in" link can be focused with Tab but pressing Enter does nothing.

The root cause is that the `onPress` handler was placed on the inner `Text` component rather than on the focusable `Pressable`. The `Pressable` (with `accessibilityRole=\"button\"`) is what receives keyboard focus, but it had no `onPress`, while the `Text` that had `onPress` is not keyboard-activatable. Moving `onPress` to the `Pressable` fixes Enter-key activation.

Note: as described in the issue, the space key still does not activate the button after this fix. That appears to be an upstream React Native behavior (space is not mapped for `Pressable`), so it is out of scope here — this PR restores Enter-key activation.

Fixes #11214

## Test plan

1. Open the app as a guest so the welcome modal appears
2. Press Tab until the "Sign in" link is focused
3. Press Enter -> the sign in flow is triggered

### DIFF
--- a/src/components/WelcomeModal.tsx
+++ b/src/components/WelcomeModal.tsx
@@ -188,6 +188,7 @@ export function WelcomeModal({control}: WelcomeModalProps) {
                     ]}>
                     <Trans>Already have an account?</Trans>{' '}
                     <Pressable
+                      onPress={onPressSignIn}
                       onPointerEnter={() => setSignInLinkHovered(true)}
                       onPointerLeave={() => setSignInLinkHovered(false)}
                       accessibilityRole="button"
@@ -201,8 +202,7 @@ export function WelcomeModal({control}: WelcomeModalProps) {
                             fontSize: undefined,
                           },
                           signInLinkHovered && a.underline,
-                        ]}
-                        onPress={onPressSignIn}>
+                        ]}>
                         <Trans>Sign in</Trans>
                       </Text>
                     </Pressable>


ds-boyce:
Ope, same fix in #11308

PR #11222, opened 2026-07-21
APP-2670: add multipart video upload transport

vineyardbovines:
## Summary

- add client-side multipart video upload using the first-party Lumi proxy contract
- chunk compressed videos with server-selected part sizes and upload with bounded concurrency, retry, progress, and real cancellation
- finalize uploads with state-aware recovery, missing-part re-upload, and safe abort-before-legacy fallback
- gate rollout with `video:multipart_upload:enable` while preserving the existing single-request transport
- surface asynchronous Lumi validation failures with specific user-facing messages


src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,248 @@
...
+            return await abortThenFallbackOrResolve(jobId, token, err)
+          }
+          createdFailures++
+          if (createdFailures < MULTIPART_FINISH_ATTEMPTS) {
+            await delay(500 * 2 ** (createdFailures - 1), signal)
+            continue
+          }
+          return await abortThenFallbackOrResolve(jobId, token, finishError)
+        case 'finishing':
+          // Finalization owns the reservation and may already have assembled
```
estrattonbailey: This first sentence is a doozy lol, what mean lol

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,248 @@
...
+  throw lastError
+}
+
+function isRetryableStatusError(err: unknown) {
+  return (
+    err instanceof TypeError ||
+    (err instanceof MultipartUploadError &&
+      (err.error === 'ServiceOverloaded' ||
+        err.status === undefined ||
+        err.status >= 500))
```
estrattonbailey: Any overlap with the common `RETRYABLE_ERRORS` we've used elsewhere? Do we need to handle 429?

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,248 @@
...
+  if (result.state === 'aborted') {
+    throw new MultipartFallbackError(
+      cause instanceof Error ? cause.message : 'Multipart upload failed',
+    )
+  }
+  if (result.state === 'completed' && result.completedJobId) {
+    const status = await getUploadStatus(jobId, token)
+    const completed = completedStatus(status)
+    if (completed) return completed.jobStatus
+  }
```
estrattonbailey, line 234: Interesting that we can abort but still return a completed job 🤔

Seems like basically if something throws, we abort, which is also a recovery mechanism?

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,248 @@
...
+
+function mintToken(agent: AtpAgent) {
+  return getServiceAuthToken({
+    agent,
+    lxm: 'com.atproto.repo.uploadBlob',
+    exp: Date.now() / 1000 + 60 * 30,
+  })
+}
+
+function throwIfAborted(signal: AbortSignal) {
```
estrattonbailey, line 290: Nit: I like this, could use in a few more places in these files, looks like, just for consistency

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+  signal.addEventListener('abort', abortOnCancel, {once: true})
+  let reader: ReturnType<typeof createChunkReader> | undefined
+  let parts: ReturnType<typeof planParts> = []
+  try {
+    try {
+      reader = createChunkReader(video)
+      parts = planParts(video.size, session.partSizeBytes)
+      if (parts.length !== session.partCount) {
+        throw new Error('Video service returned an invalid multipart plan')
+      }
```
estrattonbailey, line 73: Is this just defensive or an actual concern?

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  async function get(forceRefresh = false) {
+    if (!forceRefresh && token && Date.now() < expiresAt - 60_000) return token
+    if (!refresh) {
+      const exp = Math.floor(Date.now() / 1000) + 60 * 30
+      refresh = getServiceAuthToken({
+        agent,
+        lxm: 'com.atproto.repo.uploadBlob',
+        exp,
+      })
```
estrattonbailey: Might be good to wrap this with a retry too (if the error code is retryable). Looks like uploadParts retries, but if the last `get(true)` before finish throws, we fail the whole thing, iiuc

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  return {get}
+}
+
+function throwIfAborted(signal: AbortSignal) {
+  if (signal.aborted) throw new AbortError()
+}
+
+function delay(ms: number, signal: AbortSignal) {
```
estrattonbailey: Duplicate of the util in uploadParts :P

src/lib/media/video/multipart/uploadPart.ts
```diff
@@ -0,0 +1,94 @@
...
+        err instanceof MultipartUploadError &&
+        (err.status === 401 || err.error === 'AuthRequired')
+      ) {
+        args.onProgress(0)
+        return await sendPart(jobId, await getToken(true), args)
+      }
+      throw err
+    }
+  }
+}
```
estrattonbailey, line 24: Subsequent failures of `sendPart` here bubble up to `uploadPartWithRetry` and that's where the actual retry loop happens.

What do you think about hoisting this logic up to that retry loop so we can flatten out this curried method? We may also want to consider what status codes are retryable and which aren't, which would be better done at the `uploadPartWithRetry` level so we can exit that early if need be.

Retries on `getToken` also kinda good for this purpose, and may have different status code retry rules that make sense to handle at the token getter rather than here 🤔

src/lib/media/video/multipart/uploadParts.ts
```diff
@@ -0,0 +1,136 @@
...
+  signal,
+  concurrency = MULTIPART_CONCURRENCY,
+  maxAttempts = MULTIPART_MAX_ATTEMPTS,
+}: {
+  parts: PartPlan[]
+  reader: ChunkReader
+  uploadPart: UploadPartFn
+  totalBytes: number
+  setProgress: (progress: number) => void
+  signal: AbortSignal
```
estrattonbailey, line 33: Non-blocking nit: since there are multiple signals, might be nice to name the outer one like more specifically "user signal" or "composer signal" or something. Like everywhere it's used, so it's like `userSignal.aborted` and reads a little nicer throughout these files. Nbd.

estrattonbailey: Predates this and not at all related to this work but not a fan calling it just "AbortError" either, it's a known happy path exit case, reads weird like that across the composer files... anyway, just an aside

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+        setProgress,
+        signal,
+      })
+    } catch (err) {
+      if (signal.aborted) throw new AbortError()
+      return await abortThenFallbackOrResolve(
+        jobId,
+        await tokenProvider.get(),
+        err,
+      )
```
estrattonbailey, line 88: Just double check my understanding: this handle does double duty. Here it's used to clean up the backend state if something failed on the frontend.

Within `finishAndRecover`, the `if (completed)` block comes into play, since the finish call could have failed on the frontend but succeeded on the backend.

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+}: {
+  jobId: string
+  getToken: () => Promise<string>
+  signal: AbortSignal
+  resendMissingParts: (receivedPartNumbers: number[]) => Promise<boolean>
+}): Promise<AppBskyVideoDefs.JobStatus> {
+  let createdFailures = 0
+  while (true) {
+    throwIfAborted(signal)
+    const token = await getToken()
```
estrattonbailey: We just called `await tokenProvider.get(true)` before `finishAndRecover`, right? Should we just remove the first call and pass `true` here for the same functionality?

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+        await tokenProvider.get(),
+        err,
+      )
+    }
+
+    // Finish stores this credential for the later PDS blob upload, so use a
+    // fresh token rather than the one that may have aged during transfer.
+    await tokenProvider.get(true)
+    const activeReader = reader
+    if (!activeReader) throw new Error('Video chunk reader is unavailable')
```
estrattonbailey, line 93: Nit: if this is just for types, lil comment would be rad

src/lib/media/video/multipart/upload.ts
estrattonbailey: Also might be good here so that retries result in a fresh token 🤔 or maybe we want to keep the same token for each retry?

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,281 @@
...
+              createdFailures = 0
+              continue
+            }
+          } catch (err) {
+            throwIfAborted(signal)
+            return await abortThenFallbackOrResolve(jobId, token, err)
+          }
+          createdFailures++
+          if (createdFailures < MULTIPART_FINISH_ATTEMPTS) {
+            await delay(500 * 2 ** (createdFailures - 1), signal)
```
estrattonbailey, line 165: Could parallel token refresh during this backoff

estrattonbailey (commented):
Looking good! Haven't tested yet, will do that now

estrattonbailey (commented):
Looking good to me on web/ios. Android env is borked, will get it tested too.

Edit: Android looks good too!

src/lib/media/video/multipart/uploadParts.ts
estrattonbailey: Not really worth digging into here, was just noting

src/lib/media/video/multipart/upload.ts
estrattonbailey: If we want a fresh token each attempt at finishing, that is

src/lib/media/video/multipart/upload.ts
vineyardbovines: Right. Abort is both cleanup and the gate that makes fallback safe. If the backend says the reservation was aborted, we can use the legacy path. If finish actually succeeded before the client observed the failure, abort reports completed and we return that completed job instead of uploading twice.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Reworded this to be more direct: the service may already have assembled the upload even though the finish request failed, so we poll and retry instead of starting a second upload.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Done in the earlier feedback commit. Status lookup now goes through getUploadStatusWithRetry, with abort-aware exponential backoff for network and retryable service errors.

src/lib/media/video/multipart/upload.ts
vineyardbovines: This is guarding an actual protocol invariant rather than only a type concern. The client plans parts from the returned part size, so a mismatched count means the server plan is internally inconsistent and we should stop before sending data.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Yep, there is overlap. I pulled the shared HTTP status check out of the existing retry helper and use it here now, including 429. This also avoids treating every 5xx response as retryable.

src/lib/media/video/multipart/upload.ts
vineyardbovines: I chose to keep the same token across finish retries. Finish stores that credential for the later PDS blob upload, so once we submit it, polling/recovery should stay tied to that credential rather than refreshing on every attempt. The initial finish attempt still forces a fresh token.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Done. I extracted the abort-aware delay into a shared multipart utility and use it from both finish/status recovery and part retries.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Yes, exactly. During ordinary frontend failure it cleans up the backend reservation and enables safe legacy fallback. If finish succeeded server-side but the response was lost, the completed result prevents fallback from creating a duplicate upload.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Added a comment explaining that the alias preserves TypeScript narrowing inside the recovery callback.

src/lib/media/video/multipart/uploadPart.ts
vineyardbovines: I moved the retryability decision into uploadPartWithRetry, so 4xx responses now fail immediately while network errors, the shared transient HTTP statuses (including 429), and ServiceOverloaded retry. I kept the single 401/AuthRequired refresh next to sendPart because that layer owns the token provider; token acquisition itself also has its own retry rules.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Fixed now by extracting the abort-aware delay into a shared multipart utility.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Yep. I removed the separate pre-finish refresh and moved forceRefresh=true into the first finishAndRecover iteration, so there is only one token lookup there.

src/lib/media/video/multipart/upload.ts
estrattonbailey: Lovely thanks!

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,302 @@
...
+
+function isRetryableStatusError(err: unknown) {
+  return (
+    err instanceof TypeError ||
+    (err instanceof MultipartUploadError &&
+      (err.error === 'ServiceOverloaded' ||
+        err.status === undefined ||
+        isRetryableHttpStatus(err.status)))
+  )
+}
```
estrattonbailey: This is also duplicated now

src/lib/media/video/multipart/upload.ts
```diff
@@ -0,0 +1,302 @@
...
+          createdFailures++
+          if (createdFailures < MULTIPART_FINISH_ATTEMPTS) {
+            await delay(500 * 2 ** (createdFailures - 1), signal)
+            continue
+          }
+          return await abortThenFallbackOrResolve(jobId, token, finishError)
+        case 'finishing':
+          // The service may have assembled the upload even though the finish
+          // request failed. Poll and retry instead of starting a second upload.
+          await delay(1000, signal)
```
estrattonbailey, line 172: Reasonable shout from Claude, also reasonable to leave as is. Just making sure you saw it.

> 'finishing' branch has no ceiling — upload.ts:169-173 polls every 1s indefinitely, exited only by user abort. The author's rationale (finalization owns the reservation, so legacy fallback would be unsafe) is sound, and this was consciously left as-is. Flagging only as the one path that can spin against a stuck server with the composer open; a high max-attempts cap surfacing a "still processing, try again" error would bound the worst case without giving up idempotent retry. Author's call.

estrattonbailey (approved):
LGTM, one more nit and a question for ya

src/lib/media/video/multipart/upload.ts
vineyardbovines: Fixed in 33d387a. I moved the shared network/multipart-status retry predicate into multipart/utils.ts and both status polling and part uploads now use it.

src/lib/media/video/multipart/upload.ts
vineyardbovines: Yep, saw it. I’m leaving this intentionally unbounded: once the service owns finalization, timing out locally cannot safely fall back or release that reservation, and a retry-later error would leave the composer without a reliable resume path. User cancellation still aborts the client-side polling. We can add a resumable/background finalization flow separately if needed.

PR #11226, opened 2026-07-21
Add new search results tab for starter packs

ds-boyce:
search

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "prettier": "prettier --check ."
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


estrattonbailey (approved):
Hell yes!

Small nit: can we nix this border at the end of results or nah? Hardly noticeable, just figured I'd check.

Image

ds-boyce:
Ah, looks like it’s part of `ListFooter` and I’m not sure how far-reaching that change would be at the moment.

PR #11227, opened 2026-07-21
Update presentation of trending topics in Explore

ds-boyce:
| Before | After |
| - | - |
| before | after |

### DIFF
--- a/src/screens/Search/components/ModuleHeader.tsx
+++ b/src/screens/Search/components/ModuleHeader.tsx
@@ -31,7 +31,7 @@ export function Container({
         a.px_lg,
         a.pt_2xl,
         a.pb_md,
-        a.gap_sm,
+        a.gap_xs,
         t.atoms.bg,
         bottomBorder && [a.border_b, t.atoms.border_contrast_low],
         style,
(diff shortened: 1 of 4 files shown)


src/screens/Search/components/ModuleHeader.tsx
```diff
@@ -82,19 +82,21 @@ export function Icon({
...
+      <Comp width={iconSize} fill={t.atoms.text.color} />
     </View>
   )
 }
 
 export function TitleText({style, ...props}: TextProps) {
   return (
     <Text
-      style={[a.font_semi_bold, a.flex_1, a.text_xl, style]}
+      style={[a.font_medium, a.flex_1, a.text_lg, a.leading_snug, style]}
```
estrattonbailey: Medium seems a little light. Looks like designs call for semi bold still

src/screens/Search/modules/ExploreInterestsCard.tsx
```diff
@@ -63,8 +57,8 @@ export function ExploreInterestsCard() {
             t.atoms.border_contrast_medium,
           ]}>
           <View style={[a.flex_row, a.gap_sm, a.align_center]}>
-            <Shapes />
-            <Text style={[a.text_xl, a.font_semi_bold, a.leading_tight]}>
+            <ShapesIcon fill={t.atoms.text.color} />
+            <Text style={[a.text_lg, a.font_medium, a.leading_snug]}>
```
estrattonbailey: Maybe another semi to restore?

estrattonbailey (approved):
LGTM! Just maybe some font tweaks, unless you have insider info from the design team :P

src/screens/Search/components/ModuleHeader.tsx
ds-boyce: It’s actually normal in the Figma, but that felt a little light to me. I’ll check with the design team.

ds-boyce: Oh wait, layer properties says “Medium/Semi” but inspector says `font-weight: 400`. Strange. I’ll go with the layer properties.

mozzius:
Would be nice to ensure the skeleton is a matching size if possible. currently there's a fair bit of layout shift

PR #11230, opened 2026-07-22
Potential fix to silent failures when un/subscribing to labellers past ~25

RetroSunstar:
When a user has ~25 labellers associated with their account, several parts of the labeller subscription flow silently fail:

* Unsubscribing from a labeller does nothing
* The 'max labellers reached' warning fails to show
* Subscribing to new labellers is also blocked, though largely irrelevant since Bluesky already enforces a 20-labeller limit

Under normal circumstances, this may not be an issue 

But with Bluesky's additional moderation authorities, we're already pushing exactly 25 in a worst case where a country code could not be obtained. Only because the other 6 additional services are not yet declared labellers and are regular profiles. 

Theoretically if anything changed in this equation, this bug would surface naturally. In which case, we shouldn't have a bug that prevents a user from removing labellers, or one that doesn't inform them of the limits at hand

## Foot note

I'm not familiar with this area, so take it more as a demonstration piece, but I've tested it up to 40+ labellers and I can't see any particular issues in the client itself

Any other platform would be able to ignore the limits at hand, allowing users to exceed the max count, and cause this bug to show in Bluesky

This should also reduce the only real hard limitation with #11180 🐙


src/state/queries/labeler.ts
```diff
@@ -119,10 +119,18 @@ export function useLabelerSubscriptionMutation() {
...
       const invalidLabelers: string[] = []
       if (labelerDids.length) {
-        const profiles = await agent.getProfiles({actors: labelerDids})
-        if (profiles.data) {
+        try {
+          const CHUNK_SIZE = 25
+          const chunks: string[][] = []
+          for (let i = 0; i < labelerDids.length; i += CHUNK_SIZE) {
+            chunks.push(labelerDids.slice(i, i + CHUNK_SIZE))
+          }
```
mozzius: `@atproto/common-web` has a chunk util

RetroSunstar: Oh, perfect! Swapped out and hopefully that's better? 🐙

PR #11237, opened 2026-07-22
Sort follows and following behind gate

ds-boyce:
Sorts follows/following to `"top"` when viewing someone’s profile and defaults to `"latest"` otherwise (explicitly `"latest"` for your own profile).

This is gated and a new `"sort"` param has been added to relevant metrics events.


ds-boyce:
Need a feature gate here.

vineyardbovines (approved):

PR #11248, opened 2026-07-23
APP-1645: Improve video upload progress indicator

vineyardbovines:
## Summary

- present compression, upload, and backend processing as one monotonic progress timeline
- map compression to 0–30%, upload to 30–50%, and backend `getJobStatus.progress` to 50–100%
- prevent multipart retries and legacy fallback from moving progress backwards
- retain phase-specific labels and use indeterminate rotation only when backend progress is unavailable

### DIFF
--- a/src/view/com/composer/Composer.tsx
+++ b/src/view/com/composer/Composer.tsx
@@ -2672,9 +2672,14 @@ function VideoUploadToolbar({state}: {state: VideoState}) {
   const t = useTheme()
   const {t: l} = useLingui()
   const progress = state.progress
+  const processingProgress =
+    state.status === 'processing' ? state.jobStatus?.progress : undefined
   const shouldRotate =
-    state.status === 'processing' && (progress === 0 || progress === 1)
-  let wheelProgress = shouldRotate ? 0.33 : progress
+    state.status === 'processing' &&
+    (processingProgress === undefined ||
+      processingProgress <= 0 ||
+      processingProgress >= 100)
+  let wheelProgress = progress
 
   const rotate = useDerivedValue(() => {
     if (shouldRotate) {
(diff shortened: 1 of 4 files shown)


src/view/com/composer/Composer.tsx
```diff
@@ -2694,7 +2696,7 @@ function VideoUploadToolbar({state}: {state: VideoState}) {
       break
     case 'error':
       text = l`Error`
-      wheelProgress = 100
+      wheelProgress = 1
```
mozzius, line 2708: I think the idea here is that's it's a red circle

mozzius (changes requested):
I tried it out. It seems to start at a fixed 30%, I guess because it skipped compression? Maybe if there's no compression needed, remove it from the calculation?

And when the backend doesn't return progress it spins momentarily which is way more jarring now, perhaps get rid of the spinning and just keep where it's at?

https://github.com/user-attachments/assets/12bf4618-0830-4f0f-9c16-2962e20a9475

vineyardbovines:
Addressed in a2fd0fa5ef68:

- skipped compression now gives upload the full 0–50% range, so progress starts at 0 instead of 30%
- processing now holds its current progress until the backend reports a percentage; the transient spinner is removed
- the thumbnail compression overlay converts global progress back to phase-local progress, so it still fills from 0–100%

Also confirmed the error state remains a full red circle: react-native-progress uses a 0–1 range, so progress=1 is the corrected equivalent of the previous progress=100.

mozzius:
@vineyardbovines needs a rebase

mozzius (approved):
Def would be nice to spend half a sec in a "completed" state, would be more satisfying. But big improvement regardless!

mozzius:
I do wonder if it would be nicer to fake 50% -> 60% as a 15 second steady increase. It does wait a long time at 50% as it waits for the job to start processing

PR #11254, opened 2026-07-23
Fix issue where form controls appear above sticky headers

ds-boyce:
| Before | After |
| - | - |
| before | after |


mozzius:
Is it possible to fix this without a RN patch? would like to avoid that if possible. For example, setting smaller zIndex values on our text fields

estrattonbailey (approved):

PR #11260, opened 2026-07-24
Lingui v5 - rip off bandaid, migrate every file with codemod

mozzius:
Ran codemod then manually fixed the broken files. I had to update oxlint to use the autofix of the `import/no-duplicates` rule

Have not tested particularly extensively but typecheck passes


ds-boyce (approved):
In codemod we trust (stamp).

PR #11276, opened 2026-07-25
Fix inline profile badge alignment in notifications on iOS new arch

mozzius:
The verification/bot/beta badges rendered inline after the author name in a
notification row ("liked by xyz") sat too high on iOS after the new
architecture switch.

The badges are a `View` nested inside a `<Text>`, so iOS lays them out as a
text attachment. The old architecture measured the attachment box with
`sizeThatFitsMinimumSize:` (which folds the inline view's margins into the box)
and positioned it at `lineFragmentBottom - boxHeight + font.descender`, so the
`marginBottom: -6` hack shifted the badges down. On the new architecture
`ParagraphShadowNode` measures the box with `LayoutableShadowNode::measure`,
which returns the node's own frame size with margins excluded, and then
overwrites the box origin with the frame `RCTTextLayoutManager` computed at
`lineTop + baseline - boxHeight`. Margins and `top` are both ignored, leaving
the badges pinned to the text baseline.

Box height is the only remaining lever, so add an `inlineFontSize` prop to
`ProfileBadges` that constrains the container to the cap height of the
surrounding text. The box then covers exactly the capital letters and the
existing `align_center` centers the badges on them, regardless of which badges
are visible or how tall they are.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Pgeau3P52qsMTcnDeJm8Zn

### DIFF
--- a/src/components/ProfileBadges.tsx
+++ b/src/components/ProfileBadges.tsx
@@ -8,10 +8,18 @@ import {BotBadge, BotBadgeButton, isBotAccount} from '#/components/BotBadge'
 import {useSimpleVerificationState} from '#/components/verification'
 import {VerificationCheck} from '#/components/verification/VerificationCheck'
 import {VerificationCheckButton} from '#/components/verification/VerificationCheckButton'
+import {IS_IOS} from '#/env'
 import type * as bsky from '#/types/bsky'
 import {BetaBadge, BetaBadgeButton, useIsBetaBadgeVisible} from './BetaBadge'
 
 export type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl'
+
+/*
+ * Cap height of Inter as a fraction of its em size, from the font's OS/2
+ * table. The system font (used when the "system font" setting is on) is within
+ * 0.015em of this, so a single constant covers both.
+ */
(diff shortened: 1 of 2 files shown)


src/components/ProfileBadges.tsx
```diff
@@ -8,11 +8,19 @@ import {BotBadge, BotBadgeButton, isBotAccount} from '#/components/BotBadge'
...
 import {BetaBadge, BetaBadgeButton, useIsBetaBadgeVisible} from './BetaBadge'
 
 export type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl'
 
+/*
+ * Cap height of Inter as a fraction of its em size, from the font's OS/2
+ * table. The system font (used when the "system font" setting is on) is within
+ * 0.015em of this, so a single constant covers both.
+ */
+const CAP_HEIGHT_RATIO = 0.7275
```
vineyardbovines, line 22: ugh lol

vineyardbovines (approved):

mozzius (commented):
This is kinda gross, will put in draft until I can verify properly

PR #11277, opened 2026-07-25
fix: prevent boot crash when transition_animation_scale setting is malformed

mozzius:
## Why

Fixes a fatal Android boot crash ([APP-T2EW](https://blueskyweb.sentry.io/issues/APP-T2EW)):

```
Error: Call to function 'ExpoPlatformInfo.getIsReducedMotionEnabled' has been rejected.
Caused by: java.lang.NumberFormatException: For input string: "STRING_FROM_SETTINGS_GLOBAL_WITH_KEY_transition_animation_scale"
```

On devices with a broken settings provider (seen on Waydroid; possible on odd OEM ROMs), `Settings.Global.TRANSITION_ANIMATION_SCALE` returns a non-numeric string. The existing Kotlin guard was `catch (_: Error)`, but `NumberFormatException` is an `Exception`, so it escaped and rejected the sync module call. Since `src/state/persisted/schema.ts` calls `getIsReducedMotionEnabled()` at module scope during boot, the rejection is an unhandled fatal and the app cannot start on affected devices.

## How

Two layers of hardening:

1. **Kotlin**: replace `toFloat()` + broken catch with `toFloatOrNull() == 0f`, which never throws.
2. **JS wrapper** (`PlatformInfo/index.native.ts`): wrap the native call in try/catch returning `false`, so any future native failure of this call degrades gracefully instead of crashing boot.

Note: reanimated's built-in `useReducedMotion` has the identical unguarded `Float.parseFloat` on Android (`NativeProxy.java:520` in v4.3.2), so migrating to it was not an option here - worth an upstream PR separately.

### DIFF
--- a/modules/expo-bluesky-swiss-army/android/src/main/java/expo/modules/blueskyswissarmy/platforminfo/ExpoPlatformInfoModule.kt
+++ b/modules/expo-bluesky-swiss-army/android/src/main/java/expo/modules/blueskyswissarmy/platforminfo/ExpoPlatformInfoModule.kt
@@ -13,12 +13,7 @@ class ExpoPlatformInfoModule : Module() {
       Function("getIsReducedMotionEnabled") {
         val resolver = appContext.reactContext?.contentResolver ?: return@Function false
         val scale = Settings.Global.getString(resolver, Settings.Global.TRANSITION_ANIMATION_SCALE) ?: return@Function false
-
-        try {
-          return@Function scale.toFloat() == 0f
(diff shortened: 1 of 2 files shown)


vineyardbovines (commented):
do we know if this gets resolved in updating RN / Expo on the next go-around?

mozzius:
No, reanimated has an equivalent function that would suffer from the same crash

This is caused by a shitty linux shim providing something other than a number for this API

PR #11278, opened 2026-07-25
fix: stop grouping all signup failures into one sentry issue

mozzius:
Fixes APP-S3WR

[APP-S3WR](https://blueskyweb.sentry.io/issues/6706737577/) ("Signup Flow Error") has 511k events / 100k users because four call sites log the same constant string at error level, so Sentry groups every possible signup failure – expected user errors, user behavior, and genuine bugs – into one unactionable issue. It also sends PII (the raw server error containing the user's email, plus the registration handle) to Sentry.

## Changes

**`useSubmitSignup` catch block** (`src/screens/Signup/state.ts`)
- Expected failures (invalid/taken handle, invalid password, unsupported domain, email already taken, network errors) are now counted via a new `signup:createAccountFailure` metric with a PII-free `reason`, instead of being reported to Sentry.
- Unexpected failures go to Sentry as `Signup: unexpected createAccount failure` with `safeMessage`.
- Classification runs on the **raw** error (typed `ComAtprotoServerCreateAccount` errors + string match for email-taken, which has no typed error). `cleanError` output can't be used for this since it's localized; it's still used for the user-facing error display, which is unchanged.

**Captcha step** (`src/screens/Signup/StepCaptcha/index.tsx`)
- Webview error → own Sentry issue: `Signup: captcha response error`.
- Back-press on the captcha step is user behavior, not an error – replaced `logger.error` with a `signup:captchaBackPress` metric, matching the neighboring captcha metrics.

**Other**
- The captcha-code-missing invariant in `state.ts` keeps error level but gets its own message.
- No more `errorMessage`/`registrationHandle` attached to any of these events.
- New metric events registered in `src/analytics/metrics/types.ts`.

## Test plan

- `pnpm typecheck`, `pnpm lint`, `pnpm prettier` all pass. No existing tests cover these files.
- User-facing behavior is intentionally unchanged: all `setError`/`setStep`/`isLoading` dispatches, including the `isHandleError` routing, are untouched.

### DIFF
--- a/src/analytics/metrics/types.ts
+++ b/src/analytics/metrics/types.ts
@@ -103,6 +103,10 @@ export type Events = {
   }
   'signup:captchaSuccess': {}
   'signup:captchaFailure': {}
+  'signup:captchaBackPress': {}
+  'signup:createAccountFailure': {
+    reason: string
+  }
   'signup:fieldError': {
     field: string
     errorCount: number
--- a/src/screens/Signup/StepCaptcha/index.tsx
+++ b/src/screens/Signup/StepCaptcha/index.tsx
@@ -127,23 +127,17 @@ function StepCaptchaInner({
         value: _(msg`Error receiving captcha response.`),
       })
       ax.metric('signup:captchaFailure', {})
-      logger.error('Signup Flow Error', {
-        registrationHandle: state.handle,
-        error,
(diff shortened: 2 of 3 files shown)


vineyardbovines (approved):
yes

Issue #8770, opened 2025-08-02
Unwanted auto-signout every few weeks


axelfontaine:
### Steps to Reproduce

Every few weeks I get signed out of all my Bluesky accounts in the Android app. No such issue on desktop web. It is very annoying, especially with the tedious 2FA sign in process.

Please fix this or allow me to trust my own device forever.

### What platform(s) does this occur on?

Android

### Device Info

Pixel 7 Pro / Android 16

### What version of the app are you using?

1.105.0.679

### Additional Information

I have two Bluesky accounts and both get signed out. I can still choose the usernames to sign in again, but the session is definitely gone.

antoniovj1:
I'm having the same issue, but it's not related to the current version — it's been happening for a long time. I'm also using 2FA.

### Device
Pixel 9 Pro XL / Android 16

### Version
1.105.0.679

djeikyb:
on desktop (macos, safari 18.4), it's a signout every few hours

[hopefully related?](https://github.com/bluesky-social/social-app/issues/8695)

ribcakes:
I am also having the same problem with the android app. First on a Pixel 6 Pro and now on my Pixel 10 Pro. The earliest occurrence was May 19th, and the most recent on Nov 11th. I unfortunately do not have a list of all the app versions I've been using over that time span, but I'm currently on 1.110.1 on Android 16.

DanielToft:
I am also experiencing this issue on my Pixel 10 Pro. I have to log in every day.

FlareHeart:
I am also getting signed out approximately every 48 hours. 

If there are logs I can provide, I would be happy to. 

Pixel 8 Pro - GrapheneOS (Android 16)

antoniovj1:
> I'm having the same issue, but it's not related to the current version — it's been happening for a long time. I'm also using 2FA.
> 
> ### Device
> Pixel 9 Pro XL / Android 16
> 
> ### Version
> 1.105.0.679

Still happening with V1.110.1.741

pancake-tongs:
I'm having this problem too, sometimes more than once per day. I'm on a pixel 9 pro, with grapheneOS

Bandit:
This issue has been happening to me for over a year, resulting in me not using Bluesky anymore. Sad but true.

FWIW this issue doesn't happen in the Graysky Android app - I've _never_ had to relogin; but it's a piece of crap so I don't want to use it.

Here's a support reply from 7 months ago on the issue:

[...]

PR #11281, opened 2026-07-26
Move sessions to secure storage

mozzius:
## Summary

- move session persistence behind a session-owned repository instead of the global persisted-state blob
- store native credentials synchronously in Expo SecureStore using split, SHA-256-keyed records
- migrate and validate legacy sessions before scrubbing tokens, with retry, corruption recovery, logout journaling, and storage-full telemetry
- retain web localStorage and BroadcastChannel synchronization in a dedicated session store

Addresses #8770.

## Test plan

- Test login persistence across all platforms
- Ensure old sessions are migrated cleanly

Roast review completed; all actionable findings were addressed.

### DIFF
--- a/app.config.js
+++ b/app.config.js
@@ -240,6 +240,7 @@ module.exports = function (_config) {
       plugins: [
         'expo-video',
         'expo-localization',
+        'expo-secure-store',
         'expo-web-browser',
         [
           'react-native-edge-to-edge',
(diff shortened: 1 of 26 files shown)


vineyardbovines:
> Scenario: native user with a full disk (or a keychain write that keeps failing) cold-launches the app → blank screen indefinitely, no messaging.

i actually think this is worth addressing, i can absolutely see this happening on a lower device class

mozzius:
Superseded by #11483, which reimplements this on top of the SDK-migrated session core with a dual-write + gated-boot-read design (and resolves the open blank-screen-on-keychain-failure finding: boot init is now synchronous, non-throwing, and always falls back to the legacy blob).

PR #11283, opened 2026-07-26
fix: prevent blank flash on mentions tab when there are no mentions

tayyabjamil:
Noticed the mentions tab flickers with a blank screen when you have no mentions.
Happens every time you switch to the tab.

Traced it back to the `else if (data)` branch — when `isFetching` is true,
`isEmpty` stays false so we fall into that branch and loop through the pages.
But if you have zero mentions every page comes back empty, so `arr` ends up as
`[]` and we're passing an empty array to List which causes the flash.

Simple fix — if `arr` is still empty after the loop just show the empty state
instead. Doesn't affect anyone who actually has mentions since their `arr` would
have items and never hit that check.

### DIFF
--- a/src/view/com/notifications/NotificationFeed.tsx
+++ b/src/view/com/notifications/NotificationFeed.tsx
@@ -62,13 +62,7 @@ export function NotificationFeed({
     enabled: enabled && !!moderationOpts,
     filter,
   })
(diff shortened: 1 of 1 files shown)


estrattonbailey:
@claude whatcha think

PR #11289, opened 2026-07-26
Enter build numbers manually for production OTA deploys

mozzius:
Production OTA updates are bound to the specific native build they
target, but the deploy workflow read build numbers from the global EAS
counters, which advance with every testflight build. Deploying an OTA
to an older production release therefore required manually resetting
the global counter first.

Add iosBuildNumber and androidVersionCode workflow_dispatch inputs to
the EAS update workflow, validate that both are provided (numeric) when
the channel is production, and pass them into the deploy steps. The
use-build-number wrapper now respects build numbers already present in
the environment and only falls back to the EAS counters when they are
unset, so testflight and PR preview deploys are unchanged.

Also fix the runtimeVersion validation step, whose inverted emptiness
check meant the format regex never actually ran.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MGjPfvP5nRMWr53bVvb7SZ

### DIFF
--- a/.github/workflows/bundle-deploy-eas-update.yml
+++ b/.github/workflows/bundle-deploy-eas-update.yml
@@ -17,6 +17,14 @@ on:
         type: string
         description: Runtime version (in x.x.x format) that this update is for
         required: true
+      iosBuildNumber:
+        type: string
+        description: iOS build number of the native build this update targets
+          (required for production)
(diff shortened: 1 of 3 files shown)


vineyardbovines (approved):
yessss

PR #11293, opened 2026-07-27
Fix IllegalStateException: Required value was null.

tomekzaw:
## Summary

Extends our `react-native@0.81.5` patch to fix the fatal Android crash [APP-T20Q](https://blueskyweb.sentry.io/issues/APP-T20Q) (15 events / 14 users over 14 days on 1.128.0/1.129.0, new arch, wide device spread):

```
IllegalStateException: Required value was null.
  at com.facebook.react.views.view.ReactViewGroup.updateSubviewClipStatus(ReactViewGroup.kt:469)
  at com.facebook.react.views.view.ReactViewGroup.updateClippingToRect(ReactViewGroup.kt:435)
  at com.facebook.react.views.view.ReactViewGroup.updateClippingRect(ReactViewGroup.kt:411)
  at com.facebook.react.views.scroll.ReactScrollView.updateClippingRect(ReactScrollView.java:604)
  ...
  at com.facebook.react.views.scroll.ReactScrollView.onScrollChanged(ReactScrollView.java:472)
  at android.view.View.setScrollY(View.java:19296)  (ObjectAnimator - animated smooth scroll)
```

**Root cause:** during a `removeClippedSubviews` clipping pass (triggered here by animated smooth scrolls via `ReactScrollView.onScrollChanged`), clipping a view can synchronously trigger reentrant child removal (layout-change listeners, animation-end callbacks, Fabric mounting on the UI thread). That compacts `ReactViewGroup.allChildren` mid-loop and leaves a `null` slot below `allChildrenCount`; the `checkNotNull(allChildren?.get(idx))` in `updateSubviewClipStatus` then throws `IllegalStateException: Required value was null.` Upstream already catches the `IndexOutOfBoundsException` flavor of this corruption with diagnostics, but the null-child flavor escapes as a fatal crash.

**Fix:** two null-guards in `ReactViewGroup.kt` — a null slot means the view is already detached, so skip it and count it as clipped (keeps the `clippedSoFar` index math aligned) instead of throwing.

**Upstream status:** unfixed as of July 2026 (identical `checkNotNull` on `main`, zero matching issues); the sibling fix attempt for the same bookkeeping corruption ([facebook/react-native#57365](https://github.com/facebook/react-native/pull/57365), different stack via `onLayoutChange`) was abandoned. Worth filing upstream separately so the hunk can eventually be dropped.

## Where the fix applies

Production Android builds compile react-android from source (`buildReactNativeFromSource: IS_PRODUCTION` via expo-build-properties injects the `includeBuild` dependency-substitution block at prebuild), **so this fix is active in production releases once merged.** Local dev builds prebuilt in a non-production env consume the prebuilt AAR from Maven Central and do not pick up ReactAndroid source patches — noted in the patch md so nobody is surprised in a debug build. (iOS hunks in this patch file are always active since CocoaPods builds RN from source.)

## Verification

Verified with a deterministic state-injection repro on a Pixel emulator, with react-android built from source using the same dependency-substitution mechanism production uses (temporary volume-down hook: find an on-screen `ReactViewGroup`, enable `removeClippedSubviews`, detach the last child via `removeViewInLayout` + reflectively null its `allChildren` slot - the exact corrupt state from the crash reports - then call `updateClippingRect()`):

- **Unpatched control:** fatal `IllegalStateException: Required value was null.` with the byte-for-byte Sentry stack (`ReactViewGroup.kt:469` → `435` → `411`).
- **Patched:** logs `SURVIVED clipping pass`, no exception, app remains fully functional (feed scrolls and renders normally).

Also attempted an organic repro on a fully stock build (80 automated cycles of deep feed flings + animated scroll-to-top): no crash, consistent with the ~1/day production rate being a rare timing race — hence the state-injection approach for verification.

The `pnpm-lock.yaml` diff is purely the `patch_hash` propagation from `pnpm patch-commit`.

Since the fix ships with the next production release after merge, the Sentry issue can be resolved once that release is out (deliberately not using the auto-close keyword so it does not resolve at merge time).


mozzius (approved):
LGTM - we can try it out

PR #11298, opened 2026-07-27
Reenable wake-from-background Android OTA

mozzius:
No longer crashes (confirmed when originally reproduced)


vineyardbovines (approved):
sick

PR #11301, opened 2026-07-27
Fix Android pager crash during screen transitions

mozzius:
## Summary

- patch react-native-screens so screen removal transitions treat RecyclerView as one unit
- avoid marking recyclable RecyclerView children as transitioning
- keep the matching end-transition traversal symmetric

## Why

react-native-screens recursively calls startViewTransition on every descendant when a screen is removed. Android deliberately keeps a transitioning child parented after removal so it can continue drawing until endViewTransition.

If a ViewPager2 is still settling, its RecyclerView can remove a page holder and immediately try to recycle it. Because react-native-screens marked that holder as transitioning, Android leaves its parent set and RecyclerView crashes with Scrapped or attached views may not be recycled.

This also explains why the crash occurs specifically when pressing back during a swipe and why the pager had not begun Fabric teardown in the diagnostic logs.

The v5 alpha retains the same behavior in its legacy compatibility stack. Its new Stack.Host / Stack.Screen implementation does not recursively transition descendants.

Related:
- https://github.com/callstack/react-native-pager-view/issues/1005
- https://github.com/software-mansion/react-native-screens/issues/2461
- https://blueskyweb.sentry.io/issues/7612736192/

## Testing

- reproduced before the patch by swiping a pager and pressing Android back while it was moving
- confirmed the crash no longer reproduces with the patch
- compiled react-native-screens and react-native-pager-view debug Kotlin targets
- pnpm frozen-lockfile validation passes

### DIFF
--- a/patches/react-native-screens@4.24.0.patch
+++ b/patches/react-native-screens@4.24.0.patch
@@ -1,0 +1,33 @@
+diff --git a/android/src/main/java/com/swmansion/rnscreens/Screen.kt b/android/src/main/java/com/swmansion/rnscreens/Screen.kt
+index c99ca362df5baee81fa48d1c9ef09b40a2c23f07..725c07aa0ed516f9dd09f81e522f2d3191d9b5d0 100644
+--- a/android/src/main/java/com/swmansion/rnscreens/Screen.kt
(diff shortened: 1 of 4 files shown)


tomekzaw (commented):
LGTM

tomekzaw (approved):

